PR #840, opened 2016-05-20
systemd, README: Edit to account for the recent spike in reports

wmark:
Condenses the README file a bit, stresses that you need a up-to-date *systemd*, and highlights how to debug a failed start of *caddy* using system logs.


faddat:
Hm.  I bet that I need to update systemd, then.... #841 

Thanks!

mholt:
Thanks!

Issue #854, opened 2016-05-28
proxy_header Host doesn't appear to work


kaihendry:
#### 1. What version of Caddy are you running (`caddy -version`)?

238250e

#### 2. What are you trying to do?

Reverse proxy to a site with a Host: defined

#### 3. What is your entire Caddyfile?
```text
localhost
proxy / testserver:8080 {
    proxy_header Host foobar.com
}
```

#### 4. How did you run Caddy (give the full command and describe the execution environment)?

`caddy`

```
$ hostnamectl
   Static hostname: teefour
         Icon name: computer-laptop
           Chassis: laptop
        Machine ID: 08fa83b18a364e11aa225f5b23ff3a19
           Boot ID: 1460ab922a064e10b811bf38c03282bd
  Operating System: Arch Linux
            Kernel: Linux 4.5.1-1-t460s
      Architecture: x86-64
```

#### 5. What did you expect to see?

Request on _testsever_ with **Host: foobar.com**

#### 6. What did you see instead (give full error messages and/or log)?

Wrong site server since **Host: localhost:2015** is sent over

#### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

http://s.natalian.org/2016-05-28/debug.go is how I setup _testserver_

Btw this is how I've done a reverse proxy with golang in my own tests: https://github.com/kaihendry/go-rproxy-test/blob/master/main.go which works well.

abiosoft:
I've reproduced this. Working on a fix.

abiosoft:
Fixed in https://github.com/mholt/caddy/commit/a0e6eb3ba950e94eb0d061c0c7697d3f8005f975

mholt:
Oops. Thanks :smile: I think I broke this a few weeks ago when fixing proxy for something else.

wesm-outreach:
Where is the docs for `proxy_header`? I don't see that here: https://caddyserver.com/docs/proxy

francislavoie:
@wesm-outreach it's now called `header_upstream` and `header_downstream`. This issue is from nearly two years ago, things have changed a bunch since then.

wesm-outreach:
@francislavoie ya they have! Lol. I was doing some google searches around converting from nginx to Caddy, and this issue popped up. I didn't quite equate the `header_downstream` from the docs to my use case until now. Thanks for the reply! Sorry for resurrecting an old issue.

PR #855, opened 2016-05-29
Fix for #854

abiosoft:
The bug was introduced in e2234497b79603388b58ba226abb157aa4aaf065

### DIFF
--- a/middleware/proxy/proxy.go
+++ b/middleware/proxy/proxy.go
@@ -107,7 +107,7 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 					outreq.Host = rHost
 				}
 				if v, ok := host.UpstreamHeaders["Host"]; ok {
-					r.Host = replacer.Replace(v[len(v)-1])
+					outreq.Host = replacer.Replace(v[len(v)-1])
 				}
 				// Modify headers for request that will be sent to the upstream host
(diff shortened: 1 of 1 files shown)


abiosoft:
Pretty basic. Merging.

kaihendry:
Just tested it after a `go get -u github.com/mholt/caddy` and it now works as expected!

Thank you!

Now it would be good if the Docker images updated with the fix. ;)

abiosoft:
@mholt any release in sights ?

It'll make it to the docker image after the next release.

mholt:
Seeing the fix, I think I might have caused this when I fixed a different bug in proxy related to headers.

So, thanks for finding this and making it right. :smile:

The next release is slated for Caddy 0.9, I'm working vigorously to get the beta versions out as soon as possible!

kaihendry:
I was wondering if a test could be made for this?

Or is it too complex to run  another server to catch Host: in a test?

mholt:
A test can definitely be written for this. It's on my list but I'd be happy to accept PRs on that!

kaihendry:
I don't quite understand how testing works in Caddy. You recurse through directories running `go test`?
http://s.natalian.org/2016-06-01/1464743896_2548x1380.png

mholt:
go test ./... gets all subdirectories too.

kaihendry:
Cool. Didn't know about that syntax.
http://s.natalian.org/2016-06-01/1464745956_2548x1380.png

PR #859, opened 2016-06-01
Fix for cleaned URL.Path

### LINKED ISSUE #643: Encoded slash in URL
I want to address a resource with the name `/`. So rabbitmq api requests look like this: `/api/queues/%2F/queue.fetchtasks`. I can't proxy this requests with caddy, because caddy modify the url path.

My simple Caddyfile:

```
localhost:8888 {
  # proxy to "nc -l localhost 9999"
  proxy /test http://localhost:9999
}
```

If I run `curl localhost:8888/test/%2F/foo`, then caddy decodes the path before proxy (`%2F` to `/`). This results in `/test///foo` but it should be `/test/%2F/foo`. For apache there is an option [AllowEncodedSlashes](https://httpd.apache.org/docs/2.4/mod/core.html#allowencodedslashes). Is there any equivalent in caddy?

jupiter:
Fixes #643

### DIFF
--- a/middleware/proxy/proxy.go
+++ b/middleware/proxy/proxy.go
@@ -165,6 +165,11 @@ func createUpstreamRequest(r *http.Request) *http.Request {
 	outreq := new(http.Request)
 	*outreq = *r // includes shallow copies of maps, but okay
 
+	// Restore URL Path if it has been modified
+	if outreq.URL.RawPath != "" {
+		outreq.URL.Opaque = outreq.URL.RawPath
+	}
+
 	// Remove hop-by-hop headers to the backend.  Especially
 	// important is "Connection" because we want a persistent
 	// connection, regardless of what the client sent to us.  This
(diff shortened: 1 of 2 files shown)


mholt:
Interesting, I wonder why the Travis CI builds didn't run...

jupiter:
Must be because there is no `URL.RawPath` - this must be why it is better to use `URL.EscapedPath()` rather than reading `URL.RawPath`.

mholt:
I have a hunch that most middlewares will not be proxying requests, and that the "norm" will be to use the URL to load a file. Do you think that's accurate?

If so, I'm fine with making the fix in the middleware handlers to handle these edge cases.

We have to sanitize the URL to prevent tampering with the paths and accessing parts of the system that one shouldn't, which is why we do it up front.

jupiter:
That makes sense. However, it does become impossible to match against an 'unclean' path as all middlewares use `URL.Path` at the moment, but it would probably be very rare that anyone would want to. It just seems a bit heavy-handed to me. I would have said that all middlewares that actually traverse resources similar to a file system should do their own cleaning, but I don't know how many that is.

mholt:
I think it's best that we clean up the URL.Path and if a middleware wants to use a potentially haphazard path, it can use the RawPath explicitly. The way you have done that here shows it's not too difficult.

One last thing, I think we should add a test case to prevent this from popping up in the future. Could you do that when you have a chance? :smile:

mholt:
Great! Thanks @jupiter !

Issue #115, opened 2015-06-07
Add upstream directive


lenovouser:
My caddyfile gets very unorganized when I try proxy to a lot of different backend servers.

It looks kinda like this:

```
proxy / api.backend.mydomain.tld:6000 api.backend.mydomain.tld:6001 api.backend.mydomain.tld:6002 api.backend.mydomain.tld:6003 api.backend.mydomain.tld:6005 api.backend.mydomain.tld:6006 api.backend.mydomain.tld:6007 api.backend.mydomain.tld:6008 api.backend.mydomain.tld:6010 {
	policy round_robin
	websocket
	proxy_header Host {host}
}
```

It would be nice to have something similar to the NGINX **[upstream](http://nginx.org/en/docs/http/ngx_http_upstream_module.html)** directive. Maybe even with a few advancements to it. E.g. the ability to use port-ranges. Ideally I would like it to look something like this:

```
upstream api {
	api.backend.mydomain.tld:{6001:6010}
	api.backend.backup.mydomain.tld:{6011:6020} backup
}
proxy / api {
	policy round_robin
	websocket
	proxy_header Host {host}
}
```

Important options used in the NGINX upstream are:

  * **weight:** You can set it from 0 to 10. The higher the number the more load is being directed to the server. Useful if you have servers with different memory and processor configurations.
  * **backup:** Marks a server as backup. This server only gets used when the servers without the **backup** directive are not responding.
  * **down:** Marks a server down. Useful because you don't have to delete the line in case you need it again.
  * **resolve:** Looks for backend IP address changes. Useful when having a backend that gets moved a lot.

mholt:
That's a long line, indeed. I like the port ranges idea, although I don't know how commonly that is needed. I would be interested to know if others have similar needs.

Without saying yes or no to these features either way, mind if I suggest an alternative syntax, just for the sake of discussion?

```
proxy / api.backend.mydomain.tld:6001-6020 {
    policy round_robin
    websocket
    proxy_header Host {host}
}
```

I realize that doesn't allow the 4 options you mentioned, but I do like this better than defining upstreams and stuff - since doing it the way you've proposed changes the rules of the Caddyfile.

Curious about some of the options, mostly because I haven't set up a complex architecture like you have. The **backup** option - why keep any of the backends idle when you could be using them to share the load? Any one backend is a backup to any other backend already. The **down** option -- why not just put it in a comment instead? The **resolve** option -- why does the IP address matter when you use domain names to specify the backends anyway?

[...]

PR #861, opened 2016-06-03
Add upstream directive

marc-gr:
Hi! First PR here!

I added the upstream directive mentioned in #115 

This adds the ability to add an upstream directive to a proxy block, and is also compatible with the common format. Also support for port ranges is added, so the formats below will be supported:

```yaml
proxy / localhost:80 {
    upstream example.com:8080-8090
    ....
}
```

```yaml
proxy / {
    upstream example.com:8080-8090
    ....
}
```

```yaml
proxy / localhost:80 {
    ....
}
```

```yaml
proxy / {
    upstream example.com:8080-8090
    upstream localhost:2222-2225
    ....
}
```

...and so on.

I also added some tests for it.

Expecting your impressions about it!

### DIFF
--- a/caddy/setup/proxy_test.go
+++ b/caddy/setup/proxy_test.go
@@ -18,7 +18,7 @@ func TestUpstream(t *testing.T) {
 			"proxy / localhost:80",
 			false,
 			map[string]struct{}{
-				"http://localhost:80": struct{}{},
+				"http://localhost:80": {},
 			},
 		},
 
@@ -27,9 +27,9 @@ func TestUpstream(t *testing.T) {
 			"proxy / localhost:8080-8082",
 			false,
 			map[string]struct{}{
-				"http://localhost:8080": struct{}{},
-				"http://localhost:8081": struct{}{},
-				"http://localhost:8082": struct{}{},
+				"http://localhost:8080": {},
+				"http://localhost:8081": {},
+				"http://localhost:8082": {},
 			},
 		},
 
@@ -38,7 +38,7 @@ func TestUpstream(t *testing.T) {
 			"proxy / {\n upstream localhost:8080\n}",
 			false,
 			map[string]struct{}{
-				"http://localhost:8080": struct{}{},
+				"http://localhost:8080": {},
 			},
 		},
 
@@ -47,8 +47,8 @@ func TestUpstream(t *testing.T) {
 			"proxy / {\n upstream localhost:8080-8081\n}",
 			false,
 			map[string]struct{}{
-				"http://localhost:8080": struct{}{},
-				"http://localhost:8081": struct{}{},
+				"http://localhost:8080": {},
+				"http://localhost:8081": {},
 			},
 		},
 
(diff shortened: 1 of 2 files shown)


mholt:
Woah, this is awesome. Nice work!! Will take a closer look very soon.

caddy/setup/proxy_test.go
```diff
@@ -0,0 +1,122 @@
...
+		// test #4 test to destination with upstream directive
+		{
+			"proxy / localhost:8080 {\n upstream localhost:8081-8082\n}",
+			false,
+			map[string]struct{}{
+				"http://localhost:8080": {},
+				"http://localhost:8081": {},
+				"http://localhost:8082": {},
+			},
+		},
```
mholt: How about adding a test case that has 2 lines of upstream? i.e. `proxy / localhost:8080 {\n upstream localhost:8081-8082\n upstream localhost:8083-8085}`

caddy/setup/proxy_test.go
```diff
@@ -0,0 +1,122 @@
...
+		if err != nil && !test.shouldErr {
+			t.Errorf("Test case #%d received an error of %v", i, err)
+		} else if test.shouldErr {
+			continue
+		}
+
+		upstreams := receivedFunc(nil).(proxy.Proxy).Upstreams
+		for _, upstream := range upstreams {
+			val := reflect.ValueOf(upstream).Elem()
+			hosts := val.FieldByName("Hosts").Interface().(proxy.HostPool)
```
mholt: Fancy reflection work here. I guess this is a good idea in case a concrete type other than `staticUpstream` is in the list. (Did you do that on purpose?)

middleware/proxy/upstream.go
```diff
@@ -56,17 +57,38 @@ func NewStaticUpstreams(c parse.Dispenser) ([]Upstream, error) {
...
+					return upstreams, err
+				}
 			}
 		}
 
+		if len(to) == 0 {
+			return upstreams, c.ArgErr()
+		}
+
 		upstream.Hosts = make([]*UpstreamHost, len(to))
```
mholt: My intuition tells me it is, but have you checked that it's to leave `upstream.Hosts` nil in the case `len(to) == 0` (just above)?

caddy/setup/proxy_test.go
marc-gr: This makes a lot of sense, I'm going to add it right away 👍

mholt:
I cannot see anything obviously wrong with this, the tests are a breath of fresh air, and the cross-platform CI checks pass. Are you sure this is really your first pull request? :+1:

I'll squash and merge this so it's sure to go out for the 0.9 betas coming up in a week or two. That way folks can put it through the wringer before the stable release.

mholt:
... and my twitchy finger did a merge instead of a squash and merge 😞  Oh well.

@marc-gr I'm going to give you push rights so you can more easily contribute to the code base - reviewing pull requests, committing to branches, etc. Feel free to contribute again when you can / feel like it!

Thanks for doing this!

PR #869, opened 2016-06-07
Update go get path for Caddy binary

tpng:


mholt:
Oh thanks :smile: That reminds me I'm gonna revise the whole readme pretty soon. It hasn't changed much since Caddy was first released over a year ago!

PR #871, opened 2016-06-07
Log Requests

### LINKED ISSUE #858: Add ability to debug log http requests
any

I would like to be able to debug the incoming http request that I'm proxying to another server/service.  Specifically, I would like to debug output the HTTP headers, but it might be useful to debug the http content as well.

See this forum post for reference: https://forum.caddyserver.com/t/logging-http-headers-for-debugging/88/6?u=klauern

gchaincl:
Add a {request} placeholder to the replacer.

Closes #858.

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -107,7 +107,6 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
 			"{request}": func() string {
 				dump, err := httputil.DumpRequest(r, false)
 				if err != nil {
-					// should we return err.Error()?
 					return ""
 				}
 
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -3,12 +3,21 @@ package httpserver
 import (
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


caddyhttp/httpserver/replacer.go
```diff
@@ -95,6 +104,15 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
 				dir, _ := path.Split(r.URL.Path)
 				return dir
 			}(),
+			"{request}": func() string {
+				dump, err := httputil.DumpRequest(r, false)
+				if err != nil {
+					// should we return err.Error()?
+					return ""
```
mholt: > // should we return err.Error()?

Only if you want the error showing up where you don't expect it. :wink: A quick glance inside the DumpRequest() function with false as the second argument reveals that an error is very unlikely... the system would have to be completely out of memory I think. So I think returning empty string is fine.

gchaincl: makes sense then

mholt:
Thanks @gchaincl! Glad to have your contribution. :smile: I'll give you push rights to the repository so, in the future if you wish to contribute, you can create a branch and review other PRs, etc.

PR #873, opened 2016-06-08
Fix restart on USR1 not loading new Caddyfile

tpng:


caddyhttp/httpserver/plugin.go
```diff
@@ -51,7 +51,8 @@ var contexts []*httpContext
 
 func newContext() caddy.Context {
 	context := &httpContext{keysToSiteConfigs: make(map[string]*SiteConfig)}
-	contexts = append(contexts, context)
+	// put the new context at start to allow setup of directives on new instance
+	contexts = append([]*httpContext{context}, contexts...)
```
tpng: @mholt Is there any reason in using a contexts slice instead of a single context?

mholt: In case caddy.Start() is called more than once; it would definitely be unusual but since Caddy can be used as a library I'm not gonna rule it out.

tpng: Ah I see, that make sense now.

mholt: I wonder: should the context have some attachment to or knowledge of the Instance so that the order doesn't matter? I am trying to see if there is a way to do it without having to pass it into httpserver.GetConfig() which is obtained from the Controller, along with the Key.... yuck...

mholt: I think I'll just leave it like it is for now, but I'll still be looking at this issue after the PR is merged.

caddy.go
```diff
@@ -405,13 +410,10 @@ func startWithListenerFds(cdyfile Input, inst *Instance, restartFds map[string]r
...
-		// run startup callbacks since this is not a restart
-		for _, startupFunc := range inst.onStartup {
-			err := startupFunc()
-			if err != nil {
-				return err
-			}
+	for _, startupFunc := range inst.onStartup {
+		err := startupFunc()
+		if err != nil {
+			return err
```
mholt: I'm not sure we want to run the startup callbacks during restarts. Was this change intentional?

caddy.go
```diff
@@ -160,6 +163,8 @@ func (i *Instance) Restart(newCaddyfile Input) (*Instance, error) {
 	}
 	instancesMu.Unlock()
 
+	i.Stop()
```
mholt: Turns out `i.Stop()` does the same thing as the for loop starting on line 155/158 does (removing it from the list).

caddy.go
tpng: Yes, otherwise the `log` and `errors` directives will write to nil and panic on each request since they open the log file in the onStartup call.
Or we can fix that in `log` or `errors` directives instead, what do you think?

caddy.go
tpng: Oops, right. PR updated.

caddy.go
mholt: Oof. You're right. Maybe we need more specific callbacks, instead of just startup, restart, and shutdown. The older versions of caddy had a OnFirstStartup callback, but even then, log and errors didn't use it (oops).

This PR has been good, I'm re-thinking quite of few of these details now. I need to reconsider the relationship between `caddy.Instance` and `httpserver.httpContext`.

I also will look into adding more specific callbacks, so that we have:

- OnFirstStartup: When the Instance starts for the first time (not during restarts)
- OnStartup: When the Instance starts, even during restarts
- OnRestart: When the Instance begins to be restarted
- OnShutdown: When the Instance stops, even during restarts
- OnFinalShutdown: When the Instance stops for good (not a restart)

What do you think of that?

If that's good, I'll be happy to merge this PR and then make these changes.

tpng: That's cool, please merge, thanks!

mholt: These other callbacks were implemented in 15fa5cf2da0b721066210afab45cd2599ee0dd68.

PR #876, opened 2016-06-09
Add minify directive

hacdias:
[See this also.](https://github.com/caddyserver/caddyserver.com/pull/52)

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -328,10 +328,10 @@ var directives = []string{
 	"git",    // github.com/abiosoft/caddy-git
 
 	// directives that add middleware to the stack
-	"minify", // github.com/hacdias/caddy-minify
 	"log",
 	"gzip",
 	"errors",
+	"minify", // github.com/hacdias/caddy-minify
 	"ipfilter", // github.com/pyed/ipfilter
 	"search",   // github.com/pedronasser/caddy-search
 	"header",
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -331,7 +331,7 @@ var directives = []string{
 	"log",
 	"gzip",
 	"errors",
-	"minify", // github.com/hacdias/caddy-minify
+	"minify",   // github.com/hacdias/caddy-minify
 	"ipfilter", // github.com/pyed/ipfilter
 	"search",   // github.com/pedronasser/caddy-search
 	"header",
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -328,6 +328,7 @@ var directives = []string{
 	"git",    // github.com/abiosoft/caddy-git
 
 	// directives that add middleware to the stack
+	"minify", // github.com/hacdias/caddy-minify
 	"log",
 	"gzip",
 	"errors",


caddyhttp/httpserver/plugin.go
```diff
@@ -328,6 +328,7 @@ var directives = []string{
 	"git",    // github.com/abiosoft/caddy-git
 
 	// directives that add middleware to the stack
+	"minify", // github.com/hacdias/caddy-minify
```
mholt: Since minify can generate errors, it should at least go after the errors middleware. And you definitely don't want to be minifying after the gzip writer has closed.

caddyhttp/httpserver/plugin.go
```diff
@@ -331,6 +331,7 @@ var directives = []string{
 	"log",
 	"gzip",
 	"errors",
+	"minify", // github.com/hacdias/caddy-minify
 	"ipfilter", // github.com/pyed/ipfilter
```
mholt: You're gonna find this annoying, but there is a single space that your commit needs in order for the CI tests to pass; run `go fmt` on this file and you'll see that the comment should be lined up with the others. Sorry to be nit-picky!

hacdias:
What's that "invalid-email-address" in there?

mholt:
It means you pushed a commit done with a different email address than the one you signed the CLA with, I think. It seems to think that one of the commits had an "invalid-email-address" (whatever that means).

hacdias:
Oh.. I see. Is that going to change anything?

mholt:
Ummm I dunno. It might be a bug in @CLAassistant. I can use my "admin powers" to override that check in this case, since I know you've signed the CLA.

hacdias:
Thanks then.

andrewhamon:
Its because you updated you used the github UI to update your branch. Check your git logs and you will see that the author is Github on the last two commits.

I fixed this in my PR by amending the author and force pushing :)

hacdias:
Hey @andrewhamon 

Thanks for that! :D I check my log and it is:

```
commit 492d5aa37f7cb717ce272418426a667ee0692796
Merge: 1e4a410 daa4de5
Author: Henrique Dias <hacdias@gmail.com>
Date:   Fri Jun 10 07:34:42 2016 +0100

    Merge branch 'master' into patch-1

commit 1e4a4109a7216761511c9660b60c7a7f19f6aa0d
Author: Henrique Dias <hacdias@gmail.com>
Date:   Fri Jun 10 07:31:07 2016 +0100

    Update plugin.go

commit daa4de572e92668e8b40024c641cdd3d9fa6f21c
Author: Matthew Holt <mholt@users.noreply.github.com>
Date:   Thu Jun 9 19:12:11 2016 -0600

    Ensure certificate has a non-nil config when caching (fixes #875)

    Also we change the scheme of the site's address if TLS is enabled and
    no other scheme is explicitly set; this makes it appear as "https" when
    we print it; otherwise it would show "http" when TLS is turned on
    implicitly, and that is confusing/incorrect.

commit 83451ea2a0f6177214ec7e9f4e2d1a496d0b4a4d
Author: Henrique Dias <hacdias@gmail.com>
Date:   Thu Jun 9 16:06:50 2016 +0100

    Update plugin.go

commit 06fed0db1722b640749f1e2f1402fe644f4b1e77
Author: Henrique Dias <hacdias@gmail.com>
Date:   Thu Jun 9 15:14:46 2016 +0100

    Add minify directive
```

I guess the problem is not with my email, right @mholt? :P

andrewhamon:
Thats strange, this looks exactly like what I encountered. Even more weird is on the commits tab of this PR those commits still say "hacdias committed on GitHub" and those are the ones that fail the CLA check, but the email on them is definitely yours.

PR #880, opened 2016-06-12
Balance round robin evenly when some hosts are down

andrewhamon:
Before, when load balancing across multiple hosts, if a host went down
then the next host in line would be sent a double share of requests.
This is because the round robin counter was only incremented once per
request, regardless of the health of the selection. If current
selection was unhealthy then the policy would advance to the next host,
but this would not be reflected in the policy counter. To fix this, the
counter is now incremented for every attempted host.

This commit adds a test case that identifies the issue, and a fix.

### DIFF
--- a/caddyhttp/proxy/policy.go
+++ b/caddyhttp/proxy/policy.go
@@ -2,7 +2,7 @@ package proxy
 
 import (
 	"math/rand"
-	"sync/atomic"
+	"sync"
 )
 
 // HostPool is a collection of UpstreamHosts.
@@ -83,20 +83,21 @@ func (r *LeastConn) Select(pool HostPool) *UpstreamHost {
 // RoundRobin is a policy that selects hosts based on round robin ordering.
 type RoundRobin struct {
 	robin uint32
+	mutex sync.Mutex
 }
 
 // Select selects an up host from the pool using a round robin ordering scheme.
 func (r *RoundRobin) Select(pool HostPool) *UpstreamHost {
 	poolLen := uint32(len(pool))
-	selection := atomic.AddUint32(&r.robin, 1) % poolLen
-	host := pool[selection]
-	// if the currently selected host is not available, just ffwd to up host
-	for i := uint32(1); !host.Available() && i < poolLen; i++ {
-		selection = atomic.AddUint32(&r.robin, 1) % poolLen
-		host = pool[selection]
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	// Return next available host
+	for i := uint32(0); i < poolLen; i++ {
+		r.robin++
+		host := pool[r.robin%poolLen]
+		if host.Available() {
+			return host
+		}
 	}
-	if !host.Available() {
-		return nil
-	}
-	return host
+	return nil
 }
--- a/caddyhttp/proxy/policy.go
+++ b/caddyhttp/proxy/policy.go
@@ -92,7 +92,8 @@ func (r *RoundRobin) Select(pool HostPool) *UpstreamHost {
 	host := pool[selection]
 	// if the currently selected host is not available, just ffwd to up host
 	for i := uint32(1); !host.Available() && i < poolLen; i++ {
-		host = pool[(selection+i)%poolLen]
+		selection = atomic.AddUint32(&r.Robin, 1) % poolLen
+		host = pool[selection]
 	}
 	if !host.Available() {
 		return nil
--- a/caddyhttp/proxy/policy.go
+++ b/caddyhttp/proxy/policy.go
@@ -82,17 +82,17 @@ func (r *LeastConn) Select(pool HostPool) *UpstreamHost {
 
 // RoundRobin is a policy that selects hosts based on round robin ordering.
 type RoundRobin struct {
-	Robin uint32
+	robin uint32
 }
 
 // Select selects an up host from the pool using a round robin ordering scheme.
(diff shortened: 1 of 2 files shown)


andrewhamon:
If you'd like to see this with your own eyes, I have reproduced the issue with docker here:

https://gist.github.com/andrewhamon/ae3bfbd66420bc3e1147e2bc0ece08d6

If you kill any of the servers you will clearly see that the number of requests will not be split evenly among the remaining servers.

andrewhamon:
Now that I've thought about it, I believe I've introduced a concurrency error. If there are multiple requests at the same time, then each request might not get a chance to look at each host in succession -- there would be skips and jumps from the counter being incremented in another goroutine.

As far as I can tell the most straightforward solution is to wrap the whole operation in a mutex.

mholt:
Hi @andrewhamon - thanks for the pull request, the test case, and the clear description. I gotta get a few beta bugs fixed real quick but I'll be looking at this more closely very soon, since I want to see your fix go out with 0.9; maybe even the next beta version.

Host selection I don't believe is very expensive; but do you expect any performance impacts with locking it?

andrewhamon:
Thanks @mholt! This is my first real open source pull request.

As far as performance -- I'm really not sure. I wouldn't expect this to be a hostspot though. I decided to do a simple benchmark to see the difference between a mutex and atomic under different circumstances.

## Mutex vs Atomic
### Code
```go
// syncing.go
package main

import (
	"sync"
	"sync/atomic"
)

type Counter struct {
	counter uint32
	mutex   sync.Mutex
}

// g is number of go routines
// n is number of times counter gets incremented
// per gouroutine

// Synchronize access using atomic
func GoIncAtomic(counter *Counter, g int, n int) {
	wg := sync.WaitGroup{}
	wg.Add(g)
	for i := 0; i < g; i++ {
		go func() {
			defer wg.Done()
			for j := 0; j < n; j++ {
				atomic.AddUint32(&counter.counter, 1)
			}
		}()
	}
}

// Synchronize access using a mutex
func GoIncMutex(counter *Counter, g int, n int) {
	wg := sync.WaitGroup{}
	wg.Add(g)
	for i := 0; i < g; i++ {
		go func() {
			defer wg.Done()
			counter.mutex.Lock()
			defer counter.mutex.Unlock()
			for j := 0; j < n; j++ {
				counter.counter++
			}
		}()
	}
}
```
### Benchmarks
```go
package main

import "testing"

var numGoRoutines = 100

// Simulate a best case scenario, where a healthy server is immediately found
func BenchmarkGoIncAtomicServersUp(b *testing.B) {
	c := Counter{}
	for n := 0; n < b.N; n++ {
		GoIncAtomic(&c, numGoRoutines, 1)
	}
}

func BenchmarkGoIncMutexServersUp(b *testing.B) {
	c := Counter{}
	for n := 0; n < b.N; n++ {
		GoIncMutex(&c, numGoRoutines, 1)
	}
}

// Simulate worst case, where 100 servers are all down
func BenchmarkGoIncAtomicServersDown(b *testing.B) {
	c := Counter{}
	for n := 0; n < b.N; n++ {
		GoIncAtomic(&c, numGoRoutines, 100)
	}
}

func BenchmarkGoIncMutexServersDown(b *testing.B) {
	c := Counter{}
	for n := 0; n < b.N; n++ {
		GoIncMutex(&c, numGoRoutines, 100)
	}
}
```

### Results

```
PASS
BenchmarkGoIncAtomicServersUp-4  	   50000	     24244 ns/op
BenchmarkGoIncMutexServersUp-4   	   50000	     35794 ns/op
BenchmarkGoIncAtomicServersDown-4	   10000	    210493 ns/op
BenchmarkGoIncMutexServersDown-4 	   20000	     62358 ns/op
ok  	github.com/andrewhamon/mutex_test	7.716s
```

### Analysis

So in a situation where all servers are healthy and there are 100 requests in flight, it looks like atomic is about 50% faster. Conversely, if you have a bunch of servers that are all unhealthy, the mutex is nearly an order of magnitude faster. The amortized difference in processing time per "request" is 116ns when all servers are healthy.

### Conclusion

Technically, yes, using a mutex appears to be a performance regression. But, difference is only a fraction of a fraction of a millisecond. The variance of response time from whatever is upstream will probably eclipse this.

### Caveats

This is obviously a very simplified test. It would be nice to compare two builds of Caddy using actual requests, but that would be substantially more work to get started.

mholt:
Excellent work @andrewhamon - nice benchmarks! Quite a bit more detailed than I was expecting :smile: but really good to know.

After reading this through a couple times, and comparing it with what we had before, it seems correct and logical (although I am sometimes wrong!) and I'm happy to merge it in. Thanks for the improvement, and congrats on your first PR!

Here's what I really liked about this PR:

- Even though there wasn't an issue open about this behavior, you clearly described it and gave instructions to replicate it
- You vetted your own code and fixed an error in it
- The change is bite-size enough to comprehend
- You modified test to expose the bug
- A for your performance analysis (it doesn't use the actual code in the PR but it's basically the same thing which I think works, assuming the `Available()` call is quick)

I'll grant you push rights so that you can contribute more directly in the future by making branches right in the repository, reviewing others' pull requests, merging them after reviews, etc! Thanks for being thorough. Feel free to contribute in the future :+1:

andrewhamon:
Great! I have a few more contributions I'm planning to make, I just happened upon this while reviewing all the proxy related code. I look forward to making Caddy even more awesome!

PR #885, opened 2016-06-15
Fix rewrite bug with URL query + test case (#884)

### LINKED ISSUE #884: rewrite bug with URL query
0.9 (master)

Use `rewrite` directive with URL on query
 
```text
0.0.0.0:2015
rewrite {
    r (.*)
    to /proxy/{path}?{query}
}
```

`go run main.go`

Caddy should rewrite any URL without changing the query.

Example:
```
/example?url=http://caddyserver.com -> /proxy/example?url=http://caddyserver.com
```

Caddy is removing a `/` from the query.

Example
```
/example?url=http://caddyserver.com -> /proxy/example?url=http:/caddyserver.com
```

Just use the caddyfile

pedronasser:
Prevent `rewrite.To()` method to use `path.Clean()` on URL's query.
Closes #884.

### DIFF
--- a/caddyhttp/rewrite/to.go
+++ b/caddyhttp/rewrite/to.go
@@ -18,8 +18,15 @@ func To(fs http.FileSystem, r *http.Request, to string, replacer httpserver.Repl
 
 	// try each rewrite paths
 	t := ""
+	query := ""
 	for _, v := range tos {
-		t = path.Clean(replacer.Replace(v))
+		t = replacer.Replace(v)
+		tparts := strings.Split(t, "?")
+		t = path.Clean(tparts[0])
+
+		if len(tparts) > 1 {
+			query = tparts[1]
+		}
 
 		// add trailing slash for directories, if present
 		if strings.HasSuffix(v, "/") && !strings.HasSuffix(t, "/") {
@@ -47,9 +54,9 @@ func To(fs http.FileSystem, r *http.Request, to string, replacer httpserver.Repl
 
 	// perform rewrite
 	r.URL.Path = u.Path
-	if u.RawQuery != "" {
+	if query != "" {
 		// overwrite query string if present
-		r.URL.RawQuery = u.RawQuery
+		r.URL.RawQuery = query
 	}
 	if u.Fragment != "" {
 		// overwrite fragment if present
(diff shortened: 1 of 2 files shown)


guilhermebr:
:+1

caddyhttp/rewrite/to.go
```diff
@@ -18,8 +18,15 @@ func To(fs http.FileSystem, r *http.Request, to string, replacer httpserver.Repl
 
 	// try each rewrite paths
 	t := ""
+	query := ""
 	for _, v := range tos {
-		t = path.Clean(replacer.Replace(v))
+		t = replacer.Replace(v)
+		tparts := strings.Split(t, "?")
```
mholt: Is this a safe way to split the URL?

pedronasser: Not really sure? I think more about possible cases and other solution to get the query.
And we can't really use url.Parse() to solve that problem.

mholt: Hmm, well I like the simplicity of your PR as-is, I'll look at this again tomorrow after I've had some sleep and try to think of ways to break it. In the meantime it looks like this is easy enough to test as you've shown, so we can always add those later if they come up.

abiosoft:
Well, I am not in favour of this. if you really want to use that url, then escape it.

`path.Clean` cleans up the url and prevents accessing parent directories.

caddyhttp/rewrite/to.go
abiosoft: why not `tparts := strings.SplitN(t, "?", 2)` ?

abiosoft:
Your description was a bit misleading. I thought you removed `path.Clean` in its entirety.
> Prevent rewrite.To() method to use path.Clean() on URL's query.

abiosoft:
This is why we provided `{path_escaped}`, `{uri_escaped}`, `{query_escaped}` placeholders. Though it did not occur to me users may need to put in their own urls.

There was once a similar issue but caused by regex and was handled here.
https://github.com/mholt/caddy/blob/master/caddyhttp/rewrite/rewrite.go#L171.

caddyhttp/rewrite/to.go
pedronasser: Agreed. SplitN is better in this case. I'll replace it.

PR #887, opened 2016-06-16
Add timeout to health_check

dulldave:
The `DefaultClient` in net/http doesn't specify a timeout and is 0 by default. Which means if a server is causing a connection to hang the request will hang.

This is to add a way to specify a "timeout" on health_check, as well as set a sensible 60 second default.

Example:

```
localhost {
    proxy / localhost:8080 {
        health_check /health interval=10s timeout=15s
    }
}
```

Looking forward to everyones thoughts. 😄

### DIFF
--- a/caddyhttp/proxy/upstream.go
+++ b/caddyhttp/proxy/upstream.go
@@ -31,6 +31,7 @@ type staticUpstream struct {
 	MaxFails    int32
 	MaxConns    int64
 	HealthCheck struct {
+		Client   http.Client
 		Path     string
 		Interval time.Duration
 		Timeout  time.Duration
@@ -100,6 +101,9 @@ func NewStaticUpstreams(c caddyfile.Dispenser) ([]Upstream, error) {
 		}
 
 		if upstream.HealthCheck.Path != "" {
+			upstream.HealthCheck.Client = http.Client{
+				Timeout: upstream.HealthCheck.Timeout,
+			}
 			go upstream.HealthCheckWorker(nil)
 		}
 		upstreams = append(upstreams, upstream)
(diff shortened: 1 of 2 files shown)


caddyhttp/proxy/upstream.go
```diff
@@ -287,9 +304,12 @@ func parseBlock(c *caddyfile.Dispenser, u *staticUpstream) error {
 }
 
 func (u *staticUpstream) healthCheck() {
+	client := http.Client{
+		Timeout: u.HealthCheck.Timeout,
+	}
```
mholt: It'd be nice if we re-used this client for all health checks, I think the Go docs say something about reusing clients being a good thing.

mholt:
Hey thanks for the PR!

I like the idea of being able to customize the health check timeout; however, using the `key=value` notation on a single line isn't the Caddy way; instead:

```
health_check_timeout 15s
health_check_interval 10s
```

will be more consistent with all other directives (even though `health_check_*` is kind of long).

weingart:
As an FYI, I think that the context package will enable this to be handled in a different manner in the near future...

dulldave:
Thanks for the comments guys!

So I've updated the code now to reflect the suggestions from @mholt:
* Client is now called once.
* Updated from the `key=value` notation to doing it the caddy way using more consistent directives. 

Example:
```
localhost {
    proxy / localhost:8080 {
        health_check /health
        health_check_interval 50s
        health_check_timeout 30s
    }
}
```

mholt:
This is much better. Thanks @dulldave!

@weingart Isn't that only useful when initiating timeout between handlers? Here we're proxying a request upstream to a different process.

weingart:
No, I think it will make sense even for proxying requests.  IE: the context will likely contain a timeout/deadline value, which can then be passed on and cancelled on during the backside request when it times out.

mholt:
I need to start reading up on Context...

PR #889, opened 2016-06-17
Condition upgrades

abiosoft:
* Makes `if` condition reusable.
* Adds `if` condition to `redir`
* Adds `if_op` config to specify `and` or `or`. Defaults to `and` if omitted.

#### Example
Or is now possible. This will be `or`ed instead of `and`ed.
```
rewrite {
    if       {path} is /home
    if       {path} is /blog
    if_op    or
}
```
If condition is now available to `redir`. [This](https://forum.caddyserver.com/t/redirect-based-on-header-value-from-load-balancer/165/5?u=abiosoft) is now possible.
```
redir {
    if  {>X-Forwarded-Proto} is http
    /   https://{uri}
}
```
Any middleware/plugin can now integrate and support `if` condition using `httpserver.SetupIfMatcher` as it is now done with `redir` and `rewrite`.

### DIFF
--- a/caddyhttp/httpserver/condition.go
+++ b/caddyhttp/httpserver/condition.go
@@ -11,7 +11,6 @@ import (
 
 // SetupIfMatcher parses 'if' or 'if_type' in the current dispenser block.
 // It returns a RequestMatcher and an error if any.
-//  // Usage Example
 //  // Embed Matcher in plugin struct.
 //  type MyPlugin struct {
 //    ...
--- a/caddyhttp/httpserver/condition.go
+++ b/caddyhttp/httpserver/condition.go
@@ -1,4 +1,4 @@
-package rewrite
+package httpserver
 
 import (
 	"fmt"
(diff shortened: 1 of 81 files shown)


caddyhttp/httpserver/condition.go
```diff
@@ -0,0 +1,227 @@
...
+//        continue
+//      }
+//    }
+//  }
+//  ...
+//  // Check conditions in middleware/plugin httpserver.Handler.
+//  func (m Middleware) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
+//    for _, c := range m.Configs { if c.Match(r) { ... } ... }
+//  }
+//
```
mholt: Are these comments needed for this commit?

mholt:
It looks like `if_type` (from your comment) is actually `if_cond` (in the code) - which do you think is better?

caddyhttp/rewrite/rewrite.go
```diff
@@ -230,6 +228,10 @@ func (r *ComplexRule) matchExt(rPath string) bool {
 	return true
 }
 
+func newReplacer(r *http.Request) httpserver.Replacer {
+	return httpserver.NewReplacer(r, nil, "")
+}
+
```
mholt: I think this is only used once (at least, that's what a quick Cmd+F on this page tells me) - maybe not needed then?

abiosoft: It is also used in tests. No real benefits other being shorter.

caddyhttp/httpserver/condition.go
abiosoft: Not really. Should be removed in favour of a wiki entry.

abiosoft:
> It looks like `if_type` (from your comment) is actually `if_cond` (in the code) - which do you think is better?

Apparently, proof that I'm human 😄. I think I vote for `if_type`.

abiosoft:
How about `if_logic` ?

mholt:
`if_logic` is good! Let's do that one.

caddyhttp/httpserver/condition.go
mholt: I agree. Please do that so we can eliminate this TODO.

mholt:
Thanks Abiola! This is great. :smile:  :tada:

PR #890, opened 2016-06-17
fix for new rewrite test case and add table test to replacer

pedronasser:


caddyhttp/rewrite/to_test.go
```diff
@@ -22,7 +22,8 @@ func TestTo(t *testing.T) {
 		{"/?a=b", "/testfile /index.php?{query}", "/testfile?a=b"},
 		{"/?a=b", "/testdir /index.php?{query}", "/index.php?a=b"},
 		{"/?a=b", "/testdir/ /index.php?{query}", "/testdir/?a=b"},
-		{"/test?url=http://caddyserver.com", " /p/{path}?{query}", "/p/test?url=http://caddyserver.com"},
```
abiosoft: any reason why you removed this test ?

caddyhttp/rewrite/to.go
```diff
@@ -28,6 +28,10 @@ func To(fs http.FileSystem, r *http.Request, to string, replacer httpserver.Repl
 			query = tparts[1]
 		}
 
+		if strings.HasSuffix(tparts[0], "/") && !strings.HasSuffix(t, "/") {
+			t += "/"
+		}
+
```
abiosoft: With this, I do not think the similar check below is needed anymore. Maybe move the `// add trailing slash for directories, if present` comments over here and delete the other one.

caddyhttp/rewrite/to_test.go
pedronasser: Not removed, just replaced.
No need for the `caddyserver.com`. The problem is in the `http://`

caddyhttp/rewrite/to.go
pedronasser: I think they are different cases, but I will work on it

caddyhttp/rewrite/to_test.go
abiosoft: For readability since the emphasis is on trailing slash, there can be additional test for `http://caddyserver.com/`

pedronasser: The enphasis on that test was on the double slashes. But we could keep the full address as well

caddyhttp/rewrite/to.go
mholt: Finally getting around to this PR - what have you decided about this?

mholt:
Thanks @pedronasser! :+1:

PR #892, opened 2016-06-20
Add support for Alt-Svc header

wendigo:
This change will advertise QUIC availability using [Alt-Svc & Alternate-Protocol (deprecated) header](http://httpwg.org/http-extensions/alt-svc.html#alt-svc).

In the future, when the Chromium [host whitelist restriction](https://groups.google.com/a/chromium.org/forum/#!msg/proto-quic/1K79SVSLEiU/bvMLV0juEwAJ) will be removed clients will automatically switch to QUIC without special configuration described in [wiki](https://github.com/mholt/caddy/wiki/QUIC).

Furthermore instead of forcing QUIC, host can be whitelisted using command line flag:

```
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome \
    --user-data-dir=/tmp/chrome \
    --no-proxy-server \
    --enable-quic \
    --quic-host-whitelist="YOUR_SITE" "YOUR_SITE"
```

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -71,6 +71,7 @@ func NewServer(addr string, group []*SiteConfig) (*Server, error) {
 	// Enable QUIC if desired
 	if QUIC {
 		s.quicServer = &h2quic.Server{Server: s.Server}
+		s.Server.Handler = s.wrapWithSvcHeaders(s.Server.Handler)
 	}
 
 	// We have to bound our wg with one increment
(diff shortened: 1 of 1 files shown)


mholt:
Ah nice, I like that better. Will give it a try soon and update the wiki and merge this. Thanks!!

wendigo:
It's worth to mention that if UDP traffic is blocked Chrome/Chromium with such configuration will fallback to TCP (instead of displaying ERR_QUIC_PROTOCOL_ERROR) - it mimics better real behaviour :)

mholt:
Bah, okay, I'm having a few troubles testing the command that launches Chrome because of my local dev environment today - did you confirm it works, that indeed the page is being served via QUIC? If so I'll merge it in. :smile:

wendigo:
You can test it on my homepage: https://serafin.tech :) It works! Also I've tested it with UDP traffic blocked on :443

mholt:
Ah, that's a beautiful thing. Thanks!

wendigo:
Thx 👍

PR #894, opened 2016-06-20
Refactoring to remove lint

RobbieMcKinstry:
Hi Caddy community!

I took [Matt's advice](https://twitter.com/mholt6/status/744976277552062464) and wanted to get involved. 

I added comments to the `markdown` package to remove `golint` warnings. This commit removes all warnings that are possible to be removed. The interface required by `blackfriday`, the markdown library forces the remaining two warnings to remain.

I threw this together pretty quickly so if you'd like me to change something, feel free to comment with feedback.

Thanks!

PR #895, opened 2016-06-20
Fix a few lint warnings.

zmb3:


mholt:
Ohai Zac! Glad to see another PR from you. It appears your patch was submitted just a few minutes after #894 with the CLI signed, and that one was more comprehensive and it was his first PR, so I'm going to use that one. But thanks for putting in the effort! I really do appreciate it.

PR #907, opened 2016-06-29
Support for placeholders in fastcgi env vars.

abiosoft:


mholt:
Yay! This is a wonderful surprise. Thank you! :smile:

pwFoo:
So it's only supported with fastcgi? Would it (easy) possible to support env vars in the entire caddyfile ?

mholt:
These are FastCGI environment variables (like fastcgi_param in nginx conf) - not process environment variables. Those are supported in the caddyfile. See https://caddyserver.com/docs/caddyfile

PR #908, opened 2016-06-29
Proposal: Middleware Config

abiosoft:
* Created a generic config interface for middlewares.
* Matches by longest path approach.
* Port **rewrite** to use it.


caddyhttp/httpserver/middleware.go
```diff
@@ -48,20 +47,48 @@ type (
...
 	}
+
+	// Config is a middleware configuration.
+	// This makes it possible for middlewares to have a common
+	// configuration interface.
+	//
+	// TODO The long term plan is to get all middleware implement this
+	// interface for configurations.
+	Config interface {
+		RequestMatcher
```
mholt: The name `httpserver.Config` is a little ambiguous and looks more like a config for an HTTP server; maybe name this type `MiddlewareConfig` or `HandlerConfig` instead (I think HandlerConfig is technically more accurate, but I'm not married to it if you disagree)?

caddyhttp/rewrite/rewrite.go
```diff
@@ -31,18 +29,13 @@ const (
 type Rewrite struct {
 	Next    httpserver.Handler
 	FileSys http.FileSystem
-	Rules   []Rule
+	Rules   []httpserver.Config
```
mholt: Ah, this is elegant.

caddyhttp/httpserver/middleware.go
```diff
@@ -48,20 +47,48 @@ type (
...
+		BasePath() string
```
mholt: Hmm, if we can eliminate this method, then we may not even need the whole interface and can just use RequestMatcher.

caddyhttp/rewrite/to.go
```diff
@@ -12,8 +12,7 @@ import (
 
 // To attempts rewrite. It attempts to rewrite to first valid path
 // or the last path if none of the paths are valid.
-// Returns true if rewrite is successful and false otherwise.
-func To(fs http.FileSystem, r *http.Request, to string, replacer httpserver.Replacer) Result {
+func To(fs http.FileSystem, r *http.Request, to string, replacer httpserver.Replacer) (result Result) {
```
mholt: I typically try to avoid named returns; I don't really know what we gain from this one here.

caddyhttp/httpserver/middleware.go
```diff
@@ -48,20 +47,48 @@ type (
...
+// Select selects a Config.
+// This chooses the config with the longest length.
+//
+// TODO A better solution.
+func (c ConfigSelector) Select(r *http.Request) (config Config) {
+	for i := range c {
+		if !c[i].Match(r) {
+			continue
+		}
+		if config == nil || len(c[i].BasePath()) > len(config.BasePath()) {
```
mholt: This is the only place that BasePath() is used; I can see why it is, but I wonder if, instead of a Config interface as you have it here, maybe a new interface that compares whether one match is superior to another. In fact, this can be reduced to a sorting problem! And there's a standard `sort` package. What do you think about that train of thought?

mholt:
I really like where this is going!

caddyhttp/httpserver/middleware.go
abiosoft: Yeah. `HandlerConfig` sounds better.

caddyhttp/httpserver/middleware.go
abiosoft: Yeah but the goal is to abstract path matching from middlewares. And if that is gonna happen, we need a property in the middleware config to check to determine which config to pick.

caddyhttp/rewrite/to.go
abiosoft: That's fine. Just reducing lines of code.

caddyhttp/httpserver/middleware.go
abiosoft: I considered the middleware config implementing `sort.Interface` but that does not look like an ideal solution on the surface. Also, I do not want to demand too much from the middlewares.

caddyhttp/httpserver/middleware.go
mholt: You're right, but what if we don't limit this to the base path in the future? In the end, it's a sorting problem, and the highest-sorted config gets dibs. :) I think we should look for a more general way to handle this.

caddyhttp/httpserver/middleware.go
mholt: I can understand that. It is idiomatic at least. I'd still be interested in seeing if it's actually too difficult to do.

abiosoft: I think that basically takes us back to where we are coming from.

What I am thinking of is a way to abstract middleware config selection from the middlewares. A middleware can still override this by simply not using `ConfigSelector`.

mholt:
As discussed in Slack briefly, I'll wait for some minor changes to this PR (like marking these new types as experimental in the godoc) and then we'll merge it in. Abiola will continue to iterate on this until we get a really nice, elegant solution for all the middleware handlers.

mholt:
Thanks Abiola! Will be interesting to consider taking this further over time.

PR #909, opened 2016-06-29
Test that the host header forwarding on proxy

evermax:
This tests test two cases:
- When no host is specified, the address/host of the proxy should be used as host
- When the host is specified, it should appear in the request


mholt:
Yesss, finally putting an end to those pesky bug reports on this. :wink: Thank you Maxime!

I keep saying and I really mean it: next time you find yourself in Utah, I'll treat you to lunch. (Actually there are really cheap flights from San Diego to where I live starting in October I think.)

evermax:
My pleasure, happy to help!

I'll keep in mind your kind offer, Utah is definitely not far away 🙂

Issue #906, opened 2016-06-28
Make storage pluggable


mholt:
If caddytls.Storage was pluggable, it would make it possible to use something other than the local disk for TLS asset storage. In fact, the abstraction doesn't have to be limited to caddytls; if we think of storage as a simple key-value store, we may only need a few methods on an interface for all disk-related functions to be abstracted.

So two ways to do it:

1. Abstract caddytls.Storage, this makes only the TLS assets pluggable
2. Abstract assets.go, which makes all storage pluggable, however I don't know how feasible this is

Someone is welcome to look into this before I have a chance to get around to it.

\* _Adopted from [this forum topic which has more details](https://forum.caddyserver.com/t/any-ssl-solution-for-multi-server-setup-behind-same-domain/215?u=matt) but continue discussion here_

cretz:
From what I see, assets is only used for the TLS storage. I don't think its reasonable to abstract the assets stuff (and the moveStorage stuff or any of that). Instead, the default impl of the TLS storage interface should just leverage the `AssetsPath()` as necessary. If, in the future, there are other assets that leverage the filesystem, no reason to require them to share state like TLS.

mholt:
Well, Caddy does need a place to store assets; the search plugin does, for instance. (The search plugin was written before the auto TLS stuff and there was no concept of a storage place for Caddy assets, but there is now, and it will soon need to use it.)

So the storage abstraction will indeed need to be available to plugins somehow. I think that will be valuable in the long run!

cretz:
@mholt - Here is an early commit of what I am thinking: https://github.com/cretz/caddy/commit/84771da9fd041d534e5c0bd87d28a2251d57fae5. It will allow plugins to add storage just like they add middleware today. I took great care w/ my refactoring scalpel to not upset the code too much.

Please take a peek. Hopefully the benefits of keeping it specific to TLS, such as keeping the changes lean and not touching the rest of the codebase, are clear. I think a generic KV store is orthogonal to this. Also, this data is highly secret, so not sure it should be mixed with other arbitrary storage.

mholt:
@cretz Wow, that's quite the change! I can tell you put some time into it. I have made some comments in-line along with @weingart.

You make some good points about keeping TLS asset storage separate from "regular" Caddy storage, although I hope that all Caddy assets will be stored with care.

[...]

PR #913, opened 2016-07-01
Pluggable TLS Storage

cretz:
This addresses issue #906. I may have gone a bit overboard w/ the test harness :-). Quick notes:

* Test harness added in a separate package (`storagetest`) in the same vein as `iotest` or `httptest` since I knew I would end up creating one anyways when I get around to writing my own impl
* In-memory storage impl in the `storagetest` package for use by others during testing if desired
* Basic approach would be for plugins to set `httpserver.GetConfig(c).TLS.StorageCreator` (or course w/ `TLS` field null check) in a similar way to how middleware is added. This allows plugins to register themselves as normal, create directives as normal, etc.
* I will write the wiki for how to create a custom TLS storage impl once this PR is given the thumbs up
* I likely will be writing either a Vault or Consul+AES impl

As a reminder the whole point of wanting to do this, at least for me, is so I can finally have multiple load balanced end-to-end encrypted web servers that respond to HTTP challenges. Basically we'll be able to update the auto-https faq at https://github.com/caddyserver/caddyserver.com/blob/f27a911a9276cdb6d15ce762504b22db8d082721/public/docs/automatic-https.html#L71 to include the fact that if you use an alternative shared storage plugin, your stuff will work.

Another thing I want to ask: I didn't dig too deeply, but I trust that Caddy obviously caches the cert loaded from storage. So remote storage wouldn't constantly be hit, right?


cretz:
Note, all comments between my initial description at the top and this comment were based on an initial review of the first commit.

mholt:
Nice, I'll look at this in more scrutiny later! Thanks for the PR. Really looking good from what I saw earlier.

> Another thing I want to ask: I didn't dig too deeply, but I trust that Caddy obviously caches the cert loaded from storage. So remote storage wouldn't constantly be hit, right?

Right. That's what `certCache` is in certificates.go.

Regarding InMemoryStorage:

> I was going to use it as an example of an alternative impl and to show that my test harness works on just the interface so it validates both this and the file one. Since it has no runtime purpose per se, shall I remove it?

Ah, gotcha. Yeah, let's remove it if it's not used; feel free to keep what you need for tests in the test file. :)

Is this level of abstraction sufficient for your needs? Do you foresee maybe even a general-purpose plugin that we could put on the download page for others to use, thanks to this change? :smile:

cretz:
Once you peek at the code you'll see where I just made the in memory store as a tool for other testers in the `storagetest` package. I think it's fine there.

Also, I think this level of abstraction is perfect. I think one I (or someone else) develops a good storage plugin (i.e. one that has its own directives and what not) then it definitely would be worth including in the download. If only Go had better shared lib support it wouldn't have to be bundled.

caddytls/storagetest/memorystorage.go
```diff
@@ -0,0 +1,158 @@
...
+var memoryMutex sync.Mutex
+
+// memoryStoragesByCAURL is a map keyed by a CA URL string with values of
+// instantiated memory stores. Do not access this directly, it is used by
+// InMemoryStorageCreator.
+var memoryStoragesByCAURL = make(map[string]*InMemoryStorage)
+
+// InMemoryStorageCreator is a caddytls.Storage.StorageCreator to create
+// InMemoryStorage instances for testing.
+func InMemoryStorageCreator(caURL *url.URL) (caddytls.Storage, error) {
```
mholt: This type is only used for testing; can we move it into a test file?

caddytls/storagetest/storagetest.go
```diff
@@ -0,0 +1,304 @@
...
+		return err
+	}
+	if s.AfterUserEmailStore != nil {
+		s.AfterUserEmailStore("foo1@example.com")
+	}
+	if e := s.MostRecentUserEmail(); e != "foo1@example.com" {
+		return fmt.Errorf("Unexpected most recent email after second user reg: %v", e)
+	}
+	return nil
+}
```
mholt: This is a little unconventional; let's move the test logic into a true `_test.go` file.

caddytls/storagetest/memorystorage.go
cretz: I figured this was used for testing like those items in `iotest` and `httptest` are only used for testing. This can be used by other packages for their testing (same with the test harness), ref [this SO answer](http://stackoverflow.com/questions/31794141/can-i-create-shared-test-utilities-in-go)

caddytls/storagetest/storagetest.go
cretz: It's not test logic. It's for others to use in their implementations. Now when I write my plugin I can use this test harness to test my implementation. Reference [this SO answer](http://stackoverflow.com/questions/31794141/can-i-create-shared-test-utilities-in-go). Think of it as a "compatibility kit" or a "spec conformance suite" or a "black box harness" or any term that makes sense.

caddytls/storagetest/memorystorage.go
mholt: Oh, okay. I get you now.

cretz:
@mholt Before I dig too deep in the code, can you help me understand cache invalidation and auto renewal? Specifically, it seems like there is a `CacheManagedCertificate` that caches in a local map but would not get updated if renewal happened on another server.

Specifically, take the following scenario. Say I am using a multi-server setup (i.e. server A and server B) w/ shared storage to some DB backend or something. When server A and server B both feel the need to renew, what kind of race conditions persist here I wonder. Are overwrites and multiple challenges acceptable to the CA? Should I add some logic in `RenewManagedCertificates` that can tell the storage interface to use a mutex because we are in renewal? Basically this is the same question as if I were running two Caddy processes on the same machine for the same host (different ports) and sharing the same assets dir.

Now I'm thinking that I may need to more abstractions in the storage interface: `LockRenewal(domain string) (bool, error)` and `UnlockRenewal(domain string) error` so servers don't step on each other.

mholt:
@cretz 

>  Before I dig too deep in the code, can you help me understand cache invalidation and auto renewal? Specifically, it seems like there is a CacheManagedCertificate that caches in a local map but would not get updated if renewal happened on another server.

Correct; that doesn't synchronize, just loads from storage into memory.

> When server A and server B both feel the need to renew, what kind of race conditions persist here I wonder.

Each Caddy operates independently, renewing the certificates it is managing separately from other Caddy instances. Caddy would have to be programmed to coordinate with a fleet in order to synchronize renewals and other changes. (I'm working on a tool to do this, independently of Caddy, which will one day plug into Caddy. I do not believe this kind of logic belongs in the caddytls package, as it is highly diverse and complicated on its own...)

cretz:
> Each Caddy operates independently, renewing the certificates it is managing separately from other Caddy instances. Caddy would have to be programmed to coordinate with a fleet in order to synchronize renewals and other changes. (I'm working on a tool to do this, independently of Caddy, which will one day plug into Caddy. I do not believe this kind of logic belongs in the caddytls package, as it is highly diverse and complicated on its own...)

So, my whole idea of using shared storage to solve HA-multi-server HTTP challenges fails under race condition here (technically keys and certs stored at the same time could get out of order)? Would you mind if I added those two functions? I don't want Caddy to do the coordination either, but I do need to solve storage race conditions. Otherwise, do you foresee an issue of two servers issuing renewals at the same time? I am open to any solution here, but my whole purpose for pluggable storage was to allow multiple servers to share certs (and the renewal of them).

mholt:
> So, my whole idea of using shared storage to solve HA-multi-server HTTP challenges fails under race condition here (technically keys and certs stored at the same time could get out of order)?

Yep, right. Across multiple machines, each Caddy would do their renewals independently which would run into LE rate limits faster. So yeah, you'll need a way to sync their access (including checking their expiration date when scanning for renewals - making sure to update the in-memory cached cert inside the lock).

> Would you mind if I added those two functions?

Go for it.

> I don't want Caddy to do the coordination either (but I do want my plugin to), but I do need to solve storage race conditions.

Good clarification. That's fine.

cretz:
Welp, I may have been a bit too eager with this PR...I'll have some changes incoming at some point.

mholt:
I probably should have mentioned it, I just hadn't thought much of Caddy sharing its certificates with other processes on different machines.

Anyway, looking forward to the revisions whenever they show up. :)

cretz:
Ok, I did my best to try to alleviate my problems. I reduced site/user loading/storing to one call each. I also added a locking mechanism at strategic parts of the code to prevent duplicate renewals.

I hope this hasn't introduced too much complexity. If it has, we can take this PR off the table. I fear the locking parts I added to the somewhat-confusing code paths already taken for obtain/renew cert only make the code look a bit more confusing. I took care to remain compatible with how it works today. In my impl, I'll also have to watch for changes and call `CacheManagedCertificate` when I notice a change elsewhere.

Review appreciated and if the benefit of including this layer doesn't outweigh the cost of complexity added, we can kill this PR and I will find other approaches.

caddytls/filestorage.go
```diff
@@ -0,0 +1,239 @@
...
+		email = emptyEmail
+	}
+	email = strings.ToLower(email)
+	return filepath.Join(s.users(), email)
+}
+
+// emailUsername returns the username portion of an email address (part before
+// '@') or the original input if it can't find the "@" symbol.
+func emailUsername(email string) string {
+	at := strings.Index(email, "@")
```
weingart: I think this will be cleaner using `strings.SplitN(email, "@", 2)`

caddytls/filestorage.go
```diff
@@ -0,0 +1,239 @@
...
+	if at == -1 {
+		return email
+	} else if at == 0 {
+		return email[1:]
```
weingart: This return behaviour is not specified in the function comment.

caddytls/filestorage.go
cretz: While that might be true, this is taken verbatim from https://github.com/mholt/caddy/blob/fdb6d64f9d2afdeb6f843862da7dd243abb2399b/caddytls/storage.go#L119 and I wasn't wanting to change any of that logic during this refactoring. I think to clean up that should be a separate effort.

caddytls/filestorage.go
cretz: This is a problem also in master (ref: https://github.com/mholt/caddy/blob/fdb6d64f9d2afdeb6f843862da7dd243abb2399b/caddytls/storage.go#L119) and likely best addressed there. But if this function must be cleaned up as part of this effort, I can.

caddytls/filestorage.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	err = ioutil.WriteFile(s.siteCertFile(domain), data.Cert, 0600)
+	if err == nil {
+		err = ioutil.WriteFile(s.siteKeyFile(domain), data.Key, 0600)
+	}
+	if err == nil {
+		err = ioutil.WriteFile(s.siteMetaFile(domain), data.Meta, 0600)
+	}
+	return err
+}
```
weingart: Possibly this can be done like:
```go
func (s FileStorage) StoreSite(domain string, data *SiteData) error {
 	if err := os.MkdirAll(s.site(domain), 0700); err != nil {
 		return err
 	}
 	if err := ioutil.WriteFile(s.siteCertFile(domain), data.Cert, 0600); err != nil {
 		return err
 	}
 	if err := ioutil.WriteFile(s.siteKeyFile(domain), data.Key, 0600); err != nil {
 		return err
 	}
 	if err := ioutil.WriteFile(s.siteMetaFile(domain), data.Meta, 0600); err != nil {
 		return err
 	}
 	return nil
 }
```

caddytls/filestorage_test.go
```diff
@@ -0,0 +1,6 @@
+package caddytls
+
+// *********************************** NOTE ********************************
+// Due to circular package dependencies with the storagetest sub package and
+// the fact that we want to use that harness to test file storage, the tests
+// for file storage are done in the storagetest package.
```
weingart: Possible option for stubbing out file operations: https://talks.golang.org/2012/10things.slide#8

caddytls/storagetest/storagetest_test.go
```diff
@@ -0,0 +1,39 @@
+package storagetest
```
weingart: Possibly a way to mock out the filesystem: https://talks.golang.org/2012/10things.slide#8

caddytls/filestorage_test.go
cretz: This comment doesn't relate to the fact that I can't mock the filesystem or anything. It's more related to the fact that I built a test harness for use by any other storage provider implementors to use, but filestorage_test.go can't use it because it would cause a circular dependency. Therefore, the filestorage impl is actually tested in the `storagetest` package.

weingart:
FYI, take my comments as suggestions.  This is looking like it's going in the right direction.  😄

caddytls/filestorage.go
cretz: Yes, there are a million ways to do this. I tried to stay consistent with existing code (in most places) which seems to favor setting the var outside of the if statement. I also left the top part the way it was previously, otherwise I would have written:

```go
func (s FileStorage) StoreSite(domain string, data *SiteData) error {
	err := os.MkdirAll(s.site(domain), 0700)
	if err == nil {
		err = ioutil.WriteFile(s.siteCertFile(domain), data.Cert, 0600)
	}
	if err == nil {
		err = ioutil.WriteFile(s.siteKeyFile(domain), data.Key, 0600)
	}
	if err == nil {
		err = ioutil.WriteFile(s.siteMetaFile(domain), data.Meta, 0600)
	}
	return err
}
```

cretz:
Gladly. Thank you for looking! :-)

caddytls/storagetest/storagetest_test.go
cretz: I don't think we need to mock out the filesystem. Actually, in many of these cases use of the local FS is actually great integration/functional test anyways. Also note the existing tests, which I left intact as much as possible, use the FS. For example, I got Windows and Linux dir mod time testing for free when the CI ran.

caddytls/filestorage.go
mholt: I do actually find `if err != nil` to be more common/idiomatic in Go programs; this won't stop this PR but maybe consider it in the future if you ever go back to refactor this. :)

mholt:
I want this to go in beta 3.

@cretz Is this a change you are willing to maintain (at least in the foreseeable future) if problems or questions arise? I'll make you a collaborator so you have push rights when I merge this PR.

cretz:
Sure, I can maintain for the near future, though I can't promise I'll be able to for a really long time (though I hope it's clear enough for anyone to understand/contribute). I also plan on providing a consul plugin right away. I may request help w/ integration testing the multi-server-HTTP-challenge setup w/ the CA.

mholt:
Thanks @cretz for all your work! Looking forward to what this makes possible.

caddytls/maintain.go
```diff
@@ -93,7 +93,10 @@ func RenewManagedCertificates(allowPrompts bool) (err error) {
 				continue
 			}
 
-			// this works well because managed certs are only associated with one name per config
+			// This works well because managed certs are only associated with one name per config.
+			// Note, the renewal inside here may not actually occur and no error will be returned
+			// due to renewal lock (i.e. because a renewal is already happening). This lack of
+			// error is by intention to force cache invalidation as though it has renewed.
 			err := cert.Config.RenewCert(allowPrompts)
```
mholt: @cretz Just a thought: what if another renewal process updates the certificate between the beginning of this for loop (above on line 67) and actually calling RenewCert? Even though we have a read lock on the certCache, something else could have renewed it by now, and finished, which would cause this certificate to be renewed twice.

Could you double-check my thinking here and see if that's a plausible 'race' condition? I wonder if the lock needs to be put over this whole function. (It's still early here; I could be wrong...)

cretz: I thought about this. It appears RenewCert re-obtains the site data from storage but yeah, it doesn't check if it still needs to be renewed. So yes, in theory it could double renew here. With how the RenewCertificate function is set up, I am not sure of the best place to place the lock. Open to suggestions. If I place the lock in the for loop (remember it is a domain specific lock), you still suffer the problem where it might need to reload from cert cache. Instead, maybe we just do a "continue" on renewal lock?

mholt: Can you move the logic that checks the expiration date into that method, where the expiration check happens inside the lock?

cretz: That's what I'm thinking. I'll send across another PR at some point.

caddytls/client.go
```diff
@@ -204,28 +202,33 @@ Attempts:
...
-		return err
-	}
-	keyBytes, err := ioutil.ReadFile(storage.SiteKeyFile(name))
-	if err != nil {
+	// We must lock the renewal with the storage engine
+	if lockObtained, err := storage.LockRegister(name); err != nil {
 		return err
+	} else if !lockObtained {
+		log.Printf("[INFO] Certificate for %v is already being renewed elsewhere", name)
+		return nil
```
mholt: @cretz I've been looking at this for a new class of Caddy plugins for TLS storage. In this branch, some other instance is already renewing the certificate, but here we return nil instead of waiting until the renew finishes and then loading it below and continuing along. When does the renewed certificate get loaded? What if the lock was blocking instead?

cretz: @mholt - I simply added this for good measure, but best I can see, this method is never called anyways. See `Config::obtainCertName` and `Config::renewCertName` for where renewal actually appears to happen.

mholt: Okay, thanks @cretz. Was there a specific reason that LockRegister() was not put into `ACMEClient.Obtain()` but was put into `ACMEClient.Renew()`?

cretz: @mholt - Tbh, I don't remember exactly why. If I had to guess, it is because I wanted to lock the storage more than the renewal/registration process. `Obtain` only uses a single (atomic) `StoreSite` whereas `Renew` does a `LoadSite` and then a `StoreSite` later and I need to make that combination atomic.

PR #919, opened 2016-07-04
Add the ACL directive

atenart:
ACL is a middleware allowing to control access of locations to certain client addresses. See https://github.com/atenart/caddy-acl.

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -343,6 +343,7 @@ var directives = []string{
 
 	// directives that add middleware to the stack
 	"log",
+	"acl",      // github.com/atenart/caddy-acl
 	"gzip",
 	"locale", // github.com/simia-tech/caddy-locale
 	"errors",


abiosoft:
Is this not similar to what [ipfilter](https://caddyserver.com/docs/ipfilter) does ?

atenart:
@abiosoft It's similar but with a different implementation. I personally find ipfilter quite odd: it does not deal with CIDR subnets and do not uses golang net functions to check client IPs against the allowed/denied subnets. Patching ipfilter would be quite difficult, to keep its way of dealing with subnets in addition to the one proposed here.

I do not intend to replace ipfilter, as some may be happy with it, but I'd like to give an alternative closer to what's usually available when dealing with subnets in other softwares/devices.

mholt:
@atenart Thanks for the PR! I agree with @abiosoft this seems quite similar to ipfilter and we do not want to duplicate or overlap functionality at this amount I think.

> Patching ipfilter would be quite difficult, to keep its way of dealing with subnets in addition to the one proposed here.

Would a breaking change to ipfilter be accepted? Especially if it was an improvement? /cc @pyed  Or do these in fact have different purposes?

pyed:
`ipfilter` serves the same propose and more.

it is right that `ipfilter` doesn't support CIDR notation, but it supports the better way of doing it in my opinion where you can write just `192.168.1` for `192.168.1.0/24` or just `32` for `32.0.0.0/8`

also recently we added the use of multiple `ipfilter` blocks which allows more complicated use cases, thanks to @Merccy

`ipfilter` is [well-tested](https://github.com/pyed/ipfilter/blob/master/ipfilter_test.go), if you have a use-case that `ipfilter` fall short with, let's talk about it at https://github.com/pyed/ipfilter

atenart:
@pyed I don't agree, but I understand your point. Let's close this PR.

mholt:
@atenart And just to clarify, I think your plugin looks great; it's just that we're currently working through a transition with how add-ons (now called "plugins") work with Caddy, including the policies around adding them to the Caddy source or the download page.

Anyone is of course welcome to write plugins for Caddy, but we use discretion when choosing which ones to add to the download page or to the Caddy source. I don't intend to follow the "app store" model because of the tight integration with Caddy and also for the sake of the user's experience. I think this methodology will encourage developers to improve plugins already on the download page rather than creating discordant plugins leading to a dizzying number of choices for users.

We'll build a plugin manager that people can install for custom Caddy builds, including installing plugins that aren't on the download page, but we haven't gotten around to that yet. :)

atenart:
@mholt Yes, I really understand your point of not having duplicated plugins, no worries :) I was not expecting to get on the Caddy download page, just to be able to integrate my middleware nicely without having a patch on top of Caddy.

Your plugin manager idea seems nice, keep up the good work :)

@pyed Well,  ipfilter does have use cases not supported: IPv6, or nice way to describe other than /32, /24, /16 and /8 IPv4 subnets. But that's use cases not supported by design. I'm usually a "patch the existing" guy, but I'm not convinced here. Don't get me wrong, ipfilter provides a nice feature and if people are happy with it, keep it :)

PR #920, opened 2016-07-05
Changing refs from /usr/bin to /usr/local/bin

ianwalter:
`/usr/local/bin` is a more appropriate location for the Caddy binary to be installed. I believe `/usr/bin` is more geared towards distribution-managed programs and so there's a possibility that that directory may be changed during a distribution upgrade while `/usr/local/bin` wouldn't be.

As a side note, this change would help me in my efforts to automate the installation of Caddy on my servers.


lenovouser:
I can confirm that `/usr/local/bin` is a more appropriate install location for applications. At least on all the *NIX-Systems I know.

mholt:
LGTM. Thanks!

PR #929, opened 2016-07-13
import should get absolute path before glob

pedronasser:


mholt:
Thanks Pedro! What does this fix, exactly? Can you provide some context for this change?

pedronasser:
Giving some context:

If you run caddy using a configuration from a different directory (not the cwd)
`caddy -conf ../other/path`
And inside that configuration you have:
```
# glob relative to this caddyfile location
import ../other/different/path/*.conf
```
Caddy might not resolve that import correctly before using glob.

mholt:
Ah, got it. Thanks for the fix!

Issue #916, opened 2016-07-03
Wildcard host bind with proxy cause hostname been cached.


vibbow:
Caddy version: 0.9 beta 2
Running on Windows Server 2012 R2

Caddy config:
```
https://*.example.com {
	gzip
	header / -Server
	tls cert.crt cert.key
	header / Strict-Transport-Security "max-age=31536000; includeSubDomains; preload"
	proxy / https://internal.example.com {
		transparent
		header_downstream -Server ""
	}
}
```

Expect behavior:
When I access a.example.com, the proxy pass a.example.com as host to backend server
When I access b.example.com, the proxy pass b.example.com as host to backend server.

Real behavior:
The first host I access will be cached in proxy.
For example, after caddy are started, I first access a.example.com, everything works fine.
Then I access b.example.com, the proxy still pass a.example.com as host to backend server.

mholt:
Thanks for trying the beta! My guess is this is either due to some sort of persistent connection (reusing the http.Client maybe?) or something wrong with how we're setting the headers going upstream. We'll look into it! (I'm away from my computer for a bit.)

mholt:
I am able to replicate this behavior. I used this Caddyfile though to simply slightly:

```plain
:2015
gzip
header / -Server
tls cert.pem key.pem
proxy / https://localhost:2016 {
	transparent
	header_downstream -Server ""
}
```

And I added a fmt.Println to reverseproxy.go so it printed `outreq.Host` when the request was proxied to the backend. First I loaded `https://localhost:2015/abc` and saw `localhost:2015` being passed upstream, as expected. Then I loaded `https://127.0.0.1:2015/abc` and still saw `localhost:2015` being passed upstream, which was _not_ expected. /cc @evermax

mholt:
Leaving this issue open since @pedronasser and I agree there is a better way to fix this.

martindale:
I'd like to bump this, hoping there's an easy fix.  It's the only remaining issue before we switch to Caddy completely.

martindale:
@mholt can you point us in the right direction?  Perhaps we can write a patch ourselves.

mholt:
I've been working on this today... seems the replacer is not replacing a `{host}` value on the second request. (The function that returns r.Host from within the replacer's map is not being called.)

mholt:
[...]

PR #931, opened 2016-07-14
fix transparent host header #916

pedronasser:


mholt:
Hey Pedro, thanks for working on this.

I'm actually not sure this is the best way to go about fixing this. I don't really like the idea of introducing a "transparent" mode (the keyword is just a name for the preset, all it does it set headers). I feel like there is an error underlying the logic we already have in place, which should be fixed, instead of a flag to work around it.

Also if you could allow ~24-48 hours on average so one or two other collaborators have a chance to review a pull request, that would be helpful. :) I'll try to remember to add that to our guidelines for future reference.

pedronasser:
Yes, I actually thought that this was not the best and final solution for this problem. But I think it's a solution, and I actually like the idea of solving the problem first (if there is a simple solution) then look for a complete solution. I haven't refactored everything because I think would be better to be someone that knows the best this directive to refactor it.

Sorry for not waiting that amount of time to merge it. I thought wouldn't be a great deal because the change is not so big. But I will wait more next time.

And sorry about the amount of commits, I forgot to squash it. (I'll pay more attention later)

Thanks for the feedback :)

mholt:
Sounds good! I usually like to let each PR get reviewed by at least one other dev before it gets merged. Sometimes waiting a day or two is needed, but at least this change is small and non-destructive.

I think @evermax was looking at this issue too, maybe between you and him you can find the error in the existing code and test against it!

PR #935, opened 2016-07-17
Allow for UDP servers

miekg:
Extend the Server interface with ServePacket and ListenPacket - this is
in the same vein as the net package.

Plumb the packetconn through the start and restart phases.

Rename RestartPair to RestartTriple as it now also contains a Packet.
Not that these can now be nil, so we need to check for that when
restarting.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -148,12 +148,24 @@ func (i *Instance) Restart(newCaddyfile Input) (*Instance, error) {
 	}
 
 	// Add file descriptors of all the sockets that are capable of it
-	restartFds := make(map[string]restartPair)
+	restartFds := make(map[string]restartTriple)
 	for _, s := range i.servers {
 		gs, srvOk := s.server.(GracefulServer)
 		ln, lnOk := s.listener.(Listener)
-		if srvOk && lnOk {
-			restartFds[gs.Address()] = restartPair{server: gs, listener: ln}
+		pc, pcOk := s.packet.(PacketConn)
+		if srvOk {
+			if lnOk && pcOk {
+				restartFds[gs.Address()] = restartTriple{server: gs, listener: ln, packet: pc}
+				continue
+			}
+			if lnOk {
+				restartFds[gs.Address()] = restartTriple{server: gs, listener: ln}
+				continue
+			}
+			if pcOk {
+				restartFds[gs.Address()] = restartTriple{server: gs, packet: pc}
+				continue
+			}
 		}
 	}
 
@@ -223,28 +235,6 @@ func listenerAddrEqual(ln net.Listener, addr string) bool {
 	return false
 }
 
-/*
-// TODO: We should be able to support UDP servers... I'm considering this pattern.
-
-type UDPListener struct {
-	*net.UDPConn
-}
-
-func (u UDPListener) Accept() (net.Conn, error) {
-	return u.UDPConn, nil
-}
-
-func (u UDPListener) Close() error {
-	return u.UDPConn.Close()
-}
-
-func (u UDPListener) Addr() net.Addr {
-	return u.UDPConn.LocalAddr()
-}
-
-var _ net.Listener = UDPListener{}
-*/
-
 // Server is a type that can listen and serve. A Server
 // must associate with exactly zero or one listeners.
 type Server interface {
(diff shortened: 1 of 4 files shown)


miekg:
Yes, no tests (yet). Oh and I prolly forgot to change the tests..

caddy.go
```diff
@@ -253,12 +243,24 @@ type Server interface {
 	// connections.
 	Listen() (net.Listener, error)
 
+	// ListenPacket starts listening by creating a new packetconn
+	// and returning it. It does not start accepting
+	// connections.
+	ListenPacket() (net.PacketConn, error)
+
```
mholt: It would probably be worth explaining in the godoc whether this is required for all server types to implement or if this can be a no-op for TCP servers, etc.

caddy.go
```diff
@@ -253,12 +243,24 @@ type Server interface {
...
 	// loop begins. Serve blocks indefinitely, or in other
 	// words, until the server is stopped.
 	Serve(net.Listener) error
+
+	// ServePacket starts serving using the provided packetconn.
+	// Serve must start the server loop nearly immediately,
+	// or at least not return any errors before the server
+	// loop begins. Serve blocks indefinitely, or in other
+	// words, until the server is stopped.
+	ServePacket(net.PacketConn) error
```
mholt: Change instances of `Serve` to `ServePacket`. And again, explain when this should be implemented vs. no-op; if only TCP is used, maybe only Listen() and Serve() need to be implemented? Should ServePacket always be implemented if ListenPacket is (i.e. do they pair up)? etc.

caddy.go
```diff
@@ -299,6 +301,15 @@ type Listener interface {
...
 }
 
+// PacketConn is a net.PacketConn with an underlying file descriptor.
+// A server's packetconn should implement this interface if it is
+// to support zero-downtime reloads (in sofar this holds true for datagram
+// connections).
+type PacketConn interface {
+	net.PacketConn
+	File() (*os.File, error)
+}
```
mholt: Good godoc here.

caddy/caddymain/run.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/mholt/caddy"
 	// plug in the HTTP server type
 	_ "github.com/mholt/caddy/caddyhttp"
+	// plugin the DNS server type
+	_ "github.com/miekg/caddy-coredns/coredns"
```
mholt: Yep, this is exactly how easy it is to add the DNS server type!

However we probably won't ship the DNS server _by default_ (HTTP will remain the default). I've already updated the Download page so DNS can be checked very easily just like other features. So we can remove this line and people can still get the DNS server quite easily. Hope that's okay.

miekg: [ Quoting <notifications@github.com> in "Re: [mholt/caddy] Allow for UDP ser..." ]
>> @@ -19,6 +19,8 @@ import (
>>  	"github.com/mholt/caddy"
>>  	// plug in the HTTP server type
>>  	_ "github.com/mholt/caddy/caddyhttp"
>> +	// plugin the DNS server type
>> +	_ "github.com/miekg/caddy-coredns/coredns"
>
>Yep, this is exactly how easy it is to add the DNS server type!
>
>However we probably won't ship the DNS server _by default_ (HTTP will remain the default). I've already updated the Download page so DNS can be checked very easily just like other features. So we can remove this line and people can still get the DNS server quite easily. Hope that's okay.


You're right, I forgot to remove this line -- *I* need it for testing :)

/Miek

--
Miek Gieben

mholt:
Great! Did you test the restarts to make sure they still work?

miekg:
[ Quoting <notifications@github.com> in "Re: [mholt/caddy] Allow for UDP ser..." ]
>Great! Did you test the restarts to make sure they still work?

Yes, restarts work. For both caddy -type=http and -type=dns.

miekg:
Pushed updated godoc.

mholt:
I hereby dub thee, "Caddy DNS master" :smile:

I'll give you push rights in the interest of preserving this functionality. Thank you Miek!

miekg:
Hehe :-) Cool and thank you!

PR #947, opened 2016-07-20
Headers Patch 

### LINKED ISSUE #818: Add Headers instead of Set Headers
Currently the way that cloudflare is supporting server push is by having the server (caddy) set a Link header for each item to be pushed.

Their example:
`Link: </asset/to/push.js>; rel=preload;`

The problem with Caddy is that it sets each header of that name (Link) each time, instead of adding it. So if you wanted to tell cloudflare to push multiple files it isn't currently possible. I have tested it myself by changing Line 30 of the headers.go file from:
`w.Header().Set(header.Name, replacer.Replace(header.Value))`
to
`w.Header().Add(header.Name, replacer.Replace(header.Value))`

Is this something worth changing permanently for caddy?

https://github.com/mholt/caddy/blob/da016f8d5a14cdce2925556ae0e5a7bc72ddbb1a/middleware/headers/headers.go#L30

gsquire:
This is a trivial patch that allows for multiple header values to be used. Fixes #818.

### DIFF
--- a/caddyhttp/header/header.go
+++ b/caddyhttp/header/header.go
@@ -27,7 +27,7 @@ func (h Headers) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 				if strings.HasPrefix(header.Name, "-") {
 					w.Header().Del(strings.TrimLeft(header.Name, "-"))
 				} else {
-					w.Header().Set(header.Name, replacer.Replace(header.Value))
+					w.Header().Add(header.Name, replacer.Replace(header.Value))
 				}
 			}
 		}


mholt:
Thanks Garrett! Unfortunately with this patch we can no longer set headers; only add.

We need something like proxy's header control where we use `+` to add and `-` to remove headers, with the default being to set the header.

gsquire:
Ah I misread. I'll update it then!

gsquire:
If I understood you correctly, the changes should be good now. Thanks for the help.

mholt:
Now we're talkin'. Will look at this soon, thanks!

mholt:
This looks good. Squash at will and we can merge it in!

gsquire:
Closing in favor of #949.

PR #949, opened 2016-07-21
HTTP Headers

### LINKED ISSUE #818: Add Headers instead of Set Headers

gsquire:
Rebase of #947 after I incorrectly rebased that PR :/

Thank goodness for `git reflog`.

Fixes #818

### DIFF
--- a/caddyhttp/header/header.go
+++ b/caddyhttp/header/header.go
@@ -24,8 +24,12 @@ func (h Headers) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 	for _, rule := range h.Rules {
 		if httpserver.Path(r.URL.Path).Matches(rule.Path) {
 			for _, header := range rule.Headers {
+				// One can either delete a header, add multiple values to a header, or simply
+				// set a header.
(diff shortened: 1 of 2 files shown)


mholt:
Oops. Wrong button. Thanks!

gsquire:
You're welcome. Congratulations on the project. I'm happy it's getting a lot of support 👍

PR #960, opened 2016-07-24
Change position of locale directive

phifty:
First, great job on the 0.9 release! It seems caddy's path lead into a bright future. Thanks also for including the locale plugin.

Trying it, I've figured out, that there might be a problem with the order of the directives. In the typical use case, the result of the locale detection might be used in the `rewrite` and `log` plugin. If I'm not mistaken, it makes sense to put the `locale` directive before those.

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -342,11 +342,11 @@ var directives = []string{
 	"git",    // github.com/abiosoft/caddy-git
 
 	// directives that add middleware to the stack
+	"locale", // github.com/simia-tech/caddy-locale
 	"log",
 	"rewrite",
 	"ext",
 	"gzip",
-	"locale", // github.com/simia-tech/caddy-locale
 	"errors",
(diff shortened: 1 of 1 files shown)


mholt:
Looks reasonable to me. Thank you!

phifty:
Ah, one more thing: after re-ordering the plugins, the extention of the replacer doesn't seem to work.

https://github.com/simia-tech/caddy-locale/blob/master/middleware.go#L38

The provided `ResponseWriter` is not of type `*httpserver.ResponseRecorder` yet. Any hints how to solve this?

mholt:
@phifty Not at the moment. :-/ Maybe you could make the Replacer (and httpserver.ResponseRecorder) yourself in this middleware, but that seems unnecessarily complex.

Maybe a better solution will be introduced using context.Context after Go 1.7 is released.

phifty:
Ok, I've removed the replacer code. Could you update the documentation?

mholt:
Sure, just did it locally; will go out with next push.

PR #966, opened 2016-07-28
Proxy middleware response codes less than 400 (fix #801).

### LINKED ISSUE #801: Rewrite directive can only return 200 status code for any 2xx success class
*If you are filing a bug report, please answer these questions. If your issue is not a bug report, you do not need to use this template. Either way, please consider donating if we've helped you. Thanks!*

$ ./caddy -version
Caddy 0.8.3

Get return code 2xx via rewrite directive 

```
$ cat Caddyfile2 
localhost
rewrite {
    regexp ^/1
    status 403
}
rewrite {
    regexp ^/2
    status 201
}
rewrite {
    regexp ^/3
    status 202
}
rewrite {
    regexp ^/4
    status 203
}
rewrite {
    regexp ^/5
    status 204
}
rewrite {
    regexp ^/6
    status 401
}
rewrite {
    regexp ^/7
    status 402
}
```

./caddy -conf Caddyfile2

$ curl -I http://localhost:2015/5
HTTP/1.1 204 No Content

$ curl http://localhost:2015/5 -I
HTTP/1.1 200 OK
Server: Caddy

vchimishuk:
I'm trying to fix https://github.com/mholt/caddy/issues/801 and finish Sean's work https://github.com/mholt/caddy/pull/802

It looks like that we have no any reason to block middleware from returning any response code, and not only after 400.  Docs also says that only 0 indicates that response has been already written [1]. Currently default 200 (written by Go's net.http.Server internally) is sent.

@mholt, what do you think?

[1]  https://github.com/mholt/caddy/blob/39030d9e1bbbc1e31be49ab7f34bc800f2f2f8d8/middleware/middleware.go#L16-L33

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -210,7 +210,7 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	status, _ := s.serveHTTP(w, r)
 
 	// Fallback error response in case error handling wasn't chained in
-	if status >= 400 {
+	if status != 0 {
 		DefaultErrorFunc(w, r, status)
 	}
 }


mholt:
Well, error codes are only >= 400 -- why serve an error page for a 200 response?

I think the way to fix 801 is probably in the rewrite directive itself, although I admit I haven't looked into it too much yet.

vchimishuk:
Now I see: ServeHTTP results indicates errors only. I thought that middleware can produces response in two ways: writes response by itself or returns as ServeHTTP result.

AFAIU, in this case rewrite looks broken, it should writes response by itself, and doesn't return status, since it is not an error but correctly processed request. I can try this way.

vchimishuk:
Here is what I came up with. https://github.com/vchimishuk/caddy/commit/cd301e3a379031198971e45f38961f2ce1de56b8
It fixes broken 2xx rewrites and looks logically more correct.

P. S.
IMHO, it is much more Go idiomatic way to return special error type from ServeHTTP than (int, error) pair.

mholt:
I think that's probably okay, but I'll ask @abiosoft to vet it for correctness when he has a chance, too.

> P. S.
> IMHO, it is much more Go idiomatic way to return special error type from ServeHTTP than (int, error) pair.

In fact, this was inspired from [a post on the Go blog](https://blog.golang.org/error-handling-and-go), except we return two values instead of one. Indeed, there are many ways to do this. I just settled on this over a year ago when I first started the project and it has worked pretty well so far.

mholt:
Actually, will [this change](https://github.com/vchimishuk/caddy/commit/cd301e3a379031198971e45f38961f2ce1de56b8) honor custom error pages in the case of an error status, since it's writing its own error message? Can test it with the `errors` directive...

vchimishuk:
Yes, you are absolutely right about errors. Then forget it. :)

mholt:
Yeah :-/ I still think it's weird to rewrite to a non-error status. Will close this for now, until a good solution can be discovered. But thank you for working on it!

PR #973, opened 2016-07-30
redirect to site root explicitly

tw4452852:
fix issue #969


mholt:
This issue is causing my head to spin, I need to look into it some more. I'm afraid that a redirect here might cause redirect loops in some cases but I haven't yet confirmed that. I also wonder if there's a way to correct the logic we already have rather than adding a new condition.

Right now, this redirect assumes that any path prefix leads to a directory (trailing "/") which I'm not sure is always the case. It makes sense in theory, I suppose, for base paths to be a directory (because, it is a *base path*, not a final endpoint necessarily)...

I just need to think about this some more. Thanks for working on it!

mholt:
So, Tw, I think I'm going to decline this PR simply because, although it 'fixes' the issue linked to, I think it's a bit hasty to assume that the base path must always be a directory and should always have the trailing slash.

I don't yet know a better solution, but I'll recommend in the issue that we just update the documentation for now. Thank you!!

tw4452852:
Ok @mholt , make sense.

PR #976, opened 2016-07-30
Re-enable HTTP/2 for Go 1.7.

### LINKED ISSUE #975: no HTTP/2 on Go 1.7
My "production" server, with some proprietary patches, serves only HTTP/1.1 instead of HTTP/2. I can't tell since when. Maybe since I switched from 0.8 to 0.9, maybe since my recent Let's Encrypt cert cycle; I can't tell really.

QUIC works.

I have tried this again with vanilla Caddy with a minimal setup, still HTTP/2 does not work. I've also tried to `git bisect`, but I gave up when I realized that some API of dependent libraries changed since 0.9 released, letting my naïve bisect build script fail. 

current master
Caddy 0.9.0 (+7157bdc Sat Jul 30 13:04:36 UTC 2016)

GET via http/2

```
caddy:2015 { tls self_signed }
```

```
# inside docker container named caddy
echo "caddy:2015 { tls self_signed }" | ./caddy -log stderr
```

```
# inside docker container linked to container caddy, i.e. hostname caddy resolves.
curl --insecure --http2 -v https://caddy:2015/ 2>&1 | grep ALPN
```

```
* ALPN, offering h2
* ALPN, offering http/1.1
* ALPN, server accepted to use h2
```
( no log output )

```
* ALPN, offering h2
* ALPN, offering http/1.1
* ALPN, server did not agree to a protocol
```
( no log output )

See above. "Expected" output generated by 

```
curl --insecure --http2 -v https://google.com/ 2>&1 | grep ALPN
```

wjkohnen:
This re-enables HTTP/2 with Go 1.7. I am defensively setting the map to {"h2"}, if caddytls.MakeTLSConfig didn't set anything beforehand. Does that make sense?

It works on my public server.

fixes #975

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -92,6 +92,10 @@ func NewServer(addr string, group []*SiteConfig) (*Server, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Since Go 1.7 HTTP/2 is enabled only if TLSConfig.NextProtos includes the string "h2".
+	if HTTP2 && s.Server.TLSConfig != nil && len(s.Server.TLSConfig.NextProtos) == 0 {
+		s.Server.TLSConfig.NextProtos = []string{"h2"}
+	}
 
 	// Compile custom middleware for every site (enables virtual hosting)
 	for _, site := range group {


caddyhttp/httpserver/server.go
```diff
@@ -92,6 +92,10 @@ func NewServer(addr string, group []*SiteConfig) (*Server, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Since Go 1.7 HTTP/2 is enabled only if TLSConfig.NextProtos includes the string "h2".
+	if HTTP2 && s.Server.TLSConfig != nil && len(s.Server.TLSConfig.NextProtos) == 0 {
+		s.Server.TLSConfig.NextProtos = []string{"h2"}
+	}
```
mholt: At first I was thinking this logic should go into the `caddytls.MakeTLSConfig()` function, but this pertains specifically to HTTP, not all TLS. So I think this is actually in the right place. :+1: Setting "h2" for another type of server might mess things up.

mholt:
Cheers Wolfgang!

The only issue I wonder about -- not a showstopper for now -- is if the MakeTLSConfig() function will ever set next protocols so that it's not empty. In that case, HTTP/2 would not be enabled again. But I can't see why MakeTLSConfig would be modified to set NextProtos because that function is agnostic of the application layer (like HTTP). So I think I'll merge this in, but we should probably get some tests in place to ensure HTTP/2 is enabled when it is supposed to be. (Although these tests will require a new harness that doesn't exist yet, and is a different topic entirely.)

wjkohnen:
Hey Matt, exactly my thoughts and that's why I didn't want to merge w/o review. 

> I wonder about is if the MakeTLSConfig() function will ever set next protocols so that it's not empty. In that case, HTTP/2 would not be enabled again.

Maybe "hard setting" would be better in order to "fail early" when someone fiddles with NextProtos() inside MakeTLSConfig(). But then again I think spending some energy on some sort of test harness is a better investment in order to prevent regressions. When I was writing this fix I was looking at the test files at the vicinity of this code and didn't find anything except trivial or very sparse special case tests. So I've stopped there. Or to put it differently: Lack of tests discourages writing tests.

PR #978, opened 2016-07-31
added ip_hash load balancing WIP

krishamoud:
This would allow for "sticky sessions" based on client IP.  The basic algorithm is to hash the client IP address and then take `hash % len(pool)` which gives a fairly evenly distributed load but with the assurance that the request will be routed to the same host every time.

### DIFF
--- a/caddyhttp/proxy/policy.go
+++ b/caddyhttp/proxy/policy.go
@@ -1,8 +1,11 @@
 package proxy
 
 import (
+	"hash/fnv"
 	"math"
 	"math/rand"
+	"net"
+	"net/http"
 	"sync"
 )
 
@@ -11,20 +14,21 @@ type HostPool []*UpstreamHost
 
 // Policy decides how a host will be selected from a pool.
 type Policy interface {
-	Select(pool HostPool) *UpstreamHost
+	Select(pool HostPool, r *http.Request) *UpstreamHost
 }
 
 func init() {
 	RegisterPolicy("random", func() Policy { return &Random{} })
 	RegisterPolicy("least_conn", func() Policy { return &LeastConn{} })
 	RegisterPolicy("round_robin", func() Policy { return &RoundRobin{} })
+	RegisterPolicy("ip_hash", func() Policy { return &IPHash{} })
 }
 
 // Random is a policy that selects up hosts from a pool at random.
 type Random struct{}
 
 // Select selects an up host at random from the specified pool.
-func (r *Random) Select(pool HostPool) *UpstreamHost {
+func (r *Random) Select(pool HostPool, request *http.Request) *UpstreamHost {
 
 	// Because the number of available hosts isn't known
 	// up front, the host is selected via reservoir sampling
@@ -53,7 +57,7 @@ type LeastConn struct{}
 // Select selects the up host with the least number of connections in the
 // pool.  If more than one host has the same least number of connections,
 // one of the hosts is chosen at random.
-func (r *LeastConn) Select(pool HostPool) *UpstreamHost {
+func (r *LeastConn) Select(pool HostPool, request *http.Request) *UpstreamHost {
 	var bestHost *UpstreamHost
 	count := 0
 	leastConn := int64(math.MaxInt64)
@@ -86,7 +90,7 @@ type RoundRobin struct {
 }
 
 // Select selects an up host from the pool using a round robin ordering scheme.
-func (r *RoundRobin) Select(pool HostPool) *UpstreamHost {
+func (r *RoundRobin) Select(pool HostPool, request *http.Request) *UpstreamHost {
 	poolLen := uint32(len(pool))
 	r.mutex.Lock()
 	defer r.mutex.Unlock()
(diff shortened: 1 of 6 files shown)


caddyhttp/proxy/policy.go
```diff
@@ -100,3 +104,35 @@ func (r *RoundRobin) Select(pool HostPool) *UpstreamHost {
...
+func (r *IPHash) Select(pool HostPool, request *http.Request) *UpstreamHost {
+	var clientIP string
+	poolLen := uint32(len(pool))
+	clientIP, _, err := net.SplitHostPort(request.RemoteAddr);
+	if err != nil {
+		return nil
+	}
+	hash := hash(clientIP)
+	for {
+		if poolLen == 0 { break }
```
vchimishuk: This is definitely is not Go coding style. Please, use gofmt to format your code before commit.

caddyhttp/proxy/proxy_test.go
```diff
@@ -664,15 +664,15 @@ func basicAuthTestcase(t *testing.T, upstreamUser, clientUser *url.Userinfo) {
 	}))
 	defer backend.Close()
 
-	backURL, err := url.Parse(backend.URL)
+	backUrl, err := url.Parse(backend.URL)
```
vchimishuk: backURL had a right name. You can use golint to check your naming.

caddyhttp/proxy/upstream_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"strings"
 	"testing"
 	"time"
-
+	"net/http"
```
vchimishuk: You should gofmt code first. It will also organize your imports.

caddyhttp/proxy/policy.go
```diff
@@ -100,3 +104,35 @@ func (r *RoundRobin) Select(pool HostPool) *UpstreamHost {
...
+
+
+// Select selects an up host from the pool using a round robin ordering scheme.
+func (r *IPHash) Select(pool HostPool, request *http.Request) *UpstreamHost {
+	var clientIP string
+	poolLen := uint32(len(pool))
+	clientIP, _, err := net.SplitHostPort(request.RemoteAddr);
+	if err != nil {
+		return nil
+	}
```
mholt: It's possible that the remote address will not have a port, in which case you would get a non-nil error; you might be fine to assume that the remote address is only a host or IP; I do this in a few places in Caddy's code.

mholt:
Hey @krishamoud, thanks for making the PR! This is looking good - love the tests - except for the things mentioned in comments; use `gofmt -s` and `go lint` to make the CI builds pass and check out the other inline remarks and we'll be in a better place to merge.

krishamoud:
@mholt Thanks!  I'll fix those issues and have it pushed soon.

caddyhttp/proxy/policy.go
```diff
@@ -100,3 +103,30 @@ func (r *RoundRobin) Select(pool HostPool) *UpstreamHost {
...
+func hash(s string) uint32 {
+	h := fnv.New32a()
+	h.Write([]byte(s))
+	return h.Sum32()
+}
+
+// Select selects an up host from the pool using a round robin ordering scheme.
+func (r *IPHash) Select(pool HostPool, request *http.Request) *UpstreamHost {
+	poolLen := uint32(len(pool))
+	hash := hash(request.RemoteAddr)
```
s7v7nislands: what about strings.Split(requests.RemoteAddr, ":")[0] ?

I think it's better not include the port.

s7v7nislands:
http://nginx.org/en/docs/http/ngx_http_upstream_module.html#ip_hash

using request.RemoteAddr is too strict?

caddyhttp/proxy/policy.go
mholt: You may be right. I would recommend `net.SplitHostPort()` instead, however. IPv6 addresses would otherwise be problematic.

krishamoud: I was originally using `net.SplitHostPort()`  which looked like this.

```
clientIP, _, err := net.SplitHostPort(request.RemoteAddr);
if err != nil {
	return nil
}
hash := hash(clientIP)
```
However, `It's possible that the remote address will not have a port, in which case you would get a non-nil error`, was brought to my attention.  I'm happy to change it though.

mholt: If stripping the port is necessary, use `net.SplitHostPort()`; when I do, I handle the error case there of no port by just assuming the whole RemoteAddr is only a host or IP (since there was no port, might as well).

```go
host, _, err := net.SplitHostPort(request.RemoteAddr)
if err != nil {
    host = request.RemoteAddr
}
hash := hash(host)
```

Something like that.

vchimishuk:
Actually Nginx allows you to configure it for hashing 3 or 4 octets.
They have here some details why it is done that way https://www.nginx.com/blog/choosing-nginx-plus-load-balancing-techniques/ 
Maybe it can be better to add some flexibility to the user here? In this case we can add two policies: "ip_hash3" and "ip_hash4". Or do it in some other way.

mholt:
@vchimishuk You're right, and I'll quote the relevant portion here for reference:

> If the client has an IPv6 address, the hash is based on the entire address. If it has an IPv4 address, the hash is based on just the first three octets of the address. This is designed to optimize for ISP clients that are assigned IP addresses dynamically from a subnetwork (/24) range. In case of reboot or reconnection, the client’s address often changes to a different one in the /24 network range, but the connection still represents the same client, so there’s no reason to change the mapping to the server.

> If, however, the majority of the traffic to your site is coming from clients on the same /24 network, IP Hash doesn’t make sense because it maps all clients to the same server. In that case (or if you want to hash on all four octets for another reason), instead use the Hash method with the $remote_addr variable.

Let's start simple for now and just use the whole IP address, since I don't think the IP reassignment is _that_ common (my IP hasn't changed in months).

I think this is good to merge now. Thank you @krishamoud! I will grant you push rights so you can maintain this change and help by reviewing other pull requests, and this way you can work on branches in this repo without having to fork the project. Please remember to review [contributor guidelines](https://github.com/mholt/caddy/blob/master/CONTRIBUTING.md) so we can maintain a high-quality code base. Thanks!

krishamoud:
Thanks @mholt!  I'll change the policy in the future to hash the first 3 octets of IPv4 and the full IP of IPv6.  Hopefully something I can get to this weekend.

mholt:
Well, remember:

> If, however, the majority of the traffic to your site is coming from clients on the same /24 network  IP Hash doesn’t make sense because it maps all clients to the same server.

So maybe the behavior of hashing only the first 3 octets is not desirable.

Issue #979, opened 2016-08-01
Send downloads for protected content using X-Accel-Redirect


slightfoot:
#### 1. What version of Caddy are you running (`caddy -version`)?
Caddy 0.9.0 (7157bdc79d38e203386d7e12d7065d884166c431 Mon Aug 01 18:30:00 UTC 2016)

#### 2. What are you trying to do?
Send downloads for protected content using `X-Accel-Redirect`.

#### 3. What is your entire Caddyfile?
```text
localhost:2015
log stdout
errors stdout
fastcgi / 127.0.0.1:9000 php
rewrite {
	to {path} {path}/ /index.php/{path}
}
internal /protected
```

#### 4. How did you run Caddy (give the full command and describe the execution environment)?
`caddy -log stdout`
Test URL: `http://localhost:2015/download.txt?key=secret`

#### 5. What did you expect to see?
Download dialog in browser for the user to save the file because of the script provided the
`Content-Disposition: attachment` header. 

#### 6. What did you see instead (give full error messages and/or log)?
Content of file directly in browser.

#### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?
See attached zip file for example project that includes small example script and content for test URL above.

[test.zip](https://github.com/mholt/caddy/files/392741/test.zip)

slightfoot:
I've tracked it down to the [internal.internalResponseWriter.ClearHeader](https://github.com/mholt/caddy/blob/master/caddyhttp/internalsrv/internal.go#L71) which is removing all headers from the script results. Changing this function so that it just removes the `X-Accel-Redirect` header resolves my immediate issue.

The question is what is the expected behaviour. It might be nice to have compatibility with `X-Sendfile` for porting (my issue) [mod_xsendfile](https://tn123.org/mod_xsendfile/) just removes these headers. Code comment says `as we dropped all the content this field is not valid anymore!` which is a fair point.
```
X-Sendfile
Content-Length
Content-Encoding
```
[Nginx's sendfile topic](https://www.nginx.com/resources/wiki/start/topics/examples/xsendfile/) just states it will leave this set of headers unmodified. This doesn't mean it it will let other headers through. Just that these headers from the original will not be overwritten with the redirected content's headers.
```
Content-Type
Content-Disposition
Accept-Ranges
Set-Cookie
Cache-Control
Expires
```
I am going to submit a pull-request with my fix that will simply remove these headers.
```
X-Accel-Redirect
Content-Length
Content-Encoding
```
[...]

PR #980, opened 2016-08-01
Enable downloading of protected content. See issue #979

slightfoot:
Please see issue #979

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ Thumbs.db
 _gitignore/
 Vagrantfile
 .vagrant/
+/.idea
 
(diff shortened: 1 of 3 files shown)


mholt:
Thanks for the PR Simon! This looks good but perhaps we should have a test to assert that only the desired headers are removed. Could you write a test before we merge this?

slightfoot:
I'll see what I can do.

slightfoot:
@mholt Sorry for the delay. Tests added.

mholt:
Nice! Thanks for your contribution, @slightfoot! We'll give it a go.

PR #982, opened 2016-08-01
improve rlimit usage

s7v7nislands:
using syscall instead of exec

### DIFF
--- a/rlimit_posix.go
+++ b/rlimit_posix.go
@@ -15,10 +15,9 @@ func checkFdlimit() {
 	// Warn if ulimit is too low for production sites
 	rlimit := &syscall.Rlimit{}
 	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, rlimit)
-	if err == nil {
-		if rlimit.Cur < min {
-			fmt.Printf("WARNING: File descriptor limit %d is too low for production servers. "+
-				"At least %d is recommended. Fix with \"ulimit -n %d\".\n", rlimit.Cur, min, min)
-		}
+	if err == nil && rlimit.Cur < min {
+		fmt.Printf("WARNING: File descriptor limit %d is too low for production servers. "+
+			"At least %d is recommended. Fix with \"ulimit -n %d\".\n", rlimit.Cur, min, min)
 	}
+
 }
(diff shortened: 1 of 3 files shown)


mholt:
Hi, thanks for the PR. Can you explain why? I'm not aware of any issues with this code; why use syscall (which requires another file, protected with build tags) over os/exec?

Xe:
These limits only matter on `!windows`, windows has each process have an unlimited file descriptor count. This removes the branch and lifts it to a compiler time choice, which I don't think Go optimizes for. Trivial patch, no reason to reject it.

mholt:
Well I do agree the code is cleaner. We'll give this a try. Thank you!

rlimit_posix.go
```diff
@@ -0,0 +1,24 @@
...
+// checkFdlimit issues a warning if the OS limit for
+// max file descriptors is below a recommended minimum.
+func checkFdlimit() {
+	const min = 8192
+
+	// Warn if ulimit is too low for production sites
+	rlimit := &syscall.Rlimit{}
+	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, rlimit)
+	if err == nil {
+		if rlimit.Cur < min {
```
mholt: I know this might seem like a nit-pic, but can we combine this to one line: `if err == nil && rlimit.Cur < min {` - and save a level of indentation? I prefer that style over separate if statements.

s7v7nislands: ok, I will update it tomorrow.

s7v7nislands:
update the code style

mholt:
Excellent, thank you!

weingart:
@mholt: This should likely be changed to be: https://paperairoplane.net/?tag=ulimit

The ulimit you're invoking is a shell built-in, and the information it returns has little to no bearing as to the limit that *your* currently executing program has.  The shell could (through .bashrc, etc) set its own file descriptor rlimit.

mholt:
Nice link, I'll take a look. Thank you!

mholt:
@weingart Wait, so what needs to change exactly?

wjkohnen:
Nothing. This PR does what it should. Before we were measuring the ulimit of a subshell, which potentially did not inherit Caddy's ulimit. Thanks to this PR we're measuring the ulimit of the Caddy process itself. On POSIX, that is.

weingart:
I must have been very confused.  Yes, this does the right thing.  Sorry about the bad feedback.

PR #983, opened 2016-08-01
dist/init/linux-sysvinit: caddy for SysVinit

djvdorp:
In addition to `linux-upstart` and `linux-systemd`, I think this one might be very useful too.

The script is based on [this script](https://git.devuan.org/fredg/stuffs/blob/master/caddy/init.d/caddy) by @fredg, found via [Installation du serveur Caddy sous Devuan &middot;  Frédéric Galusik](http://galusik.xyz/installation-caddy-server-devuan/#démon:d7570338f345f168f3c50f22e7f8c47c). I have modified it into an extended version myself, since I had the need for this.

### DIFF
--- a/dist/init/linux-sysvinit/README.md
+++ b/dist/init/linux-sysvinit/README.md
@@ -1,0 +1,11 @@
+SysVinit conf for Caddy
+=======================
+
+Usage
+-----
+
+* Download the appropriate Caddy binary in `/usr/local/bin/caddy` or use `curl https://getcaddy.com | bash`.
+* Save the SysVinit config file in `/etc/init.d/caddy`.
+* Ensure that the folder `/etc/caddy` exists and that the subfolder `ssl` is owned by `www-data`.
+* Create a Caddyfile in `/etc/caddy/Caddyfile`
+* Now you can use `sudo /etc/init.d/caddy start|stop|restart|reload|status`.
--- a/dist/init/linux-sysvinit/caddy
+++ b/dist/init/linux-sysvinit/caddy
@@ -1,0 +1,93 @@
+#!/bin/sh
+### BEGIN INIT INFO
+# Provides:          caddy
+# Required-Start:    $local_fs $network $named $time $syslog
+# Required-Stop:     $local_fs $network $named $time $syslog
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: starts the caddy web server
+# Description:       starts caddy using start-stop-daemon
+### END INIT INFO
+
+# Original Author: Frédéric Galusik (fredg)
+# Maintainer: Daniel van Dorp (djvdorp)
+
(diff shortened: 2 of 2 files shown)


mholt:
Cool, thanks! Would you mind adding your GitHub username to the readme and/or the comment in the top of the script file so that people know who they can ping for support (as these scripts are not officially supported)?

(Wonder what's up with the failing build on Go tip. Maybe broken for now, looks unrelated.)

djvdorp:
Thanks for your quick reply! I have added the Original Author (credit where credit is due) and myself as the maintainer for this, do you also think this makes the most sense? I am open for improvements or suggestions.

mholt:
That'll do @djvdorp, thank you!

djvdorp:
Thanks for merging @mholt ! 

Speaking about maintaining the script, is there any possibility of getting Contributor access on this repo, or should it be done via more Pull Requests? (I am not sure how you handle this for Caddy)

mholt:
There ya go! Thanks for helping to maintain it.

dist/init/linux-sysvinit/README.md
```diff
@@ -0,0 +1,11 @@
...
+=======================
+
+Usage
+-----
+
+* Download the appropriate Caddy binary in `/usr/local/bin/caddy` or use `curl https://getcaddy.com | bash`.
+* Save the SysVinit config file in `/etc/init.d/caddy`.
+* Ensure that the folder `/etc/caddy` exists and that the subfolder `ssl` is owned by `www-data`.
+* Create a Caddyfile in `/etc/caddy/Caddyfile`
+* Now you can use `sudo /etc/init.d/caddy start|stop|restart|reload|status`.
```
weingart: This `sudo` command is an anti-pattern that will usually result in all sorts of pain at some point.  It is much better to use the `/usr/sbin/service` command, which will make sure that environment variables, ulimits, etc, are changed and done as the system itself would make it happen during startup, instead of whatever current directory, environments variables sudo allows, and so on, to be part of Caddy's invocation.

dist/init/linux-sysvinit/caddy
```diff
@@ -0,0 +1,93 @@
...
+
+reload() {
+    start-stop-daemon --stop --quiet --signal USR1 --pidfile $PIDFILE \
+        --name $NAME
+}
+
+status() {
+    if [ -f $PIDFILE ]; then
+        PID=`cat $PIDFILE`
+        if [ -z "`ps axf | grep ${PID} | grep -v grep`" ]; then
```
weingart: The `pgrep` command will be a much better choice here.  Or possibly `kill -0` if all you want to test that a process exists with the particular pid within `$PIDFILE`.

djvdorp:
Thank you for your feedback @weingart, I have tried to process this feedback in https://github.com/mholt/caddy/pull/1008

Issue #938, opened 2016-07-19
Exhaustion of dynamic TCP ports by reverse proxy middleware


tomasdeml:
#### 1. What version of Caddy are you running (`caddy -version`)?
v0.8.3 / v0.9.0
OS Windows Server 2012 R2 running on Azure instance 'Standard D3 v2' (4 cores, 14 GB memory)

#### 2. What are you trying to do?  
Measure performance of caddy reverse proxy middleware under heavy load using Apache Bench.

#### 3. What is your entire Caddyfile?
caddyfile_proxy:
```text
http:// {
	errors CaddyErrors.log
	header / {
		-Server
	}
	proxy / localhost:580 {
		proxy_header Host {host}
		proxy_header X-Real-IP {remote}
		proxy_header X-Forwarded-Proto {scheme}
	}
}
```
caddyfile_upstream:
```text 
http://:580 {
	errors CaddyErrors_upstream.log
    header / {
		-Server
	}
	root WebRoot
}
```

The `WebRoot` folder contains file `index.html`:
```html
<!DOCTYPE html><html><head><meta charset="utf-8"><meta name="format-detection" content="telephone=no"><meta name="viewport" content="width=device-width,initial-scale=1"><title>1234567 12345</title></head><body></body></html>
```

#### 4. How did you run Caddy (give the full command and describe the execution environment)?
```text
caddy.exe -conf=Caddyfile_proxy
```

and then in new shell:
```text
caddy.exe -conf=Caddyfile_upstream
```

#### 5. What did you expect to see?
Apache Bench completing with 0 failed requests.

#### 6. What did you see instead (give full error messages and/or log)?
Apache Bench with the following result (note the failed requests):
```text
This is ApacheBench, Version 2.3 <$Revision: 1748469 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking *** (be patient)

Server Software:        
Server Hostname:        ***
Server Port:            80

Document Path:          /
Document Length:        224 bytes

Concurrency Level:      10000
Time taken for tests:   104.036 seconds
Complete requests:      1000000
Failed requests:        953057
   (Connect: 0, Receive: 0, Length: 953057, Exceptions: 0)
Non-2xx responses:      953057
Keep-Alive requests:    1000000
Total transferred:      179237926 bytes
HTML transferred:       25764144 bytes
[...]

PR #984, opened 2016-08-01
Proxy: Add keepalive directive to proxy to set MaxIdleConnsPerHost on transport

nemothekid:
More discussion in #938 - 

if the requests coming into Caddy are much greater than the maximum idle connections allowed in the transport backend, Caddy will continue to create more and more connections until the OS runs out of sockets.

This PR adds a [keepalive](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) option, like nginx that allows you set the number of idle connections for a proxy backend, allowing you to increase it from the default, defined in golang as 2 (or disable it).

To keep consistent with the internal `MaxIdleConnsPerHost` option on `http.Transport`, setting `keepalive` to 0, will actually keep the default value of 2. Setting it to a negative number (-1) will actually disable keepalives (new connection for every request).

I'm not sure what the recommended values should be yet, as the golang ones may be tuned for normal client/server type requests.

[Edited by @mholt: Fixed link to keepalive docs]

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -122,7 +122,10 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 		rp.Transport = &http.Transport{
 			Dial: socketDial(target.String()),
 		}
-	} else if keepalive != 0 {
+	} else if keepalive != http.DefaultMaxIdleConnsPerHost {
+		// if keepalive is equal to the default,
+		// just use default transport, to avoid creating
+		// a brand new transport
 		rp.Transport = &http.Transport{
 			Proxy: http.ProxyFromEnvironment,
 			Dial: (&net.Dialer{
@@ -132,7 +135,7 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 			TLSHandshakeTimeout:   10 * time.Second,
 			ExpectContinueTimeout: 1 * time.Second,
 		}
-		if keepalive < 0 {
+		if keepalive == 0 {
 			rp.Transport.(*http.Transport).DisableKeepAlives = true
 		} else {
 			rp.Transport.(*http.Transport).MaxIdleConnsPerHost = keepalive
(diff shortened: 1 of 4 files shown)


caddyhttp/proxy/reverseproxy.go
```diff
@@ -122,10 +122,44 @@ func NewSingleHostReverseProxy(target *url.URL, without string) *ReverseProxy {
...
+				KeepAlive: 30 * time.Second,
+			}).Dial,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+		}
+		if keepalive < 0 {
+			rp.Transport.(*http.Transport).DisableKeepAlives = true
+		} else {
+			rp.Transport.(*http.Transport).MaxIdleConnsPerHost = keepalive
+		}
```
mholt: Phew, finally got a chance to review this! 😅 

Setting a negative keepalive is less intuitive than just using 0; what if we made a `const` with a value of 2, and passed that around instead of zeroes?

nemothekid: It's up to you honestly - the reason why I kept 0 (which I agree is less intuitive), is I prefer "the default" to be a the "golang default", and if in golang 1.8, the idleconns are changed to 5, Caddy would "reflect" that. 

Another option is leaving it unset means 0, but setting `keepalive 0` actually sets it to -1 in the code.

mholt: Hmm. I think either way is fine, the more I think about it. If we hard-code a default (like 2) and then _if_ people report problems in the future due to some change, then I guess we can just update Caddy's default. It wouldn't be any harder than recompiling with a new version of Go, since a new compile would have to happen anyway.

I might regret this later but let's change it so that 0 disables keepalives; nginx apparently doesn't enable them by default but Caddy should, so there should be an easy way to turn them off.

nemothekid: Actually, now that *I* think about it, we could just use `net/http.DefaultMaxIdleConnsPerHost` anyways, which would be cleaner.

mholt: Haha, yes, that's good! I didn't see that before...

nemothekid:
If that looks good, I'll update the branch, and then merge #987 as well.

mholt:
LGTM - ship it! We can merge both now.

Thank you again!!

abiosoft:
Good to see your PR @nemothekid :)

PR #764, opened 2016-04-17
Proxy: Add a failing test which replicates #763

edrex:
2 websocket connections are made instead of one

Issue #746, opened 2016-04-12
Caddy Websocket Reverse Proxy


kryptodev:
*If you are filing a bug report, please answer these questions. If your issue is not a bug report, you do not need to use this template. Either way, please consider donating if we've helped you. Thanks!*

#### 1. What version of Caddy are you running (`caddy -version`)?
Untracted Dev build, latest basically.

#### 2. What are you trying to do?
I'm trying to reverse proxy a websocket connection.

#### 3. What is your entire Caddyfile?
```text
my.domain {
	proxy / localhost:8080 {
		proxy_header Host {host}
		proxy_header X-Real-IP {remote}
		proxy_header X-Forwarded-For {remote}
		websocket
	}
}
```

#### 4. How did you run Caddy (give the full command and describe the execution environment)?
Just caddy, the execution environment is Windows Operating System 2008 R2

#### 5. What did you expect to see?
I expected to see a normal websocket connection like I do without Caddy.

#### 6. What did you see instead (give full error messages and/or log)?
read: websocket: close 1006 (abnormal close): unexpected EOF
recv: 313330313023e3933313031

Here's the backend websocket server: https://gist.github.com/kryptodev/2ee48eb9b4c68cf9722d182ec4aba7b3

Please note, the EOF error only occurs with Caddy in between. So far I have tried with TLS off, TLS on, HTTP2 off and also on port 80. Everywhere it's the same scenario. Makes me wonder if this could be windows related? 

Thank you.

mholt:
Thanks for the report. Can you run `./ecaddy -version` after building it with `build.bash`? That will give important information in case we can't solve this right away.

kryptodev:
$ ./ecaddy -version
Caddy 0.8.2 (+b149a86 Wed Apr 13 00:49:24 UTC 2016)

edrex:
I was able to reproduce, and posted a modified version of @kryptodev's gist with results inline https://gist.github.com/edrex/105a4c9577aa13c7c2f52048be3295ce

edrex:
`upgrade:websocket: origin not allowed`

edrex:
It works for me through a proxy if I instantiate the `Upgrader` with a permissive `checkOrigin`:

```
var upgrader = websocket.Upgrader{
	CheckOrigin: func(r *http.Request) bool { return true },
} // use default options
```

@kryptodev, what client are you using to test this? I can't reproduce using the client setup in my version of your gist.

edrex:
[...]

PR #987, opened 2016-08-02
Proxy: Single WebSocket connection

nemothekid:
This PR depends on #984 because they would need to be merged and I didn't want to do that work.

The relevant commits to this PR are c4e65df and d534a21. 

How it essentially works is it created a new transport for Websockets connections (kind of wasteful), but then hijacks the connection on that transport and uses it to continue the websocket connection.

It passes the test created in #764, but I'm not sure if it actually fixes the problems outlined in #746.

### DIFF
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


edrex:
I can confirm that quassel-webserver, which uses socket.io, is working behind Caddy with this change. Nice!

mholt:
Wow, nice work!! The "wasteful" code you refer to is only executed on WebSocket requests, so it would hardly impact performance I think in normal proxy scenarios.

mholt:
This is looking pretty good to me. Thank you so much for solving a challenging problem! Should we go ahead and merge this in @nemothekid?

nemothekid:
Ok, I'll merge it in when we agree on #984 (I have to merge that into this, then update the branch).

Issue #990, opened 2016-08-02
Proxy upstream depending on hostname


meyskens:
It would be nice if Caddy could use a different proxy upstream dependent on the host that is requested.
eg.

```
proxy / http://{host}.internal.somedomain
```

in Nginx we do this by using `server_name ~^(?<subdomain>.+)\.domain\.com$;` and `proxy_pass http://$subdoain.int.domain.com;`

Not sure how hard it is to add this.

mholt:
Thanks for your question, Maartje. Try this:

```
host1.com {
    proxy / http://host1.com.internal.somedomain
}
host2.com {
    proxy / http://host2.com.internal.somedomain
}
```

Much easier I think, no?

meyskens:
Not in our usecase. We use it to proxy dynamic subdomains (client's username) to their specific internal ip for a container where a server runs.

mholt:
It could be possible to enable placeholders for upstream addresses, we'll have to look into that! Anyone is welcome to look into this, not sure I'll have time right now.

vchimishuk:
I can try, if nobody minds.

meyskens:
Was just thinking the same, never worked with the Caddy codebase but i imagine the upstream values just have to be passed through some parser. Will take a look myself later

meyskens:
@vchimishuk didn't see your reply, sorry. Feel free to try!

meyskens:
I've been able to add `outreq.Host = replacer.Replace(outreq.Host)` which should do what I think it has to. But it seems on the setup net/url parses the URL where it returns the error that it doesn't like the `{}`. Not sure how to solve that issue

tpng:
Seems you will have to create a new func like `NewDynamicHostReverseProxy` similar to [`NewSingleHostReverseProxy`](https://github.com/mholt/caddy/blob/master/caddyhttp/proxy/reverseproxy.go#L86) and call it instead at https://github.com/mholt/caddy/blob/master/caddyhttp/proxy/upstream.go#L152 if `uh.Name` contain `{host}`.

meyskens:
Ok I added that and parse the URL in there (with a dirty strings.Replace 😛 to get it parsed). The code runs but still `parse http://{host}:8080: invalid character "{" in host name`. Seems like it still parses it before passing it on. Will keep on looking.

meyskens:
Pushed my work (https://github.com/meyskens/caddy) if anybody spots my mistake you'll get many love.

[...]

PR #998, opened 2016-08-04
Allow {host} in proxy upstream

meyskens:
This PR allow to use {host} in the proxy upstream to you can proxy the request to a specific server/path based on what hostname was requested. (discussed in #990 )

*Note: I have barely any experience in Go, and the implementation I used seems very hacky to me. Any suggestions are welcome*

### DIFF
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -710,7 +710,7 @@ func basicAuthTestcase(t *testing.T, upstreamUser, clientUser *url.Userinfo) {
 }
 
 func newFakeUpstream(name string, insecure bool) *fakeUpstream {
-	if strings.Contains(name,"{") && strings.Contains(name,"}") {
+	if strings.Contains(name, "{") && strings.Contains(name, "}") {
 		ReverseProxy := NewDynamicHostReverseProxy(name, "")
 	} else {
 		uri, _ := url.Parse(name)
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -710,7 +710,7 @@ func basicAuthTestcase(t *testing.T, upstreamUser, clientUser *url.Userinfo) {
 }
 
 func newFakeUpstream(name string, insecure bool) *fakeUpstream {
-	if strings.Contains(name,"{") && strings.Contains(name,"{") {
+	if strings.Contains(name,"{") && strings.Contains(name,"}") {
 		ReverseProxy := NewDynamicHostReverseProxy(name, "")
 	} else {
 		uri, _ := url.Parse(name)
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -710,13 +710,18 @@ func basicAuthTestcase(t *testing.T, upstreamUser, clientUser *url.Userinfo) {
 }
 
 func newFakeUpstream(name string, insecure bool) *fakeUpstream {
-	uri, _ := url.Parse(name)
+	if strings.Contains(name,"{") && strings.Contains(name,"{") {
+		ReverseProxy := NewDynamicHostReverseProxy(name, "")
+	} else {
+		uri, _ := url.Parse(name)
+		ReverseProxy := NewSingleHostReverseProxy(uri, "")
+	}
 	u := &fakeUpstream{
 		name: name,
 		from: "/",
 		host: &UpstreamHost{
 			Name:         name,
-			ReverseProxy: NewSingleHostReverseProxy(uri, ""),
+			ReverseProxy: ReverseProxy,
 		},
 	}
 	if insecure {
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -137,7 +137,7 @@ func NewDynamicHostReverseProxy(templateUrl string, without string) *ReverseProx
 		targetQuery := target.RawQuery
 
 		req.URL.Scheme = target.Scheme
-		req.URL.Host = replacer.Replace(strings.Replace(target.Host, "HOST", "{host}", -1))
+		req.URL.Host = target.Host
 
 		req.URL.Path = singleJoiningSlash(target.Path, req.URL.Path)
 		if targetQuery == "" || req.URL.RawQuery == "" {
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -86,38 +86,8 @@ func socketDial(hostName string) func(network, addr string) (conn net.Conn, err 
 // Without logic: target's path is "/", incoming is "/api/messages",
 // without is "/api", then the target request will be for /messages.
 func NewSingleHostReverseProxy(target *url.URL, without string) *ReverseProxy {
-	targetQuery := target.RawQuery
 	director := func(req *http.Request) {
-		if target.Scheme == "unix" {
-			// to make Dial work with unix URL,
-			// scheme and host have to be faked
-			req.URL.Scheme = "http"
-			req.URL.Host = "socket"
-		} else {
-			req.URL.Scheme = target.Scheme
-			req.URL.Host = target.Host
-		}
-		req.URL.Path = singleJoiningSlash(target.Path, req.URL.Path)
-		if targetQuery == "" || req.URL.RawQuery == "" {
-			req.URL.RawQuery = targetQuery + req.URL.RawQuery
-		} else {
-			req.URL.RawQuery = targetQuery + "&" + req.URL.RawQuery
-		}
-		// Trims the path of the socket from the URL path.
-		// This is done because req.URL passed to your proxied service
-		// will have the full path of the socket file prefixed to it.
-		// Calling /test on a server that proxies requests to
-		// unix:/var/run/www.socket will thus set the requested path
-		// to /var/run/www.socket/test, rendering paths useless.
-		if target.Scheme == "unix" {
-			// See comment on socketDial for the trim
-			socketPrefix := target.String()[len("unix://"):]
-			req.URL.Path = strings.TrimPrefix(req.URL.Path, socketPrefix)
-		}
-		// We are then safe to remove the `without` prefix.
-		if without != "" {
-			req.URL.Path = strings.TrimPrefix(req.URL.Path, without)
-		}
+		req = buildHttpRequest(req, target, without)
 	}
 	rp := &ReverseProxy{Director: director, FlushInterval: 250 * time.Millisecond} // flushing good for streaming & server-sent events
 	if target.Scheme == "unix" {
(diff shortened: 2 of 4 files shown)


mholt:
Cool, thanks for your work @meyskens! Will take a look at it later today.

In the meantime, run `go fmt` on your code which will automatically fix indentation and other formatting issues.

theblazehen:
I've been running this by reverse proxying to nginx behind caddy, would be great to have it implemented directly

caddyhttp/proxy/reverseproxy.go
```diff
@@ -126,6 +127,40 @@ func NewSingleHostReverseProxy(target *url.URL, without string) *ReverseProxy {
 	return rp
 }
 
+func NewDynamicHostReverseProxy(templateUrl string, without string) *ReverseProxy {
+	templateUrl = strings.Replace(strings.ToLower(templateUrl), "{host}", "HOST", -1)
```
mholt: This doesn't generalize well, for example, if you want {hostonly}...

caddyhttp/proxy/reverseproxy.go
```diff
@@ -126,6 +127,40 @@ func NewSingleHostReverseProxy(target *url.URL, without string) *ReverseProxy {
...
+	target, err := url.Parse(templateUrl)
+	if err != nil {
+		return nil
+	}
+	targetQuery := target.RawQuery
+	director := func(req *http.Request) {
+		replacer := httpserver.NewReplacer(req, nil, "")
+
+		req.URL.Scheme = target.Scheme
+		req.URL.Host = replacer.Replace(strings.Replace(target.Host, "HOST", "{host}", -1))
```
mholt: I see why we do this {host}->HOST replacement, but I can also see how that might bite if the input had HOST in it.

mholt:
Nice first attempt! This is not an easy problem to solve elegantly. I have been looking at this for about an hour trying to find a good solution independently as well. The director function is set up when the server is started, not at request-time. We do want to do as much processing as we can before actual requests start coming in in order to reduce latency. But that's what makes it a little hacky, because replacements require a *http.Request.

Will have some suggestions soon.

mholt:
I think the most proper fix will involve making a httpserver.NewReplacer and parsing the upstream URL on each request, since the Director function needs it at request-time. This will introduce some overhead, but I haven't checked how much. Is that something we're willing to accept?

caddyhttp/proxy/reverseproxy.go
meyskens: True. Maybe checking for {} would be better

caddyhttp/proxy/reverseproxy.go
meyskens: That's why the ToLower but that might also cause issues on second thought. url.Parse seems to allow many special characters but {}

meyskens:
Seems like a good idea. Overhead is unavoidable in this situation imo

meyskens:
I think i know a way to do this. I'll try this out and update this PR

meyskens:
This should parse the URL in the director itself on every request, this adds a little overhead on every request but I don't think an url parse will kill your server 😛. It should also allow to use parameters everywhere in the URL (so you can also set the path different if you like). Support for Unix sockets is not included as it seems to me as a security risk.

caddyhttp/proxy/reverseproxy.go
```diff
@@ -126,6 +127,35 @@ func NewSingleHostReverseProxy(target *url.URL, without string) *ReverseProxy {
...
+	director := func(req *http.Request) {
+		replacer := httpserver.NewReplacer(req, nil, "")
+
+		templateUrl = replacer.Replace(templateUrl)
+
+		target, _ := url.Parse(templateUrl)
+		targetQuery := target.RawQuery
+
+		req.URL.Scheme = target.Scheme
+		req.URL.Host = replacer.Replace(strings.Replace(target.Host, "HOST", "{host}", -1))
```
mholt: How about doing the replacement before parsing, then you won't have to hack around the { } like this?

caddyhttp/proxy/reverseproxy.go
```diff
@@ -13,6 +13,7 @@ package proxy
 
 import (
 	"crypto/tls"
+	"github.com/mholt/caddy/caddyhttp/httpserver"
```
mholt: I feel like goimports would move this import line by the other non-std packages... can you double-check that? Run goimports or set up your editor to run it on save. :+1:

caddyhttp/proxy/reverseproxy.go
meyskens: oops, it is already done but forgot to remove the old one
```
templateUrl = replacer.Replace(templateUrl)
target, _ := url.Parse(templateUrl)
```

caddyhttp/proxy/reverseproxy.go
meyskens: ` go fmt` put it there, will look into it!

caddyhttp/proxy/reverseproxy.go
mholt: Oh yeah :smile: ... is it weird that I read PRs from the bottom-up? 😛

meyskens: I do that with magazines so no. (disclaimer: I'm weird)

caddyhttp/proxy/reverseproxy.go
```diff
@@ -126,6 +128,35 @@ func NewSingleHostReverseProxy(target *url.URL, without string) *ReverseProxy {
...
+
+		// We are then safe to remove the `without` prefix.
+		if without != "" {
+			req.URL.Path = strings.TrimPrefix(req.URL.Path, without)
+		}
+	}
+	rp := &ReverseProxy{Director: director, FlushInterval: 250 * time.Millisecond} // flushing good for streaming & server-sent events
+
+	return rp
+}
```
mholt: This is the same thing as NewSingleHostReverseProxy, isn't it, but with a different director and without support for unix sockets?

Any chance we could reuse that code? i.e. call NewSingleHostReverseProxy somehow (maybe its signature has to change slightly?) and then just change the director?

Thanks for working through this. :) Not an easy change! But it's looking better already.

Once we get this looking good, it will need some tests...

meyskens: Calling NewSingleHostReverseProxy seem hard as the URL has to change inside the director. The director is also where 90% of the code changes are. What I was thinking of is putting the property setters of the req inside a separate function.

mholt: That's a possibility too. I'll let you experiment with what is most elegant/clean!

meyskens: oh yay 😛 If suddenly caddy has a bug that it burns down houses: don't blame me 😛

meyskens:
Moved almost all logic of the director in a function that sends back the req. Since there is unix sockets in there and you probably don't want to let `/var/run/{host}.socket` become `/var/run/mysql-without-a-password.socket` I added a check that they always go to SingleHostProxy.

meyskens:
Added the first part for the tests, now only finding out how to use it.

caddyhttp/proxy/proxy_test.go
```diff
@@ -710,13 +710,18 @@ func basicAuthTestcase(t *testing.T, upstreamUser, clientUser *url.Userinfo) {
 }
 
 func newFakeUpstream(name string, insecure bool) *fakeUpstream {
-	uri, _ := url.Parse(name)
+	if strings.Contains(name,"{") && strings.Contains(name,"{") {
```
arikfr: I think the second `strings.Contains` call was meant to be with a `}` instead of `{`.

meyskens:
From testing on my server I noticed the health system isn't useful as are the fail counts. I've disabled them for dynamic hosts. Imo we need a property that tells if a backend is dynamic is the string.Contains gets overused. I also added some more explaining errors (maybe improve the logging, but that is for another PR). 

On tests: anybody a hint what URL it tries to fetch 😛

meyskens:
Also a configurable tryDuration may be handy as not everybody can wait a minute for a 502.

meyskens:
Once I got this tested and wrote the tests it should be ready for review + merge. (may look into a tryDuration property in the caddyfile)

caddyhttp/proxy/upstream.go
```diff
@@ -138,6 +138,9 @@ func (u *staticUpstream) NewHost(host string) (*UpstreamHost, error) {
 				if uh.Unhealthy {
 					return true
 				}
+				if uh.Dynamic {
+					return false
+				}
```
mholt: Backends with hostnames that depend on the request are always up?

caddyhttp/proxy/upstream.go
```diff
@@ -320,13 +329,17 @@ func parseBlock(c *caddyfile.Dispenser, u *staticUpstream) error {
 
 func (u *staticUpstream) healthCheck() {
 	for _, host := range u.Hosts {
-		hostURL := host.Name + u.HealthCheck.Path
-		if r, err := u.HealthCheck.Client.Get(hostURL); err == nil {
-			io.Copy(ioutil.Discard, r.Body)
-			r.Body.Close()
-			host.Unhealthy = r.StatusCode < 200 || r.StatusCode >= 400
+		if host.Dynamic {
+			host.Unhealthy = false //dynamic hosts can never be unhealthy
```
mholt: Hmm, okay, so this is my same question as above; can you explain that to me? 😅

caddyhttp/proxy/upstream.go
meyskens: yeah, since caddy can't see the difference between 2 different backends with this unfortunately.

mholt: Yikes. That is a problem... will think on this...

caddyhttp/proxy/upstream.go
meyskens: since the backend can change with every request checking if it is up is not really possible.

mholt: Okay. I need to see the problem the way you are seeing it. I'll come back to this after I've cleared my head. This might be OK, I just need to convince myself of that.

meyskens: What I think can solve this is either keep a cache of all past requested backends or add an option to shorten the time caddy tries to connect so it doesn't take 1 minute before returning 502 (that is how nginx does it I think)

mholt:
I will be able to get back to this in a little while... this is a bigger change than I thought it would be!

mholt:
I am kind of wondering if this change would merit a re-design of the entire proxy middleware...

meyskens:
Depends on how you want it to work imo. The most efficient for dynamic will indeed be a redesign of the middleware.

mholt:
@meyskens I think without another one or two people proficient with writing proxy code in Go to review this (as I don't trust such a significant change to myself), I just don't feel comfortable with this change. It's disruptive because of the design of the existing proxy code.

I believe a better way to approach this might be to design the proxy middleware with this feature in mind, rather than kind of hacking it into place. I really appreciate your time and efforts on this issue, it made a lot of progress. But I think this is the end of it for now, until proxy is redesigned and built by contributors.

Thank you!! Hope you will still contribute in the future. :smile:

mholt:
This feature has been implemented into Caddy 2, as of 1e31be8.

It was possible in large part due to the experience gained from this pull request, so thank you for that!

PR #1002, opened 2016-08-05
dist/init/linux-sysvinit: improve legacy compatibility

djvdorp:


mholt:
I don't really know what this change means, but thanks for maintaining this! :)

Issue #995, opened 2016-08-04
Custom latency format


wenerme:
#### 1. What version of Caddy are you running (`caddy -version`)?

Custom build from matser a24e36176105d9d42166156c1147e5e0465e2064

#### 2. What are you trying to do?

log latency in custom format instead of `30.021232ms`

#### 3. What is your entire Caddyfile?
```text
:80 {
    log / internal/access.log "{remote} - [{when}] {method} {path} {status} {size} {latency} {host}"
}

```

#### 4. How did you run Caddy (give the full command and describe the execution environment)?

./caddy -conf dev.conf -pidfile internal/pid -log internal/log -http2=false

#### 5. What did you expect to see?

I want some latency format like `33.33` without trailing `ms`, or event more flexible use go time format.

#### 6. What did you see instead (give full error messages and/or log)?

```
111.11.11.11 - [04/Aug/2016:12:28:12 +0800] GET /product/type 200 347 18.869068ms 111.11.11.11
```

mholt:
If we just round this number, is that sufficient? Will anyone mind?

- < 1s: Round to nearest whole number
- >= 1s: Round to nearest thousandth

definitelycarter:
I can take a look at this one.

seedifferently:
@definitelycarter @mholt I stumbled across this while trying to figure out how to use [GoAccess](https://goaccess.io/) with caddy for log analysis, and believe that the original issue here was misunderstood.

Many log analysis tools expect the "latency" param to be just a number (e.g. `123` or `123.45`, but not `123ms`), which is what I believe @wenerme was trying to get at. (cf. the `%L` param here: https://goaccess.io/man#custom-log)

I'd be interested in seeing caddy support this so that the barrier of entry for pairing caddy with common log analysis tools is lowered, and can happily submit a PR if there's agreement that this is a valuable addition (and that the list of placeholders wouldn't become too polluted by the addition of something like a `{latency_ms}` placeholder).

mholt:
@seedifferently Hmm, let's discuss that in a new issue then. Reference this one in the new issue. Thanks!

PR #1005, opened 2016-08-06
Rounding the latency in certain scenarios

definitelycarter:
Rounding the latency based on the suggestions from @mholt.  This addresses #995.


mholt:
This is beautiful. 😭 I think we have a winner here. Thanks @definitelycarter!

Issue #708, opened 2016-04-02
WWW-Authenticate: Basic authentication requires a realm


ntauthority:
I'm using Caddy to provide a service with the 'basicauth' option, and trying to consume it through libavformat's HTTP client. The latter client [expects](http://git.videolan.org/?p=ffmpeg.git;a=blob;f=libavformat/httpauth.c;h=18cf36bcfece25c4f86117f444212c2608fd4bce;hb=HEAD#l95) 'WWW-Authenticate: Basic' be followed by a space, containing a realm - Caddy currently only [sends](https://github.com/mholt/caddy/blob/47c5b6c9c40200cb7483bb890ac3a0e092771021/middleware/basicauth/basicauth.go#L61) the string 'Basic', without any trailing parameter.

As RFC 7617 says:

```
   The Basic authentication scheme utilizes the Authentication Framework
   as follows.

   In challenges:

   o  The scheme name is "Basic".

   o  The authentication parameter 'realm' is REQUIRED ([RFC7235],
      Section 2.2).
```

it seems to be Caddy not adhering to the specification, and libavformat not handling 'broken' (according to the RFC) web servers properly.

mholt:
Right, thanks for pointing this out. Let's get this fixed. How should realm be defined in a way that is both useful and user-friendly (configuration-friendly)? Can we just use the base path argument or should it be something else?

FWIW I think the basicauth middleware could use an upgrade all-around to make it easier to add multiple users or multiple paths per user, and even to load users/passwords from a file (NOT htpasswd) rather than listing it right in the Caddyfile would be cool. But that's another topic for another day.

Tsaukpaetra:
Perhaps a sane default with configurable parameter in Advanced Settings?
(Not a Caddy user, just followed a link)

ntauthority:
> How should realm be defined in a way that is both useful and user-friendly (configuration-friendly)? Can we just use the base path argument or should it be something else?

RFC 7235 states the realm is meant as a sub-namespace of the scheme and authority components (i.e. 'components up to the path'), so I'd assume the base path is a sufficient parameter for this purpose - outside, of course, of vanity/potential-rare-legacy purposes of specifying a custom realm (which could be done in a parameter in to the basicauth option in the configuration).

jupiter:
[...]

PR #1007, opened 2016-08-07
Add default "Restricted" realm to HTTP Basic auth

skorokithakis:
This is hopefully an acceptable workaround for #708, which is currently breaking HTTP Basic auth for many Android apps.


mholt:
Thanks -- I guess "Restricted" is as good as any ("default", etc.) -- and we can put this out in Caddy 0.9.1. And thanks for the test case!

Did you confirm this works for your needs?

skorokithakis:
No, I couldn't build this version 🙁 I commented on the most recent issue, I'm getting some compilation errors.

skorokithakis:
I can confirm that this PR fixes my Android app authentication issue. Eagerly waiting for 0.9.1!

mholt:
Thank you!

PR #1008, opened 2016-08-07
dist/init/linux-sysvinit: process @weingart's feedback

djvdorp:
As he provided in: https://github.com/mholt/caddy/pull/983

I think these changes are correct, but would love a double-check from @weingart before we merge this?


weingart:
looks good to me 👍

mholt:
Thank you both!

PR #1009, opened 2016-08-07
Use P384 for TestUser (privateKey)

elcore:
We are using (see below) to generate a new user

```
func newUser(email string) (User, error) {
	user := User{Email: email}
	privateKey, err := ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
	if err != nil {
		return user, errors.New("error generating private key: " + err.Error())
	}
	user.key = privateKey
	return user, nil
}
```

We should do the same as a test (for the sake of integrity)!

### DIFF
--- a/caddytls/user_test.go
+++ b/caddytls/user_test.go
@@ -2,8 +2,9 @@ package caddytls
 
 import (
 	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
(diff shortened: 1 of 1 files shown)


mholt:
*shrug* I don't think it really matters for this test but you do have a point about integrity or maybe consistency. I originally used 128-bit RSA keys for this test because they're so fast to generate (see crypto_test.go) but I guess ECC keys are too.

PR #1013, opened 2016-08-08
Updated httpserver to include the caddy app details as the product token

slightfoot:
"Updated httpserver to include the caddy app details as the product token."
Kinda says it all. See https://forum.caddyserver.com/t/caddy-and-server-product-tokens/433

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -207,7 +207,7 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}()
 
-	w.Header().Set("Server", "Caddy")
+	w.Header().Set("Server", caddy.AppName+"/"+caddy.AppVersion)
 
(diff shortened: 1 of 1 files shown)


mholt:
Thanks, but we intentionally set it to just "Caddy" since this is the caddy HTTP library (no matter what app is running it) and we hide the version to make it a _little_ less obvious to attackers if the site is running something vulnerable. So I'm going to close this, even though I appreciate the attention to detail. :)

slightfoot:
There are reason to have this included for cluster setups. I understand the attach surface issue. Perhaps to have it default to Product name only with a directive to enable higher level of detail?

mholt:
> Perhaps to have it default to Product name only with a directive to enable higher level of detail?

This is already possible, I think, with the `header` directive. Haven't tried it though. I don't recommend it anyway. See my reply in your forum thread for what I think is a better idea. :wink:

Issue #1015, opened 2016-08-08
Add ability to log request body.


itskenny0:
#### 1. What version of Caddy are you running (`caddy -version`)?
any

#### 2. What are you trying to do?
Log the entire request body.

In #858, the ability to log the request headers was added. However, my use case would require me to log the entire input. Would this be possible using a secondary placeholder, maybe?

definitelycarter:
Do we want to log only certain request bodies either by type, length, both?  I don't think we want to log out binary for example.

itskenny0:
I personally would be interested in POST requests with application/json.

mholt:
- How do we handle newlines (you want a log entry that spans an arbitrary number of lines?)
- How do we handle binary data?
- How do we handle huge request bodies (10+ MB... ?)

itskenny0:
We already have a method to print the headers, which print newlines as \n. That sounds reasonable to me. The implementation by @definitelycarter restricts it to XML and JSON inputs, so binary shouldn't be an issue. A size limitation would be a good idea, however I'm not sure where the limitation should be put.

Maybe this feature would be good in a seperate plugin, which would allow for some configurability.

definitelycarter:
> How do we handle newlines

I saw [this code](https://github.com/mholt/caddy/blob/59b1e8b0bc1f14832a721b9177dbd3884e47d0b4/caddyhttp/httpserver/replacer.go#L20) used in a couple of replacers; However, In my testing I found newlines were not getting logged so I decided to leave it out.

> How do we handle binary data?

I'm currently limiting the logging done based on the Content-Type parameter.  Now that I think of it, It probably won't stop someone from posting binary data with Content-Type of application/json.  Should we use [DetectContentType](https://golang.org/pkg/net/http/#DetectContentType)?

> How do we handle huge request bodies (10+ MB... ?)

Currently I'm copying the entire body.  What would be a good default limit?

slightfoot:
Would this feature just be used for development/debugging? If so then wouldn't a middle-ware host dashboard be better. Like `/debug/requests` and see a live request log with the ability to expand/collapse stuff and all that jazz.

mholt:
@slightfoot That's a totally separate beast, and it'll come, I'm working on it.

itskenny0:
I would definitely require a log though (part of the reason I want this is to be able to replay for debugging). A live view wouldn't suffice as it wouldn't save the requests to disk.

definitelycarter:
[...]

PR #1023, opened 2016-08-10
Now logging the request body

definitelycarter:
Logging the request body if the Content-Type is application/json or
application/xml.  This addresses #1015.

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -127,12 +127,9 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
 					return ""
 				}
 
-				body, err := ioutil.ReadAll(r.Body)
-				// Create a new ReadCloser to keep the body from being drained.
-				r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
-
+				body, err := readRequestBody(r)
 				if err != nil {
-					log.Printf("[WARNING] Cannot read request body %v", err)
+					log.Printf("[WARNING] Cannot copy request body %v", err)
 					return ""
 				}
 
@@ -162,6 +159,18 @@ func canLogRequest(r *http.Request) (canLog bool) {
 		}
 	}
 	return
+}
+
+// readRequestBody reads the request body and sets a
+// new io.ReadCloser that has not yet been read.
+func readRequestBody(r *http.Request) ([]byte, error) {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+	// Create a new ReadCloser to keep the body from being drained.
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	return body, nil
 }
 
 // Replace performs a replacement of values on s and returns
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -1,6 +1,7 @@
 package httpserver
 
 import (
+	"bytes"
 	"io/ioutil"
 	"log"
 	"net"
@@ -127,6 +128,9 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
 				}
 
 				body, err := ioutil.ReadAll(r.Body)
+				// Create a new ReadCloser to keep the body from being drained.
+				r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
 				if err != nil {
 					log.Printf("[WARNING] Cannot read request body %v", err)
 					return ""
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -147,17 +147,16 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
 	return rep
 }
 
-func canLogRequest(r *http.Request) (canLog bool) {
+func canLogRequest(r *http.Request) bool {
 	if r.Method == "POST" || r.Method == "PUT" {
 		for _, cType := range r.Header[headerContentType] {
 			// the cType could have charset and other info
 			if strings.Index(cType, contentTypeJSON) > -1 || strings.Index(cType, contentTypeXML) > -1 {
-				canLog = true
-				break
+				return true
 			}
 		}
 	}
-	return
+	return false
 }
 
 // readRequestBody reads the request body and sets a
@@ -280,5 +279,5 @@ const (
 	headerContentType = "Content-Type"
 	contentTypeJSON   = "application/json"
 	contentTypeXML    = "application/xml"
-	maxLogBodySize    = 100 * 1000
+	maxLogBodySize    = 100 * 1024
 )
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -163,17 +163,21 @@ func canLogRequest(r *http.Request) (canLog bool) {
 // readRequestBody reads the request body and sets a
 // new io.ReadCloser that has not yet been read.
 func readRequestBody(r *http.Request, n int64) ([]byte, error) {
+	defer r.Body.Close()
+
 	body, err := ioutil.ReadAll(io.LimitReader(r.Body, n))
 	if err != nil {
 		return nil, err
 	}
 
-	mr := io.MultiReader(
-		bytes.NewBuffer(body),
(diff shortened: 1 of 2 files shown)


abiosoft:
That is going to drain the request body before it is gets to the relevant middlewares. You will need to find a way to replace the body after draining it.

definitelycarter:
Thank you @abiosoft.  I am now setting the request body to a new ReadCloser.

caddyhttp/httpserver/replacer.go
```diff
@@ -132,6 +148,31 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
+			}
+		}
+	}
+	return
+}
+
+// readRequestBody reads the request body and sets a
+// new io.ReadCloser that has not yet been read.
+func readRequestBody(r *http.Request) ([]byte, error) {
+	body, err := ioutil.ReadAll(r.Body)
```
mholt: This should probably be an io.LimitReader, like 1 MB or even just 100 KB. Not sure what would be ideal yet.

caddyhttp/httpserver/replacer_test.go
```diff
@@ -161,3 +162,25 @@ func TestRound(t *testing.T) {
...
+	}
+	defer r.Body.Close()
+
+	body, err := readRequestBody(r)
+	if err != nil {
+		t.Error("readRequestBody failed", err)
+	}
+
+	var data = make([]byte, len(body))
+	_, err = r.Body.Read(data)
```
mholt: This would be a good place to check to make sure the body was not drained.

caddyhttp/httpserver/replacer.go
```diff
@@ -119,6 +122,19 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
 				return requestReplacer.Replace(string(dump))
 			},
+			"{request_body}": func() string {
+				if !canLogRequest(r) {
+					return ""
+				}
+
+				body, err := readRequestBody(r)
+				if err != nil {
+					log.Printf("[WARNING] Cannot copy request body %v", err)
```
mholt: Although technically an error, I don't think it merits cluttering up the process log in this case -- might as well just return an empty string?

caddyhttp/httpserver/replacer.go
```diff
@@ -119,6 +122,19 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
+					return ""
+				}
+
+				return string(body)
```
mholt: Might be good to condense newlines into \n and \r literals. See `requestReplacer` in this file.

caddyhttp/httpserver/replacer.go
definitelycarter: For now, the limit is set to 100 KB.

caddyhttp/httpserver/replacer_test.go
definitelycarter: I added a second part of the test to compare the Body with the original.

caddyhttp/httpserver/replacer.go
definitelycarter: I removed the log

caddyhttp/httpserver/replacer.go
definitelycarter: In my testing I didn't see new line / carriage returns printed out;  However, I went ahead and added it for consistency sake.

caddyhttp/httpserver/replacer.go
```diff
@@ -132,6 +147,36 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
+			}
+		}
+	}
+	return
+}
+
+// readRequestBody reads the request body and sets a
+// new io.ReadCloser that has not yet been read.
+func readRequestBody(r *http.Request, n int64) ([]byte, error) {
+	body, err := ioutil.ReadAll(io.LimitReader(r.Body, n))
```
mholt: Actually, I totally forgot about [http.MaxBytesReader](https://golang.org/pkg/net/http/#MaxBytesReader) - that might be better for this purpose.

definitelycarter: I'm not sure it is the right fit for this scenario.  Here we just want to log a portion of the body (100 KB).  If the request's body is less then 100 KB then all is well and we'll get the full body size.

Looking at the implementation of the MaxBytesReader, it wants a ResponseWriter to set the status code `Request Too Large` if request's body size is less than the 100 KB limit that we're trying to read.

mholt: Oh, good point. We don't need to penalize the client in this case, just read the first 100 KB at most with no other side-effects. This works as-is.

caddyhttp/httpserver/replacer.go
```diff
@@ -132,6 +147,36 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
+	if err != nil {
+		return nil, err
+	}
+
+	mr := io.MultiReader(
+		bytes.NewBuffer(body),
+		r.Body,
+	)
+
+	r.Body = ioutil.NopCloser(mr)
```
mholt: Will this allow the underlying r.Body to close?

definitelycarter: Fixed.  I decided to read the remaining data from the HttpRequest body and closing it.

caddyhttp/httpserver/replacer.go
```diff
@@ -132,6 +147,40 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
 	return rep
 }
 
+func canLogRequest(r *http.Request) (canLog bool) {
```
mholt: These are stylistic comments, I hope you don't mind :)

I tend to avoid named returns in these situations. I use them only rarely because they make the return statements a little more confusing, since their value lacks context. Better in this situation: leave this return value just `bool`.

caddyhttp/httpserver/replacer.go
```diff
@@ -132,6 +147,40 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
+	if r.Method == "POST" || r.Method == "PUT" {
+		for _, cType := range r.Header[headerContentType] {
+			// the cType could have charset and other info
+			if strings.Index(cType, contentTypeJSON) > -1 || strings.Index(cType, contentTypeXML) > -1 {
+				canLog = true
+				break
```
mholt: `return true` instead

caddyhttp/httpserver/replacer.go
```diff
@@ -132,6 +147,40 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
+			}
+		}
+	}
+	return
```
mholt: `return false`

caddyhttp/httpserver/replacer.go
```diff
@@ -226,6 +275,10 @@ func (r *replacer) Set(key, value string) {
...
 
 const (
-	timeFormat     = "02/Jan/2006:15:04:05 -0700"
-	headerReplacer = "{>"
+	timeFormat        = "02/Jan/2006:15:04:05 -0700"
+	headerReplacer    = "{>"
+	headerContentType = "Content-Type"
+	contentTypeJSON   = "application/json"
+	contentTypeXML    = "application/xml"
+	maxLogBodySize    = 100 * 1000
```
mholt: I think, technically, this should be `100 * 1024`... but I won't let what you have stop the merge. Just seems about 24 short of a nice, round number :wink:

caddyhttp/httpserver/replacer_test.go
```diff
@@ -161,3 +163,28 @@ func TestRound(t *testing.T) {
...
+	r, err := http.NewRequest("POST", "/", bytes.NewReader(payload))
+	if err != nil {
+		t.Error(err)
+	}
+	defer r.Body.Close()
+
+	logBody, err := readRequestBody(r, readSize)
+	if err != nil {
+		t.Error("readRequestBody failed", err)
+	} else if !bytes.EqualFold(payload[0:readSize], logBody) {
```
mholt: EqualFold is an interesting choice here.

mholt:
I can see why having #1027 will be important with this change, since writing the log involves reading the whole body a second time.

Thanks @definitelycarter! This is good.

mholt:
Merge away

PR #1028, opened 2016-08-11
Add support for default (wildcard) error page

evvvvr:
Add ability to specify default (fallback) error page in Caddyfile  via '*' as status code. This addresses #752

PR #1032, opened 2016-08-12
Check for duplicate status code entries in 'errors' directive

evvvvr:
Check for duplicate status code entries in 'errors' directive as  was proposed by @vchimishuk in #1028


mholt:
Nice! Thanks.

PR #1035, opened 2016-08-14
Added support for environment variables to 'templates' module.

slightfoot:
Feature request from [forum post](https://forum.caddyserver.com/t/environment-variables-in-templates/321)
* Environment variables in templates work as:

   `{{.Env.NAME}}` ~~and for debugging you can do `{{.PrintEnv}}` to get them all with new-line breaks.~~

Includes test cases.

### DIFF
--- a/caddyhttp/httpserver/context.go
+++ b/caddyhttp/httpserver/context.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/russross/blackfriday"
+	"os"
 )
 
 // This file contains the context and functions available for
@@ -55,6 +56,24 @@ func (c Context) Cookie(name string) string {
 // Header gets the value of a request header with field name.
 func (c Context) Header(name string) string {
 	return c.Req.Header.Get(name)
+}
+
+// Env gets a map of the environment variables.
+func (c Context) Env() map[string]string {
+	osEnv := os.Environ()
+	envVars := make(map[string]string, len(osEnv))
+	for _, env := range osEnv {
+		data := strings.SplitN(env, "=", 2)
+		if len(data) == 2 {
+			envVars[data[0]] = data[1]
+		}
+	}
+	return envVars
+}
+
+// PrintEnv gets the values of all environment variables (for debug purposes).
+func (c Context) PrintEnv() string {
+	return c.Join(os.Environ(), "\n")
 }
 
 // IP gets the (remote) IP address of the client making the request.
@@ -243,9 +262,14 @@ func (c Context) ToUpper(s string) string {
 	return strings.ToUpper(s)
 }
 
-// Split is a passthrough to strings.Split. It will split the first argument at each instance of the separator and return a slice of strings.
+// Split is a pass-through to strings.Split. It will split the first argument at each instance of the separator and return a slice of strings.
 func (c Context) Split(s string, sep string) []string {
 	return strings.Split(s, sep)
+}
+
+// Join is a pass-through to strings.Join. It will join the first argument slice with the separator in the second argument and return the result.
+func (c Context) Join(a []string, sep string) string {
+	return strings.Join(a, sep)
 }
 
 // Slice will convert the given arguments into a slice.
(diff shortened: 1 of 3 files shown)


caddyhttp/httpserver/context.go
```diff
@@ -57,6 +58,24 @@ func (c Context) Header(name string) string {
 	return c.Req.Header.Get(name)
 }
 
+// Env gets a map of the environment variables.
+func (c Context) Env() map[string]string {
+	os_env := os.Environ()
```
tboerger: Convention for variable names is camel-case

slightfoot: Will update shortly. Noted for next time ;)

slightfoot:
Updated and squashed.

tboerger:
LGTM as far as I can see.

caddyhttp/httpserver/replacer.go
```diff
@@ -141,6 +141,14 @@ func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Repla
...
 	}
 
+	// Environment variable placeholders
+	for _, osEnv := range os.Environ() {
+		env := strings.SplitN(osEnv, "=", 2)
+		if len(env) == 2 {
+			rep.replacements[environReplacer+strings.ToLower(env[0])+"}"] = func() string { return env[1] }
+		}
+	}
+
```
mholt: Environment variables are already evaluated when parsing the Caddyfile. https://caddyserver.com/docs/caddyfile

slightfoot:
Reverted changes to the replacer. I thought their was already that feature, but I couldn't locate it in the docs or code. I would have liked to also have updated the docs, are they in a repo?

caddyhttp/httpserver/context.go
```diff
@@ -57,6 +58,24 @@ func (c Context) Header(name string) string {
 	return c.Req.Header.Get(name)
 }
 
+// Env gets a map of the environment variables.
+func (c Context) Env() map[string]string {
```
mholt: How about instead of getting all of them we just accept an argument which is the env var name and then return its value? Then it's one line: os.GetEnv(). I'd like that much better...

caddyhttp/httpserver/context.go
```diff
@@ -57,6 +58,24 @@ func (c Context) Header(name string) string {
...
+			envVars[data[0]] = data[1]
+		}
+	}
+	return envVars
+}
+
+// PrintEnv gets the values of all environment variables (for debug purposes).
+func (c Context) PrintEnv() string {
+	return c.Join(os.Environ(), "\n")
+}
```
mholt: I don't think this is needed.

caddyhttp/httpserver/replacer_test.go
```diff
@@ -121,7 +121,7 @@ func TestSet(t *testing.T) {
 
 	request, err := http.NewRequest("POST", "http://localhost", reader)
 	if err != nil {
-		t.Fatalf("Request Formation Failed \n")
+		t.Fatalf("Request Formation Failed: %s\n", err.Error())
```
mholt: Oh, thanks :blush:

caddyhttp/httpserver/context.go
slightfoot: I thought the same thing, but other go templating uses the map notion. So keeping it the same might be good for portability and the 'barer to entry', so to speak. Perhaps caching the map up front on Context creation to so avoid processing on each call?

caddyhttp/httpserver/context.go
slightfoot: This is most definitely needed. When your wondering why your Environment Var is not coming up.. its nice to be able to dump them all and check to see if everything is as it should be.

caddyhttp/httpserver/context.go
mholt: Let's just leave it as-is for now, we can optimize later. This function is good for now.

caddyhttp/httpserver/context.go
mholt: But you can iterate a map in templates with just a couple lines using `.range` - which, IMO, is better because you can control line breaks (if you want `<br>` for example). I still vote we remove this method.

mholt:
Great, thanks @slightfoot! :+1:

PR #1036, opened 2016-08-15
Fix error which lead to skipping tests in 'errors.TestErrorsParse'

evvvvr:
Fix mistake in checking error conditions


mholt:
Oops, thanks!

PR #1037, opened 2016-08-15
Fix for invalid environment variable names used on windows that start…

slightfoot:
… with an equals symbol. Even though this contradicts the Microsoft docs.


mholt:
Cool, thank you! (Still think it's weird that Windows does that.)

Issue #1034, opened 2016-08-14
Support for lego's timeout option


dangmai:
Lego added an option to specify timeout for reaching the ACME server: https://github.com/xenolf/lego/issues/207 Is there any chance that we can change this option from Caddy?

Why it'd be helpful: I run Caddy in containers on IBM Bluemix, and for some reason it takes a long time for Lego to contact the ACME server from the containers (usually more than the default 10 second time out), so the Caddy would sometimes be unable to activate the HTTPS feature.

Thanks!

mholt:
The HTTP timeout is set by changing the whole ACME package's `http.Client` value, which means we could make this a command line flag at best - it won't be customizable on a per-site basis, which I guess is fine.

dangmai:
Yes that'd be enough for my use case. Thanks!

PR #1041, opened 2016-08-16
Support for lego's timeout option

tehscorpion:
I hope i've done nothing wrong but at least it compiles and runs :8ball: 
#1034

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -34,6 +34,7 @@ func init() {
 	flag.StringVar(&cpu, "cpu", "100%", "CPU cap")
 	flag.BoolVar(&plugins, "plugins", false, "List installed plugins")
 	flag.StringVar(&caddytls.DefaultEmail, "email", "", "Default ACME CA account email address")
+	flag.IntVar(&caddytls.HTTPTimeout, "catimeout", 10, "ACME CA HTTP timeout in seconds")
 	flag.StringVar(&logfile, "log", "", "Process log file")
 	flag.StringVar(&caddy.PidFile, "pidfile", "", "Path to write pid file")
 	flag.BoolVar(&caddy.Quiet, "quiet", false, "Quiet mode (no initialization output)")
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -49,7 +49,6 @@ func init() {
 // Run is Caddy's main() function.
 func Run() {
 	flag.Parse()
-	moveStorage() // TODO: This is temporary for the 0.9 release, or until most users upgrade to 0.9+
 
 	caddy.AppName = appName
 	caddy.AppVersion = appVersion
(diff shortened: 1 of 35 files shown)


caddytls/client.go
```diff
@@ -60,6 +61,11 @@ var newACMEClient = func(config *Config, allowPrompts bool) (*ACMEClient, error)
 		return nil, fmt.Errorf("%s: insecure CA URL (HTTPS required)", caURL)
 	}
 
+	// Set timeout by replacing ACME's http.Client
+	if HTTPTimeout > 1 {
+		acme.HTTPClient = http.Client{Timeout: time.Duration(HTTPTimeout) * time.Second}
```
mholt: This should probably be assigned only once, not ever time we create a new ACME client. @xenolf can you confirm when you have a chance?

caddy/caddymain/run.go
```diff
@@ -34,6 +34,7 @@ func init() {
 	flag.StringVar(&cpu, "cpu", "100%", "CPU cap")
 	flag.BoolVar(&plugins, "plugins", false, "List installed plugins")
 	flag.StringVar(&caddytls.DefaultEmail, "email", "", "Default ACME CA account email address")
+	flag.IntVar(&caddytls.HTTPTimeout, "catimeout", 10, "ACME CA HTTP timeout in seconds")
```
mholt: This should be a DurationVar. Then you don't have to convert it when we actually use it.

caddytls/tls.go
```diff
@@ -157,6 +157,9 @@ var (
 	// DefaultEmail represents the Let's Encrypt account email to use if none provided.
 	DefaultEmail string
 
+	// Sets the maximum waiting time for the CA's HTTP responses
+	HTTPTimeout int
```
mholt: As above, change to time.Duration.

caddytls/client.go
xenolf: It's a package level variable and not connected to a single client instance. Assigning it once is enough.

mholt: Okay, thanks. @tehscorpion can you make this so that it's only assigned once?

tehscorpion:
I changed it according to your request, but i have actually no idea where to put it...at least where you want it.
@mholt could you refine it according to your preference?

mholt:
Thanks @tehscorpion! I'll comment inline.

caddytls/client.go
```diff
@@ -60,6 +61,9 @@ var newACMEClient = func(config *Config, allowPrompts bool) (*ACMEClient, error)
 		return nil, fmt.Errorf("%s: insecure CA URL (HTTPS required)", caURL)
 	}
 
+	// Set timeout by replacing ACME's http.Client
+	acme.HTTPClient = http.Client{Timeout: HTTPTimeout}
```
mholt: This creates a new http client every time we perform an ACME request. Actually, what if we just changed the timeout value? `acme.HTTPClient.Timeout = HTTPTimeout` - then we don't have to replace the whole client each time.

mholt: Better yet, just remove this line and change the flag line to be:

```
flag.DurationVar(&acme.HTTPClient.Timeout, "catimeout", acme.HTTPClient.Timeout, "Default ACME CA HTTP timeout")
```

mholt:
Thank you!

Issue #1012, opened 2016-08-07
rewrite to status 404 does not show custom 404 error page


peterychuang:
I am using rewrite-to-404 as a way of denying access to a certain files in my site root. While I expect it to show the custom 404 page I defined in the errors block, it instead shows me a plain text "404 Not Found".

Below is my sample Caddyfile:

```
localhost:5000 {
  root ./public
  gzip

  errors {
    404 404.html
  }

  rewrite /something {
    status 404
  }
}
```

Not a big deal, but it would be nice if it behaves like other 404 error pages.

abiosoft:
This is because `errors` is after `rewrite` in the middleware chain. 

@mholt don't you think errors should be higher up in the chain considering it also has the panic/recover handler.

mholt:
This is kind of a regression after fixing #914. We can't move errors to be chained in before rewrite,  otherwise error pages won't be compressed because gzip is chained in later. @abiosoft and I were chatting in Slack about one possible solution. 

We could separate the status code rewriting out of rewrite directive. The rewrite directive would then only rewrite URI, not status code. The new middleware, say, `status` could then be chained in after `errors`, allowing the rewritten status code to be served with custom error pages again.

I don't know what Abiola thinks of this yet, but I think I like it. :smile: I don't mind that it breaks the some rewrite directives at this stage (we're pre-1.0), I'm mostly concerned about usability. Will it be confusing? We could clarify the docs to say that rewrite is for rewriting the request URI internally.

peterychuang:
I would say that having a separate `status` is actually less confusing than using `rewrite` for the purpose of setting status code. It could be just me though.

mholt:
The middleware proposed in #370 is very similar to this `status` directive we're considering, in that it just iterates rules (even if the 'rule' is just a request path prefix) and writes a status code (and nothing else, I guess) if it finds a matching rule.

Looking at it differently, `ext` is just a common kind of `rewrite` that was made for convenience, because it's much easier to use `ext` than it is to write the `rewrite` rules yourself.

So perhaps we could also look at `forbid` and `hide` as potentially common cases of `status`. In the future. Maybe not right now. I think `status` would be a good start, and should be a middleware that's built into Caddy.

evvvvr:
Would like to take it. Am I right  that `status` directive should have the same syntax/support the same rules as `rewrite` (except of course `status` and `to` entries) i.e. it should look like:
`status _path_ _status code_`
or
`status [basepath] {
[...]

PR #1042, opened 2016-08-16
Add 'status' middleware instead of 'status' directive for 'rewrite' middleware

evvvvr:
Simple implementation of middleware for responding with specified status as discussed in #1012

### DIFF
--- a/caddyhttp/status/setup.go
+++ b/caddyhttp/status/setup.go
@@ -1,0 +1,93 @@
+package status
+
+import (
+	"strconv"
+
+	"github.com/mholt/caddy"
+	"github.com/mholt/caddy/caddyhttp/httpserver"
+)
+
+// init registers Status plugin
+func init() {
+	caddy.RegisterPlugin("status", caddy.Plugin{
+		ServerType: "http",
+		Action:     setup,
+	})
+}
+
+// setup configures new Status middleware instance.
+func setup(c *caddy.Controller) error {
+	rules, err := statusParse(c)
+	if err != nil {
+		return err
+	}
+
+	cfg := httpserver.GetConfig(c)
+	mid := func(next httpserver.Handler) httpserver.Handler {
+		return Status{Rules: rules, Next: next}
+	}
+	cfg.AddMiddleware(mid)
+
+	return nil
+}
+
+// statusParse parses status directive
+func statusParse(c *caddy.Controller) ([]httpserver.HandlerConfig, error) {
+	var rules []httpserver.HandlerConfig
+
+	for c.Next() {
+		hadBlock := false
+		args := c.RemainingArgs()
+
+		switch len(args) {
+		case 1:
+			status, err := strconv.Atoi(args[0])
+			if err != nil {
+				return rules, c.Errf("Expecting a numeric status code, got '%s'", args[0])
+			}
+
+			for c.NextBlock() {
+				hadBlock = true
+				basePath := c.Val()
+
+				for _, cfg := range rules {
+					rule := cfg.(*Rule)
+					if rule.Base == basePath {
+						return rules, c.Errf("Duplicate path: '%s'", basePath)
+					}
+				}
+
+				rule := NewRule(basePath, status)
+				rules = append(rules, rule)
+
+				if c.NextArg() {
+					return rules, c.ArgErr()
+				}
+			}
+
+			if !hadBlock {
+				return rules, c.ArgErr()
+			}
+		case 2:
+			status, err := strconv.Atoi(args[0])
+			if err != nil {
+				return rules, c.Errf("Expecting a numeric status code, got '%s'", args[0])
+			}
+
+			basePath := args[1]
+			for _, cfg := range rules {
+				rule := cfg.(*Rule)
+				if rule.Base == basePath {
+					return rules, c.Errf("Duplicate path: '%s'", basePath)
+				}
+			}
+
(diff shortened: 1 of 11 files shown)


mholt:
Cool - thanks @evvvvr! Going to be looking through this soon.

mholt:
This is looking nice so far, thank you @evvvvr. I would like @abiosoft to review it when he has a chance (I know he's busy).

caddyhttp/status/setup.go
```diff
@@ -0,0 +1,88 @@
...
+		if _, exists := res[path]; exists {
+			return c.Errf("Duplicate path: '%s'", path)
+		}
+		res[path] = statusCode
+
+		return nil
+	}
+
+	for c.Next() {
+		hadBlock = false
```
mholt: Not sure we need block support do we? `status path code` is a pretty simple syntax. Do you think there will be lists of many items?

mholt: Okay, I just saw the tests below. I'm not sure the blocks are really worth the extra syntax complexity. _Maybe_ if they all shared the same status code... what do you think?

evvvvr: I made it by analogy with `errors` directive, where we can specify pages for different status codes in one block:
```
errors {
   what where
   what where... 
}
```  
IMO it is convenient to be able  to specify rules in one block, instead of writing the same directive multiple times like:
```
status path code
status path code...
```

abiosoft: Both can be supported.

mholt: Okay, so if you do allow blocks, what if it was for the same status code:

```
status 404 {
    /path1
    /path2
}
```

I figure there won't be so many different statuses that you'd have to write the word "status" a lot.

caddyhttp/status/setup.go
```diff
@@ -0,0 +1,88 @@
...
+
+// statusParse parses status directive
+func statusParse(c *caddy.Controller) (map[string]int, error) {
+	res := make(map[string]int)
+	hadBlock := false
+
+	saveStatusRule := func(path, statusCodeString string) error {
+		statusCode, err := strconv.Atoi(statusCodeString)
+		if err != nil {
+			return c.Err(fmt.Sprintf("Expecting a numeric status code, got '%s'", statusCodeString))
```
abiosoft: why not just `c.Errf` ?

caddyhttp/status/status.go
```diff
@@ -0,0 +1,23 @@
...
+// Status is a middleware to return status code for request
+type Status struct {
+	Rules map[string]int
+	Next  httpserver.Handler
+}
+
+// ServeHTTP implements the httpserver.Handler interface
+func (status Status) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
+	if statusCode, ok := status.Rules[r.URL.Path]; ok {
+		return statusCode, nil
```
mholt: Hmm, this middleware is an odd exception in the way I've described how the [contract](https://godoc.org/github.com/mholt/caddy/caddyhttp/httpserver#Handler) for a Handler works:

> If ServeHTTP writes to the response body, it should return a status code of 0. This signals to other handlers above it that the response body is already written, and that they should not write to it also.

In this case we don't write a response body but we do essentially "finish" handling the request (which is what I meant by "writes to the response body" - perhaps I should clarify that in the godoc by changing it to "writes the response header").

So what happens here is two-fold:

- If we return statusCode >= 400, great; an error page can be served by the errors middleware.
- If we return statusCode < 400... not so great, because there's no code to write a response body in that case, or even write the status code to the header of the response.

In other words, we need a condition, something like:

```go
if statusCode < 400 {
    w.WriteHeader(statusCode)
    return 0, nil
}
return statusCode, nil
```

mholt:
@evvvvr The latest changes are looking great! Just one more inline comment that I just made and I think it's getting really close. :)

evvvvr:
Changed directive syntax to `status code path` and
```
status code {
   path
   path...
} 
``` 
Is it ok?
Please, be advised that code in PR supports **exact** path only. I propose
```
status code {
  basepath base_path_value
}
```
syntax for it.

mholt:
I just noticed the exact path matching. I think we should choose either base or exact, but probably not both (confusing). I'm leaning towards base path matching since it will allow you to hide a whole folder, and if you just need to hide a file, you can list the full file path.

evvvvr:
Replaced exact path matching with base path check and used [`httpserver.HandlerConfig`](https://godoc.org/github.com/mholt/caddy/caddyhttp/httpserver#HandlerConfig) for middleware configuration.

caddyhttp/status/status.go
```diff
@@ -0,0 +1,64 @@
...
+		RequestMatcher: httpserver.PathMatcher(basePath),
+	}
+}
+
+// BasePath implements httpserver.HandlerConfig interface
+func (rule *Rule) BasePath() string {
+	return rule.Base
+}
+
+// Match implements httpserver.HandlerConfig interface
```
abiosoft: This method is redundant since `Rule` embeds `httpserver.RequestMatcher`.

abiosoft:
Finally, looks like this branch is behind master and a rebase is needed.

abiosoft:
Thanks :)

mholt:
Thank you @evvvvr! I'll write up some docs for this.

PR #1048, opened 2016-08-19
Implement missing bits for an external servertype

miekg:
Make ServerListeners public and add two helper methods to get access
to the address they listen on. This is useful for tests (among other
things)

Also make DefaultConfigFile a var so it can be overridden by server
types.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -76,7 +76,7 @@ type Instance struct {
 	context Context
 
 	// servers is the list of servers with their listeners.
-	servers []serverListener
+	servers []ServerListener
 
 	// these callbacks execute when certain events occur
 	onFirstStartup  []func() error // starting, not as part of a restart
@@ -85,6 +85,9 @@ type Instance struct {
 	onShutdown      []func() error // stopping, even as part of a restart
 	onFinalShutdown []func() error // stopping, not as part of a restart
 }
+
+// Servers returns the ServerListeners in i.
+func (i *Instance) Servers() []ServerListener { return i.servers }
 
 // Stop stops all servers contained in i. It does NOT
 // execute shutdown callbacks.
@@ -202,7 +205,7 @@ func (i *Instance) Restart(newCaddyfile Input) (*Instance, error) {
 // internally-kept list of servers that is running. For
 // saved servers, graceful restarts will be provided.
 func (i *Instance) SaveServer(s Server, ln net.Listener) {
-	i.servers = append(i.servers, serverListener{server: s, listener: ln})
+	i.servers = append(i.servers, ServerListener{server: s, listener: ln})
 }
 
 // HasListenerWithAddress returns whether this package is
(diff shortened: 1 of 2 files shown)


miekg:
How do you sign the CLA?

mholt:
This is weird, there should be CI checks too from both Travis and AppVeyor. The CLA is powered by CLA Assistant, but I don't know why it hasn't showed up. Anyway, you've already signed it.

caddy.go
```diff
@@ -836,7 +839,7 @@ var (
 	instancesMu sync.Mutex
 )
 
-const (
+var (
```
mholt: Remember that Caddy can (soon) be compiled with multiple server types built in - the HTTP server will look for Caddyfile by default but looking for Corefile in that case just because it's built with the DNS server plugged in would be confusing. Instead of changing this, have your server type specify default Caddyfile contents.

miekg: [ Quoting <notifications@github.com> in "Re: [mholt/caddy] Implement missing..." ]
>> @@ -836,7 +839,7 @@ var (
>>  	instancesMu sync.Mutex
>>  )
>>
>> -const (
>> +var (
>
>Remember that Caddy can (soon) be compiled with multiple server types built in
>- the HTTP server will look for Caddyfile by default but looking for Corefile
>in that case just because it's built with the DNS server plugged in would be
>confusing. 

Ok, but this is a regression for me. How can we/I make it so that if -type=dns
is will look for a Corefile instead of a Caddyfile?

>Instead of changing this, have your server type specify default Caddyfile contents.

I don't have a sane default for an empty Caddyfile. I prefer to not do anything
at that point, i.e. don't start at all.

/Miek

--
Miek Gieben

miekg: [ Quoting <notifications@github.com> in "Re: [mholt/caddy] Implement missing..." ]
>> @@ -836,7 +839,7 @@ var (
>>  	instancesMu sync.Mutex
>>  )
>>
>> -const (
>> +var (
>
>Remember that Caddy can (soon) be compiled with multiple server types built in - the HTTP server will look for Caddyfile by default but looking for Corefile in that case just because it's built with the DNS server plugged in would be confusing. Instead of changing this, have your server type specify default Caddyfile contents.

Even so, the default is *still* Caddyfile even with CoreDNS compiled in? I just
change it in *my* startup....?

mholt: > Ok, but this is a regression for me. How can we/I make it so that if -type=dns
is will look for a Corefile instead of a Caddyfile?

Set the DefaultInput field of your server type to a function that loads a file called Corefile, perhaps? But if there is a Caddyfile in the directory already, then it will use that anyway... :-/

mholt: > Even so, the default is *still* Caddyfile even with CoreDNS compiled in? I just
change it in *my* startup....?

If you change it only when the DNS server is being _used_, I guess that's okay. How are you doing that?

miekg: Add it here: https://github.com/miekg/coredns/blob/caddy/coredns.go ?

mholt: Ohh, from your main function, not your plugin, I see... that should work. Okay.

miekg: [ Quoting <notifications@github.com> in "Re: [mholt/caddy] Implement missing..." ]
>> @@ -836,7 +839,7 @@ var (
>>  	instancesMu sync.Mutex
>>  )
>>
>> -const (
>> +var (
>
>Ohh, from your main function, not your plugin, I see... that should work. Okay.

Yep! :-) (sorry if that wasn't clear)

PR #1049, opened 2016-08-19
capture request body normally

tw4452852:


mholt:
Hey, thanks Tw! Does this imply a performance hit, though, since we always create a 100KB buffer for requests with a body? (And I need to check, is request.Body nil if there is no request body?)

tw4452852:
@mholt 

> Does this imply a performance hit, though, since we always create a 100KB buffer for requests with a body?

No. It only allocates a default `bytes.Buffer` for the request, and the buffer only expands when others read the request's body.

> And I need to check, is request.Body nil if there is no request body?

I think so.

mholt:
Nice, thank you Tw!

There is just one lint warning, though:

> caddyhttp\httpserver\replacer.go:304:2: comment on exported const MaxLogBodySize should be of the form "MaxLogBodySize ..."

Otherwise I think this is okay to merge. This is a big change but I can see why it's necessary and, in some ways, better. Will be interesting later on to evaluate performance of the replacer and how it might be able to be optimized, if it is a bottleneck.

mholt:
Brilliant! Thanks @tw4452852 !

PR #1058, opened 2016-08-21
caddymain: Fix setting and displaying versioning

miekg:
Move setting the versioning into the init function and use
caddy.AppVersion and friends when displaying it.

This is useful for servertype plugin that re-use `run.go`, but want to
set their own versioning and naming.

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -27,6 +27,7 @@ import (
 func init() {
 	caddy.TrapSignals()
 	setVersion()
+	setName()
 
 	flag.BoolVar(&caddytls.Agreed, "agree", false, "Agree to the CA's Subscriber Agreement")
 	flag.StringVar(&caddytls.DefaultCAUrl, "ca", "https://acme-v01.api.letsencrypt.org/directory", "URL to certificate authority's ACME server directory")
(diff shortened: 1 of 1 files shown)


mholt:
FYI I'm holding out on this until we figure out how to differentiate branding from the application. Discussion seems to be happening in the forums.

miekg:
Ack. Where is the discussion exactly? I'm sorta waiting for something like
this to be merged, but I can see why you would hesitate.

mholt:
There's one here: https://forum.caddyserver.com/t/caddy-defaultconfigfile/506

And another here: https://forum.caddyserver.com/t/server-types-other-than-http/65/54

It would probably be good to move the discussion over to the first link.

miekg:
Fair enough. I think this calls for you to make a decision, and provide a
way to do this, or at least tell what we should do 😁

mholt:
I know. I want to preserve your brand since the server type is the dominant function of the application, but that would make it exclusive of others. So the question is should we just take the greatest common denominator of the server types (which is the `caddy` package)... maybe we need to separate branding from technical function better.

Hmm, deep questions... :)

miekg:
Hmmm, do you for see multiple types to be enabled at the same time? That is
a usecase I have not foreseen... Would it make sense to have both dns and
http? I now disable http when you compile coredns...

mholt:
> Hmmm, do you for see multiple types to be enabled at the same time?

Probably not 'enabled' (as in, being used in the same process) but definitely could have more than one plugged in.

miekg:
I'll copy (and edit out most of it) run.go and use that for the time being.

PR #1063, opened 2016-08-23
Make ACME listener ports configurable

### LINKED ISSUE #918: Add directive to change plain HTTP port
Would be nice if we could have a directive to let caddy listen on a >1024 port for the plain HTTP requests.
Currently the vhost:port notation is only for HTTPS traffic, but I want the HTTP listener to listen on >1024 too to circumvent the bind permission denied problem.

Something like `http_port 8080` would do fine.

Ideas?

tdsmith:
Adds new command-line options that describe which ports are externally
    visible as ports 80 and 443 and instructs the ACME listener to use them
    instead of the default ports. This allows non-DNS ACME challenges to work
    without special privileges or privilege delegation mechanisms.

This is my first PR to Caddy and my first time playing with Go. :) I've probably made some poor architectural decisions in the interest of producing a simple, concrete artifact for feedback. I tested it on a AWS instance with [iptables port forwarding](https://wiki.jenkins-ci.org/display/JENKINS/Running+Jenkins+on+Port+80+or+443+using+iptables) and it looks like it behaves correctly. I haven't attempted unit tests.

I hope this isn't too frustrating to review!

Fixes #918.

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -40,6 +40,8 @@ func init() {
 	flag.StringVar(&revoke, "revoke", "", "Hostname for which to revoke the certificate")
 	flag.StringVar(&serverType, "type", "http", "Type of server to run")
 	flag.BoolVar(&version, "version", false, "Show version")
+	flag.StringVar(&caddytls.HTTPChallengePort, "external-port-80-is-local-port", "80", "Local port exposed externally as port 80")
+	flag.StringVar(&caddytls.TLSSNIChallengePort, "external-port-443-is-local-port", "443", "Local port exposed externally as port 443")
 
 	caddy.RegisterCaddyfileLoader("flag", caddy.LoaderFunc(confLoader))
 	caddy.SetDefaultCaddyfileLoader("default", caddy.LoaderFunc(defaultLoader))
--- a/caddyhttp/httpserver/https.go
+++ b/caddyhttp/httpserver/https.go
@@ -108,6 +108,10 @@ func enableAutoHTTPS(configs []*SiteConfig, loadCertificates bool) error {
 func makePlaintextRedirects(allConfigs []*SiteConfig) []*SiteConfig {
 	for i, cfg := range allConfigs {
 		if cfg.TLS.Managed &&
+			// don't trigger automatic redirects if port 80 and 443 might not be
+			// bindable
+			caddytls.HTTPChallengePort == "80" &&
+			caddytls.TLSSNIChallengePort == "443" &&
 			!hostHasOtherPort(allConfigs, i, "80") &&
 			(cfg.Addr.Port == "443" || !hostHasOtherPort(allConfigs, i, "443")) {
 			allConfigs = append(allConfigs, redirPlaintextHost(cfg))
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -528,15 +528,15 @@ var defaultCiphers = []uint16{
 	tls.TLS_RSA_WITH_AES_128_CBC_SHA,
 }
 
-const (
-	// HTTPChallengePort is the officially designated port for
-	// the HTTP challenge.
+var (
+	// HTTPChallengePort must be externally visible as port 80.
 	HTTPChallengePort = "80"
 
-	// TLSSNIChallengePort is the officially designated port for
-	// the TLS-SNI challenge.
+	// TLSSNIChallengePort must be externally visible as port 443.
 	TLSSNIChallengePort = "443"
+)
 
+const (
 	// DefaultHTTPAlternatePort is the port on which the ACME
 	// client will open a listener and solve the HTTP challenge.
 	// If this alternate port is used instead of the default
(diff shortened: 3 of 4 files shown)


caddy/caddymain/run.go
```diff
@@ -40,6 +40,8 @@ func init() {
 	flag.StringVar(&revoke, "revoke", "", "Hostname for which to revoke the certificate")
 	flag.StringVar(&serverType, "type", "http", "Type of server to run")
 	flag.BoolVar(&version, "version", false, "Show version")
+	flag.StringVar(&caddytls.HTTPChallengePort, "external-port-80-is-local-port", "80", "Local port exposed externally as port 80")
+	flag.StringVar(&caddytls.TLSSNIChallengePort, "external-port-443-is-local-port", "443", "Local port exposed externally as port 443")
```
mholt: Woah, these flag names gotta get shorter.

mholt:
Thanks for the initiative, @tdsmith - especially a first one with Go! I'm gonna give this one some serious thought since this is kind of a scary change. Not frustrating :smile: Just scary. :wink:

caddy/caddymain/run.go
tdsmith: The suggestion is a little tongue-in-cheek but I do like that there's not much mystery about valid configurations. :) Maybe `--external-port-80` and `--external-port-443`?

caddyhttp/httpserver/https.go
```diff
@@ -108,6 +108,10 @@ func enableAutoHTTPS(configs []*SiteConfig, loadCertificates bool) error {
 func makePlaintextRedirects(allConfigs []*SiteConfig) []*SiteConfig {
 	for i, cfg := range allConfigs {
 		if cfg.TLS.Managed &&
+			// don't trigger automatic redirects if port 80 and 443 might not be
+			// bindable
+			caddytls.HTTPChallengePort == "80" &&
+			caddytls.TLSSNIChallengePort == "443" &&
```
tdsmith: This feels bad; should the caddytls variables be renamed? Does it make sense to write a helper function called e.g. `defaultPortsExposed`?

caddytls/config.go
```diff
@@ -528,15 +528,15 @@ var defaultCiphers = []uint16{
...
-	// the HTTP challenge.
+var (
+	// HTTPChallengePort must be externally visible as port 80.
 	HTTPChallengePort = "80"
 
-	// TLSSNIChallengePort is the officially designated port for
-	// the TLS-SNI challenge.
+	// TLSSNIChallengePort must be externally visible as port 443.
 	TLSSNIChallengePort = "443"
+)
```
mholt: I think changing these into vars is not a good idea, these are consts because they are what the ACME spec requires.

caddyhttp/httpserver/https.go
```diff
@@ -108,6 +108,10 @@ func enableAutoHTTPS(configs []*SiteConfig, loadCertificates bool) error {
...
 			!hostHasOtherPort(allConfigs, i, "80") &&
 			(cfg.Addr.Port == "443" || !hostHasOtherPort(allConfigs, i, "443")) {
```
mholt: Here you're checking to see if redirects can/need to be set up, but you're checking ports 80 and 443. Issue #918 isn't about changing the ACME ports, it's about changing the plaintext HTTP port, which affects the redirects as well.

caddy/caddymain/run.go
mholt: But this is changing only the ACME challenge ports, which are hard-coded into the spec. I don't think these should change.

caddyhttp/httpserver/https.go
tdsmith: Sure, that makes sense.

caddy/caddymain/run.go
tdsmith: Can you clarify what you mean by "I don't think these should change"? Do you mean that Caddy should never try to open local ports other than 80 or 443 to respond to ACME challenges? I'm specifically interested in responding to ACME challenges on alternative ports because we have a NAT router forwarding i.e. external_ip:80 to internal_ip:8080. Or do you mean that these values should be stored in a different set of variables?

mholt: By "I don't think these should change," I mean that those are consts which represent the ports defined by the ACME spec. Flags don't change the spec and I don't think we should be modifying those consts either.

> Do you mean that Caddy should never try to open local ports other than 80 or 443 to respond to ACME challenges?

No. It already opens an alternate port for the HTTP challenge when the server is already running. But you're asking to change the _default_ port. But that is hard-coded into the spec.

> I'm specifically interested in responding to ACME challenges on alternative ports because we have a NAT router forwarding i.e. external_ip:80 to internal_ip:8080.

Only ACME HTTP challenges or all HTTP traffic? I guess I'm not sure clear on what problem you're trying to solve, between this PR and the linked issue which talks about the HTTP port for everything.

mholt:
Hey @tdsmith, thanks for contributing this. I'm going to close this for now since it's not clear yet what the change is trying to do, exactly -- but feel free to continue working on this, clarify, etc, reopen when ready. Cheers!

caddy/caddymain/run.go
tdsmith: Solely handling the ACME challenge is sufficient to scratch my itch but I see that the linked issue wants to handle all "default" uses of ports 80 and 443.

PR #1064, opened 2016-08-23
Keep quic protocol headers only between one hop

stp-ip:
Removing quic protocol headers from being persisted during proxy requests.
Not removing them could lead to the client attempting to connect to the wrong port.
This makes the quic headers consistent with other protocol headers.


mholt:
@stp-ip And you tested this that it works? (I admit we need some more tests here eventually.)

stp-ip:
Yeah that is tested and works.
Should the Server header also be added there?

mholt:
I think that's debatable (about the Server header). Might be nice to know what the request traveled through, or not. Either way, proxy has header controls to allow that to be customized. Thanks for this PR!

PR #1070, opened 2016-08-25
Actually set tls.Config.PreferServerCipherSuites

FiloSottile:
It was set by default on the caddy-internal config object, and even
checked for conflicts, but it was never actually reflected on the
tls.Config.

This will have user-visible changes: a client that prefers, say, AES-CBC
but also supports AES-GCM would have used AES-CBC befor this, and will
use AES-GCM after.

This is desirable and important behavior, because if for example the
server wanted to support 3DES, but *only if it was strictly necessary*,
it would have had no way of doing so with PreferServerCipherSuites
false, as the client preference would have won.

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -350,6 +350,7 @@ func MakeTLSConfig(configs []*Config) (*tls.Config, error) {
 		if i > 0 && cfg.PreferServerCipherSuites != configs[i-1].PreferServerCipherSuites {
 			return nil, fmt.Errorf("cannot both use PreferServerCipherSuites and not use it")
 		}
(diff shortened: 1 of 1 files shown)


mholt:
Whoops. 😐 🌴 

Could you do me one more quick favor and just add it to [the test case in config_test.go](https://github.com/mholt/caddy/blob/17709a7d3f1a6ee0c34bdb81c6332b1339ea7085/caddytls/config_test.go#L13-L19)? (I know that we need to greatly expand these tests but I want to at least start with the things that bit us.)

mholt:
Nevermind, I want to work on these tests anyway. Merging now, as discussed on Slack. Thanks!!

PR #1084, opened 2016-08-31
faster fastcgi

Echsecutor:
This is to fix the fastcgi performance issue
https://github.com/mholt/caddy/issues/1082

By keeping the fastcgi connection open, answer latency is reduced by a factor 100 in my local setup.

This is a kind of quick fix, it would be better to include an option into the fastcgi directive to enable/disable connection upkeep and to set timeouts. See
http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_keep_conn
for the nginx equivalent.

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -18,7 +18,7 @@ import (
 
 	"github.com/mholt/caddy"
 	// plug in the HTTP server type
-	_ "github.com/mholt/caddy/caddyhttp"
+	_ "github.com/echsecutor/caddy/caddyhttp"
 
 	"github.com/mholt/caddy/caddytls"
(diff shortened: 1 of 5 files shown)


Echsecutor:
ok, I have signed the cla

Echsecutor:
Sorry.
a) I have messed up the pull request
b) It turns out that keeping connections alive was disabled for a reason. Apparently the de-mixing of queries is not implemented. (the fastcgi protocol allows to mix answers to multiple queries in the same stream. Those need to be properly sorted by the receiver.  A more real life example shows that this is not happening.)

So to really address https://github.com/mholt/caddy/issues/1082 one would probably need to improve the fastcgi client significantly.

PR #1087, opened 2016-09-01
persistent fastcgi connections

Echsecutor:
This fixes issue https://github.com/mholt/caddy/issues/1082#issuecomment-243919623 .

fastcgi connections are kept open, but, since the fcgiclient seems to be incapable of demuxing answers, I need to serialize the communication 'by hand'. If someone manages to include fastcgi demuxing, the serialization should be removed.
On linux, opening new connections didn't take noticeable time, hence here the new system could actually lead to a slowdown, due to serialization.

It would hence be nice to have a switch in the Caddy file to switch the persistent but serial fastcgi version provided here on and off. I have included a bool for this switch but did not work my way through the code to find the right spots to implement it. I guess someone familiar with the Caddyfile system can add this in no time.

### DIFF
--- a/caddyhttp/browse/browse_test.go
+++ b/caddyhttp/browse/browse_test.go
@@ -96,7 +96,7 @@ func TestSort(t *testing.T) {
 func TestBrowseHTTPMethods(t *testing.T) {
 	tmpl, err := template.ParseFiles("testdata/photos.tpl")
 	if err != nil {
-		t.Fatalf("An error occured while parsing the template: %v", err)
+		t.Fatalf("An error occurred while parsing the template: %v", err)
 	}
 
 	b := Browse{
@@ -134,7 +134,7 @@ func TestBrowseHTTPMethods(t *testing.T) {
 func TestBrowseTemplate(t *testing.T) {
 	tmpl, err := template.ParseFiles("testdata/photos.tpl")
 	if err != nil {
-		t.Fatalf("An error occured while parsing the template: %v", err)
+		t.Fatalf("An error occurred while parsing the template: %v", err)
 	}
 
 	b := Browse{
@@ -290,18 +290,19 @@ func TestBrowseJson(t *testing.T) {
 	for i, test := range tests {
 		var marsh []byte
 		req, err := http.NewRequest("GET", "/photos"+test.QueryURL, nil)
-
-		if err == nil && test.shouldErr {
-			t.Errorf("Test %d didn't error, but it should have", i)
-		} else if err != nil && !test.shouldErr {
-			t.Errorf("Test %d errored, but it shouldn't have; got '%v'", i, err)
+		if err != nil && !test.shouldErr {
+			t.Errorf("Test %d errored when making request, but it shouldn't have; got '%v'", i, err)
 		}
 
 		req.Header.Set("Accept", "application/json")
 		rec := httptest.NewRecorder()
 
 		code, err := b.ServeHTTP(rec, req)
-
+		if err == nil && test.shouldErr {
+			t.Errorf("Test %d didn't error, but it should have", i)
+		} else if err != nil && !test.shouldErr {
+			t.Errorf("Test %d errored, but it shouldn't have; got '%v'", i, err)
+		}
 		if code != http.StatusOK {
 			t.Fatalf("In test %d: Wrong status, expected %d, got %d", i, http.StatusOK, code)
 		}
--- a/caddyhttp/browse/browse_test.go
+++ b/caddyhttp/browse/browse_test.go
@@ -300,7 +300,7 @@ func TestBrowseJson(t *testing.T) {
 		req.Header.Set("Accept", "application/json")
 		rec := httptest.NewRecorder()
 
-		code, _ := b.ServeHTTP(rec, req)
+		code, err := b.ServeHTTP(rec, req)
 
 		if code != http.StatusOK {
 			t.Fatalf("In test %d: Wrong status, expected %d, got %d", i, http.StatusOK, code)
--- a/caddyhttp/browse/browse_test.go
+++ b/caddyhttp/browse/browse_test.go
@@ -300,7 +300,7 @@ func TestBrowseJson(t *testing.T) {
 		req.Header.Set("Accept", "application/json")
 		rec := httptest.NewRecorder()
 
-		code, err := b.ServeHTTP(rec, req)
+		code, _ := b.ServeHTTP(rec, req)
 
 		if code != http.StatusOK {
 			t.Fatalf("In test %d: Wrong status, expected %d, got %d", i, http.StatusOK, code)
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -19,7 +19,7 @@ import (
 
 // for persistent fastcgi connections
 var persistentConnections map[string]*FCGIClient = make(map[string]*FCGIClient)
-var metaMutex sync.RWMutex
+var metaMutex sync.Mutex
 var poorMansSerialisation map[string]*sync.Mutex = make(map[string]*sync.Mutex)
 
 // Handler is a middleware type that can handle requests as a FastCGI client.
@@ -86,16 +86,12 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 			var fcgiBackend *FCGIClient
 			reuse := false
 			if usePersistentFcgiConnections {
-				metaMutex.RLock()
+				metaMutex.Lock()
 				// re use connection, if possible
 				if _, ok := poorMansSerialisation[network+address]; !ok {
-					metaMutex.RUnlock()
-					metaMutex.Lock()
 					poorMansSerialisation[network+address] = new(sync.Mutex)
-					metaMutex.Unlock()
-					metaMutex.RLock()
 				}
-				metaMutex.RUnlock()
+				metaMutex.Unlock()
 
 				poorMansSerialisation[network+address].Lock()
 				defer poorMansSerialisation[network+address].Unlock()
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -16,6 +16,9 @@ import (
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
+// for persistent fastcgi connections
+var persistent_connections map[string]*FCGIClient
+
 // Handler is a middleware type that can handle requests as a FastCGI client.
 type Handler struct {
 	Next    httpserver.Handler
@@ -73,9 +76,22 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 
 			// Connect to FastCGI gateway
 			network, address := rule.parseAddress()
-			fcgiBackend, err := Dial(network, address)
-			if err != nil {
-				return http.StatusBadGateway, err
+
+			
+			// re use connection, if possible
+			if persistent_connections == nil{
+					persistent_connections = make(map[string]*FCGIClient)
+			}
+			fcgiBackend, ok := persistent_connections[network+address]
+			
+			// otherwise dial:
+			if(!ok || fcgiBackend == nil){
+				var err error
+				persistent_connections[network+address], err = Dial(network, address)
+				if err != nil {
+					return http.StatusBadGateway, err
+				}
+				fcgiBackend = persistent_connections[network+address]				
 			}
 
 			var resp *http.Response
@@ -91,11 +107,10 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 				resp, err = fcgiBackend.Post(env, r.Method, r.Header.Get("Content-Type"), r.Body, contentLength)
 			}
 
-			if resp.Body != nil {
-				defer resp.Body.Close()
-			}
-
+			
 			if err != nil && err != io.EOF {
+				persistent_connections[network+address].Close()
+				persistent_connections[network+address] = nil
 				return http.StatusBadGateway, err
 			}
 
@@ -112,6 +127,8 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 			if fcgiBackend.stderr.Len() != 0 {
 				// Remove trailing newline, error logger already does this.
 				err = LogError(strings.TrimSuffix(fcgiBackend.stderr.String(), "\n"))
+				persistent_connections[network+address].Close()
+				persistent_connections[network+address] = nil
 			}
 
 			// Normally we would return the status code if it is an error status (>= 400),
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -12,12 +12,14 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
 // for persistent fastcgi connections
 var persistent_connections map[string]*FCGIClient
+var poor_mans_serialisation map[string]*sync.Mutex
 
 // Handler is a middleware type that can handle requests as a FastCGI client.
 type Handler struct {
@@ -77,21 +79,28 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 			// Connect to FastCGI gateway
 			network, address := rule.parseAddress()
 
-			
 			// re use connection, if possible
-			if persistent_connections == nil{
-					persistent_connections = make(map[string]*FCGIClient)
+			if persistent_connections == nil {
+				persistent_connections = make(map[string]*FCGIClient)
+				poor_mans_serialisation = make(map[string]*sync.Mutex)
 			}
+			mut, ok := poor_mans_serialisation[network+address]
+			if !ok || mut == nil {
+				poor_mans_serialisation[network+address] = new(sync.Mutex)
+			}
+			poor_mans_serialisation[network+address].Lock()
+			defer poor_mans_serialisation[network+address].Unlock()
+
 			fcgiBackend, ok := persistent_connections[network+address]
-			
+
 			// otherwise dial:
-			if(!ok || fcgiBackend == nil){
+			if !ok || fcgiBackend == nil {
 				var err error
 				persistent_connections[network+address], err = Dial(network, address)
 				if err != nil {
 					return http.StatusBadGateway, err
 				}
-				fcgiBackend = persistent_connections[network+address]				
+				fcgiBackend = persistent_connections[network+address]
 			}
 
 			var resp *http.Response
@@ -107,7 +116,6 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 				resp, err = fcgiBackend.Post(env, r.Method, r.Header.Get("Content-Type"), r.Body, contentLength)
 			}
 
-			
 			if err != nil && err != io.EOF {
 				persistent_connections[network+address].Close()
 				persistent_connections[network+address] = nil
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -17,22 +17,23 @@ import (
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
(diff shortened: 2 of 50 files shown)


caddyhttp/fastcgi/fastcgi.go
```diff
@@ -35,6 +40,9 @@ type Handler struct {
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	for _, rule := range h.Rules {
 
+		//TODO: add an option in Caddyfile and pass it down to here
+		use_persistent_fcgi_connections := true
```
abiosoft: This is Python's naming convention :)
Go uses camel case.

Echsecutor:
Ok, I fixed the inaffective assignments (although they were not due to me) and changed to camel case.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -12,10 +12,15 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
+// for persistent fastcgi connections
+var PersistentConnections map[string]*FCGIClient
```
abiosoft: `persistentConnections` is preferred. It is not used outside of this package.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -12,10 +12,15 @@ import (
...
+var PoorMansSerialisation map[string]*sync.Mutex
```
abiosoft: same as above, lower case preferred.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -35,6 +40,9 @@ type Handler struct {
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	for _, rule := range h.Rules {
 
+		//TODO: add an option in Caddyfile and pass it down to here
+		UsePersistentFcgiConnections := true
```
abiosoft: same as above, lower case preferred.

caddyhttp/staticfiles/fileserver.go
```diff
@@ -103,7 +103,7 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
 				defer ff.Close()
 				dd, err := ff.Stat()
 				if err == nil {
-					name = index
+					//name = index //ineffectual
```
abiosoft: we can remove the line

mholt:
Thanks!! Excited to review this.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -35,6 +41,9 @@ type Handler struct {
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	for _, rule := range h.Rules {
 
+		//TODO: add an option in Caddyfile and pass it down to here
+		usePersistentFcgiConnections := runtime.GOOS == "windows"
```
mholt: Is there a technical reason this is limited to Windows (for now)? Is there any performance gain on Linux too? (maybe under load?)

caddyhttp/browse/browse_test.go
```diff
@@ -300,7 +300,7 @@ func TestBrowseJson(t *testing.T) {
 		req.Header.Set("Accept", "application/json")
 		rec := httptest.NewRecorder()
 
-		code, err := b.ServeHTTP(rec, req)
+		code, _ := b.ServeHTTP(rec, req)
```
mholt: These ineffectual assignment changes intrigue me. Our CI tests have been checking for these for months; how come the tests haven't been failing? How did you catch these?

elcore: @mholt 

```
ineffassign .

/src/github.com/mholt/caddy/caddyhttp/browse/browse_test.go:303:9: ineffectual assignment to err
/src/github.com/mholt/caddy/caddyhttp/fastcgi/fcgiclient.go:548:6: ineffectual assignment to err
/src/github.com/mholt/caddy/caddyhttp/staticfiles/fileserver.go:106:6: ineffectual assignment to name
/src/github.com/mholt/caddy/caddyhttp/staticfiles/fileserver_test.go:165:12: ineffectual assignment to err
/src/github.com/mholt/caddy/caddytls/client.go:58:5: ineffectual assignment to err
```

mholt: @elcore Of course, but I use it too and don't see that output. O.o

And the CI tests have been running it too for the last several months without showing it.

elcore: @mholt this is kind of spooky!

Echsecutor: Well, in my case the CI test where complaining about my pull request, pointing me to what @elcore mentioned. I would not have been picky about the style of your code, if your CI tools hadn't forced me ;)

mholt: Ahhh, I figured it out -- it's not your fault, @Echsecutor. ineffassign [was updated just a few days ago](https://github.com/gordonklaus/ineffassign/commit/99762ed65232b1477364c8c53778f8b6503e868c) to expose more ineffectual assignments. My local clone needed to be updated and now I see the same warnings. And it just so happens that this is the first pull request since that commit so it's the first one to show the warnings.

Thanks for addressing them!

caddyhttp/fastcgi/fcgiclient.go
```diff
@@ -545,7 +545,7 @@ func (c *FCGIClient) PostFile(p map[string]string, data url.Values, file map[str
 		if e != nil {
 			return nil, e
 		}
-		_, err = io.Copy(part, fd)
+		io.Copy(part, fd)
```
mholt: I'm wondering if we should handle this error instead of ignore it... nice catch by the updated ineffassign!

caddyhttp/staticfiles/fileserver.go
```diff
@@ -103,7 +103,6 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
 				defer ff.Close()
 				dd, err := ff.Stat()
 				if err == nil {
-					name = index
```
mholt: Woah, yikes. Was this always useless? 😛 Will have to investigate...

mholt:
I'm kind of thinking that all these errors should be treated instead of ignored, glad that the updated tool is catching them.

@Echsecutor I think we should split these pull requests into two -- unless you want to help us properly fix the ineffectual assignments (by not ignoring errors). :smile: I'm fine either way, but I think having two commits (one for the ineffassign fixes and the other for your fastcgi improvements) or two PRs would be best...

Echsecutor:
Sure, that sounds like a good idea. ;) I have removed the ineffectual assignment stuff from this pull request. (Now the CI checks are failing again)

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -91,14 +124,18 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 				resp, err = fcgiBackend.Post(env, r.Method, r.Header.Get("Content-Type"), r.Body, contentLength)
 			}
 
-			if resp.Body != nil {
-				defer resp.Body.Close()
-			}
-
 			if err != nil && err != io.EOF {
+				persistentConnections[network+address].Close()
+				persistentConnections[network+address] = nil
```
mholt: Perhaps this should be `delete(persistentConnections, network+address)` to free up the allocation?

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -10,12 +10,18 @@ import (
...
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
+// for persistent fastcgi connections
+var persistentConnections map[string]*FCGIClient
+var poorMansSerialisation map[string]*sync.Mutex
```
mholt: Use `make()` for both of these here so you don't have to have extra logic below (see comment)

caddyhttp/fastcgi/fastcgi.go
mholt: Is there a technical reason this is limited to Windows (for now)? Is there any performance gain on Linux too? (maybe under load?)

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -73,9 +82,33 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
+
+			var fcgiBackend *FCGIClient
+			var mut *sync.Mutex
+			ok := false
+			if usePersistentFcgiConnections {
+				// re use connection, if possible
+				if persistentConnections == nil {
+					persistentConnections = make(map[string]*FCGIClient)
+					poorMansSerialisation = make(map[string]*sync.Mutex)
+				}
```
mholt: This whole `if` block can go away if you just call `make` when you declare them above.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -112,6 +149,8 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 			if fcgiBackend.stderr.Len() != 0 {
 				// Remove trailing newline, error logger already does this.
 				err = LogError(strings.TrimSuffix(fcgiBackend.stderr.String(), "\n"))
+				persistentConnections[network+address].Close()
+				persistentConnections[network+address] = nil
```
mholt: Same here, consider delete() instead.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -73,9 +82,33 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
+			var mut *sync.Mutex
+			ok := false
+			if usePersistentFcgiConnections {
+				// re use connection, if possible
+				if persistentConnections == nil {
+					persistentConnections = make(map[string]*FCGIClient)
+					poorMansSerialisation = make(map[string]*sync.Mutex)
+				}
+				mut, ok = poorMansSerialisation[network+address]
+				if !ok || mut == nil {
```
mholt: If you change the code below to use `delete()` then you can reduce this just `if !ok`

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -73,9 +82,33 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
+					poorMansSerialisation[network+address] = new(sync.Mutex)
+				}
+				poorMansSerialisation[network+address].Lock()
+				defer poorMansSerialisation[network+address].Unlock()
+
+				fcgiBackend, ok = persistentConnections[network+address]
+			}
+
+			// otherwise dial:
+			if !ok || fcgiBackend == nil {
```
mholt: Same here, don't need the second check if `ok` entries always have a non-nil value.

mholt:
Cool, thanks! We're almost there! :+1:

caddyhttp/fastcgi/fastcgi.go
mholt: Also is this racy? We need to lock these map operations, no?

mholt:
@Echsecutor I have taken care of the ineffectual assignments on master; feel free to sync up when you are ready to finalize your PR!

caddyhttp/fastcgi/fastcgi.go
Echsecutor: I though in go there is a garbage collector, i.e. no explicit deallocation? Deleting from the map is a good idea, anyway, that saves one comparison. 

My intention was to use 
```
poorMansSerialisation[network+address].Lock()
defer poorMansSerialisation[network+address].Unlock()
```
to lock the map until returning from ServeHTTP. Maybe I misunderstood the way mutex work in go?

mholt: The garbage collector cleans up memory that is no longer reachable; map entries are very definitely still accessible.

The locks you've added appear to lock the FCGIClient for a specific backend, but these map reads/writes happen unsynchronized for each request that comes in. If one of them does happen concurrently with another, the runtime will panic. (It won't kill your server, because Caddy has a recover higher up, but it will kill the request.)

Consider a RWMutex to protect the map during concurrent access too.

caddyhttp/fastcgi/fastcgi.go
Echsecutor: I have tested a little ajax load using 
https://github.com/Echsecutor/fastcgi-test
on a ubuntu Laptop. Result: 

usePersistentFcgiConnections := true
the time for 101 requests varies between 0.7s and 1.6s

usePersistentFcgiConnections := false
the time for 101 requests varies between 0.6s and 1.5s

In other words, this measurements where not conclusive. ;) Not keeping the connection open and not serialising requests feels a little bit faster on Linux, but this might be a misconception.

caddyhttp/fastcgi/fastcgi.go
Echsecutor: access to the persistentConnections is locked via poorMansSerialisation. But you are right, I also need a lock for creating the second lists entries. Thanks!

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -86,16 +86,12 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
 			if usePersistentFcgiConnections {
-				metaMutex.RLock()
+				metaMutex.Lock()
 				// re use connection, if possible
 				if _, ok := poorMansSerialisation[network+address]; !ok {
-					metaMutex.RUnlock()
-					metaMutex.Lock()
 					poorMansSerialisation[network+address] = new(sync.Mutex)
-					metaMutex.Unlock()
-					metaMutex.RLock()
```
Echsecutor: This whole lock/unlock was not only ugly, unlocking in between or allowing for multiple reads at that point was not a good idea, anyway. ;)

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -73,9 +82,30 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
+			if usePersistentFcgiConnections {
+				metaMutex.Lock()
+				// re use connection, if possible
+				if _, ok := poorMansSerialisation[network+address]; !ok {
+					poorMansSerialisation[network+address] = new(sync.Mutex)
+				}
+				metaMutex.Unlock()
+
+				poorMansSerialisation[network+address].Lock()
+				defer poorMansSerialisation[network+address].Unlock()
```
mholt: These map accesses should probably be protected too, or better yet, assign it to a variable once so you don't have to keep reading into the map.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -73,9 +82,30 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
+
+				fcgiBackend, reuse = persistentConnections[network+address]
```
mholt: Here's another unprotected read.

Concurrency can be hard :wink:

Echsecutor: I would think that this is protected by the lock in line 96?

caddyhttp/fastcgi/fastcgi.go
Echsecutor: You are saying that I need to protect locking/unlocking the mutex because it's stored in a map?
If that's the case my map construction is pretty useless. I will change it.

caddyhttp/fastcgi/fastcgi.go
mholt: Those locks seem to protect the *FCGIClient, not the map that stores them.

caddyhttp/fastcgi/fastcgi.go
mholt: Well, you need to protect the map accesses, not merely the values inside it. Whenever you access a map concurrently, it must be synced with mutex.

Echsecutor:
Ok, I have learned a little bit about go today. ;)
Hope that it's now thread safe.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -12,10 +12,29 @@ import (
...
+	sync.Mutex
+	backend *FCGIClient
+}
+type concurrentPersistentConnectionsMap struct {
+	// for thread safe acces to the map
+	sync.Mutex
+	clientMap map[string]*serialClient
+}
+
+var persistentConnections *concurrentPersistentConnectionsMap = &(concurrentPersistentConnectionsMap{clientMap: make(map[string]*serialClient)})
```
mholt: You can remove the type declaration here (`*concurrentPersistentConnectionsMap) - see the lint warnings. It will shorten the line quite a bit :)

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -12,10 +12,29 @@ import (
...
+
+//TODO: add an option in Caddyfile and pass it down to here
+var UsePersistentFcgiConnections bool = true
```
mholt: Also remove the `bool` here.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -73,9 +92,25 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
-			if err != nil {
-				return http.StatusBadGateway, err
+
+			var fcgiBackend *FCGIClient
+			var client *serialClient
+			reuse := false
+			if UsePersistentFcgiConnections {
+				persistentConnections.Lock()
+				client, reuse = persistentConnections.clientMap[network+address]
+				persistentConnections.Unlock()
```
mholt: Now we're talkin'! :+1:

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -91,14 +126,18 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 				resp, err = fcgiBackend.Post(env, r.Method, r.Header.Get("Content-Type"), r.Body, contentLength)
 			}
 
-			if resp.Body != nil {
-				defer resp.Body.Close()
-			}
```
mholt: Where do we close the response body?

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -110,8 +149,12 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 
 			// Log any stderr output from upstream
 			if fcgiBackend.stderr.Len() != 0 {
+				persistentConnections.Lock()
 				// Remove trailing newline, error logger already does this.
 				err = LogError(strings.TrimSuffix(fcgiBackend.stderr.String(), "\n"))
```
mholt: Only lock what is absolutely necessary - do you need to lock the log line?

caddyhttp/fastcgi/fastcgi.go
Echsecutor: We don't if we want to UsePersistentFcgiConnections.

See fcgiclient.go, line 456
`		resp.Body = clientCloser{c, httputil.NewChunkedReader(rb)}`
line 408 ff
```
// clientCloser is a io.ReadCloser. It wraps a io.Reader with a Closer
// that closes FCGIClient connection.
type clientCloser struct {
	*FCGIClient
	io.Reader
}

func (f clientCloser) Close() error { return f.rwc.Close() }
```
i.e. closing the "body" actually closes the connection.

I have replaced this by fastcgi.go lin 139
`				defer fcgiBackend.Close()`

caddyhttp/fastcgi/fastcgi.go
Echsecutor: Ah you are right. I wanted to include the close (otherwise another thread might try to reuse the closed connection) but not the log. Even better to first lock and delete, than close. Thx.

Echsecutor:
I did some quick test/benchmarks (results in https://github.com/mholt/caddy/issues/1082#issuecomment-246039603 ) and also briefly tested the fastcgi backend that I am actually interested in, which yields similar results. I did not actually test the latest version against some php backend.

mholt:
Phew. :sweat_smile: There we go. Stay tuned in case there are any reports from users of problems related to this change, but I think this is definitely an overall improvement. Thank you so much for your work!

Echsecutor:
thanks for running this project! :)
I am currently convincing my team at work to use caddy instead of nginx as our web frontend's webserver. :)

PR #1091, opened 2016-09-05
Caddyfiles read from STDIN now have server types associated with them.

### LINKED ISSUE #1090: Caddyfiles read from STDIN have empty server types

0.9.1

Use a config file from STDIN.

```text
:2016 {
     root /var/www/public
}
```

`cat Caddyfile | caddy -conf stdin -log stdout`

Caddy running

```text
2016/09/05 18:25:03 no action found for directive 'root' with server type '' (missing a plugin?)
```

`echo -e ":2016 {\n root /var/www/public \n}" | caddy -conf stdin -log stdout`

------

I’ve found and fixed the problem; basically files parsed from STDIN do not have a server type associated with them, so HTTP plugins/directives are not correctly parsed.

I’m attaching a Pull Request which fixes this bug.

mikepulaski:
Fixes #1090.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -388,7 +388,7 @@ func (i *Instance) Wait() {
 // but the Input value will be nil. An error is only returned
 // if there was an error reading the pipe, even if the length
 // of what was read is 0.
-func CaddyfileFromPipe(f *os.File) (Input, error) {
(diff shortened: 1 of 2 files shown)


mholt:
Looks good - thanks for reporting and fixing it!! :100:

PR #1098, opened 2016-09-08
Updated FreeBSD init script

dprandzioch:
Hey Matt,

I just updated the FreeBSD init script to fix a bug which caused Caddy to not stop properly when running service caddy stop. Should work now, at least on the three machines I've tested it on, it works :-)

David


mholt:
Thank you David! :+1:

PR #1099, opened 2016-09-09
awslambda plugin

coopernurse:
This plugin proxies requests to AWS Lambda functions.  Related forum thread is here: https://forum.caddyserver.com/t/idea-aws-lambda-gateway-plugin/438/10

Documentation PR is here: https://github.com/caddyserver/caddyserver.com/pull/73

Thanks!

### DIFF
--- a/caddyhttp/awslambda/config.go
+++ b/caddyhttp/awslambda/config.go
@@ -1,0 +1,166 @@
+package awslambda
+
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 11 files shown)


coopernurse:
Looks like there may be some gofmt and golint issues. I'll look into those in the next day or two.

mholt:
Oh, thanks for the PR!

I think this will be better suited as a non-default plugin. In other words, you commit the plugin code to your own repo and the only PR we need to make here is to add your directive to the list of directives for the HTTP server.

coopernurse:
OK sounds good.  I'll close this PR and open a new one with that approach.

PR #1102, opened 2016-09-10
awslambda external plugin

coopernurse:
It looks like this is the only file that needed to be updated.  Please let me know if that's not the case.


mholt:
Thanks! I'll ping you if there's anything else that needs to be done to make the plugin available on the download page for the next release.

mholt:
@coopernurse Oh, can you add your plugin to the list at https://github.com/caddyserver/buildsrv/blob/master/features/registry.go ? - order doesn't matter there, but I try to group similar plugin types together. I need you to do the PR to assert that you consent to having your plugin on the download page, agree to maintain it, etc... (sorry it's a clumsy system right now -- working on that!)

coopernurse:
@mholt excellent - here you go!  https://github.com/caddyserver/buildsrv/pull/28

PR #1112, opened 2016-09-17
proxy: don't append predefined headers

tw4452852:
This PR contains two parts:

1. Fix a subtle bug in `TestDownstreamHeadersUpdate` and `TestUpstreamHeadersUpdate`: wrong logic when checking `Merge-Me` header. Also do a minor refactor.

2. Don't append a predefined header: Some plugins may already set response headers(e.g. MIME plugin). I think the proxy shouldn't append them which may lead to misleading or duplication.

### DIFF
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"reflect"
 	"runtime"
 	"strings"
 	"sync/atomic"
@@ -407,16 +408,19 @@ func TestUpstreamHeadersUpdate(t *testing.T) {
 	replacer := httpserver.NewReplacer(r, nil, "")
 
 	headerKey := "Merge-Me"
-	values, ok := actualHeaders[headerKey]
-	if !ok {
-		t.Errorf("Request sent to upstream backend does not contain expected %v header. Expected header to be added", headerKey)
-	} else if len(values) < 2 && (values[0] != "Initial" || values[1] != replacer.Replace("{hostname}")) {
-		t.Errorf("Values for proxy header `+Merge-Me` should be merged. Got %v", values)
+	got := actualHeaders[headerKey]
+	expect := []string{"Initial", "Merge-Value"}
+	if !reflect.DeepEqual(got, expect) {
+		t.Errorf("Request sent to upstream backend does not contain expected %v header: expect %v, but got %v",
+			headerKey, expect, got)
 	}
 
 	headerKey = "Add-Me"
-	if _, ok := actualHeaders[headerKey]; !ok {
-		t.Errorf("Request sent to upstream backend does not contain expected %v header", headerKey)
+	got = actualHeaders[headerKey]
+	expect = []string{"Add-Value"}
+	if !reflect.DeepEqual(got, expect) {
+		t.Errorf("Request sent to upstream backend does not contain expected %v header: expect %v, but got %v",
+			headerKey, expect, got)
 	}
 
 	headerKey = "Remove-Me"
@@ -425,12 +429,11 @@ func TestUpstreamHeadersUpdate(t *testing.T) {
 	}
 
 	headerKey = "Replace-Me"
-	headerValue := replacer.Replace("{hostname}")
-	value, ok := actualHeaders[headerKey]
-	if !ok {
-		t.Errorf("Request sent to upstream backend should not remove %v header", headerKey)
-	} else if len(value) > 0 && headerValue != value[0] {
-		t.Errorf("Request sent to upstream backend should replace value of %v header with %v. Instead value was %v", headerKey, headerValue, value)
+	got = actualHeaders[headerKey]
+	expect = []string{replacer.Replace("{hostname}")}
+	if !reflect.DeepEqual(got, expect) {
+		t.Errorf("Request sent to upstream backend does not contain expected %v header: expect %v, but got %v",
+			headerKey, expect, got)
 	}
 
 	if actualHost != expectHost {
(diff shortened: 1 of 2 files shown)


caddyhttp/proxy/proxy_test.go
```diff
@@ -407,16 +407,19 @@ func TestUpstreamHeadersUpdate(t *testing.T) {
...
 
 	headerKey := "Merge-Me"
-	values, ok := actualHeaders[headerKey]
-	if !ok {
-		t.Errorf("Request sent to upstream backend does not contain expected %v header. Expected header to be added", headerKey)
-	} else if len(values) < 2 && (values[0] != "Initial" || values[1] != replacer.Replace("{hostname}")) {
-		t.Errorf("Values for proxy header `+Merge-Me` should be merged. Got %v", values)
+	got := actualHeaders[headerKey]
+	expect := []string{"Initial", "Merge-Value"}
+	if len(got) != len(expect) || got[0] != expect[0] || got[1] != expect[1] {
```
mholt: Ah, yes, this is better.

elcore:
LGTM 👍

caddyhttp/proxy/reverseproxy.go
```diff
@@ -254,6 +254,10 @@ func (rp *ReverseProxy) copyResponse(dst io.Writer, src io.Reader) {
 
 func copyHeader(dst, src http.Header) {
 	for k, vv := range src {
+		// some predefined headers (e.g. MIME plugin), skip.
+		if _, ok := dst[k]; ok {
+			continue
+		}
```
mholt: So basically, this says that if the destination (whether upstream or downstream) already has the header field specified, we do not add it again. I can see how this fixes Content-Type problems, but what if we want to follow the Server header through all the proxy points?

I wonder if this check should be for certain fields only... what do you think?

tw4452852: Yes, it definitely makes sense. But what kind of fields should be skipped?
Also I think the expected result should be like this: suppose we set `Content-Type = text/css` with MIME plugin and upstream responses with `Content-Type = text/html, encoding=utf8`, the expected result should be `Content-Type = text/css, encoding=utf8` (just combination).

Above is just for `Content-Type`, what about other header fields?

mholt: Maybe Content-Length? I'm not sure. It would be good to find out how other web servers handle this.

tw4452852: Ok, I will do some research to try to find out a better way.

mholt: @tw4452852 I've been looking into it too, but I ran out of time pretty quick so I asked @adamwathan and he says: https://twitter.com/adamwathan/status/779007715670499328

> nginx will use the mime that comes back from the proxy if it's there, but if not it adds the correct header itself
> I just figured that out now though, by passing an empty Content-Type header I can trigger nginx to do the look up for me.

mholt: And then, according to [this tweet](https://twitter.com/adamwathan/status/779018371090178053), there's the docs for XSendfile: https://www.nginx.com/resources/wiki/start/topics/examples/xsendfile/

Which apparently doesn't modify these headers:

> Content-Type
Content-Disposition
Accept-Ranges
Set-Cookie
Cache-Control
Expires

However, this seems like it has more to do with the `internal` directive than proxy. Where do you think is the best place to implement this fix?

tw4452852: I think we could observe these headers from proxy, but if we have predefined some of them, they will be overwritten, is it a problem (user may think its plugin not works, e.g. MIME)? Maybe we could clarify this in documents.

Also I think proxy directive is a proper place.

mholt: @tw4452852 So you're saying if they're using the `mime` directive, for instance, it might break the logic here even if we ignore / don't modify these headers?

tw4452852: Oh, I misunderstood...

mholt:
@adamwathan Would you be able to test this branch and see if it fixes your situation with Valet? (cf. #1086) - it'll require building from source (`go build`) -- let me know if you would rather I provide a binary for you to test instead.

tw4452852:
Just do a minor revise. Use `reflect.DeepEqual` for `[]string` comparison.

adamwathan:
Hey @mholt sure happy to test! I'd love to learn how to build it myself, currently facing:

```
commands.go:8:2: cannot find package "github.com/flynn/go-shlex" in any of:
       	/usr/local/go/src/github.com/flynn/go-shlex (from $GOROOT)
       	($GOPATH not set)
caddy.go:29:2: cannot find package "github.com/mholt/caddy/caddyfile" in any of:
       	/usr/local/go/src/github.com/mholt/caddy/caddyfile (from $GOROOT)
       	($GOPATH not set)
```

Is there some dependency manager I need to install/setup?

adamwathan:
I got it built and running but not seeing any difference. All files still coming back with `Content-Type: text/html`:

image

This really simple repo makes it easy to reproduce: https://github.com/adamwathan/caddy-mime-bug

What should I expect to be seeing or what should I be trying in order to take advantage of the changes in this PR?

adamwathan:
I also tried what worked with nginx, which was to explicitly send an empty `Content-Type` header to get nginx to figure out the Content-Type itself, but in Caddy it seems to just keep the empty header:

image

It's difficult to not send the header at all because PHP FPM specifies a default Content-Type.

caddyhttp/proxy/reverseproxy.go
```diff
@@ -252,8 +252,25 @@ func (rp *ReverseProxy) copyResponse(dst io.Writer, src io.Reader) {
...
+
 func copyHeader(dst, src http.Header) {
 	for k, vv := range src {
+		if _, ok := dst[k]; ok {
+			// some predefined headers (e.g. MIME plugin) shoule skip.
+			if _, shouldSkip := skipHeaders[k]; shouldSkip {
+				continue
+			}
+			// otherwise, overwrite
+			dst.Del(k)
```
mholt: If we use `dst.Set()` below instead of `dst.Add()`, then maybe we can just get rid of this `Del()` line?

caddyhttp/proxy/reverseproxy.go
```diff
@@ -252,8 +252,25 @@ func (rp *ReverseProxy) copyResponse(dst io.Writer, src io.Reader) {
...
+	"Accept-Ranges":       {},
+	"Set-Cookie":          {},
+	"Cache-Control":       {},
+	"Expires":             {},
+}
+
 func copyHeader(dst, src http.Header) {
 	for k, vv := range src {
+		if _, ok := dst[k]; ok {
+			// some predefined headers (e.g. MIME plugin) shoule skip.
```
mholt: I'm not 100% sure this has anything to do with the mime plugin, maybe we should use this comment to refer to issue #1086 instead.

caddyhttp/proxy/reverseproxy.go
```diff
@@ -252,8 +252,25 @@ func (rp *ReverseProxy) copyResponse(dst io.Writer, src io.Reader) {
 	io.CopyBuffer(dst, src, buf.([]byte))
 }
 
+var skipHeaders = map[string]struct{}{
```
mholt: Could we add a comment referring to this comment (or the linked nginx docs page): https://github.com/mholt/caddy/pull/1112#discussion_r80092582 -- which explains why this map exists.

adamwathan:
@mholt Replied to your comment on the issue also thinking it was the PR 😄

https://github.com/mholt/caddy/issues/1086#issuecomment-249380150

caddyhttp/proxy/reverseproxy.go
tw4452852: Done.

caddyhttp/proxy/reverseproxy.go
tw4452852: Done.

caddyhttp/proxy/reverseproxy.go
mholt: Gotcha; let's not change this then.

mholt:
Woo hoo! Thanks Tw! And @adamwathan for your patience. :wink:

PR #1113, opened 2016-09-17
Attempt to fix broken absolute path imports (issue #1026)

### LINKED ISSUE #1026: import converting absolute path to relative path
for example: in my Caddyfile I have an `import /home/avail/.caddy/*.conf`, in a conf there, let's say `/home/avail/.caddy/avail.pw.conf` I have `import /etc/caddy/common.conf`, and it tries importing `/home/avail/.caddy/etc/caddy/common.conf`

mholt:
Removing these lines resolves #1026 and tests pass but I'm not sure what implications there might be. Shall we give this a go? /cc @pedronasser 

Will merge by default if no concerns in a day or two.

### DIFF
--- a/caddyfile/parse.go
+++ b/caddyfile/parse.go
@@ -244,27 +244,6 @@ func (p *parser) doImport() error {
 		if err != nil {
 			return err
 		}
-		var importLine int
-		importDir := filepath.Dir(importFile)
-		for i, token := range newTokens {
-			if token.Text == "import" {
-				importLine = token.Line
-				continue
-			}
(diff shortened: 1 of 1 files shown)


tw4452852:
Hi @mholt , I just look through this part of code, I think it is used to handle the recursive import case which is useful. For issue #1026, I think the culprit is [this line](https://github.com/mholt/caddy/blob/master/caddyfile/parse.go#L266) We should't join the path while the import file path is already absoluted.

So please hold on, I will give a pr later.

Issue #1107, opened 2016-09-13
Execute redirections before rewrites


hlidotbe:
### 1. What version of Caddy are you running (`caddy -version`)?

Caddy 0.9.1 (+e8e5595 Thu Sep 01 09:31:58 UTC 2016)
### 2. What are you trying to do?

Redirect old uri to their new location AND rewrite missing paths to default handler
### 3. What is your entire Caddyfile?

``` text
:8080
gzip
log stdout
errors stderr
root /var/www
fastcgi / /var/run/php-fpm.sock php

realip {
  from 172.17.0.0/16
}

redir 307 {
# Lots of from/to (about 200)
}

rewrite {
  r ^/(.*)$
  to {path} {path}/ /index.php?q={1}&{query}
}
```
### 4. How did you run Caddy (give the full command and describe the execution environment)?

Caddy is running in a docker container under runit supervision alongside php-fpm in an alpine 3.4 image.

``` sh
#!/bin/sh
export CADDYPATH=/tmp
exec chpst -u caddy /sbin/caddy -conf=/etc/caddy.conf
```
### 5. What did you expect to see?

I expect/want redirections to be run BEFORE rewrites. So if a redirect matches, it should be executed and stop further processing
### 6. What did you see instead (give full error messages and/or log)?

Not an error per-se but the final url is rewritten to `/index.php?q={1}&{query}` and never matches the redir paths
### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

```
:9000

root .
log stdout
errors stderr

redir 302 {
  /something /something-else
}

rewrite {
  to {path} /foo?q={path}
}
```

If I comment the rewrite block I'm redirected to /something-else but otherwise I see in the logs Caddy trying to fetch /foo?q=/something

hlidotbe:
Looking a bit further, I see why it happens. in `caddyhttp/httpserver/plugin.go` "redir" is after "header", way after "rewrite".

Is it preferable that I make a PR and discuss it there?

elcore:
Hello @hlidotbe,

feel free to create a PR 😄 (my opinion)

@mholt what do you think?

hlidotbe:
I was going to do that after the expires plugin 😀

elcore:
@hlidotbe 👍 😄

mholt:
What kind of effects would this change have for those who expect rewrites to happen before redirects?

hlidotbe:
[...]

PR #1125, opened 2016-09-22
Add expires directive

hlidotbe:
See https://github.com/caddyserver/caddyserver.com/pull/78

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -412,6 +412,7 @@ var directives = []string{
 	"ratelimit", // github.com/xuqingfeng/caddy-rate-limit
 	"search",    // github.com/pedronasser/caddy-search
 	"header",
+	"expires", // github.com/epicagency/caddy-expires
 	"redir",
 	"status",
 	"cors", // github.com/captncraig/cors/caddy


elcore:
LGTM 👍

mholt:
So, Expires is just a header -- why does it need a separate plugin to do this? Why not just the `header` directive?

elcore:
"It allows you to set different expiration durations depending on a matching regular expression." -- @hlidotbe 

> So, Expires is just a header -- why does it need a separate plugin to do this? Why not just the `header` directive?

hlidotbe:
@mholt because the expiration date is relative to the request time. It is to avoid the browser fetching static (and supposedly unchanging) assets. 

It is a simpler version of the expires modules of nginx and Apache (will extend it as time allows but already fit basic needs) 

I'm on mobile but I think there was an issue about the same need.

hlidotbe:
Redirect don't need expiration header no. I guess it could be moved closer to header to be more logical. I'll update the PR.

That being said, redirection should be moved much higher above rewrite (or rewrites lower), see #1107. In this case it has no influence though.

mholt:
Thanks! You'll just want to add it to the [build server](https://github.com/caddyserver/buildsrv/blob/master/features/registry.go) and then we can get it out with the next release.

PR #1126, opened 2016-09-22
Make FastCGI persistent connections optional; add tests

Echsecutor:
Here I deliver some code which I still owe you. ;)
This should have been in my last pull request already.

I have added a "persistent" (bool) option to the fastcgi directive to toggle the use of the new persistent fastcgi connections. Furthermore I have added test for the new directive, but more importantly also for the persistent connections. This test has already revealed, that the new persistent connection can be broken by to many parallel requests. I am investigating...
This is a quick fix for https://github.com/mholt/caddy/issues/1123

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: go
 
 go:
   - 1.6.3
-  - 1.7
+  - 1.7.1
   - tip
 
 matrix:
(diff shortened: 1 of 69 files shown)


mholt:
Sebastian, thank you very much for this PR! I'll have to see how this is different from #1129.

mholt:
Do you think the tests from this PR could be morphed pretty easily to work with #1129? /cc @abiosoft

mholt:
@Echsecutor I think we will go with @abiosoft's PR. However, yours has the TestPersistence function which I think would be good to have. Basically, you've already written a test for what his PR fixes, and I think it would be very good to have that test.

Echsecutor:
I have pulled the master with @abiosoft s pool and adapted the tests. Should I file a new PR or do you want to re-open this one?

abiosoft:
Create a new PR.

Thanks :)

PR #1129, opened 2016-09-23
Fastcgi persistent fix

abiosoft:
This builds on https://github.com/mholt/caddy/pull/1087 and fixes https://github.com/mholt/caddy/issues/1123. Reconnects if required and allows a configurable number of connection to reuse in the pool.

This instructs fastcgi to attempt to persist and reuse up to 5 connections. 
```
fastcgi / 127.0.0.1:900 php {
    pool 5
}
```
It is disabled by default unless the configuration is added and `pool > 0`.

`pool 1` works fine from all my tests.

#### Persistent Connection
Caddyfile
```
0.0.0.0
root /src
fastcgi / fpm:9000 php {
     pool 1
}
```
Result
```
❯ hey -n 1000 -c 20 http://localhost:2015
Summary:
  Total:	8.8103 secs
  Slowest:	0.3320 secs
  Fastest:	0.0379 secs
  Average:	0.1743 secs
  Requests/sec:	113.5040

Status code distribution:
  [200]	1000 responses

Response time histogram:
  0.038 [1]	|
  0.067 [4]	|∎
  0.097 [20]	|∎∎∎
  0.126 [69]	|∎∎∎∎∎∎∎∎∎
  0.156 [237]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.185 [293]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.214 [224]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.244 [99]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.273 [36]	|∎∎∎∎∎
  0.303 [14]	|∎∎
  0.332 [3]	|

Latency distribution:
  10% in 0.1281 secs
  25% in 0.1482 secs
  50% in 0.1717 secs
  75% in 0.1989 secs
  90% in 0.2263 secs
  95% in 0.2460 secs
  99% in 0.2856 secs
```
#### Non-Persistent Connection
Caddyfile
```
0.0.0.0
root /src
fastcgi / fpm:9000 php
```
Result
```
❯ hey -n 1000 -c 20 http://localhost:2016
Summary:
  Total:	10.2372 secs
  Slowest:	0.4034 secs
  Fastest:	0.0261 secs
  Average:	0.2017 secs
  Requests/sec:	97.6830

Status code distribution:
  [200]	1000 responses

Response time histogram:
  0.026 [1]	|
  0.064 [4]	|∎
  0.102 [25]	|∎∎∎
  0.139 [71]	|∎∎∎∎∎∎∎∎∎
  0.177 [227]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.215 [303]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.252 [210]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.290 [97]	|∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.328 [48]	|∎∎∎∎∎∎
  0.366 [8]	|∎
  0.403 [6]	|∎

Latency distribution:
  10% in 0.1388 secs
  25% in 0.1677 secs
  50% in 0.1988 secs
  75% in 0.2340 secs
  90% in 0.2727 secs
  95% in 0.2970 secs
  99% in 0.3450 secs
```
Not much difference in the benchmark results and persistent connection is a bit faster.

### DIFF
--- a/caddyhttp/fastcgi/dialer.go
+++ b/caddyhttp/fastcgi/dialer.go
@@ -1,0 +1,58 @@
+package fastcgi
+
+import "sync"
+
+type dialer interface {
+	Dial() (*FCGIClient, error)
+	Close(*FCGIClient) error
+}
+
+// basicDialer is a basic dialer that wraps default fcgi functions.
+type basicDialer struct {
+	network, address string
+}
+
+func (b basicDialer) Dial() (*FCGIClient, error) { return Dial(b.network, b.address) }
+func (b basicDialer) Close(c *FCGIClient) error  { return c.Close() }
+
+// persistentDialer keeps a pool of fcgi connections.
+// connections are not closed after use, rather added back to the pool for reuse.
+type persistentDialer struct {
+	size    int
+	network string
+	address string
+	pool    []*FCGIClient
+	sync.Mutex
+}
+
+func (p *persistentDialer) Dial() (*FCGIClient, error) {
+	p.Lock()
+
+	// connection is available, return first one.
+	if len(p.pool) > 0 {
+		client := p.pool[0]
+		p.pool = p.pool[1:]
+		p.Unlock()
+
+		return client, nil
+	}
+
+	p.Unlock()
+
+	// no connection available, create new one
+	return Dial(p.network, p.address)
+}
+
+func (p *persistentDialer) Close(client *FCGIClient) error {
+	p.Lock()
+	if len(p.pool) < p.size {
+		// pool is not full yet, add connection for reuse
+		p.pool = append(p.pool, client)
+		p.Unlock()
+	} else {
+		// otherwise, close the connection.
+		p.Unlock()
+		return client.Close()
+	}
+	return nil
+}
--- a/caddyhttp/fastcgi/dialer.go
+++ b/caddyhttp/fastcgi/dialer.go
@@ -49,10 +49,12 @@ func (p *persistentDialer) Close(client *FCGIClient) error {
 		// pool is not full yet, add connection for reuse
 		p.pool = append(p.pool, client)
 		p.Unlock()
-	} else {
-		// otherwise, close the connection.
-		p.Unlock()
-		return client.Close()
+
+		return nil
 	}
-	return nil
+
+	p.Unlock()
+
+	// otherwise, close the connection.
+	return client.Close()
 }
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -12,28 +12,9 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
-
-// persistent fastcgi connections
-
-type serialClient struct {
-	// for read/write serialisation
-	sync.Mutex
-	backend *FCGIClient
-}
-type concurrentPersistentConnectionsMap struct {
-	// for thread safe acces to the map
-	sync.Mutex
-	clientMap map[string]*serialClient
-}
-
-var persistentConnections = &(concurrentPersistentConnectionsMap{clientMap: make(map[string]*serialClient)})
-
-// UsePersistentFcgiConnections TODO: add an option in Caddyfile and pass it down to here
-var UsePersistentFcgiConnections = true
 
 // Handler is a middleware type that can handle requests as a FastCGI client.
 type Handler struct {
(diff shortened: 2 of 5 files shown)


caddyhttp/fastcgi/dialer.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	p.Unlock()
+
+	// no connection available, create new one
+	return Dial(p.network, p.address)
+}
+
+func (p *persistentDialer) Close(client *FCGIClient) error {
+	p.Lock()
```
abiosoft: I was trying to avoid `client.Close()` to bite into the lock. But considering it's not really an overhead, your suggestion is cleaner.

mholt: Ah, gotcha. That makes sense. I trust your judgment then, either way is fine with me.

caddyhttp/fastcgi/dialer.go
```diff
@@ -0,0 +1,60 @@
...
+func (p *persistentDialer) Dial() (*FCGIClient, error) {
+	p.Lock()
+
+	// connection is available, return first one.
+	if len(p.pool) > 0 {
+		client := p.pool[0]
+		p.pool = p.pool[1:]
+		p.Unlock()
+
+		return client, nil
```
Echsecutor: Notice that the persistentDialer's network and address must not be changed after the first Dial(), otherwise the returned connection might be a pooled one using the old adress/network.

Echsecutor: (I do not think that this is of any concern in Caddy. Here Network and Adress are fixed for the fastcgi component after reading the Caddyfile.)

abiosoft: Each `Rule` has its own persistent connection pool which is initiated with the network and address configured for that rule. So it is not gonna change.

abiosoft:
@Echsecutor I should thank you. If you hadn't created that PR, Caddy would not have this feature now.

abiosoft:
Nope, if the connection is dead, it won't complete serving the request i.e. an error will occur. So it won't get to line of code that adds it back to the pool.

caddyhttp/fastcgi/dialer.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	p.Unlock()
+
+	// no connection available, create new one
+	return Dial(p.network, p.address)
+}
+
+func (p *persistentDialer) Close(client *FCGIClient) error {
```
Echsecutor: Unfortunately this function is unreachable, i.e. never called. (Insert some trivial fmt.Println("...") to verify).

The problem is, that in fcgiclient.go line 182 the closer for the fastcgiclient is always registered to be the dialer.Dial, i.e.  actual tcp connection and not the new dialer.

We can not simply replace the rwc member, this does more than just closing, but we could modify 
func (c *FCGIClient) Close() error

abiosoft: Oh my I see what you mean. https://github.com/mholt/caddy/pull/1129/files#diff-9900852dfb974061eb746138baa758b9R93 should be `defer rule.dialer.Close(fcgiBackend)`. My mistake.

abiosoft: @Echsecutor can you change fastcgi.go:93 to `defer rule.dialer.Close(fcgiBackend)` and build and see if that fixes it.

I do not have a Windows machine to test the behaviour.

Echsecutor: @abiosoft yepp, that's a way to fix the problem. 

(You do not need Windows, my tests in PR https://github.com/mholt/caddy/pull/1134 (or just chekcout my master) report on the number of used connections with test -v.)

Echsecutor:
> Nope, if the connection is dead, it won't complete serving the request i.e. an error will occur. So it won't get to line of code that adds it back to the pool.

Currently, `defer rule.dialer.Close(fcgiBackend)` is on line 93, right? That's before the errors occur. We should move it down to below the  `// Write the response body` block to prevent connections from re-entering the pool.

PR #1135, opened 2016-09-24
proxy: Improve failover logic and retries

### LINKED ISSUE #1069: Memory/cpu leak in proxy directive when max_fails 0 and backend fails
v0.9.1

Serving traffic via the proxy directive using max_fails 0 as only one backend is available.

```text
:2015 {
  proxy / 127.0.0.1:8080 {
    transparent
    max_fails 0
  }
}
```

./caddy --agree --conf=caddyfile_test

No memory leak and no 100% CPU usage on 4 cores.

4 cores at 100%, 4GB memory filled in under 1 minute after the first backend failure arose.

Also stdout seems to go crazy with:
```
goroutine 2826341 [sleep, 2 minutes]:

time.Sleep(0x2540be400)

        /usr/local/go/src/runtime/time.go:59 +0xe1

github.com/mholt/caddy/caddyhttp/proxy.Proxy.ServeHTTP.func1(0xc4202c5420, 0x2540be400)

        /tmp/custombuild_111_575206995/src/github.com/mholt/caddy/caddyhttp/proxy/proxy.go:157 +0x2b

created by github.com/mholt/caddy/caddyhttp/proxy.Proxy.ServeHTTP

        /tmp/custombuild_111_575206995/src/github.com/mholt/caddy/caddyhttp/proxy/proxy.go:159 +0x44e



goroutine 2831798 [sleep, 2 minutes]:

time.Sleep(0x2540be400)

        /usr/local/go/src/runtime/time.go:59 +0xe1

github.com/mholt/caddy/caddyhttp/proxy.Proxy.ServeHTTP.func1(0xc4202c5420, 0x2540be400)

        /tmp/custombuild_111_575206995/src/github.com/mholt/caddy/caddyhttp/proxy/proxy.go:157 +0x2b

created by github.com/mholt/caddy/caddyhttp/proxy.Proxy.ServeHTTP

        /tmp/custombuild_111_575206995/src/github.com/mholt/caddy/caddyhttp/proxy/proxy.go:159 +0x44e
```

* use proxy with max_fails 0
* have a high traffic server
* force a backend unreachable error
* see how it goes crazy

As always edge cases \o/

The fix for now was to use max_fails to 50, which doesn't get reached under usual circumstances.

mholt:
This PR should fix #1069 and resolve a few other outstanding irritants in proxy behavior for backends that are less-than-robust. In summary:

- Proxy no longer retries by default when a request fails, because `tryDuration` (now a configurable value on the Upstream itself) defaults to 0. This means that, by default, the downstream client must retry their request when a host initially goes down or no hosts are available.

- New configuration subdirectives: `try_duration` and `try_interval`. try_duration specifies how long to try selecting hosts (used to be 60s) until one succeeds. The introduction of try_interval (with a default of 1s) is there to prevent tight loops that spin the CPU in case all the requests fail gloriously fast; this also operates under the assumption that hosts won't likely come back up within a fraction of a second and that a client can wait 1s when an upstream host initially goes down.

- fail_timeout now defaults to 0, which means that request failure counting is disabled. In other words, a failed request does not count against the upstream host's availability. Since max_fails defaults to 1 (which is sensible), a single request failure could mark the upstream host down for every client for 10 seconds. This is still the case, but now the user has to opt-in to this behavior.

- max_fails can no longer be < 1. People used to set this to 0 to work around the sporadic failures or "long-lived" downtime (of 10 seconds) even if the actual downtime may have been a fluke or a second or two. This resulted in the CPU spinning in a loop and 10s of "502 Bad Gateway" even though the host came up 9 seconds earlier. Anyway, this value is now irrelevant if fail_timeout is 0. The user has to explicitly enable fail_timeout in order to count request failures.

- We report the actual backend error rather than "unreachable backend".

I have also [updated the proxy docs](https://github.com/caddyserver/caddyserver.com/commit/06f46e3875a347efabd34bb8b839c4e467497861) in an attempt to be clearer in what these parameters do.

@nemothekid: Would love to have your review/approval on this! :smile:

### DIFF
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -13,8 +13,6 @@ import (
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
-var errUnreachable = errors.New("unreachable backend")
-
 // Proxy represents a middleware instance that can proxy requests.
 type Proxy struct {
 	Next      httpserver.Handler
@@ -92,10 +90,14 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	// since Select() should give us "up" hosts, keep retrying
 	// hosts until timeout (or until we get a nil host).
 	start := time.Now()
+	var backendErr error
 	for time.Now().Sub(start) < tryDuration {
 		host := upstream.Select(r)
 		if host == nil {
-			return http.StatusBadGateway, errUnreachable
+			if backendErr == nil {
+				backendErr = errors.New("no hosts available upstream")
+			}
+			return http.StatusBadGateway, backendErr
 		}
 		if rr, ok := w.(*httpserver.ResponseRecorder); ok && rr.Replacer != nil {
 			rr.Replacer.Set("upstream", host.Name)
@@ -141,7 +143,7 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 
 		// tell the proxy to serve the request
 		atomic.AddInt64(&host.Conns, 1)
-		backendErr := proxy.ServeHTTP(w, outreq, downHeaderUpdateFn)
+		backendErr = proxy.ServeHTTP(w, outreq, downHeaderUpdateFn)
 		atomic.AddInt64(&host.Conns, -1)
 
 		// if no errors, we're done here; otherwise failover
@@ -159,7 +161,7 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 		}(host, timeout)
 	}
 
-	return http.StatusBadGateway, errUnreachable
+	return http.StatusBadGateway, backendErr
 }
 
 // match finds the best match for a proxy config based
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -94,17 +94,33 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	// outreq is the request that makes a roundtrip to the backend
 	outreq := createUpstreamRequest(r)
 
-	// since Select() should give us "up" hosts, keep retrying
-	// hosts until timeout (or until we get a nil host).
+	// The keepRetrying function will return true if we should
+	// loop and try to select another host, or false if we
+	// should break and stop retrying.
 	start := time.Now()
+	keepRetrying := func() bool {
+		// if we've tried long enough, break
+		if time.Now().Sub(start) >= upstream.GetTryDuration() {
+			return false
+		}
+		// otherwise, wait and try the next available host
+		time.Sleep(upstream.GetTryInterval())
+		return true
+	}
+
 	var backendErr error
 	for {
+		// since Select() should give us "up" hosts, keep retrying
+		// hosts until timeout (or until we get a nil host).
 		host := upstream.Select(r)
 		if host == nil {
 			if backendErr == nil {
 				backendErr = errors.New("no hosts available upstream")
 			}
-			return http.StatusBadGateway, backendErr
+			if !keepRetrying() {
+				break
+			}
+			continue
 		}
 		if rr, ok := w.(*httpserver.ResponseRecorder); ok && rr.Replacer != nil {
 			rr.Replacer.Set("upstream", host.Name)
@@ -170,12 +186,9 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 		}
 
 		// if we've tried long enough, break
-		if time.Now().Sub(start) >= upstream.GetTryDuration() {
+		if !keepRetrying() {
 			break
 		}
-
-		// otherwise, wait and try the next available host
-		time.Sleep(upstream.GetTryInterval())
 	}
 
 	return http.StatusBadGateway, backendErr
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -80,10 +80,6 @@ func (uh *UpstreamHost) Available() bool {
 	return !uh.Down() && !uh.Full()
 }
 
-// tryDuration is how long to try upstream hosts; failures result in
-// immediate retries until this duration ends or we get a nil host.
-var tryDuration = 60 * time.Second
-
 // ServeHTTP satisfies the httpserver.Handler interface.
 func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	// start by selecting most specific matching upstream config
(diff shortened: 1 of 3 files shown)


caddyhttp/proxy/proxy.go
```diff
@@ -89,13 +94,33 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
...
-	// since Select() should give us "up" hosts, keep retrying
-	// hosts until timeout (or until we get a nil host).
+	// The keepRetrying function will return true if we should
+	// loop and try to select another host, or false if we
+	// should break and stop retrying.
 	start := time.Now()
-	for time.Now().Sub(start) < tryDuration {
+	keepRetrying := func() bool {
+		// if we've tried long enough, break
+		if time.Now().Sub(start) >= upstream.GetTryDuration() {
```
tw4452852: Maybe `time.Since()`?

mholt: Good idea. Done!

elcore:
It is beautiful and a welcome change 😄 -- thank you @mholt 

P.S.: I was working on basically the same thing, time to delete my private branch.

mholt:
Great, I will merge this just before we roll out 0.9.3 or whenever @nemothekid gives it a look (whichever comes first).

nemothekid:
1. A default `try_interval` of 1s seems really high - if you have more than one host. For the end user, if they get a bad server, and theres a chance they could get a new good server, having to wait an entire second for the page to load isn't very attractive.

 2. The docs should also probably highlight the fact that if you want to properly take advantage of load balancing among multiple servers, while not routing requests to bad servers, you have to set `fail_timeout`. 

Other than that LGTM.

mholt:
@nemothekid Excellent points - thank you.

1. I'm reducing the default try_interval to 250ms. That should be enough to prevent spinning and still get a new host reasonably quickly relative to the speed of most networks. (And, that only applies to a request when the host initially goes down -- not all requests, fortunately.)

2. Updating the docs now to clarify that.

Merging in a few minutes!

jazoom:
I am happy to see this because I've been having 502 issues with Caddy. Just FYI about this:

1. https://caddyserver.com/docs/proxy  -  this page mentions "interval_duration" under "health_check", but there doesn't seem to be anything by that name?

2. I've read the release notes for 0.9.3 multiple times, as well as the notes on that proxy pages linked to in 1. It's not clear to me why this is true: "Note that in order to do proper, redundant load balancing in the event of failures, you must set fail_timeout and try_duration to values > 0." Why do you need that with the health checks? Why would I want my server being marked as down if it returns max_fails number of 404 responses in fail_timeout? Do only 5XX responses get added to the tally?

mholt:
@jazoom 

1. I'll change that; if you look at the syntax declaration, interval_duration refers to the health_check_interval value.

> Why would I want my server being marked as down if it returns max_fails number of 404 responses in fail_timeout?

It doesn't. If your _health checks_ get a 404, that's an error, but max_fails only caps the connection failures that occur with actual client requests. Health checks and max_fails control different things.

jazoom:
That's what I mean. Why would anyone want a 404 or 400 or 401 or anything like that for a regular request to mark the server as down? Those are issues with the request, not the server. I'm guessing I'm just not understanding the docs because making Caddy work this way makes no sense.

mholt:
@jazoom

> Why would anyone want a 404 or 400 or 401 or anything like that for a regular request to mark the server as down?

Like I said, that's not what happens. Client requests that get a 4xx error don't mark the backends as down.

jazoom:
So it doesn't count 5XX errors either then? It's lower level than that? Because the docs repeatedly say "failed request", not "connection error". To me, and perhaps to most people, a failed request refers to an HTTP request that gets back a non-2XX/3XX response. I might be wrong.

mholt:
@jazoom Requests from clients that get a 5xx or 4xx response don't count against max_fails. Only connection errors. A "failed" request is one that never gets through.

jazoom:
Okay, it all makes sense now. Thank you for taking the time to clarify that. Keep up the great work!

tomasdeml:
@mholt Live docs for the `proxy` directive say that the default value for `try_interval` is 1s. Shouldn't the page be updated to say 250ms?

mholt:
Yup, I'll fix that too.

jazoom:
I'm still getting 502 bad gateway errors frequently, that I never got with HAProxy. Maybe I've misconfigured the proxy?

    policy least_conn
    fail_timeout 10s
    max_fails 1
    try_duration 2s
    try_interval 250ms
    health_check /healthCheck
    health_check_interval 20s
    health_check_timeout 3s
    keepalive 100
    transparent

While trying to work out what's going on, I've realised I'm still not 100% clear on some things. By my config, do the health checks get called every 20 seconds, and if they don't return within 3s, or if they are non 2xx/3xx, that counts as one fail, which adds to the max_fails tally, which is the SAME tally as the connection errors you were talking about?

Or, does Caddy only do health checks when an upstream server is down (by the connection failure method), and this is Caddy's method of determining when it's back up? This is what I'm assuming since the docs say this: "specifies the time between each health check on unhealthy backends", and I don't see the health checks being logged. If this is so, is there a way to use the health checks as the sole way of determining server health? I'm interested in this since Caddy continues to throw out "bad gateway", even though I had the same setup running for a year with HAProxy and never ran into these issues. I'm wondering if it's related to caddy thinking there are connection errors when there really aren't. Perhaps increasing max_fails will help?

mholt:
It's probably because your backend is still, well, "flaky" (for lack of a better word) - and you've configured the proxy to act like it used to. Your health checks are running every 20 seconds. Health checks do not count against the max_fails count, because they are not client requests. If a health check fails, the backend is marked down until it comes back up.

But yes, perhaps in your case, increasing max_fails from 1 would be a good idea. So start there. With max_fails of 1 and the rest of your config, you're saying that if even 1 client gets a connection error, mark the whole backend as down for everyone for 10 seconds. This is why we changed the default settings.

jazoom:
I'd be pretty surprised if all the backends were that flaky since there are 3 of them connected over a Digital Ocean private network, and a completely separate cluster connected over a Vultr private network (all direct via private IP). All of this worked with no bad gateway errors until the day I switched from HAProxy to Caddy. I would just switch back but I believe in this project and it's a pain to set up Let's Encrypt with HAProxy.

I'll try increasing max_fails.

jazoom:
Okay, I did actually increase max_fails to 3 last night. It seems it solved the problem. The interesting thing is, I get notified of all communication failures, and there were none, just like there were never any before switching to Caddy. I don't understand where the bad gateways come from when max_fails is 1, but it seems to be all good now.

mholt:
Any errors communicating with the backend should be logged as of 0.9.3, so you should see it if you have `errors` directive logging them somewhere.

jazoom:
Nice. I'll be sure to let you know what the logs say if this issue comes back.

jazoom:
By the way, the docs don't actually mention how often health checks are called. It says how often they're called on unhealthy backends, what about how often they are called on healthy backends? That should probably be configurable too.

mholt:
@jazoom It's the same either way, I believe.

PR #1139, opened 2016-09-27
Fastcgi persistent fix.. again

abiosoft:


mholt:
Okay, closing this PR since, as I understand it (and see it), the other PR (linked above) has this change plus tests, so we'll go with that one. :+1:

PR #1143, opened 2016-09-27
1136 fix (IP hash policy no longer removes down host from pool)

krishamoud:
This should fix https://github.com/mholt/caddy/issues/1136.

Instead of modifying the pool directly I make a copy of it and remove hosts from the copy.  I'm not sure if this is the best way to do this or not.

It's necessary to keep track of the host pool because simply decreasing the `poolLen` could result in traffic still being sent to the down host.

### DIFF
--- a/caddyhttp/proxy/policy.go
+++ b/caddyhttp/proxy/policy.go
@@ -121,18 +121,13 @@ func (r *IPHash) Select(pool HostPool, request *http.Request) *UpstreamHost {
 	if err != nil {
 		clientIP = request.RemoteAddr
 	}
-	hash := hash(clientIP)
-	for {
-		if poolLen == 0 {
-			break
-		}
-		index := hash % poolLen
-		host := pool[index]
+	index := hash(clientIP) % poolLen
+	for i := uint32(0); i < poolLen; i++ {
+		index += i
+		host := pool[index%poolLen]
 		if host.Available() {
 			return host
 		}
-		pool = append(pool[:index], pool[index+1:]...)
-		poolLen--
 	}
 	return nil
 }
--- a/caddyhttp/proxy/policy_test.go
+++ b/caddyhttp/proxy/policy_test.go
@@ -163,14 +163,14 @@ func TestIPHashPolicy(t *testing.T) {
 	request.RemoteAddr = "172.0.0.1"
 	pool[1].Unhealthy = true
 	h = ipHash.Select(pool, request)
-	if h != pool[0] {
-		t.Error("Expected ip hash policy host to be the first host.")
+	if h != pool[2] {
+		t.Error("Expected ip hash policy host to be the third host.")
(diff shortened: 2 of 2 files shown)


mholt:
Thanks for working on this by the way. :smile: We just need to make sure the logic is correct.

krishamoud:
Thanks for letting me be a contributor to this awesome project! 😄 

If you have a 3 host pool and one host is unhealthy the length of the pool will still be 3.  Assuming the IP doesn't change the hash will always be the same, the index to return will always be the same.

If `hash % len(pool)` returns an unhealthy host and the pool isn't modified then requests will always be sent to the down host even if healthy hosts are available.

mholt:
Ah, right, I was just thinking about this when I saw your reply. Let me just double-check my understanding of slices in Go and make sure this doesn't affect the _actual_ pool.

In the meantime, do you think you could write a test for this policy? Well, you already have several, but maybe a test to ensure that a downed host that comes back up will "go back into" the pool.

mholt:
@krishamoud Ruh-roh: https://play.golang.org/p/R9-qBBk1i9

Note the final line should be the same as the first; changing the underlying array like this still changes the original, just the slice header gets copied. Maybe there's a better way of doing this, like a map iteration...?

krishamoud:
That's unfortunate.  I should brush up on my understanding of slices as well...

I'll fix this and I'll add a test for this case as well.

tw4452852:
I think we'd better not modify the pool because it may lead unexpected result later...

Maybe this is better?

```
func (r *IPHash) Select(pool HostPool, request *http.Request) *UpstreamHost {
	poolLen := len(pool)
	clientIP, _, err := net.SplitHostPort(request.RemoteAddr)
	if err != nil {
		clientIP = request.RemoteAddr
	}
	index := hash(clientIP) % poolLen
	for i := 0; i < poolLen; i++ {
		index += i
		host := pool[index]
		if host.Available() {
			return host
		}
	}
	return nil
}
```

mholt:
^ That is pretty good, close to what I was thinking. But it could run out of bounds. If `index > 0`, we wouldn't be able to reach the hosts at the beginning of the pool. So maybe we could start `i` at `index` and then do a bounds check so we loop back to 0, and terminate once `i == index` again, or something.

tw4452852:
Oh, yes. `s/pool[index]/pool[index%poolLen]/`

mholt:
That's even better :smile:

krishamoud:
@tw4452852 can you clarify that for me?

tw4452852:
@krishamoud Of course. we first select a host according to the hash. If the host is available, just return it, otherwise try the next one until we loop all the hosts in the pool.

mholt:
Thanks! This is great now (I think, haha) - really appreciate your help, both of you. :+1:

PR #1144, opened 2016-09-28
Remove the eager check in the browse middleware

pbeckmann:
Hello,

it is just the removing of the eager check in the browse middleware as stated in https://github.com/mholt/caddy/issues/1085

Caddy will start and throw a 404-error until the directory will be created.

Hopes that helps you ;)

### DIFF
--- a/caddyhttp/browse/setup_test.go
+++ b/caddyhttp/browse/setup_test.go
@@ -66,4 +66,16 @@ func TestSetup(t *testing.T) {
 			}
 		}
 	}
+
+	// test case #6 tests startup with missing root directory in combination with default browse settings
+	controller := caddy.NewTestController("http", "browse")
+	cfg := httpserver.GetConfig(controller)
+
+	//Manipulate the root to a nonexistent directory, must be done, cause the default test root exists
+	cfg.Root = nonExistantDirPath
+	err = setup(controller)
+
+	if err != nil {
+		t.Errorf("Test case #%d received an error of %v", 6, err)
+	}
 }
--- a/caddyhttp/browse/setup_test.go
+++ b/caddyhttp/browse/setup_test.go
@@ -18,7 +18,7 @@ func TestSetup(t *testing.T) {
 	if err != nil {
 		t.Fatalf("BeforeTest: Failed to find an existing directory for testing! Error was: %v", err)
 	}
-	nonExistantDirPath := filepath.Join(tempDirPath, strconv.Itoa(int(time.Now().UnixNano())))
+	nonExistentDirPath := filepath.Join(tempDirPath, strconv.Itoa(int(time.Now().UnixNano())))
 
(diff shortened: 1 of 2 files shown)


tw4452852:
Add a test?

pbeckmann:
Hey,

i'm not really sure how to write this test. I tried

`{"root " + nonExistantDirPath + "\n browse", []string{nonExistantDirPath, "/"}, false},`

in the https://github.com/mholt/caddy/blob/master/caddyhttp/browse/setup_test.go file.

But the test never fails. I'm looking into it. :) But I'm a Golang newbie. So it will need some time.

Sorry for the inconvenience

tw4452852:
Maybe you could use `httpserver.GetConfig` to get the config and then manipulate the site root by `config.Root`. If it is appointed to a non exist directory, you will get the expected failure.

For the reason your method doesn't trigger the failure is because test config's site root is the default one which is `.`

caddyhttp/browse/setup_test.go
```diff
@@ -66,4 +66,16 @@ func TestSetup(t *testing.T) {
 			}
 		}
 	}
+
+	// test case #6 tests startup with missing root directory in combination with default browse settings
+	controller := caddy.NewTestController("http", "browse")
+	cfg := httpserver.GetConfig(controller)
+
+	//Manipulate the root to a nonexistent directory, must be done, cause the default test root exists
+	cfg.Root = nonExistantDirPath
```
mholt: "nonExistentDirPath" (spelling)

caddyhttp/browse/setup_test.go
```diff
@@ -66,4 +66,16 @@ func TestSetup(t *testing.T) {
 			}
 		}
 	}
+
+	// test case #6 tests startup with missing root directory in combination with default browse settings
+	controller := caddy.NewTestController("http", "browse")
+	cfg := httpserver.GetConfig(controller)
+
+	//Manipulate the root to a nonexistent directory, must be done, cause the default test root exists
```
mholt: `go fmt` should put a space between `//M` so it becomes `// M`

caddyhttp/browse/setup_test.go
```diff
@@ -66,4 +66,16 @@ func TestSetup(t *testing.T) {
...
+	cfg.Root = nonExistantDirPath
+	err = setup(controller)
+
+	if err != nil {
+		t.Errorf("Test case #%d received an error of %v", 6, err)
```
mholt: Change to `"Test for non-existent browse path received an error, but shouldn't have: %v"`

caddyhttp/browse/setup_test.go
mholt: Could probably even simplify the comment to just: "Make sure non-existent root path doesn't return error"

mholt:
Excellent. Thank you!

PR #1155, opened 2016-10-01
Fix [1152]? Fix unexpected behavior in redir

### LINKED ISSUE #1152: Infinite Redirect with Caddy 0.9.3

Caddy 0.9.3

Serve https with automatic SSL

```text
https://sub.domain.com

log stdout
prometheus 0.0.0.0:9180

tls user@domain.com

proxy / upstream:80 {
  transparent
}
```

docker run -d --link upstream -p 0.0.0.0:80:80 -p 0.0.0.0:443:443 my-caddy-image

The content of https://sub.domain.com/

Infinite redirects from https://sub.domain.com/ to https://sub.domain.com/

Use this Dockerfile:
```text
FROM alpine:3.4
RUN apk --update --no-cache add \
  ca-certificates \
  curl \
  git \
  openssh-client \
  && rm -rf /var/cache/apk/*
RUN mkdir -p /bin /srv
RUN curl -fsSL "https://caddyserver.com/download/build?os=linux&arch=amd64&features=cors%2Cgit%2Chugo%2Cminify%2Cprometheus%2Cratelimit%2Crealip%2Csearch%2Ccloudflare%2Cgooglecloud%2Croute53" -o /tmp/caddy.tar.gz \
  && cd /tmp \
  && tar xvzf caddy.tar.gz \
  && mv /tmp/caddy /bin/caddy \
  && rm -rf /tmp/caddy.tar.gz
VOLUME [ "/srv" ]
WORKDIR /srv
ADD Caddyfile /Caddyfile
EXPOSE 80 443 9180
CMD [ "/bin/caddy", "-http2=false" ]
```
and the above Caddyfile, run:
```
docker build -t my-caddy-image .
docker run -d -p 0.0.0.0:80:80 -p 0.0.0.0:443:443 my-caddy-image
```

elcore:
Hello,

Caddyfile:

```
www.domain.com {
   redir https://domain.com{uri}
}
domain.com {
  ...
}
```

http://www.domain.com/abc/def redirects to https://domain.com//abc/def

This PR fixes this behavior => http://www.domain.com/abc/def redirects to https://domain.com/abc/def

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -245,9 +245,9 @@ func (r *replacer) getSubstitution(key string) string {
 		}
 		return port
 	case "{uri}":
-		return r.request.URL.RequestURI()
+		return strings.TrimLeft(r.request.URL.RequestURI(), "/")
 	case "{uri_escaped}":
-		return url.QueryEscape(r.request.URL.RequestURI())
+		return url.QueryEscape(strings.TrimLeft(r.request.URL.RequestURI(), "/"))
 	case "{when}":
 		return time.Now().Format(timeFormat)
 	case "{file}":
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -245,9 +245,9 @@ func (r *replacer) getSubstitution(key string) string {
 		}
 		return port
 	case "{uri}":
-		return strings.TrimLeft(r.request.URL.RequestURI(), "/")
+		return r.request.URL.RequestURI()
 	case "{uri_escaped}":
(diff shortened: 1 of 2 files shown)


mholt:
Thanks for looking into this!! :smile:

elcore:
> So, wait a second. What's causing the extra slash? And can you add a test case?

We already have a test (https://github.com/mholt/caddy/blob/master/caddyhttp/redirect/redirect_test.go#L91-L107), but it is not "waterproof"!!! -- I am not sure how we can "waterproof" it.

> Currently, this feels like a hack, singling out the special case of the redirect rule going to {uri} only... I think we need to understand the problem better first.

You are right!!! This is a "hack"

I will close this PR now -- Somebody with more experience in Golang needs to troubleshoot our issue.

> Thanks for looking into this!! :smile:

Thank you @mholt

Issue #1154, opened 2016-09-30
Proxy "without" option does not work if percent-encoded characters are in the URL


lluchs:
### 1. What version of Caddy are you running (`caddy -version`)?
Caddy 0.9.3

### 2. What are you trying to do?
Use the "without" option to "proxy".

### 3. What is your entire Caddyfile?
```text
:2015

log stdout

proxy /proxy http://localhost:2015 {
	without /proxy
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
No options, AUR package caddy-full-bin 0.9.3-8 on Arch Linux

### 5. What did you expect to see?
Caddy should proxy requests to /proxy/foobar to /foobar and requests to /proxy/foobar%3D to /foobar%3D.

### 6. What did you see instead (give full error messages and/or log)?
Caddy routes requests to /proxy/foobar correctly, but adding a percent-encoded character such as %3D causes Caddy to proxy e.g. /proxy/foobar%3D to /proxy/foobar%3D, i.e. the /proxy prefix is not removed.

Example log:
```
% caddy
Activating privacy features... done.
http://:2015
WARNING: File descriptor limit 1024 is too low for production servers. At least 8192 is recommended. Fix with "
ulimit -n 8192".
# In separate terminal: % curl http://localhost:2015/proxy/foobar
::1 - [30/Sep/2016:17:37:55 +0200] "GET /foobar HTTP/1.1" 404 14
::1 - [30/Sep/2016:17:37:55 +0200] "GET /foobar HTTP/1.1" 404 14
# In separate terminal: % curl http://localhost:2015/proxy/foobar%3D
::1 - [30/Sep/2016:17:38:12 +0200] "GET /proxy/foobar%3D HTTP/1.1" 502 16
::1 - [30/Sep/2016:17:38:12 +0200] "GET /proxy/foobar%3D HTTP/1.1" 502 16
::1 - [30/Sep/2016:17:38:12 +0200] "GET /proxy/foobar%3D HTTP/1.1" 502 16
::1 - [30/Sep/2016:17:38:12 +0200] "GET /proxy/foobar%3D HTTP/1.1" 502 16
::1 - [30/Sep/2016:17:38:12 +0200] "GET /proxy/foobar%3D HTTP/1.1" 502 16
[lots more]
```

### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?
Above is the minimal reproduction, real case is proxying to another server.

PR #1158, opened 2016-10-02
proxy: handle 'without' option in encoded form of URL path

tw4452852:
fix issue #1154


mholt:
Love the new tests :+1:

PR #1163, opened 2016-10-03
(WIP) Add maxrequestbody directive

Ngo-The-Trung:
@mholt I'm looking for opinions on this implementation before I proceed further

[Issue](https://github.com/mholt/caddy/issues/1027)

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -438,6 +438,7 @@ var directives = []string{
 	"mailout",   // github.com/SchumacherFM/mailout
 	"awslambda", // github.com/coopernurse/caddy-awslambda
 	"filter",    // github.com/echocat/caddy-filter
+	"maxrequestbody",
 }
 
 const (
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -393,6 +393,7 @@ var directives = []string{
 	// primitive actions that set up the fundamental vitals of each config
 	"root",
 	"bind",
+	"maxrequestbody",
 	"tls",
 
 	// services/utilities, or other directives that don't necessarily inject handlers
@@ -438,7 +439,6 @@ var directives = []string{
 	"mailout",   // github.com/SchumacherFM/mailout
 	"awslambda", // github.com/coopernurse/caddy-awslambda
 	"filter",    // github.com/echocat/caddy-filter
-	"maxrequestbody",
 }
 
 const (
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -4,6 +4,7 @@ package httpserver
 import (
 	"crypto/tls"
 	"fmt"
+	"io"
 	"log"
 	"net"
 	"net/http"
@@ -265,6 +266,19 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 		}
 	}
 
+	// Apply the path-based request body size limit
+	// The error returned by MaxBytesReader is meant to be handled
+	// by whichever middleware/plugin that receives it when calling
+	// .Read() or a similar method on the request body
+	if r.Body != nil {
+		for _, pathlimit := range vhost.MaxRequestBodySizes {
+			if Path(r.URL.Path).Matches(pathlimit.Path) {
+				r.Body = MaxBytesReader(w, r.Body, pathlimit.Limit)
+				break
+			}
+		}
+	}
+
 	return vhost.middlewareChain.ServeHTTP(w, r)
 }
 
@@ -391,6 +405,74 @@ func (ln tcpKeepAliveListener) Accept() (c net.Conn, err error) {
 // File implements caddy.Listener; it returns the underlying file of the listener.
 func (ln tcpKeepAliveListener) File() (*os.File, error) {
 	return ln.TCPListener.File()
+}
+
+// MaxBytesExceeded is the error type returned by MaxBytesReader
+// when the request body exceeds the limit imposed
+type MaxBytesExceeded struct{}
+
+func (err MaxBytesExceeded) Error() string {
+	return "http: request body too large"
+}
+
+// MaxBytesReader and its associated methods are borrowed from the
+// Go Standard library (comments intact). The only difference is that
+// it returns a MaxBytesExceeded error instead of a generic error message
+// when the request body has exceeded the requested limit
+func MaxBytesReader(w http.ResponseWriter, r io.ReadCloser, n int64) io.ReadCloser {
+	return &maxBytesReader{w: w, r: r, n: n}
+}
+
+type maxBytesReader struct {
+	w   http.ResponseWriter
+	r   io.ReadCloser // underlying reader
+	n   int64         // max bytes remaining
+	err error         // sticky error
+}
+
+func (l *maxBytesReader) Read(p []byte) (n int, err error) {
+	if l.err != nil {
+		return 0, l.err
+	}
+	if len(p) == 0 {
+		return 0, nil
+	}
+	// If they asked for a 32KB byte read but only 5 bytes are
+	// remaining, no need to read 32KB. 6 bytes will answer the
+	// question of the whether we hit the limit or go past it.
+	if int64(len(p)) > l.n+1 {
+		p = p[:l.n+1]
+	}
+	n, err = l.r.Read(p)
+
+	if int64(n) <= l.n {
+		l.n -= int64(n)
+		l.err = err
+		return n, err
+	}
+
+	n = int(l.n)
+	l.n = 0
+
+	// The server code and client code both use
+	// maxBytesReader. This "requestTooLarge" check is
+	// only used by the server code. To prevent binaries
+	// which only using the HTTP Client code (such as
+	// cmd/go) from also linking in the HTTP server, don't
+	// use a static type assertion to the server
+	// "*response" type. Check this interface instead:
+	type requestTooLarger interface {
+		requestTooLarge()
+	}
+	if res, ok := l.w.(requestTooLarger); ok {
+		res.requestTooLarge()
+	}
+	l.err = MaxBytesExceeded{}
+	return n, l.err
+}
+
+func (l *maxBytesReader) Close() error {
+	return l.r.Close()
 }
 
 // DefaultErrorFunc responds to an HTTP request with a simple description
--- a/caddyhttp/httpserver/siteconfig.go
+++ b/caddyhttp/httpserver/siteconfig.go
@@ -30,6 +30,16 @@ type SiteConfig struct {
 	// standardized way of loading files from disk
 	// for a request.
 	HiddenFiles []string
+
+	// Max amount of bytes a request can send on a given path
+	MaxRequestBodySizes []PathLimit
+}
+
+// PathLimit is a mapping from a site's path to its corresponding
+// maximum request body size (in bytes)
+type PathLimit struct {
+	Path  string
+	Limit int64
 }
 
 // AddMiddleware adds a middleware to a site's middleware stack.
--- a/caddyhttp/maxrequestbody/maxrequestbody.go
+++ b/caddyhttp/maxrequestbody/maxrequestbody.go
@@ -43,7 +43,7 @@ func setupMaxRequestBody(c *caddy.Controller) error {
 		// Format: { <path> <limit> ... }
 		for c.NextBlock() {
 			path := c.Val()
-			if !c.NextBlock() {
+			if !c.NextArg() {
 				// Uneven pairing of path/limit
 				return c.ArgErr()
 			}
--- a/caddyhttp/maxrequestbody/maxrequestbody.go
+++ b/caddyhttp/maxrequestbody/maxrequestbody.go
@@ -22,6 +22,12 @@ func init() {
 	})
 }
 
+// pathLimitUnparsed is a PathLimit before it's parsed
+type pathLimitUnparsed struct {
+	Path  string
+	Limit string
+}
+
 func setupMaxRequestBody(c *caddy.Controller) error {
 	config := httpserver.GetConfig(c)
 
@@ -30,7 +36,39 @@ func setupMaxRequestBody(c *caddy.Controller) error {
 	}
 
 	args := c.RemainingArgs()
-	pathLimit, err := parseArguments(args)
+	argList := []pathLimitUnparsed{}
+
+	switch len(args) {
+	case 0:
+		// Format: { <path> <limit> ... }
+		for c.NextBlock() {
+			path := c.Val()
+			if !c.NextBlock() {
+				// Uneven pairing of path/limit
+				return c.ArgErr()
+			}
+			argList = append(argList, pathLimitUnparsed{
+				Path:  path,
+				Limit: c.Val(),
+			})
+		}
+	case 1:
+		// Format: <limit>
+		argList = []pathLimitUnparsed{{
+			Path:  "/",
+			Limit: args[0],
+		}}
+	case 2:
+		// Format: <path> <limit>
+		argList = []pathLimitUnparsed{{
+			Path:  args[0],
+			Limit: args[1],
+		}}
+	default:
+		return c.ArgErr()
+	}
+
+	pathLimit, err := parseArguments(argList)
 	if err != nil {
 		return c.ArgErr()
 	}
@@ -42,26 +80,15 @@ func setupMaxRequestBody(c *caddy.Controller) error {
 	return nil
 }
 
-func parseArguments(args []string) ([]httpserver.PathLimit, error) {
+func parseArguments(args []pathLimitUnparsed) ([]httpserver.PathLimit, error) {
 	pathLimit := []httpserver.PathLimit{}
-	// Arguments appear in pairs of "<path> <limit>"
-	// If there's an argument that's not paired, it's assumed to be
-	// the limit used for all remaining paths ("/")
-	if len(args)%2 == 1 {
-		args = append([]string{"/"}, args...)
-	}
 
-	i := 0
-	for i < len(args) {
-		path := args[i]
-		limitStr := args[i+1]
-		i = i + 2
-
-		size := parseSize(limitStr)
+	for _, pair := range args {
+		size := parseSize(pair.Limit)
 		if size < 1 { // also disallow size = 0
 			return pathLimit, errors.New("Parse failed")
 		}
-		pathLimit = addPathLimit(pathLimit, path, size)
+		pathLimit = addPathLimit(pathLimit, pair.Path, size)
 	}
 	return pathLimit, nil
 }
--- a/caddyhttp/maxrequestbody/maxrequestbody.go
+++ b/caddyhttp/maxrequestbody/maxrequestbody.go
@@ -1,0 +1,151 @@
+package maxrequestbody
+
+import (
+	"errors"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/mholt/caddy"
+	"github.com/mholt/caddy/caddyhttp/httpserver"
+)
+
+const (
+	serverType = "http"
+	pluginName = "maxrequestbody"
+)
+
+func init() {
+	caddy.RegisterPlugin(pluginName, caddy.Plugin{
+		ServerType: serverType,
+		Action:     setupMaxRequestBody,
+	})
+}
+
+func setupMaxRequestBody(c *caddy.Controller) error {
+	config := httpserver.GetConfig(c)
+
+	if !c.Next() {
+		return c.ArgErr()
+	}
+
+	args := c.RemainingArgs()
+	pathLimit, err := parseArguments(args)
+	if err != nil {
+		return c.ArgErr()
+	}
+
+	SortPathLimits(pathLimit)
+
+	config.MaxRequestBodySizes = pathLimit
+
+	return nil
+}
+
+func parseArguments(args []string) ([]httpserver.PathLimit, error) {
+	pathLimit := []httpserver.PathLimit{}
+	// Arguments appear in pairs of "<path> <limit>"
+	// If there's an argument that's not paired, it's assumed to be
+	// the limit used for all remaining paths ("/")
+	if len(args)%2 == 1 {
+		args = append([]string{"/"}, args...)
+	}
+
+	i := 0
+	for i < len(args) {
+		path := args[i]
+		limitStr := args[i+1]
+		i = i + 2
+
+		size := parseSize(limitStr)
+		if size < 1 { // also disallow size = 0
+			return pathLimit, errors.New("Parse failed")
+		}
+		pathLimit = addPathLimit(pathLimit, path, size)
+	}
+	return pathLimit, nil
+}
+
+var validUnits = []struct {
+	symbol     string
+	multiplier int64
+}{
+	{"KB", 1024},
+	{"MB", 1024 * 1024},
+	{"GB", 1024 * 1024 * 1024},
+	{"B", 1},
+	{"", 1}, // defaulting to "B"
+}
+
+// parseSize parses the given string as size limit
+// Size are positive numbers followed by a unit (case insensitive)
+// Allowed units: "B" (bytes), "KB" (kilo), "MB" (mega), "GB" (giga)
+// If the unit is omitted, "b" is assumed
+// Returns the parsed size in bytes, or -1 if cannot parse
+func parseSize(sizeStr string) int64 {
+	sizeStr = strings.ToUpper(sizeStr)
+
+	for _, unit := range validUnits {
+		if strings.HasSuffix(sizeStr, unit.symbol) {
+			size, err := strconv.ParseInt(sizeStr[0:len(sizeStr)-len(unit.symbol)], 10, 64)
+			if err != nil {
+				return -1
+			}
+			return size * unit.multiplier
+		}
+	}
+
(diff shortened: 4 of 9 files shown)


caddyhttp/httpserver/siteconfig.go
```diff
@@ -30,6 +30,14 @@ type SiteConfig struct {
 	// standardized way of loading files from disk
 	// for a request.
 	HiddenFiles []string
+
+	// Max amount of bytes a request can send on a given path
```
mholt: "Max body size for requests to their respective paths"

caddyhttp/httpserver/siteconfig.go
```diff
@@ -30,6 +30,14 @@ type SiteConfig struct {
...
+	MaxRequestBodySizes []PathLimit
+}
+
+type PathLimit struct {
```
mholt: This'll need some godoc of course

caddyhttp/httpserver/server.go
```diff
@@ -406,3 +417,14 @@ func WriteTextResponse(w http.ResponseWriter, status int, body string) {
 	w.WriteHeader(status)
 	w.Write([]byte(body))
 }
+
+// HandleCommonError checks if the given error is an error that could pop up
+// when handling request. Currently only checks for MaxBytesReader error.
+// Returns a boolean - whether the given error is a common error, and a int -
+// the suggested HTTP status code to return to the user
+func IsCommonError(err error) (bool, int) {
+	if err.Error() == "http: request body too large" {
```
mholt: Reading the string content of an error value is a little scary -- I like to avoid this where possible. It could change any time.

After looking into MaxBytesReader, I see it [makes use of an unexported error value](https://github.com/golang/go/blob/89c4cbd7acc36cf627746f379f0cc002b5b60383/src/net/http/request.go#L982), hence the need for this function which has to read the error string...

Finding https://github.com/golang/go/issues/8479, it seems like the Go team isn't interested in having the error value exported and reused, but the implementation of MaxBytesReader is trivial; we could just borrow from it and change it to meet our needs by using a pre-defined error value. Also it would be useful to mention how big the read body was, just so it's easier to read in the logs.

However, we should decide if we want to centralize the use of a MaxBodyReader like this, or if each middleware that reads the response body should use it instead. Either way, they'll have to check for the kind of error value returned and, if an EntityTooLargeError, they return a special status code in that case.

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"regexp"
+	"strconv"
+
+	"github.com/mholt/caddy"
+	"github.com/mholt/caddy/caddyhttp/httpserver"
+)
+
+var SIZE_RE = regexp.MustCompile("[1-9][0-9]*[kmKM]?")
```
mholt: To prevent maintenance headaches in the future, can we not use a regular expression? :) This is a simple job for parsing. If it's a positive integer, use it as bytes. If it's a string suffixed with any case variation of KB, MB, (or GB, I suppose), split it out and parse the non-negative number before it.

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,59 @@
...
+		mult = 1024
+		arg = arg[0:L]
+	} else if arg[L] == 'M' {
+		mult = 1024 * 1024
+		arg = arg[0:L]
+	}
+	if size, err := strconv.ParseInt(arg, 10, 64); err != nil {
+		return c.ArgErr()
+	} else {
+		config.MaxRequestBodySizes = append(config.MaxRequestBodySizes, httpserver.PathLimit{path, size * mult})
```
tw4452852: Should we check the possible path duplication?

caddyhttp/httpserver/server.go
```diff
@@ -265,6 +265,17 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 		}
 	}
 
+	// Apply the path-based request body size limit
+	// The error returned by MaxBytesReader is meant to be handled
+	// by whichever middleware/plugin that receives it when calling
+	// .Read() or a similar method on the request body
+	for _, pathlimit := range vhost.MaxRequestBodySizes {
+		if Path(r.URL.Path).Matches(pathlimit.Path) {
+			r.Body = http.MaxBytesReader(w, r.Body, pathlimit.Limit)
```
tw4452852: `r.Body` may be `nil`.

caddyhttp/httpserver/server.go
tw4452852: Agree with @mholt , Use the type assertion instead.

caddyhttp/httpserver/server.go
mholt: I've heard this debated before, and so I'm not sure if that's true -- but I definitely would not oppose to a nil check here!

caddyhttp/maxrequestbody/maxrequestbody.go
Ngo-The-Trung: I've been meaning to ask this - do we:
- clean up the path (add slashes)
- sort the paths (so longer/more specific paths are checked first)

caddyhttp/maxrequestbody/maxrequestbody.go
Ngo-The-Trung: Okay

caddyhttp/maxrequestbody/maxrequestbody.go
tw4452852: I think we could clean up the path here and if the path exists, overwrite.
Also sorting paths for specific path matching above makes sense.

mholt: Hmm, I thought there was a facility in the `httpserver` package to help with the longest-path-matching thing, but maybe it was only the `proxy` middleware that got that blessed feature. (Man, I still really think we need to totally standardize the code to match requests. But that's a different topic.) Check out the proxy middleware maybe...

caddyhttp/httpserver/plugin.go
```diff
@@ -437,6 +437,7 @@ var directives = []string{
 	"hugo",        // github.com/hacdias/caddy-hugo
 	"mailout",     // github.com/SchumacherFM/mailout
 	"awslambda",   // github.com/coopernurse/caddy-awslambda
+	"maxrequestbody",
```
mholt: Also I wonder if it should be called `max_req_body` or something a little shorter (and readable).

Ngo-The-Trung: Doesn't seem like anything other directives use underscore at the moment.

Also I will leave renaming last.

Ngo-The-Trung:
Updated. Addressed the comments above (except for the directive name, that comes later)

- Copied MaxBytesReader from the Go std here and use type checking for error
- Updated parsing to allow multiple paths (such as `maxrequestbody / 100MB /index 200GB`)
- Allow sizes of B, KB, MB, GB (optional B, case insensitive) 
- Parse sorts by path lengths
- Add some parsing tests

I don't understand the CI failures.

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,83 @@
...
+	for caseNum, c := range cases {
+		output, err := parseArguments(c.arguments)
+		if c.hasError && (err == nil) {
+			t.Errorf("Expecting error for case %v but none encountered", caseNum)
+		}
+		if !c.hasError && (err != nil) {
+			t.Errorf("Expecting no error for case %v but encountered %v", caseNum, err)
+		}
+
+		if len(c.expected) != len(output) {
```
tw4452852: Maybe `reflect.DeepEqual` helps.

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,83 @@
...
+		// Parse errors
+		{[]PathLimit{}, []PathLimit{}},
+		{[]PathLimit{{"/index", 100}}, []PathLimit{{"/index", 100}}},
+		{[]PathLimit{{"/static", 1}, {"/static/images", 100}, {"/index", 200}}, []PathLimit{{"/static/images", 100}, {"/static", 1}, {"/index", 200}}},
+	}
+
+	for caseNum, c := range cases {
+		output := append([]PathLimit{}, c.arguments...)
+		SortPathLimits(output)
+		if len(c.expected) != len(output) {
```
tw4452852: Ditto.

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,83 @@
+package maxrequestbody
+
+import (
+	"testing"
+
+	. "github.com/mholt/caddy/caddyhttp/httpserver"
```
tw4452852: Explicit import may be better.

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const (
```
tw4452852: Could these be combined with `validUnits` in `maxrequestbody.go` ?

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,152 @@
...
+	args := c.RemainingArgs()
+	pathLimit, err := parseArguments(args)
+	if err != nil {
+		return c.ArgErr()
+	}
+
+	SortPathLimits(pathLimit)
+
+	config.MaxRequestBodySizes = pathLimit
+	fmt.Println(pathLimit)
```
tw4452852: Debug code?

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,152 @@
...
+
+	return nil
+}
+
+func parseArguments(args []string) ([]httpserver.PathLimit, error) {
+	pathLimit := []httpserver.PathLimit{}
+	// Arguments appear in pairs of "<path> <limit>"
+	// If there's an argument that's not paired, it's assumed to be
+	// the limit used for all remaining paths ("/")
+	if len(args)%2 == 1 {
```
tw4452852: This could be wrong. Suppose we have this sequence: `/a 100KB 200KB`  or `/a 100KB /b`

caddyhttp/httpserver/server.go
```diff
@@ -406,3 +488,14 @@ func WriteTextResponse(w http.ResponseWriter, status int, body string) {
...
+
+// HandleCommonError checks if the given error is an error that could pop up
+// when handling request. Currently only checks for MaxBytesReader error.
+// Returns a boolean - whether the given error is a common error, and a int -
+// the suggested HTTP status code to return to the user
+func IsCommonError(err error) (bool, int) {
+	if _, ok := err.(MaxBytesExceeded); ok {
+		return true, http.StatusRequestEntityTooLarge
+	}
+	return false, -1
```
tw4452852: `-1` maybe a invalid HTTP status.

caddyhttp/httpserver/server.go
```diff
@@ -392,6 +406,74 @@ func (ln tcpKeepAliveListener) File() (*os.File, error) {
 	return ln.TCPListener.File()
 }
 
+// MaxBytesExceeded is the error type returned by MaxBytesReader
+// when the request body exceeds the limit imposed
+type MaxBytesExceeded int
```
tw4452852: Is there any reason for using a `int` here ?

tw4452852:
@Ngo-The-Trung 
Also for CI issue, you should check `TestStandardPlugins`, I think you will find the answer.

caddyhttp/httpserver/server.go
Ngo-The-Trung: No. Could be anything. Should I use `interface{}` instead?

caddyhttp/maxrequestbody/maxrequestbody.go
Ngo-The-Trung: Forgot to remove

caddyhttp/maxrequestbody/maxrequestbody.go
Ngo-The-Trung: Well I think everything should have a pair, with the only exception being the first argument.

The 2 cases you mentioned would fail due to the ParseInt below, also I have no idea how to interpret them.

caddyhttp/maxrequestbody/maxrequestbody_test.go
Ngo-The-Trung: Not easy. 100 * KB is more readable than say 100 * validUnits["KB"] or 100 * validUnits[1], depending on whether I'm making it a map or a list.

Also I traverse the units in order in the parser code so if I were to rewrite validUnits to a map I would need to make an additional variable to store the order of the keys.

I would keep the constants.

caddyhttp/maxrequestbody/maxrequestbody.go
tw4452852: Hmm, Is there any reason for this exception?

caddyhttp/httpserver/server.go
tw4452852: Maybe `struct{}` better.

caddyhttp/httpserver/server.go
Ngo-The-Trung: My intention is that the status code is advisory only, and is only applicable when the returned bool is `true`. It should be discarded if it is `false`.

caddyhttp/maxrequestbody/maxrequestbody.go
Ngo-The-Trung: For convenience

`maxrequestbody 200MB` is understood as "limit everything (/) to 200MB"

`maxrequestbody 200MB /upload 500GB` is understood as "limit upload to 500GB, and limit everythign else to 200MB"

caddyhttp/maxrequestbody/maxrequestbody_test.go
Ngo-The-Trung: Edit: I meant to reply to the `validUnits` comment

caddyhttp/httpserver/server.go
tw4452852: Make it clear in comment.

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,150 @@
...
+	for i < len(args) {
+		path := args[i]
+		limitStr := args[i+1]
+		i = i + 2
+
+		if size := parseSize(limitStr); size < 1 { // also disallow size = 0
+			return pathLimit, errors.New("Parse failed")
+		} else {
+			pathLimit = addPathLimit(pathLimit, path, size)
+		}
```
elcore: You do not need to use else here

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,150 @@
...
+	sizeStr = strings.ToUpper(sizeStr)
+
+	for _, unit := range validUnits {
+		if strings.HasSuffix(sizeStr, unit.symbol) {
+
+			if size, err := strconv.ParseInt(sizeStr[0:len(sizeStr)-len(unit.symbol)], 10, 64); err != nil {
+				return -1
+			} else {
+				return size * unit.multiplier
+			}
```
elcore: You do not need to use else here

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (s *pathLimitSorter) Swap(i, j int) {
+	s.pathLimits[i], s.pathLimits[j] = s.pathLimits[j], s.pathLimits[i]
+}
+
+func (s *pathLimitSorter) Less(i, j int) bool {
+	return s.by(&s.pathLimits[i], &s.pathLimits[j])
+}
+
+func LengthDescending(p1, p2 *httpserver.PathLimit) bool {
```
elcore: We need some godoc here

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,118 @@
...
+			},
+			hasError: false},
+		{
+			arguments: []string{"999", "/home", "12345MB"},
+			expected: []httpserver.PathLimit{
+				httpserver.PathLimit{Path: "/", Limit: 999},
+				httpserver.PathLimit{Path: "/home", Limit: 12345 * MB},
+			},
+			hasError: false,
+		},
```
elcore: Please use gofmt

caddyhttp/httpserver/siteconfig.go
elcore: Yes, please add some godoc.

caddyhttp/maxrequestbody/maxrequestbody_test.go
Ngo-The-Trung: My gofmt seems to be ok with this for some reason

Ngo-The-Trung: Oh gofmt wasn't the reason. You guys are using simplify mode (-s).

I'll update this.

Ngo-The-Trung:
Well at least the CI is green now

caddyhttp/maxrequestbody/maxrequestbody_test.go
elcore: Well, it´s part of gofmt 😉 -- Our code should be simple 😄

elcore:
> Well at least the CI is green now

@Ngo-The-Trung mission accomplished 😄

Ngo-The-Trung:
Pinging @mholt @tw4452852 for review 

Not sure how to test the limiting itself.

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,150 @@
...
+
+// addPathLimit appends the path-to-request body limit mapping to pathLimit
+// Slashes are checked and added to path if necessary. Duplicates are ignored.
+func addPathLimit(pathLimit []httpserver.PathLimit, path string, limit int64) []httpserver.PathLimit {
+	// Enforces preceding slash
+	if path[0] != '/' {
+		path = "/" + path
+	}
+
+	// Avoid duplicates
```
tw4452852: I think it's better to overwrite the exist one.

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for caseNum, c := range cases {
+		output, err := parseArguments(c.arguments)
+		if c.hasError && (err == nil) {
+			t.Errorf("Expecting error for case %v but none encountered", caseNum)
+		}
+		if !c.hasError && (err != nil) {
+			t.Errorf("Expecting no error for case %v but encountered %v", caseNum, err)
+		}
+
+		if reflect.DeepEqual(c.expected, len(output)) {
```
tw4452852: Should be: `!reflect.DeepEqual(c.expected, output)`

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,118 @@
...
+				{Path: "/static/images", Limit: 100},
+				{Path: "/static", Limit: 1},
+				{Path: "/index", Limit: 200}},
+		},
+	}
+
+	for caseNum, c := range cases {
+		output := append([]httpserver.PathLimit{}, c.arguments...)
+		SortPathLimits(output)
+		if reflect.DeepEqual(c.expected, len(output)) {
```
tw4452852: Ditto.

caddyhttp/maxrequestbody/maxrequestbody_test.go
Ngo-The-Trung: Oh lol the 2 wrongs make it right

Ngo-The-Trung:
Updated

mholt:
Making progress! :) I'll look at it more when I'm back from my trip.

caddyhttp/httpserver/server.go
```diff
@@ -406,3 +488,15 @@ func WriteTextResponse(w http.ResponseWriter, status int, body string) {
 	w.WriteHeader(status)
 	w.Write([]byte(body))
 }
+
+// IsCommonError checks if the given error is an error that could pop up
+// when handling request. Currently only checks for MaxBytesReader error.
+// Returns a boolean - whether the given error is a common error, and an int -
+// the advisory HTTP status code to return to the user (or -1 if the given error
+// is not a common error)
+func IsCommonError(err error) (bool, int) {
```
mholt: I still don't quite feel good about this function. It's only used by MaxBytesReader, I think error handling should not be here. It should probably remain in the handlers that use the MaxBytesReader.

Ngo-The-Trung: Ok

Ngo-The-Trung:
Updated

caddyhttp/maxrequestbody/maxrequestbody.go
mholt: I think this is a little confusing... A more Caddy-like way of doing pairs would be a "table" format. How about these valid forms instead:

```
maxrequestbody 200MB
maxrequestbody /upload 500GB
maxrequestbody {
    /ten 10MB
    /twenty 20MB
}
```

I think this way is much clearer.

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,151 @@
...
+		path = "/" + path
+	}
+
+	// Use the last value if there are duplicates
+	for i, p := range pathLimit {
+		if p.Path == path {
+			pathLimit[i].Limit = limit
+			return pathLimit
+		}
+	}
```
mholt: Nice, glad you're thinking of stuff like this. :+1:

caddyhttp/proxy/proxy.go
```diff
@@ -174,6 +174,10 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 			return 0, nil
 		}
 
+		if _, ok := backendEerr.(httpserver.MaxBytesExceeded); ok {
+			return http.StatusRequestEntityTooLarge, backendErr
+		}
```
mholt: _NICE._ This is what I was looking for. :tada:

mholt:
Cool, I'm almost done with a final review -- looking pretty good so far (I might massage a few things after merging, no big deal). Just need a few more minutes. :) Perhaps meanwhile @tw4452852 and @elcore would be able to update their reviews or approve the changes.

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/mholt/caddy/caddyhttp/httpserver"
+)
+
+const (
+	KB = 1024
+	MB = 1024 * 1024
+	GB = 1024 * 1024 * 1024
+)
+
```
tw4452852: Add `TestSetupMaxRequestBody` ? (just test the all kinds of syntax formats)

caddyhttp/maxrequestbody/maxrequestbody_test.go
```diff
@@ -0,0 +1,127 @@
...
+		arguments []pathLimitUnparsed
+		expected  []httpserver.PathLimit
+		hasError  bool
+	}{
+		// Parse errors
+		{arguments: []pathLimitUnparsed{{"/", "123.5"}}, expected: []httpserver.PathLimit{}, hasError: true},
+		{arguments: []pathLimitUnparsed{{"/", "200LB"}}, expected: []httpserver.PathLimit{}, hasError: true},
+		{arguments: []pathLimitUnparsed{{"/", "path:999MB"}}, expected: []httpserver.PathLimit{}, hasError: true},
+		{arguments: []pathLimitUnparsed{{"/", "1_234_567"}}, expected: []httpserver.PathLimit{}, hasError: true},
+
```
tw4452852: Add test with zero size.

Ngo-The-Trung:
@tw4452852 done

caddyhttp/maxrequestbody/maxrequestbody.go
```diff
@@ -0,0 +1,178 @@
...
+
+	args := c.RemainingArgs()
+	argList := []pathLimitUnparsed{}
+
+	switch len(args) {
+	case 0:
+		// Format: { <path> <limit> ... }
+		for c.NextBlock() {
+			path := c.Val()
+			if !c.NextBlock() {
```
tw4452852: I think we should use `NextArg` here.

mholt:
@tw4452852 When you are satisfied, approve and merge away!

caddyhttp/maxrequestbody/maxrequestbody.go
Ngo-The-Trung: Done

Ngo-The-Trung:
@mholt moved to below "bind"

tw4452852:
LGTM. Thanks @Ngo-The-Trung !

mholt:
@Ngo-The-Trung Thank you for your work on this!

Would you like to be a collaborator? I'll send you the invite and you may accept it if you want. This means you'd be able to contribute to Caddy without having to fork the project (you could commit to branches) and you can help by reviewing PRs, etc. Really appreciate your efforts. :)

PR #1167, opened 2016-10-04
Added directive "filter" to whitelist for support of caddy-filter plugin

blaubaer:
All required pull request to enable full support of github.com/echocat/caddy-filter plugin:
* https://github.com/mholt/caddy/pull/1167
* https://github.com/caddyserver/buildsrv/pull/33
* https://github.com/caddyserver/caddyserver.com/pull/84


mholt:
Thanks for writing this plugin @blaubaer - looks cool!

I have a couple recommendations before we make this available, if you don't mind.

- The `filter` syntax has nested blocks (with `rule`) and I think we should try to avoid that (I will probably remove support for nested blocks in the parser in the future). Instead, the user can use `filter` multiple times and specify one rule per `filter` directive.

- Make `maximumBufferSize` consistent and maybe shorter, like `max_buffer_size` instead.

- Tests should be runnable without Java installed. When the new build server is finished, it will run `go test` to test your plugin in an automated way, and we won't have Java installed. (And will the tests even run on Linux? I see a .bat file there.)

blaubaer:
Hi @mholt!

> The ``filter`` syntax has nested blocks (with ``rule``) and I think we should try to avoid that (I will probably remove support for nested blocks in the parser in the future). Instead, the user can use ``filter`` multiple times and specify one rule per ``filter`` directive.

The plugin does already also support this way. I changed the documentation for this for caddyserver.com and my project README.md.
 
> Make ``maximumBufferSize`` consistent and maybe shorter, like ``max_buffer_size`` instead.

Thank you for hint. Done! 😄

> Tests should be runnable without Java installed. When the new build server is finished, it will run ``go test`` to test your plugin in an automated way, and we won't have Java installed. (And will the tests even run on Linux? I see a .bat file there.)

This project also runs perfectly with a native ``go test`` only. I added this hint to my project README.md. Gradle in this case just helps me with the rest. For example test in a temporary directory, coverage upload etc.

mholt:
I'll check this out again when I have a chance, but I'm traveling for a few days.

PR #1169, opened 2016-10-04
Adding Go Report Card badge

bengadbois:
Not sure if this is useful, but Caddy is (currently) an A+, seems nice to show it off 😉


mholt:
Oh neat, thanks! Can you add `?style=flat-square` to the image URL so it's consistent with the others?

bengadbois:
Updated to flat style!

Note that it's not an official shields.io badge (unlike the rest of the Caddy badges), but otherwise matches the rest

Issue #1173, opened 2016-10-08
Websocket problem after upgrading from 0.9.2 to 0.9.3


jbub:
### 1. What version of Caddy are you running (`caddy -version`)?

0.9.3

### 2. What are you trying to do?

I am trying to proxy to `drone` server (https://github.com/drone/drone). 

### 3. What is your entire Caddyfile?
```text
https://xx.yy.zz {
    tls {
        dns digitalocean
    }
    proxy / localhost:8000 {
        websocket
        header_upstream Host {host}
        header_upstream X-Real-IP {remote}
        header_upstream X-Forwarded-For {remote}
        header_upstream X-Forwarded-Proto {scheme}
    }
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

Ubuntu 14.04 with following command:

```DO_AUTH_TOKEN=xx /usr/local/bin/caddy -agree=true -http2=false -conf=/etc/caddy/Caddyfile -pidfile=/var/run/caddy.pid -log=/var/log/caddy/caddy.log```

### 5. What did you expect to see?

Working websocket connections, everything was working on 0.9.2 with no config changes.

### 6. What did you see instead (give full error messages and/or log)?

For websockets connection i get errors like:

```WebSocket connection to 'wss://xx.yy.zz/ws/feed' failed: Error during WebSocket handshake: Unexpected response code: 200```

### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

The hard way:

1. Setup drone server: http://readme.drone.io/0.5/installation/server/
2. Setup caddy proxy using provided config.

mholt:
Hey @jbub, thanks for the report. Do you have the ability to build from source? If so, can you try building it at a661007a559e23aaef7952b9a87fb1608da127ad ? (I'm traveling so a little too busy to install drone, etc. at the moment.) Or try bisecting until you find the commit that broke it? Apparently we need more tests...

jbub:
Hello @mholt, just tried with a661007 and its working. I also tried bisecting and the last working commit is 617012c3fb701502f98ad9b7c27617a94560c581, original behaviour was broken by d0ddfc849df4c7e50487a8d0f1c2cac6d057be3f. Thank you.

mholt:
Oops. @tw4452852, apparently that change broke something with websockets. I won't have a chance to check it out for a while, if you wanted to take a look maybe.

tw4452852:
Sure, I will check it. Sorry for regression.

[...]

PR #1174, opened 2016-10-10
header: implement http.Hijacker for responseWriterWrapper

tw4452852:
fix issue #1173

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -21,6 +21,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
 var bufferPool = sync.Pool{New: createBuffer}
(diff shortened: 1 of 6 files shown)


caddyhttp/proxy/reverseproxy.go
```diff
@@ -189,7 +189,7 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request,
 		res.Body.Close()
 		hj, ok := rw.(http.Hijacker)
 		if !ok {
-			return nil
+			panic("not a hijacker")
```
mholt: Hmm, maybe we should just return an error here instead. And if we can be a little more specific with the message, that would be good: describe what specifically is not a hijacker.

tw4452852: Well, I think it should panic here because it's a programmer's error instead of a common user error and this panic should be captured by [here](https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/server.go#L205).

Also if it returns specific error here, proxy's `ServeHTTP` will retry some times.
Although we could make a exception for this specific error, but I think it adds some verbosity.

mholt: Okay, I can understand that. Thanks for explaining!

Issue #1177, opened 2016-10-11
Reverse proxy removes trailing slash


nicolasazrak:
### 1. What version of Caddy are you running (`caddy -version`)?

Compiled from master branch, not yet released

### 2. What are you trying to do?

Use reverse proxy and url ending with `/`

### 3. What is your entire Caddyfile?
```text
caddy.test {
        tls off
        proxy / localhost:8000
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

./caddy -conf Caddyfile

### 5. What did you expect to see?

When I do a request to `/apath/` the backend server receives the same path.

### 6. What did you see instead (give full error messages and/or log)?

The backend server receives a request to `/apath` instead of `/apath/`

### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

Using the above caddyfile, starting a python server and navigating to a folder. Searching in the code I've notice that the bug comes from here https://github.com/mholt/caddy/blob/master/caddyhttp/proxy/reverseproxy.go#L99 and it was changed to solve #1154 The problem is that using `path.Join`. This is problematic in some servers that redirects to add trailing slash, it causes an infinite redirect.

tw4452852:
Ah,  thanks for investigation @nicolasazrak . Could you have a try again to confirm?

nicolasazrak:
Thanks @tw4452852  now it's working again

PR #1178, opened 2016-10-11
proxy: preserve path trailing slash if it was there

tw4452852:
fix issue #1177


mholt:
Thank you! This is good.

Issue #1192, opened 2016-10-19
[Feature] Make fastcgi load balanceable too


HazCod:
It would be nice if we could support load-balanced fastcgi backend too. (e.g. multiple PHP-FPM backends with a singly caddy server)

wendigo:
I've attempted to implement it and waiting for comments right now

wendigo:
Closing this issue as https://github.com/mholt/caddy/pull/1264 was merged

PR #1196, opened 2016-10-20
Make fastcgi load balanceable too

wendigo:
My attempt to address #1192 😃 #hacktoberfest

This change extends fastcgi directive syntax to: 

```
fastcgi backend1 backend2 backend3... {
    preset php
}
```

and **is not backward compatible** with previous syntax.

### DIFF
--- a/caddyhttp/caddyhttp.go
+++ b/caddyhttp/caddyhttp.go
@@ -17,6 +17,7 @@ import (
 	_ "github.com/mholt/caddy/caddyhttp/internalsrv"
 	_ "github.com/mholt/caddy/caddyhttp/log"
 	_ "github.com/mholt/caddy/caddyhttp/markdown"
+	_ "github.com/mholt/caddy/caddyhttp/maxrequestbody"
 	_ "github.com/mholt/caddy/caddyhttp/mime"
 	_ "github.com/mholt/caddy/caddyhttp/pprof"
 	_ "github.com/mholt/caddy/caddyhttp/proxy"
--- a/caddyhttp/caddyhttp_test.go
+++ b/caddyhttp/caddyhttp_test.go
@@ -11,7 +11,7 @@ import (
 // ensure that the standard plugins are in fact plugged in
 // and registered properly; this is a quick/naive way to do it.
 func TestStandardPlugins(t *testing.T) {
-	numStandardPlugins := 27 // importing caddyhttp plugs in this many plugins
+	numStandardPlugins := 28 // importing caddyhttp plugs in this many plugins
 	s := caddy.DescribePlugins()
 	if got, want := strings.Count(s, "\n"), numStandardPlugins+5; got != want {
 		t.Errorf("Expected all standard plugins to be plugged in, got:\n%s", s)
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -393,6 +393,7 @@ var directives = []string{
 	// primitive actions that set up the fundamental vitals of each config
 	"root",
 	"bind",
+	"maxrequestbody",
 	"tls",
 
 	// services/utilities, or other directives that don't necessarily inject handlers
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -268,7 +268,9 @@ func (r *replacer) getSubstitution(key string) string {
 		}
 		_, err := ioutil.ReadAll(r.request.Body)
 		if err != nil {
-			return r.emptyValue
+			if _, ok := err.(MaxBytesExceeded); ok {
+				return r.emptyValue
+			}
 		}
 		return requestReplacer.Replace(r.requestBody.String())
 	case "{status}":
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -4,6 +4,7 @@ package httpserver
 import (
 	"crypto/tls"
 	"fmt"
+	"io"
 	"log"
 	"net"
 	"net/http"
@@ -265,6 +266,19 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 		}
 	}
 
+	// Apply the path-based request body size limit
+	// The error returned by MaxBytesReader is meant to be handled
+	// by whichever middleware/plugin that receives it when calling
+	// .Read() or a similar method on the request body
+	if r.Body != nil {
+		for _, pathlimit := range vhost.MaxRequestBodySizes {
+			if Path(r.URL.Path).Matches(pathlimit.Path) {
+				r.Body = MaxBytesReader(w, r.Body, pathlimit.Limit)
+				break
+			}
+		}
+	}
+
 	return vhost.middlewareChain.ServeHTTP(w, r)
 }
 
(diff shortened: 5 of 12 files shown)


caddyhttp/fastcgi/setup_test.go
```diff
@@ -207,3 +217,32 @@ func TestFastcgiParse(t *testing.T) {
 	}
 
 }
+
+func areDialersEqual(current, expected dialer, t *testing.T) bool {
+	var equal = true
```
tw4452852: remove this line.

caddyhttp/fastcgi/setup_test.go
```diff
@@ -207,3 +217,32 @@ func TestFastcgiParse(t *testing.T) {
...
+
+	switch actual := current.(type) {
+	case *loadBalancingDialer:
+		if expected, ok := expected.(*loadBalancingDialer); ok {
+			for i := 0; i < len(actual.dialers); i++ {
+				if !areDialersEqual(actual.dialers[i], expected.dialers[i], t) {
+					return false
+				}
+			}
```
tw4452852: add `return true`

caddyhttp/fastcgi/setup_test.go
```diff
@@ -207,3 +217,32 @@ func TestFastcgiParse(t *testing.T) {
...
+		} else {
+			equal = false
```
tw4452852: change to `return false`

caddyhttp/fastcgi/setup_test.go
```diff
@@ -207,3 +217,32 @@ func TestFastcgiParse(t *testing.T) {
...
+		}
+	case basicDialer:
+		equal = current == expected
```
tw4452852: change to `return current == expected`

caddyhttp/fastcgi/setup_test.go
```diff
@@ -207,3 +217,32 @@ func TestFastcgiParse(t *testing.T) {
...
+	case *persistentDialer:
+		if expected, ok := expected.(*persistentDialer); ok {
+			equal = actual.Equals(expected)
```
tw4452852: change to `return actual.Equals(expected)`

caddyhttp/fastcgi/setup_test.go
```diff
@@ -207,3 +217,32 @@ func TestFastcgiParse(t *testing.T) {
...
+		} else {
+			equal = false
```
tw4452852: change to `return false`

caddyhttp/fastcgi/setup_test.go
```diff
@@ -207,3 +217,32 @@ func TestFastcgiParse(t *testing.T) {
...
+		}
+	default:
+		t.Errorf("Unknown dialer type %T", current)
+	}
+
+	return equal
```
tw4452852: change to `return false`

caddyhttp/fastcgi/setup.go
```diff
@@ -55,26 +56,21 @@ func fastcgiParse(c *caddy.Controller) ([]Rule, error) {
...
 
-		network, address := parseAddress(rule.Address)
-		rule.dialer = basicDialer{network: network, address: address}
+		rule.Path = args[0]
+		lastIndex := len(args)
+
+		// Last argument is preset
+		if err := fastcgiPreset(args[len(args)-1], &rule); err == nil {
+			lastIndex--
+		}
```
tw4452852: if `err != nil` ?

caddyhttp/fastcgi/dialer_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// load balancing starts from index 1
+	if mockDialer2.client != client {
+		t.Errorf("Expected Close() to be called on referenced dialer")
+	}
+}
+
+type mockDialer struct {
+	dialCalled  int
+	closeCalled int
```
tw4452852: I don't see `closeCalled` used in test.

caddyhttp/fastcgi/dialer_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	dialer := &loadBalancingDialer{dialers: []dialer{mockDialer1, mockDialer2}}
+	client, _ := dialer.Dial()
+
+	// when
+	err := dialer.Close(client)
+
+	// then
+	if err != nil {
+		t.Error("Expected error not to occur")
```
tw4452852: should be `"Expected error to be nil"`

caddyhttp/fastcgi/setup_test.go
wendigo: ok

caddyhttp/fastcgi/setup_test.go
wendigo: ok

caddyhttp/fastcgi/setup_test.go
wendigo: ok

caddyhttp/fastcgi/setup_test.go
wendigo: ok

caddyhttp/fastcgi/setup.go
wendigo: Nope. If it's preset - than don't treat last index as fastcgi backend address.

So: ```fastcgi address1 address2 address3``` will yield rule with loadBalancingDialer and 3 dialers,
And ```fastcgi address1 address2 php``` will yield rule with loadBalancingDialer and 2 dialers and PHP preset loaded

caddyhttp/fastcgi/dialer_test.go
wendigo: Added to test

caddyhttp/fastcgi/dialer_test.go
wendigo: Ok

mholt:
Wow, I'll take a detailed look at this as soon as I can! (It's midterm season...)

wendigo:
Good luck then @mholt :)

caddyhttp/fastcgi/setup.go
tw4452852: But how do you handle this: `fastcgi address1 address2 invalid` ?

caddyhttp/fastcgi/setup_test.go
```diff
@@ -230,19 +229,19 @@ func areDialersEqual(current, expected dialer, t *testing.T) bool {
 				}
```
tw4452852: Just `return true` here and then you could remove the following `else`

caddyhttp/fastcgi/setup_test.go
```diff
@@ -230,19 +229,19 @@ func areDialersEqual(current, expected dialer, t *testing.T) bool {
...
+			return actual.Equals(expected)
 		}
+		return false
+
 	default:
 		t.Errorf("Unknown dialer type %T", current)
 	}
 
-	return equal
+	return true
```
tw4452852: should be `return false`

caddyhttp/fastcgi/dialer_test.go
```diff
@@ -15,17 +15,22 @@ func TestLoadbalancingDialer(t *testing.T) {
...
 		}
 	}
 
 	// then
 	if mockDialer1.dialCalled != mockDialer2.dialCalled && mockDialer1.dialCalled != 50 {
 		t.Errorf("Expected dialer to load balance to multiple backend dialers %d times [actual: %d, %d]", 50, mockDialer1.dialCalled, mockDialer2.dialCalled)
 	}
+
+	if mockDialer1.closeCalled != mockDialer2.closeCalled && mockDialer1.closeCalled != 50 {
+		t.Errorf("Expected dialer to load balance to multiple backend dialers %d times [actual: %d, %d]", 50, mockDialer1.closeCalled, mockDialer2.closeCalled)
```
tw4452852: Amend this error message, shouldn't be same as above.

caddyhttp/fastcgi/setup.go
wendigo: I check if the last argument is a valid preset. If it's not - it is treated as an address.

caddyhttp/fastcgi/setup_test.go
wendigo: ok

caddyhttp/fastcgi/setup_test.go
wendigo: ok

caddyhttp/fastcgi/dialer_test.go
wendigo: ok

wendigo:
@tw4452852: is this code ok for you now?

caddyhttp/fastcgi/setup.go
tw4452852: I mean when you handle `fastcgi address1 address2 invalid`, you should return the invalid preset error instead of others, like `invalid address`. Also if `parseAddress` encounters `invalid`, it will return `"tcp", "invalid` which isn't what we expect. I think we should amend it to return some kind of error.

mholt: Hmm, I agree, this could use some improvement. What if the user has a local (internal) hostname of `php` because that's what they run their PHP app on?

wendigo:
@mholt: I agree but otherwise this change will break backward compatibility. I can move preset inside the fastcgi block so the new syntax will be:

```
fastcgi server1, server2, server3... {
    preset php
}
```

WDYT?

wendigo:
@tw4452852 I've changed the syntax. Can you take a look?

tw4452852:
The new syntax makes more sense. @mholt ?

mholt:
It's better, but darn, I really dislike moving the preset inside a block like that. I think, between proxying to multiple backends and setting the `php` preset, the php preset is going to be more common.

I vote we move the list of backends inside the server block with some subdirective instead of listing them along the top like we do for proxy.

Ideally, in the future, we could share a lot of code between proxy and fastcgi so that they can share syntaxes and each do their job with minimal code duplication.

wendigo:
@mholt: so what is your recommendation? Proxy syntax is: 
```
proxy from backend1 backend2 {
   preset
}
```

I can try to unify both proxy and fastcgi to share some code.

mholt:
@wendigo I'm still thinking about it; if proxy and fastcgi share code, they'll probably share syntax too... which is slightly problematic since the `php` preset is so common for FastCGI, but proxy has multiple presets, sometimes they are both used together. But maybe moving `php` to the inside like how proxy does its presets isn't a bad idea in the long-term?

Reusing the code between the two would result in great functionality but would also be a very ambitious project -- I guess it depends what you're willing/able to do! :)

For now, what about something like this, which preserves current syntax compatibility:

```plain
fastcgi from backend1 preset {
    backends backend2 backend3 backend4
}
```

wendigo:
@mholt I will think how to do that :)

tw4452852:
LGTM

wendigo:
I'm okay with that - I have more spare time following weeks so I will work on refactoring it either way

mholt:
Oh, shoot, I forgot about the syntax. I don't want to move the php preset inside the block, as we've discussed:

```plain
fastcgi from backend1 preset {
    upstream backend2
    upstream backend3
    upstream backend4
}
```

Sorry. Can we have that change and then I'll merge it? ;)

wendigo:
Yes, sure, I will change syntax in a couple of hours :) Stay tuned!

wendigo:
Oh crap, I've removed my fork :) I will have to push this PR to new branch and make a new PR instead.

wendigo:
Ok, I've created new branch containing changes from this PR + syntax change (as @mholt suggested):

```
fastcgi from backend1 preset {
    upstream backend2
    upstream backend3
    upstream backend4
}
```

PR #1207, opened 2016-10-24
proxy: set request's body to nil explicitly

tw4452852:
fix issue #1206

### DIFF
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -218,6 +218,11 @@ func (p Proxy) match(r *http.Request) Upstream {
 func createUpstreamRequest(r *http.Request) *http.Request {
 	outreq := new(http.Request)
 	*outreq = *r // includes shallow copies of maps, but okay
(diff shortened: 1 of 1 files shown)


wendigo:
How did you find that?

tw4452852:
Just bisect and find [this commit](https://github.com/golang/go/commit/4859f6a416b053d57fcc9d8f43e81e9d218280e9). It splits the returned body and content length. Before this commit, when content length is 0 && can't peek one byte from original body, the body will be set to nil. But now it will remain the original body. 
And for our case, the outgoing request is from a server request which its body is never nil:

```
// For server requests the Request Body is always non-nil
// but will return EOF immediately when no body is present.
// The Server will close the request body. The ServeHTTP
// Handler does not need to.
```

So we should reset its body to nil explicitly.

wendigo:
Makes sense - thank you for clarification :) I've been trying to fix that using delve but it's a lot of code to go through...

mholt:
Thank you Tw! :+1:

Issue #1205, opened 2016-10-22
Startup needs to be logged when running as service.


pwhodges:
When starting Caddy at the command line, it is clear what is happening, and in particular the sites being served are listed out.  When Caddy is running as a service this information is lost, because only stderr is captured in the log file. (Tested on Windows; but I presume *nix will be the same.)

Having the initial console messages up to and including the list of domains in the log would make it much easier to see when Caddy has been restarted, which would be be useful in assessing the errors before and after. It might even be useful to precede it in the log by a standard message saying "Caddy started" to make scripts which search in the log a little easier.

tobya:
@elcore @mholt 

in[ server.go OnStartupComplete the call to fmt.Println](https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/server.go#L365) outputs the servername being served.

However replacing

    		fmt.Println(output)
with 

		log.Println(output)

will cause a number of issues

1. list of hosts will only be show when -log is used on command line
1. also the date is printed infront of the host name `2016/10/23 09:15:34 http://127.0.0.1:1234` which is not a problem but is less asthetically pleasing

Is there a simple solution to these issues with log ?  I did wonder about keepting fmt.Println and building a list of hosts to be output when all servers have been listed and then outputing via log.Println but I would need to store the string list on a global level as OnStartupComplete is called (and logged) for each host.

Any pointers of how this should be implemented?

elcore:
Hello @tobya,

In Linux (e.g. Ubuntu/Debian) one can use `sudo systemctl status caddy`

```
Oct 22 10:10:10 host systemd[1]: Started Caddy HTTP/2 web server .
Oct 22 10:10:10 host caddy[3314]: Activating privacy features... done.
Oct 22 10:10:10 host caddy[3314]: https://domain.tld
Oct 22 10:10:10 host caddy[3314]: http://domain.tld
```

You could use something like (non-tested):

```
// OnStartupComplete lists the sites served by this server
// and any relevant information, assuming caddy.Quiet == false.
func (s *Server) OnStartupComplete() {
        if caddy.Quiet {
                return
        }

        for _, site := range s.sites {
                output := site.Addr.String()
                if caddy.IsLoopback(s.Address()) && !caddy.IsLoopback(site.Addr.Host) {
                        output += " (only accessible on this machine)"
                }
                if runtime.GOOS == "windows" {
                        l := log.New(os.Stdout, "[Caddy] Serving: ", 0)
                        l.Println(output)
[...]

PR #1210, opened 2016-10-24
Log Site start to -log. Fix for #1205

tobya:
This is a 1 line fix for #1205. 

Details of each site started will be written to Log as well as stdout.

I dont think this needs a test case.

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -363,6 +363,8 @@ func (s *Server) OnStartupComplete() {
 			output += " (only accessible on this machine)"
 		}
 		fmt.Println(output)
+		//Also print to -log.
+		log.Println(output)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/server.go
```diff
@@ -363,6 +363,8 @@ func (s *Server) OnStartupComplete() {
 			output += " (only accessible on this machine)"
 		}
 		fmt.Println(output)
+		//Also print to -log.
```
mholt: This comment isn't interesting. :) Let's remove it.

caddyhttp/httpserver/server.go
```diff
@@ -363,6 +363,8 @@ func (s *Server) OnStartupComplete() {
...
+		log.Println(output)
```
mholt: Do you want to suppress this output if Caddy is running quietly?

tobya: I'm not sure I understand.  By my basic testing if I run `caddy -quiet` I get no output. 

If I run `caddy -log stdout -quiet` I get the Startup and SIGINT log but not the output of line 367

Am I missing something?

mholt:
Thanks Tobya!

Issue #1183, opened 2016-10-13
Headers not removed on error pages


HazCod:
### 1. What version of Caddy are you running (`caddy -version`)?

Caddy 0.9.3
### 2. What are you trying to do?

Remove the Server header at all times.
### 3. What is your entire Caddyfile?

``` text
hazcod.com {

    log stdout
    errors stderr

#    log /tmp/caddy.log {
#        rotate {
#            size 10
#            keep 5
#        }
#    }

#    errors {
#        log /tmp/caddy.log {
#            size 10
#            keep 5
#        }
#    }

    header / {
        "Strict-Transport-Security" "max-age=63072000; includeSubdomains; preload"
        "X-Frame-Options" "DENY"
        "X-Content-Type-Options" "nosniff"
        "X-XSS-Protection" "1; mode=block"
        "Access-Control-Allow-Origin" "{host}"
        "Access-Control-Allow-Methods" "GET, POST, OPTIONS"
        -Server
    }

    gzip
    minify

    fastcgi / php-fpm:8080 php {
        except /css/,/js/,/img/,/fonts/
    }

    expires {
        match .js$ 1w
        match .css$ 1w
        match .png$ 1w
    }
    root /assets/
}

```
### 4. How did you run Caddy (give the full command and describe the execution environment)?

sudo ./caddy -conf Caddyfile -email hazcod@xxx.xx -agree
### 5. What did you expect to see?

No `Server: Caddy` header when receiving a 404
### 6. What did you see instead (give full error messages and/or log)?

`Server: Caddy` on a 404
### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

`header -Server`, query non-existant file and check response headers.

Edit: This is also the case when querying existing directories it seems.

tw4452852:
This problem is due to the header middleware only influences the following chain, while the `Server` header is set at the start of chain. There is quick fix for this case: 

```
diff --git a/caddyhttp/header/header.go b/caddyhttp/header/header.go
index dff925d..18deaa2 100644
--- a/caddyhttp/header/header.go
+++ b/caddyhttp/header/header.go
@@ -104,6 +104,9 @@ func (rww *responseWriterWrapper) addHeader(key, value string) {
 
 // delHeader registers a http.Header.Del operation
 func (rww *responseWriterWrapper) delHeader(key string) {
+       // remove the existing one if any
+       rww.w.Header().Del(key)
+
        rww.ops = append(rww.ops, func(h http.Header) {
[...]

PR #1212, opened 2016-10-25
header: only register deletion operation

tw4452852:
fix issue #1183

Signed-off-by: Tw tw19881113@gmail.com


caddyhttp/header/header_test.go
```diff
@@ -25,21 +25,23 @@ func TestHeader(t *testing.T) {
...
+		{"/a", "Server", ""},
 		{"/a", "ServerName", hostname},
 		{"/b", "Foo", ""},
 		{"/b", "Bar", "Removed in /a"},
 	} {
 		he := Headers{
 			Next: httpserver.HandlerFunc(func(w http.ResponseWriter, r *http.Request) (int, error) {
 				w.Header().Set("Bar", "Removed in /a")
-				fmt.Fprint(w, "This is a test")
+				w.WriteHeader(http.StatusOK)
```
mholt: Is this significant, or just because writing a body isn't necessary?

caddyhttp/header/header_test.go
```diff
@@ -50,6 +52,8 @@ func TestHeader(t *testing.T) {
 		}
 
 		rec := httptest.NewRecorder()
+		// preset header
+		rec.Header().Set("Server", "Caddy")
```
mholt: Do we assert against this somewhere? What exactly is this testing?

caddyhttp/header/header_test.go
tw4452852: Just want to confirm that there is no need to write a body.

caddyhttp/header/header_test.go
tw4452852: Just want to confirm the headers set before `Header.ServeHTTP` are removed after all.

tw4452852: Just want to confirm that the headers set before `header.HTTP` are removed after all.

mholt:
Thanks Tw! I really, really appreciate your help. :)

PR #1213, opened 2016-10-25
Remove dead code, do struct alignment, simplify code

wendigo:
I've passed code through gometaliner and made some changes: aligned structs, removed dead/unused code, simplified some statements.

### DIFF
--- a/caddyfile/parse.go
+++ b/caddyfile/parse.go
@@ -16,8 +16,7 @@ import (
 // pass in nil instead.
 func Parse(filename string, input io.Reader, validDirectives []string) ([]ServerBlock, error) {
 	p := parser{Dispenser: NewDispenser(filename, input), validDirectives: validDirectives}
-	blocks, err := p.parseAll()
-	return blocks, err
+	return p.parseAll()
 }
 
 // allTokens lexes the entire input, but does not parse it.
@@ -62,12 +61,7 @@ func (p *parser) parseAll() ([]ServerBlock, error) {
 func (p *parser) parseOne() error {
 	p.block = ServerBlock{Tokens: make(map[string][]Token)}
 
-	err := p.begin()
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return p.begin()
 }
 
 func (p *parser) begin() error {
@@ -76,6 +70,7 @@ func (p *parser) begin() error {
 	}
 
 	err := p.addresses()
+
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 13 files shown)


caddyhttp/fastcgi/fcgiclient_test.go
```diff
@@ -155,7 +155,7 @@ func sendFcgi(reqType int, fcgiParams map[string]string, data []byte, posts map[
 	fcgi.Close()
 	time.Sleep(1 * time.Second)
 
-	if bytes.Index(content, []byte("FAILED")) >= 0 {
+	if bytes.Contains(content, []byte("FAILED")) {
```
wendigo: bytes has Contains method as well

caddyhttp/fastcgi/fcgiclient.go
```diff
@@ -261,29 +260,6 @@ func (c *FCGIClient) writePairs(recType uint8, pairs map[string]string) error {
 	return nil
 }
 
-func readSize(s []byte) (uint32, int) {
```
wendigo: These two methods were unused

caddyhttp/fastcgi/fcgiclient.go
```diff
@@ -44,7 +44,6 @@ const FCGINullRequestID uint8 = 0
 
 // FCGIKeepConn describes keep connection mode.
 const FCGIKeepConn uint8 = 1
-const doubleCRLF = "\r\n\r\n"
```
wendigo: This was unused across the codebase

caddyfile/parse.go
```diff
@@ -76,6 +70,7 @@ func (p *parser) begin() error {
 	}
 
 	err := p.addresses()
+
```
mholt: I'm going to be pedantic here and suggest that this empty line shouldn't be here, as the error check on the next line is directly associated with this line. ;)

Anyway, I won't make this a stopper for the PR if it's the only change I would want.

caddytls/config.go
```diff
@@ -418,27 +418,6 @@ var supportedCiphersMap = map[string]uint16{
 	"RSA-3DES-EDE-CBC-SHA":          tls.TLS_RSA_WITH_3DES_EDE_CBC_SHA,
 }
 
-// List of supported cipher suites in descending order of preference.
-// Ordering is very important! Getting the wrong order will break
-// mainstream clients, especially with HTTP/2.
-//
-// Note that TLS_FALLBACK_SCSV is not in this list since it is always
-// added manually.
-var supportedCiphers = []uint16{
```
mholt: Wow, I guess I must've stopped using this with the 0.9 rewrite and not noticed it.

Makes sense though, I don't see why we would need this now. The order is still important, but on the defaultCiphers slice (below) -- I'll have to update the godoc comments (aside from this) - thanks!

PR #1214, opened 2016-10-26
Use http.Header instead of custom type

wendigo:
Get rid of TODO and custom Header type.

### DIFF
--- a/caddyhttp/header/header.go
+++ b/caddyhttp/header/header.go
@@ -26,15 +26,21 @@ func (h Headers) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 	rww := &responseWriterWrapper{w: w}
 	for _, rule := range h.Rules {
 		if httpserver.Path(r.URL.Path).Matches(rule.Path) {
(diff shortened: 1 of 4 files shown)


wendigo:
Actually that was only todo in the source-code and now it's gone :sunglasses:

caddyhttp/header/header.go
```diff
@@ -44,16 +49,9 @@ func (h Headers) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 
 type (
 	// Rule groups a slice of HTTP headers by a URL pattern.
-	// TODO: use http.Header type instead?
```
mholt: YES, I love getting rid of TODOs! Thanks!

mholt:
Oops. I merged #1212 before this which created a conflict here. Sync up and we can merge it!

caddyhttp/header/header.go
wendigo: Same here. Done - ready to merge

mholt:
Thank you @wendigo! :tada:

PR #1217, opened 2016-10-26
Update README.md

chris-bandce:


mholt:
Thanks, but I think it's fine as-is.

Issue #1200, opened 2016-10-21
Using proxy with a query results in an added slash being forwarded 


elvarb:
### 1. What version of Caddy are you running (`caddy -version`)?

Caddy 0.9.3
### 2. What are you trying to do?

I'm trying to make a simple proxy for the http endpoint on NSQ
Any data posted to the site root should be forwarded to the proxy url where the topic is defined

In the NSQ log files I can see that the request goes to the wrong address
### 3. What is your entire Caddyfile?

``` text
localhost:8001
proxy / localhost:4151/pub?topic=caddy
```
### 4. How did you run Caddy (give the full command and describe the execution environment)?

Ran the binary, it read the CaddyFile
### 5. What did you expect to see?

Proxy should forward the connection to the correct destination defined in the config

/pub?topic=caddy
### 6. What did you see instead (give full error messages and/or log)?

NSQ got the query like this

/pub/?topic=caddy
### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

If you want to use NSQ then just start NSQd and it will run with the default config. Monitor stdout.

mholt:
When you say "latest" do you mean latest release or a commit? Please be specific, this project is in motion! For example, we have a recent commit, #1178, that may affect this issue.

elvarb:
Sorry, version added :) more motion!

That commit very well could solve this

mholt:
Thanks! Do you have the ability to build from source? Perhaps you would have a chance to check it before I do (I have midterms for another week).

elvarb:
Not at the moment no, possibly next week.

tobya:
With my limited testing #1178 4baca884c5e4767b114f3d363188e6570a24e90d  does not seem to fix this issue, the extra / is still added to the forward-proxy url

mholt:
Okay -- I think that was for a different issue (but still could have affected this one). So I guess we need to look into this one more.

tobya:
I've looked at the code responsible & it seems to assume the proxy will be an upstream directory rather than a file. I will see if I can make progress

tobya:
I need some discussion on expected behaviour.

Currently what happens is that the proxy is assumed to be a directory which can have first directories and then query parameters added to it. 

so given

    localhost:8001
    proxy / localhost:4151/pub?topic=caddy

visiting `localhost:8001` will add the extra `/` not desired to get `localhost:4151/pub/?topic=caddy` and I could just prevent this from happening.  

[...]

PR #1219, opened 2016-10-26
Starting possible fix for #1200

tobya:
I wanted to see if we were happy with the direction of this fix for #1200

If the proxy address has a query on the end ?some=value then discard any additional path information .
eg

```
localhost:8001
proxy / localhost:4151/pub?topic=caddy
```

then the url `localhost:8001/big/long/url/index.php?other=value`

would proxy to `localhost:4151/pub?topic=caddy&other=value`

the information contained in `big/long/url/index.php` is discarded.

Is this the desired behaviour?

ps. I know I will need to add test before submitting.

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -98,11 +98,17 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 			}
 		}
 
-		hadTrailingSlash := strings.HasSuffix(req.URL.Path, "/")
-		req.URL.Path = path.Join(target.Path, req.URL.Path)
-		// path.Join will strip off the last /, so put it back if it was there.
-		if hadTrailingSlash && !strings.HasSuffix(req.URL.Path, "/") {
-			req.URL.Path = req.URL.Path + "/"
+		//If there is no query, there is no ? so proxy destination is the begining of a path.
+		//Otherwise simply proxy to address and discard additional path information.
+		if target.RawQuery == "" {
+			hadTrailingSlash := strings.HasSuffix(req.URL.Path, "/")
+			req.URL.Path = path.Join(target.Path, req.URL.Path)
+			// path.Join will strip off the last /, so put it back if it was there.
+			if hadTrailingSlash && !strings.HasSuffix(req.URL.Path, "/") {
+				req.URL.Path = req.URL.Path + "/"
+			}
+		} else {
+			req.URL.Path = target.Path
 		}
 
 		// Trims the path of the socket from the URL path.


tw4452852:
Actually I propose to disable query in the destination endpoint to proxy to. So `proxy / localhost:4151/pub?topic=caddy' will be invalid.

tobya:
@elvarb Do you have an opinion on this?

elvarb:
@tobya this solution would probably work in my case. This is a tougher problem to solve than I thought.

In this case would the url with other=value1 replace the proxy value of other=value2 ? 

URL: localhost:8001/big/long/url/index.php?other=value1

Configured proxy: localhost:4151/pub?topic=caddy&other=value2

Results: localhost:4151/pub?topic=caddy&other=value1
Or
Results: localhost:4151/pub?topic=caddy&other=value2

tobya:
I would need to test it but I would guest it will be

    Results: localhost:4151/pub?topic=caddy&other=value2&other=value1

What happens then is up to the server.

I am starting to agree with @tw4452852 that this may just be illegal to have a query in a proxy url.

I think what you want is really a redir?

mholt:
Hmm, what does it really mean to proxy to a URL with a query string? Are you adding those parameters, overwriting them, or replacing all of them?

- Adding the parameters seems reasonable
- Overwriting them (setting them) seems a bit cavalier
- Replacing the entire query string with the one in the proxy config seems destructive

Duplicate QS parameters are allowed, I think that would be okay, no?

Either way, I think that the current (?) behavior of detecting whether the upstream is a directory or a file by looking for a file extension is not very robust. What are the consequences of not adding the `/` at all?

tobya:
@mholt  The problem really is what to do in all instances.

We can **not** add the `/` when there is no extra path added to the incoming path to be proxied, but what do we do when there is an incoming path to be proxied?  

eg. given caddyfile we are dealing with

    localhost:8001
    proxy / localhost:4151/pub?topic=caddy

then the url `localhost:8001?other=value`

is fine, it would proxy to 
    
    localhost:4151/pub?topic=caddy&other=value

but what happens when there is an incoming path 

    localhost:8001/big/long/url/index.php?other=value

Now you need to make a  choice.  do we treat `pub` as part of the path or a file?

You can proxy to

-  localhost:4151/pub/big/long/url/index.php?topic=caddy&other=value

or

-  localhost:4151/pub?topic=caddy&other=value

We need to have a very clear indication as to what happens.  I am begining to agree with @tw4452852 that this should not be allowed.  I think what @elvarb is trying to do is better achieved with a `redir`

mholt:
Thanks for the explanation @tobya. Do we know how other servers do this? e.g. Nginx.

But as I read that, with the instance given here, I'm feeling like the first destination is probably more correct: `localhost:4151/pub/big/long/url/index.php?topic=caddy&other=value`

PR #1221, opened 2016-10-27
Add support for OCSP Must-Staple for Let's Encrypt certs

elcore:
We need to fix Caddy <caddytls> due to https://github.com/xenolf/lego/commit/72914df00f027074ea70c878dde401f02fa22f09


tehscorpion:
Should fix #1222

caddytls/client.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/xenolf/lego/acme"
 )
 
+// Add the must staple TLS extension to the CSR generated by lego/acme
+var mustStaple bool
```
mholt: This should not be 'global' -- probably scoped to the config struct like the other values in the TLS directive.

caddytls/client.go
```diff
@@ -197,7 +200,7 @@ Attempts:
 	for attempts := 0; attempts < 2; attempts++ {
 		namesObtaining.Add([]string{name})
 		acmeMu.Lock()
-		certificate, failures := c.acmeClient.ObtainCertificate([]string{name}, true, nil)
+		certificate, failures := c.acmeClient.ObtainCertificate([]string{name}, true, nil, mustStaple)
```
mholt: Replace with `c.MustStaple` once it gets moved into the config struct

elcore:
> (Update: Must-Staple will definitely not be the default in the foreseeable future in Caddy.)

@mholt okay :smile: any reason why?

mholt:
Fantastic -- thank you, @elcore!!

Issue #1198, opened 2016-10-20
List files in a directory inside a template


tristanrollet:
Hi,

I would like to know if it was possible to have a feature which would allow to list files in a directory, inside a template. The "browse" directive isn't really what I'm trying to achieve.

Here's a link to the forum thread: https://forum.caddyserver.com/t/list-files-in-a-directory-inside-a-template/841

Sorry if this isn't precise enough, if you have any questions, I'll be glad to answer them.

crackerplace:
Am starting with go and would be interested to work on this.

mholt:
@crackerplace  Go for it. :)

ericdreeves:
I also started on this earlier in the week. I'll submit a PR, but feel free to reject it if @crackerplace has dibs. I don't want to step on any toes!

Since @tristanrollet nrollet cited Readdir() as an example, the new Files template action behaves similarly to it and Readdirnames, but doesn't take an integer as an argument.

I assumed that the directory would be within the root of the context. Let me know if that was incorrect.

crackerplace:
@ericdreeves Sure , Please continue.I will work on some other issue.Not a problem.

crackerplace:
@mholt Please let me know any issue which I can work on.Am looking into the issues too.

k0nsl:
To @all working on this I just want to say: thank you. I am also very interested in improvements / changes regarding this.

I run a lot of mirrors for various projects (FOSS, etc,.) and I typically create a very simple directory index for the files, e.g something like this:
http://devuan.mirror.k0nsl.org/

But if this can be improved, I'd glad to implement those changes in my projects.

georgelesica-wf:
Was this ever actually resolved? When I try to use `{{.Files "path"}}` I get the following error:

```
30/Apr/2017:13:14:23 -0600 [ERROR 500 /] template: :13:14: executing "" at <.Include>: error calling Include: template: meta/tmpl/_index.html:7:2: executing "meta/tmpl/_index.html" at <.Files>: Files has arguments but cannot be invoked as function
```

I'm on Caddy 0.10.0 installed through Homebrew and Go 1.7.1.

georgelesica-wf:
Ah, so I think I know why I'm getting that error. There's already a `.Files` in scope in a Markdown document:

```go
	mdData := Data{
		Context: ctx,
		Doc:     mdata.Variables,
		Styles:  c.Styles,
		Scripts: c.Scripts,
		Meta:    meta,
		Files:   files,
[...]

PR #1226, opened 2016-10-29
Add Files action to template context. (#1198)

ericdreeves:
See #1198.

Since @tristanrollet cited Readdir() as an example, the new Files template action behaves similarly to it and Readdirnames(), but doesn't take an integer as an argument.

I assumed that the directory would be within the root of the context. Let me know if that was incorrect.


ericdreeves:
Investigating failures.

wendigo:
Hi, @ericdreeves: first `gofmt -s -w .` the source code :)

caddyhttp/httpserver/context_test.go
```diff
@@ -684,3 +685,103 @@ func TestTemplates(t *testing.T) {
...
+		},
+	}
+
+	for i, test := range tests {
+		context := getContextOrFail(t)
+		testPrefix := getTestPrefix(i + 1)
+
+		// Create directory / files from test case.
+		if test.fileNames != nil {
+			dirPath := fmt.Sprintf("%s/%s", context.Root, test.dirName)
```
wendigo: use `filepath.Join()` (https://golang.org/pkg/path/filepath/#Join)

ericdreeves:
Yeah, sorry about that. It looks like there's still some issues with the test and how it creates files.

caddyhttp/httpserver/context_test.go
```diff
@@ -684,3 +685,103 @@ func TestTemplates(t *testing.T) {
...
+			}
+			defer func() {
+				if err := os.RemoveAll(dirPath); err != nil && !os.IsNotExist(err) {
+					t.Fatalf(testPrefix+"Expected no error removing directory, got: '%s'", err.Error())
+				}
+			}()
+
+			for _, name := range test.fileNames {
+				absFilePath := filepath.Join(dirPath, name)
+				if err := ioutil.WriteFile(absFilePath, []byte(""), os.ModePerm); err != nil {
```
wendigo: This line fails on travis, but not locally.

wendigo:
@ericdreeves, now one test fails on AppVeyor: 

```
context_test.go:745: Test [4]: Expected no error removing directory, got: 'remove C:\Users\appveyor\AppData\Local\Temp\1\caddy_test708967809\file1: The process cannot access the file because it is being used by another process.'
```

mholt:
@wendigo You're calling a function with `defer` in a for loop. This is almost always a bug (in fact, I'm surprised `go vet` didn't call this out). Maybe you mean to call the function at the end of each iteration of the for loop?

wendigo:
@mholt It's not my PR :)

mholt:
Oops. I meant @ericdreeves. Sorry.

ericdreeves:
Thanks, @mholt and @wendigo. That still doesn't satisfy AppVeyor, but maybe I missed something. Could it be that I'm removing all the files in the directory via os.RemoveAll? At any rate, thanks for your patience. I understand if you want to reject the PR so that I can rework the tests.

ericdreeves:
I feel silly for not noticing that before. Again, thank you both for working with me. I set up AppVeyor against my fork so it won't happen again. Obviously there's lots of commits here, so let me know what else I can do to assist.

caddyhttp/httpserver/context.go
```diff
@@ -288,3 +288,33 @@ func (c Context) Map(values ...interface{}) (map[string]interface{}, error) {
...
+	if !stat.IsDir() {
+		return nil, fmt.Errorf("%v is not a directory", name)
+	}
+
+	dirInfo, err := dir.Readdir(0)
+	if err != nil {
+		return nil, err
+	}
+
+	var names []string
```
wendigo: instead of append (and growing names) you can init it to len(dirInfo): 
```names := make([]string, len(dirInfo))``` and then:

```
for i, fileInfo := range dirInfo {
    names[i] = fileInfo.Name()
}
```

In this case you know exactly what is the size of ```names```

mholt:
@ericdreeves Thanks for the revisions. Looks good from here, so I've merged it. I'll give you collaborator status so you can maintain it at will, help review PRs, and/or contribute more things without having to fork the project. Again, thank you!

PR #1232, opened 2016-11-01
proxy: record request Body for retry (fixes #1229)

### LINKED ISSUE #1229: proxy: retrying POST request failed with http: invalid Read on closed Body
(Are you asking for help with using Caddy? Please use our forum instead: https://forum.caddyserver.com. If you are filing a bug report, please answer the following questions. If your issue is not a bug report, you do not need to use this template. Either way, please consider donating if we've helped you. Thanks!)

I tested the newest master caddy today.

I try to test the proxy with POST method and body. When proxying to backend servers with load balance policy, if an error happened such as the backend server doesn't work well caddy would request another backend server. 
And I add one log in following location.
res, err := transport.RoundTrip(outreq)
	if err != nil {
		log.Println("err:", err)
		return err
	}

While if this failing request is a POST request and whose body has data, the second try would fail either because of the error "http: invalid Read on closed Body" even if the backend server is normal and it should be forwarded successfully in this second try.

```text
localhost:2018

proxy / 192.168.206.231:9606 localhost:9608{
        policy round_robin
        max_fails 5
        try_interval 60s
}
```
PS: 192.168.206.231:9606 workes well while localhost:9608 doesn't exist.

OS: win7 x64
go version: go1.7
First I "go install" the source and then cd /bin directory, finally I run the caddy command
image

I expect to see caddy can proxy successfully if I configure several endpoints even if there are some wrong endpoints.

I got the log "err: http: invalid Read on closed Body" 
image

I have Googled this issue and found an extremely similar one.  They have solved this.
https://github.com/flynn/flynn/issues/872

tpng:


mholt:
Hey @tpng, thanks for the PR! Really glad you're helping out. :smile:

Did you see the fix for a similar issue in flynn? https://github.com/flynn/flynn/pull/875/files - they make "Close()" fake so that they only close it when they really mean to. What do you think of that approach instead of cloning? Seems to be a little simpler. (I haven't looked super thoroughly into the issue, but it seems like their problem was pretty much the same one.)

tpng:
The reason for clone is that the request body maybe read by upstream already. In this case, fake close will end up with a dirty body for us to retry the request. It seems it is not possible for us to replay the downstream request without recording the body as it is read.

tpng:
I think we don't need to clone it every retry, we can just record the downstream body once and reuse it.

tpng:
@mholt PR updated.

caddyhttp/proxy/body.go
```diff
@@ -0,0 +1,39 @@
...
+	return err
+}
+
+// newBody returns *body to use in place of src. Closes src and returns Read
+// error on src. All content from src is buffered in *body.
+func newBody(src io.ReadCloser) (*body, error) {
+	if src == nil {
+		return nil, nil
+	}
+	b, err := ioutil.ReadAll(src)
```
tw4452852: I think we should close the original body here: `src.Close()`

caddyhttp/proxy/body.go
```diff
@@ -0,0 +1,39 @@
+package proxy
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+)
+
+type body struct {
```
tw4452852: Make a more sensitive name?

tpng:
@tw4452852 Thanks! PR updated.

tw4452852:
I think we'd better add a test case in `proxy_test.go` for issue #1229.

tpng:
Any idea on how to write the test case for #1229?

tpng:
@tw4452852 Test added for #1229.

tw4452852:
Sorry for late. LGTM.

PR #1239, opened 2016-11-06
Filter empty headers

### LINKED ISSUE #1225: Caddy passing empty headers
During some testing on #1170 it was found that when a `header_upstream` is defined, even if the incoming request didn't contain one, an empty one is provided to the upstream server. @ericdreeves confirmed this in his testing [here](https://github.com/mholt/caddy/issues/1170#issuecomment-255552317):

```
Adding an upstream header via:

header_upstream Sec-WebSocket-Protocol {>Sec-WebSocket-Protocol}

does indeed add the header to the upstream request, however it will be empty
if the header did not exist in the original request.
```

Effectively, this means Caddy is sending malformed HTTP to the upstream ;( More on that [here](http://stackoverflow.com/a/12131993). Ideally, we should be intelligently checking for this first ;)

gyulavoros:
Some web servers (e.g. Jetty 9.3) don’t like HTTP headers with empty values. This commit filters header replacements with zero length.

Should fixes #1225

### DIFF
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -273,7 +273,10 @@ func mutateHeadersByRules(headers, rules http.Header, repl httpserver.Replacer) 
 	for ruleField, ruleValues := range rules {
 		if strings.HasPrefix(ruleField, "+") {
 			for _, ruleValue := range ruleValues {
-				headers.Add(strings.TrimPrefix(ruleField, "+"), repl.Replace(ruleValue))
+				var replacement = repl.Replace(ruleValue)
+				if len(replacement) > 0 {
+					headers.Add(strings.TrimPrefix(ruleField, "+"), replacement)
+				}
 			}
 		} else if strings.HasPrefix(ruleField, "-") {
 			headers.Del(strings.TrimPrefix(ruleField, "-"))
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -273,7 +273,7 @@ func mutateHeadersByRules(headers, rules http.Header, repl httpserver.Replacer) 
 	for ruleField, ruleValues := range rules {
 		if strings.HasPrefix(ruleField, "+") {
 			for _, ruleValue := range ruleValues {
-				var replacement = repl.Replace(ruleValue)
+				replacement := repl.Replace(ruleValue)
 				if len(replacement) > 0 {
 					headers.Add(strings.TrimPrefix(ruleField, "+"), replacement)
 				}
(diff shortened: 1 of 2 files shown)


wendigo:
Hi, this behaviour must be tested. Please add test case that covers that.

caddyhttp/proxy/proxy.go
```diff
@@ -278,7 +278,10 @@ func mutateHeadersByRules(headers, rules http.Header, repl httpserver.Replacer)
 		} else if strings.HasPrefix(ruleField, "-") {
 			headers.Del(strings.TrimPrefix(ruleField, "-"))
 		} else if len(ruleValues) > 0 {
-			headers.Set(ruleField, repl.Replace(ruleValues[len(ruleValues)-1]))
+			var replacement = repl.Replace(ruleValues[len(ruleValues)-1])
+			if len(replacement) > 0 {
+				headers.Set(ruleField, replacement)
+			}
```
mholt: This applies only in the "set" headers case, but what about "adding" a header (field prefixed with `+`)?

gyulavoros: I have an alternative idea to implement this. How about to define an extra prefix, let's say `?` and apply conditional filtering only on headers marked with this. This way it wouldn't change existing behavior, and people have issues with empty headers can use this method e.g.:

`header_upstream ?Upgrade {>Upgrade}`

mholt: I haven't heard of anyone who needs the current, existing behavior... and I'd rather not introduce another prefix if possible, to keep things simpler.

caddyhttp/proxy/proxy.go
```diff
@@ -273,12 +273,18 @@ func mutateHeadersByRules(headers, rules http.Header, repl httpserver.Replacer)
 	for ruleField, ruleValues := range rules {
 		if strings.HasPrefix(ruleField, "+") {
 			for _, ruleValue := range ruleValues {
-				headers.Add(strings.TrimPrefix(ruleField, "+"), repl.Replace(ruleValue))
+				var replacement = repl.Replace(ruleValue)
```
mholt: Stylistic nit-pick: change to `replacement := ...` rather than using the `var` longform. (Same below)

gyulavoros:
@mholt Should I rebase on master and squash commits?

mholt:
I got it ;) Thank you!!

mholt:
@gyulavoros Would you like to be a collaborator on the project? It'll give you the ability to create branches and submit PRs without having to fork the project, and you can also help review other pull requests and help with issues in the future.

gyulavoros:
@mholt That would be great, we wouldn't mind investing more in Caddy, as we use it and like a lot.

PR #1245, opened 2016-11-08
Reorder Basicauth

### LINKED ISSUE #1244: Basicauth not working when the url is rewritten
Don't know if it's the indented behavior or not.

```text
root@server1:~/test# ./caddy --version
Caddy 0.9.3
```

Using basicauth to protect an url that is in a rewrite statement.

```text
[website]:8000

tls off

root /root/test/html

fastcgi / unix:/run/php5-fpm.sock php

rewrite {
        to {path} {path}/ /index.php?{query}
}

basicauth /test user pass
```

`./caddy`

I expect the browser ask for credentials when i visit the url
`http://[website]:8000/test`

The browser didn't ask for credentials and showed directly the requested page.

This behavior can be reproduced using the Caddyfile that i provided at point 2.
If you comment the rewrite block, the basicauth works fine.

elcore:
In order to fix #1244 we need to reorder basicauth

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -405,6 +405,7 @@ var directives = []string{
 	// directives that add middleware to the stack
 	"locale", // github.com/simia-tech/caddy-locale
 	"log",
+	"basicauth",
 	"rewrite",
 	"ext",
 	"gzip",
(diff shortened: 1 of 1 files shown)


elcore:
You are right, but basicauth needs to be executed before rewrite (to fix the issue) -- Any idea?

mholt:
I do not think, at this point, that basicauth should be moved; I think there is a better way for the asker to fix their issue.

PR #1247, opened 2016-11-09
Basicauth: Use Caddy-Rewrite-Original-URI header

elcore:
Is this better @mholt ?

### DIFF
--- a/caddyhttp/basicauth/basicauth.go
+++ b/caddyhttp/basicauth/basicauth.go
@@ -32,6 +32,11 @@ type BasicAuth struct {
 	Rules    []Rule
 }
 
+// When a rewrite is performed, a header field of this name
+// is added to the request
+// It contains the original request URI before the rewrite.
+const internalRewriteFieldName = "Caddy-Rewrite-Original-URI"
+
 // ServeHTTP implements the httpserver.Handler interface.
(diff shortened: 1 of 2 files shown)


mholt:
But the question is, is this what the user intends? If they do a rewrite, maybe they're doing the rewrite so that it matches the URI accepted by basicauth...

elcore:
> But the question is, is this what the user intends? If they do a rewrite, maybe they're doing the rewrite so that it matches the URI accepted by basicauth...

I don't know! I thought of that too, but I don't know any case where it might be applied to.

mholt:
I think we just have to decide one way or the other what the semantics of 'rewrite' are. For now, rewrite is a change to the URI as if the original request had been to the rewritten URI in the first place; having basicauth ignore that as an exception doesn't seem like a good idea at this point. But again, thank you so much for jumping in and contributing! I think we would need a lot more discussion on this if we were to make a change though.

elcore:
>  I think we just have to decide one way or the other what the semantics of 'rewrite' are. For now, rewrite is a change to the URI as if the original request had been to the rewritten URI in the first place; having basicauth ignore that as an exception doesn't seem like a good idea at this point. But again, thank you so much for jumping in and contributing! I think we would need a lot more discussion on this if we were to make a change though.

No problem :smile: --  I want to be a partner for you and for the community!  An active member of the community! The guy who cares about Caddy!

PR #1256, opened 2016-11-11
trial fix for #1200 with log

tobya:
So I have fixed the immediate issue but need help with the next step.  The code submitted does as requested.

caddyfile

    localhost:8001
    proxy / localhost:4151/pub?topic=caddy

request `localhost:8001`  will now create a proxy url out to  `localhost:4151/pub?topic=caddy`

You can see this if you run `caddy -log` as you will get 

    [INFO] Proxied Query:localhost:4151/pub?topic=caddy

However the actual proxied url is 

    localhost:4151/pub/?topic=caddy

so more code somewhere else is adding the / in.  From investigations I would imagine the url is being split again and then joined which causes this extra / to be added. 

I do not know enough about `go` or `caddy` to be able to find the possible places this is happening.  Perhaps some pointers from @mholt  or others would help.

PS. If building and testing from this branch, I recommend using curl to test as the browser caches and makes it difficult to test.

### DIFF
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -56,9 +56,7 @@ func TestReverseProxy(t *testing.T) {
 	// Make sure {upstream} placeholder is set
 	rr := httpserver.NewResponseRecorder(httptest.NewRecorder())
 	rr.Replacer = httpserver.NewReplacer(r, rr, "-")
-
 	p.ServeHTTP(rr, r)
-
 	if got, want := rr.Replacer.Replace("{upstream}"), backend.URL; got != want {
 		t.Errorf("Expected custom placeholder {upstream} to be set (%s), but it wasn't; got: %s", want, got)
 	}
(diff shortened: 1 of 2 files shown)


tobya:
Sorry failing

tobya:
This PR is a step on the way to solve #1200 however I'm getting failing tests that I can't figure out how I've affected.

     proxy_test.go:63: Expected custom placeholder {upstream} to be set (http://127.0.0.1:62567), but it wasn't; got: -

I don't understand why {upstream} is unavailable.  When I remove my code this error goes away.

I added a test case which checks that the / is not added in this one case.

mholt:
Huh, I'm not sure either, about either of the questions. I'd need more time to look into it but I don't have it right now. (Somebody else?)

tw4452852:
Should be fixed now, the reason is the target path is `""` while the request's path is `/`, so the result will be `""` which isn't expected.

tobya:
Thanks I'll take a look

tobya:
@tw4452852 Thanks that fixed the failing test.  I still have the issue that something else in the code is adding in the final '/'

tobya:
I think this fix should be abandoned.  There is something weird going on with / being added sometimes and not others.  It is a very niche example and introduces unpredictable behaviour.   Most of the time https://example.com/dir and https://example.com/dir/ should be the same and if not i would suggest site configuration is less transparent than it should be.

mholt:
@tobya I agree, this is not as straightforward as expected, and I'm really wondering if anything is actually broken. I agree with you about the trailing forward slash... by convention, a trailing forward slash indicates a directory, and a lack of a trailing forward slash indicates a file.

PR #1267, opened 2016-11-18
launchd service file for Mac

DenBeke:
I added the *launchd* service file from my blogpost to the *init* folder.

It's not as good as the Linux ones because paths are not that fixed on Mac. So users can put Caddy wherever they want. But I figured out that adding a sample plist file would be useful.


mholt:
Cool, thanks!

DenBeke:
With pleasure, @mholt!

PR #1272, opened 2016-11-23
Init switch from HOME to Caddy

genofire:
Maybe make the path also equal, i found:
- `/etc/ssl/caddy`
- `/etc/caddy`
- `/etc/caddy/ssl`

And for MAC, i do not know :sweat_smile:

### DIFF
--- a/dist/init/linux-systemd/caddy.service
+++ b/dist/init/linux-systemd/caddy.service
@@ -12,7 +12,7 @@ User=www-data
 Group=www-data
 
 ; Letsencrypt-issued certificates will be written to this directory.
-Environment=HOME=/etc/ssl/caddy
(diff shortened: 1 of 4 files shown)


tobya:
@genofire Thanks for the PR.  

Could you provide some explanation or discussion as to why you wish to make this change?

genofire:
Great.
If the init/systemd script use HOME, caddy creates there a folder `.caddy` for CADDYPATH.
This is a hidden folder, so it is a little bit strange for the systemwide config file folders under `/etc`.

To look into:
docs: https://caddyserver.com/docs/automatic-https
source: https://github.com/mholt/caddy/blob/master/assets.go

mholt:
I think this is a reasonable change. Thank you!

Issue #1094, opened 2016-09-07
Feature request:fastcgi timeout options


Echsecutor:
In nginx, there are various fastcgi  timeout options, e.g.:
http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_connect_timeout
http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_read_timeout

It would be nice to also have these options (or at least know the current values) in caddy.

ericdreeves:
I'll give this a shot.

ericdreeves:
Hello, just wanted to touch base on this request. I'm getting pretty close to a fix and hope to soon have a PR to review!

ericdreeves:
I've noticed an issue with the read_timeout directive. I'm working on the changes. Also, as requested, I'll set default values for both read_timeout and connect_timeout.

ericdreeves:
Work continues. I'll be investigating a send_timeout property next.

ericdreeves:
Hello, @Echsecutor!

We've added the following timeout properties to the fastcgi directive:
* connect_timeout
* read_timeout
* send_timeout

Each of these have a default value of 60 seconds.

Are these updates sufficient? If so then I'll close this issue.

ericdreeves:
Closing.

PR #1274, opened 2016-11-25
Fix read timeout and add default timeout values.

ericdreeves:
By setting the read deadline in streamReader.Read(), the deadline was
extended by the read timeout on each subsequent call. To avoid this, the
deadline is set in FCGIClient.Request(), before the first read occurs.

Additionally, the read_timeout and connect_timeout properties now have a
default value of 60 seconds.

See #1094.


ericdreeves:
As I look over this and start on the write_timeout property I'm thinking we might extract the rwc field on FCGIClient into a (unexported?) type that implements the net.Conn interface.

That would simplify things a bit. Thoughts?

mholt:
Thanks for taking initiative on this, @ericdreeves ! I don't have a chance to look at it yet but maybe somebody else does?

caddyhttp/fastcgi/setup.go
```diff
@@ -53,15 +53,13 @@ func fastcgiParse(c *caddy.Controller) ([]Rule, error) {
...
-		var rule Rule
-
 		args := c.RemainingArgs()
 
 		if len(args) < 2 || len(args) > 3 {
 			return rules, c.ArgErr()
 		}
 
-		rule.Path = args[0]
+		rule := Rule{Path: args[0], ReadTimeout: 60 * time.Second}
```
wendigo: I know that 60 seconds is a lot of time but this change can actually break existing configurations. I'm not really sure what value should be used as default. Suggestions @mholt ?

mholt: Why can this break existing configs?

ericdreeves: Thanks for pointing it out, @wendigo.

@mholt some configurations may rely on a read timeout of greater than 60 seconds, whether users know it or not. Affected users would have to update their Caddyfile with the new read_timeout property in order to preserve the previous behavior. And even if the property is used, the exact duration might not be known.

Some options:
1. I can remove the default value and users must explicitly declare the read_timeout
2. Make the default duration greater than the currently proposed one

Are there others?

wendigo: @mholt If it's explicitly stated in the documentation - I'm fine with default value 👍

ericdreeves: Thanks for the feedback.

Speaking of documentation, I'm happy to make those changes as well.

mholt:
Great, thank you for making progress on this!! :tada:

Issue #1276, opened 2016-11-27
Numeric markdown metadata can't be used in template


cork:
Numeric values in markdown metadata becomes `"<no value>"` in template.

template:
```
<!DOCTYPE html>
<html>
	<head><title></title></head>
	<body>
		...
		<input name="test" value="{{.Doc.number1}}">
		<input name="test" value="{{.Doc.number2}}">
		...
	</body>
</html>
```

markdown:
```
---
number1: 5
number2: "5"
---
```

Output:
```
<!DOCTYPE html>
<html>
	<head><title></title></head>
	<body>
		...
		<input name="test" value="<no value>">
		<input name="test" value="5">
		...
	</body>
</html>
```

Expected Output:
```
<!DOCTYPE html>
<html>
	<head><title></title></head>
	<body>
		...
		<input name="test" value="5">
		<input name="test" value="5">
		...
	</body>
</html>
```

Tested with Caddy 0.9.3.

wendigo:
You're right @cork - there's type assertion in https://github.com/mholt/caddy/blob/master/caddyhttp/markdown/metadata/metadata.go#L67

Providing fix in a couple of minutes

wendigo:
Here you go: https://github.com/mholt/caddy/pull/1278

wendigo:
It will now work for floats too ;)

mholt:
Why do we even need to do the conversion, again? I haven't looked at this code in a long time, but I feel like having to enumerate every possible type is impractical. (Maybe not for 99% of uses, but still...)

wendigo:
@mholt It's due to automatic type detection/conversion made in the yaml decoder: `something: number` and `something: 1234` will be decoded to `string : int` and `string : string` when decoding into `map[string]interface{}`. I'm fine with the change I made - it works in YAML, TOML and JSON files

tobya:
I find myself agreeing with @mholt on this. Could the markdown not simply specify that it can only be a string.  Non enclosed values are simply not allowed?

mholt:
@wendigo I like your change too, I'm just wondering if we can improve it a little more by making it kind of agnostic of the type. :) (If not, I'll merge your change anyway, but I just thought I'd ask while we had it out.)

wendigo:
@mholt: sure, check it now - now it's type agnostic :)

PR #1278, opened 2016-11-29
Fix for #1276 - support integers and floats as metadata in markdown

wendigo:
Adds support for all types of data in markdown


wendigo:
Waiting for review @mholt :)

caddyhttp/markdown/metadata/metadata.go
```diff
@@ -68,7 +68,7 @@ func (m *Metadata) load(parsedMap map[string]interface{}) {
 		switch v := val.(type) {
 		case bool:
 			m.Flags[key] = v
-		case string:
+		default:
```
mholt: This is great. Do you know why we separate Flags and Variables? (/cc @abiosoft) - do they need to stay separate?

wendigo: @mholt: I think they have to: you can access variables through ```{{ .Doc.Name }}``` and flags through ```{{ .DocFlags.Name }}```. I could of course change it so flags and variables will be accessible through ```{{ .Doc.Name }}``` only but it will break for existing clients.

mholt:
I dunno why they need to be separate anymore other than that compatibility,
let's combine them. I can mention the change in the change log.

wendigo:
@mholt: I've removed DocFlags entirely

PR #1284, opened 2016-12-03
Add send_timeout property to fastcgi directive

ericdreeves:
See #1094.

This update adds the send_timeout property to the fastcgi directive.

	fastcgi / 127.0.0.1:9000 {
		send_timeout 10s
	}

Like the connect_timeout and read_timeout properties, send_timeout has a default value of 60 seconds.

I added/refactored some tests along the way.

### DIFF
--- a/caddyhttp/fastcgi/dialer.go
+++ b/caddyhttp/fastcgi/dialer.go
@@ -19,8 +19,11 @@ type basicDialer struct {
 	timeout time.Duration
 }
 
-func (b basicDialer) Dial() (Client, error) { return Dial(b.network, b.address, b.timeout) }
-func (b basicDialer) Close(c Client) error  { return c.Close() }
+func (b basicDialer) Dial() (Client, error) {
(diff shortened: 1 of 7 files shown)


mholt:
@ericdreeves Can you clarify for me (I'm writing the docs) -- does read_timeout and send_timeout count against reading or sending the whole response/request or is it just each successive read/send call (like nginx)?

ericdreeves:
@mholt They count against the request/response.

mholt:
@ericdreeves Thanks. Could you suggest a better way to word this than I have, then, in the [docs](https://caddyserver.com/docs/fastcgi)?

ericdreeves:
@mholt Sorry about the delay. Busy week! How about:

read_timeout is the time allowed to read a response from the backend.
send_timeout is the time allowed to send a request to the backend.

mholt:
Works for me. Updating now, thanks!

Issue #1282, opened 2016-12-02
Lazy gzip/brotli


wendigo:
As mentioned in https://forum.caddyserver.com/t/lazy-gzip-possibly-a-feature-request/1031 it would be nice for Caddy to support pre-gzipped files.

PR #1289, opened 2016-12-05
Feature #1282 - Support serving statically compressed .gz and .br files

wendigo:
This change adds support for serving pre-compressed files (```.gz``` for ```gzip``` encoding and ```.br``` for ```brotli```).

How it works:

If ```file.x``` is requested and request's ```Accept-Encoding``` contains ```br```, try to serve ```file.x.br``` instead if it exists. ```Vary: Content-Encoding``` header and ```Content-Encoding: br``` headers are then added to response.

If ```file.x``` is requested and request's ```Accept-Encoding``` contains ```gzip```, try to serve ```file.x.gz``` instead if it exists. ```Vary: Content-Encoding``` header and ```Content-Encoding: gzip``` headers are then added to response.

This works for index files as well 😁

It also modifies ```gzip plugin``` so it no longer removes ```Accept-Encoding``` header from requests when gzip is enabled. This allows various other plugins in the chain (like fastcgi or proxy) to return compressed responses that will be omitted entirely by gzip plugin (backend can emit already compressed files).

Tip 1.: when manually testing gzip files with ```gzip -k filename```
Tip 2.: when using git plugin on linux all files can be gzipped using ```then_long find ROOT -type f -exec gzip -kf {} ;```


wendigo:
Btw. test cases in ```fileserver_test.go``` were broken.

rickb777:
Please don't merge this. The algorithm for matching Content-Encoding is inadequate and the response headers are not all set correctly (Vary is missing for example).

wendigo:
@rick777: Vary is added. Please clarify what do you mean by "algoritm for matching..." so I can fix that :)

wendigo:
P.S. Appveyor is broken due to: ```remote: Daily bandwidth rate limit exceeded for IP 74.205.54.20```

wendigo:
Waiting for review & thoughts @mholt

mholt:
I will look at it -- school is ending, gimme a few days, sorry. :)

caddyhttp/gzip/responsefilter.go
```diff
@@ -25,6 +25,14 @@ func (l LengthFilter) ShouldCompress(w http.ResponseWriter) bool {
 	return l != 0 && int64(l) <= length
 }
 
+// SkipGzipped is ResponseFilter that will discard already gzipped files
```
mholt: Technically, it will skip compressing responses that have a non-empty Content-Encoding header; it could be any value...

caddyhttp/gzip/setup.go
```diff
@@ -106,6 +106,8 @@ func gzipParse(c *caddy.Controller) ([]Config, error) {
 			config.RequestFilters = append(config.RequestFilters, DefaultExtFilter())
 		}
 
+		config.ResponseFilters = append(config.ResponseFilters, SkipGzipped{})
```
mholt: This is extremely important now that the Accept-Encoding header stays in the request, so add a test to ensure this response filter is always added.

caddyhttp/staticfiles/fileserver.go
```diff
@@ -121,13 +125,36 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
+
+				stat, err := gf.Stat()
+				if err == nil {
+					d = stat
+					f = gf
+
+					w.Header().Add("Vary", "Accept-Encoding")
+					w.Header().Set("Content-Encoding", encoding)
+
+					break encodings
```
mholt: Did you intend for this to be a labeled break? I guess I like that it's readable, but it's not a nested loop.

caddyhttp/staticfiles/fileserver.go
```diff
@@ -121,13 +125,36 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
+encodings:
+	for _, encoding := range staticEncodingPriority {
+		if strings.Contains(r.Header.Get("Accept-Encoding"), encoding) {
+			gf, err := fs.Root.Open(location + staticEncoding[encoding])
+			if err == nil {
+				defer gf.Close()
+
+				stat, err := gf.Stat()
+				if err == nil {
+					d = stat
```
mholt: Is `filename` updated properly in this case, when `d` is re-assigned?

mholt:
@rickb777 Have your concerns been addressed, by having the Vary header added?

caddyhttp/gzip/responsefilter.go
wendigo: Yeah, I've added brotli support after initialy adding gzip support only. Name is misleading. I will change that :)

caddyhttp/staticfiles/fileserver.go
wendigo: It don't need to. Filename is only needed so the ```http.ServeContent``` will return valid mime-type - it has to point to original filename, not compressed one :)
So: ```http.ServeContent(w, r, "index.html", compressedIndex.ModTime(), compressedIndexFile)``` is intended :)

caddyhttp/staticfiles/fileserver.go
wendigo: You're right - it's redundant.

caddyhttp/gzip/responsefilter.go
wendigo: I've fixed this filter so it now checks if response is compressed

wendigo:
Any more thoughts @mholt ?

caddyhttp/staticfiles/fileserver.go
```diff
@@ -121,13 +125,35 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
 		return http.StatusNotFound, nil
 	}
 
+	filename := d.Name()
+
+	for _, encoding := range staticEncodingPriority {
+		if strings.Contains(r.Header.Get("Accept-Encoding"), encoding) {
+			gf, err := fs.Root.Open(location + staticEncoding[encoding])
+			if err == nil {
+				defer gf.Close()
```
mholt: I think this defer could potentially leak file descriptors. If a file takes a long time to download (network congestion or client leaves connection open or something), this handler may never terminate and the loop that we're in could open multiple files that never end up being used.

caddyhttp/staticfiles/fileserver.go
mholt: Ah, indeed

nhooyr:
@mholt what do you think of gzipping files that can be gzipped (json/css/html/javascript) and storing them in memory and serving from there? This won't pollute the file system.

caddyhttp/staticfiles/fileserver.go
wendigo: What do you suggest then?

wendigo:
@nhooyr: I'm personally against - it will require synchronization between original and compressed image (in memory) - so invalidation of cache, memory hard limits and other things. There is also no big difference between serving file from memory and from disk - kernel will cache recently accessed files anyway.

nhooyr:
@wendigo I didn't mean that we should do it because of the minimal performance gains, it's just that it's a lot easier as you don't have to manually gzip the files manually and pollute your file system. You can just have caddy do it and update them whenever necessary.

caddyhttp/staticfiles/fileserver.go
wendigo: Ok, I see that now - I'm calling defer inside loop - I've fixed that :)

wendigo:
@nhooyr you don't have to manually gzip files - a lot of javascript builders (like webpack) can output gzipped files for you. This change is about serving pre-compressed file if it exists, not compression per se.

caddyhttp/gzip/setup.go
wendigo: Test added

caddyhttp/staticfiles/fileserver.go
```diff
@@ -121,13 +125,41 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
+	for _, encoding := range staticEncodingPriority {
+		if strings.Contains(r.Header.Get("Accept-Encoding"), encoding) {
+			gf, err := fs.Root.Open(location + staticEncoding[encoding])
+
+			if err == nil {
+				stat, err := gf.Stat()
+
+				if err == nil {
+					// Close previous file - release fd
+					f.Close()
```
mholt: Is closing `f` a second time problematic? It's already closed above with a `defer`.

wendigo: I've checked and in case that file was previously closed ```Close()``` will return error (```ErrInvalid  - invalid argument```). As we don't check ```Close()``` return error calling it is no-op.

lhecker: @wendigo @mholt You could also replace [this line](https://github.com/mholt/caddy/blob/cffeb22f522dd9488cf21a2ec423ae079bb76dd2/caddyhttp/staticfiles/fileserver.go#L69) with the following:

```go
defer func() { _ = f.Close() }()
```

That way no file handle is closed twice...
...as long as you remember to always close `f` before assigning it a new value, but considering how "dangerous" Go is anyways already I'd consider such a requirement quite normal.

wendigo: https://github.com/golang/go/blob/master/src/os/file_unix.go#L128

lhecker: @wendigo Did you mean [this](https://github.com/golang/go/blob/master/src/os/file_unix.go#L143) line? That would generate a EINVAL error. Otherwise I wouldn't know how the `*File` pointer would be nil.

Well either way it's nearly a no-op. In that case I'm sorry for "accusing" you. I know that my opinion doesn't "really" count since I'm neither maintainer nor contributor of this PR, but in the end I'd still say that the critique of the double-close is valid -  on one hand because defer is on the heavy side of inbuilt operations and on the other because it appears to be cleanlier in my opinion if you reduce the amount of duplicate (apparent) work. Kinda like how you use bounded channels in Go instead of unbounded ones.

caddyhttp/staticfiles/fileserver.go
```diff
@@ -121,13 +125,41 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
+					f = gf
+
+					w.Header().Add("Vary", "Accept-Encoding")
+					w.Header().Set("Content-Encoding", encoding)
+
+					defer f.Close()
+					break
+				}
+
+				gf.Close()
```
mholt: This loop is constructed a bit "deep" in my opinion. Can we do this instead, I find this much easier to follow (untested, please try it out if you like it!):

```go
// see if there is a compressed version of the file available; if so, use it
for _, encoding := range staticEncodingPriority {
	if !strings.Contains(r.Header.Get("Accept-Encoding"), encoding) {
		continue
	}
	encodedFile, err := fs.Root.Open(location + staticEncoding[encoding])
	if err != nil {
		encodedFile.Close()
		continue
	}
	encodedFileInfo, err := encodedFile.Stat()
	if err != nil {
		encodedFile.Close()
		continue
	}
	
	f.Close() // release fd of previously-selected file
	f = encodedFile
	d = encodedFileInfo

	w.Header().Set("Content-Encoding", encoding)

	defer f.Close()
	break
}
```

caddyhttp/staticfiles/fileserver.go
mholt: Well, I'd rather have one too many calls to `Close()` than one too few. We can clean up things as needed later (as long as we have sufficient tests)

caddyhttp/staticfiles/fileserver.go
```diff
@@ -168,3 +200,13 @@ var IndexPages = []string{
...
+
+var staticEncoding = map[string]string{
+	"gzip": ".gz",
+	"br":   ".br",
+}
+
+var staticEncodingPriority = []string{
+	"br",
+	"gzip",
+}
```
mholt: Can you comment on what these variables do? godoc style (I know they're unexported, but it's not obvious what they're used for, since this isn't the compression module.)

caddyhttp/staticfiles/fileserver.go
wendigo: Yeah, it's easier to follow 👍 First ```encodedFile.Close()``` is not needed as Open resulted in error :)

wendigo: I refactored also one extra loop - when iterating over ```IndexPages``` - same comments apply

caddyhttp/staticfiles/fileserver.go
wendigo: @lhecker totally agree. By writing "no-op" I meant "leaving *File in the same, terminal state". As we are not checking ```Close()``` error anyway calling ```Close()``` multiple times has no side effects.

caddyhttp/staticfiles/fileserver.go
mholt: Good point about the first Close(), thanks.

mholt:
Thank you very much for your work, Mateusz! :)

wendigo:
Thx! :)

Issue #1246, opened 2016-11-09
syslog to external server


myReaper:
Hi,

This is basically a feature request :-)

Currently I'm missing an option to output error and access logs directly to our centralized syslog server, or at least I haven't found it :-)

Nginx for example has an option to directly log to a syslog server via ip and port:
https://nginx.org/en/docs/syslog.html

The main benefit of this is:
- Using a centralized syslog server (of course :-)) for multiple instances to easier monitor them
- Optimize I/O on the Caddy/Webserver as no local files need to get written anymore

Right now, we are running most sites in development or low usage systems but we'd like to move caddy to our production environment in about 2-3 weeks, would be great to have this implemented "soon" [can't really estimate the work amount for this as I'm no coder :( ]

btw. great job on caddy, I'm using it already on almost all of my private sites for a few months now and it works super smooth :-)

Best Regards
Michael

wendigo:
Would you like to stream access/errors log to syslog or every log message?

myReaper:
Sorry for my late reply :-)

Right now, streaming access and error log to an remote syslog server currently would be sufficient for us. Of course, receiving all messages would also not be wrong.

Our main goal is to have all caddy log messages available in our central Logstash installation. So the best option for us would be to directly output error/access logs as json to the logstash server.

However, the quicker integration option into caddy might be to "simply" stream current error and {combined] access log format to a remote syslog (which would also be ok for us as we could parse this in Logstash then).

mholt:
You can specify `syslog` for error and access logs (using the `errors` and `log` directives) already. Does that meet your needs?

myReaper:
@mholt no, just specifying syslog will log to the internal syslog, we require caddy to log to an external syslog server (using ip:port) like nginx offers for example.

The local log should not be written at all, also in terms of disk i/o performance.

wendigo:
Can you use relay to bounce logs to an external syslog using local one?

myReaper:
@wendigo for performance reasons it would be good to have caddy directly log to an external syslog server (or send json to logstash) instead of first writing to local syslog and then relaying using the local syslog server.

wendigo:
@myReaper as Caddy is using https://github.com/hashicorp/go-syslog syslog wrapper I've submitted PR to allow dialling remote syslog server https://github.com/hashicorp/go-syslog/pull/5/files. When it will be merged I can add support for remote syslog in Caddy.

[...]

PR #1301, opened 2016-12-18
Feature #1246 - Remote syslog

wendigo:
This change introduces ability to log to remote syslog server.

I've made small refactoring to share code between ```errors``` and ```log``` plugins.


mholt:
I'll look at this soon, very swamped with things!

wendigo:
Fully understand :)

mholt:
I'm still busy... sorry. Would somebody else like to review this if they have a chance before I do?

caddyhttp/httpserver/logger.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+)
+
+var remoteSyslogPrefixes = map[string]string{
+	"syslog+tcp://": "tcp",
+	"syslog+udp://": "udp",
+	"syslog://":     "udp",
+}
+
+type Logger struct {
```
mholt: Check the golint output -- will need godoc on the exported things. ;)

caddyhttp/httpserver/logger.go
```diff
@@ -0,0 +1,117 @@
...
+	Output string
+	*log.Logger
+	Roller *LogRoller
+	writer io.Writer
+}
+
+func NewTestLogger(buffer *bytes.Buffer) *Logger {
```
abiosoft: This will also need godoc as long as it is not in a `_test.go` file.

mholt:
Fabulous. Thank you @wendigo! Before I click Merge, can you provide some docs that I can put on the website? Either a PR on the caddyserver.com repo or just a comment here explaining precisely how it is used would be fine. I want to be sure I don't miss anything.

mholt:
@wendigo I imagine you're pretty busy. When do you think you'll be able to document this? (Again, it doesn't have to go into the site directly, just write something up for us to get us started.)

wendigo:
Yeah, I'll write some docs @mholt

wendigo:
Hi @mholt!
Changes in usage are as follow:

both `log` and `errors` adds support for following log "locations":

```syslog``` - logs to local syslog
```syslog://host[:port]``` - logs via UDP to local or remote syslog server
```syslog+udp://host[:port]``` - logs via UDP to local or remote syslog server
```syslog+tcp://host[:port]``` - logs via TCP to local or remote syslog server

Final syntax is i.e.:

```log syslog://remotehost:442 {``` and
```errors syslog+udp://remotehost:442 {``` and
```
errors {
  log syslog://remotehost:442
}
```

Basically - everywhere where ```[logfile]``` can be passed proper syslog scheme can be :)

wendigo:
I've also merged your change with RWMutex

wendigo:
AppVeyor failed on `test -race TestStandaloneTLSTicketKeyRotation`

tobya:
@wendigo Sorry to ask a silly question but did you merge the doc changes into the docs?  

If not I could look at creating the changes to docs for caddyserver.com based on your notes, unless someone beats me to it.

mholt:
I was also just looking at this now that I'm catching up with life a bit :) If someone can prepare a PR for the docs on the website that would be splendid, and we'll merge this ASAP.

mholt:
Thanks for your patience and effort @wendigo!

tobya:
@wendigo @mholt syslog doesnt seem to be implemented on windows

https://golang.org/pkg/log/syslog/#pkg-note-BUG

I cant build.  If I got some suggestions around what to do (eg an if block, use another library etc) I could take a look, as is however though I think this won't build on windows (unless I'm missing something)

http://stackoverflow.com/questions/42083059/getting-syslog-writer-undefined

wendigo:
Hi, I'm using https://github.com/hashicorp/go-syslog?files=1 which wrapps suslog using build tags

wendigo:
Did you fetch it?

tobya:
I just built it.  Should I fetch separately?

tobya:
@wendigo Thank you I assumed go build would fetch latest but obviously not.  Builds fine now.

wendigo:
@tobya good to hear that. Thx!

mholt:
Phew, you guys scared me for a while. 😅

PR #1309, opened 2016-12-26
Fixed #1292 and resulting issues from #1300

### LINKED ISSUE #1292: websocket connect failure over https proxy

lhecker:
@mholt We already talked in length about the contents of this PR and here it is!

I tried to split up this PR in 4 commits to clearly separate the fixes. The commit messages probably say it all. 🙂

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -205,10 +205,6 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request, 
 	}
 
 	rp.Director(outreq)
-	outreq.Proto = "HTTP/1.1"
-	outreq.ProtoMajor = 1
-	outreq.ProtoMinor = 1
-	outreq.Close = false
 
 	res, err := transport.RoundTrip(outreq)
 	if err != nil {
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -148,7 +148,9 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 		} else {
 			transport.MaxIdleConnsPerHost = keepalive
 		}
-		http2.ConfigureTransport(transport)
+		if httpserver.HTTP2 {
+			http2.ConfigureTransport(transport)
+		}
 		rp.Transport = transport
 	}
 	return rp
@@ -168,10 +170,15 @@ func (rp *ReverseProxy) UseInsecureTransport() {
 			TLSHandshakeTimeout: 10 * time.Second,
 			TLSClientConfig:     &tls.Config{InsecureSkipVerify: true},
 		}
-		http2.ConfigureTransport(transport)
+		if httpserver.HTTP2 {
+			http2.ConfigureTransport(transport)
+		}
 		rp.Transport = transport
 	} else if transport, ok := rp.Transport.(*http.Transport); ok {
 		transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+		// No http2.ConfigureTransport() here.
+		// For now this is only added in places where
+		// an http.Transport is actually created.
 	}
 }
 
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -211,10 +211,27 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request, 
 		return err
 	}
 
+	isWebsocket := res.StatusCode == http.StatusSwitchingProtocols && strings.ToLower(res.Header.Get("Upgrade")) == "websocket"
+
+	// Remove hop-by-hop headers listed in the
+	// "Connection" header of the response.
+	if c := res.Header.Get("Connection"); c != "" {
+		for _, f := range strings.Split(c, ",") {
+			if f = strings.TrimSpace(f); f != "" {
+				res.Header.Del(f)
+			}
+		}
+	}
+
+	for _, h := range hopHeaders {
+		res.Header.Del(h)
+	}
+
 	if respUpdateFn != nil {
 		respUpdateFn(res)
 	}
-	if res.StatusCode == http.StatusSwitchingProtocols && strings.ToLower(res.Header.Get("Upgrade")) == "websocket" {
+
+	if isWebsocket {
 		res.Body.Close()
 		hj, ok := rw.(http.Hijacker)
 		if !ok {
@@ -246,13 +263,30 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request, 
 		go pooledIoCopy(backendConn, conn) // write tcp stream to backend
 		pooledIoCopy(conn, backendConn)    // read tcp stream from backend
 	} else {
-		defer res.Body.Close()
-		for _, h := range hopHeaders {
-			res.Header.Del(h)
-		}
 		copyHeader(rw.Header(), res.Header)
+
+		// The "Trailer" header isn't included in the Transport's response,
+		// at least for *http.Transport. Build it up from Trailer.
+		if len(res.Trailer) > 0 {
+			trailerKeys := make([]string, 0, len(res.Trailer))
+			for k := range res.Trailer {
+				trailerKeys = append(trailerKeys, k)
+			}
+			rw.Header().Add("Trailer", strings.Join(trailerKeys, ", "))
+		}
+
 		rw.WriteHeader(res.StatusCode)
+		if len(res.Trailer) > 0 {
+			// Force chunking if we saw a response trailer.
+			// This prevents net/http from calculating the length for short
+			// bodies and adding a Content-Length.
+			if fl, ok := rw.(http.Flusher); ok {
+				fl.Flush()
+			}
+		}
 		rp.copyResponse(rw, res.Body)
+		res.Body.Close() // close now, instead of defer, to populate res.Trailer
+		copyHeader(rw.Header(), res.Trailer)
 	}
 
 	return nil
@@ -305,16 +339,17 @@ func copyHeader(dst, src http.Header) {
 // Hop-by-hop headers. These are removed when sent to the backend.
 // http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html
 var hopHeaders = []string{
+	"Alt-Svc",
+	"Alternate-Protocol",
 	"Connection",
 	"Keep-Alive",
 	"Proxy-Authenticate",
 	"Proxy-Authorization",
-	"Te", // canonicalized version of "TE"
-	"Trailers",
+	"Proxy-Connection", // non-standard but still sent by libcurl and rejected by e.g. google
+	"Te",               // canonicalized version of "TE"
+	"Trailer",          // not Trailers per URL above; http://www.rfc-editor.org/errata_search.php?eid=4522
 	"Transfer-Encoding",
 	"Upgrade",
-	"Alternate-Protocol",
-	"Alt-Svc",
 }
 
 type respUpdateFn func(resp *http.Response)
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -253,7 +253,7 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request, 
 }
 
 func (rp *ReverseProxy) copyResponse(dst io.Writer, src io.Reader) {
-	buf := bufferPool.Get()
+	buf := bufferPool.Get().([]byte)
 	defer bufferPool.Put(buf)
 
 	if rp.FlushInterval != 0 {
@@ -268,7 +268,10 @@ func (rp *ReverseProxy) copyResponse(dst io.Writer, src io.Reader) {
 			dst = mlw
 		}
 	}
-	io.CopyBuffer(dst, src, buf.([]byte))
+
+	// `CopyBuffer` only uses `buf` up to it's length and
+	// panics if it's 0 => Extend it's length up to it's capacity.
+	io.CopyBuffer(dst, src, buf[:cap(buf)])
 }
 
 // skip these headers if they already exist.
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -349,9 +349,14 @@ func newConnHijackerTransport(base http.RoundTripper) *connHijackerTransport {
 		MaxIdleConnsPerHost: -1,
 	}
 	if b, _ := base.(*http.Transport); b != nil {
+		tlsClientConfig := b.TLSClientConfig
+		if tlsClientConfig.NextProtos != nil {
+			tlsClientConfig = cloneTLSClientConfig(tlsClientConfig)
+			tlsClientConfig.NextProtos = nil
+		}
+
 		t.Proxy = b.Proxy
-		t.TLSClientConfig = cloneTLSClientConfig(b.TLSClientConfig)
-		t.TLSClientConfig.NextProtos = nil
+		t.TLSClientConfig = tlsClientConfig
 		t.TLSHandshakeTimeout = b.TLSHandshakeTimeout
 		t.Dial = b.Dial
 		t.DialTLS = b.DialTLS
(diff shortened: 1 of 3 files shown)


caddyhttp/proxy/reverseproxy.go
```diff
@@ -261,7 +267,10 @@ func (rp *ReverseProxy) copyResponse(dst io.Writer, src io.Reader) {
 			dst = mlw
 		}
 	}
-	io.CopyBuffer(dst, src, buf.([]byte))
+
+	// `CopyBuffer` only uses `buf` up to it's length and
+	// panics if it's 0 => Extend it's length up to it's capacity.
```
mholt: This comment is hard to understand. Can it be phrased a little more clearly (and fix the gammar, haha)? Maybe something like:

> CopyBuffer only uses buf up to its length and panics if it's 0, so we extend its length up to its capacity.

caddyhttp/proxy/reverseproxy.go
```diff
@@ -331,51 +340,148 @@ type connHijackerTransport struct {
...
+			err := tlsConn.Handshake()
+			if timer != nil {
+				timer.Stop()
+			}
+			errc <- err
+		}()
+		if err := <-errc; err != nil {
+			plainConn.Close()
+			return nil, err
+		}
```
mholt: `errc` can buffer 2 values but only 1 is ever consumed. And only the plainConn is closed if there's an error. Is that expected?

caddyhttp/proxy/reverseproxy.go
```diff
@@ -331,51 +340,148 @@ type connHijackerTransport struct {
...
+		if !t.TLSClientConfig.InsecureSkipVerify {
+			serverName := t.TLSClientConfig.ServerName
+			if serverName == "" {
+				serverName = addr
+				idx := strings.LastIndex(serverName, ":")
+				if idx != -1 {
+					serverName = serverName[:idx]
+				}
```
mholt: Use `net.SplitHostPort()` instead.

caddyhttp/proxy/reverseproxy.go
```diff
@@ -331,51 +340,148 @@ type connHijackerTransport struct {
...
 	}
-	if base != nil {
-		if baseTransport, ok := base.(*http.Transport); ok {
-			transport.Proxy = baseTransport.Proxy
-			transport.TLSClientConfig = baseTransport.TLSClientConfig
-			transport.TLSHandshakeTimeout = baseTransport.TLSHandshakeTimeout
-			transport.Dial = baseTransport.Dial
-			transport.DialTLS = baseTransport.DialTLS
-			transport.MaxIdleConnsPerHost = -1
+	if b, _ := base.(*http.Transport); b != nil {
```
mholt: This is interesting; is there a specific reason you're not using the `ok` (the second value, which is currently `_`) for the check here?

lhecker: Nice catch! If `ok` is false (and the cast thus failed) `b` must be `nil` right? (As per Go spec.)

So if we have this code:
```go
if base != nil {
    if b, ok := base.(*http.Transport); ok == true {
        ...
    }
}
```

It's the same as this:
```go
if b, _ := base.(*http.Transport); b != nil {
}
```

because if `base` is `nil` it stays `nil` and if it's not a `*http.Transport` it will be converted to `nil`. I usually use this coding pattern in Go if I want to write sth. along the former snippet. Should I revert that or add a comment?

caddyhttp/proxy/reverseproxy.go
lhecker: I copied this straight out of the Go source code from [here](https://github.com/golang/go/blob/230a376b5a67f0e9341e1fa47e670ff762213c83/src/net/http/transport.go#L1017-L1052) and thus didn't dare to fiddle with this rather sensitive piece of code.
I forgot to add a comment with the above URL though. I'll add one tomorrow.

`errc` probably has a capacity of 2 to ensure that the 2 concurrent operations (handshake and timeout) will never block while sending (and thus finishing the Goroutine as fast as possible), especially since the (memory) overhead of a small buffered channel is neglectible anyways.
I'm not entirely sure why the call to `tlsConn.Close()` is missing, but that's what every Go 1.7 app is running on so I guess what they're doing there is fine.

caddyhttp/proxy/reverseproxy.go
lhecker: Okay... [That's quite a method though.](https://github.com/golang/go/blob/866e01457f69b58b31ccb95f223aac80e1285332/src/net/ipsock.go#L107-L163) The reason I wrote it myself is because I had to replace the call to `cm.tlsHost()` in the original Go source, which also didn't use `net.SplitHostPort()` for some reason (performance maybe?). It's written like [this](https://github.com/golang/go/blob/230a376b5a67f0e9341e1fa47e670ff762213c83/src/net/http/transport.go#L1183-L1187) and my code was basically a inlined version of that.

caddyhttp/proxy/reverseproxy.go
lhecker: Ah... I always forget about "its" because I pronounce it the same as "it's".

> CopyBuffer only uses buf up to its length and panics if it's 0. Due to that we extend buf's length to its capacity here and ensure it's always non-zero.

Is that fine too?

lhecker:
> And you've confirmed it fixes your use case as described in #1292 and #1300?

Yeah passing HTTP/2 requests and WebSockets generally works flawless now. I tested it using self signed certs and Go 1.8 beta 2.

<hr/>

I'm still having issues with my gRPC backend though, because requests and responses seem to buffered by caddy somehow. Maybe they're even dropped - I've got to properly investigate that first. Due to that my gRPC server seems to only receive the first and last stream chunks. Any idea why that could be the case?
(I don't think it's because of the missing support for HTTP/2 push, because in my case it's a client sending a stream to the server!)

caddyhttp/proxy/reverseproxy.go
mholt: Yep!

caddyhttp/proxy/reverseproxy.go
mholt: I don't know, that's a good question -- SplitHostPort handles IPv6 addresses. If this function doesn't see IPv6 addresses then I imagine it's fine as-is.

caddyhttp/proxy/reverseproxy.go
lhecker: I'm not sure if I understand… Do you agree that the code is fine? (My English skills are seriously lacking lately 😩)

Basically what I said is that the first code snipped (the old code) above is the same as:
```go
if b, ok := base.(*http.Transport); b != nil && ok == true {
}
```

But since `b` being non-nil **must** imply that `ok` is true we can drop that check and get the latter snipped (the new code). You can try that in your Playground by making it return e.g. an `&sync.Mutex{}` instead.

mholt: Sorry! Your English is actually really good. Yes, you are right. Sorry for the confusion. Been a long day.

caddyhttp/proxy/reverseproxy.go
mholt: It's probably fine as-is then. Didn't realize this came from the std lib. (Wonder if we should just put a comment in saying that we've adapted this code from the standard library and modified it a bit.)

mholt:
> I'm still having issues with my gRPC backend though, because requests and responses seem to buffered by caddy somehow. Maybe they're even dropped - I've got to properly investigate that first. Due to that my gRPC server seems to only receive the first and last stream chunks. Any idea why that could be the case?

Nope :( Although I wonder if there is something up with the buffering somewhere in the proxy. For example, in #1310 there is a report of high memory use when proxying for big files.

caddyhttp/proxy/reverseproxy.go
lhecker: You where right and the function was broken. I forgot about IPv6 addresses.
I fixed this problem and moved the code to a `stripPort()` method which is basically 4 LOC, easy to understand and makes the code quite clean now IMO.

lhecker:
@sanderhahn I was able to reproduce your problem. This only happens if you don't specify `insecure_skip_verify`.

@mholt Due to this I'd consider this PR incomplete. I'll try to fix this but I'm not entirely sure yet where to start, since unlike before this does not cause any debug/error output to be generated by Go.

mholt:
@lhecker I'm not entirely clear on what the error is, but maybe dumping the HTTP request would be a helpful start? See https://golang.org/pkg/net/http/httputil/.

lhecker:
@mholt I actually found the root cause again only minutes after writing the comment. It's because I only add the `DialTLS` method to the transport if a `transport.TLSClientConfig` is provided, but that's not the case for verified TLS connections (i.e. `transport.TLSClientConfig` is `nil`).

So… I fixed that, but now to the next issue:
```
2016/12/28 15:59:35 http: TLS handshake error from [::1]:2816: EOF
```

_sigh…_

lhecker:
AHA! I forgot to add the `errors stderr` directive to that particular domain I'm testing on. I can't claim that I particularly like that behavior (swallowing errors _by default_) and IMHO you might want to consider changing it to `stderr` as the safer default.

Anyways: Fix inbound.

mholt:
@lhecker Nice work :) Sorry about the default of hiding errors. We can probably open a discussion on that in a new issue if you want.

lhecker:
@sanderhahn Do you mind trying out the newest commit (cd8a992)?

lhecker:
@mholt I found the cause for #1310 and I think I fixed it for the most part. Unfortunately I'm still experiencing issues with my gRPC backend so I can't be fully sure yet. Do you mind if I attach the fix for that to this PR in case I finish with it before you merge this?

It's caused by [this line](https://github.com/mholt/caddy/blob/d5fe4928f284b5d10c0bcebaaa2b27d641e058c9/caddyhttp/proxy/proxy.go#L98). I fixed it by introducing a special case where the request body is not buffered if `len(proxy.Upstreams)` is 1 or less. If only one upstream exists we don't ever have to rewind the body and thus don't have to cache it right? That should really help improve performance for the common proxy case!

sanderhahn:
Additional note is that the -http2=false flag doesn't seem to be necessary anymore. The proxy works both with and without http2 enabled.

mholt:
@lhecker Maybe make a new PR for the gRPC issue? I'll probably merge this soon unless somebody tells me to hold the presses.

Let's make you a collaborator on the project too so you can have push access. This way you can work in branches instead of forks and also help by reviewing and merging pull requests, etc. I really appreciate your efforts, you've done a great job!

@sanderhahn Thanks for testing it!

mholt:
Oh, missed the second half of your comment @lhecker:

> It's caused by this line. I fixed it by introducing a special case where the request body is not buffered if len(proxy.Upstreams) is 1 or less. If only one upstream exists we don't ever have to rewind the body and thus don't have to cache it right? That should really help improve performance (especially latency obviously) for the common proxy case!

I had never considered that. I guess not, then. I don't see where or why else we're using `body` other than to replay it for the next upstream.

I wonder if buffering has been causing some of other proxy-related issues too. It'd be interesting to see if their issues are fixed with that change too.

Let me know when all the changes for this PR are pushed. :)

lhecker:
> Let's make you a collaborator on the project too so you can have push access. This way you can work in branches instead of forks and also help by reviewing and merging pull requests, etc. I really appreciate your efforts, you've done a great job!

Thanks! I probably won't be able to contribute much, because most of the current work resulted from me needing a viable nginx alternative as the frontend for a commercial product with HTTP/2 upstream support while being easy to mod. Caddy turned out to not be quite bug free yet so I simply contributed my fixes. It's quid pro quo right? 😅
Oh and I'll of course promise to handle in good faith with my new access rights to this repo though! And if I ever have a fix for caddy in the future I'll make sure to create a PR. 🙂

> Let me know when all the changes for this PR are pushed. :)

Since @sanderhahn confirmed that it's working for him too I'd consider this PR to be finished and you can merge it whenever you see fit. This PR turned out to be mostly a fix for #1292, but since the issues from #1300 are quite closely related I think it's okay we squashed both together.

lhecker:
@mholt I pushed one further change just now. Please review it carefully, because I don't know wether it has any implications:
I looked into the git history of those 4 removed lines in 533039e but never found a explanation as to why they where added originally. Well apart from `outreq.Close = false` which can be removed without any concerns, because `false` is the default anyways.

But I found that the removal of those 4 lines has no side effects for my sites and seemed to be essential to make gRPC proxying work properly.

mholt:
@lhecker We added those lines originally because we borrowed the ServeHTTP for ReverseProxy in https://golang.org/src/net/http/httputil/reverseproxy.go?s=3924:3999#L134 of the Go standard library.

It's quite possible, given all the handling/configuration of transports and directors and such in our case that we don't need to set these there.

But it's good to hear that's what makes gRPC work! And the tests still pass, so that's promising. I'll take a look.

Others are welcome/invited to review carefully too. @nemothekid - hope you're having a happy holiday, I dunno where you're at but maybe you'd have a chance to look this over and see if you notice anything?

lhecker:
@mholt I found one more bug… 😕 You're not handling any trailers in the proxy! Seeing as trailers are gaining quite some interest with HTTP/2 I'll update this PR with a fix soon.

lhecker:
@mholt I pushed the support for HTTP trailers just now. As usual I made sure to add tests (I extended `TestReverseProxy`).
Again: Without the fix it doesn't work at all, but after applying it it works flawless (trailers in general as well as proxying gRPC metadata, which partially relies on it). I also tested everything on my staging system (basic REST, WebSockets, gRPC).

The code stems from the current [`reverseproxy.go`](https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go) code just like it did before. The big difference is that the version in caddy was outdated, missing the trailer feature as well as containing a trailer-related bug (`"Trailers" != "Trailer"`).

mholt:
That's awesome, thanks @lhecker . Can't wait to take a look this weekend. :+1:

caddyhttp/proxy/reverseproxy.go
```diff
@@ -331,51 +376,169 @@ type connHijackerTransport struct {
...
+		ClientCAs:                   cfg.ClientCAs,
+		InsecureSkipVerify:          cfg.InsecureSkipVerify,
+		CipherSuites:                cfg.CipherSuites,
+		PreferServerCipherSuites:    cfg.PreferServerCipherSuites,
+		ClientSessionCache:          cfg.ClientSessionCache,
+		MinVersion:                  cfg.MinVersion,
+		MaxVersion:                  cfg.MaxVersion,
+		CurvePreferences:            cfg.CurvePreferences,
+		DynamicRecordSizingDisabled: cfg.DynamicRecordSizingDisabled,
+		Renegotiation:               cfg.Renegotiation,
```
mholt: Just as a note, I think this won't be needed in Go 1.8 as TLS configs have a Clone() method ^_^

mholt:
Thank you @lhecker!! You are welcome to hack on Caddy at any time. :smile:

caddyhttp/proxy/reverseproxy.go
lhecker: Ah cool. I'll take note of that and fix this after 1.8 is out.

I've seen a insane amount of code already which copies structs like `tls.Config` by value instead of creating a new proper clone and thus causing all sorts of very subtle issues by inproperly copying the embedded mutexes. Things like that really make me personally believe that Go might be our best tool but the worst language we've come up with. 😅

PR #1314, opened 2016-12-29
proxy: Unbuffered request optimization

### LINKED ISSUE #1310: High RAM usage when proxying WebDAV (& possibly normal traffic as well)
`Caddy 0.9.4`

Proxy an entire (sub)domain through to another server (which mostly handles WebDAV)

```text
example.com {
  header / -Server
  log / /srv/www/log/access.log "{host}: {scheme} request by {remote}: {method} {proto} {path} {query}" {
    rotate {
      size 256
      age 14
      keep 3
    }
  }
  errors {
    log /srv/www/log/error.log {
      size 256
      age 14
      keep 3
    }
  }
  gzip
  root /srv/www/dummy

  proxy / 10.10.10.10:81 {
    header_upstream Host {host}
    header_upstream X-Real-IP {remote}
    header_upstream X-Forwarded-For {remote}
    header_upstream X-Forwarded-Proto {scheme}
    header_upstream X-Frame-Options SAMEORIGIN
  }
}
```

I'm using systemd to start Caddy & the command the systemd unit executes is  
`/usr/bin/caddy -log stdout -agree=true -conf=/srv/www/Caddyfile -root=/var/tmp -email hostmaster@example.com -quic`

Caddy serving/proxying the service as usual

High RAM usage  
In Idle state, caddy uses about 100M of RAM,  
After starting a big WebDAV operation that was going through the proxy it jumped to about 1G,  
10 minutes later it was at 2G and shortly after the (successfully) proxied operation finished the memory usage went back down to about 100M again.

Although this can probably be reproduced easier, my setup was:  
nginx serving a (fresh) Nextcloud instance on 10.10.10.10:81 (lan) (http (no TLS)),  
Caddy proxying the connection as shown in the Caddyfile,  
Then i mounted the Nextcloud instance locally using WebDAV & pushed about 50G through (with a gbit uplink)  

IIRC the memory usage started jumping high after a few gigs already

lhecker:
> If only one upstream is defined we don't need to buffer the body.
> Instead we directly stream the body to the upstream host, which reduces memory usage as well as latency.
> Furthermore this enables different kinds of HTTP streaming applications like gRPC for instance.

This probably fixes #1310.

Again **NOTE:** This only works if your `proxy` directive specifies only a single upstream host.

### DIFF
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -39,6 +39,9 @@ type Upstream interface {
 	// Gets how long to wait between selecting upstream
 	// hosts in the case of cascading failures.
 	GetTryInterval() time.Duration
+
+	// Gets the number of upstream hosts.
+	GetHostCount() int
 }
 
 // UpstreamHostDownFunc can be used to customize how Down behaves.
@@ -94,13 +97,26 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	// outreq is the request that makes a roundtrip to the backend
 	outreq := createUpstreamRequest(r)
 
-	// record and replace outreq body
-	body, err := newBufferedBody(outreq.Body)
-	if err != nil {
-		return http.StatusBadRequest, errors.New("failed to read downstream request body")
-	}
-	if body != nil {
-		outreq.Body = body
+	// If we have more than one upstream host defined and if retrying is enabled
+	// by setting try_duration to a non-zero value, caddy will try to
+	// retry the request at a different host if the first one failed.
+	//
+	// This requires us to possibly rewind and replay the request body though,
+	// which in turn requires us to buffer the request body first.
+	//
+	// An unbuffered request is usually preferrable, because it reduces latency
+	// as well as memory usage. Furthermore it enables different kinds of
+	// HTTP streaming applications like gRPC for instance.
+	requiresBuffering := upstream.GetHostCount() > 1 && upstream.GetTryDuration() != 0
+
+	if requiresBuffering {
+		body, err := newBufferedBody(outreq.Body)
+		if err != nil {
+			return http.StatusBadRequest, errors.New("failed to read downstream request body")
+		}
+		if body != nil {
+			outreq.Body = body
+		}
 	}
 
 	// The keepRetrying function will return true if we should
@@ -173,15 +189,25 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 			downHeaderUpdateFn = createRespHeaderUpdateFn(host.DownstreamHeaders, replacer)
 		}
 
-		// rewind request body to its beginning
-		if err := body.rewind(); err != nil {
-			return http.StatusInternalServerError, errors.New("unable to rewind downstream request body")
+		// Before we retry the request we have to make sure
+		// that the body is rewound to it's beginning.
+		if bb, ok := outreq.Body.(*bufferedBody); ok {
+			if err := bb.rewind(); err != nil {
+				return http.StatusInternalServerError, errors.New("unable to rewind downstream request body")
+			}
 		}
 
 		// tell the proxy to serve the request
-		atomic.AddInt64(&host.Conns, 1)
-		backendErr = proxy.ServeHTTP(w, outreq, downHeaderUpdateFn)
-		atomic.AddInt64(&host.Conns, -1)
+		//
+		// NOTE:
+		//   The call to proxy.ServeHTTP can theoretically panic.
+		//   To prevent host.Conns from getting out-of-sync we thus have to
+		//   make sure that it's _always_ correctly decremented afterwards.
+		func() {
+			atomic.AddInt64(&host.Conns, 1)
+			defer atomic.AddInt64(&host.Conns, -1)
+			backendErr = proxy.ServeHTTP(w, outreq, downHeaderUpdateFn)
+		}()
 
 		// if no errors, we're done here
 		if backendErr == nil {
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -954,6 +954,90 @@ func TestReverseProxyRetry(t *testing.T) {
 	}
 }
 
+func TestReverseProxyLargeBody(t *testing.T) {
+	log.SetOutput(ioutil.Discard)
+	defer log.SetOutput(os.Stderr)
+
+	// set up proxy
+	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		io.Copy(ioutil.Discard, r.Body)
+		r.Body.Close()
+	}))
+	defer backend.Close()
+
+	su, err := NewStaticUpstreams(caddyfile.NewDispenser("Testfile", strings.NewReader(`proxy / `+backend.URL)))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p := &Proxy{
+		Next:      httpserver.EmptyNext, // prevents panic in some cases when test fails
+		Upstreams: su,
+	}
+
+	// middle is required to simulate closable downstream request body
+	middle := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err = p.ServeHTTP(w, r)
+		if err != nil {
+			t.Error(err)
+		}
+	}))
+	defer middle.Close()
+
+	// Our request body will be 100MB
+	bodySize := uint64(100 * 1000 * 1000)
+
+	// We want to see how much memory the proxy module requires for this request.
+	// So lets record the mem stats before we start it.
+	begMemstats := &runtime.MemStats{}
+	runtime.ReadMemStats(begMemstats)
+
+	r, err := http.NewRequest("POST", middle.URL, &noopReader{len: bodySize})
+	if err != nil {
+		t.Fatal(err)
+	}
+	resp, err := http.DefaultTransport.RoundTrip(r)
+	if err != nil {
+		t.Fatal(err)
+	}
+	resp.Body.Close()
+
+	// Finally we need the mem stats after the request is done...
+	endMemstats := &runtime.MemStats{}
+	runtime.ReadMemStats(endMemstats)
+
+	// ...to calculate the total amount of allocated memory during the request.
+	totalAlloc := endMemstats.TotalAlloc - begMemstats.TotalAlloc
+
+	// If that's as much as the size of the body itself it's a serious sign that the
+	// request was not "streamed" to the upstream without buffering it first.
+	if totalAlloc >= bodySize {
+		t.Fatalf("proxy allocated too much memory: %d bytes", totalAlloc)
+	}
+}
+
+type noopReader struct {
+	len uint64
+	pos uint64
+}
+
+var _ io.Reader = &noopReader{}
+
+func (r *noopReader) Read(b []byte) (int, error) {
+	if r.pos >= r.len {
+		return 0, io.EOF
+	}
+	n := int(r.len - r.pos)
+	if n > len(b) {
+		n = len(b)
+	}
+	for i := range b[:n] {
+		b[i] = 0
+	}
+	r.pos += uint64(n)
+	return n, nil
+}
+
 func newFakeUpstream(name string, insecure bool) *fakeUpstream {
 	uri, _ := url.Parse(name)
 	u := &fakeUpstream{
@@ -998,6 +1082,7 @@ func (u *fakeUpstream) Select(r *http.Request) *UpstreamHost {
 func (u *fakeUpstream) AllowedPath(requestPath string) bool { return true }
 func (u *fakeUpstream) GetTryDuration() time.Duration       { return 1 * time.Second }
 func (u *fakeUpstream) GetTryInterval() time.Duration       { return 250 * time.Millisecond }
+func (u *fakeUpstream) GetHostCount() int                   { return 1 }
 
 // newWebSocketTestProxy returns a test proxy that will
 // redirect to the specified backendAddr. The function
@@ -1049,6 +1134,7 @@ func (u *fakeWsUpstream) Select(r *http.Request) *UpstreamHost {
 func (u *fakeWsUpstream) AllowedPath(requestPath string) bool { return true }
 func (u *fakeWsUpstream) GetTryDuration() time.Duration       { return 1 * time.Second }
 func (u *fakeWsUpstream) GetTryInterval() time.Duration       { return 250 * time.Millisecond }
+func (u *fakeWsUpstream) GetHostCount() int                   { return 1 }
 
 // recorderHijacker is a ResponseRecorder that can
 // be hijacked.
--- a/caddyhttp/proxy/upstream.go
+++ b/caddyhttp/proxy/upstream.go
@@ -423,6 +423,10 @@ func (u *staticUpstream) GetTryInterval() time.Duration {
 	return u.TryInterval
 }
 
+func (u *staticUpstream) GetHostCount() int {
+	return len(u.Hosts)
+}
+
 // RegisterPolicy adds a custom policy to the proxy.
 func RegisterPolicy(name string, policy func() Policy) {
 	supportedPolicies[name] = policy


mholt:
Cool, thanks for giving this a go @lhecker!

For the test, how about using `testing.Short()` as described at the top of the overview here: https://golang.org/pkg/testing/

Or maybe the test could be sized down just a notch?

lhecker:
@mholt That would still not lead to a reproducible, reliable test (since only OOM errors would be a proof for it to be not working). I just found `debug. SetGCPercent()` though with which we could disable the GC. If we then proxy a relatively large request (e.g. 100MB) and check wether the  memory usage after running it is still _below_ the request size we can confirm that it was streaming the request in chunks with a small buffer (32KB currently) instead of fully loading it into the memory first.

I'll add such a test okay? 🙂

mholt:
That's... A really neat idea. I've never heard of doing that for a test
like this but it makes sense. Give a whirl! I wonder how the CI environment
will like it...

lhecker:
@mholt Found an even easier solution by simply calculating the difference between the `runtime.MemStats.TotalAlloc` before and after the request and added it to this PR just now.

I wrote the new test case based on `TestReverseProxyRetry` using a request body of 100MB. If I disable the optimizations in this PR my test reliably fails with:
```
proxy allocated too much memory: 268847840 bytes
```

and if I enable the optimizations it reliably succeeds with `582320` allocated bytes.
(The test fails if more than 100MB are allocated, because it's very likely a sign that the request body must have been buffered. And if not it's still a sign that the proxy module allocates too much memory.)

For some strange reason the test runs **a lot** faster if I run it with `go test -cpu 1` (2.2s vs 0.2s). Any idea why that might be the case?

mholt:
That's interesting -- I don't know why that is! I just tried running the test on my Macbook Pro and I see similar results.

caddyhttp/proxy/body.go
```diff
@@ -23,9 +30,19 @@ func (b *bufferedBody) rewind() error {
...
 }
 
+type unbufferedBody struct {
+	io.ReadCloser
+}
+
+var _ rewindableReader = &unbufferedBody{}
+
+func (b *unbufferedBody) rewind() error {
+	panic("cannot rewind unbuffered body")
```
mholt: Y u no return error? 😓

mholt:
More thinking out loud here...

The issue that motivated the buffering of proxy bodies is here: https://github.com/mholt/caddy/issues/1229 - and the subsequent PR merged the change.

From what I can tell, other load balancers like nginx and haproxy will try other upstreams only if they couldn't establish a connection. This makes sense, since it means the request body wouldn't be drained: no was connection was ever established to drain it.

Currently, Caddy may try the next upstream for any error returned by the reverseproxy's ServeHTTP method, which (currently) returns its first possible error at the RoundTrip(). It's unclear to me whether RoundTrip() returns an error only on connection problems or potentially after establishing the connection. The [net/http docs](https://golang.org/pkg/net/http/#RoundTripper) only talk about whether a "response" was obtained, saying:

> RoundTrip must return err == nil if it obtained a response, regardless of the response's HTTP status code. A non-nil err should be reserved for failure to obtain a response.

I'm not sure whether this is interpreted as "complete response body" or "part of a response"... well anyway, I think the way we do it now is alright, but I'm not sure it's ideal.

What I'm getting at is:

The server operator should be able to control whether the proxy retries with another upstream. Right now, it is off by default and, if enabled, only happens if RoundTrip() returns an error (presumably a connection error?). This is good.

It would be _really_ neat if a "proxy-mode error level" could be defined. Right now, that definition is something like this: "failure to obtain a response" (according to the Go docs), which I guess includes "failure to establish a connection". This is a good default and is what other LBs do. But what if the operator could define "a timeout of 10s" or "a 500 response" as proxy-mode error levels? That way, if an upstream serve was accepting connections but unable to handle them, the operator _could_ tell Caddy to treat a 500 like a connection error: retry the request on a different upstream. (The operator would have to understand this has the potential to retry non-idempotent POST requests and buffer response bodies, but... that's how it goes.)

So -- what does this have to do with your PR? It touches on this in a very significant way. Ideally, the only time we should have to buffer the body is if it started being drained, then was stopped, then we had to retry. All other errors -- connection errors, whatever -- shouldn't need it, even if we have multiple upstreams.

What do you think?

caddyhttp/proxy/body.go
lhecker: Ah yes... When I wrote this method it wasn't hidden behind a common interface. Calling `rewind()` on it was thus basically a precondition failure, akin to how `501 Not Implemented` is a server error code instead of being a client error like 404. I still kind of think that it's a precondition failure, but since this struct is only used using the `rewindableReader` it's surely not the right choice anymore. I'll change it.

lhecker:
> I have mixed feelings on this -- since ideally an upstream shouldn't apply changes when a POST fails!

The issues begin when not the request to the upstream but it's response fails and you apply a POST request multiple times due to this. E.g. instead of ordering one expensive product you order a hundred and won't notice because the response from the backend is never reaching caddy.

> To enable this retry [...], multiple upstreams need to be provided and `try_duration` has to be set to a value > 0.

Oh from the docs [1] I thought that a `try_duration` of 0 just makes the proxy try all upstreams in a loop forever? I guess it should be updated to say that "0s" disables the retrying instead? If that's correct I'll change the code to check for `try_duration == 0`.

> It's unclear to me whether RoundTrip() returns an error only on connection problems or potentially after establishing the connection.

It returns errors on connection failures, timeouts (as configured on the transport) _and_ on protocol errors (TLS & HTTP). This does not include the response body though because it's not parsed until you actually read from the `res.Body`.

> It would be really neat if a "proxy-mode error level" could be defined. [...]

Yeah. [Here](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream) are the nginx docs for this. As you can see it provides a flexibility similar to what you wrote.

> Ideally, the only time we should have to buffer the body is if it started being drained, then was stopped, then we had to retry.

I'm not sure if I understand. At the point you try to "retry" without having buffered the body already you cannot retry anymore because you failed to buffer in the first place. So either you start buffering right from the start and be able to retry upstream requests or you don't buffer and can't retry. But I believe there is no other way apart from deciding it before you even try to connect to the first upstream.

Even if you say "Well but what if we first see wether we can connect to the first upstream?" we still have to buffer it. The reason is simple: Calls to `res.Body.Read()` can still fail, because the HTTP body might be in a malformed encoding, or because the TCP connection disconnects while reading the body. That's why you have to buffer it anyways.

I don't think we should spend too much time in optimizing the buffering logic _for now_ apart from the 2 obvious cases though (only one upstream or `try_duration` is 0). Most of the time you'll want to only use the retry logic for GET requests anyways (which don't have a large body!) due to the major issue with e.g. POST requests I mentioned above.

[1]:
> By default, this retry is disabled ("0s"). Clients may hang for this long while the proxy tries to find an available upstream host.

mholt:
@lhecker 

> The issues begin when not the request to the upstream but it's response fails and you apply a POST request multiple times due to this. E.g. instead of ordering one expensive product you order a hundred and won't notice because the response from the backend is never reaching caddy.

IMO a "request" should fail or be rejected if any part of the RoundTrip fails, but apparently most backend services don't think about that. (I can see why, it's just not ideal.) Like rolling back a DB transaction if there's an error.

> Oh from the docs [1] I thought that a try_duration of 0 just makes the proxy try all upstreams in a loop forever? I guess it should be updated to say that "0s" disables the retrying instead?

Huh? It doesn't say anything about retrying forever. This is what it says: "try_duration is how long to try selecting available upstream hosts for each request. By default, this retry is disabled (0s)."

> If that's correct I'll change the code to check for try_duration == 0.

That'd be a good change in the meantime, until we figure out exactly how we want to configure retries. I'm not sure yet if it will affect your PR any more than this, so at least make the change and I'll take another look.

> It returns errors on connection failures, timeouts (as configured on the transport) and on protocol errors (TLS & HTTP).

By HTTP protocol errors, do you mean if the backend returns a 4xx or 5xx response, an error value is returned? My understanding is that it only happens on connection failures (and maybe timeouts).

> Yeah. Here are the nginx docs for this. As you can see it provides a flexibility similar to what you wrote.

Oh, I didn't realize nginx did this. Thanks. Those docs say that nginx can't retry if the body has been read, so I guess they aren't buffering it. I still think buffering isn't all bad in some use cases, as long as the user configures it. So I like where this PR is going.

> I'm not sure if I understand. At the point you try to "retry" without having buffered the body already you cannot retry anymore because you failed to buffer in the first place. So either you start buffering right from the start and be able to retry upstream requests or you don't buffer and can't retry.

Right, that's what I mean -- buffer from the start. But we should know ahead of time whether we need to buffer: retries are enabled and there are multiple backends.

> I don't think we should spend too much time in optimizing the buffering logic for now apart from the 2 obvious cases though (only one upstream or try_duration is 0). Most of the time you'll want to only use the retry logic for GET requests anyways (which don't have a large body!) due to the major issue with e.g. POST requests I mentioned above.

Fair enough -- one thing at a time. :)

lhecker:
> but apparently most backend services don't think about that.

No most really don't, especially as soon as you start layering your abstraction on top, where e.g. your DB transaction does not extend until you finished replying to the client.
In fact it's actually quite hard to do since you need to e.g. commit first before you know wether the transaction succeeded. And then you have to reply to the client which can still fail if meanwhile the connection broke down for whatever reason, but you can't rollback anymore because you commited already and... _aaaaaaah..._ You just noticed you'd need to implement something like the 2-phase commit protocol over HTTP and you start wondering if growing potatoes in your basement isn't the better option over all this trouble. 😂

Anyways:

> Huh? It doesn't say anything about retrying forever.

Ah I misread. I thought it said that the client might hang for a long time if retry is disabled. (Take another look at the docs and you'll notice how close my initial understanding and the correct version are. 😅)

> That'd be a good change in the meantime, until we figure out exactly how we want to configure retries.

Done & Pushed.

> By HTTP protocol errors, do you mean if the backend returns a 4xx or 5xx response, an error value is returned?

No actual protocol errors like e.g. replying with `foobar/1.1 200 OK` in the first line. That would make `RoundTrip` return an error. Status codes, headers (e.g. cookies) or the body aren't interpreted. It's basically just one single level above a raw TCP connection.

> [...] only meeting all 3 of those conditions should induce buffering.

Yeah. To make it easier to extend I already moved the check to a [free-standing assignment](https://github.com/mholt/caddy/pull/1314/commits/f8c0e08388ac0bce7cb8619b728aed331420e37f#diff-5dd020071b9de2477b5734a9bd3aed57R110). If it get's too complicated we can surely move it to a seperate method in the future.

oliverpool:
I want to serve a [tus server](https://tus.io/) behind caddy (via a reverse proxy) and this PR might fix my issue:

**Scenario:**
A large file upload is interrupted (due to bad network for instance).
The file upload is retried.

**Expected behavior** (when the request is directly sent to tus listening port)
The tus server manages to store the data that has been sent so far.
On client retry, only the missing part of the file is sent.

**Observed behavior** (when caddy acts as a reverse proxy in front of tus):
The tus server does not see any request and the client has to send the whole file again

---

If think that this might fix my issue, I can try to compile this branch on my system and let you know if it works!

In this case, it could be used as a test case for this PR:
send an *incomplete* request and see if it is passed to one of the backend

oliverpool:
(This could also be seen as unrelated to this issue that the caddyserver [does not forward incomplete requests](https://github.com/mholt/caddy/blob/f8c0e08388ac0bce7cb8619b728aed331420e37f/caddyhttp/proxy/proxy.go#L116))

oliverpool:
I just tried this branch with the tus uploader and resuming works perfectly!

I'm looking forward for this PR to be merged :smiley:

oliverpool:
> I assume incomplete requests work as well then? Because I'm not entirely sure if that's the case…

I think so. 
Without your patch it didn't work (the tus server didn't receive any request, probably because caddy filters out when a [request is incomplete](https://github.com/mholt/caddy/blob/f8c0e08388ac0bce7cb8619b728aed331420e37f/caddyhttp/proxy/proxy.go#L116)). 
With your patch it works perfectly (the tus server receives the incomplete request and is capable of handling it)

lhecker:
@mholt Is everything fine with this PR? And if so: When do you plan on merging it?

caddyhttp/proxy/body.go
```diff
@@ -23,9 +31,19 @@ func (b *bufferedBody) rewind() error {
...
 }
 
+type unbufferedBody struct {
+	io.ReadCloser
+}
+
+var _ rewindableReader = &unbufferedBody{}
+
+func (b *unbufferedBody) rewind() error {
+	return errors.New("cannot rewind unbuffered body")
```
tw4452852: Could this just return `nil ` instead of a error? so you don't a check below.

caddyhttp/proxy/proxy.go
```diff
@@ -173,15 +192,20 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 			downHeaderUpdateFn = createRespHeaderUpdateFn(host.DownstreamHeaders, replacer)
 		}
 
-		// rewind request body to its beginning
-		if err := body.rewind(); err != nil {
-			return http.StatusInternalServerError, errors.New("unable to rewind downstream request body")
+		// NOTE: We check for requiresBuffering, because
+		// only a bufferedBody supports rewinding.
+		if requiresBuffering {
```
tw4452852: I think we could clear this checking if `unbufferBody`'s rewind() just return nil

caddyhttp/proxy/body.go
mholt: Oh, you're suggesting to just make this a no-op?

If it's a no-op, maybe the bigger design question is, Why does `unbufferedBody` implement `rewindableReader` if it is not a rewindable reader?

tw4452852: Hmm, it makes sense.

caddyhttp/proxy/proxy.go
tw4452852: Actually we only need to rewind the body when the previous one failed.

caddyhttp/proxy/proxy.go
```diff
@@ -231,6 +224,16 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
...
 			break
 		}
+
+		// Before we retry the request we have to make sure
+		// that the body is rewound to it's beginning.
+		//
+		// NOTE:
+		//   Reaching this point implies that keepRetrying() above returned true,
+		//   which in turn implies that body must be a *bufferedBody.
+		if err := body.(*bufferedBody).rewind(); err != nil {
```
mholt: _Juuuuust_ to be safe, maybe this code should have one extra check. If the type assertion fails right now, that will lead to a panic (I know, it shouldn't fail, but I suspect we lack tests to enforce this right now). Either a test should be written -- this particular case might be tedious to test for, I'm not sure -- or try this, which won't panic:

```go
if bb, ok := body.(*bufferedBody); ok {
    err := bb.rewind()
    if err != nil {
        // ...
    }
}
```

lhecker: @mholt Fixed.

Can I squash the last 3 commits into a single one before you merge it to clean up the history? It'd result in a force push though of course.

caddyhttp/proxy/proxy.go
```diff
@@ -173,15 +189,20 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
...
 
 		// tell the proxy to serve the request
-		atomic.AddInt64(&host.Conns, 1)
-		backendErr = proxy.ServeHTTP(w, outreq, downHeaderUpdateFn)
-		atomic.AddInt64(&host.Conns, -1)
+		func() {
+			atomic.AddInt64(&host.Conns, 1)
+			defer atomic.AddInt64(&host.Conns, -1)
+			backendErr = proxy.ServeHTTP(w, outreq, downHeaderUpdateFn)
+		}()
```
mholt: This won't stop merging, but I just noticed this wrapped in a func(). Why is that? Just curious.

caddyhttp/proxy/proxy.go
```diff
@@ -197,10 +218,10 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 		timeout := host.FailTimeout
 		if timeout > 0 {
 			atomic.AddInt32(&host.Fails, 1)
-			go func(host *UpstreamHost, timeout time.Duration) {
+			go func() {
 				time.Sleep(timeout)
 				atomic.AddInt32(&host.Fails, -1)
-			}(host, timeout)
+			}()
```
mholt: Wait a sec -- why are these parameters removed? It is a bug to start a goroutine in a `for` loop and not pass in the values as they are on this iteration of the loop, they might change before the goroutine starts.

caddyhttp/proxy/proxy.go
oliverpool: Probably for the `defer`

mholt: Just move line 203 to right after line 204... I don't think there's a need to use defer here.

oliverpool: I think it's probably for it to look cleaner (but I'm not the author).
The increment and decrement are now spatially near one another: the rest of the function (only one line here) will then be in between (even in case of an exception or an early return)

mholt:
If that's the only reason, then I'd like it to change to not use the defer.

caddyhttp/proxy/proxy.go
oliverpool: For reference: https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

> It is also important to note that variables declared within the body of a loop are not shared between iterations, and thus can be used separately in a closure. 

So this should actually work fine, but it is a dangerous portion of code (for instance if the `for` loop is changed to `for host := range upstream.Selector(r) {`)

mholt: Sure, I still think it should be reverted.

caddyhttp/proxy/proxy.go
lhecker: It's because `proxy.ServeHTTP` can panic under certain conditions, but previously the `host.Conns` counter was not decremented properly if it did. That way - in my understanding - the counter was stuck and could sooner or later lead to a broken proxy state.

caddyhttp/proxy/proxy.go
lhecker: I'll revert it. I changed it because it was not written in the same style as other parts of the code, which all use implicit binding of closure parameters.

caddyhttp/proxy/proxy.go
mholt: Hm, makes sense. It's a little awkward but we'll roll with it for now. Maybe place a comment explaining why it's in an anonymous func.

caddyhttp/proxy/proxy.go
mholt: Okay; will await that, I'll feel better if they get passed into the func ;)

tw4452852:
I think the root cause for this issue is we have to read full body beforehand to rewind it later when retrying  . Here, we just work around it by checking the `requiresBuffering`, however this issue still exists if checking failed. So I think the proper fix is that we buffer the body only if retrying the request.  Anyway we could keep this patch for now and I will try to implement it later.

oliverpool:
@tw4452852 : something like a [TeeReader](https://golang.org/pkg/io/#TeeReader) with a [NopCloser](https://golang.org/pkg/io/ioutil/#NopCloser) (to have a `ReadCloser`)? (thus you can copy the forwarded request to a local buffer and use this buffer to repeat if needed)

tw4452852:
@oliverpool Yes, I actually have adopted this way in [here](https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/replacer.go#L92). :smile:

lhecker:
> So I think the proper fix is that we buffer the body only if retrying the request.

That's inherently impossible. You can't decide to buffer the body _after_ you've already consumed it by proxying it to the first upstream host. Some way or the other you have to decide upfront wether you'd like to buffer the body, especially since you have to keep in mind that request payloads don't have an upper bound on their size.

The only optimizations I see is to better abstract this logic away and to optimize `bufferedBody` to transparently backup the body into a buffer while it's read from (which reduces latency).

mholt:
@lhecker I'm ready to merge this, as soon as the goroutine gets passed the arguments again. :+1:

@tw4452852 I appreciate your thinking on this. I'm not sure I understand fully what you have in mind but perhaps you could explain more? Or submit a PR... but yeah, you'll have to decide to buffer before consuming the body.

@lhecker Maybe he meant the body will not be buffered unless we know we _might_ retry the request?

lhecker:
@mholt I pushed the requested changes just now. Can/May I now squash the later commits into the first 3?

mholt:
@lhecker Sure! Go ahead and clean up your commits as you please and when ready I'll merge this. 

@tw4452852 It looks like your concerns have been addressed so I'll assume you're good with my merging this. We can always go back and make little changes if we find anything else.

lhecker:
@mholt I cleaned up the commits. As you can see it's a very compact PR now. 🙂

mholt:
Excellent!! Thanks so much for your patience and persistence @lhecker, and for your help reviewing, @tw4452852 and others. 

At some point I intend to revisit this (unless somebody else wants to first! I'm busy for a while) to make retries more configurable; i.e. to only buffer if explicitly configured to do so, even with multiple backends. Otherwise, to only retry on connection failures where the request body wasn't drained at all.

PR #1323, opened 2017-01-03
Fix position of "filter" in directives

blaubaer:
...to fix [#2 Replacement doesn't happen - of caddy-fitler](https://github.com/echocat/caddy-filter/issues/2).

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -411,8 +411,8 @@ var directives = []string{
 	"header",
 	"errors",
 	"filter", // github.com/echocat/caddy-filter - Required at this position because if it is
-	          //                                   enabled it captures the response of following
-	          //                                   middlewares and modifies it.
+	//                                             enabled it captures the response of following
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/plugin.go
```diff
@@ -410,6 +410,9 @@ var directives = []string{
 	"gzip",
 	"header",
 	"errors",
+	"filter", // github.com/echocat/caddy-filter - Required at this position because if it is
+	//                                             enabled it captures the response of following
+	//                                             middlewares and modifies it.
```
mholt: We can probably lob off the explanation -- if we did that for every directive, we'd have a bit of an unwieldy list. :) Keep the package path though. (Sorry, I know this absolute ordering isn't a great system, but it's the only correct way I can think of doing it!)

blaubaer: No problem. I just think it could be a good idea to left a hint for this strange thing - that not someones reorders this and we trap in new trouble. :)

mholt:
Great, thanks!

Issue #1324, opened 2017-01-03
Basicauth and proxy not working


oliverpool:
### 1. What version of Caddy are you running (`caddy -version`)?
Caddy 0.9.4 (on windows 10)

### 2. What are you trying to do?
Protect a reverse proxy via basicauth

### 3. What is your entire Caddyfile?
```text
example.com:801 {
  errors errors.log
  log access.log
  tls off
  
  basicauth /admin "r" "r"
  proxy /admin/monitor 192.168.3.9:80
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
`caddy.exe` in the command prompt

### 5. What did you expect to see?
```
> curl -u r:r http://example.com:801/admin/monitor
<!doctype html>...
> curl -u hacker:r http://example.com:801/admin/monitor
401 Unauthorized
```
The first command should return the page from the proxy, while the second should return a 404 (the `r` is missing - meaning the auth was successful)

### 6. What did you see instead (give full error messages and/or log)?
```
> curl -u r:r http://example.com:801/admin/monitor
{"message":"Basic auth failed"}
> curl -u hacker:r http://example.com:801/admin/monitor
401 Unauthorized
```
The first command fails to authenticate!

If I request a page not behind the proxy `admin/test` for instance, I get an error 404 or 401 as expected (depending if the auth was successful).

mholt:
Caddy doesn't write a JSON response when basic auth fails... it looks like Caddy's basic auth is succeeding but your backend is seeing it as failed and writing some sort of JSON response.

oliverpool:
Thanks for your very quick reply!
Can I prevent caddy from forwarding the basic auth info to the backend?

mholt:
Try this in your proxy directive:

```
header_upstream -Authorization
```

We changed proxy recently so that it _will_ forward the Authorization header, but in the cases where Caddy's basicauth successfully completes the authorization, perhaps that header should be removed.

In any case, what I showed you should work.

oliverpool:
Great, it works!

Thank you very much for this quick and efficient support :1st_place_medal:

mholt:
@oliverpool No problem, see #1325. This _should_ mean you won't need to manually unset that header. I don't think my change will break anyone but we'll see. (If it does break someone, I think they're doing something strange...)

PR #1325, opened 2017-01-04
basicauth: Remove Authorization header on successful authz (issue #1324)

mholt:
If a site owner protects a path with basicauth, no need
to use the Authorization header elsewhere upstream, especially since it
contains credentials.

If this breaks anyone, it means they're double-dipping. It's usually
good practice to clear out credentials as soon as they're not needed
anymore. (Note that we only clear credentials after they're used,
they stay for any other reason.)

### DIFF
--- a/caddyhttp/basicauth/basicauth.go
+++ b/caddyhttp/basicauth/basicauth.go
@@ -34,8 +34,7 @@ type BasicAuth struct {
 
 // ServeHTTP implements the httpserver.Handler interface.
 func (a BasicAuth) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
-	var hasAuth bool
-	var isAuthenticated bool
+	var protected, isAuthenticated bool
 
 	for _, rule := range a.Rules {
 		for _, res := range rule.Resources {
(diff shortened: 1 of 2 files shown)


mholt:
Would someone like to review this? @tw4452852 or @elcore perhaps?

mholt:
@radek-senfeld Why do you need to authenticate twice? If you authenticate with Caddy, you don't need it with Gogs. If Gogs is doing the authentication, don't have Caddy do it...?

radek-senfeld:
There's no other way I aware of to hide Gogs completely from unauthorized users.

Also we are not sure how secure it is and as it contains quite sensitive data (Git repos) better to start with authorization in Caddy.

I know it's not very elegant but "it worked for us (tm)".. ;-)

mholt:
So did this change break your setup? That means that you were authenticating twice, which ... is nonsensical to me still. If you have specific instructions to reproduce this and you think this is actually a bug in Caddy (and not in your configuration), feel free to open a new issue.

radek-senfeld:
@mholt Yes, I'm pretty sure this line is the show-stopper:

https://github.com/mholt/caddy/blob/master/caddyhttp/basicauth/basicauth.go#L62

I'm gonna try to compile Caddy with the mentioned line removed and will eventually open a new issue if it fixes our setup.

mholt:
@radek-senfeld The problem is that keeping that line in there broke someone else in #1324.

radek-senfeld:
@mholt Confirmed. After removing the aforementioned line everything works as before.

I'm aware there are two groups of users, each demanding quite opposite. :-)

I guess it could be either user-configurable (seems like removing Authorization header should be default behavior) or Caddy could add header (e.g. X-WEBAUTH-USER) after successful auth.

Because of this issue I found out Gogs can be configured to respect such a header which would de-dupe auths and I would stop you bothering on this subject.. ;-) Win - Win situation? ;-)

Should I start working on a PR? ;-)

mholt:
If Gogs shouldn't be showing the homepage to unauthorized users, that sounds like a bug in Gogs, not in Caddy. Have you tried submitting a PR with Gogs instead?

radek-senfeld:
It's not a bug rather than a feature I guess. Nevermind, we will stick with nginx.

mholt:
It's a feature of Gogs to have the homepage unprotected?

radek-senfeld:
I was rather aiming for "kept undisclosed". That's the main reason why we do auth in a reverse proxy. Honestly I don't think is an uncommon use case.

You're right that passing Authorization header after successful auth should not be necessary and is not optimal.

As I wrote I found out just today that Gogs can be configured to use username sent by proxy in a header after successful auth - e.g. X-Auth-User.

I'm pretty sure this feature could be great for some other applications as well.

mholt:
I see. You're right, it's not uncommon; I just have yet to see a valid use case for authenticating in the reverse proxy and then also authenticating with the same method and credentials upstream.

Ah, so if we set the X-Auth-User header after a successful authorization, would that help with Gogs? I can't see that breaking anything.

radek-senfeld:
According to Gogs docs it should help indeed:

https://gogs.io/docs/advanced/configuration_cheat_sheet#security-(security)
REVERSE_PROXY_AUTHENTICATION_USER: Header name for reverse proxy authentication username

https://gogs.io/docs/advanced/configuration_cheat_sheet#service-(service)
ENABLE_REVERSE_PROXY_AUTHENTICATION: Enable this to allow reverse proxy authentication

mholt:
That's cool, if that would help then feel free to submit a PR (I gotta get back to my homework after this morning's release). But does Gogs trust that header implicitly? Hopefully it is only accessible via reverse proxy, otherwise clients could just set the header themselves and pretend that they've been authorized.

radek-senfeld:
That's my concern as well. For security reasons the header should be unset/ignored on incoming requests. Which could then break cascaded proxying.. Although I'm not sure how common this setup with Caddy in the "middle" is.

Ok then, thanks! Hopefully I start working on it on friday.

PR #1330, opened 2017-01-06
Add StartupHooks to Plugins

hacdias:
StartupHook enables the developer to create plugins that are executed in the very beginning of the Caddy process, even before it has started running. It allows, among many other things, to use flags and take advantage of the fact that Caddy hasn't started running yet to configure and enable a variety of features. I think this hook will open a whole new range of plugins.

I've already talked with @mholt about this because I was having a problem while I wanted to develop a plugin and he suggested this.

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -70,14 +70,9 @@ func Run() {
 		})
 	}
 
-	err := caddy.StartupHooks(serverType)
-	if err != nil {
-		mustLogFatalf(err.Error())
-	}
-
 	// Check for one-time actions
 	if revoke != "" {
-		err = caddytls.Revoke(revoke)
+		err := caddytls.Revoke(revoke)
 		if err != nil {
 			mustLogFatalf(err.Error())
 		}
@@ -97,7 +92,13 @@ func Run() {
 	}
 
 	// Set CPU cap
-	err = setCPU(cpu)
+	err := setCPU(cpu)
+	if err != nil {
+		mustLogFatalf(err.Error())
+	}
+
+	// Execute Startup Hooks
+	err = caddy.StartupHooks(serverType)
 	if err != nil {
 		mustLogFatalf(err.Error())
 	}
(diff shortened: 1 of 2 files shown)


plugins.go
```diff
@@ -176,6 +200,10 @@ type Plugin struct {
 	// Action is the plugin's setup function, if associated
 	// with a directive in the Caddyfile.
 	Action SetupFunc
+
+	// StartupHook is the plugin's function that is executed
+	// imediatelly after the flag parsing.
```
abiosoft: typo, should be `immediately`

plugins.go
```diff
@@ -69,6 +69,30 @@ func DescribePlugins() string {
 	return str
 }
 
+// StartupHooks executes the startup hooks defined when the
+// plugins where registred and returns the first error
```
abiosoft: typo, should be `were registered`

abiosoft:
Well done @hacdias 👏

hacdias:
Thanks @abiosoft! Typos corrected :smile:

plugins.go
```diff
@@ -69,6 +69,30 @@ func DescribePlugins() string {
 	return str
 }
 
+// StartupHooks executes the startup hooks defined when the
+// plugins were registred and returns the first error
```
abiosoft: There is still one typo, regist**e**red.

hacdias:
@abiosoft thanks 😄

wendigo:
LGTM

caddy/caddymain/run.go
```diff
@@ -97,6 +97,12 @@ func Run() {
 		mustLogFatalf(err.Error())
 	}
 
+	// Execute Startup Hooks
```
mholt: A more useful comment might be, "Execute plugins that are registered to run as the process starts"

caddy/caddymain/run.go
```diff
@@ -97,6 +97,12 @@ func Run() {
...
+	err = caddy.StartupHooks(serverType)
+	if err != nil {
+		mustLogFatalf(err.Error())
+	}
```
mholt: Change to `"%v", err` just to be safe (I know, I know, we don't do that anywhere else in this file, I just thought of it.)

hacdias:
Sweet 😄

Issue #1327, opened 2017-01-04
Caddy rewrites example.com/foo/bar to example.com/bar/


Mxbonn:
I am trying to host a site on a specific path on my domain, so my caddy file looks like this:
```text
example.com/foo/
```
My directory structure looks like this
```text
root/
| -- Caddyfile
| -- index.html
| -- bar/
|     | -- index.html
````
So when I want to see the content of index.html on my site I would go to:
`example.com/foo/bar/index.html` or `example.com/foo/bar/`.
This works all like expected, but if I don't include that final '/' and type `example.com/foo/bar` it gets redirected to `example.com/bar/`, which is not a valid page, nor what I intended.

I know that `foo/bar` and `foo/bar/` are not the same by default but the fact that Caddy does add a trailing slash make me think the intention is to make them both equal when there is no file called bar.

Is this a bug or what should I do to redirect `example.com/foo/bar` to `example.com/foo/bar/`?

**Comment by mholt** [comment]
Does the same thing happen if using curl? Use curl -v and post the exact, unmodified output.

**Comment by tobya** [comment]
Can you post your caddyfile?

**Comment by Mxbonn** [comment]
@tobya I posted it in the first part of the issue (although abstract) but concretely its just:
```text
bonnaerens.me/best/
```

mholt:
Oh, I think I might know what's going on. Somebody have a chance to look at it? https://github.com/mholt/caddy/blob/3f770603bc8d55dfd6d90131f62bbf61e70c7ebd/caddyhttp/httpserver/https.go#L151 -- I'm willing to bet that URI has `/best` stripped from the URI because the site definition includes `/best` -- see https://github.com/mholt/caddy/blob/3f770603bc8d55dfd6d90131f62bbf61e70c7ebd/caddyhttp/httpserver/server.go#L258-L266.

But I'm speculating. Would love if someone could verify and fix. I'm super busy until next week. ;)

ericdreeves:
@mholt Indeed, pathPrefix is trimmed from the path.

As a quick test, I used RequestURI from http.Request (+"/") [here](https://github.com/mholt/caddy/blob/3f770603bc8d55dfd6d90131f62bbf61e70c7ebd/caddyhttp/staticfiles/fileserver.go#L87).

Here's the result when allowing curl to follow redirects (some curl output omitted).

	# curl -v -L https://example.com/foo/bar
	> GET /foo/bar HTTP/1.1
	> User-Agent: curl/7.38.0
[...]

PR #1331, opened 2017-01-07
Use RequestURI when redirecting to canonical path.

ericdreeves:
Caddy may trim a request's URL path when it starts with the path that's
associated with the virtual host. This change uses the path from the request's
RequestURI when performing a redirect.

Fix issue #1327.


caddyhttp/staticfiles/fileserver.go
```diff
@@ -80,17 +80,35 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
+	u := r.URL.Path
 	if d.IsDir() {
 		// Ensure / at end of directory url
-		if !strings.HasSuffix(url, "/") {
-			Redirect(w, r, path.Base(url)+"/", http.StatusMovedPermanently)
+		if !strings.HasSuffix(u, "/") {
+			// Since the path in u may have been trimmed by Caddy, and Redirect
+			// requires the full path, parse the request's RequestURI and use the
+			// path.
+			redirurl, err := url.ParseRequestURI(r.RequestURI)
```
abiosoft: let's rename this to `redirURL` or anything else that won't give lint/vet error.

tw4452852:
Maybe [https](https://github.com/mholt/caddy/blob/3f770603bc8d55dfd6d90131f62bbf61e70c7ebd/caddyhttp/httpserver/https.go#L151) also is the case.

caddyhttp/staticfiles/fileserver.go
ericdreeves: Thanks for taking a look, @abiosoft. I've made the change.

ericdreeves:
@tw4452852 Acknowledged. I'm investigating.

caddyhttp/httpserver/server.go
```diff
@@ -265,6 +266,11 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 		}
 	}
 
+	if vhostURL, err := url.Parse(vhost.Addr.String()); err == nil {
+		r.URL.Scheme = vhostURL.Scheme
+		r.URL.Host = vhostURL.Host
+	}
```
mholt: Can you add a comment explaining why this is necessary?

caddyhttp/staticfiles/fileserver.go
```diff
@@ -80,17 +80,37 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
+			redirURL, err := url.ParseRequestURI(r.RequestURI)
+			if err != nil {
+				return http.StatusInternalServerError, err
+			}
+			if !strings.HasSuffix(redirURL.Path, "/") {
+				redirURL.Path = redirURL.Path + "/"
+			}
+			toURL := r.URL.ResolveReference(redirURL)
+
+			http.Redirect(w, r, toURL.String(), http.StatusMovedPermanently)
```
mholt: Hmm, it's unfortunate that all this is required. Sorry about that. Since you've now worked with this a bit, do you have any recommendations for how we can make it easier/simpler?

caddyhttp/httpserver/server.go
ericdreeves: Will do!

caddyhttp/staticfiles/fileserver.go
ericdreeves: My concern with changing fields in r.URL is that it might cause issues for other middleware Handlers that rely on them. Otherwise, only the RequestURI seems to have the original path. Let me know if that's not the case.

The check for "/" on redirURL.Path is probably redundant and can be removed.

I'll make that change and add the requested comment. Thanks for taking a look!

mholt:
I think AppVeyor is having trouble, not a problem with this PR:

```
ERR Error running script (call to f_5b8e3b60e050400311bc8b6ef2cbba37bc0b9cb3): @user_script:1: @user_script: 1: -OOM command not allowed when used memory > 'maxmemory'.
```

mholt:
@ericdreeves Just had an idea. What if Caddy didn't chop off the path in the request URI to adjust for the site's address in the Caddyfile, but instead set a field on the SiteConfig and then middlewares that want to access the file system based on the request path have to take that into account...? (Maybe we provide a method to make this easy? It's just a path.Join though. So maybe not needed.)

But then we wouldn't have to do all these kinda ugly hacks -- which I appreciate you for diving into this. :) What do you think?

tw4452852:
I don't think this is a good choice, because all exist middles will be changed if they use the `r.URL`.
Actually I have another way to handle this: if any middleware want to redirect, it just returns the special status code (maybe `http.StatusTemporaryRedirect`) and saves the destination in the response header. After all the middleware chains returns, we handle the redirection actually if any. ([here is the proper place](https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/server.go#L281)).

mholt:
@tw4452852 That's an interesting idea!

So, two options:

1) We don't chop off the URI if the site address contains a path (like we do now) -- instead, middlewares that want to access static files on disk just have to take into account the site's path. Slightly error prone, if middlewares don't think about that. They'll have to store what the base path is then do a TrimPrefix or something.

2) Middlewares that want to redirect just set the Location header and return a 3xx status code. This is really quite elegant. However, will this break anything like errors or log or gzip middlewares? How does this work with the `redir` middleware? I guess there's no response body, unless the `redir` middleware writes a body because of a "meta tag" redirect...

I like Tw's idea most, but I'll have to give it some thought and also somebody should experiment to make sure it won't break anything. :smile:

ericdreeves:
If we go with the second option, how would it solve the original issue? If the middleware has no way of knowing that the URL path was changed, what would it write to the Location header?

tw4452852:
@ericdreeves , Location header will only be the relative path, it will be joint with site root finally to make up a right redirection URL.

tw4452852:
@mholt I think every middleware now should work with it w/o changes. It just delegates(defers) the redirection action.

ericdreeves:
@tw4452852 That's what I figured. Thanks for confirming.

@mholt Once you've thought it over, I'd be happy to make the change. I should have some time later in the week to do it.

mholt:
If we do go with 2, and we continue to trim the path like we do now, it only helps in the cases where redirects are performed. Are there any other situations where the full Path would be needed to be known? (Asking out loud here, because this is what I'm currently thinking over.)

ericdreeves:
@mholt The {path} request placeholder.

For example, this Caddyfile:

	example.com:80/foo {
		tls off
		log / stdout "URL: {scheme}://{host}/{path}"
	}

and this request:

	curl -v -L http://example.com/foo
	* Hostname was NOT found in DNS cache
	*   Trying 172.21.0.2...
	* Connected to example.com (172.21.0.2) port 80 (#0)
	> GET /foo HTTP/1.1
	> User-Agent: curl/7.38.0
	> Host: example.com
	> Accept: */*

produce this log:

	example.com_1  | URL: http://example.com//

ericdreeves:
Once I discovered that the {path} placeholder is also affected I became convinced that redirects are a symptom of the underlying issue. Option 2 is indeed elegant but if Caddy continues to provide a middleware (like redir) the ability to do its own redirects then it should somehow have access to the full path.

mholt:
Okay. So maybe we need to rethink this.

Maybe middlewares which access files from disk need to access the path portion of the address in the Caddyfile and trim it. We could provide a helper method (maybe?) in the httpserver package to do it for the middlewares, they just have to call that method instead of accessing r.URL.Path directly.

How does that sound?

If it sounds good, then what about fastcgi and proxy middlewares? Trim or not trim?

ericdreeves:
I like the sound of that, for two reasons. First, middlewares can decide which path is appropriate for the task they're performing. It seems clear that there's a need for both. Second, providing a helper method should provide a better alternative tha obtaining the full path through unsupported means (i.e. RequestURI 😏).

I'll need to dig through the httpserver package a bit more to have a better understanding of where such a method might exist.

mholt:
Sounds good. Thanks for picking up my slack @ericdreeves! Looking forward to seeing where this goes.

ericdreeves:
@mholt Do you think [this comment](https://github.com/mholt/caddy/blob/dc6c986b3fc5863454bad3ff891101b0df855138/caddyhttp/httpserver/server.go#L347-L348) is still accurate? My observation is that RawPath is set only when url.Parse() does it.

mholt:
@ericdreeves From the docs on net/url, it does seem that Parse() is how that gets set. So if we can't guarantee that we call Parse(), it probably is no longer accurate.

ericdreeves:
Thanks for taking a look.

RawPath may be another option, then. The [EscapedPath()](https://golang.org/pkg/net/url/#URL.EscapedPath) method in url package ignores RawPath if it is not a valid escaping of Path, and since httpserver's ServeHTTP cleans Path via [sanitizePath()](https://github.com/mholt/caddy/blob/dc6c986b3fc5863454bad3ff891101b0df855138/caddyhttp/httpserver/server.go#L353) it can't be assumed that RawPath would be consistent with Path anyway.

Storing the sanitized but untrimmed path in RawPath is a better alternative than using RequestURI. Also, it may be simpler than giving middlewares a helper method.

> If it sounds good, then what about fastcgi and proxy middlewares? Trim or not trim?

The proxy middleware [already looks for RawPath](https://github.com/mholt/caddy/blob/dc6c986b3fc5863454bad3ff891101b0df855138/caddyhttp/proxy/proxy.go#L272) when creating its upstream request. In my initial testing I observed that the trimmed path affects the upstream request in the same fashion-perhaps proxy middleware would benefit without changes to the middleware so long as the change in behavior is correct.

One drawback is that using RawPath in this way may not be consistent with its typical use. And while Caddy isn't using RawPath extensively now, it may complicate future use.

>We could provide a helper method (maybe?) in the httpserver package to do it for the middlewares, they just have to call that method instead of accessing r.URL.Path directly.

I've spent some time researching the httpserver package to determine how this would work however it's not clear. Perhaps you could elaborate when you have some time? Appreciate the help!

mholt:
Yikes, this is complicated. I will take a look at this as soon as I can, but I'm going to be out of town and away from the computer for a week starting Monday. Don't let me slow you down. :) Reduce this to the simplest possible way to think about the problem: separate sanitizing, trimming, and using the URL and consider using the Context value on the request if necessary. Ideally, we'll change as little as possible. We should do what makes the most semantic sense for each field in the URL. We want to keep things as possible and obvious for the middlewares as possible, but I understand if it can't be perfectly obvious or the same for every possible situation.

Sorry I can't be more help right now. :) I'm drifting to sleep with a cold and am not quite in the right state of mind to be doing this... ha. I believe in you contributors! 🙌 (Thank you very much for taking such care, too!)

ericdreeves:
Understand completely! Your reply was very helpful, as it reminded me of Context-so, thank you.

I've updated the PR to store the original URL path as a value in the request's context. If we agree on the approach then I'll see about subsequent PRs to fix the {path} request placeholder and proxy middleware.

caddy.go
```diff
@@ -867,3 +867,19 @@ var (
 	// by default if no other file is specified.
 	DefaultConfigFile = "Caddyfile"
 )
+
+// contextKey is a value for use with context.WithValue. Inspired by the
+// net/http source.
+type contextKey struct {
+	name string
+}
```
mholt: I wonder if this belongs in the httpserver package. (My PR, #1430 creates a type called `CtxKey` so maybe we should consider merging these implementations.) I like the simplicity of it just being a string... any reason to use a struct?

caddyhttp/httpserver/server.go
```diff
@@ -265,6 +269,14 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 		}
 	}
 
+	// URL fields other than Path and RawQuery will be empty for most server
+	// requests. Hence, the request URL is updated with the scheme and host
+	// from the virtual host's site address.
+	if vhostURL, err := url.Parse(vhost.Addr.String()); err == nil {
+		r.URL.Scheme = vhostURL.Scheme
+		r.URL.Host = vhostURL.Host
```
mholt: Is this just for convenience in upstream middlewares?

caddyhttp/staticfiles/fileserver.go
```diff
@@ -80,17 +82,34 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
 	if d.IsDir() {
-		// Ensure / at end of directory url
-		if !strings.HasSuffix(url, "/") {
-			Redirect(w, r, path.Base(url)+"/", http.StatusMovedPermanently)
+		// Ensure / at end of directory url. If the original URL path is
+		// used then ensure / exists as well.
+		if !strings.HasSuffix(r.URL.Path, "/") {
+			toURL, _ := url.Parse(r.URL.String())
+
+			path, ok := r.Context().Value(caddy.URLPathContextKey).(string)
```
mholt: Hmm, do we want to use the context's path value here?

A Caddyfile like:

```
example.com/subpath
root /www/mysite
```

And then a request to `example.com/subpath/test.html` should load `/www/mysite/test.html` -- is that what you would expect? I think that's how it's currently implemented, anyway.

caddyhttp/staticfiles/fileserver.go
```diff
@@ -80,17 +82,34 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
...
+		if strings.HasSuffix(r.URL.Path, "/") {
+			toURL, _ := url.Parse(r.URL.String())
+
+			path, ok := r.Context().Value(caddy.URLPathContextKey).(string)
+			if ok && strings.HasSuffix(path, "/") {
+				toURL.Path = path
+			}
+			toURL.Path = strings.TrimSuffix(toURL.Path, "/")
+
+			http.Redirect(w, r, toURL.String(), http.StatusMovedPermanently)
```
mholt: The diff shows that this used to be `Redirect(...)` not `http.Redirect(...)`. Why the change?

caddy.go
mholt: @ericdreeves FYI, I have merged the branch that contains httpserver.CtxKey so you can now use that.

ericdreeves: Will do!

caddyhttp/httpserver/server.go
ericdreeves: The URL in request r has neither the scheme nor the host.

http.Redirect will [set the Location header to a path](https://github.com/golang/go/blob/64c02460d703f718a44326646d19c417eeb1999a/src/net/http/server.go#L2011) (not absolute URI) when
the provided urlStr is missing both.

Since the virtual host's site address has both, and the virtual host
configuration was matched to r.URL.Path, r.URL is updated so that invoking
http.Redirect will:

1. Set the Location header to an absolute URI
2. Redirect to a URI with the correct scheme

caddyhttp/staticfiles/fileserver.go
ericdreeves: If it's guaranteed that no middleware will modify r.URL.Path then I agree that we can instead use context's path.

mholt: Pretty sure that can't be guaranteed. Is the context storing the original URI or the modified one?

ericdreeves: I'll verify the behavior with the Caddyfile that you provided above.

caddyhttp/staticfiles/fileserver.go
ericdreeves: staticfiles.Redirect does not convert to absolute URI.

While refreshing my memory regarding the reason for this change, I was reminded that the browse package uses staticfiles.Redirect for a similar reason: the path refers to a directory but doesn't have a trailing '/'. We should investigate whether the browse package relies on always redirecting to a relative path.

ericdreeves:
Thanks for being patient as well! I've provided some details that I hope address your questions. I have some behavior to verify, and a conflict to resolve. I should be able to address both over the weekend.

caddyhttp/staticfiles/fileserver.go
ericdreeves: The context stores the original URI.

caddyhttp/httpserver/server.go
mholt: That's cool, but I don't think a fully qualified URL is needed; the path seems to do okay. I guess setting them isn't bad either. We can try it.

caddyhttp/staticfiles/fileserver.go
ericdreeves: Your Caddyfile does produce the expected behavior.

caddy.go
ericdreeves: As is, the staticfiles package cannot use CtxKey because importing httpserver would result in an import cycle. My rationale for putting my contextKey type in the caddy package was not only to avoid the import cycle but also to make the predefined keys available to the caddytls package where HTTPChallengeHandler is implemented.

If you don't like the idea of putting CtxKey definition in the caddy package then how about a new package within caddyhttp? Then, httpserver and other middlewares (staticfiles, fastcgi, basicauth, etc.) could each import the package and have access to the predefined keys declared therein.

I have some local commits that prove it out, but wanted to get others' take before pushing.

mholt: @ericdreeves Ah, I didn't think about that. Yes, we can probably put `CtxKey` in the `caddy` package. Good idea. I don't see any negative consequences of that.

mholt:
Sorry for my delay. Working through some unexpected things, but will get to this soon!

caddyhttp/staticfiles/fileserver.go
mholt: @ericdreeves If I merge this PR this week, could you look into that?

ericdreeves: @mholt This?

> the browse package uses staticfiles.Redirect for a similar reason: the path refers to a directory but doesn't have a trailing '/'. We should investigate whether the browse package relies on always redirecting to a relative path.

If so then yes, I can look into it. This week is very busy but late this week / early next should be possible. I did not remove `staticfiles.Redirect` as part of this PR since the browse package still depends on it.

mholt: @ericdreeves Yep - that. Sounds good, no hurry. But it would be nice to get that consistent between this package and the browse package before the next release (I think it'll be at least a few weeks, so you should be good). Thank you!!

PR #1335, opened 2017-01-08
Generate meta elements from prelude items description and keywords

pbeckmann:
At the moment this functionality is limited to the allowed meta tags from `getAllowedMeta`

There are at the moment no test cases, but i'll look into them, if they are possible and useful.

Criticism and discussion is very welcome.

Should fix https://github.com/mholt/caddy/issues/1326


pepa65:
I still think keywords is good for consideration, not for search engines or SEO, but because they could be taken as tagnames. It makes sense semantically.
Language is not very useful, because there are a number of more functional ways to set language.

pbeckmann:
Do we should test if somebody already use a front matter with the name `meta`?
At the moment we would break that usage.

pbeckmann:
@mholt Shall i extend the test? I'm not realy sure, cause i didn't wanted to test too much around the functionality.

caddyhttp/markdown/process.go
```diff
@@ -75,3 +83,12 @@ func (c *Config) Markdown(title string, r io.Reader, dirents []os.FileInfo, ctx
...
 }
+
+func getAllowedMeta() []string {
+	return []string{
+		"author",
+		"copyright",
+		"description",
+		"subject",
+	}
+}
```
mholt: This could just be a variable, no? `recognizedMetaTags` or something.

mholt:
> Shall i extend the test? I'm not realy sure, cause i didn't wanted to test too much around the functionality.

When in doubt, usually yes. :) What is the behavior when meta is specified?

pbeckmann:
At the moment we would override the original meta value. That question implies, that I'll implement a solution tomorrow ^^

pepa65:
If a default meta has been given in a template, the most commonly expected behaviour would be that the page settings override the template. (That is, after this change is part of the release and the documentation... Right now it could potentially lead to surprises, but likely nothing that would cause serious breakage, unless the javascript is making use of certain meta values.)

pbeckmann:
We do not touch user defined templates. We are just changed the default template and the user has the possibility to create their meta tags in a simpler way! So NO we don't override meta tags in templates.

caddyhttp/markdown/template.go
```diff
@@ -32,6 +33,7 @@ func execTemplate(c *Config, mdata metadata.Metadata, files []FileInfo, ctx http
 		Doc:     mdata.Variables,
 		Styles:  c.Styles,
 		Scripts: c.Scripts,
+		Meta:    mdata.Variables["meta"].(map[string]string),
```
tw4452852: `Meta` is just one of fields of `Doc`, I suggest no to introduce the new one.

pbeckmann:
Hey @mholt,

Basicly this function adds an array with all front matters, that are known meta values.
So every user of the markdown middleware could just add 

`{{range $key, $val := .Meta}}
	<meta name="{{$key}}" content="{{$val}}">
{{end}}`

to it's template and got all the meta values from the front matters :)

Greetings from Germany

P.S.: If you think this is a stupid idea, i can live with it, if you close the PR ^^

caddyhttp/markdown/process_test.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		toml := "+++"
+		for key, val := range meta {
+			toml = toml + "\n" + key + "= \"" + val + "\""
+		}
+		toml = toml + "\n+++"
+
+		res, _ := config.Markdown("Test title", strings.NewReader(toml), []os.FileInfo{}, httpserver.Context{})
+		sRes := string(res[:])
```
mholt: Just FYI I don't think the `[:]` is necessary

mholt:
@tw4452852 Do you approve of this PR or are there any other critiques you have?

caddyhttp/markdown/process.go
```diff
@@ -63,6 +70,14 @@ func (c *Config) Markdown(title string, r io.Reader, dirents []os.FileInfo, ctx
 		mdata.Variables["title"] = title
 	}
 
+	// move available and valid front matters to the meta values
+	meta := make(map[string]string)
+	for _, val := range recognizedMetaTags {
+		if _, ok := mdata.Variables[val]; ok {
+			meta[val] = mdata.Variables[val].(string)
```
tw4452852: Two lines above could be combined:

```
if var, ok := mdata.Variables[val]; ok {
    meta[val] = var.(string)
```

tw4452852:
Except one minor refinement, LGTM.

pbeckmann:
Hey,

@tw4452852 I've pushed it :) 

Thank you both, for your guidance and patience :+1: 

Greetings

mholt:
Thank you @pbeckmann ! Nice work, really appreciate you sticking to it.

Issue #1332, opened 2017-01-07
Unable to setup reverse-proxy for weechat websocket


joelnb:
Following up on a [forum post](https://forum.caddyserver.com/t/proxy-weechat-relay-with-caddy/801/4) about something which seems like a possible issue with caddy.

On that thread I suggested it was related to headers sent by Caddy which could not be removed but I am now not convinced that is the issue after further testing. Should I create a separate issue for the fact that those headers cannot be removed? As it could possibly do with being added to the docs anyway.

### 1. What version of Caddy are you running (`caddy -version`)?

Caddy 0.9.4

### 2. What are you trying to do?

Setup a reverse-proxy for weechat equivalent to this nginx config:

```
map $http_upgrade $connection_upgrade {
    default upgrade;
    '' close;
}
upstream weechat {
    server 127.0.0.1:8001;
}

server {
    listen       9000 ssl;
    server_name  _;

    ssl_certificate ...;
    ssl_certificate_key ...;

    location /weechat {
        proxy_pass http://weechat;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_read_timeout 8h;
    }
}
```

### 3. What is your entire Caddyfile?

```text
irc.something.com:9000 {
    proxy /weechat 127.0.0.1:8001 {
        websocket
    }

    log stdout
    errors stdout
}
```

And also:

```
irc.something.com:9000 {
    proxy /weechat 127.0.0.1:8001 {
        header_upstream -Accept-Encoding
        header_upstream -User-Agent
        header_upstream -X-Forwarded-For

        websocket
    }

    log stdout
    errors stdout
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

I'm running using the systemd service file:

```
[Unit]
Description=Caddy HTTP/2 web server
Documentation=https://caddyserver.com/docs
After=network-online.target
Wants=network-online.target
; Wants=network-online.target systemd-networkd-wait-online.service

[Service]
Restart=on-failure
StartLimitInterval=86400
StartLimitBurst=5

; User and group the process will run as.
User=www-data
Group=www-data

; Letsencrypt-issued certificates will be written to this directory.
Environment=CADDYPATH=/etc/ssl/caddy

; Always set "-root" to something safe in case it gets forgotten in the Caddyfile.
[...]

PR #1338, opened 2017-01-09
proxy: capitalize some request headers

tw4452852:
Go std lib will [normalize](https://github.com/golang/go/blob/master/src/net/textproto/reader.go#L600) the request header, so something like `Sec-WebSocket-Key` will be changed to `Sec-Websocket-Key`. while some servers (e.g. weechat in issue #1332) may depend on the original one case-sensitive.

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -194,6 +194,28 @@ func (rp *ReverseProxy) UseInsecureTransport() {
 	}
 }
 
+var capitalHeaders = map[string]string{
+	strings.ToLower("Sec-WebSocket-Key"):        "Sec-WebSocket-Key",
+	strings.ToLower("Sec-WebSocket-Extensions"): "Sec-WebSocket-Extensions",
+	strings.ToLower("Sec-WebSocket-Accept"):     "Sec-WebSocket-Accept",
+	strings.ToLower("Sec-WebSocket-Protocol"):   "Sec-WebSocket-Protocol",
+	strings.ToLower("Sec-WebSocket-Version"):    "Sec-WebSocket-Version",
+}
+
+func capitalizeHeaders(headers http.Header) {
+	// Set the request headers using the capitalization for names.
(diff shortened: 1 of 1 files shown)


joelnb:
I've tested your branch and I can confirm this fixes issue #1332 for me completely. Thanks for your work on this!

mholt:
Umm, wait a sec. I'm not saying No to this PR, but WeeChat is broken if it treats headers as case-sensitive. Why should we fix something in Caddy that is not broken?

(Thanks for working on this @tw4452852! I'm not criticizing your fix, I'm upset at weechat.)

tw4452852:
@mholt Maybe [weechat shouldn't treats headers as case-sensitive](https://github.com/weechat/weechat/blob/master/src/plugins/relay/relay-websocket.c#L153).
But I actually find [something similar in gorilla](https://github.com/gorilla/websocket/blob/master/client.go#L221).

mholt:
Gorilla caved... I'm torn on this. I'll have to think about it!

mholt:
Thanks for your thorough research on this @tw4452852 -- I've opened an issue upstream. I'll leave this open for a little bit but I will probably close it. If this change is what causes weechat to work with Caddy, then it's a bug in weechat and not in Caddy.

mholt:
It looks like they're working on it upstream! 🙌  I'm going to close this now, since it looks like weechat 1.7 will have the fix. Thank you for your work on it!

tobya:
Interesting.  While I can understand @mholt point of view and I agree with it, I also think it is not a good idea for caddy to modify headers as they go through.  

My understanding here is that caddy is modifying the headers on the request.

Should we allow @tw4452852 fix and also have weechat fix their side?

mholt:
> My understanding here is that caddy is modifying the headers as the request is sent downstream.

The standard library is -- they are being normalized according to convention. Caddy itself doesn't change them, per-se. In HTTP/2, header field names are lowercased anyway. Let's just let weechat fix their end, and if Go shouldn't be modifying the casing of header fields, then that should be an issue with the Go standard library.

PR #1340, opened 2017-01-10
Enable golint

wendigo:
As per last PR of mine I would like to suggest to enable golint in CI so contributors will get feedback faster


mholt:
Huh, Caddy used to fail golint because one of its dependencies requires an interface that is not Golint-compliant, I wonder why the warnings don't cause a failure anymore.

I guess since that's the case we can merge this... I will do that later today!

.travis.yml
```diff
@@ -27,6 +27,7 @@ script:
   - ineffassign .
   - go vet ./...
   - go test -race ./...
+  - $HOME/gopath/bin/golint ./...
 
 after_script:
   - golint ./...
```
mholt: Why do we run golint during and after? I also don't think you'll need the full path like that.

mholt:
@wendigo I'm going to close this, if that's alright. Maybe you just didn't notice that golint is already being run?

PR #1343, opened 2017-01-10
Add misspell to travis build

bengadbois:
This can catch some of things like https://github.com/mholt/caddy/pull/1342 automatically


mholt:
Thanks! Is there a way to get it to exit with a code other than 0 if there are misspellings?

bengadbois:
Now it'll exit 2 upon finding an error, which Travis should catch

bengadbois:
Alright, it's about the same file size and total word count. Hopefully it doesn't cause issues, but would make sense to treat it like golint.

Issue #1328, opened 2017-01-05
Add a cli parameter to validate a Caddyfile


enko:
At our company we are deploying caddy via Gitlab CI if someone pushes a new Caddyfile to a repository. It would help us big times if caddy had a parameter like `caddy -validate /etc/Caddyfile` that returns a non-zero code if something obvious (f.e. there is a syntax error) is wrong with the caddyfile. In that case the Caddyfile would not be deployed and the caddy instance would not crash.

You may ask if I could just start the caddy with the config on the ci server. This will not work since we are using let's encrypt for this config and thous caddy cant activate the privacy features and will fail to start.

mholt:
So if we implemented `-validate` to only parse the Caddyfile (thus reporting syntax errors), would that be sufficient?

enko:
@mholt Yes, that would be excellent.

tobya:
I'll take a look at this

tobya:
Still working on this but the implementation is designed to be as follows

Addition of a new `-validate` command line parameter which should parse the caddyfile but not run the server.  Also no TLS certs should be requested.

 - If there is **an error** in the caddyfile caddy will exit with an error as normal
 - If there is **no error** in the caddyfile caddy will exit with no further output. 

@enko do you have any comments on the proposed implementation.

justinmayer:
If there is no error, are there any downsides to displaying something like the following?

    Caddyfile syntax validated successfully.

Just curious. 😄

mholt:
Maybe not. But the "unixy" way of doing things is to not output anything on success.

enko:
@tobya yeah, thats what I need, just the error codes are fine.

tobya:
If you can build from source this is implemented in branch issue1328d #1344

tobya:
This has been implemented & merged in #1344

PR #1344, opened 2017-01-10
Add a cli parameter to -validate a Caddyfile. Issue #1328

tobya:
This seems to be a straightforward implementation of #1328 .

I created a new flag -validate which will simply exit after the Caddyfile is parsed.  

If Caddyfile is valid will log

 > [INFO] Caddyfile Valid

otherwise if Caddyfile is not valid the system will output an error as usual.

Questions?
==

1. Should a value other than 0 be returned from 

    os.Exit(0);

if the caddyfile is not valid?

2. Also validate will output even if -log is not present.   I assume this is the desired behaviour.

3. I couldn't see any tests for flags, so I didn't add one.


caddy.go
```diff
@@ -182,7 +182,7 @@ func (i *Instance) Restart(newCaddyfile Input) (*Instance, error) {
 	newInst := &Instance{serverType: newCaddyfile.ServerType(), wg: i.wg}
 
 	// attempt to start new instance
-	err := startWithListenerFds(newCaddyfile, newInst, restartFds)
+	err := startWithListenerFds(newCaddyfile, newInst, restartFds, false)
```
tobya: I assume there will never be a scenario where justValidate is expected to be true on a restart

mholt: Correct. I can't think of one.

caddy.go
```diff
@@ -421,13 +422,13 @@ func (i *Instance) Caddyfile() Input {
 // Start starts Caddy with the given Caddyfile.
 //
 // This function blocks until all the servers are listening.
-func Start(cdyfile Input) (*Instance, error) {
+func Start(cdyfile Input, justValidate bool) (*Instance, error) {
```
mholt: I don't want to change the signature of Start().

caddy/caddymain/run.go
```diff
@@ -40,6 +40,7 @@ func init() {
 	flag.StringVar(&revoke, "revoke", "", "Hostname for which to revoke the certificate")
 	flag.StringVar(&serverType, "type", "http", "Type of server to run")
 	flag.BoolVar(&version, "version", false, "Show version")
+	flag.BoolVar(&validate, "validate", false, "Validate Caddyfile and return")
```
mholt: "Parse the Caddyfile but do not start the server"

mholt:
Also, to address some of your questions, here's my take on `-validate`:

- If the Caddyfile is "valid", exit code 0 and don't print anything. (I am not sure how useful logging an info message would be, either. Let's not do that unless we get specific requests with good reasons for it.)

- If the Caddyfile is not valid, exit code 1 with the error message.

- Don't worry about flags for now.

tobya:
@mholt I have this working now in a different manner.  

However there is a difficulty in that `caddy.executeDirectives` will attempt to retrieve TLS certs via acme if required.  I don't think this is expected with the  -validate flag.

Do you have a suggestion for how to avoid this?

caddy/caddymain/run.go
```diff
@@ -104,6 +106,18 @@ func Run() {
...
+		_, err := caddyfile.Parse(caddyfile.Path(), bytes.NewReader(caddyfile.Body()), validDirectives)
+		if err != nil {
+			mustLogFatalf(err.Error())
+		}
+
+		//Caddyfile is valid so exit without starting
+		os.Exit(0)
+
+	}
+
```
tobya: Ignore this section I remove it in the next commit

tobya:
@mholt `executeDirectives` seems to be very abstract and runs through all directives executing them.  I am unsure if there is a straightforward way to prevent the tls code from being run.

mholt:
Maybe we can extract some of `executeDirectives` out into a separate function -- the inside loop over the server blocks -- and then skip [the part that does the directive callbacks](https://sourcegraph.com/github.com/mholt/caddy@32681adea67511d103c4f5eb6245e1196d53343c/-/blob/caddy.go#L589-597). We could pass another argument to executeDirectives so that if it's false, it doesn't execute the callbacks.

Would that work?

tobya:
I'll take a look and see if that does the trick

caddy/caddymain/run.go
```diff
@@ -105,11 +105,12 @@ func Run() {
...
 
 	if validate {
-
-		_, err := caddy.ValidateCaddyFile(caddyfileinput, nil)
+		justValidate := true
+		_, err := caddy.ValidateAndExecuteDirectives(caddyfileinput, nil, justValidate)
 		if err != nil {
 			mustLogFatalf(err.Error())
 		}
+		log.Println("[INFO] Caddyfile Valid")
```
tobya: I know you asked to remove the INFO log earlier.  But I felt that if the user starts with -log they should receive an INFO log item

tobya:
This seems to work well.  I have done some basic testing by creating various server blocks and creating errors and seems to all be reported the same way with -validate and without.  

ValidateAndExecuteDirectives is a new function that is exported.  I have added the justValidate parameter to the executeDirectives function.  I will add a PR to caddyserver.com repo with update to documentation

tobya:
Added PR for caddyserver.com update https://github.com/caddyserver/caddyserver.com/pull/103

caddy.go
```diff
@@ -357,6 +357,7 @@ type AfterStartup interface {
 // specify any default Caddyfile value, then an empty Caddyfile
 // is returned. Consequently, this function never returns a nil
 // value as long as there are no errors.
+//func LoadCaddyfile(serverType string, justValidate bool) (Input, error) {
```
mholt: Stray comment ;)

caddy.go
```diff
@@ -516,9 +493,48 @@ func startWithListenerFds(cdyfile Input, inst *Instance, restartFds map[string]r
 	return nil
 }
 
-func executeDirectives(inst *Instance, filename string,
-	directives []string, sblocks []caddyfile.ServerBlock) error {
+func ValidateAndExecuteDirectives(cdyfile Input, inst *Instance, justValidate bool) (*Instance, error) {
+
+	//If parsing only inst will be nil, create an instance for this function call only.
```
mholt: Hmm, I'm surprised `go fmt` doesn't put a space between the `//` and `If`, but please do. :)

caddy/caddymain/run.go
```diff
@@ -98,13 +99,24 @@ func Run() {
...
 	}
 
+	if validate {
+		justValidate := true
+		_, err := caddy.ValidateAndExecuteDirectives(caddyfileinput, nil, justValidate)
+		if err != nil {
+			mustLogFatalf(err.Error())
+		}
+		log.Println("[INFO] Caddyfile Valid")
+		//Caddyfile is valid so exit without starting
```
mholt: Same thing here about the comment spacing, but I don't think this comment is even necessary

caddy/caddymain/run.go
```diff
@@ -98,13 +99,24 @@ func Run() {
...
+	caddyfileinput, err := caddy.LoadCaddyfile(serverType)
 	if err != nil {
 		mustLogFatalf(err.Error())
 	}
 
+	if validate {
+		justValidate := true
+		_, err := caddy.ValidateAndExecuteDirectives(caddyfileinput, nil, justValidate)
+		if err != nil {
+			mustLogFatalf(err.Error())
```
mholt: Hmm, just to be safe, use `"%v"` as the first argument, then `err` as the second. (If an error message has "%s" in it, it could create problems I think.) I realize now I didn't do that above and below, as I hadn't thought of this before -- maybe you could do me a favor and change those too? :smile:

caddy.go
```diff
@@ -516,9 +492,48 @@ func startWithListenerFds(cdyfile Input, inst *Instance, restartFds map[string]r
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = executeDirectives(inst, cdyfile.Path(), stype.Directives(), sblocks, justValidate)
+	if err != nil {
+		return nil, err
+	}
+
+	return inst, nil
```
mholt: One more thing I just noticed; I don't this function needs to return an *Instance. Just an error. Right?

tobya: ahh yes.

caddy.go
```diff
@@ -516,9 +492,48 @@ func startWithListenerFds(cdyfile Input, inst *Instance, restartFds map[string]r
...
+	if err != nil {
+		return err
+	}
+
+	err = executeDirectives(inst, cdyfile.Path(), stype.Directives(), sblocks, justValidate)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
mholt: Just for future reference, this could be condensed to `return executeDirectives(...)`

mholt:
Thanks so much!

PR #1353, opened 2017-01-14
Add {whenISO} to record timestamp in ISO 8601 format in UTC.

maruel:
ISO 8601 is the standard time format and is easy to parse.

This change assumes users desiring ISO 8016 generally prefer UTC for simplicity.
This results in {whenISO} to be significantly shorter than {when}:
{when}    = "02/Jan/2006:15:04:05 +0000"
{whenISO} = "2006-01-02T15:04:12Z"

Add unit test to verify both, as there was no unit test for {when}.

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -23,6 +23,8 @@ var requestReplacer = strings.NewReplacer(
 	"\n", "\\n",
 )
 
+var now = time.Now
+
 // Replacer is a type which can replace placeholder
 // substrings in a string with actual values from a
 // http.Request and ResponseRecorder. Always use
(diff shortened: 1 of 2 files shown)


caddyhttp/httpserver/replacer.go
```diff
@@ -23,6 +23,8 @@ var requestReplacer = strings.NewReplacer(
 	"\n", "\\n",
 )
 
+var now = time.Now
```
mholt: This is a very elegant way to make time testable. :+1:

caddyhttp/httpserver/replacer.go
```diff
@@ -249,7 +251,9 @@ func (r *replacer) getSubstitution(key string) string {
 	case "{uri_escaped}":
 		return url.QueryEscape(r.request.URL.RequestURI())
 	case "{when}":
-		return time.Now().Format(timeFormat)
+		return now().Format(timeFormat)
+	case "{whenISO}":
```
mholt: How do you feel about `{when_iso}` to be consistent with a couple of other placeholders?

maruel: Done and on caddyserver.com PR too.

mholt:
Thank you, @maruel! Appreciate your contribution.

PR #1354, opened 2017-01-15
Set session ticket keys properly (fixed in Go 1.8)

mholt:
The reason this line was needed in the first place was due to https://github.com/golang/go/issues/15421.

This change should resolve the race in #1157 as well.


wmark:
Yes, this line can be removed for Go 1.8, but I am not entirely convinced it eliminates a race condition in practice, or the error you are trying to hunt down in the other thread. Could be premature to close the other one.

mholt:
I agree that there is more than one thing going on in that issue. Still actively investigating it (it's why I haven't gotten more done over here with anything else.)

PR #1356, opened 2017-01-17
proxy: Fixed #1352: invalid use of the HTTP hijacker

### LINKED ISSUE #1352: caddyhttp/proxy: invalid use of Hijack in reverseproxy.go
Hello, in preparation for Go1.8, I detected this misuse of the http.Hijacker API.

[reverseproxy.go makes a call to Hijack](https://github.com/mholt/caddy/blob/4babe4b201eef3e9794851b74f734a03338f7a20/caddyhttp/proxy/reverseproxy.go#L241), but ignores the returned `bufio.ReadWriter` and proceeds to directly use the connection. In Go1.8, the probability that data is buffered in the `bufio.Reader` is increased, such that there is a higher change that this logic fails. The proper fix is to handle the data in the read buffer (accessed via `brw.Reader.Peek(brw.Reader.Buffered())`) and forward that to the `backendConn` before performing the pair of `io.Copy` calls.

See https://golang.org/cl/35232 for more information.

\cc @mholt

lhecker:


mholt:
Very nice. Do you think we can get a test to accompany this? Will ping @dsnet in case you would like help, if he is willing & able.

dsnet:
You could try something similar to the test added in https://golang.org/cl/35232.

lhecker:
I'm not sure how we'd do that in a test, since we can't just add this part to the proxy to reliably enforce that some bytes are buffered up:

```
select {
case <-w.(CloseNotifier).CloseNotify():
case <-time.After(5 * time.Second):
	t.Error("timeout")
	return
}
```

lhecker:
@mholt IMHO you should just merge it, because AFAIK there is no reliable and reproducible way to write a test for this. You'd have to add something like the lines 5197-5202 from [CL 35232](https://go-review.googlesource.com/#/c/35232/5/src/net/http/serve_test.go) and we can't do that in our proxy code right? This means that this PR is sadly only correct in theory, but since it's a rather straightforward implementation I don't think it can hurt.

If anyone has an idea how to write a test though I'll try to implement it of course. 🙂

caddyhttp/proxy/reverseproxy.go
```diff
@@ -260,8 +260,23 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request,
...
+		//
+		// NOTE: Hijack() sometimes returns buffered up bytes in brw which
+		// would be lost if we didn't read them out manually below.
+		if brw != nil {
+			if n := brw.Reader.Buffered(); n > 0 {
+				rbuf, err := brw.Reader.Peek(n)
+				if err != nil {
+					return err
+				}
+				backendConn.Write(rbuf)
```
mholt: Why is the Peek necessary? Wouldn't it be sufficient to write them out without peeking?

lhecker: The `brw` is a alternative, low-level interface to the `conn`. If we `Read()` more from it than we have buffered up it will delegate the `Read()` to the `conn`. We don't want that, because we already do a regular `io.Copy` from the `conn` right after. What we want instead is `brw`'s internal `buf` and `Peek()` is the most straightforward choice for that (as it guarantees to only return buffered bytes). After we read those bytes out the `brw` can be garbage collected since we actually don't need it at all (apart from circumventing this design bug in Go's stdlib).

mholt:
I'm about to  push merge but my gut is asking that I double-check to make sure that you've at least tested this somehow... or somebody has seen that this works? I mean, I believe it, by looking at it, but... I've been bitten before by accepting changes by just assuming/hoping they work. 😅

caddyhttp/proxy/reverseproxy.go
dsnet: This implementation looks correct.

It could be simplified by avoiding the check if `n > 0` since its okay to `Peek` zero bytes:
```go
rbuf, err := brw.Reader.Peek(brw.Reader.Buffered())
```

lhecker: I guess you could consider it a premature optimization, because it only prevents executing whatever is done inside `Peek()` and `Write()`.
On the other hand doing nothing in case nothing needs to be done is something I've learned to be a very valuable optimization. Reason being that it's a) usually very straightforward to implement & understand and b) it automatically prevents regressions in the future if the internal implementations you rely on ever change.

lhecker:
@mholt Unfortunately I didn't test it. I've never experienced problems of "missing bytes" with Caddy myself and I don't know how I'd create even a test setup for this as I don't really understand what's causing Go to read bytes into the buffer returned by the Hijacker.

As far as I understand CL 35232 all of this is due to Go reading from the connection in the background and `Hijack()` stopping this process. You could thus say that the only reason some bytes might already be buffered up is basically due to a race condition. I'm not sure how to test for this without waiting for the `CloseNotify` or how to test for this in a general test setting (without spending more time on it than I can on this right now). If my understand above is correct I'd consider this as a theoretical issue which thus has a theoretical solution.

PR #1358, opened 2017-01-18
proxy: use a new context for the outgoing request

tw4452852:
fix issue #1345

### DIFF
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -3,6 +3,7 @@ package proxy
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"fmt"
 	"io"
@@ -11,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/http/httptrace"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


mholt:
Interesting. I would have to learn more about Context to accept this without a test; for now, let's see if @Tester98 can confirm whether this solves the issue. I think it would still be good to have a test, if it's not too difficult.

@tw4452852 Thank you as always for the PR. :)

tw4452852:
Add a test.

mholt:
Thanks Tw! I'm about ready to merge this. But before I do, why do you think the standard lib doesn't use context at all with their implementation of httputil.ReverseProxy? Also, is this change consistent with [the docs for the `context` package](https://golang.org/pkg/context/)? It states:

> The chain of function calls between them must propagate the Context, optionally replacing it with a derived Context created using WithCancel, WithDeadline, WithTimeout, or WithValue. When a Context is canceled, all Contexts derived from it are also canceled.

tw4452852:
`httputil.ReverseProxy` indeed uses [the context which derived from the incoming request](https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L157). I have ever adopt the same way, but it adds additional complexity(we can't recognize whether the request is canceled according to the error `transport.RoundTrip` returns. [This error isn't export](https://github.com/golang/go/blob/master/src/net/http/transport.go#L1806)). So I just give it a try even the original request is canceled. I think this isn't harmful.

Anyway if you think it is necessary to cancel the outgoing request if the incoming request is cancelled. I will revise this patch a bit.

mholt:
Hey, thanks. It's good to have that documented in this issue, so I appreciate the clarification. Let's start simple and I'll accept this now. If it comes up again, at least we know what might have to happen to fix something.

PR #1360, opened 2017-01-18
proxy: fix TestReverseProxy failure on go tip

tw4452852:
Because of [this commit](https://github.com/golang/go/commit/6e36811c37399d60cbce587b7c48e611009c5aec) on go tip,
it will probe the request's body to determine whether to use chunked transfer
encoding which trailers depend on it.
So we just offer a non empty body to make trailers work.

fix issue #1359

### DIFF
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -78,7 +78,7 @@ func TestReverseProxy(t *testing.T) {
 	}
 
 	// create request and response recorder
-	r := httptest.NewRequest("GET", "/", nil)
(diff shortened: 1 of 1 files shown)


tw4452852:
BTW @mholt , Is there any reason why allow failures on go tip?

mholt:
@tw4452852 We allow failures on tip because it changes rapidly out from under our feet; we still want to know when Caddy may break in a future version of Go, but we don't want it to block pull requests for stable versions. Sometimes tip has new bugs too, I've seen it before.

Thanks for the fix :) I'm gonna re-run the build since Travis seems to be stuck...

mholt:
Nice find! Thanks for the fixed test.

Issue #1185, opened 2016-10-14
Should {path} in access logs show value for {>Caddy-Rewrite-Original-URI}


tobya:
This is an related to  #1153 assuming Pull Request #1184 is merged.

In an access log the placeholder  `{>Caddy-Rewrite-Original-URI}` can be used with a fastcgi site that does a rewrite to output the origional path requested.  {path} will output the path after redirection

eg.

with a caddyfile

```
localhost:1234 {
log /  access.log "{path} {>Caddy-Rewrite-Original-URI} "

rewrite / {
 r (.*)
 to /redir.php
}

 fastcgi / localhost:34345  php
 root  D:\webroot\phpcgi
}
```

for the url  http://localhost:1234/customers/department/sales/2233 the log will show

> /redir.php /customers/department/sales/2233 

So, there is an argument here for {path} in the caddy file to log the same value as {>Caddy-Rewrite-Original-URI} as really it is the requested path that is of interest in a log rather than the rewritten path.

Is it worth considering another log placeholder such as `{rewrite_path}` which would contain what `{path}` currently contains, ie. the path after rewriting, and have `{path}` output what `{>Caddy-Rewrite-Original-URI}` does?

abiosoft:
This is my view as well. https://github.com/mholt/caddy/issues/1153#issuecomment-250827388.

elcore:
@tobya @abiosoft like this?

```
diff --git a/caddyhttp/httpserver/replacer.go b/caddyhttp/httpserver/replacer.go
index 02ad54f..bcaa2d8 100644
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -221,8 +221,12 @@ func (r *replacer) getSubstitution(key string) string {
 		}
 		return host
 	case "{path}":
-		return r.request.URL.Path
+		return r.request.Header.Get("Caddy-Rewrite-Original-URI")
 	case "{path_escaped}":
+		return url.QueryEscape(r.request.Header.Get("Caddy-Rewrite-Original-URI"))
+	case "{rewrite_path}":
+		return r.request.URL.Path
+	case "{rewrite_path_escaped}":
 		return url.QueryEscape(r.request.URL.Path)
 	case "{query}":
 		return r.request.URL.RawQuery
diff --git a/caddyhttp/redirect/redirect_test.go b/caddyhttp/redirect/redirect_test.go
index 27998ab..a71cb3a 100644
[...]

PR #1364, opened 2017-01-18
Introduce new Replacer field {rewrite_path} issue #1185

tobya:
This is a continuation of implementation of #1185 started by  @elcore 

 The basic concept here is that we introduce two new replacer variables  {rewrite_path} and {rewrite_path_escaped} and change the behaviour of {path} and {path_escaped}

{rewrite_path} will always output the path that is the result of a rewrite block or the plain vanilla URI if no rewrite block exists.

{path} will always output the requested URI before any rewrite blocks are executed.  This is the same as {>Caddy-Rewrite-Original-URI} if a rewrite block exists.

Prior to this change {path} would output the rewritten path when a rewrite block was present.


caddyhttp/rewrite/rewrite_test.go
```diff
@@ -24,7 +24,8 @@ func TestRewrite(t *testing.T) {
 	regexps := [][]string{
 		{"/reg/", ".*", "/to", ""},
 		{"/r/", "[a-z]+", "/toaz", "!.html|"},
-		{"/url/", "a([a-z0-9]*)s([A-Z]{2})", "/to/{path}", ""},
+		{"/path/", "[a-z0-9]", "/to/{path}", ""},
+		{"/url/", "a([a-z0-9]*)s([A-Z]{2})", "/to/{rewrite_path}", ""},
```
tobya: @elcore Could you explain why it was necessary to add this extra regex item?

tobya: @elcore Would you be able to explain these tests, I'm not sure I follow.

tobya:
I am currently getting these errors @elcore .  I think I need a quick discussion around if I need to change the tests.

> --- FAIL: TestRewrite (0.00s)
        rewrite_test.go:113: Test 18: Expected URL to be '/to/' but was '/to/path/a1b2c'
        rewrite_test.go:113: Test 19: Expected URL to be '/to/' but was '/to/path/d3e4f'
--- FAIL: TestTo (0.00s)
        to_test.go:44: Test 9: expected /p/?url=http:// found /p/test?url=http://
FAIL
FAIL    github.com/mholt/caddy/caddyhttp/rewrite        0.754s

tobya:
I have fixed the tests.  I think the issue was that prior to 	5a27e2b {path} would be empty if a rewrite had not taken place.  Now {path} will always have a value.

tobya:
I need to make some further changes before completion

caddyhttp/httpserver/replacer.go
```diff
@@ -223,8 +223,22 @@ func (r *replacer) getSubstitution(key string) string {
...
-		return r.request.URL.Path
+		// if a rewrite has happened, the original URI should be used as the path
+		// rather than the rewritten URI
+		path := r.request.Header.Get("Caddy-Rewrite-Original-URI")
+		if path == "" {
+			path = r.request.URL.Path
+		}
+		return path
 	case "{path_escaped}":
+		path := r.request.Header.Get("Caddy-Rewrite-Original-URI")
```
tobya: This Header field name is used in fastcgi.go, rewrite.go and now replacer.go . I tried creating it as an exported constant but ran into import cycle issues.

Should this be a const?  What file should the const be declared in?

tobya:
I think this works fine.

@mholt I know in the original issue #1185 you suggested that may the necessity for the `"Caddy-Rewrite-Original-URI"` header field might be removed by the use of contexts.  

However I still feel this fix is needed as it establishes the pattern going forward that the `{path}' placeholder will always output the requested URI which is important.

caddyhttp/httpserver/replacer.go
mholt: For a value that is shared between middlewares, probably should be in the `httpserver` package, which is this one. I am still wondering if context would be a better way to do this...

mholt:
@tobya Merging now since there's a release tomorrow, but feel free to come back and iterate on this more if you want to, little changes welcome. :) e.g. using a const, or changing to using context.Context instead of header (if we decide to do that), etc. Thanks!

Issue #1297, opened 2016-12-15
Use tls with explicit certificates and HTTP redirects


xrstf:
### 1. What version of Caddy are you running (`caddy -version`)?

Caddy 0.9.3

### 2. What are you trying to do?

Run a frontend proxy that forces HTTPS and uses explicitely configured certificates. The manual redirect works just fine as long as I don't enable the `tls` section in my Caddyfile. Putting `redir` before or after the `tls` has no effect. I _can_ just have two server blocks (one for http, one for https), but this would require me to duplicate a few things.

### 3. What is your entire Caddyfile?
```text
http://0.0.0.0, https://0.0.0.0 {
  root /var/www

  # redirect to HTTPS
  redir 301 {
    if {scheme} not https
    / https://{host}{uri}
  }

  tls my.crt.pem my.key.pem

  proxy / myapp-container:3002 {
    transparent
  }

  log stdout

  errors {
    log stdout
  }
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

    bash% sudo caddy -conf Caddyfile.proxy
    Activating privacy features... done.
    http://0.0.0.0
    https://0.0.0.0
    WARNING: File descriptor limit 1024 is too low for production servers. At least 8192 is recommended. Fix with "ulimit -n 8192".

### 5. What did you expect to see?

    bash% curl -I http://localhost/foo
    HTTP/1.1 301 Moved Permanently
    Location: https://localhost/foo
    Server: Caddy
    Date: Thu, 15 Dec 2016 12:47:28 GMT
    Content-Type: text/plain; charset=utf-8

### 6. What did you see instead (give full error messages and/or log)?

    bash% curl -I http://localhost/foo
    HTTP/1.1 502 Bad Gateway
    Content-Type: text/plain; charset=utf-8
    Server: Caddy
    X-Content-Type-Options: nosniff
    Date: Thu, 15 Dec 2016 12:51:00 GMT
    Content-Length: 16

The 502 is expected, as this Caddy is not running within our Docker environment and there is no `myapp-container` hostname reachable. That doesn't matter though, it's only important to note that it's not the expected 301 redirect.

### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

Take the Caddyfile, run Caddy and perform any HTTP request against it.

tobya:
Maybe I'm missing something here, but caddy automatically serves on https you do not need a redirect.

    0.0.0.0 {
      root  /var/www

      etc....
     }

[...]

PR #1365, opened 2017-01-21
redirect: determine the FromScheme at runtime (#1297)

tw4452852:
`cfg.TLS.Enabled` may be changed after plugin setup ([here](https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/plugin.go#L132)), but redirect plugin doesn't catch it.

### DIFF
--- a/caddyhttp/redirect/setup.go
+++ b/caddyhttp/redirect/setup.go
@@ -34,10 +34,11 @@ func redirParse(c *caddy.Controller) ([]Rule, error) {
 	cfg := httpserver.GetConfig(c)
 
 	initRule := func(rule *Rule, defaultCode string, args []string) error {
-		if cfg.TLS.Enabled {
-			rule.FromScheme = "https"
-		} else {
-			rule.FromScheme = "http"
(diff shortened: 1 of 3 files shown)


caddyhttp/redirect/setup.go
```diff
@@ -34,10 +34,11 @@ func redirParse(c *caddy.Controller) ([]Rule, error) {
...
 	initRule := func(rule *Rule, defaultCode string, args []string) error {
-		if cfg.TLS.Enabled {
-			rule.FromScheme = "https"
-		} else {
-			rule.FromScheme = "http"
+		rule.FromScheme = func() string {
+			if cfg.TLS.Enabled {
+				return "https"
+			}
+			return "http"
```
mholt: Is this thread-safe?

(If not, and if the solution is to introduce locking, let's reconsider how we do the scheme changing instead...)

tw4452852: `cfg.TLS.Enabled` is setup at the server's initialization. Unless we changed it on the fly later, it is thread-safe IMO.

PR #1368, opened 2017-01-23
New 'timeouts' directive to configure timeouts; default timeouts enabled

### LINKED ISSUE #1311: Implement configurable ReadTimeout, WriteTimeout, and IdleTimeout
Some recent events have reminded me that we should finally implement [this TODO](https://github.com/mholt/caddy/blob/d5fe4928f284b5d10c0bcebaaa2b27d641e058c9/caddyhttp/httpserver/server.go#L46-L49) in Caddy: a way to set WriteTimeout, ReadTimeout, and (coming in Go 1.8) IdleTimeout.

Required reading:

- https://blog.gopheracademy.com/advent-2016/exposing-go-on-the-internet/
- https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

I figure this will be either one or three new directives for the Caddyfile.

Since these timeouts [can't be set on a per-request basis](https://github.com/golang/go/issues/16100), we have to reduce any repeated invocations of these directives across sites on a listener to a single value -- probably the lowest value among them. This is not too bad. We can print a warning if that's the case.

While we're at it, we should also make the max header size configurable too.

mholt:
This PR enables default timeout values and makes timeouts configurable with the Caddyfile. This is important for mitigating slowloris attacks (a type of DoS attack) which will now be mostly prevented right out of the box. I'd been meaning to do this, since it's not hard, but decided it was finally time to get around to it.

Closes #1311 and fixes this forum issue: https://forum.caddyserver.com/t/non-tls-vulnerable-to-slowloris-attack/1322?u=matt

I suspect some users who are doing long-polling or have other types of extenuating proxy/websocket scenarios would need to raise or disable these timeouts. However, I tested a simple websockets web page (just an echo server using the websocket directive) and even with the default timeouts, did not experience any interruption in service. Standard HTTP proxying, if the upstream request is slow, may be different and require higher or no timeouts. Depends on your circumstances.

Default timeouts:

- Read: 10s
- ReadHeader: 10s
- Write: 20s
- Idle: 2m

Caddyfile usage:

```
# set all timeouts to same value
timeouts 30s

# disable all timeouts
timeouts none

# configure each timeout
timeouts {
    read   10s
    header 15s
    write  20s
    idle   2m
}

# disable the write timeout
timeouts {
    write none
}
```

Note that idle and header timeouts won't be available until Go 1.8, but this PR is Go 1.8-ready. Also, if an Idle timeout is disabled but a Read timeout is set, the Go standard library will use the Read timeout as the Idle timeout.

Since timeouts are set for an entire server, a group of sites will have their timeout settings reduced to the minimum values (including 0 or "none") across all sites that share that server.

### DIFF
--- a/caddyhttp/caddyhttp.go
+++ b/caddyhttp/caddyhttp.go
@@ -26,6 +26,7 @@ import (
 	_ "github.com/mholt/caddy/caddyhttp/root"
 	_ "github.com/mholt/caddy/caddyhttp/status"
 	_ "github.com/mholt/caddy/caddyhttp/templates"
+	_ "github.com/mholt/caddy/caddyhttp/timeouts"
 	_ "github.com/mholt/caddy/caddyhttp/websocket"
 	_ "github.com/mholt/caddy/startupshutdown"
 )
--- a/caddyhttp/caddyhttp_test.go
+++ b/caddyhttp/caddyhttp_test.go
@@ -11,7 +11,7 @@ import (
 // ensure that the standard plugins are in fact plugged in
 // and registered properly; this is a quick/naive way to do it.
 func TestStandardPlugins(t *testing.T) {
-	numStandardPlugins := 28 // importing caddyhttp plugs in this many plugins
+	numStandardPlugins := 29 // importing caddyhttp plugs in this many plugins
 	s := caddy.DescribePlugins()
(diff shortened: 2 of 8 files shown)


mholt:
Will be merging this by tomorrow in time for the new release.

PR #1376, opened 2017-01-25
Implement curve X25519 (Golang 1.8)

elcore:
I promised to implement X25519 😄 (even though I have no time)

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -311,6 +311,11 @@ func MakeTLSConfig(configs []*Config) (*tls.Config, error) {
 		config.CipherSuites = append([]uint16{tls.TLS_FALLBACK_SCSV}, config.CipherSuites...)
 	}
 
+	// Default curves
+	if len(config.CurvePreferences) == 0 {
+		config.CurvePreferences = defaultCurves
+	}
+
 	// Set up client authentication if enabled
 	if config.ClientAuth != tls.NoClientCert {
 		pool := x509.NewCertPool()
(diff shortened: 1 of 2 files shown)


lhecker:
@mholt @elcore What about the `tls.Config.CurvePreferences`? [Cloudflare strongly recommends](https://blog.cloudflare.com/exposing-go-on-the-internet/) setting that value and I believe this PR might be a good fit for doing so. What do you think?

elcore:
@lhecker 

> What about the tls.Config.CurvePreferences? Cloudflare strongly recommends setting that value and I believe this PR might be a good fit for doing so. What do you think?

We are using the default configuration until the user decided to modify the curves.

Default:

`tls.Config.CurvePreferences -> empty -> default curves`

Custom Configuration:

`tls.Config.CurvePreferences -> value -> []tls.CurveID{value}`

https://github.com/mholt/caddy/blob/master/caddytls/config.go#L274-L280

https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L486-L488

lhecker:
Yeah I'm aware of that - I didn't even ask what we currently use...
What I was saying is that the Go defaults are not safe for public internet services. It's about [this](https://github.com/golang/go/tree/release-branch.go1.8/src/crypto/elliptic) folder not having optimized assembly code for `CurveP384` and `CurveP521`, which causes them to indeed take up to about **one second** of CPU time when using them.
Since `caddy` is surely trying to be a "serious" project I figured that we should also be serious and thus wary about possible implementation details inside the Go code.

elcore:
@lhecker 

Feel free to implement it, I am currently too sick to do it. I have 39°C fever!!

lhecker:
@elcore Sure! Get well soon!

elcore:
@lhecker thank you

Issue #1362, opened 2017-01-18
Encoded commas (%2C) make proxy omit base directory


mbirth:
### 1. What version of Caddy are you running (`caddy -version`)?

`0.9.4` (also happened with `0.9.1`)

### 2. What are you trying to do?

Downloading an eBook from my local [COPS](https://github.com/seblucas/cops) instance. The (generated from COPS) URL is:

`https://cops.myserver.com/download/2731/abc%2Cdef.mobi`

### 3. What is your entire Caddyfile?
```text
:80 {
    proxy / http://localhost:81/
    log stdout
    errors stderr
}

cops.myserver.com {
    proxy / http://otherserver:80/cops/ {
        header_upstream Host cops.myserver.com
        header_upstream Scheme https
    }
    log stdout
    errors stderr
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

`/usr/sbin/caddy -log stdout -agree=true -email="my@emailaddr.es" -conf="/etc/caddy/Caddyfile" -root=/var/tmp`

Run from systemd as user `www-data`. Also tested it by running as `root` from command line. Same effect.

### 5. What did you expect to see?

A request to `/cops/download/2731/abc%2Cdef.mobi` on the upstream server.

### 6. What did you see instead (give full error messages and/or log)?

A request to `/download/2731/abc%2Cdef.mobi` on the upstream server. Which returns a 404 page.

Caddy shows:

    Jan 18 16:05:37 myserver caddy[15171]: 145.243.xxx.xxx - [18/Jan/2017:16:05:37 +0100] "GET /download/2731/abc%2Cdef.mobi HTTP/2.0" 404 2968

Upstream Apache (`otherserver`) shows:

    127.0.0.1 - - [18/Jan/2017:16:05:37 +0100] "GET /download/2731/abc%2Cdef.mobi HTTP/1.1" 404 2968 "-" "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.87 Safari/537.36"

### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?

Setup Caddy as a proxy as above. Watch the requests on the upstream server. Experiment with URLs containing `%2C`.

**If I omit the `%2C` from the URL, Caddy queries the correct URL on the upstream server.**

[...]

PR #1378, opened 2017-01-26
proxy: handle encoded path in URL

tw4452852:
fix issue #1362

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -18,7 +18,6 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"path"
 	"strings"
 	"sync"
 	"time"
@@ -89,6 +88,18 @@ func socketDial(hostName string) func(network, addr string) (conn net.Conn, err 
 	}
 }
 
+func singleJoiningSlash(a, b string) string {
+	aslash := strings.HasSuffix(a, "/")
+	bslash := strings.HasPrefix(b, "/")
+	switch {
+	case aslash && bslash:
+		return a + b[1:]
+	case !aslash && !bslash && b != "":
+		return a + "/" + b
+	}
+	return a + b
+}
+
 // NewSingleHostReverseProxy returns a new ReverseProxy that rewrites
 // URLs to the scheme, host, and base path provided in target. If the
 // target's path is "/base" and the incoming request was for "/dir",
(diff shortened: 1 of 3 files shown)


caddyhttp/proxy/reverseproxy.go
```diff
@@ -89,6 +88,18 @@ func socketDial(hostName string) func(network, addr string) (conn net.Conn, err
...
+	aslash := strings.HasSuffix(a, "/")
+	bslash := strings.HasPrefix(b, "/")
+	switch {
+	case aslash && bslash:
+		return a + b[1:]
+	case !aslash && !bslash && b != "":
+		return a + "/" + b
+	}
+	return a + b
+}
```
mholt: Is this different from path.Join?

mholt:
Thanks for the PR, Tw! This seems like a complicated change for what looks like a simple problem. What about the encoded comma is causing the comma to drop path components?

caddyhttp/proxy/reverseproxy.go
tw4452852: `path.Join` will also strip the middle slash: `/abc/` + `/de` = `abc/de` instead of `abc//de`. This case will often occur in the encoded URL and it is the way that [standard lib uses](https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L95).

tw4452852:
I think [doc](https://github.com/golang/go/blob/master/src/net/url/url.go#L704-L723) will explain the everything.

caddyhttp/proxy/reverseproxy.go
```diff
@@ -119,12 +130,38 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
...
+				a = target.Path
+			}
+
+			if req.URL.RawPath != "" {
+				b = req.URL.RawPath
+			} else {
+				b = req.URL.Path
+			}
+
+			req.URL.RawPath = singleJoiningSlash(a, b)
```
mholt: I think I'm OK with this logic, but could you comment the relevant parts, like why the `if` statements are the way they are, or what the point of these `if` statements is? Maybe this chunk belongs in a separate function.

tw4452852:
@mholt Got it. I have revised it a bit. Wish it looks promising now.

mholt:
Thanks Tw! Looks great. :+1:

Issue #958, opened 2016-07-23
Reloading (USR1) with QUIC produces error


mholt:
When serving with QUIC enabled, reloading Caddy (using SIGUSR1) results in this error:

```
2016/07/23 16:08:22 [INFO] Reloading
2016/07/23 16:08:23 [ERROR] listening for QUIC connections: listen udp :443: bind: address already in use
2016/07/23 16:08:23 [ERROR] listening for QUIC connections: listen udp :80: bind: address already in use
2016/07/23 16:08:28 [INFO] Reloading complete
```

The site still runs fine and I think the reload even completes successfully, but the site will no longer serve QUIC. Caddy 0.9 can handle UDP listeners now; @lucas-clemente I wonder if you could help implement [UDPServer](https://godoc.org/github.com/mholt/caddy#UDPServer) with me when you have a chance? (This is not a high-priority issue but should probably be fixed at some point.)

lucas-clemente:
Interesting, I've never tried reloading :)

I will take a look at this and the UDPServer some time this week and let you know my thoughts.

mholt:
I know UDP doesn't really have the concept of "graceful" reloads (or the definition is fuzzy) but what Caddy can still do is hand the UDP server a PacketConn after a reload so it doesn't need to make a new one (and fail, because the old one is still bound).

Essentially we just need to make [this ListenPacket method](https://github.com/mholt/caddy/blob/1bbad72ff162bdd92960fb8d69ede9e75250a70f/caddyhttp/httpserver/server.go#L150) return [this *net.UDPConn](https://github.com/lucas-clemente/quic-go/blob/9ef936c94153b1308d346fca54533917346696bc/server.go#L75-L78).

Then [during the restart](https://github.com/mholt/caddy/blob/1bbad72ff162bdd92960fb8d69ede9e75250a70f/caddy.go#L575-L635), Caddy will hand over that UDPConn to the new server instance that's created by calling `Serve()` directly, passing it in. Which means that [the ServePacket method in Caddy's HTTP server](https://github.com/mholt/caddy/blob/1bbad72ff162bdd92960fb8d69ede9e75250a70f/caddyhttp/httpserver/server.go#L193) needs a couple lines of code to call Serve on your QUIC/UDP server. I think.

Anyway, I hope that's helpful. If we did our part right with Caddy I don't think this will be too difficult. :smile: Let me know!

miekg:
Cool :-)
(told you so :) )

[...]

PR #1381, opened 2017-01-26
httpserver: support QUIC reload

tw4452852:
fix issue #958

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -145,8 +145,17 @@ func (s *Server) Listen() (net.Listener, error) {
 	return ln.(*net.TCPListener), nil
 }
 
-// ListenPacket is a noop to implement the Server interface.
-func (s *Server) ListenPacket() (net.PacketConn, error) { return nil, nil }
+// ListenPacket creates udp connection for QUIC if it is enabled,
+func (s *Server) ListenPacket() (net.PacketConn, error) {
+	if QUIC {
+		udpAddr, err := net.ResolveUDPAddr("udp", s.Server.Addr)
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/server.go
```diff
@@ -146,7 +146,16 @@ func (s *Server) Listen() (net.Listener, error) {
 }
 
 // ListenPacket is a noop to implement the Server interface.
```
mholt: No longer a no-op. ;) Need to update the godoc comment.

caddyhttp/httpserver/server.go
```diff
@@ -189,7 +189,12 @@ func (s *Server) Serve(ln net.Listener) error {
 }
 
 // ServePacket is a noop to implement the Server interface.
-func (s *Server) ServePacket(pc net.PacketConn) error { return nil }
+func (s *Server) ServePacket(pc net.PacketConn) error {
+	if QUIC {
+		return s.quicServer.Serve(pc.(*net.UDPConn))
```
mholt: Could we wrap this error, something like what we used to have:

```go
return fmt.Errorf("listening for QUIC connections: %v", err)
```

caddyhttp/httpserver/server.go
tw4452852: Done.

caddyhttp/httpserver/server.go
tw4452852: Done

tw4452852:
Yes, I try to reload the server with `SIGUSR1`, it works well.

mholt:
Thank you! This is awesome.

PR #1389, opened 2017-01-30
tls: Per-site TLS configs using GetClientConfig, including http2 switch

wendigo:
Hi,

This PR introduced per vhost tls.Config using GetConfigForClient :)

I've tested it an now it's possible to set tls.Config using ```tls``` directive for every subdomain on it's own (i.e. customize curves, tls version, cyphers etc)

As per requested I've also added: 
```
tls {
  http2 off
}
```

so HTTP/2 can be turned of completely for a subdomain.

Addresses: https://github.com/mholt/caddy/issues/1375, https://github.com/mholt/caddy/issues/189


mholt:
Awesome, I just got back in town and will take a look at this ASAP!

caddytls/handshake.go
```diff
@@ -58,11 +58,30 @@ func (cg configGroup) getConfig(name string) *Config {
...
 
+// GetCertificate gets a certificate to satisfy clientHello. In getting
+// the certificate, it abides the rules and settings defined in the
+// Config that matches clientHello.ServerName. It first checks the in-
+// memory cache, then, if the config enables "OnDemand", it accesses
+// disk, then accesses the network if it must obtain a new certificate
+// via ACME.
+//
+// This method is safe for use as a tls.Config.GetCertificate callback.
+func (cg ConfigGroup) GetConfigForClient(clientHello *tls.ClientHelloInfo) (*tls.Config, error) {
```
mholt: Amazingly, I'm having a hard time finding something other than whitespace to nit-pick about here, but I finally found something. :) The godoc for this function is not relevant...

mholt:
> I think that it will be good idea to move certificate obtaining from handshake.go (configGroup) to caddytls.Config (as generate will be obtained per config).

You might be right. However, we might still want to keep those methods in handshake.go since that is a significant part of the handshake...

mholt:
@wendigo I made a merge conflict for you -- you're welcome. 😅 I just merged @elcore's X25519-related PR, so just make sure this branch sets curve preferences.

caddytls/handshake.go
mholt: @wendigo Ready to merge as soon as this godoc is revised! :tada:

wendigo: Done :)

caddytls/setup.go
```diff
@@ -164,6 +164,22 @@ func setupTLS(c *caddy.Controller) error {
...
+				args := c.RemainingArgs()
+				if len(args) != 1 {
+					return c.ArgErr()
+				}
+
+				switch args[0] {
+				case "off":
+					config.DisableHTTP2 = true
+				case "on":
+					config.DisableHTTP2 = false
```
mholt: Is there a case where HTTP/2 would need to be turned on, since it's enabled by default? (I would rather not have people thinking they need to explicitly turn on HTTP/2.) If not, let's remove the "on" option.

wendigo: No, there isn't. That's exactly why I've introduced DisableHTTP2 (and used golang default boolean value 'false')

wendigo: Removed "on"

wenerme:
Any schedule for next release ?

mholt:
@wenerme The release window is in motion right now as I am doing a lot of work behind the scenes for a big release in a few weeks or a couple months.

FraaJad:
Did this ever make it in?  I get `Error during parsing: Unknown subdirective 'http2'`

mholt:
Yes, you'll need to use the `alpn` subdirective. See the docs: https://caddyserver.com/docs/tls

FraaJad:
Thanks!  My head is spinning from the speed of the response.

ronschaeffer:
> Yes, you'll need to use the `alpn` subdirective. See the docs: https://caddyserver.com/docs/tls

Not very experienced with Caddy...

I tried disabling http/2 in my Caddyfile, but I can’t work out what the correct format should be. I've tried various locations/formats/bracketing of `http2 off` using the (sub)directives `alpn` and `tls`. Everything I’ve tried results in an unknown directive or unknown subdirective error. Could you please give an example of a complete working configuration? Thanks very much.

Here is the beginning of my Caddyfile.

```
xxxxxx.duckdns.org {
    header / {
    Strict-Transport-Security "max-age=31536000; includeSubdomains"
    X-XSS-Protection "1; mode=block"
    X-Frame-Options "SAMEORIGIN"
    Referrer-Policy "same-origin"
    }
    proxy / localhost:8123 {
        websocket
        transparent
    }
}
```

francislavoie:
> alpn is a list of space-separated protocols to use for Application-Layer Protocol Negotiation (ALPN). For HTTPS servers, HTTP versions can be enabled/disabled with this setting. Default is `h2 http/1.1`.

So you need to set `alpn http/1.1` to exclude `h2`.

Next time, please ask on https://caddy.community for usage questions, old issues like this aren't the best place to ask.

ronschaeffer:
Worked. Thanks for the quick reply and noted re the community forum.

PR #1392, opened 2017-02-01
Add request placeholder support for querying request cookies.

augustoroman:
This adds the ability to query the request cookies for placeholders
using the syntax "@cookiename".

For example, this would allow rewriting based on a cookie:
```
  rewrite {
    if {:version} is 'dev'
    to /dev/index.html
  }
```

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -199,7 +199,7 @@ func (r *replacer) getSubstitution(key string) string {
 		}
 	}
 	// next check for cookies
-	if key[1] == '@' {
+	if key[1] == ':' {
 		name := key[2 : len(key)-1]
 		if cookie, err := r.request.Cookie(name); err == nil {
 			return cookie.Value
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -199,7 +199,7 @@ func (r *replacer) getSubstitution(key string) string {
 		}
 	}
 	// next check for cookies
-	if key[1] == ':' {
+	if key[1] == '~' {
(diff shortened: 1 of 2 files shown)


mholt:
@augustoroman Thanks for the PR! Do you mean this instead?

```plain
  rewrite {
    if {@version} is 'dev'
    to /dev/index.html
  }
```

That's cool. I'm not sure the "@" character is the right one, though, since @ these days feels like a mention. What about `^` or `~`?

augustoroman:
@mholt Ooops, yes that's what I meant.  I fixed the PR description.

I don't have any strong preference on the particular symbol, I just picked @ since it's round like a cookie. :-) 
We could use 🍪 ([ref](http://www.fileformat.info/info/unicode/char/1f36a/index.htm))!

In rough order of preference, I think I like:
```
{@cookie}
{🍪cookie}
{:cookie}
{~cookie}
{^cookie}
{*cookie}
{©cookie}
```

Let me know what you preference is and I'll update the PR.

tobya:
Although I like the idea of {🍪cookie} it may be somewhat impractical for the variation in text editors still used today.  

My vote would be with {:cookie} if this is going to be implemented.

mholt:
The colon looks like chips in a cookie... ha. But... sorry, I'm not sure I like the `:` because it looks too much like a key-value pair that is missing the key.

I know it's easy to draw correlations to anything else (e.g. "{^name}" looks like a regular expression) but maybe we should reserve the colon for something else if we need to.

I'll think about it? And get back to you. :smile:

augustoroman:
Sure, reserving the colon for something like key-value pairs later on is fine, and for some reason I don't like the `^` -- maybe the regular expression similarity is nagging at me.  Let's do `~`.

PR #1404, opened 2017-02-06
Create only one log roller per file across whole process (fixes #1363)

### LINKED ISSUE #1363: Log rotate issues
Current release

Access logs rotation

```
log / access.log "{remote} {when} {method} {uri} {proto} {status} {size} {>Referer} {>User-Agent} {latency}" {
        rotate {
                size 100 # Rotate after 100 MB
                age  30  # Keep log files for 14 days
                keep 10  # Keep at most 10 log files
        }
}

```

Log file created, trimmed at proper size and new log files created. however logfiles of greater size was created, then a small file was created. If this bug triggers in production, then randomly previous access files could get deleted as keep limit would touch

wrong sized file, see ls -lh access*
-rw-r--r--  1  root  wheel   100M Jan 18 06:56 access-2017-01-17T09-01-49.164.log
-rw-r--r--  1 root  wheel   106M Jan 18 17:58 access-2017-01-18T17-39-41.106.log
-rw-r--r--  1 root  wheel   344B Jan 18 17:40 access-2017-01-18T17-45-42.329.log
-rw-r--r--  1 root  wheel   70K Jan 18 17:41 access.log

Firing lot of requests

mholt:


caddyhttp/httpserver/roller.go
```diff
@@ -19,14 +20,30 @@ type LogRoller struct {
 }
 
 // GetLogWriter returns an io.Writer that writes to a rolling logger.
+// This should be called only from the main goroutine (like during
```
tw4452852: How about just making it thread-safe?

mholt: I guess I could, I just didn't bother to since this is only called during the setup which happens in the main goroutine. Maybe if the need arises we can make it thread-safe?

tw4452852: Make sense.

PR #1409, opened 2017-02-07
return 404 for "not a directory" errors

### LINKED ISSUE #1408: caddy returns 503 if path contains a file
master branch

GET request on `existingfile/something`

```text
:2015
```

`caddy`

Expected a 404 response

Got a 503  response

create `foo.html`
run `caddy`
run `curl localhost:2015/foo.html/somepath`

mastercactapus:
Fixes #1408

### DIFF
--- a/caddyhttp/staticfiles/fileserver.go
+++ b/caddyhttp/staticfiles/fileserver.go
@@ -56,6 +56,9 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
 
 	f, err := fs.Root.Open(name)
 	if err != nil {
+		// TODO: remove when http.Dir handles this
+		// Go issue #18984
+		err = mapFSRootOpenErr(err)
 		if os.IsNotExist(err) {
 			return http.StatusNotFound, nil
 		} else if os.IsPermission(err) {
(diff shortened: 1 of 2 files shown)


mholt:
Thanks! I'm kind of watching https://github.com/golang/go/issues/18974 and now https://github.com/golang/go/issues/18984 to see what happens and when/if it gets fixed in the standard lib. If it does get fixed there, we could remove this hack later, which means we'd want to accompany it with a TODO of some sort.

caddyhttp/staticfiles/fileserver.go
```diff
@@ -40,6 +40,15 @@ func (fs FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, err
...
 
+func isNotDirErr(err error) bool {
+	perr, ok := err.(*os.PathError)
+	if !ok {
+		return false
+	}
+
+	return perr.Err.Error() == "not a directory"
+}
+
```
mholt: Can we move this lower in the file to keep the serveFile and ServeHTTP methods close together?

mholt: Also a godoc on this would be useful.

mastercactapus: Yeah, I'm actually going to revisit this and update the PR with the fix I'm using/submitting to golang. I'll add the TODO and comments as well -- hopefully sometime today.

mastercactapus:
@mholt I think this is ready to be looked at again when you get a chance. The fix has now been merged into golang/go. I've used that implementation here, and added TODO comments to remove it when the fix is released.

Let me know if you want me to revise anything!

mholt:
Thanks @mastercactapus  - I see they merged a fix into Go. Am I correct in understanding that this patch is temporary, then, until they release a Go version that has your patch?

mastercactapus:
@mholt That is correct. I also tested without this patch against Go master to make sure it was resolved (and how I found an additional case). This is the same code that got merged into Go, adapted of course.

mholt:
@mastercactapus Cool. When that patch is released, will you update Caddy so this patch won't be redundant? If so, I'll merge this.

mastercactapus:
@mholt definitely, I can own it.

mholt:
@mastercactapus Mind if I invite you to be a collaborator? I can't remember if I've done this for you yet, but it doesn't look like I have. This way you could contribute in a branch instead of a fork, and help with reviewing pull requests and issues when you are able.

mastercactapus:
@mholt I don't mind, I'd be happy to

PR #1410, opened 2017-02-07
Replace magic number 308 with http.StatusPermanentRedirect

JRaspass:


mholt:
This must be from a newer version of Go. Thanks!

Issue #1396, opened 2017-02-02
Allow placeholders in log filenames


skpy:
I recognize that the documentation does not claim to support placeholders in the filename portion of logs, so please consider this a feature request more than a bug report!

### 1. What version of Caddy are you running (`caddy -version`)?
Caddy 0.9.5

### 2. What are you trying to do?
I'd like to have multiple vhosts (in a WordPress multi-site setup) served by a single Caddy configuration block, but have each site's activities logged to vhost-specific log files.

### 3. What is your entire Caddyfile?
```
# for local testing purposes only
*:80
gzip
tls off
rewrite {
  if {path} not_match ^\/wp-admin
  to {path} {path}/ /index.php?_url={uri}
}
fastcgi / 127.0.0.1:9000 php
log /var/log/caddy/{hostonly}.log {
  rotate {
    size 100 # Rotate after 100 MB
    age 7 # Keep rotated log files for 7 days
    keep 10 # Keep at most 10 log files
  }
}
errors {
  log /var/log/caddy/{hostonly}-e.log {
    size 50 # Rotate after 50 MB
    age 7 # Keep rotated files for 7 days
    keep 5 # Keep at most 5 log files
  }
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
I copied the `init/linux-systemd/caddy.service` file to `/etc/systemd/system`, and ran `sudo systemctl start caddy.service` on an Ubuntu 16.04 VM running under VirtualBox.

### 5. What did you expect to see?
I expected to see `/var/log/caddy` populated with files that matched the names I might use to access the sites.

### 6. What did you see instead (give full error messages and/or log)?
I see files named `/var/log/caddy/{hostonly}.log` and `/var/log/caddy/{hostonly}-e.log`

### 7. How can someone who is starting from scratch reproduce this behavior as minimally as possible?
Ensure that the site running Caddy is accessible via multiple host names.  If DNS is not set up for this, add multiple entries to `/etc/hosts` for the IP address at which Caddy is running.  Then try to access each of those host names.

Strictly speaking even that's not necessary.  Just try to use a placeholder in the filename portion of a `log` directive and observe that the placeholder is not replaced.

tobya:
I have the start of a PR for this issue  with commit https://github.com/mholt/caddy/commit/4efdf66abdb4fb01b44d9cbfeb387570e3ffce61

@mholt  There are some comments inline if you have time to take a look.

Essentially I call replacer on the path for the log file.  

[...]

PR #1414, opened 2017-02-08
Allow replacement params in log filename

tobya:
I have the start of a PR for this issue #1396 

@mholt There are some comments inline if you have time to take a look.

Essentially I call replacer on the path for the log file.

If the logfile returned is differnt to the currently open one, I close it and open the new log file? I have questions around

- Am I closing the existing logfile correctly?
- Would a better approach be to leave a file open for each different log file opened rather than closing and reopening?
- Is this desired behaviour are we happy that this Issue is to be implemented?
- Is #1404 relevant to this at all?

This is not ready to merge yet, I need to add tests and tidy up some code but need some feedback on direction.

### DIFF
--- a/caddyhttp/log/log.go
+++ b/caddyhttp/log/log.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+
 	"sync"
 
 	"github.com/mholt/caddy"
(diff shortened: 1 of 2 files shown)


tobya:
This needs to be merged with #1246 before it can work.  I will merge it and repush

mholt:
Hey, sorry, I will get to this. Just working through a backlog.

tobya:
Actually my fix here needs to be rewritten due to #1246 - I will try to get around to it or someone else can take my branch and modify.

mholt:
Sounds good. Sorry about that!

mholt:
Actually... I dunno, is this even thread-safe? It would need testing as well... I'm not sure opening and closing logs per-request is a good idea either.

tobya:
I was concerned about that myself.  However it seems that the new logger.go  code ensures that a single logger per file is kept open.  If the code opens a log file if not open and then on each log request simply uses  the correct logger instance it might be valid (eg no need to close and reopen files).

Also the use of replacer would need to be restricted (eg {host} only).  

However while a nice to have I don't think this is an essential feature and certainly we can park it for the time being.  I might come back to it in a while if I can find an elegant solution.

mholt:
@tobya Thanks, I also just thought of the security implications of this which I commented in the other issue. Sorry about that. But I think you're right, it's a lesser priority right now than other features/bugs that need to be worked out.

PR #1425, opened 2017-02-11
Register cgi plugin

jung-kurt:
CGI middleware for Caddy. Directive syntax has been flattened and problems raised in issues forum have been addressed.


mholt:
Thank you @jung-kurt!

PR #1426, opened 2017-02-11
Store name of authenticated user in basicauth

jung-kurt:
The fastcgi and (pending) cgi middleware are currently unable to set the REMOTE_USER environment variable of users authenticated with the basicauth plugin. This PR sets the HTTP_AUTHORIZED key in the request header to the name of the authenticated user in order to do that. The password is not retained.


mholt:
Thanks for the PR, @jung-kurt! This is not a standard field, correct? Do you think this will be likely to overwrite an existing header field?

mholt:
Actually, how about using request.Context instead?

And can you add a test case? :innocent:

jung-kurt:
> Actually, how about using request.Context instead?

Perfect! That makes much better sense than co-opting a header field.

> And can you add a test case? 😇

Will do.

caddyhttp/basicauth/basicauth.go
```diff
@@ -60,6 +61,10 @@ func (a BasicAuth) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 
 			// remove credentials from request to avoid leaking upstream
 			r.Header.Del("Authorization")
+
+			// let upstream middleware (e.g. fastcgi and cgi) know about authenticated
+			// user; this replaces the request with a wrapped instance
+			r = r.WithContext(context.WithValue(r.Context(), "remote_user", username))
```
mholt: Instead of using a primitive string type as a context key (which is asking for collisions, though I admit, less likely in Caddy-specific packages), we should wrap this in `httpserver.CtxKey`. It doesn't exist yet but it will when #1430 is merged! I should have done that change separately, sorry. Also see #1331 which may be blocked on the same thing.

caddyhttp/basicauth/basicauth_test.go
```diff
@@ -14,19 +14,31 @@ import (
 )
 
 func TestBasicAuth(t *testing.T) {
+	var i int
+	// This handler is registered for tests in which the only authorized user is
+	// "okuser"
+	upstreamHandler := func(w http.ResponseWriter, r *http.Request) (int, error) {
+		remoteUser, _ := r.Context().Value("remote_user").(string)
```
mholt: The `, _` isn't necessary for type assertions, but why ignore the bool/error? Wouldn't it panic then? I guess it's a test case so it's not the end of the world, but perhaps better to check the error and t.Fatal if there's a problem.

Edit: Nevermind, it doesn't panic. Clever.

jung-kurt: Admittedly, it could be _too_ clever. Maybe something like [this](https://github.com/mholt/caddy/blob/30de2e47590bdd63219b259c87b281e0f5ec00d1/caddyhttp/httpserver/context.go#L327-L332) is clearer for maintenance. Does it make sense to have some helper functions for setting and getting context variables?

mholt: The more I think about it, the net result is the same. It's not technically an error value, it's a boolean telling you whether the assertion was successful. I like the brevity of your approach. I just forgot that type assertions don't panic if you assign to two variables, even if one is discarded. This is fine.

caddyhttp/basicauth/basicauth.go
mholt: @jung-kurt Okay, I've merged the other branch, so if you want to wrap your string key in httpserver.CtxKey, then I think we're good to go!

jung-kurt: Thanks, @mholt!

mholt:
Thank you, @jung-kurt ! If you'd like, I'll make you a collaborator so you can submit changes in branches instead of forks, and help review pull requests and help with issues.

jung-kurt:
Thanks, @mholt -- I'd be honored to be a collaborator. Caddy is an impressive project -- keep up the great work!

PR #1428, opened 2017-02-13
fixing panic when root is symlink

pennywisdom:
When the root path that is set is a symlink, caddyserver will not start and panics. 
This PR is adding a check to determine if the path is a symlink first and if it is then delegate the responsibility that it is a valid path to the underlying OS, otherwise continue with the existing checks.


mholt:
@pennywisdom Why did you close this?

mholt:
@pennywisdom Nevermind -- I just saw your other PR. 😅

PR #1429, opened 2017-02-13
fixing panic when root is symlink

pennywisdom:
Re-submitting PR (tests passing running locally!)
When the root path that is set is a symlink, caddyserver will not start and panics.
This PR is adding a check to determine if the path is a symlink first and if it is then delegate the responsibility that it is a valid path to the underlying OS, otherwise continue with the existing checks.

### DIFF
--- a/caddyfile/dispenser.go
+++ b/caddyfile/dispenser.go
@@ -120,12 +120,6 @@ func (d *Dispenser) NextBlock() bool {
 	return true
 }
 
-// IncrNest adds a level of nesting to the dispenser.
-func (d *Dispenser) IncrNest() {
-	d.nesting++
-	return
-}
-
 // Val gets the text of the current token. If there is no token
 // loaded, it returns empty string.
 func (d *Dispenser) Val() string {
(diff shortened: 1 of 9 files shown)


mholt:
Thanks @pennywisdom! I am surprised that it straight-up panicked in that case... do you think you could write a test so that it preserves the fix contained in your patch?

pennywisdom:
yes @mholt  i'll get one/some created.

pennywisdom:
Not sure how to handle the issue on Windows (not server. You need to be running as Administrator to create a symlink  and if you are not then an error throw, however I'm not sure this is a good case for failing the tests, seems like a bit of an edge case, and we only need to create the symlink in the test, its not part of running Caddy. 
For this reason, the test verifies that the error is thrown when privileges need elevating and skips the test, which will be logged by the runner.

mholt:
I also really appreciate your effort in writing the test case and considering the permissions issues. I think you handled it fine. Would you like to be added as a collaborator on the project? That means you could contribute without having to fork, you could just do it in a branch, and you could help review pull requests and issues.

pennywisdom:
No probs at all, really happy to help out. Yes please do add me as a collaborator that would be great!

PR #1430, opened 2017-02-13
Detect HTTPS interception

### LINKED ISSUE #1420: Implement MITM detection described by Durumeric, et. al.
A recent paper, [The Security Impact of HTTPS Interception](https://jhalderm.com/pub/papers/interception-ndss17.pdf) published by Zakir Durumeric, Richard Barnes, Nick Sullivan, J. Alex Halderman, and others demonstrates the use of heuristics for detecting active MITM attacks/proxies on TLS connections:

> First, we show that web servers can detect interception by identifying a mismatch between the HTTP User-Agent header and TLS client behavior. We characterize the TLS handshakes of major browsers and popular interception products, which we use to build a set of heuristics to detect interception and identify the responsible product.

The tl;dr of the heuristics described in the paper is to capture the information in the ClientHello of TLS handshakes and then capture the User-Agent value from the subsequent HTTP request headers. Since the TLS configuration of the major user agents is known and is mostly predictable (see paper for caveats), the paper suggests that by comparing the User-Agent header field to the actual client inferred from the characteristics of the TLS ClientHello is a way to detect active MITM attacks/proxies.

Despite benevolent intentions of some TLS proxies, [they actually do more harm than good](https://users.encs.concordia.ca/~mmannan/publications/ssl-interception-ndss2016.pdf), and end users are not always able to know when their "trusted" proxy is helping or hurting them.

Because of the adverse effects on user privacy and the technical problems of TLS proxies of all kinds, I propose we give site owners the ability to handle clients who are being MITM'ed in the way that seems best to them.

I propose these actions should be possible:

- Close the connection entirely. This would involve a new directive or a special property in the TLS directive that would allow the site owner to specify base paths (maybe?) which are considered sensitive enough to not even send a response in the case of a MITM.
- Perform rewrites, redirects, or log when a connection is thought to be MITM'ed. This would involve a new [request placeholder](https://caddyserver.com/docs/placeholders), perhaps `{mitm}` that contains a yes/no, 1/0, or true/false. This could also be used to set headers on the request for upstream services to be aware of the attack.
- Warn the client in HTML or other kinds of responses of the attack. This would require a new [template action](https://caddyserver.com/docs/template-actions) such as `{{.IsMITM}}` (returns true/false or nonempty/empty) so that the site owner could inject custom HTML or scripts to help warn the client. (Yes, MITM proxies can of course try to detect these and strip them out, however, that is unlikely with the volume of Caddy sites at this point and would be difficult to generalize across all sites anyway.)

Possible challenges:

- Maintaining a list of major browser TLS configurations. Maybe this won't change often, but it will change.
- Mapping connection addresses to ClientHellos (`map[string]*tls.ClientHelloInfo`) may be inefficient, since the map will have to be protected with a RWMutex. We'll have to do some benchmarking to see if the impact is negligible under load, or if this feature will have to be turned on by the site owner before it can be used. Also, this requires Go 1.8.
- ~~We need to use GetConfigForClient since GetCertificate is not always called (for example, I've seen requests with `curl --insecure` that don't invoke GetCertificate).~~ We don't need this callback at all.
- Reading TLS Heartbeat support and other certain raw pieces from the ClientHello will be tricky because the ClientHelloInfo does not check for this (understandably so).

Feedback welcome. I will be starting on this as soon as I get a chance amid the other drastic changes that Caddy is undergoing right now (build server infrastructure, new website, etc).

Update: I have a PoC started. I can map requests to their ClientHelloInfos. `wrk` indicates no significant overhead in doing this. I will invite others to verify when I have something to show. Now all that's left is the ability to detect the browser from the ClientHelloInfo and expose the results of the detection to templates and configuration.

mholt:
This is a WIP branch which closes #1420. It makes it possible for Caddy to detect when it is likely that the TLS connection is being intercepted. This is possible because TLS proxies are mostly careless. It works by comparing the properties of the TLS Client Hello message to what is expected of the client indicated by the User-Agent HTTP header. If there is a mismatch, a MITM is likely.

The heuristics used here are a partial implementation of [The Security Impact of HTTPS Interception](https://jhalderm.com/pub/papers/interception-ndss17.pdf) by Durumeric, Halderman, and others.

I would like people to try this out -- you will need to compile this branch with Go 1.8.

This PR also adds:

- `{{.IsMITM}}` [template action](https://caddyserver.com/docs/template-actions) that returns true or false.
- `{mitm}` [placeholder](https://caddyserver.com/docs/placeholders) which can be used in the Caddyfile to initiate redirects or add the value to a log entry. It returns likely, unlikely, or unknown.

TODO:

- [x] Write tests
- [x] Flush out the Safari heuristics
- [x] Seek clarification on the IE/Edge heuristics as they seem to overlap with Chrome
- [x] Reading the TLS Client Hello in Accept() is currently blocking, I think. Verify if this is the case and, if it is, find a way to read the Hello without blocking other connections...
- [x] A way to close the connection right away if a MITM is suspected (will be exposed via Caddyfile directive later if this feature is useful)

An easy way of testing this is to use Firefox to load a HTTPS page, then use the Dev Tools to "Edit and Resend" the HTTP request, changing the User-Agent header to replace "Firefox" with something like "Chrome" or any other browser.

I would like to encourage anyone to test this out and submit feedback, this is a highly experimental feature but one I feel could be important. Any other collaborators that would like to participate in reviewing, please do!


caddyhttp/httpserver/mitm.go
```diff
@@ -0,0 +1,547 @@
...
+	extensionSupportedCurves   = 10 // also called "SupportedGroups"
+	extensionSupportedPoints   = 11
+	extensionHeartbeat         = 15
+
+	scsvRenegotiation = 0xff
+
+	// cipher suites missing from the crypto/tls package,
+	// in no particular order here
+	TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 = 0xcca9
+	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256   = 0xcca8
```
elcore: ChaCha20 cipher suites are not missing from the crypto/tls package.

https://tip.golang.org/pkg/crypto/tls/#pkg-constants

```
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305    uint16 = 0xcca8
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305  uint16 = 0xcca9
```

`TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 ` is not missing in the crypto/tls package.

mholt: Oh, thanks.

elcore: No problem 😄, in my spare time I will try to review this PR

tobya:
Not sure if this shows something or nothing, but i tested by editing firefox request as suggested and got expected results.  Then I connected using my phone so using the mobile network without editing anything.  Got the following results

    178.167.254.199 - [15/Feb/2017:15:48:19 +0000] 'GET  /list.php /list.php  - HTTP/2.0 200 51060' https unlikely
    178.167.254.199 - [15/Feb/2017:15:48:23 +0000] 'GET  /favicon.ico /favicon.ico  - HTTP/2.0 404 58442' https likely

Would there be a reason that the file I requested list.php is `unlikely` but the favicon.ico (which doesnt exist) would be `likely`?

mholt:
The [heuristics used in the research](https://github.com/zakird/tlsfingerprints) will be published probably no sooner than March 1, 2017, but I may merge this PR before then and then iterate on this feature later at that time.

mholt:
@tobya Interesting... can you add `{remote}:{port}` to your log format and see if they are the same? Also, since you're building this from source, if you dump the request headers (`fmt.Printf("Headers: %+v", r.Header)` might be helpful too.

tobya:
@mholt  {remote} and {port} are there (just apart)

178.167.254.199:51060   /list.php   https unlikely
178.167.254.199:58442 /favicon.ico' https likely

The ports are differnt.  

I can reproduce consistently with 3G on my mobile  on Chrome for iOS.  This is output of headers. 

    178.167.254.34:58772 - [15/Feb/2017:21:59:47 +0000] 'GET asite4.bcs.cookingisfun.ie  /list.php /list.php  - HTTP/2.0 200 ' https unlikely
    
    Headers: map[Accept-Encoding:[gzip, deflate] Cookie:[_ga=GA1.2.1285204035.1421831904; __utma=71723596.1285204035.1421831904.1478764756.1478864814.78;     __utmz=71723596.1478705289.76.15.utmcsr=jobsforcooks.com|utmccn=(referral)|utmcmd=referral|utmcct=/employers/] Accept:[text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8] User-Agent:[Mozilla/5.0 (    iPhone; CPU iPhone OS 10_0_2 like Mac OS X) AppleWebKit/602.1.50 (KHTML, like Gecko) CriOS/56.0.2924.79 Mobile/14A456 Safari/602.1] Accept-Language:[en-ie]]
    
    178.167.254.34:50453 - [15/Feb/2017:21:59:48 +0000] 'GET asite4.bcs.cookingisfun.ie  /favicon.ico /favicon.ico  - HTTP/2.0 404 ' https likely
    
    Headers: map[User-Agent:[Mozilla/5.0 (iPhone; CPU iPhone OS 10_0_2 like Mac OS X) AppleWebKit/602.1.50 (KHTML, like Gecko) CriOS/56.0.2924.79 Mobile/14A456 Safari/602.1] Accept-Encoding:[gzip, deflate, sdch]     Accept-Language:[en-US,en;q=0.8]]

As you guessed the ports are different.  This **only happens on my mobile phone and only on Chrome**, firefox for iOS and Safari for iOS dont cause this.

I have uploaded the log with headers for various calls with 3 differnt browsers  https://gist.github.com/tobya/8ec9e4b6553264b6e2022b228a58f631

tobya:
On a related / unrelated note.  i wonder should we reconsider the use of likely / unlikely.

It is quite difficult to search a log file for likely when unlikely is also there on each line.  Perhaps I could suggest

islikely / unlikely /unknown
possible / unlikely /unknown
maybe / unlikely /unknown

but I guess if you want to keep it as is, it is possible to simple search for [space]likely

mholt:
@tobya Thanks, this might be a flaw in how I'm guessing the browser from the User-Agent string. It probably thinks it is Safari and not Chrome because Chrome doesn't appear in the UA. :confused: That's annoying. What the heck is CriOS?

I don't have iOS so I can't test this. If you want to help fix it, put some print/log statements in the ServeHTTP of mitm.go and see which browser it thinks the client is based on the UA and let's see if it's getting it right, which I bet it's not. It doesn't explain why only some are false positives though... but maybe that will be clear later...

tobya:
Should I move this to a seperate issue?

From my position of ignorance it seems it may be that Chrome on iOS is using a cipher that is in your list of ciphers not supported by Chrome specifically I output ext [at this line](https://github.com/mholt/caddy/blob/mitmdetect/caddyhttp/httpserver/mitm.go#L443) and output uint16=49188

Would a fix to be to add CriOS as another specific browser and remove that cipher from the list?

mholt:
@tobya This issue is fine, we're almost done I think.

I need from you:

- The hex encoding of the TLS ClientHello (the current commit on this branch prints this out for you)
- The exact browser and iOS versions

And yes, I think that is a reasonable course of action: add CriOS as an alternative to Chrome in our conditions and then remove that cipher from the list of exclusions. If you want to try that in your build too and see if it fixes it, that would be great!

And with those two pieces of information I asked you for I'll add this to our test corpus.

wendigo:
LGTM :)

mholt:
@wendigo Thanks. Have you used the "review" feature on Github before? Just so you know, you can leave a review with inline comments and express approval or request changes that way.

PR #1434, opened 2017-02-14
Revised PR 1432

rickb777:
See https://github.com/mholt/caddy/pull/1432.

This revision uses strong ETags only. If-None-Match requests are handled correctly (by the standard library).

However the possibility of making the strong vs weak choice a configurable option hasn't been explored.


rickb777:
I still seem to have difficulty with the CLA filter. Why does it think I'm not me?

mholt:
It's because the email address on your commit isn't the one on your GitHub account, or the one you signed the CLA with. Don't worry about it, I can override that.

PR #1435, opened 2017-02-14
Revised fileserver Accept-Encoding and ETag

rickb777:
The parsing of Accept-Encoding was inadequate, not recognising word boundaries. The returned ETag was  incorrectly weak variant.

The ETag calculation is now done in base-36.

This PR addresses these issues and improves the inadequacies in the related tests.

See https://github.com/mholt/caddy/pull/1432 and https://github.com/mholt/caddy/pull/1434.


rickb777:
PRs https://github.com/mholt/caddy/pull/1432 and https://github.com/mholt/caddy/pull/1434 were rejected by the CLA filter.

elcore:
> PRs #1432 and #1434 were rejected by the CLA filter.

That's not a big issue

"Hm, yeah, the CLA tool doesn't work if your email isn't on your GitHub profile. It's fine though, I can override that check." -- Matt Holt

mholt:
Well, I responded to your comments over in #1432 but you can continue working out of this one I guess. :)

mholt:
Given my response in the other thread, is there anything else you wanted to do to this before I review it?

rickb777:
Thanks for your responses. I have revisited this, as you can see. It's ready now, if you're happy to merge it.

caddyhttp/staticfiles/fileserver.go
```diff
@@ -40,6 +40,14 @@ func (fs FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, err
 	return fs.serveFile(w, r, r.URL.Path)
 }
 
+// calculateEtag produces a strong etag by default. Prefix the result with "W/" to convert this into a weak one.
+// see https://tools.ietf.org/html/rfc7232#section-2.3
+func calculateEtag(d os.FileInfo) string {
+	t := strconv.FormatInt(d.ModTime().Unix(), 36)
+	s := strconv.FormatInt(d.Size(), 36)
+	return fmt.Sprintf(`"%s%s"`, t, s)
+}
```
mholt: This is my only nit-pick, and it is a nit-pick. :) I suspect `t + s` will be slightly faster than using the reflection features of `fmt.Sprintf`. Might be worth changing it to `return t + s`.

rickb777: Yes, well spotted, it would be (and I asked myself the same question). But the double quote marks are required (note these are enclosed in the back-ticks). It's a feature of ETags that there need to be double-quote marks. A bit unnecessary I agree, but that's what the spec requires.

rickb777: We could have
```
`"` + t + s + `"`
```
instead, perhaps. Suggestions are welcome.

mholt: I didn't see the quotes :smile:

Not that it matters I guess, but I did a quick benchmark:

```go
func BenchmarkSprintf(b *testing.B) {
	for n := 0; n < b.N; n++ {
		fmt.Sprintf(`"%s%s"`, s1, s2)
	}
}
func BenchmarkConcat(b *testing.B) {
	for n := 0; n < b.N; n++ {
		_ = `"` + s1 + s2 + `"`
	}
}
```

Results:

```
BenchmarkSprintf-8       5000000      267 ns/op
BenchmarkConcat-8       30000000      42.7 ns/op
```

I'll leave it up to you to decide if we should change Sprintf to `+`. :) Let me know when you're ready for me to merge it in!

mholt:
@rickb777 DO NOT WORRY ABOUT THE CLA problem. I know you've signed it. :P I will merge this PR. Do not make a new one.

mholt:
@rickb777 I think that does it, then! I'll merge it now. Thank you for your contribution!

I'd like to invite you to be a collaborator on the project, so you can submit patches from branches instead of forks, and also help review pull requests and issues. Generally I'm OK with merging pull requests after one or two other collaborators review and approve the pull request, if it has tests and is relatively small. Thank you again!

PR #1438, opened 2017-02-15
Fix data race for max connection limiting in proxy directive.

augustoroman:
The Conns and Unhealthy fields are updated concurrently across all active
requests.  Because of this, they must use atomic operations for reads and
writes.

Prior to this change, Conns was incremented atomically, but read unsafely.
Unhealthly was updated & read unsafely.  The new test
TestReverseProxyMaxConnLimit exposes this race when run with -race.

Switching to atomic operations makes the race detector happy.

### DIFF
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -49,6 +49,8 @@ type UpstreamHostDownFunc func(*UpstreamHost) bool
 
 // UpstreamHost represents a single proxy upstream
 type UpstreamHost struct {
+	// This field is read & written to concurrently, so all access must use
+	// atomic operations.
 	Conns             int64 // must be first field to be 64-bit aligned on 32-bit systems
 	MaxConns          int64
 	Name              string // hostname of this upstream host
@@ -59,7 +61,10 @@ type UpstreamHost struct {
 	WithoutPathPrefix string
 	ReverseProxy      *ReverseProxy
 	Fails             int32
-	Unhealthy         bool
+	// This is an int32 so that we can use atomic operations to do concurrent
+	// reads & writes to this value.  The default value of 0 indicates that it
+	// is healthy and any non-zero value indicates unhealthy.
+	Unhealthy int32
 }
 
 // Down checks whether the upstream host is down or not.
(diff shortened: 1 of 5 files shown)


caddyhttp/proxy/proxy.go
```diff
@@ -59,7 +61,10 @@ type UpstreamHost struct {
 	WithoutPathPrefix string
 	ReverseProxy      *ReverseProxy
 	Fails             int32
-	Unhealthy         bool
+	// This is an int32 so that we can use atomic operations to do concurrent
+	// reads & writes to this value.  The default value of 0 indicates that it
+	// is healthy and any non-zero value indicates unhealthy.
+	Unhealthy int32
```
mholt: Just to confirm, this is int32 so it's not subject to the last bug on this page, right? https://golang.org/pkg/sync/atomic/#pkg-note-BUG - if it is, this would be exposed on ARM systems, for example (Raspberry Pi).

caddyhttp/proxy/proxy_test.go
```diff
@@ -143,6 +144,83 @@ func TestReverseProxyInsecureSkipVerify(t *testing.T) {
...
+	}
+
+	// set up proxy
+	p := &Proxy{
+		Next:      httpserver.EmptyNext, // prevents panic in some cases when test fails
+		Upstreams: su,
+	}
+
+	// testComplete := make(chan bool)
+	// go func() {
```
mholt: Impressive test case. Thank you! I notice a few stray commented lines that made their way in.

caddyhttp/proxy/proxy.go
augustoroman: Oh neat, I hadn't considered that, actually.  I just selected int32 since it seemed large enough to hold the single bit value that we are putting in there. :-)

caddyhttp/proxy/proxy_test.go
augustoroman: Ooops!  Thanks, removed.

mholt:
Thank you @augustoroman! I really appreciate the initiative you took on this, and the care to write a thorough test. I'm going to invite you to be a collaborator so you can have push rights to the repository. This means you can contribute changes in a branch instead of a fork, and you can help review PRs and help with issues, etc. Once a PR has one or two reviews and is accepted from other collaborators, you can usually merge them (esp. if they're small and have tests).

augustoroman:
Thanks!  I'm happy to help where I can, and I appreciate the effort you've already put into making caddy great!  :-)

PR #1453, opened 2017-02-18
Push down headers from client

wendigo:
Addresses comments from #1215


caddyhttp/push/handler.go
```diff
@@ -62,9 +65,51 @@ outer:
...
+
+	return out
+}
+
+func (h Middleware) filterProxiedHeaders(headers http.Header) http.Header {
+
+	filter := http.Header{}
+
+	for _, header := range proxiedHeaders {
+		if val, ok := headers[http.CanonicalHeaderKey(header)]; ok {
```
mholt: It looks like the header names are already in the canonical format, as defined below.

caddyhttp/push/handler.go
```diff
@@ -44,14 +40,21 @@ outer:
...
+	// Serve later
+	code, err := h.Next.ServeHTTP(w, r)
+
 	if links, exists := w.Header()["Link"]; exists {
-		h.pushLinks(pusher, links)
+		h.servePreloadLinks(pusher, headers, links)
+	}
+
+	if flusher, ok := w.(http.Flusher); ok {
+		flusher.Flush()
```
wendigo: I can remove that. It seems it's not needed.

caddyhttp/push/handler.go
wendigo: You're right :) Removing

PR #1456, opened 2017-02-19
Disable TLS completely if there is no listener with tls enabled

wendigo:
Fixes https://github.com/mholt/caddy/issues/1455 (hope so)


wendigo:
Needs review @tobya @mholt

mholt:
Thanks @wendigo!

PR #1457, opened 2017-02-19
Don't return tls.Config if client does not support SNI

wendigo:
Clients without SNI support are very rare (http://caniuse.com/#search=SNI). 

Calling Caddy without SNI support ends with connection being established via TLS 1.0 and bad certificate (first in the list of domains). It should not be supported given the fact that 98.18% clients support it (lowering security for unsecured devices is not really a good idea).


mholt:
I'll look at this in a bit! One thing to consider is that some modern clients (like Firefox) do not use SNI when connecting to IP addresses.

wendigo:
If you're hosting multiple domains on the same IP it does not make sense at all to connect to IP (which domain/vhost should be presented in this situation?)

wendigo:
I guess we should add support for following:

- if there is single domain - we can always fallback on it
- if there are multiple (sub)domains - mark which one is default for fallback if SNI is disabled

This adds more control from the user perspective

mholt:
Hope you don't mind, but I took this into consideration (somewhat, at least!) when I did #1466, which I think supersedes this change.

PR #1462, opened 2017-02-20
Revert removed method

### LINKED ISSUE #1461: Removed method IncrNest breaks caddy filter
https://github.com/mholt/caddy/pull/1379 removed IncrNest() method which is used by caddy filter plugin

wendigo:
Fixes #1461

### DIFF
--- a/caddyfile/dispenser.go
+++ b/caddyfile/dispenser.go
@@ -120,6 +120,12 @@ func (d *Dispenser) NextBlock() bool {
 	return true
 }
 
+// IncrNest adds a level of nesting to the dispenser.
+func (d *Dispenser) IncrNest() {
+	d.nesting++
+	return
+}
+
 // Val gets the text of the current token. If there is no token
 // loaded, it returns empty string.
 func (d *Dispenser) Val() string {


wendigo:
Calling @tobya, @mholt

tobya:
This brings up an interesting point about breaking plugins.  Do we need/watn to run plugin builds with travis-ci as well as core caddy, to see when core code changes break plugins.

mholt:
Good question, and the way I've approached this in the new build server is that Caddy core is the boss: nothing blocks the release of Caddy (as long as its tests pass, etc.) - even if it means that a plugin breaks.

That's kind of harsh, but:

1) We're pre-1.0 so we move quick and break things (I'm still not 100% sure what the 1.0 compatibility guarantee looks like, but there will be one).
2) Depending on compute resources available, a possible future upgrade to the new build server would be to re-build all plugins with Caddy and notify plugin authors if their plugin has broken due to a Caddy update.

We took out IncrNest because we decided we don't want the Caddyfile to have deep nesting (more than 2 blocks) -- in other words, directives cannot nest blocks inside them.

I'm going to revert this change since we want to get rid of IncrNest, and we do not let plugins block the advancement of Caddy core.

@echocat You have some time to update the filter plugin -- I'm sorry about this -- if you would like any assistance please let me know. :)

PR #1466, opened 2017-02-21
tls: Refactor internals related to TLS configurations

mholt:
This builds on @wendigo's recent PRs to use Go 1.8 features in the caddytls package.

I moved some logic around and simplified a few functions, as well as -- I think -- fixing some bugs or edge cases, like what to do with clients who don't use SNI: we return _any_ tls.Config for a handshake that doesn't match any configs by the hostname. I figure any tls.Config is better than none: a TLS alert will be deferred to loading the certificate, or browsers will at least show warnings. The certificate store is process-wide, so which config is used shouldn't entirely matter when accessing it. It also means we can serve clients who don't support SNI, although I care about that less and less as time goes on. I hope this decision won't bite, but I don't think there's a perfect solution.

A couple syntax changes:

- `muststaple` --> `must_staple` (not commonly used, which is probably a good thing for now, so I'm not too worried)
- `http2 off` --> `alpn http/1.1`. I realized that the TLS directive should not really know about HTTP, but it can be told what protocols to send for ALPN. So disabling HTTP/2 is now done by telling the TLS config that it supports only HTTP/1.1.

I also revised how we read the ClientHello for our MITM detector. I defer all reading off the wire to the standard library's crypto/tls package Read() code, because it's highly robust and handles all the hard stuff. I tee off of that into a small buffer that is cleared after the ClientHello is collected and parsed. Load tests showed no noticeable performance degradation.


caddyhttp/httpserver/server.go
```diff
@@ -122,6 +124,61 @@ func NewServer(addr string, group []*SiteConfig) (*Server, error) {
...
+	}
+	if !min.WriteTimeoutSet {
+		min.WriteTimeout = defaultTimeouts.WriteTimeout
+	}
+	if !min.IdleTimeoutSet {
+		min.IdleTimeout = defaultTimeouts.IdleTimeout
+	}
+
+	// set the final values on the server
+	s := &http.Server{Addr: addr}
```
wendigo:
```
&http.Server{
  Addr: addr,
  ReadTimeout: min.ReadTimeout,
  ReadHeaderTimeout: min.ReadHeaderTimeout,
  WriteTimeout: min.WriteTimeout,
  IdleTimeout: min.IdleTimeout,
}
```

mholt: Oh yeah. Thanks. :)

PR #1474, opened 2017-02-25
Adding support for ServerIP context

jtyr:
This PR is adding support for `ServerIP` context which can be used to get the local server IP. This is useful for example to print out the server IP in a template (e.g. `Server IP: {{.ServerIP}}`) when testing several instances of Caddy behind a load balancer.

### DIFF
--- a/caddyhttp/httpserver/context.go
+++ b/caddyhttp/httpserver/context.go
@@ -84,6 +84,29 @@ func (c Context) IP() string {
 	return ip
 }
 
+// To mock the net.InterfaceAddrs from the test.
+var networkInterfacesFn = net.InterfaceAddrs
+
+// ServerIP gets the (local) IP address of the server.
+// TODO: The bind directive should be honored in this method (see PR #1474).
+func (c Context) ServerIP() string {
+	addrs, err := networkInterfacesFn()
+	if err != nil {
+		return ""
+	}
+
+	for _, address := range addrs {
+		// Validate the address and check if it's not a loopback
+		if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil || ipnet.IP.To16() != nil {
+				return ipnet.IP.String()
+			}
+		}
+	}
+
+	return ""
+}
+
 // URI returns the raw, unprocessed request URI (including query
 // string and hash) obtained directly from the Request-Line of
 // the HTTP request.
--- a/caddyhttp/httpserver/context_test.go
+++ b/caddyhttp/httpserver/context_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -298,6 +299,44 @@ func TestIP(t *testing.T) {
 
 		if actualIP != test.expectedIP {
 			t.Errorf(testPrefix+"Expected IP %s, found %s", test.expectedIP, actualIP)
+		}
+	}
+}
+
+type myIP string
+
+func (ip myIP) mockInterfaces() ([]net.Addr, error) {
+	a := net.ParseIP(string(ip))
+
+	return []net.Addr{
+		&net.IPNet{IP: a, Mask: nil},
+	}, nil
+}
+
+func TestServerIP(t *testing.T) {
+	context := getContextOrFail(t)
+
+	tests := []string{
+		// Test 0 - ipv4
+		"1.1.1.1",
+		// Test 1 - ipv6
+		"2001:db8:a0b:12f0::1",
+	}
+
+	for i, expectedIP := range tests {
+		testPrefix := getTestPrefix(i)
+
+		// Mock the network interface
+		ip := myIP(expectedIP)
+		networkInterfacesFn = ip.mockInterfaces
+		defer func() {
+			networkInterfacesFn = net.InterfaceAddrs
+		}()
+
+		actualIP := context.ServerIP()
+
+		if actualIP != expectedIP {
+			t.Errorf("%sExpected IP \"%s\", found \"%s\".", testPrefix, expectedIP, actualIP)
 		}
 	}
 }


mholt:
Thank you, @jtyr! Really appreciate you first contribution. :smile: Could you add a test for this function? ... not sure how you'd hard-code it, you might have to do a kind of round-about test that makes sure that a loopback address is not returned... or something like that.

jtyr:
Unfortunately I'm quite new to Go programming (this PR is my first code!) so I'm afraid that I don't know how to write and run Go tests. Any hints would be appreciated.

jtyr:
I tried to implement the test:

https://github.com/jtyr/caddy/commit/8fd2867124de82c976d627c74ebe8920f51ac814

It probably doesn't work as I don't know how to test it.

tobya:
You can run tests by calling

     go test -v github.com/mholt/caddy/caddyhttp/httpserver

that should run all the tests on all the sub dir from httpserver.

mholt:
^ Yes, or just `cd` into that folder and run `go test`. Let us know how that goes. :)

jtyr:
I don't know how to create the `net.Addr` data structure. I was trying to use `net.ParseCIDR("1.1.1.1/32")` to get the `net.IPNet` but I don't know how to cast it to `net.Addr` :o(

mholt:
@jtyr I don't think you have to create one. Just see if the output of the function is within the requirements: not empty, not 127.* ... that'd probably do for starters.

jtyr:
But I have to somehow simulate the `addrs` from `addrs, err := net.InterfaceAddrs()`. And that's an array of `net.Addr`. So how to create it?

mholt:
You don't have to simulate it, you just have to check the return value of the function.

jtyr:
But the `ServerIP` function is reading my local IP. In order to test the function I have to be able to simulate the output of `net.InterfaceAddrs` in order to compare the value the function returns with the expected value. For that I wanted to extend the `Context` type and update it from the test function (see the linked commit above).

mholt:
To do that approach, define an unexported variable outside of the function in context.go:

```go
var networkInterfacesFn = net.InterfaceAddrs
```

Then call `networkInterfacesFn()` inside your ServerIP function. The test can then replace the `networkInterfacesFn` variable with something custom just for the test, to return contrived, consistent values.

Since `net.Addr` is an interface, your test function can define any unexported type that implements [its two methods](https://sourcegraph.com/github.com/golang/go@go1.8/-/blob/src/net/net.go#L109:6-109:10) and use that. Just define a struct with 2 string values and have those two methods return the values.

jtyr:
I really appreciate your help. Unfortunately I know too little about Go which doesn't allow me to implement the proper test. Instead, I will try to implement the simple test which will test for empty string and string starting with `127.`.

jtyr:
I have added the simple test. It seems to run just fine:

```
=== RUN   TestServerIP
--- PASS: TestServerIP (0.00s)
```

jtyr:
Btw. I tried to implement the test with the simulated `net.Addr` but I wasn't able to assert it to `net.IPNet` type. Any idea what I'm doing wrong?

https://play.golang.org/p/_OIrBZIPX9

mholt:
@jtyr The value there isn't a *net.IPNet type, so the assertion fails. You don't need a type assertion. Good start -- I made a few revisions/simplifications; give this a try: https://play.golang.org/p/ci2ChGZTfI

jtyr:
I see. As per your example, the `myAddr` should be `net.IPNet` compatible to make the code working:

https://play.golang.org/p/2lVFRItpca

jtyr:
I don't understand why the `addrs` (from `addrs, err := net.InterfaceAddrs()`) is of type `[]net.Addr` (which is an array of `net.Addr`, right?) but the `addrs[0]` is of type `*net.IPNet`. Is there any explanation for that?

jtyr:
The [`net.newAddr()`](https://golang.org/src/net/interface_linux.go) should return `Addr` but the `return` has as its argument an address to the `net.NetIP` instead. How's that possible?

mholt:
@jtyr net.IPNet implements the net.Addr interface. It's usually good practice to return a concrete type rather than an interface, where possible. Concrete types are easily, implicitly used as interfaces as long as they implement it. See this: https://tour.golang.org/methods/10

jtyr:
I see! I was looking into the `net.IPNet` struct only and did not see that there are actually `IPNet` methods (functions with the `IPNet` receiver) which are the same like in the `net.Addr` interface (`Network` and `String`). It's quite difficult to get my head around the OOP in Go without classical classes and explicit inheritance.

jtyr:
@mholt I got it working! ;o) I'm just not sure if it's the best way to set the `mockInterfaces` function through the `context`. Is there a better way?

caddyhttp/httpserver/context.go
```diff
@@ -27,6 +27,7 @@ type Context struct {
 	Req  *http.Request
 	URL  *url.URL
 	Args []interface{} // defined by arguments to .Include
+	NIFn func() ([]net.Addr, error)
```
mholt: I don't think this belongs in the Context struct; it can be a package-level variable, like in the playground I showed you. The reason is that it isn't changed per Context: we only change it for tests, and the whole process is either for testing or not. Make sure it's un-exported, so, starts with a lower-case letter.

caddyhttp/httpserver/context.go
```diff
@@ -84,6 +85,30 @@ func (c Context) IP() string {
 	return ip
 }
 
+// ServerIP gets the (local) IP address of the server.
+func (c Context) ServerIP() string {
+	if c.NIFn == nil {
+		c.NIFn = net.InterfaceAddrs
+	}
```
mholt: This if won't be necessary because NIFn will always have a value

caddyhttp/httpserver/context.go
mholt: Also, rename NIFn to something more descriptive, like in the play example.

caddyhttp/httpserver/context.go
```diff
@@ -84,6 +85,30 @@ func (c Context) IP() string {
...
+
+	addrs, err := c.NIFn()
+
+	if err != nil {
```
mholt: This is a nit-pick, but conventionally in Go, you'll do the error check on the line immediately after calling your function, without an empty line in between.

caddyhttp/httpserver/context_test.go
```diff
@@ -302,6 +303,41 @@ func TestIP(t *testing.T) {
...
+		// Test 1 - ipv6
+		"2001:db8:a0b:12f0::1",
+	}
+
+	for i, testIP := range tests {
+		testPrefix := getTestPrefix(i)
+
+		// Fake the network interfaces
+		ip := myIP(testIP)
+		context.NIFn = ip.mockInterfaces
```
mholt: Once you make this a package variable, just set this at the top of the TestServerIP function and use `defer func() { ... }` to set it back to what it was (just as a nice-thing-to-do, keeping your tests clean and have fewer side-effects in case something else would be changed by it).

jtyr:
@mholt I have implemented all your comments. Please feel free to review it.

mholt:
Thanks @jtyr! Does this function respect the `bind` directive? In other words, if the server listener is bound specifically to a certain interface, does this function return that? For reference, the ListenHost is set on the `SiteConfig`. I'm not sure if that would matter here, but I wonder if it would be confusing for somebody who binds to some IP address to have another one shown when invoking .ServerIP in templates. Does that make sense?

jtyr:
Yes, It makes sense. I will have a look into it.

jtyr:
@mholt Please could you give me a hint how to access the `cfg.TLS.ListenHost` value?

mholt:
@jtyr You want the ListenHost on the `SiteConfig` struct in siteconfig.go, not the TLS config. ;)

jtyr:
I was thinking to add something like this at the beginning of the  ServerIP method:

```
// WHERE TO GET THE c *caddy.Controller???
cfg := config := httpserver.GetConfig(c)
bindIP := net.ParseIP(cfg.ListenHost)
if bindIP == nil {
    return ""
}

// Check if the bind configuration is a specific non-loopback IP
if cfg.ListenHost != "0.0.0.0" && cfg.ListenHost != "::" && !bindIP.IsLoopback() {
    return cfg.ListenHost
}
```

But I don't know where to get the `c *caddy.Controller` ;o(

mholt:
Oh... yeah, I don't either... ha.

I kinda just want to get this merged in. We'll see if anyone complains about this and handle it then. Can you just put a "TODO" comment in the code, reminding that the `bind` directive will need to be honored in the function? Then I can merge this.

jtyr:
I have added the TODO note. Please feel free to merge it.

mholt:
Thanks for your effort and patience! :+1:

Issue #1346, opened 2017-01-11
caddy download tarballs are tarbombs?


winny-:
caddy download tarballs are tarbombs? There is probably a well thought out reason behind a root-less tarball, but I figured it wouldn't hurt to ask on the issue tracker.

https://en.wikipedia.org/wiki/Tar_(computing)#Tarbomb

dprandzioch:
Actually also drives me crazy from time to time when setting up a host... +1 👍

mholt:
I guess now's a better time than ever to change this, since I'm rewriting the website and all the build server code.

mholt:
Hmm, wait. This would be the responsibility of whatever extraction program you're using, isn't it? My Mac always extracts to a folder when I open an archive, even if the contents weren't originally in one (sometimes that drives me crazy).

winny-:
That's one of the nicer features of Finder. It always ensures an archive has a root directory. Unfortunately that isn't the case for `tar`.

mholt:
Hmm, so isn't it good practice to untar into an empty directory, since tar doesn't do that for you?

pbeckmann:
Yes, it should be good practice. In fact i think, we are just too lazy. We could look into the tarball before with `tar tf` and see if there is a root folder or not. Anyway i think it would be nice to have the root folder there. But if its too much overhead, then I can deal with it.

Maybe we should create a PR to the tar source code :thinking: :laughing:

mholt:
Would love that. Saves me lot of time so I can focus on (frankly) more important things. ;)

I was looking into doing this in [archiver](https://github.com/mholt/archiver). If the archive name had a trailing file separator (filepath.Separator), it would bundle the top-level files within a folder called after the name of the archive (without the extension).

For example:

```
$ archiver make mytar.tar.gz/ item1.txt item2.txt dir1
```

Would create a file called "mytar.tar.gz" and within it, a folder called "mytar", and within that, three items (two text files and a directory with its contents). Make sense? Anyway, I'd accept PRs there. It was just taking a little more time than I wanted right now.

Issue #1476, opened 2017-02-26
{path} is logging {uri}


tobya:
PR #1364 introduced a bug where `{path}` is now actually logging the full path?query rather than just the path when a rewrite has been performed.

I have a fix for this but I need to write tests to try to prevent something similar in the future.  Just wanted to log it so we know it is in progress.

tobya:
Currently defered until after #1496

tobya:
Fixed with #1481

PR #1481, opened 2017-02-27
Fix issue #1346 {path} logging {uri} and add {rewrite_uri) placeholder

tobya:
Fix for issue #1476 . PR #1364 introduced a bug where when a rewrite was performed  {path} logged path?query rather than just {path}.

This PR fixes it and adds a test.

Also while working on this I realised we needed a {rewrite_uri} placeholder and {uri} now correctly provides the non rewritten value.


caddyhttp/httpserver/replacer.go
```diff
@@ -232,15 +232,23 @@ func (r *replacer) getSubstitution(key string) string {
...
 		// if a rewrite has happened, the original URI should be used as the path
 		// rather than the rewritten URI
-		path := r.request.Header.Get("Caddy-Rewrite-Original-URI")
-		if path == "" {
+		var path string
+		origpath := r.request.Header.Get("Caddy-Rewrite-Original-URI")
+		if origpath == "" {
 			path = r.request.URL.Path
+		} else {
+			origURL, _ := url.Parse(origpath)
```
tobya: I am assuming that it is safe to ignore err here as Caddy-Rewrite-Origional-URI should always be a valid URL.  Is this ok?

caddyhttp/httpserver/replacer.go
```diff
@@ -232,15 +232,23 @@ func (r *replacer) getSubstitution(key string) string {
...
+			path = origURL.Path
 		}
 		return path
 	case "{path_escaped}":
-		path := r.request.Header.Get("Caddy-Rewrite-Original-URI")
-		if path == "" {
+		var path string
+		origpath := r.request.Header.Get("Caddy-Rewrite-Original-URI")
```
tobya: This Header field name Caddy-Rewrite-Original-URI  is used in several files (Fastcgi etc) is there a way to have a system wide const?

caddyhttp/httpserver/replacer.go
mholt: Yeah, I dunno what else you would do, since we can't return an error.

caddyhttp/httpserver/replacer.go
mholt: This reminds me, we're switching to using the `context` package... #1331 does this for something similar, and frankly I only used a header field because context didn't exist at the time so it was kind of a hack. What do you think if we switch this to use req.Context? (It can be a separate PR, I prefer multiple smaller PRs!)

tobya: I'll take a look.

tobya:
This PR has now been updated with changes made in #1496 and test have been updated to use context

caddy.go
```diff
@@ -877,3 +877,7 @@ type CtxKey string
 // context.WithValue to access the original request URI that accompanied the
 // server request. The associated value will be of type string.
 const URLPathCtxKey CtxKey = "url_path"
+
+// URIxRewriteCtxKey is a context key used to store original unrewritten
+// URI in context.WithValue
+const URIxRewriteCtxKey CtxKey = "caddy_rewrite_original_uri"
```
mholt: Now that I see URLPathCtxKey and URIxRewriteCtxKey constants, I'm beginning to think they won't necessarily be used outside of the httpserver and am thinking these should be declared in the `httpserver` package instead.

caddyhttp/rewrite/to.go
```diff
@@ -49,7 +51,7 @@ func To(fs http.FileSystem, r *http.Request, to string, replacer httpserver.Repl
 
 	// take note of this rewrite for internal use by fastcgi
 	// all we need is the URI, not full URL
-	r.Header.Set(headerFieldName, r.URL.RequestURI())
+	*r = *r.WithContext(context.WithValue(r.Context(), caddy.URIxRewriteCtxKey, r.URL.RequestURI()))
```
mholt: I'm a little confused by the need to dereference `r` at all... can you help me understand that choice?

tobya: This was the bit that gave me a little bit of a headache, but essentially if I just added context to r since this call `To` just returns and doesnt not pass the request further down (if that makes sense), the context that I had added was not available outside the function unless I dereferenced r.

caddy.go
tobya: Ok, I'll take a look at doing this.

tobya: should I add them to httpserver/server.go or is there another file that makes more sense

caddyhttp/rewrite/to.go
mholt: Ah. That makes sense. Thank you.

caddy.go
mholt: @tobya That could work, although I think httpserver/middleware.go might make more sense, since that file is more directly about helping middleware with what they need.

tobya: @mholt The Context keys seems to be used in fastcgi, rewrite and staticfiles packages.  I can move them, but perhaps it makes more sense just to leave them in caddy.go

mholt: @tobya But what use does a certain Context key have with the other server types? I think they should go in httpserver...

tobya: @mholt I seem to get a circular reference between `httpserver` and `staticfiles` since `httpserver`  imports `staticfile` and in order to see the context keys in the `httpserver` package `staticfile` would need to import `httpserver`.  Am I missing something obvious here?

mholt: Nuts. This goes back to the debate I had when I designed Caddy 0.9 about whether the staticfiles package should be in the httpserver package as one package. It's kind of an edge case because the staticfiles is technically a middleware but it's the "base" middleware that is always used with the httpserver. So, uh... I guess that's one fail of my decision earlier.

Sigh. Let's leave the consts where they are, but put a TODO in a comment saying we want them in the httpserver package eventually, but we haven't done that yet since it creates a circular import with the staticfiles package. Then we can merge.

mholt:
Phew, thanks @tobya. Really appreciate your patience and persistence. It paid off. :)

PR #1482, opened 2017-02-28
Set right MIME type in middleware templates

crvv:
The issue is caddy will return wrong "Content-Type" (application/x-gzip) if "gzip" and "templates" are used at the same time.

To reproduce, run these code
```
cat << EOF > Caddyfile
localhost:2015 {
    gzip
    templates
}
EOF
echo test > index.html

caddy &
curl http://localhost:2015 -H 'accept-encoding: gzip' -v
```
and got
```
< HTTP/1.1 200 OK
< Content-Encoding: gzip
< Last-Modified: Tue, 28 Feb 2017 10:42:03 GMT
< Server: Caddy
< Vary: Accept-Encoding
< Date: Tue, 28 Feb 2017 10:43:11 GMT
< Content-Length: 29
< Content-Type: application/x-gzip
```

And the pull request fixed this.

### DIFF
--- a/caddyhttp/templates/templates.go
+++ b/caddyhttp/templates/templates.go
@@ -4,6 +4,7 @@ package templates
 
 import (
 	"bytes"
+	"mime"
 	"net/http"
 	"os"
 	"path"
@@ -61,6 +62,16 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 				if err != nil {
 					return http.StatusInternalServerError, err
 				}
+
+				// If Content-Type isn't set here, http.ResponseWriter.Write
+				// will set it according to response body. But other middleware
+				// such as gzip can modify response body, then Content-Type
+				// detected by http.ResponseWriter.Write is wrong.
+				ctype := mime.TypeByExtension(ext)
+				if ctype == "" {
+					ctype = http.DetectContentType(buf.Bytes())
+				}
+				w.Header().Set("Content-Type", ctype)
 
 				templateInfo, err := os.Stat(templatePath)
 				if err == nil {


mholt:
Hey @crvv, thanks for the PR!

I don't suppose you could add a test case could you? :no_mouth:

crvv:
I tried to add a simple test in `templates`. But I am not sure this is right way.

Should I add a test which use the Caddyfile mentioned above to start a full Caddy HTTP server and send real HTTP request to make sure that Go won't set wrong `Content-Type`?

mholt:
@crvv Thank you!

> But I am not sure this is the right way.

This is good enough for now -- in fact, we only probably need one of the assertions, not both, since I don't really see the difference between them.

Does the new test fail without your change? If so, then that's good.

> Should I add a test which use the Caddyfile mentioned above to start a full Caddy HTTP server and send real HTTP request to make sure that Go won't set wrong Content-Type?

No, we definitely need a way to run full system tests like this, which I plan to build near/around Caddy version 1.0. @hackintoshrao started something like this a while back but it's not a simple task, so it didn't come full circle. But we'll get there eventually.

crvv:
> Does the new test fail without your change? If so, then that's good.

I tested this today, but it passed without my change.

The reason is `templates.ServeHTTP` function calls `http.ResponseWriter.Write`. `http.ResponseWriter.Write` will "adds a Content-Type set to the result of passing the initial 512 bytes of written data to DetectContentType."
This is also the reason why this bug only occurs when use `templates` and `gzip` simultaneously.

There are similar situations in middlewares `markdown` and `browse`, and they both set `Content-Type` in `ServeHTTP` function.

Maybe I should remove this useless test.

caddyhttp/templates/templates.go
```diff
@@ -62,6 +63,12 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 					return http.StatusInternalServerError, err
 				}
 
+				ctype := mime.TypeByExtension(ext)
```
mholt: Hmm, yeah, okay - I've given it some more thought, and I can see why the test isn't effective. We'll need a more advanced test harness to take care of that. Go ahead and remove the test case, but put a comment above this line explaining why we set the Content-Type here. Thanks!

mholt:
Great, thank you @crvv! Really appreciate your contribution. :)

PR #1488, opened 2017-03-02
Fixed #1484

### LINKED ISSUE #1484: Nil pointer errors when using websockets

0.9.5

Setup Caddy as reverse HTTPS proxy over [Rancher](//rancher.com) and its Kubernetes service.

```text
rancher.example.com
bind 0.0.0.0
log stdout
errors stderr
tls email@address.com
timeouts 0
proxy / http://rancher:8080 {
  keepalive 30
  header_upstream Host {host}
  header_upstream X-Real-IP {remote}
  header_upstream X-Forwarded-For {remote}
  header_upstream X-Forwarded-Proto {scheme}
  websocket
}
```

Docker-Compose section:

```yaml
caddy:
    image: abiosoft/caddy
    command: ["-conf=/etc/Caddyfile", "-agree", "-http2=false"]
    ports:
      - 80:80
      - 443:443
    restart: always
    volumes:
      - ./caddy:/root/.caddy:rw
      - ./Caddyfile:/etc/Caddyfile:ro
    links:
      - rancher
```

Rancher opening a WebSocket for its web-based kubectl console and Caddy proxying that to the backend.

Caddy answered with a 500, the websocket was not passed to the backend, and Caddy logs showed this:

```
28/Feb/2017:21:50:02 +0000
[PANIC /v2-beta/projects/1a10/subscribe?eventNames=resource.change&limit=-1&sockId=12]
caddyhttp/proxy/reverseproxy.go:408 -
runtime error: invalid memory address or nil pointer dereference
```

(log message broken up on multiple lines for clarity)

This is where it's panicking: https://github.com/mholt/caddy/blob/v0.9.5/caddyhttp/proxy/reverseproxy.go#L408

Not sure. I'll try setting up local caddy with a sample websocket server and see if I can replicate.

lhecker:
#1484. The fix is luckily very simple (it's literally only `tlsClientConfig != nil &&`). It can currently be tested by setting the default value of `httpserver.HTTP2` to `false` in `caddyhttp/httpserver/plugin.go` and then running `go test -run=TestWebSocketReverseProxyFromWSClient` (which I did).

@passcod If you know how and if you have the time it'd be really cool if you could test this PR by building caddy from source. 🙂

P.S.: This PR sadly does not contain a real testcase for 4e31a26. Reason being that we currently run all of our tests with `httpserver.HTTP2` set to `true`, which causes `TLSClientConfig` to be set and thus this error to never appear. Adding another testcase just for this very simple bug does not seem particularly helpful to me. Instead we should probably rather finally rewrite all the proxy tests to be more modular. Currently the entire `proxy_test.go` file is >1000 lines of largely redundant code. If it were more modular we could then easily run the same test e.g. twice: Once with `httpserver.HTTP2` set to `true` and once while it's set to `false`. But this should probably rather happen in a seperate PR.

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -224,7 +224,10 @@ func (rp *ReverseProxy) UseInsecureTransport() {
 		}
 		rp.Transport = transport
 	} else if transport, ok := rp.Transport.(*http.Transport); ok {
-		transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+		if transport.TLSClientConfig == nil {
+			transport.TLSClientConfig = &tls.Config{}
+		}
+		transport.TLSClientConfig.InsecureSkipVerify = true
 		// No http2.ConfigureTransport() here.
 		// For now this is only added in places where
 		// an http.Transport is actually created.
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -441,7 +441,7 @@ func newConnHijackerTransport(base http.RoundTripper) *connHijackerTransport {
 	}
 	if b, _ := base.(*http.Transport); b != nil {
 		tlsClientConfig := b.TLSClientConfig
-		if tlsClientConfig.NextProtos != nil {
+		if tlsClientConfig != nil && tlsClientConfig.NextProtos != nil {
 			tlsClientConfig = tlsClientConfig.Clone()
 			tlsClientConfig.NextProtos = nil
 		}


passcod:
Ah, apologies, I don't have the setup I used for this anymore, so I can't test it easily.

mholt:
@lhecker Did you replicate the behavior or was this change just a guess?

I just want _some_ verification that this does what it is supposed to.

lhecker:
@mholt I verified the change manually. Here's how I did it:

Check out the current `master` branch and set the [default of `httpserver.HTTP2`](https://github.com/mholt/caddy/blob/4e31a26af9762172e4e39f3c35d8477e87b5bb9b/caddyhttp/httpserver/plugin.go#L26) to `false`. Afterwards just run the following in the ` caddyhttp/proxy` subfolder:
```
`go test -run=TestWebSocketReverseProxyFromWSClient`
```

You'll see that the test fails even though it doesn't rely on SSL nor HTTP/2. If you check out this branch though and again set the default above to `false` it'll work, proving that my PR fixes #1484.

mholt:
Fantastic!! Thanks @lhecker, much appreciated.

PR #1491, opened 2017-03-03
Support realms with basic authentication

jung-kurt:
Web browsers can be configured to save authenticated credentials for a combination of server and realm. Currently, the basicauth middleware always uses the default realm "Restricted". This PR supports the association of a custom realm with one or more protected resources. The benefit is that different credentials can be remembered by a browser for different groups of resources, something that applies to one of my use cases.

The changes are backward compatible. However, within a nested block, a resource to be protected can be placed after the `path` subdirective. This is in addition to simply naming a file or directory by itself as is supported currently. The new subdirective support seems to be more in line with most other middleware directives and looks a bit nicer when used with the proposed `realm` keyword.

### DIFF
--- a/caddyhttp/basicauth/basicauth_test.go
+++ b/caddyhttp/basicauth/basicauth_test.go
@@ -25,11 +25,20 @@ func TestBasicAuth(t *testing.T) {
 		}
 		return http.StatusOK, nil
 	}
-	rw := BasicAuth{
-		Next: httpserver.HandlerFunc(upstreamHandler),
-		Rules: []Rule{
-			{Username: "okuser", Password: PlainMatcher("okpass"),
-				Resources: []string{"/testing"}, Realm: "Resources"},
+	rws := []BasicAuth{
+		{
+			Next: httpserver.HandlerFunc(upstreamHandler),
+			Rules: []Rule{
+				{Username: "okuser", Password: PlainMatcher("okpass"),
+					Resources: []string{"/testing"}, Realm: "Resources"},
+			},
+		},
+		{
+			Next: httpserver.HandlerFunc(upstreamHandler),
+			Rules: []Rule{
+				{Username: "okuser", Password: PlainMatcher("okpass"),
+					Resources: []string{"/testing"}},
+			},
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


caddyhttp/basicauth/basicauth_test.go
```diff
@@ -70,7 +71,7 @@ func TestBasicAuth(t *testing.T) {
 		if test.result == http.StatusUnauthorized {
 			headers := rec.Header()
 			if val, ok := headers["Www-Authenticate"]; ok {
-				if got, want := val[0], "Basic realm=\"Restricted\""; got != want {
+				if got, want := val[0], "Basic realm=\"Resources\""; got != want {
```
pennywisdom: We still have a default for "Restricted" I think it would be good to have a test still checking for this, so we are aware of this default and any changes. Looks like the change of test has replaced the existing string and maybe adding an additional check for the new changes plus the original default value would be nice.

jung-kurt: I just committed a revision that tests the case in which the realm subdirective is missing.

pennywisdom:
@mholt ok to merge?

caddyhttp/basicauth/setup.go
```diff
@@ -68,6 +61,31 @@ func basicAuthParse(c *caddy.Controller) ([]Rule, error) {
...
+			switch len(args) {
+			case 0:
+				// For backwards compatibility, assume single argument is path resource
+				rule.Resources = append(rule.Resources, val)
+			case 1:
+				if val == "realm" {
+					if rule.Realm == "" {
+						rule.Realm = strings.Replace(args[0], `"`, `\"`, -1)
+					} else {
+						return rules, c.Errf("\"realm\"subdirective can only be specified once")
```
mholt: There's a space missing before subdirective

caddyhttp/basicauth/setup_test.go
```diff
@@ -70,6 +70,30 @@ md5:$apr1$l42y8rex$pOA2VJ0x/0TwaFeAF9nX61`
...
+		{`basicauth user pwd {
+			path /resource1
+			path /resource2
+		}`, false, "pwd", []Rule{
+			{Username: "user", Resources: []string{"/resource1", "/resource2"}},
+		}},
+		{`basicauth user pwd {
+			/resource1
+			realm "Secure resources"
+			path /resource2
```
mholt: The path/realm syntax is slightly awkward... not bad, but I wonder if we could improve it. What if we moved the realm up to the first line as an optional third argument?

```plain
basicauth user pwd realm {
    paths...
}
```

That way we wouldn't need the `path`/`realm` subdirectives and you can't specify the realm more than once.

caddyhttp/basicauth/setup.go
jung-kurt: Good catch -- will correct that

caddyhttp/basicauth/setup_test.go
jung-kurt: That was my first approach. The problem (not insurmountable) is that when there are three arguments, the intent could be 
```
basicauth path user pwd
```
or
```
basicauth user pwd realm {
  resource 1
  ...
  resource N
}
```
That is, it depends on whether the directive line precedes a nested block. After I had coded it that way, the whole thing seemed fragile and not very future-proof. That's when I decided the use of subdirectives seemed cleaner. I even thought of supporting subdirectives for username and password so the basicauth directive could be free of arguments. I'm fine with reverting to the optional third/fourth argument, but I think that if this directive is ever extended again, it could get messy.

jung-kurt: One additional note, the example you highlight showing both the isolated resource and the one preceded by the `path` subdirective is not intended as an example, just as a test to show that both forms are supported. I would hope that going forward, people would use the `path` and `realm` subdirectives.

mholt: Good point, about the optional path. I forgot about that.

What if we keep your changes as-is, but drop the "path" keyword? In other words, each line is a list of resources, unless the line starts with "realm" (and has two arguments). That way, there's only one way to do it in that case.

jung-kurt: Sounds good. I'll revise the caddyserver.com documentation to suit.

mholt:
Excellent. Thank you @jung-kurt and @pennywisdom both!

jung-kurt:
Thanks, @mholt and @pennywisdom. I just revised the caddyserver.com documentation to account for the change.

PR #1492, opened 2017-03-04
staticfiles: add Content-Length header (closes #1479)

### LINKED ISSUE #1479: gzip: no Content-Length header is set for file downloads

Caddy 0.9.5 (abiosoft/caddy:latest Docker container)

Downloading a file, using gzip. `curl -LO --compressed -v http://example.com/url/to/big/file`

```text
example.com:80 {
        tls off
        gzip
        root /www
}
```

Default abiosoft/caddy:0.9.5 Docker container and the above Caddyfile

A **Content-Length** header set like this: 
```
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Content-Length: 50020064
< Content-Type: application/octet-stream
< Etag: W/"58b43808-2fb3ee0"
< Last-Modified: Mon, 27 Feb 2017 14:30:32 GMT
< Server: Caddy
< Date: Mon, 27 Feb 2017 15:58:20 GMT
```
This is the curl output from running `curl -LO -v http://example.com/url/to/big/file` (no gzip)

No **Content-Length** header present in the response.

```
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Content-Encoding: gzip
< Content-Type: application/octet-stream
< Etag: W/"58b43808-2fb3ee0"
< Last-Modified: Mon, 27 Feb 2017 14:30:32 GMT
< Server: Caddy
< Vary: Accept-Encoding
< Date: Mon, 27 Feb 2017 16:02:36 GMT
< Transfer-Encoding: chunked
```
This is the curl output from running `curl -LO --compressed -v http://example.com/url/to/big/file` (with gzip)

#### Note: **Content-Length**  is set when I download a small .html file
```
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Content-Encoding: gzip
< Content-Type: text/html; charset=utf-8
< Etag: W/"588f5fc8-1900"
< Last-Modified: Mon, 30 Jan 2017 15:46:16 GMT
< Server: Caddy
< Vary: Accept-Encoding
< Date: Mon, 27 Feb 2017 16:11:15 GMT
< Content-Length: 1686
```

- turn on gzip in your Caddyfile and specify a root
- download the file using `curl curl -LO -v {URL}` 
- download the file using `curl curl -LO --compressed -v {URL}`

mbertschler:
This adds a Content-Length header when the file `myfile` is requested, gzip encoding is accepted and `myfile.gz` is present and served. I also added Content-Length checks to the fileserver tests.

For details see issue #1479

### DIFF
--- a/caddyhttp/staticfiles/fileserver.go
+++ b/caddyhttp/staticfiles/fileserver.go
@@ -201,6 +201,7 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
 
 		w.Header().Add("Vary", "Accept-Encoding")
 		w.Header().Set("Content-Encoding", encoding)
+		w.Header().Set("Content-Length", strconv.FormatInt(encodedFileInfo.Size(), 10))
 
 		defer f.Close()
 		break
--- a/caddyhttp/gzip/gzip.go
+++ b/caddyhttp/gzip/gzip.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"errors"
+
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
--- a/caddyhttp/staticfiles/fileserver_test.go
+++ b/caddyhttp/staticfiles/fileserver_test.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -76,15 +77,16 @@ func TestServeHTTP(t *testing.T) {
 	movedPermanently := "Moved Permanently"
 
 	tests := []struct {
-		url                 string
-		cleanedPath         string
-		acceptEncoding      string
-		expectedLocation    string
-		expectedStatus      int
-		expectedBodyContent string
-		expectedEtag        string
-		expectedVary        string
-		expectedEncoding    string
+		url                   string
+		cleanedPath           string
+		acceptEncoding        string
+		expectedLocation      string
+		expectedStatus        int
+		expectedBodyContent   string
+		expectedEtag          string
+		expectedVary          string
+		expectedEncoding      string
+		expectedContentLength string
 	}{
 		// Test 0 - access without any path
 		{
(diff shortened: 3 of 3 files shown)


caddyhttp/staticfiles/fileserver.go
```diff
@@ -201,6 +201,7 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request, name stri
 
 		w.Header().Add("Vary", "Accept-Encoding")
 		w.Header().Set("Content-Encoding", encoding)
+		w.Header().Set("Content-Length", strconv.FormatInt(encodedFileInfo.Size(), 10))
```
mholt: Isn't ServeContent() supposed to set the Content-Length for us?

mbertschler: Hmm interesting question. ServeContent() doesn't set Content-Length when Content-Encoding is set.

```go
if w.Header().Get("Content-Encoding") == "" {
	w.Header().Set("Content-Length", strconv.FormatInt(sendSize, 10))
}
```

I opened an issue in the go project: https://github.com/golang/go/issues/19420

mholt: Very interesting! I subscribed to the issue.

wendigo:
Nice one @mbertschler !

mbertschler:
Since golang/go#19420 was closed:

> It's never necessary to set Content-Length. If Caddy wants to set its own headers before ServeContent, then it can also set Content-Length. ServeContent is being conservative here and not interfering if the caller is trying to do its own thing.

I guess this pull request can be merged now?

mbertschler:
Thank you. Yes the test fails without this fix in these two cases:
- Test 18 - try to get pre-gzipped file.
- Test 19 - try to get pre-brotli encoded file.
```
$ go test github.com/mholt/caddy/caddyhttp/staticfiles
--- FAIL: TestServeHTTP (0.01s)
	fileserver_test.go:332: Test 18: Expected Content-Length header 17, found 
	fileserver_test.go:332: Test 19: Expected Content-Length header 16, found 
FAIL
FAIL	github.com/mholt/caddy/caddyhttp/staticfiles	0.015s
```

mholt:
Thank you again!

Would you like to be a collaborator on the project so that you can submit patches in a branch instead of a fork, and have power to help review pull requests, etc? I'll invite you and you're welcome to accept, as I always welcome more humanpower to the project.

mbertschler:
I already joined and hope that I can help. 😀

PR #1496, opened 2017-03-05
Replace use of Header to store Original URI for rewrite to Context

tobya:
This PR replaces the use of an internal request header

> Caddy-Rewrite-Original-URI

with `context`.  This should be merged before #1481 which will be updated to use this new code.

I'm not sure where to put a test if any.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -877,4 +877,7 @@ type CtxKey string
 // context.WithValue to access the original request URI that accompanied the
 // server request. The associated value will be of type string.
 const URLPathCtxKey CtxKey = "url_path"
+
+// URIxRewriteCtxKey is a context key used to store origional unrewritten
+// URI in context.WithValue
 const URIxRewriteCtxKey CtxKey = "caddy_rewrite_original_uri"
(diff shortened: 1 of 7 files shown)


caddy.go
```diff
@@ -877,3 +877,7 @@ type CtxKey string
 // context.WithValue to access the original request URI that accompanied the
 // server request. The associated value will be of type string.
 const URLPathCtxKey CtxKey = "url_path"
+
+// URIxRewriteCtxKey is a context key used to store original unrewritten
+// URI in context.WithValue
+const URIxRewriteCtxKey CtxKey = "caddy_rewrite_original_uri"
```
mholt: Oh, I guess I mentioned/pressed this point in the other issue, that this should probably go into the httpserver package. In fact, so should the const above this (URLPathCtxKey). These are specific to the HTTP server.

mholt:
Once the consts are moved, this looks good! Let me know if that move ends up being problematic.

tobya:
@mholt it should be fine

tobya:
All code will be merged with #1481 so I will close this.

PR #1497, opened 2017-03-05
browse: Use helper functions in staticfiles to redirect

ericdreeves:
Previously the browse package invoked staticfiles.Redirect when
redirecting clients who requested a directory but with a Request-URI
that did not contain a trailing '/'. staticfiles.Redirect only used a
relative URI. This change defers the decision of how to format the
Location header value to the helper methods in the staticfiles package.

### DIFF
--- a/caddyhttp/browse/browse.go
+++ b/caddyhttp/browse/browse.go
@@ -299,7 +299,7 @@ func (b Browse) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	// Browsing navigation gets messed up if browsing a directory
 	// that doesn't end in "/" (which it should, anyway)
 	if !strings.HasSuffix(r.URL.Path, "/") {
-		staticfiles.Redirect(w, r, r.URL.Path+"/", http.StatusTemporaryRedirect)
+		staticfiles.RedirectToDir(w, r)
 		return 0, nil
 	}
 
--- a/caddyhttp/browse/browse_test.go
+++ b/caddyhttp/browse/browse_test.go
@@ -13,7 +13,6 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 	"github.com/mholt/caddy/caddyhttp/staticfiles"
 )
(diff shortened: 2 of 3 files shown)


mholt:
Thanks Eric! Help me understand a little bit, is the relative redirect problematic?

ericdreeves:
Thanks for being patient while waiting for a response.

This change is based on our discussion [here](https://github.com/mholt/caddy/pull/1331#discussion_r103444896), as we wanted to make the redirect behavior between the browse and staticfiles packages consistent.

Regarding relative vs. absolute URI in the Location header, I made the change because the virtual site configuration provides the scheme and host components. Those two components are missing from most server requests, and `http.Redirect` requires both in order to use an absolute URI in the Location header. RFC2616 says that the Location header is an absolute URI, but we all know that isn't always the case. On the other hand, using a relative URI leaves it up to the client or web server to resolve the URI itself.

If you're uncomfortable with using absolute URI then we can change it back. Fortunately, we can change it in the helper functions such that both packages use the same URI format.

mholt:
@ericdreeves FYI, I'm investigating a breakage this introduced for a Caddyfile like:

```
:2015 {
   ...
}
```

Then requesting a directory without the trailing forward slash (which invokes RedirectToDir) causes a redirect to:

```
http://:2015/myfolder/
```

Which is not correct.

mholt:
@ericdreeves Or maybe this was introduced by #1331, I'm actually not sure. :) But, I think I'm gonna change this to a relative redirect. Let me know if you have any tips/comments on it.

mholt:
@ericdreeves Aha, it's introduced with https://github.com/mholt/caddy/commit/024833b2e3b0e08b39d40ceb1b012ed563671332#diff-e74f72185700e304b2047c4af2ad1a48R271 - I'm gonna remove that if statement. I don't think filling in those values is going to be good.

I didn't catch this earlier, but the reasoning [in this comment](https://github.com/mholt/caddy/pull/1331/files#r101875522) makes a false assumption:

> Since the virtual host's site address has both

A Caddyfile's site address can be simply `:2015` (port only), not both.

ericdreeves:
Thanks for finding the cause, @mholt, and for taking care of it. Let me know how I can help.

mholt:
@ericdreeves No problem :) I'm pushing a fix now. My lack of a sufficient test harness for the httpserver is to blame, ultimately. My bad.

mholt:
@ericdreeves By the way, I didn't mean to blame you; I was just informing you as I thought you might be interested and if I hit a wall you could help. Your contributions have made Caddy _better_ for sure!

ericdreeves:
@mholt I didn't take it that way. I appreciate you keeping me in the loop!

PR #1498, opened 2017-03-06
Fix for missing content-length header when using QUIC

ssut:
Since some chrome releases or quic clients are not sending the Content-Length
in the request headers, caddy should re-calculate content-length by the body and fix it.

This fixes a potential issue where FastCGI POST requests body empty when QUIC is enabled. (#1370)

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -6,6 +6,7 @@ package fastcgi
 import (
 	"errors"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -92,6 +93,7 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 
 			var resp *http.Response
 			contentLength, _ := strconv.Atoi(r.Header.Get("Content-Length"))
+			
 			switch r.Method {
 			case "HEAD":
 				resp, err = fcgiBackend.Head(env)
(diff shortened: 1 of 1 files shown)


caddyhttp/fastcgi/fastcgi.go
```diff
@@ -100,6 +102,20 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 			case "OPTIONS":
 				resp, err = fcgiBackend.Options(env)
 			default:
+				// since some chrome releases or quic clients are not sending the Content-Length
+				// in the request headers, caddy should re-calculate content-length by the body and fix it.
+				if contentLength == 0 {
+					var body []byte
+					if r.Body != nil {
+						body, _ = ioutil.ReadAll(r.Body)
```
mholt: Unfortunately, this line is a show-stopper right now. Not only are we ignoring the error, but we're reading the whole stream into memory over the network. This presents a number of problems, as you can imagine.

I think the whole reason the client is required by spec to send a Content-Length header is so that the server doesn't have to buffer the whole thing first; if the client isn't sending that, doesn't it mean the client is broken, and not Caddy?

ssut: I know it will create some problems with the memory management. even though it is not the most efficient way and also not often the case, I just noticed my stable chrome release does not send Content-Length header even though there is content inside the request so I just tried it with that way. (ye even if it's a special case, that's not good way though.) Well, I totally agree that content-length header should be included in every POST requests but It used "should" instead of "must", see content below.

```
A user agent SHOULD send a Content-Length in a request message when
   no Transfer-Encoding is sent and the request method defines a meaning
   for an enclosed payload body.  For example, a Content-Length header
   field is normally sent in a POST request even when the value is 0
   (indicating an empty payload body).  A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.
```

I haven't checked that it's working when only Transfer-Encoding is given, I think that caddy can (and should) handle this particular matter because it's impossible to ensure that clients behave as intended. or at least it could be better to send 411 Length Required if Content-Length is missing.

ssut:
now I figured out where the problem comes from, creating a patch for it.

caddyhttp/fastcgi/fastcgi.go
ssut: what i mentioned "I just noticed my stable chrome release does not send Content-Length header" above was wrong, it has been sending Content-Length header correctly but I didn't catch it (#1501). However, I think it would be great if caddy can handle the request which does not contain Content-Length header.

Issue #468, opened 2015-12-29
Caddy LE - SSL Only? No port 80


j-mcnally:
Hey Matt,

Im trying to get Caddy to work over 443 with lets encrypt, we have a different server running on port 80.

I saw something in the docs about tls-sni-01, anyway to force that?

mholt:
Not yet, but wait for the 0.8.1 release (or 0.8.2 for what you want - not sure yet how this'll play out) - we're working on that right now.

j-mcnally:
k, ive worked around it with some proxying from my other nginx instances.

pwFoo:
Is that fixed? because the issue ist closed. 
Tried to start 0.8.1 with port 443 only, but doesn't work. At the moment an apache is listening at port 80.

mholt:
Caddy needs port 80 to serve the http-01 challenge, which is the only challenge that works after the server has started (without implementing custom SNI which I'm not planning on doing). So if I were you I would relinquish port 80 to Caddy and proxy to Apache or something.

pwFoo:
Ok. Should be a quick solution with caddy https only. 
I have to move it to another host / ip. Thanks for your fast reponse! :)

seanhealy:
Is there any way to use the `tls-sni-01` challenge instead of `http-01`?

mholt:
No, because Caddy need to guarantee port 80 in order to not break reloads. So right now we still don't let the user disable http-01 challenge or the HTTP->HTTPS redirects.

heri16:
This ticket should be reopened. It should be possible now to disable port 80.

mholt:
@heri16 What do you mean? (Did I miss something?)

Place1:
I'd like to know what's happening with this issue as well. Having caddy work with letsencrypt using only 443 would be amazing as my ISP blocks port 80

mholt:
@Place1 There are issues #918 and #1293 that I think are both different ways to address this.

mholt:
@Place1 the patch I submitted doesn't disable the HTTP-01 challenge, but that can come in another patch.

Place1:
right, so if i've read these correctly they're about moving caddy to different ports and having something else proxy traffic to them.

my reason for getting caddy to to TLS-SNI-01 only is because my ISP blocks inbound connections to port 80 meaning I can't do an http challenge. Basically my server is 443 only and so the only challenge I can do (that isn't DNS) is TLS-SNI-01.

mholt:
[...]

PR #1500, opened 2017-03-07
httpserver: Flags to customize HTTP and HTTPS ports (including for ACME challenges)

### LINKED ISSUE #918: Add directive to change plain HTTP port

### LINKED ISSUE #1293: Support passing an alternate host:port for ACME verification
I would like to host non-HTTP TLS services on port 443 (specifically, OpenVPN, to get around silly firewalls). I've seen a past bug for this closed because Caddy wants to focus on HTTP, which is fair.

My alternative is to use a TLS SNI proxy, which inspects the requested hostname in the TLS handshake, and routes either to Caddy listening on localhost:1234, or to my other TLS-requiring services. The proxy itself doesn't terminate TLS, it just proxies bytes after picking the right backend, so this is mostly invisible to Caddy.

However, this currently breaks automatic TLS cert issuance, because Caddy uses the bind address when configuring the ACME client, and localhost doesn't work well over the internet. So, to make my setup work, I would need a way to tell Caddy use some other address for ACME issuance, and trust me to plumb things together.

Is this something you'd be open to adding as an option to the `tls` directive? Alternatively, if Caddy's stance has changed on doing TLS proxying in addition to HTTP handling... I'd be happy to discuss how to contribute that :).

mholt:
This commit removes _almost_ all instances of hard-coded ports 80 and
443 strings, and now allows the user to define what the HTTP and HTTPS
ports are by the `-http-port` and `-https-port` flags.

(One instance of "80" is still hard-coded in tls.go because it cannot
import httpserver to get access to the HTTP port variable. I don't
suspect this will be a problem in practice, but one workaround would be
to define an exported variable in the caddytls package and let the
httpserver package set it as well as its own HTTPPort variable.)

The port numbers required by the ACME challenges HTTP-01 and TLS-SNI-01
are hard-coded into the spec as ports 80 and 443 for good reasons,
but the big question is whether they necessarily need to be the HTTP
and HTTPS ports. Although the answer is probably no, they chose those
ports for convenience and widest compatibility/deployability. So this
commit also assumes that the "HTTP port" is necessarily the same port
on which to serve the HTTP-01 challenge, and the "HTTPS port" is
necessarily the same one on which to serve the TLS-SNI-01 challenge. In
other words, changing the HTTP and HTTPS ports also changes the ports
the challenges will be served on.

If you change the HTTP and HTTPS ports, you are responsible for
configuring your system to forward ports 80 and 443 properly.

Closes #918 and closes #1293. Also related: #468.

/cc @dncohen @danderson @HazCod @stp-ip @FiloSottile (phew)

### DIFF
--- a/caddyhttp/httpserver/context.go
+++ b/caddyhttp/httpserver/context.go
@@ -111,7 +111,7 @@ func (c Context) Port() (string, error) {
 	if err != nil {
 		if !strings.Contains(c.Req.Host, ":") {
 			// common with sites served on the default port 80
-			return "80", nil
+			return HTTPPort, nil
 		}
 		return "", err
 	}
--- a/caddyhttp/httpserver/https.go
+++ b/caddyhttp/httpserver/https.go
@@ -93,7 +93,7 @@ func enableAutoHTTPS(configs []*SiteConfig, loadCertificates bool) error {
 			cfg.TLS.Enabled &&
 			(!cfg.TLS.Manual || cfg.TLS.OnDemand) &&
 			cfg.Addr.Host != "localhost" {
-			cfg.Addr.Port = "443"
+			cfg.Addr.Port = HTTPSPort
 		}
 	}
 	return nil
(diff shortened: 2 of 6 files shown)


elcore:
> (One instance of "80" is still hard-coded in tls.go because it cannot
import httpserver to get access to the HTTP port variable. I don't
suspect this will be a problem in practice, but one workaround would be
to define an exported variable in the caddytls package and let the
httpserver package set it as well as its own HTTPPort variable.)

👍 sounds good

mholt:
Thanks both of you!

Skilgarriff:
Matt you are the best 👍

dncohen:
Huh.  I was about to comment that I tried, and haven't had any luck.  I'll try master branch now that I see it is merged.

HazCod:
@mholt sorry to bring this back up, but do you have any idea when you're going to bake a new release?

mholt:
@HazCod I'm thinking Thursday :wink: https://www.facebook.com/events/1413078512092363/

PR #1501, opened 2017-03-07
Fix for missing content-length header when using QUIC

ssut:
If request.ContentLength is set then it will be used instead of getting it from request.Header map since quic-go(lucas-clemente/quic-go@bb24be8) will not store (and pass) the Content-Length header using its header map.

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -91,7 +91,14 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 			fcgiBackend.SetSendTimeout(rule.SendTimeout)
 
 			var resp *http.Response
-			contentLength, _ := strconv.Atoi(r.Header.Get("Content-Length"))
+
+			var contentLength int64
+			// if ContentLength is already set
+			if r.ContentLength > 0 {
+				contentLength = r.ContentLength
+			} else {
+				contentLength, _ := strconv.ParseInt(r.Header.Get("Content-Length"), 10, 64)
+			}
 			switch r.Method {
 			case "HEAD":
 				resp, err = fcgiBackend.Head(env)
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -97,7 +97,7 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 			if r.ContentLength > 0 {
 				contentLength = r.ContentLength
 			} else {
-				contentLength, _ := strconv.ParseInt(r.Header.Get("Content-Length"), 10, 64)
+				contentLength, _ = strconv.ParseInt(r.Header.Get("Content-Length"), 10, 64)
 			}
 			switch r.Method {
 			case "HEAD":
--- a/caddyhttp/fastcgi/fcgiclient.go
+++ b/caddyhttp/fastcgi/fcgiclient.go
@@ -112,7 +112,7 @@ type Client interface {
 	Get(pair map[string]string) (response *http.Response, err error)
 	Head(pair map[string]string) (response *http.Response, err error)
 	Options(pairs map[string]string) (response *http.Response, err error)
-	Post(pairs map[string]string, method string, bodyType string, body io.Reader, contentLength int) (response *http.Response, err error)
+	Post(pairs map[string]string, method string, bodyType string, body io.Reader, contentLength int64) (response *http.Response, err error)
 	Close() error
 	StdErr() bytes.Buffer
 	SetReadTimeout(time.Duration) error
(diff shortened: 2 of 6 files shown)


caddyhttp/fastcgi/fastcgi.go
```diff
@@ -91,7 +91,14 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 			fcgiBackend.SetSendTimeout(rule.SendTimeout)
 
 			var resp *http.Response
-			contentLength, _ := strconv.Atoi(r.Header.Get("Content-Length"))
+
+			var contentLength int64
```
mholt: Any particular reason you chose int64 instead of int? This requires adding type conversions throughout the code below.

mholt:
Thanks @ssut - did you verify that this fixes an issue? Is the issue documented?

caddyhttp/fastcgi/fastcgi.go
ssut: I chose int64 because quic-go(https://github.com/lucas-clemente/quic-go/blob/master/h2quic/request.go#L49) uses it. I thought it would be better to use the same type with quic-go for overflow safe.

ssut:
@mholt I'm pretty sure this fixes the exact issue such like #1370, in this case https://github.com/mholt/caddy/issues/1483#issuecomment-283955017.

caddyhttp/fastcgi/fastcgi.go
mholt: Fair enough, not a big deal I guess.

mholt: Fair enough, not a big deal I guess.

mholt:
@ssut By "pretty sure" do you mean it doesn't work before your change, but does work after? If you're sure that's the case, then I'm okay with merging this.

ssut:
@mholt Yes indeed.

mholt:
Great, thank you @ssut!

caddyhttp/fastcgi/fastcgi.go
jaroslavtyc: @mholt @ssut Hi, sorry for writting it here, but it seems to me that this is the code responsible for my problem.
I am using php-fpm with Caddy v2.2.0 on localhost and v1.0.4 in production. Both versions just scrap my `Content-Length` header. It simply dissapears. Every other headers remains untouched.
That cause real problem with mobile Firefox on Android, as it considers the content length as zero, so downloaded file is empty.
Is there just a simple catch I do not see?
Thanks a lot for any clue!
```php
$file = __DIR__ . '/documents/CV_2020.pdf';
header('Content-type: application/pdf');
header('Last-Modified: ' . gmdate('D, d M Y H:i:s', filemtime($pdfFile)));
header('Content-Length: ' . filesize($pdfFile));
header(sprintf('Content-Disposition: attachment; filename="%s"', basename($pdfFile)));
readfile($pdfFile);
```

francislavoie: Please instead open a new issue using the following template. We can't really help you on a 4 year old pull request.


## Template

<pre><code>## 1. Environment

```
paste here
```


```
paste here
```


### 1c. Go version (if building Caddy from source; run `go version`)

```
paste here
```


## 2. Description

### 2a. What happens (briefly explain what is wrong)




### 2b. Why it's a bug (if it's not obvious)




### 2c. Log output

```
paste terminal output or logs here
```



### 2d. Workaround(s)




### 2e. Relevant links




</code></pre>

## Helpful tips

1. **Environment**: Please fill out your OS and Caddy versions, even if you don't think they are relevant. (They are _always_ relevant.) If you built Caddy from source, provide the commit SHA and specify your exact Go version.

2. **Description:** Describe at a high level what the bug is. What happens? Why is it a bug? Not all bugs are obvious, so convince readers that it's actually a bug.
	- **2c) Log output:** Paste terminal output and/or complete logs in a code block. [DO NOT REDACT INFORMATION except for credentials.](https://caddy.community/t/how-to-get-help-with-caddy-more-effectively/5222)
	- **2d) Workaround:** What are you doing to work around the problem in the meantime? This can help others who encounter the same problem, until we implement a fix.
	- **2e) Relevant links:** Please link to any related issues, pull requests, docs, and/or discussion. This can add crucial context to your report.

3. **Tutorial:** What are the _minimum required specific steps_ someone needs to take in order to experience the same bug? Your goal here is to make sure that anyone else can have the same experience with the bug as you do. You are writing a tutorial, so make sure to carry it out yourself before posting it. Please:
	- Start with an empty config. Add _only_ the lines/parameters that are _absolutely required_ to reproduce the bug.
	- Do not run Caddy inside containers.
	- Run Caddy manually in your terminal; do not use systemd or other init systems.
	- If making HTTP requests, avoid web browsers. Use a simpler HTTP client instead, like `curl`.
	- Do not redact any information from your config (except credentials). Domain names are public knowledge and [often necessary for quick resolution of an issue](https://caddy.community/t/how-to-get-help-with-caddy-more-effectively/5222)!
	- Note that ignoring this advice may result in delays, or even in your issue being closed. 😞 Only actionable issues are kept open, and if there is not enough information or clarity to reproduce the bug, then the report is not actionable.

Example of a tutorial:

<blockquote>Create a config file:

```
{ ... }
```

Open terminal and run Caddy:

```
$ caddy ...
```

Make an HTTP request:

```
$ curl ...
```

Notice that the result is ___ but it should be ___.</blockquote>

PR #1505, opened 2017-03-08
proxy: add root_ca parameter to configure upstream root CAs

tsusanka:
This adds the `root_ca` parameter to the proxy module to configure upstream root CAs.

Some of the example use cases might be to use internal roots for internal usage or to trim down roots when connecting to a service.

If the idea seems right, I can add docs to document this.

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -14,6 +14,7 @@ package proxy
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"io"
 	"net"
 	"net/http"
@@ -189,12 +190,8 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 		// if keepalive is equal to the default,
 		// just use default transport, to avoid creating
 		// a brand new transport
-		transport := &http.Transport{
-			Proxy:                 http.ProxyFromEnvironment,
-			Dial:                  defaultDialer.Dial,
-			TLSHandshakeTimeout:   10 * time.Second,
-			ExpectContinueTimeout: 1 * time.Second,
-		}
+		transport := createTransport()
+		transport.ExpectContinueTimeout = 1 * time.Second
 		if keepalive == 0 {
 			transport.DisableKeepAlives = true
 		} else {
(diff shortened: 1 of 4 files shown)


caddyhttp/proxy/reverseproxy.go
```diff
@@ -231,6 +232,26 @@ func (rp *ReverseProxy) UseInsecureTransport() {
...
 }
 
+func (rp *ReverseProxy) SetTransportCARoots(certPool *x509.CertPool) {
+	if rp.Transport == nil {
+		transport := &http.Transport{
+			Proxy:               http.ProxyFromEnvironment,
+			Dial:                defaultDialer.Dial,
+			TLSHandshakeTimeout: 10 * time.Second,
+			TLSClientConfig:     &tls.Config{RootCAs: certPool},
+		}
```
tsusanka: Makes sense. Something like this?

```go
func (rp *ReverseProxy) UseInsecureTransport() {
	rp.createHttp();
	rp.Transport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
}

func (rp *ReverseProxy) SetTransportCARoots(certPool *x509.CertPool) {
	rp.createHttp();
	rp.Transport.(*http.Transport).TLSClientConfig = &tls.Config{RootCAs: certPool}
}

func (rp *ReverseProxy) createHttp() {
	if rp.Transport == nil {
		transport := &http.Transport{
			Proxy:               http.ProxyFromEnvironment,
			Dial:                defaultDialer.Dial,
			TLSHandshakeTimeout: 10 * time.Second,
		}
		if httpserver.HTTP2 {
			http2.ConfigureTransport(transport)
		}
		rp.Transport = transport
	}
}
```

I've removed the

```go
if transport, ok := rp.Transport.(*http.Transport); ok {
```
block, since the `rp.Transport.(*http.Transport)` is now always created, is that right?

I think there is one more place where the function will be called. I'll replace it, if you find this correct.

tsusanka:
Thank you for your feedback. This was a result of a pair programming with @FiloSottile, who came up with the feature. We concluded we might throw in a PR directly to see what you think.

---------

I'd be happy to add tests, however I think I need a little bit of guidance. Is the `proxy_test.go` file the right place to put it? Should I use a mocked `.pem` file?

I'm afraid a brief howto is needed, sorry for that. I've tried to get inspired by the `TestReverseProxyMaxConnLimit` test, but this issue is quite different.

caddyhttp/proxy/reverseproxy.go
mholt: Yes, something like that. Although I was thinking of a function that just returned an http.Transport that you could further customize before setting on the ReverseProxy.Transport struct field.

> the `rp.Transport.(*http.Transport)` is now always created, is that right?

Hmm, that does seem to be true now. And if it wasn't, I suppose the reverse proxy would basically break, because we depend on that concrete type so much.

mholt:
Yeah, the test would involve at least connecting to an upstream (an httptest.Server probably) that has a self-signed certificate, that is trusted because it's a member of rootCertPool. The connection should succeed without disabling cert checking.

You'll want a test in upstream_test.go to assert that root_ca is being set properly, and another one in proxy_test.go to assert that it is being used properly. Hope that helps!

FiloSottile:
> (Maybe open an issue first next time?)

Yep, that would be my fault :)

tsusanka:
Sorry for a late response. I've done some of the modifications we've discussed and I've also tried to write the tests. I gotta be honest with you, I'm not sure I can finish it completely due to my limited Go knowledge.

Anyway, I've tried my best and I've squashed the commits and pushed it here. The tests (especially the `proxy_test.go`) unfortunately need your further assistance, I'm sorry for that. If you want you can give me another hints, but maybe it makes more sense if you finish up and I'll learn from that. But of course if you want ;). Sorry again.

mholt:
@tsusanka I'm likewise sorry, been busy getting ready for next week!

I appreciate the work you've put into this, I think it's pretty close to functional, but yes, the test will have to be done first for a change like this. I don't have the availability to assume the task and do it right now, either, as I'm preparing for the next release.

I'm closing the pull request, but open to having this feature implemented by someone who will pick it up in the future. Thank you again!

tsusanka:
Sure, I understand. I might also pick it up later myself. Thanks for your time.

mholt:
That'd be great. Thanks for understanding! (I hate to close PRs, tbh.)

PR #1511, opened 2017-03-12
Added Const for use of CtxKeys

tobya:
Replaced use of 

    caddy.CtxKey("remote_user")
and 

    caddy.CtxKey("mitm")

with constants

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -883,3 +883,9 @@ const URLPathCtxKey CtxKey = "url_path"
 // URIxRewriteCtxKey is a context key used to store original unrewritten
 // URI in context.WithValue
 const URIxRewriteCtxKey CtxKey = "caddy_rewrite_original_uri"
+
+// RemoteUserCtxKey is a context key used to store remote user for request
+const RemoteUserCtxKey CtxKey = "remote_user"
+
+// MitmCtxKey stores Mitm result
+const MitmCtxKey CtxKey = "mitm"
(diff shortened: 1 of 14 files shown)


caddy.go
```diff
@@ -883,3 +883,9 @@ const URLPathCtxKey CtxKey = "url_path"
 // URIxRewriteCtxKey is a context key used to store original unrewritten
 // URI in context.WithValue
 const URIxRewriteCtxKey CtxKey = "caddy_rewrite_original_uri"
+
+// RemoteUserCtxKey is a context key used to store remote user for request
+const RemoteUserCtxKey CtxKey = "remote_user"
+
+// MitmCtxKey stores Mitm result
+const MitmCtxKey CtxKey = "mitm"
```
mholt: While we're thinking about this, I have another idea... what if we move these const definitions into the `httpserver` package like we talked about before, except for URLPathCtxKey, which can be moved into the `staticfiles` package. Leave `type CtxKey string` here in the `caddy` package, since contexts can be used across different server types.

Also, I recommend using this form for multiple consts:

```go
const (
    // ...
)
```

tobya: @mholt  I have made those changes now.

tobya: actually I broke the tests - just  one moment.

tobya:
@mholt Should be good now.

mholt:
Thanks Toby. Looks like the tests are still failing though...

tobya:
:confounded:  @mholt should be good now.

mholt:
Fantastic! Thanks @tobya! :tada:

PR #1512, opened 2017-03-12
Fixed #1502: Proxying of unannounced trailers

### LINKED ISSUE #1502: Reverse proxying GRPC calls does not work
Caddy 0.9.5

I am trying to use caddy as a reverse proxy to a GRPC server.

See 7. below.

See 7. below.

Working GRPC calls.

No errors or log output of any kind. The grpc-server never receives anything from caddy.

The client only sees that the connection was closed by the (caddy) server before any sensible data could be sent or received.

I have sniffed the traffic using wireshark. The entire communication of one call consists of the following steps:
1. TCP SYN+ACK Client <-> Server
2. HTTP/2 SETTINGS package Client -> Server
3. TCP FIN+ACK Client <-> Server (FIN initiated by the sever)

It seems like caddy doesn't know what to do with a SETTINGS package or doesn't support something requested in that package and just terminates the connection.

I know my way around HTTP/1 a little but I don't really know anything about HTTP/2 or SETTINGS packages. The only thing I found out is that a SETTINGS package does not contain any request (GET/POST etc.) and thus no URL. So caddy can't know to which vserver/middleware it should route the package, thus the problem is most likely not *in* the proxy middleware but somewhere in the initial connection handling. I haven't looked at the caddy code yet though. I will try to come around to doing that soon.

I have researched GRPC proxing and caddy before and I found these related issues. I have created a new one though because I would consider them related but different enough from this to justify a new issue.

- #631 Apparently grpc proxing does not work out-of-the-box with go1.6
- #54 This seems to ask for a far larger amount of features. I don't need any fancy REST <-> GRPC converter, I simply want to have a caddy server on ports 80/443 and at the same time want to be able to have a grpc server on those ports.

I have created a little test suite in go, which mostly automates testing if GRPC works. See [this repo](https://github.com/targodan/caddyGrpcTest) for the source code, including a Caddyfile with which I tested this.

It provides a go package where you can run `go test -v` on and it will tell you if everything is working or not.

lhecker:
This is a rather controversial PR I believe, @mholt, since it contains some… more aggressive changes. Even though I believe that my changes are quite solid, I think we'll have to review this carefully… 🙂

### DIFF
--- a/caddyhttp/caddyhttp.go
+++ b/caddyhttp/caddyhttp.go
@@ -14,6 +14,7 @@ import (
 	_ "github.com/mholt/caddy/caddyhttp/fastcgi"
 	_ "github.com/mholt/caddy/caddyhttp/gzip"
 	_ "github.com/mholt/caddy/caddyhttp/header"
+	_ "github.com/mholt/caddy/caddyhttp/index"
 	_ "github.com/mholt/caddy/caddyhttp/internalsrv"
 	_ "github.com/mholt/caddy/caddyhttp/log"
 	_ "github.com/mholt/caddy/caddyhttp/markdown"
--- a/caddyhttp/caddyhttp_test.go
+++ b/caddyhttp/caddyhttp_test.go
@@ -11,7 +11,7 @@ import (
 // ensure that the standard plugins are in fact plugged in
 // and registered properly; this is a quick/naive way to do it.
 func TestStandardPlugins(t *testing.T) {
-	numStandardPlugins := 30 // importing caddyhttp plugs in this many plugins
+	numStandardPlugins := 31 // importing caddyhttp plugs in this many plugins
 	s := caddy.DescribePlugins()
 	if got, want := strings.Count(s, "\n"), numStandardPlugins+5; got != want {
 		t.Errorf("Expected all standard plugins to be plugged in, got:\n%s", s)
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -434,6 +434,7 @@ func RegisterDevDirective(name, before string) {
 var directives = []string{
 	// primitive actions that set up the fundamental vitals of each config
 	"root",
+	"index",
 	"bind",
 	"maxrequestbody", // TODO: 'limits'
 	"timeouts",
--- a/caddyhttp/index/index.go
+++ b/caddyhttp/index/index.go
@@ -1,0 +1,33 @@
+package index
+
+import (
+	"github.com/mholt/caddy"
+	"github.com/mholt/caddy/caddyhttp/staticfiles"
+)
+
+func init() {
+	caddy.RegisterPlugin("index", caddy.Plugin{
+		ServerType: "http",
+		Action:     setupIndex,
+	})
+}
+
+func setupIndex(c *caddy.Controller) error {
+	var index []string
+
+	for c.Next() {
+		args := c.RemainingArgs()
+
+		if len(args) == 0 {
+			return c.Errf("Expected at least one index")
+		}
+
+		for _, in := range args {
+			index = append(index, in)
+		}
+
+		staticfiles.IndexPages = index
+	}
+
+	return nil
+}
--- a/caddyhttp/index/index_test.go
+++ b/caddyhttp/index/index_test.go
@@ -1,0 +1,39 @@
+package index
+
+import (
+	"testing"
+
+	"github.com/mholt/caddy"
+	"github.com/mholt/caddy/caddyhttp/staticfiles"
+)
+
+func TestIndexIncompleteParams(t *testing.T) {
+	c := caddy.NewTestController("", "index")
+
+	err := setupIndex(c)
+	if err == nil {
+		t.Error("Expected an error, but didn't get one")
+	}
+}
+
+func TestIndex(t *testing.T) {
+	c := caddy.NewTestController("", "index a.html b.html c.html")
+
+	err := setupIndex(c)
+	if err != nil {
+		t.Errorf("Expected no errors, got: %v", err)
+	}
+
+	expectedIndex := []string{"a.html", "b.html", "c.html"}
+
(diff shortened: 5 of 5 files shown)


mholt:
Thanks @lhecker -- I'll take a look... do you want to explain what your changes do or why you made them so I know what to look for specifically in this PR?

lhecker:
@mholt:

037bc2d is the most controversial change IMO and introduces the `shallowCopyHeader`. When working with `copyHeader` I noticed that it was written in a rather inefficient way. The Get/Set/Add/Del methods on `http.Header` always transform their arguments into the canonicalized version, which means writing this is not good:

```go
for _, v := range vv {
	dst.Add(k, v)
}
```

Instead it should be:

```go
dst[k] = append([]string(nil), vv...)
```

This is especially true because `copyHeader` handles _only_ `http.Header` instances which means all keys (and values) are already in their canonical form! No need to use Get/Set/Add/Del here. 🙂

But afterwards I realized the next thing: All our uses of `copyHeader` only copy from headers which won't be modified later on. That means we can do a shallow copy and reduce the amount of allocations to almost zero. If that's too dangerous I'd be okay with reverting that and only doing the above `append()` optimization. But again: All current uses of `copyHeader` only copy from static/final headers to mutable ones.

<hr/>

9ad316f is a change I'd like you to look closely at. I happened to see that code and wondered for quite some time why it was written that complicated, instead of simply appending the new value.

<hr/>

75680bd is quite straight forward because it only splits up `ServeHTTP` into 2 sub-functions for better readability etc.

<hr/>

a2261e6 is the actually relevant change and adds the `shallowCopyTrailers` method. It's identical to `shallowCopyHeader` apart from not validating wether the headers are to be skipped (a small optimization). This method is then used in `serveGeneral` in place of the previous `copyHeader` method at the end of it. As a optimization for the normal case where all trailers are "announced" in the `Trailer:` header I added the `trailerKeyFunc` which allows one to switch between regularly adding trailers like it was before and forcing trailers by employing `http.TrailerPrefix`. The code contains comments to explain my reasoning there.

Tell me what you think, but please don't merge it yet: There are some minor things I'd like to force push beforehand.

mholt:
Thanks @lhecker, I hope to get to this soon. In addition to my review, I wonder if @nemothekid would have anything to say about the shallow header copy, as I believe he is the author of this part of the original proxy code (I'm mobile so I can't verify right now).

lhecker:
@mholt The original version comes directly from [Go's reverseproxy.go code](https://github.com/golang/go/blob/b71ed4edc6a5663caac434f9c2bea47dbc37db15/src/net/http/httputil/reverseproxy.go#L109-L115) and has since only been modified to add the `skipHeaders` feature.
It can't hurt if more people review this though.

mholt:
@lhecker I know you're busy moving right now, and I'm sorry for taking so long to review this that you were already moving before I got around to it!

Your explanation is _super_ helpful, thank you. I am really looking forward to going over these 4 commits more, but already as I begin to do so, I'm getting nervous because I have to either accept all 4 unrelated commits together (which might break something if merged prematurely), or reject all 4 unrelated commits even though only 1 or 2 need more work.

I guess I'm asking if you could please create four small PRs instead? The commits are unrelated so I would hate to hold up 3 just because 1 needs more time, or something like that. Smaller PRs are much easier to handle, when the commits are unrelated. :)

nemothekid:
w.r.t 9ad316f732c46cba4681da6d99f5ec1f1fe5a4a5,

The header for X-Forwarded-For was written that way because using `Header.Add` writes 2 headers in succession. For example currently the code does:

```
... Headers
X-Forwarded-For: 10.0.0.1, 10.0.0.2
... More Headers
```

While your change would result in:

```
... Headers
X-Forwarded-For: 10.0.0.1
X-Forwarded-For: 10.0.0.2
... More Headers
```

I don't think X-Forwarded-For is formally standardized, but other projects typically use the comma delimited version (https://en.wikipedia.org/wiki/X-Forwarded-For#Format).

nemothekid:
w.r.t. 037bc2d1a9f74a181258a35eadd5e53ec82c6159

You are probably correct - `reverseproxy.go` doesn't modify the headers again. I'm not sure about the rest of the Caddy codebase (are there other middlewares that modify headers?) or the golang stdlib (probably unlikely that they modify headers as well).

nemothekid:
Without 9ad316f it LGTM. However I'm also not 100% sure of 037bc2d - it should probably be in its own PR? Like you mentioned it came directly from the Go stdlib - however I'm not sure why Golang Authors went down this route if a shallow version is faster.

mholt:
Also, for the commits which are purely performance driven, it would be really great to see Benchmark functions in the tests that validate the performance improvements.

lhecker:
@nemothekid Thanks for the review. I'll soon split up this PR and leave out the change to the X-Forwarded-For header code.

The copyHeader code stems from httputil which isn't exactly the most well reviewed and optimized part of the Go stdlib. They probably went with full copies because they are simpler and safer.
I'll re-check that commit soon again and make sure that indeed no part of the proxy module modifies the headers though.
I'm in fact a bit scared of that commit myself because in better languages than Go you could just use a copy-on-write wrapper for the Headers and be done with it. But there's nothing like that that's possible with Go.

mholt:
@lhecker Just a heads up, the next release is in 6 days; if you can make separate PRs there's a good chance the little/easy ones will get in before then.

lhecker:
@mholt I'll update this PR with a force push to only include 75680bd and a2261e6 by tomorrow evening CET and I'll open a PR with 037bc2d later on.
Is that alright?

mholt:
@lhecker I'd still prefer one PR per commit in this case! Smaller PRs are fabulous. :) Makes review much easier and less... scary.

lhecker:
Sorry @mholt I misused git accidentally which caused GitHub to close this PR. And even though the fix_1502 branch has a commit it doesn't seem to be referenced by this PR anymore, which means we probably can't reopen this one.

PR #1513, opened 2017-03-13
Add the first policy  which sends the request to the first available host

pbeckmann:
Hey,

as promised an implementation of the load-balancing algorithm first.

Should be help fixing:
 - https://github.com/mholt/caddy/issues/1463
 - https://github.com/mholt/caddy/issues/1103

Greetings

### DIFF
--- a/caddyhttp/proxy/policy_test.go
+++ b/caddyhttp/proxy/policy_test.go
@@ -226,3 +226,20 @@ func TestIPHashPolicy(t *testing.T) {
 		t.Error("Expected ip hash policy host to be nil.")
 	}
 }
+
+func TestFirstPolicy(t *testing.T) {
+	pool := testPool()
+	firstPolicy := &First{}
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+
+	h := firstPolicy.Select(pool, req)
+	if h != pool[0] {
+		t.Error("Expected first policy host to be the first host.")
+	}
+
+	pool[0].Unhealthy = 1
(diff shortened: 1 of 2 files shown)


caddyhttp/proxy/policy_test.go
```diff
@@ -226,3 +226,20 @@ func TestIPHashPolicy(t *testing.T) {
...
+
+	h := firstPolicy.Select(pool, req)
+	if h != pool[0] {
+		t.Error("Expected first policy host to be the first host.")
+	}
+
+	pool[0].Unhealthy = 1
+	h = firstPolicy.Select(pool, req)
+	if h != pool[1] {
+		t.Error("Expected first policy host to be the first host.")
```
elcore: Isn't this the second host?

Everything else LGTM 👍

mholt: It's because for the test, the first host was set to Unhealthy, so the first available host would be the second one.

elcore: Thank you @mholt 😄 -- Even though that was clear to me 

I was just asking for the sake of clarification for the comment, sometimes the code says more than the comment :)

pbeckmann: Oh yes, i'll update the error message, so it's more clear 👍  Copy & Pasted it 😆

mholt: Oh, I see, Eldin was referring to the message, not the number.

mholt:
Thanks so much for your contribution, @pbeckmann !

PR #1522, opened 2017-03-17
Do not lowercase path in httpContext

### LINKED ISSUE #1517: Uppercase site address paths not working, lc alway performed
Further details https://forum.caddyserver.com/t/adresses-with-uppercase-path-does-not-work/1616

Caddy 0.9.5

Setup site addresses with Uppercase chars.

```text
:80/Transfer/ , :80/download/ {

	tls off

	timeouts {
	    read  30m 
	    write 30m
	    idle  30m
	}

	root /share	
}

```

/usr/bin/caddy --conf /etc/Caddyfile --log stdout
Tried also CASE_SENSITIVE_PATH=0

-  wget $host/Transfer/file.ext should be working

2017/03/15 15:06:29 http:///transfer/ 

Setup site adresses with paths, which has uppercase chars.

jeremiahmegel:
Closes #1517 

`httpContext` assumed that the entire address string could be case-insensitive and therefore lowercased it in order to detect duplicate entries. However, an address' path _is_ case-sensitive (e.g. `http://localhost:80/foo` and `http://localhost:80/Foo` are unique). This pull request modifies the lowercasing behavior to exclude the path.

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -344,7 +344,15 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 	// the URL path, so a request to example.com/foo/blog on the site
 	// defined as example.com/foo appears as /blog instead of /foo/blog.
 	if pathPrefix != "/" {
-		r.URL.Path = strings.TrimPrefix(r.URL.Path, pathPrefix)
+		if CaseSensitivePath {
+			r.URL.Path = strings.TrimPrefix(r.URL.Path, pathPrefix)
+		} else {
+			// The following is essentially a case-insensitive implementation
+			// of strings.TrimPrefix
+			if strings.HasPrefix(strings.ToLower(r.URL.Path), strings.ToLower(pathPrefix)) {
+				r.URL.Path = r.URL.Path[len(pathPrefix):]
+			}
+		}
 		if !strings.HasPrefix(r.URL.Path, "/") {
 			r.URL.Path = "/" + r.URL.Path
 		}
--- a/caddyhttp/httpserver/vhosttrie.go
+++ b/caddyhttp/httpserver/vhosttrie.go
@@ -33,6 +33,10 @@ func (t *vhostTrie) Insert(key string, site *SiteConfig) {
 // insertPath expects t to be a host node (not a root node),
 // and inserts site into the t according to remainingPath.
 func (t *vhostTrie) insertPath(remainingPath, originalPath string, site *SiteConfig) {
+	if !CaseSensitivePath {
+		remainingPath = strings.ToLower(remainingPath)
+		originalPath = strings.ToLower(originalPath)
+	}
 	if remainingPath == "" {
 		t.site = site
 		t.path = originalPath
@@ -103,6 +107,9 @@ func (t *vhostTrie) matchHost(host string) *vhostTrie {
 // remainingPath, and returns its node.
 func (t *vhostTrie) matchPath(remainingPath string) *vhostTrie {
 	var longestMatch *vhostTrie
+	if !CaseSensitivePath {
+		remainingPath = strings.ToLower(remainingPath)
+	}
 	for len(remainingPath) > 0 {
 		ch := string(remainingPath[0])
 		next, ok := t.edges[ch]
--- a/caddyhttp/httpserver/vhosttrie_test.go
+++ b/caddyhttp/httpserver/vhosttrie_test.go
@@ -16,6 +16,7 @@ func TestVHostTrie(t *testing.T) {
 		"example.com/foo/bar",
 		"*.example.com/test",
 	})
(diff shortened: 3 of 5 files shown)


tobya:
@jeremiahmegel  Thanks for the pull request.  This looks good.  Would you be able to add a test also?

abiosoft:
How about leveraging on `httpserver.CaseSensitivePath` to decide how this is handled.

https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/middleware.go#L147

jeremiahmegel:
Speaking of `httpserver.CaseSensitivePath`, it doesn't seem like that has any effect on the `staticfiles` server; [it relies on `http.Dir`](https://github.com/mholt/caddy/blob/36d2027493a65e76463d7075d37cd0a139dcb7f7/caddyhttp/httpserver/server.go#L117), which is case-sensitive (at least on my Linux system). Is this by design?

mholt:
@jeremiahmegel I'm not sure how to go about doing case-insensitive static file serving on Linux without trying each case combination for the whole path. Is there a way to do it with only one hit to the file system?

Thanks for this PR, I'm giving the whole situation more thought before going forward. :)

jeremiahmegel:
@mholt You could use [`ioutil.ReadDir()`](https://golang.org/pkg/io/ioutil/#ReadDir) and then case-insensitive search through the resulting slice for any filenames that match.

Of course, this would run in O(_n_ * _m_), where _n_ is the number of directories deep the ultimate file is from the webroot and _m_ is the average number of files in a directory.

The other problem would be handling conflicts, e.g. `abc.txt` vs. `ABC.txt` in the same directory.

wmark:
> […] http.Dir, which is case-sensitive (at least on my Linux system). Is this by design?

Yes.

This property is inherited from the backing filesystem(s).

The original issue emerged from converting *middleware paths* to simplify matching, and the remedy (duplicating entries) is something that fs implementers want to avoid.—Extending case-insensitive matching to static files is something, imho, that can remain out of scope of this PR.

caddyhttp/httpserver/server.go
```diff
@@ -344,7 +344,15 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
 	if pathPrefix != "/" {
-		r.URL.Path = strings.TrimPrefix(r.URL.Path, pathPrefix)
+		if CaseSensitivePath {
+			r.URL.Path = strings.TrimPrefix(r.URL.Path, pathPrefix)
+		} else {
+			// The following is essentially a case-insensitive implementation
+			// of strings.TrimPrefix
+			if strings.HasPrefix(strings.ToLower(r.URL.Path), strings.ToLower(pathPrefix)) {
+				r.URL.Path = r.URL.Path[len(pathPrefix):]
+			}
```
wmark: A comment why, how, and if `vhosts.Match` matched correctly would be helpful for someone reading this the first time.

caddyhttp/httpserver/vhosttrie.go
```diff
@@ -103,6 +107,9 @@ func (t *vhostTrie) matchHost(host string) *vhostTrie {
 // remainingPath, and returns its node.
 func (t *vhostTrie) matchPath(remainingPath string) *vhostTrie {
 	var longestMatch *vhostTrie
+	if !CaseSensitivePath {
+		remainingPath = strings.ToLower(remainingPath)
+	}
```
wmark: Remember to document that case sensitivity handling is guaranteed for *ASCII* and *Latin* only.

Else you'd convert it using `unicode.ToLower()`, and not doing so for performance reasons should be an intentional decision.

A *complete* handling (if ever wanted) would be *normalizing* → *convert to narrow* → *to lowercase*. ;-)

caddyhttp/httpserver/vhosttrie_test.go
```diff
@@ -16,6 +16,7 @@ func TestVHostTrie(t *testing.T) {
 		"example.com/foo/bar",
 		"*.example.com/test",
 	})
+	CaseSensitivePath = true
```
wmark: The tests are not run concurrently? (I don't know. Better check.)

jeremiahmegel:
Thanks for your feedback @wmark. I'm pretty busy for the next couple weeks, but I'll implement those changes when I get a chance.

> Extending case-insensitive matching to static files is something, imho, that can remain out of scope of this PR.

Might it be helpful to update the documentation to reflect this? In a separate pull request, of course.

mholt:
@jeremiahmegel I'm able to go over this again whenever you'd like another review. :)

jeremiahmegel:
@mholt Thanks! Still pretty busy as I'm finishing up college. I'll try to get to it soon! (In a couple weeks maybe?)

mholt:
Congrats, I did that not too long ago :P

mholt:
Closing, due to inactivity...

PR #1527, opened 2017-03-19
Sort Browse listing by dir first

tobya:
Following [this discussion on the caddy forum](https://forum.caddyserver.com/t/browse-function-display-folders-first/1613) this PR changes default ordering of a directory listing to show Directories first in alphabetical order then files in alphabetical order.  

This is the standard ordering that people expect to see in any gui (web or desktop based) display of files and folders.  The change simply modifies the main name sort.  When sorting on modified date, I have left it the same so it sorts files and folders together as a list.

I modified the test and added a folder to the test directory to show that the sorting works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ Caddyfile
 og_static/
 
 .vscode/
+
+caddyhttp/browse/temp*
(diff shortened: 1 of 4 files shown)


tobya:
@pbeckmann  Yes that is correct.

mholt:
Thanks Toby!

Issue #1534, opened 2017-03-25
browse: wrong alphabetic ordering


wmark:
With Caddy from HEAD (0.9.5),
I am *browsing* a directory with lots of files (binhost from Gentoo's portage tree, by the way) *alphabetically*.

### expected
`glibc` next to `glibc-2.25-r22.tgz`. (Exemplary.)

### experienced
Directory entries are scattered. Finding an entry becomes cumbersome. If it were for a few files I would just *spot* them and don't care about the order.

wmark:
Output obtained using `application/json` is in the wrong order, too, while it's supposed to be already sorted.

#1527 has introduced the breaking change.

If any special non-standard ordering is wanted for `text/html` (I'd veto'd it), it should not be done in that *view* only and not in the model, where it affects all views. ← Which would fix this issue, as would reverting the aforementioned commit.

tobya:
@wmark Thanks for the issue, I'll take a look

wmark:
You might wonder why sorting by *size* got a special treatment. For a consistent user experience we sort of standardized it.

This is due to filesystems interpreting *size* of a directory differently. Some set it to 0, some to (i)node-size, node-size + embedded content, some to the sum of the *actual* or *apparent* size of its contents (one level or ∞ levels deep). Go figure.

tobya:
@wmark I have pushed a PR that introduces a new sort type 'alpha' which will provide the original sort order ignoring file or dir type.  You simply need to specify sort in url

     curl --header "Accept: application/json"   http://courgette.kinchik.ie?sort=alpha

Does this solve your issue?

tobya:
@wmark I have pushed a PR that introduces a new sort type 'alpha' which will provide the original sort order ignoring file or dir type. You simply need to specify sort in url

    curl --header "Accept: application/json"   http://courgette.kinchik.ie?sort=alpha

Does this solve your issue? If possible I would rather not introduce differnt behavior based on a differnt view of the same data.

tobya:
I'll close this now - fixed in PR #1551

PR #1536, opened 2017-03-26
Added new alpha sort for Browse

tobya:
Issue #1534 raised the issue that json requests on browse folder were using new name sort introduced by #1527.

This PR introduces a new sort type 'alpha' which returns all files an folders in strict alphabetical order on name without taking into account if it is a dir or not. The sort can be set like the other sort types in the url.  

If a json request wishes to get all files in alpha order by curl can do

    curl --header "Accept: application/json"   https://courgette.kinchik.ie?sort=alpha

### DIFF
--- a/caddyhttp/browse/browse.go
+++ b/caddyhttp/browse/browse.go
@@ -137,7 +137,14 @@ func (l byName) Swap(i, j int) { l.Items[i], l.Items[j] = l.Items[j], l.Items[i]
 
 // Treat upper and lower case equally
 func (l byName) Less(i, j int) bool {
-	return strings.ToLower(l.Items[i].Name) < strings.ToLower(l.Items[j].Name)
+
+	// if both are dir or file sort normally
+	if l.Items[i].IsDir == l.Items[j].IsDir {
+		return strings.ToLower(l.Items[i].Name) < strings.ToLower(l.Items[j].Name)
+	} else {
+		// always sort dir ahead of file
+		return l.Items[i].IsDir
+	}
 }
 
 // By Size
--- a/caddyhttp/browse/browse.go
+++ b/caddyhttp/browse/browse.go
@@ -21,10 +21,10 @@ import (
 )
 
 const (
+	sortByName         = "name"
 	sortByNameDirFirst = "namedirfirst"
 	sortBySize         = "size"
 	sortByTime         = "time"
-	sortByName         = "name"
 )
 
 // Browse is an http.Handler that can show a file listing when
@@ -128,10 +128,19 @@ func (fi FileInfo) HumanModTime(format string) string {
 }
 
 // Implement sorting for Listing
+type byName Listing
 type byNameDirFirst Listing
 type bySize Listing
 type byTime Listing
-type byName Listing
+
+// By Name
+func (l byName) Len() int      { return len(l.Items) }
+func (l byName) Swap(i, j int) { l.Items[i], l.Items[j] = l.Items[j], l.Items[i] }
+
+// Treat upper and lower case equally
+func (l byName) Less(i, j int) bool {
+	return strings.ToLower(l.Items[i].Name) < strings.ToLower(l.Items[j].Name)
+}
 
 // By Name
 func (l byNameDirFirst) Len() int      { return len(l.Items) }
@@ -170,43 +179,34 @@ func (l byTime) Len() int           { return len(l.Items) }
 func (l byTime) Swap(i, j int)      { l.Items[i], l.Items[j] = l.Items[j], l.Items[i] }
 func (l byTime) Less(i, j int) bool { return l.Items[i].ModTime.Before(l.Items[j].ModTime) }
 
-// By Name
-func (l byName) Len() int      { return len(l.Items) }
-func (l byName) Swap(i, j int) { l.Items[i], l.Items[j] = l.Items[j], l.Items[i] }
-
-// Treat upper and lower case equally
-func (l byName) Less(i, j int) bool {
-	return strings.ToLower(l.Items[i].Name) < strings.ToLower(l.Items[j].Name)
-}
-
 // Add sorting method to "Listing"
 // it will apply what's in ".Sort" and ".Order"
 func (l Listing) applySort() {
 	// Check '.Order' to know how to sort
 	if l.Order == "desc" {
 		switch l.Sort {
+		case sortByName:
+			sort.Sort(sort.Reverse(byName(l)))
 		case sortByNameDirFirst:
 			sort.Sort(sort.Reverse(byNameDirFirst(l)))
 		case sortBySize:
 			sort.Sort(sort.Reverse(bySize(l)))
 		case sortByTime:
 			sort.Sort(sort.Reverse(byTime(l)))
-		case sortByName:
-			sort.Sort(sort.Reverse(byName(l)))
 		default:
 			// If not one of the above, do nothing
 			return
 		}
 	} else { // If we had more Orderings we could add them here
 		switch l.Sort {
+		case sortByName:
+			sort.Sort(byName(l))
 		case sortByNameDirFirst:
 			sort.Sort(byNameDirFirst(l))
 		case sortBySize:
 			sort.Sort(bySize(l))
 		case sortByTime:
 			sort.Sort(byTime(l))
-		case sortByName:
-			sort.Sort(byName(l))
 		default:
 			// If not one of the above, do nothing
 			return
@@ -364,7 +364,7 @@ func (b Browse) handleSortOrder(w http.ResponseWriter, r *http.Request, scope st
 		if sortCookie, sortErr := r.Cookie("sort"); sortErr == nil {
 			sort = sortCookie.Value
 		}
-	case sortByNameDirFirst, sortBySize, sortByTime, sortByName:
+	case sortByName, sortByNameDirFirst, sortBySize, sortByTime:
 		http.SetCookie(w, &http.Cookie{Name: "sort", Value: sort, Path: scope, Secure: r.TLS != nil})
 	}
 
--- a/caddyhttp/browse/browse.go
+++ b/caddyhttp/browse/browse.go
@@ -21,10 +21,10 @@ import (
 )
 
 const (
-	sortByName  = "name"
-	sortBySize  = "size"
-	sortByTime  = "time"
-	sortByAlpha = "alpha"
+	sortByNameDirFirst = "namedirfirst"
+	sortBySize         = "size"
+	sortByTime         = "time"
+	sortByName         = "name"
 )
 
 // Browse is an http.Handler that can show a file listing when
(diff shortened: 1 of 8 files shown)


tobya:
hmm.  too soon.

mholt:
I will await a comment from @wmark before merging this, but this LGTM.

wmark:
So the behaviour know to professionals for 28 years+ – the old behaviour is under a new flag, and the new one under the old flag…. I don't think this is right. Whoever upgrades needs to change templates and scripts.

Use the novel order with the new flag perhaps? (Remember to set it in the template.) That way we'd have a win-win.

tobya:
I disagree that the previous behaviour was the standard for a UI.  All file browsing UIs (which a web browser is) show Directories first then files.  

@mholt do you have an opinion on this?  

If we are not happy with this change, there are a number of other possibilities

 - We could provide different sort orders for `application/text` or `application/json`?
 -  we could provide a default sort for the browse plugin eg

     localhost:2015
     root c:\docs\
     browse / alpha

this would browse with each directory defaulting to alpha sorting?  Is this desirable?  
- Revert to prior to  #1527

wmark:
If you set the corresponding column in *browse's* template to accordingly, as per the suggestion, you'd have the desired outcome in the browser, nothing would break, and other mime types were unaffected.

mholt:
When navigating through a "deep" tree of files (more than 2 or 3 folders), it can be quicker to have all the folders near the top because it reduces the amount of scrolling you have to do. With the current proposed change, as I understand it, you'll only have to potentially scroll a lot on the last page, rather than on every page before it.

Personally, _I'm_ okay with the change. I don't think it'll be the end of the world. I do believe most users probably use the visual interface in the browser and the consistency across the different MIME types is desirable.

I do like Mark's suggestion too, that alpha sorting be an option in the browse template somehow, so that users can still choose pure alphabetic ordering if they want.

wmark:
To clarify, I suggested:

 * keep the old `name` as it was, no new algorithm under the old flag
 * the new dir-first order can remain the new default for `text/html`
 * just rename it (for example, to `name2`), — fans of `O(2 log n)` seeking still will be happy

So, a new name and used in the shipped `text/html` template = used by default when clicking on the column header.

mholt:
How about we add a new "sort arrow" to the left of the "Name" column heading (in the same column, but with a little right margin) that sorts by "Name" ascending by default, with the directories first, then the files below them. And we change the "Name" sort to be the alpha sort. I think that would make the most sense for most people.

wmark:
IMHO, every variant is fine as long as the behaviour for `name` doesn't change.

tobya:
@wmark I presume that as well as name behaviour not changing, you also require the default behaviour (eg without specifying any sort order in the url) to also remain the same?

mholt:
Why is the current behavior for "name" so important to not change? It seems very strange to have "alpha" not be the truly alphabetic order.

wmark:
Then rename it from `alpha` to `name2`. I don't understand the resistance.

Existing templates will remain working as expected with the `name` of ≤0.9.5, and output in json for scripts will remain sorted the expected way.

I don't really see any obstacle for implementing a new behaviour under a new name. There's no downsides to that. Forcing the *old behaviour* to a *new flag*, and dropping in *a new order* for the *old flag* as per the hero-edit that created facts while an discussion was still open — not backward compatible for what benefit exactly? The burden is on justifying a change, not the other way around.

tobya:
The issue here is the *default* for the gui view.  Whether we call it `name` or `alpha` or `name2` is not the issue, the issue is the default sort order when no sort order is specified.

The previous behaviour when no sort order was specified was to sort by alphabetical order ignoring Dir & Folder differences (now `alpha`), this was the sort order for both html and json.  

There is no problem not changing the behaviour of the name sort, but fundamentally it all comes down to the default sort when sort  is *not* specified.  In this instance if we create name2 we would still default to the new order name2 when none is specified, in order to get the new sorting for the GUI view.

I assume this would still cause you a problem @wmark ?

Do you have a particular use case that you are currently using this feature for, that is broken by the change to default behaviour (without specifying a sort order explicitly)?

The suggestion by @mholt I think still has the same issue which is that if we simply add the sorting to the template then the default behaviour is still to order by name2 before the link is clicked.

Considering the resistance to the change from @wmark I think an argument can be made for the 'default' behaviour being name (  Dir first) for html and (alpha ) for json response.  The underlying implementation of name would still change if explicitly specified in the url.

Would this be acceptable?

wmark:
Then change the default template, too.

The change will not be a breaking one iff output of `https://…/…/?sort=name` `×{asc,desc}` remains the same going from 0.9.5 to a later version. (This is orthogonal to Matt's suggestion.)

tobya:
broken tests, fixing

tobya:
This update

- name sort has been reverted to previous behaviour
- Behaviour is now the same for urls where sort=name is included such as  `https://…/…/?sort=name ×{asc,desc}`
- New sort created namedirfirst that sorts dir then file in alpha order
- Default order when no sort defined is namedirfirst

caddyhttp/browse/browse.go
```diff
@@ -168,30 +170,43 @@ func (l byTime) Len() int           { return len(l.Items) }
...
 // Add sorting method to "Listing"
 // it will apply what's in ".Sort" and ".Order"
 func (l Listing) applySort() {
 	// Check '.Order' to know how to sort
 	if l.Order == "desc" {
 		switch l.Sort {
-		case sortByName:
-			sort.Sort(sort.Reverse(byName(l)))
+		case sortByNameDirFirst:
+			sort.Sort(sort.Reverse(byNameDirFirst(l)))
```
wmark: Leave `case sortByName` where it was. Don't modify the lines here. Just at new code above or below.

This is due to how source code control software, like **git**, works and helps retain change history.

caddyhttp/browse/browse.go
```diff
@@ -345,11 +360,11 @@ func (b Browse) handleSortOrder(w http.ResponseWriter, r *http.Request, scope st
 	// If the query 'sort' or 'order' is empty, use defaults or any values previously saved in Cookies
 	switch sort {
 	case "":
-		sort = sortByName
+		sort = sortByNameDirFirst
 		if sortCookie, sortErr := r.Cookie("sort"); sortErr == nil {
 			sort = sortCookie.Value
 		}
-	case sortByName, sortBySize, sortByTime:
+	case sortByNameDirFirst, sortBySize, sortByTime, sortByName:
```
wmark: Again, leave `sortByName` where it was. Just append or prepend.

caddyhttp/browse/browse_test.go
```diff
@@ -70,6 +70,13 @@ func TestSort(t *testing.T) {
 		t.Errorf("The listing isn't time sorted: %v", listing.Items)
 	}
 
+	// sort by name dir first
+	listing.Sort = "namedirfirst"
+	listing.applySort()
+	if !sort.IsSorted(byNameDirFirst(listing)) {
+		t.Errorf("The listing isn't namedirfirst sorted: %v", listing.Items)
+	}
+
```
wmark: Those tests are not needed, IMHO. There's no formal *contract* as to what order is the default. Adding code = adding complexity.

But there's admittedly several families of thinking regarding what and how granular to test.

caddyhttp/browse/setup.go
```diff
@@ -342,12 +342,20 @@ footer {
...
+							<a href="?sort=name&order=desc{{if ne 0 .ItemsLimitedTo}}&limit={{.ItemsLimitedTo}}{{end}}">(a-z) <svg width="1em" height=".4em" version="1.1" viewBox="0 0 12.922194 6.0358899"><use xlink:href="#up-arrow"></use></svg></a>
 							{{- else if and (eq .Sort "name") (ne .Order "asc")}}
-							<a href="?sort=name&order=asc{{if ne 0 .ItemsLimitedTo}}&limit={{.ItemsLimitedTo}}{{end}}">Name <svg width="1em" height=".4em" version="1.1" viewBox="0 0 12.922194 6.0358899"><use xlink:href="#down-arrow"></use></svg></a>
+							<a href="?sort=name&order=asc{{if ne 0 .ItemsLimitedTo}}&limit={{.ItemsLimitedTo}}{{end}}">(a-z) <svg width="1em" height=".4em" version="1.1" viewBox="0 0 12.922194 6.0358899"><use xlink:href="#down-arrow"></use></svg></a>
 							{{- else}}
-							<a href="?sort=name&order=asc{{if ne 0 .ItemsLimitedTo}}&limit={{.ItemsLimitedTo}}{{end}}">Name</a>
+							<a href="?sort=name&order=asc{{if ne 0 .ItemsLimitedTo}}&limit={{.ItemsLimitedTo}}{{end}}">(a-z)</a>
```
wmark: This addresses Matt's suggestion, I guess. I don't mind skipping this and just changing `name` → `namedirfirst` here to keep the template clean and minimal.

caddyhttp/browse/browse.go
tobya: I'll modify to leave sortByName where it is.

tobya:
I think these changes address @wmark concerns.  I think the commits will get squashed when merging with master.

caddyhttp/browse/browse.go
```diff
@@ -137,6 +139,15 @@ func (l byName) Swap(i, j int) { l.Items[i], l.Items[j] = l.Items[j], l.Items[i]
 
 // Treat upper and lower case equally
 func (l byName) Less(i, j int) bool {
+	return strings.ToLower(l.Items[i].Name) < strings.ToLower(l.Items[j].Name)
+}
```
wmark: This still appears as having been modified, although it's restored code.

Please start your PR with the result of `git revert` as first commit.

caddyhttp/browse/browse_test.go
```diff
@@ -257,6 +272,9 @@ func TestBrowseJson(t *testing.T) {
 			Mode:    f.Mode(),
 		})
 	}
+
+	// Test that sort=name returns correct listing, as per filesystem
```
wmark: That actually not *as per filesystem*, which usually doesn't guarantee any order. Remove everything after the comma.

tobya:
#1551 has the necessary changes submitted as a clean PR with a revert of the origional commit.

wmark:
Awesome!

By the way, you can overwrite existing PRs as follows, including flattening/squashing:
```bash
# save the most latest version of the file(s)
cp -a http/browse/{a.go,b.go,c/d.html} /tmp
#git rebase -i HEAD~6  # fixup, excluding the first commit
#git format-patch …    # in case you want a second backup, or like patches more

git checkout mholt/master # or any other branch, just leave the one to be deleted
git checkout -d browsesort4

git checkout -b browsesort4
git revert c0c0c0c0
cp -a /tmp/… http/browse/
# or: <00*patch git apply
git add -A
git commit
git push tobya +browsesort4
```

mholt:
Thanks, Toby and Mark!

PR #1537, opened 2017-03-26
Introducing Event Hooks Plugins

hacdias:
The goal of this is to create plugins that can register Event Hooks and process information according to the type of event that is fired.

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -1,8 +1,8 @@
 package caddy
 
 import (
-	"errors"
 	"fmt"
+	"log"
 	"net"
 	"sort"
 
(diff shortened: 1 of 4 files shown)


plugins.go
```diff
@@ -228,6 +204,44 @@ func RegisterPlugin(name string, plugin Plugin) {
...
+type EventHook func(eventType EventType) error
+
+// RegisterEventHook plugs in hook. All the hooks should register themselves
+// and they must have a name.
+func RegisterEventHook(name string, hook EventHook) {
+	if name == "" {
+		panic("event hook must have a name")
+	}
+	if _, dup := eventHooks[name]; dup {
+		panic("hook named " + name + " already registered")
```
mholt: This reminds me, though is not relevant to this PR directly, the names of these plugins should be prefixed with something, like "hook." or "events." so they don't clutter a global namespace as much. What do you think? (Again, I don't think this involves a code change here, but it will be needed in the dev portal I think.)

plugins.go
```diff
@@ -228,6 +204,44 @@ func RegisterPlugin(name string, plugin Plugin) {
 	plugins[plugin.ServerType][name] = plugin
 }
 
+// EventType represents the event type to be used with event hooks
```
mholt: Actually, can we call it "EventName"?

plugins.go
hacdias: I don't know. A different section in the Plugins (like "Directives/Middleware", "Server Types") or using a prefix. 😄

hacdias:
Changes made! @mholt

PR #1540, opened 2017-03-28
Add Event Hook plugins to DescribePlugins function

hacdias:


plugins.go
```diff
@@ -65,6 +65,11 @@ func DescribePlugins() string {
 			others = append(others, s)
 		}
 	}
+
+	for hookPlugin := range eventHooks {
+		others = append(others, hookPlugin)
+	}
+
```
mholt: Move this above the "Let's alphabetize these" comment and give it their own section: "Event hooks" - you can just print the name as you iterate them. Thanks!

hacdias: Should they be prefixed with "hook."?

mholt: Yes. :) That would be good.

hacdias:
Done!

mholt:
Spoke too soon. I didn't notice the tests are failing. How about only printing event hooks (including the header) if there are at least one; make sure the test passes.

PR #1541, opened 2017-03-28
Make CADDYPATH equal across init scripts

lucafavatella:
See also https://github.com/mholt/caddy/pull/1272#issue-191256343

### DIFF
--- a/dist/init/linux-upstart/caddy.conf
+++ b/dist/init/linux-upstart/caddy.conf
@@ -14,12 +14,12 @@ respawn limit 10 5
 reload signal SIGUSR1
 
 # Let's Encrypt certificates will be written to this directory.
-env CADDYPATH=/etc/caddy
+env CADDYPATH=/etc/ssl/caddy
 
 limit nofile 1048576 1048576
 
 script
-        cd /etc/caddy
+        cd /etc/ssl/caddy
         rootdir="$(mktemp -d -t "caddy-run.XXXXXX")"
         exec /usr/local/bin/caddy -agree -log=stdout -conf=/etc/caddy/Caddyfile -root=$rootdir
 end script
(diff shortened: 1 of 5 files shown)


dist/init/linux-upstart/caddy.conf
```diff
@@ -14,12 +14,12 @@ respawn limit 10 5
 reload signal SIGUSR1
 
 # Let's Encrypt certificates will be written to this directory.
-env CADDYPATH=/etc/caddy
+env CADDYPATH=/etc/ssl/caddy
```
mholt: I'm just an outsider (I don't use these init scripts), but `/etc/ssl/caddy` is a weird path. Why not just `/etc/caddy`? SSL is an old technology anyways.

lucafavatella: As far as content is concerned, I see that `/etc/caddy` contains only `Caddyfile` so I do not expect conflicts. The main problem is permissions:
* systemd has specific [instructions](https://github.com/mholt/caddy/blob/6cabc9bfe3dc5f06c49aa8fb58ac19a09a4f42e9/dist/init/linux-systemd/README.md#instructions) and [code](https://github.com/mholt/caddy/blob/6cabc9bfe3dc5f06c49aa8fb58ac19a09a4f42e9/dist/init/linux-systemd/caddy.service#L34-L38) for distinct permissions of the two folders;
* The other init systems have notes on permissions, but not as specific as systemd.

Having the two folders distinct rather than one-subfolder-than-the-other shall make permission setting simpler.

elcore: How about /etc/tls/caddy (/etc/caddy/certs)?

mholt: So /etc/caddy is actually the HTTP site root? And you need a separate Caddypath which stores OCSP staples and certs... I guess the name 'ssl' is fine. 🤷‍♂️

mholt:
Thanks for your contribution, @lucafavatella !

PR #1542, opened 2017-03-28
Add basicauth {user} placeholder for logs

tobya:
This PR adds in {user} placeholder.  The value is provided by basicauth.

If `{user}` does not exist a `-` is output to conform with the 'common log' format 

I will submit a PR against caddyserver.com with updated docs.


jamesboswell:
My testing shows this PR fixes my issue #1543

caddyhttp/basicauth/basicauth.go
```diff
@@ -63,7 +63,7 @@ func (a BasicAuth) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 
 			// let upstream middleware (e.g. fastcgi and cgi) know about authenticated
 			// user; this replaces the request with a wrapped instance
-			r = r.WithContext(context.WithValue(r.Context(),
+			*r = *r.WithContext(context.WithValue(r.Context(),
```
mholt: Why is this dereference needed?

caddyhttp/httpserver/replacer.go
```diff
@@ -309,6 +309,12 @@ func (r *replacer) getSubstitution(key string) string {
 	case "{dir}":
 		dir, _ := path.Split(r.request.URL.Path)
 		return dir
+	case "{user}":
+		user, _ := r.request.Context().Value(RemoteUserCtxKey).(string)
+		if user == "" {
+			user = "-"
```
mholt: I don't think we need an empty value here; any middleware that wants a default value for empty placeholders will pass in what that value should be.

jamesboswell: I disagree @mholt.  `"-"` should always be the default

What @tobya and I have been discussing on the forums is fixing the **username** logging *and* keeping  the {common} and {combined} log formats compliant with the [Common Log Format ](https://en.wikipedia.org/wiki/Common_Log_Format) which is a common format established by NCSA/W3C for many many years.

The standard says: 
> If a field is unused in a particular entry dash "-" marks the omitted field. 

Why stick with Common Log Format (CLF) for Caddy?   Tooling and log analysis.  Many, many tools such as Logstash/Elasticsearch, AWStats, Splunk, etc.   Most of them are out of the box ready to understand CLF

Most of them use this grok pattern or something similar
```
COMMONAPACHELOG %{IPORHOST:clientip} %{USER:ident} %{USER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
COMBINEDAPACHELOG %{COMMONAPACHELOG} %{QS:referrer} %{QS:agent}
```
These lines setup a default `"-"` and allows plugins that perform authentication to pass the username into the Caddy log

mholt: @jamesboswell `"-"` **is** the default empty value.

See the third argument to NewReplacer:

https://github.com/mholt/caddy/blob/6cabc9bfe3dc5f06c49aa8fb58ac19a09a4f42e9/caddyhttp/httpserver/replacer.go#L90

And the log always passes in `"-"`:

https://github.com/mholt/caddy/blob/6cabc9bfe3dc5f06c49aa8fb58ac19a09a4f42e9/caddyhttp/log/log.go#L34

Consequently, the line highlighted in this PR will break that argument if some middleware ever wants to use something other than "-" (which some do, some just want an empty string because they aren't always using this for logging).

jamesboswell: @mholt  doh!  well I appreciate your passion as well 😄,  and I was all fired up! , but I see what you mean.   Yeah lines 314-315 can go.

caddyhttp/basicauth/basicauth.go
jamesboswell: I should let Toby answer, but I wanted to better understand myself, so pardon me @tobya  😬 

@mholt  I believe its because "under the hood" it's still [passing by value](https://groups.google.com/forum/m/#!topic/golang-nuts/INedfATw74A) [Rob Pike] and not by reference.   

`WithContext` is [returning  a new `http.Request` (r2)](https://golang.org/src/net/http/request.go?s=11679:11738#L317).   So we need to dereference to modify the original `http.Request` that r is pointing to,  otherwise you get an entirely new `r` and the new context key doesn't get passed down the chain.   My own testing shows this to be the case.

I have output dumps of this if you need to see it.    My research shows the only way to avoid this is to go all the way back up the chain to where the first ServeHTTP is called,  or just do the dereference within the middleware

tobya: sort of basically what @jamesboswell said.  Simply put if I don't dereference r when calling `withContext` the context value is not available to the replacer.

caddyhttp/httpserver/replacer.go
tobya: @mholt this is fixed.  If user is empty I do not return a value and `-` is written to log.

caddyhttp/basicauth/basicauth.go
jamesboswell: @mholt @tobya just an FYI,  For the past 9 days, I've been running in production a build of Caddy with this PR cherry-picked and my own auth module that [closely copies the `basicauth` ](https://github.com/jamesboswell/caddy-radius/compare/caddyPR1542) changes in this PR.

Caddyfile is now using the updated `{combined}` log format instead of my previous custom format.

So far everything looks great in regards to logging `{user}` in access logs.

mholt: @jamesboswell Thanks for the explanation - but it doesn't seem right. The method returns a _copy_ of the original request with the new value in the context, hence the re-assignment of `r`.

@tobya I do remember that issue, but that was different because we had to modify the Request in a function outside of ServeHTTP; in this function, there's no use of a replacer, so why is the dereference necessary?

I'm worried that doing the dereference could easily introduce race conditions in the event a handler ever starts a goroutine (I think some already do) and needs to use the context...

tobya: OK. However I think I need a bit of help, because I can't figure out what is going on.  

If I don't dereference `r`, then the context value is not available to the replacer.  Perhaps when someone has time they could try to figure out where the issue is.  I have even tried changing the context key just incase it was being overwritten.

mholt: Ohhhh, is this because "the replacer" is "the replacer" that is being created by the logger in an earlier handler, and _that_ replacer is the one that holds a reference to the first `r`?

Yuck. Every time we do this it means we can't use the context for cancelling goroutines that may need it because it could be racey to pass it into a goroutine that accesses the context concurrently. (Maybe middlewares that need to use the context values can pull out what they need before the goroutine starts and pass them into the goroutine separately?)

jamesboswell: @mholt  @tobya  
The problem is that within the scope of `basicauth`  the pointer to `http.Request` is being changed as `WithContext` allocates a new `r (r2)`.   Thus the dereference is needed.

if we're concerned about race issues in goroutines,  then I think we need to capture the Authorization header username before the middleware chain, to have `{user}` placeholder available

Perhaps something like this [here before r](https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/server.go#L295) is passed down to middlwareChain?   Then `{user}` is available to all downstream middleware
Quck testing looks good so far,  can look into it more later.
```diff
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -292,6 +292,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {

        w.Header().Set("Server", "Caddy")
        c := context.WithValue(r.Context(), staticfiles.URLPathCtxKey, r.URL.Path)
+       // if r has an Authorization header add username
+       // to context for {user} placeholder in downstream middleware
+       username, _, ok := r.BasicAuth()
+       if ok {
+               c = context.WithValue(c, RemoteUserCtxKey, username)
+       }
        r = r.WithContext(c)
```

mholt: @jamesboswell That's all fine and well, but I strongly want to avoid creating special cases, which this certainly is, for there will only be more and more. Can we generalize the fix instead?

jung-kurt: > If I don't dereference r, then the context value is not available to the replacer.

One clue is in the standard library source for [WithContext](https://golang.org/src/net/http/request.go?s=11679:11738#L317). The `context.WithValue()` call returns a wrapped context containing all values and cancellation functions plus the new specified value. Then the call to `r.WithContext()` returns a shallow copy of the request receiver with the new, recently extended, context. The pointer dereference occurs in `r.WithContext()`. I believe the original line, `r = r.WithContext()`, functions properly. The line it was replaced with, `*r = *r.WithContext()`, changes the request that all downstream middleware have pointers to. This is a potential race risk, and provides them access to values that they do not need (after all, it was an upstream middleware that added the value) and, maybe, should not see at all.

tobya:
@mholt Any suggestions on where to go next with this issue?

jung-kurt:
I'm willing to look into this further but I need some clarification of the original and subsequent problems. Does the radiusauth plugin provide the same end result as the basicauth plugin? (That is, on success it grants access by storing the user name in the request context and calling Next.ServeHTTP, otherwise it returns an unauthorized code?)

My CGI plugin is able to retrieve the user name that is stored by the basicauth plugin earlier in the middleware chain. It can do this whether basicauth stores the username as follows:

`r = r.WithContext(context.WithValue(r.Context(), httpserver.RemoteUserCtxKey, username))`

(which I believe is the correct way), or

`*r = *r.WithContext(context.WithValue(r.Context(), httpserver.RemoteUserCtxKey, username))`

(which I maintain is unsafe when goroutines may still access the request).

I am not clear why [radiusauth](https://github.com/jamesboswell/caddy-radius/blob/446f849e942e81e35dc98139e356cbe8328b9f5b/radius.go#L74-L76) doesn't use the same scheme for storing the username as basicauth. Logging middleware could then retreive the username just as, for example, the CGI plugin does. If all authorization plugins followed the scheme used by basicauth, the benefit would be that all subsequent middleware would have access to the username in a standard way.

tobya:
@jung-kurt Thank you very much for helping out with this.  The issue here however doesn't have anything to do with the radiusauth plugin or how it accesses the user value.  The issue is to do with the `replacer` 

I will try and explain but may get a bit muddled.

As matt hinted at, it is to do with 2 paths of execution and 2 copies of the request.

1 copy is used by the main code that calls `ServeHTTP` on all the various packages and plugins and another that gets used by `replacer` - somehow they are not the same, so if `withContext` is called without dereferencing, then the `RemoteUserCtxKey` is not available to the replacer.

Again to reiterate that this has nothing (at this point) to do with the radiusauth plugin that Jim is writing and infact this code is simply to work with basicauth.  If we can figure out how to make the withContext call work with the replacer for basicauth we can use the same method with the radiusauth plugin.

If we can work around what Matt is talking about here

> Ohhhh, is this because "the replacer" is "the replacer" that is being created by the logger in an earlier handler, and that replacer is the one that holds a reference to the first r?

we can make a single replacer that is available to all.

jung-kurt:
Thank you, @tobya, for the explanation. Everything is a lot clearer now and I appreciate the issue that has to be resolved.

I'll look more carefully at that code execution divergence. (Without digging in yet, I assume the logger runs in a separate goroutine and requires access to fields like the authorized user that get assigned in a different goroutine.)

mholt:
I need to give this more thought but I'll try to come back to this before the next release.

mholt:
I'm pretty sure this change (and any other place we've dereferenced a request) will introduce a race condition just waiting to explode when a handler needs to fire off a goroutine and passes it the request.

If I can't think of a solution by, well, really soon, it may not go in the release on Thursday, but that's ok -- we can get it in a later release. I'll keep this open. Suggestions welcome in the meantime.

jung-kurt:
> Suggestions welcome in the meantime.

My only thought is something like this:
* implement a simple goroutine-safe instance for saving and retrieving key-value pairs
* allocate an instance and save a pointer to it in the request's context when the request is first encountered
* use this instance for setting and getting values by key instead of using the request's context for this purpose

With this scheme, the execution path may diverge without losing access to shared values. The lifetime of the getter-setter instance would be that of its associated request.

mholt:
@jung-kurt That's an interesting idea. I need to think about whether we want to go that route! If there is a (good) way the Replacer can be implemented such that the Replacer doesn't store that pointer to the request, then this becomes a non-issue.

One possible fix for the dereference would be, I think, to move `basicauth` to be executed before `log`. But I think that will not be good/right, since basicauth can return errors which the `errors` directive needs to handle (among other middlewares in between the two that might need to run...)

jamesboswell:
>One possible fix for the dereference would be, I think, to move basicauth to be executed before log. But I think that will not be good/right, since basicauth can return errors which the errors directive needs to handle (among other middlewares in between the two that might need to run...)

@mholt  if I may humbly suggest again, just adding the username to the context before starting the middleware chain?  https://github.com/mholt/caddy/pull/1542#discussion_r110530309
 
Then the issue of `basicauth` before `log` is moot as well.  I know you're concerned about special cases, but a HTTP basic authorization header is RFC standard, not special, yes?  If a browser presents the credentials we just need the username for logs,  downstream middleware(s) can do whatever they like with authentication.

The issues with Replacer will need more work to avoid race conditions, but setting the Context values before any downstream middleware avoids that issue for now, does it not?

tobya:
My golang knowledge is nowhere near sufficient to solve this issue, but I do think it needs to be solved.  In this one instance it is not possible to have the {user} from either basicauth or any other login/auth plugin used by the replacer for output to the logs or use in caddyfile etc.  

There must be a clever golang solution to this problem and it will keep causing us problem until we solve it,. #1769 is going to run into similar issues.  If anyone involved in this thread @jamesboswell @jung-kurt @mholt or anyone else has a bit of time it would be great to move this forward.

mholt:
I know, I'm sorry. 😞 I'll try to get to this ASAP. I frankly don't know a good answer yet, but I'm pretty strongly set on _not_ introducing race conditions (i.e. performing a copy/dereference).

mholt:
@tobya Simply adding this to basicauth.go just after the `r = r.WithContext(...)` line (and no changes to that line) made it work for me in that I was able to add a `{user}` placeholder to my log format and it successfully showed the basicauth'ed user:

```go
if rr, ok := w.(*httpserver.ResponseRecorder); ok && rr.Replacer != nil {
	rr.Replacer.Set("user", username)
}
```

See if that does what we're trying to achieve; I think it will. If so, revise the PR with this change instead.

tobya:
I'll try that out @mholt and try to move this forward.  Thanks.

tobya:
@mholt Not sure what I'm doing wrong.  I reset my code back to master and added your line but the if statement is evaluating as false for me so the block is never run.  I'm not sure why. 

     if rr, ok := w.(*httpserver.ResponseRecorder); ok

this seems to be causing some sort of error.  I started from master so there shouldnt be any other code that is causing this issue.

mholt:
@tobya You need the log directive, and then you can use the {user} placeholder in your log format.

magikstm:
I tested latest change suggested by @mholt (https://github.com/mholt/caddy/pull/1542#issuecomment-321395635).

Tested locally with latest commit (https://github.com/mholt/caddy/commit/49e98a151827a3aae7e6ceed194c3180d1bcbe68) on a Windows 7 x64 laptop.

New tests pass and it works with this minimal caddyfile in the cmd prompt:
```
http://localhost:8080 {
  log stdout  
  basicauth /secret bob hiccup
}
```

I really like this PR. Will it be added in the next release?

mholt:
@magikstm Thanks for testing it! I'm glad you could confirm the minimal case of what I suggest works. However, Toby has recently discovered a flaw in my plan, where the `header` directive doesn't see the `{user}` placeholder that is set by the `basicauth` directive. I think I know why this is but now the question is one of design: do we change the current design of the Replacer or change how we implement it so that all middlewares share & use the same one?

magikstm:
Oh.

What is the flaw's negative impact for the end-user?

mholt:
@magikstm Probably nothing, other than a delay for a good solution. We just need a good solution. :)

tobya:
I'm going to close this and will submit a new PR assuming #1931 is merged at some point.

PR #1546, opened 2017-03-29
browse: fix arbitrary ordering of directories with sort=size

wmark:
While reading **browse** I noticed this one. We've given **directories** a special treatment due to differences in filesystem implementations, but actually forgot to sort them further by the next dimension.

As their order has been undefined until now (their apparent size is set to `-1` anyway), this does not change any expectations of legacy scripts.


mholt:
Thank you @wmark! Appreciate the attention to detail.

PR #1547, opened 2017-03-29
Allow setting directive tls to "no_email"

AndreKR:
This allows to suppress the prompt while still registering with LE without an email address.

I kept the string "no_email" in `caddytls.Config.ACMEEmail` all the way down and implemented the mangling to an empty string only in the `GetEmail()` receiver function that implements the `github.com/xenolf/lego/acme.User` interface. This seemed reasonable because the string is also used to identify the user when his credentials are saved to disk.

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -95,9 +95,11 @@ type Config struct {
 	DNSProvider string
 
 	// The email address to use when creating or
-	// using an ACME account (fun fact: if this
-	// is set to "off" then this config will not
-	// qualify for managed TLS)
+	// using an ACME account. If this is set to
+	// "off" then this config will not qualify
+	// for managed TLS. If it is set to "no_email"
+	// registration will be done with an empty
+	// email address.
 	ACMEEmail string
 
 	// The type of key to use when generating
--- a/caddytls/user.go
+++ b/caddytls/user.go
@@ -25,7 +25,11 @@ type User struct {
 
 // GetEmail gets u's email.
 func (u User) GetEmail() string {
-	return u.Email
+	if (u.Email == "no_email") {
+		return ""
+	} else {
+		return u.Email
+	}
 }
 
 // GetRegistration gets u's registration resource.
--- a/caddytls/user_test.go
+++ b/caddytls/user_test.go
@@ -40,6 +40,18 @@ func TestUser(t *testing.T) {
 	}
 }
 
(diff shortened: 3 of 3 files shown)


mholt:
Thanks for the PR, but I am not inclined to accept this, because entering no email address is considered bad practice, and we want to encourage Caddy users to be good members of the PKI community. Without an email address, the ACME CA cannot contact you if something goes wrong and you cannot recover your account, which does you a disservice.

AndreKR:
From https://github.com/mholt/caddy/blob/master/caddytls/user.go#L81 it didn't seem like a big deal. Also according to https://community.letsencrypt.org/t/what-are-accounts-do-i-need-to-backup-them/21318/2 you can't even recover accounts through email anymore. :man_shrugging:

mholt:
@AndreKR Instead of no_email, how about a change that simply allows the email to be blank? (Note that ToS agreement must still be enforced, so make sure the prompt isn't bypassed on a first run that is missing the `-agree` flag.)

AndreKR:
@mholt That was what I had at first, but then I noticed that the email address is also used as the directory name when the user is stored on disk, so I would have to mangle it to something at that point. I felt like that is against the **principle of least surprise** and given the fact that we already had special values (`off` and `self_signed`) I decided to use `no_email`, but I'm not married to the that.

mholt:
@AndreKR Yep, empty email becomes "default" folder name, no worries.

An empty `tls` directive is not needed. The tls directive can be omitted entirely in that case. To bypass the prompt, the -agree and -email flags can be used, or `tls ""` is fine too I think.

mholt:
(oops, I didn't mean to self-request a review, that's a weird notion anyway)

Hmm, I think I need to give this more thought; I'm getting a little confused. If we want to make an email address optional, we can skip the checks for when an email is blank; but we still need to ensure terms agreement (maybe by the presence of an account?) - anyway, thanks for the PR, but I'm going to close until we decide how to handle this.

AndreKR:
Yeah, I felt confused too, and I was going to make a decision table of the possible combinations of `tls`, `-agree` and `-email` and the possible use cases, but haven't gotten around to doing that, sorry.

mholt:
It's okay, I'm still open to discussing it later, I just don't have the time to squeeze it into the next release right now. Feel free to bring it up again though (maybe as an issue first) later!

PR #1551, opened 2017-03-30
Add new browse sort - namedirfirst

tobya:
I have submitted this as a new PR for #1536 as @wmark  requested that the origional PR #1527 be reverted.

This PR has the revert and then the necessary changes applied.

- `name` sort has been reverted to previous behaviour
- Behaviour is now the same for urls where sort=name is included such as  `https://…/…/?sort=name ×{asc,desc}`
- New sort created `namedirfirst` that sorts dir then file in alpha order
- Default order when no sort defined is `namedirfirst`
- json tests updated to include sort=name in url

Sorry to create a new PR but this seemed the most straightforward way to make a clean PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,3 @@ Caddyfile
 og_static/
 
 .vscode/
-
-caddyhttp/browse/temp*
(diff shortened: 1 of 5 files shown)


wmark:
LGTM

PR #1558, opened 2017-04-01
Fix#1556

pbeckmann:
Hey,

fixing https://github.com/mholt/caddy/issues/1556.
Unfortunately does not contain a functional test.

Ideas and tips for the test are appreciated.

Greetings

### DIFF
--- a/caddyhttp/proxy/upstream_test.go
+++ b/caddyhttp/proxy/upstream_test.go
@@ -297,7 +297,9 @@ func TestParseBlock(t *testing.T) {
 			}
 		}
 	}
+}
 
+func TestHealthSetUp(t *testing.T) {
 	// tests for insecure skip verify
 	isv_tests := []struct {
 		config string
--- a/caddyhttp/proxy/upstream_test.go
+++ b/caddyhttp/proxy/upstream_test.go
@@ -279,7 +279,7 @@ func TestParseBlock(t *testing.T) {
 	for i, test := range tests {
 		upstreams, err := NewStaticUpstreams(caddyfile.NewDispenser("Testfile", strings.NewReader(test.config)))
 		if err != nil {
-			t.Error("Expected no error. Got:", err.Error())
+			t.Errorf("Expected no error. Got: %s", err.Error())
 		}
 		for _, upstream := range upstreams {
 			headers := upstream.Select(r).UpstreamHeaders
@@ -313,7 +313,7 @@ func TestParseBlock(t *testing.T) {
 	for i, test := range isv_tests {
 		upstreams, err := NewStaticUpstreams(caddyfile.NewDispenser("Testfile", strings.NewReader(test.config)))
 		if err != nil {
-			t.Error("Expected no error. Got:", err.Error())
+			t.Errorf("Expected no error. Got: %s", err.Error())
 		}
 		for _, upstream := range upstreams {
 			staticUpstream, ok := upstream.(*staticUpstream)
--- a/caddyhttp/proxy/upstream_test.go
+++ b/caddyhttp/proxy/upstream_test.go
@@ -6,8 +6,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/mholt/caddy/caddyfile"
 	"reflect"
+
+	"github.com/mholt/caddy/caddyfile"
 )
 
 func TestNewHost(t *testing.T) {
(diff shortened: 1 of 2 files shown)


caddyhttp/proxy/upstream_test.go
```diff
@@ -225,4 +226,39 @@ func TestParseBlock(t *testing.T) {
...
+		upstreams, err := NewStaticUpstreams(caddyfile.NewDispenser("Testfile", strings.NewReader(test.config)))
+		if err != nil {
+			t.Error("Expected no error. Got:", err.Error())
+		}
+		for _, upstream := range upstreams {
+			staticUpstream, ok := upstream.(*staticUpstream)
+			if !ok {
+				t.Error("Type mismatch")
+			}
+			transport, ok := staticUpstream.HealthCheck.Client.Transport.(*http.Transport)
```
pbeckmann: At this point the type conversion does not run. Instead Transport seems to be `nil`.

mholt: The path that the code requires to configure a healthcheck requires that the health check path is not `""`, but the Caddyfiles passed in to the parser don't specify a health check path. Try that?

francislavoie: Yup, the client doesn't get created unless health_check is defined: https://github.com/mholt/caddy/blob/74664fe0467cc99f84fa64aa7175267b016366b0/caddyhttp/proxy/upstream.go#L109

francislavoie: Try with 

`{"proxy / localhost:8080 {\n health_check / \n}", false}`

`{"proxy / localhost:8080 {\n health_check / \n insecure_skip_verify \n}", true},`

I guess

pbeckmann: 🤦‍♂️ 
Thank you both.

caddyhttp/proxy/upstream_test.go
```diff
@@ -225,4 +227,38 @@ func TestParseBlock(t *testing.T) {
...
+			if !ok {
+				t.Error(reflect.TypeOf(transport.TLSClientConfig))
+				t.Error("Type mismatch")
```
mholt: This is a little clumsy, and I don't think we need reflection here. How about `t.Errorf("type mismatch: %#v", transport.TLSClientConfig)` instead?

caddyhttp/proxy/upstream_test.go
```diff
@@ -225,4 +227,38 @@ func TestParseBlock(t *testing.T) {
...
+				continue
+			}
+			if test.flag != transport.TLSClientConfig.InsecureSkipVerify {
+				t.Error("Test case #" + string(i) + "failed")
```
mholt: This will not have the intended output, I think. And a more useful error message is good. Instead, use `t.Errorf("test %d: expected transport.TLSClientConfig.InsecureSkipVerify=%v, got %v", i, ...)` (you get the idea)

caddyhttp/proxy/upstream_test.go
```diff
@@ -225,4 +227,38 @@ func TestParseBlock(t *testing.T) {
...
+		{"proxy / localhost:8080 {\n health_check / \n}", false},
+
+		// Test #2: with flag
+		{"proxy / localhost:8080 {\n health_check / \n insecure_skip_verify \n}", true},
+	}
+
+	for i, test := range isv_tests {
+		upstreams, err := NewStaticUpstreams(caddyfile.NewDispenser("Testfile", strings.NewReader(test.config)))
+		if err != nil {
+			t.Error("Expected no error. Got:", err.Error())
```
mholt: As I've suggested below, you can use t.Errorf to create properly-formatted strings with values of arbitrary types.

caddyhttp/proxy/upstream_test.go
```diff
@@ -297,4 +297,38 @@ func TestParseBlock(t *testing.T) {
...
+			transport, ok := staticUpstream.HealthCheck.Client.Transport.(*http.Transport)
+			if !ok {
+				t.Errorf("type mismatch: %#v", staticUpstream.HealthCheck.Client.Transport)
+				continue
+			}
+			if test.flag != transport.TLSClientConfig.InsecureSkipVerify {
+				t.Errorf("test %d: expected transport.TLSClientCnfig.InsecureSkipVerify=%v, got %v", i, test.flag, transport.TLSClientConfig.InsecureSkipVerify)
+			}
+		}
+	}
```
mholt: Sorry, I should have noticed this earlier (been a crazy weekend! our internet has been down so I'm doing this away from my office) - but I think these tests should be in a different function, they are not testing the parsing of the configuration; rather, they test the proper set up of health checks.

francislavoie:
@mholt https://www.youtube.com/watch?v=8Cmqxe6P4Ms :laughing: :laughing: :laughing:

Sorry, I had to, first thing I thought of.

elcore:
@francislavoie that's hilarious 😂😂😂😂 (childhood memories)

mholt:
There's a beautiful PR - thank you!

PR #1559, opened 2017-04-02
Allow markdown to match index file for extensions.  Fix #1418

### LINKED ISSUE #1418: Markdown does not apply to default directory index file
(Are you asking for help with using Caddy? Please use our forum instead: https://forum.caddyserver.com. If you are filing a bug report, please answer the following questions. If your issue is not a bug report, you do not need to use this template. Either way, please consider donating if we've helped you. Thanks!)

0.9.5

Use .html as the markdown extension and have file load as index file for directory with markdown

aubergine.kinchik.ie:80 {
    root d:\caddy\webroot\mdissue
    markdown /md {
        ext .html
    }

}

caddy -conf=d:\godev\bin\caddymd.txt -log stdout

when i visit
> http://aubergine.kinchik.ie/md/

I expect to see http://aubergine.kinchik.ie/md/index.html  rendered in markdown

http://aubergine.kinchik.ie/md/index.html rendered as html 

If I visit http://aubergine.kinchik.ie/md/index.html directly then I see the markdown rendered properly, but not when the file is omitted.

Should there be a index directive like there is for the fastcgi and other directives.  This would also remove the necessity to use .html as a md ext

Use the caddyfile provided.

tobya:
This PR ensures that an index file for each markdown extension is check for. Fixes #1418 

Edit: I realised I had explained this wrong.

The only index file that was getting matched for markdown was index.md even if the extension was different.

This PR adds an index file to be matched for each extension provided for markdown.  Index files are now also matched on a block basis.

As ever my newness to GO may betray me so I am open to any comments about what changes I should make to structure my code in a better way.

The default extensions (.md, .markdown and .mdown) are represented as a hash rather than array so  if an index.md, index.markdown and index.mdown exist in a directory it is random as to which gets displayed.  I decide this was unlikely to happen so ok to leave.

### DIFF
--- a/caddyhttp/markdown/markdown.go
+++ b/caddyhttp/markdown/markdown.go
@@ -107,7 +107,6 @@ func (md Markdown) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 
 		// Set path to found index file
 		fpath = idx
-
 	}
 
 	// If not supported extension, pass on it
--- a/caddyhttp/markdown/markdown.go
+++ b/caddyhttp/markdown/markdown.go
@@ -29,9 +29,6 @@ type Markdown struct {
 
 	// The list of markdown configurations
 	Configs []*Config
-
-	// The list of index files to try
-	IndexFiles []string
 }
 
 // Config stores markdown middleware configurations.
@@ -50,6 +47,9 @@ type Config struct {
 
 	// List of JavaScript files to load for each markdown file
 	Scripts []string
+
+	// The list of index files to try
+	IndexFiles []string
 
 	// Template(s) to render with
 	Template *template.Template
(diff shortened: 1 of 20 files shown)


caddyhttp/markdown/markdown.go
```diff
@@ -107,6 +107,7 @@ func (md Markdown) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 
 		// Set path to found index file
 		fpath = idx
+
```
mholt: Not a showstopper, but in the future, I prefer that closing curly braces don't have extra blank lines before them, it makes it easier to see exactly what it's enclosing.

caddyhttp/markdown/setup.go
```diff
@@ -75,6 +75,14 @@ func markdownParse(c *caddy.Controller) ([]*Config, error) {
...
 		}
 
+		// Make a list of index files to match extensions
+		i := 0
+		md.IndexFiles = make([]string, len(md.Extensions))
+		for ext := range md.Extensions {
+			indexName := "index" + ext
+			md.IndexFiles[i] = indexName
+			i++
+		}
```
mholt: I think this whole change can be simplified:

```go
for ext := range md.Extensions {
    md.IndexFiles = append(md.IndexFiles, "index" + ext)
}
```

Pre-allocating is good for a slice that will be very large, but when there's like 5 or 10 things, I prefer shorter, less error-prone code.

tobya: @mholt understood. I've made that change now

mholt:
Fabulous - thanks again Toby!

PR #1577, opened 2017-04-14
Fix #1574, Proxy health check does not respect hostname when upstream Host header is configured 

### LINKED ISSUE #1574: Proxy health check does not respect upstream headers
This is a followup issue to https://github.com/mholt/caddy/issues/1556

My upstream server expects headers from the `transparent` directive to be passed through for the health check to work. My upstream server is a caddy server listening on a specific hostname (i.e. not `0.0.0.0` but instead for the domain name).

I'm trying to write the fix, essentially we just need to mutate the headers similarly to https://github.com/mholt/caddy/blob/7d15435361c95aa029be244f962b47950487a539/caddyhttp/proxy/proxy.go#L185-L191, but the problem I see is that we don't have a request as a base for the replacer when doing a health check, since it's an out-of-band check.

We need a request to pass into this:

```Go
replacer := httpserver.NewReplacer(req, nil, "")
```

Any ideas?

Side question, since I'm quite new to Go, it feels gross to be copying the `mutateHeadersByRules` func from `proxy.go` into `upstream.go` for this, what would be the ideal way to re-use that code in this case?

francislavoie:
Fixes #1574

---
- Change `NewStaticUpstreams` func signature to add the host string as a parameter
- Change all usages of `NewStaticUpstreams` to pass `""` for the new parameter
- Change `staticUpstream.HealthCheck` struct to store the host string
- Change `NewStaticUpstreams` to set `staticUpstream.HealthCheck.Host`
- Set `req.Host` if `upstreamHeaders` includes `Host` with the `{host}` placeholder
---
- ~~Still needs tests~~


francislavoie:
That's odd, the build from Travis CI had a spelling error from outside my changes. Hopefully I'm not stepping on any toes by fixing it.

pbeckmann:
Hey,
great work!
- For the test; Maybe we could do it like for the insecureSkipVerify and test if the headers and host are sent?
- About changing the function signature. I'm fine with it, just as an idea, wouldn't it be better to use the whole Caddy config as a parameter?

caddyhttp/proxy/upstream.go
```diff
@@ -50,7 +51,7 @@ type staticUpstream struct {
 
 // NewStaticUpstreams parses the configuration input and sets up
 // static upstreams for the proxy middleware.
```
pbeckmann: Maybe we should add here a small comment, why we are adding the host?

francislavoie:
Regarding the func signature, I was just thinking principle of least privilege, only pass the info it needs. That would probably be fine though. A step further, we could pass the controller in (since we can get the config with the controller)?

Thanks for the comments, I'll work on it later today,

pbeckmann:
You are right ✌️ You should pass at least information possible into the middleware.
Otherwise there would be the danger, that the configuration could be changed from inside the middleware. So let it, like it is.
Just the comment would be nice, and of course the test....

francislavoie:
Alright, so I added the test and improved the comment for `NewStaticUpstreams`.

I also realized that I could improve flow by doing the string replace in `NewStaticUpstreams` instead of during the actual `HealthCheck()` (very marginal efficiency gain), because it shouldn't change because it's based off config.

Also, I took the liberty of cleaning up your test @pbeckmann, I hope that's not a problem - changes `isv_tests` to `tests` and made the error messages uppercase, for consistency.

Apparently the build failed on Travis CI, but I don't understand why. 
`The command "diff <(echo -n) <(gofmt -s -d .)" exited with 1.`
What's that about? :thinking:

elcore:
Hello @francislavoie,

>Apparently the build failed on Travis CI, but I don't understand why.
The command "diff <(echo -n) <(gofmt -s -d .)" exited with 1.
What's that about? 🤔

This should help you 😄

francislavoie:
@elcore aha, thanks. I'm still new to this. :smile:

elcore:
> @elcore aha, thanks. I'm still new to this. 😄

You are welcome 😄 ; I wish you a nice time in the world of Go and Caddy!

francislavoie:
Fyi this PR also includes https://github.com/mholt/caddy/pull/1586 (I fixed it here because I didn't like Travis CI complaining), hope that won't be a problem.

caddyhttp/proxy/upstream.go
```diff
@@ -371,7 +382,16 @@ func (u *staticUpstream) healthCheck() {
 	for _, host := range u.Hosts {
 		hostURL := host.Name + u.HealthCheck.Path
 		var unhealthy bool
-		if r, err := u.HealthCheck.Client.Get(hostURL); err == nil {
+
+		// set up request, needed to be able to modify headers
+		req, _ := http.NewRequest("GET", hostURL, nil)
```
mholt: Check the error - if the URL is bad (thus resulting in an error), the upstream is considered unhealthy.

caddyhttp/proxy/upstream_test.go
```diff
@@ -312,24 +312,65 @@ func TestHealthSetUp(t *testing.T) {
...
+			t.Errorf("Expected no error. Got: %s", err.Error())
+		}
+		for _, upstream := range upstreams {
+			staticUpstream, ok := upstream.(*staticUpstream)
+			if !ok {
+				t.Errorf("Type mismatch: %#v", upstream)
+				continue
+			}
+			if test.flag != (staticUpstream.HealthCheck.Host == test.host) {
+				t.Errorf("Test %d: expected staticUpstream.HealthCheck.Host=%v, got %v", i, test.host, staticUpstream.HealthCheck.Host)
```
mholt: Test looks good! :+1:

caddyhttp/proxy/upstream.go
```diff
@@ -49,8 +50,10 @@ type staticUpstream struct {
 }
 
 // NewStaticUpstreams parses the configuration input and sets up
-// static upstreams for the proxy middleware.
-func NewStaticUpstreams(c caddyfile.Dispenser) ([]Upstream, error) {
+// static upstreams for the proxy middleware. The host string parameter,
+// if not empty, is used for setting the upstream Host header for the
+// health checks if the upstream header config requires it.
```
mholt: LOVE the comment, good explanation.

caddyhttp/proxy/upstream.go
francislavoie: Ah okay, didn't really read into the error cases for `NewRequest`. Will do.

francislavoie: Should be fixed now. :+1:

mholt:
Thanks @francislavoie! Would you like to be added as a collaborator so you can help preview pull requests, fix issues, etc, on a branch rather than in a fork? (We just ask that all PRs have at least one other collaborator approve before merging assuming the PR is not large; and we squash most PRs with multiple commits - we ask that large PRs get spliced up into multiple smaller ones).

francislavoie:
Sure! That'd be cool. :+1:

Issue #1529, opened 2017-03-21
Log duplicated to ./access.log


PetterS:
https://forum.caddyserver.com/t/caddy-runs-out-of-file-descriptors/1659/7 has the Caddyfile and configuration.

Caddy logs everything to ./access.log as well, even though it is not in the Caddyfile.

mholt:
@julianvmodesto Would you have a moment to look at this? I was able to reproduce this behavior.

Baughn:
Can confirm. This leads to failure on startup if CWD isn't inside a writeable directory, even if nothing in the configuration file writes to CWD.

Tweenagedream:
Pretty sure this is caused by the nested rotate, flattening those options with #1379 I believe fixes the issue (kinda)

Technically you can still put
rotate {
}
tags in there even if they don't do anything and cause this behavior. 

unknown whats seem to be silently ignored: https://github.com/mholt/caddy/blob/60415c8d0e365189a6014b610ef72c8a8a84b2e7/caddyhttp/log/setup.go#L44

You can actually add any block at all inside of log on two separate lines:
log ./log/stuff.log {
  thisblockwillmakeitlogtocwd {
  }
}

If interested in fixing this patch would look something like:
diff --git a/caddyhttp/log/setup.go b/caddyhttp/log/setup.go
index 2366aec..e5e5e09 100644
--- a/caddyhttp/log/setup.go
+++ b/caddyhttp/log/setup.go
@@ -30,11 +30,11 @@ func logParse(c *caddy.Controller) ([]*Rule, error) {
 
        for c.Next() {
                args := c.RemainingArgs()

                var logRoller *httpserver.LogRoller
                logRoller = httpserver.DefaultLogRoller()

+               wasNested := false
                for c.NextBlock() {
+                       wasNested = true
                        what := c.Val()
                        if !c.NextArg() {
                                return nil, c.ArgErr()
@@ -49,7 +49,9 @@ func logParse(c *caddy.Controller) ([]*Rule, error) {
                                }
                        }
                }

+               if wasNested {
+                       c.Next()
+               }
                if len(args) == 0 {
                        // Nothing specified; use defaults
                        rules = appendEntry(rules, "/", &Entry{

I can make a PR if it's worth it...

Baughn:
My (autogenerated) Caddyfile looks like this: http://sprunge.us/DOLA

...but this still produces an access.log. In fact, all requests to any server defined there gets logged *three times* to access.log, so it looks like this:
[...]

PR #1583, opened 2017-04-17
log: only allow new roller related options in a block

tw4452852:
fix issue #1529


mholt:
Misspelling is fixed in another PR, so even though that check failed, still merging. Thanks Tw!

Issue #1582, opened 2017-04-17
Encoded slash in URL


Cinerar:
### 1. What version of Caddy are you running (`caddy -version`)?

Caddy 0.9.5 (+7d15435 Sun Apr 16 00:21:33 UTC 2017)

### 2. What are you trying to do?
 Proxy request

### 3. What is your entire Caddyfile?
```text
localhost:8888 {
  # proxy to "nc -l localhost 9999"
  proxy /test http://localhost:9999
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

Just  run caddy with this caddyfile

### 5. Please paste any relevant HTTP request(s) here.

```
 curl localhost:8888/test/%2F/foo
```

### 6. What did you expect to see?
```
nc -l localhost 9999
GET /test/%2F/foo HTTP/1.1
Host: localhost:9999
User-Agent: curl/7.47.0
Accept: */*
X-Forwarded-For: 127.0.0.1
Accept-Encoding: gzip
```
### 7. What did you see instead (give full error messages and/or log)?
```
nc -l localhost 9999
GET /test/foo HTTP/1.1
Host: localhost:9999
User-Agent: curl/7.47.0
Accept: */*
X-Forwarded-For: 127.0.0.1
Accept-Encoding: gzip
```

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
1. Caddyfile:
```
localhost:8888 {
  #proxy to "nc -l localhost 9999"
  proxy /test http://localhost:9999
}
```
2. Run
```
nc -l localhost 9999
```
3. Run
```
curl localhost:8888/test/%2F/foo
```
4. See
```
GET /test/foo HTTP/1.1
Host: localhost:9999
User-Agent: curl/7.47.0
Accept: */*
X-Forwarded-For: 127.0.0.1
Accept-Encoding: gzip
```

I believe this issue is #643 reintroduced with https://github.com/mholt/caddy/commit/c37481cc7b2e2476fe50e17f31421bf20c35cc75
and there are no slashes becouse of https://github.com/mholt/caddy/commit/04bee0f36d218edf0593b2c2254974f35d00276e #1298 

@tw4452852 please check it out.

tw4452852:
@Cinerar 
Patch is ready, could you please check it again? Thanks!

Cinerar:
@tw4452852 looks like everything ok now.

mholt:
[...]

PR #1584, opened 2017-04-17
proxy: use untouched URL for concatenating

tw4452852:
fix issue #1582 

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -25,6 +25,7 @@ import (
 	"golang.org/x/net/http2"
 
 	"github.com/mholt/caddy/caddyhttp/httpserver"
+	"github.com/mholt/caddy/caddyhttp/staticfiles"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


caddyhttp/proxy/reverseproxy.go
```diff
@@ -154,7 +155,12 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 				prefer(target.RawPath, target.Path),
 				prefer(req.URL.RawPath, req.URL.Path))
 		}
-		req.URL.Path = singleJoiningSlash(target.Path, req.URL.Path)
+		req.URL.Path = singleJoiningSlash(target.Path,
+			prefer(req.Context().Value(staticfiles.URLPathCtxKey).(string), req.URL.Path))
+		// req.URL.Path must be consistent with decoded form of req.URL.RawPath if any
+		if req.URL.RawPath != "" && req.URL.RawPath != req.URL.EscapedPath() {
+			panic("RawPath doesn't match Path")
```
mholt: I know this function doesn't return an error, but we shouldn't panic. Would this if statement be better in a test instead?

Cinerar: BTW, i'm not sure but looks like that there are tests to test this case #1582, but as long as i understand they pass when they should not.
https://github.com/mholt/caddy/blob/master/caddyhttp/proxy/proxy_test.go#L518

tw4452852: @mholt I still think we should panic here because if the final request's `RawPath` and `Path` are not consistent. There must be a bug in our director logic and we should recognized it ASAP.
BTW I have add a test in the new patchset.

tw4452852: @Cinerar Because we have sanitize the original request's URL ([here](https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/server.go#L297). So if the original request's URL is `/%2F/test`, it will be changed to `/test` before proxy middleware. This is the reason why some tests don't catch.

tw4452852:
Uploaded a new patch set to add a test and make `go test -race` happy.

caddyhttp/proxy/reverseproxy.go
mholt: Okay, that's reasonable; we can panic for programmer error. Should be recovered in the handler chain anyway.

mholt:
Nice work Tw, thanks for the revision. This will go out on the new release this week.

mholt:
@tw4452852 Unfortunately this breaks proxy paths when `rewrite` or `ext` changes the req.URL.Path value... We need a better way to preserve the slash.

PR #1585, opened 2017-04-17
fix a mispell to make CI happy

tw4452852:


mholt:
Oh, thanks for catching this! As you probably saw, there's a larger PR that came in earlier that also fixes this, and it will probably be merged. Thanks though!

PR #1588, opened 2017-04-17
Fixed #1502: Proxying of unannounced trailers

### LINKED ISSUE #1502: Reverse proxying GRPC calls does not work

lhecker:
As per #1512 this PR only contains the necessary changes to fix #1502 and nothing more.
Sorry for closing the other PR and sorry for the delays - I'm having a kinda hard time in life right now and things like that are the result I guess. 😕

P.S.: I have no idea why one of the commit tests failed... Maybe a race condition? Because I haven't seen such a failure personally yet. <sup>With Rust that wouldn't have happened.</sup>

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -318,30 +318,61 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request, 
 		}
 		pooledIoCopy(backendConn, conn)
 	} else {
+		// NOTE:
+		//   Closing the Body involves acquiring a mutex, which is a
+		//   unnecessarily heavy operation, considering that this defer will
+		//   pretty much never be executed with the Body still unclosed.
+		bodyOpen := true
+		closeBody := func() {
+			if bodyOpen {
+				res.Body.Close()
+				bodyOpen = false
+			}
+		}
+		defer closeBody()
+
+		// Copy all headers over.
+		// res.Header does not include the "Trailer" header,
+		// which means we will have to do that manually below.
 		copyHeader(rw.Header(), res.Header)
 
-		// The "Trailer" header isn't included in the Transport's response,
-		// at least for *http.Transport. Build it up from Trailer.
-		if len(res.Trailer) > 0 {
-			trailerKeys := make([]string, 0, len(res.Trailer))
+		// The "Trailer" header isn't included in res' Header map, which
+		// is why we have to build one ourselves from res.Trailer.
+		//
+		// But res.Trailer does not necessarily contain all trailer keys at this
+		// point yet. The HTTP spec allows one to send "unannounced trailers"
+		// after a request and certain systems like gRPC make use of that.
+		announcedTrailerKeyCount := len(res.Trailer)
+		if announcedTrailerKeyCount > 0 {
+			vv := make([]string, 0, announcedTrailerKeyCount)
 			for k := range res.Trailer {
-				trailerKeys = append(trailerKeys, k)
+				vv = append(vv, k)
 			}
-			rw.Header().Add("Trailer", strings.Join(trailerKeys, ", "))
+			rw.Header()["Trailer"] = vv
 		}
 
+		// Now copy over the status code as well as the response body.
 		rw.WriteHeader(res.StatusCode)
-		if len(res.Trailer) > 0 {
+		if announcedTrailerKeyCount > 0 {
 			// Force chunking if we saw a response trailer.
-			// This prevents net/http from calculating the length for short
-			// bodies and adding a Content-Length.
+			// This prevents net/http from calculating the length
+			// for short bodies and adding a Content-Length.
 			if fl, ok := rw.(http.Flusher); ok {
 				fl.Flush()
 			}
 		}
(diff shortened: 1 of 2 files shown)


mholt:
Thanks @lhecker. It's not a problem - take care of things you need to. I hope things look up for you soon!

I'm not sure about the failing test either, but it's something worth looking into! Even if this change doesn't go into this next release, it'll get in one of the next ones.

Know that I and many others (perhaps unknowingly) appreciate your work. I'll review this soon, but might also wait until you're ready to circle back around to it.

lhecker:
@mholt My life is slowly getting back to normal and starting next month everything should finally be a bit more relaxed again. 😂

I just rebased this PR on the current master. So how about we take another look at this PR now? The new & shiny support for some glorious gRPC proxying has been waiting far too long already. 🙂

P.S.: When the tests on this PR originally failed after submitting this PR it I bet a race condition was at fault. Since the underlying issue might not have been fixed yet I think we should carefully look out for that in caddy in the future. I mean: You do run the tests with `-race` enabled but who knows if that catches the rather spurious bugs (e.g. those who only happen if caddy is preempted or interrupted at a critical moment etc.).

lhecker:
@mholt I fixed the 2 issues you mentioned. The PR indeed was quite a bit messy, but I cleaned it up now (no TODOs anymore or removed code).

The issue with the test was - or rather "is" - that it's only a fake HTTP request test. Since we never actually serialize and submit `r` (the `http.Request`), but instead submit the fake request/response directly to the proxy's `ServeHTTP` we will never know if proxying of unannounced trailers actually works. That will only ever be the case if we do a proper E2E test over localhost.
Code wise on my side everything's ready for a proper E2E though, after I added the new `trailerTestStringReader` struct just now. 🙂

caddyhttp/proxy/reverseproxy.go
```diff
@@ -318,30 +318,61 @@ func (rp *ReverseProxy) ServeHTTP(rw http.ResponseWriter, outreq *http.Request,
 		}
 		pooledIoCopy(backendConn, conn)
 	} else {
+		// NOTE:
+		//   Closing the Body involves acquiring a mutex, which is a
```
mholt: Why would this require a mutex? I'm trying to see where it would be closed in a different goroutine.

lhecker: The Go stdlib uses a Mutex for locking the `res.Body` when calling `Close()` on it. See [here](https://github.com/golang/go/blob/66b5a2f3f0b1d94f69763114a85a977f5bc0277a/src/net/http/transfer.go#L905-L906). I thought that having a bit additional logic in our code is a okay-ish cost considering that the above `defer` will never ever once do anything meaningful until that very rare occurence that our code panics.

mholt: Ah, I see, we don't want to call Close() more than once. Makes sense.

mholt:
Feel free to submit small PRs with the other changes you had in #1512 if you still want to do those. :+1:

lhecker:
Yeah I'll do that. 🙂 And if I have the opportunity to contribute something again afterwards it'd probably be something to integrate caddy with kubernetes or so... I don't really want to use linkerd for that. 😂 And maybe something to make the config file more generic - I honestly fear that the current format might become a mayor painpoint as soon as we accumulate more and more plugins which aren't necessarely strictly centered around HTTP or domain names at all.

mholt:
Let's discuss it more, either in an issue, or I've also sent you an email with some information. Thanks again! :+1:

PR #1593, opened 2017-04-18
Fixes #1576 -- automatically support CT for sites!

### LINKED ISSUE #1576: Support for static certificate transparency files
Like [this module](https://github.com/grahamedgecombe/nginx-ct) allows it for NGINX

alex:
There's a handful of TODOs in the code, and I think it needs some serious design feedback from @mholt, and then some "how do we CT best" feedback from the community.

But it works! I tested it! (If I was smart, I'd have left the server running, instead of deleting the VM it was running in)


caddytls/config.go
```diff
@@ -119,6 +119,10 @@ type Config struct {
 	// Protocol Negotiation (ALPN).
 	ALPN []string
 
+	// The list of CT logs to attempt to submit the certificate to in order to
+	// obtain SCTs to server in the handshake.
```
mholt: "to serve" in the handshake ;)

caddytls/setup.go
```diff
@@ -174,6 +174,14 @@ func setupTLS(c *caddy.Controller) error {
...
 			case "must_staple":
 				config.MustStaple = true
+			case "ct_logs":
+				args := c.RemainingArgs()
+				if len(args) == 0 {
+					return c.ArgErr()
+				}
+				for _, arg := range args {
+					config.CTLogURLs = append(config.CTLogURLs, arg)
+				}
```
mholt: Could be worth adding a test to setup_test.go that ct_logs does indeed fill out the CTLogURLs field. (easy test to do)

caddytls/certificates.go
```diff
@@ -182,6 +182,23 @@ func makeCertificate(certPEMBlock, keyPEMBlock []byte) (Certificate, error) {
 	if err != nil {
 		log.Printf("[WARNING] Stapling OCSP: %v", err)
 	}
+	// TODO: We don't have the Config.CTLogURLs here, just hardcoded some
+	// Config.Managed (with Let's Encrypt) specific values here.
```
mholt: Easy enough I think: `makeCertificate` is called only one place where Managed certificates are loaded: `CacheManagedCertificate` which does have access to a Config. Maybe the list can be passed in as an argument? And for the other two places where this function is called, it's just one more function up. We can start with that but if it gets too messy then we can look at relocating this particular part of the code.

caddytls/certificates.go
```diff
@@ -182,6 +182,23 @@ func makeCertificate(certPEMBlock, keyPEMBlock []byte) (Certificate, error) {
...
+	// TODO: Venafi was just distrusted; one idea would be to fetch the
+	// Chrome list of trusted logs and just submit to everything until we
+	// have 1 Google and 1 non-Google. This would be hilarious, and pretty
+	// resillient.
+	sctLogURLs := []string{
+		"https://ct.googleapis.com/icarus",
+		"https://ctlog.api.venafi.com",
+	}
+	// TODO: We can skip this if the leaf has SCTs in an X.509 extension.
```
mholt: Can this be done easily? Like, before merging? If not, let's do it ASAP after the release on Thursday.

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, err
+	}
+	return sct, nil
+}
+
+// GetSCTSForCertificateChain takes a certificate chain, and a list of target
+// log URLs, and returns a list of SCTs (byte slices) or an error.
+
+// TODO: support submitting to multiple logs and just getting as many SCTs as
+// we can, even if some logs error.
```
mholt: When this is implemented, do so concurrently perhaps for speed?

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,126 @@
...
+func submitSCT(url string, payload []byte) (*signedCertificateTimestamp, error) {
+	if !strings.Contains(url, "://") {
+		// Allow people to omit the protocol and default to HTTPS.
+		url = "https://" + url
+	}
+	if !strings.HasSuffix(url, "/") {
+		url = url + "/"
+	}
+
+	response, err := http.Post(url+"ct/v1/add-chain", "application/json", bytes.NewReader(payload))
```
mholt: Maybe worth setting up an http.Client with a timeout value so as to avoid hanging/lingering connections that block startup or maintenance along the way?

Also, does this have a noticeable impact on startup time? Or is it pretty quick usually? If it noticeably slows down startup, maybe we can load the certificate like we've always done but then attach the SCTs in the background after it's started, making them available just a few seconds later. But that doesn't have to happen now, and maybe should only happen at all if it's a problem. Just thinking out loud... but still worth setting up a client with timeouts I think.

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,126 @@
...
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("HTTP error %d", response.StatusCode)
+	}
+	responseBody, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+	sct := &signedCertificateTimestamp{}
+	err = json.Unmarshal(responseBody, &sct)
```
mholt: I bet we can get these last 4-5 lines down to `err = json.NewDecoder(response.Body).Decode(&sct)`. Voila, no need to buffer per-se.

caddytls/setup.go
alex: Done.

caddytls/certificate_transparency.go
alex: I haven't measured, so I can't say for sure.

caddytls/certificates.go
```diff
@@ -149,14 +164,26 @@ func makeCertificateFromDisk(certFile, keyFile string) (Certificate, error) {
 	if err != nil {
 		return Certificate{}, err
 	}
-	return makeCertificate(certPEMBlock, keyPEMBlock)
+	// XXX: Get a Config and pass the real CTLogURLs
+	return makeCertificate(certPEMBlock, keyPEMBlock, []string{})
+}
+
+var x509SCTOid = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 2}
```
mholt: I think I know what this means, but maybe put a comment on this, it's a bit obscure. :)

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,123 @@
...
+	addReq := addChainRequest{}
+	for _, cert := range certChain {
+		addReq.Chain = append(addReq.Chain, base64.StdEncoding.EncodeToString(cert))
+	}
+	payload, err := json.Marshal(addReq)
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient := http.Client{Timeout: 30 * time.Second}
```
mholt: I see that you're creating just one client for each log URL, that is good. But I wonder if we should just declare this value once outside of the function: `var sctClient = http.Client{Timeout: 30 * time.Second}` so it can be reused for all SCT things?

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Timestamp
+	binary.Write(b, binary.BigEndian, sct.Timestamp)
+
+	// Extensions
+	extensionBytes, err := base64.StdEncoding.DecodeString(sct.Extensions)
+	if err != nil {
+		return nil, err
+	}
+	if len(extensionBytes) > 65535 {
```
mholt: Good check!

caddytls/certificates.go
```diff
@@ -128,7 +142,8 @@ func cacheUnmanagedCertificatePEMFile(certFile, keyFile string) error {
 //
 // This function is safe for concurrent use.
 func cacheUnmanagedCertificatePEMBytes(certBytes, keyBytes []byte) error {
-	cert, err := makeCertificate(certBytes, keyBytes)
+	// XXX: Get a Config and pass the real CTLogURLs
```
mholt: Probably the way to do this for now (and for the below `XXX:`) is to pass the Config into this function so its log URLs can be passed to makeCertificate. I'm still thinking if the actual submitting and stapling the SCT should be done elsewhere so we don't have to lob the Config everywhere but it won't stop a merge, we can refactor later.

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,123 @@
...
+	return b.Bytes(), nil
+}
+
+// Makes an HTTP request to the log server and returns the parsed SCT response.
+// If the cert is already in the log, the log will simply return the previously
+// generated SCT, making this idempotent.
+func submitSCT(client http.Client, url string, payload []byte) (*signedCertificateTimestamp, error) {
+	if !strings.Contains(url, "://") {
+		// Allow people to omit the protocol and default to HTTPS.
+		url = "https://" + url
```
AGWA: Certificate Transparency requires the use of HTTPS, so I suggest always prepending `https://` and requiring logs to be specified without the protocol, as this is how log addresses are conventionally represented.

mholt: Would there be any use case for a local CT server for testing or development that the operator would just want to use HTTP with?

caddytls/certificates.go
```diff
@@ -99,7 +100,20 @@ func (cfg *Config) CacheManagedCertificate(domain string) (Certificate, error) {
...
 	}
-	cert, err := makeCertificate(siteData.Cert, siteData.Key)
+
+	ctLogURLs := cfg.CTLogURLs
+	if len(ctLogURLs) == 0 {
+		// TODO: Venafi was just distrusted; one idea would be to fetch the
+		// Chrome list of trusted logs and just submit to everything until we
+		// have 1 Google and 1 non-Google. This would be hilarious, and pretty
+		// resillient.
+		ctLogURLs = []string{
```
elcore: > [...] Venafi was just distrusted [...]

If Venafi is distrusted we should replace it with another log!

> one idea would be to fetch the Chrome list of trusted logs and just submit to everything until we have 1 Google and 1 non-Google. This would be hilarious, and pretty

This sounds amazing 😄

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,122 @@
...
+	response, err := httpClient.Post(url+"ct/v1/add-chain", "application/json", bytes.NewReader(payload))
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("HTTP error %d", response.StatusCode)
+	}
+	sct := &signedCertificateTimestamp{}
+	err = json.NewDecoder(response.Body).Decode(&sct)
```
mholt: This is the paranoia in me, but it should be easy to wrap `response.Body` with [`io.LimitReader`](https://golang.org/pkg/io/#LimitReader) using a sane size in case a CT server is compromised or buggy. I'm not sure what a reasonable size would be; it should definitely be large enough to handle any valid response we could possibly see, maybe times 2?

caddytls/certificates.go
elcore: @mholt / @alex I would like to hear your thoughts 😄

AGWA:
I strongly encourage you not to hardcode a list of CT logs in Caddy.  It is very important to the ecosystem that CT logs can be easily distrusted - in just the last year, two logs (Izenpe and Venafi) had to be distrusted because they made fatal mistakes that completely undermined their security guarantees.  If someone has an out-of-date version of Caddy and is relying on Caddy's automatic SCT support to achieve mandatory CT qualification, their site will stop working when one of the hardcoded logs is distrusted.  If enough sites are in this situation, user agents will have a difficult time distrusting logs, much like distrusting CAs is difficult today.  This situation is detrimental to security and should be avoided with CT if at all possible.

Retrieving Chrome's trusted log list on-the-fly, as one comment suggests, is a much better approach.  Chrome already publishes its list (see https://groups.google.com/a/chromium.org/forum/#!msg/ct-policy/rQIM7JI99lI/AECBJ_cOAAAJ), and there are discussions in progress to define a common data format that all user agents can use to publish their trusted log lists.

This problem can also exist with log lists that an administrator hardcodes in their config file.  I encourage you to think carefully about the use cases for configuring a custom log list.  I suspect that the usefulness of specifying custom logs will be limited to a few very special cases.  Is there any way to reflect this in how the option is configured?

I'm very excited to see automatic SCT management in TLS servers, as when done well it can be a great boon to ecosystem agility.  This PR is an excellent start, but I don't think it's quite ready for merging yet.

mholt:
@AGWA Fair enough -- I didn't realize CT logs were so easily/often distrusted. We can retrieve the list at startup and then refresh it every 24 hours; is that a good interval?

We can start with Chrome's list and update the retrieval code as a standard emerges.

As for configuring log lists in the Caddyfile, we could either:

**A)** Major "caveat emptor" - you better know what you're doing kind of thing (we already allow server admins to enable less-than-ideal cipher suites and protocols if needed)

or

**B)** They specify a location from which to obtain a log list instead of specifying the log list itself.

Thoughts?

mholt:
Link to relevant thoughts on CT by Ryan Sleevi: https://gist.github.com/sleevi/08e578719d92d9a5a3c2e1d59839b77e

alex:
(travis test failure is a transient network thing and should be re-run)

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,207 @@
...
+	operators []struct {
+		name string
+		id   int
+	}
+}
+
+// GetTrustedCTLogs returns a list of CT logs trusted by Chrome. As the
+// browser/CT ecosystem evolves it may return other CT logs as well.
+func GetTrustedCTLogs() ([]ctLog, error) {
+	response, err := http.Get("https://www.gstatic.com/ct/log_list/log_list.json")
```
alex: This should be `httpClient.Get` so we get the timeline

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+type ctLog struct {
+	url string
+	// Chrome's CT policy requires one SCT from a Google log, and one SCT from
+	// a non-Google log, so we track whether a log is Google or not.
+	is_google bool
+}
+
+type logList struct {
```
elcore: @alex please use "Mixed Caps" 😄

caddytls/setup.go
```diff
@@ -174,6 +176,13 @@ func setupTLS(c *caddy.Controller) error {
 				}
 			case "must_staple":
 				config.MustStaple = true
+			case "certificate_transparency":
+				args := c.RemainingArgs()
+				if len(args) == 0 || args[0] == "yes" {
+					config.CertificateTransparency = true
+				} else if args[0] == "no" {
+					config.CertificateTransparency = false
+				}
```
mholt: What if we just call it `ct` instead of `certificate_transparency` (the Caddyfile is designed to be easy to write, and I think the "CT" acronym is pretty common)? Also, since it's on by default, there's no reason IMO to turn it on. The `tls` directive for example doesn't allow `tls on` -- just `tls off`. I think the only variant of this certificate_transparency line that we should allow is `ct off`.

It also means you could drop one of the test functions below.

caddytls/setup_test.go
```diff
@@ -358,6 +358,58 @@ func TestSetupParseWithOneTLSProtocol(t *testing.T) {
 	}
 }
 
+func TestSetupParseCertificateTransparencyDefault(t *testing.T) {
+	// This isn't must_staple specific, we just need something in that block.
```
mholt: Haha, I guess this is kind of a funny quirk of testing this directive. Sorry about that.

elcore: You can move the test to 'TestSetupParseBasic'

alex:
Ok, did all the config updates as suggested; going to defer to you on some of these plumbing tasks.

alex:
If anyone wants to see a deployed version of this, https://caddy.dysfunctional-testing.com/

One thing, ATM this prints a handful of warnings about various logs rejecting submissions, this is kinda annoying/just noise for most users. We should find a way to filter out ignorable errors, but I'm not sure what the right way is.

alex:
Ok, log annoyance dealt with.

mholt:
Now that I'm ready to revisit this, where does this PR stand? I see a couple TODOs but I'm trying to get a gauge for how serious/urgent they are. Off-hand, they don't seem particularly urgent, but definitely nice to have or at least worth considering in the longer run.

It looks like most of my TODO list from above still stands? (Which is fine! I can work on it -- not sure how concurrently with the proxy rewrite starting soon.)

Anyway, if there's anything in particular I should look for, let me know. I'll be reviewing this over the next little while though.

alex:
Both TODOs that are left are related to life-cycle management; I'll defer to you on that.

lenovouser:
Is it possible to manually specify SCT's too? Or is that for another PR?

mholt:
@lenovouser Why would you want to do it manually? The point here is to automate it. We don't want people doing this manually.

lenovouser:
@mholt for non-letsencrypt certificates

alex:
It should work fine with any publicly trusted certificate.

lenovouser:
@alex ah okay, so if I manually specify a lets say digicert / comodo / etc. certificate it will generate the `.sct`'s for it too? What about the log servers? Some only accept Let's Encrypt certificates, some don't accept Let's Encrypt certificates. (One of the Google servers does only accept LE certs, the others from Google don't accept LE certs IIRC).

By the way, here is the most extensive list of current certificate transparency servers I have found: [ct.grahamedgecombe.com](https://ct.grahamedgecombe.com/). Might be useful.

alex:
It should handle that case fine -- this code will automatically grab Google's list of trusted log servers and attempt to submit your certificate to all of them.

mholt:
@alex I'm afraid I won't have the time to pick up the rest of this myself, as I'm still working on the proxy rewrite and some other things. HOWEVER, I will be happy and able to guide you (or anyone else working on this) -- perhaps quite in detail if needed -- in finishing this, especially as it relates to Caddy's internals.

I know you're busy too, but if some time does crop up, I'm totally willing to help hit this home. You know how to reach me on Slack.

mholt:
(Incidentally, @alex, if merged, this feature would merit its own blog post on the Caddy blog. I could definitely write up the Caddy side of it, and you could co-author as well, to talk about the state of CT and what's coming.)

alex:
I'll see if I can make some time, maybe next weekend. For myself, here are the remaining TODOs:

- [ ] Update SCT server list every ~24 hours? This can be done in maintain.go and can call a function in certificate_transparency.go, for example.
- [ ] Cache the list of CT servers so we can check if it changed.
- [ ] Write SCTs to disk.
- [ ] If the CT server list changes, replace all SCTs.
- [ ] If the SCTs in the cert or OCSP response are not trusted, don't use them

alex:
I got all but two of the TODOs from that list. I don't think either of them need to be blocking though.

I haven't performed an integration test of this code yet, so let me know if that's something you'd like me to do (it should probably be done by someone before this lands!)

alex:
Ok, performed some integration testing and everything appears to work swimmingly!

elcore:
Hello @alex,

you should add some linters to your local environment. Furthermore, you should test before you commit. It makes a lot of things easier!

I can recommend VS Code (https://code.visualstudio.com/) with the Go extension (https://marketplace.visualstudio.com/items?itemName=lukehoban.Go)

alex:
My local go environment is... a mess. This is exacerbated by the fact that integration testing to verify that it actually _works_ requires me to move the code over to my server which has DNS records. (Right now I'm doing this by 1) Push to github from laptop, 2) Pull from Github on server). Sorry for the noise!

It's all working now though!

elcore:
> My local go environment is... a mess. This is exacerbated by the fact that integration testing to verify that it actually works requires me to move the code over to my server which has DNS records. (Right now I'm doing this by 1) Push to github from laptop, 2) Pull from Github on server). Sorry for the noise!

>It's all working now though!

No problem at all! 😄

caddytls/certificates.go
```diff
@@ -178,7 +179,19 @@ func makeCertificate(certPEMBlock, keyPEMBlock []byte) (Certificate, error) {
 	if err != nil {
 		log.Printf("[WARNING] Stapling OCSP: %v", err)
 	}
-
+	if certificateTransparency {
+		logs, err := getTrustedCTLogs()
```
mholt: It would be nice if we could get this at startup and then cache it for some time, so that each time we load a certificate (which might even be during TLS handshakes), we don't have to do a GET request and rely on that succeeding each time.

caddytls/certificate_transparency.go
```diff
@@ -0,0 +1,395 @@
...
+		return nil, fmt.Errorf("HTTP response error: %d", response.StatusCode)
+	}
+	list := logList{}
+	// Limit the amount we read to 10MB to defend against a DoS. Currently this
+	// file is ~2.5KB so this leaves us plenty of breathing room.
+	err = json.NewDecoder(io.LimitReader(response.Body, 10*1024*1024)).Decode(&list)
+	if err != nil {
+		return nil, err
+	}
+
```
mholt: What if we wrap the code from the start of this function to this point in another function that loads the cached `logList` or else gets a new one, if the cached one is old. That way we just call `list, err := getLogList()` and most of the time it won't have the overhead of a network request.

caddytls/maintain.go
```diff
@@ -318,4 +329,43 @@ func freshOCSP(resp *ocsp.Response) bool {
...
+			scts, err := getSCTSForCertificateChain(cert.Certificate.Certificate, newLogs)
+			if err != nil {
+				log.Printf("[WARNING] Fetching SCTs: %v", err)
+			} else {
+				updates[name] = scts
+			}
+		}
+		certCacheMu.RUnlock()
+
+		certCacheMu.Lock()
```
mholt: :+1: for separating the read and update locks.

caddytls/certificates.go
```diff
@@ -195,6 +208,8 @@ func fillCertFromLeaf(cert *Certificate, tlsCert tls.Certificate) error {
 		return err
 	}
 
+	cert.Leaf = leaf
```
mholt: 🤔 I realize you reference the Leaf cert elsewhere, but I'm wondering if setting this field has any other side-effects or implications? Or was it a mistake that we didn't do this before?

francislavoie:
@alex any chance of finishing this up soon?

alex:
Probably not, at this point folks should just wait for Let's Encrypt to roll out embedded SCTs in their certs.

francislavoie:
Okay. I guess we should close this then? @mholt @alex

lenovouser:
What about non-le certificates=

mholt:
Due to mutual consensus, I am going to close this PR. Certificate transparency infrastructure has matured and improved over the last year or so, and the industry is moving toward requiring CAs to embed SCTs into certificates. That is ultimately a better solution. But always let it be known that Alex chose to experiment with auto-CT-in-web-servers with a pull request to Caddy. :) I think we all learned quite a bit in this process, and we're looking forward to a more robust future of HTTPS all the same.

Thanks again, Alex, for all your work on this!

PR #1595, opened 2017-04-18
Add hostname template action

jannickfahlbusch:
This adds the ability to display the remote hostname of the visitors IP with template actions.

### DIFF
--- a/caddyhttp/httpserver/context.go
+++ b/caddyhttp/httpserver/context.go
@@ -63,6 +63,18 @@ func (c Context) Header(name string) string {
 	return c.Req.Header.Get(name)
 }
 
+// Hostname gets the (remote) hostname of the client making the request.
+func (c Context) Hostname() string {
+	ip := c.IP()
+
+	hostnameList, err := net.LookupAddr(ip)
+	if err != nil || len(hostnameList) == 0 {
+		return c.Req.RemoteAddr
+	}
+
+	return hostnameList[0]
+}
+
 // Env gets a map of the environment variables.
 func (c Context) Env() map[string]string {
 	osEnv := os.Environ()
--- a/caddyhttp/httpserver/context_test.go
+++ b/caddyhttp/httpserver/context_test.go
@@ -244,6 +244,37 @@ func TestHeader(t *testing.T) {
 	}
 }
 
+func TestHostname(t *testing.T) {
+	context := getContextOrFail(t)
+
+	tests := []struct {
+		inputRemoteAddr  string
+		expectedHostname string
+	}{
+		// Test 0 - ipv4 with port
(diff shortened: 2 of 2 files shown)


caddyhttp/httpserver/context.go
```diff
@@ -63,6 +63,18 @@ func (c Context) Header(name string) string {
...
+// Hostname gets the (remote) hostname of the client making the request.
+func (c Context) Hostname() string {
+	ip := c.IP()
+
+	address, err := net.LookupAddr(ip)
+	if err != nil {
+		return c.Req.RemoteAddr
+	}
+
+	return address[0]
```
francislavoie: Is there a chance that this ever returns an empty array? Maybe check `err != nil && len(address) > 0`?

jannickfahlbusch: I never seen `net.LookupAddr` return an empty slice.

However, the docs doesn't mention this case, so I'll add the check.

francislavoie: Yeah, I did a bunch of digging and I couldn't find an appropriate example proving that it would never return empty. Just being a bit paranoid just in case!

jannickfahlbusch:
@francislavoie  I added the length-check and renamed the variable from `address` to `hostnameList`.

mholt:
Hey Jan, thanks for the PR! What's the use case for this? Most clients won't have a registered hostname, and DNS lookups are very slow.

jannickfahlbusch:
I wanted to have a simple page on which I get my current IP and hostname (like myip.is). 

At least here in Germany most of the users get IPs with a hostname registered to it. 

Am 19. April 2017 22:34:36 MESZ schrieb Matt Holt <notifications@github.com>:

francislavoie:
@jannickfahlbusch Do German ISPs provide hostnames for their users? That's interesting.

elcore:
> @jannickfahlbusch Do German ISPs provide hostnames for their users? That's interesting.

@francislavoie Yes some ISPs do it.

JoshHarmon:
I don't know about Germany, but I can tell you that Comcast does here in the U.S. (Pattern: c-aa-bb-cc-dd.hsd1.YOUR_STATE.comcast.net). I've also seen hostnames on TWC (U.S.) customers, Verizon Wireless users, and other ISPs over the years (vaguely seem to remember Charter also doing so).

To me, this means there's a significant group of people who this feature *could* be useful to. I'm not sure how useful it would be in practice, but figured I'd just chime in to say that there are definitely visitors that *could* benefit from the feature.

jannickfahlbusch:
You're welcome :) 

Thanks for merging this in!

Issue #1592, opened 2017-04-18
No way to disable "insecure CA URL" error in some private networks.


ntrrgc:
### 1. What version of Caddy are you running (`caddy -version`)?
Caddy 0.9.5

### 2. What are you trying to do?
Test my server deployment with virtual machines before deploying it in the public Internet.

### 4. How did you run Caddy (give the full command and describe the execution environment)?

Systemd service inside a CentOS 7 virtualized with libvirt in a Linux host. This same host runs a [Boulder](https://github.com/letsencrypt/boulder) instance, running in Docker, whose port 4000 is forwarded to the host and made accessible to the CentOS 7 server.

```
/usr/local/bin/caddy -log stdout -agree=true -conf=/etc/caddy/Caddyfile -root=/var/tmp -ca http://192.168.122.1:4000/directory
```

### 6. What did you expect to see?

There should be a way to use this "fake" CA server so that I can use my servers with a self signed CA.

### 7. What did you see instead (give full error messages and/or log)?

```
caddy[10172]: Activating privacy features...2017/04/18 12:22:54 http://192.168.122.1:4000/directory: insecure CA URL (HTTPS required)
```

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

Set the `-ca` parameter in Caddy to any URL you want to test. The error, if any, is shown before any connection is made.

The actual problem is here in the code: https://github.com/mholt/caddy/blob/6bc3e7536e915c0f7dd0c22be36201760e9d816d/caddytls/client.go#L64-L66

Non HTTPS CAs are only accepted if the host is a loopback address or it starts with `10.` (one of the several prefixes used for private networks in IPv4, but not the one I'm using). I could workaround this in several ways, for instance, running a forwarding daemon in localhost like this:

```
ncat -k -l 4000 -e "$(which ncat) 192.168.122.1 4000"
```

But it is quite hacky. It would be nicer if just a configuration option was added to remove this check, or at least private network ranges were checked properly.

mholt:
Hey @ntrrgc thanks for the issue! How about if we add "192.168." to the list of allowed-insecure prefixes?

ntrrgc:
@mholt It would be an incomplete fix, as those IPv4 addresses between 172.16.0.0 and 172.31.255.255 (inclusive) are also private, so are IPv6 in the `fc00::/7` range plus arguably the link-local range (`fe80::/10`). Also, the host may not be an IP address but a hostname that resolves to one of these IPs.

[...]

PR #1599, opened 2017-04-19
Disable warning for insecure CA if located on private network.

cez81:
My first Caddy PR :)

Fix for #1592. It doesn't handle hostnames that resolves to private IPs.

### DIFF
--- a/caddy_test.go
+++ b/caddy_test.go
@@ -61,6 +61,48 @@ func TestIsLoopback(t *testing.T) {
 	}
 }
 
+func TestIsPrivateNetwork(t *testing.T) {
+	for i, test := range []struct {
+		input  string
+		expect bool
+	}{
+		{"10.0.0.0", true},
+		{"10.0.0.5", true},
+		{"10.255.255.255", true},
+		{"10.0.0.5:1234", true},
+		{"172.16.0.0", true},
+		{"172.19.1.1", true},
+		{"172.31.255.255", true},
+		{"192.168.0.0", true},
+		{"192.168.1.12", true},
+		{"192.168.255.255", true},
+		{"fc00::1", true},
+		{"fd12:3456:789a:1::1", true},
+		{"fd12:3456:789a:1::1:1234", true},
+		{"fd3e:00c3:62a6:e2b3::", true},
+		{"fdff:ffff:ffff:ffff:ffff:ffff:ffff:ffff", true},
+		{"example.com", false},
+		{"localhost", false},
+		{"localhost:1234", false},
+		{"localhost:", false},
+		{"127.0.0.1", false},
+		{"127.0.0.1:443", false},
+		{"127.0.1.5", false},
+		{"12.7.0.1", false},
+		{"[::1]", false},
+		{"[::1]:1234", false},
+		{"::1", false},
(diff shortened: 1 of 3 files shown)


francislavoie:
@mholt it helps to simplify IPv6, otherwise we'd need separate logic for IPv6 than IPv4. The address spaces are listed here: https://en.wikipedia.org/wiki/Private_network, for reference

cez81:
Checking the 172.16. -> 172.31. are all internal. But yh could be done with a number of ifs

caddy_test.go
```diff
@@ -61,6 +61,48 @@ func TestIsLoopback(t *testing.T) {
...
+		input  string
+		expect bool
+	}{
+		{"10.0.0.0", true},
+		{"10.0.0.5", true},
+		{"10.255.255.255", true},
+		{"10.0.0.5:1234", true},
+		{"172.16.0.0", true},
+		{"172.19.1.1", true},
+		{"172.31.255.255", true},
```
mholt: Add tests for the edge cases on both sides of the boundary (both `true` and `false` sides).

caddy_test.go
```diff
@@ -61,6 +61,48 @@ func TestIsLoopback(t *testing.T) {
...
+		{"fd12:3456:789a:1::1:1234", true},
+		{"fd3e:00c3:62a6:e2b3::", true},
+		{"fdff:ffff:ffff:ffff:ffff:ffff:ffff:ffff", true},
+		{"example.com", false},
+		{"localhost", false},
+		{"localhost:1234", false},
+		{"localhost:", false},
+		{"127.0.0.1", false},
+		{"127.0.0.1:443", false},
+		{"127.0.1.5", false},
```
mholt: Localhost and 127.* are not a private/internal network? I am thinking I disagree...

francislavoie: Technically, loopback is different than private. We already have `IsLoopback` here: https://github.com/cez81/caddy/blob/1f928fc764da0b88ffa5117bdd2853aa20bd4446/caddy.go#L756 so we can just do `IsLoopback || IsPrivateNetwork` depending on the use-case.

mholt: Except I feel like private/internal is a _superset_ of loopback... in other words, almost feels like we should add a check for IsLoopback into this function.

francislavoie: Maybe this is a weird idea, but maybe we make this function `IsPrivate` (for naming consistency, we're not calling the other `IsLoopbackNetwork`) and make another helper called `IsPrivateNetwork` which gives us the union. I think there's value in having `IsPrivate` its own thing for potential future use.

mholt: Well, both are for networks, so how about this instead:

- `IsLoopback` (keep)
- `IsInternal` (rename this new function)
- and then we just use `IsLoopback() && IsInternal()` when we need both.

Make sure the godoc comment for IsInternal states that it does not include loopback addresses.

francislavoie: It would be `||` not `&&`, it can never be both loopback and internal 😉 
Sounds good to me.

mholt:
Looks good, @cez81! Thanks for contributing, and nice work!

Issue #1604, opened 2017-04-21
v0.10 proxy without does not work


lilee:
### 1. What version of Caddy are you running (`caddy -version`)?
v 0.10

### 2. What are you trying to do?
in `proxy` directive use `without` to trim prefix

### 3. What is your entire Caddyfile?
like this
```text
proxy /help localhost:60165 {
    without /help
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
caddy -conf <Caddyfile>

### 5. Please paste any relevant HTTP request(s) here.

curl myhost.com/help

### 6. What did you expect to see?

see what `curl localhost:60165` returns

### 7. What did you see instead (give full error messages and/or log)?

404 Page not found

the `without` prefix dose not work. the request uri is /help not /

sgsfak:
I had the exact same problem noticed some minutes ago after the upgrade to 0.10. The 0.9.5 release does not have this bug, so I reverted to it.

tw4452852:
Hmm, this issue is introduced by [this commit](https://github.com/mholt/caddy/commit/344017dc21e80380447eedf515bbc3389451a672).
My fault for not taking `without` into account...

jazoom:
Yup, just got bitten by this

jgsqware:
Same here, with this Caddyfile:

```
http://www.lam.local {
  log stdout
  proxy /mattermost gitlab:80 {
    without mattermost
    header_upstream Host mattermost.lam.local
    header_upstream X-Real-IP {remote}
    header_upstream X-Forwarded-For {remote}
    header_upstream X-Forwarded-Proto {scheme}
  }
  tls off
}
``` 

Caddy log:

```
caddy_1   | 172.20.0.1 - - [24/Apr/2017:14:51:40 +0000] "GET /mattermost HTTP/1.1" 200 1382
```

Gitlab Log:

```
lab_1  | 172.20.0.3 - - [24/Apr/2017:14:51:40 +0000] "GET /mattermost HTTP/1.1" 200 1394 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.138 Safari/537.36 Vival
di/1.8.770.56"
```

tw4452852:
@jazoom This issue should be fixed on the latest master, I checked 0775f91, it works as expected.

[...]

PR #1605, opened 2017-04-21
proxy: take original URL path into account when remove prefix

tw4452852:
fix issue #1604


mholt:
Ah, there we go -- that looks like the right behavior. Thanks for the quick fix @tw4452852! (And for the review, elcore.)

PR #1612, opened 2017-04-23
dist: better way to get systemd version

emersion:
Use `systemctl --version` instead of a distribution-specific command.


mholt:
Thank you @emersion (and @francislavoie for the review)!

PR #1613, opened 2017-04-24
Context: add Push action

tw4452852:
During writing [my blog caddy plugin](https://github.com/tw4452852/caddy-totorow), I find that there is no way to add push links in template files. I think this is very useful to indicate server push in templates when combining with `push`.
For example, something in `index.html` looks like: 

```
...
<script src="/public/js/jquery.js" type="text/javascript" charset="utf-8"></script>
<link rel="stylesheet" type="text/css" href="/public/css/style.css" />
{{.AddLink /public/js/jquery.js}}
{{.AddLink /public/css/style.css}}
...
```

Client will receive server push for `/public/js/jquery.js` `/public/css/style.css`

### DIFF
--- a/caddyhttp/httpserver/context.go
+++ b/caddyhttp/httpserver/context.go
@@ -29,6 +29,20 @@ type Context struct {
 	Req  *http.Request
 	URL  *url.URL
 	Args []interface{} // defined by arguments to .Include
+
+	// just used for adding preload links for server push
+	responseHeader http.Header
+}
+
+// NewContextWithHeader creates a context with given response header.
+//
+// To plugin developer:
+// The returned context's exported fileds remain empty,
+// you should then initialize them if you want.
+func NewContextWithHeader(rh http.Header) Context {
+	return Context{
+		responseHeader: rh,
+	}
 }
 
 // Include returns the contents of filename relative to the site root.
@@ -408,6 +422,15 @@ func (c Context) RandomString(minLen, maxLen int) string {
 	}
 
 	return string(result)
+}
+
+// Push adds a preload link in response header for server push
+func (c Context) Push(link string) string {
+	if c.responseHeader == nil {
+		return ""
+	}
+	c.responseHeader.Add("Link", "<"+link+">; rel=preload")
+	return ""
 }
 
 // buffer pool for .Include context actions
--- a/caddyhttp/httpserver/context_test.go
+++ b/caddyhttp/httpserver/context_test.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


caddyhttp/httpserver/context.go
```diff
@@ -410,6 +411,15 @@ func (c Context) RandomString(minLen, maxLen int) string {
 	return string(result)
 }
 
+// AddLink adds a link in response for future server push
+func (c Context) AddLink(link string) string {
```
mholt: Does this have to return something?

caddyhttp/httpserver/context.go
```diff
@@ -410,6 +411,15 @@ func (c Context) RandomString(minLen, maxLen int) string {
...
+	if c.Res == nil {
+		return ""
+	}
+	c.Res.Header().Add("Link", "<"+link+">")
```
mholt: Wouldn't you need rel=preload? See https://www.w3.org/TR/preload/#server-push-http-2

caddyhttp/httpserver/context.go
```diff
@@ -26,6 +26,7 @@ import (
 // Context is the context with which Caddy templates are executed.
 type Context struct {
 	Root http.FileSystem
+	Res  http.ResponseWriter
```
mholt: This gets exposed to the templates, so just with this line, templates can now do:

```
{{.Res.Header.Add "Field" "Value"}}
```

I can see why having a convenience method for push would be useful though:

```
{{.Push "/filename.ext"}}
```

mholt: However, by exposing this, it also lets people call Write and WriteHeader on the ResponseWriter from a template. I'm not sure if this is desirable, in fact, probably not. What if we just expose the Header field instead of the whole ResponseWriter?

tw4452852: I like your idea and we could get even better: don't export response's Header at all. This will forbid templates manipulating response header (e.g. Add, Set, Del). See my new patchset.

caddyhttp/httpserver/context.go
tw4452852: I think so according to [this](https://golang.org/pkg/text/template/#hdr-Pipelines)

> The output of a command will be either one value or two values, the second of which has type error. If that second value is present and evaluates to non-nil, execution terminates and the error is returned to the caller of Execute.

caddyhttp/httpserver/context.go
tw4452852: Done.

mholt:
Is it actually bad if the template has access to the headers? I'm kind of thinking that wouldn't be a problem and could actually be useful. And it would simplify a lot of the code (get rid of the NewContextWithHeader constructor, etc).

tw4452852:
Just concerning the probably bug after export headers to templates, through it can give templates more control over the response header. But it also introduces the possibility of the weird bug which is hard to find according to the `Caddyfile` only, we need template files as well if user add some headers in it. So I think we should limit the functionality of templates, just export what we claims. 

Anyway this is only myself opinion, it's up to you @mholt .

mholt:
Sounds good to me. Thanks! {{.Push}} template action may be useful.

PR #1616, opened 2017-04-25
Don't sanitize path when it contains escaped parts

tw4452852:
fix issue #1582

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -415,6 +415,12 @@ func sanitizePath(r *http.Request) {
 	if r.URL.Path == "/" {
 		return
 	}
+
+	// skip if there are some escaped forms in the original path (see issue 1582)
+	if r.URL.RawPath != "" {
+		return
+	}
+
 	cleanedPath := CleanPath(r.URL.Path)
 	if cleanedPath == "." {
 		r.URL.Path = "/"


francislavoie:
I'm thinking this probably needs some tests? What are the failure cases that this solves?

mholt:
I think I'm gonna work on this issue more broadly. I'll probably replace the sanitizePath function with a simple string replace of "../" and "./" values, as well as experiment with the actual harm any extra slashes can do. I'm basically going to explore simplifying this whole thing, maybe even to the point where we don't need any sanitizing or Context values to hold the "original" URL.

caddyhttp/httpserver/server.go
```diff
@@ -415,6 +415,12 @@ func sanitizePath(r *http.Request) {
 	if r.URL.Path == "/" {
 		return
 	}
+
+	// skip if there are some escaped forms in the original path (see issue 1582)
+	if r.URL.RawPath != "" {
+		return
```
mhoIt: This should give an error because caddy is stupid

elcore: > This should give an error because caddy is stupid

????

mholt:
FYI, I'm working on a larger change locally right now and hope to push a PR up tomorrow or early next week! It builds on your work with paths etc. recently, but it takes a new approach to handling them that I hope will simplify some things.

tw4452852:
Wait for your change :smile:

mholt:
@tw4452852 Okay, change is up at #1633. It removes the `sanitizePath()` function entirely, so I will close this PR now. Thanks for looking at it, and for your efforts. 😃 I appreciate it!

Issue #735, opened 2016-04-09
High GC pressure causing slowdowns (was: gzip awfully slow)


joushou:
#### 1. What version of Caddy are you running (`caddy -version`)?

Caddy 0.8.2
#### 2. What are you trying to do?

Simply running a comparison between my own, simple Go web server and Caddy in various modes.

My tests were run with https://github.com/rakyll/boom:

``` text
boom -n 10000 -c 100 http://mysite/index.html
```
#### 3. What is your entire Caddyfile?

``` text
0.0.0.0:8080
root /somewhere
log /var/log/caddy.log
gzip
```
#### 4. How did you run Caddy (give the full command and describe the execution environment)?

./caddy -conf Caddyfile
#### 5. What did you expect to see?

A total execution time of the test somewhere around 12s±5%, which is the speed it takes for my own Go server to respond to the test when set up to serve from disk with gzip enabled (which SHA256 and gzips the content on every request).
#### 6. What did you see instead (give full error messages and/or log)?

A total execution time of 39.6s±5%. That's a lot. The same run against Caddy without gzip enabled on the server takes 6.8s±5% in comparison. That's 3 times slower than my server, and while I suspected that my server would be faster, I had thought a few percent, not factor 3.

I don't use Caddy myself, but I thought I'd give you a heads up about my findings. I have not investigated why Caddy is so slow in this case, but it should be possible to get similar speeds to my server, which isn't magic when serving data from disk.

mholt:
Can you run your test again without the `log` directive? I feel like that's going to add noise to your tests.

joushou:
A single run with gzip, but without access logs: 37.72s. Also note that the non-gzip run at ~6.5s also included access logs.

joushou:
(But as a performance test against my own server, I wanted it to include access logs. Otherwise, it wasn't comparing real setups.)

joushou:
Something I just noticed: Caddy uses level 9 by default, not the gzip packages default. For the 10KB input file, Caddy outputs 4498 bytes, while the default level my own server uses output 4563, which is quite an almost unnoticeable difference. I'll try to run with Caddy set to a lower level, although the docs don't seem to state exactly what the default level is...

mholt:
Thanks for trying a few things. We'll look into this soon, unless somebody wants to dive deeper sooner. (I haven't spent a lot of time in optimizations yet since I'm still trying to get the code base stable.)

[...]

PR #1618, opened 2017-04-25
gzip: pool gzip.Writer to reduce allocation

tw4452852:
```
benchmark           old ns/op     new ns/op     delta
BenchmarkGzip-4     2219756       2123427       -4.34%

benchmark           old allocs     new allocs     delta
BenchmarkGzip-4     25             8              -68.00%

benchmark           old bytes     new bytes     delta
BenchmarkGzip-4     959566        146507        -84.73%
```

### DIFF
--- a/caddyhttp/gzip/gzip_test.go
+++ b/caddyhttp/gzip/gzip_test.go
@@ -117,3 +117,37 @@ func nextFunc(shouldGzip bool) httpserver.Handler {
 		return 0, nil
 	})
 }
+
+func BenchmarkGzip(b *testing.B) {
+	pathFilter := PathFilter{make(Set)}
+	badPaths := []string{"/bad", "/nogzip", "/nongzip"}
+	for _, p := range badPaths {
+		pathFilter.IgnoredPaths.Add(p)
+	}
+	extFilter := ExtFilter{make(Set)}
+	for _, e := range []string{".txt", ".html", ".css", ".md"} {
+		extFilter.Exts.Add(e)
+	}
+	gz := Gzip{Configs: []Config{
+		{
+			RequestFilters: []RequestFilter{pathFilter, extFilter},
+		},
+	}}
+
+	w := httptest.NewRecorder()
+	gz.Next = nextFunc(true)
+	url := "/file.txt"
+	r, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		b.Fatal(err)
+	}
+	r.Header.Set("Accept-Encoding", "gzip")
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err = gz.ServeHTTP(w, r)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+}
--- a/caddyhttp/gzip/gzip_test.go
+++ b/caddyhttp/gzip/gzip_test.go
@@ -1,6 +1,7 @@
 package gzip
 
 import (
+	"compress/gzip"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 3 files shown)


caddyhttp/gzip/setup.go
```diff
@@ -119,3 +124,49 @@ func gzipParse(c *caddy.Controller) ([]Config, error) {
...
+	writerPool             = map[int]*sync.Pool{}
+	defaultWriterPoolIndex int
+)
+
+func initWriterPool() {
+	var i int
+	for i = gzip.BestSpeed; i <= gzip.BestCompression; i++ {
+		writerPool[i] = &sync.Pool{
+			New: func() interface{} {
+				w, _ := gzip.NewWriterLevel(ioutil.Discard, i)
```
nussjustin: There's a bug here.

`i` is closed over, so all pools (exception the default writer pool) will use the last value of `i` when called, which is 10 (gzip.BestCompression+1). Passing 10 to gzip.NewWriterLevel returns an error and a nil writer making the call to Reset in getWriter panic.

See https://play.golang.org/p/W9Chscrkme

Just declare a new variable `l := i` inside the loop and use it in Pool.New. This will fix the bug.

See https://play.golang.org/p/__42IlPQI8

tw4452852: Ah, Thanks for finding this bug. I have just updated a new patchset, please take a look.

caddyhttp/gzip/setup.go
```diff
@@ -119,3 +124,52 @@ func gzipParse(c *caddy.Controller) ([]Config, error) {
...
+	}
+}
+
+func getWriter(level int) *gzip.Writer {
+	index := defaultWriterPoolIndex
+	if level >= gzip.BestSpeed && level <= gzip.BestCompression {
+		index = level
+	}
+	w := writerPool[index].Get().(*gzip.Writer)
+	w.Reset(ioutil.Discard)
```
mholt: I DID NOT KNOW RESET() WAS A THING ON GZIP WRITERS. THIS IS AMAZING

👏

caddyhttp/gzip/gzip_test.go
```diff
@@ -77,6 +78,22 @@ func TestGzipHandler(t *testing.T) {
...
+		r, err := http.NewRequest("GET", "/file.txt", nil)
+		if err != nil {
+			t.Error(err)
+		}
+		r.Header.Set("Accept-Encoding", "gzip")
+		_, err = gz.ServeHTTP(w, r)
+		if err != nil {
+			t.Error(err)
+		}
+	}
```
mholt: :+1: testing all the compression levels. Was it intentional to start at `0` instead of `gzip.BestSpeed`?

caddyhttp/gzip/setup.go
```diff
@@ -119,3 +124,52 @@ func gzipParse(c *caddy.Controller) ([]Config, error) {
 
 	return configs, nil
 }
+
+// pool gzip.Writer according to compress level
+// so we can reuse allocations over time
+var (
+	writerPool             = map[int]*sync.Pool{}
```
mholt: :100: Huge win; I didn't realize these could be reused.

caddyhttp/gzip/setup.go
```diff
@@ -16,6 +19,8 @@ func setup(c *caddy.Controller) error {
 		return err
 	}
 
+	initWriterPool()
```
mholt: `setup` can be called multiple times, even while a server is running, for example when it gets reloaded. This will run while the server is already up and running. Either the writer pool map should be guarded with a RWMutex -OR- it should only be initialized once, like in `init()` instead of `setup`.

caddyhttp/gzip/setup.go
mholt: Are we sure this map doesn't need a mutex? Requests are constantly reading from it to access the pools.

caddyhttp/gzip/gzip_test.go
tw4452852: Yes. Starting from `0` is intentional.

caddyhttp/gzip/setup.go
tw4452852: Indeed, `init()` is good place.

caddyhttp/gzip/setup.go
tw4452852: I don't think we need a lock here, we only write this map during initialization and then only read.

caddyhttp/gzip/gzip_test.go
tw4452852: Yes, starting from `0` is intentional.

caddyhttp/gzip/setup.go
mholt: I think you're right. The writes happen inside the underlying pool, not the map values directly.

mholt:
Great work @tw4452852 :) I think this will help with #735 -- probably enough that I will close that (unless they can show it's still a significant problem and is the same problem).

elcore:
@tw4452852 awesome job 😄

Issue #1610, opened 2017-04-22
Bad Rewrite parameters break fastcgi matching


tobya:
### 1. What version of Caddy are you running (`caddy -version`)?

0.10 but also present in 0.9.5

### 2. What are you trying to do?

rewrite a url to a php page

### 3. What is your entire Caddyfile?
```text
basil.kinchik.ie:80 {
  
  #breaks php fastcgi - renders as html
  rewrite /admin wp-admin/admin.php
  
  root  d:\caddy\webroot\basil
 
  fastcgi /  localhost:49249 php
  startup d:\php7\php-cgi.exe -b 49249 &
}
```

It works correctly if the rewrite rule is replaced with 

     rewrite /admin /wp-admin/admin.php

note the `/` before `/wp-admin/admin.php`  

### 4. How did you run Caddy (give the full command and describe the execution environment)?

caddy -log stdout

### 5. Please paste any relevant HTTP request(s) here.

    http://basil.kinchik.ie/admin

### 6. What did you expect to see?

http://basil.kinchik.ie/wp-admin/index.php executed as php and returned as html

### 7. What did you see instead (give full error messages and/or log)?

http://basil.kinchik.ie/wp-admin/index.php rendered as text 

eg.  

    <?php
     echo '<P>this is a php page';

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

Use the caddyfile provided.  Ensure the file wp-admin/index.php exists and contains php code.

There should never be a situation where a php file is served without going through the php engine this is a security risk.  It would be very easy for a configuration to slip through that caused this.

tobya:
I found the place where the mismatching  happens

https://github.com/mholt/caddy/blob/master/caddyhttp/fastcgi/fastcgi.go#L40

I will try to submit a PR

mholt:
Thanks Toby! Be careful though, we just need to decide what rewriting a URI to `wp-admin/admin.php` really means (what's the final result given a base path of `/` vs `/foo` vs `/foo/` etc)) and see if this is actually a bug in Caddy or just user error.

tobya:
@mholt  I am working on a very simple fix at the moment and will look at what happens when I check for a `/` at the base of any url being matched by fastcgi and add it if required.  

An alternative would be to raise a syntax error on parsing the caddyfile if rewrite to destination does not begin with `/` (all examples do).

[...]

PR #1629, opened 2017-04-27
Raise syntax error if rewrite path does not begin with / #1610

tobya:

### 1. What issue(s) is this pull request related to?

 Issue #1610 
### 2. What does this change do, exactly?

The change raises a syntax error if the rewrite path being rewritten to does not begin with a /

### 3. What documentation changes (if any) go along with this PR?

I can submit a seperate pull request to update documentation to say that rewrite paths should start with a /

### 4. Checklist

- [x] This change has comments for package types, values, and functions or non-obvious lines of code

### DIFF
--- a/caddyfile/dispenser.go
+++ b/caddyfile/dispenser.go
@@ -207,7 +207,7 @@ func (d *Dispenser) SyntaxErr(expected string) error {
 // RewritePathErr returns an error indicating the path being rewritten to
 // is missing an initial /
 func (d *Dispenser) RewritePathErr() error {
-	msg := fmt.Sprintf("%s:%d - Rewrite error: Rewrite path must begin with '/'. Provided: %s", d.File(), d.Line(), d.Val())
+	msg := fmt.Sprintf("%s:%d - Syntax error: Rewrite path must begin with '/'. Provided: '%s'", d.File(), d.Line(), d.Val())
 	return errors.New(msg)
 }
 
--- a/caddyfile/dispenser.go
+++ b/caddyfile/dispenser.go
@@ -204,13 +204,6 @@ func (d *Dispenser) SyntaxErr(expected string) error {
 	return errors.New(msg)
 }
 
-// RewritePathErr returns an error indicating the path being rewritten to
-// is missing an initial /
-func (d *Dispenser) RewritePathErr() error {
-	msg := fmt.Sprintf("%s:%d - Syntax error: Rewrite path must begin with '/'. Provided: '%s'", d.File(), d.Line(), d.Val())
-	return errors.New(msg)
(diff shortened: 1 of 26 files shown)


mholt:
Hey Toby, thanks for this PR! I will look at it but I'm concurrently working on a change that touches the rewrite middleware slightly, so I will submit my PR of that before I act on this one, to make sure it doesn't conflict (and if it does, we can work it out).

caddyfile/dispenser.go
```diff
@@ -204,6 +204,13 @@ func (d *Dispenser) SyntaxErr(expected string) error {
 	return errors.New(msg)
 }
 
+// RewritePathErr returns an error indicating the path being rewritten to
+// is missing an initial /
+func (d *Dispenser) RewritePathErr() error {
```
abiosoft: Since `rewrite` is simply a directive, I do not think it is best to have it's error as part of the dispenser. This looks out of place to me.

mholt: I agree. Why not just use `d.Err()`?

caddyhttp/rewrite/setup.go
```diff
@@ -90,14 +94,20 @@ func rewriteParse(c *caddy.Controller) ([]httpserver.HandlerConfig, error) {
...
 				return nil, c.ArgErr()
 			}
+
 			if rule, err = NewComplexRule(base, pattern, to, ext, matcher); err != nil {
 				return nil, err
 			}
 			rules = append(rules, rule)
 
-		// the only unhandled case is 2 and above
+		// the only unhandled case is 2 and above 'from to'
```
mholt: This comment is confusing; clarify, maybe this:

> handle case of 2 arguments: "from to"

caddyfile/dispenser.go
tobya: hmm, well all the other parsing/syntax errors seem to be raised in dispenser.go.  

I can use `d.Err()` but I don't get access to `d.Val()` which would be nice.  I can just raise a syntax error if we don't want new errors created.

>  Syntax error: Unexpected token 'wpadmin/admin.php', expecting '/wpadmin/admin.php'

but IMHO it is a little too subtle better to tell the user exactly what the issue is.

Could I create another general purpose method to raise an error that includes `d.Val()` as a parameter?  I will see if i can get the output correct.

mholt:
@tobya I can get this out in the 0.10.1 release today or tomorrow. Let me know if you want help finishing up the last couple changes. :+1:

tobya:
@mholt those two changes made now.

mholt:
Fabulous, @tobya - thank you!

Issue #1624, opened 2017-04-26
Rewrite broken in 0.10?


jleclanche:
The following caddyfile's `rewrite` directive is ignored in 0.10.0 (I reverted to 0.9.5 and confirmed it working):

```
api.example.com {
        root /opt/example.com/latest/source
        redir / /v1 307 {
                if {path} is /
        }
        rewrite {
                r ^/v1(.*)
                to /api/{path}
        }
        gzip
        proxy / [::1]:8000 {
                transparent
                except /static /robots.txt /favicon.ico
        }
}
```

I can't figure it out. I can't get it to match at all, it's like it's being completely ignored. Tried a bunch of variations including a plain `rewrite / /foo` and it's just like it's commented out.

jleclanche:
Download url, in case it's a plugin's fault (haven't tried on vanilla yet):

`https://caddyserver.com/download/linux/amd64?plugins=http.awslambda,http.cors,http.expires,http.ratelimit,http.realip,tls.dns.cloudflare,tls.dns.gandi`

mholt:
Thanks @jleclanche -- but why did you not follow the issue template? ☹️ - we need all that information to help you.

jleclanche:
Sorry, i tried to reduce the test case as much as possible. I'm still currently trying to reduce it further :). I'm using the systemd service file to run caddy (the one shipped with the tarball).

mholt:
Thanks, it's helpful. :) How do you know rewrite is being ignored?

jleclanche:
Okay, so my reduced set up so far: I have uwsgi serving a web page at exactly `/api/v1/`.

I have the following (literal) Caddyfile right now:

```
example.com:80 {
        rewrite {
                r ^/v1(.*)
                to /api/{path}
        }
        proxy / [::1]:8000 {
                transparent
        }
}
```

If I send a `HTTP HEAD Host: example.com /v1/` to Caddy, on 0.9.5, I receive a 200. On 0.10.0, I receive a 404 (uwsgi logs tells me I tried to query unmodified `/v1/`, which is a 404 on its own).

```
163e60a943f22b185b75e66ff70585f44cf872661a88a46740fe604b7a926ee7  caddy-0.10
bd06a3b76527244cf43e4b41ad3e3273604233f88da402ff31ba6dc92ff7b4ef  caddy-0.9.5
```

Trying to remove plugins now.

jleclanche:
[...]

Issue #256, opened 2015-09-25
FastCGI: REQUEST_URI after rewrite


DenBeke:
When querying the `REQUEST_URI` environmental variable without rewrites, it correctly gives the whole request uri, but when a rewrite took place, it only returns the actual script name called...

* Without rewrite:  `REQUEST_URI`: `/index.php/apps/files/`
* With rewrite:  `REQUEST_URI`: `/index.php`

mholt:
Uh oh. Well, unless I'm missing something obvious there's currently no way to deal with this... but I have an idea.

We could make the rewrite middleware set a request header for internal Caddy use only, a header that indicates what the original URI was. Then the fastcgi middleware could check for that header and use it. Kind of a hack but a fair way to keep some simple context between handlers. What do you think @abiosoft? Is there a better way to do this?

DenBeke:
> Uh oh. Well, unless I'm missing something obvious there's currently no way to deal with this... but I have an idea.

That why I didn't even start looking into it, because it's how the middleware is designed, isn't it?

mholt:
Yep. It was a good design in theory :smile: But I think we can make it work. I'll experiment on the train today.

mholt:
@DenBeke Would you please pull the latest and try commit be6fc35326051094b3dcbd5fa7c106ae57705dc1? I don't think it broke anything but just be sure. Also see if it fixes this issue and let me know.

DenBeke:
@mholt This solved the "main error" I was experiencing. Will check again with a cleaner ownCloud installation, because I messed a lot in it's source.

DenBeke:
@mholt This fixes all problems I was experiencing.
Next up is enhancing rewrites for all pages to obtain a completely working ownCloud installation...

mholt:
@DenBeke Sounds good, I will continue that via email with you. Thanks!

PR #1633, opened 2017-04-29
httpserver/all: Clean up and standardize request URL handling

mholt:
### 1. What issue(s) is this pull request related to?
I'm hoping this will address issues with #1624, #1584, #1582, #256 (by not breaking it), and others.

### 2. What does this change do, exactly?
The HTTP server now always creates a context value on the request which
is a copy of the request's URL struct. It should not be modified by
middlewares, but it is safe to get the value out of the request and make
changes to it locally-scoped. Thus, the value in the context always
stores the original request URL information as it was received. Any
rewrites that happen will be to the request's URL field directly.

The HTTP server no longer cleans /sanitizes the request URL. It made too
many strong assumptions and ended up making a lot of middleware more
complicated, including upstream proxying (and fastcgi). To alleviate
this complexity, we no longer change the request URL. Middlewares are
responsible to access the disk safely by using http.Dir or, if not
actually opening files, they can use httpserver.SafePath().

### 3. What documentation changes (if any) go along with this PR?
Nothing of note, except instructions for middleware authors:

Middlewares should NOT do this:

	os.Open(filepath.Join(siteRoot, filepath.FromSlash(r.URL.Path)))

But this is safe:

	jailedDisk := http.Dir(siteRoot)
	jailedDisk.Open(r.URL.Path)

If not opening files, you can use `SafePath()`:

	httpserver.SafePath(siteRoot, r.URL.Path)

But `SafePath()` is good for when you don't need to open the file.

### 4. Checklist

- [x] This change has comments for package types, values, and functions or non-obvious lines of code

I'm considering having SafePath() take the entire *http.Request rather than just a string (so right now you have to pass req.URL.Path) - but it's not a big deal.

**This is a big change, but the highlights are:**

- The only context key relevant to path rewriting is `httpserver.OriginalURLCtxKey` ("original_url") which stores a copy of the original url.URL value before any middlewares even see the request.
- The HTTP server no longer sanitizes paths.
- http.Dir or httpserver.SafePath() should be used by middlewares that want to access the disk.
- Improved the handling in the static file server as well.
- Overall, this should lead to much simpler code and fewer problems. (I hope. Please test.)


caddyhttp/browse/browse.go
```diff
@@ -333,9 +333,14 @@ func (b Browse) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 
 	// Browsing navigation gets messed up if browsing a directory
 	// that doesn't end in "/" (which it should, anyway)
-	if !strings.HasSuffix(r.URL.Path, "/") {
-		staticfiles.RedirectToDir(w, r)
-		return 0, nil
+	u := r.Context().Value(httpserver.OriginalURLCtxKey).(url.URL)
```
tw4452852: This maybe not what we want. If `rewrite` occurred before, `r.URL.Path` contains the result and we should use it. Otherwise `rewrite` middleware will disfunction when combining with `browser`.

mholt: Yeah, good catch. This is tricky. The reason I chose to get the path of the original URL is because we end up causing a redirect with it; the redirect is external, but the rewrite is internal.

Maybe... should we check for a trailing slash using `r.URL.Path` and then issue the direct by modifying the original path stored in the context?

caddyhttp/extensions/ext.go
```diff
@@ -31,22 +31,14 @@ type Ext struct {
 // ServeHTTP implements the httpserver.Handler interface.
 func (e Ext) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	urlpath := strings.TrimSuffix(r.URL.Path, "/")
-	if path.Ext(urlpath) == "" && len(r.URL.Path) > 0 && r.URL.Path[len(r.URL.Path)-1] != '/' {
+	if len(r.URL.Path) > 0 && path.Ext(urlpath) == "" && r.URL.Path[len(r.URL.Path)-1] != '/' {
 		for _, ext := range e.Extensions {
-			if resourceExists(e.Root, urlpath+ext) {
+			_, err := os.Stat(httpserver.SafePath(e.Root, urlpath) + ext)
+			if err == nil {
```
elcore: I am nitpicking, but I guess this should be fine

```
if _, err := os.Stat(httpserver.SafePath(e.Root, urlpath) + ext); err == nil {
	r.URL.Path = urlpath + ext
	break
}
```

caddyhttp/browse/browse.go
```diff
@@ -333,9 +333,14 @@ func (b Browse) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
...
+	if u.Path == "" {
+		u.Path = "/"
+	}
+	if u.Path[len(u.Path)-1] != '/' {
```
tobya: Is there a reason not to use `strings.HasSuffix` here?

caddyhttp/staticfiles/fileserver.go
```diff
@@ -25,146 +30,151 @@ import (
...
+	if u.Path == "" {
+		u.Path = "/"
+	}
 	if d.IsDir() {
-		// Ensure / at end of directory url. If the original URL path is
-		// used then ensure / exists as well.
-		if !strings.HasSuffix(r.URL.Path, "/") {
-			RedirectToDir(w, r)
+		// ensure there is a trailing slash
+		if u.Path[len(u.Path)-1] != '/' {
```
tobya: Similar here, any reason not to use `strings.HasSuffix`

tobya:
I think this provides a much more straightforward way of dealing with `OriginalURLCtxKey` which is great.

caddyhttp/browse/browse.go
mholt: 🤷‍♂️ Beats me, the standard library does it this way. Maybe a single byte comparison is faster than needing do all the logic of strings.HasSuffix.

caddyhttp/browse/browse.go
tw4452852: I think the original logic is okay.

caddyhttp/extensions/ext.go
mholt: I prefer the two-line version in this case because there's already a lot of logic in that first line. Thanks though!

elcore: Sure 😄

caddyhttp/browse/browse.go
mholt: @tw4452852 Alrighty -- updated.

caddyhttp/httpserver/server.go
```diff
@@ -558,3 +547,22 @@ func WriteTextResponse(w http.ResponseWriter, status int, body string) {
...
 	w.Write([]byte(body))
 }
+
+// SafePath joins siteRoot and reqPath and converts it to a path that can
+// be used to access a path on the local disk. It ensures the path does
+// not traverse outside of the site root.
+//
+// If opening a file, use http.Dir instead.
+func SafePath(siteRoot, reqPath string) string {
+	if filepath.Separator != '/' {
```
mholt: Yep, indeed. Updated!

caddyhttp/staticfiles/fileserver.go
mholt: Same as above: no particular reason. Std lib does it this way. /shrug

mholt:
@tw4452852 Feel free to approve (or request changes to) the PR when you have a chance. I can get this out later today if so. Thanks so much for your time!

tw4452852:
LGTM.

Issue #1628, opened 2017-04-27
fastcgi on Windows: Caddyfile is accessible


l5h5t7:
Just created a Caddyfile, then run a site, I am surprised that Caddyfile is accessible through https://example.com/Caddyfile .
If add `browser /` to Caddyfille, then Caddyfile is inaccessible https://example.com/Caddyfile through https://example.com/Caddyfile .
Does this meet expectations?
I add `basicauth` to Caddyfile for another site, so I think it's unsafe.

mholt:
Hi @l5h5t7  - I cannot reproduce this behavior:

```
::1 - - [27/Apr/2017:05:38:48 -0600] "GET /Caddyfile HTTP/1.1" 404 14
```

Next time, please follow the issue template so we have enough information to see what you're seeing.

l5h5t7:
OK! I will follow the issue template next time.

### 1. What version of Caddy are you using (`caddy -version`)?

Web Server: Caddy 0.10.0;
PHP: 5.4.28.0 (In fact, in any FastCGI);
OS: Windows 2003, Windows 10.

### 3. What is your entire Caddyfile?

```
:2020 {
	fastcgi / 127.0.0.1:9000 php
}
```

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

The Caddyfile is in webroot.
Access http://localhost:2020/Caddyfile ,then you can reproduce it.

mholt:
What's your backend doing? If it's serving static files, then of course it can read the Caddyfile. Caddy can't stop your fastcgi process from reading it. You'll have to move it out of the site or change permissions to forbid it.

tobya:
Caddy prevents the caddyfile from being served as a static file.

It should prevent the file from being passed to fastcgi also.

mholt:
I disagree. It's not Caddy's place to meddle in what upstreams are doing on the file system. The OS and the site owner are responsible for that.

l5h5t7:
Even FastCGI backend doesn't work (close php-cgi.exe), Caddyfile is still accessible.
My backend runs PHP using `fastcgi / 127.0.0.1:9000 php` only.

mholt:
@l5h5t7 Not sure I follow. You mean you took out the fastcgi directive and you could still access the Caddyfile? (Did you restart the server after changing the Caddyfile?)

l5h5t7:
Yes, I restarted caddy after changing the Caddyfile.
[...]

PR #1637, opened 2017-05-01
Bugfix for issue #1628 where `Caddyfile` is not being hidden correctl…

slightfoot:
Fixes the path handling on windows when storing the `Caddyfile` into `HiddenFiles`

Issue #1628 

None

### 4. Checklist


mholt:
Brilliant, Simon!

A test for this should be easy. There's already a test in the file server that it hides what it is asked to hide -- the flaw here, if I understand correctly, was that we were telling it to hide something that didn't exist ("\Caddyfile") (on Windows). So all we would need is a `TestHideCaddyfile()` function that ensures the slash is converted to the canonical forward slash `/`. That should be sufficient, I think. If you agree, go ahead and write that up and then add it to this PR, and I'll merge it to go out on a release this week. :+1:

PR #1641, opened 2017-05-02
rename Push action for more general use

tw4452852:
I have just read the `LinkHeader` [spec](https://www.w3.org/wiki/LinkHeader), found that we should support it more generally, like `Apache` and `Nginx` do.

rename context's `Push` action to `AddLink` for more general use case.

N/A.

Add a new template action for `AddLink`.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/tplcontext.go
+++ b/caddyhttp/httpserver/tplcontext.go
@@ -424,12 +424,13 @@ func (c Context) RandomString(minLen, maxLen int) string {
 	return string(result)
 }
 
-// Push adds a preload link in response header for server push
-func (c Context) Push(link string) string {
+// AddLink adds a link header in response
+// see https://www.w3.org/wiki/LinkHeader
+func (c Context) AddLink(link string) string {
 	if c.responseHeader == nil {
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


mholt:
Is any use for the Link header other than server push common though? (Especially from templates.) I have never seen them used myself... :confused: Do you use them?

tw4452852:
Yes, it's just for server push now.  But it also allows user to give more control over its links, like [extensions](https://www.w3.org/TR/preload/#link-element-interface-extensions) or `nopush` target attribute.

> Do you use them?

Yes, I have used it in my [template files](https://github.com/tw4452852/totorow/blob/master/web/templates/header.html#L19).

mholt:
@tw4452852 Huh; does setting the link header without the `rel=preload` and without the `< >` brackets still work for server push?

Anyway, I'm fine with this I guess, and will merge it pending an answer. :)

tw4452852:
@mholt I think so, but we'd better comply with spec.

mholt:
@tw4452852 I was asking because in your example, it doesn't look spec-compliant, it's missing the `< >` and rel=preload.

tw4452852:
In my example, I do a little tricks in code, so it is surrounded by `<` in the end, but without `preload`. It seems works.

PR #1648, opened 2017-05-03
caddy 'login' plugin

smancke:
I would like to add my login plugin as an official 3rd party plugin.
It allows the login using multiple backends and plays nicely together with the existing caddy-jwt plugin. 

See the README's for more details: 
  https://github.com/tarent/loginsrv
  https://github.com/tarent/loginsrv/tree/master/caddy

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -468,6 +468,7 @@ var directives = []string{
 	"status",
 	"cors", // github.com/captncraig/cors/caddy
 	"mime",
+	"login",     // github.com/tarent/loginsrv/caddy
 	"jwt",       // github.com/BTBurke/caddy-jwt
 	"jsonp",     // github.com/pschlump/caddy-jsonp
 	"upload",    // blitznote.com/src/caddy.upload


mholt:
This is cool! Have you seen this, @captncraig? I know he was working on something similar...

Couple questions:

- You'll need to remove the RegisterDevDirective() call before deploying your plugin
- The docs mention two directives: `loginsrv` and `login`. Did you mean to have both? I prefer `login`.

smancke:
> You'll need to remove the RegisterDevDirective() call before deploying your plugin
Yes, I will remove this and I still have to do some cleanup an testing before releasing it. 

> The docs mention two directives: loginsrv and login. Did you mean to have both? I prefer login.
Yeah, I renamed it to login and had missed this one on the docs .. fixed.

mholt:
Great! I'm going to wait on @captncraig to answer because I know he had some oauth plugin in the works. I'm not sure how far he got on it.

smancke:
I already found it:
https://github.com/captncraig/caddy-oauth

captncraig:
That code is old. Pushing my newest version now.

captncraig:
I'd say mine is about 90% complete. Just need to handle the cases of what to do when auth is denied with redirects and so forth. And documentation.

captncraig:
Just pushed what I've been working on.

mholt:
Cool - so in order to make sure we don't have multiple plugins doing the same thing, how are these two different? (If they aren't, can they be merged easily?)

captncraig:
Honestly, I don't have as much free time for this as I would like. If your thing can do oauth well, I'd be happy to help contribute to a unified effort. My goals were these:

1. Super easy to add new providers. A provider should be simply oauth config and a function to get user data specific to the site. Also add ability to protect based on provider specific things like username / github org membership and so forth. My parsing code allows provider specific config directives of the form `github_whatever` without requiring explicit handling.
2. Allow "simple" mode where you simply protect pages, as well as "advanced" mode where we pass various headers to a proxied service. I would like to use it for a proxied go app that need github logins, including api integration. I tried to make it possible to pass (and trust) headers for things like github username, access token and so forth.
3. Security of course. I was using encrypted cookies to store oauth state and final user info headers and so forth. I don't know a lot about jwt, but I assume it accomplished similar things. 
4. Ease of use. I was proud of my solution to generate a cookie secret on the machine and store it in the .caddy directory for example. Makes a secure, machine specific secret by default, but still allows user to supply it themselves.

Shall we work together to create a single plugin?

smancke:
Hi Craig,

it would be great to work together !!!

From what I saw by now, loginsrv is already doing a little bit more than caddy-oauth and has already a documentation and good test coverage. So I would propose to take over missing parts from caddy-oauth to loginsrv, if any. 

**About your goals:**
> Super easy to add new providers. A provider should be simply oauth config and a function to get user data specific to the site.

It is exactly that. Please take a look at: https://github.com/tarent/loginsrv/blob/master/oauth2/github.go

> ... protect pages ...  pass various headers ...

Loginsrv itself is only about the login and cookie creation. It is designed to play nicely together with @BTBurke's [caddy-jwt plugin](https://github.com/BTBurke/caddy-jwt). So the protection can be done by caddy-jwt. The plugin allows enough rules for most cases and is able to forward the cookie info via headers. 

> Security of course. I was using encrypted cookies to store oauth state and final user info headers and so forth. I don't know a lot about jwt, but I assume it accomplished similar things.

The benefit of JWT is, that it is a standard, so different applications on your server side are able to check the Token, as log as they have the secret. This feature is essential for splitting up the infrastructure in independent microservices. But JWT is only about singning the token, not about encryption. So you can't put secret infomation in the token.
There is also a specification out there for JSON Web Encryption (JWE), but I guess in our usecase, the signing is enough, because the user is the owner of the information.

For easy usage, loginsrv is also able to create a random secret (as well as take one from config or environment variable). The generated secret is only held in memory: It is a good idea to store it on disk.

**Additional features of loginsrv:**
* It is not limited to oauh, but also provides a login with password check against pluggable providers. Especially an htaccess file is supported.
* It has a minimal frontend/Login mask.
* It is also usable as standalone login service or golang middleware (without caddy)

> Shall we work together to create a single plugin?

Yes! 
Since I'm using loginsrv already in some production projects, I would prefer to continue the work in this repository. But I can give you enough rights to the repository, as long as we agree on review processes. Yesterday I created a [Gitter Room](https://gitter.im/tarent/loginsrv) for loginsrv. So if you like, we can discuss further ideas and features there.

captncraig:
Hmm, ok, I need to play with jwt to see how the two work together. I agree we should proceed with your repo until I find sufficient reason to think otherwise. I do have some concerns, but will raise issues over there.

hacdias:
I just love love how it looks (just read the README)! Good work!! I'll keep a look on this and I'll be waiting for this to be published 😄 Then I'll add a mention to it on Hugo and FileManager docs probably.

mholt:
Great -- you're making progress.

Is this plugin's directive name (`login`) set in stone, then? And is the placement in the middleware chain correct?

If so, I will merge this in.

smancke:
> If so, I will merge this in.

Yes, please.

mholt:
Thanks, @smancke and @captncraig ! Looking forward to getting the next version out so you can publish the plugin. :+1:

smancke:
:+1:

Issue #1611, opened 2017-04-22
templates: Render whatever is written as a template, rather than loading a static file


mholt:
This is a memo to improve how `templates` works. We could potentially extend this idea to `markdown` as well.

Right now, the `templates` directive loads a static file from disk, parses it as a template, and executes it. Unfortunately, there's no way in Go (that I know of) to render the contents of an io.Reader as a template; it has to be fully buffered in memory. So, that's how we do it.

It could be really useful if `templates` would wrap the ResponseWriter with its own that buffers the body, then renders whatever is written to it as a template. The buffering _shouldn't_ be a new problem since we're doing that already anyway. But this would allow us to proxy to a backend that returns template HTML that Caddy then renders.

This idea could extend to `markdown` too, because that middleware _also_ has to load the full input into memory anyway.

I'm going to open this to a "help wanted" if anyone would like to take a crack at it. REMEMBER: When wrapping the ResponseWriter, it must satisfy a few interfaces:

```
var _ http.Pusher = (*MyResponseRecorder)(nil)
var _ http.Flusher = (*MyResponseRecorder)(nil)
var _ http.CloseNotifier = (*MyResponseRecorder)(nil)
var _ http.Hijacker = (*MyResponseRecorder)(nil)
```

^ These lines can be used as "interface guards" to ensure the type that records the response satisfies the interfaces. See httpserver.ResponseRecorder for an example. (Maybe it would be worth making `httpserver.limitWriter` an exported type that other middlewares could use?)

eticzon:
@mholt first off thanks for `caddy`! I'd like to get involved. Would you mind if take a stab at this? I'll drop some notes here as I go along so I can check if I'm on the right track.

elcore:
> @mholt first off thanks for caddy! I'd like to get involved. Would you mind if take a stab at this? I'll drop some notes here as I go along so I can check if I'm on the right track.

Let's Go 😇

mholt:
@eticzon Sure -- don't be afraid to submit a PR -- smaller ones are better. i.e. one for templates, one for markdown, for example. We'll review and iterate on it until it's ready to merge.

PR #1649, opened 2017-05-05
templates: Execute template loaded by later middlewares

mholt:

This is the beginning of an attempt to make the staticfiles file server
the only middleware that hits the disk and loads content. This may have
unknown implications. But the goal is to reduce duplication without
sacrificing performance. (We now call ServeContent here.)

This change loses about 15% of the req/sec of the old way of doing it,
but this way is arguably more correct since the file server is good at
serving static files; duplicating that logic in every middleware that
needs to hit the disk is not practical.

UPDATE: Performance numbers have been improved back to their
original levels, even slightly better in some cases.

#1611 

Hopefully none

### 4. Checklist

I would like some others to try using this with `templates` to see if they can uncover any broken scenarios. Also, if anyone would like to help profile to understand why this change is slower, that'd be great. I don't want to merge this in while it's slower.

Before this change (serving the Caddy site's homepage):

```
$ wrk -t12 -c400 -d10s http://localhost:2015/
Running 10s test @ http://localhost:2015/
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    42.95ms   42.64ms 387.61ms   85.72%
    Req/Sec     0.87k   161.81     1.32k    70.08%
  103478 requests in 10.02s, 0.97GB read
  Socket errors: connect 0, read 279, write 0, timeout 0
Requests/sec:  10330.04
Transfer/sec:     99.45MB
```

After:

```
$ wrk -t12 -c400 -d10s http://localhost:2015/
Running 10s test @ http://localhost:2015/
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    57.35ms   66.82ms 653.25ms   86.64%
    Req/Sec   704.43    150.37     1.26k    69.50%
  84314 requests in 10.03s, 811.25MB read
  Socket errors: connect 0, read 280, write 0, timeout 0
Requests/sec:   8409.06
Transfer/sec:     80.91MB
```

UPDATE: Performance has been improved...


caddyhttp/templates/templates.go
```diff
@@ -4,90 +4,92 @@ package templates
...
-				// If Content-Type isn't set here, http.ResponseWriter.Write
-				// will set it according to response body. But other middleware
-				// such as gzip can modify response body, then Content-Type
-				// detected by http.ResponseWriter.Write is wrong.
-				ctype := mime.TypeByExtension(ext)
-				if ctype == "" {
-					ctype = http.DetectContentType(buf.Bytes())
-				}
-				w.Header().Set("Content-Type", ctype)
+				modTime, _ := time.Parse(http.TimeFormat, w.Header().Get("Last-Modified"))
```
tw4452852: `modTime` may be invalid here because `Last-Modified` is nil or its format is invalid. We'd better check error.

caddyhttp/templates/templates.go
```diff
@@ -4,90 +4,92 @@ package templates
...
+				buf.Reset()
+				defer t.BufPool.Put(buf)
+				respRec := &httptest.ResponseRecorder{
+					HeaderMap: make(http.Header),
+					Body:      buf,
+				}
+
+				// pass request up the chain to let another middleware provide us the template
+				code, err := t.Next.ServeHTTP(respRec, r)
+				if code >= 300 {
```
tw4452852: Also check `err` ?

caddyhttp/templates/templates.go
mholt: You're right, but if it is, there's nothing we can do about it; we shouldn't stop serving the content just because the Last-Modified header is missing or invalid IMO.

caddyhttp/templates/templates.go
mholt: I guess we could, will update soon.

caddyhttp/templates/templates.go
```diff
@@ -4,90 +4,92 @@ package templates
...
-				w.Header().Set("Content-Type", ctype)
+				modTime, _ := time.Parse(http.TimeFormat, w.Header().Get("Last-Modified"))
 
-				templateInfo, err := os.Stat(templatePath)
-				if err == nil {
-					// add the Last-Modified header if we were able to read the stamp
-					httpserver.SetLastModifiedHeader(w, templateInfo.ModTime())
-				}
-				buf.WriteTo(w)
+				http.ServeContent(w, r, templateName, modTime, bytes.NewReader(buf.Bytes()))
```
tw4452852: > // TODO: This performs ~15% worse (req/s) than the old method; why?

I did some investigation find that the culprit is `http.ServeContent`. It will use `io.Copy` to write content to the `w` which allocates a lot of memory. For now, there are two places using `http.ServeContent`, one is here and the other one is `staticfiles` middleware.

I take a memory profile as reference:

```
         .          .     19:func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
         .          .     20:   // TODO: This performs ~15% worse (req/s) than the old method; why?
         .          .     21:   // TODO: Need to somehow embed upcoming httpserver.ResponseWriterWrapper
         .          .     22:   // into the ResponseRecorder, so it implements the needed interfaces
         .          .     23:
         .          .     24:   for _, rule := range t.Rules {
         .          .     25:           if !httpserver.Path(r.URL.Path).Matches(rule.Path) {
         .          .     26:                   continue
         .          .     27:           }
         .          .     28:
         .          .     29:           // check for index files
         .          .     30:           fpath := r.URL.Path
         .          .     31:           if idx, ok := httpserver.IndexFile(t.FileSys, fpath, rule.IndexFiles); ok {
         .          .     32:                   fpath = idx
         .          .     33:           }
         .          .     34:
         .          .     35:           // check if extension matches
         .          .     36:           reqExt := path.Ext(fpath)
         .          .     37:           for _, ext := range rule.Extensions {
         .          .     38:                   if reqExt == ext {
         .          .     39:                           // get a buffer from the pool and make a response recorder
         .          .     40:                           buf := t.BufPool.Get().(*bytes.Buffer)
         .          .     41:                           buf.Reset()
         .          .     42:                           defer t.BufPool.Put(buf)
         .          .     43:                           respRec := &httptest.ResponseRecorder{
       4MB        4MB     44:                                   HeaderMap: make(http.Header),
       2MB        2MB     45:                                   Body:      buf,
         .          .     46:                           }
         .          .     47:
         .          .     48:                           // pass request up the chain to let another middleware provide us the template
         .     1.29GB     49:                           code, err := t.Next.ServeHTTP(respRec, r)
         .          .     50:                           if code >= 300 {
         .          .     51:                                   return code, err
         .          .     52:                           }
         .          .     53:
         .          .     54:                           // copy the buffered response headers to the real response writer
         .          .     55:                           for field, val := range respRec.HeaderMap {
         .          .     56:                                   w.Header()[field] = val
         .          .     57:                           }
         .          .     58:
         .          .     59:                           // create a new template
         .          .     60:                           templateName := filepath.Base(fpath)
         .          .     61:                           tpl := template.New(templateName)
         .          .     62:
         .          .     63:                           // set delimiters
         .          .     64:                           if rule.Delims != [2]string{} {
         .          .     65:                                   tpl.Delims(rule.Delims[0], rule.Delims[1])
         .          .     66:                           }
         .          .     67:
         .          .     68:                           // add custom functions
         .          .     69:                           tpl.Funcs(httpserver.TemplateFuncs)
         .          .     70:
         .          .     71:                           // parse the template
         .    67.01MB     72:                           parsedTpl, err := tpl.Parse(respRec.Body.String())
         .          .     73:                           if err != nil {
         .          .     74:                                   return http.StatusInternalServerError, err
         .          .     75:                           }
         .          .     76:
         .          .     77:                           // create execution context for the template template
         .          .     78:                           ctx := httpserver.NewContextWithHeader(w.Header())
         .          .     79:                           ctx.Root = t.FileSys
         .          .     80:                           ctx.Req = r
         .          .     81:                           ctx.URL = r.URL
         .          .     82:
         .          .     83:                           // execute the template
         .          .     84:                           buf.Reset()
    2.50MB   179.11MB     85:                           err = parsedTpl.Execute(buf, ctx)
         .          .     86:                           if err != nil {
         .          .     87:                                   return http.StatusInternalServerError, err
         .          .     88:                           }
         .          .     89:
         .          .     90:                           modTime, _ := time.Parse(http.TimeFormat, w.Header().Get("Last-Modified"))
         .          .     91:
         .          .     92:                           http.ServeContent(w, r, templateName, modTime, bytes.NewReader(buf.Bytes()))
         .          .     93:
         .     1.22GB     94:                           return http.StatusOK, nil
         .          .     95:                   }
         .          .     96:           }
         .          .     97:   }
```

tw4452852:
To avoid memory allocation in `io.Copy` which `http.ServeContent` calls at two places: 

1. implement `io.ReaderFrom` in `ResponseWriter` for `staticfiles.ServeHTTP`.
2. use the original `buf.WriteTo(w)` for `templates.ServeHTTP`.

Here is the patch:

```
diff --git a/caddyhttp/templates/templates.go b/caddyhttp/templates/templates.go
index 6ea09c4..cd0fb87 100644
--- a/caddyhttp/templates/templates.go
+++ b/caddyhttp/templates/templates.go
@@ -5,12 +5,11 @@ package templates
 import (
        "bytes"
        "net/http"
-       "net/http/httptest"
        "path"
        "path/filepath"
        "sync"
        "text/template"
-       "time"
+       _ "time"
 
 )
@@ -40,19 +39,20 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
                                buf := t.BufPool.Get().(*bytes.Buffer)
                                buf.Reset()
                                defer t.BufPool.Put(buf)
-                               respRec := &httptest.ResponseRecorder{
-                                       HeaderMap: make(http.Header),
-                                       Body:      buf,
+                               rb := &responseBuffer{
+                                       status: http.StatusOK,
+                                       header: make(http.Header),
+                                       Buffer: buf,
                                }
 
                                // pass request up the chain to let another middleware provide us the template
-                               code, err := t.Next.ServeHTTP(respRec, r)
+                               code, err := t.Next.ServeHTTP(rb, r)
                                if code >= 300 {
                                        return code, err
                                }
 
                                // copy the buffered response headers to the real response writer
-                               for field, val := range respRec.HeaderMap {
+                               for field, val := range rb.header {
                                        w.Header()[field] = val
                                }
 
@@ -69,7 +69,7 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
                                tpl.Funcs(httpserver.TemplateFuncs)
 
                                // parse the template
-                               parsedTpl, err := tpl.Parse(respRec.Body.String())
+                               parsedTpl, err := tpl.Parse(rb.Buffer.String())
                                if err != nil {
                                        return http.StatusInternalServerError, err
                                }
@@ -87,9 +87,10 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
                                        return http.StatusInternalServerError, err
                                }
 
-                               modTime, _ := time.Parse(http.TimeFormat, w.Header().Get("Last-Modified"))
+                               // modTime, _ := time.Parse(http.TimeFormat, w.Header().Get("Last-Modified"))
 
-                               http.ServeContent(w, r, templateName, modTime, bytes.NewReader(buf.Bytes()))
+                               // http.ServeContent(w, r, templateName, modTime, bytes.NewReader(buf.Bytes()))
+                               buf.WriteTo(w)
 
                                return http.StatusOK, nil
                        }
@@ -99,6 +100,20 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
        return t.Next.ServeHTTP(w, r)
 }
 
+type responseBuffer struct {
+       status int
+       header http.Header
+       *bytes.Buffer
+}
+
+func (rb *responseBuffer) Header() http.Header {
+       return rb.header
+}
+
+func (rb *responseBuffer) WriteHeader(status int) {
+       rb.status = status
+}
+
 // Templates is middleware to render templated files as the HTTP response.
 type Templates struct {
        Next    httpserver.Handler
```

This patch is only for test.

mholt:
@tw4452852 That's amazing, thank you!

I'll give it a try as soon as I get a chance, and do some benchmarks to compare.

I imagine we'll want the responseBuffer to implement [httpserver.HTTPInterfaces](https://godoc.org/github.com/mholt/caddy/caddyhttp/httpserver#HTTPInterfaces) by wrapping it with ResponseWriterWrapper?

tw4452852:
@mholt 

> I imagine we'll want the responseBuffer to implement httpserver.HTTPInterfaces by wrapping it with ResponseWriterWrapper?

Yes, I think so.

mholt:
@tw4452852 So far, your patch does indeed seem to be faster. Even slightly faster than how we had it before my PR in the first place. :+1: But can you elaborate on this for me?

> implement io.ReaderFrom in ResponseWriter for staticfiles.ServeHTTP

I understand your second point, but I'm not sure what you mean by this exactly. I see that io.CopyN will use ReaderFrom and WriterTo to avoid new allocations. Sounds yummy, I'm looking into how to do that in the file server, is that right?

mholt:
@tw4452852  Aha, I've been digging way too deep on this for hours now and I think I know what you meant.

With the `responseBuffer` type you defined and used instead of `httptest.ResponseRecorder`, we are implementing io.ReaderFrom because `responseBuffer` _embeds_ `*bytes.Buffer` which is an io.ReaderFrom! That explains why there is a speedup: not because ServeContent() in this ServeHTTP is slow, but because ServeContent in  the staticfiles ServeHTTP was slow. Because http.ResponseRecorder is _not_ a ReaderFrom, it cannot benefit from the optimization in io.Copy when ServeContent is called in the fileserver.

At this point, I am using your patch, but without the buf.WriteTo; I'm still using ServeContent in the templates ServeHTTP, because `bytes.NewReader(buf.Bytes())` is allocation-free and still satisfies io.WriterTo, which also benefits io.Copy. Plus, ServeContent handles range requests properly.

The performance numbers are looking much better. I'm going to do some cleanup and see what I can push next!

mholt:
@tw4452852 I've implemented your suggestions to improve performance, and learned a lot in the process!

I introduced a new type, httpserver.ResponseBuffer, so other middlewares can take advantage of it if need be. I verified with lots of `wrk` tests that implementing io.ReaderFrom is a great idea. :wink: Please take another look and review when you have a chance!

elcore:
I am amazed

mholt:
@tw4452852 Can you take another look at this when you have a chance? I'd hate to merge without a review, but might have to do so in order to get this on the next release.

tw4452852:
@mholt 
Sorry for late. Look good to me. Let's give it a try.

mholt:
Yay - thanks. I hope things are going well for you lately!

PR #1653, opened 2017-05-06
Support template arguments for Markdown.

glesica:

Adds the second `args ...interface{}` parameter to the version of `Include` used with the Markdown plugin. This allows arguments to included templates when using that plugin.

I don't think this will require any changes. The documentation already implies that passing arguments to included templates should be possible.

### 4. Checklist

- [ ] I have written tests and verified that they fail without my change - the file is untested at the moment and the change is pretty trivial, I can write tests if necessary, though.


glesica:
@mholt not sure what the review process is, but this is a pretty small change... :-)

mholt:
Thanks for contributing, @glesica! Welcome to the project. Thanks for making the code consistent with the documentation. ;)

Issue #1652, opened 2017-05-06
proxy: A websocket test is not synchronized


mholt:
### 1. What version of Caddy are you using (`caddy -version`)?

v0.10.2

### 2. What are you trying to do?

Run `go test -race ./...` without failures

### 3. What is your entire Caddyfile?

n/a

### 4. How did you run Caddy (give the full command and describe the execution environment)?

`go test ./...` from `github.com/mholt/caddy` folder.

### 5. Please paste any relevant HTTP request(s) here.

n/a

### 6. What did you expect to see?

All tests pass

### 7. What did you see instead (give full error messages and/or log)?

```
--- FAIL: TestWebSocketReverseProxyServeHTTPHandler (0.00s)
        proxy_test.go:281: Expected 1 websocket connection, got 0
FAIL
FAIL    github.com/mholt/caddy/caddyhttp/proxy  5.229s
```

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

It's spotty. Re-running the tests succeeds.

Apparently the test `TestWebSocketReverseProxyServeHTTPHandler` is not synchronized. I am not 100% sure where the "pseudo-race" is (there IS a race, but it's not a "memory" race) but maybe it's just needing some sort of synchronization. Thanks to @hacdias and @bruhs for the report!

PR #1654, opened 2017-05-07
proxy: synchronize websocket test

tw4452852:
fix issue #1652

Fix `TestWebSocketReverseProxyServeHTTPHandler` failure.

#1652 

N/A.

### 4. Checklist

### DIFF
--- a/caddyhttp/proxy/proxy_test.go
+++ b/caddyhttp/proxy/proxy_test.go
@@ -247,8 +247,8 @@ func TestWebSocketReverseProxyNonHijackerPanic(t *testing.T) {
 func TestWebSocketReverseProxyServeHTTPHandler(t *testing.T) {
 	// No-op websocket backend simply allows the WS connection to be
 	// accepted then it will be immediately closed. Perfect for testing.
-	var connCount int32
-	wsNop := httptest.NewServer(websocket.Handler(func(ws *websocket.Conn) { atomic.AddInt32(&connCount, 1) }))
+	accepted := make(chan struct{})
+	wsNop := httptest.NewServer(websocket.Handler(func(ws *websocket.Conn) { close(accepted) }))
 	defer wsNop.Close()
 
 	// Get proxy to use for the test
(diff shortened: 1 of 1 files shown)


caddyhttp/proxy/proxy_test.go
```diff
@@ -247,8 +247,8 @@ func TestWebSocketReverseProxyNonHijackerPanic(t *testing.T) {
 func TestWebSocketReverseProxyServeHTTPHandler(t *testing.T) {
 	// No-op websocket backend simply allows the WS connection to be
 	// accepted then it will be immediately closed. Perfect for testing.
-	var connCount int32
```
mholt: Part of the importance of this test, if I recall correctly, is to ensure that only one connection is made to the backend, hence the counter.

I think we shouldn't delete the count. I don't recall the issue anymore, and I wish the comments in the code mentioned it, but I do think the count is necessary.

elcore: @mholt The PR is #987 (issue #746)

mholt: BINGO. That's the one, Eldin! Thanks!!

tw4452852: @mholt I think we could still make sure only one connection is made here: If there are two connections reaching the backend, the second one will trigger a panic due to close of a closed channel, the test will fail as a result. Also, if there is no connection issued, the test will also fail because of timeout.

mholt: I just tested this, and indeed, you're right...

As for sleeping vs. time.After, it's okay either way, I think the first way you did it made sense too. But I'll merge this!

caddyhttp/proxy/proxy_test.go
```diff
@@ -279,8 +279,14 @@ func TestWebSocketReverseProxyServeHTTPHandler(t *testing.T) {
 	if !bytes.Equal(actual, expected) {
 		t.Errorf("Expected backend to accept response:\n'%s'\nActually got:\n'%s'", expected, actual)
 	}
-	if got, want := atomic.LoadInt32(&connCount), int32(1); got != want {
-		t.Errorf("Expected %d websocket connection, got %d", want, got)
+
+	// wait a minute for backend handling, see issue 1654.
+	time.Sleep(10 * time.Millisecond)
```
mholt: This isn't a lot of time in CI environments where the CPU is shared, but we'll see how it goes. We may have to raise this.

mholt:
Thanks Tw!! This is a great help.

PR #1665, opened 2017-05-11
Support internationalized domain name (IDN) in Caddyfile

cez81:
It adds support for using an IDN as a site address in Caddyfiles. Instead of writing `xn--bcher-kva.ch`. it's now possible to write `bücher.ch`.

Ref: https://www.wikiwand.com/en/Internationalized_domain_name

N/A

Could add a note [in site address section](https://caddyserver.com/docs/http-caddyfile#addresses)

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -14,6 +14,8 @@ import (
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyfile"
 	"github.com/mholt/caddy/caddytls"
+
+	"golang.org/x/net/idna"
 )
 
 const serverType = "http"
@@ -311,10 +313,11 @@ func (a Address) String() string {
 // VHost returns a sensible concatenation of Host:Port/Path from a.
 // It's basically the a.Original but without the scheme.
 func (a Address) VHost() string {
-	if idx := strings.Index(a.Original, "://"); idx > -1 {
-		return a.Original[idx+3:]
+	vhost := a.String()
+	if idx := strings.Index(vhost, "://"); idx > -1 {
+		return vhost[idx+3:]
 	}
-	return a.Original
+	return vhost
 }
 
 // standardizeAddress parses an address string into a structured format with separate
@@ -338,6 +341,15 @@ func standardizeAddress(str string) (Address, error) {
 		if err != nil {
 			host = u.Host
 		}
+	}
+
+	// convert non-ascii (IDNA) host to Punycode
+	if isIDN(host) {
+		asciiHost, err := idna.ToASCII(host)
+		if err != nil {
+			return Address{}, fmt.Errorf("Failed to convert IDN to ASCII: %s", input)
+		}
+		host = asciiHost
 	}
 
 	// see if we can set port based off scheme
@@ -369,6 +381,16 @@ func standardizeAddress(str string) (Address, error) {
 	}
 
 	return Address{Original: input, Scheme: u.Scheme, Host: host, Port: port, Path: u.Path}, err
+}
+
+// isIDN return true if s contains non-ascii characters
+func isIDN(s string) bool {
+	for _, c := range s {
+		if c > 127 {
+			return true
+		}
+	}
+	return false
 }
 
 // RegisterDevDirective splices name into the list of directives
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -345,11 +345,10 @@ func standardizeAddress(str string) (Address, error) {
 
 	// convert non-ascii (IDNA) host to Punycode
 	if isIDN(host) {
-		asciiHost, err := idna.ToASCII(host)
+		host, err = idna.ToASCII(host)
 		if err != nil {
-			return Address{}, fmt.Errorf("Failed to convert IDN to ASCII: %s", input)
+			return Address{}, fmt.Errorf("failed to convert IDN to ASCII: %s", input)
 		}
-		host = asciiHost
 	}
 
 	// see if we can set port based off scheme
--- a/caddyhttp/httpserver/plugin_test.go
+++ b/caddyhttp/httpserver/plugin_test.go
@@ -52,6 +52,10 @@ func TestStandardizeAddress(t *testing.T) {
 		{`host:80/path`, "", "host", "80", "/path", false},
 		{`host:https/path`, "https", "host", "443", "/path", false},
 		{`/path`, "", "", "", "/path", false},
+		{`http://idnhost-öåä`, "http", "xn--idnhost--7zaj0q", "80", "", false},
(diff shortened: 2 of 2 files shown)


caddyhttp/httpserver/plugin.go
```diff
@@ -311,10 +313,11 @@ func (a Address) String() string {
...
 // It's basically the a.Original but without the scheme.
 func (a Address) VHost() string {
-	if idx := strings.Index(a.Original, "://"); idx > -1 {
-		return a.Original[idx+3:]
+	vhost := a.String()
+	if idx := strings.Index(vhost, "://"); idx > -1 {
+		return vhost[idx+3:]
 	}
-	return a.Original
+	return vhost
```
mholt: Why does this function need to change at all?

caddyhttp/httpserver/plugin.go
```diff
@@ -340,6 +343,15 @@ func standardizeAddress(str string) (Address, error) {
 		}
 	}
 
+	// convert non-ascii (IDNA) host to Punycode
+	if isIDN(host) {
```
mholt: What's the significance of this check? AFAIK converting an ASCII name like `example.com` will remain unchanged. Is the extra function, this if statement, etc etc, all necessary?

caddyhttp/httpserver/plugin_test.go
```diff
@@ -80,16 +84,37 @@ func TestStandardizeAddress(t *testing.T) {
...
-		{Address{Original: "host:1234"}, "host:1234"},
-		{Address{Original: "host:1234/foo"}, "host:1234/foo"},
-		{Address{Original: "host/foo"}, "host/foo"},
-		{Address{Original: "http://host/foo"}, "host/foo"},
-		{Address{Original: "https://host/foo"}, "host/foo"},
+		{Address{Original: "host:1234", Scheme: "", Host: "host", Port: "1234", Path: ""}, "host:1234"},
+		{Address{Original: "host:1234/foo", Scheme: "", Host: "host", Port: "1234", Path: "/foo"}, "host:1234/foo"},
+		{Address{Original: "host/foo", Scheme: "", Host: "host", Port: "", Path: "/foo"}, "host/foo"},
+		{Address{Original: "http://host/foo", Scheme: "http", Host: "host", Port: "", Path: "/foo"}, "host/foo"},
+		{Address{Original: "https://host/foo", Scheme: "https", Host: "host", Port: "", Path: "/foo"}, "host/foo"},
```
mholt: I am not sure rewriting these existing tests is desirable, or at least, doesn't seem relevant for this change.

caddyhttp/httpserver/plugin.go
```diff
@@ -340,6 +343,15 @@ func standardizeAddress(str string) (Address, error) {
...
+		asciiHost, err := idna.ToASCII(host)
+		if err != nil {
+			return Address{}, fmt.Errorf("Failed to convert IDN to ASCII: %s", input)
```
mholt: Error strings should start with a lower-case character (lint). Change to `"failed to ..."`

caddyhttp/httpserver/plugin.go
```diff
@@ -340,6 +343,15 @@ func standardizeAddress(str string) (Address, error) {
...
+		}
+		host = asciiHost
```
mholt: This re-assignment is not necessary, is it? Just assign directly to `host`.

caddyhttp/httpserver/plugin.go
cez81: Otherwise the insertion in the vhost trie will use the non-ascii domain name (standardizeAddress only converts host) and lookups will fail.
https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/server.go#L125
https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/server.go#L354

caddyhttp/httpserver/plugin.go
cez81: Test fails for the following two cases when converting all hosts.
```
--- FAIL: TestStandardizeAddress (0.00s)
	plugin_test.go:66: Test 26 (::1): Expected error, but had none
	plugin_test.go:66: Test 27 (localhost::): Expected error, but had none
```

caddyhttp/httpserver/plugin.go
cez81: Updated

caddyhttp/httpserver/plugin.go
cez81: Updated

caddyhttp/httpserver/plugin_test.go
cez81: I had to update these or I couldn't use a.String() in VHost() but there is probably better ways to do it.

mholt:
I'll be getting back to this PR soon, I have a few things on my plate, and this is a pretty significant change, so I want to make sure it's correct. Thanks!

caddyhttp/httpserver/plugin.go
tw4452852: We couldn't use `Address.String()` here, because it contains additional scheme and path if not supplied.

Why not just translate non-ascii characters in `a.Original` ?

mholt:
@cez81 I've given this a lot of thought and been experimenting with IDNs locally, as well as talking to a few people who know more than I do about them.

I really appreciate the time you put into this, but I'm going to close the PR as I don't think it's where I want to take the Caddyfile. (This change is disruptive enough, believe or not -- and although not necessarily bad just because of that, perhaps this should have been discussed in an issue first.)

IDNA is intended primarily for display purposes; and all underlying servers, services, and protocols (including DNS and ACME) use the ASCII encoding (A-labels). There might be a few service *providers* that allow Unicode input, but server operators (Caddy users) will likely be interfacing with the punycode variants most of the time as they inspect certificates, configure DNS, etc. Browsers display unicode, but even copying+pasting the URL results in punycode.

I do not intend for the Caddyfile to be the final, outermost configuration layer. Eventually, a web UI will sit on top of Caddy and that component can do the conversion. Adding a new dependency to Caddy (even though it is in the /x/ packages), increasing binary size a notable amount (the Unicode tables are not small), adding this complexity to Caddy core, it just doesn't seem worth it to me. Also, I'm a little skeptical that all hostnames will flow through the `standardizeAddress()` function -- they should, I think, but it's not guaranteed; that leads to inconsistencies, potentially really dangerous ones when it comes to certificate management.

Perhaps instead we should return an error, asking site owners to [convert the domain to punycode first](https://www.punycoder.com/), if it is a non-ASCII domain. Then again, maybe there's nothing wrong with it (I'm not sure?); it's just that browsers make requests using the punycode formats. Either way, we can definitely make note of this in the documentation on the website.

Certbot [also had this discussion](https://github.com/certbot/certbot/issues/3614) a while ago and I think it's good we follow their wisdom.

cez81:
No worries :) Appreciate your detailed explanation for your decision.

Issue #1660, opened 2017-05-09
browse shows symlinked folders as files


theel0ja:
### 1. What version of Caddy are you using (`caddy -version`)?
0.9.5

### 2. What are you trying to do?
Use caddy with browse.

### 3. What is your entire Caddyfile?
I do not have Caddyfile, but the CLI command has `browse` as it.

### 4. How did you run Caddy (give the full command and describe the execution environment)?
`caddy browse`

### 5. Please paste any relevant HTTP request(s) here.

Just http://localhost:2015/

### 6. What did you expect to see?
Something like this:
![image](https://cloud.githubusercontent.com/assets/5832930/25861755/489a23a2-34ee-11e7-80f4-4bf9df328c36.png)

### 7. What did you see instead (give full error messages and/or log)?
![image](https://cloud.githubusercontent.com/assets/5832930/25861729/38cf7e86-34ee-11e7-8917-864d287d716f.png)

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

```
mkdir folder/
cd folder/
mkdir hello/
mkdir hello/world/
ln -s hello/world/ ./
caddy browse
```

Go to http://localhost:2015/ and see `world` as file, while it's folder.

Like this:
![image](https://cloud.githubusercontent.com/assets/5832930/25861857/aa1c82c8-34ee-11e7-8007-2d94bb5db342.png)

mholt:
Nice issue, @theel0ja  - thanks for all the info. I think this should be easy enough for someone new to the project to be able to contribute to, so I'll try to leave this for someone who has little (but some) Go experience to see if they want to submit a pull request!

henkburgstra:
I would like to fix this issue if that's ok :-)

mholt:
@henkburgstra @cez81 already beat you to it, but we do need some icons if you are savvy with SVG graphics.

henkburgstra:
I missed that, sorry.  Also not very good with SVG :( 
isSymlinkTargetDir by @cez81 doesn't seem to take into account that a symlink can point to another symlink. If that's the case it should recurse and protect itself against endless recursion.

mholt:
@henkburgstra No problem, maybe make that comment in-line on the pull request itself. :) Thanks!

cekvenich:
Has this been fixed? Working w/ ln -s?
And what about a mounted filesystem?

elcore:
>Has this been fixed?

Yes (in 0.10.4)

[...]

PR #1667, opened 2017-05-12
browse: Show symbolic links and target's type properly

cez81:
Symbolic links pointing to a directory will show as directory instead if file.

#1660 

https://caddyserver.com/docs/browse
Update examples and line numbers for "this struct type"-link

### 4. Checklist

I need to add some tests but I'll start with this to see if I'm on the right track :)


caddyhttp/browse/setup.go
```diff
@@ -426,7 +426,9 @@ footer {
 								<span class="name">{{html .Name}}</span>
 							</a>
 						</td>
-						{{- if .IsDir}}
+						{{- if .IsSymlink }}
+						<td data-order="-1">symbolic link</td>
```
mholt: Instead of showing the words "symbolic link" under the "Size" column, how about we introduce new icons for symbolic link to file and symbolic link to directory?

caddyhttp/browse/browse.go
```diff
@@ -258,12 +259,13 @@ func directoryListing(files []os.FileInfo, canGoUp bool, urlPath string, config
 		url := url.URL{Path: "./" + name} // prepend with "./" to fix paths with ':' in the name
 
 		fileinfos = append(fileinfos, FileInfo{
-			IsDir:   f.IsDir(),
-			Name:    f.Name(),
-			Size:    f.Size(),
-			URL:     url.String(),
-			ModTime: f.ModTime().UTC(),
-			Mode:    f.Mode(),
+			IsDir:     f.IsDir() || (isSymlink(f) && isSymlinkTargetDir(f)),
```
mholt: We could move the check for isSymlink() into isSymlinkTargetDir().

caddyhttp/browse/setup.go
cez81: Yes, that was my initial thought but graphics is not my thing :)

caddyhttp/browse/browse.go
cez81: :+1:

caddyhttp/browse/setup.go
mholt: I can update the graphics. How about we take this out (and leave the template unchanged) for now and I will see about getting icons with little arrows or something to indicate a symbolic link?

cez81: Sounds good to me!

mholt:
I'm waiting on this until I have a chance to update the icons, at which point I'll update your branch. Maybe this weekend?

mholt:
I have revised icons for the default template; I will get those pushed later tonight or tomorrow.

They look good I think. :)

cez81:
Looks good :)

mholt:
Just noting: a little bug snuck in, related to checking if the file is a symlink. `os.Readlink(f.Name())` assumes that `f` is in the current directory. I'll work on a fix soon...

Issue #1668, opened 2017-05-14
Extra " in Caddyfile log directive causes hosts to be skipped


tobya:
### 1. What version of Caddy are you using (`caddy -version`)?

0.10.0

### 2. What are you trying to do?

Run all my sites

### 3. What is your entire Caddyfile?
```text
recipes.cookingisfun.ie {

  tls {
   protocols tls1.0 tls1.2
  }
    log / recipesx.log  "{remote} - [{when}] "{method} {path} {>Caddy-Rewrite-Original-URI} {proto} {status} {port}" {scheme} {mitm} " 
         
         
    errors  errorsrecipesx.log   
      proxy / http://recipes.cookingisfun.ie:3032

}

recipesmove.bcs.cookingisfun.ie {

  tls {
   protocols tls1.0 tls1.2
  }
    log / recipesmovex.log "{remote} - [{when}] "{method} {path} {>Caddy-Rewrite-Original-URI} {proto} {status} {port}" {scheme} {mitm} "   
         
         
    errors  errorsrecipesmovex.log   


    fastcgi / 127.0.0.1:35454 php
    root "C:/BCSSERVER/Websites/Recipes.cookingisfun.ie/api_v1.0/recipes/"
    startup "c:\php\php56\php-cgi.exe -b 35454" & 




}

another.bcs.cookingisfun.ie {
  root "C:/BCSSERVER/Websites/Recipes.cookingisfun.ie/api_v1.0/recipes/"
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

caddy -log stdout

### 5. Please paste any relevant HTTP request(s) here.

(paste curl command, or full HTTP request including headers and body, here)

### 6. What did you expect to see?

All my sites served

### 7. What did you see instead (give full error messages and/or log)?
 Only http/https:// recipes.cookingisfun.ie and http/https:// another.bcs.cookingisfun.ie are served.  recipesmove.bcs.cookingisfun.ie is not.

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

Use the caddyfile with your own domains.

### Details

I came across this bug while modifying my log string.  I was copying and pasting and accidentally copied in a " inside a log string

>  log / recipesx.log  "{remote} - [{when}] "{method} {path} {>Caddy-Rewrite-Original-URI} {proto} {status} {port}" {scheme} {mitm} " 

note the " before {method}

[...]

PR #1672, opened 2017-05-15
Add check for maximum number of arguments to log directive

cez81:
This will add a check for maximum number of arguments to the log directive. If len(args) > 3 it will return an error. This can happen when having unmatched number of quotes.

#1668

None

### 4. Checklist

### DIFF
--- a/caddyhttp/log/setup.go
+++ b/caddyhttp/log/setup.go
@@ -72,7 +72,7 @@ func logParse(c *caddy.Controller) ([]*Rule, error) {
 				format = strings.Replace(format, "{combined}", CombinedLogFormat, -1)
 			}
 		default:
-			// Maxiumum number of args in log directive is 3.
+			// Maximum number of args in log directive is 3.
 			return nil, c.ArgErr()
 		}
 
--- a/caddyhttp/log/setup.go
+++ b/caddyhttp/log/setup.go
@@ -71,6 +71,9 @@ func logParse(c *caddy.Controller) ([]*Rule, error) {
 				},
 				Format: DefaultLogFormat,
 			})
+		} else if len(args) > 3 {
+			// Maxiumum number of args in log directive is 3.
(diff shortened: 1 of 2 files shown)


caddyhttp/log/setup.go
```diff
@@ -71,6 +71,9 @@ func logParse(c *caddy.Controller) ([]*Rule, error) {
 				},
 				Format: DefaultLogFormat,
 			})
+		} else if len(args) > 3 {
```
tw4452852: I prefer to translate these `else if` into `switch` for more readability.

caddyhttp/log/setup.go
```diff
@@ -53,7 +53,8 @@ func logParse(c *caddy.Controller) ([]*Rule, error) {
 			}
 		}
 
-		if len(args) == 0 {
+		switch len(args) {
```
tw4452852: Hmm, we could do even better like this:

```
path = "/"
format = DefaultLogFormat
output = DefaultLogFilename

switch len(args) {
    case 0:
       // nothing to change
    case 1:
      // Only an output file specified
      output = args[0]
    case 2, 3:
      // Path scope, output file, and maybe a format specified
      path = args[0]
      output = args[1]
      if len(args) > 2 {
          format = strings.Replace(args[2], "{common}", CommonLogFormat, -1)
          format = strings.Replace(format, "{combined}", CombinedLogFormat, -1)
      }
    default:
        // Maxiumum number of args in log directive is 3.
 	return nil, c.ArgErr()
}
rules = appendEntry(rules, path, &Entry{
		Log: &httpserver.Logger{
			Output: output,
			Roller: logRoller,
		},
		Format: format,
	})
```

cez81: Yes, that looks pretty neat :)

mholt:
Thank you both! :+1:

Issue #1445, opened 2017-02-17
markdown template auto reload for local develop


jet10000:
### 1. What version of Caddy are you running (`caddy -version`)?

0.9.5

### 2. What are you trying to do?

markdown template can't auto refresh when I change it.

mholt:
FYI, for anyone who wants to contribute: this involves changing markdown's behavior to work like templates does, where it loads and parses the template file at request-time instead of during startup.

francislavoie:
I'm thinking an option like `no_cache_template` to implement this would be good. Maybe you have a better idea for the option name @mholt?

mholt:
I was wondering if we should just change markdown's logic so that it loads and parses the template at every request like `templates` does. It's not efficient for high-load servers (I mean thousands of page loads per second), but med-low loads will not even sweat it. There's another issue #1489 about optimizing this, but I'd rather changing the template file just works for now.

mholt:
Markdown should now be loading the template at every page load (for now). Thanks, Tw!

PR #1682, opened 2017-05-22
markdown: reload template on each request and fix fake tests

tw4452852:
This implements the control of markdown's template as said in [this comment](https://github.com/mholt/caddy/issues/1445#issuecomment-294700924).

While working with this issue, I found that markdown's tests won't failed even its template file is missing.
After tracking this down, this issue even exists dating back to ac4fa2c, it seems missed markdown's testdata when migrating.

add template control option and fix fake tests issue.

#1445 

we should inform the user about this option.

### 4. Checklist

### DIFF
--- a/caddyhttp/markdown/markdown_test.go
+++ b/caddyhttp/markdown/markdown_test.go
@@ -1,16 +1,12 @@
 package markdown
 
 import (
-	"bufio"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
-	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 	"text/template"
-	"time"
 
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 	"github.com/russross/blackfriday"
@@ -79,19 +75,23 @@ func TestMarkdown(t *testing.T) {
 		}),
 	}
 
-	req, err := http.NewRequest("GET", "/blog/test.md", nil)
-	if err != nil {
-		t.Fatalf("Could not create HTTP request: %v", err)
+	get := func(url string) string {
+		req, err := http.NewRequest("GET", url, nil)
+		if err != nil {
+			t.Fatalf("Could not create HTTP request: %v", err)
+		}
+		rec := httptest.NewRecorder()
+		code, err := md.ServeHTTP(rec, req)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if code != http.StatusOK {
+			t.Fatalf("Wrong status, expected: %d and got %d", http.StatusOK, code)
+		}
+		return rec.Body.String()
 	}
 
-	rec := httptest.NewRecorder()
-
-	md.ServeHTTP(rec, req)
-	if rec.Code != http.StatusOK {
-		t.Fatalf("Wrong status, expected: %d and got %d", http.StatusOK, rec.Code)
-	}
-
-	respBody := rec.Body.String()
+	respBody := get("/blog/test.md")
 	expectedBody := `<!DOCTYPE html>
 <html>
 <head>
@@ -99,7 +99,6 @@ func TestMarkdown(t *testing.T) {
 </head>
 <body>
 <h1>Header for: Markdown test 1</h1>
-
 Welcome to A Caddy website!
 <h2>Welcome on the blog</h2>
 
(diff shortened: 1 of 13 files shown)


mholt:
Hey Tw, so sorry I haven't gotten a chance to look at this yet! It's kind of a big change, I will try to get it to it very soon. :)

tw4452852:
@mholt Never mind :smile:

caddyhttp/markdown/setup.go
```diff
@@ -115,28 +117,48 @@ func loadParams(c *caddy.Controller, mdc *Config) error {
...
+
+		paths, err := filepath.Glob(pattern)
+		if err != nil {
+			return c.Errf("glob %q failed: %v", pattern, err)
+		}
+		for _, path := range paths {
+			mdc.TemplateFiles[filepath.Base(path)] = path
+		}
+		return nil
+	case "no_cache_template":
```
mholt: I wonder if we should change this to `cache_templates`, so that the default behavior mirrors that of `templates` directive, which is to not cache them. That would make more sense to me, anyway.

caddyhttp/markdown/template.go
```diff
@@ -52,6 +61,7 @@ func SetTemplate(t *template.Template, name, filename string) error {
 	// Read template
 	buf, err := ioutil.ReadFile(filename)
 	if err != nil {
+		println(filename)
```
tw4452852: Ah, yes.

caddyhttp/markdown/setup.go
tw4452852: Make sense.

tw4452852:
> Doing this caching seems to add considerable complexity to the code. How necessary do you think it is that we have it?

Though it adds some sort of complexity, but it give us some performance benefits as well (Parsing template file takes time). I think users may disable template cache in developing environment, because they may adjust their template files again and again, once they are ready to go to product environment, they may enable template cache because templates are fine enough. This also improve the performance as said above.

> Should we leave caching to some other component of Caddy (some other plugin), rather than trying to bake caching into the markdown directive? 

I have thought of this, but can't find a general way for now. Maybe this could also be used by `template`.

> It would simplify a lot of code here, and we could still keep the part of this change so that templates are loaded and parsed at every request, to be consistent with the templates directive. What do you think?

I have just split the previous commit. The first one is fixing the fake tests and the other one is for template cache option. Feel free to use them.

mholt:
Okay, phew, I finally am getting around to this and the other PRs. (Gearing up for the 0.10.4 release!) So sorry about the long delay.

I'm not sure how you'll feel about this, but I think for now we should remove the code that controls whether to cache the templates. At this point, I prefer simplicity, and I am hoping that caching will be performed in a general way by some other middleware, which would cover the parsing of templates. But I do agree with what you're saying.

So: I'm ready to accept this PR, but first let's get rid of the change(s) to the Caddyfile so the only change in this PR is that templates are parsed on each page load, much like how the `templates` directive already works. Yes, caching would help here, but if anyone's site is that high-traffic anyway, they should be behind a CDN or something.

OH -- and thanks for fixing/restoring the tests! Let's keep those changes too. :smile:

tw4452852:
@mholt Alright, I have just amended the second commit for reloading the template simply.

caddyhttp/markdown/markdown_test.go
```diff
@@ -79,27 +77,30 @@ func TestMarkdown(t *testing.T) {
...
+		}
+		if code != http.StatusOK {
+			t.Fatalf("Wrong status, expected: %d and got %d", http.StatusOK, code)
+		}
+		return rec.Body.String()
 	}
 
-	respBody := rec.Body.String()
+	respBody := get("/blog/test.md")
 	expectedBody := `<!DOCTYPE html>
```
mholt: respBody and expectedBody are ineffective assignments, according to the CI tests.

tw4452852: Done.

caddyhttp/markdown/markdown_test.go
```diff
@@ -79,27 +77,33 @@ func TestMarkdown(t *testing.T) {
...
+		}
+		rec := httptest.NewRecorder()
+		code, err := md.ServeHTTP(rec, req)
+		if err != nil {
+			t.Fatal(err)
+			return ""
+		}
+		if code != http.StatusOK {
+			t.Fatalf("Wrong status, expected: %d and got %d", http.StatusOK, code)
+			return ""
```
mholt: Lines after a `Fatalf` call won't be executed. You can remove these returns! ;)

tw4452852: But it seems `gometalinter` recognizes this as ineffective assignments. Don't know why...

mholt: Oh, really? So if you take out these returns, gometalinter throws the ineffective assignment error?

That's probably a bug in ineffassign. (Is it reproducible?) I won't require it of you, and will merge this PR as-is, but you are welcome to open a bug [upstream in the ineffassign repo](https://github.com/gordonklaus/ineffassign).

So, let me know if this is ready to be merged then, and I will do it.

tw4452852: Yes, I think this is a bug in ineffassign, this is the reason of my [last commit](https://github.com/mholt/caddy/pull/1682/commits/655e61ab32a72831e897a46cf8763577fa719612).

mholt:
Thanks as always, Tw! You do great work. I'm super glad you're around, and the community is too.

tw4452852:
Hi @mholt . Today I just try to figure out the possible bug in ineffassign. As a result, there is nothing wrong with it. There is a real ineffectual assignments in test and I have just make a pr(#1734) to fix it.

BTW, I still don't understand why CI passed after I just add the unnecessary returns. But on my local environment, it doesn't pass. Maybe there are some bugs in CI, I guess...

mholt:
@tw4452852 Sometimes, I don't have an answer. 🤷‍♂️  Thanks for working on it! :+1:

PR #1686, opened 2017-05-24
Add grpcwebproxy plugin

pieterlouw:
New directive for the `http` server type called `grpcwebproxy`

N/A

No document changes needed. Link to plugin's documentation --> [here](https://github.com/pieterlouw/caddy-grpcwebproxy/blob/master/README.md)


abiosoft:
Wow this is awesome. 👏

mholt:
What if it was just called `grpc`? Would there be anything other than a GRPC web proxy?

pieterlouw:
Hi Matt,

That's a good question.

My initial idea was to have it as a grpc-web client to normal grpc server
proxy,but it could just as well serve as a transparent grpc client
(mobile,IoT, etc) to grpc server proxy as well.

The other implementation in the grpc ecosystem is grpc-gateway (REST to
grpc proxy),but I'm not exactly sure how and if that could be dynamically
implemented in the plugin

Pieter

mholt:
In general, Caddy plugins should be as cohesive as possible (meaning that a `grpc` plugin would handle all things GRPC related), unless there are some crazy dependency reasons or some other extenuating circumstance that requires they be split into different plugins/directives. (Remember a plugin can also register more than one directive. But I dunno if you'd want to do that if configuring a GRPC proxy and gateway are similar.)

Simpler directive names are usually better as well.

Once you choose the name, it's not easily changed, so choose wisely. I think I might lean more toward `grpc` than `grpcwebproxy` personally.

pieterlouw:
Naming it `grpc` makes more and more sense to me.

pieterlouw:
Closing this PR as I created a new one [here](https://github.com/mholt/caddy/pull/1694) with plugin name as `grpc` instead

Issue #1683, opened 2017-05-22
Exposing TLS certicate metrics


miekg:
This is a question/feature request.

For caddy-prometheus (github.com/miekg/caddy-prometheus) I would like to expose metrics from the certs caddy has configured. Like time left per cert, failures to update, etc. etc.

Is there an API for this, should caddy expose some of these bits or should I hack in into ACME (? believe that is used for handling LE comms)?

elcore:
Hello @miekg,

this sounds like a great idea! 

I believe the best approach would be to expose those metrics in `github.com/xenolf/lego/acme` -- It might be useful for other projects using `lego/acme`

Basically you should "hack in into ACME" 😄

__ 

On the other side, I believe it would be easier to create those metics directly in Caddy and expose them to `caddy-prometheus`

miekg:
[ Quoting <notifications@github.com> in "Re: [mholt/caddy] Exposing TLS cert..." ]
>Hello @miekg,
>
>this sounds like a great idea!
>
>I believe the best approach would be to expose those metrics in `github.com/xenolf/lego/acme` -- It might be useful for other projects using `lego/acme`
>
>Basically you should "hack in into ACME" 😄

Interesting thought. I'll look into that, there could either be an API in ACME,
or directly expose prometheus metrics. I'll open an issue there.

>On the other side, I believe it would be easier to create those metics directly in Caddy and expose them to `caddy-prometheus`

But how do I get access to the information that those metrics will expose?

/Miek

--
Miek Gieben

elcore:
>But how do I get access to the information that those metrics will expose?

You could import `caddy/caddytls` in `caddy-prometheus` and use those metrics -- I have not really thought deeply about this ..........

elcore:
> Interesting thought. I'll look into that, there could either be an API in ACME,
or directly expose prometheus metrics. I'll open an issue there.

Awesome :smile:

mholt:
I think we can do something like this; Caddy can [emit events](https://godoc.org/github.com/mholt/caddy#EmitEvent) but right now it only emits one (startup) because we're adding events based on need. I don't have a page on the wiki yet explaining how to hook into events (I'll do that soon) but it's very easy; you can look at the [caddy-service plugin and see how it does it](https://github.com/bruhs/caddy-service/blob/5675e04a449b33b202aca35c4bbd3e2dd33636ac/service.go#L21).

[...]

PR #1692, opened 2017-05-26
[Breaking Change] Add ExtraInfo to EventHook

hacdias:
Add ExtraInfo to EventHook.


hacdias:
I think so @elcore! I chatted with Matt yesterday and he told me we needed to change the signature of `EventHook` function so extra information could be passed. I haven't read #1683 until now but it must be it 😄

mholt:
Thanks @hacdias ! Would you mind updating the hook.service plugin when this is released?

PR #1693, opened 2017-05-27
Register authz plugin.

hsluoyz:


mholt:
Hey Yang, thanks so much for opening this pull request! Always glad to see people writing tools that integrate with Caddy.

I agree with Peer: the pull request itself looks good, but the plugin will need more documentation on its readme or something. It's not clear how a Caddy user would go about using this.

hsluoyz:
Hi @mholt ,

I have added the documentation section in [README](https://github.com/casbin/caddy-authz#documentation). In fact the authorization plugin [caddy-authz](https://github.com/casbin/caddy-authz) is a wrapper for the more general authorization library [Casbin](https://github.com/casbin/casbin). So I also add a pointer to Casbin's documentation in the plugin's documentation.

mholt:
I did see the existing README, but that's Go code, it doesn't actually explain how to use the Caddy plugin with Caddy.

mholt:
@hsluoyz I am ready to merge this as soon as the plugin docs are more geared towards Caddy users. :+1:

hsluoyz:
Hi @mholt , thanks for the concern. I have improved the docs here: https://github.com/casbin/caddy-authz :)

mholt:
Great, thanks. The Caddyfile example is much better than what was missing before. :)

mholt:
@hsluoyz You may now deploy your plugin!

hsluoyz:
Hi @mholt , please point me to how to deploy a plugin?

francislavoie:
@hsluoyz https://github.com/mholt/caddy/wiki/Publishing-a-Plugin-to-the-Download-Page

hsluoyz:
I think I have already deployed it by registering it at ``Developer Portal account``?

mholt:
@hsluoyz Oh, yes, sorry -- it's already on the Download page, so I must have missed that it was already up.

tobya:
I know I'm a bit late to the party here, but I'm thinking about naming. We now have quite a few 3rd party Authenticator plugins which is awesome but the names all end up being very general. 

 - reauth 
 - basicauth
 - login 
 - authz 

Now might be the time to start asking for better names. 

In this instance I would like to ask @hsluoyz to perhaps rename his plugin & directive to one of the following or something similar 

 - authcasbin
 - auth_casbin
 - casbin

If everyone else thinks I am nitpicking that's fine but I think this is a problem easier to solve early

hsluoyz:
AFAIK, reauth, basicauth, login are all authentication plugins (aka authn). Only Casbin is the authorization (aka authz). So I think there's no much overlapping between Casbin and other plugins.

I think we should first separate the plugins into two categories: authn and authz, and rename them. Then discuss about the renaming inside one category.

Of course, if there are multiple authz plugins, I'm OK to rename Casbin into other names, like authz_casbin.

mholt:
Good discussion here. Renaming plugins is very tricky and I strongly discourage it. I think we're okay so far. I'm trying to avoid having the plugins become a "marketplace", but rather keep them consolidated so there is one plugin that does a certain thing, not several.

PR #1694, opened 2017-05-28
Add grpc plugin directive

pieterlouw:
New directive for the http server type called `grpc`

N/A

No document changes needed. Link to plugin's documentation --> [here](https://github.com/pieterlouw/caddy-grpc/blob/master/README.md)

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -490,6 +490,8 @@ var directives = []string{
 	"hugo",      // github.com/hacdias/caddy-hugo
 	"mailout",   // github.com/SchumacherFM/mailout
 	"awslambda", // github.com/coopernurse/caddy-awslambda
+
+	"grpc", // github.com/pieterlouw/caddy-grpc
 }
 
 const (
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/plugin.go
```diff
@@ -490,6 +490,8 @@ var directives = []string{
 	"hugo",      // github.com/hacdias/caddy-hugo
 	"mailout",   // github.com/SchumacherFM/mailout
 	"awslambda", // github.com/coopernurse/caddy-awslambda
+
```
mholt: Let's take out this extra line.

pieterlouw: What is the best way to determine where in the chain are the best place to put the directive?

mholt: Where ever it works correctly and doesn't break other middlewares in the chain. We can always move it later, but I always ask to double check and make sure that the author at least considered that the plugin is in the right place. Whatever makes the most logical sense in the order of things done by handlers, really.

pieterlouw: Okay. I have left it at the bottom of the chain for now. I tested it along with tls,log and prometheus plugins on my dev machine and it all works fine

mholt:
Cool, thanks. Just fix the spacing with the comment (gofmt -s) and I'll merge this in! (Check CI failures)

mholt:
@pieterlouw You may now deploy your plugin to the Caddy website!

pieterlouw:
Thanks @mholt! I just registered it now.

What's the steps if there's a new version of the plugin? Redeploy and then users need to build a new download of Caddy?

mholt:
@pieterlouw Yep.

PR #1696, opened 2017-05-31
Rewrite: treat "if a not_op b" uniformly by negating "op".

lbogdan:

As there were some inconsistencies with `if` conditions in `rewrite` directives (conditions not having the `not_cond` counterpart, like `starts_with`, fixed by Taylor [here](https://github.com/mholt/caddy/pull/1687), but also `ends_with`, which is not yet "fixed"), I took a stab at adding a `not` unary operator to `if` conditions. This change adds support for conditions like `not a cond b`, which is equivalent with `a not_cond b`.
All `not_cond` conditions were left unchanged and still work.

Please note that this is my first ever golang contribution, so there might be some non go idiomatic bits.

https://github.com/mholt/caddy/pull/1687

The `rewrite` directive documentation at https://caddyserver.com/docs/rewrite#if-conditions needs updating, I'll do it if this gets merged.

### 4. Checklist


mholt:
Hi @lbogdan - sorry I didn't get to this until now!

I like the idea of consistency among the if statements, although I'm not sure I like the _inconsistency_ of having `not` out front. What's wrong with the operators like `if a not_op b` that necessitates a need for `if not a op b`?

abiosoft:
Thanks for this @lbogdan.

I considered this initially when if conditions were being added but @mholt wants this to be as simple as possible and expresses the fear of it becoming more complex and eventually turn to full scripting.
Adding `not` keyword is a step towards validating his fear.

I think we can have `not_` prefix for all the conditions. i.e. intelligently strip the `not_` prefix and negate the condition outcome. That can be more generic than explicitly handling a negative version of all conditions.

mholt:
At this point I think I'm OK with adding `not_` variants of all the operators -- but I'd rather not change the syntax.

lbogdan:
I think I got a bit confused because I needed not_starts_with and couldn't find it, so I thought you should be able to negate conditions. :-) But I completely get your argument about wanting to keep the syntax as simple as possible.

So would it make sense to rewrite this using a not_op b syntax? It should be an easy change and would remove the "duplicated" conditions.

mholt:
Yup, use `a not_op b` syntax instead of introducing the other `not` syntax. I'll continue to leave this open for @abiosoft to review after that!

lbogdan:
@mholt @abiosoft you guys need anything from me to get this reviewed?

mholt:
Sorry, been really busy lately -- haven't had time to do anything on the computer... What's the syntax now? We settled on standardizing with `not_` prefixes, right? And not changing the number of arguments or types of operators, etc?

abiosoft:
Yeah, the syntax is exactly the same. The only addition is that every condition can now be prefixed with `not_`. e.g. `if a match b` can be negated with `if a not_match b` e.t.c.

mholt:
Thank you @lbogdan! If you have a chance, I'd welcome a PR to update the docs on https://github.com/caddyserver/website.

lbogdan:
@mholt should this be stated explicitly, or should it be just an implementation detail? If the latter, then we should only add `not_ends_with`, which is not documented right now.

mholt:
@lbogdan I think each one should be stated explicitly. (Did you mean "if the former"?)

lbogdan:
Done. Also added a note about the general rule of negating a condition by prefixing it with `not_`.

PR #1699, opened 2017-06-01
add optional 'ca' tls directive, closes #1689

### LINKED ISSUE #1689: TLS CAUrl Option
Currently the ACME CA URL is configured via the `-ca` command line argument. This means changing the value involves restarting the process and that all domains must use the same ACME CA.  I think this option would be better suited for the `tls` directive, e.g.:

```
staging.mydomain.com {
  tls {
    ca https://acme-staging.api.letsencrypt.org/directory
  }
  ...
}

other.mydomain.com {
  tls {
    ca https://acme-v01.api.letsencrypt.org/directory
  }
  ...
}
```

I think I can whip up a PR for this idea if it has value.

zikes:
This change adds an optional `ca` option to the `tls` directive, which lets you specify an ACME endpoint within the Caddyfile.

#1689 

The new option should be added at https://caddyserver.com/docs/tls

### 4. Checklist


caddytls/setup_test.go
```diff
@@ -277,6 +277,25 @@ func TestSetupParseWithClientAuth(t *testing.T) {
 	}
 }
 
+func TestSetupParseWithCAUrl(t *testing.T) {
+	testUrl := "https://acme-staging.api.letsencrypt.org/directory"
```
elcore: Rename 'testUrl' to 'testURL'

zikes: @elcore Updated and squashed

mholt:
Welcome @zikes! Thanks for your contribution. I'll give this a second review before long. :)

caddytls/setup.go
```diff
@@ -66,6 +66,12 @@ func setupTLS(c *caddy.Controller) error {
 		for c.NextBlock() {
 			hadBlock = true
 			switch c.Val() {
+			case "ca":
+				arg := c.RemainingArgs()
```
mholt: We should use only the first argument; if there are more or less than 1 argument, it should be an error.

elcore: Sounds good to me :)

zikes: I've replaced the check with `len(args) != 1` similar to how `dns` and `storage` check their args, and added test cases for too many/too few args.

mholt:
Thank you @zikes! If you have a chance, submit a PR to https://github.com/caddyserver/website to add the docs for the new `ca` option. :)

PR #1702, opened 2017-06-03
Configure systemd to send SIGQUIT on stop

gmacon:

This updates the systemd unit to configure the use of SIGQUIT to trigger a graceful shutdown when the service is stopped.

I didn't create an issue for this, but I did look at open issues matching "systemd", and none of them appeared to be relevant.

The exact behavior of `systemctl start caddy`, `systemctl reload caddy`, and `systemctl stop caddy` are not currently discussed in the README. This modification does not, in my opinion, violate the principle of least surprise for the behavior of `systemctl stop caddy`.

### 4. Checklist


mholt:
I haven't forgotten about this PR, @gmacon -- thanks for your contribution. I'll catch up on PRs later this week, or more likely next week as I won't be on the computer much for a bit.

PR #1703, opened 2017-06-04
Update Initscripts

messyidea:
"$(which caddy)" is not work at system startup.
After this change, I can run "insserv -d caddy" to start caddy automatically on boot.


djvdorp:
LGTM!

PR #1709, opened 2017-06-09
using caddy.AppName as a replacement for the fixed ’Caddy’ string in …

malud:
…‘Server’ response-header field

Sets the variable `caddy.AppName` as replacement for the fixed string `Caddy` as value for the `Server`-header-field which is tracked down to the const variable caddymain.appName which value is also `Caddy`. In summary not a change at all but enables someones custom caddy build to overwrite the caddy.AppName and thus the `Server` Reponse-Header-Field to some custom (dynamic) value without writing a plugin to have more options for e.g. the `header` directive.

n/a

n/a

### 4. Checklist


mholt:
Thank you @malud!

Issue #1590, opened 2017-04-18
Proposal: Add a request id placeholder


GlurG:
Something similar to what nginx already provides: [$request_id](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_id).
In a microservice design it is common to assign each request an id (also called correlation id) so an error can be traced through all services processing the request.

tobya:
@GlurG Thank you for the feature request.

Can you provide a bit more detail as to how this would be used?  Do you really mean it to be a placeholder to be used in logs and proxy urls?

Would it be added to headers anywhere?

Could you give some more detail about how you would envisage it being used?

GlurG:
Sure,

I'm currently using the $request_id for nginx in the following way:

```
upstream api-gateway {
	server			api-gateway:3000;
}

location /api {
	proxy_pass		http://api-gateway;
	proxy_set_header	X-Request-Id $request_id;
}
```
This way, nginx will generate a unique identifier for each request to /api (which is my [api gateway](http://microservices.io/patterns/apigateway.html)). The header "X-Request-Id" is extracted in the microservice and passed along to all subsequent calls to other microservices. If an error occurrs during the request, the logs will contain the request-id so I can trace all calls between the microservices which generated the error.
You can read more about loggin/correlation ids here: https://homeadvisor.tech/software-antipatterns-microservices/ (under "Logging Can Wait Pitfall")

mholt:
Is generating a request ID just a matter of using math/rand to generate a random string or number of  16 bytes? That's easy enough.

We'd either want to do this in replacer.go where we only attach the request ID to the context if needed (although we might run into the request dereferencing issues holding up #1542) or we should generate it for all requests always in httpserver/server.go and attach it to the request there (which will be safe -- will it be fast?).

tobya:
There seems to be a go package for generating uuids 

https://github.com/nu7hatch/gouuid

Perhaps we could use that.  I think I would favour generating on usage if possible.

tobya:
A few thoughts.

With the issues @mholt raises in mind I was wondering if it was worth having a `request_id` directive?  The requestid could then be generated in `server.go` but only for server blocks that needed it, rather than for ever request - which would be an unnecessary overhead.  `{request_id}` would return '' (or other empty  placeholder value} unless the `request_id` directive was also used.  

eg.  This would be used in the Caddyfile

    example.com {
      root c:\website
      request_id
      header X-Request-Id {request_id}
     }

[...]

PR #1711, opened 2017-06-10
Create request_id directive #1590

tobya:

Provides a `request_id` directive which when present in Caddyfile will populate a {request_id} placeholder which can then be used to follow a request through proxies etc.

This implements #1590 

I will add documentation to Caddyserver for this.

### 4. Checklist

### Other details

I use a third party library to generate the UUID used as the request id.  With Vendor inclusion we may need to judge if this is the right library to uses.

Caddyfile example

```
courgette.kinchik.ie{
  
  gzip
  log / d:\caddy\logs\courgettelog.log   "[{when}] {request_id} {path}  {uri}" 
  root d:\Development\websites\courgette
  browse 
  
  # add upstream header with request_id
  header /  "X-Request-Id" {request_id}

  # Switch on request_id generation
  request_id 

}
```
if `request_id` is not present {request_id} will be blank
I am not sure whether tests are required.


caddyhttp/httpserver/server.go
```diff
@@ -333,6 +334,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Server", "Caddy")
 
+	if requestid.IsActive {
+		reqid := requestid.UUID()
+		c := context.WithValue(r.Context(), caddy.CtxKey("REQUESTID"), reqid)
+		r = r.WithContext(c)
+	}
```
mholt: - Move this into an HTTP handler like every other middleware should be; you can then lose the whole `IsActive` variable.
- I generally try to avoid storing stuff in the context if possible, but I think a request ID is one of the few things that would be expected to belong in a context. So that is OK. However...
- ... There's a little trick you can use to set a value on a replacer. The proxy middleware, for example, defines an `{upstream}` placeholder like this:

```
if rr, ok := w.(*httpserver.ResponseRecorder); ok && rr.Replacer != nil {
	rr.Replacer.Set("upstream", host.Name)
}
```

You probably should do this for request ID, and then you can also add it to the context as well.

caddyhttp/httpserver/server.go
```diff
@@ -333,6 +334,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Server", "Caddy")
 
+	if requestid.IsActive {
+		reqid := requestid.UUID()
+		c := context.WithValue(r.Context(), caddy.CtxKey("REQUESTID"), reqid)
```
mholt: Use a context key of `"request_id"` instead.

caddyhttp/requestid/requestid.go
```diff
@@ -0,0 +1,43 @@
+package requestid
+
+import (
+	"fmt"
+
+	"github.com/mholt/caddy"
+	"github.com/nu7hatch/gouuid"
```
mholt: This dependency will have to be vendored. Use https://github.com/FiloSottile/gvt to do it.

caddyhttp/httpserver/plugin.go
```diff
@@ -488,6 +488,7 @@ var directives = []string{
 	"markdown",
 	"templates",
 	"browse",
+	"request_id",
```
mholt: Move this _way_ higher in the chain (like, after "shutdown"), and I'm thinking about whether this should be renamed to "requestid".

caddyhttp/requestid/requestid.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	IsActive = true
+	return nil
+}
+
+// UUID returns U4 UUID
+func UUID() string {
+	u4, err := uuid.NewV4()
+	if err != nil {
+		fmt.Println("error:", err)
```
mholt: Log this instead, something like `log.Printf("[ERROR] generating request ID: %v", err)`

caddyhttp/httpserver/server.go
mholt: ... actually, I just realized that setting a value on the replacer isn't probably going to work since, once the directive is moved up to the beginning of the chain, it's quite probable that the ResponseWriter won't be a Replacer type (yet)... until, for example, the `log` middleware is executed.

So perhaps you should ignore that advice for now, continue on with using context.

tobya:
I'll have to leave this now, I have too much other stuff on.  I'll come back to it in the Autumn (Fall!) or if anyone else wants to move it forward feel free.

elcore:
I will work on this 😄

elcore:
@mholt please take a look at my changes 😄

elcore:
AppVeyor fails randomly, I fixed (I hope so) the Test in `caddy_test.go` 😄

tobya:
@elcore Thank you.  I didn't really expect anyone to take it up, but I am delighted you are.  I will make a few comments on the code I hope you dont mind.

caddyhttp/requestid/requestid.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// Handler is a middleware handler
+type Handler struct {
+	Next httpserver.Handler
+}
+
+func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
+	reqid := UUID()
+	c := context.WithValue(r.Context(), caddy.CtxKey("request_id"), reqid)
```
tobya: I think @mholt was hoping that there might be a constant created like there is here

https://github.com/mholt/caddy/blob/d5371aff2265540fb8600541bfdf3c3c4fc963e4/caddyhttp/httpserver/middleware.go#L207

    RequestID caddy.CtxKey = "requestid"

then you can do this

    c := context.WithValue(r.Context(), RequestID, reqid)

caddyhttp/requestid/requestid_test.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func TestRequestID(t *testing.T) {
+	request, err := http.NewRequest("GET", "http://localhost/", nil)
+	if err != nil {
+		t.Fatal("Could not create HTTP request:", err)
+	}
+
+	reqid := UUID()
+	c := context.WithValue(request.Context(), caddy.CtxKey("request_id"), reqid)
```
tobya: AS above

caddyhttp/httpserver/replacer.go
```diff
@@ -243,6 +243,11 @@ func (r *replacer) getSubstitution(key string) string {
 	case "{path_escaped}":
 		u, _ := r.request.Context().Value(OriginalURLCtxKey).(url.URL)
 		return url.QueryEscape(u.Path)
+	case "{request_id}":
+		reqid, _ := r.request.Context().Value(caddy.CtxKey("request_id")).(string)
```
tobya: See comment on RequestID.go

caddyhttp/requestid/requestid.go
elcore: :+1: sure

caddyhttp/httpserver/replacer.go
```diff
@@ -243,6 +243,11 @@ func (r *replacer) getSubstitution(key string) string {
 	case "{path_escaped}":
 		u, _ := r.request.Context().Value(OriginalURLCtxKey).(url.URL)
 		return url.QueryEscape(u.Path)
+	case "{request_id}":
+		reqid, _ := r.request.Context().Value(RequestIDCtxKey).(string)
+		if reqid != "" {
+			return reqid
```
mholt: You can elide the `if` check for empty string, because if it's not a value that is set, the string will be empty anyway.

caddy_test.go
```diff
@@ -142,13 +142,13 @@ func TestListenerAddrEqual(t *testing.T) {
...
-		{ln1, ":1234", false},
-		{ln1, "0.0.0.0:1234", false},
+		{ln1, ":" + ln2port, false},
+		{ln1, "0.0.0.0:" + ln2port, false},
 		{ln1, "0.0.0.0", false},
 		{ln1, ":" + ln1port, true},
 		{ln1, "0.0.0.0:" + ln1port, true},
 		{ln2, ":" + ln2port, false},
-		{ln2, "127.0.0.1:1234", false},
+		{ln2, "127.0.0.1:" + ln1port, false},
```
mholt: What do these changes have to do with the request ID?

caddyhttp/httpserver/replacer.go
elcore: Sure :)

caddy_test.go
elcore: I can create a separate PR if you want :)

mholt: Oh, does this have to do with the "random test failures"? Sorry, it wasn't clear to me before. This is fine.

mholt: Oh. Wait, the changes are gone...

mholt: Ahhh I see the new PR now, sorry. That's good. Thanks

tobya:
Thanks @elcore Great to get this in.

elcore:
It was a pleasure @tobya

mholt:
@tobya I forgot to add the docs page for `request_id`. Would you want to propose a PR and then I can review/refine it? https://github.com/caddyserver/website

tobya:
@mholt  I'll do that.

tobya:
@mholt While I'm at it the works here are a bit confused.  I changed request_id to requestid but not in both places.

The directive is `requestid` but the placeholder is `{request_id}` .  Should I update the placeholder to be `{requestid}` or does it not matter now?

sample working Caddyfile

    root c:\website
    requestid
    header /  "X-Request-Id" {request_id}


Probably best to make a decision before doing the docs

mholt:
@tobya Oh, man. I somehow missed that.

This is the first standard directive that's more than one word. That's why I was thinking of changing it to `requestid` but I don't remember actually confirming that we should. We used to have a multi-word directive (`maxrequestbody`) but it was removed and changed to `limits` which is still one word. The [wiki says](https://github.com/mholt/caddy/wiki/Things-HTTP-Middleware-Developers-Should-Know) that directives should be underscore_convention when there are multiple words. 

Anyway, either the change from `request_id` to `requestid` went unnoticed by me, or I don't remember confirming that we should change it. (Totally my bad either way.) Since it's not documented yet, we can still change it.

Do you think we should make the name of the directive `request_id` to be consistent with the context key, then?

tobya:
I dont mind, I can make both `request_id` . I will submit a PR today.  and update the changes to documentation also.

PR #1566, opened 2017-04-06
update systemd script to account for bug #1565

coolaj86:
update systemd script to account for bug #1565

PR #1719, opened 2017-06-16
Increase restart rate limit

### LINKED ISSUE #1718: systemd service file goes into bad state and won't restart caddy

`0.10.3`

I'm trying to restart caddy while testing new configs and after a while it will not restart.

```text

{{config.site.domains[0]}}:8080 {
  tls off

  root {{config.dev.path.devDir}}public
  gzip

  fastcgi / 127.0.0.1:9000 php
  rewrite / {
    r /(.*)
    to {path} index.php?url={1}&{query}
  }
}
```

I'm using Ubuntu 16.04 and restarting in the terminal with `systemctl restart caddy`.

@klingtnet posted this on https://github.com/mholt/caddy/issues/711#issuecomment-308885617

It looks like the settings for StartLimitInterval, StartLimitBurst are to conservative.

```
$ journalctl -xefu caddy
Jun 15 18:02:08 m253 systemd[1]: caddy.service: Unit entered failed state.
Jun 15 18:02:08 m253 systemd[1]: caddy.service: Failed with result 'start-limit-hit'.
Jun 15 18:02:10 m253 systemd[1]: caddy.service: Start request repeated too quickly.
```

A workaround is to [deactivate the start limit check](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=) until we decided which values StartLimitInterval, StartLimitBurst should be set to.

To reproduce the problem simply start caddy's service `systemctl start caddy` call `systemctl restart caddy` and it will fail to restart eventually (after ~10 retries on my machine).

klingtnet:
The previous setting caused the service to hit a rate-limit when it was
restarted more than 5 times in 24h.
Editing the Caddyfile and restarting the service could also easily
trigger this rate limit.
One could argue that users could simply call `systemctl reset-failed
caddy` to reset the rate-limit counter, but this is counterintuitive
because most users won't know this command and are possibly unaware that
they had hit a rate-limit.

The service is now allowed to restart 10 times in 10 seconds before
hitting a rate limit.
This should be conservative enough to rate limit quickly failing
services and to allow users to edit and test their caddy configuration.

This closes #1718

### DIFF
--- a/dist/init/linux-systemd/caddy.service
+++ b/dist/init/linux-systemd/caddy.service
@@ -6,11 +6,6 @@ Wants=network-online.target systemd-networkd-wait-online.service
 
 [Service]
 Restart=on-failure
-; Renamed to StartLimitIntervalSec for systemd>=230
-; Set StartLimitInterval=0 to disable rate limiting.
-; To reset the restart counter run 'systemctl reset-failed caddy'
-StartLimitInterval=10
-StartLimitBurst=10
 
 ; User and group the process will run as.
 User=www-data
--- a/dist/init/linux-systemd/caddy.service
+++ b/dist/init/linux-systemd/caddy.service
@@ -6,8 +6,11 @@ Wants=network-online.target systemd-networkd-wait-online.service
 
 [Service]
 Restart=on-failure
-StartLimitInterval=86400
-StartLimitBurst=5
+; Renamed to StartLimitIntervalSec for systemd>=230
+; Set StartLimitInterval=0 to disable rate limiting.
+; To reset the restart counter run 'systemctl reset-failed caddy'
+StartLimitInterval=10
+StartLimitBurst=10
 
 ; User and group the process will run as.
 User=www-data


dist/init/linux-systemd/caddy.service
```diff
@@ -6,8 +6,11 @@ Wants=network-online.target systemd-networkd-wait-online.service
 
 [Service]
 Restart=on-failure
-StartLimitInterval=86400
-StartLimitBurst=5
+; Renamed to StartLimitIntervalSec for systemd>=230
+; Set StartLimitInterval=0 to disable rate limiting.
```
mholt: I don't think we should recommend/encourage this...

klingtnet:
[systemd.unit](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=) states:
> .. Note that units which are configured for Restart= and which reach the start limit are not attempted to be restarted anymore; however, they may still be restarted manually at a later point, from which point on, the restart logic is again activated. ...
Unfortunately, it does not care what the services return value was, instead only the bare number of restarts is taken into account.

dist/init/linux-systemd/caddy.service
klingtnet: It was not thought to be a recommendation, rather a hint for user that want to debug their caddy configuration.

klingtnet:
@wmark Do you know if it is possible to disable restarts from being counted in `StartLimitBurst`?

Chronial:
I think this should just be removed and let systemd fall back to the default (5 attempts in 10 seconds). I see no reason why you would need to manually restart caddy more than 5 times in 10 seconds.

klingtnet:
@Chronial I reset both `StartLimit` settings to use the default restart limit.

wmark:
Please see also #1566.

We arrived at:
```diff
-Restart=on-failure
+Restart=on-abnormal
-StartLimitInterval=86400
-StartLimitBurst=5
+StartLimitInterval=10
+StartLimitBurst=3
```

(Please note I am for `Restart=no`, but then I acknowledge that `≠no` promises fewer opened issues with *Caddy's* target audience.)

Explicit is better than implicit in this case, for one due to the history of the file (existing users can spot the difference better) and then to be used as template. A shockingly high amount of operators wouldn't know how to get any or all default values, to compare. (`systemctl show caddy.service`)

klingtnet:
@wmark I will integrate the changes you quoted into this pull request because #1566 will probably not be merged anytime soon.
Is there a reason to use `StartLimitBurst=3` and not the default `StartLimitBurst=5`? I would propose to state

```
StartLimitInterval=10
StartLimitBurst=5
```

explicitly.

klingtnet:
@wmark Will systemd still count `systemctl restart caddy` calls to the `RestartLimit` when `Restart` is set to `on-abnormal`?
`systemctl restart` will send a `SIGHUP` which is ignored by `on-abnormal` so this should not increase the `RestartLimit` counter, right?

i-s-o-g-r-a-m:
I'd like to +1 this one. If you're new to systemd, as I am, you end up in a pretty confusing situation. Thanks!

wmark:
I still would change this to `Restart=on-abnormal`, everything else LGTM. (`systemctl restart …` should not be affected by this, or else it's a bug.)

mholt:
I'm not really sure what's going on with this PR anymore, since it seems a bit inactive, and there's no consensus. I'm gonna close it, but feel free to discuss it, work out something you agree on - and then we can re-open this. Thanks!

Chronial:
@mholt I would ask you to please just merge this Pull Request. There might be a lack of consensus here regarding the exact values, but everybody agrees that the current values are bad. This PR just makes the service fall back to the systemd defaults, which I would argue are probably an acceptable compromise till a better solution is agreed upon.

The current state is just really bad and I am sure made a lot of people loose quite a bit of time.

klingtnet:
@wmark I think too that [`Restart=on-abnormal`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=) is the way to go because it would not restart the service if it fails with a non-zero exit code. This looks reasonable. I will change this in my fork.

@mholt If you want to merge this then please wait until I made this last change mentioned above.

mholt:
@Chronial I'm sorry, but I can't merge it with the limits taken out. People have complained about systemd putting Caddy into a restart loop otherwise.

Chronial:
@mholt Without the limits, caddy will stop starting after 5 failed attempts in 10 seconds – how would that result in an infinite loop?

mholt:
@Chronial What makes it stop after 5 attempts in 10 seconds? I don't see anything in the systemd file that guarantees that.

Chronial:
Sry, you are right. What actually happens is that it will not restart more often than 5 times per 10 seconds. It will keep doing that forever. But I don't see how that's a big problem – that will not take up a lot of system resources. And is a lot better that the current situation in which caddy suddenly stops working while you're trying to configure it.

mholt:
@Chronial Let's Encrypt has complained about systemd misconfigurations that cause Caddy to hammer their servers.

Chronial:
@mholt Will Caddy contact the Let's Encrypt service on every startup or only after certain kinds of failures (failed to get a cert)? If yes, are these kinds of failures recognizable from the exit code?

I have the impression that this is something that can't be truly fixed by only the init service. Maybe caddy itself should have something build in that prevents it from hammering the let's encrypt servers, but at least it should report these kinds of issues in its CLI.

wmark:
@mholt I don't feel you're doing an assessment in good faith here. Let's Encrypt cannot tell what causes a hammering unless they can log into any machine. To me it seems most issues that arise from *systemd's* use are caused by committers tinkering with *service files* without having understood the extent of their modifications.

Any *start limits* are applied to automatic (like, on `on-abort`) as well as manual (`systemctl try-restart caddy`) restarts. Counting this over a window of 86400 seconds (24h) is ridiculous.

Excerpt from the description of the restart behaviour:

> Note that units which are configured for Restart= and which reach the start limit are not attempted to be restarted anymore;
> — https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=

mholt:
@Chronial This isn't a bug in Caddy. Caddy's updated exit codes only recently changed and haven't been released yet. When it is released in 0.10.7, an exit code of 1 should tell the launcher to NOT re-execute Caddy. That would be a better change than removing limits in systemd altogether.

Chronial:
@mholt I wasn't trying to imply that this is a bug in Caddy, just saying that this would be easier with some help from Caddy. But it sounds like that is already underway.

> That would be a better change than removing limits in systemd altogether.

Just to clarify: That is **not** what this PR is doing. This PR is just resetting these limits to their defaults, *not* removing them.

mholt:
The correct change for this PR, IMO, would be to see if the status code is 1. If so, do not restart. If not, wait a few seconds before restarting, and cap it at some maximum attempts.

klingtnet:
> would be to see if the status code is 1. If so, do not restart. If not, wait a few seconds before restarting, and cap it at some maximum attempts.

@mholt This is exactly the behaviour of `Restart=on-abnormal` and, for non-zero exit codes, it would stop caddy from restarting after 5 failed attempts in 10 seconds (which is systemd's default).

mholt:
@klingtnet 

> This is exactly the behaviour of Restart=on-abnormal and, for non-zero exit codes, it would stop caddy from restarting after 5 failed attempts in 10 seconds (which is systemd's default).

That sounds good, but right now the PR is "on-failure". Also, there are other exit codes where an immediate restart could be allowed (but still within limits to prevent looping) - it's just 1 where it should be forbidden at all, for optimal configuration.

klingtnet:
@mholt I deleted my comment because I was a bit two fast with the following conclusion _for non-zero exit codes, it would stop caddy from restarting_. I will check tomorrow if the changes let caddy's service restart as expected and reopen the PR if this is the case.

mholt:
@klingtnet Oh, sorry. GitHub sends me emails of all comments on this repo, and I didn't look higher. (I've had some problems with GitHub notifications in the past.) Even deleted comments appear on my news feed...

Thanks for looking into it!

klingtnet:
I tested the `Restart=on-abnormal` directive and the behaviour is as desired. Unfortunately, I can't reopen this PR because I rebased my branch on upstream master and force pushed the branch in my fork. See #1823 for the followup.

PR #1724, opened 2017-06-19
Fix memory leak

lostinblue:
I think the way you had it, you were initializing new variables and using more memory every time you iterated through the loop. I think that would cause a memory leak. This should achieve some slight memory improvements, also since err is being returned, we can set it during function initialization.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -606,16 +606,15 @@ func executeDirectives(inst *Instance, filename string,
 	return nil
 }
 
-func startServers(serverList []Server, inst *Instance, restartFds map[string]restartTriple) error {
+func startServers(serverList []Server, inst *Instance, restartFds map[string]restartTriple) (err error) {
 	errChan := make(chan error, len(serverList))
+	var (
+		ln  net.Listener
+		pc  net.PacketConn
+	)
 
 	for _, s := range serverList {
-		var (
-			ln  net.Listener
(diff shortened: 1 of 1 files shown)


caddy.go
```diff
@@ -606,16 +606,15 @@ func executeDirectives(inst *Instance, filename string,
...
-func startServers(serverList []Server, inst *Instance, restartFds map[string]restartTriple) error {
+func startServers(serverList []Server, inst *Instance, restartFds map[string]restartTriple) (err error) {
 	errChan := make(chan error, len(serverList))
+	var (
+		ln  net.Listener
+		pc  net.PacketConn
+	)
 
 	for _, s := range serverList {
-		var (
```
tw4452852: I don't think we could move these out of the loop, because we check `ln` and `pc` below (line 651 and 657).

lostinblue:
Thanks for clarifying, sorry about that.

lostinblue:
I thought about this more, and wouldn't you allocate less memory just zeroing out the variables at the end of the loop and using a single instance of the variables instead of creating a set for each time through the loop? Gains would be dependent on the active number of servers and number of resets.

mholt:
Not really, these are on the stack and the impact here is negligible. Try doing some profiling, you'll see what I mean.

PR #1731, opened 2017-06-27
Introduce `rotate_compress` option for log and errors directives in Caddyfile

fern4lvarez:

This PR addresses https://github.com/mholt/caddy/issues/1728 and provides a new `rotate_compress` option for `log` and `errors` directive to the Caddyfile, which will enable gzip compression for rotated log files.

It also updates the Lumberjack package dependencies in the `vendor` directory, and adds go-syslog, which was not present in `vendor` and was breaking tests when this was not present.

https://github.com/mholt/caddy/issues/1728

https://github.com/caddyserver/website/pull/18

### 4. Checklist

### DIFF
--- a/caddyhttp/errors/setup_test.go
+++ b/caddyhttp/errors/setup_test.go
@@ -85,7 +85,7 @@ func TestErrorsParse(t *testing.T) {
 				Roller: httpserver.DefaultLogRoller(),
 			},
 		}},
-		{`errors errors.txt { rotate_size 2 rotate_age 10 rotate_keep 3 rotate_compress true }`, false, ErrorHandler{
+		{`errors errors.txt { rotate_size 2 rotate_age 10 rotate_keep 3 rotate_compress }`, false, ErrorHandler{
 			ErrorPages: map[int]string{},
 			Log: &httpserver.Logger{
 				Output: "errors.txt", Roller: &httpserver.LogRoller{
--- a/caddyhttp/errors/setup_test.go
+++ b/caddyhttp/errors/setup_test.go
@@ -85,13 +85,14 @@ func TestErrorsParse(t *testing.T) {
 				Roller: httpserver.DefaultLogRoller(),
 			},
 		}},
-		{`errors errors.txt { rotate_size 2 rotate_age 10 rotate_keep 3 }`, false, ErrorHandler{
+		{`errors errors.txt { rotate_size 2 rotate_age 10 rotate_keep 3 rotate_compress true }`, false, ErrorHandler{
 			ErrorPages: map[int]string{},
 			Log: &httpserver.Logger{
 				Output: "errors.txt", Roller: &httpserver.LogRoller{
 					MaxSize:    2,
 					MaxAge:     10,
 					MaxBackups: 3,
+					Compress:   true,
 					LocalTime:  true,
 				},
 			},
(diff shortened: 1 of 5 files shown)


caddyhttp/errors/setup_test.go
```diff
@@ -85,13 +85,14 @@ func TestErrorsParse(t *testing.T) {
 				Roller: httpserver.DefaultLogRoller(),
 			},
 		}},
-		{`errors errors.txt { rotate_size 2 rotate_age 10 rotate_keep 3 }`, false, ErrorHandler{
+		{`errors errors.txt { rotate_size 2 rotate_age 10 rotate_keep 3 rotate_compress true }`, false, ErrorHandler{
```
mholt: Let's get rid of "true" here, since the presence of "rotate_compress" will make that obvious. (And lack of it is obviously "false".) In other words, the line would be "rotate_compress" by itself, no arguments.

caddyhttp/httpserver/roller.go
```diff
@@ -89,10 +99,14 @@ const (
 	// defaultRotateAge is 14 days.
 	defaultRotateAge = 14
 	// defaultRotateKeep is 10 files.
-	defaultRotateKeep   = 10
-	directiveRotateSize = "rotate_size"
-	directiveRotateAge  = "rotate_age"
-	directiveRotateKeep = "rotate_keep"
+	defaultRotateKeep = 10
+	// defaultRotateCompress is false.
```
mholt: This comment isn't super helpful, it can be removed. ;)

fern4lvarez:
@mholt re: broken dependency

I just reverted the commit that adds the `gopkg.in/mcuadros/go-syslog.v2` package to `vendor/` and run tests:

```
$ git revert 6722284                                            
[1728-log-rotate-compress 9e990b6] Revert "vendor: add go-syslog dep"
...
go test -v ./caddyhttp/httpserver               
# github.com/mholt/caddy/caddyhttp/httpserver
caddyhttp/httpserver/logger_test.go:15:2: cannot find package "gopkg.in/mcuadros/go-syslog.v2" in any of:
	/home/fernando/src/github.com/mholt/caddy/vendor/gopkg.in/mcuadros/go-syslog.v2 (vendor tree)
	/usr/local/go/src/gopkg.in/mcuadros/go-syslog.v2 (from $GOROOT)
	/home/fernando/src/gopkg.in/mcuadros/go-syslog.v2 (from $GOPATH)
FAIL	github.com/mholt/caddy/caddyhttp/httpserver [setup failed]
```

Of course, I can `go get gopkg.in/mcuadros/go-syslog.v2` and the tests would pass. But this dependency is not tracked into `vendor/`. If you think we should not added to `vendor/`, I'm happy to revert that commit :)

fern4lvarez:
@mholt bases in your comments:

* I agree that either we need to vendor `gopkg.in/mcuadros/go-syslog.v2` or not, this ain't part of this PR, so I removed it.
* No need to set `false` or `true` along with `rotate_compress`. If the option is declared into the Caddyfile, we want to compress logs.
* Get rid of `defaultRotateCompress`. We don't compress by default, hence it is `false`, so no need to create an wrapping var for this.

I'll also update caddyserver/website#18 accordingly.

PR #1740, opened 2017-07-02
iteration on init script for freebsd

richardbowden:
* uses more of the builtin functionality for starting and stopping of the process by using command and command_args along with procname
* removed -f from daemon as this was hiding error message that were sent to stdout on startup, now writing stdout to the logfile directly

for example, this was being hidden:

“Activating privacy features.. [www.domain.com] failed to get certificate: Error presenting token: Could not find the start of authority”

it now shows up in the log

* aded “caddy_env” to allow the setting of environment variables that caddy might need, for example when setting creds for “DNS Challenge”

* added a check to ensure caddy_config_path file exists

N/A

None! Documented in the init script

### 4. Checklist

### DIFF
--- a/dist/init/freebsd/caddy
+++ b/dist/init/freebsd/caddy
@@ -23,8 +23,18 @@
 # caddy_config_path (str):    Set to "/usr/local/www/Caddyfile" by default.
 #                             Defines the path for the configuration file caddy will load on boot
 #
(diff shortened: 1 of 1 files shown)


mholt:
Cool, nobody has objected, so I'll merge it in. Thanks for contributing!

PR #1746, opened 2017-07-05
proxy: Fix CI fail from format token in Error call

JoshHarmon:

`go vet` began failing after 078c991574647c27 and causing CI failures. 

This patch changes the Error call added in commit 078c991574647c27 to
an Errorf call to support the use of the %d and %s tokens that are needed.

https://travis-ci.org/mholt/caddy/jobs/248392875
https://ci.appveyor.com/project/mholt/caddy/build/2979

`caddyhttp/proxy/upstream_test.go:480::error: possible formatting directive in Error call (vet)`

None

### 4. Checklist
This is a ~one~four-character fix, so I'm pretty sure I can safely check these. ;)


mholt:
Thank you both, @JoshHarmon and @tw4452852 !

Issue #1749, opened 2017-07-06
Requesting dynamic client-side redirects


diazona:
I'm making this feature request to consolidate previous requests for dynamic redirects, by which I mean the ability to issue client-side redirects (HTTP 3xx status codes) based on matching the URL to a pattern - presumably a regular expression - not just a fixed string. `redir` already issues client-side redirects but cannot match a URL pattern, whereas `rewrite` can match patterns using regular expressions, but doesn't do client-side redirection, so neither of these existing directives are significant.

# Background and justification

This has been previously brought up in

- Issue #726 
- Issue #856 
- [One forum thread](https://caddy.community/t/rewrite-with-redirect-or-redirect-with-regex/600) related to #726 
- [Another forum thread](https://caddy.community/t/any-chance-of-dynamic-redirects/2261) (my own, which prompted opening this issue)

Those pages identify several situations in which dynamic redirection would be useful:

- redirect all URLs with a given file extension, with the new URL depending on the old one
- redirect all URLs under a directory tree, with the new URL depending on the old one
- stripping or adding terminal slashes
- redirecting URLs with single-digit months to have leading zeros

In each case internal rewriting as performed by `rewrite` is not suitable because it's necessary to signal a change in the page's location to search engines and other visitors.

All of these cases can be addressed with workarounds, but in some cases the workaround is rather impractical, like creating a long list of exact-match redirect rules ([1](https://caddy.community/t/rewrite-with-redirect-or-redirect-with-regex/600/4?u=diazona), [2](https://caddy.community/t/any-chance-of-dynamic-redirects/2261/5?u=diazona)). That wouldn't even work if, say, the destination URL is proxied on to a dynamic backend site where it's not possible to enumerate all valid destinations. (Though in that case one could probably let the backend handle the redirection, but it's not hard to imagine reasons that might be difficult.) So even though it could be argued that dynamic redirection isn't _necessary_, it would massively simplify the configuration in a wide variety of situation.

Plus, dynamic client-side redirects is something that webmasters typically expect a server to do. Most (all?) other popular web servers include this functionality. I think there are enough cases justifying it that it's worth implementing in Caddy too.

astei:
I've actually been looking to implement this feature, so I'm going to consider giving this a go.

mholt:
This is made much easier in Caddy 2.

Issue #856, opened 2016-05-29
Redirects and rewrites that prevent duplicate content


razor-x:
Caddy has limited tools for preventing [duplicate content](https://support.google.com/webmasters/answer/66359?hl=en).

Here is a common situation for a static site. Suppose I am serving `example.com` with the files

```
./index.html
./about.html
./example/index.html
```

Let's assume I will use `ext .html` in my Caddyfile, but most of the issues below still apply without using `ext`.
1. The content of `index.html` is served at `https://example.com/` AND `https://example.com/index.html`.
2. The content of `./example/index.html` is served at `https://example.com/example/` AND `https://example.com/example/index.html`. Caddy forces a redirect from `https://example.com/example` to `https://example.com/example/`
3. The content of `./about.html` is served at `https://example.com/about` AND `https://example.com/about.html`.
4. The location `https://example.com/about/` returns a 404 error.

Now, what can I do with Caddy to fix any of these?

Note that `rewrite` cannot solve any of these issues since rewrite does not support a redriect code (e.g., [like NGINX does](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite)).
1. I can manually redirect `/index.html` to `https://example.com/`.
2. I can manually redirect `/example/index.html` to `https://example.com/`, **BUT** `redir` is not dynamic, so I would need a manual redirect for EVERY page like this (intractable).
3. Same issue as 2.
4. I could add a rewrite rule so `https://example.com/about/` serves `https://example.com/about`, and I might be able to make this dynamic to handle all such cases, but then I've created a new duplicate content issue where the about page is served from 3 URIs.

Finally, suppose I want to be consistent and all decide all locations should be served without a terminal `/`. Then `https://example.com/about` works fine, but there is no way to serve from `https://example.com/example` since Caddy automatically redirects this to `https://example.com/example/`.

Or suppose I want to opposite: always have a terminal `/`. Then I'm back to adding manual rewrites for all paths and introducing more duplicate content issues.

I believe most or all of these issues could be solved by giving `rewrite` a redirect option and removing the forced directory redirect.

mholt:
Hey @razor-x, sorry it took me so long to get around to responding to this!

It seems that you are asking about dynamic redirects; that is only one of the suggestions on the linked article. The rest are mainly content-related, regardless of the web server being used.

If we give `rewrite` the ability to redirect, how would it differ from `redir`? Would we even need `redir` anymore? @abiosoft, what do you think.

[...]

Issue #726, opened 2016-04-05
How could caddy redirect from url by regexp to another domain ?


wspl:
Hi,

I want to redirect all urls like:
`https://mydomain.com/xxx.jpg`
to
`https://anotherdomain.com/xxx.jpg`
by regexp: `^.+\.jpg`

Redirect and Redir seems like cannot do this. Is there other way to realize this?

Thanks.

mholt:
Only rewrite supports regular expressions and other fancy rules. It seems that these features have been popular and no reports of slowness yet, so we may add it to redir eventually. What do you think @abiosoft? Can we do it without duplicating the code?

abiosoft:
Sure, there should be a way without code repetition. I can have a look.

mholt:
Abiola and I have discussed this a bit and we are willing to implement the functionality *if* we can find a good, elegant way to do it. We're open to well-thought-out proposals for how this would look in the Caddyfile, preferably without breaking `redir` as it currently stands.

abiosoft:
Adding to what Matt said, we have experimented with regexp and conditions in **rewrite** and we have gotten many positive feedbacks. With the number of requests for similar feature for **redir**, we are now willing to add the features to **redir**.

The challenge is supporting **rewrite** features (basically `if` and `regexp`) and ensuring it is backward compatible with the current **redir** syntax.

For example, how can **redir** support `if` and `regexp` and still support this syntax.
```
redir 307 {
	/foo     /info/foo
	/todo    /notes
	/api-dev /api       meta
}
```

wmark:
How about:

Left column: If ```!strings.HasPrefix(…, "/")``` then is a regexp.

Right column can have a scheme. Or start with ```//``` in which case the current request's scheme is prepended.

razor-x:
Related: #856.

mholt:
I think this is now possible with redir's advanced features, like:

```
redir 307 {
    if {path} ends_with .jpg
    to https://anotherdomain.com{uri}
}
```

Bregor:
But what can I do in case my backend is changed when old URLs are still flowing on the internet?
In my particular case, I ran single language blog with URLs like: `/posts/<title>/`, but now blog became multilanguage, and URLs changed to `/en/<title>/`, `/ru/<title>/` and so on.
Default language is Russian, so, for now, rewrite looks like:

    rewrite /posts {
      r (.*)
      to /ru{1}
    }

[...]

PR #1750, opened 2017-07-06
rewrite: Add client-side redirect support

astei:

This adds a new "redirect" directive to the complex rewrite directive. Providing an HTTP response code will force Caddy to issue an HTTP redirect instead of internally rewriting the request.

#1749, #726, #856 

Information about the new `redirect` directive should be added to https://caddyserver.com/docs/rewrite.

### 4. Checklist

### DIFF
--- a/caddyhttp/rewrite/rewrite.go
+++ b/caddyhttp/rewrite/rewrite.go
@@ -34,7 +34,16 @@ type Rewrite struct {
 // ServeHTTP implements the httpserver.Handler interface.
 func (rw Rewrite) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 	if rule := httpserver.ConfigSelector(rw.Rules).Select(r); rule != nil {
-		rule.(Rule).Rewrite(rw.FileSys, r)
+		if result := rule.(Rule).Rewrite(rw.FileSys, r); result == RewriteDone {
+			// Check if this is a complex rule that requires a redirect.
+			if complexRule, ok := rule.(*ComplexRule); ok {
+				if complexRule.RedirectCode != 0 {
+					// Issue the redirect
+					http.Redirect(w, r, r.URL.String(), complexRule.RedirectCode)
+					return 0, nil
+				}
+			}
+		}
 	}
 
 	return rw.Next.ServeHTTP(w, r)
@@ -81,6 +90,9 @@ type ComplexRule struct {
 	// Extensions to filter by
 	Exts []string
 
+	// Desired redirect code
+	RedirectCode int
+
 	// Request matcher
 	httpserver.RequestMatcher
 
@@ -89,7 +101,7 @@ type ComplexRule struct {
 
 // NewComplexRule creates a new RegexpRule. It returns an error if regexp
 // pattern (pattern) or extensions (ext) are invalid.
-func NewComplexRule(base, pattern, to string, ext []string, matcher httpserver.RequestMatcher) (*ComplexRule, error) {
+func NewComplexRule(base, pattern, to string, ext []string, redirectCode int, matcher httpserver.RequestMatcher) (*ComplexRule, error) {
 	// validate regexp if present
 	var r *regexp.Regexp
 	if pattern != "" {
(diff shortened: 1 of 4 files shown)


diazona:
That was quick! Thanks :-) Before the people "in charge" get in here, I noticed a few things that might be worth thinking about:

- If `rewrite` can now issue client-side redirects, does `redir` provide any functionality that `rewrite` (with `redirect`)  doesn't? (Other than backwards compatibility, of course)
- Should it be prohibited to have more than one `redirect` directive in a given `rewrite` block? I don't read Go so I could be wrong, but I don't see that the code checks for this.
- Will the `redirect` directive only work when a regular expression is provided for the rewrite, or can it apply to string-match rules too? If it only works with regexes, I'd consider that kind of unintuitive.

I don't intend any of this as negative criticism, just things that I think are worth having thought about before the change is finalized.

astei:
@diazona 

* `redir` can still serve as a convenient one-line directive for quick redirects.
* If more than one redirect directive is included, it'll simply use the last one it finds. (Even then, why would you use multiple redirect directives?)
* It will work not just for regex matches - you can use this for `if` matches as well, but this works best for regexp rewrites.

mholt:
Hey Andrew, thanks so much for taking the time to implement this. I have a few reservations, but nothing we can't work through.

I really don't believe the rewrite middleware should be issuing redirects. This behavior is confusing. Do we just need regular expression matching for `redir` instead? If so, we could consider having `redir` and `rewrite` share a large chunk of the parsing code so that they both have similar functionality.

astei:
@mholt 

I chose to implement this as a part of the `rewrite` directive as I felt it was the least disruptive approach I could take.

I wouldn't have an issue with refactoring `redir` and `rewrite` to share the vast majority of their code when I can find the time.

mholt:
Okay. If my understanding _is_ correct that all `redir` needs is support for regular expressions, I think we need to go that route to avoid weird bugs and confusing behavior. Rewrite for internal rewrites, redir for external redirects. Maybe all we need to do is share the regexp- and if-related code.

Want me to leave this PR open in the meantime?

diazona:
> Okay. If my understanding is correct that all redir needs is support for regular expressions, I think we need to go that route to avoid weird bugs and confusing behavior. Rewrite for internal rewrites, redir for external redirects. Maybe all we need to do is share the regexp- and if-related code.

Indeed, what I had in mind when proposing the feature request was to either give `redir` support for regular expressions, or consolidate both `redir` and `rewrite` into one directive that would handle both internal and external redirects. I left it open-ended because I wasn't sure how you'd prefer to have this implemented.

Sharing the request-matching code (regexp and conditionals) between `rewrite` and `redir` would be exactly in line with what I'd be looking for. I do think it makes sense, in general, to give the two directives equivalent power to select which requests they apply to.

astei:
@mholt I'll probably open another PR for this. For the moment, I'd like it if this PR stayed open until I can open a new one.

mholt:
Sounds good, consider this your active TODO item. Thanks for tackling this!! ❤️

astei:
Going to check in:

I've been busy with other things, which has understandingly limited my ability to work on this. I will hopefully be able to start working on the refactor this week.

mholt:
Thanks for the check-in. We appreciate it! (Likewise, I have a state holiday tomorrow so I'm off. :smile:)

mholt:
Hey @astei - just checking in. Let us know if you need help and would like someone else to finish this up...

astei:
I have moved into my dorm to start classes in a few days and don't have my computer yet. I will have it tomorrow.On Aug 4, 2017 9:03 PM, Matt Holt <notifications@github.com> wrote:Hey @astei - just checking in. Let us know if you need help and would like someone else to finish this up...

mholt:
Ah, I know how that goes. Good luck with the transition!

astei:
I finally have a little time to work on this.

mholt:
Great :smile: Looking forward to it.

mholt:
I think it's been long enough that perhaps we should close this for now. Much has changed in the last few months, including #1948, which might alleviate the need for these changes. Feel free to comment if this PR needs further discussion or another look, just closing to clear out my TODO list. Thanks!

Issue #1463, opened 2017-02-20
proxy: Add header and uri load balancing policies


vicanso:
Can caddy support more load balancing algorithms such as haproxy. 

- `round-robin` for short connections, pick each server in turn
- `leastconn`  for long connections, pick the least recently used of the servers
    with the lowest connection count
- `source` for SSL farms or terminal server
    farms, the server directly depends on the client's source address
- `uri` for
    HTTP caches, the server directly depends on the HTTP URI
- `hdr` the server
    directly depends on the contents of a specific HTTP header field
- `first` for short-lived virtual machines, all connections are packed on the
    smallest possible subset of servers so that unused ones can be powered
    down

abiosoft:
Anyone can propose a new policy. The requirement really is adding the implementation here https://github.com/mholt/caddy/blob/master/caddyhttp/proxy/policy.go#L21-L24.

mholt:
Caddy already supports round robin, least conn, and source (IP hash).

I think implementing `first` would also satisfy #1103. Anyone is welcome to implement that and submit a pull request, of course!

pbeckmann:
Hey,

i would implement these feature. But i've a question. Where do we want to get the number of "firsts" machine? 

I think we could make it in a static way, or about questioning an url for it.

Greetings.

mholt:
@pbeckmann Good question... I'm not sure how haproxy decides that. Anyone know?

pbeckmann:
@mholt 

I looked in the documentation of haproxy and found 

`first       The first server with available connection slots receives the
              connection. The servers are chosen from the lowest numeric
              identifier to the highest (see server parameter "id"), which
              defaults to the server's position in the farm. Once a server
              reaches its maxconn value, the next server is used. It does
              not make sense to use this algorithm without setting maxconn.
              The purpose of this algorithm is to always use the smallest
              number of servers so that extra servers can be powered off
              during non-intensive hours. This algorithm ignores the server
              weight, and brings more benefit to long session such as RDP
              or IMAP than HTTP, though it can be useful there too. In
              order to use this algorithm efficiently, it is recommended
              that a cloud controller regularly checks server usage to turn
              them off when unused, and regularly checks backend queue to
              turn new servers on when the queue inflates. Alternatively,
              using "http-check send-state" may inform servers on the load.`
Source: http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4.2-balance

[...]

PR #1751, opened 2017-07-07
add Header proxy policy

zikes:

Add a new proxy policy which routes to upstream based on the hash of a given request header's value.

Example:

```
proxy {
  policy header My-Header
}
```

This also changes the `Policy` registration process to allow policies to receive a single optional `string` argument.

#1463

The new policy would need added at https://caddyserver.com/docs/proxy

### 4. Checklist

### DIFF
--- a/caddyhttp/proxy/upstream.go
+++ b/caddyhttp/proxy/upstream.go
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	supportedPolicies = make(map[string]func() Policy)
+	supportedPolicies = make(map[string]func(string) Policy)
 )
 
 type staticUpstream struct {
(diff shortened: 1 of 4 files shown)


caddyhttp/proxy/upstream.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	supportedPolicies = make(map[string]func() Policy)
+	supportedPolicies = make(map[string]func(string) Policy)
```
abiosoft: Or maybe `func(args ...string) Policy` to make it dynamic in case of any future policy that may require more than one args.

zikes: Per @mholt's recommendation in #1463 it's just `string`, but changing it to `...string` would be fairly trivial.

mholt: This is fine, I'm rewriting this part of the proxy middleware now and will take it into account.

zikes:
I'll try to submit a PR to caddyserver/website soon to update the documentation for this.

mholt:
Perfect, thanks

PR #1753, opened 2017-07-08
gzip: change ETag to weak etag after gzip

spacewander:

According to https://tools.ietf.org/html/rfc7232#section-2.1
> Likewise, a validator is weak if it is shared by two or more
representations of a given resource at the same time, unless those
representations have identical representation data.  For example, if
the origin server sends the same validator for a representation with
a gzip content coding applied as it does for a representation with no
content coding, then that validator is weak.

Although it seems to be harmless at most time, follow the RFC would be appreciate.

Not yet.

No documentation need to be changed.

### 4. Checklist

### DIFF
--- a/caddyhttp/gzip/gzip.go
+++ b/caddyhttp/gzip/gzip.go
@@ -105,6 +105,10 @@ func (w *gzipResponseWriter) WriteHeader(code int) {
 	w.Header().Del("Content-Length")
 	w.Header().Set("Content-Encoding", "gzip")
 	w.Header().Add("Vary", "Accept-Encoding")
+	originalEtag := w.Header().Get("ETag")
+	if originalEtag != "" && !strings.HasPrefix(originalEtag, "W/") {
+		w.Header().Set("ETag", "W/"+originalEtag)
+	}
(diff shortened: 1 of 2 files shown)


caddyhttp/gzip/gzip.go
```diff
@@ -105,6 +105,10 @@ func (w *gzipResponseWriter) WriteHeader(code int) {
 	w.Header().Del("Content-Length")
 	w.Header().Set("Content-Encoding", "gzip")
 	w.Header().Add("Vary", "Accept-Encoding")
+	originalEtag := w.Header().Get("ETag")
+	if originalEtag != "" {
+		w.Header().Set("ETag", "W/"+originalEtag)
```
mholt: We should not prepend "W/" if it is already a weak ETag; we will have to perform this check here. It should be pretty straightforward to add a test for that case too.

spacewander:
@mholt 
Just updated and rebased my branch. Thanks for your review!

mholt:
Great job, thank you so much!! Really appreciate your contribution.

PR #1757, opened 2017-07-09
Rename directive requestid to request_id

tobya:
Following #1711 I caused confusion by using requestid and request_id.  I have changed all instances to request_id including the name of the folder.

Caddyfile now is like this

    request_id
    header /  "X-Request-Id" {request_id}

This is a breaking change for next release.

I will update documentation also https://github.com/caddyserver/website/pull/22


elcore:
> Following #1711 I caused confusion by using requestid and request_id. I have changed all instances to request_id including the name of the folder

It's rather my fault, because I changed it to `requestid` (see below)  

__

Matt H. comment on your changes and I implemented it :) 

> Use a context key of "request_id" instead.

> Move this way higher in the chain (like, after "shutdown"), and I'm thinking about whether this should be renamed to "requestid".

tobya:
It seems to work with the folder named requestid so unless there is a good reason to change it I think I would like to keep it consistent.

elcore:
> It seems to work with the folder named requestid so unless there is a good reason to change it I think I would like to keep it consistent.

https://golang.org/doc/effective_go.html#names

"By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps."

tobya:
ok, I'll update it.

elcore:
I would like to know Matt's opinion

mholt:
Let's keep the folder name `requestid` to match the package name, which should not have underscores in it. Thanks for checking!

tobya:
OK.

I think that's it, I have changed the code to use `request_id` in both places.  Thanks to everyone for patience. Documentation will be done once I make changes suggested by matt.  https://github.com/caddyserver/website/pull/22

PR #1767, opened 2017-07-16
Adding nobots directive

Xumeiquer:
It adds nobots directive used in nobots plugin

There is no issue

There is no changes

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -453,7 +453,6 @@ var directives = []string{
 	// directives that add middleware to the stack
 	"locale", // github.com/simia-tech/caddy-locale
 	"log",
-	"nobots",    // github.com/Xumeiquer/nobots
 	"cache", // github.com/nicolasazrak/caddy-cache
 	"rewrite",
 	"ext",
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/plugin.go
```diff
@@ -498,6 +498,7 @@ var directives = []string{
 	"grpc",      // github.com/pieterlouw/caddy-grpc
 	"gopkg",     // github.com/zikes/gopkg
 	"restic",    // github.com/restic/caddy
+	"nobots",    // github.com/Xumeiquer/nobots
```
mholt: Are you sure the very end of the middleware chain is the proper place for your directive? Have you tried it in conjunction with `proxy` or `templates` or `browse`, for example?

Xumeiquer: No, I believe a proper place should be at the top. Actually behind log. I am going to modify my commit.

caddyhttp/httpserver/plugin.go
```diff
@@ -453,6 +453,7 @@ var directives = []string{
 	// directives that add middleware to the stack
 	"locale", // github.com/simia-tech/caddy-locale
 	"log",
+	"nobots",    // github.com/Xumeiquer/nobots
```
mholt: You'll probably want to be after `rewrite` though, since rewrite could change the URI that your middleware reads. So consider putting this after `status` or `cors` instead. Also be sure to use `go fmt` so the CI tests pass!

mholt:
The new version is out, so you may publish your plugin!

Issue #1761, opened 2017-07-12
Log example in documentation has incorrect rotate_compress line


gregmac:
The documentation at https://caddyserver.com/docs/log shows the example:

````
log requests.log {
	rotate_size 50  # Rotate after 50 MB
	rotate_age  90  # Keep rotated files for 90 days
	rotate_keep 20  # Keep at most 20 log files
	rotate_compress # Compress rotated log files in gzip format
}
````

This example should actually have the last line as `rotate_compress gzip`

**Comment by tobya** [comment]
@gregmac  Thanks for the issue.  If you wanted we would be delighted with a PR to https://github.com/caddyserver/website on the log file
https://github.com/caddyserver/website/blob/master/site/docs/log.html

**Comment by mholt** [comment]
Hi @gregmac - are you sure? [It is a boolean value, not a string](https://sourcegraph.com/github.com/mholt/caddy@95366e41c464ce8e7d05d32e7df75e1b26906541/-/blob/caddyhttp/httpserver/roller.go#L76-77).

**Comment by gregmac** [comment]
@mholt With Caddy 0.10.4, when I put that line in as-is, I get the error:

    Parse error: Wrong argument count or unexpected line ending after 'rotate_compress'

and caddy exits with code 1.

**Comment by gregmac** [comment]
@mholt Ah, I think I see what's happening. It's treating that directive the same as all other logrotate directives -- assuming they have an integer-type value -- but explicitly ignores the integer conversion error: https://github.com/mholt/caddy/blob/bf7b25482ede58728589b5b0868b294db48cbcdf/caddyhttp/httpserver/roller.go#L65

So I'm pretty sure basically *any* value will cause it to work, as long as there is a value. That's actually kind of misleading, because, for example, the following all _enable_ compression:

    rotate_compress 1
    rotate_compress 0
    rotate_compress false
    rotate_compress off
    rotate_compress none

----

I just ran into this -- twice, sadly -- because I copied+pasted that example into my config. As I write this I'm only three days into using Caddy so right now relying heavily on documentation and examples.

**Comment by mholt** [comment]
Ah, this is a bug then! Must have been missed by the tests. I wonder if @fern4lvarez would have a chance to help us fix it.

**Comment by fern4lvarez** [comment]
Not sure if there was a follow-up of https://github.com/mholt/caddy/pull/1731, but the implementation states that if you want to compress logs, you just specify

```
rotate_compress
```

[...]

PR #1775, opened 2017-07-20
log,error: fix roller parser issue

tw4452852:

Revise log and error roller parser.

1) Each subdirective must be on its own line.
2) `rotate_compress` subdirective doesn't accept any parameter.

#1761 

N/A.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/roller.go
+++ b/caddyhttp/httpserver/roller.go
@@ -1,6 +1,7 @@
 package httpserver
 
 import (
+	"errors"
 	"io"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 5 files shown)


caddyhttp/httpserver/roller.go
```diff
@@ -54,17 +55,32 @@ func IsLogRollerSubdirective(subdir string) bool {
...
 	}
-	var value int
-	var err error
-	value, err = strconv.Atoi(where)
-	if what != directiveRotateCompress && err != nil {
-		return err
+
+	// rotate_compress doesn't accept any parameters.
+	// others only accept one parameter
+	if (what == directiveRotateCompress && len(where) != 0) ||
```
gregmac: If anyone was using `rotate_compress` and put something after it to get it to work (like I did) this is a breaking change. Probably worth noting in release notes.

PR #1780, opened 2017-07-25
Encapsulate WriteSiteNotFound error

sergeyfrolov:
Encapsulates writing of error response when no served vhosts match the requested host.
If HTTP version is 2.0 or above, 421 status code is written, otherwise - 404. Error now is a bit more descriptive.

N/A

N/A

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -366,7 +366,7 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 		if err != nil {
 			remoteHost = r.RemoteAddr
 		}
-		WriteTextResponse(w, http.StatusNotFound, "No such site at "+s.Server.Addr)
+		WriteSiteNotFound(w, r) // don't add headers outside of this function
 		log.Printf("[INFO] %s - No such site at %s (Remote: %s, Referer: %s)",
 			hostname, s.Server.Addr, remoteHost, r.Header.Get("Referer"))
 		return 0, nil
@@ -488,6 +488,19 @@ var ErrMaxBytesExceeded = errors.New("http: request body too large")
 // of the specified HTTP status code.
 func DefaultErrorFunc(w http.ResponseWriter, r *http.Request, status int) {
 	WriteTextResponse(w, status, fmt.Sprintf("%d %s\n", status, http.StatusText(status)))
+}
+
+const httpStatusMisdirectedRequest = 421 // RFC 7540, 9.1.2
+
+// WriteSiteNotFound writes appropriate error code to w, signaling that
+// requested host is not served by Caddy on a given port.
+func WriteSiteNotFound(w http.ResponseWriter, r *http.Request) {
+	status := http.StatusNotFound
+	if r.ProtoMajor >= 2 {
+		// TODO: use http.StatusMisdirectedRequest when it gets defined
+		status = httpStatusMisdirectedRequest
+	}
+	WriteTextResponse(w, status, fmt.Sprintf("%d Site %s is not served on this interface\n", status, r.Host))
 }
 
 // WriteTextResponse writes body with code status to w. The body will


caddyhttp/httpserver/server.go
```diff
@@ -490,6 +490,20 @@ func DefaultErrorFunc(w http.ResponseWriter, r *http.Request, status int) {
 	WriteTextResponse(w, status, fmt.Sprintf("%d %s\n", status, http.StatusText(status)))
 }
 
+const errStatusMisdirectedRequest = 421 // RFC 7540, 9.1.2
+// WriteSiteNotFound writes appropriate error code to w, signaling that
```
mholt: For readability, put an empty line before this comment and after the `const`.

caddyhttp/httpserver/server.go
```diff
@@ -490,6 +490,20 @@ func DefaultErrorFunc(w http.ResponseWriter, r *http.Request, status int) {
...
+// requested host is not served by Caddy on a given port.
+func WriteSiteNotFound(w http.ResponseWriter, r *http.Request) {
+	var status int
+	if r.ProtoMajor >= 2 {
+		// TODO: use http.StatusMisdirectedRequest when it gets defined
+		status = errStatusMisdirectedRequest
+	} else {
+		status = http.StatusNotFound
+	}
```
mholt: Can shorten this whole section to:

```go
status := http.StatusNotFound
if r.ProtoMajor >= 2 {
	// TODO: use http.StatusMisdirectedRequest when it gets defined
	status = errStatusMisdirectedRequest
}
```

caddyhttp/httpserver/server.go
```diff
@@ -490,6 +490,20 @@ func DefaultErrorFunc(w http.ResponseWriter, r *http.Request, status int) {
 	WriteTextResponse(w, status, fmt.Sprintf("%d %s\n", status, http.StatusText(status)))
 }
 
+const errStatusMisdirectedRequest = 421 // RFC 7540, 9.1.2
```
mholt: This isn't technically an `error` value, so maybe renaming this const to `httpStatusMisdirectedRequest` would be more fitting.

caddyhttp/httpserver/server.go
```diff
@@ -490,6 +490,20 @@ func DefaultErrorFunc(w http.ResponseWriter, r *http.Request, status int) {
...
+// requested host is not served by Caddy on a given port.
+func WriteSiteNotFound(w http.ResponseWriter, r *http.Request) {
+	var status int
+	if r.ProtoMajor >= 2 {
+		// TODO: use http.StatusMisdirectedRequest when it gets defined
+		status = errStatusMisdirectedRequest
+	} else {
+		status = http.StatusNotFound
+	}
+	WriteTextResponse(w, status, fmt.Sprintf("%d No site %s served at this port\n", status, r.Host))
```
mholt: For clarity, rephrase to `Site %s is not served on this interface` - if that sounds good to you?

caddyhttp/httpserver/server.go
sergeyfrolov: Agreed.

caddyhttp/httpserver/server.go
sergeyfrolov: That sounds better, assuming it is fine to call this an interface

francislavoie:
The RFC mentions:
>This status code MUST NOT be generated by proxies.
I haven't really looked at the code flow, but this seems to be true, right? This code doesn't get hit if proxying?

mholt:
@francislavoie Yeah, if we're proxying, the user would receive a proxied site, rather than this error message.

mholt:
Welcome to the project, @sergeyfrolov - thanks for your first contribution!

sergeyfrolov:
@mholt thank you! Glad to be part of the project!

Issue #999, opened 2016-08-05
Feature Request: Proxy into QUIC


msample:
For request dispatch to microservices it is often useful to have a local reverse proxy acting as a smart, load balancing, connection-reusing client.  This improves actual client language support by working around poor http clients as they are used only for local connections and keep-alive is less important.  Clients can even skip TLS on local hop to the egress proxy in environments where policy allows it.

Given the above, it would be great if Caddy in proxy mode could convert regular HTTP, HTTPS and HTTP2 client connections over TCP to use QUIC as their transport to the backend.

mholt:
Hi, thanks for your feature request!

We'd first need a pure Go QUIC client implementation, which there isn't one yet (quic-go may eventually have one). Also I'm not sure I can make this a priority in the near future. As such, I'm going to close this since there's nothing really to do on it for the time being. But thank you for your feedback!

mholt:
Note: https://twitter.com/progrium/status/771021678524506112

> This is the Go impl of QUIC that Caddy uses. But, it's server only. Can I sponsor somebody to contribute a client?

Skilgarriff:
Appears that the client is now done: https://github.com/lucas-clemente/quic-go/pull/399 -> I'd love to get on this soon as a quic proxy would be a huge benefit for me. Any pointers on how I would go about doing this?

mholt:
I have no idea... :confused: probably start in reverseproxy.go which is where Caddy initiates connections to backends.

Skilgarriff:
Sounds good, I'll check it out and see if I can't figure out a plan.

heri16:
anyone working on this?

francislavoie:
@heri16 not that I know of, go for it! :+1:

heri16:
Is there a certain style-guide I should follow when CONTRIBUTING?

The plan atttempt replacing http.Transport with h2quic.QuicRoundTripper .
https://github.com/lucas-clemente/quic-go/blob/master/example/client/main.go

francislavoie:
Just run gofmt and golint before you commit. Anything else can be caught in code review in your PR before it gets merged.

See here if you haven't already https://github.com/mholt/caddy/blob/master/.github/CONTRIBUTING.md 😄

slightfoot:
If I understand the request correctly the OP wants to connect to the backend service with QUIC. This sounds like a job for the additional protocol support which lets you register a round-tripper. Suggest this is continued as part of #1638.

[...]

PR #1782, opened 2017-07-26
Proxy can now use QUIC for upstream connections

twdkeule:
Adds QUIC proxy support for upstream connections.

#999 

Yes, updating the `proxy` command in Caddyfile to include a QUIC scheme: `quic://`. This is identical to `https://` but using QUIC. 
Using this feature does not require the proxy to use the `-quic` flag, but the server still has to!

### 4. Checklist

I will write tests at a later point (September maybe).
Note: when updating go-quic, h2quic.QuicRoundTripper should change to
h2quic.RoundTripper


twdkeule:
Fyi: this does not work (yet) when the upstream server pushes something to the proxy.

mholt:
Thanks for your work on this, @twdkeule ! Also on the upstream quic-go library, we all really appreciate it. I'll be looking at this soon!

twdkeule:
Just a quick update that receiving server pushes will probably not be supported in the near future as even the HTTP/2 implementation of the Go standard lib does not support it yet.

caddyhttp/proxy/upstream.go
```diff
@@ -150,7 +150,8 @@ func (u *staticUpstream) From() string {
 
 func (u *staticUpstream) NewHost(host string) (*UpstreamHost, error) {
 	if !strings.HasPrefix(host, "http") &&
-		!strings.HasPrefix(host, "unix:") {
+		!strings.HasPrefix(host, "unix:") &&
+		!strings.HasPrefix(host, "quic:") {
```
mholt: Is it really necessary to have the user specify this, or can the reverse proxy infer QUIC from the upstream's Alt-Svc headers?

mholt:
Tests _would_ prove really helpful here, especially as I prepare to rewrite the proxy middleware; I would hate to see my rewrite break this change. Do you want someone to help you write the test for it?

caddyhttp/proxy/upstream.go
twdkeule: It can.
But one of the features of QUIC is the 1- or 0-RTT connection setup and you're losing that if you set up a connection over TCP first. So I would really keep it there. 
Besides, I don't know if Caddy supports the Alt-Svc headers as a client?

This is also what Google does inside Chrome for it's own QUIC-enabled servers, directly connecting over QUIC.

twdkeule:
I really don't have the time to write tests until the end of August. But anybody is free to write them :wink:

caddyhttp/proxy/upstream.go
mholt: Okay, so just to be sure I'm clear, using `quic:` in the Caddyfile this way means that the upstream server _has to use QUIC each and every time_. That's acceptable, I just want to be sure that's clear for the documentation. It does not only enable QUIC to upstream, it _forces_ it.

caddyhttp/proxy/reverseproxy.go
```diff
@@ -185,6 +186,8 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 		rp.Transport = &http.Transport{
 			Dial: socketDial(target.String()),
 		}
+	} else if target.Scheme == "quic" {
+		rp.Transport = &h2quic.QuicRoundTripper{}
```
mholt: What about all the stuff the director func above does? Does the QUIC transport not need/use it?

twdkeule: Do you mean the `rp.Transport = &http.Transport{Dial: socketDial(target.String())}` ?
Or 
```
// if keepalive is equal to the default,
// just use default transport, to avoid creating
// a brand new transport
transport := &http.Transport{
	Proxy:                 http.ProxyFromEnvironment,
	Dial:                  defaultDialer.Dial,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
if keepalive == 0 {
	transport.DisableKeepAlives = true
} else {
	transport.MaxIdleConnsPerHost = keepalive
}
if httpserver.HTTP2 {
	http2.ConfigureTransport(transport)
}
rp.Transport = transport
```?
I think those options are not part of `h2quic.QuicRoundTripper`, so they cannot be set. But I'll take another look when I'll write those tests.

Comment (mholt) [inline caddyhttp/proxy/reverseproxy.go]: Yeah, I meant that large block of logic. :) Okay, if they're not even available with the QuicRoundTripper then I wouldn't worry about it for now.

**Comment by twdkeule** [comment]
I added a test to check if a QUIC RoundTripper was created. 
Does the repo have an integration test suite? Because it would be good to run a test where 2 Caddy servers use QUIC to communicate.

**Comment by mholt** [comment]
@twdkeule Thanks for working on it! I don't have a suite of integration tests set up quite yet, but it's not that hard to add one already; we do this for other kinds of proxying if you look at the other Test functions. Just spin up a server on some port (that's configured to use QUIC) and then proxy a request to it through ServeHTTP. You don't have to spin up a whole Caddy instance, per-se, just a QUIC server.

Checking for the RoundTripper is a good start. Would you want to try the other test that actually makes sure proxying QUIC works? ;) (I would highly, strongly recommend it!)

By the way:

> Note: when updating go-quic, h2quic.QuicRoundTripper should change to h2quic.RoundTripper

What do you mean by this? Should we just update go-quic in our vendor folder with this PR?

**Comment by twdkeule** [comment]
Yes, I'll definitely add that test. Maybe next week.

You can ignore that comment. You already updated quic-go sufficiently I'm the last vendor update :-)

**Comment by mholt** [comment]
Gotcha. Cool. An "integration" test should just be a few lines of code, I think. Looks like this is all you need to set up the QUIC server (taken from quic-go's readme):

```
h2quic.ListenAndServeQUIC("localhost:4242", "/path/to/cert/chain.pem", "/path/to/privkey.pem", nil)
```

**Comment by twdkeule** [comment]
I vendored in the certificate from quic-go to run the quic server

**Comment by mholt** [comment]
@twdkeule Cool, almost there! The tests are failing because "cannot bind requested address" - likely because a test server is not being shut down or we are re-using the same address? Do the tests pass locally for you?

Anyway, if I understand this properly, once the tests pass, we can merge this PR and deploy this in 0.10.7. Is that right?

**Comment by twdkeule** [comment]
The tests run successful locally.
I don't stop the goroutine running the QUIC server, but that should stop after the test.go file is finished, no?
I changed the port number just in case.

**Comment by twdkeule** [comment]
OK, the tests are a succes.
@mholt this is ready to merge

**Comment by mholt** [comment]
Thanks @twdkeule for tackling the tests. Sorry about the quic-go tests having troubles. This is a recent thing that changed upstream. I'd rather not add test dependencies to the vendor folder if we can avoid it. Do you have any idea what caused the problem? We're discussing it in https://github.com/lucas-clemente/quic-go/issues/797 in case it can be fixed there, rather than adding test baggage to the vendor folder.

**Comment by mholt** [comment]
Okay, updating the CI to Go 1.9 fixed the issue. @twdkeule if you could undo your changes to the vendored folders and only include the changes for your PR, then we can merge this. If it happens early enough today, it's likely to go out on the release I'm planning.

**Comment by twdkeule** [comment]
@mholt The Travis tests are succesful now, but Appveyor cannot find the file.
I don't have a Windows machine for Go development, so I'm not sure what the problem is. 
Go should handle `/` correctly as path separator, right?

File: caddyhttp/proxy/proxy_test.go [hunk 3]
```diff
@@ -1470,3 +1472,57 @@ func TestChunkedWebSocketReverseProxy(t *testing.T) {
...
+	}
+
+	// start QUIC server
+	go func() {
+		dir, err := os.Getwd()
+		if err != nil {
+			t.Errorf("Expected no error. Got: %s", err.Error())
+			return
+		}
+		certDir := path.Join(dir, "../../vendor/github.com/lucas-clemente/quic-go/example/")
```
mholt: Aha -- this won't work on Windows. You need to use `filepath.Join(...)` where each argument is one element of the path. Go will use the proper seperator for the OS.

Hrm, not really sure I like this hard-coded path. Is it possible to copy just the certificate directory into a local testdata folder?

Xe: The windows API's take both forward and backslashes

mholt:
Oh. Then I was wrong. 😅 Maybe moving the certs I to a local folder is
still a good way to do it...

caddyhttp/proxy/proxy_test.go
mholt: Oh. Then I was wrong. 😅 Maybe moving the certs I to a local folder is still a good idea though.

mholt:
@twdkeule The vendor folder does not include test dependencies (this is intentional).

twdkeule:
Maybe it's easier to use a public website we know uses QUIC

mholt:
@twdkeule Maybe, but then we'd have external network dependencies for the tests. I bet we can just move the certificates into a local test folder for proxy and get them to work. :)

mholt:
Wow, this is looking better. Thank you for the changes! I will do another review this week!

mholt:
Sorry, I've gotten a little delayed...

heri16:
Looking forward to the release. Awesome.

mholt:
@twdkeule I just realized after running the tests locally that the certificate chain is a real one (I didn't bother to decode it in the code review, oops) that expires on October 17, and that the hosts file must direct `quic.clemente.io` to localhost. For now, I've disabled that test outside of CI environments, and later, we'll have to update it every 3 months or just disable it entirely after October 17. Is there a more permanent or low-maintenance way of testing connections to QUIC upstreams? (pinging @marten-seemann in case you have any ideas as well!)

twdkeule:
As far as I am aware, the QUIC server needs a certificate to start. 
We could use a self-signed certificate with a 5 year validity, but also only on a  CI environment as we need to add it to the trusted certificates. Unless `insecure-skip-verify` works with quic-go

marten-seemann:
`tls.Config.InsecureSkipVerify` works with QUIC.
Alternatively, you could generate a CA certificate, sign the server public key with that and use it in the `tls.Config.ClientCAs` cert pool.

mholt:
Thanks, I fixed the tests easily enough in cccfe3b4efb45b443dcbef9ab26a4ec4ffb46eab by your suggestion.

PR #1784, opened 2017-07-27
Move fallbackHosts to vhostTrie

sergeyfrolov:
 * Moves fallbackHosts to vhostTrie. Therefore, this list is no longer global, rather is per-trie(and, as a result, is per-listener).
 * SiteConfig now has an option to make it a fallback site for all unmatched requests.

None

None

### 4. Checklist


sergeyfrolov:
Thank you!

Issue #1794, opened 2017-07-29
rewrite if match {file} doesn't work like expected


onodera-punpun:
### 1. What version of Caddy are you using (`caddy -version`)?
`Caddy 0.10.6 (+e7f08bf Sat Jul 29 13:43:27 UTC 2017)`

### 2. What are you trying to do?
Trying to rewrite requests like `/Sn3d.jpg` to `/uploads/Sn3d.jpg` using `if {file} match...`.
So files that are exactly 4 randomized characters long, with an optional extension.

The following works:
```
rewrite {
    regexp ^/[[:alnum:]]{4}(?:\.[[:alnum:]]+)?/?$
    to /uploads/{file}
}
```

but using the a `{file}` or even `{path}` match equivalent doesn't work:
```
rewrite {
    if {file} match ^[[:alnum:]]{4}(?:\.[[:alnum:]]+)?$
    to /uploads/{file}
}
```

It seems that the `$` somehow causes this? It might also be the case that I'm a complete noob at regexp, hehe.

### 6. What did you expect to see?
A rewrited request

### 7. What did you see instead (give full error messages and/or log)?
A 404.

mholt:
@onodera-punpun I'm seeing this too, but am unable to proceed because frankly I'm stuck. 

@abiosoft: this is strange, have you seen this before? I'm using this rewrite given above:

```
rewrite {
    if {file} match ^[[:alnum:]]{4}(?:\.[[:alnum:]]+)?$
    to /uploads/{file}
}
```

And when I `fmt.Println(rules)` in setup.go, just before `return rules, nil` [here](https://github.com/mholt/caddy/blob/e7f08bff38988c3049b7fda301c52a681af63cd8/caddyhttp/rewrite/setup.go#L105), this is the output:

```
[%!v(PANIC=runtime error: invalid memory address or nil pointer dereference)]
```

Indeed, printing out `rules[0]` shows the panic too. However, if you print rules[0] using `fmt.Printf("%#v\n", rules[0])`, you get a NON-NIL value:

```
[...]

PR #1796, opened 2017-07-30
Fix for #1794: Fixes issues with IfMatcher and regular expressions.

slightfoot:
* Fixes bug caused by Replacer running on the regular expressions in IfMatcher. We also now compile regular expressions up front to detect errors.

* Fixes rewrite bugs that come from formatting a rule as a string and failing with nil dereference caused by embedding Regexp pointer in a Rule. Re: Issue #1794

#1794 

It might be good to update the docs to let users know that they cannot use replacement vars in regular expressions.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/condition.go
+++ b/caddyhttp/httpserver/condition.go
@@ -53,59 +53,8 @@ const (
 	matchOp      = "match"
 )
 
-func operatorError(operator string) error {
-	return fmt.Errorf("Invalid operator %v", operator)
-}
-
 // ifCondition is a 'if' condition.
-type ifCondition func(string, string) bool
-
-var ifConditions = map[string]ifCondition{
-	isOp:         isFunc,
-	notOp:        notFunc,
-	hasOp:        hasFunc,
-	startsWithOp: startsWithFunc,
-	endsWithOp:   endsWithFunc,
-	matchOp:      matchFunc,
-}
-
-// isFunc is condition for Is operator.
-// It checks for equality.
-func isFunc(a, b string) bool {
-	return a == b
-}
-
-// notFunc is condition for Not operator.
-// It checks for inequality.
-func notFunc(a, b string) bool {
-	return !isFunc(a, b)
-}
-
-// hasFunc is condition for Has operator.
-// It checks if b is a substring of a.
-func hasFunc(a, b string) bool {
-	return strings.Contains(a, b)
-}
-
-// startsWithFunc is condition for StartsWith operator.
-// It checks if b is a prefix of a.
-func startsWithFunc(a, b string) bool {
-	return strings.HasPrefix(a, b)
-}
-
-// endsWithFunc is condition for EndsWith operator.
-// It checks if b is a suffix of a.
-func endsWithFunc(a, b string) bool {
-	return strings.HasSuffix(a, b)
-}
-
-// matchFunc is condition for Match operator.
-// It does regexp matching of a against pattern in b
-// and returns if they match.
-func matchFunc(a, b string) bool {
-	matched, _ := regexp.MatchString(b, a)
-	return matched
-}
+type ifFunc func(a, b string) bool
 
 // ifCond is statement for a IfMatcher condition.
 type ifCond struct {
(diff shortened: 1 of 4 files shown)


caddyhttp/rewrite/setup_test.go
```diff
@@ -175,13 +175,25 @@ func TestRewriteParse(t *testing.T) {
 			}
 
 			if actualRule.Regexp != nil {
-				if actualRule.String() != expectedRule.String() {
+				actualRuleStr := fmt.Sprintf("%v", actualRule)
+				expectedRuleStr := fmt.Sprintf("%v", actualRule)
```
slightfoot: This should be `fmt.Sprintf("%v", expectedRule)`

slightfoot:
Fixed the bug I saw in my review.

caddyhttp/httpserver/condition.go
```diff
@@ -123,7 +119,12 @@ func newIfCond(a, operator, b string) (ifCond, error) {
 		operator = operator[4:]
 	}
 	if _, ok := ifConditions[operator]; !ok {
-		return ifCond{}, operatorError(operator)
+		return ifCond{}, fmt.Errorf("Invalid operator %v", operator)
+	}
+	if operator == matchOp {
+		if _, err := regexp.Compile(b); err != nil {
```
mholt: Should we save the compiled regular expression so it doesn't need to be recompiled for every request? /cc @abiosoft

caddyhttp/rewrite/setup.go
```diff
@@ -90,10 +90,11 @@ func rewriteParse(c *caddy.Controller) ([]httpserver.HandlerConfig, error) {
 			if to == "" {
 				return nil, c.ArgErr()
 			}
-			if rule, err = NewComplexRule(base, pattern, to, ext, matcher); err != nil {
+			cRule, err := NewComplexRule(base, pattern, to, ext, matcher)
+			if err != nil {
 				return nil, err
 			}
-			rules = append(rules, rule)
+			rules = append(rules, *cRule)
```
mholt: Is there a reason that NewComplexRule is still returning a pointer, if it's just as good without?

caddyhttp/rewrite/setup_test.go
```diff
@@ -175,13 +175,23 @@ func TestRewriteParse(t *testing.T) {
...
+		testRewriteParseString(t, actual, i)
 	}
+}
 
+func testRewriteParseString(t *testing.T, rules []httpserver.HandlerConfig, testIndex int) {
+	defer func() {
+		if r := recover(); r != nil {
+			t.Errorf("Test %d: Failed to string encode: %#v", testIndex, r)
+		}
+	}()
```
mholt: Does this test actually work? I was under the impression that `fmt` kind of recovered from panics by encoding the panic into a string.

caddyhttp/httpserver/condition.go
slightfoot: I tried doing that, ended up being quite painful. But I think I fixed the issue that put a stopper on it. I'll review my old stash and go from there.

caddyhttp/rewrite/setup.go
slightfoot: Good call. Will change.

caddyhttp/httpserver/condition.go
abiosoft: Yeah, that is better.`rewrite` does that.

slightfoot:
All changes implemented.

caddyhttp/httpserver/condition.go
```diff
@@ -53,100 +53,88 @@ const (
...
+	if strings.HasPrefix(op, "not_") {
+		i.neg = true
+		i.op = op[4:]
 	}
-	if _, ok := ifConditions[operator]; !ok {
-		return ifCond{}, operatorError(operator)
+
+	switch i.op {
+	case isOp:
+		// It checks for equality.
```
mholt: "It" => "is" ?

They all say "It" though so I dunno if that was intentional.

slightfoot: They are the same comments as the previous methods that the switch replaces.

Issue #1717, opened 2017-06-15
Push more than one file (multiple links per Link header)


si14:
I'm sorry for not following the template, but I'm not sure if it makes sense in this case.

I use Caddy to push resources from the backend. To do this conditionally, I use Link header and a little bit of logic in the backend. However, if I understand Caddy's code correctly, it's currently implied that one Link header = one resource. Unfortunately, a lot of backends (e.g. Django) don't really support multiple headers with the same name, limiting http2 push to one resource.

Cloudflare had exactly this problem and solved it by allowing multiple links per Link header: https://blog.cloudflare.com/http-2-server-push-with-multiple-assets-per-link-header/ Do you think it's reasonable to do the same in Caddy?

xfalcox:
Oh just got this problem too, looks like in Rails it's the same.

wendigo:
Here you go: https://github.com/mholt/caddy/pull/1798/files

PR #1798, opened 2017-07-31
Allow pushing multiple resources via Link header

wendigo:
Addresses #1717

### DIFF
--- a/caddyhttp/push/handler.go
+++ b/caddyhttp/push/handler.go
@@ -58,24 +58,39 @@ outer:
 }
 
 func (h Middleware) servePreloadLinks(pusher http.Pusher, headers http.Header, links []string) {
+outer:
 	for _, link := range links {
-		parts := strings.Split(link, ";")
+		resources := strings.Split(link, ",")
 
-		if link == "" || strings.HasSuffix(link, "nopush") {
-			continue
-		}
+		for _, resource := range resources {
+			parts := strings.Split(resource, ";")
 
-		target := strings.TrimSuffix(strings.TrimPrefix(parts[0], "<"), ">")
+			if link == "" || strings.HasSuffix(resource, "nopush") {
+				continue
+			}
 
-		err := pusher.Push(target, &http.PushOptions{
-			Method: http.MethodGet,
-			Header: headers,
-		})
+			target := strings.TrimSuffix(strings.TrimPrefix(parts[0], "<"), ">")
 
-		if err != nil {
-			break
+			if h.IsRemoteResource(target) {
+				continue
+			}
+
+			err := pusher.Push(target, &http.PushOptions{
+				Method: http.MethodGet,
+				Header: headers,
+			})
+
+			if err != nil {
+				break outer
+			}
 		}
 	}
+}
+
+func (h Middleware) IsRemoteResource(resource string) bool {
+	return strings.HasPrefix(resource, "//") ||
+		strings.HasPrefix(resource, "http://") ||
+		strings.HasPrefix(resource, "https://")
 }
 
 func (h Middleware) mergeHeaders(l, r http.Header) http.Header {
--- a/caddyhttp/push/handler_test.go
+++ b/caddyhttp/push/handler_test.go
@@ -269,6 +269,52 @@ func TestMiddlewareShouldInterceptLinkHeader(t *testing.T) {
 	comparePushedResources(t, expectedPushedResources, pushingWriter.pushed)
 }
 
+func TestMiddlewareShouldInterceptLinkHeaderWithMultipleResources(t *testing.T) {
+	// given
+	request, err := http.NewRequest(http.MethodGet, "/index.html", nil)
+	writer := httptest.NewRecorder()
+
+	if err != nil {
+		t.Fatalf("Could not create HTTP request: %v", err)
+	}
+
+	middleware := Middleware{
+		Next: httpserver.HandlerFunc(func(w http.ResponseWriter, r *http.Request) (int, error) {
+			w.Header().Add("Link", "</assets/css/screen.css?v=5fc240c512>; rel=preload; as=style,</content/images/2016/06/Timeouts-001.png>; rel=preload; as=image,</content/images/2016/06/Timeouts-002.png>; rel=preload; as=image")
+			w.Header().Add("Link", "<//cdn.bizible.com/scripts/bizible.js>; rel=preload; as=script")
+			return 0, nil
+		}),
+		Rules: []Rule{},
+	}
+
+	pushingWriter := &MockedPusher{ResponseWriter: writer}
+
+	// when
+	_, err2 := middleware.ServeHTTP(pushingWriter, request)
+
+	// then
+	if err2 != nil {
+		t.Error("Should not return error")
+	}
+
+	expectedPushedResources := map[string]*http.PushOptions{
+		"/assets/css/screen.css?v=5fc240c512": {
+			Method: http.MethodGet,
+			Header: http.Header{},
+		},
+		"/content/images/2016/06/Timeouts-001.png": {
+			Method: http.MethodGet,
+			Header: http.Header{},
+		},
+		"/content/images/2016/06/Timeouts-002.png": {
+			Method: http.MethodGet,
+			Header: http.Header{},
+		},
+	}
+
+	comparePushedResources(t, expectedPushedResources, pushingWriter.pushed)
+}
+
 func TestMiddlewareShouldInterceptLinkHeaderPusherError(t *testing.T) {
 	// given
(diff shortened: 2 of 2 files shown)


caddyhttp/push/handler.go
```diff
@@ -58,26 +58,41 @@ outer:
...
+		resources := strings.Split(link, ",")
 
-		if link == "" || strings.HasSuffix(link, "nopush") {
-			continue
-		}
+		for _, resource := range resources {
+			parts := strings.Split(resource, ";")
 
-		target := strings.TrimSuffix(strings.TrimPrefix(parts[0], "<"), ">")
+			if link == "" || strings.HasSuffix(resource, "nopush") {
```
twdkeule: Shouldn't `nopush` be checked for every item in `parts`?
Maybe add a `nopush` link to the unit test.

wendigo: No, it's fine as it is. I've added test case to confirm this :)

caddyhttp/push/handler.go
```diff
@@ -58,26 +58,41 @@ outer:
 }
 
 func (h Middleware) servePreloadLinks(pusher http.Pusher, headers http.Header, links []string) {
```
mholt: Can you add a godoc-style comment for this function that describes the syntax of the Link headers that it parses and applies?

caddyhttp/push/handler.go
```diff
@@ -58,26 +58,41 @@ outer:
...
+			})
+
+			if err != nil {
+				break outer
+			}
 		}
 	}
 }
 
+func (h Middleware) IsRemoteResource(resource string) bool {
```
mholt: Does this have to be exported? If it's not used outside the package, let's un-export it. Add a godoc comment either way.

caddyhttp/push/handler_test.go
```diff
@@ -269,6 +269,52 @@ func TestMiddlewareShouldInterceptLinkHeader(t *testing.T) {
...
+			w.Header().Add("Link", "</assets/css/screen.css?v=5fc240c512>; rel=preload; as=style,</content/images/2016/06/Timeouts-001.png>; rel=preload; as=image,</content/images/2016/06/Timeouts-002.png>; rel=preload; as=image")
+			w.Header().Add("Link", "<//cdn.bizible.com/scripts/bizible.js>; rel=preload; as=script,</resource.png>; rel=preload; as=script; nopush")
+			return 0, nil
+		}),
+		Rules: []Rule{},
+	}
+
+	pushingWriter := &MockedPusher{ResponseWriter: writer}
+
+	// when
```
mholt: when what?

caddyhttp/push/handler_test.go
```diff
@@ -269,6 +269,52 @@ func TestMiddlewareShouldInterceptLinkHeader(t *testing.T) {
...
+	_, err2 := middleware.ServeHTTP(pushingWriter, request)
+
+	// then
```
mholt: then what?

Sorry, just a little confused by these

caddyhttp/push/handler.go
wendigo: Sure

caddyhttp/push/handler_test.go
wendigo: This is a way I'm separating test parts from each other:

// given

// when

// then

See: https://martinfowler.com/bliki/GivenWhenThen.html

wendigo:
Take a look at this now @mholt

caddyhttp/push/handler_test.go
mholt: Gotcha. Thanks

caddyhttp/push/link_parser.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+const (
+	commaSeparator     = ","
+	semicolonSeparator = ";"
+	equalSeparator     = "="
+)
+
+var (
+	resourceRegexp = regexp.MustCompile("^<(.*?)>;?(.*)$")
```
mholt: How well does this work if resource URIs have commas or semicolons in them?

Also -- this isn't a showstopper -- I wonder how this pre-compiled regex performs compared to strings.Split. I'll try to whip up a benchmark today and see, just to satisfy my curiosity.

wendigo: It's a valid question, I'll dig into: https://tools.ietf.org/html/rfc5988#section-5.1

caddyhttp/push/handler_test.go
wendigo: I guess it will be much more readable with colons:

// given:

// when:

// then:

caddyhttp/push/link_parser.go
mholt: Ah yeah, that's what the angle brackets are for then. Got it.

caddyhttp/push/link_parser.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, link := range strings.Split(header, commaSeparator) {
+		l := linkResource{params: make(map[string]string)}
+
+		groups := resourceRegexp.FindAllStringSubmatch(link, -1)
+
+		if len(groups) == 0 {
+			continue
+		}
+
+		l.uri = strings.TrimSpace(groups[0][1])
```
mholt: What if `len(groups[0]) <= 1`? This will panic...

caddyhttp/push/link_parser.go
```diff
@@ -0,0 +1,70 @@
...
+
+		for _, param := range strings.Split(strings.TrimSpace(groups[0][2]), semicolonSeparator) {
```
mholt: Similar potential problem here; possible out of bounds.

caddyhttp/push/link_parser.go
mholt: For what it's worth, I wrote a simple function (not thoroughly tested, but I did get it working with your test case above) that extracts the text between `< >` (but not the other properties; those can also be parsed easily enough). Here's how it performs compared to FindAllStringSubmatch:

```
BenchmarkLinkRegex-8     	  200000	      6334 ns/op
BenchmarkLinkStrings-8   	 5000000	       332 ns/op
```

Performance is not by any means a halter for merging something like this -- but the work is half done now. :smile: Do you think we should finish it up? Here is the test/bench file:

```go
package main

import (
	"regexp"
	"strings"
	"testing"
)

var resourceRegexp = regexp.MustCompile("^<(.*?)>;?(.*)$")

var link = `</assets/css/screen.css?v=5fc240c512>; rel=preload; as=style,</content/images/2016/06/Timeouts-001.png>; rel=preload; as=image,</content/images/2016/06/Timeouts-002.png>; rel=preload; as=image")`

func BenchmarkLinkRegex(b *testing.B) {
	for n := 0; n < b.N; n++ {
		resourceRegexp.FindAllStringSubmatch(link, -1)
	}
}

func TestLinkString(t *testing.T) {
	var items []string
	end := 0
	for {
		openCaret := strings.Index(link[end:], "<") + end + 1
		if openCaret < end+1 {
			break
		}
		endCaret := strings.Index(link[openCaret:], ">") + openCaret
		items = append(items, link[openCaret:endCaret])
		end = endCaret + 1
	}
}

func BenchmarkLinkStrings(b *testing.B) {
	for n := 0; n < b.N; n++ {
		linkString(link)
	}
}

// this is the function that parses the Link header (it deserves a better name, and
// should probably parse the other components than just the resource path)
func linkString(link string) []string {
	var items []string
	end := 0
	for {
		openCaret := strings.Index(link[end:], "<") + end + 1
		if openCaret < end+1 {
			break
		}
		endCaret := strings.Index(link[openCaret:], ">") + openCaret
		items = append(items, link[openCaret:endCaret])
		end = endCaret + 1
	}
	return items
}
```

mholt:
@wendigo I think we'd be able to get this in the next release with a few changes I mentioned. Did you have any questions? Let me know if you have no time, I might be able to take over, but I don't want to conflict.

wendigo:
I will handle this :)

mholt:
I'm going to try to do a new release today, if this PR doesn't make it in today; that's OK, I guess we can just get it in 0.10.8.

caddyhttp/push/link_parser.go
wendigo: I've got rid of regexp.

wendigo:
Sorry for long delay @mholt :) I've been busy building my new blazzing fast development pc :smiling_imp:

mholt:
@wendigo No problem! Fascinating, one of the test runs caught a data race in an unrelated test. I'll open an issue about it. What's the status of this PR now? Is it ready for another review?

wendigo:
Yes it is @mholt

mholt:
Thanks so much @wendigo ! You're a champ. Merged in and we're good to go here.

PR #1800, opened 2017-08-02
Distinguishable exit codes

mholt:

Makes exit codes different for different exit reasons.

https://github.com/abiosoft/caddy-docker/issues/65

Probably none at this point?

### 4. Checklist

New code explanations:

0: normal or expected exit
1: error before server finished starting
2: double SIGINT (force quit)
3: error stopping with SIGQUIT
4: shutdown callback(s) returned error(s)

### DIFF
--- a/sigtrap.go
+++ b/sigtrap.go
@@ -33,7 +33,7 @@ func trapSignalsCrossPlatform() {
 				if PidFile != "" {
 					os.Remove(PidFile)
 				}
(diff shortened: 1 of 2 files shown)


abiosoft:
This is helpful. Thanks.

francislavoie:
Should definitely have a mention in the release notes in case someone relies on exit code 1 for everything (although I doubt that)

PR #1814, opened 2017-08-08
SIGUSR2 triggers graceful binary upgrades (spawns new process)

### LINKED ISSUE #814: Replace graceful restart code for binary upgrades (SIGUSR2)

mholt:

Implements zero-downtime, graceful binary upgrades with SIGUSR2. This is the first step towards enabling automatic upgrades in production for the continued security and privacy of all sites and their connections served by Caddy. Next step in evil plan is to figure out a good, secure way to trigger rolling updates. (However, this PR is not the place to discuss those details -- try the [forum](https://caddy.community) or a new issue instead.)

I think this progress will interest @paragonie-scott as it's a step toward auto updates.

Closes #814

Update the CLI page to explain signal USR2.

### 4. Checklist

Edit: D'oh, I forgot about Windows, as usual... Update: fixed.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -17,6 +17,7 @@ package caddy
 
 import (
 	"bytes"
+	"encoding/gob"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -51,7 +52,7 @@ var (
 	// isUpgrade will be set to true if this process
 	// was started as part of an upgrade, where a parent
 	// Caddy process started this one.
-	isUpgrade bool
+	isUpgrade = os.Getenv("CADDY_UPGRADE") == "1"
 
 	// started will be set to true when the first
 	// instance is started; it never gets set to
@@ -360,6 +361,16 @@ type AfterStartup interface {
 // is returned. Consequently, this function never returns a nil
 // value as long as there are no errors.
 func LoadCaddyfile(serverType string) (Input, error) {
+	// If we are finishing an upgrade, we must obtain the Caddyfile
+	// from our parent process, regardless of configured loaders.
+	if IsUpgrade() {
+		err := gob.NewDecoder(os.Stdin).Decode(&loadedGob)
+		if err != nil {
+			return nil, err
+		}
+		return loadedGob.Caddyfile, nil
+	}
+
 	// Ask plugged-in loaders for a Caddyfile
 	cdyfile, err := loadCaddyfileInput(serverType)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


elcore:
👍I will take a look tomorrow

caddyhttp/httpserver/https.go
```diff
@@ -44,9 +44,14 @@ func activateHTTPS(cctx caddy.Context) error {
 	// renew all relevant certificates that need renewal. this is important
 	// to do right away so we guarantee that renewals aren't missed, and
 	// also the user can respond to any potential errors that occur.
-	err = caddytls.RenewManagedCertificates(true)
-	if err != nil {
-		return err
+	// (skip if upgrading, because the parent process is likely already listening
+	// on the ports we'd need to do ACME before we finish starting; parent process
+	// already running renewal ticker, so renewal won't be missed anyway.)
+	if !caddy.IsUpgrade() {
```
elcore: :+1:

upgrade.go
```diff
@@ -0,0 +1,212 @@
...
+// signalParentOnce is used to make sure that the parent is only
+// signaled once; doing so more than once breaks whatever socket is
+// at fd 4 (TODO: the reason for this is still unclear - to reproduce,
+// call Stop() and Start() in succession at least once after a
+// restart, then try loading first host of Caddyfile in the browser
+// - this was pre-v0.9; this code and godoc is borrowed from the
+// implementation then, but I'm not sure if it's been fixed yet, as
+// of v0.10.7). Do not use this directly; call signalSuccessToParent
+// instead.
+var signalParentOnce sync.Once
```
elcore: This isn't a showstopper, right?

caddy.go
```diff
@@ -51,7 +52,7 @@ var (
 	// isUpgrade will be set to true if this process
 	// was started as part of an upgrade, where a parent
 	// Caddy process started this one.
-	isUpgrade bool
+	isUpgrade = os.Getenv("CADDY_UPGRADE") == "1"
```
elcore: That is a really intelligent way to handle this situation! Well done Matt!!

upgrade.go
mholt: Nope, not at all.

elcore: Great :)

francislavoie:
Hmm. I noticed that I do use some env variables to help configure my Caddyfile and I've been using a `CADDY_` prefix for those. I don't have one called `CADDY_UPGRADE`, but it got me thinking that there might be overlap in these kinds of cases. Why not write to a file in the `/root/.caddy` directory or whatever instead? Just a thought.

elcore:
> Why not write to a file in the ~/.caddy directory or whatever instead? Just a thought.

:/ - This might add unnecessary complexity, as we would need to write a file, parse it and delete it afterwards

mholt:
@francislavoie Okay, I'll give it two underscores or something.

francislavoie:
To clarify, the point I'm getting at is that env vars are kinda external/global things that aren't fully under Caddy's control, but writing to a file in a directory managed by Caddy or something like that is probably safer to depend on. I'm sure it'll be "fine" but it's one of those edge cases that could happen ¯\_(ツ)_/¯

mholt:
It's a nice idea, but I think an env var is much more practical and less error-prone in this case. ;)

mholt:
Updated the env var to use two underscores; it's unlikely enough that will collide. Will probably merge this in a day or so.

PR #1815, opened 2017-08-08
browse: Support absolute and recursive directory symlinks

wader:

Support absolute symlink to directory.
Support recursive symlink to directory.

Could there be some security issue with absolute symlinks? make sure they are inside root or path spec?

### 4. Checklist

### DIFF
--- a/caddyhttp/browse/browse.go
+++ b/caddyhttp/browse/browse.go
@@ -246,7 +246,9 @@ func directoryListing(files []os.FileInfo, canGoUp bool, urlPath string, config 
 			}
 		}
 
-		if f.IsDir() {
+		isDir := f.IsDir() || isSymlinkTargetDir(f, urlPath, config)
+
+		if isDir {
 			name += "/"
 			dirCount++
 		} else {
@@ -260,7 +262,7 @@ func directoryListing(files []os.FileInfo, canGoUp bool, urlPath string, config 
 		url := url.URL{Path: "./" + name} // prepend with "./" to fix paths with ':' in the name
 
 		fileinfos = append(fileinfos, FileInfo{
-			IsDir:     f.IsDir() || isSymlinkTargetDir(f, urlPath, config),
+			IsDir:     isDir,
 			IsSymlink: isSymlink(f),
 			Name:      f.Name(),
 			Size:      f.Size(),
@@ -291,19 +293,19 @@ func isSymlinkTargetDir(f os.FileInfo, urlPath string, config *Config) bool {
 	if !isSymlink(f) {
 		return false
 	}
-	fullPath := func(fileName string) string {
-		fullPath := filepath.Join(string(config.Fs.Root.(http.Dir)), urlPath, fileName)
-		return filepath.Clean(fullPath)
-	}
-	target, err := os.Readlink(fullPath(f.Name()))
+
+	// a bit strange but we want Stat thru the jailed filesystem to be safe
+	target, err := config.Fs.Root.Open(filepath.Join(urlPath, f.Name()))
 	if err != nil {
 		return false
 	}
-	targetInfo, err := os.Lstat(fullPath(target))
+	defer target.Close()
+	targetInto, err := target.Stat()
 	if err != nil {
 		return false
 	}
-	return targetInfo.IsDir()
+
+	return targetInto.IsDir()
 }
 
 // ServeHTTP determines if the request is for this plugin, and if all prerequisites are met.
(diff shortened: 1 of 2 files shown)


mholt:
Hi Mattias, thanks for your contribution! I or another collaborator will look at this more soon. I'll have to think and research about absolute symlinks. I figure if the site owner puts the symlinks within the site root, that's kind of their own doing, right? Is it really a security issue then? It could be useful. But I can see how it could be an accident. (Still, it wouldn't be Caddy's fault.)

wader:
Ok thanks! Yeah i'm leaning towards the same conclusion too. Currently i think you already can symlink to things outside root/pathspec as long as the symlink is relative (haven't seen any checks  in the source).

One scenario i guess is if caddy is serving content under some kind of user control and an attacker can somehow indirectly create symlinks via extracted tar files etc. But that is a bit unrelated to this PR. My change is actually just about when to add a "/" at the end of the URL and have correct icon 😄

If more options is an option they could be a "follow symlinks" and "no symlinks pointing outside root/pathspec".

caddyhttp/browse/browse.go
```diff
@@ -291,18 +293,13 @@ func isSymlinkTargetDir(f os.FileInfo, urlPath string, config *Config) bool {
...
-		return filepath.Clean(fullPath)
-	}
-	target, err := os.Readlink(fullPath(f.Name()))
-	if err != nil {
-		return false
-	}
-	targetInfo, err := os.Lstat(fullPath(target))
+
+	fullPath := filepath.Join(string(config.Fs.Root.(http.Dir)), urlPath, f.Name())
+	targetInfo, err := os.Stat(fullPath)
```
wader: `fullPath` function and `os.Readlink` should not be needed anymore. When we use `os.Stat` instead of `os.Lstat` the OS will take care resolving the symlink and stat what it points to.

I get a bit concern that `os.*` functions are used here without cleaning `urlPath` (seems to under total user control). Original code cleaned fullPath but then it's too late i think. Maybe we can stat thru `config.Fs.Root.Open(..).Stat()` somehow?

wader: This seems to work:
```go
func isSymlinkTargetDir(f os.FileInfo, urlPath string, config *Config) bool {
	if !isSymlink(f) {
		return false
	}

	target, err := config.Fs.Root.Open(filepath.Join(urlPath, f.Name()))
	if err != nil {
		return false
	}
	defer target.Close()
	targetInto, err := target.Stat()
	if err != nil {
		return false
	}

	return targetInto.IsDir()
}
```

That should make sure the symlink we are following is located inside `config.Fs.Root`.

With the original code im not really sure what an attacker could do, maybe stat files outside the root using a relative urlPath joined with the symlink target path.

mholt: @wader Okay, this change looks like a good improvement. Let's try it out.

caddyhttp/browse/browse_test.go
```diff
@@ -453,3 +457,146 @@ func TestBrowseRedirect(t *testing.T) {
 		}
 	}
 }
+
+func TestDirSymlink(t *testing.T) {
```
mholt: I am loving these tests. :+1:

caddyhttp/browse/browse_test.go
```diff
@@ -453,3 +457,146 @@ func TestBrowseRedirect(t *testing.T) {
...
+	testCases := []struct {
+		source       string
+		target       string
+		pathScope    string
+		url          string
+		expectedName string
+		expectedURL  string
+	}{
+		// test case can expect a existing diretory "dir" and a symlink to it called "symlink"
```
wader: "diretory" typo, will fix

caddyhttp/browse/browse_test.go
wader: 😄

wader:
Added a comment about needing to `Open` to `Stat` using `http.FileSystem`

mholt:
Thanks, this LGTM. (Glad you added the comment about having to Open to Stat; I'm thinking of making my own interface that has a Stat method...)

Ready to merge? If so I'll do it/

wader:
Ready. Yeah own interface is probably a good idea. Thanks and thanks for caddy!

mholt:
Thanks @wader! Appreciate your contribution and the careful thought and the tests. I'll invite you to be a collaborator on the repo so that you can continue to contribute to Caddy, but using a branch rather than a fork. And also help with code reviews if you want, etc. I have a little guide here that might be helpful: https://github.com/mholt/caddy/blob/master/.github/CONTRIBUTING.md#collaborator-instructions

Welcome to the project!

wader:
Ok nice 👍

PR #1823, opened 2017-08-13
Fix restart restart behaviour of the systemd service

klingtnet:
This is a follow up to #1719 

In short, we are now using systemd's restart limit default that is 5 allowed service restarts in 10 seconds and the restart directive was changed to `on-abnormal`, i.e. it won't restart on non-zero exit codes (e.g. on a bad Caddyfile).

### DIFF
--- a/dist/init/linux-systemd/caddy.service
+++ b/dist/init/linux-systemd/caddy.service
@@ -5,9 +5,7 @@ After=network-online.target
 Wants=network-online.target systemd-networkd-wait-online.service
 
 [Service]
-Restart=on-failure
-StartLimitInterval=86400
-StartLimitBurst=5
+Restart=on-abnormal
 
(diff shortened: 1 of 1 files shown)


wmark:
I am commenting here because I've been invited by the OP in the original thread.

The description of *systemd's* behaviour is off. `on-abnormal` (equivalent to `on-abort` in this context) doesn't restart if the processes exited on his own terms, the return code is irrelevant. Restarts only happen if, for example, the kernel kills the process due to accessing invalid memory or the such.  
→ https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

LGTM, but squash all commits into one.

klingtnet:
Squashed.

mholt:
Thank you!

Issue #1723, opened 2017-06-18
Caddy panics on request, "failed to create new OS thread" and "may need to increase max user processes"


Southclaws:
### 1. What version of Caddy are you using (`caddy -version`)?
0.10.3 - was using 0.10.2 and installed the latest yesterday hoping it would resolve this issue

### 2. What are you trying to do?
Run Caddy as a reverse proxy to some services and basic static site host for a couple of extremely small, single-page portfolios, it was working well for months then randomly decided to start dying on requests.

The machine is one of these: https://documentation.online.net/en/dedicated-server/offers/2016/server-dedibox-classic-2016 and is pretty under-used in terms of resources, the largest apps I have running on there are some Java services (Confluence, Bitbucket and Jira - which I use Caddy to proxy to, successfully for a few months)

I stripped away all the reverse-proxies and the issue is still happening on a single basic static `root` site, sometimes it loads once but most of the time it just panics as soon as the request is sent.

I found one other post about this issue that mentioned `ulimit -u` (mine is 128150, no idea what's considered "normal" but this was what my Ubuntu 16.04 installation came with by default)

### 3. What is your entire Caddyfile?
The stripped down version is as simple as it gets, the page loads sometimes (and it loaded every single time, consistently, up until recently) so I know the directories are all set up correctly.

```text
https://southcla.ws {
        root /var/www/southcla.ws
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
The stock systemd unit file

### 5. Please paste any relevant HTTP request(s) here.
Just a simple `curl https://southcla.ws` or browser load causes this.

### 6. What did you expect to see?
A working site.

### 7. What did you see instead (give full error messages and/or log)?

```
Jun 18 16:28:02 southclaws caddy[27042]: Activating privacy features... done.
Jun 18 16:28:02 southclaws caddy[27042]: https://southcla.ws
Jun 18 16:28:02 southclaws caddy[27042]: 2017/06/18 16:28:02 https://southcla.ws
Jun 18 16:28:02 southclaws caddy[27042]: http://southcla.ws
Jun 18 16:28:02 southclaws caddy[27042]: 2017/06/18 16:28:02 http://southcla.ws
(I curl or browse to https://southcla.ws to trigger the panic)
Jun 18 16:28:07 southclaws caddy[27042]: runtime: failed to create new OS thread (have 13 already; errno=11)
[...]

PR #1825, opened 2017-08-14
Double systemd LimitNPROC to allow more threads

thejmazz:

`LimitNPROC` in `caddy.service` changed from `64` to `512`

This change solved #1723 for me. I arbitrarily doubled 64 to 128, not sure what is a better value. The error seems to be that caddy requests a 65th thread and then gets denied. Not sure if relevant to #1823 since perhaps that will affect restarting (no more "restart too quickly" errors?). Pinging @wmark

None.

### 4. Checklist

- [x] I have squashed any insignificant commits (only one commit)

### DIFF
--- a/dist/init/linux-systemd/caddy.service
+++ b/dist/init/linux-systemd/caddy.service
@@ -26,7 +26,7 @@ TimeoutStopSec=5s
 ; Limit the number of file descriptors; see `man systemd.exec` for more limit settings.
 LimitNOFILE=1048576
 ; Unmodified caddy is not expected to use more than that.
-LimitNPROC=64
+LimitNPROC=512
 
 ; Use private /tmp and /var/tmp, which are discarded after caddy stops.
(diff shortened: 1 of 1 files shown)


wmark:
I'd set this to 512 to be on the safe side.

`ulimit -u` or `LimitNPROC` limits processes and threads (in the whole tree, with `caddy` as its parent). You can run into this limit on servers with lots of cores (*Golang* will start C×cores threads if not instructed otherwise) or if you use a plugin which starts its own processes (≠ *goroutines*).

Anyway, operators configure a limit here to catch misbehaviour early. (See also `max smb processes` on *Samba*.)

The natural limit is the »largest« thread count you can expect these days on a 8-way server. 28 cores it is since Broadwell on non-public cpus, and since Skylake for the public. 8× 28×2 +1 = 449.

mholt:
Thank you!

PR #1832, opened 2017-08-18
Resource NOFILE rlimit support

zaolin:
Sets the NOFILE rlimit if not set to the minimal required value.
Offers a commandline option for setting a custom NOFILE rlimit value.

Signed-off-by: Philipp Deppenwiese <zaolin@das-labor.org>

Sets the NOFILE rlimit for unix operating systems. Normally this could be done as well by
the init system. The problem comes with non systemd OS and privilege separation if used.

This patch sets the default rlimit to 8192 instead of log a warning and offers the ability to set
a custom NOFILE rlimit value via "rlimit" commandline option.

It does checks for the UID in order to figure out if the resource capability is enabled. In general
golang does not support capabilities and privilege separation (see issue https://github.com/golang/go/issues/1435 ) until now.

Remove unneccessary NOFILE limits in init scripts for posix operating systems. 

-

An additional commandline option needs to be documented. See https://caddyserver.com/docs/cli

-rlimit
Sets the open file descriptors limit for the caddy process. Default limit is 8192. For non-root user
usage please refer to [setcap](http://man7.org/linux/man-pages/man8/setcap.8.html). The resource capability can be set by executing:
```
sudo setcap cap_sys_resource+ep caddy
```

### 4. Checklist

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -43,7 +43,7 @@ func init() {
 	flag.StringVar(&serverType, "type", "http", "Type of server to run")
 	flag.BoolVar(&version, "version", false, "Show version")
 	flag.BoolVar(&validate, "validate", false, "Parse the Caddyfile but do not start the server")
-	flag.Uint64Var(&rlimitNofile, "rlimit-nofile", 8192, "Sets the NOFILE rlimit for the caddy process")
+	flag.Uint64Var(&rlimitNofile, "rlimit-nofile", 16384, "Sets the NOFILE rlimit for the caddy process")
 
 	caddy.RegisterCaddyfileLoader("flag", caddy.LoaderFunc(confLoader))
 	caddy.SetDefaultCaddyfileLoader("default", caddy.LoaderFunc(defaultLoader))


zaolin:
No problem!. I planned to write tests but I am not sure how to do it in this case. Maybe you can give me a hint. :)

mholt:
Re: tests - is there a way to do a system call of some sort to verify that the new value is set properly?

rlimit_posix.go
```diff
@@ -4,20 +4,26 @@ package caddy
...
-// checkFdlimit issues a warning if the OS limit for
-// max file descriptors is below a recommended minimum.
-func checkFdlimit() {
-	const min = 8192
-
-	// Warn if ulimit is too low for production sites
+// SetRlimitOpenFiles set rlimit for NOFILE
+// setcap cap_sys_resource+ep for the caddy binary is needed
+// if it is executed as non-root user
+func SetRlimitOpenFiles(limit uint64) {
```
slightfoot: Shouldn't this function also check to see if the value passed in meets the minimum value of 8192?

slightfoot:
@zaolin Regarding tests, you could use the `Getrlimit` syscall the same as the previous `checkFdlimit` function to verify the limit has been set. I suggest you also check how other command flags are tested and you can do the same test on yours. Perhaps a test to make sure a value smaller than the minimum is correctly rejected.

rlimit_posix.go
zaolin: Good question. In my opinion the user should be aware of what he is doing. If someone wants to do that, he should do it. I could add a warning in this case.

zaolin:
My plan wasn't to test the SetRlimit syscall.(golang has already tests for it) :) Instead I planned to test the setrlimit feature with/without root (caps) access. Is that possible via travis ci ?

mholt:
Probably not without modifying the Travis script. Doable, but something to consider as we implement a better test suite in the future.

However, while we're on this topic, any harm in increasing the limit even higher than 8192? Would setting it to double that work well?

The way I'm seeing this change so far is, instead of the user running `ulimit -n` they have to run `setcap` (which isn't available in some constrained systems apparently). Wouldn't it be better for the sysadmin to just configure the system themselves instead, since they have to run a command either way?

zaolin:
Increasing limit to 16384 shouldn't be a problem.

I think it's an issue on non systemd OS because systemd supports setting resources. Normally on server systems there is no systemd and the sysadmin wants to do privilege separation in order to keep his machine safe. But you can only set ulimit user based which means there is no way setting ulimit if you use a no shell user, e.g. www-data user.

Capabilities are in my opinion the best solution. It's also easy for linux package maintainers. The solution doesn't work a cross distributions but that's how it is.

mholt:
Dangit, merge conflict. Sorry :smile: I'm starting to catch up with things...

mholt:
@zaolin Any chance you could finish up this PR? (Sorry for the merge conflict. Shouldn't be too bad.)

mholt:
Closing due to inactivity. :(

PR #1833, opened 2017-08-18
Add forwardproxy to directives' list

sergeyfrolov:
Adds forwardproxy to directives' list

N/A

ASAIK, the documentation for plugins is handled at registration.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -459,13 +459,14 @@ var directives = []string{
 	"gzip",
 	"header",
 	"errors",
-	"authz",     // github.com/casbin/caddy-authz
-	"filter",    // github.com/echocat/caddy-filter
-	"minify",    // github.com/hacdias/caddy-minify
-	"ipfilter",  // github.com/pyed/ipfilter
-	"ratelimit", // github.com/xuqingfeng/caddy-rate-limit
-	"search",    // github.com/pedronasser/caddy-search
-	"expires",   // github.com/epicagency/caddy-expires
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/plugin.go
```diff
@@ -449,6 +449,7 @@ var directives = []string{
 
 	// directives that add listener middleware to the stack
 	"proxyprotocol", // github.com/mastercactapus/caddy-proxyprotocol
+	"forwardproxy",  // github.com/caddyserver/forwardproxy
```
mholt: This group is actually for plugins which implement [httpserver.ListenerMiddleware](https://godoc.org/github.com/mholt/caddy/caddyhttp/httpserver#ListenerMiddleware) which this does not (although I can see why you put it here) -- no big deal, just move it down to between `// directives that add middleware to the stack` and `"locale",` - so, still in the same place, but right now the grouping is just a little confusing. I should probably improve the wording of the comments too.

sergeyfrolov:
Oh, sure, my bad.
I moved it, not sure if new place is the best, but we can reorder later.

elvizlai:
When will this plugin work? I saw doc on caddy website, but no choice for built.

mholt:
As soon as @sergeyfrolov releases it. :) The website is ready for it.

sergeyfrolov:
@elvizlai @mholt thanks for your interest, `forwardproxy` is released!
It works for me, but if you encounter any issues, please file them at [forwardproxy repo](https://github.com/caddyserver/forwardproxy).

Issue #1698, opened 2017-06-01
[Feature] renew hook for automatically obtained certificates


fheusel:
Caddy is automatically obtaining LetsEncrypt certificates but I need to automate restarting services (e.g. dovecot, postfix) so that they load the renewed certificates. Would it be possible to implement some kind of renew hook, like it's available in [certbot](https://certbot.eff.org/docs/using.html#renewing-certificates)?

francislavoie:
I think this is related to https://github.com/mholt/caddy/issues/1683

mholt:
@fheusel Do you need the hook available from outside the Caddy process, or from within it (i.e. as part of a plugin or something)?

fheusel:
@mholt I'm not sure if I understand you correctly. I'd like caddy to be able to trigger a restart of some services each time caddy obtains a new certificate. Otherwise services like dovecot or postfix won't pick up the new certificates.

mholt:
@fheusel How does that restart happen? (i.e. push or pull method?)

francislavoie:
Could you use something like `inotify` to watch for file changes? I'd think that could also work.

elcore:
Theoretically, we could add an event hook to... 

https://github.com/mholt/caddy/blob/b5ec4622993655d69ee68f3b833a69179a830f3f/caddytls/client.go#L268-L335

... and build a plugin to execute a command [`os/exec`], triggered by the event

fheusel:
@mholt e.g `service dovecot restart` - so I suppose it's a push method?

@francislavoie that would be possible, but then one had to build another error prone "watching" service. And on different operating systems you'll need different solutions to watch for filechanges.

@elcore this sounds like the way I was hoping for.

elcore:
Hello @fheusel, 

I am currently working on this 😄

francislavoie:
Update, the new plan is the following:

1. Add new event hooks: `CertRenewEvent` (and `InstanceStartupEvent` @elcore?)
2. Add a new Caddyfile directive `on [event] [cmd]`
3. Deprecate the `startup` and `shutdown` directives and change them to be aliases for `on startup [cmd]`

Reference code: 
- Original PR: https://github.com/mholt/caddy/pull/1842
- Original plugin code: https://github.com/elcore/caddy-command

elcore:
>[...] and `InstanceStartupEvent` @elcore?

Caddy Command does not support `StartupEvent`. The event is emitted before the Caddyfile is loaded!

[...]

PR #1842, opened 2017-08-27
Add 'command' plugin

elcore:

1.  Add the 'command' plugin
2. Register a new Event Hook `CertRenewEvent`
3. Register a new Event Hook `InstanceStartupEvent`

#1698 

None

### 4. Checklist


abiosoft:
Mind adding a readme to the command repo to indicate how it is meant to be used ?

elcore:
@abiosoft 

> Mind adding a readme to the command repo to indicate how it is meant to be used ?

I am working on it 😄

abiosoft:
I can't help but feel `command` sounds generic. `tlscommand` maybe, considering this is limited to when tls cert has been renewed.

elcore:
> I can't help but feel command sounds generic. tlscommand maybe, considering this is limited to when tls cert has been renewed.

This might change in the near future, it a feature request by Matt H.

francislavoie:
Imo this could (should?) be extended to support multiple different events. I'm thinking more of a `command [event] [cmd] [args]` syntax, where this event specifically would be `certrenew`. Making it more generic would be nice.

Even better, why don't we replace `startup` and `shutdown` with `command startup` and `command shutdown` instead of this being a plugin?

elcore:
>Imo this could (should?) be extended to support multiple different events. I'm thinking more of a command [event] [cmd] [args] syntax, where this event specifically would be certrenew. Making it more generic would be nice.

> Even better, why don't we replace startup and shutdown with command startup and command shutdown instead of this being a plugin?

That is my plan (see above)

francislavoie:
Sorry, maybe I'm blind, but I don't see your plan anywhere?

elcore:
Hello @francislavoie,

> Sorry, maybe I'm blind, but I don't see your plan anywhere?

I have only expressed it vaguely 😄 -- "This might change in the near future, it a feature request by Matt H."

Anyway, I have a "working" local branch

## Syntax

`command [event] [cmd] [args]`

## Example

`command certrenew echo "This is an example"`

plugins.go
```diff
@@ -217,10 +217,12 @@ func RegisterPlugin(name string, plugin Plugin) {
 // EventName represents the name of an event used with event hooks.
 type EventName string
 
-// Define the event names for the startup and shutdown events
+// Define the event names for the startup, shutdown, instancestartup and certrenew events
```
francislavoie: Might as well just change this to "names for the various events", it'll get pretty wordy if you keep adding each one to the comment every time you set up a new event 😆

elcore: Right 😆

elcore:
I have pushed my local branch 😄

francislavoie:
Awesome @elcore :smiley: 

@mholt IMO this is something that should just be built-into Caddy, I don't think this should only be available via a plugin. I think there is room for plugins to do cool things with events, but just running a command seem like something that should be core. What do you think?

Like I suggested above, I think this should replace the `startup` and `shutdown` directives. I guess for backwards compat they can be kept as aliases to `command startup [args]` and `command shutdown [args]` until later and just marked as deprecated?

elcore:
> IMO this is something that should just be built-into Caddy, I don't think this should only be available via a plugin. [...] What do you think? 

ping @mholt

mholt:
Hi -- sorry, been really busy with other things.

I agree with the others. Let's switch startup and shutdown to use the event hooks, but let's keep those directives as syntactic as sugar. I think we can wrap this all into a standard `run` or `on` directive to run commands on certain events.

Thanks @elcore and others who reviewed!

PR #1848, opened 2017-08-29
Make filename column fill out space

wader:

Make filename column fill out extra space

### 4. Checklist

### DIFF
--- a/caddyhttp/browse/setup.go
+++ b/caddyhttp/browse/setup.go
@@ -222,11 +222,16 @@ th svg {
 }
 
 td {
+	white-space: nowrap;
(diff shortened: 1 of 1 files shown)


caddyhttp/browse/setup.go
```diff
@@ -222,11 +222,16 @@ th svg {
...
 	font-size: 14px;
 }
 
 td:first-child {
-	width: 50%;
+	width: 100%;
+}
+
+td:nth-child(2) {
+	padding: 0 20px 0 20px;
```
wader: this makes sure there is some padding space around the size column

Issue #1846, opened 2017-08-28
Path routing breaks FastCGI SCRIPT_NAME


jchv:
Hi. First of all, I apologize for the length of this issue report, as well as the fact that I did not follow the template. This is not quite a bug report as I am not confident that there is a bug in Caddy, but it does regard a behavior of Caddy that is unintuitive. As such, I had trouble fitting it in the template and gave up.

(I'd be happy to work on a PR for the issue if we did gain consensus that this was a bug.)

Basically, I'm trying to transition an existing PHP-based site over to Caddy. In NGINX, we might do something like this:
(Simplified, but mostly valid config.)

```
worker_processes 1;
events {
    worker_connections 1024;
}
http {
    server {
        listen 8000;
        location / {
            root /www/home;
        }
        location /forum {
            root /www/phpbb;
            location ~ \.php$ {
                fastcgi_pass php:9000;
            }
        }
        location /wiki {
            root /www/mediawiki;
            location ~ \.php$ {
                fastcgi_pass php:9000;
            }
        }
    }
}
```

What is intended to happen is pretty simple, /forum will be separate for purposes of configuration. In Caddy, at first glance, this maps well to the path-based routing, and in fact, for some software, this works. (I've noticed it works for MediaWiki.) It's worth noting that the above NGINX config isn't totally right because it will end up serving `/forum` to `/www/phpbb/forum/index.php` or so, so there are some simplifications made here, but hopefully the intent is clear.

Anyway, trying to do this sort of thing in Caddy yields something like this:
```
localhost:8000 {
    root /www/home
}
localhost:8000/forum {
    root /www/phpbb
    fastcgi / php:9000 php
}
localhost:8000/wiki {
    root /www/mediawiki
    fastcgi / php:9000 php
}
```

Which looks nice, but unfortunately it has a pretty debilitating issue: `$_SERVER["SCRIPT_NAME"]` and `$_SERVER["PHP_SELF"]` will end up evaluating to paths relative to the subdirectory routed, not the domain

This may not be a bug. I am not sure if this is intentional or not. I guess there's no actual standard for CGI. Drafts suggest that the consensus is that the "URI" information passed to the script need not match reality:
![image](https://user-images.githubusercontent.com/938744/29797463-2e027718-8c25-11e7-900e-6ddbc71401b7.png)

[...]

PR #1852, opened 2017-09-01
Fix SCRIPT_NAME for FastCGI

jchv:
As described in #1846, many PHP apps depend on the `SCRIPT_NAME` CGI environment variable. The Caddy FastCGI plugin sets `SCRIPT_NAME` to roughly what would be expected by NGINX (essentially, a path to the script relative to the URL.) Unfortunately, when using path-based routing with VHosts, the path prefix gets lost, leading to PHP software such as phpBB and anything else depending on Symphony's requestUri logic failing (and probably more stuff too.)

This patch does exactly one thing: it finds the path prefix via the request context and prepends it to the calculated `SCRIPT_NAME`. That's it. This simple change seems to be all you need to do in the circumstances I was able to think of.

This patch affects #1846.

I don't believe any changes need to be made, since this simply brings Caddy's behavior into accordance with what one would expect from any other FastCGI server.

### 4. Checklist

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -18,6 +18,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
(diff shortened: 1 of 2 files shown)


caddyhttp/fastcgi/fastcgi.go
```diff
@@ -248,6 +249,11 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 	// Strip PATH_INFO from SCRIPT_NAME
 	scriptName = strings.TrimSuffix(scriptName, pathInfo)
 
+	// Add vhost path prefix to scriptName. Otherwise, some PHP software will
+	// have difficulty discovering its URL.
+	pathPrefix, _ := r.Context().Value(caddy.CtxKey("path_prefix")).(string)
+	scriptName = pathPrefix + scriptName
```
abiosoft: Maybe `path.Join` here for consistency in case pathPrefix has a trailing slash.

jchv: Makes sense to me. I actually didn't think the standard library `path.Join` was applicable, but upon checking it seems like it is.

jchv:
Thanks @mholt. I'm a big fan and user of Caddy and hope I can contribute more over time.

PR #1857, opened 2017-09-06
should use bytes.Contains(content, []byte(FAILED)) instead and should…

wgliang:
… use strconv.Itoa instead of strconv.FormatInt

should use bytes.Contains(content, []byte("FAILED")) instead and should use strconv.Itoa instead of strconv.FormatInt

NONE
NONE

### 4. Checklist


mholt:
Fair enough. Thank you!

PR #1861, opened 2017-09-08
httpserver: Fix #1859 by cleaning paths when matching them

### LINKED ISSUE #1859: Multiple forward slashes in paths
Multiple forward slashes are not collapsed in paths

Caddy 0.10.7

Use path matches with `status`, `rewrite` and `internal`.

```text
:8001 {
    status 401 /secret/files
}

:8002 {
    rewrite /secret/files {
        regexp ^/(.*)
        to /notsecret/files/{1}
    }
}

:8003 {
    internal /secret/files
}
```

`caddy` with no arguments, Windows 64 bit

* `curl localhost:8001/secret/files/file.txt` returns 401 as expected
* `curl localhost:8001//secret//files/file.txt` serves the secret file (all variations of multiple slashes serves the file.
* `curl localhost:8002/secret/files/file.txt` serves the public file after rewriting, as expected
* `curl localhost:8002/secret//files/file.txt` serves the secret file
* `curl localhost:8003/secret/files/file.txt` returns 404 as expected
* `curl localhost:8003/secret//files/file.txt` serves the secret file.

That multiple slashes in the path was collapsed or matched by the path expressions.

They were not

I have attached a small test site as a zip.
[caddy_bug.zip](https://github.com/mholt/caddy/files/1285420/caddy_bug.zip)

mholt:

When checking if one path matches another using Path.Matches(), call path.Clean() on the paths to compress/merge multiple slashes. Note that it does NOT effectively match paths where `p == "/foo"` and `base == "//foo"` (this will be a match even though it looks like it shouldn't be) - however I think this is not a concern, as removing slashes is rare... also note that this change doesn't mutilate the request URI at all, it's only a side-effect-less change for path matching. It should fix an issue for all middlewares that use path.Matches. If they use their own path matcher, though, they're on their own...

#1859

Maybe a note somewhere that multiple slashes are merged for path matching, similar to nginx's location block, assuming `merge_slashes` is NOT `off`.

### 4. Checklist

/cc @hstaugaard @magikstm

### DIFF
--- a/caddyhttp/browse/browse_test.go
+++ b/caddyhttp/browse/browse_test.go
@@ -227,7 +227,7 @@ func TestBrowseTemplate(t *testing.T) {
 func TestBrowseJson(t *testing.T) {
 	b := Browse{
 		Next: httpserver.HandlerFunc(func(w http.ResponseWriter, r *http.Request) (int, error) {
-			t.Fatalf("Next shouldn't be called")
+			t.Fatalf("Next shouldn't be called: %s", r.URL)
 			return 0, nil
 		}),
 		Configs: []Config{
--- a/caddyhttp/httpserver/path.go
+++ b/caddyhttp/httpserver/path.go
@@ -2,6 +2,7 @@ package httpserver
 
 import (
 	"net/http"
+	"path"
 	"strings"
 )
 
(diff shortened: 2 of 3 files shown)


tgulacsi:
ide://open?url=file:///home/user/text.txt&line=15 should be ide://open?url=file:%2F%2F%2Fhome%2Fuser%2Ftext.txt&line=15 anyway...

PR #73, opened 2015-05-12
Fix for Issue 72: Markdown: 500 for YAML metadata

abiosoft:

PR #1870, opened 2017-09-14
Add support for SRV upstreams in fastcgi

Gufran:
Adds support in fastcgi module to discover upstreams using SRV lookup

closes: https://github.com/mholt/caddy/issues/1803

The fastcgi module documentation (https://caddyserver.com/docs/fastcgi) requires an update:

``` diff
- endpoint is the address or Unix socket of the FastCGI server.
+ endpoint is the address or Unix socket of the FastCGI server. Service discovery using SRV lookup is supported. If the endpoint start with `srv://` it will be considered as a service locator and caddy will attempt to resolve available services via SRV DNS lookup.
```

``` diff
- upstream specifies an additional backend to use. Basic load balancing will be performed. This can be specified multiple times.
+ upstream specifies an additional backend to use. Basic load balancing will be performed. This can be specified multiple times. upstream directive is not supported if endpoint is a service locator.
```

### 4. Checklist

I'm sorry that the tests are not super clean code, I had to employ a dirty trick to prevent tests from making network calls.

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -20,6 +20,7 @@ package fastcgi
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -107,7 +108,11 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 			}
 
 			// Connect to FastCGI gateway
-			network, address := parseAddress(rule.Address())
+			address, err := rule.Address()
+			if err != nil {
+				return http.StatusBadGateway, err
+			}
+			network, address := parseAddress(address)
 
 			ctx := context.Background()
 			if rule.ConnectTimeout > 0 {
@@ -381,7 +386,7 @@ type Rule struct {
 type balancer interface {
 	// Address picks an upstream address from the
 	// underlying load balancer.
-	Address() string
+	Address() (string, error)
 }
 
 // roundRobin is a round robin balancer for fastcgi upstreams.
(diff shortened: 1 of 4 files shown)


Gufran:
@francislavoie I could use some feedback here 😃

francislavoie:
You know more than I do @Gufran, I was just helping with issue triage :smile: 

Thanks for the PR, I'll defer to @abiosoft for review.

My understanding of the issue was to have `proxy` load balancing via a DNS policy, not fastcgi, but this might be fine too.

Gufran:
I'll work on proxy package today in a separate PR, fastcgi was the quicker and easier change so I did this first.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -18,6 +18,8 @@ import (
 	"sync/atomic"
 	"time"
```
abiosoft: delete empty line to have `gofmt` rearrange this.

caddyhttp/fastcgi/setup.go
```diff
@@ -147,13 +159,28 @@ func fastcgiParse(c *caddy.Controller) ([]Rule, error) {
...
+		} else {
+			rule.balancer = &roundRobin{addresses: upstreams, index: -1}
+		}
 
 		rules = append(rules, rule)
 	}
 	return rules, nil
 }
 
+func parseSRV(locator string) (*srv, error) {
```
abiosoft: a panic is bound to happen when the `len(location) <= 6`, a check could be for that and an error can be returned.

caddyhttp/fastcgi/setup.go
```diff
@@ -62,8 +64,14 @@ func fastcgiParse(c *caddy.Controller) ([]Rule, error) {
 			Root: absRoot,
 			Path: args[0],
 		}
+
 		upstreams := []string{args[1]}
 
+		srvUpstream := false
+		if upstreams[0][:6] == "srv://" {
```
abiosoft: Though it is a user error if `len(upstreams[0]) < 6`, it is preferred to avoid panics. `strings.HasPrefix` is a better candidate here.

abiosoft:
Looks like you know more than many of us 😄 
> You know more than I do @Gufran, I was just helping with issue triage 😄

This looks good, just some bunch of little cleanups.

Thanks for this and welcome to the list of Caddy contributors.

caddyhttp/fastcgi/setup.go
Gufran: It will not panic because control will only reach `parseSRV` if the check on line #73 guarantees at least 6 characters and those will be `srv://`. I however added a check to raise an error if the locator consists of only those 6 characters and nothing else.

Gufran:
@abiosoft Thanks for feedback and a warm welcome. I addressed the points you raised. We should be ready to roll now.

mholt:
Is this ready for a final review then? @abiosoft I think you're the most qualified to do it, let me know if you want another set of eyes though...

Gufran:
@mholt Yep, I guess we are ready for final review.

Gufran:
@abiosoft @mholt what's the status on this? I am waiting for this to conclude before I start working on proxy package.

caddyhttp/fastcgi/setup.go
```diff
@@ -2,10 +2,14 @@ package fastcgi
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 	"path/filepath"
+	"strings"
 	"time"
```
abiosoft: delete whitespace and gofmt to rearrange imports.

caddyhttp/fastcgi/setup_test.go
```diff
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"testing"
```
abiosoft: same as above.

caddyhttp/fastcgi/setup.go
Gufran: @abiosoft both fixed

Gufran:
@mholt I could really use your input here :) Please let me know if you need me to make any other change. Otherwise I can start my work on other packages.

mholt:
Right, this looks good, sorry. Thank you!

PR #1864, opened 2017-09-12
Implement 'on' plugin

elcore:
### 1. What does this change do, exactly?

1. Implement the 'on' plugin

### 2. Please link to the relevant issues.

#1698, #1842

### 3. Which documentation changes (if any) need to be made because of this PR?

- Add documentation for `on`

### 4. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have squashed any insignificant commits
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I am willing to help maintain this change if there are issues with it later

PR #1879, opened 2017-09-16
Implement CertRenewEvent

elcore:

1. Register a new Event Hook `CertRenewEvent`

#1864 

/

### 4. Checklist

### DIFF
--- a/caddytls/client.go
+++ b/caddytls/client.go
@@ -345,6 +345,9 @@ func (c *ACMEClient) Renew(name string) error {
 		return errors.New("too many renewal attempts; last error: " + err.Error())
 	}
 
+	// Executes Cert renew events
+	caddy.EmitEvent(caddy.CertRenewEvent, name)
+
 	return saveCertResource(storage, newCertMeta)
 }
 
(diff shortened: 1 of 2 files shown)


elcore:
**NOTE: Merge this PR before #1864**

caddytls/client.go
```diff
@@ -331,6 +331,9 @@ func (c *ACMEClient) Renew(name string) error {
 		return errors.New("too many renewal attempts; last error: " + err.Error())
 	}
 
+	// Executes Cert renew events
+	caddy.EmitEvent(caddy.CertRenewEvent, nil)
```
mholt: What should we pass as data in the second argument?

caddy/caddymain/run.go
```diff
@@ -126,6 +126,9 @@ func Run() {
 		mustLogFatalf("%v", err)
 	}
 
+	// Executes Instance Startup events
+	caddy.EmitEvent(caddy.InstanceStartupEvent, nil)
```
mholt: What should we pass as data to this event?

elcore: instance

caddytls/client.go
elcore: Nothing

mholt: But won't hooks want to know which certificate was renewed, at least?

elcore: Great idea!

Issue #1882, opened 2017-09-18
Option for http.errors to handle errors thrown from php (fastcgi)


infinityCounter:
### 1. What version of Caddy are you using (`caddy -version`)?
0.10.9

### 2. What are you trying to do?
Log error responses (HTTP Status Code >=400) using the errors middleware

### 3. What is your entire Caddyfile?
```
localhost { 
  tls self_signed
  root ./App/
  gzip
  log / ./App/Logs/caddyaccess.log "{remote} - {user} [{when}] \"{method} {uri} -> {rewrite_path} {proto}\" {status} {size}" {
  	  rotate_size 50  # Rotate after 50 MB
  	  rotate_age  90  # Keep rotated files for 90 days
  	  rotate_keep 20  # Keep at most 20 log files
  	  rotate_compress # Compress rotated log files in gzip format
  }
  errors ./App/Logs/caddyerrors.log {
	  rotate_size 50 
	  rotate_age  90 
	  rotate_keep 20
	  rotate_compress 
  }

  # Following rewrite blocks does an internal
  # URL rewrite for all resources (NOT PHP FILES) to ./App/Assets folder
  rewrite / {
 	  regexp /.*\..+
      ext    !.php
      to     /Assets/{path}
  }

  # Rewrite all requests for anything else to root
  rewrite / {
   	  to     /
  }

  fastcgi / /var/run/php/php7.0-fpm.sock {
      ext .php
      split .php
      index index.php
      connect_timeout "120s"
  }
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
caddy (Ran from born again shell in directory with Caddyfile)

### 5. Please paste any relevant HTTP request(s) here.

Not relevant

### 6. What did you expect to see?
I expected the content of my 500 response to be logged to into ./App/Log/caddyerrors.log

### 7. What did you see instead (give full error messages and/or log)?
No 500 response were logged

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
Start a php application and try the following code in an index.php and test from any browser
```
try{
    throw new Exception("Problemo");
}catch(Exception $e)
{
    http_response_code(500);
    echo $e->getMessage() . ", this should be in your log file but it's not.";
    flush();
}
```

tobya:
Thank you for the issue.  I assume by the way you have phrased your question that if you set the reponse code to < 400 it does get printed?  Does anything ever get printed to the error log?

I'm wondering does the server consider a status code actually an error?

[...]

PR #1883, opened 2017-09-18
Updated fastcgi.go to return the response body, as an error, of any r…

infinityCounter:
…esponse with a status code indicating an error.

Updated the fastcgi middleware to now return an error status code as long as the http response from the fastcgi server indicates an error, along with the response body as the error.

#1882 

Don't think so

### 4. Checklist

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -18,6 +18,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"io/ioutil"
+
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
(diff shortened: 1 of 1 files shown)


caddyhttp/fastcgi/fastcgi.go
```diff
@@ -157,17 +159,20 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 				return http.StatusBadGateway, err
 			}
 
+			respBody, err := ioutil.ReadAll(resp.Body)
```
mholt: This is a little unsafe, what if the response body is a large file download (GBs in size)?

Also, the body has already been drained above...

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -157,17 +159,20 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
 				// Remove trailing newline, error logger already does this.
 				err = LogError(strings.TrimSuffix(fcgiBackend.stderr.String(), "\n"))
 			}
 
-			// Normally we would return the status code if it is an error status (>= 400),
-			// however, upstream FastCGI apps don't know about our contract and have
-			// probably already written an error page. So we just return 0, indicating
-			// that the response body is already written. However, we do return any
-			// error value so it can be logged.
+			// Return any error value so it can be logged.
```
mholt: I don't think we can just ignore the behavior that is very clearly explained in the comments. :) It fixed a bug, hence the comment.

infinityCounter: Sorry about that, I didn't want to leave the comment there since I had changed the reasoning. I'll definitely put it back :)

caddyhttp/fastcgi/fastcgi.go
infinityCounter: I could not think of a scenario where someone would response with a multi GB response with an error status code, but I see your point.

infinityCounter:
@mholt  Working on those tests and fixing those bug! Definitely putting those comments back as well, sorry for the trouble.

mholt:
No problem, @infinityCounter - I will go ahead and close this PR for now, until it is in a state where it can be reviewed, so feel free to reopen it if you push new commits.

PR #1884, opened 2017-09-18
Fix pid error in linux sysvinit

timothywlewis:

Fixes two issues with the `dist/init/linux-sysvinit/caddy` init script, specifically it fixes an issue where the log isn't created and secondly it fixes a spurious error in the log file because the daemon user isn't responsible for writing the pid file.

None as far as I am aware.

None.

### 4. Checklist

- [N/A] I have written tests and verified that they fail without my change
- [Yes] I have squashed any insignificant commits
- [N/A] This change has comments for package types, values, functions, and non-obvious lines of code
- [Yes] I am willing to help maintain this change if there are issues with it later

### DIFF
--- a/dist/init/linux-sysvinit/caddy
+++ b/dist/init/linux-sysvinit/caddy
@@ -36,6 +36,7 @@ ulimit -n 8192
 
 start() {
     $USERBIND $DAEMON
+    touch $LOGFILE && chown $DAEMONUSER $LOGFILE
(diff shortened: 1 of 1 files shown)


fakeid30:
Looks good.

timothywlewis:
How does this look @djvdorp?

djvdorp:
I think the change is sane and looks good @timothywlewis!
Thank you for your contribution! That's a good improvement here.

PR #1913, opened 2017-10-12
Allow nonblocking onevent command with no other arguments

lbguilherme:
Take this for example:

    on startup /path/to/my/script.sh &

Currently it will run this script as blocking and passing "&" as if it were a parameter because it only checks for "&" if there is *more than one* argument. If the same is written this way, it works:

    on startup /path/to/my/script.sh dummy &

Now this is run non-blocking.

The change itself is very minimal. I don't know how to write a test to check this new behavior.

### DIFF
--- a/onevent/hook/hook.go
+++ b/onevent/hook/hook.go
@@ -16,7 +16,7 @@ func (cfg *Config) Hook(event caddy.EventName, info interface{}) error {
 	}
 
 	nonblock := false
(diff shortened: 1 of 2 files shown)


mholt:
Thanks for your contribution, @lbguilherme! I'll have Eldin look at this since he recently worked on this code and is more familiar with it.

lbguilherme:
Updated with @elcore 's test

mholt:
Great work everyone! Thanks :+1:

PR #1914, opened 2017-10-12
Fix upstart script for Centos 6 (and Amazon Linux)

ekini:

Fixes upstart script so it actually works on Centos 6 and Amazon Linux. The script had comments that Centos6 upstart doesn't support some features, but it still used them.

https://github.com/mholt/caddy/pull/1189

None I can think of.

### 4. Checklist

### DIFF
--- a/dist/init/linux-upstart/caddy.conf.centos-6
+++ b/dist/init/linux-upstart/caddy.conf.centos-6
@@ -3,18 +3,10 @@ description "Caddy HTTP/2 web server"
 start on runlevel [2345]
 stop on runlevel [016]
 
-# centos 6 upstart version does not support console
-console log
-
-# centos 6 upstart version does not support setuid/setgid
-setuid www-data
-setgid www-data
+console output
 
 respawn
 respawn limit 10 5
-
-# centos 6 upstart version does not support reload
-reload signal SIGUSR1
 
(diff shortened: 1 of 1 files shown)


mholt:
Thanks -- but what about CentOS 7 users? These are kind of important (for example, not running Caddy as root).

ekini:
Centos 7 uses Systemd, so it should be covered by https://github.com/mholt/caddy/blob/master/dist/init/linux-systemd/caddy.service

mholt:
Okay. We can merge this but first can you make a comment or at least a note in the readme that systemd or a newer version of upstart should be used instead, to handle permissions properly?

ekini:
Sorry, I don't get it. Centos 6 uses older upstart, and there's no way to update it. Amazon Linux is based on Centos 6 and uses the same upstart. This change just updates upstart script `caddy.conf.centos-6` to do work with these upstart versions, doing the same things, and handling permissions correctly.

On the other hand, Centos7 uses systemd and doesn't even have upstart.

So if one uses Centos 6, they take `linux-upstart/caddy.conf.centos-6`.
If they use Centos 7, then take `linux-systemd/caddy.service`.

mholt:
Ah, I see now. That's unfortunate. I guess this is the best that can be done (?). Thanks for the fix, then!

PR #1922, opened 2017-10-15
Handle the case of an OCSP responder certificate expiring before an OCSP response it issued

alex:

If an OCSP response's issuing OCSP responder certificate expires before the OCSP response itself does, consider the OCSP responder's expiration date to be the end of the OCSP response's vality period, so that we obtain a new OCSP response sooner.

### 4. Checklist

### DIFF
--- a/caddytls/maintain.go
+++ b/caddytls/maintain.go
@@ -334,8 +334,15 @@ func DeleteOldStapleFiles() {
 // meaning that it is not expedient to get an
 // updated response from the OCSP server.
 func freshOCSP(resp *ocsp.Response) bool {
+	nextUpdate := resp.NextUpdate
+        // If there is an OCSP responder certificate, and it expires before the
+        // OCSP response, use its expiration date as the end of the OCSP
+        // response's validity period.
+	if resp.Certificate && resp.Certificate.NotAfter.Before(nextUpdate) {
+		nextUpdate = resp.Certificate.NotAfter
+	}
 	// start checking OCSP staple about halfway through validity period for good measure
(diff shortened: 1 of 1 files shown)


alex:
(untested)

caddytls/maintain.go
```diff
@@ -334,8 +334,15 @@ func DeleteOldStapleFiles() {
 // meaning that it is not expedient to get an
 // updated response from the OCSP server.
 func freshOCSP(resp *ocsp.Response) bool {
+	nextUpdate := resp.NextUpdate
+        // If there is an OCSP responder certificate, and it expires before the
+        // OCSP response, use its expiration date as the end of the OCSP
+        // response's validity period.
+	if resp.Certificate && resp.Certificate.NotAfter.Before(nextUpdate) {
```
mholt: This isn't JavaScript; it'll have to be `if resp.Certificate != nil && ...` :)

Otherwise this looks pretty good!

caddytls/maintain.go
```diff
@@ -334,8 +334,15 @@ func DeleteOldStapleFiles() {
 // meaning that it is not expedient to get an
 // updated response from the OCSP server.
 func freshOCSP(resp *ocsp.Response) bool {
+	nextUpdate := resp.NextUpdate
+        // If there is an OCSP responder certificate, and it expires before the
```
mholt: Ah, now the CI tests are complaining about the formatting. Needs `gofmt` - i.e. Comments need to be tab-indented and aligned with the rest of this block. Sorry I didn't catch this before. 😇

mholt:
Thanks @alex! I'm pushing a commit now to complement your change as an extra check.

Issue #1769, opened 2017-07-17
Log Privacy


myReaper:
Hi,

i just add this directly as "feature request" as I can't find any option for this in the documentation :-)

For privacy reasons it is crucial to strip the last or the last two octets of an IP address in the access.log . I'm currently using the "{combined}" format to log my websites and can't find any option in the documentation to strip them.

I think this is also mandatory for many caddy users because of privacy regulations in many countries.

Best Regards

francislavoie:
I think adding an additional replacer as an alternative to `{remote}` would be the solution here. Maybe something like `{remote_private}`? That way you can just specify the log format by hand, for example:

```
log / access.log "{remote_private} - - [{when}] \"{method} {uri} {proto}\" {status} {size}"
```

myReaper:
I'm no programmer, so I can't tell you what code-vise the best option would be :-)

Just from a user perspective, it would be great to still use the {combined} option and maybe just set an additional parameter like "anonymize-ip 1".

With the option to set 0 (default, for logging all octets) to 4 depending on how many octets i would like to anonymize (1 for 123.123.123.0, 3 for 123.0.0.0)

dmke:
Please note that for IPv6, the last two octets are not enough to get satisfying results. To quote this [SO answer](https://stackoverflow.com/a/6098698/766783):

> At the very least you want to strip the EUI-64 off, i.e the last 64 bits of the address.
> more realistically you want to strip quite a lot more to really be private, since the
> remaining part will still identify only one subnet (i.e. one house possibly)
>
> [...]
>
> You can implement this with a bitmask exactly like you would for an IPv4 address,
> the question becomes a legal one though of "how much do I need to strip to comply
> with the specific legislation", not a technical one at that point though

Based on the suggestion, I propose the following config block options:

    log path file [format] {
      rotate_...
      anonymize-v4    number-of-bits   # 0..32
      anonymize-v6    number-of-bits   # 0..128
    }

This could relatively easy be implemented using [`net#CIDRMask`](https://golang.org/pkg/net/#CIDRMask) and [`net.IP#Mask`](https://golang.org/pkg/net/#IP.Mask). Here's a small demo: https://play.golang.org/p/Ib6y4FwrCt

If performance is critical, this approach might not be suitable, since it requires parsing the `RemoteAddr` via `net#ParseIP`, and both `net.IP#Mask` and `net.IP#String` are not allocation-free...

tobya:
[...]

PR #1930, opened 2017-10-21
Allow Masking of IP address in Logfile.

tobya:
Fix for #1769.  Allows for a mask to be specified to mask the remote IP

new sub directive for the log directive

      log / mylogfile.log {common} {
        ipmask "255.255.0.0" , "ffff:ffff:ffff:ff00::"
      }

The first value after `ipmask` is the IP4 mask and the second the IP6 mask.  IP6 mask can be left out.  If you only want an IP6 mask for some reason just supply "" for IP4

This will provide output like this in the log

> 0.0.0.1 - - [21/Oct/2017:09:06:21 +0100] "GET /mailchimp/mailchimp-api/.git/ HTTP/1.1" 200 5255

I have implemented this by changing the value of r.RemoteAddr just before log calls replace.  This seemed the least bad way of doing it.  

I have left in log lines etc while in development.  I would appreciate any feedback on this PR.

I am not 😞 able to test IP6 values or masks so would be great if someone would test with an IP6 network and IP6 mask to see if it works, it should but any feedback or errors would be greatly appreciated.

I will work on Tests now.

Fix for #1769.   Thank you to @dmke for linking to a simple working example.

This will need an update to docs on `log` to explain ipmask as a new subdirective.  Once close to merging I can provide these documentation changes.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/logger.go
+++ b/caddyhttp/httpserver/logger.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"io"
 	"log"
+	"net"
 	"os"
 	"strings"
 	"sync"
@@ -37,9 +38,11 @@ var remoteSyslogPrefixes = map[string]string{
 type Logger struct {
 	Output string
 	*log.Logger
-	Roller *LogRoller
-	writer io.Writer
-	fileMu *sync.RWMutex
+	Roller   *LogRoller
+	writer   io.Writer
+	fileMu   *sync.RWMutex
+	V4ipMask string
+	V6ipMask string
 }
 
 // NewTestLogger creates logger suitable for testing purposes
@@ -62,6 +65,35 @@ func (l Logger) Printf(format string, args ...interface{}) {
 	l.fileMu.RLock()
 	l.Logger.Printf(format, args...)
 	l.fileMu.RUnlock()
+}
+
+func (l Logger) MaskIP(ip string) string {
+	var myIP net.IP
+	var maskIP net.IP
+	log.Println(ip)
+	log.Printf("%V+", net.ParseIP(ip))
+
+	// If unable to parse, simply return IP as provided.
+	myIP = net.ParseIP(ip)
+	if myIP == nil {
+		return ip
+	}
+
+	log.Println(l.V4ipMask)
+
+	if myIP.To4() == nil {
+		// ipV6
+		maskIP = net.ParseIP(l.V6ipMask)
+	} else {
+		maskIP = net.ParseIP(l.V4ipMask)
+
+	}
+	mask := net.IPMask(maskIP)
+	log.Println(mask.String())
+	log.Println(myIP.String())
+	log.Println(myIP.Mask(mask).String())
+	return myIP.Mask(mask).String()
+
 }
 
 // Attach binds logger Start and Close functions to
--- a/caddyhttp/httpserver/logger.go
+++ b/caddyhttp/httpserver/logger.go
@@ -41,8 +41,8 @@ type Logger struct {
 	Roller   *LogRoller
 	writer   io.Writer
 	fileMu   *sync.RWMutex
-	V4ipMask string
-	V6ipMask string
+	V4ipMask net.IPMask
+	V6ipMask net.IPMask
 }
 
 // NewTestLogger creates logger suitable for testing purposes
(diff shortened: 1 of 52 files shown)


tobya:
I have added tests.  I think this is ready to be reviewed and merged if the approach is right.

caddyhttp/log/log.go
```diff
@@ -66,6 +67,12 @@ func (l Logger) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 
 			// Write log entries
 			for _, e := range rule.Entries {
+				// Mask IP Address, will not modify if default
+				hostip, hostport, err := net.SplitHostPort(r.RemoteAddr)
+				if err == nil {
+					r.RemoteAddr = e.Log.MaskIP(hostip) + ":" + hostport
```
dmke: This won't work for IPv6 addresses. `2001:feed:c0fe::` + Port 1234 should be `[2001:feed:c0fe::]:1234`. Use `net.JoinHostPort`.

caddyhttp/httpserver/logger.go
```diff
@@ -64,6 +67,25 @@ func (l Logger) Printf(format string, args ...interface{}) {
...
+	var myIP, maskIP net.IP
+
+	// If unable to parse, simply return IP as provided.
+	myIP = net.ParseIP(ip)
+	if myIP == nil {
+		return ip
+	}
+
+	if myIP.To4() == nil {
+		maskIP = net.ParseIP(l.V6ipMask)
```
dmke: Is it feasible to move the `net.ParseIP` calls into the setup function? This would waste CPU cycles on each request.

Side note: if `l.V4ipMask`/`l.V6ipMask` were of type `*net.IP`, this would allow for an early return avoiding the `net.IPMask` call. Useful, when no masking should happen.

caddyhttp/log/setup.go
```diff
@@ -47,21 +48,45 @@ func logParse(c *caddy.Controller) ([]*Rule, error) {
...
-			if !httpserver.IsLogRollerSubdirective(what) {
+			if what == "ipmask" {
+
+				if len(where) == 0 {
+					return nil, c.ArgErr()
+				}
+
+				if where[0] != "" {
+					ip4Mask = where[0]
+					testmask := net.ParseIP(ip4Mask).To4()
```
dmke: `net.ParseIP` can return nil: https://golang.org/pkg/net/#ParseIP

dmke: Ah, nevermind, this won't panic.

tobya:
Thanks for the comments, I’ll make the changes over the weekend

tobya:
I have updated the code to take into account @dmke comments.   Also some tidying up.

tobya:
The documentation should be added to log page

IP Masking 
-----

IP addresses in log can be masked to provide annonymisation for corporate or legal reasons.

    log / access.log "{common}" {
      ipmask 255.255.0.0 "2001:db8:a0b:12f0::1"
    }

This will mask both IPV4 and IPV6 IP addresses.  IPV6 can be left out.  If only IPV6 is required, IPV4 can be an empty string and will not be masked.

dmke:
Are the quotation marks nececcary?

Also using `2001:db8:a0b:12f0::1` as sample mask would be confusing, as it would mask seemlingly random bits of the address. It shoud be made clear that the first *n* bits of the mask should be 1, and the remaining bits be 0.

(This is also why I originally suggested the *number of bits* to be configurable, not the actual bitmask, i.e. `ipmask 16 72` → `255.255.0.0` + `ffff:ffff:ffff:ffff:ff00::`.)

tobya:
Quotes probably are not necessary. I’ll update ip6 mask also. Thanks.

caddyhttp/log/log.go
```diff
@@ -66,6 +67,14 @@ func (l Logger) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 
 			// Write log entries
 			for _, e := range rule.Entries {
+				if e.Log.IPMaskExists {
+					// Mask IP Address, will not modify if default
+					hostip, hostport, err := net.SplitHostPort(r.RemoteAddr)
+					if err == nil {
+						r.RemoteAddr = net.JoinHostPort(e.Log.MaskIP(hostip), hostport)
```
tw4452852: Just a concern how to get the real `RemoteAddr` if we replace here.

tw4452852: Just a concern how to get the real `RemoteAddr` later if we replace it here.

tobya: I made an assumption, very possibly wrong that Log is the last thing to be called.  If not I can look at a way around this.

mholt: I think it's unlikely that log is the last middleware in the chain.

tobya:
Perhaps there may be a straightforward fix for @tw4452852 concern once #1937 is merged.

tobya:
Finally got around to updating this.  I simply set the replacer custom value for 'remote' if an ipmask exists.  This will ensure that the custom value is used instead of the code in the switch statement.  If no ipmask exists the code executes in replacer as normal.

tobya:
@mholt The documentation changes below should be added just below the 'Log Rolling' section on log page.  I'm not sure how to do a diff without access to the source for the /doc pages.

IP Masking 
-----

IP addresses in log can be masked to provide annonymisation for corporate or legal reasons.

    log / access.log "{common}" {
      ipmask 255.255.0.0 ffff:ffff:ffff:ff00::
    }

This will mask both IPV4 and IPV6 IP addresses.  IPV6 can be left out.  If only IPV6 is required, IPV4 can be an empty string and will not be masked.

tobya:
@tw4452852 Hopefully this update resolves your concern.

tw4452852:
LGTM

PR #1931, opened 2017-10-22
make log directive replacer available on request context

kujenga:
This change allows custom log values to be made available as placeholders while preventing other directives from breaking that functionality.

This issue I ran into was that the currently available method of retrieving the `httpserver.Replacer`, with a type assertion on the `http.ResponseWriter`, as is dine in the snippet here, breaks down in conjunction with other plugins that wrap the response writer, such as `header`.

https://github.com/mholt/caddy/blob/59a8ada4a8d287abd7e1025293f5d7894a7788c2/caddyhttp/log/log_test.go#L31-L33

The solution in-flight in this PR https://github.com/mholt/caddy/pull/1542 allows for a custom `{user}` placeholder to be logged, but does not facilitate arbitrary custom logs.

This change allows the log replacer to be retrieved at any point, when present.

There were a few issues I saw referencing this class of problem, but none of the proposed or in-flight solutions fit the use case I was experimenting with.

- https://caddy.community/t/custom-user-placeholder-along-with-header-directive/1676/9
- https://github.com/mholt/caddy/issues/1543

Since this is for plugin authors, the godoc from this added method is likely sufficient for minimal documentation. However, if there are other places to document this type of functionality, I'd be happy to update them.

### 4. Checklist

### DIFF
--- a/caddyhttp/log/log.go
+++ b/caddyhttp/log/log.go
@@ -16,6 +16,7 @@
 package log
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 
@@ -28,6 +29,19 @@ func init() {
 		ServerType: "http",
 		Action:     setup,
 	})
+}
+
+type logReplacerKey struct{}
+
+var logReplacerCtxKey logReplacerKey
+
+// GetReplacer returns the log replacer for the given request, when present.
+// If no log directive in the configuration applied to this request, then the
+// returned value is nil. This method allows plugins to robustly make
+// placeholder values available to the logging directive.
+func GetReplacer(r *http.Request) httpserver.Replacer {
+	replacer, _ := r.Context().Value(logReplacerCtxKey).(httpserver.Replacer)
+	return replacer
 }
 
 // Logger is a basic request logging middleware.
@@ -47,6 +61,8 @@ func (l Logger) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
 			// set their own placeholders if they want to.
 			rep := httpserver.NewReplacer(r, responseRecorder, CommonLogEmptyValue)
 			responseRecorder.Replacer = rep
+
+			r = r.WithContext(context.WithValue(r.Context(), logReplacerCtxKey, rep))
 
 			// Bon voyage, request!
 			status, err := l.Next.ServeHTTP(responseRecorder, r)
--- a/caddyhttp/log/log_test.go
+++ b/caddyhttp/log/log_test.go
@@ -31,6 +31,9 @@ func (erroringMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) (int
 	if rr, ok := w.(*httpserver.ResponseRecorder); ok {
 		rr.Replacer.Set("testval", "foobar")
 	}
+	if repl := GetReplacer(r); repl != nil {
+		repl.Set("customval", "barbaz")
+	}
 	return http.StatusNotFound, nil
 }
 
@@ -40,7 +43,7 @@ func TestLoggedStatus(t *testing.T) {
 	rule := Rule{
 		PathScope: "/",
 		Entries: []*Entry{{
-			Format: DefaultLogFormat + " {testval}",
+			Format: DefaultLogFormat + " {testval} {customval}",
 			Log:    httpserver.NewTestLogger(&f),
 		}},
 	}
@@ -74,6 +77,10 @@ func TestLoggedStatus(t *testing.T) {
 	// check custom placeholder
 	if !strings.Contains(logged, "foobar") {
 		t.Errorf("Expected the log entry to contain 'foobar' (custom placeholder), but it didn't: %s", logged)
+	}
+	// check contains other placeholder
+	if !strings.Contains(logged, "barbaz") {
+		t.Errorf("Expected the log entry to contain 'barbaz' (custom placeholder from context), but it didn't: %s", logged)
 	}
 }
 


tobya:
Thanks for the PR.   This has been an issue for a while.

caddyhttp/log/log.go
```diff
@@ -30,6 +31,19 @@ func init() {
 	})
 }
 
+type logReplacerKey struct{}
+
+var ctxKey logReplacerKey
```
tobya: Can you change the name `ctxKey` here to something like logReplacerCtxKey to make it consistent with other context keys.

tobya:
Sorry for my ignorance here but would you be able to show how this can be used for example to store the user in basicauth, I'm not clear how I can call the `GetReplacer` function?  Perhpas you could give an example of calling it in arbitrary code?

caddyhttp/log/log.go
tobya: actually I think we should create a const like the other context keys eg from middleware.go

    const (
         RemoteUserCtxKey caddy.CtxKey = "remote_user"

kujenga: @tobya thanks for the review!

I'm happy to make these these changes, but I'm curious what the utility of the shared `caddy.CtxKey` type is (I'm fairly new to the Caddy codebase). The utility I see in the private type is that it prevents any unintentional collisions between packages, which one can imagine resulting in confusing bugs in the general case.

kujenga: I amended my commit to use `logReplacerCtxKey` as the variable name. If you'd like me to make the switch to the shared context key type as well, just let me know, happy to adhere to the existing conventions.

tobya: Thanks for making change. I’m fine with your explanation unless others disagree. We’ll
Need to do some more testing to make sure this change doesn’t introduce any unforeseen issues, but on the face of it this looks like a great solution

mholt:
Hey, thanks for the contribution! Been looking to do this for a while, as Toby said (see #1542) -- I do want to think about the implications of using a private type; what's wrong with using caddy.CtxKey?

kujenga:
Sure thing! The private type pattern is derived from the standard library documentation, which I've copied in below. The reason I used it here was that I didn't see a use case for overriding this context value outside of the `log` package. If you prefer the flexibility provided by the publicly declared `caddy.CtxKey`, I'm happy to make that switch.

```
	// 	// Package user defines a User type that's stored in Contexts.
	// 	package user
	//
	// 	import "context"
	//
	// 	// User is the type of value stored in the Contexts.
	// 	type User struct {...}
	//
	// 	// key is an unexported type for keys defined in this package.
	// 	// This prevents collisions with keys defined in other packages.
	// 	type key int
	//
	// 	// userKey is the key for user.User values in Contexts. It is
	// 	// unexported; clients use user.NewContext and user.FromContext
	// 	// instead of using this key directly.
	// 	var userKey key
	//
	// 	// NewContext returns a new Context that carries value u.
	// 	func NewContext(ctx context.Context, u *User) context.Context {
	// 		return context.WithValue(ctx, userKey, u)
	// 	}
	//
	// 	// FromContext returns the User value stored in ctx, if any.
	// 	func FromContext(ctx context.Context) (*User, bool) {
	// 		u, ok := ctx.Value(userKey).(*User)
	// 		return u, ok
	// 	}
```

https://github.com/golang/go/blob/006bc57095e7b1221ecc292735a510a52a27292c/src/context/context.go#L124-L150

kujenga:
I also found one existing place in the codebase where this improvement could be leveraged, in the `proxy` directive where it sets the `"upstream"` placeholder: https://github.com/mholt/caddy/blob/fc75527eb5ea9d0252bb3079a0137dbbfb754790/caddyhttp/proxy/proxy.go#L178-L180

Would it make sense to update that to use the new getter this PR provides?

tobya:
I think it would make sense to make the change in the proxy.go file.  If this is going to be merged then this is the sort of use.  This also provides an example of its use in the core code.

tobya:
I think I might have been wrong the use of the Replacer in proxy is seperate to the use of the replacer in Log so they would need to be 2 seperate instances.  This I think i why the test is failing.  Lets keep testing this.  For the moment I think you should rollback the change perhaps

kujenga:
Tt looks like that test is specifically creating it's own "replacer" and not using the log package, which is why the test is failing. What is the intent/desire for the functionality of custom values in replacers? I see that placeholders are available for use in several directives, are they all just leveraging the context values for most things, with spotty support for custom values?

kujenga:
I've removed that commit for now. Here is the diff from it for future context:

```patch
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -26,6 +26,7 @@ import (
        "time"
 
+       "github.com/mholt/caddy/caddyhttp/log"
 )
 
 // Proxy represents a middleware instance that can proxy requests.
@@ -175,8 +176,8 @@ func (p Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
                        }
                        continue
                }
-               if rr, ok := w.(*httpserver.ResponseRecorder); ok && rr.Replacer != nil {
-                       rr.Replacer.Set("upstream", host.Name)
+               if replacer := log.GetReplacer(r); replacer != nil {
+                       replacer.Set("upstream", host.Name)
                }
 
                proxy := host.ReverseProxy
```

mholt:
For some added context (pun intended): if every request has a Replacer added to the request's Context, I can complete the new `if` directive which allows really powerful request matching for any other directive, rather than just base path or having to use rewrite hacks.

kujenga:
Neat! I'll try to take a stab at it.

kujenga:
I think I've figured out a nice way to handle this without needing to change the existing APIs, and have it pushed to a branch here: https://github.com/mholt/caddy/compare/master...kujenga:per-request-replacer All existing tests pass, but I still need to figure out what the best way to test it is. Let me know if you have thoughts!

tobya:
@kujenga Would you be able to add it as a new PR so we can test and comment.  We can discuss between the two options.

kujenga:
Closing in favor of https://github.com/mholt/caddy/pull/1937

Issue #1933, opened 2017-10-22
On Hook runs once for each host rather than once for each block


tobya:
### 1. What version of Caddy are you using (`caddy -version`)?

Latest Dev build 

### 2. What are you trying to do?

Run shutdown script once!

### 3. What is your entire Caddyfile?
```text
courgette.kinchik.ie:80 pepper.kinchik.ie:80 {
  log / d:\caddy\logs\courgettelog.log   {common} 

  root d:\Development\websites\bcsstudents\Attached
  browse  

  on startup d:\php7\php-cgi.exe -b 34345  &
  on shutdown d:\caddy\shutdown.bat 
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
caddy -log stdout

### 6. What did you expect to see?

Startup and shutdown scripts run once 

### 7. What did you see instead (give full error messages and/or log)?

Startup and shutdown scripts run twice.
### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
use caddyfile included.

----

I'm not sure if this was part of the discussion for  #1864 but prior to the `on` hook being merged the `startup` and `shutdown` commands were run once per block - or as @mholt  puts it in the documentation once for each time it appears in the caddyfile.  

Now they are run for each host so with the above caddyfile the startup and shutdown scripts are run twice rather than once.

PR #1934, opened 2017-10-22
onevent/startupshutdown: run command once per server block

elcore:

Execute onevent commands once per server block

#1933 

n/a

### 4. Checklist

### DIFF
--- a/onevent/on.go
+++ b/onevent/on.go
@@ -20,9 +20,12 @@ func setup(c *caddy.Controller) error {
 	}
 
 	// Register Event Hooks.
-	for _, cfg := range config {
-		caddy.RegisterEventHook("on-"+cfg.ID, cfg.Hook)
-	}
+	c.OncePerServerBlock(func() error {
+		for _, cfg := range config {
+			caddy.RegisterEventHook("on-"+cfg.ID, cfg.Hook)
+		}
+		return nil
+	})
 
 	return nil
 }
--- a/startupshutdown/startupshutdown.go
+++ b/startupshutdown/startupshutdown.go
@@ -36,9 +36,12 @@ func Startup(c *caddy.Controller) error {
 	}
 
 	// Register Event Hooks.
-	for _, cfg := range config {
-		caddy.RegisterEventHook("on-"+cfg.ID, cfg.Hook)
-	}
+	c.OncePerServerBlock(func() error {
+		for _, cfg := range config {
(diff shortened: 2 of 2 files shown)


elcore:
> This seems fine. I pulled and built it and it seems to fix the issue. I guess we might eventually get on events that should be executed for each host, but perhaps we can wait until that happens before worry about it.

👍

tobya:
I'm not sure if this is related, but i get odd log entries .  Here is the log I tested 

https://gist.github.com/tobya/6254f1fd406994cc3f8621d339f26f06

If you take a look you'll see interspersed in my .bat batch file that runs on shutdown is startup log lines, these come from caddy not from my script.  The script I have just writes the date to a file.

mholt:
Don't batch scripts print their commands as they're executed? `@ECHO off` or something is needed? (I haven't programmed .bat in like 10 years...)

elcore:
> I'm not sure if this is related, but i get odd log entries . Here is the log I tested

I do not think this is related, `REM Should run once` sounds like an issue with your batch script

Reference: `error on 'on-fa727352-e277-49bf-b980-6771e9c37a00' hook`

tobya:
No sorry its a bit confusing, the issue is not that my bat script is echoing nonsense (it doesnt do anything its just a test) its that caddy is for somereason putting the startup message in the output as though the on startup scripts are being run again.

> 2017/10/22 22:18:50 [INFO] Nonblocking Command "d:\php7\php-cgi.exe -b 49209" with ID e3bb5110-d7cb-42a9-b5b0-2ebbebd6a578

elcore:
`2017/10/22 22:18:50 [INFO] Nonblocking Command "d:\php7\php-cgi.exe -b 49209" with ID e3bb5110-d7cb-42a9-b5b0-2ebbebd6a578` is the normal log output

https://github.com/mholt/caddy/blob/e5d04f9a96c3beb74f350688feba0d5003a22bea/onevent/hook/hook.go#L31

> Caddy is for somereason putting the startup message in the output as though the on startup scripts are being run again.

I am not familiar with Windows batch scripts, Caddy logs the message before its execution. Have you tried `@ECHO off` as suggested by @mholt

tobya:
I tracked it, nothing to do with batch scripts or this change!  I'll take the discussion offline.

elcore:
> I tracked it, nothing to do with batch scripts or this change! I'll take the discussion offline.

Great 😄. What was/is the issue

Issue #1543, opened 2017-03-28
Unable to set BasicAuth username placeholder for access log, header directive overrides


jamesboswell:
### 1. What version of Caddy are you running (`caddy -version`)?
0.9.5  (fbd6412359be76bcfbb2ba5cfdecbc273e040674)

### 2. What are you trying to do?
Log BasicAuth username to access.log with custom {user} placeholder

### 3. What is your entire Caddyfile?
```text
localhost {
  tls off

  log / /tmp/access.log "{remote} - {user} [{when}] \"{method} {uri} {proto}\" {status} {size} \"{>Referer}\" \"{>User-Agent}\""
  errors /tmp/error.log

  radiusauth {
          server 127.0.0.1:1812 127.0.0.1:9012
          secret testing123
          realm  "RADIUS Auth"
          except /public /img
          cache /tmp
          cachetimeout 60
          }

  root /var/www/hugo

  fastcgi / 127.0.0.1:9000 php

  # Hardening headers
  header / {
    X-Content-Type-Options "nosniff"
}

}

localhost/img {
  tls off
  basicauth / bob hello
  log / /tmp/access.log "{remote} - {user} [{when}] \"{method} {uri} {proto}\" {status} {size} \"{>Referer}\" \"{>User-Agent}\""
  errors /tmp/error.log
  root /data/tmp/img/

}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
Testing local checkout
`go run main.go -conf ~jboswell/caddyfile3`

### 5. Please paste any relevant HTTP request(s) here.
```
curl -I -u bob:hello http://localhost:2015
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 7363
Content-Type: text/html; charset=utf-8
Etag: "nqf40l5oj"
Last-Modified: Tue, 23 Jun 2015 22:05:09 GMT
Server: Caddy
X-Content-Type-Options: nosniff
X-Xss-Protection: 1; mode=block
Date: Tue, 28 Mar 2017 22:28:03 GMT
```
### 6. What did you expect to see?
Expect to see "bob" in the access log,  instead see  "-"

### 7. What did you see instead (give full error messages and/or log)?
```
::1 - - [28/Mar/2017:18:28:03 -0400] "HEAD / HTTP/1.1" 200 0 "-" "curl/7.53.1"
```

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
[...]

PR #1937, opened 2017-10-29
Setup a Replacer and ResponseRecorder per-request, making interactions with placeholders more coherent

kujenga:

This PR sets up a `Replacer` and `ResponseRecorder` object per-request, so that the functions that "create" them preferentially retrieve them from the request itself before creating a new one. This allows directives throughout the middleware chain to leverage shared information collected over the course of the request.

I have tested this locally using a custom plugin that adds custom placeholders, in combination with the `headers` directive, which were previously in conflict, causing the placeholders to not be available in the `log ` directive, and that now works as expected!

Tests here still need to be added.

Follows on from a comment on a previous PR attacking a manifestation of this underlying issue: https://github.com/mholt/caddy/pull/1931#pullrequestreview-71723741

Previous PR: https://github.com/mholt/caddy/pull/1931

Issues linked to from previous PR:
- https://caddy.community/t/custom-user-placeholder-along-with-header-directive/1676/9
- #1543

There is a slight behavioral change if plugin implementers were relying on the previous under the hood behavior of these where separate replacers/response recorders could be created, though it's hard to imagine a use case for that.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -102,20 +102,30 @@ func (lw *limitWriter) String() string {
 // emptyValue should be the string that is used in place
 // of empty string (can still be empty string).
 func NewReplacer(r *http.Request, rr *ResponseRecorder, emptyValue string) Replacer {
-	rb := newLimitWriter(MaxLogBodySize)
-	if r.Body != nil {
-		r.Body = struct {
-			io.Reader
-			io.Closer
-		}{io.TeeReader(r.Body, rb), io.Closer(r.Body)}
+	repl := &replacer{
+		request:          r,
+		responseRecorder: rr,
+		emptyValue:       emptyValue,
 	}
-	return &replacer{
-		request:            r,
-		requestBody:        rb,
-		responseRecorder:   rr,
-		customReplacements: make(map[string]string),
-		emptyValue:         emptyValue,
+
+	// extract customReplacements from a request replacer when present.
+	if existing, ok := r.Context().Value(ReplacerCtxKey).(*replacer); ok {
+		repl.requestBody = existing.requestBody
+		repl.customReplacements = existing.customReplacements
+	} else {
+		// if there is no existing replacer, build one from scratch.
+		rb := newLimitWriter(MaxLogBodySize)
+		if r.Body != nil {
+			r.Body = struct {
+				io.Reader
+				io.Closer
+			}{io.TeeReader(r.Body, rb), io.Closer(r.Body)}
+		}
+		repl.requestBody = rb
+		repl.customReplacements = make(map[string]string)
 	}
+
+	return repl
 }
 
 func canLogRequest(r *http.Request) bool {
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -356,6 +356,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	c := context.WithValue(r.Context(), OriginalURLCtxKey, urlCopy)
 	r = r.WithContext(c)
 
+	// Setup a replacer for the request that keeps track of placeholder
+	// values across plugins.
+	replacer := NewReplacer(r, nil, "")
+	c = context.WithValue(r.Context(), ReplacerCtxKey, replacer)
+	r = r.WithContext(c)
+
 	w.Header().Set("Server", caddy.AppName)
 
 	status, _ := s.serveHTTP(w, r)
--- a/caddyhttp/httpserver/middleware.go
+++ b/caddyhttp/httpserver/middleware.go
@@ -214,6 +214,9 @@ func SameNext(next1, next2 Handler) bool {
 
 // Context key constants.
 const (
+	// ReplacerCtxKey is the context key for a per-request replacer.
+	ReplacerCtxKey caddy.CtxKey = "replacer"
+
 	// RemoteUserCtxKey is the key for the remote user of the request, if any (basicauth).
 	RemoteUserCtxKey caddy.CtxKey = "remote_user"
 


mholt:
Yesss, I am looking forward to reviewing this soon!! Thank you!

tobya:
I can confirm that this change allows me to set the basicauth use to allow it to be written to logs to replace `{user}`

			repl := httpserver.NewReplacer(r, nil, "-")
			repl.Set("user", username)

good work, thanks.

caddyhttp/httpserver/replacer.go
```diff
@@ -41,6 +41,9 @@ var requestReplacer = strings.NewReplacer(
 
 var now = time.Now
 
+// ReplacerCtxKey is the context key for a per-request replacer.
+var ReplacerCtxKey = (caddy.CtxKey)("replacer")
```
mholt: This syntax is interesting - did you mean `caddy.CtxKey("replacer")`?

caddyhttp/httpserver/server.go
```diff
@@ -356,6 +356,16 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	c := context.WithValue(r.Context(), OriginalURLCtxKey, urlCopy)
 	r = r.WithContext(c)
 
+	// Setup a response recorder for this request.
+	rr := NewResponseRecorder(w)
+	w = rr
```
mholt: I am not confident that this will have no impact on performance. A ResponseRecorder is necessary for implementing the response placeholders, but these are not always needed. Can we only attach the ResponseRecorder in middlewares that need it?

caddyhttp/httpserver/server.go
```diff
@@ -356,6 +356,16 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	// Setup a replacer for the request that keeps track of placeholder
+	// values across plugins.
+	replacer := NewReplacer(r, rr, "")
```
mholt: The `log` directive (for example) uses an empty value of `"-"`. How can we make this something customizable per-middleware?

caddyhttp/httpserver/replacer.go
```diff
@@ -102,20 +105,33 @@ func (lw *limitWriter) String() string {
...
-	return &replacer{
-		request:            r,
-		requestBody:        rb,
-		responseRecorder:   rr,
-		customReplacements: make(map[string]string),
-		emptyValue:         emptyValue,
+
+	// extract customReplacements from a request replacer when present.
+	existing, ok := r.Context().Value(ReplacerCtxKey).(*replacer)
+	if ok {
```
mholt: This could be wrapped up in a `if existing, ok := r.Context().Value(ReplacerCtxKey).(*replacer); ok {` to keep the variables contained in that scope.

caddyhttp/httpserver/recorder.go
```diff
@@ -47,6 +47,9 @@ type ResponseRecorder struct {
 // explicitly, this constructor initializes with a status code
 // of 200 to cover the default case.
 func NewResponseRecorder(w http.ResponseWriter) *ResponseRecorder {
+	if rr, ok := w.(*ResponseRecorder); ok {
+		return rr
+	}
```
mholt: I do wonder what kinds of implications there are of sharing a single ResponseRecorder across all middlewares...

caddyhttp/httpserver/replacer.go
tobya: I would suggest here that we should create a new const in middleware.go with the other context keys.

caddyhttp/httpserver/recorder.go
tobya: Would it be worth providing a test binary with this change in it that people could test out to give us some early indication if it is likely to cause any issues.  I'm not sure there is an answer to this question without extensive testing.

caddyhttp/httpserver/replacer.go
kujenga: @mholt I've come to like the casting syntax above because it makes the case visually distinct from a function call, but I realize it's unusual. Moving it to middleware.go will avoid the issue entirely, since those keys declare the type to the left of the equals 🙂

caddyhttp/httpserver/server.go
kujenga: In the body of `NewReplacer`, I maintain some functions from the replacer for the request, but not all. The `emptyValue` is one of the things that gets set every time, so I'm pretty sure that this functionality it preserved. I'll see if a test can be crafted for that.

caddyhttp/httpserver/server.go
kujenga: Hmm, yeah, that's a fair point. The reason I set this up globally was that the replacer takes a  recorder as a parameter, and I couldn't seem to find a good way to do it ad-hoc within the middleware chain, since it requires re-defining the `*http.Request` being propagated. I'll take another pass at this to see if there are other options.

kujenga:
Okay, I believe that the the concerns raised have been addressed! I was able to remove everything having to do with customizing the `ResponseRecorder` behavior, I must have first introduced all that for reasons that turned out to be a red herring.

caddyhttp/httpserver/server.go
```diff
@@ -356,6 +356,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	c := context.WithValue(r.Context(), OriginalURLCtxKey, urlCopy)
 	r = r.WithContext(c)
 
+	// Setup a replacer for the request that keeps track of placeholder
+	// values across plugins.
+	replacer := NewReplacer(r, nil, "")
```
mholt: Wanted to double-check: does the `log` middleware still set its own "empty" value (should default to `-` at least for the default log format)?

kujenga: Yep! That's correct. Every call to `NewReplacer` it a partial clone of the shared one that sets it's own `emptyValue`: https://github.com/kujenga/caddy/blob/per-request-replacer/caddyhttp/httpserver/replacer.go#L108

The log package still does that for each request: https://github.com/kujenga/caddy/blob/per-request-replacer/caddyhttp/log/log.go#L48

mholt: Okay, cool. So isn't the next step to get the middlewares to pull the Replacer off the context instead of making their own? (I guess your code helps by using the context's one, but I imagine it'd be "cleaner" if they just pulled from the context directly?)

mholt:
Thank you for this contribution, @kujenga! Am I correct in understanding that the best way to access and use this Replacer is to modify the middlewares to get it from the Context directly? Or should we leave them the same.

kujenga:
@mholt delighted to assist!

I think leaving them as-is makes sense because of the issue you raised about customizing fields like `emptyValue`. In the future it might make sense to separate out a new type, having something like a `ReplacerValues` that gets propagated through the request context, and the `Replacer` would be a thin wrapper over that which handles rendering specific to the middleware? That would also allow the existing API of the `Replacer` type to remain the same. Just a thought.

kujenga:
Also, any idea when the next release will be tagged?

mholt:
Interesting idea, we can pursue later, but probably not expedient right now. As for next release, I'm not sure when; just follow the 0.10.11 milestone!

Issue #1847, opened 2017-08-29
Check with backend if should do On-Demand TLS


francislavoie:
I had a "showerthought" this morning, a idea to potentially make on-demand TLS more viable for some SaaS modeled applications. Disclaimer: I personally don't have a use for this yet, but it might make it easier for when I do!

So with `proxy`, we have the `health_check` feature. I was thinking we could leverage the same kind of idea with `tls` to do a request to an internal API that the application would need to expose that would let Caddy ask the app if it "should" attempt on-demand TLS. That way Caddy would only make certs for domains that actually make sense to the application.

The only thing we have as a limiter right now (as far as I can tell?) is `max_certs`, which is a bit too hard of a limit if there's legitimate subdomains that should exist. Specifically, I'm thinking of the `{customername}.example.com` model specifically here.

I'm not totally sure what the API would look like, but it could be as simple as a GET to the endpoint specified with domain as a URL parameter like `/api/domains?domain={subdomain}` and the app should return a `2xx` as a go-ahead or `4xx` as a no-go.

I guess this might go along with https://github.com/mholt/caddy/issues/8 a bit, but in this case it would be a request that Caddy makes to reach out to the internally proxied/fastcgi'd application like with `health_check`.

mholt:
I like this idea. Caddy should have a way of determining whether it's authorized to obtain a certificate for a certain hostname, rather than just "have we hit the limit yet?" - I think ultimately this is what I wanted to do, but I got lazy at first.

mholt:
I envision a subdirective to `tls` that lets you set a URL where Caddy can ask whether a hostname should grant the certificate: `ask localhost:1234/on-demand` which then sends a request with the hostname. Something like that.

tantalic:
@mholt  I would be willing to work on this. A couple of questions/clarifications for how you would like to see this:

1. Should the request be a GET or POST? Since we are retrieving informaiton I propose a GET request.
2. Should the request include any information other then the domain being requested?
3. I propose any `2xx` response would be considered an approval and any other response code would be considered a rejection. The response body would not be considered. Do you agree?
4. Should we follow `3xx` redirects? At this time I would suggest we do not follow redirects. I believe redirects may provide an avenue for abuse and unnecessarily complicate the feature.

mholt:
@tantalic 

1. I agree: `GET` - just no request body
2. No, probably not. (We can change this later if there is an actual need)
3. Yes, let's keep it as simple as possible!
[...]

PR #1939, opened 2017-10-31
Add option for a backend check to approve use of on-demand TLS

tantalic:
This adds the ask sub-directive to tls that defines the URL of a backend HTTP service to be queried during the TLS handshake to determine if an on-demand TLS certificate should be acquired for incoming hostnames. When the ask sub-directive is defined, Caddy will query the URL for permission to acquire a cert by making a HTTP GET request to the URL including the requested domain in the query string. If the backend service returns a 2xx response Caddy will acquire a cert. Any other response code (including 3xx redirects) are be considered a rejection and the certificate will not be acquired.

#1847

The TLS docs will need a reference to the `ask` directive and some updated language around on-demand TLS in general. 

### 4. Checklist

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -148,6 +148,11 @@ type OnDemandState struct {
 	// Set from max_certs in tls config, it specifies the
 	// maximum number of certificates that can be issued.
 	MaxObtain int32
+
+	// The url to call to check if an on-demand tls certificate should
+	// be issued. If a request to the URL fails or returns a non 2xx
+	// status on-demand issuances must fail.
+	AskURL *url.URL
 }
 
 // ObtainCert obtains a certificate for name using c, as long
--- a/caddytls/handshake.go
+++ b/caddytls/handshake.go
@@ -19,6 +19,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -135,8 +137,8 @@ func (cfg *Config) getCertDuringHandshake(name string, loadIfNecessary, obtainIf
 
 			name = strings.ToLower(name)
 
-			// Make sure aren't over any applicable limits
-			err := cfg.checkLimitsForObtainingNewCerts(name)
+			// Make sure the certificate should be obtained based on config
+			err := cfg.checkIfCertShouldBeObtained(name)
 			if err != nil {
 				return Certificate{}, err
 			}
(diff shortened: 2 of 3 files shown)


caddytls/handshake.go
```diff
@@ -159,10 +161,49 @@ func (cfg *Config) getCertDuringHandshake(name string, loadIfNecessary, obtainIf
...
+		return cfg.checkURLForObtainingNewCerts(name)
+	}
+
+	// Otherwise use the limit defined by the "max_certs" setting
+	return cfg.checkLimitsForObtainingNewCerts(name)
+}
+
+func (cfg *Config) checkURLForObtainingNewCerts(name string) error {
+	client := http.Client{
+		Timeout: 10 * time.Second,
```
tantalic: I set a timeout of 10s here. Since this should be a backend service that should be able to respond quickly and it will be holding open the TLS negotiation I think we could drop this quite a bit lower as well. It could be configurable as well, but that seems unnecessary to me at this time.

caddytls/setup.go
```diff
@@ -164,6 +164,9 @@ func setupTLS(c *caddy.Controller) error {
 			case "max_certs":
 				c.Args(&maxCerts)
 				config.OnDemand = true
+			case "ask":
+				c.Args(&askURL)
```
mholt: See if you can parse the URL here, and if it fails, return an error. This will help prevent configuration errors after the server has already started...

caddytls/handshake.go
```diff
@@ -159,10 +161,49 @@ func (cfg *Config) getCertDuringHandshake(name string, loadIfNecessary, obtainIf
...
+	query.Add("domain", name)
+	checkURL.RawQuery = query.Encode()
+
+	resp, err := client.Get(checkURL.String())
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return fmt.Errorf("Non-200 status code returned (%d)", resp.StatusCode)
```
mholt: Start the error string with a lower-case letter, since it is usually used in the middle of a sentence when printing to the logs. Also, be a little more descriptive about the implications of this. For example: "certificate for hostname '%s' not allowed, non-2xx status code %d returned"

caddytls/config.go
```diff
@@ -148,6 +148,11 @@ type OnDemandState struct {
 	// Set from max_certs in tls config, it specifies the
 	// maximum number of certificates that can be issued.
 	MaxObtain int32
+
+	// The url to call to check if an on-demand tls certificate should
+	// be issued. If a request to the URL fails or returns a non 2xx
+	// status on-demand issuances must fail.
+	CheckUrl string
```
mholt: "CheckURL" or, for consistency with the Caddyfile, "AskURL" -- or for more descriptive: "OnDemandAskURL" or something.

caddytls/setup.go
tantalic: See line 220 in the latest revision which validates the URL can be parsed and the scheme is either HTTP or HTTPS.

tantalic:
@mholt thanks for the review, this now includes the improvements requested.

caddytls/handshake.go
```diff
@@ -159,10 +161,48 @@ func (cfg *Config) getCertDuringHandshake(name string, loadIfNecessary, obtainIf
...
+
+	query := askURL.Query()
+	query.Set("domain", name)
+	askURL.RawQuery = query.Encode()
+
+	resp, err := client.Get(askURL.String())
+	if err != nil {
+		return err
+	}
+
```
mholt: The response body must be closed:

```go
defer resp.Body.Close()
```

tobya:
Is there any benefit of thinking about this as an event for on? Eg

    On tls_anyhost https://example.com

Just an idea

tantalic:
@mholt this has been updated as follows:

- Added the missing `resp.Body.Close()` as requested
- Updated the backend HTTP request so that it doesn't follow 3xx redirects (`handshake.go` lines 181-183), this matches the discussion in  #1847

mholt:
LGTM. Thank you!!

PR #1946, opened 2017-11-06
Checking that a server listener is not nil before printing verbose in…

insomniacslk:

It adds a check to make sure that a non-nil server listener is used before printing its file descriptor notice in non-quiet mode.
Without this patch, returning a nil listener alongside with a nil error from `Server.Listener` and `Server.PacketListener` would cause a segmentation fault.

In the plugins documentation[1] it's stated that
```
If your server only uses TCP, the *Packet() methods may be no-ops (i.e. they return nil).
The inverse is true for non-TCP servers. A server may also use both TCP and UDP, and
implement all four methods.
```

However the function `startWithListenerFds` (before this patch) would dereference `srvln.listener` in non-quiet mode in any case.

[1] https://github.com/mholt/caddy/wiki/Writing-a-Plugin:-Server-Type

no issue associated to this PR

None, this should more closely match the existing documentation

### 4. Checklist

- no new test has been written. There is currently no other test for `startWithListenerFds`
- I am willing to help maintain this change if there are issues with it later

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -518,9 +518,13 @@ func startWithListenerFds(cdyfile Input, inst *Instance, restartFds map[string]r
 		}
 		if !Quiet {
 			for _, srvln := range inst.servers {
-				if !IsLoopback(srvln.listener.Addr().String()) {
-					checkFdlimit()
-					break
+				// only show FD notice if the listener is not nil.
+				// This can happen when only serving UDP or TCP
+				if srvln.listener != nil {
+					if !IsLoopback(srvln.listener.Addr().String()) {
+						checkFdlimit()
+						break
+					}
 				}
 			}
 		}
--- a/caddy.go
+++ b/caddy.go
@@ -520,11 +520,12 @@ func startWithListenerFds(cdyfile Input, inst *Instance, restartFds map[string]r
 			for _, srvln := range inst.servers {
 				// only show FD notice if the listener is not nil.
 				// This can happen when only serving UDP or TCP
-				if srvln.listener != nil {
-					if !IsLoopback(srvln.listener.Addr().String()) {
-						checkFdlimit()
-						break
-					}
+				if srvln.listener == nil {
+					continue
+				}
+				if !IsLoopback(srvln.listener.Addr().String()) {
+					checkFdlimit()
+					break
 				}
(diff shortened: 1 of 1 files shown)


caddy.go
```diff
@@ -518,9 +518,13 @@ func startWithListenerFds(cdyfile Input, inst *Instance, restartFds map[string]r
 		}
 		if !Quiet {
 			for _, srvln := range inst.servers {
-				if !IsLoopback(srvln.listener.Addr().String()) {
-					checkFdlimit()
-					break
+				// only show FD notice if the listener is not nil.
+				// This can happen when only serving UDP or TCP
+				if srvln.listener != nil {
```
mholt: Instead of adding a layer of nesting, how about using `if srvln.listener == nil { continue }` (with line breaks of course). Just reads a little cleaner, and I prefer it.

insomniacslk: yes, much better! New commit coming shortly

mholt:
Thanks, LGTM (the change doesn't hurt anyway). Just curious since there aren't tests, did you see a panic and confirm that this eliminates it?

insomniacslk:
Yes, I got the following:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x4f406a]

goroutine 1 [running]:
github.com/mholt/caddy.startWithListenerFds(0x5da540, 0xc4200516c0, 0xc42006eea0, 0x0, 0x4051bc, 0xc42006a058)
	/home/insomniac/go/src/github.com/mholt/caddy/caddy.go:521 +0x36a
github.com/mholt/caddy.Start(0x5da540, 0xc4200516c0, 0x5da540, 0xc4200516c0, 0x0)
	/home/insomniac/go/src/github.com/mholt/caddy/caddy.go:459 +0xdf
main.main()
	/home/insomniac/test/main.go:66 +0xd1
```

when the custom-defined `Server.Listen` was returning `nil, nil` to make it work as UDP-only

insomniacslk:
Also, setting `caddy.Quiet = true` before this patch would make the segfault disappear. After the patch, it won't crash (and I see the daemon correctly serving only over UDP) even when `caddy.Quiet` is set to false

Issue #1862, opened 2017-09-09
Improve Caddy path matching


dhaavi:
I am currently using Caddy on multiple production server in different settings. So far I really love the simplicity, speed and stability of Caddy. In the last weeks I have also experienced two limits of Caddy. Both of them are, I think, very important to discuss, as they may become a dealbreaker easily. In order to secure the future Caddy (you want companies to buy [these plans](https://caddyserver.com/pricing), right?), these issues should be thoroughly discussed - as they are also breaking changes. 

The first issue is about path matching and has also been partly discussed in #167. If there are other similar issues, or this is already currently being discussed in another issue (I couldn't find anything current), please tell me.

This is not beautiful:

    www.example.com {
      proxy /en 10.0.0.1:80
      redir 301 {
        if {path} not_starts_with /en
        / www.example.org
      }
    }

This is:

    www.example.com {
      proxy /en 10.0.0.1:80
      redir / www.example.org
    }

This is just a small example, the underlying issue is much more important and complex. How I experienced Caddy, plugins are asked sequentially if they handle a certain path, if they do, the path matching phase ends, no matter how many plugins would have been left to check. Correct me if I am wrong - if I am, please explain what creates this behaviour.

Looking at the example, I had to exclude the subsite proxy directive from redirection, as apparently `redir` is processed before the `proxy` plugin. In this case there was a solution, but with lots of plugins, you will quickly encounter situations where a plugin will not have the ability to exclude a path. Also, the process of finding the correct configuration is a lot of effort and the resulting configuration is ugly.

This behavior makes writing Caddy config needlessly complex, as I have to know in which order plugins are being called and have to exclude all paths from earlier plugins that I want later plugins to handle.

I would suggest to handle paths similar to nginx does with its [location](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) directive, as it works really well and I guess A LOT of Caddy users will already know nginx and how paths work there, especially if you are trying to attract companies to use Caddy. (Just a quick side thought: a nginx to Caddy config converter - at least for simple things - may help in getting more users)

Though the most important thing is that the Caddy core handles path matching and not the plugins directly.

Until this issue is resolved, please update to documentation to explain how the current path matching works exactly. While searching for a solution for my use case I did not find any piece of documentation that was helpful in this sense and discovered the solution by chance. This is not what Caddy stands for.

Whatever you do, please do one thing: keep up this awesomeness.

mholt:
Hi, thanks for your comments!

[...]

Issue #619, opened 2016-02-20
Proposal: Upgrades to the Caddyfile


mholt:
A number of issues (#587, #566, #310, #167, #103, and *maybe* #584 and #209) and discussions [in gitter chat](https://gitter.im/mholt/caddy) have some things in common: they want to match requests by more than just the path prefix, and they want to be able to use variable or pre-defined values elsewhere in the Caddyfile.

In an attempt to *mostly* resolve *most* of those issues, I propose a few changes/upgrades to the Caddyfile.

### 1. More powerful path matching

Right now, most directives only accept a base path to match:

```plain
redir /foo https://sub.example.com/foo
```

If we replace `/foo` (the base path) with a more expressive matching block, we can meet most users' needs. For example:

```plain
redir {
    {path} not /
    and {>Referer} has google.com or {>Referer} has bing.com
} https://sub.example.com{uri}
```

Notice here that we've specified three conditions, where the last two are grouped because they are on the same line. We're using [placeholders](https://caddyserver.com/docs/placeholders) for the dynamic values. It is the equivalent of this pseudocode:

```go
req.URL.Path != "/" && (strings.Contains(header["Referer"], "google.com") || strings.Contains(header["Referer"], "bing.com")
```

### 2. Virtual hosts defined by more than an explicit host:port

We've had numerous requests for variable site roots, but [that's](https://github.com/mholt/caddy/issues/167#issuecomment-176822199) complicated since some directives need to know the site root when the server starts.

So how about we describe "virtual hosts" not so much as a `host:port` but as a "site" instead, and sites can be defined with wildcards and even request paths:

```
# What you're used to: a server block is headed by a host name
example.com {
    root /www/example.com
}

# But you could specify a path to have variable roots
example.com/foo {
    root /www/foo
}
example.com/bar {
    root /www/bar
}

# You could also use wildcards
# (but the *-value would not be accessible by a placeholder)
*.example.com {
    root /www/sub.example.com
}
```

However, if wildcards are used in a hostname, we are not able to issue certificates for those names. Even if Let's Encrypt enabled wildcard issuance, we wouldn't be able to issue certificates for `example*.com`, at least, not until Caddy gets On-Demand TLS (which issues certificates during the TLS handshake, about which I'm in talks with Let's Encrypt right now, hoping they'll let me do it).

### 3. Re-using blocks and arguments

[...]

PR #1948, opened 2017-11-09
'if' statements for the HTTP Caddyfile

mholt:

Implements `if` statements for the Caddyfile.

This a rough, but working, implementation of the 'if' directive (kind of
a faux-directive hacked together in the HTTP server core). It allows any
directive which supports it to conditionally execute based on params of
the request.

**This is a WIP and needs lots of attention still. It is NOT READY TO MERGE.**

Example uses:

```plain
if {remote} not_in_range 192.168.1.11/32 {
    status 402
}

if {path} ends_with foobar/ or {when} before "4:00 AM" {
    ...
}
```

General syntax:

```plain
if A op B [conj C op D [conj E op F]]...  {
    # directives as usual here, but probably
    # without the usual "base path" argument
}
```

In other words, there are two values joined by an operator. That is a condition. If there is more than one condition, each condition is joined by a conjunction (`and|or`). There may be multiple conditions and they may even span multiple lines. There is no grouping or precedence ordering. As-is, I've implemented lazy evaluation for `and` conjunctions.

The plumbing is mostly complete, but only the 'status' directive has
been upgraded to support this RequestMatcher. All standard directives
that add an HTTP middleware should be upgraded to support it. Directives
that implemented their own 'if' matching (like 'rewrite', etc) will need
to use this instead.

Other than that, tests need to be added, plugin boilerplate code on the
wiki needs to be updated, and the final interface / usage details will
need to be ironed out, especially in the httpserver package.

It may be useful to implement more condition operators too, so that
we can condition on time of day, etc. We also need to add support for
regular expression captures so they can be used by the middleware
within it...

Possible operators to consider adding:

- before/after (for date and time)
- exists (for files/directories -- but this'll have to be binary somehow)
- less/more/less_than/more_than/`>`/`<` ?? (for numbers, obviously)

/cc @abiosoft

#619 and #1862 among others

Lots! We'll need to add an `if` docs page for the HTTP server. I will draft that up. Might need to update the Placeholders docs page too, maybe? (Maybe not.) We'd also need to update the wiki, especially boilerplate code for plugin developers, so that they all use this. Oh, and every directive will make the base path argument optional (if not already) inside an `if` block.

Part of the change for plugin developers is that they should use this inside their main parsing loop (`for c.Next() {`):

```go
reqMatcher := httpserver.GetRequestMatcher(c)
```

and embed that RequestMatcher inside their httpserver.HandlerConfig struct.

### 4. Checklist

### 5. TODO

**I'll need help to get all this done. I'd love to have some contributions related to this, especially from collaborators. Feel free to ask if you aren't sure how to get started with something.**

- [ ] How does the interface look for plugin developers who want to use it? Can/should we improve it?
- [ ] All directives that add HTTP middleware need to be upgraded to support this. That implies making the base path argument for each directive optional _if a RequestMatcher is not nil_ (meaning that it's in an `if` statement).
- [ ] Regular expression captures need to be implemented so that middlewares inside the if statement can use them.
- [ ] Write tests!
- [ ] While we're at it, it would be good to get every middleware using the HandlerConfig interface so that when base paths are used to choose a middleware config to run, the longest base path always has precedence. Not sure what is required for this.
- [ ] Remove the if and if_op subdirectives from rewrite, redir, and any other directive which uses them.
- [ ] Please try it out! We need feedback from people before this is merged in and released. 

Thanks to all who help!

(NOTE: Let's not let our `if` become evil: https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)


caddyfile/parse.go
```diff
@@ -33,6 +33,19 @@ func Parse(filename string, input io.Reader, validDirectives []string) ([]Server
 	return p.parseAll()
 }
 
+// ParseBlockTokens takes a list of tokens that come from the
+// body of a single block into a nicely parsed map of lists of
+// tokens, keyed by directive name. It's essentially parsing
+// only a partial input into lists of tokens by directive name;
+// no server blocks or labels.
+func ParseBlockTokens(tokens []Token, validDirectives []string) (map[string][]Token, error) {
+	p := parser{Dispenser: NewDispenserTokens("(no filename)", tokens), validDirectives: validDirectives}
```
lindaxie: testing SG

pwhodges:
I'd like to add a request for "exists" to be implemented, as I have a use case for it.  See [this thread](https://caddy.community/t/testing-file-existence/3431).  It can be made binary by writing, for instance:

    if {path} contains "testfilename"

maciekish:
Is there way to be notified when this is eventually merged? It would cut my Caddyfile by more than half as i do conditional redirects and reverse proxying based on `client ip`.

steamp0rt:
This would be be EXTREMELY useful for pre-compressed gz files. I can't figure out how to have the server avoid double-compressing gz files without sending false content-encoding headers to un-precompressed files

magikstm:
> 
> 
> This would be be EXTREMELY useful for pre-compressed gz files. I can't figure out how to have the server avoid double-compressing gz files without sending false content-encoding headers to un-precompressed files

Double-compressing of gz files shouln't happen. It doesn't for me on Windows 7. It is also documented as such in the docs here:
https://caddyserver.com/docs/gzip

Are your files named .gz?

If you're using latest version of Caddy, I suggest opening an issue about it for further investigation.

steamp0rt:
@magikstm Yeah, they're named .gz

for example: game.txt.gz

steamp0rt:
@magikstm It appears there's already an issue: https://github.com/mholt/caddy/issues/2230

ShadowJonathan:
My experience with Nginx told me that `if` statements were evil somehow, and "stole" requests (according from some personal suspicions, and documentation i started reading about it).

How are if statements implemented here? Are they iterated sequentially till one matches? Are all of them iterated? With what rules/entries inside a block statement will an if statement stop? `proxy`? `fastcgi`?

My suggestion, if all if statements are iterated regardless of their contents, is to also add a `break` statement, which stops the iteration of if statements, and finishes a request.

elcore:
@ShadowJonathan

That's the goal

> (NOTE: Let's not let our if become evil: https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)

prophit987:
- [ REVIEW]

ShadowJonathan:
@mholt

mholt:
Man... I kinda hate this.

Closing for now.

ShadowJonathan:
Why has this pull request been delayed?

francislavoie:
@ShadowJonathan just read the TODO section of the original post, there's still a lot of questions unanswered. It also kinda went stale due to other priorities. It's also a very big/risky feature addition to add.

Also, in general, there are existing workarounds to if statements, like using rewrite to match certain conditions, prepending something to the path, then matching on that path. This doesn't solve everything, but it can help. If you have a specific usecase where you're thinking `if would help, try asking on https://caddy,community, someone might be able to help you find a workaround. If not, PRs/discussion are always welcome to try and help support those usecases.

daiaji:
I am still looking forward to the completion of this feature.
In fact, I need to do basic authentication except for a certain path under a certain domain name.
For example, I am building the aria2 webUI, aria2's RPC itself can use the original authorization token for authentication, and aria2 webUI requires basic authentication.
The existing path matching made me have to use aria2RPC with a different subdomain of aria2 webUI.

mholt:
@daiaji There might be a way to do what you need to do using the `rewrite` directive as a workaround; it supports complex conditions like you need.

However, this PR will probably never be finished since Caddy 2's middleware architecture is even more powerful than these `if` statements in the Caddyfile. It allows you to describe any kind of matching logic you need.

mholt:
Caddy 2's configuration is solely declarative but provides the power and flexibility of "if-then" logic. Quoting from nginx's blog post about `if` being evil:

> Directive “if” is part of rewrite module which evaluates instructions imperatively. On the other hand, NGINX configuration in general is declarative. At some point due to users demand an attempt was made to enable some non-rewrite directives inside “if”, and this lead to situation we have now. It mostly works, but... see above.

In Caddy 2, we have avoided this problem of mixing declarative and imperative syntaxes, which is what this PR almost did. It is something I have been wary of a long time. But fortunately, in Caddy 2, there's also an embedded scripting language that provides true imperative syntax where necessary, without it becoming a "split-personality config," so to speak.

Best of both worlds in Caddy 2, without the weakness of nginx's config design.

coolaj86:
There's two cases I've come across where the `handle @my-matcher { ... }` doesn't seem to apply:
- If a DNS provider ENV is set, I want to use it for `tls`
- If a API_TOKEN_DIGEST is set, I want to enforce `basicauth`

This would allow me to distribute a Caddyfile as a template (in a Docker instance, for example) and the precense of the `API_TOKEN_DIGEST` when the container is run would enable authentication, etc.

Makes it easier to package an in-a-box "appliance".

coolaj86:
Or maybe that's what the CEL stuff is about?

Something like this?

```hcl
@match-enforce-auth `{env.API_TOKEN_DIGEST}.size() > 0`
handle @match-enforce-auth {
    basicauth {
        {env.API_KEY_ID} {env.API_TOKEN_DIGEST}        
    }
}
```

I've seen that section before but I didn't understand what it was (and I've never encountered another project mentioning it) so I just read past it.

mholt:
Yeah, you can use CEL to do what you're wanting. It confines the imperative logic to its matcher "sandbox" and keeps your config declarative. Even better than `if`.

coolaj86:
I CAN'T BELIEVE IT'S NOT BUTTER!

HOLY CHEESE SNIPS THAT WORKED IN REAL LIFE!

`.env`:
```sh
OLLAMA_API_ID='my-api-token'
OLLAMA_API_TOKEN='xxxxxxxxxxxxxxxx'
# generated by passing the token to `caddy hash-password`
OLLAMA_API_TOKEN_DIGEST='$2a$14$iyyuawykR92xTHNR9lWzfu.uCct/9/xUPX3zBqLqrjAu0usNRPbyi'
```

```Caddyfile
localhost {
    ## HTTP Basic Auth
    ## (set OLLAMA_API_TOKEN_DIGEST to enable)
    @match-enforce-auth `"{$OLLAMA_API_TOKEN_DIGEST}".size() > 0`
    handle @match-enforce-auth {
        basicauth {
            # see .example.env for how to generate tokens
            {$OLLAMA_API_ID} {$OLLAMA_API_TOKEN_DIGEST}
        }
    }

    handle /api/* {
        reverse_proxy localhost:11434
    }

    file_server {
        root ./ollama-webui/
    }
}
```

coolaj86:
Oh, actually that **didn't work**. The cached page just happened to load the first time after the authentication request.

`handle` terminates the request, so either have to duplicate the other handlers or make a snippet that's included in two places, or so it seems.

francislavoie:
`handle` doesn't terminate, but it is mutually exclusive to other `handle` at the same level.

Next time, please consider opening a topic on the forums instead of commenting on a very old issue, especially one from Caddy v1 which is long-since EOL.

coolaj86:
> Next time, please consider opening a topic on the forums

Absolutely not! I fight for the users!

The forums are where bad information goes stale. Here, which is among the first results for the search terms, I was looking for, I can make a comment, and someone else can benefit from the knowledge I've learned.

you may, however, consider unlocking the forums. There are dozens of questions that have bad information or no answer or only old answers that once were correct that I have to comment on with links to the correct information.

Issue #1952, opened 2017-11-13
Browse - directory modified date column misalignment


jungle-boogie:
### 1. What version of Caddy are you using?

Caddy 0.10.10 (+9a22cda Sun Nov 12 22:13:44 UTC 2017) (unofficial)

### 2. What are you trying to do?

Use the `browse` feature without the Modified column being misaligned 

### 3. What is your entire Caddyfile?
```text
browse
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

`sudo caddy -conf /usr/local/www/Caddyfile`

### 6. What did you expect to see?

Properly aligned columns. All columns but the modified are correct.

### 7. What did you see instead (give full error messages and/or log)?

[Imgur](https://i.imgur.com/VduuKvz.png) <-- pictorial description of the problem. Notice the Modified column isn't correct. 

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

Probably just using the `browse` feature, you will notice the alignment is off in the modified column.

On a related note, how can the YYYY-MM-DD format be implemented within this `browse` feature?

jungle-boogie:
On further review, I think this would be corrected if all DD and MM used two digits.
The clock should also be 24hours.

Before:
```
6/6/2017, 4:24:17 PM
7/10/2017, 12:09:05 AM
```

After:
```
06/06/2017, 14:24:17
07/10/2017, 12:09:05
```

I'd still prefer a YYYY-MM-DD 24hr. 

Or something like this:
`11-Nov-2017 01:04`
From here:
https://ftp.openbsd.org/pub/OpenBSD/snapshots/octeon/

tobya:
I think you can specify your own template for the browse directive, which would allow you to specify how you wanted the date to be formatted.

jungle-boogie:
Thanks for the input @tobya.

I'd still like to see ISO 8601 implemented in the default template.

tobya:
We are always delighted to accept a pull request for an update.  I will mark this as a feature request.

jungle-boogie:
Hi @tobya, there's a pull request to align the column: https://github.com/mholt/caddy/pull/1954

tobya:
This was fixed in #1954

PR #1954, opened 2017-11-13
Correct browse modified date alignment

magikstm:
Correct Browse "modified date" alignment that may be a bit misaligned depending on time and date.

I'm unsure if it is the best way to correct the issue.

#1952

None.

### 4. Checklist

### DIFF
--- a/caddyhttp/browse/setup.go
+++ b/caddyhttp/browse/setup.go
@@ -499,7 +499,11 @@ footer {
 						return;
 					}
 				}
-				e.textContent = d.toLocaleString();
+				//Remove +00:00 from e.textContent
+				e.textContent = e.textContent.substring(0, e.textContent.lastIndexOf(" "));
+				
+				//Add a ',' to separate date and time
+				e.textContent = e.textContent.replace(' ',', ');
 			}
 			var timeList = Array.prototype.slice.call(document.getElementsByTagName("time"));
 			timeList.forEach(localizeDatetime);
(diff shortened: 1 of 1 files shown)


francislavoie:
In addition to my note in the review, I don't think this will still be perfectly aligned depending on the font. It'll need testing to see how it actually behaves.

jungle-boogie:
I updated the issue that caused this PR with a, IMO, better solution.

magikstm:
Thanks for the review. I could adjust. Please confirm which solution would be preferred (FrancisLavoie's suggestion or the suggestion in the PR).

francislavoie:
@jungle-boogie my suggestion matches your first suggestion. I disagree with your 2nd one because it complicates localization using the JS Date API (unless you're willing to write up a solution that keeps that format but is still properly localized)

magikstm:
@francislavoie I tested your suggestion locally on a Windows 7 x64 machine on IE11 (latest), Firefox (latest) and Chrome (latest) and it works reasonably well.

Current font which is "sans-serif" makes it not perfectly aligned, but I think it's quite usable.

jungle-boogie:
@francislavoie, What's complicated about it? I'm not saying show the local date/time, just format it as per ISO8601.

francislavoie:
@jungle-boogie Yeah, ISO8601 would be fine too. I was reading your example of `11-Nov-2017 01:04` which would require localization if using text for the month. I just think this solution is the easiest short-term without changing the date format.

jungle-boogie:
Thanks for the reply, @francislavoie! Looking forward to this change.

mholt:
Thanks everyone!

PR #1959, opened 2017-11-15
Add dontstrip plugin

dhaavi:

Adds an option for sites not to strip the path.
This is important to be compatible to software that does not work well with path stripping/trimming.
My use case was that the webdav plugin did not work with the Linux Gnome Desktop, as trimming the path would confuse the client.

My guess is that this will be very useful to a lot of users, as custom protocols may use paths within headers and payload, which will stop being valid when only the request URI is changed.

No issue was created.

Create a new documentation page for this plugin, stating something like:

    If this option is used, caddy will not trim the original request path for this site.

### 4. Checklist

  - There is so little code, that additional tests don't make any sense.

### DIFF
--- a/caddyhttp/caddyhttp.go
+++ b/caddyhttp/caddyhttp.go
@@ -22,6 +22,7 @@ import (
 	_ "github.com/mholt/caddy/caddyhttp/basicauth"
 	_ "github.com/mholt/caddy/caddyhttp/bind"
 	_ "github.com/mholt/caddy/caddyhttp/browse"
+	_ "github.com/mholt/caddy/caddyhttp/dontstrip"
 	_ "github.com/mholt/caddy/caddyhttp/errors"
 	_ "github.com/mholt/caddy/caddyhttp/expvar"
 	_ "github.com/mholt/caddy/caddyhttp/extensions"
--- a/caddyhttp/dontstrip/dontstrip.go
+++ b/caddyhttp/dontstrip/dontstrip.go
@@ -1,0 +1,33 @@
+// Copyright 2015 Light Code Labs, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


mholt:
Sorry, but this isn't a good fit for Caddy (I always recommend opening an issue before a pull request if unsure). We deliberately do this for security and functionality reasons. If the webdav plugin doesn't work well with it, it's best to open an issue with them about it. The unmodified path can be obtained from the request Context.

dhaavi:
Interesting.

I think it's best to fix a problem where it is beautiful by design.
Going into the webdav plugin and adding lots of header and maybe even payload mangling just introduces bugs and is a lot of work - on the other side adding a small flag to caddy seems like a very elegant solution.

Also, caddy plugins is just the small part of the issue. I guess this plugin will be mostly used in combination with proxy, having a external application deal with the content. Introducing changes there might even be a lot harder or impossible (closed source).
I know that `proxy` does not trim the path, but the great thing about this plugin is that you can have separate plugin stacks for proxies served under the same domain, while not modifying the path anywhere, staying compatible with software.

Can you elaborate on the security and functionality reasons?

mholt:
The technical problem with this PR is that it breaks all other middlewares that expect the path to be stripped, (which is all of them, if implemented correctly). It also allows auth to be bypassed because the base paths won't match.

For when you do need the original URI, you can get the original URI [from the context](https://github.com/mholt/caddy/pull/1959/files#diff-e74f72185700e304b2047c4af2ad1a48R356) with one line of code. Quite elegant.

dhaavi:
Of course the user has to adapt the configuration of the other plugins used in stack to work with the full path, not the trimmed one. I don't see how that breaks anything if the user is aware of that.
Could you provide an example?

Yes, quite elegant indeed. This would be an option for the webdav plugin. Still a problem for external software.

PR #1963, opened 2017-11-22
Allow graceful restart of TLS listener

yuewko:
Previously, graceful restart of TLS listener does not work because:
1) tls.Listener does not implement the caddy.Listener interface as it
   lacks the File() method.
2) tls.Listener utilizes an underlying net.Listener which breaks the
   code in startServers() that attempt to do a dup.

This change introduces a new caddy.DuppableListener interface that
encapsulates the dup logic for TLS listener.

Fixes issue: https://github.com/coredns/coredns/issues/1244

(please see commit msg)

https://github.com/coredns/coredns/issues/1244

Document additional interface, caddy.DuppableListener

### 4. Checklist

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -177,15 +177,22 @@ func (i *Instance) Restart(newCaddyfile Input) (*Instance, error) {
 	restartFds := make(map[string]restartTriple)
 	for _, s := range i.servers {
 		gs, srvOk := s.server.(GracefulServer)
-		ln, lnOk := s.listener.(Listener)
+		dupLn, dupLnOk := s.listener.(DuppableListener)
+		if !dupLnOk {
+			ln, lnOk := s.listener.(Listener)
+			if lnOk {
+				dupLn = makeDuppableListener(ln)
+				dupLnOk = true
+			}
+		}
 		pc, pcOk := s.packet.(PacketConn)
 		if srvOk {
-			if lnOk && pcOk {
-				restartFds[gs.Address()] = restartTriple{server: gs, listener: ln, packet: pc}
+			if dupLnOk && pcOk {
+				restartFds[gs.Address()] = restartTriple{server: gs, listener: dupLn, packet: pc}
 				continue
 			}
-			if lnOk {
-				restartFds[gs.Address()] = restartTriple{server: gs, listener: ln}
+			if dupLnOk {
+				restartFds[gs.Address()] = restartTriple{server: gs, listener: dupLn}
 				continue
 			}
 			if pcOk {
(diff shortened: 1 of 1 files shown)


johnbelamaric:
@mholt we need this for a CoreDNS fix. Any chance you can take a look?

mholt:
I'm taking a look at it. But pardon my delay as I get over a cold and travel to NIPS.

yuewko:
@mholt, no problem!! Thanks for reviewing it - and hope you feel better soon!

UladzimirTrehubenka:
Could someone approve and merge the PR?

johnbelamaric:
@yuewko perhaps you should re-base this one

yuewko:
@johnbelamaric - Will do!

mholt:
I have spent the better part of this evening trying to confirm that graceful reloads over a TLS connection are broken, but my experiments seem to show that it is fully functional.

Caddyfile:

```
https://localhost {
	tls localhost.crt localhost.key
	browse
}
```

I also added these lines into the ServeHTTP of browse.go:

```go
fmt.Print("**SLEEPING** ")
time.Sleep(10 * time.Second)
fmt.Println(" **DONE**")
```

And ran Caddy:

```
caddy -log stderr -grace 30s
```

Then I loaded `https://localhost` in the browser. It took 10 seconds, then showed me the directory listing, as expected.

Then I modified the Caddyfile to remove the `browse` directive. I refreshed the web page, and the 10 second sleep started. While sleeping, I quickly reloaded Caddy: `sudo pkill -SIGUSR1 caddy` and went back to my browser. I went to a new tab, `https://localhost`. The directory listing successfully reloaded in the first tab while the second tab showed me a 404 page (because I don't have an index file), as expected.

After the reload completed, refreshing the first web page that still showed the directory listing then showed the 404, confirming that the new config was reloaded. No connections were dropped; the browser showed both pages.

Is this inconsistent with your own experience?

yuewko:
Umm... that's certainly inconsistent with what we observed as per https://github.com/coredns/coredns/issues/1244

Basically, the restart would fail because the old listener was not closed.

Let me try out your exact scenario in the morning and get back to you.

BTW - thanks for taking the time to review this!!

yuewko:
@mholt ,

OK, I tested your exact scenario - and yes, it is consistent with what you observed.

The reason why it did not exhibit the reported issue is that, specifying `tls` in caddy, it creates a `httpserver/tcpKeepAliveListener`, which basically wraps around a `net.TCPListener`:

```
type tcpKeepAliveListener struct {
	*net.TCPListener
}
```
Hence, it satisfies the `caddy.Listener` interface.

However, in the CoreDNS issue (https://github.com/coredns/coredns/issues/1244) that is reported, the listener used is a `crypto/tls.listener`, which does NOT implement the `caddy.Listener`.

Guess I could have been clearer in the original description.

However, the GOOD NEWS is, I have run your test case with the changes in the PR - and it still works. In fact, I made it an explicit goal to not break any existing code.

Hope this helps!

mholt:
Thanks -- that's very helpful! I have some good news. (I think/hope!) If CoreDNS is using "tcp" as the network type when calling `tls.Listen`, then you might be able to fix this in CoreDNS.

Yes, `tls.Listen` returns a `net.Listener`, but you can do the same thing (I think!) by calling net.Listen, then asserting that to a `*net.TCPListener`, which has the needed `File()` method.

I'm just armchair debugging, but I'm crossing my fingers anyway, because this is how Caddy's HTTPS server does it.

yuewko:
@mholt , thanks for sharing your thoughts!

I have gone down a similar path before - no joy though. In short, the object returned by `tls.Listen()` cannot be asserted to `*net.TCPListener`. However, the _inner_ `net.Listener` of that object can (which is how `Dup()` is implemented [here](https://github.com/coredns/coredns/pull/1254/files#diff-4aefc06e72d42d0c6427e4856b1b456eR35) )

Which give me an idea - I can just have the `caddy/TCPServer.Listen()` method in CoreDNS returns the  _inner_ `net.Listener`, instead of the object returned by `tls.Listen()`.

Thanks again - I'll report back.

yuewko:
@mholt , I have implemented a fix in CoreDNS that does not require this change. (FYI, this is the CoreDNS PR https://github.com/coredns/coredns/pull/1389). Hence closing this one.

Thank you for your time and suggestions!!

mholt:
Great! Glad I could be of help. Thanks for thinking of the problem in a different way to find (what I think is) a better solution.

Issue #1858, opened 2017-09-07
need to update support chrome 61 quic/39


favortel:
The current version of caddy cann't support the newewst version quic/39 with chrome61

KEY word: QUIC 39

wendigo:
It will when quic-go 0.6 will be released: https://github.com/lucas-clemente/quic-go/blob/master/Changelog.md

favortel:
wo, thank you

marten-seemann:
Please don't wait for quic-go releases. We're really lazy tagging releases on Github.
Please just go ahead and use the current master. We're making sure that the current master always works with the current version of Chrome.

LFlare:
Ping @wendigo @mholt :)

PR #1968, opened 2017-12-02
Updated quic-go for QUIC 39

LFlare:
This is just a simple PR pointing towards the latest commit on lucas-clemente/quic-go's master branch.

#1858. 

Just that it's updated for QUIC 39?

### 4. Checklist

- [x] I have written tests and verified that they fail without my change (`go test` passes)
- [x] I am not willing to help maintain this change if there are issues with it later (I merely updated the vendoring files).


lbguilherme:
The lastest commit on master of https://github.com/lucas-clemente/quic-go is not a released version.

https://github.com/lucas-clemente/quic-go/releases

https://github.com/lucas-clemente/quic-go/blob/master/Changelog.md

It is probably better to wait for a release.

LFlare:
@lbguilherme, I understand your concerns over this but the maintainers of quic-go has stated that they will not be tagging releases and rather would just work towards ensuring that the master branch keeps up-to-date with Google Chrome's QUIC support.

https://github.com/mholt/caddy/issues/1858

marten-seemann:
Author of quic-go here. @LFlare is right, we're currently not tagging any releases of quic-go. 
Please always use the current master of the quic-go repo. This applies to this PR as well as to future Caddy releases.

LFlare:
Hello @mholt,

Would you review this?

mholt:
Just got back from a trip, please give me a few weeks to catch up with the end of school and the holidays.

LFlare:
No problem @mholt, just try to prioritize this as soon as you can, just for it's importance level :)

LFlare:
Hey @mholt, are you able to review this any time soon?

marten-seemann:
I recently created a new release of quic-go: https://github.com/lucas-clemente/quic-go/releases/tag/v0.6.0. Feel free to use that one, if you prefer to vendor on git tags.

13111107:
When the official will be out of caddy support quic-39

quasipedia:
This is awesome.... but it seems like this PR has stalled?  Support for QUIC is _the_ reason why I use Caddy... :(

francislavoie:
@quasipedia this'll be merged (or done manually in `master`) right before the next release by @mholt.

quasipedia:
That's even more awsome @francislavoie ! :)  Ths obvious follow-up question is: when will the next realise be?

The thing is: there is only production-ready one FLOSS server supporting QUIC (caddy)

mholt:
Working on it. Some critical changes need to get into the next release. I'm mainly eyeing ACMEv2 and wildcard support, which should be hopefully in February.

marten-seemann:
I just released [quic-go v0.7.0](https://github.com/lucas-clemente/quic-go/releases/tag/v0.7.0). This release implements packet pacing, which should significantly increase the throughput of QUIC connections. Furthermore, it contains a bunch of optimizations, which will result in a lower CPU consumption.
Please make sure to update the vendored quic-go version in order to take advantage of these changes.

LFlare:
@mholt, @lbguilherme and @marten-seemann, I've updated the repository to reflect the latest commit on lucas-clemente/quic-go. Hopefully this time it can be merged.

lbguilherme:
I believe this should be merged

LFlare:
@mholt, in that case, please mark it with the milestone for next release. Honestly, this update has been delayed for nearly half a year. For a product that claims to be the easiest zero-configuration secure-by-default web server, not having the latest QUIC version is just depressing.

marten-seemann:
Updating the vendored quic-go just before the release is a good idea, since QUIC (the protocol) is moving fast, and quic-go is as well. I don't know your build and release process, but maybe it would even make sense to automate that?

Nevertheless, I think this PR could be merged right now. All the people building Caddy from source would be able to use QUIC again (after support was effectively broken for almost half a year). I got a few bug reports from Caddy users in the past, mostly for bugs that would've stayed undiscovered for a long time otherwise, so making the quic-go changes of the 6 months or so available to people for testing would actually be helpful for me.

quasipedia:
> For a product that claims to be the easiest zero-configuration secure-by-default web server, not having the latest QUIC version is just depressing

@LFlare - Me too have been terribly disappointed by caddy not supporting v.39, and had to fix it myself by merging your PR locally instead (thank you for that!).  However I think is a bit unfair to attack @mholt for not giving precedence to an experimental feature supporting an experimental protocol (experimental implies no guarantees about security).  I have started with QUIC literally a couple of weeks ago - so I may be wrong here - but my understanding is that Chromium used to support a few versions of the protocol at once, while the latest version only supports v.39 and this exacerbated the problem.  Not a caddy fault, I'd say...

marten-seemann:
> I have started with QUIC literally a couple of weeks ago - so I may be wrong here - but my understanding is that Chromium used to support a few versions of the protocol at once, while the latest version only supports v.39 and this exacerbated the problem. Not caddy's fault, I'd say...

While the QUIC protocol itself supports version negotiation (i.e. both client and server both support a set of versions, and agree on a common version during connection establishment), Chrome never actually supported more than one version at a time. That's why in quic-go, we always support all QUIC versions for which Chromes are out there (with a non negligible market share).  Currently, this is only QUIC 39, since Google hasn't released a new version for quite some time, but there were times where quic-go used to support 4 different QUIC versions or so.
Unfortunately, the fact that Chrome only supports a single version means that in order to continue using QUIC, you'll have to update your server as soon as Google rolls out a new QUIC version into Chrome.

itsnotvalid:
@marten-seemann it's fine for only supporting the supported version used in Chrome, as non-functioning versions of QUIC would gracefully degrade to HTTP/2 and no other browsers are supports QUIC on default setting anyways. Note that one should always test if it's faster for using QUIC over HTTP/2 in their own scenarios.

mholt:
@marten-seemann Thanks for working on quic-go. Expect this to go out Monday or so. I noticed the quic-go library is gaining some pretty large dependencies. Any idea if that trend is going to continue? (This update added about 22,000 lines of code.)

marten-seemann:
@mholt The only dependency of quic-go (in production code, and not just for tests or code generation) is mint, a TLS 1.3 library. It will be needed for the QUIC Version that is currently being specified by the IETF. I'm not planning to add any other dependencies beyond a TLS library.

mholt:
Thanks Marten -- that's good then. I hope, when Go gets TLS 1.3 into the standard lib, maybe that dependency will not even be necessary! Thank you again.

Issue #1962, opened 2017-11-20
CPU cap (flag -cpu) with a small percentage is silently ignored


detaoin:
### 1. What version of Caddy are you using (`caddy -version`)?
Caddy 0.10.10 (non-commercial use only)
However it is also present on current master (commit 1125a236eabb61bbccb5b6a1af1a48e39da59a20)

### 2. What are you trying to do?
N/A (just reading the source code)

### 3. What is your entire Caddyfile?
_empty_

### 4. How did you run Caddy (give the full command and describe the execution environment)?
GODEBUG=schedtrace=1000 caddy -cpu 1%

### 5. Please paste any relevant HTTP request(s) here.
N/A

### 6. What did you expect to see?
The _console_ (stderr) output should contain the scheduler state summary line every second:
```
SCHED 0ms: gomaxprocs=4 idleprocs=1 threads=5 spinningthreads=1 idlethreads=0 runqueue=0 [0 0 0 0]
Activating privacy features... done.
http://localhost:1515
SCHED 1009ms: gomaxprocs=1 idleprocs=1 threads=8 spinningthreads=0 idlethreads=3 runqueue=0 [0]
```

The `gomaxprocs=X` information changes from 4 to 1 after the command line flags have been parsed; we requested (`-cpu 1%`) a very small cpu cap, thus logically _saturate_ at 1 GOMAXPROC, the minimum possible value. (*Note* on this machine I have 4 logical CPUs).

### 7. What did you see instead (give full error messages and/or log)?
```
SCHED 0ms: gomaxprocs=4 idleprocs=1 threads=4 spinningthreads=1 idlethreads=0 runqueue=0 [0 0 0 0]
Activating privacy features... done.
http://localhost:1515
SCHED 1009ms: gomaxprocs=4 idleprocs=4 threads=7 spinningthreads=0 idlethreads=2 runqueue=0 [0 0 0 0]
```

At https://github.com/mholt/caddy/blob/97710ced7ea5b5d8025dc6a3e43d82e9a280920e/caddy/caddymain/run.go#L238 the available number of logical CPUs (`availCPU`) is multiplied by the `1%` specified cap, however the result is rounded down to the nearest integer (0), and ignored by https://golang.org/pkg/runtime/#GOMAXPROCS because smaller than 1.

`numCPU` should have a lower cap of 1, this can be implemented with

```
if numCPU < 1 {
        numCPU = 1
}
```

[...]

PR #1969, opened 2017-12-04
caddymain: fix setCPU silently ignoring small percent values

detaoin:
caddy/run.go:238
the percent value is resolved in a GOMAXPROCS relative number by simple
division, thus rounding down the non-integer quotient. If zero, the call
to runtime.GOMAXPROCS is silently ignored.

We decide here to exceptionally round up the CPU cap in case of percent
values that are too small.

#1962 

### 4. Checklist


detaoin:
No problem, it wasn't much.

PR #1970, opened 2017-12-04
Update README.md

seanlane:
I believe the owner and group of the `chown` command here are mixed up. As it was caused a permissions issue, with the service being unable to read the directory.

Change the command for changing the ownership of the `/etc/ssl/caddy` directory from `www-data:root` to `root:www-data`.

I don't believe there are any relevant issues at the time of writing.

This change consists of the needed documentation change.

### 4. Checklist

### DIFF
--- a/dist/init/linux-systemd/README.md
+++ b/dist/init/linux-systemd/README.md
@@ -46,7 +46,7 @@ sudo useradd \
 sudo mkdir /etc/caddy
 sudo chown -R root:www-data /etc/caddy
 sudo mkdir /etc/ssl/caddy
-sudo chown -R www-data:root /etc/ssl/caddy
+sudo chown -R root:www-data /etc/ssl/caddy
 sudo chmod 0770 /etc/ssl/caddy
 ```
 
--- a/dist/init/linux-systemd/README.md
+++ b/dist/init/linux-systemd/README.md
@@ -44,9 +44,9 @@ sudo useradd \
   --system --uid 33 www-data
 
 sudo mkdir /etc/caddy
-sudo chown -R www-data:www-data /etc/caddy
+sudo chown -R root:www-data /etc/caddy
 sudo mkdir /etc/ssl/caddy
-sudo chown -R www-data:www-data /etc/ssl/caddy
+sudo chown -R root:www-data /etc/ssl/caddy
 sudo chmod 0770 /etc/ssl/caddy
 ```
 
--- a/dist/init/linux-systemd/README.md
+++ b/dist/init/linux-systemd/README.md
@@ -44,9 +44,9 @@ sudo useradd \
   --system --uid 33 www-data
 
 sudo mkdir /etc/caddy
-sudo chown -R root:www-data /etc/caddy
+sudo chown -R www-data:www-data /etc/caddy
 sudo mkdir /etc/ssl/caddy
-sudo chown -R root:www-data /etc/ssl/caddy
+sudo chown -R www-data:www-data /etc/ssl/caddy
 sudo chmod 0770 /etc/ssl/caddy
 ```
 


Whitestrake:
Correct me if I'm wrong, but isn't `0770` user and group writable? Why does it matter whether the user or the group is `www-data`?

Shouldn't they all be `www-data` anyway (since `root` can execute them regardless of who owns it)?

seanlane:
I think you're right, moving both owner and group to be `www-data` should be correct, I'll change that accordingly.

seanlane:
@Whitestrake The problem I was having was running Caddy as a user that wasn't `www-data` but was part of the `www-data` group. When the permissions are set to `0770 www-data:root`, Caddy didn't have rights to read `/etc/caddy/ssl`. So changing it to either `www-data:www-data` or `root:www-data` should work, as long as `www-data` is the group. Otherwise, as it is, Caddy has to be run by the `root` user to read/modify that directory.

I've changed the request to have both lines 47 and 49 chown to `www-data:www-data` which seems to be consistent with the rest of the README. Hope this is helpful!

Whitestrake:
@seanlane I'm not convinced that this fixes your issue for the reason you think it does.

`chmod 0770` means:

"Special" bit: neither `setuid` nor `setgid` nor `sticky`
"User/Owner" bit: executable
"Group" bit: executable
"World" bit: non-readable

So as long as Caddy's running user is a member of `www-data` (which it should be), they should be able to execute `/etc/caddy/ssl` just fine on completion of the guide prior to the changes in this PR.

That said, I agree with changing the lines to `www-data:www-data` (I think owning anything as root makes no sense in general unless it shouldn't be accessed).

seanlane:
@Whitestrake Sorry for the additional comment, but I'd like to understand the implications of the change and I'm not following your reasoning just yet.

The original version on line 49 in the guide read:

```
$ sudo chown -R www-data:root /etc/ssl/caddy
```
As I understand it, since the permissions are set to `0770`, this should mean that `www-data` is the owner of the directory and `root` was set as the directory's group. So the running user would have to be either the `www-data` user or part of the `root` group, otherwise the directory would be unreadable. Since `www-data` is a service account, and only the `root` account is normally part of the `root` group, this would mean that the directory is unreadable to any user that isn't running as a superuser. Is that not the case?

Since the latest commit then changes the aforementioned line to 

```
$ sudo chown -R www-data:www-data /etc/ssl/caddy
```
then now any user that is part of the `www-data` group, which the running user should have already been added to, can now read, write, and execute on this directory. That is my intuition behind why the change fixes the issue I was having, but I'd really appreciate any feedback if I'm not thinking this through properly.

Whitestrake:
Oh, you're right, I mixed up `/etc/caddy` and `/etc/ssl/caddy` a bit there.

Regardless, you're running Caddy as `www-data`, right? And `www-data` is a member of the `www-data` group, right?

If that's the case, only one of the user or group bits needs to be executable, for Caddy to read it, as long as Caddy is running as `www-data`.

wjkohnen:
Why should Caddy be able to mutate files below /etc/caddy? Having `root:www-data` with mode 640,660,770,750 is a basic security pattern. As long as Caddy runs with user and group id www-data the permissions should be fine. Please don't give daemon users ownership of files they do not need to mutate!

Whitestrake:
Good point @wjkohnen. `/etc/caddy` is still on default mod (I believe that's `0755`), so its owner should still be something other than `www-data` as it needs to read that data but should not write it.

seanlane:
I've reverted the changes back to the original suggestion, based on the comments by @wjkohnen and @Whitestrake

mholt:
Thanks!

Issue #550, opened 2016-01-29
Import should wrap imported file in curly braces if applicable


mholt:
This was mentioned on gitter the other day. If your Caddyfile looks like this:

``` plain
import site1.conf
import site2.conf
```

Then both site1.conf and site2.conf need to use curly braces:

``` plain
site1.example.com {
    ...
}
```

``` plain
site2.example.com {
    ...
}
```

For import statements that appear at the top level of a Caddyfile, perhaps it should check if the imported file uses curly braces, and if not, it should add them internally, so that the individual conf files can stay "flat":

``` plain
site1.example.com
...
```

``` plain
site2.example.com
...
```

I'm just opening this idea as a memo to self, but I haven't decided on it entirely quite yet. I like the idea of import being a simple drop-in replacement.

captncraig:
would it make sense to have an additional import statement for this? `import` vs `importFlat` or something like that? If you can detect that it needs curlys I would say do that, but it sounds hard.

mholt:
It's only hard if we don't assume that the import is not well formed.

Given these files:

**Caddyfile**
```plain
import site1.conf
site2.dev {
    gzip
    browse
}
```

**site1.conf**
```plain
site1.dev

gzip
templates
```

Because the import occurs in the Caddyfile where an address should be, we can assume that the first line in the imported file is an address. Thus if that first line doesn't end with `{` then we know we have to wrap it in curly braces.

However, a malformed import would be the same Caddyfile, for example, but with this site1.conf:

```plain
gzip
templates
```

If the imported file doesn't, in fact, start with an address, it will interpret `gzip` as the address, and wrap `templates` in curly braces. This is obviously an error, but the import was poorly formed.

Any import directive inside a server block will not have the imported file wrapped in curly braces. Just the ones that appear where an address should be.

alexandreteles:
I think that we should have only one import directive, the one that we already have, with the exactly behavior that we have today. The user should be responsible for writing the imported files using the correct syntax for where he will import it. If we start to make assumptions, we can't predict where edge cases will appear and how bad it can turn. Just leave it as it is.

tw4452852:
At first look at this issue, I don't like this feature. As @alexandreteles says, it will lead some edge cases.

But for the possible way of implementation, I think we could drop the braces entirely. We could add braces internally, but we should check the conflict scope definition:

1. If the imported file doesn't start with address, it belongs the scope which imports it. If it is the uppermost one, it belong the default global scope.

[...]

PR #1973, opened 2017-12-13
confLoader: use import to handle globbed values for -conf flag

Whitestrake:
Implements handling for glob values for the -conf flag.

https://github.com/mholt/caddy/issues/1910

https://caddyserver.com/docs/cli should be updated to mention the ability to use globs.

### 4. Checklist

---
95% of the time spent preparing this pull request was figuring out how to fork, branch, then fix my branch, then rebase and start over, then squash two commits into one branch.

This proposed PR feels pretty hacky to me, but it works and I think the only alternative is to copy a whole lot of code out of `doImport` itself.

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -170,10 +170,18 @@ func confLoader(serverType string) (caddy.Input, error) {
 		return caddy.CaddyfileFromPipe(os.Stdin, serverType)
 	}
 
-	contents, err := ioutil.ReadFile(conf)
-	if err != nil {
-		return nil, err
+	var contents []byte
+	if strings.Contains(conf, "*") {
+		// Let caddyfile.doImport logic handle the globbed path
+		contents = []byte("import " + conf)
+	} else {
+		var err error
+		contents, err = ioutil.ReadFile(conf)
+		if err != nil {
+			return nil, err
+		}
 	}
+
 	return caddy.CaddyfileInput{
 		Contents:       contents,
 		Filepath:       conf,


mholt:
Thanks for the change! Nice work.

This has a slightly different behavior than I would expect. If I have 3 separate Caddyfiles and one site defined in each, they may not have curly braces around them, and using import like this change does will cause that to break, because import merges them into a single Caddyfile.

What do you think we should do about this?

francislavoie:
Do we have a good way built-in of determining if the `import`ed string begins with a label? If so then I think just artificially wrapping the rest with `{ }` when importing seems probably fine. It's undefined behavior otherwise, I'd think. The only case I can think of that this would break is if you for some reason in your root `Caddyfile` only have some directives without a label... which I don't think is valid outside of just specifying a couple directives in CLI, unless I'm mistaken.

mholt:
> Do we have a good way built-in of determining if the imported string begins with a label?

In theory, an import statement where a site label is expected should be enough of an indicator whether the contents should contain a site label and thus have curly braces. But the hard part can be knowing where the label ends and the site definition starts. (Labels can span multiple lines, technically. Not my favorite design decision but oh well.)

I proposed this in #550 but I decided to decline it because it didn't seem popular.

francislavoie:
Is there any actual downside to that? I think it's just worth doing, if someone can set it up.

Is there a way to access the list of valid directives (I guess including plugins' attributes)? If so then we could read until we find a directive.

Looking at the docs, it seems like the only valid way to have labels on multiple lines is with `,\h*\n` (comma + zero or more horizontal whitespace + newline), right? Then as soon as you find a newline without a comma, you know you're done with labels. :man_shrugging:

mholt:
>  Then as soon as you find a newline without a comma, you know you're done with labels.

That's true, I think. Yeah.

Whitestrake:
Not even the import docs themselves mention this caveat - honestly, I thought it would be implicit behaviour. Perhaps an update to the import docs and a pointer to those docs from the cli docs?

The implementation of this PR means that any kind of special handling of globbed files, such as reformatting flat Caddyfiles into braced site blocks, would have to be implemented in `caddyfile.doImport` instead.

mholt:
@Whitestrake 

> The implementation of this PR means that any kind of special handling of globbed files, such as reformatting flat Caddyfiles into braced site blocks, would have to be implemented in caddyfile.doImport. Unless we want to handle things differently for -conf globs specifically?

Yes, if I'm reading you right. And that's how it currently works. Basically, we would need to update the docs to mention that the presence of a wildcard `*` performs a glob import under the hood as a convenience.

Fair enough? I think we can merge this.

francislavoie:
@mholt is the sub-discussion we had here actionable? Should it be broken out into a separate issue?

Whitestrake:
In my opinion, if there's agreement that this kind of automatic reformatting should be done, it should be handled in another issue/PR specifically aimed at modifying `doImport` which would affect the `import` directive as well as the `-conf` flag assuming this PR is merged.

mholt:
I agree with @Whitestrake -- I think the sub-discussion here was interesting and probably more productive than what we had in #550 but let us handle that in a separate PR.

I'll merge this soon if we're all on the same page on it!

PR #1953, opened 2017-11-13
Add {user} placeholder to CommonLogFormat

magikstm:
### 1. What does this change do, exactly?
Make code in line with current documentation for CommonLogFormat for {user} placeholder.

I tested with latest version of the plugin https://github.com/tarent/loginsrv.

It works as noted in the documentation https://caddyserver.com/docs/log under the predefined format `{common} (default)`.

I've not added a new test yet.

### 2. Please link to the relevant issues.
#1951

### 3. Which documentation changes (if any) need to be made because of this PR?
None.

### 4. Checklist

- [ ] I have written tests and verified that they fail without my change
- [X] I have squashed any insignificant commits
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I am willing to help maintain this change if there are issues with it later

PR #1978, opened 2017-12-17
Add basicauth {user} value to replacer for logfile.

tobya:

Following on from #1937 and #1931 this change adds {user} value in basicauth to be used by replacer

 - #1937
- #1931
- #1953 

I don't think it requires any but I can add if necessary.

### 4. Checklist

- [x ] I am willing to help maintain this change if there are issues with it later


tobya:
will resubmit

PR #1989, opened 2018-01-02
Correct latency placeholder with lower than 1ms elapsed duration

### LINKED ISSUE #1988: {latency} placeholder should probably return 0ms instead of 0s (for lower than 1ms requests)
0.10.10 (for Windows 7 x64)

Test latency speed using {latency} placeholder.

```
:2015 {
	root F:\caddy2\caddySites
	log / access_log "{remote} | {user} | {when} | {method} | {uri} | {proto} | {status} | {size} | {latency}"
	
	rewrite {
		to {path} {path}/ /maintenance.html /
	}
}
```

Double-click the .exe. (no additional commands)

Some simple navigations and Ctrl-F5 refreshes.

0ms for fast requests.
(other requests are listed in ms such as 1ms, 5ms or 20ms)

It's a minor issue, but I think 0ms requests should be noted as being "0ms" instead of "0s" to use a common base.

It's kinda odd seeing 1ms, 5ms 7ms then 0s (for quick requests).

It's also kinda odd when both the {latency} and {latency_ms} placeholders are used and they return:
latency: 0s
latency_ms: 0

I'm also unsure the microsecond part in this function works:
https://github.com/mholt/caddy/blob/master/caddyhttp/httpserver/replacer.go#L183
(I've not seen microsecond results locally.)

0s for fast requests.

Navigate in a site with low latency locally.

magikstm:
Fixes #1988. I also adjusted some test comments to make them closer to specs.

Tested on Windows 7 x64. I'm intermediate in Go, so it may not be the be the best way to correct the issue.

See above.

None.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -359,7 +359,7 @@ func (r *replacer) getSubstitution(key string) string {
 			return r.emptyValue
 		}
 		elapsedDuration := time.Since(r.responseRecorder.start)
-		
+
 		if elapsedDuration == 0 {
 			return "0ms"
 		}
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -358,7 +358,12 @@ func (r *replacer) getSubstitution(key string) string {
 		if r.responseRecorder == nil {
 			return r.emptyValue
 		}
-		return roundDuration(time.Since(r.responseRecorder.start)).String()
+		elapsedDuration := time.Since(r.responseRecorder.start)
+		
+		if elapsedDuration == 0 {
+			return "0ms"
+		}
+		return roundDuration(elapsedDuration).String()
 	case "{latency_ms}":
 		if r.responseRecorder == nil {
 			return r.emptyValue
--- a/caddyhttp/httpserver/replacer_test.go
+++ b/caddyhttp/httpserver/replacer_test.go
@@ -205,13 +205,13 @@ func TestPathRewrite(t *testing.T) {
 
 func TestRound(t *testing.T) {
 	var tests = map[time.Duration]time.Duration{
-		// 599.935µs -> 560µs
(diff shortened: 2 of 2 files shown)


caddyhttp/httpserver/replacer.go
```diff
@@ -358,7 +358,12 @@ func (r *replacer) getSubstitution(key string) string {
 		if r.responseRecorder == nil {
 			return r.emptyValue
 		}
-		return roundDuration(time.Since(r.responseRecorder.start)).String()
+		elapsedDuration := time.Since(r.responseRecorder.start)
+
+		if elapsedDuration == 0 {
+			return "0ms"
+		}
```
mholt: time.Since returns a duration value, which has accuracy to the nanosecond, I believe. So while 0ms is technically true in that case, the likelihood of getting an actual value of 0 is extremely low, I think. Is there a test that exhibits this behavior?

magikstm: I get 0s durations for about 40% of status 404 and 20% of status 200 on pages with little content.

I'm unsure if this issue affects only Windows 7 x64 or if it affects other Operating Systems. I'm unable to test on others at the moment.

I'm unsure if a test that is valid 100% of the time would be doable. I think it may need more user testing first.

mholt: I do wonder if that's OS-dependent. time.Since should return granularity down to the nanosecond, at least, as far as Go is concerned: we'll have to try to find out if your OS is what's causing a value of 0 to be returned.

mholt:
Thanks for this PR, but after looking it over again, I think it's not doing what we would want it to do. I suspect the reason for the 0 value has something to do with Windows (see linked discussion in a review comment).

In order to go forward with a change like this, a test that reliably reproduces the behavior you're seeing and shows that it fixes it would be necessary, I think.

Again, thank you for your time though! And for being an important part of the project.

PR #1990, opened 2018-01-05
Systemd - disable privatedevices as it may not work for some devices

magikstm:
Sets privatedevices in systemd to 'false' as it may not work for some devices.

I didn't create one yet. It was discussed in the forum recently here:
https://caddy.community/t/systemd-fails-to-load-caddy-on-a-raspberry-pi/3170

It is related to: 
https://github.com/mholt/caddy/issues/1853
https://serverfault.com/questions/871635/error-running-otherwise-working-caddy-under-systemd-v232
https://caddy.community/t/error-running-otherwise-working-caddy-under-systemd-v232/2638

More details on this feature here:
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PrivateDevices=

None.

### 4. Checklist

### DIFF
--- a/dist/init/linux-systemd/caddy.service
+++ b/dist/init/linux-systemd/caddy.service
@@ -30,8 +30,8 @@ LimitNPROC=512
 
 ; Use private /tmp and /var/tmp, which are discarded after caddy stops.
 PrivateTmp=true
-; Use a minimal /dev
-PrivateDevices=true
+; Use a minimal /dev (May bring additional security if switched to 'true', but it may not work on Raspberry Pi's or other devices, so it has been disabled in this dist.)
+PrivateDevices=false
 ; Hide /home, /root, and /run/user. Nobody will steal your SSH-keys.
 ProtectHome=true
 ; Make /usr, /boot, /etc and possibly some more folders read-only.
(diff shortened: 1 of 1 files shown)


francislavoie:
Maybe add a note to the [README](https://github.com/mholt/caddy/blob/master/dist/init/linux-systemd/README.md) with a similar note to your comment, plus a mention that it can be turned on for some other devices for added security?

@mholt this is the first I've seen PrivateDevices, so sorry, can't really give much better of a review 😛

mholt:
Ha, alright, thanks @francislavoie  :smile:

dist/init/linux-systemd/caddy.service
```diff
@@ -30,8 +30,8 @@ LimitNPROC=512
 
 ; Use private /tmp and /var/tmp, which are discarded after caddy stops.
 PrivateTmp=true
-; Use a minimal /dev
-PrivateDevices=true
+; Use a minimal /dev (May bring additional security if switched to 'true', but it may not work on Raspberry Pi's or other devices, so it has been disabled in this dist.)
+PrivateDevices=false
```
mholt: What if this line (and its associated comment) was just removed altogether?

magikstm: I'm unsure all devices set it to "false" by default. This line forces it.

As it adds security, I think it'll make it's possible use easier if left in that file.

PR #1995, opened 2018-01-13
shutdown: allow graceful shutdown for SIGTERM on posix

### LINKED ISSUE #1993: sigtrap only triggers on INT
killing caddy (and coredns) should trigger OnFinalShutdown callbacks, while looking at the code (on why this isn't happening), I see https://github.com/mholt/caddy/blob/master/sigtrap.go#L40 is only trapping on SIGINT. 

Is this intentional? Add SIGTERM here would make a lot of sense of kill sends this by default.

miekg:
The signal is already trapped; make it do the same thing as SIGQUIT to
be more inline with Unix/Linux shutdown expectations.

Fixes #1993

SIGTERM is detailed on various pages/READMEs. I haven't followed up with those yet.

### DIFF
--- a/sigtrap_posix.go
+++ b/sigtrap_posix.go
@@ -25,6 +25,8 @@ import (
 
 // trapSignalsPosix captures POSIX-only signals.
 func trapSignalsPosix() {
+	signalToString := map[os.Signal]string{syscall.SIGTERM: "SIGTERM", syscall.SIGQUIT: "SIGQUIT"}
+
 	go func() {
 		sigchan := make(chan os.Signal, 1)
 		signal.Notify(sigchan, syscall.SIGTERM, syscall.SIGHUP, syscall.SIGQUIT, syscall.SIGUSR1, syscall.SIGUSR2)
@@ -32,18 +34,13 @@ func trapSignalsPosix() {
 		for sig := range sigchan {
 			switch sig {
 			case syscall.SIGTERM:
-				log.Println("[INFO] SIGTERM: Terminating process")
-				if PidFile != "" {
-					os.Remove(PidFile)
-				}
-				os.Exit(0)
-
+				fallthrough
 			case syscall.SIGQUIT:
-				log.Println("[INFO] SIGQUIT: Shutting down")
-				exitCode := executeShutdownCallbacks("SIGQUIT")
+				log.Printf("[INFO] %s: Shutting down", signalToString[sig])
+				exitCode := executeShutdownCallbacks(signalToString[sig])
 				err := Stop()
 				if err != nil {
-					log.Printf("[ERROR] SIGQUIT stop: %v", err)
+					log.Printf("[ERROR] %s stop: %v", signalToString[sig], err)
 					exitCode = 3
 				}
 				if PidFile != "" {
--- a/sigtrap_posix.go
+++ b/sigtrap_posix.go
@@ -25,7 +25,7 @@ import (
 
 // trapSignalsPosix captures POSIX-only signals.
 func trapSignalsPosix() {
-	signalToString := map[os.Signal]string{syscall.SIGTERM: "SIGTERM", syscall.SIGQUIT: "SIGQUIT"}
+	signalToString := map[os.Signal]string{syscall.SIGHUP: "SIGHUP", syscall.SIGUSR1: "SIGUSR1"}
 
 	go func() {
 		sigchan := make(chan os.Signal, 1)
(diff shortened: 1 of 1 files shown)


mholt:
Thanks, I was just about to get around to this. I'm OK with this change overall, but want to make sure there's still a way to forcefully kill the process immediately without shutdown hooks. I guess that's what SIGKILL is for?

miekg:
yes, but SIGKILL is usually used for that, but then nothing is run, i.e. the process is immediately killed.

mholt:
I think I'm OK with that. Basically people should use SIGKILL instead of SIGTERM to immediately kill the process: it just means that the Caddy-managed pidfile remains. (Not sure if that will be a problem...)

Will get back to this later tonight. Thanks again!

mholt:
Instead of having two signals do the same thing, what if we switch around TERM and QUIT?

In other words, from most forceful to most graceful:

- KILL will do what it does best, as usual (unchanged)
- QUIT will clean up the pidfile, but otherwise exit immediately (not graceful on the listeners, and does not execute shutdown callbacks)
- INT will run shutdown callbacks, clean up the pidfile, but NOT stop the server gracefully (exits immediately after the shutdown callbacks and cleaning up the pidfile)
- HUP... should we ignore HUP? Right now it stops the servers gracefully (without shutdown hooks), but... that's never useful, is it?
- TERM will do what you want: run shutdown callbacks, clean up the pidfile, then stop the server gracefully and exit the process
- USR1 and USR2 are reload and upgrade signals, respectively (unchanged)

mholt:
Oops. Mouse jitter.

francislavoie:
I'll be happy if TERM does graceful shutdowns. That's the default signal that Docker sends when stopping containers, so the more graceful the better for TERM, IMO.

miekg:
[ Quoting <notifications@github.com> in "Re: [mholt/caddy] shutdown: allow g..." ]
>Instead of having two signals do the same thing, what if we switch around TERM and QUIT?
>
>In other words, from most forceful to most graceful:
>
>- KILL will do what it does best, as usual (unchanged)
>- QUIT will clean up the pidfile, but otherwise exit immediately (not graceful on the listeners, and does not execute shutdown callbacks)
>- INT will run shutdown callbacks, clean up the pidfile, but NOT stop the server gracefully (exits immediately after the shutdown callbacks and cleaning up the pidfile)
>- HUP... should we ignore HUP? Right now it stops the servers gracefully (without shutdown hooks), but... that's never useful, is it?
>- TERM will do what you want: run shutdown callbacks, clean up the pidfile, then stop the server gracefully and exit the process
>- USR1 and USR2 are reload and upgrade signals, respectively (unchanged)

+1 to the above list, with one exception, HUP is usually also used to reload config.

/Miek

--
Miek Gieben

miekg:
updated PR wrt latest comments.

francislavoie:
I feel like HUP should be the same as TERM, honestly. It means "hang up", which is a shutdown meaning. I don't think HUP really has anything to do with config reloading.

mholt:
One thing that bugs me about SIGHUP is that it tends to be sent by the system arbitrarily? (And not just at logout) -- hence `nohup`. It has bitten me before -- especially on AWS EC2 instances -- where the server shut down because the system sent it a HUP. I am kind of in the camp where we should ignore HUP since, well, having to use `nohup` is a bit annoying.

miekg:
I don't care too much about HUP. But it has been repurposed as reload
config in a multitude of daemons..like Apache, bind and loads of others.

mholt:
K. Thanks, I'll merge this but I might make a few minor changes, preserving the ones you care about. :)

miekg:
Thanks!

sigtrap_posix.go
```diff
@@ -25,58 +25,53 @@ import (
...
 					os.Remove(PidFile)
 				}
 				os.Exit(0)
 
-			case syscall.SIGQUIT:
-				log.Println("[INFO] SIGQUIT: Shutting down")
-				exitCode := executeShutdownCallbacks("SIGQUIT")
+			case syscall.SIGTERM:
+				log.Println("[INFO] SIGTERM: Shutting down")
+				exitCode := executeShutdownCallbacks(signalToString[sig])
```
rdrozhdzh: signalToString map doesn't have syscall.SIGTERM key

miekg:
Nice catch! Want to send pr for this?

sigtrap_posix.go
mholt: Good point, I'll fix this locally and push soon. Thanks!

mholt:
@miekg Fixed in 106d62b0674b582c8fe26a5719378a4261849a94 - and I changed HUP to be ignored, I never did find it useful anyways.

PR #2004, opened 2018-01-24
Update README.md

zottelchin:
Adding the bash command for downloading the caddy.service file from the reposetory. Because it was easy to forget where you find it.

### 4. Checklist


mholt:
Fair enough -- thanks!

Issue #1927, opened 2017-10-18
Encoded forward slash (%2F) in URI not honored.


thedodd:
### 1. What version of Caddy are you using (`caddy -version`)?
Using 0.10.10

### 2. What are you trying to do?
Use the reverse proxy directive to proxy traffic back to the rabbitmq control panel (not AMQP, just standard tcp/http).

### 3. What is your entire Caddyfile?
```text
:80/admin/mq-management/ {
    # Log everything to stdout. Pattern is as follows:
    # {remote} - [{when}] "{method} {upstream} {uri} {proto}" {status} {size} "{>Referer}" "{>User-Agent}"
    log / stdout "{combined}"

    proxy / rabbitmq:15672 {
        transparent
    }
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
`caddy` (defaults to using the Caddyfile mentioned above as it is in the same directory).

### 5. Please paste any relevant HTTP request(s) here.
This fails with `405`:
`curl -XPUT -i http://localhost/admin/mq-management/api/permissions/%2F/guest`

This succeeds with `401` (as in, that is what is expected):
`curl -XPUT -i http://localhost:15672/api/permissions/%2F/guest`

### 6. What did you expect to see?
I expected the behavior between the two curl commands above to be the same & that the proxy would not manipulate the URI.

### 7. What did you see instead (give full error messages and/or log)?
The curl command which goes through the proxy appears to be collapsing the forward slashes. 

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
Using docker (this is obviously the easiest way):
- spin up a rabbitmq instance named `rabbitmq`. Expose the needed ports (`15672` mainly).
- spin up a container with caddy 0.10.10 installed with the caddy file above in the CWD; start the container with command `caddy` will read the Caddyfile above as long as it is in the same directory.

**NOTE:** looks like there have been a **decent number of issues reported about this already.** I've pored over them, none of them appear to present a fix.

lirener:
Although I know it is a question and the suggestion,But！（虽然我知道这是一个问题与建议,但是！）

Thank you very much!Finally let me find the regarding print log output to the CMD command line writing I have seen a lot of tutorials didn't mention it...【log / stdout "{combined}"】
（谢谢！终于让我找到了 关于打印日志输出到cmd命令行的写法了 我看过很多教程 都没提到它……【log / stdout "{combined}"】）
[...]

PR #2007, opened 2018-01-29
Fix for #1927

tobya:
This is a fix for #1927 .  Ensures that encoded / (%2f) in url is retained when the site definition includes a directory.

The problem turned out to be a call to strings.TrimPrefix which also decoded any encoded values in the string.  It has been replaced by code to simply perform the trim.
#1927 

None

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -413,7 +413,12 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 	// the URL path, so a request to example.com/foo/blog on the site
 	// defined as example.com/foo appears as /blog instead of /foo/blog.
 	if pathPrefix != "/" {
-		r.URL.Path = strings.TrimPrefix(r.URL.Path, pathPrefix)
+		// We are unable to use strings.TrimPrefix to trim the pathPrefix as
+		// it will decode any url encoded characters. See #1927
+		idx := strings.Index(pathPrefix, r.URL.Path)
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/server.go
```diff
@@ -413,7 +413,12 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 	// the URL path, so a request to example.com/foo/blog on the site
 	// defined as example.com/foo appears as /blog instead of /foo/blog.
 	if pathPrefix != "/" {
-		r.URL.Path = strings.TrimPrefix(r.URL.Path, pathPrefix)
+		// We are unable to use strings.TrimPrefix to trim the pathPrefix as
+		// it will decode any url encoded characters. See #1927
+		idx := strings.Index(pathPrefix, r.URL.Path)
+		if idx > -1 {
```
tobya: This may be unnecessary.  What should I do if idx == -1?  Can we be confident it never will?

mholt:
That's so interesting if this really works like you say. Do you think you could add a test? You'd probably have to write a new Test function but it could be simple. I'd like to verify this behavior.

tobya:
I’ll take a look and see if I can create a new test.

tobya:
Sorry the PR is a complete mess, I'll update when I'm sure I have the correct behaviour.

mholt:
No problem! Thank you for working on it! I will be happy to take another look when it is tested. I'm sure it'll clear things right up.

Issue #1974, opened 2017-12-13
Change EventHooks to sync.map 


Zeymo:
when customer plugin `RegisterEventHook ` with new events name and trigger this hook with posix signal(e.g.  `SIGUSR2`)
https://github.com/mholt/caddy/blob/97710ced7ea5b5d8025dc6a3e43d82e9a280920e/plugins.go#L254
and
https://github.com/mholt/caddy/blob/97710ced7ea5b5d8025dc6a3e43d82e9a280920e/plugins.go#L261
will cause data race warning ,so must change eventHooks to `sync.map`

mholt:
Thanks for the heads up. Have you observed this race using the race detector?

Zeymo:
@mholt yes with `go tool race` , I will paste print later if necessary

mholt:
That would be great if you could. It would be nice to have for completeness sake. Thank you!

Zeymo:
```
WARNING: DATA RACE
Read at 0x00c4200710b0 by goroutine 12:
  runtime.mapiterinit()
      /xxx/go/src/runtime/hashmap.go:709 +0x0
  mwcs/framework.EmitEvent()
      /xxx/caddy/plugins.go:261 +0x9d
  mwcs/caddy.trapSignalsPosix.func1()
     /xxx/caddy/sigtrap_posix.go:90 +0x2a8

Previous write at 0x00c4200710b0 by main goroutine:
  runtime.mapassign_faststr()
     /xxx/go/src/runtime/hashmap_fast.go:598 +0x0
  mwcs/framework.RegisterEventHook()
      /xxx/caddy/plugins.go:254 +0xcc
  mwcs/http/configserver.setup.func1()
      /xxx/configserver/setup.go:128 +0x8ae
  mwcs/framework.startWithListenerFds()
      /xxx/caddy.go:307 +0x17d
  mwcs/framework.Start()
     /xxx/caddy.go:270 +0x179
  mwcs/mwcsmain.Run()
     /xxx/caddymain/run.go:90 +0x3b9
  main.main()
      /xxx/caddy/main.go:8 +0x46
```

**I customize caddy so I change  Line.No not match to caddy master**

sigtrap_posix.go:90
```
func trapSignalsPosix() {
	go func() {
	       sigchan := make(chan os.Signal, 1)
[...]

PR #2009, opened 2018-01-30
make eventHooks thread safe

tw4452852:

Use `sync.Map` to make `eventHooks` concurrent r/w safe

#1974 

N/A

### 4. Checklist

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"net"
 	"sort"
+	"sync"
 
 	"github.com/mholt/caddy/caddyfile"
 )
@@ -38,7 +39,7 @@ var (
 
 	// eventHooks is a map of hook name to Hook. All hooks plugins
 	// must have a name.
-	eventHooks = make(map[string]EventHook)
+	eventHooks = sync.Map{}
 
 	// parsingCallbacks maps server type to map of directive
 	// to list of callback functions. These aren't really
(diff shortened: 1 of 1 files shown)


tw4452852:
Hi @mholt,

I think the main advantage of `sync.Map` is that it doesn't incur the blocking while `sync.Mutex` does.
Because `sync.Map` use atomic operation underneath. In the case of #1974 , the signal handle may wait  until the plugin initializes done if we use a mutex.

mholt:
For the record: This change will require Caddy to be built using Go 1.9 instead of 1.8. Which I suppose is fine, with Go 1.10 around the corner.

plugins.go
```diff
@@ -38,7 +39,7 @@ var (
 
 	// eventHooks is a map of hook name to Hook. All hooks plugins
 	// must have a name.
-	eventHooks = make(map[string]EventHook)
+	eventHooks = sync.Map{}
```
mholt: Should this be a pointer 🤔 ? I know we don't pass this value around, but [all the methods are defined on `*sync.Map`](https://golang.org/pkg/sync/#Map). (I know it says the zero value is valid, but I just want to double-check.)

tw4452852: No, this isn't necessary, please check this:
https://stackoverflow.com/a/47825590

mholt:
Thanks as always, Tw!

PR #2014, opened 2018-02-03
Ensure %2f is not converted into / for hosts with directories.  Fix for #1927

tobya:
This is a fix for #1927 

If a host is declared to include a directory, any %2f in the url will be lost by being converted to `/`.  This change ensures they are not.

I have put some inline comments in my code for some things I was unsure about.

It is a known issue of the URL type that Path cannot distinguish between a `/` or `%2f` . 
https://golang.org/pkg/net/url/#URL

#1927 

N/A
### 4. Checklist


caddyhttp/httpserver/server.go
```diff
@@ -413,7 +413,12 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 	// the URL path, so a request to example.com/foo/blog on the site
 	// defined as example.com/foo appears as /blog instead of /foo/blog.
 	if pathPrefix != "/" {
-		r.URL.Path = strings.TrimPrefix(r.URL.Path, pathPrefix)
+		// We need to use URL.EscapedPath() when trimming the pathPrefix as
+		// URL.Path is ambiguous about / or %2f - see docs. See #1927
+		r.URL, err = r.URL.Parse(strings.TrimPrefix(r.URL.EscapedPath(), pathPrefix))
```
tobya: I have to reparse here as when I tried to simply assign to r.URL.Path I got unexpected results

caddyhttp/httpserver/server.go
```diff
@@ -413,7 +413,12 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
+		if err != nil {
```
tobya: hmmm.  Do I need to check for err here or can I ignore it.  The likelyhood of the URL not being able to be parsed at this point is very minimal. 

Should I remove this check and error?

mholt:
Thanks Toby. We have to be very careful here in this part of the code. I will try to test it out and give it a more thorough review as soon as I push up my changes to the caddytls package.

caddyhttp/httpserver/server.go
abiosoft: I think golint reports unchecked errors, you can leave it like this.

elcore: Better safe than sorry :)

tobya:
This pr now passes all the tests I wrote. 

Calling url.Parse rather than u.Parse fixed the issue with /// at the beginning of a url being collapsed to /

As @mholt said this is a place to be careful so would be great if it could be reviewed @abiola @elcore

caddyhttp/httpserver/server.go
```diff
@@ -413,15 +413,26 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
...
+	// We need to use URL.EscapedPath() when trimming the pathPrefix as
+	// URL.Path is ambiguous about / or %2f - see docs. See #1927
+	trimmed := strings.TrimPrefix(u.EscapedPath(), prefix)
+	if !strings.HasPrefix(trimmed, "/") {
+		trimmed = "/" + trimmed
+	}
+	u, err := url.Parse(trimmed)
+	if err != nil {
+		log.Printf("[ERROR] unable to parse url: %s, %v", u.Path, err)
+	}
```
mholt: Maybe name the return value of `url.Parse()` something else, like `newURL` or `trimmedURL` and then return `u` if there is an error here. Also, change error message to: `"[ERROR] Unable to parse trimmed URL %s: %v", trimmed,  err`

mholt:
Oh, and if you could resolve the merge conflict too, that'd be great. (Thanks, sorry for the trouble!) :+1:

tobya:
@mholt Made that change and added a couple more tests for good measure.

Issue #1918, opened 2017-10-13
Consul KV and Vault for TLS certificate storage


Gufran:
This issue is a feature proposal, discussion thread, and progress tracker for said feature.

### 2. What are you trying to do?

I propose two new storage backends for SSL certificate:

 - [Consul][] KV
 - [Vault][]

currently Caddy supports only disk or in-memory storage, both of these are far from ideal in an
environment with hundreds of certificates being served by redundant machines, SaaS businesses are a good example here. It is easy to spam the certificate issuer if each of those machine has its own storage. There are ways to work around this limitation with shared volumes but at the very least it is inconvenient and prone to human errors or inconsistent configuration.

There are also community supported plugins available but they have their own development lifecycle, decisions around feature and behaviour, and usual obligations for the maintainers that come with an open source project.

Native support for these storages will tie previously mentioned loose ends and declare the standards.

### Proposed changes

`tls` block has an undocumented directive `storage` which lets one change the storage backend. This doc proposes a backwards compatible change to `storage` directive as follows:

```text
tls {
  storage [name] {
    key value
    ...
  }
}
```

The content of the block will be a map which will be passed down to the storage backend as it is.
`caddytls.Storage` interface will be updated to provide a new method:

``` go
ValidateConfig(map[string]string) error
```

which will be called as part of setup to validate the additional configuration provided to `storage` directive and can raise an error if the configuration is not understood by the storage driver.

`caddytls.StorageConstructor` will be updated to accept the configuration as second parameter:

```go
type StorageConstructor func(caURL *url.URL, config map[string]string) (Storage, error)
```

These changes together will provide scaffolding necessary for more complex storage backends than
disk or in-memory.

### Consul KV
Consul KV storage backend is the simpler one. Possible configuration options are:

| Attribute | Possible Value | Required | Description |
|-----------|----------------|----------|-------------|
| Host | valid URL | yes | Consul Host URL including the protocol and port |
| Prefix | any string | yes | KV path prefix under which the certificates will be stored |
| Token | any string | no | Consul ACL token with permissions to read and write under `prefix` |
| Datacenter | any string | no | Consul datacenter |
| Auth | tuple (username, password) | no | basic auth credentials for Consul client |
| TLS  | tuple (certificate, key) or `trust` or `off` | no | TLS configuration. On by default |
| AES Key | 32 characters long string | no | used to encrypt the contents. contents will be stored in plain text if not provided |

[...]

PR #2015, opened 2018-02-04
tls: Restructure and improve certificate management

### LINKED ISSUE #1303: Caddy serves wrong SSL cert for site that is not served on HTTPS port

```
Caddy 0.9.3
```

I want to have multiple virtual hosts, some http**s**, some http only. I want the http only hosts to not respond to http**s** requests with a wrong certificate.

```text
https://foo.example.com {
  tls self_signed
  root /html/foo
}

http://bar.example.com {
  tls off
  root /html/bar
}
```

I created the reproducible example from above on my Mac (`10.11.6`) but I am having this problem in production on `Ubuntu 16.04.1 LTS ` so I don’t think it is OS related. The working directory contains the `Caddyfile` from above.

```
$ sudo caddy
```

Caddy starts fine and serves both vhosts, one with https and the other with http. I expect Caddy to **not respond at all** when I try to access `https://bar.example.com` as that virtual hosts is explicitly configured as http, not http**s**.

When accessing `https://bar.example.com` (the host which is explicitly http), Caddy sends the cert for `foo.example.com`, which results in an SSL error. (I know that in this example it always results in an error because the cert is self signed, but in production this would serve a trusted cert from Let’s Encrypt).

When ignoring the warning, Caddy responds with `No such site at :443`. While this is technically correct, Caddy should not accept the connection in the first place.

In addition to the SSL error, this also **leaks information** about at least one other site (the one Caddy takes the certificate from) that is set up on the host.

1. Use the `Caddyfile` from step 3.
2. Update the `root` directive so that Caddy finds some files to serve
3. Edit your `/etc/hosts` to include the following
4. Start Caddy as described in step 4.

```
# /etc/hosts
127.0.0.1 foo.example.com
127.0.0.1 bar.example.com
```

---

Thanks so much for any responses. I’m sorry if there is already an issue for this or if this is a known bug, I couldn’t find anything related to this.

### LINKED ISSUE #1991: Wrong certificate when mixing on-demand and manual TLS

```
Caddy 0.10.10
```

Serving both a manual certificate and a Let's Encrypt certificate on two different domains.

```text
whoami.filippo.io {
	tls bip@filippo.io
	redir https://blog.filippo.io/ssh-whoami-filippo-io/ 302
}

https://cve-2016-2107.filippo.io {
	realip cloudflare
	tls cert.pem key.pem {
		protocols tls1.2 tls1.2
		ciphers ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256
		# clients origin-pull-ca.pem
	}
	proxy / localhost:31337
}

https://ticketbleed.filippo.io {
	realip cloudflare
	tls cert.pem key.pem {
		protocols tls1.2 tls1.2
		ciphers ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256
		# clients origin-pull-ca.pem
	}
	proxy / localhost:41337
}
```

```
cd /etc/caddy
ulimit -n 4096
setcap cap_net_bind_service=+ep /usr/local/bin/caddy
exec setuidgid www-data env HOME=/etc/caddy /usr/local/bin/caddy \
	-agree -pidfile /var/run/caddy.pid
```

Let's Encrypt on whoami.filippo.io, my custom certificate on ticketbleed.filippo.io and cve-2016-2107.filippo.io.

The custom certificate is served on whoami.filippo.io.

Even if I deleted the ACME certificate, Caddy obtained a new (correct) one and placed it at `/etc/caddy/.caddy/acme/acme-v01.api.letsencrypt.org/sites/whoami.filippo.io/whoami.filippo.io.crt` while starting. However it then doesn't use it.

This is a regression, the previous version (alas lost to the wind) was working correctly.

### LINKED ISSUE #1994: old certificate is used after switching to on-demand let's encrypt
Unfortunately this issue is hard to track and I could find a way to reproduce it on my test systems, yet.
I hope by describing the issue you'll find a reason by checking the code.
It might be solved by the solution discussed in #1991.

Caddy 0.10.10

Switch certificate of a domain from third-party to on-demand let's encrypt.

* Caddyfile
```text
http://*.eu-central-1.compute.amazonaws.com http://www.www.www.* {
    status 403 /
    ratelimit 1 2 hour
    import /etc/caddy/load.d/log
    import /etc/caddy/load.d/ipfilter
}

https://*.eu-central-1.compute.amazonaws.com https://www.www.www.* {
    status 403 /
    ratelimit 1 2 hour
    tls /etc/ssl/default.crt.pem /etc/ssl/default.key.pem
    import /etc/caddy/load.d/hsts
    import /etc/caddy/load.d/log
    import /etc/caddy/load.d/ipfilter
}

http:// {
    redir / {scheme}://www.{host}{uri} 301
    import /etc/caddy/load.d/log
    import /etc/caddy/load.d/ratelimit
    import /etc/caddy/load.d/ipfilter
    tls off
}
https:// {
    redir / {scheme}://www.{host}{uri} 301
    import /etc/caddy/load.d/log
    import /etc/caddy/load.d/letsencrypt_default
    import /etc/caddy/load.d/hsts
    import /etc/caddy/load.d/ratelimit
    import /etc/caddy/load.d/ipfilter
}

import /etc/caddy/conf.d/*.conf
```
* Example for conf.d (currently we have ~280 files):
```text
http://*.example.com http://example.com {
    redir 301 {
        / {scheme}://www.{host}{uri}
    }
    import /etc/caddy/load.d/log
    import /etc/caddy/load.d/ratelimit
    import /etc/caddy/load.d/ipfilter
    tls off
}
https://*.example.com https://example.com {
    redir 301 {
        / {scheme}://www.{host}{uri}
    }

    import /etc/caddy/load.d/log
    import /etc/caddy/load.d/letsencrypt_profile1
    import /etc/caddy/load.d/hsts
    import /etc/caddy/load.d/ratelimit
    import /etc/caddy/load.d/ipfilter
}
```
* log:
```text
request_id
log / syslog "({request_id}) {combined} {host} {latency_ms}"
errors syslog
```
* hsts:
```text
header / Strict-Transport-Security "max-age=31536000; includeSubDomains";
```
* letsencrypt_default:
```text
tls "foo+letsencrypt-default@example.com" {
    max_certs 10
    key_type rsa4096
}
```
* letsencrypt_profile1:
```text
tls "foo+letsencrypt-profile1@example.com" {
    max_certs 10
    key_type rsa4096
}
```
* ipfilter:
```text
# currently empty
```
* ratelimit:
```text
#ratelimit / 300 400 minute {
#    whitelist 128.199.222.65/32
#}
```

I just provided a short part of the configuration as all would result in 31026 lines of configuration.
Most of it is vhost configuration which just differ in the matching hostname.
'ratelimit' is deactivated because of https://github.com/xuqingfeng/caddy-rate-limit/issues/23

Started using the sysv-init script extended by me with PR #1984 .

* started as www-data
* CADDYPATH=/etc/ssl/caddy/
```text
/usr/local/bin/caddy -agree=true -log=/var/log/caddy.log -conf=/etc/caddy/Caddyfile -disable-tls-sni-challenge
```

see 8.

After reloading Caddy it should start generating a let's encrypt certificate on the first request made to the domain.

It delivers the old certificate until you switch to not-on-demand mode, reload Caddy and a certificate is generated.
The behaviour does not depend on the validity of the old certificate. It happens with invalid and with valid certificates and certificate chains.

Thiis is the hard part.
With a minimal config using just one or a few domains Caddy behaves as expected.
We have 986 domains with https configuration. 
316 are using let's encrypt configuration.
233 are using on-demand let's encrypt.

Following the steps I can use to reproduce it on our production system:

1. curl with given certificate:
```text
$ curl -Ikv https://example.com
* Rebuilt URL to: https://example.com/
*   Trying 52.XX.XX.XX...
* TCP_NODELAY set
* Connected to example.com (52.XX.XX.XX) port 443 (#0)
{...}
* Server certificate:
*  subject: C=DE; ST=Saxony; O=root360; OU=testing; CN=*.*; emailAddress=admin@*.*
*  start date: Sep  7 10:41:55 2015 GMT
*  expire date: Sep  4 10:41:55 2025 GMT
*  issuer: C=DE; ST=Saxony; L=Leipzig; O=root360; OU=testing; CN=*.*; emailAddress=admin@*.*
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
{...}
> HEAD / HTTP/2
> Host: example.com
> User-Agent: curl/7.57.0
> Accept: */*
> 
{ [5 bytes data]
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
} [5 bytes data]
< HTTP/2 301 
< location: https://www.example.com/
< server: Caddy
< strict-transport-security: max-age=31536000; includeSubDomains
< content-type: text/plain; charset=utf-8
< date: Fri, 12 Jan 2018 09:08:01 GMT
< 
* Connection #0 to host example.com left intact
HTTP/2 301 
location: https://www.example.com/
server: Caddy
strict-transport-security: max-age=31536000; includeSubDomains
content-type: text/plain; charset=utf-8
date: Fri, 12 Jan 2018 09:08:01 GMT
```
2. switch tls config to use on-demand let's encrypt
```text
tls test+letsencrypt@example.com {
  max_certs 10
}
```
4. reload caddy
5. curl after switching to on-demand let's encrypt
```text
$ curl -Ikv https://example.com
* Rebuilt URL to: https://example.com/
*   Trying 52.XX.XX.XX...
* TCP_NODELAY set
* Connected to example.com (52.XX.XX.XX) port 443 (#0)
{...}
* Server certificate:
*  subject: C=DE; ST=Saxony; O=root360; OU=testing; CN=*.*; emailAddress=admin@*.*
*  start date: Sep  7 10:41:55 2015 GMT
*  expire date: Sep  4 10:41:55 2025 GMT
*  issuer: C=DE; ST=Saxony; L=Leipzig; O=root360; OU=testing; CN=*.*; emailAddress=admin@*.*
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
{...}
> HEAD / HTTP/2
> Host: example.com
> User-Agent: curl/7.57.0
> Accept: */*
> 
{ [5 bytes data]
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
} [5 bytes data]
< HTTP/2 301 
< location: https://www.example.com/
< server: Caddy
< strict-transport-security: max-age=31536000; includeSubDomains
< content-type: text/plain; charset=utf-8
< date: Fri, 12 Jan 2018 09:11:09 GMT
< 
* Connection #0 to host example.com left intact
HTTP/2 301 
location: https://www.example.com/
server: Caddy
strict-transport-security: max-age=31536000; includeSubDomains
content-type: text/plain; charset=utf-8
date: Fri, 12 Jan 2018 09:11:09 GMT
```
6. restart caddy (just in case):
7. curl after restart:
```text
$ curl -Ikv https://example.com
* Rebuilt URL to: https://example.com/
*   Trying 52.XX.XX.XX...
* TCP_NODELAY set
* Connected to example.com (52.XX.XX.XX) port 443 (#0)
{...}
* Server certificate:
*  subject: C=DE; ST=Saxony; O=root360; OU=testing; CN=*.*; emailAddress=admin@*.*
*  start date: Sep  7 10:41:55 2015 GMT
*  expire date: Sep  4 10:41:55 2025 GMT
*  issuer: C=DE; ST=Saxony; L=Leipzig; O=root360; OU=testing; CN=*.*; emailAddress=admin@*.*
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
{...}
> HEAD / HTTP/2
> Host: example.com
> User-Agent: curl/7.57.0
> Accept: */*
> 
{ [5 bytes data]
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
} [5 bytes data]
< HTTP/2 301 
< location: https://www.example.com/
< server: Caddy
< strict-transport-security: max-age=31536000; includeSubDomains
< content-type: text/plain; charset=utf-8
< date: Fri, 12 Jan 2018 09:14:36 GMT
< 
* Connection #0 to host example.com left intact
HTTP/2 301 
location: https://www.example.com/
server: Caddy
strict-transport-security: max-age=31536000; includeSubDomains
content-type: text/plain; charset=utf-8
date: Fri, 12 Jan 2018 09:14:36 GMT
```
8. switch tls config to use let's encrypt without on-demand mode
```text
tls test+letsencrypt@example.com
```
9. reload caddy, certificate is requested from let's encrypt
```text
2018/01/12 09:25:12 [INFO][example.com] acme: Obtaining bundled SAN certificate                                                           
2018/01/12 09:25:13 [INFO][example.com] AuthURL: https://acme-v01.api.letsencrypt.org/acme/authz/4UwvWG1hCecjtpTRCsGVaP1yp0t90sgUKfbrJGFer-I
2018/01/12 09:25:13 [INFO][example.com] acme: Trying to solve HTTP-01                                                                     
2018/01/12 09:25:14 [INFO][example.com] Served key authentication                                                                         
2018/01/12 09:25:16 [INFO][example.com] The server validated our request                                                                  
2018/01/12 09:25:16 [INFO][example.com] acme: Validations succeeded; requesting certificates                                              
2018/01/12 09:25:17 [INFO] acme: Requesting issuer cert from https://acme-v01.api.letsencrypt.org/acme/issuer-cert                        
2018/01/12 09:25:17 [INFO][example.com] Server responded with a certificate.                                                               
2018/01/12 09:25:17 [INFO][example.com] Certificate written to disk: /etc/ssl/caddy/acme/acme-v01.api.letsencrypt.org/sites/example.com/example.com.crt
```
10. curl with not-on-demand mode of let's encrypt
```text
$ curl -Ikv https://example.com
* Rebuilt URL to: https://example.com/
*   Trying 52.XX.XX.XX...
* TCP_NODELAY set
* Connected to example.com (52.XX.XX.XX) port 443 (#0)
{...}
* Server certificate:
*  subject: CN=example.com
*  start date: Jan 12 08:25:17 2018 GMT
*  expire date: Apr 12 08:25:17 2018 GMT
*  issuer: C=US; O=Let's Encrypt; CN=Let's Encrypt Authority X3
*  SSL certificate verify ok.
{...}
> HEAD / HTTP/2
> Host: example.com
> User-Agent: curl/7.57.0
> Accept: */*
> 
{ [5 bytes data]
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
} [5 bytes data]
< HTTP/2 301 
< conf: example.com
< location: https://www.example.com/
< server: Caddy
< strict-transport-security: max-age=31536000; includeSubDomains
< content-type: text/plain; charset=utf-8
< date: Fri, 12 Jan 2018 09:28:34 GMT
< 
* Connection #0 to host example.com left intact
HTTP/2 301 
conf: example.com
location: https://www.example.com/
server: Caddy
strict-transport-security: max-age=31536000; includeSubDomains
content-type: text/plain; charset=utf-8
date: Fri, 12 Jan 2018 09:28:34 GMT
```

mholt:
**Please help test and review this PR!**

Most of the changes are internal to the code, but there are bug fixes and a couple minor breaking changes for server type plugins that use the `caddytls` package.

- Expose the list of Caddy instances through caddy.Instances()

- Added arbitrary storage to caddy.Instance

- The cache of loaded certificates is no longer global; now scoped
  per-instance, meaning upon reload (like SIGUSR1) the old cert cache
  will be discarded entirely, whereas before, aggressively reloading
  config that added and removed lots of sites would cause unnecessary
  build-up in the cache over time.

- Key certificates in the cache by their SHA-256 hash instead of
  by their names. This means certificates will not be duplicated in
  memory (within each instance), making Caddy much more memory-efficient
  for large-scale deployments with thousands of sites sharing certs.

- Perform name-to-certificate lookups scoped per caddytls.Config instead
  of a single global lookup. This prevents certificates from stepping on
  each other when they overlap in their names.

- Do not allow TLS configurations keyed by the same hostname to be
  different; this now throws an error.

- Updated relevant tests, with a stark awareness that more tests are
  needed.

- Change the NewContext function signature to include an *Instance.

- Strongly recommend (basically require) use of caddytls.NewConfig()
  to create a new *caddytls.Config, to ensure pointers to the instance
  certificate cache are initialized properly.

- Update the TLS-SNI challenge solver (even though TLS-SNI is disabled
  currently on the CA side). Store temporary challenge cert in instance
  cache, but do so directly by the ACME challenge name, not the hash.
  Modified the getCertificate function to check the cache directly for
  a name match if one isn't found otherwise. This will allow any
  caddytls.Config to be able to help solve a TLS-SNI challenge, with one
  extra side-effect that might actually be kind of interesting (and
  useless): clients could send a certificate's hash as the SNI and
  Caddy would be able to serve that certificate for the handshake.

- Do not attempt to match a "default" (random) certificate when SNI
  is present but unrecognized; return no certificate so a TLS alert
  happens instead.

- Store an Instance in the list of instances even while the instance
  is still starting up (this allows access to the cert cache for
  performing renewals at startup, etc). Will be removed from list again
  if instance startup fails.

- Laid groundwork for ACMEv2 and Let's Encrypt wildcard support.

Server type plugins will need to be updated slightly to accommodate
minor adjustments to their API (like passing in an Instance). This
commit includes the changes for the HTTP server.

Certain Caddyfile configurations might error out with this change, if
they configured different TLS settings for the same hostname.

This change trades some complexity for other complexity, but ultimately
this new complexity is more correct and robust than earlier logic.

Fixes #1991
Fixes #1994
Fixes #1303 

The `tls` directive page will definitely need updating. TLS configurations for equal hostnames cannot be different. Any description about how certificates are stored internally might need to change.

OH, and something I forgot on the original commit message: Caddy can now better support being used in conjunction with other Caddy instances that share the same certificate storage. In other words, if you have multiple instances all using the same $CADDYPATH, that's been a bad idea because they will all try to manage the certificates themselves (keep them renewed, etc, which runs you into rate limits fast). Now, Caddy will attempt to check if the certificate in storage still needs to be renewed just before it fires off a renewal. So if you stagger Caddy deployments by more time than it takes to complete an ACME challenge (usually just a few seconds, but if you use DNS challenge maybe minutes or more), they might play better together in an acceptable way, for small-medium scale use cases. I wouldn't rely on this for high-volume deployments (like, hundreds or thousands of instances sharing the same certificate storage and all trying to manage them) but I bet it'll be better for dozens of Caddy instances!

** **UPDATE** **: The latest changes I pushed to this branch now eliminate the need to stagger the timing of running each Caddy instance. If multiple Caddy instances (even thousands) share the same certificate storage on disk, they will coordinate renewals and not step on each others' toes. However, testing this is a little tricky, so we'll have to see how it goes. It's an advanced use case, so I'm not expecting many people to be benefitted by it, but I hope they will let us know how it goes!

### 4. Checklist

/cc @pieterlouw and @miekg -- I'll be getting in touch with you soon with more details for what this means about your server type plugins. I can probably submit a PR to caddy-net, and I'm not sure that CoreDNS will have to change much except a line where `NewContext()` is invoked.


francislavoie:
I had a few questions/ideas as to what this means for Caddy in the future. I'm not really reviewing the code, so I'm sorry if I'm asking things that are painfully obvious by looking at the code.

>- Added arbitrary storage to caddy.Instance

This means we'll now have support for plugins adding their own storage drivers, correct? E.g. Consul (see #1918) - for situations where Caddy is a load balancer, there needs to be a way to ensure instances don't conflict with eachother when attempting to renew. Which brings me to my next question:

>Caddy can now better support being used in conjunction with other Caddy instances that share the same certificate storage. In other words, if you have multiple instances all using the same $CADDYPATH, that's been a bad idea because they will all try to manage the certificates themselves (keep them renewed, etc, which runs you into rate limits fast). Now, Caddy will attempt to check if the certificate in storage still needs to be renewed just before it fires off a renewal. So if you stagger Caddy deployments by more time than it takes to complete an ACME challenge (usually just a few seconds, but if you use DNS challenge maybe minutes or more), they might play better together in an acceptable way, for small-medium scale use cases. I wouldn't rely on this for high-volume deployments (like, hundreds or thousands of instances sharing the same certificate storage and all trying to manage them) but I bet it'll be better for dozens of Caddy instances!

So this is kinda part two of the requirements in #1918, locking. This PR doesn't implement locking for renewals, based on your description, right? For that usecase you described, it sounds like a lock file would be ideal (that could contain a timestamp as well to indicate the last time renewal was attempted) to alert other instances that renewal is already happening, and those other instances could give up. Maybe that's out of scope for this PR though. A lock interface would help for implementing Consul or other storage drivers.

> - The cache of loaded certificates is no longer global; now scoped
per-instance

Maybe I'm getting confused by the terminology here, but does `instance` here mean individual running instances of Caddy or does it mean something smaller/internal (like an instance per site)? If the former, my understanding is that the old behaviour was a single cache for every instance running on the same machine? If the later, then was the old behaviour that each instance had its own cache (this seems like the ideal goal, so I'm leaning towards the former being the meaning)?

I ask that, because:

>- Expose the list of Caddy instances through caddy.Instances()

So does each running instance now know about the other instances on the same machine? If so, neat!

> The tls directive page will definitely need updating. TLS configurations for equal hostnames cannot be different. Any description about how certificates are stored internally might need to change.

This sounds like a pretty good case for recommending using the new snippet syntax (#1921) to avoid duplicated TLS blocks.

mholt:
@francislavoie Great thoughts, let me respond to each one:

> This means we'll now have support for plugins adding their own storage drivers, correct? E.g. Consul (see #1918) - for situations where Caddy is a load balancer, there needs to be a way to ensure instances don't conflict with eachother when attempting to renew

No, sorry for the confusion. This storage is simply a `map[interface{}]interface{}` that allows plugins to store things scoped to an Instance, instead of globally. This makes for better correctness and efficiency with regards to reloads (SIGUSR1, etc). Our certificate cache is stored in Instance storage instead of globals now. This is not the same as TLS storage, which I'm still thinking of how to develop it fully.

>  So this is kinda part two of the requirements in #1918, locking. This PR doesn't implement locking for renewals, based on your description, right? For that usecase you described, it sounds like a lock file would be ideal (that could contain a timestamp as well to indicate the last time renewal was attempted) to alert other instances that renewal is already happening, and those other instances could give up. Maybe that's out of scope for this PR though. A lock interface would help for implementing Consul or other storage drivers.

Right: this PR does not totally solve the issue of shared, managed certificates among multiple Caddy instances. However, it should make any problem virtually go away as long as the instances are staggered in their execution by a few seconds to a few minutes, and as long as there aren't too many of them (hundreds or more might be too many to be safe). The TLS storage stuff I am working on in my head should introduce locking and do it properly, but those will be plugins. This solution is baked right into Caddy and should "just work" for the vast majority of people who need something like it.

> Maybe I'm getting confused by the terminology here, but does instance here mean individual running instances of Caddy or does it mean something smaller/internal (like an instance per site)? If the former, my understanding is that the old behaviour was a single cache for every instance running on the same machine? If the later, then was the old behaviour that each instance had its own cache (this seems like the ideal goal, so I'm leaning towards the former being the meaning)?

Each Caddy process instantiates a literal `*Instance` value which holds all the "servers" it is running. A single Caddy process can run more than 1 instance, for example momentarily during graceful reloads, or if being used as a library. This is uncommon but still possible, and scoping things to an Instance helps us avoid global variables and keep things more correct and efficient. Before, there was a single certificate cache for the whole PROCESS, meaning reloading the config and changing sites didn't always flush the certificate cache properly. Now, each Instance value has its own cache, so a reload will just dispose of the old cache.

> So does each running instance now know about the other instances on the same machine? If so, neat!

In the same process, yes. If they have to know about them, they can. The updated certificate maintenance routines run one-per-process and have to iterate the list of Instances in order to access their individual certificate caches. Then they iterate the list of certificates and perform renewals, etc.

> This sounds like a pretty good case for recommending using the new snippet syntax (#1921) to avoid duplicated TLS blocks.

Good point.

Thanks for your comments! I hope this makes a little more sense!

francislavoie:
Ah, yeah that helps. I got a little excited that this would cover #1918 because there's a few scenarios I ran into that it would really help with, i.e. load balancing your load balancers, or having certs managed by the underlying instances that are being load balanced. That's currently not possible with Caddy because they don't share certs - if one renews, the other won't know about it unless they share the same filesystem.

nodesocket:
> a lock file would be ideal (that could contain a timestamp as well to indicate the last time renewal was attempted) to alert other instances that renewal is already happening, and those other instances would wait until the lock is removed.

👍 

That also means that incoming requests would be blocked until the TLS renew lock is removed right?

mholt:
@nodesocket 

> > a lock file would be ideal (that could contain a timestamp as well to indicate the last time renewal was attempted) to alert other instances that renewal is already happening, and those other instances could give up.
>
> 👍
> That also means that incoming requests would be blocked until the TLS renew lock is removed right?

I wouldn't go too far with the quoted suggestion (no offense Francis, ha) -- what we'd rather have is a truly atomic "create-file-if-not-exists" operation, provided by the OS. Right now, Caddy does not utilize any such features, which is why cert management, even with this PR, is not truly synchronized.

And no, incoming requests are NOT blocked while certificate maintenance happens. Maintenance happens in the background and is entirely non-blocking.

mholt:
@nodesocket I just pushed some changes which make Caddy synchronize its renewals using the file system, so multiple instances can run at the same time and safely coordinate renewals. You won't have to stagger the execution of Caddy processes anymore, and this should scale well to thousands of instances. Would you please test this and report back?

nodesocket:
@mholt thanks for this. Awesome. Can you give a brief overview of how it works? I assume still need to use a shared file system like NFS (Amazon Elastic Filesystem) correct? This is not doing any rsync magic is it?

mholt:
@nodesocket It should "just work" - yes, it'll use the file system. No rsync; the .caddy folder (well, $CADDYPATH) must be locally mounted. This design may not be 100% perfect -- I'm guessing more like 99.7% or thereabouts, assuming there are no implementation bugs (ha!) -- but it doesn't need to be perfect, since rate limits have a little bit of tolerance.

By the way, if your company is relying on Caddy, I highly suggest ordering [extended support](https://caddyserver.com/products/support). If you encounter any bugs, it'll ensure a quicker resolution for your business. :+1:

nodesocket:
@mholt can you provide instructions on how to build this? I assume I need to compile myself?

mholt:
@nodesocket 

> can you provide instructions on how to build this? I assume I need to compile myself?

Yes :) Just like you've been building it up to now, except on this branch instead of master. (If you haven't been building from source, then you should get a commercial license if you haven't already.)

The build instructions are [in the README](https://github.com/mholt/caddy#build) -- just switch to this branch.

dschwar:
Did some testing between this branch and master, with the following caddyfile

```
    subdomain.host.com:443 {
      proxy / web:8000 {
        transparent
        header_upstream X-Request-Start "t={when_unix}000"
      }
      tls /star-subdomain-host-cert/tls.crt /star-subdomain-host-cert/tls.key
    }
    *.subdomain.host.com:443 {
      proxy / web:8000 {
        transparent
        header_upstream X-Request-Start "t={when_unix}000"
      }
      tls /star-subdomain-host-cert/tls.crt /star-subdomain-host-cert/tls.key
    }
    :443 {
      proxy / web:8000 {
        transparent
        header_upstream X-Request-Start "t={when_unix}000"
      }
      tls {
        max_certs 5000
        storage mysql
      }
    }
    :80 {
      proxy / web:8000 {
        transparent
        header_upstream X-Request-Start "t={when_unix}000"
      }
      tls {
        max_certs 5000
        storage mysql
      }
    }
```

I find that when I visit `subdomain.host.com` I get cert a let encrypt certificate. On master this serves the certificate that I've provided.

mholt:
@dschwar What's the `storage` thing you have going on? That seems custom. Please try again without modifications so I can work on reproducing the issue.

dschwar:
After removing the mysql plugin, I tested master vs your branch. On master it loads the provided cert of `subdomain.host.com`, on this branch. it looks like it's attempting to get a LE cert, but fails. I probably don't have the filesystem configured properly though.

```
2018/02/14 18:52:18 [INFO] Obtaining new certificate for subdomain.host.com
2018/02/14 18:52:19 [INFO] acme: Registering account for email@host.com
2018/02/14 18:52:20 http: TLS handshake error from 10.111.0.2:15721: open /root/.caddy/acme/acme-v01.api.letsencrypt.org/sites/subdomain.host.com/subdomain.host.com.crt.lock: no such file or directory
```

The weird thing is, if I browse a site  like `secondary.subdomain.host.com`, then subsequent requests to `subdomain.host.com` will be served using the provided cert.

mholt:
Thanks for giving it another go! We're getting closer.

First thing -- _and you should do this immediately_ -- use the Let's Encrypt [staging endpoint](https://letsencrypt.org/docs/staging-environment/) for testing! Set it [with the `-ca` flag](https://caddyserver.com/docs/cli#ca). Or you may hit rate limits.

Are you running caddy as the same user as who owns that directory, etc? Also, is this subdomain new (i.e. does the containing folder listed in the error message already exist on disk)?

dschwar:
>Are you running caddy as the same user as who owns that directory, etc? 

Yup

```
~/.caddy/acme/acme-v01.api.letsencrypt.org # ls -la
total 12
drwx------    3 root     root          4096 Feb 14 18:52 .
drwx------    3 root     root          4096 Feb 14 18:52 ..
drwx------    3 root     root          4096 Feb 14 18:52 users
~/.caddy/acme/acme-v01.api.letsencrypt.org # whoami
root
~/.caddy/acme/acme-v01.api.letsencrypt.org # ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 /usr/bin/caddy --conf /caddy/Caddyfile --log stdout --email email@host.com --host subdomain.host.com
```


>Also, is this subdomain new (i.e. does the containing folder listed in the error message already exist on disk)?

The containing folder doesn't exist (see output pasted above). Should caddy be trying to provision a LE cert for `subdomain.host.com` since that cert is part of the `tls /star-subdomain-host-cert/tls.crt /star-subdomain-host-cert/tls.key`?

mholt:
Now we're getting somewhere. So there's 2 things going on: opening a lock file where the parent dir doesn't exist, and the subdomain thing. I'm looking into it...

mholt:
@dschwar PS. If you want to help dive into the code and help debug it (since I have to work to reproduce it) -- specifically help with the problem of serving the wrong certificate -- I would really appreciate it. And it would help us get it fixed sooner. You'd want to look in `caddytls/handshake.go` -- probably putting some prints in some of the getConfig or getCertificate functions should help!

mholt:
@dschwar I just pushed a change which should fix the error about opening the file. But I could use your help to pinpoint the other problem with the wrong certificate, if you have a chance today!

dschwar:
Yeah, I've got time. Let me put in some prints and get something more helpful

dschwar:
Not sure if this will be very helpful

```
2018/02/14 21:54:26 [INFO] Successfully loaded TLS assets from /star-subdomain.hostcom-cert/tls.crt and /star-subdomain.hostcom-cert/tls.key
2018/02/14 21:54:26 [INFO] Successfully loaded TLS assets from /star-subdomain.hostcom-cert/tls.crt and /star-subdomain.hostcom-cert/tls.key
Activating privacy features... done.
2018/02/14 21:54:26 [WARNING] TLS disabled for http://subdomain.host.com
https://subdomain.host.com
2018/02/14 21:54:26 https://subdomain.host.com
https://*.subdomain.host.com
2018/02/14 21:54:26 https://*.subdomain.host.com
https://subdomain.host.com
2018/02/14 21:54:26 https://subdomain.host.com
http://subdomain.host.com
2018/02/14 21:54:26 http://subdomain.host.com
2018/02/14 21:54:46 [INFO] GetConfigForClient
(*tls.ClientHelloInfo)({
 CipherSuites: ([]uint16) (len=14 cap=14) {
  <max depth reached>
 },
 ServerName: (string) (len=30) "subdomain.host.com",
 SupportedCurves: ([]tls.CurveID) (len=4 cap=4) {
  <max depth reached>
 },
 SupportedPoints: ([]uint8) (len=1 cap=1) {
  <max depth reached>
 },
 SignatureSchemes: ([]tls.SignatureScheme) (len=9 cap=9) {
  <max depth reached>
 },
 SupportedProtos: ([]string) (len=2 cap=2) {
  <max depth reached>
 },
 SupportedVersions: ([]uint16) (len=4 cap=4) {
  <max depth reached>
 },
 Conn: (*httpserver.clientHelloConn)({
  <max depth reached>
 })
})

2018/02/14 21:54:46 [INFO] getConfig
(string) (len=30) "subdomain.host.com"
(caddytls.configGroup) (len=2) {
 (string) (len=30) "subdomain.host.com": (*caddytls.Config)({
  <max depth reached>
 }),
 (string) (len=32) "*.subdomain.host.com": (*caddytls.Config)({
  <max depth reached>
 })
}
2018/02/14 21:54:46 [INFO] getCertDuringHandshake
(*caddytls.Config)({
 Hostname: (string) (len=30) "subdomain.host.com",
 Enabled: (bool) true,
 ProtocolMinVersion: (uint16) 770,
 ProtocolMaxVersion: (uint16) 771,
 Ciphers: ([]uint16) (len=13 cap=16) {
  <max depth reached>
 },
 PreferServerCipherSuites: (bool) true,
 CurvePreferences: ([]tls.CurveID) (len=2 cap=2) {
  <max depth reached>
 },
 ClientAuth: (tls.ClientAuthType) 0,
 ClientCerts: ([]string) <nil>,
 Manual: (bool) false,
 Managed: (bool) true,
 OnDemand: (bool) true,
 SelfSigned: (bool) false,
 CAUrl: (string) "",
 ListenHost: (string) "",
 AltHTTPPort: (string) "",
 AltTLSSNIPort: (string) "",
 DNSProvider: (string) "",
 ACMEEmail: (string) "",
 KeyType: (acme.KeyType) "",
 StorageProvider: (string) "",
 OnDemandState: (caddytls.OnDemandState) {
  <max depth reached>
 },
 MustStaple: (bool) false,
 ALPN: ([]string) (len=2 cap=2) {
  <max depth reached>
 },
 Certificates: (map[string]string) {
  <max depth reached>
 },
 certCache: (*caddytls.certificateCache)({
  <max depth reached>
 }),
 tlsConfig: (*tls.Config)({
  <max depth reached>
 })
})
2018/02/14 21:54:46 [INFO] Obtaining new certificate for subdomain.host.com
2018/02/14 21:54:47 [INFO] acme: Registering account for email@host.com
2018/02/14 21:54:47 [INFO][subdomain.host.com] acme: Obtaining bundled SAN certificate
2018/02/14 21:54:47 [INFO][subdomain.host.com] AuthURL: https://acme-staging.api.letsencrypt.org/a....
2018/02/14 21:54:47 [INFO][subdomain.host.com] acme: Could not find solver for: dns-01
2018/02/14 21:54:47 [INFO][subdomain.host.com] acme: Trying to solve HTTP-01
2018/02/14 21:54:48 [INFO][subdomain.host.com] Served key authentication
2018/02/14 21:54:48 [INFO][subdomain.host.com] Served key authentication
2018/02/14 21:54:48 [INFO][subdomain.host.com] Served key authentication
2018/02/14 21:54:48 [INFO][subdomain.host.com] Served key authentication
2018/02/14 21:54:49 [INFO][subdomain.host.com] The server validated our request
2018/02/14 21:54:49 [INFO][subdomain.host.com] acme: Validations succeeded; requesting certificates
2018/02/14 21:54:50 [INFO] acme: Requesting issuer cert from https://acme-staging.api.letsencrypt.org/acme/issuer-cert
2018/02/14 21:54:50 [INFO][subdomain.host.com] Server responded with a certificate.
2018/02/14 21:54:50 [INFO][subdomain.host.com] Certificate written to disk: /root/.caddy/acme/acme-staging.api.letsencrypt.org/sites/subdomain.host.com/subdomain.host.com.crt
2018/02/14 21:54:50 [INFO] getCertDuringHandshake
(*caddytls.Config)({
 Hostname: (string) (len=30) "subdomain.host.com",
 Enabled: (bool) true,
 ProtocolMinVersion: (uint16) 770,
 ProtocolMaxVersion: (uint16) 771,
 Ciphers: ([]uint16) (len=13 cap=16) {
  <max depth reached>
 },
 PreferServerCipherSuites: (bool) true,
 CurvePreferences: ([]tls.CurveID) (len=2 cap=2) {
  <max depth reached>
 },
 ClientAuth: (tls.ClientAuthType) 0,
 ClientCerts: ([]string) <nil>,
 Manual: (bool) false,
 Managed: (bool) true,
 OnDemand: (bool) true,
 SelfSigned: (bool) false,
 CAUrl: (string) "",
 ListenHost: (string) "",
 AltHTTPPort: (string) "",
 AltTLSSNIPort: (string) "",
 DNSProvider: (string) "",
 ACMEEmail: (string) (len=26) "email@host.com",
 KeyType: (acme.KeyType) "",
 StorageProvider: (string) (len=4) "file",
 OnDemandState: (caddytls.OnDemandState) {
  <max depth reached>
 },
 MustStaple: (bool) false,
 ALPN: ([]string) (len=2 cap=2) {
  <max depth reached>
 },
 Certificates: (map[string]string) {
  <max depth reached>
 },
 certCache: (*caddytls.certificateCache)({
  <max depth reached>
 }),
 tlsConfig: (*tls.Config)({
  <max depth reached>
 })
})
```

nodesocket:
@mholt probably unrelated to these changes, but when I set the `ca` to the Let's Encrypt staging environment 

```
tls {
        ca https://acme-staging.api.letsencrypt.org/directory
}
```

I am getting the error:

```
failed to get certificate: Failed to HTTP POST to https://acme-staging.api.letsencrypt.org/acme/new-cert -> Post https://acme-staging.api.letsencrypt.org/acme/new-cert: net/http: timeout awaiting response headers
```

Which is super strange it is timing out. I can curl the endpoint:

```
curl -i https://acme-staging.api.letsencrypt.org/acme/new-cert -X POST
HTTP/1.1 400 Bad Request
Server: nginx
Content-Type: application/problem+json
Content-Length: 96
Replay-Nonce: YkFKsSUx3F8ixYiEHXQHdnrfhpDhbdHURatfiQikdlM
Expires: Thu, 15 Feb 2018 00:16:12 GMT
Cache-Control: max-age=0, no-cache, no-store
Pragma: no-cache
Date: Thu, 15 Feb 2018 00:16:12 GMT
Connection: close

{
  "type": "urn:acme:error:malformed",
  "detail": "Parse error reading JWS",
  "status": 400
}
```

mholt:
@dschwar Thanks; I'm wondering, though, how `[WARNING] TLS disabled for http://subdomain.host.com` made its way into there from your Caddyfile. And also that there is a TLSConfig for `"subdomain.host.com"` at the same time. This isn't a behavior I'm observing...

@nodesocket Yes, that seems unrelated, maybe just a sporadic network failure.

nodesocket:
@mholt it is consistence. Even created a new AWS instance and still getting:

```
Feb 15 01:22:11 ip-172-31-30-2 caddy[4785]: 2018/02/15 01:22:11 [portal.phishprotection.duocircle.info] failed to get certificate: Failed to HTTP POST to https://acme-staging.api.letsencrypt.org/acme/new-cert -> Post https://acme-staging.api.letsencrypt.org/acme/new-cert: net/http: timeout awaiting response headers
```

I am very confused as this has worked in the past.

```
portal.phishprotection.duocircle.info {
    gzip
    tls support@duocircle.com
    errors /var/log/caddy/error.log

    header / Strict-Transport-Security "max-age=15768000;"

    root /var/www/portal

    tls {
        ca https://acme-staging.api.letsencrypt.org/directory
    }

    fastcgi / 127.0.0.1:3000 php {
        env RDS_ENDPOINT {$RDS_ENDPOINT}
        index index.php
    }
}
```

mholt:
@nodesocket I'm seeing that too now. I haven't made any changes since now and when it was working. Switching back to master branch and I'm seeing the same thing. I think staging is having issues; we should both [report it to Let's Encrypt](https://community.letsencrypt.org).

@dschwar I'm having a hard time reproducing the behavior you're describing. Can you please post the full, unmodified, un-redacted Caddyfile and log outputs here?

nodesocket:
@mholt Let's Encrypt staging is still timing-out for me. I can't believe others are not encountering this after 4+ hours. Just curious, are you testing from AWS EC2 instances? I am.

dschwar:
> @dschwar Thanks; I'm wondering, though, how [WARNING] TLS disabled for http://subdomain.host.com made its way into there from your Caddyfile. And also that there is a TLSConfig for "subdomain.host.com" at the same time. This isn't a behavior I'm observing...

This seems to happen because I have the command line `-host` arg set to `subdomain.host.com`.

I tried removing the `-host` command line arg, which gives me the log
`[WARNING] TLS disabled for http://`
and then `subdomain.host.com` is served using the appropriate provided certificate, but Let's Encrypt no longer provisions.

So I tried changing the `-host` arg to `https://subdomain.host.com` and now I get
`WARNING] TLS disabled for http://https://subdomin.host.com`

mholt:
@nodesocket 

>  Let's Encrypt staging is still timing-out for me. I can't believe others are not encountering this after 4+ hours. Just curious, are you testing from AWS EC2 instances? I am.

Nope, not using EC2. Thanks for commenting on their forums -- hopefully they'll find it soon. Not many people use staging except client developers I think.

@dschwar  -- okay, I think what you'll need to do is simplify your config down to bare-bones. Run caddy without any extra command line args (except to set the staging endpoint!) and then run Caddy again with a simple Caddyfile. I can help debug it but only if I have all the information: how Caddy was run, what env variables are set, the Caddyfile, log output, requests that are made, etc -- all unredacted and unmodified. Otherwise, I can only give you tips to help you debug it on your end.

(The `-host` flag, [per the docs](https://caddyserver.com/docs/cli#host) specifies a default host to listen on; so if the hostname defined in the Caddyfile is empty, it will be filled in with `-host`. And `https://...` is not a host, of course. Remember, Caddy is very easy to use, but has enough power that you can easily shoot yourself in the foot. Unlike other web servers, with Caddy, you'll want to trim back the config when things start getting complicated.)

dschwar:
@mholt I can test this tomorrow but my current hypothesis is that if the ‘-host’ flag is supplied and there is a Caddyfile with a supplied tls cert for that host and a wildcard on-demand config, then when you visit the host, it will choose to generate an on demand cert.

mholt:
@dschwar Ahhhh, wait, it's only generating the certificate on-demand (i.e. not at startup)? Is `subdomain.host.com` that you visit in your browser the exact same one as the first site in your Caddyfile, `subdomain.host.com:443`? If so, then what you found may not be a bug with the TLS changes. If you set `-host subdomain.host.com` and then define a site in the Caddyfile with an empty hostname (`:443` for example) then that site is assigned a default hostname of `subdomain.host.com` according to your flag. Thus there is a duplicate site name and that should be caught at startup. Let me know what you think.

dschwar:
> @dschwar Ahhhh, wait, it's only generating the certificate on-demand (i.e. not at startup)?

Yup, only on demand. Should have mentioned that earlier, sorry.

> Is subdomain.host.com that you visit in your browser the exact same one as the first site in your Caddyfile, subdomain.host.com:443?

Yup. When redacting, I made sure everything matches up and is in the same form.   

> hostname (:443 for example) 

I thought a :443 would give me a wildcard host? So that I can generate certs for any domain that points at my server (hence why I redact, don’t want spam)

mholt:
@dschwar Thanks for clarifying.

> I thought a :443 would give me a wildcard host? So that I can generate certs for any domain that points at my server

Well, the `-host` flag sets the host when it is missing for site configurations. Generally, `-host` is used without a Caddyfile to even more quickly get a simple site up and running. The `-host` flag _can_ be used in conjunction with one in some cases for convenience, but your case is not one of them.

> (hence why I redact, don’t want spam)

Won't matter -- all hostnames with Let's Encrypt certificates go to public CT logs anyway. :)

Anyway, I'm thinking of making your situation a "collision" at startup, because otherwise it's not obvious. In other words, using `-host subdomain.host.com` and with sites `subdomain.host.com` and `:443` in your Caddyfile should be an error because the second site is indistinguishable from the first. Sound good?

mholt:
@dschwar I've pushed that fix to this branch; feel free to try it out and let me know. You should now get an error when you try to start Caddy with the configuration you had. The reason you were getting the wrong certificate is because Caddy couldn't disambiguate between the two site configurations because of your `-host` flag: the sites weren't distinct, even though they had different configurations. So now we prevent this case, by raising an error at startup.

dschwar:
> @dschwar I've pushed that fix to this branch; feel free to try it out and let me know. You should now get an error when you try to start Caddy with the configuration you had. The reason you were getting the wrong certificate is because Caddy couldn't disambiguate between the two site configurations because of your -host flag: the sites weren't distinct, even though they had different configurations. So now we prevent this case, by raising an error at startup.

I think I'm missing something now. I've removed the `-host subdomain.host.com` arg and now when I visit `www.davesdomain.com` which CNAME's to `subdomain.host.com`, Chrome gives me a `This site can’t provide a secure connection`

and the logs show

```
2018/02/15 14:49:29 [INFO] Successfully loaded TLS assets from /star-subdomain-hostcom-cert/tls.crt and /star-subdomain-hostcom-cert/tls.key
2018/02/15 14:49:29 [INFO] Successfully loaded TLS assets from /star-subdomain-hostcom-cert/tls.crt and /star-subdomain-hostcom-cert/tls.key
Activating privacy features... done.
2018/02/15 14:49:29 [WARNING] TLS disabled for http://
https://subdomain.host.com
2018/02/15 14:49:29 https://subdomain.host.com
https://*.subdomain.host.com
2018/02/15 14:49:29 https://*.subdomain.host.com
https://
2018/02/15 14:49:29 https://
http://
2018/02/15 14:49:29 http://
2018/02/15 14:52:30 [INFO] getCertificate for name: www.davesdomain.com
(map[string]string) (len=2) {
 (string) (len=32) "*.subdomain.host.com": (string) (len=64) "41z9d420f85af211afdsfsdgdsfhfghfgdhjghjghfkhjkjhkjhgkjhgkjhgc122b7",
 (string) (len=30) "subdomain.host.com": (string) (len=64) "41z9d420f85af211afdsfsdgdsfhfghfgdhjghjghfkhjkjhkjhgkjhgkjhgc122b7"
}
2018/02/15 14:52:30 [INFO] onDemand: false, loadIfNecessary: true, obtainIfNecessary: true
2018/02/15 14:52:30 http: TLS handshake error from 10.234.0.25:23064: no certificate available for www.davesdomain.com
```

mholt:
@dschwar Thanks, I think I'm finally onto a fix. Just a hunch, can you verify that replacing the site named `:443` in your Caddyfile with `*:443` acts different?

dschwar:
> Just a hunch, can you verify that replacing the site named :443 in your Caddyfile with *:443 acts different?

Seems to do the same thing

mholt:
@dschwar Hmm well that's discouraging. Are you sure? If you could pull the latest and try again -- with and without the `*`, that would be good. I was able to reproduce what I think is the issue, in that it was getting the wrong config for the name. If you could try it after pulling my latest commit, then let me know, that would be great!

dschwar:
It works!

without the `*`

mholt:
@dschwar Great!! It should work with the `*` as well.

I'll merge this PR as Go 1.10 is released, unless someone speaks up with another bug report before then.

dschwar:
> It should work with the * as well.

Testing that (for fun), when you say it should with `*`, do you mean 
```
    *:443 {
      proxy / web:8000 {
        transparent
        header_upstream X-Request-Start-Edge {>X-Request-Start}
        header_upstream X-Request-Start "t={when_unix}000"
      }
      tls {
        max_certs 5000
      }
    }
```

or 
```
    * {
      proxy / web:8000 {
        transparent
        header_upstream X-Request-Start-Edge {>X-Request-Start}
        header_upstream X-Request-Start "t={when_unix}000"
      }
      tls {
        max_certs 5000
      }
    }
```


I just attempted the first one and it doesn't work for me.

mholt:
The first one. That should work, (it works for me), so if you want to investigate feel free!

dschwar:
>The first one. That should work, (it works for me), so if you want to investigate feel free!

It seems to work if I remove the `subdomain.host.com:443` and `*.subdomain.host.com:443` from my Caddyfile, the resulting Caddyfile becomes
```
*:443 {
  proxy / web:8000 {
    transparent
    header_upstream X-Request-Start-Edge {>X-Request-Start}
    header_upstream X-Request-Start "t={when_unix}000"
  }
  tls {
    max_certs 5000
  }
}
*:80 {
  proxy / web:8000 {
    transparent
    header_upstream X-Request-Start-Edge {>X-Request-Start}
    header_upstream X-Request-Start "t={when_unix}000"
  }
  tls {
    max_certs 5000
  }
}
```

with the command line args
`["--conf", "/caddy/Caddyfile", "--log", "stdout", "--email", "myemail@host.com", "--ca", "https://acme-staging.api.letsencrypt.org/directory"]`

mholt:
@dschwar Ah, you know what, I'm sorry -- that is correct and expected behavior; I was wrong before. Wildcard characters (`*`) only match a single domain label (e.g. `sub.domain.com` cannot be matched by `*` but or even `*.*` but can be matched by `*.*.*`), as per the RFC related to wildcard SSL certificates.

So actually I think that remaining "bug" is a non-issue. However, the investigation led me to fix what I think was some logical inconsistencies in the code, so it was all good. I'm gonna merge this real soon. Probably today.

Issue #1926, opened 2017-10-18
Server header is included in HTTPS redirect response even if disabled in config


cskr:
### 1. What version of Caddy are you using (`caddy -version`)?

Caddy 0.10.10 (unofficial)

### 2. What are you trying to do?

Disable the `Server` header in all responses of a site.

### 3. What is your entire Caddyfile?
```text
blog.thecskr.in {
        root /var/www/blog

        header / {
                Strict-Transport-Security "max-age=31536000; includeSubDomains"
                -Server
        }
        header /images {
                Cache-Control "public, max-age=604800"
        }
        gzip

        log /post /var/log/caddy/blog-access.log {combined}
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?

```
/usr/local/bin/caddy -log /var/log/caddy/caddy.log -agree=true -email=my@email -conf=/etc/caddy/Caddyfile -root=/var/tmp
```

### 5. Please paste any relevant HTTP request(s) here.

```
$ curl -i http://blog.thecskr.in/
HTTP/1.1 301 Moved Permanently
Location: https://blog.thecskr.in/
Server: Caddy
Date: Wed, 18 Oct 2017 04:23:21 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 59
Accept-Ranges: none
Connection: close

<a href="https://blog.thecskr.in/">Moved Permanently</a>.
```

### 6. What did you expect to see?

```
$ curl -i http://blog.thecskr.in/
HTTP/1.1 301 Moved Permanently
Location: https://blog.thecskr.in/
Date: Wed, 18 Oct 2017 04:23:21 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 59
Accept-Ranges: none
Connection: close

<a href="https://blog.thecskr.in/">Moved Permanently</a>.
```

### 7. What did you see instead (give full error messages and/or log)?

The `Server` header was present.

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

Try creating a site with `-Server` specified in Caddyfile and access site with `http://` instead of `https://`.

tobya:
Thanks for the issue.

mholt:
This is probably because the automatic HTTP->HTTPS redirects don't use the middleware chain defined by the user. Perhaps we should see if there is a way to get the `header` middleware copied over to the HTTP one when we set up the custom redirect. (It's a special case.)

theodesp:
[...]

PR #2016, opened 2018-02-05
Change "Server" header set location so that it is set only possibly once

magikstm:
Change "Server" header set location so that it is set only possibly once.

I investigated #1926 a bit more and made additional tests locally with this caddyfile:

```
http://localhost:8084/
{
	root C:\webroot
	proxy / localhost:8085
}

http://localhost:8085/
{
	root C:\webroot
}
```

I noted that if I visit http://localhost:8084/2.jpg "Server" header is set twice.

This commit should correct the issue mentioned above as well as #1926 which is related.

Please review this PR to confirm if it may be a good solution for this issue.

#1926 

None.

### 4. Checklist


mholt:
The Server header having two values (or being set twice -- equivalent in HTTP) is a feature, I think, not a bug. It's useful to see which servers are handling the request. I rely on this when serving the Caddy forum, for example.

A solution for the linked issue would be to bring in the header middleware into the middleware chain that simply redirects HTTP to HTTPS. Or something like that.

Thanks for the PR! But I don't think it's something we're looking for.

magikstm:
Np. I didn't see that use-case before submitting my PR.

Issue #2017, opened 2018-02-06
Allow to skip logging of certain requests


ghostsquad:
## General Description

I don't want to whitelist every endpoint, I just want to ignore logging or change logging for specific endpoints.

This might be easily done with something like (edited 2018-02-14):
```diff
log access.log {
    not /health
}
```

### 1. What version of Caddy are you using (`caddy -version`)?
v0.10.10

### 2. What are you trying to do?
Skip logging for `/health` endpoints

### 3. What is your entire Caddyfile?
```text
http://*/ {
  log stdout
  errors stderr

  bind 0.0.0.0
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
n/a

### 5. Please paste any relevant HTTP request(s) here.
`/health`

### 6. What did you expect to see?
`/health` log requests cannot be ignored. Log directive only supports base path matching.

### 7. What did you see instead (give full error messages and/or log)?
Everything is logged, but I have an `/health` endpoint that I'd like to ignore

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
Run above caddyfile, and hit `/health`, try to modify `Caddyfile` to not log requests to `/health`

mholt:
For the record, a more Caddy-omatic way to do this would be:

```
log access.log {
    not /health
}
```

ghostsquad:
@mholt I figured out another way to accomplish this, but it's still not great:

```
http://*/health {
  bind 0.0.0.0
}

http://*/ {
  log stdout
  errors stderr

  bind 0.0.0.0
}
```

But this is pretty heavy for anything but this simple example.

tobya:
I'll take a look at implementing this.

tobya:
@mholt this seems an easy enough feature to add.

However I’m wondering does it warrant a little more discussion?

Is not writing entries in the log desirable? I may seems so in the short term but cause numerous issues in the long term. Perhaps a heartbeat url is not logged and the stops working? There is no indication in the logs anything has changed.  If a ddos or other attack is attempted on a url that is not logged will this hamper investigation?

I know all of this is individual choice but I thought it was worth a little more discussion

tobya:
I have an initial implementation it will have the following rules

      log / log.txt {common} {
         except /test
       }

Will not log 
  - example.com/test 
  - example.com/testing 
  - example.com/test/another/directory.pdf

Will log
  - example.com/atest

Example2 

      log / log.txt {common} {
         except /test/
       }

[...]

PR #2028, opened 2018-02-18
Allow certain paths to be skipped in log

tobya:

The change adds an `except` sub keyword to the log directive.  If the path being logged matches the except path the request will not be written to the logs

#2017 

The log directive can now have an `except` sub directive

    log / log.txt {common} {
         except /debug /testing/ 
         except /pages/heartbeat.php
    }

### 4. Checklist
I will add some tests


caddyhttp/httpserver/logger.go
```diff
@@ -84,6 +85,17 @@ func (l Logger) MaskIP(ip string) string {
 
 }
 
+func (l Logger) ShouldLog(path string) bool {
+	shouldLog := true
+	for ex := 0; ex < len(l.Exceptions); ex++ {
+		if Path(path).Matches(l.Exceptions[ex]) {
+			shouldLog = false
+			break
```
mholt: Replace with `return false` ;) - then you can get rid of `shouldLog`

caddyhttp/log/log.go
```diff
@@ -67,17 +67,19 @@ func (l Logger) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
...
 			for _, e := range rule.Entries {
-
-				// Mask IP Address
-				if e.Log.IPMaskExists {
-					hostip, _, err := net.SplitHostPort(r.RemoteAddr)
-					if err == nil {
-						maskedIP := e.Log.MaskIP(hostip)
-						// Overwrite log value with Masked version
-						rep.Set("remote", maskedIP)
+				if len(e.Log.Exceptions) == 0 || e.Log.ShouldLog(r.URL.Path) {
```
mholt: This is just a style preference, but instead of adding another level of indent, how about if we invert this statement to `len(e.Log.Exceptions) > 0 && !e.Log.ShouldLog(r.URL.Path) {` and then you can just `continue` instead of nesting all the logic.

tobya: I think this is less clear, but I've made the change :)

caddyhttp/httpserver/logger.go
tobya: Done.

tobya:
@mholt I've made those changes now.  Tests passed locally.

Issue #1837, opened 2017-08-24
Support more response placeholders


vicanso:
I want to log response header in log directive, is it possible to add a placeholder to support it? 

```
{>Header}  --> request header
{<Header}  --> response header
```

tobya:
hi viscanso, thank you for the feature request.  Do you have a specific use case in mind for this?

vicanso:
No specific use case. I want this feature to be similar as the feature of Nginx

mholt:
Wish granted. Thanks @tobya!

PR #2029, opened 2018-02-18
Allow Response Headers as a placeholder

tobya:
Allows Response header values to be used by the replacer and log.

It seemed best to use the less than symbol `<` so a response header would be `{<Server}`

Only some response headers seem to be available to the replacer when called .

eg - the following are shown in chrome

Content-Length:16
Content-Type:text/plain; charset=utf-8
Date:Sun, 18 Feb 2018 08:07:13 GMT
Server:Caddy
X-Content-Type-Options:nosniff

only 

    {<Content-Type}
    {<Server}
    {<X-Content-Type-Options}

will return a value.

#1837 

We simply need to add to the placeholder page that {<Header} is allowed as a response placeholder.

### 4. Checklist
I will write tests

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -225,7 +225,6 @@ func (r *replacer) getSubstitution(key string) string {
 			}
 		}
 	}
-	log.Printf("ResponseHEaders: %+v, %+V", r.responseRecorder.Header(), r.responseRecorder.Header())
 	// search response headers then
 	if key[1] == '<' {
 		want := key[2 : len(key)-1]
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"io"
 	"io/ioutil"
+	"log"
 	"net"
 	"net/http"
 	"net/http/httputil"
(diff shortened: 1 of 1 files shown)


caddyhttp/httpserver/replacer.go
```diff
@@ -224,6 +224,16 @@ func (r *replacer) getSubstitution(key string) string {
 			}
 		}
 	}
+	// search response headers then
+	if key[1] == '<' {
+		want := key[2 : len(key)-1]
+		for key, values := range r.responseRecorder.Header() {
```
mholt: Note from comment way above in the constructor, NewReplacer: `rr may be nil if it is not available.` -- so do a nil check before line 228 to make sure `r.responseRecorder` isn't nil.

tobya: @mholt I have added this check and tests now.

caddyhttp/httpserver/replacer.go
```diff
@@ -225,12 +225,14 @@ func (r *replacer) getSubstitution(key string) string {
...
 	}
 	// search response headers then
-	if key[1] == '<' {
-		want := key[2 : len(key)-1]
-		for key, values := range r.responseRecorder.Header() {
-			// Header placeholders (case-insensitive)
-			if strings.EqualFold(key, want) {
-				return strings.Join(values, ",")
+	if r.responseRecorder != nil {
+		if key[1] == '<' {
```
mholt: These two if statements can be combined with && so we move out one level of nesting. Sorry I didn't clearly suggest that before. :)

tobya:
@mholt I have made that change to the if statements.  This seems to work well now.  At the moment {<Content-Length} has no value.  Perhaps that could be investigated seperately.

mholt:
@tobya Nice! Thanks for checking. Yes, that Content-Length issue can be looked at separately.

What's with the `caddyhttp/browse/tempTemplate247651759` file being included with this change? 🤔

tobya:
Sorry I'm not sure what happened, removed now.

PR #2032, opened 2018-02-19
dep

tgulacsi:
Use the new dep package management tool to manage dependencies and the vendor/ directory.


tgulacsi:
Sorry, plugins haven't been intended to be pushed here. Just the dep'd dependencies.

tobya:
Thanks for the PR. We would need some more justification to take a PR this large. Perhaps you could provide a bit more explanation

mholt:
It seems like Go is not necessarily settling on dep either. https://research.swtch.com/vgo-intro

Right now we don't have a compelling reason to switch from using the vendor folder with gvt. Thanks for your PR though! Feel free to continue discussion, but I'm going to close this.

tgulacsi:
Understood.

Matt Holt <notifications@github.com> ezt írta (időpont: 2018. febr. 21.,
Sze, 20:40):

Issue #2038, opened 2018-02-20
httpserver: Trimmed URI doesn't reconstruct query string


sirkon:
### 1. What version of Caddy are you using (`caddy -version`)?
[0.10.11](https://github.com/mholt/caddy/commit/d3f338ddab9ee24b376b5c9c51e202581e2d43ba)

### 2. What are you trying to do?

Custom simplistic plugin, just to show a error

```go
package main

import (
	"io/ioutil"
	"log"
	"os"

	"net/http"

	"io"

	_ "github.com/mholt/caddy/caddyhttp"

	"github.com/mholt/caddy"
	"github.com/mholt/caddy/caddyhttp/httpserver"
	"github.com/sanity-io/litter"
)

func setup(c *caddy.Controller) error {
	for c.Next() {
		// get configuration
	}
	// now what?
	cfg := httpserver.GetConfig(c)
	cfg.AddMiddleware(func(next httpserver.Handler) httpserver.Handler {
		return handler{}
	})
	return nil
}

// provide loader function
func defaultLoader(serverType string) (caddy.Input, error) {
	contents, err := ioutil.ReadFile("Caddyfile")
	if err != nil {
		if os.IsNotExist(err) {
			return nil, nil
		}
		return nil, err
	}
	return caddy.CaddyfileInput{
		Contents:       contents,
		Filepath:       "Caddyfile",
		ServerTypeName: serverType,
	}, nil
}

func init() {
	caddy.RegisterPlugin(
		"custom",
		caddy.Plugin{
			ServerType: "http",
			Action:     setup,
		},
	)
	httpserver.RegisterDevDirective("custom", "")
	caddy.SetDefaultCaddyfileLoader("default", caddy.LoaderFunc(defaultLoader))
}

type handler struct{}

func (handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
	_, err := io.WriteString(w, litter.Sdump(r.URL))
	return http.StatusOK, err
}

func main() {
	caddyfile, err := caddy.LoadCaddyfile("http")
	if err != nil {
		log.Fatal(err)
	}

	instance, err := caddy.Start(caddyfile)
	if err != nil {
		panic(err)
	}

	instance.Wait()
}

```

### 3. What is your entire Caddyfile?
```text
0.0.0.0:2015/s {
    custom
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
```
```

### 5. Please paste any relevant HTTP request(s) here.

[...]

PR #2039, opened 2018-02-21
 Fix Issue #2038 - Querystring being lost from URI

tobya:

Fixes a bug introduced by #2014

I did not origionally test with query string.  I have added tests and fixed code.  The url passed in to trimPathPrefix will never have scheme / host / port 

I may never have a fragement either but I have added test to ensure if it is it willwork.

Please review code and test if possible. 
 
 #2014 
#2038 

NA

### 4. Checklist


sirkon:
This PR fixes a bug when no query string was passing to further middlewares. Yet it is not really a fix for #2038, scheme and host are lost somewhere:

```go
&url.URL{
  Scheme: "",
  Opaque: "",
  User: nil,
  Host: "",
  Path: "/123213",
  RawPath: "",
  ForceQuery: false,
  RawQuery: "r87ewfhdghfdfkjghdfkghdfklvckjhvdfs",
  Fragment: "",
}
```

tobya:
Thanks for checking this out.  
I'm not sure that scheme and host were passed in 0.10.10 .

mholt:
Thanks Toby! Really appreciate your contributions. :)

Issue #2044, opened 2018-02-22
onevent bug: need to deregister previous hooks after graceful restart


UladzimirTrehubenka:
Observed with CoreDNS that uses latest https://github.com/mholt/caddy/tree/master/onevent

On start with Corefile:
```
.:53 {
	forward . 8.8.8.8:53
	errors
	on startup touch /tmp/ready
}
```
In logs:
```
[INFO] Blocking Command "touch /tmp/ready" with ID 0b73050f-c09b-4188-9df3-422fdec0bf5b
```
Change Corefile to:
```
.:53 {
	forward . 8.8.8.8:53
	errors
	on startup touch /tmp/ready2
}
```
and make graceful restart: killall -USR1 coredns
In logs:
```
[INFO] Blocking Command "touch /tmp/ready" with ID 0b73050f-c09b-4188-9df3-422fdec0bf5b
[INFO] Blocking Command "touch /tmp/ready2" with ID de1d2119-8822-41a5-92eb-f9d0b10d558b
```

mholt:
Pinging @elcore. Something that will help with this is a change in #2015, which introduces instance-scoped storage. You can "store" the event callbacks in there, which will be disposed of when the instance is disposed of (which happens during a reload: a whole new instance is made in its place).

This is what I am referring to: https://github.com/mholt/caddy/blob/a03eba6fbc2770ef6bbe7defa184ec4a6817f0c2/caddy.go#L116-L122

And this is how to use it in the setup function: https://github.com/mholt/caddy/blob/a03eba6fbc2770ef6bbe7defa184ec4a6817f0c2/caddytls/setup.go#L52

That should be of some help!

elcore:
Hi @mholt,

Unfortunately, I do not have enough time to work on this :(

I kindly ask you or somebody else to work on this.

P.S.: I am not sure if a instance-scoped storage should be implemented in onevent, as onevent registers an event hook, which is stored at https://github.com/mholt/caddy/blob/master/plugins.go#L42

elcore:
Hello @UladzimirTrehubenka,

according to my test, my PR should fix your problem, it would be nice if you would try it yourself :)

UladzimirTrehubenka:
I have tested - PR fix the issue.

PR #2047, opened 2018-02-24
caddy: Purge event hooks after USR1 reload

elcore:

- Purge event hooks after USR1 reload
- ~Create a new func to delete event hooks~

#2044

/

### 4. Checklist

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -271,6 +271,19 @@ func EmitEvent(event EventName, info interface{}) {
 	})
 }
 
+// purgeEventHooks purges all event hooks from the map
+func purgeEventHooks() {
+	eventHooks.Range(func(k, _ interface{}) bool {
+		eventHooks.Delete(k)
+		return true
+	})
+}
+
+// deleteEventHook deletes one event hook from the map
+func deleteEventHook(name string) {
+	eventHooks.Delete(name)
+}
+
 // ParsingCallback is a function that is called after
 // a directive's setup functions have been executed
 // for all the server blocks.
--- a/plugins.go
+++ b/plugins.go
@@ -279,11 +279,6 @@ func purgeEventHooks() {
 	})
 }
 
-// deleteEventHook deletes one event hook from the map
-func deleteEventHook(name string) {
-	eventHooks.Delete(name)
-}
-
 // ParsingCallback is a function that is called after
 // a directive's setup functions have been executed
 // for all the server blocks.
--- a/plugins.go
+++ b/plugins.go
@@ -39,7 +39,7 @@ var (
 
 	// eventHooks is a map of hook name to Hook. All hooks plugins
 	// must have a name.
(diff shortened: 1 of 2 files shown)


sigtrap_posix.go
```diff
@@ -76,6 +76,9 @@ func trapSignalsPosix() {
 					caddyfileToUse = newCaddyfile
 				}
 
+				// Purge the old event hooks
+				purgeEventHooks()
```
rdrozhdzh: the potential drawback is that we'll lose valid hooks in case of unsuccessful restart

elcore: @rdrozhdzh I fixed this, thank you!

elcore: @rdrozhdzh I fixed this, thank you!

elcore: @rdrozhdzh I fixed this, thank you!!!

sigtrap_posix.go
```diff
@@ -76,9 +76,17 @@ func trapSignalsPosix() {
...
+				// Backup old event hooks
+				oldEventHooks := cloneEventHooks()
+
+				// Purge the old event hooks
+				purgeEventHooks()
+
 				// Kick off the restart; our work is done
 				_, err = inst.Restart(caddyfileToUse)
 				if err != nil {
+					eventHooks = oldEventHooks
```
tw4452852: Actually there is race condition here, because pointer assignment isn't atomic on some platform.
But this is a rare case we should hit, we can just let it go for now. Maybe this is one of the TODO list.

elcore: @tw4452852 do you have a source for that?

elcore: We could use a mutex to prevent a race

elcore: @tw4452852 I implemented sync.Mutex to protect `eventHooks` itself

elcore: @tw4452852 I implemented `sync.Mutex` to protect `eventHooks` itself

elcore: Okay, I have found a better solution :)

plugins.go
```diff
@@ -271,6 +271,36 @@ func EmitEvent(event EventName, info interface{}) {
...
+}
+
+// restoreEventHooks restores eventHooks with a provided *sync.Map
+func restoreEventHooks(m *sync.Map) {
+	// Purge old event hooks
+	purgeEventHooks()
+
+	// Restore event hooks
+	m.Range(func(k, v interface{}) bool {
+		eventHooks.Store(k, v)
```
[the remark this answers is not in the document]
elcore: @mholt We restore hooks after an unsuccessful reload to maintain a "known" working state.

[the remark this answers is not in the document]
mholt: Oh! Sorry, you're right.

Issue #2055, opened 2018-03-05
Placeholders for TLS protocol version and cipher


omz13:
Would it be possible to add placeholders for, where applicable, the TLS protocol version (e.g. `ssl_protocol`) and TLS CIPHER (e.g. `ssl_cipher`) that a client is using?

This would be very useful as it would allow not only such things to be logged, but also to do redirection when a user is using an older TLS version (e.g. 1.0) to a page that tells them to upgrade their client to something a bit more modern (instead of simply being unable to handshake their connection and just dropping it).

PR #2062, opened 2018-03-11
Implement placeholders for tls_protocol and tls_cipher

omz13:

It implements two new placeholders: ```tls_protocol``` and ```tls_cipher``` which will contain, when using a secure connection, the protocol and cipher respectively.

For fastcgi, it passes these two placeholders in a manner that is compatible with Apache's [mod_ssl](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html), i.e. as environment variables ```SSL_PROTOCOL``` and ```SSL_CIPHER```.

#2055

The documentation for placeholders will need to incorporate details of these two placeholders. Essentially:

```{tls_protocol}``` The protocol used for a secure transport connection (e.g. "tls1.2").

```{tls_cipher}``` The cipher used for a secure transport connection (e.g. "ECDHE-ECDSA-WITH-CHACHA20-POLY1305").

### 4. Checklist

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -34,6 +34,7 @@ import (
 	"time"
 
 	"crypto/tls"
+
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 	"github.com/mholt/caddy/caddytls"
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -33,10 +33,10 @@ import (
 	"sync/atomic"
 	"time"
 
+	"crypto/tls"
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 	"github.com/mholt/caddy/caddytls"
-	"crypto/tls"
 )
 
 // Handler is a middleware type that can handle requests as a FastCGI client.
@@ -327,7 +327,7 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 		env["HTTPS"] = "on"
 		// and pass the protocol details like apache's mod_ssl
 		v, ok := sslProtocolToStringMap[r.TLS.Version]
-		if (ok) {
+		if ok {
 			env["SSL_PROTOCOL"] = v
 		}
 		// and pass the cipher like apache's mod_ssl
@@ -483,7 +483,7 @@ func (l LogError) Error() string {
 // Map of supported protocols to Apache ssl_mod format
 // Note that these are slightly different from SupportedProtocols in caddytls/config.go's
 var sslProtocolToStringMap = map[uint16]string{
-	tls.VersionTLS10 : "TLSv1",
-	tls.VersionTLS11 : "TLSv1.1",
-	tls.VersionTLS12 : "TLSv1.2",
+	tls.VersionTLS10: "TLSv1",
+	tls.VersionTLS11: "TLSv1.1",
+	tls.VersionTLS12: "TLSv1.2",
 }
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -35,6 +35,8 @@ import (
 
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
+	"github.com/mholt/caddy/caddytls"
+	"crypto/tls"
 )
 
 // Handler is a middleware type that can handle requests as a FastCGI client.
@@ -323,6 +325,18 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 	// Some web apps rely on knowing HTTPS or not
 	if r.TLS != nil {
 		env["HTTPS"] = "on"
+		// and pass the protocol details like apache's mod_ssl
+		v, ok := sslProtocolToStringMap[r.TLS.Version]
+		if (ok) {
+			env["SSL_PROTOCOL"] = v
+		}
+		// and pass the cipher like apache's mod_ssl
+		for k, v := range caddytls.SupportedCiphersMap {
+			if v == r.TLS.CipherSuite {
+				env["SSL_CIPHER"] = k
+				break
+			}
+		}
 	}
 
 	// Add env variables from config (with support for placeholders in values)
@@ -464,4 +478,12 @@ type LogError string
 // Error satisfies error interface.
 func (l LogError) Error() string {
 	return string(l)
+}
+
+// Map of supported protocols to Apache ssl_mod format
+// Note that these are slightly different from SupportedProtocols in caddytls/config.go's
+var sslProtocolToStringMap = map[uint16]string{
+	tls.VersionTLS10 : "TLSv1",
+	tls.VersionTLS11 : "TLSv1.1",
+	tls.VersionTLS12 : "TLSv1.2",
 }
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -326,12 +326,13 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 	// Some web apps rely on knowing HTTPS or not
 	if r.TLS != nil {
 		env["HTTPS"] = "on"
-		// and pass the protocol details like apache's mod_ssl
-		v, ok := sslProtocolToStringMap[r.TLS.Version]
+		// and pass the protocol details in a manner compatible with apache's mod_ssl
+		// (which is why they have a SSL_ prefix and not TLS_).
+		v, ok := tlsProtocolStringToMap[r.TLS.Version]
 		if ok {
 			env["SSL_PROTOCOL"] = v
 		}
-		// and pass the cipher like apache's mod_ssl
+		// and pass the cipher suite in a manner compatible with apache's mod_ssl
 		for k, v := range caddytls.SupportedCiphersMap {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by elcore** [comment]
I think we should change `ssl_*` to `tls_*`

File: caddyhttp/fastcgi/fastcgi.go [hunk 1]
```diff
@@ -323,6 +326,18 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 	// Some web apps rely on knowing HTTPS or not
 	if r.TLS != nil {
 		env["HTTPS"] = "on"
+		// and pass the protocol details like apache's mod_ssl
+		v, ok := sslProtocolToStringMap[r.TLS.Version]
+		if ok {
+			env["SSL_PROTOCOL"] = v
```
mholt: Rename to `TLS_PROTOCOL`

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -323,6 +326,18 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
...
+		}
+		// and pass the cipher like apache's mod_ssl
+		for k, v := range caddytls.SupportedCiphersMap {
+			if v == r.TLS.CipherSuite {
+				env["SSL_CIPHER"] = k
```
mholt: Likewise, `TLS_CIPHER`

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -465,3 +480,11 @@ type LogError string
 func (l LogError) Error() string {
 	return string(l)
 }
+
+// Map of supported protocols to Apache ssl_mod format
+// Note that these are slightly different from SupportedProtocols in caddytls/config.go's
+var sslProtocolToStringMap = map[uint16]string{
```
mholt: Rename to `tlsProtocolStringToMap`

Also, was the second sentence unfinished? Include punctuation; to clarify. :)

caddyhttp/httpserver/replacer.go
```diff
@@ -375,6 +376,26 @@ func (r *replacer) getSubstitution(key string) string {
 		}
 		elapsedDuration := time.Since(r.responseRecorder.start)
 		return strconv.FormatInt(convertToMilliseconds(elapsedDuration), 10)
+	case "{ssl_protocol}":
```
mholt: Change to `{tls_protocol}` or, perhaps, `{tls_version}`. Open for discussion!

caddyhttp/httpserver/replacer.go
```diff
@@ -375,6 +376,26 @@ func (r *replacer) getSubstitution(key string) string {
...
+		if r.request.TLS != nil {
+			for k, v := range caddytls.SupportedProtocols {
+				if v == r.request.TLS.Version {
+					return k
+				}
+			}
+			return "yes" // this should never happen, but guard in case
+		}
+		return "none" // because not using a secure channel
+	case "{ssl_cipher}":
```
mholt: Change to `tls_cipher` or perhaps `{tls_ciphersuite}` since technically it is not a cipher, it is a cipher suite. Maybe it doesn't matter though.

caddyhttp/httpserver/replacer.go
```diff
@@ -375,6 +376,26 @@ func (r *replacer) getSubstitution(key string) string {
...
+		if r.request.TLS != nil {
+			for k, v := range caddytls.SupportedCiphersMap {
+				if v == r.request.TLS.CipherSuite {
+					return k
+				}
+			}
+			return "UNKNOWN" // this should never happen, but guard in case
+		}
+		return "na" // "not applicable" because not using a secure connection
```
mholt: Use empty string instead of "na"

caddyhttp/httpserver/replacer.go
```diff
@@ -375,6 +376,26 @@ func (r *replacer) getSubstitution(key string) string {
...
+	case "{ssl_protocol}":
+		if r.request.TLS != nil {
+			for k, v := range caddytls.SupportedProtocols {
+				if v == r.request.TLS.Version {
+					return k
+				}
+			}
+			return "yes" // this should never happen, but guard in case
+		}
+		return "none" // because not using a secure channel
```
mholt: Return empty string instead of "none"

caddytls/config.go
```diff
@@ -532,12 +532,14 @@ var supportedKeyTypes = map[string]acme.KeyType{
...
 // Map of supported protocols.
 // HTTP/2 only supports TLS 1.2 and higher.
-var supportedProtocols = map[string]uint16{
+var SupportedProtocols = map[string]uint16{
 	"tls1.0": tls.VersionTLS10,
 	"tls1.1": tls.VersionTLS11,
 	"tls1.2": tls.VersionTLS12,
 }
 
+// NOTE: if updating the above map, also update sslProtocolToStringMap in caddyhttp/fastcgi/fastcgi.go
```
mholt: Put this in the godoc of SupportedProtocols instead. Preferably we would just use one map though. Why not just use this one?

caddyhttp/fastcgi/fastcgi.go
mholt: Although if bajillions of PHP sites expect SSL_PROTOCOL then we can leave it

omz13: I’m using SSL_PROTOCOL to maintain compatibility with, as you say, the bajillions of sites that use mod_ssl and so expect it.

caddytls/config.go
omz13: Yes, one map would be better, but the format used in Caddyfile (for the tls settings) is different from the format used by fastcgi/mod_ssl, which is why two maps are required (otherwise I would have used what was already there).

caddyhttp/httpserver/replacer.go
omz13: I don’t think it really matters (and it was chosen to maintain some sense of consistency with mod_ssl’s use of ```SSL_CIPHER```).

BTW, if you look at the documentation for the ```tls``` directive and its  ```ciphers``` keyword, that refers to ciphers and not cipher suites... but I’m guessing that anybody mucking around at that level knows the technicalities.

caddyhttp/httpserver/replacer.go
omz13: You mean ```return r.emptyValue``` or something else?

caddyhttp/httpserver/replacer.go
omz13: To avoid any arguments, alias; i.e. how about using either ```{tls_...}``` or ```{ssl_...}``` (or is an axtra case statement too much?) Discuss!

caddyhttp/fastcgi/fastcgi.go
omz13: Is there a preference for how these things should be named? I noticed some inconsistency in the caddy source and just went with something that made sense (to me, but YMMV).

caddyhttp/fastcgi/fastcgi.go
elcore: We could implement a mod_ssl compatible mode....

I do not like the use of "SSL", it is outdated

caddyhttp/fastcgi/fastcgi.go
elcore: It should be `tlsProtocolStringToMap` (to be consistent)

caddyhttp/httpserver/replacer.go
elcore: This would lead to unnecessary complication...

caddyhttp/fastcgi/fastcgi.go
mholt: I don't either, but let's just use it for now with the env vars because, well, most PHP programs are too ;) (no offense)

caddyhttp/fastcgi/fastcgi.go
mholt: Okay, two maps it is; but yes rename to `tls...`

caddyhttp/httpserver/replacer.go
mholt: Just use `tls_` :wink:

caddyhttp/httpserver/replacer.go
mholt: Fair enough, this is fine. :) But rename to `tls_`.

caddyhttp/httpserver/replacer.go
mholt: Oh! Yes, the emptyValue.

caddyhttp/httpserver/replacer.go
mholt: (Actually return emptyValue, as you pointed out)

caddytls/config.go
mholt: That's fine then -- thanks. But move it to the godoc for that map, rather than floating by itself

caddyhttp/fastcgi/fastcgi.go
omz13: “SSL” may be outdated, but there is a lot of, for example, PHP code that tests against the SSL_PROTOCOL environment variable (due to mod_ssl)... and if you want to implement something compatible with mod_ssl you have you use what it uses vis-à-vis env vars...

mholt: Yeah; to clarify, I meant let's use "SSL_" for now, to be compatible, since that is the point of setting the environment variables. But we can/should be able to use TLS everywhere else.

omz13:
@mholt The last commit (d4fd1a5) that I made should have addressed the changes you requested... please let me know if you need anything else.

francislavoie:
Could probably use some tests.

mholt:
I agree, if you find opportunity to write some tests that would be good. :) But I'll merge for now. Thanks!

PR #2066, opened 2018-03-13
httpserver: Add geoip directive (closes #1819)

### LINKED ISSUE #1819: geoip functionality
I'm currently using nginx and its geoip module to add the `X-Forwarded-For-Latitude` and `X-Forwarded-For-Longitude` headers to requests before passing them to my app server.  I love Caddy and its automatic https integration, and would like to use it instead of nginx.  However, I don't see any geoip functionality mentioned in the Caddy documentation.  Does this exist and I'm missing it?  If not, can it be added?

kodnaplakal:
Add geoip directive to Caddyfile for [caddy-geoip](https://github.com/kodnaplakal/caddy-geoip)

https://github.com/mholt/caddy/issues/1819

No

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -544,6 +544,7 @@ var directives = []string{
 	"grpc",      // github.com/pieterlouw/caddy-grpc
 	"gopkg",     // github.com/zikes/gopkg
 	"restic",    // github.com/restic/caddy
+	"geoip",     // github.com/kodnaplakal/caddy-geoip
 }
 
 const (
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -500,6 +500,7 @@ var directives = []string{
 	"ext",
 	"gzip",
 	"header",
+	"geoip", // github.com/kodnaplakal/caddy-geoip
 	"errors",
 	"authz",        // github.com/casbin/caddy-authz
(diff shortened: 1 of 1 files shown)


mholt:
Cool, thanks for writing this plugin!

We'll merge this, but a couple thoughts first. Instead of `set_header_*` subdirectives, how about just make [placeholders](https://caddyserver.com/docs/placeholders) that have the geoip information? Then people can just use the standard `header` directive. And you can vastly simplify your syntax to `geoip database_path` without needing a block. Then a user can just do `header Country-Code {country_code}` for example, which is a little more idiomatic.

Also then you can (and should) move the geoip directive much higher on the list. It won't actually handle requests or fiddle with the header at all by itself, it'll just set placeholder values. For an example, [see how proxy does it](https://github.com/mholt/caddy/blob/13dfffd203102fe21b454df756d9e4296fe75cfd/caddyhttp/proxy/proxy.go#L179-L181).

kodnaplakal:
Hi, thanks for feedback! I'm fixed plugin following your recommendations, please look again and tell me if need some other changes.

mholt:
Looks great!! Way more Caddy-esque. Nice job.

Since your middleware comes before the errors directive, make sure your handler doesn't create and return errors (it doesn't currently, just make sure it doesn't in the future either).

Thanks! Will go out with the next Caddy release, at which point you can publish your plugin to the website.

PR #2070, opened 2018-03-15
Enabled configurable Timeout for the proxy directive

tchaudhry91:
Allows setting an upstream timeout to a proxy via the Caddyfile.
E.x
proxy / 8.8.8.8 backend1.local {
    timeout 10
}

https://github.com/mholt/caddy/issues/2050

https://caddyserver.com/docs/proxy
New property to be added there.

### 4. Checklist

Gladly


tchaudhry91:
This is still a WIP.
Looking for validation before completing and finalizing this.

mholt:
Thanks @tchaudhry91 for submitting this! I have been really busy with the ACMEv2 stuff and other things, so I invite someone else to try to look this over if they can make a few minutes. Sorry I can't get to it right now -- we really do appreciate the PR.

theodesp:
@tchaudhry91 are you still able to finalize this task?

tchaudhry91:
@theodesp Yes, I'll finalize it. 
Just looking for some help on how to test this.
To test a timeout I'll actually have to run the test for the timeout duration and measure time. 
Is there a better way?

theodesp:
You can add small timeouts. For inspiration check how the Go authors test timeouts:

https://github.com/golang/go/blob/2951f909eae0afc4bea4067c98fef1a4f760006c/src/time/sleep_test.go

tchaudhry91:
@theodesp 
I have added the test with a small timeout.
Can you please review and let me know if more changes are needed?
Thanks!

caddyhttp/proxy/proxy_test.go
```diff
@@ -1378,11 +1403,29 @@ func (r *noopReader) Read(b []byte) (int, error) {
...
+			ReverseProxy: NewSingleHostReverseProxy(uri, "", http.DefaultMaxIdleConnsPerHost, 30*time.Second),
+		},
+	}
+	if insecure {
+		u.host.ReverseProxy.UseInsecureTransport()
+	}
+	return u
+}
+
+func newFakeUpstreamWithTimeout(name string, insecure bool, timeout time.Duration) *fakeUpstream {
```
theodesp: This is duplicate of `newFakeUpstream` except for the timeout param. It's better if you supply a timeout parameter to `newFakeUpstream` and use that value or the default.

caddyhttp/proxy/proxy_test.go
```diff
@@ -1433,21 +1478,36 @@ func newWebSocketTestProxy(backendAddr string, insecure bool) *Proxy {
...
 func newPrefixedWebSocketTestProxy(backendAddr string, prefix string) *Proxy {
 	return &Proxy{
 		Next:      httpserver.EmptyNext, // prevents panic in some cases when test fails
-		Upstreams: []Upstream{&fakeWsUpstream{name: backendAddr, without: prefix}},
+		Upstreams: []Upstream{&fakeWsUpstream{name: backendAddr, without: prefix, timeout: 30 * time.Second}},
+	}
+}
+
+// newWebSocketTestProxyWithTimeout returns a WebSocket proxy with a configurable upstream timeout
+func newWebSocketTestProxyWithTimeout(backendAddr string, insecure bool, timeout time.Duration) *Proxy {
```
theodesp: The same applies here. This is duplicate of `newPrefixedWebSocketTestProxy` except for the timeout param. It's better if you supply a timeout parameter to`newPrefixedWebSocketTestProxy` and use that value or the default.

caddyhttp/proxy/proxy_test.go
tchaudhry91: Is the approach in the latest commit acceptable?
I do not want to make timeout a dependency of creating an upstream because that will mean changing the function call in all the callers. 
This change should remove duplication and serve as the default value supplier in callers that choose not to provide a timeout.

tchaudhry91: Thanks for the input. I can refactor them. I was not quite sure on the correct approach, but what you say makes sense.
I'll make the changes.

tchaudhry91:
Suggestion from @theodesp incorporated.

caddyhttp/proxy/upstream.go
```diff
@@ -464,6 +466,15 @@ func parseBlock(c *caddyfile.Dispenser, u *staticUpstream, hasSrv bool) error {
 			return c.ArgErr()
 		}
 		u.KeepAlive = n
+	case "timeout":
+		if !c.NextArg() {
+			return c.ArgErr()
+		}
+		dur, err := time.ParseDuration(c.Val())
+		if err != nil {
+			return err
```
mholt: `return c.Err/f(...)` for a correctly-formatted, consistent parse-time error message. :)

caddyhttp/proxy/reverseproxy.go
```diff
@@ -226,15 +230,21 @@ func NewSingleHostReverseProxy(target *url.URL, without string, keepalive int) *
 		}
 	}
 
+	var dialer = *defaultDialer
```
mholt: This is strange syntax to me, let's use `dialer := *defaultDialer` instead (also, might be worth noting that this isn't a deep copy, but I guess it works in this case)

tchaudhry91:
@mholt Incorporated. Thanks!

mholt:
Great, thank you!

PR #2074, opened 2018-03-16
add plugin 'http.concat'

MrToy:
This plugin's usage is similar to https://github.com/alibaba/nginx-http-concat


mholt:
Thanks @MrToy -- can you help me understand this a bit more? Why use this instead of HTTP/2 (especially with server push)? Caddy already supports push functionality, and HTTP/2 supports multiplexed streams, so generally concatenating files is no longer necessary for performance.

MrToy:
Yeah, I think you are right ! It's not necessary now, so I'll close this pull request. 
This plugin only works for which browser does not support http/2 , just like ie9 .

PR #2075, opened 2018-03-16
httpserver.Replacer: Rework loop to ignore escaped placeholder braces

whitestrake:
Makes placeholder braces escapable by backslash character (`\`) for strings fed to a httpserver.Replacer.

It achieves this by using loops to identify the first unescaped opening brace and closing brace, then cleaning up escaped braces as it works its way through the string.

This allows, for example, JSON to be set as a `header` or a `fastcgi` env var and even allows placeholders to continue working inside that JSON.

Fixes https://github.com/mholt/caddy/issues/2063

A quick search determined that the following areas call some form of `replacer.Replace`:
- Env vars in `fastcgi`
- Log formats for `log`
- Upstream and downstream headers in `proxy`
- The location targets of both `redirect` and `rewrite`
- `httpserver.IfMatcher` when evaluating `ifCond.True()`

There may be more I wasn't able to find. Public documentation of the above should note the new escape capability.

### 4. Checklist

---

As I don't have much experience with Golang PRs yet, I'd welcome any feedback, but especially regarding:
- The method used for looping and breaking the loops
- The tests used to determine whether the brace is escaped
- The method used to clean up the escaped braces
- Any additional tests that could be helpful

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -140,6 +140,14 @@ func canLogRequest(r *http.Request) bool {
 	return false
 }
 
+// unescapeBraces finds escaped braces in s and returns
+// a string with those braces unescaped.
+func unescapeBraces(s string) string {
+	ret := strings.Replace(s, "\\{", "{", -1)
+	ret = strings.Replace(s, "\\}", "}", -1)
+	return ret
+}
+
 // Replace performs a replacement of values on s and returns
 // the string with the replaced values.
 func (r *replacer) Replace(s string) string {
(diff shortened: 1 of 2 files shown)


caddyhttp/httpserver/replacer.go
```diff
@@ -149,32 +157,59 @@ func (r *replacer) Replace(s string) string {
...
 	result := ""
+Placeholders: // process each placeholder in sequence
 	for {
-		idxStart := strings.Index(s, "{")
-		if idxStart == -1 {
-			// no placeholder anymore
-			break
+		var idxOffset, idxStart, idxEnd int
+
+		idxOffset = 0
```
mholt: This variable always already equals 0 here because the line before declares it.

caddyhttp/httpserver/replacer.go
```diff
@@ -149,32 +157,59 @@ func (r *replacer) Replace(s string) string {
...
+		for { // find first unescaped opening brace
+			searchSpace := s[idxOffset:]
+			idxStart = strings.Index(searchSpace, "{")
+			if idxStart == -1 {
+				// no more placeholders
+				break Placeholders
```
mholt: I love this feature of Go

caddyhttp/httpserver/replacer.go
```diff
@@ -149,32 +157,59 @@ func (r *replacer) Replace(s string) string {
...
+			}
+			if idxStart == 0 || searchSpace[idxStart-1:idxStart] != "\\" {
```
mholt: Simpler second condition: `searchSpace[idxStart-1] != '\\'`

caddyhttp/httpserver/replacer.go
```diff
@@ -149,32 +157,59 @@ func (r *replacer) Replace(s string) string {
...
+
+		idxOffset = 0
+		for { // find first unescaped closing brace
+			searchSpace := s[idxStart+idxOffset:]
+			idxEnd = strings.Index(searchSpace, "}")
+			if idxEnd == -1 {
+				// unpaired placeholder
+				break Placeholders
+			}
+			if idxEnd == 0 || searchSpace[idxEnd-1:idxEnd] != "\\" {
```
mholt: Also here (re: the 'simpler' comment above)

caddyhttp/httpserver/replacer.go
whitestrake: I like the idea of having `idxOffset` obviously being set to zero just before the loop as a preference for code transparency, so I'd opt to remove it from the declaration instead, if that's suitable?

whitestrake:
For requests that don't have escaped placeholders, I believe the extra code executed is:

- Three variable declarations (`idxOffset` and two `searchSpace`s, one in each brace-seeking loop)
- Two integer comparisons
- Two string slicing operations, one per brace-seeking loop
- Two string comparisons
- Two arithmetic operations
- Six calls to `strings.Replace`
- One call to `strings.TrimPrefix`

I'd absolutely bet it's slower... not sure how to measure it, though my hunch is not very much worse, since I don't believe any of those are very heavy operations.

mholt:
I won't require it, but to help answer this, you could try doing a [`Benchmark` function](https://golang.org/pkg/testing/#hdr-Benchmarks) which will tell us the performance.

Either way, I will probably merge this -- it's up to you about the benchmarks.

whitestrake:
I had a go at running those benchmarks. I use an identical setup to `TestReplace` to create the request, then I reset the timer and benchmark `repl.Replace`.

The results from the code prior to this PR:

```
goos: darwin
goarch: amd64
pkg: github.com/mholt/caddy/caddyhttp/httpserver
BenchmarkReplace-4   	  300000	      4970 ns/op	     144 B/op	       5 allocs/op
```

The results for the new code, with no escaped placeholders:

```
BenchmarkReplace-4   	  300000	      5079 ns/op	     144 B/op	       5 allocs/op
```

And with escaped JSON containing a placeholder:

```
BenchmarkReplaceEscaped-4   	  300000	      5407 ns/op	     208 B/op	      10 allocs/op
```

mholt:
@whitestrake Sure! Sounds good! Performs looks great. Commit those benchmark tests, they may be useful in the future. Then I'll merge.

PR #2079, opened 2018-03-23
Caddy telemetry: a global, server-side perspective of the health of the Internet

mholt:

Implements telemetry as a series of reports to a remote server which aggregates the global data set and makes it queryable.

Please try out this branch! We're trying to build a data set that observes the health of the Internet from a server-side perspective. Such a global, non-proprietary data set does not yet exist, and this has deep implications for future research and development.

This PR still under very active development, but should generally be functional, so try it out! The more instances we have running this branch, the better the final solution will be. Participants in this project will be able to view and query the database while development is ongoing.

(The original PR was #2027 but I couldn't figure out how to keep that one when the branch got renamed.)

For more information about this project, see [The Caddy Telemetry Project on the forum](https://caddy.community/t/the-caddy-telemetry-project/3224?u=matt) and our recent announcement blog post: https://caddyserver.com/blog/caddy-0_11-telemetry

**This PR is not the right place to discuss ideas or concerns about telemetry; please [read this blog post](https://caddyserver.com/blog/caddy-0_11-telemetry) for instructions on where to discuss telemetry. We invite anyone interested to join us there and participate.** This PR is specifically for implementation, building, testing, and usage; i.e. "the code." It's very important that we keep the discussion together since there are several parties contributing to this. Thanks!

We'll need at least a whole new page in the docs describing how telemetry works and what exactly is reported.

### 4. Checklist


caddy/caddymain/run.go
```diff
@@ -291,4 +358,6 @@ var (
 	gitCommit        string // git rev-parse HEAD
 	gitShortStat     string // git diff-index --shortstat
 	gitFilesModified string // git diff-index --name-only HEAD
+
+	enableTelemetry = true
```
Xe: This defaults to enabled.

telemetry/collection.go
```diff
@@ -0,0 +1,245 @@
...
+		panic("already initialized")
+	}
+	if str := instanceID.String(); str == "" ||
+		str == "00000000-0000-0000-0000-000000000000" {
+		panic("empty UUID")
+	}
+	instanceUUID = instanceID
+	disabledMetricsMu.Lock()
+	for _, key := range disabledMetricsKeys {
+		disabledMetrics[key] = false
```
omz13: The whole premise is that everything is enabled unless explicitly disabled, i.e. opt-out. Wouldn't it be far better to be opt-in instead?

Xe: I've tried to talk with him in DM about this, he seems dead set on this course of action. I'm looking for another HTTP server in the meantime.

elcore: @Xe your reaction is childish!!

caddy/caddymain/run.go
```diff
@@ -87,6 +89,13 @@ func Run() {
 		})
 	}
 
+	// initialize telemetry client
+	if enableTelemetry {
+		initTelemetry()
+	} else if disabledMetrics != "" {
+		mustLogFatalf("[ERROR] Cannot disable specific metrics because telemetry is disabled")
```
Xe: Please explain how this is an unrecoverable state. If you cannot disable specific metrics because telemetry is disabled, who cares?

mholt: > If you cannot disable specific metrics because telemetry is disabled, who cares?

It's a logical error. To not get an error is to imply that disabling metrics worked, which implies that telemetry is enabled, which is a contradiction.

omz13: For these situations, rather than labeling as `[ERROR]`, it might be worth using something more appropriate (me, I tend to use `[INSANITY]`, and if it ever appears its easy to trap for and know that Things Have Gone Really Wrong).

elcore: That's not appropriate @omz13

caddy.go
```diff
@@ -869,7 +873,7 @@ func Stop() error {
 // explicitly like a common local hostname. addr must only
 // be a host or a host:port combination.
 func IsLoopback(addr string) bool {
-	host, _, err := net.SplitHostPort(addr)
+	host, _, err := net.SplitHostPort(strings.ToLower(addr))
```
elcore: Could you please explain why you changed that?

caddy/caddymain/run.go
```diff
@@ -266,18 +295,86 @@ func setCPU(cpu string) error {
...
+		err := ioutil.WriteFile(uuidFilename, []byte(id.String()), 0600) // human-readable as a string
+		if err != nil {
+			log.Printf("[ERROR] Persisting instance UUID: %v", err)
+		}
+		return id
+	}
+
+	var id uuid.UUID
+
+	// load UUID from storage, or create one if we don't have one
```
elcore: This is a very complex statement, it may be worthwhile in the future to find a better solution.

caddyhttp/httpserver/mitm.go
```diff
@@ -97,6 +106,13 @@ func (h *tlsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if checked {
 		r = r.WithContext(context.WithValue(r.Context(), MitmCtxKey, mitm))
+		if mitm {
```
elcore: I would have used something like this to reduce complexity.

```
switch checked {
case true:
	r = r.WithContext(context.WithValue(r.Context(), MitmCtxKey, mitm))

	if mitm {
		go telemetry.AppendUnique("http_mitm", "likely")
	} else {
		go telemetry.AppendUnique("http_mitm", "unlikely")
	}
default:
	go telemetry.AppendUnique("http_mitm", "unknown")
}
```

caddytls/handshake.go
```diff
@@ -97,7 +99,27 @@ func (cg configGroup) GetConfigForClient(clientHello *tls.ClientHelloInfo) (*tls
 //
 // This method is safe for use as a tls.Config.GetCertificate callback.
 func (cfg *Config) GetCertificate(clientHello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	if ClientHelloTelemetry && len(clientHello.SupportedVersions) > 0 {
+		// If no other plugin (such as the HTTP server type) is implementing ClientHello telemetry, we do it.
+		// NOTE: The values in the Go standard lib's ClientHelloInfo aren't guaranteed to be in order.
```
elcore: Do you have a source for that?

caddy.go
mholt: Because I do comparisons against lowercase strings like "localhost"

elcore: Okay. Thank you!

caddytls/handshake.go
mholt: Mainly because I'm not in control of that code -- the docs make no such guarantees as far as I know. And the Go standard lib doesn't necessarily use the "raw" values as extracted from the wire; they'll make slight modifications for convenience to most people; but for this case it's necessary to know that the values obtained may not be 100% representative of the how the client sent the data.

telemetry/telemetry.go
```diff
@@ -0,0 +1,406 @@
...
+// Since the buffer is reset by this, if the
+// resulting byte slice is lost, the payload is
+// gone with it.
+func makePayloadAndResetBuffer() ([]byte, error) {
+	bufCopy := resetBuffer()
+
+	// encode payload in preparation for transmission
+	payload := Payload{
+		InstanceID: instanceUUID.String(),
+		Timestamp:  time.Now().UTC(),
```
wmark: Don't trust this time. Instead, get it from the TLS handshake with the telemetry endpoint.

Unless, of course, this is for a survey of clock drift(s).

telemetry/telemetry.go
```diff
@@ -0,0 +1,406 @@
...
+		list = append(list, Item{Value: k, Count: v})
+	}
+
+	return json.Marshal(list)
+}
+
+var (
+	// httpClient should be used for HTTP requests. It
+	// is configured with a timeout for reliability.
+	httpClient = http.Client{Timeout: 1 * time.Minute}
```
wmark: Allow for a custom `tls.Config` (`http.Transport{TLSClientConfig: …}` and so forth) for these reasons:

1. Use *TLS tickets* in your client. ;-)  
```go
myTLSSessionCache = tls.NewLRUClientSessionCache(0) // capacity < 1 → use a default capacity
```
2. Configure a list of known TLS cipher suites.  
Some forks of Go (like the one I use) might have suites (and curves) unknown to your endpoint. By using said list you avoid operators reporting errors for which they cannot completely describe the environment.

telemetry/telemetry.go
```diff
@@ -0,0 +1,406 @@
...
+	endpoint = "https://telemetry-staging.caddyserver.com/v1/update/"
+
+	// defaultUpdateInterval is how long to wait before emitting
+	// more telemetry data if all retires fail. This value is
+	// only used if the client receives a nonsensical value, or
+	// doesn't send one at all, or if a connection can't be made,
+	// likely indicating a problem with the server. Thus, this
+	// value should be a long duration to help alleviate extra
+	// load on the server.
+	defaultUpdateInterval = 1 * time.Hour
```
wmark: ±Add some random seconds to spread this.

telemetry/telemetry.go
wmark: * Be prepared for a new UUID every time a throwaway instance is spun up.
 * Keep in mind that some operators will constrain the process `caddy`, not allow any outgoing connections (other than to a few ACME hosts perhaps) or/and writing to disk (does it break the part that generates the UUID? I didn't check).

mholt: Does the Go standard lib expose the timestamp of the server contained in the TLS handshake?

And, the contingencies you mentioned are acceptable I think. (And if the UUID can't be written to disk, it'll continue gracefully but with a log entry.)

telemetry/telemetry.go
wmark: You still need a contingency in case your endpoint cannot be reached.

Like when 10k EC2 instances are started at once, being the cause and victim.

telemetry/telemetry.go
wmark: I'd configure HTTP 1.0 (and no keep-alive).

The *session cache* does not interfere with HTTP/2. And I wouldn't expect it to break TLSv1.3 once that one is in.

telemetry/telemetry.go
mholt: At the _exact same moment_? Like, _within milliseconds_? That does seem unlikely. In fact, spreading them out might make them more likely to converge to about the same moment for sending updates.

telemetry/telemetry.go
wmark: Re time: come to think of it, leave as-is and just log with the endpoint's local datetime. A survey of clock drift is a good thing to have given how many short-lived tokens are used.

telemetry/telemetry.go
wmark: That's a *yes* to your question expressing disbelief. FYI, some gateways/routers do hold back and batch messages some milliseconds.

In any case, spreading this some seconds will do no harm, and as you said you avoid the convergence by adding variance once the client has reached the endpoint.

telemetry/telemetry.go
mholt: Sounds good, I think I agree!

telemetry/telemetry.go
mholt: Well, I was referring more to HTTP/2 needing to be specially configured if creating your own tls.Config on the client-side: `// If non-nil, HTTP/2 support may not be enabled by default.` -- it doesn't really say what is required to enable h2.

But I'm down for no keepalive (the receiving server will enforce this too) -- since I don't really see the benefit of it in this case.

Thanks for your reviews/feedback, Mark!

telemetry/telemetry.go
mholt: Fair enough; I've revised this locally with:

```go
reply.NextUpdate = defaultUpdateInterval + time.Duration(rand.Intn(int(1*time.Minute)))
```

and will push soon.

telemetry/telemetry.go
wmark: TBH, I wouldn't care if this is HTTP/2 or not. It's one request every so often (1h I gather). Scenarios:

* Use HTTP/2, but go through the whole TLS handshake every time (+how many roundtrips is that‽).
* HTTP 1.0, but resumed sessions. IW on Linux is 10 or higher, giving you at least 12000 octets in one go. Better encoding of HTTP headers could be negligible here, and sth. like head-of-line–blocking is irrelevant when just pushing some JSON.

elcore:
```
The DefaultTransport supports HTTP/2.
// To explicitly enable HTTP/2 on a transport, use golang.org/x/net/http2
// and call ConfigureTransport. See the package docs for more about HTTP/2.
```


https://github.com/golang/go/blob/cd1976dbef0a4a474cccb9e7fcb1e9cf2310df66/src/net/http/transport.go#L72-L74

mholt: Thanks Eldin. I'll look into that, I guess it's true we don't really need h2 for this.

That's a clever trick -- although I think we have other ways to measure that too.

elcore:
> I'd log the hostid as well. It's all zeroes within containers (checked with Docker and rkt). You can use this to estimate how many instances are run within those, and if need be clean up your aggregated data.

I think that's a bit too much, Caddy does not need the machine id/hostid.

wmark:
Then as ternary field perhaps, like `{isContainer: yes/unknown}`.

mholt:
Ooo, I love that idea, and I think will be really valuable. From a support perspective, containers add a lot of complexity with our issues and I think knowing that would be really helpful.

If I don't add that to this PR, it'll be added soon after merging I think.

telemetry/telemetry.go
elcore: > I guess it's true we don't really need h2 for this.

We don't really need it, but it is nice to have and h2 is the future.

elcore:
> Then as ternary field perhaps, like {isContainer: yes/unknown}.

That's okay.

telemetry/telemetry.go
```diff
@@ -0,0 +1,413 @@
...
+	if err != nil {
+		log.Printf("[ERROR] Sending telemetry: %v", err)
+	}
+}
+
+// emit sends an update to the telemetry server.
+// Set final to true if this is the last call to emit.
+// If final is true, no future updates will be scheduled.
+// Otherwise, the next update will be scheduled.
+func emit(final bool) error {
```
elcore: I would break down this function.

mholt:
Great, thank you everyone for your feedback and reviews! :tada: Will be doing a release in coming days.

francislavoie:
@wmark re `hostid` that's not strictly true, it depends on the container and what base image it uses. In one of my images, I'm using `phusion/baseimage` and I get a non-0 value, but in most of my alpine images, I do get 0.

francislavoie:
Nope, I get `14ac0c00` on one machine, `12ac0900` on the other, same container (they were built independently of eachother though, same base image). Ubuntu host for both. :man_shrugging: FWIW I don't actually like that base image that much, I just happen to be using it for a random project and never bothered to swap it out cause it's just a utils thing for me (I don't even run Caddy in it, just saw your comment and wanted to test it out and got a non-0 on the first one I tried it on :stuck_out_tongue: )

mholt:
While we're on the topic, is there a dependable way to know if you're running in a container? That metric would be extremely valuable information for debugging, since many issues relate to being run in a container.

wmark:
Docker creates a file `/.dockerenv` (empty).

With *rkt* and *systemd-nspawn* you'd see a `/run/systemd/nspawn/incoming` (directory).

And a `/proc/1/exe` (a symlink) which would point to a shell (`/bin/sh`, `/bin/bash`) or `caddy` itself  will give away containers. Or if the symlink points to a binary that does not exist.

I would start with that and worry about full jails or secure enclaves later.

mholt:
That's great info, thanks Mark!

wmark:
I wouldn't use `/proc/1/cgroup` for that. For one, `/proc/1` can be locked down even in containers. And then, you can run containers without *cgroups*.

mholt:
Is that common though?

wmark:
I've no representative survey, and I guess it's not common (to lock down any `/proc/[0-9]+` not belonging to a given process descendancy). The more experienced a sysadmin is, the more likely this is the case.

Anyway, checking for a `/.dockerenv` before reading from any *cgroup* files will significantly reduce error reports with this. (Though, the *significantly* is a guess.)

PR #2082, opened 2018-03-24
Regular expression support for simple rewrite rule

abiosoft:
- Adds regexp support for simple rewrite rule. 
- Also includes an optional `not` syntax to negate the regex. Negating regex can be pain for an average user and Go's regex support makes it even more difficult. I think an optional `not` syntax should not hurt.

Reason:
- The simple rewrite rule is too basic and it is rarely used.
- Many rewrite config blocks are only used for regexp.
- Users that do not need `if` conditions do not need to open to config block, hence a cleaner Caddyfile.

Outcome:

Sample wordpress config can become a one liner.
```
rewrite not ^/wp-admin {path} {path}/ /index.php?{query}
```

No issues reported. Noticed the trend from users' Caddyfiles.

Rewrite:
 
Indication of regexp support and an additional option `not` field.
```
rewrite [not] from to... 
```

### 4. Checklist

### DIFF
--- a/caddyhttp/rewrite/rewrite_test.go
+++ b/caddyhttp/rewrite/rewrite_test.go
@@ -131,6 +131,43 @@ func TestRewrite(t *testing.T) {
 	}
 }
 
+// TestWordpress is a test for wordpress usecase.
+func TestWordpress(t *testing.T) {
+	rw := Rewrite{
+		Next: httpserver.HandlerFunc(urlPrinter),
+		Rules: []httpserver.HandlerConfig{
+			newSimpleRule(t, "^/wp-admin", "{path} {path}/ /index.php?{query}", true),
+		},
+		FileSys: http.Dir("."),
+	}
+	tests := []struct {
+		from       string
+		expectedTo string
+	}{
+		{"/wp-admin", "/wp-admin"},
+		{"/wp-admin/login.php", "/wp-admin/login.php"},
+		{"/not-wp-admin/login.php?not=admin", "/index.php?not=admin"},
+		{"/loophole", "/index.php"},
+		{"/user?name=john", "/index.php?name=john"},
+	}
+
+	for i, test := range tests {
+		req, err := http.NewRequest("GET", test.from, nil)
+		if err != nil {
+			t.Fatalf("Test %d: Could not create HTTP request: %v", i, err)
+		}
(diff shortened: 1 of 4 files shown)


caddyhttp/rewrite/rewrite_test.go
```diff
@@ -131,6 +131,45 @@ func TestRewrite(t *testing.T) {
 	}
 }
 
+// TestWordpress is a test for wordpress usecase.
+func TestWordpress(t *testing.T) {
```
elcore: Does this need to be Wordpress specific?

elcore: Does this need to be Wordpress specific?

abiosoft: not really, I was using to test while writing the code and decided to leave it there and let you all judge if it should stay.

elcore: Okay, I'll leave this to Matt :smile:

mholt:
I may not get this into the 0.10.12 release tomorrow morning, but that's okay -- we can probably get it into the next one. Thanks Abiola! I'll look this over soon. :)

caddyhttp/rewrite/rewrite_test.go
tobya: I think its fine to leave this as a test.

francislavoie:
I'm only reading this now, just asking so I understand; I have the following currently for most of my sites:

```
    rewrite {
        to {path} {path}/ /index.php?{query}
    }
```

With this change, I could now shorten it to the following, correct?

```
    rewrite / {path} {path}/ /index.php?{query}
```

Is that right? Was the above not possible before?

mholt:
And now that I think of it, I should ask: is this backwards-compatible with existing one-liner `rewrite` uses?

abiosoft:
@francislavoie yeah, that is correct. ` rewrite / {path} {path}/ /index.php?{query}` is not possible before. It does exact matching and `/` only matches requests to the site root i.e. it would not even match `/index.php`.

@mholt it is not 100% backward compatible, `from` is now with regex support. 

My argument is that the one-liner should be a sensible default and people should only open a config block for more control, but what is currently happening is no one is using the one liner and many people open a config block to only use regex.

PR #2097, opened 2018-03-30
Fix errors caught by fuzzing

### LINKED ISSUE #2096: Caddyfile Parser Crashes by Fuzzing
I thought it might be worth it to fuzz the parser of Caddyfile to find crasher. There were 31 crashers. Linked are the crashers texts and their outputs. The majority of them seem to be due to glob expansion.

The fuzzer is not getting past the import directive. I'll keep running it every now and then hoping it finds other stuff. If it doesn't, I'll have to wait until it's fixed. Also, let's brainstorm how to include the fuzzer. Should it be in its own repo along with its corpus? Should the corpus be in its own repo but the fuzzing code is here, and the corpus is included in the main repo as a submodule? Or not include fuzzing at all?

*Warning:* Lots of text....

Link to gist containing description of crashers: [https://gist.github.com/Mohammed90/ea28afa044cff5e9aacca0ad0d8ece3d](https://gist.github.com/Mohammed90/ea28afa044cff5e9aacca0ad0d8ece3d)

mholt:

Fix errors caught by fuzzing.

The fix for hanging involves limiting the number of wildcards in an
import pattern to just 1. Otherwise some patterns can expand to the
entire disk.

The other fix requires that the end string for an environment variable
expansion come after the start string.

Fix #2096 

- The `import` directive can have only one wildcard (`*` and `?`), instead of unlimited. And no `[ ]` globbing.

### 4. Checklist

/cc @Mohammed90

### DIFF
--- a/caddyfile/parse.go
+++ b/caddyfile/parse.go
@@ -263,6 +263,10 @@ func (p *parser) doImport() error {
 		} else {
 			globPattern = importPattern
 		}
+		if strings.Count(globPattern, "*") > 1 {
+			// See issue #2096 - a pattern with many wildcards can hang for too long
+			return p.Errf("Glob pattern may only contain one wildcard, but have multiple: %s", globPattern)
+		}
 		matches, err = filepath.Glob(globPattern)
 
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mholt:
Okay, well, it's been a few days and I think I can merge these fixes. Thanks for finding the bugs, @Mohammed90 !

PR #2108, opened 2018-04-07
Let "transparent" subdirective imply "Forwarded" header

simnalamburt:
Let `transparent` subdirective imply `Forwarded` header. The `Forwarded` Header is the standardized version of `X-Forwarded-For`, `X-Forwarded-Host` and `X-Forwarded-Proto` headers. Since Caddy is now one of the most popular web servers, I editted it to follow RFC7239 standards.

Alternatively, we can choose to make Caddy always serve `Forwarded` header whenever `proxy` directive is used, just like `X-Forwarded-For` header. But I was not sure if it is acceptable to the maintainers, so I hereby just append it to existing `transparent` subdirective.

- https://github.com/mholt/caddy/pull/881
- https://github.com/mholt/caddy/blob/88edca/caddyhttp/proxy/proxy.go#L347-L355
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded
- https://tools.ietf.org/html/rfc7239#section-4

- https://caddyserver.com/docs/proxy#presets

### 4. Checklist

### DIFF
--- a/caddyhttp/proxy/upstream.go
+++ b/caddyhttp/proxy/upstream.go
@@ -436,6 +436,7 @@ func parseBlock(c *caddyfile.Dispenser, u *staticUpstream, hasSrv bool) error {
 		u.upstreamHeaders.Add("Host", "{host}")
 		u.upstreamHeaders.Add("X-Real-IP", "{remote}")
 		u.upstreamHeaders.Add("X-Forwarded-Proto", "{scheme}")
+		u.upstreamHeaders.Add("Forwarded", "for={remote}; host={host}; proto={scheme}")
 	case "websocket":
 		u.upstreamHeaders.Add("Connection", "{>Connection}")
 		u.upstreamHeaders.Add("Upgrade", "{>Upgrade}")


theodesp:
I think it's not so trivial to add it to the current header. For example, read the following section:

https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/

This will end up having 2 headers one for Forwarded and one for X-Forwarded thus causing confusion.

simnalamburt:
Since `Forwarded` is a new version of `X-Forwarded-For`, I think coexistence of those two headers are inevitable to preserve backward compatibility.

And I took a read of that [nginx document](https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/), but I couldn't get which part is non-trivial. This PR just adds `Forwarded` to existing headers, and multiple presence of `Forwarded` header is well-defined by RFC7239, so it looks OK to me. Can you please elaborate it a bit more if you don't mind?

theodesp:
Then the safest thing to do is to let the clients decide if they want to add a Forwarded header in a format they choose to add. Now we add a format that may or not be suitable and they will have to remove it.  For example they can remove the `by` clause.

Also if you look at the RFC section for proper transition from X-Forwarded-For to Forwarded 
https://tools.ietf.org/html/rfc7239#section-7.4 you will see that for multiple IP addresses there should be multiple `For` clauses. So that's what I meant by not trivial.

For me I don't find a valid reason to add something on the headers now especially for the `transparent` directive which is mainly a preset.  But again maybe I'm just too resistant to change sometimes. 

Let's see what the other commiters have to say about this.

caddyhttp/proxy/upstream.go
```diff
@@ -436,6 +436,7 @@ func parseBlock(c *caddyfile.Dispenser, u *staticUpstream, hasSrv bool) error {
 		u.upstreamHeaders.Add("Host", "{host}")
 		u.upstreamHeaders.Add("X-Real-IP", "{remote}")
 		u.upstreamHeaders.Add("X-Forwarded-Proto", "{scheme}")
+		u.upstreamHeaders.Add("Forwarded", "for={remote}; host={host}; proto={scheme}")
```
tobya: As per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded if the {remote} is an IPV6 address it needs to be quoted and in square brackets [] .

simnalamburt: Oh sorry I missed that

tobya:
@abiosoft @elcore @whitestrake 
Anyone have an opinion on whether or not this is a good addition to the transparent preset.

simnalamburt:
Now I see that this change is not very trivial as I expected. Since there's no good way to know if remote IP is IPv6 or IPv4 in <https://github.com/mholt/caddy/pull/2108/commits/642cf594b33de7c6b56c2670efc36deec7631168#diff-70219fd88fe3f36834f474ce6537ed26R439>, I think I should implement this logic in https://github.com/mholt/caddy/blob/88edca65/caddyhttp/proxy/proxy.go#L347-L355 this side.

Even though I don't have enough time to update this PR for right now, I still have the willingness to update this PR if this way is considered to be OK with the caddy team. How do you think?

mholt:
Sure, go ahead and work on it if you'd like! We can find a way to make it happen, once it's RFC-compliant. Thanks!

tobya:
You can use ip.to4() to check if an ip address is v4 or v6

https://golang.org/pkg/net/#IP.To4

simnalamburt:
Closing due to inactivity. I'll reopen this if I implmente this properly later.

mholt:
Sounds good. Thanks for the effort!

PR #2111, opened 2018-04-09
Vertical align the file browser's icons.

tosone:

Vertical align the file browser's icons.

No issues reported.

No.

### 4. Checklist

### DIFF
--- a/caddyhttp/browse/setup.go
+++ b/caddyhttp/browse/setup.go
@@ -132,6 +132,20 @@ a {
 	text-decoration: none;
 }
 
+.file-obj-item {
+	position: relative;
+}
+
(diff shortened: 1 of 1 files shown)


tosone:
So, it looks like this. https://files.tosiney.com/

mholt:
We kept them aligned as they currently are to fix this problem, which arises with your changes:

screen shot 2018-04-09 at 7 42 25 am

So, while I appreciate your consideration in making this PR, I don't think we will accept it; it looks fine as-is and even looks better when filenames wrap.

francislavoie:
@mholt that should be easy to fix by just making sure the text is in a div/block which won't wrap the text on the second line to further to the left than it should.

PR #2128, opened 2018-04-21
bind: support multiple values

zhsj:

bind directive supports multiple values.

Although not entirely fix https://github.com/mholt/caddy/issues/864, but related. 

https://caddyserver.com/docs/bind

syntax change to

```
bind host...
```

### 4. Checklist

The syntax is backward compatible, so old conf won't fail.


caddyhttp/httpserver/plugin.go
```diff
@@ -368,19 +377,24 @@ func groupSiteConfigsByListenAddr(configs []*SiteConfig) (map[string][]*SiteConf
...
 		// https://caddy.community/t/wildcard-virtual-domains-with-wildcard-roots/221/5?u=matt
 
 		if conf.Addr.Port == "" {
 			conf.Addr.Port = Port
 		}
-		addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(conf.ListenHost, conf.Addr.Port))
-		if err != nil {
-			return nil, err
+		if len(conf.ListenHosts) == 0 {
+			conf.ListenHosts = append(conf.ListenHosts, "127.0.0.1")
```
mholt: What is the reason for this?

caddyhttp/httpserver/server.go
```diff
@@ -419,7 +419,7 @@ func (s *Server) serveHTTP(w http.ResponseWriter, r *http.Request) (int, error)
 
 	// we still check for ACME challenge if the vhost exists,
 	// because we must apply its HTTP challenge config settings
-	if caddytls.HTTPChallengeHandler(w, r, vhost.ListenHost) {
+	if caddytls.HTTPChallengeHandler(w, r, vhost.ListenHosts[0]) {
```
mholt: How do we know the first ListenHost is the right one for the ACME challenge?

caddyhttp/httpserver/plugin.go
```diff
@@ -239,9 +239,18 @@ func (h *httpContext) MakeServers() ([]caddy.Server, error) {
...
+					!caddy.IsLoopback(listenHost) &&
+					(caddytls.QualifiesForManagedTLS(cfg) ||
+						caddytls.HostQualifies(cfg.Addr.Host)) {
+					atLeastOneSiteLooksLikeProduction = true
+				}
+			}
+		}
 		if !atLeastOneSiteLooksLikeProduction {
 			if !caddy.IsLoopback(cfg.Addr.Host) &&
-				!caddy.IsLoopback(cfg.ListenHost) &&
```
mholt: Can you explain this change and the added lines above this?

caddyhttp/httpserver/https_test.go
```diff
@@ -43,13 +44,13 @@ func TestRedirPlaintextHost(t *testing.T) {
...
+			Host:        "foohost",
+			ListenHosts: []string{"93.184.216.34"},
 		},
 		{
-			Host:       "foohost",
-			Port:       "1234",
-			ListenHost: "93.184.216.34",
+			Host:        "foohost",
+			Port:        "1234",
+			ListenHosts: []string{"93.184.216.34"},
```
mholt: Are there any test cases that actually test multiple hosts?

caddyhttp/httpserver/plugin.go
zhsj: oh, I misread the comment above. It said 'We would add a special case here so that localhost addresses bind to 127.0.0.1 if conf.ListenHost is not already set', but it also said 'that was problematic'. I will change it.

caddyhttp/httpserver/server.go
zhsj: We can't know. What I thought is, ListenHosts just behaves like the old ListenHost value. Maybe document is the right place to tell people that the first value will be used for ACME. This makes the implemetetion much simpler.

caddyhttp/httpserver/plugin.go
zhsj: lines added: loop over cfg.ListenHosts, same logic like the old.
line removed: if cfg.ListenHosts is empty, just don't check it.

zhsj: seems i forget to ensure the second block only runs when cfg.ListenHosts is empty

zhsj:
> How does this handle IPv4 and IPv6 hosts combined?

you can just use `bind 127.0.0.1 ::1`.

mholt:
Cool, thanks, looking into this!

JonLundy:
What ever happened to this?

mholt:
Nevermind. This PR causes all HTTP requests to panic. Reverted in b7091650f82c43fd27c41e0b725baee3d9bf146f.

Please add more tests @zhsj or any other contributors who would like to see this added in.

zhsj:
@mholt does the test now fail or you have other test case? Could you provide the detail? I haven't looked at the changes made since I last comment. But I've run with this patch in production and the server runs well for months. And sorry I don't have time to touch these code now, if others want to pick up, please just do it.

mholt:
@zhsj Sure, this Caddyfile:

```
localhost
```

Causes a panic on this line: https://github.com/mholt/caddy/pull/2128/files#diff-e74f72185700e304b2047c4af2ad1a48R422

Because the slice is empty.

PR #2130, opened 2018-04-23
Fix configuration permissions in systemd integration.

smlx:

This fixes the permissions on /etc/caddy to match standard linux
permissions for /etc, and makes the Caddyfile read-only for the caddy
user.

n/a

This is a documentation change.

### 4. Checklist


fab120:
I'm using caddy launched by systemd on several servers.

The standard permission for configuration files and folders under /etc are the ones that this pr is proposing.

www-data user will be able to read configuration files but not edit them.

mholt:
Excellent. Thanks for the review!

PR #2134, opened 2018-04-24
proxy: initialize ReverseProxy.Transport earlier and fix TCP connection leak

crvv:
Fix a bug introduced in 98de336, #2070
https://github.com/mholt/caddy/commit/98de336a2130aa102e52af4d8150cb3c05101d15#diff-a71769529a289e215ba922059611256bL309

Before 98de336, `proxy` used `http.DefaultTransport` for every request.
After 98de336, `proxy` creates a new `http.Transport` for every request.
Then `caddy` will create a lot of TCP connections and won't reuse them.

I found this because `github.com/transmission/transmission` can only have 1024 file descriptors, And it encounters error `too many files open` frequently when I upgraded `caddy` to 0.10.13.

See 1.

None.

### 4. Checklist

### DIFF
--- a/caddyhttp/proxy/reverseproxy.go
+++ b/caddyhttp/proxy/reverseproxy.go
@@ -85,7 +85,6 @@ type ReverseProxy struct {
 	Director func(*http.Request)
 
 	// The transport used to perform proxy requests.
-	// If nil, http.DefaultTransport is used.
 	Transport http.RoundTripper
 
 	// FlushInterval specifies the flush interval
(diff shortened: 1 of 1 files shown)


crvv:
CI failed because of
https://github.com/golang/go/issues/25048
https://github.com/golang/lint/issues/397

mholt:
Re-running CI tests.

Pinging @tchaudhry91

tchaudhry91:
Ah! Yes, this makes sense.
I can check the broken test also tomorrow morning. Unfortunately, do not have access to my dev system right now.

crvv:
I am very glad to become a collaborator. Thanks.

PR #2151, opened 2018-05-04
proxy.AllowedPath: Re-add pre-existing trailing slashes

whitestrake:
When checking a requested path's eligibility for a proxy config, the request path is compared to a provided list of excepted paths. Since both the request path and the excepted path are `path.Clean`ed before comparison, they lose trailing slashes, making `except /foo/` effectively equivalent to `except /foo` for proxy configuration, which can be undesirable.

This PR introduces basic string checking for existing trailing slashes in either the request path or the excepted path prior to cleaning. If the result of the `path.Clean` produces a string without a trailing slash, we simply append a new one to the result.

Fixes https://github.com/mholt/caddy/issues/2107

The existing proxy documentation doesn't currently specify whether trailing slashes are counted for comparison, and since I believe there's no reason to expect they wouldn't be, this change actually brings the functionality in line with the implied functionality of the documentation. (No change needed)

### 4. Checklist


caddyhttp/proxy/upstream_test.go
```diff
@@ -153,6 +153,8 @@ func TestAllowedPaths(t *testing.T) {
 		{"/proxy//static", false},
 		{"/proxy//static//download", false},
 		{"/proxy//download", false},
+		{"/proxy/trailingslash", true},
+		{"/proxy/trailingslash/", false},
```
mholt: Can you please add a test case for a path consisting of only "/"?

whitestrake: In order to test this, I'd need to add `/` to the list of ignored sub paths as this code only executes on ignored sub paths.

That would make every possible path ignored by the proxy, causing every `AllowedPath` test to return `false`.

While I _would_ also argue "who would ever `except /` from a proxy?", I acknowledge that someone might do it sometime and it needs to not break horrifically...

Should I set up another testing function with only `/` ignored? Called `TestAllowedPathTrailingSlashOnly` or something? 🤔

whitestrake: Those are good concerns. I will revise this to use `strings.HasSuffix` and to check both that the string originally had a trailing slash _and_ no longer has one after cleaning.

I am still not sure what the best way to test those cases would be, though.

PR #2158, opened 2018-05-09
fastcgi: strip PATH_INFO from SCRIPT_FILENAME (mirroring SCRIPT_NAME)

caldwell:

Currently with fastcgi, `SCRIPT_NAME` gets `PATH_INFO` stripped from it. `SCRIPT_FILENAME`, however, does not and that seems like an oversight. This is causing problems with [fcgiwrap](https://github.com/gnosek/fcgiwrap), which tries to use `SCRIPT_FILENAME` to locate a script in the filesystem.

I didn't file an issue, I just decided to fix it :-).

This is a bug fix, so I don't think this needs any documentation changes.

### 4. Checklist


caddyhttp/fastcgi/fastcgi_test.go
```diff
@@ -238,6 +238,15 @@ func TestBuildEnv(t *testing.T) {
 	envExpected = newEnv()
 	envExpected["SCRIPT_NAME"] = "/test/fgci_test.php"
 	testBuildEnv(r, rule, fpath, envExpected)
+
+	// 6. Test SCRIPT_NAME,SCRIPT_FILENAME do not include PATH_INFO
```
francislavoie: You could probably number this test `7.` (the one before this one is already `6.` 😛 )

caldwell: Oops! Good catch. Fixed.

caldwell:
I had to update the patch a couple times to fix Windows path issues (yay appveyor). It's not the one-liner of the original but it's still pretty simple.

mholt:
Hey @caldwell , thanks for the contribution!

The way I understood these variables, [as far as PHP goes](http://www.php.net/manual/en/reserved.variables.server.php):

> **'SCRIPT_NAME'**
> Contains the current script's path. This is useful for pages which need to point to themselves. The __FILE__ constant contains the full path and filename of the current (i.e. included) file.

> **'SCRIPT_FILENAME'**
> The absolute pathname of the currently executing script.
> Note: If a script is executed with the CLI, as a relative path, such as file.php or ../file.php, $_SERVER['SCRIPT_FILENAME'] will contain the relative path specified by the user.

But Apache apparently mirrors PATH_TRANSLATED (not SCRIPT_NAME) with SCRIPT_FILENAME:

> Note: As of PHP 4.3.2, PATH_TRANSLATED is no longer set implicitly under the Apache 2 SAPI in contrast to the situation in Apache 1, where it's set to the same value as the SCRIPT_FILENAME server variable when it's not populated by Apache. This change was made to comply with the CGI specification that PATH_TRANSLATED should only exist if PATH_INFO is defined. Apache 2 users may use AcceptPathInfo = On inside httpd.conf to define PATH_INFO.

It's been several years since I've worked in this environment, though. So maybe your change is in fact correct: my brain isn't really working today. :smile: Does this seem compatible with the change?

caldwell:
I think so. The `PATH_TRANSLATED` stuff a little hard to understand just by reading, so I did a quick test on an apache server I have set up and this is what it does (the config does not have `AcceptPathInfo` set, document root is `/home/david/www`):

1. With http://example.com/echo.cgi/extra/paths:

        PATH_INFO=/extra/paths
        PATH_TRANSLATED=/home/david/www/extra/paths
        SCRIPT_FILENAME=/home/david/www/echo.cgi
        SCRIPT_NAME=/echo.cgi

2. With http://example.com/echo.cgi:

        SCRIPT_FILENAME=/home/david/www/echo.cgi
        SCRIPT_NAME=/echo.cgi

So, `PATH_TRANSLATED` is weird, but it doesn't look relevant to this change. "`PATH_TRANSLATED` is to `PATH_INFO` as `SCRIPT_FILENAME` is to `SCRIPT_NAME`" seems to be the rule.

I think what the docs you quoted are saying is that in Apache 1 the above example #2 would have `PATH_TRANSLATED` set to `/home/david/www/echo.cgi` instead of not being set at all.

caldwell:
Haven't heard any objections. Are we still good to merge?

PR #2161, opened 2018-05-10
caddy: move EmitEvent(InstanceStartupEvent, instance)

elcore:

Move `EmitEvent(InstanceStartupEvent, instance)` to `caddy.Start()`

https://github.com/hacdias/caddy-service/issues/21

n/a

### 4. Checklist


elcore:
I will close this PR. After some thought, I came to the conclusion that the plugin is responsible, as it is managing instances.

mholt:
Hmm, I actually kind of like this change though.

elcore:
> Hmm, I actually kind of like this change though.

Of course we can do it anyway, but I personally believe that the plugin should be responsible emitting these events.

tobya:
@elcore What is the status on this?  Did we decide to merge it here or leave it to plugin authors.  Can we merge or close this PR?

elcore:
We haven't decided on this

Rubyj:
@elcore Just tested this fix using the service plugin. Can confirm it does not solve the issue with caddy not starting when using service plugin.

mholt:
@elcore Can you do one more change for this PR? In `onevent/hook/config.go` we need to update `SupportedEvents`. We need to distinguish between instance startup and startup, and I think I want to use `snake_case` for `instance_startup` for its string representation.

elcore:
> Can you do one more change for this PR? In onevent/hook/config.go we need to update SupportedEvents. We need to distinguish between instance startup and startup, and I think I want to use snake_case for instance_startup for its string representation.

I have implemented your suggestion, please keep in mind that this will change the behavior of this plugin, as `on startup` used to run on `caddy.InstanceStartupEvent`!

PR #2165, opened 2018-05-12
WIP - add zstd compression

darshanime:
[WIP] - Add zstd compression

Currently, the new `zstd` package is a placeholder and just adds a new header. 
I realized that it will be very similar to the `gzip` package, with the same block in the Caddyfile etc. Instead of duplicating much of the code, would it be a better way to refactor compression as a general directive and have gzip, zstd (and something else too in the future) add just the appropriate `Writer`?

Comments, @abiosoft?

This can be done without breaking existing Caddyfiles. 

https://github.com/mholt/caddy/issues/2153

Plugin docs

### 4. Checklist


darshanime:
> Instead of duplicating much of the code, would it be a better way to refactor compression as a general directive and have gzip, zstd (and something else too in the future) add just the appropriate Writer?

comments @abiosoft @mholt?

abiosoft:
> > Instead of duplicating much of the code, would it be a better way to refactor compression as a general directive and have gzip, zstd (and something else too in the future) add just the appropriate Writer?
> 
> comments @abiosoft @mholt?

I'm in favour of this as well, the code is easily reusable. Maybe we can consider depreciating `gzip` and introducing a `compress` directive that takes `gzip`, `brotli`, `zstd` e.t.c. as config param. 

I think the main issue would be if the compression is provided by a plugin. But in this case `zstd` is gonna be part of standard Caddy plugins.

tobya:
I think we would need to do this before 1.0 and slowly deprecate the `gzip` directive.

If @mholt like the idea would you be willing to work on a `compress` directive @darshanime ?

mholt:
Yes, now's a great time to do it. Let's have `compress` (and keep `gzip` for a while so we can deprecate it gracefully) and find a way to put all the compression code into there.

Note that brotli compression is only supported for pre-compressed, static assets (mainly because brotli is expensive to compress, and no good pure-Go encoders exist last time I checked).

darshanime:
yeah sure @tobya, I'll love to work on this. 
On it! 👍

tobya:
Could we open an issue for this so we can discuss as an issue rather than on a PR

darshanime:
@tobya sure. I've raised https://github.com/mholt/caddy/issues/2173

caddyhttp/compress/zstd.go
```diff
@@ -0,0 +1,82 @@
...
+// limitations under the License.
+
+package compress
+
+import (
+	"io/ioutil"
+	"io"
+	"sync"
+
+	"github.com/DataDog/zstd"
```
elcore: This library is using `cgo`, I don't like it...

99,8 % of the code is C!!!!!!

elcore: There will be cross compile issues!!!!

mholt: Unfortunately, cgo is an unacceptable dependency for Caddy core. You'll have to find a package that doesn't use C; it has to be a pure Go implementation.

elcore: @mholt I had the same thought (see above)

darshanime:
There are only 2 official bindings (http://facebook.github.io/zstd/#other-languages) and both use cgo. We can defer `zstd` until someone implements it in pure go.
Are we merging the `compress` directive change or that is redundant now?

mholt:
I don't think there's much of a need for it yet; let's just wait until there's a Go implementation of another compression standard that is fit for a web server. Thanks!

tobya:
@darshanime Thanks for all your work on this.

mholt:
Zstd compression is implemented in Caddy 2 (see the v2 branch).

PR #2171, opened 2018-05-16
Fix TestCaddyStartStop

SmilingNavern:
Move this test to caddymain to fix import problems
Minor fixes to bypass locks and race conditions

Fixes TODO test.

-

There shouldn't be documentation changes

### 4. Checklist

### DIFF
--- a/caddy/caddymain/run_test.go
+++ b/caddy/caddymain/run_test.go
@@ -15,8 +15,12 @@
 package caddymain
 
 import (
+	"net/http"
 	"runtime"
 	"testing"
+	"time"
+
+	"github.com/mholt/caddy"
 )
 
 func TestSetCPU(t *testing.T) {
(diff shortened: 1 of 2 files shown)


SmilingNavern:
Its my first contribution to this project. There is minor change with fixed TODO test. Let me know if this is required change or shouldn't be done in a future.

mholt:
Hi @SmilingNavern - welcome to the project, and thanks for submitting a PR!

Can you help me understand, what exactly does this fix? It looks the same to me, except it moved to a different package...

SmilingNavern:
Hi, @mholt 

Okay, i will break it down as i understand it:
This test checks if caddy can start. When we use caddyfile and want to start caddy we need to have some plugins(for server_type support). When test was ran at caddy there was no plugins because they should be imported(http plugin for example). And we can't import http into caddy because of cyclic import problem.

So i moved it to caddymain as it seems most logical way to fix this problem.

Also i changed a two lines: we get testInst to correctly close instances(or we get some kind of deadlock) and added sleep to prevent bind problem.

If you need any further explanation let me know.

mholt:
Hmm, well this test was disabled because it has kind of a race condition, especially on Windows. It has nothing to do with the plugins. I think I intend to remove integration tests such as this one and put a suite of tests in a separate repository to do them more thoroughly and correctly.

SmilingNavern:
Oh, ok. So it wasn't needed after all:) But i've got problems described above after enabling this test.

So should i close this pull request?

mholt:
Yeah, sorry. I think a full integration + systems test suite will best be implemented in a new repo.

PR #2172, opened 2018-05-17
Add REQUEST_SCHEME to fastcgi envs

SmilingNavern:
Fixes https://github.com/mholt/caddy/issues/2152

This changes provides REQUEST_SCHEME env variable for fastcgi

https://github.com/mholt/caddy/issues/2152

-

### 4. Checklist

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -285,6 +285,11 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 	// Retrieve name of remote user that was set by some downstream middleware such as basicauth.
 	remoteUser, _ := r.Context().Value(httpserver.RemoteUserCtxKey).(string)
 
+	requestScheme := "http"
+	if r.TLS != nil {
+		requestScheme = "https"
+	}
+
 	// Some variables are unused but cleared explicitly to prevent
 	// the parent environment from interfering.
 	env = map[string]string{
@@ -301,6 +306,7 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 		"REMOTE_IDENT":      "", // Not used
 		"REMOTE_USER":       remoteUser,
 		"REQUEST_METHOD":    r.Method,
+		"REQUEST_SCHEME":    requestScheme,
 		"SERVER_NAME":       h.ServerName,
 		"SERVER_PORT":       h.ServerPort,
 		"SERVER_PROTOCOL":   r.Proto,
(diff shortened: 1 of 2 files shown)


SmilingNavern:
I am not sure if assignment should be inside anonymous function. It's more about code style issues.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -302,6 +308,7 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 		"REMOTE_IDENT":      "", // Not used
 		"REMOTE_USER":       remoteUser,
 		"REQUEST_METHOD":    r.Method,
+		"REQUEST_SCHEME":    requestScheme,
```
abiosoft: You don't need a separate function or block, use `r.URL.Scheme`

SmilingNavern: I thought about it, but checked another caddy code for getting scheme and did it similiar.

https://github.com/mholt/caddy/blob/95514da91bbf134baaa6b18e3e9037a5dd165d1c/caddytls/httphandler.go#L56

I am not sure that you always have scheme in url.

SmilingNavern: Should i rewrite this to use r.URL.Scheme?

abiosoft: What you have done is the safest but from my experience the url scheme is also accurate. The ACME challenge may be a special case so I'll wait for @mholt and others on this.

abiosoft:
Congrats on your first PR @SmilingNavern

SmilingNavern:
Actually it's my second pull request:) but first was almost useless:) thank you, anyway

caddyhttp/fastcgi/fastcgi.go
mholt: From the godoc on net/http.Request's URL field:

```
        // For server requests the URL is parsed from the URI
        // supplied on the Request-Line as stored in RequestURI.  For
        // most requests, fields other than Path and RawQuery will be
        // empty. (See RFC 2616, Section 5.1.2)
```

So checking r.TLS is the correct way to test for HTTPS.

caddyhttp/fastcgi/fastcgi.go
```diff
@@ -286,6 +286,12 @@ func (h Handler) buildEnv(r *http.Request, rule Rule, fpath string) (map[string]
 	// Retrieve name of remote user that was set by some downstream middleware such as basicauth.
 	remoteUser, _ := r.Context().Value(httpserver.RemoteUserCtxKey).(string)
 
+	requestScheme := "http"
+
```
mholt: Just a nit, but remove the empty line here. :)

SmilingNavern: ok:)

PR #2176, opened 2018-05-21
Add env file

SmilingNavern:
It adds ability to provide file with env variables via cli

https://github.com/mholt/caddy/issues/1941

Actually i don't know which part of documentation i should change

### 4. Checklist


SmilingNavern:
Hello,

Here is pretty naive implementation of required feature.

I have found out that you are using 'gvt' for vendoring(maybe that should be added to contribution documentation). I've added godotenv to vendor.

I am not sure about error-checking and this implementation. I follow your style with fail early if some errors happen.

There is no unit-tests cause i don't know how to provide data for this unit tests. I would be glad to hear your suggestions.

caddy/caddymain/run.go
```diff
@@ -50,6 +51,7 @@ func init() {
 	flag.StringVar(&disabledMetrics, "disabled-metrics", "", "Comma-separated list of telemetry metrics to disable")
 	flag.StringVar(&conf, "conf", "", "Caddyfile to load (default \""+caddy.DefaultConfigFile+"\")")
 	flag.StringVar(&cpu, "cpu", "100%", "CPU cap")
+	flag.StringVar(&envFile, "env-file", "", "File with environment variables to load")
```
tobya: I think I would prefer just `env` as the name of the cli directive.

SmilingNavern: Hm...i believe env-file is better name because "env" parameter is used to provide single env value usually. 

Docker run for example:
```
 % docker run --help 2>&1|grep env
  -e, --env list                       Set environment variables
      --env-file list                  Read in a file of environment variables
```

abiosoft: I can not see the use case for using `-env` to supply env vars to caddy. It makes sense for docker because the container does not have access to any env vars unless those passed to it; which is not the case with caddy. 

I vote for `-env` as well.

SmilingNavern: Oh, ok. I didn't thought about it. I will rename it and renew PR.

tobya:
Thanks for the PR

SmilingNavern:
Shall i write unit tests for this function? I havn't figured out the best way to do so, because i need to have some file already or create/delete it inside of unit test.

mholt:
Thanks for the PR. Just glanced at it for now -- why do we need to pull in a dependency for this? Isn't it just a list of KEY=VALUE pairs in a file? Looks like we're importing a dependency for a grand total of about 10 lines of code. (I don't think we need the other stuff like autoloading or ignoring lines, etc?)

Tests are always preferred, but generally you don't need to (and shouldn't) test the library. If we don't end up pulling in a new dependency, that'll probably change how you approach testing this.

I ask because we don't take new dependencies lightly at this point.

abiosoft:
> Looks like we're importing a dependency for a grand total of about 10 lines of code. (I don't think we need the other stuff like autoloading or ignoring lines, etc?)

That is a good point, considering it is mainly splitting each non-empty line by the first `=` and ignoring lines starting with `#`.

SmilingNavern:
Hello, this dependency was proposed at the initial issue and i went with naive approach to solve this problem.

I can remove this lib and rewrite it for KEY=VALUE in bash style with os.Setenv(). I don't see a problem with this.

How should i test this after rewriting? Should i go with creating file with contents inside test?

mholt:
> How should i test this after rewriting? Should i go with creating file with contents inside test?

You can just pass a string into the function and see what the output is, no need to put them in separate files.

SmilingNavern:
> You can just pass a string into the function and see what the output is, no need to put them in separate files.

Yep, but then i should change logic. I don't understand what you suggest, but i will rewrite this function with removed dependency and then decide how to do this.

SmilingNavern:
@mholt  i have pushed new version. I've reworked my implementation with simple parsing and change command line argument to env(env-file in previous version).

Check this please. If this suits you i will write tests for this

caddy/caddymain/run.go
```diff
@@ -409,13 +415,61 @@ func initTelemetry() error {
...
+		return err
+	}
+
+	envMap := make(map[string]string)
+
+	scanner := bufio.NewScanner(file)
+	var line string
+
+	for scanner.Scan() {
+		line = scanner.Text()
```
abiosoft: I think you should `TrimSpace` for the line instead.

caddy/caddymain/run.go
```diff
@@ -409,13 +415,61 @@ func initTelemetry() error {
...
 }
 
+// LoadEnvFromFile loads additional envs if file provided and exists
+//Envs in file should be in KEY=VALUE format
+func LoadEnvFromFile(envFile string) error {
+	if envFile == "" {
+		return nil
+	}
+
+	file, err := os.Open(envFile)
```
abiosoft: Add a `defer file.Close()` here to close the file after reading.

abiosoft: Sorry, add it after the error check below instead.

SmilingNavern: Ok, sorry, i missed it:)

caddy/caddymain/run.go
```diff
@@ -409,13 +415,61 @@ func initTelemetry() error {
...
+		}
+
+		//skip empty line
+		if len(line) == 0 {
+			continue
+		}
+
+		fields := strings.SplitN(line, "=", 2)
+		key := strings.TrimSpace(fields[0])
+		val := fields[1]
```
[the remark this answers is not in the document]
SmilingNavern: Ok, i will add check here for len of fields.

caddy/caddymain/run.go
SmilingNavern: Why? i want to sanitize only KEY part, and let value be everything that users want.

SmilingNavern:
I've updated my merge request. I've fixed all remarks, except trimspace. Check it please

caddy/caddymain/run.go
mholt: Extra whitespace is annoying; `export KEY=VAL      ` doesn't set a value of `"VAL      "`, it sets a value of `"VAL"` (obviously without quotes). I think trimming the space is helpful.

SmilingNavern: Oh, ok, i got your point but i need to sanitize key part as well, so i can call TrimSpace on value below? What do you think?

mholt: Actually, I suggest this with the assumption that a user can encapsulate a value in quotes, since values could also have newlines, rendering trimming of whitespace harmless.

caddy/caddymain/run.go
```diff
@@ -409,13 +415,70 @@ func initTelemetry() error {
...
+
+	envMap := make(map[string]string)
+
+	scanner := bufio.NewScanner(file)
+	var line string
+
+	for scanner.Scan() {
+		line = scanner.Text()
+
+		//skip lines starting with comment
```
mholt: Put a space between `//` and your comment -- apply to all comments

caddy/caddymain/run.go
mholt: Sigh... GitHub didn't show me your latest comment... yes, although on second thought maybe we shouldn't at least for now, unless we honor quotes around the value. But don't worry about it for now.

caddy/caddymain/run.go
```diff
@@ -409,13 +415,70 @@ func initTelemetry() error {
...
+// LoadEnvFromFile loads additional envs if file provided and exists
+//Envs in file should be in KEY=VALUE format
+func LoadEnvFromFile(envFile string) error {
+	if envFile == "" {
+		return nil
+	}
+
+	file, err := os.Open(envFile)
+
+	if err != nil {
```
mholt: Move error check up by opening the file (no empty line)

caddy/caddymain/run.go
```diff
@@ -409,13 +415,70 @@ func initTelemetry() error {
...
+		}
+
+		//skip empty line
+		if len(line) == 0 {
+			continue
+		}
+
+		fields := strings.SplitN(line, "=", 2)
+		if len(fields) != 2 {
+			return errors.New("The line should be in KEY=VALUE format")
```
mholt: For this and other occurrences, use `fmt.Errorf` and give the line number of the error.

SmilingNavern:
I will fix code-style and update my pull request in one-two hours:)

caddy/caddymain/run.go
SmilingNavern: Vscode linter actually yells at me on this one:)

SmilingNavern:
I've updated once more:) still didn't fix place with trimspace. If i understood correctly i can leave it as it is now.

caddy/caddymain/run.go
```diff
@@ -409,13 +415,71 @@ func initTelemetry() error {
...
+		}
+
+		key := strings.TrimSpace(fields[0])
+		val := fields[1]
+
+		if key == "" {
+			return fmt.Errorf("Can't parse line %d; KEY can't be empty string", lineNumber)
+		}
+		envMap[key] = val
+	}
```
mholt: Check the scanner's error after the loop: `if err := scanner.Err(); err != nil { ... }`

caddy/caddymain/run.go
```diff
@@ -409,13 +415,71 @@ func initTelemetry() error {
...
+
+	for k, v := range envMap {
+		// fail early if we can't set env
```
mholt: Comment is unnecessary I think

mholt:
Implementation is looking pretty good -- just needs some tests now. Just splice the parsing logic out into a function that takes lines in as a string and returns the map, then you can test that you get the map back that you expect. I don't think we have to test the scanner or the Setenv calls, those are better left to integration tests. You may do those too if you'd like, just be sure to clean up temporary files when done.

caddy/caddymain/run.go
SmilingNavern: Oh, ok. Didn't know about scanner error, heh, but it's obvious. I will fix this

SmilingNavern:
Ok, i start at writing unit tests.

Can i put parse function into caddymain or i should it move somewhere?

SmilingNavern:
I've added some tests for this function.
Parsing logic is in separated function now.

Check it please:) sorry for such a inconvenient PR:)

caddy/caddymain/run.go
```diff
@@ -409,13 +416,84 @@ func initTelemetry() error {
...
+	scanner := bufio.NewScanner(envInput)
+	var line string
+	lineNumber := 0
+
+	for scanner.Scan() {
+		line = scanner.Text()
+		lineNumber++
+
+		// skip lines starting with comment
+		if strings.HasPrefix(line, "#") {
```
abiosoft: It is not guaranteed that `#` will be the first character on the line, hence my suggestion for trimming the entirely line of whitespaces. You should at least trim the leading spaces or change this line to `strings.HasPrefix(strings.TrimSpace(line), "#")`.

caddy/caddymain/run.go
```diff
@@ -409,13 +416,84 @@ func initTelemetry() error {
...
+			continue
+		}
+
+		// skip empty line
+		if len(line) == 0 {
```
abiosoft: Another thing that would be handled by trimming the line of whitespaces. With this, a line containing only whitespace characters will (wrongfully) not be skipped.

caddy/caddymain/run.go
```diff
@@ -409,13 +416,84 @@ func initTelemetry() error {
...
+			continue
+		}
+
+		fields := strings.SplitN(line, "=", 2)
+		if len(fields) != 2 {
+			return nil, fmt.Errorf("Can't parse line %d; line should be in KEY=VALUE format", lineNumber)
+		}
+
+		key := strings.TrimSpace(fields[0])
```
abiosoft: considering a key cannot have a trailing whitespace anyway, this is another case that trimming the line of whitespace would've handled.

abiosoft:
Sorry for drawing us back a bit when it looks ready to get merged. Well done for what you have done so far which works but I still noticed some minor issues.

From what I can see, it looks like the attempt to cater for trailing whitespaces in the value is the reason for reluctance to trim the line of whitespaces, which is fine if we want to support trailing whitespaces.
You can either call `strings.TrimSpace` when needed as I've commented or trim the line once at the beginning. I'm fine with either.

Also consider adding a test or two that test the behaviour when whitespaces are present.

SmilingNavern:
Oh, no problem:) thank you for your comments and pointing out potential problems. I will improve my code later at night and make pull request. Maybe i should trim space on the whole line because it helps with most cases.

caddy/caddymain/run.go
SmilingNavern: Actually it can if you split line by '=' space and someone puts it like this:
```
A = B
```

then actual key would be "A " and not "A". Its pretty hard to understand and debug from the user point of view. I prefer to leave this trim space and add another one to whole line.

abiosoft: `A = B` is not a valid env syntax.

abiosoft: Since Docker is your reference point, here is what Docker gives.
screenshot from 2018-05-24 11-40-30

mholt:
Thank you both for your continued work on this!! :blush:

caddy/caddymain/run.go
SmilingNavern: Yep, you are right:) I didn't think that i can return error on this case:)

SmilingNavern:
I've updated my PR.
Add trimspace at the start of loop.
Add tests for whitespace behaviour.
Add check for whitespace in KEY.

abiosoft:
I think it's good to go now.

SmilingNavern:
@mholt @abiosoft  hey guys, are there anything that left to be fixed?:) Did i miss something?)

caddy/caddymain/run_test.go
```diff
@@ -15,8 +15,12 @@
 package caddymain
 
 import (
+	"reflect"
 	"runtime"
+	"strings"
 	"testing"
+
+	_ "github.com/mholt/caddy/caddyhttp"
```
mholt: What is this import for?

SmilingNavern: Let me check it, i believe it could be here from previous iterations:)

SmilingNavern: Fixed this import:)

caddy/caddymain/run_test.go
```diff
@@ -57,3 +59,34 @@ func TestSetCPU(t *testing.T) {
...
 }
+
+func TestParseEnvFile(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   string
+		want    map[string]string
+		wantErr bool
+	}{
+		{"parsing KEY=VALUE", "PORT=4096", map[string]string{"PORT": "4096"}, false},
```
tobya: I like your test with a comment explaining what it is tesing.

caddy/caddymain/run.go
```diff
@@ -409,13 +416,88 @@ func initTelemetry() error {
...
+
+	return nil
+}
+
+// ParseEnvFile implements parse logic for environment files
+func ParseEnvFile(envInput io.Reader) (map[string]string, error) {
+	envMap := make(map[string]string)
+
+	scanner := bufio.NewScanner(envInput)
+	var line string
```
mholt: No need to declare this outside the loop; can use `line := ...` on line 455.

SmilingNavern: Okay, i will look up for this kind of things in future:) thanks

SmilingNavern:
It's cool:) thank you) i need to re-read effective go:)

What should be done for this pull request to merge it?

SmilingNavern:
Yey! Thank you for your guidance on this:)  i'm glad that it's merged finally:)

PR #2178, opened 2018-05-23
instance restart (reload) event

sirkon:
New event type called *restart* added which is emitted at server restart (at SIGUSR1)

We are using caddy as embedded server and we need to make preparations before restarts

Nothing

### 4. Checklist

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -263,9 +263,10 @@ type EventName string
 // Define names for the various events
 const (
 	StartupEvent         EventName = "startup"
-	ShutdownEvent        EventName = "shutdown"
-	CertRenewEvent       EventName = "certrenew"
-	InstanceStartupEvent EventName = "instancestartup"
+	ShutdownEvent                  = "shutdown"
+	CertRenewEvent                 = "certrenew"
+	InstanceStartupEvent           = "instancestartup"
+	InstanceRestartEvent           = "instancerestart"
 )
 
 // EventHook is a type which holds information about a startup hook plugin.
--- a/sigtrap_posix.go
+++ b/sigtrap_posix.go
@@ -90,6 +90,7 @@ func trapSignalsPosix() {
 				purgeEventHooks()
 
 				// Kick off the restart; our work is done
+				EmitEvent(InstanceRestartEvent, nil)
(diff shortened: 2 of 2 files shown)


plugins.go
```diff
@@ -263,9 +263,10 @@ type EventName string
 // Define names for the various events
 const (
 	StartupEvent         EventName = "startup"
-	ShutdownEvent        EventName = "shutdown"
-	CertRenewEvent       EventName = "certrenew"
-	InstanceStartupEvent EventName = "instancestartup"
+	ShutdownEvent                  = "shutdown"
+	CertRenewEvent                 = "certrenew"
+	InstanceStartupEvent           = "instancestartup"
+	InstanceRestart                = "restart"
```
mholt: Can we make the string consistent with the name? "instancerestart"?

sirkon: sure, done

plugins.go
```diff
@@ -263,9 +263,10 @@ type EventName string
 // Define names for the various events
 const (
 	StartupEvent         EventName = "startup"
-	ShutdownEvent        EventName = "shutdown"
-	CertRenewEvent       EventName = "certrenew"
-	InstanceStartupEvent EventName = "instancestartup"
+	ShutdownEvent                  = "shutdown"
+	CertRenewEvent                 = "certrenew"
+	InstanceStartupEvent           = "instancestartup"
+	InstanceRestart                = "instancerestart"
```
lbguilherme: Maybe it also should be `InstanceRestartEvent` for consistency

sirkon: indeed, pushed

sigtrap_posix.go
```diff
@@ -90,6 +90,7 @@ func trapSignalsPosix() {
 				purgeEventHooks()
 
 				// Kick off the restart; our work is done
+				EmitEvent(InstanceRestart, nil)
```
francislavoie: You missed this one 😛 needs to be `InstanceRestartEvent` too

sirkon: oops, fixed :)

mholt:
Thanks!

PR #2205, opened 2018-06-16
Fix two import problems: snippets and import literals.

### LINKED ISSUE #2204: snippets don't work across files
0.10.10

Define a common snippet then use that in other imported files.

`/Caddyfile` (top-level file)
```text
(common) {
  browse
  gzip
}
import apps/*/Caddyfile
```
`/apps/123/Caddyfile`
```text
localhost:9871 {
  import common
}
```

```text
(✔) aroman-mb ~/tmp $ cat Caddyfile
(common) {
  browse
  gzip
}
import apps/*/Caddyfile
(✔) aroman-mb ~/tmp $ cat apps/123/Caddyfile
localhost:9871 {
  import common
}
(✔) aroman-mb ~/tmp $ caddy
2018/06/15 17:01:36 /Users/aroman/tmp/apps/123/Caddyfile:2 - Error during parsing: File to import not found: /Users/aroman/tmp/apps/123/common
(✘) aroman-mb ~/tmp $
```

Server starting up.

```
2018/06/15 16:59:31 /Users/aroman/tmp/apps/123/Caddyfile:2 - Error during parsing: File to import not found: /Users/aroman/tmp/apps/123/common
```

augustoroman:
Two problems are fixed by this code simplification:
1. Snippets defined in one import file are strangely not available in
   another.
2. If an imported file had a directive with an argument "import", then
   the rest of the tokens on the line would be converted to absolute
   filepaths.

An example of 2 would be the following directive in an imported file:
```
server {
    basicauth / import secret
}
```

In this case, the password would actually be an absolute path to the
file 'secret' (whether or not it exists) in the directory of the imported
Caddyfile.

The problem was the blind token processing to fix import paths in the
imported tokens without considering the context of the 'import' token.

My first inclination was to just add more context (detect 'import' tokens
at the beginning of lines and check the value tokens against defined
snippets), however I eventually realized that we already do all of this
in the parser, so the code was redundant. Instead we just use the current
token's File property when importing. This works fine with imported tokens
since they already have the absolute path to the imported file!

Fixes #2204

### DIFF
--- a/caddyfile/parse_test.go
+++ b/caddyfile/parse_test.go
@@ -527,15 +527,15 @@ func testParser(input string) parser {
 }
 
 func TestSnippets(t *testing.T) {
-	p := testParser(`(common) {
-					gzip foo
-					errors stderr
-					
-				}	
-				http://example.com {
-					import common
-				}
-			`)
+	p := testParser(`
+		(common) {
+			gzip foo
+			errors stderr
+		}
+		http://example.com {
+			import common
+		}
+	`)
 	blocks, err := p.parseAll()
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


mholt:
Ooo, nice! I'm looking forward to going over these changes. Of course, I invite any other collaborators to give it a review too, since they'll probably be able to sooner than I am.

caddyfile/parse_test.go
```diff
@@ -561,3 +561,81 @@ func TestSnippets(t *testing.T) {
...
+		panic(err)
+	}
+	if err := file.Close(); err != nil {
+		panic(err)
+	}
+	return file.Name()
+}
+
+func TestImportedFilesIgnoreNonDirectiveImportTokens(t *testing.T) {
+	file2 := writeStringToTempFileOrDie(t, `
```
theodesp: Why name it `file2` when you can name it just `fileName` ?

augustoroman: sure, renamed

augustoroman:
ping?  Is this ready to merge?

mholt:
Thanks @augustoroman  -- I love how the solution was basically to delete code and add tests. If you'd like to be a collaborator on the project to help review PRs and issues and get push access, let me know . :)

PR #2211, opened 2018-06-27
fix: correct import path for filemanager plugins

hacdias:
That ^ 😄


mholt:
Thanks for maintaining these, Henrique! We love your plugins. :)

Issue #2237, opened 2018-07-25
Caddyfile in .gitignore?


sssilver:
Hi,

We're using Caddy as a dependency for our project through `dep`, and we check in our vendored dependencies. However, because of the fact that Caddy's `.gitignore` contains an entry `Caddyfile`, that particular directory under vendor/ remains ignored and doesn't get pushed into our repository. This causes the build to be broken for whoever clones the repo, as they have to delete the vendor Caddy directory and re-run `$ dep ensure` to get `caddyfile` back.

It may be intentional that the .gitignore lists `Caddyfile` with an uppercase C, and the directory itself is called `caddyfile` with a lowercase c, however on a macOS system `git` disregards the case difference.

Please advise.

Thanks,
Areg

sssilver:
This is a little confusing. Is there a difference between `caddyfile` and `Caddyfile`?

francislavoie:
Yes, `Caddyfile` is the filename for the config file Caddy uses. `caddyfile` is the name of a folder of Go code which implements parsing of the `Caddyfile` file format.

sssilver:
Ah, that makes it quite prone to problems.

I vote for `!caddyfile/`, since it disambiguates the pattern and the intentions. Shall I open a PR?

francislavoie:
Yeah, sounds good :+1:

PR #2238, opened 2018-07-25
Don't ignore the caddyfile/ package (#2237)

sssilver:
Unignores the `caddyfile/` package, which causes problems on systems with case-insensitive filesystems (such as macOS).

#2237 

None

### 4. Checklist


sssilver:
@francislavoie not sure why the build is broken. Can we merge this?

PR #2239, opened 2018-07-25
Register Web Key Directory directive

emersion:
This adds a little [WKD](https://tools.ietf.org/html/draft-koch-openpgp-webkey-service-06) plugin. WKD is a simple PGP public key discovery system.

https://github.com/emersion/caddy-wkd


mholt:
Cool, a little web application/service in a plugin! Thanks for the submission.

CI failure is unrelated.

PR #2244, opened 2018-07-30
Fix for travis-ci

SmilingNavern:
Fix golang version, because build is failing with golang 1.11beta2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ addons:
     - quic.clemente.io
 
 go:
-  - 1.x
+  - "1.10"
   - tip
 
 matrix:


SmilingNavern:
@mholt @elcore look please, i've fixed travis-ci :)

SmilingNavern:
@elcore but go 1.11 wasn't released yet? So this fix can help with ci for now and after release of go 1.11 we can fix other problems?

elcore:
> @elcore but go 1.11 wasn't released yet? So this fix can help with ci for now and after release of go 1.11 we can fix other problems?

No, it's not released. Matt should decide if we should merge this PR.

SmilingNavern:
I think it's not actual anymore)

Issue #2202, opened 2018-06-14
plugins: don't replace a HTTP response code 204 with 200


wmark:
With Caddy 0.11.0 (and `HEAD`) I've noticed using my plugin http.upload that something inside `caddy` changes my response **code 204** with no content (`StatusNoContent`).

Instead, the response code served **becomes 200** (`StatusOk`). Which is wrong, because the response does not contain a body or anything describing the response itself.

Proof, with *HTTP method* `DELETE`:
* [step 1 within the plugin](https://github.com/wmark/caddy.upload/blob/df37404c760d4f206beddffc7b58b9a5ff2391db/upload.go#L136)
* [step 2 and final step wwithin the plugin](https://github.com/wmark/caddy.upload/blob/df37404c760d4f206beddffc7b58b9a5ff2391db/upload.go#L324-L340) (yes, the comment is wrong; I'll fix that)
* [tests pass](https://github.com/wmark/caddy.upload/blob/df37404c760d4f206beddffc7b58b9a5ff2391db/upload_test.go#L574-L578) — test of the *happy path*

Steps from my README (Github won't allow to link to lines within a *Markdown* file; see below) show the infringing response:
```bash
$ curl -vX DELETE http://localhost:8000/web/path/to-release

* Connected to localhost (::1) port 8000 (#0)
> DELETE /web/path/to-release HTTP/1.1
> Host: localhost:8000
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: Caddy
< Content-Length: 0
< 
```

You can reproduce this like this (with *http.upload* being the only external plugin):
```bash
cat <<EOF >/tmp/Caddyfile
*:8000
upload / {
  to "/tmp"
}
EOF

TARGET="http://localhost:8000/web/path/to-release"
curl -T /etc/os-release "${TARGET}"
curl -vX DELETE "${TARGET}" |& grep -F "HTTP/"
```

I suggest adding a test with a mock plugin that serves a 204 as first step when addressing this.

francislavoie:
Is it possible that this is related to https://github.com/mholt/caddy/issues/2168?

francislavoie:
@wmark do you think you could help with PR #2252 to write the test as you suggested?

mholt:
I'm 99.9% sure this isn't a problem in v2, so -- fixed. Probably :) Feel free to try it out!

PR #2252, opened 2018-08-05
Write header when status code 204 (#2202)

u5surf:

when response status code 204, it can return instead of 200

#2202 

### 4. Checklist

- [y? ] I have written tests and verified that they fail without my change
no but, i tried to reproduce according to #2204 comment
```
cat <<EOF >/tmp/Caddyfile
*:8000
upload / {
  to "/tmp"
}
EOF
$ curl -T foobar http://localhost:8000/foo/bar
$ curl -vX DELETE http://localhost:8000/foo/bar | grep -F "HTTP/"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8000 (#0)
> DELETE /foo/bar HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 204 No Content
< Server: Caddy
< Date: Sun, 05 Aug 2018 08:42:53 GMT
<
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
```
i would like to write this status code test, would you mind telling me how to test status code?
- [y] I have squashed any insignificant commits
nop them
- [y] This change has comments for package types, values, functions, and non-obvious lines of code

- [y] I am willing to help maintain this change if there are issues with it later


caddyhttp/httpserver/server.go
```diff
@@ -380,6 +380,10 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	status, _ := s.serveHTTP(w, r)
 
+	if status == 204 {
+		w.WriteHeader(http.StatusNoContent)
+	}
```
mholt: Hmm, probably we should instead write whatever `status` is returned instead of coding a special case.

caddyhttp/httpserver/server.go
```diff
@@ -380,6 +380,10 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	status, _ := s.serveHTTP(w, r)
 
+	if status >= 200 {
```
francislavoie: What if the status is `3xx` or even `5xx`? Feels like it could cause unintended side-effects, but maybe I'm misunderstanding.
```suggestion
	if status >= 200 && status < 300 {
```
Thoughts @mholt?

Issue #2250, opened 2018-08-03
Add additional method OnRestartFailed() to caddy.Controller


ekleiner:
### 1. What version of Caddy are you using (`caddy -version`)?

caddy dev version (build from master)

### 2. What are you trying to do?

I am working on external plugin which makes self-tests. It sends request to server and intercepts own request while it goes through it's Handler.
While restart (SIGUSR1) it is disabled in favor to a new plugin when OnRestart() is called.
And it is removed totally when OnShutdown is called. But there is no hook for the case when reload failed. I need to have a hook like OnRestartFailed(). So every OnRestart for plugin would finish by either OnShutdow() or OnRestartFailed().

### 3. What is your entire Caddyfile?
```text
localhost:2020 {
    atepmonitor
}
```

ekleiner:
Also adding this method should simplify other plugins (like prometheus) to support the reload feature.

elcore:
Feel free to create a PR :smile:

ekleiner:
Ok. I will. Just wanted to make sure that others are ok with the changes.

elcore:
> Ok. I will. Just wanted to make sure that others are ok with the changes.

Changes are always welcome, we will review your PR and decide if it is suitable :smile:

fturib:
@ekleiner : while you are here, I think we need to think how to handle the errors during a Reload process.

- What if a OnRestart() fails ?
- I guess if the new Server fails, then we call all OnRestartFailed() on the initial instance.
- But what if one of the OnRestartFailed() fails ... ?

Maybe we should introduce some fatal error when we are not able to catch-up ...
Or not allow the event function to return an error.

fturib:
@elcore , @mholt : I guess it is worth to validate that the proposition is correct before proposing a PR. At least that nobody is against this proposition of adding an event OnRestartFailed().

Is this issue the right location to get a pre-approval ?

tobya:
Hi @fturib this is the right place to discuss and get approval for a PR.

On the face of it I would think that this is an interesting idea for a PR.  It seems @elcore suggests the same.

I wouldnt go too far down the rabbit hole of worrying about failing failures.  We can simply do that same as what happens now.

ekleiner:
Working on it.

PR #2262, opened 2018-08-09
Add callback onRestartFailed to caddy.Controller

ekleiner:
Adds callback OnRestartFailed to caddy.Controller. It is triggered if restart failed.

https://github.com/mholt/caddy/issues/2250

### 4. Checklist


ekleiner:
@elcore, I'v added the callback. Could you take a look?

elcore:
@ekleiner I am on vacation.... I can't review your PR right now.

ekleiner:
Thanks for responding. Have a good vacation.

elcore:
Thank you :smile:

fturib:
hum... gometalinter fail on checking a file that is not part of this PR ...
```
caddytls/handshake.go:1::warning: file is not gofmted with -s (gofmt)
caddytls/handshake.go:1::warning: file is not goimported (goimports)
```

ekleiner:
@fturib, I think this issue is fixed [here](https://github.com/mholt/caddy/pull/2241).

elcore:
I invite everybody to review this PR 😄

ekleiner:
I want to implement a  new callback for the case when reload failed old plugin to have an event to recover.

ekleiner:
And I think I mismatched naming here. I meant not hook, but callback.

caddy_test.go
```diff
@@ -48,6 +53,77 @@ func TestCaddyStartStop(t *testing.T) {
...
+				}
+			},
+			// If MakeServersFail.MakeServersFail is true then restart wil fail because of context failure
+			NewContext: func(inst *Instance) Context { return &TestContext{MakeServersFail: test.restartFail} },
+		})
+		c := NewTestController(serverName, "")
+		c.instance = &Instance{
+			serverType: serverName,
+			wg:         new(sync.WaitGroup),
+			Storage:    make(map[interface{}]interface{}),
```
elcore: That's not necessary (`Storage`)

caddy_test.go
```diff
@@ -48,6 +53,77 @@ func TestCaddyStartStop(t *testing.T) {
...
+		expectedCalls []string
+	}{
+		{false, []string{"OnRestart", "OnShutdown"}},
+		{true, []string{"OnRestart", "OnRestartFailed"}},
+	} {
+		serverName := fmt.Sprintf("%v", i)
+		// RegisterServerType to make successful restart possible
+		RegisterServerType(serverName, ServerType{
+			Directives: func() []string { return []string{} },
+			DefaultInput: func() Input {
```
elcore: That's not necessary (`DefaultInput`)

caddy_test.go
```diff
@@ -48,6 +53,77 @@ func TestCaddyStartStop(t *testing.T) {
...
+		})
+		c.OnShutdown(func() error {
+			calls = append(calls, "OnShutdown")
+			return nil
+		})
+
+		c.instance.Restart(CaddyfileInput{Contents: []byte(""), ServerTypeName: serverName})
+
+		if !reflect.DeepEqual(calls, test.expectedCalls) {
+			t.Errorf("Test %d: Callbacks expected: %v. But werecalled: %v", i, calls, test.expectedCalls)
```
elcore: This should be: `t.Errorf("Test %d: Callbacks expected: %v, got: %v", i, test.expectedCalls, calls)`

caddy_test.go
```diff
@@ -48,6 +53,77 @@ func TestCaddyStartStop(t *testing.T) {
...
+			NewContext: func(inst *Instance) Context { return &TestContext{MakeServersFail: test.restartFail} },
+		})
+		c := NewTestController(serverName, "")
+		c.instance = &Instance{
+			serverType: serverName,
+			wg:         new(sync.WaitGroup),
+			Storage:    make(map[interface{}]interface{}),
+		}
+
+		// Register hoks which save the calls order
```
elcore: // [...] callbacks [...]

caddy_test.go
```diff
@@ -48,6 +53,77 @@ func TestCaddyStartStop(t *testing.T) {
...
+		// RegisterServerType to make successful restart possible
+		RegisterServerType(serverName, ServerType{
+			Directives: func() []string { return []string{} },
+			DefaultInput: func() Input {
+				return CaddyfileInput{
+					Contents:       []byte(fmt.Sprintf("")),
+					ServerTypeName: serverName,
+				}
+			},
+			// If MakeServersFail.MakeServersFail is true then restart wil fail because of context failure
```
elcore: `If MakeServersFail is true then the restart will fail due to context failure`

elcore:
This is just a superficial review, I will take a deeper look after you implement the requested changes :smile:

ekleiner:
I dont' think that  AppVeyor failure "exec: "gcc": executable file not found in %PATH%" is related to my code.

elcore:
> I dont' think that AppVeyor failure "exec: "gcc": executable file not found in %PATH%" is related to my code.

It is not

mholt:
I don't know how to fix that but I want to do that before anything else, tbh.

Just that nobody else seems to be having the issue and I have no idea where to start (other than filing an issue on their GitHub, which I haven't had a chance to do properly yet).

caddy.go
```diff
@@ -180,15 +181,31 @@ func (i *Instance) ShutdownCallbacks() []error {
 // Restart replaces the servers in i with new servers created from
 // executing the newCaddyfile. Upon success, it returns the new
 // instance to replace i. Upon failure, i will not be replaced.
-func (i *Instance) Restart(newCaddyfile Input) (*Instance, error) {
+func (i *Instance) Restart(newCaddyfile Input) (inst *Instance, err error) {
```
mholt: Rather than a named return in this instance (sorry, I had to pun it), let's just declare `var` before the `defer` below.

caddy.go
```diff
@@ -180,15 +181,31 @@ func (i *Instance) ShutdownCallbacks() []error {
...
 	log.Println("[INFO] Reloading")
 
 	i.wg.Add(1)
 	defer i.wg.Done()
 
+	// if something went wrong on restart then run onRestartFailed callbacks
```
mholt: Declare `var err error` here instead of using named returns.

caddy.go
```diff
@@ -736,7 +753,8 @@ func startServers(serverList []Server, inst *Instance, restartFds map[string]res
 			if old, ok := restartFds[addr]; ok {
 				// listener
 				if old.listener != nil {
-					file, err := old.listener.File()
+					var file *os.File
+					file, err = old.listener.File()
```
mholt: Why this change?

caddy.go
```diff
@@ -751,7 +769,8 @@ func startServers(serverList []Server, inst *Instance, restartFds map[string]res
 				}
 				// packetconn
 				if old.packet != nil {
-					file, err := old.packet.File()
+					var file *os.File
+					file, err = old.packet.File()
```
mholt: ... and this one?

caddy.go
ekleiner: Because after initial declaration of "err" linter warning:
```declaration of "err" shadows declaration at ./caddy.go:716```
So, I need to use assignment of err instead of declaration.

File: caddy.go (hunk 9 above)
Comment (ekleiner) [reply caddy.go]: Same as previous.

**Comment by ekleiner** [comment]
@mholt , @elcore , any additional comments to the review?

**Comment by fturib** [comment]
@ekleiner : do you think you can rebase ?

**Comment by ekleiner** [comment]
@fturib, the code seems good to me. And from my point of view I have made all changes advised by @mholt and @elcore. Just waiting for them to merge the changes or add review comments.

**Comment by fturib** [comment]
@ekleiner : I was wondering if the branch status "This branch is out-of-date with the base branch" requires a rebase from master in order for maintainer to be able to merge, after approving the changes.

**Comment by ekleiner** [comment]
@fturib, Looks like it didn't rebase automatically. I will have a closer look in a day or 2.

**Comment by ekleiner** [comment]
@mholt, is there anything else I can do to speed up the merge of this pr?

**Comment by mholt** [comment]
@ekleiner LGTM now, thank you! The CLA just needs to be signed and then I think we're good to go.

/cc @elcore

**Comment by elcore** [comment]
@mholt LGTM too

**Comment by mholt** [comment]
@ekleiner Usually that happens then if your commit email does not match your GitHub email; make sure your commit email is associated with your GitHub account.

**Comment by ekleiner** [comment]
@mholt, thank you for information, it wasn't obvious.
The email I used for the last commit is not attached to my github account anymore. And it belongs to company where I don't work anymore. So, I don't have an access there.
Is there any workaround for such situation?

**Comment by mholt** [comment]
You could fix the email in your git config, then force-push so the only commits on this PR have right email address.

**Comment by ekleiner** [comment]
Thanks for advice. I will try to fix that at weekend.

File: caddy.go [hunk 10]
```diff
@@ -186,9 +187,26 @@ func (i *Instance) Restart(newCaddyfile Input) (*Instance, error) {
...
+	var err error
+	// if something went wrong on restart then run onRestartFailed callbacks
+	defer func() {
+		r := recover()
+		if err != nil || r != nil {
+			for _, fn := range i.onRestartFailed {
+				err = fn()
+				if err != nil {
+					return
+				}
```
rdrozhdzh: I see several issues here.
1. The actual error from `OnRestartFailed` handler is ignored, i.e. you use return value as boolean type. Consider at least logging the error.
2. If one of `OnRestartFailed` handlers fails (returns error) other handlers might not be called because you exit the loop. This means that you don't achieve your goal as described in #2250 - 
> So every OnRestart for plugin would finish by either OnShutdow() or OnRestartFailed().

 **Note**, that with current `onShutdown` related code
https://github.com/mholt/caddy/blob/d3e3fc533f7046d47d60818234f3bb50d62ea1b6/caddy.go#L233-L238
by the similar reasons both `OnShutdown()` and `OnRestartFailed()` handlers can be called for the same plugin. This could happen if `OnShutdown()` succeded for plugin _A_ but failed for plugin _B_, i.e. bot `OnShutdown()` and `OnRestartFailed()` handlers of plugin _A_ will be called
3. If panic is caught and one of `onRestartFailed` handlers fails (returns error) you won't re-issue panic at line 202, because you have exited at line 198

ekleiner: 1: I don't see why it should be ignored. This is defered function and err is assigned to a new value.
2, 3: Agree with your statements, but I am wondering what should be the most correct behavior here? Should I compose errors somehow? Or just log them right on place? @mholt, @elcore, any preferences?

ekleiner:
Couldn't resolve the license issue on weekend. I will try one more today at evening.

caddy.go
rdrozhdzh: `defer` assigns new value to variable `err` too late. Statement `return` copies the original `err` value to return value before calling deferred functions. You can use named return value if you want deferred function to change the return value. See example at https://play.golang.org/p/QCJBUPc7ePK

ekleiner: You are right. I don't use named value anymore here. So, return should be changed.

ekleiner: Made the log message. Usef format like in https://github.com/mholt/caddy/blob/master/sigtrap.go#L83 .

ekleiner:
@mholt, license is fixed, current comments are resolve. Please, review the last code change.

mholt:
Great, thank you!

PR #2265, opened 2018-08-15
Add fastcgi timeouts

SmilingNavern:
Add default timeouts for fastCGI
Timeouts still can be disabled if you explicitly set timeout to 0.

https://github.com/mholt/caddy/issues/2113

There should be documentation changes at this link https://caddyserver.com/docs/fastcgi, but i don't know how to update documentation.

### 4. Checklist

### DIFF
--- a/caddyhttp/fastcgi/setup.go
+++ b/caddyhttp/fastcgi/setup.go
@@ -27,6 +27,8 @@ import (
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
+var defaultTimeout = 60 * time.Second
+
 func init() {
 	caddy.RegisterPlugin("fastcgi", caddy.Plugin{
 		ServerType: "http",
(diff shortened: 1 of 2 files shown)


SmilingNavern:
@mholt check it please if everything that you want is actually here.

Also i've written only test for SendTimeout, maybe i should write test for all timeouts(but i believe it would be too verbose)

SmilingNavern:
Retriggering ci)

SmilingNavern:
@mholt  are there any issues with PR?

caddyhttp/fastcgi/setup.go
```diff
@@ -27,6 +27,10 @@ import (
 	"github.com/mholt/caddy/caddyhttp/httpserver"
 )
 
+//default timeout for fastcgi
+//specific timeouts can be disabled by setting it to 0 in config
+var defaultTimeout = 60 * time.Second
```
mholt: I think these comments can be deleted -- the first line is obvious, and the second one isn't necessarily true if something changes elsewhere in the code; instead that should be noted in the docs on the website which is a different matter.

SmilingNavern: Ok, that was my question) How can i contribute changes to docs?

SmilingNavern:
@mholt I have updated PR, look please. Appveyor build has failed but there is some missing gcc issue)

mholt:
Great, thank you!

I don't know what's wrong with AppVeyor :(

PR #2271, opened 2018-08-21
Update paths for filebrowser plugins

cmulk:
Filebrowser changed their github structure, this just updates the paths in plugins.go

None

### 4. Checklist

- [ X] I have written tests and verified that they fail without my change
- [X ] I have squashed any insignificant commits


hacdias:
Yees. Thanks!

PR #2274, opened 2018-08-22
caddytls: Implement CT monitor

cgatrell:
This change creates a new thread that will query the cert spotter API to get a list of certificates that have been issued for your domain names, and compares them to the ones that you have for caddy.

### 4. Checklist


JosephSummerhays:
my instance has this error

```
2018/08/23 21:38:37 Error occured on line 185 of ctmonitor
2018/08/23 21:38:37 strconv.Atoi: parsing "": invalid syntax
```
@cgatrell

mholt:
@cgatrell A good first step would be getting the tests to pass -- take a look at the CI failures and see what you can do to make the tests pass. The error messages / failures should be highlighted for convenience.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	writer := bufio.NewWriter(fileHandle)
+	defer fileHandle.Close()
+	writeValue := strconv.Itoa(currentId)
+	fmt.Fprintln(writer, writeValue)
+	writer.Flush()
+}
+
+func startMonitoring(eventType caddy.EventName, eventInfo interface{}) error {
+	go monitorCerts()
```
mholt: Check the event type to make sure this is the right event -- but let's look at this in more detail after your first pass

mholt:
@cgatrell @JosephSummerhays  I'll have a chance to review the latest push on Monday or Tuesday!

In the meantime, the PR can't be merged until the CI tests pass. So click on "Details" for those tests and make sure that all the errors/failures are resolved.

caddyhttp/caddyhttp_test.go
```diff
@@ -25,7 +25,7 @@ import (
 // ensure that the standard plugins are in fact plugged in
 // and registered properly; this is a quick/naive way to do it.
 func TestStandardPlugins(t *testing.T) {
-	numStandardPlugins := 31 // importing caddyhttp plugs in this many plugins
+	numStandardPlugins := 34 // importing caddyhttp plugs in this many plugins
 	s := caddy.DescribePlugins()
```
JosephSummerhays: We wrote (I think) only one plug in, so we should have incremented numStandardPlugins to 32. But that didn't pass all the test cases. Ultimately we changed it to 34, which is the number it expected, and it passed. We're not usre if this was a valid fix, or if it was just symptomatic of a larger problem.

mholt:
Hmm, there seem to be a lot of extra/unnecessary commits here that were pulled in. I'm not sure what happened, but only your own commits should be in the PR. You'll probably need to look up how to do a proper merge of master into your branch so that only your commits are in the PR.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	certSpotterAPIBase string = "https://api.certspotter.com/v1/issuances"
+)
+
+var (
+	filePath     string = filepath.Join(caddy.AssetsPath(), "ct_id")
+	ctConfigFile string = filepath.Join(caddy.AssetsPath(), "ct_config")
+)
+
+// CtConfig struct will allow me to get the config data from a file
+type CtConfig struct {
```
mholt: Rename to `CTConfig`

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	IncludeSubdomains bool `json:"subdomains"`
+	IncludeWildCards  bool `json:"wildCards"`
```
mholt: Make the JSON fields match the field name: `include_subdomains` and `include_wildcards`

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+const (
+	certSpotterAPIBase string = "https://api.certspotter.com/v1/issuances"
+)
+
+var (
+	filePath     string = filepath.Join(caddy.AssetsPath(), "ct_id")
+	ctConfigFile string = filepath.Join(caddy.AssetsPath(), "ct_config")
+)
+
+// CtConfig struct will allow me to get the config data from a file
```
mholt: `// CTConfig is the configuration for the CT monitor.`

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	NotAfter  string `json:"not_after"`
+	Cert      struct {
+		ID     string `json:"id"`
+		Type   string `json:"type"`
+		SHA256 string `json:"sha256"`
+		Data   string `json:"data"`
+	} `json:"cert"`
+}
+
+// QueryConfig is the configuration information for my queries.
```
mholt: Eliminate "my" from the godoc comments.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+type QueryConfig struct {
+	Subdomains bool
+	WildCards  bool
```
mholt: Instead of repeating these fields, consider embedding the `CTConfig` struct (look up embedded structs).

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+			log.Printf("[WARNING] Certificate found that caddy is not monitoring, issued by: %v\n", certSpotterCerts[key])
+		}
+	}
+}
+
+// getCaddyCerts retrieves the certificates that caddy monitors and returns them
+// as a map with the key being the bytes of the certificate cast to a string.
+func getCaddyCerts() ([]string, map[string]struct{}) {
+	var (
+		caddyCerts = make(map[string]struct{}) //caddyCerts consists of the certificates that Caddy is serving.
```
mholt: Put a space after the comment slashes: `// `

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+		caddyDNS   = make([]string, 0, 10)     //caddyDNS consists of the Subject Alternate Names that Caddy is hosting.
```
mholt: Why make with a capacity of 10? (Seems arbitrary)

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+				caddyCerts[certBytes] = struct{}{}
+			}
+		}
+		certCache.RUnlock()
+	}
+	return caddyDNS, caddyCerts
+}
+
+func getLatestIndex(fileName string) (int, error) {
+	fmt.Printf("ct_id FilePath: %v\n", fileName)
```
mholt: Remove all stray prints

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	indexBytes, err := ioutil.ReadFile(fileName)
+	if os.IsNotExist(err) {
+		log.Println("getLatestIndex failed, could not find the file")
```
mholt: Remember the error handling pattern: `return fmt.Errorf("could not load last index file: %v", err)`

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+		return 0, nil
+	}
+	if err != nil {
+		return 0, err
+	}
+	indexStr := strings.TrimSpace(string(indexBytes))
+	index, err := strconv.Atoi(indexStr)
+	if err != nil {
+		log.Println("Error converting file string to int.")
```
mholt: Error handling :wink:

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	return index, nil
+}
+
+func loadConfig() (config CtConfig) {
+	configJSON, err := os.Open(ctConfigFile)
+	if err != nil {
+		log.Printf("loadConfig error: %v", err)
+	}
+	defer configJSON.Close()
+	//var config Config
```
mholt: Remove stray comments

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+		log.Println("Error converting file string to int.")
+		return 0, err
+	}
+	return index, nil
+}
+
+func loadConfig() (config CtConfig) {
+	configJSON, err := os.Open(ctConfigFile)
+	if err != nil {
+		log.Printf("loadConfig error: %v", err)
```
mholt: You probably don't want to keep going after the file fails to open.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	defer configJSON.Close()
+	//var config Config
+	err = json.NewDecoder(configJSON).Decode(&config)
+	if err != nil {
+		log.Printf("[NOTICE] jsonDecode failed, error: %v\nUsing default values", err)
```
mholt: Remove these stray / debug log lines.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+// It then adds them to a set and returns a map of each certificate  mapped to a string
+// that contains identifying information for the cert.
+func lookUpNames(caddyCertsSANs []string, config QueryConfig) (map[string]string, int) {
+	// retrievedCerts is the bytes of a certificate mapped to the
+	// ID, issuer name, and before/after values.
+	retrievedCerts := make(map[string]string)
+
+	// biggestID is the most recent certificate id returned from CertSpotter.
+
+	// timeToWait is the amount of time you need to wait before querying again.
```
mholt: Keep these comments together

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	// retryAfter is the timeToWait value from the response headers.
+	var retryAfter string
+
+	// If the caddyCertsSANs is empty, it shouldn't run this at all.
+	for _, domainName := range caddyCertsSANs {
+		retryAfter = queryDomainName(domainName, config, &biggestID, retrievedCerts)
+	}
+	err := putLatestID(biggestID, filePath)
+	if err != nil {
+		//return fmt.Errorf("writing latest ID: %v", err)
```
mholt: Remove extraneous comments

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+		log.Printf("[WARNING] writing latest ID (%v): %v", biggestID, err)
+	}
+	timeToWait, err = strconv.Atoi(retryAfter)
+	if err != nil {
+		log.Printf(err.Error())
+		log.Print("Error retrieving time to wait, waiting 1 hour\n")
+		timeToWait = 3600
+	}
+	// returning retrievedCerts because it is created here and timeToWait
+	// so that I wait enough time before trying again.
```
mholt: Remove references to yourself in the comments :)

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+
+// monitorCerts continuously monitors the certificates that Caddy serves,
+// it queries again after the specified time.
+func monitorCerts() {
+	var queryConfig QueryConfig
+	config := loadConfig()
+	for {
+		namesToLookUp, caddyCerts := getCaddyCerts()
+		if len(namesToLookUp) == 0 {
+			log.Print("Could not retrieve DNS names from Caddy Certificate\n" +
```
mholt: Log messages should always be one line; reformat this, and prepend `[ERROR] ` to designate the type of message that this is.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// If the caddyCertsSANs is empty, it shouldn't run this at all.
+	for _, domainName := range caddyCertsSANs {
+		retryAfter = queryDomainName(domainName, config, &biggestID, retrievedCerts)
+	}
+	err := putLatestID(biggestID, filePath)
+	if err != nil {
+		//return fmt.Errorf("writing latest ID: %v", err)
+		log.Printf("[WARNING] writing latest ID (%v): %v", biggestID, err)
+	}
```
mholt: Are you sure you want the code to keep running if there is an error? Check all error handling occurrences.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+		namesToLookUp, caddyCerts := getCaddyCerts()
+		if len(namesToLookUp) == 0 {
+			log.Print("Could not retrieve DNS names from Caddy Certificate\n" +
+				"Make sure that you are serving on port 80 & 443\n" +
+				"Terminating monitorCerts.")
+			break
+		}
+		startingIndex, err := getLatestIndex(filePath)
+		if err != nil {
+			log.Printf("Error %v while getting starting index, starting at 0", err)
```
mholt: Here again, the code continues on error -- is that intended?

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	response, err := http.Get(certQuery)
+	if err != nil {
+		log.Printf("https get request failed on input %s \nError: %v", prepQuery(domainName, config), err)
+	}
+	defer response.Body.Close()
+
+	err = json.NewDecoder(response.Body).Decode(&issuanceObjects) // handle error
+	if err != nil {
+		//return fmt.Errorf("decoding json stream: %v", err)
+		log.Printf("[WARNING] error decoding json stream: %v", err)
```
mholt: :frowning_face: The error handling here was so perfect, now it's not...

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	numOfIssuanceObjects = len(issuanceObjects)
+	fmt.Printf("Length of issuanceObjects: %v\n", numOfIssuanceObjects)
```
mholt: Remove debug lines

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+				log.Printf("Decoding failed: %v", err)
+			}
+			aKey := string(bytes)
+			if _, ok := retrievedCerts[aKey]; ok {
+				continue
+			}
+			if issuance.Cert.Type == "precert" {
+				continue
+			}
+			value := "ID: " + issuance.Cert.ID + " " + issuance.Issuer.Name + " not valid before: " + issuance.NotBefore +
```
mholt: Is there a particular reason you chose this string syntax/value?

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,298 @@
...
+					log.Print("Error occurred on line 277 of ctmonitor")
+				}
+				if issuanceID > *biggestID {
+					*biggestID = issuanceID
+				}
+			}
+		}
+	} else {
+		retryAfter = response.Header.Get("Retry-After")
+		log.Printf("retryAfter: %v", retryAfter)
```
mholt: Again, remove debugging lines

caddytls/ctmonitor.go
cgatrell: It was intended, instead of completely halting the process, it is given a default value of 0.  Unless they are including expired certs, which the default is to not include expired certs, it should get them close to what we were originally going to use.

caddytls/ctmonitor.go
cgatrell: I had the default be false and false, so if the file couldn't open, it would use the defaults, would it be better to just stop execution at that point?

caddytls/ctmonitor.go
cgatrell: I wasn't sure of a better way, is it possible to do something similar to a formatted string?  If so, I could do something like: value := "ID: %v %v not valid before: %v.....", then list the variables.

caddytls/ctmonitor.go
cgatrell: I made it so that the process will terminate after comparing all of the certificates that they found.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	Index int
+}
+
+// CompareCerts compares the certificates that caddy is serving against the certificates
+// that certSpotter has found, if there are any that don't match the caddy certificates,
+// they are reported to the user.
+func CompareCerts(caddyCerts map[string]struct{}, certSpotterCerts map[string]string) {
+	for key := range certSpotterCerts {
+		if _, ok := caddyCerts[key]; !ok {
+			log.Printf("[WARNING] Certificate found that caddy is not monitoring, issued by: %v\n", certSpotterCerts[key])
```
mholt: @JosephSummerhays - we need to add an entry to telemetry here in addition to the log entry.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("error %v while converting file string to int, possible file corruption.", err)
+	}
+	return index, nil
+}
+
+func loadConfig() (config QueryConfig, err error) {
+	configJSON, err := os.Open(ctConfigFile)
+	if err != nil {
+		log.Printf("loadConfig error: %v", err)
```
mholt: I wouldn't bother logging this -- if the error is `os.IsNotExist()`, just return a default config and a nil error. If it's any other kind of error, return it immediately.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	defer configJSON.Close()
+	err = json.NewDecoder(configJSON).Decode(&config)
+	if err != nil {
+		config.CTConfig.Include_subdomains = false
+		config.CTConfig.Include_wildcards = false
+		log.Printf("[NOTICE] jsonDecode failed, error: %v\nUsing default values", err)
```
mholt: Make sure log entries are just a single line. `log.Printf("[ERROR] Decoding CT monitor configuration failed: %v", err)` is all you need here.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+		return config, fmt.Errorf("jsonDecode failed, error: %v", err)
```
mholt: Return the default config, but a nil error; since we don't want/need it to stop the whole CT monitor.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	// returning retrievedCerts because it is created here and timeToWait
+	// so that ctmonitor waits enough time before trying again.
+	return retrievedCerts, timeToWait, err
+}
+
+// monitorCerts continuously monitors the certificates that Caddy serves,
+// it queries again after the specified time.
+func monitorCerts() {
+	//var queryConfig QueryConfig
```
mholt: Remove stray commented lines

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+	}
+	err = putLatestID(biggestID, filePath)
+	if err != nil {
+		timeToWait = 3600
+		return retrievedCerts, timeToWait, fmt.Errorf("[ERROR] writing latest ID failed, err: %v", err)
+	}
+	timeToWait, err = strconv.Atoi(retryAfter)
+	if err != nil {
+		log.Printf("Error %v retrieving time to wait, waiting 1 hour\n", err)
```
mholt: `log.Printf("[ERROR] Bad Retry-After value from CT server: %v", err)`

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	//var queryConfig QueryConfig
+	config, err := loadConfig()
+	if err != nil {
+		log.Printf("[ERROR] loading config: %v", err)
+		return
+	}
+	for {
+		namesToLookUp, caddyCerts := getCaddyCerts()
+		if len(namesToLookUp) == 0 {
+			log.Printf("[ERROR]Could not retrieve DNS names from Caddy Certificate, " +
```
mholt: `log.Println("[INFO] Caddy is not serving any certificates; CT monitor disabled.")`

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+				"make sure that you are serving on port 80 & 443. " +
+				"Terminating monitorCerts.\n")
+		}
+		startingIndex, err := getLatestIndex(filePath)
+		if err != nil {
+			log.Printf("Error %v while getting starting index, starting at 0", err)
```
mholt: `log.Printf("[ERROR] Getting starting index: %v", err)`

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+			break
+		}
+		config.Query = certSpotterAPIBase
+		config.Index = startingIndex
+		fetchedCerts, pause, err := lookUpNames(namesToLookUp, config)
+
+		// CompareCerts runs even if there was an error.  If the error occurred after
+		// retrieving all of the certs, we can still compare them before quitting.
+		CompareCerts(caddyCerts, fetchedCerts)
+		if err != nil {
```
mholt: This `err` value is not assigned before the last check.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	return ioutil.WriteFile(fileName, []byte(writeValue), 0600)
+}
+
+func queryDomainName(domainName string, config QueryConfig, biggestID *int, retrievedCerts map[string]string) (retryAfter string, err error) {
+	var querySize int
+
+	for ok := true; ok; ok = querySize > 0 {
+		querySize, config.Index, retryAfter, err = getCertSpotterCerts(domainName, config, biggestID, retrievedCerts)
+		if err != nil {
+			return retryAfter, fmt.Errorf("error: %v while getting certs from certSpotter", err) // Finish error checking.
```
mholt: Remove the comment at the end of this line; doesn't seem helpful (?)

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	issuanceID = config.Index
+	var (
+		issuanceObjects []CertSpotterResponse
+		certQuery       string
+	)
+	certQuery = prepQuery(domainName, config)
+
+	response, err := http.Get(certQuery)
+	if err != nil {
+		log.Printf("https get request failed on input %s \nError: %v", prepQuery(domainName, config), err)
```
mholt: Return the error here immediately

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+		retryAfter = "3600"
+		return numOfIssuanceObjects, issuanceID, retryAfter, fmt.Errorf("decoding json stream: %v", err)
+	}
+
+	numOfIssuanceObjects = len(issuanceObjects)
+	if numOfIssuanceObjects > 0 {
+		for i, issuance := range issuanceObjects {
+			bytes, err := base64.StdEncoding.DecodeString(issuance.Cert.Data)
+			if err != nil {
+				log.Printf("Decoding failed: %v", err)
```
mholt: The body of this loop should probably be a function, then you can just return this error immediately; don't continue if this decode fails.

caddytls/ctmonitor.go
mholt: Consider `map[string]Issuance` (or something like that) instead. You can still have a String() method to encode it as a string if you want to output it in the logs.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+				continue
+			}
+			value := "ID: " + issuance.Cert.ID + " " + issuance.Issuer.Name + " not valid before: " + issuance.NotBefore +
+				" and not valid after: " + issuance.NotAfter
+			retrievedCerts[aKey] = value
+			if i == numOfIssuanceObjects-1 {
+				issuanceID, err = strconv.Atoi(issuance.ID)
+				if err != nil {
+					log.Printf(err.Error())
+					log.Print("Error occurred on line 277 of ctmonitor")
```
mholt: Return this error instead of logging it and continuing. (Also the line number is wrong anyway)

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+
+	response, err := http.Get(certQuery)
+	if err != nil {
+		log.Printf("https get request failed on input %s \nError: %v", prepQuery(domainName, config), err)
+	}
+	defer response.Body.Close()
+
+	err = json.NewDecoder(response.Body).Decode(&issuanceObjects) // handle error
+	if err != nil {
+		retryAfter = "3600"
```
mholt: Still prefer the Retry-After header even if the response body is malformed.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+					log.Printf(err.Error())
+					log.Print("Error occurred on line 277 of ctmonitor")
+				}
+				if issuanceID > *biggestID {
+					*biggestID = issuanceID
+				}
+			}
+		}
+	} else {
+		retryAfter = response.Header.Get("Retry-After")
```
mholt: Always use this header value -- not just in in the else case.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,308 @@
...
+	caddy.RegisterEventHook("ctmonitor", startMonitoring)
+}
+
+const (
+	certSpotterAPIBase string = "https://api.certspotter.com/v1/issuances"
+)
+
+var (
+	filePath     string = filepath.Join(caddy.AssetsPath(), "ct_id")
+	ctConfigFile string = filepath.Join(caddy.AssetsPath(), "ct_config")
```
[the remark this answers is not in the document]
cgatrell: The current way of doing it overwrites the file every time, is there an easy way of preserving the other contents of the file and only overwriting the one line?

caddytls/ctmonitor.go
cgatrell: The comment before explains that it runs even if there was an error, the error would have occurred previously but compareCerts will check any that it has already found.

caddytls/ctmonitor.go
cgatrell: I think that this value is only returned if the issuance endpoint returns an empty array.
From their page: 

Polling For New Issuances

When the issuances endpoint returns an empty array, it includes a Retry-After HTTP header containing a number of seconds. After the specified time has elapsed, you may retry the request. If Cert Spotter has discovered new issuances matching your query, they will be returned. You can then request subsequent pages until another empty response is returned.

cgatrell: This is what I'm doing now, to make sure that there is some waiting no matter what happens.
if retryAfter == "" {
     retryAfter = defaultWaitTime
}

caddytls/ctmonitor.go
cgatrell: Ok, I took out the new line between the comment and the function before that assigns the err value.

caddytls/ctmonitor.go
```diff
@@ -26,10 +25,7 @@ const (
 	defaultWaitTime    string = "3600"
 )
 
-var (
-	filePath     string = filepath.Join(caddy.AssetsPath(), "ct_id")
-	ctConfigFile string = filepath.Join(caddy.AssetsPath(), "ct_config")
-)
+var ctConfigFile string = filepath.Join(caddy.AssetsPath(), "ct_config")
```
mholt: Let's call it `ctmonitor_config` for accuracy and clarity.

caddytls/ctmonitor.go
```diff
@@ -132,16 +128,16 @@ func getCaddyCerts() ([]string, map[string]struct{}) {
...
 		issuanceObjects []CertSpotterResponse
 		certQuery       string
 	)
+	issuanceID = config.Index
 	certQuery = prepQuery(domainName, config)
 
 	response, err := http.Get(certQuery)
 	if err != nil {
-		log.Printf("https get request failed on input %s \nError: %v", prepQuery(domainName, config), err)
+		log.Printf("https get request failed on input %s, Error: %v", prepQuery(domainName, config), err) // TODO fix this according to changes on git.
```
mholt: TODO?

Consider this instead:

```
log.Printf("[ERROR] CT monitor: %v (url=%s)", err, certQuery)
```

caddytls/ctmonitor.go
```diff
@@ -132,16 +128,16 @@ func getCaddyCerts() ([]string, map[string]struct{}) {
 
 func getCertSpotterCerts(domainName string, config QueryConfig, biggestID *int, retrievedCerts map[string]CertSpotterResponse) (numOfIssuanceObjects int, issuanceID int, retryAfter string, err error) {
 	// issuanceObjects consists of the issuanceObjects returned from CertSpotter.
-	issuanceID = config.Index
 	var (
 		issuanceObjects []CertSpotterResponse
 		certQuery       string
 	)
+	issuanceID = config.Index
 	certQuery = prepQuery(domainName, config)
```
mholt: This function returns a full URL, not just a query string; the name is a little misleading. Maybe `reqURL` and `prepareRequestURL` instead?

caddytls/ctmonitor.go
```diff
@@ -178,34 +174,39 @@ func getCertSpotterCerts(domainName string, config QueryConfig, biggestID *int,
...
 	}
 	if err != nil {
 		return 0, err
 	}
-	indexStr := strings.TrimSpace(string(indexBytes))
-	index, err := strconv.Atoi(indexStr)
-	if err != nil {
-		return 0, fmt.Errorf("error converting file string to int: %v", err)
+	if err := json.Unmarshal(configBytes, &holder); err != nil {
+		return 0, fmt.Errorf("problem reading ct_config: %v", err)
```
mholt: "error reading configuration file: %v"

caddytls/ctmonitor.go
```diff
@@ -306,9 +306,11 @@ func prepQuery(domainName string, config QueryConfig) (query string) {
 	return query
 }
 
-func putLatestID(currentID int, fileName string) error {
-	writeValue := strconv.Itoa(currentID)
-	return ioutil.WriteFile(fileName, []byte(writeValue), 0600)
+func putLatestIndex(config QueryConfig) error {
+	//writeValue := strconv.Itoa(currentID)
```
mholt: Remove stray comment, and remove extra line below

caddytls/ctmonitor.go
```diff
@@ -306,9 +306,11 @@ func prepQuery(domainName string, config QueryConfig) (query string) {
...
+
+	jsonBlob := []byte(`{"jsonString":config.CTConfig.IncludeSubdomains. "jsonString":config.CTConfig.IncludeWildcards", "jsonString":config.CTConfig.LastID}`)
+	return ioutil.WriteFile(ctConfigFile, jsonBlob, 0600)
```
mholt: This is not correct. :)

Use `json.NewEncoder(outFile).Encode(config)`, where `outFile` is the opened ctmonitor_config file.

mholt:
For reporting potential misissuances via telemetry, I think using [Append](https://godoc.org/github.com/mholt/caddy/telemetry#Append) will do nicely; something like this maybe:

```go
go telemetry.Append("ct_unrecognized_issuances", issuance)
```

caddytls/ctmonitor.go
mholt: Here, I did something similar today:

```go
f, err := os.Create(ctConfigFile)
if err != nil {
	return fmt.Errorf("opening %s file for writing: %v", ctConfigFile)
}
defer f.Close()
err = json.NewEncoder(f).Encode(config)
if err != nil {
	return fmt.Errorf("writing CT monitor config: %v", err)
}
```

Simple, elegant.

caddytls/ctmonitor.go
cgatrell: I wanted to make sure that I had this up to date with the requested changes on here.

caddytls/ctmonitor.go
cgatrell: I like it

caddytls/ctmonitor.go
cgatrell: @mholt I added the line to the program, but telemetry wasn't recognized, am I supposed to include a telemetry header, or tell it where to find the telemetry tools inside of caddy?

caddytls/ctmonitor.go
cgatrell: How would I go about testing this portion?  Is it possible to write a test that would send the startupEvent signal and then stop it with the ShutdownEvent?

caddytls/ctmonitor.go
cgatrell: I got the telemetry package imported properly and the line for the telemetry added.

cgatrell:
The code is ready for another code review.

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,414 @@
...
+	if _, ok := ctmonitor.retrievedCerts[aKey]; ok {
+		return nil
+	}
+	if ctmonitor.issuance.Cert.Type == "precert" {
+		return nil
+	}
+
+	ctmonitor.retrievedCerts[aKey] = ctmonitor.issuance
+	if ctmonitor.lastIssuance {
+		ctmonitor.issuanceID, err = strconv.Atoi(ctmonitor.issuance.ID)
```
AGWA: Per the Cert Spotter API spec, the issuance ID is an opaque string.  Although it looks like an integer, there is no guarantee that this is the case now or in the future.  So you should just treat it as a string

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,414 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to convert issuance.ID to an int: %v", err)
+		}
+		if ctmonitor.issuanceID > ctmonitor.biggestID {
```
AGWA: Since issuance IDs are opaque, you can't make any inferences about the relation of two different IDs.  I'm curious what the motivation behind this comparison is?

caddytls/ctmonitor.go
```diff
@@ -0,0 +1,414 @@
...
+func buildMapEntry() error {
+	bytes, err := base64.StdEncoding.DecodeString(ctmonitor.issuance.Cert.Data)
+	if err != nil {
+		return fmt.Errorf("Decoding failed: %v", err)
+	}
+	aKey := string(bytes)
+	if _, ok := ctmonitor.retrievedCerts[aKey]; ok {
+		return nil
+	}
+	if ctmonitor.issuance.Cert.Type == "precert" {
```
AGWA: This seems to be ignoring precertificates?  That is not a good idea, since misissued certificates might appear in logs only as precertificates.

mholt:
Closing due to inactivity.

PR #2276, opened 2018-08-22
Cert issuer and subject string conversion fix.

laszlo-kiss:
The original code causes compilation failures:

```
server1|vagrant$ go get github.com/mholt/caddy/caddy
# github.com/mholt/caddy/caddyhttp/httpserver
src/github.com/mholt/caddy/caddyhttp/httpserver/replacer.go:464:22: cert.Issuer.String undefined (type pkix.Name has no field or method String)
src/github.com/mholt/caddy/caddyhttp/httpserver/replacer.go:476:23: cert.Subject.String undefined (type pkix.Name has no field or method String)

```

Used RawIssuer/RawSubject instead of missing .String() method.

Fixes a compilation error due to missing method call (.String()). This method is not in the "pkix.Name" structure.

https://github.com/mholt/caddy/issues/2275#issue-353008788

None.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -461,7 +461,7 @@ func (r *replacer) getSubstitution(key string) string {
 	case "{tls_client_i_dn}":
 		cert := r.getPeerCert()
 		if cert != nil {
-			return cert.Issuer.String()
+			return string(cert.RawIssuer)
 		}
 		return r.emptyValue
 	case "{tls_client_raw_cert}":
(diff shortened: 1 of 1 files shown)


mholt:
I think you need to be running Go 1.10. What version of Go are you using?

laszlo-kiss:
I'm using 1.9 (I have two, my Mac (macOS High Sierra) and Linux (Mint) both fail. The build instructions probably need to change, it calls for 1.9.

laszlo-kiss:
I see also that the 'fix' is not really a fix, since it fails the test. **Sorry about that.** I think we should drop this.

francislavoie:
Okay @laszlo-kiss, closing per your request :smile:

PR #2279, opened 2018-08-24
Update Build instructions to use Go 1.10

tobya:
Build instructions need to be updated to specifiy GO 1.10


mholt:
Thanks, but #2267 beat you to it, I think.

Issue #2306, opened 2018-09-30
Configurable `FallbackDelay` for reverse proxies


jakelucas:
This issue is somewhat related to a PR I made recently (#2305) where the default dialer that's created for reverse proxies needed the `DualStack` field set to `true`.

The reason I was looking into the issue raised in that PR was because response speed matters to me a lot, and `DualStack` being enabled really helps to bring responses to a somewhat acceptable level.

Having said that however, if Caddy could provide an extra bit of configuration for the `FallbackDelay` value in the `net.Dialer` instance then it would really help people like me tune servers to respond even faster when the response time really matters.

`FallbackDelay` is only used when `DualStack` is enabled, which since my PR it now is for reverse proxies.

When the `FallbackDelay` value is not set the `fallbackDelay` function in the standard library returns `300 * time.Millisecond` as a default.
Because of this, any use of a transport with the `net.Dialer`'s `DualStack` field set to `true` always takes at least 300ms after the timeout set in the `TLSHandshakeTimeout` has passed, or for any new connection.

I've been using Caddy for quite some time now, and always wondered why I would see these spikes in response time, and now I know it's because of this setting.
It would be useful to make the `FallbackDelay` configurable for those users (like me) where having spikes of at least 300ms delay either feels like a constant thorn in my side, or in some cases, completely unacceptable.

Currently I'm manually setting it to a more reasonable value for my applications by hard coding it myself and building from source, but a new directive would be a much nicer way to solve the problem.

Since this directly relates to reverse proxies (in my case) and reverse proxies are the only place I can see in the source that use a custom dialer anyway, I thought perhaps a new sub-directive in the `proxy` directive could work.
It also looks like configuration is set per reverse proxy as well, so I'm under the assumption it wouldn't be too difficult to add this.

Does this sound like an acceptable change, and if not what would be the issues with it?

tobya:
Personally I cant see why if you provided a PR allowing `fallbackdelay` to be configured via the caddyfile that we wouldnt be able to approve it.

Add it as a new PR and we'll take it from there.

magikstm:
PR #2309 added this feature.

I think this issue can be closed.

mholt:
Appreciating the triage work, @magikstm! Do you want to be made a collaborator?

magikstm:
Sure. I would like to be one.

PR #2309, opened 2018-10-02
Add a new `fallback_delay` sub-directive to the `proxy` directive

jakelucas:

A previous PR I made (#2305) just enabled `DualStack` in the default dialer that was being used in reverse proxies to improve initial connection speed, or connection speed after the TLS handshake timeout had already passed.

In the Go standard library's `net` package, the `net.Dialer` uses a fallback delay as part of its implementation of "Happy Eyeballs" when `DualStack` is enabled. This fallback delay is just a measure of how much of a head start the first connection gets before allowing another "racer" to attempt a second connection.

By default when this duration is not set the `fallbackDelay` function in the `net` package defaults to returning `300 * time.Millisecond`.

It would be useful to configure this fallback delay duration from a caddyfile instead of having to edit the source and recompile.

So this PR just adds a new `fallback_delay` sub-directive to the existing `proxy` directive, which gets used in the `NewSingleHostReverseProxy` function in the `reverseproxy.go` file to overwrite the default value if it's different from the one set on the default dialer.

The default dialer has been left with this value unset so it can always use the default Go standard library's implementation, since it's possible the Go standard library could change it in the future, in which case we would probably want to pick up those changes by default.

This PR relates directly to issue #2306 and is a result of a previous PR I made (#2305).

A change to the proxy documentation (https://caddyserver.com/docs/proxy) would be needed to add an explanation of the new sub-directive.

### 4. Checklist

I checked the first box about tests, but what I mean here is that I updated the existing tests, checked that they failed, and then added the new directive and tested again.

### DIFF
--- a/caddyhttp/proxy/proxy.go
+++ b/caddyhttp/proxy/proxy.go
@@ -47,6 +47,12 @@ type Upstream interface {
 	// Checks if subpath is not an ignored path
 	AllowedPath(string) bool
 
+	// Gets the duration of the headstart the first
+	// connection is given in the Go standard library's
+	// implementation of "Happy Eyeballs" when DualStack
+	// is enabled in net.Dialer.
+	GetFallbackDelay() time.Duration
+
 	// Gets how long to try selecting upstream hosts
 	// in the case of cascading failures.
 	GetTryDuration() time.Duration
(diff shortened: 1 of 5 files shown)


mholt:
Thanks for these high quality contributions! I will try to look at them soon, but it's likely another collaborator will have a chance before I do.

mholt:
Thanks for the review! That helps speed things up. And thank you again for the contribution!

PR #2315, opened 2018-10-11
tls: Add root CA to obtain certificates.

jiusanzhou:

Some words from mitmproxy's docs:

> This CA is used for on-the-fly generation of dummy certificates for each of the SSL sites that your client visits. Since your browser won’t trust the mitmproxy CA out of the box, you will see an SSL certificate warning every time you visit a new SSL domain through mitmproxy. When you are testing a single site through a browser, just accepting the bogus SSL cert manually is not too much trouble, but there are a many circumstances where you will want to configure your testing system or browser to trust the mitmproxy CA as a signing root authority.

*I'm trying to make Caddy can work like mitmproxy while I dealing with [MITM](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) things with my own plugin, but only works with modified Caddy, so I apply this pull request first.*

So, I need Caddy to load a root CA, and then generate a certificate with domain(wildcard or not) on-the-fly.

Steps: 

1. handle sub directive in tls: `root_signed`
2. load the cert and key of CA
3. generate cert with the root cert and key in the fly if needs

**Do I need to open an issue to discuss about this feature?**

1. add to a sub directive for doc of `tls`
2. add `root_signed_wild` for doc of `tls` settings block, will obtain and manage a wildcard certificate for this name by replacing the left-most label with *, ignore the DNS challenge.

For example, part of Caddyfile with enable wildcard for root signed:

```
	tls root_signed ./ssl/rootCA.crt ./ssl/rootCA.key {
		root_signed_wild
	}
```

Thing looks like,

caddy_tls_wildcard

### 4. Checklist

### DIFF
--- a/caddytls/certificates.go
+++ b/caddytls/certificates.go
@@ -196,6 +196,8 @@ func (cfg *Config) cacheUnmanagedCertificatePEMBytes(certBytes, keyBytes []byte)
 	}
 	cfg.cacheCertificate(cert)
 	telemetry.Increment("tls_manual_cert_count")
+
+	cfg.SelfCAName = cert.Names[0]
 	return nil
 }
 
--- a/caddytls/client.go
+++ b/caddytls/client.go
@@ -25,13 +25,164 @@ import (
 	"sync"
 	"time"
 
+	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/telemetry"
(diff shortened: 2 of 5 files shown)


mholt:
Hi, thank you for your pull request! This is a relatively large change; would you mind filling out the PR template so we know what it is about? What problem(s) does it solve, what issues was it discussed in, etc? Thanks

jiusanzhou:
@mholt thank you for your reminding. I have updated the comment, please feel free to review, if you have any question, let me know. Thanks a lot.

francislavoie:
Honestly this change doesn't feel like it fits with Caddy. I think a better change would be some bindings to allow a plugin to make this kind of thing possible.

Ultimately this really feels like a debugging feature, not something the general public would use. So maybe it's better to keep it in your own fork for now.

I'll close this, but feel free to keep discussing, it could be re-opened later.

PR #2323, opened 2018-10-14
new template action: TLS protocol version

henrydcase:
Introduces new template action to get TLS protocol version

None

Template action should be documented here: https://caddyserver.com/docs/template-actions

### 4. Checklist

- [x]  I have written tests and verified that they fail without my change

### DIFF
--- a/caddyhttp/httpserver/tplcontext.go
+++ b/caddyhttp/httpserver/tplcontext.go
@@ -17,6 +17,7 @@ package httpserver
 import (
 	"bytes"
 	"crypto/rand"
+	"crypto/tls"
 	"fmt"
 	"io/ioutil"
 	mathrand "math/rand"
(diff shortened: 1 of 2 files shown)


henrydcase:
let me know if this is interesting. then I'll add tests if needed

caddyhttp/httpserver/tplcontext.go
```diff
@@ -448,6 +449,20 @@ func (c Context) AddLink(link string) string {
 	return ""
 }
 
+// Resturns TLS protocol versions
```
francislavoie: Typo:
```suggestion
// Returns TLS protocol versions
```

caddyhttp/httpserver/tplcontext.go
```diff
@@ -448,6 +449,20 @@ func (c Context) AddLink(link string) string {
 	return ""
 }
 
+// Returns TLS protocol version
+func (c Context) TLSVersion() string {
+	switch c.Req.TLS.Version {
```
mholt: `c.Req.TLS` might be nil -- resulting in a panic.

henrydcase: Good point. I'll add tests and fix this. Thnx!

henrydcase:
I've decided to change implementation and reuse "GetSupportedProtocolName" to get a string with protocol version. Pro - Context.TLSVersion doesn't need to be changed after TLS 1.3 is added, Cons - instead of getting string like "1.2" we get "tls1.2"

mholt:
Thank you!

Issue #1920, opened 2017-10-13
Const Etag and Last-Modified header when using template


crvv:
### 1. What version of Caddy are you using (`caddy -version`)?
Caddy 0.10.10

### 2. What are you trying to do?
```
echo '{{.RandomString 64 64}}' > random.txt
caddy &
curl http://localhost:2015/random.txt -v
curl http://localhost:2015/random.txt -v
```

### 3. What is your entire Caddyfile?
```text
:2015 {
    templates
}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
`caddy`

### 5. Please paste any relevant HTTP request(s) here.
see 2.

### 6. What did you expect to see?
Because the responses are different, the `Etag` and `Last-Modified` should be different.

### 7. What did you see instead (give full error messages and/or log)?
The `Etag` and `Last-Modified` values didn't change for different responses.

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
see 2 and 3.

Besides `templates`, I guess some other middlewares have the same issue.

PR #2338, opened 2018-11-12
templates: delete ETag and Last-Modified headers in ServeHTTP

crvv:
In tempates middleware, delete the two headers related to cache.

#1920

None.

### 4. Checklist

### DIFF
--- a/caddyhttp/templates/templates.go
+++ b/caddyhttp/templates/templates.go
@@ -110,6 +110,10 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 		// set the actual content length now that the template was executed
 		w.Header().Set("Content-Length", strconv.Itoa(buf.Len()))
 
+		// delete the headers related to cache
+		w.Header().Del("ETag")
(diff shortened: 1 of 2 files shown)


caddyhttp/templates/templates.go
```diff
@@ -110,6 +110,10 @@ func (t Templates) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 		// set the actual content length now that the template was executed
 		w.Header().Set("Content-Length", strconv.Itoa(buf.Len()))
 
+		// delete the headers related to cache
```
mholt: Note for future work: I think a comment for the next two lines is good, but this comment just restates what is obvious from the code. Better would be to have a comment explaining WHY the next two lines are present.

francislavoie: FWIW I think it's at least more useful than just stating the obvious, because it's specific domain knowledge that ETag and Last-Modified are cache-related. Not everyone would know that. But I agree that it could explain why those headers are being deleted.

PR #2347, opened 2018-11-15
Prevent errors test fail if $GOPATH is something like /home/user/caddy

linquize:
To pass `errors` unit test when source code `$GOPATH` is `/home/user/caddy`.
caddy repository is at `/home/user/caddy/src/github.com/mholt/caddy`
This issue happens on Linux and Windows too.

### 4. Checklist


tobya:
I think this should be fine, we just need you to sign the Contributor agreement. You seem to have submitted with a differnt email address than you github account has you need to add it to your github account.  See link in notice above.

PR #2363, opened 2018-11-22
Implement `{when_iso_local}` placeholder

### LINKED ISSUE #2362: Add `{when_iso_local}` placeholder

Git commit 8f7a1d6a2 from Nov 18, 2018.

I prefer the ISO 8601 timestamp format but when perusing the log file it is friendlier if the human readable timestamps are in the local timezone of the server rather than UTC. So I'd like to propose adding a `{when_iso_local}` placeholder. The implementation is trivial:

```
diff --git a/caddyhttp/httpserver/replacer.go b/caddyhttp/httpserver/replacer.go
index 945b93e..8755f40 100644
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -368,6 +368,8 @@ func (r *replacer) getSubstitution(key string) string {
                return url.QueryEscape(r.request.URL.RequestURI())
        case "{when}":
                return now().Format(timeFormat)
+       case "{when_iso_local}":
+               return now().Format(timeFormatISO)
        case "{when_iso}":
                return now().UTC().Format(timeFormatISOUTC)
        case "{when_unix}":
@@ -539,6 +541,7 @@ func (r *replacer) Set(key, value string) {

 const (
        timeFormat        = "02/Jan/2006:15:04:05 -0700"
+       timeFormatISO     = "2006-01-02T15:04:05" // ISO 8601 with timezone to be assumed as local
        timeFormatISOUTC  = "2006-01-02T15:04:05Z" // ISO 8601 with timezone to be assumed as UTC
        headerContentType = "Content-Type"
        contentTypeJSON   = "application/json"
```

For automated log file analysis I also include the `{when_unix_ms}` placeholder. This gives analysis tools an unambiguous timestamp while giving humans an easier to interpret string via the `{when_iso_local}` placeholder. Yes, this is itself problematic if you have personnel in different time zones who need to peruse the logs. But in that case you can either use the UTC timestamps and/or a tool to convert the UTC timestamps to each person's local timezone. This enhancement is primarily for people who are using Caddy as a personal web server.

krader1961:
This implements the `{when_iso_local}` placeholder. This is like the
`{when_iso}` placeholder but the output is in the current timezone
rather than UTC.

Resolves #2362


krader1961:
The comments I added about the use of `FixedTimezone()` is because I was surprised the value was in seconds versus hours. So my first attempt at adding a test case failed. Yes, an experienced Go programmer would know that but it seems to me to be sufficiently obscure to warrant a comment.

krader1961:
Why is the build failing with the errors below since my change does not change any imports and `go fmt` on my system doesn't modify my change?

```
gometalinter --disable-all -E vet -E gofmt -E misspell -E ineffassign -E goimports -E deadcode --tests --vendor ./...
caddyhttp\httpserver\replacer.go:1::warning: file is not gofmted with -s (gofmt)
caddyhttp\httpserver\replacer.go:1::warning: file is not goimported (goimports)
```

tobya:
> Also, I would have updated the documentation at https://caddyserver.com/docs/placeholders but it wasn't obvious how to do so.

The docs repo is looked after by @mholt. If you put here the changes that need to be added to docs Matt will add them to the repo.

krader1961:
> If you put here the changes that need to be added to docs Matt will add them to the repo.

Okay. In this case it's pretty simple. Modify https://caddyserver.com/docs/placeholders by adding 

```
{when_iso_local} | Timestamp in the format 2006-01-02T15:04:05 in local TZ
```

tobya:
If you run 
   
    gofmt -s -w replacer.go 

on the two files and submit any changes that should do it.

krader1961:
> gofmt -s -w replacer.go

Done

caddy/main.go
```diff
@@ -20,6 +20,8 @@
 package main
 
 import "github.com/mholt/caddy/caddy/caddymain"
+import _ "github.com/krader1961/caddy-geoip"
+import _ "github.com/krader1961/ipfilter"
```
francislavoie: This seems wrong

krader1961: /facepalm /embarrassment Yes, my custom config commit should not have been included. I've force pushed the proper change.

tobya:
@francislavoie Are you happy to approve?

tobya:
Thank you  @krader1961 .  Sorry took so long to merge.

Issue #2356, opened 2018-11-19
Unexpected TLS alert for IP-only certificate


jung-kurt:
### 1. What version of Caddy are you using (caddy -version)?

Caddy (untracked dev build) (unofficial) 

commit f6e50890b3d81cb04146bd7b2c2b59e99830849a

### 2. What are you trying to do?

Use TLS with IP addresses (not DNS) for local development. Everything has worked fine up to and including commit 22dfb140d0b3587dc9de4f2f15f93a9f319becf2 right before the TLS alert commit. I am not sure if this is a problem with (1) Caddy correctly handling invalid certificates or (2) Caddy incorrectly handling valid certificates. Nothing jumped out at me when examining the commit diffs so I will not be surprised if the fault is with the certificates. Any insights are appreciated.

### 3. What is your entire Caddyfile?
```
https://192.168.1.2 {
        tls cert/pki/issued/192.168.1.2.bundle.crt cert/pki/private/192.168.1.2.key
}
```
More details on the certificates can be found below.

### 4. How did you run Caddy (give the full command and describe the execution environment)?

`caddy -conf ./Caddyfile`

### 5. Please paste any relevant HTTP request(s) here.

`curl --cacert cert/pki/ca.crt https://192.168.1.2`

### 6. What did you expect to see?

`<h1>Hello from Caddy</h1>`

This is what I get with versions of Caddy before the TLS alert commit.

### 7. What did you see instead (give full error messages and/or log)?

`curl: (35) error:14094438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error`

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

Here are the steps in the form of a shell script. Sorry for the involved sequence, but nothing is easy when it comes to TLS certificates.
```
# --- Obtain, install and initialize easyrsa ---

wget https://github.com/OpenVPN/easy-rsa/releases/download/v3.0.5/EasyRSA-nix-3.0.5.tgz
mkdir -p ~/tmp/caddy
tar -xzf EasyRSA-nix-3.0.5.tgz -C ~/tmp/caddy
mv ~/tmp/caddy/EasyRSA-3.0.5 ~/tmp/caddy/cert
cd ~/tmp/caddy/cert
cat > vars  <<end-of-vars
set_var EASYRSA_PKI		"$PWD/pki"
set_var EASYRSA_DN	"cn_only"
set_var EASYRSA_KEY_SIZE	2048
[...]

PR #2367, opened 2018-11-24
Partially fix raw IP request regression (#2356)

Zenexer:

Partially resolves regression that resulted from a security patch, as described in #2356

#2356

This still needs to be discussed.

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -391,6 +391,12 @@ func groupSiteConfigsByListenAddr(configs []*SiteConfig) (map[string][]*SiteConf
 // parts of an address. The component parts may be
 // updated to the correct values as setup proceeds,
 // but the original value should never be changed.
+//
+// Important: In order for the fix for #2356 to work,
+// if Host is an IP address, it must be in a
+// normalized form (as presented by net.IP.String()).
+// For SNI matching in general, it needs to be
+// lowercase.
 type Address struct {
 	Original, Scheme, Host, Port, Path string
 }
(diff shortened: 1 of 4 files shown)


Zenexer:
I did my best to minimize the potential for unintended side effects.  Initially, I tried to get it to intelligently fall back to the most likely intended certificate under certain circumstances, but the changes were too disruptive, and I couldn't easily gauge the fallout.

jung-kurt:
Nice work, @Zenexer. Thanks for your prompt attention to this issue. Your solution addresses some subtleties that I would have certainly missed. My simple example and my more complicated working setup both work properly with your changes.

lukenowak:
Thank you @Zenexer , I checked your PR code and it works perfectly on my quite complex configuration (a lot of sites with many name-based certificates and few endpoints with IP based -- IPv4, IPv6 and others).

For the ip access I have caddyfile like:
```
https:[::1]:5555, https:127.0.0.1:5555 {
  bind 127.0.0.1
  tls ip.crt ip.key
}
```

And my cert has two IPs in altSubjectName with meaningless common name. All works well.

Of course! If at least one entry for IP-based access has wrong certificate, like adding to my Caddyfile:

```
https:[::1]:5555/cous, https:127.0.0.1:5555/cous {
  bind 127.0.0.1
  tls other.crt other.key
}
```

all endpoints returns error.

This is expected, as it is said in the docs, that all configurations shall share exactly the same TLS stanzas. Improving error reporting in this case is subject to another thing I think.

mholt:
Mostly superseded by https://github.com/mholt/caddy/commit/a7aeb979be5a59ae82fac0783fe54b83c7fc57fd - but I would like to at least merge in your tests, @Zenexer, either here or CertMagic, wherever they belong.

PR #2368, opened 2018-11-24
browse: filter on document load

simon04:
This applies the filter when using the browser history to go the
previous directory.


tobya:
@simon04 Thank you very much for the PR.  I have just recently been wishing that my browse directories would keep the filter.  Welcome to Caddy.

PR #2374, opened 2018-12-02
Expose JSON <-> Caddyfile conversion via command line flags

linquize:
There is already `caddyfile/json.go` but it was only used in unit test.
Now expose JSON <-> Caddyfile conversion via command line flags.

https://caddyserver.com/docs/cli

### 4. Checklist


mholt:
Interesting... do you find this conversion useful? What are you using it for?

linquize:
For scripting,  to edit Caddyfile using JSON tools, such as using jq command line.
`cat Caddyfile | caddy --jsonto | jq ....... | caddy --jsonfrom > Caddyfile`

j8r:
If we can convert from/to json, why Caddy won't also accept directly json configuration?
- H2O configuration is in YAML
- Envoy configuration is in either YAML or JSON

francislavoie:
Because we don't want to encourage people to use json by default. All the documentation is written using Caddyfile syntax and it provides things like comments as well.

francislavoie:
I think this PR should help with that though.

j8r:
@francislavoie please don't merge `master` in a branch that will be merged in `master` :skull_and_crossbones:
=> `git rebase -i master`

francislavoie:
Fair enough, I just clicked the "Resolve conflicts" button in Github. I think this will be squash merged anyways, so I doubt it'll matter (correct me if I'm wrong).

j8r:
Good to know. Yes hopefully squash integrates commits into one.

caddy/caddymain/run.go
```diff
@@ -55,6 +56,8 @@ func init() {
 	flag.BoolVar(&plugins, "plugins", false, "List installed plugins")
 	flag.StringVar(&certmagic.Email, "email", "", "Default ACME CA account email address")
 	flag.DurationVar(&certmagic.HTTPTimeout, "catimeout", certmagic.HTTPTimeout, "Default ACME CA HTTP timeout")
+	flag.BoolVar(&jsonFrom, "jsonfrom", false, "From JSON stdin to Caddyfile stdout")
+	flag.BoolVar(&jsonTo, "jsonto", false, "From Caddyfile stdin to JSON stdout")
```
mholt: What if we call the flag `from-json` and `to-json` ... makes much more sense in English.

caddy/caddymain/run.go
```diff
@@ -127,6 +130,32 @@ func Run() {
 		fmt.Println(caddy.DescribePlugins())
 		os.Exit(0)
 	}
+	if jsonFrom {
+		jsonBytes, err1 := ioutil.ReadAll(os.Stdin)
+		if err1 != nil {
+			os.Exit(1)
```
mholt: Do we want to print the error?

caddy/caddymain/run.go
```diff
@@ -127,6 +130,32 @@ func Run() {
...
+		caddyfileBytes, err1 := ioutil.ReadAll(os.Stdin)
+		if err1 != nil {
+			os.Exit(1)
+		}
+		jsonBytes, err2 := caddyfile.ToJSON(caddyfileBytes)
+		if err2 != nil {
+			fmt.Fprintf(os.Stderr, "Convert to JSON failed: %v", err2)
+			os.Exit(2)
+		}
+		fmt.Print(string(jsonBytes))
```
mholt: Would `Println` be desirable here?

j8r:
Any update?

francislavoie:
@mholt your review notes were addressed, anything else missing from this PR?

caddy/caddymain/run.go
```diff
@@ -127,6 +130,34 @@ func Run() {
...
+			os.Exit(1)
+		}
+		jsonBytes, err2 := caddyfile.ToJSON(caddyfileBytes)
+		if err2 != nil {
+			fmt.Fprintf(os.Stderr, "Convert to JSON failed: %v", err2)
+			os.Exit(2)
+		}
+		fmt.Println(string(jsonBytes))
+		os.Exit(0)
+	}
```
mholt: This function is getting long and complex. We should move the handling of fromJSON and toJSON into a separate function.

tobya:
I think it would be good if this could accept a file from disk as well as reading from stdin.  @linquize are you still interested in this ?  I am happy to add to your PR to see if we can get it merged.

mholt:
Do people need it to read from or write to disk? If it's speculative, let's wait and see if it's needed first.

tobya:
This appears to convert a caddyfile into json with all directives written back in alphabetical order.  Is this an issue.  Should the `caddyfile.ToJSON` code retain directive order?  

This is more a question about existing `cadddyfile.ToJSON` and `caddyfile.FromJSON` functions.

tobya:
Since this `cli` exits when called and only does a conversion.  For example, you cant actually use this to pass a JSON Caddyfile to caddy to start, I would like to suggest we change the name of the cli to something like this - though I appreciate it is rather long.

(2 options for from)

    -convert-caddyfile-from-json
    -convert-json-to-caddyfile

    -convert-caddyfile-to-json

or alternatively allow a param

    -convert-caddyfile=tojson
    -convert-caddyfile=fromjson
 or

    -convert=caddyfiletoJSON
    -convert=JSONtoCaddyfile

We may in the future require other convert options, this also makes it a little more obvious that a simple conversion is being performed rather than getting our actual caddyfile for the run from a JSON stream.

j8r:
It's too long @tobya, what about `--export` and `--import`? The fact it's a json can be in the command description. Or, having a `--format=json` option.

tobya:
I think I prefer `-convert` to `-export` so maybe

    -convert=tojson
    -convert=fromjson

tobya:
Is the concensus to keep `-json-to` and `-json-from` ?

I dont need to load from file but its a bit more of a windows way to do things.  Now that we have [wsl](https://docs.microsoft.com/en-us/windows/wsl/about) (Windows Subsystem for Linux) it makes it easier to pipe etc.  No problem to not have it.  

I do want to make sure we are all happy with naming though as it will be difficult to change once we have merged.

tobya:
Anyone for `-json-in` and `-json-out`

mholt:
Actually, `-from-json` and `-to-json` reads a little more naturally to me. (Is it obvious that the Caddyfile is what is being converted?)

j8r:
I think the main use case, or at least for me, is:
- having a way to export the Caddyfile in the filesystem into a json in stdout
- do JSON manipulations
- then import it, either from file or stdin, to overwrite the Caddyfile on filesystem
- reload the service

whitestrake:
Just gonna throw `-json-to-caddyfile` and `-caddyfile-to-json` in the ring.

tobya:
I have changed to

    -caddyfile-to-json
    -json-to-caddyfile

If I get one more LGTM I will merge.  Thank you to @linquize for PR and early work.

PR #2389, opened 2018-12-14
optimized parse cli's disabledMetrics flag string to initTelemetry

alimy:
run caddy:
```
$ ./caddy --disabled-metrics "os,,, arch, cpu, caddy_version,"
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
github.com/mholt/caddy/caddy/caddymain.initTelemetry(0x0, 0x0)
	/go/src/github.com/mholt/caddy/caddy/caddymain/run.go:407 +0x8d4
github.com/mholt/caddy/caddy/caddymain.Run()
	/go/src/github.com/mholt/caddy/caddy/caddymain/run.go:102 +0xb99
main.main()
	/go/src/github.com/mholt/caddy/caddy/main.go:27 +0x27
```
This patch fixed above problem.
### 4. Checklist

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -392,29 +392,31 @@ func initTelemetry() error {
 	}
 
 	// parse and check the list of disabled metrics
-	var disabledMetricsSlice []string
+	var fixedDisabledMetrics []string
 	if len(disabledMetrics) > 0 {
 		if len(disabledMetrics) > 1024 {
 			// mitigate disk space exhaustion at the collection endpoint
 			return fmt.Errorf("too many metrics to disable")
 		}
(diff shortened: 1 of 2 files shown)


caddy/caddymain/run.go
```diff
@@ -392,35 +392,53 @@ func initTelemetry() error {
 	}
 
 	// parse and check the list of disabled metrics
-	var disabledMetricsSlice []string
+	var fixedDisabledMetrics []string
```
francislavoie: I'm not sure this variable name needs to be changed, the previous name made more sense I think.

alimy: Yes, I had restore previous name.

PR #2390, opened 2018-12-14
Add tls_client_provided_cert placeholder

virtualdxs:
Adds a placeholder that resolves to true if the client offered a cert to authenticate themselves, and false otherwise

None

https://caddyserver.com/docs/placeholders needs to list the new placeholder

### 4. Checklist

### DIFF
--- a/caddyhttp/httpserver/replacer.go
+++ b/caddyhttp/httpserver/replacer.go
@@ -446,6 +446,12 @@ func (r *replacer) getSubstitution(key string) string {
 			}
 		}
 		return r.emptyValue
+	case "{tls_client_provided_cert}":
+		if r.getPeerCert() != nil {
+			return "true"
+		} else {
+			return "false"
+		}
 	case "{tls_client_escaped_cert}":
 		cert := r.getPeerCert()
 		if cert != nil {


tobya:
Thank you for PR.  This should be fine.  

Can you please add a test to the PR.

virtualdxs:
Sorry for taking so long to get back to this. I have no idea how to write tests, and I wasn't able to understand the tests surrounding the other placeholders, so what should I do? I'm just not a programmer.

mholt:
Thanks for the contribution! Nice job getting this far. What do you use this for? Most people would want to know something about the certificate, I'd imagine, rather than just whether there is one.

virtualdxs:
Our application needs to accept both TLS authenticated and IP whitelisted clients, so the config I wrote up basically says "If provided, validate and reject if failed; if not provided, pass through", and passes through that flag. So basically, we check to see if that flag is set, and it indicates to us that (due to the config) the client successfully authenticated with TLS and can be trusted.

virtualdxs:
Another, likely more useful idea, is a `tls_client_provided_valid_cert` placeholder. It would do the same thing for me, since I have caddy set to reject invalid certs.

mholt:
Oh, yes, I like that idea better. How about we call it `tls_valid_client_cert` instead.

virtualdxs:
I almost have it, but I'm stuck on one thing: How can I access the `caddytls.Config` object for the current connection from the context of a `caddyhttp.httpserver.replacer` object?

mholt:
Sorry for the delay on this. I think at this point, the efforts spent on this feature would be more relevant with Caddy 2, which has different mechanisms for placeholders and more powerful configurability. Just so you know, it will require a degree of skill though which is probably suitable for programmers with moderate Go / TLS experience, since we're still building the foundations of all these new features in Caddy 2.

Thank you for your contributions! I hope you will contribute again, to Caddy 2.

PR #2398, opened 2018-12-19
Terminate goroutine used for logging errors

yuewko:
Terminate the goroutine used for logging errors by using a
WaitGroup (stopWg) to track termination of servers.

This PR addresses https://github.com/mholt/caddy/issues/2358. It now terminates the goroutine used for logging errors by using a WaitGroup (stopWg) to track termination of servers.

https://github.com/mholt/caddy/issues/2358

n/a

### 4. Checklist

I haven't written any test, but did verify that goroutine count no longer increases (using the same test method described in the issue).


francislavoie:
@mholt anything actionable here?

caddy.go
```diff
@@ -701,6 +701,11 @@ func executeDirectives(inst *Instance, filename string,
 func startServers(serverList []Server, inst *Instance, restartFds map[string]restartTriple) error {
 	errChan := make(chan error, len(serverList))
 
+	// used for signaling to error logging goroutine to terminate
+	stopChan := make(chan struct{})
+	// used to track termination of servers
+	stopWg := &sync.WaitGroup{}
```
mholt: Is this different from `inst.wg`?  It looks like both do the same thing: wait for the instance to shut down.

yuewko: The `Wait()` for `inst.wg` happens at the end of `Run()` in `caddymain/run.go` - i.e. only when the instance shutdown.

What I'm trying to do here is to kill the error logging loop every time another `startServers()` is called (as a result of calling `Restart()`). That's why I need to use `stopWg` instead of piggy-backing on `inst.wg`.

Hope I'm explaining myself. And thanks for reviewing the PR.

Issue #2080, opened 2018-03-23
TLS 1.3


rugk:
I could not find an open issue, and I hope it is not already implemented, but TLS 1.3 support should be added!

TLS 1.3 is finalized AFAIK, browsers are soon going to implement the latest "draft" (not a draft now anymore, don't know the name).

mholt:
Waiting for Go to implement it into their standard library.

Fortunately, @FiloSottile -- who [already implemented an earlier draft of TLS 1.3 in Go](https://github.com/cloudflare/tls-tris) -- now works on the Go team doing cryptography. Sooooo I'm pretty confident it could happen by Go 1.11 or definitely Go 1.12. Right? :wink:

faradaytrs:
They seem to be aiming for go 1.12

mholt:
Yep. We're looking at ~Feb. 2019 I think.

rugk:
Uuh, so late??

mholt:
Stability over speed, when it comes to things like this...

Henrocker:
I've compiled Caddy together with the `tls-tris` lib from cloudflare and released a how to here: 

https://www.henrock.net/md/2._caddy+TLS1.3.md.

TLS1.3 version is RFC8446.

PoC is https://www.henrock.net/

vladbondarenko:
> I've compiled Caddy together with the `tls-tris` lib from cloudflare and released a how to here:
> 
> https://www.henrock.net/md/2._caddy+TLS1.3.md.
> 
> Patch of caddy (pretty-printed) is [here](https://www.henrock.net/md/caddy.patch.html).
> 
> TLS1.3 version is RFC8446.
> 
> PoC is https://www.henrock.net/

Hi, Matt, sorry for chatting in caddy repo, but @Henrocker seems there is no https://golang.org/src/crypto/internal/ now.
GOARCH=amd64 GOROOT="/root/tls-tris/_dev/GOROOT/linux_amd64" go install -v std
_dev/GOROOT/linux_amd64/src/crypto/tls/common.go:10:2: cannot find package "crypto/internal/cipherhw" in any of:
        /root/tls-tris/_dev/GOROOT/linux_amd64/src/vendor/crypto/internal/cipherhw (vendor tree)
        /root/tls-tris/_dev/GOROOT/linux_amd64/src/crypto/internal/cipherhw (from $GOROOT)
        /root/go/src/crypto/internal/cipherhw (from $GOPATH)

[...]

PR #2399, opened 2018-12-20
caddytls: add TLS 1.3 support

crvv:
add TLS 1.3

#2080 

https://caddyserver.com/docs/tls
1. "tls1.3" is a new suppoted protocol if caddy is built by Go 1.12 or newer.
2. TLS 1.3 cipher suites are not configurable.

### 4. Checklist

If caddy should be able to be built by Go 1.11 after the releasing of 1.12, I think this PR can be accepted now.
If not, the new file with build tag "go1.12" is useless and the change can be applied to config.go directly when Go 1.12 is released.

### DIFF
--- a/caddytls/setup.go
+++ b/caddytls/setup.go
@@ -34,6 +34,10 @@ import (
 )
 
 func init() {
+	// opt-in TLS 1.3 for Go1.12
+	// TODO: remove this line when Go1.13 is released.
+	os.Setenv("GODEBUG", os.Getenv("GODEBUG")+",tls13=1")
+
 	caddy.RegisterPlugin("tls", caddy.Plugin{Action: setupTLS})
 
 	// ensure the default Storage implementation is plugged in
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -407,7 +407,7 @@ func SetDefaultTLSParams(config *Config) {
 		config.ProtocolMinVersion = tls.VersionTLS12
 	}
 	if config.ProtocolMaxVersion == 0 {
-		config.ProtocolMaxVersion = tls.VersionTLS12
+		config.ProtocolMaxVersion = tls.VersionTLS13
 	}
 
 	// Prefer server cipher suites
@@ -430,6 +430,7 @@ var SupportedProtocols = map[string]uint16{
 	"tls1.0": tls.VersionTLS10,
 	"tls1.1": tls.VersionTLS11,
 	"tls1.2": tls.VersionTLS12,
+	"tls1.3": tls.VersionTLS13,
 }
 
 // GetSupportedProtocolName returns the protocol name
@@ -489,10 +490,6 @@ var defaultCiphers = []uint16{
 	tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
 	tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
 	tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
-	tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
-	tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
-	tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
-	tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
 }
 
 // List of ciphers we should prefer if native AESNI support is missing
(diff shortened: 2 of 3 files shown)


mholt:
Nice, thanks for jumping on this. (Realizing that, of course, this is a temporary solution, as once Go 1.12 is released, we can just update it in-place.)

While we're at it, can we remove the CBC ciphers from the defaultCiphers lists? I think we should take this opportunity to modernize Caddy's TLS config to be up to snuff with the latest: TLS 1.2 and 1.3.

crvv:
I tested this on SSL Labs. The result is https://www.ssllabs.com/ssltest/analyze.html?d=tw.crvv.me&hideResults=on

With Let's Encrypt RSA certificates, the failed clients are `IE 11 on Windows 7/8` and `Safari 6/7/8`

Is that OK?

abiosoft:
I do not think this should be an issue.

> I tested this on SSL Labs. The result is https://www.ssllabs.com/ssltest/analyze.html?d=tw.crvv.me&hideResults=on
>
> With Let's Encrypt RSA certificates, the failed clients are IE 11 on Windows 7/8 and Safari 6/7/8
>
> Is that OK?

elcore:
> While we're at it, can we remove the CBC ciphers from the defaultCiphers lists? I think we should take this opportunity to modernize Caddy's TLS config to be up to snuff with the latest: TLS 1.2 and 1.3.

We definitely should add TLS 1.3 cipher suites 😄

crvv:
@elcore 
TLS 1.3 ciphers are not configurable via `Config.CipherSuites`.
Please see golang/go#29349 for details.

elcore:
> @elcore
TLS 1.3 ciphers are not configurable via Config.CipherSuites.
Please see golang/go#29349 for details.

Oh... Thank you!

brink668:
Is there an ETA on this?

mholt:
To use Caddy with TLS 1.3 today, go ahead and build from @crvv's branch -- and when Go 1.12 comes out next month, this PR should be updated to make TLS 1.3 the default (without the build constraint which it currently has), and then we'll merge in the PR.

mholt:
@crvv Will we need to set an env variable to enable TLS 1.3 by default? https://github.com/golang/go/issues/30055

crvv:
We need to set that.
But let's wait for some responses to my comments in that issue.

mholt:
@crvv Great! Looks like a doc change will be merged, and I'm good with it if we include that Setenv in this PR to enable it by default.

Henrocker:
Made some tests today, compiling caddy with go1.12rc1 and [this](https://hnrk.io/md/caddy.patch.html) patch. 
Result: TLS 1.3 did not work anymore.

Switched then back to go1.12beta2 and applied the same patch and boom, TLS 1.3 works again... Wonder where that comes from.

mholt:
@Henrocker I think it's because rc1 now requires an environment variable to be set. See the issue I linked to in a comment above. Would love it if you could help figure out if that's what is needed!

Henrocker:
Thanks for pointing me out. Gonna do some further tests tomorrow and will report back if the envvar fixes this. Thanks

Henrocker:
Applied [patch](https://hnrk.io/md/caddy.patch) and Caddy with go1.12rc1 Build command:
```
$ GODEBUG=tls13=1 go run build.go
```

Command to run Caddy:
```
$ GODEBUG=tls13=1 ./caddy
```

TLS 1.3 works! :-)

mholt:
@Henrocker Any chance we can set that env var in code instead of the user having to do it?

Henrocker:
Maybe something like https://github.com/golang/go/issues/30055#issuecomment-463146031 could be done, but since my understanding in golang is limited, I'm afraid I can only do some tests, if there's something new to try out.

caddytls/setup.go
```diff
@@ -34,6 +34,9 @@ import (
 )
 
 func init() {
+	// opt-in TLS 1.3 for Go1.12
+	os.Setenv("GODEBUG", os.Getenv("GODEBUG")+",tls13=1")
```
crvv: @Henrocker  @mholt 
I have added this env in code.

whalehub: @mholt I can confirm that compiling Caddy with go1.12rc1 and @crvv's change successfully enables TLS 1.3 support in Caddy without the need to manually set an env variable in the OS.

mholt: Excellent!

Can we put a TODO in the comment so we remember to revisit this later (i.e. after Go 1.13)?

crvv: Yes, I have added the TODO.

mholt:
Go 1.12 is released; just waiting for CI systems to pick up the new version.

mholt:
AppVeyor is refusing to pick up the merge that upgrades to Go 1.12, but the tests on Travis are passing, so I'm merging this sucker.

Thanks everyone!! How exciting.

kleveralt:
tls1.0 and tls1.2 is no longer available?

I use this config but only tls1.2 and tls1.3 are supported, since this commit...
protocols tls1.0 tls1.3

whitestrake:
The documentation does state that "Supported protocols and default protocol versions may be changed at any time", but if that's the case, it should be updated at the next opportunity to reflect the current available protocols.

mholt:
TLS 1.0 and 1.1 are supported but not enabled by default, you have to enable those manually, but I highly, highly, strongly, super duper discourage that.

kleveralt:
I understand, but ssllabs.com replys only tls1.2 and 1.3 are supported with this cofig,
    tls {
        protocols tls1.0 tls1.3
    }

image

mholt:
Oh, I see what you mean. Hmm. Can you open a new issue please?

crvv:
With this PR, there are only AEAD ciphers in the default cipher suites.
AEAD is new in TLS 1.2
So TLS 1.0 and 1.1 won't work with default cipher suites.

If you want TLS 1.0 and 1.1, you must specify cipher suites you want in the Caddyfile.

mholt:
Annnd we have a winner, totally forgot about that. 😅  That would probably be the reason! Thanks @crvv.

Issue #2346, opened 2018-11-14
Occasional sigsegv with QUIC in 0.11.1


vcsjones:
### 1. What version of Caddy are you using (`caddy -version`)?

0.11.1.

### 2. What are you trying to do?

Run caddy with `-quic`.

### 3. What is your entire Caddyfile?

Omitted for now, but can try to make a sample that reproduces the problem if relevant.

### 4. How did you run Caddy (give the full command and describe the execution environment)?

```
/usr/local/bin/caddy -log stdout -quic -agree=true -conf=/opt/caddy/Caddyfile/Caddyfile -root=/var/tmp
```

`lsb_release -a`:

```
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
Codename:	xenial
```

Caddy starts up fine under this scenario, but Caddy appears to occasionally crash with the backtrace below in the `journalctl` log. I can't reliably reproduce it, and I'm not sure if the issue is upstream in `quic-go` or if it is Caddy's use of `quic-go` (I do not see an issue in quic-go that looks like this backtrace).

I did not have a single crash with months of uptime in Caddy with 0.11.0, but observed this in 0.11.1 a few hours afters updating.

Removing `-quic` from the startup appears to have resolved the issue.

```
Nov 14 05:57:51 ip-10-0-0-24 caddy[1296]: panic: runtime error: invalid memory address or nil pointer dereference
Nov 14 05:57:51 ip-10-0-0-24 caddy[1296]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x8b734c]
Nov 14 05:57:51 ip-10-0-0-24 caddy[1296]: goroutine 3829 [running]:
Nov 14 05:57:51 ip-10-0-0-24 caddy[1296]: github.com/mholt/caddy/vendor/github.com/lucas-clemente/quic-go/internal/crypto.(*certChain).getCertForSNI(0xc00000c5c0, 0xc0000315c0, 0x10, 0xc000400ea0, 0xfde33730, 0x28f7167692445711)
Nov 14 05:57:51 ip-10-0-0-24 caddy[1296]: #011/tmp/gopath_11-14-0232.202407298/src/github.com/mholt/caddy/vendor/github.com/lucas-clemente/quic-go/internal/crypto/cert_chain.go:66 +0x6c
[...]

PR #2404, opened 2018-12-21
set a GetCertificate callback in the tls.Config

marten-seemann:

It sets a fake `tls.Config.GetCertificate` callback in the `tls.Config`. According to the Go standard library, a `tls.Config` must have `Certificates` or `GetCertificate` set, otherwise it will be rejected by `tls.Listen`. quic-go recently implemented the same check, and rejects the `tls.Config` as currently constructed by Caddy.
I think it should be permissible to have a `tls.Config` that only sets the `GetConfigForClient` callback, and I opened https://github.com/golang/go/issues/29139, but it looks like this was postponed to Go 1.13.

#2346, especially https://github.com/mholt/caddy/issues/2346#issuecomment-445248107

### 4. Checklist

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -269,6 +269,13 @@ func MakeTLSConfig(configs []*Config) (*tls.Config, error) {
 	}
 
 	return &tls.Config{
+		// A tls.Config must have Certificates or GetCertificate
+		// set, in order to be accepted by tls.Listen and quic.Listen.
+		// TODO: remove this once the standard library allows a tls.Config with
+		// only GetConfigForClient set.
+		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return nil, fmt.Errorf("all certificates configured via GetConfigForClient")
+		},
 		GetConfigForClient: configMap.GetConfigForClient,
 	}, nil
 }


caddytls/config.go
```diff
@@ -269,6 +270,11 @@ func MakeTLSConfig(configs []*Config) (*tls.Config, error) {
 	}
 
 	return &tls.Config{
+		// A tls.Config must have Certificates or GetCertificate
+		// set, in order to be accepted by tls.Listen and quic.Listen.
+		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
```
mholt: Can you add a TODO here so that we remember to remove this after it is safe to do so?

francislavoie:
@marten-seemann just wondering, what's the status on this? Could you address @mholt's feedback?

marten-seemann:
> > quic-go recently implemented the same check
>
> I do wonder, why did you implement the same check, if 1) the standard library already does it, and 2) you disagree with it?

In the development version of QUIC (which has preliminary IETF QUIC support), we're using tls-tris, which is based on standard library code. The change I'm proposing requires changes to way the standard library retrieves the applicable `tls.Config`. While technically not necessary, since the gQUIC branch of quic-go we're using in Caddy doesn't use tris, I prefer to be consistent.

I updated the PR according to your review.

mholt:
Great, thanks for the explanation! Merging now. Thanks again for the patch.

mholt:
[Go 1.14 will supersede this workaround](https://tip.golang.org/doc/go1.14#crypto/tls), yay!

> Config.Certificates and Config.GetCertificate can now both be nil if Config.GetConfigForClient is set. If the callbacks return neither certificates nor an error, the unrecognized_name is now sent.

PR #2415, opened 2019-01-03
Disable basic authentication for OPTIONS method

sebastianhutter:
The pull request disables basic auth on OPTION method calls.

https://github.com/mholt/caddy/issues/1568

No. The pull request is minimal and changes the behaviour of the basic auth plugin without adding new configuration options

### 4. Checklist

### DIFF
--- a/caddyhttp/basicauth/basicauth.go
+++ b/caddyhttp/basicauth/basicauth.go
@@ -52,37 +52,39 @@ func (a BasicAuth) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 	var protected, isAuthenticated bool
 	var realm string
 
-	for _, rule := range a.Rules {
-		for _, res := range rule.Resources {
-			if !httpserver.Path(r.URL.Path).Matches(res) {
-				continue
+	if r.Method != http.MethodOptions {
+		for _, rule := range a.Rules {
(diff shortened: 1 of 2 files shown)


caddyhttp/basicauth/basicauth.go
```diff
@@ -52,37 +52,39 @@ func (a BasicAuth) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 	var protected, isAuthenticated bool
 	var realm string
 
-	for _, rule := range a.Rules {
-		for _, res := range rule.Resources {
-			if !httpserver.Path(r.URL.Path).Matches(res) {
-				continue
+	if r.Method != http.MethodOptions {
```
tobya: Could we return early if this is the case
```
// Do not check basicauth for OPTIONS
if r.Method == http.MethodOptions {
// Pass-through when no paths match
	return a.Next.ServeHTTP(w, r)
}

for _, rule := range a.Rules {
//etc
```
I think it works better for usability

tobya:
Thank you for the PR its great to have this looked at.

caddyhttp/basicauth/basicauth.go
sebastianhutter: good point 👍 
i changed the check accordingly

Issue #2378, opened 2018-12-03
Allow using import within proxy directive


euh2:
### 1. What version of Caddy are you using (`caddy -version`)?
Caddy 0.11.1 (+6f580c6 Sun Dec 02 00:15:50 UTC 2018) (unofficial)

### 2. What are you trying to do?
I'm trying to import a configuration within a `proxy directive`

### 3. What is your entire Caddyfile?
The relevant part of the Caddyfile
```text
example.com {
  import /etc/caddy/Caddyfile.d/log
  import /etc/caddy/Caddyfile.d/ipfilter
  import /etc/caddy/Caddyfile.d/realip
  proxy / 172.20.0.4:80/ {
    import /etc/caddy/Caddyfile.d/security_header_proxy ## <-- This Line causes the Error
    transparent
  }
```
The snippet I want to import with `import /etc/caddy/Caddyfile.d/security_header_proxy`
```text
header_downstream Strict-Transport-Security "max-age=31536000;"
header_downstream X-XSS-Protection "1; mode=block"
header_downstream X-Content-Type-Options "nosniff"
header_downstream X-Frame-Options "DENY"
header_downstream x-powered-by ""
header_downstream Server ""
header_downstream Referrer-Policy "no-referrer"
```
### 4. How did you run Caddy (give the full command and describe the execution environment)?
I'm not sure if this is relevant, but I run Caddy with the Caddy File Loader for Docker (https://github.com/lucaslorentz/caddy-docker-proxy)

### 5. Please paste any relevant HTTP request(s) here.
_I add this if relevant_

### 6. What did you expect to see?
_I add this if relevant_

### 7. What did you see instead (give full error messages and/or log)?
``[ERROR] CaddyFile error: :60 - Error during parsing: unknown property 'import'``

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
_I add this if relevant_

francislavoie:
Unfortunately, I don't think `import` is implemented to work inside of directives currently.

I think some logic to call `doImport` would need to be added around here: https://github.com/mholt/caddy/blob/master/caddyfile/parse.go#L369

A PR (with tests!) is welcome if you're willing to implement it!

PR #2428, opened 2019-01-14
add support for directive level imports

awoodbeck:
Adds support for import statements to Caddyfile directives.

#2378 

The [import](https://caddyserver.com/docs/import) docs 

### 4. Checklist

### DIFF
--- a/caddyfile/parse_test.go
+++ b/caddyfile/parse_test.go
@@ -371,6 +371,68 @@ func TestRecursiveImport(t *testing.T) {
 	}
 }
 
+func TestDirectiveImport(t *testing.T) {
+	testParseOne := func(input string) (ServerBlock, error) {
+		p := testParser(input)
+		p.Next() // parseOne doesn't call Next() to start, so we must
+		err := p.parseOne()
+		return p.block, err
+	}
+
+	isExpected := func(got ServerBlock) bool {
+		if len(got.Keys) != 1 || got.Keys[0] != "localhost" {
+			t.Errorf("got keys unexpected: expect localhost, got %v", got.Keys)
+			return false
+		}
+		if len(got.Tokens) != 2 {
+			t.Errorf("got wrong number of tokens: expect 2, got %d", len(got.Tokens))
+			return false
+		}
+		if len(got.Tokens["dir1"]) != 1 || len(got.Tokens["proxy"]) != 8 {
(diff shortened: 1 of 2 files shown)


caddyfile/parse_test.go
```diff
@@ -371,6 +371,55 @@ func TestRecursiveImport(t *testing.T) {
...
+		}
+		return true
+	}
+
+	directiveFile, err := filepath.Abs("testdata/directive_import_test")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = ioutil.WriteFile(directiveFile, []byte("prop1 1"), 0644)
```
francislavoie: Should probably also test that `prop1 1` is seen after the import. Doesn't seem like that happens. The `len != 6` check doesn't mean much because `import testdata/directive_import_test` is two tokens and `prop1 1` is two tokens as well. Am I off-base here?

awoodbeck: You're not off base at all.  I'll update it.

francislavoie:
I just realized - this could potentially be a breaking change if a plugin uses "import" as a directive option. I really doubt that's the case though, because I think it's pretty clear that import was special in Caddyfiles. This is more of a note to @mholt for when writing the changelog later.

francislavoie:
I'm not sure if it really makes sense - `import` is still not a valid directive option, so the error will probably be misleading, saying `Error during parsing: unknown property 'import'`. I think it's better to actually have it error out with a missing file message.

awoodbeck:
But erroring out while parsing the import within a directive doesn't address your concern of:
> this could potentially be a breaking change if a plugin uses "import" as a directive option.

... unless I misunderstand.  Have a look at the latest changes I pushed.  If the code cannot successfully import within another directive ("proxy" in the op's case), the line will be left as-is.  No error is returned in that case and parsing continues.

francislavoie:
My concern about it being a breaking change wasn't really directed to you, but more as a note for the changelogs. I don't think there's any good way of making this non-breaking, because we still want to properly validate bad Caddyfiles, not just make it be quiet if something went wrong in this sort of situation.

awoodbeck:
Understood.  I'll revert the latest change and address the error message.

tobya:
Thanks for the PR. Would it be possible to add a little note here about what and where this change can be used.

Issue #2400, opened 2018-12-20
Caddy produces account rate limit errors when compiled with latest certmagic package


whalehub:
### 1. What version of Caddy are you using (`caddy -version`)?

Working build: Caddy 0.11.1 (+0684cf8 Wed Dec 19 15:48:39 UTC 2018) (unofficial)
Broken build: Caddy 0.11.1 (+0b83014 Thu Dec 20 10:24:39 UTC 2018) (unofficial)

### 2. What are you trying to do?

I'm trying to run a build of Caddy that has been compiled against the [latest commit](https://github.com/mholt/caddy/commit/0b83014ff81b68b8fde21521662339396c277ab8).

### 3. What is your entire Caddyfile?

```
(headers) {

   header / {
        Strict-Transport-Security "max-age=31536000; includeSubDomains; preload"
        X-XSS-Protection "1; mode=block"
        X-Content-Type-Options "nosniff"
        X-Frame-Options "sameorigin"
        Referrer-Policy "same-origin"
        Expect-CT "max-age=86400, enforce, report-uri=\"<redacted>""
        Feature-Policy "geolocation 'none'; camera 'none'; microphone 'none'"
        Content-Security-Policy "
          default-src 'none' ;
          script-src 'self' 
            'sha256-cecXKP7KMOZMsFpYnnSCvdcCU6aS02wUdHhXQz+rzBY=' 
            'sha256-/7v/RBsCYSwEKfNm3GKddoMHIwFQ+LfJpRWy80yIag4=' ;
          style-src 'self' 
            'sha256-PNU+0KaVWZj7tiWyHGw3pEsZuO7kRmnu5/l6YGevQT8=' 
            'sha256-aqNNdDLnnrDOnTNdkJpYlAxKVJtLt9CtFLklmInuUAE=' 
            'sha256-JzqfVewwP10gJQP4wR5aRX/QBEDQqm4aNzeNDSlmFMA=' 
            'sha256-6HXkwoFqJhqd+PP4Vxf73UzS42TYUDqU+itLbeA5gkg=' 
            'sha256-xU719XNBJeO/eH/EjSQB+kSzlP6mmB8IQMtMQXFyp+o=' 
            'sha256-aIzxslm0qrBPPo+5scCOmx10sNTFpcpSd4f3MmyjjFM=' 
            'sha256-H8+aYwf/B0lGGDHV2VNReccOGfuYjzG52CWxEhAPqJs=' ;
          img-src * ;
[...]

PR #2432, opened 2019-01-17
caddytls: set certmagic.Config.Email when parsing config file

crvv:
Fix a bug. caddy doesn't set the `Email` field of certmagic.Config.Email.

I encountered #2400 when I upgraded caddy today.
And found that my email setting in Caddyfile didn't work for some time.
This PR is related to #2400 but doesn't fix it.

It looks like `~/.caddy/acme/acme-staging-v02.api.letsencrypt.org/users/default/default.json` had some other formats which caddy doesn't recognize. So that caddy can't read the account setting and caddy never updates that file. Every time caddy is about to obtain a certificate, it registers a new account.

I can't confirm this. It is just my guess. (I have deleted my old `~/.caddy`. I shouldn't have done that.)

None

### 4. Checklist


mholt:
Thanks. So, I'm not really clear on how this affects you: does this fix the behavior you're seeing, or not?

mholt:
Also, does your Caddyfile have only a single site? Or are there > 10?

crvv:
As I explained in #2400, this resolved by issue, partially.
In my Caddyfile, there are 8 domains, so the rate limit error occurred when I started caddy for the second time.

Issue #2402, opened 2018-12-20
caddy not logging basic authentication failures to log/error


calisro:
I'm attempting to use tools like fail2ban to protect basic authentication from brute force.  Caddy implements basic auth but doesn't log failures.  

I'm using Caddy 0.11.1.

tobya:
Hello @calisro 

Thank you for your issue.  Can you be as clear as possible what you are looking for.  As far as I can see caddy logs failed authorization attempts in the domain log file with a 401 code.  However it doesnt log the username that the failed attempt used.

calisro:
Hi.  Yes 401s are logged but those are unauthorized errors not necessarily authentication failures I believe.  For example, when a user comes to a site and see's the basic challenge and simply cancels the request to authorize we'll still see 401's for resources such as favicon.ico.  That wasn't an authentication failure.  What I am looking for is the basic module to actually log password error attempts.  

I believe apache and nginx both do this where they capture the actual attempts at an authorization and just just a 'unauthorized' 401.  If you look at the fail2ban config files for nginx, for example, we see this:

failregex = ^ \[error\] \d+#\d+: \*\d+ user "\S+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$

Its logging an actual attempt at passing a username/password to the nginx auth module.  

This example is a SINGLE authorization attempt on caddy.  One time.  
Current state:

```
xxxx - xxxxx:0000 - - [21/Dec/2018:08:38:22 -0500] "GET /sdf/sdf HTTP/2.0" 401 17 - /sdf/sdf
xxxx - xxxxx:0000 - - [21/Dec/2018:08:38:27 -0500] "GET /sdf/sdf HTTP/2.0" 401 17 - /sdf/sdf
xxxx - xxxxx:0000 - - [21/Dec/2018:08:38:28 -0500] "GET /favicon.ico HTTP/2.0" 401 17 - /file_not_found

```

In the above example, we get 3 401's for a single attempt.  That is just showing the resources attempted.

In this example, the user didn't even attempted to authorize.  They simple cancelled the challange:
xxxx - xxxxx:0000 - - [21/Dec/2018:08:41:58 -0500] "GET /sdf/sdf HTTP/2.0" 401 17 - /sdf/sdf
[...]

PR #2434, opened 2019-01-18
Log error when BasicAuth fails

tobya:
Raises an error which is logged when a basicauth attempt fails.

The error logged is 

 > BasicAuth: user \"%s\" was not found or password was incorrect. {remote} {host} {uri} {proto}

which comes out as 

  >  18/Jan/2019:20:07:10 +0000 [ERROR 401 /admin] BasicAuth: user "admin" was not found or password was incorrect. 127.0.0.1 salmon.kinchik.ie /admin?withinfo=alskjfd HTTP/1.1

I updated the tests that when the error is raised they check it is the correct error.

#2402 

It could be added that BasicAuth will log an error when authentication fails

### 4. Checklist


tobya:
I  noticed there are some unneeded lines etc and some tidying up, I will do that now.

mholt:
Thanks Toby! Looking forward to reviewing this when I have a chance.

Always gotta be extra careful in the basicauth middleware...

sumnerboy12:
Hi - I am testing this on v2.0 and I am not seeing any additional logging when BasicAuth fails. Looking at the code in https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/caddyauth/basicauth.go I can't the extra logging added in this PR.

Was this an intentional regression or something that was missed? I am trying to configure fail2ban on my reverse proxies but keep getting false positives due to the initial 401 sent when first trying to authenticate.

Sorry to re-open an old PR!

mholt:
We rewrote the auth modules in v2, so it's not a "regression" per-se, as we have all-new logging facilities and logic in v2. We can add more logging in v2, but please open a new issue to request it. Thanks!

sumnerboy12:
added #3363

PR #2446, opened 2019-01-28
Add extra dash

jsoref:

Change readme to use `--` instead of `-` for `disabled-metrics`

### 4. Checklist

I'm not actually sure this change is correct. Based on everything else, I suspect I'm right, if I'm wrong, please accept my apologies (I will probably send a larger minor PR for something else in February).


francislavoie:
See here: https://caddyserver.com/docs/cli

Single dash is correct.

Thanks for the attention to detail though!

PR #2447, opened 2019-01-28
Removed redeclaration of config.Manager

### LINKED ISSUE #2407: Caddy ignores -https-port and breaks custom port mapping with recent change

broken (latest master):
0b83014ff81b68b8fde21521662339396c277ab8

a working revision:
4f5df39bdd9ce05146da14bb60f5a17a163d5262

Proper security by explicit port forwarding and explicitly limited capabilities.

```text
http://example.com:8080 {
  redir https://asdas.net{uri}
}

https://example.comt:8443 {
  tls postmaster@example.com
  limits 1mb
  log stdout
  errors stdout
}
```

``` 
caddy -agree -http-port 8080 -https-port 8443 -conf /path/to/config/Caddyfile -log stdout
```

+ external port mapping from external ports 80 and 443 to internal 8080 and 8443 by SDN.

Older versions including the "working" revision respect the flags and are able to request and and renew LE/ACME certificates fine with the config. At least the version defined as "broken" tries to bind to port `443` ignoring the given CLI flag.

oscartbeaumont:
The config.Manager is redeclared causing it to lose the AltTLSALPNPort value which in turn causes CertMagic to not correctly issue certificates when the http-port command line option is set. It now passes through the existing configuration.

Fixes #2407

^

N/A

### 4. Checklist

- [-] I have written tests and verified that they fail without my change
- [-] I have squashed any insignificant commits
- [-] This change has comments for package types, values, functions, and non-obvious lines of code
- [-] I am willing to help maintain this change if there are issues with it later

### DIFF
--- a/caddytls/setup.go
+++ b/caddytls/setup.go
@@ -64,7 +64,7 @@ func setupTLS(c *caddy.Controller) error {
 		})
 		c.Set(CertCacheInstStorageKey, certCache)
 	}
-	config.Manager = certmagic.NewWithCache(certCache, certmagic.Config{})
+	config.Manager = certmagic.NewWithCache(certCache, *config.Manager)
 
 	// we use certmagic events to collect metrics for telemetry
 	config.Manager.OnEvent = func(event string, data interface{}) {


oscartbeaumont:
Yep, That seems fair.

oscartbeaumont:
Instead of the pointer dereference should it just parse the AltTLSALPNPort and possibly the AltHTTPPort?

mholt:
Hmm, you're right. Let's stick with pointer dereference for now, then -- but, first, we need to understand better whether this is actually the cause, and why it is/was happening for you, but not for me.

mholt:
Will be ready to merge this as soon as we understand why it works outside of Docker builds but not inside them.

mholt:
Thanks for the PR, but I think this will be superseded by #2452. Please test that out ASAP!

PR #2453, opened 2019-02-03
Rollback changes creating an additional instance_startup event in #2161

tobya:

When fixing #2161 by moving the emitEvent call to ensure on startup was called at the right point, a new event was created instance_startup.  After discussion it has been decided we dont need this new event and it was a breaking change so it has been reverted.

## Testing

- [ ] Can someone please test that this works on linux.  I have a windows only. 
- [ ] Can someone on linux please check that whatever is supposed to happen after SIGUSR1 still happens? 
- [x]  see if it fixes the issue with caddy not executing `on startup` if caddy.service was included.  __Tested and Works__

 - Anyone else who would like to test. 

#2161 

https://github.com/hacdias/caddy-service/issues/21

none

### 4. Checklist

### DIFF
--- a/onevent/hook/config.go
+++ b/onevent/hook/config.go
@@ -14,8 +14,7 @@ type Config struct {
 
 // SupportedEvents is a map of supported events.
 var SupportedEvents = map[string]caddy.EventName{
-	"startup":          caddy.StartupEvent,
-	"instance_startup": caddy.InstanceStartupEvent,
-	"shutdown":         caddy.ShutdownEvent,
-	"certrenew":        caddy.CertRenewEvent,
+	"startup":   caddy.InstanceStartupEvent,
+	"shutdown":  caddy.ShutdownEvent,
+	"certrenew": caddy.CertRenewEvent,
 }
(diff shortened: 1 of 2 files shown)


mholt:
Looking at this more, I would actually like to see these events related to startup/shutdown:

- `launch` - The Caddy process has been launched
- `startup` - A Caddy instance has been started
- `shutdown` - A Caddy instance has been shut down
- `exit` - The Caddy process is exiting cleanly

@tobya Will that be sufficient for your needs? And this way it doesn't have to be a breaking change.

elcore:
> `launch` - The Caddy process has been launched

Using `caddy.StartupEvent` will not work (Slack comments)

mholt:
Why won't it work? I don't understand -- explain exactly what is needed.

tobya:
@elcore I never quite understood why moving the call to emitEvent solved the issue with hook.service. What is the timing that was preventing the startup event from firing in the main caddy code?

tobya:
@mholt I think the idea of creating new events is a good one.  However what elcore says is correct, the way the code is at the moment the launch event (if it gets called in the same place as  caddy.Startupevent will never run anything in the caddyfile.  

I presume this event is used for something else other than the caddyfile event.

Maybe we should leave any additional changes until after the next release.

PR #2456, opened 2019-02-06
Add a function to SetCAAgreement for embedded Caddy.

sundbry:
Use case: Embedded Caddy server

Setting certmagic.Agreed = true from outside the caddy package does not
set the variable in the vendored instance, so we need a method to set
it safely from outside the package.

The vendored certmagic package gets a different memory address as we see in these logs:

 ```go
log.Printf("Mark4 certmagic.Agreed = %t @ %p", certmagic.Agreed, &certmagic.Agreed)
```

```text
2019/02/06 01:55:13 Mark10 certmagic.Agreed = false @ 0x24a6c2d
2019/02/06 01:55:13 [INFO] Reloading
2019/02/06 01:55:13 Mark4 certmagic.Agreed = true @ 0x24a6bf3
```

The first access is from my application code, and the second access is from inside the caddy package.

### Checklist

- [ x] I have written tests and verified that they fail without my change

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -16,7 +16,7 @@
 //
 // To use this package:
 //
-//   1. Set the AppName and AppVersion variables.
+//   1. Set the AppName and AppVersion variables, and SetCAAgreement(true)
 //   2. Call LoadCaddyfile() to get the Caddyfile.
 //      Pass in the name of the server type (like "http").
 //      Make sure the server type's package is imported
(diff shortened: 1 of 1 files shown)


mholt:
We may move to Go modules very soon... I wonder if this would be necessary then...

sundbry:
That's up to you, that would probably fix the underlying issue if the
modules work like one would expect. However I think if embedded caddy is a
supported use case, there should be an interface to set these options
(rather than hacking on internal packages) from the main package. Or
perhaps adding an optional parameter to pass a flags object through to
Start()?

mholt:
We no longer vendor, so... I think this is no longer necessary. Feel free to correct me if wrong, and we can re-open to look at this.

sundbry:
@mholt We're embedding caddy in wehco/caddy-ingress-controller (for
Kubernetes) where I use this

I'm fine building off a fork for now, rewriting it as a plugin would almost
certainly be possible but a large effort.

PR #2459, opened 2019-02-08
caddytls: Change clustering to be a plugin to the caddytls package

mholt:

Should resolve the failure in
https://github.com/coredns/coredns/pull/2541.

This change is breaking to clustering plugin developers (not Caddy
users), but logical, since only the caddytls package uses CertMagic
directly (the httpserver package also uses it, but only because it also
uses the caddytls plugin); and it is early enough that no clustering
plugins really exist yet.

This will also require a change of devportal
so that it looks for a different registration function, which has moved
to the caddytls package.

https://github.com/coredns/coredns/pull/2541

No user-facing changes, but update the wiki about how to write (and register) a clustering plugin. Devportal needs updating too, since the registration function moved.

### 4. Checklist

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -999,7 +999,6 @@ var (
 	DefaultConfigFile = "Caddyfile"
 )
 
-var clusterPluginSetup int32 // access atomically
 
 // CtxKey is a value type for use with context.WithValue.
 type CtxKey string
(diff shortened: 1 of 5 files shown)


miekg:
Compiled a coredns binary against this change and that starts (normally). So LGTM

mholt:
Awesome, thanks. I'm just gonna make sure these tests pass now and then merge it.

Issue #2366, opened 2018-11-23
Caddyfile lexer bug


dafanasiev:
### 1. What version of Caddy are you using (`caddy -version`)?
Caddy 0.11.1 (+ce0988f Mon Nov 19 11:57:34 UTC 2018) (unofficial)

### 2. What are you trying to do?
I'm try to setup very simple configuration (see below), but caddy wont work normally (because configuration file lexer bug?)

### 3. What is your entire Caddyfile?
```text
https://localhost:8888 {

    header / X-Cert-ClientSubjectDn "{tls_client_s_dn}"

    tls self_signed {clients verify_if_given ca.crt}

}
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
Just put config file near the bin and run `./caddy`

### 5. Please paste any relevant HTTP request(s) here.

```
 curl --insecure -E client.crt:1234 --key client.key -i https://localhost:8888
```

### 6. What did you expect to see?
```
<h1>Its work!</h1>
```

### 7. What did you see instead (give full error messages and/or log)?
```
curl: (35) error:14094438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error
```

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
//you can generate client certs using this manual https://gist.github.com/mtigas/952344
Ok, and now about 'real' bug.
I see that lexer in [line 118](https://github.com/mholt/caddy/blob/master/caddyfile/lexer.go#L118) tokenize input and think that lexems separates by `unicode.IsSpace(ch)`. Its ok, but in my case lexer produce `{client` and `ca.crt}`  (yes, with curves) tokens (bug?)

Im not sure, that  this is not expected, can u explain caddy configuration file rules more detailed?

mholt:
No, I don't think that's the bug. From [the docs](https://caddyserver.com/docs/caddyfile):

> Directives may span multiple lines by opening a block. Blocks are enclosed by curly braces { }. The opening curly brace { must be at the end of the directive's first line, and the closing curly brace } must be the only token on its line:

So your Caddyfile is malformed, but the bug is that Caddy should show an error when parsing the Caddyfile, not silently allowing it to pass.

PR #2464, opened 2019-02-12
Ensure closing curly brace is on its own line

awoodbeck:
This change enforces the documentation that, "the closing curly brace } must be the only token on its line" without breaking the requirement that an environment variable is, "enclosed in curly braces { } without extra whitespace."

#2366 

None.  This PR brings parsing functionality in line with the documentation.

### 4. Checklist

### DIFF
--- a/caddyfile/parse_test.go
+++ b/caddyfile/parse_test.go
@@ -447,6 +447,24 @@ func TestParseAll(t *testing.T) {
 			{"localhost:1234"},
 		}},
 
+		{`localhost {
+			tls self_signed {}
+			}`, false, [][]string{
+			{"localhost"},
+		}},
+
+		{`localhost {
+			tls self_signed {clients verify_if_given ca.crt}
+			}`, true, [][]string{}},
+
+		{`https://localhost:8888 {
+
+			header / X-Cert-ClientSubjectDn "{tls_client_s_dn}"
+	
+			tls self_signed {clients verify_if_given ca.crt}
+	
+			}`, true, [][]string{}},
+
 		{`localhost:1234 {
 		  }
 		  localhost:2015 {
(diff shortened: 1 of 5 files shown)


tobya:
Thanks for the PR.  

This seems to raise a parse error when dealing with placeholders.  The following exits with a parse error.

````
localhost:80 {
    log / d:\caddy\logs\courgettelog.log   "--{user}-- {request_id}"
}
````

but these seems  fine

```
localhost:80 {
    log / d:\caddy\logs\courgettelog.log   "{user} {request_id}"
    log / d:\caddy\logs\courgettelog.log   "{user} {request_id}--"
    log / d:\caddy\logs\courgettelog.log   {common}
}
```

File: caddyfile/parse_test.go [hunk 1]
```diff
@@ -447,6 +447,24 @@ func TestParseAll(t *testing.T) {
 			{"localhost:1234"},
 		}},
 
+		{`localhost {
```
tobya: I think we should move these tests into a seperate function making it obvious we are testing the final } issue.

elcore: I agree

caddyfile/parse_test.go
```diff
@@ -597,6 +615,56 @@ func TestEnvironmentReplacement(t *testing.T) {
...
+			tls self_signed {clients verify_if_given ca.crt}
+	
+			}`, true, [][]string{}},
+
+		{`localhost:80 {
+			log / d:\caddy\logs\default.log   "--{user}-- {request_id}"
+			}`, false, [][]string{}},
+
+		{`localhost:80 {
+			log / d:\caddy\logs\default.log   "request_id}"
```
awoodbeck: This test case and the one above would fail without adding the `Quoted` property to the `Token`.  I could see no other reliable way to infer a token was derived from a quoted string and not a block.

tobya:
Thanks for the changes.  Placeholders can be in lots of places in a caddy file and don't need to be quoted, so  I think if there is a solution it may be in looking at the kind of token that the parser is dealing with and if it is a } checking it is on its own on a line

I'm not sure the updates you've made are a way to solve this issue.  You might need to take a different approach.  Perhaps @mholt might have some pointers.

Its failing now on my caddyfile at 

```
  rewrite / {
  r  (.*)/\.
  to /index.php?path={1}  // this line
}
```

awoodbeck:
Please excuse my ignorance of the code base.  I'm reviewing the code as time allows.  But it seems the challenge here is curly braces are used for delimiters _and_ place holders.  Maybe the correct approach is to better manage the state in the `parser.directive`'s for loop.

PR #2472, opened 2019-02-18
Raise parsing error if errors directive has more than one argument.

tobya:
I found I had written the following  in several places in my caddyfile

````
example.com {
  root c:\web
   log / c:\logs\mylogfilt.txt
  errors /  c:\logs\myerrorlog.txt
}
````
However `error` does not require a path param so the `/` is unneeded

This silently fails and the only place the log is written to is stdout.

This PR raises a parsing error in this case.

I'm not sure how to write tests for this.

### DIFF
--- a/caddyhttp/errors/setup.go
+++ b/caddyhttp/errors/setup.go
@@ -123,6 +123,10 @@ func errorsParse(c *caddy.Controller) (*ErrorHandler, error) {
 			}
 		}
 
+		if len(args) > 1 {
+			return handler, c.Errf("Only 1 Argument expected for errors directive")
+		}
+
 		// Configuration may be in a block
 		err := optionalBlock()
 		if err != nil {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by francislavoie** [comment]
I think you can just add a test like this:

```go
{`errors first.txt second.txt`, true, ErrorHandler{ErrorPages: map[int]string{}, Log: &httpserver.Logger{}}},
```

Over here: https://github.com/mholt/caddy/blob/master/caddyhttp/errors/setup_test.go#L181

tobya:
Thanks @francislavoie  I added those tests now.

caddyhttp/errors/setup.go
```diff
@@ -123,6 +123,10 @@ func errorsParse(c *caddy.Controller) (*ErrorHandler, error) {
 			}
 		}
 
+		if len(args) > 1 {
+			return handler, c.Errf("Only 1 Argument expected for errors directive")
+		}
```
abiosoft: I think this should be lower case. `go lint` would've complained if this was part of `fmt.Errorf`.

mholt:
Thanks everyone!

caddyhttp/errors/setup.go
mholt: Yeah that was bad planning for convention on my part.

PR #2489, opened 2019-02-27
fastcgi: Improve fastcgi SCRIPT_FILENAME generation

longsleep:
This change change the logic how the SCRIPT_FILENAME for fastcgi is
created. IndexFiles are now checked in the matches rule's root instead
of the sites root. In addition a new optional sub-directive 'without'
is added to fastcgi, which allows stripping prefix paths from the
generated SCRIPT_FILENAME.

```
fastcgi /blog/ 127.0.0.1:9000 php {
  without /blog/
  root /usr/share/some-blog-software
}
```

Requests to `/blog/` or `/blog/index.php` will properly find and
load 'index.php' in /usr/share/some-blog-software without any further
rewrites.

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -42,10 +42,8 @@ import (
 
 // Handler is a middleware type that can handle requests as a FastCGI client.
 type Handler struct {
-	Next    httpserver.Handler
-	Rules   []Rule
-	Root    string
-	FileSys http.FileSystem
+	Next  httpserver.Handler
+	Rules []Rule
 
 	// These are sent to CGI scripts in env variables
 	SoftwareName    string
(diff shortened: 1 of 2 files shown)


longsleep:
Will add tests once getting positive signals that this change is something you guys consider worth merging.

abiosoft:
Having specific root for fastcgi was deliberate. You can have caddy and fastcgi server running on different hosts.

longsleep:
> Having specific root for fastcgi was deliberate. You can have caddy and fastcgi server running on different hosts.

Sure - behavior is unchanged on that regard.

tobya:
I wonder should `root` here get a name change to something like `alternative_root`?

longsleep:
> I wonder should root here get a name change to something like alternative_root?

Well it's more the `fastcgi_root` - but that's already implied since it's inside the fastcgi directive. But u agree that it could be clearer. Might be a good idea to change while retraining backwards compatibility.

mholt:
@longsleep Thanks for this PR. 

> Will add tests once getting positive signals that this change is something you guys consider worth merging.

I think at this point, the efforts are better spent in making this work in Caddy 2 (it's on the v2 branch). We don't quite have fastcgi support migrated yet, but when we do, I'd like your help verifying that SCRIPT_FILENAME is correct. Would you be willing to help with that?

(Closing this PR for now, further discussion is welcome.)

longsleep:
> Would you be willing to help with that?

Sure - will check Caddy 2 and see what needs to be done there. Will open another PR/Issue if required. Thanks!

PR #2498, opened 2019-03-03
Easy source

Conor-Burns:
### Hi
This makes it easier to build from source with plugins.
- adds a list of plugin into run.go that are commented by default
- add usage into README.md


whitestrake:
Hi @Conor-Burns, thanks for submitting a PR.

That said, I'm pretty ambivalent on these additions.

On one hand, this will have to be consistently maintained as plugins come and go; on the other hand, we already maintain a middleware chain in `plugin.go`.

There are already some tools and examples for handling plugins when building Caddy from source (such as [abiosoft/caddyplug](https://github.com/abiosoft/caddyplug), and [this script](https://github.com/abiosoft/caddy-docker/blob/master/builder/builder.sh) that makes use of it).

I just don't think this is necessary. The problem it solves just isn't a big problem at all.

Conor-Burns:
ah ok wasn't aware of caddyplug and I can fully understand that the plugin list changes frequently.

PR #2506, opened 2019-03-05
Add TLSv1.3 to list of known ciphers.

### LINKED ISSUE #2505: tls_cipher is not logged correctly with TLSv1.3

0.11.5.

Have caddy log the `tls_cipher` that the client and server negotiated with TLSv1.3

Can re reproduced simply by running `caddy`.

I expect the logs to contain the TLS cipher like "TLS_AES_128_GCM_SHA256".

"UNKNOWN" is seen in the logs.

It looks like `config.go` needs the 1.3 cipher suites added to `SupportedCiphersMap` which is used by `GetSupportedCipherName`

https://github.com/mholt/caddy/blob/72d0debde6bf01b5fdce0a4f3dc2b35cba28241a/caddytls/config.go#L457

The lack of the TLSv1.3 cipher suite IDs means that the logged suite will be "UNKNOWN", and that the `tls_cipher` environment variable that the fastcgi module sets here:

https://github.com/mholt/caddy/blob/053373a38519d8cdf4ee7582ed9dc6ce239597cc/caddyhttp/fastcgi/fastcgi.go#L341

The [go docs][1] seem to indicate that the TLSv1.3 constants are there, so I think this is a matter of adding a few more IDs to the map, something like:

```go
...
"TLS_AES_128_GCM_SHA256": tls.TLS_AES_128_GCM_SHA256,
"TLS_AES_256_GCM_SHA384": tls.TLS_AES_256_GCM_SHA384,
"TLS_CHACHA20_POLY1305_SHA256": tls.TLS_CHACHA20_POLY1305_SHA256
```

Looking a bit hard through, it looks like this map is also used for configuring caddy. Since Go's ciphers are not configurable for TLSv1.3, adding them to the existing map is likely not correct.

https://github.com/mholt/caddy/blob/72d0debde6bf01b5fdce0a4f3dc2b35cba28241a/caddytls/setup.go#L200

[1]: https://golang.org/pkg/crypto/tls/

vcsjones:
This adds the TLSv1.3 ciphers to the list of known ciphers which is required for logging. It also introduces a means of making a cipher disallowed from configuration but still getting it be used in the `SupportedCiphersMap` map.

The disallow list appears to be necessary. If the TLSv1.3 suite is in the `SupportedCiphersMap` and a user attempts to put it in their Caddyfile, Caddy will panic later during the configuration process. Simplest approach for now seemed to be to just disallow them.

Note I am not a Go developer. This is my second go PR ever. 😅

Fixes #2505

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -444,7 +444,8 @@ func GetSupportedProtocolName(protocol uint16) (string, error) {
 	return "", fmt.Errorf("name: unsupported protocol")
 }
 
-// SupportedCiphersMap has supported ciphers, used only for parsing config.
+// SupportedCiphersMap has supported ciphers, used only for parsing config
+// and logging.
 //
 // Note that, at time of writing, HTTP/2 blacklists 276 cipher suites,
 // including all but four of the suites below (the four GCM suites).
@@ -455,6 +456,9 @@ func GetSupportedProtocolName(protocol uint16) (string, error) {
 //
 // This map, like any map, is NOT ORDERED. Do not range over this map.
 var SupportedCiphersMap = map[string]uint16{
+	"AES-128-GCM-SHA256":                 tls.TLS_AES_128_GCM_SHA256,
+	"AES-256-GCM-SHA384":                 tls.TLS_AES_256_GCM_SHA384,
+	"CHACHA20-POLY1305-SHA256":           tls.TLS_CHACHA20_POLY1305_SHA256,
 	"ECDHE-ECDSA-AES256-GCM-SHA384":      tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
 	"ECDHE-RSA-AES256-GCM-SHA384":        tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
 	"ECDHE-ECDSA-AES128-GCM-SHA256":      tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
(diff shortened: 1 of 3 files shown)


vcsjones:
Ah. Hm. AppVeyor is using Go 1.11. What is the minimum go version required for building Caddy? Is 1.11 support intended?

francislavoie:
Ignore the AppVeyor CI failures for now. They don't have Go 1.12 support yet :-1: https://github.com/appveyor/ci/issues/2875

vcsjones:
@francislavoie thanks for that. The Travis output is a bit uh, verbose from linting errors. Did Travis failure only from linting errors, or where should I be looking to see if I broke anything? I (thought) I ran tests locally, but it's entirely possible (read: likely) I did not do that correctly.

mholt:
When there is a CI error, Travis expands the whole thing. The actual failure here is this error near the top:

```
$ gometalinter --disable-all -E vet -E gofmt -E misspell -E ineffassign -E goimports -E deadcode --tests --vendor ./...
caddytls/config.go:1::warning: file is not gofmted with -s (gofmt)
The command "gometalinter --disable-all -E vet -E gofmt -E misspell -E ineffassign -E goimports -E deadcode --tests --vendor ./..." exited with 1.
```

So you need to run `gofmt -s` on the file you edited, config.go.

caddytls/config.go
```diff
@@ -471,6 +475,14 @@ var SupportedCiphersMap = map[string]uint16{
 	"RSA-3DES-EDE-CBC-SHA":               tls.TLS_RSA_WITH_3DES_EDE_CBC_SHA,
 }
 
+// NonConfigurableCiphersMap is a list of ciphers that are understood but
+// cannot be configured such as the case of TLSv1.3.
+var NonConfigurableCiphersMap = map[string]struct{}{
+	"AES-128-GCM-SHA256":       struct{}{},
+	"AES-256-GCM-SHA384":       struct{}{},
+	"CHACHA20-POLY1305-SHA256": struct{}{},
+}
```
mholt: I suspect using `struct{}{}` instead of `{}` is the reason for the failure.

mholt:
@vcsjones Hiya! I can still get this in before 1.0, but with all the changes to our CI and stuff, maybe you could re-sync with master and submit a new PR? Would love to have you credited (otherwise I'd just do it myself...).

vcsjones:
@mholt thank you for the nudge, apologies for the falling off on this. I'll close this for now and open a new PR soon.

Issue #2466, opened 2019-02-13
memory leak with SIGUSR1


chengsurf:
I have used caddy as proxy engine for a period of time. Recently I find the free memory of the device is decreasing slowly.

The  corresponding scenario is like this: At the device, we launch a process(named agent) which provides interface for site configuration(add, delete or change) and then notify caddy to reload new  configurations through sending the USR1 signal. 

After caddy running one week, we can get the top memory-consumed functions and list below:
      flat  flat%   sum%        cum   cum%
 1284.11MB 12.44% 12.44%  1284.11MB 12.44%  github.com/mholt/caddy/caddyfile(*parser).directive
  856.70MB  8.30% 20.74%  1195.28MB 11.58%  github.com/mholt/caddy/vendor/golang.org/x/net/http2.configureTransport
  706.98MB  6.85% 27.59%   706.98MB  6.85%  bytes.makeSlice
  577.11MB  5.59% 33.18%  1772.39MB 17.17%  github.com/mholt/caddy/caddyhttp/proxy.NewSingleHostReverseProxy
  573.15MB  5.55% 38.73%  3151.70MB 30.53%  github.com/mholt/caddy/caddyhttp/proxy.NewStaticUpstreams
  430.62MB  4.17% 52.63%   440.62MB  4.27%  net/textproto.MIMEHeader.Add
  409.11MB  3.96% 56.59%   409.11MB  3.96%  github.com/mholt/caddy/caddytls.NewConfig
  365.13MB  3.54% 60.13%   886.57MB  8.59%  github.com/mholt/caddy/caddyhttp/httpserver.(*httpContext).InspectServerBlocks

The function "proxy.NewStaticUpstreams" has taken up so much memory. I doubt if there is 
memory leak.  After reading the code of the function “NewStaticUpstreams”, I still can't find the reason.

As a comparison，I restart the caddy process and find the function "proxy.NewStaticUpstreams" only consume about 3M.

### 1. What version of Caddy are you using (`caddy -version`)?
0.11.0

### 2. What are you trying to do?
add new site periodicity and then caddy reload new site configuration

### 3. What is your entire Caddyfile?
site example:
http://www.testexample.com {
    proxy / 10.67.8.123 {
        websocket
        transparent
        keepalive 100
        timeout 3m
    }
}

total site number: about 4000

### 4. How did you run Caddy (give the full command and describe the execution environment)?
caddy -conf ./Caddyfile

[...]

PR #2507, opened 2019-03-07
caddy: close keep-alive connections after reloads

chengsurf:
#2466 

Thank your review!


mholt:
Can you sign our CLA?

Also, sync up with master. We have updated our CI systems. Are you sure this works for you? There are no tests... does this break websocket connections?

Issue #2479, opened 2019-02-24
A way to print the environment


mholt:
## 1. What would you like to have changed?

Print the environment

`caddy -env` could be all that is needed. Just loop through `os.Environ()` and print it out.

## 2. Why is this feature a useful, necessary, and/or important addition to this project?

A lot of times the environment is not set how people think because of complexities running in containers or with systemd.

## 3. What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

Currently recompiling to print debug info manually

jimen0:
Hi.

I'd love to start contributing to Caddy and I think this one could be a perfect first contribution. Could you please ellaborate a bit more what do you want it to do?

My guess is that you want a new flag to print the current `os.Environ()` values, am I right? But as far as I see, current master already has an `-env` flag defined [here](https://github.com/mholt/caddy/blob/397d67876cb3063138754250a1e4a01881e60ec2/caddy/caddymain/run.go#L56). My idea is to add a new flag (`-print-env`, `-printenv`, `-show-env`, please, let me know which one you think fits better) that prints the `os.Environ()` values right after loading the environment from file [here](https://github.com/mholt/caddy/blob/397d67876cb3063138754250a1e4a01881e60ec2/caddy/caddymain/run.go#L119).

Also, how would you like me to show the results? One on each line `VARIABLE=VALUE`? In that case, PR would add this code + the new flag.

```go
for _, v := range os.Environ() {
	fmt.Printf("%s\n", v)
}
```

Kind regards

mholt:
> But as far as I see, current master already has an -env flag defined here.

Dang, you're right. I think it'd be better to rename that flag to `-envfile` in the long run, so let's go ahead and do that.

And yes, your proposed output will probably be sufficient! At least for an initial PR.

PR #2517, opened 2019-03-11
caddy: add a way to print the environment

jimen0:
This PR implements #2479.

There are 2 commits in here:

1. Rename the original `env` flag to `envfile`.
2. Add the new `env` flag implementation which prints environment variables.

Let me know if you wish me to add a log message before printing environment variables and which text should it have.

I navigated [the docs](https://godoc.org/github.com/mholt/caddy) looking for places where it was necessary to update the meaning of each flag, but found nothing.

Kind regards

### DIFF
--- a/caddy/caddymain/run.go
+++ b/caddy/caddymain/run.go
@@ -53,7 +53,7 @@ func init() {
 	flag.StringVar(&disabledMetrics, "disabled-metrics", "", "Comma-separated list of telemetry metrics to disable")
 	flag.StringVar(&conf, "conf", "", "Caddyfile to load (default \""+caddy.DefaultConfigFile+"\")")
 	flag.StringVar(&cpu, "cpu", "100%", "CPU cap")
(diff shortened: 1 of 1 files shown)


tobya:
I like this change, but I am a little concerned at the breaking change.  @mholt 

It does not seem necessary to me for this to be breaking.  May I suggest leaving env as is and adding another

````
flag.StringVar(&envFile, "env", "", "Path to file with environment variables to load in KEY=VALUE format")
flag.BoolVar(&printEnv, "printenv", false, "Print environment variables")

````

mholt:
I think `envfile` is more descriptive, accurate, and is still concise, and is important to make the change before we solidify at 1.0. We'll definitely mention the change in the release notes so it's clear.

mholt:
@tobya Hey good news, `-env` was never actually documented on our CLI page, so... it probably won't break anyone, or at least not anyone who wasn't involved with the creation of the flag.

PR #2523, opened 2019-03-16
caddy: optimize the code order to make the logic more accurate

johncming:
The original logic is a little fuzzy. 

For example

if serverType is equal to  "",  it happens that there is such a key in the map, but it is not the only key. 

https://github.com/mholt/caddy/blob/4a3c9b428622f91744f2e74f791e2affdc186d73/caddy.go#L839-L843

This code is easy to skip.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -833,10 +833,6 @@ func startServers(serverList []Server, inst *Instance, restartFds map[string]res
 }
 
 func getServerType(serverType string) (ServerType, error) {
-	stype, ok := serverTypes[serverType]
-	if ok {
-		return stype, nil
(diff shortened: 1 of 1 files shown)


johncming:
@mholt

mholt:
Good change, thank you! Sorry I missed this.

Can you sync with master again? We updated our CI systems so they actually work now.

mholt:
This slipped through my fingers again. 😞 But fortunately this is irrelevant in Caddy 2 since this code is very different in Caddy 2. Thanks for the contribution! Would you like to help with Caddy 2?

PR #2527, opened 2019-03-17
caddyhttp/httpserver: remove unused listener.

johncming:
Caching `net.listener` is actually not necessary. The original code didn't make much sense.
Also, if TLS is supported, the net.listener needs to be updated. Considering that there is no practical use for it at present, it is better to remove it.

https://github.com/mholt/caddy/blob/bc5df3b383ee0add2e7015fabc83f7c8d093b30e/caddyhttp/httpserver/server.go#L46-L47

https://github.com/mholt/caddy/blob/bc5df3b383ee0add2e7015fabc83f7c8d093b30e/caddyhttp/httpserver/server.go#L313-L315

https://github.com/mholt/caddy/blob/bc5df3b383ee0add2e7015fabc83f7c8d093b30e/caddyhttp/httpserver/server.go#L323


mholt:
Interesting! This might be an artifact from an older version of Caddy. Thanks for finding this.

PR #2528, opened 2019-03-17
New config option tls.dump_cert

furkanmustafa:
This option adds support for dumping auto generated, self-signed certificate into a file in specified path.

Please let me know if I should improve variable/conf naming, or any part of the code at all.

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -79,6 +79,9 @@ type Config struct {
 	// that we generated in memory for convenience
 	SelfSigned bool
 
+	// Dump the self-signed certificate to a path
+	DumpSelfSignedCert string
+
 	// The email address to use when creating or
 	// using an ACME account (fun fact: if this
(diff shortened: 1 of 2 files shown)


furkanmustafa:
Sorry for sending without discussion first. This was required for my case and worked for me (I'm using my fork).

Feel free to close without merge. We(somebody?) can re-introduce with proper planning.
Because I'm not good with golang and just hacked this with 30 google searches :D 

Also, I believe that there should also be an option to provide a CA key/cert, for self-signing certificates, so user can just set one CA certificate as trusted, and caddy can just re-issue as many signed certificates as necessary. I guess that'd even (partially) cover for the issue you've mentioned. And that'd also eliminate my need for extracting the issued certificate.

mholt:
Closing since Caddy 2 approaches this differently. Thanks for the contribution!

PR #2553, opened 2019-04-07
caddy: remove unused code.

johncming:
there are two same definations:

https://github.com/mholt/caddy/blob/04789a24467cbf47a6e9adfc270df01b57295230/caddy.go#L64-L65

https://github.com/mholt/caddy/blob/04789a24467cbf47a6e9adfc270df01b57295230/caddyhttp/httpserver/plugin.go#L716-L717

the first is unused, whether it is legacy code?

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -41,7 +41,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/mholt/caddy/caddyfile"
 	"github.com/mholt/caddy/telemetry"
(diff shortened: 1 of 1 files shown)


mholt:
Hmm, actually I think we meant to use that... it should force a graceful shutdown to terminate after some time. In fact, I think we just need to use it right here to give the context a timeout: https://github.com/mholt/caddy/blob/04789a24467cbf47a6e9adfc270df01b57295230/caddyhttp/httpserver/server.go#L487

mholt:
Closing for inactivity (should probably apply the default rather than remove the variable entirely); irrelevant for Caddy 2, anyway. :) Thanks for the contribution! Would you like to help contribute to Caddy 2? Checks like that could be very helpful.

Issue #2514, opened 2019-03-11
Add read/write timeouts for proxy module


akawhy:
## 1. What would you like to have changed?

Add read/write timeouts for proxy module. At this time, `time` directive in proxy module just set the connection timeout.   

## 2. Why is this feature a useful, necessary, and/or important addition to this project?

The backend server maybe buggy or broken.  

## 3. What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

undef

## 4. Please link to any relevant issues, pull requests, or other discussions.

undef

mholt:
Done in v2: https://github.com/caddyserver/caddy/wiki/v2:-Documentation#httphandlersreverse_proxy

PR #2564, opened 2019-04-12
proxy module: new idle_conn_timeout, response_header_timeout directiv…

akawhy:
Add two direcives:

* idle_conn_timeout: idle timeout for keepalive connections
* response_header_timeout: timeout for reading upstream response header

Add one placeholder:

* upstream_conn_ms: how long it takes to connect to the upstream

related issue: #2514


akawhy:
Example caddyfile:

```

:8080 {
    log / stdout "{remote} - {user} [{when}] \"{method} {uri} {proto}\" {status} {size} \"{>Referer}\" \"{>User-Agent}\" {upstream_conn_ms}"
    proxy / http://backend:8080 {
        transparent
        keepalive 10
        timeout 10s
        idle_conn_timeout 30s
        response_header_timeout 30s
    }
}

```

mholt:
Thanks for contributing this, @akawhy!

Sorry it took a while to get around to this. We've been working on Caddy 2, which is where almost all of our development effort is going now. Would you be interested in contributing this to Caddy 2 instead? I think the effort will be much better spent there. We are still laying the foundation of the new proxy though: https://github.com/caddyserver/caddy/issues/2679

(Closing this PR then, but further discussion welcome.)

PR #2575, opened 2019-04-22
Support relative unix socket file path

stefansundin:
Hi there,

I like to provide a Caddyfile that is portable, and that does not consume a port. Socket files are perfect for this, but I could not figure out a way to make a relative path work.

Here's the Caddyfile that I would like to add to my [github-activity](https://github.com/stefansundin/github-activity) project:

```
http://github-activity.localhost {
  on startup bundle exec puma -b unix://tmp/puma.sock
  proxy / unix:tmp/puma.sock {
    transparent
  }
}
```

Let me know if you want me to change anything. Thanks!

### DIFF
--- a/caddyhttp/proxy/upstream.go
+++ b/caddyhttp/proxy/upstream.go
@@ -26,6 +26,7 @@ import (
 	"net/textproto"
 	"net/url"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


stefansundin:
I have another issue that is a bit related to this.. if I have a couple of projects, each with their own Caddyfile, and then import them all in a central Caddyfile, like so:

```
import github-activity/Caddyfile
import rssbox/Caddyfile
```

Then there are a bunch of things that don't work. For example, the relative paths for `unix:` that I just added support for, and also my `on startup` command above won't work. I see that in the `import` code, `p.Dispenser.File()` is used to get a path relative to the original Caddyfile. That should work to fix the `unix:` relative file paths. To solve the `on startup` issue, we'd need to set `cmd.Dir` in `hook.go` to the directory of the Caddyfile that is being parsed.

I don't know how to refactor the code so that all the paths and directories go where they should be. Can I get some help with that?

mholt:
Relative paths in config files are hard, especially when files can include other files. What is the basis for the relative paths?

In Caddy 2, there are no config files (sort of), so this problem goes away. All paths are relative to the cwd. The Caddyfile will still be an option, along with importing (including) other files, but I suppose any relative paths would be relative to that file's path, not the base Caddyfile path or the cwd.

Would you like to help with the Caddyfile adapter in this regard? I've mostly ported over the v1 adapter and once I work out its fundamental architecture I'll push it to GitHub... 

(Closing this PR so as to focus efforts on Caddy 2.)

PR #2582, opened 2019-04-24
 fix variable 'sort' collides with package name

Inconnu08:


tobya:
Hello, Thank you for the pull request.

Can you please provide more information about why this is needed and where the collision occurs?

Inconnu08:
@tobya, variable `sort` collides with the sort package used here - `Import "sort"` hence refactored to `sortListing`

Inconnu08:
@mholt any update?

PR #2590, opened 2019-04-26
implement brotli compression

paolobarbolini:
This PR adds the option to compress responses with brotli when the client supports it.

### DIFF
--- a/caddyhttp/brotli/brotli.go
+++ b/caddyhttp/brotli/brotli.go
@@ -1,0 +1,163 @@
+// Copyright 2015 Light Code Labs, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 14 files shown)


mholt:
Thanks for the contribution!

Any reason this has to be included as a default plugin though? (I'm of the impression that brotli is slower at compressing than gzip, thus resources should be pre-compressed before serving responses.)

mholt:
Thanks for all your work on this.

This has been added to Caddy 2, however we found that the encoder is extremely slow. Like, waaay slow causing network requests to hang. It's still available in Caddy 2 but its use will probably be discouraged in favor of pre-compressing assets with brotli until encoder performance improves. As such, it probably shouldn't be a standard module (it may not be bundled in Caddy 2 by default either; we'll see).

nanmu42:
May be give [github.com/google/brotli/go/cbrotli](https://pkg.go.dev/github.com/google/brotli/go/cbrotli?tab=doc) a try?

mholt:
No, because it uses cgo. We need a pure-Go brotli encoder.

Azareal:
https://github.com/valyala/fasthttp/issues/798 There have been some big perf improvements over the past week, you may want to take another look at it when the release comes out.

PR #2618, opened 2019-05-21
Fix graceful shutdown

dvic:
Currently, the instance waitgroup is decremented twice in `startServers()`: once when `Serve()` is finished and once when `ServePacket()` is finished. However, with a graceful shutdown, `Serve()` returns before the server has actually finished shutting down all active connections. This patch increases the wait group by one when the server is shut down so that the program only exits when all the server instances have finished serving their connections.

Steps for me that showed graceful shutdown not to work currently:
1. `mkfile 2m foo.txt`
2. `go run github.com/mholt/caddy/caddy -log stdout -grace 1m`
3. `curl -v -o /dev/null  --limit-rate 300k http://localhost:2015/foo.txt`
4. `kill -s SIGTERM PID_OF_SERVER`

Without this patch, the curl command fails: 
```
curl: (18) transfer closed with 746120 bytes remaining to read
```


dvic:
The `fastcgi` test seems to be failing on Mac OS.

dvic:
@apat5358 Does anyone else need to look at this before it gets merged?

dvic:
@mholt could you have a look at this? Can you confirm the graceful shutdown bug or am I missing something? Thanks 👍

PR #2630, opened 2019-05-29
If an encoding has a qvalue strip it

rouzier:
---
name: Pull request
about: Take in consideration the the qvalue when matching Accept-Encoding
title: ''
labels: ''
assignees: ''
---

Take in consideration the the qvalue when matching Accept-Encoding.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.3

## 2. Checklist

### DIFF
--- a/caddyhttp/staticfiles/fileserver.go
+++ b/caddyhttp/staticfiles/fileserver.go
@@ -191,7 +191,12 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request) (int, err
 		// see if the client accepts a compressed encoding we offer
 		acceptEncoding := strings.Split(r.Header.Get("Accept-Encoding"), ",")
 		accepted := false
-		for _, acc := range acceptEncoding {
+		for _, acc_q := range acceptEncoding {
+			acc, qval := parseQvalue(acc_q)
+			if qval == 0.0 {
+				continue
+			}
+
 			if strings.TrimSpace(acc) == encoding {
 				accepted = true
 				break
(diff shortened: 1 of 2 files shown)


mholt:
Thanks... someone will review this eventually. :)

What is your specific use case / need for this (advanced content negotiation)? i.e. what are your actual requirements?

rouzier:
None.
I was reviewing the code and I noticed that qvalue was not properly handled.
So I supplied a fix.

mholt:
Okay. I appreciate it, I really do! But since there isn't an actual use case for it yet, it isn't a huge priority for me personally (Caddy 2 is currently on my plate, and if q-factor'ed content negotiation is important, I can work it into Caddy 2). But I see that this is a decent change since it has tests; albeit the feature is only partially implemented (which is still OK).

If someone considers this a priority they are welcome to review it and I can see about merging it in, I just can't get to it myself quite right now. Thanks!

rouzier:
Feel free to close the PR if it does not fit your needs or if you know if
that section of code will be rewritten

mholt:
FWIW, Caddy 2 does have content negotiation respecting q-values with regards to dynamic encodings: https://sourcegraph.com/github.com/caddyserver/caddy@0d3f99e85a089a6a772ae38f426b5cd5f2f4583f/-/blob/modules/caddyhttp/encode/encode.go#L232

However, content negotiation _in general_ still needs work. I am tracking this in https://github.com/caddyserver/caddy/issues/2665. Although, I am not sure if a matcher is the right approach anymore. (Discuss before implementing!) But would be happy to have your help on it if you're interested.

PR #2639, opened 2019-06-07
Add permission plugin directive

dhaavi:
I'm finally looking to make my plugin for caddy officially available.

The code currently lives at https://github.com/dhaavi/caddy-authplugger, but will be moved to https://github.com/dhaavi/caddy-permission within the next days.

Like a couple other plugins, this plugin is about authentication, but, additionally also about authorization:
Instead of just relaying authenticating users to another piece of software, it aims to bring fine-grained permission management to otherwise unsupported web applications.

For example, I've been using this code for over a year now to enforce fine-grained permissions on the filebrowser plugin. It plays fairly nice for the basic feature set, I must say. (I have not yet bothered to integrate all features with the permission model)

Technically speaking, it fetches and authenticates users via (1) the config file (HTTP Basic Auth), (2) an easy json-based api and/or (3) TLS client certificates. It then enforces a HTTP Method + Request Path permission model on the user. It also handles websockets.

I changed the name in order for users to better differentiate this plugin from other auth plugins.

EDITED TO ADD: There will be minor improvements and some maintenance within the next days (go modules!).

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -656,13 +656,14 @@ var directives = []string{
 	"s3browser", // github.com/techknowlogick/caddy-s3browser
 	"nobots",    // github.com/Xumeiquer/nobots
 	"mime",
-	"login",     // github.com/tarent/loginsrv/caddy
-	"reauth",    // github.com/freman/caddy-reauth
-	"extauth",   // github.com/BTBurke/caddy-extauth
-	"jwt",       // github.com/BTBurke/caddy-jwt
-	"jsonp",     // github.com/pschlump/caddy-jsonp
-	"upload",    // blitznote.com/src/caddy.upload
-	"multipass", // github.com/namsral/multipass/caddy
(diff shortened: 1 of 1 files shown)


dhaavi:
I done so far, just waiting for this to be merged in order to publish.
See https://github.com/dhaavi/caddy-permission.

dhaavi:
If I understand correctly, I will have to wait for the next release of caddy to be able to use the new plugin with official images.

This is mildly confusing, as the plugin is already published on the docs page of caddy. Which means, that a download with plugin would succeed, but when trying to actually use the plugin it will complain: `Error during parsing: Unknown directive 'permission'`.

While technically not a bug, users will perceive this as a broken build and might even report it as such.

(I fixed the license)

abiosoft:
@dhaavi you can temporarily disable your plugin in the developer portal

dhaavi:
Thanks for the hint!

I didn't want to disable it, as users could be confused. I addad a note about compatibility, which I will remove as soon as caddy >=1.0.1 ships.

PR #2644, opened 2019-06-11
Fix goroutine leak when restarting Caddy

### LINKED ISSUE #2642: Goroutine leak in caddytls NewConfig
## Which version of Caddy are you using?

tags/v1.0.0

## What are you trying to do?

This bug reproduces when I instruct Caddy to reload the Caddyfile multiple times before the final shutdown. A goroutine leaks each time I send the `SIGUSR1` signal. 

## What is your Caddyfile?

```text
:33000 {

    pprof

    prometheus

    timeouts {
        idle 4m
        header 2m
    }

}
```

## Please paste any relevant HTTP request(s) here.

If I send several `SIGUSR1` signals to Caddy and then look at the number of goroutines, I see that there is a leak:

```
curl "localhost:9180/debug/pprof/goroutine?debug=1"
```

```
10 @ 0x403172f 0x4007f7a 0x4007f50 0x4007c7b 0x449270b 0x405ed21
#	0x449270a github.com/mholt/caddy/caddytls.NewConfig.func2+0x8a	github.com/mholt/caddy/caddytls/config.go:145
```

## What did you expect to see?

I expected to see a single goroutine for the `NewConfig` function. 

## Why is this a bug, and how do you think this should be fixed?

This is a bug for those for which Caddy is long running and reloads the configs many times throughout its execution. The number of goroutines could easily grow to thousands if an automated event triggered a Caddyfile reload. 

--- 

I have made the following code change (in caddy/caddytls/config.go) to resolve this bug and will happily make a pull request.

```go
storageCleaningTicker := time.NewTicker(12 * time.Hour)
done := make(chan bool)
go func() {
	for {
		select {
		case <-done:
			storageCleaningTicker.Stop()
			return
		case <-storageCleaningTicker.C:
			certmagic.CleanStorage(certmagic.Default.Storage, certmagic.CleanStorageOptions{
				OCSPStaples: true,
			})
		}
	}
}()
inst.OnShutdown = append(inst.OnShutdown, func() error {
	certCache.Stop()
	done <- true
	close(done)
	return nil
})
```

jaredririe:
## What does this change do, exactly?

Each time the Caddyfile reloads and Caddy is restarted, `caddytls.NewConfig` starts a goroutine for cleaning the certificate storage. This goroutine ranges over a `time.Ticker` channel; although `Stop` is called on this ticker, `Stop` does not close the underlying channel so the goroutine never exits.

This change adds an additional channel that is listened to in the certificate cleaning goroutine so it can exit on restarts.

## Please link to the relevant issues.

Fixes #2642 

## Which documentation changes (if any) need to be made because of this PR?

None

## Checklist

## Open Discussion/Questions

1. I did not write any new tests for this code because it is contained within a function `caddytls.NewConfig` which is currently untested. I can carve out some time to test the entire function, though it feels out of scope for this PR. 
2. I observed that calling `certmagic.CleanStorage` is delayed by 12 hours. If a restart always occurs before 12 hours, this method is never called. Is this concerning? Is the certificate storage cleaned on restart or by some other means? (I don't believe that I am introducing this problem as calling `Stop` would have the same effect.)

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -140,17 +140,26 @@ func NewConfig(inst *caddy.Instance) (*Config, error) {
 				return certmagic.Default, nil
 			},
 		})
+
 		storageCleaningTicker := time.NewTicker(12 * time.Hour)
+		done := make(chan bool)
 		go func() {
-			for range storageCleaningTicker.C {
-				certmagic.CleanStorage(certmagic.Default.Storage, certmagic.CleanStorageOptions{
-					OCSPStaples: true,
-				})
+			for {
+				select {
+				case <-done:
+					storageCleaningTicker.Stop()
+					return
+				case <-storageCleaningTicker.C:
+					certmagic.CleanStorage(certmagic.Default.Storage, certmagic.CleanStorageOptions{
+						OCSPStaples: true,
+					})
+				}
 			}
 		}()
 		inst.OnShutdown = append(inst.OnShutdown, func() error {
 			certCache.Stop()
-			storageCleaningTicker.Stop()
+			done <- true
+			close(done)
 			return nil
 		})
 


mholt:
> I did not write any new tests for this code because it is contained within a function caddytls.NewConfig which is currently untested. I can carve out some time to test the entire function, though it feels out of scope for this PR.

Honestly, Caddy 2 is coming quickly enough that I wouldn't worry about it much.

> I observed that calling certmagic.CleanStorage is delayed by 12 hours. If a restart always occurs before 12 hours, this method is never called. Is this concerning? Is the certificate storage cleaned on restart or by some other means? (I don't believe that I am introducing this problem as calling Stop would have the same effect.)

Eh, it's not too common that configs are reloaded that often; it just means that some stale OCSP staples might stick around in storage longer than necessary, but it's unlikely that anything detrimental would happen.

Thanks for the PR!

PR #2648, opened 2019-06-14
Verify client cert filenames

### LINKED ISSUE #2641: Failed to request client certs with multiple addresses

```
Caddy v1.0.0 (h1:KI6RPGih2GFzWRPG8s9clKK28Ns4ZlVMKR/v7mxq6+c=)
```

Enable client cert verification for multiple addresses.

## 3. What is your Caddyfile?
```text
$ cat Caddyfile

(tls) {
  tls webmaster@namunamu.com {
    dns cloudflare
    clients verify_if_given /etc/certs/ca/certs/ca.crt
  }
}

namunamu.com {
  import tls
}
namunamu.com/var/ {
  import tls
}
```

```
env CLOUDFLARE_EMAIL=<REDACTED> CLOUDFLARE_API_KEY=<REDACTED> caddy -agree
```

## 5. Please paste any relevant HTTP request(s) here.

skipped

```
Activating privacy features... done.

Serving HTTPS on port 443
https://namunamu.com
https://namunamu.com/var/


Serving HTTP on port 80
http://namunamu.com
http://namunamu.com/var/
```

```
Activating privacy features... done.
incompatible TLS configurations for the same SNI name (namunamu.com) on the same listener: multiple hosts requiring client authentication ambiguously configured
```

I think the tls config compatibility detector may have flaw in it. I'm not familiar with the code base/architecture so it's just my guess.

## 9. What are you doing to work around the problem in the meantime?

I use a subdomain instead of a path for the particular block. i.e. replace `namunamu.com/var/` with `var.namunamu.com`. But it doesn't solve the problem.

## 10. Please link to any related issues, pull requests, and/or discussion.

This issue (https://github.com/mholt/caddy/issues/2251) may be related, but the situations we stuck with are different.

## Bonus: What do you use Caddy for? Why did you choose Caddy?

Less boilerplate config than nginx - ideal for personal projects!

shouya:

Fix #2641.

#2641.

Nothing I believe.

## 4. Checklist

### DIFF
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -562,6 +562,11 @@ func assertClientCertsCompatible(cfg1, cfg2 *Config) error {
 	// The order of client CAs matters
 	for i, v := range ccerts1 {
 		if v != ccerts2[i] {
+			// Two hosts defined on the same listener are not compatible if they
+			// have ClientAuth enabled, because there's no guarantee beyond the
+			// hostname which config will be used (because SNI only has server name).
+			// To prevent clients from bypassing authentication, require that
+			// ClientAuth be configured in an unambiguous manner.
 			return fmt.Errorf("multiple hosts requiring client authentication ambiguously configured")
 		}
 	}
--- a/caddytls/config.go
+++ b/caddytls/config.go
@@ -395,9 +395,9 @@ func assertConfigsCompatible(cfg1, cfg2 *Config) error {
 		return fmt.Errorf("maximum TLS version mismatch")
 	}
 
-  if err := assertClientCertsCompatible(cfg1, cfg2); err != nil {
-    return err
-  }
+	if err := assertClientCertsCompatible(cfg1, cfg2); err != nil {
+		return err
+	}
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


caddytls/config.go
```diff
@@ -397,7 +397,7 @@ func assertConfigsCompatible(cfg1, cfg2 *Config) error {
 	if c1.ClientAuth != c2.ClientAuth {
 		return fmt.Errorf("client authentication policy mismatch")
 	}
-	if c1.ClientAuth != tls.NoClientCert && c2.ClientAuth != tls.NoClientCert && c1.ClientCAs != c2.ClientCAs {
+	if c1.ClientAuth != tls.NoClientCert && c2.ClientAuth != tls.NoClientCert && !StringSliceEqual(cfg1.ClientCerts, cfg2.ClientCerts) {
```
mholt: This line is getting a little long; let's move the conditions to at least two lines (maybe three).

caddytls/config.go
```diff
@@ -550,6 +550,18 @@ func getPreferredDefaultCiphers() []uint16 {
 	return defaultCiphersNonAESNI
 }
 
+func StringSliceEqual(a, b []string) bool {
```
mholt: Unexport this function. Document whether order matters (it does).

shouya:
@mholt: I have adjusted the implementation accordingly and added the test.

caddytls/config.go
```diff
@@ -394,16 +394,9 @@ func assertConfigsCompatible(cfg1, cfg2 *Config) error {
...
 	}
-	if c1.ClientAuth != c2.ClientAuth {
-		return fmt.Errorf("client authentication policy mismatch")
-	}
-	if c1.ClientAuth != tls.NoClientCert && c2.ClientAuth != tls.NoClientCert && c1.ClientCAs != c2.ClientCAs {
-		// Two hosts defined on the same listener are not compatible if they
-		// have ClientAuth enabled, because there's no guarantee beyond the
-		// hostname which config will be used (because SNI only has server name).
-		// To prevent clients from bypassing authentication, require that
-		// ClientAuth be configured in an unambiguous manner.
```
mholt: Can you preserve the comment (and update it and move it as needed)? It's a little confusing what is going on in the changes otherwise.

shouya: Sure, I'll put it back. To be honest I don't quite get what this comment means here.

shouya: Added it back.

mholt: I guess it does a bad job of explaining, but basically: if you have two sites defined in the Caddyfile: `example.com/foo` and `example.com/bar`, and one has ClientAuth but the other does not, the expected behavior is ambiguous since we do not know the path at TLS-handshake-time, thus it is possible that a config not requiring client auth would be chosen even if it is needed, which is bad.

shouya: @mholt: Thanks for the explanation. It's got more clear with your example!

PR #2655, opened 2019-06-28
Document how to run without root priv's on FreeBSD

hartzell:

Add  `dist/init/freebsd/README.md` that describes how to configure the system so that `caddy` can be run without root privileges.

Perhaps this could be added to the main Caddy documentation?

Note that #2652 also adds a `dist/init/freebsd/README.md`, but merging them should be straightforward (the order of the sections is un-important).

## 4. Checklist

### DIFF
--- a/dist/init/freebsd/README.md
+++ b/dist/init/freebsd/README.md
@@ -1,3 +1,55 @@
+# Running caddy without root privileges
+
+FreeBSD systems can use the mac_portacl module to allow access to
+ports below 1024 by specific users (by default, non-root users are not
+able to open ports below 1024).
+
+On a stock FreeBSD system, you need to:
+
+1. Add the following line to `/boot/loader.conf`, which tells the boot
+   loader to load the `mac_portacl` kernel module:
+
+    ``` shell
+    mac_portacl_load="YES"
+    ```
+
+2. Add the following lines to `/etc/sysctl.conf`
+
+    ``` shell
+    net.inet.ip.portrange.reservedlow=0
+    net.inet.ip.portrange.reservedhigh=0
+    security.mac.portacl.port_high=1023
+    security.mac.portacl.suser_exempt=1
+    security.mac.portacl.rules=uid:80:tcp:80,uid:80:tcp:443
+    ```
+
+    The first two lines disable the default restrictions on ports <
+    1023, the third makes the `mac_portacl` system responsible for ports
+    from 0 (the default) up to 1023, and the fourth ensures that the
+    superuser can open *any* port.
+
+    The final/fifth line specifies two rules, separated by a `,`:
+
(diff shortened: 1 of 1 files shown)


mholt:
@hartzell Great, thanks! Would be happy to merge this if you want to take a second and resolve the conflict.

Also, could you help us by running Caddy 2 on FreeBSD? It's in the v2 branch and instructions are in the readme there, and on the wiki...

hartzell:
> [...] Would be happy to merge this if you want to take a second and resolve the conflict.

Done.

In retrospect, the syslog change (the other FreeBSD PR) probably deserves a line in the release notes for the next release, it might surprise someone who's expecting the old behavior. 

> Also, could you help us by running Caddy 2 on FreeBSD? It's in the v2 branch and instructions are in the readme there, and on the wiki...

Happily.  It's already on my things-to-look-at list.

g.

mholt:
K, I'll try to mention it, although it's not directly related to the binary itself, so I don't know why updating the binary would have any effect on existing installations.

> Happily. It's already on my things-to-look-at list.

Awesome! Hope you like it!

hartzell:
> so I don't know why updating the binary would have any effect on existing installations.

Simply installing the new binary wouldn't.

But an automated build system, like [my personal FreeBSD port for Caddy](https://github.com/hartzell/freebsd-ports/blob/master/www/caddy-custom/Makefile#L51-L52), that grabs the init script as part of the install would install the new one and might lead to a bit of head scratching that a simple HEADS UP could avoid.

mholt:
Oof. Well, I think that's a little scary. :) Does your automated system read the release notes too?

hartzell:
The guy who runs it does.  That's why a note there calling attention to the change would be useful.  *Thanks*!

mholt:
I see, so it's _sorta_ automated (you still have to trigger it). Thanks for clarifying. If it was any other way, we'd have to update the compatibility guarantee which I'm not inclined to do right now. I will try to remember to add it to the release notes then.

hartzell:
I figure that the `/etc/rc.conf` variables used to configure the `caddy` startup (assuming one uses the unsupported-but-in-the-tree sample `/etc/rc.d` script) aren't part of Caddy *per-se*, so you have wiggle room.  But, one heads-up might be worth hours of collective head-scratching across the FreeBSD community, so...

Thanks again for the tool and your work!

hartzell:
> Also, could you help us by running Caddy 2 on FreeBSD? It's in the v2 branch and instructions are in the readme there, and on the wiki...

Haven't had time to dig in, but can report that it built and ran the little file browser demo from the README.md without any trouble what-so-ever.  Stood up a FreeBSD-12 server on Digital Ocean, added git and go, cloned, built, curl'ed and browsed.  I'll dig in more.  *Thanks*

Issue #2657, opened 2019-07-07
v2: caddy stop does not stop on windows


tobya:
## 1. Which version of Caddy are you using (`caddy -version`)?
 fdd871e1775d1b6d85de43e8ecf7ddc71a72c503
dev build V2

## 2. What are you trying to do?
Stop Caddy

## 3. What is your Caddyfile?
I dont think its relevant but here it is.

```text
{
  "apps": {
    "http": {
      "servers": {
        "example": {
          "listen": [
            "127.0.0.1:2082",
            "127.0.0.1:2182"

          ],
          "routes": [
            {
              "respond": {
                "responder": "file_server",
                "browse": {}
              }
            }
          ]
        },"example2": {
          "listen": [
            "127.0.0.1:3082",
            "127.0.0.1:3182"

          ],
          "routes": [
            {
              "respond": {
                "responder": "file_server",
                "root" : "d:\\godev\\bin",
                "browse": {}
              }
            }
          ]
        },"example3": {
          "listen": [
            "127.0.0.1:4082",
            "127.0.0.1:4182"

          ],
          "routes": [
            {
              "respond": {
                "responder": "file_server",
                "root" : "d:\\godev\\src",
                "browse": {}
              }
            }
          ]
        }
      }
    }
  }
}
```

## 4. How did you run Caddy (give the full command and describe the execution environment)?
caddy start --config caddyfile.txt

## 5. Please paste any relevant HTTP request(s) here.
NA

## 6. What did you expect to see?
When running 

> caddy stop

the following is shown

> pid=21924                                                                                                                                     > Graceful stop...2019/07/07 17:00:35 taskkill: exit status 1     

Caddy doesnt stop it needs to be stopped manually using /f

taskkill /PID /f 21924 

https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill

if /f is left out caddy wont terminate.

## 7. What did you see instead (give full error messages and/or log)?

caddy stops

## 8. Why is this a bug, and how do you think this should be fixed?

Caddy shoudl stop

## 9. What are you doing to work around the problem in the meantime?

Manually stop

tobya:
I have a PR for this

mholt:
Does /f do a forceful stop though?

Ideally, Caddy would be notified that it is shutting down; i.e. send an "interrupt" (os.SIGINT or in other words, Ctrl+C) so it can shut down gracefully.

tobya:
[...]

PR #2670, opened 2019-07-14
V2: Implement taskkill /f to force quit on windows when caddy stop

tobya:
---
name: Implement taskkill /f to force quit on windows when caddy stop
about: Gracefull shutdown on windows
title: ''
labels: ''
assignees: ''
---

As discussed this is the first iteration of forced shutdown on windows.

The current code wont shutdown caddy on windows.  taskkill /f is required. 

Shutsdown a running instance of caddy by using `taskkill /f` to force shutdown of the process.  Also ensure process is found on windows systems

#2657 

Important to tell windows users that shutdown is not graceful on windows.  Ideally recommend they call stop on admin interface before calling `caddy stop`

I will make a change to 
## 4. Checklist

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -194,7 +194,7 @@ func cmdStop() (int, error) {
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, fmt.Errorf("listing processes: %v", err)
 	}
-	thisProcName := getAppName()
+	thisProcName := getProcessName()
 	var found bool
 	for _, p := range processList {
 		// the process we're looking for should have the same name but different PID
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -27,7 +27,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
-	"path/filepath"
+
 	"strings"
 
 	"github.com/caddyserver/caddy/v2"
@@ -194,11 +194,11 @@ func cmdStop() (int, error) {
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, fmt.Errorf("listing processes: %v", err)
 	}
-	thisProcName := filepath.Base(os.Args[0])
+	thisProcName := getAppName()
 	var found bool
 	for _, p := range processList {
 		// the process we're looking for should have the same name but different PID
-		if matchProcess(p.Executable(), thisProcName) && p.Pid() != os.Getpid() {
+		if p.Executable() == thisProcName && p.Pid() != os.Getpid() {
 			found = true
 			fmt.Printf("pid=%d\n", p.Pid())
 
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -15,285 +15,285 @@
 package caddycmd
 
 import (
-    "bytes"
-    "crypto/rand"
-    "encoding/json"
-    "flag"
-    "fmt"
-    "io"
-    "io/ioutil"
-    "log"
-    "net"
-    "net/http"
-    "os"
-    "os/exec"
-    "path/filepath"
-    "strings"
+	"bytes"
+	"crypto/rand"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
 
-    "github.com/caddyserver/caddy/v2"
-    "github.com/mholt/certmagic"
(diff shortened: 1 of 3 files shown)


cmd/commands.go
```diff
@@ -15,285 +15,285 @@
...
-	"io"
-	"io/ioutil"
-	"log"
-	"net"
-	"net/http"
-	"os"
-	"os/exec"
-	"path/filepath"
-	"strings"
+    "bytes"
```
tobya: I'm not sure why the spaces are differnt here.  Do I need to set some config in go fmt?

mholt: Go fmt uses tabs. No configuration is necessary. Your editor can probably integrate with gofmt and run it on save.

mholt: To clarify, the spaces will have to be converted back into tabs, before this is feasible to review.

tobya: 👍

tobya: @mholt, this is fixed now.

cmd/proc_windows.go
```diff
@@ -21,9 +21,16 @@ import (
...
+	cmd := exec.Command("taskkill", "/pid", strconv.Itoa(pid), "/f")
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("taskkill: %v", err)
 	}
 	return nil
 }
+
+func matchProcess(appname string, processname string) bool {
+	// on Windows the process name will include the extension
+	return appname == processname+".exe"
```
mholt: On Windows, doesn't `os.Args[0]` include the `.exe`?

tobya: This was bothering me also, so 
After a bit of investigation

- `os.Args[0]` returns whatever was typed to launch caddy so

> caddy start

os.Args[0] will return `caddy` but

> caddy.exe start

will return `caddy.exe`

and to further make fun `get_process` in powershell returns `caddy` no matter what and tasklist on cmd window will return `caddy.exe` no matter what.

I think I will expand the windows function to match process name with or without `.exe`.

Also I noticed I have swapped around the appname and processname to make it even more confusing.

mholt: Oh lovely! So I guess we need to standardize the name?

Wait, what if we use `taskkill /im` instead?

> **/im** Specifies the image name of the process to be terminated. Use the wildcard character (\*) to specify all image names.

I don't really know what an "image name" is, but maybe it just means the process name. Could we experiment with that?

tobya: The danger there is that `taskkill /im caddy*` will kill `caddy` `caddy.exe` or even `caddyshack.exe` so that's not really possible.  I'll push another change that should do the trick.

mholt: Well what about `taskkill /im os.Args[0]`?

tobya: no because they are two seperate processes and 1 may have started as `caddy` and the other started as `caddy.exe`

tobya: I've pushed a change now which consistently returns the appname with .exe

tobya: The process list provided by `go-ps` always return process names with `.exe` so this will work consistently to find the process.

cmd/proc_posix.go
```diff
@@ -22,9 +22,14 @@ import (
...
 func gracefullyStopProcess(pid int) error {
+	fmt.Printf("Graceful stop...")
 	err := syscall.Kill(pid, syscall.SIGINT)
 	if err != nil {
 		return fmt.Errorf("kill: %v", err)
 	}
 	return nil
 }
+
+func getAppName() string {
```
mholt: `getProcessName()` will be less ambiguous, since Caddy has "apps" that it runs internally.

cmd/proc_windows.go
```diff
@@ -16,14 +16,29 @@ package caddycmd
...
 	"fmt"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 )
 
 func gracefullyStopProcess(pid int) error {
-	cmd := exec.Command("taskkill", "/pid", strconv.Itoa(pid))
+	fmt.Printf("Stop...")
```
mholt: Maybe clarify that this is a forceful stop: "Forceful stop..."

cmd/proc_posix.go
tobya: Thats fine, I was trying to somehow differentiate between the processes in the process list and the name of the .exe (or process), but I think getProcessName should be fine

tobya:
@mholt Those changes made now

PR #2671, opened 2019-07-14
V2:Add stop command to admin.  

tobya:

Add a stop command to  admin  website.

This gracefully stops all servers and exits the go process.

This enables graceful shutdown on Windows.

Gives the following output on windows 

````
D:\godev\src\github.com\caddyserver\caddy\cmd\caddy>caddy start
2019/07/14 22:50:14 Caddy 2 admin endpoint listening on localhost:2019
Successfully started Caddy

D:\godev\src\github.com\caddyserver\caddy\cmd\caddy>2019/07/14 22:50:20 [INFO][cache:0xc0000a8780] Started certificate maintenance routine
2019/07/14 22:50:24 Shutting down...

2019/07/14 22:50:24 [INFO][cache:0xc0000a8780] Stopped certificate maintenance routine
2019/07/14 22:50:24 Shut down.
````

#2670 
#2657 

Add `stop` command to admin website

## 4. Checklist

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -176,14 +176,14 @@ func handleStop(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	log.Println("[ADMIN] Initiating shutdown")
-	if err := stopandCleanup(); err != nil {
+	if err := stopAndCleanup(); err != nil {
 		log.Printf("[ADMIN][ERROR] stopping: %v \n", err)
 	}
 	log.Println("[ADMIN] Exiting")
 	os.Exit(0)
 }
 
-func stopandCleanup() error {
+func stopAndCleanup() error {
 	if err := Stop(); err != nil {
 		return err
 	}
--- a/admin.go
+++ b/admin.go
@@ -22,13 +22,14 @@ import (
 	"io"
 	"log"
 	"net"
-	"os"
 	"net/http"
 	"net/http/pprof"
+	"os"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/mholt/certmagic"
 	"github.com/rs/cors"
 )
 
@@ -85,7 +86,6 @@ func StartAdmin(initialConfigJSON []byte) error {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/load", handleLoadConfig)
 	mux.HandleFunc("/stop", handleStop)
-
 
 	///// BEGIN PPROF STUFF (TODO: Temporary) /////
 	mux.HandleFunc("/debug/pprof/", pprof.Index)
@@ -171,13 +171,24 @@ func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
 }
 
 func handleStop(w http.ResponseWriter, r *http.Request) {
-	if r.Method == http.MethodPost {
-		log.Println("Shutting down...")
-		Stop()		
-		log.Println("Shut down.")
-		os.Exit(0)
+	if r.Method != http.MethodPost {
+		http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+		return
 	}
-	return
+	log.Println("[ADMIN] Initiating shutdown")
+	if err := stopandCleanup(); err != nil {
+		log.Printf("[ADMIN][ERROR] stopping: %v \n", err)
+	}
+	log.Println("[ADMIN] Exiting")
+	os.Exit(0)
+}
+
+func stopandCleanup() error {
+	if err := Stop(); err != nil {
+		return err
+	}
+	certmagic.CleanUpOwnLocks()
+	return nil
 }
 
 // Load loads and starts a configuration.
--- a/admin.go
+++ b/admin.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"log"
 	"net"
+	"os"
 	"net/http"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tobya** [comment]
I do however get this error when I POST the stop command 

> curl -X POST "http://localhost:2019/stop"

> curl: (52) Empty reply from server

File: admin.go [hunk 1]
```diff
@@ -149,7 +152,7 @@ type AdminRoute struct {
 
 func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
 	r.Close = true
-	if r.Method != "POST" {
+	if r.Method != http.MethodPost {
```
mholt: Good idea, for some reason I never thought to use the constants!

admin.go
```diff
@@ -167,6 +170,16 @@ func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func handleStop(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodPost {
```
mholt: Maybe a little better would be

```go
if r.Method != http.MethodPost {
    http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
    return
}
```

admin.go
```diff
@@ -167,6 +170,16 @@ func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
...
+		log.Println("Shutting down...")
```
mholt: `log.Println("[ADMIN] Initiating shutdown")`

admin.go
```diff
@@ -167,6 +170,16 @@ func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
...
+		Stop()
```
mholt: While currently this function does not return any errors, might as well handle it anyway:

```go
if err := Stop(); err != nil {
    log.Printf("[ADMIN][ERROR] stopping: %v", err)
}
```

admin.go
```diff
@@ -167,6 +170,16 @@ func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func handleStop(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodPost {
+		log.Println("Shutting down...")
+		Stop()		
+		log.Println("Shut down.")
+		os.Exit(0)
```
mholt: This line is the reason for the `empty response from server` message you're getting; the process is quitting before the server can reply to the API client. I suppose that's to be expected.

mholt: Before calling os.Exit(), we absolutely need to be sure to call `certmagic.CleanUpOwnLocks()`. Maybe the steps to a graceful shutdown (it's really only two: Stop() and then that) should be put into their own function...

admin.go
```diff
@@ -167,6 +170,27 @@ func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
...
+func handleStop(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+		return
+	}
+	log.Println("[ADMIN] Initiating shutdown")
+	if err := stopandCleanup(); err != nil {
+		log.Printf("[ADMIN][ERROR] stopping: %v \n", err)
+	}
+	log.Println("[ADMIN] Exiting")
```
tobya: I feel like this is still needed due to odd behaviour on windows in the console.  After exiting the cursor is not shown in normal place (see below)
ScreenHunter_190 16-Jul-19 19 55

I can now type 'dir' and it will work, but it looks odd.  If I leave out Exiting then its unclear if process is finished.  There are a few odd things that happen on the windows console perhaps we can chat about them elsewhere

admin.go
tobya: done

admin.go
tobya: done

admin.go
tobya: ok, do you want to handle anything other than post such as delete?

admin.go
tobya: I put these two in their own function

tobya:
Thank you @mholt for your patience as ever.

admin.go
```diff
@@ -167,6 +170,27 @@ func handleLoadConfig(w http.ResponseWriter, r *http.Request) {
...
+	os.Exit(0)
+}
+
+func stopandCleanup() error {
```
mholt: `stopAndCleanup()`

And perhaps this should also be called by `gracefulStop()` here: https://sourcegraph.com/github.com/caddyserver/caddy@b44a22a9d45443aa736a04495dfe716b62b58adc/-/blob/sigtrap.go#L61-68 (we'd want to preserve the logic but reuse the code which calls `Stop()` then `CleanUpOwnLocks()`) - in order to reduce code duplication.

admin.go
mholt: That's just because Windows is writing stderr to the same place that stdout and stdin are displayed; happens on Linux too. Pretty normal. (In a production setting, the output would be redirected to a log, which Caddy 2 can't do quite yet.)

admin.go
tobya: done

admin.go
mholt: Nah, POST is fine.

PR #2674, opened 2019-07-16
v2: Performance improvements to Replacer implementation (placeholders)

mholt:
## 1. What does this change do, exactly?

Improves performance of placeholders in Caddy 2. Placeholders are the equivalent of variables in other web servers (but better). They allow you to refer to dynamic values in static configuration.

I was lazy with the initial implementation because I wanted to get it done quickly.

Note that this optimization affects ALL usage of placeholders, but the tests I performed were using them in a response body.

Using this sample config:

```
{
	"apps": {
		"http": {
			"servers": {
				"myserver": {
					"listen": [":8080"],
					"routes": [
						{
							"handle": [{
								"handler": "rewrite",
								"method": "FOO",
								"uri": "/test/abc?foo=bar"
							},
							{
								"handler": "static_response",
								"body": "Method: {http.request.method} URI: '{http.request.uri}' Path: '{http.request.uri.path}' Query: '{http.request.uri.query}'"
							}]
						}
					]
				}
			}
		}
	}
}
```

Here are some high-level load tests on my 8-core Macbook Pro:

```
NO PLACEHOLDERS, BASELINE:

$ wrk -t12 -c400 -d10s http://127.0.0.1:8080
Running 10s test @ http://127.0.0.1:8080
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    10.25ms   49.25ms 992.92ms   98.28%
    Req/Sec     6.67k     1.49k   26.95k    85.37%
  798668 requests in 10.10s, 99.78MB read
  Socket errors: connect 0, read 248, write 0, timeout 0
Requests/sec:  79074.88
Transfer/sec:      9.88MB



PLACEHOLDERS, BEFORE OPTIMIZATION:

$ wrk -t12 -c400 -d10s http://127.0.0.1:8080
Running 10s test @ http://127.0.0.1:8080
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    18.18ms   63.94ms   1.02s    97.73%
    Req/Sec     3.11k   512.10    11.11k    92.02%
  372368 requests in 10.10s, 46.52MB read
  Socket errors: connect 0, read 286, write 0, timeout 0
Requests/sec:  36860.00
Transfer/sec:      4.60MB



PLACEHOLDERS, AFTER OPTIMIZATION:

$ wrk -t12 -c400 -d10s http://127.0.0.1:8080
Running 10s test @ http://127.0.0.1:8080
[...]

Issue #2673, opened 2019-07-16
v2: Improve placeholder performance


mholt:
Placeholders in v2 have a much simpler implementation than in v1, but their performance is not great on busy servers.

We should avoid allocating strings unless necessary. In other words, instead of building a full map of all possible replacements and calling `strings.ReplaceAll()`, we should only obtain string values for placeholders that are used. I suspect this will help speed things up quite a bit.

aligator:
I would like to contribute to caddy. Do you think this is a good starting point?

mholt:
@aligator I would say so, since it would get you familiar with several parts of the v2 code base. I would rate this task as moderate. It requires doing some performance testing before starting, and probably some profiling. Then any changes will need to be tested again with the same tests to ensure they are faster.

aligator:
Ok, great. Do you already have something for performance tests?

mholt:
Yes, but my results will be different on my machine than they are on yours. So you'll need to set up your own tests (you can use wrk, for example, to do some high-level load testing) and see how they compare on your own system.

First, here's a simple config I used to get a baseline:

```
{
	"apps": {
		"http": {
			"servers": {
				"myserver": {
					"listen": [":8080"],
					"routes": [
						{
							"handle": [{
								"handler": "static_response",
								"body": "Method: GET URI: '/' Path: '/' Query: ''"
							}]
						}
					]
				}
			}
		}
	}
}
```

I started Caddy with that config then used wrk:

```
$ wrk -t12 -c400 -d10s http://127.0.0.1:8080
```

(all the caveats of performance testing apply here; for example make sure your machine is in a consistent state every time, such as plugged in to power, not running energy-intensive apps, etc. etc.)

Then I used an equivalent config with several placeholders:

```
{
	"apps": {
		"http": {
			"servers": {
				"myserver": {
					"listen": [":8080"],
					"routes": [
						{
							"handle": [{
								"handler": "static_response",
								"body": "Method: {http.request.method} URI: '{http.request.uri}' Path: '{http.request.uri.path}' Query: '{http.request.uri.query}'"
							}]
						}
					]
				}
			}
		}
	}
}
```

And did the same load test again.

To ascertain the specific reason for the performance drop, some profiling is needed. pprof can help with that: 

- https://golang.org/pkg/net/http/pprof/
- https://blog.golang.org/profiling-go-programs

aligator:
Thank you, I'll look into it.

mholt:
Great!

[...]

PR #2675, opened 2019-07-16
V2: WIP: Tests for Replacer

aligator:
---
name: Tests for Replacer
about: Testing the Replacer
---
This PR adds tests for the Replacer.

Currently there are tests for Set and Delete. (Both use ReplaceAll to check if they work, but I also want to write an extra test for it.)

Question: Should I test only through the Interface or the replacer type directly?

https://github.com/caddyserver/caddy/issues/2673
https://github.com/caddyserver/caddy/pull/2674

--
## 4. Checklist

### DIFF
--- a/replacer.go
+++ b/replacer.go
@@ -25,7 +25,7 @@ import (
 type Replacer interface {
 	Set(variable, value string)
 	Delete(variable string)
-	Map(func() map[string]string)
+	Map(ReplacementFunc)
 	ReplaceAll(input, empty string) string
 }
 
@@ -34,23 +34,22 @@ func NewReplacer() Replacer {
 	rep := &replacer{
 		static: make(map[string]string),
 	}
-	rep.providers = []ReplacementsFunc{
-		defaultReplacements,
-		func() map[string]string { return rep.static },
+	rep.providers = []ReplacementFunc{
+		globalDefaultReplacements,
+		rep.fromStatic,
 	}
 	return rep
 }
 
 type replacer struct {
-	providers []ReplacementsFunc
+	providers []ReplacementFunc
 	static    map[string]string
 }
 
-// Map augments the map of replacements with those returned
-// by the given replacements function. The function is only
-// executed at replace-time.
-func (r *replacer) Map(replacements func() map[string]string) {
-	r.providers = append(r.providers, replacements)
+// Map adds mapFunc to the list of value providers.
+// mapFunc will be executed only at replace-time.
+func (r *replacer) Map(mapFunc ReplacementFunc) {
+	r.providers = append(r.providers, mapFunc)
 }
 
 // Set sets a custom variable to a static value.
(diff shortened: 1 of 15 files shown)


aligator:
Just noticed that I based the test on the old implementation. Updated it now to the new one (just used string(phOpen) instead of phOpen) and... 
the tests fail:
` replacer_test.go:81: Expectd 'val1123098765öö_äüspace valuetest-1231.2.3.4EMPTY{MyNotSetVariable}', got 'val1{asdf}098765{23456789}space value{1}1.2.3.4{testEmpty}{testEmpty}{MyNotSetVariable}' for input '{test1}{asdf}{äöü}{23456789}{with space}{1}{mySuper_IP}{testEmpty}{MyNotSetVariable}'`

Seems that it always missed one, because my test string has no spaces between the } and {. This worked with your previous implementation.

aligator:
Ok, thank you for your suggestions. I will update my tests accordingly.

I just added one for something like {l{test1}... 
This test fails with my current fixes in my fix-branch.

mholt:
> I just added one for something like {l{test1}...
> This test fails with my current fixes in my fix-branch.

Excellent. :) That doesn't surprise me. You can either propose a fix in this PR or I can commit a fix. I'll be happy with either option.

aligator:
@mholt could you please review this? I rewrote the tests and added new ones. Also I included two little fixes for the replacer logic.

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// test if all keys are still there (by length)
+	length := len(rep.static)
+	if len(rep.static) != 7 {
+		t.Errorf("Expectd length '%v' got '%v'", 7, length)
+	}
+}
+
+func TestReplacerReplaceAll(t *testing.T) {
+	rep := fakeReplacerFilled()
```
mholt: This is the only place the function is called; just inline it so it's easier to read.

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+	actual := rep.ReplaceAll(testInput, "EMPTY")
+
+	// test if all are replaced as expected
+	if actual != expected {
+		t.Errorf("Expectd '%s' got '%s' for '%s'", expected, actual, testInput)
+	}
+}
+
+func TestReplacerDelete(t *testing.T) {
+	rep := fakeReplacerFilledStatic()
```
mholt: Again, just inline the value here; no need for a separate function. Less indirection.

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+		rep.Delete(key)
+
+		// test if key is removed from static map
+		if _, ok := rep.static[key]; ok {
+			t.Errorf("Expectd '%s' to be removed. It is still in static map.", key)
+		}
+	}
+
+	// check if static slice is smaller
+	expected := len(fakeReplacerFilledStatic().static) - len(toDel)
```
mholt: Get the start length before delete operations to avoid needing to call the whole function again

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+	toDel := []string{
+		"key2", "key4",
+	}
+
+	for _, key := range toDel {
+		rep.Delete(key)
+
+		// test if key is removed from static map
+		if _, ok := rep.static[key]; ok {
+			t.Errorf("Expectd '%s' to be removed. It is still in static map.", key)
```
mholt: Do a global search+replace to fix spelling: `Expected`

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+				return tc.value, true
+			}
+			return "NO", false
+		})
+
+		// test if function (which listens on specific key) is added bychecking length
+		if len(rep.providers) == i+1 {
+			val, _ := rep.providers[i](tc.key) // never fails, as we just checked the length
+			// check if the last function is the one we just added
+			if val != tc.value {
```
mholt: A better way to do this is to just compare the pointers, i.e. `rep.providers[i] == testCaseProviderFunc` (where testCaseProviderFunc is the function you passed into `Map()`).

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+			// test with non existing vars
+			testInput: "{test1} {nope} {1} ",
+			expected:  "val1 {nope} test-123 ",
+		},
+	} {
+		actual := rep.ReplaceAll(tc.testInput, "EMPTY")
+
+		// test if all are replaced as expected
+		if actual != tc.expected {
+			t.Errorf("Expectd '%s' got '%s' for '%s'", tc.expected, actual, tc.testInput)
```
mholt: Add the test case number (`i`) so it's easy to find in the list: `t.Errorf("Test %d: Expected ...", i, ...)`

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+package caddy
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+func fakeReplacer() replacer {
```
mholt: Move helper functions to bottom; prefer tests at top of file

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+		},
+		{
+			variable: "testEmpty",
+			value:    "",
+		},
+	} {
+		rep.Set(tc.variable, tc.value)
+
+		// test if key is added
+		if val, ok := rep.static[tc.variable]; ok {
```
mholt: This might be better as two loops through the test cases:

1. Call `Set()` and do nothing else
2. Then check each variable and make sure the value is correct.

This way we can guarantee that values aren't disappearing and aren't being overwritten either when they shouldn't be.

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+		if _, ok := rep.static[key]; ok {
+			t.Errorf("Expectd '%s' to be removed. It is still in static map.", key)
+		}
+	}
+
+	// check if static slice is smaller
+	expected := len(fakeReplacerFilledStatic().static) - len(toDel)
+	actual := len(rep.static)
+	if len(rep.static) != expected {
+		t.Errorf("Expectd length '%v' got lenth '%v'", expected, actual)
```
mholt: `length`

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+func TestReplacerReplaceAllDefaults(t *testing.T) {
+	rep := NewReplacer()
+	testInput := "{system.hostname} {system.slash} {system.os} {system.arch}"
+	expected := getHostnameForTest() + " " + string(filepath.Separator) + " " + runtime.GOOS + " " + runtime.GOARCH
+
+	// test env.
+	testInput += " {env.GOPATH}"
```
mholt: What's the purpose of this exactly?

replacer.go
```diff
@@ -115,12 +115,13 @@ func (r *replacer) ReplaceAll(input, empty string) string {
 			}
 		}
 		if !found {
+			lastWriteCursor = i
```
mholt: Just wondering, which test case does this fix?

replacer.go
```diff
@@ -115,12 +115,13 @@ func (r *replacer) ReplaceAll(input, empty string) string {
...
 			continue
 		}
 
 		// advance cursor to end of placeholder
-		i = end + 1
-		lastWriteCursor = i
+		i = end
+		lastWriteCursor = i + 1
```
mholt: Just wondering, which test case does this fix?

replacer_test.go
```diff
@@ -0,0 +1,275 @@
...
+			variable: "test1",
+			value:    "val1",
+		},
+		{
+			variable: "asdf",
+			value:    "123",
+		},
+		{
+			variable: "äöü",
+			value:    "öö_äü",
```
mholt: Good thinking to test unicode

replacer.go
aligator: If you uncomment this fix, the ReplaceAll test fails:
`    replacer_test.go:155: Expected '{teval1{as{{dftest-123' got '{teval1{as{as{{as{{dftest-123' for '{te{test1}{as{{df{1}'
    replacer_test.go:155: Expected 'val1 {nope} test-123 ' got 'val1  {nope} test-123 ' for '{test1} {nope} {1} '`
--> it doesn't count correctly if an invalid placeholder is found. This results for example in the three "{as" and the additional empty space before "{nope}"

replacer.go
aligator: This fixes placeholders which are directly behind each other:
`replacer_test.go:155: Expected 'val1123öö_äütest-123space value1.2.3.4' got 'val1{asdf}öö_äü{1}space value{mySuper_IP}' for '{test1}{asdf}{äöü}{1}{with space}{mySuper_IP}'`

and it results in each second placeholder to be ignored (if there is no char between the placeholders) because `i` is set to one position too late and this results in one char to be ignored after a matching placeholder.

replacer_test.go
aligator: done

replacer_test.go
aligator: done

replacer_test.go
aligator: done

replacer_test.go
aligator: done

replacer_test.go
aligator: It doesn't work how you suggested it. I now use Sprintf to get the pointer value like it is suggested here: https://stackoverflow.com/a/34901677

replacer_test.go
aligator: done

replacer_test.go
aligator: testing if env. can be used. I test now with a predefined environment variable instead of GOPATH and moved all tests for the default replacements into the new-test because New should return replacer with a provider which replaces these default replacements. (now without using ReplaceAll for testing instead calling the first provider directly)

replacer_test.go
mholt: Ah, hmm. TIL. Good idea. That'll do.

aligator:
Yes, you can merge this.

mholt:
Thanks for your contribution! It's nice to have that under some testing.

PR #2682, opened 2019-07-19
Used goimports to fix import order

muesli:
Keeps the list sorted and prevents future merge conflicts.


muesli:
No problem, thanks for merging!

Issue #2681, opened 2019-07-18
v2: Optimize with compiler's inliner


mholt:
Go over the Caddy 2 code base at some point and see where we can take advantage of inlining for performance improvements. See: https://blog.filippo.io/efficient-go-apis-with-the-inliner/

dominikbraun:
The most basic approach to do this is to find all functions that return a reference - since these functions can be optimized if the referenced value has been allocated inside it.

So I've written a [small tool called refreturn](https://github.com/dominikbraun/refreturn) that finds functions returning a reference. For the Caddy v2 branch, the output is:

```sh
$ refreturn .
caddy.go:211:1: GoModule
modules/caddyhttp/encode/encode.go:95:1: openResponseWriter
modules/caddyhttp/fileserver/browse.go:135:1: browseWriteJSON
modules/caddyhttp/fileserver/browse.go:141:1: browseWriteHTML
modules/caddyhttp/fileserver/staticfiles.go:178:1: openFile
modules/caddyhttp/responsewriter.go:39:1: Hijack
modules/caddyhttp/responsewriter.go:156:1: Buffer
modules/caddyhttp/reverseproxy/healthchecker.go:80:1: NewHealthCheckWorker
modules/caddyhttp/reverseproxy/upstream.go:283:1: roundRobin
modules/caddyhttp/reverseproxy/upstream.go:305:1: random
modules/caddyhttp/reverseproxy/upstream.go:355:1: newUpstream
modules/caddyhttp/reverseproxy/upstream.go:430:1: newReverseProxy
modules/caddyhttp/reverseproxy/reverseproxy.go:116:1: NewSingleHostReverseProxy
modules/caddyhttp/reverseproxy/reverseproxy.go:196:1: ServeHTTP
modules/caddyhttp/staticresp_test.go:60:1: fakeRequest
modules/caddytls/connpolicy.go:37:1: TLSConfig
```

I've optimized `openResponseWriter()` for example and the escape analysis shows that it seems to work quite well. Before the optimization:

```
modules/caddyhttp/encode/encode.go:102:3: &responseWriter literal escapes to heap
```

And after (the output above disappeared):

```
modules/caddyhttp/encode/encode.go:84:29: inlining call to (*Encode).openResponseWriter
```

However, we may discuss which function calls should be optimized this way. For instance, `browseWriteJSON` calls `new(bytes.Buffer)` and I'm not sure if this is a intentional heap allocation or if it can be optimized away.

[...]

PR #2687, opened 2019-07-22
Part 1: Optimize using compiler's inliner

dominikbraun:
Allocating a value and returning a reference to it in a single function causes the Go compiler to escape the value to the heap, since it has to outlive the function's lifetime. Optimizing the function calls (as shown in the first commit) can enable mid-stack inlining of the function and avoiding a heap allocation.

https://github.com/caddyserver/caddy/issues/2681

This is an internal change.

## 4. Checklist

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -15,6 +15,7 @@
 package fileserver
 
 import (
+	"bytes"
 	"fmt"
 	"html/template"
 	weakrand "math/rand"
@@ -25,6 +26,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
@@ -46,6 +48,8 @@ type FileServer struct {
 	Hide       []string `json:"hide,omitempty"`
 	IndexNames []string `json:"index_names,omitempty"`
 	Browse     *Browse  `json:"browse,omitempty"`
+
+	bufPools map[string]*sync.Pool
 	// TODO: Content negotiation
 }
 
@@ -53,6 +57,19 @@ type FileServer struct {
 func (fsrv *FileServer) Provision(ctx caddy.Context) error {
 	if fsrv.IndexNames == nil {
 		fsrv.IndexNames = defaultIndexNames
+	}
+
+	fsrv.bufPools = map[string]*sync.Pool{
+		"json": &sync.Pool{
+			New: func() interface{} {
+				return new(bytes.Buffer)
+			},
+		},
+		"html": &sync.Pool{
+			New: func() interface{} {
+				return new(bytes.Buffer)
+			},
+		},
 	}
 
 	if fsrv.Browse != nil {
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -49,7 +49,6 @@ type FileServer struct {
 	IndexNames []string `json:"index_names,omitempty"`
 	Browse     *Browse  `json:"browse,omitempty"`
 
-	bufPools map[string]*sync.Pool
 	// TODO: Content negotiation
 }
 
@@ -57,19 +56,6 @@ type FileServer struct {
 func (fsrv *FileServer) Provision(ctx caddy.Context) error {
 	if fsrv.IndexNames == nil {
 		fsrv.IndexNames = defaultIndexNames
-	}
-
-	fsrv.bufPools = map[string]*sync.Pool{
-		"json": &sync.Pool{
-			New: func() interface{} {
-				return new(bytes.Buffer)
-			},
-		},
-		"html": &sync.Pool{
-			New: func() interface{} {
-				return new(bytes.Buffer)
-			},
-		},
 	}
 
 	if fsrv.Browse != nil {
@@ -317,6 +303,12 @@ func calculateEtag(d os.FileInfo) string {
 }
 
 var defaultIndexNames = []string{"index.html"}
+
+var bufPool = sync.Pool{
+	New: func() interface{} {
+		return new(bytes.Buffer)
+	},
+}
 
 const minBackoff, maxBackoff = 2, 5
 
--- a/caddy.go
+++ b/caddy.go
@@ -209,7 +209,15 @@ func (d *Duration) UnmarshalJSON(b []byte) error {
 // value will still be returned, but with an
 // unknown version.
 func GoModule() *debug.Module {
-	mod := &debug.Module{Version: "unknown"}
+	var mod debug.Module
+	return goModule(&mod)
+}
+
+// goModule holds the actual implementation of GoModule.
+// Allocating debug.Module in GoModule() and passing a
+// reference to goModule enables mid-stack inlining.
+func goModule(mod *debug.Module) *debug.Module {
+	mod.Version = "unknown"
 	bi, ok := debug.ReadBuildInfo()
 	if ok {
 		mod.Path = bi.Main.Path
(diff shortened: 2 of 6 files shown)


mholt:
Awesome, this LGTM so far. Great start. The only thing on my wishlist would be a simple `Bench` function in a test file to see what the speedup is like, but I don't suppose we have to do that for _every_ change. But I would like to justify all optimizations with data if possible. Thanks for working on this!! How exciting.

mholt:
Just noticed your latest commits 👍 looking great! It is a little too bad you have to keep the old implementation in the test file. I wonder if it would be OK to nix those and just report the difference in your pull request or commit messages. What do you think?

(Run tests with `-test.benchmem` to get allocation reports.)

dominikbraun:
I also didn't like that the previous implementations were copied into the test files (but that seemed to be the simplest approach and I've got some comparisons now).

As expected, the benefit of the optimizations is largely dependent on the method. `BrowseWriteJSON` gets a slight but consistent speedup, even though the number of allocations stays the same:

```
BenchmarkOldBrowseWriteJSON-4   	   10000	    122096 ns/op	   25370 B/op	     303 allocs/op
BenchmarkNewBrowseWriteJSON-4   	   10000	    105656 ns/op	   25370 B/op	     303 allocs/op
``` 

In contrast, `BrowseWriteHTML` has been slowed down.

```
BenchmarkOldBrowseWriteHTML-4   	 3000000	       486 ns/op	     272 B/op	       5 allocs/op
BenchmarkNewBrowseWriteHTML-4   	 3000000	       492 ns/op	     272 B/op	       5 allocs/op
```

For `openResponseWriter` there is a better speedup including a 'dropped' allocation.

```
BenchmarkOldOpenResponseWriter-4   	 5000000	       270 ns/op	     128 B/op	       3 allocs/op
BenchmarkNewOpenResponseWriter-4   	10000000	       200 ns/op	      64 B/op	       2 allocs/op
```

It's up to you how significant a speedup has to be to justify these code changes (I guess this is dependent on how often the respective function will be called).

I'm going to do some other comparisons and see how the shown allocations may be avoided.

mholt:
Ah yeah, the `openResponseWriter` improvement is definitely significant. (There's only 3 allocations to begin with, but reducing 3->2 is a ~33% improvement! and at half the number of bytes/op.)

I'm guessing that the HTML and JSON ones don't improve because, compared to encoding JSON or executing HTML templates, this one allocation is miniscule. Although I bet I should be pooling those buffers, I doubt inlining those will be very beneficial, at least until then. (Using a pool would be easy; `sync.Pool` makes it pretty straightforward; I do this in other packages including the encode one you already inlined.)

In general, any function in the HTTP request handler chain is a candidate for inlining, especially when there isn't much other work being done on the value that would be slow in comparison.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -46,6 +48,8 @@ type FileServer struct {
 	Hide       []string `json:"hide,omitempty"`
 	IndexNames []string `json:"index_names,omitempty"`
 	Browse     *Browse  `json:"browse,omitempty"`
+
+	bufPools map[string]*sync.Pool
```
mholt: This should actually be a global variable, we don't want to be making new pools just because configuration has changed.

And I think we can simply make one pool, unless you suspect the JSON and HTML outputs will have very different sizes (I doubt they will be very different -- at least, within 1 order of magnitude; i.e. JSON wouldn't be KBs while HTML is MBs I think).

dominikbraun: Yap, I took this from your `Encode` example.

> But if you think otherwise and/or have evidence that different pools perform better, I could be convinced.

I guess this depends on how widely the pool sizes differ from each other. If you think that the JSON and HTML outputs will have similar sizes, a global buffer would be perfectly fine for me.

mholt: Okay, in that case let's just use one global pool like here: https://sourcegraph.com/github.com/caddyserver/caddy@c8b0a97b1c2524f7486aef38b2cf78a93d7b98b0/-/blob/modules/caddyhttp/templates/tplcontext.go#L275-279

dominikbraun:
Using `sync.Pool` for the JSON and HTML buffers works just fine. The version with `new(bytes.Buffer)`:

```
BenchmarkBrowseWriteJSON-4   	   10000	    108608 ns/op	   25448 B/op	     303 allocs/op
BenchmarkBrowseWriteHTML-4   	 3000000	       524 ns/op	     347 B/op	       5 allocs/op
```

And the pooled version:

```
BenchmarkBrowseWriteJSON-4   	   10000	    104678 ns/op	   49519 B/op	     301 allocs/op
BenchmarkBrowseWriteHTML-4   	 3000000	       413 ns/op	     224 B/op	       4 allocs/op
```

Note that this reduces the number of allocations by 20% for `BrowseWriteHTML`, even with 75.000 iterations.

mholt:
Cool, I think that is expected. The HTML writer is just building a string, whereas the JSON writer is using the encoding/json package, which uses its own buffer pools already as it marshals things piece-by-piece. So I expect the gains for the HTML writer to be greater.

Thanks for that improvement!

dominikbraun:
The escape analysis shows that the `FileServer`'s `openFile()` method seems to get inlined out of the box:

```
modules/caddyhttp/fileserver/staticfiles.go:182:22: inlining call to os.Open
```

Correspondingly the ns/op were reduced only a little bit after optimizing the method.

```
BenchmarkOpenFile-4   	  300000	      3993 ns/op	     682 B/op	      13 allocs/op
```

After:

```
BenchmarkOpenFile-4   	  500000	      3908 ns/op	     682 B/op	      13 allocs/op
```

mholt:
@dominikbraun Sure, we can probably leave it if it's already being inlined, right? I didn't realize it could be inlined with such complex error handling there... https://sourcegraph.com/github.com/caddyserver/caddy@c8b0a97b1c2524f7486aef38b2cf78a93d7b98b0/-/blob/modules/caddyhttp/fileserver/staticfiles.go#L179-193

Maybe the compiler is smart enough to inline error handling too.

mholt:
@dominikbraun Do you want to do this piecewise (a little at a time) or in one (bigger) PR? If you won't be able to get around to more until later, how about just resolving the merge conflict and I can merge in this change.

dominikbraun:
First of all I'm sorry that I wasn't able to optimize `ServeHTTP` yet (been busy the last days).

> Sure, we can probably leave it if it's already being inlined, right?

Yes, I would suggest to leave it as it is.

> how about just resolving the merge conflict and I can merge in what you've got so far.

I'll continue my tests today, but you may merge the commits just as you wish - I'm fine with both options. :smile:

mholt:
@dominikbraun 

> First of all I'm sorry that I wasn't able to optimize ServeHTTP yet (been busy the last days).

No problem! It happens. That's why I am offering to review and merge multiple smaller PRs so you can invest smaller time chunks and still yield real progress as you go.

> I'll continue my tests today, but you may merge the commits just as you wish - I'm fine with both options. 😄

Okie. Go ahead and submit what you'd like today and when you're ready for this PR to be merged, just say so -- and we can always do another one later!

dominikbraun:
The call to ``ServeHTTP` could been inlined, however there is hardly any improvement in performance since this method is pretty large.

I'd suggest merging the current changes and I'll spend some time trying to optimize `ServeHTTP` as well as the request handling chain.

mholt:
@dominikbraun Thanks, will do! Can you just resolve the merge conflict first?

doodeehee70:
ซ้ำซ้ำของ #

mholt:
Excellent, thanks so much @dominikbraun! Your contributions are welcome any time.

dominikbraun:
Many thanks for your patience and support so far! 👍

PR #2688, opened 2019-07-24
Fix: unix build in v2

Dreamacro:

Add some missing import to fix build on my macOS

None

None

## 4. Checklist

### DIFF
--- a/cmd/proc_posix.go
+++ b/cmd/proc_posix.go
@@ -18,6 +18,8 @@ package caddycmd
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"syscall"
 )
 


mholt:
Huh, thanks!

/cc @tobya - FYI

tobya:
hmmm.  Do we need to set up CI builds on azure for this branch.  Should they not be automatic?

mholt:
We do, although I haven't been able to get it to work yet. I just wanted to let you know so that you can make sure the program compiles before you push. :+1:

tobya:
I found this https://github.com/mitchellh/gox which should make it easier to check cross compiles

PR #2712, opened 2019-08-17
Fixing a compilation error

agbaraka:
name: Pull request
title: 'Fixing a compilation error'
labels: ''
assignees: ''
---

Small bug fix during compilation.

./caddy.go:230:12: cannot use *dep (type debug.Module) as type *debug.Module in return argument
./caddy.go:233:12: cannot use bi.Main (type debug.Module) as type *debug.Module in return argument

## 4. Checklist


mholt:
Thanks, but this will conflict with #2699 which already resolve this... 🤔I hope to get that merged in next week. If not, I'll merge this instead.

agbaraka:
No problem.

Issue #2719, opened 2019-08-22
Incorrect Zone parsing in the Caddyfile


jgsqware:
## 1. Which version of Caddy are you using (`caddy -version`)?
Latest

## 2. What are you trying to do?

Hello, when in my Corefile, I have a server block no space between the end of the name and the `{`, the server block is wrongly interpreted.

## 3. What is your Caddyfile?

Corefile who is working:
```
www.mydomain.com:1053 {
  cache 30
  forward . 8.8.8.8
}
```

CoreDNS logs:

```
# /coredns -conf ./c/Corefile
www.mydomain.com.:1053
2019-08-21T10:44:02.806Z [INFO] plugin/reload: Running configuration MD5 = 6740acee942add7a7ad7a726502e6ab4
2019-08-21T10:44:02.806Z [INFO] CoreDNS-1.6.2
2019-08-21T10:44:02.806Z [INFO] linux/amd64, go1.12.7, 25632d6a
CoreDNS-1.6.2
linux/amd64, go1.12.7, 25632d6a
^C[INFO] SIGINT: Shutting down
```

Corefile who's not working:
```
www.mydomain.com:1053{
  cache 30
  forward . 8.8.8.8
}
```

CoreDNS logs:

```
# /coredns -conf ./c/Corefile
www.mydomain.com:1053{.:53
2019-08-21T10:44:02.806Z [INFO] plugin/reload: Running configuration MD5 = 6740acee942add7a7ad7a726502e6ab4
2019-08-21T10:44:02.806Z [INFO] CoreDNS-1.6.2
2019-08-21T10:44:02.806Z [INFO] linux/amd64, go1.12.7, 25632d6a
CoreDNS-1.6.2
linux/amd64, go1.12.7, 25632d6a
^C[INFO] SIGINT: Shutting down
```

## 4. How did you run Caddy (give the full command and describe the execution environment)?
It use CoreDNS. But related to Caddyfile Parser

## 6. What did you expect to see?
If i use block separator `{` at the end of a line, I would expect a space before the zone, if not I would raise a parsing error.

## 7. What did you see instead (give full error messages and/or log)?

Corefile:
```
www.mydomain.com:1053{
  cache 30
  forward . 8.8.8.8
}
```

CoreDNS logs:

```
# /coredns -conf ./c/Corefile
www.mydomain.com:1053{.:53
[...]

PR #2720, opened 2019-08-22
Modifying Lexer to remove tailing { if not quoted

jgsqware:
When a address ends with a `{` and is not quoted, it discard the `{` in the address name.

This is meant to avoid this 

```
localhost:1234{
    dir1 foo bar
    dir2
}
```
parsed as `localhost:1234` and not `localhost:1234{`

but 

```
"localhost:1234{"
    dir1 foo bar
    dir2
```
parsed as `localhost:1234{`

https://github.com/caddyserver/caddy/issues/2719

## 4. Checklist

### DIFF
--- a/caddyfile/parse_test.go
+++ b/caddyfile/parse_test.go
@@ -217,8 +217,8 @@ func TestParseOneAndImport(t *testing.T) {
 		{`localhost:1234{
 		    dir1 foo bar
 		    dir2
-		  }`, false, []string{
-			"localhost:1234",
+		  }`, true, []string{
+			"localhost:1234{",
 		}, map[string]int{
 			"dir1": 3,
 			"dir2": 1,
--- a/caddyfile/parse_test.go
+++ b/caddyfile/parse_test.go
@@ -214,6 +214,36 @@ func TestParseOneAndImport(t *testing.T) {
 			"dir3": 1,
 		}},
 
+		{`localhost:1234{
+		    dir1 foo bar
+		    dir2
+		  }`, false, []string{
+			"localhost:1234",
+		}, map[string]int{
+			"dir1": 3,
+			"dir2": 1,
+		}},
+
+		{`"localhost:1234"{
+		    dir1 foo bar
+		    dir2
+		  }`, false, []string{
+			"localhost:1234",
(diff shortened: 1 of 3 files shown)


jgsqware:
Yes sure, I didn't implemented it like this as currently there is no error from the lexer itself.
I can return an error on the lexer. but this error must bubble up. 
 wdyt?

mholt:
It should lex just fine; I think this needs to be a parse error. The parser will either see the closing curly brace and have no opening one to match it, or it will see the opening one at the end of another token, which is an error.

jgsqware:
@mholt Now, it detect in directives when the `}` that ends the block has no `{` in the block server and error

mholt:
Sorry I've fallen behind on this review, will get to it soon!

caddyfile/parse_test.go
```diff
@@ -214,6 +214,36 @@ func TestParseOneAndImport(t *testing.T) {
...
+		    dir1 foo bar
+		    dir2
+		  }`, true, []string{
+			"localhost:1234{",
+		}, map[string]int{
+			"dir1": 3,
+			"dir2": 1,
+		}},
+
+		{`"localhost:1234"{
```
mholt: Do you think this should be an error? 🤔 It looks untidy...

jgsqware: You mean because of the missing space?
I assume here is intended with the double quote. but we can make this case invalid too.

mholt: > You mean because of the missing space?

Correct.

I think technically this is one token... but, I guess, now that we talk about it, I have never really thought about this before. If you have an input like `"a" b` then that is clearly 2 tokens, where the first token is interpreted literally. If you have an input like `"a"b`... should that be one or two?

I'm inclined to say that the lexer _should_ lex it as 1 token, where the quoted part is literal and the b is still part of the same token (similar to bash shell, I believe).

So, I think it should be treated the same as the case for `localhost:1234{`.

If that is kind of complex or gnarly, though, maybe we need to patch the lexer on that matter instead.

mholt:
Any interest in completing this?

Issue #2680, opened 2019-07-18
v2: TLS client authentication


mholt:
Implement TLS client auth into Caddy 2.

Caddy 2 provides advanced config facilities to enable highly-specific TLS client auth controls. We just need to take advantage of it.

- [x] Whitelist CA certs
- [x] Whitelist each leaf cert
- [ ] 'ask' endpoint

alexandrestein:
Hi,

What is the status of this?

I think this should goes into the **http/servers/tls_connection_policies**.
The name of the field can be something link **client_CAs**.
It would take the form of **DER encoded certificate authorities** themselves encoded into **base64** to be compatible with **JSON**.

 **http/servers/tls_connection_policies**:
```json
{
    "match": {},
    "alpn": [],
    "cipher_suites": [],
    "certificate_selection": {},
    "client_CAs": [
        "MIIDSzCCAjOgAwI......E/lYx0qGtr0xHQ=="
    ]
}
```

If this value is provided the verification is required as it is with **Caddy 1**.

Encoding certificate into this form is trivial:
```bash
openssl x509 -inform PEM -outform DER -in ca.crt | base64
```

If this is looking OK for you I can start working on it.
I can't try version 2 for the moment because I need TLS client authentication.

mholt:
Cool, the only thing left to do after #2731 is the ask_url.

If we want the "asking" to be done solely via TCP/TLS, and not even use HTTPS, maybe we just need an `ask_address` or `ask_endpoint` instead of a full URL.

mholt:
Will revisit this later if there's demand for it.

zakcutner:
@mholt Is there a way to use this from a Caddyfile, similar to v1's `clients` subdirective for `tls`? If not, are there any plans to add this back in the future? Thanks!

mholt:
Probably around ~v2.1.

zakcutner:
Cool, I'll use the JSON for now in that case — thanks for such a quick response 😇

sahib:
> Cool, the only thing left to do after #2731 is the ask_url. [...]
> Will revisit this later if there's demand for it.

Just commenting here to indicate I have a use-case for this: I have several services that require a client certificate to be accessed.
They check the serial of the cert and do some internal database lookups to determine if it is valid. Only if valid I would forward the request to the actual service. For this an ``ask_url`` would be very useful. This would allow me to put those services behind caddy (with its great certificate management) and not expose them directly like I have to do right now.

[...]

PR #2731, opened 2019-08-28
Add support for client TLS authentication Caddy V2

alexandrestein:
---
name: Add support for client TLS authentication Caddy V2
about: Adding support for client TLS authentication for the next Caddy release
title: 'Add support for client TLS authentication :smile:'
labels: 'V2'
assignees: '@alexandrestein'
---

The thing is pretty obvious and it was available in **Caddy 1**.

[An issue](https://github.com/caddyserver/caddy/issues/2680) is open about this.

I'm really new to the project and I don't know if this is the best way to this.
To test this update I used a similar configuration:
```json
{
    "apps": {
        "http": {
            "http_port": 80,
            "https_port": 443,
            "grace_period": "2s",
            "servers": {
                "mainServices": {
                    "listen": [
                        ":443"
                    ],
                    "tls_connection_policies": [
                        {
                            "match": {
                                "sni": [
                                    "0.dom.com",
                                    "dom.com"
                                ]
                            },
                            "client_cas": [
                                "MIIDSzCCAjOgAwIBAgIUfIRObjWNUA4jxQ/0x8BOCvE2Vw4wDQYJKoZIhvcNAQELBQAwFjEUMBIGA1UEAwwLRWFzeS1SU0EgQ0EwHhcNMTkwODI4MTYyNTU5WhcNMjkwODI1MTYyNTU5WjAWMRQwEgYDVQQDDAtFYXN5LVJTQSBDQTCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAK5m5elxhQfMp/3aVJ4JnpN9PUSz6LlP6LePAPFU7gqohVVFVtDkChJAG3FNkNQNlieVTja/bgH9IcC6oKbROwdY1h0MvNV8AHHigvl03WuJD8g2ReVFXXwsnrPmKXCFzQyMI6TYk3m2gYrXsZOU1GLnfMRC3KAMRgE2F45twOs9hqG169YJ6mM2eQjzjCHWI6S2/iUYvYxRkCOlYUbLsMD/AhgAf1plzg6LPqNxtdlwxZnA0ytgkmhK67HtzJu0+ovUCsMv0RwcMhsEo9T8nyFAGt9XLZ63X5WpBCTUApaAUhnG0XnerjmUWb6eUWw4zev54sEfY5F3x002iQaW6cECAwEAAaOBkDCBjTAdBgNVHQ4EFgQU4CBUbZsS2GaNIkGRz/cBsD5ivjswUQYDVR0jBEowSIAU4CBUbZsS2GaNIkGRz/cBsD5ivjuhGqQYMBYxFDASBgNVBAMMC0Vhc3ktUlNBIENBghR8hE5uNY1QDiPFD/THwE4K8TZXDjAMBgNVHRMEBTADAQH/MAsGA1UdDwQEAwIBBjANBgkqhkiG9w0BAQsFAAOCAQEAKB3V4HIzoiO/Ch6WMj9bLJ2FGbpkMrcb/Eq01hT5zcfKD66lVS1MlK+cRL446Z2b2KDP1oFyVs+qmrmtdwrWgD+nfe2sBmmIHo9m9KygMkEOfG3MghGTEcS+0cTKEcoHYWYyOqQh6jnedXY8Cdm4GM1hAc9MiL3/sqV8YCVSLNnkoNysmr06/rZ0MCUZPGUtRmfd0heWhrfzAKw2HLgX+RAmpOE2MZqWcjvqKGyaRiaZks4nJkP6521aC2Lgp0HhCz1j8/uQ5ldoDszCnu/iro0NAsNtudTMD+YoLQxLqdleIh6CW+illc2VdXwj7mn6J04yns9jfE2jRjW/yTLFuQ=="
                            ]
                        }
                    ],
                    "routes": [
                        {
                            "match": [
                                {
                                    "host": [
                                        "0.dom.com",
                                        "dom.com"
                                    ]
                                }
                            ],
                            "handle": [
                                {
                                    "handler": "file_server",
                                    "browse": {}
                                }
                            ]
                        },
                        {
                            "match": [
                                {
                                    "host": [
                                        "1.dom.com"
                                    ]
                                }
                            ],
                            "handle": [
                                {
                                    "handler": "file_server",
                                    "browse": {}
                                }
                            ]
                        }
                    ]
                }
            }
        }
    },
    "tls": {
        "management": {
            "module": "acme",
            "ca": "https://acme-staging-v02.api.letsencrypt.org/directory",
            "email": "astein58@gmail.com",
            "on_demand": false
        }
    }
}
```
[Here](https://github.com/caddyserver/caddy/files/3552078/caddyTestingClient.zip) is a client certificate you can use to try. The password is "caddy".
And [here](https://github.com/caddyserver/caddy/files/3552081/ca.zip) the CA certificate.

I didn't notice any side effect and it was asking for certificate.

Let me know what change would you like to see. :smile: 

The part related to the **tls_connection_policies** in the Wiki V2.

## 4. Checklist

**No but I'm ready to do it. I don't see where you are managing the CI**

### DIFF
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -75,7 +75,12 @@ func (app *App) Provision(ctx caddy.Context) error {
 			srv.AutoHTTPS = new(AutoHTTPSConfig)
 		}
 
-		if len(srv.TLSConnPolicies) != 0 {
+		// disallow TLS client auth bypass which could
+		// otherwise be exploited by sending an unprotected
+		// SNI value during TLS handshake, then a protected
+		// Host header during HTTP request later on that
+		// connection
+		if srv.hasTLSClientAuth() {
 			srv.StrictSNIHost = true
 		}
 
@@ -163,8 +168,7 @@ func (app *App) Start() error {
 					return fmt.Errorf("%s: listening on %s: %v", network, addr, err)
 				}
 
-				// enable HTTP/2 (and support for solving the
-				// TLS-ALPN ACME challenge) by default
+				// enable HTTP/2 by default
 				for _, pol := range srv.TLSConnPolicies {
 					if len(pol.ALPN) == 0 {
 						pol.ALPN = append(pol.ALPN, defaultALPN...)
@@ -298,19 +302,11 @@ func (app *App) automaticHTTPS() error {
 				return fmt.Errorf("%s: managing certificate for %s: %s", srvName, domains, err)
 			}
 
-			// tell the server to use TLS by specifying a TLS
-			// connection policy (which supports HTTP/2 and the
-			// TLS-ALPN ACME challenge as well)
+			// tell the server to use TLS if it is not already doing so
 			if srv.TLSConnPolicies == nil {
-				// build a new TLSConnPolicies if it does not exist
 				srv.TLSConnPolicies = caddytls.ConnectionPolicies{
 					&caddytls.ConnectionPolicy{ALPN: defaultALPN},
 				}
-			} else {
-				// append the default policy at the end of the existing policies
-				srv.TLSConnPolicies = append(srv.TLSConnPolicies,
-					&caddytls.ConnectionPolicy{ALPN: defaultALPN},
-				)
 			}
 
 			if srv.AutoHTTPS.DisableRedir {
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -75,6 +75,10 @@ func (app *App) Provision(ctx caddy.Context) error {
 			srv.AutoHTTPS = new(AutoHTTPSConfig)
 		}
 
+		if len(srv.TLSConnPolicies) != 0 {
+			srv.StrictSNIHost = true
+		}
+
 		// TODO: Test this function to ensure these replacements are performed
 		for i := range srv.Listen {
 			srv.Listen[i] = repl.ReplaceAll(srv.Listen[i], "")
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -297,8 +297,16 @@ func (app *App) automaticHTTPS() error {
 			// tell the server to use TLS by specifying a TLS
 			// connection policy (which supports HTTP/2 and the
 			// TLS-ALPN ACME challenge as well)
-			srv.TLSConnPolicies = caddytls.ConnectionPolicies{
-				{ALPN: defaultALPN},
+			if srv.TLSConnPolicies == nil {
+				// build a new TLSConnPolicies if it does not exist
+				srv.TLSConnPolicies = caddytls.ConnectionPolicies{
+					&caddytls.ConnectionPolicy{ALPN: defaultALPN},
+				}
+			} else {
+				// append the default policy at the end of the existing policies
+				srv.TLSConnPolicies = append(srv.TLSConnPolicies,
+					&caddytls.ConnectionPolicy{ALPN: defaultALPN},
+				)
 			}
 
 			if srv.AutoHTTPS.DisableRedir {
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -17,6 +17,7 @@ package caddytls
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -117,7 +118,7 @@ type ConnectionPolicy struct {
 	ProtocolMin  string   `json:"protocol_min,omitempty"`
 	ProtocolMax  string   `json:"protocol_max,omitempty"`
 
-	// TODO: Client auth
+	ClientCAs []string `json:"client_cas,omitempty"`
 
 	matchers     []ConnectionMatcher
 	certSelector certmagic.CertificateSelector
@@ -156,6 +157,38 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 
 	// session tickets support
 	cfg.SessionTicketsDisabled = tlsApp.SessionTickets.Disabled
+
+	// test client authentication is required
+	if len(p.ClientCAs) != 0 {
+		// build client cert pool, use it and set client certificates as
+		// "RequireAndVerify"
+		cliCertPool := x509.NewCertPool()
+		cfg.ClientCAs = cliCertPool
+		cfg.ClientAuth = tls.RequireAndVerifyClientCert
+
+		// parse and add certificate to client cert pool
+		for _, clientCAString := range p.ClientCAs {
+			// decode base64
+			derBytes, err := base64.StdEncoding.DecodeString(clientCAString)
+			if err != nil {
+				continue
+			}
+			if len(derBytes) == 0 {
+				continue
+			}
+
+			// parse the DER encoded certificate
+			var clientCA *x509.Certificate
+			clientCA, err = x509.ParseCertificate(derBytes)
+			if err != nil {
+				continue
+			}
+
+			// add the certificate to cliCertPool
+			cliCertPool.AddCert(clientCA)
+		}
+
+	}
 
 	// session ticket key rotation
 	tlsApp.SessionTickets.register(cfg)
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -112,12 +112,11 @@ type ConnectionPolicy struct {
 	Matchers      map[string]json.RawMessage `json:"match,omitempty"`
 	CertSelection json.RawMessage            `json:"certificate_selection,omitempty"`
 
-	CipherSuites []string `json:"cipher_suites,omitempty"`
-	Curves       []string `json:"curves,omitempty"`
-	ALPN         []string `json:"alpn,omitempty"`
-	ProtocolMin  string   `json:"protocol_min,omitempty"`
-	ProtocolMax  string   `json:"protocol_max,omitempty"`
-
+	CipherSuites         []string              `json:"cipher_suites,omitempty"`
+	Curves               []string              `json:"curves,omitempty"`
+	ALPN                 []string              `json:"alpn,omitempty"`
+	ProtocolMin          string                `json:"protocol_min,omitempty"`
+	ProtocolMax          string                `json:"protocol_max,omitempty"`
 	ClientAuthentication *ClientAuthentication `json:"client_authentication,omitempty"`
 
 	matchers     []ConnectionMatcher
@@ -213,8 +212,9 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 		return fmt.Errorf("protocol min (%x) cannot be greater than protocol max (%x)", p.ProtocolMin, p.ProtocolMax)
 	}
 
-	// test client authentication is required
 	if p.ClientAuthentication != nil {
+		cfg.ClientAuth = tls.RequireAnyClientCert
+
 		parseCertFromBase64Func := func(certStr string) (*x509.Certificate, error) {
 			// decode base64
 			derBytes, err := base64.StdEncoding.DecodeString(certStr)
@@ -231,7 +231,6 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 			// build client cert pool, use it and set client certificates as
 			// "RequireAndVerify"
 			cfg.ClientCAs = cliCertPool
-			cfg.ClientAuth = tls.RequireAndVerifyClientCert
 
 			// parse and add certificate to client cert pool
 			for _, clientCAString := range p.ClientAuthentication.TrustedCACerts {
@@ -244,14 +243,45 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 				// add the certificate to cliCertPool
 				cliCertPool.AddCert(clientCA)
 			}
+
+			verifyPeerCertificateFunc := func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+				if rawCerts == nil {
+					return fmt.Errorf("no certificate provided")
+				}
+
+				interPool := x509.NewCertPool()
+				var cert *x509.Certificate
+
+				for i, rawCert := range rawCerts {
+					remoteCert, err := x509.ParseCertificate(rawCert)
+					if err != nil {
+						return fmt.Errorf("can't parse the given certificate: %s", err.Error())
+					}
+
+					// don't add the leaf certificate to the pool but save it for verification
+					if i < len(rawCerts)-1 {
+						interPool.AddCert(remoteCert)
+					} else {
+						cert = remoteCert
+					}
+				}
+
+				_, err := cert.Verify(x509.VerifyOptions{
+					Roots:         cliCertPool,
+					Intermediates: interPool,
+					KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+				})
+				if err == nil {
+					return nil
+				}
+
+				return fmt.Errorf("can't validate remote certificate with trusted_ca_certs")
+			}
+			cfg.VerifyPeerCertificate = verifyPeerCertificateFunc
 		}
 
 		if len(p.ClientAuthentication.TrustedLeafCerts) != 0 {
 			trustedLeafCerts := []*x509.Certificate{}
-			// ask for client certificate
-			if cfg.ClientAuth < tls.RequireAnyClientCert {
-				cfg.ClientAuth = tls.RequireAnyClientCert
-			}
 
 			// parse given leafs to check
 			for _, clientCertString := range p.ClientAuthentication.TrustedLeafCerts {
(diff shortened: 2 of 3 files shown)


alexandrestein:
> One crucial thing we'll need to do is force-enable `strict_sni_host` (field `StrictSNIHost`) on the associated caddyhttp.Server instance, otherwise client auth can be bypassed.

I tried to bypass the authentication but I did not manage to do it (I did not spend much time on it).
Does the modification I made should solve this bypass?

alexandrestein:
I implemented a struct similar to what you exposed in the issue. (this is a much better approach)

Leafs and CA verification should be working fine.

Tell me what you think about this.

If you like it :crossed_fingers:
Please explain me what you expect in the "ask_url" field.
And what should be done with it?

francislavoie:
I think what @mholt has in mind for "ask_url" is the idea of allowing a backend service to determine if the client certificate is valid, instead of just having a whitelist configured in Caddy. I.e. send a GET request with a specific payload to some internal service, say for example `http://127.0.0.1:8888/check-cert`, which should reply with 200 if the certificate is acceptable, some 4xx status otherwise if it should be denied.

See the ask option here: https://caddyserver.com/docs/tls which was used to determine whether a certain domain should be allowed to be dynamically issued a certificate from whatever ACME backend is configured. I think the "ask_url" field is similar to that but for client certs. Allow some backend service to implement the logic dynamically based on the cert input rather than just a whitelist (there could be all kinds of business reasons to do it this way).

mholt:
@alexandrestein 

> I tried to bypass the authentication but I did not manage to do it (I did not spend much time on it). Does the modification I made should solve this bypass?

That's the right idea, it's just in the wrong place. I left a review comment in-line. I guarantee that without that protection, it can be bypassed. Simply connect using an unprotected SNI value and then send an HTTP requests with that connection with a protected hostname in the Host header. See https://github.com/caddyserver/caddy/commit/4d9ee000c8d2cbcdd8284007c1e0f2da7bc3c7c3 for details. There's actually 2 other ways to do it. The third one in that list may also be relevant here (different client auth configs), could you look into that if you have a chance? I will also review it later.

> Please explain me what you expect in the "ask_url" field. And what should be done with it?

@francislavoie hit the nail on the head. I figure Caddy could make a request to the "ask URL" (or maybe it should just be a TLS connection on the given host and port? do we even need HTTPS here?) using the given client certificate and if that connection succeeds, then Caddy will allow the client cert.

> Leafs and CA verification should be working fine.

Would you have a chance to verify that? Either perform some tests or write tests (preferred -- though I admit I haven't gotten around to adding many tests yet myself).

Thank you by the way!! This is looking great.

modules/caddytls/connpolicy.go
```diff
@@ -212,7 +213,83 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 		return fmt.Errorf("protocol min (%x) cannot be greater than protocol max (%x)", p.ProtocolMin, p.ProtocolMax)
 	}
 
-	// TODO: client auth, and other fields
+	// test client authentication is required
```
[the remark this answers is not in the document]
alexandrestein: I removed it

alexandrestein:
@francislavoie thank.
I totally agree. I was more asking for the way to do it and @mholt has pretty much replied to it.

About testing the bypass I did not have any unsecured virtual host.
I tried similar things but only with protected domains.
I will try to reproduce the bypass and then check if the fixe works properly.

I verified :smile: most of the case by hand.
It works but I didn't try with a chain of certificate for example.
But I tried with one option and with multiple options. 
It looks OK.

Let's close this one and I will try to do the ask option after this one.

Do you have any starting point for the testing mechanism I should use or I need to start it from scratch?

mholt:
@alexandrestein Sounds good. If you want to add a test function (which I would thank you very much for doing, haha), I would test `buildStandardTLSConfig`. Just use https://golang.org/pkg/testing/ for tests. Something simple is fine: just test the part of `buildStandardTLSConfig` that sets up the Client Auth. Make sure that the resulting struct has the necessary, proper values in its fields.

Then we can wrap up this PR!

alexandrestein:
Awesome :smile: 

Ok I will try to do this tomorrow.
It's evening here. :v:

mholt:
@alexandrestein If you'd like to keep contributing to Caddy after this, then let me know and I can invite you to our development Slack. I just need to know which email to invite.

alexandrestein:
@mholt what does this mean? 🥺
I can't guarantee that I will always has time to spend on the project. 😔

But I would be glad to be part of it. 😊

modules/caddytls/connpolicy.go
```diff
@@ -221,6 +298,13 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 	return nil
 }
 
+// ClientAuthentication defines client authentication solutions
```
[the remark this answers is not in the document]
alexandrestein: I used base64 because it can be easily embedded directly into JSON.
I don't think file name would be easier to manage.

alexandrestein:
I confirme that the bypass is fixed. (hand tested with **cURL**)
I manage to get a protected domain through a unprotected one when the `StrictSNIHost` enforcement was not active.
When active, the server reply by an error. (I will need to make a test for this)
Great!!!

I started to write some tests to verify that all pieces are working fine together. (not only the part I worked on)
For this I start a Caddy server and I inject some configurations.

Then I try to reach the server as a regular `HTTP` client.

I did not manage to make it works.
It looks like `tls.certificates.load_files` doesn't even look for the files.
I'm not sure if this is a bug or if I'm missing something.

I don't know if I will have much time for this next weeks and unlikely this weekend neither.
But if you could take a look why I can't connect to the server with `HTTPS` I think I could make a "_base function_" to test **Caddy** as close as possible from real conditions (load various configurations and test if all parts of the project behave as expected  together)

In first I would like to test only the TLS authentication just to close this PR.
And after I will try to build a reusable test suite for the project.

Have a good day and good week-end. :v: :sunglasses:

connpolicy_test.go
```diff
@@ -0,0 +1,177 @@
+package caddy_test
```
mholt: These can be white-box tests, so `package caddyhttp` is fine.

connpolicy_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"bytes"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	// this is where modules get plugged in
```
mholt: You shouldn't need to plug in any modules for the tests

connpolicy_test.go
```diff
@@ -0,0 +1,177 @@
...
...jAAMBEGCWCGSAGG+EIBAQQEAwIGQDArBglghkgBhvhCAQ0EHhYcR2VuZXJhdGVkIFNlcnZlciBDZXJ0aWZpY2F0ZTAdBgNVHQ4EFgQUnQSqjMYpErYOmO5U4fxbBl8oAkUwgbIGA1UdIwSBqjCBp4AUHTg2XTOhyEBEkwSs9A6bkevOfoOheaR3MHUxCzAJBgNVBAYTAkdCMRcwFQYDVQQIDA5Vbml0ZWQgS2luZ2RvbTEPMA0GA1UEBwwGTG9uZG9uMRIwEAYDVQQKDAlBQ01FIEluYy4xGTAXBgNVBAsMEEFudmlsIERlcGFydG1lbnQxDTALBgNVBAMMBHJvb3SCFGOtyOOeL8byY03yYIqyKW1SQMKXMA4GA1UdDwEB/wQEAwIFoDATBgNVHSUEDDAKBggrBgEFBQcDATAyBgNVHR8EKzApMCegJaAjhiFodHRwczovLzEyNy4wLjAuMTo4MjAwL3YxL3BraS9jcmwwDQYJKoZIhvcNAQELBQADggIBACXuuAVzNuqdVt1irPHnWcDOkwJoFAIeVmI0/BZPaMPEZJc/vd3/66+sRJ5pX7j8Bl6CCMT6RZauQk7engiAfgbHmRt0Cngo+0K4KChcbsG5RWsjB7fXtI/Xup8D0pfNwDUqPgQt50P0Gz1MOgFPEDtb0OcIiKIEVGEa+noHMQkspAKnINzgROApjljGcwgpSenU7Y9r4NkcZKsZkHrCp5HL3cZCJAI9LwUAE40gNAo10lfQrzgVzeCj1mpfghrYVGU9AO4wcyad7zBHbb0ZhudP39RsdfjotngiB1ANsa0lfuKXZ2cXBZFsv3JZOfRQz02UyPqg3X1yaxktYuaxc/KejeceTu73wkM/ir3topUg616KC7YIqtUq2Qu3n0vhBSwY+fzEudbfxp2nuqefqmnYuENwkNxdiXuqqWQ5k8ISBgECVc8VY27441wU7lTJFz/CMpU7Z9jx5Wf3YlMALHRVyyzdWJGBW1/wXxaGvBRzaEDAl4VqsRX1mij/4xAcToTQhCzSzKE/MGokMLpB+HrWCyqG+55wcwVBcyA7oFYs3f7a7BynPuOkV64rNO7ouaxbXyhpsMmE0kjvEIJD7XGyLuOJa6lThlEokLK/StXv620PPQ5zrGGHG6R4IQquToXxcdz4h5pdpCaQ7LTGG6BBm4bRNlcWobIX1DzLX+vn"
```
mholt: Prefix these with `test` so that we can make sure it's clear they're only used in tests (even though it's in a test file, it's just good form I feel like) -- for example, `testRootCA`.

connpolicy_test.go
```diff
@@ -0,0 +1,177 @@
...
+	rootCertPool = x509.NewCertPool()
+	caBuff, _ := base64.StdEncoding.DecodeString(rootCA)
+	cert, _ := x509.ParseCertificate(caBuff)
+
+	rootCertPool.AddCert(cert)
+
+	caddy.StartAdmin([]byte(`{}`))
+}
+
+func loadConf(input []byte) error {
```
mholt: Also unnecessary; remove this

connpolicy_test.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+func init() {
```
mholt: None of this is necessary, the test file does not need and should not have an init function.

connpolicy_test.go
```diff
@@ -0,0 +1,177 @@
...
+				{
+					"certificate": "./server.crt",
+					"key": "./server.key",
+					"format": "pem",
+					"tags": []
+				}
+			]
+		}
+	}
+}`)
```
mholt: None of this is necessary either.

Just make a TLSConnPolicy that has the values you want to test.

Going through all these other parts of the software is bad form in unit testing.

connpolicy_test.go
```diff
@@ -0,0 +1,177 @@
...
+				"`+t+`": [
+					"`+v+`"
+				]
+			}
+		}
+	],`)...)
+	return append(ret, end...)
+}
+
+func TestTLSAuth(t *testing.T) {
```
mholt: Really, this function should be all that is needed. Hard-code a TLS connection policy value and call the relevant method(s) on it.

modules/caddytls/connpolicy.go
```diff
@@ -221,6 +333,18 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
...
+type ClientAuthentication struct {
+	// TrustedCACerts contains a list of CA to verify the peer certificates.
+	// TrustedLeafCerts contains a list of accepted certificates.
+	// For both certificates are base64 DER encoded.
+	TrustedCACerts   []string `json:"trusted_ca_certs,omitempty"`
+	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
+
+	// TODO
+	AskURL string `json:"ask_url,omitempty"`
```
mholt: Remove for now, until it gets implemented. We can always add it later.

server.crt
```diff
@@ -0,0 +1,34 @@
+-----BEGIN CERTIFICATE-----
```
mholt: Remove these two files; or, if _absolutely_ needed for tests, put them into a `testdata` folder. However, I don't think they are needed. If so, maybe put them into a const like you did with the others.

server.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
mholt: Remove this file too, as above

alexandrestein:
> Thanks! It seems you may be new to writing tests. Please review my comments and simplify the tests _a lot_, there's way too much here for effective unit tests. The tests can and should be much simpler. Once these changes are made, we can see about merging it in.
> 
> Thank you for working on this!

Well you get it wrong sorry.
This is not supposed to be a final test suite for the part I worked on. It was to have tip on my error or to show you a reproducible bug. Sorry it wasn't clear.

It's just a **WIP** which faced a bug. (one purpose of testing in fact)
I made all the handmade tests with the automatic **TLS**. To test in any environment we can't use this method. So we need to load some static certificates. This is where I'm missing something or the software is buggy.

The way you would like to have the test would be probably more complex (implementing caddy.Context is not obvious to me) to be a relevant testing. I think you forgot that this needs a working server to be fully tested. I made a custom verification process (which is called only during handshake) to be able to chain all the possible authentications. We are talking about authentication and security feature consumers need to rely on.
To test this properly we **NEED TO DO** the authentication which require a **TLS** server. 
Why no use the software to test itself? At least for parts which require a full stack to be fully/correctly tested.

I can revert to commit [commit #ba44e88](https://github.com/caddyserver/caddy/pull/2731/commits/ba44e88a45d78c1ed429de72b0b1576ae228d1a3) and apply the rest of the comments to close this PR.
And later you make your self happy with the test you want.

Does this is OK for you?
Otherwise close the PR and leave it for the next.

alexandrestein:
> @francislavoie hit the nail on the head. I figure Caddy could make a request to the "ask URL" (or maybe it should just be a TLS connection on the given host and port? do we even need HTTPS here?) using the given client certificate and if that connection succeeds, then Caddy will allow the client cert. That is optional for this PR.

I just rethink about this. After all I don't think you could do this.
You will need the private key of the given certificate. Which is not send by the client :wink: 

I did not test it but if the server does not check the private key ownership the all authentication is probably is not that secure.

It's maybe not the place to say it but we talk about this in this conversation.

mholt:
@alexandrestein 

> I just rethink about this. After all I don't think you could do this.
You will need the private key of the given certificate. Which is not send by the client 😉

That's not the case, actually. The standard library will verify the client's certificate (meaning, the client must prove that they possess the private key) and then call the `VerifyPeerCertificate` callback. In other words, Caddy (via the standard lib) will negotiate the TLS handshake and then can ask this custom endpoint whether the given public key is allowable. Here are the godocs:

```go
        // VerifyPeerCertificate, if not nil, is called after normal
        // certificate verification by either a TLS client or server. It
        // receives the raw ASN.1 certificates provided by the peer and also
        // any verified chains that normal processing found. If it returns a
        // non-nil error, the handshake is aborted and that error results.
        //
        // If normal verification fails then the handshake will abort before
        // considering this callback. If normal verification is disabled by
        // setting InsecureSkipVerify, or (for a server) when ClientAuth is
        // RequestClientCert or RequireAnyClientCert, then this callback will
        // be considered but the verifiedChains argument will always be nil.
```

> I did not test it but if the server does not check the private key ownership the all authentication is probably is not that secure.

That is definitely correct (assuming the server wishes to perform authentication of the client, which is likely the case), but fortunately the standard lib will do that check for us.

mholt:
@alexandrestein Oops, I only saw your second reply, let me reply to the first now.

> Well you get it wrong sorry.
This is not supposed to be a final test suite for the part I worked on. It was to have tip on my error or to show you a reproducible bug. Sorry it wasn't clear.

Ah, I'm sorry I misunderstood then.

If it's okay with you, let's remove the tests from this PR and consider a more comprehensive test suite as part of a new project in a different issue + series of PRs. (I intend to go back and improve the test coverage quite a bit anyway.)

In PRs that add features, I prefer _unit tests_ which are really very simple tests that directly access the function that is being tested, with as minimal external setup as possible. Unit tests do not need to (and shouldn't) test the whole server; that is definitely a separate project.

> I think you forgot that this needs a working server to be fully tested. ... To test this properly we NEED TO DO the authentication which require a TLS server.

This isn't true, and I'll show you what I mean when I get the chance. Basically, you've written little functions [like this one](https://github.com/caddyserver/caddy/pull/2731/files#diff-506f14bb805212d0680508b152f0a607R297-R322) which take inputs (`rawCerts [][]byte, verifiedChains [][]*x509.Certificate`) and produce an output (`error`). All we need to do is call that function with some bytes and maybe a chain of x509.Certificates and ensure it produces the right output (error value, or lack thereof). That's a unit test.

The rest of your change, like setting the VerifyPeerCertificate field in the tls.Config, is a separate unit test.

In other words, there's no need to test the whole system end-to-end, just to make sure client auth works. That's a larger project that will require some careful consideration and discussion in an issue before a PR.

But we can revisit that later! (We're still in the early stages after all.)

Let's get this PR merged. :) If you want to remove the tests (and the crt and key files), we can revisit tests soon in a new PR.

alexandrestein:
I think it's OK.

Sorry for all the noise, I hope it will helps.

modules/caddyhttp/caddyhttp.go
```diff
@@ -297,8 +301,16 @@ func (app *App) automaticHTTPS() error {
...
+			if srv.TLSConnPolicies == nil {
+				// build a new TLSConnPolicies if it does not exist
+				srv.TLSConnPolicies = caddytls.ConnectionPolicies{
+					&caddytls.ConnectionPolicy{ALPN: defaultALPN},
+				}
+			} else {
+				// append the default policy at the end of the existing policies
+				srv.TLSConnPolicies = append(srv.TLSConnPolicies,
+					&caddytls.ConnectionPolicy{ALPN: defaultALPN},
+				)
```
mholt: @alexandrestein Just a quick question -- does appending the default policy have any particular utility? i.e. is there something that breaks if you don't do this?

alexandrestein:
Thanks to you @mholt for the package, your hard work and a very good knowledge of the all mechanisms you put into this.

It was a pleasure :smile: 
If I got some time I hope to help again. :v:

Issue #2461, opened 2019-02-09
Caddy startup output is incorrect when hosting HTTP on port 443


whitestrake:
### 1. What version of Caddy are you using (`caddy -version`)?

Caddy 0.11.3 (non-commercial use only)

### 2. What are you trying to do?

Host a website on port 443.

### 3. What is your entire Caddyfile?

N/A - easily reproducible without a Caddyfile

### 4. How did you run Caddy (give the full command and describe the execution environment)?

`caddy -port 443`

### 5. Please paste any relevant HTTP request(s) here.

`curl https://localhost` - fails, naturally, because we haven't enabled TLS

`curl http://localhost:443` - succeeds as expected

### 6. What did you expect to see?

```text
Activating privacy features... done.
http://:443
```

### 7. What did you see instead (give full error messages and/or log)?

```text
Activating privacy features... done.
https://
```

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?

See point 4.

---

Pretty simple issue. Startup output says `https://` if you're hosting a site on port 443, regardless of whether it's actually HTTPS or not...

tobya:
Is a solution to this simply to disallow hosting a http site on port 443?

It seems to me that anyone wishing to do this is just trying to cause themselves trouble?

mholt:
These are good questions... I kind of designed Caddy to fight against unconventional choices which are "only confusing or problematic" but I can't foresee all _valid_ uses either.

Are there good reasons to serve HTTP on port 443? Or HTTPS on port 80? I feel like that defies the port's IANA reservations.

mholt:
@whitestrake What is your use case? Why do you need HTTP on port 443?

magikstm:
> 
> 
> Is a solution to this simply to disallow hosting a http site on port 443?
> 
> It seems to me that anyone wishing to do this is just trying to cause themselves trouble?
> 
> Would simply raising an error and exiting be a valid fix?

I think it would be a good fix. As @mholt noted, it defies the IANA reservations.

Ref: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=443

PR #2755, opened 2019-09-17
Disallow port 443 for http sites

anarchyrucks:
---
name: Pull request
title: 'Disallow port 443 for http sites'
---

Disallow port 443 for http sites.

This PR is based on the discussion at #2461 

Need help here. New to caddy project.

## 4. Checklist

- [X] This change has comments explaining package types, values, functions, and non-obvious lines of code

I'm not sure if this is the correct place to put this guard code. Please let me know and I can make changes accordingly.

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -50,6 +50,11 @@ func init() {
 	caddy.RegisterServerType(serverType, caddy.ServerType{
 		Directives: func() []string { return directives },
 		DefaultInput: func() caddy.Input {
+			if Port == DefaultHTTPSPort && serverType == "http" {
+				fmt.Println("Port 443 can only be used for https sites.")
+				os.Exit(1)
+			}
+
 			if Port == DefaultPort && Host != "" {
 				// by leaving the port blank in this case we give auto HTTPS
 				// a chance to set the port to 443 for us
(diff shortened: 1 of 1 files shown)


mholt:
The "http" server type serves HTTP applications, but still can use the HTTPS protocol; this fix is not correct, sorry. The linked issue needs fixing only in the output that is printed to the screen. However, Caddy 1's days are numbered, as we'll be moving to Caddy 2 soon. Would be happy if you wanted to jump into contributing to Caddy 2, but please discuss your changes before spending much time on them, Thanks!

anarchyrucks:
It was mentioned at the end of #2461 discussion that serving HTTP on 443 defies the IANA reservations. I was under the assumption that raising an error and exiting would be the valid fix.

Thanks for your time and I'd love to contribute to Caddy 2.

caddyhttp/httpserver/plugin.go
```diff
@@ -50,6 +50,11 @@ func init() {
 	caddy.RegisterServerType(serverType, caddy.ServerType{
 		Directives: func() []string { return directives },
 		DefaultInput: func() caddy.Input {
+			if Port == DefaultHTTPSPort && serverType == "http" {
```
mholt: >  I was under the assumption that raising an error and exiting would be the valid fix.

But this is not how to raise the error. I think reading the godoc to learn about Caddy's architecture would help. :) But again, Caddy 2 is around the corner, so Caddy 1 is becoming less relevant.

PR #2756, opened 2019-09-17
tls/acme: Ability to customize trusted roots for ACME servers

### LINKED ISSUE #2702: Specify root certificate(s) to use for ACME

It's currently possible to tell Caddy to connect to a specific ACME server using the `tls.ca` directive, like this:

```
foo.internal {
  root /var/run/www
  tls {
    ca https://acme.internal/acme/directory
  }
}
```

but there doesn't appear to be any way to specify which root certificate(s) are trusted when Caddy connects to the specified server. I would like this to be configurable.

This is useful if you're connecting to an ACME server whose certificate is not issued by a CA in your system's default trust store. This is likely to be the case for an internal ACME server that's connected to an internal CA (in which the ACME server's certificate is likely issued by the internal CA, not by a public CA). An internal CA is useful for internal service-to-service mTLS and in development & other pre-prod environments.

More specifically, I'm working on [an open source certificate authority](https://github.com/smallstep/certificates). We're adding ACME support and would like to support issuing internal certificates to Caddy. We ran into this issue during testing with Caddy.

This is *only* necessary if the relevant root certificate is not already trusted by default (i.e., its certificate isn't in the default system trust store). For example, this isn't necessary for Let's Encrypt's staging server because that server uses a certificate issued by (non-staging) Let's Encrypt, which is trusted by default. However, an internal ACME server is likely to use a certificate issued by an internal CA. In that case, the internal CA's root certificate likely isn't in the system's trust store and won't be trusted by Caddy by default.

It's possible to work around this limitation by adding your internal root certificate to your system's default trust store. But this is a global operation: it affects most TLS communication on a machine. There are lots of situations where this is not desirable.

mholt:

Adds the ability to customize trusted root CAs for ACME servers. Makes integrations with things like Smallstep CA a lot easier and more secure.

/cc @mmalone

Closes #2702

Add `trusted_roots_pem_files` to docs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/klauspost/compress v1.7.1-0.20190613161414-0b31f265a57b
 	github.com/klauspost/cpuid v1.2.1
 	github.com/lucas-clemente/quic-go v0.7.1-0.20190908032346-fc962d18373a
-	github.com/mholt/certmagic v0.7.2
+	github.com/mholt/certmagic v0.7.3-0.20190917224939-65d418add14f
 	github.com/muhammadmuzzammil1998/jsonc v0.0.0-20190902132743-e4903c4dea48
 	github.com/rs/cors v1.6.0
 	github.com/russross/blackfriday/v2 v2.0.1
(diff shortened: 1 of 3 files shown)


mholt:
@mmalone Pending a review/feedback by you, I can merge this in by tomorrow morning and get this in the beta 2 release this weekend.

mmalone:
Nice. Looks great! +1,000,000.

PR #2770, opened 2019-09-28
replace own implementation with os.UserHomeDir

aca:
It would be better to use standard library for this.

none

none

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -17,7 +17,6 @@ package caddy
 import (
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


aca:
@mholt , oh, I think I missed it. I would close this pr for now.

mholt:
Well, it's not a bad thing; just something we need to discuss and see if that's better than what happens right now. What do you think?

aca:
@mholt I will create an issue for this.

Issue #2694, opened 2019-08-05
OpenBSD - Caddy immediately dies on http requests


jungle-boogie:
Hello,

## 1. Which version of Caddy are you using (`caddy -version`)?
Caddy v1.0.1 (h1:oor6ep+8NoJOabpFXhvjqjfeldtw1XSzfISVrbfqTKo=)

## 2. What are you trying to do?
I would like to use Caddy to serve a 404 page.

## 3. What is your Caddyfile?
```
$ cat Caddyfile
localhost:8080
log access.log
log err.log
```

## 4. How did you run Caddy (give the full command and describe the execution environment)?
`$ ./Downloads/caddy -log err2.log`

## 5. Please paste any relevant HTTP request(s) here.
```
$ curl http://localhost:8080
curl: (56) Recv failure: Connection reset by peer
```

## 6. What did you expect to see?
The default 404 not found page.

## 7. What did you see instead (give full error messages and/or log)?
See curl response above.
```
$ cat err.log err2.log
2019/08/05 07:56:52 [INFO] Serving http://localhost:8080
2019/08/05 07:56:52 [INFO][cache:0xc000092730] Started certificate maintenance routine
```
(note err.log is blank, but err2 contains the text above.)

## 8. Why is this a bug, and how do you think this should be fixed?
This is a bug because Caddy is dying each time there is an http request. I see no reason why it should be dying.

## 9. What are you doing to work around the problem in the meantime?
I know of know work around for this.

You didn't ask about the OS or go version, so I will provide that...

```
$ sysctl kern.version
kern.version=OpenBSD 6.5-current (GENERIC.MP) #171: Sun Aug  4 11:11:52 MDT 2019
    deraadt@amd64.openbsd.org:/usr/src/sys/arch/amd64/compile/GENERIC.MP
```

```
$ go version
go version go1.12.7 openbsd/amd64
```

mholt:
Does this still happen if you remove one of the log directives?

jungle-boogie:
```
$ cat Caddyfile
localhost:8080
```
```
$ ./Downloads/caddy
Activating privacy features... done.

Serving HTTP on port 8080
http://localhost:8080
```
```
$ curl http://localhost:8080
curl: (52) Empty reply from server
```
# Second try

```
$ cat Caddyfile
localhost:8080
```

```
$ ./Downloads/caddy -log err.log
Activating privacy features... done.

Serving HTTP on port 8080
http://localhost:8080
```

```
[...]

PR #2787, opened 2019-10-04
no SetKeepAlivePeriod in openbsd

didil:
I played a bit with this issue https://github.com/caddyserver/caddy/issues/2694 and I was able to reproduce it. The logs don't show anything because the function calls in this function do not return errors https://github.com/caddyserver/caddy/blob/master/caddyhttp/httpserver/server.go#L557
After restoring the errors temporarily to debug I could see the error was:  
````set tcp 127.0.0.1:xxxx->127.0.0.1:yyyy : protocol not available````
After digging around a bit I found out that openbsd has not user-settable per-socket keepalive times http://golang.ir/src/net/tcpsockopt_openbsd.go

Adds a runtime check to skip the call to tc.SetKeepAlivePeriod for openbsd

https://github.com/caddyserver/caddy/issues/2694

None

## 4. Checklist

 (*Can't figure out which tests that would be in this case*)

### DIFF
--- a/caddyhttp/httpserver/server.go
+++ b/caddyhttp/httpserver/server.go
@@ -554,18 +554,19 @@ type tcpKeepAliveListener struct {
 }
 
 // Accept accepts the connection with a keep-alive enabled.
-func (ln tcpKeepAliveListener) Accept() (c net.Conn, err error) {
+func (ln tcpKeepAliveListener) Accept() (net.Conn, error) {
 	tc, err := ln.AcceptTCP()
 	if err != nil {
-		return
+		return nil, err
 	}
 	if err = tc.SetKeepAlive(true); err != nil {
-		return
+		return nil, err
 	}
 	// OpenBSD has no user-settable per-socket TCP keepalive
+	// https://github.com/caddyserver/caddy/pull/2787
 	if runtime.GOOS != "openbsd" {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mholt** [comment]
Oh wow, thanks for looking into this!

I'm surprised that function call does not return errors. What did you change to emit the logs that you found?

File: caddyhttp/httpserver/server.go [hunk 1]
```diff
@@ -562,8 +562,11 @@ func (ln tcpKeepAliveListener) Accept() (c net.Conn, err error) {
 	if err = tc.SetKeepAlive(true); err != nil {
 		return
 	}
-	if err = tc.SetKeepAlivePeriod(3 * time.Minute); err != nil {
-		return
+	// OpenBSD has no user-settable per-socket TCP keepalive
```
mholt: Can you link to the GitHub issue and/or PR so that it is easy for future readers to find out more about this?

didil: done

jungle-boogie:
Thanks @didil for taking a look at this! I see it was merged into master so I'll build from there and try it out.

jungle-boogie:
YES!!! In my simple test, caddy doesn't die anymore.
tested with be2fdb6af6bab03d83839a4b3ea0e01da74ae498

PR #2794, opened 2019-10-08
cmd register use RegisterCommand

yzongyue:
---
name: Pull request
title: ''
labels: 'v2'
assignees: ''
---

sort cmd && use RegisterCommand function to register cmd.

in v2 barnch, cmd registered by variable definition, init function and RegisterCommand, the behavior is inconsistent and confusing.

only manual testing:
2019-10-08 17-04-33屏幕截图

none

none

## 4. Checklist

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -28,6 +28,7 @@ import (
 	"os/exec"
 	"reflect"
 	"runtime/debug"
+	"sort"
 	"strings"
 
 	"github.com/caddyserver/caddy/v2"
@@ -470,7 +471,13 @@ usage:
 
 commands:
 `
-		for _, cmd := range commands {
+		keys := make([]string, 0, len(commands))
+		for k := range commands {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+		for _, k := range keys {
+			cmd := commands[k]
 			short := strings.TrimSuffix(cmd.Short, ".")
 			s += fmt.Sprintf("  %-15s %s\n", cmd.Name, short)
 		}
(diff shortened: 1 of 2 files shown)


mholt:
Cool, I love the sorting of the output.

I almost used `init()` and `RegisterCommand()` for the standard/built-in commands, but realized that if a module tried to register a command named the same as one of the built-in commands, possibly one of the built-in commands could fail to register instead of the third-party one because `init()` runs in an arbitrary order. It's not a huge deal, but it seemed weird that the built-in command would be the one that errors.

I'm fine with introducing more consistency here as your PR does, but can you explain what you mean by:

> the behavior is inconsistent and confusing.

just so I understand the same.

yzongyue:
> Cool, I love the sorting of the output.
> 
> I almost used `init()` and `RegisterCommand()` for the standard/built-in commands, but realized that if a module tried to register a command named the same as one of the built-in commands, possibly one of the built-in commands could fail to register instead of the third-party one because `init()` runs in an arbitrary order. It's not a huge deal, but it seemed weird that the built-in command would be the one that errors.
> 
> I'm fine with introducing more consistency here as your PR does, but can you explain what you mean by:
> 
> > the behavior is inconsistent and confusing.
> 
> just so I understand the same

When i first read the cmd's source code, the help command define separately, although
the comments explain why, i can't understand until finish reading  function `cmdHelp`'s code.

In my opinion this PR's code may more readable and easy to understand.

> but realized that if a module tried to register a command named the same as one of the built-in commands, possibly one of the built-in commands could fail to register instead of the third-party one because `init()` runs in an arbitrary order

I just realized this problem, it seemed weird when program panic but i think people who write the module has same name as one of built-in commands will easy to find out the reason why error occur, just like what you said,  it's not a huge deal.

mholt:
Thanks for your contribution, @yzongyue - hope you'll participate again in the future.

PR #2801, opened 2019-10-10
Add RemoteAddr placeholders

pascalgn:

Add additional placeholders for RemoteAddr (the IP + port of the client)

- `http.request.remote`, e.g. "70.116.186.215:34231"
- `http.request.remote.host`, e.g. "70.116.186.215"
- `http.request.remote.port`, e.g. "34231"

No issue created as this PR is so small. It's fine to close it if it doesn't fit!

List of placeholders needs to be expanded!

## 4. Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,10 @@ Caddyfile
 *.prof
 *.test
 
+# build artifacts
+cmd/caddy/caddy
+cmd/caddy/caddy.exe
+
 # mac specific
 .DS_Store
 
--- a/modules/caddyhttp/replacer.go
+++ b/modules/caddyhttp/replacer.go
@@ -65,13 +65,24 @@ func addHTTPVarsToReplacer(repl caddy.Replacer, req *http.Request, w http.Respon
 					return req.Host, true // OK; there probably was no port
 				}
 				return host, true
-			case "http.request.hostport":
-				return req.Host, true
-			case "http.request.method":
-				return req.Method, true
 			case "http.request.port":
 				_, port, _ := net.SplitHostPort(req.Host)
 				return port, true
+			case "http.request.hostport":
+				return req.Host, true
+			case "http.request.remote":
+				return req.RemoteAddr, true
+			case "http.request.remote.host":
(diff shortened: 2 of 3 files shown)


mholt:
Bizarre, we did this at the same time! Look what I have staged locally:

Screen Shot 2019-10-10 at 12 47 39 PM

Implemented almost exactly the same way, too.

I like yours better though, and it has tests.

One notable difference is I called mine `http.request.remote.ip` and you called it `http.request.remote.host`. While I guess both are correct (since it should never be a host _name_, i.e. a DNS address), `host` is more consistent.

Will review in just a sec.

mholt:
Will accept this PR as soon as the CLA is signed

pascalgn:
This is really bizarre, and also a bit spooky, but it is spooktober, after all! 👻

One more thing I wanted to clarify before we merge is the naming. I didn't write this in the description because I wasn't sure how much response I'd get.

When we have a HTTP request from say 70.116.186.215:34231 to example.com:80 then the following variables would be set:

- http.request.host: example.com
- http.request.port: 80
- http.request.hostport: example.com:80
- http.request.remote: 70.116.186.215:34231
- http.request.remote.host: 70.116.186.215
- http.request.remote.port: 34231

So it's a bit inconsistent, especially the "hostport" part. We could for example rename http.request.remote to http.request.remote.hostport. WDYT?

mholt:
> So it's a bit inconsistent, especially the "hostport" part.

Hmm, I see what you mean, but I disagree, because:

> We could for example rename `http.request.remote` to `http.request.remote.hostport`.

The Request's `Host` header is not the same as the "remote" (client) of the request. So we can't do this.

- The `Host` is where the client is sending the request to.
- The "remote" is where the server received the request from.

I think we better leave the naming as-is...

pascalgn:
OK, I'm also fine with the current naming and I also signed the CLA now! 👍

mholt:
Excellent, thanks for your contribution! Hope to have you contribute again :)

Issue #2805, opened 2019-10-11
v2: reverse proxy HTTPTransport may need add param MaxIdleConnsPerHost


yzongyue:
## 1. Which version of Caddy are you using (`caddy -version`)?

branch v2

## 2. What are you trying to do?

measure reverse proxy latency

## 3. What is your Caddyfile?
```text
http://127.0.0.1:8081 {
    reverse_proxy / 127.0.0.1:8080 {
    }
}

```

a simple hello world app run on :8080, 
helloworld app avg latency: 365us
caddy avg latency: 2.94ms
caddy with param `MaxIdleConnsPerHost` avg latency: 679us

after use pprof debug caddy v2, i realized that there's too many sycall/connect, 
so i try to add  `MaxIdleConnsPerHost: 1000` in `HTTPTransport` , it works, the
avg latency fallen from 2.94ms to 679us

modules/caddyhttp/reverseproxy/reverseproxy.go:95

![image](https://user-images.githubusercontent.com/6824651/66642783-f7793c80-ec4f-11e9-94ee-897e02b6739b.png)

![image](https://user-images.githubusercontent.com/6824651/66642828-12e44780-ec50-11e9-884d-9f3cec0f6d80.png)

mholt:
Ohh this looks great! Thank you for testing this. Do you want to submit a pull request?

PR #2809, opened 2019-10-12
optimization reverse proxy MaxIdleConnsPerHost

yzongyue:
---
name: Pull request
title: ''
labels: ''
assignees: ''
---

Add param `MaxIdleConnsPerHost` in HTTPTransport to optimize reverse proxy latency.
I test some values like 32 as well as 16,100,200, 32 is enough indeed . 
But i think the suiteable value may depend on hardware, linux kernel params, qps... it's difficult for me to found out it, we may need more test or make it configable in the future.

[https://github.com/caddyserver/caddy/issues/2805](https://github.com/caddyserver/caddy/issues/2805)

## 4. Checklist

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -91,8 +91,9 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 	if h.Transport == nil {
 		t := &HTTPTransport{
 			KeepAlive: &KeepAlive{
-				ProbeInterval:   caddy.Duration(30 * time.Second),
-				IdleConnTimeout: caddy.Duration(2 * time.Minute),
+				ProbeInterval:       caddy.Duration(30 * time.Second),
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -93,6 +93,7 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 			KeepAlive: &KeepAlive{
 				ProbeInterval:   caddy.Duration(30 * time.Second),
 				IdleConnTimeout: caddy.Duration(2 * time.Minute),
+				MaxIdleConnsPerHost: 32,
```
mholt: Thanks!

The value is indeed configurable already, so a default of 32 sounds good. We can always lower it in the future if there's a better reason.

Can you please gofmt this file and then I can merge it in!

yzongyue: oh, sorry, gofmt && force update PR's branch has been completed

mholt:
Great, thank you!! Please contribute again :)

Issue #2802, opened 2019-10-10
Allow placeholders in matchers


pascalgn:
## 1. What would you like to have changed?

We would like to use placeholders (especially env. placeholders) in the `match` part:

```json
          "routes": [
            {
              "match": [
                {
                  "host": ["{env.DOMAIN_NAME}"]
                }
              ]
```

## 2. Why is this feature a useful, necessary, and/or important addition to this project?

Especially when running in Docker, it can be useful to have an image with the configuration and run it with a different domain, like `docker run -e DOMAIN_NAME=example.com ...`

## 3. What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

Currently we use some script inside the Docker image which replaces placeholders in the configuration before starting Caddy

## 4. Please link to any relevant issues, pull requests, or other discussions.

None that I know of

pascalgn:
This is for v2, by the way!

I would give it a shot, but I'm not sure where to add the ReplaceAll call...

mholt:
Sure, I can see why this would be useful.

The replacement would probably happen in between these lines: https://sourcegraph.com/github.com/caddyserver/caddy@fe36d26b63b6398592e46604d1795f84ce0477d4/-/blob/modules/caddyhttp/matchers.go#L122-123

pascalgn:
I had a look, but I'm not sure about this: When adding it in matchers.go, it would mean that every incoming request would get a (maybe small?) performance hit due to the replacement logic.

I was thinking more about adding it in the configuration setup, maybe in https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/routes.go#L178? Or maybe at the end of loadConfig, https://github.com/caddyserver/caddy/blob/v2/cmd/main.go#L96? WDYT?

mholt:
@pascalgn Replacers in v2 are very efficient: https://github.com/caddyserver/caddy/pull/2674

> (maybe small?) performance hit

You can test this. [Write a benchmark](https://golang.org/pkg/testing/#hdr-Benchmarks) for the Host matcher without replacer support, then add the ReplaceAll call and run the benchmark again. Post the results here.

pascalgn:
Getting the numbers is easy, but making sense of them... 😄

Without placeholder replacements (status quo)

```
$ go test -bench .
goos: darwin
goarch: amd64
pkg: github.com/caddyserver/caddy/v2/modules/caddyhttp
[...]

PR #2810, opened 2019-10-13
Replace placeholders in host matcher

pascalgn:

Allow environment placeholders in the `match` part:

```json
          "routes": [
            {
              "match": [
                {
                  "host": ["{env.DOMAIN_NAME}"]
                }
              ]
```

#2802

In the `routes.match` section, the `host` part supports the following placeholders:

- `system.hostname`
- `system.os`
- `system.arch`

In addition, environment placeholders are supported, in the form `env.ENV_NAME`.

This allows to use configurations like this:

```json
"routes": [
  {
    "match": [{ "host": ["{system.hostname}"] }]
  }
]
```
(match the hostname that's assigned to the OS caddy is running on)

```json
"routes": [
  {
    "match": [{ "host": ["{env.ENVIRONMENT}.example.com"] }]
  }
]
```
(given the environment variable `ENVIRONMENT` set to `staging`, this will match the hostname `staging.example.com`)

## 4. Checklist

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -118,8 +118,11 @@ func (m MatchHost) Match(r *http.Request) bool {
 		reqHost = strings.TrimSuffix(reqHost, "]")
 	}
 
+	repl := r.Context().Value(caddy.ReplacerCtxKey).(caddy.Replacer)
+
 outer:
 	for _, host := range m {
+		host = repl.ReplaceAll(host, "")
 		if strings.Contains(host, "*") {
 			patternParts := strings.Split(host, ".")
 			incomingParts := strings.Split(reqHost, ".")
--- a/modules/caddyhttp/matchers_test.go
+++ b/modules/caddyhttp/matchers_test.go
@@ -539,22 +539,32 @@ func TestResponseMatcher(t *testing.T) {
 	}
 }
 
-func BenchmarkHostMatcherWithoutPlaceholders(b *testing.B) {
+func BenchmarkHostMatcherWithoutPlaceholder(b *testing.B) {
 	req := &http.Request{Host: "localhost"}
+	repl := caddy.NewReplacer()
+	ctx := context.WithValue(req.Context(), caddy.ReplacerCtxKey, repl)
+	req = req.WithContext(ctx)
+
 	match := MatchHost{"localhost"}
+
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		match.Match(req)
 	}
 }
 
-func BenchmarkHostMatcherWithPlaceholders(b *testing.B) {
+func BenchmarkHostMatcherWithPlaceholder(b *testing.B) {
 	err := os.Setenv("GO_BENCHMARK_DOMAIN", "localhost")
 	if err != nil {
 		b.Errorf("error while setting up environment: %v", err)
 	}
+
 	req := &http.Request{Host: "localhost"}
+	repl := caddy.NewReplacer()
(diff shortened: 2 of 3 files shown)


modules/caddyhttp/matchers.go
```diff
@@ -118,8 +118,11 @@ func (m MatchHost) Match(r *http.Request) bool {
 		reqHost = strings.TrimSuffix(reqHost, "]")
 	}
 
+	repl := caddy.NewReplacer()
```
mholt: This needs to be obtained from the http.Request context; see https://sourcegraph.com/github.com/caddyserver/caddy@fcd8869f51b78f260e3c3aed53d7d87c2051ffe9/-/blob/modules/caddyhttp/matchers.go#L192 for an example.

modules/caddyhttp/matchers.go
```diff
@@ -118,8 +118,11 @@ func (m MatchHost) Match(r *http.Request) bool {
...
+
 outer:
 	for _, host := range m {
+		host = repl.ReplaceAll(host, "")
```
mholt: This is correct :)

modules/caddyhttp/matchers_test.go
```diff
@@ -518,3 +518,12 @@ func TestResponseMatcher(t *testing.T) {
...
 }
+
+func BenchmarkHostMatcherWithoutPlaceholders(b *testing.B) {
+	req := &http.Request{Host: "localhost"}
+	match := MatchHost{"localhost"}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		match.Match(req)
+	}
+}
```
mholt: It would be interesting to also benchmark when a replacement happens. i.e. use a placeholder in `MatchHost` instead of `"localhost"`.

modules/caddyhttp/matchers.go
pascalgn: I did not think of also using the http.Request for replacement, but mainly because there's also this place where we need to replace:

https://github.com/caddyserver/caddy/blob/fcd8869f51b78f260e3c3aed53d7d87c2051ffe9/modules/caddyhttp/caddyhttp.go#L290-L305

But this happens without any request, so we cannot use HTTP request placeholders in the host matcher, I think. But I'm also not entirely sure if we need to, I thought about it a bit and could not come up with a use case. Maybe let's only consider this when someone presents a reasonable use case for it?

mholt: We could do replacements there, sure -- even without an HTTP request, a replacer can still replace environment variable placeholders, which exist outside of HTTP requests. But we definitely don't need to enable automatic HTTPS for dynamic host matchers (that's what On-Demand TLS is for).

pascalgn:
This should work now! 👍

Could you please have a look?

I was unsure if I should move the init logic in the tests
```
	err := os.Setenv("GO_BENCHMARK_DOMAIN", "localhost")
	if err != nil {
		t.Errorf("error while setting up environment: %v", err)
	}
```
into a separate method...

modules/caddyhttp/matchers.go
mholt: @pascalgn This line (NewReplacer()) still needs to be replaced with `repl := r.Context().Value(caddy.ReplacerCtxKey).(caddy.Replacer)` -- we have to use the existing Replacer in order to take advantage of all the HTTP-request-specific variables. It also performs better!

pascalgn:
Fixed! 👍

pascalgn:
I just did an end-to-end test with a Docker image built from the pascalgn:v2-matcher-placeholders branch and everything is looking good! 👍

PR #2821, opened 2019-10-18
v2 CI: cache the GOCACHE directory

Mohammed90:

Use CacheBeta to cache the contents of GOCACHE directory so builds don't need to re-download previously downloaded deps that haven't changed. This should speed up the builds.

Let's see how this works for us.

N/A

N/A

## 4. Checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -62,7 +62,7 @@ steps:
     $ProgressPreference = 'SilentlyContinue'
     Write-Host "Downloading Go..."
     (New-Object System.Net.WebClient).DownloadFile("https://dl.google.com/go/$(LATEST_GO).windows-amd64.zip", "$(LATEST_GO).windows-amd64.zip")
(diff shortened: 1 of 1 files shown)


Mohammed90:
This is weird. Even with a cache hit, it could take as much time to get deps as with cache miss 🤷‍♂  Comparing a few builds, it shaves 1 minute at most, but could be any where between a few seconds to a minute.

mholt:
Wow, thanks for experimenting with this!

> This is weird. Even with a cache hit, it could take as much time to get deps as with cache miss 🤷‍♂ Comparing a few builds, it shaves 1 minute at most, but could be any where between a few seconds to a minute.

So... I mean... is it worth adding this extra complexity to our pipeline?

Mohammed90:
Nope, ain't worth it.

mholt:
Thank you for trying it anyway :)

Issue #847, opened 2016-05-25
Follow XDG Base Dir Spec


nesl247:
Caddy should probably support https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html rather than writing to `$HOME/.caddy`.

While OS X doesn't follow this, linux does. The behavior I've seen frequently enough is to just always use the basedir spec regardless of platform. It makes it easier.

nesl247:
The following is a WIP patch for this. I followed the handling of paths from NeoVim which followed ActiveState which based it off of MSDN docs. 

I didn't submit this as a PR because technically this should be `$XDG_DATA_HOME` instead of `XDG_CONFIG_HOME`. XDG specifies a multitude of directories, with config and data being the ones we would need. From what I can see, the current use for `~/.caddy` is for ssl certs. I'm not 100% sure if that should be considered configuration or data.

If it is data, then we need to be able to support two types of paths, data and configuration. I think I could figure that out, but didn't want to waste my time doing all that if it wasn't going to be used.

The other question I have is that when I went to go do the tests (which I started with, I try to follow TDD), I had no idea on how to stub/mock the `os` package to be able to write tests for each of the different branches. Maybe that's not how things are done in golang, or just in caddy, but it was definitely something I could use help with.

```
diff --git a/caddy/assets/path.go b/caddy/assets/path.go
index 46b883b..bd3469b 100644
--- a/caddy/assets/path.go
+++ b/caddy/assets/path.go
@@ -8,22 +8,21 @@ import (
 
 // Path returns the path to the folder
 // where the application may store data. This
-// currently resolves to ~/.caddy
+// currently resolves to $XDG_HOME_CONFIG/caddy
 func Path() string {
-	return filepath.Join(userHomeDir(), ".caddy")
+	return filepath.Join(userHomeDir(), "caddy")
 }
 
 // userHomeDir returns the user's home directory according to
 // environment variables.
-//
-// Credit: http://stackoverflow.com/a/7922977/1048862
 func userHomeDir() string {
 	if runtime.GOOS == "windows" {
-		home := os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
-		if home == "" {
-			home = os.Getenv("USERPROFILE")
-		}
-		return home
+		os.Getenv("LOCALAPPDATA")
 	}
-	return os.Getenv("HOME")
+
+	home := os.Getenv("XDG_CONFIG_HOME")
+	if home == "" {
[...]

PR #2826, opened 2019-10-23
replace XDG Directory configuration to standard os.UserConfigDir

aca:
v2

Remove xdg config directory for universal support.

v2 introduces XDG specification. 
While it is popular in linux, It's only for linux. 
I don't think we need non-standard, complex third party packages for this.
But as go 1.13 now has os.UserConfigDir I think we should use it.
If we simply use os.UserConfigDir, we can also just remove homedir.(#2770)

os.UserConfigDir uses XDG_CONFIG_HOME instead of XDG_CONFIG_DATA for linux. 
But it seems quite common. https://github.com/golang/go/issues/29960#issue-403761007

And getting current directory using "." is too fragile process. Replaced it with proper function.

cancel xdg configuration

## 4. Checklist

- [o] I have written tests and verified that they fail without my change
- [o] I have squashed any insignificant commits
- [o] This change has comments explaining package types, values, functions, and non-obvious lines of code
- [o] I am willing to help maintain this change if there are issues with it later

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -17,7 +17,6 @@ package caddy
 import (
 	"os"
 	"path/filepath"
-	"runtime"
 
 	"github.com/mholt/certmagic"
 )
@@ -31,34 +30,15 @@ type StorageConverter interface {
 	CertMagicStorage() (certmagic.Storage, error)
 }
 
-// homeDir returns the best guess of the current user's home
-// directory from environment variables. If unknown, "." (the
-// current directory) is returned instead.
-func homeDir() string {
-	home := os.Getenv("HOME")
-	if home == "" && runtime.GOOS == "windows" {
-		drive := os.Getenv("HOMEDRIVE")
-		path := os.Getenv("HOMEPATH")
-		home = drive + path
-		if drive == "" || path == "" {
-			home = os.Getenv("USERPROFILE")
-		}
-	}
-	if home == "" {
-		home = "."
-	}
-	return home
-}
-
 // dataDir returns a directory path that is suitable for storage.
-// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables
+// If the location cannot be determined, use current directory.
 func dataDir() string {
-	baseDir := filepath.Join(homeDir(), ".local", "share")
-	if xdgData := os.Getenv("XDG_DATA_HOME"); xdgData != "" {
-		baseDir = xdgData
+	baseDir, err := os.UserConfigDir()
+	if err != nil {
+		baseDir, err = os.Getwd()
+		if err != nil {
+			panic("failed to locate current directory")
+		}
 	}
 	return filepath.Join(baseDir, "caddy")
 }
-
-// TODO: Consider using Go 1.13's os.UserConfigDir() (https://golang.org/pkg/os/#UserConfigDir)
-// if we are going to store the last-loaded config anywhere


mholt:
Thanks for the PR, but I think this needs some more discussion first.

We've had a lot of feature requests for XDG though (#847 is just some of them). If you don't want to use it, you don't have to use it. I do not think it is harmful to support it by default.

> I don't think we need non-standard, complex third party packages for this.

Which "non-standard, complex third party packages" are we using for it?

> But as go 1.13 now has os.UserConfigDir I think we should use it.

The problem with this is that this function returns the user's _config_ directory, not their _data_ directory:

> On Unix systems, it returns $XDG_CONFIG_HOME as specified by https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html if non-empty, else $HOME/.config. On Darwin, it returns $HOME/Library/Application Support. On Windows, it returns %AppData%. On Plan 9, it returns $home/lib.

But we need the data directory, $XDG_DATA_DIR.

> And getting current directory using "." is too fragile process. 

I don't understand why that is fragile.

aca:
Obviously, there was not enough explanation.


> Which "non-standard, complex third party packages" are we using for it?

> We've had a lot of feature requests for XDG though (#847 is just some of them). If you don't want to use it, you don't have to use it. I do not think it is harmful to support it by default.

I also think it is good to support it by default. Only problem with XDG base directory is that is absolutely Linux specific. What I meant is that to support this stuffs in more robust way, we might need to implement / use third party libraies like this.  https://github.com/shibukawa/configdir. As most reliable go standard library is provided to fix this, I think It would be better to use it.

>   Mac and Windows don't obey XDG, so they only have one persistent per-user directory
    Even on Linux, most apps store all their data in their config dir. For example: Chromium, Libreoffice, GIMP, and gcloud all seem to ignore XDG_DATA_HOME
    Most programs only need one persistent per-user directory, and it's often for config files; see the links above for examples

https://github.com/golang/go/issues/29960 explains about, XDG_DATA_HOME vs XDG_CONFIG_HOME





> And getting current directory using "." is too fragile process.

I shouldn't have said that as I can't explain exact reason for this.
https://github.com/golang/go/blob/5d000a8b6268c09697c64c76bade1daa86f43a9e/src/os/getwd.go#L27
But when I looked at how standard library implement "get current directory". It seems It is much more complicated than just relying on '.' variable.

aca:
@mholt any more thoughts with this?

mholt:
@aca Sort of, I still need to think about it more...

> I also think it is good to support it by default. Only problem with XDG base directory is that is absolutely Linux specific. What I meant is that to support this stuffs in more robust way, we might need to implement / use third party libraies like this. https://github.com/shibukawa/configdir. As most reliable go standard library is provided to fix this, I think It would be better to use it.

It *is* kind of nice to have everything in one place no matter the OS -- makes troubleshooting and documentation easier. Plus, I suppose it is more likely that the process will have write permissions to the $HOME folder than other paths in the file system. But I can also see the benefit of adopting an OS's usual storage paths...

> I shouldn't have said that as I can't explain exact reason for this.
https://github.com/golang/go/blob/5d000a8b6268c09697c64c76bade1daa86f43a9e/src/os/getwd.go#L27
But when I looked at how standard library implement "get current directory". It seems It is much more complicated than just relying on '.' variable.

Ah, that's because that function returns the literal, expanded path of the current directory. Whereas using `.` simply tells the OS to expand it for us. We don't actually need to know what the current directory is; we just need to make paths relative to the current directory and we can let the OS do the expansion. So that's the difference. Using `.` is fine for our needs.

mholt:
Going to close this for now since I don't think it's quite the right solution (I'm still not convinced it's important to change what we have, as it's not really broken AFAIK).

PR #2828, opened 2019-10-23
feature: Add pledge for OpenBSD

### LINKED ISSUE #2817: OpenBSD pledge
I think it would be beneficial to add a call to [`pledge`](https://man.openbsd.org/pledge.2) at startup for OpenBSD servers.

`pledge` would restrict the caddy process to only being able to invoke the system calls that it needs to function. This would in turn reduce the chances of any errant system calls that could potentially cause issues on the host. It is an easy way to reduce the attack vector and has been implemented in several [OpenBSD](https://www.openbsd.org/plus65.html) [releases](https://www.openbsd.org/plus66.html).

This isn't strictly necessary to add to caddy and could pose problems. If the pledge promises are incorrect, functionality could be limited and cause issues for OpenBSD users. It would require some investigation at first to ensure that the correct subset of promises gets implemented.

It would also add another dependency to the binary to be able to abstract the call via Go's [sys package](https://github.com/golang/sys/blob/master/unix/pledge_openbsd.go). Of course this would be conditionally compiled but there is no way to conditionally add it to the list of modules as far as I know.

gsquire:
This patch adds a pledge system call for OpenBSD builds to better
protect caddy on this operating system.

Signed-off-by: Garrett Squire <garrettsquire@gmail.com>

It adds a call to `pledge` for OpenBSD builds.

Closes #2817 

It may be worth noting this somewhere but I am not sure yet.

## 4. Checklist

### DIFF
--- a/pledge_openbsd.go
+++ b/pledge_openbsd.go
@@ -1,3 +1,5 @@
+// +build openbsd
+
 package caddy
 
 import (
--- a/pledge_openbsd.go
+++ b/pledge_openbsd.go
@@ -1,0 +1,27 @@
+package caddy
+
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/sys/unix"
+)
+
+// pledge invokes the pledge system call with the associated promises. It passes `NULL` to the
+// `execpromises` argument.
+func pledge() {
+	const (
+		// Allows things such as read, write, and socket operations.
+		io = "stdio"
+		// Allows setting up a socket to listen.
+		net = "inet"
+		// Allows calling `setrlimit`.
+		limit = "proc"
+	)
+
+	promises := strings.Join([]string{io, net, limit}, " ")
+	err := unix.PledgePromises(promises)
+	if err != nil {
+		fmt.Printf("WARNING: Error calling pledge: %s", err)
+	}
+}
--- a/caddy.go
+++ b/caddy.go
@@ -469,6 +469,9 @@ func (i *Instance) Caddyfile() Input {
 //
 // This function blocks until all the servers are listening.
 func Start(cdyfile Input) (*Instance, error) {
+	// This restricts system operations on OpenBSD and is a no-op on all other operating systems.
+	pledge()
+
 	inst := &Instance{serverType: cdyfile.ServerType(), wg: new(sync.WaitGroup), Storage: make(map[interface{}]interface{})}
 	err := startWithListenerFds(cdyfile, inst, nil)
 	if err != nil {
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/naoina/toml v0.1.1
 	github.com/russross/blackfriday v0.0.0-20170610170232-067529f716f4
 	golang.org/x/net v0.0.0-20190328230028-74de082e2cca
+	golang.org/x/sys v0.0.0-20191023151326-f89234f9a2c2 // indirect
 	gopkg.in/mcuadros/go-syslog.v2 v2.2.1
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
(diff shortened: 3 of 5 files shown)


pledge_openbsd.go
```diff
@@ -0,0 +1,27 @@
+package caddy
```
gsquire: Forgot to add a build tag but I will update it.

francislavoie:
Since Caddy v1 isn't going to get a whole lot of additional changes from here on out, do you think you could look into contributing this change the v2 as well? See the v2 branch.

gsquire:
@francislavoie Should I close this PR in favor of the V2 version? Or keep both?

jungle-boogie:
hi @francislavoie,

Binary releases of `caddy` v1 are available for OpenBSD. Do you think binary releases of `caddy` will be available for OpenBSD?

francislavoie:
I don't see why they wouldn't be. I figure @mholt just hasn't built them due to lack of demand during the beta.

jungle-boogie:
Thanks for the reply, @francislavoie. FWIW, I haven't yet tried building v2 on my OpenBSD machines yet.

mholt:
Should this be optional? Are we sure that these 3 pledges are all that will be needed, ever? What about plugins that need other system calls? What specific threats does this protect against?

gsquire:
How could it be optional? That would be good to start with I think.

> Are we sure that these 3 pledges are all that will be needed, ever?

I'm *fairly* confident that it will but it does require some investigation. The other pledge values are pretty specialized or centered around audio, video, etc.

> What about plugins that need other system calls?

It's hard to say from a quick glance. Presumably you could call `pledge` after all plugins are loaded unless they are running alongside the server.

> What specific threats does this protect against?

It allows you to effectively sandbox your application by isolating it to certain profiles. In a worst case scenario, an attacker that got control of a caddy process wouldn't be able to exploit it to do certain things.

mholt:
If this has the potential to break any Caddy 1 users, we should not merge this into Caddy 1... so we'll need to be very careful.

In that sense, it might be safer to develop this for Caddy 2 (and then we can always remove them later if they are problematic, right?), as long as you are willing to help maintain this feature!

Also, is there any documented cases of a Go program being controlled by an attacker? Go programs aren't vulnerable to memory exploits that typically allow these kinds of attacks.

gsquire:
I'd be happy to help maintain this and see it through. With that being said I am also alright with tabling this feature for now. I certainly don't want to cause issues for any existing or new users.

I don't know if Go's memory safety precludes it from any type of attacks but I am not a threat researcher by any means. I see it as more of a "nice-to-have" than anything.

jungle-boogie:
@gsquire, speaking on only myself, would you be able to work this in `caddy` v2? IMO, if there's a chance things _could_ break, v2 would probably be the best place to do it. 

OpenBSD's `httpd` has this for pledge:
https://github.com/openbsd/src/blob/master/usr.sbin/httpd/httpd.c#L227

Would caddy need those extra permissions?

mholt:
That's good input, thanks @jungle-boogie.

I still would like to know the actual risk of this being possible in Caddy before adding it as a "nice to have", since the consequences aren't fully understood.

> I don't know if Go's memory safety precludes it from any type of attacks

Yeah, I don't mean that it's not impossible, I just can't think of a way to do it off the top of my head.

A PoC attack on Caddy would be _ideal_ or at least a documented case of something similar happening in another pure-Go web server, if possible.

mholt:
@gsquire I fail to see how either of those vulnerabilities are relevant. Neither of those grant the attacker control over the process.

I think, considering the risks that this introduces, and the lack of understanding as to the consequences of this change, and the lack of motivation for considering this change necessary, I'm going to close this. But, thanks for the PR, and if things are different we can revisit this later.

Issue #2829, opened 2019-10-23
File server could download folders as zip file


jaysonsantos:
## 1. What would you like to have changed?

It would be great if file server could download folders a zip files.

## 2. Why is this feature a useful, necessary, and/or important addition to this project?

Sometimes you want to serve a big amount of files and it is hard to download each one, a zip older would make life easier.

## 3. What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

I implemented a POC against branch v2 and I intend to submit a WIP PR for discussion

## 4. Please link to any relevant issues, pull requests, or other discussions.

https://github.com/caddyserver/caddy/pull/2830

francislavoie:
Great idea! Thanks for putting forward the effort to make a POC :smile:

mholt:
Interesting idea!

Why should this be baked into the file server, instead of a separate plugin?

tobya:
I think part of fileserver makes sense, but as an option

francislavoie:
I'll close this for now as it was decided during discussion in the PR that this should probably be done in a plugin: https://github.com/caddyserver/caddy/pull/2830#issuecomment-678792010

PR #2830, opened 2019-10-23
Implement a POC folder downloader

jaysonsantos:

For now this only changes browse's template to include a link to download folders as zip files.

https://github.com/caddyserver/caddy/issues/2829

I would say tha this feature should be optional, if it is an accepted idea that should for sure be documented

## 4. Checklist

I would like to hear from you before writing tests as I am not experienced in go development

## 5. Missing points
- [ ] Use accept header for negotiation
- [ ] Create svg icons
- [ ] Respect hidden files and avoid shallow download?

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -100,6 +100,8 @@ github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4/go.mod h1:ybxpYRF
 github.com/aliyun/alibaba-cloud-sdk-go v0.0.0-20190808125512-07798873deee h1:NYqDBPkhVYt68W3yoGoRRi32i3MLx2ey7SFkJ1v/UI0=
 github.com/aliyun/alibaba-cloud-sdk-go v0.0.0-20190808125512-07798873deee/go.mod h1:myCDvQSzCW+wB1WAlocEru4wMGJxy+vlxHdhegi1CDQ=
 github.com/aliyun/aliyun-oss-go-sdk v0.0.0-20190307165228-86c17b95fcd5/go.mod h1:T/Aws4fEfogEE9v+HPhhw+CntffsBHJ8nXQCwKr0/g8=
+github.com/andybalholm/brotli v0.0.0-20190621154722-5f990b63d2d6 h1:bZ28Hqta7TFAK3Q08CMvv8y3/8ATaEqv2nGoc6yff6c=
+github.com/andybalholm/brotli v0.0.0-20190621154722-5f990b63d2d6/go.mod h1:+lx6/Aqd1kLJ1GQfkvOnaZ1WGmLpMpbprPuIOOZX30U=
 github.com/anmitsu/go-shlex v0.0.0-20161002113705-648efa622239/go.mod h1:2FmKhYUyUczH0OGQWaF5ceTx0UBShxjsH6f8oGKYe2c=
 github.com/antihax/optional v0.0.0-20180407024304-ca021399b1a6/go.mod h1:V8iCPQYkqmusNa815XgQio277wI47sdRh1dUOLdyC6Q=
 github.com/antlr/antlr4 v0.0.0-20190819145818-b43a4c3a8015 h1:StuiJFxQUsxSCzcby6NFZRdEhPkXD5vxN7TZ4MD6T84=
@@ -182,6 +184,8 @@ github.com/dlclark/regexp2 v1.2.0 h1:8sAhBGEM0dRWogWqWyQeIJnxjWO6oIjl8FKqREDsGfk
 github.com/dlclark/regexp2 v1.2.0/go.mod h1:2pZnwuY/m+8K6iRw6wQdMtk+rH5tNGR1i55kozfMjCc=
 github.com/dnaeon/go-vcr v0.0.0-20180814043457-aafff18a5cc2/go.mod h1:aBB1+wY4s93YsC3HHjMBMrwTj2R9FHDzUr9KyGc8n1E=
 github.com/dnsimple/dnsimple-go v0.30.0/go.mod h1:O5TJ0/U6r7AfT8niYNlmohpLbCSG+c71tQlGr9SeGrg=
+github.com/dsnet/compress v0.0.1/go.mod h1:Aw8dCMJ7RioblQeTqt88akK31OvO8Dhf5JflhBbQEHo=
+github.com/dsnet/golib v0.0.0-20171103203638-1ea166775780/go.mod h1:Lj+Z9rebOhdfkVLjJ8T6VcRQv3SXugXy999NBtR9aFY=
 github.com/dustin/go-humanize v1.0.0 h1:VSnTsYCnlFHaM2/igO1h6X3HA71jcobQuxemgkq4zYo=
 github.com/dustin/go-humanize v1.0.0/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25KnS6fMYU6eOk=
 github.com/dustin/go-humanize v1.0.1-0.20200219035652-afde56e7acac h1:opbrjaN/L8gg6Xh5D04Tem+8xVcz6ajZlGCs49mQgyg=
@@ -247,6 +251,8 @@ github.com/gogo/protobuf v1.2.0/go.mod h1:r8qH/GZQm5c6nD/R0oafs1akxWv10x8SbQlK7a
 github.com/gogo/protobuf v1.2.1/go.mod h1:hp+jE20tsWTFYpLwKvXlhS1hjn+gTNwPg2I6zVXpSg4=
 github.com/gogo/protobuf v1.3.1/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
 github.com/goji/httpauth v0.0.0-20160601135302-2da839ab0f4d/go.mod h1:nnjvkQ9ptGaCkuDUx6wNykzzlUixGxvkme+H/lnzb+A=
+github.com/golang/gddo v0.0.0-20190419222130-af0f2af80721 h1:KRMr9A3qfbVM7iV/WcLY/rL5LICqwMHLhwRXKu99fXw=
+github.com/golang/gddo v0.0.0-20190419222130-af0f2af80721/go.mod h1:xEhNfoBDX1hzLm2Nf80qUvZ2sVwoMZ8d6IE2SrsQfh4=
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekfRL6a582T4P37/31XEstQ5p58=
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
 github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
@@ -274,6 +280,7 @@ github.com/golang/protobuf v1.3.3/go.mod h1:vzj43D7+SQXF/4pzW/hwtAqwc6iTitCiVSaW
 github.com/golang/protobuf v1.3.4 h1:87PNWwrRvUSnqS4dlcBU/ftvOIBep4sYuBLlh6rX2wk=
 github.com/golang/protobuf v1.3.4/go.mod h1:vzj43D7+SQXF/4pzW/hwtAqwc6iTitCiVSaWz5lYuqw=
 github.com/golang/snappy v0.0.0-20180518054509-2e65f85255db/go.mod h1:/XxbfmMg8lxefKM7IXC3fBNl/7bRcc72aCRzEWrmP2Q=
+github.com/golang/snappy v0.0.1/go.mod h1:/XxbfmMg8lxefKM7IXC3fBNl/7bRcc72aCRzEWrmP2Q=
 github.com/golangci/check v0.0.0-20180506172741-cfe4005ccda2/go.mod h1:k9Qvh+8juN+UKMCS/3jFtGICgW8O96FVaZsaxdzDkR4=
 github.com/golangci/dupl v0.0.0-20180902072040-3e9179ac440a/go.mod h1:ryS0uhF+x9jgbj/N71xsEqODy9BN81/GonCZiOzirOk=
 github.com/golangci/errcheck v0.0.0-20181223084120-ef45e06d44b6/go.mod h1:DbHgvLiFKX1Sh2T1w8Q/h4NAI8MHIpzCdnBUDTXU3I0=
@@ -398,12 +405,14 @@ github.com/kisielk/gotool v0.0.0-20161130080628-0de1eaf82fa3/go.mod h1:jxZFDH7IL
 github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+oQHNcck=
 github.com/klauspost/compress v1.4.0/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
 github.com/klauspost/compress v1.4.1/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
+github.com/klauspost/compress v1.9.2/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
 github.com/klauspost/compress v1.10.3 h1:OP96hzwJVBIHYU52pVTI6CczrxPvrGfgqF9N5eTO0Q8=
 github.com/klauspost/compress v1.10.3/go.mod h1:aoV0uJVorq1K+umq18yTdKaF57EivdYsUV+/s2qKfXs=
 github.com/klauspost/cpuid v0.0.0-20180405133222-e7e905edc00e/go.mod h1:Pj4uuM528wm8OyEC2QMXAi2YiTZ96dNQPGgoMS4s3ek=
 github.com/klauspost/cpuid v1.2.0/go.mod h1:Pj4uuM528wm8OyEC2QMXAi2YiTZ96dNQPGgoMS4s3ek=
 github.com/klauspost/cpuid v1.2.3 h1:CCtW0xUnWGVINKvE/WWOYKdsPV6mawAtvQuSl8guwQs=
 github.com/klauspost/cpuid v1.2.3/go.mod h1:Pj4uuM528wm8OyEC2QMXAi2YiTZ96dNQPGgoMS4s3ek=
+github.com/klauspost/pgzip v1.2.1/go.mod h1:Ch1tH69qFZu15pkjo5kYi6mth2Zzwzt50oCQKQE9RUs=
 github.com/kolo/xmlrpc v0.0.0-20190717152603-07c4ee3fd181/go.mod h1:o03bZfuBwAXHetKXuInt4S7omeXUu62/A845kiycsSQ=
 github.com/konsorten/go-windows-terminal-sequences v0.0.0-20180402223658-b729f2633dfe/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
 github.com/konsorten/go-windows-terminal-sequences v1.0.1/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
@@ -468,6 +477,7 @@ github.com/mattn/go-sqlite3 v1.10.0/go.mod h1:FPy6KqzDD04eiIsT53CuJW3U88zkxoIYsO
 github.com/mattn/go-tty v0.0.0-20180219170247-931426f7535a/go.mod h1:XPvLUNfbS4fJH25nqRHfWLMa1ONC8Amw+mIA639KxkE=
 github.com/mattn/goveralls v0.0.2/go.mod h1:8d1ZMHsd7fW6IRPKQh46F2WRpyib5/X4FOpevwGNQEw=
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
+github.com/mholt/archiver/v3 v3.3.0/go.mod h1:YnQtqsp+94Rwd0D/rk5cnLrxusUBUXg+08Ebtr1Mqao=
 github.com/microcosm-cc/bluemonday v1.0.1/go.mod h1:hsXNsILzKxV+sX77C5b8FSuKF00vh2OMYv+xgHpAMF4=
 github.com/miekg/dns v1.1.15 h1:CSSIDtllwGLMoA6zjdKnaE6Tx6eVUxQ29LUgGetiDCI=
 github.com/miekg/dns v1.1.15/go.mod h1:W1PPwlIAgtquWBMBEV9nkV9Cazfe8ScdGz/Lj7v3Nrg=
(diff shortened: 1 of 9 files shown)


jaysonsantos:
Hey there, an update on the performance of this process.
I ran curl piping to pv to measure the speed that the server can output the zip file with different buffer sizes.


```
4096      -> 3.69GiB 0:00:10 [ 370MiB/s]
4096 * 2  -> 3.69GiB 0:00:09 [ 409MiB/s]
4096 * 5  -> 3.69GiB 0:00:04 [ 894MiB/s]
4096 * 10 -> 3.69GiB 0:00:03 [1.21GiB/s]
4096 * 20 -> 3.69GiB 0:00:04 [ 790MiB/s]
```

The sweet spot is 40 KB but for 1k clients that would use at least 40 MB of memory, would it still be worth it?

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -94,6 +94,7 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, _ cadd
 	root := repl.ReplaceAll(fsrv.Root, ".")
 	suffix := repl.ReplaceAll(r.URL.Path, "")
 	filename := sanitizedPathJoin(root, suffix)
+	streamAsZip := r.URL.Query()["zip"] != nil
```
tobya: I think we should use a more descriptive query parameter name to reduce lieklyhood of being used elsewhere in template.

Perhaps `zipfolder` or `downloadasZip`

modules/caddyhttp/fileserver/browsetpl.go
```diff
@@ -358,6 +358,9 @@ footer {
 								{{- end}}
 								<span class="name">{{html .Name}}</span>
 							</a>
+							{{- if .IsDir}}
```
tobya: I am unsure if this should be in the default template?

Maybe see if anyone has an opinion @whitestrake @mholt @elcore

go.sum
```diff
@@ -38,6 +38,7 @@ github.com/baiyubin/aliyun-sts-go-sdk v0.0.0-20180326062324-cfa1a18b161f/go.mod
 github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973/go.mod h1:Dwedo/Wpr24TaqPxmxbtue+5NUziq4I4S80YR8gNf3Q=
 github.com/beorn7/perks v1.0.0/go.mod h1:KWe93zE9D1o94FZ5RNwFwVgaQK1VOXiVxmqh+CedLV8=
 github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6rlkpw=
+github.com/caddyserver/caddy v1.0.3 h1:i9gRhBgvc5ifchwWtSe7pDpsdS9+Q0Rw9oYQmYUTw1w=
```
mholt: Must not use caddy v1

modules/caddyhttp/fileserver/browse.go
```diff
@@ -146,6 +150,89 @@ func (fsrv *FileServer) browseWriteHTML(listing browseListing) (*bytes.Buffer, e
 	return buf, err
 }
 
+func (fsrv *FileServer) streamFolderAsZip(baseFolderName, downloadFolderName string, w http.ResponseWriter) error {
+	w.Header().Set("Content-type", "application/zip")
```
mholt: Of course, I know the first question we're going to get is requests for other formats: tar.gz, tar.bz2, and .tar.zstd among the more likely I think.

modules/caddyhttp/fileserver/browsetpl.go
mholt: Yeah, good question... I'm not sure yet either. Seems a bit niche.

modules/caddyhttp/fileserver/staticfiles.go
mholt: If possible I don't think we need a query parameter at all; there's already a way to specify how you want your content, and that's through content negotiation. The `Accept` header should be sufficient.

tobya: I disagree.  I think the concept of a link added into the browse template is very elegant and this would not be possible if using the Accept header.  Perhaps both but I would keep the query param also.

jaysonsantos: Hey there, the accept parameter would not work as a link to be downloaded on the browser unless someone explicitly sends it while requesting a folder like on curl.
About the parameter renaming, I will do it and send later.

modules/caddyhttp/fileserver/browsetpl.go
jaysonsantos: Let me know where to put it

go.sum
jaysonsantos: Probably I did something wrong there, I will remove it.

jaysonsantos:
@tobya 
> Is there any benefit in caching the zip file for reuse? Is the extra storage space used better than the extra cpu space used?
Depending on the size of the folder and/or it could also kill the server if a cache is added.

> Also could you please provide basic documentation additions if you can?
Do you mean on user docs? 

@mholt 
>Bulk compression (esp. without caching) can add significant load to busy servers, so this feature should be opt-in.
I also agree it should be opt-in

> Does this feature belong here by default in the first place? Why not a plugin?
I am not sure to be honest, let me know what to do here if I would keep on this pr or move somewhere else.

> I think if we do add it, we should use (and/or improve) https://github.com/mholt/archiver instead of re-inventing the wheel.
I thought of zip as it is a universal format and is already implemented in the stdlib

> The Accept request header should be used to specify an archive type, instead of using query string parameters for this, I think.
And what about browsers downloading?

mholt:
The more I think about it, the more I like this, but I'm going to be kind of picky about it :)

- We should definitely use `archiver` -- there's no question that people will want different formats, as `unzip` isn't even available on some systems AFAIK.

- If we want to support point-and-click downloads within a web browser, we need to find an elegant way to do that, if possible. A query string parameter is "easy" but is kind of hacky.

- Regardless of the query string, we absolutely need to honor the Accept header for something like this. Archive downloads shouldn't be allowed if the client doesn't want it.

- Yeah, don't add a cache here. We have a [separate handler](https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/httpcache/httpcache.go) for caching stuff.

Good start, I think we can work out the details!

jaysonsantos:
@mholt 


> We should definitely use archiver -- there's no question that people will want different formats, as unzip isn't even available on some systems AFAIK.

I changed to use archiver and create a function that simply returns the Writer interface

> If we want to support point-and-click downloads within a web browser, we need to find an elegant way to do that, if possible. A query string parameter is "easy" but is kind of hacky.

In the beginning, I thought of having something like `folder/asZip` but that would make people confused if asZip was a directory, thus the query parameter.

> Regardless of the query string, we absolutely need to honor the Accept header for something like this. Archive downloads shouldn't be allowed if the client doesn't want it.

Should I check accept and make the query parameter override it?

mholt:
Cool, I think this is looking better. I'll give it another review later, maybe tomorrow or Tuesday.

Your feature request bumps into a weird niche where the HTTP standard has tools to support this exact kind of functionality, but web browsers don't implement it! Well, they don't expose it to users at all. If a browser's "File -> Save Page As..." feature would just make the same GET request again but with `Accept: application/zip` (or whatever format) header, this would be a done deal. Servers that can't satisfy the Accept header can send a 406 status code in return.

I wonder if we should file this as a feature request for web browsers.

In the meantime, a query string parameter is probably OK, but we'd have to support it forever, which I kinda hate supporting a hack in brand new software... for forever.

Let's first find out if we can get browsers to implement it. Until then, I'd be more willing to merge this sooner if we nix the download link and query string stuff, and only use the Accept header. (For now! We can always add the download link back later.)

I even probed Twitter if there is a better way: https://twitter.com/mholt6/status/1188556552795344897

jaysonsantos:
@mholt About the optional flag, should I implement it on FileServer struct, on the command line and on caddyfile parser?

modules/caddyhttp/fileserver/browsetpl.go
jaysonsantos: Also, this has to be thought because it is quite ugly the way i put, maybe a new <td> with an svg icon?

tobya:
I think the option should be part of the `template` directive

https://caddyserver.com/v1/docs/templates

````
templates {
	path    basepath
	ext     extensions...
	between open_delim close_delim
        folder_compression gzip 

}
````

If `folder_compression` (or whatever) exists then streaming compressed folders is available

**Comment by jaysonsantos** [comment]
Hey @tobya but wouldn't that make the module share state/configs? To decouple them would be quite hard in the future.

**Comment by mholt** [comment]
@tobya I think we're confused because this change has nothing to do with the templates handler. This is a change in the static file server.

@jaysonsantos Incidentally, how much size would this add to the resulting binary? If it's nontrivial then we should consider making this a plugin so its compilation can be optional.

**Comment by jaysonsantos** [comment]
@mholt 
```
git checkout v2
Already on 'v2'
Your branch is up to date with 'origin/v2'.
~/g/s/g/c/caddy ❯❯❯ go build cmd/caddy/main.go
~/g/s/g/c/caddy ❯❯❯ du -h main
24M	main

~/g/s/g/c/caddy ❯❯❯ git checkout download-zip
Switched to branch 'download-zip'
Your branch is up to date with 'jaysonsantos/download-zip'.
~/g/s/g/c/caddy ❯❯❯ rm main
rm: remove regular file 'main'? y
~/g/s/g/c/caddy ❯❯❯ go build cmd/caddy/main.go
~/g/s/g/c/caddy ❯❯❯ du -h main
24M	main
```

**Comment by mholt** [comment]
That's surprising, considering all the dependencies that the `archiver` package has.

In summary, I think this is how I'd like to see this feature work, considering the unfortunate reality browsers are lame-sauce with the Accept header:

- This feature is off by default, but can be enabled in the [`browse` property](https://github.com/caddyserver/caddy/wiki/v2:-Documentation#httphandlersfile_server) by specifying `"dir_archives": [""]`, where the list is a list of allowed formats: `["zip", "tar.gz", "tar.xz"]` (etc).

- If enabled: A request to a directory (`/foo/dir/`) can invoke an archive to be streamed with the query string parameter `archive` set to one of the values in "dir_archives"`: `/foo/dir/?archive=VAL` (where "VAL" is one of "zip", "tar.gz", etc.)

- If enabled: Even if there is no query string parameter, the `Accept` header is still honored: this is for non-browser clients that want a proper HTTP interface.

- If enabled: The default browse template will show a Download link next to directories in the list and at the top, which uses the query string because ... browsers ... yeah. ☹️  Perhaps a little drop-down menu should allow the user to choose which format they want.

- Any compression work is done using the `archiver` package. Archives must be streamed directly to the client, not written (in whole) to disk or memory first.

How does that sound?

**Comment by jaysonsantos** [comment]
Hey there, the PR is already getting a bit big.
What do you think of merging it (if all is good) and we create two issues, one for prettier svg icons and one for the accept?

**Comment by jaysonsantos** [comment]
I can see that there are many failing tests on the build but they are not breaking the status of the PR.
https://dev.azure.com/mholt-dev/Caddy/_build/results?buildId=375&view=ms.vss-test-web.build-test-results-tab&runId=1358&resultId=100053&paneView=debug

**Comment by Mohammed90** [comment]
Those are failing linting preferences. Our setup is to only fail the CI of PR for non-lint matters. Lint is ignored because it's often too pedantic.

**Comment by mholt** [comment]
I hope to review this again this week

File: modules/caddyhttp/fileserver/staticfiles.go [hunk 5]
```diff
@@ -95,6 +96,13 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, _ cadd
 	root := repl.ReplaceAll(fsrv.Root, ".")
 	suffix := repl.ReplaceAll(r.URL.Path, "")
 	filename := sanitizedPathJoin(root, suffix)
+	contentTypes := r.URL.Query()["downloadAs"]
```
mholt: Let's rename it, also use Get:

```go
contentType := r.URL.Query().Get("archive")
```

If the archive type isn't supported, return a HTTP 406 (Not Acceptable). No fallbacks.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -47,6 +47,7 @@ type FileServer struct {
 	IndexNames    []string `json:"index_names,omitempty"`
 	Browse        *Browse  `json:"browse,omitempty"`
 	CanonicalURIs *bool    `json:"canonical_uris,omitempty"`
+	DirArchives   []string `json:"dir_archives,omitempty"`
```
mholt: I think the "archive download" feature should be moved into the Browse configuration, since we require browse to be enabled for this anyway.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -144,7 +152,11 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, _ cadd
 	// to browse or return an error
 	if info.IsDir() {
 		if fsrv.Browse != nil && !fileHidden(filename, filesToHide) {
-			return fsrv.serveBrowse(filename, w, r)
+			if streamAsArchive {
+				return fsrv.streamFolderAsArchive(suffix, filename, contentType, w)
+			} else {
+				return fsrv.serveBrowse(filename, w, r)
+			}
```
mholt: When this feature is moved into the Browse struct, this change will not be necessary; i.e. it will all happen within `serveBrowse()`.

modules/caddyhttp/fileserver/browse.go
```diff
@@ -34,6 +37,18 @@ type Browse struct {
 	template *template.Template
 }
 
+const (
+	zipContentType = "application/zip"
+	zipExtension   = "zip"
+	tarContentType = "application/tar"
+	tarExtension   = "tar"
```
mholt: Probably want `.tar.gz` instead of `.tar` (which also changes the content type)

modules/caddyhttp/fileserver/browse.go
```diff
@@ -146,6 +161,72 @@ func (fsrv *FileServer) browseWriteHTML(listing browseListing) (*bytes.Buffer, e
 	return buf, err
 }
 
+func (fsrv *FileServer) streamFolderAsArchive(baseFolderName, downloadFolderName string, extension string, w http.ResponseWriter) error {
```
mholt: `baseFolderName, downloadFolderName, extension string, ...`

modules/caddyhttp/fileserver/browse.go
```diff
@@ -146,6 +161,72 @@ func (fsrv *FileServer) browseWriteHTML(listing browseListing) (*bytes.Buffer, e
...
+	if err != nil {
+		return err
+	}
+
+	folder, err := os.Open(downloadFolderName)
+	if err != nil {
+		return err
+	}
+	// TODO: This is a POC return a the full contents instead of a shallow copy
+	files, err := folder.Readdir(-1)
```
mholt: Use `filepath.Walk` here instead. This will also greatly simplify your logic.

Here's some code I wrote for another project that streams a folder to an archive writer over HTTP:

```go
dirName := "name of top-level directory"

// ensure HTTP response headers are set properly for a large binary download
w.Header().Set("Content-Type", contentType)
w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="%s.%s"`, dirName, o.format))

// open the archive streaming writer
err := writer.Create(w)
if err != nil {
	return fmt.Errorf("creating archive writer: %v", err)
}
defer writer.Close()

err = filepath.Walk(source, func(fpath string, info os.FileInfo, err error) error {
	if err != nil {
		return err
	}
	if info == nil {
		return fmt.Errorf("nil file info")
	}


	// open the file, if it has any content
	var file io.ReadCloser
	if info.Mode().IsRegular() {
		file, err = os.Open(fpath)
		if err != nil {
			return fmt.Errorf("%s: opening: %v", fpath, err)
		}
		defer file.Close()
	}

	// make its archive-internal name
	internalName, err := archiver.NameInArchive(sourceInfo, source, fpath)
	if err != nil {
		return fmt.Errorf("making internal archive name for %s: %v", fpath, err)
	}

	// write the file to the archive
	err = writer.Write(archiver.File{
		FileInfo: archiver.FileInfo{
			FileInfo:   info,
			CustomName: path.Join(dirName, internalName),
		},
		ReadCloser: file,
	})
	if err != nil {
		return fmt.Errorf("writing file to archive: %v", err)
	}

	return nil
})
if err != nil {
	return fmt.Errorf("walking %s: %v", source, err)
}
```

modules/caddyhttp/fileserver/browse.go
jaysonsantos: Hey there, what about both? Some times it is better to just use tar to avoid the cpu time compressing a gzip afterwards

mholt: I'd be surprised if most people's download speeds were faster than the gzip compression. I'd also be surprised if most people didn't want it to be compressed when they downloaded it.

jaysonsantos: I am more cocerned about cpu usage on the server side and some stuff won't even be compressed like a folder with pictures or movies.

modules/caddyhttp/fileserver/staticfiles.go
jaysonsantos: I used here [1] 415 unsupported media type, don't you think it is more appropriate?
[1] https://github.com/jaysonsantos/caddy/blob/660f8deb2e240741ff998286057252faba3e3cfd/modules/caddyhttp/fileserver/browse.go#L272-L275

modules/caddyhttp/fileserver/staticfiles.go
jaysonsantos: And what about the caddy file?
Would be something like this?

```go
case "dir_archives":
	if fsrv.Browse == nil {
		fsrv.Browse = new(Browse)
	}
	selectedArchives := h.RemainingArgs()
	if err := validateArchiveSelection(selectedArchives); err != nil {
		return nil, err
	}
	fsrv.Browse.DirArchives = selectedArchives
```

jaysonsantos:
I've squashed my commits and rebased with v2 branch

modules/caddyhttp/fileserver/staticfiles.go
mholt: No, because [415 means](https://httpstatuses.com/415):

> The origin server is refusing to service the request because the payload is in a format not supported by this method on the target resource.

Which doesn't apply here. [406 is precisely what we want](https://httpstatuses.com/406):

> The target resource does not have a current representation that would be acceptable to the user agent, according to the proactive negotiation header fields received in the request, and the server is unwilling to supply a default representation.

modules/caddyhttp/fileserver/staticfiles.go
mholt: Sure, yeah! Something like that looks good.

mholt:
Awesome, will take a look at this again soon

jaysonsantos:
Hey there, I just made the move for the folder stream to browse

modules/caddyhttp/fileserver/browse.go
```diff
@@ -45,14 +64,20 @@ func (fsrv *FileServer) serveBrowse(dirPath string, w http.ResponseWriter, r *ht
 		return nil
 	}
 
+	repl := r.Context().Value(caddy.ReplacerCtxKey).(caddy.Replacer)
+	suffix := repl.ReplaceAll(r.URL.Path, "")
```
mholt: I don't think this is what we want... what is this for, exactly?

modules/caddyhttp/fileserver/browse.go
```diff
@@ -45,14 +64,20 @@ func (fsrv *FileServer) serveBrowse(dirPath string, w http.ResponseWriter, r *ht
...
+	contentType := r.URL.Query().Get("archive")
+
+	if contentType != "" {
+		return fsrv.streamFolderAsArchive(suffix, dirPath, contentType, w, r, next)
+	}
```
mholt: Nice, I like this approach a lot better.

modules/caddyhttp/fileserver/browse.go
```diff
@@ -146,6 +171,98 @@ func (fsrv *FileServer) browseWriteHTML(listing browseListing) (*bytes.Buffer, e
...
+
+func (fsrv *FileServer) getArchiveWriter(contentType string) (archiver.Writer, error) {
+	switch contentType {
+	default:
+		return nil, fmt.Errorf("A file format with content type %v is not supported", contentType)
+	case zipContentType:
+		return &archiver.Zip{
+			CompressionLevel:       0,
+			MkdirAll:               true,
+			SelectiveCompression:   false,
```
mholt: I think usually SelectiveCompression is preferable, we should enable this.

modules/caddyhttp/fileserver/browse.go
```diff
@@ -17,23 +17,42 @@ package fileserver
...
+	zipExtension       = "zip"
+	tarContentType     = "application/tar"
+	tarExtension       = "tar"
+	tarGzipContentType = "application/tar+gzip"
+	tarGzipExtension   = "tar.gz"
+)
+
+var (
+	extensionToContentType = map[string]string{zipExtension: zipContentType, tarExtension: tarContentType, tarGzipExtension: tarGzipContentType}
+	contentTypeToExtension = map[string]string{zipContentType: zipExtension, tarContentType: tarExtension, tarGzipContentType: tarGzipExtension}
```
mholt: This appears never to be used? (either that or my browser's Find In Page feature is broken)

modules/caddyhttp/fileserver/command.go
```diff
@@ -53,6 +56,7 @@ respond with a file listing.`,
 			fs.String("root", "", "The path to the root of the site")
 			fs.String("listen", "", "The address to which to bind the listener")
 			fs.Bool("browse", false, "Whether to enable directory browsing")
+			fs.String("dir_archives", "", fmt.Sprintf("Formats to download directories separated by comma. Possible values: %v", reflect.ValueOf(extensionToContentType).MapKeys()))
```
mholt: This should be `dir-achives` (no underscores in CLI flags)

jaysonsantos:
Hey there @mholt sorry for the long time, work is being crazy these days.
I've changed the things you asked and deleted some, i think they were left overs from the refactoring.

mholt:
Thanks for the update! Similarly, I'm taking a break for the holiday but will get back to this soon. Merry Christmas!

mholt:
@jaysonsantos Hey, I'd be able to give this another pass -- sorry for the crazy delay, after getting really sick for a while and then a bunch of other stuff... would you be able to sync up with the v2 branch and then I'll clone this down and try it out for final reviews?

jaysonsantos:
Hey there, I will try do it ASAP

jaysonsantos:
Hey there, I've just rebased it.

francislavoie:
Thanks!

So yesterday, the v2 branch was just renamed to master. Could you rebase onto master? Then you can should be able to change the target branch in the PR to master (use the edit button).

mholt:
Sure, or, I was fine with merging it into v2 and then merging v2 into master. Either way.

jaysonsantos:
Ok, I just rebased onto master

mholt:
Thanks, I will get back to this shortly!

mholt:
Thanks for keeping this updated!

I have cloned it down and tried it out a bit, pretty cool stuff.

I'm a bit nervous to put this v2 this late in the release cycle as I don't want to introduce more surface for bugs. One option is to re-evaluate at around v2.1 or v2.2. But I have an idea that can help us make this available sooner: We might be able to ship this as a separate module (plugin) that either integrates with the file server or comes before it in the handler chain.

I suppose, if we can make the default browse template extensible as well, this should be doable with minimal disruption to the core.

Would you be alright with that instead?

jaysonsantos:
Hey there, whatever you think its best for the project I agree with. :)
Let me know what to change and I will try to do it.

mholt:
Thanks for understanding -- I will revisit this after the 2.0 release.

mholt:
Sorry for the delay. I am going to close this PR, since I don't think we can accept this change into the core of Caddy.

But hold onto the code because I think it could make a great third-party HTTP handler module.

The template integration is nice but not required I think, for this to work; if we need to extend the browse template and/or the file handler, we can discuss that in a separate issue. Thanks again, and thanks for understanding!

Issue #2084, opened 2018-03-26
Extend ipmask directive to error logs


Gunni:
Note that this is a feature request, not an issue or bug.

### 1. What version of Caddy are you using (`caddy -version`)?
Caddy 0.10.11 (unofficial)

### 2. What are you trying to do?
Have consistent logs with no privacy violations.

### 3. What is your entire Caddyfile?
Relevant part:

```
log / stdout "\"when\": \"{when_iso}\", \"id\": \"{request_id}\", \"src_prefix\": \"{remote}\", \"scheme\": \"{scheme}\", \"proto\": \"{proto}\", \"host\": \"{host}\", \"status\": \"{status}\", \"path\": \"{path}\"" {
	ipmask 255.255.0.0 ffff:ffff:ffff:ffff::
}

errors stderr

request_id
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
`/usr/bin/caddy -conf /etc/caddy/caddy.conf -log stdout -root /tmp -quic -agree`

systemd runs it, journald catches stdout and stderr.

### 5. Please paste any relevant HTTP request(s) here.
ssllabs scan, and a normal browser, see logs

### 6. What did you expect to see?
Error logs should filter ips, and to be frank, be formattable...

### 7. What did you see instead (give full error messages and/or log)?
    2018/03/26 03:29:49 http: TLS handshake error from [2001:db8:1020:4202::ac10:826b]:59062: EOF
    "when": "2018-03-26T04:05:54Z", "id": "813a9f4a-ed5d-47ee-adf2-bbde289f59f3", "src_prefix": "2001:db8:58c::", "scheme": "https", "proto": "HTTP/2.0", "host": "example.com", "status": "304", "path": "/"

### 8. How can someone who is starting from scratch reproduce the feature as minimally as possible?
Put the above config on any domain block that doesn't already have them.

mholt:
Thanks for opening the issue!

I'm not sure we can change that particular log example you show, since the Go standard library generates it internally. You should open this issue in https://github.com/golang/go.

I'm afraid that when it comes to the Caddy process log, your best bet for now is to just purge them often if you are required to do that because of privacy laws (make sure they actually apply to you first!).

We can revamp the logging to be entirely under our control and discard any of the Go standard library's logs, but it's not on my roadmap anytime soon, especially not for free. If a business has this need and is willing to pay for it to be expedited, I can get it done.

Gunni:
[...]

Issue #2019, opened 2018-02-08
Syslog - request logs and errors are both logged at error, request logs should be info


chrisvanderpennen:
It looks like in #1301 when the logger for the log and errors directives were merged, the difference in log level between the two was missed and all syslog messages are sent as errors.  This is making it difficult to differentiate between the two types of logs when shipping them to an elk instance for analysis.

### 1. What version of Caddy are you using (`caddy -version`)?
Caddy 0.10.10 (unofficial)

### 2. What are you trying to do?
Log requests and errors to an elk instance via remote syslog, parsing the combined log format into structured data and leaving error messages unmodified

### 3. What is your entire Caddyfile?
```text
https://{$PUBLIC_HOST_NAME}
log / syslog://logstash:5000 "{combined} {latency_ms}"
errors syslog://logstash:5000

[snip 200 lines of reverse-proxy mapping]
```

### 4. How did you run Caddy (give the full command and describe the execution environment)?
Environment: Docker container based on `alpine:3.6` and `abiosoft/caddy:builder`, hosted on ubuntu 16.04
Command line is `/usr/bin/caddy --conf /etc/Caddyfile --log stdout --agree`

### 5. Please paste any relevant HTTP request(s) here.
n/a

### 6. What did you expect to see?
Request logs at severity info, errors at severity error

### 7. What did you see instead (give full error messages and/or log)?
Request logs and errors are both logged at severity error

### 8. How can someone who is starting from scratch reproduce the bug as minimally as possible?
```text
http://localhost
log / syslog "{combined}"
errors syslog

proxy /fail/ host.that.doesnt.exist
```
navigate to http://localhost/fail/

mholt:
@wendigo What do you think about this? Is it possible to differentiate the two levels per-directive for the syslog?

@chrisvanderpennen For what it's worth, Caddy doesn't have "log levels" officially; on some messages we control we usually output "ERROR" or "INFO" or whatever, but we'll have to see if it's feasible to separate these with our syslog package.

chrisvanderpennen:
[...]

PR #2831, opened 2019-10-24
v2: Logging!

### LINKED ISSUE #425: Is there a way caddy can write access logs in JSON format?
For example as below - unmarshalling this type of struct? 

```
   // HTTP detailed message
        HTTP struct {
                ResponseHeaders http.Header
                Request         struct {
                        Method     string
                        URL        *url.URL
                        Proto      string // "HTTP/1.0"
                        ProtoMajor int    // 1
                        ProtoMinor int    // 0
                        Header     http.Header
                        Host       string
                        Form       url.Values
                        PostForm   url.Values
                        Trailer    http.Header
                        RemoteAddr string
                        RequestURI string
                }
        }
```

```
{"StartTime":"2015-06-17T03:43:22.289131896Z","Duration":457855,"StatusMessage":"OK","ContentLength":"0B","HTTP":{"ResponseHeaders":{"Accept-Ranges":["bytes"],"Connection":["close"],"Content-Length":["0"],"Content-Type":["application/xml"],"Server":["Minio"]},"Request":{"Method":"HEAD","URL":{"Scheme":"","Opaque":"","User":null,"Host":"","Path":"/updates","RawQuery":"","Fragment":""},"Proto":"HTTP/1.1","ProtoMajor":1,"ProtoMinor":1,"Header":{"Authorization":["AWS4-HMAC-SHA256 Credential=AC5NH40NQLTL4D2W92PM/20150617/milkyway/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=70bbb1cf8f133b1fd33eca46d77859527b4bb267208856f1b7322f75e2565d92"],"User-Agent":["Minio/Fri, 12 Jun 2015 02:05:16 GMT (/home/minio/go/bin/mc; linux; amd64)  minio-go/0.1.0 (linux; amd64)"],"X-Amz-Content-Sha256":["e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"],"X-Amz-Date":["20150617T034322Z"]},"Body":{"Closer":{"Reader":null}},"ContentLength":0,"TransferEncoding":null,"Close":false,"Host":"dl.minio.io:9000","Form":null,"PostForm":null,"MultipartForm":null,"Trailer":null,"RemoteAddr":"10.134.253.170:51043","RequestURI":"/updates","TLS":{"Version":771,"HandshakeComplete":true,"DidResume":false,"CipherSuite":49199,"NegotiatedProtocol":"","NegotiatedProtocolIsMutual":true,"ServerName":"dl.minio.io","PeerCertificates":null,"VerifiedChains":null,"TLSUnique":"GRVvG32qlXzFR4Lp"}}}}
```

If not possible does caddy provide support for external middleware / handlers? - I couldn't see much in documentation.

### LINKED ISSUE #1109: Generate json log
It would be gread if caddy could generate a full json like this example from kibana:

```
{
 "type": "response",
 "@timestamp": "2016-09-15T20:26:54Z",
 "tags": [],
 "pid": 12,
 "method": "post",
 "statusCode": 200,
 "req": {
  "url": "\/elasticsearch\/_msearch?timeout=0&ignore_unavailable=true&preference=1473971177442",
  "method": "post",
  "headers": {
   "host": "kibana.example.loc",
   "user-agent": "Mozilla\/5.0 (X11; Linux x86_64) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/52.0.2743.82 Safari\/537.36",
   "content-length": "789",
   "accept": "application\/json, text\/plain, *\/*",
   "accept-language": "de-DE,de;q=0.8,en-US;q=0.6,en;q=0.4",
   "content-type": "application\/json;charset=UTF-8",
   "kbn-version": "4.6.1",
   "origin": "http:\/\/kibana.example.loc",
   "referer": "http:\/\/kibana.example.loc\/app\/kibana",
   "x-forwarded-for": "172.18.0.1",
   "x-forwarded-proto": "http",
   "x-real-ip": "172.18.0.1",
   "accept-encoding": "gzip"
  },
  "remoteAddress": "172.18.0.14",
  "userAgent": "172.18.0.14",
  "referer": "http:\/\/kibana.example.loc\/app\/kibana"
 },
 "res": {
  "statusCode": 200,
  "responseTime": 249,
  "contentLength": 9
 },
 "message": "POST \/elasticsearch\/_msearch?timeout=0&ignore_unavailable=true&preference=1473971177442 200 249ms - 9.0B"
}
```

That would really simplify our logstash rules...

### LINKED ISSUE #2678: v2: Logging
Version 2 needs a flexible logging module.

Main goals:

- Performance: zero allocation loggers would be ideal in the hot paths. Consider using https://github.com/uber-go/zap
- Structure: logs should be programmable.
- Modular backends: logs should be able to go to anywhere, whether it be a file, stdout/stderr, a system log, or over the network to some third-party logging service.
- Customizable output formats: logs should be able to be customized as to how they look and what information they include; for example the standard Apache log format should be possible as well as JSON or binary log formats.
- Customizable emissions: choose specifically what gets emitted, even beyond just log levels. Do we want to emit errors about clients closing a connection, or just ACME certificate errors instead?
- Standard process log: most/all dependencies use the standard logger or won't integrate specifically with our logging system, so we need a way to dump those (secondary) logs to a file or a stream of the user's choice.
- Reduce repetition considering tracing/metrics/telemety: if logs use logic similar to tracing/metrics (which tracks a connection or request through various handlers / parts of the config) or telemetry (which helps understand global caddy usage), it should reuse the same code or infrastructure when possible. (If we add telemetry, it will be a very watered-down version of what Caddy 1 had, which was too expensive to maintain.)

mholt:

This adds logging functionality to Caddy 2.

Ahem. That is an understatement. Let me try again:

This adds **really awesome, fun-to-use, super flexible, very powerful, high performance logging** to Caddy 2.

I have [already started adding logging to the v2 docs](https://github.com/caddyserver/caddy/wiki/v2:-Documentation#logging).

- Writers are modular, so you can emit logs to your favorite destinations: console, files (with or without automatic rolling), syslog, external services, etc.
- Encoders are also modular, so you can emit logs in your favorite formats: JSON, human-readable (with or without coloring), logfmt, common log format, etc.
- Multiple logs are supported.
- Each log can filter which Caddy modules are allowed to write to them.
- Each log can also filter individual fields, including in a nested structure. So you can mask IP addresses or remove passwords before they hit certain logs.
- We use [zap](https://github.com/uber-go/zap) under the hood, so all logs are structured and high-performing (most should be zero-alloc).
- Because we use zap, we also get log sampling for free, so that is configurable too.
- The "sink" logger (what I call the default global logger of Go's `log` package) can be redirected as well, but without all the flexibility of course. Basically the "sink" is where "everything else" goes, or log emissions that use `log.Print` from any code that wasn't designed to be used with Caddy. But all Caddy modules (including Caddy's core) should emit using Caddy's logger. Docs are forthcoming on that.

Still TODO:

- [x] Documentation
- [x] General polish
- [x] Fine-tune the defaults. For example, do we assume production or development environment by default? How accessible/human-readable do we want it to be out-of-the-box, vs. performant?
- [x] HTTP access and error logging
- [x] Change all of Caddy's `log.Print*()` calls to use its shiny new `Log()` logger.

Closes #1109
Fixes #425
Closes #2678
Related to #2019
#2084 too, I think

Will add a hecka lot of JSON and explanatory paragraphs to a new section of the docs just about logging.


tobya:
Fabulous.  This seems to be the  system log.  Will the access log use the same system?  Is the access log configurable yet?

mholt:
@tobya Good question, here's how it works:

You define logs centrally in the top-level `logging` property of Caddy's config. There's the "default" log implicitly, which you can customize, and you can add any others by your own name. A log consists of a writer, encoder, and some other settings such as log levels to allow, sampling, and including/excluding specific Caddy modules.

By default, all log emissions go to all defined logs. That includes logs emitted by individual Caddy modules such as the HTTP log handler (still WIP, I should be able to push it today).

Caddy modules, including the HTTP log handler, have the ability to allow a custom log to be defined in-line with it, like this:

```json
{
    "handler": "log",
    "logs": [
        {
            "writer": {},
            "encoder": {}
        }
    ]
}
```

This is useful when you have multiple instances of the log handler, for example, and you want each one to be configured differently. In the case where an individual module allows its own log configuration, it's up to the module whether to broadcast the logs to all loggers or to use only the locally-defined log.

So basically, by default logs are "broadcast" (for lack of a better word) which means that all logs go to all defined logs. And by default, there's only 1 log. So by default, everything goes into 1 place and has the same encoding (format).

From there, it's up to you to configure how many logs you want, where you want them to go, what you want them to look like, what you want them to include, etc.

mholt:
And that's a wrap.

Geeeeez, that was harder than I thought it would be. 😅But so worth it. Have fun with this!

The transition to using the new loggers (instead of the global standard "sink" logger) will be happening gradually over the next few beta releases.

We'll also be fine-tuning logging according to experience and feedback. Once we reach stable, we can't change some stuff, so try it out ASAP!

The docs are updated too.

PR #2834, opened 2019-10-26
admin listener as optin for initial config

### LINKED ISSUE #2833: Options to disable admin listener

Under some circumstances, the admin listener might cause a problem or is simply not necessary. If the dynamic configuration at runtime via HTTP is not needed, it would be good to be able to disable the admin port altogether.

I would even propose that if caddy runs on Caddyfile, the admin port is disabled by default (to be closer to caddy v1) and introduce a Caddyfile directive to easily enable and configure the admin port if needed.

I think there are two use cases for caddy 2: professional hosting, reverse proxying, etc. where you greatly benefit from dynamic configuration you can read/write during runtime using a HTTP listener.

Then there is also the "simple user" that only needs a quick or easy webserver. This is also where the Caddyfile shines and what (for me) always is/was the great benefit of caddy in contrast to its competitors. In this case, an additional port is only another potential attack vector (since machines can be shared) and is another port that could conflict with other running services. Also in these cases there is simply no gain from the admin feature, since all configuration is done through the Caddyfile anyway.

So far I stay with caddy 1 to keep it simple for myself.

## Remarks

This ticket is mostly intended for discussion. If you / someone jumps on that train right away and implements it I would not be sad, of course. But if the proposal is considered valid, I would also be willing to implement it and create a PR.

aksdb:

After these changes, the admin endpoint is no longer enabled by default if a configuration file is specified. Instead, the admin listener needs to be configured explicitly.

For the `Caddyfile` this has been simplified with the "admin" option. Without parameters, it will use the (current) default listen address. If a parameter is given, it is treated as listener address.

Example:

```
{
    admin
}
```

Example 2:

```
{
    admin 0.0.0.0:1234
}
```

Closes #2833

* It needs to be stated that `run` only initializes the admin listener if no config file is given.
* The admin endpoint documentation needs to mention that if a config file is given, the listen address has to be specified for the admin endpoint to be available.
* The `Caddyfile` documentation needs to be enhanced to mention the new `admin [listen address]` option (see examples above).

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -148,7 +148,7 @@ func StartAdmin(initialConfigJSON []byte) error {
 
 	go cfgEndptSrv.Serve(ln)
 
-	fmt.Println("Caddy 2 admin endpoint listening on", adminConfig.Listen)
+	Log().Named("admin").Info("Caddy 2 admin endpoint started.", zap.String("listenAddress", adminConfig.Listen))
 
 	return nil
 }
--- a/admin.go
+++ b/admin.go
@@ -150,14 +150,6 @@ func StartAdmin(initialConfigJSON []byte) error {
 
 	fmt.Println("Caddy 2 admin endpoint listening on", adminConfig.Listen)
 
-	if len(initialConfigJSON) > 0 {
-		err := Load(bytes.NewReader(initialConfigJSON))
-		if err != nil {
-			return fmt.Errorf("loading initial config: %v", err)
-		}
-		fmt.Println("Caddy 2 serving initial configuration")
-	}
-
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


mholt:
Awesome! I am looking forward to reviewing this :)

I'm super impressed you figured out how to integrate it into the Caddyfile...

aksdb:
I will rebase my branch tomorrow and adjust the logging 🙂

aksdb:
@mholt Done. Btw I like that you use `zap`. I love that logger; use it where ever I can.

mholt:
Excellent, thank you!! I appreciate that you tidied up the commits and paid attention to details.

Looking forward to reviewing with you again. Would you be willing to contribute more? If so, I can make you a collaborator so you can work on a branch instead of a fork.

PR #2862, opened 2019-11-05
update quic-go to v0.13.0

### LINKED ISSUE #2450: Upgrade QUIC to HTTP/3
You know, HTTP/3 is the new name for QUIC, also (previously?) called "HTTP over QUIC" as it seems.

AFAIK it is not yet standardized and no one supports it yet, (seems there are still some technical differences to QUIC, see test site below) but it may soon… :smile: 
And Caddy can be one of the first to support it. :smiley: 

There is even already a test site:
https://http3.info/

marten-seemann:
Fixes #2450.

This version supports IETF QUIC draft-23.


Henrocker:
Having this in v2 would be nice, too!

Henrocker:
Have a good sleep :-)

chaosong:
Is there a release or an official build binary working for this? My build is not working by using the master code according to http3check.net. The result is 'QUIC is supported', no http3 result

PR #2866, opened 2019-11-08
fix(systemd): update caddy.service on Ubuntu 18

yuqianma:

Move the config field in this script to the right part.
i.e.
move `StartLimitIntervalSec` and `StartLimitBurst` from [Service] to [Unit]
ref: https://lists.freedesktop.org/archives/systemd-devel/2017-July/039255.html
systemd version: `systemd 237`
(You can check with `man systemd.directives`, mine is `systemd 237`)

I am not sure if it is valid with old ubuntu and others.
Maybe we can mention it in the document, and keep a new version script?

none

Maybe, need a further discuss.

## 4. Checklist

### DIFF
--- a/dist/init/linux-systemd/caddy.service
+++ b/dist/init/linux-systemd/caddy.service
@@ -4,13 +4,13 @@ Documentation=https://caddyserver.com/docs
 After=network-online.target
 Wants=network-online.target systemd-networkd-wait-online.service
 
-[Service]
-Restart=on-abnormal
-
 ; Do not allow the process to be restarted in a tight loop. If the
 ; process fails to start, something critical needs to be fixed.
 StartLimitIntervalSec=14400
 StartLimitBurst=10
+
+[Service]
+Restart=on-abnormal
 
 ; User and group the process will run as.
 User=www-data


mholt:
Thanks for the PR.

@hairyhenderson @Ilyes512 @carlwgeorge  Is this something we need to consider for our new packages? I wasn't aware that systemd made this change.

Ilyes512:
Not relevant to the Docker image.

(a services manager like systemd should not be used because you want the container to crash when the (main) process crashes. Restarting is handled by Docker.)

fab120:
@mholt According to systemd documentation,
**StartLimitIntervalSec** and **StartLimitBurst** are under the [Unit] section.

[Reference](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval) (You can see those config are under *[Unit] Section Options*).

carlwgeorge:
The [unit file in caddyserver/dist](https://github.com/caddyserver/dist/blob/master/init/caddy.service) isn't using those directives currently.  Last we talked about it I believe the conclusion was that automatically restarting was not desirable.  If the process fails something is wrong and it should be investigated.

These directives also apply to manual restarts, not just automatic ones.  The default is to allow five restarts in ten seconds.  What is the benefit of switching that to ten restarts in four hours?

mholt:
@carlwgeorge Ah yes, you are right -- and I think we should keep our new unit file as the authoritative one at this point.

> What is the benefit of switching that to ten restarts in four hours?

I believe this had something to do with misuse from a certain deployment that caused a certain CA to be hammered pretty significantly.

PR #2872, opened 2019-11-09
Add support for placeholders in some logging config

### LINKED ISSUE #2870: Logging configuration should support placeholders

The following config.json won't work (`run: loading initial config: loading new config: setting up default log: unrecognized log level: {env.LOG_LEVEL}`), but it would be nice if it did:

```json
{
  "logging": {
    "sink": {
      "writer": {}
    },
    "logs": {
      "default": {
        "level": "{env.LOG_LEVEL}",
        "encoder": {
          "format": "json"
        }
      }
    }
  },
  "apps": {
    "http": {
      "servers": {
        "server-1": {
          "listen": [":443"],
          "routes": [
            {
              "match": [
                {
                  "host": ["localhost"]
                }
              ],
              "handle": [
                {
                  "handler": "reverse_proxy",
                  "upstreams": [
                    {
                      "dial": "localhost:8080"
                    }
                  ]
                }
              ]
            }
          ]
        }
      }
    }
  }
}
```

Especially when running in Docker or Kubernetes, it's common to configure applications via environment variables. Having the option to have a common base configuration which works as-is, but is still customizable, is very useful. One application could set `LOG_LEVEL=debug` and another `LOG_LEVEL=info`, but both could use the same config.json.

For the same reason, it could also be useful to support placeholders in the `filename` option of the `file` output module.

There might be other fields where placeholders can be useful, but those are the most useful ones, in my opinion.

No config.json reuse, each application has their own

## 4. Please link to any relevant issues, pull requests, or other discussions.

--

iamd3vil:
Fixes #2870 

Adds support for placeholders in config file.

#2870 

The config file can have placeholders for any environment variables like `{env.LOG_LEVEL}`.

## 4. Checklist

- [X] This change has comments explaining package types, values, functions, and non-obvious lines of code

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -280,8 +280,15 @@ type CustomLog struct {
 }
 
 func (cl *CustomLog) provision(ctx Context, logging *Logging) error {
+	// Replace placeholder for log level
+	repl := NewReplacer()
+	level, err := repl.ReplaceOrErr(cl.Level, true, true)
+	if err != nil {
+		return fmt.Errorf("invalid log level: %v", err)
+	}
+
 	// set up the log level
-	switch cl.Level {
+	switch level {
 	case "debug":
 		cl.levelEnabler = zapcore.DebugLevel
 	case "", "info":
@@ -351,7 +358,7 @@ func (cl *CustomLog) provision(ctx Context, logging *Logging) error {
 	if cl.writerOpener == nil {
 		cl.writerOpener = StderrWriter{}
 	}
-	var err error
+
 	cl.writer, _, err = logging.openWriter(cl.writerOpener)
 	if err != nil {
 		return fmt.Errorf("opening log writer using %#v: %v", cl.writerOpener, err)
--- a/modules/logging/filewriter.go
+++ b/modules/logging/filewriter.go
@@ -47,12 +47,21 @@ func (FileWriter) CaddyModule() caddy.ModuleInfo {
 	}
 }
 
-func (fw FileWriter) String() string {
+// Provision sets up the module
+func (fw *FileWriter) Provision(ctx caddy.Context) error {
 	// Replace placeholder in filename
 	repl := caddy.NewReplacer()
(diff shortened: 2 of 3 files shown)


logging.go
```diff
@@ -280,8 +280,15 @@ type CustomLog struct {
 }
 
 func (cl *CustomLog) provision(ctx Context, logging *Logging) error {
+	// Replace placeholder for log level
+	repl := NewReplacer()
+	level, err := repl.ReplaceOrErr(cl.Level, true, true)
+	if err != nil {
+		return fmt.Errorf("invalid log level: %v", err)
+	}
+
```
mholt: This part looks great! :+1:

modules/logging/filewriter.go
```diff
@@ -47,7 +48,11 @@ func (FileWriter) CaddyModule() caddy.ModuleInfo {
 }
 
 func (fw FileWriter) String() string {
-	fpath, err := filepath.Abs(fw.Filename)
+	// Replace placeholder in filename
+	repl := caddy.NewReplacer()
+	filename := repl.ReplaceKnown(fw.Filename, "")
```
mholt: This is a good idea, to support placeholders in log file names.

But we should do the replacement only once.

All Caddy modules go through a provisioning phase before they are used: https://github.com/caddyserver/caddy/wiki/v2:-Writing-a-Module#provisioning

So, I would:

1) make a `Provision()` method that implements the `caddy.Provisioner` interface,
2) set an [interface guard](https://github.com/caddyserver/caddy/wiki/v2:-Writing-a-Module#interface-guards) - (see examples in other files, you can do a search in the code for "interface guard")
3) in the `Provision()` method, do the replacement there.

This way, we can use `fw.Filename` just as before, because it will have already been evaluated for replacements during provisioning.

iamd3vil:
Thanks for being patient and welcoming @mholt :heart:

Issue #2899, opened 2019-11-26
v2: query matcher


sarge:
Edit: removed "quotes" from query matcher as quotes in the config are not recommended.

## 1. Which version of Caddy are you using (`caddy -version`)?
v2 beta 10

## 2. What are you trying to do?
I am attempting to use a query parameter from the url and rewrite the request

```
matcher debug {
    query show=1
}

rewrite match:debug /debug/index.html
```

Extension: I would also like to add an extension to the query matcher to allow wildcards, this would allow matching on just the show parameter the value is irrelevant.

```
matcher debug {
    query show=*
}

rewrite match:debug /debug/index.html
```

Extension: use case 2 - passing a variable upstream if it exists

```
matcher debug {
    query version=*
}

rewrite match:debug /debug/{http.request.uri.query.version}/index.html
```

## 3. What is your Caddyfile?
```text
http://localhost:8099
tls off
root * . 

matcher debug {
  query "showdebug=1"
}

# redirect to the specified installer version
rewrite match:debug /debug/index.html

file_server
```

## 4. How did you run Caddy (give the full command and describe the execution environment)?
Fedora 31

caddy run --config Caddyfile2 --adapter caddyfile

## 5. Please paste any relevant HTTP request(s) here.
N/A

## 6. What did you expect to see?
The Caddyfile should  parse

## 7. What did you see instead (give full error messages and/or log)?
2019/11/27 09:52:32 run: adapting config using caddyfile: Caddyfile2:6 - Error during parsing: malformed query matcher token: query; must be in param=val format

## 8. Why is this a bug, and how do you think this should be fixed?
The syntax is valid

## 9. What are you doing to work around the problem in the meantime?
Avoid using the query matcher

## 10. Please link to any related issues, pull requests, and/or discussion.
MR to come :)

## Bonus: What do you use Caddy for? Why did you choose Caddy?
Clean extension model

francislavoie:
Did you try without quotes? Reading the docs leads me to believe quotes should be excluded.

sarge:
@francislavoie Thanks, I have just rechecked and it also fails without quotes. I have fixed the parsing locally, which does trigger another bug. So I am pretty sure no one is using the caddyfile with a query matcher.

mholt:
Aha, indeed, I have yet to write tests for most (all?) of the matchers -- would definitely love to accept and review pull requests for this! Thanks!

sarge:
@mholt I will make a start on the tests.
[...]

PR #2901, opened 2019-11-27
v2: fixes query matcher parsing

sarge:
---
name: Pull request
about: This addresses a bug in the caddy file parser when parsing query matchers
title: ''
labels: ''
assignees: ''
---

As discussed in #2899 

I have had a go at what a set of parser tests could look like. Open to suggestions about getting the state of a particular parsed entity and verifying it contains the parsed data.

#2899

https://github.com/caddyserver/caddy/wiki/v2:-Documentation#matcher

> query <key=val...>

becomes

> query <key=[val...]>  | val can be the value to match or the wildcard '*' to match any value.

 

## 4. Checklist

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -241,8 +241,16 @@ func (MatchQuery) CaddyModule() caddy.ModuleInfo {
 
 // UnmarshalCaddyfile implements caddyfile.Unmarshaler.
 func (m *MatchQuery) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	if *m == nil {
+		*m = make(map[string][]string)
+	}
+
 	for d.Next() {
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/matchers.go
```diff
@@ -241,8 +241,16 @@ func (MatchQuery) CaddyModule() caddy.ModuleInfo {
...
 func (m *MatchQuery) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	if *m == nil {
+		*m = make(map[string][]string)
+	}
+
 	for d.Next() {
-		parts := strings.SplitN(d.Val(), "=", 2)
+		var query string
+		if !d.Args(&query) {
+			return d.Errf("malformed query matcher token: %s; must be in param=val format", d.Val())
```
mholt: A more proper error might be `return d.ArgErr()`, since the argument is missing.

PR #2909, opened 2019-12-02
v2: add remaining subdirectives to rewrite directive

bemasher:
This PR adds the remaining subdirectives for rewrite to match the full JSON config. The 'to' argument has been made optional if subdirectives are also provided.

N/A

The Caddyfile documentation for rewrite will need to be updated to include the new subdirectives and a statement that the 'to' argument is optional if subdirectives are also provided.

https://github.com/caddyserver/caddy/wiki/v2:-Documentation#rewrite

## 4. Checklist

I am not sure where tests for this directive or the changes belong. Let me know and I will work on them.

### DIFF
--- a/modules/caddyhttp/rewrite/caddyfile.go
+++ b/modules/caddyhttp/rewrite/caddyfile.go
@@ -15,8 +15,6 @@
 package rewrite
 
 import (
-	"strconv"
-
 	"github.com/caddyserver/caddy/v2/caddyconfig/httpcaddyfile"
 	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
 )
@@ -27,8 +25,8 @@ func init() {
 
 // parseCaddyfile sets up the handler from Caddyfile tokens. Syntax:
 //
-//     rewrite [<matcher>] [<to>] {
-//         to                <string>
+//     rewrite [<matcher>] [<uri>] {
+//         uri                <string>
 //         method            <string>
 //         strip_path_prefix <string>
 //         strip_path_suffix <string>
@@ -40,7 +38,7 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 	var (
 		rewr Rewrite
 
-		toIsSet        bool
+		uriIsSet       bool
 		nextBlockIsSet bool
 		rehandleIsSet  bool
 		redirectIsSet  bool
@@ -51,7 +49,7 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 		switch len(args) {
 		case 0:
 		case 1:
-			toIsSet = true
+			uriIsSet = true
 			rewr.URI = h.Val()
 		default:
 			return nil, h.ArgErr()
@@ -60,9 +58,9 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 		for h.NextBlock(0) {
 			nextBlockIsSet = true
 			switch h.Val() {
-			case "to":
-				if toIsSet {
-					return nil, h.Err("to is already set")
+			case "uri":
+				if uriIsSet {
+					return nil, h.Err("uri is already set")
 				}
 				if !h.Args(&rewr.URI) {
 					return nil, h.ArgErr()
@@ -71,17 +69,11 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 				if !h.Args(&rewr.Method) {
 					return nil, h.ArgErr()
 				}
-				switch rewr.Method {
-				case "GET", "HEAD", "POST", "PUT", "DELETE",
-					"CONNECT", "OPTIONS", "TRACE", "PATCH":
-				default:
-					return nil, h.Errf("unknown method '%s'", rewr.Method)
-				}
-			case "strip_path_prefix":
+			case "strip_prefix":
 				if !h.Args(&rewr.StripPathPrefix) {
 					return nil, h.ArgErr()
 				}
-			case "strip_path_suffix":
+			case "strip_suffix":
 				if !h.Args(&rewr.StripPathSuffix) {
 					return nil, h.ArgErr()
 				}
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -25,20 +27,116 @@ func init() {
...
+				}
+			case "method":
+				if !h.Args(&rewr.Method) {
+					return nil, h.ArgErr()
+				}
+				switch rewr.Method {
+				case "GET", "HEAD", "POST", "PUT", "DELETE",
+					"CONNECT", "OPTIONS", "TRACE", "PATCH":
+				default:
+					return nil, h.Errf("unknown method '%s'", rewr.Method)
```
mholt: It might not be a bad idea to allow arbitrary methods in case some upstream app requires them, for example, WebDAV...

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -25,20 +27,116 @@ func init() {
...
+				if !h.Args(&rehandle) {
+					return nil, h.ArgErr()
+				}
+
+				switch rehandle {
+				case "true":
+					rewr.Rehandle = true
+				case "false":
+				default:
+					return nil, h.Errf("invalid rehandle argument '%s', expected bool", rehandle)
```
mholt: For booleans in the Caddyfile, I like to simply use the presence of the subdirective to indicate a true value.

mholt: I actually wonder if we need to / should expose this in the Caddyfile at all, since rehandling is kind of a detail of the underlying JSON routes, which is abstracted away from the user...

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -25,20 +27,116 @@ func init() {
...
 // parseCaddyfile sets up the handler from Caddyfile tokens. Syntax:
 //
-//     rewrite [<matcher>] <to>
+//     rewrite [<matcher>] [<to>] {
+//         to                <string>
+//         method            <string>
+//         strip_path_prefix <string>
+//         strip_path_suffix <string>
+//         rehandle          <bool>
+//         http_redirect     <status_code>
```
mholt: I am wondering if it is useful to expose the http_redirect feature in the Caddyfile...

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -25,20 +27,116 @@ func init() {
 
 // parseCaddyfile sets up the handler from Caddyfile tokens. Syntax:
 //
-//     rewrite [<matcher>] <to>
+//     rewrite [<matcher>] [<to>] {
+//         to                <string>
```
mholt: I wonder if "to" should be called "uri" since that makes it more consistent with the "method" subdirective, which rewrites the method.

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -25,20 +27,116 @@ func init() {
...
+//         method            <string>
+//         strip_path_prefix <string>
+//         strip_path_suffix <string>
```
mholt: If "path" is obvious here, I wonder if we could omit that, to make this easier to write?

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -25,20 +27,116 @@ func init() {
...
+				}
+
+				statusNum, err := strconv.Atoi(status)
+				if err != nil {
+					return nil, h.Errf("bad status value '%s': %v", status, err)
+				}
+
+				if statusNum < 300 || statusNum > 399 {
+					return nil, h.Errf("bad status value '%s', must be in 3xx range", status)
+				}
```
mholt: (If we keep this feature)

Is the status code ever assigned to `rewr`? (Also, status code validation will happen shortly after this, as the underlying module is loaded and used, so we don't necessarily have to validate it again here in the Caddyfile.)

modules/caddyhttp/rewrite/caddyfile.go
bemasher: That's an easy enough fix to check for presence.

In my case I need to prevent it from re-handling rewrites so an upstream system can be served under a path prefix, and strip the prefix from requests before sending them on their way. Unless there's a more appropriate way to handle that.

modules/caddyhttp/rewrite/caddyfile.go
bemasher: There is already the `redir` directive, so I'm not sure how useful it is to have it here as well.

modules/caddyhttp/rewrite/caddyfile.go
bemasher: That makes more sense to me too.

modules/caddyhttp/rewrite/caddyfile.go
bemasher: `rewr.HTTPRedirect` definitely never got set after all that checking.

modules/caddyhttp/rewrite/caddyfile.go
mholt: Ah. See #2911 -- I think that we'll turn off rehandling by default and make it opt-in...

modules/caddyhttp/rewrite/caddyfile.go
mholt: Let's remove it then, if that's OK with you. We'll expose as little as possible until it is needed.

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -25,20 +25,99 @@ func init() {
...
+				if !h.Args(&rewr.Method) {
+					return nil, h.ArgErr()
+				}
+			case "strip_prefix":
+				if !h.Args(&rewr.StripPathPrefix) {
+					return nil, h.ArgErr()
+				}
+			case "strip_suffix":
+				if !h.Args(&rewr.StripPathSuffix) {
+					return nil, h.ArgErr()
```
mholt: I had a second thought, that perhaps `strip_prefix` and `strip_suffix` should be separate directives -- especially since they would conflict with the "uri" parameter, I think.

Would that be OK with you? Making them separate directives is suuuper easy... and I think having more directives that do fewer things is good, especially for nuanced stuff like rewriting.

bemasher: Sounds reasonable to me.

modules/caddyhttp/rewrite/caddyfile.go
bemasher: Will do.

modules/caddyhttp/rewrite/caddyfile.go
bemasher: I'll change it to check for presence only. I think the `route` directive would fit my use case.

bemasher:
@mholt: Do you have a preferred place to test these changes? How about a template for testing caddyfile directives?

mholt:
I think just the usual tests in a `Test*` function in a `_test.go` file would be just fine!

mholt:
Sorry about the delay on this PR. I had to address some underlying design issues which I am _hoping_ are mostly fixed now, but it involved some semi-significant changes like getting rid of rehandling entirely (I'm actually glad for that).

Now, the rewrite directive could still use `method` (which your PR adds), strip_path_prefix, `strip_path_suffix`, and `uri_replace`.

If you have a chance, that'd be great -- otherwise, I can take care of it eventually.

PR #2913, opened 2019-12-02
ability to specify that client cert must be present in SSL

evtr:
---
name: Pull request
title: ''
labels: ''
assignees: ''
---

Ability to specify in tls connection policy that a client SSL cert be present.

https://caddy.community/t/v2-client-auth/6581/2

https://github.com/caddyserver/caddy/wiki/v2:-Documentation#httpserverstls_connection_policies need to be updated to include
"require" boolean "requires a client certificate, but will not verify it"

## 4. Checklist

- [x ] I have written tests and verified that they fail without my change
- [ x] This change has comments explaining package types, values, functions, and non-obvious lines of code

### DIFF
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -258,7 +258,7 @@ type ClientAuthentication struct {
 
 // Active returns true if clientauth has an actionable configuration.
 func (clientauth ClientAuthentication) Active() bool {
-	return len(clientauth.TrustedCACerts) > 0 || len(clientauth.TrustedLeafCerts) > 0 || len(clientauth.ClientAuthType) > 0
+	return len(clientauth.TrustedCACerts) > 0 || len(clientauth.TrustedLeafCerts) > 0 || len(clientauth.Mode) > 0
 }
 
 // ConfigureTLSConfig sets up cfg to enforce clientauth's configuration.
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -244,9 +244,12 @@ type ClientAuthentication struct {
 	// which are not in this list will be rejected.
 	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
 
-	// requires a client certificate, but will not verify it
-	ClientAuthType string `json:"client_auth_type,omitempty"`
-
+	// The mode of the client authentication - the allowed values are
+	// 'request'		- A client certificate is retrieved if available (but is not otherwise verified), if no client cert is presented this is also ok
+	// 'require'		- A client certificate must be presented (but is not otherwise verified)
+	// 'verify_if_given'	- Verify the presented client certificate if a client cert is presented, if no client cert is presented this is also ok
+	// 'require_and_verify'	- Verify the presented client certificate. A client certificate must be presented
+	Mode string `json:"mode,omitempty"`
 
 	// state established with the last call to ConfigureTLSConfig
 	trustedLeafCerts       []*x509.Certificate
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the PR! I'm starting to get caught up so I'll look at this soon.

mholt:
Thanks for the PR!

I wonder if a better approach is to have different modes of enabling Client Auth, like what the Go standard library has here: https://golang.org/pkg/crypto/tls/#ClientAuthType

```
NoClientCert
RequestClientCert
RequireAnyClientCert
VerifyClientCertIfGiven
RequireAndVerifyClientCert
```

evtr:
Absolutely...I will fix it.

/Eva

Sendt fra min iPhone

evtr:
Left out the 'NoClientCert' as i didnt think it made much sense.
In the config you should specify the values for the clientauthtype in the same way as in caddy1 https://caddyserver.com/v1/docs/tls

mholt:
Awesome, I like where this is going. I will take a closer look after the holiday. Merry Christmas!

mholt:
Going to look at this again after Go 1.14 comes out, just in case we can benefit from anything there.

modules/caddytls/connpolicy.go
```diff
@@ -244,14 +244,18 @@ type ClientAuthentication struct {
 	// which are not in this list will be rejected.
 	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
 
+	// requires a client certificate, but will not verify it
+	ClientAuthType string `json:"client_auth_type,omitempty"`
+
+
```
mholt: Remove extra line

modules/caddytls/connpolicy.go
```diff
@@ -244,14 +244,18 @@ type ClientAuthentication struct {
 	// which are not in this list will be rejected.
 	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
 
+	// requires a client certificate, but will not verify it
```
mholt: Flush out this godoc comment more. Will need much more explanation! What does it do? What are the possible values? What do they do? I can write this if you want a suggestion.

modules/caddytls/connpolicy.go
```diff
@@ -244,14 +244,18 @@ type ClientAuthentication struct {
...
+	ClientAuthType string `json:"client_auth_type,omitempty"`
```
mholt:
```suggestion
	Mode string `json:"mode,omitempty"`
```

Since we're already in the ClientAuthentication struct, the fact that it's Client Auth is obvious, and doesn't need to be repeated, I think. I also think of this as an operational mode than a type. Tbh I'm not a huge fan of the std lib's client auth configuration...

modules/caddytls/connpolicy.go
```diff
@@ -262,8 +266,29 @@ func (clientauth *ClientAuthentication) ConfigureTLSConfig(cfg *tls.Config) erro
...
+				cfg.ClientAuth = tls.RequireAnyClientCert
+			}
+			case "verify_if_given": {
+				cfg.ClientAuth = tls.VerifyClientCertIfGiven
+			}
+			case "require_and_verify": {
+				cfg.ClientAuth = tls.RequireAndVerifyClientCert
+			}
+			default: {
+				return fmt.Errorf("client_auth_type %s not allowed", clientauth.ClientAuthType)
```
mholt:
```suggestion
				return fmt.Errorf("client auth mode %s not recognized", clientauth.Mode)
```

evtr:
Hi Matt, I tried to incorporate your comments. Hope it is ok now

mholt:
@evtr Can you give me push privileges to your repo? I have a commit with some minor cleanups ready to push but I am getting a 403.

evtr:
Certainly. It should be ok now.

PR #2916, opened 2019-12-04
update quic-go to v0.14.0

marten-seemann:
v0.14.0 supports draft-24.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/keybase/go-ps v0.0.0-20190827175125-91aafc93ba19
 	github.com/klauspost/compress v1.8.6
 	github.com/klauspost/cpuid v1.2.1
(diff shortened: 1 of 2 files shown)


marten-seemann:
draft-25 will be intended for deployment, so the QUIC working group can collect some real-world data. It will probably ship early next year.

Do you have any instrumentation in Caddy that could yield some useful results in that regard?

mholt:
Great, that's good to hear.

> Do you have any instrumentation in Caddy that could yield some useful results in that regard?

We used to via telemetry, but then everyone freaked out about it so we don't have it anymore, sadly.

marten-seemann:
So we’ll have QUIC support based on the hunch that it will be faster, but won’t have any data to verify that it actually is?
That makes me sad.

mholt:
Correct 😞

PR #2917, opened 2019-12-05
Trim those characters because they serve PHP as plain text if appended

MisterDuval:
…ed after .php on Windows @mholt  check your email for more info

### DIFF
--- a/caddyhttp/fastcgi/fastcgi.go
+++ b/caddyhttp/fastcgi/fastcgi.go
@@ -85,6 +85,8 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) 
 		// but we also want to be flexible for the script we proxy to.
 
 		fpath := r.URL.Path
+		// We trim those characters because they are served as plain text if appended after .php on Windows 
+		fpath = strings.TrimRight(fpath, " .")
 
 		if idx, ok := httpserver.IndexFile(h.FileSys, fpath, rule.IndexFiles); ok {
 			fpath = idx


mholt:
Thanks for the change! I saw your emails, but will carry on the conversation here I guess.

I want to understand better what is going on here. I have a few questions, but no PHP setup or Windows to investigate further:

1. Does this happen only on Windows?
2. What is actually causing this? Is the file system just ignoring trailing dots and spaces? That seems odd.
3. Why do these characters not have the same effect in other parts of the path?
4. What is the full character set that causes this? What about invisible and/or non-ASCII characters?

Along with answers to the questions, having some tests would be required. Would you be able to help out with that?

MisterDuval:
> Does this happen only on Windows?
not sure on Apple, but that's not the case on Linux

> What is actually causing this? Is the file system just ignoring trailing dots and spaces? That seems odd.
I think this is related to Windows filesystem. If you do `dir "index.php "` (space at the end), it lists the file. Same for `dir index.php.`

> Why do these characters not have the same effect in other parts of the path?
No sure, but it is not affected when putting dot or space before the filename

> What is the full character set that causes this? What about invisible and/or non-ASCII characters?
I tried them, and I think they are handled/removed by Go net/url

> Would you be able to help out with that?
Sure!

mholt:
@francislavoie tells me that it does not happen on Linux, so I assume it doesn't happen on macOS either (but it'd be good to test it out if someone has a PHP setup there).

Does it happen if the characters appear before the filename, e.g. `/%20index.php`?

What about `index.php%20%20` or `index.php..`?

If this doesn't happen with nginx on Windows, do we know why?

tobya:
Yes to 

Index.php..

No

.index.php 404

tobya:
It may be to do with how split is done since 

Index.php.txt returns  the correct index.php file - not the source, but really should return 404

tobya:
@mholt This is the caddyfile block I'm using.


````
test2917.example.ie {
     root "C:\BCSDEV\www\Z\apps\test2917"
  
   gzip
   errors C:\BCSDEV\www\Z\apps\test2917\testerror.log
   log  /  C:\BCSDEV\www\Z\apps\test2917\testaccess.log "{common}"

   # Connect over fastcgi to php 
   # Requires fastcgi server to be running.
   on startup C:\BCSDEV\php70\php-cgi.exe -b 54715 &
   fastcgi / 127.0.0.1:54715 php

   tls {
        ciphers ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-WITH-CHACHA20-POLY1305 ECDHE-RSA-WITH-CHACHA20-POLY1305 ECDHE-RSA-AES256-CBC-SHA ECDHE-RSA-AES128-CBC-SHA ECDHE-ECDSA-AES256-CBC-SHA ECDHE-ECDSA-AES128-CBC-SHA RSA-AES256-CBC-SHA RSA-AES128-CBC-SHA ECDHE-RSA-3DES-EDE-CBC-SHA RSA-3DES-EDE-CBC-SHA       
      }
}
````

MisterDuval:
@mholt , this security issue should be patched fast. 
2nd security issue: 
As index.php.txt is served as index.php. Imagine you upload a file : bluesky.php.jpg, it is allowed because file extension matches but it is executed as PHP...

mholt:
This feels like a hack, so before I'm comfortable merging this, I want to make sure we understand what is happening. Why does Windows ignore `%20` and `.` at the end of a filename? We absolutely need to know the answer to this question first.

tobya:
I agree - there is more going on here than just needing to trim extra chars. Something is going wrong with the split of file and ext , I can’t see it at a glance.

I don’t have a working build at the moment but ill try to set up.

MisterDuval:
This space and dot characters is a serious issue as it delivers the PHP source code. @tobya fix is OK (even if we could serve this as Apache and Nginx do). @mholt, please push this in production fast. 

The second issue could be fixed a bit later as it is a security issue but only on file upload, it is way more complex to solve (without breaking things) because the behavior seems to happen in the PHP FastCGI. I debugged fpath in fastcgi.go, and it outputs index.php.jpg when calling index.php.jpg, so PHP is serving this file. 
Caddy should prevent this if there is no "/" or "?" at the end of the allowed extension.

tobya:
@MisterDuval I can provide a binary if you wish, or you can build my source.

The second issue although we should fix it is not serious because

- index.php.jpg (if it doesnt exist) will server index.php (not ideal but not a security risk)
- index.php.jpg (if it does exist) is just servered as a file, no execution, so no security risk.

MisterDuval:
@tobya, I tested your change and it's OK for the PHP source issue ex: index.php.%20..%20

I do not agree with you on the second issue. I tested and index.php.jpg executes PHP (both Windows + Linux).
If I upload a file bluesky.php.jpg containing PHP, upload will be allowed by the website because .jpg is a allowed filetype, BUT Caddy will serve it as PHP. So my server is compromised.

tobya:
I understand what your saying, but that is not the behaviour of the code.

- if bluesky.php.jpg exists on disk it will be served without being executed.  

You can try it and see what happens.

MisterDuval:
@tobya you're right, only bluesky.php will allow bluesky.php.jpg to be executed. A bluesky.php.jpg file will serve a JPG.

MisterDuval:
> This feels like a hack, so before I'm comfortable merging this, I want to make sure we understand what is happening. Why does Windows ignore `%20` and `.` at the end of a filename? We absolutely need to know the answer to this question first.

@mholt , I'm trying to get some info, but I'm pretty confident that the dot and space is the character set to match: https://superuser.com/questions/1508994/windows-wierd-behaviour-for-dir-command

MisterDuval:
@mholt @tobya when do we fix that huge security issue? My fix acts like Nginx/Apache (serves as PHP), Tobya's makes a 404. Please take a decision and push :-)

MisterDuval:
@mholt we need a 1.0.5 for this issue, it's the same kind of security issue as 1.0.4 (case fix). Thank you :-)

mholt:
Sorry for the delay on this, I've been busy patching a few issues affecting production use of v2 for a customer.

Thank you for the patches, both of you. I'll get around to a release as soon as I can but right now I'm required to work on v2 to meet some important deadlines.

@MisterDuval I made the equivalent fix for v2 in https://github.com/caddyserver/caddy/commit/dae4913fe35ff3f8a97383061ea8d44c1e98279e -- if you want to take a look at that, let me know what you think.

MisterDuval:
@mholt No problem Matt, that was a bit long for this kind of security issue :-) 
The patch seams to be OK to me, I only checked visually. Haven't tried the v2 yet!

mholt:
Honestly, I think there's only one person using PHP on Windows in production, and he's already involved in this thread. :P

Anyway, let me know if you do get a chance to try it on v2.

tobya:
@mholt It would be great if we could get a release of 1.05 with this fix included as soon as you have time.  

Otherwise if that is not possible we should prevent download of Windows build from build server.

mholt:
I'm on holiday right now, and v1 releases take about half a day on average lately, so I'll get around to it as soon as possible after the holidays.

tobya:
Just pinging this @mholt  - we really really need a release of 1.05.  I understand that everyone is busy but 1.05 could concievably be last V1 release and it needs this fix in it

mholt:
@tobya Actually, I thought I tagged v1.0.5 several weeks ago. It should have been tagged with this commit: https://github.com/caddyserver/caddy/commit/11ae1aa6b88e45b077dd97cb816fe06cd91cca67

But I can't find the tag. It might have gotten lost in my OS reinstall or something, somehow? I dunno.

I remember I tried doing a full release of v1.0.5 that weekend, but MacOS Catalina broke my deployment script. For reasons I don't understand -- and couldn't figure out after hours of troubleshooting -- Catalina refuses to run the `go` commands the same way it did in Mojave, sigh. It's not worth my time to fix at this point.

Anyway, that commit should be v1.0.5. I'll try to give it another pass sometime, but right now it's more than just a matter of "everyone is busy" -- it's been a funding problem.

Feel free to build from that commit in the meantime.

PR #2921, opened 2019-12-10
Return 404 if exact filename does not exist on disk

tobya:
use

> os.Stat(filename)

before calling 

> fileSystem.Open 

On windows `fileSystem.Open` will still open a file that contains trailing `.` or ` ` .  `os.Stat` will raise an error.

Related to #2917


tobya:
ping @MisterDuval

francislavoie:
What happens re case insensitivity in this case? i.e. `Index.php`

tobya:
I'm getting some odd behaviour so I'll investigate

PR #2923, opened 2019-12-10
Fix for 2917 - Replace trailing . and [ ] with x to prevent match.

tobya:
This is a fix for #2917 

Additional trailing . and [  ] (space) can cause a php page to be returned without execution on *Windows*.

There was a suggestion in #2917 to trim these in fastcgi, but I feel a better place to do this is in fileserver.go as really these should never match.

The solution is a little bit of a kludge and I am happy if someone can find another solution ( I have one idea which I have outlined below)

Reasons for this issue
--

- A request for a file on windows will ignore any trailing . or ' ' 
- This mean a index.php. request will be rejected by fastcgi for not matching .php but was then being matched by fileserver via the FileSystem.Open  call.
- There are two ways to prevent this, 
- I have chosen the first which is to replace the trailing characters with a character that will not be ignored such as x 
- The second way is to get a listing of all files in the directory and compare against the filename being requested before opening.

Security issues
--

The only possible issue I can think of is that if someone can manage to create a file on the server index.phpxxx and request index.php... it will be shown, but I'm not sure how much use that would be as it will not be executed.

If we are concerned about this perhaps we could randomly pick a letter that replaces the . and [ ]

Testing
--
- updated 2 Tests that checked for .. at the end of url and now return 404
- I am now running this in production.

### DIFF
--- a/caddyhttp/staticfiles/fileserver.go
+++ b/caddyhttp/staticfiles/fileserver.go
@@ -79,13 +79,15 @@ func (fs FileServer) serveFile(w http.ResponseWriter, r *http.Request) (int, err
 
 	// Prevent absolute path access on Windows.
 	// TODO remove when stdlib http.Dir fixes this.
-	if runtime.GOOS == "windows" && len(reqPath) > 0 && filepath.IsAbs(reqPath[1:]) {
-		return http.StatusNotFound, nil
-	}
+	if runtime.GOOS == "windows" {
+		if len(reqPath) > 0 && filepath.IsAbs(reqPath[1:]) {
+			return http.StatusNotFound, nil
+		}
 
-	// Github 2917 Replace Trailing . and [ ] since windows wont match propertly.
-	// Otherwise index.php. matches index.php and returns.  Should be 404
-	reqPath = fs.replaceTrailingDotSpace(reqPath)
+		// Github 2917 Replace Trailing . and [ ] since windows wont match propertly.
+		// Otherwise index.php. matches index.php and returns.  Should be 404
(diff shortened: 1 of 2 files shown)


tobya:
Ping @MisterDuval

MisterDuval:
FYI, i tested: Apache and Nginx on Windows serve index.php..%20 as index.php (so they TrimRight)

tobya:
@MisterDuval Thanks for testing.  I still think 404 is the desired behaviour though.

tobya:
Since my last update where I made the replacement Windows only the updated tests are failing. 

Would appreciate any suggestions on way forward

Mohammed90:
The new behavior causes different platforms to respond with different codes. This is why the tests are failing. Tests expectations aside, I think it's problematic to have non-uniform behavior across operating systems. There's also the issue of allowing leakage of underlying operating system, but this being an issue is probably debatable.

tobya:
Just in reference to serving index.php or 404

I tried with a website I use called huntoffice.  I stuck a `.` after their cart to see what would happen, I got a 404

ScreenHunter 207

They seem to be using Nginx

ScreenHunter 208

MisterDuval:
@tobya, my test was on Nginx and Apache on Windows, not on Linux. They both serve PHP with dot or space at the end of the filename.

tobya:
#2917 has been merged so I'll close this.  Though I feel this is a better place to fix the issue.

mholt:
Thanks Toby. I agree that a 404 is more correct, but I think I'd rather not fight Windows' file system here.

PR #2936, opened 2019-12-17
v2: Proposal: Support tls_trusted_ca_certs in reverse_proxy.transport

zaquestion:
Allows specifying ca certs with by filename in
`reverse_proxy.transport`.

Example
```
:443

root * app

# Make HTML file extension optional
try_files {path}.html {path}

# Send requests to /api to backend
reverse_proxy /api api:443 {
    transport http {
        tls
        tls_trusted_ca_certs certs/rootCA.pem
    }
}

# Serve everything else from the file system
file_server

tls certs/cert.pem certs/key.pem
```

These changes should be backwards compatible by simply opting to add a new way of specifying certs in `modules/reverseproxy/httptransport.go`

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"reflect"
@@ -166,6 +167,9 @@ func (h *HTTPTransport) setScheme(req *http.Request) {
 
 // Cleanup implements caddy.CleanerUpper and closes any idle connections.
 func (h HTTPTransport) Cleanup() error {
+	if h.Transport == nil {
+		return nil
+	}
 	h.Transport.CloseIdleConnections()
 	return nil
 }
@@ -174,6 +178,8 @@ func (h HTTPTransport) Cleanup() error {
 // TLS configuration for the transport/client.
 type TLSConfig struct {
 	RootCAPool []string `json:"root_ca_pool,omitempty"`
+	// Added to the same pool as above, but brought in from files
+	RootCAPemFiles []string `json:"root_ca_pem_files,omitempty"`
 	// TODO: Should the client cert+key config use caddytls.CertificateLoader modules?
 	ClientCertificateFile    string         `json:"client_certificate_file,omitempty"`
 	ClientCertificateKeyFile string         `json:"client_certificate_key_file,omitempty"`
@@ -203,7 +209,7 @@ func (t TLSConfig) MakeTLSClientConfig() (*tls.Config, error) {
 	}
 
 	// trusted root CAs
-	if len(t.RootCAPool) > 0 {
+	if len(t.RootCAPool) > 0 || len(t.RootCAPemFiles) > 0 {
 		rootPool := x509.NewCertPool()
 		for _, encodedCACert := range t.RootCAPool {
 			caCert, err := decodeBase64DERCert(encodedCACert)
@@ -211,6 +217,14 @@ func (t TLSConfig) MakeTLSClientConfig() (*tls.Config, error) {
 				return nil, fmt.Errorf("parsing CA certificate: %v", err)
 			}
 			rootPool.AddCert(caCert)
+		}
+		for _, pemFile := range t.RootCAPemFiles {
+			pemData, err := ioutil.ReadFile(pemFile)
+			if err != nil {
+				return nil, fmt.Errorf("failed reading ca cert: %v", err)
+			}
+			rootPool.AppendCertsFromPEM(pemData)
+
 		}
 		cfg.RootCAs = rootPool
 	}
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -425,6 +425,7 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 //         tls_client_auth <cert_file> <key_file>
 //         tls_insecure_skip_verify
 //         tls_timeout <duration>
+//         tls_trusted_ca_certs <cert_files...>
 //         keepalive [off|<duration>]
 //         keepalive_idle_conns <max_count>
 //     }
@@ -500,6 +501,17 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 					h.TLS = new(TLSConfig)
 				}
 				h.TLS.HandshakeTimeout = caddy.Duration(dur)
+
+			case "tls_trusted_ca_certs":
+				args := d.RemainingArgs()
+				if len(args) == 0 {
+					return d.ArgErr()
+				}
+				if h.TLS == nil {
+					h.TLS = new(TLSConfig)
+				}
+
+				h.TLS.RootCAPemFiles = args
 
 			case "keepalive":
 				if !d.NextArg() {


modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -166,6 +165,9 @@ func (h *HTTPTransport) setScheme(req *http.Request) {
 
 // Cleanup implements caddy.CleanerUpper and closes any idle connections.
 func (h HTTPTransport) Cleanup() error {
+	if h.Transport == nil {
+		return nil
+	}
```
zaquestion: This prevents a panic in the case where `Provision` fails and `h.Transport` is left `nil` during the `Cleanup()` call 
https://github.com/caddyserver/caddy/blob/v2/context.go#L327

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -205,12 +207,8 @@ func (t TLSConfig) MakeTLSClientConfig() (*tls.Config, error) {
...
 	if len(t.RootCAPool) > 0 {
 		rootPool := x509.NewCertPool()
-		for _, encodedCACert := range t.RootCAPool {
-			caCert, err := decodeBase64DERCert(encodedCACert)
-			if err != nil {
-				return nil, fmt.Errorf("parsing CA certificate: %v", err)
-			}
-			rootPool.AddCert(caCert)
+		for _, caCert := range t.RootCAPool {
+			rootPool.AppendCertsFromPEM([]byte(caCert))
```
zaquestion: This is comparable to the approach over here: 
https://github.com/caddyserver/caddy/blob/5ab17a3a371c1f84ef5a649d9310971eb7959271/modules/caddytls/acmemanager.go#L130-L142

We could adapt `t.RootCAPool` to also just be the file paths as well, either way this change would be incompatible with any existing v2 configs that were putting the pem block directly in the config. Still I'd argue this is the right thing to do while v2 is still in beta.

modules/caddyhttp/reverseproxy/httptransport.go
ardan-bkennedy: Shouldn't this be an error since this should never happen? I get the desire to reduce the possibility of panic but if Provisioning fails, is the server working as intended?

modules/caddyhttp/reverseproxy/httptransport.go
ardan-bkennedy: Assuming the functionality of putting a pem block directly in the config is there because it was required by someone, how can the server maintain backwards compatibility? I hate adding flags/switches but it seems your needs are a different requirement or option someone may want to leverage?

modules/caddyhttp/reverseproxy/httptransport.go
zaquestion: I can't speak authoritatively, suffice to say that in the context of cleanup I wouldn't consider this an error but rather the absence of anything to clean up.

ardan-bkennedy: This is defensive programming to allow a potential bug to go un-noticed, IMHO. This is the kind of code that scares me. If `h.Transport` should never be `nil`, or if it is `nil` this call should not be made, then the application should log these scenarios. Returning an error is important.

modules/caddyhttp/reverseproxy/httptransport.go
zaquestion: Given the inconsistency with the acmemanager it self, it seems worthwhile to validate the specific requirement to load the pem block directly. Given that v2 is still in beta, I'd push to consider that this _is the time_ for backwards incompatible changes. Loading the pem block directly seems a bit usual to begin with IMO, the only context I've seen that before is when building keys directly into binaries to make it less obvious that it's a cert, which doesn't apply here.

That said, if we wanted to support both, I'd suggest we add an addition field here for files, and leave the pem block logic as it is

https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/reverseproxy/httptransport.go#L176


Then we can follow the same approach shown in `caddy/modules/caddytls/acmemanager.go`, but again given v2's status as a beta, this is the best time to standardize the approach.

ardan-bkennedy: I get this is in beta but I don't want to use that as a blank check to break others.

Right now Matt is the only developer on the project and he has some high priority items he needs to get done by mid January. If you can make a non-breaking change that is reasonable, it will be easier to make a decision to merge.

This being open source, you could fork the project for now, make your change and then early next year we can attempt to bring those changes back in.

zaquestion:
Tangentially, if we do take a breaking approach, I'd be open to doing the leg work to redo the config structure itself, if desired. Talked with @mholt I understand we're trying to optimize for flat structure in the `Caddyfile`, which make sense this is mostly a repost of the message I sent him.

Again in the interest of consistency, since the reverse_proxy.transport block has a lot of `tls_*` fields which seems out of line with the `tls` block (at the root or `http` level IIRC)
```
	"tls": {
		"root_ca_pool": ["..."],
		"client_certificate_file": "cert.pem",
		"client_certificate_key_file": "key.pem",
		"insecure_skip_verify": false,
		"handshake_timeout": "2s",
		"server_name": ""
	},
```

vs the tls  bool and tls_* fields
```
transport http {
	read_buffer  <size>
	write_buffer <size>
	dial_timeout <duration>
	tls  # use HTTPS instead of HTTP
	tls_client_auth <cert_file> <key_file>
	tls_insecure_skip_verify
	tls_timeout <duration>
	keepalive [off|<duration>]
	keepalive_idle_conns <max_count>
}
```
Something like, seems desirable, but I'm open to either.
``
tls {
    client_auth <cert> <key>
    insecure_skip_verify
    timeout
    root_ca_pool
}
```

File: modules/caddyhttp/reverseproxy/httptransport.go (hunk 1 above)
Comment (zaquestion) [reply modules/caddyhttp/reverseproxy/httptransport.go]: Seems fair, I couldn't say if this should never happen or not, but given your confidence and the fact that it's the `Transport` field on a struct called `HTTPTransport` that stance seems reasonable. 

Thoughts on:
```
    return errors.New("transport is nil, nothing to cleanup")
```

For context: from what I recall the code that calls this is largely ignorant to the modules themselves. If `Provision` fails, it simply runs the `Cleanup` logic as if it was shutting down. I'm not readily familiar enough to reference the relevant blocks offhand, but I'm not quite sure what the config loading logic should do differently, in the event that `Provision` fails (like due to a bad config, which is how I hit this), how would you know whether or not there's something to cleanup given the abstractions at play? Provision could return more information perhaps, or even a special error indicating cleanup needs, but that seems overly complex. Just fruit for thought ❤️ I know you like it

ardan-bkennedy:
@zaquestion I will leave the rest of this discussion to @mholt. I would like you to have what you need but at the same time Matt has some high priority items he needs to get done.

mholt:
Thanks for the PR, @zaquestion!

I do think I prefer to keep the Caddyfile as flat as possible (within reason) -- and I don't mind `tls_` prefix for a few properties -- most of which probably won't be specified most of the time anyway.

As for the PR itself, I like where it's headed -- gimme a chance to review it a little more and I'll come back with some feedback soon. Merry Christmas!

modules/caddyhttp/reverseproxy/httptransport.go
mholt: Ah, good catch!

Great discussion here; all good points. You're both right. :)

Normally, I avoid (overly) defensive programming as well, since errors should be exposed in the proper place. Indeed, this is not the right place to raise an error.

However, this function _does_ need to be defensive.

To clarify, `Cleanup()` **is supposed** to be defensive since it is called even if `Provision()` returns an error. From the [godoc](https://pkg.go.dev/github.com/caddyserver/caddy/v2@v2.0.0-beta11?tab=doc#CleanerUpper):

> Cleanup should work even if Provision returns an error, to allow cleaning up from partial provisionings.

The reason this is necessary is because `Provision()` may partially complete before returning an error; i.e. there's no way to guarantee that an error'ed `Provision()` call has rolled itself back. And duplicating that logic in a defer in `Provision()` and also in `Cleanup()` would be undesirable, tedious, and error-prone. The easiest thing to do I think is to simply have a defensive `Cleanup()` instead.

In short: this nil check is good, but we shouldn't return an error if `Cleanup` is nil. In that case, there is simply nothing to clean up and it may be a no-op. If there was an error earlier that prevented the Transport from being set, it should have already been reported as an error, hence (possibly) why this function is being called.

Hope that makes sense! Remove the error return here and this will be good.

ardan-bkennedy:
I still disagree because you want to log these situations since they shouldn’t be happening. There should be ways to prevent these situations to begin with IMO. I’ve learned over the years this can bite you hard.

It’s your call since it’s your product.

mholt:
It's okay if the situation happens. That's why Cleanup exists. Having nothing to clean up is not an error, I'm fact Cleanup is an optional method entirely. Any real error will definitely have been reported already by returning from Provision.

Anyway, it is truly fine, as long as the docs are met. I promise! Thanks for the feedback on the holiday!

Merry Christmas 🎄

PR #2945, opened 2019-12-22
FreeBSD init: PID file should not be executable

maxatome:
`install` use 0755 mode by default.


tobya:
Thank you for your submission, someone will try to review it as soon as they can.

PR #2947, opened 2019-12-24
v2: Added missing context for embedding

siredwin:
modified:   caddyhttp/httpserver/https.go
	modified:   caddytls/tls.go

### DIFF
--- a/caddy.iml
+++ b/caddy.iml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="GO_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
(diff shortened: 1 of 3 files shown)


Mohammed90:
Thanks for your contribution!
Just few issues:
1- The PR name says `v2:` but it's PR-ed against `master` which is still at v1
2- The context argument was added to the API 8 days ago (release v0.9.0 of certmagic) while `master` is still using v0.8.3
3- The type `context.Context` is an interface, so `certmagic` is receiving an interface of type `context.Context` but of value `nil`. Using that will cause `nil pointer dereference`
4- Remove the ` caddy.iml` file

siredwin:
In that case, I need to download the v2 branch and try this again

Issue #2959, opened 2020-01-06
v2: Caddyfile enhancements


mholt:
The Caddyfile is one of the last things to be worked on in v2 since it is just a thin layer over Caddy 2's config, in fact, it is a separate plugin (whereas in v1, it was the central mode of configuration). What we have at this point was just brought over from v1 with very few changes -- most of the changes are totally internal (like the fact that it just generates JSON now). The matchers are the only real external change. A good one, but clunky.

Anyway, it's time to get it up to speed, and get it documented, so people will actually want to use Caddy again.

There are a number of deficiencies in the Caddyfile from v1 and in v2 currently that have been studied over the last few months (and years, frankly), which we now have a chance to improve on. If I have time I will try to link to them here, but if you've done anything mildly advanced with the Caddyfile, you know what I'm talking about.

I don't want to sacrifice the Caddyfile's simplicity and usability, though, either.

Since we're still in beta, now is the time to make breaking changes to get it right. Also, since the current Caddyfile documentation is scant, I suspect very few people will be impacted by the changes.

These solutions/enhancements could go a lot of different ways, but here's what I've settled on as of now:

1) **Path matching will be an exact match instead of a prefix match. Prefix matching can still be done simply by ending in `*`.**

Example: Proxying all API endpoints, before:

```
reverse_proxy /api/ localhost:8080
```

And after:

```
reverse_proxy /api/* localhost:8080
```

2) **Within multiple instances of a directive, the Caddyfile adapter will sort them in descending order of specificity of their path matcher.** (Specificity is hard/impossible to define with other matchers.)

Before, this example would be impossible, since the first line would always match those in the second line, and would never evaluate the second line since proxying is terminal:

```
reverse_proxy /api/ localhost:8081
reverse_proxy /api/v1/ localhost:8080
```

After this change, the above will be valid, since it will be sorted in the other order: first matching handler wins.

3) **The `rewrite` directive will be mutually exclusive -- only the first matching rewrite will be executed. It will not "rehandle" -- this means rewrites do not cascade.**

Example: Our new v2 documentation site uses something like this:

```
rewrite /docs/json/* /docs/json/index.html
rewrite /docs/*      /docs/index.html
```

But this ends up redundant because the first rewrite makes it into the second rewrite, since rewrites "rehandle". With the change, only the first rewrite would be evaluated for requests to /docs/json/*.

[...]

PR #2963, opened 2020-01-06
transform a caddyfile with environment variables

sarge:
This will convert a caddy file with windows or unix replacements `{$VAR}` to a Caddy json syntax while preserving the variables.

A couple of observations.

Firstly I have kept the original parsing logic as it does provide some helpful error messages.

Secondly, this config will transform to 
```caddyfile
{$HOSTNAME}{
}
```

```json
{
        "apps": {
                "http": {
                        "servers": {
                                "srv0": {
                                        "listen": [
                                                ":2015"
                                        ],
                                        "routes": [
                                                {
                                                        "match": [
                                                                {
                                                                        "host": [
                                                                                "{env.HOSTNAME}{"
                                                                        ]
                                                                }
                                                        ],
                                                        "handle": [
                                                               ...
                                                        ]
                                                }
                                        ]
                                }
                        }
                }
        }
}
```
will take the default port number 2015 due to this line
https://github.com/caddyserver/caddy/blob/v2/caddyconfig/httpcaddyfile/addresses.go#L164

Since the variable will not match a valid hostname it cannot make an assumption about the port to use. But it does seem that this would be the least useful port. I was thinking that defaulting to `caddyhttp.DefaultHTTPSPort` instead.

### DIFF
--- a/caddyconfig/caddyfile/parse_test.go
+++ b/caddyconfig/caddyfile/parse_test.go
@@ -481,52 +481,52 @@ func TestEnvironmentReplacement(t *testing.T) {
 	// basic test; unix-style env vars
 	p := testParser(`{$ADDRESS}`)
 	blocks, _ := p.parseAll()
-	if actual, expected := blocks[0].Keys[0], "servername.com"; expected != actual {
+	if actual, expected := blocks[0].Keys[0], "{env.ADDRESS}"; expected != actual {
 		t.Errorf("Expected key to be '%s' but was '%s'", expected, actual)
 	}
 
 	// basic test; unix-style env vars
 	p = testParser(`di{$PARTIAL_DIR}`)
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Keys[0], "dir1"; expected != actual {
+	if actual, expected := blocks[0].Keys[0], "di{env.PARTIAL_DIR}"; expected != actual {
 		t.Errorf("Expected key to be '%s' but was '%s'", expected, actual)
 	}
 
 	// multiple vars per token
 	p = testParser(`{$ADDRESS}:{$PORT}`)
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Keys[0], "servername.com:8080"; expected != actual {
+	if actual, expected := blocks[0].Keys[0], "{env.ADDRESS}:{env.PORT}"; expected != actual {
 		t.Errorf("Expected key to be '%s' but was '%s'", expected, actual)
 	}
 
 	// windows-style var and unix style in same token
 	p = testParser(`{%ADDRESS%}:{$PORT}`)
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Keys[0], "servername.com:8080"; expected != actual {
+	if actual, expected := blocks[0].Keys[0], "{env.ADDRESS}:{env.PORT}"; expected != actual {
 		t.Errorf("Expected key to be '%s' but was '%s'", expected, actual)
 	}
 
 	// reverse order
 	p = testParser(`{$ADDRESS}:{%PORT%}`)
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Keys[0], "servername.com:8080"; expected != actual {
+	if actual, expected := blocks[0].Keys[0], "{env.ADDRESS}:{env.PORT}"; expected != actual {
 		t.Errorf("Expected key to be '%s' but was '%s'", expected, actual)
 	}
 
 	// env var in server block body as argument
 	p = testParser(":{%PORT%}\ndir1 {$FOOBAR}")
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Keys[0], ":8080"; expected != actual {
+	if actual, expected := blocks[0].Keys[0], ":{env.PORT}"; expected != actual {
 		t.Errorf("Expected key to be '%s' but was '%s'", expected, actual)
 	}
-	if actual, expected := blocks[0].Segments[0][1].Text, "foobar"; expected != actual {
+	if actual, expected := blocks[0].Segments[0][1].Text, "{env.FOOBAR}"; expected != actual {
 		t.Errorf("Expected argument to be '%s' but was '%s'", expected, actual)
 	}
 
 	// combined windows env vars in argument
 	p = testParser(":{%PORT%}\ndir1 {%ADDRESS%}/{%FOOBAR%}")
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Segments[0][1].Text, "servername.com/foobar"; expected != actual {
+	if actual, expected := blocks[0].Segments[0][1].Text, "{env.ADDRESS}/{env.FOOBAR}"; expected != actual {
 		t.Errorf("Expected argument to be '%s' but was '%s'", expected, actual)
 	}
 
@@ -540,21 +540,21 @@ func TestEnvironmentReplacement(t *testing.T) {
 	// malformed (non-existent) env var (unix)
 	p = testParser(`:{$PORT$}`)
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Keys[0], ":"; expected != actual {
+	if actual, expected := blocks[0].Keys[0], ":{env.PORT$}"; expected != actual {
 		t.Errorf("Expected key to be '%s' but was '%s'", expected, actual)
 	}
 
 	// in quoted field
 	p = testParser(":1234\ndir1 \"Test {$FOOBAR} test\"")
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Segments[0][1].Text, "Test foobar test"; expected != actual {
+	if actual, expected := blocks[0].Segments[0][1].Text, "Test {env.FOOBAR} test"; expected != actual {
 		t.Errorf("Expected argument to be '%s' but was '%s'", expected, actual)
 	}
 
 	// after end token
 	p = testParser(":1234\nanswer \"{{ .Name }} {$FOOBAR}\"")
 	blocks, _ = p.parseAll()
-	if actual, expected := blocks[0].Segments[0][1].Text, "{{ .Name }} foobar"; expected != actual {
+	if actual, expected := blocks[0].Segments[0][1].Text, "{{ .Name }} {env.FOOBAR}"; expected != actual {
 		t.Errorf("Expected argument to be '%s' but was '%s'", expected, actual)
 	}
 }
--- a/caddyconfig/caddyfile/parse_test.go
+++ b/caddyconfig/caddyfile/parse_test.go
@@ -15,6 +15,7 @@
 package caddyfile
 
 import (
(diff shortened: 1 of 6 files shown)


mholt:
Another excellent PR, @sarge ! I will take a look soon. Sorry I've been getting a bit behind on PRs these last few days... lots of things to do.

mholt:
So after thinking on this more, I think I want to solve this by deleting code rather than adding code.

It seems to me that the Caddyfile parser should have no notion of environment variable placeholders. Loosely speaking, the Caddyfile parser's only job is to unmarshal string tokens into a JSON structure.

So, a config like this:

```
{env.HOST} {
   ...
}
```

Would serialize to:

```
...
"host": ["{env.HOST}"]
...
```

Which is fine -- except, as you pointed out, the way it currently stands is that this won't qualify for automatic HTTPS because host matchers are not evaluated until request-time. We could potentially make a special case for env placeholders, and evaluate them up front, assuming that an HTTP handler doesn't change the environment (I think that's probably a safe assumption).

I think, then, all we need to do is rip out the stuff in the Caddyfile parser that deals with environment variable replacements.

sarge:
I think that is a fair place to land. 
We would keep the {$VAR} -: {env.VAR} transformation?
Default port to be `caddyhttp.DefaultHTTPSPort` rather than the admin port?
What happens to the automatic HTTPS support? Is that something that the `caddyfile` is responsible for?

mholt:
> We would keep the {$VAR} -: {env.VAR} transformation?

I don't see a need for it.

> Default port to be caddyhttp.DefaultHTTPSPort rather than the admin port?

No; the default port is 2015 (which is not the admin port) since most "default" configurations are those used in development, where you can't easily bind to low ports.

> What happens to the automatic HTTPS support? Is that something that the caddyfile is responsible for?

Nope, it's baked into the core. That was very important to the goals of the project: automatic HTTPS no matter where your configuration comes from, and on by default (where there's a domain name, of course). Automatic HTTPS gets enabled when the server is started: https://sourcegraph.com/github.com/caddyserver/caddy@78e98c40d35c0b3bc933886ce11cbf2d0cf44c99/-/blob/modules/caddyhttp/caddyhttp.go#L224-225

sarge:
Understood, that will simplify a bunch of things:)

mholt:
Do you want to revise this PR? Or, I am working on the Caddyfile this week and can make the changes. Either way.

sarge:
I'll take another run at it. What would be awesome is the `SubRoute` support for the reverseproxy:) I the forum a better place to discuss the status of a bunch of the v2 milestone issues? Chat?

mholt:
> I'll take another run at it. 

Sounds good, rip 'er out!

> What would be awesome is the SubRoute support for the reverseproxy:)

What do you mean by this?

> I the forum a better place to discuss the status of a bunch of the v2 milestone issues? Chat?

Actionable development items? Issue tracker probably.

sarge:
This issue, https://caddy.community/t/reverse-proxying-sub-path-to-golang-webserver-v2/6719
In v1 there was the `strip_prefix` which have now been implemented as commands?, but there is a bit of work to wire up a `strip_prefix` + `reverseproxy` which needed to implement a `route` concept to scope the `strip_prefix` to the same route as the `reverseproxy`.
I would take run at it but the plumbing is a bit beyond my caddy experience to date.

mholt:
Ah, right. I'm fixing that now, but not by embedding `strip_prefix` into the `reverse_proxy`. See https://github.com/caddyserver/caddy/issues/2959.

mholt:
Here's an interesting use-case for env variables in the Caddyfile parser that just popped up: https://caddy.community/t/caddy-v2-reusable-snippets/6744?u=matt

Specifically, the use of an environment variable in an `import` statement (which is basically a Caddyfile macro).

Not sure how I feel about this. I can see why it'd be nice to support this. Can we do it in a really easy way?

For example, maybe in the code for importing a snippet (or Caddyfile), we call `replacer := caddy.NewReplacer()` and then `replacer.ReplaceAll()` before we try looking for the snippet or filename?

smebberson:
I've just documented another use-case for env variables in the Caddyfile parser... https://caddy.community/t/caddy-v2-reusable-snippets/6744/6?u=smebberson

We use this feature religiously in Caddy v1 to ensure that our Caddy configuration is closer between environments.

We run different domains in development, beta and production and simply change the `APP_DOMAIN` environment variable for each environment.

This is a really important feature to us.

smebberson:
I think the updated approach to this PR would still work fine though as:

```
{env.APP_DOMAIN} {
}
```

If `APP_DOMAIN` was `https://example.com` should still translate to:

```
{
        "apps": {
                "http": {
                        "servers": {
                                "srv0": {
                                        "listen": [
                                                ":443"
                                        ],
                                        "routes": [
                                                {
                                                        "match": [
                                                                {
                                                                        "host": [
                                                                                "example.com"
                                                                        ]
                                                                }
                                                        ],
                                                        "handle": [
                                                                {
                                                                        "handler": "subroute"
                                                                }
                                                        ]
                                                }
                                        ]
                                }
                        }
                }
        }
}
```

mholt:
@smebberson I guess this begs the question: Do you want the placeholders evaluated at Caddyfile adaptation-time, or at request-time? Our current plan is to defer evaluation of placeholders until the modules are loaded and taking requests.

Environment variables are kind of a special case since they likely won't change while the process is running. But, when using a config adapter, it may be adapted at one time, then executed much later in a different environment.

So... we should probably figure this out.

- If Caddy is given JSON (i.e. no config adapter), then the env var placeholders must appear in the JSON, obviously, and the production environment must have those variables.

- If the a config adapter is used, the production environment might be different from the development environment where the `adapt` command was run (if that's the workflow). Alternatively, the `--adapter` flag might be used so that it is converted to JSON right before being used, in the same production environment. Both are possible.

- Do we treat env var placeholders as special? Maybe a Caddyfile-adapter-only feature that accepts environment variables and evaluates them like a pre-processor (i.e. before parsing, first substitute all env variable placeholders)? Should it be configurable?

Hope this makes sense. Let me know your thoughts. I guess the big question here is: **Do we substitute env vars into the Caddyfile _before_ even parsing and adapting the Caddyfile? Or should we persist all env var placeholders into the final config and let the modules do the replacements at runtime?**

Now that I've seen an actual use case or two that would benefit from Caddyfile-adapt-time substitution, I am kind of on the fence again. Maybe the Caddyfile should substitute all `$ENV_VARIABLE` instances before parsing. That should be easy, right? Then any `{env.PLACEHOLDER}` instances would be persisted into the final config. What do you think?

smebberson:
@mholt yep, all makes sense. Thanks for including me in the discussion.

In response to the big question... I think the answer for me is **We should persist all env var placeholders into the final config and let the modules do the replacements at runtime.**

I think this provides the most flexibility about when and where config adapters are run (i.e. at dev time, at Docker image time, automated at container runtime).

It could be neat to offer `{$ENV_VARIABLE}` substitution at Caddyfile-adapt-time and then persist any `{env.PLACEHOLDER}` into the final config. This gives the best of both worlds allowing adapt-time to be influenced by the environment using `{$ENV_VARIABLE}` if desired and runtime to be influenced by the environment using `{env.PLACEHOLDER}` if desired.

One could achieve `{$ENV_VARIABLE}` adapt-time substitution outside of Caddy however, using something like Mustache (of course you'd have to use a different syntax). So it's definitely not a must but if Caddy offers it, it's just another tool and process we don't have to use. It could be pushed to v2.1 though if you get enough requests for it?

But I think **We should persist all env var placeholders into the final config and let the modules do the replacements at runtime.** is important for v2.

sarge:
Ok so the parser will seek out and replace `$ENV_[\w\d]+`.

The address parsing needs to stay in because there needs to know how to split between the host:port.

Anything special about the import process?

mholt:
Actually, sorry, I meant something more along the lines of this regular expression: `\$\w+` (in other words, no need for ENV_ prefix)

Although... this means that `$` would need to be escapable (not a big deal), and also that we'd need a way to terminate the variable if it appeared in the middle of a word, like `foo$EXAMPLEbar` ... bash has its own weird ways of doing this, and I don't love any of them. There's `${ENV}` but that looks like a Caddy placeholder, just different -- and then how bash handles and escapes quotes is different too, and I don't want people to think that this is bash.

Hmm.

Maybe we should just stick with `{$ENV_VAR}` -- is that weird? Or should we do `${ENV}`? Or just `$ENV` and hope that nobody needs it in the middle of a word?

sarge:
I'll admit that I was confused initially with the different replacement formats and figuring out where they are valid. 
I would be comfortable with {$VAR} at parsing time and {env.VAR} at runtime. I think we can remove {%VAR%} support. Because caddy has the adapt feature I think we can document these two phases of the configuration clearly

mholt:
That sounds good to me. It's easy to document and still fits with our conventions pretty consistently. And I'm 100% for removing the %VAR% support, doesn't seem particularly useful; most users are on Linux anyway.

sarge:
@mholt there are a couple of caveats on this. Address `[::1]:{env.PORT}` is not supported by the `ParseWithVariables` function would require a more sophisticated parser. Work arounds would be to pass the host and port as variables.

mholt:
@sarge Cool, I had an epiphany, so I just went ahead and implemented it, building on your changes. I hope that's OK. :)

Check it out. Basically, we just pre-process the Caddyfile and do replacements before any parsing begins. It's a global, unstructured search-and-replace.

This allows environment variables to expand out to multiple tokens (if desired), and it should make the special handling of `{env.VAR}` placeholders irrelevant (at least, at this layer).

caddyconfig/caddyfile/parse_test.go
```diff
@@ -473,75 +474,81 @@ func TestParseAll(t *testing.T) {
...
+			input:  "{$FOOBAR}{$FOOBAR}",
+			expect: "foobarfoobar",
+		},
+		{
+			input:  "{$FOOBAR",
+			expect: "{$FOOBAR",
+		},
+		{
+			input:  "{$LONGER_NAME $FOOBAR}",
+			expect: "",
```
smebberson: @mholt Is this correct? I would have expected to not be replaced as `{$LONGER_NAME $FOOBAR}` is not in `{$ENV}` format?.

smebberson: For that input, I would have expected `{$LONGER_NAME $FOOBAR}` as the output? What am I missing?

mholt: I think this is a correct test case. This means to use an environment variable with a name of `LONGER_NAME $FOOBAR`, which of course there isn't one. I dunno, I just tried some wacky input to see what happens, but I think I would expect this to be empty string.

smebberson: @mholt, right, okay, that makes sense now :) That's a strange environment variable! But I have seen stranger things.

sarge:
@mholt sure no drama - I will close

mholt:
No please don't close it, I'm about to merge it! :)

mholt:
Tada! Thanks for your help, @sarge! Couldn't have done this without it.

PR #2970, opened 2020-01-09
fix replacing variables on imported files

sarge:
Fixes small issue when importing files.

The parse time replacements are not happening.

### DIFF
--- a/caddyconfig/caddyfile/parse.go
+++ b/caddyconfig/caddyfile/parse.go
@@ -16,7 +16,6 @@ package caddyfile
 
 import (
 	"bytes"
-	"io"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


caddyconfig/caddyfile/parse.go
```diff
@@ -368,7 +369,14 @@ func (p *parser) doSingleImport(importFile string) ([]Token, error) {
...
 
-	importedTokens, err := allTokens(importFile, file)
+	input, err := ioutil.ReadAll(file)
+	if err != nil {
+		return nil, p.Errf("Could not read imported file %s: %v", importFile, err)
+	}
+
+	input = replaceEnvVars(input)
+
+	importedTokens, err := allTokens(importFile, bytes.NewReader(input))
```
mholt: Maybe we should move replaceEnvVars into allTokens instead? We'd have to change the function signature from an io.Reader to []byte, but that's OK I think, it's unexported anyway.

sarge: Ok, I might still read into an array - reworking the parsing logic to be stream based I think is a bit of work, it is friday here:)

Issue #2977, opened 2020-01-14
v2: Expose TLS placeholders


moorereason:
## Proposal

Born out of https://github.com/caddyserver/caddy/issues/2505#issuecomment-574224265.

### TLS Placeholders

```text
{http.request.tls.version}
	The TLS version name (e.g. "tls1.3")
{http.request.tls.cipher_suite}
	The TLS cipher suite (e.g. "TLS_CHACHA20_POLY1305_SHA256").
{http.request.tls.resumed}
	The TLS connection resumed a previous connection (e.g. "true").
{http.request.tls.proto}
	The negotiated next protocol (e.g. "h2").
{http.request.tls.proto_mutual}
	The negotiated next protocol was advertised by server (e.g. "true").
{http.request.tls.server_name}
	The server name requested by client, if any.

{http.request.tls.client.cert}
	The client certificate in PEM format.
{http.request.tls.client.fingerprint}
	The SHA256 checksum of the client certificate.
{http.request.tls.client.issuer}
	The issuer DN string of the client certificate.
{http.request.tls.client.serial}
	The serial number of the client certificate.
{http.request.tls.client.subject}
	The subject DN string of the client certificate.
```

### Shorthand Placeholders

| Shorthand  | Replaces                        |
|------------|---------------------------------|
| {tls_cipher}      |  {http.request.tls.cipher_suite}              |
| {tls_version}      |  {http.request.tls.version}              |
| {tls_client_cert}      |  {http.request.tls.client.cert}              |
| {tls_client_fingerprint}      |  {http.request.tls.client.fingerprint}              |
| {tls_client_issuer}      |  {http.request.tls.client.issuer}              |
| {tls_client_serial}      |  {http.request.tls.client.serial}              |
| {tls_client_subject}      | {http.request.tls.client.subject}              |

### Comments

1. ~~It's not clear to me if I should be using dashes or underscores for these placeholders.~~
1. Should the "client" placeholders be renamed to "peer" to match `tls.ConnectionState`?
1. I've renamed several items to be clearer (`issuer` vs `i-dn`).

I'm willing to work on a PR once we agree on the details.

### Revisions 📓 

- Use underscores instead of hyphens.
- Changed "i.e." to "e.g."
- Add underscores instead of mashing words together
- Removed `http.request.tls.client.cert_escaped`
- Changed `http.request.tls.client.cert_raw` to `http.request.tls.client.cert`
- Removed `http.request.tls.client.valid_*`
[...]

PR #2982, opened 2020-01-17
v2:  Expose TLS placeholders

### LINKED ISSUE #2505: tls_cipher is not logged correctly with TLSv1.3

moorereason:
This PR will fix #2505 and #2977.

I updated the const names in `caddyhttp/replacer.go` which creates a little diff churn.  The naming was a little inconsistent.  I couldn't help myself.  :smiley:

Is `{http.request.tls.client.cert}` useful as `string(cert.Raw)`?  I ported the v1 behavior, but it seems like that byte slice should be encoded to printable text.  I'm not sure what it's used for.

Second commit may be off-topic, but it's trivial.  Let me know if I need to revert it.

I'm new to Caddy, so suggest any changes.


mholt:
Woah, this looks like a great PR. I can't wait to review it! Lemme get beta 13 out the door first and I'll head back to this. Thanks!

modules/caddytls/values.go
```diff
@@ -127,9 +128,35 @@ var SupportedProtocols = map[string]uint16{
 	"tls1.3": tls.VersionTLS13,
 }
 
+// UnsupportedProtocols is a map of unsupported protocols.
```
mholt: It might be worth noting in the godoc that this is just for logging purposes, not enforcement.

modules/caddyhttp/replacer.go
```diff
@@ -208,12 +217,79 @@ func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.Respo
 	repl.Map(httpVars)
 }
 
+func getReqTLSReplacement(req *http.Request, key *string) (string, bool) {
```
mholt: Why is `key` a pointer? I believe empty string is also invalid, so there is probably no need for a nil key, right?

modules/caddyhttp/replacer.go
```diff
@@ -208,12 +217,79 @@ func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.Respo
...
+	}
+}
+
+// getTLSPeerCert retrieves the first peer certificate from a TLS session.
+// Returns nil if no peer cert is in use.
+func getTLSPeerCert(cs *tls.ConnectionState) *x509.Certificate {
+	if len(cs.PeerCertificates) == 0 {
+		return nil
+	}
+	return cs.PeerCertificates[0]
```
mholt: I can't remember, if the peer presents a certificate chain (bundle), will this be the leaf certificate or the CA? Which one will we want? Does it depend? We should consider these questions.

modules/caddytls/cipher_suites.go
```diff
@@ -0,0 +1,107 @@
+// Derived from https://github.com/golang/go/commit/0ee22d9 which should be
+// included in Go 1.14.  Once Go 1.14 is released, we can remove this file.
```
mholt: Go 1.14 is in RC, and I think we won't get around to merging this until Go 1.14 is released, probably within the next couple weeks. So, just keep in mind, I think we will be removing this file before the merge.

modules/caddyhttp/replacer.go
moorereason: Good question.  I'll try to find an answer.  I'm pretty sure I mirrored the behavior of v1 for this.

modules/caddytls/cipher_suites.go
moorereason: :+1: I'm in no rush.  Happy to wait for Go 1.14 to land.

modules/caddyhttp/replacer.go
moorereason: I'm was trying to reduce memory allocations.  Let me know if you would rather we just pass a string.

mholt: Let's just pass a string for now; an optimization like this should be profiled first.

modules/caddytls/values.go
moorereason: Pushed a commit to update the godoc and not export `unsupportedProtocols`.  Let me know if you agree with the change.

modules/caddyhttp/replacer.go
moorereason: Fixed.

modules/caddyhttp/replacer.go
moorereason: Here's what I see in the stdlib:

The godocs for [`tls.ConnectionState`](https://golang.org/pkg/crypto/tls/#ConnectionState) `PeerCertificates` are not explicit about the leaf cert being first.  There's a hint in [`tls.Certificate`](https://golang.org/pkg/crypto/tls/#Certificate), which says "A Certificate is a chain of one or more certificates, leaf first," but `PeerCertificates` is not of type `tls.Certificate`.

Browsing through [`crypto/tls/handshake_server.go#L545`](https://github.com/golang/go/blob/88e564edb13f1596c12ad16d5fd3c7ac7deac855/src/crypto/tls/handshake_server.go#L545), it certainly appears that the first certificate is assumed to be the leaf.

mholt: Indeed, it also seems to be the case here: https://github.com/golang/go/blob/33d20760a23acffd3b200b922c5687ed2aeb74cc/src/crypto/tls/conn.go#L1419

It's not strictly documented anywhere, but let's run with that, I guess.

mholt:
@moorereason It's looking good! After Go 1.14 is released, let's remove that redundant file and then I'll approve this for merging if there are no conflicts.

mholt:
> Is {http.request.tls.client.cert} useful as string(cert.Raw)? I ported the v1 behavior, but it seems like that byte slice should be encoded to printable text.

I just saw this, somehow. So actually, I think you make a good point... let's remove that particular placeholder and see what the actual needs are in Caddy 2 first. If we get requests for it, we can add it (maybe base64-encoded?) but no need to have it for right now.

moorereason:
> > Is {http.request.tls.client.cert} useful as string(cert.Raw)? I ported the v1 behavior, but it seems like that byte slice should be encoded to printable text.
> 
> I just saw this, somehow. So actually, I think you make a good point... let's remove that particular placeholder and see what the actual needs are in Caddy 2 first. If we get requests for it, we can add it (maybe base64-encoded?) but no need to have it for right now.

Removed in latest commit.  Also rebased to fix a merge conflict.

I'll be back once Go 1.14 is released.  👀 👍

mholt:
@moorereason Go 1.14 was released today, shall we finish this up!? :+1:

moorereason:
Pushed a commit for Go 1.14 support.  🚀

Mohammed90:
If this PR needs Go 1.14, then the go.mod file needs to be updated too.

mholt:
Bumped to Go 1.14 in https://github.com/caddyserver/caddy/commit/78760c0ddcd890702977c876fdb126026e13f937

PR #2986, opened 2020-01-17
v2: some documentation & rename of circuit breaker config parameter

Mohammed90:
Adds docs for circuit breaker config and "random selection" load balancing policy.

The other config I know of that are missing docs is for NTLM transport here:
https://caddyserver.com/docs/json/apps/http/servers/errors/routes/handle/reverse_proxy/transport/http_ntlm/

Most of them are straightforward though.

### DIFF
--- a/modules/caddyhttp/reverseproxy/circuitbreaker.go
+++ b/modules/caddyhttp/reverseproxy/circuitbreaker.go
@@ -130,9 +130,14 @@ func (c *localCircuitBreaker) checkAndSet() {
 
 // Config represents the configuration of a circuit breaker.
 type Config struct {
+	// The threshold over sliding window that would trip the circuit breaker
 	Threshold float64 `json:"threshold"`
-	Type      string  `json:"type"`
-	TripTime  string  `json:"trip_time"`
+	// Possible values: latency, error_ratio, and status_ratio. It
+	// defaults to latency.
+	Type string `json:"type"`
+	// How long to wait after the circuit is tripped before allowing operations to resume.
+	// The default is 5s.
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/reverseproxy/circuitbreaker.go
```diff
@@ -130,9 +130,14 @@ func (c *localCircuitBreaker) checkAndSet() {
 
 // Config represents the configuration of a circuit breaker.
 type Config struct {
+	// The threshold over sliding window that would trip the circuit breaker
 	Threshold float64 `json:"threshold"`
-	Type      string  `json:"type"`
-	TripTime  string  `json:"trip_time"`
+	// Possible values: latency, error_ratio, and status_ratio. It
+	// defaults to latency.
+	Type string `json:"type"`
```
mholt: Thinking on it more, I actually really like your idea to rename `type` to `factor`.

Mohammed90: Done

Mohammed90:
I'm thinking we should also add more descriptive comments on these parameters:

https://github.com/caddyserver/caddy/blob/372540f0ee7f534d15ce000af176d0880efba5a3/modules/caddyhttp/caddyauth/hashes.go#L53-L67

Something along what Filippo did in his [blogpost](https://blog.filippo.io/the-scrypt-parameters/) but succinct for documentation lines.

mholt:
@Mohammed90 Careful though, I don't want to get into the business of documenting primitives of cryptographic functions. That's not really our MO. Instead, let's just link to some authoritative documentation that already exists.

PR #2990, opened 2020-01-19
php_fastcgi: No longer try index.php by default

### LINKED ISSUE #2989: v2+windows+php-cgi: Display index when file does not exist
My Windows is in this situation:
https://gist.github.com/fu-sen/490442e1a1b832c5623ef7c8168a3925
https://caddy.community/t/caddy-v2-php-cgi-on-windows/6825

At this time, When referring to a file that does not exist, the index is displayed without issuing 404.
Returns 404 normally when `php_fastcgi` is removed.

This bug has been confirmed with Caddy v2 + Windows + php-cgi.

mholt:
The user can still explicitly configure this manually with a try_files if they want to fall back to a PHP file if there's no match.

fix #2989

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
@@ -93,7 +93,7 @@ func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 //     }
 //     redir @canonicalPath {path}/ 308
 //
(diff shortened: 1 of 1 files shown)


mholt:
@francislavoie Not every PHP app does its own routing through an index.php file. As exhibited in #2989, without this change it is difficult to make the server return a 404 if an index.php file exists that is not designed as a 404 page.

So yes, in your case you would need `try_files {path} index.php`, since your index.php file is acting as a router. But in general, that is not always the case.

PR #2997, opened 2020-01-22
v2: Implement RegExp Var Matcher

Mohammed90:
This PR adds var matcher against regular expressions rather than just static values. It should help a lot with the nginx-adapter, and can serve as wildcard matcher too.

### DIFF
--- a/modules/caddyhttp/matchers_test.go
+++ b/modules/caddyhttp/matchers_test.go
@@ -547,77 +547,87 @@ func TestHeaderREMatcher(t *testing.T) {
 
 func TestVarREMatcher(t *testing.T) {
 	for i, tc := range []struct {
+		desc       string
 		match      MatchVarRE
 		input      VarsMiddleware
 		expect     bool
 		expectRepl map[string]string
 	}{
 		{
+			desc:   "match static value within var set by the VarsMiddleware succeeds",
 			match:  MatchVarRE{"Var1": &MatchRegexp{Pattern: "foo"}},
-			input:  VarsMiddleware{"Var1": "foo"},
+			input:  VarsMiddleware{"Var1": "here is foo val"},
 			expect: true,
 		},
 		{
+			desc:   "value set by VarsMiddleware not satisfying regexp matcher fails to match",
 			match:  MatchVarRE{"Var1": &MatchRegexp{Pattern: "$foo^"}},
 			input:  VarsMiddleware{"Var1": "foobar"},
 			expect: false,
 		},
 		{
+			desc:       "successfully matched value is captured and its placeholder is added to replacer",
 			match:      MatchVarRE{"Var1": &MatchRegexp{Pattern: "^foo(.*)$", Name: "name"}},
 			input:      VarsMiddleware{"Var1": "foobar"},
 			expect:     true,
 			expectRepl: map[string]string{"name.1": "bar"},
 		},
 		{
+			desc:   "matching against a value of standard variables succeeds",
 			match:  MatchVarRE{"{http.request.method}": &MatchRegexp{Pattern: "^G.[tT]$"}},
 			input:  VarsMiddleware{},
 			expect: true,
 		},
 		{
+			desc:   "matching agaist value of var set by the VarsMiddleware and referenced by its placeholder succeeds",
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/vars.go
```diff
@@ -88,6 +89,74 @@ func (m VarsMatcher) Match(r *http.Request) bool {
...
 }
 
+// MatchVarRE matches the value of the context variables by a given regular expression.
+//
+// Upon a match, it adds placeholders to the request: `{http.regexp.name.capture_group}`
+// where `name` is the regular expression's name, and `capture_group` is either
+// the named or positional capture group from the expression itself. If no name
+// is given, then the placeholder omits the name: `{http.regexp.capture_group}`
+// (potentially leading to collisions).
+type MatchVarRE map[string]*MatchRegexp
```
francislavoie: Is there a reason you called it `MatchVarRE`? I think `MatchVarRegexp` would fit the project's naming conventions a bit better.

Mohammed90: On mobile now so I'm not able to grab the links... The `-RE` suffix is analogous to the other regexp based matchers, i.e. `MatchHeaderRE` and `MatchPathRE`.

francislavoie: Ah, okay, I missed that. I was looking more at like `http.regexp` and such. 😄

modules/caddyhttp/vars.go
```diff
@@ -25,6 +25,7 @@ import (
 func init() {
 	caddy.RegisterModule(VarsMiddleware{})
 	caddy.RegisterModule(VarsMatcher{})
+	caddy.RegisterModule(MatchVarRE{})
```
zaquestion: nit: It seems like it would be more consistent to name this `RegexVarsMatcher` and a little clearly to spell out `Regex` IMO

zaquestion: Overlooked the collapsed conversation below, looks like you've already thought specifically about the naming. 👍

modules/caddyhttp/matchers_test.go
```diff
@@ -545,6 +545,82 @@ func TestHeaderREMatcher(t *testing.T) {
...
+			match:  MatchVarRE{"{http.request.method}": &MatchRegexp{Pattern: "^G.[tT]$"}},
+			input:  VarsMiddleware{},
+			expect: true,
+		},
+		{
+			match:  MatchVarRE{"{http.vars.Var1}": &MatchRegexp{Pattern: "[vV]ar[0-9]"}},
+			input:  VarsMiddleware{"Var1": "var1Value"},
+			expect: true,
+		},
+	} {
```
zaquestion: Though now followed in the other testcases in this file, this test is well positioned to leverage sub tests in go. 

Ref: https://blog.golang.org/subtests

I would suggest adding a `desc` field to your test case structure above, which would help articulate what you're trying to test in each case in addition to meeting the needs for sub tests. It also looks like these could be run in parallel no probs, but I could be missing something

```
tc := tc // capture range value
t.Run(tc.desc, func(t *testing.T) {
    t.Parallel()
    ... contents of for loop ...
}
```

modules/caddyhttp/vars.go
```diff
@@ -88,6 +89,74 @@ func (m VarsMatcher) Match(r *http.Request) bool {
...
+// where `name` is the regular expression's name, and `capture_group` is either
+// the named or positional capture group from the expression itself. If no name
+// is given, then the placeholder omits the name: `{http.regexp.capture_group}`
+// (potentially leading to collisions).
+type MatchVarRE map[string]*MatchRegexp
+
+// CaddyModule returns the Caddy module information.
+func (MatchVarRE) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "http.matchers.var_regexp",
```
mholt: For consistency with the `vars` handler

```suggestion
		ID:  "http.matchers.vars_regexp",
```

modules/caddyhttp/vars.go
mholt:
```suggestion
type MatchVarsRE map[string]*MatchRegexp
```

For consistency with `VarsMiddleware`

Mohammed90:
Done, done, and done.

PR #3002, opened 2020-01-25
Update orig_uri placeholder docs

### LINKED ISSUE #3001: V2: incorrect documentation for original uri placeholders
Hi,

Just to let you know that [this documentation page](https://caddyserver.com/docs/json/apps/http/) indicates that the request's original URI placeholder is `{http.request.orig.uri}`, but it happens to be wrong. It is actually `{http.request.orig_uri}`.

Thanks for the amazing work :pray: .

francislavoie:
Fixes #3001

See https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/replacer.go#L112-L128, this updated the documentation to match the replacers as implemented.

### DIFF
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -59,12 +59,12 @@ func init() {
 // `{http.request.host}` | The host part of the request's Host header
 // `{http.request.hostport}` | The host and port from the request's Host header
 // `{http.request.method}` | The request method
-// `{http.request.orig.method}` | The request's original method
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the PR Francis!

This discrepancy is probably because I waffled between the two forms for some time, unable to decide which is better or more proper. I still wonder if we should spell out `original` instead of just using `orig` for brevity. What do you think?

francislavoie:
I think `orig` is fine, personally. We can just document what `orig` is short for if you're concerned.

Issue #3004, opened 2020-01-26
v2: if custom TLS certificate is set, Let's Encrypt sites break


theel0ja:
```caddyfile
example.com {
        root * /var/www/example.com/public
        file_server

       tls /etc/ssl/certs/example.com.pem /etc/ssl/private/example.com.key
}

cdn.example.com {
       root * /var/www/cdn.example.com/public
       file_server

       tls /etc/ssl/certs/example.com.pem /etc/ssl/private/example.com.key
}

example.org {
	respond "example.org"
}
```

In this case, `example.com` and `cdn.example.com` work, but `example.org` returns error message in browsers (in Firefox, `SSL_ERROR_INTERNAL_ERROR_ALERT`) and into Caddy's logs.

If I remove the `example.com` blocks, `example.org` works.

```
caddy[1228]: 2020/01/26 02:22:18 http: TLS handshake error from CLIENTIP:CLIENTPORT: no server TLS configuration available for ClientHello: &{CipherSuites:[19018 4865 4866 4867 49195 49199 49196 49200 52393 52392 49171 49172 156 157 47 53 10] ServerName:example.org SupportedCurves:[56026 29 23 24] SupportedPoints:[0] SignatureSchemes:[1027 2052 1025 1283 2053 1281 2054 1537 513] SupportedProtos:[h2 http/1.1] SupportedVersions:[43690 772 771 770 769] Conn:0xc0000f2028}
```

```
$ caddy version
v2.0.0-beta.13 h1:QL0JAepFvLVtOatABqniuDRQ4HmtvWuuSWZW24qVVtk=
```

Henrocker:
Yes, this is exactly the same that happend to me: https://caddy.community/t/v2-automatic-https-certificate-errors/6847/21

mholt:
I believe I've fixed this in https://github.com/caddyserver/caddy/commit/4a07a5d41e0f54d1a1ec998b9d956ccf2a880d90 -- please pull the latest from the v2 branch and try again, if you would!

Henrocker:
I've just rebuild Caddy v2 with latest master (including https://github.com/caddyserver/caddy/commit/57c6f22684e74191814a30f9de83a05c11ac4b78) and I get panics again, when 1. using LE sites only in Caddyfile, and 2. in combination with sites with specified certificate paths in Caddyfile. 

Go version 1.13.8 and System is: Debian 10 x86_64.

[...]

PR #3005, opened 2020-01-26
v2: only compare custom TLS protocol relative to each other if both are set

Mohammed90:
The `max` argument for the `protocols` setting of the `tls` directive is optional. If not specified, the created `*tls.Config` has its MaxVersion set to TLS 1.3. That is, given such Caddyfile:

```
example.com {
    root * /path/to/root
    tls cert.pem cert.key {
        protocols tls1.2
    }
}
```
Caddy should have TLS config with minimum of tls1.2 and max tls1.3 (which is the default max). Because max is not specified, the `ConnectionPolicy` struct has an empty string as value of `ProtocolMax`. So the comparison `p.ProtocolMin > p.ProtocolMax` results in `true` because string with content is greater than an empty string.

We should check if both knobs have custom values in them before comparing them, and we should check first before actually changing values in `*tls.Config`.

Found via: https://caddy.community/t/caddy-and-staticfiles-django/6889/3

### DIFF
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -222,14 +222,14 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 	}
 
 	// min and max protocol versions
+	if (p.ProtocolMin != "" && p.ProtocolMax != "") && p.ProtocolMin > p.ProtocolMax {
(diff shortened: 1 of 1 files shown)


mholt:
Ah, thanks!

mholt:
(Oops, commit 4a07a5d linking to this PR was supposed to link to issue #3004 instead.)

Issue #3019, opened 2020-02-03
v2: Expose a caddyfile directive ordering API to modules


roblabla:
Currently, when building a Caddy module that provides a caddyfile directive, module authors have to modify caddy's [directiveOrder](https://github.com/caddyserver/caddy/blob/v2/caddyconfig/httpcaddyfile/directives.go#L29). This sort makes building out-of-tree caddy modules a bit of a pain.

Maybe we should have some broad directive categories instead (say, Authentication, Routing, ServeContent, ErrorHandling...) that would have a fixed order, and then directives can register in one of those categories? Or maybe allow directives to specify they want to register "before" another directive?

mholt:
Use the `order` global option: https://caddyserver.com/docs/caddyfile/options

You can change or add directives in any place in the list using it, without changing any source code.

I looked into the buckets/bins idea about a year ago when I was designing it, but they are not sufficient and are too unwieldy in practice.

roblabla:
Ah, that does work! Maybe the `order` global option should be mentioned in the [extending caddy](https://caddyserver.com/docs/extending-caddy) guide though?

mholt:
Definitely. It's a newer thing. I need to expand the Extending Caddy guide too, so I'll swing around and do that at some point.

PR #548, opened 2016-01-29
Making directives externally registerable

captncraig:
This will allow other packages using caddy as a library to register directives without modifying caddy's source. 

I can also make a custom fork of caddy and register plugins in an init function in a separate file to make a custom build of caddy.

Issue #409, opened 2015-12-09
The order of directives


lilee:
Now the caddy directives's order is hard code. 

I think the directives should be ordered as the directives write in the Caddyfile

captncraig:
Do you have an example of a case where you would like it changed? Can you share a caddyfile snippet? 

Initially this sounds like an ok idea, but it is complicated by the fact that there are built-in directives that don't go in the caddyfile. 

My worry is that often I put the "most important (to me)" directives first in my Caddyfile, not necessarily considering the ordering consequences. Having them hard-coded allows the caddy and plugin authors define a relatively "safe" ordering.

mholt:
Thanks for the suggestion @lilee.

But I disagree.

Consider this Caddyfile:

```
example.com

markdown
log access.log
header -Server
```

Here the site owner intends to log all requests and, as a security precaution, remove the "Server" header.  But markdown handles requests for Markdown files, and it doesn't pass requests down the middleware chain because it already wrote to the response (with the rendered Markdown body). If other middleware like the core file server at the end of the chain tried to write the response, it'd be a major malfunction.

As such, requests for markdown files wouldn't be logged and the header would never be stripped. Similar bad things happen with error and gzip middlewares. Each directive is carefully placed in the chain where it belongs.

lilee:
As I mentioned in the issue #408, I want ```my_middleware``` to be handled before ```proxy```. Because ```my_middleware``` will provide some context for ```proxy```

mholt:
That's fine; when you create a new middleware you can put its directive higher in the list than `proxy` and it will be handled before it.

lilee:
@mholt but I think hardcode is not flexible

mholt:
If there's anything else to discuss on this, feel free to re-open and clarify, but I think we answered your question.

PR #3024, opened 2020-02-07
v2: Enabling order in RegisterHandlerDirective for better external plugin development

milifili:
Currently a http module directive can be registered through RegisterHandlerDirective, but to make the module works there is a need to change directiveOrder slice in directive.go file at caddyconfig package. This approach doesn't work if we use caddy as library and adding extra http module. So with this PR, we can register a module with order!

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -107,6 +107,28 @@ func RegisterHandlerDirective(dir string, setupFunc UnmarshalHandlerFunc) {
 	})
 }
 
+// RegisterHandlerDirectiveWithOrder is like RegisterHandlerDirective, but
+// adding dir param to directive order list after the specified insertAfterDirective
+func RegisterHandlerDirectiveWithOrder(dir string, setupFunc UnmarshalHandlerFunc, insertAfterDirective string) {
+	RegisterHandlerDirective(dir, setupFunc)
+	if insertAfterDirective == "" {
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the PR @milifili and welcome to the project!

While I appreciate it, I don't think we can accept this change. There's a bit of history here; see #409, #548, #1473, and #3019 for starters - I think there's even more discussion in the forums.

Directives are registered in an arbitrary order because `init()`s are executed in a mostly non-deterministic order. Thus, ordering directives relatively during registration is sure to result in a non-deterministic ordering, which can lead to inconsistent and incorrect results.

While the order of HTTP handlers is important, the "correct" order varies from case to case. Our hard-coded default ordering is a good default that works for most people, but when it doesn't, they can use the [`route` directive](https://caddyserver.com/docs/caddyfile/directives/route) to force a certain order. Directives appearing in a `route` block do not need to be in the hard-coded list of directives at all! Alternatively, users can use the [`order` option in their Caddyfile](https://caddyserver.com/docs/caddyfile/options) to change the default order and add new directives to the ordering.

I've re-approached this problem several times over the past few years, but I do not think there is a better way to do it like what you're proposing. The order needs to be absolutely defined or relatively defined in a deterministic and extensible way, but this approach satisfies neither of those requirements.

I'm open to ideas on how to improve on the situation, but now that Caddy 2 lets you change the directive ordering using the `order` option or the `route` directive, I think we've satisfactorily solved this problem for most people. There's definitely no need to change Caddy's source code.

PR #3026, opened 2020-02-08
browse: allow filter init via `filter` query param

### LINKED ISSUE #3025: browse: allow filter init via `filter` query param
It would be nice to be able to create links that initialize the file browsing filter so only certain files are displayed by default. For example, if a directory contains many .png and .jpg files, `https://example.com/photos/?filter=.png` would show just the ones with the `.png` substring.

jjlin:
This allows creating links that display only a subset of files in a directory.

Fixes #3025 for Caddy v1.

### DIFF
--- a/caddyhttp/browse/setup.go
+++ b/caddyhttp/browse/setup.go
@@ -326,7 +326,7 @@ footer {
 }
 </style>
 	</head>
-	<body onload='filter()'>
+	<body onload='initFilter()'>
 		<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" height="0" width="0" style="position: absolute;">
 			<defs>
 				<!-- Folder -->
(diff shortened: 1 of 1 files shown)


francislavoie:
@mholt this was for v1, not the same as the other PR

jjlin:
I thought it would be stating the obvious, so sorry if it wasn't clear. I updated the first comment to clarify the Caddy version being addressed.

caddyhttp/browse/setup.go
```diff
@@ -477,6 +477,16 @@ footer {
 			var filterEl = document.getElementById('filter');
 			filterEl.focus();
 
+			function initFilter() {
+				if (!filterEl.value) {
```
jjlin: This check ensures that any user changes to the initial filter value aren't clobbered when navigating forward/backward.

caddyhttp/browse/setup.go
```diff
@@ -477,6 +477,16 @@ footer {
...
+					var filterParam = new URL(window.location.href).searchParams.get('filter');
+					if (filterParam) {
+						filterEl.value = filterParam;
+					}
+				}
+				filter();
```
jjlin: `filter()` needs to be executed unconditionally on load to preserve the behavior added in #2368.

mholt:
Oh. Thanks for clarifying.

PR #3027, opened 2020-02-08
browse: allow filter init via `filter` query param

### LINKED ISSUE #3025: browse: allow filter init via `filter` query param

jjlin:
This allows creating links that display only a subset of files in a directory.

Fixes #3025 for Caddy v2.

### DIFF
--- a/modules/caddyhttp/fileserver/browsetpl.go
+++ b/modules/caddyhttp/fileserver/browsetpl.go
@@ -228,7 +228,7 @@ footer {
 }
 </style>
 	</head>
-	<body onload='filter()'>
+	<body onload='initFilter()'>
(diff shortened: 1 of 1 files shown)


mholt:
Why did you open three PRs? (At first I thought I just got the same notification 3 times, but...)

francislavoie:
One is for v2 and the other is for v1 @mholt

mholt:
Hmm, I see. I'm still getting used to GitHub's new notification thingie. It's kinda throwing me off.

PR #3028, opened 2020-02-08
WIP: Add default to tls connection

sarge:
Here is another attempt at providing support for SNI-less connections.  
The intent is to provide a TLS connection if a client does not pass a SNI name when initiating the connection, relevant to older TLS clients typically.

```
...
                "srv0": {
                    "listen": [
                        ":9443"
                    ],
                    "routes": [
                        {
                            "handle": [
                                {
                                    "handler": "subroute",
                                    "routes": [
                                        {
                                            "handle": [
                                                {
                                                    "body": "hi secure",
                                                    "handler": "static_response",
                                                    "status_code": 200
                                                }
                                            ],
                                            "match": [
                                                {
                                                    "path": [
                                                        "/version"
                                                    ]
                                                }
                                            ]
                                        }
                                    ]
                                }
                            ],
                            "match": [
                                {
                                    "remote_ip": {
                                        "ranges":["0.0.0.0/0"]
                                    }
                                }
                            ],
                            "terminal": true
                        }
                    ],
                    "tls_connection_policies": [
                        {
                            "match": {
                                "sni": [
                                    "api.local.security.io"
                                ],
                                "default": "api.local.security.io"
                            }
                        }
                    ]
                }...
```
Support for the Caddyfile to come.

When there is no SNI, we need a name to lookup a certificate by as all certificates are stored using a certificate name. ` "default": "api.local.security.io"` essentially defaults the SNI name if none is found. Without some other way to find or name a certificate perhaps this is as good as it gets, allows you to run an SNI-less connection with this 
```
...
"match": {
      "default": "api.local.security.io"
}
...
```

There is no link between the tls connection policy and the host matching so have used a wildcard matcher `remote_ip` to accept all requests. To support an SNI-less connection this configuration would need to be placed at the end of the config as this wildcard will match any request, this would be true for any wildcard matcher.

I wouldn't say I am super sold on this solution, you could argue that I have abused the matcher concept to make this work. An alternative would be to extend the `ConnectionPolicy`?

testing this is done using
```
curl -i -k https://127.0.0.1:9443/version
```


mholt:
Thanks for the PR @sarge!

Like you suggested, I think we can do it without needing to implement a new/hacky/special-case matcher.

Instead of implementing a new matcher, what if we simply use the `default` property and set that [in the underlying CertMagic config](https://sourcegraph.com/github.com/mholt/certmagic@9893774184aa0786275ae6cba4a1f7ae40584c57/-/blob/config.go#L103-106) (see connpolicy.go around ~L165)?

Then, regardless of what SNI matchers are used, the `default` property could set an empty ServerName to its value before matching, and then all other logic proceeds as usual.

Would that work?

sarge:
Closing in favour of https://github.com/caddyserver/caddy/pull/3031

PR #3031, opened 2020-02-09
WIP: add support for sni-less tls connections

sarge:
Here is attempt++ at providing support for SNI-less connections.
The intent is to provide a TLS connection if a client does not pass a SNI name when initiating the connection, relevant to older TLS clients typically.

```
"tls_connection_policies": [
                        {
                            "match": {
                                "sni": [
                                    "api.local.security.io"
                                ],
                                "default_server_name": "api.local.security.io"
                            }
                        }
                    ]
```
ntipick: `default_server_name` is actually the name of the certificate, this was highlighted in my case as I have a wildcard certificate so it might require `*.local.security.io`.

Wildcard matching on a host. We want to give every more specific match a chance, so would place this at the end. Is this the best implementation of a wildcard host match? 

```
"match": [
                                {
                                    "remote_ip": {
                                        "ranges":["0.0.0.0/0"]
                                    }
                                }
                            ],
```

Testing
```
curl -i -k https://127.0.0.1:9443/version
```

### DIFF
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -57,6 +57,19 @@ func (cp ConnectionPolicies) TLSConfig(ctx caddy.Context) (*tls.Config, error) {
 		}
 	}
 
+	// check for a default server name, in the case that no SNI information is passed.
+	var defaultCP *ConnectionPolicy
+	for _, p := range cp {
+		if p.DefaultServerName != "" {
+			if defaultCP == nil {
+				caddy.Log().Info("setting default connection policy")
+			} else {
+				caddy.Log().Warn("multiple default connection policies, using last one")
+			}
+			defaultCP = p
+		}
+	}
+
 	// pre-build standard TLS configs so we don't have to at handshake-time
 	for i := range cp {
 		err := cp[i].buildStandardTLSConfig(ctx)
@@ -100,6 +113,11 @@ func (cp ConnectionPolicies) TLSConfig(ctx caddy.Context) (*tls.Config, error) {
 				return pol.stdTLSConfig, nil
 			}
 
+			// attempt to use the default TLS
+			if defaultCP != nil {
+				return defaultCP.stdTLSConfig, nil
+			}
+
 			return nil, fmt.Errorf("no server TLS configuration available for ClientHello: %+v", hello)
 		},
 	}, nil
@@ -137,6 +155,9 @@ type ConnectionPolicy struct {
 	// Enables and configures TLS client authentication.
 	ClientAuthentication *ClientAuthentication `json:"client_authentication,omitempty"`
 
+	// Default Server Name is used if no SNI information is available
+	DefaultServerName string `json:"default_server_name,omitempty"`
+
 	matchers     []ConnectionMatcher
 	certSelector certmagic.CertificateSelector
 
@@ -159,6 +180,7 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 		PreferServerCipherSuites: true,
 		GetCertificate: func(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
 			cfgTpl, err := tlsApp.getConfigForName(hello.ServerName)
+			cfgTpl.DefaultServerName = p.DefaultServerName
 			if err != nil {
 				return nil, fmt.Errorf("getting config for name %s: %v", hello.ServerName, err)
 			}


mholt:
Aww yis -- I'm actually headed to sleep for now but at a quick glance I'm liking this approach a lot better. Do you like it?

> ntipick: default_server_name is actually the name of the certificate, this was highlighted in my case as I have a wildcard certificate so it might require *.local.security.io.

Hmm, I was kind of hoping that the default server name would be irrelevant to whether it matched a cert exactly or only by wildcard match. I'd need to look (or anyone can look, heh), but maybe we need to adjust CertMagic. The idea is that if an empty SNI comes in, we can pretend it came in as `sub.example.com` for instance, and then CertMagic should find the closest matching cert, even if it's only for `*.example.com`. (That might be something to fix in CertMagic, if it's truly not using the wildcard cert if that's the closest match.)

sarge:
I'll take a look a certmagic, the indexing appeared to include the certname and any SAN's.

I think you are right they are independent and possibly too tied to my current use case, but I have a TLS client connecting to the IP address of the hostname (client resolves the DNS itself, and uses the IP address). TLS without SNI is the first challenge, no hostname in the request is the second:) Lets separate these use cases, thinking this through I might be able to depend on a static IP address to make this work with no further changes.

mholt:
> Should this work? Because it currently doesn't because it automatically generates an sni match on 127.0.0.1.

If the cert has 127.0.0.1 in its subjects, then I think CertMagic will choose it. Don't you want it to match on 127.0.0.1 if there is a match?

I guess I don't really understand what is happening. Can you be more specific?

sarge:
First I have a client that does not support SNI. It also connects via IP address (no host header). All fairly odd and non standard. 

So I think we have a solution for the SNI, by setting the DefaultServerName. There is some interesting behaviour based on the type of server host name being used. I think some of my confusion is due to a recent change.

Things I have tried.

Test 1.
```
local.security.io:9443 {
  tls /test.cer /test.key {
    default local.security.io
  }
  respond /version 200 {
     body "hi secure"
  }
}
```

`curl -i -k local.security.io:9443/version` - works

`curl -i -k https://127.0.0.1:9443/version` - is unable to find the certificate despite the `default local.security.io`. This appears to be due to a default connection policy introduced - https://github.com/caddyserver/caddy/issues/3004 - as this does not have the default server name configured on it.


Test 2.
```
:9443 {
  tls /test.cer /test.key {
    default local.security.io
  }
  ...
}
```

`curl -i -k local.security.io:9443/version` - works

`curl -i -k https://127.0.0.1:9443/version` - works


Test 3.
```
127.0.0.1:9443 {
  tls /test.cer /test.key {
    default local.security.io
  }
  ...
}
```

`curl -i -k local.security.io:9443/version` - TLS connection works but host name does not resolve (expected)

`curl -i -k https://127.0.0.1:9443/version` - does not work for the same reason as Test 1.

Conclusions

Not sure what the correct fix is here, there is a bunch of subtle interplay between these features. In my view I think all of these scenarios should work.

mholt:
@sarge Great, that is way helpful, thanks for explaining that.

What's your JSON output for those various configs? (`caddy adapt`) -- I can't generate them because there's no code yet for the `default` subdirective.

sarge:
@mholt sure here you go. These adapters tell a little bit of a story, we can see an empty connection policy has been injected in scenarios 1 and 3.

Test 1.
```
local.security.io:9443 {
  tls /test.cer /test.key {
    default local.security.io
  }
  respond /version 200 {
     body "hi secure"
  }
}

---

{
    "apps": {
        "http": {
            "http_port": 9080,
            "https_port": 9443,
            "servers": {
                "srv0": {
                    "listen": [
                        ":9443"
                    ],
                    "routes": [
                        {
                            "handle": [
                                {
                                    "handler": "subroute",
                                    "routes": [
                                        {
                                            "handle": [
                                                {
                                                    "body": "hi secure B",
                                                    "handler": "static_response",
                                                    "status_code": 200
                                                }
                                            ],
                                            "match": [
                                                {
                                                    "path": [
                                                        "/version"
                                                    ]
                                                }
                                            ]
                                        }
                                    ]
                                }
                            ],
                            "match": [
                                {
                                    "host": [
                                        "api.local.security.io"
                                    ]
                                }
                            ],
                            "terminal": true
                        }
                    ],
                    "tls_connection_policies": [
                        {
                            "certificate_selection": {
                                "policy": "custom",
                                "tag": "cert0"
                            },
                            "default_server_name": "local.security.io",
                            "match": {
                                "sni": [
                                    "api.local.security.io"
                                ]
                            }
                        },
                        {}
                    ]
                }
            }
        },
        "tls": {
            "certificates": {
                "load_files": [
                    {
                        "certificate": "/home/fedora/workspaces/third-party/caddy/certificates/ccp-gateway-app.ssl.cer",
                        "key": "/home/fedora/workspaces/third-party/caddy/certificates/ccp-gateway-app.ssl.key",
                        "tags": [
                            "cert0"
                        ]
                    }
                ]
            }
        }
    }
}


```

Test 2.
```
:9443 {
  tls /test.cer /test.key {
    default local.security.io
  }
  respond /version 200 {
     body "hi secure B"
  }
}

---

{
    "apps": {
        "http": {
            "http_port": 9080,
            "https_port": 9443,
            "servers": {
                "srv0": {
                    "listen": [
                        ":9443"
                    ],
                    "routes": [
                        {
                            "handle": [
                                {
                                    "body": "hi secure B",
                                    "handler": "static_response",
                                    "status_code": 200
                                }
                            ],
                            "match": [
                                {
                                    "path": [
                                        "/version"
                                    ]
                                }
                            ]
                        }
                    ],
                    "tls_connection_policies": [
                        {
                            "certificate_selection": {
                                "policy": "custom",
                                "tag": "cert0"
                            },
                            "default_server_name": "local.security.io"
                        }
                    ]
                }
            }
        },
        "tls": {
            "certificates": {
                "load_files": [
                    {
                        "certificate": "/home/fedora/workspaces/third-party/caddy/certificates/ccp-gateway-app.ssl.cer",
                        "key": "/home/fedora/workspaces/third-party/caddy/certificates/ccp-gateway-app.ssl.key",
                        "tags": [
                            "cert0"
                        ]
                    }
                ]
            }
        }
    }
}
```

Test 3.
```
127.0.0.1:9443 {
  tls /test.cer /test.key {
    default local.security.io
  }
  ...
}

---


{
    "apps": {
        "http": {
            "http_port": 9080,
            "https_port": 9443,
            "servers": {
                "srv0": {
                    "listen": [
                        ":9443"
                    ],
                    "routes": [
                        {
                            "handle": [
                                {
                                    "handler": "subroute",
                                    "routes": [
                                        {
                                            "handle": [
                                                {
                                                    "body": "hi secure B",
                                                    "handler": "static_response",
                                                    "status_code": 200
                                                }
                                            ],
                                            "match": [
                                                {
                                                    "path": [
                                                        "/version"
                                                    ]
                                                }
                                            ]
                                        }
                                    ]
                                }
                            ],
                            "match": [
                                {
                                    "host": [
                                        "127.0.0.1"
                                    ]
                                }
                            ],
                            "terminal": true
                        }
                    ],
                    "tls_connection_policies": [
                        {
                            "certificate_selection": {
                                "policy": "custom",
                                "tag": "cert0"
                            },
                            "default_server_name": "local.security.io",
                            "match": {
                                "sni": [
                                    "127.0.0.1"
                                ]
                            }
                        },
                        {}
                    ]
                }
            }
        },
        "tls": {
            "certificates": {
                "load_files": [
                    {
                        "certificate": "/home/fedora/workspaces/third-party/caddy/certificates/ccp-gateway-app.ssl.cer",
                        "key": "/home/fedora/workspaces/third-party/caddy/certificates/ccp-gateway-app.ssl.key",
                        "tags": [
                            "cert0"
                        ]
                    }
                ]
            }
        }
    }
}
```

mholt:
Thanks!

Indeed...

So... a couple solutions come to mind, but I'm in a different head-space right now, so let me know your feedback before jumping into implement it right away.

Perhaps putting default_server_name in connection policies is the wrong place. Maybe the server name should already be established by the time we start choosing a policy. Maybe we need a new sibling property to `tls_connection_policies` called `tls_default_sni` instead. And maybe we need to move the Caddyfile config for this property from the `tls` directive into the global options block, a new option named `default_sni`, that applies across all site blocks. (Makes sense, I think.)

What do you think? Would that be more correct and cleaner? Would that still be useful?

sarge:
I think that would meet my usecase. The only thing you would lose is having different default sni names for different ports which I haven't come across. 
The implementation becomes `certmagic.DefaultServerName` = `default_sni` each time `certmagic` is called?

mholt:
> The only thing you would lose is having different default sni names for different ports which I haven't come across.

Yeah, this is pretty niche, and would still be possible in the JSON I think.

> The implementation becomes `certmagic.DefaultServerName` = `default_sni` each time certmagic is called?

Better yet, set `certmagic.Default.DefaultServerName` once, and all new configs derived from that will carry the value.

sarge:
Closing in favour of https://github.com/caddyserver/caddy/pull/3039

PR #3039, opened 2020-02-12
add support for sni-less tls connections

sarge:
Adds support for SNI-less connections.

Nailed this first try /s

[edit] 
btw I did attempt to use the `certmagic.Defaults.DefaultServerName` but it never applied to all future instances.


sarge:
Closing with a mix of sadness and embarrassment.  In favour of https://github.com/caddyserver/caddy/pull/3047

mholt:
No need to be sad or embarrassed @sarge :smile: We'll figure it out!

PR #3047, opened 2020-02-16
add global default sni

sarge:
So mistakes were made.

This is what I propose for the global SNI change.


caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -805,6 +808,14 @@ func tryInt(val interface{}, warnings *[]caddyconfig.Warning) int {
 	return intVal
 }
 
+func tryString(val interface{}, warnings *[]caddyconfig.Warning) string {
+	stringVal, ok := val.(string)
+	if val != nil && !ok && warnings != nil {
+		*warnings = append(*warnings, caddyconfig.Warning{Message: "not an string type"})
```
mholt:
```suggestion
		*warnings = append(*warnings, caddyconfig.Warning{Message: "not a string type"})
```

modules/caddyhttp/caddyhttp.go
```diff
@@ -124,6 +129,8 @@ func (app *App) Provision(ctx caddy.Context) error {
 
 	repl := caddy.NewReplacer()
 
+	certmagic.Default.DefaultServerName = app.DefaultSNI
```
mholt: This might be racey, esp. with regard to config reloads, since this variable is global, but app.DefaultSNI is not. I am looking at a different way to set DefaultServerName on the individual certmagic configs that are made [around here](https://github.com/caddyserver/caddy/blob/dd103a67876bd19afd000bda4410fbf83ee5b57d/modules/caddytls/connpolicy.go#L165).

sarge: I did wonder about that, in some options I went down I did set this directly at this point.

sarge:
Having an integration test around this would help prove the result. In my manual testing it worked for me, but if it is racey then we can get a bit lucky on smalller configurations

modules/caddyhttp/caddyhttp.go
sarge: Setting the DefaultServerName inside getCertificate would be very explicit

mholt:
Gah, as soon as I figure out where the actual test output is in our CI system, I will fix/merge this... sigh...

:mega: Anyone know where the CI job output is??

mholt:
Sigh... totally compiles and passes all tests on my machine. I don't know what's going on.

PR #3050, opened 2020-02-17
Don't lie: default basicauth hash to bcrypt

roblabla:
Came across this small inconsistency between docs and implementation. basicauth docs says algorithm defaults to bcrypt, but implementation will raise an error if no hash is specified.


mholt:
Oops. Where did you see that?

Note that Caddyfile defaults might be different from JSON defaults, since JSON is supposed to be more explicit/verbose by design; the Caddyfile is more along the lines of usable defaults.

mholt:
@roblabla Would you like an invite to our dev Slack? Just let me know an email

roblabla:
Sure, send an invite to slack@roblab.la

I saw that in the [http_basic config JSON documentation](https://caddyserver.com/docs/json/apps/http/servers/routes/handle/authentication/providers/http_basic/)

PR #3063, opened 2020-02-20
Add handler for unhandled errors in errorChain

### LINKED ISSUE #3053: v2: Error chains override the default error handling behavior
If a config has an error handling route, it will [bypass the default codepath that sets the status code](https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/server.go#L224) (it will go in the if branch instead of the else).

This is problematic, because it's possible for the error handler to never actually handle an error. What if our error handling route handles only some paths, but not others? For instance in the config at the bottom of the post, we actually don't handle anything at all! What happens in this case is, we accidentally return a 200 OK response instead of the promised 500 we should receive (as set by the `error` Handler).

Ideally, I think we should automatically append a suffix middleware that sets the error code. I think we could do this by [replacing this line](https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/caddyhttp.go#L188) with something along the lines of

```go
srv.errorHandlerChain = srv.Errors.Routes.Compile(HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
    status_code := repl.Get("http.error.status_code")
    if status_code_int, err := strconv.Atoi(status_code); err != nil {
        w.WriteHeader(status_code_int)
    } else {
        w.WriteHeader(500)
    }
}))
```

---

The problematic json. You can run it with `caddy run -config test.json -adapter json5`

```json5
{
  // Boilerplate to make it easy to run config
  "admin": { "disabled": true },
  "apps": {
    "http": {
      "http_port": 4444,
      "servers": {
        "srv0": {
          "listen": [":4444"],
          "routes": [{
            "handle": [{"handler": "error"}]
          }],
          // Imagine we have routes in there that match based on status_code for instance, and don't handle all cases
          "errors": {
            "routes": [
              {} // Noop route. I'm... surprised this works at all X)
            ]
          }
        }
      }
    }
  }
}
```

roblabla:
Currently, when an error chain is defined, the default error handler is bypassed entirely - even if the error chain doesn't handle every error. This may result in pages returning a blank 200 OK page instead of the error status defined by the error.

For instance, it's possible for an error chain to match on the error status code and only handle a certain subtype of errors (like 403s). In this case, we'd want any other errors to still go through the default handler and return an empty page with the status code.

This PR changes the "suffix handler" passed to errorChain.Compile toset the status code of the response to the error status code.

Fixes #3053

### DIFF
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -419,11 +419,11 @@ var emptyHandler Handler = HandlerFunc(func(http.ResponseWriter, *http.Request) 
 // Error chain does not actually handle the error (for instance, it matches only
 // on some errors). See #3053
 var errorEmptyHandler Handler = HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
-	http_error := r.Context().Value(ErrorCtxKey)
-	if handler_err, ok := http_error.(HandlerError); ok {
-		w.WriteHeader(handler_err.StatusCode)
+	httpError := r.Context().Value(ErrorCtxKey)
+	if handlerError, ok := httpError.(HandlerError); ok {
+		w.WriteHeader(handlerError.StatusCode)
 	} else {
-		w.WriteHeader(500)
+		w.WriteHeader(http.StatusInternalServerError)
 	}
 	return nil
 })
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -185,7 +185,21 @@ func (app *App) Provision(ctx caddy.Context) error {
 			if err != nil {
 				return fmt.Errorf("server %s: setting up server error handling routes: %v", srvName, err)
 			}
-			srv.errorHandlerChain = srv.Errors.Routes.Compile(emptyHandler)
+
+			// Add an implicit suffix middleware that, if reached, sets the
+			// StatusCode to the error stored in the ErrorCtxKey. This is to
+			// prevent situations where the Error chain does not actually handle
+			// the error (for instance, it matches only on some errors). See
+			// #3053
+			srv.errorHandlerChain = srv.Errors.Routes.Compile(HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
+				http_error := r.Context().Value(ErrorCtxKey)
+				if handler_err, ok := http_error.(HandlerError); ok {
+					w.WriteHeader(handler_err.StatusCode)
+				} else {
+					w.WriteHeader(500)
+				}
+				return nil
+			}))
 		}
 	}
 
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -186,20 +186,7 @@ func (app *App) Provision(ctx caddy.Context) error {
 				return fmt.Errorf("server %s: setting up server error handling routes: %v", srvName, err)
 			}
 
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/caddyhttp.go
```diff
@@ -185,7 +185,21 @@ func (app *App) Provision(ctx caddy.Context) error {
 			if err != nil {
 				return fmt.Errorf("server %s: setting up server error handling routes: %v", srvName, err)
 			}
-			srv.errorHandlerChain = srv.Errors.Routes.Compile(emptyHandler)
```
mholt: `emptyHandler` is already at the end of the chain, so do you think we could just replace `emptyHandler` here with your handler that you're currently appending down on lines 194-202? Maybe just pull it out to a new unexported variable next to `emptyHandler` called `emptyErrorHandler`.

roblabla: Done

modules/caddyhttp/caddyhttp.go
```diff
@@ -413,6 +414,20 @@ type MiddlewareHandler interface {
...
+// An implicit suffix middleware that, if reached, sets the StatusCode to the
+// error stored in the ErrorCtxKey. This is to prevent situations where the
+// Error chain does not actually handle the error (for instance, it matches only
+// on some errors). See #3053
+var errorEmptyHandler Handler = HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
+	http_error := r.Context().Value(ErrorCtxKey)
+	if handler_err, ok := http_error.(HandlerError); ok {
+		w.WriteHeader(handler_err.StatusCode)
+	} else {
+		w.WriteHeader(500)
```
mholt:
```suggestion
		w.WriteHeader(http.StatusInternalServerError)
```

modules/caddyhttp/caddyhttp.go
```diff
@@ -413,6 +414,20 @@ type MiddlewareHandler interface {
 // emptyHandler is used as a no-op handler.
 var emptyHandler Handler = HandlerFunc(func(http.ResponseWriter, *http.Request) error { return nil })
 
+// An implicit suffix middleware that, if reached, sets the StatusCode to the
+// error stored in the ErrorCtxKey. This is to prevent situations where the
+// Error chain does not actually handle the error (for instance, it matches only
+// on some errors). See #3053
+var errorEmptyHandler Handler = HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
+	http_error := r.Context().Value(ErrorCtxKey)
```
mholt:
```suggestion
	httpError := r.Context().Value(ErrorCtxKey)
```

(Go doesn't conventionally use snake_case)

modules/caddyhttp/caddyhttp.go
```diff
@@ -413,6 +414,20 @@ type MiddlewareHandler interface {
...
+	if handler_err, ok := http_error.(HandlerError); ok {
```
mholt:
```suggestion
	if handlerErr, ok := http_error.(HandlerError); ok {
```

modules/caddyhttp/caddyhttp.go
```diff
@@ -413,6 +414,20 @@ type MiddlewareHandler interface {
...
+		w.WriteHeader(handler_err.StatusCode)
```
mholt:
```suggestion
		w.WriteHeader(handlerError.StatusCode)
```

PR #3074, opened 2020-02-24
Update README.md

huchenme:
removing `-R` because of this error in MacOS `chown: -R: No such file or directory`


mholt:
Thanks for the fix!

PR #3075, opened 2020-02-24
Add insecure_disable_sni_matching directive to tls

danlsgiga:
~~This basically disables `StrictHostMatching` for Caddy with single server configurations and TLS Client auth enabled.~~

~~Since its a single server configuration, the concerns about someone being able to bypass the TLS Client Auth by connecting to another server running without TLS Client auth and using the `Host` header is not applicable anymore.~~

This PR adds a new `insecure_disable_sni_matching` subdirective to the `tls` directive and it will disable the default behaviour of caddy denying a TLS client authenticated connection using a mismatched TLS SNI and `Host` request header

### DIFF
--- a/caddyhttp/httpserver/plugin.go
+++ b/caddyhttp/httpserver/plugin.go
@@ -251,9 +251,9 @@ func (h *httpContext) MakeServers() ([]caddy.Server, error) {
 	// 2) if QUIC is enabled, TLS ClientAuth is not, because
 	//    currently, QUIC does not support ClientAuth (TODO:
 	//    revisit this when our QUIC implementation supports it)
-	// 3) if TLS ClientAuth is used, StrictHostMatching is on
+	// 3) if TLS ClientAuth is used and there is more than 1 server instance, StrictHostMatching is on
 	var atLeastOneSiteLooksLikeProduction bool
-	for _, cfg := range h.siteConfigs {
+	for cfgIndex, cfg := range h.siteConfigs {
 		// see if all the addresses (both sites and
 		// listeners) are loopback to help us determine
 		// if this is a "production" instance or not
@@ -292,12 +292,15 @@ func (h *httpContext) MakeServers() ([]caddy.Server, error) {
 			if QUIC {
 				return nil, fmt.Errorf("cannot enable TLS client authentication with QUIC, because QUIC does not yet support it")
 			}
-			// this must be enabled so that a client cannot connect
+			// this must be enabled for configs with more than 1 server instance
+			// so that a client cannot connect 
 			// using SNI for another site on this listener that
 			// does NOT require ClientAuth, and then send HTTP
 			// requests with the Host header of this site which DOES
 			// require client auth, thus bypassing it...
-			cfg.StrictHostMatching = true
+			if cfgIndex > 0 {				
+				cfg.StrictHostMatching = true
+			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


danlsgiga:
That's actually a good point! 

I can make adjustments to add a config flag to this just like its done for v2 via `strict_sni_host`. Works for you?

mholt:
Sure -- what do you have in mind specifically? A Caddyfile subdirective of `tls`?

danlsgiga:
Exactly... my only concern is that the default is already set to enable strict SNI host match, so we need to add something like the code snippet below to actually disable it:

```
tls cert.crt cert.key {
  clients /etc/pki/myca.pem
  allow_sni_host_mismatch
}
```

mholt:
How about `insecure_disable_sni_matching`? In keeping with `insecure_skip_verify`.

danlsgiga:
The new commits add a new `insecure_disable_sni_matching` subdirective to the `tls` directive to allow mismatching TLS SNI and Host headers when using TLS client authentication.

This is required whenever someone is using a different address to target the upstreams other than the Host header.

For example:

```
tls cert.crt cert.key {
  clients /etc/pki/ca.crt
  insecure_disable_sni_matching
}
proxy / srv+https://myapp.service.consul
```

The config above will allow Caddy's DNS SRV discovery to use a different address (myapp.service.consul) without strictly requiring the `Host` header to be the same. This effectively allows us to to mTLS between Caddy instances and still serve whatever the website is requested via `Host` header.

Obviously, the user must be aware of the security trade-offs of this directive and that must be present in the docs.

caddyhttp/httpserver/plugin.go
```diff
@@ -292,12 +292,13 @@ func (h *httpContext) MakeServers() ([]caddy.Server, error) {
...
-			// this must be enabled so that a client cannot connect
+			// this must be false so that a client cannot connect
 			// using SNI for another site on this listener that
 			// does NOT require ClientAuth, and then send HTTP
 			// requests with the Host header of this site which DOES
-			// require client auth, thus bypassing it...
-			cfg.StrictHostMatching = true
+			// require client auth, thus bypassing it.
+			// if set to true via `insecure_disable_sni_matching` directive, the behaviour above is allowed
+			cfg.TLS.InsecureDisableSNIMatching = false
```
mholt: Does using `insecure_disable_sni_matching` in your Caddyfile work with this line present? 🤔  My brain-that-needs-to-go-to-sleep is telling me that this overwrites any `true` value that may have been set.

danlsgiga: hey, yeah... it works. [here](https://github.com/caddyserver/caddy/pull/3075/files#diff-e74f72185700e304b2047c4af2ad1a48R449) I'm defaulting to the same *must* match behaviour. So, if `insecure_disable_sni_matching` is not set, the default will continue to be a strict host matching

danlsgiga: wait... your-brain-that-needs-to-go-to-sleep beats my brain-that-just-woke-up.

danlsgiga:
I believe GH is having issues today and the Multiplatform Tests are out of sync... its finished in the Azure pipelines but GH Actions is still waiting for it

danlsgiga:
Yup, I was thinking about it and I can give a go to it too. I'm also willing to add support to DynamoDB to Certmagic. Lets see how it goes! ;)

mholt:
@danlsgiga Ohhh, that'd be awesome! Let me know if I can help. You can either open an issue to track its development or post on the forums.

PR #3090, opened 2020-02-27
cmd: Add `caddy fmt` command.

### LINKED ISSUE #3020: v2: Add `caddy fmt` command to restructure/tab Caddyfiles
As someone who's been frequenting the https://caddy.community forums recently, I see a lot of people who don't bother to tab their Caddyfiles, and it makes it a bit annoying to read when helping them out.

I think it would be pretty cool if we had a  `caddy fmt /path/to/Caddyfile` command that just rewrote the Caddyfile while cleaning up whitespace.

vrongmeal:
This takes the config file as input and formats it.

Code currently implements a `Formatter` interface
which can be implemented by various adapter to format
a file.

Currently only Caddyfile formatter is implemnented.

Fixes #3020 

I still have to write tests (`TODO`).

An example of formatting (I hope this covers everything):

**Before `caddy format`:**

```
# Caddyfile with { comments } and "quotes"
# and improper \indentation

localhost { # starts with a comment
{$SOME_ENV}
		encode zstd gzip
  templates
 file_server browse\this\server
}

:8080, :8081 { respond "I am \"80\""
}
```

**After `caddy format`:**

```
# Caddyfile with { comments } and "quotes"
# and improper \indentation

localhost {
	# starts with a comment
	{$SOME_ENV}
	encode zstd gzip
	templates
	file_server browse\this\server
}

:8080, :8081 {
	respond "I am \"80\""
}
```

### DIFF
--- a/caddyconfig/caddyfile/formatter.go
+++ b/caddyconfig/caddyfile/formatter.go
@@ -1,0 +1,137 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package caddyfile
+
+import (
+	"bytes"
+	"io"
+	"unicode"
+)
+
+// Format formats a Caddyfile to conventional standards.
+func Format(body []byte) []byte {
+	reader := bytes.NewReader(body)
+	result := new(bytes.Buffer)
+
+	var (
+		commented,
+		quoted,
+		escaped,
+		block,
+		environ,
+		lineBegin bool
+
+		firstIteration = true
+
+		prev,
+		curr,
+		next rune
+
+		err error
+	)
+
+	for {
+		prev = curr
+		curr = next
+
+		if curr < 0 {
+			break
+		}
+
+		next, _, err = reader.ReadRune()
+		if err != nil {
+			if err == io.EOF {
+				next = -1
+			} else {
+				panic(err)
+			}
+		}
+
+		if firstIteration {
+			firstIteration = false
+			lineBegin = true
+			continue
+		}
+
+		if quoted {
+			if escaped {
+				escaped = false
+			} else {
+				if curr == '\\' {
+					escaped = true
+				}
+				if curr == '"' {
+					quoted = false
+				}
+			}
+			if curr == '\n' {
+				quoted = false
+			}
+		} else if commented {
+			if curr == '\n' {
+				commented = false
+			}
+		} else {
+			if curr == '"' {
+				quoted = true
+			}
+			if curr == '#' {
+				commented = true
+			}
+			if curr == '}' {
+				if environ {
+					environ = false
+				} else if block {
+					block = false
+				}
+			}
+			if curr == '{' {
+				if unicode.IsSpace(next) {
+					block = true
+
+					if !unicode.IsSpace(prev) {
+						result.WriteRune(' ')
+					}
+				} else {
+					environ = true
+				}
+			}
+			if lineBegin {
+				if curr == ' ' || curr == '\t' {
+					continue
+				} else {
+					lineBegin = false
+					if block {
+						result.WriteRune('\t')
+					}
+				}
+			} else {
+				if prev == '{' &&
+					(curr == ' ' || curr == '\t') &&
+					(next != '\n' && next != '\r') {
+					curr = '\n'
+				}
+			}
+		}
+
+		if curr == '\n' {
+			lineBegin = true
+		}
+
+		result.WriteRune(curr)
+	}
+
+	return result.Bytes()
+}
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 	"github.com/mholt/certmagic"
 	"go.uber.org/zap"
 )
@@ -534,6 +535,35 @@ func cmdValidateConfig(fl Flags) (int, error) {
 	}
 
 	fmt.Println("Valid configuration")
+
+	return caddy.ExitCodeSuccess, nil
+}
+
+func cmdFormatConfig(fl Flags) (int, error) {
+	// Default path of file is Caddyfile
+	formatCmdConfigFile := fl.Arg(0)
+	if formatCmdConfigFile == "" {
+		formatCmdConfigFile = "Caddyfile"
+	}
+
+	formatCmdWriteFlag := fl.Bool("write")
+
+	input, err := ioutil.ReadFile(formatCmdConfigFile)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup,
+			fmt.Errorf("reading input file: %v", err)
+	}
+
+	output := caddyfile.Format(input)
+
+	if formatCmdWriteFlag {
+		err = ioutil.WriteFile(formatCmdConfigFile, output, 0644)
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, nil
+		}
+	} else {
+		fmt.Print(string(output))
+	}
 
 	return caddy.ExitCodeSuccess, nil
 }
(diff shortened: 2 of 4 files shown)


francislavoie:
Awesome! Could we call it `caddy fmt`? The idea was to match [`gofmt`](https://golang.org/cmd/gofmt/) 😅

francislavoie:
Awesome! Could we call it `caddy fmt`? The idea was to match [`gofmt`](https://golang.org/cmd/gofmt/) 😅

vrongmeal:
> Awesome! Could we call it caddy fmt? The idea was to match gofmt 😅

Yeah! I'll change that. Also right now it takes a flag `--config` as the name of the file (just like `caddy adapt` and `caddy validate`). I guess we could change that to an argument as well.

cmd/commandfuncs.go
```diff
@@ -538,6 +538,37 @@ func cmdValidateConfig(fl Flags) (int, error) {
...
+	formatter := caddyconfig.GetFormatter(formatCmdFormatterFlag)
+	if formatter == nil {
+		return caddy.ExitCodeFailedStartup,
+			fmt.Errorf("unrecognized config formatter: %s", formatCmdFormatterFlag)
+	}
+
+	output := formatter.Format(input)
+	// fmt.Print(string(output))
+
+	err = ioutil.WriteFile(formatCmdConfigFlag, output, 0644)
```
mholt: I think the command must not mutate the input file. Dangerous and unexpected! Print to stdout instead. User can redirect to a file if they want. Or we can add a flag for it. This would be my first suggestion for a revision.

cmd/commands.go
```diff
@@ -242,6 +242,19 @@ provisioning stages.`,
 		}(),
 	})
 
+	RegisterCommand(Command{
+		Name: "format",
```
mholt: Let's call it `fmt` as an homage to `go fmt` :)

cmd/commandfuncs.go
francislavoie: FWIW I'd actually expect it to mutate the input. That's how most `fmt` things work. I feel like having a `--dry-run` option or similar would make more sense, which would just output to stdout.

mholt: @francislavoie 

>  That's how most fmt things work.

Citation needed. 😁  (But they won't change my mind, I'm afraid.)

Here, I'll give you a counterexample: https://golang.org/cmd/gofmt/ -- does not overwrite the input.

Overwriting input files is risky, and I feel pretty strongly that we shouldn't do that.

If you want to do that, it's easy with my proposed change:

```
caddy fmt > Caddyfile
```

This also makes the command more convenient for scripting and other use.

Plus, I've found that it's quite common in practice that the caddy process does not have write privileges on its input config file.

The sweet spot for this feature is potential editor integration in the future.

francislavoie: > Here, I'll give you a counterexample: golang.org/cmd/gofmt -- does not overwrite the input.

Could've sworn it did. Haven't played with Go in a while so 🤷‍♂ I stand corrected

I don't feel super strongly about this point, so I concede 😄

tobya: I'm with Matt here

vrongmeal: > caddy fmt > Caddyfile

So I think I should also default `--config` to take `Caddyfile` so we can use it this way.

And is it better to accept file as an argument or flag?

cmd/commands.go
vrongmeal: I actually named it `format` since all the other command names were expressive and this seemed like an exception among them. But I do like `fmt` too.

mholt: Gotcha. I'm torn. I don't care either way :)

vrongmeal:
Updated this:

1. Changed the name from `format` to `fmt`.
1. Changed path of config from flag to argument.
1. Default path is `Caddyfile` (so is the corresponding formatter).
1. Added `--write` flag if changes are to be written to the file.

cmd/commandfuncs.go
roblabla: FWIW `caddy fmt > Caddyfile` won't work because it will cause Caddyfile to be replaced with an empty file before `caddy fmt` runs. You can try it easily, the following will empty up file:

```
echo test > file
cat file > file
```

So you'd have to do it in a two step process of `caddy fmt > Caddyfile.tmp && mv Caddyfile.tmp Caddyfile`. Not a big deal though.

mholt: Wow, I hate bash.

caddyconfig/caddyformatters.go
```diff
@@ -0,0 +1,41 @@
...
+	return nil
+}
+
+// GetFormatter returns the Formatter with the given name,
+// or nil if one with that name is not registered.
+func GetFormatter(name string) Formatter {
+	return configFormatters[name]
+}
+
+var configFormatters = make(map[string]Formatter)
```
mholt: As cool as it would be to have formatters be extensible, I really can't think of a need for formatters beyond just the Caddyfile. I mean, I wouldn't want/expect Caddy to be able to format any other general-purpose config language like NGINX configs or YAML or any other config adapters. That'd be kinda out of scope, right?

If you agree, I think we can simplify things a lot. This file wouldn't be needed. No need to register formatters. No `--formatter` flag. Just call the `Format()` function from the command, since there's only the one.

cmd/commands.go
```diff
@@ -242,6 +242,21 @@ provisioning stages.`,
 		}(),
 	})
 
+	RegisterCommand(Command{
+		Name:  "fmt",
+		Func:  cmdFormatConfig,
+		Usage: "[--formatter <name>] [--write] [<path>]",
+		Short: "Formats a configuration file",
+		Long: `
+Formats the configuration file and prints the result on standard output.`,
```
mholt: This doc text will need some expansion before the final merge (take cues from some of the other commands).

cmd/commandfuncs.go
```diff
@@ -538,6 +538,47 @@ func cmdValidateConfig(fl Flags) (int, error) {
...
+	formatter := caddyconfig.GetFormatter(formatCmdFormatterFlag)
+	if formatter == nil {
+		return caddy.ExitCodeFailedStartup,
+			fmt.Errorf("unrecognized config formatter: %s", formatCmdFormatterFlag)
+	}
+
+	output := formatter.Format(input)
+
+	if formatCmdWriteFlag {
+		err = ioutil.WriteFile(formatCmdConfigFile, output, 0644)
```
mholt: We must be careful not to change a file's permissions. Be sure the original permissions are preserved.

cmd/commandfuncs.go
vrongmeal: Yeah, and so I added the `--write` option :D

caddyconfig/caddyformatters.go
vrongmeal: Umm, we could remove it. The first thing I thought about this was we could extend it to JSON for starters easily but I get why we wouldn't like to do that.

cmd/commands.go
vrongmeal: Yeah! I'll add more to it

cmd/commandfuncs.go
vrongmeal: This actually does not change permissions. If the file does not exist (which in this case we're sure it does) it just creates a file with these permissions.

From docs:
> WriteFile writes data to a file named by filename. If the file does not exist, WriteFile creates it with permissions perm (before umask); otherwise WriteFile truncates it before writing.

vrongmeal:
Update:
1. Remove `Formatter` interface (support only Caddyfile).
1. Improve description of command.

cmd/commandfuncs.go
mholt: Ah, I stand corrected. Thanks!

caddyconfig/caddyfile/formatter.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"bytes"
+	"io"
+	"unicode"
+)
+
+// Formatter is a type that implements caddyconfig.Formatter.
+// This is used to format a Caddyfile.
+type Formatter struct {
+}
```
mholt: This seems outdated now. This type is not needed, probably!

caddyconfig/caddyfile/formatter.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Format formats the body of Caddyfile.
+func (f Formatter) Format(body []byte) []byte {
```
mholt: This method can just be a function!
```suggestion
// Format formats a Caddyfile to conventional standards.
func Format(body []byte) []byte {
```

cmd/commands.go
```diff
@@ -242,6 +242,24 @@ provisioning stages.`,
...
 	})
 
+	RegisterCommand(Command{
+		Name:  "fmt",
+		Func:  cmdFormatConfig,
+		Usage: "[--write] [<path>]",
+		Short: "Formats a Caddyfile",
+		Long: `
+Formats the Caddyfile by adding proper indentation and spaces to improve
+human readability. It prints the output on the STDOUT by default.
```
mholt:
```suggestion
human readability. It prints the result to stdout.
```

vrongmeal:
Updated this with above suggestions.

Added tests too :)

PR #3111, opened 2020-03-02
remove the certificate tag tracking from global state

sarge:
Remove the tag tracking from global state. This better supports reloading configuration via the admin reload function.

@mholt thoughts about the `logCounter` it will mutate the log files unnecessarily I think https://github.com/caddyserver/caddy/pull/3111/files#diff-3235e86b75f6440c822f6d7d8f4beb61R540

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -524,10 +524,12 @@ func parseLog(h Helper) ([]ConfigValue, error) {
 
 		var val namedCustomLog
 		if !reflect.DeepEqual(cl, new(caddy.CustomLog)) {
+			logCounter := h.state["logCounter"].(int)
 			cl.Include = []string{"http.log.access"}
 			val.name = fmt.Sprintf("log%d", logCounter)
 			val.log = cl
 			logCounter++
+			h.state["logCounter"] = logCounter
 		}
 		configValues = append(configValues, ConfigValue{
 			Class: "custom_log",
@@ -536,5 +538,3 @@ func parseLog(h Helper) ([]ConfigValue, error) {
 	}
 	return configValues, nil
 }
-
-var logCounter int
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -153,7 +153,16 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 			// loaded won't find it, and TLS handshakes will fail (see end)
 			// of issue #3004)
 
-			tlsCertTags := h.state["tlsCertTags"].(map[string]string)
+			// tlsCertTags maps certificate filenames to their tag.
+			// This is used to remember which tag is used for each
+			// certificate files, since we need to avoid loading
+			// the same certificate files more than once, overwriting
+			// previous tags
+			tlsCertTags, ok := h.State["tlsCertTags"].(map[string]string)
+			if !ok {
+				tlsCertTags = make(map[string]string)
+				h.State["tlsCertTags"] = tlsCertTags
+			}
 
 			tag, ok := tlsCertTags[certFilename]
 			if !ok {
@@ -524,12 +533,17 @@ func parseLog(h Helper) ([]ConfigValue, error) {
 
 		var val namedCustomLog
 		if !reflect.DeepEqual(cl, new(caddy.CustomLog)) {
-			logCounter := h.state["logCounter"].(int)
+
+			logCounter, ok := h.State["logCounter"].(int)
+			if !ok {
+				logCounter = 0
+			}
+
 			cl.Include = []string{"http.log.access"}
 			val.name = fmt.Sprintf("log%d", logCounter)
 			val.log = cl
 			logCounter++
-			h.state["logCounter"] = logCounter
+			h.State["logCounter"] = logCounter
 		}
 		configValues = append(configValues, ConfigValue{
 			Class: "custom_log",
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -152,6 +152,9 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 			// policy that is looking for any tag but the last one to be
 			// loaded won't find it, and TLS handshakes will fail (see end)
 			// of issue #3004)
+
+			tlsCertTags := h.state["tlsCertTags"].(map[string]string)
+
 			tag, ok := tlsCertTags[certFilename]
 			if !ok {
 				// haven't seen this cert file yet, let's give it a tag
@@ -533,12 +536,5 @@ func parseLog(h Helper) ([]ConfigValue, error) {
 	}
 	return configValues, nil
 }
-
-// tlsCertTags maps certificate filenames to their tag.
-// This is used to remember which tag is used for each
-// certificate files, since we need to avoid loading
-// the same certificate files more than once, overwriting
-// previous tags
-var tlsCertTags = make(map[string]string)
 
 var logCounter int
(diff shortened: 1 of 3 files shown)


caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -152,6 +152,9 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 			// policy that is looking for any tag but the last one to be
 			// loaded won't find it, and TLS handshakes will fail (see end)
 			// of issue #3004)
+
+			tlsCertTags := h.state["tlsCertTags"].(map[string]string)
```
mholt: If this value doesn't exist (do an `, ok := ...` check), we should initialize and store it

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -43,6 +43,13 @@ func (st ServerType) Setup(originalServerBlocks []caddyfile.ServerBlock,
 	var warnings []caddyconfig.Warning
 	gc := counter{new(int)}
 
+	// tlsCertTags maps certificate filenames to their tag.
+	// This is used to remember which tag is used for each
+	// certificate files, since we need to avoid loading
+	// the same certificate files more than once, overwriting
+	// previous tags
+	var tlsCertTags = make(map[string]string)
```
mholt: Instead of initializing this here, we should do so where we use it.

Preserve the comment though, when deleting this line - thanks!

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -133,14 +140,19 @@ func (st ServerType) Setup(originalServerBlocks []caddyfile.ServerBlock,
...
 				Dispenser:    caddyfile.NewDispenser(segment),
 				options:      options,
 				warnings:     &warnings,
 				matcherDefs:  matcherDefs,
 				parentBlock:  sb.block,
 				groupCounter: gc,
-			})
+				state:        make(map[string]interface{}),
+			}
+			h.state["tlsCertTags"] = tlsCertTags
```
mholt: Again, initialize this where it is used; can remove this line

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -133,14 +140,19 @@ func (st ServerType) Setup(originalServerBlocks []caddyfile.ServerBlock,
...
+
+			results, err := dirFunc(h)
+
```
mholt: I like to keep `err != nil` checks immediately after the line that can produce the error, for readability

caddyconfig/httpcaddyfile/httptype.go
sarge: This is here to track tags across all the server blocks, so I think this should stay outside all of the server block loops.

caddyconfig/httpcaddyfile/builtins.go
sarge: Sure ok, but I don't think there is another code path today that will mean this is uninitialised. Initialising at this point will mean we may load different certificates with the same tags. Would an error be better?

caddyconfig/httpcaddyfile/httptype.go
sarge: Moved the initialisation of `state`

caddyconfig/httpcaddyfile/httptype.go
sarge: Sure thing

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -119,6 +119,7 @@ type Helper struct {
 	matcherDefs  map[string]caddy.ModuleMap
 	parentBlock  caddyfile.ServerBlock
 	groupCounter counter
+	state        map[string]interface{}
```
mholt: Maybe this should be exported so (external/third-party) plugins can also use it.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -43,6 +43,16 @@ func (st ServerType) Setup(originalServerBlocks []caddyfile.ServerBlock,
...
 
+	state := make(map[string]interface{})
+	// tlsCertTags maps certificate filenames to their tag.
+	// This is used to remember which tag is used for each
+	// certificate files, since we need to avoid loading
+	// the same certificate files more than once, overwriting
+	// previous tags
+	state["tlsCertTags"] = make(map[string]string)
+	// this tracks the logCounter across all servers
+	state["logCounter"] = 0
```
mholt: So, the reason I want to avoid initializing up here in Setup() is because the Caddyfile adapter is extensible, and I don't want to put plugins' logic into the core of the adapter. Otherwise we'll have to keep changing the core to support all the plugins.

Instead, let's move all this initialization to where we first use them, like so:

```go
tlsCertTags, ok := h.state["tlsCertTags"].(map[string]string)
if !ok {
    tlsCertTags = make(map[string]string)
    h.state["tlsCertTags"] = tlsCertTags
}
// then proceed to use it
```

Does that make sense?

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -133,14 +143,17 @@ func (st ServerType) Setup(originalServerBlocks []caddyfile.ServerBlock,
...
-			results, err := dirFunc(Helper{
+			h := Helper{
 				Dispenser:    caddyfile.NewDispenser(segment),
 				options:      options,
 				warnings:     &warnings,
 				matcherDefs:  matcherDefs,
 				parentBlock:  sb.block,
 				groupCounter: gc,
-			})
+				state:        state,
```
mholt: We can just initialize state right here
```suggestion
				state:        make(map[string]interface{}),
```

caddyconfig/httpcaddyfile/httptype.go
sarge: I see but leave state where is it, outside of all of the server blocks.

caddyconfig/httpcaddyfile/httptype.go
sarge: Hmm i'm not so sure about that.

mholt: Oh wait. You are right

caddyconfig/httpcaddyfile/httptype.go
mholt: Yes, exactly -- ignore my wrong comment below. :smile:

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -43,6 +43,16 @@ func (st ServerType) Setup(originalServerBlocks []caddyfile.ServerBlock,
 	var warnings []caddyconfig.Warning
 	gc := counter{new(int)}
 
+	state := make(map[string]interface{})
```
mholt: Maybe even bump this up a line so it's together with warnings and gc, which are also state we share in a very similar way.

caddyconfig/httpcaddyfile/directives.go
sarge: Done

caddyconfig/httpcaddyfile/httptype.go
sarge: Done

caddyconfig/httpcaddyfile/builtins.go
sarge: Done

caddyconfig/httpcaddyfile/httptype.go
sarge: Done

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -119,6 +119,7 @@ type Helper struct {
 	matcherDefs  map[string]caddy.ModuleMap
 	parentBlock  caddyfile.ServerBlock
 	groupCounter counter
+	State        map[string]interface{}
```
mholt: Add a godoc comment to this, maybe put it right after caddyfile.Dispenser near the top, because I personally like for exported fields to be at the top, but I don't care too much as far as merging this. Just needs godoc.

mholt:
Woo hoo! I like it! Just needs a godoc comment. You tested that the latest changes still fix the problem? :) If so, then let's merge it after that.

sarge:
@mholt Yep tested.

PR #3121, opened 2020-03-06
v2: Rework Replacer loop to ignore escaped braces

### LINKED ISSUE #3116: Unable to set header containing JSON in Caddy v2 b15
I'm trying to set the `Report-To` header to a value that is JSON. I've tried various attempts at escaping braces and quotes but have been unable to set the header value to valid JSON. For more details on the attempts I have tried, see below.

I believe this is a sensible feature request based on increased use of this specification: [Reporting API, W3C Working Draft, 25 September 2018](https://www.w3.org/TR/reporting-1/).

This feels like a feature request rather than a bug as the documentation doesn't not explicitly indicate that JSON header values are supported. If we don't want to support JSON (or the use of `{`, `}`) in header values, we could detail this behaviour in the documentation to clarify this.

I haven't yet tracked down the lines in the source code responsible for parsing header definitions in the Caddyfile. I'll update this issue if I do.

This appears to have been fixed in: #2063 but I believe the behaviour described there is reproducible in the latest beta.

---

### 1. My Caddy version (`caddy version`):
`v2.0.0-beta.15 h1:Td1esMk7bebftnoBuT3gOqUGxew5HqdIKw3s36S8tNw=`

#### a. System environment:

```plain
Linux blog 4.15.0-74-generic #84-Ubuntu SMP Thu Dec 19 08:06:28 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

I'm using the DIgital Ocean image but have upgraded Caddy to the latest version `v2.0.0-beta.15`.

#### b. Command:

I'm using `systemctl`. The following command is how I reload after a configuration change.
```
systemctl reload caddy
```

`cat /etc/systemd/system/caddy.service`

```plain
[Unit]
Description=Caddy Web Server
After=network.target

[Service]
User=caddy
Group=caddy
ExecStart=/usr/bin/caddy run --config /etc/caddy/Caddyfile --adapter caddyfile --resume --environ
ExecReload=/usr/bin/caddy reload --config /etc/caddy/Caddyfile --adapter caddyfile
TimeoutStopSec=5s
LimitNOFILE=1048576
LimitNPROC=512
PrivateTmp=true
ProtectSystem=full
AmbientCapabilities=CAP_NET_BIND_SERVICE

[Install]
WantedBy=multi-user.target
```

```plain
blog.glvr.io {
  root * /var/www/html
  file_server
  encode gzip zstd

  header {
    -Server
    Referrer-Policy no-referrer
    X-Xss-Protection "1; mode=block; report='https://bill.report-uri.com/r/d/xss/enforce'"
    Report-To "\{'group':'default','max_age':3600,'endpoints':[\{'url':'https://bill.report-uri.com/a/d/g'\}],'include_subdomains':true\}"
    X-Content-Type-Options nosniff
    Content-Security-Policy "default-src 'self'"
  }
}
```

### 3. The problem I'm having:

I'm trying to set the `Report-To` header to a value that is JSON.

```plain
{'group':'default','max_age':3600,'endpoints':[{'url':'https://bill.report-uri.com/a/d/g'}],'include_subdomains':true}
```
I've tried various attempts at escaping braces and quotes but have been unable to set the header value to valid JSON. For more details on the attempts I have tried, see below.

### 4. Error messages and/or full log output:

```plain
~# curl -XGET -I https://blog.glvr.io

HTTP/2 200 
accept-ranges: bytes
content-security-policy: default-src 'self'
content-type: text/html; charset=utf-8
etag: "q6kdmr6l9"
last-modified: Mon, 02 Mar 2020 11:37:39 GMT
referrer-policy: no-referrer
report-to: \],'include_subdomains':true\}
x-content-type-options: nosniff
x-xss-protection: 1; mode=block; report='https://bill.report-uri.com/r/d/xss/enforce'
content-length: 8541
date: Wed, 04 Mar 2020 10:50:52 GMT
```

### 5. What I already tried:

I've tried a couple of things:

1. No escaping at all

   **config:** `Report-To "{'group':'default','max_age':3600,'endpoints':[{'url':'https://bill.report-uri.com/a/d/g'}],'include_subdomains':true}"`
   **result:** `report-to: ],'include_subdomains':true}`

2. Escape all the `{`

   **config:** `Report-To "\{'group':'default','max_age':3600,'endpoints':[{'url':'https://bill.report-uri.com/a/d/g'}],'include_subdomains':true\}"`
   **result:** `report-to: ],'include_subdomains':true}`

3. Escape the outer `{`

   **config:** `Report-To "\{'group':'default','max_age':3600,'endpoints':[{'url':'https://bill.report-uri.com/a/d/g'}],'include_subdomains':true\}"`
   **result:** `report-to: \],'include_subdomains':true\}`

### 6. Links to relevant resources:

* I've found this previous question that appears to describe the same issue: [Header containing JSON](/t/header-containing-json/3468).
* This question links through to this PR: [ httpserver.Replacer: Rework loop to ignore escaped placeholder braces](https://github.com/caddyserver/caddy/pull/2075/files) which has been merged.
* This fix does not appear to have made it into the `v2` branch: https://github.com/caddyserver/caddy/blob/v2/modules/caddyhttp/replacer.go

billglover:
This PR fixes #3116. I have added test cases to existing tests. I've made the assumption that escaping the closing brace is optional (i.e. I'm not checking that an escaped opening brace is matched with an escaped closing brace). See test cases for specific examples.


billglover:
No problem. I'll add a couple of benchmarks and share the results.

billglover:
Benchmark before this change:

```plain
BenchmarkReplacer-8   	18244922	        61.0 ns/op	      48 B/op	       1 allocs/op
```

Benchmark after this change:

```plain
BenchmarkReplacer-8   	 3793706	       310 ns/op	     176 B/op	       5 allocs/op
```

Good catch, you are correct. I'll have a think about how to do this more efficiently. It is almost certainly possible to do this during the main `for` loop as part of the `replace` function.

mholt:
Awesome, thanks. Yeah, I wonder if it's possible to just skip the escaping backslash when building the string...

billglover:
Benchmark after optimisation:

```plain
BenchmarkReplacer-8   	10560511	       107 ns/op	      48 B/op	       1 allocs/op
```

Same number of allocs and bytes per operation. Still a little slower though. I'd be very interested in feedback on clarity.

mholt:
> Still a little slower though.

I hypothesize that it's only slower on the case of JSON input (i.e. input where escapes actually happen). You can vary your input string on the benchmark test and see if that's true.

Still, allocations are the big deal -- which is fixed now -- and the speed decrease is minimal. If that speed decrease is in fact just for escaped inputs, then that's probably totally acceptable.

I have a comment/question I'll inline.

Thanks for working on this!

replacer_test.go
```diff
@@ -59,6 +59,26 @@ func TestReplacer(t *testing.T) {
...
+			expect: `{"json": "object"}`,
+		},
+		{
+			input:  `\{"json": \{"nested": "object"\}\}`,
+			expect: `{"json": {"nested": "object"}}`,
+		},
+		{
+			input:  `\{"json": \{"nested": "{bar}"\}\}`,
+			expect: `{"json": {"nested": ""}}`,
+		},
```
mholt: Can we add more test cases? Specifically:

- backslashes that aren't acting as escapes
- malformed escapes (if there is such a thing)
- escaping only the closing brace
- repeated backslashes
- content before and after escaped brace segments

Or any others you can think of, but I think this would be pretty comprehensive.

replacer.go
```diff
@@ -125,6 +125,16 @@ func (r *Replacer) replace(input, empty string,
...
+
+		// check for the escape character
+		if i > 0 && input[i-1] == phEscape {
+			if input[i] == phClose || input[i] == phOpen {
+				sb.WriteString(input[lastWriteCursor : i-1])
+				lastWriteCursor = i
+			}
+			continue
+		}
+
```
mholt: For some reason, the placement of this block isn't sitting well with me. My bug sense is tingling. Maybe new test cases will help validate or dissolve my concerns.

I'm not sure that `continue` is in the right place. Does that skip all backslashes?

Doesn't this immediately write all content (minus backslash) since the `lastWriteCursor`, and then skip all the logic below? I feel like we need to maybe put backslash-skipping logic around existing `WriteString` calls. (Does that make sense?)

I'm just supposing here -- sorry if it turns out to be wrong!

billglover: This makes a lot of sense and I do believe there is a bug here. Will confirm this with some tests and the fix before considering code placement.

replacer_test.go
billglover: I've added additional test cases, along with one that prompted issue #3116 and these all pass.

replacer.go
billglover: Interestingly this works as expected. It doesn't skip all backslashes as `lastWriteCursor` isn't modified and so they are picked up in the next write. That said, I've collapsed the `if` block as the way it is structured is open to confusion.

Moving things closer to the existing `WriteString` calls works well for the check around opening braces but I haven't yet found a way to avoid making the check for the closing brace more complex. Will continue to think about this one.

replacer_test.go
mholt: Awesome, they LGTM!

replacer.go
```diff
@@ -112,7 +112,7 @@ func (r *Replacer) ReplaceFunc(input string, f ReplacementFunc) (string, error)
 func (r *Replacer) replace(input, empty string,
 	treatUnknownAsEmpty, errOnEmpty, errOnUnknown bool,
 	f ReplacementFunc) (string, error) {
-	if !strings.Contains(input, string(phOpen)) {
+	if !strings.Contains(input, string(phOpen)) && !strings.Contains(input, string(phClose)) {
```
mholt: Interesting, was this needed for a test case to pass? I figured if there is no opening brace, there is definitely no placeholder -- we don't even have to check for a closing one.

replacer.go
mholt: Okay. Does that mean you want to continue working on this PR? Or is it finished do you think?

mholt:
Oh, and just before we merge, could you post the final benchmarks? Thanks again!

replacer.go
billglover: This was needed for cases like this:

```plain
{
	input:  `\}`,
	expect: `}`,
},
```

Without it, you don't get the chance to remove the escape character from a closing braces. It's a valid question as to whether this is a case we want to handle.

billglover:
Benchmarks for three different scenarios:

* no_placeholder - `simple string`
* placeholder - `{"json": "object"}`
* escaped_placeholder - `\{"json": \{"nested": "{bar}"\}\}`

```plain
BenchmarkReplacer/no_placeholder-8         	100000000	        11.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkReplacer/placeholder-8            	21644844	        53.0 ns/op	      32 B/op	       1 allocs/op
BenchmarkReplacer/escaped_placeholder-8    	10888167	       108 ns/op	      48 B/op	       1 allocs/op
```

replacer.go
billglover: I've found another case where the result doesn't match what a user might expect.

* input: `placeholder {with \} brace} in name`
* got: `placeholder  brace} in name`
* expected: `placeholder  in name`

I haven't had a look to see how nested placeholders would fair against existing tests and I think this case starts to get close to a nested braces scenario. Trying to address this risks expanding scope of the PR a bit, so am open to guidance.

mholt: Ah, nice find... my first thought is that it's a bug, it should output what you expected.

Hopefully the fix is not too complex.

replacer.go
mholt: That's definitely a good edge case... but one I'd be surprised if we ever hit in real life (at this point). Let's remove the extra check and gain back a couple nanoseconds. I think it's OK if input `\}` outputs `\}` unless we have users complaining otherwise...

mholt:
Performance-wise, it's looking good. If we keep it in this range for any remaining, last-minute changes, LGTM!

replacer.go
billglover: Will push a version with the extra check removed and tests modified accordingly. Benchmarks are in the same ballpark but I'll re-post for confirmation.

billglover:
Benchmarks for: e385e0d485eb3780f374875d7359811ced7dbcb9

```plain
BenchmarkReplacer/no_placeholder-8         	160360214	         7.38 ns/op	       0 B/op	       0 allocs/op
BenchmarkReplacer/placeholder-8            	21445204	        54.8 ns/op	      32 B/op	       1 allocs/op
BenchmarkReplacer/escaped_placeholder-8    	 9841627	       122 ns/op	      48 B/op	       1 allocs/op
```

replacer.go
billglover: This one still has me stumped.

The key challenge is that this line is a very convenient way of finding the closing brace.

https://github.com/caddyserver/caddy/blob/217419f6d97885daf0fb8832a407626e578523f0/replacer.go#L133

Unfortunately this means it finds closing braces even if they are escaped. Going to spend a little more time this evening and see if I can add a check to see if the closing braces was escaped.

The alternative is to re-write the `for` loop to iterate until closing braces are found. I'm trying to avoid doing this.

mholt: I think a loop is fine; if the one that was found is preceded by a `\`, then start at that point and look ahead for the next one, until one that is not preceded by `\` is found. I bet for 99% of inputs it'll usually be the first one, so the loop won't have to iterate. When it is escaped, then there's a slight performance hit, but that's what we get for being able to escape syntactic characters...

billglover: Thank you for outlining this approach, and specifically why it might be an acceptable trade-off. I've added this check and pushed. I also added one more test to satisfy myself I wasn't going to run into an 'off by one' scenario. New benchmarks to follow.

billglover:
Benchmarks for: https://github.com/caddyserver/caddy/pull/3121/commits/681531b0d211878ba44a01512f4304a0c930b2bb

```plain
BenchmarkReplacer/no_placeholder-8         	158008672	         7.57 ns/op	       0 B/op	       0 allocs/op
BenchmarkReplacer/placeholder-8            	22138293	        53.8 ns/op	      32 B/op	       1 allocs/op
BenchmarkReplacer/escaped_placeholder-8    	10015346	       122 ns/op	      48 B/op	       1 allocs/op
```

billglover:
Many thanks for all the support and patience as I worked through this. I'm in the process of deploying Caddy for a small site at the moment. This should give me a little more experience using the application. As I do, I'll take a look at the open issues and see if anything stands out.

Issue #3021, opened 2020-02-04
Caddy+Smallstep integration (trusted certs for localhost and internal sites)


mholt:
Thanks to a recent agreement, I'll soon be working on integrating [Smallstep](https://smallstep.com/) into Caddy. :+1:

This should allow us to serve HTTPS for localhost and internal sites automatically, much like we already do for public sites via Let's Encrypt. Essentially, we want `https://localhost` or `https://example.local` to be able to be the default using _trusted_ certificates and _properly managed_ local/internal PKI.

In Caddy 1, `tls self_signed` simply generated a private key in memory and used it to self-sign a certificate that lasted for a week, and that's all: it wasn't trusted, the key wasn't reused, and there was no proper PKI. With this integration in v2, Caddy will be able to serve your local dev and internal sites over HTTPS using trusted certificates with a proper, long-lived CA. This brings the ACME protocol into local and internal environments, rather than hacking private keys together in memory.

In the future, we can probably extend this to properly-managed mTLS that is (near-)fully automatic in a cluster.

This feature is not a release blocker for Caddy 2.0. It may get into 2.1.

This issue is for tracking discussion of the proposal and getting feedback from interested users and companies. Please get involved!

Here are some questions to help bootstrap development discussion:

- What do you want/need most from this feature?
- Should HTTPS be the default for all sites, including local and internal ones?
- What kind of configuration should be exposed?
- What should the cert and CA default lifetimes be?
- (Anything else that comes to your mind)

/cc Smallstep: @mmalone @maraino @mikemaxey

noor3430n:
# noor3430n
https://story.snapchat.com/s/s:FakP-LFnRgyShBF1VsUTRA?loop=true

mmalone:
Hi everyone. I'm Mike, from smallstep.

We're super interested in uncovering use cases for this integration. To help get a conversation going I wanted to drop a couple ideas to gauge interest, and share a few links for anyone who wants to learn more about this stuff.

## Use cases

Integrating Caddy with smallstep's CA (called `step-ca`) will make it easy for users to run their own internal PKI -- like your own internal Let's Encrypt. In general, a certificate is a sort of credential that you can use to authenticate a server or a client (e.g., for clients it's a replacement for a username/password or API access token). Running your own CA means you can issue these credentials yourself, however you want, whenever you want.

Some use cases that we think may be interesting to the Caddy community are:

[...]

Issue #2502, opened 2019-03-05
Tls self_signed - Not working correctly since a few versions


magikstm:
## 1. Which version of Caddy are you using (`caddy -version`)?

Caddy 0.11.5 (non-commercial use only)

## 2. What are you trying to do?

Use tls self_signed for a local development Caddy server.

## 3. What is your Caddyfile?

    :2021 {
    root C:\sites\1
    log “” access2019.log “{remote} | {>User-Agent} | {user} | {when} | {method} | {uri} | {proto} | {status} | {size} | {latency_ms}”

    gzip
    on startup php_cgi2.bat &
    fastcgi / 127.0.0.1:6545 php
    tls self_signed
    }

## 4. How did you run Caddy (give the full command and describe the execution environment)?

I'm using Windows 7 x64.

I typed "caddy" in cmd in the folder containing caddy.exe.

## 5. Please paste any relevant HTTP request(s) here.

N/A

## 6. What did you expect to see?

Server starting and running normally.

## 7. What did you see instead (give full error messages and/or log)?

> 2019/03/04 19:39:27 self-signed: certificate has no names

## 8. Why is this a bug, and how do you think this should be fixed?

It doesn't respect docs here:
https://caddyserver.com/docs/tls

It worked correctly with Caddy 0.11.1.

## 9. What are you doing to work around the problem in the meantime?

I commented this line:
> tls self_signed

## 10. Please link to any related issues, pull requests, and/or discussion.

https://caddy.community/t/tls-self-signed-windows-7-x64-caddy-0-11-5/5299

mholt:
Thanks for the report. I could reproduce it. This one might require a change in CertMagic to fix fully.

STaRDoGG:
Any word on a fix for this? I just upgraded from 0.11.1 to 0.11.5 and have run into the same issue.

francislavoie:
@STaRDoGG see PR #2531, it's a work in progress.

linquize:
Should this issue be fixed in 1.0.0? Many users use self sign cert for development purpose

mholt:
I'll get around to it soon. It's not a blocker for 1.0, and I'm currently redesigning how self-signed certs are managed.

magikstm:
Is this issue still considered for Caddy 1.x?

I tested with Caddy 1.0.3 and I still have the issue.

mholt:
[...]

PR #3125, opened 2020-03-07
v2: Implement 'pki' app powered by Smallstep for localhost certificates

mholt:
This pull request adds support for localhost certificates. It creates a new App module called `pki` and a new Issuer module called `internal`.

In Caddy v1, users would write `tls self_signed` in the Caddyfile (#2502) to have Caddy generate a quick, self-signed certificate for use in local/dev environments. But the cert was only stored in memory, so its key had to continually be re-trusted every time you restarted Caddy. It was a quick and bad implementation.

In Caddy 2, we can do better than that. These days, internal/local certificates can be very useful when done properly, and there is now tooling in Go to make this possible.

This design implements a proper, fully-managed, self-contained PKI (Public Key Infrastructure) system. Caddy generates a root and intermediate certificate, stores the root key offline, and uses the intermediate to sign leaf certificates. The intermediate is short-lived by default (7 days) and the leaf certificates are even shorter (12 hours -- maybe shorter before long).

Features:

- Generated roots can be installed into the system trust store automatically (with authorization). We might even do this by default if the Caddyfile is used, since we expect most developers will want this.

- Intermediates get renewed automatically (and soon roots will too, but their default lifetime is currently 10 years, so we'll get around to this later)

- Command to uninstall root certificates (`caddy untrust`)

[mkcert](https://github.com/FiloSottile/mkcert) is a tool by @FiloSottile which generates certificates and installs them into the root stores for you. It's really popular because it's so useful in local dev environments. This PR is basically mkcert, but with short cert lifetimes, proper renewals, and totally automatic and integrated into your web server! Think of this like a "mkcert server" and powered by [Smallstep](https://smallstep.com/).

## Configuration

The minimal required configuration to add to your Caddy JSON is an automation policy that uses an "internal" issuer:

```
{
	"issuer": {
		"module": "internal"
	}
}
```

That will assume a CA named "local". You can define and customize CAs, including the default "local" CA with the `pki` app:

```
{
	"apps": {
		"pki": {
			"certificate_authorities": {
				"local": {
					"install_trust": true
				}
			}
		}
	},
	...
}
```

That config will install the root into your system trust store. Then:

Localhost certs

Of course, this can be used for any hostname or IP address, not just `localhost`. We expect most users will want to use `*.localhost` as well.

## WORK IN PROGRESS

This is not a finished PR, but please try it out and submit your feedback! Lots of TODOs to implement, and Caddyfile support is yet to come.

A later PR will add support for running an ACME server.

cf. #3021

/cc @mmalone @maraino -- I will have specific questions for your review sooner or later. Thanks!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,9 +4,9 @@ go 1.14
 
 require (
 	github.com/Masterminds/sprig/v3 v3.0.2
-	github.com/alecthomas/chroma v0.7.1
+	github.com/alecthomas/chroma v0.7.2-0.20200305040604-4f3623dce67a
 	github.com/andybalholm/brotli v1.0.0
-	github.com/caddyserver/certmagic v0.10.0
+	github.com/caddyserver/certmagic v0.10.1
 	github.com/dustin/go-humanize v1.0.1-0.20200219035652-afde56e7acac
 	github.com/go-acme/lego/v3 v3.4.0
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e
@@ -24,8 +24,8 @@ require (
 	github.com/smallstep/cli v0.14.0-rc.3
 	github.com/smallstep/truststore v0.9.4
 	github.com/vulcand/oxy v1.0.0
-	github.com/yuin/goldmark v1.1.24
-	github.com/yuin/goldmark-highlighting v0.0.0-20200218065240-d1af22c1126f
+	github.com/yuin/goldmark v1.1.25
+	github.com/yuin/goldmark-highlighting v0.0.0-20200307114337-60d527fdb691
 	go.uber.org/zap v1.14.0
 	golang.org/x/crypto v0.0.0-20200302210943-78000ba7a073
 	golang.org/x/net v0.0.0-20200301022130-244492dfa37a
@@ -33,3 +33,5 @@ require (
 	gopkg.in/square/go-jose.v2 v2.4.1 // indirect
 	gopkg.in/yaml.v2 v2.2.8
 )
+
+replace github.com/caddyserver/certmagic => ../../certmagic
--- a/go.mod
+++ b/go.mod
@@ -33,5 +33,3 @@ require (
 	gopkg.in/square/go-jose.v2 v2.4.1 // indirect
 	gopkg.in/yaml.v2 v2.2.8
 )
-
-replace github.com/caddyserver/certmagic => ../../certmagic
--- a/modules/caddypki/crypto.go
+++ b/modules/caddypki/crypto.go
@@ -1,0 +1,150 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package caddypki
+
+import (
+	"bytes"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"strings"
+)
+
+func pemDecodeSingleCert(pemDER []byte) (*x509.Certificate, error) {
+	pemBlock, remaining := pem.Decode(pemDER)
+	if pemBlock == nil {
+		return nil, fmt.Errorf("no PEM block found")
+	}
+	if len(remaining) > 0 {
+		return nil, fmt.Errorf("input contained more than a single PEM block")
+	}
+	if pemBlock.Type != "CERTIFICATE" {
+		return nil, fmt.Errorf("expected PEM block type to be CERTIFICATE, but got '%s'", pemBlock.Type)
+	}
+	return x509.ParseCertificate(pemBlock.Bytes)
+}
+
+func pemEncodeCert(der []byte) ([]byte, error) {
+	return pemEncode("CERTIFICATE", der)
+}
+
+// pemEncodePrivateKey marshals a EC or RSA private key into a PEM-encoded array of bytes.
+// TODO: this is the same thing as in certmagic. Should we reuse that code somehow? It's unexported.
+func pemEncodePrivateKey(key crypto.PrivateKey) ([]byte, error) {
+	var pemType string
+	var keyBytes []byte
+	switch key := key.(type) {
+	case *ecdsa.PrivateKey:
+		var err error
+		pemType = "EC"
+		keyBytes, err = x509.MarshalECPrivateKey(key)
+		if err != nil {
+			return nil, err
+		}
+	case *rsa.PrivateKey:
+		pemType = "RSA"
+		keyBytes = x509.MarshalPKCS1PrivateKey(key)
+	case *ed25519.PrivateKey:
+		var err error
+		pemType = "ED25519"
+		keyBytes, err = x509.MarshalPKCS8PrivateKey(key)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("unsupported key type: %T", key)
+	}
+	return pemEncode(pemType+" PRIVATE KEY", keyBytes)
+}
+
+// pemDecodePrivateKey loads a PEM-encoded ECC/RSA private key from an array of bytes.
+// Borrowed from Go standard library, to handle various private key and PEM block types.
+// https://github.com/golang/go/blob/693748e9fa385f1e2c3b91ca9acbb6c0ad2d133d/src/crypto/tls/tls.go#L291-L308
+// https://github.com/golang/go/blob/693748e9fa385f1e2c3b91ca9acbb6c0ad2d133d/src/crypto/tls/tls.go#L238)
+// TODO: this is the same thing as in certmagic. Should we reuse that code somehow? It's unexported.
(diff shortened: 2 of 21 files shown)


modules/caddytls/internalissuer.go
```diff
@@ -0,0 +1,199 @@
...
+	cfg := &authority.Config{
+		Address:          "placeholder_Address:1",
+		Root:             []string{"placeholder_Root"},
+		IntermediateCert: "placeholder_IntermediateCert",
+		IntermediateKey:  "placeholder_IntermediateKey",
+		DNSNames:         []string{"placeholder_DNSNames"},
+		AuthorityConfig: &authority.AuthConfig{
+			Provisioners: provisioner.List{},
+		},
+	}
```
mholt: @mmalone @maraino Do you know if there's any chance we can drop these placeholders/meaningless values?

modules/caddytls/internalissuer.go
```diff
@@ -0,0 +1,199 @@
...
+	return func(p x509util.Profile) error {
+		fn := x509util.WithNotBeforeAfterDuration(notBefore, notAfter, time.Duration(d))
+		if err := fn(p); err != nil {
+			return err
+		}
+		crt := p.Subject()
+		crt.NotBefore = crt.NotBefore.Add(backdate)
+		return nil
+	}
+}
```
mholt: @mmalone @maraino It would be nice if there was an easier way to simply specify the certificate lifetime. I just feel kinda not-awesome copying unexported code out of your repo, especially for keeping up with changes. Open to discussing ideas here!

modules/caddypki/crypto.go
```diff
@@ -0,0 +1,150 @@
...
+	"crypto/ed25519"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"strings"
+)
+
+func pemDecodeSingleCert(pemDER []byte) (*x509.Certificate, error) {
```
mholt: @mmalone @maraino Would you be willing to add exported versions of these functions (the ones in this file here) in your `x509util` package? At least `pemDecodeSingleCert`, `pemEncodePrivateKey`, and `pemDecodePrivateKey`? (Or maybe they do exist and I just glossed over them)

mholt:
Some remaining questions pertain mostly to automatic HTTPS.

I intend for all sites -- including localhost/dev ones -- to use HTTPS by default. But this is a bit tricky in practice.

- Most dev machines can't conveniently bind to low ports (80 and 443)
- It's not uncommon for production configs to have both public-facing names (e.g. `example.com`) mixed with local-bound ones (e.g. `127.0.0.1:5000`, for internal services). The public-facing name would need to use a public CA and the IP address would use Caddy's CA, but that could potentially cause trust issues if accessed from another machine...

These are the two main pain points I haven't resolved yet.

More visually, what behavior results from these Caddyfiles:

```
localhost

# dev site
...
```

```
example.com {
   # production site
    ...
}

10.0.1.3:5000 {
   # internal service; and what if the port was omitted?
   ...
}
```

We'll have the ability to serve all these over HTTPS, but in the first case, binding to port 443 might be problematic. In the second case, using a non-publicly trusted cert means that any client accessing it must trust Caddy's root -- not a big deal, but just not something that is automatic yet.

Oh, and what about this:

```
:5000
...
```

?

xdevs23:
I'd suggest use tls by default only on hostnames without a port, i. e. `localhost`, `foo.bar.localhost`, `foo.bar`, ...

Don't use tls on `localhost:5050`, `foo.bar:8080`, `127.0.0.1`, `0.0.0.0:8088`, `:5000`, ...

Only exception would be for the port `443` in which case I'd expect a HTTPs endpoint.

mholt:
@xdevs23 Thanks for your input! I don't think that's going to fly though, because most dev environments can't (freely) bind to low ports, so defaulting to low ports for internal/localhost names will usually result in an error.

I think this is what I'll end up doing for now:

- Keep public-looking hosts (e.g. `example.com`) the same in terms of auto HTTPS
- Make HTTPS opt-in for non-public-looking hosts (e.g. `localhost` or `127.0.0.1`); to opt-in:
   - Use `tls internal` directive,
   - Or specify `https://` (with or without port -- as a special case, port 443 implies `https://`)
   - A global option to use the internal issuer
- ACME and internal issuers cannot be mixed in a site block
   - If `https://localhost` and `example.com` are specified for the same site block, use local issuer, I guess...

xdevs23:
> because most dev environments can't (freely) bind to low ports, so defaulting to low ports for internal/localhost names will usually result in an error

Hasn't this always been the case? Using automatic tls only changes the bound ports from just 80 to 80 and 443. Also, since there is a Docker image of Caddy v2, I think that might be the best solution overall and would also solve this issue.

mholt:
> Hasn't this always been the case?

Yes, and no, because automatic HTTPS never triggered in dev environments by default. If anyone wanted to use a public DNS name in a local dev environment, they'd turn off automatic HTTPS using `http://` or port `:80` unless they're set up to get a public cert for the name on their local environment.

> Using automatic tls only changes the bound ports from just 80 to 80 and 443.

That is so untrue. Automatic HTTPS does a lot more to work properly: https://caddyserver.com/docs/automatic-https#effects - and the default port is 2015, not 80 (because of the low port permissions problem). This has always been the case, since Caddy v0.5.0 back in early 2015.

> Also, since there is a Docker image of Caddy v2, I think that might be the best solution overall and would also solve this issue.

I don't think Docker has anything to do with it.

xdevs23:
> That is so untrue. Automatic HTTPS does a lot more to work properly: https://caddyserver.com/docs/automatic-https#effects - and the default port is 2015, not 80 (because of the low port permissions problem). This has always been the case, since Caddy v0.5.0 back in early 2015.

I got something mixed up then, my bad.

> I don't think Docker has anything to do with it.

Docker has the permission to bind low ports which means Caddy can do that, too, as long as Caddy itself inside the Docker container has that permission, too.

I think you're right — binding 80 and 443 by default for non-public domains is not a good idea. Perhaps a global directive explicitly allowing this could be the solution – or `tls on`.

mholt:
In the case of Docker or any other environment where you are able to bind to low ports, you can certainly opt-in to the automatic HTTPS for an internal name by omitting the port and using `https://`:

- `https://localhost`
- `localhost:443`

or

```
localhost

tls internal
```

... I think, anyway. That's what I'm working on today.

xdevs23:
I think the `https://` approach might be the best in this case.

mholt:
Well, my point is that both methods are valid; `tls internal` will be needed if the site address _looks_ like a public DNS name but isn't (or when you want to use an internal cert anyway).

xdevs23:
Yeah, I think that's a good approach.

mholt:
@mmalone @maraino Ready for review! Will probably merge this soon.

mholt:
Wow our CI tests disappeared

francislavoie:
Woohoo! CI is back :tada:

mholt:
@Mohammed90 I got our CI back, but it's still broken... I had to delete the old pipelines and re-make it... any chance you could help? 😅

go.mod
```diff
@@ -24,12 +24,14 @@ require (
...
+	github.com/yuin/goldmark-highlighting v0.0.0-20200307114337-60d527fdb691
 	go.uber.org/zap v1.14.0
 	golang.org/x/crypto v0.0.0-20200302210943-78000ba7a073
 	golang.org/x/net v0.0.0-20200301022130-244492dfa37a
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 	gopkg.in/square/go-jose.v2 v2.4.1 // indirect
 	gopkg.in/yaml.v2 v2.2.8
 )
+
+replace github.com/caddyserver/certmagic => ../../certmagic
```
Mohammed90: @mholt  Remove this -- should fix CI ;)

mholt: D'oh. Not even sure how that got committed. Nice catch... sigh.

mholt: FWIW I had to reset the FUZZIT_API_KEY too, so I'm not sure if I did that right. We'll see. I'm not even sure how to re-enable the checks, it seems I had to manually do a run on the pki branch from the Azure control panel.

mholt:
Gah, I pushed a small commit and the checks disappeared again, now the only check is CLA Assistant again... 😓

go.mod
Mohammed90: I think the Fuzzit API key wasn't affected because the job runs fine and shows progress

https://dev.azure.com/mholt-dev/Caddy/_build/results?buildId=964&view=logs&j=6b957288-3301-5435-9f73-758a5678d59e&t=a29c7330-30ca-5cb6-0170-5e46cdbd1b2e

mholt: Ah good! Thanks for checking!

Mohammed90:
Let's try adding [PR trigger](pr:
- master) to the `azure-pipelines.yml` file. At the top, add this:

```yaml
pr:
- v2
```

mholt:
@Mohammed90 Hm, but we didn't have to do that before, right?

Anyway, we can try it if you want, but in a separate PR / let's chat on Slack -- I'm gonna merge this up for now. Thanks!

mholt:
There is one question as to the auto-redirects on port 80. If a user can't bind to low ports, they can just specify higher ones in their config, but auto-redirects still happen from port 80 (unless the `http_port` is changed). I suppose that's fine for now, but I bet we're gonna get questions about this. 👌

Bawa75:
Hello, finally you have found a way to launch https://localhost on the current V2. If yes please help me

Bawa75:
Hello @mholt  , finally you have found a way to launch https://localhost on the current V2. If yes please help me

mohammed90:
> Hello, finally you have found a way to launch https://localhost on the current V2. If yes please help me

Thanks for your question, and we're thrilled that you're using Caddy! This looks more like a question about how to use Caddy rather than a bug report or feature request. Since this issue tracker is reserved for actionable development items, I'm going to close this, but we have a [community forum](https://caddy.community) where more people will be exposed to your question, including people who may be more expert or experienced with the specific question you're facing. I hope you'll ask your question there, and thanks for understanding!

PR #3128, opened 2020-03-09
caddyhttp: New algorithm for auto HTTP->HTTPS redirects

mholt:
It's still not perfect but I think it should be more correct for
slightly more complex configs. Might still fall apart for complex
configs that use on-demand TLS or at a large scale (workarounds are
to just implement your own redirects, very easy to do anyway).

Should fix #3127 but will need to be more verified first.

/cc @anoadragon453 - please try it soon and let me know!

PR #3129, opened 2020-03-09
caddyhttp: Fix terms error by provisioning new ACMEIssuer

anoadragon453:
This is an attempt to fix the `must agree to terms of service` error that originated in discussion [here](https://github.com/caddyserver/caddy/pull/3128#issuecomment-596708316). Somehow we were attempting ACME requests without first agreeing to the terms (at least as far as the ACME server knew).

This commit spawned from instructions in this comment https://github.com/caddyserver/caddy/pull/3128#issuecomment-596807622 however are likely not a proper fix.

While it appeared that the change had solved everything, it seemed more likely that while jumping about older commits, `go build`ing and restarting caddy, the certificates were provisioned successfully thus making it so I no longer had the `must agree to terms of service` error, fix or not.

I suspect that perhaps deleting my certificates would allow me to test again, but I'll refer to @mholt's judgement before attempting that.

Also be aware that it's quite late here, so I'm going to pop off now, but will be back in the morning :)

### DIFF
--- a/modules/caddyhttp/autohttps.go
+++ b/modules/caddyhttp/autohttps.go
@@ -394,6 +394,7 @@ func (app *App) createAutomationPolicy(ctx caddy.Context) error {
 	if acmeIssuer.Challenges.HTTP.AlternatePort == 0 {
 		// don't overwrite existing explicit config
 		acmeIssuer.Challenges.HTTP.AlternatePort = app.HTTPPort
+		acmeIssuer.Provision(ctx)
 	}
 	if acmeIssuer.Challenges.TLSALPN == nil {
 		acmeIssuer.Challenges.TLSALPN = new(caddytls.TLSALPNChallengeConfig)


mholt:
Thanks for helping so much with this, it made this really easy!

>  While it appeared that the change had solved everything, it seemed more likely that while jumping about older commits, go building and restarting caddy, the certificates were provisioned successfully thus making it so I no longer had the must agree to terms of service error, fix or not.

Right. I do think we're onto something with a fix, but to be sure... first make sure we're on Let's Encrypt's staging endpoint ([see this comment](https://github.com/caddyserver/caddy/pull/3128#issuecomment-596823633)) and then try again (with the broken commit). To retry after that, first delete the certificates from disk that are in the data folder for the staging endpoint.

modules/caddyhttp/autohttps.go
```diff
@@ -394,6 +394,7 @@ func (app *App) createAutomationPolicy(ctx caddy.Context) error {
 	if acmeIssuer.Challenges.HTTP.AlternatePort == 0 {
 		// don't overwrite existing explicit config
 		acmeIssuer.Challenges.HTTP.AlternatePort = app.HTTPPort
+		acmeIssuer.Provision(ctx)
```
mholt: Move this down to line 405 (old) or 406 (new).

mholt:
@anoadragon453 Thank you for reporting and helping confirm a fix for this bug! Given that it's just one line and I've had a fair bit of issues over in the `pki` branch (#3125), I ended up just needing to add this line over there along with a refactor. So, I am going to close this as it is superseded by the other PR. Thanks again, and hope you will continue to participate!

anoadragon453:
Heh, sorry I went out for a long vacation and only just got caught up with things. Glad you got the fix in anyways!

PR #3132, opened 2020-03-10
v2: adapter: register config adapters as Caddy modules

Mohammed90:
This adds the benefit of being able to know what config adapters are available in a given Caddy binary by running `caddy list-modules`

### DIFF
--- a/caddyconfig/json5/json5.go
+++ b/caddyconfig/json5/json5.go
@@ -29,7 +29,7 @@ func init() {
 // Adapter adapts JSON5 to Caddy JSON.
 type Adapter struct{}
 
-func (_ Adapter) CaddyModule() caddy.ModuleInfo {
+func (Adapter) CaddyModule() caddy.ModuleInfo {
 	return caddy.ModuleInfo{
 		ID: "adapters.json5",
 		New: func() caddy.Module {
--- a/caddyconfig/json5/json5.go
+++ b/caddyconfig/json5/json5.go
@@ -23,7 +23,7 @@ import (
 )
 
 func init() {
-	caddy.RegisterModule(Adapter{})
+	caddyconfig.RegisterAdapter("json5", Adapter{})
 }
 
 // Adapter adapts JSON5 to Caddy JSON.
--- a/caddyconfig/json5/json5.go
+++ b/caddyconfig/json5/json5.go
@@ -17,7 +17,6 @@ package json5adapter
 import (
 	"encoding/json"
 
-	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
 	"github.com/ilibs/json5"
 )
@@ -28,15 +27,6 @@ func init() {
 
 // Adapter adapts JSON5 to Caddy JSON.
 type Adapter struct{}
-
-func (Adapter) CaddyModule() caddy.ModuleInfo {
-	return caddy.ModuleInfo{
-		ID: "adapters.json5",
-		New: func() caddy.Module {
-			return Adapter{}
-		},
-	}
-}
 
 // Adapt converts the JSON5 config in body to Caddy JSON.
 func (a Adapter) Adapt(body []byte, options map[string]interface{}) (result []byte, warnings []caddyconfig.Warning, err error) {
--- a/caddyconfig/json5/json5.go
+++ b/caddyconfig/json5/json5.go
@@ -17,16 +17,27 @@ package json5adapter
 import (
 	"encoding/json"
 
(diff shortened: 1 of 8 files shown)


Mohammed90:
That would be really nice, alas...
```
can't load package: import cycle not allowed
package github.com/caddyserver/caddy/v2
        imports github.com/caddyserver/caddy/v2/caddyconfig
        imports github.com/caddyserver/caddy/v2
```

Mohammed90:
I don't like this approach a whole-lot, but it centralizes the adapter/module registration and avoids the import cycle. How do you like it?

mholt:
Hmm, I don't like it either, because then we have to hard-code exceptions for every type of plugin.

How about, instead, we find a way to make it so the `caddy/v2` package doesn't import `caddyconfig`? I think we can do it if we're clever about the admin endpoint...

This might take some time/effort. If you're busy, this might not be a top priority right now. :+1:

Mohammed90:
After playing with it for a few days, this seems like it will require some serious refactoring.

mholt:
Yeah, it's not trivial.

mholt:
@Mohammed90 I've refactored the troublesome import out of the caddy package; see the (poorly-named) [admin-package](https://github.com/caddyserver/caddy/tree/admin-package) branch and see what magic is possible there!

Mohammed90:
Yes, that works like a charm! 😃 I'll clean it up and do a push in a few days

Mohammed90:
I'm on the fence about whether we should panic if registering any of the standard adapters fails. Like, should our registration be like this:

```go
func init() {
	if err := caddyconfig.RegisterAdapter("caddyfile", caddyfile.Adapter{ServerType: ServerType{}}); err != nil {
		panic(err)
	}
}
```

or like this as it is now:

```go
func init() {
	caddyconfig.RegisterAdapter("caddyfile", caddyfile.Adapter{ServerType: ServerType{}})
}
```

mholt:
Good question. I don't know if there's a "right" answer in this case either, but I decided to have `RegisterModule()` return an error and let the caller decide how they wanted to handle it, although in `init()` I dunno what you can do other than panic or log.Fatal -- and I think either are fine.

So, go with the error return I guess. Even if the error is non-nil and goes unchecked, the developer will very quickly realize their plugin isn't working...

caddyconfig/json5/json5.go
```diff
@@ -28,6 +29,15 @@ func init() {
 // Adapter adapts JSON5 to Caddy JSON.
 type Adapter struct{}
 
+func (_ Adapter) CaddyModule() caddy.ModuleInfo {
```
mholt: For all these `(_ Adapter)` you can just do `(Adapter)` for a cleaner look, FYI :+1:

mholt: And actually, instead of having each adapter declare their own ModuleInfo, you could even do this for them, since you have their name and instance value... does that make sense?

Mohammed90: Nope, you lost me there 😅

mholt: Oh... maybe a dumb idea / won't work. brb

caddyconfig/load.go
```diff
@@ -0,0 +1,153 @@
...
+					Code: http.StatusBadRequest,
+					Err:  fmt.Errorf("adapting config using %s adapter: %v", adapterName, err),
+				}
+			}
+			if len(warnings) > 0 {
+				respBody, err := json.Marshal(warnings)
+				if err != nil {
+					caddy.Log().Named("admin.api.load").Error(err.Error())
+				}
+				w.Write(respBody)
```
francislavoie: Nice, glad to see golangci/reviewdog doing their job :joy:

mholt: @francislavoie That's the only place I see it. Where else are you seeing it?

francislavoie: https://github.com/caddyserver/caddy/pull/3132/checks?check_run_id=522599373 gives a summary of all the issues

mholt: Ah, thanks.

I'll revisit this PR this weekend!

Mohammed90: Should I do
```go
if _, err := w.Write(respBody); err != nil {
    caddy.Log().Named("admin.api.load").Error(err.Error())
}
```
?

mholt: To be honest it's not an error I care about at all :joy: Can we just ignore it?

Mohammed90: We could explicitly ignore it 😅 As a project, we chose to have errcheck linting, so we should obey the linter we chose to be our boss  😇

PR #3145, opened 2020-03-14
WIP: install local ca prerequisities

sarge:


azure-pipelines.yml
```diff
@@ -56,12 +56,16 @@ jobs:
   - bash: |
       wget "https://dl.google.com/go/$(LATEST_GO).linux-amd64.tar.gz"
       sudo tar -C $(gorootDir) -xzf "$(LATEST_GO).linux-amd64.tar.gz"
+      sugo apt install libnss3-tools
```
francislavoie:
```suggestion
      sudo apt install libnss3-tools
```

sarge: lol thanks

mholt:
I wonder how much this will slow down our CI tests...

(Gosh darn it, where did they disappear to??)

Is this needed after https://github.com/caddyserver/caddy/commit/acf4dde1dd1fa2d30b76c19009e5a7019cce0b2b?

sarge:
They have been pretty fast, but of course not free. I did some premature optimisation to only check the necessary prerequisites once. My feeling is that some tests like this will be useful to prevent regressions, there is a lot of subtle variations to how the server can be configured it is pretty challenging to be aware of them all.

Thanks for the heads up that certificate trust is a warning. We don't need trust for these tests to be effective, because they are all self signed we skip verification anyway.

mholt:
Oh, this PR does two things. It installs nss libs _and_ runs the integration tests.

Aren't the integration tests already run by CI with our regular `go test` flow?

sarge:
Currently the integration tests expect a running instance of caddy to load the config into. 

I didn't want to assume that the version of caddy in the path was the one to start, so if the integration tests do not find a server on port 2019 the tests are skipped. I could change that and start and stop caddy, but that feels like it should be part of a build process? I don't know, everyones dev loop is a little bit different. What do you think?

I will pull out the nss code once the build server is running again.

sarge:
Also I had write a separate integration test script for windows path separators /\

mholt:
I think tests should be as self-contained as possible: `go test` should be what it takes to do the tests, and that probably means spinning up a server and shutting it down when complete. (Is that what you're asking?)

sarge:
That is possible to start caddy, we would have to make a bunch of assumptions. 
Where was it built? Is it in the path or is it under `/cmd/caddy` or %GOPATH%/bin/

We probably would need to put a lock in to prevent concurrent tests.

I run these tests once caddy is build and running

`go test ./caddytest/... -v -count=1`

sarge:
One last thought, is that instead of starting caddy as an external process. Would it be possible to host the server in-process?

mholt:
Yeah, just call `Main()`: https://github.com/caddyserver/caddy/blob/v2/cmd/main.go#L38

cmd/main.go
```diff
@@ -77,6 +77,11 @@ func Main() {
 	os.Exit(exitCode)
 }
 
+// GetCommand is provided to allow tests to extract the commands
+func GetCommand(name string) Command {
```
sarge: @mholt this is pretty cool, change running the app locally, but I did hit a snag getting hold of the run command. So this is my work around.

It is called here.
https://github.com/caddyserver/caddy/blob/62d86d892b0c9c7865903c6d2b9771a72c3eef35/caddytest/caddytest.go#L151

mholt: Instead, can you just manipulate `os.Args`? I know that's normally a no-no, but these are integration tests after all...

mholt:
@sarge We switched to GH Actions recently, when you are able, feel free to rebase and then we'll see about getting this merged. Thanks!

sarge:
@mholt I have these running, and pleased to have discovered the windows issue. It was all hanging on a password when attempting to place certs into the trust store.

Solved in the json case like this.
https://github.com/caddyserver/caddy/blob/7bd553ed6ebec8997fbdd65ca5579db8de1ff7e7/caddytest/integration/caddyfile_test.go#L218-L223

There isn't currently a way to disable trust in the `caddyfile`, perhaps we could something in there or command line or ENV to prevent attempting the cert install. Or perhaps there is an easier way to configure this?

mholt:
@sarge Sure, but, probably later. Can any tests that require that be removed (preferably not just "disabled" or commented out -- trying to avoid dead code in the code base) and then we can merge this?

PR #3146, opened 2020-03-14
added sni tests

sarge:
Expanding the SNI tests. 

Ideally with an SNI less connection we don't to know the destination IP address, things in the cloud are load balanced, DNS changes.

This change supports this port only matching configuration
```
 {
    http_port     9080
    https_port    9443
    default_sni   a.caddy.localhost
  }

:9443 {
    tls /a.caddy.localhost.crt /a.caddy.localhost.key {
    }
    respond /version 200 {
      body "hello from a.caddy.localhost"
    }	
}
```

The configuration I would like to support is 

```
 {
    http_port     9080
    https_port    9443
    default_sni   a.caddy.localhost
  }

 a.caddy.localhost:9443 {
    tls /a.caddy.localhost.crt /a.caddy.localhost.key {
    }
    respond /version 200 {
      body "hello from a.caddy.localhost"
    }	
  }
```
Which would respond to `curl https://127.0.0.1/version`. In this case the CP matches perfectly but the host matcher does not. 

I guess an alternative could be, with a wildcard IP matcher.
```
 a.caddy.localhost:9443, 0.0.0.0:9443 {
    tls /a.caddy.localhost.crt /a.caddy.localhost.key {
    }
    respond /version 200 {
      body "hello from a.caddy.localhost"
    }	
  }
```


caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -519,10 +519,10 @@ func (st *ServerType) serversFromPairings(
 					for _, h := range hosts {
 						if h == defaultSNI {
 							hosts = append(hosts, "")
-							cp.DefaultSNI = defaultSNI
 							break
 						}
 					}
+					cp.DefaultSNI = defaultSNI
```
mholt: I don't think adding this value to _every_ connection policy is needed or useful... why is this necessary?

sarge: In this example it is needed to support this test.

```
{
    http_port     9080
    https_port    9443
    default_sni   a.caddy.localhost
  }

:9443 {
    tls /a.caddy.localhost.crt /a.caddy.localhost.key {
    }
    respond /version 200 {
      body "hello from a.caddy.localhost"
    }	
}
```
I think this is a reasonable config, what do you think? Alternatively I was thinking about adding an wildcard IP matcher `0.0.0.0:9443`

mholt: (BTW: did you know you have an unnecessary `{\n }` in your `tls` directive?)

The config is reasonable, I just don't think the fix is. Specifically: there is/should be _one_ TLS connection policy that matches the empty SNI case (or is a catch-all), and that one should have DefaultSNI set. Not all of them. Makes for kind of a sloppy JSON...

sarge: Hmm ok, I saw the default being fallback to this if there is no other option. But I guess when can calculate when this will be needed.

So we can add defaultSNI for the following scenarios:
  1. the hostname and the sni name are the same
  2. if this will be the default CP by setting the DefaultSNI https://github.com/caddyserver/caddy/blob/v2/caddyconfig/httpcaddyfile/httptype.go#L533?
  3. if there is an IP matcher specified in the host list `0.0.0.0` which I think is required for https://github.com/caddyserver/caddy/pull/3146/files/3ee69e48e63abcd385ff6a57fa228828cde4ef53#diff-4119c85db3fe58fb3e309290dd219b39R119

Does that sound better?

mholt: > [edit] Some time ago we were looking at certmagic.Default.DefaultServerName = defaultSNI which applies to call cert lookups

Yeah, but I determined that it's not necessary to set this (and is almost certainly racey).

DefaultSNI is not needed to select a TLS connection policy, only to select a certificate; connection policies are already capable of being matched to an empty SNI.

> the hostname and the sni name are the same

If by this, you mean https://github.com/caddyserver/caddy/blob/7a4548c5823e85bab0a2e2f40a3ea00f64ce8552/caddyconfig/httpcaddyfile/httptype.go#L520, then yes; that's what we already do.

> if this will be the default CP by setting the DefaultSNI

We already do that too, but lower down: https://github.com/caddyserver/caddy/blob/7a4548c5823e85bab0a2e2f40a3ea00f64ce8552/caddyconfig/httpcaddyfile/httptype.go#L609-L610

> if there is an IP matcher specified in the host list 0.0.0.0

I don't understand what you mean by this.

sarge: Lets bite off the first issue.

In the test `TestDefaultSNIWithPortMappingOn` 

Because there is no hostname to match to the defaultSNI, the defaultSNI is not set.
Because there are no hosts, it will then set `hasCatchAllTLSConnPolicy` to true, so never enters
https://github.com/caddyserver/caddy/blob/v2/caddyconfig/httpcaddyfile/httptype.go#L609-L611

Which results in a connection policy with no defaultSNI configured.

A suggestion I have is to set the defaultSNI when no hostnames are matched
https://github.com/caddyserver/caddy/blob/v2/caddyconfig/httpcaddyfile/httptype.go#L533

mholt: > Because there is no hostname to match to the defaultSNI, the defaultSNI is not set.

Are you sure? The variable `defaultSNI` is _always_ obtained from the global options and should definitely be set, regardless of any server blocks: https://github.com/caddyserver/caddy/blob/7a4548c5823e85bab0a2e2f40a3ea00f64ce8552/caddyconfig/httpcaddyfile/httptype.go#L450

The line you linked to _does_ set the SNI...:

> `srv.TLSConnPolicies = append(srv.TLSConnPolicies, &caddytls.ConnectionPolicy{DefaultSNI: defaultSNI})`

sarge: Yes the defaultSNI is read, but never applied. I was trying to link to the guard clause which checks `hasCatchAllTLSConnPolicy`. Which is set further up when no hosts were matched.

Just like caddy load you can use all of the adaptable formats. I do think it is valuable including both, there are a lot of subtle checks going on within the adaptive process, being able to make a curl request against the server is pretty definitive result.

```
func TestDefaultSNIWithPortMappingOnly(t *testing.T) {

// arrange
caddytest.InitServer(t, ` 
  {
    http_port     9080
    https_port    9443
    default_sni   a.caddy.localhost
  }
  
  :9443 {
    tls /a.caddy.localhost.crt /a.caddy.localhost.key
    respond /version 200 {
      body "hello from a.caddy.localhost"
    }	
  }
  `, "caddyfile")

	// act and assert
	caddytest.AssertGetResponse(t, "https://a.caddy.localhost:9443/version", 200, "hello from a.caddy.localhost")
	// makes a request with no sni
	caddytest.AssertGetResponse(t, "https://127.0.0.1:9443/version", 200, "hello from a.caddy.localhost")
}

```


When the test fails it does a load from caddy admin and prints it, which I have found useful.



```
    TestDefaultSNIWithPortMappingOnly: caddytest.go:225: failed to call server Get "https://127.0.0.1:9443/version": remote error: tls: internal error
    TestDefaultSNIWithPortMappingOnly: caddytest.go:211: requesting "https://127.0.0.1:9443/version" expected response body "hello from a.caddy.localhost" but got ""
    TestDefaultSNIWithPortMappingOnly: caddytest.go:78: ----------- failed with config -----------
        {
          "apps": {
            "http": {
              "http_port": 9080,
              "https_port": 9443,
              "servers": {
                "srv0": {
                  "listen": [
                    ":9443"
                  ],
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "hello from a.caddy.localhost",
                          "handler": "static_response",
                          "status_code": 200
                        }
                      ],
                      "match": [
                        {
                          "path": [
                            "/version"
                          ]
                        }
                      ]
                    }
                  ],
                  "tls_connection_policies": [
                    {
                      "certificate_selection": {
                        "policy": "custom",
                        "tag": "cert0"
                      }
                    }
                  ]
                }
              }
            },
            "tls": {
              "certificates": {
                "load_files": [
                  {
                    "certificate": "/home/fedora/workspaces/gallagher/third-party/caddy/caddytest/a.caddy.localhost.crt",
                    "key": "/home/fedora/workspaces/gallagher/third-party/caddy/caddytest/a.caddy.localhost.key",
                    "tags": [
                      "cert0"
                    ]
                  }
                ]
              }
            }
          }
        }

```



Btw what syntax are you using to include the code snippets? I would like use them to help in our discussion.

mholt: Oh, gotcha.

Yeah, okay, we probably do need to set DefaultSNI here then: https://github.com/caddyserver/caddy/blob/7a4548c5823e85bab0a2e2f40a3ea00f64ce8552/caddyconfig/httpcaddyfile/httptype.go#L532-L534

To get the snippets, press Y when you're viewing a code file, and GitHub will turn the URL into a permanent one (by pinning it to the specific commit) and then I think that's what causes the snippet to be inlined.

sarge: Okay so my final SNI scenario is not so much about SNI but host matching on any IP address.

```
  {
    http_port     9080
    https_port    9443
    default_sni   a.caddy.localhost
  }
  
  a.caddy.localhost:9443 {
    tls /a.caddy.localhost.crt /a.caddy.localhost.key
    respond /version 200 {
      body "hello from a.caddy.localhost"
    }	
  }
```

Which translates to 

```
  {
          "apps": {
            "http": {
              "http_port": 9080,
              "https_port": 9443,
              "servers": {
                "srv0": {
                  "listen": [
                    ":9443"
                  ],
                  "routes": [
                    {
                      "handle": [
                        {
                          "handler": "subroute",
                          "routes": [
                            {
                              "handle": [
                                {
                                  "body": "hello from a.caddy.localhost",
                                  "handler": "static_response",
                                  "status_code": 200
                                }
                              ],
                              "match": [
                                {
                                  "path": [
                                    "/version"
                                  ]
                                }
                              ]
                            }
                          ]
                        }
                      ],
                      "match": [
                        {
                          "host": [
                            "a.caddy.localhost"
                          ]
                        }
                      ],
                      "terminal": true
                    }
                  ],
                  "tls_connection_policies": [
                    {
                      "certificate_selection": {
                        "policy": "custom",
                        "tag": "cert0"
                      },
                      "default_sni": "a.caddy.localhost",
                      "match": {
                        "sni": [
                          "a.caddy.localhost",
                          ""
                        ]
                      }
                    },
                    {
                      "default_sni": "a.caddy.localhost"
                    }
                  ]
                }
              }
            },
            "tls": {
              "certificates": {
                "load_files": [
                  {
                    "certificate": "/home/fedora/workspaces/gallagher/third-party/caddy/caddytest/a.caddy.localhost.crt",
                    "key": "/home/fedora/workspaces/gallagher/third-party/caddy/caddytest/a.caddy.localhost.key",
                    "tags": [
                      "cert0"
                    ]
                  }
                ]
              }
            }
          }
        }

```

calling this server using  `curl https://127.0.0.1/version` will match the empty SNI match but fail to return the correct result, because the hostname failed to match on the handler.

I can make this will work as long as I know the IP address exactly, I would do this by adding the IP to the hostname e.g. ` a.caddy.localhost:9443,127.0.0.1:9443 {`.
But often the IP addresses are dynamic, load balanced etc. What I really want to do is add a matcher to match on any IP address, if a request comes in without a hostname use this server to fullfill the request. Similar to how we identify the empty hostname for SNI.

I was thinking about ` a.caddy.localhost:9443, 0.0.0.0:9443 {` to indicate that requests without hostnames should be handled by this server.

```
 "host": [
    "a.caddy.localhost",
    "0.0.0.0"
]
```
`*` might be better than `0.0.0.0` or perhaps we could use a CDIR mask instead? Or use an empty hostname "" like the SNI to match against any request with no hostname, like an IP address?

mholt: > calling this server using curl https://127.0.0.1/version will match the empty SNI match but fail to return the correct result, because the hostname failed to match on the handler.

So then, the TLS handshake is correct and working properly?

It sounds like what is happening is expected behavior. That handler is only reached for requests to an HTTP Host of `a.caddy.localhost`, not `127.0.0.1`.

sarge: I think you are right, I should use the host header-less version. I will remove the failing scenario. Thanks for talking it through with me:)

caddytest/integration/caddyfile_test.go
```diff
@@ -78,14 +78,61 @@ func TestDefaultSNI(t *testing.T) {
...
+  }
+  `, "caddyfile")
+
+	// act and assert
+	caddytest.AssertGetResponse(t, "https://a.caddy.localhost:9443/version", 200, "hello from a.caddy.localhost")
+	// makes a request with no sni
+	caddytest.AssertGetResponse(t, "https://127.0.0.1:9443/version", 200, "hello from a.caddy.localhost")
+}
+
+func xTestDefaultSNIWithPortMappingOnly(t *testing.T) {
```
mholt: Intentional to skip this test?

caddytest/integration/caddyfile_test.go
```diff
@@ -78,14 +78,61 @@ func TestDefaultSNI(t *testing.T) {
...
+      body "hello from a.caddy.localhost"
+    }	
+  }
+  `, "caddyfile")
+
+	// act and assert
+	caddytest.AssertGetResponse(t, "https://127.0.0.1:9443/version", 200, "hello from a.caddy.localhost")
+}
+
+func xTestDefaultSNIWithNamedHostAndExplicitIP(t *testing.T) {
```
mholt: And this one too

caddytest/integration/caddyfile_test.go
sarge: I disabled it because it relies on the hostname in the test. What I should do it comment out / delete the request that makes the hostname call instead.
The lack of hostname support is a bit of a pain, I will investigate again once the new CI is up and running.

mholt:
Cool, so should we merge this?

PR #3149, opened 2020-03-15
WIP: Set up work-in-progress github actions workflows

francislavoie:
This may or may not be nicer than Azure Pipelines (same backend, different UI, slightly different config)

PR #3152, opened 2020-03-17
Switch to Github Actions

francislavoie:
This may or may not be nicer than Azure Pipelines (same backend, different UI, slightly different config)

Supersedes #3149, trying to get CI to trigger.


francislavoie:
Well that's good, CI triggered but it's obviously not ready yet. Still WIP.

francislavoie:
Well, tests work, lint doesn't yet, integration tests also don't work yet.

Gonna put this off until another day 😥

mholt:
Why do we need to use "reviewdog" anyway - what is it?

/cc @sarge about the integration tests

sarge:
Yes @mholt we can drop the integration tests, we can run them in process - so we don't have to start and stop a separate caddy process.

francislavoie:
> Why do we need to use "reviewdog" anyway - what is it?

golangci-lint wasn't installing on windows, so I googled alternatives and found that - thought it looked cool and figured it might be worth trying, adds comments inline to PRs for the reported issues. See https://github.com/reviewdog/action-golangci-lint

I'm not married to it, I was just trying something :stuck_out_tongue: 

Acknowledged @sarge I'll take that out :+1:

francislavoie:
I... think I did it, guys!

🎉 

I'm not sure what to do about test/coverage reporting, would appreciate some thoughts there.

Please review!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: ${{ matrix.go-version }}
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Install test and coverage analysis tools
```
Mohammed90: Those were needed because Azure uses JUnit XML format for test result reporting. I haven't found anything as straightforward for Github Actions. There's a workaround described here:
https://github.community/t5/GitHub-Actions/Publishing-Test-Results/m-p/37236/highlight/true#M2860

Seems like a bit of a hassle though. Let's research more. Otherwise, we'll have to pull more tooling, unfortunately.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+      run: |
+        go get -v -t -d ./...
+        # mkdir test-results
+
+    # Force bash shell, this command fails on windows if using pwsh
+    - name: Build Caddy
+      working-directory: ./cmd/caddy
+      shell: bash
+      run: |
+        CGO_ENABLED=0 go build -trimpath -a -ldflags="-w -s" -v
```
Mohammed90: What if we set `CGO_ENABLED=0` as `env:`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+      run: |
+        mkdir coverage
+        gocov convert cover-profile.out > coverage/coverage.json
+        # Because Windows doesn't work with input redirection like *nix, but output redirection works.
+        (cat ./coverage/coverage.json | gocov-xml) > coverage/coverage.xml
+
+    # TODO publish coverage/lint/test results ?
+
+    # To return the correct result even though we set 'continue-on-error: true'
+    - name: Coerce correct build result (Windows)
```
Mohammed90: These were needed earlier to bypass the test failures so we can publish the test results. Keeping or removing this step depends on how end up setting up the publishing of test/coverage/lint results.

.github/workflows/fuzzing.yml
```diff
@@ -0,0 +1,75 @@
+name: Fuzzing
+
+on:
+  schedule:
```
Mohammed90: Fuzzers run in 2 scenarios:
- On PR, where they'll run with `--fuzz-type local-regression` in which it'll download the corpus from Fuzzit and run them against the new code.
- On scheduled run, where new fuzzer binaries will be built from HEAD of the branch and submitted to Fuzzit with `--fuzz-type fuzzing` to run through the next day.

How can we reuse the steps listed here?

.github/workflows/fuzzing.yml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: ${{ matrix.go-version }}
+
+    # - name: Checkout code
+    #   uses: actions/checkout@v2
```
Mohammed90: It needs the project code to build the fuzzers 😄

.github/workflows/fuzzing.yml
```diff
@@ -0,0 +1,75 @@
...
+
+    # TODO env GOBIN may not be set
+    # TODO should have if condition to disable for forks?
```
Mohammed90: I think it's fine to let forks trigger the fuzzers. They'll only run the regression fuzzers, so there's no possibility for abuse of CPU. We had to disable forks on Azure because Azure doesn't provide the secrets (e.g. env var API keys) to runners of forks.

.github/workflows/fuzzing.yml
```diff
@@ -0,0 +1,75 @@
...
+        sudo add-apt-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-7 main"
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+        sudo apt update && sudo apt install -y clang-7 lldb-7 lld-7
+
+        go get -v github.com/dvyukov/go-fuzz/go-fuzz github.com/dvyukov/go-fuzz/go-fuzz-build
+        wget -q -O fuzzit https://github.com/fuzzitdev/fuzzit/releases/download/v2.4.77/fuzzit_Linux_x86_64
+        chmod a+x fuzzit
+        mv fuzzit $(GOBIN)
+
+    # TODO $(Build.Reason) is probably not set
```
Mohammed90: This depends on how we set up having local-regression vs. scheduled fuzzers. It's just needed to figure out what should the var `fuzz_type` be.

francislavoie:
Some links/references I used:

- https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions
- https://github.com/mvdan/github-actions-golang
- https://github.com/reviewdog/action-golangci-lint
- https://stackoverflow.com/questions/57850553/github-actions-check-steps-status

.github/workflows/ci.yml
francislavoie: Ah, yeah that works. That would let us avoid `shell: bash`, right?

.github/workflows/ci.yml
francislavoie: Yep. I just wanted to keep these ideas intact, I wasn't sure how we'd do result reporting.

.github/workflows/ci.yml
Mohammed90: Yep, I believe so.

.github/workflows/fuzzing.yml
francislavoie: I think we use `${{ github.event }}` for this, after a bit of reading.

.github/workflows/fuzzing.yml
```diff
@@ -0,0 +1,77 @@
...
+        sudo add-apt-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-7 main"
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+        sudo apt update && sudo apt install -y clang-7 lldb-7 lld-7
+
+        go get -v github.com/dvyukov/go-fuzz/go-fuzz github.com/dvyukov/go-fuzz/go-fuzz-build
+        wget -q -O fuzzit https://github.com/fuzzitdev/fuzzit/releases/download/v2.4.77/fuzzit_Linux_x86_64
+        chmod a+x fuzzit
+        mv fuzzit $(go env GOPATH)/bin
+
+    - name: Download go-fuzz tools and the Fuzzit CLI, move Fuzzit CLI to GOBIN
```
Mohammed90: The name needs to be updated to match the actual content

francislavoie: Yep, was already doing that 😄 just forgot to fix that one when copy-pasting from the azure yml

francislavoie:
Fuzzing failure now is just due to the missing API key.

@mholt I think we'll need you to set the secret. See the instructions here https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

The secret key is `FUZZIT_API_KEY`

mholt:
Wow, great work @francislavoie! I can even see the test output in the Checks tab.

I set the API key as a secret. Please take another look!

This will be a relief to have working again :)

mholt:
```
FUZZER: ==18== ERROR: libFuzzer: fuzz target overwrites it's const input
FUZZER: SUMMARY: libFuzzer: out-of-memory
```

Is that different from a "crasher" caught by fuzzing?

francislavoie:
Yeah, regression testing fails for some reason. I think we might need to bug the fuzzit guys for some help. I have no idea what to do with that.

mholt:
I opened an issue upstream. https://github.com/fuzzitdev/fuzzit/issues/57

.github/workflows/ci.yml
```diff
@@ -0,0 +1,116 @@
...
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    # - name: Install test and coverage analysis tools
+    #   run: |
+    #     go get github.com/axw/gocov/gocov
+    #     go get github.com/AlekSi/gocov-xml
+    #     go get -u github.com/jstemmer/go-junit-report
+    #     echo "::add-path::$(go env GOPATH)/bin"
```
mholt: For the commented lines, can we add additional comment explaining why it's commented? #meta

(For all the commented sections, not just this one)

francislavoie: Done 👍

mholt: LOVE 'em, thanks!

PR #3159, opened 2020-03-19
reverseproxy: Add Alt-Svc to Hop-by-hop headers list

paolobarbolini:
Adds `Alt-Svc` to the list of headers that get removed when proxying to a backend.

This fixes the issue of having the contents of the Alt-Svc header duplicated when proxying to another Caddy server.

This was already blacklisted in Caddy v1

https://github.com/caddyserver/caddy/blob/891446d06340db2912c9a970bfe862bd54efbb70/caddyhttp/proxy/reverseproxy.go#L584-L600


mholt:
Thanks. Is this a result of enabling experimental HTTP/3?

paolobarbolini:
Exactly

PR #3166, opened 2020-03-21
Use matrix includes to set per-os variables, simplify cross-platform

francislavoie:
I felt it was weird that I had to make separate jobs per environment, I just didn't find any info on how to do it initially. Did a bit more digging, found this thread which explained it pretty clearly.

https://github.community/t5/GitHub-Actions/Create-matrix-with-multiple-OS-and-env-for-each-one/m-p/38574#M3418

Trying this out. Let's see if it works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,23 @@ jobs:
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         go-version: [ 1.14.x ]
+
+        # Set some variables per OS, usable via ${{ matrix.VAR }}
+        # CADDY_BIN_PATH: the path to the compiled Caddy binary, for artifact publishing
+        # SUCCESS: the typical value for $? per OS (Windows/pwsh returns 'True')
+        include:
+        - os: ubuntu-latest
+          CADDY_BIN_PATH: ./cmd/caddy/caddy
+          SUCCESS: 0
+
+        - os: macos-latest
+          CADDY_BIN_PATH: ./cmd/caddy/caddy
+          SUCCESS: 0
+
+        - os: windows-latest
+          CADDY_BIN_PATH: ./cmd/caddy/caddy.exe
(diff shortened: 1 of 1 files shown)


francislavoie:
Awesome, that's much better. 🎉

Mohammed90:
Neat! Can you add one more improvement into the same PR? Cache the downloaded Go modules to be efficient:

https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds

.github/workflows/ci.yml
```diff
@@ -59,19 +76,11 @@ jobs:
...
         name: caddy_v2_${{ matrix.os }}
-        path: ./cmd/caddy/caddy.exe
-
-    - name: Publish Build Artifact (Linux/Mac)
-      if: matrix.os != 'windows-latest'
-      uses: actions/upload-artifact@v1
-      with:
-        name: caddy_v2_${{ matrix.os }}
-        path: ./cmd/caddy/caddy
+        path: ${{ matrix.CADDY_BIN_PATH }}
```
mholt: It is _kind_ of a bummer that the artifacts names are all the same ("latest") so if one is downloaded, one can quickly forget which commit it's built from (since `caddy version` shows `(devel)`).

1. Would it be difficult to add the first 7-8 chars of the commit SHA to the filename, instead of `latest`? (Also, `ubuntu` -> `linux` maybe?)
2. And/or would it be difficult to make `caddy version` show the proper version?

francislavoie:
@Mohammed90 @mholt Those are all really good ideas, but I think I'll try to tackle those things in separate PRs if that's okay. I think there's value in getting this in `v2` now, I think Sarge can use this to set up CoreDNS per-platform for integration tests or whatever if that's something he still wants to attempt.

@mholt I think for the `caddy version` thing we should just use the `Magefile` that's planned, right? I definitely don't know enough about configuring the build to tackle that myself.

PR #3167, opened 2020-03-21
ci: Enable caching of go modules

francislavoie:
From https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,14 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    - name: Cache Go modules
+      uses: actions/cache@v1
+      with:
+        path: ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


francislavoie:
Well, the build is actually not any faster with caching. 🤔

It caches ~160MB, and it takes just as long to copy and untar than to just `go get`. It's only faster by a few seconds at best, totally within the margin of error.

Do we care to merge this? @mholt @Mohammed90

mholt:
Huh, you're right... I was just comparing your second CI run here with some previous ones and... it's not much faster, if at all.

Since it doesn't appear to help, let's skip this one, I guess?

francislavoie:
I'd say let's just go with the simplicity of no cache. Curious to hear @Mohammed90's thoughts before closing.

Mohammed90:
Same experience as #2821. I thought it'd probably do better, but no.

PR #3173, opened 2020-03-22
chore: ci: preliminary CD support with goreleaser

Mohammed90:
Let's see if this works

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,6 @@ on:
   push:
     branches: 
       - v2
-    tags:
-      - '*'
   pull_request:
     branches: 
       - v2
@@ -68,7 +66,6 @@ jobs:
         env
         # Calculate the short SHA1 hash of the git commit
         echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
-        echo "::set-output name=closest_tag::$(git describe --abbrev=0)"
 
     # The following steps are for 
     - name: Get dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,10 +48,6 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
-    # So GoReleaser can generate the changelog properly
-    - name: Unshallow
-      run: git fetch --prune --unshallow
-
     # These tools would be useful if we later decide to reinvestigate
     # publishing test/coverage reports to some tool for easier consumption
     # - name: Install test and coverage analysis tools
@@ -73,22 +69,6 @@ jobs:
         # Calculate the short SHA1 hash of the git commit
         echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
         echo "::set-output name=closest_tag::$(git describe --abbrev=0)"
-
-    # GoReleaser will take care of publishing those artifacts into the release
-    - name: Run GoReleaser
-      uses: goreleaser/goreleaser-action@v1
-      # if: startsWith(github.ref, 'refs/tags/')
-      with:
-        version: latest
-        args: release --rm-dist --skip-publish
-        # Figure out signing
-        # key: ${{ secrets.YOUR_PRIVATE_KEY }}
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        TAG: ${{ steps.vars.outputs.closest_tag }} #${{ github.ref }}
-
-    - name: Show project tree
-      run: tree
 
     # The following steps are for 
     - name: Get dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,6 @@ jobs:
         # Calculate the short SHA1 hash of the git commit
         echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
 
-    # The following steps are for 
     - name: Get dependencies
       run: |
         go get -v -t -d ./...
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,8 @@ on:
   push:
     branches: 
       - v2
+    tags:
+      - '*'
   pull_request:
     branches: 
       - v2
@@ -46,6 +48,10 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    # So GoReleaser can generate the changelog properly
+    - name: Unshallow
+      run: git fetch --prune --unshallow
+
     # These tools would be useful if we later decide to reinvestigate
     # publishing test/coverage reports to some tool for easier consumption
     # - name: Install test and coverage analysis tools
@@ -66,7 +72,25 @@ jobs:
         env
         # Calculate the short SHA1 hash of the git commit
         echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
+        echo "::set-output name=closest_tag::$(git describe --abbrev=0)"
 
+    # GoReleaser will take care of publishing those artifacts into the release
+    - name: Run GoReleaser
+      uses: goreleaser/goreleaser-action@v1
+      # if: startsWith(github.ref, 'refs/tags/')
+      with:
+        version: latest
+        args: release --rm-dist --skip-publish
+        # Figure out signing
+        # key: ${{ secrets.YOUR_PRIVATE_KEY }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        TAG: ${{ steps.vars.outputs.closest_tag }} #${{ github.ref }}
+
+    - name: Show project tree
+      run: tree
+
+    # The following steps are for 
     - name: Get dependencies
       run: |
         go get -v -t -d ./...
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,10 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    # So GoReleaser can generate the changelog properly
+    - name: Unshallowify the repo clone
+      run: git fetch --prune --unshallow
+
     - name: Print Go version and environment
       id: vars
       run: |
@@ -35,12 +39,7 @@ jobs:
         go env
         printf "\n\nSystem environment:\n\n"
         env
-        # Calculate the short SHA1 hash of the git commit
         echo "::set-output name=closest_tag::$(git describe --abbrev=0)"
-
-    # So GoReleaser can generate the changelog properly
-    - name: Unshallowify the repo clone
-      run: git fetch --prune --unshallow
 
     # GoReleaser will take care of publishing those artifacts into the release
     - name: Run GoReleaser
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,6 +43,7 @@ jobs:
 
     - name: Prep GPG env
       run: |
+        export GPG_TTY="$(tty)"
         mkdir -p ~/.gnupg
         echo use-agent >> ~/.gnupg/gpg.conf
         echo pinentry-mode loopback >> ~/.gnupg/gpg.conf
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,6 @@ jobs:
       with:
         version: latest
         args: release --rm-dist
-        key: ${{ secrets.SIGNING_KEY }}
       env:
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -46,6 +48,10 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    # So GoReleaser can generate the changelog properly
+    - name: Unshallow
```
francislavoie: Could use a better name here, at a glance it's unclear what this does when looking at the CI output

francislavoie: Also maybe only run this if it's a tagged run?

.github/workflows/ci.yml
```diff
@@ -66,7 +72,25 @@ jobs:
...
+        version: latest
+        args: release --rm-dist --skip-publish
+        # Figure out signing
+        # key: ${{ secrets.YOUR_PRIVATE_KEY }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        TAG: ${{ steps.vars.outputs.closest_tag }} #${{ github.ref }}
+
+    - name: Show project tree
+      run: tree
```
francislavoie: Need `sudo apt install tree`, also `if: matrix.os == 'ubuntu-latest'`

Mohammed90: Will only run it on Ubuntu, so I'm splitting the release process into its own yaml.

francislavoie:
This looks awesome 🎉

Mohammed90:
While it's mostly ready, don't merge it yet. I need to remove ` --skip-publish` and we need to figure out the signing of binaries.

.goreleaser.yml
```diff
@@ -0,0 +1,68 @@
...
+  homepage: https://caddyserver.com
+  description: Caddy Web Server
+  license: Apache 2.0
+  bindir: /usr/bin
+  formats:
+  - deb
+  - rpm
+  empty_folders:
+  - /etc/caddy
+  files:
```
Mohammed90: The `pre-install` and `post-install` scripts for the `deb` and `rpm` also need to be decided. I'm thinking we need to break the `packer/scripts/20-caddy.sh` file from the dist repo into 2 scripts: one to create the user and prep the directories, and the other to download the Caddy binary. Reuse the former script as one of the scripts to include as part of the packaging.

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
```
mholt: Do you know if this includes `-beta` releases (and similar)?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+  pull_request: # temporarily
+    branches: 
+      - v2
```
mholt: Probably changing branches approximately tomorrow, just FYI

.goreleaser.yml
mholt: I like reusing code :+1:

.goreleaser.yml
```diff
@@ -0,0 +1,68 @@
...
+    linux: Linux
+    windows: Windows
+    386: i386
+    amd64: x86_64
+checksum:
+  algorithm: sha256
+# signs:
+# - artifacts: all
+#   cmd: signify
+#   args: ["-S", "-s", "../seckey", "-m", "${artifact}"]
```
mholt: Do I need to set a private key somewhere?

.goreleaser.yml
```diff
@@ -0,0 +1,68 @@
...
+  - linux
+  - windows
+  - freebsd
+  flags:
+  - -trimpath
+  ldflags:
+  - -s -w
+archives:
+- replacements:
+    darwin: Darwin
```
mholt: Can we call it "Mac"? I don't know how many non-Go users know what Darwin is.

.goreleaser.yml
francislavoie: https://github.com/caddyserver/caddy/pull/3173/files#diff-56c30d9224897b08c28075052d4595c5R51

Yeah in the repo secrets like we did for fuzzit.

https://github.com/goreleaser/goreleaser-action#signing

.goreleaser.yml
francislavoie: It even has "win" in the name 😂

.goreleaser.yml
mholt: K, thanks. I've set the signing key as a secret named `SIGNING_KEY`.

.github/workflows/release.yml
Mohammed90: I think it should. It says the glob should match one or more characters [in the docs](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet), without any disqualifications. I debated between this pattern and mere `'*'`, but thought more explicit is better.

.goreleaser.yml
Mohammed90: It seems like managing {pre,post}{inst,rm} scripts is a [very tedious](https://wiki.debian.org/MaintainerScripts) [matter](https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html). It doesn't seem to be as simple as packaging our scripts as-is. Let's put off producing the rpm/deb packages for now until we learn this better. Is that fine?

For posterity: https://www.debian.org/doc/manuals/maint-guide/index.en.html

mholt: Yuck. That's a pain. Alright.

.goreleaser.yml
Mohammed90: Is it a `signify` secret key or `gpg` or something else? I put `singinfy` just for testing, but [it can be any command](https://goreleaser.com/customization/#Signing).

Here comes the paradox of choice... we need to decide on these.

Checksum algorithm:
- [ ] sha256
- [ ] sha512
- [ ] sha1
- [ ] crc32
- [ ] md5
- [ ] sha224
- [ ] sha384

Signing command:
- [ ] signify
- [ ] gpg
- [ ] gpg2
- [ ] minisign

What to sign:
- [ ] checksums
- [ ] all (executables, archives, checksums)

mholt: It's a GPG key.

Probably sign everything with sha256 or sha512 using gpg or gpg2. How's that sound?

.goreleaser.yml
```diff
@@ -0,0 +1,71 @@
...
+  - amd64
+  - 386
+  - arm
+  - arm64
+  goarm:
+  - 6
+  - 7
+  ignore:
+    - goos: darwin
+      goarch: 386
```
hairyhenderson: Probably also worth ignoring darwin-arm6/7. It's probably also not worth releasing a darwin-arm64 binary either - does anyone actually want a Caddy binary for iOS?

```suggestion
    - goos: darwin
      goarch: 386
    - goos: darwin
      goarch: arm
```

.goreleaser.yml
```diff
@@ -0,0 +1,71 @@
...
+  flags:
+  - -trimpath
+  ldflags:
+  - -s -w
+archives:
+  -
+    format_overrides:
```
hairyhenderson: nit: awkward yaml

```suggestion
  - format_overrides:
```

.goreleaser.yml
```diff
@@ -0,0 +1,71 @@
...
+      - goos: windows
+        format: zip
+    replacements:
+      darwin: Mac
```
hairyhenderson: The proper name is `macOS`, if that's what we're going for here... Though is there a reason these need to be capitalized?

```suggestion
      darwin: macOS
```

.goreleaser.yml
```diff
@@ -0,0 +1,71 @@
...
+      linux: Linux
+      windows: Windows
+      386: i386
+      amd64: x86_64
+checksum:
+  algorithm: sha512
```
hairyhenderson: Any specific reason you're choosing `sha512` here? `sha256` should be enough for this purpose... I'm already using `sha256` in the [`Dockerfile`s](https://github.com/caddyserver/caddy-docker/blob/d6ce5d55b4dbff40f0452f5478fdd0d648c17a15/alpine/Dockerfile#L20), though I can certainly switch to `sha512` if need be.

.goreleaser.yml
francislavoie: GoReleaser has it that way in their docs - but I agree it's weird

.goreleaser.yml
mholt: 256 is probably fine, we should match the Dockerfiles.

Mohammed90: I researched this and [found reports](https://crypto.stackexchange.com/questions/26336/sha512-faster-than-sha256) that sha512 is faster than sha256. I was able to reproduce the performance locally. So `sha512` seemed like future-proof collision resistance for free, with rebate too.

This is my local run:
```
Doing sha256 for 3s on 16 size blocks: 11893333 sha256's in 2.93s
Doing sha256 for 3s on 64 size blocks: 6140405 sha256's in 2.90s
Doing sha256 for 3s on 256 size blocks: 2638625 sha256's in 2.93s
Doing sha256 for 3s on 1024 size blocks: 749622 sha256's in 2.86s
Doing sha256 for 3s on 8192 size blocks: 109542 sha256's in 2.96s
Doing sha512 for 3s on 16 size blocks: 9164488 sha512's in 2.91s
Doing sha512 for 3s on 64 size blocks: 9295270 sha512's in 2.96s
Doing sha512 for 3s on 256 size blocks: 3468409 sha512's in 2.97s
Doing sha512 for 3s on 1024 size blocks: 1182612 sha512's in 2.97s
Doing sha512 for 3s on 8192 size blocks: 168891 sha512's in 2.97s
LibreSSL 2.6.5
built on: date not available
options:bn(64,64) rc4(16x,int) des(idx,cisc,16,int) aes(partial) blowfish(idx)
compiler: information not available
The 'numbers' are in 1000s of bytes per second processed.
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes
sha256           64946.53k   135512.39k   230541.98k   268396.13k   303164.89k
sha512           50388.94k   200978.81k   298960.51k   407742.32k   465843.46k
```

hairyhenderson: @Mohammed90 ok, I can switch to `sha512` in the `Dockerfile`!

hairyhenderson:
@Mohammed90 looks like this needs to be rebased to `master` (can that even be done without resubmitting the PR?)

.goreleaser.yml
mholt: Wow. Neat!

francislavoie:
Did a bit of reading in the GitHub help, and yes it seems possible.

Just rebase this onto master instead (rebase ignores commits already in the target branch so it should work just fine, but check git log before pushing) then you can click the edit button in the PR to change to target to master. I think.

Mohammed90:
Yep, as Francis says, it's possible. It's giving me this warning:

> Some commits from the old base branch may be removed from the timeline, and old review comments may become outdated.

The commits seem to be correct as well:
https://github.com/caddyserver/caddy/compare/master...add-goreleaser

Will the review comments be removed, or just marked outdated?

mholt:
Fwiw, I'm okay with merging this into v2 and then merging v2 into master.

francislavoie:
I think review comments are tied the diff, they should stick around regardless pretty sure. But I'm not 100% sure.

hairyhenderson:
> Fwiw, I'm okay with merging this into v2 and then merging v2 into master.

That's probably the easier option 😂

.github/workflows/release.yml
mholt: Should this be updated now? @Mohammed90

.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+        printf "Go version: $(go version)\n"
+        printf "\n\nGo environment:\n\n"
+        go env
+        printf "\n\nSystem environment:\n\n"
+        env
+    
+    # https://github.community/t5/GitHub-Actions/How-to-get-just-the-tag-name/m-p/32167/highlight/true#M1027
+    - name: Get the version
+      id: get_version
+      run: echo ::set-output name=version_tag::${GITHUB_REF/refs\/tags\//}
```
francislavoie: Doesn't actually matter cause we're only running this on ubuntu, but quoting this for safety is ideal (breaks on windows)

```suggestion
      run: echo "::set-output name=version_tag::${GITHUB_REF/refs\/tags\//}"
```

mholt:
Thanks for your hard work on this, @Mohammed90 !

In time, we can add signing and debian/fedora packages.

PR #3174, opened 2020-03-22
caddyauth: Add Metadata field to caddyauth.User

roblabla:
Long overdue PR to allow Authentication modules to add arbitrary data to a User.

### DIFF
--- a/modules/caddyhttp/caddyauth/caddyauth.go
+++ b/modules/caddyhttp/caddyauth/caddyauth.go
@@ -78,6 +78,9 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 	repl.Set("http.authentication.user.id", user.ID)
+	for k, v := range user.Metadata {
+		repl.Set("http.authentication.user.metadata." + k, v)
+	}
 
 	return next.ServeHTTP(w, r)
 }
@@ -93,6 +96,7 @@ type Authenticator interface {
 // User represents an authenticated user.
 type User struct {
 	ID string
+	Metadata map[string]string
 }
 
 // Interface guards
--- a/modules/caddyhttp/caddyauth/caddyauth.go
+++ b/modules/caddyhttp/caddyauth/caddyauth.go
@@ -77,9 +77,9 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 	}
 
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
-	repl.Set("http.authentication.user.id", user.ID)
+	repl.Set("http.auth.user.id", user.ID)
 	for k, v := range user.Metadata {
-		repl.Set("http.authentication.user.metadata."+k, v)
+		repl.Set("http.auth.user."+k, v)
 	}
 
 	return next.ServeHTTP(w, r)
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/caddyauth/caddyauth.go
```diff
@@ -78,6 +78,9 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 	repl.Set("http.authentication.user.id", user.ID)
+	for k, v := range user.Metadata {
+		repl.Set("http.authentication.user.metadata." + k, v)
```
mholt: > modules/caddyhttp/caddyauth/caddyauth.go:82: File is not `goimports`-ed (goimports)

Go home reviewdog, you're drunk

Mohammed90: I was wondering what it's complaining about. This is the `goimports` diff 😄 

```diff
--- caddyauth.go.orig   2020-03-23 01:23:26.000000000 +0300
+++ caddyauth.go        2020-03-23 01:23:26.000000000 +0300
@@ -79,7 +79,7 @@
        repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
        repl.Set("http.authentication.user.id", user.ID)
        for k, v := range user.Metadata {
-               repl.Set("http.authentication.user.metadata." + k, v)
+               repl.Set("http.authentication.user.metadata."+k, v)
        }
 
        return next.ServeHTTP(w, r)
@@ -95,7 +95,7 @@
 
 // User represents an authenticated user.
 type User struct {
-       ID string
+       ID       string
        Metadata map[string]string
 }
```

mholt: So... goimports is doing fmt now?

Mohammed90: 🤷‍♂ I thought it only sorted imports

mholt: Haha me too.

Okay, well this PR looks good to me, except for the formatting omissions.

francislavoie: https://godoc.org/golang.org/x/tools/cmd/goimports

> In addition to fixing imports, goimports also formats your code in the same style as gofmt so it can be used as a replacement for your editor's gofmt-on-save hook. 

Apparently yes

modules/caddyhttp/caddyauth/caddyauth.go
```diff
@@ -78,6 +78,9 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 	repl.Set("http.authentication.user.id", user.ID)
+	for k, v := range user.Metadata {
+		repl.Set("http.authentication.user.metadata."+k, v)
```
mholt:
```suggestion
		repl.Set("http.authentication.user."+k, v)
```

@roblabla Would this be OK with you? I can't think of a reason we need to separate out the metadata fields from anything else really.

mholt: Part of me wonders if we even need the `.authentication` namespace at all, wouldn't it be obvious that a user is from some kind of authentication? i.e. we could do just `http.user.*` - what do you think?

PR #3176, opened 2020-03-23
Support placeholders in key_type

pascalgn:
Support the following configuration:

```json
{
  "apps": {
    "tls": {
      "automation": {
        "policies": [
          {
            "key_type": "{env.KEY_TYPE}"
          }
        ]
      }
    }
  }
}

```

Also adds a check to validate key_type is valid (rsa2048, rsa4096, etc.)

### DIFF
--- a/modules/caddytls/automation.go
+++ b/modules/caddytls/automation.go
@@ -183,8 +183,14 @@ func (ap *AutomationPolicy) Provision(tlsApp *TLS) error {
 		ap.Issuer = val.(certmagic.Issuer)
 	}
 
+	keyType := ap.KeyType
+	if keyType != "" {
+		repl := caddy.NewReplacer()
+		keyType = repl.ReplaceAll(ap.KeyType, "")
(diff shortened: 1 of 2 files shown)


modules/caddytls/automation.go
```diff
@@ -183,8 +183,14 @@ func (ap *AutomationPolicy) Provision(tlsApp *TLS) error {
...
 
+	keyType := ap.KeyType
+	if keyType != "" {
+		repl := caddy.NewReplacer()
+		keyType = repl.ReplaceAll(ap.KeyType, "")
+	}
+
 	keySource := certmagic.StandardKeyGenerator{
-		KeyType: supportedCertKeyTypes[ap.KeyType],
+		KeyType: supportedCertKeyTypes[keyType],
```
mholt: Since you're setting the key type here, would be best to do the validation at this point, too (probably no need to use the Validate() method in your case)

PR #3180, opened 2020-03-23
reverse_proxy: Add support for SRV backends

### LINKED ISSUE #3179: v2: Support SRV DNS discovery for proxy upstreams
This is to replicate the same feature existing in `v1` here: https://caddyserver.com/v1/docs/proxy

mholt:
This change allows you to specify the `"lookup_srv"` field on a proxy upstream to lookup a name's SRV records to get the upstream dial address. Resolves #3179.

In the Caddyfile, prefix proxy upstream addresses with `srv+`, for example:

```
reverse_proxy srv+example.com
```

or

```
reverse_proxy srv+https://example.com
```

(It's a little weird but that's how it goes I guess.)

@danlsgiga Can you please try this? Once the CI is finished, it should have a binary you can download and use, or you can just pull it yourself and build. Let me know!

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -41,7 +41,7 @@ type (
 	// especially A/AAAA pointed at your server.
 	//
 	// Automatic HTTPS can be
-	// [customized or disabled](/docs/json/apps/http/servers/automatic_https/).
+	// [customized or disabled](/docs/modules/http#servers/automatic_https).
 	MatchHost []string
 
 	// MatchPath matches requests by the URI's path (case-insensitive). Path
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -177,13 +177,36 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 		return net.JoinHostPort(host, port), nil
 	}
 
+	// appendUpstream creates an upstream for address and adds
+	// it to the list. If the address starts with "srv+" it is
+	// treated as a SRV-based upstream, and any port will be
+	// dropped.
+	appendUpstream := func(address string) error {
+		isSRV := strings.HasPrefix(address, "srv+")
+		if isSRV {
+			address = strings.TrimPrefix(address, "srv+")
+		}
+		dialAddr, err := upstreamDialAddress(address)
+		if err != nil {
+			return err
+		}
+		if isSRV {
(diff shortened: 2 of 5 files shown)


francislavoie:
Why not `srv/` like `unix/`?

mholt:
> Why not `srv/` like `unix/`?

Because it's not a network type; you could in theory have TCP or UDP protocol for SRV lookups, see here: https://golang.org/pkg/net/#Resolver.LookupSRV

danlsgiga:
@mholt, the SRV functionality works as expected. I can provide only the address without ports now and it can proxy requests successfully.

I see you added the `https` option for the `Caddyfile`, but how can I tell via `json` (using `"lookup_srv"`) for the upstream to use https/tls and potentially `insecure_skip_verify`?

mholt:
@danlsgiga Yay! That's awesome. Thanks for verifying!

If you use a Caddyfile with:

```
reverse_proxy srv+https://example.com {
    insecure_skip_verify
}
```

in it, you can get the resulting JSON with:

```
$ caddy adapt
```

and then you will see how TLS is enabled with the insecure option :persevere:

danlsgiga:
Yeah, I know... sometimes we don't have an option but to not verify upstreams certs (self signed).

So, I did what you asked and:

```
adapt: parsing caddyfile tokens for 'reverse_proxy': Caddyfile:4 - Error during parsing: unrecognized subdirective insecure_skip_verify
```

That might not be supported yet :/

mholt:
@danlsgiga I edited my comment a minute after I posted it :) The correct syntax is documented here: https://caddyserver.com/docs/caddyfile/directives/reverse_proxy

danlsgiga:
Ha! it works!!!!!
It just needed the `transport` block. Awesome!!!!

```
                     "handle": [
                        {
                          "handler": "reverse_proxy",
                          "transport": {
                            "protocol": "http",
                            "tls": {
                              "insecure_skip_verify": true,
                              "server_name": "myapp.service.consul"
                            }
                          },
                          "upstreams": [
                            {
                              "lookup_srv": "app.service.consul"
                            }
                          ]
                        }
                      ]
```

One gotcha though: the json `server_name` is not documented in https://caddyserver.com/docs/caddyfile/directives/reverse_proxy

mholt:
@danlsgiga That's a JSON option, not a Caddyfile option (the Caddyfile adapter sets that implicitly so that SNI works with the upstream.) https://caddyserver.com/docs/json/apps/http/servers/routes/handle/reverse_proxy/transport/http/tls/

danlsgiga:
np... works fine for me! Thanks a lot!

PR #3193, opened 2020-03-25
Fixes SNI being applied to http servers

sarge:
Here are some tests were we can agree on what the output should be. Testing against json is a bit of an eye sore, ~I was thinking of adding in a json diff tool to helpout.~ I have added a diffing tool

In the `TestHttpsMultiHostWithAcme` I removed the TLS Connection Policy from the `:80` server.

~I was unsure about the need for the skip https on servers mapped to `:80` as well.~ This was already fixed, I was looking at an earlier beta.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -367,7 +367,7 @@ func (st *ServerType) serversFromPairings(
 			return specificity(iLongestHost) > specificity(jLongestHost)
 		})
 
-		var hasCatchAllTLSConnPolicy bool
+		var requiresCatchAllTLSConnPolicy bool
 
 		// create a subroute for each site in the server block
 		for _, sblock := range p.serverBlocks {
@@ -376,45 +376,65 @@ func (st *ServerType) serversFromPairings(
 				return nil, fmt.Errorf("server block %v: compiling matcher sets: %v", sblock.block.Keys, err)
 			}
 
-			hosts := sblock.hostsFromKeys(false, false)
+			var httpHosts, httpsHosts []string
+			for _, key := range sblock.block.Keys {
+				addr, err := ParseAddress(key)
+				if err != nil {
+					return nil, err
+				}
+				addr = addr.Normalize()
 
-			// tls: connection policies
-			if cpVals, ok := sblock.pile["tls.connection_policy"]; ok {
-				// tls connection policies
-				for _, cpVal := range cpVals {
-					cp := cpVal.Value.(*caddytls.ConnectionPolicy)
+				if addr.Scheme == "http" {
+					httpHosts = append(httpHosts, addr.Host)
+				} else {
+					httpsHosts = append(httpsHosts, addr.Host)
+				}
 
-					// make sure the policy covers all hostnames from the block
-					for _, h := range hosts {
-						if h == defaultSNI {
-							hosts = append(hosts, "")
-							cp.DefaultSNI = defaultSNI
-							break
+				// exclude any hosts that were defined explicitly with
+				// "http://" in the key from automated cert management (issue #2998)
+				if addr.Scheme == "http" {
+					if addr.Host != "" {
+						if srv.AutoHTTPS == nil {
+							srv.AutoHTTPS = new(caddyhttp.AutoHTTPSConfig)
+						}
+						if !sliceContains(srv.AutoHTTPS.Skip, addr.Host) {
+							srv.AutoHTTPS.Skip = append(srv.AutoHTTPS.Skip, addr.Host)
+							sort.Strings(srv.AutoHTTPS.Skip)
 						}
 					}
-
-					if len(hosts) > 0 {
-						cp.MatchersRaw = caddy.ModuleMap{
-							"sni": caddyconfig.JSON(hosts, warnings), // make sure to match all hosts, not just auto-HTTPS-qualified ones
-						}
-					} else {
-						cp.DefaultSNI = defaultSNI
-						hasCatchAllTLSConnPolicy = true
-					}
-
-					srv.TLSConnPolicies = append(srv.TLSConnPolicies, cp)
 				}
 			}
 
-			// exclude any hosts that were defined explicitly with
-			// "http://" in the key from automated cert management (issue #2998)
-			for _, addr := range sblock.keys {
-				if addr.Scheme == "http" && addr.Host != "" {
-					if srv.AutoHTTPS == nil {
-						srv.AutoHTTPS = new(caddyhttp.AutoHTTPSConfig)
-					}
-					if !sliceContains(srv.AutoHTTPS.Skip, addr.Host) {
-						srv.AutoHTTPS.Skip = append(srv.AutoHTTPS.Skip, addr.Host)
+			// sort hosts for a predictable config file
+			sort.Strings(httpHosts)
+			sort.Strings(httpsHosts)
+
+			// tls: connection policies
+			if len(httpsHosts) > 0 {
+				if cpVals, ok := sblock.pile["tls.connection_policy"]; ok {
+					// tls connection policies
+					for _, cpVal := range cpVals {
+						cp := cpVal.Value.(*caddytls.ConnectionPolicy)
+						// make sure the policy covers all hostnames from the block
+						for _, h := range httpsHosts {
+							if h == defaultSNI {
+								httpHosts = append(httpHosts, "")
+								cp.DefaultSNI = defaultSNI
+								requiresCatchAllTLSConnPolicy = true
+								break
+							}
+						}
+
+						if len(httpsHosts) > 0 {
+							cp.MatchersRaw = caddy.ModuleMap{
+								"sni": caddyconfig.JSON(httpsHosts, warnings), // make sure to match all hosts, not just auto-HTTPS-qualified ones
+							}
+						} else {
+							cp.DefaultSNI = defaultSNI
+							requiresCatchAllTLSConnPolicy = true
+						}
+
+						srv.TLSConnPolicies = append(srv.TLSConnPolicies, cp)
 					}
 				}
 			}
@@ -475,7 +495,7 @@ func (st *ServerType) serversFromPairings(
 		// TODO: maybe a smarter way to handle this might be to just make the
 		// auto-HTTPS logic at provision-time detect if there is any connection
 		// policy missing for any HTTPS-enabled hosts, if so, add it... maybe?
-		if !hasCatchAllTLSConnPolicy && (len(srv.TLSConnPolicies) > 0 || defaultSNI != "") {
+		if requiresCatchAllTLSConnPolicy && (len(srv.TLSConnPolicies) > 0 || defaultSNI != "") {
 			srv.TLSConnPolicies = append(srv.TLSConnPolicies, &caddytls.ConnectionPolicy{DefaultSNI: defaultSNI})
 		}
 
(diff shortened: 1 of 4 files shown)


sarge:
@mholt I realise this is a bit bold to be in here rummaging about, without all scenarios well understood. I do want to add some more use cases as I understand them better.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -410,53 +410,66 @@ func (st *ServerType) serversFromPairings(
...
 				addr, err := ParseAddress(key)
 				if err != nil {
 					return nil, err
 				}
 				addr = addr.Normalize()
-				if addr.Scheme == "http" && addr.Host != "" {
-					if srv.AutoHTTPS == nil {
-						srv.AutoHTTPS = new(caddyhttp.AutoHTTPSConfig)
+
+				//TODO: do we need to do something about this default port??
```
sarge: I am trying to determine if this pass is for an http or https server, so probably need `app.httpPort()`

mholt:
There's a lot of non-test changes here sarge, what do these all do?

sarge:
@mholt - you know when you start pulling a thread on a jumper and it starts unravelling?

So initially I wanted to exclude sni being set for http only servers, it was forcing a TLSConnectionPolicy to be set unnecessarily. 

I appreciate that this is a complex area and started this PR to capture the current behaviour, and fix the bug.

https://github.com/caddyserver/caddy/blob/cdafeee45fd0ce351f55c75522fe092711b89532/caddytest/integration/caddyfile_adapt_test.go#L227-L234

In testing that I discovered that in the case of a mixed http/https mode web server the hostnames were getting applied to both incorrectly (I believe)

https://github.com/caddyserver/caddy/blob/cdafeee45fd0ce351f55c75522fe092711b89532/caddytest/integration/caddyfile_adapt_test.go#L336

Now it groups them 
https://github.com/caddyserver/caddy/blob/cdafeee45fd0ce351f55c75522fe092711b89532/caddytest/integration/caddyfile_adapt_test.go#L356-L359

and
https://github.com/caddyserver/caddy/blob/cdafeee45fd0ce351f55c75522fe092711b89532/caddytest/integration/caddyfile_adapt_test.go#L411-L413

My feeling is that we need a barrage of tests to avoid regressions, plus I think it would be a useful approach to share what we think should happen. I have placed some sorting logic to keep the output consistent. Would love to have some more test scenarios, at this stage the tests are far more valuable than the code I have written to make them pass.

Happy to have a chat about the approach

mholt:
Considering the non-test changes here, I think it is best to not merge this in before 2.0 unless it is a critical bug fix.

caddyconfig/httpcaddyfile/httptype.go
mholt: Yeah, using the configured HTTP port is always better than hard-coding a port.

caddyconfig/httpcaddyfile/tlsapp.go
```diff
@@ -382,5 +382,9 @@ func consolidateAutomationPolicies(aps []*caddytls.AutomationPolicy) []*caddytls
 		return len(aps[i].Subjects) > len(aps[j].Subjects)
 	})
 
+	for i := 0; i < len(aps); i++ {
+		sort.Strings(aps[i].Subjects)
```
mholt, line 386: Why is this needed?

caddytest/caddytest.go
```diff
@@ -238,6 +240,55 @@ func createTestingTransport() *http.Transport {
 	}
 }
 
+// AssertAdapt -
```
mholt, line 243: Finish godoc

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -410,53 +410,66 @@ func (st *ServerType) serversFromPairings(
...
-					} else {
-						cp.DefaultSNI = defaultSNI
-						hasCatchAllTLSConnPolicy = true
-					}
-
-					srv.TLSConnPolicies = append(srv.TLSConnPolicies, cp)
-				}
-			}
+			httpHosts := make([]string, 0)
+			httpsHosts := make([]string, 0)
```
mholt, line 414:
```suggestion
			var httpHosts, httpsHosts []string
```

caddytest/caddytest.go
```diff
@@ -19,6 +19,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aryann/difflib"
```
mholt, line 22: How necessary is this dependency?

caddyconfig/httpcaddyfile/tlsapp.go
sarge: So the diffing approach treats the out puts as text, there are few places that iterate over maps and things which will randomly reorder the outputs

caddytest/caddytest.go
sarge: It makes understanding what is broken significantly easier. I'm not sold on this particular tool, could use a json diffing tool which should not require the sorting code.

caddyconfig/httpcaddyfile/tlsapp.go
mholt: Cool, sounds good to me.

caddytest/caddytest.go
mholt: Should/could the sorting be done in the tests instead of in the actual code? Not saying it's bad in the code, just wondering since it's extra work just to cater to the tests.

sarge: I would view it as reducing the friction to writing tests. One reason to exclude it would be because it could end up in the binary. Not sure if the compiler understands that this for tests only.
The alternative would be to include a function in a test file that does the diffing instead. Not sure if that would be materially different.

caddyconfig/httpcaddyfile/httptype.go
sarge: From my tests it appears that only the scheme is needed

caddytest/caddytest.go
sarge: Done in test harness PR

caddyconfig/httpcaddyfile/httptype.go
sarge: Done

caddytest/caddytest.go
```diff
@@ -369,5 +375,53 @@ func AssertAdapt(t *testing.T, rawConfig string, adapterName string, expectedRes
...
 }
+
+type load struct {
+	Apps apps `json:"apps"`
+}
+type apps struct {
+	HTTP hTTP        `json:"http"`
+	TLS  interface{} `json:"tls,omitempty"`
+}
+type hTTP struct {
```
mholt, line 390: This casing is weird. I know why you did it like this, but maybe just rename it instead, like `httpServers` or something.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -376,45 +376,65 @@ func (st *ServerType) serversFromPairings(
 				return nil, fmt.Errorf("server block %v: compiling matcher sets: %v", sblock.block.Keys, err)
 			}
 
-			hosts := sblock.hostsFromKeys(false, false)
+			var httpHosts, httpsHosts []string
+			for _, key := range sblock.block.Keys {
+				addr, err := ParseAddress(key)
```
mholt, line 381: Keys are already parsed in the latest master; iterate sblock.keys instead.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -376,45 +376,65 @@ func (st *ServerType) serversFromPairings(
...
-
-					// make sure the policy covers all hostnames from the block
-					for _, h := range hosts {
-						if h == defaultSNI {
-							hosts = append(hosts, "")
-							cp.DefaultSNI = defaultSNI
-							break
-						}
-					}
+				if addr.Scheme == "http" {
```
mholt, line 387: No consideration for port? Semantically, http:// and port 80 are equivalent, as are https:// and :443.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -376,45 +376,65 @@ func (st *ServerType) serversFromPairings(
...
-						srv.AutoHTTPS = new(caddyhttp.AutoHTTPSConfig)
-					}
-					if !sliceContains(srv.AutoHTTPS.Skip, addr.Host) {
-						srv.AutoHTTPS.Skip = append(srv.AutoHTTPS.Skip, addr.Host)
+			// sort hosts for a predictable config file
+			sort.Strings(httpHosts)
+			sort.Strings(httpsHosts)
+
+			// tls: connection policies
+			if len(httpsHosts) > 0 {
```
mholt, line 413: Keys that don't explicitly provide the `https://` scheme can still add TLS connection policies.

caddyconfig/httpcaddyfile/httptype.go
sarge: Nice thanks!

caddyconfig/httpcaddyfile/httptype.go
sarge: I don't think we need to be concerned with the port.So I tested with server hostnames like `:80` `http://:80`

caddyconfig/httpcaddyfile/httptype.go
sarge: Good point, how about a different name like `tlsHosts`? With the assumption that caddy is all about the TLS, the scheme check above looks for `http` and assumes tls for everything else.

caddytest/caddytest.go
sarge: Yeah it felt iffy to me too:)

mholt:
@sarge I found a much, much simpler way to fix this on the [`default-sni-fix`](https://github.com/caddyserver/caddy/tree/default-sni-fix) branch. Can you try it out?

sarge:
@mholt That works needs a test though.

https://github.com/caddyserver/caddy/commit/61266df53b8c32aa14b36a6054dc0e4b2117e41a#diff-bcfd3742388849a127292f52b12a1288R481-R483
Isn't much fun to parse :) What could be done to tidy that up?


Is this a valid thing to ask, I see there is now an error message "adapting config using caddyfile adapter: server listening on [:80] is HTTP, but attempts to configure TLS connection policies"
```
       a.caddy.localhost, b.caddy.localhost, http://c.caddy.localhost {
		tls admin@example.com
		respond /version 200 {
			body "hello from localhost"
		}
	}
```


Test for you :)
```
func TestHttpOnlyOnDomainWithSNI(t *testing.T) {
	caddytest.AssertAdapt(t, `
	{
		default_sni a.caddy.localhost
	}
	:80 {
		respond /version 200 {
			body "hello from localhost"
		}
	}
  `, "caddyfile", `{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":80"
					],
					"routes": [
						{
							"match": [
								{
									"path": [
										"/version"
									]
								}
							],
							"handle": [
								{
									"body": "hello from localhost",
									"handler": "static_response",
									"status_code": 200
								}
							]
						}
					]
				}
			}
		}
	}
}`)
```

mholt:
> 61266df#diff-bcfd3742388849a127292f52b12a1288R481-R483
Isn't much fun to parse :) What could be done to tidy that up?

"If it uses TLS, does not have a catch-all TLS policy, and either has a TLS conn policy or has default SNI, then add a catch-all TLS conn policy."

It's pretty tidy, there are just a lot of possibilities and edge cases. I'm not sure how to make it any simpler.

Thanks for the test. Want to open a PR against that branch? (Then you'll get credited in the history.)

PR #3194, opened 2020-03-26
ci: fuzz: switch engine from libfuzzer to native go-fuzz

Mohammed90:
Fewer deps & use only native Go fuzzing engine. Should make our life a bit easier.


Mohammed90:
`go-fuzz currently only supports fuzzing`, meaning it doesn't support running regression of fuzzing crashers on PR. @mholt & @francislavoie , thoughts?

francislavoie:
I don't have much of an opinion on this, I trust your judgement 😁

PR #3207, opened 2020-03-31
caddyfile: block imports of files within sensitive dirs

Mohammed90:
This PR addresses a crasher found by Fuzzit: `import /proc/self/fd/1`


mholt:
As discussed in Slack, I think the solutions for working around the fuzzer to accomodate its unusual environment (i.e. `import *`) is harder than just telling the fuzzer to skip `import` inputs. The complexity for making these all work is a bit too high using known/proposed solutions, for problems that we aren't sure really exist outside of a fuzzing environment.

PR #3208, opened 2020-03-31
Allow specifying multiple matcher sets in "not" matcher

mholt:
A "not" matcher now accepts an array of matcher sets, rather than a single matcher set.

Matcher sets are OR'ed (i.e. if any matcher sets return true, this matcher returns false; if all matcher sets return false, this matcher returns true). This is analogous with how matcher sets work in routes.

Matchers within the same set are the same (i.e. they are AND'ed).

In the Caddyfile, if `not` appears multiple times in the same set, each occurrence creates a new matcher set within a single "not" matcher; in other words, multiple `not`s are effectively OR'ed because they create a single matcher (that is AND'ed with the others).

See https://caddy.community/t/v2-matcher-or-in-not/7355/

/cc @princemaple - please try this out and confirm that it works for you! Thanks!

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -99,13 +99,15 @@ type (
 		cidrs []*net.IPNet
 	}
 
-	// MatchNot matches requests by negating its matchers' results.
-	// To use, simply specify a set of matchers like you normally would;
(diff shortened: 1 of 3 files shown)


princemaple:
👍 thanks!

PR #3228, opened 2020-04-05
caddyhttp: Support single-line not matcher shortcut

francislavoie:
Simple addition to make it possible to have `not` negate a matcher on the same line instead of requiring it to be contained in a block.

E.g. before:

```
@notStatic {
    not {
        path /static*
    }
}
```

after:

```
@notStatic {
    not path /static*
}
```

easy

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -559,7 +559,7 @@ func (m *MatchNot) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	for d.Next() {
 		var mp matcherPair
 		matcherMap := make(map[string]RequestMatcher)
-		for d.NextBlock(0) {
+		for d.NextArg() || d.NextBlock(0) {
 			matcherName := d.Val()
 			mod, err := caddy.GetModule("http.matchers." + matcherName)
 			if err != nil {
--- a/caddyconfig/httpcaddyfile/httptype_test.go
+++ b/caddyconfig/httpcaddyfile/httptype_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 )
 
-func TestServerType(t *testing.T) {
+func TestMatcherSyntax(t *testing.T) {
 	for i, tc := range []struct {
 		input       string
 		expectWarn  bool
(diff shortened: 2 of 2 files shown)


sarge:
Nice, I wonder what a good testing approach is for this?

Perhaps a usecase in here

https://github.com/caddyserver/caddy/blob/84c729e96a32a52a41195fe3414fd0801e779f34/caddyconfig/httpcaddyfile/httptype_test.go#L9-L33

francislavoie:
I don't think `TestServerType` is the right place for testing matchers. We don't have any tests for Caddyfile matchers currently as far as I can tell, so I didn't bother making any - my experience in Go is extremely limited. I can read the code, but barely write anything 😛

sarge:
Understood, I think that test name is bit misleading because it is actually testing the parsing of a matcher.

francislavoie:
I'll rename `TestServerType` to `TestMatcherSyntax` and add some tests 👍

francislavoie:
Added a couple test cases, but I don't think there's a huge amount of value here because it didn't return a warning or error before with `not path /somepath*`, it just generated bad JSON.

```
"match": [
        {
                "not": [
                        {},
                        {},
                        {}
                ]
        }
],
```

modules/caddyhttp/matchers.go
```diff
@@ -559,7 +559,7 @@ func (m *MatchNot) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	for d.Next() {
 		var mp matcherPair
 		matcherMap := make(map[string]RequestMatcher)
-		for d.NextBlock(0) {
+		for d.NextArg() || d.NextBlock(0) {
```
mholt, line 562: This is clever, but I do admit I think it's kinda weird. We can go with it for now and fix it later if people complain.

francislavoie: What about it do you find weird?

mholt: It's just an unusual condition for a parsing loop.

francislavoie: Idk, reads perfectly fine to me. 🤷‍♂

"While there's still an arg or a block ahead, loop to read for additional matchers"

PR #3242, opened 2020-04-08
caddyfile: Support backticks as quotes

### LINKED ISSUE #2591: Caddyfile support backtick strings

Caddyfile support multi line string

without supporting of multi line string, it's difficult to develop some new plugins which need user to configure some complex string such as json.

```
directive here_might_be_a_json

# now
directive here_json_file_path
```

mholt:
Slated to land just after 2.0 release.

Closes #2591

### DIFF
--- a/caddyconfig/caddyfile/lexer.go
+++ b/caddyconfig/caddyfile/lexer.go
@@ -73,7 +73,7 @@ func (l *lexer) load(input io.Reader) error {
 // a token was loaded; false otherwise.
 func (l *lexer) next() bool {
 	var val []rune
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@mholt Why use backticks instead of `"""`? That would align similarly with other languages that do this (like Python, Ruby, _et al_)?

francislavoie:
Since the primary usecase for this is JSON strings, backticks make more sense here I think. Easier to read, and JS has support for backticks as "template literals", i.e. interpolated strings.

PR #3245, opened 2020-04-08
chore: make the linter happier

Mohammed90:
This PR enables 2 additional linters:
- [prealloc](https://github.com/alexkohler/prealloc), which finds slices that can be preallocated and subsequently [leads to better performance and fewer allocations](https://github.com/alexkohler/prealloc#purpose)
- [unconvert](https://github.com/mdempsky/unconvert), which removes unnecessary conversions

I've taken action on all of `unconvert` and most of `prealloc`.  [`maligned`](https://github.com/mdempsky/maligned) reports 4 structs whose size can be reduced:
- 64 B -> 56 B: AdminConfig, AutoHTTPSConfig, responseRecorder
- 152 B -> 144 B: AutomationPolicy

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,6 @@
 linters-settings:
+  maligned:
+    suggest-new: true
   errcheck:
     ignore: fmt:.*,io/ioutil:^Read.*,github.com/caddyserver/caddy/v2/caddyconfig:RegisterAdapter,github.com/caddyserver/caddy/v2:RegisterModule
     ignoretests: true
@@ -8,6 +10,8 @@ linters-settings:
 linters:
   enable:
     - bodyclose
+    - prealloc
+    - unconvert
     - errcheck
     - gofmt
     - goimports
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,4 @@
 linters-settings:
-  maligned:
-    suggest-new: true
   errcheck:
     ignore: fmt:.*,io/ioutil:^Read.*,github.com/caddyserver/caddy/v2/caddyconfig:RegisterAdapter,github.com/caddyserver/caddy/v2:RegisterModule
     ignoretests: true
--- a/caddyconfig/httpcaddyfile/addresses.go
+++ b/caddyconfig/httpcaddyfile/addresses.go
@@ -138,7 +138,7 @@ func (st *ServerType) mapAddressToServerBlocks(originalServerBlocks []serverBloc
 // association from multiple addresses to multiple server blocks; i.e. each element of
 // the returned slice) becomes a server definition in the output JSON.
 func (st *ServerType) consolidateAddrMappings(addrToServerBlocks map[string][]serverBlock) []sbAddrAssociation {
-	var sbaddrs []sbAddrAssociation
+	sbaddrs := make([]sbAddrAssociation, 0, len(addrToServerBlocks))
 	for addr, sblocks := range addrToServerBlocks {
 		// we start with knowing that at least this address
 		// maps to these server blocks
@@ -199,7 +199,7 @@ func (st *ServerType) listenerAddrsForServerBlockKey(sblock serverBlock, key str
 	}
 
 	// the bind directive specifies hosts, but is optional
-	var lnHosts []string
+	lnHosts := make([]string, 0, len(sblock.pile))
 	for _, cfgVal := range sblock.pile["bind"] {
 		lnHosts = append(lnHosts, cfgVal.Value.([]string)...)
 	}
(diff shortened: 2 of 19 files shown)


mholt:
Very cool!

I don't know much about the `maligned` linter, but I do know that in some cases, field order in a struct matters. For example, see the [Bugs section in the sync/atomic package](https://golang.org/pkg/sync/atomic/#pkg-note-BUG):

> On ARM, x86-32, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned. 

So, if we aren't careful, moving the wrong field could yield panics on some architectures. (I've encountered this in the past before.)

Mohammed90:
Good thing I didn't take action on its reports, I guess. I'll remove it to avoid CI lint failures and not risk someone taking action blindly.

mholt:
@Mohammed90 Actually the responseRecorder might be worth reordering, since they can be used in hot paths.

mholt:
Actually I hope that the `maligned` linter would be smart enough to know about that bug, it's a fairly well-known issue... if not, maybe we could request it as a feature?

Mohammed90:
maligned hasn't been updated since July of 2018. So not sure there be much update, unless mdempsky delegates it to another maintainer or picks up back on it. Maybe we could do local runs periodically? Not the best idea, but don't really like CI failures to show up that aren't really actionable.

.golangci.yml
```diff
@@ -1,4 +1,6 @@
 linters-settings:
+  maligned:
+    suggest-new: true
```
mholt, line 3: Are these still needed?

Mohammed90:
I've reordered the responseRecorder fields as per the suggestion of maligned

.golangci.yml
Mohammed90: Oops. No.

caddyconfig/httpcaddyfile/addresses.go
```diff
@@ -199,7 +199,7 @@ func (st *ServerType) listenerAddrsForServerBlockKey(sblock serverBlock, key str
 	}
 
 	// the bind directive specifies hosts, but is optional
-	var lnHosts []string
+	lnHosts := make([]string, 0, len(sblock.pile))
```
mholt, line 202: This is not the right length. I don't think we know the right length without an extra iteration. I guess that's OK though since the third parameter is just capacity, not length, and it will never be too large right?

modules/caddytls/tls.go
```diff
@@ -457,7 +457,7 @@ func (t *TLS) moveCertificates() error {
 	}
 
 	// get list of used CAs
-	var oldCANames []string
+	oldCANames := make([]string, 0, len(oldAcmeCas))
```
mholt, line 460: Similarly, I suppose it's OK if the capacity isn't exactly correct, right? Here, the realized length might be slightly under the capacity. But again this is not a hot path so it's probably fine :P

modules/caddyhttp/responsewriter.go
```diff
@@ -74,15 +74,26 @@ type HTTPInterfaces interface {
...
+struct of size 64 bytes could be of size 56 bytes:
+	ResponseWriterWrapper	*github.com/caddyserver/caddy/v2/modules/caddyhttp.ResponseWriterWrapper,
+	statusCode           	int,
+	buf                  	*bytes.Buffer,
+	shouldBuffer         	github.com/caddyserver/caddy/v2/modules/caddyhttp.ShouldBufferFunc,
+	size                 	int,
+	header               	net/http.Header,
+	wroteHeader          	bool,
+	stream               	bool,
+*/
```
mholt, line 87: I do not think this is necessary to keep in the code. The order of fields is seldom changed after the struct is created, especially on unexported types. We can add it to the commit message though!

caddyconfig/httpcaddyfile/addresses.go
Mohammed90: Yep. There a couple of them where the length is just the lower bound, others are upper bound, and some are exactly as expected. The point is to avoid constant reallocation by the runtime, and at times avoiding spendy allocation. So there is 1 of 3 scenarios where the capacity is lower than the real number, and we'll endure 1 or 2 more allocations.

modules/caddyhttp/responsewriter.go
Mohammed90: 🤦‍♂

caddyconfig/httpcaddyfile/addresses.go
mholt: Alrighty, sounds fine to me. Thanks!

modules/caddyhttp/responsewriter.go
Mohammed90: Removed

mholt:
Thanks for the contribution @Mohammed90 !

Issue #3212, opened 2020-04-01
v2: HTTP-only wildcard sites still redirect to HTTPS if another site is configured with automatic HTTPS


lukasbestle:
**Caddy version:** `v2.0.0-beta9.0.20200401064340-809e72792c50 h1:r0vSakzB/zmOVhKZINtlCNi1NfaagKG2NP7oAZUSjL4=`

**Caddyfile:**

```
tls.test {
   tls internal

   respond "HTTPS"
}

http://* {
  respond "HTTP"
}
```

**Expected behavior:** Requests to `http://tls.test` should be automatically redirected to `https://tls.test`, while requests to any other site that matches the second site block should be served via HTTP only and *not* redirect to HTTPS.

**Actual behavior**: Redirection occurs. If you comment out the first site block, it works again. So the issue only occurs if there is also a site listening on HTTPS.

Henrocker:
@lukasbestle Can you reproduce this behaviour with latest beta 20? Your version is beta 9.

lukasbestle:
@Henrocker No, it's a build of commit 809e72792c50, which is more recent than beta 20.

Henrocker:
@lukasbestle I'm a little bit confused because your Caddy version clearly states that you're on beta 9:
```
Caddy version: v2.0.0-beta9.0.20200401064340-809e72792c50 h1:r0vSakzB/zmOVhKZINtlCNi1NfaagKG2NP7oAZUSjL4=
```

lukasbestle:
I don't blame you, I was confused as well. But that seems to be because `go build`'s version detection is broken.

Take a look at the part behind the last dash: There's the commit hash `809e72792c50`. If you don't believe me, try building Caddy yourself from any recent commit hash of the `master` branch.

mholt:
That's a sorting bug in Go modules. I think it will get fixed in rc1 soon. But the commit follows the tag which is correct.

Henrocker:
@mholt Ahh okay, got it.

mholt:
@lukasbestle 

> http://* 

That matches only single-label domain names like `localhost` or `foobar`, etc. Similarly, `http://*.example.com` matches only `foo.example.com` (etc.) but not `foo.bar.example.com`.

To catch all hostnames, omit the `*`.

lukasbestle:
Ah, thanks!

I tried it with the following configuration:

```
tls.test {
   tls internal

   respond "HTTPS"
}

http:// {
  respond "HTTP"
[...]

PR #3246, opened 2020-04-08
caddyhttp: Add placeholder for domains that should have HTTPS redirects

francislavoie:
WIP implementation of https://github.com/caddyserver/caddy/issues/3212#issuecomment-611229153

This isn't working yet, when I try to compile the CEL expression, it fails with this message:

```
validate: loading http app module: provision http: server srv1: setting up route matchers: route 0: loading matcher modules: module name 'expression': provis
ion http.matchers.expression: CEL request matcher expects return type of bool, not dyn:<>
```

Using this Caddyfile as a testcase:

```
abc.com {
  respond "HTTPS"
}

bcd.com {
  respond "HTTPS"
}

cde.com {
  respond "HTTPS"
}

http:// {
  @shouldRedir {
    expression {host} in {http.autohttps.redir_domains}
  }
  redir @shouldRedir https://{host}{uri} 308

  respond "HTTP"
}
```

```
$ ./caddy validate --config Caddyfile
```

### DIFF
--- a/modules/caddyhttp/celmatcher.go
+++ b/modules/caddyhttp/celmatcher.go
@@ -112,8 +112,11 @@ func (m *MatchExpression) Provision(_ caddy.Context) error {
 
 	// request matching is a boolean operation, so we don't really know
 	// what to do if the expression returns a non-boolean type
-	if !proto.Equal(checked.ResultType(), decls.Bool) {
-		return fmt.Errorf("CEL request matcher expects return type of bool, not %s", checked.ResultType())
+	// TODO: for now, we're allowing Dyn because certain expressions can't have
+	// their type inferred if there's ambiguity.
+	// See https://github.com/caddyserver/caddy/pull/3246#issuecomment-611256387
+	if !proto.Equal(checked.ResultType(), decls.Bool) && !proto.Equal(checked.ResultType(), decls.Dyn) {
+		return fmt.Errorf("CEL request matcher expects return type of bool or dyn, not %s", checked.ResultType())
 	}
 
 	// compile the "program"
--- a/modules/caddyhttp/celmatcher.go
+++ b/modules/caddyhttp/celmatcher.go
@@ -112,11 +112,8 @@ func (m *MatchExpression) Provision(_ caddy.Context) error {
 
 	// request matching is a boolean operation, so we don't really know
 	// what to do if the expression returns a non-boolean type
-	// TODO: for now, we're allowing Dyn because certain expressions can't have
-	// their type inferred if there's ambiguity.
-	// See https://github.com/caddyserver/caddy/pull/3246#issuecomment-611256387
-	if !proto.Equal(checked.ResultType(), decls.Bool) && !proto.Equal(checked.ResultType(), decls.Dyn) {
-		return fmt.Errorf("CEL request matcher expects return type of bool or dyn, not %s", checked.ResultType())
+	if !proto.Equal(checked.ResultType(), decls.Bool) {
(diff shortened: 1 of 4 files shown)


francislavoie:
@TristonianJones maybe you could help here, I might not understand how to use CEL correctly - how would I do `string in list` here? Any thought why it would tell me the return type is `dyn`?

TristonianJones:
@francislavoie the `string in list` is declared to  return type `bool`, so there could be a different issue in the compilation. I'll see if I can debug it.

TristonianJones:
I tried to setup the matcher code as best (and as quickly) as I believe the issue is that since CEL can't determine which function overload to call `in_list` or `in_map` it marks the return value as dynamic. The reason for this internal type-checking confusion is because the `caddyPlaceholder` has a return type of `Any`.

```
caddyPlaceholder(request, "http.request.host")
    in caddyPlaceholder(request, "http.autohttps.redir_domains")
```

It happens that all of the function overloads have the same return type (not always the case), so I should be able to improve the type-resolution here to clearly mark the overall expression as `bool`.  In the mean time @mholt and @francislavoie, I would recommend updating the acceptable return types at the provision step to include `decls.Dyn` in addition to `decls.Bool`.

francislavoie:
Thanks, will do!

francislavoie:
Interesting, @TristonianJones I tried to eval the expression, and I'm actually getting "no such overload" in the output. Is that related to the compiled type being `dyn`?

TristonianJones:
The `dyn` type only exists at type check time. At runtime this means that the arg types didn't match any of the implementations. 

If you add `cel.EvalOptions(cel.OptTrackState)` to the `Program` call, the second return value will show you all of the intermediate values observed during eval. That will hopefully get you started on debugging

francislavoie:
👍 that helped, turns out my array is nil. Debugging further.

francislavoie:
@mholt I got this working, but you'll probably hate the design of this feature... but it works!

modules/caddyhttp/celmatcher.go
```diff
@@ -108,8 +108,11 @@ func (m *MatchExpression) Provision(_ caddy.Context) error {
 
 	// request matching is a boolean operation, so we don't really know
 	// what to do if the expression returns a non-boolean type
-	if !proto.Equal(checked.ResultType(), decls.Bool) {
-		return fmt.Errorf("CEL request matcher expects return type of bool, not %s", checked.ResultType())
+	// TODO: for now, we're allowing Dyn because certain expressions can't have
+	// their type inferred if there's ambiguity.
+	// See https://github.com/caddyserver/caddy/pull/3246#issuecomment-611256387
+	if !proto.Equal(checked.ResultType(), decls.Bool) && !proto.Equal(checked.ResultType(), decls.Dyn) {
```
francislavoie, line 114: We can remove this `&&` for `Dyn` once https://github.com/google/cel-go/pull/327 is available

TristonianJones: Released in cel-go v0.4.2

francislavoie: Thank you! I'll update the PR to pull that in soon

francislavoie: Confirmed, looks to work! Thanks!

francislavoie:
Rebased, updated the deps and removed the Dyn workaround

mholt:
Thanks for proposing this, but it isn't sitting right with me -- I think we need more valid uses cases before considering something that deals with this kind of complexity (the HTTP->HTTPS redirs are some of the most complex parts of Caddy's logic).

Closing for now, but we'll see if this is needed in the future.

PR #3255, opened 2020-04-11
fastcgi: Add new `php_fastcgi` subdirectives to override shortcut behaviour

francislavoie:
The [php_fastcgi](https://caddyserver.com/docs/caddyfile/directives/php_fastcgi) Caddyfile directive acts as a shortcut to a quite long expanded form. There are many situations where users might want to override some of the behaviour generated by the shortcut.

This change makes it possible to override some aspects of the shortcut by first reading the subdirective block once, handling any subdirectives that are relevant, and importantly, deleting the tokens as it goes. Then, it rewinds the Dispenser's cursor to the point just before so that the `reverse_proxy`'s UnmarshalCaddyfile can read it for itself. We delete as we go, because otherwise the `default:` case in `reverse_proxy` would return `d.Errf("unrecognized subdirective %s", d.Val())`.

Adds the following subdirectives:
- `split`: passthrough to the `split_path` subdirective of the `fastcgi` transport, but also affects the path matcher for the list of PHP file extensions to support
- `env`: passthrough to the `env` subdirective of the `fastcgi` transport
- `root`: passthrough to the `root` subdirective of the `fastcgi` transport
- `index`: defaults to `index.php`, controls the file that is used as a `try_files` fallback. If the value is set to `off`, it disables the path canonicalization redirect and the `try_files` rewrite altogether.

Caddyfile:

```
:8884

php_fastcgi localhost:9000 {
        # some php_fastcgi-specific subdirectives
        split .php .phpt
        env PATH /something
        root /var/somewhere
        index off

        # passed through to reverse_proxy (directive order doesn't matter!)
        lb_policy random
}
```

JSON output:

```
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":8884"
          ],
          "routes": [
            {
              "match": [
                {
                  "path": [
                    "*.php",
                    "*.phpt"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "reverse_proxy",
                  "load_balancing": {
                    "selection_policy": {
                      "policy": "random"
                    }
                  },
                  "transport": {
                    "env": {
                      "PATH": "/something"
                    },
                    "protocol": "fastcgi",
                    "root": "/var/somewhere",
                    "split_path": [
                      ".php",
                      ".phpt"
                    ]
                  },
                  "upstreams": [
                    {
                      "dial": "localhost:9000"
                    }
                  ]
                }
              ]
            }
          ]
        }
      }
    }
  }
}
```


francislavoie:
I was reading through the codebase some more and I realized a better approach than adding 2 new funcs to Dispenser to manipulate the cursor. I'm pretty sure I can use `NewFromNextSegment()`, mutate that new Dispenser as usual, then `Reset()` it and pass it down to `reverse_proxy`'s unmarshal.

I'll refactor this soon.

mholt:
@francislavoie When you get a chance, could you fix the merge conflict? Then I'll do a final review. Looking good though.

Note that I'll be relying on you for fixes to any potential bug reports that come in related to this in the future 😅  (esp. since I don't really have a PHP environment!)

Thanks!

francislavoie:
Yep, this one will need a bit of reworking actually, because of the `split_path` thing we added.

mholt:
Gotcha... no hurry. I can see the huge value in this though. Thanks!

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -123,47 +123,133 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
...
+		}
+
+		routes = append(routes, redirRoute, rewriteRoute)
 	}
 
 	// route to actually reverse proxy requests to PHP files;
 	// match only requests that are for PHP files
+	pathList := []string{}
+	for i := 0; i < len(extensions); i++ {
+		pathList = append(pathList, "*"+extensions[i])
```
mholt, line 249:
```suggestion
	for _, ext := range extensions {
		pathList = append(pathList, "*"+ext)
```

Not a big deal, but a little easier to read, and a little less that can go wrong for such a simple loop.

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -123,47 +123,133 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
...
+				if fcgiTransport.EnvVars == nil {
+					fcgiTransport.EnvVars = make(map[string]string)
+				}
+				fcgiTransport.EnvVars[args[0]] = args[1]
+
+			case "index":
+				args := dispenser.RemainingArgs()
+				dispenser.Delete()
+				for i := 0; i < len(args); i++ {
+					dispenser.Delete()
```
mholt, line 183: Check this out:

```suggestion
				for range args {
					dispenser.Delete()
```

caddytest/integration/caddyfile_adapt_test.go
```diff
@@ -45,3 +45,193 @@ func TestCaddyfileAdaptToJSON(t *testing.T) {
...
+								}
+							]
+						}
+					]
+				}
+			}
+		}
+	}
+}`)
+}
```
mholt, line 237: Is this one of the test cases that should be in a separate file now?

francislavoie: Yep! Forgot to do that, thanks for the reminder

a3dmorteza:
> Thanks! Code looks nice, I'll assume it works. Anyone with a PHP environment could please test it out, thank you!

I have the same problem at v2.9.1 in docker:

root of the php container is /app/public
root of the caddyfile is /srv/example.com

and when I send any request to caddy it says:
"error":"building environment: lstat /app: no such file or directory"

francislavoie:
This is an extremely old PR and no longer applicable to any issue you may be having.

PR #3265, opened 2020-04-15
httpcaddyfile: Make global options pluggable

francislavoie:
This is a refactor to make global options be registerable like Caddyfile directives.

It makes it possible for plugins/modules to add their own global options if their configuration doesn't make sense to be inside a server block.

The only functional change here should be the error messages.

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -120,6 +120,17 @@ func RegisterHandlerDirective(dir string, setupFunc UnmarshalHandlerFunc) {
 	})
 }
 
+// RegisterGlobalOption registers a unique global option opt with
+// an associated unmarshaling (setup) function. When the global
+// option opt is encountered in a Caddyfile, setupFunc will be
+// called to unmarshal its tokens.
+func RegisterGlobalOption(opt string, setupFunc UnmarshalGlobalFunc) {
+	if _, ok := registeredGlobalOptions[opt]; ok {
+		panic("global option " + opt + " already registered")
+	}
+	registeredGlobalOptions[opt] = setupFunc
+}
+
 // Helper is a type which helps setup a value from
 // Caddyfile tokens.
 type Helper struct {
@@ -454,6 +465,13 @@ type (
 	// for you. These are passed to a call to
 	// RegisterHandlerDirective.
 	UnmarshalHandlerFunc func(h Helper) (caddyhttp.MiddlewareHandler, error)
+
+	// UnmarshalGlobalFunc is a function which can unmarshal Caddyfile
+	// tokens into a global option config value using a Helper type.
+	// These are passed in a call to RegisterGlobalOption.
+	UnmarshalGlobalFunc func(d *caddyfile.Dispenser) (interface{}, error)
 )
 
 var registeredDirectives = make(map[string]UnmarshalFunc)
+
+var registeredGlobalOptions = make(map[string]UnmarshalGlobalFunc)
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -284,39 +284,18 @@ func (ServerType) evaluateGlobalOptionsBlock(serverBlocks []serverBlock, options
 		var val interface{}
 		var err error
 		disp := caddyfile.NewDispenser(segment)
-		switch dir {
-		case "debug":
-			val = true
-		case "http_port":
-			val, err = parseOptHTTPPort(disp)
-		case "https_port":
-			val, err = parseOptHTTPSPort(disp)
-		case "default_sni":
-			val, err = parseOptSingleString(disp)
-		case "order":
-			val, err = parseOptOrder(disp)
-		case "experimental_http3":
-			val, err = parseOptExperimentalHTTP3(disp)
-		case "storage":
-			val, err = parseOptStorage(disp)
-		case "acme_ca", "acme_dns", "acme_ca_root":
-			val, err = parseOptSingleString(disp)
-		case "email":
-			val, err = parseOptSingleString(disp)
-		case "admin":
-			val, err = parseOptAdmin(disp)
-		case "on_demand_tls":
-			val, err = parseOptOnDemand(disp)
-		case "local_certs":
-			val = true
-		case "key_type":
-			val, err = parseOptSingleString(disp)
-		default:
-			return nil, fmt.Errorf("unrecognized parameter name: %s", dir)
+
+		dirFunc, ok := registeredGlobalOptions[dir]
+		if !ok {
+			tkn := segment[0]
+			return nil, fmt.Errorf("%s:%d: unrecognized global option: %s", tkn.File, tkn.Line, dir)
 		}
+
+		val, err = dirFunc(disp)
 		if err != nil {
-			return nil, fmt.Errorf("%s: %v", dir, err)
+			return nil, fmt.Errorf("parsing caddyfile tokens for '%s': %v", dir, err)
 		}
+
 		options[dir] = val
 	}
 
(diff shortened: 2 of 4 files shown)


mholt:
Cool! I'm surprised this works 🤔 Will look at this more after 2.0!

caddyconfig/httpcaddyfile/options.go
```diff
@@ -23,7 +23,29 @@ import (
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
 )
 
-func parseOptHTTPPort(d *caddyfile.Dispenser) (int, error) {
+func init() {
```
hairyhenderson, line 26: Instead of using `init` can you make this a regular function, and call it explicitly at the beginning of `evaluateGlobalOptionsBlock`?

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -282,39 +282,18 @@ func (ServerType) evaluateGlobalOptionsBlock(serverBlocks []serverBlock, options
...
+		dirFunc, ok := registeredGlobalOptions[dir]
+		if !ok {
+			tkn := segment[0]
+			return nil, fmt.Errorf("%s:%d: unrecognized global option: %s", tkn.File, tkn.Line, dir)
 		}
+
+		val, err = dirFunc(disp)
 		if err != nil {
-			return nil, fmt.Errorf("%s: %v", dir, err)
+			return nil, fmt.Errorf("parsing caddyfile tokens for '%s': %v", dir, err)
```
hairyhenderson, line 294: Better idea to wrap `err` (now that [we can](https://golang.org/doc/go1.13#error_wrapping)!) 😉

```suggestion
			return nil, fmt.Errorf("parsing caddyfile tokens for '%s': %w", dir, err)
```

hairyhenderson: (note that the GitHub formatting doesn't recognize `%w` as a valid printf token yet - that's why it shows up red)

francislavoie: That's news to me! Still new in Go 😛

caddyconfig/httpcaddyfile/options.go
francislavoie: Hmm, I'm not sure I see the advantage to that. I'm matching the way it's done in `caddyconfig/httpcaddyfile/builtins.go`

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -120,6 +120,17 @@ func RegisterHandlerDirective(dir string, setupFunc UnmarshalHandlerFunc) {
 	})
 }
 
+// RegisterGlobalOption registers a unique global option opt with
+// an associated unmarshaling (setup) function. When the global
+// option opt is encountered in a Caddyfile, setupFunc will be
+// called to unmarshal its tokens.
+func RegisterGlobalOption(opt string, setupFunc UnmarshalGlobalFunc) {
```
hairyhenderson, line 127: Terminology question - these are called "global options" in the code, but the parallel non-global versions of these are called "directives"... Should this be `RegisterGlobalDirective`? Or does that have different semantics?

francislavoie: That's more a Q for @mholt but I'm following the nomenclature as seen here https://caddyserver.com/docs/caddyfile/options

mholt: Global options are for configuring things that are configured outside of an HTTP request / handler chain. If your directive just adds HTTP route(s), then you don't need/want global option.

caddyconfig/httpcaddyfile/options.go
francislavoie: After a separate discussion on slack, we can dismiss this; there was a misunderstanding in the goal of this PR 😄

hairyhenderson: Well I still think using `init()` where not completely necessary is not a great idea. In this case especially, there are already other `init()` functions, which means it all of a sudden becomes hard to know which code will get executed first.

Have a look at https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html for some discussion around this (and other things)

francislavoie: Order of execution doesn't matter here, all the work being done is just adding closures to a map. I'm just gonna follow Matt's lead here and do it the way it was already done for regular Caddyfile directives

francislavoie: FWIW I don't disagree with your take, I just think being consistent is more important here.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -270,39 +270,18 @@ func (ServerType) evaluateGlobalOptionsBlock(serverBlocks []serverBlock, options
...
+		dirFunc, ok := registeredGlobalOptions[dir]
+		if !ok {
+			tkn := segment[0]
+			return nil, fmt.Errorf("%s:%d: unrecognized global option: %s", tkn.File, tkn.Line, dir)
 		}
+
+		val, err = dirFunc(disp)
 		if err != nil {
-			return nil, fmt.Errorf("%s: %v", dir, err)
+			return nil, fmt.Errorf("parsing caddyfile tokens for '%s': %w", dir, err)
```
mholt, line 282: Are you sure you want to expose the error types to the exported API?

francislavoie: 🤷‍♂️ I just went off what @hairyhenderson recommended [here](https://github.com/caddyserver/caddy/pull/3265#discussion_r415425648)

caddyconfig/httpcaddyfile/httptype.go
mholt: @hairyhenderson Any particular reason to do this? It exposes error types to the exported API, potentially unintentionally -- we should be sure we want to do this, i.e. have a good reason before doing that, since once we export it, it'll be hard to change/remove it later.

caddyconfig/httpcaddyfile/options.go
hairyhenderson: Fair enough. Personally I think all the `init()`s used in the codebase should be revisited 😉

caddyconfig/httpcaddyfile/httptype.go
hairyhenderson: @mholt How do you mean?

From https://golang.org/pkg/fmt/#Errorf:

> If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand. It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. The %w verb is otherwise a synonym for %v.

So if there's an unexported error type, it should be safe, no?

caddyconfig/httpcaddyfile/options.go
mholt: @hairyhenderson Inits are really useful for registration functions.

hairyhenderson: @mholt agreed, except in this case the registration is something that can be done on-demand (and the current code _is_ on-demand)...

It's not a big deal, just makes things a little harder to reason about, and can make tests a bit more awkward...

caddyconfig/httpcaddyfile/httptype.go
mholt: /shrug Maybe, although I still question the need for this change until there's a need for it.

hairyhenderson: Fair enough. I just think that it's important to wrap errors where possible, to make downstream handling simpler. In this case it's no big deal.

francislavoie: So what's the verdict, do I revert the change?

caddyconfig/httpcaddyfile/httptype.go
francislavoie: @mholt so what's the verdict, so I revert this change based on the discussion above?

mholt: I don't care too strongly I guess, but I'd say let's revert it unless we have a specific reason to wrap errors in just this one spot.

francislavoie:
I'd like if we could merge this one earlier rather than later, useful for plugin authors and it would avoid merge conflicts to merge earlier if anything else needs to add/change global options

francislavoie:
Fixed :+1:

PR #3267, opened 2020-04-16
caddyconfig: Only parse `#` as the start of a comment if preceded by whitespace

francislavoie:
See discussion in https://caddy.community/t/example-for-redir/7475

Also updated `caddy fmt` to skip adding a space before `#` when it shouldn't

### DIFF
--- a/caddyconfig/caddyfile/lexer.go
+++ b/caddyconfig/caddyfile/lexer.go
@@ -74,7 +74,6 @@ func (l *lexer) load(input io.Reader) error {
 func (l *lexer) next() bool {
 	var val []rune
 	var comment, quoted, escaped bool
-	var spacePrior = true
 
 	makeToken := func() bool {
 		l.token.Text = string(val)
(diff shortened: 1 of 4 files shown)


caddyconfig/caddyfile/lexer.go
```diff
@@ -138,9 +140,12 @@ func (l *lexer) next() bool {
 			continue
 		}
 
-		if ch == '#' {
+		if ch == '#' && spacePrior {
```
mholt, line 143: Rather than adding a new variable to keep track of spaces, maybe a better approach would be to utilize the existing state, for example, `&& len(val) == 0` instead? Worth looking into, since it'd be simpler and would reuse existing token logic.

francislavoie: Ah yeah - `len(val) == 0` works perfectly, I didn't grok how val was used until you mentioned it. 👍

PR #3271, opened 2020-04-16
chore: Pull .gitattributes from v1 branch

francislavoie:
Not sure how much you care about this but I figured I'll pull it forwards, seems like a good idea to have 🤷‍♂️


mholt:
I don't think most of this stuff applies in our current master branch... I'd rather wait to see if it's needed first. Thanks though.

PR #3273, opened 2020-04-17
ci: Cache the GOCACHE directory to speed up builds and tests

francislavoie:
I'm thinking caching GOCACHE might speed up builds in CI... let's see.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ${{ steps.vars.outputs.go_cache }}
-        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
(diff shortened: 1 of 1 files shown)


francislavoie:
Maybe this is a dumb idea, but it definitely has a pretty noticeable speedup.

What do you guys think?

mholt:
Wow that is way faster. The results are still correct I imagine?

francislavoie:
I would assume so. I don't really want to assume that there's no side effects though.

mholt:
Alrighty, well, I'm willing to give it a go! Will let Mohammed review first when he has a chance.

francislavoie:
Hmm. @Mohammed90 can we try with just CI first then enable it for GoReleaser later? I don't want to risk breaking releases (especially since v2.0 stable is the next one coming) just for a speed increase.

PR #3276, opened 2020-04-17
reverseproxy: Set X-Forwarded-Proto

### LINKED ISSUE #3275: v2 - add forward-related headers by default
I just found out Caddy, firstly I'd like to say it looks amazing, new `reverse-proxy` mode seems so handy, even certificates are automatically generated S2

I just missed one thing; I'm running this command:

```
caddy reverse-proxy --from https://public.host:443 --to http://localhost:8081
```

And my upstream server expects some headers to know the request is actually being received through a proxy. It ends up generating a URL to redirect, and it uses `http://public.host`, instead of `https://public.host`, which breaks my app.

Any chances of getting this working?

mholt:
This sets the X-Forwarded-Proto header to upstreams by default; seems like a good idea for wider compatibility by default; overrideable if necessary.

@brunojcm can you please give this a try? Either build from this branch or download the CI artifacts when it completes.

Closes #3275.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -74,11 +74,19 @@ type Handler struct {
 	// Upstreams is the list of backends to proxy to.
 	Upstreams UpstreamPool `json:"upstreams,omitempty"`
 
+	// Adjusts how often to flush the response buffer. A
+	// negative value disables response buffering.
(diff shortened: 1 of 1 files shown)


brunojcm:
@mholt Fix is perfect!

``````

mholt:
Yep, it's the same as X-Forwarded-For in that regard.

Thanks, both!

Issue #3266, opened 2020-04-15
Strip path for reverse proxied requests


zakcutner:
#### Background

The [`reverse_proxy` directive](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy) does not strip the path of the request before forwarding the request upstream. For example, in the following configuration a request to `example.com/foo/bar` will result in an upstream request of `upstream:80/foo/bar` *(rather than `upstream:80/bar`)*.

```
example.com {
  route /foo* {
    reverse_proxy upstream:80
  }
}
```

At least from my experience, I quite often want the path to not be included in the upstream request, as the upstream server is a microservice which is not aware that it is being reverse-proxied. A good example given in #2813 (where this issue originated) is using phpMyAdmin at the URL `/phpmyadmin`. The [`uri` directive](https://caddyserver.com/docs/caddyfile/directives/uri) can be used to achieve this and the configuration can be changed to something like this.

```
example.com {
  route /foo* {
    uri strip_prefix /foo
    reverse_proxy upstream:80
  }
}
```

#### Issue

As this seems to be a relatively common pattern, I imagine configurations like the following will become quite commonplace.

```
example.com {
  route /api {
    route /v1* {
      uri strip_prefix /api/v1
      reverse_proxy oldapi:80
    }

    route /v2* {
      uri strip_prefix /api/v1
      reverse_proxy newapi:80
    }
  }

  route /docs* {
    uri strip_prefix /docs
    reverse_proxy docs:80
  }

  reverse_proxy website:80
}
```

Here there is a lot of duplication of paths, i.e. `/docs` is written twice. I would like to avoid this to make the configuration more succinct and because it could very easily lead to errors being accidentally introduced. For instance, if I change `/docs` to be `/documentation` I may forget to update `uri strip_prefix /docs` to `uri strip_prefix /documents`. In a more complex config like this one you can see how there may be more subtle errors with the api/ path as there is nesting too.

#### Ideas

I suggested a few ideas in #2813 and @francislavoie made some good points, also mentioning that in a lot of cases a subdomain may be preferred. Ideas that I have come up with are...

1. Change the default behaviour of `reverse_proxy` to strip out the path. Then you could simply add it back in by changing `reverse_proxy upstream:80` to `reverse_proxy upstream:80/foo`. However, @francislavoie pointed out that this could be a confusing change from the current behaviour in Caddy (both v1 and v2) and may be unintuitive to users who expect the path to be preserved.

[...]

PR #3281, opened 2020-04-19
caddyconfig: Add new `handle_path` directive for easier subpath grouping

francislavoie:
Implementation of https://github.com/caddyserver/caddy/issues/3266

~This is incomplete/non-functional, still has debug prints etc. The general idea is there but it's not working as expected.~

Edit: Figured it out! Didn't realize I should be using `h.NewRoute()` which does the trick.

Caddyfile:
```
:80

handle_path /api/v1 {
    respond "API v1"
}

file_server
```

JSON:
```
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":80"
          ],
          "routes": [
            {
              "match": [
                {
                  "path": [
                    "/api/v1*"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "handler": "rewrite",
                          "strip_path_prefix": "/api/v1"
                        }
                      ]
                    },
                    {
                      "handle": [
                        {
                          "body": "API v1",
                          "handler": "static_response"
                        }
                      ]
                    }
                  ]
                }
              ]
            },
            {
              "handle": [
                {
                  "handler": "file_server",
                  "hide": [
                    "../../../../caddy/Caddyfile-handle-path"
                  ]
                }
              ]
            }
          ]
        }
      }
    }
  }
}
```


caddyconfig/httpcaddyfile/directives.go
```diff
@@ -250,6 +251,63 @@ func (h Helper) NewBindAddresses(addrs []string) []ConfigValue {
 	return []ConfigValue{{Class: "bind", Value: addrs}}
 }
 
+// ParseSegmentAsSubroute parses the segment such that its subdirectives
+// are themselves treated as directives, from which a subroute is built
+// and returned.
+func ParseSegmentAsSubroute(h Helper) (caddyhttp.MiddlewareHandler, error) {
```
francislavoie, line 257: The only change here is `parseSegmentAsSubroute` -> `ParseSegmentAsSubroute`, but I moved the function up in the file because I think the convention is to keep exported functions together?

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -110,3 +114,53 @@ func parseCaddyfileURI(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, err
...
+	}
+	subroute, ok := handler.(*caddyhttp.Subroute)
+	if !ok {
+		return nil, h.Errf("segment was not parsed as a subroute")
+	}
+
+	// make a matcher on the path and everything below it
+	pathMatcher := caddy.ModuleMap{
+		"path": h.JSON(caddyhttp.MatchPath{path + "*"}),
+	}
```
francislavoie, line 151: I figure we could validate the path argument a bit more - could make sure the user didn't suffix it with `*` themselves.

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -110,3 +114,53 @@ func parseCaddyfileURI(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, err
...
+	path := h.Val()
+	if !strings.HasPrefix(path, "/") {
+		path = "/" + path
+	}
+
+	// the ParseSegmentAsSubroute function expects the cursor
+	// to be at the token just before the block opening,
+	// so we need to rewind because we already read past it
+	h.Reset()
+	h.Next()
```
francislavoie, line 136: This is a bit funky, but aside from changing how `ParseSegmentAsSubroute` works to avoid the extra `h.Next()`, I'm not sure what to do.

zakcutner:
This is super cool, thanks! My only concern is that I actually find the name `handle_path` slightly confusing as it doesn't express that it matches a prefix (rather than an exact match as usual) and that it strips the prefix rather than simply handling it — which I think is the killer behaviour here 😎

I don't have an amazing understanding of all the directives so perhaps this wouldn't make sense, but how about splitting up this functionality into two options which could be added to the existing `handle` directive?

1. `handle /foo prefix { ... }` would cause `/foo*` to be matched rather than `/foo`.
2. `handle /foo strip { ... }` would cause the first argument, `/foo`, to be stripped after matching.

Then, the functionality of `handle_path /foo { ... }` that you have implemented could be expressed using these options as `handle /foo prefix strip { ... }`. 

Alternatively, maybe it would be cleaner to add the `prefix` option to the path matcher in a similar way to how NGINX does in `location` blocks (the syntax `location foo { ... }` vs `location = /foo`). I think it also makes more sense for 'prefix mode' to actually be enabled by default as (just from my intuition 😬) I imagine this is the more common use case.

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -110,3 +114,53 @@ func parseCaddyfileURI(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, err
...
+		return nil, h.ArgErr()
+	}
+	if !h.NextArg() {
+		return nil, h.ArgErr()
+	}
+
+	// read the prefix to strip
+	path := h.Val()
+	if !strings.HasPrefix(path, "/") {
+		path = "/" + path
```
mholt, line 129: I think this should be an error. (See comment below)

modules/caddyhttp/rewrite/caddyfile.go
mholt: Hm, yeah, this is an interesting case... I think we shouldn't change their input _at all_, though, so things will stay consistent. i.e. they should use `/foo/*` if they want `/foo/*` (but we should obviously strip the `*` when trimming the path -- but I think the path matcher should be consistent to the user.

If the matcher token doesn't start with `/`, then we should return an error. This directive only works with path matchers, so the user should enter one, even if we do treat it a little special... it should be consistent.

caddyconfig/httpcaddyfile/directives.go
mholt: Sure -- doesn't matter too much. The go std lib puts functions all over the place, doesn't really matter if you know how to use your editor in the slightest.

mholt:
@zakcutner I propose we update this PR to accept a first argument that looks and acts like a standard path matcher token, i.e. `handle_path /foo/* ...` where we trim the `*` if present. Omitting it would still be an exact match, and it would still strip it, essentially turning the path into `/`. How's that?

francislavoie:
I agree with that - I'll try to finish this one off soon.

zakcutner:
Sound good 😃 It looks like this will solve a pretty specific use case so I agree there's not much point generalising it further like I was suggesting, particularly as it seems like that would be a lot of work. Thanks again to both of you for all your work on this!

francislavoie:
Alright, cleaned this up. I think this is done?

zakcutner:
Just tried this out and it works great, thank you both 😃

mholt:
Nice work on this one @francislavoie

PR #3285, opened 2020-04-20
refactored caddytest helpers

sarge:
Refactored the caddytests helpers. I am shooting for making the simple cases easy, and the complex ones possible.

I have exposed a low level function

```func AssertResponseCode(t *testing.T, req *http.Request, expectedStatusCode int) *http.Response {```

And have build a some of the higher level functions for some of the common cases. Building wrappers for all of the http verbs seems unnecessary.

@greenpau would you mind having a look and seeing if these are suitable?

### DIFF
--- a/caddytest/caddytest.go
+++ b/caddytest/caddytest.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"net/http/cookiejar"
 	"os"
 	"path"
 	"regexp"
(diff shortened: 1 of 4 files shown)


sarge:
@greenpau thanks for the feedback and suggestions. Interesting your observations about control over the http client. I had been considering an approach similar to the `http.DefaultClient` adding that as something you can do but I suspect seldom need to. It is global state, but it is familiar. 

Something I just needed to look up was about 'keeping the client state as one progresses through tests'. As I understand it cookies won't be remembered, what other state were you hoping to track? In general `useragent` should not be encouraged to drive behaviour, though I can see in practice it might be. `caddytest.DefaultClient` would give you that.

I also wonder how deep into http testing caddy really needs to get. It seems we will need to make a call about all the other httpverbs? I definitely wanted to shoulder the complexity of starting and loading configuration.

Not many answers here I know

greenpau:
@sarge , i implemented it here. Please Review https://github.com/caddyserver/caddy/pull/3299

greenpau:
> cookies won't be remembered

@sarge , I think with [this approach](https://github.com/caddyserver/caddy/pull/3299/commits/e5c521d1ccd66394fec644fa47711bf3e6c1528b#diff-04cda1e6d4da174d1919b0c57a3c86f9R31-R34) you can extend it to have a cookie jar (https://golang.org/pkg/net/http/cookiejar/)

sarge:
@greenpau I have taken another pass through these changes. I had added a cookie jar and a functioning test. I have also added few more verbs.

Take a look I think I have captured the spirit of what you needed.

greenpau:
> Take a look I think I have captured the spirit of what you needed.

@sarge , LGTM :+1:

greenpau:
@sarge , I approved the changes, but it looks like it does not count myself as an approver.

cc: @mholt

PR #3289, opened 2020-04-21
httpserver: Add H2C support

mholt:
This draft enables optional H2C support, both in the server and in the reverse proxy.

It also fixes a bug when recording responses (used primarily with HTTP access logging enabled, or the templates middleware) where trailers weren't properly getting flushed to the underlying response writer. As a bonus, that particular fix deletes code, eliminates copies, and should slightly improve performance!

Need to finalize the API before we merge this, but it should be ready shortly after the 2.0 release. Do not merge before 2.0.

@Zetanova could you please test this out? :+1:

Edit: Should be ready. On a server, you can set `"allow_h2c": true` to enable H2C support as a server, and on the proxy you can specify `"h2c"` as one of the Versions values of the HTTP Transport to enable the H2C client.


mholt:
I still need to do a little cleanup/finalization of the h2c-related config, then this PR will be out of draft phase.

modules/caddyhttp/responsewriter.go
```diff
@@ -122,46 +121,34 @@ type responseRecorder struct {
...
 //
-// After a response has been buffered, remember that any upstream header
-// manipulations are only manifest in the recorder's Header(), not the
-// Header() of the underlying ResponseWriter. Thus if you wish to inspect
-// or change response headers, you either need to use rec.Header(), or
-// copy rec.Header() into w.Header() first (see caddyhttp.CopyHeader).
+// The header map is not buffered; i.e. the ResponseRecorder's Header()
+// method returns the same header map of the underlying ResponseWriter.
+// This is a crucial design decision to allow HTTP trailers to be
+// flushed properly (https://github.com/caddyserver/caddy/issues/3236).
```
francislavoie, line 127: I'm a little unclear on how this works - does this mean the headers can't be modified (i.e. by `header` or `header_down`) after we get the response from the proxied service? I'm not sure I see the big picture of when that 
would happen.

mholt: (To be clear, this particular commit was already cherry-picked into master.)

I believe this change does not affect the manipulation of headers, it only eliminates the map-in-the-middle so to speak.

modules/caddyhttp/responsewriter.go
```diff
@@ -173,12 +160,11 @@ func (rr *responseRecorder) WriteHeader(statusCode int) {
 	if rr.shouldBuffer == nil {
 		rr.stream = true
 	} else {
-		rr.stream = !rr.shouldBuffer(rr.statusCode, rr.header)
+		rr.stream = !rr.shouldBuffer(rr.statusCode, rr.ResponseWriterWrapper.Header())
```
francislavoie, line 163: Okay I follow now after reading this line. I glossed over this one. 👍

AkinoKaede:
Will it support Caddyfile?

mholt:
@AkinoKaede The Caddyfile does not currently support this, because the Caddyfile is not structured around server-level settings (but rather, site-level settings), in other words: it is not possible to enable H2C for one site and disable it for another, it has to either be enabled or disabled for the whole server, but the Caddyfile doesn't have a good way to express that unless we enable H2C for _all_ servers globally.

francislavoie:
The issue #3669 is tracking Caddyfile support for `allow_h2c`

PR #3290, opened 2020-04-21
reverseproxy: always set the req.URL.Host with the upstream

westwin:
Issue Description: reverse proxy with round-robin policy always forwards the traffic to one upstream node.

How to Reproduce: 
1. set up upstream http service with two nodes(keep-alive enabled, on port 9010 and 9020)
2. set up caddy as a reverse proxy with config: 
:2015 {
    reverse_proxy /* 127.0.0.1:9010 127.0.0.1:9020 {
        lb_policy round_robin
        lb_try_duration 100ms
        lb_try_interval 250ms
    }
}
3. run the reverse proxy, and do a infinite loop test against :2015 to see the traffics are round-robin forwarded to 9010/9020
4. during the test, kill the service on 9010, and restart 9010
5. check the traffics will never forward to 9010, which always sticks to 9020

Test on macos

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -613,19 +613,17 @@ func (lb LoadBalancing) tryAgain(start time.Time, proxyErr error, req *http.Requ
 // directRequest modifies only req.URL so that it points to the upstream
 // in the given DialInfo. It must modify ONLY the request URL.
 func (h Handler) directRequest(req *http.Request, di DialInfo) {
-	if req.URL.Host == "" {
-		// we need a host, so set the upstream's host address
-		reqHost := di.Address
+	// we need a host, so set the upstream's host address
+	reqHost := di.Address
 
-		// if the port equates to the scheme, strip the port because
-		// it's weird to make a request like http://example.com:80/.
-		if (req.URL.Scheme == "http" && di.Port == "80") ||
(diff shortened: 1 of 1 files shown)


francislavoie:
Discussion: https://caddy.community/t/how-reverse-proxy-round-robin-works/7607

mholt:
Thanks for the PR!

I don't know what I was thinking when I wrote that `if` statement. It seems to go way back to this commit, waaay early (or before?) the beta versions: https://github.com/caddyserver/caddy/commit/0830fbad0347ead1dbea60e664556b263e44653f#diff-3820b4f14e2fee163a1894dbf2560bd9R476

It makes sense to remove it, since if the first request fails, it will try with the second backend, but the req.URL.Host will already be set to the host of the first backend 🤔 

So... I do think this fixes a bug. It's a pretty minor change and you've confirmed it works for you, too. I think we can get this out in v2.0.

Any objections? @francislavoie @sarge

mholt:
@westwin Before we can merge, we need you to sign the CLA.

sarge:
Looks fine to me:)

francislavoie:
@westwin for the CLA, you'll need to either add your email address `fengxi@fengxi.com` to your westwin Github account as a secondary email address [on this settings page](https://github.com/settings/emails), or make the commit again with git configured with an email address that's already on your account (see the settings page to see which ones you have)

PR #3299, opened 2020-04-22
caddytest: add AssertPostResponse and test client struct

greenpau:
Changes per my comment here: https://github.com/caddyserver/caddy/pull/3285#pullrequestreview-396450281

### DIFF
--- a/caddytest/caddytest.go
+++ b/caddytest/caddytest.go
@@ -27,6 +27,25 @@ import (
 	_ "github.com/caddyserver/caddy/v2/modules/standard"
 )
 
+// Tester represents an instance of a test client.
+type Tester struct {
+	Client *http.Client
+	t      *testing.T
+}
+
+// NewTester returns an instance of a test client.
+func NewTester(t *testing.T) *Tester {
+	return &Tester{
+		Client: &http.Client{
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+				return http.ErrUseLastResponse
+			},
+			Transport: createTestingTransport(),
+		},
+		t: t,
+	}
+}
+
 // Defaults store any configuration required to make the tests run
 type Defaults struct {
 	// Port we expect caddy to listening on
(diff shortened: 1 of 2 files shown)


greenpau:
cc: @sarge , @mholt

caddytest/caddytest.go
```diff
@@ -371,3 +392,75 @@ func AssertAdapt(t *testing.T, rawConfig string, adapterName string, expectedRes
...
+// AssertPostResponse asserts whether POST request returns expected status code and body.
+func AssertPostResponse(t *testing.T, requestURI string, requestHeaders []string, requestBody *bytes.Buffer, statusCode int, expectedBody string) (*http.Response, string) {
+	tester := NewTester(t)
+	return tester.AssertPostResponse(requestURI, requestHeaders, requestBody, statusCode, expectedBody)
+}
+
+// AssertPostResponse asserts whether POST request returns expected status code and body.
+func (tc *Tester) AssertPostResponse(requestURI string, requestHeaders []string, requestBody *bytes.Buffer, statusCode int, expectedBody string) (*http.Response, string) {
+	resp, body := AssertPostResponseBody(tc, requestURI, requestHeaders, requestBody, statusCode)
+	if !strings.Contains(body, expectedBody) {
```
sarge, line 405: You had an observation about assertions with html that was stripped of whitespaces. If there is a tester object perhaps we could deal with this by putting a transformation on the body.

The null implementation would be
`tc.TransformBody := func(resp *http.Response, body string) string { return body }`
This would allow for testing code bases to figure out how to transform the response to make the assertion cleaner.

caddytest/caddytest.go
```diff
@@ -371,3 +392,75 @@ func AssertAdapt(t *testing.T, rawConfig string, adapterName string, expectedRes
...
+		tc.t.Errorf("requesting \"%s\" expected response body \"%s\" but got \"%s\"", requestURI, expectedBody, body)
+	}
+	return resp, body
+}
+
+// AssertPostResponseBody performs http call to assert whether POST request returns expected status code and body.
+func AssertPostResponseBody(tc *Tester, requestURI string, requestHeaders []string, requestBody *bytes.Buffer, expectedStatusCode int) (*http.Response, string) {
```
sarge, line 412: I think we should push the requests through a common AssertStatusCode function that will take a http.Request object which will mean we can do other verbs too

https://github.com/caddyserver/caddy/pull/3285/files#diff-04cda1e6d4da174d1919b0c57a3c86f9R265

I am fine if want ot

caddytest/caddytest.go
```diff
@@ -371,3 +392,75 @@ func AssertAdapt(t *testing.T, rawConfig string, adapterName string, expectedRes
 		t.Fail()
 	}
 }
+
+// AssertPostResponse asserts whether POST request returns expected status code and body.
+func AssertPostResponse(t *testing.T, requestURI string, requestHeaders []string, requestBody *bytes.Buffer, statusCode int, expectedBody string) (*http.Response, string) {
+	tester := NewTester(t)
```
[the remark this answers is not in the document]
sarge: I think we could remove this, and always require a `Tester` object.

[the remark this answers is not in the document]
greenpau: > AssertPostResponse

@sarge , I actually like the flexibility :-)

caddytest/caddytest.go
```diff
@@ -27,6 +28,26 @@ import (
...
+}
+
+// NewTester returns an instance of a test client.
+func NewTester(t *testing.T) *Tester {
+	return &Tester{
+		Client: &http.Client{
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+				return http.ErrUseLastResponse
+			},
+			Transport: createTestingTransport(),
```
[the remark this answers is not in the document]
greenpau: Will look into it.

caddytest/caddytest.go
greenpau: > You had an observation about assertions with html that was stripped of whitespaces. If there is a tester object perhaps we could deal with this by putting a transformation on the body.

@sarge , it looks like it got fixed by someone. I don't have that problem now.

caddytest/caddytest.go
greenpau: >> I think we should push the requests through a common AssertStatusCode function that will take a http.Request object which will mean we can do other verbs too

:+1:

caddytest/caddytest.go
sarge: OK we can leave this out for now.

greenpau:
Closing as the changes are in #3285

PR #3300, opened 2020-04-22
caddyhttp: Fix common_log format's user ID placeholder

francislavoie:
I was digging around for the `http.auth.user.id` placeholder in the code and noticed that the common_log format was using the wrong placeholder.


francislavoie:
As an aside, I don't think `{http.auth.user.id}` is documented anywhere. That should probably be added in a couple different spots (`basicauth` directive and in the authentication module docs I figure)

mholt:
Thanks for the fix!

You are right that it's not documented anywhere -- yet; this is intentional as the auth features aren't really finished/set in stone yet, but there's some good progress on auth plugins coming along by the community so I think we can document them better soon enough!

Issue #3294, opened 2020-04-21
v2: log params not taken into account


fmarcia:
I'm trying to adjust log to my need:
```
app1.mydomain.tld,
app2.mydomain.tld,
app3.mydomain.tld {
    header -server
    log {
        output stderr
        format single_field common_log
        level error
    }
    file_server
    root * /srv/http/www
}
```
which is translated in `autosave.json` to:
```json
{
    // ...
    "logging": {
        "logs": {
            "log5": {
                "encoder": {
                    "field": "common_log",
                    "format": "single_field"
                },
                "include": ["http.log.access.log5"],
                "level": "error",
                "writer": {
                    "output": "stderr"
                }
            }
        }
    }
}
```

But I keep getting a complete json log and for any level:
```
2020-04-21T17:57:35+0200 homeserver caddy[84787]: {"level":"info","ts":1587484655.2077143,"logger":"http.log.access.log5","msg":"handled request","request":{"method":"GET","uri":"/favicon.ico","proto":"HTTP/2.0","remote_addr":"{ip_address}:44450","host":"app1.mydomain.tld","headers":{"User-Agent":["Mozilla/5.0 (X11; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0"],"Accept":["image/webp,*/*"],"Accept-Language":["fr,en-US;q=0.7,en;q=0.3"],"Accept-Encoding":["gzip, deflate, br"],"Cache-Control":["max-age=0"],"Te":["trailers"]},"tls":{"resumed":false,"version":772,"ciphersuite":4865,"proto":"h2","proto_mutual":true,"server_name":"app1.mydomain.tld"}},"common_log":"{ip_address} - - [21/Apr/2020:17:57:35 +0200] \"GET /favicon.ico HTTP/2.0\" 200 0","latency":0.000502759,"size":0,"status":200,"resp_headers":{"Etag":["\"q95azobvy\""],"Content-Type":[],"Last-Modified":["Tue, 21 Apr 2020 15:57:24 GMT"],"Accept-Ranges":["bytes"],"Content-Length":["15406"]}}
```

mholt:
That's probably because you're seeing the output of the default log (`"default"`) which is always enabled unless you customize it: https://caddyserver.com/docs/json/logging/#docs

clonejo:
for the record, reconfiguring the default logger can look like this:
```Caddyfile
{
    log default {
        output stderr
        level ERROR
        format console
    }
}

[...]

PR #3301, opened 2020-04-22
caddyhttp: General improvements to access logging

mholt:
Even though any logs can just be ignored, most users don't seem to like configuring an access log to go to a file only to have it doubly appear in the default log.

This also fixes a few little bugs related to access log controls in the Caddyfile. Users now have the ability for fine-grained customization over which requests to which hosts get access-logged.

Related to:
- #3294
- https://caddy.community/t/v2-logging-format/7642/4?u=matt
- https://caddy.community/t/caddyfile-questions/7651/3?u=matt
- #3310

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -397,7 +397,7 @@ func (s *Server) shouldLogRequest(r *http.Request) bool {
 type ServerLogConfig struct {
 	// The default logger name for all logs emitted by this server for
 	// hostnames that are not in the LoggerNames (logger_names) map.
-	LoggerName string `json:"logger_name,omitempty"`
+	DefaultLoggerName string `json:"default_logger_name,omitempty"`
 
 	// LoggerNames maps request hostnames to a custom logger name.
 	// For example, a mapping of "example.com" to "example" would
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/server.go
```diff
@@ -367,17 +367,52 @@ func (*HTTPErrorConfig) WithError(r *http.Request, err error) *http.Request {
...
+// enabled without customization, all requests to this server are
+// logged to the default logger; logger destinations may be
+// customized per-request-host.
 type ServerLogConfig struct {
-	// The logger name for all logs emitted by this server unless
-	// the hostname is found in the LoggerNames (logger_names) map.
-	LoggerName string `json:"log_name,omitempty"`
+	// The default logger name for all logs emitted by this server for
+	// hostnames that are not in the LoggerNames (logger_names) map.
+	LoggerName string `json:"logger_name,omitempty"`
```
francislavoie, line 400: Technically a breaking change, isn't it?

Would this make more sense as `default_logger_name` to indicate it's a fallback?

mholt: Yeah nobody is using this yet. (Also it's still in pre-release.)

Hmm, yeah, that could be good too. Just didn't know if that was too long. I guess it's not.

PR #3302, opened 2020-04-23
matcher: Add `split_path` option to file matcher; used in php_fastcgi

francislavoie:
Discussion: https://caddy.community/t/help-to-migrate-caddyfile-v1-to-v2-for-nextcloud/7647/18

A common pattern in PHP apps is to have a PHP file that acts as a router entrypoint to the application. This is typically the `index.php` file at the webroot or some other subpath of the application. These entrypoints typically accept requests that have a path appended to the filename, for example `/index.php/blog/123`. 

Some PHP apps happen to have more than one router entrypoint, for example, `/router.php/some/path`, in addition to the `index.php` file. NextCloud is one of those applications; it has `remote.php`, `public.php` and more that act as entrypoints.

With the default `php_fastcgi` Caddyfile directive, many of those pages didn't work correctly in NextCloud. The problem has to do with the `try_files` matcher and rewrite that is included as part of the `php_fastcgi` shortcut:

```
try_files {path} {path}/index.php index.php
```

The issue is that on a request to the path `/remote.php/dav/files`, Caddy would check if that exact path exists as a file on disk (i.e. the first argument, `{path}`), but this doesn't work because there's extra stuff after the actual filename. 

We already have an option in the reverse_proxy's fastcgi transport to allow splitting the path into two parts, with the left-hand side being the filename (including the split value), and the right-hand side being the path, i.e. `split .php`.

The fix we landed on for this issue is to add a new `split` option to the file matcher (`split_path` in JSON) that allows for splitting the path the same way, before testing if the file exists on disk. It looks like this:

```
@phpFiles {
    file {
        split .php
        try_files {path} {path}/index.php index.php
    }
}
```

So now for a request to `/remote.php/dav/files`, Caddy will instead look for `/remote.php` on disk while matching.

The `php_fastcgi` shortcut is also updated to internally add this new split option for its `try_files` matcher. This should not be a breaking change, because we assume that any path that involved `.php` in the middle of it was assumed to be the filename boundary, but this wasn't working as expected.


francislavoie:
@mholt I'll let you pick the milestone, not sure when you think this should go in.

francislavoie:
I kept that mostly to match the reverse_proxy behaviour. I figure if a request to `/Remote.PHP/some/path` comes in, it should still work.

mholt:
I'd be willing to merge this in before 2.0 if the original reporters of the bug confirm that it works for them.

tobya:
We had some issues in 1 around split_path for php files and incorrect matches.  do you think there is any possibility this will suffer from similar issues.

eg.

Would it be possible to upload a file 

myhackfile.php.txt and have it executed as myhackfile.php?

This would obviously be a security issue.

francislavoie:
> Would it be possible to upload a file myhackfile.php.txt and have it executed as myhackfile.php?

No, because `try_files` would look for `myhackfile.php` on disk and not find it, so it would serve that path the path `myhackfile.php.txt` through `index.php` instead.

But, this does mean this change would break requests to files that have `.php` in the middle of them though.

Hmm. Should we only return the split path if the next character after the split is `/` or is the end of the string?

francislavoie:
Okay I tried adding this bit just before the return in the loop:

```go
			// skip the split if it's not the final part of the filename
			if pos != len(path) && ! strings.HasPrefix(path[pos:], "/") {
				continue
			}
```

Caddyfile:
```
:8886

root ./test

@try {
  file {
    split .php
    try_files {path} {path}/index.php index.php
  }
}

rewrite @try {http.matchers.file.relative}

respond "Path: {path} | Relative: {http.matchers.file.relative}"
```

Tree:
```
.
├── caddy
├── main.go
└── test
    ├── index.php
    ├── myhackfile.php.txt
    └── remote.php
```

Tests:
```
$ curl localhost:8886
Path: /index.php | Relative: /index.php

$ curl localhost:8886/index.php
Path: /index.php | Relative: /index.php

$ curl localhost:8886/index.php/somewhere
Path: /index.php | Relative: /index.php

$ curl localhost:8886/remote.php
Path: /remote.php | Relative: /remote.php

$ curl localhost:8886/remote.php/somewhere
Path: /remote.php | Relative: /remote.php

$ curl localhost:8886/myhackfile.php
Path: index.php | Relative: index.php

$ curl localhost:8886/myhackfile.php.txt
Path: /myhackfile.php.txt | Relative: /myhackfile.php.txt

$ curl localhost:8886/myhackfile.php.txtxt
Path: index.php | Relative: index.php

$ curl localhost:8886/myhackfile.php.txt/
Path: /myhackfile.php.txt | Relative: /myhackfile.php.txt

$ curl localhost:8886/myhackfile.php.txt/somewhere
Path: index.php | Relative: index.php
```

With that change, as you can see, it allows direct requests to `myhackfile.php.txt` (or if the file is followed by `/` which I think is harmless? Hmm.) but not if followed by anything else.

francislavoie:
In other news, we got word from the user on the forums that this change fixed the problem for them with NextCloud:

https://caddy.community/t/help-to-migrate-caddyfile-v1-to-v2-for-nextcloud/7647/25?u=francislavoie

mholt:
@tobya You are referring to this issue? https://github.com/caddyserver/caddy/pull/2917

modules/caddyhttp/fileserver/matcher.go
```diff
@@ -68,6 +68,15 @@ type MatchFile struct {
...
 	// Default is first_exist.
 	TryPolicy string `json:"try_policy,omitempty"`
+
+	// A list of delimiters to use to split the path in two
+	// when trying files. If empty, no splitting will
+	// occur, and the path will be tried as-is. For each
+	// split value, the left-hand side of the split,
+	// including the split value, will be the path tried.
+	// For example, the path `/remote.php/dav/` using the
+	// split value `.php` would try the file `/remote.php`.
```
mholt, line 78: Do you think the godoc should mention that all delimiters will be suffixed with `/`?

modules/caddyhttp/fileserver/matcher.go
```diff
@@ -256,6 +270,26 @@ func strictFileExists(file string) bool {
...
+// result is the first piece of the path that ends with
+// in the split value. Returns the path as-is if the
+// path cannot be split.
+func (m MatchFile) firstSplit(path string) string {
+	lowerPath := strings.ToLower(path)
+	for _, split := range m.SplitPath {
+		if idx := strings.Index(lowerPath, strings.ToLower(split)); idx > -1 {
+			pos := idx + len(split)
+			// skip the split if it's not the final part of the filename
+			if pos != len(path) && !strings.HasPrefix(path[pos:], "/") {
```
mholt, line 284: Would it be simpler (and still correct?) if we just suffix each split with `/`?

modules/caddyhttp/fileserver/matcher.go
francislavoie: Technically they're not all suffixed, because `/remote.php` still works (without a final `/`), but yeah it could mention it

modules/caddyhttp/fileserver/matcher.go
francislavoie: No because then `/remote.php` wouldn't work

mholt:
Are untrusted users ever able to upload directories?

```

$ curl localhost:8886/myhackfile.php.txt/
Path: /myhackfile.php.txt | Relative: /myhackfile.php.txt
```

^ that test, if a user uploads a file in a directory like `myhackfile.php.txt/index.php`, would a request to `/myhackfile.php.txt` invoke the index.php file within it?

francislavoie:
At that point it's not Caddy's problem, it's the apps. If they allow things like that, then it's an insecure app. I don't think we need to worry about it.

mholt:
Not that I disagree with you, but, allow me to play devil's advocate here, since I know we will get complainers if we don't drill down ahead of them: is there a reason that is the app's fault but not allowing file uploads with filename like `foo.php.php`? Where do we draw the line between app responsibility and Caddy's responsibility?

modules/caddyhttp/fileserver/matcher.go
mholt: Good point

modules/caddyhttp/fileserver/matcher.go
mholt: Well, at least mention that the delimiters have to be at the end of either the path or the filename, i.e. have to be followed by a `/` if not at the end of the path.

francislavoie:
I added a file as described on disk:

```
$ tree
.
├── caddy
├── main.go
└── test
    ├── foo.php.php
    │   └── index.php
    ├── index.php
    ├── myhackfile.php.txt
    ├── nonphpfile.txt
    └── remote.php
```

And tried requesting it (same Caddyfile as above):
```
$ curl localhost:8886/foo.php.php/index.php
Path: index.php | Relative: index.php
```

It ends up falling back to `/index.php` because it tries for `/foo.php` which doesn't exist on disk. This is totally fine, this is the expected case. If a user has a legitimate use-case for a file named that way, they can add their own rewrite rule ahead of the one built into `php_fastcgi` so it goes through.

Also, here is the matcher in the Caddyfile above vs what is now generated by the `php_fastcgi` directive (only difference is `"group"` which is an implementation detail):

```
Caddyfile:
            {
              "group": "group0",
              "match": [
                {
                  "file": {
                    "try_files": [
                      "{http.request.uri.path}",
                      "{http.request.uri.path}/index.php",
                      "index.php"
                    ],
                    "split_path": [
                      ".php"
                    ]
                  }
                }
              ],
              "handle": [
                {
                  "handler": "rewrite",
                  "uri": "{http.matchers.file.relative}"
                }
              ]
            },
php_fastcgi:
            {
              "match": [
                {
                  "path": [
                    "*.php"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "reverse_proxy",
                  "transport": {
                    "protocol": "fastcgi",
                    "split_path": [
                      ".php"
                    ]
                  },
                  "upstreams": [
                    {
                      "dial": "some:9000"
                    }
                  ]
                }
              ]
            }
```

Then tried with this Caddyfile (PHP 7.4), each file just has their own filename in it:

```
:8886

root ./test

php_fastcgi unix//run/php/php7.4-fpm.sock
```

```
$ curl localhost:8886/index.php
index

$ curl localhost:8886/index.php/somewhere
index

$ curl localhost:8886/remote.php
remote

$ curl localhost:8886/remote.php/somewhere
remote

$ curl localhost:8886/myhackfile.php
index

$ curl localhost:8886/myhackfile.php.txt
index

$ curl localhost:8886/myhackfile.php.txtxt
index

$ curl localhost:8886/myhackfile.php.txt/
index

$ curl localhost:8886/myhackfile.php.txt/somewhere
index

$ curl localhost:8886/foo.php.php/index.php
index
```

I think we're safe here, I don't see any issues. I still want @whitestrake's input though!

francislavoie:
Okay sorry about that - I edited the comment above, I made a booboo and I ran all of that under the wrong code 🤦‍♂️

francislavoie:
Hey @sarge I wouldn't mind your help on this one to add matcher tests similarly to those `curl` requests I did above! I'm not sure where that should go or how to set up the tests

francislavoie:
Don't mind the change to `.github/CONTRIBUTING.md`... that was just a `dos2unix` line endings change. I mucked up the file endings when I made the file the other day. The joys of working on Windows.

francislavoie:
Perfect! Thanks @whitestrake that's essentially what I expected to hear 😄 Just wanted someone to check my sanity here ❤️

francislavoie:
Bah, I hate that Github dismisses reviews after rebasing... but I rebased. This is ready to go @mholt.

mholt:
I think I can turn the auto-dismissal off, but I specifically want it on for PRs to the master branch to make sure something doesn't slip in too easily. No worries.

PR #3305, opened 2020-04-24
httpcaddyfile: Add shorthands for parameterized placeholders

francislavoie:
Adds the following new placeholder shorthands to the Caddyfile (plus one non-`*` placeholder):

```
{port}      ->  {http.request.port}
{query.*}   ->  {http.request.uri.query.*}
{labels.*}  ->  {http.request.host.labels.*}
{header.*}  ->  {http.request.header.*}
{path.*}    ->  {http.request.uri.path.*}
{re.*.*}    ->  {http.regexp.*.*}
```

These are quite commonly used placeholders. Because they're relatively long, it's hard to remember the exact syntax on the fly, requiring frequent referencing of the docs to verify.

~Currently I implemented this via just a quick prefix replacement, it's unlikely users would use these string patterns on their own. Probably the most concerning would be `{r.` because it's such a short sequence.~

~I chose the very short `{r.*.*}` as the syntax for the regexp placeholder, because when users provide a name argument, it's pretty obvious what `r` is referencing, since the name will be associated with a `path_regexp` or `header_regexp` matcher. It's easy to trace in the Caddyfile.~

Edit: Switched to `{re.*.*}` for some added clarity.

~As an improvement to this, in second iteration I think we could implement this via regexp replacements instead, it would probably be safer as well because we'd verify the final `}` is actually present. Just wanted to get out a quick proof of concept first.~

Edit: Done, implemented using regexp. Pretty easy.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -109,7 +109,7 @@ func (st ServerType) Setup(inputServerBlocks []caddyfile.ServerBlock,
 		{regexp.MustCompile(`{labels\.([\w-]*)}`), "{http.request.host.labels.$1}"},
 		{regexp.MustCompile(`{header\.([\w-]*)}`), "{http.request.header.$1}"},
 		{regexp.MustCompile(`{path\.([\w-]*)}`), "{http.request.uri.path.$1}"},
-		{regexp.MustCompile(`{r\.([\w-]*)\.([\w-]*)}`), "{http.regexp.$1.$2}"},
+		{regexp.MustCompile(`{re\.([\w-]*)\.([\w-]*)}`), "{http.regexp.$1.$2}"},
 	}
 
 	for _, sb := range originalServerBlocks {
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -96,9 +97,30 @@ func (st ServerType) Setup(originalServerBlocks []caddyfile.ServerBlock,
...
+		// for those, so we use a regexp to replace
+		regexpReplacements := []struct {
+			search  string
+			replace string
+		}{
+			{search: `{query\.([\w-]*)}`, replace: "{http.request.uri.query.$1}"},
+			{search: `{labels\.([\w-]*)}`, replace: "{http.request.host.labels.$1}"},
+			{search: `{header\.([\w-]*)}`, replace: "{http.request.header.$1}"},
+			{search: `{path\.([\w-]*)}`, replace: "{http.request.uri.path.$1}"},
+			{search: `{r\.([\w-]*)\.([\w-]*)}`, replace: "{http.regexp.$1.$2}"},
```
francislavoie, line 112: I'm not certain if `[\w-]` (a.k.a. `[a-zA-Z0-9_-]`) is the right set of chars to allow. Maybe that's too restrictive. What else would appear there?

mholt: That ought to cover 99.9% of use cases, I think.

mholt:
@francislavoie When you have a chance could you help with the merge conflict? I'll do another review then.

francislavoie:
Rebased! Btw, easier to review with whitespace changes hidden.

mholt:
Thanks, looks good!

My **only** remaining question is about the use of `r`. I like short, but I almost never see `r` used to mean `regular expression`. I usually see `regex` or `regexp` or _at least_ `re`.

What do you think of `re`?

francislavoie:
`re` is okay :+1: does make me think of like `Re:` (regarding) in emails, but I don't find it particularly unclear.

PR #3280, opened 2020-04-19
ci: Enable GoReleaser .deb support

francislavoie:
This needs https://github.com/caddyserver/dist/pull/23 before it can actually be tested.

Work in progress.

TODO:
- [ ] Set up uploading to https://gemfury.com
- [x] Improve scripts to deal with `caddy-api` service file
  - In postinstall:
    - if `caddy` is enabled, restart that one
    - if `caddy-api` is enabled, then restart that one
    - else enable and start `caddy`
  - In preremove:
    - if `caddy` is enabled, then stop and disable that one
    - if `caddy-api` is enabled, then stop and disable that one
- [ ] Clean up CI config afterwards to put it back in "release mode"
  - Turn off the PR trigger
  - Remove `--skip-publish` in GoReleaser
  - Remove publishing the `.deb` as Github artifacts

PR #3309, opened 2020-04-26
ci: Enable GoReleaser .deb support

francislavoie:
Replaces #3280

I need to do this from a branch on this repo to be able to test using secrets, i.e. for gemfury.

~Work in progress.~

~TODO:~
- [x] Set up uploading to https://gemfury.com
- [x] Improve scripts to deal with `caddy-api` service file
  - In postinstall:
    - if `caddy` is enabled, restart that one
    - if `caddy-api` is enabled, then restart that one
    - else enable and start `caddy`
  - In preremove:
    - if `caddy` is enabled, then stop and disable that one
    - if `caddy-api` is enabled, then stop and disable that one
- [x] Clean up CI config afterwards to put it back in "release mode"
  - Turn off the PR trigger
  - Remove `--skip-publish` in GoReleaser
  - Remove publishing the `.deb` as Github artifacts

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -72,9 +72,9 @@ nfpms:
       ./caddy-dist/config/Caddyfile: /etc/caddy/Caddyfile
 
     scripts:
-      preinstall: ./caddy-dist/scripts/preinstall.sh
       postinstall: ./caddy-dist/scripts/postinstall.sh
       preremove: ./caddy-dist/scripts/preremove.sh
+      postremove: ./caddy-dist/scripts/postremove.sh
 
 
 release:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -50,8 +50,6 @@ nfpms:
   - id: default
     package_name: caddy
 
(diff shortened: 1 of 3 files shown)


.goreleaser.yml
```diff
@@ -43,12 +45,45 @@ archives:
       darwin: mac
 checksum:
   algorithm: sha512
+
+nfpms:
+  - id: default
+    package_name: caddy
+
+    # file_name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}"
```
mholt, line 53: Is this a TODO or something?

francislavoie: Ah, no, just wanted to leave it in as a note that we can modify the filename template. I think we'll probably stick with the default because it matches the rest of our archives.

mholt:
Thanks for your hard work, @francislavoie, and for the review @Mohammed90 !

Issue #2920, opened 2019-12-09
Modify status code of reverse proxy response 


hackwaly:
Could it be done by only config?
And are there reverse_proxy response placeholders? Such as headers, status_code, body etc.

mholt:
Hi, welcome to the issue tracker!

I am not entirely sure that I understand your post. Do you have 2 questions here?

1. How to modify the response status code from an upstream?
2. Whether there are placeholders for reverse proxy? Can you explain more what you mean?

AbdelmalekIhdene:
Arguably, you'd say that those configurations should be the responsibility of the upstreams you are passing the request to, i.e. a proxy would only act as a middle-man in a request transaction.
If you want conditionals for different responses then caddy would not be the right answer for you since I'm sure @mholt is steering away from directives like [if](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)

mholt:
Hmmm no I believe this should be possible with Caddy 2, without imperative syntax -- if it's not possible already, then we should implement it. I just need to know exactly what needs to be implemented. So first, we need some clarification from @hackwaly.

hackwaly:
I need to use config to achieve:

* Rewrite headers and status_code of reverse_proxy response

mholt:
I almost understand -- can you give an actual example of one such configuration you're trying to accomplish?

hackwaly:
BTW: I already wrote a go program to get what i want.

Config may like this:

```json
{
  "apps": {
    "proxy_app": {
      "servers": {
        "proxy_server": {
          "routes": [
            {
              "handle": [
                {
                  "handler": "reverse_proxy",
                  "upstreams": [
                    {
                      "dial": "{host}"
                    }
                  ],
                  "headers": {
                    "handler": "headers",
                    "response": {
                      "delete": ["Set-Cookie"],
                      "set": {
                        "X-Set-Cookie": ["{http.handlers.reverse_proxy.response.header.Set-Cookie}"],
                        "Access-Control-Allow-Origin": ["*"]
                      },
                      "set_status_code": "{http.handlers.reverse_proxy.response.status_code >= 400 ? 500 : 200}"
                    }
                  }
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    }
  }
}
```

mholt:
@hackwaly Neat, is the source anywhere?

mholt:
Ping @hackwaly  -- how did you implement that feature?

[...]

PR #3313, opened 2020-04-26
caddyhttp: Implement status_code handler

### LINKED ISSUE #3312: Add a handler to allow modifying the HTTP status code of a response
Currently, we have a `respond` (or `static_response` in JSON) handler which can serve an arbitrary request body with an arbitrary status code, and we have a `file_server` handler which can serve arbitrary files from disk, but no way to serve a file with an arbitrary status code.

This would be particularly useful in situations where users want to serve specific files on error or as a fallback, for example a `/404.html` file should be combined with status code 404.

I propose we add a `status_code` handler which would simply set the status code of the response as a request middleware.

For example, in JSON it might look like this:
```
{
    "handler": "status_code",
    "status_code": 404
}
```

In Caddyfile:
```
status_code [<matcher>] 404
```

In the Caddyfile directive ordering, I suggest adding it under `encode` and before `templates`.

Example usage:
```
# Try to serve the requested file at {path}, fallback to 404.html otherwise
try_files {path}.php {path} 404.html

# Change the status code to 404 if the file being served is 404.html
status_code /404.html 404

file_server
```

francislavoie:
Closes #3312 

Adds a simple `status_code` handler which simply modifies the HTTP status code of a response.

JSON:
```
{
    "handler": "status_code",
    "status_code": 404
}
```

Caddyfile:
```
status_code [<matcher>] <status>
```

---

### Example usecase:

Caddyfile:
```
:8884

root .

try_files {path} /404.html
status_code /404.html 404

file_server
```

Request:
```
$ curl -v localhost:8884/nope
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8884 (#0)
> GET /nope HTTP/1.1
> Host: localhost:8884
> User-Agent: curl/7.58.0
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Server: Caddy
< Date: Mon, 27 Apr 2020 06:21:34 GMT
< Content-Length: 4
<
This is a 404 response page.
```

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -35,6 +35,7 @@ func init() {
 	RegisterHandlerDirective("root", parseRoot)
 	RegisterHandlerDirective("redir", parseRedir)
 	RegisterHandlerDirective("respond", parseRespond)
+	RegisterHandlerDirective("status_code", parseStatusCode)
 	RegisterHandlerDirective("route", parseRoute)
 	RegisterHandlerDirective("handle", parseHandle)
 	RegisterDirective("handle_errors", parseHandleErrors)
@@ -386,6 +387,16 @@ func parseRedir(h Helper) (caddyhttp.MiddlewareHandler, error) {
 // parseRespond parses the respond directive.
 func parseRespond(h Helper) (caddyhttp.MiddlewareHandler, error) {
 	sr := new(caddyhttp.StaticResponse)
+	err := sr.UnmarshalCaddyfile(h.Dispenser)
+	if err != nil {
+		return nil, err
+	}
+	return sr, nil
+}
+
+// parseStatusCode parses the status_code directive.
+func parseStatusCode(h Helper) (caddyhttp.MiddlewareHandler, error) {
+	sr := new(caddyhttp.StatusCode)
 	err := sr.UnmarshalCaddyfile(h.Dispenser)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


Mohammed90:
Isn't the functionality duplicate of [`static_response`](https://caddyserver.com/docs/json/apps/http/servers/routes/handle/static_response/)? You can already do:

```
{
	"handler": "static_response",
	"status_code": "400",
	"close": false
}
```
and the `"close": false` is to allow the handler chain to continue.

francislavoie:
Yes that's true, but there's no way to specify an empty body via Caddyfile, and the syntax wouldn't allow for it either.

francislavoie:
Yeah I see what you mean. I made this PR before actually trying it cause I was overconfident 😅 

I needed to return `next.ServeHTTP(w, r)` which actually does work, but it spits out this warning every time:

```
2020/04/27 01:35:39 http: superfluous response.WriteHeader call from github.com/caddyserver/caddy/v2/modules/caddyhttp/fileserver.(*FileServer).ServeHTTP (staticfiles.go:239)
```

francislavoie:
Turns out `NewResponseRecorder` is exactly what we need here 👍 

It works now, I think this is what you were thinking of.

That said, do you think it would make sense at all to change StaticResponse to wrap the response writer like this and chain if only the status is specified and no body? In other words, keep the response body the same, but just update the status code if that's the only parameter.

mholt:
I think the solution for this problem and https://github.com/caddyserver/caddy/issues/2920 are the same, and I wonder if the right place for it is actually in the existing `subroute` handler.

The idea is that you capture the response, like you're doing, but then when WriteHeader is called, we invoke a subroute (handler chain) to be evaluated. In theory, this would allow the user to customize the status code that gets written, along with the headers and body if they want to, _based on the response headers_, which is something that can't currently be done.

francislavoie:
As noted in https://github.com/caddyserver/caddy/issues/3312#issuecomment-632358395, this isn't really the right approach. Closing.

PR #3323, opened 2020-04-29
caddy: Add support for `d` duration unit

francislavoie:
This has come up a few times, most recently https://twitter.com/_francislavoie/status/1255374426612383745

Adds support for a `d` suffix for number of days (i.e. 24 hours) in duration strings wherever Caddy parses them.

Caddyfile:
```
:80

log {
	output file /var/log/access.log {
		roll_size 1gb
		roll_keep 5
		roll_keep_for 90d
	}
}
```

JSON:
```json
{
	"logging": {
		"logs": {
			"default": {
				"exclude": [
					"http.log.access.log0"
				]
			},
			"log0": {
				"writer": {
					"filename": "/var/log/access.log",
					"output": "file",
					"roll_keep": 5,
					"roll_keep_days": 90,
					"roll_size_mb": 954
				},
				"include": [
					"http.log.access.log0"
				]
			}
		}
	},
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":80"
					],
					"logs": {
						"default_logger_name": "log0"
					}
				}
			}
		}
	}
}
```

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -486,7 +486,7 @@ func Validate(cfg *Config) error {
 // Duration can be an integer or a string. An integer is
 // interpreted as nanoseconds. If a string, it is a Go
 // time.Duration value such as `300ms`, `1.5h`, or `2h45m`;
-// valid units are `ns`, `us`/`µs`, `ms`, `s`, `m`, and `h`.
+// valid units are `ns`, `us`/`µs`, `ms`, `s`, `m`, `h`, and `d`.
 type Duration time.Duration
 
 // UnmarshalJSON satisfies json.Unmarshaler.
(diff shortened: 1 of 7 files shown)


caddy.go
```diff
@@ -497,14 +497,29 @@ func (d *Duration) UnmarshalJSON(b []byte) error {
...
 	} else {
 		err = json.Unmarshal(b, &dur)
 	}
 	*d = Duration(dur)
 	return err
 }
 
+// ParseDuration parses a duration string, adding
+// support for the "d" suffix for number of days.
+func ParseDuration(s string) (time.Duration, error) {
```
mholt, line 510: This function will need tests (especially because it is so easy to test).

caddy.go
```diff
@@ -497,14 +497,29 @@ func (d *Duration) UnmarshalJSON(b []byte) error {
...
+	// Add support for days suffix (24 hours)
+	if strings.HasSuffix(s, "d") {
```
mholt, line 512: How does it handle `30d12h` or `12h30d`, etc? You'll have to find the "day" part in the string and parse it out, rather than assuming days are the only unit of the string.

francislavoie: Sounds like it's time to whip out https://regex101.com/ again :sweat_smile:

caddy.go
francislavoie: I did add an adapt test which indirectly tests it, but yeah I'll add a `caddy_test.go` file I guess which does more direct ones

caddy.go
mholt: Nooo, just iterate the string until you find "d" -- if at all. If you don't find it, just pass it to time.ParseDuration. If you do find it, extract the characters since the last non-letter (note that decimals can appear in numbers) and replace that segment with hours.

francislavoie: Fair enough. Still think it would be easier to do with a regexp rather than a memoized loop though.

How would it behave if `30d12h` -> `720h12h` :thinking:

caddy.go
mholt: I saw that -- thankful for that -- but we absolutely can unit test this sucker :)

caddy.go
mholt: That's nifty, I didn't know that! That should make things simpler then.

francislavoie: https://regex101.com/r/vQW0M4/1/ This is what I came up with. Loop through all the `days` matches, parse as float, then simple string replace the match value with `string(value * 24.0) + "h"`.

mholt: What about this? https://regex101.com/r/vQW0M4/2 -- isn't that simpler?

francislavoie:
I just did a rebase, cleaner commit history, easier to review IMO :sweat_smile: 

Thanks so much for finishing this one up, you're much better at parsing logic like that than I am :+1:

Issue #3314, opened 2020-04-26
caddyhttp: Implement placeholders in templates


bagbag:
(Feature request out of <https://caddy.community/t/use-placeholders-in-templates/7697>)

I would like to see the ability of using placeholders like `{http.error.status_code}` in templates. This would allow us, for example, to create a single custom error page in nginx style. For now we have to create a page for every status code we would like to support (maybe using `respond` with include and arguments would work too, but I haven't tested that).

Instead of adding an extra layer of complexity to substitute placeholders in templates, my suggestion is to either implement a new template function like `{{.Placerholder "http.error.status_code"}}` (with a better name than `.Placeholder`) or if variables in Go templates work the way I guess they work, populate them (if that makes sense, I have no idea about Go templates yet). But I think a function is the better solution.

When this is implemented, caddys template functions `.RemoteIP` and `.Host` could be deprecated and later removed, as the placeholders includes them with `http.request.remote.host` and `http.request.host`.

I guess this feature would fit for caddy v2.1. I have no idea what the release schedule for caddy looks like, but if it takes a few days or more from now until the release of 2.1, I would like to try to implement this feature, as it is probably quite easy to implement and therefore a good candidate for me to get started with Go and contribute to this great project (hints welcomed).

francislavoie:
> When this is implemented, caddys template functions `.RemoteIP` and `.Host` could be deprecated and later removed

I don't think this is necessary, shorter/easier APIs are beneficial.

mholt:
I assume you mean `.Req.RemoteIP` and `.Req.Host`?

We can't really get rid of those since they're on the standard lib's http.Request struct.

We could expose a way to access replacer values from the templates instead; it just involves getting a context value out of the request, type-asserting it, then calling Get() on the Replacer type.

francislavoie:
No he means `.RemoteIP` and `.Host` as documented here: https://caddyserver.com/docs/modules/http.handlers.templates

mholt:
Oh, I think that's pretty old docs. It'll be updated at 2.0.

mholt:
My bad -- those are actually different.

florence443:
Iam doing good

florence443:
Iam almost done

florence443:
Ok

florence443:
Its good

PR #3324, opened 2020-04-29
caddyhttp: Implement placeholders in template

bagbag:
Pull request for #3314

I don't think that I have to say anything about the code, as it is little and straight-forward.

I decided to go for an template-function, I haven't found a better name than `placeholder`, feel free to recommend another name. In my opinion, the braces should be omitted in the context of templates. 

I'm not sure what text to return if the placeholder does not exist. Copy the behaviour of the `env`-function or return the placeholder unchanged, as it is done for placeholders in Caddyfile. I decided to go for the first one.

I'm not sure whether `modules/caddyhttp/placeholdershorthands.go` is the correct directory for that file or not or if there is a better/preferred way in Go/this code-base to avoid defining the list of shorthands multiple times.

I'm not sure about the documentation. Is the addition in `modules/caddyhttp/templates/templates.go` all I need for it to appear in the docs?
How can I edit <https://caddyserver.com/docs/conventions> to add an exception for the braces?

#### How to test this pull request

Create the following files next to the caddy binary (or the working directory for the running process):

`Caddyfile`
```
http://localhost {
  file_server
  templates
}
```

`template.html`
```
{{placeholder "http.request.host"}} - {{placeholder "path"}}
```

Run caddy: `./caddy run --config Caddyfile --adapter caddyfile`
Navigate to the file: <http://localhost/template.html>


francislavoie:
FYI this implementation conflicts with https://github.com/caddyserver/caddy/pull/3305

I don't think it's right to expose the shorthands to the templates yet. I think that should be a separate change later.

bagbag:
Good to know, I reverted these changes. If this pull request is merged using squash, there shouldn't be a conflict I think(?). Otherwise I can rewrite the history.

francislavoie:
Nope, you should be good! Thanks!

FYI this may take a bit of time to get merged since we're still in a feature freeze until v2.0 stable is released.

francislavoie:
For bonus points, you could add some tests for this to `tplcontext_test.go` if you're up for it.

modules/caddyhttp/templates/templates.go
```diff
@@ -64,6 +64,16 @@ func init() {
 // {{env "VAR_NAME"}}
 // ```
 //
+// ##### `placeholder`
+//
+// Gets an [placerholder variable](/docs/conventions#placeholders). Shorthands are not supported at the moment.
```
mholt, line 69:
```suggestion
// Gets an [placeholder variable](/docs/conventions#placeholders).
```

Shorthand placeholders are a Caddyfile-only concept, doesn't make sense to document them in the JSON doc.

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -164,6 +166,18 @@ func (c templateContext) executeTemplateInBuffer(tplName string, buf *bytes.Buff
...
+	repl := c.Req.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
+
+	value, has := repl.Get(name)
+
+	if !has {
+		return ""
+	}
+
+	return fmt.Sprintf("%v", value)
+}
```
mholt, line 179:
```suggestion
	repl := c.Req.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
	value, _ := repl.GetString(name)
	return value
}
```

Might need to rebase with master -- I know this GetString() method was added recently but I can't recall when, maybe it's still in a PR... in any case, please try this out.

mholt:
Any interest in finishing this? If not, that's fine, I can close the PR so it's not hanging over our heads :)

bagbag:
Oh, I totally forgot that PR. Yes I will finish this.

bagbag:
Okay, so I commited your suggestions, rebased to master, built it, tested it (it still works) and forced pushed. Anything else missing for a merge?

mholt:
Great! At this point I'm just trying to decide if `placeholder` is the right/best function name for this. I'm mostly indifferent about this, but, at first glance it doesn't strike me quite the right way. I wonder if there's a better term.

bagbag:
I had similar thoughts when I started this PR. My best argument was to use the same terminology everywhere, eg. at https://caddyserver.com/docs/conventions#placeholders

When I searched for that feature, my query was "caddy placeholder in templates", as this was the term I knew from caddyfile.

mholt:
Fair enough, we can roll with it for now and then change it later if needed (templates are all kind of experimental anyway, as the docs say).

Thanks for working on this and seeing it through to completion!

PR #3335, opened 2020-05-03
httpcaddyfile: Add client certificate config to `tls`

### LINKED ISSUE #3334: Caddyfile: support tls option "clients" in tls config
Caddy v1 allowed it to configure client certificates in the tls configuration. Caddy v2 allows this just using the API. It would be nice to implement this options in the Caddyfile parser.

How it looked in Caddy v1: ( [source](https://caddyserver.com/v1/docs/tls) )
```
tls [cert key] {
    ca        uri
    protocols min max
    ciphers   ciphers...
    curves    curves...
    clients   [request|require|verify_if_given] clientcas...
    load      dir
    ask       url
    key_type  type
    dns       provider
    alpn      protos...
    must_staple
    wildcard
}
```
How it should look to cover all the options from the JSON structure of Caddy v2:
```
clients {
  mode [request|require|verify_if_given|require_and_verify]
  trusted_ca_certs <trusted_ca_certs...>
  trusted_leaf_certs <trusted_leaf_certs...>
}
```

nwhirschfeld:
Fixes #3334


francislavoie:
Could you add some tests (likely an adapt test, see the `caddytest` package)?

Could you also post an example Caddyfile and the adapted JSON output (using `caddy adapt`)?

Thanks for working on this!

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -59,6 +59,11 @@ func parseBind(h Helper) ([]ConfigValue, error) {
 //         protocols <min> [<max>]
 //         ciphers   <cipher_suites...>
 //         curves    <curves...>
+//         clients	 {
+//                     mode	              [request|require|verify_if_given|require_and_verify]
+//                     trusted_ca_certs   <trusted_ca_certs...>
+//                     trusted_leaf_certs <trusted_leaf_certs...>
+//					 }
```
francislavoie, line 66: Just some whitespace cleanup on this comment (tabs -> spaces, indentation fixes)

```suggestion
//         clients {
//             mode	              [request|require|verify_if_given|require_and_verify]
//             trusted_ca_certs   <trusted_ca_certs...>
//             trusted_leaf_certs <trusted_leaf_certs...>
//         }
```

nwhirschfeld:
wow, you're fast!

nwhirschfeld:
This is my Caddyfile and the example output:

nwhirschfeld:
Here is the Caddyfile I used for testing and the *caddy adapt* run:

caddytest/integration/caddyfile_adapt_test.go
```diff
@@ -283,3 +283,73 @@ func TestHttpOnlyOnNonStandardPort(t *testing.T) {
...
+                mode request 
+                trusted_ca_certs MIIDSzCCAjOgAwIBAgIUfIRObjWNUA4jxQ/0x8BOCvE2Vw4wDQYJKoZIhvcNAQELBQAwFjEUMBIGA1UEAwwLRWFzeS1SU0EgQ0EwHhcNMTkwODI4MTYyNTU5WhcNMjkwODI1MTYyNTU5WjAWMRQwEgYDVQQDDAtFYXN5LVJTQSBDQTCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAK5m5elxhQfMp/3aVJ4JnpN9PUSz6LlP6LePAPFU7gqohVVFVtDkChJAG3FNkNQNlieVTja/bgH9IcC6oKbROwdY1h0MvNV8AHHigvl03WuJD8g2ReVFXXwsnrPmKXCFzQyMI6TYk3m2gYrXsZOU1GLnfMRC3KAMRgE2F45twOs9hqG169YJ6mM2eQjzjCHWI6S2/iUYvYxRkCOlYUbLsMD/AhgAf1plzg6LPqNxtdlwxZnA0ytgkmhK67HtzJu0+ovUCsMv0RwcMhsEo9T8nyFAGt9XLZ63X5WpBCTUApaAUhnG0XnerjmUWb6eUWw4zev54sEfY5F3x002iQaW6cECAwEAAaOBkDCBjTAdBgNVHQ4EFgQU4CBUbZsS2GaNIkGRz/cBsD5ivjswUQYDVR0jBEowSIAU4CBUbZsS2GaNIkGRz/cBsD5ivjuhGqQYMBYxFDASBgNVBAMMC0Vhc3ktUlNBIENBghR8hE5uNY1QDiPFD/THwE4K8TZXDjAMBgNVHRMEBTADAQH/MAsGA1UdDwQEAwIBBjANBgkqhkiG9w0BAQsFAAOCAQEAKB3V4HIzoiO/Ch6WMj9bLJ2FGbpkMrcb/Eq01hT5zcfKD66lVS1MlK+cRL446Z2b2KDP1oFyVs+qmrmtdwrWgD+nfe2sBmmIHo9m9KygMkEOfG3MghGTEcS+0cTKEcoHYWYyOqQh6jnedXY8Cdm4GM1hAc9MiL3/sqV8YCVSLNnkoNysmr06/rZ0MCUZPGUtRmfd0heWhrfzAKw2HLgX+RAmpOE2MZqWcjvqKGyaRiaZks4nJkP6521aC2Lgp0HhCz1j8/uQ5ldoDszCnu/iro0NAsNtudTMD+YoLQxLqdleIh6CW+illc2VdXwj7mn6J04yns9jfE2jRjW/yTLFuQ==
```
mholt, line 295: Don't you think people using a Caddyfile would be more likely to have the certs in a PEM file than in base64-encoded DER format?

Maybe we should expect filepaths instead.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -59,6 +59,11 @@ func parseBind(h Helper) ([]ConfigValue, error) {
 //         protocols <min> [<max>]
 //         ciphers   <cipher_suites...>
 //         curves    <curves...>
+//         clients {
+//             mode	              [request|require|verify_if_given|require_and_verify]
```
mholt, line 63: Fix the comment alignment here

caddytest/integration/caddyfile_adapt_test.go
nwhirschfeld: I see your point. The idea was to "just build an adapter" to the JSON config model, without making any security related changes in this point. 
Do you think this should be implemented in a way, that the JSON structure also accepts filepaths or should the Caddyfile interpreter read the file?
For both solutions, should both formats, DER format, as well as filepath, be allowed in the same argument list?

mholt: > without making any security related changes in this point.

This isn't a "security-related" change; it's just changing how the input is read. Instead of decoding a huge base64 blob from a handwritten file, it'd be better to read a filename and then just load the PEM data from the file.

> should the Caddyfile interpreter read the file?

Probably this one. The Caddyfile adapter can load and parse the file.

> For both solutions, should both formats, DER format, as well as filepath, be allowed in the same argument list? 

No, let's keep it simple. Just a filename to a PEM file.

shouya: I personally using the full text in config is fine, after all, we already store hashed password etc in based64 encoded text, this "single-file-config" philosphy is pretty good for me. A single file is easy to config and move around. Another thing that might be useful to do is to implement a place holder expression to read from file and expand upon parsing.

shouya: > without making any security related changes in this point.

@nwhirschfeld might mean that with the file loading and admin api, an attacker may be able to exploit this functionality to read arbitrary files on the system.

mholt: @shouya

> I personally think that using this piece of full text PEM in config is fine. After all, we already store hashed password, cloudflare API token, etc in encoded text and this "single-file-config" philosphy is pretty good for me. A single file is easy to config and move around.

I like it too, but the thing is that nobody is asking for this, AFAIK. Nobody wants to open their existing PEM file, PEM-decode it, then re-encode it as base64, then paste it into their nice comfy-looking Caddyfile. Certificates are much longer than a password hash (even EC certs).

People usually just already have their PEM file on disk and simply write the path to that PEM file.

> One thing that might be useful to do is to implement a place holder expression to read from file and expand upon parsing, so the users can express their logic in the way the like.

That's an interesting idea (but a really bad one to be used inside a request pipeline, for performance reasons). So it could be easily misused...

> with the file loading and admin api, an attacker may be able to exploit this functionality to read arbitrary files on the system.

Well, once your system is popped, it's game over anyway. I don't think this has anything to do with that (unless there's a viable, practical proof of concept to show it).

francislavoie: > Nobody wants to open their existing PEM file, PEM-decode it, then re-encode it as base64

PEM is base64 already, none of those steps would be necessary btw. PEM is just base64-encoded DER, but with some ASCII armor before and after (the `-----` stuff)

mholt: Oh yeah. Forgot...

So I guess that's fine... hmm...

Should we have both?

nwhirschfeld: I think to add the possibility to import a certificate file using the preamble **file:<PATH>** and to support base64 encoded certificates in the same list is no problem.

mholt: Let's just follow Caddy convention and have two subdirectives instead:

```
trusted_ca_cert      ... # base64 PEM goes here, can be specified multiple times
trusted_ca_cert_file ... # PEM filepath goes here, can be specified multiple times
```

velosol: Sorry to butt in but I found this while looking to add a feature request for specifying a binary (DER) or base64-DER file (ideally with many certificates) and place it in either a caddyfile or a list in the JSON config for both `trusted_ca_cert` and `trusted_leaf_cert`.  I'm hoping this would be extended to `trusted_leaf_cert_file` similar to `trusted_ca_cert_file` in this same merge?

Please let me know if you'd like a separate issue opened for any part of the above!

mholt: @velosol This is a good point, we should double-check whether this is only for CAs or whether we should just trust any cert in the chain that matches.

velosol: Ah! I misunderstood the purpose of 'leaf' vs 'ca' (I had thought leaf were intermediate CAs issued by a 'ca', not _client_ certs that would be accepted).  Is there documentation on how these interact (i.e. leaf must be issued by one of the 'ca's)?

caddyconfig/httpcaddyfile/builtins.go
nwhirschfeld: no problem, but which style do you prefer? I currently use something like a.
```
aaa [aaa aaa aaa] {
    aaaaaaa <aaa>
    aaaaa   <aaa...>
    aaaa {
        aaaaaa <aaaaa>
        aaa    <aaaa>
    }
}

bbb [bbb bbb bbb] {
    aaaaaaa <bbb>
    bbbbb   <bbb...>
    bbbb {
            bbbbbb <bbbbb>
            bbb    <bbbb>
    }
}


ccc [ccc ccc ccc] {
    aaaaaaa    <ccc>
    ccccc      <ccc...>
    cccc {
        cccccc <ccccc>
        ccc    <cccc>
    }
}
```

modules/caddytls/connpolicy.go
```diff
@@ -410,7 +410,7 @@ func (clientauth ClientAuthentication) verifyPeerCertificate(rawCerts [][]byte,
 }
 
 // decodeBase64DERCert base64-decodes, then DER-decodes, certStr.
```
mholt, line 412:
```suggestion
// DecodeBase64DERCert base64-decodes, then DER-decodes, certStr.
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+					for ch.NextBlock(0) {
+						switch ch.Val() {
+						case "mode":
+							for ch.NextArg() {
+								cp.ClientAuthentication.Mode = ch.Val()
+							}
+						case "trusted_ca_certs":
+							for ch.NextArg() {
+								if _, err := caddytls.DecodeBase64DERCert(ch.Val()); err != nil {
+									return nil, h.Errf("cannot decode trusted CA certificate '%s'", ch.Val())
```
mholt, line 207: Instead of printing the entire base64 input, print the error message here instead.

mholt: On second thought, this decoding step is not necessary. It will be decoded at Provision-time and so there is no need to do the check here.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+				ch := h.NewFromNextSegment() // new helper for client auth config
+				cp.ClientAuthentication = &caddytls.ClientAuthentication{}
+				for ch.Next() {
+					for ch.NextBlock(0) {
+						switch ch.Val() {
+						case "mode":
+							for ch.NextArg() {
+								cp.ClientAuthentication.Mode = ch.Val()
+							}
+						case "trusted_ca_certs":
```
mholt, line 204: Let's rename this to `trusted_ca_cert` since it is most likely that the user will only specify one, and they can specify it multiple times too if they need to -- as opposed to the JSON field which is a slice.
```suggestion
						case "trusted_ca_cert":
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+							for ch.NextArg() {
+								if _, err := caddytls.DecodeBase64DERCert(ch.Val()); err != nil {
+									return nil, h.Errf("cannot decode trusted CA certificate '%s'", ch.Val())
+								}
+								cp.ClientAuthentication.TrustedCACerts = append(cp.ClientAuthentication.TrustedCACerts, ch.Val())
+							}
+						case "trusted_ca_certs_file":
```
mholt, line 211: This rename also flows better with the `_file` variants.
```suggestion
						case "trusted_ca_cert_file":
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+									return nil, err
+								}
+								// safe repack PEM file
+								block, _ := pem.Decode(certDataPEM)
+								if block == nil || block.Type != "CERTIFICATE" {
+									return nil, h.Errf("cannot decode trusted CA certificate file'%s'", ch.Val())
+								}
+								cp.ClientAuthentication.TrustedCACerts = append(cp.ClientAuthentication.TrustedCACerts, base64.StdEncoding.EncodeToString(block.Bytes))
+							}
+						case "trusted_leaf_certs":
```
mholt, line 226:
```suggestion
						case "trusted_leaf_cert":
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+							for ch.NextArg() {
+								if _, err := caddytls.DecodeBase64DERCert(ch.Val()); err != nil {
+									return nil, h.Errf("cannot decode trusted leaf certificate '%s'", ch.Val())
+								}
+								cp.ClientAuthentication.TrustedLeafCerts = append(cp.ClientAuthentication.TrustedLeafCerts, ch.Val())
+							}
+						case "trusted_leaf_certs_file":
```
mholt, line 233:
```suggestion
						case "trusted_leaf_cert_file":
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+								if block == nil || block.Type != "CERTIFICATE" {
+									return nil, h.Errf("cannot decode trusted CA certificate file'%s'", ch.Val())
+								}
+								cp.ClientAuthentication.TrustedCACerts = append(cp.ClientAuthentication.TrustedCACerts, base64.StdEncoding.EncodeToString(block.Bytes))
+							}
+						case "trusted_leaf_certs":
+							for ch.NextArg() {
+								if _, err := caddytls.DecodeBase64DERCert(ch.Val()); err != nil {
+									return nil, h.Errf("cannot decode trusted leaf certificate '%s'", ch.Val())
+								}
```
mholt, line 230: As above, this check is not necessary here.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+								certDataPEM, err := ioutil.ReadFile(filename)
+								if err != nil {
+									return nil, err
+								}
+								// safe repack PEM file
+								block, _ := pem.Decode(certDataPEM)
+								if block == nil || block.Type != "CERTIFICATE" {
+									return nil, h.Errf("cannot decode trusted CA certificate file'%s'", ch.Val())
+								}
+								cp.ClientAuthentication.TrustedCACerts = append(cp.ClientAuthentication.TrustedCACerts, base64.StdEncoding.EncodeToString(block.Bytes))
```
mholt, line 224: I think PEM blocks are just base64-encoded DER already, aren't they? Might not have to decode it; try just using block.Bytes directly and see if it works!

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+								certDataPEM, err := ioutil.ReadFile(filename)
+								if err != nil {
+									return nil, err
+								}
+								// safe repack PEM file
+								block, _ := pem.Decode(certDataPEM)
+								if block == nil || block.Type != "CERTIFICATE" {
+									return nil, h.Errf("cannot decode trusted leaf certificate file'%s'", ch.Val())
+								}
+								cp.ClientAuthentication.TrustedLeafCerts = append(cp.ClientAuthentication.TrustedLeafCerts, base64.StdEncoding.EncodeToString(block.Bytes))
```
mholt, line 246: Same here as above, might not need to decode the PEM blocks at all.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -181,6 +191,66 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 					cp.Curves = append(cp.Curves, h.Val())
 				}
 
+			case "clients":
```
mholt, line 194:
```suggestion
			case "client_auth":
```

Makes more sense I think.

modules/caddytls/connpolicy.go
```diff
@@ -409,8 +409,8 @@ func (clientauth ClientAuthentication) verifyPeerCertificate(rawCerts [][]byte,
 	return fmt.Errorf("client leaf certificate failed validation")
 }
 
-// decodeBase64DERCert base64-decodes, then DER-decodes, certStr.
-func decodeBase64DERCert(certStr string) (*x509.Certificate, error) {
+// DecodeBase64DERCert base64-decodes, then DER-decodes, certStr.
+func DecodeBase64DERCert(certStr string) (*x509.Certificate, error) {
```
mholt, line 413: Since we don't need to double-verify the input, I don't think this needs to be exported anymore.

PR #3344, opened 2020-05-04
tls/client auth: verify all certificates in client request

KarolBedkowski:
Hi,
I have problem with certificate-based authentication. In my env Caddy can't authorized connection despite  correct certificates and configuration. After some debugging I notice that Caddy check only last certificate in request but client certificate is fist and second is CA cert.
This patch change this behavior - all certs from request are verified.

--------

When client certificate is enabled Caddy check only last certificate from
request. When this cert is not in list of trusted leaf certificates,
connection is rejected. But in some cases client certificate is not
last certificate in list; sometimes is first, before certificates of CA.

This patch add checking all sent certificates - if any match with trusted
certificates connection is accepted.

### DIFF
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -382,7 +382,7 @@ func (clientauth ClientAuthentication) verifyPeerCertificate(rawCerts [][]byte, 
 		return fmt.Errorf("no client certificate provided")
 	}
 
-	remoteLeafCert, err := x509.ParseCertificate(rawCerts[len(rawCerts)-1])
+	remoteLeafCert, err := x509.ParseCertificate(rawCerts[0])
 	if err != nil {
 		return fmt.Errorf("can't parse the given certificate: %s", err.Error())
 	}


francislavoie:
Thanks! Looks pretty simple. We definitely want to be careful here because it affects security.

Don't forget to sign the CLA!

mholt:
It does look like we got it backwards; the spec ([RFC 5246](https://tools.ietf.org/html/rfc5246#section-7.4.2)) requires:

```
   certificate_list
      This is a sequence (chain) of certificates.  The sender's
      certificate MUST come first in the list.  Each following
      certificate MUST directly certify the one preceding it.  Because
      certificate validation requires that root keys be distributed
      independently, the self-signed certificate that specifies the root
      certificate authority MAY be omitted from the chain, under the
      assumption that the remote end must already possess it in order to
      validate it in any case.
```

In my own testing I've only used one-cert chains, so I never encountered this.

I'm also told that TLS 1.3 abandoned the strict ordering of certificates other than the first one.

I will review this soon!

In the meantime please sign the CLA @KarolBedkowski . Thanks!

KarolBedkowski:
Yes, according to specification checking first certificate should be enough. 
Thanks.

PR #3325, opened 2020-04-30
goreleaser: add s390x and ppc64le builds for linux

grooverdan:
dan@fstn4-p1:~/go/src/github.com/caddyserver/caddy/cmd/caddy$ file ./caddy 
./caddy: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), dynamically linked, interpreter /lib64/l, for GNU/Linux 3.10.0, BuildID[sha1]=1fc956e7b143268aec43dd6850e6af65562ed412, not stripped

dan@fstn4-p1:~/go/src/github.com/caddyserver/caddy/cmd/caddy$ ./caddy version
(devel)

dan@fstn4-p1:~/go/src/github.com/caddyserver/caddy/cmd/caddy$ ./caddy build-info
path: github.com/caddyserver/caddy/v2/cmd/caddy
main: github.com/caddyserver/caddy/v2 (devel) 
dependencies:
cloud.google.com/go v0.54.0 h1:3ithwDMr7/3vpAMXiH+ZQnYbuIsh+OPhUPMFC9enmn0=
...

Builds and runs seamlessly here (ppc64le, haven't tested s390x).

PR #3355, opened 2020-05-06
Fixing goreleaser syntax error

hairyhenderson:
Fixing syntax error introduced in #3325, and adding a `goreleaser check` command so we can avoid these in the future 😉

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,10 +4,10 @@ name: Cross-Platform
 
 on:
   push:
-    branches: 
+    branches:
       - master
   pull_request:
-    branches: 
+    branches:
       - master
 
 jobs:
@@ -133,3 +133,15 @@ jobs:
         # uses: docker://reviewdog/action-golangci-lint:v1 # pre-build docker image
         with:
           github_token: ${{ secrets.github_token }}
+
+  goreleaser-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - uses: goreleaser/goreleaser-action@v1
+        with:
+          version: latest
+          args: check
+        env:
+          TAG: ${{ steps.vars.outputs.version_tag }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -34,13 +34,18 @@ builds:
   ignore:
     - goos: darwin
       goarch: arm
-    - goarch:
-      - ppc64le
-      - s390x
-      goos: 
-        - darwin
-        - windows
-        - freebsd
+    - goos: darwin
+      goarch: ppc64le
+    - goos: darwin
+      goarch: s390x
+    - goos: windows
+      goarch: ppc64le
+    - goos: windows
+      goarch: s390x
+    - goos: freebsd
+      goarch: ppc64le
+    - goos: freebsd
+      goarch: s390x
   flags:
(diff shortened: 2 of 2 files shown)


hairyhenderson:
PPC build timed out? That'd be unrelated to this change I think... Can it just be re-triggered?

Mohammed90:
Yeah, it does timeout more often than desirable. Just (force-)push a commit. I don't see any buttons to retrigger.

francislavoie:
@grooverdan FYI the TravisCI tests seem pretty flaky. Is there a chance you could provide hardware we could use permanently to test on (maybe as an external gitlab actions server)?

grooverdan:
Ok, looking at the travis history I see two important failures, both s390x:

https://travis-ci.org/github/caddyserver/caddy/jobs/682382430
https://travis-ci.org/github/caddyserver/caddy/jobs/683516295

@hairyhenderson I assume the ppc timeout was actually the s390x one. If not, do you have a URL.

I've raised these with an internal team to look at/escalate to travis (which probably could be done directly).

The other failures look to be commits that where removed by the time the CI got to them or initial config problems.

In the mean time for travis, while no where close to idea [allow failures](https://docs.travis-ci.com/user/customizing-the-build/#jobs-that-are-allowed-to-fail) for `arch: s390x`

On permanent access to VMs, contact Elizabeth (contact at bottom of [this page](https://community.ibm.com/community/user/ibmz-and-linuxone/blogs/elizabeth-k-joseph1/2020/02/14/linuxone-open-source-software-report-january-2020?CommunityKey=87042487-940e-49cc-ae6d-5a5b76c5ea7e&Tab=).

I looked at [github runners](https://github.com/caddyserver/caddy/pull/3328#issuecomment-622289165) and their agents are C#/.Net based and appears to be impossible/excessive effort to port to s390x.

I've been looking through [other options](https://ligurio.github.io/awesome-ci/) with GH hooks and haven't found any that are largely a cloud service where you can add your deploy your own executor/agent. Ask Elizabeth for suggestions.

francislavoie:
Here's one where ppc timed out: https://travis-ci.org/github/caddyserver/caddy/builds/683625139

francislavoie:
@grooverdan the idea was to install a self-hosted runner for github actions, see https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

Thanks for that contact link, that looks ideal! We'll look into it.

grooverdan:
Thanks for the ppc timeout. I'll get that reported to travis, saving cache shouldn't timeout.

On self-hosted github actions, it requires this agent runner - https://github.com/actions/runner, which I couldn't build on s390x.

francislavoie:
> On self-hosted github actions, it requires this agent runner - actions/runner, which I couldn't build on s390x.

Oh, that's a bummer :cry:

grooverdan:
Yeh I know :sad:, when you get to the point your asking yourself if you can write a JIT compiler for a dotnet runtime you need to step back. Also tried getting a qemu tcg emulator of the runner going. I found the azure runners and they look similar technology wise.

On travis failures, reported:
- https://travis-ci.community/t/s390x-timeout-got-as-far-as-build-system-information/8490
- https://travis-ci.community/t/s390x-updating-gimme-failed-to-run-gimme/8489
- https://travis-ci.community/t/ppc64le-store-build-cache-causing-output-timeout/8488

PR #3372, opened 2020-05-07
caddycmd: check file exists before validate

josedonizetti:
Calling `validate` if the config file doesn't exist has a misleading error:

```
./caddy validate
validate: decoding config: unexpected end of JSON input
```

This PR adds a validation to output a better error

```
./caddy validate
validate: file doesn't exist: stat Caddyfile: no such file or directory
```

The check is done on `cmdValidate` because `loadConfig` shouldn't error if the file doesn't exist (https://github.com/caddyserver/caddy/blob/master/cmd/main.go#L139).  The func `loadConfig` is used by `run` and `start`,  if the config file doesn't exist, it will boot with blank config.

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -506,6 +506,15 @@ func cmdValidateConfig(fl Flags) (int, error) {
 	validateCmdConfigFlag := fl.String("config")
 	validateCmdAdapterFlag := fl.String("adapter")
 
+	if validateCmdConfigFlag == "" {
+		validateCmdConfigFlag = "Caddyfile"
+	}
+	_, err := os.Stat(validateCmdConfigFlag)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup,
+			fmt.Errorf("file doesn't exist: %v", err)
+	}
+
 	input, _, err := loadConfig(validateCmdConfigFlag, validateCmdAdapterFlag)
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, err


cmd/commandfuncs.go
```diff
@@ -506,6 +506,15 @@ func cmdValidateConfig(fl Flags) (int, error) {
 	validateCmdConfigFlag := fl.String("config")
 	validateCmdAdapterFlag := fl.String("adapter")
 
+	if validateCmdConfigFlag == "" {
+		validateCmdConfigFlag = "Caddyfile"
+	}
+	_, err := os.Stat(validateCmdConfigFlag)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup,
+			fmt.Errorf("file doesn't exist: %v", err)
```
mholt, line 515: We don't know that the file doesn't exist. It could be any number of errors. I would just return the error.

```suggestion
		return caddy.ExitCodeFailedStartup, err
```

cmd/commandfuncs.go
```diff
@@ -506,6 +506,15 @@ func cmdValidateConfig(fl Flags) (int, error) {
 	validateCmdConfigFlag := fl.String("config")
 	validateCmdAdapterFlag := fl.String("adapter")
 
+	if validateCmdConfigFlag == "" {
+		validateCmdConfigFlag = "Caddyfile"
+	}
```
mholt, line 511: `loadConfig()` already does this.

cmd/commandfuncs.go
josedonizetti: Agreed. It could be a permission error.

cmd/commandfuncs.go
josedonizetti: @mholt doing it because on the next line we do the `os.Stat` validation. Didn't think we want to run an `os.Stat` for a empty string. wdty?

mholt: We shouldn't be duplicating the logic that is in `loadConfig()`.

josedonizetti: @mholt Not sure if got it. Do prefer to avoid this change at all? And you okay with the current return from `loadConfig`? 

Or you okay with the change, but doesn't think we should set the `validateCmdConfigFlag` here?

Thanks

mholt: The `loadConfig()` function already accounts for whether a config file was specified, or if a config file is not found (as you noted in your OP). This code steps on that logic, since now there are two places where we assume a default config file instead of one.

The change should be made in the one place that an explicitly-defined config file is loaded and, if it is not found, then return an error.

josedonizetti: @mholt gotcha! let me see if I can fit this in a better way.

mholt:
@josedonizetti Why did you close this? Is it no longer needed/wanted?

Issue #3365, opened 2020-05-07
Add support in hash-password for reading from terminals/stdin


0az:
I have an initial implementation in my fork, in the improve-hash-password branch.

We only read from stdin if plaintext is not specified via argument. If stdin doesn't belong to a controlling terminal, we just hash everything up to the first newline. Otherwise, we ask the user to confirm the entered plaintext. This matches htpasswd's behavior, or at least the version I have access to.

0az:
The implementing commit: https://github.com/0az/caddy/commit/c5bafbb75c092f7c368d3323d44886ae18038e95.

francislavoie:
Thanks, that's interesting!

Could you make a pull request with that change?

PR #3373, opened 2020-05-07
http: Add support in hash-password for reading from terminals/stdin (#3365)

0az:
Implements #3365.

### DIFF
--- a/modules/caddyhttp/caddyauth/command.go
+++ b/modules/caddyhttp/caddyauth/command.go
@@ -71,7 +71,6 @@ func cmdHashPassword(fs caddycmd.Flags) (int, error) {
 			fmt.Print("Enter password: ")
 			plaintext, err = terminal.ReadPassword(int(os.Stdin.Fd()))
 			fmt.Println()
-
 			if err != nil {
 				return caddy.ExitCodeFailedStartup, err
 			}
@@ -79,7 +78,6 @@ func cmdHashPassword(fs caddycmd.Flags) (int, error) {
 			fmt.Print("Confirm password: ")
 			confirmation, err := terminal.ReadPassword(int(os.Stdin.Fd()))
 			fmt.Println()
-
 			if err != nil {
 				return caddy.ExitCodeFailedStartup, err
 			}
@@ -90,7 +88,6 @@ func cmdHashPassword(fs caddycmd.Flags) (int, error) {
 		} else {
 			rd := bufio.NewReader(os.Stdin)
 			plaintext, err = rd.ReadBytes('\n')
-
 			if err != nil {
 				return caddy.ExitCodeFailedStartup, err
 			}
--- a/modules/caddyhttp/caddyauth/command.go
+++ b/modules/caddyhttp/caddyauth/command.go
@@ -15,25 +15,33 @@
 package caddyauth
 
 import (
+	"bufio"
+	"bytes"
 	"encoding/base64"
 	"flag"
 	"fmt"
+	"os"
 
 	"github.com/caddyserver/caddy/v2"
 	caddycmd "github.com/caddyserver/caddy/v2/cmd"
 	"golang.org/x/crypto/bcrypt"
 	"golang.org/x/crypto/scrypt"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 func init() {
 	caddycmd.RegisterCommand(caddycmd.Command{
 		Name:  "hash-password",
 		Func:  cmdHashPassword,
-		Usage: "--plaintext <password> [--salt <string>] [--algorithm <name>]",
+		Usage: "[--algorithm <name>] [--salt <string>] [--plaintext <password>]",
 		Short: "Hashes a password and writes base64",
 		Long: `
 Convenient way to hash a plaintext password. The resulting
 hash is written to stdout as a base64 string.
+
+--plaintext, when omitted, will be read from stdin. If
+Caddy is attached to a controlling tty, the plaintext will
+not be echoed.
 
 --algorithm may be bcrypt or scrypt. If script, the default
 parameters are used.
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,43 @@ be provided (scrypt).
...
+			plaintext, err = terminal.ReadPassword(int(os.Stdin.Fd()))
+			fmt.Println()
+
+			fmt.Print("Confirm Plaintext: ")
+			confirmation, err = terminal.ReadPassword(int(os.Stdin.Fd()))
+			fmt.Println()
+
+			if !bytes.Equal(plaintext, confirmation) {
+				return caddy.ExitCodeFailedStartup, fmt.Errorf("plaintext does not match")
+			}
```
0az, line 83: What's a good, idiomatic way to handle these errors?

francislavoie: I think just
```go
if err != nil {
}
```

mholt: There's no error for this though; it needs to be made.

How about `fmt.Errorf("passwords do not match")`

modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,43 @@ be provided (scrypt).
...
+	var err error
+
 	algorithm := fs.String("algorithm")
 	plaintext := []byte(fs.String("plaintext"))
 	salt := []byte(fs.String("salt"))
 
 	if len(plaintext) == 0 {
-		return caddy.ExitCodeFailedStartup, fmt.Errorf("password is required")
+		if terminal.IsTerminal(int(os.Stdin.Fd())) {
+			var confirmation []byte
```
mholt, line 71: This can be declared on the line before it is assigned. (This isn't C! :wink:)

modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,43 @@ be provided (scrypt).
...
+			plaintext = plaintext[:len(plaintext)-1] // Trailing newline
+		}
+
+		if len(plaintext) == 0 {
+			return caddy.ExitCodeFailedStartup, fmt.Errorf("plaintext is required")
+		}
+
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, err
+		}
```
mholt, line 96: Error check should come first

modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,43 @@ be provided (scrypt).
...
 	algorithm := fs.String("algorithm")
 	plaintext := []byte(fs.String("plaintext"))
 	salt := []byte(fs.String("salt"))
 
 	if len(plaintext) == 0 {
-		return caddy.ExitCodeFailedStartup, fmt.Errorf("password is required")
+		if terminal.IsTerminal(int(os.Stdin.Fd())) {
+			var confirmation []byte
+
+			fmt.Print("Plaintext: ")
```
mholt, line 73:
```suggestion
			fmt.Print("Enter password: ")
```

modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,43 @@ be provided (scrypt).
...
+			plaintext, err = terminal.ReadPassword(int(os.Stdin.Fd()))
+			fmt.Println()
+
+			fmt.Print("Confirm Plaintext: ")
```
mholt, line 77:
```suggestion
			fmt.Print("Confirm password: ")
```

modules/caddyhttp/caddyauth/command.go
0az: > There's no error for this though; it needs to be made.

Specifically, in the `ReadPassword` call. I guess I should just move/copy the error block from the bottom.

> How about `fmt.Errorf("passwords do not match")`

I was leaning towards `plaintext` for consistency with the CLI arg.

modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,50 @@ be provided (scrypt).
...
+			confirmation, err := terminal.ReadPassword(int(os.Stdin.Fd()))
+			fmt.Println()
+
+			if err != nil {
+				return caddy.ExitCodeFailedStartup, err
+			}
+
+			if !bytes.Equal(plaintext, confirmation) {
+				return caddy.ExitCodeFailedStartup, fmt.Errorf("plaintext does not match")
```
0az, line 88:
```suggestion
				return caddy.ExitCodeFailedStartup, fmt.Errorf("password does not match")
```

modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,50 @@ be provided (scrypt).
...
 	if len(plaintext) == 0 {
-		return caddy.ExitCodeFailedStartup, fmt.Errorf("password is required")
+		if terminal.IsTerminal(int(os.Stdin.Fd())) {
+			fmt.Print("Enter password: ")
+			plaintext, err = terminal.ReadPassword(int(os.Stdin.Fd()))
+			fmt.Println()
+
+			if err != nil {
+				return caddy.ExitCodeFailedStartup, err
+			}
```
mholt, line 77: Error check should come immediately after the line that produces the error, for better readability and maintainability (same below).

Unless, does the terminal not go to the next line _at all_ without the `Println()`? Or is the Println adding an _extra_ (empty) line?

modules/caddyhttp/caddyauth/command.go
```diff
@@ -52,16 +60,50 @@ be provided (scrypt).
...
+
+			if !bytes.Equal(plaintext, confirmation) {
+				return caddy.ExitCodeFailedStartup, fmt.Errorf("password does not match")
+			}
+		} else {
+			rd := bufio.NewReader(os.Stdin)
+			plaintext, err = rd.ReadBytes('\n')
+
+			if err != nil {
```
mholt, line 94: Close this gap -- the error check is very tied to the ReadBytes line, not a different "phrase" or thought.

modules/caddyhttp/caddyauth/command.go
0az: I tested it, and ReadPassword does not preserve the newline.

PR #3378, opened 2020-05-08
adding wildcard matching of logger names

### LINKED ISSUE #3375: v2: Wildcard subdomains defined in Caddyfile not logging correctly
When defining some access logs, it seems like domains with wildcards aren't caught for some reason. I'm trying to redirect all of the subdomains that aren't defined into a different log for parsing later. Any of the hardcoded subdomains can get pushed into a standard log file just fine, but anything else seems to only appear in the stdout of the caddy process. I'm using a service to run caddy. I've included a minimal Caddyfile and the output I'm seeing.

Of note, the permissions on the directory are such that it is owned by the user & group caddy, and I'm able to login as caddy and write to wildcard.log manually.

# Caddyfile
```
{
    http_port 80
}

http://domain.com {
    log {
        output file /var/log/caddy/caddy.log
    }
    respond "root" 200
}

http://*.domain.com {
    log {
        output file /var/log/caddy/wildcard.log
    }
    respond "wildcard" 200
}
```
# Autosaved Config (/var/lib/caddy/.config/caddy/autosave.json)
```
{
  "apps": {
    "http": {
      "http_port": 80,
      "servers": {
        "srv0": {
          "automatic_https": {
            "skip": [
              "*.domain.com",
              "domain.com"
            ]
          },
          "listen": [
            ":80"
          ],
          "logs": {
            "logger_names": {
              "*.domain.com": "log1",
              "domain.com": "log0"
            }
          },
          "routes": [
            {
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "wildcard",
                          "handler": "static_response",
                          "status_code": 200
                        }
                      ]
                    }
                  ]
                }
              ],
              "match": [
                {
                  "host": [
                    "*.domain.com"
                  ]
                }
              ],
              "terminal": true
            },
            {
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "root",
                          "handler": "static_response",
                          "status_code": 200
                        }
                      ]
                    }
                  ]
                }
              ],
              "match": [
                {
                  "host": [
                    "domain.com"
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    }
  },
  "logging": {
    "logs": {
      "default": {
        "exclude": [
          "http.log.access.log1",
          "http.log.access.log0"
        ]
      },
      "log0": {
        "include": [
          "http.log.access.log0"
        ],
        "writer": {
          "filename": "/var/log/caddy/caddy.log",
          "output": "file"
        }
      },
      "log1": {
        "include": [
          "http.log.access.log1"
        ],
        "writer": {
          "filename": "/var/log/caddy/wildcard.log",
          "output": "file"
        }
      }
    }
  }
}
```

# Example of root domain
```sh
$ curl -i -H 'Host: domain.com' -k http://192.168.1.10/
HTTP/1.1 200 OK
Server: Caddy
Date: Fri, 08 May 2020 15:45:02 GMT
Content-Length: 4

root
```
```sh
$ cat /var/log/caddy/caddy.log
{"level":"info","ts":1588953466.9726162,"logger":"http.log.access.log0","msg":"handled request","request":{"method":"GET","uri":"/","proto":"HTTP/1.1","remote_addr":"192.168.1.50:50296","host":"domain.com","headers":{"User-Agent":["curl/7.68.0"],"Accept":["*/*"]}},"common_log":"192.168.1.50 - - [08/May/2020:10:57:46 -0500] \"GET / HTTP/1.1\" 200 4","duration":0.000098646,"size":4,"status":200,"resp_headers":{"Server":["Caddy"],"Content-Type":[]}}
```

# Example of wildcard domain
```sh
$ curl -i -H 'Host: asdf.domain.com' -k http://192.168.1.10/
HTTP/1.1 200 OK
Server: Caddy
Date: Fri, 08 May 2020 16:00:07 GMT
Content-Length: 8

wildcard
```
```sh
$ cat /var/log/caddy/wildcard.log
cat: /var/log/caddy/wildcard.log: No such file or directory
```
```sh
$ cat /etc/systemd/system/caddy.service
[Unit]
Description=Caddy
After=network.target

[Service]
User=caddy
Group=caddy
ExecStart=/usr/bin/caddy run --environ --config /caddy/Caddyfile
ExecReload=/usr/bin/caddy reload --config /caddy/Caddyfile
TimeoutStopSec=5s
LimitNOFILE=1048576
LimitNPROC=512
PrivateTmp=false
ProtectSystem=full
AmbientCapabilities=CAP_NET_BIND_SERVICE

[Install]
WantedBy=multi-user.target
```
```sh
$ sudo journalctl -u caddy
...
May 08 11:00:55 hostname caddy[9569]: {"level":"info","ts":1588953655.81634,"logger":"http.log.access","msg":"handled request","request":{"method":"GET","uri":"/","proto":"HTTP/1.1","remote_addr":"192.168.1.50:50224","host":"asdf.domain.com","headers":{"User-Agent":["curl/7.68.0"],"Accept":["*/*"]}},"common_log":"192.168.1.50 - - [08/May/2020:11:00:55 -0500] \"GET / HTTP/1.1\" 200 8","duration":0.00008724,"size":8,"status":200,"resp_headers":{"Content-Type":[],"Server":["Caddy"]}}
```

GregoryDosh:
This should fix #3375. Please let me know if you would like any other changes, or how I can add some tests to this. I didn't see a way to run any tests, but based on my own server it looks like this enables the functionality I was looking for.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
+	"github.com/caddyserver/certmagic"
 	"github.com/lucas-clemente/quic-go/http3"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
@@ -464,8 +465,10 @@ func (slc ServerLogConfig) wrapLogger(logger *zap.Logger, host string) *zap.Logg
 }
 
 func (slc ServerLogConfig) getLoggerName(host string) string {
-	if loggerName, ok := slc.LoggerNames[host]; ok {
-		return loggerName
+	for potentialHost, loggerName := range slc.LoggerNames {
+		if certmagic.MatchWildcard(host, potentialHost) {
+			return loggerName
+		}
 	}
 	return slc.DefaultLoggerName
 }
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -465,6 +465,10 @@ func (slc ServerLogConfig) wrapLogger(logger *zap.Logger, host string) *zap.Logg
 }
 
 func (slc ServerLogConfig) getLoggerName(host string) string {
+	if loggerName, ok := slc.LoggerNames[host]; ok {
+		return loggerName
+	}
+	// Try matching wildcard domains if other non-specific loggers exist
 	for potentialHost, loggerName := range slc.LoggerNames {
 		if certmagic.MatchWildcard(host, potentialHost) {
 			return loggerName
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
-	"github.com/caddyserver/certmagic"
 	"github.com/lucas-clemente/quic-go/http3"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
@@ -468,12 +467,20 @@ func (slc ServerLogConfig) getLoggerName(host string) string {
 	if loggerName, ok := slc.LoggerNames[host]; ok {
 		return loggerName
 	}
+
 	// Try matching wildcard domains if other non-specific loggers exist
-	for potentialHost, loggerName := range slc.LoggerNames {
-		if certmagic.MatchWildcard(host, potentialHost) {
+	labels := strings.Split(host, ".")
+	for i := range labels {
+		if labels[i] == "" {
+			continue
+		}
+		labels[i] = "*"
+		wildcardHost := strings.Join(labels, ".")
+		if loggerName, ok := slc.LoggerNames[wildcardHost]; ok {
 			return loggerName
 		}
 	}
+
 	return slc.DefaultLoggerName
 }
 


modules/caddyhttp/server.go
```diff
@@ -467,6 +468,12 @@ func (slc ServerLogConfig) getLoggerName(host string) string {
 	if loggerName, ok := slc.LoggerNames[host]; ok {
 		return loggerName
 	}
+	// Try matching wildcard domains if other non-specific loggers exist
+	for potentialHost, loggerName := range slc.LoggerNames {
+		if certmagic.MatchWildcard(host, potentialHost) {
```
mholt, line 473: This extra loop can be avoided if we borrow some of the logic from the `MatchWildcard()` function. If we split `host` by its labels and then replace each next-left-most consecutive label with a `*` all we have to do is the map lookup. Does that make sense?

Not a big deal, but an optimization that should reduce the complexity a notch.

GregoryDosh: I *think* I understand. So if my list of loggers is `domain.com`, `subdomain1.domain.com`, and `*.domain.com` and I'm looking up `something-else.domain.com` I can replace the leftmost label `something-else` with `*` and try the map lookup again with `*.domain.com`?

mholt: Yep! It'll somewhat resemble `MatchWildcard` I think. (See its implementation.)

GregoryDosh: Yep I think I get it. I just haven't ever had a use case (at home) where I was doing wildcard as segments of a subdomain like `ads.*.domain.com` and the like.

PR #3380, opened 2020-05-08
docs: Fix TOC/section header mismatch

ChandlerSwift:
Table of Contents says "Get Started"; Section Header says "Quick Start". Changed table of contents to align with section header.


francislavoie:
Thanks!

Issue #3386, opened 2020-05-10
templates and SplitFrontMatter do not play nice with windows newline characters 


oganga:
When on Windows and creating template and Markdown files with native Windows newline the SplitFrontMatter work in an unexpected way.

The Frontmatter is handled correctly due to the extra control character in the Windows newline

Issue #3191, opened 2020-03-25
Fix race/inconsistent results in integration tests


mholt:
Opening an issue to track the inconsistent results of integration tests, especially on Windows, in our CI workflow. /cc @sarge

It's not a data race per-se, but does appear to fail sometimes or another.

mholt:
(Also, if a config fails to load, the test should not continue as if the loading succeeded, which will actually use the config from the previous test.)

sarge:
Just an update on this, I would need a windows machine to load the https://hub.docker.com/_/microsoft-windows-servercore as the os of the container and the os of the docker image must match.

francislavoie:
@sarge Amazon does have Windows images if you're willing to try that, shouldn't be too expensive. You can RDP into them and mess around.

francislavoie:
I think this is solved by https://github.com/caddyserver/caddy/pull/3966 /cc @Mohammed90

PR #3387, opened 2020-05-10
windows newline issue withSplitFrontMatter, fix for #3386

oganga:
Had an issue with templates and SplitFrontMatter that did not work when on Windows.
Created an test case and found that the issue was in the way fronmatter was detected.

Due to the extra control characters in Windows newline it did not match correctly.

Fixed by removing the extra "\r" 

Fix for https://github.com/caddyserver/caddy/issues/3386

And to some degree cloud also add value to #3191

### DIFF
--- a/modules/caddyhttp/templates/frontmatter.go
+++ b/modules/caddyhttp/templates/frontmatter.go
@@ -29,8 +29,10 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 		lineEmpty = lineEmpty && unicode.IsSpace(b)
 	}
 	firstLine := input[firstLineStart:firstLineEnd]
-	//ensure residue windows newline is removed
+
+	// ensure residue windows newline is removed
 	firstLine = strings.Trim(firstLine, "\r")
+
 	// see what kind of front matter there is, if any
 	var closingFence string
 	var fmParser func([]byte) (map[string]interface{}, error)
--- a/modules/caddyhttp/templates/frontmatter.go
+++ b/modules/caddyhttp/templates/frontmatter.go
@@ -31,7 +31,7 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 	firstLine := input[firstLineStart:firstLineEnd]
 
 	// ensure residue windows newline is removed
-	firstLine = strings.Trim(firstLine, "\r")
+	firstLine = strings.TrimSpace(firstLine)
 
 	// see what kind of front matter there is, if any
 	var closingFence string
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/templates/frontmatter.go
```diff
@@ -29,7 +29,8 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 		lineEmpty = lineEmpty && unicode.IsSpace(b)
 	}
 	firstLine := input[firstLineStart:firstLineEnd]
-
+	//ensure residue windows newline is removed
+	firstLine = strings.Trim(firstLine, "\r")
```
francislavoie, line 33: Just a minor nit, we typically have a space at the start of comments, and leaving some empty lines around this is good because it's a different "idea" than the either section before and after it.

```suggestion

	// ensure residue windows newline is removed
	firstLine = strings.Trim(firstLine, "\r")

```

francislavoie:
FYI don't worry about the TravisCI failure, it's a bit flaky.

modules/caddyhttp/templates/frontmatter.go
```diff
@@ -30,6 +30,9 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 	}
 	firstLine := input[firstLineStart:firstLineEnd]
 
+	// ensure residue windows newline is removed
+	firstLine = strings.Trim(firstLine, "\r")
```
mholt, line 34: Should we just be generous and elide all extra whitespace? `strings.TrimSpace`

oganga: Then maybe in the for loop with the" if b == '\n'" switch to "if unicode.IsSpace(b)"
Then the strings.Trim will be redundant. 
Would require some benchmarks but could be faster all in all.

mholt: @oganga `unicode.IsSpace(b)` would also return true for literal space characters ` ` which is not what we want...

oganga: Ahe...well then it is not a good idea :)
I have no strong feeling for either version of trim.

modules/caddyhttp/templates/frontmatter.go
```diff
@@ -30,6 +30,9 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 	}
 	firstLine := input[firstLineStart:firstLineEnd]
 
+	// ensure residue windows newline is removed
```
mholt, line 33:
```suggestion
	// ensure residue windows carriage return byte is removed
```

PR #3389, opened 2020-05-10
caddyhttp: reverse-proxy command add --insecure flag

linquize:
caddyhttp: reverse-proxy command add --insecure flag

### DIFF
--- a/modules/caddyhttp/reverseproxy/command.go
+++ b/modules/caddyhttp/reverseproxy/command.go
@@ -59,6 +59,7 @@ default, all incoming headers are passed through unmodified.)
 			fs.String("from", "localhost", "Address on which to receive traffic")
 			fs.String("to", "", "Upstream address to which to to proxy traffic")
 			fs.Bool("change-host-header", false, "Set upstream Host header to address of upstream")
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/command.go
```diff
@@ -59,6 +59,7 @@ default, all incoming headers are passed through unmodified.)
 			fs.String("from", "localhost", "Address on which to receive traffic")
 			fs.String("to", "", "Upstream address to which to to proxy traffic")
 			fs.Bool("change-host-header", false, "Set upstream Host header to address of upstream")
+			fs.Bool("insecure", false, "Skip verify TLS, insecure")
```
mholt, line 62:
```suggestion
			fs.Bool("insecure", false, "Disable TLS verification (WARNING: DISABLES SECURITY, WHY ARE YOU EVEN USING TLS)")
```

I'm partly kidding but mostly serious about this suggestion

linquize: updated

mholt: Yessss thank you!

PR #3401, opened 2020-05-12
reverseproxy: Apply response header ops before copying it (fix #3382)

### LINKED ISSUE #3382: v2: header_down removes headers added by caddy too
I'm using an import for my security headers. They are in a snippets folder located where the Caddyfile is.

### snippets/sec-headers
```
header {
  X-Xss-Protection "1; mode=block"
  X-Robots-Tag "none"
  X-Permitted-Cross-Domain-Policies "none"
  X-Download-Options "noopen"
  Referrer-Policy "strict-origin-when-cross-origin"
  Cache-Control "public, max-age=15, must-revalidate"
  Server "SomeServer"
  X-Frame-Options "SAMEORIGIN"
  Content-Security-Policy "upgrade-insecure-requests"
}
```

### Caddyfile
```
example.com {

  import snippets/sec-headers

  reverse_proxy http://backend.local {
      header_down -Server
  }

}
```

The `backend.local` service sends it's own `Server` header which I want to remove, so that there is only the `Server` header from the imported `snippets/sec-headers` file. But if I use the `header_down` directive like this, no `Server` header reaches the client and if I remove the directive, the client receives two headers, one with the value from the backend and one from the `sec-headers` file. In Caddy v1 this worked like a charm with the header_downstream directive.

I know that I can overwrite the headers in the header_down directive like
```
reverse_proxy http://backend.local {
  header_down Server "SomeServer"
}
```
which works, but that means that I have to set all headers for every reverse proxy I use (which are many) separately. This would be really impractical if I'd like to change or add something in the future.

So what can I do to fix it, or is this a bug?

(I'm using the latest `caddy` docker image with only the standard plugins loaded.)

mholt:
I confirmed that this fixes #3382. Thanks for the report @MexHigh - please try this out. There are build artifacts available in the Actions / CI.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -510,6 +510,15 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, di Dia
 		res.Header.Del(h)
 	}
 
+	// apply any response header operations
+	if h.Headers != nil && h.Headers.Response != nil {
+		if h.Headers.Response.Require == nil ||
+			h.Headers.Response.Require.Match(res.StatusCode, res.Header) {
+			repl := req.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
+			h.Headers.Response.ApplyTo(res.Header, repl)
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


mholt:
Great, thanks!

PR #3407, opened 2020-05-13
docs: link to CEL standard definitions

Mohammed90:
As @abiosoft recommended

### DIFF
--- a/modules/caddyhttp/celmatcher.go
+++ b/modules/caddyhttp/celmatcher.go
@@ -42,7 +42,7 @@ func init() {
 // MatchExpression matches requests by evaluating a
 // [CEL](https://github.com/google/cel-spec) expression.
 // This enables complex logic to be expressed using a comfortable,
-// familiar syntax.
+// familiar syntax. The standard definitions of CEL functions and operators can be found [here](https://github.com/google/cel-spec/blob/master/doc/langdef.md#standard-definitions).
 //
 // This matcher's JSON interface is actually a string, not a struct.
 // The generated docs are not correct because this type has custom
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/celmatcher.go
```diff
@@ -42,7 +42,7 @@ func init() {
 // MatchExpression matches requests by evaluating a
 // [CEL](https://github.com/google/cel-spec) expression.
 // This enables complex logic to be expressed using a comfortable,
-// familiar syntax.
+// familiar syntax. The standard definitions of CEL functions and operators can be found [here](https://github.com/google/cel-spec/blob/master/doc/langdef.md#standard-definitions).
```
abiosoft, line 45: I know it has no effect but my eyes can't help. Is that line not too long?

Mohammed90: The sight of it bothers me too. I think it'll be wrapped when rendered though. Putting that sentence on a whole line just by itself doesn't feel right. I don't know whether new-lines within sentence will be rendered into actual <br>. @mholt, how does your doc generator handle it?

francislavoie:
```suggestion
// familiar syntax. The standard definitions of CEL functions and operators can be found
// [here](https://github.com/google/cel-spec/blob/master/doc/langdef.md#standard-definitions).
```

Somewhat matches how the line above with `[CEL]` looks :shrug:

mholt: Ha, I had the same question as @abiosoft . Do the best you can with long links, even if they get their own wrapped line, it's OK; better than the whole sentence extending waaaay out there IMO.

mholt, line 45:
```suggestion
// familiar syntax. Please refer to
// [the standard definitions of CEL functions and operators](https://github.com/google/cel-spec/blob/master/doc/langdef.md#standard-definitions).
```

PR #3408, opened 2020-05-13
add CookieHashSelection for session affinity

utick:


francislavoie:
Hmm. We typically prefer to discuss new features in an issue before accepting an implementation. This seems like a pretty significant change.

mholt:
Adding a ResponseWriter parameter to the selector interface is technically a breaking change... but I don't know of any code aside from this repo that depends on that particular interface.

Still, we need to have a really good reason for adding it. Is this not something that can/should be done with a handler instead?

modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -380,6 +381,48 @@ func (s *HeaderHashSelection) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+	cookie, err := req.Cookie(s.Field)
+	var cookieValue string
+	if err != nil || cookie == nil {
+		cookieValue = caddy.RandomString(16)
+		http.SetCookie(w, &http.Cookie{Name: s.Field, Value: cookieValue, Secure: false})
+	} else {
+		cookieValue = cookie.Value
+	}
+	return hostByHashing(pool, cookieValue)
+}
```
francislavoie, line 413: I think how it should work is more like this:

- Check if there's a cookie.
- If there's no cookie:
  - Get a random host (see RandomChoiceSelection)
  - Hash (maybe HMAC with some key for privacy), or encrypt (with some key, for privacy + reversibility) the upstream.Dial string as the cookie value, write the cookie.
- If there's a cookie:
  - If you chose to hash, then loop over the upstreams until you find a match.
  - If you chose to encrypt, then decrypt and lookup the upstream.
  - If there's no match, then give a random upstream again (i.e. the upstream is no longer available), write the cookie.
- Return the upstream.

mholt:
@utick Thank you for working on this. If we can resolve the issues and implement a slightly better design, I think I'd be willing to accept a change like this, even if it means adding the ResponseWriter to the signature...

francislavoie:
@utick are you still working on this?

I'll close this for now, please let us know if you plan to continue working on it.

mholt:
I haven't decided on this yet. We could add the response writer to the method, but it's just a bit weird as I hadn't considered whether a LB policy should be able to modify the response. Maybe we should just pass the response _header_ so that it can only manipulate that. Thoughts?

francislavoie:
Alright this got inactive, but has been rebased and updated by @d-masson in https://github.com/caddyserver/caddy/pull/3809 so I'll close this one.

Thanks for your work on this @utick!

PR #3414, opened 2020-05-15
caddyfile: Allow empty files to be imported

bhcleek:
Allow empty files to be imported.

This may not be the right change to make. It's possible that you'd rather the io.EOF be handled at https://github.com/caddyserver/caddy/blob/master/caddyconfig/caddyfile/parse.go#L380-L382

### DIFF
--- a/caddyconfig/caddyfile/lexer.go
+++ b/caddyconfig/caddyfile/lexer.go
@@ -49,6 +49,9 @@ func (l *lexer) load(input io.Reader) error {
 	// discard byte order mark, if present
 	firstCh, _, err := l.reader.ReadRune()
 	if err != nil {
+		if err == io.EOF {
+			return nil
+		}
 		return err
 	}
 	if firstCh != 0xFEFF {
--- a/caddyconfig/caddyfile/parse_test.go
+++ b/caddyconfig/caddyfile/parse_test.go
@@ -23,21 +23,47 @@ import (
 )
 
 func TestAllTokens(t *testing.T) {
-	input := []byte("a b c\nd e")
-	expected := []string{"a", "b", "c", "d", "e"}
-	tokens, err := allTokens("TestAllTokens", input)
-
-	if err != nil {
-		t.Fatalf("Expected no error, got %v", err)
+	tests := []struct {
+		name     string
+		input    string
+		expected []string
+	}{
+		{
+			name:     "not-empty",
+			input:    "a b c\nd e",
+			expected: []string{"a", "b", "c", "d", "e"},
+		}, {
+			name:  "empty",
+			input: "",
+		}, {
+			name:  "newline",
+			input: "\n",
+		}, {
+			name:  "space",
+			input: " ",
+		}, {
+			name:  "tab and newline",
+			input: "\t\n",
+		},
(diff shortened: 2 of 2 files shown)


mholt:
Is there an issue discussing this change?

What do you need this for? We have found that most errors reading files are in fact errors due to permissions or disk, not because the sysadmin intentionally wanted them to be empty, so we treat it as an error instead of letting it fly under the radar.

bhcleek:
I didn't create an issue discussing this change.

The use case I have is running CoreDNS in kubernetes with a configmap that's mounted as files. The key in the configmap for the file that's `import`ed in the `Corefile` has an empty value, so CoreDNS doesn't start because it goes to parse the imported file, finds it empty, and crashes.

Here's an example configmap:
```
Corefile: |
    .:53 {
        errors
        health
        ready

        autopath @kubernetes
        kubernetes {$CLUSTER_DOMAIN} in-addr.arpa ip6.arpa {
          pods verified
          endpoint_pod_names
          fallthrough in-addr.arpa ip6.arpa
        }
        template ANY AAAA {
          match "^ns\.dns\.{$CLUSTER_DOMAIN}\.$"
          authority "{$CLUSTER_DOMAIN} 30 IN SOA ns.dns.{$CLUSTER_DOMAIN} hostmaster.{$CLUSTER_DOMAIN} (1 60 60 60 60)"
          fallthrough
        }
        prometheus :10054
        forward . /etc/resolv.conf
        cache 60
        loop
        reload
        loadbalance

        import ingresses.conf
    }
  ingresses.conf: ""
```

bhcleek:
@mholt this shouldn't break your the use cases you mentioned; when reading a file returns an error due to permissions or disk issues, the error won't be `io.EOF`.

mholt:
Hm, I do seem to recall situations where the disk or system was letting the file be opened but for whatever reason, reads truncated at size 0.

Anyway, I'm open to the change, and can give it a proper review next week -- but why have an empty file at all?

Mohammed90:
For the record, the PR is against master, which affects v2, while CoreDNS is still dependent on v1.

bhcleek:
> Anyway, I'm open to the change, and can give it a proper review next week -- but why have an empty file at all?

So that we don't have to vary the Corefile between clusters and can let another process control the contents of the imported file.

mholt:
@bhcleek I don't totally understand, but that's fine.

Don't you think this change would be better suited for the v1 branch then?

bhcleek:
Honestly, we have two decent workarounds. One of which, a file with only a byte order mark in it, will work generally, so I don't feel a strong need to get this into the v1 branch, but I'll happily make a PR for it, too, if you want.

I'm happy to explain the issue in more detail if you want so you can understand what we're bumping up against: just let me know if you'd like me to try to clarify.

mholt:
@bhcleek I thought CoreDNS was based on Caddy 1? Isn't @Mohammed90 right, in that this should probably go into the v1 branch then?

bhcleek:
#3421 is the same change for the v1 branch

mholt:
Closing, unless there's a compelling reason that v2 needs this also.

PR #3416, opened 2020-05-15
caddyauth: Fix hash-password broken terminal state on SIGINT

francislavoie:
If you run `caddy hash-password` then hit `^C` to interrupt, the terminal would get stuck in a broken state. Not good.

This fixes that by storing the state before starting to read passwords (which modifies the terminal state) and restores it if `SIGINT` is caught.

This also has the side-effect of avoiding the default cleanup handling from Caddy which prints out some log lines that aren't helpful in this situation, e.g.:

```
2020/05/15 14:52:47.348	INFO	shutting down	{"signal": "SIGINT"}
2020/05/15 14:52:47.349	ERROR	stopping admin endpoint	{"signal": "SIGINT", "error": "no admin server"}
2020/05/15 14:52:47.349	INFO	shutdown done	{"signal": "SIGINT"}
```

And also, I changed the prompts to be printed to STDERR instead of STDOUT, so now you can do `caddy hash-password > hashed.txt` and it works as expected.

/cc @0az FYI

### DIFF
--- a/modules/caddyhttp/caddyauth/command.go
+++ b/modules/caddyhttp/caddyauth/command.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/signal"
 
 	"github.com/caddyserver/caddy/v2"
 	caddycmd "github.com/caddyserver/caddy/v2/cmd"
@@ -67,17 +68,29 @@ func cmdHashPassword(fs caddycmd.Flags) (int, error) {
 	salt := []byte(fs.String("salt"))
 
 	if len(plaintext) == 0 {
-		if terminal.IsTerminal(int(os.Stdin.Fd())) {
-			fmt.Print("Enter password: ")
-			plaintext, err = terminal.ReadPassword(int(os.Stdin.Fd()))
-			fmt.Println()
+		fd := int(os.Stdin.Fd())
+		if terminal.IsTerminal(fd) {
+			// ensure the terminal state is restored on SIGINT
+			state, _ := terminal.GetState(fd)
+			c := make(chan os.Signal)
+			signal.Notify(c, os.Interrupt)
+			go func() {
+				<-c
+				_ = terminal.Restore(fd, state)
+				os.Exit(caddy.ExitCodeFailedStartup)
+			}()
+			defer signal.Stop(c)
+
+			fmt.Fprint(os.Stderr, "Enter password: ")
+			plaintext, err = terminal.ReadPassword(fd)
+			fmt.Fprintln(os.Stderr)
 			if err != nil {
 				return caddy.ExitCodeFailedStartup, err
 			}
 
-			fmt.Print("Confirm password: ")
-			confirmation, err := terminal.ReadPassword(int(os.Stdin.Fd()))
-			fmt.Println()
+			fmt.Fprint(os.Stderr, "Confirm password: ")
+			confirmation, err := terminal.ReadPassword(fd)
+			fmt.Fprintln(os.Stderr)
 			if err != nil {
 				return caddy.ExitCodeFailedStartup, err
 			}
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -148,6 +148,8 @@ func cmdStart(fl Flags) (int, error) {
 }
 
 func cmdRun(fl Flags) (int, error) {
+	caddy.TrapSignals()
+
 	runCmdConfigFlag := fl.String("config")
 	runCmdConfigAdapterFlag := fl.String("adapter")
 	runCmdResumeFlag := fl.Bool("resume")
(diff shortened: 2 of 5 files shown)


modules/caddyhttp/caddyauth/command.go
```diff
@@ -67,17 +68,28 @@ func cmdHashPassword(fs caddycmd.Flags) (int, error) {
...
+		if terminal.IsTerminal(fd) {
+			// ensure the terminal state is restored on SIGINT
+			state, _ := terminal.GetState(fd)
+			c := make(chan os.Signal)
+			signal.Notify(c, os.Interrupt)
+			go func() {
+				<-c
+				_ = terminal.Restore(fd, state)
+				os.Exit(caddy.ExitCodeFailedStartup)
+			}()
```
mholt, line 81: I think this goroutine leaks if no signal is received.

modules/caddyhttp/caddyauth/command.go
```diff
@@ -67,17 +68,28 @@ func cmdHashPassword(fs caddycmd.Flags) (int, error) {
...
+
+			fmt.Fprint(os.Stderr, "Enter password: ")
```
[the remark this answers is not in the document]
francislavoie: I think you misunderstood. If you do `caddy hash-password > file.txt` then it will contain:

```
Enter password: 
Confirm password: 
JDJhJDEwJGRCVzVGY0o1UHhZWDJqczRCdnF6aXVxTUppM0xaQmhUZDViUG9ZT0NsU0NvLmZhSDhHOFMu
```

That doesn't make sense, it should only contain the password. Also, if you use `>` then you won't see the prompts. It's better to print those on stderr.

modules/caddyhttp/caddyauth/command.go
francislavoie: :thinking: I don't know what that means frankly. Still a relative noob at Go :smile:

modules/caddyhttp/caddyauth/command.go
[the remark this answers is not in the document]
mholt: Ah :-/

This is unfortunate. Stderr is reserved for logs and errors.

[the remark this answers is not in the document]
francislavoie: Well this is a special command, I don't think it's really that big of a problem, ultimately. I think usability here is more important :man_shrugging:

modules/caddyhttp/caddyauth/command.go
francislavoie: @mholt what should I do here? :thinking:

modules/caddyhttp/caddyauth/command.go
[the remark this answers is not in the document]
francislavoie: FWIW, there's precedent for this: https://pymotw.com/3/getpass/index.html

That's a python lib, it specifically recommends using stderr if the stdout output is useful, which is the case here.

modules/caddyhttp/caddyauth/command.go
mholt: I guess that's true, it's just not as clean as it could be, probably.

What I mean is that `<-c` never unblocks if the program never receives an interrupt. So if the user goes through the process normally, this goroutine will never exit. Of course, as you note, this is a short-lived command, I'm just noting it's not the cleanest Go code since it does technically leak a goroutine.

If we're going to leave it unfixed, we should at least comment to explain that the goroutine never unblocks (feel free to verify by adding `log.Println()` on both sides of the `<-c` and see what happens) but that we don't care because the program terminates quickly anyway.

A proper fix probably involves closing the channel (`close(c)`) when we're no longer accepting signals... oh -- actually maybe https://golang.org/pkg/os/signal/#Stop needs to be used (instead?).

francislavoie: I added `defer signal.Stop(c)`, does what we need I think.

francislavoie:
Big note: this PR now also moves the `caddy.TrapSignals()` call to inside the subcommands that it's relevant for instead of at the start of `Main()`. This could be a breaking change for plugins that add long-running commands.

/cc @lucaslorentz this is relevant to you. You'll need to add `caddy.TrapSignals()` at the top of your `cmdFunc` function once this PR is merged.

0az:
> That's a python lib, it specifically recommends using stderr if the stdout output is useful, which is the case here.

> The stderr thing is a bit weird but I'll let it slide since this is kind of a dusty corner of the code base that I don't expect will be used much anyway... and I guess it makes sense... sigh I hate shells.

The getpass module is actually part of the Python stdlib: https://docs.python.org/3/library/getpass.html. The module docs say it defaults to /dev/tty, and falls back to stderr, so that's something that could get added in a following PR. This might be better than using stderr 100% of the time.

francislavoie:
@0az if you want to make a PR for that, that'd be great. I feel out of my depth here, I just wanted to fix the terminal state getting clobbered.

PR #3421, opened 2020-05-16
v1: Allow empty files to be imported

bhcleek:
* add tests for an empty file
* return nil from `lexer.load` when the file is empty

This is #3414 for v1, too.

### DIFF
--- a/caddyfile/lexer.go
+++ b/caddyfile/lexer.go
@@ -48,6 +48,9 @@ func (l *lexer) load(input io.Reader) error {
 	// discard byte order mark, if present
 	firstCh, _, err := l.reader.ReadRune()
 	if err != nil {
+		if err == io.EOF {
+			return nil
+		}
 		return err
 	}
 	if firstCh != 0xFEFF {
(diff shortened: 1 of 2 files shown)


francislavoie:
Can you confirm the tests run locally?

CI is kinda busted on the v1 branch, so you can probably ignore that, as long as the tests pass locally.

bhcleek:
Yeah, the tests pass locally. I verified that the tests fail in the commit that adds them and then the tests pass in the second commit.

bhcleek:
@mholt it would probably be surprising to users to have this fixed on v1, upgrade to v2 and have this regress. Can I convince you to bring it into v2, too?

mholt:
@bhcleek It's a pretty niche requirement, yours is the only use case I've heard of, and it doesn't use v2.

Let's wait and see if there are compelling reasons to put it into v2 as well. (I can't find it, but again, I seem to recall users reporting problems with empty files due to some misconfig in their system or environment, and reporting an error was a good thing. I might be wrong but my sense is tingling.)

PR #3423, opened 2020-05-18
caddyfile: Add support for args on imports

francislavoie:
Stems from the discussion in https://caddy.community/t/logging-snippet-using-host-placeholder-for-log-filename/8218

The idea here is that it's common for users to want to have one element of an imported snippet or file be replaced with a context-specific value.

This PR adds support for `{args.*}` placeholders at Caddyfile-parse time when importing tokens from either a snippet or file, where `*` is the positional argument number.

The `import` syntax becomes:

```
import <pattern> <args...>
```

---

For example, consider this Caddyfile, where a user wants to enable logging using a common config for each of their sites. You'll see that the log filename is expanded to `/var/log/caddy/a.example.com.access.log`:
```
(logging) {
  log {
    output file /var/log/caddy/{args.0}.access.log
  }
}

a.example.com {
  import logging a.example.com
}

b.example.com {
  import logging b.example.com
}
```

Adapted JSON output:
```json
{
	"logging": {
		"logs": {
			"default": {
				"exclude": [
					"http.log.access.log0",
					"http.log.access.log1"
				]
			},
			"log0": {
				"writer": {
					"filename": "/var/log/caddy/a.example.com.access.log",
					"output": "file"
				},
				"include": [
					"http.log.access.log0"
				]
			},
			"log1": {
				"writer": {
					"filename": "/var/log/caddy/b.example.com.access.log",
					"output": "file"
				},
				"include": [
					"http.log.access.log1"
				]
			}
		}
	},
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"match": [
								{
									"host": [
										"a.example.com"
									]
								}
							],
							"terminal": true
						},
						{
							"match": [
								{
									"host": [
										"b.example.com"
									]
								}
							],
							"terminal": true
						}
					],
					"logs": {
						"logger_names": {
							"a.example.com": "log0",
							"b.example.com": "log1"
						}
					}
				}
			}
		}
	}
}
```

---

This also supports an arbitrary number of arguments, and supports imported files.

The file `respond.txt`
```
respond "I am {args.0}, hears {args.1}"
```

Caddyfile:
```
example.com

import respond.txt Groot Rocket
import respond.txt you "the confused man"
```

Adapted JSON:
```json
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"match": [
								{
									"host": [
										"example.com"
									]
								}
							],
							"handle": [
								{
									"handler": "subroute",
									"routes": [
										{
											"handle": [
												{
													"body": "I am groot, hears rocket",
													"handler": "static_response"
												},
												{
													"body": "I am you, hears the confused man",
													"handler": "static_response"
												}
											]
										}
									]
								}
							],
							"terminal": true
						}
					]
				}
			}
		}
	}
}
```


sumnerboy12:
Looks great @francislavoie, thanks!

caddyconfig/caddyfile/parse.go
```diff
@@ -348,10 +359,20 @@ func (p *parser) doImport() error {
 		}
 	}
 
+	// copy the tokens so we don't overwrite p.definedSnippets
+	tokensCopy := make([]Token, len(importedTokens))
+	copy(tokensCopy, importedTokens)
```
Mohammed90, line 364: These two lines can be replaced by just this:
```suggestion
	tokensCopy := importedTokens[:]
```

mholt: @Mohammed90 They are different, though -- your suggestion does not perform a copy, so any changes to the elements of the slice will be reflected in both importedTokens and tokensCopy.

Mohammed90: Nope, it won't reflect on the original. I've tested on the playground.
https://play.golang.org/p/P2MjWaDexhL

It's `tokensCopy := importedTokens` vs `tokensCopy := importedTokens[:]` that determines whether it's value copy or pointer copy.

Mohammed90: Am I missing something slice internals pitfall?

Mohammed90: I see another one-liner here citing more benefits: https://github.com/go101/go101/wiki/How-to-perfectly-clone-a-slice%3F

mholt: @Mohammed90 Your example is wrong, I think, to demonstrate a copy. Here is the corrected version: https://play.golang.org/p/EVDWDqA9ntL (see line 12)

francislavoie: I just tried the above @Mohammed90, doesn't work. Both log filenames end up being `a.example.com`.

Mohammed90: Ah I see. It goes in one direction not the other.

francislavoie: This does work though, but looks really ugly IMO.

```go
tokensCopy := append(importedTokens[:0:0], importedTokens...)
```

mholt: @Mohammed90 The `[:]` notation creates a new slice _header_ over an underlying array (often used for converting arrays to slices, or for copying a header of an existing slice -- but the underlying elements stay the same, so if you change `a[0]` or `b[0]` it changes it for both.

Mohammed90: I've just played with the example more and realized what I missed 🤦‍♂️

PR #3430, opened 2020-05-19
httpcaddyfile: Support inline path prefix matcher tokens

francislavoie:
This makes it possible to use a suffix matcher like `*.txt` inline wherever you can have a `/prefix/*` matcher as well. We already use `*` as the wildcard, so expanding that to also allow a suffix isn't a bit stretch. Example:

```
respond *.txt "yep it's a text file"
```

I think the likelihood of this breaking someone's config is pretty low, but there is a slight chance if they rely on it as the first argument to some directive. Not sure I have anything but contrived examples here.

I think we just need to update https://caddyserver.com/docs/caddyfile/matchers#syntax to mention another option for the inline syntax.

I'll probably need to update my syntax highlighter for this :sweat_smile: and the caddy-docker-proxy will probably need a slight update as well to support this. No big deal though.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -890,8 +890,8 @@ func matcherSetFromMatcherToken(
 	if tkn.Text == "*" {
 		// match all requests == no matchers, so nothing to do
 		return nil, true, nil
-	} else if strings.HasPrefix(tkn.Text, "/") {
-		// convenient way to specify a single path match
+	} else if strings.HasPrefix(tkn.Text, "/") || strings.HasPrefix(tkn.Text, "*") {
+		// convenient way to specify a single prefix or suffix path match
 		return caddy.ModuleMap{
 			"path": caddyconfig.JSON(caddyhttp.MatchPath{tkn.Text}, warnings),
 		}, true, nil
--- a/caddytest/integration/caddyfile_adapt/matcher_syntax.txt
+++ b/caddytest/integration/caddyfile_adapt/matcher_syntax.txt
@@ -9,6 +9,8 @@
 
 	@matcher3 not method PUT
 	respond @matcher3 "not put"
+
+	respond *.txt "text"
 }
 ----------
 {
@@ -20,6 +22,21 @@
 						":80"
 					],
 					"routes": [
+						{
+							"match": [
+								{
+									"path": [
+										"*.txt"
+									]
+								}
+							],
+							"handle": [
+								{
+									"body": "text",
+									"handler": "static_response"
+								}
+							]
+						},
 						{
 							"match": [
 								{


mholt:
I don't really like the idea of adding more special cases at this point. And technically, this is a breaking change...

whitestrake:
Technically breaking, yes, but in practice I expect nobody's config to break (although I'd welcome anyone to indicate a real example where it would).

I'd support this change, I think it's a good one.

francislavoie:
I went through the [list of directives](https://caddyserver.com/docs/caddyfile/directives) and I can't think of a single case that would break. Maybe a plugin might break, but I don't think we're too concerned about that right now.

mholt:
I don't have a concrete reason why I'm against this, but a few weeks ago (maybe a couple months ago now!) I remember I actually implemented this on a local branch and then fiddled with it a bit and then disbanded it, and I remember I thought had a good reason, but I can't remember what the reason was.

I'm going to sit on this -- maybe close it? (to avoid raising expectations) -- until this change sits better with me. I dunno that adding yet another special case to the Caddyfile is a good idea, when the alternative is one very easy to read line:

```
@txt path *.txt
```

I think part of the reason I don't love this is because it's a path matcher that doesn't start with `/`, which is definitely not obvious / not what you'd expect...

francislavoie:
> That's a path matcher that doesn't start with /, which is definitely not obvious / not what you'd expect...

IMO it feels pretty valid because we already have `*` as a wildcard and `*` as a suffix. I don't think adding `*` as a prefix match is much of a stretch for that reason.

But fair enough, we could mark as draft and put it as 2.2 milestone?

polarathene:
> That's a path matcher that doesn't start with `/`, which is definitely not obvious / not what you'd expect...

You could just require the `/`? So it'd be `/*.txt`. Would follow the rule as outlined [here](https://caddyserver.com/docs/caddyfile/matchers#path-matchers) with the `/` being required. Although if that became available, one would expect a future request to support something like `/*.{css, js}` or `/*.(css|js)`.

mholt:
@polarathene 

> You could just require the /? So it'd be /*.txt. Would follow the rule as outlined here with the / being required. 

Yep, I would love that -- but a path matcher does not do prefix matching when the `*` appears in the middle of the string. So, that would match `/foo.txt` but not `/foo/bar.txt`. In fact, this already works. It's very useful sometimes. But we probably shouldn't/can't change that behavior.

francislavoie:
I think `**` is how some glob parsers deal with that, i.e. many path segments. Not sure if we can easily support that or not but it's something to consider.

mholt:
I've thought about that too, and I think we can do that, but we'd have to implement it ourselves.

mholt:
Going to push this out of the 2.2 milestone until we can figure out a better solution... I don't hate the `**` idea but it depends on the implementation.

jbritton916:
https://github.com/caddyserver/caddy/pull/3430#issuecomment-632284710

jbritton916:
https://github.com/caddyserver/caddy/pull/3430#issuecomment-632284710

jbritton916:
Find location

francislavoie:
I'll close this for now, I'm not really planning on working on the glob idea.

PR #3431, opened 2020-05-20
httpcaddyfile: Improve error on matcher declared outside site block

francislavoie:
We don't support defining matchers outside of site blocks. Some users try it, but get confusing error messaging because it still gets read as a site address. Instead, we can improve the error message in those cases because a site address starting with `@` isn't valid anyways.

Consider this Caddyfile:
```
@foo {
	path /bar/*
}

localhost {
	respond "hello"
}
```

Before:
```
$ ./caddy adapt --config Caddyfile
adapt: Caddyfile:2: unrecognized directive: path
```

After:
```
$ ./caddy adapt --config Caddyfile
adapt: cannot define a matcher outside of a site block: '@foo'
```


francislavoie:
Good point. Updated.

PR #3436, opened 2020-05-22
chore: simplify goreleaser flow, add bash completions to .deb

Mohammed90:
There's no need to create a whole new directory and copy `main.go` into it. It's already there, so just re-purpose it. Also, let's package bash-completions now that we have them.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -82,9 +82,9 @@ nfpms:
       ./caddy-dist/init/caddy.service: /lib/systemd/system/caddy.service
       ./caddy-dist/init/caddy-api.service: /lib/systemd/system/caddy-api.service
       ./caddy-dist/welcome/index.html: /usr/share/caddy/index.html
+      ./caddy-dist/scripts/completions/bash-completion: /etc/bash_completion.d/caddy
     config_files:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -142,6 +158,6 @@ jobs:
       - uses: goreleaser/goreleaser-action@v1
         with:
           version: latest
-          args: check
+          args: build --snapshot
```
Mohammed90, line 161: If this attempt is successful, this won't be needed

.goreleaser.yml
```diff
@@ -86,6 +84,7 @@ nfpms:
       ./caddy-dist/welcome/index.html: /usr/share/caddy/index.html
     config_files:
       ./caddy-dist/config/Caddyfile: /etc/caddy/Caddyfile
+      ./caddy-dist/scripts/completions/bash-completion: /etc/bash_completion.d/caddy
```
francislavoie, line 87: Actually this should be a `files`, not `config_files`. The difference is that `apt remove` doesn't remove `config_files`, but does remove `files`, and `apt purge` removes both. Doesn't make sense to keep bash completions around after remove.

PR #3437, opened 2020-05-23
chore: Add issue templates

francislavoie:


francislavoie:
> but issue templates have proven problematic

I don't really understand this, what's the evidence for that? What problems?

I feel pretty strongly that redirecting people to the forums for questions up-front would save time and confusion from users.

mholt:
What kind of evidence do you want? Links to past drama where issue templates upset users? 😛  Most users ignore issue templates, and they're even more problematic when we require them before they even say anything. Canned replies are much better because it lowers the barrier to entry and gives us the chance to tailor our response and the issue to their specific needs with a single click, which also increases community interactivity.

Also, most users -- especially the same ones that ignore or skim on the issue templates -- don't know the difference between a bug report and a question.

The lowest-friction thing we can do for both us and them is to just have a canned reply and do the work of that one heavy click to ask that they take their question to the forum or fill out a template for more information. On this topic, you might just have to trust my judgment based on 5+ years of maintaining this particular project. Sorry that I don't have more to say on it, but I feel strongly about this, and we'll do without.

vaaski:
I'd like to give my input on this, how about just adding a small and short comment to the issue template that states something like "This is for bug reports & feature requests, for help on using Caddy, please refer to https://caddy.community".

I think that could only help and shouldn't cause any drama or anyone to be upset, and it would've definitely prevented me from opening #3499.

mholt:
@vaaski We tried that, but the problem is that people don't know how to properly classify their post as a feature request or a bug report or a request for help. We also know from years of experience that people are too lazy to read and follow issue templates properly.

Issue #3402, opened 2020-05-12
v2: Writing unit tests for Caddyfile parsing logic


kujenga:
I'm looking at porting a plugin from v1 to v2 and have gotten stuck migrating over tests. I previously had tests that utilized the `caddy.NewTestController` method on Caddy v1, in order to pass that test controller to the same `setup` function that I used in the initialization of my plugin.

```go
	caddy.RegisterPlugin("my_plugin", caddy.Plugin{
		ServerType: "http",
		Action:     setup,
	})
```

At a high level, my goals are to port over the following types of tests:
- unit tests that assert that a given set of Caddyfile directives are parsed in the expected manner. The tests currently take in values like the following, run them through `caddy.NewTestController`, and setup the plugin from there. They then verify that the output configuration looks as expected.
```
my_plugin test:1234 {
    key value
    other
}
```
- Tests that verify that the middleware is present in the Caddy server, and inspect values on it to make sure that it is configured correctly. An example of this:
```go
ctrlr := caddy.NewTestController(...)
setup(ctrlr)

cfg := httpserver.GetConfig(ctrlr)
v = cfg.Middleware()[0](nil).(*MyPlugin)

// ...various assertions
```

I started looking at the `httpcaddyfile.ServerType.Setup` [1] function as a possible replacement but it seemed unclear if that was a productive route to go. Maybe there is a better way to test plugins that requires re-thinking tests more holistically as part of porting them off of v1?

[1] https://pkg.go.dev/github.com/caddyserver/caddy/v2@v2.0.0/caddyconfig/httpcaddyfile?tab=doc#ServerType.Setup

francislavoie:
There's a handful of helpers in the `caddytest` package that you could use, such as `AssertAdapt`. See https://github.com/caddyserver/caddy/blob/master/caddytest/caddytest.go

BTW you linked docs from a pretty old version of Caddy, that's from beta 11, from December. Things have changed quite a bit since then!

mholt:
@kujenga  Thanks for working on a Caddy plugin!

In v2, you'll implement the caddyfile.Unmarshaler interface: https://pkg.go.dev/github.com/caddyserver/caddy/v2@v2.0.0/caddyconfig/caddyfile?tab=doc#Unmarshaler - this gives you a Dispenser; in tests you can call NewDispenser: https://pkg.go.dev/github.com/caddyserver/caddy/v2@v2.0.0/caddyconfig/caddyfile?tab=doc#NewDispenser

[...]

PR #3439, opened 2020-05-23
caddyfile: Make NewTestDispenser accessible for plugin authors

francislavoie:
See discussion in #3402, we've had a couple suggestions to move this out of the `_test` file to make it usable for plugin authors in their own tests.

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -17,6 +17,8 @@ package caddyfile
 import (
 	"errors"
 	"fmt"
+	"io"
+	"log"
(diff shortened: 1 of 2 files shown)


caddyconfig/caddyfile/dispenser.go
```diff
@@ -37,6 +39,16 @@ func NewDispenser(tokens []Token) *Dispenser {
 	}
 }
 
+// NewTestDispenser parses input into tokens and creates a new
+// Disenser for test purposes only; any errors are fatal.
```
nchagrass-bud, line 43: typo

PR #3442, opened 2020-05-24
fmt: preserve the input config file permissions

Mohammed90:
The earlier code set the permissions to 0644. It's possible the file has different permissions set by the file owner, so we don't want to alter anything beyond the content we're permitted and instructed to change.


francislavoie:
🤔 does it change permissions? I read this: https://github.com/golang/go/issues/35711

Mohammed90:
Huh, so it doesn't. Closing? The file always exists, so it will never use the new "permissions".

PR #3447, opened 2020-05-25
log: implement `formatted` log encoder which accepts custom template for log lines

### LINKED ISSUE #3417: Re-introduce formatter logger in v2?
In v1, the `log` directive could receive a format string, and print that exact thing to syslog or stdout.

Can this be re-introduced in v2? I appreciate the move to JSON-first but my environment is one of `journald`-based logging, with minimal use and pressure to go to any sort of centralised logging system, so formatting an entry to something that's comfortable at first glance is very helpful.

Here's an example of one of my journald-based formats;
`"{remote} - {user} [{when}] \"{method} {host}{uri} {proto}\" {status} {size} \"Referer: {>Referer}\" \"{>User-Agent}\""`

Mohammed90:
This implements a logger that accepts custom format. I picked `template` for the format because the encoder name is determined by an existing `format` key. This is extremely an inefficient implementation. It takes the request to encode the entries, encodes it as JSON, parses the JSON and collects keys/values (using [buger/jsonparser](https://github.com/)) and sets them in a new replacer, then runs the replacer against the provided template. The placeholders are namespaced by the calling app name followed by the series of keys traversed to reach the designated value. For example, the below is the JSON structure logged by the HTTP app in response to handling a request. To log the protocol, the placeholder would be `{http.request.proto}`.

In the Caddyfile, the absence of template implies configuring the template as Common Log Format.

```
{
	"request": {
		"method": "GET",
		"uri": "/",
		"proto": "HTTP/2.0",
		"remote_addr": "[::1]:50257",
		"host": "localhost",
		"headers": {
			"Te": [
				"trailers"
			],
			"User-Agent": [
				"Mozilla/5.0 (Windows NT 10.0; rv:68.0) Gecko/20100101 Firefox/68.0"
			],
			"Accept-Language": [
				"en-US,en;q=0.5"
			],
			"Accept-Encoding": [
				"gzip, deflate, br"
			],
			"Referer": [
				"https://localhost/"
			],
			"Cache-Control": [
				"max-age=0"
			],
			"Accept": [
				"text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8"
			],
			"Dnt": [
				"1"
			],
			"Upgrade-Insecure-Requests": [
				"1"
			]
		},
		"tls": {
			"resumed": false,
			"version": 772,
			"ciphersuite": 4865,
			"proto": "h2",
			"proto_mutual": true,
			"server_name": "localhost"
		}
	},
	"common_log": "::1 - - [25/May/2020:10:03:00 +0300] \"GET /.tooling/ HTTP/2.0\" 200 12851",
	"duration": 0.008519377,
	"size": 12851,
	"status": 200,
	"resp_headers": {
		"Server": [
			"Caddy"
		],
		"Content-Type": [
			"text/html; charset=utf-8"
		]
	}
}
```

Closes #3417

### DIFF
--- a/modules/logging/formattedencoder.go
+++ b/modules/logging/formattedencoder.go
@@ -1,0 +1,147 @@
+package logging
+
+import (
+	"bytes"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/buger/jsonparser"
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"go.uber.org/zap/buffer"
+	"go.uber.org/zap/zapcore"
+)
+
+func init() {
+	caddy.RegisterModule(FormattedEncoder{})
+}
+
+const commonLogFormat = `{http.common_log}`
+
+// FormattedEncoder allows the user to provide custom template for log prints. The
+// encoder builds atop the json encoder, thus it follows its message structure. The placeholders
+// are namespaced by the name of the app logging the message.
+type FormattedEncoder struct {
+	zapcore.Encoder `json:"-"`
+	LogEncoderConfig
+	Template string `json:"template,omitempty"`
+}
+
+func (FormattedEncoder) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID: "caddy.logging.encoders.formatted",
+		New: func() caddy.Module {
+			return &FormattedEncoder{
+				Encoder: new(JSONEncoder),
+			}
+		},
+	}
+}
+
+// Provision sets up the encoder.
+func (se *FormattedEncoder) Provision(ctx caddy.Context) error {
+	if se.Template == "" {
+		return fmt.Errorf("missing template for formatted log encoder")
+	}
+	se.Encoder = zapcore.NewJSONEncoder(se.ZapcoreEncoderConfig())
+	return nil
+}
+
+// Clone wraps the underlying encoder's Clone. This is
+// necessary because we implement our own EncodeEntry,
+// and if we simply let the embedded encoder's Clone
+// be promoted, it would return a clone of that, and
+// we'd lose our FormattedEncoder's EncodeEntry.
+func (se FormattedEncoder) Clone() zapcore.Encoder {
+	return FormattedEncoder{
+		Encoder:  se.Encoder.Clone(),
+		Template: se.Template,
+	}
+}
+
+// EncodeEntry partially implements the zapcore.Encoder interface.
+func (se FormattedEncoder) EncodeEntry(ent zapcore.Entry, fields []zapcore.Field) (*buffer.Buffer, error) {
+	repl := caddy.NewReplacer()
+	buf, err := se.Encoder.EncodeEntry(ent, fields)
+	if err != nil {
+		return nil, err
+	}
+	appName := strings.SplitN(ent.LoggerName, ".", 2)[0]
+	// set the vals in the replacer
+	err = jsonparser.ObjectEach(buf.Bytes(), visitor(appName, repl))
+	buf.Reset() // the buffer is only used to collect placeholders' values anyway
+	buf.Free()
+	if err != nil {
+		return nil, err
+	}
+
+	out := repl.ReplaceKnown(se.Template, "")
+	// Unescape escaped quotes
+	buf.AppendString(strings.Replace(out, `\"`, `"`, -1))
+	if !strings.HasSuffix(out, "\n") {
+		buf.AppendByte('\n')
+	}
+
+	return buf, err
+}
+
+func visitor(prefix string, repl *caddy.Replacer) func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) error {
+	format := fmt.Sprintf("%s.%%s", prefix)
+	keyFormat := func(itemKey string) string {
+		return fmt.Sprintf(format, itemKey)
+	}
+	return func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) error {
+		switch dataType {
+		case jsonparser.NotExist:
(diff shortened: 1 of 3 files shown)


modules/logging/formattedencoder.go
```diff
@@ -0,0 +1,147 @@
...
+	err = jsonparser.ObjectEach(buf.Bytes(), visitor(appName, repl))
+	buf.Reset() // the buffer is only used to collect placeholders' values anyway
+	buf.Free()
+	if err != nil {
+		return nil, err
+	}
+
+	out := repl.ReplaceKnown(se.Template, "")
+	// Unescape escaped quotes
+	buf.AppendString(strings.Replace(out, `\"`, `"`, -1))
```
Mohammed90, line 81: @mholt, is there anything besides this that needs to be unescaped?

modules/logging/formattedencoder.go
```diff
@@ -0,0 +1,147 @@
...
+		case jsonparser.Boolean:
+			v, _ := strconv.ParseBool(string(value))
+			repl.Set(keyFormat(string(key)), v)
+		case jsonparser.Null:
+			panic("unimplemented jsonparser visitor for Null")
+		case jsonparser.Unknown:
+			panic("unimplemented jsonparser visitor for Unknown")
+		default:
+			panic("completely unknown dataType")
+		}
```
Mohammed90, line 126: How should these cases (and `jsonparser.NotExist` above) be handled?

modules/logging/formattedencoder.go
mholt: Probably 😅 

I think I have a better idea though, lemme see if I can explain it clearly in the review.

modules/logging/formattedencoder.go
mholt: Hm, yeah instead of panicking we should probably just return an empty value. With structured logs we can't guarantee that a field will exist, so gracefully treat it as empty.

Mohammed90: The `panic` isn't meant to stay. It's meant tingle your spidey senses and get you to comment on it for feedback 😛  You're suggesting `repl.Set(keyFormat(string(key)), "")` this, right?

modules/logging/formattedencoder.go
```diff
@@ -67,15 +65,28 @@ func (se FormattedEncoder) EncodeEntry(ent zapcore.Entry, fields []zapcore.Field
...
+	})
 
 	out := repl.ReplaceKnown(se.Template, "")
+	// The buffer is only used to find the values of placeholders.
+	// The content has served its purpose. It's time for it to go to repurpose the buffer.
+	buf.Reset()
+
 	// Unescape escaped quotes
 	buf.AppendString(strings.Replace(out, `\"`, `"`, -1))
```
mholt: Why is this needed?

modules/logging/formattedencoder.go
```diff
@@ -17,7 +15,7 @@ func init() {
 	caddy.RegisterModule(FormattedEncoder{})
 }
 
-const commonLogFormat = `{http.common_log}`
+const commonLogFormat = `{common_log}`
```
mholt, line 18: Where is this used?

mholt:
Nice, this is looking a lot better!

modules/logging/formattedencoder.go
Mohammed90: It's used in the `UnmarshalCaddyfile` as fallback if no template is given. The JSON config doesn't have the same fallback. Should I implement the same?

modules/logging/formattedencoder.go
Mohammed90: Because strings with quotes inside the JSON end up being printed with the escape backslashes, e.g. `::1 - - [28/May/2020:22:45:40 +0300] \"GET / HTTP/2.0\" 200 78212`, while they only needed to be escaped because they were embedded inside a JSON string.

mholt: Interesting... so bruger's lib doesn't unescape quotes then?

modules/logging/formattedencoder.go
mholt: Ah, no, I think this is fine. Thanks

modules/logging/formattedencoder.go
Mohammed90: Nope. Honestly, I don't expect it to manipulate the strings. If it did, I'd call it a bug. It should just hand out the content as-is.

mholt: Make sense. I guess this is fine then! If we find this to be buggy or difficult to maintain, we could always surround the string with quotes and then call `json.Unmarshal()` to do the dirty work for us. But this LGTM for now.

mholt:
@Mohammed90 I think this is great -- but I am not confident about adding more logging to the core at this point. (I'm still trying to clean up the mess we already have in the core, heh.) What if we made a separate repo for this and made it a plugin people could add if they need it?

Mohammed90:
> @Mohammed90 I think this is great -- but I am not confident about adding more logging to the core at this point. (I'm still trying to clean up the mess we already have in the core, heh.) What if we made a separate repo for this and made it a plugin people could add if they need it?

Sure, that's fine too.

modules/logging/formattedencoder.go
```diff
@@ -0,0 +1,128 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
```
mholt, line 1: Since this is going into a different repo, and you wrote the whole thing, you can claim the copyright for it I think.

Mohammed90: It says `and The Caddy Authors` so it seems inclusive of anyone who contributed to the body of code.

Mohammed90:
The module is now available at https://github.com/caddyserver/format-encoder

PR #3451, opened 2020-05-26
reverseproxy: Fix HTTPS detection for active health checks #3450

### LINKED ISSUE #3450: Active health checks do not support https
I have reverse proxy setup going to a couple of https only local services however the active health check doesn't seem to respect the tls flag.

Route: 

```

{
              "match": [],
              "handle": [
                {
                  "handler": "reverse_proxy",
                  "transport": {
                    "protocol": "http",
                    "tls": {
                      "insecure_skip_verify": true
                    },
                    "keep_alive": {
                      "enabled": true
                    }
                  },
                  "health_checks": {
                    "active": {
                      "path": "/",
                      "expect_status": 200
                    }
                  },
                  "load_balancing": {
                    "selection_policy": {
                      "policy": "least_conn"
                    }
                  },
                  "upstreams": [
                    { "dial": "localhost:8030" },
                    { "dial": "localhost:8031" }
                  ]
                }
              ]
            }
```

```
2020/05/26 10:20:44.998 [34mINFO[0m   http.handlers.reverse_proxy.health_checker.active       HTTP request failed
        {"host": "localhost:8030", "error": "Get \"http://localhost:8030/\": EOF"}
```
Have I missed something or perhaps this check needs improving?

https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/healthchecks.go#L201

AndyBan:
Fixes #3450 - tested http+https

### DIFF
--- a/modules/caddyhttp/reverseproxy/healthchecks.go
+++ b/modules/caddyhttp/reverseproxy/healthchecks.go
@@ -198,7 +198,7 @@ func (h *Handler) doActiveHealthCheckForAllHosts() {
 func (h *Handler) doActiveHealthCheck(dialInfo DialInfo, hostAddr string, host Host) error {
 	// create the URL for the request that acts as a health check
 	scheme := "http"
-	if ht, ok := h.Transport.(*http.Transport); ok && ht.TLSClientConfig != nil {
+	if ht, ok := h.Transport.(TLSTransport); ok && ht.TLSEnabled() {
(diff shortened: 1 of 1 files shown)


francislavoie:
So you can confirm that fix worked for you?

AndyBan:
Yeah it works - tested with and out without the tls block and it sent the request using the correct protocol

PR #3463, opened 2020-05-31
ci: add CI on s390x

Mohammed90:
I tried to make it as simple as possible and as minimal as possible. The flow is:
- Read the current short-sha of git HEAD
- rsync the repo to the remote server into a directory whose name is the short-sha
- Store the exit code
- Delete the remote directory
- Exit with the exit code of the test run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,7 +121,7 @@ jobs:
     #     exit 1
 
   s390x-test:
-    name: Test IBM Linux on Z
+    name: test (s390x on IBM Z)
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code into the Go module directory
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,25 +126,25 @@ jobs:
     steps:
       - name: Checkout code into the Go module directory
         uses: actions/checkout@v2
-      # - name: Run Tests
-      #   run: |
-      #     mkdir -p ~/.ssh && echo -e "${SSH_KEY//_/\\n}" > ~/.ssh/id_rsa && chmod og-rwx ~/.ssh/id_rsa
+      - name: Run Tests
+        run: |
+          mkdir -p ~/.ssh && echo -e "${SSH_KEY//_/\\n}" > ~/.ssh/id_ecdsa && chmod og-rwx ~/.ssh/id_ecdsa
 
-      #     # short sha is enough?
-      #     short_sha=$(git rev-parse --short HEAD)
+          # short sha is enough?
+          short_sha=$(git rev-parse --short HEAD)
 
-      #     # The environment is fresh, so there's no point in keeping accepting and adding the key.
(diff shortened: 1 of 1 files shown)


francislavoie:
LGTM so far 👍

hairyhenderson:
> Is there any situation where a build would fail but tests would pass?

There shouldn't be - `go test` will first build all of the packages it's asked to test. In this case that's `./...`, which covers everything.

Mohammed90:
> Should we do a `go build` and maybe pull an artifact out? Is there any value in that? Is there any situation where a build would fail but tests would pass? 🤔

In such scenario, it should fail on the other platforms as well. I think such scenario is only possible if we have conditional build tags for this platform, which we don't. It's possible to do, but it'll involve convoluted way of breaking the flow into multiple steps and could complicate the clean-up after the run.

> Maybe for consistency we can name the test something like `test (s390x on IBM Z)`, it would look better alongside the existing check names I think.

Done

> Cross-Platform / test (s390x on IBM Z)
> succeeded 4 minutes ago **in 39s** 
> Cross-Platform / test (ubuntu-latest, 1.14.x)
> succeeded 2 minutes ago **in 2m 29s** 

Um... remote machine finishes test run in about ~25% 🤯

mholt:
@Mohammed90 

> Um... remote machine finishes test run in about ~25% 🤯

Is that because the s390x VM is on more dedicated hardware, probably?

francislavoie:
I think it's mainly because it has less steps to run; doesn't need to install Go every time and doesn't need to pull in as many packages from `go get` every run because they're kept around between runs.

francislavoie:
FYI @grooverdan

Mohammed90:
> I think it's mainly because it has less steps to run; doesn't need to install Go every time and doesn't need to pull in as many packages from `go get` every run because they're kept around between runs.

It's also faster than the specific test step `Run tests`. The timing of the overall run was just easier for me to copy. The timing listed here is just for the `Run tests`.

> test (ubuntu-latest, 1.14.x)
> 1m 35s 

vs.

> test (s390x on IBM Z)
> 37s

I'm leaning towards @mholt thinking it could be due to dedicated hardware; though I wonder if the architecture plays a role here.

grooverdan:
your ubuntu-latest - x86_64 test is with the go test -race -coverprofile options (and the IBM Z doesn't support race) so is probably a bit quicker because of that, in addition to decent dedicated hardware.

On the free tier signup I got similar to 37s on the test case only (not the first run some some caching effects are in play).
<pre>
[linux1@rhel7testmachine caddy]$ uname -a
Linux rhel7testmachine.novalocal 3.10.0-1062.12.1.el7.s390x #1 SMP Thu Dec 12 06:45:30 EST 2019 s390x s390x s390x GNU/Linux

[linux1@rhel7testmachine caddy]$ time env CC=gcc go test -v -coverprofile=cover-profile.out  ./...
..
real	0m23.364s
user	0m26.940s
sys	0m4.991s
</pre>

Thanks for continuing the s390x enabling. Did you come up with a good option for ppc64le?

Mohammed90:
@grooverdan, so far we only received a VM for s390x. I could go back to the contact at IBM once we find the approach stable and comfortable with it.

PR #3465, opened 2020-06-01
caddyauth: Cache basicauth results

### LINKED ISSUE #3462: Enabling basicauth causes CPU usage to spike
Enabling basicauth for a route causes CPU usage to spike significantly for very modest request rates. On a Linode tiny instance (1 vcpu and 1GB of ram, admittedly modest hardware) 10-20 requests per second are sufficient to occupy 60%-80% of the single CPU available.

Here's a profile:
- Generated using /debug/pprof/
- Caddy was under modest load. I suspect one could reproduce this with any synthetic load, but I noticed it while browsing an authenticated reverse_proxy route to a netdata instance. So this was a single web-browser using javascript to fetch metrics maybe 10-ish times per second for in-browser display.
- The profile shows that that >19s of CPU time were used during the 30s sampling period, from this single user traffic.
- The profile shows that over 92% of that CPU time were occupied by blowfish/encryptBlock, presumably repeatedly doing the expensive hashing of the user-password for each request:

profile001

Perhaps caddyauth/BcryptHash or some similar function could be memoized to cache results for a given password input and reduce the number of bcrypt calculations that must be performed for a logged in user submitting multiple requests with the same username/password.

mholt:
Cache capacity is currently hard-coded at 1000 with random eviction.

It is not enabled by default, but the Caddyfile _will_ enable it by default because I assume this is the most common preference.

I used this Caddyfile to test, but you can use anything, really:

```
:5555

basicauth {
	test JDJhJDEwJEdSbzFmZWcxOGZXc3k2Tm5vdE9OamVZWTFGTmQ2SUE2aGJaWDZIWFVGczVRMVpoUi56RjNL
}
respond "Yahaha! You found me!"
```

Please give it a try @mikelococo ! Thank you for the report.

Fixes #3462

### DIFF
--- a/modules/caddyhttp/caddyauth/basicauth.go
+++ b/modules/caddyhttp/caddyauth/basicauth.go
@@ -16,15 +16,21 @@ package caddyauth
 
 import (
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	weakrand "math/rand"
 	"net/http"
+	"sync"
+	"time"
 
 	"github.com/caddyserver/caddy/v2"
 )
 
 func init() {
 	caddy.RegisterModule(HTTPBasicAuth{})
+
+	weakrand.Seed(time.Now().UnixNano())
 }
 
 // HTTPBasicAuth facilitates HTTP basic authentication.
@@ -37,6 +43,17 @@ type HTTPBasicAuth struct {
 
 	// The name of the realm. Default: restricted
 	Realm string `json:"realm,omitempty"`
+
+	// If non-nil, a mapping of plaintext passwords to their
+	// hashes will be cached in memory (with random eviction).
+	// This can greatly improve the performance of traffic-heavy
+	// servers that use secure password hashing algorithms, with
+	// the downside that plaintext passwords will be stored in
+	// memory for a longer time (this should not be a problem
+	// as long as your machine is not compromised, at which point
+	// all bets are off, since basicauth necessitates plaintext
+	// passwords being received over the wire anyway).
+	HashCache *Cache `json:"hash_cache,omitempty"`
 
 	Accounts map[string]Account `json:"-"`
 	Hash     Comparer           `json:"-"`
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/caddyauth/basicauth.go
```diff
@@ -126,6 +146,43 @@ func (hba HTTPBasicAuth) Authenticate(w http.ResponseWriter, req *http.Request)
...
+		return hba.Hash.Compare(account.password, plaintextPassword, account.salt)
+	}
+
+	// if no caching is enabled, simply return the result of hashing + comparing
+	if hba.HashCache == nil {
+		return compare()
+	}
+
+	// compute a cache key that is unique for these input parameters
+	cacheKey := hex.EncodeToString(append(append(account.password, account.salt...), plaintextPassword...))
```
francislavoie, line 160: I'm super late on this, but I think what we should do is hash this string with sha256 or sha512 so that the password is not actually contained in the hash key written to memory.

mholt: That's probably fine, though Go's memory safety guarantees make it pretty difficult to exploit your way into this data.

francislavoie: Yeah, I realize. It's a bit of a "why not do it _just in case_" situation. Better to do it than not :man_shrugging:

mholt: But I only want to do it if it provides a _measurable_ or _justifiable_ security (or performance) win. The password is still in plaintext in other parts of memory, so...

francislavoie: But it's only in memory in other places for the duration of a single request. This keeps passwords around for the lifetime of the server. That's my main concern.

mholt: It's likely in memory longer than a single request. We don't know when or if that memory will ever be cleared. That's all handled by Go's runtime + the kernel.

PR #3472, opened 2020-06-03
Opt caddyauth error handle #3471

sharego:
Add authentication provider error delivery, to change server response


mholt:
Thank you for opening a PR -- unfortunately, this code does not compile, nor is the motivation well explained or understood. :( Closing for now. Can continue discussion in related issue.

PR #3483, opened 2020-06-08
ci: Fix gemfury upload condition, move triggers to publish event

francislavoie:
@mholt no need to retag for this one, but it should fix things for next time.


francislavoie:
Hmm. Looks like it didn't trigger after publishing.

francislavoie:
My best guess is the workflow needed to be in the release that is being published for it to trigger. I guess we wait until the next release to see if it worked?

PR #3491, opened 2020-06-12
fastcgi: Make sure splitPos handles empty SplitPath correctly

### LINKED ISSUE #3490: PATH_INFO is not set correctly when SplitPath is empty
## Background

I use caddy to host a fairly custom setup of cgit that runs via [slowcgi](https://man.openbsd.org/slowcgi.8) (a FastCGI wrapper). Recently I began looking into migrating to Caddy 2 and set up a test environment for it. I managed to migrate the Caddyfile, but sadly could not get my setup to work without one *rather awkward* change...

*Note:* v1 refers to 1.0.4, v2 refers to 2.0.0 and current master.

## Configuration

For the purposes of this bug report, a minimal Caddyfile that works as expected with v1:

```
localhost:2001 {
	fastcgi / /run/slowcgi.sock {
		env SCRIPT_FILENAME /bin/cgit
	}
}
```

And one for v2 that doesn't:

```
http://localhost:2001 {
	reverse_proxy unix//run/slowcgi.sock {
		transport fastcgi {
			env SCRIPT_FILENAME /bin/cgit
		}
	}
}
```

## Behaviour

Browsing to `localhost:2001/foo/bar/baz` and inspecting the environment variables passed to the CGI script gives the following:

| Version | PATH_INFO | SCRIPT_NAME |
| --- | --- | --- |
| v1 | /foo/bar/baz | / |
| v2 |  | /foo/bar/baz |

For some reason, on v2, `PATH_INFO` remains empty. Instead, its contents end up in `SCRIPT_NAME`.

## Workaround

At this point, I had a weird inkling and decided to try out something strange. The following Caddyfile works as expected with v2:

```
http://localhost:2001 {
	reverse_proxy unix//run/slowcgi.sock {
		transport fastcgi {
			split ""
			env SCRIPT_FILENAME /bin/cgit
		}
	}
}
```

Now, `PATH_INFO` contains `/foo/bar/baz` and `SCRIPT_NAME` is empty.

## Root cause

I dug around in the codebase a bit to try and find the cause for this. In https://github.com/caddyserver/caddy/commit/f2ce81cc8b5adcd453566d292ac0b0b47fc64e99 support for multiple path splitters was added. This [changed](https://github.com/caddyserver/caddy/commit/f2ce81cc8b5adcd453566d292ac0b0b47fc64e99#diff-b3eb55bbe16c5572513e0d685dd3b2c9R54) the type of `SplitPath` from `string` to `[]string`.

Previously, `SplitPath` was initialized to `""` (the empty string). `splitPos` returns 0 here:

https://github.com/caddyserver/caddy/blob/2cab475ba516fa725d012f53ca417c3e039607de/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go#L291

Now, however, it is initalized to `[]`, the empty list. Because of this, the for loop in `splitPos` is skipped entirely, returning -1:
https://github.com/caddyserver/caddy/blob/d55c3b31ebb77df65cc052dbddc137cbe07b297e/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go#L307-L312

In turn, the code that sets `pathInfo` is never reached here:
https://github.com/caddyserver/caddy/blob/d55c3b31ebb77df65cc052dbddc137cbe07b297e/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go#L186-L191

## Conclusion

This subtle change in behaviour goes unmentioned in the commit message for https://github.com/caddyserver/caddy/commit/f2ce81cc8b5adcd453566d292ac0b0b47fc64e99. Therefore I'm assuming it was not intended. The parent commit https://github.com/caddyserver/caddy/commit/2cab475ba516fa725d012f53ca417c3e039607de works fine.

wylfen:
In commit f2ce81c, support for multiple path splitters was added. The
type of SplitPath changed from string to []string, and splitPos was
changed to loop through all values in SplitPath.

Before that commit, if SplitPath was empty, strings.Index returned 0 and
PATH_INFO was set correctly in buildEnv.

Currently, however, splitPos returns -1 for empty values of SplitPath,
behaving as if a split position could not be found at all. PATH_INFO is
then never set in buildEnv and remains empty.

Restore the old behaviour by explicitly checking whether SplitPath is
empty and returning 0 in splitPos.

Closes #3490

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
@@ -303,6 +303,10 @@ func (t Transport) splitPos(path string) int {
 	// if httpserver.CaseSensitivePath {
 	// 	return strings.Index(path, r.SplitPath)
 	// }
+	if len(t.SplitPath) == 0 {
+		return 0
+	}
+
 	lowerPath := strings.ToLower(path)
 	for _, split := range t.SplitPath {
 		if idx := strings.Index(lowerPath, strings.ToLower(split)); idx > -1 {


mholt:
Yes, we wish for more bug reports and contributions following this pattern. 🙂  Thank you!

PR #3494, opened 2020-06-12
ci: don't run s390x tests on PRs of forks

Mohammed90:
Seems like #3494 wasn't successful, so I'm tackling it again from a fork to properly test it.


Mohammed90:
For posterity... the trick is found here:
https://github.community/t/dont-run-actions-on-pull-request-from-fork/17560/4

It traverses the webhook payload structure from here:
https://developer.github.com/webhooks/event-payloads/#pull_request

PR #3498, opened 2020-06-15
templates: Add support for dots to close yaml frontmatter

### LINKED ISSUE #3497: splitFrontMatter fails when YAML metadata is terminated with ...
When using YAML front matter in the following form:

```markdown
---
title: Hello world!
author: Greg Anders
...
```

Caddy's template engine throws the following error:

```
executing "/" at <splitFrontMatter>: error calling splitFrontMatter: unterminated front matter 
```

While less common, using `...` as the closing delimiter for YAML front matter is valid [as defined by the spec](https://yaml.org/spec/1.0/#l-l-after-document) and is also used in e.g. Pandoc.

(I personally prefer to use the `...` closing syntax as it degrades more gracefully than `---` in Markdown parsers that don't support YAML front matter.)

francislavoie:
Fixes #3497

Small refactor to the template frontmatter extraction code: 
- Now generalizes a frontmatter type into a struct which defines the fence delimiters and the parse function.
- Now loops over the supported types list instead of using a switch to detect the frontmatter type.
- Each type can now have more than one fence closing string, each are searched for in order they appear (currently only yaml supports more than one).

### DIFF
--- a/modules/caddyhttp/templates/frontmatter.go
+++ b/modules/caddyhttp/templates/frontmatter.go
@@ -34,25 +34,28 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 	firstLine = strings.TrimSpace(firstLine)
 
 	// see what kind of front matter there is, if any
-	var closingFence string
+	var closingFence []string
 	var fmParser func([]byte) (map[string]interface{}, error)
-	switch firstLine {
-	case yamlFrontMatterFenceOpen:
-		fmParser = yamlFrontMatter
-		closingFence = yamlFrontMatterFenceClose
-	case tomlFrontMatterFenceOpen:
-		fmParser = tomlFrontMatter
-		closingFence = tomlFrontMatterFenceClose
-	case jsonFrontMatterFenceOpen:
-		fmParser = jsonFrontMatter
-		closingFence = jsonFrontMatterFenceClose
-	default:
+	for _, fmType := range supportedFrontMatterTypes {
+		if firstLine == fmType.FenceOpen {
+			closingFence = fmType.FenceClose
+			fmParser = fmType.ParseFunc
+		}
+	}
+
+	if fmParser == nil {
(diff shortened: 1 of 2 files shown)


gpanders:
The error I mentioned [here](https://github.com/caddyserver/caddy/issues/3497#issuecomment-643858837) is due to line 72:

```go
body := input[fmEndFenceStart+len(closingFence):]
```

`closingFence` is now an array, so `len(closingFence)` gives 2 for YAML and 1 for TOML or JSON, instead of providing the actual length of the closing fence.

The following patch is a possible fix:
```patch
diff --git a/modules/caddyhttp/templates/frontmatter.go b/modules/caddyhttp/templates/frontmatter.go
index 337a4183..cddc0a38 100644
--- a/modules/caddyhttp/templates/frontmatter.go
+++ b/modules/caddyhttp/templates/frontmatter.go
@@ -50,10 +50,12 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 
 	// find end of front matter
 	fmEndFenceStart := -1
+	var fmEndFence string
 	for _, fence := range closingFence {
 		index := strings.Index(input[firstLineEnd:], "\n"+fence)
 		if index >= 0 {
 			fmEndFenceStart = index
+			fmEndFence = fence
 		}
 	}
 	if fmEndFenceStart < 0 {
@@ -69,7 +71,7 @@ func extractFrontMatter(input string) (map[string]interface{}, string, error) {
 	}
 
 	// the rest is the body
-	body := input[fmEndFenceStart+len(closingFence):]
+	body := input[fmEndFenceStart+len(fmEndFence):]
 
 	return fm, body, nil
 }
```

francislavoie:
Haha, I only noticed your comment after making the fix, I did the same fix independently but used a different variable name, but I liked yours more 👍 

Also updated the tests to also compare the body to make sure we don't break this again in the future.

`len()` working for both strings and arrays is always a tricky one when refactoring 😛

PR #3523, opened 2020-06-25
ci: Apparently only single-quote strings are supported

francislavoie:
https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#literals

https://github.com/caddyserver/caddy/actions/runs/147953515


mholt:
Heh, thanks, both!

PR #3526, opened 2020-06-27
cel: fix validation of expression result type

Mohammed90:
This Caddy JSON config is rejected with the error `loading http app module: provision http: server srv0: setting up route handlers: route 0: loading handler modules: position 0: loading module 'subroute': provision http.handlers.subroute: setting up subroutes: route 1: loading matcher modules: module name 'not': provision http.matchers.not: loading matcher sets: module name 'expression': provision http.matchers.expression: CEL request matcher expects return type of bool, not primitive:BOOL`:

```json
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"match": [
								{
									"expression": "{http.request.uri.query} != ''"
								}
							],
							"handle": [
								{
									"handler": "static_response",
									"body": "Hello"
								}
							]
						}
					]
				}
			}
		}
	}
}
```

Checking the types with a few log lines prints (manually prettified):
```
&expr.Type{
	state:impl.MessageState{
		NoUnkeyedLiterals:pragma.NoUnkeyedLiterals{}, 
		DoNotCompare:pragma.DoNotCompare{}, 
		DoNotCopy:pragma.DoNotCopy{}, 
		atomicMessageInfo:(*impl.MessageInfo)(0xc000114138)
	}, 
	sizeCache:0, 
	unknownFields:[]uint8(nil), 
	TypeKind:(*expr.Type_Primitive)(0xc00036f400)
}

&expr.Type{
	state:impl.MessageState{
		NoUnkeyedLiterals:pragma.NoUnkeyedLiterals{}, 
		DoNotCompare:pragma.DoNotCompare{}, 
		DoNotCopy:pragma.DoNotCopy{}, 
		atomicMessageInfo:(*impl.MessageInfo)(0xc000114138)
	},
	sizeCache:0, 
	unknownFields:[]uint8(nil), 
	TypeKind:(*expr.Type_Primitive)(0xc00036f400)
}
```
The types appear to be exactly the same. However, the `proto.Equal()` method prints:
`proto: don't know how to compare []`

The earlier code used the proto.Equals from github.com/gogo/protobuf, which failed to compare two messages of the same type for some reason. Switching to proto.Equal from the canonical github.com/golang/protobuf fixes the issue.

Reported in the forum: https://caddy.community/t/example-configure-wordpress-with-a-static-cache/8215/7


modules/caddyhttp/celmatcher.go
```diff
@@ -25,7 +25,7 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
-	"github.com/gogo/protobuf/proto"
+	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/timestamp"
```
Mohammed90, line 29: Note that the package github.com/golang/protobuf/proto is deprecated. We're instructed to use the "google.golang.org/protobuf/proto" package instead. I didn't want to change it now to avoid intrusive changes whose consequences are, currently, unknown.

mholt: I wonder if we should, though, since deprecated packages spells lack of updates and fixes?

Mohammed90: Done. For a second I was worried this will break the code because cel-go uses the deprecated package `github.com/golang/protobuf/ptypes/timestamp`. I checked the code in `github.com/golang/protobuf/ptypes/timestamp` and found[ it uses type aliases](https://github.com/golang/protobuf/blob/d04d7b157bb510b1e0c10132224b616ac0e26b17/ptypes/timestamp/timestamp.pb.go#L15), so it will all work fine. Kinda cool the type alias works perfectly [as intended and argued for](https://talks.golang.org/2016/refactor.article).

PR #3542, opened 2020-07-02
[file-server] Add "from" parameter to browse complementary to "limit"

snu-ceyda:
A simple update to the **browse** functionality within the **file-server** 

- Added an optional **from** URL parameter complementary to the existing **limit** parameter.

Use case -> To efficiently browse folders with many... many... files

## Usage

caddy file-server --listen 8000 --browse

http://localhost:8000/browse
Shows all files

http://localhost:8000/browse?limit=10
Shows only files 10 files starting from 0

http://localhost:8000/browse?limit=10&from=5
Shows only files 10 files starting from 5

### DIFF
--- a/modules/caddyhttp/fileserver/browse.go
+++ b/modules/caddyhttp/fileserver/browse.go
@@ -106,6 +106,7 @@ func (fsrv *FileServer) browseApplyQueryParams(w http.ResponseWriter, r *http.Re
 	sortParam := r.URL.Query().Get("sort")
 	orderParam := r.URL.Query().Get("order")
 	limitParam := r.URL.Query().Get("limit")
+	fromParam := r.URL.Query().Get("from")
 
 	// first figure out what to sort by
 	switch sortParam {
@@ -130,7 +131,7 @@ func (fsrv *FileServer) browseApplyQueryParams(w http.ResponseWriter, r *http.Re
 	}
 
 	// finally, apply the sorting and limiting
-	listing.applySortAndLimit(sortParam, orderParam, limitParam)
+	listing.applySortAndLimit(sortParam, orderParam, limitParam, fromParam)
 }
 
 func (fsrv *FileServer) browseWriteJSON(listing browseListing) (*bytes.Buffer, error) {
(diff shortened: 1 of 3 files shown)


mholt:
Thanks for the PR! I'll try to give it a pass soon.

snu-ceyda:
> I think `offset` would be better than `from`.
that's reasonable too, should I change it?

modules/caddyhttp/fileserver/browse.go
```diff
@@ -106,6 +106,7 @@ func (fsrv *FileServer) browseApplyQueryParams(w http.ResponseWriter, r *http.Re
 	sortParam := r.URL.Query().Get("sort")
 	orderParam := r.URL.Query().Get("order")
 	limitParam := r.URL.Query().Get("limit")
+	fromParam := r.URL.Query().Get("from")
```
mholt, line 109: As discussed, change to `offset`.
```suggestion
	offsetParam := r.URL.Query().Get("offset")
```

modules/caddyhttp/fileserver/browselisting.go
```diff
@@ -101,6 +101,9 @@ type browseListing struct {
 
 	// If ≠0 then Items have been limited to that many elements.
 	ItemsLimitedTo int
+
+	// If ≠0 then Items starting from that many elements.
+	ItemsFrom int
```
mholt, line 106:
```suggestion
	ItemOffset int
```

modules/caddyhttp/fileserver/browselisting.go
```diff
@@ -131,7 +134,7 @@ func (l browseListing) Breadcrumbs() []crumb {
 	return result
 }
 
-func (l *browseListing) applySortAndLimit(sortParam, orderParam, limitParam string) {
+func (l *browseListing) applySortAndLimit(sortParam, orderParam, limitParam string, fromParam string) {
```
mholt, line 137:
```suggestion
func (l *browseListing) applySortAndLimit(sortParam, orderParam, limitParam string, offsetParam string) {
```

modules/caddyhttp/fileserver/browselisting.go
snu-ceyda: Also changed the browse template file to 'offset'.

PR #3549, opened 2020-07-05
caddyfile: Expose Tokenize function for modules to use the lexer

francislavoie:
As requested by @lucaslorentz in this comment https://github.com/lucaslorentz/caddy-docker-proxy/pull/160/files#r449789804

This exposes a `Tokenize` function which was only in the lexer test previously to do the work of lexing bytes as input to get the list of tokens.

One quirk that I'm not sure how best to handle is the `filename` param. In other languages I'd make that an optional param, but Go doesn't have that, so I'm just using `""` as the empty value to keep `File` nil.

### DIFF
--- a/caddyconfig/caddyfile/lexer.go
+++ b/caddyconfig/caddyfile/lexer.go
@@ -16,6 +16,7 @@ package caddyfile
 
 import (
 	"bufio"
+	"bytes"
 	"io"
 	"unicode"
 )
@@ -167,4 +168,22 @@ func (l *lexer) next() bool {
 
 		val = append(val, ch)
 	}
+}
+
+// Tokenize takes bytes as input and lexes it into
+// a list of tokens that can be parsed as a Caddyfile.
+// Also takes a filename to fill the token's File as
+// the source of the tokens, which is important to
+// determine relative paths for `import` directives.
+func Tokenize(input []byte, filename string) ([]Token, error) {
+	l := lexer{}
+	if err := l.load(bytes.NewReader(input)); err != nil {
+		return nil, err
+	}
+	var tokens []Token
+	for l.next() {
+		l.token.File = filename
+		tokens = append(tokens, l.token)
+	}
+	return tokens, nil
 }
--- a/caddyconfig/caddyfile/lexer_test.go
+++ b/caddyconfig/caddyfile/lexer_test.go
@@ -15,37 +15,35 @@
 package caddyfile
 
 import (
-	"log"
-	"strings"
 	"testing"
 )
 
 type lexerTestCase struct {
-	input    string
+	input    []byte
 	expected []Token
 }
 
 func TestLexer(t *testing.T) {
 	testCases := []lexerTestCase{
 		{
-			input: `host:123`,
+			input: []byte(`host:123`),
 			expected: []Token{
 				{Line: 1, Text: "host:123"},
 			},
 		},
 		{
-			input: `host:123
+			input: []byte(`host:123
 
-					directive`,
+					directive`),
 			expected: []Token{
 				{Line: 1, Text: "host:123"},
 				{Line: 3, Text: "directive"},
 			},
 		},
 		{
-			input: `host:123 {
+			input: []byte(`host:123 {
 						directive
-					}`,
+					}`),
 			expected: []Token{
 				{Line: 1, Text: "host:123"},
 				{Line: 1, Text: "{"},
@@ -54,7 +52,7 @@ func TestLexer(t *testing.T) {
 			},
 		},
 		{
-			input: `host:123 { directive }`,
+			input: []byte(`host:123 { directive }`),
 			expected: []Token{
 				{Line: 1, Text: "host:123"},
 				{Line: 1, Text: "{"},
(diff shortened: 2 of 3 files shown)


caddyconfig/caddyfile/lexer.go
```diff
@@ -168,3 +169,23 @@ func (l *lexer) next() bool {
...
+func Tokenize(input []byte, filename string) ([]Token, error) {
+	l := lexer{}
+	if err := l.load(bytes.NewReader(input)); err != nil {
+		return nil, err
+	}
+	var tokens []Token
+	for l.next() {
+		if filename != "" {
+			l.token.File = filename
+		}
```
mholt, line 187: Doesn't this just happen anyway? (regardless of `if`)

caddyconfig/caddyfile/lexer.go
```diff
@@ -168,3 +169,23 @@ func (l *lexer) next() bool {
 		val = append(val, ch)
 	}
 }
+
+// Tokenize takes bytes as input and lexes it into
+// a list of tokens that can be parsed as a Caddyfile.
+// Also takes a filename to fill the token's File as
+// the source of the tokens, which is important to
+// determine relative paths for `import` directives.
+func Tokenize(input []byte, filename string) ([]Token, error) {
```
mholt, line 178: Any good reason we need to change from `string` to `[]byte` here? Now all the tests have to change...

caddyconfig/caddyfile/lexer.go
```diff
@@ -168,3 +169,23 @@ func (l *lexer) next() bool {
 		val = append(val, ch)
 	}
 }
+
+// Tokenize takes bytes as input and lexes it into
+// a list of tokens that can be parsed as a Caddyfile.
+// Also takes a filename to fill the token's File as
+// the source of the tokens, which is important to
+// determine relative paths for `import` directives.
```
mholt, line 177: Add a scary warning that this function is experimental and is not guaranteed to be stable.

caddyconfig/caddyfile/lexer.go
francislavoie: Then they would be empty string instead of nil which seems like a relevant difference, no?

If not then sure I can remove the condition (or feel free to fix the PR yourself since I'm away for a bit)

caddyconfig/caddyfile/lexer.go
francislavoie: Because parser.go uses a []byte instead. I tried the other way but converting the []byte back to string doesn't seem like a good idea.

caddyconfig/caddyfile/lexer.go
francislavoie: I don't think it's experimental... It's the exact same code that was in parser.go

caddyconfig/caddyfile/lexer.go
mholt: `filename` is a string, it can't be nil.

caddyconfig/caddyfile/lexer.go
mholt: Whether it's experimental or not, we need to say that it's not stable ("experimental" is a good consistent term) because I wasn't ever expecting to export this API when I designed it.

caddyconfig/caddyfile/lexer.go
mholt: Fair enough

caddyconfig/caddyfile/lexer.go
francislavoie: I'm not sure I understand how it wouldn't be stable if it's actually used by the Caddyfile parser itself.

mholt: I just mean that right now we can change it whenever we want, because we know it's unexported and nobody else is using it. I just try to avoid exporting things.

lucaslorentz: As I mentioned in the other PR, this is not a must-have.
We can give it some time and see how caddy-docker-proxy code will evolve.
And also see if the lexer will change that often in caddy

mholt: @lucaslorentz Is `Tokenize()` the right function to export? Or do you need env var replacements, I can't remember. If you need env var replacements, maybe allTokens is better: https://github.com/caddyserver/caddy/pull/3549/files#diff-06ecc7b9618d6f991e3d8846d817e0dcR85

francislavoie: I'm pretty certain we don't want env var replacement to happen because the step that caddy-docker-proxy takes is just rebuilding a valid Caddyfile from the labels the user inputs. The env vars can be replaced later when Caddy actually runs. If env var replacement was done at that stage, then it would use the env of the controller, which might not be the correct stuff.

lucaslorentz: I don't need env var replacements.

I have my own in memory representation of caddyfile, that doesn't depend plugins structs.
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/caddyfile/caddyfile.go

I also have a tiny parser, that parses all tokens into my in memory caddyfile struct:
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/caddyfile/marshal.go#L164

Currently, the tokenization is an overlap between caddy implementation and my own implementation. Thus, we're considering exposing that feature.

But, another option, is to expose in caddy a function similar to the existing parseCaddyfile, but that keeps env variables and snippets untouched. Then I could change my parser to read siteBlocks returned by parseCaddyfile and tokens from each siteBlock.

mholt: I see, okay. Thanks!

Issue #3556, opened 2020-07-07
H2 and H2C bi-direction streaming reverse proxy need non-nil req.body and flush headers before copyResponse


masknu:
I found that haproxy is capable of proxying h2 clients to h2c upstream servers, but caddy's not working even with merged h2c branch #3289.  For instance, client and server I used were v2ray(h2 outbound and h2c inbound).
After some digging and testing, I found the cause and have made caddy work in this scenario.

**First**, HTTP2 allows bi-direction streaming, and not using Transfer-Encoding and Content-Length headers for this. Thus we probably need to keep req.Body valid before invoking RoundTrip to upstream.
https://github.com/caddyserver/caddy/blob/7a99835dab64f7864186185761bbf5194216f8b6/modules/caddyhttp/reverseproxy/reverseproxy.go#L424

```golang
        req.Body = originBody
	start := time.Now()
	res, err := h.Transport.RoundTrip(req)
```

https://github.com/caddyserver/caddy/blob/7a99835dab64f7864186185761bbf5194216f8b6/modules/caddyhttp/reverseproxy/reverseproxy.go#L500

**Second**, in order to signal h2 clients begin sending streaming body, we need to flush response headers to client before copyResponse from res.Body to http.ResponseWriter. And use -1 as flushInterval for http2 connections.

```golang
	rw.WriteHeader(res.StatusCode)

	if h2 {
		if wf, ok := rw.(http.Flusher); ok {
			wf.Flush()
		}
		err = h.copyResponse(rw, res.Body, -1)
	} else {
		err = h.copyResponse(rw, res.Body, h.flushInterval(req, res))
	}
```

https://github.com/caddyserver/caddy/blob/7a99835dab64f7864186185761bbf5194216f8b6/modules/caddyhttp/reverseproxy/reverseproxy.go#L605

The config file I used:

```json
{
    "logging": {
        "logs": {
            "default": {
                "level": "DEBUG"
            }
        }
    },
    "apps": {
        "http": {
            "servers": {
                "srv0": {
                    "listen": [
                        ":443"
                    ],
                    "routes": [
                        {
                            "match": [
                                {
                                    "host": [
                                        "my.org"
                                    ]
                                }
                            ],
                            "handle": [
                                {
                                    "handler": "subroute",
                                    "routes": [
                                        {
                                            "handle": [
                                                {
                                                    "handler": "vars",
                                                    "root": "/usr/share/caddy"
                                                }
                                            ]
                                        },
                                        {
[...]

PR #3561, opened 2020-07-08
reverse proxy: support some apps like v2ray which needs response head…

masknu:
Support some apps like v2ray which needs response header before initiate streaming data (#3556)

### DIFF
--- a/caddytest/integration/stream_test.go
+++ b/caddytest/integration/stream_test.go
@@ -1,0 +1,201 @@
+package integration
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/caddyserver/caddy/v2/caddytest"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
+)
+
+func TestH2ToH2CStream(t *testing.T) {
+	tester := caddytest.NewTester(t)
+	tester.InitServer(` 
+  {
+    "apps": {
+      "http": {
+        "http_port": 9080,
+        "https_port": 9443,
+        "servers": {
+          "srv0": {
+            "listen": [
+              ":9443"
+            ],
+            "routes": [
+              {
+                "handle": [
+                  {
+                    "handler": "reverse_proxy",
+                    "transport": {
+                      "protocol": "http",
+                      "compression": false,
+                      "versions": [
+                        "h2c",
+                        "2"
+                      ]
+                    },
+                    "upstreams": [
+                      {
+                        "dial": "localhost:54321"
+                      }
+                    ]
+                  }
+                ],
+                "match": [
+                  {
+                    "path": [
+                      "/tov2ray"
+                    ]
+                  }
+                ]
+              }
+            ],
+            "tls_connection_policies": [
+              {
+                "certificate_selection": {
+                  "any_tag": ["cert0"]
+                },
+                "default_sni": "a.caddy.localhost"
+              }
+            ]
+          }
+        }
+      },
+      "tls": {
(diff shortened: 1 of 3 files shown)


mholt:
Thanks for this PR! Looks simple enough. I'd feel more comfortable merging it soon if it had a self-contained unit test or integration test case. Do you think you could add one that demonstrates this is effective (and won't break other connections)?

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -604,6 +604,13 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, di Dia
 
 	rw.WriteHeader(res.StatusCode)
 
+	// for some apps which need a correct response header to start http2 streaming,
+	// it's important to explicit flush headers to the client before copy streaming data.
```
francislavoie, line 608: I think this might read better:

> some apps need the response headers before starting to stream content with http2, so it's important to explicitly flush the headers to the client before streaming the data.

masknu:
Sure, just added an integration test, and adjusted the comment. :innocent:

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -96,6 +96,11 @@ func (h Handler) flushInterval(req *http.Request, res *http.Response) time.Durat
 		return -1 // negative means immediately
 	}
 
+	// for h2 and h2c upstream streaming data to client
```
mholt, line 99:
```suggestion
	// for h2 and h2c upstream streaming data to client (issue #3556)
```

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -604,6 +604,13 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, di Dia
 
 	rw.WriteHeader(res.StatusCode)
 
+	// some apps need the response headers before starting to stream content with http2,
+	// so it's important to explicitly flush the headers to the client before streaming the data.
```
mholt, line 608:
```suggestion
	// some apps need the response headers before starting to stream content with http2,
	// so it's important to explicitly flush the headers to the client before streaming the data.
	// (see https://github.com/caddyserver/caddy/issues/3556 for use case)
```

caddytest/integration/stream_test.go
```diff
@@ -0,0 +1,201 @@
...
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Nanosecond)
+		defer cancel()
+		server.Shutdown(ctx)
+	}()
+
+	r, w := io.Pipe()
+	req := &http.Request{
+		Method: "PUT",
+		// Host:   httpSettings.getRandomHost(),
```
mholt, line 110: Can remove this probably?

caddytest/integration/stream_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/caddyserver/caddy/v2/caddytest"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
+)
+
+func TestH2ToH2CStream(t *testing.T) {
```
mholt, line 20: If you could add a comment with a link to https://github.com/caddyserver/caddy/issues/3556, I think that would be helpful if this is revisited in the future!

masknu:
Updated comments.
Yes, this would be a very helpful feature of caddy, thank you for your approving.
(resolves v2ray/discussion#759, resolves v2ray/discussion#749, resolves v2ray/discussion#296)

francislavoie:
You can use v2.2.0-rc.1 which has these changes.

ha-ku:
Thanks!

PR #3570, opened 2020-07-11
fastcgi: Comply with RFC3875, ensure a leading slash on index rewrite

### LINKED ISSUE #3568: Don’t set SERVER_PORT to empty string (reverse proxy)
I ran into an issue with a PHP framework that got confused about `SERVER_PORT` being set to an empty string.

Assuming this framework has been used with many other web servers, it implies that the de facto standard is to either have a value, or not set it at all, which I therefore propose that Caddy does as well.

That said, I did open an issue with the PHP framework in question (bcosca/fatfree#1194) as I believe their code should be updated, but the problem may also affect other frameworks, which is why I am also opening an issue with Caddy.

### LINKED ISSUE #3569: FastCGI: SCRIPT_NAME must have leading slash (or be empty)
I have `/index.php` that serves requests for non-existing files (custom router).

When requesting `/` then `SCRIPT_NAME` is correctly set to `/index.php`, but when requesting e.g.  `/about` then `SCRIPT_NAME` is `index.php` (notice that there is no leading slash).

According to [RFC3875](https://tools.ietf.org/html/rfc3875#section-4.1.13) then `SCRIPT_NAME` *must* either be empty, or have a leading slash:

    The SCRIPT_NAME variable MUST be set to a URI path (not URL-encoded)
    which could identify the CGI script (rather than the script's
    output).  The syntax is the same as for PATH_INFO (section 4.1.5)

       SCRIPT_NAME = "" | ( "/" path )

    The leading "/" is not part of the path.  It is optional if the path
    is NULL; however, the variable MUST still be set in that case.

    The SCRIPT_NAME string forms some leading part of the path component
    of the Script-URI derived in some implementation-defined manner.  No
    PATH_INFO segment (see section 4.1.5) is included in the SCRIPT_NAME
    value.

It is causing a compatibility problem with a [PHP framework](https://github.com/bcosca/fatfree) that calls `dirname` on the value of `SCRIPT_NAME`.

francislavoie:
Closes #3568
Closes #3569 

See https://tools.ietf.org/html/rfc3875#section-4.1.13 for SCRIPT_NAME requiring leading slash
See https://tools.ietf.org/html/rfc3875#section-4.1.15 for SERVER_PORT requiring omission if empty

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
@@ -202,6 +202,12 @@ func (t Transport) buildEnv(r *http.Request) (map[string]string, error) {
 	pathPrefix, _ := r.Context().Value(caddy.CtxKey("path_prefix")).(string)
 	scriptName = path.Join(pathPrefix, scriptName)
 
+	// Ensure the SCRIPT_NAME has a leading slash for compliance with RFC3875
+	// Info: https://tools.ietf.org/html/rfc3875#section-4.1.13
+	if scriptName != "" && !strings.HasPrefix(scriptName, "/") {
+		scriptName = "/" + scriptName
+	}
+
 	// Get the request URL from context. The context stores the original URL in case
 	// it was changed by a middleware such as rewrite. By default, we pass the
 	// original URI in as the value of REQUEST_URI (the user can overwrite this
@@ -244,7 +250,6 @@ func (t Transport) buildEnv(r *http.Request) (map[string]string, error) {
 		"REQUEST_METHOD":    r.Method,
 		"REQUEST_SCHEME":    requestScheme,
 		"SERVER_NAME":       reqHost,
-		"SERVER_PORT":       reqPort,
 		"SERVER_PROTOCOL":   r.Proto,
 		"SERVER_SOFTWARE":   t.serverSoftware,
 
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -237,7 +237,7 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
 		// route to rewrite to PHP index file
 		rewriteMatcherSet := caddy.ModuleMap{
 			"file": h.JSON(fileserver.MatchFile{
-				TryFiles:  []string{"{http.request.uri.path}", "{http.request.uri.path}/" + indexFile, indexFile},
+				TryFiles:  []string{"{http.request.uri.path}", "{http.request.uri.path}/" + indexFile, "/" + indexFile},
```
mholt, line 240: Am I reading this right, that it'll be, e.g. `/foo/bar/index.php/index.php`?

francislavoie: There's a comma there, so no. It's `/index.php`. The change is to just ensure the index file has a `/` prefixed.

But this entire change can probably be thrown away because of the other change in `fastcgi.go` that ensures it at a lower level.

mholt: Oh, duh -- thanks. I need more sleep. :eyes:

> because of the other change in fastcgi.go that ensures it at a lower level.

You mean this one?

https://github.com/caddyserver/caddy/pull/3570/files#diff-b3eb55bbe16c5572513e0d685dd3b2c9R205-R210

I do think that is the better way to solve it, no need for both I think.

francislavoie: Yep. Okay I'll remove the Caddyfile change.

ttys3:
the implementation has a problem.

```
4.1.15.  SERVER_PORT

   The SERVER_PORT variable MUST be set to the TCP/IP port number on
   which this request is received from the client.  This value is used
   in the port part of the Script-URI.

      SERVER_PORT = server-port
      server-port = 1*digit

   Note that this variable MUST be set, even if the port is the default
   port for the scheme and could otherwise be omitted from a URI.
```

**Note that this variable MUST be set**

PR #3573, opened 2020-07-11
push: Implement HTTP/2 server push

### LINKED ISSUE #3551: HTTP2 Push missing in Caddy 2?
I've noticed while migrating from v1 to v2 that http2 push feature is not available in Caddy v2.  
Are there any plans to bring it back?

mholt:
Implement HTTP/2 server push. Borrowed much of the code from v1, but the standard lib does most of the hard work. All I had to do was expose the config surface.

Push is a new handler in the standard HTTP chain. Caddy 1 supported server push but it wasn't a very common feature so it wasn't a dev priority for the initial release of v2. Well, here it is now.

Supports the Caddyfile, too:

```
# basically just pushes using Link headers from upstream
push

# simple case, push a style sheet for the homepage
push /index.html /style.css

# multiple resources
push /index.html {
    /resource1.css
    /resource2.jpg
}

# you can customize methods and headers of synthetic push requests
push /example.html {
    GET /foo.css
    HEAD /bar.jpg
    headers {
        Authorization {header.Authorization}
    }
}
```

For now, see the code comments for more info about syntax and capabilities.

We copy some essential headers such as Accept and Cache-Control by default, but that's all (for now). We want to be conservative until server push deployments are better understood.  The user can configure pretty much everything about the headers in the meantime. It's still early for server push and there's a lot of room for innovation here, and Caddy is the ideal place for that IMO. This will lay the foundation for that.

/cc @dunglas

closes #3551

### DIFF
--- a/modules/caddyhttp/push/handler.go
+++ b/modules/caddyhttp/push/handler.go
@@ -91,20 +91,19 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhtt
 		}
 	}
 
-	// serve only after pushing!
-	if err := next.ServeHTTP(w, r); err != nil {
-		return err
+	// wrap the response writer so that we can initiate push of any resources
+	// described in Link header fields before the response is written
+	lp := linkPusher{
+		ResponseWriterWrapper: &caddyhttp.ResponseWriterWrapper{ResponseWriter: w},
+		handler:               h,
+		pusher:                pusher,
+		header:                hdr,
+		request:               r,
 	}
 
-	// finally, push any resources described by Link fields that were
-	// written to the response header, only if another push handler
-	// hasn't already done so
-	if links, ok := w.Header()["Link"]; ok {
-		if val := caddyhttp.GetVar(r.Context(), pushedLink); val == nil {
-			h.logger.Debug("pushing Link resources", zap.Strings("linked", links))
-			caddyhttp.SetVar(r.Context(), pushedLink, true)
-			h.servePreloadLinks(pusher, hdr, links)
-		}
+	// serve only after pushing!
+	if err := next.ServeHTTP(lp, r); err != nil {
+		return err
 	}
 
 	return nil
@@ -141,7 +140,6 @@ func (h Handler) initializePushHeaders(r *http.Request, repl *caddy.Replacer) ht
 // attribute or describes an external entity (meaning, the resource
 // URI includes a scheme), it will not be pushed.
 func (h Handler) servePreloadLinks(pusher http.Pusher, hdr http.Header, resources []string) {
-outer:
 	for _, resource := range resources {
 		for _, resource := range parseLinkHeader(resource) {
 			if _, ok := resource.params["nopush"]; ok {
@@ -154,7 +152,7 @@ outer:
 				Header: hdr,
 			})
 			if err != nil {
-				break outer
+				return
 			}
 		}
 	}
@@ -173,6 +171,30 @@ type Resource struct {
 // HeaderConfig configures headers for synthetic push requests.
 type HeaderConfig struct {
 	headers.HeaderOps
+}
+
+// linkPusher is a http.ResponseWriter that intercepts
+// the WriteHeader() call to ensure that any resources
+// described by Link response headers get pushed before
+// the response is allowed to be written.
+type linkPusher struct {
+	*caddyhttp.ResponseWriterWrapper
+	handler Handler
+	pusher  http.Pusher
+	header  http.Header
+	request *http.Request
+}
+
+func (lp linkPusher) WriteHeader(statusCode int) {
+	if links, ok := lp.ResponseWriter.Header()["Link"]; ok {
+		// only initiate these pushes if it hasn't been done yet
+		if val := caddyhttp.GetVar(lp.request.Context(), pushedLink); val == nil {
+			lp.handler.logger.Debug("pushing Link resources", zap.Strings("linked", links))
+			caddyhttp.SetVar(lp.request.Context(), pushedLink, true)
+			lp.handler.servePreloadLinks(lp.pusher, lp.header, links)
+		}
+	}
+	lp.ResponseWriter.WriteHeader(statusCode)
 }
 
 // isRemoteResource returns true if resource starts with
@@ -210,4 +232,5 @@ const pushedLink = "http.handlers.push.pushed_link"
 var (
 	_ caddy.Provisioner           = (*Handler)(nil)
 	_ caddyhttp.MiddlewareHandler = (*Handler)(nil)
+	_ caddyhttp.HTTPInterfaces    = (*linkPusher)(nil)
 )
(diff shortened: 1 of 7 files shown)


modules/caddyhttp/push/handler.go
```diff
@@ -0,0 +1,213 @@
...
+
+// HeaderConfig configures headers for synthetic push requests.
+type HeaderConfig struct {
+	headers.HeaderOps
+}
+
+// isRemoteResource returns true if resource starts with
+// a scheme or is a protocol-relative URI.
+func isRemoteResource(resource string) bool {
+	return strings.HasPrefix(resource, "//") ||
```
dunglas, line 181: While it’s probably good enough for now, the spec allows a server to push any resource it has authority (as defined in the HTTPS spec) on. Basically, a server can push resources for any domain present in its TLS certificate. This probably require a new config option to be handled properly.

dunglas:
To test this feature until Push support land in the Go HTTP client, it’s easy to use Symfony HTTP client (PHP). It’s what I’ve done to test the reference implementation of Vulcain: https://github.com/dunglas/vulcain/blob/master/gateway/server_test.go#L110

modules/caddyhttp/push/handler.go
```diff
@@ -0,0 +1,213 @@
...
+outer:
+	for _, resource := range resources {
+		for _, resource := range parseLinkHeader(resource) {
+			if _, ok := resource.params["nopush"]; ok {
+				continue
+			}
+			if isRemoteResource(resource.uri) {
+				continue
+			}
+			err := pusher.Push(resource.uri, &http.PushOptions{
```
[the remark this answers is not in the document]
mholt: That seems like it's a bug in the upstream then. Do we want to add allocations to cover up upstream bugs?

modules/caddyhttp/push/handler.go
```diff
@@ -0,0 +1,213 @@
...
+	"Accept-Encoding",
+	"Accept-Language",
+	"Accept",
+	"Cache-Control",
+	"User-Agent",
+}
+
+// pushHeader is a header field that gets added to push requests
+// in order to avoid recursive/infinite pushes.
+const pushHeader = "X-Caddy-Push"
```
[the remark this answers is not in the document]
mholt: Push headers get leaked to the client? Can you explain?

modules/caddyhttp/push/handler.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// serve only after pushing!
+	if err := next.ServeHTTP(w, r); err != nil {
+		return err
+	}
+
+	// finally, push any resources described by Link fields that were
+	// written to the response header, only if another push handler
+	// hasn't already done so
+	if links, ok := w.Header()["Link"]; ok {
```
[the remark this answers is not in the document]
mholt: That's... a good point... wonder why nobody (including me) brought this up in v1... I will re-work this.

modules/caddyhttp/push/handler.go
mholt: Yeah, this is complicated to get 100% correct. But I don't even know if the Go standard lib works well in that case, since the push requests are all synthetic and within the same Server instance; i.e. we can't push resources handled by external servers (without some sort of proxy configured separately within the current server).

I'm hoping/expecting that it's a very rare edge case anyway.

modules/caddyhttp/push/handler.go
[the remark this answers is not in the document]
dunglas: Apache goes even a step further and deduplicate for the whole TCP connection. The potential CPU time and bandwidth gain is worth some allocations imo.

modules/caddyhttp/push/handler.go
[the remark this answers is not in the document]
dunglas: By the spec push headers are sent to the client as part of the push promise: https://tools.ietf.org/html/rfc7540#section-8.2.1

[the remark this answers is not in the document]
mholt: I see, that's by design. I misunderstood your use of the term "leaked".

modules/caddyhttp/push/handler.go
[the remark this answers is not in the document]
mholt: We tend to try to do things differently (better) than other web servers around here :) Frankly it's not convincing to me that just because Apache covers up a user's misconfiguration or bug, that we should too, while reducing server performance at the same time.

modules/caddyhttp/push/handler.go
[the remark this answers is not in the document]
mholt: @dunglas I have addressed this issue in the latest commit. Please give it a try when you have a chance!

kelunik:
@dunglas If PHP is an option, [amphp/http-client with its push handler API](https://github.com/amphp/http-client/blob/ba5d824253c282e1865ad95f9aaaa1db50a67f09/src/Request.php#L284) might be a good choice for testing.

dunglas:
@kelunik the Symfony HTTP client is able to use AMPHP as well as libcurl. IMHO in this context it is better to use libcurl because it ensures that the feature will work at least with curl/libcurl, in all languages having bindings for it (which are plenty).

kelunik:
I guess it depends on how low level the tests should be. :)

modules/caddyhttp/push/handler.go
```diff
@@ -0,0 +1,236 @@
...
+	pusher, ok := w.(http.Pusher)
+	if !ok {
+		return next.ServeHTTP(w, r)
+	}
+
+	// short-circuit recursive pushes
+	if _, ok := r.Header[pushHeader]; ok {
+		return next.ServeHTTP(w, r)
+	}
+
```
kelunik, line 70: This means something like index.html > main.css > img.png would not be possible to push on a request to index.html?

mholt: I don't really know what you mean by that. Can you explain the "something like" part more?

kelunik: If index.html is configured to push main.css and main.css is configured to push img.png, and a user requests index.html, Caddy will only push main.css, because the synthetic request to main.css contains the header and suppresses further pushes, no?

mholt: Correct

dunglas: It's possible to "workaround" this behavior by computing recursively all resources to push, send the push promises and the push responses in background, but only start pushing the content of the explicit response when all resources to push has been resolved (and all push promises have been sent).

It's what we do in the Vulcain Gateway Server, it should be relatively easy to port the code.

kelunik: This is only possible if pushes are configured statically.

dunglas: It can work with dynamic pushes too (Vulcain supports dynamic client-requested pushes).

mholt: Is that a merge blocker? Can we implement that later?

dunglas: We definitely can implement this later (and I can work on this, I've already done it once, I can do it again 😄).

mholt: Great! I'll merge this in a moment then, and we can improve on it later. Thanks for the feedback, and I'd be very happy to work with you on improvements and/or review a PR for it.

PR #3576, opened 2020-07-13
reverseproxy: Fix Caddyfile parsing for empty non-http transports

### LINKED ISSUE #3566: Caddy speaks plain HTTP to FastCGI backend
I've recently converted my setup from Caddy 1 to Caddy 2. I have a custom application that speaks FastCGI, but it's not a PHP application.

I noticed that attempting to access this application just resulted in a hang. I've put together what I think is a minimal reproducer and stuck it, along with output when I run it, at https://github.com/gmacon/fcgi-repro. If you look at the packet capture of the traffic between Caddy and the backend, you'll see that Caddy tries to speak what looks like plain HTTP.

If you have `caddy`, `tcpdump`, and `python3` on your `$PATH`, you should be able to reproduce the issue by cloning that repo and running `./repro.sh`.

I just reproduced using a Caddy 2.1.1 binary from the release page.

(Aside: The contributing guidelines say to fill out the issue template when reporting a bug, but I can't tell that there is an issue template...)

Edit: I realize I should say: I've worked around this issue for now by switching to the HTTP transport.

francislavoie:
Fixes #3566


modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -561,7 +561,12 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 		} else if commonScheme == "https" {
 			return d.Errf("upstreams are configured for HTTPS but transport module does not support TLS: %T", transport)
 		}
-		if !reflect.DeepEqual(transport, reflect.New(reflect.TypeOf(transport).Elem()).Interface()) {
+
+		// if the transport is the same as the default,
+		// i.e. http and with no transport options, then
+		// we'll omit setting TransportRaw to avoid redundant
+		// JSON configuration.
+		if transportModuleName != "http" || !reflect.DeepEqual(transport, reflect.New(reflect.TypeOf(transport).Elem()).Interface()) {
```
mholt, line 569: This can be much simpler, comment included:

```suggestion
		// no need to encode empty default transport
		if !reflect.DeepEqual(transport, new(HTTPTransport)) {
```

mholt:
Thanks for the patch!!

PR #3578, opened 2020-07-14
logging: Implement Caddyfile support for filter encoder

### LINKED ISSUE #3694: Add filter encoder support to caddyfile log directive
I would like to be able to prevent certain headers (Authorization: X-Auth-Token:, etc ) from being included in the caddy logs. 

I found https://github.com/caddyserver/caddy/issues/3562, which is asking to be able to set an allowlist of headers, which makes sense for their use case.

In my case, a blocklist, makes more sense, as I only want to prevent a single header from being included in the logs.  

As a workaround, I'm current using "format single_field common_log", but the documentation mentions this is depreciated and I would like all the additional logging available, aside from security sensitive auth tokens.

francislavoie:
Closes #3694

~~This is essentially done but lacks tests and proper support for the `ip_mask` filter.~~

Caddyfile:

```
:80

log {
	output stdout
	format filter {
		wrap console
		fields {
			request>headers>Authorization delete
			request>headers>Server delete
			request>remote_addr ip_mask {
				ipv4 24
				ipv6 32
			}
		}
	}
}
```

Adapted JSON:

```json
{
	"logging": {
		"logs": {
			"default": {
				"exclude": [
					"http.log.access.log0"
				]
			},
			"log0": {
				"writer": {
					"output": "stdout"
				},
				"encoder": {
					"fields": {
						"request\u003eheaders\u003eAuthorization": {
							"filter": "delete"
						},
						"request\u003eheaders\u003eServer": {
							"filter": "delete"
						},
						"request\u003eremote_addr": {
							"filter": "ip_mask",
							"ipv4_cidr": 24,
							"ipv6_cidr": 32
						}
					},
					"format": "filter",
					"wrap": {
						"format": "console"
					}
				},
				"include": [
					"http.log.access.log0"
				]
			}
		}
	},
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":80"
					],
					"logs": {
						"default_logger_name": "log0"
					}
				}
			}
		}
	}
}
```

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -82,7 +82,35 @@ func (IPMaskFilter) CaddyModule() caddy.ModuleInfo {
 }
 
 // UnmarshalCaddyfile sets up the module from Caddyfile tokens.
-func (IPMaskFilter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+func (m *IPMaskFilter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	for d.Next() {
+		for d.NextBlock(0) {
+			switch d.Val() {
+			case "ipv4":
+				if !d.NextArg() {
+					return d.ArgErr()
+				}
+				if _, err := strconv.Atoi(d.Val()); err == nil {
+					m.IPv4MaskRaw = json.RawMessage(d.Val())
+				} else {
+					m.IPv4MaskRaw = json.RawMessage(`"` + d.Val() + `"`)
+				}
+
+			case "ipv6":
+				if !d.NextArg() {
+					return d.ArgErr()
+				}
+				if _, err := strconv.Atoi(d.Val()); err == nil {
+					m.IPv6MaskRaw = json.RawMessage(d.Val())
+				} else {
+					m.IPv6MaskRaw = json.RawMessage(`"` + d.Val() + `"`)
+				}
+
+			default:
+				return d.Errf("unrecognized subdirective %s", d.Val())
+			}
+		}
+	}
 	return nil
 }
 
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -15,11 +15,8 @@
 package logging
 
 import (
-	"encoding/json"
-	"fmt"
 	"net"
 	"strconv"
-	"unicode/utf8"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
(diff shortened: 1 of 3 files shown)


Mohammed90:
Is the `\u003e` part of the actual adapter output or an artifact of copy-paste/console output?

`request​\u003e​headers​\u003e​Authorization`

francislavoie:
Yeah I found it strange, I think it's just Golang's JSON encoder being funky. I don't do anything special here. I asked Matt and he said he's not concerned about it.

mholt:
Apparently we could switch to an Encoder instead of json.Marshal if we don't want the HTML escaping:

> String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune. So that the JSON will be safe to embed inside HTML <script> tags, the string is encoded using HTMLEscape, which replaces "<", ">", "&", U+2028, and U+2029 are escaped to "\u003c","\u003e", "\u0026", "\u2028", and "\u2029". This replacement can be disabled when using an Encoder, by calling SetEscapeHTML(false).

&mdash; https://golang.org/pkg/encoding/json/#Marshal

mholt:
When you have a chance, I'm hoping to do a final pass for 2.2 reviews next week. :)

francislavoie:
Okay, this now supports IP filters. I also added support for CIDR in the format that `ParseCIDR` supports (which tbh required some JSON gymnastics lol, but whatever, I think it's kinda useful, even if the IP portion is thrown away).

```
:80

log {
    output stdout
    format filter {
        wrap console
        fields {
            request>remote_addr ip_mask {
                ipv4 24
                ipv6 2001:db8::/32
            }
        }
    }
}
```

```json
{
  "logging": {
    "logs": {
      "default": {
        "exclude": [
          "http.log.access.log0"
        ]
      },
      "log0": {
        "writer": {
          "output": "stdout"
        },
        "encoder": {
          "fields": {
            "request\u003eremote_addr": {
              "filter": "ip_mask",
              "ipv4_cidr": 24,
              "ipv6_cidr": "2001:db8::/32"
            }
          },
          "format": "filter",
          "wrap": {
            "format": "console"
          }
        },
        "include": [
          "http.log.access.log0"
        ]
      }
    }
  },
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":80"
          ],
          "logs": {
            "default_logger_name": "log0"
          }
        }
      }
    }
  }
}
```

francislavoie:
Okay, this is all done.

I got rid of that abstraction I made that allows stuff like `2001:db8::/32` because frankly that's pretty dumb. I was working under incorrect assumptions with what the `int` for CIDR was at first (I thought it was itself the mask, as an int), which is why I went down the rabbithole of implementing that string thing. But it's not useful.

I kept the optimization I made of parsing the int into an IP mask at provision time, it's a super tiny optimization but it removes that overhead from every request being logged, so I think it's good to have. Small wins.

modules/logging/filters.go
```diff
@@ -68,6 +78,58 @@ func (IPMaskFilter) CaddyModule() caddy.ModuleInfo {
...
+// Provision parses m's IP masks, from integers.
+func (m *IPMaskFilter) Provision(ctx caddy.Context) error {
+	parseRawToMask := func(rawField int, bitLen int) net.IPMask {
+		if rawField == 0 {
+			return nil
+		}
+
+		// we assume the int is a subnet size CIDR
+		// e.g. "16" being equivalent to masking the last
+		// two bytes of an ipv4 address, like "255.255.0.0"
```
mholt, line 123: Why do we do this? What was wrong with what we had before?

francislavoie: Hmm I might've missed that comment, that should no longer be the case

francislavoie: Nvm - that's correct.

mholt: I mean why is an IP required at all?

francislavoie: It's not, are you misreading the comment? That's an example of what the mask result would look like

francislavoie: It's basically saying "if the input is `16`, then the output IP after being masked looks something like `255.255.0.0`"

mholt: I might be misreading it. I'll give this another pass tomorrow. :)

mholt (approved):
Kewl, LGTM. Shall we merge this?

francislavoie:
Yes 😄

utilitynerd:
A big thank you to @francislavoie for implementing this and @mholt for getting it merged!

PR #3587, opened 2020-07-18
fastcgi: Implement resolve symlinks feature

endelwar:
This PR will enable resolving real path on symlinked releases.
When using a symlink based directory schema PHP, through Caddy, will serve old symlinked directory instead of resolving the actual real path of new released code: due to the nature of PHP opcache the new release path is not read by PHP because it’s cached and symlink path doesn’t change form PHP prospective.
Adding the new `resolverootsymlink` subdirective on php_fastcgi will enable resolution of real path, thus permitting Caddy to server the actual codebase instead of the old one.

See [this forum post](https://caddy.community/t/resolve-real-path-on-symlinked-releases/8963) to further details.

Example Caddyfile:
```
http://example.test {
    root * /var/www/deploy/current
    php_fastcgi 127.0.0.1:9000 {
        resolverootsymlink
    }
}
```

Where `/var/www/deploy/current` is a symlink to real directory

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
@@ -39,6 +39,7 @@ func init() {
 //         root <path>
 //         split <at>
 //         env <key> <value>
+//         resolve_root_symlink
 //     }
 //
 func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
@@ -66,6 +67,9 @@ func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 					t.EnvVars = make(map[string]string)
 				}
 				t.EnvVars[args[0]] = args[1]
+
+			case "resolve_root_symlink":
+				t.ResolveRootSymlink = true
 
 			default:
 				return d.Errf("unrecognized subdirective %s", d.Val())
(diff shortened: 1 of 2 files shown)


francislavoie:
Thanks!

I think `resolve_symlinks` would be a better name for the option, for consistency and brevity. We typically use underscores between words in option names.

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
```diff
@@ -54,6 +54,9 @@ type Transport struct {
 	// that 404s if the fastcgi path info is not found.
 	SplitPath []string `json:"split_path,omitempty"`
 
+	// Resolve symlink in root path
```
mholt, line 57: This will be shown to users reading documentation, so explain what this is about.

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -39,6 +39,7 @@ func init() {
 //         root <path>
 //         split <at>
 //         env <key> <value>
+//         resolverootsymlink
```
mholt, line 42: Change to `resolve_root_symlink`

francislavoie: What about `resolve_symlinks`? I don't think `root` needs to be mentioned, and the internal golang function has the plural on it.

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
endelwar: Added some background on opcache and detailed subdirective effects: I hope it's not to lengthy now

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
endelwar: I went with `resolve_root_symlink`: it's the root path that is resolved, not any other symlink inside it, so IMO being detailed here is correct.

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
mholt: Detailed is good :+1:

PR #3595, opened 2020-07-23
httpcaddyfile: Bring `enforce_origin` and `origins` to admin config

Vigilans:
Currently, `admin` field in caddyfile's global option could only affects two fields in `caddy.AdminConfig`: `Disabled` and `Listen`. This PR makes modification of `EnsureOrigin` and `Origins` field available to caddyfile config.

### DIFF
--- a/caddyconfig/httpcaddyfile/options.go
+++ b/caddyconfig/httpcaddyfile/options.go
@@ -223,17 +223,39 @@ func parseOptSingleString(d *caddyfile.Dispenser) (interface{}, error) {
 }
 
 func parseOptAdmin(d *caddyfile.Dispenser) (interface{}, error) {
-	if d.Next() {
-		var listenAddress string
-		if !d.AllArgs(&listenAddress) {
-			return "", d.ArgErr()
+	adminCfg := new(caddy.AdminConfig)
+	for d.Next() {
+		if d.NextArg() {
+			listenAddress := d.Val()
+			if listenAddress == "off" {
+				adminCfg.Disabled = true
+				if d.Next() { // Do not accept any remaining options including block
+					return nil, d.Err("No more option is allowed after turning off admin config")
+				}
+			} else {
+				adminCfg.Listen = listenAddress
+				if d.NextArg() { // At most 1 arg is allowed
+					return nil, d.Err("At most 1 argument is allowed in admin config")
+				}
+			}
 		}
-		if listenAddress == "" {
-			listenAddress = caddy.DefaultAdminListen
+		for nesting := d.Nesting(); d.NextBlock(nesting); {
+			switch d.Val() {
+			case "enforce_origin":
+				adminCfg.EnforceOrigin = true
+
+			case "origins":
+				adminCfg.Origins = d.RemainingArgs()
+
+			default:
+				return nil, d.Errf("unrecognized parameter '%s'", d.Val())
+			}
 		}
-		return listenAddress, nil
 	}
-	return "", nil
+	if adminCfg.Listen == "" && !adminCfg.Disabled {
+		adminCfg.Listen = caddy.DefaultAdminListen
+	}
+	return adminCfg, nil
 }
 
 func parseOptOnDemand(d *caddyfile.Dispenser) (interface{}, error) {
--- a/caddyconfig/httpcaddyfile/options.go
+++ b/caddyconfig/httpcaddyfile/options.go
@@ -235,7 +235,7 @@ func parseOptAdmin(d *caddyfile.Dispenser) (interface{}, error) {
 			} else {
 				adminCfg.Listen = listenAddress
 				if d.NextArg() { // At most 1 arg is allowed
-					return nil, d.Err("At most 1 argument is allowed in admin config")
+					return nil, d.ArgErr()
 				}
 			}
 		}
--- a/caddytest/integration/caddyfile_adapt/global_options_admin.txt
+++ b/caddytest/integration/caddyfile_adapt/global_options_admin.txt
@@ -1,0 +1,80 @@
+{
+	debug
+	http_port 8080
+	https_port 8443
+	default_sni localhost
+	order root first
+	storage file_system {
+		root /data
+	}
+	acme_ca https://example.com
+	acme_ca_root /path/to/ca.crt
+
+	email test@example.com
+	admin {
+		origins localhost:2019 [::1]:2019 127.0.0.1:2019 192.168.10.128
+	}
+	on_demand_tls {
+		ask https://example.com
+		interval 30s
+		burst 20
+	}
+	local_certs
+	key_type ed25519
+}
+
+:80
+----------
+{
+	"admin": {
+		"listen": "localhost:2019",
+		"origins": [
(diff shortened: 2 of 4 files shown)


caddytest/integration/caddyfile_adapt/global_options_admin.txt
```diff
@@ -0,0 +1,80 @@
...
+	storage file_system {
+		root /data
+	}
+	acme_ca https://example.com
+	acme_ca_root /path/to/ca.crt
+
+	email test@example.com
+	admin {
+		origins localhost:2019 [::1]:2019 127.0.0.1:2019 192.168.10.128
+	}
```
Vigilans, line 16: This admin config typically does:
* Listen to default address `localhost:2019` (address argument left empty in the `admin <address> | off { ... }`)
* Disable `ensure_origin` (not set in the caddyfile)
* Allow outbound IP with default HTTP(s) port as origin (192.168.10.128 without port), in ordered to be called remotely via reverse_proxy.

caddyconfig/httpcaddyfile/options.go
```diff
@@ -223,17 +223,39 @@ func parseOptSingleString(d *caddyfile.Dispenser) (interface{}, error) {
...
-			return "", d.ArgErr()
+	adminCfg := new(caddy.AdminConfig)
+	for d.Next() {
+		if d.NextArg() {
+			listenAddress := d.Val()
+			if listenAddress == "off" {
+				adminCfg.Disabled = true
+				if d.Next() { // Do not accept any remaining options including block
+					return nil, d.Err("No more option is allowed after turning off admin config")
+				}
```
Vigilans, line 234: This check forbids block after admin config off: 
```
admin off {
    enforce_origin
    origins 192.168.1.1:2020 127.0.0.1:2020
}
```

caddyconfig/httpcaddyfile/options.go
```diff
@@ -223,17 +223,39 @@ func parseOptSingleString(d *caddyfile.Dispenser) (interface{}, error) {
...
+			} else {
+				adminCfg.Listen = listenAddress
+				if d.NextArg() { // At most 1 arg is allowed
+					return nil, d.Err("At most 1 argument is allowed in admin config")
```
Vigilans, line 238: This check forbids multiple ips to listen:
```
admin 192.168.1.1:2020 127.0.0.1:2020 {
    enforce_origin
    origins 192.168.1.1:2020 127.0.0.1:2020
}
```

caddyconfig/httpcaddyfile/options.go
```diff
@@ -223,17 +223,39 @@ func parseOptSingleString(d *caddyfile.Dispenser) (interface{}, error) {
...
+		for nesting := d.Nesting(); d.NextBlock(nesting); {
+			switch d.Val() {
+			case "enforce_origin":
+				adminCfg.EnforceOrigin = true
+
+			case "origins":
+				adminCfg.Origins = d.RemainingArgs()
+
+			default:
+				return nil, d.Errf("unrecognized parameter '%s'", d.Val())
```
Vigilans, line 251: This check forbids unrecognized fields like:
```
admin {
    disabled false
}
```
(this is a legacy unit test)

caddyconfig/httpcaddyfile/options.go
mholt: You can also return `d.ArgErr()` which is designed for this. It's not required but the error message in this case is very similar anyway.
```suggestion
				adminCfg.Listen = listenAddress
				if d.NextArg() { // At most 1 arg is allowed
					return nil, d.ArgErr()
```

caddytest/integration/caddyfile_adapt/global_options_admin.txt
mholt: > Allow outbound IP with default HTTP(s) port as origin (192.168.10.128 without port), in ordered to be called remotely via reverse_proxy.

I'm not sure I understand this part...

FYI reverse proxying to the admin endpoint _within the same Caddy instance/config_ will not work (because of a chicken-egg problem). I am not sure if that is what you are referring to, but just wanted to clarify that.

Vigilans: Not sure what kind of scenario you are referring to, may I just clarify my usecase here:
1. I have established a Virtual LAN using [ZeroTierOne](https://github.com/zerotier/ZeroTierOne) with all trusted devices joined, and the IP of my caddy VPS is `10.147.17.12`. 
2. I started an https server on 10.147.17.12, with admin endpoint reverse proxied on path `/api/caddy`, so only trusted devices can get to the admin endpoint of my caddy server, and the data is encrypted in the public network. The config is as follows:
   ```
   https://10.147.17.12 {
     route /api/caddy/* {
       uri strip_prefix /api/caddy
       reverse_proxy http://localhost:2019
     }
   }
   ```
3. I tried to GET the caddy config through the private VLAN IP, but `host not allowed` reported:
   ```
   $ curl -k https://10.147.17.12/api/caddy/config/
   {"error":"host not allowed: 10.147.17.12"}
   ```
4. I added the `10.147.17.12` to the admin config:
   ```
   {
     admin {
       origins localhost:2019 [::1]:2019 127.0.0.1:2019 10.147.17.12
     }
   }
   ```
   And now I can get the caddy config from remote machine:
   ```
   $ curl -k https://10.147.17.12/api/caddy/config/
   {"admin":{"listen":"localhost:2019","origins":["localhost:2019","[::1]:2019","127.0.0.1:2019","10.147.17.12"]}, "apps": ... }
   ```
   **NOTE:** If I replace `10.147.17.12` with `10.147.17.12:443` in origins field, I cannot get the config whether I specified port in client request or not:
   ```
   $ curl -k https://10.147.17.12/api/caddy/config/
   {"error":"host not allowed: 10.147.17.12"
   $ curl -k https://10.147.17.12:443/api/caddy/config/
   {"error":"host not allowed: 10.147.17.12"}}
   ```
   I cannot accurately interpret this behavior, and I think maybe caddy automatically removed the 443 port and pass `10.147.17.12` as host for checking, so only `10.147.17.12` without specifying port with work.

mholt: Thanks for explaining!

>     `reverse_proxy http://localhost:2019`

This will be problematic. Config changes through this proxy will wait for the admin endpoint to return, obviously, but the admin endpoint will be waiting for the reverse proxy to return, for a graceful reload. So it's a chicken-and-egg problem that I don't know how to solve yet.

Unrelated to this PR, but I thought you should know.

> but host not allowed reported: 

Correct; alternatively, you can have the proxy change the Host and Origin headers as they go upstream to the backend.

> NOTE: If I replace 10.147.17.12 with 10.147.17.12:443 in origins field, I cannot get the config whether I specified port in client request or not: 

Also correct, because the `Host` header won't usually include a port for standard ports (that is up to the HTTP client). In this case, `curl` does not append `:443` to the header. Caddy does not change them.

mholt:
Thank you very much for the great contribution! It'll go out with 2.2 beta soon.

Issue #3615, opened 2020-07-28
Caddy Build Fails for Solaris/Illumos and Plan9


mohammed90:
Initially reported by @Toasterson

> The main issue is that badger@1.5.3 uses golangs internal syscall library and that does not support many operatingsystems notably openbsd, netbsd, and illumos. Badger uses a constant for MADVISE and that only came into the go for these Unixes after the internal syscall library got frozen. Right now the internal syscall library only really works for linux and every software that wants to support something else than linux and uses the syscall library directly must use x/sys/unix instead. I've personally changed quite a few direct dependencies and pushed people to use x/sys/unix instead of syscall so that today this is mostly happening due the dependency chain. I could not upgrade badger to 1.6.1 as there where api changes but nosql bump worked. So until that is resolved caddy cannot be built on quite a few unix versions I suppose. I know it won't build for illumos right now. Are you monitoring cross-builds?
>

> Also here the exact error for documentation
> 
> ```
> # github.com/dgraph-io/badger/y
> ../../../pkg/mod/github.com/dgraph-io/badger@v1.5.3/y/mmap_unix.go:57:30: undefined: syscall.SYS_MADVISE
> ```

_Originally posted by @Toasterson in https://github.com/caddyserver/caddy/pull/3602#issuecomment-663876698_

Build errors:
- Solaris/Illumos
```
$ GOOS=solaris go build
# github.com/dgraph-io/badger/y
../../../../../go/pkg/mod/github.com/dgraph-io/badger@v1.5.3/y/mmap_unix.go:57:30: undefined: syscall.SYS_MADVISE
```

- plan9:
```
$ GOOS=plan9 go build
# github.com/dgraph-io/badger/y
../../../../../go/pkg/mod/github.com/dgraph-io/badger@v1.5.3/y/file_dsync.go:24:21: undefined: unix.O_DSYNC
../../../../../go/pkg/mod/github.com/dgraph-io/badger@v1.5.3/y/mmap_unix.go:32:11: undefined: unix.PROT_READ
../../../../../go/pkg/mod/github.com/dgraph-io/badger@v1.5.3/y/mmap_unix.go:34:12: undefined: unix.PROT_WRITE
../../../../../go/pkg/mod/github.com/dgraph-io/badger@v1.5.3/y/mmap_unix.go:36:9: undefined: unix.Mmap
[...]

PR #3617, opened 2020-07-29
ci: Include tracking of GOOS for which Caddy fails to build

Mohammed90:
We learned about the build failure for Solaris by accident. Perhaps it's better we have track of which OSs Caddy fails to build for to have an understanding our areas of improvements and where we are probably unable to support.

Updates #3615


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 # Used as inspiration: https://github.com/mvdan/github-actions-golang
 
-name: Cross-Platform
+name: Cross-Platform Tests
```
francislavoie, line 3: Idk I think we can shorten this frankly

```suggestion
name: Tests
```

.github/workflows/cross-build.yml
```diff
@@ -0,0 +1,51 @@
+name: Cross-Build (Advisory Check)
```
francislavoie, line 1: :man_shrugging: 

```suggestion
name: Cross-Build
```

francislavoie:
FYI @mholt unfortunately github actions doesn't have a "warning" state for jobs yet, but there's an open issue discussing it here https://github.com/actions/toolkit/issues/399

So for now, we have `continue-on-error: true` set for this stuff so that it doesn't make the PR go red for any cross-build failures, but the only other option is green. To see the failures you can look at the annotations on the workflow: https://github.com/caddyserver/caddy/actions/runs/187524216

mholt:
This looks cool, neat idea!

I can see why this would be even better with some sort of "warning" result instead of just green/red.

I guess something we can add to our TODO list is to submit patches upstream, or at least request that they make their build tags more correct, so that we can continue to build x-platform.

Mohammed90:
> Is this ready to merge or should we wait? Is it even useful if it always emits a green status?

It is ready. The results always being green bothers me too, but I hope the workflow being listed reminds to check for annotations... until Github Actions supports tri-state results.

Mohammed90:
Blocked by #3625 😓 (kinda, sorta)

mholt:
My bad. Will get that figured out soon.

mholt:
Okie all fixed

PR #3619, opened 2020-07-30
Fix a typo in modules.go

gulien:
Hello 👋 

While diving into the `modules` code, I've found this weird typo.

As far as I know, it appears to be an error, but I might be wrong. 🤔

### DIFF
--- a/modules.go
+++ b/modules.go
@@ -212,7 +212,6 @@ func GetModules(scope string) []ModuleInfo {
 	}
 
 	var mods []ModuleInfo
-iterateModules:
 	for id, m := range modules {
 		modParts := strings.Split(id, ".")
 


Mohammed90:
On the contrary, the code is correct and this PR isn't. Notice the failing CI runs. It is called [label](https://www.ardanlabs.com/blog/2013/11/label-breaks-in-go.html).

francislavoie:
Nope, that's correct actually. It's a go-to label (I think golang may call them tags?). See the matching `continue` a little deeper in that function.

Best link I could find to explain on short notice from my phone: https://github.com/golang/go/issues/36098

gulien:
Thanks for the explanations!

PR #3620, opened 2020-07-30
reverse_proxy: fix bi-h2stream breaking gzip encode handle(#3606).

### LINKED ISSUE #3606: after e9b1d7d commit, reverse proxy to the web page of GOGS will become garbled
The commit of e9b1d7dcb4cbf85da7fb4cf8c411a4f840a98cf1 solves the #3556 issue, but may lead to new issues. 
This problem may be due to the content returned by GOGS in chunked mode and lack of the Content-Length field. I am not sure whether all chunked pages are affected by this commit.

masknu:
Sorry for breaking the encode handle.  This PR fixes #3606 . Add an integration test.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -614,7 +614,12 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, di Dia
 	// some apps need the response headers before starting to stream content with http2,
 	// so it's important to explicitly flush the headers to the client before streaming the data.
 	// (see https://github.com/caddyserver/caddy/issues/3556 for use case)
-	if req.ProtoMajor == 2 && res.ContentLength == -1 {
+	// we have to check encoding here, only flush headers with identity encoding.
+	// Non-identity encoding might combine with "encode gzip" directive,
+	// in that case, if body size larger than enc.MinLength,
+	// upper level encode handle might have Content-Encoding header to write.
+	// (see https://github.com/caddyserver/caddy/issues/3606 for use case)
+	if req.ProtoMajor == 2 && res.ContentLength == -1 && req.Header.Get("Accept-Encoding") == "identity" {
 		if wf, ok := rw.(http.Flusher); ok {
 			wf.Flush()
 		}
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -619,7 +619,7 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, di Dia
 	// in that case, if body size larger than enc.MinLength,
 	// upper level encode handle might have Content-Encoding header to write.
 	// (see https://github.com/caddyserver/caddy/issues/3606 for use case)
-	if req.ProtoMajor == 2 && res.ContentLength == -1 && req.Header.Get("Accept-Encoding") == "identity" {
+	if h.biStream(req, res) {
 		if wf, ok := rw.(http.Flusher); ok {
 			wf.Flush()
 		}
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -614,7 +614,12 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, di Dia
 	// some apps need the response headers before starting to stream content with http2,
 	// so it's important to explicitly flush the headers to the client before streaming the data.
 	// (see https://github.com/caddyserver/caddy/issues/3556 for use case)
-	if req.ProtoMajor == 2 && res.ContentLength == -1 {
+	// we have to check encoding here, only flush headers with identity encoding.
+	// Non-identity encoding might combine with "encode gzip" directive,
+	// in that case, if body size larger than enc.MinLength,
+	// upper level encode handle might have Content-Encoding header to write.
+	// (see https://github.com/caddyserver/caddy/issues/3606 for use case)
+	if req.ProtoMajor == 2 && res.ContentLength == -1 && req.Header.Get("Accept-Encoding") == "identity" {
```
mholt, line 622: If Accept-Encoding is empty (not specified by the client), would that also be permissible for flushing?

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -96,8 +96,8 @@ func (h Handler) flushInterval(req *http.Request, res *http.Response) time.Durat
 		return -1 // negative means immediately
 	}
 
-	// for h2 and h2c upstream streaming data to client (issue #3556)
-	if req.ProtoMajor == 2 && res.ContentLength == -1 {
+	// for h2 and h2c upstream streaming data to client (issue #3556,#3606)
+	if req.ProtoMajor == 2 && res.ContentLength == -1 && req.Header.Get("Accept-Encoding") == "identity" {
```
mholt, line 100: Can we abstract these conditions into a function instead of repeating it? And also split the conditions onto separate lines by `&&`.
```suggestion
	if req.ProtoMajor == 2 &&
		res.ContentLength == -1 &&
		req.Header.Get("Accept-Encoding") == "identity" {
```

modules/caddyhttp/reverseproxy/reverseproxy.go
masknu: According to [rfc2616-sec14](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html), it's permissible.
Should we consider dealing with the following form? It means accepting all but "identity".

```
Accept-Encoding: identity;q=0, *;q=1
```

modules/caddyhttp/reverseproxy/streaming.go
masknu: Yes, that would look better.

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -97,14 +97,27 @@ func (h Handler) flushInterval(req *http.Request, res *http.Response) time.Durat
...
+	if h.biStream(req, res) {
 		return -1
 	}
 
 	// TODO: more specific cases? e.g. res.ContentLength == -1? (this TODO is from the std lib)
 	return time.Duration(h.FlushInterval)
 }
 
+// biStream returns whether we should work in bi-directional stream mode.
+func (h Handler) biStream(req *http.Request, res *http.Response) bool {
```
francislavoie, line 109: I think the function name should be `isBidirectionalStream`. Clearer the result is a boolean, and "bistream" doesn't have meaning on its own.

masknu: Sounds good.

mholt:
@masknu This looks great! I took the liberty of applying the last few minor cleanups. If it looks good to you I'm ready to merge this.

masknu:
That looks great to me.

mholt:
Thank you so much for your contribution and for the discussion! Please contribute again. 🙂

PR #3621, opened 2020-07-30
caddytls: Replace lego with acmez

mholt:
Persuant to https://github.com/caddyserver/certmagic/issues/71, this change replaces [lego](https://github.com/go-acme/lego) with [acmez](https://github.com/mholt/acmez).

Caddy has used lego since day 1 -- in fact, lego was originally written _for use in Caddy_, way back in early 2015. Lego has been a wonderful library but new requirements in Caddy 2 necessitate a change to a lighter, simpler, and more robust ACME library. (ACMEz has built-in retries at every level and can iterate multiple challenge types, smartly converges to most successful challenge type, customizable logging, a simpler and more elegant API, etc. See linked issue above for some details behind the improvements.)

Lego has also focused on being a CLI tool + Traefik library which is different from what Caddy is and needs. With this change, we now have complete control of the whole ACME / certificate management stack.

Some benefits/improvements:

- Reduced go.sum by about 200 lines, i.e. dropped about 100 indirect dependencies - this makes builds (especially cross-platform builds) less brittle.
- Faster, more reliable challenge solving.
- Context cancellation is supported all the way down, meaning that config reloads will be faster and lighter: Caddy will be even more nimble!
- Can take full advantage of [providers](https://github.com/libdns) implemented for [libdns APIs](https://github.com/libdns/libdns).
- Removed temporary/shim DNS-01 solver, and cleaned it up and put it in [CertMagic](https://github.com/caddyserver/certmagic) so that any CertMagic users can benefit from all libdns providers and use them for the DNS challenge.
- Consistent logs! We use structured logging all the way down the stack (well, down to the Go standard lib of course), so you won't see logs with inconsistent looks and time zones as ACME challenges are solved.

This change builds on an all-new, low-level ACME client implementation called acmez which has been tested with pebble, Boulder, Let's Encrypt staging, Let's Encrypt production, ZeroSSL's ACME endpoint, and Caddy's internal ACME server. It also builds on necessary changes in the middle layer, CertMagic, which does the hard work of managing the certificates over time. Caddy had most of that abstracted away pretty well in front of our config structure, so most of this change in Caddy is simply adjusting the layer that connects configuration to CertMagic + acmez stuff.


mholt:
@Mohammed90 Protobuf, yeah? I had a hard time during the squash and I wonder if I messed that up. I _thought_ it was what we had in master currently, but I think it was genproto (which did give me compile errors until I reverted to what master had) -- but it builds now and all the tests pass, and I am hoping we can upgrade some dependencies soon before the 2.2 beta. So protobuf will get upgraded then too probably.

Thanks for the once-overs!

mafrosis:
I was reading your source code trying to understand some of the imports from `lego` and how all the pieces fit together. Then the pages started 404ing.. I thought I was going mad. Turns out this P/R was merged and removed the things I was looking at!

I'll try a build off `master` :p

mholt:
@mafrosis Ha, but that's weird, the only file that disappeared was dnssolver.go. (Bad luck I guess?)

mafrosis:
@mholt Yep! I have been seeing strange behaviour with DNS resolution in `caddy` whilst solving DNS.. So that was exactly where I was looking.

I'll try with `master` and then maybe open a forum post.

Thanks for all the hard work on this software!

PR #3624, opened 2020-07-31
httpcaddyfile: Support `compression` in http transport config

Vigilans:
I have successfully established h2c connection between caddy and my local service with the commits in https://github.com/caddyserver/caddy/pull/3561. 

But the json config in https://github.com/caddyserver/caddy/issues/3556#issuecomment-655274156 cannot be simply ported to caddyfile because of the absence of `compression` field in `reverse_proxy -> transport http` config. This PR is just a minor modification to make this field available to caddyfile's http transport config.

Config example:
```
    transport http {
      versions h2c 2
      compression off
    }
```

### DIFF
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -719,6 +719,14 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 					return d.ArgErr()
 				}
 
+			case "compression":
+				if d.NextArg() {
+					if d.Val() == "off" {
+						var disable bool
+						h.Compression = &disable
+					}
+				}
+
 			default:
 				return d.Errf("unrecognized subdirective %s", d.Val())
 			}
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_h2c.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_h2c.txt
@@ -1,0 +1,96 @@
+
+https://example.com {
+	reverse_proxy /path http://localhost:54321 {
+		header_up Host {host}
+		header_up X-Real-IP {remote}
+		header_up X-Forwarded-For {remote}
+		header_up X-Forwarded-Port {server_port}
+		header_up X-Forwarded-Proto "http"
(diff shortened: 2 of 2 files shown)


modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -719,6 +719,14 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 					return d.ArgErr()
 				}
 
+			case "compression":
+				if d.NextArg() {
+					if d.Val() == "off" {
+						var disable bool
+						h.Compression = &disable
```
francislavoie, line 726: Why use this approach to setting it with a reference? Why not just set it to false directly?

```suggestion
						h.Compression = false
```

francislavoie: Also could you please add an adapt test for this? See `caddytest/integration/caddyfile_adapt/*`

Vigilans: > Why use this approach to setting it with a reference? Why not just set it to false directly?

This approach is imitating a legacy practice in setting another field:
https://github.com/caddyserver/caddy/blob/19cc2bd3c3e6b551d47eca7a2efdee95eec2a86f/modules/caddyhttp/reverseproxy/caddyfile.go#L684-L695

Here `h.Compression` is of type `*nool`, so cannot just bind false to it...

Vigilans: > Also could you please add an adapt test for this? See `caddytest/integration/caddyfile_adapt/*`

I've added an adapt test for a typical h2c reverse_proxy setup that conforms to the JSON config in https://github.com/caddyserver/caddy/issues/3556#issuecomment-655274156.

francislavoie: Thanks!

mholt: The reason the boolean is a pointer is because we need a way to distinguish between the user explicitly setting it to false and it being unset, because the default is true.

mholt:
Feel free to contribute again sometime :+1:

PR #3625, opened 2020-07-31
fileserver: First attempt to fix failing test on Linux

mholt:
I think I updated the wrong test case before. I don't have Linux handy so it's going to be trial-and-error until we can figure this out.

This is a new test case. If I recall correctly, we had problems with `.PHP` (non-lowercased path splits) that led to unfavorable handling of those requests in the past.

### DIFF
--- a/modules/caddyhttp/fileserver/matcher_test.go
+++ b/modules/caddyhttp/fileserver/matcher_test.go
@@ -39,6 +39,11 @@ func TestPHPFileMatcher(t *testing.T) {
 			matched:      true,
 		},
 		{
+			path:         "/index.PHP/somewhere",
+			expectedPath: "/index.PHP",
+			matched:      true,
+		},
+		{
 			path:         "/remote.php",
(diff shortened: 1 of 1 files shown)


mholt:
This test might only pass on case-insensitive file systems.

Vigilans:
I debugged the `matcher_test`, the failure occured after the `os.Stat` call:
https://github.com/caddyserver/caddy/blob/af5c148ed1d8e405978b71e7efe15d0a829cab50/modules/caddyhttp/fileserver/matcher.go#L255-L267

Here it tried to stat `testdata/foo.php.PHP/index.php`, but linux's file system is case sensitive:
```
caddy/modules/caddyhttp/fileserver/testdata$ ls
foo.php.php  index.php  notphp.php.txt  remote.php
```
So the flow fell into `err != nil` block and false was returned.

mholt:
Thank you @Vigilans ! I bypassed that part of the code for the test in the end so it won't matter what FS it runs on.

PR #3629, opened 2020-08-04
reverseproxy: Add Caddyfile scheme shorthand for h2c

### LINKED ISSUE #3628: Support h2c URI scheme for reverse proxy upstream address
caddy 2.1.0 added support for HTTP2 cleartext connections but the syntax is a bit cumbersome:

```
https://grpc-api.example.com {
        transport http {
            versions h2c 2
        }
    }
}
```

It would be nice if we could use `h2c://` instead:
```
https://grpc-api.example.com {
    reverse_proxy h2c://localhost:8080
}
```
Traefik is using the same syntax: https://docs.traefik.io/user-guides/grpc/

francislavoie:
Closes #3628

### DIFF
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_h2c_shorthand.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_h2c_shorthand.txt
@@ -1,0 +1,38 @@
+:8884
+
+reverse_proxy h2c://localhost:8080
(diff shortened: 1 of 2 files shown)


mholt (changes requested):
LGTM, just one nit about readability.

mholt (approved):
Let's do this

PR #3633, opened 2020-08-04
caddytls: Add support for ZeroSSL; add Caddyfile support for issuers

mholt:
Configuring issuers explicitly in a Caddyfile is not easily compatible
with existing ACME-specific parameters such as email or acme_ca which
infer the kind of issuer it creates (this is complicated now because
the ZeroSSL issuer wraps the ACME issuer)... oh well, we can revisit
that later if we need to.

New Caddyfile global option:

    {
        cert_issuer <name> ...
    }

Or, alternatively, as a tls subdirective:

    tls {
        issuer <name> ...
    }

For example, to use ZeroSSL with an API key:

    {
        cert_issuser zerossl API_KEY
    }

For now, that still uses ZeroSSL's ACME endpoint; it fetches EAB
credentials for you. You can also provide the EAB credentials directly
just like any other ACME endpoint:

    {
        cert_issuer acme {
            eab KEY_ID MAC_KEY
        }
    }

All these examples use the new global option (or tls subdirective). You
can still use traditional/existing options with ZeroSSL, since it's
just another ACME endpoint:

    {
        acme_ca  https://acme.zerossl.com/v2/DV90
        acme_eab KEY_ID MAC_KEY
    }

That's all there is to it. You just can't mix-and-match acme_* options
with cert_issuer, because it becomes confusing/ambiguous/complicated to
merge the settings.

----

The latest commits on this PR make it possible to implicitly configure the ZeroSSL issuer by specifying a CA endpoint containing `acme.zerossl.com` and an email address; no API key required:

```
{
    acme_ca https://acme.zerossl.com/v2/DV90
    email   you@yours.com
}
```

(And similarly for with the `tls` directive.)

In other words, even though using ZeroSSL (the first time) requires the `ZeroSSLIssuer` type (to generate the EAB credentials when making an ACME account), it's very easy to configure ZeroSSL: Caddy will mostly auto-detect it and wrap the ACMEIssuer properly with ZeroSSLIssuer. You can also explicitly configure ZeroSSL as shown above.

No issuer defaults are changed with this PR.

----

With version 2.3 (to clarify: not 2.2, the release for which this change is slated), we can then work on making one CA a fallback of another. For example, if ZeroSSL is the primary CA but is down, then switch to Let's Encrypt. Or vice-versa, if LE is down (or you've hit rate limits), switch to ZeroSSL.

### DIFF
--- a/modules/caddytls/acmeissuer.go
+++ b/modules/caddytls/acmeissuer.go
@@ -20,9 +20,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 	"github.com/caddyserver/certmagic"
 	"github.com/mholt/acmez"
 	"github.com/mholt/acmez/acme"
@@ -86,12 +88,12 @@ func (ACMEIssuer) CaddyModule() caddy.ModuleInfo {
 }
 
 // Provision sets up m.
-func (m *ACMEIssuer) Provision(ctx caddy.Context) error {
-	m.logger = ctx.Logger(m)
+func (iss *ACMEIssuer) Provision(ctx caddy.Context) error {
+	iss.logger = ctx.Logger(iss)
 
 	// DNS providers
-	if m.Challenges != nil && m.Challenges.DNS != nil && m.Challenges.DNS.ProviderRaw != nil {
-		val, err := ctx.LoadModule(m.Challenges.DNS, "ProviderRaw")
+	if iss.Challenges != nil && iss.Challenges.DNS != nil && iss.Challenges.DNS.ProviderRaw != nil {
+		val, err := ctx.LoadModule(iss.Challenges.DNS, "ProviderRaw")
 		if err != nil {
 			return fmt.Errorf("loading DNS provider module: %v", err)
 		}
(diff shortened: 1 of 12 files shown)


mholt:
Sigh -- failing test on s390x VM is unrelated.

modules/caddytls/acmeissuer.go
```diff
@@ -86,12 +88,12 @@ func (ACMEIssuer) CaddyModule() caddy.ModuleInfo {
 }
 
 // Provision sets up m.
```
francislavoie, line 90:
```suggestion
// Provision sets up iss.
```

modules/caddytls/acmeissuer.go
```diff
@@ -170,32 +172,156 @@ func (m *ACMEIssuer) makeIssuerTemplate() (certmagic.ACMEManager, error) {
...
+//
+//     ... acme {
+//         dir <directory_url>
+//         test_dir <test_directory_url>
+//         email <email>
+//         timeout <duration>
+//         disable_http_challenge
+//         disable_tlsalpn_challenge
+//         alt_http_port    <port>
+//         alt_tlsalpn_port <port>
```
francislavoie, line 219: Could these be generalized or would that be a waste of time?

```
challenge http disable

# and/or:

challenge tlsalpn {
    # disable
    alt_port 8443
}
```

francislavoie:
LGTM 👍

modules/caddytls/acmeissuer.go
mholt: Thought about that, but I felt like it was more natural to remember and write something that sounded like "disable HTTP challenge". There's only two knobs for these two challenges, and one is "off". So, I figured... this is fine? But yes, technically your approach is more scalable. I don't foresee standardized, general-purpose challenges growing much more in the next decade or so, but maybe we'll look back on this comment and see that it didn't age well.

francislavoie: Fair enough. Just being nitpicky :smile: I don't disagree.

PR #3658, opened 2020-08-14
httpcaddyfile: Improve directive sorting logic

francislavoie:
See discussion in https://caddy.community/t/v2-help-with-path-regexp-and-rewrite/9442/11

I'm pretty sure that it's almost always incorrect to sort `root` handlers from the Caddyfile in the same order as other directives, because it's essentially setting a variable, so only the last one run will apply - so you want the last one run to be the most-specific one.

Also, see the discussion in https://caddy.community/t/presenting-different-content-based-on-origin-of-request/9467/7

If a directive has any kind of matcher, it should still sort before a directive of the same name with no matcher at all.


caddyconfig/httpcaddyfile/directives.go
```diff
@@ -398,6 +398,21 @@ func sortRoutes(routes []ConfigValue) {
 		if len(jPM) > 0 {
 			jPathLen = len(jPM[0])
 		}
+
+		// if both directives have no path matcher, use whichever one
+		// has any kind of matcher defined first.
+		if iPathLen == 0 && jPathLen == 0 {
+			return len(iRoute.MatcherSetsRaw) > len(jRoute.MatcherSetsRaw)
+		}
```
mholt, line 406: Technically, this will sort by which one has more matcher sets than the other, but we don't know that that's the right order. What you want here instead is a return value conditional on whether the `i` or `j` route has 0 matchers and the other one has more than 0.

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -398,6 +398,21 @@ func sortRoutes(routes []ConfigValue) {
...
+
+		// sort the "root" directive in reverse because we want the
+		// most-specific path last such that the root variable
+		// overrides a less-specific one.
+		if iDir == "root" {
+			return iPathLen < jPathLen
+		}
```
mholt, line 413: I think we have to be careful before we decide to do this... though I'm not opposed, the implications here for consistency and correctness run deeper. By similar logic, we should also sort `header` directives this way. And who knows which other ones. If we're going to treat one as a special case for a certain reason, we need to treat all of them where the same logic applies: any middleware directive that effectively overwrites any instances of the previous directive.

caddyconfig/httpcaddyfile/directives.go
francislavoie: True, but I don't think it's ever possible for a Caddyfile directive to have more than one matcher set. But it's an easy fix.

caddyconfig/httpcaddyfile/directives.go
francislavoie: We can easily do that with a lookup table beside the other directive order map, but the question obviously is then how do we want it to work for plugins, do we want to give them an API to plug into this as well?

mholt: Let's not bite off more than we can chew here... and just keep it simple for now.

Remember this is just a convenience, there's always the `route` directive to do things manually.

francislavoie: Sounds like `root`, `rewrite` and `header`, not sure if we need `handle` and `handle_path` as well :thinking:

francislavoie:
Hmm. Looks like we have a flaky test. For some reason on some environments, the first `group` is `0`, and sometimes it's `1`. :thinking:

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -69,6 +69,18 @@ var directiveOrder = []string{
 	"acme_server",
 }
 
+// directiveShouldSortReversed specifies whether a particular
+// directive should be sorted in reverse, compared to the
+// default of most-specific path matcher ordered first.
+// This is typically relevant for directives that have
+// "override" behaviour, where you want the most specific
```
francislavoie, line 76: Sigh... golangci is trying to make me American.

> [golangci] reported by reviewdog 🐶
> `behaviour` is a misspelling of `behavior` (misspell)

https://dictionary.cambridge.org/us/dictionary/english/behaviour

```suggestion
// "override" behavior, where you want the most specific
```

mholt: Ha, that's annoying. Sorry!

caddyconfig/httpcaddyfile/directives.go
francislavoie: Alright so we talked offline, didn't need this bit ultimately... the mutual exclusivity (the `group` thing) should make this work correctly anyways, we just needed the fix for non-path matchers for this to do the right thing.

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -398,6 +398,21 @@ func sortRoutes(routes []ConfigValue) {
...
+		// has any kind of matcher defined first.
+		if iPathLen == 0 && jPathLen == 0 {
+			var iMatchers, jMatchers int
+			if len(iRoute.MatcherSetsRaw) > 0 {
+				iMatchers = 1
+			}
+			if len(jRoute.MatcherSetsRaw) > 0 {
+				jMatchers = 1
+			}
+			return iMatchers > jMatchers
```
mholt, line 412: This is quite a bit shorter:

```suggestion
			return len(iRoute.MatcherSetsRaw) > 0 && len(jRoute.MatcherSetsRaw) == 0
```

Double-check my logic here :) But I think this is the same idea: i < j if i has a matcher and j doesn't.

mholt:
Note that the issue at https://caddy.community/t/v2-help-with-path-regexp-and-rewrite/9442 is not resolved by this because we don't (can't?) sort `path_regexp` matcher by length; only `path`.

francislavoie:
> Note that the issue at caddy.community/t/v2-help-with-path-regexp-and-rewrite/9442 is not resolved by this because we don't (can't?) sort path_regexp matcher by length; only path.

Actually it is, because having a named matcher will now sort ahead of `*`, but mutual exclusivity will prevent the `*` one from taking effect altogether.

```
:8881

root * /foo

@regexp path_regexp foo ^/z(.*)$
root @regexp /bar

respond "Root: {http.vars.root}"
```

mholt:
Ah, indeed, I was thinking more generally I guess, that this won't sort path_regexp matchers relative to path matchers.

Will merge this in a sec.

PR #3673, opened 2020-08-20
ci: Tweaks for multi go version tests

francislavoie:
- Build cache should be per go version
- We can shorten the matrix var to just `go`
- Publish separate CI artifacts per go version, drop the `_v2` infix (kinda meaningless)
- Specify the go version in a short format without the `.x`, looks cleaner
- Bump release to go 1.15, (but not fuzzing, it's not ready for 1.15 yet)
- Bump setup-go to v2, cache to v2, goreleaser-action to v2


francislavoie:
@mholt merge this before releasing, because otherwise the release will be with go 1.14 builds

.github/workflows/fuzzing.yml
```diff
@@ -12,14 +12,14 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        go-version: [ 1.14.x ]
+        go: [ '1.15' ]
```
Mohammed90, line 15: Keep the Go version here at 1.14. There's a [known bug in the go/printer package in Go 1.15](https://github.com/golang/go/issues/40546) which [breaks go-fuzz](https://github.com/dvyukov/go-fuzz/issues/294).

francislavoie: :cry: ok

Issue #3675, opened 2020-08-21
handle_path is not stripping prefix, or root is using the wrong path


ranisalt:
I want to have static images served by a file_server in a subdirectory. I tried several variations of this:

```
handle_path /static/* {
  root * /static
  file_server
}
```
```
handle_path /static/* {
  file_server {
    root /static
  }
}
```
```
handle /static/* {
  uri strip_prefix /static
  root * /static
  file_server
}
```
```
handle /static/* {
  uri strip_prefix /static
  file_server {
    root /static
  }
}
```

The files are correctly there:
```
$ docker-compose exec proxy ls /static -R
/static:
img

/static/img:
image.png
```

However, the file is NOT accessible in `/static/img/image.png`. Instead, it is not stripping the prefix, and serving at `/static/static/img/image.png`. So either `handle_path` is not stripping `/static`, or `root` does not use whatever variable `handle_path` stores the path at.

ranisalt:
```
handle_path /static/* {
  root * /static
  try_files {path}
}
```
Does not work too...

[Documentation for `file_server` is very lacking](https://caddyserver.com/docs/caddyfile/directives/file_server), by the way. I find it unclear what the `[<matcher>]` part is about. If I use `file_server /static/*`, what paths will match? Where will it look for files? Will it modify the path in some way?

mholt:
Hi, thanks for opening an issue.

> However, the file is NOT accessible in /static/img/image.png. Instead, it is not stripping the prefix, and serving at /static/static/img/image.png. So either handle_path is not stripping /static, or root does not use whatever variable handle_path stores the path at.

`handle_path` works as expected for me. What is your full, unmodified/unredacted config file? And what are the exact requests and responses? (Use `curl`) I want to get the same behavior you're seeing.

> Documentation for file_server is very lacking, by the way. I find it unclear what the [<matcher>] part is about.

That's all explained [here on the Directives Syntax](https://caddyserver.com/docs/caddyfile/directives#syntax) section of the docs.

>  If I use file_server /static/*, what paths will match?

This is explained [here in the Matchers](https://caddyserver.com/docs/caddyfile/matchers) page of the docs.

> Where will it look for files? Will it modify the path in some way?

This is explained [on that very page](https://caddyserver.com/docs/caddyfile/directives/file_server) you linked to:

> It works by appending the request's URI path to the [site's root path](https://caddyserver.com/docs/caddyfile/directives/root).

ranisalt:
[...]

PR #3676, opened 2020-08-21
httpcaddyfile: Ensure handle_path is sorted as equal to handle

francislavoie:
Fixes an issue found in #3675

I don't love this fix frankly, any hard-coded logic is 😬 but I'm not sure of a better way to do this. We want `handle` and `handle_path` to have different `dirFunc` logic, but we need them to sort equivalently because they are both just handles, where one is just a shortcut to the other.

Having done this, I think we could probably remove `handle_path` from the directive order since it'll never actually be seen by the sort function.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -862,7 +862,17 @@ func buildSubroute(routes []ConfigValue, groupCounter counter) (*caddyhttp.Subro
 		// root directives would overwrite previously-matched ones; they should not cascade
 		"root": {},
 	}
-	for meDir, info := range mutuallyExclusiveDirs {
+
+	// we want to deterministicly loop over the set of dirs
+	keys := make([]string, 0, len(mutuallyExclusiveDirs))
+	for k := range mutuallyExclusiveDirs {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	for _, meDir := range keys {
+		info := mutuallyExclusiveDirs[meDir]
+
 		// see how many instances of the directive there are
 		for _, r := range routes {
 			if r.directive == meDir {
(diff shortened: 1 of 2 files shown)


francislavoie:
I'm so confused. Why is `groupCounter` flaky? It adapts to `group3` on my machine but CI says it should be `group4`? 😬

francislavoie:
I hate Go's `map`. It sucks. 😠 

Well, now it should be deterministic. So that's helpful.

I ran 10k iterations of the adapt tests before/after to make sure.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -854,7 +863,17 @@ func buildSubroute(routes []ConfigValue, groupCounter counter) (*caddyhttp.Subro
 		// root directives would overwrite previously-matched ones; they should not cascade
 		"root": {},
 	}
-	for meDir, info := range mutuallyExclusiveDirs {
+
+	// we want to deterministicly loop over the set of dirs
```
mholt, line 867:
```suggestion
	// we need to deterministically loop over each of these directives
```

mholt (changes requested):
Just fixing a typo that was bothering me :)

Otherwise LGTM. I can see why this definitely isn't the most elegant fix, but it's pretty simple, hopefully it works well! If not, it's purely internal so we should be able to fix it up again if needed.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -854,7 +863,17 @@ func buildSubroute(routes []ConfigValue, groupCounter counter) (*caddyhttp.Subro
 		// root directives would overwrite previously-matched ones; they should not cascade
 		"root": {},
 	}
-	for meDir, info := range mutuallyExclusiveDirs {
+
+	// we need to deterministically loop over each of these directives
```
mholt, line 867:
```suggestion
	// we need to deterministically loop over each of these directives
	// in order to keep the group numbers consistent
```

mholt (approved):
LGTM, I hate it. 🙃 

Actually though I think this is worth a shot. There are worse parts of the Caddyfile adapter's internals, so...

PR #3682, opened 2020-08-23
caddyfile: Add support for env var defaults, tests

### LINKED ISSUE #2313: Allow users to provide default values for unset environment variables
Reading up on [Environment Variables](https://caddyserver.com/docs/caddyfile#env) I don't see a way to specify default values for environment variables. This is a common feature in similar types of configuration files. Docker for example supports the standard bash syntax:

> `${variable:-word}` indicates that if variable is set then the result will be that value. If variable is not set then word will be the result.

I would very much like to have a self contained `Caddyfile` with sane defaults so you can safely do `caddy` in prod. (I don't want to say "oh, when you deploy in prod you need to follow instruction A, B and C", I would like it to be "Just run `caddy`".) And then something like `DOMAIN=staging caddy` / `DOMAIN=dev caddy` to override relevant parts in staging and dev.

francislavoie:
Closes #2313

This adds support for an optional default in Caddyfile environment variable placeholders with a `:` separator. This is specifically for the ones like `{$FOO}`, not the ones like `{env.FOO}` which are replaced at runtime.

For example:

```
{$DOMAIN:localhost} {
	respond "Hello world!"
}
```

If the `DOMAIN` environment variable exists, the site label will be set to the value of that environment variable, but if not set, will be set to `localhost`.

FYI @mholt I also noticed while working on this that we were doing `os.GetEnv` and feeding that into `os.ExpandEnv`, which actually means that the env var will be replaced twice in chain. I retained this behaviour for now and added a test for it, but it feels like that wasn't intentional. It would technically be a minor breaking change to fix it so I want to wait for your input.

### DIFF
--- a/caddyconfig/caddyfile/parse.go
+++ b/caddyconfig/caddyfile/parse.go
@@ -60,21 +60,31 @@ func replaceEnvVars(input []byte) ([]byte, error) {
 		end += begin + len(spanOpen) // make end relative to input, not begin
 
 		// get the name; if there is no name, skip it
-		envVarName := input[begin+len(spanOpen) : end]
-		if len(envVarName) == 0 {
+		envString := input[begin+len(spanOpen) : end]
+		if len(envString) == 0 {
 			offset = end + len(spanClose)
 			continue
 		}
 
+		// split the string into a key and an optional default
+		envParts := strings.SplitN(string(envString), ":-", 2)
+
+		// do a lookup for the env var, replace with the default if not found
+		envVarValue, found := os.LookupEnv(string(envParts[0]))
+		if !found && len(envParts) == 2 {
+			envVarValue = envParts[1]
+		}
+
 		// get the value of the environment variable
-		envVarValue := []byte(os.ExpandEnv(os.Getenv(string(envVarName))))
+		// note that this causes one-level deep chaining
+		envVarBytes := []byte(os.ExpandEnv(envVarValue))
 
 		// splice in the value
 		input = append(input[:begin],
-			append(envVarValue, input[end+len(spanClose):]...)...)
+			append(envVarBytes, input[end+len(spanClose):]...)...)
 
 		// continue at the end of the replacement
-		offset = begin + len(envVarValue)
+		offset = begin + len(envVarBytes)
 	}
 	return input, nil
 }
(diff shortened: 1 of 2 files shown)


mholt (changes requested):
I guess I'm alright with this. Thanks!

Is the bash-ism `:-` just to keep things easier to remember so you don't have to learn a second syntax? (Because we're def. not bash)

As for the redundant ExpandEnv + Getenv, why is fixing it a breaking change?

Note the reviewdog comment

francislavoie:
> As for the redundant ExpandEnv + Getenv, why is fixing it a breaking change?

Someone could be specifically using that behaviour :man_shrugging: 

But if you're okay with fixing that, I can.

> Is the bash-ism `:-` just to keep things easier to remember so you don't have to learn a second syntax? (Because we're def. not bash)

I guess. What alternative syntax would you suggest?

mholt:
I am more familiar with `||` from several programming languages; JS actually uses that literally as a "otherwise" operator in some cases.

francislavoie:
How about `??`, as in the commonly used null coalescing operator? I think it's more technically correct because we only apply the default if the env doesn't exists, not if it's "falsey".

mholt:
I'd be fine with that too. Honestly I guess I don't care if it's `:-` or `??`. Should we put it to a vote? :P

francislavoie:
Alright decided to go ahead with `??`.

I also removed the "chaining" behaviour since I really doubt anyone actually relies on it, frankly. I just wanted to be safe by preserving it but it should be fine.

mholt:
I think `??` is fine.

> Also we have #3807 in the pipeline which uses ? for header defaults so I think ?? is likely natural to use here.

lol, I also just reopened this thread with the intent of linking to that issue. My OCD is bothered that one uses `?` and the other uses `??` but I'm not against it... still, might be worth asking... can we get them to be the same (`?`) -- would that make sense? Is it worth it?

mholt:
Now I have another idea. Doesn't something like `{$FOO:bar}` read a little more naturally/easier to you? Using `:` as the default operator in this sense. I feel like it reads easier, and reminds me more of ternary statements in a lot of languages, i.e. `A ? B : C` where `: C` indicates the default or fallback value.

Once I settle my mind on the syntax I'm good with merging this.

francislavoie:
Sure, that's fine to me. Just FYI, shell uses `:-` because it can also do other things followed by `:`, see https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

mholt:
K I've made up my mind, let's use `:` :smile: I'm done now, I promise we can merge this then.

mholt (approved):
Oh you already implemented it. K I guess we're good then.

PR #3689, opened 2020-08-26
reverseproxy: Logging for streaming and upgrades

francislavoie:
https://caddy.community/t/caddy-reverse-proxy-to-another-domain/9589/6

Looks like we forgot to set up logging here. I'm not sure if using the sugared logger is ideal, but it's the easiest way to get this working as similarly to the stdlib implementation I think.

### DIFF
--- a/modules/caddyhttp/reverseproxy/streaming.go
+++ b/modules/caddyhttp/reverseproxy/streaming.go
@@ -32,8 +32,7 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
 	reqUpType := upgradeType(req.Header)
 	resUpType := upgradeType(res.Header)
 	if reqUpType != resUpType {
-		// TODO: figure out our own error handling
-		// p.getErrorHandler()(rw, req, fmt.Errorf("backend tried to switch protocol %q when %q was requested", resUpType, reqUpType))
+		h.logger.Sugar().Errorf("backend tried to switch protocol %q when %q was requested", resUpType, reqUpType)
 		return
 	}
 
@@ -41,12 +40,12 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
 
 	hj, ok := rw.(http.Hijacker)
 	if !ok {
-		// p.getErrorHandler()(rw, req, fmt.Errorf("can't switch protocols using non-Hijacker ResponseWriter type %T", rw))
+		h.logger.Sugar().Errorf("can't switch protocols using non-Hijacker ResponseWriter type %T", rw)
 		return
 	}
 	backConn, ok := res.Body.(io.ReadWriteCloser)
 	if !ok {
-		// p.getErrorHandler()(rw, req, fmt.Errorf("internal error: 101 switching protocols response with non-writable body"))
+		h.logger.Sugar().Errorf("internal error: 101 switching protocols response with non-writable body")
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -32,21 +32,20 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
 	reqUpType := upgradeType(req.Header)
 	resUpType := upgradeType(res.Header)
 	if reqUpType != resUpType {
-		// TODO: figure out our own error handling
-		// p.getErrorHandler()(rw, req, fmt.Errorf("backend tried to switch protocol %q when %q was requested", resUpType, reqUpType))
+		h.logger.Sugar().Errorf("backend tried to switch protocol %q when %q was requested", resUpType, reqUpType)
```
mholt, line 35: Not a server error, so maybe DEBUG level

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -32,21 +32,20 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
...
 		return
 	}
 
 	copyHeader(res.Header, rw.Header())
 
 	hj, ok := rw.(http.Hijacker)
 	if !ok {
-		// p.getErrorHandler()(rw, req, fmt.Errorf("can't switch protocols using non-Hijacker ResponseWriter type %T", rw))
+		h.logger.Sugar().Errorf("can't switch protocols using non-Hijacker ResponseWriter type %T", rw)
```
mholt, line 43: Can be ERROR level structured log

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -32,21 +32,20 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
...
 		return
 	}
 	backConn, ok := res.Body.(io.ReadWriteCloser)
 	if !ok {
-		// p.getErrorHandler()(rw, req, fmt.Errorf("internal error: 101 switching protocols response with non-writable body"))
+		h.logger.Sugar().Errorf("internal error: 101 switching protocols response with non-writable body")
```
mholt, line 48: Can be ERROR level structured log

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -65,17 +64,17 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
 
 	conn, brw, err := hj.Hijack()
 	if err != nil {
-		// p.getErrorHandler()(rw, req, fmt.Errorf("Hijack failed on protocol switch: %v", err))
+		h.logger.Sugar().Errorf("Hijack failed on protocol switch: %v", err)
```
mholt, line 67: Can be ERROR level structured log

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -65,17 +64,17 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
...
 		return
 	}
 	defer conn.Close()
 	res.Body = nil // so res.Write only writes the headers; we have res.Body in backConn above
 	if err := res.Write(brw); err != nil {
-		// p.getErrorHandler()(rw, req, fmt.Errorf("response write: %v", err))
+		h.logger.Sugar().Errorf("response write: %v", err)
```
mholt, line 73: Can be DEBUG level structured log, since writes can fail for client reasons, which I bet is more common.

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -65,17 +64,17 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
...
 		return
 	}
 	if err := brw.Flush(); err != nil {
-		// p.getErrorHandler()(rw, req, fmt.Errorf("response flush: %v", err))
+		h.logger.Sugar().Errorf("response flush: %v", err)
```
mholt, line 77: Hmm maybe DEBUG level on this too

modules/caddyhttp/reverseproxy/streaming.go
francislavoie: When you say structured log, do you mean throw away the `%T` bit of the error? Or how do I encode that with zap? I don't think there's a `zap.TypeOf` or whatever... 🤔

modules/caddyhttp/reverseproxy/streaming.go
francislavoie: Should I continue to use the sugar here cause it's using a formatted string?

modules/caddyhttp/reverseproxy/streaming.go
mholt: I mean use `h.Logger.Error(...)` :)

modules/caddyhttp/reverseproxy/streaming.go
mholt: Hmmm well we don't use sugared logs anywhere else in Caddy so I dunno why we need to start here. I usually prefer to keep it consistently structured, even if individual fields require reflection to fill. (which is rare; as are these logs)

mholt (approved):
Thanks for helping with this!

PR #3695, opened 2020-08-31
reverseproxy: return validation error if upstream address has port range size of more than 1

Mohammed90:
Currently the Provision-er of the reverse proxy handler accepts upstream network address with range size more than 1. This means the following Caddyfile is accepted and validates successfully:

```
localhost
reverse_proxy localhost:2000-2020
```

This is despite the comment on the `Dial` field of the `Upstream` struct saying only a single socket is accepted:

https://github.com/caddyserver/caddy/blob/e2f913bb7f813ac2f79cd3644066eff3815accd4/modules/caddyhttp/reverseproxy/hosts.go#L68-L78

The code doesn't validate that until request time, when the method `fillDialInfo` is called:

https://github.com/caddyserver/caddy/blob/e2f913bb7f813ac2f79cd3644066eff3815accd4/modules/caddyhttp/reverseproxy/hosts.go#L161-L164

This PR adds validation of the port range size of an upstream address to the `Provision` step. We can sugar the Caddyfile upstream to accept port range and convert it to the JSON array for the entire specified range.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -204,6 +204,16 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 
 	// set up upstreams
 	for _, upstream := range h.Upstreams {
+		addr, err := caddy.ParseNetworkAddress(upstream.Dial)
+		if err != nil {
+			return err
+		}
+		if addr.PortRangeSize() != 1 {
+			h.logger.Error("multiple addresses (upstream must map to only one address)",
+				zap.String("address", upstream.Dial),
+			)
+			return fmt.Errorf("multiple addresses (upstream must map to only one address): %v", addr)
+		}
 		// create or get the host representation for this upstream
 		var host Host = new(upstreamHost)
 		existingHost, loaded := hosts.LoadOrStore(upstream.String(), host)
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -209,9 +209,6 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 			return err
 		}
 		if addr.PortRangeSize() != 1 {
-			h.logger.Error("multiple addresses (upstream must map to only one address)",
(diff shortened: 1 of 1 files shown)


francislavoie:
Hmm, yeah I think we should continue to support that, not reject it.

Mohammed90:
It didn't work before. It would've just failed an runtime rather than during validation.

francislavoie:
Understood - but we should actually fix it instead of rejecting it :slightly_smiling_face:

Mohammed90:
Hmm okay, how should we handle them? Should we expand them into the upstream slice? This makes it simpler for the selection policy to work.

The other approach is to treat network addresses whose port range size is greater than 1 as a single unit, but the selection policy might need to be applied recursively or at random.

francislavoie:
:thinking: I think the Caddyfile adapter can expand it out. Probably the easiest way to deal with it.

Mohammed90:
Oh so you're saying the JSON should reject upstream network addresses whose port range size is greater than 1, but the Caddyfile should accept it, right? I understood your earlier comment to apply to both JSON and Caddyfile.

francislavoie:
Either that, or we expand it out early at provisioning time so it's supported everywhere?

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -204,6 +204,16 @@ func (h *Handler) Provision(ctx caddy.Context) error {
...
 	// set up upstreams
 	for _, upstream := range h.Upstreams {
+		addr, err := caddy.ParseNetworkAddress(upstream.Dial)
+		if err != nil {
+			return err
+		}
+		if addr.PortRangeSize() != 1 {
+			h.logger.Error("multiple addresses (upstream must map to only one address)",
+				zap.String("address", upstream.Dial),
+			)
```
mholt, line 214: Since we're returning the error with the same content, we don't need to log it here; it will get bubbled up and emitted that way.

Mohammed90: Done

mholt (approved):
Excellent, look good! Thank you as always, Mohammed!

PR #3710, opened 2020-09-04
reverseproxy: Add `buffer_requests` option to `reverse_proxy` directive

francislavoie:
https://caddy.community/t/adding-buffer-requests-to-caddyfile-for-reverse-proxy/9675/2

Issue #3707, opened 2020-09-02
caddyfile: add handle_response blocks to reverse_proxy


maxatome:
Caddyfile support for the `handle_response` option of `reverse_proxy` would be nice.

See the JSON configuration here:

https://caddyserver.com/docs/json/apps/http/servers/errors/routes/handle/reverse_proxy/handle_response/

Maybe it could look something like this in the Caddyfile:

```
reverse_proxy ... {
	...
	handle_response header X-Accel-Redirect {
		... 
	}
	handle_response status 4xx {
		...
	}
	handle_response {
		...
	}
}
```

(Responses can be matched by headers or status code. Just showing a few examples here.)

_Originally posted by @mholt in https://github.com/caddyserver/caddy/issues/3347#issuecomment-683959748_

maxatome:
@mholt do you have an idea of how represent "routes" list in handle_response block?

```
handle_response [header <field>|status <status>] {
    status <status>
    routes?
}
```
I didn't find an other place where it is `UnmarshalCaddyfile`d so I prefer to ask before doing it the wrong way...

francislavoie:
I don't think you'd have `status <status>` inside of it, I think that would need to be isolated to the same-line args because it's not a real handler directive.

But for "routes", take a look at how the `handle` directive is parsed, i.e. `parseHandle` -> `ParseSegmentAsSubroute`:

https://github.com/caddyserver/caddy/blob/24f34780b61fc82df6507cf95535c3dfc9f78058/caddyconfig/httpcaddyfile/directives.go#L268-L278

maxatome:
Thanks @francislavoie, what do you think of #3712?
I am not sure I did it the right way, so tell me...

mholt:
Done in #4021.

PR #3712, opened 2020-09-06
reverseproxy: Add `handle_response` blocks to `reverse_proxy` (#3710)

maxatome:

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -263,6 +263,13 @@ func (h Helper) NewBindAddresses(addrs []string) []ConfigValue {
 	return []ConfigValue{{Class: "bind", Value: addrs}}
 }
 
+// WithDispenser returns a new instance based on d. All others Helper
+// fields are copied, so typically maps are shared with this new instance.
+func (h Helper) WithDispenser(d *caddyfile.Dispenser) Helper {
+	h.Dispenser = d
+	return h
+}
+
 // ParseSegmentAsSubroute parses the segment such that its subdirectives
 // are themselves treated as directives, from which a subroute is built
 // and returned.
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -38,14 +38,14 @@ func init() {
 
 func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error) {
 	rp := new(Handler)
-	err := rp.UnmarshalCaddyfile(h.Dispenser)
+	err := rp.ParseCaddyfileReverseProxy(h)
 	if err != nil {
 		return nil, err
 	}
 	return rp, nil
 }
 
-// UnmarshalCaddyfile sets up the handler from Caddyfile tokens. Syntax:
+// ParseCaddyfileReverseProxy sets up the handler from Caddyfile tokens. Syntax:
 //
 //     reverse_proxy [<matcher>] [<upstreams...>] {
 //         # upstreams
@@ -86,13 +86,20 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 //         transport <name> {
 //             ...
 //         }
+//
+//         # handle responses
+//         handle_response [header <field>|status <status>] {
+//             ...
+//         }
 //     }
 //
 // Proxy upstream addresses should be network dial addresses such
 // as `host:port`, or a URL such as `scheme://host:port`. Scheme
 // and port may be inferred from other parts of the address/URL; if
 // either are missing, defaults to HTTP.
-func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+func (h *Handler) ParseCaddyfileReverseProxy(helper httpcaddyfile.Helper) error {
+	d := helper.Dispenser
+
 	// currently, all backends must use the same scheme/protocol (the
 	// underlying JSON does not yet support per-backend transports)
 	var commonScheme string
@@ -578,6 +585,45 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				}
 				transport = rt
 
+			case "handle_response":
+				var rm caddyhttp.ResponseMatcher
+				args := d.RemainingArgs()
+				switch len(args) {
+				case 2:
+					switch args[0] {
+					case "header":
+						rm = caddyhttp.ResponseMatcher{
+							Headers: http.Header{args[1]: []string{}},
+						}
+					case "status":
+						statusNum, err := strconv.Atoi(args[1])
+						if err != nil {
+							return d.Errf("bad status value '%s': %v", args[1], err)
+						}
+						rm = caddyhttp.ResponseMatcher{
+							StatusCode: []int{statusNum},
+						}
+					default:
+						return d.Err("handle_response only accepts header|status")
+					}
+				case 0: // Any status or header
+				default:
+					return d.ArgErr()
+				}
+				handler, err := httpcaddyfile.ParseSegmentAsSubroute(helper.WithDispenser(d.NewFromNextSegment()))
+				if err != nil {
+					return err
+				}
+				subroute, ok := handler.(*caddyhttp.Subroute)
+				if !ok {
+					return helper.Errf("segment was not parsed as a subroute")
+				}
+				h.HandleResponse = append(h.HandleResponse,
+					caddyhttp.ResponseHandler{
+						Match:  &rm,
+						Routes: subroute.Routes,
+					})
+
 			default:
 				return d.Errf("unrecognized subdirective %s", d.Val())
 			}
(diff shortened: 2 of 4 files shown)


modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -743,6 +789,5 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 // Interface guards
 var (
-	_ caddyfile.Unmarshaler = (*Handler)(nil)
```
mholt, line 746: Why is this no longer a caddyfile Unmarshaler? It needs to be if it is to be used to unmarshal the Caddyfile.

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -265,6 +265,13 @@ func (h Helper) NewBindAddresses(addrs []string) []ConfigValue {
 	return []ConfigValue{{Class: "bind", Value: addrs}}
 }
 
+// WithDispenser returns a new instance based on d. All others Helper
+// fields are copied, so typically maps are shared with this new instance.
+func (h Helper) WithDispenser(d *caddyfile.Dispenser) Helper {
+	h.Dispenser = d
+	return h
+}
```
mholt, line 273: I'm not entirely clear on why this is necessary?

mholt (changes requested):
Thanks for working on a PR!

I just did a quick high-level pass for starters. We shouldn't have to change the dispenser or remove the CaddyfileUnmarshaler interface...

I also noticed that with this approach -- and maybe I'm wrong, feel free to correct me -- only filters on a single status code or header value. That seems like a severe limitation; I think adding more status codes is easily doable but giving proper control over headers is going to need more consideration.

caddyconfig/httpcaddyfile/directives.go
francislavoie: The problem is `httpcaddyfile.ParseSegmentAsSubroute` takes a `Helper` rather than a `Dispenser`, which makes it necessary to chain parsing for other directives.

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: I agree it's not exactly right (because otherwise `reverse_proxy` still need to be a caddyfile.Unmarshaler), but this was done because with those changes, `httpcaddyfile.Helper` is needed to do the parse chaining.

maxatome:
> I also noticed that with this approach -- and maybe I'm wrong, feel free to correct me -- only filters on a single status code or header value. That seems like a severe limitation; I think adding more status codes is easily doable but giving proper control over headers is going to need more consideration.

I pushed a change in `reverse_proxy_handle_response.txt` where several headers and statuses are used: each case generates a new handle_response block.

But perhaps you want to be able to combine several headers/statuses for one handle_response?

I applied the format you proposed in the original issue #3707 but maybe I misunderstood your proposition.

Do you want a handle_response block be like:
```
reverse_proxy ... {
	...
	handle_response header X-Accel-Redirect header X-Another status 401 {
		... 
	}
}
```
? Meaning both `X-Accel-Redirect` and (or "or"?) `X-Another` headers be present with the status 401. When you say several statuses, of course you mean a "or" between them, but for headers it is a "and". I am not sure to fully understand, and so the future caddyfile author.

Feel free to provide me a complete spec of what you want so we can go further.

modules/caddyhttp/reverseproxy/caddyfile.go
maxatome: Thanks @francislavoie, you are right.

@mholt if you tell me how to cope with the absence of `httpcaddyfile.Helper` in `(*Handler).UnmarshalCaddyfile()` (as I need this `httpcaddyfile.Helper` instance to call `httpcaddyfile.ParseSegmentAsSubroute()` as suggested by @francislavoie in #3707), I will be happy to change this.

caddyconfig/httpcaddyfile/directives.go
maxatome: Thanks again @francislavoie, yes this simple method is to simplify code in such cases.

francislavoie:
Just a heads up, https://github.com/caddyserver/caddy/pull/3740 may change the way the parsing of `handle_response` would work; you might need make sure to consume the args on the `handle_response` subdirective line before calling ParseSegmentAsSubroute

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: I did a bit more digging in the source, and actually, I think it's okay for `reverse_proxy` to not be a `caddyfile.Unmarshaler`.

This is because it's never used as a guest module to something else. Regular directives don't strictly need to be Unmarshalers because their parsing is handled by callbacks registered with `RegisterHandlerDirective` or `RegisterDirective`, and those are passed `Helper`.

Most directives that do implement `caddyfile.Unmarshaler` only do so for the registered callbacks to call `UnmarshalCaddyfile`, but that's not actually necessary.

mholt: It's good practice / conventional to implement that, so you _can_ unmarshal it where necessary. But true, all a directive needs is an unmarshaling _function_, not necessarily to implement an interface.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -81,13 +81,20 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
...
+//             ...
+//         }
 //     }
 //
 // Proxy upstream addresses should be network dial addresses such
 // as `host:port`, or a URL such as `scheme://host:port`. Scheme
 // and port may be inferred from other parts of the address/URL; if
 // either are missing, defaults to HTTP.
-func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+func (h *Handler) ParseCaddyfileReverseProxy(helper httpcaddyfile.Helper) error {
```
francislavoie, line 95:
```suggestion
func (h *Handler) parseCaddyfileReverseProxy(helper httpcaddyfile.Helper) error {
```

I don't think this needs to be exported (it's not used anywhere else)

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: I understand @mholt but in this case I don't think it's possible since we actually need to use `Helper` all the way down the callchain for `handle_response` to work.

modules/caddyhttp/reverseproxy/caddyfile.go
maxatome: It is used in [`modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go`](https://github.com/maxatome/caddy/blob/reverse_proxy-handle_response/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go#L286).

francislavoie: Oh, right, makes sense 👍

maxatome:
> Just a heads up, #3740 may change the way the parsing of `handle_response` would work; you might need make sure to consume the args on the `handle_response` subdirective line before calling `ParseSegmentAsSubroute`

Aren't the args consumed by `d.RemainingArgs()`? I thought they were. If not, I'll fix this.

modules/caddyhttp/reverseproxy/caddyfile.go
maxatome: Any news on that topic?

francislavoie:
Sorry to come back so late on this.

> Do you want a handle_response block be like:
> 
> ```
> reverse_proxy ... {
> 	...
> 	handle_response header X-Accel-Redirect header X-Another status 401 {
> 		... 
> 	}
> }
> ```
> 
> ? Meaning both `X-Accel-Redirect` and (or "or"?) `X-Another` headers be present with the status 401. When you say several statuses, of course you mean a "or" between them, but for headers it is a "and". I am not sure to fully understand, and so the future caddyfile author.
> 
> Feel free to provide me a complete spec of what you want so we can go further.

@maxatome So I had a bit of a think 🤔 and I believe what we should do to make this more natural is to reuse the matcher syntax that we already have for this, so like:

```
reverse_proxy ... {
	@accel {
		header X-Accel-Redirect
		header X-Another
		status 401
	}
	handle_response @accel {
		...
	}
}
```

This would be done by storing the named matchers in a map which is then referenced when parsing the `handle_response` subdirective.

I think we could keep the existing approach where if you just need one matcher, you can put it inline, but for anything more complex you'd use a named matcher.

One annoyance is that since response matchers aren't the same as request matchers, we can't quite share that parsing code.

francislavoie:
Also - if you don't have the time to finish this off, let me know and I'd be glad to take it over (you'll still get author credit)

maxatome:
Hi @francislavoie thanks for your reply.
I do not think I will have the time to work on this in the short-term, so if you are ready to take it over, be my guest :)

francislavoie:
Thanks for your work on this @maxatome, opened #4021 to replace this PR!

PR #3729, opened 2020-09-16
Fatal not a valid object name 'GitHub desktop j143/xdiff as submodule' 

jorope:


francislavoie:
This doesn't look right.

PR #3709, opened 2020-09-03
Add support for metrics with Prometheus

hairyhenderson:
_(edited for brevity - see previous edits if you're interested in the history!)_

Fixes #3390 

_At long last_ I've found time to work on this. Sorry about the delay - it's been a bit of a rough summer!

This adds PR does a number of things:
- adds a new `admin.api.metrics` module for scraping metrics on the admin API endpoint at `/metrics`. This is always mounted as long as the admin API is enabled, giving Caddy _automatic_ support for Prometheus metrics.
- adds a new `http.handlers.metrics` module for configuring which port/path metrics should be scraped on. This will be mounted as-needed, and can be used when the admin API is disabled. This module _may_ get more configurable in future PRs.
- instruments all HTTP Middleware modules to track request/error counts, request duration, time-to-first-byte, request size, and response size
- instruments all Admin handlers to track request and error counts
- instruments the Go runtime with the [Go Collector](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus?tab=doc#NewGoCollector)
- exports a `go_build_info` metric (see [`prometheus.NewBuildInfoCollector`](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus?tab=doc#NewBuildInfoCollector))
- adds a new `metrics` directive to Caddyfile

It's worth noting also that this fully enables any module author to register and observe their own custom Prometheus metrics. I can add a brief tutorial in the docs at some point to that end (in short: just register them with `prometheus.DefaultRegisterer`).

Many thanks to @mholt who helped me understand heaps more about the Caddy internals and make decisions on how to put the modules together!

<details>
<summary><b>Expand to see a sample (warning: verbose!)</b></summary>

Config file:

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [":9180"],
          "routes": [
            { 
              "match": [{"path":["/metrics"]}],
              "handle": [{ "handler": "metrics" }]
            },
            {
              "match": [{ "path": ["/foo", "/bar"] }],
              "handle": [{"handler": "static_response", "body": "Hello world!\n"}]
            },
            {
              "match": [{ "path": ["/error"] }],
              "handle": [{ "handler": "error" }]
            },
            {
              "match": [{ "path": ["*"]}],
              "handle": [{"handler":"static_response", "status_code": "404"}]
            }
          ]
        }
      }
    }
  }
}
```

Doing a few requests to seed some metrics observations:

```console
$ curl localhost:9180/foo                                                         
[...]

PR #3733, opened 2020-09-18
metrics: Fixing panic while observing with bad exemplars

hairyhenderson:
This fixes a bug that I introduced in #3709, that @mholt patched in c82c231ba7ec7c400e4d48ab943f57a6b29a2a2a.

The behaviour was that with this `Caddyfile`:

```
localhost
respond "Hello, world!"
```

This happened:

```console
$ curl -v http://localhost
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 80 (#0)
> GET / HTTP/1.1
> Host: localhost
> User-Agent: curl/7.64.1
> Accept: */*
> 
* Empty reply from server
curl: (52) Empty reply from server
* Closing connection 0
```

Effectively, the headers were sent, but then the connection was closed. Based on the metrics now exposed, I discovered that in the `metricsInstrumentedHandler.ServeHTTP` function, the call to `h.mh.ServeHTTP` was not returning at all, since neither the request counter nor the error counter were incrementing. This lead me to believe that there was a panic which was later being swallowed. A quick `defer func() { fmt.Printf(recover()) }()` showed me that the `ObserveWithExemplar` call was failing:

```
exemplar labels have 74 runes, exceeding the limit of 64
```

Effectively, I misunderstood the purpose of the labels in the [ObserveWithExemplar](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#ExemplarObserver), and simply added the same labels being observed.

The reason we didn't get hit by this earlier is that the `server` label was usually quite short, but in the case of HTTP-to-HTTPS redirection, the server is named `remaining_auto_https_redirects`, just pushing the total length of labels past the limit.

To solve this, for now I've simply switched back to using `Observe` instead of `ObserveWithExemplar`. However, once we get tracing added to Caddy, we should consider using this to track a span ID in an exemplar (with appropriate length safeguards, naturally).

Based on conversation with @mholt, I've also changed how the server name is injected into the labels. Instead of being done at provision-time, it's now done at request-time. This isn't strictly necessary, but it's slightly cleaner.

### DIFF
--- a/caddytest/integration/autohttps_test.go
+++ b/caddytest/integration/autohttps_test.go
@@ -1,0 +1,22 @@
+package integration
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/caddyserver/caddy/v2/caddytest"
+)
+
+func TestAutoHTTPtoHTTPSRedirects(t *testing.T) {
+	tester := caddytest.NewTester(t)
+	tester.InitServer(`
+	{
+		http_port     9080
+		https_port    9443
+	}
+	localhost:9443
+	respond "Yahaha! You found me!"
+  `, "caddyfile")
+
+	tester.AssertRedirect("http://localhost:9080/", "https://localhost/", http.StatusPermanentRedirect)
+}
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -155,6 +155,7 @@ func (app *App) Provision(ctx caddy.Context) error {
 
 	// prepare each server
 	for srvName, srv := range app.Servers {
+		srv.name = srvName
 		srv.tlsApp = app.tlsApp
 		srv.logger = app.logger.Named("log")
 		srv.errorLogger = app.logger.Named("log.error")
--- a/modules/caddyhttp/metrics.go
+++ b/modules/caddyhttp/metrics.go
@@ -82,20 +82,14 @@ func initHTTPMetrics() {
 	}, httpLabels)
 }
 
-type ctxKeyServerName struct{}
-
 // serverNameFromContext extracts the current server name from the context.
-// Returns "UNKNOWN" if none is available (should probably never happen?)
+// Returns "UNKNOWN" if none is available (should probably never happen).
 func serverNameFromContext(ctx context.Context) string {
-	srvName, ok := ctx.Value(ctxKeyServerName{}).(string)
-	if !ok {
+	srv, ok := ctx.Value(ServerCtxKey).(*Server)
+	if !ok || srv == nil || srv.name == "" {
 		return "UNKNOWN"
 	}
-	return srvName
-}
-
-func contextWithServerName(ctx context.Context, serverName string) context.Context {
-	return context.WithValue(ctx, ctxKeyServerName{}, serverName)
+	return srv.name
 }
 
 type metricsInstrumentedHandler struct {
--- a/modules/caddyhttp/metrics.go
+++ b/modules/caddyhttp/metrics.go
@@ -94,31 +94,31 @@ func serverNameFromContext(ctx context.Context) string {
 	return srvName
 }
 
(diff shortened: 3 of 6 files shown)


mholt (approved):
Thank you Dave, for working with me to fix this! I appreciate your deep dive to find the problem.

I have pushed a slight cleanup and an integration test. The cleanup is simple: the Server is already part of the context, so we can use that to get the name, rather than adding `s.name` separately. The integration test failed without the patch, and passes with it. This seems like an obvious one to have anyway, and it would have caught this bug much earlier if it had existed. Oh well, now it does!

I'm merging this and will release RC3 now. Thank you for making this so easy to fix quickly!

I would like to follow up though on a few things when we have a chance this week, hopefully before 2.2 final is tagged ~next week-ish.

One thing we will need to take up upstream with Prometheus authors is that sneaky panic: no library should panic based on user input. Instead, an error should be returned or some sane default/override should happen instead of a panic. Literally anything except a panic. Panics in libraries should only happen if it's a _programmer_ error, but in this case it was user input that  "exceeded the limit of 64" (which, by the way, "remaining_auto_https_redirects" does not, it is only 30 characters long) -- either it should return an error or the label should be truncated.

I'm also confused as to why the panic disappeared entirely. I didn't see it any logs; shouldn't it be in a log somewhere? Most other panics in the HTTP handlers are logged very definitely. I want to make sure we don't lose that...

I don't fully understand the difference between observing and observing with exemplar and why the limits are in place but this is something I'd like to learn more about from you when you have a chance.

Thank you again for handling this so well!

hairyhenderson:
> Panics in libraries should only happen if it's a _programmer_ error, but in this case it was user input that "exceeded the limit of 64" (which, by the way, "remaining_auto_https_redirects" does not, it is only 30 characters long) -- either it should return an error or the label should be truncated.

TBH I think in this case it's intended to be a programmer error to have long labels in this case. So it was _absolutely_ programmer in this case 😅. My understanding was totally scrambled on this particular feature, but now I understand a bit better what it's all about.

I think the most succinct explanation of what an exemplar can do is here: https://grafana.com/blog/2020/07/23/the-future-of-prometheus-and-its-ecosystem/#218 (a couple pictures speak a few thousand words in this case!) The proposal for exemplars is [here](https://docs.google.com/document/d/1ymZlc9yuTj8GvZyKz1r3KDRrhaOjZ1W1qZVW_5Gj7gA), with a few use-cases and examples.

Implementation-wise, my understanding is that the label on the exemplar should be a trace ID and possibly a span ID. For the tracing specs I've worked with the trace IDs are up to 128bits (32 hex characters) and span IDs are up to 64bits (16 hex characters), so even those combined would be within the 64-character limit.

Maybe @beorn7 wouldn't mind commenting on this - it seems he had a hand in putting the exemplar support into the Prometheus client...

> I'm also confused as to why the panic disappeared entirely.

Yeah, this was a mystery to me as well. The server will swallow panics that come from `ServeHTTP`, but it's supposed to be logged with the server's error logger (as per the [http.Handler docs](https://golang.org/pkg/net/http/#Handler)). But digging into this a bit more, this is because that particular logger is set to debug level (as of https://github.com/caddyserver/caddy/pull/3668). So, had I altered the log level, I would've probably seen that 😉. It's probably worth revisiting [the decision to make it debug instead of error level!](https://github.com/caddyserver/caddy/pull/3668#issuecomment-677804106)

beorn7:
Yeah, the examplar is definitely not supposed to be generated from user input. The most common use should be something like a trace ID of fixed format and length. From the use cases I thought through, the handling of an error returned by `ObserveWithExemplar` would have boiled down to `if err != nil { panic(fmt.Errorf("this should never happen: %w", err)) }`. I might be wrong, and we can still fix things as exemplar support is marked as experimental.

PR #3740, opened 2020-09-21
caddyconfig: Error on extra args for route and handle directive family

francislavoie:
Brought up here: https://caddy.community/t/v2-how-do-i-selectively-rewrite-multiple-times/9854

Basically, this disallows extra args on the line after the matcher for `route`, `handle`, `handle_errors` and `handle_path`. For example, these are now errors:

```
route @foo bar {
	respond "Foo"
}
```

```
handle bar {
	respond "Foo"
}
```

This is purely an adapter change to be stricter so that users don't have args in their config that do nothing useful (like thinking some args act like a matcher).


mholt (approved):
Thanks, I think this should even go in 2.2.

PR #3742, opened 2020-09-22
metrics: Always track method label in uppercase

hairyhenderson:
This is for consistency. The admin request counter always [sanitizes](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/instrument_server.go#L292) the `method` label by lower-casing it, but the admin error counter and all of the HTTP metrics used the HTTP metric as-is.

This will not do:

```console
$ curl https://localhost/
$ curl -Xget https://localhost/
$ curl -s http://localhost:2019/metrics | grep ^caddy_http_request_duration_seconds_count | grep 'handler="file_server"'
caddy_http_request_duration_seconds_count{code="200",handler="file_server",method="GET",server="srv0"} 1
caddy_http_request_duration_seconds_count{code="200",handler="file_server",method="get",server="srv0"} 1
```

🤦‍♂️

This PR normalizes the method name. In `promhttp.InstrumentHandlerCounter`, the method is lower-cased, but that's contrary to the HTTP RFCs ([RFC 7230 s3.1.1](https://tools.ietf.org/html/rfc7230#section-3.1.1) and [RFC 7231 s4.1](https://tools.ietf.org/html/rfc7231#section-4.1)), and prevailing convention. So I've chosen to upper-case the methods instead.

Side note - I'm not actually sure why the methods are lower-cased instead of upper-cased in the `promhttp` code, but that's the way it's been for [the past ~3yrs](https://github.com/prometheus/client_golang/pull/285/commits/90dca5d416dd7e1fcbbb1238cddaf1a134f2720a#diff-9c0c00018610d6efebb26be9d72ed9d8R229). Maybe this is for a good reason and @stuartnelson3 or @beorn7 could enlighten me? 😅

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -46,7 +46,7 @@ func instrumentHandlerCounter(counter *prometheus.CounterVec, next http.Handler)
 		next.ServeHTTP(d, r)
 		counter.With(prometheus.Labels{
 			"code":   sanitizeCode(d.status),
-			"method": sanitizeMethod(r.Method),
+			"method": strings.ToUpper(r.Method),
 		}).Inc()
 	})
 }
@@ -73,22 +73,5 @@ func sanitizeCode(s int) string {
 		return "200"
 	default:
 		return strconv.Itoa(s)
-	}
-}
-
-func sanitizeMethod(m string) string {
-	switch m {
-	case "GET", "get":
-		return "GET"
-	case "PUT", "put":
-		return "PUT"
-	case "HEAD", "head":
-		return "HEAD"
-	case "POST", "post":
-		return "POST"
-	case "DELETE", "delete":
-		return "DELETE"
-	default:
-		return strings.ToUpper(m)
 	}
 }
--- a/metrics.go
+++ b/metrics.go
@@ -1,6 +1,10 @@
 package caddy
 
 import (
+	"net/http"
+	"strconv"
+	"strings"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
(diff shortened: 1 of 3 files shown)


metrics.go
```diff
@@ -30,3 +34,61 @@ var adminMetrics = struct {
...
+	case "PUT", "put":
+		return "PUT"
+	case "HEAD", "head":
+		return "HEAD"
+	case "POST", "post":
+		return "POST"
+	case "DELETE", "delete":
+		return "DELETE"
+	default:
+		return strings.ToUpper(m)
```
francislavoie, line 92: Is it actually faster to use a switch rather than just always running `strings.ToUpper`? :thinking:

beorn7:
https://github.com/prometheus/client_golang/issues/387 – or in other words: v2 of client_golang will use upper case methods. It was a mistake to use lower case, but at the time we had that revelation, the library was already too widespread.

metrics.go
beorn7: Yes. (o:

hairyhenderson: @francislavoie I had that same question (this code is copied from the Prometheus client), and wrote (then threw away) some benchmarks.

For `sanitizeMethod`, the switch-based benchmark was actually _slightly_ slower (by ~8ns - they're both _around_ 150ns on my laptop), but with fewer allocations. I was a bit surprised about the speed TBH.

For `sanitizeCode`, at least the one in the Prometheus client (I didn't copy it here originally), the switch-based benchmark is about twice as fast time-wise (~120ns for the `if` plus `strconv.Itoa` vs. ~60ns for the switch-based one), and also with fewer allocations.

What I might do (eventually) is pull both the `sanitizeCode` and `sanitizeMethod` functions into an internal package so I don't end up with different implementations for different purposes. But, later. 😉

hairyhenderson: I also just noticed https://github.com/prometheus/client_golang/pull/447#discussion_r225461218 where @beorn7 has effectively explained the same thing, but with a lot more clarity 😉

hairyhenderson:
@beorn7 looks like that fix went in to the `dev-0.10` branch, while client_go is at 1.7.1 - I'm guessing the `dev-0.10` branch never made it into the 0.10 release?

hairyhenderson:
Ah - there never _was_ a v0.10 - v1.0 was released right after. And I'm noticing now that there's a `dev-v2` branch and indeed the change is [in there](https://github.com/prometheus/client_golang/blob/dev-v2/prometheus/promhttp/instrument_server.go#L341).

beorn7:
That's pre-Go-modules version numbering. The 0.10 branch is now called the [dev-v2 branch](https://github.com/prometheus/client_golang/tree/dev-v2). Development thereof is pretty much on hold anyway because the plan is to wait for OpenMetrics to publish at least a preliminary spec so that v2 can be designed around OpenMetrics from the start. However, that hasn't happened yet.

So client_golang v1 is using lower-case methods, and v2 will use upper-case methods. Once v2 is in sight, some helpers to facilitate the transition might be in order.

metrics.go
mholt: Let's not try and be more clever than the standard library -- just use `strings.ToUpper()`.

modules/caddyhttp/metrics.go
```diff
@@ -108,7 +109,8 @@ func newMetricsInstrumentedHandler(handler string, mh MiddlewareHandler) *metric
 func (h *metricsInstrumentedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request, next Handler) error {
 	server := serverNameFromContext(r.Context())
 	labels := prometheus.Labels{"server": server, "handler": h.handler}
-	statusLabels := prometheus.Labels{"server": server, "handler": h.handler, "method": r.Method, "code": ""}
+	method := sanitizeMethod(r.Method)
+	statusLabels := prometheus.Labels{"server": server, "handler": h.handler, "method": method, "code": ""}
```
mholt, line 113: In hindsight, I wonder if we should add a comment as to why "code" is `""` (empty string) -- that is significant, right? It looks like a mistake at first glance (but it's not).

modules/caddyhttp/metrics.go
```diff
@@ -181,3 +182,28 @@ func computeApproximateRequestSize(r *http.Request) int {
 	}
 	return s
 }
+
+// taken from https://github.com/prometheus/client_golang/blob/9ac0bad6062845400019e68d4fa0bf0384b5b6f6/prometheus/promhttp/instrument_server.go#L322
+// and modified to uppercase the method.
+func sanitizeMethod(m string) string {
```
mholt, line 188: Simiolarly, just use `strings.ToUpper()`. Probably don't even need a separate function for this (although Go's compiler will inline it, I bet).

mholt (changes requested):
Thanks, let's try to get this in before the 2.2 release then; hopefully tomorrow.

metrics.go
hairyhenderson: Ok. We can always go back to something like this if we need to reduce allocations later on.

hairyhenderson:
Comments addressed - @mholt PTAL!

mholt (approved):
LGTM, thanks very much Dave!

PR #3747, opened 2020-09-23
feat: add a module implementing Vulcain

dunglas:
[Vulcain](https://vulcain.rocks) is an open protocol (currently an [Internet-Draft](https://datatracker.ietf.org/doc/draft-dunglas-vulcain/)) allowing to use HTTP/2 Server Push to create fast and idiomatic client-driven REST APIs.

A schema is worth a thousand words:

image

This proposed module allows to turn any existing web API (hypermedia or not thanks to the OpenAPI integration) into a Vulcain enabled one. It is a companion of the `reverseproxy` and of the (upcoming) `cache` module. With these three modules combined, you can dramatically increase the performance of any web API.

The protocol is still subject to change, so I suggest to add this module as experimental. Having support for Vulcain directly into Caddy will make it very easy to any Caddy user to leverage the new capabilities offered by HTTP/2 and HTTP/3.


francislavoie:
🤔 So, yes, this is super cool, but I don't think it makes sense to have built-in. I think it should be a plugin.

It's very easy to build Caddy with plugins, so there's not really any downside to having it separate. It's generally preferred for things to start as plugins, and if down the line, we see that there's a lot of interest, we could consider making it a default module.

Having it built-in also adds it as a maintenance burden to @mholt which isn't ideal.

dunglas:
Hi @francislavoie. This module can definitely be hosted somewhere else.

Actually, I plan to deprecate the existing "Vulcain Gateway Server" (https://github.com/dunglas/vulcain) (a reverse proxy doing exactly the same thing than this module) and recommend to use Caddy + this module instead.

I thought that it could maybe be mutually interesting. It's obviously interesting for the users of the existing server because Caddy has much more awesome features (the current server is minimalist).
But it could also be interesting for Caddy to be the default solution for HTTP/2+ web APIs (which become more and more popular): you have an existing API and you want to turn it into a HTTP/2 Server Push enabled one? Just download Caddy server, add the `vulcain` directive to `Caddyfile` and done: thanks to Caddy you just transformed your existing web API in a very fast and modern one leveraging the advanced features of HTTP/2 (and even 3) 😄.

That being said, I totally understand the maintenance overhead that including such module can cause (even if it is marked as experimental, as the HTTP/3 one). And I'll be glad to just provide an alternative build called "Vulcain Gateway Server v2" if you think that it's better. (And of course, feel free to close this PR!).

Anyway, thank you very much for this awesome piece of software.

PS: I made a mistake in the domain name (^^). It's https://vulcain.rocks (not .org) 😅

mholt:
Oh dang, yeah this is very cool.

As Francis said, though, we're pretty strict on what we bring into this repo now as new modules (or dependencies, or any new code for that matter) because it has caused problems before in Caddy 1; with how extensible Caddy is, it makes more sense to leave most things out by default -- bringing them in is easy, taking them out later is very hard. And distributing/deploying Caddy with plugins is very common and routine anyway. :)

I would really love to see this integration -- just not in the core Caddy repo.

Can we find another way to promote the combination of Vulcain and Caddy?

dunglas:
This sounds great!

> Can we find another way to promote the combination of Vulcain and Caddy?

Sure! Maybe can we have a quick talk about that (my Twitter DM are open: @dunglas, or kevin [at] dunglas [dot] fr).

RobinHoutevelts:
( Just chiming in )

Vulcain looks amazing. Caddy looks amazing. I'm a developer for a web agency. We are using nginx because - for us - Caddy just doesn't provide anything extra we'd need. While we do feel we'd benefit from switching to Caddy, it just doesn't _scratch an itch_. And again, that's just for us, if we'd start all over we probably would be using Caddy.

Anyway, yesterday I came across Vulcain while doing research for an upcoming project but told myself "I'm never going to be able to sell this" and was ready to drop it entirely.

To my surprise I find this PR and now I'm _feeling an itch that needs to be scratched_.

---

I'm hoping you two are able to work something out. At least know I ( and probably a bunch of others, It just seems I found out about this by luck ) wouldn't mind to see a love child of both projects! 😁

mholt:
Looks like this is done then! Very cool!

PR #3756, opened 2020-09-26
reverseproxy: fix breakage in handling SRV lookup introduced by d55d50b

### LINKED ISSUE #3753: upgrade to v2.2.0 break SRV Caddyfile
I have an entry in Caddyfile like this

```
auto-keypoints.ml-service.aquabyte.ai {
    reverse_proxy srv+auto-keypoints-bridge-mode.local
}
```

It works great in v2.1.1. After I upgraded, it does not pass validation

```
validate: loading http app module: provision http: server srv0: setting up route handlers: route 0: loading handler modules: position 0: loading module 'subroute': provision http.handlers.subroute: setting up subroutes: route 0: loading handler modules: position 1: loading module 'reverse_proxy': provision http.handlers.reverse_proxy: missing port in address
```

I should not need a port. Caddy used to be able to find it from the SRV record. I added it just in case. But the validation error remains the same.

Mohammed90:
This is the naive fix. I still need to test it some more and figure whether this fix is the best.
@waiyip-aquabyte and @danlsgiga, is it possible to test using the CI artifacts? You can find them [here](https://github.com/caddyserver/caddy/actions/runs/273408022).

Fixes #3753

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -132,7 +132,7 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 	// get validation out of the way
 	for i, v := range h.Upstreams {
 		// Having LookupSRV non-empty conflicts with 2 other config parameters: active health checks, and upstream dial address.
-		// Therefore if LookupSRV is empty, then there are no immediately apparent config conflict and the iteration can be skipped.
+		// Therefore if LookupSRV is empty, then there are no immediately apparent config conflicts, and the iteration can be skipped.
 		if v.LookupSRV == "" {
 			continue
 		}
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -204,17 +204,18 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 
 	// set up upstreams
 	for _, upstream := range h.Upstreams {
-		addr, err := caddy.ParseNetworkAddress(upstream.Dial)
-		if err != nil {
-			return err
+		if upstream.LookupSRV == "" {
+			addr, err := caddy.ParseNetworkAddress(upstream.Dial)
+			if err != nil {
+				return err
+			}
+
+			if addr.PortRangeSize() != 1 {
+				return fmt.Errorf("multiple addresses (upstream must map to only one address): %v", addr)
+			}
+
+			upstream.networkAddress = addr
 		}
-
-		if addr.PortRangeSize() != 1 {
-			return fmt.Errorf("multiple addresses (upstream must map to only one address): %v", addr)
-		}
-
-		upstream.networkAddress = addr
-
 		// create or get the host representation for this upstream
 		var host Host = new(upstreamHost)
 		existingHost, loaded := hosts.LoadOrStore(upstream.String(), host)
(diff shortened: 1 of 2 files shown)


mholt (approved):
Awesome, thanks for making time to help fix this!

FWIW, I think it looks fine; I'd have probably done something similar. Since health checks don't make sense on srv upstreams, should we take this opportunity to make it an error?

danlsgiga:
Awesome @Mohammed90, I validated the build and it fixes the issue!

waiyip-aquabyte:
Thank you so much. What is the release process? I really want to upgrade because of Prometheus metrics.

danlsgiga:
@mholt is it possible to have this into a 2.2.1 release whenever its merged?

mholt (approved):
This fix LGTM! Are we comfortable merging it then? Will wait for the `do not merge` and `in progress` tags to be removed.

Mohammed90:
> This fix LGTM! Are we comfortable merging it then? Will wait for the `do not merge` and `in progress` tags to be removed.

Just want to add a few validations for incompatible config and few integration tests. I should be able to shoot them later tonight.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -129,6 +129,20 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 	h.ctx = ctx
 	h.logger = ctx.Logger(h)
 
+	// get validation out of the way
+	for i, v := range h.Upstreams {
+		// the special cases arise with LookupSRV present, so skip the iteration if it's empty.
```
mholt, line 134: What are "the special cases"? Maybe explain more here, or the comment can be removed.

mholt (approved):
Tests look great! Willing to merge this, my only nit is with a comment. Thanks!

modules/caddyhttp/reverseproxy/reverseproxy.go
Mohammed90: Active healthchecks being incompatible with lookup_srv and specifying both dial and lookup_srv. I thought it's better to explain why we're skipping the iteration.

Mohammed90: Now?

mholt (dismissed):
Yep, that'll work! As usual, thank you sooo much!

mholt (dismissed):

mholt:
CI is failing apparently because Microsoft is having a major outage right now, so GitHub is naturally affected...

francislavoie:
Okay that's pretty stupid, it looks like it didn't dismiss the failing jobs after rerunning them all... 🤔 I think if you just force push again @Mohammed90 it should be okay (force re-run CI)

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -129,6 +129,21 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 	h.ctx = ctx
 	h.logger = ctx.Logger(h)
 
+	// get validation out of the way
+	for i, v := range h.Upstreams {
+		// Having LookupSRV non-empty conflicts with 2 other config parameters: active health checks, and upstream dial address.
+		// Therefore if LookupSRV is empty, then there are no immediately apparent config conflict and the iteration can be skipped.
```
francislavoie, line 135:
```suggestion
		// Therefore if LookupSRV is empty, then there are no immediately apparent config conflicts, and the iteration can be skipped.
```

or

```suggestion
		// Therefore if LookupSRV is empty, then there is no immediately apparent config conflict, and the iteration can be skipped.
```

francislavoie (approved):

mholt (approved):
Thanks for all the helpful fixes @Mohammed90 :grin:

lorenckrzysztof:
Waszyngton

Issue #3585, opened 2020-07-17
request: additional tls client certificate placeholders


panva:
As far as I can see this is available to set as values for upstream servers, something you'd usually do when terminating required/optional mutual-TLS and needed to work with the certificate in the upstream application.

```
// `{http.request.tls.client.fingerprint}` | The SHA256 checksum of the client certificate
// `{http.request.tls.client.public_key}` | The public key of the client certificate.
// `{http.request.tls.client.public_key_sha256}` | The SHA256 checksum of the client's public key.
// `{http.request.tls.client.issuer}` | The issuer DN of the client certificate
// `{http.request.tls.client.serial}` | The serial number of the client certificate
// `{http.request.tls.client.subject}` | The subject DN of the client certificate
// `{http.request.tls.client.san.dns_names.*}` | SAN DNS names(index optional)
// `{http.request.tls.client.san.emails.*}` | SAN email addresses (index optional)
// `{http.request.tls.client.san.ips.*}` | SAN IP addresses (index optional)
// `{http.request.tls.client.san.uris.*}` | SAN URIs (index optional)
```

For OAuth 2.0 servers with mTLS support to be deployed behind Caddy I lack the following

- PEM-encoded value of the certificate (not just the public key in it)
- whether the client certificate was verified or not (in case of client_auth mode being `request`)

This is analogous [apache's variables](https://httpd.apache.org/docs/current/mod/mod_ssl.html#envvars) SSL_CLIENT_CERT, SSL_CLIENT_CERT_CHAIN and SSL_CLIENT_VERIFY (values NONE, SUCCESS, GENEROUS or FAILED:reason)

mholt:
@panva Thanks for the feature request. I'm sure we or someone will be able to work that in.

FYI:

> whether the client certificate was verified or not (in case of client_auth mode being `request`)

Note [the docs on the `request` mode](https://caddyserver.com/docs/json/apps/http/servers/tls_connection_policies/#client_authentication/mode):

> Ask clients for a certificate, but allow even if there isn't one; do not verify it

So a client cert used with "request" mode will never be verified.

panva:
Which is fine for self signed certs only. What mode would attempt to verify based kn configured CA but allow unverified requests through as well?

mholt:
> Which is fine for self signed certs only.

I am not sure why that has anything to do with it; a cert's signer is orthogonal to client authentication modes.

> What mode would attempt to verify based kn configured CA but allow unverified requests through as well?

[...]

PR #3757, opened 2020-09-27
caddyhttp: Added `is_verified` placeholder for client cert

gdhameeja:
Uses x509.Verify for verification.
For #3585 

Question: In the test, the certificate is not verified (expected value is `false`) because I'm assuming certificate used here is actually not verified right?

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -99,6 +99,8 @@ func (st ServerType) Setup(inputServerBlocks []caddyfile.ServerBlock,
 		"{tls_client_issuer}", "{http.request.tls.client.issuer}",
 		"{tls_client_serial}", "{http.request.tls.client.serial}",
 		"{tls_client_subject}", "{http.request.tls.client.subject}",
+		"{tls_client_placeholder}", "{http.request.tls.client.certificate_pem}",
+		"{tls_client_is_verified}", "{http.request.tls.client.is_verified}",
 	)
 
 	// these are placeholders that allow a user-defined final
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -78,6 +78,8 @@ func init() {
 // `{http.request.tls.client.issuer}` | The issuer DN of the client certificate
 // `{http.request.tls.client.serial}` | The serial number of the client certificate
 // `{http.request.tls.client.subject}` | The subject DN of the client certificate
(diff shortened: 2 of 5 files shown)


francislavoie:
If you could add another cert that would be - verified for the purposes of tests, that'd be good. Please also document somewhere the commands you used to generate it so it can be regenerated later if need be.

I just thought - we should make sure the expiry on these certs is very very long because we don't want to suddenly get hit by failing tests because ⏳

gdhameeja:
> If you could add another cert that would be - verified for the purposes of tests, that'd be good. Please also document somewhere the commands you used to generate it so it can be regenerated later if need be.
> 
> I just thought - we should make sure the expiry on these certs is very very long because we don't want to suddenly get hit by failing tests because

Added the test. Could you please review this again.

francislavoie:
Could you add a comment in the test with the command you used to generate that test cert, in case it needs to be re-generated later?

Also, please add this new placeholder to the godoc comment in `modules/caddyhttp/app.go` along with the rest.

gdhameeja:
> Could you add a comment in the test with the command you used to generate that test cert, in case it needs to be re-generated later?
> 
> Also, please add this new placeholder to the godoc comment in `modules/caddyhttp/app.go` along with the rest.
> 
> You could also add a new Caddyfile shortcut in `caddyconfig/httpcaddyfile/httptype.go` around line 100ish.

Added. Could you please review this again?

mholt:
Thanks for the PR.

As I understand it, this is re-verifying the certificate independently of whether it was already verified during client authentication. Is that really the right logic? I feel like we shouldn't be doing the verification a second time.

gdhameeja:
> Thanks for the PR.
> 
> As I understand it, this is re-verifying the certificate independently of whether it was already verified during client authentication. Is that really the right logic? I feel like we shouldn't be doing the verification a second time.

I will get around this fixing this PR over the weekend. Any suggestions if we're storing the result somewhere if certificate was verified during client authentication? Or if there is some other way to check the same? And I suppose, if the cert is not verified, we can use the function in the current PR and store the result?

francislavoie:
The replacer has a `Set()` function, I think we could set the verified result that way when verification actually happens, instead of re-validating when the placeholder is used.

gdhameeja:
This can be closed as the issue is now closed.

francislavoie:
Actually @gdhameeja this PR still has a couple things we'd still want, i.e. the godoc comments for the `certificate_pem` placeholder. if you want to strip the rest out of the PR, we can merge that part.

PR #3758, opened 2020-09-27
reverseproxy: ignore RFC 1521 parameters in mime type when checking the Content-Type header

### LINKED ISSUE #3765: Server Sent Events are not working in a nodejs/express app
With Caddy 2.2.0, the responses from the server are only sent unbuffered when the content type is `text/event-stream`. NodeJS express sends out a Content Type of `text/event-stream; charset=utf-8`. 

I am not so deep in the specs to know if caddy or express is at fault. I would expect caddy to gracefully accept any arguments to the `content-type: text/event-stream`.

# Workaround 1
It is possible to get it working with express (by not using express, but the underlaying nodejs function calls directly).

```
    res.writeHead(200, {
      'Content-Type': 'text/event-stream',
      'Cache-Control': 'no-cache',
      'X-Accel-Buffering': 'no',
      'Transfer-Encoding': 'chunked',
    });
```

# Workaraound 2
Setting `flush_interval` to `-1` in the Caddyfile works as well.

cmfcmf:
Without this change, a Content-Type header like "text/event-stream;charset=utf-8"
would not trigger the immediate flushing.

Fixes #3765

### DIFF
--- a/modules/caddyhttp/reverseproxy/streaming.go
+++ b/modules/caddyhttp/reverseproxy/streaming.go
@@ -21,6 +21,7 @@ package reverseproxy
 import (
 	"context"
 	"io"
+	"mime"
 	"net/http"
 	"sync"
 	"time"
@@ -88,11 +89,12 @@ func (h Handler) handleUpgradeResponse(rw http.ResponseWriter, req *http.Request
 // flushInterval returns the p.FlushInterval value, conditionally
 // overriding its value for a specific request/response.
 func (h Handler) flushInterval(req *http.Request, res *http.Response) time.Duration {
-	resCT := res.Header.Get("Content-Type")
+	resCTHeader := res.Header.Get("Content-Type")
+	resCT, _, err := mime.ParseMediaType(resCTHeader)
 
 	// For Server-Sent Events responses, flush immediately.
 	// The MIME type is defined in https://www.w3.org/TR/eventsource/#text-event-stream
-	if resCT == "text/event-stream" {
+	if err == nil && resCT == "text/event-stream" {
 		return -1 // negative means immediately
 	}
 


mholt:
Thanks for the PR! I will look at this soon.

... and I'm kinda beginning to wonder if -1 should be the default for everything at this point. Thoughts, @francislavoie / @Mohammed90 ?

francislavoie:
Idk, that sounds like too risky a change to me 🤔

mholt:
Why?

francislavoie:
Isn't there risk that we break behaviour for some clients? There must've been a good reason for it not to be set to -1 by default in stdlib, no?

The code comes from here https://golang.org/src/net/http/httputil/reverseproxy.go?h=FlushInterval#L410

Should we bubble this change upstream to see what they have to say? I know our proxy implementation is not dependent on it but it feels like we should let them know of fixes like this?

mholt:
> Isn't there risk that we break behaviour for some clients? There must've been a good reason for it not to be set to -1 by default in stdlib, no?

We don't know if it is likely to _break_ anyone; the worst I imagine could happen would be a performance degradation. (I mean, I'm sure it could break _someone_ but I doubt it'd be the majority case.)

Anyway, that's tangential, just thought I'd bring it up...

> Should we bubble this change upstream to see what they have to say? I know our proxy implementation is not dependent on it but it feels like we should let them know of fixes like this?

Sure. Sounds like a good idea.

mholt (approved):
Thanks for the fix @cmfcmf !

cmfcmf:
> > Should we bubble this change upstream to see what they have to say? I know our proxy implementation is not dependent on it but it feels like we should let them know of fixes like this?
> 
> Sure. Sounds like a good idea.

I have now opened an upstream issue about this fix: https://github.com/golang/go/issues/47359

Issue #3763, opened 2020-09-29
lookup_srv wrong behaviour for "no such host" errors


danlsgiga:
The expected error for DNS lookup errors in `lookup_srv` should be the same as `dial`.

### How to reproduce

```json
{
  "admin": {
    "disabled": true
  },
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "errors": {
            "routes": [
              {
                "handle": [                  
                  {
                    "handler": "rewrite",
                    "uri": "/502.html"
                  },
                  {
                    "handler": "file_server",
                    "hide": [
                      "caddy.json"
                    ]
                  }
                ],
                "match": [
                  {
                    "expression": "{http.error.status_code} == 502"
                  }
                ],
                "terminal": true
              }
            ]
          },
          "listen": [
            ":443"
          ],          
          "routes": [
            {
              "handle": [                
                {
                  "handler": "reverse_proxy",
                  "upstreams": [
                    {
                      "lookup_srv": "wguoihebfoindfoibnwpoefjoienbfgiberish.com"
                    }
                  ]
                }
              ],
              "match": [
                {
                  "host": [
                    "localhost"
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    }
  }  
}
```

This config will fail to render the `502` error page because Caddy is returning a `500` due to the `no such host` error from the DNS lookup failure.

However, if you change `"lookup_srv": "wguoihebfoindfoibnwpoefjoienbfgiberish.com"` to `"dial": "wguoihebfoindfoibnwpoefjoienbfgiberish.com:80"` Caddy will render the `502` error page as expected.

Both failures should result in a `502`

lorenckrzysztof:
super prgramy

aleert:
Ok, looks like i fixed it locally. I'll check if everything else if fine and then make a PR.

aleert:
#3771 
It renders 502 error page both for `lookup_srv` and `dial` config entry, however i build from #3756 to actually check `lookup_srv` since it's not working on master.

danlsgiga:
I think that's a simple and effective fix! Thanks for that @aleert

PR #3771, opened 2020-10-01
reverseproxy: Change 500 error to 502 for lookup_srv config (#3763)

aleert:
This should be checked with fix-srv-3753 branch due to #3753 bug.


mholt (approved):
I like this change, it's simple and it makes sense. Thanks for the fix!

Issue #3768, opened 2020-10-01
Add an upstream for reverse_proxy handler dynamicly?


yflau:
Refer to https://caddy.community/t/custom-select-policy-for-reverse-proxy-module/9963
If caddy can provide a public method to add upstream dynamic, then we can just configure just one reverse_proxy handler, and the upstream info will be passed from previous stage（which maybe a API gateway and get upstream information from database or some others).

mholt:
You can already do this by using a placeholder such as `{http.request.header.X-Caddy-Upstream-Dial}` -- just be aware of the implications with health checks if those are configured.

francislavoie:
@mholt I think you missed the point of the discussion in the forum thread, that's not enough for what they're looking for.

francislavoie:
Hmm, well I guess you're right. I suppose the only difference is that `MaxRequests` can't be placeholder'd, but that probably doesn't matter.

yflau:
Cool, but there is another problem, it seems that the port must be specified like this, for example, 

`Dial: {http.request.header.X-Caddy-Upstream-Dial}:80`

thus, if the upstream uses  different ports, then this solution not work, if we can use a default port 80 when not specified in Dial?

mholt:
@yflau Why is the gateway setting a dial address without a port? That seems like a bug. Dial addresses necessarily have ports. If not, it's not an address to dial. (or it's a unix socket)

> Second, all different upstreams become one upstream in such solution, and the health check and circuit breaker will become a global switch, which is not meet the need.

Please note [the documentation](https://caddyserver.com/docs/modules/http.handlers.reverse_proxy#upstreams/dial) which addresses this:

> The network address to dial to connect to the upstream. Must represent precisely one socket (i.e. no port ranges). A valid network address either has a host and port or is a unix socket address.
>
> Placeholders may be used to make the upstream dynamic, but be aware of the health check implications of this: a single upstream that represents numerous (perhaps arbitrary) backends can be considered down if one or enough of the arbitrary backends is down. Also be aware of open proxy vulnerabilities.

yflau:
@mholt 
First problem, what I mean is this: 

`
upstreams": [{
    "dial": "{http.request.header.X-Caddy-Upstream-Dial}"
}
`

Then in Provision phase, will get the error:

`
provision http.handlers.reverse_proxy: address {http.request.header.X-Caddy-Upstream-Dial}: missing port in address"
`

[...]

PR #3776, opened 2020-10-03
reverseproxy: restore allowing upstream to contain replaceable placeholders

Mohammed90:
Another regression caused by d55d50b

Updates #3768

### DIFF
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -101,6 +102,7 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	// to write tests before making any more changes to it
 	upstreamDialAddress := func(upstreamAddr string) (string, error) {
 		var network, scheme, host, port string
+		placeholder := regexp.MustCompile(`\{[[:graph:]]+\}`)
 
 		if strings.Contains(upstreamAddr, "://") {
 			// we get a parsing error if a placeholder is specified
@@ -155,7 +157,8 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 			if err != nil {
 				host = upstreamAddr
 			}
-			if port == "" {
+
+			if !placeholder.MatchString(host) && port == "" {
 				port = "80"
 			}
 		}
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -174,6 +177,12 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 		if network != "" {
 			return caddy.JoinNetworkAddress(network, host, port), nil
 		}
+		// if both network and port are empty, host address contains
+		// placeholdler (e.g. "{http.request.header.X-Caddy-Upstream-Dial}") and doing net.JoinHostPort(host, port) will
```
francislavoie, line 181:
```suggestion
		// placeholder (e.g. "{http.request.header.X-Caddy-Upstream-Dial}") and doing net.JoinHostPort(host, port) will
```

Mohammed90: I'm surprised golangci-lint didn't catch it 🤨

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -129,8 +129,12 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 	h.ctx = ctx
 	h.logger = ctx.Logger(h)
 
+	re := regexp.MustCompile(`\{[[:graph:]]+\}`)
```
mholt, line 132: Not gonna lie, I do hate this :laughing: Let me think on this for a bit and see if we can come up with something more elegant...

This didn't used to be a problem, right? Did we _add_ code that broke this use case? Can we not remove that code? Or at least change it?

Mohammed90: Yeah, the code added to fix the healthcheck on alternate port requires parsing the network address, which errors out for not having a port when using placeholders. I've moved that part into later in the `Provision` method where it only sets up the health checks. Now if the upstream contains a placeholder dial address, it will not accept active healthcheck config because it's like SRV and owned by another entity.

mholt: Do you think there's a way we can fix the original health checking bug without requiring host/port information up front? That would simplify the code, rather than complicate it.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -101,6 +102,7 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	// to write tests before making any more changes to it
 	upstreamDialAddress := func(upstreamAddr string) (string, error) {
 		var network, scheme, host, port string
+		placeholder := regexp.MustCompile(`\{[[:graph:]]+\}`)
```
mholt, line 105: Why does this logic/check have to be repeated in the Caddyfile adapter?

(If we keep it, this should be brought out into a package-level variable like `var placeholderPattern = ...` or something, then reused...)

mholt:
Thanks so much for working on this. But the growing complexity resulting from the original fix is starting to make me nervous / give me a headache. :sweat_smile:  I'm hoping we can revisit the original problem and find a better solution. I'll try taking a look at it later today.

mholt:
I think this will be mostly superseded by #3780, but not before we bring over the new integration tests this PR has (except for one, see linked PR).

Issue #3748, opened 2020-09-23
500 error for probably cancelled request (must be 499)


ont:
## Problem
There are strange 500 errors in caddy access log. After some research we suppose that this errors related to cancelled requests. Currently we have problem to separate cancelled requests from real 500 errors.

## Possible solution
Nginx use 499 error code for such situation.

## Research
Example logs entries related to 500 error during `GET /favicon.ico`:
```json
{"level":"error","ts":1600849986.420132,"logger":"http.log.error.log0","msg":"making dial info: lookup backend.service.consul on 123.123.123.123:53: dial udp 123.123.123.123:53: operation was canceled","request":{"method":"GET","uri":"/favicon.ico","proto":"HTTP/1.1","remote_addr":"222.222.222.222:38284","host":"some-site.com","headers":{"Accept":["image/webp,image/apng,image/*,*/*;q=0.8"],"Referer":["https://some-site.com/en/terms"],"Accept-Encoding":["gzip, deflate"],"Accept-Language":["id-ID,en-US;q=0.9"],"Cookie":["session=ffffffffffffffffffffffffffffffff"],"Connection":["keep-alive"],"User-Agent":["Mozilla/5.0 (Linux; Android 10; vivo 1818; wv) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.84 Mobile Safari/537.36 VivoBrowser/6.7.0.2"]},"tls":{"resumed":true,"version":771,"ciphersuite":49196,"proto":"http/1.1","proto_mutual":true,"server_name":"some-site.com"}},"duration":0.000442598} 
[...]

PR #3779, opened 2020-10-06
logging: Add error message to access logs.

ont:
Related to #3748

Error message in access log can be used in monitoring/alerting systems.
It can be used for example for filtering out "unimportant" 500 errors from access log.

My personal opinion: probably it is better to add special field `"cancelled": true` into access log when request was cancelled by user.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -158,6 +158,9 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	errLog := s.errorLogger.With(loggableReq)
 
 	var duration time.Duration
+	var errStatus int
+	var errMsg string
+	var errFields []zapcore.Field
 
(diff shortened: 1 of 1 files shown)


mholt:
Thanks, this is an interesting idea.

Typically, error logs and access logs are separate. I suppose with structured logs, this is less critical, but we have separate error logs already and it doesn't make much sense to overlap them.

We can probably add `canceled: true` to the access logs if the request is canceled, I think I like that idea more.

Upon context cancellation, a specific error value is returned, `context.Canceled`: https://golang.org/pkg/context/#pkg-variables

We can use that -- carefully, and in the right places -- to determine if the client canceled the connection. But we have to be sure we know that it was the client, not anything else...

mholt:
Superceded, as noted in https://github.com/caddyserver/caddy/issues/3748#issuecomment-733940048, by https://github.com/caddyserver/caddy/pull/3816

PR #3781, opened 2020-10-07
caddyhttp: Restore original request params before error handlers

### LINKED ISSUE #3717: handle_errors in a site block with a path matcher
This ticket follows [a question](https://caddy.community/t/handle-errors-unexpected-behavior/9753) in the Caddy Forum.

I have the following `Caddyfile` where I want to handle errors in a site block with a path matcher:

```
http://localhost:2020/www* {
    root www
    uri strip_prefix /www
    file_server
    handle_errors {
        rewrite * /404.html
        file_server
    }
}
```

I found that this code does not work and that the `handle_errors` directive is ignored. To fix it, I have to move the `handle_errors` to a site block without a path matcher:

```
http://localhost:2020/www* {
    root www
    uri strip_prefix /www
    file_server
}
http://localhost:2020 {
    handle_errors {
        rewrite * /404.html
        file_server
    }
}
```

mholt:
Fixes #3717

If the primary handler chain rewrote the request, it needs to be restored before the error handler chain gets invoked.

Right now the actual specifics of how this should work are undefined, but I think it makes sense that the primary chain, upon returning an error, should invoke the error handler chain without side-effects on the request. Works well for me.

@aitva could you please try this out? You can get build artifacts from CI.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -209,6 +209,13 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// restore original request before invoking error handler chain (issue #3717)
+	origReq := r.Context().Value(OriginalRequestCtxKey).(http.Request)
+	r.Method = origReq.Method
+	r.RemoteAddr = origReq.RemoteAddr
+	r.RequestURI = origReq.RequestURI
+	cloneURL(origReq.URL, r.URL)
+
 	// prepare the error log
 	logger := errLog
 	if s.Logs != nil {


modules/caddyhttp/server.go
```diff
@@ -209,6 +209,13 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// restore original request before invoking error handler chain (issue #3717)
+	origReq := r.Context().Value(OriginalRequestCtxKey).(http.Request)
+	r.Method = origReq.Method
+	r.RemoteAddr = origReq.RemoteAddr
+	r.RequestURI = origReq.RequestURI
+	cloneURL(origReq.URL, r.URL)
+
```
francislavoie, line 218: What about the request headers if someone uses `request_header`? :thinking:

mholt: Good point... we don't currently copy those into the OriginalReq as it is, and I've never heard of a request (pun intended) for this. Because it's a map of variable size I am cautious of its performance implications, esp. since some requests can be large (with cookies and what-not).

Maybe we just wait and see if it's a problem for anyone?

francislavoie: Yeah that's fair. Should we just add a TODO comment or whatever here to note that?

mholt: Sure, done

francislavoie (approved):

mholt:
Pinging @aitva -- can you try this out please?

mholt:
Well, going to merge this anyway. Time to tag a release. Hopefully it works for you.

xnaas:
While this PR is merged, for the sake of keeping relevant info linked: https://caddy.community/t/cant-message-users-here-reaching-out-via-thread/11063/26

PR #3785, opened 2020-10-08
docs: Add link to website for download instructions in README

coolaj86:


francislavoie:
I think this would be a better fit in the website docs: https://github.com/caddyserver/website/blob/master/src/docs/markdown/download.md

coolaj86:
I updated this commit to link to the install options.

mholt (approved):
LGTM, thanks!

PR #3794, opened 2020-10-11
ci: Use golangci's github action for linting

hairyhenderson:
I suspect that the current golangci-lint action (https://github.com/reviewdog/action-golangci-lint) isn't working quite right... I ran `golangci-lint run` locally and got a whole bunch of lint errors that I haven't observed in CI. Also golangci has a lint action available, so we should probably use that anyway!

I've also altered the config to be explicit about which linters are active (and sorted the list).

It'd probably not be a bad idea to turn on a few of the disabled linters, but that can definitely go to a separate PR.

I realize there are a lot of changes here! I'm open to creating separate PRs with some of the fixes separately 😉


modules/caddyhttp/metrics.go
```diff
@@ -122,7 +122,7 @@ func (h *metricsInstrumentedHandler) ServeHTTP(w http.ResponseWriter, r *http.Re
 
 	// This is a _bit_ of a hack - it depends on the ShouldBufferFunc always
 	// being called when the headers are written.
-	// Effectively the same behaviour as promhttp.InstrumentHandlerTimeToWriteHeader.
+	// Effectively the same behavior as promhttp.InstrumentHandlerTimeToWriteHeader.
```
francislavoie, line 125: 🙄 lol

hairyhenderson: I just realized - the problem here is that `locale` is explicitly set to `US` - but it's not an either/or setting! After looking at the docs I'm realizing that if I simply unset it, it'll allow both varieties 😬 🇨🇦🇺🇸

cmd/proc_posix.go
```diff
@@ -1,37 +0,0 @@
-// Copyright 2015 Matthew Holt and The Caddy Authors
```
hairyhenderson, line 1: neither of the functions in this file were referenced (dead code)

cmd/proc_windows.go
```diff
@@ -1,44 +0,0 @@
-// Copyright 2015 Matthew Holt and The Caddy Authors
```
hairyhenderson, line 1: neither of the functions in this file were referenced (dead code)

modules/caddyhttp/caddyhttp.go
```diff
@@ -18,18 +18,14 @@ import (
...
 	"net"
 	"net/http"
 	"strconv"
-	"time"
 
 	"github.com/caddyserver/caddy/v2"
 )
 
 func init() {
-	weakrand.Seed(time.Now().UnixNano())
```
hairyhenderson, line 31: moved to `errors.go` which is the only places `math/rand` is used in this package

modules/caddyhttp/metrics.go
francislavoie: amazing 😁

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -330,7 +330,12 @@ func (h *Handler) Cleanup() error {
 
 	// remove hosts from our config from the pool
 	for _, upstream := range h.Upstreams {
-		hosts.Delete(upstream.String())
+		// XXX: not sure if this will ever error - it's only possible if the
+		// value is a caddy.Destructor
+		_, err := hosts.Delete(upstream.String())
+		if err != nil {
+			return err
+		}
```
hairyhenderson, line 338: (commented, but wanted to bring attention)

I'm not sure if this will ever actually error in this code-path - it's only possible if the value is a `caddy.Destructor`, and the only usage I found was in the root package. So, I could just do this instead:

```suggestion
		_, _ = hosts.Delete(upstream.String())
```

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -137,9 +137,9 @@ func (h Handler) copyResponse(dst io.Writer, src io.Reader, flushInterval time.D
 		}
 	}
 
-	buf := streamingBufPool.Get().([]byte)
+	buf := streamingBufPool.Get().(*[]byte)
 	defer streamingBufPool.Put(buf)
-	_, err := h.copyBuffer(dst, src, buf)
+	_, err := h.copyBuffer(dst, src, *buf)
```
hairyhenderson, line 142: `*[]byte` is certainly awkward vs `[]byte`, but makes for almost a 50% reduction in allocations (see comment below), and speeds this up too.

I benchmarked this and got these results:

```
BenchmarkHandlerCopyResponseOld-8   	   31369	     37921 ns/op	  132364 B/op	       8 allocs/op
BenchmarkHandlerCopyResponse-8      	   31453	     36648 ns/op	  132209 B/op	       5 allocs/op
```

Note that I also tried out a `*bytes.Buffer` (which is the usual approach, and faster/fewer allocs than a `[]byte`) but using a `*[]byte` is fastest.

modules/caddypki/ca.go
```diff
@@ -309,14 +309,19 @@ func (ca CA) loadOrGenIntermediate(rootCert *x509.Certificate, rootKey interface
...
+	if rootKey == nil {
+		rootKeyPEM, err := ca.storage.Load(ca.storageKeyRootKey())
+		if err != nil {
+			return nil, nil, fmt.Errorf("loading root key to sign new intermediate: %v", err)
+		}
+		rootKey, err = pemDecodePrivateKey(rootKeyPEM)
+		if err != nil {
+			return nil, nil, fmt.Errorf("decoding root key: %v", err)
+		}
 	}
```
hairyhenderson, line 323: I put a comment in the code on this, but want to surface it here too - the `rootKey` passed into the function was never used before (overwritten before first use) - I made an assumption here on how the behaviour maybe should be, but this bears confirmation...

modules/caddytls/internalissuer.go
```diff
@@ -114,12 +119,16 @@ func (iss InternalIssuer) Issue(ctx context.Context, csr *x509.CertificateReques
...
+		lifetime = time.Until(issuerCert.NotAfter)
+		iss.logger.Warn("cert lifetime would exceed issuer NotAfter, clamping lifetime",
+			zap.Duration("orig_lifetime", time.Duration(iss.Lifetime)),
+			zap.Duration("lifetime", lifetime),
+			zap.Time("not_after", issuerCert.NotAfter),
+		)
 	}
 
-	certChain, err := auth.Sign(csr, provisioner.SignOptions{}, customCertLifetime(iss.Lifetime))
+	certChain, err := auth.Sign(csr, provisioner.SignOptions{}, customCertLifetime(caddy.Duration(lifetime)))
```
hairyhenderson, line 131: there was a bug here before, though I'm guessing the number of times the cert's lifetime would've needed to be clamped was pretty low 😉

caddytest/caddytest.go
```diff
@@ -272,7 +273,7 @@ func CreateTestingTransport() *http.Transport {
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   5 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
-		TLSClientConfig:       &tls.Config{InsecureSkipVerify: true},
+		TLSClientConfig:       &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
```
Mohammed90, line 276: Do you think it's better to have this comment or add this exclusion rule to the `.golangci.yml`?
```yml
    - path: caddytest/caddytest.go
      text: "G402" # G402: TLS InsecureSkipVerify set true. (gosec)
      linters:
        - gosec
```

I'm conflicted myself

.golangci.yml
```diff
@@ -1,21 +1,68 @@
 linters-settings:
   errcheck:
-    ignore: fmt:.*,io/ioutil:^Read.*,github.com/caddyserver/caddy/v2/caddyconfig:RegisterAdapter,github.com/caddyserver/caddy/v2:RegisterModule
+    ignore: fmt:.*,io/ioutil:^Read.*,github.com/caddyserver/caddy/v2/caddyconfig:RegisterAdapter,github.com/caddyserver/caddy/v2:RegisterModule,go.uber.org/zap/zapcore:^Add.*
```
Mohammed90, line 3: `RegisterModule` and `RegisterAdapter` no longer return errors

```suggestion
    ignore: fmt:.*,io/ioutil:^Read.*,go.uber.org/zap/zapcore:^Add.*
```

Mohammed90 (commented):
It took me an embarrassing amount of time to realize there weren't any false-positive lints because there aren't any and not because VSCode was misbehaving with golangci-lint! 😄 I had to enable one of the disable linters to ensure VSCode indeed functions as expected.

Love the added test too!

modules/caddypki/ca.go
mholt: Oh, you know, we could probably just use it directly. A quick code search shows that `genIntermediate()` should always called with a non-nil `rootKey` value. I think we can remove this bit entirely, since leaving it in might cover up a bug.

modules/caddytls/connpolicy.go
```diff
@@ -80,6 +80,7 @@ func (cp ConnectionPolicies) TLSConfig(ctx caddy.Context) *tls.Config {
 	}
 
 	return &tls.Config{
+		MinVersion: tls.VersionTLS12,
```
mholt, line 83: Why add this here?

modules/caddytls/distributedstek/distributedstek.go
```diff
@@ -145,7 +145,12 @@ func (s *Provider) storeSTEK(dstek distributedSTEK) error {
 // current STEK is outdated (NextRotation time is in the past),
 // then it is rotated and persisted. The resulting STEK is returned.
 func (s *Provider) getSTEK() (distributedSTEK, error) {
-	s.storage.Lock(s.ctx, stekLockName)
+	err := s.storage.Lock(s.ctx, stekLockName)
+	if err != nil {
+		return distributedSTEK{}, fmt.Errorf("failed to acquire storage lock: %w", err)
```
mholt, line 150:
```suggestion
		return distributedSTEK{}, fmt.Errorf("failed to acquire storage lock: %v", err)
```

modules/caddytls/internalissuer.go
```diff
@@ -52,6 +53,8 @@ type InternalIssuer struct {
 	SignWithRoot bool `json:"sign_with_root,omitempty"`
 
 	ca *caddypki.CA
+
```
mholt, line 56:
```suggestion
```

modules/caddytls/internalissuer.go
mholt: Great catch, thanks.

modules/caddytls/internalissuer.go
```diff
@@ -114,12 +119,16 @@ func (iss InternalIssuer) Issue(ctx context.Context, csr *x509.CertificateReques
 
 	// ensure issued certificate does not expire later than its issuer
 	lifetime := time.Duration(iss.Lifetime)
-	if time.Now().Add(lifetime).After(issuerCert.NotAfter) {
-		// TODO: log this
-		lifetime = issuerCert.NotAfter.Sub(time.Now())
+	if time.Now().Add(time.Duration(iss.Lifetime)).After(issuerCert.NotAfter) {
```
mholt, line 122: Shouldn't this use `lifetime`?

modules/caddytls/values.go
```diff
@@ -122,6 +122,7 @@ var SupportedProtocols = map[string]uint16{
 // unsupportedProtocols is a map of unsupported protocols.
 // Used for logging only, not enforcement.
 var unsupportedProtocols = map[string]uint16{
+	//nolint:staticcheck
```
mholt, line 125: What's this about?

mholt (changes requested):
That's quite a change!

Thanks for all the little improvements. I have questions about several, and suggestions for others.

Can we just drop the one linter than throws warnings every time we use a random number? That seems a bit obnoxious.

Anyway, 30+ comments later, let me know what you think. Thanks!!

modules/caddyhttp/reverseproxy/streaming.go
hairyhenderson: Oh yes, SA6002 indeed. I didn't commit the benchmarks since they would be referencing functions that I deleted, but here they are:

```go
func BenchmarkHandlerCopyResponseOld(b *testing.B) {
	h := Handler{}

	testdata := []string{
		"",
		strings.Repeat("a", defaultBufferSize),
		strings.Repeat("123456789 123456789 123456789 12", 3000),
	}
	dst := new(bytes.Buffer)
	for i := 0; i < b.N; i++ {
		for _, d := range testdata {
			src := bytes.NewBuffer([]byte(d))
			dst.Reset()
			h.copyResponseOld(dst, src, 0)
		}
	}
}

func BenchmarkHandlerCopyResponse(b *testing.B) {
	h := Handler{}
	testdata := []string{
		"",
		strings.Repeat("a", defaultBufferSize),
		strings.Repeat("123456789 123456789 123456789 12", 3000),
	}
	dst := new(bytes.Buffer)
	for i := 0; i < b.N; i++ {
		for _, d := range testdata {
			src := bytes.NewBuffer([]byte(d))
			dst.Reset()
			h.copyResponse(dst, src, 0)
		}
	}
}
```

I agree about the insignificance of the speedup - but the allocation reduction was the main reason I included the change 😉

modules/caddypki/ca.go
hairyhenderson: Ok, I'll remove it:

```suggestion
```

caddytest/caddytest.go
hairyhenderson: Personally I prefer the inline `//nolint` comments, because they work in all cases, and don't depend on the presence of one particular config. They also communicate that there's something fishy going on here, and we accept the risk 😉

hairyhenderson:
> Can we just drop the one linter than throws warnings every time we use a random number? That seems a bit obnoxious.

@mholt personally I agree with the linter's reasoning - it's always safer to just default to `crypto/rand`, and `math/rand` is really only useful when some vague randomness is required in a very hot path. My understanding is that `crypto/rand` is only _slightly_ less performant, especially on systems that have crypto hardware (most modern CPUs).

The unfortunate bit is that the `crypto/rand` API sucks compared to `math/rand`... There is https://github.com/FiloSottile/mostly-harmless/tree/master/cryptosource which makes the API a bit better, but that would likely still trigger the linter.

.golangci.yml
```diff
@@ -1,21 +1,68 @@
...
     - ineffassign
     - misspell
+    - prealloc
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unconvert
+    - unused
+    - varcheck
+  # these are implicitly disabled:
```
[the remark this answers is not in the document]
hairyhenderson: Nope! Mostly I left this for future-me so that I can uncomment at random and see which we could reasonably enable later. Some are quite useful (`golint` and `unparam` in particular) so those'll likely follow in a followup.

admin.go
```diff
@@ -838,7 +843,7 @@ var (
 // will get deleted before the process gracefully exits.
 func PIDFile(filename string) error {
 	pid := []byte(strconv.Itoa(os.Getpid()) + "\n")
-	err := ioutil.WriteFile(filename, pid, 0644)
+	err := ioutil.WriteFile(filename, pid, 0600)
```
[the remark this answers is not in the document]
hairyhenderson: This was from:

```
G306: Expect WriteFile permissions to be 0600 or less (gosec)
```

pidfiles only _need_ to be read/writeable by the user running the process. I've seen 644, 640 and 600 for these, but I agree with the linter - the best approach is to take the secure default: group/world-unreadable.

caddytest/caddytest.go
```diff
@@ -221,10 +221,11 @@ func isCaddyAdminRunning() error {
 	client := &http.Client{
 		Timeout: Default.LoadRequestTimeout,
 	}
-	_, err := client.Get(fmt.Sprintf("http://localhost:%d/config/", Default.AdminPort))
+	resp, err := client.Get(fmt.Sprintf("http://localhost:%d/config/", Default.AdminPort))
 	if err != nil {
 		return errors.New("caddy integration test caddy server not running. Expected to be listening on localhost:2019")
 	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
hairyhenderson: I agree, though IMO it's a bad habit to _not_ `defer` it, so since there's zero overhead (since Go 1.14) I think it's best to just leave it with the `defer` 😉

cmd/main.go
```diff
@@ -235,8 +235,11 @@ func watchConfigFile(filename, adapterName string) {
 		return false
 	}
 
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
 	// begin poller
-	for range time.Tick(1 * time.Second) {
+	for range ticker.C {
```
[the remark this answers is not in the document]
hairyhenderson: Yeah... I flip-flopped here. This was from [SA1015](https://staticcheck.io/docs/checks#SA1015):

```
SA1015: using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here (staticcheck)
```

The reason it's triggering is because this isn't the `main` package. I'll ignore this one instead.

logging.go
```diff
@@ -458,7 +458,7 @@ func (cl *CustomLog) buildCore() {
 		if cl.Sampling.Thereafter == 0 {
 			cl.Sampling.Thereafter = 100
 		}
-		c = zapcore.NewSampler(c, cl.Sampling.Interval,
+		c = zapcore.NewSamplerWithOptions(c, cl.Sampling.Interval,
```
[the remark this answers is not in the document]
hairyhenderson: `NewSampler` is deprecated:

```
Deprecated: use NewSamplerWithOptions.
```

It was caught by `staticcheck`:

```
SA1019: zapcore.NewSampler is deprecated: use NewSamplerWithOptions.  (staticcheck)
```

modules/caddyhttp/caddyauth/basicauth.go
```diff
@@ -224,6 +224,7 @@ func (c *Cache) makeRoom() {
 		// map with less code, this is a heavily skewed eviction
 		// strategy; generating random numbers is cheap and
 		// ensures a much better distribution.
+		//nolint:gosec
```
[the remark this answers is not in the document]
hairyhenderson: Yup - it's not able to tell the difference 😉 (also see my [comment](https://github.com/caddyserver/caddy/pull/3794#issuecomment-707685247) on this)

modules/caddyhttp/encode/encode.go
```diff
@@ -262,7 +262,7 @@ func acceptedEncodings(r *http.Request) []string {
 		return []string{}
 	}
 
-	var prefs []encodingPreference
+	prefs := []encodingPreference{}
```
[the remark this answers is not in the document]
hairyhenderson: IMO `prefs := []encodingPreference{}` is somewhat _more_ explicit about the fact that it's initializing an empty array. This was caught by the `prealloc` linter. If you're explicit with an actual empty array then it'll stay silent. The code below discards some of the encodings, so we can't preallocate with an actual size here.

modules/caddyhttp/errors.go
```diff
@@ -92,7 +97,8 @@ func randString(n int, sameCase bool) string {
 	}
 	b := make([]byte, n)
 	for i := range b {
-		b[i] = dict[mathrand.Int63()%int64(len(dict))]
+		//nolint:gosec
```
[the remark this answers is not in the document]
hairyhenderson: It wasn't named `weakrand` here (I changed that), and see https://github.com/caddyserver/caddy/pull/3794#issuecomment-707685247

modules/caddyhttp/fileserver/browse.go
```diff
@@ -82,9 +82,9 @@ func (fsrv *FileServer) serveBrowse(dirPath string, w http.ResponseWriter, r *ht
 		w.Header().Set("Content-Type", "text/html; charset=utf-8")
 	}
 
-	buf.WriteTo(w)
+	_, err = buf.WriteTo(w)
```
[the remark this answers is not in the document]
hairyhenderson: Ok, fair enough. I'll ignore it.

modules/caddyhttp/fileserver/browselisting.go
```diff
@@ -30,10 +30,8 @@ import (
 func (fsrv *FileServer) directoryListing(files []os.FileInfo, canGoUp bool, urlPath string, repl *caddy.Replacer) browseListing {
 	filesToHide := fsrv.transformHidePaths(repl)
 
-	var (
-		fileInfos           []fileInfo
-		dirCount, fileCount int
-	)
+	var dirCount, fileCount int
+	fileInfos := []fileInfo{}
```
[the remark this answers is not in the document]
hairyhenderson: 😂 This was `prealloc` again.

How about:

```suggestion
	dirCount, fileCount := 0, 0
	fileInfos := []fileInfo{}
```

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -226,7 +226,10 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 		}
 		w.WriteHeader(statusCode)
 		if r.Method != "HEAD" {
-			io.Copy(w, file)
+			_, err = io.Copy(w, file)
+			if err != nil {
+				return caddyhttp.Error(http.StatusInternalServerError, err)
+			}
```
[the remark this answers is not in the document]
hairyhenderson: Ok, will ignore.

modules/caddyhttp/marshalers.go
```diff
@@ -75,7 +75,8 @@ func (t LoggableTLSConnState) MarshalLogObject(enc zapcore.ObjectEncoder) error
 	enc.AddUint16("version", t.Version)
 	enc.AddUint16("cipher_suite", t.CipherSuite)
 	enc.AddString("proto", t.NegotiatedProtocol)
-	enc.AddBool("proto_mutual", t.NegotiatedProtocolIsMutual)
+	// NegotiatedProtocolIsMutual is deprecated - it's always true
+	enc.AddBool("proto_mutual", true)
```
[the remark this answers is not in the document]
hairyhenderson: The [docs say](https://github.com/golang/go/blob/release-branch.go1.15/src/crypto/tls/common.go#L238):

```
// Deprecated: this value is always true.
``` 

I'm wondering if this should simply be removed instead?

modules/caddyhttp/reverseproxy/fastcgi/client.go
```diff
@@ -242,13 +242,6 @@ func (c *FCGIClient) writeBeginRequest(role uint16, flags uint8) error {
 	return c.writeRecord(BeginRequest, b[:])
 }
 
-func (c *FCGIClient) writeEndRequest(appStatus int, protocolStatus uint8) error {
-	b := make([]byte, 8)
-	binary.BigEndian.PutUint32(b, uint32(appStatus))
-	b[4] = protocolStatus
-	return c.writeRecord(EndRequest, b)
-}
-
```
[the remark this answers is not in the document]
hairyhenderson: Good question! Looks like the fork is lost (grr bitbucket) 😞

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -170,6 +170,7 @@ func (h *HTTPTransport) NewTransport(ctx caddy.Context) (*http.Transport, error)
 		dialer.Resolver = &net.Resolver{
 			PreferGo: true,
 			Dial: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				//nolint:gosec
```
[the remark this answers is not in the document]
hairyhenderson: I'd rather not, `gosec` is super useful... We could disable the specific rule (`G404`), but I'd also rather not...

Maybe a good option would be to make a central `internal/rand` package and use Filippo's `cryptosource` there?

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -355,7 +360,10 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyht
 		buf := bufPool.Get().(*bytes.Buffer)
 		buf.Reset()
 		defer bufPool.Put(buf)
-		io.Copy(buf, r.Body)
+		_, err := io.Copy(buf, r.Body)
+		if err != nil {
+			return fmt.Errorf("failed to buffer body: %w", err)
```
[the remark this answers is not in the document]
hairyhenderson: > if we return here we don't close the body and that would be a leak

Yikes, good point - and the `r.Body.Close()` wasn't deferred first... I'll just ignore this instead.

modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -132,11 +132,11 @@ func (ash *Handler) Provision(ctx caddy.Context) error {
...
-	acmeAuth, err := acme.NewAuthority(
-		auth.GetDatabase().(nosql.DB),     // stores all the server state
-		ash.Host,                          // used for directory links; TODO: not needed
-		strings.Trim(ash.PathPrefix, "/"), // used for directory links
-		auth)                              // configures the signing authority
+	acmeAuth, err := acme.New(auth, acme.AuthorityOptions{
+		DB:     auth.GetDatabase().(nosql.DB),     // stores all the server state
+		DNS:    ash.Host,                          // used for directory links; TODO: not needed
+		Prefix: strings.Trim(ash.PathPrefix, "/"), // used for directory links
+	})
```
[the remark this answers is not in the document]
hairyhenderson: This was caught by `staticcheck`'s `SA1019` - `NewAuthority` is deprecated (https://github.com/smallstep/certificates/blob/master/acme/authority.go#L85)

modules/caddytls/connpolicy.go
hairyhenderson: This is `G402: TLS MinVersion too low. (gosec)`

modules/caddytls/distributedstek/distributedstek.go
hairyhenderson: Why not wrap it? IMO errors are much more useful when they can be unwrapped and `errors.Is` can be used on it...

modules/caddytls/internalissuer.go
hairyhenderson: You're right - this was left over from some experimentation I was doing 😅

modules/caddytls/values.go
hairyhenderson: This is from:

```
SA1019: tls.VersionSSL30 is deprecated: SSLv3 is cryptographically broken, and is no longer supported by this package. See golang.org/issue/32716.  (staticcheck)
```

Obviously we _don't_ support ssl3.0, but it's cringing at the mere mention of it 😉 (and fairly IMO)

mholt (approved):
Hey sorry it took me a while!

I think this is looking good. I don't remember looking through everything but GitHub's review tracker says I've viewed all 39/39 files. :sweat_smile:  Shall we merge it?

There is a merge conflict that needs to be resolved first.

Mohammed90 (dismissed):

modules/caddytls/distributedstek/distributedstek.go
mholt: Because then it exposes that error value to the exported API, and I don't want us or anyone to rely on those unless we have a specific reason to expose them. It adds to our maintenance burden.

mholt (approved):
Thanks very much! Really appreciate your time and attention to this. I'll merge this in now.

PR #3802, opened 2020-10-16
caddyfile: Implement heredoc support

francislavoie:
Adds support for heredoc syntax to the Caddyfile, inspired by HCL (HashiCorp Config Language), and PHP 7.3's flexible syntax (trimming the front of every line based on the indentation of the ending marker: https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes)

Here's an example:

```
example.com {
    respond <<EOF
    <html>
      <head><title>Foo</title>
      <body>Foo</body>
    </html>
    EOF 200
}
```

Adapted JSON:

```json
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"match": [
								{
									"host": [
										"example.com"
									]
								}
							],
							"handle": [
								{
									"handler": "subroute",
									"routes": [
										{
											"handle": [
												{
													"body": "\u003chtml\u003e\n  \u003chead\u003e\u003ctitle\u003eFoo\u003c/title\u003e\n  \u003cbody\u003eFoo\u003c/body\u003e\n\u003c/html\u003e\n",
													"handler": "static_response",
													"status_code": 200
												}
											]
										}
									]
								}
							],
							"terminal": true
						}
					]
				}
			}
		}
	}
}
```

Notice that the response body has the leading whitespace on each line stripped away.

---

The implementation was a bit funky here, I don't like that I added a new field on the Token struct, but I couldn't think of a better way to handle it, frankly.

Basically, heredocs don't include the newline that counts as the starting point as part of the token contents. This is ideal, because then you don't have an extra newline at the start of every token, so it reads more like a file.

The problem is this: the logic for splitting the tokens into segments involves counting the amount of newlines inside of the token text, and checking whether it matches the line numbers on the tokens. Since we drop one newline from the heredoc, this doesn't match that expectation.

So basically my "hack" was to just add a "+1" marker on the token, with a field called "Heredoc" for the purposes of counting line breaks, while keeping the line numbers intact (because they're useful for debugging etc). Can you think of a better way?

I also went ahead and commented the shit out of the lexer, because I think it deserves it, since it's getting pretty involved. The heredoc stuff is admittedly quite chunky, and probably non-optimally written cause I'm by no means a Go expert... but I think the logic is sound.

I should add some more tests for this to `lexer_test.go` but I'm too tired right now :smile: I'll come back to it soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ vendor
 dist
 caddy-build
 caddy-dist
+
+# ide specific
+.vscode/
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -378,7 +378,7 @@ func (d *Dispenser) numLineBreaks(tknIdx int) int {
 	if tknIdx < 0 || tknIdx >= len(d.tokens) {
 		return 0
 	}
-	return strings.Count(d.tokens[tknIdx].Text, "\n")
+	return strings.Count(d.tokens[tknIdx].Text, "\n") + d.tokens[tknIdx].Heredoc
 }
 
 // isNewLine determines whether the current token is on a different
--- a/caddyconfig/caddyfile/lexer.go
+++ b/caddyconfig/caddyfile/lexer.go
@@ -155,8 +155,9 @@ func (l *lexer) next() bool {
 						panic(fmt.Errorf("mismatched whitespace in heredoc on line #%d [%s], expected whitespace [%s]", i, line, paddingToStrip))
 					}
 
-					// strip, then append the line, with the newline, to the output
-					out += line[len(paddingToStrip):] + "\n"
+					// strip, then append the line, with the newline, to the output.
+					// also removes all "\r" because Windows.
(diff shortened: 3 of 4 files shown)


mholt:
So I know I've been sitting on this for a while, letting it stew in my brain... and I don't really like the added complexity for very very little benefit; a very niche use case. I don't like heredoc in bash, I don't like it in PHP, I definitely don't like it in the Caddyfile. That's mostly just my opinion.

Heredoc runs counter to the simplicity of the Caddyfile.

Not ungrateful for the contribution, but something we should have probably discussed first (as with all changes/PRs).

I'm disinclined to merge this at this point, because, as you say, the implementation is a bit clunky, and I'm not sure its cost justifies the limited benefit over quotes and backticks that we already have.

Still, thank you for working on this! It's nice to see the PoC I guess, so we can refer back to it in the future.

Will close, but feel free to discuss first if there's anything else.

francislavoie:
Reopened as https://github.com/caddyserver/caddy/pull/5385

PR #3807, opened 2020-10-18
Implement default header values in Caddyfile

### LINKED ISSUE #3804: feature request: add header only if not already set
I'm in a case where I'd like the header directive to set the header only if it wasn't already set by an upstream server. Think it as a default value for headers. This would be useful for e.g. setting `Cache-Control` to `no-store` by default, except if the upstream server defines a specific `Cache-Control` value. I don't think it is possible yet (at least with the Caddyfile adapter).

The syntax could be inspired from the syntax used to append/delete headers:

```caddyfile
#      ↓
header ?Cache-Control "no-store" {
  defer
}
```

or alternatively using an explicit parameter (but that wouldn't make much sense when used in combination with +/- IMHO):

```caddyfile
header Cache-Control "no-store" {
  if_not_set
  defer
}
```

Would you accept such contribution? Or is it out of scope?

gilbsgilbs:
closes #3804

associated docs PR: https://github.com/caddyserver/website/pull/98

Turned out to be fairly trivial to implement, so I made it. Feel free to close and just pick the relevant tests if there's already a smarter or better approach to this or if you just don't want to introduce such syntax.

I'm very impressed about how clean caddy's codebase is. Huge shout-out to maintainers and contributors :clap: .

### DIFF
--- a/modules/caddyhttp/headers/caddyfile.go
+++ b/modules/caddyhttp/headers/caddyfile.go
@@ -25,8 +25,8 @@ import (
 )
 
 func init() {
-	httpcaddyfile.RegisterHandlerDirective("header", parseCaddyfile)
-	httpcaddyfile.RegisterHandlerDirective("request_header", parseReqHdrCaddyfile)
+	httpcaddyfile.RegisterDirective("header", parseCaddyfile)
+	httpcaddyfile.RegisterDirective("request_header", parseReqHdrCaddyfile)
 }
 
 // errBadOperation is returned when trying to use a header operation in an inappropriate context.
@@ -47,21 +47,25 @@ var errBadOperation = errors.New("this header operation cannot be used here")
 // in the first line, but not both in the same directive. Header operations
 // are deferred to write-time if any headers are being deleted or if the
 // 'defer' subdirective is used.
-func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error) {
-	hdr := new(Handler)
-
-	makeResponseOps := func() {
-		if hdr.Response == nil {
-			hdr.Response = &RespHeaderOps{
-				HeaderOps: new(HeaderOps),
-			}
-		}
+func parseCaddyfile(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error) {
+	matcherSet, err := httpcaddyfile.ParseMatcher(h)
+	if err != nil {
+		return nil, err
 	}
+
+	deferred := false
+	handlers := []Handler{}
 
 	for h.Next() {
 		// first see if headers are in the initial line
 		var hasArgs bool
 		if h.NextArg() {
+			hdr := Handler{
+				Response: &RespHeaderOps{
+					HeaderOps: &HeaderOps{},
+				},
+			}
+
 			hasArgs = true
 			field := h.Val()
 			var value, replacement string
@@ -71,26 +75,30 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 			if h.NextArg() {
 				replacement = h.Val()
 			}
-			makeResponseOps()
-			if err := applyCaddyfileHeaderOp(
+			err := applyHeaderOp(
 				hdr.Response.HeaderOps,
+				hdr.Response,
 				field,
 				value,
 				replacement,
-				hdr.Response,
-			); err != nil {
+			)
+			if err != nil {
 				return nil, h.Err(err.Error())
 			}
-			if len(hdr.Response.HeaderOps.Delete) > 0 {
-				hdr.Response.Deferred = true
-			}
+			handlers = append(handlers, hdr)
 		}
 
 		// if not, they should be in a block
 		for h.NextBlock(0) {
+			hdr := Handler{
+				Response: &RespHeaderOps{
+					HeaderOps: &HeaderOps{},
+				},
+			}
+
 			field := h.Val()
 			if field == "defer" {
-				hdr.Response.Deferred = true
+				deferred = true
 				continue
 			}
 			if hasArgs {
@@ -103,23 +111,30 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 			if h.NextArg() {
 				replacement = h.Val()
 			}
-			makeResponseOps()
-			if err := applyCaddyfileHeaderOp(
+			err := applyHeaderOp(
 				hdr.Response.HeaderOps,
+				hdr.Response,
 				field,
 				value,
 				replacement,
-				hdr.Response,
-			); err != nil {
+			)
+			if err != nil {
 				return nil, h.Err(err.Error())
 			}
-			if len(hdr.Response.HeaderOps.Delete) > 0 {
-				hdr.Response.Deferred = true
-			}
+			handlers = append(handlers, hdr)
 		}
 	}
 
-	return hdr, nil
+	configValues := make([]httpcaddyfile.ConfigValue, 0, len(handlers))
+	for _, hdr := range handlers {
+		if deferred {
+			hdr.Response.Deferred = true
+		}
+
+		configValues = append(configValues, h.NewRoute(matcherSet, hdr)...)
+	}
+
+	return configValues, nil
 }
 
 // parseReqHdrCaddyfile sets up the handler for request headers
@@ -127,13 +142,23 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 //
 //     request_header [<matcher>] [[+|-|?]<field> [<value|regexp>] [<replacement>]]
 //
-func parseReqHdrCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error) {
-	hdr := new(Handler)
+func parseReqHdrCaddyfile(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error) {
+	matcherSet, err := httpcaddyfile.ParseMatcher(h)
+	if err != nil {
+		return nil, err
+	}
+
+	configValues := []httpcaddyfile.ConfigValue{}
+
 	for h.Next() {
 		if !h.NextArg() {
 			return nil, h.ArgErr()
 		}
 		field := h.Val()
+
+		hdr := Handler{
+			Request: &HeaderOps{},
+		}
 
 		// sometimes it is habitual for users to suffix a field name with a colon,
 		// as if they were writing a curl command or something; see
@@ -158,11 +183,13 @@ func parseReqHdrCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, 
 			return nil, h.Err(err.Error())
 		}
 
+		configValues = append(configValues, h.NewRoute(matcherSet, hdr)...)
+
 		if h.NextArg() {
 			return nil, h.ArgErr()
 		}
 	}
-	return hdr, nil
+	return configValues, nil
 }
 
 // CaddyfileHeaderOp applies a new header operation according to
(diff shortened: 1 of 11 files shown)


mholt:
Sorry I'm just getting around to this now. Thank you for your time to contribute this!

One question, why won't this work?

```json
{
	"handler": "headers",
	"response": {
		"require": {
			"headers": {
				"Cache-Control": [""]
			}
		},
		"set": {
			"Cache-Control": ["no-store"]
		}
	}
}
```

The `headers` handler already has a mechanism for conditionally applying operations to response headers, so why add more complexity here?

gilbsgilbs:
Thanks. I wasn't aware of this syntax and it indeed seems to fill (and generalize) the use-case. But can the Caddyfile adapter translate to this? If not, either we can use the new syntax from this PR or I can change this PR to support a "require" directive in Caddyfile syntax somehow. WDYT?

mholt:
If that'll work for you, then probably all we need is Caddyfile support for the response matcher. Here are the JSON docs for it, note the `response/require` path: https://caddyserver.com/docs/modules/http.handlers.headers

What do you propose for a Caddyfile syntax?

gilbsgilbs:
The simplest proposal I can think of is this:

```caddyfile
header Cache-Control "no-store" {
  # 1 argument => status_code
  require <status_code>

  # 2 arguments => header 
  require <header_field> <header_value>
}
```

Is that ok?

mholt:
Yeah, mostly -- except that it precludes setting any header with the name "require" -- which might be a problem... hm.

What about using your `?` syntax instead? It's still conditional, but use the existing facilities for making it conditional, rather than introducing new `SetDefault` etc.

gilbsgilbs:
> except that it precludes setting any header with the name "require"

I missed that you can set/remove/replace fields directly within the header directive's scope :man_facepalming: . There's also this issue today with the `defer` keyword right? Or not because defer doesn't take any argument? Feels wonky :yum: .

>  What about using your ? syntax instead?

Are you thinking of something like this?

```caddyfile
header ?Cache-Control "no-store"

# Which would be equivalent to the more verbose:
header Cache-Control "no-store" {
  ?Cache-Control ""
}
```

I like that it is a generalization of the syntax suggested in this PR which looks very smooth and easy to use. The only drawback I see is that it doesn't support status codes which I don't need yet but still is a bit sad. And I don't see an elegant way to extend over such syntax (as `?500` would just feel weird, right?).

Anyways, if this sounds fine to you, I'll update my PR.

mholt:
> There's also this issue today with the defer keyword right? Or not because defer doesn't take any argument? Feels wonky :yum: .

Ha, yeah it is a bit wonky, but you're right it's because it doesn't take any arguments, so it can't be confused as a header field name.

> Edit: Or maybe you just meant to only keep the syntax from this PR, but use require without additional fuss. That's also an option.

That's what I had in mind, yes! Sorry for not being clearer about that. It is true that this only works as a "default" setter instead of a more general solution, but I bet default is among the most common operations. I also on't know a better Caddyfile syntax for the more general solution for now.

And yeah, we can use the existing "require" facilities to make it happen pretty easily I think.

gilbsgilbs:
Sorry I didn't catch that earlier, but I don't think it is possible to use `require` for default values as we want. `require` invokes [matchHeaders](https://github.com/caddyserver/caddy/blob/99b8f44486b766f220a33906d84ac05af942f260/modules/caddyhttp/matchers.go#L440) under the hood which only works if the header is present in the input somehow. Consequently, doing `"require": {"headers": {"Foo": [""]}}` just tries to match empty header values, but skips missing header values. In our case the header is just absent from the response, so it fails to match it.

We could add a special case to treat empty values as missing but that sounds like a breaking change and it would add a blind spot around empty header values. An alternative would be to support negated requirements, which would allow things something like this:

```json
"require": {
  "not": {
    "headers": {
      "Cache-Control": []
    }
  }
}
```

gilbsgilbs:
Implemented my suggestion. I made it in a new commit so that we can easily rollback if you think this is overkill.

mholt:
@gilbsgilbs Ah, you're right -- I'm sorry, I was wrong in my example.

I have good news though.

Upon a closer look at my own code, the docs say:

> 	// If a list is null, the header must not exist. If the list is empty,
>	// the field must simply exist, regardless of its value.

So we see that these three values for a header field are different:

- `nil` (or `null` in JSON)
- `[]`
- `[""]`

I had proposed the last one, which I was wrong about :man_facepalming: - that one actually matches a header with an empty value as you correctly noted. The second one matches a header that exists, but with any value; the first one is what we want (`nil` / `null`).

Then I realized that doesn't have a test case :man_facepalming: so I wrote one (two actually):

```go
		{
			match:  MatchHeader{"Must-Not-Exist": nil},
			input:  http.Header{},
			expect: true,
		},
		{
			match:  MatchHeader{"Must-Not-Exist": nil},
			input:  http.Header{"Must-Not-Exist": []string{"it does, oh well"}},
			expect: false,
		},
```

and the first test fails. :man_facepalming: 

So, that's 3x :man_facepalming: on my part. I just pushed a fix in 4fc570711e170da2bd1164966d2c47f9dc3d3d6d. If you build from that, then simply setting the field to `nil` should behave the way we want, with a lot less work/complexity. :)

Let me know how that goes!

modules/caddyhttp/headers/caddyfile.go
```diff
@@ -149,13 +163,36 @@ func parseReqHdrCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler,
...
 		if ops.Add == nil {
 			ops.Add = make(http.Header)
 		}
 		ops.Add.Set(field[1:], value)
 	} else if strings.HasPrefix(field, "-") {
 		ops.Delete = append(ops.Delete, field[1:])
+	} else if strings.HasPrefix(field, "?") {
+		if respHeaderOps == nil {
+			// "?" prefix only makes sense for response headers.
+			return
```
francislavoie, line 180: Hmm, I find it odd that it'll silently ignore here. Would be nicer if it could bubble up an error instead, I think. Maybe too big a refactor though. :man_shrugging:

gilbsgilbs: It is true that it can be misleading. I made the necessary refactoring in 95aa9ab59429789a5c44b080998c1d0340ed8544 . Trying to use `?` in the wrong context would produce an error similar to this:

```
parsing caddyfile tokens for 'reverse_proxy': Caddyfile:8 - Error during parsing: ?Foo: this header operation cannot be used here
```

modules/caddyhttp/headers/caddyfile.go
```diff
@@ -27,11 +30,16 @@ func init() {
 	httpcaddyfile.RegisterHandlerDirective("request_header", parseReqHdrCaddyfile)
 }
 
+// ErrBadOperation is returned when trying to use a header operation in an inappropriate context.
+// For instance, when trying to use "?" operation (set default) with "header_down".
+var ErrBadOperation = errors.New("this header operation cannot be used here")
```
mholt, line 35: Let's not export this; avoid cluttering the public API (exporting is a big commitment).

modules/caddyhttp/headers/caddyfile.go
```diff
@@ -148,14 +174,36 @@ func parseReqHdrCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler,
...
+		if respHeaderOps == nil {
+			return fmt.Errorf("%v: %w", field, ErrBadOperation)
+		}
+
+		if respHeaderOps.Require == nil {
+			respHeaderOps.Require = &caddyhttp.ResponseMatcher{
+				Headers: make(http.Header, 1),
+			}
+		}
+		key := textproto.CanonicalMIMEHeaderKey(field[1:])
```
mholt, line 199: Why is this needed? (The response matcher will do this already, see `getHeaderFieldVals()` in matchers.go.)

modules/caddyhttp/headers/caddyfile.go
```diff
@@ -148,14 +174,36 @@ func parseReqHdrCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler,
...
+		// Header must not be set.
```
mholt, line 200: Comment seems obvious; can be removed.

mholt (changes requested):
Thanks for the progress on this! Just a few points of feedback. Getting pretty close.

gilbsgilbs:
I just realized that this:

```caddyfile
	header {
		Grace "Hopper"
		?Tim "Berners-Lee"
	}
```

Translates to:

```json
{
	"handler": "headers",
	"response": {
		"add": {
			"Tim": [
				"Berners-Lee"
			]
		},
		"require": {
			"headers": {
				"Tim": null
			}
		},
		"set": {
			"Grace": [
				"Hopper"
			]
		}
	}
}
```

Does it make any sense? I'm not sure :thinking: .

mholt:
Ah... good catch. Darn. The proper way to handle this is to return multiple handlers, one which conditionally sets the response headers, and the other which sets the others. `RegisterHandlerDirective` doesn't support returning multiple handlers, but you could just use `RegisterDirective` and then return multiple handlers (as ConfigValues, just see how RegisterHandlerDirective works) as one way of doing this (other directives already do this, see, for example, php_fastcgi).

Another option is to return just one handler like you are now, but it's a Subroute, with multiple header handlers inside it. That's probably less idiomatic but you could do it.

What do you think?

gilbsgilbs:
Thanks for the hints, that helped. This PR now puts each header op from a block in its own separate handler, fixing any risk of clash. Although I'm not sure this broad fix is perfectly safe perfomance-wise. Let me know what you think.

mholt (changes requested):
Thank you for the tests by the way!

I have given this a thorough review, I appreciate your work on it. I've decided it will be faster and easier if I just push my suggestions to your branch directly. The biggest thing I noted is we don't need to create a separate handler for every header operation, we only need the conditional ones to be in a single separate handler, so at most two handlers per directive block. (Multiple handlers does not really adversely affect performance. Only perf difference is minor, and that is a deeper call stack.) I've refactored for a cleaner JSON output and am pushing that change now. Let me know what you think, make sure it still works for you and then we can merge it.

mholt (dismissed):

modules/caddyhttp/headers/caddyfile.go
```diff
@@ -76,62 +74,68 @@ func parseCaddyfile(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error)
...
 				continue
 			}
 			if hasArgs {
-				return nil, h.Err("cannot specify headers in both arguments and block")
+				return nil, h.Err("cannot specify headers in both arguments and block") // because it would be weird
 			}
+
+			// sometimes it is habitual for users to suffix a field name with a colon,
+			// as if they were writing a curl command or something; see
+			// https://caddy.community/t/v2-reverse-proxy-please-add-cors-example-to-the-docs/7349
```
francislavoie, line 104: Just to be more specific, cause it's a long thread

```suggestion
			// https://caddy.community/t/v2-reverse-proxy-please-add-cors-example-to-the-docs/7349/19
```

gilbsgilbs:
Tested and I confirm it works well. Thanks for the work @mholt and @francislavoie . :ok_hand:

mholt (approved):

mholt:
Great, thanks so much for your effort!

cypressious:
Is this also available in `request_header`?

mholt:
> Is this also available in `request_header`?

Should be, yes.

cypressious:
Unforunately, I'm getting `run: adapting config using caddyfile: /etc/caddy/Caddyfile:20: unrecognized directive: request_header?`
when trying

```
request_header? X-Forwarded-Port {port}
```

My workaround is

```
@no_forwarded-port header !X-Forwarded-Port *
request_header @no_forwarded-port X-Forwarded-Port {port}
```

mholt:
@cypressious `request_header?` is not a directive, but `request_header` is. Make sure to correct the typo in your config.

PR #3816, opened 2020-10-22
reverseproxy: Avoid "operation was canceled" errors

danlsgiga:
Related to #3748

Do not pass the request context to the LookupSRV call (no context is used for the dial call anyways). This avoids caddy from logging the "operation was cancelled" coming from the [net package](https://golang.org/src/net/net.go#L411) constantly when the client closes the connection and the request context is canceled, which also was misleading caddy to log these as `502` in the access logs

I've also added and additional error handling for when the upstream connection timesout and we return the `http.StatusGatewayTimeout` code (504) appropriately, instead of 502.

### DIFF
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -144,7 +144,7 @@ func (u *Upstream) fillDialInfo(r *http.Request) (DialInfo, error) {
 	if u.LookupSRV != "" {
 		// perform DNS lookup for SRV records and choose one
 		srvName := repl.ReplaceAll(u.LookupSRV, "")
-		_, records, err := net.DefaultResolver.LookupSRV(r.Context(), "", "", srvName)
+		_, records, err := net.LookupSRV("", "", srvName)
 		if err != nil {
 			return DialInfo{}, err
 		}
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -144,7 +144,7 @@ func (u *Upstream) fillDialInfo(r *http.Request) (DialInfo, error) {
 	if u.LookupSRV != "" {
 		// perform DNS lookup for SRV records and choose one
 		srvName := repl.ReplaceAll(u.LookupSRV, "")
-		_, records, err := net.LookupSRV("", "", srvName)
+		_, records, err := net.DefaultResolver.LookupSRV(r.Context(), "", "", srvName)
 		if err != nil {
 			return DialInfo{}, err
 		}
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -421,6 +421,12 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyht
 			return nil
 		}
 
+		// detect connection timeout to the upstream and respond with 504
+		switch e, ok := proxyErr.(net.Error); ok {
+		case e.Timeout():
+			return caddyhttp.Error(http.StatusGatewayTimeout, proxyErr)
+		}
+
 		// if the roundtrip was successful, don't retry the request or
 		// ding the health status of the upstream (an error can still
 		// occur after the roundtrip if, for example, a response handler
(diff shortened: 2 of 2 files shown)


modules/caddyhttp/reverseproxy/hosts.go
```diff
@@ -144,7 +144,7 @@ func (u *Upstream) fillDialInfo(r *http.Request) (DialInfo, error) {
 	if u.LookupSRV != "" {
 		// perform DNS lookup for SRV records and choose one
 		srvName := repl.ReplaceAll(u.LookupSRV, "")
-		_, records, err := net.DefaultResolver.LookupSRV(r.Context(), "", "", srvName)
+		_, records, err := net.LookupSRV("", "", srvName)
```
mholt, line 147: Why not use DefaultResolver? That is what `net.LookupSRV` does under the hood, and without that, we lose context support, which is important so we can cancel requests without hanging the server during config reloads or shutdowns.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -421,6 +421,12 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyht
 			return nil
 		}
 
+		// detect connection timeout to the upstream and respond with 504
+		switch e, ok := proxyErr.(net.Error); ok {
+		case e.Timeout():
+			return caddyhttp.Error(http.StatusGatewayTimeout, proxyErr)
```
mholt, line 427: Returning here bypasses all the health check and load balancing features.

mholt (changes requested):
Thanks for the PR! I think I see where you are going with this, but we need to be careful about how we implement the change. Would love to know your thoughts on these comments...

modules/caddyhttp/reverseproxy/reverseproxy.go
danlsgiga: Oh, right, because of the retries. Indeed... I'm gonna remove it!

modules/caddyhttp/reverseproxy/hosts.go
danlsgiga: The basic reason is to avoid the error log being "flooded" with canceled requests when the client closes the connection before the response is sent. I was getting tons of "operation was canceled" errors because of the context being passed to the LookupSRV being canceled. Using `net.LookupSRV` will indeed use the `DefaultResolver` under the hood but with a new `context.Background()`.

If there's a better way to handle this, I'm all for it! This was just my dirty short way of "fixing" it on my side! :)

danlsgiga: Ah, and to avoid getting our access logs reporting 502's due to those

mholt: Not using the context will leak resources; so we can't accept this. But if you want to suppress just "context canceled" errors, I would check for [that error value directly](https://golang.org/pkg/context/#pkg-variables), like `if err == context.Canceled`.  Does that work?

danlsgiga: yeah, I tried that but the `net` pkg does not return with `context.Canceled` errors... its basically, "operation was cancelled" - https://golang.org/src/net/net.go#L411

Do you have any other suggestions to fix this one? I agree with you it may leak resources, but with my limited understanding of Go that was the fastest and simpler way I found.

mholt: Ugh, I see, that's annoying.

Worst case scenario, maybe we do text string comparison.

What about converting to net.Error, would this work? https://golang.org/pkg/net/#Error

That source you linked me to has a TODO. Why don't they just return context.Canceled???? They're literally just changing it to some unexported value which is no help.

danlsgiga: I don't have any preferred method for solving this... I'm happy with whatever way you may find to not log (except in DEBUG) the context cancelled errors for when client connections are closed. Also, not considering these as `502` is required IMO. We've lost a lot of time chasing our own tails on these 502's :) Thinking we were losing packets between caddy and upstreams when the issue was these misleading client closed connections generating 502's

mholt:
@danlsgiga I have pushed a commit that handles context cancellation within the reverse proxy as a client error, 499, which is _not_ a standard error code but presumably one that a lot of people are already using. Can you please try it out and see if it works for your needs?

danlsgiga:
This works for me Matt! Thanks for working on this!

mholt (approved):
Great, thanks for trying it out! We'll go with this for now.

Issue #3820, opened 2020-10-26
Logging filter for ip_mask does not work with ipv4-adresses


christoph-kluge:
Related: https://caddy.community/t/use-ip-mask-in-caddyfile-v2/10173/11

It seems that ipv4 filtering does not work correctly. 

PR will followup. This ticket is just servers as a reference for the PR.

mholt:
Thank you for the report and investigation!

PR #3821, opened 2020-10-26
Fixes logging filter for ip_mask with ipv4 adresses

christoph-kluge:
Addresses: #3820 
Related: https://caddy.community/t/use-ip-mask-in-caddyfile-v2/10173/11

IPv4 filter was not working correctly because `to16()` does always return a `true` value if it's an valid ip address. Independently if it's v4 or v6.

On the otherhand `to4()` does return nil when it's not an v4. So in this case it would correctly identify an IPv6 address and use the proper mask.

Logging before:

```
{
	"msg": "handled request",
	"request": {
		"remote_addr": "[::]:61671",
	},
}
```

Logging after

```
{
	"msg": "handled request",
	"request": {
		"remote_addr": "127.0.0.0:63058",
	},
}
```

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -141,7 +141,7 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 		return in
 	}
 	mask := m.v4Mask
-	if ipAddr.To16() != nil {
+	if ipAddr.To4() == nil {
 		mask = m.v6Mask
 	}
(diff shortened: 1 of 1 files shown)


christoph-kluge:
@francislavoie I am limitted to 3 comments per reply in the above mentioned discussion (default for new users).

francislavoie (approved):

mholt (approved):
Ahh, very interesting. Thanks for the fix!

PR #3824, opened 2020-10-28
httpcaddyfile: Revise automation policy generation

mholt:
This should fix a frustrating edge case where wildcard subjects are
used, which potentially get shadowed by more specific versions of
themselves; see the new tests for an example. This change is motivated
by an actual customer requirement.

Although all the tests pass, this logic is incredibly complex and
nuanced, and I'm worried it is not correct. But it took me about 4 days
to get this far on a solution. I did my best.

This is actually a better solution to https://github.com/caddyserver/caddy/commit/b6686a54d8b21bedbf042caa4a6c09d78d345fc7 I think.

### DIFF
--- a/caddyconfig/httpcaddyfile/tlsapp.go
+++ b/caddyconfig/httpcaddyfile/tlsapp.go
@@ -76,30 +76,35 @@ func (st ServerType) buildTLSApp(
 		}
 	}
 
+	// a catch-all automation policy is used as a "default" for all subjects that
+	// don't have custom configuration explicitly associated with them; this
+	// is only to add if the global settings or defaults are non-empty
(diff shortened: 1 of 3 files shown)


francislavoie (approved):
"Hide whitespace changes" helped here, some of the blocks where somewhat misaligned in the diff because the tabulation was changed.

I don't think I have a 100% grasp on this, but I think it all makes sense.

:+1:

mholt:
> I don't think I have a 100% grasp on this,

Me neither tbh :grimacing: 

> but I think it all makes sense.

Oh good! Because it doesn't for me :sweat_smile:

mholt:
Gonna merge this early and see if anyone using HEAD will report regressions as soon as possible.

PR #3829, opened 2020-10-29
reverseproxy: Make conns per host configurable, fix godocs

francislavoie:
- Makes `max_conns_per_host` configurable in Caddyfile
- Adds new `max_idle_conns_per_host` option which is passed through to `http.Transport`, plus Caddyfile support
- Fix godoc for `buffer_requests` (was in the wrong place) and add other missing options from http transport

This stuff came up after I read https://caddy.community/t/caddy-container-500-mb-ram-within-10h-1-gb-at-some-point/10070/13 which made me dig into this

I'll make a website PR as well after merged (unless you want me to do it asap?)


mholt (approved):
Great, thanks for this change!

If you submit a PR to the website before I get around to fixing it, go right ahead. Thank you again!

PR #3832, opened 2020-10-31
caddyhttp: Add Caddyfile support for merging header matchers

francislavoie:
Adds support for adding multiple headers values to match to the same field, from the Caddyfile.

This was something that was already possible in JSON, but in the Caddyfile, it was only keeping the last defined field-value pair, and not merging it into the list.

Example:
```
	@matcher7 {
		header Foo bar
		header Foo foobar
		header Bar foo
	}
	respond @matcher7 "header matcher merging values of the same field"
```

Now gives this JSON matcher (would only have `"Foo": ["foobar"]` before):
```json
"header": {
	"Bar": [
		"foo"
	],
	"Foo": [
		"bar",
		"foobar"
	]
}
```


mholt (approved):
LGTM, thanks! This is a great fix.

Issue #3669, opened 2020-08-20
Proposal: Caddyfile directives for grouping per-server options


francislavoie:
We have plenty of per-server options available in the JSON that don't yet have Caddyfile equivalents. See here: https://caddyserver.com/docs/json/apps/http/

After discussing with @whitestrake, I think we can group the ones that are missing into 3 new directives. Proposed syntax below:

```
timeouts {
	read_body   <duration>
	read_header <duration>
	write       <duration>
	idle        <duration>
}

limits {
	max_body_size   <size>
	max_header_size <size>
}

protocol {
	allow_h2c
	experimental_http3
	strict_sni_host
}
```

Notice that `max_body_size` is not one of the server options in JSON. This would map to the [`request_body`](https://caddyserver.com/docs/modules/http.handlers.request_body) handler which is currently only available in JSON, with the concession that from the Caddyfile, no matchers could be used (i.e. matching all requests), and it would be added at the top of the routes.

Every option other than `max_body_size` would just be directly passed through to the JSON as-is.

Since `experimental_http3` is already a global option, the global option would of course take precedence in enabling it for all servers, and this new per-server option wouldn't offer an opt-out. Either on globally, or on per-server.

mholt:
We did something like this in Caddy 1 but it was not a clean mapping, and can never be, as long as more than 1 site is served on a listener. The reason being that all these settings apply to the HTTP _servers_, not individual _HTTP requests_ except for, as you note, max_body_size, which is a middleware-enforced thing -- and I don't think it's a good idea to combine those because the semantics are so different.

This will need more consideration before any implementation is done. (And is another reason why I don't really love the Caddyfile, esp. for advanced use cases.)

AtjonTV:
I would love to see max_body_size find its way into the Caddyfile. I dislike the idea of writing a lot of JSON for my Webserver just as I dislike writing a lot of XML (... Apache ...). This is why I event started to use Caddy as a primary Server.

As of right now I have 302 formatted lines of Caddyfile that transform into 1997 lines of formatted JSON when fead into Caddy.

There is NO WAY I am going to touch that 2k mess of JSON by hand :roll_eyes: 

I would rather go in and hardcode that setting for the one domain I need it for, if there is no Caddyfile directive xD

francislavoie:
New proposal. Global option instead, with a syntax like this:

```
{
	servers [<listener_matcher>] {
		listener_wrappers {
			<TBD...>
		}
		timeouts {
[...]

PR #3836, opened 2020-11-01
httpcaddyfile: Configure servers via global options

francislavoie:
Implementation of the proposal here https://github.com/caddyserver/caddy/issues/3669#issuecomment-706820214

Quick example, here we're setting server options only for the server that has the listener `:443`. This will be the default for any site that doesn't explicitly define a port (or `http://`), (as long as `https_port` is not configured). Two sites, first is `:443`, second is `:8080`, so the options only apply to the first because of the `servers :443` matching only the first. We also have a `servers` with no matcher, and that one applies to the `:8080` site as a fallback.

```
{
	servers :443 {
		listener_wrappers {
			tls
		}
		timeouts {
			read_body 30s
			read_header 30s
			write 30s
			idle 30s
		}
		max_header_size 100MB
		protocol {
			allow_h2c
			experimental_http3
			strict_sni_host
		}
	}
	servers {
		max_header_size 20MB
	}
}

foo.com {
}

bar.com:8080 {
}
```

👇 

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":443"
          ],
          "listener_wrappers": [
            {
              "wrapper": "tls"
            }
          ],
          "read_timeout": 30000000000,
          "read_header_timeout": 30000000000,
          "write_timeout": 30000000000,
          "idle_timeout": 30000000000,
          "max_header_bytes": 100000000,
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "foo.com"
                  ]
                }
              ],
              "terminal": true
            }
          ],
          "strict_sni_host": true,
          "experimental_http3": true,
          "allow_h2c": true
        },
        "srv1": {
          "listen": [
            ":8080"
          ],
          "max_header_bytes": 20000000,
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "bar.com"
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    }
  }
}
```

~~Still WIP, need to implement listener wrappers support, and need to add a bunch of tests. But the core of~~ it is done!


mholt:
Wow, this is pretty awesome.

I'm starting a review of it; my initial thought is, how about we change "servers" (a tad overloaded in our application) to "sockets" since what we're really describing is configuring options that apply to all the sites/"servers" that share the socket defined by the given address?

caddyconfig/httpcaddyfile/options.go
```diff
@@ -43,6 +43,7 @@ func init() {
 	RegisterGlobalOption("local_certs", parseOptTrue)
 	RegisterGlobalOption("key_type", parseOptSingleString)
 	RegisterGlobalOption("auto_https", parseOptAutoHTTPS)
+	RegisterGlobalOption("servers", parseServerOptions)
```
mholt, line 46: A reminder that we should discuss whether to rename this to "sockets" or "listeners".

caddyconfig/httpcaddyfile/serveroptions.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
+	"github.com/dustin/go-humanize"
+)
+
+// serverOptions collects server config overrides parsed from Caddyfile global options
+type serverOptions struct {
+	// If set, will only apply to servers that contain a listener
+	// address that matches. If empty, will apply to all servers.
+	// Only the most specific one will apply.
+	ListenerMatcher string
```
mholt, line 34: This must be a listener _address_ right? Should we rename this to `ListenerAddress`?

How do you define specificity?

caddyconfig/httpcaddyfile/serveroptions.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// These will all map 1:1 to the Server struct
```
mholt, line 36:
```suggestion
	// These will all map 1:1 to the caddyhttp.Server struct
```

mholt (changes requested):
Nice, this isn't as crazy as I thought it would be. Just a couple questions, regarding naming, and how to match specific sockets/listeners.

caddyconfig/httpcaddyfile/options.go
francislavoie: I don't like the alternatives as much. When talking about request matchers, we don't do like `match @foo`, we do `handle @foo`.

For the same reason, I don't think it makes sense to say `listeners [<listener_matcher>]`. What we actually configure is the server options (as in `http.Server`), and matching them by their listener values.

caddyconfig/httpcaddyfile/serveroptions.go
francislavoie: I guess `ListenerAddressMatcher`? Specifically it's the string that'll be used to match against the list of the server's listener addresses. I think it makes sense to have `Matcher` in the name for that reason.

mholt: Crucially, this field matches one and only 1 listener, which must be an exact match, right? So `:8080` won't match all listeners on port 8080?

francislavoie: No, `:8080` won't match `127.0.0.1:8080` (for now). We could set up `*:8080` or something if you feel it's worth it, but I think that's uncommon in the Caddyfile, because `bind` is not a commonly used feature I'd say.

francislavoie: But yeah - specificity is length of the matcher, sorts them by listener matcher (empty, i.e. "any", is always last)

mholt: Exact matching is fine, but it's hardly "matching" at that point -- it's literally just the address of the listener. Matching implies some smart or cascading logic IMO.

If specificity is length, it's hard to say that `:8080` is really less specific than `:80` - the two are precisely equally specific. I think we should lose the notions of specificity and matching here, and just say that it's the address of the server listener which we're applying this to.

francislavoie: The reason I mentioned specificity is because `""` is decidedly less specific than `:80` (and will apply to "everything else").

But yeah I guess I can reword it.

damienallen:
Really looking forward to "read_timeout" being accessible through the Caddyfile. To be honest the format is _way_ more accessible than JSON and this is the only parameter that is missing for me at the moment.

Can't wait for the release!

caddyconfig/httpcaddyfile/serveroptions.go
```diff
@@ -0,0 +1,247 @@
...
+	return nil
+}
+
+// applyServerOptions sets the server options on the appropriate servers
+func applyServerOptions(servers map[string]*caddyhttp.Server, options map[string]interface{}) error {
+	// If experimental HTTP/3 is enabled, enable it on each server.
+	// We already know there won't be a conflict with serverOptions because
+	// we validated earlier that "experimental_http3" cannot be set at the same
+	// time as "servers"
+	if enableH3, ok := options["experimental_http3"].(bool); ok && enableH3 {
```
mholt, line 202: Should we just deprecate that global option? (It is experimental, after all.) (To clarify, I mean to keep the one in the servers block, and have just one way to enable it.)

mholt (changes requested):
Thanks, this is my last review pass, I think!

What's up with all the whitespace changes in the test files?

caddyconfig/httpcaddyfile/addresses.go
```diff
@@ -163,6 +164,13 @@ func (st *ServerType) consolidateAddrMappings(addrToServerBlocks map[string][]se
 
 		sbaddrs = append(sbaddrs, a)
 	}
+
+	// sort them by their first address (we know there will always be at least one)
+	// to avoid problems with non-deterministic ordering (makes tests flaky)
+	sort.Slice(sbaddrs, func(i, j int) bool {
+		return sbaddrs[i].addresses[0] < sbaddrs[j].addresses[0]
+	})
+
```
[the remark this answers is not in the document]
francislavoie: Hmm, how do you tell it to use `address[0]`? I probably misunderstand this API

caddyconfig/httpcaddyfile/serveroptions.go
francislavoie: Yeah, I figured we could mark it deprecated later but if you're happy with doing it now that works for me

caddyconfig/httpcaddyfile/addresses.go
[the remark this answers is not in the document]
francislavoie: `sbaddrs` is not a `[]string`, it's a `[]sbAddrAssociation` so I don't know how we could use `sort.Strings`?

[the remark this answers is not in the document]
mholt: Oh, you're sorting `sbaddrs`, derp. I guess that makes sense. :(

caddyconfig/httpcaddyfile/serveroptions.go
mholt: I usually prefer sooner rather than later before more people start using it.

caddyconfig/httpcaddyfile/serveroptions.go
```diff
@@ -0,0 +1,251 @@
...
+func validateServerOptions(options map[string]interface{}) error {
+	// if we don't have "servers", then no errors to report
+	_, ok := options["servers"].([]serverOptions)
+	if !ok {
+		return nil
+	}
+
+	if _, ok := options["experimental_http3"]; ok {
+		return fmt.Errorf("cannot set 'experimental_http3' while also using 'servers'; please use the 'servers > protocol > experimental_http3' option instead")
+	}
```
mholt, line 190: This still relevant then?

mholt (approved):
Cool, looking good.

As per our discussion in Slack, the changes to the test files fix (remove) Windows line endings.

PR #3839, opened 2020-11-01
caddyhttp: Merge query matchers in Caddyfile

francislavoie:
I noticed this while revising the matcher docs, `query` wasn't adapting as advertised. It was only letting you have one value per key, and if you specified more tokens, it would fail because `d.Args(&query)` would end up hitting every 2nd token because it's in a `d.Next()` loop. Switches to `Add` from `Set` which will append the value (even if it's empty) instead of overwriting every time.

Also, turns out that `Add` on headers will work even if there's nothing there yet (because it's initialized already as an empty array), so we can remove the condition I introduced in #3832 (effectively we just switched from `Set` to `Add`)


mholt (approved):
Very nice, thanks for the fix and tests!

PR #3844, opened 2020-11-02
fileserver: Improve and clarify file hiding logic

mholt:
Improvements to file hiding logic, but mostly to documentation to make it clear that hide paths are _file system paths_, NOT request paths, and that any relative paths are relative to the current working directory, NOT the site root (because site roots can be dynamic).

Hopefully these tests pass on Windows :man_shrugging:

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -46,7 +46,20 @@ type FileServer struct {
 	Root string `json:"root,omitempty"`
 
 	// A list of files or folders to hide; the file server will pretend as if
-	// they don't exist. Accepts globular patterns like "*.hidden" or "/foo/*/bar".
+	// they don't exist. Accepts globular patterns like "*.ext" or "/foo/*/bar"
+	// as well as placeholders. Because site roots can be dynamic, this list
+	// uses file system paths, not request paths. To clarify, the base of
+	// relative paths is the current working directory, NOT the site root.
+	//
+	// Entries without a path separator (`/` or `\` depending on OS) will match
+	// any file or directory of that name regardless of its path. To hide only a
+	// specific file with a name that may not be unique, always use a path
+	// separator. For example, to hide all files or folder trees named "hidden",
+	// put "hidden" in the list. To hide only ./hidden, put "./hidden" in the list.
+	//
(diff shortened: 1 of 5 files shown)


modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -330,40 +360,50 @@ func sanitizedPathJoin(root, reqPath string) string {
...
-// fileHidden returns true if filename is hidden
-// according to the hide list.
+// fileHidden returns true if filename is hidden according to the hide list.
+// filename must be a relative or absolute file system path, not a request
+// URI path. It is expected that all the paths in the hide list are absolute
+// paths or are singular filenames (without a path separator).
 func fileHidden(filename string, hide []string) bool {
-	sep := string(filepath.Separator)
+	if len(hide) == 0 {
+		return false
```
francislavoie, line 375: Does it really save anything to do this condition? It'll just skip past the loop and return false if it's empty anyways.

mholt: Saves a possible system call on the next line :point_down:

francislavoie: I guess.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -330,40 +360,50 @@ func sanitizedPathJoin(root, reqPath string) string {
...
 			// name; thus we have to compare against each component
 			// of the filename, e.g. hiding "bar" would hide "/bar"
 			// as well as "/foo/bar/baz" but not "/barstool".
 			if len(components) == 0 {
-				components = strings.Split(filename, sep)
+				components = strings.Split(filename, separator)
 			}
 			for _, c := range components {
-				if c == h {
+				if hidden, _ := filepath.Match(h, c); hidden {
```
francislavoie, line 397: So does this mean something like `/foo/*/bar` will work now?

francislavoie (approved):
Obviously fix windows first but LGTM

modules/caddyhttp/fileserver/staticfiles.go
mholt: Always did. This makes sure that `*.txt` will block all text files. (see added test cases)

PR #3849, opened 2020-11-05
Add new acme_server database options

### LINKED ISSUE #3847: acme_server directive errors on arm
I've just validated that a very simple Caddyfile adding an acme server fails when executed on an ARM Linux device (Raspberry Pi 2 and 4), but appears fine on my Darwin device. This is reproduced with Docker and the binaries for version 2.1.1 and 2.2.1.

Example Caddyfile
```
{
    local_certs
    debug
}

acme.localhost {
    acme_server
}
```

Validating with `caddy validate` on ARM yields
```
2020/11/04 18:53:31.910	INFO	using adjacent Caddyfile
2020/11/04 18:53:31.914	INFO	tls.cache.maintenance	started background certificate maintenance	{"cache": "0x246f810"}
2020/11/04 18:53:31.917	INFO	http	server is listening only on the HTTPS port but has no TLS connection policies; adding one to enable TLS	{"server_name": "srv0", "https_port": 443}
2020/11/04 18:53:31.917	INFO	http	enabling automatic HTTP->HTTPS redirects	{"server_name": "srv0"}
2020/11/04 18:53:31.917	INFO	tls	setting internal issuer for automation policy that has only internal subjects but no issuer configured	{"subjects": ["acme.localhost"]}
2020/11/04 18:53:31.946	INFO	tls.cache.maintenance	stopped background certificate maintenance	{"cache": "0x246f810"}
validate: loading http app module: provision http: server srv0: setting up route handlers: route 0: loading handler modules: position 0: loading module 'subroute': provision http.handlers.subroute: setting up subroutes: route 0: loading handler modules: position 0: loading module 'acme_server': provision http.handlers.acme_server: initializing certificate authority: Error opening database of Type badger with source /home/username/.local/share/caddy/acme_server/db: error opening Badger database: Unable to mmap RDWR log file: invalid argument
```

On my macOS it yields
```
2020/11/04 18:54:54.824	INFO	using adjacent Caddyfile
2020/11/04 18:54:54.838	INFO	tls.cache.maintenance	started background certificate maintenance	{"cache": "0xc000473030"}
2020/11/04 18:54:54.855	INFO	http	server is listening only on the HTTPS port but has no TLS connection policies; adding one to enable TLS	{"server_name": "srv0", "https_port": 443}
2020/11/04 18:54:54.855	INFO	http	enabling automatic HTTP->HTTPS redirects	{"server_name": "srv0"}
2020/11/04 18:54:54.855	INFO	tls	setting internal issuer for automation policy that has only internal subjects but no issuer configured	{"subjects": ["acme.localhost"]}
2020/11/04 18:54:54.924	INFO	tls.cache.maintenance	stopped background certificate maintenance	{"cache": "0xc000473030"}
Valid configuration
```

ViViDboarder:
As described in #3847, the ACME server fails when run on ARM devices.
This is due to smallstep using `MemoryMap` log storage for it's Badger
database.

This patch allows one to either disable `MemoryMap` and switch to `FileIO`

    acme_server {
        no_memory_map
    }

or to explicitly switch to BadgerV2

    acme_server {
        use_badgerv2
    }

If you have no existing database state, you can simply use BadgerV2, as
it will eventually become the default.

If instead you have an existing state, this will give you a manifest
error. In that case you will need to disable `MemoryMap`.

Fixes #3847

### DIFF
--- a/modules/caddypki/acmeserver/acmeserver.go
+++ b/modules/caddypki/acmeserver/acmeserver.go
@@ -65,12 +65,12 @@ type Handler struct {
 	PathPrefix string `json:"path_prefix,omitempty"`
 
 	// Whether to store using FileIO rather than MemoryMap
-	// Deafult: false
-	NoMemoryMap bool `json:"no_memory_map,omitempty`
+	// Default: false
+	NoMemoryMap bool `json:"no_memory_map,omitempty"`
 
 	// Whether to use BadgerV2 storage
-	// Deafult: false
-	UseBadgerV2 bool `json:"use_badgerv2,omitempty`
+	// Default: false
+	UseBadgerV2 bool `json:"use_badger_v2,omitempty"`
 
 	acmeEndpoints http.Handler
 }
--- a/modules/caddypki/acmeserver/acmeserver.go
+++ b/modules/caddypki/acmeserver/acmeserver.go
@@ -64,6 +64,14 @@ type Handler struct {
 	// on this property long-term; check release notes.
 	PathPrefix string `json:"path_prefix,omitempty"`
 
+	// Whether to store using FileIO rather than MemoryMap
+	// Deafult: false
+	NoMemoryMap bool `json:"no_memory_map,omitempty`
+
+	// Whether to use BadgerV2 storage
+	// Deafult: false
+	UseBadgerV2 bool `json:"use_badgerv2,omitempty`
+
 	acmeEndpoints http.Handler
 }
 
@@ -107,6 +115,20 @@ func (ash *Handler) Provision(ctx caddy.Context) error {
 		return fmt.Errorf("making folder for ACME server database: %v", err)
 	}
 
+	// Use FileIO rather than MemoryMap if provided
+	// See https://github.com/caddyserver/caddy/issues/3847
+	badgerFileLoadingMode := "MemoryMap"
+	if ash.NoMemoryMap {
+		badgerFileLoadingMode = "FileIO"
+	}
+
+	// Use BadgerV2 rather than badger if provided
+	// See https://github.com/caddyserver/caddy/issues/3847
+	dbType := "badger"
+	if ash.UseBadgerV2 {
+		dbType = "badgerV2"
+	}
+
 	authorityConfig := caddypki.AuthorityConfig{
 		AuthConfig: &authority.AuthConfig{
 			Provisioners: provisioner.List{
@@ -122,8 +144,9 @@ func (ash *Handler) Provision(ctx caddy.Context) error {
 			},
 		},
 		DB: &db.Config{
-			Type:       "badger",
-			DataSource: dbFolder,
+			Type:                  dbType,
+			DataSource:            dbFolder,
+			BadgerFileLoadingMode: badgerFileLoadingMode,
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -64,6 +64,14 @@ type Handler struct {
 	// on this property long-term; check release notes.
 	PathPrefix string `json:"path_prefix,omitempty"`
 
+	// Whether to store using FileIO rather than MemoryMap
+	// Deafult: false
+	NoMemoryMap bool `json:"no_memory_map,omitempty`
+
+	// Whether to use BadgerV2 storage
+	// Deafult: false
+	UseBadgerV2 bool `json:"use_badgerv2,omitempty`
```
francislavoie, line 73: Please see the golangci-lint errors in the diff https://github.com/caddyserver/caddy/pull/3849/files#diff-47200afac47ea8df3c0a5e4b0c7b00872d3d40287b0327ae1cccfc90161e4606R73

francislavoie: Also I think `use_badger_v2` is nicer.

ViViDboarder: Yea. I think it's nicer too. I used `bagerv2` because that's what the value is called, but I guess that's not really important and definitely makes it harder to read.

ViViDboarder:
Do you think this is possible to get into the next minor / patch release? I'm not sure how you all manage the Caddy release cycle.

francislavoie:
Possibly. @mholt will need to review when he has the time :slightly_smiling_face:

mholt:
Thanks for the PR.

I'd rather fix this without exposing any config to the user. I think the storage implementation is a deep detail the user shouldn't have to worry about, and I really don't want to add surface area here.

Instead, what if we just change the string to `"badgerV2"` -- especially considering that right now the DB doesn't hold anything particularly important or which can't be re-created instantaneously.

ViViDboarder:
That makes sense to me, however doing that will cause an error on Caddy start for anyone with an existing database.

How do you think that this should be handled? Automatically clearing it? I'm not sure if the metadata can be important in some cases.

Another possible solve could be to get rid of the options and then default to `FileIO` rather than `MemoryMap` and then wait for smallstep/certificates to to update their default badger implementation to v2.

This would get rid of the config options and provide something that will work out of the box for all users.

ViViDboarder:
Alternately... maybe there is some way we can detect the existence of a database and explicitly use `v2` if empty and then use the correct badger version if identified. Not sure if this is a thing we can easily do though. The first option would be preferable to me.

mholt:
Hmm, yeah, if it already exists, I suppose we'd have to wipe it if there's no way to upgrade it?

ViViDboarder:
No way to upgrade, only remove.

By moving the db path to a new path (either `db2` if deciding to default to badger2 or `boltdb` if deciding to switch to bbolt), it will be as simple as deleting anything in the old db path.

Actually, if changing to boltdb (which does seem to work on ARM), we can keep the path as `db` since bolt writes to a file rather than a directory. This will allow stating the path and removing it if it's a directory and keeping it if it's a file.

ViViDboarder:
Since the changes to swap out the engines were super simple, I just submitted a PR for each so you can take your pick.

francislavoie:
I figure we can close this one in favour of https://github.com/caddyserver/caddy/pull/3868. Thanks @ViViDboarder!

PR #3859, opened 2020-11-08
Add Caddyfile support for request_body

pic:
```
  request_body {
    max_size 10000000
  }
```

See: https://caddy.community/t/file-upload-size-limits/8419


pic:
@mholt 
It's a copy / paste / adapt effort. I'm a complete beginner in go. Any feedback is appreciated, thx.

modules/caddyhttp/requestbody/caddyfile.go
```diff
@@ -0,0 +1,48 @@
...
+func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error) {
+	rb := new(RequestBody)
+
+	for h.Next() {
+		// if not, they should be in a block
+		for h.NextBlock(0) {
+			field := h.Val()
+			if field == "max_size" {
+				if h.NextArg() {
+					i, err := strconv.Atoi(h.Val())
```
francislavoie, line 36: Please use `humanize.ParseBytes` to support syntax like `500MB`

See my other PR here for an example:

https://github.com/caddyserver/caddy/pull/3836/files?diff=split&w=1#diff-f05363f26f6d0155a82b76ca06bbe4bd52a152d6b147a8f5fa804cb9bacfc8e8R134-R143

modules/caddyhttp/requestbody/caddyfile.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package requestbody
+
+import (
```
francislavoie, line 17: As noted by the linter, please run `gofmt -s -w <this file>` please!

modules/caddyhttp/requestbody/caddyfile.go
```diff
@@ -0,0 +1,48 @@
...
+
+func init() {
+	httpcaddyfile.RegisterHandlerDirective("request_body", parseCaddyfile)
+}
+
+func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error) {
+	rb := new(RequestBody)
+
+	for h.Next() {
+		// if not, they should be in a block
```
francislavoie, line 31: This comment doesn't really make sense because the only thing being parsed is `max_size`.

You should reject unknown fields (you can use a `switch` statement with a `default` that returns an error). You should also reject any additional args on the same line (if we decide to only configure this via the block option, see below)

I'm not convinced that the only way to configure this should be via a block, because it only has one option. For example, the directive might be `max_request_body_size <size>` instead.

But on the other hand, aligning it with the JSON config isn't a bad idea either. It's just that we might never have any other options than just `max_size`, so it feels kinda strange to require a block.

I'm of two minds about it.

Anyways, the actual syntax as written would be:

```
request_body [<matcher>] {
	max_size <size>
}
```

francislavoie:
Also please don't forget to sign the CLA. If you already did but it's not recognizing it, it might be that you committed have a different email address than you have attached to your GitHub account. Please add that email to your GitHub account if so.

modules/caddyhttp/requestbody/caddyfile.go
pic: thanks for the example, I've adapted it now

modules/caddyhttp/requestbody/caddyfile.go
pic: yes, exactly. Should I change to `max_request_body_size` or do we keep it this way?

modules/caddyhttp/requestbody/caddyfile.go
```diff
@@ -0,0 +1,52 @@
...
+				var sizeStr string
+				if !h.AllArgs(&sizeStr) {
+					return nil, h.ArgErr()
+				}
+				size, err := humanize.ParseBytes(sizeStr)
+				if err != nil {
+					return nil, h.Errf("parsing max_size: %v", err)
+				}
+				rb.MaxSize = int64(size)
+				continue
```
francislavoie, line 44: `continue` is not necessary because you're already in a loop and `default` is not run if any other case is matched

```suggestion

```

modules/caddyhttp/requestbody/caddyfile.go
francislavoie: I'm not sure. I'll let @mholt chime in. I'm leaning towards keeping it as-is but I wanted to bring up the point of discussion.

francislavoie:
Oh - also please add an adapt test, see here: https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt just make a new file that has a minimal Caddyfile, then a separator with 10 `-` chars, then the JSON output of `caddy adapt --pretty`.

pic:
@francislavoie :+1: , done.
Let's see about the Caddyfile syntax discussion.

mholt (approved):
Thank you! This actually looks pretty good. And it's a simple improvement, which we like. Happy to accept this.

Thanks for your contribution! If you have a moment, would appreciate some help updating the docs for this new directive for our website too: https://github.com/caddyserver/website

modules/caddyhttp/requestbody/caddyfile.go
mholt: I think this is a less-commonly-used feature, so we're probably fine with the more verbose syntax for now. It also lets us add more options like replacements (etc.) in the future.

PR #3861, opened 2020-11-12
basicauth: Nitpicks and small improvements

acuteaura:
contains 3 small improvements

* roll two `if` statements into one, since `err` is guaranteed be `nil` in the second case
* unlock cache mutex after reading from the map, as this happens by-value and reduces code complexity with double unlocks
* switch cache `sync.Mutex` to `sync.RWMutex` for better concurrency on cache fast track


mholt (approved):
Thanks! Looks like this checks out. I can't find anything wrong with this from scanning the changes.

PR #3867, opened 2020-11-14
acme_server: switch to badgerV2 storage

ViViDboarder:
There have been some issues with the badgerv1 storage engine being used by the embedded acme_server. This will replace the storage engine with badgerV2

Fixes #3847
Closes #3849
Closes #3868

Issue #3865, opened 2020-11-13
acme_server: error on reload 


Mohammed90:
I started with this config:

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":80"
          ],
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "localhost"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "acme_server"
                }
              ]
            }
          ]
        }
      }
    }
  }
}
```

To which I only added logging:

```json
{
  "logging": {
    "logs": {
      "log0": {
        "writer": {
          "output": "stderr"
        },
        "encoder": {
          "format": "json"
        }
      }
    }
  },
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "logs": {
            "logger_names": {
              "localhost": "log0"
            }
          },
          "listen": [
            ":80"
          ],
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "localhost"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "acme_server"
                }
              ]
            }
          ]
        }
      }
    }
  }
}
```

Ran `caddy reload --config caddy.json` and got the error:

```
ERROR	admin.api	request error	{"error": "loading config: loading new config: loading http app module: provision http: server srv0: setting up route handlers: route 0: loading handler modules: position 0: loading module 'acme_server': provision http.handlers.acme_server: initializing certificate authority: Error opening database of Type badger with source /Users/mohammed/Library/Application Support/Caddy/acme_server/db: error opening Badger database: Cannot acquire directory lock on \"/Users/mohammed/Library/Application Support/Caddy/acme_server/db\".  Another process is using this Badger database.: resource temporarily unavailable", "status_code": 400}

reload: sending configuration to instance: caddy responded with error: HTTP 400: {"error":"loading config: loading new config: loading http app module: provision http: server srv0: setting up route handlers: route 0: loading handler modules: position 0: loading module 'acme_server': provision http.handlers.acme_server: initializing certificate authority: Error opening database of Type badger with source /Users/mohammed/Library/Application Support/Caddy/acme_server/db: error opening Badger database: Cannot acquire directory lock on \"/Users/mohammed/Library/Application Support/Caddy/acme_server/db\".  Another process is using this Badger database.: resource temporarily unavailable"}

2020/11/13 21:22:08.725	INFO	admin	stopped previous server
```

[...]

PR #3868, opened 2020-11-14
acme_server: switch to bbolt storage

### LINKED ISSUE #3847: acme_server directive errors on arm

ViViDboarder:
There have been some issues with the badger storage engine being used by the embedded acme_server. This will replace the storage engine with bbolt.

Fixes #3847
Closes #3849
Closes #3867

### DIFF
--- a/modules/caddypki/acmeserver/acmeserver.go
+++ b/modules/caddypki/acmeserver/acmeserver.go
@@ -99,13 +99,18 @@ func (ash *Handler) Provision(ctx caddy.Context) error {
 		return fmt.Errorf("no certificate authority configured with id: %s", ash.CA)
 	}
 
-	dbFolder := filepath.Join(caddy.AppDataDir(), "acme_server", "db")
+	dbFolder := filepath.Join(caddy.AppDataDir(), "acme_server")
+	dbPath := filepath.Join(dbFolder, "db2")
 
 	// TODO: See https://github.com/smallstep/nosql/issues/7
 	err = os.MkdirAll(dbFolder, 0755)
 	if err != nil {
 		return fmt.Errorf("making folder for ACME server database: %v", err)
 	}
+
+	// Clear old db paths out
+	legacyDbPath := filepath.Join(dbFolder, "db")
+	_ = os.RemoveAll(legacyDbPath)
 
 	authorityConfig := caddypki.AuthorityConfig{
 		AuthConfig: &authority.AuthConfig{
@@ -122,8 +127,8 @@ func (ash *Handler) Provision(ctx caddy.Context) error {
 			},
 		},
 		DB: &db.Config{
-			Type:       "badger",
-			DataSource: dbFolder,
+			Type:       "bbolt",
+			DataSource: dbPath,
 		},
 	}
 
--- a/modules/caddypki/acmeserver/acmeserver.go
+++ b/modules/caddypki/acmeserver/acmeserver.go
@@ -32,6 +32,7 @@ import (
 	"github.com/smallstep/certificates/authority/provisioner"
 	"github.com/smallstep/certificates/db"
 	"github.com/smallstep/nosql"
+	"go.uber.org/zap"
 )
 
 func init() {
@@ -77,6 +78,7 @@ func (Handler) CaddyModule() caddy.ModuleInfo {
 
 // Provision sets up the ACME server handler.
 func (ash *Handler) Provision(ctx caddy.Context) error {
+	logger := ctx.Logger(ash)
 	// set some defaults
 	if ash.CA == "" {
 		ash.CA = caddypki.DefaultCAID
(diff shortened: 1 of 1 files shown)


mholt:
Thanks! Letsee... I think this is better than using badger, but I also don't think it solves this problem: https://github.com/caddyserver/caddy/issues/3865, since bolt also acquires an exclusive lock on a file. During reloads, Caddy will temporarily run two configurations at the same time to ensure the new config is valid and successful before closing the old one. So for a short time, two locks are required, or ideally, we find a way to transfer those the one lock to the new config, or share it, somehow.

In order for this fix to be complete, I think we'll have to make sure we reuse the same lock. Fortunately, Caddy has a facility for sharing such global state, called a [`UsagePool`](https://pkg.go.dev/github.com/caddyserver/caddy/v2#UsagePool). Basically, it's a reference-counting map. So you'd basically encode the CA configuration down to a string (could even be something as dumb as `fmt.Sprintf("%#v", authorityConfig)` -- because this value is purely internal, and can be as cryptic as a hash if we want) and then put it in the UsagePool. Then, on reload, the UsagePool would see that the value already exists and return that, rather than try to make a new one. Here's an example of how it's used: https://github.com/caddyserver/caddy/blob/13781e67ab1b2553598d0dd1a7153ce3cdbd4879/modules/caddyhttp/reverseproxy/reverseproxy.go#L221-L225

What do you think?

ViViDboarder:
Hmm. Interesting. I'll have to investigate a bit to see how this will work. Since this does resolve #3847, would it be ok to merge this without fixing #3865 and then fix that in a different branch? That would also limit the number of changes per branch.

mholt (changes requested):
Fair enough, that sounds good to me.

However, I'd like to either keep the db path the same, and only delete it if it's badger; an easy way to detect this if it has a file called MANIFEST, I think. (Can you confirm that the bbolt implementation does not create a file called MANIFEST?)

ViViDboarder:
So `bbolt` creates a file rather than a directory, so it'll actually be pretty easy to detect and delete.

ViViDboarder:
Finished. I added a log line to let the user know something is being deleted. I could also remove it though.

For research towards the other issue, reloading with a changed config causes:

```
reload: sending configuration to instance: caddy responded with error: HTTP 400: {"error":"loading config: loading new config: loading http app module: provision http: server srv0: setting up route handlers: route 0: loading handler modules: position 0: loading module 'subroute': provision http.handlers.subroute: setting up subroutes: route 0: loading handler modules: position 0: loading module 'acme_server': provision http.handlers.acme_server: initializing certificate authority: Error opening database of Type bbolt with source /Users/vividboarder/.local/share/caddy/acme_server/db: timeout"}
```

So it will indeed doing what you described above to fix #3865. I'll try to work on that next.

mholt (approved):
Great -- so my understanding is this is ready to merge, and your subsequent PR depends on this one. Should we merge this one in and then the next one (once it has finished the review process), or should we just roll them into one PR? What do you prefer?

ViViDboarder:
Let's see if I can get the other branch ready to go some time this week. If not, then it'd be nice to merge this so ARM folks can use the ACME server.

mholt:
Thank you for your help with this!!

PR #3875, opened 2020-11-18
reverseproxy: fix random hangs on http/2-requests with Server Pushes

dtelyukh:
see https://github.com/golang/go/issues/42534


mholt (approved):
Very interesting! Thanks for diving in with others and helping to make a fix happen. We really appreciate it. :100:

PR #3892, opened 2020-11-30
headers: Fix Caddyfile parsing with request matcher

### LINKED ISSUE #3891: Looks like header directive is broken
On current master-branch such configuration fails:

```
https://caddytest.localhost {
    tls internal
    @staticAssets {
        path /img/*
    }
    header @staticAssets {
        Cache-Control "public, max-age=3600, stale-while-revalidate=86400"
    }
}
```

The error message: `run: adapting config using caddyfile: parsing caddyfile tokens for 'header': Caddyfile:7 - Error during parsing: cannot specify headers in both arguments and block`

On v2.2.1 it is OK.

francislavoie:
Closes #3891

Fixes regression from b0d5c2c8ae076393e7a3ad59ce875027f4c29304

The problem is that `RegisterHandlerDirective` also calls `h.Next()` to consume the directive name from the dispenser before calling `h.MatcherToken()`. When this was changed to use `RegisterDirective` instead and use `h.ExtractMatcherSet()`, this was forgotten, so it broke request matchers because it would first see `headers` instead of `@foo` when trying to extract the matcher.

I confirmed it still rejects the config if non-matcher args are given at the same time as block ones.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/header.txt
+++ b/caddytest/integration/caddyfile_adapt/header.txt
@@ -9,6 +9,10 @@
 		?Tim "Berners-Lee"
 		defer
 	}
+	@images path /images/*
+	header @images {
+		Cache-Control "public, max-age=3600, stale-while-revalidate=86400"
+	}
 }
 ----------
 {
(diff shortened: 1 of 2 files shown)


mholt (approved):
Aha, thank you! I wondered if something was a little off...

Q800 (commented):
@Qbm

PR #3895, opened 2020-12-01
caddyauth: fix unbuffered channel passed to signal.Notify

cuonglm:
The docs at os/signal.Notify warn about this signal delivery loss bug at
https://golang.org/pkg/os/signal/#Notify, which says:

    Package signal will not block sending to c: the caller must ensure
    that c has sufficient buffer space to keep up with the expected signal
    rate. For a channel used for notification of just one signal value,
    a buffer of size 1 is sufficient.

Caught by a static analysis tool from Orijtech, Inc. called "sigchanyzer"

### DIFF
--- a/modules/caddyhttp/caddyauth/command.go
+++ b/modules/caddyhttp/caddyauth/command.go
@@ -70,7 +70,7 @@ func cmdHashPassword(fs caddycmd.Flags) (int, error) {
 		if terminal.IsTerminal(fd) {
 			// ensure the terminal state is restored on SIGINT
 			state, _ := terminal.GetState(fd)
-			c := make(chan os.Signal)
+			c := make(chan os.Signal, 1)
 			signal.Notify(c, os.Interrupt)
 			go func() {
 				<-c


cuonglm:
Related Go issue: https://github.com/golang/go/issues/9399

francislavoie:
What issue does this actually solve here? This is just a one-off command used for hashing passwords, and it's just meant to capture `^C` to not break terminal state.

cuonglm:
> What issue does this actually solve here? This is just a one-off command used for hashing passwords, and it's just meant to capture `^C` to not break terminal state. I've not seen issues with that usecase. 
> 
> That said, this looks fine, just trying to understand.

I invite you to read https://github.com/golang/go/issues/9399

Without using buffered channel, there can be a risk that the signal is lost, because it's sent before the channel is ready. Sure it's hard to get in to trouble with this use case, but I think it's better to always use the right (and safe) way that the language offers.

mholt (approved):
I agree this is probably an insignificant change, but, we do use a buffer size of 1 in sigtrap.go already, so I guess we can be consistent here. Thanks for the PR!

Issue #3897, opened 2020-12-02
caddy validate: add ability to read the config from stdin


masipcat:
The idea is to be able to validate configurations as follows:

```sh
cat Caddyfile | caddy validate --adatper caddyfile --config -
```

This might sound unnecessary, but it makes sense when you're validating the configuration on a CI system, where Caddy runs inside a container (like docker). Otherwise, you need to mount the file in the container and it's less practical.

EDIT: I opened a PR #3898 in case this is interesting

PR #3898, opened 2020-12-02
cmd: add ability to read config from stdin

masipcat:
Close https://github.com/caddyserver/caddy/issues/3897

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -123,7 +123,11 @@ func loadConfig(configFile, adapterName string) ([]byte, string, error) {
 	var cfgAdapter caddyconfig.Adapter
 	var err error
 	if configFile != "" {
-		config, err = ioutil.ReadFile(configFile)
+		if configFile == "-" {
+			config, err = ioutil.ReadAll(os.Stdin)
+		} else {
+			config, err = ioutil.ReadFile(configFile)
+		}
 		if err != nil {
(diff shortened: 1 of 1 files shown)


francislavoie (approved):
Cool! I don't see why not! :+1:

mholt:
Thanks, it's a nice enhancement and elegant enough, but this affects all functions that call `loadConfig()` including `run` and `reload`. Have you tested how this works with `caddy start`, which utilizes stdin to run `caddy run`?

masipcat:
> Thanks, it's a nice enhancement and elegant enough, but this affects all functions that call `loadConfig()` including `run` and `reload`. Have you tested how this works with `caddy start`, which utilizes stdin to run `caddy run`?

Nope, tomorrow I'll test and report the results!

masipcat:
It doesn't work :( The load config is reading this bytes: https://github.com/caddyserver/caddy/blob/master/cmd/commandfuncs.go#L88-L101

mholt:
Thanks for checking! That's good to know.

It means that `caddy start --config -` won't work, but `caddy run --config -` will. I think we'll have to be OK with that, and mention it in the documentation.

We could probably _make_ it work, but it'll require some hoop-jumping.

mholt (approved):
If you get a chance would you mind submitting a patch to [the docs](https://github.com/caddyserver/website) mentioning this feature? With the caveat that it won't work for `caddy start`.

Issue #3881, opened 2020-11-23
QUIC reports "Network blackhole detected" with large downloads


klzgrad:
I use Caddy 2 and Chrome 87 on localhost to download a large file using QUIC version h3-29. After the first several megabytes the connection stalls and Caddy reports "Network blackhole detected."

This error seems to be generated by Chrome's QUIC stack. In Wireshark I can see the server was sending many packets before stalling for more than 30 milliseconds. The version of quic-go is v0.19.2. Caddy 2 is v2.2.1.

I have no idea how to further debug this, but am ready to test more.

marten-seemann:
Interesting. Can you run Caddy with `QUIC_GO_LOG_LEVEL=debug` and post the log file here?

klzgrad:
Thanks. After reporting this I immediately discovered quic-go actually contains an example server.

Further testing shows this issue can be reproduced with quic-go's example server alone, therefore I will move this issue to quic-go.

klzgrad:
I discovered this can be reproduced quic-go's example server alone. See https://github.com/lucas-clemente/quic-go/issues/2901.

PR #3901, opened 2020-12-04
update quic-go to v0.19.3

marten-seemann:
This patch release fixes a few bugs, most importantly (for Caddy) it disables QUIC key updates, so that large file transfers won't stall any more when using Chrome 87 (see https://github.com/caddyserver/caddy/issues/3881).


mholt (approved):
Cool, thanks for the fix and the PR!

PR #3908, opened 2020-12-08
feat: notify systemctl caddy is ready

### LINKED ISSUE #3876: Caddy doesn't currently notify systemctl when it is properly loaded
Could this be implemented by calling `sd_notify(0, "READY=1");` when initial config has loaded and is being served/requests can be accepted?

gdhameeja:
Closes #3876

Caddy notifies systemctl after config has successfully loaded

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build linux
+
 package caddycmd
 
 import (
@@ -214,6 +216,10 @@ func cmdRun(fl Flags) (int, error) {
 		return caddy.ExitCodeFailedStartup, fmt.Errorf("loading initial config: %v", err)
 	}
 	caddy.Log().Info("serving initial configuration")
+
+	if err := sdNotify("READY=1"); err != nil {
+		caddy.Log().Error("unable to notify systemctl", zap.Error(err))
+	}
 
 	// if we are to report to another process the successful start
 	// of the server, do so now by echoing back contents of stdin
@@ -698,5 +704,27 @@ func apiRequest(adminAddr, method, uri string, body io.Reader) error {
 		return fmt.Errorf("caddy responded with error: HTTP %d: %s", resp.StatusCode, respBody)
 	}
 
+	return nil
+}
+
+func sdNotify(state string) error {
+	socketAddr := &net.UnixAddr{
+		Name: os.Getenv("NOTIFY_SOCKET"),
+		Net:  "unixgram",
+	}
+
+	if socketAddr.Name == "" {
+		return nil
+	}
+
+	conn, err := net.DialUnix(socketAddr.Net, nil, socketAddr)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+
+	if _, err := io.Copy(conn, strings.NewReader(state)); err != nil {
+		return err
+	}
 	return nil
 }


gdhameeja:
@mholt @francislavoie Do we also want to notify systemctl when caddy is reloading or stopping?

francislavoie:
I guess? I'm not sure how much benefit there would be to do that though. What will systemd actually do if we notify it?

francislavoie:
I think you should use build constraints for this feature. We don't want it in non-linux versions of Caddy: https://golang.org/cmd/go/#hdr-Build_constraints

cmd/commandfuncs.go
```diff
@@ -208,6 +208,10 @@ func cmdRun(fl Flags) (int, error) {
 		}
 	}
 
+	if err := sdNotify("READY=1"); err != nil {
+		caddy.Log().Error("unable to notify systemctl", err)
+	}
```
mholt, line 213: This is before the config is loaded.

mholt (changes requested):
Thanks for the PR!

Please use build tags to constrain this to Linux only.

What are the benefits of notifying systemd? What isn't working currently?

cmd/commandfuncs.go
```diff
@@ -208,6 +208,10 @@ func cmdRun(fl Flags) (int, error) {
 		}
 	}
 
+	if err := sdNotify("READY=1"); err != nil {
+		caddy.Log().Error("unable to notify systemctl", err)
```
mholt, line 212:
```suggestion
		caddy.Log().Error("notifying systemctl", zap.Error(err))
```

francislavoie:
I think sending a reload notification would just change the result of `systemctl status caddy` to show the config is reloading, and sending "ready" again would just tell it the reload is done.

cmd/commandfuncs.go
gdhameeja: Apologies? I didn't understand the comment here. Are you saying we should `sdNotify` before the config is ready? If yes, https://github.com/caddyserver/caddy/issues/3876#issue-747061574 says "when initial config is properly loaded". So I kept it after the config was loaded.

francislavoie: `loadConfig` just parses the config, it doesn't actually "run" it. That happens next, with `caddy.Load(config, true)`. So it should be after that.

mholt: Yep, exactly. Thanks!

cmd/commandfuncs.go
```diff
@@ -700,3 +704,25 @@ func apiRequest(adminAddr, method, uri string, body io.Reader) error {
...
+		return nil
+	}
+
+	conn, err := net.DialUnix(socketAddr.Net, nil, socketAddr)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+
+	if _, err := conn.Write([]byte(state)); err != nil {
```
rumpelsepp, line 724: Shouldn't this be a `io.Copy(conn, strings.NewReader(state))` instead?

mholt: Why's that?

rumpelsepp: `conn.Write()` does not guarantee that it will write all the data. `io.Copy()` does. I use this consistently in go code I write as it avoids bugs which are hard to debug.

mholt: That's a good point.

rumpelsepp: maybe not required here, but consistency counts IMO.

gdhameeja: Updated.

cmd/commandfuncs.go
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build linux
```
Mohammed90, line 15: The build tag, i.e. `// +build linux` , means the code within this file is only readable and available when the target is Linux, which is why the build is failing for all other platform. Look at the CI results on the "Checks" tab above.

francislavoie: @gdhameeja you can see an example of how to do this properly with the `sigtrap.go`, `sigtrap_nonposix.go` and `sigtrap_posix.go` files in the root of the project.

You can't set a build tag on a file that is needed on _all_ platforms. For this feature, you'll need to abstract out the logic that you run to platform-specific files, and have a do-nothing stub for unsupported platforms.

rumpelsepp:
I guess the call to sdnotify() should be here: https://github.com/caddyserver/caddy/blob/1bd04ae35f63c1c433112b7d09f6cbfc40ffea37/cmd/commandfuncs.go#L279

mholt:
Any interest in fixing this up? If not, we can close this.

rumpelsepp:
I would like to have a try finishing this.

gdhameeja:
> Any interest in fixing this up? If not, we can close this.

Will fix it up today.

PR #3909, opened 2020-12-09
httpcaddyfile: support matching headers that do not exist

### LINKED ISSUE #3907: caddyfile: allow null values on header matchers
When using the JSON config, you can now match when a header *does not* exist using a `null` value as of 4fc570711e170da2bd1164966d2c47f9dc3d3d6d. However, this syntax is impossible to use when using the Caddyfile.

My proposed solution is to support a `null` literal value for these matchers, perhaps using the same `-` character as with the map module.

lolPants:
Closes #3907

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -475,6 +475,10 @@ func (m *MatchHeader) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 		}
 
 		if strings.HasPrefix(field, "!") {
+			if len(field) == 1 {
+				return d.Errf("malformed header matcher: must have field name following ! character")
+			}
+
 			field = field[1:]
 			headers := *m
 			headers[field] = nil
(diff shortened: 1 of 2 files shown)


lolPants:
Tests are green, but the code might not be super nice. Any feedback is appreciated 😄 
Additionally, I'm not super happy with the error messages so if someone has any input on that too, that'd be grand.

francislavoie (approved):
LGTM! Thanks!

modules/caddyhttp/matchers.go
```diff
@@ -470,13 +470,28 @@ func (m *MatchHeader) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
-			return d.Errf("malformed header matcher: expected both field and value")
+		if !d.Args(&field) {
+			return d.Errf("malformed header matcher: expected field")
 		}
 
-		// If multiple header matchers with the same header field are defined,
-		// we want to add the existing to the list of headers (will be OR'ed)
-		http.Header(*m).Add(field, val)
+		if strings.HasPrefix(field, "!") {
+			field = strings.TrimPrefix(field, "!")
```
mholt, line 478:
```suggestion
			field = field[1:]
```

This way we don't have to repeat the magic `!` symbol.

mholt (changes requested):
LGTM, just one nit for code maintainability.

Also could we add a check to make sure that there is in fact a field name after the `!` (i.e. `len(field)` > 1)?

mholt (approved):
Cool, thanks for the enhancement!

Issue #3896, opened 2020-12-02
Frequent hangs when using http/2 push


aDorofeev:
My team and I use Caddy as a reverse proxy, and we rely on HTTP/2 Push a lot. We've started with v2.2.1, on certain configurations we experienced random hangs.

We thought that the problem was fixed with  this merge request: https://github.com/caddyserver/caddy/pull/3875 (more details can be found in the net/http issue: https://github.com/golang/go/issues/42534)
Because it didn't occur for past couple weeks, since it was merged. But today we've found another configuration that makes it very easy to reproduce the problem again, on any of the recent Caddy versions.

### Caddy behavior when hangs happen
- Web browser hangs infinitely on some resources during webpage assets download:
![image](https://user-images.githubusercontent.com/1867204/100858277-72c23100-3496-11eb-9c20-c221b2ac73d3.png)
- When it happens, Caddy doesn't output any error messages
- During that hang, if I try to download those "pending" assets with a different browser or console `curl` - it might work for some assets (usually, smaller ones), but for many assets, it hangs infinitely, i.e. when the problem happens for one user - i basically makes the whole website inaccessible for other users too, till that one user hits "stop" button
- When I hit the "stop" button in a browser tab with hanging downloads, Caddy outputs lots of errors to console:
```ERROR   http.handlers.reverse_proxy     aborting with incomplete response  {"error": "http2: stream closed"}```

### How to reproduce
It depends on the proxied website and caddy config, and some random factors, thus it occurs with different frequency on different hardware. The steps are:
1. Start caddy with the configuration provided below
2. Open developer tools network tab, to visually see the hangs; checking the "disable cache" toggle will help reproduce the problem faster, but is not necessary
3. Navigate to `/` page in a browser (i.e., https://terem-pro.localhost)
4. Wait till it fully loads
5. If it didn't hang on step 4 - hit f5, and again, wait till it fully loads; repeat several times if needed

On our test server, it usually hangs after 2-3 reloads. On some devices, it might require 10-15 attempts but still hangs at some point.

### Caddy version
Reproduces on:
- v2.2.1
- v2.3.0-beta.1
- current head of master (4cff36d731390915649261f0e9c088be0eeafcf1), "caddyauth: Use buffered channel passed to signal.Notify"

Built with:
`CADDY_RACE_DETECTOR=1 xcaddy build <revision>`

### Caddy configuration
```
https://terem-pro.localhost {                                                 
    handle {                                                                                                                                                           
[...]

PR #3910, opened 2020-12-09
reverseproxy: fix random hangs on http/2-requests with Server Pushes

dtelyukh:
The solution is to use explicitly recent `x/net/http2` instead of bundled version, which updates only with releases of Go language.
Fix: https://github.com/caddyserver/caddy/issues/3896
Details: https://github.com/golang/go/issues/42534

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -306,6 +306,7 @@ func (app *App) Start() error {
 			Handler:           srv,
 			ErrorLog:          serverLogger,
 		}
+		http2.ConfigureServer(s, nil)
 
 		// enable h2c if configured
 		if srv.AllowH2C {
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -306,7 +306,10 @@ func (app *App) Start() error {
 			Handler:           srv,
 			ErrorLog:          serverLogger,
 		}
-		http2.ConfigureServer(s, nil)
+		err = http2.ConfigureServer(s, nil)
+		if err != nil {
+			return fmt.Errorf("failed to configure http2: %v", err)
+		}
 
 		// enable h2c if configured
 		if srv.AllowH2C {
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/app.go
```diff
@@ -306,6 +306,10 @@ func (app *App) Start() error {
 			Handler:           srv,
 			ErrorLog:          serverLogger,
 		}
+		err = http2.ConfigureServer(s, nil)
```
mholt, line 309: This PR is doing more than just updating the `x/net` module; I'm concerned that this could cause problems as it takes the stability out of our HTTP/2 stack by using the bleeding edge instead of tagged releases.

mholt (changes requested):
Thanks for the PR; I'm fine with updating the net dependency but I don't think we can accept the `http2.ConfigureServer()` call for this.

Go tags releases fairly often -- about once per month -- so the upstream fix should be landing soon enough. I think that will be better for the stability of the project than using HEAD.

modules/caddyhttp/app.go
dtelyukh: @mholt, what about conditional usage? It could be something like that
```
if srv.AllowH2ExternalVersion {
    ...
}
```

mholt: I definitely don't want to expose such an internal implementation detail to the end user to configure. This is something that really should be decided at compile-time. Our project should compile for stability, and if a user wants bleeding edge, they can compile it that way.

This is a permanent complication for a temporary problem...

I think we just need to wait until Go does a new release (just like with every other upstream Go issue), which will probably be soon. (The whole h2 vendoring thing is a weird decision anyway. Either HTTP/2 should have been in the std lib or it should stay outside, not... both... IMO.)

dtelyukh: OK. I agree with you.

Issue #3916, opened 2020-12-13
feat: Add support for placeholders in header matcher


SnirShechter:
The header matcher does not currently support placeholders in its values.

E.g this is not supported: (will match a literal "{http.vars.someVar}" string).
```
"match": [
  {
    "header": {
      "Some-Header": ["{http.vars.someVar}"]
    }
  }
],
```

Following an older PR regarding this issue in the query matcher:
https://github.com/caddyserver/caddy/pull/3521

Implementing this in header keys does not make much sense, so I've ignored it.

PR #3917, opened 2020-12-13
feat: add support for placeholders in headers values

SnirShechter:
Fixes https://github.com/caddyserver/caddy/issues/3916

This PR follows a previous PR: https://github.com/caddyserver/caddy/pull/3521
The only difference is that adding this functionality to the header keys does not really make much sense, so I disregarded that.

PS I'd be happy if this was approved and released ASAP since this feature is blocking our migration to Caddy :)

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -516,11 +516,10 @@ func (m MatchHeader) Match(r *http.Request) bool {
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 	header := http.Header(m)
 	for field, allowedFieldVals := range header {
-		replacedAllowedFieldVals := []string{}
 		for index, allowedFieldVal := range allowedFieldVals {
-			replacedAllowedFieldVals[index] = repl.ReplaceAll(allowedFieldVal, "")
+			allowedFieldVals[index] = repl.ReplaceAll(allowedFieldVal, "")
 		}
-		header[field] = replacedAllowedFieldVals
+		header[field] = allowedFieldVals
 	}
 
 	return matchHeaders(r.Header, header, r.Host)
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/matchers.go
```diff
@@ -513,7 +513,16 @@ func (m *MatchHeader) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 // Match returns true if r matches m.
 func (m MatchHeader) Match(r *http.Request) bool {
-	return matchHeaders(r.Header, http.Header(m), r.Host)
+	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
+	header := http.Header(m)
+	for field, allowedFieldVals := range header {
+		for index, allowedFieldVal := range allowedFieldVals {
+			allowedFieldVals[index] = repl.ReplaceAll(allowedFieldVal, "")
```
mholt, line 520: We should probably only use `ReplaceAllKnown()` here so that it's still possible to match header values that have words surrounded in braces for some reason.

mholt (changes requested):
Thanks, but I'm not sure this is the right approach; I haven't had a chance to test this thoroughly, but I'm worried that this will change the matcher itself by dereferencing the map and array pointers after the first match. This will need to be tested.

Probably a better solution would be to implement this within `matchHeaders()`.

modules/caddyhttp/matchers.go
SnirShechter: Sounds good. I Changed it.

SnirShechter:
> Thanks, but I'm not sure this is the right approach; I haven't had a chance to test this thoroughly, but I'm worried that this will change the matcher itself by dereferencing the map and array pointers after the first match. This will need to be tested.
> 
> Probably a better solution would be to implement this within `matchHeaders()`.

Sounds ok, but this means we need to alter matchHeaders to receive the Context or otherwise not use it.

mholt:
Yep, that's true. As a hint, I'd say the replacements probably need to happen right after `for _, allowedFieldVal := range allowedFieldVals {` and before the `switch {` on the next line?

mholt:
Any interest in finishing this? If not, we can close this.

mholt:
I went ahead and implemented it (much simpler) in https://github.com/caddyserver/caddy/commit/cc63c5805e1587d7e5901deaad3f5bac944a2cf6 -- please try it out!

PR #3918, opened 2020-12-13
rewrite: Use r.URL.RawPath (#3596)

### LINKED ISSUE #3596: v2: Use r.URL.RawPath for rewrite
When using e.g. `uri replace` paths get automatically unescaped. This might be undesirable. Something like this would fix it:

```diff
diff --git a/modules/caddyhttp/rewrite/rewrite.go b/modules/caddyhttp/rewrite/rewrite.go
index d47c388a..4370a77f 100644
--- a/modules/caddyhttp/rewrite/rewrite.go
+++ b/modules/caddyhttp/rewrite/rewrite.go
@@ -303,13 +303,24 @@ func (rep replacer) do(r *http.Request, repl *caddy.Replacer) bool {
        find := repl.ReplaceAll(rep.Find, "")
        replace := repl.ReplaceAll(rep.Replace, "")
 
-       oldPath := r.URL.Path
+       oldPath := r.URL.RawPath
        oldQuery := r.URL.RawQuery
 
-       r.URL.Path = strings.Replace(oldPath, find, replace, lim)
        r.URL.RawQuery = strings.Replace(oldQuery, find, replace, lim)
 
-       return r.URL.Path != oldPath && r.URL.RawQuery != oldQuery
+       if oldPath == "" {
+               oldPath = r.URL.Path
+               r.URL.Path = strings.Replace(oldPath, find, replace, lim)
+
+               return r.URL.Path != oldPath && r.URL.RawQuery != oldQuery
+       }
+
+       r.URL.RawPath = strings.Replace(oldPath, find, replace, lim)
+       if p, err := url.PathUnescape(r.URL.RawPath); err == nil {
+               r.URL.Path = p
+       }
+
+       return r.URL.RawPath != oldPath && r.URL.RawQuery != oldQuery
 }
 
 // Interface guard
```

What do you think? Maybe there's a better way?

go-d:
Prevent information loss, i.e. the encoded form that was sent by the
client, when using URL strip/replace.

Fixes #3596

### DIFF
--- a/modules/caddyhttp/rewrite/rewrite.go
+++ b/modules/caddyhttp/rewrite/rewrite.go
@@ -191,11 +191,21 @@ func (rewr Rewrite) rewrite(r *http.Request, repl *caddy.Replacer, logger *zap.L
 	// strip path prefix or suffix
 	if rewr.StripPathPrefix != "" {
 		prefix := repl.ReplaceAll(rewr.StripPathPrefix, "")
-		r.URL.Path = strings.TrimPrefix(r.URL.Path, prefix)
+		r.URL.RawPath = strings.TrimPrefix(r.URL.RawPath, prefix)
+		if p, err := url.PathUnescape(r.URL.RawPath); err == nil && p != "" {
+			r.URL.Path = p
+		} else {
+			r.URL.Path = strings.TrimPrefix(r.URL.Path, prefix)
+		}
 	}
 	if rewr.StripPathSuffix != "" {
 		suffix := repl.ReplaceAll(rewr.StripPathSuffix, "")
-		r.URL.Path = strings.TrimSuffix(r.URL.Path, suffix)
+		r.URL.RawPath = strings.TrimSuffix(r.URL.RawPath, suffix)
+		if p, err := url.PathUnescape(r.URL.RawPath); err == nil && p != "" {
+			r.URL.Path = p
+		} else {
+			r.URL.Path = strings.TrimSuffix(r.URL.Path, suffix)
+		}
 	}
 
 	// substring replacements in URI
@@ -289,10 +299,10 @@ type replacer struct {
 	Limit int `json:"limit,omitempty"`
 }
 
-// do performs the replacement on r and returns true if any changes were made.
-func (rep replacer) do(r *http.Request, repl *caddy.Replacer) bool {
+// do performs the replacement on r.
+func (rep replacer) do(r *http.Request, repl *caddy.Replacer) {
 	if rep.Find == "" || rep.Replace == "" {
-		return false
+		return
 	}
 
 	lim := rep.Limit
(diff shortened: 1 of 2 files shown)


mholt:
Awesome, thanks!! Will look at getting this in shortly after the 2.3 release.

go-d:
There is an edge case that we still need to consider:
```go
rule:   Rewrite{URISubstring: []replacer{{Find: "findme", Replace: "replaced"}}},
input:  newRequest(t, "GET", "/foo%2findme/bar"),
expect: newRequest(t, "GET", "/foo%2findme/bar"),
```
Using a regular expression in place of `strings.Replace` could be the start of a solution:
```go
re := regexp.MustCompile("(^|^[^%]|[^%][^%])" + regexp.QuoteMeta(find))
r.URL.RawPath = re.ReplaceAllString(r.URL.RawPath, "${1}"+replace)
```
Or maybe a custom version of strings.Replace?

mholt:
Definitely not a regular expression... what is the problem there exactly?

go-d:
The problem is the possibility of unexpected results. A search string having a-f or 0-9 as the first character might affect percent-encoded (%hh) bytes, like so:

1. replace *findme* with *replace*, giving *%2r*... (invalid hex)
2. replace *findme* with *0*, giving *%20* (unintended space)

A replacer that is aware of %-encoding would not alter the URL in both these cases, instead of the currently unpredictable */foo/indme/bar* and */foo%20/bar*, respectively.

mholt:
Oh; I see. So we just need to do the replacement on the decoded form of the URL, rather than the encoded form.

go-d:
I don't think so. How would you then reconstruct the encoded form (RawPath)?

My regexp example works, but could certainly be done without regexp.

mholt:
(Sorry for the delay, will get around to this shortly after the 2.3 release in a few days!)

mholt:
Ok after thinking about this more, and fiddling with it for some time today, I don't think that edge case is a problem.

The rewrite module works on the raw URL - this allows you to do replacements with `%20` versus `/` explicitly. And users can already write their own regexes.

So I think that test case is incorrect. What the rewrite module currently does (with this change) is correct.

Ready for final review?

go-d:
Sure.

Hard to say what is correct though, since the behavior isn't specified.

Agreed that it is likely not much of a problem in practice.

A modified `strings.Replace` that looks behind (in case search string 
starts with a hex digit) to ensure no `%` is there would make the result 
easier to predict in my opinion.

That would require some copying from the standard library. Of code that 
rarely changes. But still.

mholt:
That's essentially trying to parse the _decoded_ form of the path, not the RawPath which we're doing now. If we leave it raw/encoded, we don't lose information. Giving a final review soon. Thanks!

go-d:
Maybe looking for `%` in RawPath and RawQuery (before a string match index) could be considered a form of parsing. But that seems inevitable if one wants to guard against incorrect partial replacements of hex values therein.

At no point is decoding happening there, or URL.Path used for that, though. So no information loss.

For a future issue, perhaps.

mholt (approved):
> Maybe looking for % in RawPath and RawQuery (before a string match index) could be considered a form of parsing. But that seems inevitable if one wants to guard against incorrect partial replacements of hex values therein.

Right; the thing is, if we don't want the encoded characters to be rewritten we should be operating on `Path` like we are now, which is decoded, but the problem is it's harder to go from Path to RawPath. As your PR shows, it's much easier to go from RawPath to Path.

> At no point is decoding happening there, or URL.Path used for that, though. So no information loss.

Oh, yeah, haha -- exactly.

Let's roll with this for a while, I think this is better as-is, but we'll see if we get any complaints. Still, I think even those can be addressed with something like "do the rewrite differently."

Basically, we'll expect the user to configure correct rewrites -- and of course, clients need to continue making correct requests with correct URLs.

go-d:
Thank you for creating Caddy!

(Commit message now says "instead of Path", but usually, RawPath is empty, and Path is used.)

PR #3923, opened 2020-12-14
caddyfile: Introduce basic linting and fmt check

mholt:
This will help encourage people to keep their Caddyfiles tidy.

Doesn't work on imported Caddyfiles, since we don't pass the warnings into the parser (since, in theory, the Caddyfile adapter could never produce warnings because it was designed for Caddy in the first place)... could be refactored sooner or later, or we can just give up and use `log.Print` directly.

I noticed that this editor plugin auto-fmts the Caddyfile on save, which is awesome: https://github.com/matthewpi/vscode-caddyfile-support (but also made testing this inconvenient, haha)

### DIFF
--- a/caddyconfig/caddyfile/adapter.go
+++ b/caddyconfig/caddyfile/adapter.go
@@ -15,6 +15,7 @@
 package caddyfile
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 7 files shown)


mholt:
Cancelled test is due to GitHub's CI maintenance on their mac platform this week.

caddyconfig/caddyfile/adapter.go
```diff
@@ -51,11 +52,17 @@ func (a Adapter) Adapt(body []byte, options map[string]interface{}) ([]byte, []c
 		return nil, warnings, err
 	}
 
-	marshalFunc := json.Marshal
-	if options["pretty"] == "true" {
```
francislavoie, line 55: Wait I'm confused, did you remove `--pretty`?

cmd/commandfuncs.go
```diff
@@ -463,17 +463,22 @@ func cmdAdaptConfig(fl Flags) (int, error) {
...
-	}
-	opts["filename"] = adaptCmdInputFlag
+	opts := map[string]interface{}{"filename": adaptCmdInputFlag}
 
 	adaptedConfig, warnings, err := cfgAdapter.Adapt(input, opts)
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, err
 	}
 
+	if adaptCmdPrettyFlag {
```
francislavoie, line 473: Oh nevermind, it was moved here. lol why was it being done in the adapter :man_facepalming:

mholt:
Since this is a "read-only" change, I'm going to merge this sooner rather than later so people can start testing it and report any more bugs with the formatter.

PR #3932, opened 2020-12-18
ci: reject tags if not signed by Matthew Holt's key

Mohammed90:
This is an attempt at mitigating recurrence of this:
https://twitter.com/caddyserver/status/1338324878441603073

Since this touches our release workflow, I'd really love to test this on a test repo elsewhere before it's actually merged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,22 +59,9 @@ jobs:
         # Import Matt Holt's key
         curl 'https://github.com/mholt.gpg' | gpg --import
 
-        # first we verify the tag, which are only accepted if signed by Matt's key
+        echo "Verifying the tag: ${{ steps.vars.outputs.version_tag }}"
+        # tags are only accepted if signed by Matt's key
         git verify-tag "${{ steps.vars.outputs.version_tag }}" || exit 1
-
-        # Import Github's key, which is used for web merges
-        curl 'https://github.com/web-flow.gpg' | gpg --import
-
-        # now verify all the commits between the new tag and the previous tag
-        current_tag="${{ steps.vars.outputs.version_tag }}"
-        previous_tag=`git describe --tags --no-abbrev "$current_tag"^`
-
-        for commit in $(git rev-list "$previous_tag..$current_tag")
-        do
-          echo "Verifying commit $commit"
-          git verify-commit $commit || exit 1
-          echo "" # add newline for more readable log
-        done
 
     - name: Cache the build cache
       uses: actions/cache@v2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,6 +53,29 @@ jobs:
         echo "::set-output name=tag_patch::${TAG_PATCH}"
         echo "::set-output name=tag_special::${TAG_SPECIAL}"
 
+    - name: Validate commits and tag signatures
+      run: |
+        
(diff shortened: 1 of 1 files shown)


mholt:
Oohhhh Christmas came early I see :heart_eyes: 

How would you like to test this? Should I set up a repo for it? One thing we can do is tag rc.2 before the stable release just to make sure it works.

.github/workflows/release.yml
```diff
@@ -53,6 +53,29 @@ jobs:
...
         echo "::set-output name=tag_special::${TAG_SPECIAL}"
 
+    - name: Validate commits and tag signatures
+      run: |
+        
+        # Import Matt Holt's key
+        curl 'https://github.com/mholt.gpg' | gpg --import
+
+        # first we verify the tag, which are only accepted if signed by Matt's key
+        git verify-tag "${{ steps.vars.outputs.version_tag }}" || exit 1
```
francislavoie, line 63: Can we do like `echo "Bad tag" && exit 1`? Is that a thing? Just so we know where we exited since there's more than one exit point in this step

.github/workflows/release.yml
```diff
@@ -53,6 +53,29 @@ jobs:
...
+
+        # Import Github's key, which is used for web merges
+        curl 'https://github.com/web-flow.gpg' | gpg --import
```
francislavoie, line 66: Hmm. So this means that the release will be rejected if anyone else committed straight to master? I'm not sure if this is an assumption that can be kept forever. Matt is a BDFL basically but I'm not sure he'll always be the only one to commit directly? I guess we could make that a policy though. But it's a decision.

francislavoie: So like, what happens if a PR is merged but not squashed (fast-forward merge), I figure the commits remain intact, not signed by github. So if we ever don't use the squash merge button, it'll make the release fail. 🤔

Mohammed90: Good points. The commits do stay intact and not be signed by Github. We can assume [Linus' stance](http://git.661346.n2.nabble.com/GPG-signing-for-git-commit-td2582986.html) by trusting the entire tree if the tag or the HEAD is signed by a trusted committer. Which should be our preference?

.github/workflows/release.yml
flwrchild42: 🥰 I'm scared as usual lol I'm going to mess this up 😓 but I'll try this is such a great way of learning .. thank you so much you made me cry I'm sorry for all my mishaps 💕

.github/workflows/release.yml
francislavoie: I think just looking at the tag is enough. It means that Matt is 👍 on the contents.

.github/workflows/release.yml
Mohammed90: Done.

.github/workflows/release.yml
Mohammed90: Done.

francislavoie (approved):
I'd be happy to merge this now, WDYT @mholt?

mholt (approved):
Yeah, assuming this works if I change my key... this is probably a good protection against that bug. Thanks! Nice idea.

mholt:
We're going to do another RC release as a test of the updated release processes.

PR #3941, opened 2020-12-28
ci: Add pushing to cloudsmith

francislavoie:
See https://github.com/caddyserver/dist/issues/49

Still needs `CLOUDSMITH_API_KEY` to be added to the repo/org @mholt, I don't remember if you've done it already? Please let me know 😄 leaving as draft until that's confirmed and my TODO is fixed.

Also adds a condition to skip the `caddy_<version>_linux_armv7.deb` because Cloudsmith would complain and Gemfury would have it override the `armv6` one, because apparently they're both `armhf`. But maybe we should skip the `armv6` one instead of `armv7`? I don't know. I assume the `armv6` one would be more compatible 🤷‍♂️ but I don't know anything about ARM arch differences.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,6 +86,9 @@ jobs:
         TAG: ${{ steps.vars.outputs.version_tag }}
 
     # Only publish on non-special tags (e.g. non-beta)
+    # We will continue to push to Gemfury for the forseeable future, although
+    # Cloudsmith is probably better, to not break things for existing users of Gemfury.
+    # See https://gemfury.com/caddy/deb:caddy
     - name: Publish .deb to Gemfury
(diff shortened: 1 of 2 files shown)


mholt:
@francislavoie Aye, the CLOUDSMITH_API_KEY secret was created on Oct 31.

As for ARM version, I suspect v7 because most people are probably running newer hardware, but I'm soliciting feedback, since I'm not sure how that works either: https://twitter.com/mholt6/status/1343984040131477504

francislavoie:
Just waiting for @Mohammed90's review, then I think this is ready to go.

mholt (dismissed):

mholt (approved):
Thanks so much for helping with this, Francis. Excited to try it out.

PR #3944, opened 2020-12-30
metrics: allow disabling OpenMetrics negotiation

### LINKED ISSUE #3940: Metrics clashing name
I'm running [exporter-exporter](https://github.com/QubitProducts/exporter_exporter) to expose caddy metric with TLS and cert authentication and it appears `go_memstats_alloc_bytes`
is in collision with an other metric
```
déc. 27 19:24:31 reverse-proxy exporter_exporter[43467]: time="2020-12-27T19:24:31+01:00" level=error msg="Verification for module 'caddy' failed: Failed to decode metrics from proxied server: text format parsing error in line 343: second HELP line for metric name \"go_memstats_alloc_bytes\""
```

The two metrics
```
# HELP go_memstats_alloc_bytes Number of bytes allocated and still in use.
# TYPE go_memstats_alloc_bytes gauge
go_memstats_alloc_bytes 6.058576e+06
# HELP go_memstats_alloc_bytes_total Total number of bytes allocated, even if freed.
# TYPE go_memstats_alloc_bytes_total counter
go_memstats_alloc_bytes_total 4.4330096e+07
```

If I try to parse them with the prometheus client parser
```
$ python3 -c 'import sys; from prometheus_client.openmetrics import parser; list(parser.text_string_to_metric_families(sys.stdin.buffer.read().decode("utf-8")))'
# HELP go_memstats_alloc_bytes Number of bytes allocated and still in use.
# TYPE go_memstats_alloc_bytes gauge
go_memstats_alloc_bytes 6.058576e+06
# HELP go_memstats_alloc_bytes_total Total number of bytes allocated, even if freed.
# TYPE go_memstats_alloc_bytes_total counter
go_memstats_alloc_bytes_total 4.4330096e+07
```
This is the output I get
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/prometheus_client/openmetrics/parser.py", line 24, in text_string_to_metric_families
    for metric_family in text_fd_to_metric_families(StringIO.StringIO(text)):
  File "/usr/local/lib/python3.8/dist-packages/prometheus_client/openmetrics/parser.py", line 615, in text_fd_to_metric_families
    yield build_metric(name, documentation, typ, unit, samples)
  File "/usr/local/lib/python3.8/dist-packages/prometheus_client/openmetrics/parser.py", line 479, in build_metric
    raise ValueError("Clashing name: " + name + suffix)
ValueError: Clashing name: go_memstats_alloc_bytes_total
```

Openmetrics suffixes
https://github.com/prometheus/client_python/blob/f81891f57bf3d20a9bb87ea62b6c6b4a6c06f803/prometheus_client/openmetrics/parser.py#L466
```
type_suffixes = {
        'counter': ['_total', '_created'],
        'summary': ['', '_count', '_sum', '_created'],
        'histogram': ['_count', '_sum', '_bucket', '_created'],
        'gaugehistogram': ['_gcount', '_gsum', '_bucket'],
        'info': ['_info'],
    }
```

hairyhenderson:
Fixes #3940 by allowing users to disable OpenMetrics negotiation. This shouldn't usually be necessary, but there's a couple of incompatible metric names in the upstream Go collector which break strict parsing. I'll also log a bug in `prometheus/client_golang` and link it here.

Documentation PR is here: https://github.com/caddyserver/website/pull/123


hairyhenderson:
Not sure why `goreleaser-check` is failing - @francislavoie @Mohammed90 any ideas?

Mohammed90:
@hairyhenderson the failure is due to deprecation notices. See https://goreleaser.com/deprecations#nfpmsfiles and https://goreleaser.com/deprecations#nfpmsconfig_files


@francislavoie fixed it in:

https://github.com/caddyserver/caddy/commit/8286f8de51926c749793b2b80ae3654fcaf81fd8

francislavoie:
Sorry about that - I cherrypicked it into https://github.com/caddyserver/caddy/pull/3945 and I'll merge it asap, no use just leaving it in my unrelated PR.

francislavoie:
Okay merged, you could rebase to re-run CI if you like.

hairyhenderson:
Alright - rebased!

mholt (dismissed):
Thanks Dave!

Oof, the golang prometheus client is getting heavier and heavier :(

I don't love having such a deep dependency chain. Something we will have to come back and address later probably.

modules/metrics/metrics.go
```diff
@@ -71,14 +75,23 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 
 // UnmarshalCaddyfile sets up the handler from Caddyfile tokens. Syntax:
 //
-//     metrics <matcher>
+//     metrics <matcher> {
```
francislavoie, line 78: Technically the matcher is optional, so it should be:

```suggestion
//     metrics [<matcher>] {
```

modules/metrics/metrics.go
```diff
@@ -71,14 +75,23 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
...
 		args := d.RemainingArgs()
 		if len(args) > 0 {
 			return d.ArgErr()
 		}
+
+		for d.NextBlock(0) {
+			switch d.Val() {
+			case "disable_openmetrics":
+				m.DisableOpenMetrics = true
+			}
```
francislavoie, line 93: Add a `default:` to reject unsupported subdirectives pls

francislavoie: Also - minor thing but the Go property is called "DisableOpenMetrics", so should the JSON be "disable_open_metrics"? Consistency in case :thinking:

mholt: If `OpenMetrics` is cased that way only because of branding, then `disable_openmetrics` is probably better. But if it is actually two words, "Open Metrics", then `disable_open_metrics` should be preferred. IMO.

francislavoie: Seems like one-word branding: https://openmetrics.io/ so yeah, it's fine I guess. Casing rules can be tricky lol

hairyhenderson: Yeah - the convention seems to be `OpenMetrics` and `openmetrics` depending on context. For example, the MIME type for the text exposition format is `application/openmetrics-text`, not `application/open-metrics-text`.

hairyhenderson:
> Oof, the golang prometheus client is getting heavier and heavier :(
> 
> I don't love having such a deep dependency chain. Something we will have to come back and address later probably.

There's a related conversation in the `prometheus/common` repo (where most of the new dependencies come from I think): https://github.com/prometheus/common/issues/255 - my impression is many of those dependencies aren't really that relevant since they don't end up in the final binary.

mholt:
Thanks for linking that.

> my impression is many of those dependencies aren't really that relevant since they don't end up in the final binary.

And that's good, but it still makes pulling repos really slow and tedious. And easier for things to break then, which makes development hard.

modules/metrics/metrics.go
hairyhenderson: Ah right. I had a bit of a head-scracth moment because the entire block is optional too, but there seems not to be a convention for specifying that?

modules/metrics/metrics.go
hairyhenderson: > Add a `default:` to reject unsupported subdirectives pls

Fixed

modules/metrics/metrics.go
hairyhenderson: Fixed

mholt (approved):
LGTM. Will merge momentarily and cut rc2 to test the updated CI pipeline

PR #3948, opened 2020-12-31
reverseproxy: Caddyfile health check headers, host header support

yaxin-cn:
reverse_proxy:
health check headers can be set through Caddyfile using health_headers directive
health check header host can be set properly

### DIFF
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_health_headers.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_health_headers.txt
@@ -1,0 +1,49 @@
+:8884
+
+reverse_proxy 127.0.0.1:65535 {
+    health_headers {
+        Host example.com
+        X-Header-Keys VbG4NZwWnipo 335Q9/MhqcNU3s2TO
+    }
+}
+----------
+{
+    "apps": {
+        "http": {
+            "servers": {
+                "srv0": {
+                    "listen": [
+                        ":8884"
+                    ],
+                    "routes": [
+                        {
+                            "handle": [
+                                {
+                                    "handler": "reverse_proxy",
+                                    "health_checks": {
+                                        "active": {
+                                            "headers": {
+                                                "Host": [
+                                                    "example.com"
+                                                ],
+                                                "X-Header-Keys": [
+                                                    "VbG4NZwWnipo",
+                                                    "335Q9/MhqcNU3s2TO"
+                                                ]
+                                            }
+                                        }
+                                    },
+                                    "upstreams": [
+                                        {
+                                            "dial": "127.0.0.1:65535"
+                                        }
+                                    ]
+                                }
+                            ]
+                        }
+                    ]
+                }
+            }
+        }
+    }
+}
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_health_headers.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_health_headers.txt
@@ -3,7 +3,9 @@
 reverse_proxy 127.0.0.1:65535 {
 	health_headers {
 		Host example.com
+		X-Header-Key 95ca39e3cbe7
 		X-Header-Keys VbG4NZwWnipo 335Q9/MhqcNU3s2TO
+		X-Empty-Value
 	}
 }
 ----------
@@ -25,6 +27,12 @@ reverse_proxy 127.0.0.1:65535 {
 											"headers": {
 												"Host": [
 													"example.com"
+												],
+												"X-Empty-Value": [
+													""
+												],
+												"X-Header-Key": [
+													"95ca39e3cbe7"
 												],
 												"X-Header-Keys": [
 													"VbG4NZwWnipo",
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -62,6 +62,10 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 //         health_timeout <duration>
 //         health_status <status>
 //         health_body <regexp>
+//         health_headers {
+//             Host example.com
+//             X-Header-Key value
```
francislavoie, line 67: We don't tend to use examples here, it should be the syntax definition.

```suggestion
//             <field> <value>
```

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -836,6 +853,22 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	return nil
 }
 
+// parseHealthHeaders parse health_headers block
+func parseHealthHeaders(d *caddyfile.Dispenser) (http.Header, error) {
```
francislavoie, line 857: I don't think I see any value in breaking this out to a separate function. You can inline this.

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -241,7 +242,11 @@ func (h *Handler) doActiveHealthCheck(dialInfo DialInfo, hostAddr string, host H
 		return fmt.Errorf("making request: %v", err)
 	}
 	for key, hdrs := range h.HealthChecks.Active.Headers {
-		req.Header[key] = hdrs
+		if strings.ToLower(key) == "host" {
+			req.Host = h.HealthChecks.Active.Headers.Get(key)
```
francislavoie, line 246: Aha, that's a good one, thanks! 👍

francislavoie (changes requested):

francislavoie:
Please also add a Caddyfile adapt test, see https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt

modules/caddyhttp/reverseproxy/caddyfile.go
yaxin-cn: ok, I will fix it immediately

caddytest/integration/caddyfile_adapt/reverse_proxy_health_headers.txt
```diff
@@ -0,0 +1,49 @@
...
+    health_headers {
+        Host example.com
+        X-Header-Keys VbG4NZwWnipo 335Q9/MhqcNU3s2TO
+    }
+}
+----------
+{
+    "apps": {
+        "http": {
+            "servers": {
```
francislavoie, line 13: Use tabs, not spaces, in this file.

yaxin-cn: yes, I forget to run test case before push.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -312,6 +315,26 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
 
+			case "health_headers":
+				healthHeaders := make(http.Header)
+				for d.Next() {
+					for d.NextBlock(0) {
+						key := d.Val()
+						values := d.RemainingArgs()
+						if len(values) == 0 {
+							values = append(values, "")
+						}
```
mholt, line 326: Is this necessary? Did you try without this to see if Go will still set the header with an empty value?

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -62,6 +62,9 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 //         health_timeout <duration>
 //         health_status <status>
 //         health_body <regexp>
+//         health_headers {
+//             <field> <value>
```
mholt, line 66:
```suggestion
//             <field> [<values...>]
```

Since the value is technically optional and there may also be more than one.

mholt (changes requested):
Thank you for the great PR!

Will look forward to merging this shortly after the 2.3 release (today or tomorrow).

Looks good, just wanted to double check a couple things.

modules/caddyhttp/reverseproxy/caddyfile.go
yaxin-cn: if I remove this logic, the empty header value will be `null` in json format caddyfile
image
image

modules/caddyhttp/reverseproxy/caddyfile.go
yaxin-cn: ok

modules/caddyhttp/reverseproxy/caddyfile.go
mholt: Interesting... ok. thanks! I keep forgetting how that works.

mholt (approved):
Thanks for the contribution!

francislavoie (approved):

PR #3956, opened 2021-01-06
caddytls: add 'key_type' subdirective

masipcat:
It seems that `key_type` subdirective it's not implemented in v2. I don't know if this is intentional but I needed this ASAP and I implemented it myself. Here is the code if you are interested

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -87,7 +87,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 	var folderLoader caddytls.FolderLoader
 	var certSelector caddytls.CustomCertSelectionPolicy
 	var acmeIssuer *caddytls.ACMEIssuer
-	var autoPolicy *caddytls.AutomationPolicy
+	var keyType *string
 	var internalIssuer *caddytls.InternalIssuer
 	var issuers []certmagic.Issuer
 	var onDemand bool
@@ -273,10 +273,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 				if len(arg) != 1 {
 					return nil, h.ArgErr()
 				}
-				if autoPolicy == nil {
-					autoPolicy = new(caddytls.AutomationPolicy)
-				}
-				autoPolicy.KeyType = arg[0]
+				*keyType = arg[0]
 
 			case "eab":
 				arg := h.RemainingArgs()
@@ -395,10 +392,10 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 		})
 	}
 
-	if autoPolicy != nil {
+	if keyType != nil {
 		configVals = append(configVals, ConfigValue{
 			Class: "tls.key_type",
-			Value: autoPolicy.KeyType,
+			Value: *keyType,
 		})
 	}
 
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -87,6 +87,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 	var folderLoader caddytls.FolderLoader
 	var certSelector caddytls.CustomCertSelectionPolicy
 	var acmeIssuer *caddytls.ACMEIssuer
+	var autoPolicy *caddytls.AutomationPolicy
 	var internalIssuer *caddytls.InternalIssuer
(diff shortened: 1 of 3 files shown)


mholt:
Thanks for the PR. Did you know [there's a global option to change `key_type`](https://caddyserver.com/docs/caddyfile/options#key-type)? Usually Caddyfile users just want to change the key type for all their sites; currently the JSON lets you configure key types per-domain/per-site, as most advanced users do.

What is your use case?

masipcat:
Yes, I know about the global configuration and that it can be changed per-site using JSON but I prefer to continue with the Caddyfile for now :P 

My use case is: I have a VPS with a single IP address. I run multiple services (not only http server) and I use Caddy to get let's encrypt certificates for some of these other services. I want the EC certificates for the domains/subdomains served from caddy and the RSA for the other services.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -87,6 +87,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 	var folderLoader caddytls.FolderLoader
 	var certSelector caddytls.CustomCertSelectionPolicy
 	var acmeIssuer *caddytls.ACMEIssuer
+	var autoPolicy *caddytls.AutomationPolicy
```
mholt, line 90: The only field being used is the KeyType; Instead, we can probably just make a `keyType` variable here.

mholt (changes requested):
Looks pretty good, but can simplify it slightly by not using an entire automation policy where we only need a key type variable.

francislavoie:
Please also add a caddyfile adapt test, see here: https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -87,7 +87,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 	var folderLoader caddytls.FolderLoader
 	var certSelector caddytls.CustomCertSelectionPolicy
 	var acmeIssuer *caddytls.ACMEIssuer
-	var autoPolicy *caddytls.AutomationPolicy
+	var keyType *string
```
mholt, line 90: Sorry to be pendantic, but why is this a pointer? :sweat_smile:  Probably safer to use a `string` instead. Just check for `keyType != ""`, impossible to panic accidentally.

```suggestion
	var keyType string
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -395,10 +392,10 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 		})
 	}
 
-	if autoPolicy != nil {
+	if keyType != nil {
```
mholt, line 395:
```suggestion
	if keyType != "" {
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -395,10 +392,10 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
 		configVals = append(configVals, ConfigValue{
 			Class: "tls.key_type",
-			Value: autoPolicy.KeyType,
+			Value: *keyType,
```
mholt, line 398:
```suggestion
			Value: keyType,
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -273,10 +273,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 				if len(arg) != 1 {
 					return nil, h.ArgErr()
 				}
-				if autoPolicy == nil {
-					autoPolicy = new(caddytls.AutomationPolicy)
-				}
-				autoPolicy.KeyType = arg[0]
+				*keyType = arg[0]
```
mholt, line 276:
```suggestion
				keyType = arg[0]
```

mholt (changes requested):
Thanks! Almost there. We can improve the safety of this code a bit.

caddyconfig/httpcaddyfile/builtins.go
masipcat: Not a problem. I'm new to golang :)

mholt (approved):
Looks good now! Thanks very much. Not bad for a Go beginner. :)

masipcat:
Thank you! :)

PR #3966, opened 2021-01-11
ci: deflake integration tests

Mohammed90:
This is an attempt to deflake the integration tests. The flakiness occurs on machines where the goroutine of the Caddy server doesn't get a chance to run before the main goroutine calling the server to assert its response. The approach is naive but hopefully effective. It adds `ensureConfigRunning` method which pulls Caddy for the running config every 10ms until the running config matches expectation.

The second commit was the groundwork for an attempt to parallelize the integration tests via centralized registry which controls assinging the ports to the tests within the permitted user-space range. But the tests turn out not being parallelize-able due to the Caddy process having to be a singleton process. However, I kept it to discuss whether we want to structure the integration tests in this manner anyways or not.


caddytest/caddytest.go
```diff
@@ -171,17 +176,65 @@ func (tc *Tester) initServer(rawConfig string, configType string) error {
...
-			return errors.New("caddy integration prerequisites failed. see first error")
-		}
-		return nil
-	}
+	// if hasValidated {
+	// 	if !arePrerequisitesValid {
+	// 		return errors.New("caddy integration prerequisites failed. see first error")
+	// 	}
+	// 	return nil
+	// }
```
Mohammed90, line 237: The lines were commented during the attempt to parallelize the tests. I'll revert this.

francislavoie:
Hmm. I think it only needs to be a singleton if it needs the admin API.

Another idea - could just run one instance of Caddy up front and push config changes via API each time for each test instead of shutting down? Might be faster and we can rely on the API blocking while it's reloading to avoid the flake

mholt:
Neat, thanks! I'll look at this when I get a chance. :)

sarge:
Nice! The polling of the config sounds effective, my only comment is should the admin api return only once the config is running? I realise that this could well be a non trivial change to implement, as I recall trying to trace through the load process. It might be a nice guarantee for any tooling using the api. That said reducing flakiness now would be much more valuable. As I recall it was the Windows test runs that were the most problematic? Did it fix that?

Having some tests that are sequential is a small pain, but as long as they run fast enough overall is the true goal. The revised test structure is pretty solid, I had wondered about replacement variables to avoid the 'sprintf' but because we need to know the port to call it doesn't buy us much unless the 'get' calls get the same replacement treatment which might just be too much 'magic'.

francislavoie:
> my only comment is should the admin api return only once the config is running?

I was pretty sure that was already the case :thinking:

https://github.com/caddyserver/caddy/blob/master/caddyconfig/load.go#L136

Mohammed90:
> Another idea - could just run one instance of Caddy up front and push config changes via API each time for each test instead of shutting down? Might be faster and we can rely on the API blocking while it's reloading to avoid the flake

From what I've seen, it doesn't shutdown the server after every TestXXX.


> > my only comment is should the admin api return only once the config is running?
> 
> I was pretty sure that was already the case 🤔
> 
> https://github.com/caddyserver/caddy/blob/master/caddyconfig/load.go#L136

Here's where it gets weird: it does per the code, but the flakiness is observed in placed where it's obvious the new config isn't active yet. For example, this test returns empty 200 at times, which is the result of unconfigured server. I think there's a goroutine is spawned somewhere whereas the main goroutine continues.

https://github.com/caddyserver/caddy/blob/14f50d9dfb6c0da0da636829175a37b1864d23cf/caddytest/integration/caddyfile_test.go#L74-L103

> As I recall it was the Windows test runs that were the most problematic? Did it fix that?

Not sure about Windows, but I've noticed it on the s390x machine. The only way to validate is by running the test suite multiple times, and still won't be definite. What's the best way to validate absence of flakiness?
 
> Having some tests that are sequential is a small pain, but as long as they run fast enough overall is the true goal.

I tried to parallelize the tests, but it isn't possible because Caddy runs as singleton process. However, I wonder if it's possible to run multiple Caddy processes with different names, i.e. changing `os.Args[0]` to something else at every instance. Managing that would be complex though.

sarge:
@Mohammed90 I'm pretty sure I have seen that delay between applying the new config and it being active. I think your solution is good for the case in hand.

Managing starting and stopping processes running on different ports does get pretty challenging, operating systems can be temperamental about recycling ports between processes quickly, more background processes cleaning them up I suspect.

Without the parallel support the new test structure doesn't offer too much. I do like that the test harness offers the high ports. Perhaps that would be the right path for now, is we ask the test harness for an http and https port that can be used during a test. Keeping the ports out of the config should allow a more dynamic approach should any of the remaining singleton limitations be relaxed in the future.

Mohammed90:
> @Mohammed90 I'm pretty sure I have seen that delay between applying the new config and it being active. I think your solution is good for the case in hand.
> 
> Managing starting and stopping processes running on different ports does get pretty challenging, operating systems can be temperamental about recycling ports between processes quickly, more background processes cleaning them up I suspect.
> 
> Without the parallel support the new test structure doesn't offer too much. I do like that the test harness offers the high ports. Perhaps that would be the right path for now, is we ask the test harness for an http and https port that can be used during a test. Keeping the ports out of the config should allow a more dynamic approach should any of the remaining singleton limitations be relaxed in the future.

Since the new test harness doesn't provide much value in terms of parallelization, let's forgo it, at least for now.

mholt:
@sarge You're marked as a reviewer on this; have you had a chance to give this a look? I think we can merge it if you agree.

sarge (approved):
Nice job

Issue #3958, opened 2021-01-08
Add Caddyfile options to configure error log output


kujenga:
This issue is a request to add functionality to the Caddyfile to configure the error log output in a similar manner to how the `log` directive [1] allows this for "access logs". I was motived to open this issue by some of the discussion here: https://github.com/caddyserver/caddy/issues/3744#issuecomment-733288472 which resonates with needs I have run into recently. I haven't been able to find a way to do this so far, which seems to be confirmed by this somewhat recent forum post: https://caddy.community/t/where-to-put-error-log-in-caddyfile/9896/2

The specific situation I've seen motivating this request is that when you have an inbound request with sensitive headers set on it like `Cookie` or `Authorization` values that is routed to a `reverse_proxy` stanza, and if there is an error proxying the request to that backend service, you end up with an error message that looks like the following (I removed some of the default browser headers for simplicity):

```
2021/01/08 04:21:15.642	ERROR	http.log.error.log0	dial tcp [::1]:8080: connect: connection refused	{"request": {"remote_addr": "127.0.0.1:60809", "proto": "HTTP/2.0", "method": "GET", "host": "localhost", "uri": "/example", "headers": {"Accept": ["*/*"], "Accept-Encoding": ["gzip, deflate, br"], "Accept-Language": ["en-US,en;q=0.9"], "Cookie": ["secret_session_id=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"]}, "tls": {"resumed": false, "version": 772, "cipher_suite": 4865, "proto": "h2", "proto_mutual": true, "server_name": "localhost"}}, "duration": 0.009280984, "status": 502, "err_id": "5qpcc54u3", "err_trace": "reverseproxy.statusError (reverseproxy.go:783)"}
```

In this case, it would be a security issue to capture the value of `"Cookie": ["secret_session_id=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"]` in application logs on the server.

I understand that it may be possible to exclude this type of log with the JSON config [2] but it would be quite convenient to do this with the Caddyfile.

The equivalent functionality in the access logs would be a `request>headers>Cookie delete` statement similar to the examples for the access log directive: https://caddyserver.com/docs/caddyfile/directives/log#examples

[1] https://caddyserver.com/docs/caddyfile/directives/log
[2] https://caddyserver.com/docs/json/logging/

kujenga:
[...]

PR #3968, opened 2021-01-12
Allow Caddyfile configuration of arbitrary loggers, add new replace log filter

kujenga:
This PR follows up on discussion in https://github.com/caddyserver/caddy/issues/3958 about adding more functionality to the Caddyfile. I also added in another commit for a small new "replace" filter aimed at the same goal as I wrote up in the original ticket, providing more tools to remove sensitive data from logs.

Closes https://github.com/caddyserver/caddy/issues/3958

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -548,18 +548,32 @@ func parseHandleErrors(h Helper) ([]ConfigValue, error) {
 
 // parseLog parses the log directive. Syntax:
 //
-//     log {
-//         output <writer_module> ...
-//         format <encoder_module> ...
-//         level  <level>
+//     log [name] {
+//         output  <writer_module> ...
+//         format  <encoder_module> ...
+//         level   <level>
+//         include <namespaces...>
+//         exclude <namespaces...>
 //     }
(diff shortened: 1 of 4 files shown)


francislavoie:
Mmm, no, I don't think this is right. We don't want to modify the `log` directive, we want to introduce a global option.

kujenga:
Hm, I do see why you say that, I was thinking something similar when looking at what actual configs for this would be.

Might be a bit of a refactor to avoid duplicating the parsing logic, since I think we do want to share most/all of it between the two? I'll give it a try

mholt:
I've been giving this some thought, and am coming to the conclusion that Caddy's existing log emission facilities are too complex already.

I think this kind of logic belongs in a log _processing_ facility. There are plenty of external tools for processing logs that will do stuff like this.

I also see value in having something like that built into Caddy. And, I think it'd make a perfect Caddy app, called `log`, which ingests logs and processes them. Its input would be the raw stream of bytes, and then it could parse, filter, etc, all sorts of logs.

francislavoie:
I think I'll close this for now, since the impl isn't what we want. You can open a new PR if you decide to tackle configuring logging with globals.

PR #3969, opened 2021-01-12
(v2.3.0 base, temporary) Allow Caddyfile configuration of arbitrary loggers, add new replace log filter

kujenga:
This is a temporary PR to keep track of a branch with the same changes as https://github.com/caddyserver/caddy/pull/3968, except for that it is based off of the v2.3.0 tag so that it can be tested against other systems without incorporating unreleased changes in that testing process.


kujenga:
Meant to open this within my own repo, sorry for the noise

PR #3980, opened 2021-01-21
support function map for custom template of browse

snowcrumble:
I have a requirement to use `github.com/Masterminds/sprig/v3.hasSuffix` function in my custom static browse template file.


snowcrumble:
I commited with wrong git user.name, close this, reopen in #3982

Issue #3637, opened 2020-08-05
Allow `file_server browse` templates to use the same functions as `templates`


francislavoie:
See the discussion here: https://caddy.community/t/v2-http-handlers-templates-functions-in-file-server-browse-template/9214

The [`file_server` directive](https://caddyserver.com/docs/caddyfile/directives/file_server) has a `browse` option which enables a directory index to be served when a directory is requested. By default, this index is rendered using a template which can be [found in the code](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/fileserver/browsetpl.go). The default can be overridden with a template provided by the user.

It seems that the `file_server browse` template doesn't use the same code paths as the [`templates` directive](https://caddyserver.com/docs/caddyfile/directives/templates), which means that the [functions documented here](https://caddyserver.com/docs/modules/http.handlers.templates) cannot also be used.

The code should be refactored such that the logic for the templates is shared between these two such that all the template functions Caddy provides works both for the general templates _and_ for the file server directory index.

divbhasin:
Hi, new contributor here 👋.  I can work on this issue.

francislavoie:
Cool! Go for it @divbhasin 😄

divbhasin:
Here is what I am thinking: I need to make the `executeTemplateInBuffer` function in `templates/tplcontext.go` accessible to fileserver browse. This is because that function attaches the other functions like `httpInclude` to a template, and returns a template that has been executed on a buffer.

Am I on the right track @francislavoie?

francislavoie:
Yeah, that sounds about right. Maybe the `templateContext` should be shared to allow `browse` to use it.

vojkuv:
``https://www.facebook.com/tom.zdrali/about?section=relationship&lst=100001658457086%3A100036018322527%3A1598790302 #3691

vojkuv:
[](https://www.facebook.com/tom.zdrali/about?section=relationship&lst=100001658457086%3A100036018322527%3A1598790302[](url))

mholt:
Hmm, this is tricky though because in order to support of all of `templates` functions you'd need to import the `modules/caddyhttp/templates` package and use its `Templates` type, because `templateContext` depends on it.

This will need more consideration/discussion; I think there'll need to be some sort of new function or method that is exported to wrap all the unexported logic needed for this, without having to create a whole HTTP handler or something.

divbhasin:
[...]

PR #3982, opened 2021-01-21
support function map for custom template of browse

snowcrumble:
I have a requirement to use github.com/Masterminds/sprig/v3.hasSuffix function in my custom static browse template file.

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Masterminds/sprig/v3"
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


francislavoie:
This is related to https://github.com/caddyserver/caddy/issues/3637

I think we'd rather do this "right" and have the logic that provides the functions to `file_server browse` and to `templates` be the same.

Do you think you could try to do that instead? It's more work, but it would solve an existing issue we have.

Thanks!

snowcrumble:
Thanks for prompt reply. 

I'm new to caddy and I just looked #3637, then I scan related codes and I found there are two kinds of functions,  sprig's and templateContext's. The requirements of both issues is support functions in FileServer. To avoid problem of packages circle dependencies, we may create a new package for this functions.  What I confused is whether the templateContext's functions is useful in `FileServer`, cause a FileServer seems like need no `Req *http.Request` and `Args []interface{}` fields. My point is if it's not useful, the `github.com/Masterminds/sprig/v3` package is the "new" the package, so my changes is a resolution for this requirements. What do you think?

mholt:
> Ok, we needs to refactor! To avoid problem of packages circle dependencies, we may create a new package for this functions.

Why would it be circular? The `caddyhttp/fileserver` package already imports the `caddyhttp/templates` package. And what you're looking to add is a templates feature to the file server, right? So I don't think that would be a problem, unless I'm misunderstanding something.

francislavoie:
@snowcrumble do you still plan to continue work on this?

mholt:
Closing due to inactivity. Willing to reopen later if interest resurfaces!

PR #3991, opened 2021-01-25
map: Replace placeholders with regex groups

### LINKED ISSUE #3971: Feature request: map handler takes account of regex capture group
Discussion in
https://caddy.community/t/map-handler-use-regex-capture-group-in-destination/11104

I want to refer to the captured parts of the regex in destination, e.g.

```
map * {host} {my_folder} {
  ~(.+)\.localhost \\1
  default "fail"
}
```

@mholt : The map handler doesn’t set placeholders like this, but it could, someone can contribute that functionality. Would be pretty similar to what’s in the regexp matchers.

@francislavoie suggested a nice workaround for this particular case
```
@myfolder header_regexp folder Host (.+)\.localhost
handle @myfolder {
	root * /var/www/{re.folder.1}
}

handle {
	root * /var/www/fail
}
```

rajat315315:
Fixes #3971

### DIFF
--- a/caddytest/integration/map_test.go
+++ b/caddytest/integration/map_test.go
@@ -19,19 +19,19 @@ func TestMap(t *testing.T) {
 
 		map {http.request.method} {dest-1} {dest-2} {
 			default unknown1    unknown2
-			~G.T    get-called
-			POST    post-called foobar
+			~G([\w]+)   G{http.regexp.1}	called
+			POST    post-called
 		}
 
 		respond /version 200 {
 			body "hello from localhost {dest-1} {dest-2}"
-		}	
+		}
 	}
 	`, "caddyfile")
 
 	// act and assert
-	tester.AssertGetResponse("http://localhost:9080/version", 200, "hello from localhost get-called unknown2")
-	tester.AssertPostResponseBody("http://localhost:9080/version", []string{}, bytes.NewBuffer([]byte{}), 200, "hello from localhost post-called foobar")
+	tester.AssertGetResponse("http://localhost:9080/version", 200, "hello from localhost GET called")
+	tester.AssertPostResponseBody("http://localhost:9080/version", []string{}, bytes.NewBuffer([]byte{}), 200, "hello from localhost post-called unknown2")
 }
 
 func TestMapRespondWithDefault(t *testing.T) {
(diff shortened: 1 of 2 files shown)


francislavoie:
I don't think this is quite right. The linked issue is asking to essentially do the same as the `path_regexp` matcher, where it sets regexp results in the replacer to be used later. Basically, this code needs to be ported to the map handler when a regexp matches:

https://github.com/caddyserver/caddy/blob/deedf8a/modules/caddyhttp/matchers.go#L932-L944

rajat315315:
Is it good? Or better?

francislavoie:
Hmm. I'm not sure that's still quite right. The typical regexp placeholder involves a name, because you could have more than one regexp in your handler chain (i.e. a `path_regexp` or `header_regexp` matchers, etc). Your implementation will clobber those since it's placing the results at the same placeholder depth as where the names would normally be.

I think we need to figure out how to tell the `map` handler the regexp name to be used. I'm not sure what's right there. @mholt do you have any ideas?

mholt:
@francislavoie I'm just getting to this now, but I'm confused. What was wrong with the initial commit? The current commit looks way complicated. `ReplaceAllString()` is what allows you to use capture groups like `\1` in replacements. Placeholders are a separate concept. Why does this have anything to do with matchers?

francislavoie:
@mholt the linked issue wants to use the regexp result later outside of `map`.

mholt:
Hmm are you sure? Linked issue says:

> I want to refer to the captured parts of the regex in destination

francislavoie:
Oh, I see, my bad. I misunderstood how the first implementation was meant to be used because there was no explanation or associated test. I figured we would use the same regexp placeholders like `{re.*.*}` here but it makes sense to use `\*` for using the capture groups inside the map case.

mholt:
I'm pretty sure we can restore the first commit @rajat315315 -- but keep the later commits.

rajat315315:
Sure

mholt:
I believe we want to keep this one: https://github.com/caddyserver/caddy/commit/acb5ff39d7e1f98759259f4682570ad5fc6bf751

And any related test cases. :)

rajat315315:
> -- but keep the later commits.

I'm not sure about this statement..

francislavoie:
By "later commits", Matt means the changes to the tests. Basically just revert the changes after the first commit, then add tests.

francislavoie (approved):

mholt (approved):
Cool, let's try this out. Hopefully nobody wants literal `${1}` strings in their outputs. :)

rajat315315:
Hello everyone,
I request you to please nominate me, if you think I have done some meaningful contribution to the community...
https://stars.github.com/nominate/

Issue #3913, opened 2020-12-12
Add support for wildcards/ expression matching in header_up/ header_down


gucki:
It seems Caddy does not support wildcards/ expression matching in the reverse proxy's header_up. 

For example this does not remove any headers:

    header_up -X-SHIBBOLETH-*

This works as expected, but is limited in its use:

    header_up -X-SHIBBOLETH-LOGIN

It'd be great if Caddy would support  wildcards/ expression matching in the reverse proxy's header_up/ header_down.

simaotwx:
See https://github.com/caddyserver/caddy/pull/3992#issuecomment-802669782 for my suggestion

Mohit-Pawar:
@mholt I would like to work on some good first issue. can i take this up?

francislavoie:
Sure @Mohit-Pawar, just be aware of the prior work in https://github.com/caddyserver/caddy/pull/3992

Jonathanmarsh4:
https://2.148.190.199/6988.png

0640811620:
https://www.123imiwin.com/users/member

FbAungMyo:
https://www.facebook/aung.myo.8

ceoaungmyo:
http://www.facebook.com/ceoaungmyo

ALEKSHA2282:
Я ВООБЩЕ НЕЧЕГО НЕ ПОНИМАЮ

yassery7y7:
> 


https://www.snapchat.com/add/yassery7y7

yassery7y7:
yassery7y7y7y7@gmail.com

Rawadwalf:
0931415333

Rawadwalf:
Rawad Walf

Zagorrrr:
0x83f559ba8AD01F7c1A1AfE9715FcE5d344753465

francislavoie:
This is completed by https://github.com/caddyserver/caddy/pull/4831

PR #3992, opened 2021-01-25
Added regex pattern matching for Deleting Headers

rajat315315:
Fixes: #3913

### DIFF
--- a/caddytest/integration/caddyfile_adapt/header.txt
+++ b/caddytest/integration/caddyfile_adapt/header.txt
@@ -2,7 +2,7 @@
 	header Denis "Ritchie"
 	header +Edsger "Dijkstra"
 	header ?John "von Neumann"
-	header -Wolfram*
+	header -Wolfram.*
 	header {
 		Grace: "Hopper" # some users habitually suffix field names with a colon
 		+Ray "Solomonoff"
@@ -87,9 +87,7 @@
 									"response": {
 										"deferred": true,
 										"delete": [
-											{
-												"regex": "Wolfram*"
-											}
+											"Wolfram.*"
 										]
 									}
 								},
--- a/caddytest/integration/caddyfile_adapt/header.txt
+++ b/caddytest/integration/caddyfile_adapt/header.txt
@@ -2,7 +2,7 @@
 	header Denis "Ritchie"
 	header +Edsger "Dijkstra"
 	header ?John "von Neumann"
-	header -Wolfram
+	header -Wolfram*
 	header {
 		Grace: "Hopper" # some users habitually suffix field names with a colon
 		+Ray "Solomonoff"
--- a/caddytest/integration/caddyfile_adapt/header.txt
+++ b/caddytest/integration/caddyfile_adapt/header.txt
@@ -87,7 +87,9 @@
 									"response": {
 										"deferred": true,
 										"delete": [
-											"Wolfram"
+											{
+												"regex": "Wolfram*"
+											}
 										]
 									}
 								},
--- a/modules/caddyhttp/headers/headers.go
+++ b/modules/caddyhttp/headers/headers.go
@@ -119,10 +119,13 @@ type HeaderOps struct {
 	Set http.Header `json:"set,omitempty"`
 
 	// Names of HTTP header fields to delete.
-	Delete []DeleteHeader `json:"delete,omitempty"`
+	Delete []string `json:"delete,omitempty"`
 
 	// Performs substring replacements of HTTP headers in-situ.
 	Replace map[string][]Replacement `json:"replace,omitempty"`
+
+	// compiled regex-exp for Delete
+	deleteRegexps map[string]*regexp.Regexp
 }
 
 // Provision sets up the header operations.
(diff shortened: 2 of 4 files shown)


francislavoie:
Please add tests for this!

francislavoie:
Also, I think this won't be quite as efficient as it could be. I think the regexp should be compiled ahead of time during the provisioning phase if possible.

rajat315315:
Sure, would do that.

rajat315315:
Also, can anybody point out the function in tests where I can compare/match headers?

rajat315315:
Is it beeter? @francislavoie

caddytest/integration/caddyfile_adapt/header.txt
```diff
@@ -2,7 +2,7 @@
 	header Denis "Ritchie"
 	header +Edsger "Dijkstra"
 	header ?John "von Neumann"
-	header -Wolfram
+	header -Wolfram*
```
francislavoie, line 5: This should be `.*`, no? I don't think `*` is useful there unless we want to say that `Wolframmmm` is useful to match 😂 

```suggestion
	header -Wolfram.*
```

caddytest/integration/caddyfile_adapt/header.txt
```diff
@@ -87,7 +87,9 @@
 									"response": {
 										"deferred": true,
 										"delete": [
-											"Wolfram"
+											{
+												"regex": "Wolfram*"
```
francislavoie, line 91:
```suggestion
												"regex": "Wolfram.*"
```

modules/caddyhttp/headers/headers.go
```diff
@@ -119,7 +119,7 @@ type HeaderOps struct {
 	Set http.Header `json:"set,omitempty"`
 
 	// Names of HTTP header fields to delete.
-	Delete []string `json:"delete,omitempty"`
+	Delete []DeleteHeader `json:"delete,omitempty"`
```
francislavoie, line 122: I'm worried about this change though, because this would be a breaking change for any JSON users. Whatever the solution is, it should retain the fact that a string can be used to configure deleted headers, for backwards compatibility.

My suggestion might be to instead of `DeleteHeader`, make a `deleteRegexps map[string]*regexp.Regexp` field which retains the compiled regexps, by the header config string. On provision, if it's a regexp, then add the compiled regexp to the map. Then you would do a lookup in the map by the `.Delete` string at runtime to fetch the regexp if any, and if not then just delete as normal. This should keep backwards-compat, I think.

@mholt how does this sound to you?

francislavoie (changes requested):

modules/caddyhttp/headers/headers.go
rajat315315: Makes sense. I agree with you.

modules/caddyhttp/headers/headers.go
```diff
@@ -138,6 +141,15 @@ func (ops *HeaderOps) Provision(_ caddy.Context) error {
 			}
 		}
 	}
+	for _, delHeader := range ops.Delete {
+		if strings.ContainsAny(delHeader, "+|*|[") {
```
mholt, line 145: I know this may be unlikely, but I'm not sure we can safely assume that header field names are alphanumeric.

I had to look this up in the HTTP/2 spec, which actually defers to the HTTP/1.1 spec for describing acceptable header field values, and that spec says the header field names can be `token`, which I finally found, is defined as:

```
<any CHAR except CTLs or separators>
```

which is basically any ASCII character, including the "special" regex characters.

We also can't assume that a regex _does_ have one of those characters, necessarily. A regex can also be something like `foo`.

I think there'll need to be some sort of other way to indicate that the user is specifying a regular expression.

mholt (changes requested):
Thanks for the improvements! It's looking good, but I think the fundamental regex-detecting logic is flawed. What do you suggest for a good way for the user to indicate they are specifying a regex?

modules/caddyhttp/headers/headers.go
francislavoie: Maybe `-~` prefix? Is that insane? :joy:

mholt: Maybe, if it's really needed. I'd like to see if a simple prefix match would suffice first, since that'd be fairly intuitive and really really simple.

mholt:
Closing due to inactivity.

modules/caddyhttp/headers/headers.go
rajat315315: Can we go with `-~` ?

mholt: Is there an actual use case that needs more than a prefix match?

simaotwx:
Suggestion: We could use `header_up_regexp`and `header_down_regexp` for regex matching or add an option to `header_up`/`header_down` to enable prefix or regex matching.

Examples:

```
header_up_regexp -X-Some-Header[*]-A.*
```

Or

```
header_up -X-Some-Header- {
    match_prefix
}
```

fuckpass8629 (commented):
caddytest/integration/caddyfile_adapt/header.txt

mholt:
@simaotwx That's probably a reasonable option, although we might be able to scrape by with a `~` prefix, kind of like we do with the `map` handler.

(And in hindsight I wish I exposed other regular expression properties in the config as `re` instead of `regexp` in the config, for brevity; oh well.)

0640811620 (approved):
https://www.123imiwin.com/users/member

PR #3995, opened 2021-01-26
ci: update the command to run tests on the s390x machine

Mohammed90:
I've updated Go installation on the machine from go1.14.3 to go1.15.6. Actually, I've replaced the unpacked tarball with the Snap package, which enables automatic updates.


mholt (approved):
Awesome, thanks so much for maintaining this @Mohammed90 ! :100:

francislavoie (approved):

PR #4001, opened 2021-01-29
ci: make CI run faster

Mohammed90:
Speed it up using  [WillAbides/setup-go-faster](https://github.com/WillAbides/setup-go-faster) rather than actions/setup-go


Mohammed90:
The numbers aren't very optimistic. The total gain of ~10 seconds and adding `tip` obliterates any gain, so the shift isn't very beneficial.

mholt:
Bummer. Thanks for exploring this!

Issue #2268, opened 2018-08-17
Latency timer should stop early for websocket, text/event-stream connections


sundbry:
As mentioned in issue #593, "latency" should be measured as the time to first response for streaming connections, such as websockets or EventSource connections, and the current behavior should probably be called "duration" instead.

The issue comes up when analyzing access logs; event source and websocket connections show as being abnormally slow to respond.

EventSource connections can be identified with "text/event-stream" as the response content-type.

mholt:
Somehow I never saw this issue until now but you make a good point. Will fix in Caddy 2.

mholt:
@sundbry This is implemented in #4013: `{http.reverse_proxy.upstream.latency}` which tells you the time for the backend to write the headers, and `{http.reverse_proxy.upstream.duration}` gives you total streaming time.

PR #4013, opened 2021-02-08
reverseproxy: Add duration/latency placeholders (close #4012)

### LINKED ISSUE #4012: Feature Request: Placeholders (at header time) for request time & upstream time.
Hello all,

This feature request is stemming from [Any equivalent of $request_time and $upstream_header_time from nginx?](https://caddy.community/t/any-equivalent-of-request-time-and-upstream-header-time-from-nginx/11418). To summarize, there are two placeholders that I'm using in nginx now that would be very useful to me in caddy:

1) [request_time](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_time):

> request processing time in seconds with a milliseconds resolution; time elapsed since the first bytes were read from the client.

2) [upstream_header_time](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_header_time):

> keeps time spent on receiving the response header from the upstream server; the time is kept in seconds with millisecond resolution.

Note that nginx does have an [upstream_response_time](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_response_time), but this isn't available at header time because nginx forwards the headers returned by upstream before it receives the response body. I don't know if Caddy does this, or if it instead buffers the response and only computes headers after response is fully received (and fully timed), but for this feature request, `upstream_header_time` and `upstream_response_time` are close enough where a port of either one would be great.

I would use these placeholders to signify to our client how much of the latency they observed was due to their upload speeds vs. our processing speeds. For example, if I returned `X-Trip-Time == 5` (where 5 is the equivalent of `request_time`) and `X-Process-Time == 3` (where 3 is the equivalent of `upstream_header_time` / `upstream_response_time`), then they would know that they spent roughly 2 seconds uploading.

If this is a feature request that the maintainers are okay with and it's a good first issue, I might be able to take a shot at this. Just would need a pointer on where to start.

mholt:
Adds 4 placeholders, one is actually outside reverse proxy though:

`{http.request.duration}` is how long since the server decoded the HTTP request (headers).
`{http.reverse_proxy.upstream.latency}` is how long it took a proxy upstream to write the response header.
`{http.reverse_proxy.upstream.duration}` is total time proxying to the upstream, including writing response body to client.
`{http.reverse_proxy.duration}` is total time spent proxying, including selecting an upstream and retries.

Obviously, most of these are only useful at the end of a request, like when writing response headers or logs.

See also: https://caddy.community/t/any-equivalent-of-request-time-and-upstream-header-time-from-nginx/11418

Closes #4012

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -49,6 +49,7 @@ func init() {
 // ------------|---------------
 // `{http.request.body}` | The request body (⚠️ inefficient; use only for debugging)
 // `{http.request.cookie.*}` | HTTP request cookie
+// `{http.request.duration}` | Time up to now spent handling the request (after decoding headers from client)
 // `{http.request.header.*}` | Specific request header field
 // `{http.request.host.labels.*}` | Request host labels (0-based from right); e.g. for foo.example.com: 0=com, 1=example, 2=foo
 // `{http.request.host}` | The host part of the request's Host header
--- a/modules/caddyhttp/replacer.go
+++ b/modules/caddyhttp/replacer.go
@@ -36,6 +36,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
@@ -52,6 +53,8 @@ func NewTestReplacer(req *http.Request) *caddy.Replacer {
 }
 
 func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.ResponseWriter) {
+	SetVar(req.Context(), "start_time", time.Now())
+
 	httpVars := func(key string) (interface{}, bool) {
 		if req != nil {
 			// query string parameters
@@ -140,6 +143,9 @@ func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.Respo
 				return dir, true
 			case "http.request.uri.query":
 				return req.URL.RawQuery, true
+			case "http.request.duration":
+				start := GetVar(req.Context(), "start_time").(time.Time)
+				return time.Since(start), true
 			case "http.request.body":
 				if req.Body == nil {
 					return "", true
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -44,7 +44,7 @@ func init() {
 // Handler implements a highly configurable and production-ready reverse proxy.
 //
 // Upon proxying, this module sets the following placeholders (which can be used
-// both within and after this handler):
+// both within and after this handler; for example, in response headers):
 //
 // Placeholder | Description
 // ------------|-------------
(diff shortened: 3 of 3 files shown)


francislavoie:
Should add these the godocs, right?

mholt:
Yeah, good point. Doing that now.

francislavoie (approved):

MarioIshac:
Alright this is looking great, given [this caddy.json](https://gist.github.com/MarioIshac/268d7f62bae47580529648401a4d9b20) I've successfully received `{http.request.duration}` and `{http.reverse_proxy.upstream.latency}` as headers. And, as expected, `{http.reverse_proxy.upstream.duration}` and `{http.reverse_proxy.duration}` are not available at header-time because Caddy has already sent the response by then.

Note that `"buffer_requests": true` was required to make `{http.reverse_proxy.upstream.latency}` independent of file uploading time from the client. When Caddy doesn't buffer a request (by default), it streams the request bytes to upstream, meaning that upstream is contacted (and `{http.reverse_proxy.upstream.latency}` starts going) before a client finished sending the request. In my case, this caused the upstream latency to look big (and actually roughly equal the request duration) even though my client was a slow uploader.

What if `{http.reverse.proxy.upstream.*}` started going after the last byte of the request finished streaming, instead of the first byte? Would this make it independent from `buffer_requests`?

mholt:
@MarioIshac Great, thanks for trying it out.

> What if {http.reverse.proxy.upstream.*} started going after the last byte of the request finished being sent upstream, instead of the first byte?

I'm not sure that's a reasonable assumption we can make: AFAIK there's no reason a backend can't start sending the response before it has finished reading the request, and in fact I'm pretty sure quite a few servers work this way (and has been the cause of some bugs in the Go standard library in the past).

> In my case, this caused the upstream latency to look big (and actually roughly equal the request duration) even though it was my client's fault for being slow uploader.

I think that's actually good information from these placeholders -- the small difference makes it easy to show that the client uploading is the bulk of the request time, _but this deduction requires knowing how your backend works._ A backend will either wait to download the entire body before writing a response, or it will start writing a response before the body has been fully read.

francislavoie:
I figure we can merge this?

mholt:
Is the implementation of `{http.reverse_proxy.upstream.latency}` satisfactory though?

MarioIshac:
Currently using the build artifact from this PR in prod, been working well. Let me know if there's anything more I can do to help get this merged, but don't have the expertise to comment on the implementation itself though.

francislavoie:
I think the only open question was whether we agree how `.latency` behaves, but if Matt's answer is satisfactory I think we can merge it.

Thanks!

MarioIshac:
Matt's answer makes sense to me, especially after I looked into what types of servers would send response back before request is complete (sending "unauthorized" instead of waiting for file upload to complete, for example). Because this is an implementation detail of the backend, the backend can send the time it received the last byte of the request if desired outside of Caddy.

francislavoie:
I think this might resolve some of the concerns in https://github.com/caddyserver/caddy/issues/2268 as well @mholt?

mholt:
@francislavoie Hmm, yes I believe it does. Between these 4 placeholders, they should be able to get the info they need. Thanks for linking that up.

I'll go ahead and merge this then. Thank you both!

Issue #3784, opened 2020-10-08
Support per-host Prometheus metrics


muety:
Initiated through [this](https://caddy.community/t/per-host-prometheus-metrics/10035) forum post – involving @hairyhenderson –, I'd like to open the discussion about extending the `metrics` directive to support metrics (request count, etc.) on a per-host basis.

I'd like to have metrics exposed for every single site block (speaking in Caddyfile terminology), i.e. metrics about different sites / vhosts result in different Prometheus metrics labels. 

Since determining the labels automatically based on the incoming requests' host headers is problematic (see discussion above), a solution would be to manually assign a tag to the `metrics` directive. For instance:

```
example.org {
    root /var/www/html
    file_server
    metrics /metrics {
        host "example.org"
    }
}
```

I'm not really involved with Caddy's code base or development process, so this is rather supposed to be a starter for further discussion.

hairyhenderson:
Thanks for logging this @muety!

Just to bring the goal into view, the v1 prometheus module had the ability to set a value to be used by the `host` label, which effectively solved the cardinality problem (i.e. the fact that there isn't a bounded set of values that can be used if it's determined automatically by looking at the `Host` header).

Caddy v2 is quite different from v1, and one of those differences is that [request matchers](https://caddyserver.com/docs/caddyfile/matchers) are used extensively, and now the hostname in the site block is actually just a different kind of matcher.

This is maybe best explained by using `caddy adapt` to convert the following `Caddyfile` into the canonical JSON format:

```
example.com {
  respond /foo "hello world"
}
example2.com {
  respond /bar "hello world"
}
example2.com:8443 {
  respond /baz "hello world"
}
```

Here's the result of running `caddy adapt`:

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [ ":443" ],
          "routes": [ {
              "match": [ { "host": [ "example2.com" ] } ],
              "handle": [ {
                  "handler": "subroute",
                  "routes": [ {
                      "handle": [ { "body": "hello world", "handler": "static_response" } ],
                      "match": [ { "path": [ "/bar" ] } ]
                    } ]
                } ],
              "terminal": true
            },
            {
              "match": [ { "host": [ "example.com" ] } ],
              "handle": [ {
                  "handler": "subroute",
                  "routes": [ {
                      "handle": [ { "body": "hello world", "handler": "static_response" } ],
                      "match": [ { "path": [ "/bar" ] } ]
                    } ]
[...]

PR #4015, opened 2021-02-11
caddyhttp: Add content type to metrics

dtelyukh:


hairyhenderson:
Hi @dtelyukh, thanks for the contribution!

Can I get some background on your use-case for this? I've never seen a good use-case for providing `content-type` as a label in metrics. This may be something that's more appropriate to scrape with logs, or for our future plans to integrate a distributed tracing mechanism.

Also, similar to the discussion in #3784, `content-type` may be a source of unbounded high-cardinality, which will cause issues when scraping with Prometheus over long periods of time and with larger deployments.

Finally, it's best to log an issue before working on a PR, to make sure that the reason makes sense. I won't close this right away, but we do need to understand the full context for this.

dtelyukh:
The goal is to separate dynmaic and static resources, and there are two reasons for this:

1) Dynamically generated resources usually have way bigger TTFB, and mixing such resources with static files we obscure actual peaks in data, because each dynamically generated HTML document also pulls 10 to 100 static files. Content-type is a good heuristic for dynamic content, because html is usually dynamically generated, and everything else is static in most cases

2) No matter if html is dynamic or static, its performance is more critical for overall page load time, because it's the root document for the page, and each millisecond of its load time is always added to overall page load time, which is not always the case for page resources.

We intended to create two extra grafana panels, to be able to see metrics subsets that represent:
- html documents only
- small static files (js/css)
- large static files (images)

> our future plans to integrate a distributed tracing mechanism.

Tracing is a very useful method, but it's even more likely to get bloated when collecting data from all requests. I'd rather use metrics here, though I'm also interested in looking at tracing options. Might be a powerful method to diagnose slow requests.
Could you please point me at the discussion about tracing feature?

> similar to the discussion in #3784, content-type may be a source of unbounded high-cardinality

Totally agree! And thanks for pointing at that issue, it's very helpful. It probably makes sense to add a toggle for this feature. Also, I had a discussion with my colleagues about per-host metrics too, and we're interested in that feature too. I'll create an issue and link it here.

hairyhenderson:
Thanks for your explanation @dtelyukh, it makes sense! I'll add some further comments on #4016

francislavoie:
For the reasons mentioned already, I don't think this is the approach we want to take, so I'll close this for now. Discussion continues in #4016.

Issue #3828, opened 2020-10-29
Support for nginx style X-Accel headers in reverse_proxy


segevfiner:
Nginx has support for an upstream server to effect it by setting `X-Accel-*` headers in the response ([NGINX | X-Accel](https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/)).

This feature is very useful for various different scenarios, for example:
1. Disabling buffering for certain requests without having to write specialized configuration in the proxy server for them. e.g. A route that returns large files (`X-Accel-Buffering`).
2. Offloading protected/restricted static files to the proxy server, by an `internal`/named location only accessible ro such internal redirects (`X-Accel-Redirect`).
3. Offloading certain requests to a different server proxied by Nginx. Often an `internal`/named location, only accessible by such internal redirects (`X-Accel-Redirect`).

Some of this behavior is probably possible by `handle_response`, which is not currently available in the Caddyfile (https://github.com/caddyserver/caddy/issues/3707), and even if it would be available, a stock handler/config snippet for this behavior would still be useful.

Note that there are some fine details with how nginx behaves with those headers, for example: It changes the request method to a `GET` unless you redirect to a named location when using `X-Accel-Redirect`. Both changing the method (e.g. A `POST` that redirects to a static file), and not changing it (e.g. Offload to a different proxied server) are useful and need consideration how to make them available in caddy.

It also doesn't seem like Caddy has such named locations or internal locations/matcher to create such internal redirect only routes.

Also another interesting custom header of this sort is `X-Sendfile` supported by Apache.

segevfiner:
I recently encountered a bad interaction between `X-Accel-Redirect` and request buffering in Nginx. If you disable request buffering, Nginx just doesn't send the body to the redirected upstream with no error on its side, with no obvious way to direct it to hold off on the body to send after the redirect.

Nginx is always such a pain in the ass...

mholt:
Caddy's JSON config is capable of supporting very nuanced routes and handling logic, so I wouldn't worry about that. The functionality, if it does not already exist, just needs to be implemented. 

This is a pretty vague description of everything though, and there seems to be a lot to unpack. Frankly, I'm not really sure what you're getting at. I think this is a feature request? Can you be more specific please?

dmke:
> I think this is a feature request?

Yes, indeed.

> Can you be more specific please?

[...]

PR #4021, opened 2021-02-16
reverseproxy: Add `handle_response` blocks to `reverse_proxy` (#3710)

### LINKED ISSUE #2920: Modify status code of reverse proxy response

francislavoie:
Supersedes #3712, closes #2920, #3707, #3828

Adds a new `handle_response` subdirective to `reverse_proxy`, and also `@<name>` named response matchers to pair with it.

Example which might serve files that are otherwise not accessible based on some response header:

```
	@accel header X-Accel-Redirect *
	handle_response @accel {
		root * /path/to/private/files
		rewrite {http.response.header.X-Accel-Redirect}
		file_server
	}
}
```

The matcher is optional; having no matcher will match all responses. Any `handle_response` without a matcher will be sorted to the end just in case the user put them in a bad order.

This also supports an optional status code on the same line, to change the status code of the response. If using this syntax, then a block may _not_ be defined, and the matcher is _required_. (Why would you ever want to 

```
	@fooIsBad header Foo *
	handle_response @fooIsBad 500
}
```

We should also remember to document that a status code matcher like `4` matches `4xx`, as per the `caddyhttp.StatusCodeMatches()` function. It's useful.


frou:
Sorry for what's maybe an elementary question, but can this be used to add some nuance to what Caddy does when the server being reverse proxied to is completely down?

Caddy currently serves a blank page with status code 502 (Bad Gateway) in that scenario. I'd like to be able to serve a page with a readable message, rather than just a blank page.

francislavoie:
@frou I'm pretty sure you can use https://caddyserver.com/docs/caddyfile/directives/handle_errors for that.

The difference is who emits the error. If it's Caddy itself that emits it, then you use `handle_errors`, and if the backend emits it as a response, then you use `handle_response`. A 502 is usually Caddy saying it couldn't connect to the upstream... But nothing stops an upstream from responding with a 502 if it feels like it.

amyspark:
👋  I've been sorely trying to get a shadow of Gitlab Pages/nginx's `error_page 404 /404.html` with a proxy. Will this MR let me preserve the status code, while replacing *only the response body* with the result of `handle_response`?

francislavoie:
Thanks for sparking the question @amyspark 😂

I realized that there was a piece lacking to make that work. The `file_server` handler would only write a `200` status code when writing the file. To make what you want work, I wrote this PR https://github.com/caddyserver/caddy/pull/4076 that should do it, i.e. something like this:

```
	@fourhundred status 4
	handle_response @fourhundred {
		root * /path/to/errors
		file_server {
			status {http.reverse_proxy.status_code}
		}
	}
}
```

amyspark:
Close enough! but what I need is something like this: 

```
	@fourhundred status 4
	handle_response @fourhundred {
		reverse_proxy localhost:8080 {http.reverse_proxy.status_code}
	}
}
```

(i.e. rewrite the petition, send it again upstream, but keep the returned status code from the original response's)

francislavoie:
In that case you'll just need to make sure your upstream writes the correct status. I don't think it's a good idea to modify `reverse_proxy` for that.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -102,6 +113,10 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	var transport http.RoundTripper
 	var transportModuleName string
 
+	// collect the response matchers defined as subdirectives prefixed with "@"
+	// for use with "handle_response" blocks
+	responseMatchers := map[string]caddyhttp.ResponseMatcher{}
```
mholt, line 118: This is a nit, to be sure, but I get a little triggered when I see a `map` without a `make()` around it. (Am I the only one? And yes I know both are perfectly valid.)

```suggestion
	responseMatchers := make(map[string]caddyhttp.ResponseMatcher)
```

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -892,8 +997,76 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+//     @name {
+//         header <field> [<value>]
+//         status <code...>
+//     }
+//
+// Or, single line syntax:
+//
+//     @name [header <field> [<value>]] | [status <code...>]
+//
+func (h *Handler) parseNamedResponseMatcher(d *caddyfile.Dispenser, matchers map[string]caddyhttp.ResponseMatcher) error {
```
mholt, line 1011: I might be missing something, but `h` is never used here. In fact, it doesn't necessarily have to be a method on `Handler` at all (except maybe for code organization).

```suggestion
func (Handler) parseNamedResponseMatcher(d *caddyfile.Dispenser, matchers map[string]caddyhttp.ResponseMatcher) error {
```

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -892,8 +997,76 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+		matchers[definitionName] = matcher
+	}
+	return nil
+}
+
+const matcherPrefix = "@"
+
 // Interface guards
 var (
-	_ caddyfile.Unmarshaler = (*Handler)(nil)
```
mholt, line 897: This does make me a little sad. I don't know if there's any way we can keep this...

mholt (changes requested):
Just little things, I think. Thanks for this contribution. Sorry it took me a while to devote the focus to review it.

I wonder if we could keep the `UnmarshalCaddyfile()` method if we keep the Caddyfile token parsing in there, but then leave any extra setup of the handler (e.g. the hanlde_response routes) to a caller, such as `parseCaddyfile` that takes the Helper. (I could elaborate in Slack...)

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: :man_shrugging: I have absolutely no preference

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: Yeah my plan is to move this elsewhere, combined with the code from the precompression PR (which is now merged) which also uses response matchers essentially the same way.

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: Yeah, same. I don't really see a way. We need `Helper` here, and not `Dispenser`, and `Helper` is a superset of `Dispenser`, but the unmarshaler uses `Dispenser` :cry:

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: Done, moved to a `responsematchers.go`, moved a few things around.

modules/caddyhttp/reverseproxy/caddyfile.go
mholt: So, I think there is a way to do this, by keeping `UnmarshalCaddyfile()` almost exactly the same, except that tokens for `handle_response` are collected and stored in the `Handler`. Then the `parseCaddyfile()` function, which calls `UnmarshalCaddyfile()`, could take those and finish parsing them out using its `Helper`.

Note that if we do remove this interface support, then any call to `UnmarshalModule` on a reverse proxy handler will break. I don't know that is currently done (it's not in our code base), but this is a breaking change otherwise.

francislavoie: Hmm, I'll give it a shot. That would mean that any call to `UnmarshalCaddyfile()` that doesn't come from something which has `Helper` will just have the `handle_response` stuff dropped silently though, but that's probably better than not having `reverse_proxy` work at all. I'll play a bit with it this evening.

francislavoie: Got it!!

mholt (approved):
_Excellent work!_ It turned out just about as I had hoped. Yes it's a bummer the complete Caddyfile unmarshaling can't take place inside `UnmarshalCaddyfile()`, but I think maintaining that interface is important, and the workaround is pretty reasonable.

It might be worth adding to the godoc comment for `UnmarshalCaddyfile()` that an additional call to `Finalize...` is needed when a Helper is available.

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -353,12 +353,14 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
 
 	// the rest of the config is specified by the user
 	// using the reverse_proxy directive syntax
-	// TODO: this can overwrite our fcgiTransport that we encoded and
-	// set on the rpHandler... even with a non-fastcgi transport!
```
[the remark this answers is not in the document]
francislavoie: Yeah, the comment was invalid actually, I misunderstood how it worked when I first wrote the `php_fastcgi` stuff. I tested this again when working on this PR and realized it wasn't actually a problem.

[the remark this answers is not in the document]
mholt: Oh good, that's great.

mholt:
Great work. Thanks for your patience and persistence with this.

francislavoie:
And thanks for @maxatome for doing the groundwork for this! 🎉

PR #4022, opened 2021-02-16
caddyfile: reject cyclic imports

Mohammed90:
Fix an edge case of Caddyfile import itself. I don't imagine anyone deliberately self-importing a Caddyfile but could happen if glob expansion wasn't vetted by the user.

### DIFF
--- a/caddyconfig/caddyfile/lexer.go
+++ b/caddyconfig/caddyfile/lexer.go
@@ -38,8 +38,8 @@ type (
 		File        string
 		Line        int
 		Text        string
-		InSnippet   bool
-		SnippetName string
+		inSnippet   bool
+		snippetName string
 	}
 )
 
--- a/caddyconfig/caddyfile/parse.go
+++ b/caddyconfig/caddyfile/parse.go
@@ -173,10 +173,13 @@ func (p *parser) begin() error {
 		if err != nil {
 			return err
 		}
-		// Keep track of which tokens come from snippets and what the snippet name is
+		// Just as we need to track which file the token comes from, we need to
+		// keep track of which snippets do the tokens come from. This is helpful
+		// in tracking import cycles across files/snippets by namespacing them. Without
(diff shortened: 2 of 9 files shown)


Mohammed90:
Converting it to draft as @francislavoie and I found more corner cases

Mohammed90:
I opted for minimal implementation of the graph to avoid pulling an additional dep just for tracking cycles.

caddyconfig/caddyfile/lexer.go
```diff
@@ -35,9 +35,11 @@ type (
...
 	// Token represents a single parsable unit.
 	Token struct {
-		File string
-		Line int
-		Text string
+		File        string
+		Line        int
+		Text        string
+		InSnippet   bool
+		SnippetName string
```
mholt, line 42: It looks like these are used only in this package, right? Let's unexport them.

caddyconfig/caddyfile/parse.go
```diff
@@ -165,6 +173,12 @@ func (p *parser) begin() error {
 		if err != nil {
 			return err
 		}
+		// Keep track of which tokens come from snippets and what the snippet name is
```
mholt, line 176: Maybe this comment should explain why; right now it doesn't add too much to what is already obvious from the lines below.

mholt (changes requested):
Thanks for contributing this. Might save some people some head scratching.

I have just a couple requests, nothing major.

caddyconfig/caddyfile/lexer.go
Mohammed90: Done

caddyconfig/caddyfile/parse.go
Mohammed90: Done

mholt (approved):
Cool, hopefully this will help a few people and prevent some problems! Thanks for contributing this @Mohammed90 !

Issue #4017, opened 2021-02-13
Bug in Caddy HTTP/2 support


rayjlinden:
Please see this wiki thread: https://caddy.community/t/caddy-with-reverse-proxy-returns-unexpected-403/11477

In HTTP/2 some browsers like Firefox and Chrome implement connection sharing with http/2.  So the browser rather than opening multiple connections for multiple requests may open one connection and send multiple requests through to it.

However, if you are proving multiple sites on the same proxy that a connection may come to the proxy server that shouldn't.

In my case, I had multiple sites that all shared a wildcard cert.  A request to site A is opened and a page is read.  It is then requesting a JS blog from a different site B.  Both are served from the same Proxy server but have different virtual names with the the same wildcard cert.

The problem is Caddy in the request the tls.server_name does not match the virtual host name.  (The tls connection is reused.). Caddy in such a case returns a 403.  However, it should instead respond with a 421.  This is part of the HTTP/2 spec and when the browser gets the 421 it simply opens a new connection and everything would fine.

However, with Caddy returning a 403 in such cases - the Browser thinks it just a normal 403 and shit breaks.  In very weird ways...

francislavoie:
How can this be reproduced minimally? What's the smallest possible config you can make it happen with, and I suppose along with an HTML file that triggers the requests that exhibit the behaviour?

rayjlinden:
I really am not sure how.  I only see it in our test suite where it does happen quite consistently. 

The test is loading a page (via selenium and headless chrome) - that page is loading some JS from a different site (that in our test set up has the same ip as the page).  That's when it is happening.  However, if I load the page in Chrome (same version) directly - it works fine.  I really can not explain that...  (Though as I'm writing this - I'm realizing that the test suite runs on the same server, where when I do it manually I'm coming over VPN...)

The site the page is coming from and the site the JS widget come from are using the same IP address.  And both using the same wildcard cert.  However, the domain names are different.

Lastly, as I understand it - this can only happen under Chrome or Firefox.  (Safari and Edge do not do connection coalescing.)

rayjlinden:
I think the fix is that this line needs to change to return status 421 instead of 403
https://github.com/caddyserver/caddy/blob/792fca40f18b7c528b00a7dea508bdfd0821dd8c/modules/caddyhttp/server.go#L292

[...]

PR #4023, opened 2021-02-16
Fix to HTTP/2.- and Chrome coalescing issue

rayjlinden:
Made change to return status code 421 instead of 403 when StrictSNIHost matching is on.

This is related to:
https://github.com/caddyserver/caddy/issues/4017

I'm not suggesting this is the complete fix.  I just know it fixes a critical issue with Chrome (and presumable Firefox).


mholt:
Update: I'm still not sure whether this is the right place to make this change, or if the standard lib should implement this instead since only it knows if the connection is reused.

francislavoie:
We'll close this for now, until we settle on a solution in #4017.

mholt (approved):
With no movement on this from standards body, I guess we will try it out. Thank you!

rayjlinden:
Wow!  Almost forgot about this... :)

PR #4024, opened 2021-02-17
ci: Build and test on Go 1.16, bump minimum to 1.15

francislavoie:
🎉

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        go: [ '1.14', '1.15' ]
+        go: [ '1.14', '1.15', '1.16' ]
 
         # Set some variables per OS, usable via ${{ matrix.VAR }}
         # CADDY_BIN_PATH: the path to the compiled Caddy binary, for artifact publishing
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        go: [ '1.14', '1.15', '1.16' ]
+        go: [ '1.15', '1.16' ]
 
         # Set some variables per OS, usable via ${{ matrix.VAR }}
         # CADDY_BIN_PATH: the path to the compiled Caddy binary, for artifact publishing
--- a/.github/workflows/cross-build.yml
+++ b/.github/workflows/cross-build.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         goos: ['android', 'linux', 'solaris', 'illumos', 'dragonfly', 'freebsd', 'openbsd', 'plan9', 'windows', 'darwin', 'netbsd']
-        go: [ '1.14', '1.15' ]
+        go: [ '1.14', '1.15', '1.16' ]
     runs-on: ubuntu-latest
     continue-on-error: true
     steps:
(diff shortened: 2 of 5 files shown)


.goreleaser.yml
```diff
@@ -37,8 +37,6 @@ builds:
   - 6
   - 7
   ignore:
-    - goos: darwin
-      goarch: arm
```
hairyhenderson, line 41: I think we still need to ignore this - `darwin/arm` is 32-bit ARM, and it was dropped in Go 1.15, whereas what we want is `darwin/arm64` (which wasn't ignored before, but it used to refer to the iOS port).

francislavoie: Ah, okay thanks. I wonder if we need to do anything special to get `arm64` to build or if it'll just work now that we build on 1.16. 🤔

francislavoie: I think it should _just work_ https://github.com/goreleaser/goreleaser/pull/1956

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        go: [ '1.14', '1.15' ]
+        go: [ '1.14', '1.15', '1.16' ]
```
Mohammed90, line 20:
```suggestion
        go: [ '1.15', '1.16' ]
```

.github/workflows/cross-build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         goos: ['android', 'linux', 'solaris', 'illumos', 'dragonfly', 'freebsd', 'openbsd', 'plan9', 'windows', 'darwin', 'netbsd']
-        go: [ '1.14', '1.15' ]
+        go: [ '1.14', '1.15', '1.16' ]
```
Mohammed90, line 17:
```suggestion
        go: [ '1.15', '1.16' ]
```

Mohammed90 (changes requested):
The Go version in go.mod needs to be bumped up to 1.15. The commit ab80ff4fd2911afc394b9dbceeb9f71c7a0b7ec1 (admin: remote management) uses the PublicKey#Equal method, which was added in go1.15.

https://github.com/caddyserver/caddy/blob/ec3ac840cf09ec43b9130b41a46c07827b2ff0b6/admin.go#L552-L553

Which means we need to drop `'1.14'` in the CI flows as well.

.github/workflows/ci.yml
francislavoie: Thanks for checking on that. Any idea why it still compiles on 1.14 though? :thinking: shouldn't it fail to compile if we use new stuff like that? Maybe I misunderstand how the Go compiler works.

Mohammed90: Ha, good question. Now I see the trick. In go1.14, the value of `ok` will be `false`:

```go
comparer, ok := peerCert.PublicKey.(interface{ Equal(crypto.PublicKey) bool }) 
 if !ok || !comparer.Equal(allowedKey) { 
    continue
```

This means in go1.14, the loop will hit `continue` at every iteration, and the entire func ends up returning APIError:

https://github.com/caddyserver/caddy/blob/ec3ac840cf09ec43b9130b41a46c07827b2ff0b6/admin.go#L597-L603

It means users lose an entire feature by compiling with go1.14. The type-assertion is fail-safe so it **will not panic** if compiled with < go1.15, but I'd expect confused users hitting the forum.

francislavoie: Okay @mholt are we pulling the trigger on the bump to `go1.15` then?

mholt: Ah yep, I meant to say this, sorry. We need to bump the minimum to Go 1.15 for the 2.4 tree.

francislavoie:
Oops - @mholt you'll need to turn off the requirement in the repo settings that the `test (macos-latest, 1.14)` and such pass.

mholt:
Thanks, good point. Done!

francislavoie:
:+1: 

I'm thinking this is good to?

mholt (approved):

Mohammed90 (approved):

PR #4028, opened 2021-02-18
Allow configuration of loggers through Caddyfile global options

kujenga:
This PR is a follow up on the previous attempt in https://github.com/caddyserver/caddy/pull/3968 but instead implemented as a global logging option.

It follows up on discussion in https://github.com/caddyserver/caddy/issues/3958 about adding more functionality to the Caddyfile. I also added in another commit for a small new "replace" filter aimed at the same goal as I wrote up in the original ticket, providing more tools to remove sensitive data from logs.

Closes https://github.com/caddyserver/caddy/issues/3958

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -619,11 +619,50 @@ func parseHandleErrors(h Helper) ([]ConfigValue, error) {
 //     }
 //
 func parseLog(h Helper) ([]ConfigValue, error) {
+	return parseLogHelper(h, nil)
+}
+
+// parseLogHelper is used both for the parseLog directive within Server Blocks,
+// as well as the global "log" option for configuring loggers at the global
+// level. The parseAsGlobalOption parameter is used to distinguish any differing logic
+// between the two.
+func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue, error) {
+	// When the globalLogNames parameter is passed in, we make
+	// modifications to the parsing behavior.
+	parseAsGlobalOption := globalLogNames != nil
+
 	var configValues []ConfigValue
 	for h.Next() {
-		// log does not currently support any arguments
-		if h.NextArg() {
-			return nil, h.ArgErr()
+		// Logic below expects that a name is always present when a
+		// global option is being parsed.
+		var globalLogName string
+		if parseAsGlobalOption {
+			if h.NextArg() {
+				globalLogName = h.Val()
+
+				// Only a single argument is supported.
+				if h.NextArg() {
+					return nil, h.ArgErr()
+				}
+			} else {
+				// If there is no log name specified, we
+				// reference the default logger. See the
+				// setupNewDefault function in the logging
+				// package for where this is configured.
+				globalLogName = "default"
+			}
+
+			// Verify this name is unused.
+			_, used := globalLogNames[globalLogName]
+			if used {
+				return nil, h.Err("duplicate global log option for: " + globalLogName)
+			}
+			globalLogNames[globalLogName] = struct{}{}
+		} else {
+			// No arguments are supported for the server block log directive
+			if h.NextArg() {
+				return nil, h.ArgErr()
+			}
 		}
 
 		cl := new(caddy.CustomLog)
@@ -687,22 +726,48 @@ func parseLog(h Helper) ([]ConfigValue, error) {
 					return nil, h.ArgErr()
 				}
 
+			case "include":
+				// This configuration is only allowed in the global options
+				if !parseAsGlobalOption {
+					return nil, h.ArgErr()
+				}
+				for h.NextArg() {
+					cl.Include = append(cl.Include, h.Val())
+				}
+
+			case "exclude":
+				// This configuration is only allowed in the global options
+				if !parseAsGlobalOption {
+					return nil, h.ArgErr()
+				}
+				for h.NextArg() {
+					cl.Exclude = append(cl.Exclude, h.Val())
+				}
+
 			default:
 				return nil, h.Errf("unrecognized subdirective: %s", h.Val())
 			}
 		}
 
 		var val namedCustomLog
+		// Skip handling of empty logging configs
 		if !reflect.DeepEqual(cl, new(caddy.CustomLog)) {
-			logCounter, ok := h.State["logCounter"].(int)
-			if !ok {
-				logCounter = 0
+			if parseAsGlobalOption {
+				// Use indicated name for global log options
+				val.name = globalLogName
+				val.log = cl
+			} else {
+				// Construct a log name for server log streams
+				logCounter, ok := h.State["logCounter"].(int)
+				if !ok {
+					logCounter = 0
+				}
+				val.name = fmt.Sprintf("log%d", logCounter)
+				cl.Include = []string{"http.log.access." + val.name}
+				val.log = cl
+				logCounter++
+				h.State["logCounter"] = logCounter
 			}
-			val.name = fmt.Sprintf("log%d", logCounter)
-			cl.Include = []string{"http.log.access." + val.name}
-			val.log = cl
-			logCounter++
-			h.State["logCounter"] = logCounter
 		}
 		configValues = append(configValues, ConfigValue{
 			Class: "custom_log",
--- a/caddyconfig/httpcaddyfile/options_test.go
+++ b/caddyconfig/httpcaddyfile/options_test.go
@@ -1,0 +1,64 @@
+package httpcaddyfile
+
+import (
+	"testing"
+
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	_ "github.com/caddyserver/caddy/v2/modules/logging"
+)
+
+func TestGlobalLogOptionSyntax(t *testing.T) {
+	for i, tc := range []struct {
+		input       string
+		output      string
+		expectError bool
+	}{
+		// NOTE: Additional test cases of successful Caddyfile parsing
+		// are present in: caddytest/integration/caddyfile_adapt/
+		{
+			input: `{
+				log default
+			}
+			`,
+			output:      `{}`,
+			expectError: false,
+		},
+		{
+			input: `{
+				log example {
+					output file foo.log
+				}
+				log example {
+					format json
+				}
+			}
+			`,
+			expectError: true,
+		},
+		{
+			input: `{
+				log example /foo {
+					output file foo.log
+				}
+			}
+			`,
+			expectError: true,
+		},
+	} {
+
+		adapter := caddyfile.Adapter{
+			ServerType: ServerType{},
+		}
+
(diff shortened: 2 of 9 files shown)


caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -687,22 +723,46 @@ func parseLog(h Helper) ([]ConfigValue, error) {
...
+					cl.Include = append(cl.Include, h.Val())
+				}
+
+			case "exclude":
+				if !h.NextArg() {
+					return nil, h.ArgErr()
+				}
+				for h.NextArg() {
+					cl.Exclude = append(cl.Exclude, h.Val())
+				}
```
francislavoie, line 740: I don't think we want to add `include` and `exclude` to the `log` directive, but we would want it for the global option. So that would probably need to be wrapped with that boolean.

caddyconfig/httpcaddyfile/options_test.go
```diff
@@ -0,0 +1,92 @@
...
+				log {
+					output file foo.log
+				}
+				log default {
+					format json
+				}
+			}
+			`,
+			output:      `{"logging":{"logs":{"default":{"encoder":{"format":"json"}}}}}`,
+			expectError: false,
```
francislavoie, line 64: I'm confused about this one - shouldn't this be an error? I don't think it's valid to have a `log` without a name in globals.

modules/logging/filters.go
```diff
@@ -25,6 +25,7 @@ import (
 
 func init() {
 	caddy.RegisterModule(DeleteFilter{})
+	caddy.RegisterModule(ReplaceFilter{})
```
francislavoie, line 28: Let's hold off on this for now, we can do that in a separate PR. Keep this PR focused on just the global option, please!

caddyconfig/httpcaddyfile/builtins.go
kujenga: Makes sense, that was one of the things I was wondering about, I can disable these for the global case.

modules/logging/filters.go
kujenga: Sure, happy to do that 👍

caddyconfig/httpcaddyfile/options_test.go
kujenga: I was going off your comment here about how the name could be optional: https://github.com/caddyserver/caddy/issues/3958#issuecomment-756776300

This is a bit of a strange case because the second block overwrites the first one.

I like the idea of making it name required field, explicit in all cases for the global option. I'll make that change.

caddyconfig/httpcaddyfile/builtins.go
francislavoie: To be clear - we want to _allow_ `include` and `exclude` for globals, but _disallow_ it for the `log` directive. The `log` directive's purpose is to configure access logs, so the `include` part of it is already "managed".

caddyconfig/httpcaddyfile/options_test.go
francislavoie: Oh I see, well in the case yeah I think it's okay that omitting the name uses `default`, but it should be an error if the same name is used (or implied) more than once. Overriding is non-apparent behaviour and is almost always going to be misconfiguration.

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -502,7 +510,7 @@ type (
 	// tokens from a global option. It is passed the tokens to parse and
 	// existing value from the previous instance of this global option
 	// (if any). It returns the value to associate with this global option.
-	UnmarshalGlobalFunc func(d *caddyfile.Dispenser, existingVal interface{}) (interface{}, error)
+	UnmarshalGlobalFunc func(h GlobalHelper, existingVal interface{}) (interface{}, error)
```
francislavoie, line 513: I understand the purpose of this change, but this is technically a breaking change. This is meant to be available to third party plugins so they can register global options of their own.

That said, I don't think any actually do yet because support for it was incomplete (see https://github.com/caddyserver/caddy/issues/3634) but that was resolved in https://github.com/caddyserver/caddy/pull/3990 which will be released in v2.4.0.

So if we need to make this change, we'll have to make the change before the v2.4.0 release to avoid future breakages.

I'm thinking this refactor of `UnmarshalGlobalFunc` might need to also be moved into a separate PR and merged ahead of this one, but I'll let @mholt chime in first.

kujenga: Ah! Great point.

Without this or something like it, it might be a bit hacky to do the check for duplicate/colliding log configurations, and it would be something that had to be in Caddy itself, i.e. not possible as part of the public API. FWIW, this seems like it may be a beneficial construct to have going forward to make the API more easily extensible in a non-breaking way, like directives work today.

caddyconfig/httpcaddyfile/options_test.go
kujenga: Got it, makes sense to me, I'll add back the allowance for making it optional.

caddyconfig/httpcaddyfile/directives.go
kujenga: Hey @mholt @francislavoie , just wanted to check in here, any further thoughts on whether or not this is an acceptable change to make?

If there is an issue with this interface shift let me know and I take another look here.

mholt: It hasn't been released yet, so we can technically change it, though that's not really desirable.

I don't _love_ the `GlobalHelper` state (we do already have something similar in `Helper`: https://pkg.go.dev/github.com/caddyserver/caddy/v2/caddyconfig/httpcaddyfile#Helper)

If the only goal is to pass in the state, then maybe that could/should just be another argument to `UnmarsahlGlobalFunc` functions.

kujenga: Looking at this again, I'm actually able to do what's needed without the state involved by using a healthy dose of type assertions to look at the "existing values" passed into the UnmarshalGlobalFunc. I've pushed commits for that now that I can squash down if the result looks amenable.

kujenga: I went ahead and rebased on master and squashed things down

kujenga:
Hi @mholt @francislavoie , just checking in, any thoughts on the simplified approach this PR now takes? Happy to make any further iterations as wanted.

francislavoie (approved):
Looks good to me! Thanks for the effort! :tada:

caddyconfig/httpcaddyfile/options_test.go
```diff
@@ -0,0 +1,114 @@
+package httpcaddyfile
+
+import (
+	"testing"
+
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	_ "github.com/caddyserver/caddy/v2/modules/logging"
+)
+
+func TestGlobalLogOptionSyntax(t *testing.T) {
```
francislavoie, line 10: This is testing the global options in isolation, which is good, but I think we should also test global logs in combination with logs configured in sites.

An easy way to do that is by writing an adapt test, you can find the ones we have already here: https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt

You could maybe move all of the tests in this file to be adapt test files instead, they're a bit easier to read and update, and render a diff of the JSON on error.

francislavoie (changes requested):
Actually, noticed a couple more things on a quick second pass, sorry :sweat_smile:

caddyconfig/httpcaddyfile/options_test.go
kujenga: Great, thanks for pointing that out. I ended up splitting the tests so that the more complex successful test cases are in the adapt folder, and the failure cases remain unit tests, since it seems like the adapt folder does not support that? Let me know if this matches what you'd expect.

francislavoie: Yeah that's perfect, thanks

kujenga:
Alright, this is rebased on master now as well, if there's anything else just let me know!

caddyconfig/httpcaddyfile/options_test.go
francislavoie: I think we're still missing an adapt test that has both a global logger and a logger in a site. Something like this:

```
{
	log {
		output file caddy.log
	}
}

:8884 {
	log {
		output file access.log
	}
}
```

kujenga: Great idea, and it turns out that caught a subtle edge-case bug the default logger would end up with logs configured for it excluded by this logic: https://github.com/caddyserver/caddy/blob/51f35ba03f78e691d0e141b24ff19f82bfb34b67/caddyconfig/httpcaddyfile/httptype.go#L313-L322

That case is now handled better and a new test case file covers it.

francislavoie: Perfect, thank you!

francislavoie (approved):

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -619,11 +619,46 @@ func parseHandleErrors(h Helper) ([]ConfigValue, error) {
...
 //
 func parseLog(h Helper) ([]ConfigValue, error) {
+	return parseLogHelper(h, false, nil)
+}
+
+// parseLogHelper is used both for the parseLog directive within Server Blocks,
+// as well as the global "log" option for configuring loggers at the global
+// level. The globalOption parameter is used to distinguish any differing logic
+// between the two.
+func parseLogHelper(h Helper, globalOption bool, globalLogNames map[string]struct{}) ([]ConfigValue, error) {
```
mholt, line 629: Will `globalOption` ever be `true` while `globalLogNames` is nil? If not, we can drop the `globalOption` param and just assume that if `globalLogNames` is non-nil, `globalOption` must be true.

mholt (approved):
Awesome, thanks for your reviews Francis and for the work on this, @kujenga -- LGTM, just one comment where a slight improvement could be made if I'm right. Ignore if not. :)

caddyconfig/httpcaddyfile/builtins.go
kujenga: Good point! I think that was an artifact of an earlier iteration, I collapsed the parameters and added an internal variable to keep the logic explicit.

kujenga:
Alright, this has been rebased on master and I believe the last comment is now addressed!

francislavoie (approved):

mholt (approved):
Excellent, thanks for the contribution @kujenga ! Please feel free to contribute again sometime. :) Especially if any issues with these changes are reported later. Thank you!

kujenga:
Awesome, thank you! Feel free to ping me if anything does come up 👍

francislavoie:
@kujenga would you like to take a shot at writing the docs for this? You can find them here https://github.com/caddyserver/website/blob/master/src/docs/markdown/caddyfile/options.md

PR #4029, opened 2021-02-18
logging: add replace filter for static value replacement

kujenga:
This filter is intended to be useful in scenarios where you may want to
redact a value with a static string, giving you information that the
field did previously exist and was present, but not revealing the value
itself in the logs.

This was inspired by work on adding more complete support for removing
sensitive values from logs [1]. An example use case would be the
Authorization header in request log output, for which the value should
usually not be logged, but it may be quite useful for debugging to
confirm that the header was present in the request.

This PR was split out from https://github.com/caddyserver/caddy/pull/4028

[1] https://github.com/caddyserver/caddy/issues/3958

### DIFF
--- a/caddytest/integration/caddyfile_adapt/log_filters.txt
+++ b/caddytest/integration/caddyfile_adapt/log_filters.txt
@@ -5,7 +5,7 @@ log {
 	format filter {
 		wrap console
 		fields {
-			request>headers>Authorization delete
+			request>headers>Authorization replace REDACTED
 			request>headers>Server delete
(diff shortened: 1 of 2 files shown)


kujenga:
@francislavoie @mholt let me know if there are any thoughts on this PR I can address, I had originally paired it with https://github.com/caddyserver/caddy/pull/4028

francislavoie (approved):
I'm happy with this! Definitely a valid usecase, thanks for the explanation.

mholt (approved):
LGTM, thanks again very much!

PR #4033, opened 2021-02-22
caddyhttp: Implement better logic for inserting the HTTP->HTTPS redirs

francislavoie:
Taking a crack at implementing what I suggested in https://github.com/caddyserver/caddy/issues/3212

This is WIP, I haven't tested it yet, but wouldn't mind your 2c about this idea @mholt before I dive any deeper on the idea.

Basically, this should insert the redirect routes after any existing route with a host matcher but before any possibly existing user-defined catch-all route, then append our own catch-all at the end just in case as we already did before. I think this would better align with what users expect when they write a Caddyfile that has a `http://` route. This behaviour can be opted out of by setting `auto_https disable_redirects`.

### DIFF
--- a/caddytest/integration/autohttps_test.go
+++ b/caddytest/integration/autohttps_test.go
@@ -80,3 +80,26 @@ func TestAutoHTTPRedirectsWithHTTPListenerFirstInAddresses(t *testing.T) {
 `, "json")
 	tester.AssertRedirect("http://localhost:9080/", "https://localhost/", http.StatusPermanentRedirect)
 }
+
+func TestAutoHTTPRedirectsInsertedBeforeUserDefinedCatchAll(t *testing.T) {
+	tester := caddytest.NewTester(t)
+	tester.InitServer(`
+	{
+		http_port     9080
+		https_port    9443
+		local_certs
+	}
+	http://:9080 {
+		respond "Foo"
+	}
+	http://baz.localhost:9080 {
+		respond "Baz"
+	}
+	bar.localhost {
+		respond "Bar"
+	}
+  `, "caddyfile")
+	tester.AssertRedirect("http://bar.localhost:9080/", "https://bar.localhost/", http.StatusPermanentRedirect)
+	tester.AssertGetResponse("http://foo.localhost:9080/", 200, "Foo")
+	tester.AssertGetResponse("http://baz.localhost:9080/", 200, "Baz")
+}
--- a/modules/caddyhttp/autohttps.go
+++ b/modules/caddyhttp/autohttps.go
@@ -342,21 +342,22 @@ redirServersLoop:
 	for redirServerAddr, routes := range redirServers {
 		// for each redirect listener, see if there's already a
 		// server configured to listen on that exact address; if so,
-		// simply add the redirect route to the end of its route
-		// list; otherwise, we'll create a new server for all the
-		// listener addresses that are unused and serve the
-		// remaining redirects from it
-		for srvName, srv := range app.Servers {
+		// insert the redirect route to the end of its route list
+		// after any other routes with host matchers; otherwise,
+		// we'll create a new server for all the listener addresses
+		// that are unused and serve the remaining redirects from it
+		for _, srv := range app.Servers {
 			if srv.hasListenerAddress(redirServerAddr) {
-				// user has configured a server for the same address
-				// that the redirect runs from; simply append our
-				// redirect route to the existing routes, with a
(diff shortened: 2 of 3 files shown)


francislavoie:
Okay, I've confirmed this works.

Caddyfile:
```
{
        local_certs
}

http:// {
        respond "Foo"
}

http://baz.localhost {
        respond "Baz"
}

bar.localhost {
        respond "Bar"
}
```

Try to hit the domains on HTTP (some curl output omitted for length):
```
$ curl -v --resolve 'bar.localhost:80:127.0.0.1' http://bar.localhost/
< HTTP/1.1 308 Permanent Redirect
< Location: https://bar.localhost/
< Server: Caddy

$ curl -v --resolve 'baz.localhost:80:127.0.0.1' http://baz.localhost/
< HTTP/1.1 200 OK
< Server: Caddy
< Content-Length: 3
<
Baz

$ curl -v --resolve 'foo.localhost:80:127.0.0.1' http://foo.localhost/
< HTTP/1.1 200 OK
< Server: Caddy
< Content-Length: 3
<
Foo
```

francislavoie:
Shoutout to @Mohammed90 for helping me with the integration test

roobre:
Hi, just wanted to say thanks for working on this! My use case involving a caddy server reachable both via a FQDN and local-only names is also affected by #3212, so I'm very glad to see some work towards fixing it.

I'm trying to run a custom build to test this patch but unfortunately I'm not being able to make it work. When I have a `Caddyfile` like this:
```
http:// {
        file_server /usr/share/caddy
}
```
I can `curl` a local domain and get an HTTP response:
```
> $ curl gigabox -v                                                                        [±better-auto-http-redir ✓]
*   Trying 10.0.0.5:80...
* Connected to gigabox (10.0.0.5) port 80 (#0)
> GET / HTTP/1.1
> Host: gigabox
> User-Agent: curl/7.76.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: Caddy
< Date: Tue, 06 Apr 2021 12:29:09 GMT
< Content-Length: 0
< 
* Connection #0 to host gigabox left intact
```

However, when I add a host with autotls:
```
http:// {
        file_server /usr/share/caddy
}

terabox.moe {
        file_server /usr/share/caddy
}
```
It  tries to redirect me to a non-existing https host
```
> $ curl gigabox -v                                                                        [±better-auto-http-redir ✓]
*   Trying 10.0.0.5:80...
* Connected to gigabox (10.0.0.5) port 80 (#0)
> GET / HTTP/1.1
> Host: gigabox
> User-Agent: curl/7.76.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 308 Permanent Redirect
< Connection: close
< Location: https://gigabox/
< Server: Caddy
< Date: Tue, 06 Apr 2021 12:32:41 GMT
< Content-Length: 0
< 
* Closing connection 0
```

If I'm not mistaken, I should be running a binary built from 139b9f78565de9ffda10e3bc1e391607d90f7454 using `go build` from `caddy/cmd/caddy`. Am I doing something wrong?

Thanks!

roobre:
Hello again,

After some experimentation, I have traced this behavior to the existance of a non-catch-all http host in the caddyfile.

This does not work:
```
http:// {
        file_server /usr/share/caddy
}

terabox.moe {
        file_server /usr/share/caddy
}
```

But this does:
```
http:// {
        file_server /usr/share/caddy
}

http://baz.localhost {
        respond "Baz"
}

terabox.moe {
        file_server /usr/share/caddy
}
```

Could this be caused by `findLastRouteWithHostMatcher` not returning this http-only wildcard?

francislavoie:
I think this is a bit off-topic, because you have unrelated problems in your Caddyfile.

Specifically, `file_server /usr/share/caddy` doesn't really make sense; you need to use the `root` directive to set the site root. What you have instead sets a request matcher for the path `/usr/share/caddy`, and the `root` defaults to the current-working-directory used to run Caddy.

Please ask your usage questions on the [Caddy community forums](https://caddy.community). Don't forget to fill out the thread template so we can help you!

mholt (dismissed):
Okie, this is a minimally invasive change, and appears to have more intuitive results. I don't really know how it will work in practice, but I guess we should try it, and if it causes more problems than it solves, we'll have to remove it (or fix it, if the fix doesn't add considerable complexity).

Thank you for tackling this, I really didn't want to. :P

mholt (approved):

Issue #3782, opened 2020-10-07
httpcaddyfile: respond with error page from handle_errors


segevfiner:
It looks like it is not currently possible to directly [`respond`](https://caddyserver.com/docs/caddyfile/directives/respond) with an error page defined by [`handle_errors`](https://caddyserver.com/docs/caddyfile/directives/handle_errors) (Like the Nginx [`return`](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#return).

An example use case where this could be useful:
```caddyfile
respond /subpath/ 404 # <- I want to return an error page defined in handle_errors
rewrite /subpath/* /subpath/index.html
```

francislavoie:
Hi! Thanks for trying Caddy!

Are the examples in https://caddyserver.com/docs/caddyfile/directives/handle_errors#examples not sufficient?

Please ask your usage questions on the [Caddy community forums](https://caddy.community). We prefer to keep the GitHub issue board for bugs and feature requests. Don't forget to fill out the thread template so we can help you!

segevfiner:
@francislavoie 

They are sufficient. I'm talking about something else, returning an explicit error from some other route so as to trigger the registered error handlers.

Looking into the JSON config, I think the [`error`](https://caddyserver.com/docs/json/apps/http/servers/errors/routes/handle/error/) handler achieves that. But it's yet another feature not available in the Caddyfile.

francislavoie:
Ah, okay.

francislavoie:
Opened PR https://github.com/caddyserver/caddy/pull/4034 for this.

PR #4034, opened 2021-02-22
httpcaddyfile: Add `error` directive for the existing handler

francislavoie:
Closes https://github.com/caddyserver/caddy/issues/3782

Just adds Caddyfile support for the static `error` handler that already exists: https://caddyserver.com/docs/json/apps/http/servers/routes/handle/error/

Syntax:
```
error [<matcher>] <status>|<message> [<status>] {
    message <text>
}
```

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -63,6 +63,7 @@ var directiveOrder = []string{
 	"push",
 
 	// handlers that typically respond to requests
+	"error",
 	"respond",
 	"metrics",
 	"reverse_proxy",
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -63,7 +63,6 @@ var directiveOrder = []string{
 	"push",
 
 	// handlers that typically respond to requests
-	"error",
 	"respond",
 	"metrics",
 	"reverse_proxy",
@@ -71,6 +70,7 @@ var directiveOrder = []string{
 	"file_server",
 	"acme_server",
 	"abort",
+	"error",
 }
 
 // directiveIsOrdered returns true if dir is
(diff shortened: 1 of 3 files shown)


lilbee101920 (approved):

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -63,6 +63,7 @@ var directiveOrder = []string{
 	"push",
 
 	// handlers that typically respond to requests
+	"error",
```
mholt, line 66: Is this at the optimal place? (Sometimes I wish I had reversed the positions of `file_server` and `reverse_proxy`.)

For example, will it be more common to use this directive as a "fallback" (i.e. no matchers) or as a way to handle a certain class of requests (i.e. with matchers)? If a fallback, should be at the end. If specific, it's probably in a reasonable place. (As to the exact placement, though, it's still an open question.)

mholt (dismissed):
Thanks, LGTM! I do have a question regarding the directive order but no answer or preference as to that.

caddyconfig/httpcaddyfile/directives.go
francislavoie: Honestly, I'm not sure, but I'm leaning toward people using matchers with `error` more often than not :man_shrugging:

mholt: Hmm yeah I'm not sure either. I feel like returning an error could be a common fallback scenario ("Oh, if we haven't handled it yet, then return an error instead of the default 200 OK"). Then again, I can also see how the opposite could be common: "If we haven't errored, then reverse_proxy / file_server." I kinda think the first may be more common. Should we move it to the end?

francislavoie: Sure why not :man_shrugging:

mholt (approved):
Alright, cool, let's give this a try and see if it's useful!

habibalamin:
This was one of the things keeping me on an older version of Caddy, so thank you for this.

francislavoie:
@habibalamin which version? This isn't something that existed in a previous version.

habibalamin:
Sorry, I was a bit unclear. There was a “rewrite hack” for the exact use case in the linked issue that you could do in v1 but not v2, and adding this directive gets me one step closer to letting me migrate to v2.

Issue #2665, opened 2019-07-11
v2: Content negotiation


mholt:
Caddy 2 needs a generic, powerful way to do content negotiation.

Currently, the `http.handlers.encode` module negotiates its encoding from the Accept-Encoding header. However, for pre-compressed assets or other kinds of content negotiation, different facilities are required.

Content negotiation can be based on MIME type (Accept header), encoding (Accept-Encoding header), or language (Accept-Language) header.

I propose a content negotiation request matcher:

```json
"content_negotiation": {
    "mime": ["text/plain", "text/html"],
    "encoding": ["br", "gzip", "flate"],
    "language": ["en-US"]
}
```

For each negotiated variance, a placeholder would be added indicating what was negotiated, for example: `{http.matchers.content_negotiation.mime}`

TODO: We would also need a way to add a Vary header to the response. Maybe a matcher is not the right solution. This might have to be a middleware... hm.

mholt:
As discussed in https://caddy.community/t/how-to-serve-gzipped-files-automatically-in-caddy-v2/7311?u=matt it would be useful for Caddy to be able to serve pre-compressed files as well. If enabled, this would involve using the Accept-Encoding header to see if there is a matching file on disk, and if so, serve it with the proper headers. Maybe as a guest module to the file_server handler.

averri:
What is the workaround for implementing this in the actual version v2? Suppose the following scenario:

The browser requests the resource `/js/app.js` with the header `accept-encoding: gzip`. If the resource exists with an extension gz, like `/js/app.js.gz`, so the HTTP response should include the compressed file  `/js/app.js.gz` with the appropriate response header set: `content-encoding: gzip`.

polarathene:
It'd be nice for Caddy to be able to support pre-compressed brotli `.br` files. 

Unlike gzip, brotli isn't advised for on-demand compression from what I've read, would it also be something Caddy could do if encoding was requested for static assets to compress them, but write that compressed file to disk for serving in future, or is this something that should be handled outside of Caddy? Such as a scheduled compression pass if compressible files are added to the server?(eg via user uploads)

mholt:
> It'd be nice for Caddy to be able to support pre-compressed brotli .br files.

I think so too (see my reply above)! It just needs to be discussed, designed, and implemented...

ueffel:
[...]

PR #4045, opened 2021-03-02
encode: Enhancements to content negotiation, serving precompressed content

ueffel:
This pull request partly addresses #2665

It adds:
* the ability to configure the `minium_length` of a response for encoding to happen via caddyfile
* the ability to configure encoding preference on the server-side (`prefer` setting for `encode`)
* a new interface `Precompress`, which modules can implement to provide the file extension suffix for the `file_server` to search for precompressed files

Differences to the previous iteration of the changes (https://github.com/caddyserver/caddy/compare/master...ueffel:file_server-precompressed):
https://github.com/ueffel/caddy/compare/file_server-precompressed...ueffel:file_server-precompressed_v2
* module namespace for precompress modules changed to `http.precompressed.*`
* new precompress modules for gzip, zstd, br
* caddyfile directive and interface name changed to Precompress**ed**

Edit:
Instead of introducing a new dependency to "github.com/gobwas/glob" now the content-types matching uses `filepath.Match`. Maybe the "validation" by matching with an empty string does not produce a `ErrBadPattern` for every bad patttern 🤔

### DIFF
--- a/modules/caddyhttp/encode/caddyfile.go
+++ b/modules/caddyhttp/encode/caddyfile.go
@@ -40,8 +40,9 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 // UnmarshalCaddyfile sets up the handler from Caddyfile tokens. Syntax:
 //
 //     encode [<matcher>] <formats...> {
-//         gzip [<level>]
+//         gzip           [<level>]
 //         zstd
+//         prefer         <formats...>
 //     }
 //
 // Specifying the formats on the first line will use those formats' defaults.
@@ -64,19 +65,31 @@ func (enc *Encode) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 		for d.NextBlock(0) {
 			name := d.Val()
-			modID := "http.encoders." + name
-			unm, err := caddyfile.UnmarshalModule(d, modID)
-			if err != nil {
-				return err
+			switch name {
+			case "prefer":
+				var encs []string
+				for d.NextArg() {
+					encs = append(encs, d.Val())
+				}
+				if len(encs) == 0 {
+					return d.ArgErr()
+				}
+				enc.Prefer = encs
+			default:
+				modID := "http.encoders." + name
+				unm, err := caddyfile.UnmarshalModule(d, modID)
+				if err != nil {
+					return err
+				}
+				encoding, ok := unm.(Encoding)
+				if !ok {
+					return fmt.Errorf("module %s is not an HTTP encoding; is %T", modID, unm)
+				}
+				if enc.EncodingsRaw == nil {
+					enc.EncodingsRaw = make(caddy.ModuleMap)
+				}
+				enc.EncodingsRaw[name] = caddyconfig.JSON(encoding, nil)
 			}
-			encoding, ok := unm.(Encoding)
-			if !ok {
-				return fmt.Errorf("module %s is not an HTTP encoding; is %T", modID, unm)
-			}
-			if enc.EncodingsRaw == nil {
-				enc.EncodingsRaw = make(caddy.ModuleMap)
-			}
-			enc.EncodingsRaw[name] = caddyconfig.JSON(encoding, nil)
 		}
 	}
 
--- a/modules/caddyhttp/encode/caddyfile.go
+++ b/modules/caddyhttp/encode/caddyfile.go
@@ -15,8 +15,9 @@
 package encode
 
 import (
-	"fmt"
+	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
@@ -45,20 +46,25 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 //         zstd
 //         minimum_length <length>
 //         prefer         <formats...>
-//         types          <glob-patterns...>
+//         match          <matcher>
+//         @name {
+//             status <code...>
+//             header <field> [<value>]
+//         }
 //     }
 //
 // Specifying the formats on the first line will use those formats' defaults.
 func (enc *Encode) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	responseMatchers := map[string]caddyhttp.ResponseMatcher{}
 	for d.Next() {
 		for _, arg := range d.RemainingArgs() {
 			mod, err := caddy.GetModule("http.encoders." + arg)
 			if err != nil {
-				return fmt.Errorf("finding encoder module '%s': %v", mod, err)
+				return d.Errf("finding encoder module '%s': %v", mod, err)
 			}
 			encoding, ok := mod.New().(Encoding)
 			if !ok {
-				return fmt.Errorf("module %s is not an HTTP encoding", mod)
+				return d.Errf("module %s is not an HTTP encoding", mod)
 			}
 			if enc.EncodingsRaw == nil {
 				enc.EncodingsRaw = make(caddy.ModuleMap)
@@ -68,6 +74,17 @@ func (enc *Encode) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 		for d.NextBlock(0) {
 			name := d.Val()
+
+			// if the subdirective has an "@" prefix then we
+			// parse it as a response matcher for use with "match"
+			if strings.HasPrefix(name, matcherPrefix) {
+				err := enc.parseNamedResponseMatcher(d.NewFromNextSegment(), responseMatchers)
+				if err != nil {
+					return err
+				}
+				continue
+			}
+
 			switch name {
 			case "minimum_length":
 				if !d.NextArg() {
@@ -87,15 +104,30 @@ func (enc *Encode) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 					return d.ArgErr()
 				}
 				enc.Prefer = encs
-			case "types":
-				var types []string
-				for d.NextArg() {
-					types = append(types, d.Val())
+			case "match":
+				var matcher *caddyhttp.ResponseMatcher
+				args := d.RemainingArgs()
+				// the first arg should be a matcher (optional)
+				// the second arg should be a status code (optional)
+				// any more than that isn't currently supported
+				if len(args) > 1 {
+					return d.Errf("too many arguments for 'if_matches': %s", args)
 				}
-				if len(types) == 0 {
-					return d.ArgErr()
+
+				// the first arg should always be a matcher.
+				// it doesn't really make sense to support status code without a matcher.
+				if len(args) > 0 {
+					if !strings.HasPrefix(args[0], matcherPrefix) {
+						return d.Errf("must use a named response matcher, starting with '@'")
+					}
+
+					foundMatcher, ok := responseMatchers[args[0]]
+					if !ok {
+						return d.Errf("no named response matcher defined with name '%s'", args[0][1:])
+					}
+					matcher = &foundMatcher
 				}
-				enc.Types = types
+				enc.Matcher = matcher
 			default:
 				modID := "http.precompressed." + name
 				unm, err := caddyfile.UnmarshalModule(d, modID)
@@ -104,7 +136,7 @@ func (enc *Encode) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				}
 				encoding, ok := unm.(Encoding)
 				if !ok {
-					return fmt.Errorf("module %s is not an HTTP encoding; is %T", modID, unm)
+					return d.Errf("module %s is not an HTTP encoding; is %T", modID, unm)
 				}
 				if enc.EncodingsRaw == nil {
 					enc.EncodingsRaw = make(caddy.ModuleMap)
@@ -116,6 +148,73 @@ func (enc *Encode) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 	return nil
 }
+
+// Parse the tokens of a named response matcher.
+//
+//     @name {
+//         header <field> [<value>]
+//         status <code...>
+//     }
+//
+// Or, single line syntax:
+//
+//     @name [header <field> [<value>]] | [status <code...>]
+//
+func (enc *Encode) parseNamedResponseMatcher(d *caddyfile.Dispenser, matchers map[string]caddyhttp.ResponseMatcher) error {
+	for d.Next() {
+		definitionName := d.Val()
+
+		if _, ok := matchers[definitionName]; ok {
+			return d.Errf("matcher is defined more than once: %s", definitionName)
+		}
+
+		matcher := caddyhttp.ResponseMatcher{}
+		for nesting := d.Nesting(); d.NextArg() || d.NextBlock(nesting); {
+			switch d.Val() {
+			case "header":
+				if matcher.Headers == nil {
+					matcher.Headers = http.Header{}
+				}
+
+				// reuse the header request matcher's unmarshaler
+				headerMatcher := caddyhttp.MatchHeader(matcher.Headers)
+				err := headerMatcher.UnmarshalCaddyfile(d.NewFromNextSegment())
+				if err != nil {
+					return err
+				}
+
+				matcher.Headers = http.Header(headerMatcher)
+			case "status":
+				if matcher.StatusCode == nil {
+					matcher.StatusCode = []int{}
+				}
+
+				args := d.RemainingArgs()
+				if len(args) == 0 {
+					return d.ArgErr()
+				}
+
+				for _, arg := range args {
+					if len(arg) == 3 && strings.HasSuffix(arg, "xx") {
+						arg = arg[:1]
+					}
+					statusNum, err := strconv.Atoi(arg)
+					if err != nil {
+						return d.Errf("bad status value '%s': %v", arg, err)
+					}
+					matcher.StatusCode = append(matcher.StatusCode, statusNum)
+				}
+			default:
+				return d.Errf("unrecognized response matcher %s", d.Val())
+			}
+		}
+
+		matchers[definitionName] = matcher
+	}
+	return nil
+}
+
+const matcherPrefix = "@"
 
 // Interface guard
 var _ caddyfile.Unmarshaler = (*Encode)(nil)
--- a/modules/caddyhttp/encode/caddyfile.go
+++ b/modules/caddyhttp/encode/caddyfile.go
@@ -46,16 +46,18 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 //         zstd
 //         minimum_length <length>
 //         prefer         <formats...>
-//         match          <matcher>
-//         @name {
+//         # response matcher block
+//         match {
 //             status <code...>
 //             header <field> [<value>]
 //         }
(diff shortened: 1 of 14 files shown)


modules/caddyhttp/encode/br/brotli_precompressed.go
```diff
@@ -0,0 +1,31 @@
...
+		New: func() caddy.Module { return new(BrotliPrecompressed) },
+	}
+}
+
+// AcceptEncoding returns the name of the encoding as
+// used in the Accept-Encoding request headers.
+func (BrotliPrecompressed) AcceptEncoding() string { return "br" }
+
+// Suffix returns the filename suffix of precomressed files
+func (BrotliPrecompressed) Suffix() string { return "br" }
```
mholt, line 28: Isn't the suffix actually `".br"`?

modules/caddyhttp/encode/encode.go
```diff
@@ -43,12 +45,15 @@ type Encode struct {
...
-	// If the client has no strong preference, choose this encoding. TODO: Not yet implemented
-	// Prefer    []string `json:"prefer,omitempty"`
+	// If the client has no strong preference, choose these encodings in order.
+	Prefer []string `json:"prefer,omitempty"`
 
 	// Only encode responses that are at least this many bytes long.
 	MinLength int `json:"minimum_length,omitempty"`
 
+	// Only encode responses that match at least one of the content-types.
+	Types []string `json:"types,omitempty"`
```
mholt, line 55: It occurred to me, would there be any value in letting the user configure a [`ResponseMatcher`](https://pkg.go.dev/github.com/caddyserver/caddy/v2@v2.3.0/modules/caddyhttp#ResponseMatcher) here? This could let them choose based on other headers (and status code) too, but I'm not sure how useful that is.

(To be clear, what you have is fine, but before we introduce a non-breakable change I want to get your thoughts on this. If the ResponseMatcher approach is feasible, that might be better since it gives more flexibility.)

modules/caddyhttp/encode/encode.go
```diff
@@ -75,11 +80,49 @@ func (enc *Encode) Provision(ctx caddy.Context) error {
 	if enc.MinLength == 0 {
 		enc.MinLength = defaultMinLength
 	}
+
+	if len(enc.Types) == 0 {
+		enc.Types = []string{"*"} // backwards compatibility
+
+		// sane default for text-based content ?
```
mholt, line 87: We like sane defaults

modules/caddyhttp/encode/br/brotli_precompressed.go
```diff
@@ -0,0 +1,31 @@
+package caddybr
```
mholt, line 1: Thinking this should be in a folder called `brotli` (it's not a long name), and maybe `caddybrotli`, but that's just my opinion.

modules/caddyhttp/encode/gzip/gzip_precompressed.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/modules/caddyhttp/encode"
+)
+
+func init() {
+	caddy.RegisterModule(GzipPrecompressed{})
+}
+
+// GzipPrecompressed provides the file extension for files precompressed with gzip encoding
+type GzipPrecompressed struct {
```
mholt, line 13: Maybe we should just reuse the `Gzip` type, i.e. put the methods on that type instead of making a new one.

Same for zstd.

modules/caddyhttp/fileserver/caddyfile.go
```diff
@@ -77,6 +81,26 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
...
+				for h.NextArg() {
+					modID := "http.precompressed." + h.Val()
+					mod, err := caddy.GetModule(modID)
+					if err != nil {
+						return nil, h.Errf("getting module named '%s': %v", modID, err)
+					}
+					inst := mod.New()
+					precompress, ok := inst.(encode.Precompressed)
+					if !ok {
+						return nil, fmt.Errorf("module %s is not an HTTP encoding; is %T", modID, inst)
```
mholt, line 95: Prefer `h.Errf()` here to get line information

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -79,6 +80,14 @@ type FileServer struct {
 	// a 404 error. By default, this is false (disabled).
 	PassThru bool `json:"pass_thru,omitempty"`
 
+	// Selection of encoders to use to check for precompressed files.
+	PrecompressRaw caddy.ModuleMap `json:"encodings,omitempty" caddy:"namespace=http.precompressed"`
```
mholt, line 84: I usually prefer that the json name matches the field name (except "Raw" suffixes). So like `json:"precompressed,omitempty" and a field name of `PrecompressedRaw` or something like that.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -79,6 +80,14 @@ type FileServer struct {
...
+
+	// If the client has no strong preference, choose these encodings in order.
+	PrecompressOrder []string
```
mholt, line 87: Maybe explain what the default is.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -230,18 +263,49 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
...
+	// check for precompressed files
+	for _, ae := range encode.AcceptedEncodings(r, fsrv.PrecompressOrder) {
+		precompress, ok := fsrv.precompressors[ae]
+		if !ok {
+			continue
+		}
+		compressedFilename := filename + "." + precompress.Suffix()
+		compressedInfo, err := os.Stat(compressedFilename)
+		if err != nil || compressedInfo.IsDir() {
+			fsrv.logger.Debug("precompressed file not accessible", zap.String("filename", compressedFilename))
```
mholt, line 277: I would include the error too: `zap.Error(err)`. Same below.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -230,18 +263,49 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
...
+			continue
+		}
+		fsrv.logger.Debug("opening file", zap.String("filename", compressedFilename))
```
mholt, line 280:
```suggestion
		fsrv.logger.Debug("opening compressed sidecar file", zap.String("filename", compressedFilename))
```

mholt (changes requested):
Awesome, thanks so much for working on this.

This is my first pass -- see what you think!

modules/caddyhttp/encode/encode.go
ueffel: I didn't know a `ResponseMatcher` exists. I will have a look.

modules/caddyhttp/encode/br/brotli_precompressed.go
ueffel: thought I changed that. It should be `caddybrotli`.

modules/caddyhttp/encode/gzip/gzip_precompressed.go
ueffel: So just `GzipPrecompressed` as alias for `Gzip` and register it under the `http.precompressed.*` namespace?

modules/caddyhttp/fileserver/staticfiles.go
ueffel: Again a thing I forgot to change along the namespace >.>

ueffel:
Would you like changes to be in additional commit or should I rebase-edit changes into the existing commits?

francislavoie:
We squash-merge at the end, so do whatever works best for you for now :+1:

modules/caddyhttp/encode/encode.go
francislavoie: https://github.com/caddyserver/caddy/pull/4021 is something you could reference re response matchers. Not sure if that's quite the direction we want to go, but it could be.

modules/caddyhttp/encode/gzip/gzip_precompressed.go
francislavoie: I suppose the same struct can be in two separate module namespaces :thinking:

mholt: Yeah I think this can be done. I don't think you need a type named `GzipPrecompressed` at all. Just give `Gzip` the methods it needs to satisfy the interface and register it as an `http.precompressed` module. I'll be surprised if this doesn't work.

ueffel: `Gzip` still has to be register under its encoder namespace `http.encoder.gzip`. This done via `CaddyModule() caddy.ModuleInfo` (https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/encode/gzip/gzip.go#L37-L43)
How would I register the same module again with a different namespace? Wouldn't I need a different implementation of `CaddyModule()`?

francislavoie: Right, I guess not then 😅

mholt: Oops, you're right. I forgot that module info goes both ways :man_facepalming:  A module registers a type to its name, but also needs to be able to return its name given its type. So I guess you're right, a single type can't be two modules. Not a big deal, thankfully.

modules/caddyhttp/encode/br/brotli_precompressed.go
ueffel: "." included in the suffix

modules/caddyhttp/encode/encode.go
ueffel:
```caddyfile
	encode zstd gzip {
		minimum_length 256
		prefer zstd gzip
		@text {
			status 2xx 4xx
			header Content-Type text/*
			header Content-Type application/javascript
			header Content-Type application/xhtml+xml
			header Content-Type application/atom+xml
			header Content-Type application/rss+xml
			header Content-Type image/svg+xml
		}
		match @text
	}
```

something like this is possible now

modules/caddyhttp/encode/encode.go
ueffel: with the implementation via `ResponseMatcher` the default is:
```
header Content-Type text/*
header Content-Type application/javascript
header Content-Type application/xhtml+xml
header Content-Type application/atom+xml
header Content-Type application/rss+xml
header Content-Type image/svg+xml
```

modules/caddyhttp/encode/gzip/gzip_precompressed.go
ueffel: TIL: just aliasing like
```go
type GzipPrecompressed Gzip
```
does not mean the aliasing type implements the same interfaces as the aliased type.
So an interface guard like this
```
var _ encode.Precompressed = (*GzipPrecompressed)(nil)
```
would does not compile

modules/caddyhttp/encode/caddyfile.go
```diff
@@ -64,24 +74,147 @@ func (enc *Encode) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+//     @name {
+//         header <field> [<value>]
+//         status <code...>
+//     }
+//
+// Or, single line syntax:
+//
+//     @name [header <field> [<value>]] | [status <code...>]
+//
+func (enc *Encode) parseNamedResponseMatcher(d *caddyfile.Dispenser, matchers map[string]caddyhttp.ResponseMatcher) error {
```
francislavoie, line 163: We'll probably want to move this somewhere common, but since it's internal I suppose it can wait until both this PR and the `handle_response` one is merged.

ueffel: Sounds good to me. I just shamelessly copied it 😇

modules/caddyhttp/encode/encode.go
francislavoie: Hmm. I'm thinking whether it would work better like this, in this case, since it only makes sense to define one matcher for `encode`, unlike `reverse_proxy`'s `handle_response` which can appear more than once:

```
encode [<encoders...>] {
	match [<matcher> [<args...]] {
		<matcher> <args...>
	}
}
```

Basically skip the named matcher stuff and directly read the args on the same line and/or block for the matcher. This shouldn't change the code much, and should still be reusable.

modules/caddyhttp/encode/encode.go
ueffel: Changed the default to:

```
header Content-Type text/*
header Content-Type application/json*
header Content-Type application/javascript*
header Content-Type application/xhtml+xml*
header Content-Type application/atom+xml*
header Content-Type application/rss+xml*
header Content-Type image/svg+xml*
```

to account for any `;charset=xxx` that may be appended to the header.

modules/caddyhttp/encode/encode.go
francislavoie: Awesome! Looks good 😄

Small thing, I think it might be better to make `parseNamedResponseMatcher` return the response matcher rather than shoving it into `responseMatchers` and read it back out.

francislavoie:
Could you add an adapt test as well for this? See https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt

modules/caddyhttp/encode/encode.go
ueffel: I haven't touched `parseNamedResponseMatcher` intentionally. If it will be reusable in that state from your pull request, the changes required to use the hopefully soon merged implementation will be minimal.
Also it ensures right now, that `match` is only defined once in a encode block.

modules/caddyhttp/encode/gzip/gzip_precompressed.go
Mohammed90: > TIL: just aliasing like
> 
> ```go
> type GzipPrecompressed Gzip
> ```
> 
> does not mean the aliasing type implements the same interfaces as the aliased type.

This is an alias, rather it's a new type definition. The example in [the "Alias declarations" and "Type definition" sections of the spec](https://golang.org/ref/spec#Type_declarations) show the distinction between the two.

modules/caddyhttp/encode/encode.go
xdevs23: Are there any MIME types that could match these blobs except for the `;charset=...` addition?
Could we use a regexp instead (or is it too much of a performance hit, maybe?)

modules/caddyhttp/encode/encode.go
```diff
@@ -43,12 +44,14 @@ type Encode struct {
...
 
-	// If the client has no strong preference, choose this encoding. TODO: Not yet implemented
-	// Prefer    []string `json:"prefer,omitempty"`
+	// If the client has no strong preference, choose these encodings in order.
+	Prefer []string `json:"prefer,omitempty"`
 
 	// Only encode responses that are at least this many bytes long.
 	MinLength int `json:"minimum_length,omitempty"`
 
+	Matcher *caddyhttp.ResponseMatcher
```
mholt, line 53: This needs a JSON tag :blush:

modules/caddyhttp/encode/encode.go
```diff
@@ -332,6 +392,13 @@ type Encoding interface {
 	NewEncoder() Encoder
 }
 
+// Precompressed is a type which returns filename suffix of precomressed
```
mholt, line 395:
```suggestion
// Precompressed is a type which returns filename suffix of precompressed
```

mholt (changes requested):
Awesome, this is looking so good!! Nice work.

Just a couple nits. But I think we're good to go. Caddyfile tests that Francis mentioned can come either in this PR or a later one. Also, if you want to make sure godoc comments end with periods (or other proper punctuation), that'd be good for legibility when they are read on the docs website.

Almost there!

modules/caddyhttp/encode/encode.go
ueffel: Regexps aren't supported by the `ResponseMatcher` type, yet. It would be nice the have a `header_regexp` directive from `RequestMatcher`s as well, but I think this would be out of the scope of this PR. Maybe an additional one?

francislavoie: I agree we should have a header_regexp matcher. If you want to do a follow-up with that, that'd be great 👍
I am a bit worried that it would be detrimental to performance for the `encode` defaults though, so a bit of benchmarking may be beneficial to see if it makes any appreciable difference.

ueffel:
> Could you add an adapt test as well for this? See https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt, both for `encode` and your `file_server` changes. Just for the coverage.

The `file_server` adapt test is giving me a headache. It automatically adds the "Caddyfile" with a OS-specific path separator as hidden file. So the test fails either on unix/mac or on windows, because in the adapt-txt file can only be one specific path separator. Should I use a placeholder for the path separator and added a replacement step in the step before the actual comparison is done?

francislavoie:
Right... Dangit. Yeah, I think a string replacement for `".\\Caddyfile"` to (I think) `./Caddyfile` should be enough, in `caddyfile_adapt_test.go` just before calling `CompareAdapt`. We already replace windows newlines with unix ones, so there's precedent for doing that.

ueffel:
> Edit: Sorry - I think the replacement might need to happen inside `CompareAdapt` cause what you actually need to replace is the string in the adapt output, cause that's the variable. It's a bit less clean to do it in there, but 🤷‍♂️ good enough I think.

I just did it [here](https://github.com/caddyserver/caddy/commit/10a4f2c5631ae4db6a5befefd27ae2ecce822b5a#diff-2cce32ea4df268884ed931b70eaf3e0553160cd50b56745f3f3ce8c8032944a3R45-R48), if that's ok? It seems to work, tests are nit failing anymore.

francislavoie:
Yeah that seems okay to me 👍

ueffel:
So, I would consider this PR ready-to-merge or at least ready-for-final-review.

francislavoie (dismissed):
This is excellent work! Thank you!

If you find the time, it would be great if you could update the docs for https://github.com/caddyserver/website/blob/master/src/docs/markdown/caddyfile/directives/encode.md and https://github.com/caddyserver/website/blob/master/src/docs/markdown/caddyfile/directives/file_server.md to cover the changes :memo:

modules/caddyhttp/encode/gzip/gzip.go
```diff
@@ -81,6 +81,9 @@ func (g Gzip) Validate() error {
 // used in the Accept-Encoding request headers.
 func (Gzip) AcceptEncoding() string { return "gzip" }
 
+// Suffix returns the filename suffix of precompressed files.
+func (Gzip) Suffix() string { return ".gz" }
```
mholt, line 85: Is this still needed since there's the `GzipPrecompressed` type?

ueffel: I moved the implementation from `GzipPrecompressed` to `Gzip` because of [this review](https://github.com/caddyserver/caddy/pull/4045#discussion_r585891879)

But I can move it back to `GzipPrecompressed` as it makes more sense there. Same for `ZstdPrecompressed`.
You tell me how you want it :)

mholt (approved):
Awesome, this looks good -- let's give it a try. Thanks for your persistence in helping to get this ready! Please contribute again sometime.

PR #4046, opened 2021-03-03
reverseproxy: Fix upstreams with placeholders with no port

francislavoie:
See https://caddy.community/t/migrate-to-using-a-wildcard-certificate/11679/21

Fixes a bug where an upstream with a placeholder would append a `:` in the JSON.

Before:

```json
{
	"handler":"reverse_proxy",
	"upstreams":[
		{
			"dial": "{upstream}:"
		}
	]
}
```

After:

```json
{
	"handler":"reverse_proxy",
	"upstreams":[
		{
			"dial": "{upstream}"
		}
	]
}
```


mholt (approved):
Ah thanks. I guess this'll do! (Not sure what else we'd do anyway.)

PR #4050, opened 2021-03-04
reverseproxy: Implement health_uri, replaces health_path, supports query

### LINKED ISSUE #4049: Support query parameters on active healthcheck
Hi!

When using `health_path` instruction in an active health check context, [query parameters](https://www.cockroachlabs.com/docs/stable/monitoring-and-alerting.html#health-ready-1) are not supported, response from a health check query is a 404 not found:

```
reverse_proxy / {
    to 192.101.153.159:8080 192.118.109.166:8080
    health_path /health?ready=1
    health_status 200
}
```

Current behavior (404 not found) can be misleading.
Thanks for your help

francislavoie:
Closes #4049

```
reverse_proxy localhost:54321 {
	health_uri /health?ready=1
	health_status 2xx
}
```

Also fixes a bug with `health_status` Caddyfile parsing, it would always only take the first character of the status code even if it didn't end with "xx". I fixed the same bug in #4021 but it's relevant here too because it affects the adapt test I wrote (will rebase #4021 after and remove the redundant fix)

### DIFF
--- a/modules/caddyhttp/reverseproxy/healthchecks.go
+++ b/modules/caddyhttp/reverseproxy/healthchecks.go
@@ -51,8 +51,12 @@ type HealthChecks struct {
 // health checks (that is, health checks which occur in a
 // background goroutine independently).
 type ActiveHealthChecks struct {
-	// The URI path to use for health checks.
+	// The path to use for health checks.
+	// DEPRECATED: Use 'uri' instead.
 	Path string `json:"path,omitempty"`
+
+	// The URI (path and query) to use for health checks
+	Uri string `json:"uri,omitempty"`
 
 	// The port to use (if different from the upstream's dial
 	// address) for health checks.
@@ -79,6 +83,7 @@ type ActiveHealthChecks struct {
 	// body of a healthy backend.
 	ExpectBody string `json:"expect_body,omitempty"`
 
+	uri        *url.URL
 	httpClient *http.Client
 	bodyRegexp *regexp.Regexp
 	logger     *zap.Logger
@@ -218,7 +223,15 @@ func (h *Handler) doActiveHealthCheck(dialInfo DialInfo, hostAddr string, host H
 	u := &url.URL{
 		Scheme: scheme,
 		Host:   hostAddr,
-		Path:   h.HealthChecks.Active.Path,
+	}
+
+	// if we have a provisioned uri, use that, otherwise use
+	// the deprecated Path option
+	if h.HealthChecks.Active.uri != nil {
+		u.Path = h.HealthChecks.Active.uri.Path
+		u.RawQuery = h.HealthChecks.Active.uri.RawQuery
+	} else {
+		u.Path = h.HealthChecks.Active.Path
 	}
 
 	// adjust the port, if configured to be different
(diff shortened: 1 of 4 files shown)


francislavoie:
Flaky tests :cry:

mholt:
So here's a curveball, what if we introduce `health_uri` and deprecate `health_path`? The URI is just the path + optional query string in this case. That way the user doesn't have to separate them, seems kinda odd to require them separate...

francislavoie:
Sure, that works 👍 I'll rework it

francislavoie:
@mholt Done :smile:

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -51,9 +51,13 @@ type HealthChecks struct {
 // health checks (that is, health checks which occur in a
 // background goroutine independently).
 type ActiveHealthChecks struct {
-	// The URI path to use for health checks.
+	// The path to use for health checks.
+	// DEPRECATED: Use 'uri' instead.
 	Path string `json:"path,omitempty"`
 
+	// The URI (path and query) to use for health checks
+	Uri string `json:"uri,omitempty"`
```
mholt, line 59:
```suggestion
	URI string `json:"uri,omitempty"`
```

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -273,15 +274,26 @@ func (h *Handler) Provision(ctx caddy.Context) error {
...
+			h.HealthChecks.Active.Port != 0) {
+
 			h.HealthChecks.Active.logger = h.logger.Named("health_checker.active")
 
 			timeout := time.Duration(h.HealthChecks.Active.Timeout)
 			if timeout == 0 {
 				timeout = 5 * time.Second
 			}
 
+			// parse the Uri string (supports path and query)
```
mholt, line 288:
```suggestion
			// parse the URI string (supports path and query)
```

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -273,15 +274,26 @@ func (h *Handler) Provision(ctx caddy.Context) error {
...
+			if h.HealthChecks.Active.Uri != "" {
+				url, err := url.Parse(h.HealthChecks.Active.Uri)
```
mholt, line 290:
```suggestion
				parsedURI, err := url.Parse(h.HealthChecks.Active.Uri)
```

Just to avoid any confusion with the `url` package (even though scope is small).

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -51,9 +51,13 @@ type HealthChecks struct {
 // health checks (that is, health checks which occur in a
 // background goroutine independently).
 type ActiveHealthChecks struct {
-	// The URI path to use for health checks.
+	// The path to use for health checks.
+	// DEPRECATED: Use 'uri' instead.
 	Path string `json:"path,omitempty"`
```
mholt, line 56: We should probably emit a warning if this is used, too (separate from Caddyfile). (It's OK if warning is emitted twice if Caddyfile is used too.)

mholt (changes requested):
Thanks for working on this!

modules/caddyhttp/reverseproxy/healthchecks.go
francislavoie: Right - I was thinking about that but wasn't sure if I should make it double up or not. Will do.

mholt (approved):
Thank you for helping with this! This is a nice improvement.

PR #4060, opened 2021-03-12
notify: Send all sd_notify signals from main caddy process

carlwgeorge:
Initial sd_notify support was added in #3963, but that sent signals from both cmdRun and cmdReload.  This approach has two drawbacks:

- Reloads initiated via the API do not send signals.
- The signals are sent from different processes, which requires the `NotifyAccess=exec` directive in the unit file.

This change moves the NotifyReloading and NotifyReadiness invocations to Load, which address both of those drawbacks.  It also adds a complimentary NotifyStopping method which is invoked from handleStop.  All the notify methods are defined in a notify package to avoid an import loop.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -39,6 +39,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/caddyserver/caddy/v2/notify"
 	"github.com/caddyserver/certmagic"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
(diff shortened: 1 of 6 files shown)


francislavoie:
I think this looks good now!

francislavoie:
https://github.com/carlwgeorge/caddy/blob/f559bb10f64e09e4462591f841a14a2aed0f994d/notify/notify_linux.go#L45 has a typo that https://github.com/caddyserver/caddy/pull/4081 aims to fix, would you mind making the change in here (to avoid the conflict)?

carlwgeorge:
Sure thing, added that typo fix and went ahead and rebased on master.

francislavoie (approved):
This looks good to me.

@rumpelsepp Do you have any thoughts about this?

rumpelsepp:
These adjustments make sense, I would say. :+1:

mholt (approved):
Great, thank you very much for the enhancement.

PR #4063, opened 2021-03-14
fileserver: Add a few more debug lines for fileserver problems

francislavoie:
I noticed file server had debug lines for most everything else but the most common problem people have are file permissions or files not being found. Showing those cases when attempting to open the file would make debugging much easier.


simaotwx (approved):

mholt (approved):
Thanks for this Francis! Should be helpful.

PR #4077, opened 2021-03-21
httpcaddyfile: Add `auto_https ignore_loaded_certs`

### LINKED ISSUE #4048: the ignore_loaded_certificates config for Caddyfile
I can't find the ignore_loaded_certificates config for Caddyfile

I have some domain behind cloudfare strict https so I need use cloudfare wildcard certificates, but I also have some domains is direct to my host and these domains need autohttps, but now I can't do that, because I can't find this config https://caddyserver.com/docs/json/apps/http/servers/automatic_https/#ignore_loaded_certificates

francislavoie:
Closes #4048

Adds `ignore_loaded_certs` as an option for the `auto_https` global option, which turns on `ignore_loaded_certificates`, i.e. making the automation policy still run for domains in the server even if they're covered by otherwise loaded certificates (such as a wildcard cert that covers that specific domain).

~~I went with the name `always` as suggested by a user in #4048 because I think it somewhat better describes what happens in a more user-friendly way, even if it's somewhat ambiguous.~~

And obviously it's not possible to set `ignore_loaded_certs` and `disable_redirects` at the same time, but I think the likelyhood of that being a valid usecase is quite low.

Example:

```
{
	auto_https ignore_loaded_certs
}

*.example.com {
	tls {
		dns <provider>
	}
	respond "Wildcard"
}

# Without `ignore_loaded_certs`, this site would not have its own
# certificate, but instead inherit the wildcard cert from `*.example.com`
foo.example.com {
	respond "Foo"
}
```

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -126,10 +126,10 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 			// must load each cert only once; otherwise, they each get a
 			// different tag... since a cert loaded twice has the same
 			// bytes, it will overwrite the first one in the cache, and
-			// only the last cert (and its tag) will survive, so a any conn
-			// policy that is looking for any tag but the last one to be
-			// loaded won't find it, and TLS handshakes will fail (see end)
-			// of issue #3004)
+			// only the last cert (and its tag) will survive, so any conn
+			// policy that is looking for any tag other than the last one
+			// to be loaded won't find it, and TLS handshakes will fail
+			// (see end of issue #3004)
 			//
 			// tlsCertTags maps certificate filenames to their tag.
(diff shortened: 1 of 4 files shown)


mholt (changes requested):
I like the natural language sound of this, but I think it also throws me off. Hmm. This is hard to articulate, but ... "always auto-HTTPS"... if it's "always", what is it in opposition to? I feel like this creates some sort of false dichotomy, like manually-loaded certs are the only way to disable it. There's got to be a better way to call this. For example, `auto_https ignore_loaded_certs` might just be more clear, even if less "elegant" in its name. (It's kind of an obscure option anyways.)

francislavoie:
Yeah you're probably right, I'll switch it.

mholt (approved):
Great, looking better. The nice thing about this is we can later change it to support multiple arguments since they aren't mutually exclusive.

PR #4085, opened 2021-03-23
headers: Fix Caddyfile parsing for `request_header` with matchers

### LINKED ISSUE #4084: request_header directive broken with matcher
If I try to use the `request_header` with a directive such as
`request_header @internal x-custom caddy something`

I will receive the error message 
```
Error during parsing: Wrong argument count or unexpected line ending after 'something'
```

I came across this because I cannot use a matcher with this directive. As soon as I add one, the directive will never be applied.

The documentation seems to state that the `request_header` should be able to be used with matchers, and a brief look into the source suggests to me that you are trying to parse matchers as well.

My current matcher is
```
@internal {
    path *
}
```

But I get the same result no matter what I configure in the named matcher.

francislavoie:
Fixes #4084

The first token needs to be skipped with `h.Next()`, otherwise the matcher token is read as the header field.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/request_header.txt
+++ b/caddytest/integration/caddyfile_adapt/request_header.txt
@@ -1,0 +1,90 @@
+:80
+
+@matcher path /something*
+request_header @matcher Denis "Ritchie"
+
+request_header +Edsger "Dijkstra"
+request_header -Wolfram
+
+@images path /images/*
+request_header @images Cache-Control "public, max-age=3600, stale-while-revalidate=86400"
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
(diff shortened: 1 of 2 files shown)


silverbackdan:
Would you be able to give me an idea of a roadmap for this PR to be merged and available to install from a Docker image just so I stick something in my diary to sort out my hacky workaround?

francislavoie:
If you want to use it now, you can make a custom build by building from this branch:

```
xcaddy build caddyfile-request-header
```

There's instructions on https://hub.docker.com/_/caddy, see the "Adding custom Caddy modules" section (you don't need to add plugins with `--with`)

mholt (approved):
Ah, gracias!

PR #4087, opened 2021-03-24
.gitignore: add IDE files

simaotwx:
These tend to clutter up repositories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ vendor
 dist
 caddy-build
 caddy-dist
+
+# IDE files
+.idea/
+.vscode/


francislavoie (approved):

mholt:
Thank you!

(Huh, I use VS Code but don't have any .vscode folder.)

simaotwx:
> Thank you!

You're welcome :)

> (Huh, I use VS Code but don't have any .vscode folder.)

I think it depends on your local settings and configuration. Certain extensions and settings may create this folder.

Here is a more thorough .gitignore file (certain files would still be checked in if you use this one): https://github.com/github/gitignore/blob/218a941be92679ce67d0484547e3e142b2f5f6f0/Global/VisualStudioCode.gitignore

francislavoie:
I did have a `.vscode/` dir when I played around with delve a while back to debug something, because I was configuring the "Run" button in VSCode, which wrote a project-specific .json config file.

PR #4099, opened 2021-03-31
Release 386

sagudev:
Some people are using old computers as personal servers.


Mohammed90:
> Some people are using old computers as personal servers.

Thank you for the contribution. However, including x86/386 arch was a deliberate decision made on #3217 due to the very low number of users. Nonetheless, you can always use [xcaddy](github.com/caddyserver/xcaddy) to build x86 binaries from source. I hope you understand.

PR #4103, opened 2021-04-02
caddyfile: Normalize line endings before comparing fmt result

francislavoie:
This is a minor bug I noticed when working on Windows; `caddy fmt` always outputs unix line endings (LF), so if a Caddyfile crafted on Windows with CRLF line endings, the `input is not formatted with 'caddy fmt'` warning would always appear.

This change normalizes the input Caddyfile to LF line endings to avoid this problem when comparing.


mholt (approved):
Awesome, thanks!

PR #4105, opened 2021-04-04
ci: fuzz: add more fuzz targets

Mohammed90:
I'm prepping to bring back the continuous fuzzing of Caddy on another platform after the departure of FuzzIt. In the process, I found more possible targets that can be included based on the criteria of:
- accept string|[]byte as input
- function body is more than mere `json.Unmarshal`

One dilemma for now is where to store the seed corpus to be retrieved by the new platform. Any thoughts? I think some projects store them in a branch, but not sure how much of a hassle will it be to keep that branch updated. I've considered hiding them behind a `refs/<something>/corpus`, but it feels too magical and needs insider knowledge.

### DIFF
--- a/caddyconfig/caddyfile/formatter_fuzz.go
+++ b/caddyconfig/caddyfile/formatter_fuzz.go
@@ -1,0 +1,25 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


francislavoie:
I think a `fuzz-corpus` branch, or a separate repo, if possible, would be ok.

mholt (approved):
LGTM, thanks as usual @Mohammed90 :slightly_smiling_face: 

As for where to store the corpus, a branch sounds good to me also. Would it get very large though? If so, maybe a separate repo, in a (possibly vain) attempt to keep this repo as lean as possible.

Mohammed90:
> LGTM, thanks as usual @Mohammed90 🙂
> 
> As for where to store the corpus, a branch sounds good to me also. Would it get very large though? If so, maybe a separate repo, in a (possibly vain) attempt to keep this repo as lean as possible.

It won't be the generated corpus. It'll just contain the seeds and the corpus I extracted from Fuzzit before its demise. Going with a branch, I'll do something like `git checkout --orphan fuzz-seed-corpus` and commit only the seed `.zip` files.

mholt (approved):
Awesome, let's see if it helps! Thanks!

PR #4111, opened 2021-04-12
logging: Implement dial timeout for net writer (fix #4083)

### LINKED ISSUE #4083: Bug: Caddy freezes when syslog server not available
Hello,

I had a power outage over the weekend and noticed an issue with one of my sites, when looking into the issue, caddy was locking up whenever browsing the one domain that uses a syslog output

```
somewebsite.com {
        reverse_proxy host:port
        log {
                output net syslog-server:port
        }
}
```

Is the config that is being used. Is this behavior expected? or am I breaking something?

mholt:
My hypothesis is that if a dial timeout is configured, the server will still be able to respond to requests even if the log server is down, although there will probably still be a delay for up to the dial timeout.

I haven't had a chance to test this, if someone else could that would be great.

### DIFF
--- a/modules/logging/netwriter.go
+++ b/modules/logging/netwriter.go
@@ -18,7 +18,9 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"os"
 	"sync"
+	"time"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
@@ -30,7 +32,11 @@ func init() {
 
 // NetWriter implements a log writer that outputs to a network socket.
 type NetWriter struct {
+	// The address of the network socket to which to connect.
 	Address string `json:"address,omitempty"`
+
+	// The timeout to wait while connecting to the socket.
+	DialTimeout caddy.Duration `json:"dial_timeout,omitempty"`
 
 	addr caddy.NetworkAddress
 }
@@ -60,6 +66,10 @@ func (nw *NetWriter) Provision(ctx caddy.Context) error {
 		return fmt.Errorf("multiple ports not supported")
 	}
 
+	if nw.DialTimeout < 0 {
+		return fmt.Errorf("timeout cannot be less than 0")
+	}
+
 	return nil
 }
 
@@ -74,7 +84,10 @@ func (nw NetWriter) WriterKey() string {
 
 // OpenWriter opens a new network connection.
 func (nw NetWriter) OpenWriter() (io.WriteCloser, error) {
-	reconn := &redialerConn{nw: nw}
+	reconn := &redialerConn{
+		nw:      nw,
+		timeout: time.Duration(nw.DialTimeout),
+	}
 	conn, err := reconn.dial()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


modules/logging/netwriter.go
```diff
@@ -134,6 +166,8 @@ func (reconn *redialerConn) Write(b []byte) (n int, err error) {
 	// we're the lucky first goroutine to re-dial the connection
 	conn2, err2 := reconn.dial()
 	if err2 != nil {
+		// logger socket seems to be offline; instead of discarding the log, dump it to stderr
+		os.Stderr.Write(b)
```
simaotwx, line 170: Do we have some way of notifying the admin of the dial failure via stderr?
Something along the lines of:

```go
os.Stderr.Write([]byte("No connection to network log, message: "))
os.Stderr.Write(b)
```

mholt: How often do you want to be notified? Every log message? Or one log message every 10 minutes, or...?

simaotwx: I'd say every log message so that it immediately catches attention.
Since this is a temporary, probably important failure, IMO it would be appropriate.

mholt: Would a single error message not "catch attention" though? Why would you need 100s of them (especially in hot paths).

simaotwx: Was thinking that when there is only one error message, it would be hard to notice if there are many other log lines after that. Now that I think about it again, if someone was to look into the logs and see them being printed out, something can't be right as they should be sent to the network log instead. Should probably go with a single error message and see if it fits people's needs. For me it would be enough as I'd know the behavior.

simaotwx: The idea behind my initial suggestion is just to let the user know that there is an issue (and what the issue is, in words) instead of just dumping the logs without such an error message. (`err` is returned here. maybe that's already enough?)

francislavoie:
I had an idea... Could we make the netwriter do its work in a goroutine? We don't care about the result really, so we can offload it from the request's goroutine to happen async, that way the request/response isn't blocked if something goes wrong, and we can write an error to stderr if the timeout happens. Am I wrong with this assumption?

mholt:
@francislavoie It does risk piling up goroutines faster than they terminate, leading to resource exhaustion.

francislavoie:
But at least the goroutine for the request itself would be released, no? Obviously we still need the timeout here regardless, but I feel this would lessen the pain. Sucks for requests to become slow just due to logging.

mholt:
> Sucks for requests to become slow just due to logging.

Yeah, so... don't let your log server go down?

I know that sounds dumb and unhelpful, but it's either slow requests or exhaust resources due to an external problem. Unless there's another clever solution. One thing I'm considering is only retrying to redial after a certain amount of time, and just assume it's down if we retried within that interval (and immediately log to stderr instead of trying to redial every time).

francislavoie:
Yeah sounds like doing an "upstream pool" type of deal for this would be a good idea.

mholt:
I don't think so... that's much too complicated.

francislavoie:
Whoops, I think this PR will need to be recreated. TIL we should've re-targeted this PR before merging the `2.4` branch.

mholt:
Superceded by #4172

PR #4123, opened 2021-04-19
caddytls: Implement remote IP connection matcher

mholt:
I'm not sure this is a good idea as a replacement for auth, but I think this could still be useful. See https://caddy.community/t/best-solution-for-mixing-lan-and-cloudflare-authenticated-origin-pulls/12139/8

### DIFF
--- a/context.go
+++ b/context.go
@@ -430,5 +430,12 @@ func (ctx Context) Storage() certmagic.Storage {
 
 // Logger returns a logger that can be used by mod.
 func (ctx Context) Logger(mod Module) *zap.Logger {
+	if ctx.cfg == nil {
+		l, err := zap.NewDevelopment()
+		if err != nil {
+			panic("config missing, unable to create dev logger: " + err.Error())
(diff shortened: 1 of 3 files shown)


mholt:
I know this implementation using `NotRanges` might be disappointing, @francislavoie, but I realized that the connection policy matchers aren't really designed to be composed like the HTTP request matchers (conn policy matchers are only a list, whereas HTTP request matchers are a list of maps of lists). So I just went ahead with the simplest possible implementation to negate IP ranges.

francislavoie:
That's ok 👍

wazerstar:
Hey was this nuked again? I'm actually in this boat now, want to use same domain and when on lan with origin destination would avoid getting out on cloudflare and back in again before reaching internal.

francislavoie:
@wazerstar this isn't the right place to ask questions. Please ask on the forums instead, and fill out the help topic template: https://caddy.community

PR #4129, opened 2021-04-22
caddyfile: Fix `import` replacing unrelated placeholders

francislavoie:
See https://caddy.community/t/snippet-issue-works-outside-snippet/12231

So it turns out that `NewReplacer()` gives a replacer with some global defaults (like `{env.*}` and some system and time placeholders), which is not ideal when running `import` because we just want to replace `{args.*}` only, and nothing else.

### DIFF
--- a/caddyconfig/caddyfile/parse.go
+++ b/caddyconfig/caddyfile/parse.go
@@ -321,7 +321,7 @@ func (p *parser) doImport() error {
 	args := p.RemainingArgs()
 
 	// add args to the replacer
-	repl := caddy.NewReplacer()
+	repl := caddy.NewEmptyReplacer()
 	for index, arg := range args {
 		repl.Set("args."+strconv.Itoa(index), arg)
 	}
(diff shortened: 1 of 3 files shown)


francislavoie:
I should probably add a test for this, but it can be merged without. It's on my TODO.

mholt (approved):
Interesting. Thanks for the patch. Can't think of a better way to do it, so let's roll with it.

Dulanic:
Not sure that this corrected it. It is passing the env variable name not the env variable value. I looked at the autosave.json and all the other env variables show the value, but this same one just shows the name instead.

image

francislavoie:
That's working as intended. The `{env.*}` are replaced at runtime, and only if the module being configured passes it through the replacer. The `{$ENV}` style are replaced at Caddyfile adapt time, i.e. before it's transformed into JSON, and will work anywhere, but will leak the values to the `autosave.json`

francislavoie:
FYI @greenpau make sure you're running the replacer on all config values that are relevant either during `ServeHTTP` (if it makes sense for the value to be dynamic) or during `Provision` (if it only makes sense to be a single value for the lifetime of the server).

Dulanic:
OK, I'll work /w @greenpau under his issue ticket since I'm still seeing errors. https://github.com/greenpau/caddy-auth-portal/issues/122#issuecomment-828481342

PR #4132, opened 2021-04-24
fileserver: Better handling of HTTP status override

francislavoie:
Just something I noticed when reading through the code; I think the `io.Copy` was a remnant of an older version of that code. I don't see a reason why we don't just let it fall through and serve the file with `http.ServeContent` normally. More consistent that way.

~~Also, I hoisted the `status_code` override logic above the error handler detection bit, because then it allows for overriding the status code during error routes. Not necessarily that likely to be useful, but it's more correct. (Reminder that the first `w.WriteHeader()` wins, subsequent ones have no effect). Marking this as a bug just for this reason, but it's barely worth discussing as a bug 😅.~~

Edit: So calling `w.WriteHeader()` ourselves makes the `w.Header().Set()` operations in `http.ServeContent` not actually work correctly, which isn't ideal.

After a bit of googling, I found a great solution, by wrapping the `ResponseWriter` to intercept the `w.WriteHeader()` call that `http.ServeContent` does to write the header we want instead.

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"fmt"
 	"html/template"
-	"io"
 	weakrand "math/rand"
 	"mime"
 	"net/http"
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -333,10 +332,21 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
...
 	}
 
+	// if a status code override is configured, write the status code
+	// before serving the file
+	if codeStr := fsrv.StatusCode.String(); codeStr != "" {
+		intVal, err := strconv.Atoi(repl.ReplaceAll(codeStr, ""))
+		if err != nil {
+			return caddyhttp.Error(http.StatusInternalServerError, err)
+		}
+		w.WriteHeader(intVal)
```
mholt, line 342: The problem with this, is that ServeContent can add additional header fields, such as Content-Type, Content-Length, and Content-Range... by writing the header now, we forfeit all of those.

francislavoie: Are you sure? I was pretty sure you can still add headers after writing the status?

mholt: I'm sure; once you call WriteHeader, the header is written, as one might expect.

Not that `io.Copy()` is any better in that sense... i.e. this might not be a new problem, but it'd be nice if we could actually fix it.

francislavoie: A bit of googling landed me on https://www.howtobuildsoftware.com/index.php/how-do/dQY/go-serving-static-files-with-a-http-500-status which gave me the idea of wrapping the ResponseWriter to intercept the `WriteHeader` call 🎉 🧠

mholt (approved):
Thanks, this LGTM.

You did test this, right? :smile:

francislavoie:
Yep, with this Caddyfile, and a `404.html` file in $PWD

```
{
        admin off
        debug
}

:8883 {
        root * .

        handle_errors {
                rewrite * /404.html
                file_server
        }

        file_server
}
```

Request to `http://localhost:8883/` serves `/404.html` with status 404, and `Content-Length` headers are set (wasn't before)

mholt:
Perfect, thanks so much!

Issue #4135, opened 2021-04-28
[Bug] reverse_proxy can skip hosts when using any hashing policy


Timbus:
TLDR: https://github.com/caddyserver/caddy/blob/a8d45277ca387ca4769fb4e881d242a5896cca5a/modules/caddyhttp/reverseproxy/selectionpolicies.go#L526 this will skip hosts if it loops, because you're incrementing `i` as well as `index`. Please avoid mutating things in a loop..

Anyway just replace `+=` with `+` I guess, and just throw it inline or something.

Learned about this in the worst possible way (in production), because our one healthy host wasn't getting traffic some of the time.

Timbus:
Long version (was too tired to elaborate): When a host isn't healthy/available, the hash selector tries to move on to the next healthy host in the pool array. The issue is the code is incrementing the index 'pointer' by `i` each time, instead of by 1. So it's looping through each host in some fibbonacci-esque.. triangle number way. Which is just weird and won't work.
There are a few ways to improve this, such as changing the loop to a `while (i != start_i)`, or statically calculating the offset as 'index + i'

An easy testcase for this is to have a pool of 3 hosts, two unavailable: [Dead, Dead, Ok], and if you start at i = 0, index = 0, and use the above code I linked you'll miss the last element, thus returning nil and logging to the user that no healthy hosts are available.

francislavoie:
PRs welcome! :slightly_smiling_face:

simaotwx:
I proposed a fix that does not break existing tests, however I haven't added specific tests that replicate the reported behavior (yet)

mholt:
Thank you for the report, and for the fix, respectively!

PR #4137, opened 2021-04-29
caddyhttp: reverseproxy: fix hash selection policy

simaotwx:
The tests ran without an issue, so this should be the fix.

Fixes: #4135
Test: go test './...' -count=1

### DIFF
--- a/modules/caddyhttp/reverseproxy/selectionpolicies.go
+++ b/modules/caddyhttp/reverseproxy/selectionpolicies.go
@@ -523,8 +523,7 @@ func hostByHashing(pool []*Upstream, s string) *Upstream {
 	}
 	index := hash(s) % poolLen
 	for i := uint32(0); i < poolLen; i++ {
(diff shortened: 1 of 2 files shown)


francislavoie:
We'll definitely need a test that fails _before_ the change to make sure it actually fixes the problem.

simaotwx:
Test case added. It fails before applying my patch and passes afterwards.

simaotwx:
Is it appropriate to reverse the order of the commits? Maybe that would immediately prove that the commit after it fixes the issue.

francislavoie:
No, that's okay! As long as you confirmed it, that's good :+1: 

Thanks!

francislavoie (approved):

simaotwx:
Alright, perfect. :)

mholt (approved):
Excellent; thank you so much for the fix and the test case! Please feel free to contribute again.

PR #4141, opened 2021-04-30
cmd: Add `envfile` flag to `start` command

jesse996:
add envfile in start cmd

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -65,7 +65,6 @@ func cmdStart(fl Flags) (int, error) {
 	// ensure it's the process we're expecting - we can be
 	// sure by giving it some random bytes and having it echo
 	// them back to us)
-
 	cmd := exec.Command(os.Args[0], "run", "--pingback", ln.Addr().String())
 	if startCmdConfigFlag != "" {
 		cmd.Args = append(cmd.Args, "--config", startCmdConfigFlag)
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -45,6 +45,7 @@ func cmdStart(fl Flags) (int, error) {
 	startCmdConfigAdapterFlag := fl.String("adapter")
 	startCmdPidfileFlag := fl.String("pidfile")
 	startCmdWatchFlag := fl.Bool("watch")
+	startCmdEnvfileFlag := fl.Bool("envfile")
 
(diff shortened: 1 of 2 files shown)


cmd/commandfuncs.go
```diff
@@ -64,6 +65,14 @@ func cmdStart(fl Flags) (int, error) {
...
 	// sure by giving it some random bytes and having it echo
 	// them back to us)
+
+	// load all additional envs as soon as possible
+	if startCmdEnvfileFlag != "" {
+		if err := loadEnvFromFile(startCmdEnvfileFlag); err != nil {
+			return caddy.ExitCodeFailedStartup,
+				fmt.Errorf("loading additional environment variables: %v", err)
+		}
+	}
```
mholt, line 75: Instead of doing this, I would just pass the flag through to `run` like the other flags below. (Also, this section is in the way, between the comment before it which relates to the next code after it.)

mholt (changes requested):
Thanks for the PR! I think it could be even simpler/easier.

cmd/commandfuncs.go
jesse996: thanks ,i've fixed it~

cmd/commandfuncs.go
```diff
@@ -68,6 +69,9 @@ func cmdStart(fl Flags) (int, error) {
 	if startCmdConfigFlag != "" {
 		cmd.Args = append(cmd.Args, "--config", startCmdConfigFlag)
 	}
+	if startCmdEnvfileFlag != "" {
+		cmd.Args = append(cmd.Args, "--envfile", startCmdEnvfileFlag)
```
francislavoie, line 73: Hmm, this isn't working because `startCmdEnvfileFlag` is a boolean and not a string.

francislavoie (changes requested):

cmd/commandfuncs.go
jesse996: sorry,i've fixed `startCmdEnvfileFlag`  to string

mholt (approved):
Thanks, LGTM!

francislavoie (approved):
I checked out your branch and tried it out, work well! 🎉

mholt:
Thank you for the contribution @jesse996, and thanks for testing, Francis.

PR #4142, opened 2021-05-01
fix struct field alignment

calvinxiao:
`maligned` analysis shows below info:

```
modules/caddytls/automation.go:83:23: struct of size 192 bytes could be of size 184 bytes
admin.go:50:18: struct of size 80 bytes could be of size 72 bytes
modules/caddyhttp/autohttps.go:33:22: struct of size 64 bytes could be of size 56 bytes
```

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -48,24 +48,10 @@ import (
 // AdminConfig configures Caddy's API endpoint, which is used
 // to manage Caddy while it is running.
 type AdminConfig struct {
-	// If true, the admin endpoint will be completely disabled.
-	// Note that this makes any runtime changes to the config
-	// impossible, since the interface to do so is through the
-	// admin endpoint.
-	Disabled bool `json:"disabled,omitempty"`
-
 	// The address to which the admin endpoint's listener should
 	// bind itself. Can be any single network address that can be
 	// parsed by Caddy. Default: localhost:2019
 	Listen string `json:"listen,omitempty"`
-
-	// If true, CORS headers will be emitted, and requests to the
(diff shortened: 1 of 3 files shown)


calvinxiao:
Reading https://github.com/caddyserver/caddy/pull/3245, these little optimizations are probably not worth it.

francislavoie:
I would tend to agree. These structs are often only allocated once per config, so this few amount of bytes gained is basically insignificant. I think I prefer the order they are in for readability.

But I'll let @mholt make the decision.

mholt:
Hmm... the structs are basically singletons, except the automation policies, but even then, shaving just 8 bytes off won't make a difference for most users (even ones with thousands of them), and the order of the fields is how they're presented in the docs.

Then again, I personally don't care about the order too much. I don't think this is a regression in readability, personally. Maybe it's just what we're used to?

calvinxiao:
@mholt 

I think we can close this PR, but someone may send the same PR using `maligned` analysis in the future.

My performance workflows for Go projects are:

1. Run `maligned`  to improve struct padding.
2. run `go build  -gcflags '-m'` to find unnecessary heap allocation, the fix might be, replace `fmt` or `binary` package used in hot patth, `return value` instead of `return reference`.

mholt:
That's a good flow; I appreciate your contributions!

I don't have an opinion on this particular change (except perhaps that some unexported fields now come between exported fields, which does just sit a little "off" with me, since I'm not used to that, personally). If anyone wants to play tiebreaker, we can merge or close this. Or I can just decide. But again, I have no preference.

Issue #4146, opened 2021-05-04
file matcher should not match `/`


princemaple:
On Caddy 2.3

------

```
@is_file {
  file
}

handle @is_file {
  file_server
}

handle {
  reverse_proxy ...
}
```

Going to `/` seems to match `is_file` even though `/` isn't a file and there isn't a `index.html`. This causes 404 directly from caddy while the intended response is from upstream server.

```
@is_file {
  file {path}
}

handle @is_file {
  file_server
}

handle {
  reverse_proxy ...
}
```

Note the `{path}`, works as expected.

Wondering if it's by design?

francislavoie:
Hmm, yeah I guess this might be a bug.

For @mholt, the explanation:

If `TryFiles` is empty, we fill it with `r.URL.Path`. In this case, this is `/`. Then later, in `prepareFilePath()`, we run the replacer (which turns `{path}` into `/` at that point) but `file` remains the original value (and the placeholder is still the placeholder there).

So then `strings.HasSuffix(file, "/")` will be `false` for the placeholder, but `true` for the empty `TryFiles` codepath, because `file` was `/` due to being set to the actual request value beforehand.

This means that `suffix` becomes `//` in that case, so after `sanitizedPathJoin`, it becomes `./`, so `strictFileExists`'s `strings.HasSuffix(file, separator)` codepath will return true.

I think we should change the `m.TryFiles == nil` codepath to `m.TryFiles = []string{"{http.request.uri.path}"}` for consistency. (And maybe consider hoisting this to `Provision` cause there's no point doing this on every request). I don't think this "optimization" of directly using `r.URL.Path` is so valuable, cause it causes this edgecase with directories.

PR #4147, opened 2021-05-04
fileserver: Fix `file` matcher with empty `try_files`

francislavoie:
Fixes https://github.com/caddyserver/caddy/issues/4146

If `TryFiles` is empty, we fill it with `r.URL.Path`. In this case, this is `/`. Then later, in `prepareFilePath()`, we run the replacer (which turns `{path}` into `/` at that point) but `file` remains the original value (and the placeholder is still the placeholder there).

So then `strings.HasSuffix(file, "/")` will be `false` for the placeholder, but `true` for the empty `TryFiles` codepath, because `file` was `/` due to being set to the actual request value beforehand.

This means that `suffix` becomes `//` in that case, so after `sanitizedPathJoin`, it becomes `./`, so `strictFileExists`'s `strings.HasSuffix(file, separator)` codepath will return true.

I think we should change the `m.TryFiles == nil` codepath to `m.TryFiles = []string{"{http.request.uri.path}"}` for consistency. (And maybe consider hoisting this to `Provision` cause there's no point doing this on every request). I don't think this "optimization" of directly using `r.URL.Path` is so valuable, cause it causes this edgecase with directories.

### DIFF
--- a/modules/caddyhttp/fileserver/matcher.go
+++ b/modules/caddyhttp/fileserver/matcher.go
@@ -140,6 +140,7 @@ func (m *MatchFile) Provision(_ caddy.Context) error {
 		m.Root = "{http.vars.root}"
 	}
 	// if list of files to try was omitted entirely, assume URL path
+	// (use placeholder instead of r.URL.Path; see issue #4146)
 	if m.TryFiles == nil {
 		m.TryFiles = []string{"{http.request.uri.path}"}
 	}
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/fileserver/matcher.go
```diff
@@ -139,6 +139,10 @@ func (m *MatchFile) Provision(_ caddy.Context) error {
 	if m.Root == "" {
 		m.Root = "{http.vars.root}"
 	}
+	// if list of files to try was omitted entirely, assume URL path
```
mholt, line 142: Could we add the issue number to this comment? It's not immediately obvious why the placeholder would be used instead of simply `r.URL.Path` like we were before.
```suggestion
	// if list of files to try was omitted entirely, assume URL path
	// (use placeholder instead of r.URL.Path; see issue #4146)
```

mholt (approved):

Issue #4148, opened 2021-05-06
Deprecate and remove common_log field from access logs and single_field encoder


mholt:
When Caddy 2 was being designed a couple years ago, there was still a lot of dependence on common-log-based workflows. Common Log Format (CLF) is bad, because it contains fields which are almost never used in modern HTTP, and it omits obvious fields like Host. This necessitates outdated workflows that separate each log emission into a separate file based on Host. And if you want to organize your logs by more than Host, or something other than Host, it gets even more complicated. And this does not scale well on the modern web, where most servers handle multiple domains -- some Caddy instances serve tens of thousands of hosts. CLF is too rigid to be useful in advanced deployments. It is also inefficient to generate CLF in Caddy's otherwise-zero-allocation logs, and impossible to filter them at emission-time because they're unstructured.

We make the case for [structured logging in our official documentation](https://caddyserver.com/docs/logging#structured-logs).

Caddy 2 uses JSON logs. To support CLF in Caddy 2, we construct the CLF line and add it to the JSON message emitted for each HTTP request as a separate field. This has the unfortunate side-effect of bloating each and every access log message, and also makes them harder to read/scan. It contains only duplicate information that has to be extracted to be parsed as a subtext anyway.

[A log encoder module, `single_field`, was created as a hack](https://caddyserver.com/docs/caddyfile/directives/log#single-field) to be able to write purely CLF-formatted log files without any JSON. This was a short-term solution, not a well-engineered one.

Nowadays, modules like the [format encoder](https://github.com/caddyserver/format-encoder) exist which can emit access logs in a custom format. They're not zero-allocation of course, but you can do what you want: including CLF if you *really* need it still.

The common log format is a thorn in the side of Caddy's logs, and I intend to:

- Remove the `common_log` field from access logs.
- Remove the `single_field` log encoder.
- Remove the `single_field common_log` example from the docs: https://caddyserver.com/docs/caddyfile/directives/log#examples

I already have considerable weight behind this motion, but in true OSS fashion I wanted to make my intentions clear and open it to discussion before finalizing these changes shortly after the v2.4 release.

francislavoie:
I agree with this change, but it will be a breaking change for some people who rely on it to feed into tools like fail2ban and such.

I think before we remove it from Caddy, we should have a clear and easy way to maintain this behaviour. Maybe this means adding a preset to the `format` encoder to do the same thing.

[...]

PR #4149, opened 2021-05-06
logging: Prep for `common_log` removal

francislavoie:
See https://github.com/caddyserver/caddy/issues/4148#issuecomment-833207811

With the upcoming removal of `common_log` from access logs, there's a few things we should do to smooth things out to offer a path for backwards compatibility.

- Add a `common_log` time format, which matches the format from the `{time.now.common_log}` placeholder which is currently used for the `common_log` field. This will make it easier for users of https://github.com/caddyserver/format-encoder to get the same time format by doing something like this:

      log {
	    format formatted <insert common log template here> {
          time_format common_log
        }
      }

- Add the `http.auth.user.id` placeholder's value to the logs. This currently isn't in the access logs, so if we drop `common_log`, that wouldn't be visible anywhere in the access logs, and therefore not be possible to recover using https://github.com/caddyserver/format-encoder. I went with `user_id` for the field name but it could just as easily be `auth_user_id`. 🤷‍♂️

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -184,8 +184,11 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				log = logger.Error
 			}
 
+			userID, _ := repl.GetString("http.auth.user.id")
+
 			log("handled request",
 				zap.String("common_log", repl.ReplaceAll(commonLogFormat, commonLogEmptyValue)),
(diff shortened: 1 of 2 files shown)


mholt (approved):
Good first steps. Thanks @Mohammed90 !

mholt:
Next steps would probably be to remove the common_log field from the access logs.

PR #4151, opened 2021-05-07
encode: Default to order the formats are enabled for `prefer` in Caddyfile

### LINKED ISSUE #4150: Allow for Caddyfile to implicitly set prefer value for encode directive
Creating an issue for a recent [discussion on encode directive docs PR for upcoming 2.4 release](https://github.com/caddyserver/website/pull/148#issuecomment-833096333).

---

### Proposal

For a Caddyfile, `encode.prefer` could be set implicitly.

```diff
- encode zstd gzip {
-  prefer zstd gzip
- }
+ encode zstd gzip
```

That would be intuitive to the user and consistent with the new `file_server.precompressed` feature.

### Justification

Without `prefer`, only the first supported client encoding is negotiated (often gzip).

#### Impact of implicit preference:

  - Backwards compatibility concern already applies to `encode` adopting sane defaults.
  - **Consistency:**
    - The new `file_server.precompressed` has implicit preference.
    - Minimizing config verbosity and good defaults are goals of Caddy.
  - **Caddy vs Clients:**
    - **Caddy:** Only `zstd` and `gzip` are officially supported (_50% chance `gzip` remains preferred_).
    - **Browser:** Not a single browser supports `zstd` officially.
    - **Other clients:** If `gzip` is supported it often took priority, even if unintentional (eg `cURL`).

francislavoie:
Fixes #4150 

/cc @ueffel @polarathene

I don't personally have a strong opinion on whether we should do this or not, but if we do, it should be done before v2.4.0. It's not complicated, so I figured I'd just write the PR to save time if we do decide to merge it.


mholt (approved):
Thanks Francis

PR #4155, opened 2021-05-09
caddypki: Add `--storage` option as override for `caddy trust`

francislavoie:
So the problem is that when you run Caddy as a systemd service, there's no good way to manually install the trust.

When running as a service, Caddy runs as the `caddy` user, which doesn't have `sudo` permissions, so trust won't install correctly.

If you try to run `sudo caddy trust`, then it'll use the `$HOME` of the `root` user instead, which ends up being `/root` (and therefore the storage location is `/root/.local/share/caddy`), which is not the right location in that situation, it should be `/var/lib/caddy/.local/share/caddy`.

One option is to run `sudo HOME=/var/lib/caddy caddy trust` but that's sorta awkward at a conceptual level.

So this new option lets you do `sudo caddy trust --storage /var/lib/caddy/.local/share/caddy`which I think makes a bit more sense.

---

Really I think we need to add a `--config` option to `caddy trust` so it can load trust from any arbitrary storage location, including redis or whatever. But I still think there's value in this override to exist regardless. 🤷‍♂️

### DIFF
--- a/modules/caddypki/command.go
+++ b/modules/caddypki/command.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	caddycmd "github.com/caddyserver/caddy/v2/cmd"
+	"github.com/caddyserver/certmagic"
 	"github.com/smallstep/truststore"
 )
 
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the PR, but:

> One option is to run `sudo HOME=/var/lib/caddy caddy trust` but that's sorta awkward at a conceptual level.

This is the right approach. Caddy will need to be run with root privileges but using the data directory of the user it will be running as, and that's an environment setting more than a storage setting. It's also unfortunate if `--storage` is limited to just file paths.

I will be closing this, as the implementation isn't quite right and the right way to do this is with the environment as you stated.

PR #4156, opened 2021-05-10
encode: Drop `prefer` from Caddyfile

francislavoie:
Followup to #4150, #4151 /cc @ueffel @polarathene

After a bit of discussion with @mholt, we decided to remove `prefer` as a subdirective and just go with using the order implicitly always. Simpler config, simpler docs, etc.


mholt:
Effectively reverts 77764714ad83d5c6e4db2da4b8945701f146bad0 and a small part of  f35a7fa466ffb06c38dcb3216e30c13aa8e14ce5.

mholt (approved):
This is much better. Thank you!

PR #4157, opened 2021-05-10
caddyfile(formatter): fix nesting not decrementing

matthewpi:
This is an extremely weird edge-case where if you had a environment variable {}
on one line, a comment on the next line, and the closing of the block on the
following line; the rest of the Caddyfile would be indented further than it
should've been.

ref; https://github.com/matthewpi/vscode-caddyfile-support/issues/13

---

A preview of what happens without this fix can be seen at https://streamable.com/ouziak

Formatting the file shown in the video worked fine before https://github.com/caddyserver/caddy/commit/635f075f187ac73c65e6939b8402494469a4d627

Additional testing (or even a better way to handle this edge-case) would be appreciated.

### DIFF
--- a/caddyconfig/caddyfile/formatter.go
+++ b/caddyconfig/caddyfile/formatter.go
@@ -43,6 +43,7 @@ func Format(input []byte) []byte {
 
 		newLines int // count of newlines consumed
 
+		wasComment bool // whether the last line was a comment
 		comment bool // whether we're in a comment
 		quoted  bool // whether we're in a quoted segment
 		escaped bool // whether current char is escaped
@@ -64,6 +65,7 @@ func Format(input []byte) []byte {
 	nextLine := func() {
 		write('\n')
 		beginningOfLine = true
+		wasComment = false
 	}
 
 	for {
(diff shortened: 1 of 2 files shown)


caddyconfig/caddyfile/formatter.go
```diff
@@ -165,6 +168,9 @@ func Format(input []byte) []byte {
 			}
 			continue
 
+		case nesting == 1 && wasComment && ch == '}':
```
francislavoie, line 171: Are you sure it shouldn't be `nesting > 0`? Could you add a test case with some subdirectives (or a `route`/`handle`) so it's more than one level deep?

mholt:
Thanks for the quick PR! I don't really understand the bug yet, but I wonder if there's a more elegant way to fix this.

matthewpi:
> Thanks for the quick PR! I don't really understand the bug yet, but I wonder if there's a more elegant way to fix this.
> 
> i.e. What is causing this condition `case nesting == 1 && wasComment && ch == '}':` to be so special... why is the nesting not getting decremented when it should be?

I don't understand it either tbh, but I think I figured it out now with the latest commit. (Apologies for the not so wonderful commit names)

PR #4164, opened 2021-05-12
ci: Run CI on PRs targeting minor version branches

francislavoie:
We decided that we'll use branches like `2.4` as the target for any changes that we might want to release in a `2.4.x` version like `2.4.1`, so that we can continue to merge changes targeting the next minor release (e.g. `2.5.0`) on master.

Our CI config wasn't set up for this to work properly though, since it was only running checks on PRs targeting master. See https://github.com/caddyserver/caddy/pull/4163 for example. This should fix it (then I'll rebase that PR so CI runs).

I couldn't find a way to do a pattern to only match digits for the branch names from Github's docs, it just looks like a pretty generic glob syntax. But this should do until we get to 3.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,11 @@ on:
   push:
     branches:
       - master
+      - 2.*
   pull_request:
     branches:
       - master
+      - 2.*
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


francislavoie:
I'll just go ahead and merge this. CI did run on this PR so it should be enough.

I need to submit another PR and I need it to run CI, so this is a mild blocker 😄

PR #4165, opened 2021-05-12
reverseproxy: Set the headers in the replacer before `handle_response`

francislavoie:
Turns out this was an oversight, we assumed we could use `{http.response.header.*}` but that doesn't work because those are grabbed from the response writer, and we haven't copied any headers into the response writer yet.

So the fix is to set all the response headers into the replacer at a new namespace before running the handlers.

This adds the `{http.reverse_proxy.header.*}` replacer.

See https://caddy.community/t/empty-http-response-header-x-accel-redirect/12447

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -120,9 +120,10 @@ type Handler struct {
 	// handler chain will not affect the health status of the
 	// backend.
 	//
-	// Two new placeholders are available in this handler chain:
-	// - `{http.reverse_proxy.status_code}` The status code
-	// - `{http.reverse_proxy.status_text}` The status text
+	// Three new placeholders are available in this handler chain:
+	// - `{http.reverse_proxy.status_code}` The status code from the response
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks, interesting situation... I guess this'll work.

francislavoie:
I think we'd go with `http.reverse_proxy.header.request_header.*` if we ever did, which matches the `request_header` handler.

corny:
I have another related issue: If the upstream server returns a `Content-Disposition` header, it won't be copied into the response.

francislavoie:
That's normal. No headers are copied by default in `handle_response`, you need to do it yourself for anything that's relevant. It just gives you a clean slate to re-handle the request based on the response from the upstream.

corny:
I see that `Set-Cookie`, `X-Request-Id` and a couple of other headers are copied. That's a little bit confusing, isn't it?

francislavoie:
Where do you see that @corny? The code doesn't copy any headers.

https://github.com/caddyserver/caddy/blob/2.4/modules/caddyhttp/reverseproxy/reverseproxy.go#L613-L652

PR #4179, opened 2021-05-25
fileserver: Redirect within the original URL

diamondburned:
This commit changes the file_server directive to redirect using the original request's URL instead of the possibly trimmed URL. This should make file_server work with handle_path.

This fix is taken from mholt's comment in https://caddy.community/t/file-servers-on-different-paths-not-working/11698/11.

---

For testing purposes, below Caddyfile should reproduce the before and afters:

```
{
	debug
	admin off
}

http://127.0.0.1:28475 {
	root * /home/diamond

	handle_path /browse* {
		file_server browse
	}

	respond "Hi"
}
```

When accessing `http://127.0.0.1:28475/browse`, the redirection would lead to `/`; after the changes, it would lead to `/browse/`.

### DIFF
--- a/modules/caddyhttp/fileserver/browse.go
+++ b/modules/caddyhttp/fileserver/browse.go
@@ -47,10 +47,10 @@ func (fsrv *FileServer) serveBrowse(root, dirPath string, w http.ResponseWriter,
 	// URL doesn't end in a trailing slash because hrefs like
 	// "/b/c" on a path like "/a" end up going to "/b/c" instead
 	// of "/a/b/c" - so we have to redirect in this case
-	if !strings.HasSuffix(r.URL.Path, "/") {
-		fsrv.logger.Debug("redirecting to trailing slash to preserve hrefs", zap.String("request_path", r.URL.Path))
-		r.URL.Path += "/"
-		http.Redirect(w, r, r.URL.String(), http.StatusMovedPermanently)
+	oldReq := r.Context().Value(caddyhttp.OriginalRequestCtxKey).(http.Request)
+	if !strings.HasSuffix(oldReq.URL.Path, "/") {
+		fsrv.logger.Debug("redirecting to trailing slash to preserve hrefs", zap.String("request_path", oldReq.URL.Path))
+		http.Redirect(w, r, oldReq.URL.Path+"/", http.StatusMovedPermanently)
 		return nil
 	}
 
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -243,12 +243,14 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 	// trailing slash - not enforcing this can break relative hrefs
 	// in HTML (see https://github.com/caddyserver/caddy/issues/2741)
 	if fsrv.CanonicalURIs == nil || *fsrv.CanonicalURIs {
-		if implicitIndexFile && !strings.HasSuffix(r.URL.Path, "/") {
-			fsrv.logger.Debug("redirecting to canonical URI (adding trailing slash for directory)", zap.String("path", r.URL.Path))
-			return redirect(w, r, r.URL.Path+"/")
-		} else if !implicitIndexFile && strings.HasSuffix(r.URL.Path, "/") {
-			fsrv.logger.Debug("redirecting to canonical URI (removing trailing slash for file)", zap.String("path", r.URL.Path))
-			return redirect(w, r, r.URL.Path[:len(r.URL.Path)-1])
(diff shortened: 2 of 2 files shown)


modules/caddyhttp/fileserver/browse.go
```diff
@@ -48,9 +48,10 @@ func (fsrv *FileServer) serveBrowse(root, dirPath string, w http.ResponseWriter,
 	// "/b/c" on a path like "/a" end up going to "/b/c" instead
 	// of "/a/b/c" - so we have to redirect in this case
 	if !strings.HasSuffix(r.URL.Path, "/") {
-		fsrv.logger.Debug("redirecting to trailing slash to preserve hrefs", zap.String("request_path", r.URL.Path))
-		r.URL.Path += "/"
-		http.Redirect(w, r, r.URL.String(), http.StatusMovedPermanently)
+		oldReq := r.Context().Value(caddyhttp.OriginalRequestCtxKey).(http.Request)
+
+		fsrv.logger.Debug("redirecting to trailing slash to preserve hrefs", zap.String("request_path", oldReq.URL.Path))
+		http.Redirect(w, r, oldReq.URL.Path+"/", http.StatusMovedPermanently)
```
francislavoie, line 54: I think we should probably keep the `.String()` call. This makes the redirect go to the fully qualified URL rather than just the absolute path. It'll make sure the path is properly escaped as well, etc.

francislavoie (commented):
Thanks for this! I'd been meaning to look into this at some point. Glad someone got to it before me :grin: 

FWIW, I realized this was a problem in https://caddy.community/t/route-directive-file-server-not-picking-up-index-html/12562 as well.

francislavoie:
I'm also worried about the situation where the original URL already has a `/` but the rewritten URL doesn't (can happen when the `rewrite` directive is used, instead of it being inside of a `handle_path`).

I think the original request path needs to be considered as well to not do a redirect when it doesn't make sense.

modules/caddyhttp/fileserver/browse.go
diamondburned: Isn't properly escaping guaranteed by `net/http` using `path.Clean`? I'm mainly doing this just to avoid copying a `url.URL` or mutating `oldReq` itself.

francislavoie: Yeah, you're probably right.

francislavoie:
> Do you mean both the original path and the new path, or just the original path? I've changed the code to check the original path instead; I think it makes more sense that way as well.

I was thinking both, but yeah I think that makes sense to only look at the original request, because the point of the canonicalization redirect is to make sure that _from the browser's perspective_, the URL does have a trailing `/` so that relative path links go to the right place. :+1:

modules/caddyhttp/fileserver/browse.go
mholt: Redirecting to `oldReq.URL.Path+"/"` will drop the query string and other URL components, though. I think the current way is more correct.

mholt: Nevermind, `http.Redirect()` apparently reconstructs the full URL given a relative target.

mholt (approved):
Thanks for this change, let's try it out.

Edit: This PR turned out to be a regression and has been reverted in v2.4.3. See #4205.

modules/caddyhttp/fileserver/browse.go
diamondburned: I think you're actually right, and I've only realized this just a few hours ago as well. Reading the `http.Redirect` code, I don't think it would reconstruct the URL with the old query given `oldReq.URL.Path+"/"`, so I should've given it `oldReq.URL.String()`.

I can hand in another PR if my guess is right.

mholt: Could you test it out? And if so, sure, another PR would be appreciated. 👍

diamondburned: I think you're right after all. With the path redirection, I'm not getting the query parameters back.

```
―❤―▶cat main.go
...
http.Redirect(w, r, r.URL.Path+"/", 200)
...

―❤―▶ curl -v localhost:8080/path?a=b
...
< HTTP/1.1 200 OK
< Location: /path/
< 
<a href="/path/">OK</a>.
```

With the `String()` redirection, I'm getting the query parameters as expected in the `Location` header.

```
―❤―▶cat main.go
...
r.URL.Path += "/"
http.Redirect(w, r, r.URL.String(), 200)
...

―❤―▶ curl -v localhost:8080/path?a=b
< HTTP/1.1 200 OK
< Location: /path/?a=b
< 
<a href="/path/?a=b">OK</a>.
```

diamondburned: I've filed PR #4196 for this.

PR #4192, opened 2021-06-05
Caddyfile support for preferred_chains

### LINKED ISSUE #4185: Support for `preferred_chains` in Caddyfile
As mentioned in #3854, Let's Encrypt now provides [two different](https://letsencrypt.org/2020/12/21/extending-android-compatibility.html) chains when acquiring a certificate: one with a cross-signed root certificate (default) and one with a self-signed root certificate.

I'd like to use the alternative chain, which seems to be possible using the `preferred_chains` option in Caddy. From what I've understood, though, this is currently only configurable through JSON.

Would it be possible to add a new global option in the Caddyfile to support changing the preferred chain without configuring Caddy using JSON?

Klooven:
This PR adds support for the `preferred_chains` option in the Caddyfile; both as an option for the ACME issuer and as a global option. If merged, this PR would resolve #4185.

---

The structure is the following:

```
preferred_chains [smallest] {
  root_common_name <common names...>
  any_common_name  <common names...>
}
```

The Caddyfile adapt tests added in this branch show how the options can be used both "locally" and globally. More details can be found in the issue linked above.

And, I'm of course happy to answer any questions that might arise 🙂


mholt (changes requested):
This looks pretty good. The only nit I have is that the parsing code is basically identical between that of the issuer config and the global option.

Can you consolidate that?

Since package `httpcaddyfile` imports package `caddytls`, maybe the parsing code for this could be in `caddytls` as its own function.

Klooven:
> This looks pretty good. The only nit I have is that the parsing code is basically identical between that of the issuer config and the global option.
> 
> Can you consolidate that?

You're right. I'll have a look at it and send in an update to the PR when I've changed that.

Klooven:
Ok, I managed to consolidate most of the stuff into a `ParseCaddyfilePreferredChainsOptions` function I created. Don't know if it's in the best place (I can move it if you feel like it should be in another place).

I encountered some issues on the way, so I don't know if the code is as elegant as it could be (comments welcome on this too). Did some testing and everything seemed to still work like before the "consolidation".

modules/caddytls/acmeissuer.go
```diff
@@ -436,6 +447,50 @@ func onDemandAskRequest(ask string, name string) error {
...
 }
 
+func ParseCaddyfilePreferredChainsOptions(d *caddyfile.Dispenser) (*ChainPreference, error) {
+	chainPref := new(ChainPreference)
+	if d.NextArg() {
+		smallestOpt := d.Val()
+		if smallestOpt == "smallest" {
+			trueBool := true
+			chainPref.Smallest = &trueBool
+			if d.NextBlock(1) {
```
Klooven, line 457: I used to use `if d.Next() {`, but it didn't work when calling this function through the ACME issuer config. Don't know why because it worked when this wasn't wrapped in a function: https://github.com/caddyserver/caddy/blob/ecc798af4c6b3370d2a7c5442e62afab81e03aba/modules/caddytls/acmeissuer.go#L407-L441

---

`if d.NextBlock(1) {` seems to work fine both when parsing the global option and the ACME issuer option. I'm a bit confused, to be honest...

mholt: Note [the proper way to use `NextBlock()`](https://pkg.go.dev/github.com/caddyserver/caddy/v2@v2.4.1/caddyconfig/caddyfile#Dispenser.NextBlock) is:

```
for nesting := d.Nesting(); d.NextBlock(nesting); {
}
```

especially for when you don't know the current level of nesting. Give that a try.

francislavoie: I think you can do `d.NextBlock(d.Nesting())` instead, removes the hardcoded number.

`NextBlock` essentially reads the next subdirective after the next `{` if any, until the next `}` is reached. It's usually used in loops like `for nesting := h.Nesting(); h.NextBlock(nesting); {`

mholt: Francis beat me to it. That's what I meant to say :smile:

francislavoie: Ah sorry, I think I misread what you're trying to do.

I think what you wanted was actually `d.NextArg()` to check that no other argument follows `smallest`. `d.Next()` will actually read the `{` which follows which is incorrect.

Klooven: In this case I want to check that when the user sets the `smallest` option, setting `root_common_name` or `any_common_name` **shouldn't** be possible (if I've understood correctly, these options can't be combined). Using `d.NextBlock(d.Nesting())` as you suggested seems to work great for this.

I could probably add another check with `d.NextArg()` to make sure no more arguments follow `smallest`. And I guess I should also add a check to make sure that the user has configured _something_, i.e. the following config should be invalid:

```
{
  preferred_chains
}
```

mholt: Perfect. Those ideas also sound good. There's a ton of places we don't do checks like those, but they add robustness to the parser and help prevent user errors ("I typed this and didn't get an error, but it's not working" kind of things), so they're a good thing if you think of them. :+1:

Klooven: Done and pushed!

mholt (approved):

mholt:
Great, looks good now! Thank you very much! Hope you'll contribute again.

PR #4196, opened 2021-06-07
fileserver: Fix browse not redirecting query parameters

diamondburned:
This commit is a follow up to PR #4179 that introduced a bug where browse redirections to the right URL would not preserve query parameters.

Below is the copy-pasted comment that describes the issue:

---

I think you're right after all. With the path redirection, I'm not getting the query parameters back.

```
―❤―▶ cat main.go
...
http.Redirect(w, r, r.URL.Path+"/", 200)
...

―❤―▶ curl -v localhost:8080/path?a=b
...
< HTTP/1.1 200 OK
< Location: /path/
< 
<a href="/path/">OK</a>.
```

With the `String()` redirection, I'm getting the query parameters as expected in the `Location` header.

```
―❤―▶ cat main.go
...
r.URL.Path += "/"
http.Redirect(w, r, r.URL.String(), 200)
...

―❤―▶ curl -v localhost:8080/path?a=b
< HTTP/1.1 200 OK
< Location: /path/?a=b
< 
<a href="/path/?a=b">OK</a>.
```

### DIFF
--- a/modules/caddyhttp/fileserver/browse.go
+++ b/modules/caddyhttp/fileserver/browse.go
@@ -50,7 +50,8 @@ func (fsrv *FileServer) serveBrowse(root, dirPath string, w http.ResponseWriter,
 	oldReq := r.Context().Value(caddyhttp.OriginalRequestCtxKey).(http.Request)
 	if !strings.HasSuffix(oldReq.URL.Path, "/") {
 		fsrv.logger.Debug("redirecting to trailing slash to preserve hrefs", zap.String("request_path", oldReq.URL.Path))
-		http.Redirect(w, r, oldReq.URL.Path+"/", http.StatusMovedPermanently)
+		oldReq.URL.Path += "/"
+		http.Redirect(w, r, oldReq.URL.String(), http.StatusMovedPermanently)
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/fileserver/browse.go
```diff
@@ -49,8 +49,9 @@ func (fsrv *FileServer) serveBrowse(root, dirPath string, w http.ResponseWriter,
 	// of "/a/b/c" - so we have to redirect in this case
 	oldReq := r.Context().Value(caddyhttp.OriginalRequestCtxKey).(http.Request)
 	if !strings.HasSuffix(oldReq.URL.Path, "/") {
+		oldReq.URL.Path += "/"
 		fsrv.logger.Debug("redirecting to trailing slash to preserve hrefs", zap.String("request_path", oldReq.URL.Path))
```
mholt, line 53: I think we might want to do the append second (i.e. switch these two lines, as in the original) -- so we log which path is being redirected from.

mholt (approved):
Thanks for double-checking me on this! Appreciate the patch.

PR #4207, opened 2021-06-14
caddyhttp: Refactor and export SanitizedPathJoin for use in fastcgi

mholt:
Help prevent directory traversal within PHP apps. Basically we pulled `sanitizedPathJoin()` out of the file server and put it into `caddyhttp` and exported it so any middleware can use it for safe path joins.

Still needs some testing on Windows CI and in actual PHP environments.

### DIFF
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -20,7 +20,9 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"path/filepath"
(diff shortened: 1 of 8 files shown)


francislavoie:
LGTM, but I'll play with it a bit this evening :+1:

francislavoie (approved):
Ran this build on Windows to check how the `env` looks with a `C:\` root, and the path is joined correctly.

LGTM! 🎉

PR #4215, opened 2021-06-18
encode: Tweak compression settings

klauspost:
zstd: Limit window sizes to 128K to keep memory in control both server and client size.
zstd: Write 0 length frames. This may be needed for compatibility.
zstd: Create fewer encoders. Small memory improvement.
gzip: Allow -2 (Huffman only) and -3 (stateless) compression modes.

### DIFF
--- a/modules/caddyhttp/encode/zstd/zstd.go
+++ b/modules/caddyhttp/encode/zstd/zstd.go
@@ -47,7 +47,8 @@ func (Zstd) AcceptEncoding() string { return "zstd" }
 
 // NewEncoder returns a new gzip writer.
 func (z Zstd) NewEncoder() encode.Encoder {
-	writer, _ := zstd.NewWriter(nil)
+	// Limit the window to 128K.
+	writer, _ := zstd.NewWriter(nil, zstd.WithWindowSize(128<<10), zstd.WithEncoderConcurrency(1), zstd.WithZeroFrames(true))
 	return writer
 }
 
--- a/modules/caddyhttp/encode/zstd/zstd.go
+++ b/modules/caddyhttp/encode/zstd/zstd.go
@@ -47,7 +47,9 @@ func (Zstd) AcceptEncoding() string { return "zstd" }
 
 // NewEncoder returns a new gzip writer.
 func (z Zstd) NewEncoder() encode.Encoder {
-	// Limit the window to 128K.
+	// The default of 8MB for the window is
+	// too large for many clients, so we limit
+	// it to 128K to lighten their load.
 	writer, _ := zstd.NewWriter(nil, zstd.WithWindowSize(128<<10), zstd.WithEncoderConcurrency(1), zstd.WithZeroFrames(true))
 	return writer
 }
--- a/modules/caddyhttp/encode/gzip/gzip.go
+++ b/modules/caddyhttp/encode/gzip/gzip.go
@@ -15,7 +15,6 @@
 package caddygzip
 
 import (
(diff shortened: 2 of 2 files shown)


modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -47,7 +47,8 @@ func (Zstd) AcceptEncoding() string { return "zstd" }
 
 // NewEncoder returns a new gzip writer.
 func (z Zstd) NewEncoder() encode.Encoder {
-	writer, _ := zstd.NewWriter(nil)
+	// Limit the window to 128K.
```
francislavoie, line 50: Maybe this comment could be elaborated upon to explain why this is helpful?

Would it be worth making this configurable? Is there a situation where a different value would be more optional?

klauspost: @francislavoie The thing is that it both affects the server and clients. So even if your server is doing fine, clients may have a hard time.

Yes, it will likely lower compression a bit for replies >128K, but since it is also affecting client side having a server-side switch is probably not that useful anyway.

In my test of [53927 typical web files](https://files.klauspost.com/compress/webdevdata.org-2015-01-07-4GB-subset.7z) the compressed size went from 755450358 -> 762046122 bytes, which is 100.87% the size. By default the limit is 8MB which will require the client to allocate at least that much.

francislavoie: Sounds good :+1:

I think clarifying the comment to say something to that effect would be helpful for future code readers. For example, maybe:

 ```suggestion
	// The default of 8MB for the window is
	// too large for many clients, so we limit
	// it to 128K to lighten their load
```

francislavoie (dismissed):

mholt (approved):
Ah, thanks. Always a pleasure to accept your contributions. Appreciate the work you do on your compress libs!

polarathene:
> zstd: Limit window sizes to 128K to keep memory in control both server and client size.

Is there any citation for clients that have trouble with the 8MB window size?

Since Chrome landed stable support for zstd content encoding in March 2024 releases of Chrome, they've been [following RFC 8878 guidance](https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1.1.2) and limiting the allowed window size to 8MB: https://github.com/caddyserver/caddy/pull/6140#issuecomment-2058021289

Their brotli support has a 16MB window, so for browsers there's likely no concern about the 8MB window limit? Where is the 128KB limit needed? Some embedded devices? Perhaps it'd be better to clarify/document that, potentially allowing for this to be configurable in some manner?

> zstd: Create fewer encoders. Small memory improvement.

Likewise it could be useful to support more than a single thread, although I guess larger static content could now be supported with pre-compressed zstd assets instead. This doesn't affect the client decoder right? Is the memory improvement likely to make that much of an improvement on the server at the expense of reduced encoding speed?

---

Granted that for many small files it's unlikely to matter much adjusting these. Just curious about these fixed settings and decisions behind them.

klauspost:
@polarathene 

>  Where is the 128KB limit needed? 

It is not needed - but you don't gain much from the additional window, so it is just to avoid mostly needless memory use, since most encodes will be small anyway.

> Likewise it could be useful to support more than a single thread

Likewise it is pointless in this context. There is currently limited concurrency for streaming encodes anyway - and it will just add overhead.

mholt:
Thanks for chiming in @klauspost ! Your info has been very helpful.

PR #4218, opened 2021-06-21
fileserver: Fix browse name_dir_first sorting

diamondburned:
This pull request fixes browse's default template to make `name_dir_first` sorting work.

The fix is basically a minor typo fix, changing `namedirfirst` in the template to `name_dir_first` to reflect `sortByNameDirFirst = "name_dir_first"`.

### DIFF
--- a/modules/caddyhttp/fileserver/browsetplcontext.go
+++ b/modules/caddyhttp/fileserver/browsetplcontext.go
@@ -264,7 +264,7 @@ func (l byTime) Less(i, j int) bool { return l.Items[i].ModTime.Before(l.Items[j
 
 const (
 	sortByName         = "name"
-	sortByNameDirFirst = "name_dir_first"
+	sortByNameDirFirst = "namedirfirst"
 	sortBySize         = "size"
 	sortByTime         = "time"
 )


francislavoie:
I think the better fix would be to change the `sortByNameDirFirst` variable's value to `namedirfirst`, because some users will have copied the template to customize it. Changing the variable would also fix it for those users, whereas changing the template in Caddy would only fix it for users who didn't copy the template.

francislavoie:
@diamondburned Do you have time to address my comment? Thanks!

diamondburned:
> @diamondburned Do you have time to address my comment? Thanks!

I've made this change, though I still have some second thoughts about it, mostly because the `browse` template that I wrote ended up using `name_dir_first` for the form, so existing templates would likely have to make the fix once this is merged.

francislavoie:
Well you're the first person whose reported it, so I have high doubts anyone else would have fixed it in the same way you did.

I'm not so concerned for people who may have fixed it in their own apps and not reported the issue, because that's not being a good user of open source software.

I'm more concerned about users who copied the template and changed _other_ things but left that the same, which will be the vast majority of people who copied templates in total. Like probably everyone except you in that group of people.

Thanks for reporting and it for the fix! 😀

francislavoie (approved):

Issue #2741, opened 2019-09-07
v2: Add configurable (auto)-adding of trailing slash for directories in path


Henrocker:
## 1. What would you like to have changed?
Since I enjoy the behaviour of Caddy v1 rewriting the url when accessing a directory in the webpath, I really like Caddy v2 to be able to do this, too. It would be awesome to have this configurable for when a webapp breaks.

## 2. Why is this feature a useful, necessary, and/or important addition to this project?
This makes configuration of Caddy v2 easier for users that would otherwise need to add specific rewrite rules. Making this a configurable option, one could turn on or off the automatic rewrite if issues appear using a specific app. 

## 3. What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?
Caddy v1 automatically adds a trailing slash to the request in case the path is a directory. This should be the most common behaviour for most apps, users and setups. 

## 4. Please link to any relevant issues, pull requests, or other discussions.
https://github.com/caddyserver/caddy/pull/2737#issuecomment-529115073 and https://github.com/caddyserver/caddy/pull/2737#issuecomment-529141600

mholt:
@Henrocker I've pushed this to the `v2` branch; URL canonicalization is enabled by default but it can be disabled in the file server. I've also updated the docs in the wiki. Thanks!

PR #4222, opened 2021-06-27
fileserver: Add `disable_canonical_uris` Caddyfile subdirective

mritd:
add 'canonical_uris' parameter to caddyfile

reference #2741

Signed-off-by: mritd <mritd@linux.com>

### DIFF
--- a/caddytest/integration/caddyfile_adapt/file_server_disable_canonical_uris.txt
+++ b/caddytest/integration/caddyfile_adapt/file_server_disable_canonical_uris.txt
@@ -1,0 +1,32 @@
+:80
+
+file_server {
+	disable_canonical_uris
+}
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":80"
+					],
+					"routes": [
+						{
+							"handle": [
+								{
+									"canonical_uris": false,
+									"handler": "file_server",
+									"hide": [
+										"./Caddyfile"
+									]
+								}
+							]
+						}
+					]
+				}
+			}
+		}
+	}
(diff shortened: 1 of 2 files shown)


francislavoie (changes requested):
Ah, good catch! This was an oversight when Caddyfile support was added to `file_server` early on it seems.

I think the UX of this option isn't consistent with other boolean options. Since `CanonicalURIs` being true is the same as being unset/nil (since the default is to enable the canonicalization redirects), I think the Caddyfile option should be called `disable_canonical_uris` and the argument should be removed.

Also, please add a Caddyfile adapt test (or update an existing one) to cover this option. See https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt

mritd:
Thanks for your reply, I have modified the subdirective UX, but I still have a question:

Is it necessary to add strict parameter verification here?

https://github.com/mritd/caddy/blob/737ead4e3de9b27d64f0b3dbd3e654149719b16c/modules/caddyhttp/fileserver/caddyfile.go#L117

``` go
case "disable_canonical_uris":
	if h.NextArg() {
		return nil, h.ArgErr()
	}
```

francislavoie:
Not necessary, but it's a nice to have to make sure invalid configs are rejected. If it's rejected, we know for sure nobody used an argument, so in the future if we want to add an argument for whatever reason, it wouldn't be a possibly breaking change.

mritd:
Okay thank you 😁

francislavoie (approved):

mholt (approved):
Thanks, this'll probably come in handy.

PR #4226, opened 2021-06-29
cmd: Implement `add-package` command

musinit:


francislavoie:
I'm not sure that I like the UX of this. `caddy add` is kinda strange.

Maybe like `caddy module add` and `caddy module remove` and so on would feel more natural.

mholt:
Technically, _packages_ get added and removed ("plugged in"), each of which can add one or more Caddy modules.

What about `caddy add-package` and `caddy remove-package`?

Mohammed90:
> Technically, _Go packages_ get added and removed ("plugged in"), each of which can add one or more Caddy modules.
> 
> What about `caddy add-package` and `caddy remove-package`?

But then the commands aren't consistent. It's `caddy list-modules` but `caddy {add,remove}-package`?

mholt:
> But then the commands aren't consistent. It's caddy list-modules but caddy {add,remove}-package?

It's rightly-named. A config module is a specific kind of plugin that extends the JSON config structure. A package is a Go concept that holds some code, which may contain one or more Caddy plugins (some of which may be modules; for example, config adapters aren't modules because they don't extend the JSON config structure, but they do add functionality so they are plugins). The `list-modules` command only lists config modules.

Unfortunately we cannot add or remove modules only, because they are registered by importing Go packages (but this is also a good thing, as it vastly simplifies this process and reduces possible errors -- since usually one module in a package uses other modules in the package).

That's why the Download page is organized by package, not module -- a package can contain multiple modules. By importing the package, you get them all (which is good, since some modules come as clusters and aren't useful in isolation).

francislavoie (commented):
Looks alright to me. But I think the opportunity could be taken to refactor this cause there's a lot of duplicated bits between this command and the upgrade one at this point.

cmd/commands.go
```diff
@@ -291,6 +291,15 @@ Downloads an updated Caddy binary with the same modules/plugins at the
 latest versions. EXPERIMENTAL: May be changed or removed.`,
 	})
 
+	RegisterCommand(Command{
+		Name:  "add-package",
+		Func:  cmdAddPackage,
+		Usage: "<packageName>",
+		Short: "Adds Caddy package (EXPERIMENTAL)",
+		Long: `
+	  Adds caddy package`,
```
francislavoie, line 300: Could we elaborate here a bit? Explain what happens, that existing non-standard packages will be retained, etc.

cmd/commandfuncs.go
```diff
@@ -40,6 +41,8 @@ import (
 	"go.uber.org/zap"
 )
 
+const DownloadPath = "https://caddyserver.com/api/download"
+
```
francislavoie, line 45: We tend to prefer putting consts at the bottom of the file. See `/caddy.go` for example, `var` and `const` at the end.

cmd/commandfuncs.go
```diff
@@ -588,16 +591,9 @@ func cmdUpgrade(_ Flags) (int, error) {
...
-	for _, mod := range nonstandard {
-		if mod.goModule.Replace != nil {
-			return caddy.ExitCodeFailedStartup, fmt.Errorf("cannot auto-upgrade when Go module has been replaced: %s => %s",
-				mod.goModule.Path, mod.goModule.Replace.Path)
-		}
-		l.Info("found non-standard module",
-			zap.String("id", mod.caddyModuleID),
-			zap.String("package", mod.goModule.Path))
-		pluginPkgs[mod.goModule.Path] = struct{}{}
+	pluginPkgs, err := buildPlugins(nonstandard)
```
francislavoie, line 594: Maybe call this function `getPluginPackages`? `build` is ambiguous here, my first thought is this would attempt to "build" at this point, i.e. compile a build of Caddy. Instead, it just makes a map, with some validation.

cmd/commandfuncs.go
```diff
@@ -657,61 +634,118 @@ func cmdUpgrade(_ Flags) (int, error) {
 	}()
 
 	// download the file; do this in a closure to close reliably before we execute it
-	writeFile := func() error {
-		destFile, err := os.OpenFile(thisExecPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, thisExecStat.Mode())
-		if err != nil {
-			return fmt.Errorf("unable to open destination file: %v", err)
-		}
-		defer destFile.Close()
+	err = writeFile(thisExecPath, &resp.Body, thisExecStat)
```
francislavoie, line 637: I think `writeFile` -> `writeCaddyBinary`? A bit more specific about what's being done.

francislavoie (changes requested):
Thanks for the extra polish! Just a few things

francislavoie:
Ah shit, I accidentally clicked "Update branch" sorry about that 😬 two big white buttons to click, "Approve and run" and "Update branch" and I clicked the wrong one. So you might need to force push to drop the merge commit.

francislavoie:
Also, you used `io/fs`, that's not safe to use yet because we still support Go 1.15, and `io/fs` is Go 1.16+. Please undo that change!

musinit:
> Ah shit, I accidentally clicked "Update branch" sorry about that 😬 two big white buttons to click, "Approve and run" and "Update branch" and I clicked the wrong one. So you might need to force push to drop the merge commit that I introduced.

Ahah, don't worry about this😄

cmd/commands.go
musinit: Can you help me please write it well, if current version is not good enough🙏

musinit:
@francislavoie check it out please😊

cmd/commands.go
```diff
@@ -291,6 +291,17 @@ Downloads an updated Caddy binary with the same modules/plugins at the
...
 
+	RegisterCommand(Command{
+		Name:  "add-package",
+		Func:  cmdAddPackage,
+		Usage: "<packageName>",
+		Short: "Adds Caddy package (EXPERIMENTAL)",
+		Long: `
+Downloads an updated Caddy binary with new module/plugin.
+If plugin is already added - then return error.
+`,
```
francislavoie, line 302: Maybe something like this:

```suggestion
		Short: "Add a package to Caddy (EXPERIMENTAL)",
		Long: `
Downloads an updated Caddy binary with the specified package (module/plugin)
added. Retains existing packages. Returns an error if the package is already
included. EXPERIMENTAL: May be changed or removed.
`,
```

cmd/commandfuncs.go
```diff
@@ -608,32 +601,13 @@ func cmdUpgrade(_ Flags) (int, error) {
...
 	}
-	urlStr := fmt.Sprintf("https://caddyserver.com/api/download?%s", qs.Encode())
 
 	// initiate the build
-	l.Info("requesting build",
-		zap.String("os", qs.Get("os")),
-		zap.String("arch", qs.Get("arch")),
-		zap.Strings("packages", qs["p"]))
-	resp, err := http.Get(urlStr)
+	resp, err := downloadModule(qs)
```
francislavoie, line 606: I think this should be `downloadBuild` since we're requesting a new build from the download server.

francislavoie (changes requested):
Getting pretty close.

cmd/commands.go
musinit: yeah, nice

francislavoie (approved):
I think this is a good start - we might make some additional changes later.

Thanks!

musinit:
Guys, I thought, maybe it will be cool to add here also remove-package command, looks like it's pretty close to add-package..but if you want I can remove it and make another PR.

francislavoie:
I'm thinking all the commands for packaging stuff (upgrade, add, remove) should be moved to a new file at this point, because it's getting pretty long in this file.

musinit:
@francislavoie check it out please, not sure is it cleaner to host these funcs in 3 different files, on in 1 file

francislavoie:
I think it's better if those three commands and their associated functions are in one file, since they represent one concept

musinit:
@francislavoie fixed!

cmd/commandfuncs.go
```diff
@@ -910,3 +713,5 @@ type moduleInfo struct {
 	goModule      *debug.Module
 	err           error
 }
+
+const DownloadPath = "https://caddyserver.com/api/download"
```
francislavoie, line 717: You can move this line over as well

francislavoie (changes requested):
Please also copy over the Copyright header from the other file - it should be included in every file.

cmd/commandfuncs.go
musinit: Done!

musinit:
@francislavoie check it out please!

francislavoie:
@musinit please have patience, I'd like Matt to review and he's a busy guy so it may take some time.

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,434 @@
...
+
+	err = destFile.Sync()
+	if err != nil {
+		return fmt.Errorf("syncing downloaded file to device: %v", err)
+	}
+
+	return nil
+}
+
+const DownloadPath = "https://caddyserver.com/api/download"
```
mholt, line 434: Unexport unless needing to be used outside of this package

```suggestion
const downloadPath = "https://caddyserver.com/api/download"
```

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,434 @@
...
+			if err2 != nil {
+				l.Error("restoring original executable failed; will need to be restored manually",
+					zap.String("backup_path", backupExecPath),
+					zap.String("original_path", thisExecPath),
+					zap.Error(err2))
+			}
+		}
+	}()
+
+	// download the file; do this in a closure to close reliably before we execute it
```
mholt, line 94: This comment doesn't make much sense after the refactoring

```suggestion
```

mholt (changes requested):
Thanks for working on this.

I haven't done a very thorough review yet (sorry, have been swamped with other things lately, and this is a big change) -- one thing I am noticing that is making this change look bigger than it probably is, is a lot of repeated code / logic. Remove and add (and maybe even upgrade) should all be very similar operations. Can the code be reused instead of repeated?

musinit:
> Thanks for working on this.
> 
> I haven't done a very thorough review yet (sorry, have been swamped with other things lately, and this is a big change) -- one thing I am noticing that is making this change look bigger than it probably is, is a lot of repeated code / logic. Remove and add (and maybe even upgrade) should all be very similar operations. Can the code be reused instead of repeated?

yeah, sure, I will refactor it

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,307 @@
...
+	pluginPkgs, err := getPluginPackages(nonstandard)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	return upgradeBuild(pluginPkgs)
+}
+
+func cmdAddPackage(fl Flags) (int, error) {
+	packageName := fl.Args()[0]
```
mholt, line 48: Is it guaranteed to have an argument? What about more than one? (It'd be nice if you could add/remove multiple packages at once.) (Same in cmdRemovePackage below)

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, fmt.Errorf("unable to enumerate installed plugins: %v", err)
+	}
+	pluginPkgs, err := getPluginPackages(nonstandard)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	if _, ok := pluginPkgs[packageName]; !ok {
+		// package does not exists
```
mholt, line 86:
```suggestion
		// package does not exist
```

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,307 @@
...
+	pluginPkgs, err := getPluginPackages(nonstandard)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	if _, ok := pluginPkgs[packageName]; ok {
+		// package already exists
+		return caddy.ExitCodeFailedStartup, fmt.Errorf("package is already added")
+	}
+
```
mholt, line 65:
```suggestion
```

mholt (changes requested):
Looking much better! Just a couple little things I noticed, then I'm good to merge this and try it out.

cmd/packagesfuncs.go
musinit: yeah, true, fixed!

mholt: Did you push the fix?

musinit:
@mholt please, take a look:)

mholt:
Sorry for the delay, will try to get back to this soon

cmd/commands.go
```diff
@@ -291,6 +291,30 @@ Downloads an updated Caddy binary with the same modules/plugins at the
 latest versions. EXPERIMENTAL: May be changed or removed.`,
 	})
 
+	RegisterCommand(Command{
+		Name:  "add-packages",
```
mholt, line 295:
```suggestion
		Name:  "add-package",
```

I think the singular form feels more natural, even if it allows multiple arguments.

cmd/commands.go
```diff
@@ -291,6 +291,30 @@ Downloads an updated Caddy binary with the same modules/plugins at the
...
+		Func:  cmdAddPackages,
+		Usage: "[<packageName>]",
```
mholt, line 297:
```suggestion
		Usage: "<packages...>",
```

cmd/commands.go
```diff
@@ -291,6 +291,30 @@ Downloads an updated Caddy binary with the same modules/plugins at the
...
+		Short: "Adds Caddy packages (EXPERIMENTAL)",
+		Long: `
+Downloads an updated Caddy binary with the specified packages (module/plugin)
+added. Retains existing packages. Returns an error if the any of packages are 
+already included. EXPERIMENTAL: May be changed or removed.
+`,
+	})
+
+	RegisterCommand(Command{
+		Name:  "remove-packages",
```
mholt, line 307:
```suggestion
		Name:  "remove-package",
```

cmd/commands.go
```diff
@@ -291,6 +291,30 @@ Downloads an updated Caddy binary with the same modules/plugins at the
...
+		Func:  cmdRemovePackages,
+		Usage: "[<packageName>]",
```
mholt, line 309:
```suggestion
		Usage: "<packages...>",
```

cmd/packagesfuncs.go
mholt: This is still not fixed.

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,308 @@
...
+			return caddy.ExitCodeFailedStartup, fmt.Errorf("package is already added")
+		}
+		pluginPkgs[arg] = struct{}{}
+	}
+
+	return upgradeBuild(pluginPkgs)
+}
+
+func cmdRemovePackages(fl Flags) (int, error) {
+	packageName := fl.Args()[0]
```
mholt, line 72: This still only accepts one argument and will crash if there aren't any.

mholt (changes requested):
Thanks for working on this, some more work is needed; please check out the comments.

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	pluginPkgs, err := getPluginPackages(nonstandard)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	return upgradeBuild(pluginPkgs)
+}
+
+func cmdAddPackages(fl Flags) (int, error) {
```
mholt, line 47:
```suggestion
func cmdAddPackage(fl Flags) (int, error) {
```

cmd/packagesfuncs.go
```diff
@@ -0,0 +1,308 @@
...
+		if _, ok := pluginPkgs[arg]; ok {
+			return caddy.ExitCodeFailedStartup, fmt.Errorf("package is already added")
+		}
+		pluginPkgs[arg] = struct{}{}
+	}
+
+	return upgradeBuild(pluginPkgs)
+}
+
+func cmdRemovePackages(fl Flags) (int, error) {
```
mholt, line 71:
```suggestion
func cmdRemovePackage(fl Flags) (int, error) {
```

cmd/packagesfuncs.go
musinit: sorry, fixed!

musinit:
@mholt sorry for long delay and for so many error/fixes, I really want to finish that PR and hope you won't get mad in the end :D

mholt:
Haha we won't get mad.

Ok now we're talkin'... I need to look closer when I'm not mobile but this looks a lot better.

mholt (approved):
Thank you for working on this so diligently -- I'm sure some minor improvements will be needed with time, so please pay attention if you notice anything. Thanks for contributing!

PR #4232, opened 2021-07-05
reverseproxy: Keep path to unix socket as dial address

### LINKED ISSUE #4231: Active health check not working with reverse proxy to unix domain socket
Hi,

I am running Caddy v2.4.3 on FreeBSD 12.2

I have an entry in my Caddyfile that looks like this (obviously the domain names and the Cloudflare key have been substituted with placeholder values here, but other than that I am showing exactly what I have):

```text
api.example.com {
    reverse_proxy {
        to 10.13.37.7 unix//var/www/com.example.api/example-http-api/run/public.sock
        lb_policy first
	health_uri /status
        health_interval 5s
        health_timeout 450ms
    }

    tls hostmaster@example.com {
        dns cloudflare OwYPs2n3yPpmLwwOGM6ufn4gQwDRmYAl9BaqL0AS
    }
}
```

However, in `/var/log/caddy.log`, I get errors that look like the following:

```
{"level":"info","ts":1625231212.875412,"logger":"http.handlers.reverse_proxy.health_checker.active","msg":"HTTP request failed","host":"localhost","error":"Get \"http://localhost/status\": dial unix localhost: connect: no such file or directory"}
```

Somehow when it tries to perform active health check, it tries to connect to `http://localhost/status`, instead of talking to the Unix domain socket.

Whereas with passive health checks instead of active health checks, I am able to service requests both with the primary destination and with falling back to the Unix domain socket destination when the primary returns an empty response.

```text
api.example.com {
    reverse_proxy {
        to 10.13.37.7 unix//var/www/com.example.api/example-http-api/run/public.sock
        lb_policy first
        fail_duration 30s
    }

    tls hostmaster@example.com {
        dns cloudflare OwYPs2n3yPpmLwwOGM6ufn4gQwDRmYAl9BaqL0AS
    }
}
```

However, I need active health checks because when the primary is not responding then the passive health checks do not trigger.

Am I doing something wrong here? I am not able to tell from the documentation at https://caddyserver.com/docs/caddyfile/directives/reverse_proxy

elnoro:
Hello, 

This is a fix for https://github.com/caddyserver/caddy/issues/4231. I tried changing the Path variable in different ways but it did not help. 

It seems http transport relies on the address in the dial info context var which caused the issue (https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/httptransport.go#L181). 

I have also added a test for healthchecks to reproduce the bug described in the issue.

### DIFF
--- a/caddytest/integration/reverseproxy_test.go
+++ b/caddytest/integration/reverseproxy_test.go
@@ -436,3 +436,57 @@ func TestReverseProxyHealthCheckUnixSocket(t *testing.T) {
 
 	tester.AssertGetResponse("http://localhost:9080/", 200, "Hello, World!")
 }
+
+func TestReverseProxyHealthCheckUnixSocketWithoutPort(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.SkipNow()
+	}
+	tester := caddytest.NewTester(t)
+	f, err := ioutil.TempFile("", "*.sock")
+	if err != nil {
+		t.Errorf("failed to create TempFile: %s", err)
+		return
+	}
+	// a hack to get a file name within a valid path to use as socket
+	socketName := f.Name()
+	os.Remove(f.Name())
+
+	server := http.Server{
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			if strings.HasPrefix(req.URL.Path, "/health") {
+				w.Write([]byte("ok"))
+				return
+			}
+			w.Write([]byte("Hello, World!"))
+		}),
+	}
+
+	unixListener, err := net.Listen("unix", socketName)
+	if err != nil {
+		t.Errorf("failed to listen on the socket: %s", err)
+		return
+	}
+	go server.Serve(unixListener)
+	t.Cleanup(func() {
+		server.Close()
+	})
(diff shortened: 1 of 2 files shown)


francislavoie:
@ctsrc could you please try out this fix to confirm that it solves the problem for you?

You'll need to build Caddy from source. Clone the repo and checkout this branch `elnoro-fix4321` and then run:
```
$ cd cmd/caddy/
$ go build
```

If you need plugins, you can add them in `cmd/caddy/main.go` as imports in there.

Thanks!

ctsrc:
@francislavoie I will try it later today and let you know

Mohammed90 (approved):
Thanks for doing this! This fix is more consistent with the documentation comment on the `Address` field of `DialInfo` than before. It says the field follows the same semantics of the address parameter of `net.Dial`, which in turn expects path on the filesystem for Unix sockets.

ctsrc:
@francislavoie I tested it now and this fix by @elnoro solves the problem for me 👍🏻 

@elcore thank you very much for fixing this problem 😊

mholt (approved):
Thanks, this is a subtle difference, appreciate the change.

PR #4233, opened 2021-07-05
caddyfile: keep error chain info in Dispenser.Errf

ggicci:
Use `fmt.Errorf` instead of `fmt.Sprintf` in `Dispenser.Errf` method to reserve the error chain information. Make the underlying error unwrapable by using `errors.Is` or `errors.As` helpers.

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -343,15 +343,28 @@ func (d *Dispenser) EOFErr() error {
 	return d.Errf("Unexpected EOF")
 }
 
+type parseError struct {
+	file       string
+	line       int
+	innerError error
+}
+
+func (e parseError) Error() string {
+	return fmt.Sprintf("%s:%d - Error during parsing: %s", e.file, e.line, e.innerError.Error())
+}
+
+func (e parseError) Unwrap() error {
+	return e.innerError
+}
+
 // Err generates a custom parse-time error with a message of msg.
 func (d *Dispenser) Err(msg string) error {
-	msg = fmt.Sprintf("%s:%d - Error during parsing: %s", d.File(), d.Line(), msg)
-	return errors.New(msg)
+	return parseError{d.File(), d.Line(), errors.New(msg)}
 }
 
 // Errf is like Err, but for formatted error messages
 func (d *Dispenser) Errf(format string, args ...interface{}) error {
-	return d.Err(fmt.Sprintf(format, args...))
+	return parseError{d.File(), d.Line(), fmt.Errorf(format, args...)}
 }
 
 // Delete deletes the current token and returns the updated slice
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -343,28 +343,15 @@ func (d *Dispenser) EOFErr() error {
 	return d.Errf("Unexpected EOF")
 }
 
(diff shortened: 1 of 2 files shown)


mholt:
Thanks for this. What is the use case for this? And why do we have to define our own error type instead of using `%w` to do it for us?

ggicci:
> What is the use case for this? (There's no context / issue associated.)

This API `Dispenser.Errf` is useful when extending the `Caddyfile` parser. When I was writing the [Caddyfile parser](https://github.com/ggicci/caddy-jwt/blob/main/caddyfile.go#L35) for [ggicci/caddy-jwt](https://github.com/ggicci/caddy-jwt.git), I found that it's not "easy" to check the error in a test. I can't use `errors.Is` nor `errors.As` to reach some more specific error information in the error chain. The root error may be wrapped multiple times during the parsing stuff. For example:

In the codebase at this time, I returned an error object as follows:

```go
func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error) {
    // ...
    return nil, h.Errf("invalid sign_key")
}
```

And in the corresponding test, I wrote:

```go
func TestParsingCaddyfileError(t *testing.T) {
    // ...
    _, err := parseCaddyfile(helper)
    assert.Contains(t, err.Error(), "sign_key")  // fragile, broken when error content changed
}
```

Assertion statements checking some error information by string comparison can be fragile.

If `Dispenser.Errf` can wrap errors, It can help writing more confident code as follows:

```go
var ErrInvalidSignKey = errors.New("invalid sign_key")
func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error) {
    // ...
    return nil, h.Errf("%w", ErrInvalidSignKey)
}

func TestParsingCaddyfileError(t *testing.T) {
    // ...
    _, err := parseCaddyfile(helper)
    assert.ErrorIs(t, err, ErrInvalidSignKey)  // more stable
}
```

> And why do we have to define our own error type instead of using %w to do it for us?

1. It does no harm to performance nor resource usage of the program.
2. The code can be more readable.

caddyconfig/caddyfile/dispenser.go
```diff
@@ -343,15 +343,28 @@ func (d *Dispenser) EOFErr() error {
...
+
+func (e parseError) Unwrap() error {
+	return e.innerError
+}
+
 // Err generates a custom parse-time error with a message of msg.
 func (d *Dispenser) Err(msg string) error {
-	msg = fmt.Sprintf("%s:%d - Error during parsing: %s", d.File(), d.Line(), msg)
-	return errors.New(msg)
+	return parseError{d.File(), d.Line(), errors.New(msg)}
```
mholt, line 362: If Err could use Errf, or vice-versa, that might be nice to reduce a little bit of code duplication / repetition.

mholt (commented):
Gotcha, thanks for the explanation.

If using %w would achieve the same result, I think I'd rather use that just for simplicity's sake; I'm not sure what about that is harder to read than having a whole new type defintion to manually wrap the error. But if this is the only way to get the result you want, then that's OK I guess.

caddyconfig/caddyfile/dispenser.go
ggicci: I refactored it and I think it's more concise now.

mholt (approved):
LGTM now -- thanks!

PR #4243, opened 2021-07-11
caddyhttp: Fix edgecase with HTTP->HTTPS logic

francislavoie:
Followup to https://github.com/caddyserver/caddy/pull/4033, see https://caddy.community/t/how-to-set-priority-in-caddyfile/13002/8 for context

As it turns out, my earlier change for HTTP->HTTPS redirect was not enough, because it left one edgecase.

If none of the routes in the HTTP server had host matchers, then it would return the index of the last route, even if the only route was a catch-all (which is defined as having no host matcher).

This means that a config like this would not work correctly, and `Foo` would be returned on requests to `http://bar.localhost/`

```
http:// {
	respond "Foo"
}
bar.localhost {
	respond "Bar"
}
```

What we want is for `http://bar.localhost/` to return a redirect to `https://bar.localhost/` instead (and this can be overridden by using `auto_https disable_redirects` if you want it otherwise).

So the fix I went with is to keep track of whether we actually found a host matcher during the loop, and if we didn't, then return `0` as the index to insert the redirect routes, i.e. before any user-defined catch-all.

P.S. Integration tests don't actually run in CI because they're flaky and slow, but I ran them all again with my change and the new test to cover this case, and it passes.

### DIFF
--- a/caddytest/integration/autohttps_test.go
+++ b/caddytest/integration/autohttps_test.go
@@ -103,3 +103,23 @@ func TestAutoHTTPRedirectsInsertedBeforeUserDefinedCatchAll(t *testing.T) {
 	tester.AssertGetResponse("http://foo.localhost:9080/", 200, "Foo")
 	tester.AssertGetResponse("http://baz.localhost:9080/", 200, "Baz")
 }
+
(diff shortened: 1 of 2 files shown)


mholt (approved):
Let's give this a try - thank you Francis!

PR #4244, opened 2021-07-12
logging: Add missing interface guards for replace filter

leodido:
While looking at the code base I noticed that probably interface guards for `ReplaceFilter` were missing.


francislavoie (approved):
Thanks! For ref, this feature was added in https://github.com/caddyserver/caddy/pull/4029 but the guards were missed. :+1:

Issue #4106, opened 2021-04-05
Feature: Add ability to periodically pull config


YourTechBud:
This is a follow up request to #3994 -  Dynamic config loading at startup.

Currently the config loading is a startup activity. As suggested in the [PR](https://github.com/caddyserver/caddy/pull/3994#issuecomment-806671858), it would be great if we could pull this config periodically at a configurable time interval.

YourTechBud:
Wanted to quickly follow up on this since I needed this feature for the project I've been working on. I can spare some time to cough up a PR over the weekend. I had two proposals in mind

## 1] Change the function signature of the `LoadModule` method

This one is a breaking change but seems to make the most sense. Currently the signature of the `LoadModule` method is 
```go
func (cl ConfigLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
  var config []byte
  // Load the marshaled json config in the `config` variable
  return config, nil
}
````

I suggest we change it to
```go
func (cl ConfigLoader) LoadConfig(ctx caddy.Context) (chan []byte, error) {
  configC := make(chan []byte, 0) // It's important for the channel buffer size to be 0
  // Write to the configC channel as many times as necessary in a separate goroutine
  return configC, nil
}
```
In the admin module we can change the caller of this method to look something like this
```go
	// if dynamic config is requested, set that up and run it
	if cfg != nil && cfg.Admin != nil && cfg.Admin.Config != nil && cfg.Admin.Config.LoadRaw != nil {
		val, err := ctx.LoadModule(cfg.Admin.Config, "LoadRaw")
		if err != nil {
			return fmt.Errorf("loading config loader module: %s", err)
		}
		loadedConfigC, err := val.(ConfigLoader).LoadConfig(ctx)
		if err != nil {
			return fmt.Errorf("loading dynamic config from %T: %v", val, err)
		}

		// do this in a goroutine so current config can finish being loaded; otherwise deadlock
		go func() {
                        for loadedConfig := range loadedConfigC {
			        Log().Info("applying dynamically-loaded config", zap.String("loader_module", val.(Module).CaddyModule().ID.Name()))
			        currentCfgMu.Lock()
			        err := unsyncedDecodeAndRun(loadedConfig, false)
			        currentCfgMu.Unlock()
			        if err == nil {
				        Log().Info("dynamically-loaded config applied successfully")
			        } else {
				        Log().Error("running dynamically-loaded config failed", zap.Error(err))
			        }
                        }
		}()
	}
```
[...]

PR #4246, opened 2021-07-19
admin: Implement a simple timer to pull config

colinaaa:
Implemented mostly referenced to the #4106

- add a `PullInterval` filed in config
- setup a timer in a goroutine
- `select` with context and timer

A trick point is that

https://github.com/colinaaa/caddy/blob/169bb265104d0ade5f8bab05d82e53eef30c0129/caddy.go#L485-L494

here I did not use if-else to check whether `PullInterval` is non-zero, but I directly use `time.After(PullInterval)` so that with 0 intervals. 
This will return immediately which is the same as executing without `time.After`

re #4106

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -109,6 +109,12 @@ type ConfigSettings struct {
 	//
 	// EXPERIMENTAL: Subject to change.
 	LoadRaw json.RawMessage `json:"load,omitempty" caddy:"namespace=caddy.config_loaders inline_key=module"`
+
+	// The interval to pull config. With a non-zero value, will pull config
+	// from config loader (eg. a http loader) with given interval.
+	//
+	// EXPERIMENTAL: Subject to change.
+	PullInterval time.Duration `json:"pull_interval,omitempty"`
 }
 
 // IdentityConfig configures management of this server's identity. An identity
--- a/admin.go
+++ b/admin.go
@@ -114,7 +114,7 @@ type ConfigSettings struct {
 	// from config loader (eg. a http loader) with given interval.
 	//
 	// EXPERIMENTAL: Subject to change.
-	PullInterval Duration `json:"pull_interval,omitempty"`
+	LoadInterval Duration `json:"pull_interval,omitempty"`
 }
 
 // IdentityConfig configures management of this server's identity. An identity
--- a/admin.go
+++ b/admin.go
@@ -114,7 +114,7 @@ type ConfigSettings struct {
 	// from config loader (eg. a http loader) with given interval.
 	//
 	// EXPERIMENTAL: Subject to change.
-	PullInterval caddy.Duration `json:"pull_interval,omitempty"`
+	PullInterval Duration `json:"pull_interval,omitempty"`
 }
 
 // IdentityConfig configures management of this server's identity. An identity
--- a/admin.go
+++ b/admin.go
@@ -114,7 +114,7 @@ type ConfigSettings struct {
 	// from config loader (eg. a http loader) with given interval.
 	//
 	// EXPERIMENTAL: Subject to change.
-	PullInterval time.Duration `json:"pull_interval,omitempty"`
+	PullInterval caddy.Duration `json:"pull_interval,omitempty"`
 }
 
 // IdentityConfig configures management of this server's identity. An identity
--- a/admin.go
+++ b/admin.go
@@ -114,7 +114,7 @@ type ConfigSettings struct {
 	// from config loader (eg. a http loader) with given interval.
 	//
 	// EXPERIMENTAL: Subject to change.
-	LoadInterval Duration `json:"pull_interval,omitempty"`
+	LoadInterval Duration `json:"load_interval,omitempty"`
 }
 
 // IdentityConfig configures management of this server's identity. An identity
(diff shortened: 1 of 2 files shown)


mholt:
Awesome, thanks for working on this! I'll take a look shortly.

colinaaa:
Here is an example `config.json` for pull interval. Use another http server to serve this file at port 5000.
```json
{
  "admin": {
    "listen": "0.0.0.0:2019",
    "config": {
      "pull_interval": 3000000000,
      "load": { "module": "http", "url": "http://localhost:5000/config.json" }
    }
  },
  "apps": {
    "http": {
      "servers": {
        "local": {
          "listen": [":8080"]
        }
      }
    }
  }
}
```

mholt (changes requested):
Thanks, this is a good looking PR.

The only thing I'm not real sure about is the fact that all config loading is now asynchronous. That does make me a little nervous since the vast majority of users won't use the pull feature, and would probably expect synchronous config loading. I am not sure of the consequences if we unblock while the config is loaded...

I do appreciate the simplicity of this approach, but I'm not sure what kinds of effects it will have. What do you think?

colinaaa:
> The only thing I'm not really sure about is the fact that all config loading is now asynchronous. That does make me a little nervous since the vast majority of users won't use the pull feature, and would probably expect synchronous config loading. I am not sure of the consequences if we unblock while the config is loaded...
> 
> I do appreciate the simplicity of this approach, but I'm not sure what kinds of effects it will have. What do you think?

I agree that load config asynchronously is a breaking change. I'm doing this because it's the simplest way to implement this feature.

I will update this pr to keep loading config synchronously when no interval is specified as well as keep the code simple and clear.

admin.go
```diff
@@ -109,6 +109,12 @@ type ConfigSettings struct {
 	//
 	// EXPERIMENTAL: Subject to change.
 	LoadRaw json.RawMessage `json:"load,omitempty" caddy:"namespace=caddy.config_loaders inline_key=module"`
+
+	// The interval to pull config. With a non-zero value, will pull config
+	// from config loader (eg. a http loader) with given interval.
+	//
+	// EXPERIMENTAL: Subject to change.
+	PullInterval caddy.Duration `json:"pull_interval,omitempty"`
```
francislavoie, line 117: This should be just `Duration` because we're in the `caddy` package already in here

```suggestion
	PullInterval Duration `json:"pull_interval,omitempty"`
```

caddy.go
```diff
@@ -480,21 +481,29 @@ func finishSettingUp(ctx Context, cfg *Config) error {
...
-			err := unsyncedDecodeAndRun(loadedConfig, false)
-			currentCfgMu.Unlock()
-			if err == nil {
-				Log().Info("dynamically-loaded config applied successfully")
-			} else {
-				Log().Error("running dynamically-loaded config failed", zap.Error(err))
+			select {
+			// if PullInterval is positive, will wait for the interval and then run with new config
+			// otherwise, time.After(0) will return immediately, and run with new config
+			case <-time.After(cfg.Admin.Config.PullInterval):
```
francislavoie, line 489: There's still an issue here, CI failing. I think `PullInterval` needs to be wrapped with a `time.Duration()` to type convert it to a `time.Duration`.

admin.go
```diff
@@ -109,6 +109,12 @@ type ConfigSettings struct {
 	//
 	// EXPERIMENTAL: Subject to change.
 	LoadRaw json.RawMessage `json:"load,omitempty" caddy:"namespace=caddy.config_loaders inline_key=module"`
+
+	// The interval to pull config. With a non-zero value, will pull config
+	// from config loader (eg. a http loader) with given interval.
+	//
+	// EXPERIMENTAL: Subject to change.
+	PullInterval Duration `json:"pull_interval,omitempty"`
```
mholt, line 117: On second thought... this affects how often the function specified by the `load` property is run. So I wonder if, for consistency, we should call this `LoadInterval` instead. Does that make sense?

mholt:
Nice, this is a lot more like what I had envisioned. Does it work well? If so, there's just the matter of deciding the name for the property. Then we can merge this.

admin.go
colinaaa: Sounds good to me!

colinaaa:
> Does it work well?

It works as expected with the config I have given at https://github.com/caddyserver/caddy/pull/4246#issuecomment-883006200

And I tried another config without `load_interval`, which work as expected too!

caddy.go
```diff
@@ -268,7 +268,8 @@ func unsyncedDecodeAndRun(cfgJSON []byte, allowPersist bool) error {
 		newCfg != nil &&
 		newCfg.Admin != nil &&
 		newCfg.Admin.Config != nil &&
-		newCfg.Admin.Config.LoadRaw != nil {
+		newCfg.Admin.Config.LoadRaw != nil &&
+		newCfg.Admin.Config.LoadInterval <= 0 {
 		return fmt.Errorf("recursive config loading detected: pulled configs cannot pull other configs")
```
mholt, line 273:
```suggestion
		return fmt.Errorf("recursive config loading detected: pulled configs cannot pull other configs without load_interval")
```

mholt (approved):
Awesome, thanks for working on this and trying it out!

I had one more suggestion that I just noticed; otherwise, LGTM and let's merge this and give it a try.

mholt (approved):

Issue #4256, opened 2021-07-29
Erratic logging behavior with wildcard domains


ghost:
As I understand it, access logging is disabled until you add the `log` directive to each site you want to log. However, if I add the `log` directive to _any_ site, all my wildcard domains start being logged too, even though they are in separate blocks and I don't want them to be logged. The only exception to this is when I enable `log` (only) in the site block for an IP address (port 80), then it does not cause the wildcard domains to log. So it seems there is a bug.

francislavoie:
Thanks for opening an issue! We'll look into this.

It's not immediately clear to us what is going on, so we'll need your help to understand it better.

Ideally, we need to be able to reproduce the bug _in the most minimal way possible_. This allows us to write regression tests to verify the fix is working. If we can't reproduce it, then you'll have to test our changes for us until it's fixed -- and then we can't add test cases, either.

I've attached a template below that will help make this easier and faster! It will ask for some information you've already provided; that's OK, just fill it out the best you can. :+1:

I've also included some helpful tips below the template. Feel free to let me know if you have any questions!

Thank you again for your report, we look forward to resolving it!

## Template

<pre><code>## 1. Environment

### 1a. Operating system and version

```
paste here
```


### 1b. Caddy version (run `caddy version` or paste commit SHA)

```
paste here
```


### 1c. Go version (if building Caddy from source; run `go version`)

```
paste here
```


## 2. Description

### 2a. What happens (briefly explain what is wrong)




### 2b. Why it's a bug (if it's not obvious)




### 2c. Log output

```
paste terminal output or logs here
```



### 2d. Workaround(s)




### 2e. Relevant links




## 3. Tutorial (minimal steps to reproduce the bug)




</code></pre>

## Helpful tips

1. **Environment**: Please fill out your OS and Caddy versions, even if you don't think they are relevant. (They are _always_ relevant.) If you built Caddy from source, provide the commit SHA and specify your exact Go version.

2. **Description:** Describe at a high level what the bug is. What happens? Why is it a bug? Not all bugs are obvious, so convince readers that it's actually a bug.
	- **2c) Log output:** Paste terminal output and/or complete logs in a code block. [DO NOT REDACT INFORMATION except for credentials.](https://caddy.community/t/how-to-get-help-with-caddy-more-effectively/5222)
	- **2d) Workaround:** What are you doing to work around the problem in the meantime? This can help others who encounter the same problem, until we implement a fix.
[...]

PR #4258, opened 2021-07-31
httpcaddyfile: Ensure hosts to skip for logs can always be collected

m90:
Previously, some hosts that should be skipped in logging would
be missed as the current logic would only collect them after
encountering the first server that would log. This change makes sure
the ServerLogConfig is initialized before iterating over the server
blocks.

---

This connects: https://github.com/caddyserver/caddy/issues/4256#issuecomment-889676939

The condition this is fixing looks like:

- When parsing the Caddyfile, the server blocks get sorted in a different order than the one they are using in the Caddyfile
- The first server block that gets processed when calling `Setup` does not have a `log` directive
- `srv.Logs` therefore is `nil` on this iteration
- This condition does not pass https://github.com/caddyserver/caddy/blob/c131339c5cda3a541223fde4a714aab55de13b9a/caddyconfig/httpcaddyfile/httptype.go#L664 and the hostnames do not get appended to the skip list
- The next server block has a `log` directive, creating the skip list
- The next server block without a `log` directive gets appended to the skip list
- The skip list in incomplete, resulting in unwanted logging

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -522,6 +522,16 @@ func (st *ServerType) serversFromPairings(
 			}
 		}
 
+		// if needed, the ServerLogConfig is initialized beforehand so
+		// that all server blocks can populate it with data, even when not
+		// coming with a log directive
+		for _, sblock := range p.serverBlocks {
+			if len(sblock.pile["custom_log"]) != 0 {
+				srv.Logs = new(caddyhttp.ServerLogConfig)
+				break
+			}
+		}
+
 		// create a subroute for each site in the server block
 		for _, sblock := range p.serverBlocks {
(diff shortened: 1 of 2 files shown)


francislavoie:
Thanks! This looks good.

Could you add a Caddyfile adapt test for this? See https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt

m90:
@francislavoie good idea, I wasn't entirely sure where to put such a test, but now I do. Added one and also checked that it would fail on current `master`. Let me know if this is ok for you.

francislavoie (approved):
Looks good! Thanks!

mholt (approved):
Thanks for the elegant fix -- and the test case! Really appreciate it. (I hate this part of the code base, ha.)

PR #4263, opened 2021-08-02
httpcaddyfile: Add shortcut for proxy hostport placeholder

francislavoie:
I've noticed that it's a pretty common pattern to write a proxy like this, when needing to proxy over HTTPS:

```
reverse_proxy https://example.com {
	header_up Host {http.reverse_proxy.upstream.hostport}
}
```

I find it pretty hard to remember the exact placeholder to use for this, and I continually need to refer to the docs when I need it. I think a simple fix for this is to add another Caddyfile shortcut for this one to shorten it:

```
reverse_proxy https://example.com {
}
```


mholt:
I support this. How do you feel about `{upstream_hostport}` instead? (I'm mostly fine with `{proxy_hostport}`, except that the value isn't actually the host+port of the proxy, it's of the upstream.)

francislavoie:
That works too. I'll change it.

francislavoie:
I think this is ready to go too @mholt

mholt (approved):
LGTM - thanks!

Issue #4112, opened 2021-04-14
Feature: Allow 'named matchers' inside 'not' matcher


fanste:
According to my testing it is currently not possible to do something like
```
@specialHandling {
    path /App1/* /App2/* /App3/*
}

@anotherSpecialHandling {
    not @specialHandling 
}

<ANY_MATCHER_BASED_DIRECTIVE> @specialHandling {
    ...
}

<ANY_MATCHER_BASED_DIRECTIVE> @anotherSpecialHandling {
    ...
}
```

In my setup I use caddy as a TLS enabled reverse proxy for my application server. We have to use multiple application servers as there are some legacy apps, that are not yet upgraded to the newer technologies. But every app should be available from the same host:port, so users don't have to change bookmarks or links referenced in other documents once the legacy apps are upgraded.

Additionally we use the filter module to fix some path/host/port issues that this setup introduces:
- http -> https
- pathes to some assets
- ...

The requested change would allow us to only specify the legacy apps in `@specialHandling` which runs some specific filter rules for those legacy apps while all other apps should have its own filter rules - therefore the `@anotherSpecialHandling `.
Currently we have to write
```
@specialHandling {
    path /App1/* /App2/* /App3/*
}

@anotherSpecialHandling {
    not path /App1/* /App2/* /App3/*
}
```
which is a bit error prone. That will get worse if more complex matchers are required.

thedadams:
I have an implementation for this that also allows for named matchers to contain other named matchers like:
```
@AppPath {
    path /App/*
}

@HeaderFoo {
    header Foo bar
}

@specialHandling {
    @AppPath
    @HeaderFoo
}

@anotherSpecialHandling {
    not @specialHandling
}
```

I am figuring out where the tests should go, and then I can submit a PR.

francislavoie:
Interesting @thedadams! Generally we use adapt tests for any Caddyfile changes https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt

francislavoie:
@thedadams did you make any progress on this?

thedadams:
Yes, the tests revealed an issue with my implementation and then I went on vacation. I have a potential fix, but now I am touching more of the unmarshal code, so I want to get it right.

thedadams:
@francislavoie I have submitted the PR for this change. I understand that a documentation update is also necessary for this update. I would be happy to do that as well when the PR is close to being merged.

mholt:
[...]

Issue #3448, opened 2020-05-25
Provide all core matchers as CEL functions


francislavoie:
CEL is super cool and powerful, makes it easier to write complex conditional logic for request matching. We can use placeholders to pull the values from the request and write conditions with it.

One thing that would make it even better though, is if all our existing request matchers were added to the CEL context as functions.

Consider this Caddyfile (example of doing all the permutations of conditional logic):
```
localhost

@both {
	path /abc/*
	method GET
}
respond @both "Both"

@neither {
	not path /abc/*
	not method GET
}
respond @neither "Neither"

@not_both {
	not {
		path /abc/*
		method GET
	}
}
respond @not_both "Not both"

@either {
	not {
		not path /abc/*
		not method GET
	}
}
respond @either "Either"
```

This is pretty wacky. Especially the "either" case.

If we had the matchers as CEL functions, we could do something like this for each of them (not all of these make sense as expressions, but this is just an example):

```
localhost

@both expression `path("/abc/*") && method("GET")`
respond @both "Both"

@neither expression `!path("/abc/*") || !method("GET")`
respond @neither "Neither"

@not_both expression `!path("/abc/*") && !method("GET")`
respond @not_both "Not both"

@either expression `path("/abc/*") || method("GET")`
respond @either "Either"
```

This reads a lot better, especially if you're comfortable with programming boolean logic.

francislavoie:
@TristonianJones I was hoping you could clarify something - I've been digging through the CEL docs and can't figure out whether it's possible to have variadic args for function overloads. I did see that `&&` is noted as variadic, so I assume there must be a way.

Specifically, I want to be able to support all of the following:

```
path(request, "/foo/*")
path(request, "/foo/*", "/bar/*")
path(request, "/foo/*", "/bar/*", "/baz/*")
...
```

(Note that I'll be using a regexp to expand `path(args)` to `path(request, args)` so that users don't need to specify `request`, which is unnecessary in this context)

Do I just use `dyn` as the last type? Do I have to have users wrap args in a `[]` list instead?

TristonianJones:
@francislavoie The CEL proto supports variadic arguments at an AST level, but the type checker and interpreter don't currently support variadic functions. Using a list literal as an argument is the simplest approximation to what you want: `path(request, [<arg_exprs>])`. Though you could also add overloads to the type-checker and interpreter for arg counts 1 .. N where N is some reasonable number to simulate what you want:

```
path(request, string)
path(request, string, string)
[...]

PR #4264, opened 2021-08-03
Add support for embedded named matchers

thedadams:
This change adds support for caddyfile embedded named matchers, as well
as named matchers in not-blocks.

For example, the following is supported:
```
@matcher1 {
    header Foo bar
    path /bar
}

@matcher2 {
    header Foo buzz
    path /foo
    protocol https
}

@matcher3 {
    @matcher1
    @matcher2
    not query foo=bar
}

@matcher4 {
    header Foo bar
    header Foo buzz
    path /foo /bar
    protocol https
    not query foo=bar
}

@matcher5 {
    not @matcher2
}
```

Note that @matcher3 and @matcher4 are equivalent.

Related issue: https://github.com/caddyserver/caddy/issues/4112


caddytest/integration/caddyfile_adapt/not_block_with_named_matcher.txt
```diff
@@ -0,0 +1,107 @@
+:80
+
+@matcher {
+    header Abc "123"
+    header Bcd "456"
```
francislavoie, line 5: Use tabs instead of spaces, avoids a `caddy fmt` warning.

```suggestion
	header Abc "123"
	header Bcd "456"
```

caddytest/integration/caddyfile_adapt/not_block_with_named_matcher.txt
```diff
@@ -0,0 +1,107 @@
...
+	not @matcher
+}
+respond @not-matcher 403
+
+@nested-not {
+	not {
+		not @matcher
+	}
+}
+respond @nested-not 401
```
francislavoie, line 19: Single line syntax should also work; you can add a test like this as well:

```
@nested-not-single-line not not @matcher
respond @nested-not-single-line 401
```

caddytest/integration/caddyfile_adapt/matchers_embedded.txt
```diff
@@ -0,0 +1,174 @@
...
+									"handler": "static_response"
+								}
+							]
+						},
+						{
+							"match": [
+								{
+									"path_regexp": {
+										"pattern": "/bar*"
+									},
```
francislavoie, line 158: This test is flaky, unfortunately. I think something needs to be adjusted to ensure sorting happens in the Caddyfile adapter.

The issue is that maps in Golang do not have guaranteed order, so a map of matcher names to matcher sets will not be looped over in a predictable order, unless a separate slice is created to sort them.

francislavoie (changes requested):
Awesome! I'm liking the look of this! Spotted a few issues at a glance. Didn't dive deep into the details yet, this will likely involve multiple rounds of review.

It also sparked some ideas for some new matchers we could add to complement this (i.e. `any` and `all` matchers which are essentially `or` and `and` boolean operators on a list of matcher sets, which would have some helpful benefits to some complex matching patterns) but that can wait until after this PR is done.

caddytest/integration/caddyfile_adapt/not_block_with_named_matcher.txt
thedadams: Done.

caddytest/integration/caddyfile_adapt/not_block_with_named_matcher.txt
thedadams: Added this test.

caddytest/integration/caddyfile_adapt/matchers_embedded.txt
thedadams: I changed the implementation so this hopefully won't be a problem.

thedadams:
In addition to addressing the above comments, I added a new `interface` type for `MatchNot` called `UnmarshalerMatcherBlocker`. This will be the `interface` to be implemented for `all` and `any`, if they get implemented.

Also, I usually make changes in the same commit and force push because that is how my team does things. Please let me know if you would like me to make a new commit when I make changes and squash at the end.

thedadams:
@francislavoie Can you give this another review when you have the chance?

caddyconfig/caddyfile/adapter.go
```diff
@@ -109,6 +109,18 @@ type Unmarshaler interface {
...
+// from already parsed named matchers. The named matchers in embeddedMatchers should be merged.
+type EmbeddedMatcherUnmarshaler interface {
+	UnmarshalCaddyfileWithEmbeddedMatchers(d *Dispenser, matchers map[string]caddy.ModuleMap, embeddedMatchers []string) error
+}
+
+// BlockUnmarshaler is the same as Unmarshaler, but will compile named matchers to be passed to the next directive.
+// These are matcher operations like `not`.
+type BlockUnmarshaler interface {
+	UnmarshalCaddyfileWithMatchers(d *Dispenser, matchers map[string]caddy.ModuleMap) error
+}
```
francislavoie, line 122: I'm not sure I like adding _two_ new interfaces here. Can we combine these into one? I figure the `BlockUnmarshaler` usecase would be covered by using a `EmbeddedMatcherUnmarshaler` with an empty `embeddedMatchers []string`, no?

thedadams: The problem is detection: when do you know to use an empty `embeddedMatchers`?

For example, in the following scenario:

```
@matcher1 {
    path /app
}

@matcher2 {
    path /other
    header Foo bar
}

@matcher3 {
    @matcher1
    not @matcher2
}
```

The intended outcome should be equivalent to:
```
@matcher3 {
    path /app
    not {
        path /other
        header Foo bar
    }
}
```
You need to know not to pass the `embeddedMatchers` to the `not` block, otherwise the above will get parsed as:
```
@matcher3 {
    path /app
    not {
        path /app /other
        header Foo bar
    }
}
```

thedadams: It could be solved by always passing in `embeddedMatcher` and then the implementation of `not` (and eventually `any` and `all`) to ignore it.

francislavoie: Yeah I think we can pass `embeddedMatchers` to `not` anyways, and it can ignore it. I think it's probably clear that `not` is sorta like a scope, so it shouldn't inherit anything outside of it anyways. `any` and `all` would probably do the same as well.

thedadams: Cool. This is done and ready for review.

mholt:
Wow, this would be a neat feature, but the change is quite bigger than expected, with new exported identifiers... might take me a while to get around to this in depth. I feel like this change should be internal. Is there no good way to do this without external changes; each matcher really has to manually support being nested?

thedadams:
The problem is that each matcher knows how to parse itself from `Tokens`, but these are discarded after parsing. With this change, the matchers need to know how to parse themselves from `Tokens` and `JSON` from previously parsed `Tokens`.

The other option I can think of is keep the `Tokens` around while parsing, but that would involve saving two copies of each named matcher during the parsing step.

francislavoie:
I think we can get away with being "inefficient" during the Caddyfile adapt step, cause it's only done once at startup.

If you think you might be able to implement it that way, I think it might be preferable.

thedadams:
I can look into that implementation.

The other option I just thought of is making the new interface and method unexported. The implementation would stay roughly the same, but there would be far less exported.

francislavoie:
We could unexport it, but that would mean plugin authors would not be able to have the merging behaviour for their own matchers should they implement any.

It's a hard balance. I think holding onto the tokens might be better.

Maybe we need to set it up in such a way that all the embedded matchers get expanded before actually unmarshaling things 🤔 but I'm not sure if that's viable either.

mholt:
I do think we will need a solution that doesn't require every matcher plugin to add its own code to support this feature. It should be an internal feature without plugin-facing (exported) changes.

Similar to Francis' last suggestion, I also thought if it might be possible to "expand" embedded matchers before they get unmarshaled; i.e. just put the tokens from the embedded matcher right into the matcher it's embedded into, before any parsing begins.

thedadams:
I think this solution requires that we hang onto the tokens after parsing each named matcher.

I will investigate whether it is possible to get around this. Either way, I will change the implementation to be internal.

Thank you both for your feedback.

thedadams:
I was able to reduce the external changes, but I wasn't able to come up with a generic solution for `not` without making some assumptions. What I mean by that is that `not` can either be in-line or a block matcher. The generic solution would need to work in both the following cases:
```
@matcher  {
    not @other-matcher
}

@matcher {
    not {
        @other-matcher
    }
}
```
This causes enough issue that I wasn't able to come up with a generic solution without assuming that I could always turn the first case above into the second. Then this assumption would have to be true for all matcher plugins that allow for both block and inline format.

Long story, short: the `not` case seems different enough to allow for another interface. Other matchers will continue to work without embedded named matchers, or the implementation can be updated to implement the new interface.

If the is not acceptable, I will continue to modify the implementation.

mholt:
Thanks for the refactor, that's a little better, but not quite as ideal as I'm hoping for.

What do you think of a sort of pre-processing approach, where if a matcher token is found inside a matcher definition, the literal tokens for that named matcher are copied into its place? Kind of like how `import` or `{$ENV}` placeholders work.

thedadams:
Consider the following scenario:
```
@matcher1 {
    path /App
    header Foo bar
}

@matcher-not {
    not @matcher1
    header Foo fuzz
}

@matcher-other {
    @matcher1
    header Foo fuzz
}
```
In `@matcher-not`, the embedded matcher should be in its own block, not just blindly replaced. In `@matcher-other`, the embedded matcher can be blindly replaced with its contents.

How should I distinguish between these two cases? It is enough to say that if the embedded named matcher is not the first word on the line, then it should be nested? That will work for `not`, but I'm not sure if that is good enough.

francislavoie:
The "pre-processing" approach would not work for the theoretical `all` and `any` matchers we were thinking about, because it would expand blocks in a strange way that isn't really valid Caddyfile syntax.

Contrived example, but:

```
@foo path /foo
@bar path /bar

@foobar {
	any @foo @bar
}
```

👇

```
@foobar {
	any {
		path /foo
	} {
		path /bar
	}
}
```

Which is pretty strange.

mholt:
I think `any` and `all` should be left to a separate PR. They are both big changes and we should focus on one thing at a time.

Disregarding those, why can't we use the preprocess approach?

francislavoie:
@mholt yes that is the plan, to add those later. But I don't want us to design ourselves into a corner but making that not possible to implement. That's why I think it's worth considering now.

thedadams:
I wanted to let you all know that I was able to get the "preprocess replace" method working. It works for both in-line `not` and nested `not`, and does not require any new interfaces for `not`.

For example:
```
@matcher {
    path /App
    header Foo bar
}

@matcher-not {
    not {
        @matcher
    }
}

@inline-matcher-not {
    not @matcher
}
```
is parsed and turns into:
```
@matcher {
    path /App
    header Foo bar
}

@matcher-not {
    not {
        path /App
        header Foo bar
    }
}

@inline-matcher-not {
    not {
        path /App
        header Foo bar
    }
}
```

I believe it will also work for the future implementations of `any` and `all`.

I can update the PR, if it is decided that this is the implementation we should go with.

francislavoie:
Great! At this point, the implementation is what will determine if it'll get merged, so might as well push it. I suppose you can keep a backup of the current implementation just in case.

> I believe it will also work for the future implementations of any and all.

How do you envision that would that work though? Because as I said above in https://github.com/caddyserver/caddy/pull/4264#issuecomment-899920369, I don't see how that could yield valid syntax, because nothing in the current Caddyfile syntax allows for multiple blocks chained after the other.

thedadams:
> How do you envision that would that work though? Because as I said above in [#4264 (comment)](https://github.com/caddyserver/caddy/pull/4264#issuecomment-899920369), I don't see how that could yield valid syntax, because nothing in the current Caddyfile syntax allows for multiple blocks chained after the other.
The way it work is, if there are multiple embedded matchers in-line then a new block is created and the embedded matchers are parsed into that new block. So this example:
```
@foo path /foo
@bar path /bar

@foobar {
	any @foo @bar
}
```
would turn into:
```
@foo path /foo
@bar path /bar

@foobar {
	any {
              path /foo
              path /bar
         }
}
```

francislavoie:
That wouldn't really work though, the idea of `any`/`all` are that it would wrap _matcher sets_ (not _a_ matcher set, but _many_ matcher sets) and do OR/AND on the entire list of matcher sets, in the order defined.

What you have there would result in a single matcher set where `path` would end up merged with itself.

thedadams:
Ah yes. If you want to say "any of these named matchers" then you're correct: this implementation wouldn't work.

I am beginning to agree that each directive that would like to handle nested embedded matchers would need a copy of the tokens for each embedded matcher so that it could match on them correctly.

francislavoie:
Yeah. 😐 

I just want to say @thedadams thank you _so much_ for the effort you've put into this so far ❤️ 

I just wanted to make sure you hear that it's appreciated, because it's turned into quite the design quandary thus far. I hope you're not getting frustrated with the back-and-forth on this 😅

mholt:
Yes, same -- thank you very much. I need to look at this since all the effort deserves more attention. (Just been so busy!)

thedadams:
No problem at all. I am enjoying working on this issue (and hopefully the `any` and `all` stuff as well).

I will switch back to the previous implementation, but that may need to change as well based on the answers to the questions below:

Currently, the implementation for embedded matchers and embedded `not` matchers just combines the matchers into one big matcher. What I mean is the matchers `@embedded-matcher` and `@equiv-matcher` are equivalent below:
```
@matcher {
    path /App
    header Foo bar
}

@other-matcher {
    path /Other
    header Foo buzz
}

@embedded-matcher {
    @matcher 
    @other-matcher
}

@equiv-matcher {
    path /App /Other
    header Foo bar
    header Foo buzz
}
```
Is this the correct interpretation?

Similarly, for the embedded `not` matchers, the `@embedded-not-matcher` and `@equiv-not-matcher` are equivalent:
```
@matcher {
    path /App
    header Foo bar
}

@other-matcher {
    path /Other
    header Foo buzz
}

@embedded-not-matcher {
    not @matcher @other-matcher
}

@equiv-not-matcher {
    not {
        path /App /Other
        header Foo bar
        header Foo buzz
    }
}
```
However, in this case, I could see an argument for making `@embedded-not-matcher` equivalent to the following:
```
@alternate-equiv-not-matcher {
    not {
        path /App
        header Foo bar
    }
    not {
        path /Other
        header Foo buzz
    }
}
```
Depending on which of these is correct, I may need to change the implementation again.

Thank you for all of your helpful input on this. I am certainly enjoying working on this, and I hope that you are not frustrated with the back-and-forth.

francislavoie:
That's a great question. The difference is pretty subtle.

I'm just now realizing (🤦‍♂️ I should've noticed this earlier) that `not` is actually already the same as what would be `all`, but just without negation. This is because the `not` matcher has merging built-in, so using `not` multiple times in the same named matcher will result in `not` having an array of matcher sets, which get `OR`'d together.

```
@foo {
	not A
	not B
	not C
}
 
👇 

!(A || B || C) --> !A && !B && !C
```

What's crazy, is it's _also_ the same as what would be `any`, also without negation, because you can specify multiple matchers in a block (but this case gets really funky because you cannot really use two matchers of the same type to get this behaviour because they might get merged and handled some other way)

```
@bar not {
	A
	B
	C
}

👇 

!(A && B && C) --> !A || !B || !C
```

So, `any` would obviously be `A || B || C`.

And `all` would obviously be `A && B && C`.

As a sanity check for myself, I verified the above logic using this awesome tool https://web.stanford.edu/class/cs103/tools/truth-table-tool/ (seriously, boolean logic can completely break my brain sometimes if I try to think too hard):

image

So, based on the above, I would say that it would actually make the most sense to go with your `@alternate-equiv-not-matcher` result, since it would make more sense to retain the "matcher-set-ness" and not risk the merging logic that other matchers would do.

And once we add `any` and `all`, it would be really easy to do `not all @A @B @C` which would literally be `!(A && B && C)`, to get the same result (but probably more correct) as your `@equiv-not-matcher`.

Does this all make sense? Have I lost my damn mind? I dunno 🤷‍♂️

francislavoie:
Also, I have to bring up at this point, because this is honestly a completely _fucked_ DSL  😂 (pardon the french), that I we _could_ just forget about all this and try to instead solve https://github.com/caddyserver/caddy/issues/3448.

If we can get all of Caddy's matchers working as CEL functions (there's definitely plenty of other tricky issues there though), then users could _much_ more easily write complex conditional logic using operators that are more familiar like `&&` and `||` directly.

TristonianJones:
@francislavoie @mholt Let me know if there's any context or support I can provide to this conversation. I think there's a strong argument to be made for separating logic from configuration and maybe that's a tension that's surfacing as part of this design. It would be great if you could provide a CEL function as an administrator of a curated package and then expose a way to feed configuration objects to that logic to produce the desired result.

mholt:
Sorry I've had this sit for so long... been very busy with more pressing issues that are less complicated :sweat_smile: 

I am seriously beginning to wonder/think if working on CEL would be a more feasible approach. Especially with @TristonianJones' blessing... I think CEL would be more flexible and intuitive for advanced matcher logic. Or heck, if we need something Turing-complete, maybe it's time to revisit Starlark.

francislavoie:
Re CEL, see here: https://github.com/caddyserver/caddy/issues/3448#issuecomment-922379403

TristonianJones:
@mholt @francislavoie First, let me try and make it possible to upgrade CEL for you and your users, and then I'll take a look at the feature request a bit more. 

The `match` behavior reminds me of the work I did for [Firebase Rules](https://firebase.google.com/docs/rules/rules-language?hl=en). CEL was designed concurrently with Rules and has a lot of overlap with it (and many of the features in Caddy matchers):

```
// Rules snippet, the {document} uri segment is captured as a variable 
// for use in the <expression>
match /documents/{document} {
  allow read: if <expression>
}
```

Firebase Rules is geared toward authorization whereas the Caddy matchers seem to have a broader scope. From what I've read of the docs, it feels like you have a pretty standard attribute vocabulary, though `file` matchers seem to be more specialized than the other request matchers. Also, in Rules, the negation would usually be written as an affirmative rejection, e.g. 

```
match /documents/{document} {
   allow read: if document != "sensitive"
}
```

Composing negation along request paths is usually difficult to do when there are multiple matchers even when there are clear semantics for how matchers combine (logical AND versus logical OR).

In general, I highly recommend avoiding Turing-complete languages in the request path, and a bit part of what we did recently for Kubernetes was to make it easy to set time and complexity limits on expression execution. Where you need expensive computations you should defer to custom functions where the function implementation is written in native code versus in the expression language configured by users.

francislavoie:
I guess we'll close this. We have a pretty nice solution with CEL matchers to craft arbitrary boolean expressions https://github.com/caddyserver/caddy/pull/4715.

Thanks so much @thedadams for the experiment, it helped spark a lot of ideas. I hope we can eventually revisit improving named matcher syntax in the Caddyfile.

PR #4282, opened 2021-08-16
logging: Remove common_log field and single_field encoder (#4149)

mholt:
Follows up on #4149 by removing common log format and the single_field encoder, which was a hack to preserve CLF logs. However these features have compliance problems and were only introduced in Caddy 2 to help people transition to structured logs, so it's time to remove them.

Will merge this sometime after the v2.4.4 release.


francislavoie:
Hmm. Could we at least mark the `SingleFieldEncoder` as deprecated for at least 1 minor version (log a warning on provision)?

Feels weird to remove so suddenly. I'd rather at least give people a chance to move away from it. Not everyone reads release notes, but hopefully the warnings will be noticed.

mholt:
It has been announced for a while, but I'm OK with emitting a warning for a minor version, I suppose. 😢

francislavoie:
Yeah. I would love to YOLO on this, but I feel like we need to go slow to keep respect in the project from the community 😅

mholt:
> Could we at least mark the SingleFieldEncoder as deprecated for at least 1 minor version (log a warning on provision)?

Done in 8a974a4f8f1628c0b74c3a19cb8f2d0b48123cca

francislavoie:
I'll merge this now, since we know our next release will be 2.5.0.

francislavoie (approved):

PR #4283, opened 2021-08-16
ci: Build and test on Go 1.17, bump minimum to 1.16

francislavoie:
Used https://github.com/caddyserver/caddy/pull/4024 as a template. Pretty much the same change, different versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        go: [ '1.15', '1.16' ]
+        go: [ '1.16', '1.17' ]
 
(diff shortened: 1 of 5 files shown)


francislavoie:
FYI just before merging, the repo should be updated to make the 1.17 jobs required, and make the 1.15 job no longer required.

francislavoie:
To be considered: should we start building for Windows ARM64, or at least add it to the cross-build test?

Mohammed90 (approved):
The build is successful and nothing in the release notes (https://golang.org/doc/go1.17) impacts the go.{mod,sum} files to where the project won't build. Thanks for tackling this, Francis!

mholt (approved):
Required CI checks have been updated to shift out 1.15 for 1.17.

@francislavoie We have had several requests for Windows ARM 64, so we should consider building for it I think. Doesn't have to be in this PR though.

PR #4285, opened 2021-08-17
caddyhttp: Implement `http.request.uuid` placeholder

rainerborene:
`http.request.uuid` placeholder enables you to trace and troubleshoot requests end‑to‑end. 
```caddyfile
reverse_proxy web:3000 {
  header_up X-Request-Id {http.request.uuid}
}
```

### DIFF
--- a/modules/caddyhttp/replacer.go
+++ b/modules/caddyhttp/replacer.go
@@ -40,6 +40,7 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
+	"github.com/google/uuid"
 )
 
 // NewTestReplacer creates a replacer for an http.Request
@@ -54,6 +55,7 @@ func NewTestReplacer(req *http.Request) *caddy.Replacer {
 
 func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.ResponseWriter) {
 	SetVar(req.Context(), "start_time", time.Now())
+	SetVar(req.Context(), "uuid", new(requestID))
 
 	httpVars := func(key string) (interface{}, bool) {
 		if req != nil {
@@ -146,6 +148,9 @@ func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.Respo
 			case "http.request.duration":
 				start := GetVar(req.Context(), "start_time").(time.Time)
 				return time.Since(start), true
+			case "http.request.uuid":
+				id := GetVar(req.Context(), "uuid").(*requestID)
+				return id.String(), true
 			case "http.request.body":
 				if req.Body == nil {
 					return "", true
@@ -396,6 +401,20 @@ func getTLSPeerCert(cs *tls.ConnectionState) *x509.Certificate {
 		return nil
 	}
 	return cs.PeerCertificates[0]
+}
+
+type requestID struct {
+	value string
+}
+
+// Lazy generates UUID string or return cached value if present
+func (rid *requestID) String() string {
+	if rid.value == "" {
+		if id, err := uuid.NewRandom(); err == nil {
+			rid.value = id.String()
+		}
+	}
+	return rid.value
 }
 
 const (
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -50,6 +50,7 @@ func init() {
 // `{http.request.body}` | The request body (⚠️ inefficient; use only for debugging)
 // `{http.request.cookie.*}` | HTTP request cookie
 // `{http.request.duration}` | Time up to now spent handling the request (after decoding headers from client)
+// `{http.request.uuid}` | The request unique identifier
 // `{http.request.header.*}` | Specific request header field
 // `{http.request.host.labels.*}` | Request host labels (0-based from right); e.g. for foo.example.com: 0=com, 1=example, 2=foo
 // `{http.request.host}` | The host part of the request's Host header


francislavoie:
I don't think we need this in Caddy.

You should instead use this plugin: https://github.com/lolPants/caddy-requestid

rainerborene:
Ok, can you elaborate why? `nginx` and `haproxy` has this feature natively. This is much more than a "nice to have" feature. It is an essential feature for troubleshooting requests.

mholt:
I do think if we are going to integrate this natively into Caddy, it needs more consideration. This change assumes a request originates with Caddy, but for true end to end tracing we need to reuse existing request ID if there is one.

francislavoie:
Also I feel like we shouldn't _always `SetVar`, but only do it if that specific placeholder is requested. There's a bit of overhead from generating a UUID, I'd prefer avoiding it if the users didn't request it.

rainerborene:
@mholt I noticed that the [implementation](https://github.com/caddyserver/caddy/blob/v1.0.5/caddyhttp/requestid/requestid.go#L36-L47) of `requestid` in previous `caddy` version tries to parse the existing RID in `uuid` form for validation purpose, and it even generates a new one if invalid. By looking at the `nginx` [source code](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_variables.c#L2177-L2215) it seems that they don't do any sort of verification. `nginx` only generates an unique request identifier from 16 random bytes, in hexadecimal. I don't think we should worry about all that. Our only concern should be to generate an unique request identifier (if necessary and explicit used on Caddyfile).

@francislavoie I completely agree with that.

francislavoie:
Please also remember to add the new placeholder to the godoc table in `modules/caddyhttp/app.go`

mholt:
I also agree we don't need to validate the RID, it should be allowed to be any kind of string.

rainerborene:
It seems that `GetVar` and `SetVar` does not work under this scenario or am I missing something?
```golang
// func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.ResponseWriter) { ...
	case "http.request.uuid":
		requestId := GetVar(req.Context(), "uuid")
		if requestId == nil {
			requestId := uuid.New().String()
			SetVar(req.Context(), "uuid", requestId)
		}
		return requestId, true
```

mholt:
What does "doesn't work" mean?

rainerborene:
@mholt It can't be used multiple times on the same `Caddyfile`.

francislavoie:
You'll need to elaborate. What's your Caddyfile? What behaviour are you seeing?

rainerborene:
This is my `Caddyfile`:
```
:2015 {
  header X-Request-Id {http.request.uuid}
  header X-Queue-Id {http.request.uuid}
  respond "it works" 200
}
```
Response:
```
HTTP/1.1 200 OK
Content-Length: 8
Date: Tue, 17 Aug 2021 22:07:32 GMT
Server: Caddy
X-Queue-Id: a33eb854-a68a-4e7c-8a87-b3cf56b22fe0
X-Request-Id:

it works
```

francislavoie:
Hmm. I'm not sure why that's happening. You'll need to debug further.

That said though, I don't think it's safe to use `uuid.New()`, since it may panic. Panics should never happen in Caddy, because the server should continue to run. Instead, please use `uuid.NewRandom()` (like the only other usage of `uuid` in Caddy does).

rainerborene:
Okay. I finally got it working. I had to wrap `uuid` with a `struct` and lazy generate it when needed. If I use `SetVar` outside of `httpVars` function, it works fine.

rainerborene:
@mholt @francislavoie let me know if this needs more work.

francislavoie (commented):
LGTM, but @mholt has a better eye for quirks of the Go language. (First thought is will this get GC'd properly? I think so, the reference to the struct in Context should be tracked, but I'm not 100% certain)

mholt:
Yeah, I mean, I've taken a few looks at this and have been thinking about it. I can see why you went with a struct, but I'm still wondering if that's really necessary... I kind of want to perform some tests on my end when I get a chance, since I would _think_ that just storing a `*UUID` should work too.

francislavoie:
Yeah, `*UUID` is a `fmt.Stringer` as well, so that should work fine too.

mholt:
I don't think this fulfills the needs quite right though. If a request already has an ID that originated outside of Caddy, this doesn't take that into account. Doesn't there need to be some way to configure how it should get the existing request ID?

rainerborene:
@mholt I don't get it. If the request ID already exists, what's the point of using this dynamic variable? Please, see @francislavoie [here](https://github.com/caddyserver/caddy/pull/4285#issuecomment-900410759)

mholt:
@rainerborene I'm sorry, you're right. I missed that comment or forgot about it.

rainerborene:
@francislavoie @mholt Can you merge this? Let's prepare for v2.5.0 :D

modules/caddyhttp/replacer.go
```diff
@@ -398,6 +403,21 @@ func getTLSPeerCert(cs *tls.ConnectionState) *x509.Certificate {
...
+
+// Lazy generates UUID string or return cached value if present
+func (rid *requestId) String() string {
+	if rid.value != "" {
+		return rid.value
+	}
+	if id, err := uuid.NewRandom(); err == nil {
+		rid.value = id.String()
+	}
+	return rid.value
```
mholt, line 418:
```suggestion
	if rid.value == "" {
		if id, err := uuid.NewRandom(); err == nil {
			rid.value = id.String()
		}
	}
	return rid.value
```

A little more concise.

mholt (changes requested):
Ok, this is my last review I think. Looking good. I like that it's lazily generated. I don't love the allocaftion for it, but we can optimize things later.

rainerborene:
@mholt Done. Thanks

francislavoie:
@rainerborene the tests are failing because `requestID` is not the same as `requestId`

modules/caddyhttp/replacer.go
```diff
@@ -398,6 +403,20 @@ func getTLSPeerCert(cs *tls.ConnectionState) *x509.Certificate {
 	return cs.PeerCertificates[0]
 }
 
+type requestId struct {
```
mholt, line 406:
```suggestion
type requestID struct {
```

This still needs to be updated :+1: then we can merge it. (Make sure to update the name everywhere.)

mholt (changes requested):
Thanks, looks like the change from `requestId` to `requestID` got undone though.

modules/caddyhttp/replacer.go
```diff
@@ -398,6 +403,20 @@ func getTLSPeerCert(cs *tls.ConnectionState) *x509.Certificate {
...
+	value string
+}
+
+// Lazy generates UUID string or return cached value if present
+func (rid *requestId) String() string {
+	if rid.value == "" {
+		if id, err := uuid.NewRandom(); err == nil {
+			rid.value = id.String()
```
mholt, line 414: Actually, I'm pretty sure this is racey. Can you test this with multiple requests using `go build -race` (or `go run -race`)?

rainerborene: It's not racey. I tested with `go build -race` and `wrk`.

mholt: Ah, of course. I'm an idiot... these values are only used within this request, which is generally all on a single goroutine, so it's not a race. 🤦‍♂️

mholt (approved):
Thank you :) Time to merge this in.

mholt:
Before we release this, any thoughts on the placeholder name? `http.request.uuid` vs. `http.request.id` ? Is it important that it's a UUID?

rainerborene:
I think `uuid` is more clear and concise. Apache, for example, uses `UNIQUE_ID` name.

Issue #3805, opened 2020-10-18
cmd/commandfuncs: use net.ErrClosed instead of doing substring error checks [requires go 1.16]


jimen0:
Go 1.16 will introduce a new exported [`net.ErrClosed`](https://tip.golang.org/pkg/net/#ErrClosed) (see [this issue](https://github.com/golang/go/issues/4373) for context) error allowing Caddy to properly check for this error instead of using the substring matching hack.

Current code:

https://github.com/caddyserver/caddy/blob/385adf5d878939c381c7f73c771771d34523a1a7/cmd/commandfuncs.go#L115-L123

Proposed patch once Caddy starts using go 1.16:

```diff
diff --git a/cmd/commandfuncs.go b/cmd/commandfuncs.go
index 772fe012..36c686c5 100644
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -116,7 +117,7 @@ func cmdStart(fl Flags) (int, error) {
 		for {
 			conn, err := ln.Accept()
 			if err != nil {
-				if !strings.Contains(err.Error(), "use of closed network connection") {
+				if !errors.Is(err, net.ErrClosed) {
 					log.Println(err)
 				}
 				break
```

Please, let me know what do you think about it and, if you would prefer me to open the PR now or if I should wait for 1.16 to be released. Also, please feel free to close this issue if you don't think we need this code change.

Best,
Miguel

francislavoie:
This is interesting, but I'm not sure we can realistically make this change soon if it will require us to bump up the minimum Go version. We want to keep a reasonably low minimum version to maintain compatibility with the Go version used by the different platforms we support, e.g. Fedora/Centos.

Probably best to wait a couple more versions to let the ecosystem to catch up before we bump minimum.

jimen0:
Hi @mholt @francislavoie 1.17beta1 is out. Should I prepare the PR and submit it once 1.17 is out or would you prefer me waiting until 1.18 is released? 😄

francislavoie:
Thanks for the reminder @jimen0!

[...]

PR #4289, opened 2021-08-18
cmd: use net.ErrClosed for matching returned error

jimen0:
Implements #3805


mholt (approved):
LGTM, thank you very much for the improvement!

Issue #3525, opened 2020-06-26
pb.proto warning


dijiwiz:
I was upgrading to v2.1 and the build seemed to go fine.

When I attempted to print up the version to confirm I was shown this warning message:

2020/06/26 17:58:17 WARNING: proto: file "pb.proto" is already registered
A future release will panic on registration conflicts. See:
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict

v2.1.0 h1:MC4d65RCVaEKy1iOFjsD51mybOwS8qdEVBi7ESDhUfE=

It seems to run fine but thought that may be of interest to the project.

Mohammed90:
According to dgraph-io/dgraph#5290, this is due to upgrading the github.com/golang/protobuf package and can be fixed by downgrading it. Needs further digging to understand the nuance of what's going on there. Nonetheless, the fix seems simple.

dijiwiz:
Thank you!

Mohammed90:
I'll keep it open. We need to resolve it by inspecting our dependency graph and pick appropriate packages version.

The error originates from [github.com/dgraph-io/dgraph](https://github.com/dgraph-io/dgraph/blob/master/protos/pb.proto), which is in our dep graph due to github.com/smallstep/nosql. 
```bash
$ go mod why github.com/dgraph-io/badger
# github.com/dgraph-io/badger
github.com/caddyserver/caddy/v2/modules/caddypki/acmeserver
github.com/smallstep/nosql
github.com/smallstep/nosql/badger/v1
github.com/dgraph-io/badger
```

linquize:
c9049bdc24cf35a1c2a4871401428cd08daf1f7d causes the issue.

mholt:
That's weird, those are just patch updates. There should be no breaking changes in those.

francislavoie:
I think `github.com/smallstep/nosql v0.3.1` is to blame.

Technically it's not a breaking change in isolation for that lib to update its protobuf dependency, but the problem is that we also depend on CEL which also uses protobuf, but a version from a different repo.

I think those weren't marked as conflicting dependencies by the go mod dependency solver so we have two different versions pulled in of protobuf now.

I'm just thinking out loud based on the evidence I think is apparent but I can't say for sure, I don't know how go mod does its dep resolution, just making guesses here.

I think #3526 is related.

Mohammed90:
[...]

PR #4291, opened 2021-08-18
chore: Upgrade smallstep libs

francislavoie:
See https://github.com/smallstep/nosql/issues/12 for context.

/cc @ThinkChaos regarding https://github.com/caddyserver/caddy/issues/3525#issuecomment-899869317

There's a surprising amount of `go.sum` changes from this, but I don't think we need to worry about that. It's a side effect of Go not pruning the graph in Go 1.16, I think once we bump to Go 1.17, it should get shrunken down significantly. See https://golang.org/ref/mod#graph-pruning


mholt:
So to be sure I follow, this removes the pb.proto warning and removes the CGO dependency?

francislavoie:
Doesn't remove the CGO dependency _yet_. That's gonna require https://github.com/dgraph-io/badger/pull/1736 first, because `smallstep/nosql` seems to not be able to safely upgrade to badger v3 without breaking changes, so I'm trying to get badger to backport the fix for that to their v2 branch.

But AFAIK yeah it should solve the pb.proto issue and some other random problems that have shown up since.

ThinkChaos:
I can confirm it fixes the pb.proto warning for me.

mholt (approved):
Thanks for helping push this along! Looking forward to getting rid of that cgo dependency too.

PR #4296, opened 2021-08-20
cmd: Fix paths when using an env file

### LINKED ISSUE #4273: Caddy environment files do not affect config & data paths

```
1. Ubuntu 20.04
2. Alpine 3.14.1
```

```
1. v2.4.3 h1:Y1FaV2N4WO3rBqxSYA8UZsZTQdN+PwcoOcAiZTM8C0I=
2. 569ecdbd02a3bdf8589f785cca022240973dea4d
```

### 1c. Go version

```
1.15
```

### 2a. What happens
I start go with the `--envfile` flag and point it to a envfile with the following contents:
```
XDG_DATA_HOME=/home/[user]/.caddy/data
XDG_CONFIG_HOME=/home/[user]/.caddy/config
```

I start Caddy now with `caddy run --envfile /home/[user]/.caddy/.env --config /home/[user]/Caddyfile --environ`.

I would now expect Caddy to put its files into the specified directories. Instead, Caddy will create the directories in the default location.

When using `--environ`, the log output of the environment variables shows the proper values (except for `caddy.ConfigAutosavePath`).
However, in the further output where Caddy informs about the autosaved config, the common paths are used. 

Using real environment variables works as expected and the files are saved at the proper location.

### 2b. Why it's a bug
It is expected that the .env file has the same effect as setting the environment variable on the system.

### 2c. Log output

Without a fix (Stripped down to the essential parts):
```
caddy.HomeDir=/home/pascal/caddydir/ch
caddy.AppDataDir=/home/pascal/caddydir/cd/caddy
caddy.AppConfigDir=/home/pascal/caddydir/cc/caddy
caddy.ConfigAutosavePath=/home/pascal/.config/caddy/autosave.json
caddy.Version=(devel)

...
{"level":"info","ts":1628361986.0723467,"msg":"using provided configuration","config_file":"/home/pascal/caddy.json","config_adapter":""}
2021/08/07 20:46:26 [INFO] Redirecting sink to: stdout
{"level":"warn","ts":1628361986.0747495,"logger":"admin","msg":"admin endpoint disabled"}
{"level":"info","ts":1628361986.0792449,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc000508000"}
{"level":"info","ts":1628361986.0966988,"msg":"autosaved config (load with --resume flag)","file":"/home/pascal/.config/caddy/autosave.json"}
{"level":"info","ts":1628361986.096718,"msg":"serving initial configuration"}
{"level":"info","ts":1628361986.096729,"logger":"tls","msg":"cleaning storage unit","description":"FileStorage:/home/pascal/.local/share/caddy"}
{"level":"info","ts":1628361986.0967355,"logger":"tls","msg":"finished cleaning storage units"}
```

With a fix (Stripped down to the essential parts):
```
caddy.HomeDir=/home/pascal/caddydir/ch
caddy.AppDataDir=/home/pascal/caddydir/cd/caddy
caddy.AppConfigDir=/home/pascal/caddydir/cc/caddy
caddy.ConfigAutosavePath=/home/pascal/caddydir/cc/caddy/autosave.json
...
{"level":"info","ts":1628362230.0860615,"msg":"using provided configuration","config_file":"/home/pascal/caddy.json","config_adapter":""}
2021/08/07 20:50:30 [INFO] Redirecting sink to: stdout
{"level":"warn","ts":1628362230.0936966,"logger":"admin","msg":"admin endpoint disabled"}
{"level":"info","ts":1628362230.0943089,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc000230070"}
{"level":"info","ts":1628362230.112259,"msg":"autosaved config (load with --resume flag)","file":"/home/pascal/caddydir/cc/caddy/autosave.json"}
{"level":"info","ts":1628362230.1122801,"msg":"serving initial configuration"}
{"level":"info","ts":1628362230.113837,"logger":"tls","msg":"cleaning storage unit","description":"FileStorage:/home/pascal/caddydir/cd/caddy"}
{"level":"info","ts":1628362230.113868,"logger":"tls","msg":"finished cleaning storage units"}
```

### 2d. Workaround(s)

1. Use system environment variables

Only other work around is to really solve the issue.

1. Add a .env file (as described below
2. Add a basic Caddyfile (can contain anything, you just need any)
3. Start Caddy using `caddy run --envfile <path-to-env-file> --config <path-to-config>`
4. Check if Caddy created its files at the location specified in the env file

**.env file:**
```
XDG_DATA_HOME=/home/[user]/.caddy/data
XDG_CONFIG_HOME=/home/[user]/.caddy/config
```

**Caddyfile:**
```
localhost {
        root: /path/to/any/route
	encode zstd gzip
	templates
	file_server browse
}
``` 

You can try the same test again using system environment variables (e. g. `export XDG_DATA_HOME=/home/[user]/.caddy/data`) and you will notice that it worked as expected.

## 4. Solution
The issue occurs because the values are being initialized as soon as `storage.go` is used (`var xy = xy.xy`).
A possible solution would be to add a reinitialization/update of `caddy.ConfigAutosavePath` and `caddy.DefaultStorage`.

This could be done by adding a call to some function (e. g. `UpdateStoragePaths `or `ReInitializeStoragePaths` after updating the environment variables in `cmd/main.go#loadEnvFromFile` after line 362.

Then we could add a function to `storage.go` that just runs the code used to initialize the variables again. This would only being called when really a env file has been specified and not at all affect anything when not using one.

Example:
```go
// file: cmd/main.go
// line: 364
caddy.UpdateStoragePaths()
```

```go
// file: storage.go
// starting at line: 156
// --> This is the newly added function
// UpdateStoragePaths sets the config auto save and default storage path
// again. This ensures that env files that overwrite the paths
// have the expected effect.
// If used, this function must be called before ConfigAutosavePath and
// DefaultStorage are being used.
func UpdateStoragePaths() {
	ConfigAutosavePath = filepath.Join(AppConfigDir(), "autosave.json")
	DefaultStorage = &certmagic.FileStorage{Path: AppDataDir()}
}

// Here everything stays as it is now

// ConfigAutosavePath is the default path to which the last config will be persisted.
var ConfigAutosavePath = filepath.Join(AppConfigDir(), "autosave.json")

// DefaultStorage is Caddy's default storage module.
var DefaultStorage = &certmagic.FileStorage{Path: AppDataDir()}
```

This solution should be fine as the program flow does seem to first refer to this stuff by using `caddy.ConfigAutosavePath`  at `cmd/commandfuncs.go` in line 185 (and only when `--resume` is set. The update of the variables would have already happened in line 170 for the example of the `caddy run` command.

_If you think this issue should be solved and this would be an appropriate solution, I would implement it and open a PR.
I'm open for suggestions for a better or cleaner solution._

pascal-zarrad:
### Description
Fix the behaviour of the environment variables `XDG_DATA_HOME` and `XDG_CONFIG_HOME` when supplied through an env file to match the expected one.

For further information, check #4273 

### Related Issues
Resolves #4273 

### Testing
 - Caddy should create its data and config directories and files at the common location as always when no env file is specified using `--envfile`.
   - For example: `/home/user/.local/share/caddy/<files and/or folders>`
 - When an env file is specified using `--envfile`, Caddy should use the environment variables `XDG_DATA_HOME` and `XDG_CONFIG_HOME` from the env file instead (instead) of the system ones. This should cause Caddy to create its files/directories at the location specified in the env file. 
   - Example: Instead of `/home/user/.local/share/caddy/autosave.json` Caddy now uses `/home/user/caddydata/caddy/autosave.json`, when `XDG_CONFIG_HOME` was set to `/home/user/caddydata` in the env file.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -363,7 +363,8 @@ func loadEnvFromFile(envFile string) error {
 
 	// Update the storage paths to ensure they have the proper
 	// value after loading a specified env file.
-	caddy.UpdateStoragePaths()
+	caddy.ConfigAutosavePath = filepath.Join(caddy.AppConfigDir(), "autosave.json")
+	caddy.DefaultStorage = &certmagic.FileStorage{Path: caddy.AppDataDir()}
 
 	return nil
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -361,6 +361,10 @@ func loadEnvFromFile(envFile string) error {
 		}
 	}
 
+	// Update the storage paths to ensure they have the proper
+	// value after loading a specified env file.
+	caddy.UpdateStoragePaths()
+
 	return nil
 }
 
--- a/storage.go
+++ b/storage.go
@@ -153,16 +153,6 @@ func AppDataDir() string {
 	return "./caddy"
 }
 
-// UpdateStoragePaths sets the config auto save and default storage path
-// again. This ensures that env files that overwrite the paths
-// have the expected effect.
-// If used, this function must be called before ConfigAutosavePath and
-// DefaultStorage are being used.
-func UpdateStoragePaths() {
-	ConfigAutosavePath = filepath.Join(AppConfigDir(), "autosave.json")
-	DefaultStorage = &certmagic.FileStorage{Path: AppDataDir()}
-}
-
 // ConfigAutosavePath is the default path to which the last config will be persisted.
 var ConfigAutosavePath = filepath.Join(AppConfigDir(), "autosave.json")
 
--- a/storage.go
+++ b/storage.go
@@ -153,6 +153,16 @@ func AppDataDir() string {
 	return "./caddy"
 }
 
+// UpdateStoragePaths sets the config auto save and default storage path
+// again. This ensures that env files that overwrite the paths
+// have the expected effect.
+// If used, this function must be called before ConfigAutosavePath and
+// DefaultStorage are being used.
+func UpdateStoragePaths() {
+	ConfigAutosavePath = filepath.Join(AppConfigDir(), "autosave.json")
+	DefaultStorage = &certmagic.FileStorage{Path: AppDataDir()}
+}
+
 // ConfigAutosavePath is the default path to which the last config will be persisted.
 var ConfigAutosavePath = filepath.Join(AppConfigDir(), "autosave.json")
 


francislavoie (approved):

mholt (changes requested):
Thanks for the PR. (Sorry for how long it's taking me to get around to this issue.)

I'm very sensitive about adding new exported identifiers, and in this case I think we can avoid it. The function is only called in one place, so I think it'd be fair to just move that logic directly into the `loadEnvFromFile()` function, or, if you still want a separate function, move it next to `loadEnvFromFile()` and leave it unexported. (I don't think it needs to be a separate function though, personally.)

That sound alright?

pascal-zarrad:
I agree that exporting this function isn't really necessary. Exporting it brings the risk of someone using it at a wrong spot in the future.

The logic has been moved to `loadEnvFromFile()`.  I also brought the branch up to date.

mholt (approved):

mholt:
Much better - thanks so much!

PR #4266, opened 2021-08-03
reverseproxy: Pull latest proxy changes from stdlib

francislavoie:
I went through the commits that touched stdlib's `reverseproxy.go` file in the past couple of years (only a dozen or so), and copied over all the changes that are to code that was copied into Caddy.

The commits I pulled changes from (hopefully apparent from looking at those commits what the changes in this PR are doing):

- https://github.com/golang/go/commit/2cc347382f4df3fb40d8d81ec9331f0748b1c394
- https://github.com/golang/go/commit/a5cea062b305c8502bdc959c0eec279dbcd4391f
- https://github.com/golang/go/commit/ecdbffd4ec68b509998792f120868fec319de59b
- https://github.com/golang/go/commit/21898524f66c075d7cfb64a38f17684140e57675
- https://github.com/golang/go/commit/ca3c0df1f8e07337ba4048b191bf905118ebe251
- https://github.com/golang/go/commit/9c017ff30dd21bbdcdb11f39458d3944db530d7e

This may also fix https://github.com/caddyserver/caddy/issues/4247 because of the change to `copyResponse` to set `mlw.flushPending = true` right away. /cc @leafac if you'd be willing to try out this PR (you can grab a build from the CI artifacts, or build from this commit with `xcaddy`) to see if it fixes your usecase.

PR #4299, opened 2021-08-22
reverseproxy: Remove redundant flushing

francislavoie:
From reading through the code, I think this code path is now obsoleted by the changes made in https://github.com/caddyserver/caddy/pull/4266.

Basically, `h.flushInterval()` will set the flush interval to `-1` if we're in a bi-directional stream, and the recent PR ensured that `h.copyResponse()` properly flushes headers immediately when the flush interval is non-zero. So now there should be no need to call Flush before calling `h.copyResponse()`.

---

Since this was your code from https://github.com/caddyserver/caddy/issues/3556 and https://github.com/caddyserver/caddy/pull/3561 @masknu, would you mind confirming that your usecase still works correctly with this change, just as a sanity check? Thanks!

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -691,15 +691,6 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, repl *
 	}
 
 	rw.WriteHeader(res.StatusCode)
-
-	// some apps need the response headers before starting to stream content with http2,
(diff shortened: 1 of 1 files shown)


francislavoie:
Well, tests are passing, and there was a test added along with the original change (https://github.com/masknu/caddy/blob/744d04c2585d50f64cf3d43d139c310a18e78f73/caddytest/integration/stream_test.go#L23), so I _think_ we're good to go here, so we can merge unless we hear back in the next couple days.

mholt (approved):
Ok, good point. I'll merge this, and we can revert if we find out otherwise.

Issue #4298, opened 2021-08-22
handle_response errors with "read on closed response body"


polarathene:
Following the [docs for `reverse_proxy` directive with `handle_response`](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#intercepting-responses) on a very basic JSON response endpoint results in failure, but modifying the header elsewhere in Caddy works as expected.

**Caddyfile (v2.4.3):**

```Caddyfile
:80 {
  
  # Works:
  #header Cache-Control "max-age=10"
  reverse_proxy localhost:8000 {
    handle_response {
      # Errors:
      header Cache-Control "max-age=10"
    }
    # Works:
    #header_down Cache-Control "max-age=10"
  }
}
```

**Error logged:**

```
2021/08/22 02:34:00.645 ERROR   http.handlers.reverse_proxy     reading from backend    {"error": "http: read on closed response body"}
2021/08/22 02:34:00.645 ERROR   http.handlers.reverse_proxy     aborting with incomplete response       {"error": "http: read on closed response body"}
```

The upstream still received the request and responded no differently AFAIK, in the web browser network tools Caddy has sent a response header but omitted the body.

---

For added context, intention was to match on `Content-Type application/json` rather than the URI; but AFAIK this is only available via the response, and thus can only be matched within `handle_response`? As the response matcher is not required to reproduce the error I have excluded it.

polarathene:
cURL response if useful:

```
curl -v http://localhost/json
*   Trying ::1:80...
* Connected to localhost (::1) port 80 (#0)
> GET /json HTTP/1.1
> Host: localhost
> User-Agent: curl/7.74.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Cache-Control: max-age=10
< Content-Length: 18
< Content-Type: application/json
< Date: Sun, 22 Aug 2021 02:16:35 GMT
< Server: Caddy
< Server: Rocket
< 
* transfer closed with 18 bytes remaining to read
* Closing connection 0
curl: (18) transfer closed with 18 bytes remaining to read
```

francislavoie:
That's not how `handle_response` works. It's meant to match _some_ responses, and do something completely different instead of actually returning that response to the client. If you use `handle_response` with routes defined, then the response body will _not_ be copied from the proxy response to the client response. As the docs say:

[...]

PR #4300, opened 2021-08-22
reverseproxy: Move status replacement intercept to `replace_status`

francislavoie:
I was doing some thinking in https://github.com/caddyserver/caddy/issues/4298#issuecomment-903219435 and I think the UX of `handle_response` is super awkward in that it was doing two different things in one subdirective.

Semantically, it's expected that the `handle` family of options imply that subroutes will be defined (as per `handle` and `handle_path` directives).

Separating the status code replacement usecase into its own Caddyfile subdirective (even if it doesn't match the JSON 1:1) is easier to understand.

This is a breaking change, so we could defer this to v2.5.0, but I don't think many people use this yet so the impact should be very low. I haven't seen anyone on the forums asking questions using this feature yet since we released it in v2.4.0.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_handle_response.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_handle_response.txt
@@ -1,6 +1,9 @@
 :8884
 
 reverse_proxy 127.0.0.1:65535 {
+	@changeStatus status 500
+	replace_status @changeStatus 400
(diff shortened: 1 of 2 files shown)


jungjuheeer:
caddyfile-replace-status

mholt (approved):
Sorry it took me so long to get to this. I agree this is probably better; thank you!

PR #4303, opened 2021-08-22
caddyhttp: Add optional trailing slash syntax for path matchers

francislavoie:
This is an idea taken from the WICG URLPattern draft, that I thought would be a perfect fit for Caddy.

https://github.com/WICG/urlpattern/blob/main/mdn-drafts/QUICK-REFERENCE.md

It's reasonably common that users want to match both `/foo` and `/foo/*` at the same time, but not `/foobar`. This is relevant when some service is being proxied to under a subpath, but you want to make sure there's no "leakage" in the path matching, matching more than will actually function properly.

The URLPattern draft has the `/*?` syntax which means "make the wildcard and slash optional". This is a pretty neat pattern that can fold `/foo` and `/foo/*` as above into just `/foo/*?` to match the exact same thing in one rule.

This means it can also be used for inline path matchers to do both those things at once, usually saving a line of Caddyfile config.

Before:
```
example.com

@service path /service /service/*
reverse_proxy @service localhost:8000
```

After:
```
example.com

reverse_proxy /service/*? localhost:8000
```


rainerborene:
This is great! I'm currently doing this instead:
```
@admin {
  path /admin/*
  not file
}
redir /admin /admin/
reverse_proxy @admin web:3000
```

francislavoie:
That's also a valid approach. It depends whether you actually want the redirect or not, since the redirect causes an extra roundtrip to tell the browser to try again.

Both should typically should behave the same if you've configured some "base URL" option on the upstream service typically.

rainerborene:
I don't want redirects. Using this new path matcher I could use `path /admin/*?` and discard the `redir`, right?

francislavoie:
Yep. You could also just do `path /admin /admin/*` which would be identical to what this PR is doing.

mholt:
That's a weird syntax to me. To indicate an optional trailing slash, shouldn't it be `/?*`? But at that point, why not just use a regexp? (How many specific cases do we want to optimize for...)

francislavoie:
I'd honestly like for `path` to fully support `URLPattern` once that gets ratified. That would be super cool.

The `URLPattern` draft is heavily inspired by https://github.com/pillarjs/path-to-regexp which supports that syntax.

> Per the path-to-regexp behavior the optional modifier makes the preceding prefix separator, "/", also optional.

It makes more sense when you consider a pattern like `/:foo/:bar?`. This would match both `/1/2` and `/1` because `?` makes all of `/:bar` optional (the parameter and the separator).

PR #4313, opened 2021-08-30
caddyhttp: Redirect HTTP requests on the HTTPS port to https://

francislavoie:
This is very much not done yet, but I just wanted to show the proof of concept.

---

Background: in certain cases, users would like HTTP->HTTPS redirects in situations where they only actually serve HTTPS on one port; if an HTTP request is made to that port, Caddy currently responds with `Client sent an HTTP request to an HTTPS server.` This response comes directly from Go's stdlib (added in https://go-review.googlesource.com/c/go/+/143177/, see https://github.com/golang/go/issues/23689#issuecomment-362936254 for context):

https://github.com/golang/go/blob/56c3856d529d72e280ad6b185f7927657de86c37/src/net/http/server.go#L1817-1825

While it could be argued that HTTP responses should never be sent on the HTTPS port... Go already does it.

This is alright, but Caddy _could_ go a step further and do an HTTP->HTTPS redirect, which would make HTTP clients happier, skipping the error and instead going straight to HTTPS. This can help avoid confusion for users making mistakes, etc.

---

So after hacking on this a bit, the idea I came up with is to make `tlsPlaceholderWrapper` a bit more useful by peeking the first bytes of the connection. Turns out that if we just call `Peek()` on the wrapped request, then the stdlib TLS listener will return an error `tls.RecordHeaderError` that we can watch for, which is the same one that gets handled for writing the `Client sent an HTTP request to an HTTPS server.` message. We can do the same check on the first 5 bytes to see if it looks like HTTP, and write our own response instead.

Note that `tlsPlaceholderWrapper` is now _always_ registered, and is no longer skipped when performing the wrapping. If I didn't make this change, then the logic in the wrapper would never get called.

We'll probably want to rename `tlsPlaceholderWrapper` to something more meaningful now that it has a secondary purpose 🤷‍♂️

Still `TODO`, I need to figure out how to get the wrapper to read a bit more of the request in the error case to grab the `Host` header to help us perform the redirect. I have a hunch that might not be possible with this current approach, because any attempts to read would trigger the same error since the stdlib TLS listener is still under ours. I'll have to dig into this more another day.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -346,7 +346,7 @@ func (app *App) Start() error {
 					// create HTTP redirect wrapper, which detects if
 					// the request had HTTP bytes on the HTTPS port, and
 					// triggers a redirect if so.
-					ln = NewHttpRedirectListener(ln)
+					ln = &httpRedirectListener{Listener: ln}
 
 					// create TLS listener
 					tlsCfg := srv.TLSConnPolicies.TLSConfig(app.ctx)
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -343,6 +343,11 @@ func (app *App) Start() error {
 				// enable TLS if there is a policy and if this is not the HTTP port
 				useTLS := len(srv.TLSConnPolicies) > 0 && int(listenAddr.StartPort+portOffset) != app.httpPort()
 				if useTLS {
+					// create HTTP redirect wrapper, which detects if
+					// the request had HTTP bytes on the HTTPS port, and
+					// triggers a redirect if so.
+					ln = NewHttpRedirectListener(ln)
+
 					// create TLS listener
 					tlsCfg := srv.TLSConnPolicies.TLSConfig(app.ctx)
 					ln = tls.NewListener(ln, tlsCfg)
--- a/modules/caddyhttp/httpredirectlistener.go
+++ b/modules/caddyhttp/httpredirectlistener.go
@@ -1,0 +1,120 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package caddyhttp
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"net/http"
+	"sync"
+)
+
+// httpRedirectListener is listener that checks the first few bytes
+// of the request when the server is intended to accept HTTPS requests,
+// to respond to an HTTP request with a redirect.
+type httpRedirectListener struct {
+	net.Listener
+}
+
+// NewHttpRedirectListener wraps inner, handling
+func NewHttpRedirectListener(inner net.Listener) net.Listener {
+	l := &httpRedirectListener{Listener: inner}
+	return l
+}
+
+// Accept waits for and returns the next connection to the listener,
+// wrapping it with a httpRedirectConn.
+func (l *httpRedirectListener) Accept() (net.Conn, error) {
+	c, err := l.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+
+	return &httpRedirectConn{
+		Conn: c,
+		r:    bufio.NewReader(c),
+	}, nil
+}
+
+type httpRedirectConn struct {
+	net.Conn
+	once sync.Once
+	r    *bufio.Reader
+}
+
+// Read tries to peek at the first few bytes of the request, and if we get
+// an error reading the headers, and that error was due to the bytes looking
+// like an HTTP request, then we perform a HTTP->HTTPS redirect on the same
+// port as the original connection.
+func (c *httpRedirectConn) Read(p []byte) (int, error) {
+	var errReturn error
+	c.once.Do(func() {
+		firstBytes, err := c.r.Peek(5)
+		if err != nil {
+			return
+		}
+
+		// If the request doesn't look like HTTP, then it's probably
+		// TLS bytes and we don't need to do anything.
+		if !firstBytesLookLikeHTTP(firstBytes) {
+			return
+		}
+
+		// Parse the HTTP request, so we can get the Host and URL to redirect to.
+		req, err := http.ReadRequest(c.r)
+		if err != nil {
+			return
+		}
+
+		// Build the redirect response, using the same Host and URL,
+		// but replacing the scheme with https.
+		headers := make(http.Header)
+		headers.Add("Location", "https://"+req.Host+req.URL.String())
+		resp := &http.Response{
+			Proto:      "HTTP/1.0",
+			Status:     "308 Permanent Redirect",
+			StatusCode: 308,
+			ProtoMajor: 1,
+			ProtoMinor: 0,
+			Header:     headers,
+		}
+
+		err = resp.Write(c.Conn)
+		if err != nil {
+			errReturn = fmt.Errorf("couldn't write HTTP->HTTPS redirect")
+			return
+		}
+
+		errReturn = fmt.Errorf("redirected HTTP request on HTTPS port")
+		c.Conn.Close()
+	})
+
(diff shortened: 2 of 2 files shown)


francislavoie:
Caddyfile I'm using to test this:

```
{
        http_port 8880
        https_port 8881
        admin off
        debug
}

localhost {
        respond "HTTPS"
}
```

Then making these requests:

```
$ curl -vk https://localhost:8881
$ curl -vkL http://localhost:8881
```

First should just show `HTTPS` with no redirect, second should first hit the error case then get served a redirect, then see `HTTPS`.

yroc92:
For what its worth, I'm running into this issue that you're solving and look forward to your fix! I have just one port (`8844`) for a domain being served with TLS in my Caddyfile. When accessing it with `http://example.com:8844` I get back `Client sent an HTTP request to an HTTPS server.`.

I attempted using `redir` when the protocol is `http` but the error is returned before it reaches the directive as you've described.

francislavoie:
One of the open questions I had was how to parse the request headers, and I have my answer by having seen the code in https://github.com/caddyserver/caddy/blob/24fda7514ddeebf19a1e872343b45ec523eefcba/modules/caddyhttp/reverseproxy/fastcgi/client.go#L420, i.e. `textproto.NewReader(rb).ReadMIMEHeader()`. That'll let us grab the `Host` header from the request.

Still need to figure out how to properly wrap the listener so we can actually grab the bytes when we want to without getting an error from the underlying TLS listener.

francislavoie:
I figured out how to do this properly!!

I scrapped the idea of overloading the `tlsPlaceholderWrapper`. Idk what I was thinking, I just didn't understand the pipeline, really.

Instead, I created a new listener `httpRedirectListener` which checks the first 5 bytes of the request (which is fast) and returns early if it doesn't look like HTTP. If it does look like HTTP though, it will parse the request manually from the bytes, grab the `Host` and `URL`, then assemble a HTTP->HTTPS redirect response, write it out, then close the connection.

I always wrap the listener with `httpRedirectListener` just before creating the TLS listener, so that we always have that check in place.

Steps I took to test this are essentially the same. Caddyfile:

```
{
    http_port 8880
    https_port 8881
    admin off
    debug
}

localhost {
    respond "HTTPS"
}
```

Curl request which should show that it redirected to HTTPS, with the header `Location: https://localhost:8881/`
```
$ curl -vkL http://localhost:8881
```

Notice that the debug logs will contain the following on any of these kinds of requests:
```
2021/11/06 04:06:35.627 DEBUG   http.stdlib     http: TLS handshake error from [::1]:49510: redirected HTTP request on HTTPS port
```

This message can probably be adjusted, but I return this error message because otherwise the TLS listener will try to read from the connection which is already closed. I don't think it's possible to quiet it, but either way, no biggie.

mholt:
Thanks for this PR, looks interesting. I'm hoping to dive into this in more detail soon when I can give it some thought. My gut tells me this is something we have to be careful with, or that there will be some unintended side effects or something.

modules/caddyhttp/httpredirectlistener.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return c.r.Read(p)
+}
+
+// firstBytesLookLikeHTTP reports whether a TLS record header
+// looks like it might've been a misdirected plaintext HTTP request.
+func firstBytesLookLikeHTTP(hdr []byte) bool {
+	switch string(hdr[:5]) {
+	case "GET /", "HEAD ", "POST ", "PUT /", "OPTIO":
```
mholt, line 116: I don't love this. It's missing PATCH and a bunch of other verbs that happen over the HTTP port (think WebDAV, etc.). Can we instead check to see if it matches the start of a TLS handshake?

mholt (changes requested):
Very nice overall. However, does this actually work? Like, why would a client make an HTTPS request to the same port it's already connecting to if it tried HTTP to it in the first place?

francislavoie:
This is for situations where a different port than 443 is being used for HTTPS. Such as 8443. So if a link such as `example.com:8443` is given and clicked, it would typically default to `http://` by the browser. Then the user would see an error because they made an HTTP request to an HTTPS server. Instead, we can redirect to `https://` to make sure the browser actually makes a valid request.

modules/caddyhttp/httpredirectlistener.go
francislavoie: I just copied this from the stdlib: 

https://github.com/golang/go/blob/2b39d86344608423138b648b98157470d3809ee7/src/net/http/server.go#L3605-L3613

mholt: Hmm, ok. If that's the case, we can probably drop the number of bytes to 3 and check only for GET, HEA, POS, and OPT. We might still want to add QUE if QUERY becomes a thing.

francislavoie: I'm not sure I see the benefit of that. It would just increase the risk of mismatching the request with other protocols that might also start with those bytes. I think sticking to doing it the same was as stdlib has a greater advantage than we'd get by only comparing against two less bytes.

mholt: I still think checking for a TLS handshake specifically is a better solution than hard-coding some HTTP methods. I don't know of any other protocols browsers use that start with those byte sequences, do you?

francislavoie: Checking for it not being TLS would mean we might respond with HTTP to a request that _wasn't_ HTTP (i.e any other garbage request). It only makes sense to respond with HTTP if we actually received HTTP.

> I don't know of any other protocols browsers use that start with those byte sequences, do you?

Not that I know of, but I also can't predict the future, nothing's stopping anyone from writing a plain-text protocol that ends up looking sorta similar to HTTP :man_shrugging: 

Obviously we're talking hypotheticals at this point, but I think there's precedent for just yoinking logic from stdlib and trusting that it makes sense. We don't need to try to be any more clever here than we're already trying to be :sweat_smile:.

francislavoie: The error from stdlib we're trying to preempt with this code is here:

https://github.com/golang/go/blob/2b39d86344608423138b648b98157470d3809ee7/src/net/http/server.go#L1848-L1859

Which explicitly does this exact check, peek the first 5 bytes and see if it's HTTP, and if it is then write a super basic HTTP repsonse `"Client sent an HTTP request to an HTTPS server."`.

We're just _doing the same check_ but slightly earlier so that we can do a _better_ response to this than just saying "nope"

mholt: Ok, you make some valid points for looking for HTTP specifically. However, I think following the logic that this is mainly to benefit web browsers, we can nuke PUT and probably should add QUERY (which seems likely to become standardized).

francislavoie: Can we just wait until Go stdlib makes that change themselves? Again, I find it really unlikely that anything but GET (and maybe, I guess, in edgecases, POST) ever matter here. I don't think we should put `QUERY` in here until it's actually accepted in an RFC. It would just be bitrot otherwise, if it gets rejected.

mholt: Sure, as long as you help maintain it :wink: (which you're very good about, and I appreciate that).

modules/caddyhttp/app.go
```diff
@@ -343,6 +343,11 @@ func (app *App) Start() error {
 				// enable TLS if there is a policy and if this is not the HTTP port
 				useTLS := len(srv.TLSConnPolicies) > 0 && int(listenAddr.StartPort+portOffset) != app.httpPort()
 				if useTLS {
+					// create HTTP redirect wrapper, which detects if
+					// the request had HTTP bytes on the HTTPS port, and
+					// triggers a redirect if so.
+					ln = NewHttpRedirectListener(ln)
```
mholt, line 349: This isn't used outside the package, let's unexport it.

And actually, it doesn't really need to be a function at all. Simply replace with:

```suggestion
					ln = &httpRedirectListener{Listener: ln}
```

modules/caddyhttp/httpredirectlistener.go
```diff
@@ -0,0 +1,120 @@
...
+type httpRedirectListener struct {
+	net.Listener
+}
+
+// NewHttpRedirectListener wraps inner, handling
+func NewHttpRedirectListener(inner net.Listener) net.Listener {
+	l := &httpRedirectListener{Listener: inner}
+	return l
+}
+
```
mholt, line 37:
```suggestion
```

Can remove this; I don't think it's useful to export.

mholt (changes requested):
Ok ok, last pass I think :)

mholt (approved):
LGTM now. Thanks!

PR #4316, opened 2021-08-31
cmd: export CaddyVersion(), Commands()

pymnh:
Initial draft adressing caddyserver/dist#29

Because it's not yet clear where to put this package, I have put all code in one single file. 

As considered by @mholt, I could probably rewrite this as a Caddy module. However I think this would require exporting the "commands" variable from caddycmd, I am not sure if that could be a problem. Please let me know what you think

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -331,7 +331,7 @@ func cmdReload(fl Flags) (int, error) {
 }
 
 func cmdVersion(_ Flags) (int, error) {
-	fmt.Println(caddyVersion())
+	fmt.Println(CaddyVersion())
 	return caddy.ExitCodeSuccess, nil
 }
 
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -61,6 +61,12 @@ type Command struct {
 // any error that occurred.
 type CommandFunc func(Flags) (int, error)
 
+// Commands returns a list of commands initialised by
+// RegisterCommand
+func Commands() map[string]Command {
+	return commands
+}
+
 var commands = make(map[string]Command)
 
 func init() {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -420,7 +420,7 @@ func printEnvironment() {
 	fmt.Printf("caddy.AppDataDir=%s\n", caddy.AppDataDir())
 	fmt.Printf("caddy.AppConfigDir=%s\n", caddy.AppConfigDir())
 	fmt.Printf("caddy.ConfigAutosavePath=%s\n", caddy.ConfigAutosavePath)
-	fmt.Printf("caddy.Version=%s\n", caddyVersion())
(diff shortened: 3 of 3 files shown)


francislavoie:
Do you mind pasting the output of this new command in a Gist or something? Hard to envision what the template's output would look like without running it

mholt:
Yeah, I think an external module would be best for this.

Instead of exporting the `commands` variable, how about a getter function instead?

pymnh:
Of course... 😅 

So, [these](https://salsa.debian.org/go-team/packages/caddy/-/tree/debian/sid/debian/man) are the files I generated for the Debian package. Unfortunately the roff markup language is still rather opaque... 
I rendered some to html pages which resembles the final manual page shown in a terminal emulator much closer,  see [here](https://htmlpreview.github.io/?https://gist.githubusercontent.com/pymnh/58169a08c124c7267506b7854d36ddda/raw/86be98eddd8f9864a54cd0a6c69952f229fc8c99/caddy.1.html) and [here](https://htmlpreview.github.io/?https://gist.githubusercontent.com/pymnh/58169a08c124c7267506b7854d36ddda/raw/86be98eddd8f9864a54cd0a6c69952f229fc8c99/caddy-run.1.html)

pymnh:
> Yeah, I think an external module would be best for this.
> 
> Instead of exporting the `commands` variable, how about a getter function instead?

Sounds good. I will look into that

francislavoie:
Nice! Thanks for the preview!

We'll probably want to commit these to the `/dist` repo as well for the Fedora COPR and our Cloudsmith apt repos to inherit them.

We can probably also add a step in CI to automatically regenerate them for `/dist` as well on release.

pymnh:
Did some clean up and made it a module

Some notes:

I have added `Commands()` to caddycmd. Is that what you were thinking of? I knew Getters as a function to a Type, so I was a bit unsure if I understood right.

I have also exported `caddyVersion()` so that this information can be included in the manpage. Is that alright? 

I also looked at other source files and tried to be comparably explicit with comments throughout the file (maybe even too explicit?)

Please let me know what you think

mholt:
Yep, exporting those two functions like that works, I'd say.

Nice job. I think we should put this module in a new repo. Maybe `caddyserver/manpages` - what do you think?

The changes to Caddy that export the functions can stay or be opened in a new PR.

francislavoie:
It doesn't necessarily need to be in a new repo, we could put it in `/dist` maybe?

mholt:
Sure, that's an option too. Or just dist/manpages.

pymnh:
opened PR here: caddyserver/dist#73

francislavoie (approved):

pymnh:
forgot to go fmt 😬
this should do

mholt (approved):
LGTM. Thanks for the simple change, looking forward to some legit man pages :)

PR #4317, opened 2021-08-31
Enable Dependabot

milgradesec:
This PR enables github dependabot for go.mod dependencies and github actions in workflows.
Scans every day and If it finds outdated deps will create PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


francislavoie:
I think we'd rather manage dependency upgrades ourselves, on our own time. We've been frequently bitten by dependency upgrades causing breakages.

Mohammed90:
Agreeing with @francislavoie here. Also dependabot tends to be pretty spammy, which is rather annoying.

francislavoie:
I'll close this then.

Thanks for the suggestion though!

mholt:
And I wonder if we should even stop upgrading dependencies at all unless we have a reason to. :man_shrugging:

Issue #4314, opened 2021-08-30
Content-Encoding header missing with reverse-proxied chunked response


ueffel:
Affected version: 2.4.4

I noticed that the `Content-Encoding` header is missing on reverse proxied responses when they are big enough so they are transferred in chunks. The response is encoded but the header isn't sent, which results in garbled mess.

Example Caddyfile:

```caddy-d
http://:80 {
	encode zstd gzip
	reverse_proxy 127.0.0.1:8080
}

http://:8080 {
	respond / "Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
Long enough response, that would get compressed by some of the encoder modules and transferred chunked.
"
	header Content-Type text/plain;charset=utf8
}
```

Request / Response:
```shell
$ curl -v --compressed -o /dev/null -s localhost
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> Host: localhost
> User-Agent: curl/7.78.0
> Accept: */*
> Accept-Encoding: deflate, gzip, br, zstd
>
[...]

PR #4318, opened 2021-08-31
encode: ignore flushing until after first write

ueffel:
The first write will determine if encoding has to be done and will add an Content-Encoding. Until then Flushing has to be delayed so the Content-Encoding header can be added before headers and status code is written. (A passthrough flush would write header and status code)

Fixes https://github.com/caddyserver/caddy/issues/4314

As of now, no integration test included

### DIFF
--- a/modules/caddyhttp/encode/encode.go
+++ b/modules/caddyhttp/encode/encode.go
@@ -200,10 +200,10 @@ func (enc *Encode) Match(rw *responseWriter) bool {
 // until headers were written.
 func (rw *responseWriter) Flush() {
 	if !rw.wroteHeader {
-		// flushing the underlying ResponseWriter will write header and status code
+		// flushing the underlying ResponseWriter will write header and status code,
 		// but we need to delay that until we can determine if we must encode and
-		// therefore add the Content-Encoding header
-		// this happens in the first call to rw.Write
+		// therefore add the Content-Encoding header; this happens in the first call
+		// to rw.Write (see bug in #4314)
 		return
 	}
 	rw.ResponseWriterWrapper.Flush()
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/encode/encode.go
```diff
@@ -195,6 +196,17 @@ func (enc *Encode) Match(rw *responseWriter) bool {
 	return enc.Matcher.Match(rw.statusCode, rw.Header())
 }
 
+func (rw *responseWriter) Flush() {
```
francislavoie, line 199: I think this needs a proper godoc comment (i.e. `// Flush flushes the response. etc.`)

francislavoie (approved):
I've tested this fix the same way that I tried it last night according to your report https://github.com/caddyserver/caddy/issues/4314#issuecomment-908829826, and I can confirm it fixed that case.

LGTM!

If you find the time later to add an integration test, that would be great, but I don't think we need to wait on merging this for the tests.

modules/caddyhttp/encode/encode.go
```diff
@@ -195,6 +196,19 @@ func (enc *Encode) Match(rw *responseWriter) bool {
...
 }
 
+// Flush implements http.Flusher. It delays the actual Flush of the underlying ResponseWriterWrapper
+// until headers were written.
+func (rw *responseWriter) Flush() {
+	if !rw.wroteHeader {
+		// flushing the underlying ResponseWriter will write header and status code
+		// but we need to delay that until we can determine if we must encode and
+		// therefore add the Content-Encoding header
+		// this happens in the first call to rw.Write
```
mholt, line 206:
```suggestion
		// flushing the underlying ResponseWriter will write header and status code,
		// but we need to delay that until we can determine if we must encode and
		// therefore add the Content-Encoding header; this happens in the first call
		// to rw.Write (see bug in #4314)
```

mholt (changes requested):
This LGTM, thanks so much for identifying the problem and implementing a solid fix.

I just request that we link to the bug in the comment since it may not be obvious why this could be a problem. :+1:

mholt (approved):

PR #4319, opened 2021-08-31
ci: Only test cross-build on latest Go version

francislavoie:
This generated way too many test jobs, which weren't really that useful. Cross-build is just to keep us posted on which architectures are building okay, so it's not necessary to do it twice. Only plan9 is not working at this point (see https://github.com/caddyserver/caddy/issues/3615)


mholt (approved):
Yes, thank you, haha. (Is there a shortcut for "latest Go version"?)

francislavoie:
There is, but I rather update everything at the same time (including `go.mod` etc). I don't want random CI failures from using `latest` when the latest Go version is released while requiring some minor changes for us to support (e.g. `quic-go` usually needs to explicitly support new versions due to the `crypto/tls` forking)

Mohammed90 (approved):
Ah, so much lighter. Thanks, Francis!

PR #4328, opened 2021-09-03
ci: revert workaround implemented in #4306

Mohammed90:
GoReleaser v0.177.0 fixes the issue, so the workaround is no longer needed.


francislavoie (approved):
Thanks for the follow-up!

PR #4342, opened 2021-09-11
fastcgi: Fix Caddyfile parsing when `handle_response` is used

### LINKED ISSUE #4322: Unable to set root inside 'handle_response' block which is in turn inside 'php_fastcgi' directive
Hi, I tried to set up a custom error page based on response status from 'php_fastcgi'. My first attempt was like this:

```
php_fastcgi app:9000 {
  @error status 4xx
  handle_response @error {
    root * /errors
    file_server
  }
}
```

Since php_fastcgi directive is just a wrapper over a reverse proxy and (as documented [here](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#intercepting-responses)) I expected it should work but got error instead:

`Error during parsing: unrecognized directive: /errors`

Removing the line `root * /errors` solved the problem, but I needed to point 'file_server' to another directory from web root.

So I decided to expand the 'php_fastcgi' directive into respective 'reverse_proxy':

```
@index file index.php
rewrite @index {http.matchers.file.relative}

@php path *.php
reverse_proxy @php app:9000 {
    transport fastcgi {
        split .php
    }

    @error status 4xx
    handle_response @error {
        root * /errors
        file_server
    }
}
```

This worked as expected. I think 'php_fastcgi' directive has a bug or am I missing something?

francislavoie:
Fixes #4322

Super simple, actually. The problem was that if `handle_response` contained a `root` directive, then it would actually get consumed ahead of time as a subdirective for `php_fastcgi`, breaking the tokens, triggering an error when parsing when passed down to `reverse_proxy`.

The fix is to only read `php_fastcgi` subdirectives if the nesting is `1`, so we skip any that might appear within any sub-blocks.

The test fails before the change, passes after.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/php_fastcgi_handle_response.txt
+++ b/caddytest/integration/caddyfile_adapt/php_fastcgi_handle_response.txt
@@ -1,0 +1,145 @@
+:8881 {
+	php_fastcgi app:9000 {
+		env FOO bar
+
+		@error status 4xx
+		handle_response @error {
+			root * /errors
+			rewrite * /{http.reverse_proxy.status_code}.html
+			file_server
+		}
+	}
+}
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":8881"
+					],
+					"routes": [
+						{
+							"match": [
+								{
+									"file": {
+										"try_files": [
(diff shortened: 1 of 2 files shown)


mholt (approved):
Fascinating. Thanks for the elegant fix and test case, too!

Issue #4344, opened 2021-09-12
Improve `php_fastcgi` options for `try_files` overrides


ttys3:
purpose:

I have a `php-fpm` site using nginx and now I replaced it with caddy2.

The site does **not** have a `single entrypoint` to handle all requests.

(I want to make it clear: for `single entrypoint` apps, current caddy implements works perfect-_-)

at first, I simply use `php_fastcgi 127.0.0.1:9000`, and it works. but the 404 status did not passed to the `handle_errors` handler, it was eaten all by `php_fastcgi`

---------------------------------------------------------

## the original try

my first config:

```
# listen http and https both for example.com
{$SITE_ADDRESS:http://example.com, https://example.com} {
	root * /app/public

	# https://caddyserver.com/docs/caddyfile/directives/php_fastcgi
	php_fastcgi 127.0.0.1:9000

	file_server


	handle_errors {
		rewrite * /error.html
		templates
		file_server
	}
}
```
for example, any 404 pages that should happened just passed all to index.php and in actual no 404 happend, urls like:

```
/this-is-not-found.html
/blah-this-is-another-notfound/xxxxxx
/xxxxxxx
/sub/xxx/sub/xxxxx
```
this is not my desired result.

---------------------------------------------------------

## first attempt try 

So I checked the doc and found that is has an `index` option, so I changed `php_fastcgi 127.0.0.1:9000` to 
```
php_fastcgi 127.0.0.1:9000 {
    index off
}
```
but this caused another problem: when user access `/`, it can not serve `/index.php`. this is a problem.

---------------------------------------------------------

## second attempt try

so I copied the directive from the doc `Expanded form`, and changed it like this:

the change I made is changed
`try_files {path} {path}/index.php index.php`

to 

`
try_files {path} {path}/index.php
`

the full config:

```
# listen http and https both for example.com
{$SITE_ADDRESS:http://example.com, https://example.com} {
	root * /app/public

	# https://caddyserver.com/docs/caddyfile/directives/php_fastcgi
	# php_fastcgi 127.0.0.1:9000
	route {
		# Add trailing slash for directory requests
		@canonicalPath {
			file {path}/index.php
			not path */
		}
		redir @canonicalPath {path}/ 308

		# If the requested file does not exist, try index files
		@indexFiles file {
			# try_files {path} {path}/index.php index.php
			try_files {path} {path}/index.php
			split_path .php
		}
		rewrite @indexFiles {http.matchers.file.relative}

[...]

PR #4346, opened 2021-09-12
caddyhttp: Add support for triggering errors from `try_files`

francislavoie:
See #4344, this is a first step towards completing that issue.

Nginx supports a couple pieces of functionality in its implementation of `try_files` that allow handling the request either as an HTTP error, or by routing to a named location block. http://nginx.org/en/docs/http/ngx_http_core_module.html#try_files

For Caddy, I think it would be quite beneficial to add support to `try_files` for triggering an error as a fallback case, to short-circuit the request handling early. In many cases, it's not simply the case of "try_files matches, or doesn't".

The one potentially contentious bit here, implementation-wise, is the usage of a placeholder `http.matchers.error` as a way to carry the error from the matcher logic up to the HTTP handler chain logic. I'm hoping that this would have a minimal performance impact, but I'm not totally certain, haven't benchmarked. It's also I guess kinda cross-cutting concerns. But I don't really see a cleaner way to do this. We can't use `r.Context()` directly, because that's immutable at this point, so the replacer, being mutable, is the mechanism that would let us do it.

Anyways, an example of how this would look:

```
:8080 {
	root * /srv
	try_files {path} {path}/index.php =404
	php_fastcgi localhost:9000
	file_server

	handle_errors {
		respond "Error!"
	}
}
```

Given the files: `/srv/foo.js`, `/srv/index.php`, `/srv/bar.php`

```
$ curl localhost:8080/foo.js
foo.js contents

$ curl localhost:8080/
index.php executed

$ curl localhost:8080/index.php
index.php executed

$ curl localhost:8080/bar.php
bar.php executed

$ curl localhost:8080/foo.php
Error!
```

Whereas without this, `curl localhost:8080/foo.php` would be sent through fastcgi when the file is already known not to exist.

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -986,6 +986,11 @@ func (mre *MatchRegexp) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 var wordRE = regexp.MustCompile(`\w+`)
 
 const regexpPlaceholderPrefix = "http.regexp"
+
+// MatcherErrorVarKey is the key used for the variable that
+// holds an optional error emitted from a request matcher,
+// to short-circuit the handler chain, since matchers cannot
+// return errors via the RequestMatcher interface.
 const MatcherErrorVarKey = "matchers.error"
 
 // Interface guards
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -986,7 +986,7 @@ func (mre *MatchRegexp) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 var wordRE = regexp.MustCompile(`\w+`)
 
 const regexpPlaceholderPrefix = "http.regexp"
-const MatcherErrorPlaceholder = "http.matchers.error"
+const MatcherErrorVarKey = "matchers.error"
 
 // Interface guards
 var (
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -986,6 +986,7 @@ func (mre *MatchRegexp) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 var wordRE = regexp.MustCompile(`\w+`)
 
 const regexpPlaceholderPrefix = "http.regexp"
+const MatcherErrorPlaceholder = "http.matchers.error"
 
 // Interface guards
 var (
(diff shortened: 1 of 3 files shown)


francislavoie:
While reading through the code, I was reminded that `GetVar(ctx, key)` and `SetVar(ctx, key, val)` exist, which seems more appropriate for this than using the replacer. Made the change.

ttys3:
The first example has a typo, 

is it `try_files {path} {path}/index.php =404` should be put in a block ?:

```
:8080 {
	root * /srv
	php_fastcgi localhost:9000 {
       try_files {path} {path}/index.php =404
    }
	file_server

	handle_errors {
		respond "Error!"
	}
}
```

francislavoie:
No, I did that on purpose because _this_ PR doesn't add `try_files` support to `php_fastcgi`. That's the job of the other PR.

ttys3:
Got it. I like the implementation. This make `try_files` has more power ( and the `php_fastcgi` also benefit from it).

`try_files`, if matched, rewrite it, otherwise trigger an error with code after the `=` char, Am I correct in this way?

francislavoie:
Yeah, pretty much. It will rewrite to the first argument that matches a file that exists on disk, or immediately emit an error if one of the arguments looks like `=[0-9]{3}` essentially

modules/caddyhttp/matchers.go
```diff
@@ -986,6 +986,7 @@ func (mre *MatchRegexp) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 var wordRE = regexp.MustCompile(`\w+`)
 
 const regexpPlaceholderPrefix = "http.regexp"
+const MatcherErrorVarKey = "matchers.error"
```
mholt, line 989: This should probably have a comment explaining what it is and how to use it.

mholt (changes requested):
Ok, this actually looks alright. Thanks for the contribution!

modules/caddyhttp/matchers.go
francislavoie: Right - will do

francislavoie: Done!

mholt (approved):
LGTM. I still think it's good that matchers don't return errors, especially since returning an error can be an optional utility like in this case, this approach seems better than actual error return value.

PR #4347, opened 2021-09-12
fastcgi: Implement `try_files` override in Caddyfile directive

francislavoie:
See #4344, this is part 2 of completing that issue. This should be merged _after_ #4346.

This makes it possible to override the `try_files` logic in the `php_fastcgi` directive, without having to implement a `try_files` to run _before_ `php_fastcgi` (plenty of examples of users doing this on the forums, where their PHP apps didn't quite conform to what they needed).

It makes the following config possible:

```
:8080 {
	root * /srv
	php_fastcgi localhost:9000 {
		try_files {path} {path}/index.php =404
	}
	file_server

	handle_errors {
		respond "Error!"
	}
}
```

### DIFF
--- a/caddytest/integration/caddyfile_adapt/php_fastcgi_try_files_override.txt
+++ b/caddytest/integration/caddyfile_adapt/php_fastcgi_try_files_override.txt
@@ -1,0 +1,124 @@
+:8884
+
+php_fastcgi localhost:9000 {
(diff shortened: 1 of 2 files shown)


francislavoie:
@mholt with #4346 merged, this is ready to go

mholt (approved):
Great, thanks for the test.

Issue #4341, opened 2021-09-11
Active health check not separated for same upstream


ulrichSchreiner:
hi,

i already posted a question here (https://caddy.community/t/healthcheck-not-working-with-multiple-servers/13563) and the answer was to create a feature request (if needed). in the forum there is also an example `caddy.json` .

first my problem: i use caddy as a reverse proxy in front of another reverse proxy. the "backend" proxy has one IP address and distinguishes the requests with different HOST-headers to the real backend. in my concrect environment the backend is a nginx which is the default ingress controller in a K8S cluster.

for some hosts we place a caddyserver in front because caddy has some nice plugins and features (and a really nice plugin API);  my caddy config has different servers configured which all have the same upstream but use different settings for the health-check: the healthcheck overwrite the "Host" header so the backend-proxy can delegate to the correct backend system.

but although i have configured two healthchecks, caddy only uses one of them to get the healthy state of the upstream. in my question in the forum (https://caddy.community/t/healthcheck-not-working-with-multiple-servers/13563) i created an example with `httpbin.org` and two different healthcheck-URI's; this is different to my private usecase where i use host-headers but both have the same result: caddy only uses one of them, altough i configured two different healthchecks.

i'm not sure if i should request new feature or if it is a bug. imho it should be possible to chain two reverse-proxies, simply think of caddy-1 behind caddy-2. the caddy-2 must have the chance to mirror all the configured hosts of caddy-1 with a correspoding healtcheck.

what is my current workaround? well as i provision my servers dynamically, i can create hosts-entries for my backends with the same IP address. so i create host-entries "server-1", "server-2", ... all with the same IP and i reference them as upstreams. this works, but is really a hack. and i'm not sure i can do this in every environment.

i think if one has multiple server-blocks in `caddy.json` and every server has its own upstreams and its own healthchecks these should be separated. or you should at least make it clear in the documentation that multiple upstreams with the same `dial` address take one random healtheck, perhaps not the one you configured in the `server`-block. so it can happen that caddy marks your server as "healthy" although in realilty it is unhealthy.

francislavoie:
> Despite appearing in two places in the config, they are the exact same host -- they are **not** distinct -- so really, it is not possible for it to be healthy and unhealthy (not down, and down) at the same time. 

[...]

PR #4350, opened 2021-09-13
reverseproxy: Add ID field for upstreams

francislavoie:
First step for #4341, Caddyfile support would be needed to complete it. /cc @ulrichSchreiner

This makes it possible to specify upstreams in JSON with an `"id"` field, to disambiguate upstreams with the same dial address:

```json
"upstreams": [
    {
        "id": "second",
        "dial": "httpbin.org:443"
    }
],
```

Given the JSON config from https://caddy.community/t/healthcheck-not-working-with-multiple-servers/13563, with just `"id"` fields added, we see that it works, the internal state has separate upstreams:

```json
$ curl -s localhost:2019/reverse_proxy/upstreams | jq
[
  {
    "id": "second",
    "address": "second",
    "healthy": true,
    "num_requests": 0,
    "fails": 0
  },
  {
    "id": "first",
    "address": "first",
    "healthy": false,
    "num_requests": 0,
    "fails": 0
  }
]
```

To avoid it being a breaking change, I left in the `"address"` field in the admin API, deprecated for now, to be removed at a later date. It's more accurate to call it "ID" even though it could be an ID, a dial address, or a SRV address, depending on how it was configured (i.e. whatever the upstream's `String()` function returns).

### DIFF
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_ids.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_ids.txt
@@ -1,0 +1,46 @@
+:8884
+
+reverse_proxy one|http://localhost two|http://localhost {
+	to three|srv+http://localhost four|srv+http://localhost
+}
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":8884"
+					],
+					"routes": [
+						{
+							"handle": [
+								{
+									"handler": "reverse_proxy",
+									"upstreams": [
+										{
+											"dial": "localhost:80",
+											"id": "one"
+										},
+										{
+											"dial": "localhost:80",
+											"id": "two"
+										},
+										{
+											"id": "three",
+											"lookup_srv": "localhost"
+										},
+										{
+											"id": "four",
+											"lookup_srv": "localhost"
+										}
+									]
+								}
+							]
+						}
+					]
+				}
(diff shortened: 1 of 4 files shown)


francislavoie:
I made a second commit with a proposed way to specify IDs in the Caddyfile, by using `|` as a separator. I consider this part still up for debate though. I think it's an okay solution, but 🤷‍♂️

mholt:
Ok, so I appreciate the thought going into this. I think I want to see about coming up with a different fix though that doesn't require the user to name the upstream separate from the upstream's... well, its name (address, like DNS name or IP). I will try to give it some brain cycles this week.

mholt:
(With all due respect) I will be sunsetting this PR in favor of #4470 which will properly distinguish health check state from upstreams, so IDs will not be needed. (We can reopen if that's not the case for some reason, but I'm pretty sure this will be superseded by that.)

Thank you for working on it. :pray:

francislavoie:
@mholt I don't see how https://github.com/caddyserver/caddy/pull/4470 will distinguish the upstreams from two different `reverse_proxy` modules. There's still no identifier to disambiguate two hosts with the same upstream address in `var hosts = caddy.NewUsagePool()`. That's the crux of this issue.

mholt:
I'm not finished yet. When I am, the hosts in the pool will not be storing their health state. The crux of the problem will be fixed.

PR #4351, opened 2021-09-16
[ADHOC] Dirty hack for fixing the reload.

burton1982:
This will patch the reload and push the stop method into a goroutine
till its ended.


burton1982:
accident

Issue #4353, opened 2021-09-17
file_server browse lists size of symlink instead of size of linked file


bones-was-here:
On our build server we have "-latest" symlinks that always point to the most recent builds (whose filenames include date stamps).  Users downloading builds want to know the size of the file they will download when they click, not the size of the symlink.  Some users have assumed the builds were broken because the symlink size is tiny.  

Previously we were using Apache, which shows the size of the linked file that will be downloaded.

Caddy version is 2.4.5 from the official package for Debian.

francislavoie:
That should also be a pretty easy fix, if `isSymlink(f)` is true, then it can use `os.Stat` to get the real size of the file.

https://github.com/caddyserver/caddy/blob/5fda9610f907482497a7b32abb2ab2f85a4e6906/modules/caddyhttp/fileserver/browsetplcontext.go#L57

HayatoShiba:
Hi, Excuse me, can I work on this?
I have never contributed to oss and I would like to fix this one!

mholt:
Sure, go for it! We have [instructions in our contributing guide](https://github.com/caddyserver/caddy/blob/master/.github/CONTRIBUTING.md#how-to-make-a-pull-request-to-caddy) for how to make a pull request. It may then go through a few iterations of reviews. Good luck!

HayatoShiba:
Thank you very much for your quick response!

bones-was-here:
I see the correct size of the linked file, also 2.7.5 on debian.  Is the linked file inside the web root?

Roan-V:
No, does that matter?

Root is set to a "files" folder
Symlinks to files somewhere else on the disk are put inside that folder

PR #4354, opened 2021-09-18
fileserver: Fix displayed file size if it is symlink

HayatoShiba:
Fix: [#4353](https://github.com/caddyserver/caddy/issues/4353)

### DIFF
--- a/modules/caddyhttp/fileserver/browsetplcontext.go
+++ b/modules/caddyhttp/fileserver/browsetplcontext.go
@@ -27,7 +27,7 @@ import (
 	"github.com/dustin/go-humanize"
 )
 
-func (fsrv *FileServer) directoryListing(files []os.FileInfo, canGoUp bool, root, urlPath string, repl *caddy.Replacer) browseTemplateContext {
+func (fsrv *FileServer) directoryListing(files []os.FileInfo, canGoUp bool, root, urlPath string, repl *caddy.Replacer) (browseTemplateContext, error) {
 	filesToHide := fsrv.transformHidePaths(repl)
 
 	var dirCount, fileCount int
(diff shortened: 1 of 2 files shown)


francislavoie:
Looks pretty good! Thanks!

Don't forget to sign the CLA!

HayatoShiba:
Thanks for your review!
and I forgot so have signed CLA:)

modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -52,11 +52,21 @@ func (fsrv *FileServer) directoryListing(files []os.FileInfo, canGoUp bool, root
 			fileCount++
 		}
 
+		isSymlink := isSymlink(f)
```
mholt, line 55: Though technically OK, I don't think shadowing a function name with a local variable name is a good idea, as it can easily be confusing. When I initially looked at line 67 below, I was trying to figure out why passing a function value was working, then I finally saw this, that it was a local variable in this particular spot. Let's rename the variable to something different.

mholt (changes requested):
Very nice -- just one little nit. Thank you!

modules/caddyhttp/fileserver/browsetplcontext.go
francislavoie: I suggest `fileIsSymlink`

HayatoShiba: I see, thanks for the suggestion!
I fix this.

francislavoie (approved):
Thanks!

mholt (approved):
Great, thank you very much for the contribution. Welcome to the project!

PR #4356, opened 2021-09-18
fileserver: purple color for visited links in browse

SlavikCA:
Currently file browser template have all A link (visited and unvisited) in the same color (blue).

This PR make visited A links to be displayed in purpled color. 
Convenient to keep track of visited folders & opened files.


mholt (approved):
Alrighty, we'll give this a shot. Thanks :+1:

PR #4357, opened 2021-09-19
templates: Add tests for funcInclude and funcImport

rockorager:
Added unit tests for funcInclude and funcImport, related to PR #4321


modules/caddyhttp/templates/tplcontext_test.go
```diff
@@ -91,6 +91,140 @@ func TestCookie(t *testing.T) {
...
+		} else if test.shouldErr {
+			t.Errorf("Test %d: Expected error but had none", i)
+		} else if !templateWasDefined && actual != "" {
+			// template should be defined, return value should be an empty string
+			t.Errorf("Test %d: Expected template %s to be define but got %s", i, test.expect, context.tpl.DefinedTemplates())
+
+		}
+
+		if absFilePath != "" {
+			if err := os.RemoveAll(absFilePath); err != nil && !os.IsNotExist(err) {
```
mholt, line 144: Wonder if this would be safer to use `os.Remove` just in case there's a bug (even in a future change) that would end up deleting a whole directory or something unintentionally... also in other cases of RemoveAll in this change.

mholt (changes requested):
Awesome, I love test cases! Thanks for this.

modules/caddyhttp/templates/tplcontext_test.go
rockorager: Good call. Done and done!

mholt (approved):
Awesome, thank you!

PR #4364, opened 2021-09-28
style: move from deprecated ioutil to os and io packages

kallydev:
Since Caddy is currently using Go 1.16 and CI using Go 1.17, so it is safe to remove deprecated `ioutil` package, ref https://golang.org/doc/go1.16#ioutil. The relevant parameters within `.golangci.yml` have also been removed and it is no longer needed.


simaotwx (approved):

mholt (approved):
Nice, thanks. This is a change I had been wanting to make but didn't get around to yet since we bumped to Go 1.16 minimum. Appreciate it!

PR #4369, opened 2021-10-03
fix: retain and restore terminal mode on Windows

### LINKED ISSUE #4251: Running Caddy breaks terminal, ANSI escape codes are printed
After running Caddy with any command, the terminal breaks.
Non-printable keys print their ANSI escape sequence instead, (e.g. up arrow prints `^[[A` and ctrl+c prints `^C`).

image

OS: Windows Server 2016 Datacenter (Version 10.0.14393, Build 14393)
Caddy Version: v2.4.3 h1:Y1FaV2N4WO3rBqxSYA8UZsZTQdN+PwcoOcAiZTM8C0I=

This happens with both Command Prompt and PowerShell (v5).
I have a separate installation of PowerShell 7 that is unaffected.

This looks like the same issue as #4212

Mohammed90:
Given this is caused by a dependency, not possible to remove the dependency, and not easy to excise the code from the dependency due to their needs, I'm attempting to retain the original terminal mode bits and to be set before exit. Bitwise OR operations aren't reversible, so retaining the original state is the other option.

Fixes #4251

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -85,7 +85,9 @@ func Main() {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%s: %v\n", subcommand.Name, err)
 	}
-
+	if err := cleanup(); err != nil {
+		fmt.Fprintf(os.Stderr, "error restoring console to functional state: %v\n", err)
+	}
 	os.Exit(exitCode)
 }
 
--- a/cmd/cleanup_other.go
+++ b/cmd/cleanup_other.go
@@ -1,0 +1,6 @@
+//go:build !windows
+// +build !windows
+
+package caddycmd
+
+func cleanup() error { return nil }
--- a/cmd/cleanup_windows.go
+++ b/cmd/cleanup_windows.go
@@ -1,0 +1,24 @@
+//go:build windows
+// +build windows
+
+package caddycmd
+
+import "golang.org/x/sys/windows"
+
+// Retain the original mode of the console to restore later.
+// See: https://github.com/caddyserver/caddy/issues/4251
+// The evaluation of this takes place before calling init()
+// ref: https://golang.org/ref/spec#Order_of_evaluation
+var inMode, outMode uint32
+var _ = windows.GetConsoleMode(windows.Stdin, &inMode)
+var _ = windows.GetConsoleMode(windows.Stdout, &outMode)
+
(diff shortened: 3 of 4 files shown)


francislavoie:
Thanks, looks reasonable. Just need someone affected to confirm it fixes it.

cmd/main.go
```diff
@@ -85,7 +85,9 @@ func Main() {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%s: %v\n", subcommand.Name, err)
 	}
-
+	if err := cleanup(); err != nil {
```
egonelbre, line 88: This should be called before any `os.Exit` otherwise it could still mess up things.

Mohammed90: I've moved it to the top of the function to just start with clean slate

cmd/main.go
```diff
@@ -50,6 +50,9 @@ func init() {
 // Main implements the main function of the caddy command.
 // Call this if Caddy is to be the main() of your program.
 func Main() {
+	if err := resetTerminalState(); err != nil {
```
egonelbre, line 53: Doesn't this disable the terminal state altogether? Wouldn't that break the dependency -- in case it's being used?

Mohammed90: My understanding of their code is they're using it because they have interactive TUI. Running `go mod why` against the offending dependency shows:
```
2021-10-04 22:57:09  ~/projects/caddyserver/caddy git:(fix-4251) ✗ $ go mod why go.step.sm/cli-utils/ui
# go.step.sm/cli-utils/ui
github.com/caddyserver/caddy/v2/modules/caddypki
github.com/smallstep/certificates/authority
github.com/smallstep/certificates/kms
github.com/smallstep/certificates/kms/softkms
go.step.sm/cli-utils/ui
```
`softkms` uses the `go.step.sm/cli-utils/ui` package to prompt for password. The only password prompt Caddy has is on run of `caddy trust` which uses `github.com/FiloSottile/mkcert` which doesn't use `go.step.sm/cli-utils/ui` for the prompt.

egonelbre: Ok, just making sure.

mholt:
Thanks for the patch; as discussed in Slack, we're going to wait on this until we can determine if upstream can fix it first.

PR #4389, opened 2021-10-18
caddytls: Make peer certificate verification pluggable

### LINKED ISSUE #2341: Allow revocation list for client cert auth
Hello, could caddy add an option to check client certificate against a revocation list ?
This was the object of https://github.com/mholt/caddy/issues/1904, which it was closed as duplicate of https://github.com/mholt/caddy/issues/1375, but https://github.com/mholt/caddy/issues/1375 doesn't seems to implement that.
Best,

v0.11.0

Disallow revoked client certs

*Not relevant, there's no option to do that as per https://caddyserver.com/docs/tls*

Gr33nbl00d:
Extension needed to fix #2341 by providing a plugin using this namespace to extend caddy with revocation checking
Needs review and will be discussed in #2341 before merge


mholt:
Will circle back around to this soon, thanks for maintaining it!

modules/caddytls/connpolicy.go
```diff
@@ -298,6 +311,13 @@ type ClientAuthentication struct {
 	// which are not in this list will be rejected.
 	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
 
+	//List of client certificate validators used to verify the cert
+	//Validators can be used to add additional checks like revocation checks
+	ValidatorsRaw []json.RawMessage `json:"validators,omitempty" caddy:"namespace=tls.client_cert_validators inline_key=validator"`
+
+	//Decoded Validators
+	Validators []ClientCertValidator `json:"-"`
```
francislavoie, line 319:
```suggestion
	// A list of client certificate validators, for additional
	// verification. These can perform custom checks, like ensuring
	// the certificate is not revoked.
	ValidatorsRaw []json.RawMessage `json:"validators,omitempty" caddy:"namespace=tls.client_cert_validators inline_key=validator"`

	Validators []ClientCertValidator `json:"-"`
```

Comments should have a space at the start. They get rendered as markdown in our docs on the website, so they should have punctuation. Adjusted the wording a bit.

Also, I'm not sure if we need `Validators` to be exported. If it's not used outside of this package, we can change it to `validators`. The empty `json` field tag is not needed in that case.

modules/caddytls/connpolicy.go
```diff
@@ -378,52 +396,40 @@ func (clientauth *ClientAuthentication) ConfigureTLSConfig(cfg *tls.Config) erro
...
+	clientauth.existingVerifyPeerCert = cfg.VerifyPeerCertificate
+	cfg.VerifyPeerCertificate = clientauth.verifyPeerCertificate
 	return nil
 }
 
-// verifyPeerCertificate is for use as a tls.Config.VerifyPeerCertificate
-// callback to do custom client certificate verification. It is intended
-// for installation only by clientauth.ConfigureTLSConfig().
-func (clientauth ClientAuthentication) verifyPeerCertificate(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+func (clientauth *ClientAuthentication) verifyPeerCertificate(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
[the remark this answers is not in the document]
francislavoie: This function is still missing its godoc comment, it was removed by accident I think. It used to have this comment:

```suggestion
// verifyPeerCertificate is for use as a tls.Config.VerifyPeerCertificate
// callback to do custom client certificate verification. It is intended
// for installation only by clientauth.ConfigureTLSConfig().
func (clientauth *ClientAuthentication) verifyPeerCertificate(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```

modules/caddytls/connpolicy.go
```diff
@@ -461,6 +467,31 @@ func setDefaultTLSParams(cfg *tls.Config) {
 	cfg.PreferServerCipherSuites = true
 }
 
+// LeafVerificationValidator Implements Custom client certificate verification.
+// It is intended for installation only by clientauth.ConfigureTLSConfig()
```
francislavoie, line 471:
```suggestion
// LeafVerificationValidator implements custom client certificate verification.
// It is intended for installation only by clientauth.ConfigureTLSConfig().
```

modules/caddytls/connpolicy.go
```diff
@@ -481,4 +512,10 @@ type ConnectionMatcher interface {
 	Match(*tls.ClientHelloInfo) bool
 }
 
+// ClientCertValidator is a type which validates client certificates
+// It is called during verifyPeerCertificate in tls handshake
```
francislavoie, line 516:
```suggestion
// ClientCertValidator is a type which validates client certificates.
// It is called during verifyPeerCertificate in the TLS handshake.
```

Gr33nbl00d:
Hi Again,
i am back from vacation and will not cycle back to this :)

francislavoie:
> i am back from vacation and will not cycle back to this :)

Sorry, do you mean you're abandoning this PR? I'm confused about the wording here 🤔

Gr33nbl00d:
> > i am back from vacation and will not cycle back to this :)
> 
> Sorry, do you mean you're abandoning this PR? I'm confused about the wording here 🤔

Oh sorry :) That was a typo "now" cycle back to it :)

mholt:
Thanks for this. I'm also considering this PR in conjunction with the discussion at https://github.com/caddyserver/caddy/issues/4518, which might involve some of how client auth is set up anyways. :thinking:

modules/caddytls/connpolicy.go
```diff
@@ -298,6 +311,13 @@ type ClientAuthentication struct {
 	// which are not in this list will be rejected.
 	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
 
+	// A list of client certificate validators, for additional
+	// verification. These can perform custom checks, like ensuring
+	// the certificate is not revoked.
+	ValidatorsRaw []json.RawMessage `json:"validators,omitempty" caddy:"namespace=tls.client_cert_validators inline_key=validator"`
+
+	Validators []ClientCertValidator `json:"-"`
```
mholt, line 319: Can we unexport this? (and remove the `json` tag) -- that's ideal, if possible; we should only export these if modified directly outside the package, which is rare

modules/caddytls/connpolicy.go
```diff
@@ -298,6 +311,13 @@ type ClientAuthentication struct {
 	// which are not in this list will be rejected.
 	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
 
+	// A list of client certificate validators, for additional
+	// verification. These can perform custom checks, like ensuring
+	// the certificate is not revoked.
+	ValidatorsRaw []json.RawMessage `json:"validators,omitempty" caddy:"namespace=tls.client_cert_validators inline_key=validator"`
```
mholt, line 317:
```suggestion
	ValidatorsRaw []json.RawMessage `json:"validators,omitempty" caddy:"namespace=tls.client_auth inline_key=validator"`
```

Just a bit shorter, still descriptive

modules/caddytls/connpolicy.go
```diff
@@ -298,6 +311,13 @@ type ClientAuthentication struct {
 	// which are not in this list will be rejected.
 	TrustedLeafCerts []string `json:"trusted_leaf_certs,omitempty"`
```
mholt, line 312: This is basically for setting up a LeafCertValidator right? I wonder if we should deprecate this field or something. (Pending on #4518)

modules/caddytls/connpolicy.go
```diff
@@ -461,6 +470,31 @@ func setDefaultTLSParams(cfg *tls.Config) {
 	cfg.PreferServerCipherSuites = true
 }
 
+// LeafVerificationValidator implements custom client certificate verification.
+// It is intended for installation only by clientauth.ConfigureTLSConfig().
+type LeafVerificationValidator struct {
```
mholt, line 475:
```suggestion
type LeafCertValidator struct {
```

or maybe better:

```suggestion
type LeafCertClientAuth struct {
```

But let's see what we decide in #4518 first

mholt (changes requested):
Overall, I like the idea and the implementation. I would like to review this in more detail when I have time, but overall I'm not opposed to this approach.

We should see how #4518 plays out before committing to this, though, since I want to make sure we get client auth right.

modules/caddytls/connpolicy.go
Gr33nbl00d: Done :)

modules/caddytls/connpolicy.go
Gr33nbl00d: Yes, good point currently the LeafCertValidator is added programatically if TrustedLeafCerts are not empty (to be backward compatible)
But it would be an idea to extract the LeafCertValidator to an own Module so this configuration would then not be inside ClientAuthentication anymore but inside the Validators own config. Could also be extracted outside caddy.

mholt:
Thanks for keeping this updated. Still interested in this change, but need to see what the outcome / resulting consensus on #4518 is...

mholt:
CI tests seem broken / missing on this so I'm going to try closing and reopening.

modules/caddytls/connpolicy.go
```diff
@@ -55,14 +56,26 @@ func (cp ConnectionPolicies) Provision(ctx caddy.Context) error {
 		if err != nil {
 			return fmt.Errorf("connection policy %d: building standard TLS config: %s", i, err)
 		}
+
+		if pol.ClientAuthentication != nil {
+			if len(pol.ClientAuthentication.ValidatorsRaw) > 0 {
```
mholt, line 61:
```suggestion
		if pol.ClientAuthentication != nil && len(pol.ClientAuthentication.ValidatorsRaw) > 0 {
```

modules/caddytls/connpolicy.go
```diff
@@ -481,4 +515,10 @@ type ConnectionMatcher interface {
 	Match(*tls.ClientHelloInfo) bool
 }
 
+// ClientCertValidator is a type which validates client certificates.
+// It is called during verifyPeerCertificate in the TLS handshake.
+type ClientCertValidator interface {
```
mholt, line 520:
```suggestion
type ClientCertificateValidator interface {
```

Just wondering what your thoughts are on spelling out the full word here. I feel like that might be more important for the _interface_ being implemented, but I normally don't mind `Cert`, and maybe I even use it elsewhere. Not a big deal, but just thought I should suggest it.

modules/caddytls/connpolicy.go
mholt: I might make a commit to this branch to mark this as deprecated, (or not) -- but I'll play with it a bit first.

mholt (changes requested):
Finally getting back to this. I'll probably push a commit or three soon.

mholt:
Thanks so much for the enhancement, @Gr33nbl00d -- sorry it took me a few months to get around to it. Happy to merge this in!

Gr33nbl00d:
> FYI @Gr33nbl00d I spotted in the [Go 1.19 release notes](https://tip.golang.org/doc/go1.19#crypto/x509) (draft notes, it's not out yet) that they're adding a new function [`ParseRevocationList`](https://tip.golang.org/pkg/crypto/x509/#ParseRevocationList). Is that useful for you? Might perform better or well enough to obviate the need for your custom parser?

I see but according to the method signature they use a byte array as parameter and return a complete list.
This would still mean that the whole file is loaded into memory...

PR #4391, opened 2021-10-19
reverseproxy: New `copy_response` handler for `handle_response` routes

francislavoie:
Followup to #4298 and #4388.

This adds a new `copy_response` handler which may only be used in `reverse_proxy`'s `handle_response` routes, which can be used to actually copy the proxy response downstream. 

Previously, if `handle_response` was used (with routes, not the status code mode), it was impossible to use the upstream's response body at all, because we would always close the body, expecting the routes to write a new body from scratch.

To implement this, I had to refactor `h.reverseProxy()` to move all the code that came after the `HandleResponse` loop into a new function. This new function `h.finalizeResponse()` takes care of preparing the response by removing extra headers, dealing with trailers, then copying the headers and body downstream.

Since basically what we want `copy_response` to do is invoke `h.finalizeResponse()` at a configurable point in time, we need to pass down the proxy handler, the response, and some other state via a new `req.WithContext(ctx)`. Wrapping a new context is pretty much the only way we have to jump a few layers in the HTTP middleware chain and let a handler pick up this information. Feels a bit dirty, but it works.

Also fixed a bug with the `http.reverse_proxy.upstream.duration` placeholder, it always had the same duration as `http.reverse_proxy.upstream.latency`, but the former was meant to be the time taken for the roundtrip _plus_ copying/writing the response.

Example config, `copy_response` here will make sure the upstream's response body of `{"hello": "world"}` gets copied to the client. Previously, you had to write a new response with `respond` or `file_server` etc:

```
{
	admin off
	debug
}

:8881 {
	reverse_proxy 127.0.0.1:8882 {
		@200 status 200
		handle_response @200 {
			header Foo bar
			copy_response
		}
	}
}

:8882 {
	respond `{"hello": "world"}` 200
}
```

TODO: I just noticed that if no handler that writes a body is used, the `Content-Length` header from the upstream is still copied during `h.finalizeResponse()`, so we end up with a mismatch between that header and the actual body in the response (being empty). Maybe we should remove the `Content-Length` header (or set it to zero? idk) specifically if `bodyClosed` is true. Not sure yet.

Edit: Fixed the `Content-Length` issue.

### DIFF
--- a/modules/caddyhttp/reverseproxy/copyresponse.go
+++ b/modules/caddyhttp/reverseproxy/copyresponse.go
@@ -31,7 +31,7 @@ func init() {
 
 // CopyResponseHandler is a special HTTP handler which may
 // only be used within reverse_proxy's handle_response routes,
-// to copy the proxy response.
+// to copy the proxy response. EXPERIMENTAL, subject to change.
 type CopyResponseHandler struct {
 	// To write the upstream response's body but with a different
 	// status code, set this field to the desired status code.
@@ -85,7 +85,8 @@ func (h CopyResponseHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
 
 // CopyResponseHeadersHandler is a special HTTP handler which may
 // only be used within reverse_proxy's handle_response routes,
-// to copy headers from the proxy response.
+// to copy headers from the proxy response. EXPERIMENTAL;
+// subject to change.
 type CopyResponseHeadersHandler struct {
 	// A list of header fields to copy from the response.
 	// Cannot be defined at the same time as Exclude.
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -1272,7 +1272,7 @@ var bufPool = sync.Pool{
 	},
 }
 
-// HandleResponseContext carries some contextual information about the
+// handleResponseContext carries some contextual information about the
 // the current proxy handling.
 type handleResponseContext struct {
 	// handler is the active proxy handler instance, so that


fahadfa:
l3454`

francislavoie:
Alright - I've implemented `copy_response_headers` with include/exclude list support.

```
{
    admin off
    debug
}

:8881 {
    reverse_proxy 127.0.0.1:8882 {
        @200 status 200
        handle_response @200 {
            header Foo foo
            copy_response_headers {
                include Bar Baz
            }
            respond "handled"
        }
    }
}

:8882 {
    header Bar bar
    header Baz baz
    respond `{"hello": "world"}` 200
}
```

The above will copy just the `Bar` header, ignoring the `Content-Type` and the JSON response, writing a new response of `handled` instead.

I ordered the `copy_response_headers` directive _after_ `header`, I'm not sure if that's the right place for it or not. My thought is that since `header` supports `defer`, if the user wants to make things happen after the headers are copied, they can turn on `defer` mode.

rainerborene:
@francislavoie Is it possible to use `copy_response_headers` without specifying each header manually?

francislavoie:
Yes, it supports either `include` or `exclude` or neither lists at all which will copy all of them.

mholt (approved):
Okay, finally looked it over, and honestly, it's not as scary as I thought it would be, ha. Good job.

What if we mark these features as EXPERIMENTAL so we can possibly change/tweak them later if needed?

francislavoie:
Experimental sounds fine to me

modules/caddyhttp/reverseproxy/copyresponse.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+func init() {
+	caddy.RegisterModule(CopyResponseHandler{})
+	caddy.RegisterModule(CopyResponseHeadersHandler{})
+}
+
+// CopyResponseHandler is a special HTTP handler which may
+// only be used within reverse_proxy's handle_response routes,
+// to copy the proxy response.
```
mholt, line 34:
```suggestion
// to copy the proxy response. EXPERIMENTAL, subject to change.
```

modules/caddyhttp/reverseproxy/copyresponse.go
```diff
@@ -0,0 +1,189 @@
...
+	// finalizeResponse again after we've already done it here.
+	hrc.isFinalized = true
+
+	// write the response
+	return hrc.handler.finalizeResponse(rw, req, hrc.response, repl, hrc.start, hrc.logger, false)
+}
+
+// CopyResponseHeadersHandler is a special HTTP handler which may
+// only be used within reverse_proxy's handle_response routes,
+// to copy headers from the proxy response.
```
mholt, line 88:
```suggestion
// to copy headers from the proxy response. EXPERIMENTAL;
// subject to change.
```

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -1227,6 +1272,38 @@ var bufPool = sync.Pool{
 	},
 }
 
+// HandleResponseContext carries some contextual information about the
```
mholt, line 1275:
```suggestion
// handleResponseContext carries some contextual information about the
```

PR #4399, opened 2021-10-27
add placeholder like nginx $upstream_addr

yflau:
nginx ngx_http_upstream_module module has a embed var $upstream_addr, see [details](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#variables) . 

Caddy has a placeholder {http.reverse_proxy.upstream.address}, but when specify an upstream dial with a domain name(without lookup_srv), then the placeholder value will be the domain name while not a IP address, which is more useful for debug.

### DIFF
--- a/caddytest/integration/reverseproxy_test.go
+++ b/caddytest/integration/reverseproxy_test.go
@@ -1,6 +1,7 @@
 package integration
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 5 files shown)


yflau:
> Ok, but if we did this we'd have to make it so that the information is available regardless of the transport.

I make a new extention point caddy.dial_context_wrappers, and users can define their own dialer.DialContext wrappers, so if I need the upstream IP placeholder, I can write a plugin, e.g. upstreamAddrPlaceholderWrapper in test.
And I also make add the support for fastcgi transport, so please have a check if it's OK?

mholt:
Hmm, I don't think adding a new extension point is the right solution here.

Maybe an optional interface is better? I'll have to look at it when I have some time but it's a busy holiday weekend here.

yflau:
> Maybe an optional interface is better? I'll have to look at it when I have some time but it's a busy holiday weekend here.

@mholt Thank you for your reply, and could you please give some more information about the `optional interface` solution?

mholt:
I dunno... this is an awfully big change (approx. 150+ lines to test, 70+ to implement) for something so small -- something that could easily be served by a debug log, I think.

If it's alright with you, I think I'm going to decline this change for now. (In the future, it's helpful to open issues to discuss details before writing code. Unless of course you wrote it for your own needs anyway, and in those cases we appreciate that you shared it with us!) Feel free to continue discussion if needed though.

PR #4401, opened 2021-10-28
caddytls: Mark storage clean timestamp at end of routine

mholt:
See discussion on 42b7134ffa3bf3e9e86514c82407979c2627a5ab

/cc @Gr33nbl00d


Gr33nbl00d:
I think that should do it, 
Only remark is what will happen if there is a panic during cleanup. Not sure we would need to handle this case. 
But if we define a defer function after the condition which actually would set the finishTime we could avoid multiple fast retrys on panic.

mholt:
I don't think multiple fast retries will be a concern here. I could be wrong, but I'd rather wait and see if that's actually the case because timers still call this function.

PR #4403, opened 2021-11-01
caddyhttp: Split up logged remote address into IP and port

### LINKED ISSUE #4396: Separate log field for request port
`request>remote_addr` of `http.log.access` logs are in format of `<ipa>:<port>`. I think the port should be a separate field.

francislavoie:
Closes #4396, related to #4148

This is a breaking change, but we're already planning on removing  the`common_log` field as well at the same time, so might as well all do it together.

In general, the remote IP is the useful part of the remote address. The port is rarely useful, because it only identifies ephemeral information about the connection from the client. But we were logging both in one field, so certain tooling that would want to only get the remote IP would need to split it up. Splitting is non-trivial, because of IPv4, IPv6, shenanigans. So it's best if we split it up-front before logging. If the log consumer actually cares about the remote port, it can re-assemble it.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -659,11 +659,17 @@ type adminHandler struct {
 // ServeHTTP is the external entry point for API requests.
 // It will only be called once per request.
 func (h adminHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ip, port, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		ip = r.RemoteAddr
+		port = ""
+	}
 	log := Log().Named("admin.api").With(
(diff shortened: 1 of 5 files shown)


mholt (approved):
I begrudingly approve. :slightly_smiling_face:

francislavoie:
Tagging it with "do not merge", so we can put it in just before/when we're ready for `2.5.0`

mohammed90 (approved):
Whee! I'll make a note in the `format-encoder` to update it once this is released 😃

mohammed90 (approved):

mohammed90 (approved):

PR #4407, opened 2021-11-08
caddyhttp: Sanitize the path before evaluating path matchers

francislavoie:
I found an issue where it's possible to bypass path matchers by crafting a request with a path that doubles up slashes, or uses URL-encoded characters to smuggle the request past the path matcher.

This becomes a security issue, because auth gates using [`basicauth`](https://caddyserver.com/docs/caddyfile/directives/basicauth) can be bypassed 😬

The solution is to unescape (convert URL-encoded characters) then clean the path, before matching.

This could be a breaking change if exact behaviour of the matcher is required (without cleaning the path). In that case, then it's possible to use the [`expression`](https://caddyserver.com/docs/caddyfile/matchers#expression) matcher to do a direct string comparison or regexp on the `{http.request.uri.path}` placeholder (or simply `{path}` [in the Caddyfile](https://caddyserver.com/docs/caddyfile/concepts#placeholders)). For example:

```
# Simple prefix match
expression `{path}.startsWith("/foo")`

# Complex regexp match
expression `{path}.matches("^/foo.*$")`
```

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"net/textproto"
 	"net/url"
+	"path"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thanks for finding and fixing this. I'm hoping this won't break legitimate use cases, but I'm glad there's a workaround. Will release this today.

karatalaya:
V2.2

karatalaya:
[]()

PR #4410, opened 2021-11-10
fileserver: browse: do not encode the paths in breadcrumbs and page title

### LINKED ISSUE #4409: 2.4.6 regression: file_browser displays path unencoded
**Use case:**
I'm using Caddy to publish school lesson's videos & images via file_server. Here is config:

```
:80
bind 0.0.0.0

root * /var/www/html
file_server browse {
  hide "#snapshot" "@eaDir" *.db
}
```

In version 2.4.5 the name of current folder was displayed correctly on top of the page (the breadcrumb).

**Bug:**
However, in version 2.4.6 it is displayed unencoded. I suspect it caused by security patch and is probably needed. But can the path be correctly encoded and displayed in the UI?

The URL's are changed, too.

The page <title> is unencoded, too

Here are 2 screenshots of 2.4.5 vs 2.4.6:

## **Space in the name of the folder:**
Space

## **Russian characters:**
RussianChars

mohammed90:
Breadcrumbs use the `.Name` field as the visual item while `.Path` is the data-oriented item. Earlier implementation used the URL path as-is, which could have been in escaped format. This PR un-escapes the URL path to use as for the `.Name`.

Fixes #4409

### DIFF
--- a/modules/caddyhttp/fileserver/browsetplcontext.go
+++ b/modules/caddyhttp/fileserver/browsetplcontext.go
@@ -72,9 +72,9 @@ func (fsrv *FileServer) directoryListing(files []os.FileInfo, canGoUp bool, root
 			Mode:      f.Mode(),
 		})
 	}
-
+	name, _ := url.PathUnescape(urlPath)
 	return browseTemplateContext{
-		Name:     path.Base(urlPath),
+		Name:     path.Base(name),
 		Path:     urlPath,
 		CanGoUp:  canGoUp,
 		Items:    fileInfos,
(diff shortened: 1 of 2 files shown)


francislavoie:
@SlavikCA would you be able to build from this branch to confirm the fix? Thanks!

francislavoie (approved):
Seems good to me!

mholt:
@SlavikCA Can you confirm this fixes it for you?

SlavikCA:
Thank you very much!

Yes, just tested  and both page title and breadcrumbs looks good now.

Issue #4348, opened 2021-09-12
http/3 seems to stop working after any kind of reload


Zierhut-IT:
http/3 seems to stop working after any kind of reload.

- http/3 works as expected before reloading
- I went as far as v2.0.0 back (thus the two Caddyfiles, see further down) and encountered this issue every time
- Caddy seems to still listen on `udp/443` when it's happening:
    ```bash
    $ ss --listening --udp --processes # -lup
    UNCONN 0 0 *:https *:* users:(("caddy",pid=872477,fd=8))
    ```
- Verbose `curl -vvv` logs:

	```bash
    $ curl -vvv --http3 https://localhost
    *   Trying ::1:443...
    * Connect socket 5 over QUIC to ::1:443
    * Sent QUIC client Initial, ALPN: h3-29,h3-28,h3-27
    *   Trying 127.0.0.1:443...
    * Connect socket 6 over QUIC to 127.0.0.1:443
    * Sent QUIC client Initial, ALPN: h3-29,h3-28,h3-27
    * After 150000ms connect time, move on!
    * connect to ::1 port 443 failed: Connection timed out
    * After 150000ms connect time, move on!
    * connect to 127.0.0.1 port 443 failed: Connection timed out
    * Failed to connect to localhost port 443: Connection timed out
    * Closing connection 0
    curl: (28) Failed to connect to localhost port 443: Connection timed out
    ```
- <details><summary><b>Verbose caddy logs from console with annotations (// 📌):</b></summary><p>

	```json
	// 📌 start caddy
    {"level":"info","ts":1631424276.233273,"msg":"using adjacent Caddyfile"}
    {"level":"warn","ts":1631424276.233972,"msg":"input is not formatted with 'caddy fmt'","adapter":"caddyfile","file":"Caddyfile","line":2}
    {"level":"info","ts":1631424276.2348974,"logger":"admin","msg":"admin endpoint started","address":"tcp/localhost:2019","enforce_origin":false,"origins":["[::1]:2019","127.0.0.1:2019","localhost:2019"]}
    {"level":"info","ts":1631424276.2352147,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc00045f030"}
    {"level":"info","ts":1631424276.235942,"logger":"http","msg":"server is listening only on the HTTPS port but has no TLS connection policies; adding one to enable TLS","server_name":"srv0","https_port":443}
[...]

PR #2727, opened 2019-08-26
Experimental IETF-standard HTTP/3 support

mholt:
## 1. What does this change do, exactly?

Adds IETF HTTP/3 functionality to Caddy.

## 2. Please link to the relevant issues.

Closes #2450

## 3. Which documentation changes (if any) need to be made because of this PR?

Add HTTP/3 to our ad-hoc docs for Caddy 2.

## 4. Checklist

- [ ] I have written tests and verified that they fail without my change
- [x] I have squashed any insignificant commits
- [x] This change has comments explaining package types, values, functions, and non-obvious lines of code
- [x] I am willing to help maintain this change if there are issues with it later

PR #4413, opened 2021-11-12
caddyhttp: Fix HTTP/3 breaking after reload

vladbondarenko:
Fixes https://github.com/caddyserver/caddy/issues/4348

Disabling extra closing http3 listeners.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -432,25 +432,27 @@ func (app *App) Stop() error {
 	// repeated attempts (the bug manifested after a config reload; i.e.
 	// reusing a http3 server or listener was problematic), but it seems
 	// to be working fine now
-	for _, s := range app.h3servers {
-		// TODO: CloseGracefully, once implemented upstream
-		// (see https://github.com/lucas-clemente/quic-go/issues/2103)
-		err := s.Close()
-		if err != nil {
-			return err
-		}
-	}
+	// UPDATE: no need to extra closing servers and listeners as they already closing
+	// in lucas-clemente/quic-go@v0.23.0/http3/server.go L:413
+	//for _, s := range app.h3servers {
+	//	// TODO: CloseGracefully, once implemented upstream
+	//	// (see https://github.com/lucas-clemente/quic-go/issues/2103)
+	//	err := s.Close()
+	//	if err != nil {
+	//		return err
+	//	}
+	//}
 
 	// closing an http3.Server does not close their underlying listeners
 	// since apparently the listener can be used both by servers and
 	// clients at the same time; so we need to manually call Close()
 	// on the underlying h3 listeners (see lucas-clemente/quic-go#2103)
-	for _, pc := range app.h3listeners {
-		err := pc.Close()
-		if err != nil {
-			return err
-		}
-	}
+	//for _, pc := range app.h3listeners {
+	//	err := pc.Close()
+	//	if err != nil {
+	//		return err
+	//	}
+	//}
 	return nil
 }
 


mholt:
Ah, interesting. I don't seem to remember that being the case before, so that's good news. (Or if it was, I just made a dumb. But I swear this used to work, so. Anyway I'm not complaining. The best fixes delete code.)

Thanks for the PR! Will review this shortly.

francislavoie:
@marten-seemann does this make sense to you?

marten-seemann:
I feel like I'm missing context here. What do you mean by "as they already closing in lucas-clemente/quic-go@v0.23.0/http3/server.go L:413"? That would be [`Server.Close`](https://github.com/lucas-clemente/quic-go/blob/v0.23.0/http3/server.go#L413), but someone has to call that function, right?

francislavoie:
@marten-seemann the issue this is meant to solve is https://github.com/caddyserver/caddy/issues/4348

I'll update the PR description/title.

marten-seemann:
How does this reload work? Do you restart the QUIC server after the new config has been applied?

francislavoie:
@marten-seemann during a reload, the old server is kept up, and a new one is created to replace it, and finally the old server will be torn down if the new one started without error. So two servers will be running concurrently for a moment. There's a more thorough explanation of Caddy's architecture here, to bring you back up to speed on this https://caddyserver.com/docs/architecture

marten-seemann:
I'm not sure. Running them concurrently is ok, but you need to close the old server somehow, don't you?

awfulcooking:
Here's more context, fwiw:

#2727 - HTTP3 implemented. [There is trouble](https://github.com/caddyserver/caddy/pull/2727#issuecomment-526856566) with reloading, relating to closing / recreating servers, listeners with quic-go

lucas-clemente/quic-go#2111 - @marten-seemann attempts to fix an issue with quic-go, making its multiplexer recognise conns by their local address string, instead of struct equality.

[#2727 comment from @mholt just before merge (Sept 2019)](https://github.com/caddyserver/caddy/pull/2727#issuecomment-529947251) - "the issue persists, but by not closing the http3.Server, the issue is either resolved or it hides itself."

So it's merged, with h3 listeners being closed. A loop to close h3 servers is **commented out**.

[18 Feb 2020, @mholt restores the loop that closes h3 servers](https://github.com/caddyserver/caddy/commit/7f9cfcc0f2918ec3c01b3f0408442026be454dc2). This diff, surrounding comments probably most salient for context.

[#4348, "http/3 stops working after reload" filed 12 Sep 2021](https://github.com/caddyserver/caddy/issues/4348). Leading to this PR

awfulcooking:
Could it be related to lucas-clemente/quic-go#2111 preventing the new listener from starting?

mholt:
@vladbondarenko Thanks for looking into this. When you do reloads, you should see [some logs](https://github.com/caddyserver/caddy/blob/e81369e2208e47d9650f9699ad8bc7692640b275/listeners.go#L79) like:

```
[DEBUG] ... Usage counter should not go above 2 or maybe 3, is now: N
```

where `N` is a number that is hopefully less than 3. Does that number stay below 3 when you do a few reloads?

IndeedNotJames:
> [18 Feb 2020, @mholt restores the loop that closes h3 servers](https://github.com/caddyserver/caddy/commit/7f9cfcc0f2918ec3c01b3f0408442026be454dc2). This diff, surrounding comments probably most salient for context.

The mentioned commit happened on tag [v2.0.0-beta.15](https://github.com/caddyserver/caddy/releases/tag/v2.0.0-beta.15) and seems to go on par with what I wrote in https://github.com/caddyserver/caddy/issues/4348#issue-994214641:

> I went as far as v2.0.0 back (thus the two Caddyfiles, see further down) and encountered this issue every time


So I went and checked all the `v2.0.0-beta.*` but those seem to work just fine. The same applies to the actual `v2.0.0` release.
But I could swear when I tried `v2.0.0` ([`caddy:2.0.0` Docker Hub](https://hub.docker.com/layers/caddy/library/caddy/2.0.0/images/sha256-2653b633a6ecc87fdf8ab9941be387a42f25f1c17580878c9cd47055a726d734)) before, it failed.
 
Thus, I redid my testing:
- Docker Hub
  - caddy:2.0.0 from Docker Hub seems to work
  - caddy:2.1.1 from Docker Hub seems to work
  - caddy:2.2.0 from Docker Hub **does not** to work
- GitHub Releases (Linux amd64/x86_64 builds)
  - [`v2.2.0-rc.1 h1:ULGBB9efRs8yG27IBOMLljhCa4iM1IluFTDxjm+Y8vE=`](https://github.com/caddyserver/caddy/releases/tag/v2.2.0-rc.1) seems to work
  - [`v2.2.0-rc.2 h1:TIZuB1xz2YEvoFOPSGuxjGAcS4lMP604V+rwL0X4V0Q=`](https://github.com/caddyserver/caddy/releases/tag/v2.2.0-rc.2) **does not** work
- `go run ./main.go run`
  - v2.2.0 build with go1.15.15 **does not** work
  - v2.2.0 built with go1.14.15 seems to work

Commit https://github.com/caddyserver/caddy/commit/0279a57ac465b2920abf71d86203d9feac2015b5 between https://github.com/caddyserver/caddy/compare/v2.2.0-rc.1...v2.2.0-rc.2 updated `quic-go` and `Go` to 1.15 

Could this be, perhaps, somewhat related to the Go version bump from `1.14` to `1.15`?

mholt:
Interesting. Thanks for the detective work. I am not sure what the upgrade from Go 1.14 to Go 1.15 would have done to affect that...

mholt:
@vladbondarenko Can you confirm that the counter doesn't increase?

ChrisLahaye:
Unfortunately this is still an issue for me. Is there anything needed before this can get merged?

mholt:
We need confirmation that it does work and that the counter doesn't increase, indicating a leak.

ChrisLahaye:
I will look into this and get back to you guys.

ChrisLahaye:
I ran some tests and this PR indeed creates a leak. 

It seems like that the `h3servers` (and `h3listeners`) need to be closed explicitly. Without that, the number of listeners keeps increasing. This has been validated by adding a probe to the `ListenPacket` function in listeners.go (see https://github.com/caddyserver/caddy/blob/master/listeners.go#L58):
```
+	pv, loaded := listenerPool.pool[lnKey]
+	if loaded {
+		log.Printf("[DEBUG] %s: loaded %d", lnKey, atomic.LoadInt32(&pv.refs))
+	}
```

I did observe that when using `(*http.Server).Shutdown` the issue does not happen. This function comes from the http.Server embed in http3.Server, so I am not sure how far this is compatible with the http3 resources. Therefore, I will refrain from a PR and just leave it here for discussion.

mholt:
Thanks for verifying that @ChrisLahaye . Sounds like we'll need a way to shut down the HTTP/3 server without closing the socket for others who may be using it.

mholt:
I'm going to close this PR for now as it seems like it may be going inactive, and it has been confirmed that it's not a viable fix. Please feel welcome to continue discussion though if needed, although perhaps that is more relevant in the original issue unless it's about this change specifically. Thanks!

PR #4417, opened 2021-11-13
fileserver: Extract default browse template into a separate file

jjlin:
This makes it easier for users to find the default browse template if they
want to create a custom template based on that. It also makes it easier to
view the template with proper syntax highlighting.


jjlin:
Note that this PR uses the file embedding capability introduced in Go 1.16, but Caddy currently requires at least that version to build, so that should be fine.

modules/caddyhttp/fileserver/browsetpl.go
```diff
@@ -14,467 +14,7 @@
...
-				}
-				e.textContent = d.toLocaleString([], {day: "2-digit", month: "2-digit", year: "numeric", hour: "2-digit", minute: "2-digit", second: "2-digit"});
-			}
-			var timeList = Array.prototype.slice.call(document.getElementsByTagName("time"));
-			timeList.forEach(localizeDatetime);
-		</script>
-	</body>
-</html>`
+//go:embed browsetpl.html
+var defaultBrowseTemplate string
```
francislavoie, line 20: Could we just move this line to `browse.go`? This file seems redundant now.

francislavoie (commented):
Thanks for this, I was thinking of doing this change at some point, but you got to it first 😄 

Unfortunate that we lose git history from moving the contents to another file, but we gain from syntax highlighting etc.

We'll also need to update the docs once merged to link to the `.html` file instead. See https://caddyserver.com/docs/caddyfile/directives/file_server

modules/caddyhttp/fileserver/browsetpl.go
jjlin: Good idea, I updated the PR.

jjlin:
> We'll also need to update the docs once merged to link to the `.html` file instead. See https://caddyserver.com/docs/caddyfile/directives/file_server

Thanks, I didn't know there was a reference to the template. I opened caddyserver/website#196.

mholt (approved):
Thank you for this. It'll be our first time using go:embed features, so we'll give it a shot!

PR #4419, opened 2021-11-14
headers: use ReplaceKnown instead of ReplaceAll to avoid ignoring JSON values

### LINKED ISSUE #4418: headers: JSON strings deleted when setting response headers
With the following `Caddyfile`, I expect there to be a response header with the value `Foo: {"foo": "bar"}`:

```
localhost

header Foo "{\"foo\":\"bar\"}"
respond "hello world"
```

But instead, I get:

```console
$ curl -I https://localhost
HTTP/2 200 
foo: 
server: Caddy
content-length: 11
date: Sun, 14 Nov 2021 15:18:34 GMT
```

It seems that any text between the `{}` is being lost, for example, `header Foo "hello{}world"` results in a response header of `Foo: helloworld`.

hairyhenderson:
Fixes #4418

`repl.ReplaceAll` silently ignores anything between unescaped `{}`s, which is somewhat unexpected behaviour.

Replacing this usage with `repl.ReplaceKnown` fixes this, however I wonder if perhaps it would be better to use `ReplaceOrErr`, since that way we can encourage properly escaping values that aren't actually placeholders... The challenge is that `ApplyTo` doesn't return an `error` so the best we could do (without changing the signature) is log the error and allow the header to either be ignored, or the value to be set to an empty string - neither of these are desirable IMO...

### DIFF
--- a/modules/caddyhttp/headers/headers.go
+++ b/modules/caddyhttp/headers/headers.go
@@ -188,38 +188,38 @@ type RespHeaderOps struct {
 func (ops HeaderOps) ApplyTo(hdr http.Header, repl *caddy.Replacer) {
 	// add
 	for fieldName, vals := range ops.Add {
-		fieldName = repl.ReplaceAll(fieldName, "")
+		fieldName = repl.ReplaceKnown(fieldName, "")
 		for _, v := range vals {
-			hdr.Add(fieldName, repl.ReplaceAll(v, ""))
+			hdr.Add(fieldName, repl.ReplaceKnown(v, ""))
 		}
 	}
 
 	// set
 	for fieldName, vals := range ops.Set {
-		fieldName = repl.ReplaceAll(fieldName, "")
+		fieldName = repl.ReplaceKnown(fieldName, "")
 		var newVals []string
 		for i := range vals {
 			// append to new slice so we don't overwrite
 			// the original values in ops.Set
-			newVals = append(newVals, repl.ReplaceAll(vals[i], ""))
+			newVals = append(newVals, repl.ReplaceKnown(vals[i], ""))
 		}
 		hdr.Set(fieldName, strings.Join(newVals, ","))
 	}
 
 	// delete
 	for _, fieldName := range ops.Delete {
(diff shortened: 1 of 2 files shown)


francislavoie (approved):
I think using ReplaceKnown is a good-enough fix. Thanks!

mholt (approved):
Ah yeah, so when I originally wrote this, I wasn't sure which would be best here. I had figured that if placeholders were being used in headers, they might not always be set (because some placeholders aren't at some times), and the user would want an empty value instead of an un-eval'ed placeholder literally in their header fields. But I could also see the case for `ReplaceKnown()`, so I debated that and ultimately went with `ReplaceAll()`, since curly braces can be escaped with `\{ }` and this way you won't end up with a raw placeholder unintentionally in your headers. ReplaceAll is more flexible in that regard: it gives more flexibility to the user.

I'm indifferent on this personally, so I'll approve this change. Feel free to merge it if you want it (i.e. if escaping is too tedious), but be aware of the trade off here if you do merge it.

hairyhenderson:
Hmm... I hadn't considered that there may be cases where a user is _expecting_ a placeholder to resolve to an empty string in particular cases.

Technically that means this would be a breaking change, which I'm less keen on.

I think a different (and safer, non-breaking) approach could be to simply log a message (maybe even at debug level?) if the `ReplaceAll` call evaluates to an empty string, just to provide a hint to folks like me that it _might_ not have been what I was intending...

mholt:
> Hmm... I hadn't considered that there may be cases where a user is expecting a placeholder to resolve to an empty string in particular cases.

To be fair, I don't know of an actual use case where this is needed. Just was what crossed my mind when I wrote it, as the current approach allows both (empty replacements, and using curly braces with escaping) whereas the proposed change only allows one (using curly braces, but no escaping required).

Honestly I'm indifferent about it, so go ahead and I'll let the community choose on this one. 🤷

hairyhenderson:
Ok, I'm going to take the easy path and close this. If someone else runs into this issue maybe they'll find this and realize it's just a case of escaping the braces 😉

PR #4424, opened 2021-11-17
logging: add a filter for query parameters

dunglas:
Even if it's considered a bad security practice, it's common that query parameters contain sensitive information. Some examples:

* [OAuth access token](https://datatracker.ietf.org/doc/html/rfc6750#section-2.3)
* "magic links" and other temporary URLs
* [PHP session IDs](https://www.php.net/manual/en/session.idpassing.php)
* [Mercure JWT](https://github.com/dunglas/mercure/pull/562) (the reason why I started working on this patch in the first time)

This PR adds the possibility to redact such sensitive information from the logs.

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -262,8 +262,8 @@ func (m QueryFilter) Filter(in zapcore.Field) zapcore.Field {
 	for _, a := range m.Actions {
 		switch a.Type {
 		case ReplaceType:
-			if _, ok := q[a.Parameter]; ok {
-				q.Set(a.Parameter, a.Value)
+			for i := range q[a.Parameter] {
+				q[a.Parameter][i] = a.Value
 			}
 
 		case DeleteType:
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -188,7 +188,7 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
-type ActionType string
+type FilterAction string
 
 const (
 	// Replace value(s) of query parameter(s).
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -16,6 +16,7 @@ package logging
 
 import (
 	"net"
+	"net/url"
 	"strconv"
 
 	"github.com/caddyserver/caddy/v2"
@@ -27,6 +28,7 @@ func init() {
 	caddy.RegisterModule(DeleteFilter{})
 	caddy.RegisterModule(ReplaceFilter{})
 	caddy.RegisterModule(IPMaskFilter{})
+	caddy.RegisterModule(QueryFilter{})
 }
 
 // LogFieldFilter can filter (or manipulate)
@@ -185,15 +187,107 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
+type ActionType string
+
+const (
+	ReplaceType ActionType = "replace"
+	DeleteType  ActionType = "delete"
+)
+
+type queryFilterAction struct {
+	Type      ActionType `json:"type"`
+	Parameter string     `json:"parameter"`
+	Value     string     `json:"value,omitempty"`
+}
+
+// QueryFilter is a Caddy log field filter that
+// filters query parameters.
+type QueryFilter struct {
+	Actions []queryFilterAction `json:"actions"`
+}
+
+// CaddyModule returns the Caddy module information.
+func (QueryFilter) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "caddy.logging.encoders.filter.query",
+		New: func() caddy.Module { return new(QueryFilter) },
+	}
+}
+
+// UnmarshalCaddyfile sets up the module from Caddyfile tokens.
+func (m *QueryFilter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	for d.Next() {
+		for d.NextBlock(0) {
+			qfa := queryFilterAction{}
+			switch d.Val() {
+			case "replace":
+				if !d.NextArg() {
+					return d.ArgErr()
+				}
+
+				qfa.Type = ReplaceType
+				qfa.Parameter = d.Val()
+
+				if !d.NextArg() {
+					return d.ArgErr()
+				}
+				qfa.Value = d.Val()
+
+			case "delete":
+				if !d.NextArg() {
+					return d.ArgErr()
+				}
+
+				qfa.Type = DeleteType
+				qfa.Parameter = d.Val()
+
+			default:
+				return d.Errf("unrecognized subdirective %s", d.Val())
+			}
+
+			m.Actions = append(m.Actions, qfa)
+		}
+	}
+	return nil
+}
+
+// Filter filters the input field.
+func (m QueryFilter) Filter(in zapcore.Field) zapcore.Field {
+	u, err := url.Parse(in.String)
+	if err != nil {
+		return in
+	}
+
+	q := u.Query()
+	for _, a := range m.Actions {
+		switch a.Type {
+		case ReplaceType:
+			if q.Has(a.Parameter) {
+				q.Set(a.Parameter, a.Value)
+			}
+
+		case DeleteType:
+			q.Del(a.Parameter)
+		}
+	}
+
+	u.RawQuery = q.Encode()
+	in.String = u.String()
+
+	return in
+}
+
 // Interface guards
 var (
 	_ LogFieldFilter = (*DeleteFilter)(nil)
 	_ LogFieldFilter = (*ReplaceFilter)(nil)
 	_ LogFieldFilter = (*IPMaskFilter)(nil)
+	_ LogFieldFilter = (*QueryFilter)(nil)
 
 	_ caddyfile.Unmarshaler = (*DeleteFilter)(nil)
 	_ caddyfile.Unmarshaler = (*ReplaceFilter)(nil)
 	_ caddyfile.Unmarshaler = (*IPMaskFilter)(nil)
+	_ caddyfile.Unmarshaler = (*QueryFilter)(nil)
 
 	_ caddy.Provisioner = (*IPMaskFilter)(nil)
 )
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -188,18 +188,18 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
-type FilterAction string
+type filterAction string
 
 const (
 	// Replace value(s) of query parameter(s).
-	ReplaceAction FilterAction = "replace"
+	replaceAction filterAction = "replace"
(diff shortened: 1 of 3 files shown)


francislavoie (commented):
LGTM! Just a couple comments

modules/logging/filters_test.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func TestQueryFilter(t *testing.T) {
+	f := QueryFilter{[]queryFilterAction{
+		{ReplaceType, "foo", "REDACTED"},
+		{DeleteType, "bar", ""},
+	}}
+
+	out := f.Filter(zapcore.Field{String: "/path?foo=a&foo=b&bar=c&bar=d&baz=e"})
+	if out.String != "/path?baz=e&foo=REDACTED" {
```
wiese, line 16: To me it's not really intuitive (despite the obvious goal to redact the values) that two `foo` keys go in (values `a` and `b`) but only one `foo` key is expected to come out. Is this somehow rooted in the way caddy in general treats queries and logs them?

wiese (commented):
Cha­peau bas for the idea and initiative. I just saw you mention the PR on twitter, am not really involved with caddy – feel free to ignore my comment.

Maybe that's taking the second step before the first, but my first reaction when seeing the "REDACTED" value was thinking: this is going to be confusing sooner or later. Maybe the better way would be to anonymize/pseudonymize the data (i.e. not having the same log value for all incoming values). Does that resonate with you? Would you see this as an extra feature or an alternative, indeed?

modules/logging/filters.go
```diff
@@ -185,15 +187,107 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
+type ActionType string
+
+const (
+	ReplaceType ActionType = "replace"
+	DeleteType  ActionType = "delete"
+)
```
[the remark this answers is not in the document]
dunglas: > What happens if a user tries a different action via JSON config?

Go will use the invalid data without throwing an error. Maybe could I use the `Validate()` method to check the types?

dunglas:
> Maybe the better way would be to anonymize/pseudonymize the data

@wiese What would be the use case? Other similar software seems to redact and not anonymize:

* [Traefik](https://doc.traefik.io/traefik/v1.6/configuration/logs/)
* [Axway](https://docs.axway.com/bundle/axway-open-docs/page/docs/apim_administration/apigtw_admin/admin_redactors/index.html)
* [Kong](https://docs.konghq.com/gateway/2.6.x/pdk/kong.log/#konglogserialize)

Also, it's how the current filters (replace and delete) work.
Maybe could we add more advanced anonymization/hashing features later if there is a use case?

modules/logging/filters_test.go
dunglas: It's rooted in the Go stdlib but maybe can we preserve the same number of keys, I'll take a look. As it's on the hot path, we should be careful to not consume too much resources.

dunglas: Actually, it was easy to handle. Done, thanks @wiese!

wiese:
> What would be the use case? Other similar software seems to redact and not to anonymize

The use case, like for the current implementation, would be making sure the original values are not contained in the logs but (for _pseudonymization_) with the added benefit that
* different queries don't look alike, so as to avoid that wrong conclusions are drawn from that (e.g. two queries looking the same in the logs even when they were not, or spotting an illusive hot spot when grouping them)
* and vice versa, in queries which originally were the same, one can be found based on another and they can be grouped (think: by user JWT)

Of course, this functionality could totally exist _in addition_ to redacting as is and added later if there is a demand. No worries, I just wanted to run the thought in case you say "oh, that's actually closer to what I want".

francislavoie:
We could do a thing where if the replacement value is the string `HASHED` then it would SHA-256 hash the existing value (maybe truncated to 8 chars? I dunno) to deal with anonymizing without making different requests look alike. 

Obviously it should be noted that this isn't a good idea for passwords for example, i.e. anything where brute-forcing the value would be problematic. In which case, simply redacting is better.

dunglas:
Or maybe could we just add an extra action called `hash` in addition to `replace` and `delete`? It would be very easy to implement.

dunglas:
Maybe could we add this feature in a follow-up PR? (I will do it, but it will be easier to add hash support everywhere when the pending PRs will be merged).

modules/logging/filters.go
```diff
@@ -185,15 +187,107 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+
+type queryFilterAction struct {
+	Type      ActionType `json:"type"`
+	Parameter string     `json:"parameter"`
+	Value     string     `json:"value,omitempty"`
+}
+
+// QueryFilter is a Caddy log field filter that
+// filters query parameters.
```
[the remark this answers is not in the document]
dunglas: Is it ok like this or do you have something else in mind?

mholt (changes requested):
Hey Kevin, thanks for the contribution! I like where this is going.

Just wondering... could both `replace` and `delete` actions be reduced to a single `replace` action? Since deleting could be replacing with an empty string, and you wouldn't need different action types and stuff. It'd be simpler. You could then just list the replacements you want to perform on the query string.

modules/logging/filters.go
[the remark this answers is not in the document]
francislavoie: Yeah, I think we should `Validate()` to make sure invalid actions aren't configured.

FWIW I played with this idea of using an enum in another PR I wrote after having seen how you did it here. It's possible to add methods to `ActionType` to make it easy to validate. https://github.com/mholt/caddy-dynamicdns/pull/16/files#diff-3d56da1a5c865a0689b48c66b4b24ff2441ea4c442c2bc9dd074814034f78fc8R318

modules/logging/filters.go
```diff
@@ -185,15 +187,117 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+	DeleteType ActionType = "delete"
+)
+
+type queryFilterAction struct {
+	// `replace` to replace the value(s) associated with the parameter(s) or `delete` to remove them entirely.
+	Type ActionType `json:"type"`
+	// The name of the query parameter.
+	Parameter string `json:"parameter"`
+	// The value to use as replacement if the action is `replace`.
+	Value string `json:"value,omitempty"`
```
francislavoie, line 205: Style thing, we generally use a newline between fields when there's comments, to space it out a bit

```suggestion
	// `replace` to replace the value(s) associated with the parameter(s) or `delete` to remove them entirely.
	Type ActionType `json:"type"`

	// The name of the query parameter.
	Parameter string `json:"parameter"`

	// The value to use as replacement if the action is `replace`.
	Value string `json:"value,omitempty"`
```

modules/logging/filters.go
```diff
@@ -185,15 +187,117 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+}
+
+// QueryFilter is a Caddy log field filter that
+// filters query parameters from an URL.
+// It updates the filtered URL to remove or replace query parameters containing sensitive data.
+// For instance, it can be used to redact OAuth access tokens, session IDs, magic links tokens
+// or JWT when passed as query parameters (which is generally considered a bad practice, but cannot
+// always be avoided).
```
francislavoie, line 213: Since this will get rendered as Markdown, having a newline between will actually turn them into two paragraphs.

I shortened the line length a bit too, so the comment has a nicer shape 😅 and a few minor changes to wording. I removed the JWT bit because it's a bit of an unnecessary "aside".

```suggestion
// QueryFilter is a Caddy log field filter that filters
// query parameters from a URL.
//
// This filter updates the logged URL string to remove or replace query
// parameters containing sensitive data. For instance, it can be used
// to redact any kind of secrets which were passed as query parameters,
// such as OAuth access tokens, session IDs, magic link tokens, etc.
```

modules/logging/filters.go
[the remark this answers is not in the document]
francislavoie: Sorry - I missed the comment, got lost in my notifications.

Yeah, that's better! I made a comment to adjust wording separately (this review comment is now outdated)

dunglas:
Hi Matt,

This is possible but it could get messy if we add more actions such as `hash`. On the other hand, we could use @francislavoie's suggestion of using special strings.

Currently:

```
	format filter {
		fields {
			uri query {
				replace foo1 REDACTED
                                replace foo2 FOO2
				delete bar
				hash baz
			}
               }
```

With your suggestion:

```
	format filter {
		fields {
                    uri query foo1 REDACTED foo2 FOO bar `` baz HASH
                }
```

I find the first one more readable but I'm ok to update the PR to use the second option if you prefer.

Also, in we implement your solution, maybe should we also deprecate the existing `delete` filter for consistency (using an empty string as a parameter of the `replace` directive also works).

WDYT?

mholt:
Thanks for the extra information. That's very interesting! As discussed in Slack, I think the current approach is fine, since it allows other actions to be configured like hashing. It's also easier to read. Thanks!

modules/logging/filters.go
```diff
@@ -185,15 +187,120 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
+type ActionType string
```
mholt, line 190: Does this need to be a separate type? I think we could probably just use string, since this type doesn't have any special properties or behaviors. (And I suspect it definitely doesn't need to be exported.)

mholt (commented):
Just one thing I noticed. Looking good though.

modules/logging/filters.go
francislavoie: The idea was to add `IsValid()` on this, see my review above https://github.com/caddyserver/caddy/pull/4424#discussion_r753632173

francislavoie (approved):
Thanks!

modules/logging/filters.go
```diff
@@ -185,15 +188,142 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+
+const (
+	// Replace value(s) of query parameter(s).
+	ReplaceType ActionType = "replace"
+	// Delete query parameter(s).
+	DeleteType ActionType = "delete"
```
mholt, line 197:
```suggestion
	ReplaceAction ActionType = "replace"
	// Delete query parameter(s).
	DeleteAction ActionType = "delete"
```

modules/logging/filters.go
```diff
@@ -185,15 +188,142 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
+type ActionType string
```
mholt, line 191:
```suggestion
type FilterAction string
```

Think this might make more sense in a package context, as in, `logging.FilterAction` is more helpful than `logging.ActionType`.

Or we could unexport it, if it doesn't need to be exported that would be even better.

mholt (changes requested):
Thank you! LGTM except a little nit on naming / exporting one identifier.

francislavoie (approved):
Thanks!

PR #4425, opened 2021-11-17
logging: add a filter for cookies

dunglas:
Similar to https://github.com/caddyserver/caddy/pull/4424 but for cookies, we often (and legitimately) contain sensitive data that must not be logged.

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -17,6 +17,7 @@ package logging
 import (
 	"errors"
 	"net"
+	"net/http"
 	"net/url"
 	"strconv"
 
@@ -30,6 +31,7 @@ func init() {
 	caddy.RegisterModule(ReplaceFilter{})
 	caddy.RegisterModule(IPMaskFilter{})
 	caddy.RegisterModule(QueryFilter{})
+	caddy.RegisterModule(CookieFilter{})
 }
 
 // LogFieldFilter can filter (or manipulate)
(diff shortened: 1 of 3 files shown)


modules/logging/filters.go
```diff
@@ -185,15 +187,113 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
+type ActionType string
+
+const (
+	ReplaceType ActionType = "replace"
+	DeleteType  ActionType = "delete"
+)
```
francislavoie, line 195: Seems like this is code in common with https://github.com/caddyserver/caddy/pull/4424 so I suppose whichever is merged first will need to be rebased to adjust afterwards.

modules/logging/filters.go
```diff
@@ -185,15 +187,113 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+
+type cookieFilterAction struct {
+	Type  ActionType `json:"type"`
+	Name  string     `json:"name"`
+	Value string     `json:"value,omitempty"`
+}
+
+// CookiesFilter is a Caddy log field filter that
+// filters cookies.
```
francislavoie, line 204: Same as the other PR, could use a usage explanation for JSON users, at least mention the action types.

modules/logging/filters.go
```diff
@@ -185,15 +187,113 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+			}
+
+			switch a.Type {
+			case ReplaceType:
+				c.Value = a.Value
+				transformedRequest.AddCookie(c)
+				continue OUTER
+
+			case DeleteType:
+				continue OUTER
```
francislavoie, line 274: Might be worth mentioning in the godoc that if two "actions" are configured for the same cookie name, only the first will apply (because the `continue OUTER` will skip the others), although it's pretty obvious only one should be configured per.

Could be useful later to apply more than one, but that'll depend on whether any new actions added would make that beneficial.

modules/logging/filters.go
```diff
@@ -185,15 +187,113 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+
+const (
+	ReplaceType ActionType = "replace"
+	DeleteType  ActionType = "delete"
+)
+
+type cookieFilterAction struct {
+	Type  ActionType `json:"type"`
+	Name  string     `json:"name"`
+	Value string     `json:"value,omitempty"`
```
francislavoie, line 200: Could use godoc comments on these fields as well (they get pulled into the JSON docs)

modules/logging/filters.go
dunglas: indeed

francislavoie (approved):

mholt (approved):
Thanks for the contribution, will be happy to accept it!

PR #4429, opened 2021-11-20
caddyhttp: Log non-500 handler errors at debug level

### LINKED ISSUE #4428: caddyhttp.HandlerErr non-500 status logging
Hello! Is there a reason why Caddy only logs handler errors that have 500-level status codes (see https://github.com/caddyserver/caddy/blob/124ba1ba714220322c65625012c564f89b75bfc9/modules/caddyhttp/server.go#L263-L265)? I've run into [this failure case](https://github.com/caddyserver/caddy/blob/cbb045a121464527d85cce1b56250480b0515f9a/modules/caddyhttp/server.go#L291-L296) before and it wasn't obvious because that error doesn't actually end up being logged because it's a 403. Are there non-500 handler errors that we don't want logged? Otherwise, that seems like it'd be useful information for debugging.

francislavoie:
Fixes #4428

It's best to still log handler errors at debug level so that they're hidden by default, but still accessible if additional details are necessary.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -244,6 +244,8 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			// successfully, so now just log the error
 			if errStatus >= 500 {
 				logger.Error(errMsg, errFields...)
+			} else {
+				logger.Debug(errMsg, errFields...)
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


dunglas (approved):

mholt (approved):
Hm, I guess that's a good idea. Thanks for the patch!

PR #4433, opened 2021-11-23
caddyhttp: Enhance vars matcher

mholt:
Enable "or" logic for multiple values. Fall back to checking placeholders if not a var name.

Seemed useful, idk. :man_shrugging: 

This matcher is actually undocumented so far, since I wasn't sure if it'd have any purpose. But I think there might be value in matching on placeholders and variables for more advanced configurations.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/matcher_syntax.txt
+++ b/caddytest/integration/caddyfile_adapt/matcher_syntax.txt
@@ -101,7 +101,9 @@
 							"match": [
 								{
 									"vars": {
-										"{http.request.uri}": "/vars-matcher"
+										"{http.request.uri}": [
+											"/vars-matcher"
+										]
 									}
 								}
 							],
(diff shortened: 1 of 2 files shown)


mohammed90:
I like making it more powerful. Are we concerned about breaking users? Unfortunately we don't know who's using it. It's documented in JSON.

This diff fixes the tests:
```diff
--- a/caddytest/integration/caddyfile_adapt/matcher_syntax.txt
+++ b/caddytest/integration/caddyfile_adapt/matcher_syntax.txt
@@ -101,7 +101,9 @@
                                                        "match": [
                                                                {
                                                                        "vars": {
-                                                                               "{http.request.uri}": "/vars-matcher"
+                                                                               "{http.request.uri}": [
+                                                                                       "/vars-matcher"
+                                                                               ]
                                                                        }
                                                                }
                                                        ],
```

mholt:
True, but I'm pretty sure I've never seen or heard of a config where the vars matcher is being used directly in JSON config (only the Caddyfile adapter).

And thanks for the diff, I need to apply that.

PR #4426, opened 2021-11-17
feat(logging): add a regexp filter

dunglas:
Adds a more generic filter than https://github.com/caddyserver/caddy/pull/4424 and https://github.com/caddyserver/caddy/pull/4425 to filter any field using a regexp (this cannot replace the specific query and cookie filters because URIs and cookies cannot be safely parsed using a simple regexp).

PR #4434, opened 2021-11-23
logging: add support for hashing data

dunglas:
Follows #4424, #4425 and #4426. Implements a hash filter and hash actions for the query and cookie filters ask suggested by @wiese in https://github.com/caddyserver/caddy/pull/4424#issuecomment-972650688.

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -15,7 +15,9 @@
 package logging
 
 import (
+	"crypto/sha256"
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
 	"net/url"
@@ -34,6 +36,7 @@ func init() {
 	caddy.RegisterModule(QueryFilter{})
 	caddy.RegisterModule(CookieFilter{})
 	caddy.RegisterModule(RegexpFilter{})
+	caddy.RegisterModule(HashFilter{})
 }
 
 // LogFieldFilter can filter (or manipulate)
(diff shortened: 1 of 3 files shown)


modules/logging/filters.go
```diff
@@ -65,6 +68,35 @@ func (DeleteFilter) Filter(in zapcore.Field) zapcore.Field {
 	return in
 }
 
+// hash returns the first 4 bytes of the SHA-256 hash of the given data as hexadecimal
+func hash(s string) string {
+	return fmt.Sprintf("%.4x", sha256.Sum256([]byte(s)))
```
wiese, line 73: How did you decide on 4 bytes? It's shorter than what I would think it takes for inevitable collisions not to become a frequent occurrence – but I don't have data to recommend an alternative minimum.
Anecdotal: [nginx uses 16 bytes for request ids](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_id) (but those are not based on input, of course).

wiese (commented):
Thank you so much for taking this on, @dunglas!

I can imagine the hash filter become more configurable (hashing algorithm, digest length) in the future, similar to what we see in the replace filter (which has the new value passed to it), for which this is a great basis. This would then also render obsolete any (subjective) concerns about the concrete values used.

dunglas:
I thought about that but I was wondering if it's really necessary to introduce (and maintain) such features. I fail to see a use case.

francislavoie:
I agree, I don't think we need to make the hashing configurable. 4 bytes is 8 hex characters, which is _plenty_. Git gets away with less (usually 7 chars) most of the time. `16 ^ 8 = 4.2 billion`. If someone later comes up with a real usecase, we can add it without too much trouble. But I don't think we should implement something early without a good reason.

modules/logging/filters.go
dunglas: See @francislavoie's comment: https://github.com/caddyserver/caddy/pull/4434#issuecomment-978012125 (I indeed copied Git's behavior).

Also, I don't think that we care much about collisions here as the main use case is redacting data but being able to detect different patterns when analyzing the logs (being 100% reliable is not really useful for this use case). Also, a short hash allows reducing the size of the logs on disk.

modules/logging/filters.go
```diff
@@ -195,23 +227,27 @@ func (m IPMaskFilter) Filter(in zapcore.Field) zapcore.Field {
...
+	case replaceAction, deleteAction, hashAction:
 		return nil
 	}
 
 	return errors.New("invalid action type")
 }
 
 type queryFilterAction struct {
-	// `replace` to replace the value(s) associated with the parameter(s) or `delete` to remove them entirely.
+	// `replace` to replace the value(s) associated with the parameter(s), `hash` to replace them with the 4 initial bytes of the SHA-156 of their content or `delete` to remove them entirely.
```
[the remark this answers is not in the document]
dunglas: Thanks! I think it is worth it documenting the implementation details in case someone wants to emit compatible logs from another software.

francislavoie (approved):

mholt (approved):
Thanks Kevin! This looks good to me.

Issue #4245, opened 2021-07-19
Proxy load balancing retry issues


francislavoie:
There's a few different points to talk about here.

---

Firstly -- I've been playing around a bit with `lb_try_duration` and I think this doesn't necessarily always have the desired effect because of when the "duration" starts. The timer for `lb_try_duration` starts right when the [`reverse_proxy` module gets the request](https://github.com/caddyserver/caddy/blob/124ba1ba714220322c65625012c564f89b75bfc9/modules/caddyhttp/reverseproxy/reverseproxy.go#L399).

This sounds fine in isolation, but if you consider that errors that might trigger retries might take a while to actually happen (e.g. dial timeout after Caddy's default of 10s -- more on that later) then it becomes hard to configure Caddy to only retry once or twice after a failure. It's probably undesirable to set an `lb_try_duration` of 15 seconds just to make sure the 10 second timeouts are caught and trigger retries.

For example, try a config like this:

```
{
	debug
}

:7000 {
	reverse_proxy 10.0.0.255:80 :7001 {
		lb_policy first
		lb_try_duration 3s
		fail_duration 30s
	}
}

:7001 {
	respond "7001"
}
```

Assuming `10.0.0.255:80` does not exist, this should trigger a dial timeout (default 10s). Trying it with `curl localhost:7000`, you see it hang for 10 seconds, then in the logs we get `ERROR dial tcp 10.0.0.255:80: i/o timeout`, and no fallback to `:7001` happens. If we increase `lb_try_duration` to `15s` instead, then we get a hang for 10 seconds then a successful fallback to `:7001` _does_ happen, and no error appear in the logs (but a DEBUG level log does appear for the initial error).

There's a few different ways I can think of that this could be made better:

- Make the try duration timer start from the time the first attempt fails, instead of the beginning of the request.

  This would mean that the amount of times a retry would happen would be easier to predict. This has some implications though, because not all errors take the same amount of time. If some errors happen fast, you don't want to cause a stampede of stacked retries when you have a busy server, so it would be better to be able to cap the amount of retries (because potentially, a 15s try duration with a 250ms interval has a potential of 15s/0.25s = 60 retries)

- Implement a `lb_try_min_times <amount>` option which can complement or replace `lb_try_duration`; a minimum number of times to try to pick an upstream (default 1). If specified alongside `lb_try_duration`, then it should try again even if the duration has elapsed.

[...]

PR #4436, opened 2021-11-24
reverseproxy: Adjust defaults, document defaults

francislavoie:
Related to some of the issues in https://github.com/caddyserver/caddy/issues/4245, a complaint about the proxy transport defaults not being properly documented in https://caddy.community/t/default-values-for-directives/14254/6.

- Dug into the stdlib to find the actual defaults for some of the timeouts and buffer limits, documenting them in godoc so the JSON docs get them next release.

- Moved the keep-alive and dial-timeout defaults from `reverseproxy.go` to `httptransport.go`. It doesn't make sense to set defaults in the proxy, because then any time the transport is configured with non-defaults, the keep-alive and dial-timeout defaults are lost!

- Sped up the dial timeout from 10s to 3s, in practice it rarely makes sense to wait a whole 10s for dialing. A shorter timeout helps a lot with the load balancer retries, so using something lower helps with user experience.

- Also set a dial timeout for the `fastcgi` transport, it didn't have one at all. Might as well.

- Made the keep-alive interval (`ProbeInterval`) configurable via Caddyfile. This was an oversight, I guess.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_options.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_options.txt
@@ -22,6 +22,7 @@ https://example.com {
 			compression off
 			max_conns_per_host 5
 			keepalive_idle_conns_per_host 2
+			keepalive_interval 30s
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


mholt (approved):
These are all great changes. I had been wanting to get around to taking better care of defaults, so thank you!

PR #4437, opened 2021-11-24
fastcgi: Fix a TODO, prevent zap using reflection for logging env

francislavoie:
Fixes a TODO in code, should be reasonably self-explanatory. Using `zap.Any()` would fallback to using reflection, which is slower and can allocate. I doubt this would have any kind of real-world performance impact, but 🤷‍♂️ easy fix.


modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
```diff
@@ -344,6 +345,16 @@ func (t Transport) splitPos(path string) int {
 	return -1
 }
 
+// envVars is a simple type alias to allow for speeding up zap log encoding.
```
mohammed90, line 348: It's a new type but not an alias. [Aliases](https://go.dev/ref/spec#Alias_declarations) are something different.
```suggestion
// envVars is a simple type to allow for speeding up zap log encoding.
```

mholt (approved):
Nice, thanks very much.

PR #4438, opened 2021-11-24
caddyhttp: Make logging of sensitive headers opt-in

francislavoie:
Followup to https://github.com/caddyserver/caddy/commit/7d5047c1f190421528695e1cc3a4ad71c97eb022, as discussed off-Github.

Adds an opt-in flag to `http.servers.logs` which informs the Loggables about whether they should log sensitive headers (default omitted).

Also puts the flag in the request context so that it can bubble down into request handlers which may also log headers (`reverseproxy` and `push` both log their headers).

I'm finding the name of the option kiiinda verbose, but it's a balancing act. 

/cc @dunglas

### DIFF
--- a/caddyconfig/httpcaddyfile/serveroptions.go
+++ b/caddyconfig/httpcaddyfile/serveroptions.go
@@ -33,15 +33,16 @@ type serverOptions struct {
 	ListenerAddress string
 
 	// These will all map 1:1 to the caddyhttp.Server struct
-	ListenerWrappersRaw []json.RawMessage
-	ReadTimeout         caddy.Duration
-	ReadHeaderTimeout   caddy.Duration
-	WriteTimeout        caddy.Duration
-	IdleTimeout         caddy.Duration
-	MaxHeaderBytes      int
-	AllowH2C            bool
-	ExperimentalHTTP3   bool
-	StrictSNIHost       *bool
+	ListenerWrappersRaw  []json.RawMessage
+	ReadTimeout          caddy.Duration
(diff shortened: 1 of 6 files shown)


dunglas (commented):
The implementation looks good to me. Thanks for working on this feature!

I wonder if we couldn't find a name without "headers" in it. This could also allow plugins such as Mercure or Auth Portal to automatically redact sensitive query parameters such as OAuth access tokens and JWT from URLs.

May I suggest `log_credentials` (and it also make this option consistent with https://fetch.spec.whatwg.org/#credentials)?

francislavoie:
Yeah, that can work. I'm worried that might be _too_ vague though? Because we're definitely not taking care to redact everything that might be sensitive, just a select few items. But it's definitely less characters and I like the idea of generalizing for other things to use too.

dunglas:
As we redact all credentials as defined in the Fetch specification (except some safe data regarding client certificates as discussed off GitHub), I think that it's clear enough. Actually we should even mention this spec in the docs so user can find a formal definition of what we remove.

francislavoie:
Alright, good point. Let's go with that.

francislavoie:
Updated, @dunglas. I'm not sure if I'm satisfied with the godoc comments though, wording is awkward, and it's too late for me to come up with polished language 😂

dunglas (approved):

mholt (approved):
Thank you for implementing this, Francis. Saved me a lot of time. :sweat_smile:

Issue #4058, opened 2021-03-11
Rework `caddy trust` to use the admin API for fetching the root cert


tgelite:
I’m trying to script setup and configuration of caddy server based on a custom download that includes additional plugins (caddy-auth-portal, caddy-auth-jwt, caddy-trace, and various caddy-dns modules ).

During setup, the caddy unit file is configured to run caddy as a non priveledged user (by design).  

To get certificates configured properly we are attempting to use the caddy trust command as root during install/config. I think this is performing the trust install based on the default caddy CA location created upon install, rather than the custom storage file_system root declaration I’m passing in my /etc/caddy/Caddyfile (provided below)

The result is on first startup from systemctl is throwing errors indicating it can not import the root ca because the service is not running as a priveledged user.

What would be nice is to be able to run caddy trust as root, indicating the custom location for all local CA certificate files.

Here is an example of the preamble to the Caddyfile configuration I use to expose the certificate deployment for shared use by our combined components on the processing node.   In this scenario letsEncrypt/ZeroTLS is not being used...

{
        storage file_system {
                root /opt/caddy/storage
        }
        local_certs
        http_port 80
        https_port 443
}

The rest of our deployment detail is in forum issue https://caddy.community/t/new-feature-caddy-trust-with-support-for-config-switch/11606

tgelite:
systemd file (modified caddy default) 
```
# caddy.service
#
# For using Caddy with a config file.
#
# Make sure the ExecStart and ExecReload commands are correct
# for your installation.
#
# See https://caddyserver.com/docs/install for instructions.
#
# WARNING: This service does not use the --resume flag, so if you
# use the API to make changes, they will be overwritten by the
# Caddyfile next time the service is restarted. If you intend to
# use Caddy's API to configure it, add the --resume flag to the
# caddy run command or use the caddy-api.service file instead.

[Unit]
Description=Caddy
Documentation=https://caddyserver.com/docs/
After=network.target network-online.target
Requires=network-online.target

[Service]
User=caddy
Group=caddy
ExecStart=/usr/bin/caddy run --environ --config /etc/caddy/Caddyfile
ExecReload=/usr/bin/caddy reload --config /etc/caddy/Caddyfile
TimeoutStopSec=5s
LimitNOFILE=1048576
LimitNPROC=512
PrivateTmp=true
ProtectSystem=full
AmbientCapabilities=CAP_NET_BIND_SERVICE

[Install]
WantedBy=multi-user.target
```

[...]

PR #4443, opened 2021-11-25
pki: Implement `GET /pki/certificates/<id>` API, rework `caddy trust`

francislavoie:
Fixes https://github.com/caddyserver/caddy/issues/4058, once this is completed. This is basically half-way to solving the linked issue.

This implements a new admin API, `GET /pki/certificates/<id>`, which returns information about the given PKI app's CA, by ID. 

For example, this is what a response might look like, for requesting info about the `local` CA (the default CA ID):

```bash
$ curl localhost:2019/pki/certificates/local
```
```json
{
  "id": "local",
  "name": "Caddy Local Authority",
  "root": "-----BEGIN CERTIFICATE-----\nMIIBpDCCAUmgAwIBAgIQTS5a+3LUKNxC6qN3ZDR8bDAKBggqhkjOPQQDAjAwMS4w\nLAYDVQQDEyVDYWRkeSBMb2NhbCBBdXRob3JpdHkgLSAyMDIxIEVDQyBSb290MB4X\nDTIxMTEyNTA1NDIxMloXDTMxMTAwNDA1NDIxMlowMDEuMCwGA1UEAxMlQ2FkZHkg\nTG9jYWwgQXV0aG9yaXR5IC0gMjAyMSBFQ0MgUm9vdDBZMBMGByqGSM49AgEGCCqG\nSM49AwEHA0IABIFW7LqyWf7ybSXh6FREAgGL+Bz9TktcmcmiKHAZXhWH8zCqun3k\nIg2mya+GgnA0k39OPrCaVUUMb+96UfldA9ujRTBDMA4GA1UdDwEB/wQEAwIBBjAS\nBgNVHRMBAf8ECDAGAQH/AgEBMB0GA1UdDgQWBBTrVHm2qnZLcRzky7u1NO2n8jUJ\n5DAKBggqhkjOPQQDAgNJADBGAiEA2JTbqaLhEbki73/Eb4k/KBVzSTwSKzqbssKQ\n9M9t0FwCIQCAlUr4ZlFzHE/3K6dARYKusR1ck4A3MtucSSyar6lgRw==\n-----END CERTIFICATE-----\n",
  "intermediate": "-----BEGIN CERTIFICATE-----\nMIIByDCCAW2gAwIBAgIQGRTBhcgdsiLXv0AA6r8GCTAKBggqhkjOPQQDAjAwMS4w\nLAYDVQQDEyVDYWRkeSBMb2NhbCBBdXRob3JpdHkgLSAyMDIxIEVDQyBSb290MB4X\nDTIxMTEyNTA1NDIxMloXDTIxMTIwMjA1NDIxMlowMzExMC8GA1UEAxMoQ2FkZHkg\nTG9jYWwgQXV0aG9yaXR5IC0gRUNDIEludGVybWVkaWF0ZTBZMBMGByqGSM49AgEG\nCCqGSM49AwEHA0IABNHTWLDx5jxS7x3WYEBIOXfuxQP2yvfitpEPiHJNJsf44XF8\ns31IBrDcd81Xzn63b8S2qn6elSTnZK1DdrCdtnSjZjBkMA4GA1UdDwEB/wQEAwIB\nBjASBgNVHRMBAf8ECDAGAQH/AgEAMB0GA1UdDgQWBBTvu/Kpad1XWLbaTHoqNS4U\nLBSfBTAfBgNVHSMEGDAWgBTrVHm2qnZLcRzky7u1NO2n8jUJ5DAKBggqhkjOPQQD\nAgNJADBGAiEAhUhwv6EQCgS+SGMnaze5FvHk0a9yfhDCdqmAdadhsjsCIQCTTNZF\njFM8CQL3vCfpCVzTiPR8EUbaCqxj8O9sg/70dQ==\n-----END CERTIFICATE-----\n"
}
```

To get there though, I had to refactor a bit of the `admin.go` and `caddy.go` internals. The tricky bit is that this is the first `AdminRouter` module that needs a `caddy.Context` to be able to fetch Caddy apps from the current config at runtime.

To solve it, I set up a temporary slice in `AdminConfig` to remember the routers that had their routes registered (previously, they were just forgotten immediately after), so that we can later call `Provision(ctx)` on them (if they need to be provisioned).

I also had to rework how `replaceLocalAdminServer` handles the default admin config, because I needed to make sure it would write the actual admin config in use back to the `caddy.Config` so that we could later do the provisioning stuff. Previously, it would just read from the config if it was non-nil, or use a default, and spin up the admin server. But that's not really consistent with how the rest of Caddy works, because all the other apps are actually kept as references in `caddy.Config`, I think. It also gets rid of a global var, which is nice I guess.

---

All that said, this isn't enough to call the issue done, we still need to implement the changes to the `caddy trust` so that it actually uses this new API, which would make the `caddy trust` process not need to care about its own storage as a side effect, instead it would always reach out to the API to get the correct root CA cert.

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -202,7 +202,7 @@ func cmdRun(fl Flags) (int, error) {
 	// we don't use 'else' here since this value might have been changed in 'if' block; i.e. not mutually exclusive
 	var configFile string
 	if !runCmdResumeFlag {
-		config, configFile, err = loadConfig(runCmdConfigFlag, runCmdConfigAdapterFlag)
+		config, configFile, err = LoadConfig(runCmdConfigFlag, runCmdConfigAdapterFlag)
 		if err != nil {
 			return caddy.ExitCodeFailedStartup, err
 		}
@@ -277,11 +277,12 @@ func cmdRun(fl Flags) (int, error) {
 func cmdStop(fl Flags) (int, error) {
 	stopCmdAddrFlag := fl.String("address")
 
-	err := apiRequest(stopCmdAddrFlag, http.MethodPost, "/stop", nil, nil)
+	resp, err := ApiRequest(stopCmdAddrFlag, http.MethodPost, "/stop", nil, nil)
 	if err != nil {
 		caddy.Log().Warn("failed using API to stop instance", zap.Error(err))
 		return caddy.ExitCodeFailedStartup, err
 	}
+	defer resp.Body.Close()
 
 	return caddy.ExitCodeSuccess, nil
 }
@@ -293,7 +294,7 @@ func cmdReload(fl Flags) (int, error) {
 	reloadCmdForceFlag := fl.Bool("force")
 
 	// get the config in caddy's native format
-	config, configFile, err := loadConfig(reloadCmdConfigFlag, reloadCmdConfigAdapterFlag)
+	config, configFile, err := LoadConfig(reloadCmdConfigFlag, reloadCmdConfigAdapterFlag)
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, err
 	}
@@ -321,10 +322,11 @@ func cmdReload(fl Flags) (int, error) {
 		headers.Set("Cache-Control", "must-revalidate")
 	}
 
-	err = apiRequest(adminAddr, http.MethodPost, "/load", headers, bytes.NewReader(config))
+	resp, err := ApiRequest(adminAddr, http.MethodPost, "/load", headers, bytes.NewReader(config))
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, fmt.Errorf("sending configuration to instance: %v", err)
 	}
+	defer resp.Body.Close()
 
 	return caddy.ExitCodeSuccess, nil
 }
@@ -518,7 +520,7 @@ func cmdValidateConfig(fl Flags) (int, error) {
 	validateCmdConfigFlag := fl.String("config")
 	validateCmdAdapterFlag := fl.String("adapter")
 
-	input, _, err := loadConfig(validateCmdConfigFlag, validateCmdAdapterFlag)
+	input, _, err := LoadConfig(validateCmdConfigFlag, validateCmdAdapterFlag)
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, err
 	}
@@ -640,17 +642,18 @@ commands:
 	return caddy.ExitCodeSuccess, nil
 }
 
-// apiRequest makes an API request to the endpoint adminAddr with the
+// ApiRequest makes an API request to the endpoint adminAddr with the
 // given HTTP method and request URI. If body is non-nil, it will be
-// assumed to be Content-Type application/json.
-func apiRequest(adminAddr, method, uri string, headers http.Header, body io.Reader) error {
+// assumed to be Content-Type application/json. The caller should
+// close the response body.
+func ApiRequest(adminAddr, method, uri string, headers http.Header, body io.Reader) (*http.Response, error) {
 	// parse the admin address
 	if adminAddr == "" {
 		adminAddr = caddy.DefaultAdminListen
 	}
 	parsedAddr, err := caddy.ParseNetworkAddress(adminAddr)
 	if err != nil || parsedAddr.PortRangeSize() > 1 {
-		return fmt.Errorf("invalid admin address %s: %v", adminAddr, err)
+		return nil, fmt.Errorf("invalid admin address %s: %v", adminAddr, err)
 	}
 	origin := parsedAddr.JoinHostPort(0)
 	if parsedAddr.IsUnixNetwork() {
@@ -660,7 +663,7 @@ func apiRequest(adminAddr, method, uri string, headers http.Header, body io.Read
 	// form the request
 	req, err := http.NewRequest(method, "http://"+origin+uri, body)
 	if err != nil {
-		return fmt.Errorf("making request: %v", err)
+		return nil, fmt.Errorf("making request: %v", err)
 	}
 	if parsedAddr.IsUnixNetwork() {
 		// When listening on a unix socket, the admin endpoint doesn't
@@ -700,20 +703,19 @@ func apiRequest(adminAddr, method, uri string, headers http.Header, body io.Read
 
 	resp, err := client.Do(req)
 	if err != nil {
-		return fmt.Errorf("performing request: %v", err)
+		return nil, fmt.Errorf("performing request: %v", err)
 	}
-	defer resp.Body.Close()
 
 	// if it didn't work, let the user know
 	if resp.StatusCode >= 400 {
 		respBody, err := io.ReadAll(io.LimitReader(resp.Body, 1024*10))
 		if err != nil {
-			return fmt.Errorf("HTTP %d: reading error message: %v", resp.StatusCode, err)
+			return nil, fmt.Errorf("HTTP %d: reading error message: %v", resp.StatusCode, err)
 		}
-		return fmt.Errorf("caddy responded with error: HTTP %d: %s", resp.StatusCode, respBody)
+		return nil, fmt.Errorf("caddy responded with error: HTTP %d: %s", resp.StatusCode, respBody)
 	}
 
-	return nil
+	return resp, nil
 }
 
 type moduleInfo struct {
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -275,9 +275,16 @@ func cmdRun(fl Flags) (int, error) {
 }
 
 func cmdStop(fl Flags) (int, error) {
-	stopCmdAddrFlag := fl.String("address")
+	addrFlag := fl.String("address")
+	configFlag := fl.String("config")
+	configAdapterFlag := fl.String("adapter")
 
-	resp, err := ApiRequest(stopCmdAddrFlag, http.MethodPost, "/stop", nil, nil)
+	adminAddr, err := DetermineAdminAPIAddress(addrFlag, configFlag, configAdapterFlag)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, fmt.Errorf("couldn't determine admin API address: %v", err)
+	}
+
+	resp, err := AdminAPIRequest(adminAddr, http.MethodPost, "/stop", nil, nil)
 	if err != nil {
 		caddy.Log().Warn("failed using API to stop instance", zap.Error(err))
 		return caddy.ExitCodeFailedStartup, err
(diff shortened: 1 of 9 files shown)


francislavoie:
Alright, the `caddy trust` command is updated! Works well, it can target a running instance of Caddy to get the root cert using the new API.

Still TODO:
- Adjust `caddy untrust` command to also work by specifying an admin API address, in which case it would call `truststore.Uninstall()` (instead of `truststore.UninstallFile()`) with the root cert loaded from the API.

francislavoie:
Okay, this should be all done now! 🎉

tgelite:
Build worked for me over default CA.  

```
$ curl localhost:2019/pki/certificates/local
{"id":"local","name":"Caddy Local Authority","root":"-----BEGIN CERTIFICATE-----\nMIIBpDCCAUqgAwIBAgIRAMF6w1QoZjNJX6m7fVg1eLowCgYIKoZIzj0EAwIw...
Wbg1bbZunqoU0/xyFpX3G//OP+RnDGxwn32A==\n-----END CERTIFICATE-----\n","intermediate":"-----BEGIN CERTIFICATE-----\nMIIByDCCAW6gAwIBAgIRAMfweUlUgrt3/JiHF0hjAP8wCgYIKoZIzj0EAwIwMD...
lyQ1vqHdPzRpXn4AvQJKURA==\n-----END CERTIFICATE-----\n"}
```


```
$ go version
go version go1.17.3 linux/amd64
agrajag@agrajag-VirtualBox:~$ cat /etc/*release*
...
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
```

francislavoie:
@tgelite you can also play with the `caddy trust` command which now fetches from the API instead of reading from storage directly (so it should always get the right cert instead of needing to do the `HOME=~caddy` hack)

tgelite:
Yep worked as expected!
```
agrajag@agrajag-VirtualBox:~$ sudo caddy trust
2021/11/28 19:52:29.316	WARN	installing root certificate (you might be prompted for password)	{"path": "localhost:2019/pki/certificates/local"}
2021/11/28 12:52:29 not NSS security databases found
2021/11/28 12:52:29 define JAVA_HOME environment variable to use the Java trust
2021/11/28 12:52:30 certificate installed properly in linux trusts
agrajag@agrajag-VirtualBox:~$ ls -l /etc/ssl/certs/ | grep -i caddy
lrwxrwxrwx 1 root root     79 Nov 28 12:52 1e027192.0 -> Caddy_Local_Authority_-_2021_ECC_Root_6464412664498597703023796308403825355.pem
lrwxrwxrwx 1 root root    112 Nov 28 12:52 Caddy_Local_Authority_-_2021_ECC_Root_6464412664498597703023796308403825355.pem -> /usr/local/share/ca-certificates/Caddy_Local_Authority_-_2021_ECC_Root_6464412664498597703023796308403825355.crt
```

and then verifying system trust is working, use openssl to check (status code 0 means system trust is present for the tls connection...

```
agrajag@agrajag-VirtualBox:~$ openssl s_client -connect 192.168.37.170:443
CONNECTED(00000003)
Can't use SSL_get_servername
depth=2 CN = Caddy Local Authority - 2021 ECC Root
verify return:1
depth=1 CN = Caddy Local Authority - ECC Intermediate
verify return:1
depth=0 
verify return:1
---
Certificate chain
 0 s:
   i:CN = Caddy Local Authority - ECC Intermediate
 1 s:CN = Caddy Local Authority - ECC Intermediate
   i:CN = Caddy Local Authority - 2021 ECC Root
---
Server certificate
-----BEGIN CERTIFICATE-----
MIIBuDCCAV6gAwIBAgIQfYcPdmSNkTMQ05PxvTbAvjAKBggqhkjOPQQDAjAzMTEw
...
/x1jB1nLLd65hnti
-----END CERTIFICATE-----
subject=

issuer=CN = Caddy Local Authority - ECC Intermediate

---
No client certificate CA names sent
Peer signing digest: SHA256
Peer signature type: ECDSA
Server Temp Key: X25519, 253 bits
---
SSL handshake has read 1264 bytes and written 347 bytes
Verification: OK
---
New, TLSv1.3, Cipher is TLS_CHACHA20_POLY1305_SHA256
Server public key is 256 bit
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---
---
Post-Handshake New Session Ticket arrived:
SSL-Session:
    Protocol  : TLSv1.3
    Cipher    : TLS_CHACHA20_POLY1305_SHA256
    Session-ID: ED33FD226740D1BE181CD46312CD649A4DB75B5392B05215324079A6C33D054F
    Session-ID-ctx: 
    Resumption PSK: D22E500E689F46D95251BD17B37761FA634394A3C34BE1EBA3EAACB4F6ADCF69
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    TLS session ticket lifetime hint: 604800 (seconds)
    TLS session ticket:
    0000 - 85 cc e0 c7 fd 6c 7b 27-ae 86 c9 41 85 a9 76 d2   .....l{'...A..v.
    ...
    0060 - 61 4e 3a 80 82 03 52 95-89 b2 e7 ce 93 77 79 64   aN:...R......wyd
    0070 - 6a                                                j

    Start Time: 1638130173
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
    Extended master secret: no
    Max Early Data: 0
---
read R BLOCK
```
<ctrl-d> to exit the openssl interactive prompt...

modules/caddypki/adminpki.go
```diff
@@ -0,0 +1,194 @@
...
+
+// CAInfo is the response from the certificates API endpoint
+type CAInfo struct {
+	Id           string `json:"id"`
+	Name         string `json:"name"`
+	Root         string `json:"root"`
+	Intermediate string `json:"intermediate"`
+}
+
+const amdinPKICertificatesEndpoint = "/pki/certificates/"
```
mholt, line 188:
```suggestion
const adminPKICertificatesEndpoint = "/pki/certificates/"
```

modules/caddypki/adminpki.go
```diff
@@ -0,0 +1,194 @@
...
+			Err:        err,
+		}
+	}
+
+	return nil
+}
+
+// CAInfo is the response from the certificates API endpoint
+type CAInfo struct {
+	Id           string `json:"id"`
```
mholt, line 182:
```suggestion
	ID           string `json:"id"`
```

modules/caddypki/command.go
```diff
@@ -30,69 +32,153 @@ func init() {
...
+				return caddy.ExitCodeFailedStartup,
+					fmt.Errorf("unmarshaling admin listener address from config: %v", err)
+			}
+			adminAddr = tmpStruct.Admin.Listen
+		}
 	}
-	err := ca.Provision(ctx, DefaultCAID, caddy.Log())
+	if adminAddr == "" {
+		adminAddr = caddy.DefaultAdminListen
+	}
```
mholt, line 143: Yikes, this is complicated. (I know it's not new.)

If we're going to be exporting the function to perform API requests, I think the address of the admin endpoint should be irrelevant to the caller. I wonder what it would look like if we passed in the necessary information for the APIRequest() function to figure that out for us.

cmd/main.go
```diff
@@ -103,15 +103,15 @@ func handlePingbackConn(conn net.Conn, expect []byte) error {
...
 
-// loadConfig loads the config from configFile and adapts it
+// LoadConfig loads the config from configFile and adapts it
 // using adapterName. If adapterName is specified, configFile
 // must be also. If no configFile is specified, it tries
 // loading a default config file. The lack of a config file is
 // not treated as an error, but false will be returned if
 // there is no config available. It prints any warnings to stderr,
 // and returns the resulting JSON config bytes along with
 // whether a config file was loaded or not.
```
mholt, line 113:
```suggestion
// and returns the resulting JSON config bytes along with
// the name of the loaded config file (if any).
```

Not related to your change, but I just realized this comment could make more sense.

cmd/commandfuncs.go
```diff
@@ -640,17 +642,18 @@ commands:
...
 }
 
-// apiRequest makes an API request to the endpoint adminAddr with the
+// ApiRequest makes an API request to the endpoint adminAddr with the
 // given HTTP method and request URI. If body is non-nil, it will be
-// assumed to be Content-Type application/json.
-func apiRequest(adminAddr, method, uri string, headers http.Header, body io.Reader) error {
+// assumed to be Content-Type application/json. The caller should
+// close the response body.
+func ApiRequest(adminAddr, method, uri string, headers http.Header, body io.Reader) (*http.Response, error) {
```
mholt, line 649:
```suggestion
func APIRequest(adminAddr, method, uri string, headers http.Header, body io.Reader) (*http.Response, error) {
```

As mentioned in another comment, I have some concerns with this signature, especially since we're going to be exporting it. I get a lot of anxiety exporting functions and types these days 😅

Can the function take input to determine the admin address itself instead?

Or maybe can we leave this function as-is (except for, I guess, returning the response) and create a new function that takes the necessary inputs for determining the admin address. It would be realluy good if we didn't have to repeat this logic and have callers do that.

I also wonder if an exported function here could be named a little more specifically, like `AdminAPIRequest()` or something.

mholt (changes requested):
Thanks Francis, this is looking pretty good. Just had a couple nits/questions for your consideration. Thank you again, sorry it took me a while to get around to this. Long PRs are not my fave :)

modules/caddypki/adminpki.go
francislavoie: lolol oops 😅

modules/caddypki/command.go
francislavoie: I refactored this stuff to move it to a new function `caddycmd.DetermineAdminAPIAddress`

cmd/commandfuncs.go
francislavoie: Okay, this is what I did:

- Renamed it to `AdminAPIRequest()`
- Signature stayed the same, because the options are either
   1. Use `Flags` as the first arg instead and try to automatically parse the command's flags to get the address; this makes the Flags leak outside of the command's function itself, so this is kinda bleh (I did try it in code to see if it made sense)
   2. Change `adminAddr` to three string inputs instead, i.e. address, config, adapter... but that's also gross (or some setup where a struct with those strings are passed in, also gross)
   3. Keep it the same
- Moved the duplicated logic for determining the admin address to a helper `caddycmd.DetermineAdminAPIAddress()` which the 3 common CLI flags, now `run`, `stop`, `trust` and `untrust` all use this
- Moved the code that performs the admin API call and decodes the response in `trust` and `untrust` to a `caddypki.rootCertFromAdmin()` to avoid duplication.

This seems cleaner, but it doesn't _directly_ address your concerns. Not sure what the other options are, if there are any.

mholt: Thanks - of the all-not-great options, I like your approach the best I guess :)

mholt (approved):
Glorious -- let's give it a shot. Thank you for implementing this, I do think this is an overall improvement. It's neat having API access to the PKI functions and features. Hope it works! :crossed_fingers:

Issue #4329, opened 2021-09-03
404 on filenames with accents


bmichotte:
Hi there,

Caddy gives me an error 404 for files with accents like `cropped-Capture-d’écran-2019-12-02-à-16.22.37.png` (yes, I know, this is awfull), which are called `cropped-Capture-d%E2%80%99%C3%A9cran-2019-12-02-%C3%A0-16.22.37.png` by the browser.

Obviously the file does exists and the files without accents gives a `200 OK`.

Log file is 
```json
{
    "level": "error",
    "ts": 1630683589.3759747,
    "logger": "http.log.access.log16",
    "msg": "handled request",
    "request": {
        "remote_addr": "91.180.123.81:51607",
        "proto": "HTTP/2.0",
        "method": "GET",
        "host": "[REDACTED]",
        "uri": "/wp-content/uploads/2019/12/cropped-Capture-d%E2%80%99%C3%A9cran-2019-12-02-%C3%A0-16.22.37.png",
        "headers": {
            "Upgrade-Insecure-Requests": [
                "1"
            ],
            "Sec-Fetch-Dest": [
                "document"
            ],
            "User-Agent": [
                "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.4577.63 Safari/537.36"
            ],
            "Accept": [
                "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9"
            ],
            "Sec-Fetch-User": [
                "?1"
            ],
            "Accept-Encoding": [
                "gzip, deflate, br"
            ],
            "Accept-Language": [
                "fr-FR,fr;q=0.9,en-US;q=0.8,en;q=0.7"
            ],
            "Sec-Ch-Ua": [
                "\"Google Chrome\";v=\"93\", \" Not;A Brand\";v=\"99\", \"Chromium\";v=\"93\""
            ],
            "Sec-Ch-Ua-Mobile": [
                "?0"
            ],
            "Sec-Fetch-Site": [
                "none"
            ],
            "Sec-Fetch-Mode": [
                "navigate"
            ],
            "Cookie": [
                "wp-settings-time-2=1617279544; wordpress_test_cookie=WP+Cookie+check"
            ],
            "Sec-Ch-Ua-Platform": [
                "\"macOS\""
            ],
[...]

PR #4447, opened 2021-11-26
fileserver: do not double-escape paths

mohammed90:
Fixes the issue specified in https://github.com/caddyserver/caddy/issues/4329#issuecomment-976957226.


mohammed90:
@bmichotte  and @ueffel, please check if the patch resolves your issue and doesn't break new stuff.

ueffel:
```shell
$ ./caddy version
v2.4.7-0.20211126140144-80fbec0102d5 h1:mdBZC644zlsG4dWbTG3YcsU34linp0ODDoMfaogqSB0=
```

It does not seem to be fixed. 
image

I think the escaping [here](https://github.com/caddyserver/caddy/blob/80fbec0102d59ac2dd0ec2a3304c8f5c3c6b520e/modules/caddyhttp/fileserver/browsetplcontext.go#L46) is causing the double encoding. This bit:

```go
u := url.URL{Path: url.PathEscape(name)}
```

mohammed90:
> 

Now I can see why. While we're escaping it in that line, the template also calls [the `html` function from the text/template package of stdlib](https://pkg.go.dev/text/template@master#hdr-Functions):

https://github.com/caddyserver/caddy/blob/6f9b6ad78e6b9a40afbf12e53017e39abee90a6d/modules/caddyhttp/fileserver/browse.html#L390-L396

The documentation of the function is:

> html
	Returns the escaped HTML equivalent of the textual
	representation of its arguments. This function is unavailable
	in html/template, with a few exceptions.


I've updated the code. This effectively reverts the buggy earlier PR but keeps the tests, which were passing anyways.

francislavoie (approved):
Seems good 👍

PR #4449, opened 2021-11-26
ci: Add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which Caddy already runs.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
I think the issue is CIFuzz related, I have a fix - will deploy it on OSS-Fuzz shortly and ping here!

mohammed90:
> @mohammed90 there should now be a fix in place, could we try to rerun the CI?

I've triggered the re-run

mholt:
Thanks, but we already have fuzzing, and we don't have much to fuzz, and I don't think we need another CI job, especially for fuzzing at this point.

PR #4463, opened 2021-12-07
pki: Avoid provisioning the `local` CA when not necessary

### LINKED ISSUE #4461: PKI app's `local` CA attempts installation even when not configured
See https://caddy.community/t/defining-own-ca-using-the-pki-module-fails-with-no-such-file-or-directory/14330 for context.

Opening this as a reminder for myself to look into this, to figure out why installation still happens when only another CA ID is configured.

francislavoie:
Fixes #4461

Previously, we would provision the `local` CA no matter what, if the PKI app has a config. That doesn't really make sense in all cases.

Instead, we should provision it only if there's no other CA, and if either `tls internal` is configured (implicitly or otherwise), or the `acme_server` is configured with the default CA.

I think this should have slightly less surprising behaviour for users who want to configure their own CA with an ID that isn't `local`.

### DIFF
--- a/modules/caddypki/pki.go
+++ b/modules/caddypki/pki.go
@@ -34,6 +34,8 @@ func init() {
 type PKI struct {
 	// The certificate authorities to manage. Each CA is keyed by an
 	// ID that is used to uniquely identify it from other CAs.
+	// At runtime, the GetCA() method should be used instead to ensure
+	// the default CA is provisioned if it hadn't already been.
 	// The default CA ID is "local".
 	CAs map[string]*CA `json:"certificate_authorities,omitempty"`
 
@@ -53,10 +55,6 @@ func (PKI) CaddyModule() caddy.ModuleInfo {
 func (p *PKI) Provision(ctx caddy.Context) error {
 	p.ctx = ctx
 	p.log = ctx.Logger(p)
-
-	if p.CAs == nil {
-		p.CAs = make(map[string]*CA)
-	}
 
 	for caID, ca := range p.CAs {
 		err := ca.Provision(ctx, caID, p.log)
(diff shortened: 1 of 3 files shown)


modules/caddypki/pki.go
```diff
@@ -54,15 +54,9 @@ func (p *PKI) Provision(ctx caddy.Context) error {
 	p.ctx = ctx
 	p.log = ctx.Logger(p)
 
-	// if this app is initialized at all, ensure there's at
-	// least a default CA that can be used: the standard CA
-	// which is used implicitly for signing local-use certs
 	if p.CAs == nil {
 		p.CAs = make(map[string]*CA)
 	}
-	if _, ok := p.CAs[DefaultCAID]; !ok {
```
mholt, line 63: Can probably remove the lines above it too (L60-62), since ProvisionDefaultCA() makes sure the map isn't nil.

mholt (commented):
Thanks for this patch. Although, I am not really loving that other modules have to set up an app's defaults. It is much better for the app to set its own defaults, no? Does simply moving the provisioning of the default CA to after the other CAs are initialized not help solve the problem?

modules/caddypki/pki.go
francislavoie: Doesn't it need to be non-nil to `range p.CAs`?

mholt: Nope, reads on nil maps are safe.

mholt:
That's not a bad idea. We'd want to document that on the App.CAs field that it's better to use GetCA() than accessing it directly. So yeah, I'm curious what that'll look like.

modules/caddytls/internalissuer.go
```diff
@@ -68,18 +67,20 @@ func (InternalIssuer) CaddyModule() caddy.ModuleInfo {
 func (iss *InternalIssuer) Provision(ctx caddy.Context) error {
 	iss.logger = ctx.Logger(iss)
 
+	// set some defaults
+	if iss.CA == "" {
+		iss.CA = caddypki.DefaultCAID
+	}
+
```
mholt, line 74: Not sure why this moved to above L76-79, but I guess that's ok. :man_shrugging:

mholt (approved):
Looking better, thanks very much for this.

modules/caddytls/internalissuer.go
francislavoie: I moved it to match the code in `acmeserver.go` which is very similar

PR #4468, opened 2021-12-08
reverseproxy: Better handling of request headers during proxying

### LINKED ISSUE #4457: header_up -field in reverse_proxy will affect placeholders {http.request.header.field}?
### 1. Caddy version (`caddy version`):

`v2.4.6 h1:HGkGICFGvyrodcqOOclHKfvJC0qTU7vny/7FhYp9hNw=`

### 2. How I run Caddy:

```
:12345 {
    respond "hello world! {http.request.header.Origin}"
}

:23456 {
    @origin {
        header Origin *
    }
    @preflight {
        header Origin *
        method OPTIONS
    }
    header @origin {
        Access-Control-Allow-Origin "{http.request.header.Origin}"
        Access-Control-Allow-Credentials true
        Access-Control-Expose-Headers "Content-Length,Content-Range"
        defer
    }
    header @preflight {
        Access-Control-Allow-Methods "GET,POST,PUT,DELETE,HEAD,OPTIONS"
        Access-Control-Allow-Headers "{http.request.header.Access-Control-Request-Headers}"
        Access-Control-Max-Age 86400
    }
    reverse_proxy http://localhost:12345 {
        header_up -Origin
        header_down -Server
    }
}
```

#### a. System environment:

`Docker 20.10.10`

#### b. Command:

```
caddy run;
```

```
docker run --name caddy --net=host -e TZ=Asia/Shanghai -e GO111MODULE=on -e GOPROXY=https://goproxy.cn,direct --restart=always -v /data/www:/data/www:ro -v /opt/caddy/config:/etc/caddy:ro -v /opt/caddy/log:/log -w /etc/caddy -d caddy:builder-alpine sh -c "if [ ! -x /usr/bin/caddy ]; then xcaddy build --output /usr/bin/caddy --with github.com/caddyserver/format-encoder; fi; caddy run"
```

```
:12345 {
    respond "hello world! {http.request.header.Origin}"
}

:23456 {
    @origin {
        header Origin *
    }
    @preflight {
        header Origin *
        method OPTIONS
    }
    header @origin {
        Access-Control-Allow-Origin "{http.request.header.Origin}"
        Access-Control-Allow-Credentials true
        Access-Control-Expose-Headers "Content-Length,Content-Range"
        defer
    }
    header @preflight {
        Access-Control-Allow-Methods "GET,POST,PUT,DELETE,HEAD,OPTIONS"
        Access-Control-Allow-Headers "{http.request.header.Access-Control-Request-Headers}"
        Access-Control-Max-Age 86400
    }
    reverse_proxy http://localhost:12345 {
        header_up -Origin
        header_down -Server
    }
}
```

### 3. The problem I'm having:

I want to use caddy to support CORS and disable the CORS logic of the upstream.
```
curl -v http://localhost:23456 -H "Origin: test"
*   Trying ::1...
* TCP_NODELAY set
* Expire in 150000 ms for 3 (transfer 0x55695f06af50)
* Expire in 200 ms for 4 (transfer 0x55695f06af50)
* Connected to localhost (::1) port 23456 (#0)
> GET / HTTP/1.1
> Host: localhost:23456
> User-Agent: curl/7.64.0
> Accept: */*
> Origin: test
> 
< HTTP/1.1 200 OK
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Origin: 
< Access-Control-Expose-Headers: Content-Length,Content-Range
< Content-Length: 13
< Date: Sun, 05 Dec 2021 04:56:33 GMT
< Server: Caddy
< 
hello world! 
```

### 5. What I already tried:

I couldn't find the relevant information in the forum, and then pulled out the simplest Caddyfile, which contains no other information and can be reproduced 100%.

francislavoie:
Fixes #4457

Previously, we would only reset the request headers (and host, URL) in a `defer`, but this causes trouble in cases where a ResponseWriterWrapper is used (which triggers when the upstream response is copied to the writer) and the original headers are expected while the wrapper is performing its job.

I refactored what I'm calling "preserving the mutable parts of the request" into a separate function, which cuts down on the size of the `ServeHTTP()` function body (nice win).

The key part is that we pass the `resetRequest()` closure to `h.reverseProxy()` so that it can do the reset right after having done the roundtrip, and before writing the response.

The `resetRequest()` closure will probably be called more than once, but that should be totally fine, cause it's just setting a couple strings and replacing a pointer, basically, and should be a no-op on repeated calls. There's 3 places it can get called:

- In `h.reverseProxy`, this is the important one, to fix the bug.
- At the end of the loop, this is to make sure the reset happens before `h.preserveRequest(r)` is called again on the next loop iteration. Not strictly necessary, but seems like a reasonable safety check.
- In a `defer` so that any `return` or `break` inside the loop will still trigger the reset.

There was a TODO about data races in here. I'm not sure if this will have any kind of effect on that. I didn't analyze for data races, but I think this all makes sense from a sequential request handling standpoint.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -381,32 +381,6 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyht
 			fmt.Errorf("preparing request for upstream round-trip: %v", err))
 	}
 
-	// we will need the original headers and Host value if
-	// header operations are configured; and we should
-	// restore them after we're done if they are changed
-	// (for example, changing the outbound Host header
-	// should not permanently change r.Host; issue #3509)
-	reqHost := r.Host
-	reqHeader := r.Header
-
-	// sanitize the request URL; we expect it to not contain the scheme and host
-	// since those should be determined by r.TLS and r.Host respectively, but
-	// some clients may include it in the request-line, which is technically
-	// valid in HTTP, but breaks reverseproxy behaviour, overriding how the
-	// dialer will behave. See #4237 for context.
-	origURLScheme := r.URL.Scheme
-	origURLHost := r.URL.Host
-	r.URL.Scheme = ""
-	r.URL.Host = ""
-
-	// restore modifications to the request after we're done proxying
-	defer func() {
-		r.Host = reqHost     // TODO: data race, see #4038
-		r.Header = reqHeader // TODO: data race, see #4038
-		r.URL.Scheme = origURLScheme
-		r.URL.Host = origURLHost
-	}()
-
 	start := time.Now()
 	defer func() {
 		// total proxying duration, including time spent on LB and retries
@@ -449,6 +423,12 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyht
 		repl.Set("http.reverse_proxy.upstream.max_requests", upstream.MaxRequests)
 		repl.Set("http.reverse_proxy.upstream.fails", upstream.Host.Fails())
 
+		// preserve the mutable parts of the request, and
+		// set up a defer to make sure it's always reset
+		// by the end of the proxy handler
+		reqHost, reqHeader, resetRequest := h.preserveRequest(r)
+		defer resetRequest()
+
 		// mutate request headers according to this upstream;
 		// because we're in a retry loop, we have to copy
 		// headers (and the r.Host value) from the original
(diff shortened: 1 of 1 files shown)


mholt (commented):
Thanks Francis, I like this change.

Do you think copying the request before proxying would solve both this problem and the data race? (The data race occurs in the std lib, so we can't just add locking around it.)

If so, I'm wondering if copying the request -- to the extent needed to avoid this problem and the data races -- is worthwhile. Could be a simpler fix and also fix both problems.

francislavoie:
Maybe copying would be more "correct", but it would most likely be less performant, right? Probably fine to stay like this for now I think. I don't feel like benchmarking to make sure it's not a bigger regression by copying.

mholt:
Yeah, it will of course allocate. But maybe all we need to deep-copy is the headers and URL; we might not need to use the full `http.Request.Clone()` method. I'd like to explore what a copy of this nature looks like, and if it solves both these problems, quite possibly would be interested in using that instead.

mholt:
Closing in favor of #4551 which also fixes the data race

Issue #4446, opened 2021-11-26
backend weight and backup option


dazoot:
In nginx / haproxy i can set `weight` for specific backends and `backup` (stand by failover) options.
Would love to have this in caddy too.

francislavoie:
We're working on a rewrite of how upstreams are configured, and having weights is planned (because it'll support using SRV lookups for configuring upstreams, which has weights).

For `backup`, you can already do this, essentially, by using `lb_policy first` and enabling health checks so that the primary upstream is marked as unhealthy, which would trigger falling back to the next upstream.

mholt:
SRV upstreams implemented in #4470 now honor the weight (and priority) fields of SRV records, which should do what you need.

If not, feel free to reopen this issue and explain more specifically what is required. Thanks!

Issue #4174, opened 2021-05-20
SRV lookups only use first response


sorenisanerd:
https://github.com/caddyserver/caddy/blob/e3c369d4526e44f23efb10aaad8a60ce519720a0/modules/caddyhttp/reverseproxy/hosts.go#L152

Even if the the DNS randomizes the order of entries (while still observing priorities), it's still perfectly possible for all the top-priority servers to be at capacity, in which case the we should move on to entries with lower priority.

francislavoie:
Please elaborate, I don't think there's really enough detail to make this actionable.

PRs welcome, if you know the fix that needs to be made.

sorenisanerd:
An SRV lookup returns all SRV entries. Each entry has a priority and a weight.

A correct use of that response should group the entries by priority and attempt to use the top (numerically lowest) priority entries first. Their respective weights should define their odds of getting picked. Once the top priority group is exhausted with no succesful reply, only then should we move on to the next group.

However, Caddy will only use a single entry, namely the first one. Even with many SRV entries, an incoming request will fail to be served correctly if the first entry happens to be one for a server that is unhealthy.

francislavoie:
I think you need to enabled `lb_try_duration` for the retry logic to kick in.

https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#load-balancing

sorenisanerd:
The gold plated way to address this would be:
1) The addition of a selection policy that (independent of SRV lookups) allows specifying both priority and weight to backends.
2) An incoming request triggers an SRV lookup which populates the backend data for the above mentioned selection policy which then proceeds as usual.

sorenisanerd:
`lb_try_duration` will do me little good if it'll only ever try the first entry. There's no guarantee that the DNS will randomize responses and even if it did, I don't see how Caddy would ever fall back to an entry that does not have top priority.

sorenisanerd:
For clarity, I understand that most people probably have Consul or something similar maintaining their SRV records, so you should never get SRV entries for unhealthy backends. In that case, everything is fine. My use case is somewhat different.

francislavoie:
This kinda goes hand-in-hand with https://github.com/caddyserver/caddy/issues/1545

Mohammed90:
Jotting down thoughts as I took glances at the code to understand what it could touch:

[...]

Issue #3801, opened 2020-10-16
lookup_srv resolvers configuration is still using systems resolv.conf


danlsgiga:
Caddy version: `v2.2.1`

The feature introduced in https://github.com/caddyserver/caddy/pull/3479 is not using the specified `addresses` for upstream DNS resolution.

Here are is an error log that is occurring every minute in my setup:

```json
{"level":"error","ts":1602812552.1035602,"logger":"http.log.error.log0","msg":"making dial info: lookup app.service.consul on 169.254.169.253:53: dial udp 169.254.169.253:53: operation was canceled","request":{"remote_addr":"24.88.156.180:4608","proto":"HTTP/2.0","method":"OPTIONS","host":"my.app.site","uri":"/registry/user/ribbon","headers":{"Accept-Language":["en-US,en;q=0.5"],"Accept-Encoding":["gzip, deflate, br"],"User-Agent":["Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:78.0) Gecko/20100101 Firefox/78.0"],"Accept":["*/*"],"Access-Control-Request-Method":["GET"],"Access-Control-Request-Headers":["authorization,x-client-app"],"Referer":["https://app2.site/computers/29a8240d-6fd2-4464-a187-3227276b398f"],"Origin":["https://app2.site"],"Dnt":["1"],"Te":["trailers"]},"tls":{"resumed":false,"version":772,"cipher_suite":4865,"proto":"h2","proto_mutual":true,"server_name":"my.app.site"}},"duration":0.000193545,"status":502,"err_id":"3jv1hi6bp","err_trace":"reverseproxy.(*Handler).ServeHTTP (reverseproxy.go:388)"}
```

The `169.254.169.253` IP is only present in my /etc/resolv.conf and should never be used after I have specified the `resolvers` in the `transport` config.

My config:

```json
{
  "admin": {
    "disabled": true
  },
  "apps": {
    "http": {
      "http_port": 80,
      "servers": {
        "srv0": {
          "automatic_https": {
            "disable": true,
            "disable_redirects": true
          },
          "listen": [
            ":80"
          ],
          "logs": {
            "default_logger_name": "log0"
          },
          "routes": [
            {
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "group": "app",
                      "handle": [
                        {
                          "handler": "reverse_proxy",
                          "transport": {
                            "protocol": "http",
                            "resolver": {
                              "addresses": [
[...]

PR #4470, opened 2021-12-08
reverseproxy: Dynamic upstreams (with support for SRV and A/AAAA lookups)

### LINKED ISSUE #4341: Active health check not separated for same upstream

mholt:
Right now, proxy upstreams have to be hard-coded into the config (with the only dyanamism coming from placeholders, which all act as a single upstream anyway). This change adds supports for truly dynamic upstreams, with the potential for every request to have different upstreams -- not only every request, but every _retry_ within a single request, too.

Instead of (or in addition to) specifying `upstreams` in your config, you could specify `dynamic_upstreams` and then define your upstream source module. Currently I'm implementing SRV and A/AAAA lookups as sources.

Fixes or related to:

- #1545 (currently the oldest open issue)
- #4174
- A currently-private sponsor's request. Hope to reveal this later!

When we're done, it will hopefully close or supercede #4350, #3801, #4446, and #4245. Closes #4341 too.

For dynamic SRV lookups:

```json
{
    "handler": "reverse_proxy",
    "dynamic_upstreams": {
        "source": "srv",
        "name": "_service._proto.name"
     }
}
```

For dynamic A/AAAA DNS lookups:

```json
{
    "handler": "reverse_proxy",
    "dynamic_upstreams": {
        "source": "a",
        "name": "backends.example.com",
        "port": "8080"
     }
}
```

The default refresh interval for lookups is 1 minute, but can be configured with the `refresh` property. SRV lookups can also optionally be configured by their individual service, proto, and name components, or the entire domain as the name by itself.

Still need to integrate health checks, add docs and tests, and cache eviction.

May be of interest to @danlsgiga, @dazoot, @sorenisanerd, and @furkanmustafa.

CC: @jjiang-stripe and @cds2-stripe

### DIFF
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -31,6 +31,7 @@ type UpstreamPool []*Upstream
 
 // Upstream bridges this proxy's configuration to the
 // state of the backend host it is correlated with.
+// Upstream values must not be copied.
 type Upstream struct {
 	*Host `json:"-"`
 
@@ -48,7 +49,7 @@ type Upstream struct {
 
 	// DEPRECATED: Use the SRVUpstreams module instead
 	// (http.reverse_proxy.upstreams.srv). This field will be
-	// removed in a future version of Caddy.
+	// removed in a future version of Caddy. TODO: Remove this field.
 	//
 	// If DNS SRV records are used for service discovery with this
 	// upstream, specify the DNS name for which to look up SRV
@@ -68,6 +69,7 @@ type Upstream struct {
 	activeHealthCheckPort int
 	healthCheckPolicy     *PassiveHealthChecks
 	cb                    CircuitBreaker
+	unhealthy             int32 // accessed atomically; status from active health checker
 }
 
 func (u Upstream) String() string {
@@ -90,7 +92,7 @@ func (u *Upstream) Available() bool {
 // is currently known to be healthy or "up".
 // It consults the circuit breaker, if any.
 func (u *Upstream) Healthy() bool {
-	healthy := !u.Host.Unhealthy()
+	healthy := u.healthy()
 	if healthy && u.healthCheckPolicy != nil {
 		healthy = u.Host.Fails() < u.healthCheckPolicy.MaxFails
 	}
@@ -115,7 +117,7 @@ func (u *Upstream) fillDialInfo(r *http.Request) (DialInfo, error) {
 	var addr caddy.NetworkAddress
 
 	if u.LookupSRV != "" {
-		// perform DNS lookup for SRV records and choose one
+		// perform DNS lookup for SRV records and choose one - TODO: deprecated
 		srvName := repl.ReplaceAll(u.LookupSRV, "")
 		_, records, err := net.DefaultResolver.LookupSRV(r.Context(), "", "", srvName)
 		if err != nil {
@@ -156,12 +158,11 @@ func (u *Upstream) fillHost() {
 	u.Host = host
 }
 
-// Host is the basic, in-memory representation
-// of the state of a remote host.
+// Host is the basic, in-memory representation of the state of a remote host.
+// Its fields are accessed atomically and Host values must not be copied.
 type Host struct {
 	numRequests int64 // must be 64-bit aligned on 32-bit systems (see https://golang.org/pkg/sync/atomic/#pkg-note-BUG)
 	fails       int64
-	unhealthy   int32 // TODO: this should be removed from Host; this decision should be made by the health checker
 }
 
 // NumRequests returns the number of active requests to the upstream.
@@ -174,14 +175,9 @@ func (h *Host) Fails() int {
 	return int(atomic.LoadInt64(&h.fails))
 }
 
-// Unhealthy returns whether the upstream is healthy.
-func (h *Host) Unhealthy() bool {
-	return atomic.LoadInt32(&h.unhealthy) == 1
-}
-
-// CountRequest mutates the active request count by
+// countRequest mutates the active request count by
 // delta. It returns an error if the adjustment fails.
-func (h *Host) CountRequest(delta int) error {
+func (h *Host) countRequest(delta int) error {
 	result := atomic.AddInt64(&h.numRequests, int64(delta))
 	if result < 0 {
 		return fmt.Errorf("count below 0: %d", result)
@@ -189,9 +185,9 @@ func (h *Host) CountRequest(delta int) error {
 	return nil
 }
 
-// CountFail mutates the recent failures count by
+// countFail mutates the recent failures count by
 // delta. It returns an error if the adjustment fails.
-func (h *Host) CountFail(delta int) error {
+func (h *Host) countFail(delta int) error {
 	result := atomic.AddInt64(&h.fails, int64(delta))
 	if result < 0 {
 		return fmt.Errorf("count below 0: %d", result)
@@ -199,15 +195,21 @@ func (h *Host) CountFail(delta int) error {
 	return nil
 }
 
+// healthy returns true if the upstream is not actively marked as unhealthy.
+// (This returns the status only from the "active" health checks.)
+func (u *Upstream) healthy() bool {
+	return atomic.LoadInt32(&u.unhealthy) == 0
+}
+
 // SetHealthy sets the upstream has healthy or unhealthy
-// and returns true if the new value is different.
-func (h *Host) SetHealthy(healthy bool) (bool, error) {
+// and returns true if the new value is different. This
+// sets the status only for the "active" health checks.
+func (u *Upstream) setHealthy(healthy bool) bool {
 	var unhealthy, compare int32 = 1, 0
 	if healthy {
 		unhealthy, compare = 0, 1
 	}
-	swapped := atomic.CompareAndSwapInt32(&h.unhealthy, compare, unhealthy)
-	return swapped, nil
+	return atomic.CompareAndSwapInt32(&u.unhealthy, compare, unhealthy)
 }
 
 // DialInfo contains information needed to dial a
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -143,7 +143,7 @@ func (u *Upstream) fillDialInfo(r *http.Request) (DialInfo, error) {
 	}, nil
 }
 
-func (u *Upstream) setHost() {
+func (u *Upstream) fillHost() {
 	host := new(Host)
 	existingHost, loaded := hosts.LoadOrStore(u.String(), host)
 	if loaded {
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -143,6 +143,15 @@ func (u *Upstream) fillDialInfo(r *http.Request) (DialInfo, error) {
 	}, nil
 }
 
+func (u *Upstream) setHost() {
+	host := new(Host)
+	existingHost, loaded := hosts.LoadOrStore(u.String(), host)
+	if loaded {
+		host = existingHost.(*Host)
+	}
+	u.Host = host
+}
+
 // Host is the basic, in-memory representation
 // of the state of a remote host.
 type Host struct {
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -46,6 +46,10 @@ type Upstream struct {
 	// backends is down. Also be aware of open proxy vulnerabilities.
 	Dial string `json:"dial,omitempty"`
 
+	// DEPRECATED: Use the SRVUpstreams module instead
+	// (http.reverse_proxy.upstreams.srv). This field will be
+	// removed in a future version of Caddy.
+	//
 	// If DNS SRV records are used for service discovery with this
 	// upstream, specify the DNS name for which to look up SRV
 	// records here, instead of specifying a dial address.
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -26,44 +26,13 @@ import (
 	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
 )
 
-// Host represents a remote host which can be proxied to.
-// Its methods must be safe for concurrent use.
-type Host interface {
-	// NumRequests returns the number of requests
-	// currently in process with the host.
-	NumRequests() int
-
-	// Fails returns the count of recent failures.
-	Fails() int
-
-	// Unhealthy returns true if the backend is unhealthy.
-	Unhealthy() bool
-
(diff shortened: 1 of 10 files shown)


modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -0,0 +1,243 @@
...
+		return cached.upstreams, nil
+	}
+
+	// prepare parameters and perform the SRV lookup
+	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
+	service := repl.ReplaceAll(su.Service, "")
+	proto := repl.ReplaceAll(su.Proto, "")
+	name := repl.ReplaceAll(su.Name, "")
+
+	_, records, err := net.DefaultResolver.LookupSRV(r.Context(), service, proto, name)
```
francislavoie, line 99: I think if someone wants to use the SRV results by priority, they'd have to use the `first` selection policy, and also make sure to have passive health checks enabled so that Caddy actually marks the top priority backends as unhealthy so it actually tries lower priorities.

Because otherwise, the default of `random` would just do whatever, which doesn't respect the SRV response.

Also, if we're caching the result, we need to keep in mind that we're also caching the result of the RNG that happened to randomize between the weights. So only one backend out of a given priority will get routed to while the result is cached. That seems like it might be a problem.

It is kinda nice that the stdlib does this for us, but I think their assumption is that these results don't get cached and instead this gets queried every time.

Should we have two implementations of `SRVUpstreams`? One which is cached, and _doesn't_ properly support weights, and another that's not cached, and _does_ properly support weights.

Correct me if I'm wrong here, I might've made an incorrect assumption.

modules/caddyhttp/reverseproxy/hosts.go
```diff
@@ -174,34 +143,42 @@ func (u *Upstream) fillDialInfo(r *http.Request) (DialInfo, error) {
 	}, nil
 }
 
-// upstreamHost is the basic, in-memory representation
-// of the state of a remote host. It implements the
-// Host interface.
-type upstreamHost struct {
+func (u *Upstream) setHost() {
```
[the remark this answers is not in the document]
mholt: How about `fillHost()`? (Kind of like with another function here.)

modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
mholt: Ohh interesting, a somewhat non-conventional use of error returns. Good catch.

modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -0,0 +1,243 @@
...
+
+	upstreams := make([]*Upstream, len(records))
+	for i, rec := range records {
+		upstreams[i] = &Upstream{
+			Dial: net.JoinHostPort(rec.Target, strconv.Itoa(int(rec.Port))),
+		}
+		upstreams[i].setHost()
+	}
+
+	// TODO: expire these somehow
```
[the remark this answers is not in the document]
mholt: That's one way. Or my personal favorite, a size limit with random eviction once we reach the max size. It's just so dang simple.

modules/caddyhttp/reverseproxy/hosts.go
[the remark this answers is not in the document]
francislavoie: Sounds good 👍

modules/caddyhttp/reverseproxy/upstreams.go
mholt: Yes, I think if you want SRV load balancing then using `first` policy with passive health checks is the right idea, and is the same way this would be done with static upstreams too.

You're right that the caching of results also caches the randomization of upstreams with equal priority, but keep in mind that only lasts for up to the `Refresh` duration. It's still randomized, just not per-request, and this is arguably more efficient too. @jjiang-stripe do you have a preference here? Do you need the randomization of backends within the same priority to be changed at every single request or is at every interval sufficient?

(If it does have to be changed for every request, we need to be mindful that we get upstreams every loop iteration, multiple times per request...)

francislavoie: Other users may want to re-query every time, so I suppose it would make sense to allow a `-1` Refresh to turn off caching?

modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -0,0 +1,243 @@
...
+
+type srvLookup struct {
+	srvUpstreams SRVUpstreams
+	freshness    time.Time
+	upstreams    []*Upstream
+}
+
+func (sl srvLookup) fresh() bool {
+	return time.Since(sl.freshness) < sl.srvUpstreams.Refresh
+}
```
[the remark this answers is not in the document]
mholt: Ok, changed locally. Thanks

modules/caddyhttp/reverseproxy/upstreams.go
mholt: Based on a call this morning, sounds like caching this randomization will not be a problem. It could actually be beneficial for RR load balancing. Will leave this as-is unless determined otherwise.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -408,8 +414,19 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyht
 
 	var proxyErr error
 	for {
+		// get the updated list of upstreams
+		upstreams := h.Upstreams
+		if h.DynamicUpstreams != nil {
+			dUpstreams, err := h.DynamicUpstreams.GetUpstreams(r)
+			if err != nil {
+				h.logger.Error("failed getting dynamic upstreams; falling back to static upstreams", zap.Error(err))
```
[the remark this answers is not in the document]
mholt: Yeah, one code/config path would be nice. Let me give this some more thought.

modules/caddyhttp/reverseproxy/upstreams.go
francislavoie: I think we need to consider other users too though, especially since this refactor it meant to resolve a bunch of open issues.

mholt: That's fine, but one thing at a time. Let's see if it really causes a problem for real use cases before we add complexity to a growing system.

francislavoie: I'm worried about this being a regression for the few users who are using SRV already. But fair enough.

josegonzalez:
I was sent here from #1630. One thing I was thinking of doing was adding support for a `docker://` scheme, wherein a custom plugin might resolve `docker://` - or whatever scheme it specified - to a configured docker host. So `docker://app.container` might resolve to the container with the name `app`, and `docker://app.service` might resolve to any container with the label `caddy.service=app`.

Would doing this be possible with the existing codebase or this MR? Should I open an issue or discuss this sort of work somewhere?

For what its worth, my use case is adding a simple integration for the [Dokku paas](https://github.com/dokku/dokku/), and avoiding the issue we have with nginx where docker restarts may result in container IPs changing.

francislavoie:
> So `docker://app.container` might resolve to the container with the name `app`

You can just proxy to `app` if Caddy is running in the same Docker network. Nothing special to do there, it'll use Docker's resolver.

> and `docker://app.service` might resolve to any container with the label `caddy.service=app`

This would require Caddy to actually talk to the Docker API, possibly via the docker unix socket. That's out of scope of this PR specifically, but it could possibly be implemented as a custom module which implements the dynamic upstreams interface. It wouldn't be done with a special scheme, but rather with a separate module (which is what this PR makes possible).

FYI, your usecase might be already covered by https://github.com/lucaslorentz/caddy-docker-proxy though, depending on your goals. It generates Caddyfile config dynamically based on the labels on the containers in your docker network.

francislavoie:
I added Caddyfile support real quick, a couple examples here in the tests: [`dc85b7b` (#4470)](https://github.com/caddyserver/caddy/pull/4470/commits/dc85b7b577bca47d33692e8155071ea7dba55cbd#diff-8582f920b0f93a9a5e490e75d0f2d172c654d6f006eb639f1ed8388192d66f7f)

mholt:
Wow, thanks Francis! :grin: Nice job for "real quick", heh.

modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -0,0 +1,279 @@
...
+// String returns the RFC 2782 representation of the SRV domain.
+func (su SRVUpstreams) String() string {
+	return fmt.Sprintf("_%s._%s.%s", su.Service, su.Proto, su.Name)
+}
+
+func (su *SRVUpstreams) Provision(ctx caddy.Context) error {
+	su.logger = ctx.Logger(su)
+	if su.Proto != "tcp" && su.Proto != "udp" {
+		return fmt.Errorf("invalid proto '%s'", su.Proto)
+	}
```
mohammed90, line 68: The doc on the SRVUpstreams struct says:
```
// or the domain can be specified directly in name by
// leaving service and proto empty.
```

But the implementation doesn't follow.

francislavoie: I also think that's needlessly restrictive, the implementation doesn't really care what the Proto is, users might use something like `_http` instead of `_tcp` and that should be fine too :man_shrugging:

francislavoie: @mohammed90 so actually `or the domain can be specified directly in name by leaving service and proto empty.` does work, that functionality is implemented in stdlib https://pkg.go.dev/net#Resolver.LookupSRV

I need to adjust how I'm doing the Caddyfile for this, I wrote this function to parse the `_<service>._<proto>` stuff out, but I need to switch it to not do that, it's not useful because just setting `name` inline should work just fine.

mholt: > But the implementation doesn't follow.

This is done by the standard library. :+1:

francislavoie: I'm also going to remove this condition here, the stdlib does not actually care what the proto is, the stdlib docs just "suggest" that `tcp` or `udp` are recommended but it's just used to assemble the name to be queried.

modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -265,6 +343,34 @@ func (al aLookup) isFresh() bool {
...
+	// with port range of only 1. If the host is an IP address, it will be dialed directly to resolve the upstream server.
+	// If the host is not an IP address, the addresses are resolved using the [name resolution convention](https://golang.org/pkg/net/#hdr-Name_Resolution) of the Go standard library.
+	// If the array contains more than 1 resolver address, one is chosen at random.
+	Addresses []string `json:"addresses,omitempty"`
+	netAddrs  []caddy.NetworkAddress
+}
+
+// ParseAddresses parses all the configured network addresses
+// and ensures they're ready to be used.
+func (u UpstreamResolver) ParseAddresses() error {
```
mohammed90, line 343: This should be a pointer receiver because you're modifying the receiver and expecting the same to reflect back to the caller

mholt: GOOD CATCH. Tricky tricky

francislavoie: Nice, thanks. I keep forgetting to think about that, it's such a subtle thing in Go 🤦‍♂️

mholt:
@jjiang-stripe, @cds2-stripe - this is ready for an initial review / trial whenever you're ready :+1:

cds2-stripe:
Thank you! We'll set aside some time next week to test this internally.

yroc92:
Could this support fetching the upstream from an HTTP request?

mholt:
@yroc92 Yes, any function that, given the request, can return a list of upstreams, is compatible.

And any Caddy modules can get access to Caddy's global storage configured here: https://caddyserver.com/docs/json/storage/

mholt:
Alright, I'm going to merge this in preparation for the pre-release of v2.5. If needed, we can tweak these features in a separate PR and continue discussion in another issue.

francislavoie (approved):
This is definitely an awesome step forwards.

I have a few concerns, but we talked off-github and I think we're on the same page that we can probably sort them out in followup PRs before 2.5 stable (this being in the beta). In particular, I think the [GET /reverse_proxy/upstreams](https://caddyserver.com/docs/api#get-reverse-proxyupstreams) API will be broken with these changes, and I think we should take the refactor a step further with making static upstreams an `UpstreamSource` as well so we only have a single codepath in the main proxy code (instead of it being split depending on whether static or dynamic is configured).

xeruf:
Why no aaaa source that restricts it to just IPv6? Would be handy for my NAT

francislavoie:
@xeruf you'll need to elaborate, your question is severely lacking detail.

A merged PR is not the right place for further discussion. If you have a feature request, open a new issue. If you have a usage question, open a topic on the forums https://caddy.community.

mholt:
@xeruf I would be happy to hear the use case for that in a new issue!

PR #4473, opened 2021-12-09
go.mod: Update smallstep/truststore, fix build on FreeBSD

francislavoie:
See https://github.com/smallstep/truststore/pull/7#issuecomment-988544551 and https://github.com/smallstep/truststore/pull/9


mholt (approved):
Thanks!

PR #4477, opened 2021-12-12
httpcaddyfile: Fix sorting edgecase for nested `handle_path`

francislavoie:
Fixes a bug reported in https://caddy.community/t/handle-path-being-ignored-when-paired-with-handle-in-same-block/14393

We had some code in `ServerType.Setup()` that would ensure `handle_path` has the same sort level as `handle`, but we were missing the same check in `parseSegmentAsConfig` to make sure that works properly inside of any nested subroute (`handle`, `route`, `handle_errors`, `handle_response`).

To be clear, the fix is copy-paste from another spot where we do the same thing:

https://github.com/caddyserver/caddy/blob/78b5356f2b1945a90de1ef7f2c7669d82098edbd/caddyconfig/httpcaddyfile/httptype.go#L196-L202

The new test fails before the change (the `handle_path` ends up after the `handle` in the same level, and they're missing a `group` name because the group is omitted if there's no two handlers with the same level) and passes after.

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -340,6 +340,9 @@ func parseSegmentAsConfig(h Helper) ([]ConfigValue, error) {
 			if err != nil {
 				return nil, h.Errf("parsing caddyfile tokens for '%s': %v", dir, err)
 			}
+
+			dir = normalizeDirectiveName(dir)
+
 			for _, result := range results {
 				result.directive = dir
 				allResults = append(allResults, result)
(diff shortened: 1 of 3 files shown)


adityaxdiwakar:
OP from the community thread, thank you very much @francislavoie. You guys are fast! Glad I switched to Caddy.

mholt (changes requested):
Thanks for the fix. Hmm, unfortunate that we need to repeat that logic. I wonder if there's any way to combine the other function where this is from with this logic. :thinking:  (Maybe for another PR, if so.)

Can we move the change any lower? The only next use of `dir` is on L352, other than in an error message earlier on L349. It'd be helpful if the directive in the error message is the same as appears in the config. So I wonder if we move this change to between, say, L350-351 it might be a little more helpful.

francislavoie:
Yeah we should extract this to a method to do these kinds of normalizations. I'll move it.

francislavoie:
Updated

mholt (approved):
That looks better. Thanks!

PR #4482, opened 2021-12-16
reverseproxy, admin: memory leak fix

dtelyukh:
See good explanation: https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -792,10 +792,15 @@ func (lb LoadBalancing) tryAgain(ctx caddy.Context, start time.Time, proxyErr er
 	}
 
 	// otherwise, wait and try the next available host
+	timer := time.NewTimer(time.Duration(lb.TryInterval))
 	select {
-	case <-time.After(time.Duration(lb.TryInterval)):
+	case <-timer.C:
 		return true
 	case <-ctx.Done():
+		if !timer.Stop() {
+			// if the timer has been stopped then read from the channel
+			<-timer.C
+		}
 		return false
 	}
 }
--- a/caddy.go
+++ b/caddy.go
@@ -494,9 +494,10 @@ func finishSettingUp(ctx Context, cfg *Config) error {
 		if cfg.Admin.Config.LoadInterval > 0 {
 			go func() {
 				for {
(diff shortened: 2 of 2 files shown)


mholt:
Hey thanks for this. I will look at this after the holidays!

pugg11:
> 

No worries

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -792,10 +792,15 @@ func (lb LoadBalancing) tryAgain(ctx caddy.Context, start time.Time, proxyErr er
...
 	// otherwise, wait and try the next available host
+	timer := time.NewTimer(time.Duration(lb.TryInterval))
 	select {
-	case <-time.After(time.Duration(lb.TryInterval)):
+	case <-timer.C:
 		return true
 	case <-ctx.Done():
+		if !timer.Stop() {
+			// if the timer has been stopped then read from the channel
+			<-timer.C
```
vishal1132, line 802: Do you think it might be a good idea to drain the timer in it's own goroutine if the `timer.Stop()` call doesn't stop the timer. For example something like- 
```
go func(){
	<-timer.C
}()
```
Also the comment on top of it is not very useful I think. Also, thanks TIL about this memory leak.

mholt: I don't think so; see the docs: https://pkg.go.dev/time#Timer.Stop

vishal1132: @mholt If the timer has already expired and `time.Stop()` wouldn't return true and `<-timer.C` would ensure a value *atleast* after the duration, but it can be very long too. Draining this in a goroutine would ensure the normal flow of the code from this point of view and draining elsewhere. Please let me know if I am missing something and sorry for being noob in that case.

mholt: No, this only reads from the channel if the timer has already been stopped or expired. So the read will never be blocking in this case.

mholt (approved):
Thank you for this contribution -- I suppose I was just lazy, and perhaps this will save some users a little bit of trouble and keep our allocations tidy.

PR #4499, opened 2021-12-30
caddyhttp: Fix `MatchPath` sanitizing

francislavoie:
This is a followup to #4407, in response to a report on the forums: https://caddy.community/t/php-fastcgi-phishing-redirection/14542

Turns out that doing `TrimRight` to remove trailing dots, _before_ cleaning the path, will cause double-dots at the end of the path to not be cleaned away as they should. We should instead remove the dots _after_ cleaning.


mholt (approved):
Ah, nice. Simple and elegant fix, LGTM. Thank you

PR #4506, opened 2022-01-04
headers: Fix `+` in Caddyfile to properly append rather than set

francislavoie:
See https://caddy.community/t/only-one-header-link-applied/14571

In a single `header` handler, using `+` with the same field name multiple times would only use the last value, instead of adding each value.

Super simple fix, just need to use `.Add()` instead of `.Set()` if we're using the `+` operator.


mholt (approved):
Thanks, I guess that makes sense. :smile:

PR #4513, opened 2022-01-09
caddytls: Add internal Caddyfile `lifetime`, `sign_with_root` opts

francislavoie:
Options missing from Caddyfile which are configurable in JSON.

While these options are strongly recommended to not change, better to have the Caddyfile have them then not.


mholt (approved):
Another PR by Francis that LGTM. Thanks for knocking all these out recently. :100:

PR #4514, opened 2022-01-09
httpcaddyfile: Add pki app `root` and `intermediate` cert/key config

francislavoie:
Followup to #4450, I was being lazy. Adds the cert/key config.


mholt (approved):
Nice, thank you for this! Really appreciate how you're so invested into enhancing the Caddyfile for all of us.

PR #4522, opened 2022-01-13
caddytls: Fix `MatchRemoteIP` provisoning with multiple CIDR ranges

GallopingKylin:


francislavoie:
What does this fix, exactly? Please show an example config that was broken without this change.

GallopingKylin:
caddy version: v2.4.6

json config like this:
```
{
  "apps": {
    "http": {
      "servers": {
        "": {
          "listen": [
            ":443"
          ],
          "tls_connection_policies": [
            {
              "match": {
                "remote_ip": {
                  "ranges": [
                    "1.1.1.1/32",
                    "8.8.8.8/32"
                  ]
                }
              }
            }
          ]
        }
      }
    }
  }
}

```
Above is a simple json config which is used to allow authorized source IP addresses to establish TLS connections, assuming 1.1.1.1 is the allowed IP, caddy works fine when the ranges field contains only itself or when other IPs are inserted above 1.1.1.1 (e.g. 8.8.8.8), but when inserted below 1.1.1.1, caddy does not work properly, and it outputs the following error log:

```
{
    "level": "debug", 
    "ts": 1642083138.2123873, 
    "logger": "http.stdlib", 
    "msg": "http: TLS handshake error from xxx.xxx.xxx.xxx:xxx: no server TLS configuration available for ClientHello: &
{CipherSuites:[4866 4867 4865 49196 49200 159 52393 52392 52394 49195 49199 158 49188 49192 107 49187 49191 
103 49162 49172 57 49161 49171 51 157 156 61 60 53 47 255] ServerName:xxx.com SupportedCurves:[X25519 CurveP256 
CurveID(30) CurveP521 CurveP384] SupportedPoints:[0 1 2] SignatureSchemes:[ECDSAWithP256AndSHA256 
ECDSAWithP384AndSHA384 ECDSAWithP521AndSHA512 Ed25519 SignatureScheme(2056) SignatureScheme(2057) 
SignatureScheme(2058) SignatureScheme(2059) PSSWithSHA256 PSSWithSHA384 PSSWithSHA512 PKCS1WithSHA256 
PKCS1WithSHA384 PKCS1WithSHA512 SignatureScheme(771) ECDSAWithSHA1 SignatureScheme(769) PKCS1WithSHA1 
SignatureScheme(770) SignatureScheme(514) SignatureScheme(1026) SignatureScheme(1282) SignatureScheme(1538)] 
SupportedProtos:[h2 http/1.1] SupportedVersions:[772 771 770 769] Conn:0xc0000115e8 config:0xc000344780 
ctx:0xc000078f00}"
}

```

francislavoie (approved):
Ah, yeah that makes sense now. Thanks! Good find.

francislavoie (changes requested):
Looks like the linter is unhappy, actually:

> `/home/runner/work/caddy/caddy/modules/caddytls/matchers.go:89:29: cannot use cidrs (variable of type []*net.IPNet) as *net.IPNet value in argument to append`

I think you need to use the spread operator for these, since they're already slices

GallopingKylin (commented):
unpack slice...

mholt:
Thanks for finding and fixing this!

PR #4528, opened 2022-01-17
rewrite: Add `method` Caddyfile directive

### LINKED ISSUE #4523: Feature Request: Add method subdirective to caddy file
As stated [here](https://caddy.community/t/v2-rewrite-get-as-post-request/7252) it looks like the subdirective `method` is not yet available in Caddy v2 to be used in the Caddyfile.
Any plans on implementing it?

francislavoie:
Fixes #4523

Pretty simple. The order of this directive is up for debate though. Just went with what feels intuitive to me, but it's hard to say for sure how this will be used in combination with other directives.

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -45,7 +45,8 @@ var directiveOrder = []string{
 
 	"redir",
 
-	// URI manipulation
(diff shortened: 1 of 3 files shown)


mholt (approved):
LGTM -- this is exactly how I would have done it I think. Thanks!

PR #4530, opened 2022-01-18
caddyhttp: Fix HTTP->HTTPS redir not preferring HTTPS port if ambiguous

### LINKED ISSUE #4529: Auto HTTPS redirects to the wrong entry
### 1. Caddy version (caddy version):
2.4.6

### 2. How I run Caddy:

Via docker compose on Amazon ECS.

#### a. System environment:

Amazon ECS, Fargate. Docker compose file with docker client using ECS context.

#### b. Command:

Caddyfile is base64 encoded in an environment variable. Then is decoded and copied to `/etc/caddy/Caddyfile`. Then `caddy run --config /etc/caddy/Caddyfile --adapter caddyfile`.

```yaml
services:
  caddy:
    image: caddy:2.4.6-alpine
    ports:
      - "80:80"
      - "443:443"
      - "2345:2345"
    volumes:
      - caddy_data:/data
      - caddy_config:/config
    command: ["sh", "-c", "echo ${CADDYFILE_B64} | base64 -d > /etc/caddy/Caddyfile && caddy run --config /etc/caddy/Caddyfile --adapter caddyfile"]
volumes:
  caddy_data:
  caddy_config:
```

```
{
    debug
}

https://example.example.com {
    reverse_proxy http://localhost:1234
}

https://example.example.com:2345 {
    reverse_proxy http://localhost:2345
}
```

### 3. The problem I’m having:

When we do go to http://example.example.com/, we get redirected to https://example.example.com:2345/ instead of https://example.example.com/.

```
% curl -IL http://example.example.com                                                                                          
HTTP/1.1 308 Permanent Redirect
Connection: close
Location: https://example.example.com:2345/
Server: Caddy
Date: Wed, 12 Jan 2022 14:57:00 GMT

HTTP/2 200 
content-type: application/octet-stream
date: Wed, 12 Jan 2022 14:57:00 GMT
server: Caddy
content-length: 0
```

### 4. What I already tried:

Tried changing the order of the entries.

Created this issue as suggested in the [forum post ](https://caddy.community/t/entry-with-no-port-specified-redirects-to-wrong-port-entry/14708)

francislavoie:
Fix #4529 

I think it doesn't make sense to append to this slice if we actually want to prefer one port over the other. We don't need to keep a redirect for more than one port at all, I think, because we have no way to distinguish which port is the right target otherwise, when requests come in on the HTTP port.

I think the comment here was correct, but the behaviour was wrong.

This should fix it, by always preferring the HTTPS port -- if `443` is seen first then it will set it (because no value is set), then it will ignore any subsequent cause they aren't the HTTPS port. If something else comes first, and `443` follows, then it will overwrite it.

### DIFF
--- a/modules/caddyhttp/autohttps.go
+++ b/modules/caddyhttp/autohttps.go
@@ -232,7 +232,7 @@ func (app *App) automaticHTTPSPhase1(ctx caddy.Context, repl *caddy.Replacer) er
 				// port, we'll have to choose one, so prefer the HTTPS port
 				if _, ok := redirDomains[d]; !ok ||
 					addr.StartPort == uint(app.httpsPort()) {
-					redirDomains[d] = append(redirDomains[d], addr)
+					redirDomains[d] = []caddy.NetworkAddress{addr}
 				}
 			}
 		}


mholt (approved):
Thanks Francis. I've been reviewing this file a few times and I think this makes sense, though I have to admit it still hurts my head to process it. :sweat_smile: I'll happily merge this for now since I don't see a reason we need to append here, but be prepared for possible bug reports / edge cases that arise later.

francislavoie:
Yeah, I'm not 100% confident on this one either, but it seems like it's probably correct.

FWIW I wish we had a way to "adapt" the auto-https routes/etc so we could write test fixtures for that as well. Is there a way we could dump the config to JSON _after_ provisioning?

mholt:
True; I think a better way to test this would be unit tests (which might involve compartmentalizing more of the logic into individual functions, which I'd probably be OK with if it's not too convoluted that way).

jacobmischka:
Thanks for the quick fix! Is there any workaround we can use in the meantime before the next release that anyone is aware of (besides not reverse proxying multiple ports)? Or do we need to build from source until then?

mholt:
I'll probably be tagging a pre release in the next week or two!

PR #4532, opened 2022-01-18
httpcaddyfile: Fix incorrect handling of IPv6 bind addresses

francislavoie:
The `net.JoinHostPort()` function has some naive logic for handling IPv6, it just checks if the host part has a `:` and if so it wraps the host part with `[ ]` but this causes our network type prefix to get wrapped as well, which is invalid for `caddy.NetworkAddress`. Instead, we can just concatenate the host and port manually here to avoid this side-effect.


mholt (approved):
Thanks; guess I didn't think about that! LGTM

PR #4534, opened 2022-01-19
explain cryptic unix socket listener error related to process kill

ForestJohnson:
previous PR: https://github.com/caddyserver/caddy/pull/4533

here is a simple demo of the error being caught and the error message being displayed: https://go.dev/play/p/ehz0pTPjMwZ


mholt (changes requested):
Well, this is quite a bit more intense than what I had in mind. The information in this error is worded more like a documentation page. Instead, the error should be short and concise, like "always exit Caddy gracefully," or "this can happen if Caddy is forcefully killed." And we can just use fmt.Errorf with %w to modify the error message.

mholt (approved):
This looks good, thank you!

PR #4541, opened 2022-01-22
caddytls: Support custom GetCertificate modules (like Tailscale)

mholt:
Sometimes there are external tools or services that are managing certificates where we just need to call `GetCertificate()` during a handshake. The implementer of that function takes care of caching, renewal logic, etc. This adds support for `CertificateGetter` modules, in the `tls.get_certificate` namespace.

These differ from Issuer modules in that `Issue()` takes a CSR, whereas `GetCertificate()` takes a `*tls.ClientHelloInfo`. Also, using an Issuer implies that Caddy/CertMagic also generate and manage the private key, AND store the certificate and key in storage, since we're the ones maintaining the cert. In the case of GetCertificate, we don't want to be bothered about private keys or persisting in storage; we just want a cert. So basically, we use Issuers when we claim that we're managing the certificate. And we use `CustomGetCertificate` only to get a certificate from something we know is managing it for us.

The return value of our `GetCertificate()` is optional. If it returns nil+nil, it is passed by as if it was not specified at all; thus could fall back to traditional Issuer management for a name.

Specifying a certificate getter module implicitly enables On-Demand TLS for that policy (as if `on_demand: true` was set) because certificates given by `GetCertificate()` have to be called during a handshake.

Our `GetCertificate()` signature is slightly different from `tls.Config.GetCertificate` in that it adds a `bool` middle return value, which if true, tells CertMagic to add the certificate to its cache. Once cached, the module's `GetCertificate()` will only be called again once the cert nears expiration. Modules should return `false` for this value if they want to be called for every single TLS handshake.

This initial implementation integrates natively with [Tailscale](https://tailscale.com/blog/tls-certs/), although it requires being run as root or as a regular user that has access to the Tailscale socket. I'm still learning about this.

Here's an example automation policy for using Tailscale certs:

```json
{
	"subjects": ["yourhost.your-alias.ts.net"],
	"get_certificate": { "via": "tailscale" }
}
```

Or as a Caddyfile:

```
tls {
    get_certificate tailscale
}
```

A follow-up commit may also add support for an HTTP getter, to download the cert over HTTP.

This feature will remain experimental (and subject to change) even after being released, as we learn about it from experience.

Will be adding Caddyfile support soon, too.

Huge thanks to Tailscale for making this feature possible! Excited to release this and share more about it. /cc @bradfitz 

TODO:
- [x] Figure out why go.sum blew up (why is smallstep adding kms now, on this branch? :thinking: )

Builds on https://github.com/caddyserver/certmagic/pull/163


bradfitz:
> Figure out why go.sum blew up

Because of the dep to the `tailscale.com` module. That's why I'd said you might want us to pull that out separately to something smaller. Some people are offended by `go.sum` spam, even though it's rarely an actual problem.

francislavoie:
FYI, linter failure:

```
  level=warning msg="
[runner] Can't run linter goanalysis_metalinter: bodyclose: failed prerequisites: 
[buildssa@github.com/caddyserver/caddy/v2/modules/caddytls: analysis skipped: errors in package:
[/home/runner/work/caddy/caddy/modules/caddytls/certgetters.go:7:2: could not import tailscale.com/client/tailscale (
/home/runner/go/pkg/mod/tailscale.com@v1.20.1/client/tailscale/tailscale.go:28:2: could not import tailscale.com/client/tailscale/apitype (
/home/runner/go/pkg/mod/tailscale.com@v1.20.1/client/tailscale/apitype/apitype.go:8:8: could not import tailscale.com/tailcfg (
/home/runner/go/pkg/mod/tailscale.com@v1.20.1/tailcfg/tailcfg.go:19:2: could not import tailscale.com/types/key (
/home/runner/go/pkg/mod/tailscale.com@v1.20.1/types/key/disco.go:177:23: cannot convert ciphertext (variable of type []byte) to *[24]byte)
)))]]"
```

modules/caddytls/automation.go
```diff
@@ -113,7 +121,7 @@ type AutomationPolicy struct {
 
 	// If true, certificates will be managed "on demand"; that is, during
 	// TLS handshakes or when needed, as opposed to at startup or config
-	// load.
+	// load. This enable On-Demand TLS for this policy.
```
francislavoie, line 124:
```suggestion
	// load. This enables On-Demand TLS for this policy.
```

mholt:
Ah, all the tests for Go 1.16 are failing; I wonder if Go 1.17 is required.

This is pretty much done now, except for a few things:

- Exact behavior of Tailscale's GetCertificate if it doesn't have a cert for the subject
- Tailscale socket permissions
- Finishing the HTTP cert getter: what should the response body be?
- Dependency bloat

modules/caddytls/certgetters.go
```diff
@@ -0,0 +1,169 @@
...
+	// The URL from which to download the certificate. Required.
+	//
+	// The URL will be augmented with query string parameters taken
+	// from the TLS handshake:
+	//
+	// - server_name: The SNI value
+	// - signature_schemes: Comma-separated list of hex IDs of signatures
+	// - cipher_suites: Comma-separated list of hex IDs of cipher suites
+	//
+	// To be valid, the response must be HTTP 200 with ... TODO: body format?
```
francislavoie, line 83: Do we need it to return the full chain? Maybe requiring JSON is best, more flexible ultimately. If someone else wants another format they can make a plugin 🤷‍♂️

mholt: Yeah, the whole `tls.Certificate` struct, including private key (yikes). I was thinking either JSON or multipart form. Leaning towards JSON though.

francislavoie: Could add have an `accepts` option to change what data type Caddy expects to get back, but only if someone needs it/requests support for it.

bradfitz:
> Ah, all the tests for Go 1.16 are failing; I wonder if Go 1.17 is required.

Tailscale only supports the most recent version of Go. So, yes, ...

```
Error: /home/runner/go/pkg/mod/tailscale.com@v1.20.1/types/key/disco.go:177:22: cannot convert ciphertext (type []byte) to type *[24]byte
Error: /home/runner/go/pkg/mod/tailscale.com@v1.20.1/types/key/node.go:137:22: cannot convert ciphertext (type []byte) to type *[24]byte
```

We assume we can do that conversion now.

If I made a separate Go module/package for fetching certs, I can make sure it works with older Go versions.

francislavoie:
FWIW, we generally go with "current - 1" for Go versions for Caddy, mainly because our RPM packaging maintainers need to rely on the Go version they have available to them.

Thanks for splitting it out @bradfitz :grinning:

bradfitz:
And I've pushed https://github.com/tailscale/tailscale/pull/3809 to add support for a `TS_PERMIT_CERT_UID` env variable that users can add to `/etc/default/tailscaled` to grad the `caddy` user's UID access to fetch certs.

bradfitz:
Tailscale client code is now split out into https://github.com/tailscale/tscert (pkg.go.dev hasn't yet loaded it yet, so there's no godoc easily readable quite yet)

Tested back to Go 1.15, but it probably supports earlier.

mholt:
Brilliant! Thanks Brad.

I pushed a commit which adds context to the GetCertificate call and checks the Status to see if the domain in the handshake is in the list. If it's not a cert provisioned by Tailscale, false+nil is returned to CertMagic so that CertMagic knows to continue with its own management logic. I also changed to using the `tscert` package (oops, I said `tsclient` in my commit message; will try to remember to fix when squashing) which is much lighter. :100:

We'll see how the tests do now.

I'll probably figure out some simple response body for the HTTP getter just as a cherry on top, though I don't think we'll need it at this time.

mholt:
You know, it just occurred to me, that we could probably make this a default, implicit behavior. If Tailscale is running AND happens to have a certificate for the hostname, just use it. If not, then continue with our own cert.

Returning `nil, false, nil` from GetCertificate() causes CertMagic to carry on normally. So it already supports that.

If we can guarantee the latency of checking the Tailscale status is very minimal (maybe we cache the results for a certain amount of time? or set a context with a timeout of ~5ms?), maybe this is safe to do. We could even restrict any of this implicit behavior to activate only on names ending in `.ts.net`.

Of course, if `get_certificate tailscale` was _explicitly_ configured, we could ease some of these restrictions and actually return errors when errors happen (like the current implementation), since if it's explicitly configured then we _expect_ it to work.

I'm just thinking, maybe it doesn't hurt to _try_ Tailscale implicitly even if not configured, then it really will "just work". Caddy +Tailscale, ladies and gentlemen. :man_in_tuxedo:

bradfitz:
> We could even restrict any of this implicit behavior to activate only on names ending in `.ts.net.`

That's a safe heuristic. SGTM. We have no plans to offer other domain name suffixes. But we may permit people to BYODomain in the future, so if you make it automatic on `*.ts.net`, just let people explicitly enable it as well. Also, https://github.com/juanfont/headscale will likely add such support in the future and each user using that would necessarily  have their own domain name as well.

mholt:
The HTTP cert getter should be working now. It expects PEM blocks and it will decode them into a TLS certificate and private key pair.

I think we're ready to merge this, and I'll try to implement implicit Tailscale cert-getting as a cherry on top in another PR, if I can make it work. (It wasn't a requirement, I just think it'd be cool.)

francislavoie (approved):

mholt:
What happened to the CI checks...

BioEvo:
> The HTTP cert getter should be working now. It expects PEM blocks and it will decode them into a TLS certificate and private key pair.
> 
> I think we're ready to merge this, and I'll try to implement implicit Tailscale cert-getting as a cherry on top in another PR, if I can make it work. (It wasn't a requirement, I just think it'd be cool.)

It seems like HTTP cert gatter is ignored on caddy v2.6.4. I set the get_certificate section in caddyfile, caddy verified all parameters and directly goes to HTTP-01 challenge. Log of cert.mysite shows no http request from web.mysite.

`https://web.mysite {
    tls {
        get_certificate http http://cert.mysite/cert
    }
    ...
}`

mholt:
@BioEvo Could you open a new issue please? Be sure to enable debug logging and include all the logs. (And please don't redact the domain names, we'll need the actual values to troubleshoot.) Also the last Caddy version it does work on. The more specific you are, the more specific we can be. Thank you!

BioEvo:
> @BioEvo Could you open a new issue please? Be sure to enable debug logging and include all the logs. (And please don't redact the domain names, we'll need the actual values to troubleshoot.) Also the last Caddy version it does work on. The more specific you are, the more specific we can be. Thank you!

Thank you! I try to reproduce this in docker, I'm not sure how caddy was openration with HTTP CertManager, but if the config is like:
```
https://web.mysite {
    tls {
        get_certificate http http://cert.mysite/cert
    }
    ...
}
```
then i think caddy might try to request http://cert.mysite/cert?server_name=web.mysite&signature_schemes=...&cipher_suites=...
then certmgr might responds with something like below in 200:
```
-----BEGIN CERTIFICATE-----
[SOME_BASE64]
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
[SOME_BASE64]
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
[SOME_BASE64]
-----END CERTIFICATE-----
-----BEGIN EC PRIVATE KEY-----
[SOME_BASE64]
-----END EC PRIVATE KEY-----
```
https://github.com/caddyserver/caddy/issues/5415#issue-1608319601

PR #4544, opened 2022-01-22
caddyhttp: Fix test when /tmp/etc already exists

Ikke:
The TestFileListing test in tplcontext_test has one test that verifies
if directory traversal is not happening. The context root is set to
'/tmp' and then it tries to open '../../../../../etc', which gets
normalized to '/tmp/etc'.

The test then expects an error to be returned, assuming that '/tmp/etc'
does not exist on the system. When it does exist, it results in a test
failure:

```
--- FAIL: TestFileListing (0.00s)
    tplcontext_test.go:422: Test 4: Expected error but had none
    FAIL
    FAIL
    github.com/caddyserver/caddy/v2/modules/caddyhttp/templates	0.042s
```

Instead of using '/tmp' as root, use a dedicated directory created with
`os.MkdirTemp()` instead. That way, we know that the directory is empty.

### DIFF
--- a/modules/caddyhttp/templates/tplcontext_test.go
+++ b/modules/caddyhttp/templates/tplcontext_test.go
@@ -608,6 +608,9 @@ title = "Welcome"
 
 func getContextOrFail(t *testing.T) TemplateContext {
 	tplContext, err := initTestContext()
+	t.Cleanup(func() {
+		os.RemoveAll(string(tplContext.Root.(http.Dir)))
+	})
 	if err != nil {
 		t.Fatalf("failed to prepare test context: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thank you! Never seen that happen before.

Should we be deleting the temporary directory when we're done?

> It is the caller's responsibility to remove the directory when it is no longer needed.

Ikke:
> Thank you! Never seen that happen before.
> 
> Should we be deleting the temporary directory when we're done?
> 
> > It is the caller's responsibility to remove the directory when it is no longer needed.

Probably, but I wonder what a good place would be to do this cleanup, as this is used in a number of tests.

mholt:
@Ikke Hm, yeah, that's true; so either we'd need to return the folder path and the caller has to do `defer os.Remove()`, or a Cleanup() function sounds good to me as long as it doesn't break anything (or leave anything lingering).

Ikke:
> @Ikke Hm, yeah, that's true; so either we'd need to return the folder path and the caller has to do `defer os.Remove()`, or a Cleanup() function sounds good to me as long as it doesn't break anything (or leave anything lingering).

Added the cleanup now.

mholt:
Thanks! I'm mobile for a bit; did you verify the files get cleaned up then, as expected?

Ikke:
Yes, removed all left behind directories from the previous run, ran the test suite, and `ls /tmp/caddy*` returns nothing.

mholt:
Perfect, thanks!

PR #4545, opened 2022-01-23
metrics: Enforce smaller set of method labels

hairyhenderson:
Protects against unbounded cardinality on the method label.

Note: I'm finding myself copying code again between the two metrics endpoints - how do we feel about `internal/`? I ask because there's no use of it yet... Otherwise I think I'd rather throw the common code in an `internal/metrics` package or something like that...

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -1,0 +1,39 @@
+package metrics
+
+import (
+	"net/http"
+	"strconv"
+)
+
+func SanitizeCode(s int) string {
+	switch s {
+	case 0, 200:
+		return "200"
+	default:
+		return strconv.Itoa(s)
+	}
+}
+
+// Only support the list of "regular" HTTP methods, see
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+var methodMap = map[string]string{
+	"GET": http.MethodGet, "get": http.MethodGet,
+	"HEAD": http.MethodHead, "head": http.MethodHead,
+	"PUT": http.MethodPut, "put": http.MethodPut,
+	"POST": http.MethodPost, "post": http.MethodPost,
+	"DELETE": http.MethodDelete, "delete": http.MethodDelete,
+	"CONNECT": http.MethodConnect, "connect": http.MethodConnect,
+	"OPTIONS": http.MethodOptions, "options": http.MethodOptions,
+	"TRACE": http.MethodTrace, "trace": http.MethodTrace,
+	"PATCH": http.MethodPatch, "patch": http.MethodPatch,
+}
+
+// SanitizeMethod sanitizes the method for use as a metric label. This helps
+// prevent high cardinality on the method label. The name is always upper case.
+func SanitizeMethod(m string) string {
+	if m, ok := methodMap[m]; ok {
+		return m
+	}
+
+	return "OTHER"
+}
--- a/modules/caddyhttp/metrics.go
+++ b/modules/caddyhttp/metrics.go
@@ -180,7 +180,7 @@ func sanitizeMethod(m string) string {
 		return m
 	}
 
-	return "other"
+	return "OTHER"
 }
 
 // taken from https://github.com/prometheus/client_golang/blob/6007b2b5cae01203111de55f753e76d8dac1f529/prometheus/promhttp/instrument_server.go#L298
--- a/modules/caddyhttp/metrics.go
+++ b/modules/caddyhttp/metrics.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"net/http"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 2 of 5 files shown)


francislavoie (approved):

modules/caddyhttp/metrics.go
```diff
@@ -160,6 +159,30 @@ func sanitizeCode(code int) string {
...
+}
+
+// sanitizeMethod sanitizes the method for use as a metric label. This helps
+// prevent high cardinality on the method label. The name is always upper case.
+func sanitizeMethod(m string) string {
+	if m, ok := methodMap[m]; ok {
+		return m
+	}
+
+	return "other"
```
mholt, line 183: "The name is always upper case" :raised_eyebrow:

modules/caddyhttp/metrics.go
```diff
@@ -160,6 +159,30 @@ func sanitizeCode(code int) string {
 	return strconv.Itoa(code)
 }
 
+// Only support the list of "regular" HTTP methods, see
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+var methodMap = map[string]string{
```
mholt, line 164: Can we possibly not repeat this in the other metrics.go file as well? I guess that might involve exporting something but maybe that's OK... (kind of a bummer that the core caddy package uses it...)

mholt (changes requested):
Awesome, thanks Dave! Just a couple things

modules/caddyhttp/metrics.go
hairyhenderson: @mholt that's why I asked about your feelings on `internal/` in the description 😉

modules/caddyhttp/metrics.go
hairyhenderson: 😅

modules/caddyhttp/metrics.go
mholt: Ahhh wow totally missed that, sorry. I'll circle back to this shortly!

modules/caddyhttp/metrics_test.go
```diff
@@ -82,3 +83,25 @@ type middlewareHandlerFunc func(http.ResponseWriter, *http.Request, Handler) err
...
+		method   string
+		expected string
+	}{
+		{method: "get", expected: "GET"},
+		{method: "POST", expected: "POST"},
+		{method: "OPTIONS", expected: "OPTIONS"},
+		{method: "connect", expected: "CONNECT"},
+		{method: "trace", expected: "TRACE"},
+		{method: "UNKNOWN", expected: "other"},
+		{method: strings.Repeat("ohno", 9999), expected: "other"},
```
francislavoie, line 98: Test fix

```suggestion
		{method: "UNKNOWN", expected: "OTHER"},
		{method: strings.Repeat("ohno", 9999), expected: "OTHER"},
```

modules/caddyhttp/metrics.go
mholt: @hairyhenderson I'm curious what the implementation using `internal/` would look like. I think we should definitely be careful/limit what we put in there, but I think the method map is definitely one of those obscure things that doesn't make sense to truly export for other modules to use. :+1:

hairyhenderson: @mholt ok, I'll hack something up and update this PR

hairyhenderson:
@mholt I've merged the common sanitize functions into an `internal` package - PTAL!

internal/metrics/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+		return "200"
+	default:
+		return strconv.Itoa(s)
+	}
+}
+
+// Only support the list of "regular" HTTP methods, see
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+var methodMap = map[string]string{
+	"GET": http.MethodGet, "get": http.MethodGet,
```
mholt, line 20: What if a method comes in as "Get"? (This whole patch is motivated by adversarial inputs, after all.)

I kinda liked the `strings.ToUpper()` as was done previously, since it covers all case differences, and you don't need to repeat each method for different cases.

hairyhenderson: > What if a method comes in as "Get"? (This whole patch is motivated by adversarial inputs, after all.)
> 
> I kinda liked the `strings.ToUpper()` as was done previously, since it covers all case differences, and you don't need to repeat each method for different cases.

It's a fair question, but would be _very_ unusual. This code was inspired by the Prometheus client code (https://github.com/prometheus/client_golang/blob/9075cdf61646b5adf54d3ba77a0e4f6c65cb4fd7/prometheus/promhttp/instrument_server.go#L377), which assumes the same cases (but outputs lowercase instead). A map lookup is also faster than a call to `strings.ToUpper`, which is why I chose to do that instead.

Personally I have no issue using the label `"OTHER"` when the input method is abnormal. I've used the `prometheus/client_golang` code for years, and have never seen legitimate clients using different method casings.

mholt: Ok; up to you, I just think repeating the method name in the map is a bit odd, and it seems unfortunate to lose the robustness of just uppercasing the method (which is very, very cheap -- utter nanoseconds I bet).

mholt (approved):
Still don't _love_ the method map with repeated methods, but has my approval for merge either way. Thanks!

mholt:
Nice work Dave, thank you so much for helping with this!

Issue #4466, opened 2021-12-07
`file_server` doesn't properly canonicalize when inside `handle_path`


francislavoie:
For example, given this config:

```
:8880 {
	handle_path /strip* {
		root * .
		file_server browse
	}
}
```

Making a request to `http://localhost:8880/strip` does not trigger a canonicalization redirect to `/strip/`, so the file browser ends up with broken relative links while on the `/strip` page (clicking on a file will go to `/file.txt` instead of `/strip/file.txt` as it should).

mholt:
Wouldn't the correct config be `handle_path /strip/* {` ?

francislavoie:
Not necessarily, you might want to be able to go to `/strip` and have it perform `file_server`'s built-in canonicalization redir. Obviously the workaround is to use `/strip/*` and add your own `redir /strip /strip/` but that's unintuitive and unfortunate.

mholt:
@francislavoie Tbh that's what I'd do, I just don't know of an elegant solution otherwise. I do think a manual redirect is not a bad solution, personally. :man_shrugging:

francislavoie:
We _might_ be able to look at the path and if it's literally `""` (empty string) then we could assume a rewrite happened to strip the prefix and we should redirect to `{orig_path}/`. I'll have to play around with it. Not 100% certain that's correct though but it might fix some cases.

mholt:
I don't know if that would work (rewriting to an empty path?) and if it did, the implicit behavior around that kinda scares me :flushed: Maybe best to leave the config explicit for now. It's very easy and clear.

(Closing, but can continue discussion if needed.)

francislavoie:
FWIW, I was right, the path is actually `""` at this point after `handle_path` strips the path prefix, so it's trivial to add a check for that, and I really can't think of any harmful side effects. If the path was `""` for any other reason and the code reaches the `browse` code (because it mapped to a directory on disk) and the original request _did not_ have a trailing slash, then I see no way for there to be a problem by just triggering a redirect to add the slash.

PR #4549, opened 2022-01-27
fileserver: Fix canonicalization redir when within `handle_path`

francislavoie:
Fix https://github.com/caddyserver/caddy/issues/4466

See the issue for repro details.

Of note, `path.Base(r.URL.Path)` when the path is `""` returns `.`, so the base comparison check which will never be true when the path has been stripped fully.


mholt:
:weary: This patchs till makes me nervous, but alright, fine, we can try this. :cold_sweat: 

But please be prepared to help patch if there are any bug reports for cases we didn't consider. :innocent: Thanks as usual!

mholt (approved):
:weary: This patch still makes me nervous, but alright, fine, we can try this. :cold_sweat: 

But please be prepared to help patch if there are any bug reports for cases we didn't consider. :innocent: Thanks as usual!

Issue #4509, opened 2022-01-06
SIGSEGV in sendClientCertificate


hntrmrrs:
When using Caddy to reverse proxy a connection using mTLS with ACME-issued certificates, I've encountered a troubling segfault.

This machine (`ingress-001.gwar`) also has DNS record `ingress.gwar` which wants to reverse proxy all requests to `service1.gwar:4601`. `service1` requires Caddy to use a client cert issued by its trusty ca (`ca.gwar`). If `ingress-001` tries to use the same cert for client auth as it uses for serving on port 443, I get the segfault.

If I remove `tls_client_auth`, make a request (thereby getting the ACME cert issued), and then add `tls_client_auth` back, it works without issue (I haven't waited for certificate renewal yet to confirm if it continues working--certs in this test env have a ~240h expiry).

`Caddyfile`:

```
{
  email acme@mail.gwar
  acme_ca https://ca.gwar/acme/acme/directory
  acme_ca_root /opt/certs/root_ca.crt
  cert_issuer acme
}

ingress-001.gwar:443 {
  respond 404
}

ingress.gwar:443 {
  reverse_proxy {
    to https://service1.gwar:4601
    health_uri /u/up
    transport http {
      tls_client_auth ingress.gwar
    }
  }
}
```

Syslog:

```
Jan  1 00:00:00 ingress caddy[101]: {"level":"info","ts":1640995200.408815,"msg":"serving initial configuration"}
Jan  1 00:00:00 ingress caddy[101]: {"level":"info","ts":1640995200.4100556,"logger":"tls","msg":"cleaning storage unit","description":"FileStorage:/var/lib/caddy/.local/share/caddy"}
Jan  1 00:00:00 ingress caddy[101]: {"level":"info","ts":1640995200.4105682,"logger":"tls","msg":"finished cleaning storage units"}
Jan  1 00:00:00 ingress caddy[101]: {"level":"info","ts":1640995200.4106896,"logger":"tls.obtain","msg":"lock acquired","identifier":"ingress.gwar"}
Jan  1 00:00:00 ingress caddy[101]: {"level":"info","ts":1640995200.4116905,"logger":"tls.issuance.acme","msg":"waiting on internal rate limiter","identifiers":["ingress.gwar"],"ca":"https://ca.gwar/acme/acme/directory","account":"acme@mail.gwar"}
[...]

PR #4550, opened 2022-01-27
reverseproxy: Avoid returning a `nil` error during GetClientCertificate

francislavoie:
Fix https://github.com/caddyserver/caddy/issues/4509


mholt (approved):
Thanks! Not sure how I missed this.

PR #4551, opened 2022-01-27
reverseproxy: Make shallow-ish clone of the request

francislavoie:
Alternate fix for #4457, to potentially be merged instead of #4468.

As suggested in #4468, this could potentially kill two birds with one stone, fixing messy header handling (holding onto the original headers from the request and having to reset it before the response response is being written) and removing a potential data race from manipulating the incoming request.

Basically I copied the code from `http.Request.Clone()` except for the bits that touch `r.ctx` (we want to keep the same context for our shallow clone), the `r.Form` stuff (multiple fields), and `r.TransferEncoding` (I don't think this one matters?) since those are won't be modified during proxying, and we don't need the copy overhead.

I haven't tested this out any further than to confirm it does fix the bug reported in #4457, and I don't know how to check for data races etc, so I'd like some help in vetting this.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -454,43 +454,15 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyht
 	return statusError(proxyErr)
 }
 
-// prepareRequest modifies req so that it is ready to be proxied,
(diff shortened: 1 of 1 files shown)


mholt (approved):
LGTM. I refactored the request cloning into a separate function. Feel free to merge if the tests pass! Then we can close the alternate PR and the issue I think.

I imagine the performance impact will be negligible, and the correctness gains are far better.

Issue #4563, opened 2022-02-05
Add Markdown documentation


izzues:
Hey, thanks for the awesome project!

I'm interested in using Markdown as described [here](https://caddyserver.com/docs/caddyfile/directives/templates) and I'd like to know a little more about things like:

- What Markdown "dialects" are supported and what is the default? (What syntax constructs are supported?)
- Is it configurable at all?

Thanks  again!

francislavoie:
> What Markdown "dialects" are supported and what is the default?

Caddy uses https://github.com/yuin/goldmark, so CommonMark, plus Github Flavored, plus footnotes, plus syntax highlighting (which uses Chroma https://github.com/alecthomas/chroma, so any of those languages).

> Is it configurable at all?

Not currently. What would you need to configure?

izzues:
> What would you need to configure?

Thanks for the quick reply, that's a very satisfying answer for my use case already!

Other users might want to know whether you can disable or customize features (for instance, disable footnotes and use some alternative table syntax) and add plugins (to manipulate the AST). It would be helpful to know even if the docs just said "you can't do any of these". Just a few suggestions, though. Thanks again!

izzues:
I just saw your PR with comments in a Go file. It would be nice to have that info on the website as well!

francislavoie:
It is on the website: https://caddyserver.com/docs/modules/http.handlers.templates

It's the first link on the `templates` directive page. But I'll make that link more prominent.

izzues:
Oops, sorry then, I wasn't not familiar with how this works. Really nice, though! Thanks again!

PR #4564, opened 2022-02-05
templates: Elaborate on what's supported by the markdown function

francislavoie:
Fix https://github.com/caddyserver/caddy/issues/4563


mholt (approved):
Good idea, thanks.

PR #4565, opened 2022-02-06
cmd: Migrate to `spf13/cobra`, remove single-dash arg support

mohammed90:
This PR swaps the homegrown CLI parsing and management with [github.com/muesli/coral](https://github.com/muesli/coral), a fork of [github.com/spf13/cobra](https://github.com/spf13/cobra). The switch brings the benefit of being able to generate shell-completions for major shells (PowerShell, bash, zsh, fish) and manpages. It's a POC/proposal just to see the impact of the change. It brings only 1 direct dependency and 3 transitive ones.

The current tricky part is figuring out how to adapt the current `caddycmd.Command` to `*coral.Command`. The challenge is in mapping the flags. Once that is figured out, all is easy.

P.S.: Never mind the deletions in commandfunc.go. I'm only using that to track which ones aren't ported yet. Once all is ported, I'll split the funcs back into the file.

Edit: As of [Cobra v1.4.0](https://github.com/spf13/cobra/releases/tag/v1.4.0), there is not dependency on Viper, which thins its dependency to minimum and matches coral, so swapping coral for cobra does not introduce any new deps (across the two).

Supercedes caddyserver/dist#73


cmd/coral.go
```diff
@@ -0,0 +1,383 @@
...
+			Title:   "Caddy",
+			Section: "1",
+		}, "./")
+	},
+}
+
+func init() {
+	listModulesCmd.Flags().Bool("packages", false, "Print package paths")
+	listModulesCmd.Flags().Bool("versions", false, "Print version information")
+	listModulesCmd.Flags().Bool("skip-standard", false, "Skip printing standard modules")
```
francislavoie, line 347: Hmm, this is an awkward API. It doesn't have a way to define the flags along with the command `coral.Command` higher up? feels a bit detached to have to define these separately.

mohammed90: Yeah, I didn't like the ergonomics of it much. It doesn't seem to be possible from the docs, for one `flags` is a private field of `coral.Command`, but it might be cleaner if I figure out how to adapt `caddycmd.Command` to `*coral.Command` since the former has its flags exposed and available for reading.

mohammed90: And now?

francislavoie: Nice! Yeah that looks much better.

cmd/commandfuncs.go
```diff
@@ -576,70 +574,6 @@ func cmdFmt(fl Flags) (int, error) {
 	return caddy.ExitCodeSuccess, nil
 }
 
-func cmdHelp(fl Flags) (int, error) {
-	const fullDocs = `Full documentation is available at:
-https://caddyserver.com/docs/command-line`
```
francislavoie, line 581: I notice the help was removed, I assume coral has its own generated help.

Can we also add a preamble at the top of the coral help output to link to the docs site like it used to here?

francislavoie:
Worth noting this could be a breaking change for anyone who happens to use `-` single dash flags instead of `--` double dash, because this uses `pflags` instead.

But this is also a benefit in the sense that we could now provide single letter shorthands for some flags.

cmd/commandfuncs.go
mohammed90: There's a way to manipulate the help template. I've changed it so it contains the preamble and an epilogue. The current help print is:

```
Caddy is an extensible server platform.

Usage:
  caddy [command]

Available Commands:
  adapt          Adapts a configuration to Caddy's native JSON
  add-package    Adds Caddy packages (EXPERIMENTAL)
  build-info     Prints information about this build
  completion     Generate the autocompletion script for the specified shell
  environ        Prints the environment
  file-server    Spins up a production-ready file server
  fmt            Formats a Caddyfile
  hash-password  Hashes a password and writes base64
  help           Help about any command
  list-modules   Lists the installed Caddy modules
  manpage        Generate the man pages
  reload         Changes the config of the running Caddy instance
  remove-package Removes Caddy packages (EXPERIMENTAL)
  reverse-proxy  A quick and production-ready reverse proxy
  run            Starts the Caddy process and blocks indefinitely
  start          Starts the Caddy process in the background and then returns
  stop           Gracefully stops a started Caddy process
  trust          Installs a CA certificate into local trust stores
  untrust        Untrusts a locally-trusted CA certificate
  upgrade        Upgrade Caddy (EXPERIMENTAL)
  validate       Tests whether a configuration file is valid
  version        Prints the version

Flags:
  -h, --help   help for caddy

Use "caddy [command] --help" for more information about a command.

Full documentation is available at:
https://caddyserver.com/docs/command-line
```

francislavoie: Nice!

mohammed90:
The single-dash flags seem to be already used and referenced across the interwebs, examples:
- https://wiki.archlinux.org/title/Caddy
- https://mercure.rocks/docs/hub/install
- https://www.booleanworld.com/host-website-caddy-web-server-linux/

This PR introduces a breaking change to the CLI. Scripts and systemd units using the single-dash flag will break and stop working. There doesn't seem a way to do this in backwards compatible manner, unless I'm missing something.

mholt:
Thanks for working on this Mohammed. It's been a while since I looked at the manpages thing; is the main advantage of this approach over https://github.com/caddyserver/dist/pull/73 the support for multiple shells?

Given the extensive changes needed and the single-hyphen thing, I am not sure if it is worth it at this point... which is a bummer to say, but... what do you think?

francislavoie:
I'd love to get shortcuts for flags so we could do stuff like `caddy adapt -c=/path/to/Caddyfile -p`. Nice and short. Annoying to type out `--pretty` every time I'm playing around.

I personally think the breakage is worth it for the UX benefits. The help docs will be normalized nicely (no longer confusingly use `-config` form etc), and completions is a nice win.

This is also overall less code, because some stuff we did ourselves is handled by coral automatically already (like the help command).

But we do need to make it clear in the release notes when this goes out that this is a breaking change for those that do use the `-` form which we already did discourage (albeit softly). This won't break anyone's Docker or Systemd setups.

cmd/caddy/main.go
```diff
@@ -34,5 +34,5 @@ import (
 )
 
 func main() {
-	caddycmd.Main()
+	caddycmd.Execute()
```
mohammed90, line 37: Leaving a note for self when I come back to this later: changing Main to Execute this breaks xcaddy because the template uses `caddycmd.Main()`. That said, `.Execute()` doesn't have to be `.Execute()`. Whatever `.Execute()` does, `.Main()` can do, and we just keep `.Main()` and not have `.Execute()`.

mohammed90:
As of [Cobra v1.4.0](https://github.com/spf13/cobra/releases/tag/v1.4.0), there is not dependency on Viper, which thins its dependency to minimum and matches coral, so swapping coral for cobra does not introduce any new deps (across the two).

mohammed90:
I guess it's ready for review now. I hope I didn't miss anything!

mholt:
Cool, looking forward to checking this out once I am back in the office. Thanks Mohammed!

mholt:
I just realized that single-hyphen flags were documented when using the standard Go `-h` flag to get help. So that might be a bummer.

cmd/main.go
```diff
@@ -470,5 +429,9 @@ func (ss *StringSlice) Set(value string) error {
 	return nil
 }
 
+// TODO: should it implement either of those?
+// var _ pflag.Value = (*StringSlice)(nil)
+// var _ pflag.SliceValue = (*StringSlice)(nil)
```
francislavoie, line 434: Do we need to do anything with this?

cmd/cobra.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use: "caddy",
+}
+
+const docsHeader = `Caddy is an extensible server platform.`
```
mholt, line 11: This appears at the top of _every_ help command, e.g. `caddy help stop` -- I think this only needs to appear for `caddy help`, not subcommands

cmd/commands.go
```diff
@@ -346,6 +345,41 @@ EXPERIMENTAL: May be changed or removed.
 		}(),
 	})
 
+	RegisterCommand(Command{
+		Name: "manpage",
```
mholt, line 349: This is cool, how do I actually use this on my system so I can do `man caddy` and have it show me the manpages?

cmd/main.go
mholt: I'd say let's wait on that unless we really have a need for it

mholt (approved):
This is very cool, thanks. Finally pulled it down and tried it out. I like that it's fairly minimal change, not very disruptive!

Approving, but I have a couple minor nits (see other comments)

Kind of a bummer that we have 2 markdown libs in our dependencies now, since we use goldmark and cobra uses blackfriday. Oh well. Nothing we can do about that, just pointing it out.

How do we go about adding short forms to some of the flags?

I also want to try out the manpages and shell completion. How do we go about doing that?

cmd/commands.go
francislavoie: My understanding is it would look something like this (not tested):

```
caddy manpage | sudo tee /usr/share/man/man1/caddy && sudo mandb
```

But really, package installers will do this on install/update. I'll probably update the package install script for the apt repo after playing around with it for a while.

francislavoie: Confirmed this isn't quite right, I assumed `caddy manpage` would spit out to stdout but it takes a `--directory` instead. But this is probably pretty close to being correct.

cmd/cobra.go
mohammed90: Done!

mholt:
Woohoo!

cmd/commands.go
francislavoie: This worked:
```
$ mkdir man
$ ./caddy manpage --directory man
$ gzip -r man/
$ sudo cp man/* /usr/share/man/man8/
$ sudo mandb
```

Will need to add this to the deb package installation later.

PR #4569, opened 2022-02-11
admin: Stop writing 500 status on invalid requests

### LINKED ISSUE #4561: Caddy API returns a 500 when trying to POST/GET an `id` that does not exist
As you can see here:
image

Caddy is returning a 500 when trying to get an ID that does not exist.  This should be a 404 instead of a 500. We are trying to make a k8 operator which has a CreateOrUpdate function for specific routes meaning we need to check if an ID exists. It would be nice to have  404 instead of a 500.

mintbomb27:
Fixes #4561


mholt (approved):
Thank you! Did you test that this works as intended? (I would be surprised if it doesn't.)

PR #4571, opened 2022-02-11
caddyhttp: remote_ip matcher strip zone from end of IPv6

BitWuehler:
Splits the remote ip variable to exclude the zone identifier from the ip and port. Otherwise it can't be compared with ips in filters.

Just tested in separate go file yet. Okay I don't really now how to make myself a docker container for testing out of the sourcecode. So I can't test it more than that.


modules/caddyhttp/matchers.go
```diff
@@ -905,6 +905,11 @@ func (m MatchRemoteIP) getClientIP(r *http.Request) (net.IP, error) {
 			remote = strings.TrimSpace(strings.Split(fwdFor, ",")[0])
 		}
 	}
+	// Some local ipv6-adresses contain zone identifiers at the end,
```
mholt, line 908:
```suggestion
	// Some IPv6 addresses contain zone identifiers at the end,
```

mholt (approved):
Thanks for the PR!

I am not savvy enough with IPv6 to know... is this the right way to handle this? I wonder if we should be _factoring in_ the zone ID in matching, or something like that.

I also have no idea as to the security implications of zone IDs (and ignoring them) in this context

Anyway, I think this is likely a good fix for now, and if anyone reports a bug, would you be able to help us fix it? 😃

modules/caddyhttp/matchers.go
BitWuehler: At first I have to say, that I am not an expert in IPv6 too. 

As far as my research shows, I am pretty shure that zone_IDs only play a role in `link-local IPv6-adresses`. The RFC 6874 document says:
```
The syntax defined above allows a ZoneID to be added to any IPv6
address.  The 6man WG discussed and rejected an alternative in which
the existing syntax of IPv6address would be extended by an option to
add the ZoneID only for the case of link-local addresses.  It was
felt that the solution presented in this document offers more
flexibility for future uses and is more straightforward to implement.
```
So for the moment it's local but maybe in the future other IPv6-Adresses will use it too.

To the security point:
An example for the use of the zone_id is a server with serveral subnets. Because link-local adresses uses all only the fe80::/64 prefix, it could be, that if you clone a machine or very rare randomly, you could have the same link-local IPv6-adress on different machines on different subnets. The zone_id helps to speak only to the right machine in that case.
More about that [here](https://4sysops.com/archives/ipv6-tutorial-part-7-zone-id-and-unique-local-ipv6-unicast-addresses/).
So, in terms of security, I would not say, that this is not a problem, because it appears only in local networks.
 
Seems like nobody had a problem with zone_ids before, so I would say we don't have to be fast to implement a handling of the zone_id. But I would say, that it could be helpfull in the future to have implemented it allready and not only for local adresses.

Because I am really new to golang, I will need my time to write something, but I want to and I allready started. Maybe I have a few questions to specific parts. But I will do some research first and than we'll see.

And sure, if anyone reports a bug, I will give my best to help to fix it.
Thank you for your work. I love the caddy project!

BitWuehler:
@mholt 
To the security point:
An example for the use of the zone_id is a server with serveral subnets. Because link-local adresses uses all only the fe80::/64 prefix, it could be, that if you clone a machine or very rare randomly, you could have the same link-local IPv6-adress on different machines on different subnets. The zone_id helps to speak only to the right machine in that case.
More about that [here](https://4sysops.com/archives/ipv6-tutorial-part-7-zone-id-and-unique-local-ipv6-unicast-addresses/).
So, in terms of security, I would not say, that this is not a problem, because it appears only in local networks.

Seems like nobody had a problem with zone_ids before, so I would say we don't have to be fast to implement a handling of the zone_id. But I would say, that it could be helpfull in the future to have implemented it allready and not only for local adresses.

Because I am really new to golang, I will need my time to write something, but I want to and I allready started. Maybe I have a few questions to specific parts. But I will do some research first and than we'll see.

And sure, if anyone reports a bug, I will give my best to help to fix it.
Thank you for your work. I love the caddy project!

BitWuehler:
Opened a new Pull request for handling zone identifiers.  #4597

Issue #4578, opened 2022-02-16
Admin API: ETag on configuration


jhwz:
When updating Caddy configuration dynamically through the admin API, the usual flow for me is:
- Get the current configuration
- Figure out what changes need to be made to add this service to the config
- Make the changes and POST/PUT/PATCH them back to the Caddy API

This becomes an issue when multiple services are doing this at once, there is a chance for collisions. I would suggest adding an ETag header on the responses to GET requests. Initially, the ETag could just be the hash of the entire config but I guess you could compute the hash of sub-regions if there is a need.

This way, when an update is made you can optionally send the If-Match header to only apply the update if no changes have been made in the meantime and if a service receives 412 response, they know they need to re-fetch the config and try again.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#avoiding_mid-air_collisions

mholt:
Thanks, I think this is a good suggestion.

jhwz:
See https://github.com/caddyserver/caddy/pull/4579 for a basic implementation.

mholt:
Forgot to close this; thanks for the cool new feature!

Issue #3296, opened 2020-04-22
api: Support for batch config modifications


francislavoie:
The [Caddy API](https://caddyserver.com/docs/api) in v2 is very powerful.

One area that I found it lacking is in being able to make batch modifications in one atomic operation. Currently, the only possible way to do this would be to GET the whole config (or at least a superset that contains all the sections you wish to modify), make the JSON modifications offline, then POST the entire set back to the API.

An approach I thought of that could improve this is by having a [JSON-RPC](https://www.jsonrpc.org/specification) endpoint available in the API. This would allow for doing batch operations to the JSON config atomically, i.e. Caddy would only reload one time, after the entire batch is executed.

For example, here's a possible JSON-RPC request payload to modify two different servers' listen addresses at once:

```
$ curl -X POST \
    -H "Content-Type: application/json" \
	-d '
	[
	  {
	    "json-rpc": "2.0",
	    "method": "PUT",
	    "params": {
	      "path": "/config/apps/http/servers/srv0/listen/0",
	      "data": ":8080"
	    }
	  },
	  {
	    "json-rpc": "2.0",
	    "method": "PUT",
	    "params": {
	      "path": "/config/apps/http/servers/srv1/listen/0",
	      "data": ":8081"
	    }
	  }
	]' \
    "http://localhost:2019/json-rpc"
```

This would essentially be the same as doing the following, except that the following would trigger two config reloads because those two operations are not atomic:

```
$ curl -X PUT \
	-H "Content-Type: application/json" \
	-d '":8080"' \
	"http://localhost:2019/config/apps/http/servers/srv0/listen/0"

$ curl -X PUT \
	-H "Content-Type: application/json" \
	-d '":8081"' \
	"http://localhost:2019/config/apps/http/servers/srv1/listen/0"
```

This issue is mainly to guage interest to see whether anyone else would actually need this sort of feature -- I don't immediately need it myself, but I spotted the lack in the current API so I figured I'd type up a potential solution.

sospartan:
> Caddy would only reload one time, after the entire batch is executed.
One feature that bother me is why Caddy restart everything when config changed, not the app(s)  the changes do affected?

mholt:
[...]

PR #4579, opened 2022-02-16
admin: support ETags

jhwz:
For #4578.

Computes a MD5 hash of the full config when it is updated. This is added to the response headers of all GET requests.

Adds an additional argument to the `changeConfig` function which if not empty, will return an error if it doesn't match the current config hash.

This PR doesn't yet include tests. I can add tests for just the `changeConfig` function later but I'm not sure if there is an easy way to test the `adminHandler`?

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -865,6 +865,7 @@ func handleConfig(w http.ResponseWriter, r *http.Request) error {
 	switch r.Method {
 	case http.MethodGet:
 		w.Header().Set("Content-Type", "application/json")
+		w.Header().Set("ETag", configHash())
 
 		err := readConfig(r.URL.Path, w)
 		if err != nil {
@@ -904,7 +905,7 @@ func handleConfig(w http.ResponseWriter, r *http.Request) error {
 
 		forceReload := r.Header.Get("Cache-Control") == "must-revalidate"
 
-		err := changeConfig(r.Method, r.URL.Path, body, forceReload)
+		err := changeConfig(r.Method, r.URL.Path, body, r.Header.Get("If-Match"), forceReload)
 		if err != nil {
 			return err
 		}
--- a/caddy.go
+++ b/caddy.go
@@ -17,6 +17,8 @@ package caddy
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -110,7 +112,7 @@ func Load(cfgJSON []byte, forceReload bool) error {
 		}
 	}()
 
-	return changeConfig(http.MethodPost, "/"+rawConfigKey, cfgJSON, forceReload)
+	return changeConfig(http.MethodPost, "/"+rawConfigKey, cfgJSON, "", forceReload)
 }
 
 // changeConfig changes the current config (rawCfg) according to the
@@ -119,7 +121,7 @@ func Load(cfgJSON []byte, forceReload bool) error {
 // If the resulting config is the same as the previous, no reload will
 // occur unless forceReload is true. This function is safe for
 // concurrent use.
-func changeConfig(method, path string, input []byte, forceReload bool) error {
+func changeConfig(method, path string, input []byte, ifMatch string, forceReload bool) error {
 	switch method {
 	case http.MethodGet,
 		http.MethodHead,
@@ -131,6 +133,13 @@ func changeConfig(method, path string, input []byte, forceReload bool) error {
 
 	currentCfgMu.Lock()
 	defer currentCfgMu.Unlock()
+
+	if ifMatch != "" && ifMatch != rawCfgHash {
+		return APIError{
+			HTTPStatus: http.StatusPreconditionFailed,
+			Err:        fmt.Errorf("If-Match header did not match current config hash"),
+		}
+	}
 
 	err := unsyncedConfigAccess(method, path, input, nil)
 	if err != nil {
@@ -189,6 +198,8 @@ func changeConfig(method, path string, input []byte, forceReload bool) error {
 	// each config change)
 	rawCfgJSON = newCfg
 	rawCfgIndex = idx
+	md5Hash := md5.Sum(rawCfgJSON)
+	rawCfgHash = hex.EncodeToString(md5Hash[:])
 
 	return nil
 }
@@ -199,6 +210,15 @@ func readConfig(path string, out io.Writer) error {
 	currentCfgMu.RLock()
 	defer currentCfgMu.RUnlock()
 	return unsyncedConfigAccess(http.MethodGet, path, nil, out)
+}
+
+// configHash simply returns the cache of the current configs hash.
+// This is recomputed after each change to the config so should be quick
+// to access.
+func configHash() string {
+	currentCfgMu.RLock()
+	defer currentCfgMu.RUnlock()
+	return rawCfgHash
 }
 
 // indexConfigObjects recursively searches ptr for object fields named
@@ -548,6 +568,7 @@ func Stop() error {
 	currentCfg = nil
 	rawCfgJSON = nil
 	rawCfgIndex = nil
+	rawCfgHash = ""
 	rawCfg[rawConfigKey] = nil
 	return nil
 }
@@ -780,6 +801,11 @@ var (
 	// rawCfgIndex is the map of user-assigned ID to expanded
 	// path, for converting /id/ paths to /config/ paths.
 	rawCfgIndex map[string]string
+
+	// rawCfgHash is the MD5 hash of the current rawCfgJSON. Using
+	// this combined with ETags clients are able to avoid mid-air
+	// collisions. The rawCfgHash is hex encoded.
+	rawCfgHash string
 )
 
 // ImportPath is the package import path for Caddy core.


jhwz:
A good alternative to `md5` hash is [highway hash](https://github.com/minio/highwayhash) if you want to use that instead to stop the lint issues with `gosec`

mholt (changes requested):
Thanks; I'm thinking of a different way to do this though. It's probably better if the hash isn't persisted in memory -- just compute it on the fly. That will let us serve hashes for any subset of the config as well as the whole config all the same.

And yeah, definitely don't use MD5. Let's use [`fnv`](https://pkg.go.dev/hash/fnv). We use fnv in the file server for ETags and it seems to work OK and is still fast.

jhwz:
Sweet no problem at all!

What I've generally been doing is reading the entire config down, checking its state and then updating the parts of the config that I need to. If you are serving hashes for subsets of the config, then I'd need to update that same subset (it might get trickier again for deletes and so on). Would it be too restrictive?

Also, if doing global updates to the config you'd need to read it into a `map[string]interface{}` like Caddy does to preserve all existing config - instead of using structs for the subset you care about - which makes it a bit more annoying to work with.

`fnv` makes sense!

sarge:
@jhwz regarding tests, you could have a look at the caddytest examples - they update via the admin api. It would be a nice enhancement to the caddytests to check the Etag to verify that the configuration has been updated

mholt:
@jhwz 

> If you are serving hashes for subsets of the config, then I'd need to update that same subset (it might get trickier again for deletes and so on). Would it be too restrictive?
>
> Also, if doing global updates to the config you'd need to read it into a map[string]interface{} like Caddy does to preserve all existing config - instead of using structs for the subset you care about - which makes it a bit more annoying to work with.

I'm not sure I follow, maybe I'm missing something here. There isn't any need to "update" the hashes for any subset, because we aren't storing them. Just always compute the hash for every response on-the-fly. No structs, no maps, no storing hashes, just call `fnv()` or whatever and then write that value to the ETag header.

jhwz:
Sorry, I may not have explained that well.

When I talk about updating, I'm referring to a client of the Caddy server 
1. Fetching the config and the computed ETag with a HTTP GET request
2. Reading that config from the GET request into some structure it can understand
3. Make changes based on that state.
4. POST those changes back to Caddy with the `If-Match` header set to the same value as found in the ETag header.

The issues I was talking about above arise when you want to make changes to subsets of that config. Imagine I fetch the config at `localhost:2019/config/apps/http/servers` and receive the response:
```
{
  srv0: {
  ...
  },
 srv1: {
  ...
 }
}
```
with an ETag representing the hash of that JSON value.


If I then want to delete `srv1` for some reason, you would send DELETE to `localhost:2019/config/apps/http/servers/srv1`. Doing this, I wouldn't be able to use that same `ETag`  header sent with the original GET request in the `If-Match` because the route is referring to another part of the config.

The solution to that is POST to the same place you fetched from, but that means I would need to preserve the original response completely so I can send back the unmodified fields. It's also a bit more of a pain to use.

Hope that made more sense!

mholt:
Ok, I allllmost understand. Thanks for explaining. I follow until here:

> Doing this, I wouldn't be able to use that same ETag header sent with the original GET request in the If-Match because the route is referring to another part of the config.

What route refers to another part of the config? Just because srv1 is gone, doesn't mean that the paths change, does it? Can you give a specific example of requests that illustrate what you're describing?

Sorry about the trouble. It's probably just me being slow, it is after midnight, after all. 🙃

jhwz:
No worries, I may be being silly and missing something here too.

>What route refers to another part of the config? Just because srv1 is gone, doesn't mean that the paths change, does it? Can you give a specific example of requests that illustrate what you're describing?

The case where the routes do change in particular is arrays. Imagine two clients make a request to Caddy and get back an array of routes like:
```json
            [
              {
                "handle": [{ "handler": "reverse_proxy", ... }],
                "match": [{"path": ["/aaa*"]}]
              },
              {
                "handle": [{ "handler": "reverse_proxy", ... }],
                "match": [{"path": ["/bbb*"]}]
              },
              {
                "handle": [{ "handler": "reverse_proxy", ... }],
                "match": [{"path": ["/ccc*"]}]
              }
            ]
 ```


1. If client 1 wants to delete array index 0, and client 2 wants to delete array index 1 then the order of those operations matter. We want each update to be atomic (this is the entire point of the PR). If client 1 runs the DELETE operation first, then for client 2 index 1 actually refers to index 2 in the original array (because the first item is removed). What we want is a way of bailing out an update if the config has changed in the meantime.

1. I think this more addresses your question: \
Let's assume one of the clients wants to delete item 0. The `ETag` of the original request is the hash of the array of items (lets say the `ETag` of the array is `1234`). But the `ETag` of the array at index 0 is different, as that's just the hash of the object (let's say that `ETag` is `5678`). So we cannot use the returned `ETag` header from the original request as the `If-Match` request because the two paths refer to different resources (`1234 != 5678`)..\
We would need to first request the index of the array that we care about to get its `ETag`, and then try update it with the `If-Match` header set so we can be sure no-one else is making changes.


In general, I think it may be simpler to have a global hash so you can say "if anything changed since I requested let me know". It is less accurate because you may need to bail out on updates when they are performed on completely different parts of the running config, but I imagine that would happen fairly rarely out in the wild?

**Comment by mholt** [comment]
Ohhhh yes, I see what you are saying now, I think that is crystal clear this time. Thank you!  I suspect a global ETag is not a bad idea now.

I will sleep on this (it is after midnight right now, and I will only be intermittently available this week) and get back to you with more of my thoughts.

**Comment by jhwz** [comment]
No problem at all, should have started with that! Interested to see what you come up with.

**Comment by mholt** [comment]
Okay, so yes, this makes sense what you're saying with that scenario. Thinking about this independently for a while now, I agree with your conclusion:

> We would need to first request the index of the array that we care about to get its ETag, and then try update it with the If-Match header set so we can be sure no-one else is making changes.

In other words, this approach feels the most right to me. Before you make a change on a resource, the client should get its ETag so it can set If-Match on its subsequent request. That might require an extra request to get the ETag in situations like what you're describing, but I still think that's okay: the client can make a second request to get the ETag for a certain value in the config. I bet most API users won't have this problem, either.

(Another option is to document the hashing algorithm to have the client produce its own ETag that it can use without having to make the second request. If it's `fnv`, then it is pretty simple.)

So while I understand the appeal of keeping and using the hash for the whole config instead, I have a counter-perspective on this point (emphasis mine):

> It is less accurate because you may need to bail out on updates when they are performed on completely different parts of the running config, but I imagine **that would happen fairly rarely** out in the wild?

I think the reality is that a concurrent change to _any part_ of the config is **more likely** than a concurrent change to just the narrow part of the config that would affect the part that the client wants to mutate. So if we do that, and the client sends its If-Match header, the request would fail, and the client would have to start over from the beginning (make a GET request to get the new ETag, then make the POST/PUT/PATCH/DELETE request with the updated If-Match header... even if the change didn't affect the relevant config section at all.

So actually, I think an additional GET request just to get the ETag _in the relatively few scenarios like you describe that require it_ is probably a better, simpler approach (both for clients and our implementation of the admin API) than making the client start over just because another client made any change to the config whatsoever.

Another idea is that clients who expect to be manipulating the config concurrently should synchronize out-of-band to avoid having to implement "compare-and-swap" algorithms with HTTP. While they do prevent unintended changes, they also don't guarantee the change is ever actually made (it's nondeterministic) though of course the changes go up as more retries are performed. But clients that are this smart would put a cap on the number of attempts anyway, to avoid obvious problems.

So, I dunno if this made sense at all, let alone convinced you that this is a better way to go; it's after midnight here again and I might not be making any sense, sorry :)

**Comment by jhwz** [comment]
> In other words, this approach feels the most right to me. Before you make a change on a resource, the client should get its ETag so it can set If-Match on its subsequent request.

I agree, the only way this scenario differs I think is that usually an endpoint represents a single resource. In this case the Caddy API `/config/*` endpoints all refer to the same resource (kind of, for example arrays are coupled. Changes to one resource affects another). Whether or not this discrepancy matters is definitely up for debate though.

> On the contrary, I think that a concurrent change to any part of the config is more likely than a concurrent change to just the narrow part of the config that would affect the part that the client wants to mutate

Yeap, you're dead right there; must not have been thinking clearly!


>Another idea is that clients who expect to be manipulating the config concurrently should synchronize out-of-band to avoid having to implement "compare-and-swap" algorithms with HTTP. While they do prevent unintended changes, they also don't guarantee the change is ever actually made (it's nondeterministic) though of course the changes go up as more retries are performed. But clients that are this smart would put a cap on the number of attempts anyway, to avoid obvious problems.

I agree that for effective synchronisation another solution should be implemented. However something like this is opt-in (a mutation would only be rejected if the `If-Match` header was present *and* not-matching) and for most use cases 'compare-and-swap' algorithms work fine!

Not sure how you want to move forward on this. I think that hashing the relevant part of the config rather than the entire config is probably wiser. That way things are more likely to just work without breaking. I hope there's nothing I'm missing!

Thanks for taking the time to go through this - much appreciated. I'm more than happy to update this PR with whatever changes you need, or let you guys handle it.

**Comment by mholt** [comment]
Still thinking about this. I almost had a good idea then I came down with a nasty stomach bug.... so I'll try to get back around to it this week.

**Comment by jhwz** [comment]
Yep no stress - this isn't urgent.

**Comment by mholt** [comment]
Ok I have a couple ideas. Both are a bit hacky/abusive of the HTTP standard, but let's roll with this for a moment and see how it goes:

Suppose, as in your scenario above, you do a GET request and get the following response body:

```
            [
              {
                "handle": [{ "handler": "reverse_proxy", ... }],
                "match": [{"path": ["/aaa*"]}]
              },
              {
                "handle": [{ "handler": "reverse_proxy", ... }],
                "match": [{"path": ["/bbb*"]}]
              },
              {
                "handle": [{ "handler": "reverse_proxy", ... }],
                "match": [{"path": ["/ccc*"]}]
              }
            ]
```

And upon inspecting that payload, your code determines that you want to delete array element 0 (the first route). You could send a DELETE request, but you'd need to set the `If-Match` header to prevent deleting the wrong element in case the array changed in the meantime. The problem is, we don't have the `Etag` for array element 0, only for the whole array which we just requested.

So yes, we could do another GET request of just that first element and get its `Etag`, but we'd have to do a string comparison to make sure it is the same as what we saw from our first request containing the entire array -- ugh, what a pain.

Or, the `Etag` could represent the entire config document. But that's also a pain since it is very likely that any change to the config has no effect on the part of the config we're looking at, so it'd lead to unnecessary retries and races.

OR... we could slightly abuse `ETag`/`If-Match` by setting a special ETag value in the `If-Match` header ([this is allowed since ETags are just arbitrary strings](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match). We could format the ETag to look like this:

```
"/path/to/item etag"
```

In other words, we can specify the ETag for a different part of the config. In our case, the part we have an ETag for (the whole array):

```http
DELETE /config/apps/http/servers/foo/routes/0
If-Match: "/config/apps/http/servers/foo/routes asdf1234"
```

This way we don't need to make another GET request to look up the ETag, nor do we have to abort if any other irrelevant part of the config changed.

This is unconventional but I think it's technically allowed. What do you think?

jhwz:
I'm all for this, what a great idea!

The convenience far outweighs any issues of introducing a non-standard header.

I can't imagine any use cases outside the ones we've described so far so I'd hope most users would appreciate the ease of use.

mholt:
Sounds good; let's do that then. Want to update the PR when you have time? (No hurry, as we're currently focused on the 2.5 release, and maybe this could make it in since it's still just beta, but if not it's OK if it's ready for 2.5.1 or something.)

jhwz:
No problem, I'll update the PR at some point over the next day or two

sarge:
Apologies for coming in late on this, there is a json patch RFC and an implementation in go. 

It might cover a few more edge cases? 

https://github.com/evanphx/json-patch

mholt:
@sarge Thanks for the suggestion. That looks like a better fit for an alternative JSON API, but all we need here is a way to know whether the document has been patched/modified since we last looked at it. Maybe it would be more fitting over at https://github.com/caddyserver/caddy/issues/3296

RussellLuo:
Just my two cents. If Caddy supports an alternative Admin API in [JSON-Patch](https://datatracker.ietf.org/doc/html/rfc6902) style, the Etag value may be more conventional. Something like:

```http
PATCH /config/apps/http/servers/foo/routes HTTP/1.1
If-Match: "asdf1234"

[
    { "op": "remove", "path": "/0" }
]
```

jhwz:
> Just my two cents. If Caddy supports an alternative Admin API in [JSON-Patch](https://datatracker.ietf.org/doc/html/rfc6902) style, the Etag value may be more conventional. 

Yeah definitely! I don't know much about JSON patch but it would work really well in these examples (and for #3296) but having something that works with the existing API would also be nice

RussellLuo:
> but having something that works with the existing API will also be nice

Couldn't agree more!

jhwz:
OK so I have a working implementation with a basic test.

The main thing worth discussing is the use of `Trailer` headers. I'm currently adding the ETag to the end of the request rather than at the start, so we can still write directly to the response body instead of to an intermediate buffer. This may be less intuitive to users however, we can easily change to using buffers instead if need be.

Also, slightly unrelated but I noticed if you check the error for the `TestLoadConcurrent` in `admin_test.go` you get the error `loading new config: loading http app module: unknown module: http` so the test isn't actually working. But that's a separate issue

mholt:
@jhwz I haven't forgotten about this. Thanks for the updates. I'm going to tag v2.5.0-rc.1 today and aim to get this into 2.5.1 or some patch release shortly thereafter. I think we're getting closer!

admin.go
```diff
@@ -894,16 +897,30 @@ func (h adminHandler) originAllowed(origin *url.URL) bool {
...
 func handleConfig(w http.ResponseWriter, r *http.Request) error {
 	switch r.Method {
 	case http.MethodGet:
 		w.Header().Set("Content-Type", "application/json")
-
-		err := readConfig(r.URL.Path, w)
+		// Set the ETag as a trailer header.
+		// The alternative is to write the config to a buffer, and
+		// then hash that.
+		w.Header().Set("Trailer", "ETag")
```
mholt, line 911: Though unconventional, this is allowed: https://httpwg.org/specs/rfc9110.html#rfc.section.8.8.3

Yay for RFC 9110 -- HTTP Semantics -- which says:

> A sender MAY send the ETag field in a trailer section (see [Section 6.5](https://httpwg.org/specs/rfc9110.html#trailer.fields)). However, since trailers are often ignored, it is preferable to send ETag as a header field unless the entity tag is generated while sending the content.

So, I dunno. I'd say let's go ahead and try this as a trailer and if we get complaints with _valid technical reasons_ why trailer won't work for this, then maybe we can switch to header + buffering.

admin.go
```diff
@@ -894,16 +897,30 @@ func (h adminHandler) originAllowed(origin *url.URL) bool {
...
+
+		hash := etagHasher()
+		configWriter := io.MultiWriter(w, hash)
+		err := readConfig(r.URL.Path, configWriter)
 		if err != nil {
 			return APIError{HTTPStatus: http.StatusBadRequest, Err: err}
 		}
 
+		// we could consider setting up a sync.Pool for the summed
+		// hashes to reduce GC pressure.
```
mholt, line 921: Just going to guess this won't be a significant burden on the GC, since this is just the config endpoint. But maybe experience will prove me wrong (as it often does)

However if we _did_ switch to Etag header instead of trailer, then we would certainly want to pool the response buffers.

mholt (approved):
So sorry for the delay. This change is actually great and I'm excited to try it out.

Looks good now! Thanks for the great work implementing it. I'd also like to extend a collaborator invite to you if you'd like to accept it, meaning you can contribute on a branch instead of a fork. It was a joy to collaborate with you on this enhancement.

I'm sure more work will be needed on this, even if just some fine-tuning. I'm going to merge this in and put it in v2.5.2 and see if we get any feedback before the next minor release. Could you help maintain this if there are bug reports/other issues?

Thanks again!

jhwz:
No problem at all!

I would be happy to contribute and help maintain this going forward. I hope this will end up being useful to others!

jhwz:
Damn yeah we probably should, if we do we should require the `If-Match` header to have double quotes to.

The only reason we wouldn't is because we're already acting 'non-standard' by using the `[path] [hash]` format. I'd probably lean towards having the double quotes though.

If so, I'm happy to make the change

mholt:
@jhwz Yeah, let's do double quotes when you have a sec. Thanks!!

PS. Docs for your feature are in here (look in api.md for "Concurrent config changes"): https://github.com/caddyserver/website/pull/247/files

PR #4585, opened 2022-02-19
caddyhttp: Move HTTP redirect listener to an optional module

francislavoie:
Followup to https://github.com/caddyserver/caddy/pull/4313

Turns out we have some issues with this being turned on by default, so we'll have to provide it as an optional module for now.

To turn it on, this needs to be added in global options:

```
{
	servers {
		listener_wrappers {
			http_redirect
			tls
		}
	}
}
```

### DIFF
--- a/modules/caddyhttp/httpredirectlistener.go
+++ b/modules/caddyhttp/httpredirectlistener.go
@@ -20,7 +20,44 @@ import (
 	"net"
 	"net/http"
 	"sync"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 )
+
+func init() {
+	caddy.RegisterModule(HTTPRedirectListenerWrapper{})
+}
+
+// HTTPRedirectListenerWrapper provides HTTP->HTTPS redirects for
+// connections that come on the TLS port as an HTTP request,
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/httpredirectlistener.go
```diff
@@ -20,8 +20,45 @@ import (
...
+//
+// This is especially useful when using a non-standard HTTPS port.
+// A user may simply type the address in their browser without the
+// https:// scheme, which would cause the browser to attempt the
+// connection over HTTP, but this would cause a "Client sent an
+// HTTP request to an HTTPS server" error response.
+//
+// This listener wrapper must be placed BEFORE the "tls" listener
+// wrapper, for it to work properly.
+type HttpRedirectWrapper struct{}
```
mholt, line 45:
```suggestion
type HTTPRedirectListenerWrapper struct{}
```

I know it's a mouthful but I think it makes more sense this way

mholt (changes requested):
Thanks Francis, this should do the trick. Looks good! Just a nit on the naming.

modules/caddyhttp/httpredirectlistener.go
francislavoie: Done

mholt (approved):
Awesome, thanks Francis!

PR #4592, opened 2022-02-21
httpcaddyfile: Support explicitly turning off `strict_sni_host`

francislavoie:
In the Caddyfile, we were only allowing turning on `strict_sni_host` explicitly, but provided no way to explicitly turn it off, which may be necessary for situations where it gets turned on implicitly like when TLS client_auth is turned on.

This adds support for a first arg to be `on` (for symmetry, least surprise) or `insecure_off` (what's actually new in terms of behaviour). If the arg is omitted, then it's the same behaviour as before, i.e. turning it on.

```
{
	servers {
		protocol {
			strict_sni_host [on|insecure_off]
		}
	}
}
```

### DIFF
--- a/caddyconfig/httpcaddyfile/serveroptions.go
+++ b/caddyconfig/httpcaddyfile/serveroptions.go
@@ -157,11 +157,14 @@ func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (interface{}, error
 						serverOpts.ExperimentalHTTP3 = true
 
 					case "strict_sni_host":
-						if d.NextArg() {
-							return nil, d.ArgErr()
+						if d.NextArg() && d.Val() != "off" && d.Val() != "on" {
+							return nil, d.Errf("strict_sni_host only supports 'on' or 'off', got '%s'", d.Val())
 						}
-						trueBool := true
-						serverOpts.StrictSNIHost = &trueBool
+						boolVal := true
+						if d.Val() == "off" {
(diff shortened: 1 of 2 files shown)


mholt:
I don't know if I love this (because it only turns _off_ security)... so if we are going to allow this to be done so easily (which, I still am not sure whether to approve this), it should probably at least be prefixed with `insecure_`. (Like TLS' `insecure_skip_verify` option, which I also hate, heh.)

francislavoie:
Alright, like `insecure_off`? I'm down with that.

mholt:
@francislavoie Oh, I meant instead of `strict_sni_host on|off`, just have `insecure_disable_strict_sni_host` as a single token or something. But now that you mention it, `strict_sni_host insecure_off` is also fine I guess, except I don't want people thinking they need to or can turn it on at all (since the only thing to really do is force it _off_).

Still squirming about this because I hope it doesn't bite anyone who doesn't understand the docs someday...

francislavoie:
Well, we've had it as an option to turn it on for quite a long time already. Removing that would be a BC break. I'll make the change for `insecure_off`.

mholt:
Ohh, right, because it only turns on automatically if client_auth is enabled. Gotcha. I like `insecure_off` then.

francislavoie:
Donezo

mholt (approved):
Thanks, I hate it 😉

PR #4596, opened 2022-02-23
Add a override_domain option to allow DNS challenge delegation

crccw:
CNAME can be used to delegate answering the chanllenge to another DNS
zone. One usage is to reduce the exposure of the DNS credential [1].
Based on the discussion in caddy/certmagic#160, we are adding an option
to allow the user explicitly specify the domain to delegate, instead of
following the CNAME chain.

This needs caddy/certmagic#160.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -362,7 +362,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 				}
 				acmeIssuer.Challenges.DNS.Resolvers = args
 
-			case "override_domain":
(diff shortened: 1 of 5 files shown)


francislavoie:
I think maybe we should change the option to something like `dns_challenge_override_domain` or something. That is very long, but I don't think it's clear upfront that this specifically applies to the DNS challenge only (it is clear in JSON config though because of nesting)

crccw:
Changed to `dns_challenge_override_domain`

crccw:
Just signed.

mholt:
@crccw Thanks!

> Do we need to wait for a certmagic release and bump the dependency version before merging this?

Already done :wink:

PR #4597, opened 2022-02-23
caddyhttp: remote_ip matcher, handling of zone identifiers

### LINKED ISSUE #4567: Remote_ip gets local ipv6 adresses with %eth0 (zone_id) but can't handle it
Hello together,
I run caddy in a docker container on a raspberry pi. At the moment i run it in the host network, because in a bridge network I got the docker ip back and that I can’t use to restrict access from out of the local network.
My caddy version is: 2.4.6

Thats the relevant part of my Caddyfile:
```
pihole.{$DOMAIN}:443 {
  tls {$EMAIL}
  @internalv6 {
    remote_ip fe80::/64
  }
  handle @internalv6 {
    uri replace / /admin/ 1
    reverse_proxy 192.168.178.3:81
  }
```
If I use it, I get this error back. Nothing else in the log in debug mode from interesst:
```
{"level":"error","ts":1644497213.9798095,"logger":"http.matchers.remote_ip","msg":"getting client IP","error":"invalid client IP address: fe80::9457:35e3:3b21:173e%eth0"}
```

Seems like caddy can't handle zone ids, but it is a normal behavior in `Scoped link local IPv6 addresses`. I found no way to disable the `zone_id`.
I have to mention, that I adressed the local adresses of the raspberry in my homenetwork with pihole, otherwise I accessed caddy with the global ipv6, what obviosly not works.

BitWuehler:
Fixes #4567

I implemented now the handling of zone identifiers. 
Here are my answers from my first pull request to @mholt:
```
At first I have to say, that I am not an expert in IPv6 too.

As far as my research shows, I am pretty shure that zone_IDs only play a role in link-local IPv6-adresses. 
The RFC 6874 document says:


The syntax defined above allows a ZoneID to be added to any IPv6
address.  The 6man WG discussed and rejected an alternative in which
the existing syntax of IPv6address would be extended by an option to
add the ZoneID only for the case of link-local addresses.  It was
felt that the solution presented in this document offers more
flexibility for future uses and is more straightforward to implement.


So for the moment it's local but maybe in the future other IPv6-Adresses will use it too.

To the security point:
An example for the use of the zone_id is a server with serveral subnets. Because link-local adresses uses all only
the fe80::/64prefix, it could be, that if you clone a machine or very rare randomly, you could have the same 
link-local IPv6-adress on different machines on different subnets. The zone_id helps to speak only to the right 
machine in that case.
More about that: https://4sysops.com/archives/ipv6-tutorial-part-7-zone-id-and-unique-local-ipv6-unicast-addresses/
So, in terms of security, I would not say, that this is not a problem, because it appears only in local networks.

And sure, if anyone reports a bug, I will give my best to help to fix it.
Thank you for your work. I love the caddy project!
```
Of course, the thank you also goes to everyone else who is involved in this wonderful project!

This time I managed to build caddy on my PC and ich checked everthing, what I think could happen. So far it seem everything works well.

Ipv6 Adresses with Range and Filter should be written like this: ``fe80::/64%eth0``

If anyone finds more efficient ways for some parts, feel free to show me. I am new to Go (and actually also programming) so I would love to learn better ways!


modules/caddyhttp/matchers.go
```diff
@@ -909,25 +914,51 @@ func (m MatchRemoteIP) getClientIP(r *http.Request) (net.IP, error) {
 	if err != nil {
 		ipStr = remote // OK; probably didn't have a port
 	}
+	// Some IPv6-Adresses can contain zone identifiers at the end,
+	// which are separated with "%"
+	if strings.Contains(ipStr, "%") {
+		zone_id = strings.Split(ipStr, "%")[1]
+		ipStr = strings.Split(ipStr, "%")[0]
```
francislavoie, line 921: Don't need to run split twice

```suggestion
		split := strings.Split(ipStr, "%")
		ipStr = split[0]
		zoneID = split[1]
```

modules/caddyhttp/matchers.go
```diff
@@ -909,25 +914,51 @@ func (m MatchRemoteIP) getClientIP(r *http.Request) (net.IP, error) {
...
+				// Assign true or false for the handling of multible ips with or without zone_ids in Filter
+				// and handling of the "zone_id not match" info.
+				if zone_id_def == zone_id_got {
+					return true
+				} else {
+					zoneFilter = false
+				}
+			} else {
+				return true
+			}
```
francislavoie, line 956: In general, when you have an `else` that only ever returns, it's usually a signal that you should invert the condition and return early. So instead of this:

```go
if strings.Contains(zone_id_def, "%") {
	...
} else {
	return true
}
```

You should probably do

```go
if !strings.Contains(zone_id_def, "%") {
	return true
}

...
```

That way the code stays "flatter" (less indentation levels) and you get rid of the `else` altogether.

modules/caddyhttp/matchers.go
```diff
@@ -909,25 +914,51 @@ func (m MatchRemoteIP) getClientIP(r *http.Request) (net.IP, error) {
...
 
 // Match returns true if r matches m.
 func (m MatchRemoteIP) Match(r *http.Request) bool {
-	clientIP, err := m.getClientIP(r)
+	clientIP, zone_id_got, err := m.getClientIP(r)
+	zoneFilter := true
 	if err != nil {
 		m.logger.Error("getting client IP", zap.Error(err))
 		return false
 	}
```
francislavoie, line 937: This variable relates to the loop, not to the `if err != nil`, so it should go after.

```suggestion
	if err != nil {
		m.logger.Error("getting client IP", zap.Error(err))
		return false
	}
	zoneFilter := true
```

modules/caddyhttp/matchers.go
```diff
@@ -909,25 +914,51 @@ func (m MatchRemoteIP) getClientIP(r *http.Request) (net.IP, error) {
...
+					zoneFilter = false
+				}
+			} else {
+				return true
+			}
 		}
 	}
+	if zoneFilter == false {
+		m.logger.Info("zone_id from remote Client, don't matches that one in remote_ip filter.")
+	}
```
francislavoie, line 961: I don't think we need this log. It might be pretty noisy. It doesn't tell the user anything that they couldn't figure out from looking at the access logs. We _could_ log it at `Debug` level which would make it possible to reveal if they're trying to figure out if their matcher is broken, though.

Also the wording is somewhat awkward, I might write it like this:

```suggestion
	if !zoneFilter {
		m.logger.Debug("zone ID from remote '%s' did not match", zoneID)
	}
```

francislavoie (changes requested):

modules/caddyhttp/matchers.go
```diff
@@ -877,10 +877,14 @@ func (m *MatchRemoteIP) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
 	for _, str := range m.Ranges {
+		// Exclude the zone_id from the IP
+		if strings.Contains(str, "%") {
+			str = strings.Split(str, "%")[0]
+		}
 		if strings.Contains(str, "/") {
 			_, ipNet, err := net.ParseCIDR(str)
 			if err != nil {
-				return fmt.Errorf("parsing CIDR expression: %v", err)
+				return fmt.Errorf("parsing CIDR expression: %v for IP %s", err, str)
```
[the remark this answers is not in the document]
BitWuehler: Good to know!

modules/caddyhttp/matchers.go
BitWuehler: Ahh. I asked myself how it can be done this way. Nice!

modules/caddyhttp/matchers.go
BitWuehler: makes sense

modules/caddyhttp/matchers.go
BitWuehler: The first park makes sense, but I can't change the error text itself.
```
cannot use zoneID (type string) as type zapcore.Field in argument to m.logger.Debug
```

francislavoie: Oh right, that's a zap logger and not fmt, need to do this instead: `m.logger.Debug("zone ID from remote did not match", zap.String("zone", zoneID))`

BitWuehler:
Thanks @francislavoie for your nice tips. Learned a few things!

francislavoie (changes requested):

modules/caddyhttp/matchers.go
```diff
@@ -909,25 +923,47 @@ func (m MatchRemoteIP) getClientIP(r *http.Request) (net.IP, error) {
...
+			if m.zones[i] == "" {
+				return true
+			}
+			// Check if the incomming zoneID is that one in filter.
+			// Assign true or false for the handling of multible ips with or without zoneIDs in Filter
+			// and handling of the "zoneID not match" debug info.
+			if zoneID == m.zones[i] {
+				zoneFilter = true
+				return true
+			}
```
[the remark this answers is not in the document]
BitWuehler: It has to, because it could be that there is an IP with zone before, that already set the zoneFilter to false. If now an other IP with the right zone matches, is has to set the filter to true. Otherwise the debug info later would be given out wrongly.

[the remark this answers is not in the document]
francislavoie: But `return true` means that it won't continue at all, it'll exit the function right then.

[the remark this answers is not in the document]
francislavoie: Also with this change, the above condition and this one could be merged to just `if m.zones[i] == "" || zoneID == m.zones[i] { return true`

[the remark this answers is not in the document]
BitWuehler: > But `return true` means that it won't continue at all, it'll exit the function right then.

The ``for`` loop runs for every ip in every ``remote_ip`` filter. Only serveral different remote_ip filters (with one or more ips and ranges) are separated from each other. Without the setting of ``true`` it will stay ``wrong``, even if thas isn't right.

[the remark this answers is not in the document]
BitWuehler: To ne honest, im sure it's a very rare usecase and it's also not terrible, if you get a wrong debug message, but yeah, it's not correct...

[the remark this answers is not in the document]
BitWuehler: Okay. You are right! ;D

modules/caddyhttp/matchers.go
```diff
@@ -909,25 +922,40 @@ func (m MatchRemoteIP) getClientIP(r *http.Request) (net.IP, error) {
...
-	for _, ipRange := range m.cidrs {
+	zoneFilter := true
+	for i, ipRange := range m.cidrs {
 		if ipRange.Contains(clientIP) {
-			return true
+			// Check if there are zone filters assigned and if they match.
+			if m.zones[i] == "" || zoneID == m.zones[i] {
+				return true
+			}
+			zoneFilter = false
```
francislavoie, line 953: I think this needs to be wrapped in a `if m.zones[i] != ""` because otherwise the log message will be written even if none of the matchers tried to match for a zone

francislavoie (changes requested):
Almost there! Looking good now, I think

modules/caddyhttp/matchers.go
BitWuehler: oh yes

BitWuehler: Do you really think this is necessary? If its a real IP with zone id and it does not match, what else could it be?

BitWuehler: Okay. I tested it with multible matchers without zone id. Got not the debug info.

BitWuehler: okay. you are right again ^^

BitWuehler: Checked it again. It seems to work without the ``if m.zones[i] != ""``.
It shows no debug info, if I have no ips with a zone. If I use a wrong one with matching IP, it shows the info.

francislavoie: Yeah, you're right, sorry my brain made a silly :sweat_smile:

francislavoie (approved):
Looking good!

I'm sure Matt will have comments as well once he has the time to look at this.

modules/caddyhttp/matchers.go
```diff
@@ -131,7 +131,10 @@ type (
 		// to spoof request headers. Default: false
 		Forwarded bool `json:"forwarded,omitempty"`
 
+		// cidrs and zones vars should aligned always in the same
```
francislavoie, line 134: Hmm, looks like gofmt isn't happy with this. Please run `gofmt -s -w modules/caddyhttp/matchers.go` to fix it up

francislavoie:
For bonus points, if you'd like to, you could add a test to `matchers_test.go` for this.

BitWuehler:
@francislavoie I now don't know how but tommorow I can gladly try that, but now it's to late. Thanks again!

francislavoie:
You can run the tests with `go test -short ./...` from the root of the project. The `-short` flag skips integration tests which run instances of Caddy and bind to ports etc; it can mess up your running instance of Caddy if you already had one running 😬 so it's best to use `-short`, especially since these tests are just unit tests

mholt (approved):
Hey, good job. The comments are helpful and I think this approach will work for what we need. We can always improve it if people report bugs/problems. Hopefully you can help us out if that happens. :slightly_smiling_face:  Thank you!

PR #4603, opened 2022-02-26
Prevent unnecessary loading of config

YourTechBud:
Currently the config returned by `caddy.config_loaders` is applied even though there is no change in the config. This results in a lot of wasted cpu cycles to restart all the apps and modules.

I've added a basic optimization step to prevent unnecessary reloads.

It would be great if you could add this to the next release.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -481,11 +481,23 @@ func finishSettingUp(ctx Context, cfg *Config) error {
 			return fmt.Errorf("loading config loader module: %s", err)
 		}
 		runLoadedConfig := func(config []byte) {
-			Log().Info("applying dynamically-loaded config", zap.String("loader_module", val.(Module).CaddyModule().ID.Name()), zap.Int("pull_interval", int(cfg.Admin.Config.LoadInterval)))
 			currentCfgMu.Lock()
+			defer currentCfgMu.Unlock()
+
+			// Skip if there is no change in the config
+			if bytes.Equal(rawCfgJSON, config) {
+				return
+			}
+
+			Log().Info("applying dynamically-loaded config", zap.String("loader_module", val.(Module).CaddyModule().ID.Name()), zap.Int("pull_interval", int(cfg.Admin.Config.LoadInterval)))
 			err := unsyncedDecodeAndRun(config, false)
-			currentCfgMu.Unlock()
 			if err == nil {
+				// success, so update our stored copy of the encoded
+				// config to keep it consistent with what caddy is now
+				// running (storing an encoded copy is not strictly
+				// necessary, but avoids an extra json.Marshal for
+				// each config change)
+				rawCfgJSON = config
 				Log().Info("dynamically-loaded config applied successfully")
 			} else {
 				Log().Error("running dynamically-loaded config failed", zap.Error(err))
@@ -503,6 +515,7 @@ func finishSettingUp(ctx Context, cfg *Config) error {
 							Log().Error("loading dynamic config failed", zap.Error(err))
 							return
 						}
+
 						runLoadedConfig(loadedConfig)
 					case <-ctx.Done():
 						if !timer.Stop() {


mholt (approved):
Thanks for this. I will merge this and then might tweak a few things (logging, etc).

I am worried if someone wants to forcefully reload the same config on an interval; I believe this makes it impossible (without hacks). The API endpoints have cache control headers that can force a reload, but this isn't an HTTP interface so I'm not sure how we might go about doing that.

Maybe we'll wait and see if the feature request rolls in, and if so, deal with it then. This is probably better for more naive config loading systems in general.

mholt:
Follow-up commit here, simplified things by reusing code: acbee9470818a32f77634b0c0943bc0544f6e736

mholt:
Me too, but you never know :P

RussellLuo:
I'm afraid that using `changeConfig` directly — although also the first solution came up into my mind — will break the following assumption, will it?

https://github.com/caddyserver/caddy/blob/ac14b64e08a6ea63067c62cf1bad9cd6ad823d60/caddy.go#L264-L266

mholt:
@RussellLuo Ah yeah, dang... good catch.

Might need to pass through one more little bit of state...

mholt:
@RussellLuo Can you expand on that a bit?

RussellLuo:
According to my understanding, I assume that it was a pulled config if dynamic loading was enabled in `currentCfg` (the initial loading will be excluded, in which case `currentCfg` is nil). The corresponding code may look like this:

```go
func unsyncedDecodeAndRun(cfgJSON []byte, allowPersist bool) error {
	//...

	dynamicLoadEnabled := func(cfg *Config) bool {
		return cfg != nil &&
			cfg.Admin != nil &&
			cfg.Admin.Config != nil &&
			cfg.Admin.Config.LoadRaw != nil
	}

	// prevent recursive config loads; that is a user error, and
	// although frequent config loads should be safe, we cannot
	// guarantee that in the presence of third party plugins, nor
	// do we want this error to go unnoticed (we assume it was a
	// pulled config if dynamic loading was enabled in currentCfg)
	if dynamicLoadEnabled(currentCfg) &&
		dynamicLoadEnabled(newCfg) &&
		newCfg.Admin.Config.LoadDelay <= 0 {
		return fmt.Errorf("recursive config loading detected: pulled configs cannot pull other configs without positive load_delay")
	}

	//...
}
```

RussellLuo:
While trying to verify the above changes, I think I found another issue 🤔 — the dynamic reloading loop, hopefully achieved by enabling dynamic loading in the new config, will be broken once no config change is detected here:  

https://github.com/caddyserver/caddy/blob/6b385a36f9750b7a66300bbb2167ea5a4d26a61f/caddy.go#L149-L153

ChrisLahaye:
> While trying to verify the above changes, I think I found another issue 🤔 — the dynamic reloading loop, hopefully achieved by enabling dynamic loading in the new config, will be broken once no config change is detected here:  
> 
> https://github.com/caddyserver/caddy/blob/6b385a36f9750b7a66300bbb2167ea5a4d26a61f/caddy.go#L149-L153

This is an issue for us as well. We are generating our config on a separate backend which is being loaded periodically by the caddy instances. We prefer a pull- instead of push mechanism. I believe that with this change it will stop pulling after two pulls of the same config file?

mholt:
@RussellLuo 

> According to my understanding, I assume that it was a pulled config if dynamic loading was enabled in currentCfg (the initial loading will be excluded, in which case currentCfg is nil).

That might indeed be true... would you want to try out your implementation and submit a PR? I think that'd be a cleaner solution than what I was thinking.

> While trying to verify the above changes, I think I found another issue thinking — the dynamic reloading loop, hopefully achieved by enabling dynamic loading in the new config, will be broken once no config change is detected here:

Yeah, so between that and the error handling that @ChrisLahaye mentioned, I wonder if I should restore the loop I took out from [here](https://github.com/caddyserver/caddy/commit/acbee9470818a32f77634b0c0943bc0544f6e736#diff-57caf2be698e920a2599b611efd1b33dea35fd79976cacdf3354ab3f55706515L508) and make sure it only iterates if there is an error loading the config OR if the pulled config is unchanged (or nil with nil error?).

I'll look into that.

mholt:
Okie dokie, @RussellLuo and @ChrisLahaye , I think I have improved the situation a bit in ceef70d -- the delayed load is now in a loop which is iterated only on error or no-op (nil config value returned, nil error). I haven't addressed the "config unchanged" problem yet. I dunno if maybe @RussellLuo you want to take a stab at that?

RussellLuo:
@mholt Great! On the basis of your improvements, I think there is a possible way to address the "config unchanged" problem — introducing a special error such as `ErrConfigUnmodified`, inspired by the [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) header, to indicate there is no change in the config. As a result, the loop will continue (maybe the error logging needs to be adjusted):

https://github.com/caddyserver/caddy/blob/a6199cf814d676a11cc9e2abad81b48859198aa4/caddy.go#L513-L517

I think one more benefit of this solution is that the bandwidth can also be saved if there is no config change at all.

RussellLuo:
> According to my understanding, I assume that it was a pulled config if dynamic loading was enabled in currentCfg (the initial loading will be excluded, in which case currentCfg is nil).

As for the solution, it didn't work since `currentCfg.Admin.Config.LoadRaw` has been reset during the previous loading:

https://github.com/caddyserver/caddy/blob/a6199cf814d676a11cc9e2abad81b48859198aa4/caddy.go#L486

As a workaround, we can set it to an empty byte slice (non-nil) again, which may be less elegant.

mholt:
Ah, right. Forgot about that. I like the special error return value idea though. That's a good plan.

mholt:
@RussellLuo Ok, I've pushed a72acd2 which should handle the case of config unchanged according to your idea. I think it will just wait again and retry in that case.

RussellLuo:
Ah, thanks for the effort, Matt!

My initial thought was to return the special error from the `LoadConfig` method:

https://github.com/caddyserver/caddy/blob/a72acd21b0ef17d6cd9064dd105042b6edb3b8dc/caddy.go#L560-L566

I think this may have two benefits: 1) the config backend can return [304](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) instead of the full config, thus could save bandwidth; 2) whether to force a reload can be controlled by users themselves. In exchange, the [HTTPLoader](https://github.com/caddyserver/caddy/blob/master/caddyconfig/httploader.go#L35) and the config backend need to implement [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since)-like semantics.

Anyway, I think the current implementation is good, and I have just verified that it works as expected 👍

ChrisLahaye:
> [a72acd2](https://github.com/caddyserver/caddy/commit/a72acd21b0ef17d6cd9064dd105042b6edb3b8dc)

This is great, thank you!       

I see that LoadInterval has been renamed to LoadDelay. I think that with the new functionality, the old name is more applicable as now it really is an interval, might want to reverse the name change.

RussellLuo:
Per the comments:

https://github.com/caddyserver/caddy/blob/a72acd21b0ef17d6cd9064dd105042b6edb3b8dc/caddy.go#L517-L518

I think `LoadDelay` is still appropriate for successful loading, which will only happen once. For other cases described in the comments, I guess we just reuse `LoadDelay` as an interval for retry.

mholt:
@RussellLuo 

> My initial thought was to return the special error from the LoadConfig method:

Ah, I see. That's not a bad idea I guess. Want to submit a PR?

@ChrisLahaye 

> I see that LoadInterval has been renamed to LoadDelay. I think that with the new functionality, the old name is more applicable as now it really is an interval, might want to reverse the name change.

I think LoadDelay is more correct. It only loads the config once -- not on an interval. The "interval" is only if there's an error and we need to retry.

ChrisLahaye:
> @RussellLuo Ok, I've pushed a72acd2 which should handle the case of config unchanged according to your idea. I think it will just wait again and retry in that case.

Apologies, this to me suggests that the config loading is actually done with an interval if there is an error or the config remains unchanged (most common case).

RussellLuo:
@mholt I have submitted a PR (#4615), since I didn't see your comment changes. Sorry for that, and I think it's OK if this is not the feature Caddy want to provide currently.

mholt:
@RussellLuo Ah, no worries, I guess GitHub doesn't notify on edits even when tagged in a comment...

I just try to avoid adding so much complexity unless there's enough legitimate use cases (feature requests) to justify it. I'll give it a look anyway! Thanks for putting in the effort.

PR #4606, opened 2022-03-01
caddyhttp: Honor wildcard hosts in log SkipHosts

### LINKED ISSUE #4256: Erratic logging behavior with wildcard domains

mholt:
Hoping this fixes #4256.

@ledlamp I just whipped this up real quick. Can you test to confirm this fixes the issue for you?

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
+	"github.com/caddyserver/certmagic"
 	"github.com/lucas-clemente/quic-go/http3"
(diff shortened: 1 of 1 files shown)


francislavoie:
Oh, that seemed easy, didn't know we could just reuse certmagic's function for that. Nice.

francislavoie (approved):

mholt:
@ledlamp appears to have deleted their account so I'm just going to merge this. 😛

PR #4613, opened 2022-03-05
fileserver: Add `pass_thru` Caddyfile option

francislavoie:
https://caddy.community/t/serve-available-static-content-and-proxy-otherwise/15289/4

I guess it was just forgotten 🤷‍♂️


mholt (approved):
Thanks!

In practice, the `file_server` directive will probably need to be in a `route` directive (or ordered using the `order` global option) to be useful with this mode.

francislavoie:
Good point, we should note that in the docs.

PR #4615, opened 2022-03-05
caddyconfig: Support conditional requests in HTTP loader

RussellLuo:
For how conditional requests work, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -938,7 +938,7 @@ func handleConfig(w http.ResponseWriter, r *http.Request) error {
 		forceReload := r.Header.Get("Cache-Control") == "must-revalidate"
 
 		err := changeConfig(r.Method, r.URL.Path, body, forceReload)
-		if err != nil && !errors.Is(err, errSameConfig) {
+		if err != nil && !errors.Is(err, ErrConfigUnchanged) {
 			return err
 		}
 
--- a/caddy.go
+++ b/caddy.go
@@ -112,7 +112,7 @@ func Load(cfgJSON []byte, forceReload bool) error {
 	}()
 
 	err := changeConfig(http.MethodPost, "/"+rawConfigKey, cfgJSON, forceReload)
-	if errors.Is(err, errSameConfig) {
+	if errors.Is(err, ErrConfigUnchanged) {
 		err = nil // not really an error
 	}
 	return err
@@ -123,7 +123,7 @@ func Load(cfgJSON []byte, forceReload bool) error {
 // the new value (if applicable; i.e. "DELETE" doesn't have an input).
 // If the resulting config is the same as the previous, no reload will
 // occur unless forceReload is true. If the config is unchanged and not
-// forcefully reloaded, then errConfigUnchanged This function is safe for
+// forcefully reloaded, then ErrConfigUnchanged. This function is safe for
 // concurrent use.
 func changeConfig(method, path string, input []byte, forceReload bool) error {
 	switch method {
@@ -155,7 +155,7 @@ func changeConfig(method, path string, input []byte, forceReload bool) error {
 	// if nothing changed, no need to do a whole reload unless the client forces it
 	if !forceReload && bytes.Equal(rawCfgJSON, newCfg) {
 		Log().Info("config is unchanged")
-		return errSameConfig
+		return ErrConfigUnchanged
 	}
 
 	// find any IDs in this config and index them
@@ -500,11 +500,7 @@ func finishSettingUp(ctx Context, cfg *Config) error {
 
 		runLoadedConfig := func(config []byte) error {
 			logger.Info("applying dynamically-loaded config")
-			err := changeConfig(http.MethodPost, "/"+rawConfigKey, config, false)
-			if errors.Is(err, errSameConfig) {
-				return err
-			}
-			if err != nil {
+			if err := changeConfig(http.MethodPost, "/"+rawConfigKey, config, true); err != nil {
 				logger.Error("failed to run dynamically-loaded config", zap.Error(err))
 				return err
 			}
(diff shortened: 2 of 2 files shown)


RussellLuo:
This PR just implements conditional requests described here https://github.com/caddyserver/caddy/pull/4603#issuecomment-1058850899.

I think there are some remaining TODOs:
1. I guess `errSameConfig and` and `ErrConfigUnmodified` should be merged into one.
2. For users to force a reload (i.e. send the full config without regard to the `If-Modified-Since` header), we should change `forceReload` to true here:

https://github.com/caddyserver/caddy/blob/09ba9e994e6391162fbfdfbd90c72270bdff3a49/caddy.go#L503

mholt:
Thanks for the PR! Sorry it's taking me a while to get around to this. I need to give this some thought.

RussellLuo:
Never mind :)

Actually before Caddy supports dynamic config loading, I have been using the `If-Modified-Since`-like mechanism in my little [custom reloader](https://github.com/RussellLuo/olaf/blob/84b4f417013e42436adf2f35a714de6bef61fdcf/caddyconfig/reloader/reloader.go#L17-L23) for a while. Although it makes implementations a little bit more complex, I think one benefit is that it's less expensive for us to choose a smaller `load_delay` to make Caddy react faster to remote config changes.

That being said, thanks to the great extensibility of Caddy, we can still write our own implementations of [ConfigLoader](https://github.com/caddyserver/caddy/blob/93c99f67342504efe9f6b58a734aaec3929fe785/caddy.go#L564) if this feature is not officially supported by Caddy. The only drawback of the custom `ConfigLoader` is that the error message will be misleading, in case that the loader returns an error to indicate that there is no config change:

https://github.com/caddyserver/caddy/blob/93c99f67342504efe9f6b58a734aaec3929fe785/caddy.go#L523-L527

mholt:
@RussellLuo Good point!

It's mostly that I'm a little iffy on abstraction and stuff, so maybe this feature isn't a great fit for what we ship standard with Caddy, for now.

But I like the idea of it being a separate module for those who need this feature, like an enhanced HTTP loader or something.

As for the error message, we could either reword it, or maybe, would a slightly different signature for LoadConfig() be helpful? Like a boolean return value that indicates whether there is any config to load yet, or no change to signature but rather a sentinel error (like this PR introduces) that can indicate there was no error but there's no config yet either.

What do you think is better?

RussellLuo:
> It's mostly that I'm a little iffy on abstraction and stuff, so maybe this feature isn't a great fit for what we ship standard with Caddy, for now.
>
> But I like the idea of it being a separate module for those who need this feature, like an enhanced HTTP loader or something.

Yeah, I think this is a valid concern. Let's keep Caddy as simple as possible ❤️ 

> As for the error message, we could either reword it, or maybe, would a slightly different signature for LoadConfig() be helpful? Like a boolean return value that indicates whether there is any config to load yet, or no change to signature but rather a sentinel error (like this PR introduces) that can indicate there was no error but there's no config yet either.

Personally, I prefer the approach of using a sentinel error (much like [io.EOF](https://pkg.go.dev/io#pkg-variables)), which seems more conventional and thus keeps the signature for LoadConfig() more concise.

mholt:
Ok, I like the sentinel error value too.

Let's maybe just add that with this PR, and leave the more complex HTTP loading logic to a separate plugin for now. If it becomes popular etc, maybe we can look at bringing it in.

Thanks for iterating on this!

RussellLuo:
Great, I'll update this PR. Before I start, I think there are two things I'm not sure of:

- Should we merge `errSameConfig` and `ErrConfigUnmodified` into one (exported) error?
- Is the name `ErrConfigUnmodified` (derived from `If-Modified-Since`) OK? Or maybe `ErrConfigUnchanged` is more consistent with the existing naming convention (like `changeConfig`)?

caddyconfig/httploader.go
```diff
@@ -116,10 +130,22 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
...
 
+	// Upon successful response, save the new time of last modification, if any.
+	lastModified := resp.Header.Get("Last-Modified")
+	if lastModified != "" {
+		t, err := time.Parse(timeLayout, lastModified)
+		if err != nil {
+			ctx.Logger(hl).Debug("failed to parse Last-Modified header", zap.Error(err))
+		}
+		if !t.IsZero() {
+			hl.lastModified = t
```
YourTechBud, line 141: I was just wondering, does storing the lastModified field in h1 serve any purpose?
Caddy creates a new configloader instance on every change detected. So the value we store here will be lost as soon as we return a new result.

RussellLuo: > Caddy creates a new configloader instance on every change detected.

Yeah, that is the case.

The `lastModified` field — initially being zero — will typically be saved for the first polling (just `load_delay` time later after the associated loader instance was created). Then in the subsequent queries, the `If-Modified-Since` header can be filled in correctly.

Of course, when the loader finally detects a config change (i.e. receives a 200 response for the second time), the last save will be lost as you described.

RussellLuo:
> Yes, probably best to get rid fo errSameConfig entirely

@mholt Actually, I've been thinking about this too. One way to achieve this is to use `forceReload=true` when calling `changeConfig`. In other words, this gives the ability (as well as the responsibility) for "forced reloading" to `ConfigLoader` implementors.

As a consequence of this change, in the current `If-Modified-Since`-based implementation, I think the `lastModified` field should be changed to a global variable to share the same memory across different loader instances as @YourTechBud described.

RussellLuo:
> Yes, probably best to get rid fo errSameConfig entirely

@mholt Actually, I've been thinking about this too. One way to achieve this is to use `forceReload=true` when calling `changeConfig`. In other words, this gives the ability (as well as the responsibility) for "forced reloading" to `ConfigLoader` implementors.

As a consequence of this change, in the current `If-Modified-Since`-based implementation, I think the `lastModified` field should be changed to a global variable to share the same memory across different loader instances as @YourTechBud described.

RussellLuo:
Note that this PR has been closed unexpectedly, after I renamed the original branch. See #4658 instead.

PR #4616, opened 2022-03-05
reverseproxy: Refactor dial address parsing, augment command parsing

### LINKED ISSUE #4611: Feature request: Support sockets in reverse-proxy cli
Reverse proxy support sockets in its configuration but not when used in commandline:

```
$ caddy reverse-proxy -from sandbox-rabbitmq -to unix//tmp/sandbox-rabbitmq.sock
reverse-proxy: paths are not allowed: unix//tmp/sandbox-rabbitmq.sock
```

Would be amazing to support it in commandline too.

francislavoie:
Fix #4611

Quick refactor to split out the dial address parsing from the Caddyfile code to make it usable for the `reverse-proxy` command as well.

Also added a new helper method in the Dispenser because it's slightly cleaner when you already have an error from lower level parsing code that you want to throw as a parsing error.

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -350,7 +350,11 @@ func (d *Dispenser) Err(msg string) error {
 
 // Errf is like Err, but for formatted error messages
 func (d *Dispenser) Errf(format string, args ...interface{}) error {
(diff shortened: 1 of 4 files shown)


mholt (approved):
LGTM, thanks for the enhancement!

PR #4629, opened 2022-03-11
caddyhttp: retain all values of vars matcher when specified multiple times

### LINKED ISSUE #4625: Caddyfile `vars` matcher only keeps the last configured values
If the `vars` matcher is specified multiple times, only the last one's values will be retained. e.g.

```
vars foo a b c d e
vars foo 1 2 3 4 5
```

Only values `1 2 3 4 5` will be kept.

Simple fix, we just need to `append()` instead, to keep all of `a b c d e 1 2 3 4 5`

I'm feeling too lazy to write the PR right now, so I'm logging the issue for now.

mohammed90:
Fixes #4625


francislavoie (approved):
Perfect, thanks!

mholt:
Thank you :pray:

PR #4637, opened 2022-03-15
ci: Build on Go 1.18, bump actions versions

francislavoie:


mohammed90:
We should stick to go1.17 until [github.com/lucas-clemente/quic-go](https://github.com/lucas-clemente/quic-go) is compatible with go1.18.

mohammed90 (approved):
Thanks, Francis! Let's merge 🔥

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        go: [ '1.16', '1.17' ]
+        go: [ '1.16', '1.17', '1.18' ]
```
mholt, line 22: We can remove the tests for Go 1.16. We only support the most recent 2 releases of Go.

duration_fuzz.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build gofuzz
 // +build gofuzz
```
mholt, line 16: What's with both of these? Seems redundant...

.github/workflows/ci.yml
francislavoie: We're keeping 1.16 minimum until COPR can use 1.17 on all their supported platforms

duration_fuzz.go
francislavoie: See the 1.18 release notes, they're migrating from one syntax to the other, keeping both for now for compat. This is from running gofmt, I didn't add these by hand

.github/workflows/ci.yml
mholt: These are just CI tests though, not releases, right? I don't think we need to triple our CI tests for COPRs. (I really do not like the idea of external distribution platforms holding us back. The added CI tests do slow things down.)

duration_fuzz.go
mholt: Oh! TIL. Very cool.

I installed Go 1.18 today but haven't done much with it yet.

.github/workflows/ci.yml
francislavoie: It doesn't slow things down, the tests are run in parallel.

PR #4643, opened 2022-03-18
caddyfile: Support for raw token values, improve `map`, `expression`

francislavoie:
Followup to https://github.com/caddyserver/caddy/commit/93c99f67342504efe9f6b58a734aaec3929fe785 as pointed out in https://github.com/caddyserver/website/issues/221, and fixing the issue mentioned in https://github.com/caddyserver/website/issues/220

See the adapt tests, which cover how it behaves.

Note: This _is_ a breaking change, because people using `map` with numbers used to get strings, and will now get numbers.

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"strconv"
 	"strings"
 )
 
@@ -201,6 +202,43 @@ func (d *Dispenser) Val() string {
 	return d.tokens[d.cursor].Text
 }
 
+// RawVal gets the raw text of the current token (including quotes).
+// If there is no token loaded, it returns empty string.
+func (d *Dispenser) RawVal() string {
+	if d.cursor < 0 || d.cursor >= len(d.tokens) {
+		return ""
+	}
+	quote := d.tokens[d.cursor].wasQuoted
+	if quote > 0 {
+		return string(quote) + d.tokens[d.cursor].Text + string(quote)
+	}
+	return d.tokens[d.cursor].Text
+}
+
+// ScalarVal gets value of the current token, converted to the
+// appropriate scalar type. If there is no token loaded, it returns nil.
+func (d *Dispenser) ScalarVal() interface{} {
+	if d.cursor < 0 || d.cursor >= len(d.tokens) {
+		return nil
+	}
+	quote := d.tokens[d.cursor].wasQuoted
+	text := d.tokens[d.cursor].Text
+
+	if quote > 0 {
+		return text
+	}
+	if num, err := strconv.Atoi(text); err == nil {
+		return num
+	}
+	if num, err := strconv.ParseFloat(text, 64); err == nil {
+		return num
+	}
+	if bool, err := strconv.ParseBool(text); err == nil {
+		return bool
+	}
+	return text
+}
+
 // Line gets the line number of the current token.
 // If there is no token loaded, it returns 0.
 func (d *Dispenser) Line() int {
@@ -249,6 +287,19 @@ func (d *Dispenser) AllArgs(targets ...*string) bool {
 	return true
 }
 
+// CountRemainingArgs counts the amount of remaining arguments
+// (tokens on the same line) without consuming the tokens.
+func (d *Dispenser) CountRemainingArgs() int {
+	count := 0
+	for d.NextArg() {
+		count++
+	}
+	for i := 0; i < count; i++ {
+		d.Prev()
+	}
+	return count
+}
+
 // RemainingArgs loads any more arguments (tokens on the same line)
 // into a slice and returns them. Open curly brace tokens also indicate
 // the end of arguments, and the curly brace is not included in
@@ -257,6 +308,18 @@ func (d *Dispenser) RemainingArgs() []string {
 	var args []string
 	for d.NextArg() {
 		args = append(args, d.Val())
+	}
+	return args
+}
+
+// RawRemainingArgs loads any more arguments (tokens on the same line,
+// retaining quotes) into a slice and returns them. Open curly brace
+// tokens also indicate the end of arguments, and the curly brace is
+// not included in the return value nor is it loaded.
+func (d *Dispenser) RawRemainingArgs() []string {
+	var args []string
+	for d.NextArg() {
+		args = append(args, d.RawVal())
 	}
 	return args
 }
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -210,13 +210,13 @@ func (d *Dispenser) RawVal() string {
 	}
 	quote := d.tokens[d.cursor].wasQuoted
 	if quote > 0 {
-		return string(quote) + d.tokens[d.cursor].Text + string(quote)
+		return string(quote) + d.tokens[d.cursor].Text + string(quote) // string literal
 	}
 	return d.tokens[d.cursor].Text
 }
 
-// ScalarVal gets value of the current token, converted to the
-// appropriate scalar type. If there is no token loaded, it returns nil.
+// ScalarVal gets value of the current token, converted to the closest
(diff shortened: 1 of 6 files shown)


mholt:
No way, Francis, you are a WIZARD. I look forward to reviewing this, hopefully tomorrow

caddyconfig/caddyfile/dispenser.go
```diff
@@ -201,6 +202,43 @@ func (d *Dispenser) Val() string {
...
+// ScalarVal gets value of the current token, converted to the
+// appropriate scalar type. If there is no token loaded, it returns nil.
+func (d *Dispenser) ScalarVal() interface{} {
+	if d.cursor < 0 || d.cursor >= len(d.tokens) {
+		return nil
+	}
+	quote := d.tokens[d.cursor].wasQuoted
+	text := d.tokens[d.cursor].Text
+
+	if quote > 0 {
```
mholt, line 227:
```suggestion
	if quote > 0 {
		// string literal
```
Just for people like me who get confused easily

caddyconfig/caddyfile/dispenser.go
```diff
@@ -261,6 +312,18 @@ func (d *Dispenser) RemainingArgs() []string {
 	return args
 }
 
+// RawRemainingArgs loads any more arguments (tokens on the same line,
+// retaining quotes) into a slice and returns them. Open curly brace
+// tokens also indicate the end of arguments, and the curly brace is
+// not included in the return value nor is it loaded.
+func (d *Dispenser) RawRemainingArgs() []string {
```
mholt, line 319: Probably a dumb nit, but what about `RemainingRawArgs()`? That keeps the two related functions together in the docs... or maybe `RemainingArgsRaw()` is more pedantically accurate.

modules/caddyhttp/celmatcher.go
```diff
@@ -150,7 +150,11 @@ func (m MatchExpression) Match(r *http.Request) bool {
 // UnmarshalCaddyfile implements caddyfile.Unmarshaler.
 func (m *MatchExpression) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	for d.Next() {
-		m.Expr = strings.Join(d.RemainingArgs(), " ")
+		if d.CountRemainingArgs() > 1 {
+			m.Expr = strings.Join(d.RawRemainingArgs(), " ")
```
mholt, line 154: Just curious why we need this condition as opposed to always joining the remaining args? If there's only one, shouldn't it just return it anyway?

caddyconfig/caddyfile/dispenser.go
```diff
@@ -201,6 +202,43 @@ func (d *Dispenser) Val() string {
...
+	quote := d.tokens[d.cursor].wasQuoted
+	if quote > 0 {
+		return string(quote) + d.tokens[d.cursor].Text + string(quote)
+	}
+	return d.tokens[d.cursor].Text
+}
+
+// ScalarVal gets value of the current token, converted to the
+// appropriate scalar type. If there is no token loaded, it returns nil.
+func (d *Dispenser) ScalarVal() interface{} {
```
mholt, line 220: Good name choice, but `TypedVal()` also came to mind. Not sure which is better. I'm fine with this, but wanted to ask anyway to see what you thought.

caddyconfig/caddyfile/lexer.go
```diff
@@ -38,6 +38,7 @@ type (
 		File        string
 		Line        int
 		Text        string
+		wasQuoted   rune
```
mholt, line 41:
```suggestion
		wasQuoted   rune // enclosing quote character, if any
```

caddyconfig/caddyfile/lexer.go
```diff
@@ -78,16 +79,17 @@ func (l *lexer) next() bool {
...
+		l.token.wasQuoted = quoted
 		return true
 	}
 
 	for {
 		ch, _, err := l.reader.ReadRune()
 		if err != nil {
 			if len(val) > 0 {
-				return makeToken()
+				return makeToken(-1)
```
mholt, line 92: This could probably be 0, right? No quote character is going to be delimited as 0x0.

mholt (changes requested):
Awesome, thanks. This is a neat improvement. Just a few nits/questions/ideas.

caddyconfig/caddyfile/dispenser.go
francislavoie: Should we add a method to Token like `WasQuoted() bool` ?

caddyconfig/caddyfile/dispenser.go
francislavoie: Yeah we can do Raw suffix, that works for me

modules/caddyhttp/celmatcher.go
francislavoie: Because if you do Raw when it was quoted as a single token then you end up with the value being quoted wrapped which isn't valid CEL (try that change and run the tests to see if you don't follow what I mean)

caddyconfig/caddyfile/dispenser.go
francislavoie: I think Typed would be better if we supported returning somekind of struct, but I think Scalar is clearer that only scalars will be returned (for now)

caddyconfig/caddyfile/lexer.go
francislavoie: I guess? I just googled to see what the right value is for "not a rune" and a StackOverflow post had -1 🤷‍♂️ I don't know what's best but I'm indifferent, can change if you prefer

caddyconfig/caddyfile/dispenser.go
mholt: I was actually thinking of something like `quote()` which returns the quote rune (or 0 if unquoted).

caddyconfig/caddyfile/lexer.go
mholt: Technically 0 is a valid character (null byte character), but it will never be a valid quotation mark or string encapsulator in the Caddyfile. So for convenience we can use the default zero-value (0) instead of -1 if that's easier.

caddyconfig/caddyfile/dispenser.go
mholt: Yeah, and that's what I was wondering too, if in the future we'd ever deserialize a string as a structured data type. But I can't see that happening in the near future nor do I have any idea why or what that would look like. So Scalar is fine... hopefully we don't have to change it.

caddyconfig/caddyfile/dispenser.go
francislavoie: Hmm, yeah, I guess if we change it to 0 then it would be falsey, right? So we wouldn't need `> 0`, so we probably don't need a method for it.

mholt: Sort of; Go is strongly typed. But yeah, it's conventional to check for the "zero value" to see if something is set. Since 0 is technically a valid rune (just as 0 is a valid integer), it can be hard to differentiate zero value from unset (which is why many code bases use pointers in those cases, since they're nil-able), but in our case 0 is not a valid quote rune so we know that 0 just means "not quoted".

caddyconfig/caddyfile/dispenser.go
```diff
@@ -201,6 +202,43 @@ func (d *Dispenser) Val() string {
 	return d.tokens[d.cursor].Text
 }
 
+// RawVal gets the raw text of the current token (including quotes).
+// If there is no token loaded, it returns empty string.
+func (d *Dispenser) RawVal() string {
```
francislavoie, line 207: Should we name this ValRaw for symmetry? 🤷‍♂️

francislavoie (commented):
LGTM

caddyconfig/caddyfile/dispenser.go
mholt: Hmm, yes I think that makes sense!

mholt (approved):

PR #4647, opened 2022-03-19
Reverse proxy fix: copy headers as-is, don't canonicalize implicitly

KnifeMaster007:
Just simple fix to avoid implicit canonicalization of headers. In my case, it was breaking services, which were using non-canonically named headers

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -1079,7 +1079,7 @@ func cloneRequest(origReq *http.Request) *http.Request {
 func copyHeader(dst, src http.Header) {
 	for k, vv := range src {
 		for _, v := range vv {
-			dst.Add(k, v)
+			dst[k] = append(dst[k], v)
 		}
 	}
 }


mholt:
Unfortunately, services that require case-sensitive headers are broken; as, [by specification, HTTP headers are to be consumed in a _case-insensitive_ manner](https://stackoverflow.com/a/41169947/1048862). If we are to cater to applications breaking specification, it will have to be very generalizable with compelling arguments.

Can you be more specific about what app is violating the specs?

francislavoie:
I agree with Matt here, I don't think this makes sense. The problem isn't with Caddy, it's with your application, which isn't compliant with the HTTP spec.

I'll close this for now.

KnifeMaster007:
@mholt, sorry for late reply

About my case: unfortunately, sometimes we had to deal with legacy proprietary software, which isn't implementing specs carefully.

In my opinion, reverse proxy should not modify original request headers at all(with exception of hop-by-hop headers), if it's not specified explicitly by the configuration. For example, nginx handles proxy_pass in such way.

So, I think, canonicazing request end-to-end headers is simply not the responsibility of reverse proxy. And this behaviour, sure, is not expected from the reverse proxy. 

If you comletely disagree with my arguments, I think, this behaviour of reverse_proxy at least should be documented.

mholt:
Emitting headers in compliance with the specificiation is _absolutely_ the job of the reverse proxy. The broken software should be fixed.

(I will make a note to mention that in the docs; I had assumed this should be obvious.)

KnifeMaster007:
> Gotcha. I understand what you're saying. However, I think emitting headers in compliance with the specificiation is _absolutely_ the job of the reverse proxy. The broken software should be fixed.
> 
> (I will make a note to mention that in the docs; I had assumed this should be obvious.)

> Gotcha. I understand what you're saying. However, I think emitting headers in compliance with the specificiation is _absolutely_ the job of the reverse proxy. The broken software should be fixed.
> 
> (I will make a note to mention that in the docs; I had assumed this should be obvious.)

I think, this may be quite common problem. I've found, that _getHeader_ routines in, for example, Ktor framework is case-sensitive. So, if developer uses non-canonical named headers(for example, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate, non-canonical form here mentioned in documentation and is being sent in such form from the most of client apps) - they will be broken.

I agree, that emitting headers in compliance with spec is job of the reverse proxy, but I also think, that in that case we're speaking about passing original request headers, not emitting new ones. And I'd like to mention another time, that other popular reverse proxies, like Nginx, handles this case in other way, than Caddy does.

I also absolutely agree, that developers should to comply with specifications, but sometimes we have to admit, that we live in world far from ideal. For my case, fixing software behind reverse proxy is out of my responsibility.

I like, how Caddy helped to solve some of my problems, so I'd like that it could be used by wider community. Even by those, who is unlucky to work with legacy software, which written not fully compliant with spec.

Anyhow, thanks for fast feedback)

mholt:
> So, if developer uses non-canonical named headers(for example, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate, non-canonical form here mentioned in documentation and is being sent in such form from the most of client apps) - they will be broken.

That's unfortunate that the app is ignoring/violating the specification. :( What does it do when it gets an HTTP/2 request, which requires lower-case for all letters in header fields?

For the casing of `WWW-Authenticate` specifically, that is Go's doing. You could file an issue with the Go project and make your case, but it'll likely get closed as a duplicate: https://github.com/golang/go/issues/45061

> in that case we're speaking about passing original request headers, not emitting new ones.

Caddy decodes and re-encodes the request anew, so we are in fact talking about emitting headers from Caddy, not merely passing them through. The entire request header is processed; only the body is passed through uninterpreted.

> I also absolutely agree, that developers should to comply with specifications, but sometimes we have to admit, that we live in world far from ideal. For my case, fixing software behind reverse proxy is out of my responsibility.
> 
> I like, how Caddy helped to solve some of my problems, so I'd like that it could be used by wider community. Even by those, who is unlucky to work with legacy software, which written not fully compliant with spec.

I am also afraid that if we break spec compliance for the sake of apps broken one way, we will break compatibility with apps broken another way (those which expect the canonicalized header field key casing).

Sorry, but I am just not convinced that making the proposed change is the right thing to do.

PR #4649, opened 2022-03-19
Optimize PrepareRequest and req.Context lookups

renbou:
Upon analyzing a profile of Caddy during benchmarking (with `h2load`), I found out that a large chunk of CPU time is eaten up by PrepareRequest, specifically the multiple context.WithValue being used there - up to 3% of CPU time (during the 30 second benchmark+profile). Additionally, lots of CPU time was being used up by `req.Context().Value` in various places, such as matching.

**Edit: creating a map with no hint does the same thing, so I have reverted this since it's useless**
Firstly, I've added a capacity argument to the two make(map[string]interface{}) located there - for VarsCtxKey and routeGroupCtxKey. By default Caddy adds only 2 vars to the context map of vars (VarsCtxKey), and by checking out the code it seemed like only 1-2 should be rarely added, which is why using 4 here makes sense (also using <=8 allows Go to use `runtime.makemap_small` which is **significantly** faster than `runtime.makemap`). As for routeGroupCtxKey - I can't imagine using more than 4 route groups, especially since they can't be set in a Caddyfile except implicitly using try_files.

Secondly, I have changed the various context keys to their own types with `int32` as the base type. This is what gave the largest boost to performance:
- since every key is a different type, `context.Value` fails quicker - as soon as it realizes that the type is wrong (before it would see that the type is `caddy.CtxKey` and compare strings)
- since the key is an int it's comparison is instant, compared to string keys
- using `int32` basically allows the compiler to use only 4 bytes instead of 8 for `int`, which is better for 32-bit systems
This all matters pretty much due to how `context.WithValue` is implemented - it returns a simple wrapper storing 1 key/value pair, which is why using string keys results in constant meaningless string comparisons on every `context.Value` call.

I've additionally tried creating a custom "fastContext" which acted like a map for multiple values (so something like a `ContextWithValues()` function), however there was no performance gain (tried using both slices and map for storage), which makes sense since in the newer versions of Go the default `context.Value` is iterative instead of recursive.

These 2 optimizations cumulatively sped up PrepareRequest by 2 times on average (though I've seen a boost of up to 4 times). Additionally, they sped up the `context.Value` calls by 2 times on average as well. In terms of requests this gave an average boost of extra 100-150 requests per second in my benchmarks (on a 2cpu 3gb ram VPS).


renbou:
I've tested a default map size of 8 instead of 3 and it does not impact performance in any way (as it should), so I've bumped them up to that just to allow for more room without a resize. Now resizes should definitely be a rare thing for Vars and route Groups.

renbou:
I've also searched through Github to see if some plugins rely on these constants being strings and couldn't find any that do, please correct me if I'm wrong though.

renbou:
I have removed the request copy using `originalRequest()` since it was redundant. Notice, that it was used to copy some parts of the original request so that they can be later used in the error handlers as well as other places, however, we already use `request.WithContext` to make another shallow copy which copies those and even more values (!!) which is why we can simply set `OriginalRequestCtxKey` to the received request and it will not have those values overridden. This removes the call to `originalRequest` as well as a costly object allocation, now `PrepareRequest` is 3x faster on **average**

mholt:
@renbou, I love you. (Professionally...) These are great little improvements.

After just a quick glance so far, I'm on board for these changes. My only concern is, are you sure that we don't need to use `originalRequest()` to copy the request? It's very important that it remain unchanged even if middleware handlers mutate any part of the original request (the one we are copying to preserve it).

Just curious, do you have profiles or benchmark results to post? I believe you of course, but being able to see the performance improvements visually would make for a very compelling tweet. 😃 If you don't, and if it would be a lot of work to gather those again, then don't worry about it.

PS. Can I invite you to our developers Slack? We'd love to have you, as we really appreciate your recent contributions. Just let me know which email address to send the invite to.

renbou:
Thank you! :) `originalRequest()` does this:
```go
func originalRequest(req *http.Request, urlCopy *url.URL) http.Request {
	cloneURL(req.URL, urlCopy)
	return http.Request{
		Method:     req.Method,
		RemoteAddr: req.RemoteAddr,
		RequestURI: req.RequestURI,
		URL:        urlCopy,
	}
}
```
So it simply returns an `http.Request` with some copied over string fields and a shallow copy of the URL.
`request.WithContext` does this:
```go
func (r *Request) WithContext(ctx context.Context) *Request {
	if ctx == nil {
		panic("nil context")
	}
	r2 := new(Request)
	*r2 = *r
	r2.ctx = ctx
	r2.URL = cloneURL(r.URL) // legacy behavior; TODO: try to remove. Issue 23544
	return r2
}
```
The `*r2 = *r` basically means that we copy over all of the basic-type fields (such as those string fields), and we copy over the URL as well. And since `PrepareRequest` does `r = r.WithContext(ctx)`, we simply lose the reference to the original request - it is stored in `OriginalRequestCtxKey` instead, which the handlers should not change. `PrepareRequest` then returns the pointer to the **new** request, and also overwrites the original pointer `r = PrepareRequest(r, repl, w, s)`, so there's no way the original request will change. One thing that may change, though, are the headers, but `originalRequest` didn't prevent that as well. However, I've already tested out another optimization which will prevent us from losing the original headers as well, but it's more related to `Server.ServeHTTP` which is why I want to implement it after this PR.

Here's a profile without the optimization:
[profile-original.zip](https://github.com/caddyserver/caddy/files/8310276/profile-original.zip)
image
image

And with the optimization:
[profile-optimization.zip](https://github.com/caddyserver/caddy/files/8310279/profile-optimization.zip)
image
image

Here the performance of working with the context values + `PrepareRequest` differs over 2 times, however more complex examples where the context is accessed more often it should be even more visible.

I realized that what gave the consistent 3x boost was actually the temporary optimization I did to `addHTTPVarsToReplacer` - I made VarsCtxKey into a struct with 2 value fields (uuid and start_time) and a map for other variables, but I haven't finished working on that, so I reverted it in the last commit (you can check out what I tried doing in the commit before the last one). I think I'll still implement it later, since it's also on the hotpath and is a nice optimization in itself :)

renbou:
Slack would be awesome! You can invite me using artemmikheev1311@gmail.com

modules/caddyhttp/errors.go
```diff
@@ -115,6 +113,8 @@ func trace() string {
 	return ""
 }
 
+type errorCtxKeyT int32
```
hairyhenderson, line 116: Super minor as there's only one of these, but IMO `struct{}` is better as a type for context keys - an `int32` is 4 bytes, whereas a `struct{}` is 0b. Also no need to use an arbitrary `0` with no real meaning... Something like:

```go
type errorCtxKeyT struct{}

...
const ErrorCtxKey = errorCtxKeyT{}
```

It's probably more a matter of taste than anything, though, so take it or leave it!

renbou: Oh, right, I've thought about using structs but disregarded that since {} would have to be written everywhere. Somehow didn't think about doing pretty much the same thing as I did here though, lol. I'll change it to struct{} :)

renbou:
@mholt Personally I'd delay the tweet since I've tested out other optimizations which don't change the API and which can be gradually implemented and then tweeted about ;)

hairyhenderson:
@renbou this is awesome - have you thought of adding some benchmarks too?

As a related aside, it might be useful to hook up https://bencher.orijtech.com to keep track of these @mholt 😉

modules/caddyhttp/errors.go
renbou: Ah, it's going to have to be `var` though, so probably not worth it.

hairyhenderson: > Ah, it's going to have to be `var` though, so probably not worth it.

Ah, right. I suppose it matters in this case because `ErrorCtxKey` is exported. I'm used to avoiding exporting context keys to keep access limited. Too bad 😞

renbou:
I've made the allocation of the `requestID` structure lazy, since it seems like it's supposed to be used through the replacer anyways (using `"http.request.uuid"`). I couldn't find where `"uuid"` var would be used directly so this should be fine, right?

mholt:
(I'm watching this, just to let you know! I'll probably be able to give more attention to this on Monday or Tuesday when I get back into the office. Thanks for your work on this.)

renbou:
The latest commit is experimental - it adds a new `mapContext` which stores a "map" in the form of a slice, which gives us faster lookup and allocation time since it's much smaller than the smallest map. It also allows setting context values lazily if they are initialized using `make()` or `new()` (default value, basically), so the `routeGroupCtxKey` map can be initialized only if groups are actually used, which seems rare to me (?).

Made a simple `PrepareRequest` benchmark which doesn't show results as great as the probably more realistic http2 benchmark, here are the results:
1. **Without** latest experimental commit:
```
benchmark                     old ns/op     new ns/op     delta
BenchmarkPrepareRequest-8     1681          1289          -23.32%
benchmark                     old allocs     new allocs     delta
BenchmarkPrepareRequest-8     23             18             -21.74%
benchmark                     old bytes     new bytes     delta
BenchmarkPrepareRequest-8     2056          1608          -21.79%
```
2. **With** latest experimental commit:
```
benchmark                     old ns/op     new ns/op     delta
BenchmarkPrepareRequest-8     1681          1097          -34.74%
benchmark                     old allocs     new allocs     delta
BenchmarkPrepareRequest-8     23             14             -39.13%
benchmark                     old bytes     new bytes     delta
BenchmarkPrepareRequest-8     2056          1496          -27.24%
```

renbou:
Interesting results from the profile during `h2load` benchmark with the latest commit - nearly 2x faster again
image
image

This is also with an added route group, which proves that everything works.

renbou:
I've polished up the experimental fast context map and added short-circuit possibilities via `caddyhttp.RequestContextValue` and `caddyhttp.RequestContextSetValue`. Basically, they take the context from `http.Request`, lookup our fast map (renamed to `RequestContext`) and then work directly with it. For `caddyhttp.RequestContextValue` if a key is not found in the `RequestContext` map, it will not be looked up further in the parent context, which allows us to save lots of time on lookups of keys which might not be in the context, such as `ErrorCtxKey` and `routeGroupCtxKey` - before most of the time looking them up was spent digging through underlying context.valueCtx, context.cancelCtx, etc. Now pretty much all of the lookups happening in `Server.ServeHTTP` are handled by this fast map:
image

renbou:
This just became even more awesome. Since now the new `RequestContext`structure, is used which is defined by Caddy, I was able to add a `sync.Pool` so now close to none allocations other than during `r.WithContext` happen in `PrepareRequest`!

renbou:
Benchmark of concurrent 20 workers running `PrepareRequest` 5 times.
Optimized `PrepareRequest` without/with pool:
```
benchmark                                 old ns/op     new ns/op     delta
BenchmarkConcurrentPrepareRequest20-8     149091        125600        -15.76%
benchmark                                 old allocs     new allocs     delta
BenchmarkConcurrentPrepareRequest20-8     1437           1030           -28.32%
benchmark                                 old bytes     new bytes     delta
BenchmarkConcurrentPrepareRequest20-8     151519        100089        -33.94%
```

Non-optimized `PrepareRequest` and optimized `PrepareRequest` with pool:
```
benchmark                                 old ns/op     new ns/op     delta
BenchmarkConcurrentPrepareRequest20-8     219029        125600        -42.66%
benchmark                                 old allocs     new allocs     delta
BenchmarkConcurrentPrepareRequest20-8     2291           1030           -55.04%
benchmark                                 old bytes     new bytes     delta
BenchmarkConcurrentPrepareRequest20-8     204148        100089        -50.97%
```

mholt:
Very cool! Will read more in-depth soon. But for now I noticed the tests are failing:

```
 === RUN   TestStaticResponseHandler
--- FAIL: TestStaticResponseHandler (0.00s)
panic: interface conversion: interface {} is nil, not *caddyhttp.RequestContext [recovered]
	panic: interface conversion: interface {} is nil, not *caddyhttp.RequestContext

goroutine 28 [running]:
testing.tRunner.func1.2({0x19d6000, 0xc000424720})
	/opt/hostedtoolcache/go/1.17.7/x64/src/testing/testing.go:1209 +0x36c
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.17.7/x64/src/testing/testing.go:1212 +0x3b6
panic({0x19d6000, 0xc000424720})
	/opt/hostedtoolcache/go/1.17.7/x64/src/runtime/panic.go:1047 +0x266
github.com/caddyserver/caddy/v2/modules/caddyhttp.RequestContextValue(0xc000602700, {0x195c740, 0x1def928})
	/home/runner/work/caddy/caddy/modules/caddyhttp/requestcontext.go:58 +0x10d
github.com/caddyserver/caddy/v2/modules/caddyhttp.StaticResponse.ServeHTTP({{0xc00042a0bd, 0x3}, 0xc0004246f0, {0x1b6dfc0, 0x4}, 0x1, 0x0}, {0x1e1efb0, 0xc0003a8e80}, 0xc000602700, ...)
	/home/runner/work/caddy/caddy/modules/caddyhttp/staticresp.go:142 +0x551
github.com/caddyserver/caddy/v2/modules/caddyhttp.TestStaticResponseHandler(0xc0003f5860)
	/home/runner/work/caddy/caddy/modules/caddyhttp/staticresp_test.go:41 +0x370
testing.tRunner(0xc0003f5860, 0x1bdc940)
	/opt/hostedtoolcache/go/1.17.7/x64/src/testing/testing.go:1259 +0x230
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.17.7/x64/src/testing/testing.go:1306 +0x727
FAIL	github.com/caddyserver/caddy/v2/modules/caddyhttp	0.168s
?   	github.com/caddyserver/caddy/v2/modules/caddyhttp/caddyauth	[no test files]
```

renbou:
@mholt What are the chances of getting this merged this week? Does anything else have to be done on my part? This doesn't change the API other than `OriginalRequestCtxKey` having a `*http.Request` as value instead of `http.Request` (and `PrepareRequest` returns an additional `*RequestContext`, but I didn't find any code in third-party modules using it).

francislavoie:
There's a conflict now on `map_test.go` if you could fix that

renbou:
Is this ok? Doesn't look very nice, I don't rebase PRs often so I'm not sure if this is how it's supposed to look, lol

mholt:
Hmm, there does seem to be some extra cruft in this PR diff now. Not sure what caused that exactly.

I'll review this as soon as I get a chance. It started as a somewhat small change, but now it's got quite a lot of changes -- including a lot of exported surface area -- so it's going to take significantly longer to review.

To speed up the process, I usually recommend/prefer several smaller PRs, one at a time (easier to manage that way, rather than having to submit several PRs all at the same time which rely on the previous one, and updating one after the other). Best to iterate on one small change at a time. So if you want to do that, it might be faster. Or you can leave this PR open, it will just take longer. Especially in this part of the code which is quite central and critical, so I will have to scrutinize it more.

renbou:
For some reason the rebase added all of the commits in this repo over the past few days to this PR. I'll open up a new one with the changes I actually made

mholt:
Opening a new PR would lose all the discussion and context of this one though.

renbou:
Should be good now. @mholt even though a lot of files seem changed, if you look through the changes nearly all of them just add `RequestContextValue` to allow short-circuiting the context value lookup and `PrepareRequest` in the tests so that the request context is prepared properly (preparing the request using multiple `context.Value` in every test definitely isn't the way to do it)

francislavoie:
The key is that when you rebase, you have to _force push_, otherwise git might try to create a merge commit with `master`.

renbou:
After some more testing and research (found this article https://darkcoding.net/software/go-slice-search-vs-map-lookup/) it turned out that using a `map[interface{}]interface{}` here should actually be faster for the lookups even without storing extra values (using a map on the default 4-6 values is faster in most cases than the slice), so I replaced the slice-map with a proper one.
A simple benchmark of `PrepareRequest` says that this is a bit slower, though the speed-up we gain during lookups will definitely be more significant, especially with complex configs.

francislavoie:
I love O(1) lookups

modules/caddyhttp/server.go
```diff
@@ -579,21 +614,25 @@ func (slc ServerLogConfig) getLoggerName(host string) string {
...
-	ctx = context.WithValue(ctx, ServerCtxKey, s)
-	ctx = context.WithValue(ctx, VarsCtxKey, make(map[string]interface{}))
-	ctx = context.WithValue(ctx, routeGroupCtxKey, make(map[string]struct{}))
-	var url2 url.URL // avoid letting this escape to the heap
-	ctx = context.WithValue(ctx, OriginalRequestCtxKey, originalRequest(r, &url2))
-	r = r.WithContext(ctx)
+// Along with the new request it returns the RequestContext stored inside so that
+// the caller may set additional context values if needed, which should be preferred
+// over context.WithValue and r.WithContext which would be slower and result in more allocations.
+func PrepareRequest(r *http.Request, repl *caddy.Replacer, w http.ResponseWriter, s *Server) (
```
francislavoie, line 620: I think `PrepareRequest` is only exported to be used in tests, right? Maybe we should say as much in the godoc comment, to make sure nobody tries to use this at runtime other than `caddyhttp`.

renbou: If I remember correctly, PrepareRequest is indeed currently only used in tests though I can imagine a usecase where somebody wants a custom Caddy app which receives requests in some non-http form and then serves them using Caddy's http handlers. In the way PrepareRequest is currently implemented, it does not enforce the caller to return the RequestContext to the pool or even use it, so we might as well leave it exported without "marking" it as "for internal use only", no?

francislavoie (approved):
This is looking pretty great to me! 🎉

mholt:
I don't get it... if this is so much faster, why wasn't context implemented this way? What are we sacrificing? Immutability (for concurrency)?

Will need more time to finish reviewing this...

renbou:
> I don't get it... if this is so much faster, why wasn't context implemented this way? What are we sacrificing? Immutability (for concurrency)?

Well, **we** aren't sacrificing anything with this optimization, but in general context cannot be implemented this way since it has no way of knowing what ctx will be passed to `context.WithValue` (since it's an interface), which is why all it can do is add one value which will be looked up in linear time in the future... Contexts with cancels and deadlines are implemented "efficiently" since they just add a new goroutine which just blocks, using near zero resources. But a value context can't be implemented efficiently, i.e. using a map, since (like in this case) the map will have to be found first, again by using `ctx.Value`, which can be costly when we don't know what the context will consist of. Here we know that the request context will always be our custom `RequestContext`, which is why the lookup is fast.

JeDaYoshi:
I've been testing this PR as I'm interested on the optimisations, and I've had 2 random crashes so far. Might be worth taking a look into before merging this. I've not been able to reproduce this on `master` branch, just when I add this PR on top of it.

```
caddy[1501090]: panic: runtime error: invalid memory address or nil pointer dereference
caddy[1501090]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x10b0d24]
caddy[1501090]: goroutine 1108060 [running]:
caddy[1501090]: github.com/caddyserver/caddy/v2/modules/caddyhttp.(*RequestContext).Done(0xc00066e490?)
caddy[1501090]: #011<autogenerated>:1 +0x24
caddy[1501090]: context.parentCancelCtx({0x1e9dfa0, 0xc000231260})
caddy[1501090]: #011context/context.go:298 +0x2b
caddy[1501090]: context.removeChild({0x1e9dfa0?, 0xc000231260?}, {0x1e95400, 0xc00066e480})
caddy[1501090]: #011context/context.go:315 +0x31
caddy[1501090]: context.(*timerCtx).cancel(0xc00066e480, 0x1, {0x1e8c760?, 0xc00007e1a0?})
caddy[1501090]: #011context/context.go:486 +0x55
caddy[1501090]: context.WithDeadline.func3()
caddy[1501090]: #011context/context.go:459 +0x30
caddy[1501090]: net.(*Dialer).DialContext(0xc0006945a0, {0x1e9dfa0, 0xc000231260}, {0x1b1850e, 0x3}, {0xc0009bdb80, 0xe})
caddy[1501090]: #011net/dial.go:431 +0x7a2
caddy[1501090]: github.com/caddyserver/caddy/v2/modules/caddyhttp/reverseproxy.(*HTTPTransport).NewTransport.func2({0x1e9dfa0, 0xc000231260}, {0x1b1850e, 0x3}, {0xc0009bdbb0, 0xe})
caddy[1501090]: #011github.com/caddyserver/caddy/v2@v2.4.6/modules/caddyhttp/reverseproxy/httptransport.go:196 +0x153
caddy[1501090]: net/http.(*Transport).dial(0xc000231260?, {0x1e9dfa0?, 0xc000231260?}, {0x1b1850e?, 0x4b?}, {0xc0009bdbb0?, 0x1b1c284?})
caddy[1501090]: #011net/http/transport.go:1169 +0xda
caddy[1501090]: net/http.(*Transport).dialConn(0xc00039da40, {0x1e9dfa0, 0xc000231260}, {{}, 0x0, {0x1b1af84, 0x4}, {0xc0009bdbb0, 0xe}, 0x0})
caddy[1501090]: #011net/http/transport.go:1607 +0x83f
caddy[1501090]: net/http.(*Transport).dialConnFor(0xc0009da180?, 0xc00077ef20)
caddy[1501090]: #011net/http/transport.go:1449 +0xb0
caddy[1501090]: created by net/http.(*Transport).queueForDial
caddy[1501090]: #011net/http/transport.go:1418 +0x3d2
systemd[1]: caddy.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
systemd[1]: caddy.service: Failed with result 'exit-code'.
```

```
caddy[710]: panic: runtime error: invalid memory address or nil pointer dereference
caddy[710]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x10a09c7]
caddy[710]: goroutine 14176 [running]:
caddy[710]: github.com/caddyserver/caddy/v2/modules/caddyhttp.(*RequestContext).Value(0xc0004933b0, {0x176acc0?, 0x290f9c0})
caddy[710]: #011github.com/caddyserver/caddy/v2@v2.4.6/modules/caddyhttp/requestcontext.go:44 +0xa7
caddy[710]: net/http/httptrace.ContextClientTrace(...)
caddy[710]: #011net/http/httptrace/trace.go:25
caddy[710]: net/http.(*persistConn).readLoop(0xc00092b320)
caddy[710]: #011net/http/transport.go:2104 +0x2ed
caddy[710]: created by net/http.(*Transport).dialConn
caddy[710]: #011net/http/transport.go:1750 +0x173e
systemd[1]: caddy.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
systemd[1]: caddy.service: Failed with result 'exit-code'.
```

mholt:
@JeDaYoshi Thank you for testing this.

@renbou Just so you know I greatly appreciate the optimizations, as very little effort has gone into optimizing Caddy. Can you take a look at the crashes and see what might be the problem? As I said before, I think this PR would be easier to swallow if it was a bit smaller, like if we could accept it in smaller chunks. It's just easy for me to get nervous with such a big change to such a central part of Caddy's HTTP server, if you understand. 😅

renbou:
@JeDaYoshi this seems like the RequestContext's underlying Context is nil for some reason (requestcontext.go:44). If the map was also nil, it would've crashed earlier, but it didn't. Are you running Caddy with any extra/non-default modules? What handlers do you have enabled in your configuration?

renbou:
@mholt the problem is in the reverse proxy handler. It passes the request to `net/http` clients, which pool connections and, therefore, extend the context's lifespans after the request is handled, which results in a race condition since we return the context to the pool and `nil`'ify all its values before `net/http` is done working with it.

On a side note, the reverse proxy handler **should not** make a deep clone of the request, since it has every right to alter any value it wants. If a middleware requires the original request, it can do so by getting the `OriginalRequestCtxKey` value, and no middlewares should copy the request (especially deep-copy, like the reverse proxy handler does) since that degrades the performance by a lot (`http.Request` is a very big struct).

However, there can be quite a lot of cases where a request needs to be passed on to a different package in which we cannot make safe assumptions about how it will be used, so making a shallow clone (with a new `RequestContext`) is needed. For such cases it **is** safe to say that we can pass the request with `*RequestContext` as context, since the map will not be used concurrently, but it **isn't** safe to assume the lifetime of the request or the context, so we cannot return it to the pool.

renbou:
Ok, I see the problem and the solution here. Considering the fact that there might be a lot of cases where a request is handled by a different package (e.g. reverse proxy uses `net/http` client), we need to be able to preserve the request's context even *after* we have handled the request. There can be no issues with concurrent access to the `RequestContext` `values` map, because it is only known by Caddy's handlers where we can safely know that no concurrent writes/reads will happen. Thus, the only real problem is the lifespan of the actual `RequestContext`. There are two solutions:
- copying the request along with the context and not returning the new context to the `sync.Pool` to prevent this issue
- marking the `RequestContext` as non-returnable so that it is kept even after the request is handled.

Looking at these two options, the first one turns out to be the second one but with extra steps. Using the second solution will still leave us quite a lot of requests which aren't handled by specific middlewares like reverse proxy, meaning that the `sync.Pool` will still give a noticeable boost.

renbou:
It works (I've reproduced the crash and this fixes it), however, well, this still doesn't look very nice. I found out that Caddy's `cache-handler` also passes the request to an external package, and this just feels very error-prone. What we need here is to be able to keep the `RequestContext` as a valid `context.Context` even after the request is handled, **but** we don't need keep Caddy-specific values stored in the context (since their lifetime is guaranteed to be no longer than the handling of a request). This means that we can still keep a `sync.Pool` for the value map, the variable map, the replacer, etc.

@mholt TLDR:
The previously made assumptions about the single-threaded handling of the request (and thus usage of its context) as well as the lifetime of said request were incorrect, because there are cases where the request is passed to external packages (such as net/http) for which we cannot be sure that the `RequestContext` can be returned to a pool after handling.
For now, I will remove pooling altogether to simplify this PR, and implement more powerful pooling in a later one.

francislavoie:
@renbou the reason we do a clone of the request in `reverse_proxy` is because we make header modifications (i.e. `header_up` config) on the request before sending the request upstream, and we don't want those changes to persist past that point for handlers that work with the request up the chain. We used to not do a clone but just remember the old headers before modifiations, but we ran into other issues with other fields in the request, so we changed it in #4551 to do a shallow-ish clone.

I can't really speak to the pooling stuff though 🤔 that's kinda going over my head. This is just to avoid some allocations? Sounds difficult to do safely, yeah.

renbou:
> We used to not do a clone but just remember the old headers before modifiations, but we ran into other issues with other fields in the request, so we changed it in #4551 to do a shallow-ish clone.

Thank you for referencing the issue, now it makes sense.

> I can't really speak to the pooling stuff though 🤔 that's kinda going over my head. This is just to avoid some allocations? Sounds difficult to do safely, yeah.

Yeah, the problem pretty much occurred due to us pooling the **whole** context even though its lifetime isn't properly defined when we pass it outside of Caddy. We can still pool the replacer, vars map and the values map of `RequestContext` safely since we know that they aren't being leaked outside of Caddy and are of no use once the request is finished. I'll add that in a different PR after this one, though, since it's easier to digest this without pooling.

mholt:
@JeDaYoshi That's great to know, thanks for the feedback!

@renbou I haven't forgotten about this btw :) I am not sure what its future is, but I look forward to reviewing it again, if you don't mind waiting until after Caddy 2.6 is released.

mholt:
You still with us @renbou ?

I think this patch shows an amazing amount of effort, but alas with the lack of activity I think continuing to leave this open would do a disservice to you.

It's quite a large change with a lot of surface area, and the gains seem disproportionate to the complexity. I'm really glad it sounds like it's working well for a while, so there's potential here, but I think we need to take these optimizations in smaller chunks if possible.

After I close this, how do you feel about discussing ways to make more localized optimizations or maybe implement this more piecewise? (We can continue discussion here or in an issue. I definitely don't want the project to lose out on all your work.)

PR #4655, opened 2022-03-23
caddyfile: Prevent bad block opening tokens

### LINKED ISSUE #4501: Reverse proxy sporadically 502 when empty curly braces are written on the same line in Caddyfile
## Environment

Windows 10 AMD64

Latest release from `https://caddyserver.com/download`.

  ```bash
./caddy_windows_amd64.exe version
v2.4.6 h1:HGkGICFGvyrodcqOOclHKfvJC0qTU7vny/7FhYp9hNw=
```

I was reverse-proxying for a Vue project's webpack dev-server. All `*.example.com` are aliased to `localhost`.

## Replication

The following works great

```caddyfile
http://local0.example.com {
	reverse_proxy /* localhost:11451 {
		
	}
}
```

But when the curly braces are written on the same line:

```caddyfile
http://local0.example.com {
	reverse_proxy /* localhost:11451 {}
}
```

Nearly half of the times client receives HTTP 502, while other requests return normally even with content.

For 502 ones, Caddy logs:

```
2022/01/03 09:00:21.737 ←[31mERROR←[0m  http.log.error  making dial info: upstream {}: invalid dial address : missing port in address   {"request": {"remote_addr": "127.0.0.1:57018", "proto": "HTTP/1.1", "method": "GET", "host": "local0.example.com", "uri": "/js/9.js", "headers": {"Accept": ["application/signed-exchange;v=b3;q=0.7,*/*;q=0.8"], "Accept-Encoding": ["gzip, deflate"], "Accept-Language": ["en-US,en;q=0.9,zh-CN;q=0.8,zh;q=0.7,zh-TW;q=0.6"], "Cookie": ["jguid=9244a986-f872-22a9-5f42-a6c13ebcf75e"], "Connection": ["keep-alive"], "User-Agent": ["Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.110 Safari/537.36"], "Purpose": ["prefetch"], "Referer": ["http://local0.example.com/"], "If-None-Match": ["W/\"14323-gbbfEHkeFD7peaJ6qEGdT6QDbls\""]}}, "duration": 0, "status": 502, "err_id": "2famxsnvm", "err_trace": "reverseproxy.statusError (reverseproxy.go:886)"}
```

francislavoie:
Fix #4501

Basically, we want to prevent two cases where bad block openings can happen.

```
localhost
reverse_proxy { to abc:123 }
```

☝️ this one used to result in the upstreams list containing `}` (the last token on the line) because since no newline happens, `to` reads all the remaining tokens on the same line, consuming both `abc:123` and `}`, but never closing the block.

This bug is being fixed by checking if after we see a `{` token, whether the next token is on a new line; if not, err. This could break "legitimate" situations where some hypothetical `expression`-like directive wants to consume all the tokens on the line where `{` is some syntactical token of that "language", but I think this is unlikely, and can be worked around by quoting the tokens like `"{"` and `"}"` (or with backticks).

```
localhost
reverse_proxy abc:123 {}
```

☝️ this one ends up in an upstreams list with `abc:123` and `{}`; I don't think there's ever a case where there'd be a placeholder with no name, and having it at the end of line seems like almost always a mistake (where the user is trying to put an empty block in their config).

Same idea, we're checking if when we see `{}` whether the next token is on a new line, and if so, err. This can also be worked around by quoting with `"{}"` 

Worth noting that https://github.com/caddyserver/caddy/pull/4643 made the quote workaround possible in this implementation, because now we know whether a particular token was quoted. We would only know the token value previously, without context of quoting. 🎉

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -458,8 +458,28 @@ func (d *Dispenser) isNewLine() bool {
 	if d.cursor > len(d.tokens)-1 {
 		return false
 	}
-	return d.tokens[d.cursor-1].File != d.tokens[d.cursor].File ||
-		d.tokens[d.cursor-1].Line+d.numLineBreaks(d.cursor-1) < d.tokens[d.cursor].Line
+
+	prev := d.tokens[d.cursor-1]
+	curr := d.tokens[d.cursor]
+
+	// If the previous token is from a different file,
+	// we can assume it's from a different line
+	if prev.File != curr.File {
+		return true
+	}
+
+	// The previous token may contain line breaks if
+	// it was quoted and spanned multiple lines. e.g:
+	//
+	// dir "foo
+	//   bar
+	//   baz"
+	prevLineBreaks := d.numLineBreaks(d.cursor - 1)
+
+	// If the previous token (incl line breaks) ends
+	// on a line earlier than the current token,
+	// then the current token is on a new line
+	return prev.Line+prevLineBreaks < curr.Line
 }
 
 // isNextOnNewLine determines whether the current token is on a different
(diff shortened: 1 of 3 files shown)


caddyconfig/caddyfile/dispenser.go
```diff
@@ -461,3 +461,17 @@ func (d *Dispenser) isNewLine() bool {
...
+// tokens correctly. If there isn't a next token, it returns true.
+func (d *Dispenser) isNextOnNewLine() bool {
+	if d.cursor < 0 {
+		return false
+	}
+	if d.cursor >= len(d.tokens)-1 {
+		return true
+	}
+	return d.tokens[d.cursor].File != d.tokens[d.cursor+1].File ||
+		d.tokens[d.cursor].Line+d.numLineBreaks(d.cursor) < d.tokens[d.cursor+1].Line
```
mohammed90, line 476: This line broke my brain. Can you add a comment explaining the logic?

francislavoie: Broke mine too. It's stolen from `isNewLine` but just shifted to look at `current < next` instead of `isNewLine` doing `prev < current`.

But yeah I think I'll break the `||` into a short-circuiting `if`, then clarify with comments

francislavoie: Should be crystal clear now

mohammed90: Brilliant!

mohammed90 (approved):

mholt (approved):
Thanks Francis, this is cool. We can give this a try!

PR #4657, opened 2022-03-23
map: Prevent output destinations overlap with Caddyfile shorthands

francislavoie:
Essentially, this tries to prevent configs like this:

```
example.com {
    map {host} {path} {
        example.com /example
    }
}
```

Where the output `{path}` is no good, because it's one of the [placeholder shorthands](https://caddyserver.com/docs/caddyfile/concepts#placeholders). This means that the map destination would actually clobber an existing replacer output, and produce unexpected behaviour.

The implementation is kiiinda tricky though.

Since shorthand replacement happens _before_ directive parsing (this surprised me when I was working on this, but "duh" I should've realized sooner), the `map`'s Caddyfile parser _actually_ sees `{http.request.uri.path}` instead of `{path}`; this means that we need to actually look for the _replacements_ and map them back to the shorthand (for the error message).

With this change, the above Caddyfile will now spit out:

```
adapt: parsing caddyfile tokens for 'map': Caddyfile:2 - Error during parsing: destination {path} conflicts with a Caddyfile placeholder shorthand
```

Also, this adds `{vars.*}` shorthand. Nice to have now that we have a `vars` directive.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -88,34 +88,10 @@ func (st ServerType) Setup(inputServerBlocks []caddyfile.ServerBlock,
 		return nil, warnings, err
 	}
 
-	// replace shorthand placeholders (which are
-	// convenient when writing a Caddyfile) with
-	// their actual placeholder identifiers or
-	// variable names
-	replacer := strings.NewReplacer(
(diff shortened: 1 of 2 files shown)


mholt (approved):
Kinda awkward, but probably good to prevent a foot-gun. Thanks Francis

PR #4658, opened 2022-03-24
core: Allow LoadConfig to return ErrConfigUnchanged

RussellLuo:
Other changes:

- Always enable force-reload for dynamic config load
- Rename errSameConfig to ErrConfigUnchanged

See previous discussions in #4615. 

Some more thoughts behind this PR:

1. `forceReload` is always set to true for dynamic config load, since I think the responsibility is more clear and the logic can be simpler if `ErrConfigUnchanged` will be returned by `LoadConfig()` only.
    - (**EDIT**: Sorry, I realized that this change will break #4603 if used with the built-in HTTP loader, which will never return `ErrConfigUnchanged`. Therefore, I reverted this change.)
3. `changeConfig()` can still return `ErrConfigUnchanged`, which I think is helpful for callers to differentiate "config unchanged" from "config successfully changed" when `forceReload` is false.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -112,7 +112,7 @@ func Load(cfgJSON []byte, forceReload bool) error {
 	}()
 
 	err := changeConfig(http.MethodPost, "/"+rawConfigKey, cfgJSON, forceReload)
-	if errors.Is(err, errSameConfig) {
+	if errors.Is(err, ErrConfigUnchanged) {
 		err = nil // not really an error
 	}
 	return err
@@ -123,7 +123,7 @@ func Load(cfgJSON []byte, forceReload bool) error {
 // the new value (if applicable; i.e. "DELETE" doesn't have an input).
 // If the resulting config is the same as the previous, no reload will
 // occur unless forceReload is true. If the config is unchanged and not
-// forcefully reloaded, then errConfigUnchanged This function is safe for
+// forcefully reloaded, then ErrConfigUnchanged. This function is safe for
 // concurrent use.
 func changeConfig(method, path string, input []byte, forceReload bool) error {
 	switch method {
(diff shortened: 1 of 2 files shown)


caddy.go
```diff
@@ -522,15 +520,18 @@ func finishSettingUp(ctx Context, cfg *Config) error {
...
+								logger.Info("dynamically-loaded config was unchanged; will retry")
+							} else {
+								logger.Error("failed loading dynamic config; will retry", zap.Error(err))
+							}
 							continue
 						}
 						if loadedConfig == nil {
 							logger.Info("dynamically-loaded config was nil; will retry")
 							continue
 						}
```
mholt, line 533: It's been a while and we've been through several iterations now, so help me out... I'm trying to recall why simply returning `nil, nil` is insufficient for just retrying the load again later. Why do we need to return a sentinel error value at all?

RussellLuo: Ah, I remember noticing that there's a judgement of nil config, but somehow turned a blind eye to it later.

Considering that LoadConfig is a public API, if there is no available config, I would prefer it to return a sentinel error. Maybe this is just a matter of taste, or maybe this is more conventional (like [DB.QueryRow](https://pkg.go.dev/database/sql#DB.QueryRow) and https://github.com/99designs/gqlgen/discussions/1891)?

That being said, I think returning `nil, nil` is still a feasible solution to avoid producing misleading error messages I mentioned before https://github.com/caddyserver/caddy/pull/4615#issuecomment-1072004059. I think this is not a big deal, so it's also OK if this is the way Caddy prefers :)

mholt:
Sorry, just circling back to this.

I personally don't mind returning `nil, nil`, since I don't like returning "non-error" error values as much as possible, I think `nil, nil` is pretty clear: no new config, no error. :man_shrugging:  No added API surface like a custom sentinel error value, which feels more like a hack. I guess we can change this if it turns out to be buggy for implementers; we'll just have to see.

If it's alright to do so, I'm going to close this now, but feel free to continue discussion if you feel strongly that we need a sentinel error or if I overlooked some other functionality that we need. Thanks for working on this and examining it with me!

PR #4660, opened 2022-03-24
chore: Clean up adapt test line endings

francislavoie:
Lots of the files were using CRLF instead of LF. Mostly my fault cause sometimes I make the files on Windows and VSCode for some reason kept making them with the wrong line endings. Sigh.

Since .txt files typically default to spaces for indentation, I'm also adding an .editorconfig to ensure they use tabs instead


mholt (approved):
LGTM, thanks for the change for greater consistency!

PR #4661, opened 2022-03-25
httpcaddyfile: Fix Auto-HTTPS edgecase

### LINKED ISSUE #4640: Auto-HTTPS edgecase
Reproduce case:

```
http://foo:8447, http://127.0.0.1:8447 {
    reverse_proxy 127.0.0.1:8080
}
```

This produces JSON with automation policies, it shouldn't though.

As a workaround, splitting the two site addresses into separate site blocks fixes the issue.

francislavoie:
Guh, this is complicated.

Fixes #4640

This also follows up on #4398 (reverting it) which made a change that technically worked, but was incorrect. It changed the condition in `hostsFromKeysNotHTTP` from `&&` to `||`, but then the function no longer did what its name said it would do, and it would return hosts even if they were marked with `http://`, if they used a non-HTTP port. That wasn't the intent of it. The test added in there was kept though, because it is a valid usecase.

The actual fix is to check _earlier_ whether all the addresses explicitly have `http://`, and if so we can short circuit and skip considering the rest.

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -494,7 +494,7 @@ func (sb serverBlock) hostsFromKeysNotHTTP(httpPort string) []string {
 		if addr.Host == "" {
 			continue
 		}
-		if addr.Scheme != "http" || addr.Port != httpPort {
+		if addr.Scheme != "http" && addr.Port != httpPort {
 			hostMap[addr.Host] = struct{}{}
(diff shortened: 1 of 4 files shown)


mholt (approved):
Thanks for the great fix, Francis. I appreciate how you get into the trenches to fight the good fight against the Caddyfile bugz. You rock 🪨 

(Also thanks for fixing my derpy comment typo)

PR #4664, opened 2022-03-26
reverseproxy: Sync up `handleUpgradeResponse` with stdlib

francislavoie:
I had left this as a TODO for when we bump to minimum 1.17, but I should've realized it was under `internal` so it couldn't be used directly.

Copied the functions we needed for parity. Hopefully this is ok!

### DIFF
--- a/modules/caddyhttp/reverseproxy/ascii.go
+++ b/modules/caddyhttp/reverseproxy/ascii.go
@@ -1,0 +1,57 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/reverseproxy/acsii.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
```
mohammed90, line 1: Typo in the filename. It should be `ascii`.

francislavoie: 🤦‍♂️ thanks

mholt (approved):
Just tweaked a few comments and added tests. LGTM, thanks for the sync-up!

Also checked that credit is given and original copyright is included, so that's good. :100:

francislavoie:
Thanks, your fixes LGTM

PR #4665, opened 2022-03-27
chore: Fix for xcaddy builds

francislavoie:
Workaround for https://github.com/smallstep/certificates/issues/876, for now.

After upgrading to Go 1.17, running `go mod tidy` insists on pulling in `cloud.google.com/go/kms` as an indirect dependency, which causes an ambiguous import problem (see https://github.com/smallstep/certificates/issues/876 for the repro steps I wrote up).

So... for now, we need to manually remove `cloud.google.com/go/kms` from the indirect requires block every time we run `go mod tidy`, until the upstream dependency problem is fixed.

Sigh. I hate go modules. So much.

To reproduce the problem, run `xcaddy build b4f1a713` (master commit that "introduced" the problem).

To prove this branch fixes the issue, run `xcaddy build fix-build`.


RussellLuo:
Per https://go.dev/ref/mod#go-get:

> ```bash
> # Remove a dependency on a module and downgrade modules that require it
> # to versions that don't require it.
> $ go get -d golang.org/x/text@none
> ```

I think, instead of manually editing go.mod, there is an alternative workaround — we can run `go mod tidy && go get cloud.google.com/go/kms@none` to keep `cloud.google.com/go/kms` away from go.mod.

francislavoie:
Ah, good point, didn't realize `@none` is a thing.

But yeah, looks like running `go mod tidy` _after_ that still re-introduces the problem.

But it is good to have a programmatic way to remove the broken dep instead of having to do it by hand.

To be clear to anyone reading this, that command does the exact same thing as this diff produces, so there's nothing to commit.

francislavoie:
This is good to go for now @mholt

mholt (approved):

PR #4666, opened 2022-03-29
reverseproxy: Add `_ms` placeholders for proxy durations

francislavoie:
Requested in https://caddy.community/t/http-timing-placeholders-are-strings-float-desired/15515

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -58,8 +58,11 @@ func init() {
 // `{http.reverse_proxy.upstream.max_requests}` | The maximum approximate number of requests allowed to the upstream
 // `{http.reverse_proxy.upstream.fails}` | The number of recent failed requests to the upstream
 // `{http.reverse_proxy.upstream.latency}` | How long it took the proxy upstream to write the response header.
+// `{http.reverse_proxy.upstream.latency_ms}` | Same as 'latency', but in milliseconds.
 // `{http.reverse_proxy.upstream.duration}` | Time spent proxying to the upstream, including writing response body to client.
+// `{http.reverse_proxy.upstream.duration_ms}` | Same as 'upstream.duration', but in milliseconds.
 // `{http.reverse_proxy.duration}` | Total time spent proxying, including selecting an upstream, retries, and writing response.
+// `{http.reverse_proxy.duration_ms}` | Same as 'duration', but in milliseconds.
 type Handler struct {
 	// Configures the method of transport for the proxy. A transport
 	// is what performs the actual "round trip" to the backend.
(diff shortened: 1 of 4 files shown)


francislavoie:
Actually, wait a sec... Go's API returns `Milliseconds()` as `int64`, not `float64`, so that's lossy. Bleh.

I think we have to use `.Seconds()` then divide it by `1e3`. That's pretty dumb. :man_shrugging:

mholt:
@francislavoie I haven't looked yet, just reading my notifications real quick, so this might be a dumb comment, but: Can we use the raw Duration value (nanoseconds) and divide by Milliseconds ourselves to preserve the remainder?

francislavoie:
According to `time.go`, there be dragons?

```
// These methods return float64 because the dominant
// use case is for printing a floating point number like 1.5s, and
// a truncation to integer would make them not useful in those cases.
// Splitting the integer and fraction ourselves guarantees that
// converting the returned float64 to an integer rounds the same
// way that a pure integer conversion would have, even in cases
// where, say, float64(d.Nanoseconds())/1e9 would have rounded
// differently.
```

francislavoie:
Obviously the "ultimate" solution would be to have a way to type convert this when using the placeholder instead, like with CEL or something. But there's a lot of extra engineering effort needed to get there and this is probably better to just _do_ for now for it to be more useful. Pretty cheap.

bompus:
This might be a separate PR, as it's a different module, but I was also asking for `{http.request.duration}` to get the same treatment. 😬

francislavoie:
Ah, right, I'll do that too @bompus 

But uh... sigh. We might also have to contend with float to string conversions giving stuff like `1.0000000000000002e-06` :weary: 

We might need to store it as a string, formatted to a limited amount of decimals, or something. This sucks.

bompus:
Something like `fmt.Sprintf("%.6f", floatVar)` ?

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -909,6 +914,7 @@ func (h Handler) finalizeResponse(
 
 	// total duration spent proxying, including writing response body
 	repl.Set("http.reverse_proxy.upstream.duration", time.Since(start))
+	repl.Set("http.reverse_proxy.upstream.duration_sec", time.Since(start).Seconds()*1e3)
```
bompus, line 917: Did this _sec sneak in here during testing? Either way, multiplying secs by 1000 doesn't seem accurate?

francislavoie: It was too late last night for me to be making these changes, they're pretty wack right now.

mholt:
I thought it looks pretty good, except the duration_sec was a little confusing. Was that intentional?

I just pushed a commit and changed it to duration_ms, along with adding `http.request.duration_ms` as requested, @bompus . How does it look now?

francislavoie:
Alright, cool. Added another godoc comment for another of the new ones. I think this is good to go

mholt:
Oops, thanks. Will merge in a moment!

mholt (approved):

PR #4674, opened 2022-04-02
cmd: Fix defaulting admin address if empty in config, fixes `reload`

francislavoie:
Just a silly oops. Followup to #4443

I broke `caddy reload` when the admin address is not explicitly set in the config 😬 it would take the empty admin address if a `--config` is specified at all, and never fall through to use the default.

https://caddy.community/t/caddy-reload-failing/15556


mohammed90 (approved):
Thanks for fixing that!

PR #4679, opened 2022-04-05
Replace placeholder for file_server index

### LINKED ISSUE #4671: Request: add placeholder substitution for file_server fields
Problem: trying to use place holders in the `file_server`'s fields, e.g.:

```
    file_server {
      index "index.{host}.html"
    }
```

Does not substitute `{host}` with the host domain.

Request:add placeholder substitution for all `file_server` fields.

Thanks!

u5surf:
Fix #4671 

Basically I added replaceAll for indexPage.


mholt (approved):
Great, this should work. Thanks for the contribution and the test!

PR #4687, opened 2022-04-08
fileserver: Implement `show` list

### LINKED ISSUE #4683: Request to add reciprocal option show in addition to hide for file_server browsing
I’d like to have a filter for the _file_server_ browse directive to allow specifying folders using a _show_ command in addition to _hide_. 

Instead of listing what shouldn’t be present with _hide_:

 

	localhost {   
        root * a:/
	    file_server browse {
            hide "folder 1" "folder 2" "folder 4" 
	    }
    }    

I just want to list what should be present (something like this):

	localhost {   
        root * a:/
	    file_server browse {
            show "folder 3" "folder 5" 
	    }
    }    

So the listed folders, and all subdirectories and files in them, are the only things listed by the _file_server_.

Thanks!

francislavoie:
Closes #4683

I found a legitimate use for `goto`! 😅 

Pretty self explanatory, I think. `show` list is checked first, if non-empty. `hide` now further restricts on top of that, if `show` didn't cause it to be hidden.

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -519,6 +519,13 @@ func fileHidden(filename string, show []string, hide []string) bool {
 	// cache the components of the input filename
 	var components []string
 
+	// if the filename is the root of the fileserver, then
+	// it should always be shown, because otherwise the
+	// index would always be hidden
+	if filename == separator {
+		goto shown
+	}
+
 	for _, s := range show {
 		if !strings.Contains(s, separator) {
 			// if there is no separator in s, then we assume the user
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -45,6 +45,29 @@ type FileServer struct {
 	// or current working directory otherwise.
 	Root string `json:"root,omitempty"`
 
+	// A list of files or folders to show; the file server will pretend as if
+	// only these exist, hiding everything else. If not set, everything is shown
+	// by default. Accepts globular patterns like `*.ext` or `/foo/*/bar` as
+	// well as placeholders. Because site roots can be dynamic, this list uses
+	// file system paths, not request paths. To clarify, the base of relative
+	// paths is the current working directory, NOT the site root.
+	//
+	// Entries without a path separator (`/` or `\` depending on OS) will match
+	// any file or directory of that name regardless of its path. To show only a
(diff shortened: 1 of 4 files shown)


francislavoie:
That has problems, doesn't work on Windows.

@mholt I'll need your help with this one. I think you have a different mental model for how this should work than I did, so I'm not sure how to move forwards with it.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -45,6 +45,29 @@ type FileServer struct {
...
+	//
+	// When possible, all paths are resolved to their absolute form before
+	// comparisons are made. For maximum clarity and explictness, use complete,
+	// absolute paths; or, for greater portability, use relative paths instead.
+	//
+	// When both `show` and `hide` are used together, shown files or folders are
+	// checked first (hiding anything not matching), and `hide` will be checked
+	// afterwards, further restrict the results. So, a shown directory can have
+	// files within it hidden by `hide`.
+	Show []string `json:"show,omitempty"`
```
mholt, line 69: Instead of copying so much documentation, can we just explain that this works the same as Hide, and to refer to Hide's docs, except that this lists files to show.

(Or put the docs on Show and have Hide reference Show. Either way.)

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -121,6 +149,16 @@ func (fsrv *FileServer) Provision(ctx caddy.Context) error {
 		fsrv.IndexNames = defaultIndexNames
 	}
 
+	// for show paths that are static (i.e. no placeholders), we can transform them into
+	// absolute paths before the server starts for very slight performance improvement
+	for i, s := range fsrv.Show {
```
mholt, line 154: Similarly, can we combine this section and the next section (remove whitespace) and just have one comment that covers both?

mholt (changes requested):
Thanks for this contribution. It looks complex at first glance, but I realize most of it is duplicating logic from Hide.

Is it possible that we can de-duplicate some of the logic -- maybe separate the repeated stuff into functions, and then just invert the logic based on the boolean result of the function?

Like, maybe `fileIsHidden()` is changed to `fileInList()` and if true, then do something different depending on if it is a Hide or Show list. If that makes sense...

francislavoie:
I think I'm gonna close this as stale. It was more complicated than I could handle. I'm still not entirely sure how to solve this one.

PR #4691, opened 2022-04-09
caddyhttp: Implement `skip_log` handler

### LINKED ISSUE #4689: New handler to skip access logs based on request matcher
There's demand for a way to skip certain access logs from being emitted based on request path: https://caddy.community/t/exclude-certain-requests-from-logs-based-on-request-path/15607

I think we can implement this by introducing a new HTTP handler `skip_log` (name TBD) which embeds a request matcher (not the matcher sibling to the route, but a matcher _inside_ the handler, since we need the handler to get the match result back); when invoked, it would run the matcher and if the matcher returns true, then it would set a flag in the request's context (maybe in Vars, probably least invasive place to do it). Then later in `server.go`, in the `defer func()` that writes the access log, we can check for this skip flag to short circuit and skip emitting the log.

francislavoie:
Closes #4689

Introduces a new `skip_log` HTTP handler, which takes a matcher. If the request matches, it will be marked to skip log emission.

This is implemented by writing to a boolean flag set in the `Vars` map when the handler matches, then checking the value at the end of the request just before attempting to emit the log.

Caddyfile syntax:

```
skip_log <matcher>
```

Note that the matcher is _not_ optional, in this case it is required.

At the code level, this directive is registered as a non-handler directive, because `RegisterHandlerDirective` will eat the matcher tokens to put it on the route.

The `skip_log` directive can be used multiple times, and the first matching one will set the skip log flag. No matcher short circuiting happens in this case when using the Caddyfile, because I didn't implement logic for merging multiple `skip_log` handlers. Technically that could be done, but it's probably not worth the effort, it would only be a super mild performance benefit for probably a lot more code.

I set the handler order for `skip_log` to be right near the top, alongside the other `vars`-ish handlers. Having it early is probably ideal because I think it makes most sense to run the matchers on an unmodified request (before any `rewrite` etc).

When configuring with JSON, multiple matcher sets can be configured on the one handler, which gives a very slight performance advantage if `OR` conditions are needed for matching.


caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -858,3 +859,30 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 	}
 	return configValues, nil
 }
+
+// parseSkipLog parses the skip_log directive. Syntax:
+//
+//     skip_log <matcher>
```
mholt, line 865: Can `skip_log` be used within a `handle` block? (Why is a matcher _always_ required?)

francislavoie: I guess yeah you could have it nested within a handle which does have a matcher.

But the way I was thinking of it is why would you skip logging for all requests if you enabled logging? I guess we can loosen that.

mholt: I think we should loosen the _absolute_ requirement that a matcher is required, because `route` and `handle` blocks that apply to only some of the requests are a very valid way to use `skip_log`.

But yeah I agree, it doesn't make sense to use this without _any_ matcher applied. Just not sure it's worth enforcing it.

francislavoie: If you want to fix it up, feel free 😅

mholt: It just occurred to me while refactoring this (and fixing the merge conflict that I broke, d'oh) that we don't need a directive necessarily:

```
vars http.skip_log true
```

WDYT?

francislavoie: That's true but the amount of demand for this I think means we should have first-class support for this rather than hiding it behind a var. Way more discoverable as a directive. Same reason we have `root` as a directive.

mholt: That's a valid point I guess.

mholt (approved):

mholt:
Thank you for working on this, Francis!

Issue #4692, opened 2022-04-09
Support for multiple upstreams in command-line reverse-proxy


parrotmac:
👋🏻 Hey, thanks for the great project!

This is a proposal to add the ability to specify multiple upstream/`to` targets to the `reverse-proxy` command. Currently, this doesn't seem to be possible.

Proposal (not implemented -- looks like parsing a flag multiple times could be tricky):
```bash
caddy reverse-proxy --from :8080 --to host1.example.com --to host2.example.com
```

Alternatively (currently implemented):
```bash
caddy reverse-proxy --from :8080 --to host1.example.com,host2.example.com
```

Is this something worth considering? I've gone ahead and mocked up an [example for my own purposes](https://github.com/caddyserver/caddy/compare/master...parrotmac:ip/reverse-proxy-cmd-multiple-upstreams?expand=1) and would be happy to work on upstreaming this if it's desired.

francislavoie:
Sure, we could add this. But in general, I'd strongly recommend just using a config file instead. It's much more flexible and reliable.

PR #4693, opened 2022-04-10
feat: Multiple 'to' upstreams in reverse-proxy cmd

parrotmac:
From https://github.com/caddyserver/caddy/issues/4692

This is a proposal to add the ability to specify multiple upstream/`to` targets to the reverse-proxy command.

Example usage:
```bash
caddy reverse-proxy --from :8080 --to host1.example.com,host2.example.com
```

What do we think of the implementation? Any suggestions for a different interface?


mholt:
Thanks for the contribution!

I think it is more typical to repeat a flag than to have custom parsing within a flag's value, i.e. `--to host1 --to host2` instead of `--to host1,host2`. This would probably be my preference, anyway.

That said, I'm not entirely sure how nicely our flags API handles that at the moment. What do you think?

mholt:
@francislavoie Yeah, I had to look this up but I knew I did it once before, years ago. We would need to define our own type: https://stackoverflow.com/questions/45487377/golang-flag-ignore-missing-flag-and-parse-multiple-duplicate-flags

Which, I think maybe that's suitable for this. It's just a few lines and very simple. I do think the CLI looks cleaner this way, personally; but maybe that's just my preference. :man_shrugging:

francislavoie:
Ah ok, that's not so bad. Yeah I suppose we should do it that way then :+1:

parrotmac:
> I think it is more typical to repeat a flag than to have custom parsing within a flag's value, i.e. `--to host1 --to host2` instead of `--to host1,host2`. This would probably be my preference, anyway.
> 
> That said, I'm not entirely sure how nicely our flags API handles that at the moment. What do you think?

Yeah, I actually quite prefer the multiple flags form but wasn't sure if it was worth going that route. Here's [an example](https://stackoverflow.com/a/45494246) of how that could be implemented. If that's preferred I'm happy to switch to something like that.

mholt (approved):
I refactored it to use multiple `--to` flags rather than comma-separating them. I think it is slightly more natural. Thank you for the contribution! Sorry it took me a while.

PR #4694, opened 2022-04-10
Replace godotenv to read envfile

### LINKED ISSUE #4686: --envfile does not handle quotes strings
# How to test

Caddyfile
```
http://{$BASE_URL} {
    log

    handle {
        root * /usr/share/caddy/html
        try_files {path} /index.html
        file_server
    }
}
```

test1.env
```
BASE_URL="test.com"
```

test2.env
```
BASE_URL=test.com
```

Run command
```
caddy run --config /etc/caddy/Caddyfile --adapter caddyfile --envfile /etc/caddy/.env
```

# Possible solutions

1. Implement something similar to this: https://github.com/joho/godotenv/blob/master/godotenv.go#L292
2. Just use godotenv

u5surf:
Fixes #4686 
But I don’t see how and where to implement this behavior’s test. Because loadEnvFromFile is private method in cmd package, but also present cmd package doesn’t have any tests at all, moreover even there are nothing in caddytest/integrations. 
Could you tell me what your guidelines or clue of this test?


francislavoie:
Thanks!

I don't really think we need tests for this, as long as you can confirm it works the same as with the previous implementation, that's good enough for us. We don't need to test that the library works because we can assume that the library itself is tested.

mholt:
Thanks for the PR. My main concern with this is that the new dependency looks unmaintained. Lots of open issues and PRs without feedback/discussion. I like that it's a one-file, not-obnoxiously-huge dependency that has tests, but I would rather only add dependencies that are either completely finished or actively maintained.

Should we just fix the quote parsing problem instead?

u5surf:
@mholt Thanks reviewing.
In another word your suggestion, Is that we prefer to make reinvented it than using the dependency which is not maintained frequently?
 I can't help but agree with your suggestion depending its magnitude.
I'll make the function which is enable to parse the quote myself.

u5surf:
@mholt I made an another PR which doesn't include godotenv in this dependency.

PR #4695, opened 2022-04-10
cmd: Add `--diff` option for `caddy fmt`

francislavoie:
Just thought this might be fun to have.

When using no options or using `--overwrite`, it can be difficult to find out what was _actually_ "wrong" according to the formatter.

Example output where a couple of the lines had spaces instead of tabs:

```
  :80 {
  	root * /srv
  	file_server
  
  	handle_errors {
-       @404 {
-           expression {http.error.status_code} == 404
+ 		@404 {
+ 			expression {http.error.status_code} == 404
  		}
  		rewrite @404 /not-found.txt
  	}
  }
```

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -32,6 +32,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/aryann/difflib"
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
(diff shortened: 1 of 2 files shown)


AliAlhajji:
I'm curious what would be the output if there isn't a new line at the end of the file

francislavoie:
It'll show it:

```
  	}
  }
+
```

mholt:
This is cool. I do wonder if it'd be useful to output an actual patch format so that diffing tools can visualize it for the user.

francislavoie:
There's https://github.com/google/go-patchutils that we could use to do that, but it pulls in additional deps, not sure we need the complexity. I don't think we really gain that much from a full-on patch here. If you want a real patch, make it yourself with your machine's `diff` or w/e, not that difficult. Basically `caddy fmt Caddyfile | diff Caddyfile -` or whatever.

mholt:
I see...

Do we really need this at all then? If the pipe to diff can be done anyway nearly just as easily?

francislavoie:
Piping on Windows is uh, not nice :grin: 

I wrote this because I was on Windows and wanted a way to compare quickly in environments where I can't easily add random tooling. Some Windows users who see the warning in their logs might appreciate having this too for example. Or people in Docker who don't feel like getting a Caddy binary outside Docker just to compare their fmt.

AliAlhajji:
> Piping on Windows is uh, not nice 😁
> 
> I wrote this because I was on Windows and wanted a way to compare quickly in environments where I can't easily add random tooling. Some Windows users who see the warning in their logs might appreciate having this too for example. Or people in Docker who don't feel like getting a Caddy binary outside Docker just to compare their fmt.

As a Windows user I can relate to your Windows experience. Actually it was the new line at the end that kept causing the warning.

mholt (approved):

PR #4699, opened 2022-04-12
Pull the quotes off in envfile (#4686)

### LINKED ISSUE #4686: --envfile does not handle quotes strings

u5surf:
Fixes #4686 

substitute for #4694 

Without including godotenv in this dependency but inspired from godotenv is made the parsing rule about the quotes.


mholt:
Thanks for the PR, @u5surf -- although, I don't think this is _quite_ correct (nor are regular expressions needed). I went ahead and implemented my own solution in 30b6d1f47a35d00fca7cff0daa2ff59a98c5a85e -- it's a little more robust to newlines and such. (Closing, but thank you!)

PR #4703, opened 2022-04-13
ci: Ensure we always check for latest version of Go

francislavoie:
- Go `1.18.1` isn't being pulled by https://github.com/actions/setup-go right now because the versions are pulled from https://github.com/actions/go-versions and it hasn't been updated with it yet

- We were incidentally using `1.17` and `1.18` as our semver strings previously; if we change the matrix to use an actual semver like `~1.18.1` then it actually renames the job itself, which breaks the "required job" thing in github

- Mac 1.18.1 isn't out yet because of "an issue" https://github.com/golang/go/issues/52317 so I need to implement a hack for that too

- So to keep the same labels, but customize the semver string, I'm using an `include` to write to a `GO_SEMVER` variable for it, and using that as the `setup-go` version; the rest uses the short version name from the matrix.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,14 +11,21 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        go: [ '>=1.18.1' ]
+        go: [ '1.18' ]
+
+        include:
+        # Set the minimum Go patch version for the given Go minor
+        # Usable via ${{ matrix.GO_SEMVER }}
+        - go: '1.17'
+          GO_SEMVER: '^1.18.1'
+
     runs-on: ${{ matrix.os }}
 
     steps:
     - name: Install Go
       uses: actions/setup-go@v3
       with:
-        go-version: ${{ matrix.go }}
+        go-version: ${{ matrix.GO_SEMVER }}
         check-latest: true
 
     - name: Checkout code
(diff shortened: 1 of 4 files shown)


hairyhenderson (approved):

mholt (approved):

mholt (changes requested):
Hol'up, doesn't work. (Discussed in Slack) -- doesn't pick up Go 1.17.9 or Go 1.18.1...

mohammed90 (approved):
Great catch and addition. Thanks, Francis!

hairyhenderson (approved):

mholt (approved):
Thanks. I kinda hate it but hopefully it's somewhat temporary. :sweat_smile:

.github/workflows/release.yml
```diff
@@ -12,13 +12,21 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         go: [ '1.18' ]
+
+        include:
+        # Set the minimum Go patch version for the given Go minor
+        # Usable via ${{ matrix.GO_SEMVER }}
+        - go: '1.17'
+          GO_SEMVER: '~1.18.1'
```
francislavoie, line 20: Wait I did a goof :man_facepalming:

francislavoie:
```suggestion
        - go: '1.18'
          GO_SEMVER: '~1.18.1'
```

PR #4705, opened 2022-04-14
core: Refactor listeners; use SO_REUSEPORT on Unix

mholt:
Just an experiment for now.


mholt:
I know this is still undergoing testing in some settings, but I've heard from several users that this branch works great for them (as it does for me), so I'm going to merge this now and we can tune later if need be.

Issue #3833, opened 2020-10-31
consider enabling QUIC by default


marten-seemann:
[Chrome has enabled QUIC](https://blog.chromium.org/2020/10/chrome-is-deploying-http3-and-ietf-quic.html) for a large fraction of Chrome Stable. quic-go has been supporting draft-29 for a while now, and will add support for draft-32 in the next release (in case Chrome adds support for that version, which, as far as I know, they don't plan at this point).
quic-go has been the [default transport in IPFS](https://github.com/ipfs/go-ipfs/releases/tag/v0.6.0) and Filecoin for a few months already.

We might want to find a solution for #3055 first (or disable Retry entirely), so users will actually enjoy the reduced handshake latency (1 round-trip instead of 2 in the case of TLS 1.3 over TCP).

Y-A-K-E:
If quic is enabled by default, AJAX or pjax requests PHP sites to not work

marten-seemann:
@Y-A-K-E Why would that be? If so, that would be a bug. Can you share some more details?

mholt:
Overall I'm open to this. Maybe later, once HTTP/3 is finalized and has broad client support.

As a data point, here's one case study where not binding to UDP by default is preferred: https://twitter.com/itstelesphoreo/status/1333457155324846085

francislavoie:
We'll definitely want to add a `disable_http3` option, to flip it from opt-in (currently) to opt-out.

marten-seemann:
> Overall I'm open to this. Maybe later, once HTTP/3 is finalized and has broad client support.

@mholt HTTP/3 is (basically) finalized. The QUIC transport documents have completed IETF Last Call a while back. The documents are now in the RFC editor queue, so it will still take a few months until the final RFCs are published. However, nobody expects any protocol changes at this point.
Chrome has enabled IETF QUIC by default, see https://blog.chromium.org/2020/10/chrome-is-deploying-http3-and-ietf-quic.html.

> As a data point, here's one case study where not binding to UDP by default is preferred: https://twitter.com/itstelesphoreo/status/1333457155324846085

This boils down to how to configure the `address:port` in the Caddyfile. Currently, saying `domain:443` means listen on TCP port 443, and with HTTP/3 enabled to also listen on port 443 for QUIC. Do we need a way to specify an alternative UDP port?
[...]

Issue #3055, opened 2020-02-18
disable QUIC Retries when not under load


marten-seemann:
QUIC defines an address validation mechanism: https://tools.ietf.org/html/draft-ietf-quic-transport-25#section-8.1.2. A server that receives an Initial packet from a new client can send a Retry packet with an opaque token, and the client then includes this token on subsequent packets.

This costs one network round trip, so we don't really want to use this on every connection. This behavior can be configured in quic-go using the `quic.Config`'s [AcceptToken callback](https://godoc.org/github.com/lucas-clemente/quic-go#Config). Currently, Caddy doesn't set any custom callback, which means that quic-go will send a Retry for every client that hasn't connected to the server before.

Can we find a better logic for this?

mholt:
Sure, letsee, I'm not super familiar with this concept yet so I'll have to do some reading.

It looks like tokens in this context are used for "address validation" -- to ensure the client can receive packets at the address they say they are coming from, to help mitigate amplification attacks. Makes sense. Are there security benefits of just always accepting (requiring) a token?

From what I understand so far, though, ideally we only require a token when there is heavy load on the server, which could be a potential amplification attack.

If I'm right so far, then yes I'd definitely like to implement this callback. What do you think is a ballpark good estimate of "under load"? What kind of metric should I use?

marten-seemann:
> It looks like tokens in this context are used for "address validation" -- to ensure the client can receive packets at the address they say they are coming from, to help mitigate amplification attacks. Makes sense. Are there security benefits of just always accepting (requiring) a token?

Your understanding is correct. There's no security benefit of requiring a token (and a one round-trip penalty). There are two DoS benefits of doing so:
1. By requiring the client to prove that it didn't fake the sender IP address, we can defer the expensive crypto operations needed for the TLS handshake.
2. It will be harder to use the server for a reflection attack (although QUIC comes with a built-in limit for that already: a server won't send 3x more bytes than it has received from an unverified client).

> From what I understand so far, though, ideally we only require a token when there is heavy load on the server, which could be a potential amplification attack.

Amplification attack, or DoS attack against the server itself.

> If I'm right so far, then yes I'd definitely like to implement this callback. What do you think is a ballpark good estimate of "under load"? What kind of metric should I use?

I was hoping that Caddy already has some kind of metric that we could reuse here. One metric could be CPU usage, another one could be number of accepted connection with the last x seconds. Not really sure what makes sense here.

mholt:
[...]

Issue #1614, opened 2017-04-24
QUIC-only Mode


injust:
I'd like to play with Caddy's QUIC feature while maintaining my current nginx setup.
Would it be possible to implement a QUIC-only mode that could be enabled by command-line option?

mholt:
🤔 What exactly do you mean by "QUIC-only"? Is that different from using `-quic`? QUIC will be preferred where the client and protocols support it.

Injust:
By that I mean an option for Caddy to accept connections only over QUIC (only listen on UDP?), leaving the TCP port open for another server to listen on.

mholt:
Oh. No, there's no way to do that with Caddy. It's extremely niche, and severely limits the kinds of clients that can connect. Why do you need this?

Injust:
I'd like to experiment with QUIC while keeping nginx for handling other connections since Caddy's cipher suites don't support some legacy clients that I need.

mholt:
Hmm... so, I am not really inclined to implement this, I would recommend just testing on a separate machine or a different port.

Thanks for your request though!

twdkeule:
You can use --origin-to-force-quic-on=<host>:<port> in Chrome to force QUIC on the client.

Injust:
I'm aware of that flag, but I'm trying to force QUIC on the server, as opposed to the client side.

Injust:
I've found a way to redirect traffic incoming on port 443 with iptables prerouting, but if anybody knows of a better method, please do let me know.

Henrocker:
@Injust How did you manage to split the traffic, so you can do QUIC-only with Caddy?

I know the time passed since this answer but can you still explain how to do it or did you find another, better solution, yet?

anupritaisno1:
Use ufw to block the TCP port and only keep udp open?

maltris:
This would be indeed a good feature for those people who would like to use caddy only for quic on 443/udp but another webserver on 443/tcp.

I personally run caddy in a container which then only 443/udp is forwarded to.

zh99998:
I deploy my services in Docker, and this problem could be solved naturally.

nginx:
   ports:
      - '443:443'
caddy:
   ports:
     - '443:443/udp'

GongT:
+1

docker is good solution but I'm not using docker. :(

firewall require root and may need edit kernel setting.

mholt:
@injust @maltris @GongT (and all others subscribed to this)

[...]

PR #4707, opened 2022-04-15
caddyhttp: Enable HTTP/3 by default

mholt:
This PR enables HTTP/3 by default (#3833), and implements the `RequireAddressValidation` callback as recommended in #3055.

Also removes experimental_http3 options/configuration except it does leave it in the JSON config for now, but marks it as deprecated.

Still need to add a way to disable HTTP/3, probably through the use of the existing protocols configuration parameter.

We will likely merge this sometime after the v2.5.0 release.


mholt:
@marten-seemann I went ahead and implemented a very simple load statistic (number of currently active requests). Maybe it's not great, but it's so dang cheap and simple that I want to start with it and see how it goes. Let me know if I'm on the right track in terms of implementation, and making sure I have the logic right for `AcceptToken`. Thanks!

modules/caddyhttp/app.go
```diff
@@ -346,30 +346,24 @@ func (app *App) Start() error {
 					tlsCfg := srv.TLSConnPolicies.TLSConfig(app.ctx)
 					ln = tls.NewListener(ln, tlsCfg)
 
-					/////////
-					// TODO: HTTP/3 support is experimental for now
-					if srv.ExperimentalHTTP3 {
```
francislavoie, line 351: Shouldn't we just add a new option `DisableHTTP3` and flip the boolean here?

mholt: Don't we already have facilities to configure protocols? It might be ALPN, and HTTP/3 supports the "h3" identifier in ALPN: https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-3.1

listeners.go
```diff
@@ -87,12 +87,24 @@ func ListenPacket(network, addr string) (net.PacketConn, error) {
...
-func ListenQUIC(addr string, tlsConf *tls.Config) (quic.EarlyListener, error) {
+// a context other than context.Background is meaningless. If activeRequests
+// is not nil, it should represent the current load on the server.
+func ListenQUIC(addr string, tlsConf *tls.Config, activeRequests *int64) (quic.EarlyListener, error) {
 	lnKey := "quic/" + addr
 
 	sharedEl, _, err := listenerPool.LoadOrNew(lnKey, func() (Destructor, error) {
-		el, err := quic.ListenAddrEarly(addr, http3.ConfigureTLSConfig(tlsConf), &quic.Config{})
+		el, err := quic.ListenAddrEarly(addr, http3.ConfigureTLSConfig(tlsConf), &quic.Config{
+			AcceptToken: func(clientAddr net.Addr, token *quic.Token) bool {
```
marten-seemann, line 97: General explanation: When sending the first Initial packet to an unknown server, the client doesn't have any token, and thus won't send any. If the server decides to perform a Retry (which costs 1 RTT), it sends a Retry packet containing a token, which the client will then use on all subsequent Initial packets.
For reconnects to a server there's another case one needs to consider: During the connection, the server provides tokens to the client to use on future connections.
When receiving (and validating a token), a server probably wants to accept tokens provided in Retry packets for only a short time frame (10s or so), while tokens provided on a fully established connection would be valid for longer. The [quic.Token](https://pkg.go.dev/github.com/lucas-clemente/quic-go#Token) allows you to distinguish between these two cases.

You probably therefore want to:
1. First check if the server is under load. If it is not, not sending a token is probably ok. Better to save that 1 RTT! Return `true` in that case.
2. If the server is under load, it might make sense to require a token (maybe based on the `net.Addr`, if you want to go fancy). That means:
    * returning `false` if `token == nil`
    * else: validating the token. Have a look at the default token validation functions in quic-go for inspiration: https://github.com/lucas-clemente/quic-go/blob/332473668a992ce372c54c353b2bdc24344cf46e/server.go#L241-L259

marten-seemann (commented):
Super excited to see this happening! Let me know if you need anything else, happy to help!

listeners.go
mholt: @marten-seemann Thanks for the clarifications! That's really helpful.

Any chance we can/should reuse the default token validation function?

mholt: @marten-seemann I made a small update to this function based on your feedback. I think it'd be awesome if that `quic.defaultAcceptToken()` function was exported so that we could just call it from here and reuse that logic correctly.

Thanks for the guidance/help!

mholt:
I've pushed a commit that polishes this feature a bit more:

- Deprecate the `server { protocol` sub-option in global options, including `allow_h2c` and `strict_sni_host`.
- Move `strict_sni_host` to sub-option of `server`
- Add `server { protocols` sub-option, which allows setting which HTTP versions to enable: `h1 h2 h2c h3` are allowed values.

This lets the user toggle precisely which HTTP versions they want to support, including HTTP/3-only, which is currently our oldest open issue: #1614. (Yay!)

Because the Go standard library does not let us disable HTTP/1.1, we would [probably have to implement our own server](https://stackoverflow.com/questions/51521141/how-to-completely-disable-http-1-x-support) to support _only_ HTTP/2 (or h2c) without HTTP/1.1 fallback. I actually tried it:

```go
// http2Server is a server that speaks only HTTP/2.
type http2Server struct {
	h1server *http.Server
	server   *http2.Server
	handler  http.Handler
	logger   *zap.Logger
}

func (h2 http2Server) Serve(ln net.Listener) error {
	h2.server = new(http2.Server)
	opts := &http2.ServeConnOpts{
		BaseConfig: h2.h1server,
		Handler:    h2.handler,
	}
	for {
		conn, err := ln.Accept()
		if err != nil {
			return err
		}
		go h2.serve(conn, opts)
	}
}

func (h2 http2Server) serve(conn net.Conn, opts *http2.ServeConnOpts) {
	defer func() {
		if r := recover(); r != nil {
			buf := make([]byte, 64<<10)
			buf = buf[:runtime.Stack(buf, false)]
			h2.logger.Error("panic",
				zap.String("remote", conn.RemoteAddr().String()),
				zap.Any("error", r),
				zap.String("stack", string(buf)))
		}
	}()
	defer conn.Close()
	h2.server.ServeConn(conn, opts)
}
```

but I could not get it to work. Curl complained, `curl: (1) Received HTTP/0.9 when not allowed` - which I don't understand, but did I spend much time looking into why. Anyway, this seems like a huge maintenance burden & undertaking, so I decided to not support HTTP/2-only servers for now. However, HTTP/1-only and HTTP/3-only configurations work fine!

This means HTTP/3 can be disabled, and we've made protocol configuration easy and consistent.

---

**In summary:** I believe have finished this feature, all except for the AcceptToken tweak mentioned above. (I would still like `quic.defaultAcceptToken()` either exported or always called even if we set our own. /cc @marten-seemann :smiley:) 

Once that is taken care of, I think we can merge this and ship Caddy with HTTP/3 enabled by default.

TNQOYxNU:
> but I could not get it to work. Curl complained, curl: (1) Received HTTP/0.9 when not allowed - which I don't understand, but did I spend much time looking into why.

try request with `curl --http2-prior-knowledge`, curl will try to do [http upgrade](https://www.rfc-editor.org/rfc/rfc7540#section-3.2) from http1 first even with `--http2` option. h2c server's response has no http1 header, will be treated as http0.9.
I tested the code, it works on my machine.

mholt:
@TNQOYxNU Ah, that's good to know! Thank you. I will give that a try next time I revisit this (soon).

caddyconfig/httpcaddyfile/serveroptions.go
```diff
@@ -141,22 +140,51 @@ func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (any, error) {
 				}
 				serverOpts.ShouldLogCredentials = true
 
+			case "protocols":
+				protos := d.RemainingArgs()
+				for _, proto := range protos {
+					if proto != "h1" && proto != "h2" && proto != "h2c" && proto != "h3" {
```
francislavoie, line 146: Small thing, in other languages I tend to use an "array includes" pattern when I'm checking if a particular variable is one of many values. In Go, this hasn't been easy because slices don't have per-type functions to do this in a typesafe way, but with generics, we have a way now:

```go
import "golang.org/x/exp/slices"
```

```suggestion
					if slices.Contains([]string{"h1", "h2", "h2c", "h3"}, proto) {
```

If you don't want to pull this in just for this, that's fine, but I think this reads cleaner.

mholt: Oh, that's cool! I am not sure I'm ready for Caddy to use generics quite yet (I know they're one of Go's most thoroughly vetted features ever, I just want to wait a little longer), especially here where the gain is very minimal (and it adds a dependency for... string comparison) but I do see the value of this in general and I think when we are ready for generics, something like this would be a nice use case!

mholt:
Linking this upstream issue as it is the last item remaining before we merge: https://github.com/lucas-clemente/quic-go/issues/3494 -- really appreciate how thoughtful Marten is being about it. :slightly_smiling_face:

bt90:
https://github.com/lucas-clemente/quic-go/issues/2877 sounds like a possible bottleneck for servers with more traffic.

mholt:
@bt90 Possibly! Although the upstream blockers appear to have been merged since then, so it's probably just a matter of time.

mholt:
This branch is ready for final checks; I think we are ready to merge. Thank you @marten-seemann for your help and congrats on the progress!

mholt:
@TNQOYxNU I will go ahead and merge this as a starting point. If we have requests for HTTP/2-only servers that have a legitimate need, we can revisit that feature. But for now I'll just see how it goes. Thanks again for verifying that!

PR #4715, opened 2022-04-19
Expose several Caddy HTTP Matchers to the CEL Matcher

TristonianJones:
This change exposes several Caddy HTTP `Matcher` types for use within the CEL `Matcher`:

```
expression method('GET') || (method('POST') && protocol('https'))
```

The omission of the `not` matcher is intentional; however, there are a handful of 
matchers which have not been implemented: `files`, `header_regexp`, `remote_ip`,
`vars`, `vars_regexp`. 

The `vars` and `vars_regexp` require more knowledge of Caddy than I have to understand
what the implications are for CEL; however, the other three are mostly a case of picking the
right user experience and validation mechanism. 

There are basically two things that happen when a `Matcher` implements the
`celMatcherProducer` interface:
- Validate that a literal value of the appropriate type was provided to the macro
- Convert the constant literal to a `Matcher` instance during program plan time

https://github.com/caddyserver/caddy/issues/3448


TristonianJones:
@francislavoie If you could suggest some tests for setting up placeholders, that would be helpful. If they're not run ahead of time, then the `Matcher` instances which are currently statically compiled would need to be shared with a dynamic path that does not compile them.

modules/caddyhttp/celmatcher.go
```diff
@@ -83,14 +84,39 @@ func (m *MatchExpression) UnmarshalJSON(data []byte) error {
...
+	// have been configured to support CEL.
+	matcherLibProducers := []celLibraryProducer{
+		MatchHeader{},
+		MatchHost{},
+		MatchMethod{},
+		MatchPath{},
+		MatchPathRE{},
+		MatchProtocol(""),
+		MatchQuery{},
+	}
```
francislavoie, line 105: Right, so instead of hardcoding this here, I think what we should do is call `caddy.GetModules("http.matchers")` which will return a list of all the matcher modules, then they can be type asserted to check if they are a `celLibraryProducer`. That way, it doesn't need to be hard-coded, and any third-party plugin could add support. Obviously this would mean we'd need to export `CelLibraryProducer` so it can be used by third-party modules.

modules/caddyhttp/celmatcher.go
```diff
@@ -137,14 +165,15 @@ func (m *MatchExpression) Provision(_ caddy.Context) error {
...
 // Match returns true if r matches m.
 func (m MatchExpression) Match(r *http.Request) bool {
-	out, _, _ := m.prg.Eval(map[string]interface{}{
-		"request": celHTTPRequest{r},
-	})
+	celReq := celHTTPRequest{r}
+	out, _, err := m.prg.Eval(celReq)
+	if err != nil {
+		panic(err)
+	}
```
francislavoie, line 172: Hmm, I definitely don't think we should `panic`. Caddy _must_ keep running.

We added a mechanism to bubble up an error from within a matcher a while back, so instead do this (and return false):

```go
caddyhttp.SetVar(r.Context(), caddyhttp.MatcherErrorVarKey, err)
```

The godoc:
```go
// MatcherErrorVarKey is the key used for the variable that
// holds an optional error emitted from a request matcher,
// to short-circuit the handler chain, since matchers cannot
// return errors via the RequestMatcher interface.
```

modules/caddyhttp/celmatcher.go
```diff
@@ -184,10 +213,23 @@ func (m MatchExpression) caddyPlaceholderFunc(lhs, rhs ref.Val) ref.Val {
 // httpRequestCELType is the type representation of a native HTTP request.
 var httpRequestCELType = types.NewTypeValue("http.Request", traits.ReceiverType)
 
-// cellHTTPRequest wraps an http.Request with
-// methods to satisfy the ref.Val interface.
+// cellHTTPRequest wraps an http.Request with ref.Val interface methods.
```
francislavoie, line 216: Hah, we had a typo here

```suggestion
// celHTTPRequest wraps an http.Request with ref.Val interface methods.
```

modules/caddyhttp/matchers.go
```diff
@@ -1051,7 +1223,356 @@ func (mre *MatchRegexp) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+// string argument.
+//
+// The following function call is returned: <funcName>(request, arg)
+func celMatcherStringMacroExpander(funcName string) parser.MacroExpander {
+	return func(eh parser.ExprHelper, target *exprpb.Expr, args []*exprpb.Expr) (*exprpb.Expr, *common.Error) {
+		for _, arg := range args {
+			if isCELStringLiteral(arg) {
+				return eh.GlobalCall(funcName, eh.Ident("request"), arg), nil
+			}
+			return nil, &common.Error{
```
francislavoie, line 1414: The linter complains in here:

https://github.com/caddyserver/caddy/pull/4715/files#diff-906c27e0b388ea54848a849573bdf556434597d21b03514c4bea5a07989b7206R1414

> modules/caddyhttp/matchers.go#L1414
> SA4004: the surrounding loop is unconditionally terminated (staticcheck)

I figure `len(args) == 1` should be checked instead of doing a loop, and just use `args[0]` afterwards

modules/caddyhttp/celmatcher.go
mholt: I totally forgot about this 😶

mholt:
Woah, it's happening 😳 

Very very cool, @TristonianJones  -- thank you for the contribution! Can't wait to gobble this up.

modules/caddyhttp/celmatcher.go
TristonianJones: I was thinking of leaving third-party support out until all of the standard matchers had been converted and you'd gotten some feature requests from plugin writers to open up the functionality. Basically, I didn't want to commit Caddy to supporting an expanded API surface for matchers until you're convinced it's ready, safe, and desirable for others to use. Thoughts?

I'll happily change the call to use `caddy.GetModules` though.

francislavoie: I think we can just put in the godoc comment that it's still `EXPERIMENTAL` as a warning to potential users. It's what we've done for other features that aren't yet final.

modules/caddyhttp/celmatcher.go
TristonianJones: Fixed.

modules/caddyhttp/matchers.go
TristonianJones: Fixed.

modules/caddyhttp/celmatcher.go
TristonianJones: Fixed.

modules/caddyhttp/celmatcher.go
TristonianJones: The CEL related features are still marked experimental. Do you want a separate callout for the `CELLibraryProducer`? Otherwise all comments have been addressed and I've gotten the code working with placeholders as well.

TristonianJones:
PTAL, I've updated the code to work with placeholders and incorporated review feedback. The next function I would recommend adding would be the `header_regexp` implementation. @francislavoie if you want to handle it, I can do the review. Then all matchers that are exposed would feel complete. The next two of `remote_ips` and `files` can probably happen a bit later since they'll require a bit more work to validate the literal value provided as the argument to the CEL function.

TristonianJones:
Oh, I'm not sure how related it is, but with recent releases of ANTLR, they started adding semantic versions, so if you need to replace the existing ANTLR reference in the `go.mod` with `https://github.com/antlr/antlr4 v4.10.0+incompatible` that will resolve the ambiguous import dependencies if they haven't already been resolved.

modules/caddyhttp/celmatcher.go
francislavoie: IMO we can probably remove the experimental marker on CEL/`expression` at this point, I don't think it's going away, and it's seemed pretty stable thus far. WDYT @mholt?

But yeah we can mark `CELLibraryProducer` itself experimental if we do. Either way, there aren't many third-party request matchers yet so I doubt it'll matter much.

francislavoie:
I started looking into `header_regexp`; I noticed a problem with the placeholder macro, not sure how to resolve it :thinking:

Regexps can have `{n}` in them, e.g. for number of repetitions of a character, etc. For those, it gets turned into a placeholder function call which doesn't make much sense.

E.g. for this:

```
header_regexp({'Field': {'name': 'foo', 'pattern': 'fo{2}'}})
```

It ends up producing this:

```
header_regexp({'Field': {'name': 'foo', 'pattern': 'focaddyPlaceholder(request, "2")'}})
```

I'm not sure what we should do to fix this. One thought is we could support escaping the leading `{` of a placeholder with a `\` but that complicates the regexp and replacement a lot, and adds more ambiguity, like should we strip the `\` from the output always, etc.

Bleh.

It would be a big breaking change to remove placeholder support from expressions at this point, cause that's currently the main way of injecting inputs. We could add another thing like `p('http.request.path')`, i.e. a function which fetches the placeholder, but this wouldn't work with Caddyfile shortcuts (i.e. where `{path}` is transformed into `{http.request.path}` first before being put in the JSON config)

TristonianJones:
@francislavoie You could try changing the placeholder regex slightly to the following:

```go
placeholderRegexp    = regexp.MustCompile(`{([a-zA-Z][\w.-]*)}`)
```

It would just mean that the placeholders would need to start with a letter character rather than a digit. I'm not sure if that's a huge concern, but it would fix this specific case.

francislavoie:
Yeah, good point. I guess that's probably sufficient for now.

Pushed a commit with `header_regexp` implemented, some minor style changes, and added interface guards for `CELLibraryProducer`

TristonianJones (commented):
LGTM

francislavoie:
FYI @TristonianJones we won't merge this until a bit later, once we've started development for `v2.6.0` (i.e. after `v2.5.0` is released etc).

I'll work on the `file` matcher soon, I think that one is the most important remaining one to get working with this, because it's the one that has the most logic that can't otherwise be emulated by using request placeholders + CEL built-ins.

francislavoie:
Added `remote_ip` and `file`! Those were the tricky ones. I think it should do.

What would be nice is supporting the shortcut syntax for `file` where a list of strings can be given which would fill in `try_files` (leaving the other options empty/default). I tried doing it, but I'm not sure how to define a `exprpb.Type` that is a list of string consts or caddyfile placeholders (because `file({http.request.uri.path}, '/index.php')` should be possible).

Here's my WIP diff, maybe you can finish off that one last bit for me?

```diff
diff --git a/modules/caddyhttp/fileserver/matcher.go b/modules/caddyhttp/fileserver/matcher.go
index 01112584..4a75e644 100644
--- a/modules/caddyhttp/fileserver/matcher.go
+++ b/modules/caddyhttp/fileserver/matcher.go
@@ -19,6 +19,7 @@ import (
        "net/http"
        "os"
        "path"
+       "reflect"
        "strconv"
        "strings"
        "time"
@@ -146,9 +147,33 @@ func (m *MatchFile) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 // expression matchers.
 //
 // Example:
+//    expression file({http.request.uri.path}, '/index.php')
 //    expression file({'root': '/srv', 'try_files': [{http.request.uri.path}, '/index.php'], 'try_policy': 'first_exist', 'split_path': ['.php']})
 func (MatchFile) CELLibrary(ctx caddy.Context) (cel.Library, error) {
-       return caddyhttp.CelMatcherImpl(
+       shortPattern, err := caddyhttp.CelMatcherImpl(
+               "file",
+               "file_match_request_list",
+               []*exprpb.Type{caddyhttp.CelTypeListString},
+               func(data ref.Val) (caddyhttp.RequestMatcher, error) {
+                       refStringList := reflect.TypeOf([]string{})
+                       strList, err := data.ConvertToNative(refStringList)
+                       if err != nil {
+                               return nil, err
+                       }
+
+                       m := MatchFile{
+                               TryFiles: strList.([]string),
+                       }
+
+                       err = m.Provision(ctx)
+                       return m, err
+               },
+       )
+       if err != nil {
+               return nil, err
+       }
+
+       longPattern, err := caddyhttp.CelMatcherImpl(
                "file",
                "file_matcher_request_map",
                []*exprpb.Type{caddyhttp.CelTypeJson},
@@ -179,6 +204,12 @@ func (MatchFile) CELLibrary(ctx caddy.Context) (cel.Library, error) {
                        return m, err
                },
        )
+       if err != nil {
+               return nil, err
+       }
+       envOpts := append(shortPattern.CompileOptions(), longPattern.CompileOptions()...)
+       prgOpts := append(shortPattern.ProgramOptions(), longPattern.ProgramOptions()...)
+       return caddyhttp.NewMatcherCELLibrary(envOpts, prgOpts), nil
 }

 // Provision sets up m's defaults.
diff --git a/modules/caddyhttp/fileserver/matcher_test.go b/modules/caddyhttp/fileserver/matcher_test.go
index 5c60317c..16b2403e 100644
--- a/modules/caddyhttp/fileserver/matcher_test.go
+++ b/modules/caddyhttp/fileserver/matcher_test.go
@@ -283,7 +283,23 @@ var (
                        wantErr:   true,
                },
                {
-                       name: "file match index.php (MatchFile)",
+                       name: "file match short pattern index.php (MatchFile)",
+                       expression: &caddyhttp.MatchExpression{
+                               Expr: `file({http.request.uri.path}, "/index.php")`,
+                       },
+                       urlTarget:  "https://example.com/foo",
+                       wantResult: true,
+               },
+               {
+                       name: "file match short pattern foo.txt (MatchFile)",
+                       expression: &caddyhttp.MatchExpression{
+                               Expr: `file({http.request.uri.path})`,
+                       },
+                       urlTarget:  "https://example.com/foo.txt",
+                       wantResult: true,
+               },
+               {
+                       name: "file match long pattern index.php (MatchFile)",
                        expression: &caddyhttp.MatchExpression{
                                Expr: `file({"root": "./testdata", "try_files": [{http.request.uri.path}, "/index.php"]})`,
                        },
@@ -291,13 +307,21 @@ var (
                        wantResult: true,
                },
                {
-                       name: "file match foo.txt (MatchFile)",
+                       name: "file match long pattern foo.txt (MatchFile)",
                        expression: &caddyhttp.MatchExpression{
                                Expr: `file({"root": "./testdata", "try_files": [{http.request.uri.path}]})`,
                        },
                        urlTarget:  "https://example.com/foo.txt",
                        wantResult: true,
                },
+               {
+                       name: "file not match long pattern (MatchFile)",
+                       expression: &caddyhttp.MatchExpression{
+                               Expr: `file({"root": "./testdata", "try_files": [{http.request.uri.path}]})`,
+                       },
+                       urlTarget:  "https://example.com/nopenope.txt",
+                       wantResult: false,
+               },
        }
 )
```

TristonianJones:
@francislavoie I'll give the issues you described a look. It sounds like there are two challenges though:

1. The `file` matcher with support for a `root` and a list of `try_files` is hard to setup as a function
2. Placeholder support is brittle

Regarding the first issue, I think what you might want are the following signatures:

* `file(string)`
* `file(string, string)`
* `file(string, list(string))`
* `file(map(string, dyn))`

Does that sound right?

For the placeholder support, the issue is conflicting syntax between placeholders, map literals, and string literals. The solution is probably a more deliberate scan over the string that attempts to check the context in which the placeholder occurs rather than a simple `replaceAll`. I don't have a solution off the top of my head, but that's what this problem feels like it needs.

francislavoie:
> Regarding the first issue, I think what you might want are the following signatures:

No, we only need two:

- `file(...(string|placeholder))` which only fills `try_files` (not working)
- `file(map(string, dyn))` which fills everything (already works)

If that makes sense.

If you need to set `root` explicitly, use the long way. But generally you don't, because of the [`root` directive](https://caddyserver.com/docs/caddyfile/directives/root) which it can inherit the value from.

> For the placeholder support, the issue is conflicting syntax between placeholders, map literals, and string literals. The solution is probably a more deliberate scan over the string that attempts to check the context in which the placeholder occurs rather than a simple replaceAll. I don't have a solution off the top of my head, but that's what this problem feels like it needs.

Yeah. I'm not in love with the regexp approach. It was fine for our basic initial support for CEL, but I think there's probably a better way to do it. But I don't know how's best to do it 🤷‍♂️

TristonianJones:
@francislavoie I haven't forgotten about this, just needed to attend to some other things. Hope to address the request by end of week.

francislavoie:
No rush, take your time! 😊

TristonianJones:
@francislavoie I have code for the short form and long-form matcher support in this case, but I'm not sure how to test the short form since the files to match are in `./testdata` Any suggestions?

francislavoie:
The tests I wrote in https://github.com/caddyserver/caddy/pull/4715#issuecomment-1112883925 should probably work I think

TristonianJones:
There was a missing `http.vars.root` definition in the test setup that was resulting in a panic. I think I have it now. Thanks

francislavoie:
Ah right, that makes sense. The `file` matcher defaults to pulling from that placeholder if its own `root` config is nil.

TristonianJones:
@francislavoie I was also able to add support for string concatenation with placeholders. Should all be ready to go now!

francislavoie:
Looks amazing!!

I think I'll try to do the final pass and implement the `vars` ones, then this should be ready to merge!

modules/caddyhttp/celmatcher.go
```diff
@@ -241,12 +287,428 @@ func (celTypeAdapter) NativeToValue(value interface{}) ref.Val {
...
+}
+
+// CelValueToMapStrList converts a CEL value to a map[string][]string
+//
+// Earlier validation stages should guarantee that the value has this type
+// at compile time, and that the runtime value type is map[string]interface{}.
+// The reason for the slight difference in value type is that CEL allows for
+// map literals containing heterogeneous values, in this case string and list
+// of string.
+func CelValueToMapStrList(data ref.Val) (map[string][]string, error) {
```
mholt, line 574:
```suggestion
func CELValueToMapStrList(data ref.Val) (map[string][]string, error) {
```

For consistency :smile:

TristonianJones: Done

mholt:
This is going to be so cool! Nice work so far. I think Francis has one or two more little things he wants to contribute and then we can prepare to merge this in.

francislavoie:
No I don't think so. Sorry I haven't had time to get back to this! I will soon!

francislavoie:
Just rebased it, there was a conflict with 1498132ea3c4f01d3be41812dbd02364cb77d263. I'm not sure if Caddy will be logging errors twice now from this :thinking: 

```go
func (m MatchExpression) Match(r *http.Request) bool {
	celReq := celHTTPRequest{r}
	out, _, err := m.prg.Eval(celReq)
	if err != nil {
		m.log.Error("evaluating expression", zap.Error(err))
		SetVar(r.Context(), MatcherErrorVarKey, err)
		return false
	}
	if outBool, ok := out.Value().(bool); ok {
		return outBool
	}
	return false
}
```

francislavoie (approved):

francislavoie:
In https://github.com/caddyserver/caddy/pull/6594 I finally implemented `vars` and `vars_regexp`, after coming up with a solution for better escapes for placeholders

PR #4720, opened 2022-04-22
caddyauth: Speed up basicauth provision, deprecate `scrypt`

francislavoie:
When provisioning `basicauth`, there's a big performance hit on loading/reloading configs because it calculates a fake password; we don't need to do this on-the-fly, it's totally fine to hash the password once offline and just hardcode it.

I did change the `Hasher` interface unfortunately because it's really the cleanest way to implement this, but I'm not aware of any plugins that implement another hashing algorithm anyways, so the breaking change should probably be fine.

---

FWIW I still think we should drop `scrypt` altogether, and stop using base64-encoded strings.

I don't see any compelling reason to keep `scrypt`, it has way too many knobs, and having the salt not being built-in makes it very error prone to use correctly. See https://passlib.readthedocs.io/en/stable/lib/passlib.hash.scrypt.html for example, which discourages its use.

And no need for base64 because `bcrypt` already uses a safe string format (called Modular Crypt Format), so we're just base64 encoding an already-ASCII string which just obfuscates the configured work factor (which is plainly visible in the bcrypt header). For example, the `bcrypt` I hard-coded in here, base64 decoded, is `$2a$14$X3ulqf/iGxnf1k6oMZ.RZeJUoqI9PX2PM4rS5lkIKJXduLGXGPrt6` which you can clearly see at the start `$14` which means the work factor is `14`. That's its only config knob, and can tell you if it's time to re-hash the passwords if that work factor is now too weak (14 is still ridiculously strong, arguably way too high for a webserver, see https://twitter.com/TychoTithonus/status/1546270126730723329 and above).

### DIFF
--- a/modules/caddyhttp/caddyauth/basicauth.go
+++ b/modules/caddyhttp/caddyauth/basicauth.go
@@ -118,6 +118,7 @@ func (hba *HTTPBasicAuth) Provision(ctx caddy.Context) error {
 			return fmt.Errorf("account %d: username and password are required", i)
 		}
 
+		// TODO: Remove support for redundantly-encoded b64-encoded hashes
 		// Passwords starting with '$' are likely in Modular Crypt Format,
 		// so we don't need to base64 decode them. But historically, we
 		// required redundant base64, so we try to decode it otherwise.
--- a/modules/caddyhttp/caddyauth/basicauth.go
+++ b/modules/caddyhttp/caddyauth/basicauth.go
@@ -94,7 +94,7 @@ func (hba *HTTPBasicAuth) Provision(ctx caddy.Context) error {
 
 	// if supported, generate a fake password we can compare against if needed
 	if hasher, ok := hba.Hash.(Hasher); ok {
-		hba.fakePassword, err = hasher.Hash([]byte("antitiming"), []byte("fakesalt"))
+		hba.fakePassword = hasher.FakeHash()
 		if err != nil {
 			return fmt.Errorf("generating anti-timing password hash: %v", err)
 		}
@@ -271,9 +271,11 @@ type Comparer interface {
 // that require a salt). Hashing modules which implement
 // this interface can be used with the hash-password
 // subcommand as well as benefitting from anti-timing
-// features.
+// features. A hasher also returns a fake hash which
+// can be used for timing side-channel mitigation.
 type Hasher interface {
 	Hash(plaintext, salt []byte) ([]byte, error)
+	FakeHash() []byte
 }
 
 // Account contains a username, password, and salt (if applicable).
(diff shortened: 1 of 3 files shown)


francislavoie:
I decided to make the commit to deprecate `scrypt` and add support for Modular Crypt Format for bcrypt (and `caddy hash-password` now outputs them in that format, without the redundant base64).

This is a big UX simplification and avoids a lot of potential footguns that scrypt carries.

francislavoie:
@mholt can I bump this up your queue? I'd really like to get this in for 2.6.0 betas, since it involves a deprecation and a significant performance fix.

mholt:
Yeah. But events first :)

modules/caddyhttp/caddyauth/basicauth.go
```diff
@@ -117,10 +118,18 @@ func (hba *HTTPBasicAuth) Provision(ctx caddy.Context) error {
...
+		// Passwords starting with '$' are likely in Modular Crypt Format,
+		// so we don't need to base64 decode them. But historically, we
+		// required redundant base64, so we try to decode it otherwise.
+		if strings.HasPrefix(acct.Password, "$") {
+			acct.password = []byte(acct.Password)
+		} else {
+			acct.password, err = base64.StdEncoding.DecodeString(acct.Password)
+			if err != nil {
+				return fmt.Errorf("base64-decoding password: %v", err)
+			}
```
mholt, line 130: I would love to eventually remove this extra bending-over-backwards. Maybe we can have a simple way to base64-decode all the bcrypt hashes so that users can easily convert later.

Can we put this here:

```suggestion
		// TODO: Remove support for redundantly-encoded b64-encoded hashes
		// Passwords starting with '$' are likely in Modular Crypt Format,
		// so we don't need to base64 decode them. But historically, we
		// required redundant base64, so we try to decode it otherwise.
		if strings.HasPrefix(acct.Password, "$") {
			acct.password = []byte(acct.Password)
		} else {
			acct.password, err = base64.StdEncoding.DecodeString(acct.Password)
			if err != nil {
				return fmt.Errorf("base64-decoding password: %v", err)
			}
```

so I don't forget

mholt (approved):
LGTM, just a nit so I don't forget later to remove the cruft.

We'll document that scrypt is being deprecated and may be removed in the future -- not because it's broken, just because the config surface isn't ideal for our use case. Maybe I'll move it into a separate plugin repo later.

modules/caddyhttp/caddyauth/basicauth.go
francislavoie: Sure; I'd say we're in no rush to remove this though. In a later version we could probably add a warning that they should decode their configured passwords etc. Maybe we could even have `caddy fmt` do it for them.

mholt (approved):
Cool. Thank you! Sorry it took me a while to get around to this.

PR #4723, opened 2022-04-22
caddytls: Add `propagation_delay`, support `propagation_timeout -1`

francislavoie:
This has been a thorn for plenty of users in situations where their configured DNS resolvers can't actually see the TXT records Caddy ends up writing, so DNS challenges fail for them.

This makes it possible to turn off propagation, but also adds a new initial delay which can be used as a replacement for for "giving the the DNS provider some time to get their things sorted".

For example, to turn of propagation checks, but still wait 30 seconds before continuing:

```
tls {
	issuer acme {
		propagation_delay 30s
		propagation_timeout -1
	}
}
```

### DIFF
--- a/caddytest/integration/caddyfile_adapt/tls_propagation_timeout.txt
+++ b/caddytest/integration/caddyfile_adapt/tls_propagation_timeout.txt
@@ -3,7 +3,12 @@ localhost
 respond "hello from localhost"
 tls {
 	issuer acme {
-		propagation_timeout "10m0s"
+		propagation_delay 5m10s
+		propagation_timeout 10m20s
+	}
+	issuer zerossl {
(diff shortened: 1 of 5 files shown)


hvisage:
Just a Quick Q: This isn't a global option, only a per domain option, correct?

francislavoie:
Not global. This is per-issuer config.

hvisage:
Thank you, might be nice to have it "global", but hhis still works great  THANK YOU!!
```
*.silo1.pint.ovh silo1.pint.ovh {
tls {
    issuer acme {
        dns acmedns /etc/acmedns/clientstorage.json
        propagation_delay 30s
        propagation_timeout -1
    }
}
```

PR #4724, opened 2022-04-22
reverseproxy: Improve hashing LB policies with HRW

mholt:
Previously, if a list of upstreams changed, hash-based LB policies
would be greatly affected because the hash relied on the position of
upstreams in the pool. Highest Random Weight or "rendezvous" hashing
is apparently robust to pool changes. It runs in O(n) instead of
O(log n), but n is very small usually.

- https://en.wikipedia.org/wiki/Rendezvous_hashing
- https://randorithms.com/2020/12/26/rendezvous-hashing.html
- https://medium.com/i0exception/rendezvous-hashing-8c00e2fb58b0 (:arrow_backward: what I used as a starting point)

This algorithm is very simple and makes sense to me. Hopefully I implemented this correctly! :)

/cc @jjiang-stripe and @cds2-stripe

### DIFF
--- a/modules/caddyhttp/reverseproxy/selectionpolicies.go
+++ b/modules/caddyhttp/reverseproxy/selectionpolicies.go
@@ -524,7 +524,7 @@ func hostByHashing(pool []*Upstream, s string) *Upstream {
 	var highestHash uint32
 	var upstream *Upstream
 	for _, up := range pool {
-		if !upstream.Available() {
+		if !up.Available() {
 			continue
 		}
 		h := hash(s + up.String()) // important to hash key and server together
--- a/modules/caddyhttp/reverseproxy/selectionpolicies.go
+++ b/modules/caddyhttp/reverseproxy/selectionpolicies.go
@@ -514,21 +514,26 @@ func leastRequests(upstreams []*Upstream) *Upstream {
 	return best[weakrand.Intn(len(best))]
 }
 
-// hostByHashing returns an available host
-// from pool based on a hashable string s.
+// hostByHashing returns an available host from pool based on a hashable string s.
 func hostByHashing(pool []*Upstream, s string) *Upstream {
-	poolLen := uint32(len(pool))
-	if poolLen == 0 {
-		return nil
-	}
-	index := hash(s) % poolLen
-	for i := uint32(0); i < poolLen; i++ {
-		upstream := pool[(index+i)%poolLen]
-		if upstream.Available() {
-			return upstream
+	// Highest Random Weight (HRW, or "Rendezvous") hashing,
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -514,21 +514,26 @@ func leastRequests(upstreams []*Upstream) *Upstream {
...
+	// guarantees stability when the list of upstreams changes;
+	// see https://medium.com/i0exception/rendezvous-hashing-8c00e2fb58b0,
+	// https://randorithms.com/2020/12/26/rendezvous-hashing.html,
+	// and https://en.wikipedia.org/wiki/Rendezvous_hashing.
+	var highestHash uint32
+	var upstream *Upstream
+	for _, up := range pool {
+		if !up.Available() {
+			continue
+		}
```
jjiang-stripe, line 529: I think there's a tiny edge case with this!

Let's say we have upstreams A and B, where the ordering for highest hash for these is also A > B. If A was unavailable when you started making requests, and you were consistently getting routed to B, then A became healthy again, your requests would suddenly start hitting A. I'd expect to continue having my requests continue to route to B in that case.

I don't think this ends up being a huge deal, but maybe you have some ideas here? I'm a little too fried-egg-brain-friday to think of any at the moment 😅

jjiang-stripe (commented):
I really like how simple this ended up being! I just have one comment about a possible edge case.

modules/caddyhttp/reverseproxy/selectionpolicies.go
francislavoie: I can't see how it would be possible to keep it routed to B without having somekind of memory of which upstream was used. What you can do is use the `cookie` policy which would write the chosen upstream to a cookie in a response header to the client, so they can send it back to the server on subsequent requests as a hint for upstream selection.

mholt: Ah yeah, good question @jjiang-stripe -- you are right, this does not implement _sticky_ routing, only _consistent_ routing.

In other words, if a request maps to upstream A, it will always go to A unless A is down. In that case it will go to B. Once A comes back up, it will go back to A. This is consistent.

For sticky routing (i.e. always return to the same upstream you started on), you'll need some sort of memory or state like @francislavoie mentioned. I agree the cookie policy might be the best way to do this.

jjiang-stripe: That makes total sense!

cds2-stripe:
This LGTM! I had never heard of this algorithm, but it looks appropriate for this use case. +1 for the simple approach.

PR #4726, opened 2022-04-24
httpcaddyfile: Add `{vars.*}` placeholder shortcut, reverse `vars` sort order

francislavoie:
I'm yoinking the placeholder from my https://github.com/caddyserver/caddy/pull/4657 PR because I think we should get this in ASAP for v2.5.0 along with the new `vars` directive.

Also, during testing while writing the docs, I realized we should reverse the `vars` directive sort order based on matchers; since this is a directive which overwrites existing values set by other previously run directives, we actually want the least-specific matcher to run first, and most-specific last, so that the most-specific is the last one to overwrites the value.

For example, with this config:

```
vars foo bar
vars /path* foo baz

respond {vars.foo}
```

Before this change, all requests would return `bar`, because the least specific matcher would run last, overwriting any more specific matcher. After the change, `/path*` requests will correctly return `baz`.

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -424,14 +424,29 @@ func sortRoutes(routes []ConfigValue) {
 			jPathLen = len(jPM[0])
 		}
 
-		// if both directives have no path matcher, use whichever one
-		// has any kind of matcher defined first.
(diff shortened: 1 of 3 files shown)


mholt (approved):
Thanks for this. Seems to be a good change, has a test, and is minimally invasive. Will merge this for 2.5.

PR #4730, opened 2022-04-26
caddypki: Fix `caddy trust` command to use the correct API endpoint

francislavoie:
Oops.


mholt (approved):
This was my mistake. Thank you Francis for fixing it.

You're so dang fast!

Issue #4736, opened 2022-04-27
Module page http.handlers.templates renders incorrectly


mahgoh:
When visiting [https://caddyserver.com/docs/modules/http.handlers.templates](https://caddyserver.com/docs/modules/http.handlers.templates), the content renders incorrectly from the start of the import command.

The cause are missing backticks on line 103 in `modules/caddyhttp/templates/templates.go`. Because of that, all following commands are rendering incorrectly as the sections that should be snippets are not and vice-versa. See below:

<img width="1190" alt="Bildschirmfoto 2022-04-27 um 19 06 07" src="https://user-images.githubusercontent.com/66012520/165581039-73d8e8c6-d3dd-4e60-ae1b-304e7384ce96.png">

mholt:
Thanks for the report and the fix!

PR #4737, opened 2022-04-27
Add missing backticks

mahgoh:
As stated in #4736, the documentation is missing backticks that causes it to render incorrectly.


mholt (approved):
Oops, thanks for the fix!

Issue #4738, opened 2022-04-27
Automatic HTTP redirect not working if domain block has another value for bind


Xinayder:
After recently updating my Caddy from 2.4.3 to the latest version and restarting the daemon, I kept getting an error in my log:

```
Apr 27 20:58:11 yggdrasil caddy[785913]: {"level":"info","ts":1651093091.29062,"logger":"http","msg":"server is listening only on the HTTPS port but has no TLS connection policies; adding one to enable TLS","server_name":"srv0","https_port":443}

Apr 27 20:58:11 yggdrasil caddy[785913]: {"level":"info","ts":1651093091.2908275,"logger":"http","msg":"enabling automatic HTTP->HTTPS redirects","server_name":"srv0"}

Apr 27 20:58:11 yggdrasil caddy[785913]: run: loading initial config: loading new config: loading http app module: provision http: srv0: invalid listener address: :0
```

I thought I had been missing any directives or something else that was removed/deprecated, however I couldn't find anything related to this everywhere else.

I tried setting global options, explicitly enabling automatic HTTPS redirects, but none of that worked. By running `caddy adapt` on my Caddyfile I found out that it only had HTTPS listeners and no HTTP listeners. Then I found out that the default value for the `bind` is to bind to all available interfaces, and upon removing the directive from my domain block, the error was gone.

Caddyfile for reference:

```
# The Caddyfile is an easy way to configure your Caddy web server.
#
# Unless the file starts with a global options block, the first
# uncommented line is always the address of your site.
#
# To use your own domain name (with automatic HTTPS), first make
# sure your domain's A/AAAA DNS records are properly pointed to
# this machine's public IP, then replace the line below with your
# domain name.

(_unauthorized) {
        error {args.0} "Unauthorized" 403
}

*.mydomain.com, mydomain.com {
        #    bind 127.0.0.1
        bind 127.0.0.1 ::1

        header {
                # disable FLoC tracking
                Permissions-Policy interest-cohort=()

                # enable HSTS
                Strict-Transport-Security max-age=31536000;

                # disable clients from sniffing the media type
                X-Content-Type-Options nosniff

                # clickjacking protection
                X-Frame-Options DENY

                # keep referrer data off of HTTP connections
                Referrer-Policy no-referrer
        }

        tls cert@mydomain.com {
                dns vultr REDACTED
        }

        # Set this path to your site's directory.
        root * /var/www/mydomain.com
        file_server {
                hide blog tags categories
        }

        handle_errors {
[...]

PR #4740, opened 2022-04-28
caddyhttp: Improve listen addr error message for IPv6

francislavoie:
Fixes an unclear message as per https://github.com/caddyserver/caddy/issues/4738

Old message:

```
validate: loading http app module: provision http: srv0: invalid listener address: :0
```

New message:

```
validate: loading http app module: provision http: srv0: invalid listener address: ::1:443, there are too many colons, so the port is ambiguous. Did you mean to wrap the IPv6 address with [] brackets?
```

Part of the issue is we were reusing the same variable before/after parsing, so `addr` would get clobbered with the empty `NetworkAddress` value, rendering it as `:0`. Not useful. I split up the variable name to fix that, and added a condition for when there's too many colons (kinda naive cause it doesn't look at whether `[]` was already there but whatever) to suggest a fix.


mholt (approved):
I'm often down for improving error messages. This seems useful, thanks!

PR #4745, opened 2022-04-28
logging: Implement rename filter, changes field key names

### LINKED ISSUE #4744: Customise the field names for logs.
Hey,

I was looking at configuring the logging in caddy to reflect https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html. For example taking request>remote_ip and using `client.ip` instead of `remote_ip` to store our logs in a  more consistent format.

However although we can customise what appears in the logs, we don't appear to be able to change the field names.

Is this a feature I'm not seeing or something that would need to be implemented?

For example:
```json
"request": {
    "remote_ip": "client.ip",
    "remote_port": "52783",
}
```

Becomes
```json
"request": {
    "client.ip": "client.ip",
    "client.port": "52783",
}
```

Thanks,
Liam

francislavoie:
Fixes #4744 

Quick little feature; lets you rename the key of a field:

```
	log {
		format filter {
			wrap console
			fields {
				request>remote_ip rename client_ip
			}
		}
	}
```

Little implementation note. If you try to rename the same field twice, you must use the original name again; they can't be chained. E.g. this will _not_ work, and the result will be `client_ip`:

```
request>remote_ip rename client_ip
request>client_ip rename something_else
```

But this _will_ result in `something_else`:

```
request>remote_ip rename client_ip
request>remote_ip rename something_else
```

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -31,12 +31,13 @@ import (
 
 func init() {
 	caddy.RegisterModule(DeleteFilter{})
+	caddy.RegisterModule(HashFilter{})
 	caddy.RegisterModule(ReplaceFilter{})
 	caddy.RegisterModule(IPMaskFilter{})
 	caddy.RegisterModule(QueryFilter{})
 	caddy.RegisterModule(CookieFilter{})
 	caddy.RegisterModule(RegexpFilter{})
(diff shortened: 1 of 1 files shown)


mholt (approved):
:shipit: 

:ship: it

Neat contribution! Thank you.

PR #4746, opened 2022-04-28
httpcaddyfile: Fix duplicate access log when debug is on

francislavoie:
When access logging is turned on with the `log` directive, and the `debug` global option is used, the `default` logger would not correctly get the `"exclude": ["http.log.access.log0"]` attached to it; so you end up with the access log written twice, unintentionally.

The added adapt test fails before this change, i.e. `exclude` is missing from `default`.


mholt (approved):
Oh wow, thanks. Not sure how I didn't notice that before.

Issue #4747, opened 2022-04-29
Add `protocol_min` and `protocol_max` for reverseproxy transport http


git001:
Based on this question in the community  
https://caddy.community/t/https-backend-suddenly-not-reachable-getting-502-although-site-is-up/15791  
would it be helpful to add the options `protocol_min` and `protocol_max` for the reverseproxy module to the transport protocol http to the tls struct and the MakeTLSClientConfig function?

https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/httptransport.go#L290
https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/httptransport.go#L322

I think that some Servers out there are not ready to use only TLS 1.2/1.3 and the options `protocol_min` and `protocol_max` could help the users of caddyserver to move to caddyserver in front of such Servers.
What do you think.

mholt:
Hi there. Thanks for submitting a feature request.

One of the goals and values of the project is to lean on secure protocols only, and to be future-facing, not to be going backwards. The lack of advanced TLS tuning between the reverse proxy and the upstreams is intentional to avoid turning knobs that reduce security.

TLS 1.2 is 14 years old. I don't think we need to cater to broken protocol versions at this point, and it would go against the grain of this project.

(Respectfully closing, but feel free to reply if needed.)

PR #4748, opened 2022-04-29
Add protocol_min and protocol_max for reverseproxy transport http #4747

git001:
This PR add the tls protocol configuration option for the reverse proxy transport protocol http.
Related to #4747

### DIFF
--- a/modules/caddytls/values.go
+++ b/modules/caddytls/values.go
@@ -115,6 +115,8 @@ var defaultCurves = []tls.CurveID{
 
 // SupportedProtocols is a map of supported protocols.
 var SupportedProtocols = map[string]uint16{
+	"tls1.0": tls.VersionTLS10,
+	"tls1.1": tls.VersionTLS11,
 	"tls1.2": tls.VersionTLS12,
 	"tls1.3": tls.VersionTLS13,
 }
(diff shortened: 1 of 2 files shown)


modules/caddytls/values.go
```diff
@@ -115,6 +115,8 @@ var defaultCurves = []tls.CurveID{
 
 // SupportedProtocols is a map of supported protocols.
 var SupportedProtocols = map[string]uint16{
+	"tls1.0": tls.VersionTLS10,
+	"tls1.1": tls.VersionTLS11,
```
francislavoie, line 119: We specifically removed these along with Caddy v2 because we don't want to make it possible to serve insecure TLS versions. We want to make it as hard as possible for people to do bad things.

mholt:
Thanks for submitting a PR. However usually we like to discuss changes in an issue first -- especially when the change reverses deliberate, security-related decisions made in the code (in this case, moving two old, broken protocols from `unsupportedProtocols` to `supportedProtocols`). That way we can avoid volunteers spending time on code changes that aren't likely to be accepted. I notice you did open an issue first, technically, although it's good to wait for feedback first if you want to potentially spare yourself a few minutes/hours. :)

I'm going to close this PR (respectfully) as explained in the issue. Thanks though!

PR #4754, opened 2022-04-30
reverseproxy: Fix Caddyfile support for `replace_status`

francislavoie:
As per https://caddy.community/t/small-problem-with-basicauth-and-handle-error/15708, apparently I made the Caddyfile parsing a bit too strict, which prevented legitimate uses of `replace_status`.

Notably, it wasn't allowing no matcher (I thought there was no usecase, but turns out there is 😅) and it was parsing the `int` up-front instead of allowing strings for placeholders.


mholt (approved):
LGTM, thanks Francis!

PR #4757, opened 2022-05-01
templates: Add custom template function registration

### LINKED ISSUE #4568: Make `templates` pluggable
Context: https://caddy.community/t/including-files-scripts/15050/4

Right now, the `templates` module is inflexible in that only functions the functions registered by Caddy itself are available.

A good refactor would be to make it possible to pass an optional array of modules in the `http.handlers.templates.functions.*` namespace which get invoked at `Provision()` time to register additional functions.

kroppt:
Fixes #4568

This allows plugins to add functions to the template execution engine. Such plugin modules must be placed under `http.handlers.templates.functions.*`, e.g. `http.handlers.templates.functions.foo`. The interface `templates.CustomFunctions` must be implemented by the plugin module in order to provide those functions.

### DIFF
--- a/modules/caddyhttp/templates/templates.go
+++ b/modules/caddyhttp/templates/templates.go
@@ -272,8 +272,8 @@ func (Templates) CaddyModule() caddy.ModuleInfo {
 
 // Provision provisions t.
 func (t *Templates) Provision(ctx caddy.Context) error {
-	var customFuncs []template.FuncMap
 	fnModInfos := caddy.GetModules("http.handlers.templates.functions")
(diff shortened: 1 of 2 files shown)


mholt (approved):
This is a neat change, thanks for contributing! Let's try it out.

mholt:
FYI for future searchers, this is probably being superseded by #5939.

kroppt:
@mholt Sounds good. It makes sense that you'd have to explicitly "import" the custom functions at configure time that an extension is defining at compile time.

PR #4760, opened 2022-05-04
reverseproxy: Permit resolver addresses to not specify a port

francislavoie:
Context: https://caddy.community/t/caddy-2-5-dynamic-upstreams-and-consul-srv-dns/15839

I realized it probably makes sense to allow `:53` to be omitted, since it's the default port for DNS.


mholt (approved):
Yay, error string comparison :face_with_spiral_eyes: Oh well, not really sure a better way to do that for now.

Thanks! LGTM overall.

PR #4783, opened 2022-05-12
forwardauth: Support renaming copied headers, block support

francislavoie:
This twitter thread https://twitter.com/coldmacchiato_/status/1524683154237407232 made me realize there were some shortcomings to `copy_headers`; it wasn't possible to rename response headers when copied to the request, and if the header names are long, it can get unwieldy to have it all on one line.

So to fix this, I'm adding a `>` syntax where if a token has a `>`, the left side is the response header name, and the right side is the "destination" request header name.

```
	copy_headers {
		Tailscale-User>X-Webauth-User
		Tailscale-Name>X-Webauth-Name
		Tailscale-Login>X-Webauth-Login
		Tailscale-Tailnet>X-Webauth-Tailnet
		Tailscale-Profile-Picture>X-Webauth-Profile-Picture
	}
```

### DIFF
--- a/caddytest/integration/caddyfile_adapt/forward_auth_rename_headers.txt
+++ b/caddytest/integration/caddyfile_adapt/forward_auth_rename_headers.txt
@@ -1,0 +1,116 @@
+:8881
+
+forward_auth localhost:9000 {
+	uri /auth
+	copy_headers A>1 B C>3 {
+		D
+		E>5
+	}
+}
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":8881"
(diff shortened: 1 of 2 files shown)


mholt:
Probably a good improvement; could this work with a space around the `>` though? For readability sake.

francislavoie:
I think that would be way harder to parse cause then we need to deal with counting tokens and merging them if the next token is a `>` etc. And we need to delete tokens here so it's even harder to do correctly. Making each replacement a single token is way easier to handle, much less code.

mholt (approved):
Alrighty, I guess this is good. Thanks very much for the enhancement, as usual! :)

PR #4784, opened 2022-05-13
Add renegotiation option in reverseproxy tls client

yaslama:
The issue #4785 is fixed with this PR and using the following configuration:
```
route /route/* {
	reverse_proxy https://remotehost {
		transport http {
			tls_client_auth cert.pem cert.key
			tls_insecure_skip_verify
			renegotiation freely
		}
	}
}
```


francislavoie:
Please explain your usecase for this. We don't just merge PRs without an explanation.

I would've preferred if you opened an issue first so we could discuss it.

yaslama:
> Please explain your usecase for this. We don't just merge PRs without an explanation.
 I opened issue #4785 and added some explanation.
> I would've preferred if you opened an issue first so we could discuss it.
Understood.

francislavoie (approved):
Looks good, thanks!

I'll let @mholt decide whether he wants this in, but seems fine to me.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -386,6 +388,16 @@ func (t TLSConfig) MakeTLSClientConfig(ctx caddy.Context) (*tls.Config, error) {
 		cfg.RootCAs = rootPool
 	}
 
+	// Renegotitaion
```
dfunkt, line 391: Typo
```suggestion
	// Renegotiation
```

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -317,6 +317,8 @@ type TLSConfig struct {
 
 	// The server name (SNI) to use in TLS handshakes.
 	ServerName string `json:"server_name,omitempty"`
+
+	Renegotiation string `json:"renegotiation,omitempty"`
```
mholt, line 321: Put a comment above this line to document it. Otherwise this field will have no documentation on the Caddy website. Especially important since it accepts 3 specific values.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -386,6 +388,16 @@ func (t TLSConfig) MakeTLSClientConfig(ctx caddy.Context) (*tls.Config, error) {
 		cfg.RootCAs = rootPool
 	}
 
+	// Renegotiation
+	switch t.Renegotiation {
+	case "never":
+		cfg.Renegotiation = tls.RenegotiateNever
```
mholt, line 394: Since this is the default, is it needed? Maybe we should only accept "once" or "freely". Make sure to reject invalid values too.

mholt (changes requested):
Thanks, I think this is a good change. It will be useful in some cases. Just a few comments!

modules/caddyhttp/reverseproxy/httptransport.go
yaslama: "never" is _currently_ the default, and perhaps forever. But I think that it's better to let users setting explicitly to "never" to protect them from any default value change. But I'll you decide.
Invalid values are rejected now.

yaslama:
@mholt Please let me know if you need any other change

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -317,6 +317,14 @@ type TLSConfig struct {
 
 	// The server name (SNI) to use in TLS handshakes.
 	ServerName string `json:"server_name,omitempty"`
+
+	// TLS renegotiation level. TLS renegotiation is the act of performing
+	// subsequent handshakes on a connection after the first.
+	// The level can be:
+	//  "never": (the default) disables renegotiation.
+	//  "once": allows a remote server to request renegotiation once per connection.
+	//  "freely": allows a remote server to repeatedly request renegotiation.
```
mholt, line 326:
```suggestion
	//  - "never": (the default) disables renegotiation.
	//  - "once": allows a remote server to request renegotiation once per connection.
	//  - "freely": allows a remote server to repeatedly request renegotiation.
```

mholt (approved):
Awesome, looks good now. (I just added markdown bullet points since our site renders godocs as Markdown.)

Thanks for the contribution!

Issue #4788, opened 2022-05-15
SCRIPT_NAME not passed to php-fpm gateway


Elegant996:
When using the `php_fastcgi` directive, the variable `SCRIPT_NAME` is not passed to the `php-fpm` gateway. This causes issues in certain scenarios such as when `/ping` is enabled through `php-fpm`. The result is that the gateway cannot properly construct the URI and `/ping` fails to return.

In order to restore proper functionality the below is required:

```
php_fastcgi /ping unix//run/php-fpm/php-fpm.sock {
    env SCRIPT_NAME {header.SCRIPT_NAME}
}
```

It may be prudent to include this in the `caddyfile.go` as part of the `php_fastcgi` directive so that the gateway is aware of the variable by default instead of requiring it each time `/ping` or similar is enabled on the matcher.

francislavoie:
I'm confused. We _do_ set SCRIPT_NAME:

https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go#L295

Elegant996:
Yes, we do but give it a go in a demo. If you enable `ping` for `php-fpm` server and use a drop-in config such as:

```
[global]
daemonize = no

[www]
pm = dynamic
pm.max_children = 120
pm.start_servers = 12
pm.min_spare_servers = 6
pm.max_spare_servers = 18

listen = /run/php-fpm/php-fpm.sock

ping.path = /ping
```

You will find that that there is no `pong` return unless the `SCRIPT_NAME` variable is specified. Although this is `caddy`, [Fossil](https://www.fossil-scm.org/home/doc/trunk/www/scgi.wiki) had an interesting read for SCGI (FastCGI allows applies in this scenario) where it had to specified.

francislavoie:
Likely because Caddy ends up rewriting to `index.php` because there's no `/ping` file that exists in your webroot. I bet if you create an empty file called `ping`, it will work.

francislavoie:
I'm pretty sure this is just a misunderstanding of how `php_fastcgi` works. See https://caddyserver.com/docs/caddyfile/directives/php_fastcgi#expanded-form, which explains the `try_files` logic.

PR #4789, opened 2022-05-15
fastcgi: pass SCRIPT_NAME variable through directive

Elegant996:
Adds `SCRIPT_NAME` variable by default as part of the `php_fastcgi` directive. Addresses #4788.


francislavoie:
As per https://github.com/caddyserver/caddy/issues/4788#issuecomment-1126977646, I don't think this is correct.

Also, you used the Caddyfile shortcut placeholder; this won't work here because shortcut replacement runs before this point. The full placeholder needs to be used when generating the JSON config.

But I don't think it makes sense to use a request header here... SCRIPT_NAME isn't supposed to be taken from HTTP header values.

I'll close this for now, this needs further discussion first.

PR #4790, opened 2022-05-17
core: Windows service integration

### LINKED ISSUE #4670: Minimal windows service stub
It would be a nice addition to integrate [Minimal windows service stub](https://pkg.go.dev/github.com/kardianos/minwinsvc) in Caddy (built for Windows).

At the moment, it is necessary to use for example NSSM to install Caddy as a service on Windows.

See how it becomes much more natural to [install the service and remove it in the case of another project (Gitea) with the Windows tools](https://docs.gitea.io/en-us/windows-service/).

WingLim:
Implement a `svc.Handler` and run it in `init` function, so that caddy can use `sc.exe` to manage a windows service.

With this PR, we can add service like this:

1. [create](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/sc-create)
```
sc.exe create caddy start= auto binPath= "YOURPATH\caddy.exe run"
```
2. [start](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc742126(v=ws.11))
```
sc.exe start caddy
```
3. [stop](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc742107(v=ws.11))
```
sc.exe stop caddy
```

TODO:
- [ ] Add document in [Windows service](https://caddyserver.com/docs/running#windows-service)

Resolve #4670


mholt:
Interesting, will take a look at this. Thanks!

mholt (changes requested):
Sorry this took me so long to get to. This actually looks really good. But do you think we could move service_windows.go into notify_windows.go? Maybe I'm missing something but since it runs from `init()`, it doesn't seem super relevant to have it in a separate package and file.

mholt:
@WingLim Are you still interested in finishing this PR?

mholt:
@WingLim Oh, I see the difficulty of moving the files like I asked. I just went ahead and cleaned up the build tags/pragmas instead and I think we'll merge this. Thanks!

mholt (approved):

Issue #4668, opened 2022-03-29
SSLKEYLOGFILE / NSS Key Log Format support?


Manouchehri:
For diagnostics and troubleshooting (_not_ production), it's quite useful to be able to decrypt your own TLS traffic.

https://firefox-source-docs.mozilla.org/security/nss/legacy/key_log_format/index.html

It looks like Go's TLS package already supports this.

https://pkg.go.dev/crypto/tls#example-Config-KeyLogWriter

The usage would look like:

```sh
SSLKEYLOGFILE=keylog1.txt caddy
```

francislavoie:
PRs welcome, if someone would like to take a crack at it.

We definitely should log a warning when this is turned on, to make it clear that security is actively being compromised.

Generally we're not fans of having behaviour controlled via environment variables.

Config for this would probably go in connection policies: https://caddyserver.com/docs/json/apps/http/servers/tls_connection_policies/ and as an option to the `tls` directive in the Caddyfile.

mholt:
Implemented in #4808. Thanks @Manouchehri !

mildsunrise:
To save some clicks to people who saw the title of the issue and missed/misunderstood the comment above:
Caddy does NOT support `SSLKEYLOGFILE`. Key logging is enabled exclusively through directives.

mholt:
Thanks, yes -- I feel like an env var is too subtle/implicit for such a risky option.

PR #4808, opened 2022-05-23
Add SSL key logging.

Manouchehri:
Solves #4668.


Manouchehri:
@francislavoie Would you mind helping me with adding how to properly parse the config?

caddyconfig/httpcaddyfile/options.go
```diff
@@ -52,6 +52,7 @@ func init() {
 	RegisterGlobalOption("ocsp_stapling", parseOCSPStaplingOptions)
 	RegisterGlobalOption("log", parseLogOptions)
 	RegisterGlobalOption("preferred_chains", parseOptPreferredChains)
+	RegisterGlobalOption("sslkeylogfile", parseOptSingleString)
```
mholt, line 55: We try to avoid the term "SSL" now since it's obsolete. Prefer "TLS" instead.

I'm not super sure a global option is the best way to do this, but maybe it makes sense in practice. I dunno yet. Will have to give it some thought.

modules/caddytls/connpolicy.go
```diff
@@ -260,6 +260,7 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 	}
 
 	setDefaultTLSParams(cfg)
+	setSSLKeylog(cfg)
```
mholt, line 263: This appears to turn on key logging regardless of whether it has been enabled or not.

modules/caddytls/connpolicy.go
```diff
@@ -461,6 +462,16 @@ func setDefaultTLSParams(cfg *tls.Config) {
 	cfg.PreferServerCipherSuites = true
 }
 
+// SSLKEYLOGFILE
+func setSSLKeylog(cfg *tls.Config) {
+	// TODO: Take sslkeylogfile as the file to write.
+	w, err := os.OpenFile("tls-secrets.txt", os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
mholt, line 468: This whole function / approach will probably need to be revised. It's a good start, but it needs to be more flexible and configurable I think.

mholt (changes requested):
Thanks for the contribution. As this is a draft, it will need some work before we can merge it, but I think it's probably not a bad thing to have this feature. I'll give some high-level thoughts for starters.

caddyconfig/httpcaddyfile/options.go
Manouchehri: So I sorta agree, but everyone calls it SSLKEYLOGFILE. e.g.

https://firefox-source-docs.mozilla.org/security/nss/legacy/key_log_format/index.html

https://sslkeylog.readthedocs.io/en/latest/index.html

https://everything.curl.dev/usingcurl/tls/sslkeylogfile

I think it would be confusing to call it tlskeylogfile, as everyone else refers to it as SSL key logging (even when logging non-SSL keys). 

We could call it NSSKeyLogFile, but that still is non-standard. And we aren't using the NSS library, so still slightly confusing to users.

modules/caddytls/connpolicy.go
Manouchehri: Correct. I wasn't sure how to reference the config file here.

modules/caddytls/connpolicy.go
Manouchehri: Something that would be nice (that I don't know how to do), is to also pipe the key logs to the regular Caddy logs.

Maybe something like "include_sslkeylog_logs"?

This would mean the end-user has to extract out the keys from the caddyserver log, but it would be much more flexible than only writing to a single file. e.g. writing to a network socket.

caddyconfig/httpcaddyfile/options.go
mholt: I think we will need to support more than just writing to a file. There's no reason we need to use that specific name. I'd probably call the option `insecure_tls_key_export` or something.

Manouchehri: We should still allow logging to a separate file. It's a hassle to have to manually parse out keys to the "proper" format for usage in tools like Wireshark. (Sometimes it's easier, sometimes it's harder... it's a debugging/troubleshooting feature for many users, so requiring them to write a script just to check something quick isn't ideal.)

modules/caddytls/connpolicy.go
mholt: Sorry, I got busy with other Caddy projects and am now traveling to help a family move. I'll try to get back to this ASAP.

Here's a quick example of how to emit a log (you can get an instance of a logger, if you don't already have one, from the `caddy.Context` during `Provision`): https://github.com/caddyserver/caddy/blob/03b5debd958a324d5c00cc37f887130b0198e747/modules/caddypki/maintain.go#L91-L92

modules/caddytls/connpolicy.go
Manouchehri: Could I have a bit of help on this part?

mholt: Sure. I'll need a little bit to catch up on things first though!

mholt:
@Manouchehri I've pushed my version of this patch, based on your initial PR. After thinking about it some more I decided to leave things pretty simple: you specify a file name in the config and it writes it to a file. We can't control the format of the writing and other tools like Wireshark expect to read from a file, so we'll just do that. I also added a loud warning to the logs to indicate that TLS security is compromised when this feature is used.

Also rebased with master and marked this as experimental so we can change it later, after some field use. (Hopefully not in production.)

Caddyfile usage:

```
example.com

tls {
    insecure_secrets_log keys.log
}
```

That will log the TLS secrets for connections to `example.com`.

Does this change work for you? (Please try it out!)

modules/caddytls/connpolicy.go
```diff
@@ -280,6 +293,30 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
...
+			return err
+		}
+		logFile, _, err := secretsLogPool.LoadOrNew(filename, func() (caddy.Destructor, error) {
+			w, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+			return destructableWriter{w}, err
+		})
+		if err != nil {
+			return err
+		}
+		ctx.OnCancel(func() { _, _ = secretsLogPool.Delete(filename) })
```
mholt, line 312: All this just makes sure that through config reloads, only one instance of each log file is opened at a time. This isn't safe for multiple TLS configs that write to the same file, but you know what, this feature better not be used by busy servers in production -- only for testing and debugging! -- so I'm not too worried about it.

It also makes sure the log file is closed whenever the next config (if any) doesn't use it anymore.

modules/caddytls/connpolicy.go
```diff
@@ -280,6 +293,30 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 		}
 	}
 
+	if p.InsecureSecretsLog != "" {
+		filename, err := caddy.NewReplacer().ReplaceOrErr(p.InsecureSecretsLog, true, true)
```
mholt, line 297: This enables setting the filename from an env var. But it has to be done explicitly, as this definitely isn't a "feature" we want to enable implicitly and obscurely from the environment!

mholt (approved):

Manouchehri:
This looks good to me, thanks!

PR #4810, opened 2022-05-24
minor: micro-optim in run()

dunglas:


mholt (approved):
Alrighty, good call. Thanks!

PR #4811, opened 2022-05-24
ci: Fix build caching on Windows

francislavoie:
I was getting tired of Windows being slow as molasses in our CI jobs, so I went to look at our trusty source of github actions + golang information, and found a somewhat recent commit that actually fixed it. See https://github.com/mvdan/github-actions-golang/commit/4b754729baa709da219a5889c459010d4eda1888

I'll do a 2nd empty commit to re-trigger CI shortly to confirm that it actually fixes it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,12 +78,20 @@ jobs:
         printf "Git version: $(git version)\n\n"
         # Calculate the short SHA1 hash of the git commit
         echo "::set-output name=short_sha::$(git rev-parse --short HEAD)"
-        echo "::set-output name=go_cache::$(go env GOCACHE)"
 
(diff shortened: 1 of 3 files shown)


francislavoie:
Still seems really slow. Bah. Probably worth having anyways.

mholt:
You are asking Windows to be fast, Francis 😄 

Is the extra complexity of this CI job worth the lack of improvement?

francislavoie:
It's actually _less_ complex in this case, cause we're not relying on pulling the path from env anymore, and just hardcoding them. I think we should merge anyways. It's still a correctness fix, I'm just disappointed it didn't help much is all.

mholt (approved):
Ah, okay. Sounds good then. Thank you!

PR #4824, opened 2022-06-02
reverseproxy: HTTP 504 for upstream timeouts

### LINKED ISSUE #4823: reverse proxy: when timeout occurs Caddy responses as 502 Bad Gateway
When I use timeout limit in my Caddyfile and upstream server is slow, Caddy responses in such way:

`HTTP 502: net/http: timeout awaiting response headers`
`HTTP 502: dial tcp <some IP address>:443: i/o timeout`

I think, it should be 504 Gateway Timeout. Because HTTP code 504 means, that I should increase timeout limit.

mholt:
Closes #4823

/cc @dtelyukh


dtelyukh:
It works good for me

Issue #4834, opened 2022-06-09
Make reverse proxy TLS server name replaceable for SNI upstreams.


kresike:
The basic problem we're trying to solve here is the case when we're using dynamic upstreams to proxy many sites from multiple backends using HTTPS and SNI.

My relevant caddy config snippet:

```
:443 {
	reverse_proxy {
		dynamic srv {
			proto tcp
			service {host}
			name proxy.example.com
		}		
	}
}

```
The original problem is, that while caddy does allow (and does so out of the box) replacement of the Host header in the HTTP request sent upstream, it does not set the servername in the TLS client. This results in an error (421 in my case, reported by apache), looking something like below:
```
AH02032: Hostname backend01.example.com provided via SNI and hostname webpage01.example.com provided via HTTP have no compatible SSL setup
```
After implementing the fix, one can use the configuration snippet below to solve the issue:

```
:443 {
	reverse_proxy {
		dynamic srv {
			proto tcp
			service {host}
			name proxy.example.com
		}		
		transport http {
			tls_server_name {host}
		}
	}

}

```
This is a fairly simple fix, that I've already done. Will open a PR for it.

mholt:
We pre-allocate the TLSConfig at Provision time for efficiency. The only known, correct solution is to generate a TLSConfig for every connection. We can do that, but I've hesitated due to the potential performance implications. It wouldn't be much, but I'd like to do it carefully.

kresike:
Thanks for the review @mholt 
I need this done, so let me look into it, maybe I can come up with a better solution.

mholt:
Ok, I'll be happy to review a PR. I might be interested in some profiles but we'll have to see!

kresike:
@mholt please see my second solution. Not the prettiest code I've ever written but it should be loads better than the previous attempt.

francislavoie:
I think it would be a mistake to allow both HTTP and HTTPS at the same time. The srv should only do one or the other, and configure Caddy with the appropriate transport accordingly.

kresike:
Ok, I've thought about this a bit more, and I don't think there is any redesign needed. We can simply fix it using an if statement.

kresike:
Added a small change that fixes this behaviour. A small price to pay is that HTTPS cannot be used on port 80 on the upstreams, which I think is better than messing up protocols.

mholt:
@kresike The revised PR is looking much better! Thank you. Pretty close to merging it in.

kresike:
[...]

PR #4836, opened 2022-06-10
Another run at making reverse proxy TLS server name replaceable for SNI upstreams

kresike:
Updated PR for solving issue #4834

Should be thread safe now, and it only starts doing magic when it detects that magic is indeed needed.

The copying of the transport struct is a bit of a brute force approach, any tips on improving that are welcome.

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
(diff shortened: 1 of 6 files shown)


francislavoie:
I think that new code should be moved into its own function, I don't like how much bloat it adds to the `reverseProxy` function, which otherwise is decently lean.

kresike:
@francislavoie thanks for your input. I was focusing on making this efficient and working without breaking anything. I'll move this into it's own function ASAP.

Joseymartinez525 (approved):

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -730,10 +762,18 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
 	server := req.Context().Value(caddyhttp.ServerCtxKey).(*caddyhttp.Server)
 	shouldLogCredentials := server.Logs != nil && server.Logs.ShouldLogCredentials
 
+	// Default to using the transport configured during provisioning stage
+	transport := h.Transport
+
+	// If we have a HTTP transport, try to replace the TLS servername
+	if tmpTransport, ok := transport.(*HTTPTransport); ok {
+		transport = h.replaceTLSServername(tmpTransport, repl)
+	}
```
mholt, line 771: Instead of using a type assertion here, can this be done in `RoundTrip()` directly? Seeing as the transport is short-lived and all.

mholt (changes requested):
Thanks, I like this approach much better! It's nice that it's only done when needed. I wonder if we could make this a little less coupled with the proxy itself though.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -244,6 +244,7 @@ func (h *HTTPTransport) NewTransport(ctx caddy.Context) (*http.Transport, error)
 
 // RoundTrip implements http.RoundTripper.
 func (h *HTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+
```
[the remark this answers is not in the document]
kresike: Leftover from previous attempt. Sorry!

mholt:
@kresike Oops, not sure how I missed that in the review. I pushed a fix for the renegotiation parameter bug in f9b42c37723ba6fbbfc31c61119fc42963770ce9.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -261,7 +298,7 @@ func (h *HTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 func (h *HTTPTransport) SetScheme(req *http.Request) {
 	if req.URL.Scheme == "" {
 		req.URL.Scheme = "http"
-		if h.TLS != nil {
+		if h.TLS != nil && req.URL.Port() != "80" {
```
mholt, line 301: So, I've read https://github.com/caddyserver/caddy/issues/4834#issuecomment-1153694521 but I'm still having a hard time understanding why this port 80 check is necessary. If the transport is configured for HTTPS, it should probably use HTTPS, right? (This is a reverse proxy client after all, not the server or a web browser.)

Just wondering, if we leave it as-is, does this only break dynamic upstreams that try to use HTTPS on port 80? (If so, that's probably OK, since that's in violation of multiple specs I'm pretty sure.)

francislavoie: I don't think it makes sense at all for SRV to be giving you both HTTP and HTTPS endpoints. That's a system design issue. It should only do one or the other, not both.

mholt: :thinking: Are you sure you aren't looking for a forward proxy @kresike ?

We could probably discuss the matter of port 80 in another issue, but I think the solution for a dynamic ServerName value here is looking good, if we want to start with that.

mholt: @kresike Ah, I see. That's a very interesting use case indeed. Thanks for letting me know. (So often I have no idea how people are using this software and it kinda leaves me in the dark. So I really appreciate knowing what the specific needs are!)

mholt (approved):
Cool, thanks for working on this! It's probably the best we can do; and I like that this won't slow down any existing deployments (since placeholders aren't currently supported in the TLS ServerName, we know that any functioning deployment isn't affected). And the inefficiency is only there for sites using the new feature of dynamic ServerNames (via placeholders). So, I think that's acceptable. Thank you again!

PR #4838, opened 2022-06-12
core: add type registry to facilitate validation of type adherence to namespace requirements

mohammed90:
This is an attempt at helping Caddy module developers to validate their module implement all the necessary interfaces to fulfill a namespace. Of course this requires opt-in of developers who introduce a namespace to declare what interfaces are required to be fulfilled.

This should help us with the documentation system once we wire everything up.

The question of `is it possible for a namespace to require more than implementation of interfaces?` and how to design the registry to fit this use-case is open, but I'm not sure if such possibility exists.

Namespaces of Caddy core:
- caddy.config_loaders
- caddy.listeners
- caddy.logging.encoders
- caddy.logging.writers
- caddy.storage
- dns.providers
- http.authentication.hashes
- http.authentication.providers
- http.encoders
- http.handlers
- http.matchers
- http.precompressed
- http.reverse_proxy.circuit_breakers
- http.reverse_proxy.selection_policies
- http.reverse_proxy.transport
- http.reverse_proxy.upstreams
- tls.certificates
- tls.get_certificate
- tls.handshake_match
- tls.issuance
- tls.stek

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"reflect"
 	"runtime/debug"
 	"strconv"
 	"strings"
@@ -40,6 +41,15 @@ import (
 
 	"github.com/caddyserver/caddy/v2/notify"
 )
+
+func init() {
+	RegisterType("", []reflect.Type{
+		reflect.TypeOf((*App)(nil)).Elem(),
+	})
+	RegisterType("caddy.storage", []reflect.Type{
+		reflect.TypeOf((*StorageConverter)(nil)).Elem(),
+	})
+}
 
 // Config is the top (or beginning) of the Caddy configuration structure.
 // Caddy config is expressed natively as a JSON document. If you prefer
@@ -72,11 +82,15 @@ type Config struct {
 	// module is `caddy.storage.file_system` (the local file system),
 	// and the default path
 	// [depends on the OS and environment](/docs/conventions#data-directory).
+	// A storage `module` should implement the following interfaces:
+	// - [StorageConverter](https://pkg.go.dev/github.com/caddyserver/caddy/v2#StorageConverter)
 	StorageRaw json.RawMessage `json:"storage,omitempty" caddy:"namespace=caddy.storage inline_key=module"`
 
 	// AppsRaw are the apps that Caddy will load and run. The
 	// app module name is the key, and the app's config is the
 	// associated value.
+	// An `app` should implement the following interfaces:
+	// - [caddy.App](https://pkg.go.dev/github.com/caddyserver/caddy/v2?tab=doc#App)
 	AppsRaw ModuleMap `json:"apps,omitempty" caddy:"namespace="`
 
 	apps    map[string]App
--- a/caddy.go
+++ b/caddy.go
@@ -42,10 +42,10 @@ import (
 )
 
 func init() {
-	RegisterType("", []interface{}{
+	RegisterNamespace("", []interface{}{
 		(*App)(nil),
 	})
-	RegisterType("caddy.storage", []interface{}{
+	RegisterNamespace("caddy.storage", []interface{}{
 		(*StorageConverter)(nil),
 	})
 }
--- a/caddy.go
+++ b/caddy.go
@@ -27,7 +27,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"reflect"
 	"runtime/debug"
 	"strconv"
 	"strings"
@@ -43,11 +42,11 @@ import (
 )
 
 func init() {
-	RegisterType("", []reflect.Type{
-		reflect.TypeOf((*App)(nil)).Elem(),
+	RegisterType("", []interface{}{
+		(*App)(nil),
 	})
-	RegisterType("caddy.storage", []reflect.Type{
-		reflect.TypeOf((*StorageConverter)(nil)).Elem(),
+	RegisterType("caddy.storage", []interface{}{
+		(*StorageConverter)(nil),
 	})
 }
 
--- a/caddytest/integration/types_test.go
+++ b/caddytest/integration/types_test.go
@@ -7,8 +7,8 @@ import (
 	_ "github.com/caddyserver/caddy/v2/modules/standard"
 )
 
-// Validates Caddy's the registered internal types implement the necessary interfaces of their
-// namespaces
+// Validates that Caddy's registered internal modules implement the necessary interfaces of their
+// respective namespaces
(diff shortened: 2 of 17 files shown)


logging.go
```diff
@@ -306,9 +313,13 @@ func (sll *StandardLibLog) provision(ctx Context, logging *Logging) error {
 // are populated, all logs are emitted.
 type CustomLog struct {
 	// The writer defines where log entries are emitted.
+	// A `writrer` should implement the following interfaces:
```
francislavoie, line 316:
```suggestion
	// A `writer` should implement the following interfaces:
```

caddytest/integration/types_test.go
```diff
@@ -0,0 +1,22 @@
+package integration
+
+import (
+	"testing"
+
+	"github.com/caddyserver/caddy/v2"
+	_ "github.com/caddyserver/caddy/v2/modules/standard"
+)
+
+// Validates Caddy's the registered internal types implement the necessary interfaces of their
```
mohammed90, line 10: I forgot how to English here 😑 I'll fix it later

caddy.go
```diff
@@ -38,6 +39,15 @@ import (
 	"go.uber.org/zap"
 )
 
+func init() {
+	RegisterType("", []reflect.Type{
```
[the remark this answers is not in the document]
mohammed90: Good point! How about this?

```go
func RegisterType(namespace string, vals []interface{}) {
	var types []reflect.Type
	for _, v := range vals {
		reflect.TypeOf(v).Elem()
	}
	if _, ok := namespaceTypes[namespace]; ok {
		panic("namespace is already registered")
	}
	namespaceTypes[namespace] = types
}
```

To be used as such:
```go
caddy.RegisterType("http.authentication.hashes", []interface{}{
	(*Comparer)(nil),
})
```

I did consider variadic function (i.e. `func RegisterType(namespace string, vals []interface{})`) just now, but it'll silently fail if a user passes a slice, i.e.  `RegisterType("http.authentication.hashes", []interface{}{(*Comparer)(nil)})`. The slice version (i.e. `[]interface{}`) is more intentional.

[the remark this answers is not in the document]
mholt: Yeah, I like the proposed change there. And I agree that a variadic function isn't necessary.

I am still on the fence as to whether this is a useful/needed addition to Caddy, though; still, I want to thoroughly review it since you took the time to implement it.

types.go
```diff
@@ -0,0 +1,39 @@
+package caddy
+
+import (
+	"fmt"
+	"reflect"
+)
+
+var namespaceTypes map[string][]reflect.Type = make(map[string][]reflect.Type)
+
+func RegisterType(namespace string, vals []interface{}) {
```
mholt, line 10:
```suggestion
func RegisterNamespace(namespace string, vals []interface{}) {
```

How about this instead? A little more specific. With a comment that it maps namespaces to types expected from each namespace, solely for documentation purposes.

mholt (changes requested):
Thanks Mohammed. I am warming up to this idea, though like you said, a lot more work remains to be done if we are to go through with this idea. It kind of intimidates me, heh. It's a good start though.

mohammed90:
I forgot I had a PR for this branch already 😂 Good 😄 

The actual gain of this feature is implemented in caddyserver/xcaddy#161. The PRs build on each other.

types.go
mohammed90: Done.

francislavoie:
I think this list is up to date just in case this PR's list fell behind :thinking: https://caddyserver.com/docs/extending-caddy/namespaces

mohammed90:
> I think this list is up to date just in case this PR's list fell behind 🤔 https://caddyserver.com/docs/extending-caddy/namespaces

Took a cursory look just now. Probably missing `caddy.fs` and `events.handlers`, and maybe more.

mholt:
I'll look at this again soon, at a glance it looks reasonable and fairly simple.

What kinds of benefits will it bring to our docs? I'm just trying to really understand what problem this solves.

(Because initially, my thought goes: If a developer didn't use the right namespace, their module will obviously not work on first try. I'm not sure extra, optional, lines of code that every plugin has to do are the right way to help with that.)

mohammed90:
> I'll look at this again soon, at a glance it looks reasonable and fairly simple.
> 
> What kinds of benefits will it bring to our docs? I'm just trying to really understand what problem this solves.
> 
> (Because initially, my thought goes: If a developer didn't use the right namespace, their module will obviously not work on first try. I'm not sure extra, optional, lines of code that every plugin has to do are the right way to help with that.)

Currently there isn't any obvious link between a namespace and the interfaces it requires. Knowing which interface is needed for modules falling within a particular namespace currently is documented as part of a wall-of-text by the developer in the README. The mechanism introduced here allows our "docsystem" to read the `RegisterNamespace` line and its arguments to automate the documentation and linking between those namespaces and the interfaces.

In other words, instead of manually maintaining this [table](https://caddyserver.com/docs/extending-caddy/namespaces) in the [markdown here](https://github.com/caddyserver/website/blob/3ebfe769b55a26ad0039ace6e858103d6a51f240/src/docs/markdown/extending-caddy/namespaces.md), we can dynamically generate it like we do with the JSON config structure because we parse the `RegisterNamespace` line. The benefit of that line is being a flag for us and enabling a validation flow for the modules devs (as implemented in [xcaddy here](https://github.com/caddyserver/xcaddy/pull/161)).

mholt:
I like the idea of that -- generating that table to include third-party plugins as well -- and that's really a clever solution.

The thing that I hesitate with is the sheer complexity of this, and the fact that it's optional anyway, meaning our result would still require manual curation.

(Also I'm growing more aesthetically adverse to adding new files to our project root, heh.)

I am not completely decided against this yet, so I'm leaving it open, and I could still be convinced. I just worry that it might be just one more thing for plugin authors to have to think about. (They already don't document their modules with godoc comments...)

francislavoie:
> (Also I'm growing more aesthetically adverse to adding new files to our project root, heh.)

One of my biggest gripes with Go tbh

mholt:
I want to thank you very much for experimenting with this. I can see that it clearly took a lot of time and careful thought.

As we discussed in Slack, while there are benefits to this, they seem small compared to the relative cost of complexity and the burden of plugin maintainers to implement this manually, every time, for every plugin, in order to really be the most helpful.

To save you time, I think I will close this for now, but if through an issue discussion we find a lighter/more automatic solution to a problem that solving would have even greater benefit, I'd be totally open to that. This has been an informative experiment nonetheless. Thank you!!

Issue #4830, opened 2022-06-08
Allow blocking headers with underscores


giovannicimolin:
Hello!

I've been using Caddy to deploy a Django-based application and I've noticed that all headers are being passed forward through the `reverse_proxy` module. 

When looking at monitoring, I've noticed that bots are able to bypass the caddy's host blocking and smuggle a `Host` header to a Django application by exploiting Django's behaviour of converting headers to uppercase and translating `-` to `_` (so `Host` becomes `HOST_HEADER`).

After researching a bit, I've found [this article](https://github.security.telekom.com/2020/05/smuggling-http-headers-through-reverse-proxies.html) and noticed that I'm able to smuggle headers to the backend like this:
```
>curl -iH "HOST_HEADER:example.com:443" https://my-website.com    
HTTP/2 302 
content-type: text/html; charset=utf-8
date: Wed, 08 Jun 2022 13:17:48 GMT
location: /_util/login/?next=/
referrer-policy: same-origin
server: Caddy
server: gunicorn
vary: Cookie, Origin
x-content-type-options: nosniff
x-frame-options: DENY
content-length: 0
```
This arrives on Django with the spoofed host header.

Meanwhile, when using the normal host header, caddy blocks the request:
```
>curl -iH "HOST:example.com:443" https://my-website.com
HTTP/2 200 
server: Caddy
content-length: 0
date: Wed, 08 Jun 2022 13:17:54 GMT
```

Is there a way to strip all host headers containing `_` underscores? This is a default feature in `nginx`.
This was acknowledged by Django a while ago: https://www.djangoproject.com/weblog/2015/jan/13/security/.

giovannicimolin:
I've updated the issue description and added more details. 
I'm happy to have a go at this if you'll accept the contribution.

mholt:
I can probably do this; related to https://github.com/caddyserver/caddy/issues/3913.

giovannicimolin:
@mholt Woah, that was fast! 
I'll try it out tomorrow.

mholt:
@giovannicimolin How did your trial go? Does it work?

giovannicimolin:
@mholt Sorry, I'm having a hard time building this locally. Are the CI builds here available for download?

francislavoie:
Yep you can grab a build from here:

https://github.com/caddyserver/caddy/actions/runs/2464155310

[...]

Issue #4743, opened 2022-04-28
Problem with path matching and URI manipulation when involving double slashes


RussellLuo:
### 1. Path matcher containing double slashes won't match any request

Caddyfile:

```
:8080 {
    handle_path //prefix/* {
        reverse_proxy 127.0.0.1:9090
    }
    respond "oops"
}
```

Neither of the following requests will match `handle_path`:

```bash
$ curl http://127.0.0.1:8080/prefix/path
oops
$ curl http://127.0.0.1:8080//prefix/path
oops
```

### 2. Normal `handle_path` can match but can't strip prefix from requests containing double slashes

Caddyfile:

```
:8080 {
    handle_path /prefix/* {
        reverse_proxy 127.0.0.1:9090
    }
    respond "oops"
}

:9090 {
    respond {path}
}
```

As shown below, the second request containing double slashes will be forwarded to the upstream without being stripped:

```bash
$ curl http://127.0.0.1:8080/prefix/path 
/path
$ curl http://127.0.0.1:8080//prefix/path                                                     
//prefix/path
```

## Cause

I think the above problems are explainable, since "path cleaning" was introduced after the fix #4407:

> The solution is to unescape (convert URL-encoded characters) then clean the path, before matching.

## Possible solution

If the problems mentioned above are not the normal behaviors as expected by Caddy, I think maybe we should also introduce "path cleaning" while doing path matching (besides cleaning the source path, we should also clean the target path) and URI manipulation.

RussellLuo:
As for problem (2), I have tested the behavior of Nginx with the following config:

```
server {
    listen 8080;

    location /prefix/ {
        proxy_pass http://127.0.0.1:9090/; # The trailing slash matters.
    } 
}
```

And here are the corresponding error logs (since no upstream is defined):

```
..., request: "GET /prefix/path HTTP/1.1", upstream: "http://127.0.0.1:9090/path", host: "127.0.0.1:8080"
..., request: "GET //prefix/path HTTP/1.1", upstream: "http://127.0.0.1:9090/path", host: "127.0.0.1:8080"
```

RussellLuo:
Instead of cleaning path implicitly within path matchers, how about enhancing the [rewrite](https://caddyserver.com/docs/modules/http.handlers.rewrite) handler to let users do it explicitly if needed?

[...]

PR #4841, opened 2022-06-14
caddyhttp: Introduce strict HTTP mode (WIP)

mholt:
It has come up several times that across various parts of the HTTP app, malformed requests can cause problems. They're not necessarily direct security issues, but can have unintended consequences for upstream/backend applications that are being proxied to, or file systems with flimsy rules for filepaths, etc., when combined with the wrong configs.

In an effort to reduce possible misconfigurations and edge cases with various other systems like backend services and file systems, I'm proposing a kind of "strict HTTP mode" to be enabled by default. (**This PR is still in a draft/WIP state. Feedback and discussion is invited.**)

It will reject requests with HTTP 400 that have:

- Unencoded `;` in the query string
- `//`, `..`, or `NUL` (`%00` or `\x00`) in the path (see #4574 and #4743; would supersede #4614)
- Header fields with underscores (`_`) (see #4830)

Basically the idea of strict HTTP is to enforce clean, spec-compliant requests and reject all others.

To clarify, I'm not saying this is a good idea or that the web server/proxy should be responsible for this. In fact I am pretty sure this will break a few legitimate use cases and applications, which do want underscores in some header fields or which use semicolons in the query string or which require double slashes for some sort of semantic interpretation.

As mentioned, these qualities of an HTTP requests are not in and of themselves security vulnerabilities, but they can present problems from poorly-designed or buggy applications that depend on these parts of HTTP requests. (For example, a file server that does not sanitize the path could be vulnerable to directory traversal by blindly evaluating `..` components in the path. Or Django apps can be misled by clients adding a `HOST_HEADER` header field. Or Windows file systems do unexpected things with NUL bytes. Merely having `..` in a path is not a security problem for Caddy, nor is having a header called `HOST_HEADER`, and so on.)

This implementation is currently WIP. It enables the enforcement of these traits by default, but can be disabled entirely or piecewise by being lenient on query strings, paths, or headers according to your configuration.

Upsides:
- This change can prevent some subtle misconfigurations and security flaws when interacting with other systems. They do NOT patch any known vulnerability in Caddy itself.

Downsides:
- This will inevitably take some users by surprise and break legitimate use cases.
- Protections can't currently be overridden per-request; currently you can turn off strict mode (entirely or for certain request parts), but there's not necessarily a way to re-enable them in your HTTP routes for certain requests. You could probably get clever to re-implement them, for example using #4831 to drop headers that have underscores when going to a backend.
- This adds trivial latency to all requests.
- This doesn't actually solve any security issues, it just covers them.

The whole change is not my favorite, and that last point is really bothering me.

Instead of this change, I recommend that applications actually patch their vulnerabilities.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -111,6 +111,8 @@ type App struct {
 	// be forcefully closed.
 	GracePeriod caddy.Duration `json:"grace_period,omitempty"`
 
+	Strict *StrictOptions `json:"strict,omitempty"`
+
 	// Servers is the list of servers, keyed by arbitrary names chosen
 	// at your discretion for your own convenience; the keys do not
 	// affect functionality.
@@ -125,6 +127,13 @@ type App struct {
 
 	// used temporarily between phases 1 and 2 of auto HTTPS
 	allCertDomains []string
+}
+
+type StrictOptions struct {
+	Disabled            bool `json:"disable,omitempty"`
+	LenientQueryStrings bool `json:"lenient_query_strings,omitempty"`
+	LenientPaths        bool `json:"lenient_paths,omitempty"`
+	LenientHeaders      bool `json:"lenient_headers,omitempty"`
 }
 
 // CaddyModule returns the Caddy module information.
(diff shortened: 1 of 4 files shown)


mholt:
If we do enable this, it'll be optional (opt-in, not opt-out).

More likely to be a separate plugin too.

mholt:
Shelving this. May come back to it later if there's good reason or high demand.

Issue #4842, opened 2022-06-15
Feature request: don't use TLS on certain dynamic upstream ports received in SRV records


kresike:
The problem I'm facing when using dynamic upstreams with SRV records is that the http transport module doesn't have any way of determining whether to use TLS for the upstream or not based on the port number received from the SRV record.
The current implementation allows us to either configure TLS and handle all ports as HTTPS or not configure TLS and that makes all upstream requests to go over HTTP.

I'm proposing a new configuration option that permits specifying an optional list of ports which will be contacted using HTTP even when TLS is configured.

Will submit a PR shortly.

mholt:
Thanks for the issue and the PR!

PR #4843, opened 2022-06-15
Add SkipTLSPorts to http transport

kresike:
PR for issue #4842

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -386,7 +386,7 @@ type TLSConfig struct {
 	// return HTTP and HTTPS endpoints too.
 	// When specified, TLS will automatically be configured on the transport.
 	// The value can be a list of any valid tcp port numbers, default empty.
-	SkipTLSPorts []string `json:"skip_tls_ports,omitempty"`
+	ExceptPorts []string `json:"except_ports,omitempty"`
 }
 
 // MakeTLSClientConfig returns a tls.Config usable by a client to a backend.
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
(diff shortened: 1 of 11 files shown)


modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -369,6 +380,13 @@ type TLSConfig struct {
 	//  - "once": allows a remote server to request renegotiation once per connection.
 	//  - "freely": allows a remote server to repeatedly request renegotiation.
 	Renegotiation string `json:"renegotiation,omitempty"`
+
+	// Skip TLS ports specifies a list of upstream ports on which TLS should not be
+	// attempted even if it is configured. Handy when using dynamic upstreams that
+	// return HTTP and HTTPS endpoints too.
+	// When specified, TLS will automatically be configured on the transport.
+	// The value can be a list of any valid tcp port numbers, default empty.
+	SkipTLSPorts []string `json:"skip_tls_ports,omitempty"`
```
mholt, line 389:
```suggestion
	ExceptPorts []string `json:"except_ports,omitempty"`
```

What about this? (Current name suggests that we're skipping TLS ports, but the opposite is true.) Maybe also reword it in the documentation comment.

(And "TLS" might be redundant since we're in a TLSConfig struct.

mholt: Another thought is that it's more natural to express port numbers as integers. I know that Go uses strings, but I wonder if there's any reason a port would need to be a string...

mholt (changes requested):
Thanks for the contribution. I like this overall approach, but think a few tweaks are in order.

I'm also thinking of this problem more broadly: _User configures TLS but it should only be used for some upstreams. What are the conditions on which TLS use varies?_ I guess port number is the only valid or useful condition I can think of.

modules/caddyhttp/reverseproxy/httptransport.go
kresike: Thanks @mholt, your naming is much better than mine.

I agree that port 80 should be default because of it's historical status as plain HTTP port, but adding it has a slight chance of breaking current installations because it autoconfigures TLS, leaving it empty doesn't break or affect anything.

I chose the value simply because it fits in nicely with how GO handles port numbers and it looks clean this way. Building json configs will look awkward, given that you have to specify numbers as strings, but the code to handle it would look much more cumbersome.

mholt (approved):
LGTM, thanks! I might tweak a variable name or something after merging, but the logic and config surface is good. Thank you again!

PR #4847, opened 2022-06-21
forwardauth: Possible fix for double headers in non-2xx branch

francislavoie:
/cc @james-d-elliott


james-d-elliott (approved):
LGTM

francislavoie:
These changes are actually really challenging to explain. But basically, I had incorrect assumptions about response handling flows, and it caused doubled headers in many situations.

For the forwardauth error branch, we actually don't need that response handler at all, because what we wanted was to just proxy back the auth server's response as-is back to the client (usually with a `Location` redirect). We had both `copy_response_headers` _and_ `copy_response` in there, but turns out `copy_response` does copy headers too because it calls `finalizeResponse`. So it was causing doubled headers there.

For the forwardauth 2xx branch, the `Server: Caddy` header would be written twice because of `break` falling through to `finalizeResponse` which would copy headers even if the purpose was to fall through to subsequent handlers. It doesn't make sense to run finalize, we actually want to ignore the response in that case. So here I removed that `break` branch. This does mean that if someone actually does want to copy the proxy response back to the client after doing some other handlers, then they _must_ use `copy_response` but I think that's most correct; they may have relied on it happening without `copy_response` before because it didn't exist at the time, but the whole handle response was buggy anyways and not well documented so I'm not that worried it would really break anything for anyone.

PR #4848, opened 2022-06-21
reverseproxy: Fix panic when TLS is not configured

francislavoie:
There was a bug in https://github.com/caddyserver/caddy/pull/4843, it was only happening in s390x integration tests, causing a panic in active health checks.

The s390x tests don't run when the PR is done by a non-member of the org for security reasons, so we didn't notice it until I made another PR based on latest `master`. But I think this is the fix.


mholt (approved):
Thanks Francis, I missed that nil check somehow. I refactored it to make it read a little simpler. What do you think? (Feel free to merge if you want.)

PR #4856, opened 2022-06-28
forwardauth: Fix case when `copy_headers` is omitted

francislavoie:
See https://caddy.community/t/using-forward-auth-and-writing-my-own-authenticator-in-php/16410, apparently it didn't work when `copy_headers` wasn't used.

This is because we were skipping adding a handler to the routes in the "good response handler", but this causes the logic in `reverseproxy.go` to ignore the response handler since it's empty. Instead, we can just always put in the `header` handler, even with an empty `Set` operation, it's just a no-op, but it fixes that condition in the proxy code:

https://github.com/caddyserver/caddy/blob/10f85558ead15e119f8e9abd81c8ad55eb865f8b/modules/caddyhttp/reverseproxy/reverseproxy.go#L806-L810


mholt (approved):
Cool, thanks so much for working so diligently on this feature. This LGTM!

PR #4865, opened 2022-07-03
TCP Keepalive configuration

### LINKED ISSUE #4863: `net.TCPConn.SetKeepAlivePeriod` on HTTP(S) listener
`transport.http.keepalive_interval`  `defines how to communicate with the backend. Default is http`.
Something similar is needed for client (my mobile device) -> Caddy, because by default Go uses a 15-second interval, which is unnecessary traffic for a battery-powered device that could be optimized away.

Nginx and Apache both use the Linux default value, 2 hours.

I can attempt a PR soon.

Reference: https://caddy.community/t/tcp-keepalives-from-client-to-caddy-not-to-upstream/16466

karmanyaahm:
See #4863  for more info

### Implementation

This is implemented at the `fakeCloseListener` layer, because during configuration reload it's unique. However,  reloading the keepalive interval does not update it for existing connections that are being proxied. It's only set when new connections are `Accept`ed.

Why not use the obvious approach with ListenConfig? Because when reloading, it would have to stop the existing listener and start a new one with the new ListenConfig. There wouldn't be a seamless switch between the old and new listeners, since only one can bind at one time. There's no way of updating the keepalive period after creating a `net.TCPListener` afaict.

I created a new function `ListenTimeout` to avoid breaking the API to `Listen` for modules.

Does this approach work? If so, I will proceed to add comments, documentation, tests, etc.

Fixes #4863

### DIFF
--- a/listen.go
+++ b/listen.go
@@ -20,10 +20,16 @@ import (
 // the socket have been finished. Always be sure to close listeners
 // when you are done with them, just like normal listeners.
 func Listen(network, addr string) (net.Listener, error) {
-	lnKey := network + "/" + addr
+	// a 0 timeout means Go uses its default
+	return ListenTimeout(network, addr, 0)
+}
+
+func ListenTimeout(network, addr string, keepalivePeriod time.Duration) (net.Listener, error) {
+	lnKey := network + "/" + addr + "/"
 
 	sharedLn, _, err := listenerPool.LoadOrNew(lnKey, func() (Destructor, error) {
 		ln, err := net.Listen(network, addr)
+
 		if err != nil {
 			// https://github.com/caddyserver/caddy/pull/4534
 			if isUnixNetwork(network) && isListenBindAddressAlreadyInUseError(err) {
@@ -37,7 +43,13 @@ func Listen(network, addr string) (net.Listener, error) {
 		return nil, err
 	}
 
-	return &fakeCloseListener{sharedListener: sharedLn.(*sharedListener)}, nil
+	return &fakeCloseListener{sharedListener: sharedLn.(*sharedListener), keepalivePeriod: keepalivePeriod}, nil
+}
+
+// typically applies only to TCP, but using interface for future proofing
+type canSetKeepAlive interface {
+	SetKeepAlivePeriod(d time.Duration) error
+	SetKeepAlive(bool) error
 }
 
 // fakeCloseListener is a private wrapper over a listener that
@@ -50,7 +62,8 @@ func Listen(network, addr string) (net.Listener, error) {
 // is reused. This type is atomic and values must not be copied.
 type fakeCloseListener struct {
 	closed          int32 // accessed atomically; belongs to this struct only
-	*sharedListener       // embedded, so we also become a net.Listener
+	keepalivePeriod time.Duration
+	*sharedListener // embedded, so we also become a net.Listener
 }
 
 func (fcl *fakeCloseListener) Accept() (net.Conn, error) {
@@ -62,6 +75,14 @@ func (fcl *fakeCloseListener) Accept() (net.Conn, error) {
 	// call underlying accept
 	conn, err := fcl.sharedListener.Accept()
 	if err == nil {
+		// if 0, do nothing, Go's default is already set
+		if tconn, ok := conn.(canSetKeepAlive); ok && fcl.keepalivePeriod != 0 {
+			if fcl.keepalivePeriod > 0 {
+				tconn.SetKeepAlivePeriod(fcl.keepalivePeriod)
+			} else { // negative
+				tconn.SetKeepAlive(false)
+			}
+		}
 		return conn, nil
 	}
 
--- a/listen.go
+++ b/listen.go
@@ -78,10 +78,12 @@ func (fcl *fakeCloseListener) Accept() (net.Conn, error) {
 		// and if the connection allows setting KeepAlive, set it
 		if tconn, ok := conn.(canSetKeepAlive); ok && fcl.keepalivePeriod != 0 {
 			if fcl.keepalivePeriod > 0 {
-				tconn.SetKeepAlivePeriod(fcl.keepalivePeriod)
+				err = tconn.SetKeepAlivePeriod(fcl.keepalivePeriod)
 			} else { // negative
-				tconn.SetKeepAlive(false)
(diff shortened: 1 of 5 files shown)


mholt:
Hey thanks, this looks like a great start!

FWIW, I do have a WIP branch that I hope will someday replace the current reload logic: https://github.com/caddyserver/caddy/pull/4705 -- this follows a more traditional approach and I think would make this even easier. I'm just waiting for one sponsor to confirm that this approach fixes an issue they have, and if so, it'll probably become mainline.

Even if it doesn't fix their issue, it's still a simpler approach (but works only on POSIX systems AFAIK). And there's a good chance it will at least replace this code for compatible OSes.

Maybe this enhancement should branch from that branch instead of master? :thinking: Would that simplify things and make for a more elegant config you think?

karmanyaahm:
I looked at #4705 and I should be able to merge this change on top of that. That will indeed make the Linux version (`listen_linux.go`) of the code simpler and more performant.
However, the changes to `listen.go` would be pretty much identical to the contents of this PR right now, if this feature is to be supported on non-Linux platforms as well.

Should I make a new PR targeting the `reuseport` branch? Or should I rebase this PR (targeting `master`) onto reuseport?

> that I hope will someday replace the current reload logic ... I'm just waiting for one sponsor to confirm that this approach fixes an issue they have, and if so, it'll probably become mainline.

In this context, does sponsor mean #4863, a person, or something else?

mholt:
@karmanyaahm Sounds great. If it's not too much trouble for ya, then I think targeting the reuseport branch would be most productive.

> In this context, does sponsor mean https://github.com/caddyserver/caddy/issues/4863, a GitHub sponsor, or something else?

Ah, just a sponsor of the project; a company using Caddy.

karmanyaahm:
I realized I can change the target branch on GitHub, so I did that. 
Also, the Linux implementation was surprisingly simple (https://github.com/caddyserver/caddy/commit/ef502898ea480cca26159c87a0ad17b134be886d).

I will try to add some tests now.

karmanyaahm:
Also, can you authorize workflows again? I think changing the branch revoked it.

listen.go
```diff
@@ -37,7 +43,13 @@ func Listen(network, addr string) (net.Listener, error) {
...
 
-	return &fakeCloseListener{sharedListener: sharedLn.(*sharedListener)}, nil
+	return &fakeCloseListener{sharedListener: sharedLn.(*sharedListener), keepalivePeriod: keepalivePeriod}, nil
+}
+
+// typically applies only to TCP, but using interface for future proofing
+type canSetKeepAlive interface {
+	SetKeepAlivePeriod(d time.Duration) error
+	SetKeepAlive(bool) error
 }
```
mholt, line 53: This is only used on L79 right? I wonder if this should be defined on L78 or thereabouts.

Also, I'm not sure what we're future proofing against. Are you thinking of something I'm not? :sweat_smile: UDP doesn't have connections, and Unix sockets don't need keepalives right?

mholt (changes requested):
This is really nice so far! You're right, the Linux implementation was really easy.

I have mostly nits, and since this isn't going into `master` I'm inclined to merge it sooner rather than later.

Other than the comments I made, I am just wondering about the `Listen` + `ListenTimeout` duality. Since it only applies to TCP in the first place, I wonder if it should even be `ListenTCP`, and not accept a network parameter. (Oh, except the network also tells us IPv4/IPv6 potentially... hmm.)

I dunno. I just need to think about it some more. But I might merge this after your revisions and then change it later?

karmanyaahm:
> Other than the comments I made, I am just wondering about the `Listen` + `ListenTimeout` duality. Since it only applies to TCP in the first place, I wonder if it should even be `ListenTCP`, and not accept a network parameter. (Oh, except the network also tells us IPv4/IPv6 potentially... hmm.)

This, and the `canSetKeepAlive` interface is emulating the documentation of net.ListenConfig.KeepAlive (below), which also intentionally keeps TCP out of the picture. Practically, this doesn't matter and is just about code style, so I can do it whatever way you like.

https://github.com/golang/go/blob/c1a4e0fe014568501b194eb8b04309f54eee6b4c/src/net/dial.go#L610-L618

modules/caddyhttp/server.go
```diff
@@ -65,6 +65,8 @@ type Server struct {
 	// 5m is applied to help avoid resource exhaustion.
 	IdleTimeout caddy.Duration `json:"idle_timeout,omitempty"`
 
+	KeepAliveInterval caddy.Duration `json:"keepalive_interval,omitempty"`
```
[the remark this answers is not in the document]
karmanyaahm: fixed

listen.go
```diff
@@ -78,10 +78,12 @@ func (fcl *fakeCloseListener) Accept() (net.Conn, error) {
...
 			if fcl.keepalivePeriod > 0 {
-				tconn.SetKeepAlivePeriod(fcl.keepalivePeriod)
+				err = tconn.SetKeepAlivePeriod(fcl.keepalivePeriod)
 			} else { // negative
-				tconn.SetKeepAlive(false)
+				err = tconn.SetKeepAlive(false)
 			}
+			// FIXME: how to deal with these errors, on one hand, this keepalive setting is not critical
+			// on the other, the user should be informed if theses some error
 		}
```
karmanyaahm, line 87: How should I deal with this? Not sure what the right logger is in that context.

listen_linux.go
```diff
@@ -4,13 +4,19 @@ import (
...
 	"syscall"
+	"time"
 
 	"go.uber.org/zap"
 	"golang.org/x/sys/unix"
 )
 
 func Listen(network, addr string) (net.Listener, error) {
-	config := &net.ListenConfig{Control: reusePort}
+	// 0 timeout means Go picks the default
```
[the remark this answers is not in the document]
karmanyaahm: It's Go, which is why this issue even exists, unfortunately. The kernel value is configurable, but Go doesn't respect that.

listen.go
francislavoie: If you need logging from somewhere relatively global, you can use `caddy.Logger()`.

mholt:
Thank you for the contribution! I'm hoping the reuseport branch will become mainstream here pretty soon.

PR #4870, opened 2022-07-07
caddyhttp: Implement `caddy respond` command

mholt:
The `caddy respond` command is for simple HTTP servers that write hard-coded responses to the client. This is useful for testing, development, and maintenance.

This requires storing the current context instead of just the current config. The context gives access to the config and the loaded modules. This was needed so we can get the listener addresses if :0 is used.

Also disables config persistence on the file-server and reverse-proxy subcommands, since you'd probably never want `--resume` to use the last command you ran (you'd only want to use that to resume actual configs like from API requests).

This command supports port ranges to spin up multiple servers and pipes for input as well.

This enables some neat functionality, like:

```bash
$ caddy respond  # empty 200
```
```bash
$ caddy respond --status 404
```
```bash
$ caddy respond "Hello, world!"
```
```bash
$ echo "I'm server {{.N}} on port {{.Port}}" | caddy respond --listen :2000-2004

Server address: [::]:2000
Server address: [::]:2001
Server address: [::]:2002
Server address: [::]:2003
Server address: [::]:2004

$ curl 127.0.0.1:2002
I'm server 2 on port 2002
```
```bash
$ cat maintenance.html | caddy respond --status 503
```

So yeah, the body supports some very basic template functionality.

TODO:

- ~~Might rename to `caddy http` instead? Not sure. (nah)~~
- [x] For quick and easy empty bodies with a custom status code, support `caddy respond <code>`.
- [x] Ability to set headers.

Feedback welcomed!

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -971,9 +971,9 @@ func handleConfigID(w http.ResponseWriter, r *http.Request) error {
 	id := parts[2]
 
 	// map the ID to the expanded path
-	currentCfgMu.RLock()
+	currentCtxMu.RLock()
 	expanded, ok := rawCfgIndex[id]
-	defer currentCfgMu.RUnlock()
+	defer currentCtxMu.RUnlock()
 	if !ok {
 		return APIError{
 			HTTPStatus: http.StatusNotFound,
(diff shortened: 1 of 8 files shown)


francislavoie:
I think we should wait until we merge cobra before merging this, cause then we can use some of cobra's features to spiffy up the CLI to make it more ergonomic

mholt:
@francislavoie I don't think waiting is necessary; we didn't wait for any of the other subcommands. I do need to get around to that PR soon anyway.

francislavoie:
What I mean is we can use features that cobra provides that we couldn't use with Go's native CLI stuff to make the interface more ergonomic. If we merge this first, then we'd probably need to make an additional PR which might break/change the interface afterwards. I don't think we can do like `-h="Foo: Foo" -h="Bar: Bar"` right now with Go's tooling, but cobra would let us do that I think.

mholt:
The cobra PR is already a breaking change, sort of (we don't officially document single-dash flags `-` as far as I know) -- and I don' know why adding short forms would be a breaking change.

And besides, I'm thinking about waiting to document this command until later anyways. Maybe merging it soon-ish so people who want it now can use it easily, then documenting it later on after we get some feedback.

PR #4871, opened 2022-07-07
templates: Access to proxy response (closes #4420)

### LINKED ISSUE #4420: Can we make reverse proxy response available in templates
This is in response to my [question here](https://caddy.community/t/can-i-get-the-response-of-a-reverse-proxy-request-in-a-template/14195) on the community forum.

Can we add a feature where the response to a reverse proxy can be made available to or passed as argument(s) to templates?

```
:9000

reverse_proxy 127.0.0.1:3000 {
              handle_response {
                              templates
                              try_files posts.html
                              file_server
              }
}
```

And if the response from the 127.0.0.1:3000 application is:
```
[
  {
    "id": 1,
    "title": "Post 1"
  },
  {
    "id": 2,
    "title": "Post 2"
  },
  {
    "id": 3,
    "title": "Post 4"
  },
]
```

Being able to do something like this:
```
<!DOCTYPE html>
<html>
	<body>
		<main>
			<div class="post-container">
                                 {{range $field, $val := .Res.Body}}
                                         	<p id={{$val.id}}>{{$val.title}}</p>
                                  {{end}}
			</div>
		</main>
	</body>
</html>
```

mholt:
This WIP branch allows accessing the reverse proxy response in `templates`:

```
Content-Type: {{.ProxyResponse.Header.Get "Content-Type"}}
Body: {{call .ProxyResponse.Body 1024}}
```

This, of course, reads and drains the response body. I'm not really sure what I'm doing here, so I'm deferring to @francislavoie for a review on this. Maybe there's a better way to handle this. The argument 1024 here is the maximum size to read (since the body has to be buffered).

Accessing the body is admittedly a little awkward currently, since you have to use `call`. I tried without `call` and it said `Body has arguments but cannot be invoked as function`, so :man_shrugging: I guess you can't invoke functions as functions when they're struct fields: https://go.dev/src/text/template/exec.go?s=5814:5878

@byoungdale could you please give this a try?


modules/caddyhttp/templates/tplcontext.go
```diff
@@ -438,6 +439,30 @@ func (c TemplateContext) funcHumanize(formatType, data string) (string, error) {
 	return "", fmt.Errorf("no know function was given")
 }
 
+// ProxyResponse accesses the response from a reverse proxy.
+// It can only be used inside a handle_response route.
+// EXPERIMENTAL: This API is subject to change or removal.
+func (c TemplateContext) ProxyResponse() (proxyResponse, error) {
+	resp, err := reverseproxy.Response(c.Req)
```
francislavoie, line 446: FYI there's a lint warning here... but I think we can ignore it `response body must be closed (bodyclose)` because the body  actually does get closed here 
https://github.com/caddyserver/caddy/blob/f259ed52bb3764ce4fd5d88f1712cb43247c2639/modules/caddyhttp/reverseproxy/reverseproxy.go#L846

```suggestion
	//nolint:bodyclose
	resp, err := reverseproxy.Response(c.Req)
```

https://golangci-lint.run/usage/false-positives/#nolint-directive

mholt:
@byoungdale Is there still interest in this? Would be a shame to have to close it if we can't confirm this meets the requirements of the original feature request.

byoungdale:
I would still be interested, but unfortunately, I have no time at the moment to try it out.

mholt:
@byoungdale When will you try this?

mholt:
Closing due to inactivity :(

We can merge this if someone who needs it tests it to confirm it works for them (rather than me just assuming it works for a real use case).

frink:
I'm facing similar needs and can try. I know it's been awhile but If you want to reopen I'll test it.

mholt:
@frink Go for it :+1:

mholt:
Closing due to merge conflicts and inactivity. If anyone wants to take this up and fix the conflict, and test that it works for their use case, we can reopen.

MurrayGroves:
Would it be easy to modify this to allow passing the response body into another reverse_proxy? My usecase is to intercept all HTML responses from a service and add a script to their head section. From what I can tell the easiest way to do that would be to use some modified version of this feature to pass the response body to a service that returns a new HTML body with the script added.

hsmallbone:
@mholt I would like to modify the response body to inject a script into the head, just as @MurrayGroves suggests. I would be willing to test this PR if that's what's required.

mholt:
@MurrayGroves You probably want this: https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#intercepting-responses

@hsmallbone Yes, thank you!

PR #4872, opened 2022-07-08
reverseproxy: Adjust new TLS Caddyfile directive names

francislavoie:
We recently merged PRs https://github.com/caddyserver/caddy/pull/4784 and https://github.com/caddyserver/caddy/pull/4843 to add `renegotiation` and `except_ports`. Both of these are actually options inside of the `TLS` struct, so they should be prefixed with `tls_` for the Caddyfile config.

I also just moved around the cases in the switch to better align with the documentation and order of the options.


mholt (approved):
Good catch. Thanks, appreciate the consistency here.

PR #4879, opened 2022-07-11
admin api: expect quoted ETags

jhwz:
Follow up from https://github.com/caddyserver/caddy/pull/4579#issuecomment-1179329165

Expects the `If-Match` HTTP header to be quoted, and will return the `ETag` HTTP header in quotes.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -21,7 +21,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"expvar"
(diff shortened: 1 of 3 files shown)


admin.go
```diff
@@ -919,7 +919,7 @@ func handleConfig(w http.ResponseWriter, r *http.Request) error {
 
 		// we could consider setting up a sync.Pool for the summed
 		// hashes to reduce GC pressure.
-		w.Header().Set("ETag", r.URL.Path+" "+hex.EncodeToString(hash.Sum(nil)))
+		w.Header().Set("ETag", "\""+r.URL.Path+" "+hex.EncodeToString(hash.Sum(nil))+"\"")
```
mholt, line 922: Any reason we wouldn't use `makeEtag()` here?

jhwz:
Cool! I got caught up sorry and didn't have time to make the change

Issue #4860, opened 2022-06-30
Add HTTP 103 Early Hints support


otbutz:
[RFC 8297](https://httpwg.org/specs/rfc8297.html) is only a draft, but is now supported by Chrome 103:

https://caniuse.com/mdn-http_status_103

and Mozilla also intends to support it:

https://bugzilla.mozilla.org/show_bug.cgi?id=1407355

Apache already implements it: https://httpd.apache.org/docs/2.4/howto/http2.html#earlyhints

It would be nice if caddy would add (experimental) support for it. Following the RFC recommendations, this should only be added for HTTP/2 and HTTP/3:

> Therefore, a server might refrain from sending 103 (Early Hints) responses over HTTP/1.1 unless the client is known to handle informational responses correctly.

otbutz:
The very first step would be to make sure that caddy processes HTTP 103 properly while acting as a reverse proxy. That may already be the case.

francislavoie:
@dunglas has been working on adding support for early hints to the Go stdlib, I'm sure he'll follow up soon to make sure Caddy supports it sufficiently.

dunglas:
Indeed it's on my todo list. We'll have to wait for Go 1.19 first.

bt90:
@dunglas does this also add support for HTTP3 based connections?

dunglas:
@bt90 yes! I also patched the HTTP/3 library: https://github.com/lucas-clemente/quic-go/pull/3047

pc-erin:
Is there any documentation on how use this feature now that it's implemented?

If we do `respond 103` does that send hints but still maintain the connection for the main response?

mholt:
@pc-erin HTTP 103 is for applications to emit. The only "application" Caddy really offers (at least, built-in) is the static file server; I have been considering building a feature that will parse HTML and send 103 responses accordingly (like for CSS, JS, and image resources), but maybe at a future date.

For now, this relays 103s from proxy backends down to the client.

What's your use case?

bt90:
Using it in a reverse proxy scenario might also be neat.  Add a matcher for the main page and tell the clients to also load files X, Y and Z.

mholt:
Proposal implemented in #5006 -- @pc-erin and @bt90 you may want to try it out!

xfalcox:
Hey @mholt,

[...]

PR #4882, opened 2022-07-12
reverseproxy: Support 1xx status codes (HTTP early hints)

dunglas:
Fixes https://github.com/caddyserver/caddy/issues/4860.
Backport of https://github.com/golang/go/pull/53164

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -23,9 +23,11 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/httptrace"
 	"net/textproto"
 	"net/url"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -40,7 +42,11 @@ import (
 	"golang.org/x/net/http/httpguts"
 )
 
+var supports1xx bool
+
 func init() {
+	supports1xx = !regexp.MustCompile(`^go1\.1(?:7|8)\.`).Match([]byte(runtime.Version()))
+
 	caddy.RegisterModule(Handler{})
 }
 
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -730,6 +736,25 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
...
+				// Clear headers, it's not automatically done by ResponseWriter.WriteHeader() for 1xx responses
+				for k := range h {
+					delete(h, k)
+				}
+
+				return nil
+			},
+		}
+		req = req.WithContext(httptrace.WithClientTrace(req.Context(), trace))
+	}
```
francislavoie, line 756: Could we move this to a helper function instead? This functions is pretty big already, don't want to pollute it more, I'd rather chunk things up into smaller functions.

mholt: I like that idea, and since it's a non-invasive chore, I went ahead and pushed this change. Hope that's OK!

mholt (approved):
I have not tested this myself, but the change is sound IMO. I am excited to merge this so that people can start testing it more easily (they will need Go 1.19 obviously).

We just tagged a release today so this should give a fair amount of time for bleeding-edge users to try it out before the next release.

Thank you for the excellent help/support and implementation, @dunglas !

dunglas:
This should be ready now that Caddy supports Go 1.19 (#4925).

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -40,7 +42,11 @@ import (
 	"golang.org/x/net/http/httpguts"
 )
 
+var supports1xx bool
+
 func init() {
+	supports1xx = !regexp.MustCompile(`^go1\.1(?:7|8)\.`).Match([]byte(runtime.Version()))
```
francislavoie, line 48: Could this be simplified to just match Go 1.18, since that's our minimum now?

mholt:
Cool. I was waiting for a sponsor to test this out in their company, but they seem to have been delayed somewhat. I'm good with merging this whenever. I haven't had a chance to do my own testing of this but I assume it works given the history here.

subhashchy:
hey Matt, 

The setup works perfectly.  good to go.   Thanks for the making this quick. 

Took us a while to implement the necessary changes like Caching html link tags etc to fully utilise this new header. 

Thanks @dunglas 

I will share the results soon.

mholt:
That's awesome!! Thanks for testing it!

PR #4894, opened 2022-07-17
Create codeql.yml

mholt:
GitHub recommended this. I'm just curious...

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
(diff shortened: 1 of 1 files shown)


mholt:
Every single one of these alerts is a false positive.

> This log write receives unsanitized user input from here.

AFAIK it's not possible to maliciously deform zap's output due to its encoders properly sanitizing the input. This check is naive.

> Sensitive data returned by HTTP request headers is logged here.

We don't log sensitive headers, but the static analyzer doesn't know this. Again, it is naive.

> Sensitive data returned by an access to fakePassword is logged here.

lol, oh my, this one is just completely wrong. It doesn't know what "fake" means, nor does the code path make any sense for this alert. We're logging a module ID, which has nothing to do with a password, even a fake one :upside_down_face: 

Yeah, this tool is outputting rubbish, so I'm going to pass on it. (These are the same kinds of low-effort, invalid "security reports" I get quite frequently in my inbox.) Was worth checking out though.

PR #4900, opened 2022-07-20
cmd: Fix reload with stdin

### LINKED ISSUE #4899: caddy reload from stdin with default admin address fails on versions >= 2.5.0
After the 2.5.0 update, `caddy reload --config -` fails with the error: `reload: couldn't determine admin API address: adapting config using caddyfile: EOF`

# Reproduce

Here's a little shell script that depends on docker and curl:

```shell
#!/usr/bin/env bash

docker run -d --name caddy -p 8080:80 caddy:2.5.0

sleep 3

echo "--- BEFORE ---"
curl -s localhost:8080 | head -c 16
echo "--------------"

echo -e 'localhost:80\nrespond "hello world"' | docker exec -i caddy caddy reload --config - --adapter caddyfile

echo "--- AFTER ---"
curl -s localhost:8080 | head -c 16
echo "-------------"

docker logs caddy

docker stop caddy
docker rm caddy
```

Both `curl` calls return the same default html (truncated for readability), and reload returns the error above.

Changing the caddy image to `caddy:2.4.6` or lower, or adding `--address localhost:2019` to the reload line results in the second curl returning the expected `hello world`

As far as I can tell, this is due to [`cmdReload`](https://github.com/caddyserver/caddy/blob/8bdee04651d93a5a799f6816ceafaa7ac61fe26d/cmd/commandfuncs.go#L298) calling [`LoadConfig`](https://github.com/caddyserver/caddy/blob/8bdee04651d93a5a799f6816ceafaa7ac61fe26d/cmd/commandfuncs.go#L305) followed by [`DeterminAdminAPIAddress`](https://github.com/caddyserver/caddy/blob/8bdee04651d93a5a799f6816ceafaa7ac61fe26d/cmd/commandfuncs.go#L313), which then [also calls `LoadConfig`](https://github.com/caddyserver/caddy/blob/8bdee04651d93a5a799f6816ceafaa7ac61fe26d/cmd/commandfuncs.go#L747) if `--address` is not specified and the `--config` is.  `LoadConfig` will read stdin successfully the first time, but fail the second.

These changes were introduced [here](https://github.com/caddyserver/caddy/commit/bbad6931e30a2e74b3f53fff797d1115cc9dd491)

# Possible fixes
- Load config outside of DetermineAdminAPIAddress, not sure of the impact this has on other call sites
- Return a more specific error if stdin is used without `--address`.  I don't think this is desired behavior however

Thanks for taking a look!

francislavoie:
Fix #4899

The problem was that `LoadConfig` would be called twice, but when using `stdin` that isn't possible.

The fix is to pass in the already-loaded config if available to `DetermineAdminAPIAddress` (which only happens for `reload`) and `nil` otherwise, which makes it skip doing its own loading in that case.

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -280,7 +280,7 @@ func cmdStop(fl Flags) (int, error) {
 	configFlag := fl.String("config")
 	configAdapterFlag := fl.String("adapter")
 
-	adminAddr, err := DetermineAdminAPIAddress(addrFlag, configFlag, configAdapterFlag)
+	adminAddr, err := DetermineAdminAPIAddress(addrFlag, nil, configFlag, configAdapterFlag)
 	if err != nil {
 		return caddy.ExitCodeFailedStartup, fmt.Errorf("couldn't determine admin API address: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thanks for the patch!

PR #4905, opened 2022-07-22
reverseproxy: Implement read & write timeouts for HTTP transport

### LINKED ISSUE #3841: reverseproxy: Add read_timeout/send_timeout options for HTTP transport
Per the doc of [reverse_proxy](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy) directive, the timeout options available for transports are:
- HTTP: dial_timeout, tls_timeout
- FastCGI: dial_timeout, read_timeout, write_timeout (not documented, but found [in code](https://github.com/caddyserver/caddy/blob/937ec342010894f7a6239883b10f6a107ff82c9f/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go#L68-L75))

For the same reason listed [here](https://github.com/caddyserver/caddy/issues/2113#issuecomment-408840525), the read_timeout (like Nginx's [proxy_read_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout)) and send_timeout (like Nginx's  [proxy_send_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_send_timeout)) options are also very useful for HTTP transport.

mholt:
Should close #3841. /cc @RussellLuo

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
+	"go.uber.org/zap"
 	"golang.org/x/net/http2"
 )
 
@@ -210,6 +211,7 @@ func (h *HTTPTransport) NewTransport(ctx caddy.Context) (*http.Transport, error)
 			}
 
 			// if read/write timeouts are configured and this is a TCP connection, enforce the timeouts
+			// by wrapping the connection with our own type
 			if tcpConn, ok := conn.(*net.TCPConn); ok && (h.ReadTimeout > 0 || h.WriteTimeout > 0) {
 				conn = &tcpRWTimeoutConn{
 					TCPConn:      tcpConn,
@@ -528,23 +530,32 @@ type KeepAlive struct {
 }
 
 // tcpRWTimeoutConn enforces read/write timeouts for a TCP connection.
+// If it fails to set deadlines, the error is logged but does not abort
+// the read/write attempt (ignoring the error is consistent with what
+// the standard library does: https://github.com/golang/go/blob/c5da4fb7ac5cb7434b41fc9a1df3bee66c7f1a4d/src/net/http/server.go#L981-L986)
 type tcpRWTimeoutConn struct {
 	*net.TCPConn
 	readTimeout, writeTimeout time.Duration
+	logger                    *zap.Logger
 }
 
 func (c *tcpRWTimeoutConn) Read(b []byte) (int, error) {
-	err := c.TCPConn.SetDeadline(time.Now().Add(c.readTimeout))
-	if err != nil {
-		return 0, err
+	if c.readTimeout > 0 {
+		err := c.TCPConn.SetReadDeadline(time.Now().Add(c.readTimeout))
+		if err != nil {
+			c.logger.Error("failed to set read deadline", zap.Error(err))
+			return 0, err
+		}
 	}
 	return c.TCPConn.Read(b)
 }
 
 func (c *tcpRWTimeoutConn) Write(b []byte) (int, error) {
-	err := c.TCPConn.SetDeadline(time.Now().Add(c.writeTimeout))
-	if err != nil {
-		return 0, err
+	if c.writeTimeout > 0 {
+		err := c.TCPConn.SetWriteDeadline(time.Now().Add(c.writeTimeout))
+		if err != nil {
+			c.logger.Error("failed to set write deadline", zap.Error(err))
+		}
 	}
 	return c.TCPConn.Write(b)
 }
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -88,6 +88,12 @@ type HTTPTransport struct {
 	// The size of the read buffer in bytes. Default: `4KiB`.
 	ReadBufferSize int `json:"read_buffer_size,omitempty"`
 
+	// The maximum time to wait for next read from backend. Default: no timeout.
+	ReadTimeout caddy.Duration `json:"read_timeout,omitempty"`
+
+	// THe maximum time to wait for next write to backend. Default: no timeout.
+	WriteTimeout caddy.Duration `json:"write_timeout,omitempty"`
+
 	// The versions of HTTP to support. As a special case, "h2c"
 	// can be specified to use H2C (HTTP/2 over Cleartext) to the
 	// upstream (this feature is experimental and subject to
@@ -194,6 +200,7 @@ func (h *HTTPTransport) NewTransport(ctx caddy.Context) (*http.Transport, error)
 				network = dialInfo.Network
 				address = dialInfo.Address
 			}
+
 			conn, err := dialer.DialContext(ctx, network, address)
 			if err != nil {
 				// identify this error as one that occurred during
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -194,13 +200,24 @@ func (h *HTTPTransport) NewTransport(ctx caddy.Context) (*http.Transport, error)
...
 			conn, err := dialer.DialContext(ctx, network, address)
 			if err != nil {
 				// identify this error as one that occurred during
 				// dialing, which can be important when trying to
 				// decide whether to retry a request
 				return nil, DialError{err}
 			}
+
+			// if read/write timeouts are configured and this is a TCP connection, enforce the timeouts
+			if tcpConn, ok := conn.(*net.TCPConn); ok && (h.ReadTimeout > 0 || h.WriteTimeout > 0) {
```
francislavoie, line 213: Hmm, won't this mean that if a read timeout is configured but not a write timeout, then it will _always_ timeout _immediately_ on write, for example? Cause it'll call `c.TCPConn.SetDeadline(time.Now())` effectively.

mholt: Ah yep, I just implemented this real quick on my way out the door. I will fix that. Thanks!

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -510,6 +527,28 @@ type KeepAlive struct {
...
+type tcpRWTimeoutConn struct {
+	*net.TCPConn
+	readTimeout, writeTimeout time.Duration
+}
+
+func (c *tcpRWTimeoutConn) Read(b []byte) (int, error) {
+	err := c.TCPConn.SetDeadline(time.Now().Add(c.readTimeout))
+	if err != nil {
+		return 0, err
+	}
```
RussellLuo, line 540: One way to address the problem mentioned above is to set the deadline only when timeout is non-zero. In addition, I guess `SetReadDeadline` is enough here? For example:

```go
if d := c.readTimeout; d > 0 {
    err := c.TCPConn.SetReadDeadline(time.Now().Add(d))
    if err != nil {
        return 0, err
    }
}
```

By the way, I even found that [the error result is unhandled in readRequest of net/http](https://github.com/golang/go/blob/c5da4fb7ac5cb7434b41fc9a1df3bee66c7f1a4d/src/net/http/server.go#L981-L986).

mholt: Indeed. And that may be because we can't do anything useful if there is an error. :man_shrugging: Especially not in the general case of standard library code :sweat_smile: 

We can perhaps at least log if setting the deadline fails.

RussellLuo:
Here I have written a TCP server for test purpose:

```go
package main

import (
	"flag"
	"net"
	"time"
)

func main() {
	rt := flag.String("r", "0s", "read delay")
	wt := flag.String("w", "0s", "write delay")
	flag.Parse()

	rd, _ := time.ParseDuration(*rt)
	wd, _ := time.ParseDuration(*wt)

	ln, err := net.Listen("tcp", ":8080")
	if err != nil {
		panic(err)
	}

	for {
		conn, err := ln.Accept()
		if err != nil {
			panic(err)
		}

		go func() {
			time.Sleep(rd)

			buf := make([]byte, 1024)
			if _, err := conn.Read(buf); err != nil {
				panic(err)
			}

			time.Sleep(wd)

			_, _ = conn.Write([]byte("HTTP/1.1 200 OK\r\nContent-Length: 5\r\nContent-Type: text/plain; charset=utf-8\r\n\r\nHello"))
			_ = conn.Close()
		}()
	}
}
```

With the following Caddyfile (I added the Caddyfile support locally):

```
:80 {
                transport http {
                        read_timeout 2s
                        write_timeout 2s
                }
        }
}
```

Here are the test results:

1. No delay (at the server side) 

    ```console
    $ go run main.go
    ```

    Consume via Caddy:

    ```console
    $ curl -i http://localhost
    HTTP/1.1 200 OK
    Content-Length: 5
    Content-Type: text/plain; charset=utf-8
    Server: Caddy
    Date: Sat, 23 Jul 2022 10:29:19 GMT

    Hello
    ```

2. Read delay (at the server side) exceeds `write_timeout` (at the client side)

    ```console
    $ go run main.go -r 3s
    ```

    Consume via Caddy:

    ```console
    $ curl -i http://localhost
    HTTP/1.1 504 Gateway Timeout
    Server: Caddy
    Date: Sat, 23 Jul 2022 10:30:52 GMT
    Content-Length: 0

    ```

3. Write delay (at the server side) exceeds `read_timeout` (at the client side)

    ```console
    $ go run main.go -w 3s
    ```

    Consume via Caddy:

    ```console
    $ curl -i http://localhost
    HTTP/1.1 504 Gateway Timeout
    Server: Caddy
    Date: Sat, 23 Jul 2022 10:31:48 GMT
    Content-Length: 0

    ```

mholt:
@RussellLuo Awesome, thanks for testing it!

I made adjustments from the comments so far. Really appreciate your feedback!

francislavoie:
Needs Caddyfile support too

francislavoie (approved):

mholt:
Oh fine, @francislavoie -- have it your way :laughing:

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -510,6 +529,37 @@ type KeepAlive struct {
...
+	readTimeout, writeTimeout time.Duration
+	logger                    *zap.Logger
+}
+
+func (c *tcpRWTimeoutConn) Read(b []byte) (int, error) {
+	if c.readTimeout > 0 {
+		err := c.TCPConn.SetReadDeadline(time.Now().Add(c.readTimeout))
+		if err != nil {
+			c.logger.Error("failed to set read deadline", zap.Error(err))
+			return 0, err
```
RussellLuo, line 547: If I understand the docs of `tcpRWTimeoutConn` correctly, I think there's no need to return anymore?

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -194,13 +201,25 @@ func (h *HTTPTransport) NewTransport(ctx caddy.Context) (*http.Transport, error)
...
 				return nil, DialError{err}
 			}
+
+			// if read/write timeouts are configured and this is a TCP connection, enforce the timeouts
+			// by wrapping the connection with our own type
+			if tcpConn, ok := conn.(*net.TCPConn); ok && (h.ReadTimeout > 0 || h.WriteTimeout > 0) {
+				conn = &tcpRWTimeoutConn{
+					TCPConn:      tcpConn,
+					readTimeout:  time.Duration(h.ReadTimeout),
+					writeTimeout: time.Duration(h.WriteTimeout),
```
RussellLuo, line 219: The `logger` field needs a non-nil value?

modules/caddyhttp/reverseproxy/httptransport.go
mholt: Oops, yep, forgot to delete that line.

modules/caddyhttp/reverseproxy/httptransport.go
mholt: Dangit, yep :)

really tired today

mholt:
Thanks for the thorough review. I have not had a chance to test this and I am very tired today. It should be good now.

RussellLuo:
Never mind, just take a break :) I'll try to test this again.

RussellLuo:
I have tested the behavior of `read_timeout` (as well as Caddyfile support), and I think it works pretty well.

But I have trouble testing `write_timeout`, since I found `tcpRWTimeoutConn.Write()` always write the whole data immediately. I guess this is due to the TCP send buffer of the kernel, but I failed to mimic the timeout scenario even after I decrease the TCP buffer size according to the [instructions on Mac](https://wiki.geant.org/display/public/EK/MAC+OS+X+OS+Specific).

mholt:
@RussellLuo Awesome, thank you! Yeah, that makes sense about writing, and I'm afraid I don't know enough about the kernel to help there. :-/ It'd probably require adding significant latency to the writer somehow, which I'm not sure how to do.

Since the code is identical for read and write, I assume write works too, and will merge this.

Thank you for your help with testing!!

PR #4909, opened 2022-07-25
fileserver: Support virtual file systems

### LINKED ISSUE #3720: Feature Request: Operating system interface for file server
Currently, File server module only support file system.

I think the File server module could support more data sources, such as S3 or elsewhere.

This could be done by implementing a Certmagic storage interface, or roughly an OS compatible interface.

mholt:
This change replaces the hard-coded use of `os.Open()` and `os.Stat()` with
the use of the new (Go 1.16) `io/fs` APIs, enabling virtual file systems.
It introduces a new module namespace, `caddy.fs`, for such file systems.

Also improve documentation for the file server. I realized it was one of
the first modules written for Caddy 2, and the docs hadn't really been
updated since!

Anyway, this is super cool, since it enables Caddy's file server to serve not-so-static files, or files on remote resources like cloud storage or network devices. (S3? Database? No problem now.) It can serve files from any virtual file system, i.e. any Go type that has `Open()` and `Stat()` methods. So even if the content isn't strictly static, if the virtual file system can assemble the contents with a call to `Open()`, Caddy doesn't care whether it's static or not. Dynamic file server!

The first `caddy.fs` module I might implement (other than the unexported, implicitly-enabled, 2-liner `osFS` in this PR) is [`archiver.FileSystem()`](https://pkg.go.dev/github.com/mholt/archiver/v4#FileSystem) so that Caddy can serve files from zip/tar archives including transparently decompressing its contents!

I might still need to tweak a few things about this before finalizing.

Closes #3720


mholt:
Caddyfile support added here: https://github.com/caddyserver/caddy/commit/ebd6abcbd5f6b31b7b4cadca0a546d9f0eab9ad3

robbymilo:
@mholt New to caddy and don't see this mentioned in the docs - do you have an example on how to serve content directly from an archive file (e.g. .zip or .tar.gz)?

WeidiDeng:
It's very easy to write one, basically replace embed.FS in [caddy-embed](https://github.com/mholt/caddy-embed) with archive.FS, and create filesystem from archive file path.

ArchiveFS is too experimental to be used in production systems imo, [afero](https://github.com/spf13/afero) is an alternative that I use in my custom file server.

mholt:
@WeidiDeng ArchiveFS is new, but I don't want to discourage people from using it, otherwise it'll never be "production ready" :upside_down_face:

Thanks for helping me improve it -- I need to get back to those PRs soon!

Issue #4875, opened 2022-07-09
golangci-lint should run on multiple platforms


karmanyaahm:
Something I found when working on my PR was that platform specific code using `// go:build` is not checked in ci by  `.github/workflows/lint.yml`.

An example where Linux works, but macOS doesn't (due to listen.go and listen_linux.go in that branch being different): https://github.com/karmanyaahm/caddy/actions/runs/2639453990
See this for a possible fix: https://github.com/karmanyaahm/caddy/commit/68bd9d5d5c090a15104545539166d24a02a24fa9

francislavoie:
Feel free to open a PR. Thanks for letting us know!

PR #4913, opened 2022-07-26
ci: Run golangci-lint on multiple os(#4875)

u5surf:
#4875 

https://github.com/golangci/golangci-lint-action#multiple-os-support

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,10 @@ jobs:
   # From https://github.com/golangci/golangci-lint-action
   golangci:
     name: lint
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
@@ -25,6 +28,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.44
+          version: v1.47
           # Optional: show only new issues if it's a pull request. The default value is `false`.
-          # only-new-issues: true
+          only-new-issues: true


.github/workflows/lint.yml
```diff
@@ -14,7 +14,10 @@ jobs:
   # From https://github.com/golangci/golangci-lint-action
   golangci:
     name: lint
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
mholt, line 19: Should we add `windows-latest`?

mholt (changes requested):
Thanks for the PR! Just one question, otherwise LGTM.

.github/workflows/lint.yml
francislavoie: I think we should, yes.

We could also bump up the golangci-lint version to v1.47 https://github.com/golangci/golangci-lint/releases

u5surf: @francislavoie @mholt 
Hi, I'm trying to be upped the golangci-lint version, Although, I found the lint error(It has not occurred in the old version)
I consider that it should treat another issue once.

https://pkg.go.dev/github.com/google/cel-go/cel#Ast.ResultType
https://pkg.go.dev/github.com/google/cel-go/cel#Functions

```
run golangci-lint
  Running [/home/runner/golangci-lint-1.47.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Error: SA1019: checked.ResultType is deprecated: use OutputType  (staticcheck)
  Error: SA1019: checked.ResultType is deprecated: use OutputType  (staticcheck)
  Error: SA1019: cel.Functions is deprecated: use Function() instead to declare the function, its overload signatures, and the overload implementations.  (staticcheck)
  Error: SA1019: cel.Functions is deprecated: use Function() instead to declare the function, its overload signatures, and the overload implementations.  (staticcheck)
  Error: SA1019: cel.Functions is deprecated: use Function() instead to declare the function, its overload signatures, and the overload implementations.  (staticcheck)
```

mohammed90: Solving that requires uncommenting the line:

```
# only-new-issues: true
```

I vote for uncommenting the line, otherwise the lint errors might confuse the PRs with multiple unrelated changes.

francislavoie (approved):

.github/workflows/lint.yml
mholt: I am actually working on a fix for those errors as I type this.

I got it to compile but now the tests fail related to those `Function()` changes... I'm not sure how to use it properly.

mholt:
Is there also a reason this updated go.mod? I believe the only change should be on the yml file right?

u5surf:
> Is there also a reason this updated go.mod? I believe the only change should be on the yml file right?
@mholt 
Sorry, my bad, I mistake rebasing, I fixed it to be only my commit.

mholt:
@u5surf Any idea why the linter is complaining of `gofmt` on these files here which have already been `gofmt`'ed? https://github.com/caddyserver/caddy/actions/runs/2756929369

mohammed90:
> @u5surf Any idea why the linter is complaining of `gofmt` on these files here which have already been `gofmt`'ed? https://github.com/caddyserver/caddy/actions/runs/2756929369
> 
> It only happens on Windows, and running `gofmt -s -w` yields no changes.

I bet it's the line-endings. There are instructions on the action repo to add `.gitattributes` file for that.

francislavoie:
The line endings look fine to me, LF on all the files... 🤔

u5surf:
@mholt We should add the .gitattribute file to avoid the line-endings problem like this.
https://github.com/microsoft/vscode-dev-containers/blob/main/.gitattributes
......I thought but I guess it had already been fixed by @mohammed90 😄 
https://github.com/caddyserver/caddy/pull/4919

mholt:
We can do that. Thanks. Any idea why it only affects those 3 files, even though they already use LF?

u5surf:
@mholt 
> We can do that. Thanks. Any idea why it only affects those 3 files, even though they already use LF?

Could you tell me what 3 files' names are? I could not find those from the result of CI which you refer to.
Just so there's no misunderstanding, I can't execute the golangci-lint on Windows because I don't have any Windows environments. So I can only investigate the detail of those files or what is different from anything else.

mholt:
@u5surf They were metrics and notify files: https://github.com/caddyserver/caddy/actions/runs/2756929369

Ah, no worries, I also don't dev on Windows. Just thought I'd ask if you knew since none of us can figure it out, haha.

PR #4916, opened 2022-07-27
caddyhttp: Clear out matcher error immediately after grabbing it

francislavoie:
Fixes a bug reported in https://caddy.community/t/error-handling-handler-error-with-multiple-handle-in-handle-errors/16692

### DIFF
--- a/modules/caddyhttp/routes.go
+++ b/modules/caddyhttp/routes.go
@@ -205,7 +205,7 @@ func wrapRoute(route Route) Middleware {
 				err, ok := GetVar(req.Context(), MatcherErrorVarKey).(error)
 				if ok {
 					// clear out the error from context, otherwise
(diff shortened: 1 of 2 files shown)


mholt:
Do we need a `DeleteVar()`?

francislavoie:
Maybe? Up to you.

mholt:
I decided that it doesn't make much sense to change an existing key in the map to have a nil value. I could see a case where a key is initially added to the map with a nil value if all the consumers want to do is check to see if the key exists, but since changing a variable is rare enough, I can't think of good use cases for changing from a value-read to a key-exists-check, so I changed this to simply delete the map entry entirely if setting a non-nil key to nil.

mholt (approved):
Thank you!!

PR #4919, opened 2022-07-28
chore: Add .gitattributes to force *.go to LF

francislavoie:
https://github.com/golangci/golangci-lint/issues/580#issuecomment-704913317 🤷‍♂️

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.go text eol=lf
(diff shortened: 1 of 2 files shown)


mholt:
Wow, it works. So why does it only affect those 3 files?

mohammed90 (approved):
This seems to have fixed it  https://github.com/caddyserver/caddy/compare/56c139f003292005dd5dd974ffd770b8620e8aef...13e33bc. I also have no clue why those 3 files in particular 🤷🏼‍♂️ LGTM though

mholt:
Cool. I'm stumped as to why, but thanks!

Issue #4907, opened 2022-07-25
h2c server allows graceful shutdown


WeidiDeng:
Similar to websocket, h2c server internally [upgrade](https://github.com/golang/net/blob/master/http2/h2c/h2c.go#L74) the request and pass the resulting conn to http2 server to demux the stream to http handler.

Currently, h2c request, just like websocket, were left on their own. At least we can add [ConfigureServer](https://pkg.go.dev/golang.org/x/net/http2#ConfigureServer), which internally register graceful h2 shutdown on server shutdown. 

If really wants to properly cleanup, perhaps just copy the [h2c.go](https://github.com/golang/net/blob/master/http2/h2c/h2c.go) file and make some modifications. It's just 225 lines.

mholt:
@WeidiDeng That's good to know. We'll need to look into this. I appreciate you bringing it to our attention!

Copying over 200 lines of code would be a bummer. But if there's no other way, maybe that's just what we'll have to do.

WeidiDeng:
Copying code is indeed such a chore.

Alternatively, I can think of another way. h2c check condition is pretty easy to copy, but we cannot access the net.Conn after the hijack call. One way would be specify a http.Server.ConnContext to record the underlying net.Conn and extract the net.Conn. Using BaseContext and a custom listener is also an option, but I think that's an overkill.

Besides this, h2c handler has another bug to fix (but that doesn't affect caddy that much, because caddy specific context key uses caddy's own), h2c handler won't extract the correct http.Server instance from the request itself, I'll push a pr upstream. This causes a number of issues [52868](https://github.com/golang/go/issues/52868), [37089](https://github.com/golang/go/issues/37089)

WeidiDeng:
Upstream pr has been [submitted](https://github.com/golang/net/pull/139), I wonder how long that will take, since [http2.Transport.DialTlsContext](https://github.com/golang/net/pull/123) is stuck there for months.

francislavoie:
> I wonder how long that will take, since https://github.com/golang/net/pull/123 is stuck there for months.

Yeah... I've been waiting for a long time for that one 😩

mholt:
Awesome, thank you so much for helping with this @WeidiDeng

solracsf:
Sorry to ask, but what's the state here?

mholt:
IIRC it was a bit complicated, and just as Weidi started to get something finished, the upstream (x/net/http2) changed some things.

WeidiDeng:
[...]

PR #4920, opened 2022-07-29
adding h2c graceful shutdown

WeidiDeng:
Fixes [4907](https://github.com/caddyserver/caddy/issues/4907)


mholt:
Thanks! I'm looking forward to reviewing this.

mholt:
Thank you for this change @WeidiDeng  -- if it's OK with you, I'm going to bump this to after 2.6 (or at least after its initial beta) just so I can stay on a self-imposed schedule :zany_face: 

I will review this though!

modules/caddyhttp/h2chandler.go
```diff
@@ -0,0 +1,80 @@
...
+	"net/http"
+	"net/textproto"
+	"sync/atomic"
+	"time"
+
+	"golang.org/x/net/http/httpguts"
+)
+
+// h2chandler is a Handler which counts possible h2c upgrade requests
+type h2chandler struct {
```
francislavoie, line 15: Shouldn't this be uppercase, as a separate word?
```suggestion
type h2cHandler struct {
```

mholt:
Will take another look when it's updated and ready for another review (as discussed in Slack). :100:

mholt:
@WeidiDeng Just tidying things up... I might close this for now if that's alright with you? It's still in a draft state, and seems to have fallen by the wayside.

I'm open to this enhancement, but it does seem rather complex and probably could use some tests as well.

Thank you for having worked on this, and for all you contribute to the project!

WeidiDeng:
Actually the [upstream](https://github.com/golang/net/commit/7191757bc637cf79a7ece0546e33f903bf5e9709) has changed. So this needs to be reworked anyway.

Issue #4921, opened 2022-07-29
Serve http2 when listener wrapper doesn't return *tls.Conn


WeidiDeng:
I'm developing a caddy listener plugin that multiplex connection over tls, and dispatch them to different backend, ie, mysql over tls, ssh over tls etc. But unlike [caddy-l4](https://github.com/mholt/caddy-l4), this works as a listener wrapper plugin.

However because golang http.Server only directly support h2 when listener returns a *tls.Conn, I have to use h2c. Now http2 request has request.TLS field filled because [h2 server uses an interface to sniff tls related information](https://github.com/golang/net/blob/master/http2/server.go#L437). But http1.1 request is not, I wonder if caddy can set the field when the underlying conn implements the same interface.

mholt:
Sounds interesting. Can you help me understand a little better by showing some relevant code? Sounds like you need the `http.Request.TLS` field filled out even though TLS is not used?

WeidiDeng:
TLS is used, like [caddy-l4](https://github.com/mholt/caddy-l4). But I make it a listener wrapper to simplify caddyfile parsing. I can simply add a domain and it will have a tls certificate, no need to modify json configure to add domain. That domain will simply return 200 when opened in browser, but because it's not relevant to its functionality, I can name it sql.example.com to add mysql ssl support etc. Neccessary sni match can be done in my plugin too.

However, because go http server code, http1.1 will only have request.tls field filled when listener returns a [tls.Conn](https://github.com/golang/go/blob/master/src/net/http/server.go#L1891).

This will also interfere with [strict_sni_check](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/server.go#L306)

The main listener is like this 

```
type Listener struct {
	net.Listener
}

func (l *Listener) Accept() (net.Conn, error) {
	conn, err := l.Listener.Accept()
	if err != nil {
		return nil, err
	}

	return &Conn{
		Conn: conn,
	}, nil
}
```

The Conn definition is 

```
type Conn struct {
	net.Conn
	br   *bufio.Reader
	once sync.Once
}

func (c *Conn) Read(p []byte) (int, error) {
	var once bool
	c.once.Do(func() {
		c.br = handleConn(c.Conn)
		once = true
	})

	if once && c.br == nil {
		c.Conn.Close()
		return 0, net.ErrClosed
	}

	return c.br.Read(p)
}
```

[...]

PR #4929, opened 2022-08-03
Serve http2 when listener wrapper doesn't return *tls.Conn

WeidiDeng:
fixes [4921](https://github.com/caddyserver/caddy/issues/4921)


mholt:
Thanks for working on this!

So, my understanding of this is, that it allows us to serve HTTP/2 even if the underlying Conn is wrapped by a ListenerWrapper plugin so that if it's not a tls.Conn, it can still be recognized as one?

WeidiDeng:
> Thanks for working on this!
> 
> So, my understanding of this is, that it allows us to serve HTTP/2 even if the underlying Conn is wrapped by a ListenerWrapper plugin so that if it's not a tls.Conn, it can still be recognized as one?

Yes, because golang http.server won't directly serve http2 when return non tls.Conn, I have to manually configure [h2server](https://github.com/golang/go/issues/12737). tls.Conn still uses http.Server code path, because when handshake is not complete, it won't have "h2" as negotiated protocol.

WeidiDeng:
@mholt I noticed you don't allow serve [only http2](https://github.com/caddyserver/caddy/blob/c7772588bd44ceffcc0ba4817e4d43c826675379/modules/caddyhttp/app.go#L189), if you want, I can use http2.Server.ServeConn in this case.

mholt:
@WeidiDeng Yeah, we'd have to write our own HTTP server. I'm not sure I'm quite ready for that since it seems like a maintenance burden. You can see my attempt here: https://github.com/caddyserver/caddy/pull/4707#issuecomment-1204438638

I'll review this soon!

WeidiDeng:
@mholt I looked at you code, that's because http2.Server requires tls handshake to be complete, if there is any. To be fair, http2.Server can serve h2c with prior knowledge, that's like how currently reverse_proxy h2c transport is implemented.

But yes I agree it's a burden, especially if we are allowing serve http2 with non tls.Conn. Perhaps it's easier to configure tls nextproto directly, normal clients won't send h1 requests on a h2 tls.Conn.

francislavoie:
@WeidiDeng could you rebase this?

I'm not sure there's anything really blocking merging this. Seems like a reasonable fix for a usecase only used by caddy-l4 currently.

WeidiDeng:
@francislavoie I'll also need to update l4 listener module, that part is left as a todo.

There are some parts that I'm not sure, like using x/net/http2 instead of builtin http2, should there be an option?

modules/caddyhttp/server.go
```diff
@@ -213,6 +214,14 @@ type Server struct {
 
 // ServeHTTP is the entry point for all HTTP requests.
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.TLS == nil {
```
mholt, line 217: Maybe a comment explaining what this block is used for

modules/caddyhttp/server.go
```diff
@@ -195,10 +195,11 @@ type Server struct {
...
 	ctx          caddy.Context
 
-	server      *http.Server
-	h3server    *http3.Server
-	h3listeners []net.PacketConn // TODO: we have to hold these because quic-go won't close listeners it didn't create
-	addresses   []caddy.NetworkAddress
+	server         *http.Server
+	h3server       *http3.Server
+	h3listeners    []net.PacketConn // TODO: we have to hold these because quic-go won't close listeners it didn't create
+	http2listeners []*http2Listener
```
mholt, line 201:
```suggestion
	h2listeners []*http2Listener
```

Just for symmetry with `h3` vars I guess :)

modules/caddyhttp/http2listener.go
```diff
@@ -0,0 +1,96 @@
...
+	weakrand "math/rand"
+	"net"
+	"net/http"
+	"sync/atomic"
+	"time"
+
+	"golang.org/x/net/http2"
+)
+
+type http2Listener struct {
```
mholt, line 15: A comment explaining the usefulness of this type instead of just the http2 package types would be useful!

mholt (changes requested):
Thanks, I made a first pass and have some minor suggestions -- This looks like we're doing a lot more in-house but it opens up some new possibilities, yeah?

mholt (approved):
Thanks @WeidiDeng  -- let's give this a try. Sorry for the long time on this. Thank you for your work!

Issue #4928, opened 2022-08-02
Replace `strings.Index` usages with `strings.Cut` where possible


francislavoie:
We can now use the new `strings.Cut` function which should simplify some code. See https://tip.golang.org/doc/go1.18#strings

Chi3f1991:
Yuppie

Chi3f1991:
🔮

chir4gm:
Hi, I tried my hand at this Issue, and have requested a PR.
I hope the simple changes suffice.

ps. Thank you for the efforts you have put into making this awesome web server!

PR #4930, opened 2022-08-04
Replace strings.Index usages with strings.Cut

WilczynskiT:
fixed https://github.com/caddyserver/caddy/issues/4928

### DIFF
--- a/caddyconfig/load.go
+++ b/caddyconfig/load.go
@@ -189,12 +189,11 @@ func adaptByContentType(contentType string, body []byte) ([]byte, []Warning, err
 	}
 
 	// adapter name should be suffix of MIME type
-	slashIdx := strings.Index(ct, "/")
-	if slashIdx < 0 {
+	_, adapterName, slashFound := strings.Cut(ct, "/")
+	if !slashFound {
 		return nil, nil, fmt.Errorf("malformed Content-Type")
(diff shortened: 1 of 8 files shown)


mholt (changes requested):
Thank you for opening a PR!

I think the purpose of using `strings.Cut()` is to split a string into two pieces. `strings.Index()` is still the right function to see if a substring can be found.

There are some uses of `Index()` that still seem simpler than using `Cut()`, IMO.

I think we'd only want to keep the changes in listeners.go, load.go, and addresses.go. The others seem to be more complicated or not quite the right fit.

There may also be some occurrences of `SplitN()` that we could replace with `Cut()`, if we are splitting only into 2 pieces.

WilczynskiT:
@mholt Okey, thank you for review ;) I changed it. I will look at the `SplitN()` and will create new ticket to do it

francislavoie:
> I will look at the SplitN() and will create new ticket to do it

Please just do it in this PR, it'll be easier to review all in one place.

WilczynskiT:
Okey, I added the changes and signed the CLA. So it is ready to continue the review.

mholt (approved):
This LGTM!

Thank you for the contribution @WilczynskiT :smiley:

Issue #4765, opened 2022-05-06
Ability for http middleware handler to configure http.server


WeidiDeng:
Right now http.handler only needs to implement servehttp method, which is fine in most cases. However in some cases, for example, custom caddy app which also implements middleware may need to configure http.Server such as BaseContext, ConnState etc.

I have written a plugin which is similar to [forward proxy](https://github.com/klzgrad/naiveproxy). And since most of proxy connection is long lived, I have to register a http.Server OnShutdown function to close those when server configuration changes, or have to set a grace period which had lingering connection bug.

Right now I can access http.server instance through http.ServerContextKey and configure it there. But I think setting up a server is best done through an optional configure server method and before servehttp.

mholt:
Interesting use case; usually, we'd probably want to configure the server by making the server itself pluggable. For example, we have a separate namespace for listener wrappers that configures the server's socket/listener. I think it's best for the `http.handlers` namespace to stay only HTTP handlers.

So maybe if we need to extend its BaseContext or ConnState, maybe we need a new plugin namespace (or multiple?) to set those up.

If you want to do something when the server is shutting down, you can implement the `caddy.CleanerUpper` interface: https://pkg.go.dev/github.com/caddyserver/caddy/v2#CleanerUpper

The `CleanUp()` method will be called when the config is reloaded or shutting down.

WeidiDeng:
> Interesting use case; usually, we'd probably want to configure the server by making the server itself pluggable. For example, we have a separate namespace for listener wrappers that configures the server's socket/listener. I think it's best for the `http.handlers` namespace to stay only HTTP handlers.
> 
> So maybe if we need to extend its BaseContext or ConnState, maybe we need a new plugin namespace (or multiple?) to set those up.
> 
> If you want to do something when the server is shutting down, you can implement the `caddy.CleanerUpper` interface: https://pkg.go.dev/github.com/caddyserver/caddy/v2#CleanerUpper
> 
> The `CleanUp()` method will be called when the config is reloaded or shutting down.

[...]

PR #4933, opened 2022-08-05
caddyhttp: Support configuring Server from handler provisioning

WeidiDeng:
Fixes [4765](https://github.com/caddyserver/caddy/issues/4765).

When Middleware handler is provisioning, it can access caddyhttp.Server and register specific http.Server related callbacks.


mholt:
Thanks, I'll look at this soon!

modules/caddyhttp/server.go
```diff
@@ -146,6 +146,56 @@ type Server struct {
...
+	s.connContexts = append(s.connContexts, f)
+}
+
+// register OnShutdown callbacks
+func (s *Server) RegisterOnShutdown(f func()) {
+	s.onShutdowns = append(s.onShutdowns, f)
+}
+
+// apply callbacks to server
+func (s *Server) ConfigureServer(server *http.Server) {
```
mholt, line 172: Does this have to be exported? It should only be configured within this package, right?

modules/caddyhttp/app.go
```diff
@@ -173,7 +173,9 @@ func (app *App) Provision(ctx caddy.Context) error {
 	}
 
 	// prepare each server
+	oldContext := ctx.Context
 	for srvName, srv := range app.Servers {
+		ctx.Context = context.WithValue(oldContext, ServerCtxKey, srv)
```
mholt, line 178: What is the purpose of this?

mholt (changes requested):
Thanks for the enhancement! This looks mostly good, though I do have a couple questions.

modules/caddyhttp/server.go
WeidiDeng: Yes, it should be unexported.

modules/caddyhttp/app.go
WeidiDeng: In http middleware provision, this gives them to access current server by using context key. Using context value to get current server the middleware is handling request.

mholt (approved):
I went ahead and unexported that method, and just moved a couple things around; this looks useful, so I will be happy to merge it in. Thanks @WeidiDeng!

PR #4948, opened 2022-08-10
caddyhttp: Smarter path matching and rewriting

### LINKED ISSUE #4743: Problem with path matching and URI manipulation when involving double slashes

### LINKED ISSUE #4801: `path_regexp` matching whitespace breaks `uri strip_prefix`
It would seem that when matching a request path with `path_regexp`, if the resulting match includes whitespace, using the match as a part of a following `uri strip_prefix` causes the stripping to break.

Example Caddyfile:
```
{
    debug
}

localhost:80 {
    @jpegmatch {
        path /api/static/*
        path_regexp sample_params /api/static/([^/]+)/.*.(jpg|jpeg|png)$
    }

    handle @jpegmatch {
        uri strip_prefix /api/static/{http.regexp.sample_params.1}
        root * /app/data/{http.regexp.sample_params.1}/images/
        file_server
    }
}
```

Excerpt from debug log output:

```
{
    "level":"debug",
    "ts":1653058747.7992723,
    "logger":"http.handlers.file_server",
    "msg":"sanitized path join",
    "site_root":"/app/data/1517 (2)/images/",
    "request_path":"/api/static/1517 (2)/icon_da4e50dd-f069-4fcd-9088-7fbb08711a04.jpg",
    "result":"/app/data/1517 (2)/images/api/static/1517 (2)/icon_da4e50dd-f069-4fcd-9088-7fbb08711a04.jpg"
}
```

If I use the following `handle @jpegmatch` instead, everything works as expected:

```
    handle @jpegmatch {
        uri path_regexp /api/static/([^/]+)/ /
        root * /app/data/{http.regexp.sample_params.1}/images/
        file_server
    }
``` 

Log excerpt from above `handle @jpegmatch`:

```
{
    "level":"debug",
    "ts":1653058951.8403285,
    "logger":"http.handlers.file_server",
    "msg":"sanitized path join",
    "site_root":"/app/data/1517 (2)/images/",
    "request_path":"/icon_da4e50dd-f069-4fcd-9088-7fbb08711a04.jpg",
    "result":"/app/data/1517 (2)/images/icon_da4e50dd-f069-4fcd-9088-7fbb08711a04.jpg"
}
```

### LINKED ISSUE #4923: file-server returns 500 when filename contains some special characters
I have a file named "*%@.exe*" and when I use `file-server --browse` to view the directory everything is OK. While I click on this file to download it, Caddy returns 500.

In the log, I found this:

> invalid URL escape "%@."

After testing, there is no problem with nginx.

mholt:
This branch resolves several inconsistencies across Caddy's HTTP facilities regarding URI encodings in paths.

I am not entirely sure, but I suppose breaking changes might be possible if users relied on buggy behavior that has only just been determined and is being remedied here.

This PR mainly affects the `path` matcher and the `rewrite` middleware (including both the `rewrite` and `uri` Caddyfile directives). These are extremely commonly-used Caddy features.

## Background

URIs (essentially the part of the URL after the scheme and authority/host, e.g. `/foo/bar?a=b#frag` -- though servers don't really deal with `#fragment` components) are famous for being inconsistently encoded and parsed. Differences in parsing/handling between servers, proxies, and applications often lead to bugs and security vulnerabilities. For example, a path of `//foo/bar` might be considered equivalent to `/foo/bar` by one piece of infrastructure, and different to another. Similarly, `/foo%2Fbar` might or might not be the same as `/foo/bar`. To a router, they could be different. To an application, they could be the same.

A web server like Caddy is between a rock and a hard place, because it finds itself between untrusted clients who send all manner of inconsistent requests, and other servers or applications who expect the request URI to be _just right_. Caddy is often expected to route requests of all varieties and rewrite/transform them into something the backend application (even if that's just the built-in static file server) can use without confusion. The problem is the requirements and expectations vary widely!

Caddy has had several issues over the years where some users expect a URI like `/foo%2Fbar` to be transformed into `/foo/bar` before being proxied. Some want `/foo/bar` to match `/foo%2Fbar`, while others don't. Some want a matcher like `/secret/*` to match URIs like `//secret/*` or `/secret//*` because they put it behind authentication, and if it doesn't match, auth could be bypassed! Windows treats `/file.php . ..` the same as `/file.php` -- even though they technically have different suffixes and file extensions, [causing routing blunders](https://github.com/caddyserver/caddy/pull/2917). Then imagine a path prefix like `/bands/*/*/` that should match `/bands/Pink/Try/` as well as `/bands/AC%2FDC/T.N.T` -- but if the path matcher normalizes (decodes) URIs before matching, the first URI would work but the second would become `/bands/AC/DC/T.N.T` which doesn't match the pattern anymore. To make matters worse, any given URI has multiple valid encoded forms. `%2F%66%6F%6F%2F%62%61%72` can be decoded to `/foo/bar` just as well as `/foo%2Fbar` can, and everything in between can, too. If routers matched on non-normalized URIs, there would be plenty more security bugs to deal with: a pattern of `/foo/*`, which is expected to be authenticated, would no longer match `/foo%2Fbar` even though they are, according to ratified RFCs, _equivalent_.

In other words, encodings are significant to applications, but normalizing URIs to a consistent form is critical for maintaining security.

Let me restate here [what I wrote for the Laravel community](https://github.com/laravel/framework/issues/22125#issuecomment-1208810581) when I started working on this (with minor changes to make sense out of context):

---

RFC 9110, "HTTP Semantics," has a section on HTTP URI normalization, [which says](https://www.rfc-editor.org/rfc/rfc9110.html#name-https-normalization-and-com):

> Two HTTP URIs that are equivalent after normalization (using any method) can be assumed to identify the same resource, and any HTTP component MAY perform normalization. As a result, distinct resources SHOULD NOT be identified by HTTP URIs that are equivalent after normalization.

In other words, `/foo%2Fbar` and `/foo/bar` are equivalent after normalizing, and thus they _SHOULD NOT_ be used for distinct resources. So if you are encoding application data into the path, and that data could possibly have reserved characters / delimiters (like `/`), consider redesigning your API: it is not robust in the harsh HTTP environment.

Note that several RFCs, notably RFCs 3986 and 9110, continually repeat that URI parsing is dependent upon scheme. That's one other problem: we all use the `http://` or `https://` scheme and yet expect applications to handle URIs differently. So of course there's going to be head-butting: we're fighting the design.

To clarify, it is definitely possible for a URI path such as `/band/AC%2fDC/T.N.T` to be "properly" handled by a server application. For this case, simply write a server that decodes everything after `/band/` except `%2f`. :man_shrugging: The problem is that this is difficult _in general_. Depending on what situations you do this, you may be opening yourself to bugs and security holes. This is why Caddy currently handles URIs solely in the unescaped space: it's the "one true" representation of a URI, and normalized HTTP URIs are more or less clearly defined nowadays.

Others might propose a solution to double-encode application data in the path; in other words, have the client send a URI with a path of `/bands/AC%252fDC/T.N.T`. This will probably work, but it's a hack and it [violates spec](https://www.rfc-editor.org/rfc/rfc3986#section-2.4):

> Implementations **must not** percent-encode or decode the same string more than once, as decoding an already decoded string might lead to misinterpreting a percent data octet as the beginning of a percent-encoding, or vice versa in the case of percent-encoding an already percent-encoded string.

Beware of the non-conforming behavior and highlight it very prominently in documentation so you can avoid bugs.

Laravel user @alcaitiff made a comment that some of you may be thinking:

> The router should resolve the route and after that decode parameters, but it does decode the url parameters before resolving the route.

I can't speak for Laravel or what it's doing, but the Go standard library (what Caddy uses), for example, **does** do URL parsing correctly and still has this problem. Go does exactly what you and the spec recommend: it splits the URI into its components and _then_ decodes reserved characters after parsing. It preserves the original, "raw" path in the `RawPath` field and offers the decoded path in the `Path` field. Its [`EscapedPath()`](https://pkg.go.dev/net/url#URL.EscapedPath) method uses `RawPath` _if it is a valid encoding of `Path`_, which is interesting because any given path has multiple valid encodings as I noted above. So if I want to truly "normalize" the URI in Go, I have to call `url.PathEscape(req.URL.Path)` myself and ignore `RawPath` entirely (AFAIK). And guess what, this converts `/foo/bar` to... `/foo/bar`. In other words, decoding `/foo%2Fbar` is not reversible without loss of precision. (Unless the HTTP server knows your business logic, more on that in a moment.)

We can write our own logic, though, that uses `RawPath` as a "hint" (as the Go docs say) to maybe replace `/` with `%2f`, but if we've manipulated/rewritten the URI at all, this becomes infeasible because we don't know where or if that instance still exists in the string.

[RFC 3986 section 2](https://www.rfc-editor.org/rfc/rfc3986#section-2.2) states:

>    URI producing applications should percent-encode data octets that
>    correspond to characters in the reserved set unless these characters
>    are specifically allowed by the URI scheme to represent data in that
>    component. If a reserved character is found in a URI component and
>    no delimiting role is known for that character, then it must be
>    interpreted as representing the data octet corresponding to that
>    character's encoding in US-ASCII.

The `/` is in the reserved set. Thus it is up to the implementation to determine whether it is data or a delimiter. I guess Laravel doesn't know, and it's frankly safer to assume it's a delimiter and treat it in its normalized form.

So yes, this issue is frustrating. As a web server author, I feel like I need to write software that can read people's minds: is this slash data or is it a delimiter? The router needs more information, because both are very valid ways of interpreting a URI!

---

## The solution

I think the key to this problem is trying to read the developer's mind: is this character supposed to be a delimiter (part of the path) or data? Should we collapse repeated slashes or no?

The answers depend on the context. For routing / path matching, the answer may be one way, for rewriting it may be another, and for proxying it may be yet another depending on the applications being proxied to.

Nginx, Apache, and Caddy all merge slashes by default when matching. However, Nginx and Apache have options to disable that behavior and preserve the slashes, which can lead to security vulnerabilities. All three do path matching (or routing) in the normalized space to mitigate bugs but, like we saw with Laravel, makes it difficult or impossible to route requests with application data that decode as path-significant characters like `%2F` (`/`), leaving many developers frustrated.

This PR introduces a somewhat novel solution that allows the developer to convey their intent to the server when doing matching and rewriting.

**Simply put, our solution is to interpret encoded characters and multiple slashes in the configuration as a literal conveyance of the developer's intent.** In other words, we don't blanket-unescape the whole URI every time. We do it byte-for-byte in lock-step with the configured pattern to match, and only unescape if the match pattern is not escaped at that position. Similarly, if a configured path has double slashes `//` in it, we do not merge slashes when comparing paths, because we infer the user's intent is to match repeated slashes.

### Path matching

Path matching (aka routing) is still done in the normalized space. That means if you configure a path matcher of `/foo/bar`, it will match `/foo/bar`, `/foo%2Fbar` and even `%2F%66%6F%6F%2F%62%61%72` because we normalize the URI. This is unchanged from before.

But now if you have a path matcher of `/foo%2Fbar`, it will match `/foo%2Fbar` exactly (the escape sequence is case-insensitive), whereas previously it would have only matched `/foo%252Fbar` (i.e. `%` as data). Now, `/foo%2Fbar` will NOT match `/foo/bar` or `%2Ffoo/bar` because we infer intent from seeing escape sequences in the match pattern as application data, not path delimiters.

**This logic handily extends to wildcards, too.** Referring to the previous example from our Laravel discussion, if you want to use `/bands/*/*` it is impossible to match a URI of `/bands/AC%2fDC/T.N.T` (in Laravel, too). But with this change, you can use special "escape-wildcard" characters: `/bands/%*/%*` to indicate that the span matched by the wildcard should _not_ be URI-decoded and should be kept in the escaped/raw space.

So now, if you want to allow band names to have a `/` in them, you can simply write `/bands/%*/%*`.

### Double slashes

Similar to escape sequences, we now disable slash merging automatically if the configured pattern has repeated slashes. Previously, it was impossible to match `//foo` because all URIs were normalized. Now, a path matcher of `//foo` will preserve multiple slashes. (A matcher of `/foo` will still match `//foo`.)

### Rewriting

A common task of rewriting is to strip path prefix and path suffix. The logic explained above has also been implemented for these operations, allowing you to use escaped characters and multiple slashes in your prefix and suffix patterns, and now Caddy will rewrite more intuitively and correctly.

For example, if you want to strip a prefix of `//prefix` from `//prefix/foo`, it will work, whereas before it wouldn't find the prefix because it would look at a fully-normalized URI.

Similarly, you can strip prefixes or suffixes with encoded characters. For example, a prefix of `/foo%2Fbar` will rewrite a URI of `/foo%2Fbar/asdf` into `/asdf`, whereas before it wouldn't find the prefix.

## Is it perfect?

Probably not. Are there bugs? Probably. Have I overlooked things? Almost certainly yes. I'm pretty sure there might be nooks and crannies within Caddy that I missed implementing this. Please file a bug report if you need it to work but doesn't work like you expect.

I'm pretty happy with this approach though. I think it's very useful and I don't know of other mainstream servers or frameworks that implement this behavior. In true Caddy fashion, this should _just work_.

- fixes #4801
- fixes #4923
- fixes #4743

### DIFF
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -245,10 +245,32 @@ func SanitizedPathJoin(root, reqPath string) string {
 	return path
 }
 
-// CleanPath cleans path p according to path.Clean() except that
-// it preserves trailing slashes, since those are significant
-// in our application.
-func CleanPath(p string) string {
+// CleanPath cleans path p according to path.Clean(), but only
+// merges repeated slashes if collapseSlashes is true, and always
+// preserves trailing slashes.
+func CleanPath(p string, collapseSlashes bool) string {
+	if collapseSlashes {
+		return cleanPath(p)
+	}
+
+	// replace repeated slashes with an invalid/impossible URI character
+	// to avoid collisions; then clean the path, then replace them back
+	const tmpSlash = 0xff
+	var sb strings.Builder
+	for i, ch := range p {
+		if ch == '/' && i > 0 && (p[i-1] == '/' || p[i-1] == tmpSlash) {
+			sb.WriteByte(tmpSlash)
+			continue
+		}
+		sb.WriteRune(ch)
+	}
+	halfCleaned := cleanPath(sb.String())
+	halfCleaned = strings.ReplaceAll(halfCleaned, string([]byte{tmpSlash}), "/")
+
+	return halfCleaned
+}
+
+func cleanPath(p string) string {
 	cleaned := path.Clean(p)
 	if cleaned != "/" && strings.HasSuffix(p, "/") {
 		cleaned = cleaned + "/"
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -242,6 +243,17 @@ func SanitizedPathJoin(root, reqPath string) string {
 	}
 
 	return path
+}
+
+// CleanPath cleans path p according to path.Clean() except that
+// it preserves trailing slashes, since those are significant
+// in our application.
+func CleanPath(p string) string {
+	cleaned := path.Clean(p)
+	if cleaned != "/" && strings.HasSuffix(p, "/") {
+		cleaned = cleaned + "/"
+	}
+	return cleaned
 }
 
 // tlsPlaceholderWrapper is a no-op listener wrapper that marks
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -255,17 +255,17 @@ func CleanPath(p string, collapseSlashes bool) string {
 
 	// replace repeated slashes with an invalid/impossible URI character
 	// to avoid collisions; then clean the path, then replace them back
-	const tmpSlash = 0xff
+	var tmpSlash = string("\xff/")
 	var sb strings.Builder
 	for i, ch := range p {
 		if ch == '/' && i > 0 && p[i-1] == '/' {
-			sb.WriteByte(tmpSlash)
+			sb.WriteString(tmpSlash)
 			continue
 		}
 		sb.WriteRune(ch)
 	}
 	halfCleaned := cleanPath(sb.String())
-	halfCleaned = strings.ReplaceAll(halfCleaned, string([]byte{tmpSlash}), "/")
+	halfCleaned = strings.ReplaceAll(halfCleaned, "/\xff", "/")
 
 	return halfCleaned
 }
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -270,6 +270,7 @@ func CleanPath(p string, collapseSlashes bool) string {
 	return halfCleaned
 }
 
+// cleanPath does path.Clean(p) but preserves any trailing slash.
 func cleanPath(p string) string {
 	cleaned := path.Clean(p)
 	if cleaned != "/" && strings.HasSuffix(p, "/") {
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -253,19 +253,19 @@ func CleanPath(p string, collapseSlashes bool) string {
 		return cleanPath(p)
 	}
 
-	// replace repeated slashes with an invalid/impossible URI character
-	// to avoid collisions; then clean the path, then replace them back
-	var tmpSlash = string("\xff/")
+	// insert an invalid/impossible URI character into each two consecutive
+	// slashes to expand empty path segments; then clean the path as usual,
+	// and then remove the remaining temporary characters.
+	const tmpCh = 0xff
 	var sb strings.Builder
 	for i, ch := range p {
 		if ch == '/' && i > 0 && p[i-1] == '/' {
-			sb.WriteString(tmpSlash)
-			continue
+			sb.WriteByte(tmpCh)
 		}
 		sb.WriteRune(ch)
 	}
 	halfCleaned := cleanPath(sb.String())
-	halfCleaned = strings.ReplaceAll(halfCleaned, "/\xff", "/")
+	halfCleaned = strings.ReplaceAll(halfCleaned, string([]byte{tmpCh}), "")
 
 	return halfCleaned
 }
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -258,7 +258,7 @@ func CleanPath(p string, collapseSlashes bool) string {
 	const tmpSlash = 0xff
 	var sb strings.Builder
 	for i, ch := range p {
-		if ch == '/' && i > 0 && (p[i-1] == '/' || p[i-1] == tmpSlash) {
+		if ch == '/' && i > 0 && p[i-1] == '/' {
 			sb.WriteByte(tmpSlash)
 			continue
 		}
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -95,7 +95,7 @@ type (
 	//
 	// Even though path matching is done in normalized space, the special
 	// wildcard `%*` may be used in place of a span that should NOT be decoded;
-	// that is, `/bands/%*` will match `/bands/AC%2fDC` whereas `/bands/*`
+	// that is, `/bands/%*/` will match `/bands/AC%2fDC/` whereas `/bands/*/`
 	// will not.
 	//
 	// This matcher is fast, so it does not support regular expressions or
@@ -383,48 +383,55 @@ func (m MatchPath) Provision(_ caddy.Context) error {
 func (m MatchPath) Match(r *http.Request) bool {
 	// Even though RFC 9110 says that path matching is case-sensitive
 	// (https://www.rfc-editor.org/rfc/rfc9110.html#section-4.2.3),
-	// we do case in-sensitive matching to mitigate security issues
+	// we do case-insensitive matching to mitigate security issues
 	// related to differences between operating systems, applications,
 	// etc; if case-sensitive matching is needed, the regex matcher
 	// can be used instead.
-	lowerPath := strings.ToLower(r.URL.Path)
-
-	// Clean the path, merges doubled slashes, etc.
-	// This ensures maliciously crafted requests can't bypass
-	// the path matcher. See #4407. Good security posture
-	// suggests that we should do all we can to reduce any
-	// funny-looking paths into "normalized" forms such that
-	// weird variants can't sneak by.
-	lowerPath = path.Clean(lowerPath)
+	reqPath := strings.ToLower(r.URL.Path)
 
 	// See #2917; Windows ignores trailing dots and spaces
 	// when accessing files (sigh), potentially causing a
 	// security risk (cry) if PHP files end up being served
 	// as static files, exposing the source code, instead of
 	// being matched by *.php to be treated as PHP scripts.
-	lowerPath = strings.TrimRight(lowerPath, ". ")
-
-	// Cleaning may remove the trailing slash, but we want to keep it.
-	if lowerPath != "/" && strings.HasSuffix(r.URL.Path, "/") {
-		lowerPath = lowerPath + "/"
-	}
+	reqPath = strings.TrimRight(reqPath, ". ")
 
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 
-	for _, matchPath := range m {
-		matchPath = repl.ReplaceAll(matchPath, "")
+	for _, matchPattern := range m {
+		matchPattern = repl.ReplaceAll(matchPattern, "")
 
 		// special case: whole path is wildcard; this is unnecessary
 		// as it matches all requests, which is the same as no matcher
-		if matchPath == "*" {
+		if matchPattern == "*" {
 			return true
 		}
 
+		// Clean the path, merge doubled slashes, etc.
+		// This ensures maliciously crafted requests can't bypass
+		// the path matcher. See #4407. Good security posture
+		// requires that we should do all we can to reduce any
+		// funny-looking paths into "normalized" forms such that
+		// weird variants can't sneak by.
+		//
+		// How we clean the path depends on the kind of pattern:
+		// we either merge slashes or we don't. If the pattern
+		// has double slashes, we preserve them in the path.
+		//
+		// TODO: Despite the fact that the *vast* majority of path
+		// matchers have only 1 pattern, a possible optimization is
+		// to remember the cleaned form of the path for future
+		// iterations; it's just that the way we clean depends on
+		// the kind of pattern.
+
+		patternHasDoubleSlashes := strings.Contains(matchPattern, "//")
+
 		// if '%' appears in the match pattern, we interpret that to mean
 		// the intent is to compare that part of the path in raw/escaped
-		// space; i.e. "%40" == "%40" and not "@"
-		if strings.Contains(matchPath, "%") {
-			if m.matchPatternWithEscapeSequence(r, matchPath) {
+		// space; i.e. "%40"=="%40", not "@", and "%2F"=="%2F", not "/"
+		if strings.Contains(matchPattern, "%") {
+			reqPathForPattern := m.cleanPath(r.URL.EscapedPath(), !patternHasDoubleSlashes)
+			if m.matchPatternWithEscapeSequence(reqPathForPattern, matchPattern) {
 				return true
 			}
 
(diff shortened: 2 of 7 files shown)


modules/caddyhttp/matchers.go
```diff
@@ -63,20 +62,51 @@ type (
...
+	// compared with the request's raw/escaped path for those bytes only.
+	// In other words, a matcher of `/foo%2Fbar` will match a request path
+	// of precisely `/foo%2Fbar`, but not `/foo/bar`. It follows that matching
+	// the literal percent sign (%) in normalized space can be done using the
+	// escaped form, `%25`.
+	//
+	// Even though wildcards (`*`) operate in the normalized space, the special
+	// escaped wildcard (`%*`), which is not a valid escape sequence, may be
+	// used in place of a span that should NOT be decoded; that is, `/bands/%*`
+	// will match `/bands/AC%2fDC` whereas `/bands/*` will not.
```
RussellLuo, line 94: Intuitively, I expect that `/bands/*` will also match `/bands/AC%2fDC`, which is `/bands/AC/DC` in unescaped form 🤔

mholt: Oh, true. Let me put a / at the end of those and it will make more sense.

RussellLuo: Yeah, it makes sense now 😄

modules/caddyhttp/caddyhttp.go
```diff
@@ -245,10 +245,32 @@ func SanitizedPathJoin(root, reqPath string) string {
...
+	if collapseSlashes {
+		return cleanPath(p)
+	}
+
+	// replace repeated slashes with an invalid/impossible URI character
+	// to avoid collisions; then clean the path, then replace them back
+	const tmpSlash = 0xff
+	var sb strings.Builder
+	for i, ch := range p {
+		if ch == '/' && i > 0 && (p[i-1] == '/' || p[i-1] == tmpSlash) {
```
RussellLuo, line 261: I guess `p` should not contain any `tmpSlash`?

modules/caddyhttp/matchers.go
```diff
@@ -477,18 +647,10 @@ func (MatchPathRE) CaddyModule() caddy.ModuleInfo {
...
-	unescapedPath, err := url.PathUnescape(r.URL.Path)
-	if err != nil {
-		return false
-	}
-
 	// Clean the path, merges doubled slashes, etc.
 	// This ensures maliciously crafted requests can't bypass
 	// the path matcher. See #4407
-	cleanedPath := path.Clean(unescapedPath)
+	cleanedPath := path.Clean(r.URL.Path)
```
RussellLuo, line 653: I think we can merge L653~L658 into one line:

```go
cleanedPath := cleanPath(r.URL.Path)
```

modules/caddyhttp/caddyhttp.go
mholt: Ah yeah, it's not intuitive. It does contain `tmpSlash` because we put it in there:

```
1. /foo///bar
2. /foo/0xFF/bar
3. /foo/0xFF0xFFbar
```

On that third iteration, the 1-lookbehind sees 0xFF because we barely replaced the `/` with `0xFF`.

modules/caddyhttp/matchers.go
mholt: Good point, thanks!

I do realize the path_regexp matcher works a little differently than (or rather, lacks some of the new flexibility of) the path matcher (as the path_regexp matcher always operates in the normalized space only), but that's probably OK...

modules/caddyhttp/caddyhttp.go
RussellLuo: I think `p` is immutable, while `sb` grows:

| iteration | `p` | `sb` |
| --- | --- | --- |
| 1 | /foo///bar | /foo/ |
| 2 | /foo///bar | /foo/0xFF |
| 3 | /foo///bar | /foo/0xFF0xFF |

mholt: Oh... that's right. (It was late when I wrote that and it's late now.) Why does this work then?

coolaj86:
Very well reasoned, but I think this is something that you'll never regret not merging, and that you may very well regret putting into the codebase for a use case that is much more art (abusing constraints) than software engineering.

mholt:
@coolaj86 Hiya! (I know I just replied on Twitter, but)

I disagree with the part about abusing constraints, but I do think you could be right. This is not my favorite change, but it doesn't violate any standards (and in fact it is perhaps more conforming in some ways than other implementations) and it provides an elegant solution to otherwise hacky or non-compliant alternatives.

Unless I have overlooked some significant problem with this approach, I do think it's well-engineered.

But yes, I do hope I don't regret this. 😅

coolaj86:
YOLO!

modules/caddyhttp/caddyhttp.go
```diff
@@ -244,6 +245,40 @@ func SanitizedPathJoin(root, reqPath string) string {
 	return path
 }
 
+// CleanPath cleans path p according to path.Clean(), but only
+// merges repeated slashes if collapseSlashes is true, and always
+// preserves trailing slashes.
+func CleanPath(p string, collapseSlashes bool) string {
```
RussellLuo, line 251: I gave this another thought.

Typically I guess [dot segments](https://www.rfc-editor.org/rfc/rfc3986#section-5.2.4) ("." and "..") are rarely used in URLs, but I'm afraid the behavior of `CleanPath` is hard to reason about if  dot segments are used along with multiple slashes.

For example, I'm not sure if `/foo//./bar` is the result most users expect 🤔

| p | CleanPath | path.Clean |
| --- | --- | --- |
| /foo//./..//./bar | /foo//./bar | /bar |
| /foo/./.././bar | /bar | /bar |

mholt: Interesting! I'll look into this...

mholt: I guess we need to decide if `//..` is `/` or `//`.

RussellLuo: Oh, this is an important point!

Obviously, [path.Clean](https://pkg.go.dev/path#Clean) chooses `/`, but personally I think `/foo` is a better choice because [URI Empty Path Segments Matter](http://davescoolblog.blogspot.com/2011/11/uri-empty-path-segments-matter.html).

RussellLuo: The same author made an interesting point about path matching in [Application Content URI Rules wildcard syntax](http://davescoolblog.blogspot.com/2017/05/application-content-uri-rules-wildcard.html):

> For the path, an asterisk matches within the path segment. For example, http://example.com/a/*/c will match http://example.com/a/b/c and http://example.com/a//c but not http://example.com/a/b/b/c or http://example.com/a/c

mholt: @RussellLuo Thanks for those resources. I think I tend to agree, if we're not merging slashes, then treat empty path components as significant. I pushed a commit just now which makes the new test case pass.

modules/caddyhttp/caddyhttp.go
```diff
@@ -255,17 +255,17 @@ func CleanPath(p string, collapseSlashes bool) string {
...
 		if ch == '/' && i > 0 && p[i-1] == '/' {
-			sb.WriteByte(tmpSlash)
+			sb.WriteString(tmpSlash)
 			continue
 		}
 		sb.WriteRune(ch)
 	}
 	halfCleaned := cleanPath(sb.String())
-	halfCleaned = strings.ReplaceAll(halfCleaned, string([]byte{tmpSlash}), "/")
+	halfCleaned = strings.ReplaceAll(halfCleaned, "/\xff", "/")
```
RussellLuo, line 268: This fix looks good to me!

Another solution I can think of - with minimum changes - is to insert `0xff` into two consecutive slashes. For example:

```diff
	const tmpSlash = 0xff
	var sb strings.Builder
	for i, ch := range p {
		if ch == '/' && i > 0 && p[i-1] == '/' {
			sb.WriteByte(tmpSlash)
-			continue
		}
		sb.WriteRune(ch)
	}
	halfCleaned := cleanPath(sb.String())
-	halfCleaned = strings.ReplaceAll(halfCleaned, string([]byte{tmpSlash}), "/")
+	halfCleaned = strings.ReplaceAll(halfCleaned, string([]byte{tmpSlash}), "")
```

That being said, the two solutions are essentially equivalent, so just pick the one you prefer :)

mholt: @RussellLuo That's a way better solution IMO! Let's do that. Do you want to push a commit to this branch or something? That way you can get credit for it. (I just gave you write permissions. Welcome to the team!) If you don't care then I don't mind doing it. Either way.

RussellLuo: Oh, thanks! It's my pleasure 😄

RussellLuo: I have pushed a commit (also improved the docs). Hope it's ok :)

mholt: Yes! Looks good. Thank you very much!

PR #4950, opened 2022-08-11
httpcaddyfile: Fix bind when IPv6 is specified with network

stapelberg:
Commit 1e18afb5c862d62be130d563785de5c58f08ae8e broke my caddy setup.
This commit fixes it.

### DIFF
--- a/caddyconfig/httpcaddyfile/addresses.go
+++ b/caddyconfig/httpcaddyfile/addresses.go
@@ -219,7 +219,7 @@ func (st *ServerType) listenerAddrsForServerBlockKey(sblock serverBlock, key str
 		return nil, fmt.Errorf("[%s] scheme and port violate convention", key)
 	}
 
-	// the bind directive specifies hosts, but is optional
+	// the bind directive specifies hosts (and potentially network), but is optional
 	lnHosts := make([]string, 0, len(sblock.pile["bind"]))
 	for _, cfgVal := range sblock.pile["bind"] {
 		lnHosts = append(lnHosts, cfgVal.Value.([]string)...)
@@ -234,22 +234,22 @@ func (st *ServerType) listenerAddrsForServerBlockKey(sblock serverBlock, key str
 
 	// use a map to prevent duplication
 	listeners := make(map[string]struct{})
-	for _, host := range lnHosts {
-		var a string
-		if host != "" {
-			// host can have network + host (e.g. "tcp6/localhost") but
-			// will/should not have port information because this usually
-			// comes from the bind directive, so we append the port
-			network, rest, err := caddy.SplitNetwork(host)
-			if err != nil {
-				return nil, fmt.Errorf("splitting network address %q: %v", host, err)
-			}
-			rest = strings.Trim(rest, "[]")
-			a = caddy.JoinNetworkAddress(network, rest, lnPort)
-		} else {
-			a = ":" + lnPort
+	for _, lnHost := range lnHosts {
+		// normally we would simply append the port,
+		// but if lnHost is IPv6, we need to ensure it
+		// is enclosed in [ ]; net.JoinHostPort does
+		// this for us, but lnHost might also have a
+		// network type in front (e.g. "tcp/") leading
+		// to "[tcp/::1]" which causes parsing failures
+		// later; what we need is "tcp/[::1]", so we have
+		// to split the network and host, then re-combine
+		network, host, ok := strings.Cut(lnHost, "/")
+		if !ok {
+			host = network
 		}
-		addr, err := caddy.ParseNetworkAddress(a)
+		host = strings.Trim(host, "[]") // IPv6
+		networkAddr := caddy.JoinNetworkAddress(network, host, lnPort)
+		addr, err := caddy.ParseNetworkAddress(networkAddr)
 		if err != nil {
 			return nil, fmt.Errorf("parsing network address: %v", err)
 		}
--- a/caddyconfig/httpcaddyfile/addresses.go
+++ b/caddyconfig/httpcaddyfile/addresses.go
@@ -246,6 +246,7 @@ func (st *ServerType) listenerAddrsForServerBlockKey(sblock serverBlock, key str
 		network, host, ok := strings.Cut(lnHost, "/")
 		if !ok {
 			host = network
+			network = ""
 		}
 		host = strings.Trim(host, "[]") // IPv6
 		networkAddr := caddy.JoinNetworkAddress(network, host, lnPort)
--- a/caddyconfig/httpcaddyfile/addresses.go
+++ b/caddyconfig/httpcaddyfile/addresses.go
@@ -36,12 +36,12 @@ import (
 // server block that share the same address stay grouped together so the config
 // isn't repeated unnecessarily. For example, this Caddyfile:
 //
-// 	example.com {
-// 		bind 127.0.0.1
-// 	}
-// 	www.example.com, example.net/path, localhost:9999 {
-// 		bind 127.0.0.1 1.2.3.4
-// 	}
+//	example.com {
+//		bind 127.0.0.1
+//	}
+//	www.example.com, example.net/path, localhost:9999 {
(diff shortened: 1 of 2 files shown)


mholt:
Thanks; unfortunately the tests fail:

```
    caddytest.go:378: adapting config using caddyfile adapter: server block 0, key 0 (example.com): determining listener address: parsing network address: address [::]]:443: missing port in address
    caddyfile_adapt_test.go:52: failed to adapt bind_ipv6.txt
    caddytest.go:378: adapting config using caddyfile adapter: server block 0, key 0 (example.com): determining listener address: parsing network address: address [::]]:443: missing port in address
    caddyfile_adapt_test.go:52: failed to adapt global_options_default_bind.txt
```

stapelberg:
This was because of the network/ prefix syntax. Pushed a fix. Tests pass locally for me now.

stapelberg:
Still getting an error with my other caddy config :-/

```
determining listener address: splitting network address "2a02:168:4a00::d1d1": address 2a02:168:4a00::d1d1: too many colons in address
```

stapelberg:
PR updated, though we should probably add more tests — this issue I only found when updating caddy, suggesting a lack of test coverage.

mholt:
Yeah, we could use more tests in general.

Thanks! I'll review this again soon.

caddyconfig/httpcaddyfile/addresses.go
```diff
@@ -234,10 +234,23 @@ func (st *ServerType) listenerAddrsForServerBlockKey(sblock serverBlock, key str
...
+			// host can have network + host (e.g. "tcp6/localhost") but
+			// will/should not have port information because this usually
+			// comes from the bind directive, so we append the port
+			network, rest, err := caddy.SplitNetwork(host)
+			if err != nil {
+				return nil, fmt.Errorf("splitting network address %q: %v", host, err)
+			}
+			if strings.HasPrefix(rest, "[") && strings.HasSuffix(rest, "]") {
+				rest = strings.TrimSuffix(strings.TrimPrefix(rest, "["), "]")
+			}
```
mholt, line 248:
```suggestion
			rest = strings.Trim(rest, "[]")
```

listeners.go
```diff
@@ -495,6 +495,21 @@ func ParseNetworkAddress(addr string) (NetworkAddress, error) {
 	}, nil
 }
 
+// SplitNetwork splits a into a network, if any, and rest.
+func SplitNetwork(a string) (network, rest string, err error) {
+	beforeSlash, afterSlash, slashFound := strings.Cut(a, "/")
+	if slashFound {
+		network = strings.ToLower(strings.TrimSpace(beforeSlash))
+		a = afterSlash
+	}
```
mholt, line 504: The ToLower and TrimSpace probably aren't necessary (or shouldn't be, anyway)

listeners.go
```diff
@@ -495,6 +495,21 @@ func ParseNetworkAddress(addr string) (NetworkAddress, error) {
...
+	if isUnixNetwork(network) {
+		rest = a
+		return
+	}
+	rest = a
+	return
```
mholt, line 510: :thinking: 

Couldn't this whole function basically become:

```go
return strings.Cut(a, "/)
```

Thus we can eliminate the function altogether.

mholt (changes requested):
Thanks for the revisions!

Upon a closer look I think this patch can be vastly simplified. I don't think we need the `SplitNetwork` function at all! :)

listeners.go
stapelberg: I just copied this from SplitNetworkAddress below. If you think these shouldn’t be necessary, can I ask you to clean them up in a separate change? I’d like to keep the scope of my change to a bare minimum.

listeners.go
stapelberg: Almost. The third return value of strings.Cut is a bool, not an error.

But regardless of this detail, I’d argue that even if all the function does is call `strings.Cut`, having it have a descriptive name (`SplitNetwork`) is preferable to eliminating it, for clarity.

stapelberg (commented):
PTAL

listeners.go
mholt: Sure. And I can probably remove them from below too, just never really thought about it.

listeners.go
mholt: We never assign the error, do we? I think using `strings.Cut` gives us more information then, since at least the bool tells us something. Plus there's no guesswork of "how is this different from doing strings.Cut" -- because it's not :smiley: (which is a good thing!)

mholt:
@stapelberg Almost done with the tweaks on my end. But before I commit, can you share your Caddyfile that produced the error?

For me, replacing `bind tcp/::1` with `bind tcp/[::1]` made it work, even without needing a patch. I'm not sure if IPv6 is always supposed to have brackets around it, or only if there's a port. What do you think?

mholt:
@stapelberg I guess we should support specifying the IPv6 without `[ ]` since the user won't be putting a port there.

So, I just pushed my revisions. Let me know what you think / if it works for you!

stapelberg:
Your changes break listening again:

```
2022/09/14 08:38:55 gokrazy: attempt 35, starting ["/tmp/gok-run/caddy" "run" "--config" "/etc/caddy/Caddyfile"]
{"level":"info","ts":1663137535.9100008,"msg":"using provided configuration","config_file":"/etc/caddy/Caddyfile","config_adapter":""}
{"level":"info","ts":1663137535.914408,"logger":"admin","msg":"admin endpoint started","address":"localhost:2019","enforce_origin":false,"origins":["//127.0.0.1:2019","//localhost:2019","//[::1]:2019"]}
{"level":"warn","ts":1663137535.9146025,"logger":"http","msg":"server is listening only on the HTTP port, so no automatic HTTPS will be applied to this server","server_name":"srv0","http_port":80}
{"level":"info","ts":1663137535.9146535,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc000882d90"}
{"level":"info","ts":1663137535.9157517,"logger":"tls.cache.maintenance","msg":"stopped background certificate maintenance","cache":"0xc000882d90"}
Error: loading initial config: loading new config: http app module: start: 2a02:168:4a00::d1d1: listening on [2a02:168:4a00::d1d1]:80: listen 2a02:168:4a00::d1d1: unknown network 2a02:168:4a00::d1d1
2022/09/14 08:38:55 gokrazy: exit status 1
```

This is my Caddyfile:
```
{
	order realip first
}

(stdlog) {
	log {
		output file /perm/logs/access.log
		output stderr
		format transform "{common_log}"
	}
	realip {
		header "CF-Connecting-IP"
		from cloudflare
		maxhops -1
	}
}

http://repo.distr1.org:80 {
	bind 2a02:168:4a00::d1d1
	import stdlog
	root * /perm/srv/repo.distr1.org/
	encode zstd gzip
	file_server browse {
		precompressed zstd gzip
	}
}

http://repo7.distr1.org:80 {
	bind 2a02:168:4a00::d1d1
	import stdlog
	root * /perm/srv/repo.distr1.org/
	encode zstd gzip
	file_server browse {
		precompressed zstd gzip
	}
}

http://browse.distr1.org:80 {
	bind 2a02:168:4a00::d1d1
	import stdlog
	@staticassets {
		path /css/*
		path /js/*
	}
	handle @staticassets {
		header Cache-Control "max-age=3600"
		file_server browse
		root * /perm/srv/browse.distr1.org/assets
	}
	handle /* {
		reverse_proxy /* http://localhost:8047
	}
}

```

mholt:
D'oh! That's what I get for coding when I'm too tired.

I _copied_ network into host, instead of _moving_ network into host.

All the cases work for me now.

stapelberg:
> D'oh! That's what I get for coding when I'm too tired.
> 
> I _copied_ network into host, instead of _moving_ network into host.
> 
> All the cases work for me now.

Can confirm, the current PR contents work for me, too 👍

Thanks!

mholt (approved):
Thanks for confirming!

PR #4952, opened 2022-08-12
reverseproxy: Ignore context cancel in stream mode

### LINKED ISSUE #4922: caddyserver-reverseproxy + ffmpeg streaming causes invalid files/connection state
Hello,

first of all, great project, love the caddy server - love what you guys are doing, keep up the good work! :+1: 

Now for my problem:

I am using the caddy server as a reverse proxy for a python application that receives ffmpeg's streaming data and processes it (which is out of scope for this issue). Whenever the use-case has a client exiting the connection to caddy validly, we do not have any issues at all. 

Now, we are trying some low-latency streaming, which involves that ffmpeg runs with the parameter `-http_persistent 0`. Now this causes problems with caddy. Using ffmpeg's http persistent 0 causes the connection to simply drop out without a clean "exit".

It looks like, when we are using caddy as inbetween proxy, whenever the connection drops out, there's a 50:50 chance that it causes an issue on the python side (OSErrors).

If we remove caddy inbetween it works fine (streaming direct to the python server).
This is not only related to UWSGI/python, we have the same issue if you just use plain old nginx with WebDAV to PUT the files on via `ffmpeg`.

I have created a PoC project, containing a minimum viable client for this issue: https://github.com/kuhnchris/caddyserver-ffmpeg-ll-hls-poc

In the default configuration (`docker compose up -d && cd ffmpeg && sh ll_hls.sh`) we see that caddy sometimes works without any hickups, and then, randomly throws this on python (and nginx) side: 
```
poc-caddy-1  | {"level":"debug","ts":1659176760.2804174,"logger":"http.handlers.reverse_proxy","msg":"upstream roundtrip","upstream":"uwsgi:8002","duration":0.001713273,"request":{"remote_ip":"172.18.0.1","remote_port":"51190","proto":"HTTP/1.1","method":"PUT","host":"localhost:8888","uri":"/example.m3u8","headers":{"User-Agent":["Lavf/58.76.100"],"X-Forwarded-Proto":["http"],"X-Forwarded-Host":["localhost:8888"],"Icy-Metadata":["1"],"Accept":["*/*"],"X-Forwarded-For":["172.18.0.1"],"X-Server-Name":["test"]}},"error":"context canceled"}
poc-caddy-1  | {"level":"debug","ts":1659176760.2806652,"logger":"http.handlers.reverse_proxy","msg":"upstream roundtrip","upstream":"uwsgi:8002","duration":0.00282887,"request":{"remote_ip":"172.18.0.1","remote_port":"51188","proto":"HTTP/1.1","method":"PUT","host":"localhost:8888","uri":"/example190.m4s","headers":{"X-Server-Name":["test"],"Accept":["*/*"],"X-Forwarded-Proto":["http"],"X-Forwarded-Host":["localhost:8888"],"Icy-Metadata":["1"],"X-Forwarded-For":["172.18.0.1"],"User-Agent":["Lavf/58.76.100"]}},"error":"context canceled"}
```
i.e. "context canceled". But only sometimes it feels like the reverse_proxy target is not receiving everything, so it feels like this is a race condition.

```
poc-uwsgi-1  | [pid: 9|app: 0|req: 61/379] 172.18.0.3 () {38 vars in 498 bytes} [Sat Jul 30 10:25:56 2022] PUT /example188.m4s => generated 0 bytes in 1 msecs (HTTP/1.1 500) 0 headers in 0 bytes (0 switches on core 4)
poc-uwsgi-1  | Traceback (most recent call last):
poc-uwsgi-1  |   File "/root/./test_handler.py", line 12, in dispatch
poc-uwsgi-1  |     while _ := uwsgi.chunked_read(30):
poc-uwsgi-1  | OSError: unable to receive chunked part
poc-uwsgi-1  | [pid: 8|app: 0|req: 320/380] 172.18.0.3 () {38 vars in 494 bytes} [Sat Jul 30 10:25:58 2022] PUT /example.m3u8 => generated 0 bytes in 0 msecs (HTTP/1.1 500) 0 headers in 0 bytes (0 switches on core 4)
poc-uwsgi-1  | [pid: 8|app: 0|req: 320/381] 172.18.0.3 () {38 vars in 498 bytes} [Sat Jul 30 10:25:58 2022] PUT /example189.m4s => generated 2 bytes in 1 msecs (HTTP/1.1 200) 2 headers in 64 bytes (1 switches on core 2)
poc-uwsgi-1  | Traceback (most recent call last):
poc-uwsgi-1  |   File "/root/./test_handler.py", line 12, in dispatch
poc-uwsgi-1  |     while _ := uwsgi.chunked_read(30):
poc-uwsgi-1  | OSError: unable to receive chunked part
```
on the nginx side it looks as follows:
```
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:44 +0000] "PUT /example.m3u8 HTTP/1.1" 204 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:44 +0000] "PUT /example392.m4s HTTP/1.1" 201 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:46 +0000] "PUT /example393.m4s HTTP/1.1" 201 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:46 +0000] "PUT /example.m3u8 HTTP/1.1" 204 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:48 +0000] "PUT /example.m3u8 HTTP/1.1" 400 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:48 +0000] "PUT /example394.m4s HTTP/1.1" 400 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:50 +0000] "PUT /example.m3u8 HTTP/1.1" 400 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:32:50 +0000] "PUT /example395.m4s HTTP/1.1" 201 0 "-" "Lavf/58.76.100"
...
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:33:30 +0000] "PUT /example.m3u8 HTTP/1.1" 204 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:33:32 +0000] "PUT /example.m3u8 HTTP/1.1" 204 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:33:32 +0000] "PUT /example416.m4s HTTP/1.1" 201 0 "-" "Lavf/58.76.100"
poc-nginx-1  | 172.18.0.3 - - [30/Jul/2022:10:33:34 +0000] "PUT /example.m3u8 HTTP/1.1" 400 0 "-" "Lavf/58.76.100"
```

the "400" correspond to the broken issue, the 204s and 201s are OK. This happens pretty randomly.
I tried looking around for the aformentioned "context closed" and found 2 spots in the reverse_proxy.go, but I didn't really see any way to work around this issue.

Does anyone have any idea how we could get this to work without caddy randomly causing exceptions/errors when having the (client) connection closed?

mholt:
If FlushInterval is explicitly configured to -1, we assume the user is intending for low-latency streaming, and apparently in those cases clients sometimes disconnect before the request has finished its round-trip with the backend.

This patch ignores the client disconnect in those cases so the request will still be completed without error.

Even though `flushInterval()` is dynamically computed, for now I'm requiring an explicit configuration of -1 because I don't fully know the implications of this yet. It might be fine, but if there's problems I want them to be opt-in.

We do know this solution at least works though.

Resolves #4922

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -724,6 +724,13 @@ func (h Handler) addForwardedHeaders(req *http.Request) error {
 	return nil
 }
 
+type myctx struct {
+	context.Context
+	done <-chan struct{}
+}
+
+func (c myctx) Done() <-chan struct{} { return c.done }
+
 // reverseProxy performs a round-trip to the given backend and processes the response with the client.
 // (This method is mostly the beginning of what was borrowed from the net/http/httputil package in the
 // Go standard library which was used as the foundation.)
@@ -755,6 +762,11 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
 			},
 		}
 		req = req.WithContext(httptrace.WithClientTrace(req.Context(), trace))
+	}
+
+	// TODO: very spikey and experimental (issue #4922)
+	if h.FlushInterval == -1 {
+		req = req.WithContext(myctx{req.Context(), h.ctx.Done()})
 	}
 
 	// do the round-trip; emit debug log with values we know are
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -109,6 +109,11 @@ type Handler struct {
 	// response is recognized as a streaming response, or if its
 	// content length is -1; for such responses, writes are flushed
 	// to the client immediately.
+	//
+	// Normally, a request will be canceled if the client disconnects
+	// before the response is received from the backend. If explicitly
+	// set to -1, client disconnection will be ignored and the request
+	// will be completed to help facilitate low-latency streaming.
 	FlushInterval caddy.Duration `json:"flush_interval,omitempty"`
 
 	// A list of IP ranges (supports CIDR notation) from which
@@ -724,13 +729,6 @@ func (h Handler) addForwardedHeaders(req *http.Request) error {
 	return nil
 }
 
-type myctx struct {
-	context.Context
-	done <-chan struct{}
-}
-
-func (c myctx) Done() <-chan struct{} { return c.done }
-
 // reverseProxy performs a round-trip to the given backend and processes the response with the client.
 // (This method is mostly the beginning of what was borrowed from the net/http/httputil package in the
 // Go standard library which was used as the foundation.)
@@ -764,9 +762,13 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
 		req = req.WithContext(httptrace.WithClientTrace(req.Context(), trace))
 	}
 
-	// TODO: very spikey and experimental (issue #4922)
+	// if FlushInterval is explicitly configured to -1 (i.e. flush continuously to achieve
+	// low-latency streaming), don't let the transport cancel the request if the client
+	// disconnects: user probably wants us to finish sending the data to the upstream
+	// regardless, and we should expect client disconnection in low-latency streaming
+	// scenarios (see issue #4922)
 	if h.FlushInterval == -1 {
-		req = req.WithContext(myctx{req.Context(), h.ctx.Done()})
+		req = req.WithContext(ignoreClientGoneContext{req.Context(), h.ctx.Done()})
 	}
 
 	// do the round-trip; emit debug log with values we know are
@@ -1385,6 +1387,19 @@ type handleResponseContext struct {
 	// happen twice.
 	isFinalized bool
 }
+
+// ignoreClientGoneContext is a special context.Context type
+// intended for use when doing a RoundTrip where you don't
+// want a client disconnection to cancel the request during
+// the roundtrip. Set its done field to a Done() channel
+// of a context that doesn't get canceled when the client
+// disconnects, such as caddy.Context.Done() instead.
+type ignoreClientGoneContext struct {
+	context.Context
+	done <-chan struct{}
+}
+
+func (c ignoreClientGoneContext) Done() <-chan struct{} { return c.done }
 
 // proxyHandleResponseContextCtxKey is the context key for the active proxy handler
 // so that handle_response routes can inherit some config options


francislavoie:
I guess we'll need to document that it must be _explicitly_ configured to get this behaviour, and that relying on the detection stuff (i.e. `Content-Type`, bidirectional, Content-Length guessing logic) won't get this behaviour.

mholt:
@francislavoie Yep. I do in the godoc comment in this patch. (I guess we'll want to update Caddyfile docs too.)

wazerstar:
Has this been implemented? I cannot see any thing about this in change-log on release page.

I'm testing some streaming and using hdhomerun, when client leaves,stops,change channel without using "flush_interval -1"

The outcome is this
{"resumed":false,"version":772,"cipher_suite":4865,"proto":"h2","server_name":"url"}},"error":"reading: context canceled"}


then if I Add flush_interval -1 to reverse_proxy

outcome will become this
{"resumed":false,"version":772,"cipher_suite":4865,"proto":"h2","server_name":"url"}},"error":"writing: client disconnected"}

mholt:
Yes, it was released; I'm not sure why it didn't make it into the release notes, but it's definitely released with v2.6.

If you open a new issue with instructions to reproduce the behavior we can look into it.

mholt:
This behavior might possibly be changing, by splitting it out into a separate config option other than `flush_interval`, since there's really two things going on there and that's basically a bug.

See [8c9e87d](https://github.com/caddyserver/caddy/commit/8c9e87d0a60cd5395c65dd354522c0865adc47b5) -- for now I've implemented a new option, `ignore_client_gone`, to enable this behavior that until now only required setting `flush_interval` to -1. But `flush_interval` should really just change the flush interval, and the new option should configure the client disconnect handling.

Issue #4498, opened 2021-12-29
Does reverse_proxy handler support proxying H2C traffic to a UDS?


fjn308:
To proxy gRPC traffic to a gRPC inbound of Xray (similar to V2Ray), I used to let Xray listen on a port, like 4443, and let Caddy pick gRPC traffic with right Path (here it is `/example/TunMulti`) out, then proxy it to an upstream (here it is `:4443`) . The config file reads:

``` json
{
    "apps": {
        "http": {
            "servers": {
                "fall": {
                    "listen": [
                        "unix//dev/shm/fall.sock"
                    ],
                    "allow_h2c": true,
                    "routes": [
                        {
                            "handle": [
                                {
                                    "handler": "subroute",
                                    "routes": [
                                        {
                                            "match": [
                                                {
                                                    "protocol": "grpc",
                                                    "path": [
                                                        "/example/TunMulti"
                                                    ]
                                                }
                                            ],
                                            "handle": [
                                                {
                                                    "handler": "reverse_proxy",
                                                    "transport": {
                                                        "protocol": "http",
                                                        "versions": [
                                                            "h2c"
                                                        ]
                                                    },
                                                    "upstreams": [
                                                        {
                                                            "dial": ":4443"
                                                        }
                                                    ]
                                                }
                                            ]
                                        }
                                    ]
                                }
                            ]
                        }
                    ]
                }
            }
        }
    }
}
```

There is something else I need to explain:
Here a VLESS (a protocol) inbound of Xray is working behind :443 (not `:4443`), it decrypts all TLS traffic and foward everything non-VLESS to a fallback, which can be a port or a UDS (here it is `fall.sock`, on which Caddy is listening). The traffic flows through Caddy has to be H2C because it has been decrypted, and the upstream (`:4443` or `grpc.sock` in the following) is listened by a gRPC inbound of Xray.

Everything works fine until I try to replace `:4443` (a port) to `grpc.sock` (a UDS), that is, changing

``` json
"dial": ":4443"
```

to

``` json
"dial": "unix//dev/shm/grpc.sock"
```

A similar change is made in the config file of Xray. Such configs do not work, and Caddy reports some error like:

``` json
"msg": "dial tcp: lookup /dev/shm/grpc.sock: no such host"
```

My question is:
Does reverse_proxy handler support proxying H2C traffic to a UDS? If it does, I hope to get some help from you. Many thanks.

[...]

PR #4953, opened 2022-08-12
reverseproxy: Add `unix+h2c` Caddyfile network shortcut

francislavoie:
As requested in https://github.com/caddyserver/caddy/issues/4498, just adding a little shorthand to make this work:

```
reverse_proxy unix+h2c//dev/shm/grpc
```

/cc @PaTTeeL

### DIFF
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_h2c_shorthand.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_h2c_shorthand.txt
@@ -1,6 +1,8 @@
 :8884
 
 reverse_proxy h2c://localhost:8080
+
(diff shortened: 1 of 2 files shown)


mholt (approved):
I do find it strange that `h2c` is set as a scheme and that `unix+h2c` is a network type, but it kinda makes sense. I can't think of a better alternate way to do it, so I guess we can go with this.

francislavoie:
Yeah, that's just kinda how the code was set up. We have the "Caddy network addresses" which map to Go's stuff, and then we have the scheme shortcuts for Caddyfile upstream addresses... and H2C + Unix is actually both of those.

I'm using `+` cause we have precedent with `srv+https` (although that's deprecated now with `dynamic srv` but I digress)

Issue #4962, opened 2022-08-16
Race condition with file server


abdusco:
I have a caddy file server serving files from a SAMBA network share.  
Normally it's adequate for serving few big files (I can easily saturate a gigabit connection), but it struggles with many small files as the OS has to make lots of network requests to stat all files in a directory.

When I refresh a folder (call it $dump) containing many files (hundreds of entries) it usually takes a couple of seconds until Caddy lists the folder and serves the HTML.

If, while the page is loading, I refresh another folder (call it $dir) on the same mount, I get served HTML that contains the content from both the $dump and $dir, or even the entire page is replaced with the contents of $dump. 

Here's a video showing it in action:

https://user-images.githubusercontent.com/3243482/184756428-3ea2eb38-c0b4-40bd-872e-197b9a5e6ebe.mp4

<details>
<summary>folder structure shown in the video</summary>
<pre><code>
.
|-- filedump
|   |-- dump_1
|   |-- ...
|   |-- ... (1000 files)
|   `-- dump_1000
|-- root_1
|-- ...
|-- ... (10 files)
|-- root_10
`-- sub
    `-- ...
</code></pre>
</details>

<details>
<summary>poor man's disk speed test</summary>
<pre><code>
$ dd if=/dev/zero of=io.img bs=512 count=1000
1000+0 records in
1000+0 records out
512000 bytes (512 kB, 500 KiB) copied, 0.0282936 s, 18.1 MB/s

$ dd if=/dev/zero of=io.img bs=8k count=10k
10240+0 records in
10240+0 records out
83886080 bytes (84 MB, 80 MiB) copied, 1.01272 s, 82.8 MB/s

$ dd if=/dev/zero of=io.img bs=1M count=256 conv=fdatasync
256+0 records in
256+0 records out
268435456 bytes (268 MB, 256 MiB) copied, 2.61678 s, 103 MB/s

$ dd if=/dev/zero of=io.img bs=64M count=1 oflag=dsync
1+0 records in
1+0 records out
67108864 bytes (67 MB, 64 MiB) copied, 1.73932 s, 38.6 MB/s
</code></pre>
</details>

Here's the relevant caddyfile config:

```caddyfile
sub.example.com {
    root * /mnt/box/files  # CIFS mounted drive

    @private {
        not file {
            try_files {path}
        }
    }

    file_server browse
    basicauth @private {
        user hash
    }
}
```

I'm running:

```
> caddy version
[...]

PR #4963, opened 2022-08-16
fileserver: reset buffer before using it (#4962)

abdusco:
Resets buffer acquired from the pool before using it.

Fixes https://github.com/caddyserver/caddy/issues/4962


abdusco:
Can this be tested easily? I can also add some tests, if needed.

mholt (approved):
Awesome, thanks for the report and investigating where the problem was! Made this super easy.

Issue #4954, opened 2022-08-13
admin endpoint will not restart when configuration failed for new admin endpoint


WeidiDeng:
When reloading caddy configurations, if new admin endpoint address is not available, the configuration will be rollback, but admin endpoint will be permanently off.

step to reproduce.

1. run caddy with blank caddy file.
2. run another process that will use a port, say localhost:8080
3. change caddy admin endpoint to localhost:8080 using curl or something
4. err: address already in use
5. caddy no longer listens 2019

PR #4964, opened 2022-08-16
admin: Don't stop old server if new one fails

WeidiDeng:
Fixes [4954](https://github.com/caddyserver/caddy/issues/4954)

Don't stop admin server when error occurs.


mholt:
Thanks, I'll look at this soon!

mholt (approved):
Cool, I was able to verify the bug from the issue #4954 and that this patch fixes it. I revised the patch slightly to use the error value instead of creating a new boolean pointer. Still works :upside_down_face: 

LGTM. Thank you again for contributing this!

PR #4971, opened 2022-08-21
caddyhttp: Set `http.error.message` to the HandlerError message

### LINKED ISSUE #4969: Custom error message string is not output anywhere
The `error` directive accepts a message argument, but this string doesn't appear in any of the placeholders listed here https://caddyserver.com/docs/json/apps/http/#servers/errors/routes nor does it appear in the response body when a `handle_errors` template is not used (a zero size response body is returned).

If I configure something like `error "Forums are down for maintenance" 503` the placeholder values are

`http.error.status_code` "503"
`http.error.status_text` "Service Unavailable"
`http.error.message` ""
`http.error.trace` "caddyhttp.StaticError.ServeHTTP (staticerror.go:110)"
`http.error.id` "6persmrad"

Versions: 2.5.2 and fe5f5dfd6aa2f48e92897327c9c7d8c8583a3d72

francislavoie:
Fix #4969

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -526,8 +526,9 @@ func (*HTTPErrorConfig) WithError(r *http.Request, err error) *http.Request {
 	if handlerErr, ok := err.(HandlerError); ok {
 		repl.Set("http.error.status_code", handlerErr.StatusCode)
 		repl.Set("http.error.status_text", http.StatusText(handlerErr.StatusCode))
-		repl.Set("http.error.trace", handlerErr.Trace)
 		repl.Set("http.error.id", handlerErr.ID)
+		repl.Set("http.error.trace", handlerErr.Trace)
+		repl.Set("http.error.message", handlerErr.Err.Error())
 	}
 
 	return r


mholt:
Thanks! Just to make sure I understand, does this mean that `{http.error}` is the generic error message if no custom message is set, and that `{http.error.message}` is always only the custom message?

francislavoie:
For example:

```
:8883 {
    error "Custom Error Message" 503

    handle_errors {
        respond "
Message: {http.error.message}
Error: {http.error}
Status: {http.error.status_code}
Text: {http.error.status_text}
Trace: {http.error.trace}
"
    }
}
```

Will write out:

```
Message: Custom Error Message
Error: {id=1bdsihwsh} caddyhttp.StaticError.ServeHTTP (staticerror.go:110): HTTP 503: Custom Error Message
Status: 503
Text: Service Unavailable
Trace: caddyhttp.StaticError.ServeHTTP (staticerror.go:110)
```

mholt (approved):
Ah that's pretty good. LGTM! Thanks for the patch. Not sure how I forgot this... heh.

francislavoie:
Yeah it's kinda wild, we've always documented it but it never worked :rofl:

bones-was-here:
Thanks for implementing this, but there seems to be a regression.  When `file_server` generates a 404, Caddy simply terminates the connection without returning any status code or body.  It doesn't log anything to stdout/stderr but I'm guessing it doesn't like `handlerErr.Err.Error()` being called when no message is set.  It returns an error normally when the message is set by the `error` directive.

francislavoie:
Please provide your full config and a reproduce case. I don't want to make assumptions like we had to do in the previous thread.

I find it _very_ unlikely that this PR would have caused any kind of regression. It only sets a variable, that's it.

Issue #4939, opened 2022-08-08
httpError template function not documented


iUnknwn:
The Caddy template library includes the function `httpError` to return a status code like 404 when processing a template. This function is used on Caddy's [own documentation website code.](https://github.com/caddyserver/website/blob/master/src/docs/index.html)

However, the function isn't included in the Caddy template function list: https://caddyserver.com/docs/modules/http.handlers.templates

It would be great if this was included in the documentation - being able to return errors (particularly 404 errors) from templates is quite useful.

mholt:
Oh yeah, it's because it didn't work until Go ... 1.17 or 1.18 I think? So we left it undocumented. Probably time to document it now though.

Malankar:
I would like to work on this. If possible can anyone guide me through it and can you assign me this issue if no one is working on it currently

mholt:
@Malankar Sure. Here's the function to be documented, for reference:

https://github.com/caddyserver/caddy/blob/c7772588bd44ceffcc0ba4817e4d43c826675379/modules/caddyhttp/templates/tplcontext.go#L400-L402

Here's where the documentation comment is:

https://github.com/caddyserver/caddy/blob/c7772588bd44ceffcc0ba4817e4d43c826675379/modules/caddyhttp/templates/templates.go#L34-L262

That comment just needs to be updated to add this function. Remember it is formatted using Markdown.

Malankar:
> @mholt I have made the necessary changes should I make a pull request 
>  
> https://github.com/Malankar/caddy/blob/document/modules/caddyhttp/templates/templates.go#L191-L201

francislavoie:
Yes, please make a pull request.

Malankar:
Thanks for the guidance @mholt and @francislavoie

PR #4972, opened 2022-08-21
templates: Document `httpError` function

Malankar:
Added the httpError template function for this issue #4939


mholt (approved):
Thanks for the contribution! Sorry it took me a while to get to this.

PR #4975, opened 2022-08-23
templates: cap of slice should not be smaller than length

### LINKED ISSUE #4974: A bug of custom functions in templates
modules/caddyhttp/templates/templates.go
image
This should be `customFuncs := make([]template.FuncMap, 0,len(fnModInfos))`

jedy:
fix #4974


mholt (approved):
Thanks for the patch!

PR #4977, opened 2022-08-23
caddyhttp: Fix for nil `handlerErr.Err`

francislavoie:
Followup to #4971, need to check that there actually is a wrapped error before trying to grab its message.


mholt (approved):
Ah, right. Forgot that Err isn't always set. Thanks!

Issue #3803, opened 2020-10-18
Improve performance of FastCGI transport (and clean up code)


mholt:
The current `fastcgi` implementation is functional, but clunky and slow.

Go does not have a standard FastCGI client implementation (note that we are not a _responder_ we are a client), so in January 2015 we forked the code from http://bitbucket.org/PinIdea/fcgi_client (which is forked from https://code.google.com/p/go-fastcgi-client/) -- and I am very grateful to the original authors for publishing it under a liberal license, so I did not have to go learn the nitty-gritty of the FastCGI protocol and implement it myself. However, the code was never really intended for heavy use.

But now, Caddy is used in heavy production environments where performance matters. Most of our code base could benefit from optimizing, but the FastCGI client should still have lots of low-hanging fruit that makes significant optimizations without too much trouble. For example, we could reduce buffering and pool buffers.

[The client.go file needs the most attention.](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/fastcgi/client.go)

This file would benefit _greatly_ from a refactoring to clean it up, and I've determined that a refactor will be necessary to support optimizations like buffer pooling.

I don't know how soon I will be able to get around to it, so if someone wants to take up the challenge of working on this file, please feel free to comment below and discuss your plans. [The FastCGI protocol](http://www.mit.edu/~yandros/doc/specs/fcgi-spec.html) is not _particularly_ complex in its most basic form, so you may find that rewriting the client from scratch and using the existing code (and [the spec](http://www.mit.edu/~yandros/doc/specs/fcgi-spec.html)) as a guide could be the most straightforward solution.

This code is used by a lot of people in a lot of important environments. Benchmark tests (and regular unit tests) should support any changes.

Thank you to anyone who helps!

mholt:
For instance, the `newWriter()` function is extremely inefficient currently:

https://github.com/caddyserver/caddy/blob/97caf368eea8d2c33a7786fbe3471b83b5b294dc/modules/caddyhttp/reverseproxy/fastcgi/client.go#L310-L314

It appears at the top of all memory profiles as a huge allocator, because the buffer isn't pooled. (Look closely and you'll see why using `sync.Pool` is impossible without a refactor.)

raygervais:
Hey!
I've been following caddy since the release of 2.0, hoping to contribute here over time and where possible.

Would love to start with the low hanging fruit and work my way through with guidance.

raygervais:
[...]

PR #4978, opened 2022-08-23
fastcgi: Optimize FastCGI transport

WeidiDeng:
Trying to fix [3803](https://github.com/caddyserver/caddy/issues/3803).

I tried it on a small php site, no problem so far. Extensive testing is required.

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/pool.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/pool.go
@@ -1,0 +1,32 @@
+package fastcgi
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"sync"
+)
+
+var (
+	bufPool = sync.Pool{
+		New: func() any {
+			return new(bytes.Buffer)
+		},
+	}
+	bufWriterPool = sync.Pool{
+		New: func() any {
+			return bufio.NewWriterSize(nil, maxWrite)
+		},
+	}
+)
+
+func getBufWriter(w io.Writer) *bufio.Writer {
+	bw := bufWriterPool.Get().(*bufio.Writer)
+	bw.Reset(w)
+	return bw
+}
+
+func putBufWriter(bw *bufio.Writer) {
+	bw.Reset(nil)
+	bufWriterPool.Put(bw)
+}
--- a/modules/caddyhttp/reverseproxy/fastcgi/pool.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/pool.go
@@ -1,3 +1,17 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package fastcgi
 
 import (
--- a/modules/caddyhttp/reverseproxy/fastcgi/pool.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/pool.go
@@ -1,32 +1,12 @@
 package fastcgi
 
 import (
-	"bufio"
 	"bytes"
-	"io"
 	"sync"
 )
 
-var (
-	bufPool = sync.Pool{
-		New: func() any {
-			return new(bytes.Buffer)
-		},
-	}
-	bufWriterPool = sync.Pool{
-		New: func() any {
-			return bufio.NewWriterSize(nil, maxWrite)
-		},
-	}
-)
-
-func getBufWriter(w io.Writer) *bufio.Writer {
-	bw := bufWriterPool.Get().(*bufio.Writer)
-	bw.Reset(w)
-	return bw
-}
-
-func putBufWriter(bw *bufio.Writer) {
-	bw.Reset(nil)
-	bufWriterPool.Put(bw)
+var bufPool = sync.Pool{
+	New: func() any {
+		return new(bytes.Buffer)
+	},
 }
--- a/modules/caddyhttp/reverseproxy/fastcgi/client.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/client.go
@@ -131,6 +131,7 @@ type FCGIClient struct {
 	mutex     sync.Mutex
 	rwc       net.Conn
 	h         header
+	buf       *bytes.Buffer
 	stderr    bytes.Buffer
 	keepAlive bool
 	reqID     uint16
@@ -157,14 +158,14 @@ func DialWithDialerContext(ctx context.Context, network, address string, dialer 
 }
 
 // DialContext is like Dial but passes ctx to dialer.Dial.
-func DialContext(ctx context.Context, network, address string, timeout time.Duration) (fcgi *FCGIClient, err error) {
-	return DialWithDialerContext(ctx, network, address, net.Dialer{Timeout: timeout})
+func DialContext(ctx context.Context, network, address string) (fcgi *FCGIClient, err error) {
+	return DialWithDialerContext(ctx, network, address, net.Dialer{})
 }
 
 // Dial connects to the fcgi responder at the specified network address, using default net.Dialer.
 // See func net.Dial for a description of the network and address parameters.
 func Dial(network, address string) (fcgi *FCGIClient, err error) {
-	return DialContext(context.Background(), network, address, 0)
+	return DialContext(context.Background(), network, address)
 }
 
 // Close closes fcgi connection
@@ -175,16 +176,13 @@ func (c *FCGIClient) Close() {
 func (c *FCGIClient) writeRecord(recType uint8, content []byte) (err error) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
-	buf := bufPool.Get().(*bytes.Buffer)
-	buf.Reset()
-	defer bufPool.Put(buf)
 	c.h.init(recType, c.reqID, len(content))
-	if err = binary.Write(buf, binary.BigEndian, c.h); err != nil {
+	if err = binary.Write(c.buf, binary.BigEndian, c.h); err != nil {
 		return err
 	}
-	buf.Write(content)
-	buf.Write(pad[:c.h.PaddingLength])
-	_, err = buf.WriteTo(c.rwc)
+	c.buf.Write(content)
+	c.buf.Write(pad[:c.h.PaddingLength])
+	_, err = c.buf.WriteTo(c.rwc)
 	return err
 }
 
@@ -312,25 +310,32 @@ func (w *streamReader) Read(p []byte) (n int, err error) {
 // Do made the request and returns a io.Reader that translates the data read
 // from fcgi responder out of fcgi packet before returning it.
 func (c *FCGIClient) Do(p map[string]string, req io.Reader) (r io.Reader, err error) {
+	c.buf = bufPool.Get().(*bytes.Buffer)
+	c.buf.Reset()
+	defer bufPool.Put(c.buf)
+
+	writer := newWriter(c)
+	defer writer.recycle()
+
 	err = c.writeBeginRequest(uint16(Responder), 0)
 	if err != nil {
 		return
 	}
 
-	err = c.writePairs(Params, p)
+	writer.sw.recType = Params
+	err = writer.writePairs(p)
 	if err != nil {
 		return
 	}
 
-	body := newWriter(c, Stdin)
-	defer body.recycle()
+	writer.sw.recType = Stdin
 	if req != nil {
-		_, err = io.Copy(body, req)
+		_, err = io.Copy(writer, req)
 		if err != nil {
 			return nil, err
 		}
 	}
-	err = body.Close()
+	err = writer.endStream()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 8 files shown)


mholt:
Oooh, I can't wait to review this!!

WeidiDeng:
Out of curiosity, I checked [stdlib fcgi server implementation](https://pkg.go.dev/net/http/fcgi@go1.19), and found original implementation is mostly a copy of the fcgi.go. For example, writeRecord, writePairs, and record reading.

Searching through go issues, I found nobody mentioned anything about performance, I guess not many people use it anyway.

mholt:
@WeidiDeng I'd be interested in a profile to compare memory usage or a CPU benchmark to compare performance. I don't have a PHP backend installed though. Would you or someone like to do some performance testing?

WeidiDeng:
@mholt I deployed nextcloud with docker compose and tested there. Large file upload and stress test with [wrk](https://github.com/wg/wrk) only. I remember with this patch, largest memory allocator is actually http2 frame writing related? Will upload those benchmarks when I find them.

WeidiDeng:
Latest commit changes are drastic, but in my testing environment there are no problems.

[profile.tar.gz](https://github.com/caddyserver/caddy/files/9429602/profile.tar.gz)

Here is the benchmark using nextcloud, separated by with or without this patch, and in 2 use cases, large file upload and website browsing.

mholt:
@WeidiDeng Thanks for the profiles! I'm looking at them now. Interesting, so far the patch version allocates about 13% less heap memory (if I'm reading these right). Have you noticed any significant difference in req/sec?

WeidiDeng:
image

Upper one is with this patch, lower one is build from master without this patch. Using go 1.19. header.lua is setting request cookies as to simulate user browser nextcloud directories.

Around 10% improvement.

mholt:
@WeidiDeng Am I reading the results wrong? The lower one is 3 req/sec higher than the upper one. (But they're so similar anyways.)

WeidiDeng:
I write a simple file server using golang fastcgi.

image

Previous test result is because the test bottleneck is mysql server, so there is no significant difference.

[profile.tar.gz](https://github.com/caddyserver/caddy/files/9442557/profile.tar.gz)

WeidiDeng:
@mholt updated benchmark using php_info() as output.

First nginx in docker as baseline.

image

Then caddy.

image

[Uploading profile.tar.gz…]()

mholt:
@WeidiDeng That's awesome!! Way more useful information here I think.

The profile clearly shows about 10% less memory being allocated, and only ~5s spent on GC compared to 36s without the patch. Req/sec appear to be on-par with or higher than nginx. Very good!

Definitely interested in having more people test this and then getting this merged in :tada:

tgulacsi:
I've tested it with our main use, MantisBT issue tracker, and it does work flawlessly.
The speed is the same as with 2.4.5 - MantisBT is a very conservative, low request count and resource utilizing PHP app...

modules/caddyhttp/reverseproxy/fastcgi/pool.go
```diff
@@ -0,0 +1,12 @@
+package fastcgi
+
+import (
+	"bytes"
+	"sync"
+)
+
+var bufPool = sync.Pool{
```
tgulacsi, line 8: Maybe it's not a common in this codebase, but what about some type with Acquire and Release methods that is type safe and does the Reset on Release - less bolerplate on use site? Like

```
var bufPool = bufferPool{p: sync.Pool{ New: func() any { return new(bytes.Buffer) } } }
type bufferPool struct {
  p sync.Pool
}
func (p bufferPool) Acquire() *bytes.Buffer { return p.p.Get().(*bytes.Buffer) }
func (p bufferPool) Release(b *bytes.Buffer) { b.Reset(); p.p.Put(b) }
```

mholt: @tgulacsi The buffer should be reset when _retrieving_ it from the pool, since a buffer that is put back into the pool may not be used again.

Personally, I prefer the vanilla pool style without a wrapper... but that's just me. I don't know if I have a good technical reason though :smile:

tgulacsi: Are you sure? 
AFAIK a pool is for reuse (minus the GC), so items put may be reused later.

https://pkg.go.dev/sync#Pool says "Pool's purpose is to cache allocated but unused items for later reuse".

So it _is_ for reuse, and newly allocated buffers does not need to Reset, only Put'd ones.


Anyway, by calling Reset before Put, the Get won't need to call Reset, and newly allocated Buffers won't be Reset'd.

WeidiDeng: That same link contains an example in which reset is called after getting buffer from the pool.

Anyway, why bother with the wrapper since there is only one call site?

mholt: @tgulacsi Yes, I'm sure. From the same godoc:

> If the Pool holds the only reference when this happens, the item might be deallocated.

tgulacsi: "Any item stored in the Pool may be removed automatically at any time without notification. If the Pool holds the only reference when this happens, the item might be deallocated."

The second part refers to the first part. So this is in effect only when the Pool is evicted.


But if you're not used to use such a type decoration for easier and more error-proof use of a pool, I'll not argue anymore.

mholt:
Thanks for testing @tgulacsi ! That's great to know. I will try to review this soon.

mholt (approved):
@WeidiDeng This is really great, and I'm looking forward to merging this!

There's nothing obviously wrong that I can see. I'm not a PHP guy so I'm relying on those of you with PHP apps to test it out. (This will go into a beta release first.) But it LGTM and I'd be ready to merge it any time.

Have you run this with the race detector (`go run -race` or `go build -race`)?

Thanks so much for the exciting contribution!

mholt:
If we could figure out how to set the logger when the clientCloser is created, we don't have to make it allocate.

A cheap way is to pass `CaptureStderr` into the `Dial()` functions, but that's kind of ugly.

I wonder if a better way would be to create a `FastCGIClient{}` which has the `CaptureStderr` config. Then it has the `Dial()` functions as methods which can access fields on the Client. The Dial methods then return a Conn, more like how the Go standard library works.

mholt:
Here's the delta/difference between the profiles you posted above for `php_info()`, @WeidiDeng:

```
$ go tool pprof -diff_base=master/heap patch/heap 
File: caddy
Type: inuse_space
Time: Aug 29, 2022 at 8:48pm (MDT)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for -1721.29kB, 6.29% of 27380.04kB total
Dropped 5 nodes (cum <= 136.90kB)
Showing top 10 nodes out of 130
      flat  flat%   sum%        cum   cum%
-3587.50kB 13.10% 13.10% -3587.50kB 13.10%  runtime.allocm
 2056.01kB  7.51%  5.59%  2056.01kB  7.51%  bufio.NewReaderSize
-1184.27kB  4.33%  9.92% -1184.27kB  4.33%  runtime/pprof.StartCPUProfile
 1024.02kB  3.74%  6.18%  1024.02kB  3.74%  net.JoinHostPort
 -544.67kB  1.99%  8.17%  -544.67kB  1.99%  github.com/caddyserver/caddy/v2/modules/caddyhttp/reverseproxy/fastcgi.(*record).read
 -516.01kB  1.88% 10.05%  -516.01kB  1.88%  runtime/pprof.(*profMap).lookup
     515kB  1.88%  8.17%      515kB  1.88%  sync.(*Pool).pinSlow
  514.63kB  1.88%  6.29%   514.63kB  1.88%  google.golang.org/genproto/googleapis/rpc/errdetails.init
     514kB  1.88%  4.41%      514kB  1.88%  hash/crc32.archInitCastagnoli
 -512.50kB  1.87%  6.29%  -512.50kB  1.87%  go.uber.org/zap/internal/bufferpool.init.func1


$ go tool pprof -diff_base=master/profile\?seconds\=30 patch/profile\?seconds\=30 
File: caddy
Type: cpu
Time: Aug 29, 2022 at 8:47pm (MDT)
Duration: 60.04s, Total samples = 70.11s (116.78%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for -25.11s, 35.82% of 70.11s total
Dropped 824 nodes (cum <= 0.35s)
Showing top 10 nodes out of 179
      flat  flat%   sum%        cum   cum%
    -5.53s  7.89%  7.89%    -31.07s 44.32%  runtime.gcDrain
    -4.35s  6.20% 14.09%     -4.62s  6.59%  runtime.(*lfstack).pop (inline)
    -3.19s  4.55% 18.64%     -9.58s 13.66%  runtime.scanobject
    -2.13s  3.04% 21.68%     -2.13s  3.04%  runtime.(*lfstack).push
    -2.06s  2.94% 24.62%     -2.10s  3.00%  runtime.pageIndexOf (inline)
    -1.86s  2.65% 27.27%     -1.86s  2.65%  runtime.markBits.isMarked (inline)
    -1.73s  2.47% 29.74%     -2.10s  3.00%  runtime.findObject
    -1.54s  2.20% 31.94%     -1.54s  2.20%  runtime.memclrNoHeapPointers
    -1.43s  2.04% 33.98%     -1.43s  2.04%  runtime/internal/syscall.Syscall6
    -1.29s  1.84% 35.82%     -8.90s 12.69%  runtime.gentraceback
```

Look at the green boxes (heap, then CPU):

profile001

profile002

And for your file server profile:

```
$ go tool pprof -diff_base=master/heap patch/heap 
File: caddy
Type: inuse_space
Time: Aug 29, 2022 at 1:21am (MDT)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for -26.97MB, 49.32% of 54.68MB total
Dropped 10 nodes (cum <= 0.27MB)
Showing top 10 nodes out of 137
      flat  flat%   sum%        cum   cum%
  -27.04MB 49.45% 49.45%   -27.04MB 49.45%  bufio.NewWriterSize
      -2MB  3.66% 53.11%       -2MB  3.66%  runtime.allocm
    1.06MB  1.93% 51.18%     1.06MB  1.93%  github.com/yuin/goldmark/util.init
    1.03MB  1.89% 49.29%     1.03MB  1.89%  github.com/caddyserver/caddy/v2/modules/caddyhttp/reverseproxy.glob..func2
       1MB  1.84% 47.46%        1MB  1.84%  bufio.NewReaderSize
      -1MB  1.83% 49.29%       -1MB  1.83%  runtime.malg
       1MB  1.83% 47.46%        1MB  1.83%  github.com/caddyserver/caddy/v2/modules/caddyhttp/reverseproxy.cloneRequest
      -1MB  1.83% 49.29%       -1MB  1.83%  net/http.(*Server).newConn
   -0.55MB  1.00% 50.29%    -0.55MB  1.00%  html.populateMaps
    0.53MB  0.97% 49.32%     0.53MB  0.97%  github.com/alecthomas/chroma.TypeRemappingLexer


$ go tool pprof -diff_base=master/profile\?seconds\=30 patch/profile\?seconds\=30 
File: caddy
Type: cpu
Time: Aug 29, 2022 at 1:21am (MDT)
Duration: 60.32s, Total samples = 624.40s (1035.16%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for -127.22s, 20.37% of 624.40s total
Dropped 1119 nodes (cum <= 3.12s)
Showing top 10 nodes out of 267
      flat  flat%   sum%        cum   cum%
   -18.98s  3.04%  3.04%    -18.98s  3.04%  runtime.memclrNoHeapPointers
   -17.72s  2.84%  5.88%    -18.99s  3.04%  runtime.(*lfstack).pop (inline)
   -17.43s  2.79%  8.67%    -44.92s  7.19%  runtime.gcDrainN
   -15.56s  2.49% 11.16%    -42.02s  6.73%  runtime.scanobject
   -14.68s  2.35% 13.51%    -98.20s 15.73%  runtime.gcDrain
   -10.73s  1.72% 15.23%    -12.95s  2.07%  runtime.findObject
   -10.50s  1.68% 16.91%    -10.60s  1.70%  runtime.pageIndexOf (inline)
    -8.80s  1.41% 18.32%     -8.80s  1.41%  runtime.markBits.isMarked (inline)
    -7.29s  1.17% 19.49%     -7.28s  1.17%  runtime.(*lfstack).push
    -5.53s  0.89% 20.37%    -40.12s  6.43%  runtime.gentraceback
```

Heap, then CPU:

profile001

profile002

Nice job :clap:

mholt:
@WeidiDeng I pushed one more commit that refactors client, it renames FCGIClient -> client, and should have slightly fewer allocations (like, 2 fewer -- we don't allocate the client or the clientCloser I think, since they're no longer pointers).

The tests pass on my machine, but it'd be good to have some more real-world PHP testing. Also run with the race detector (`-race`) and check for resource leaks. I _think_ everything gets closed properly though. Thanks!

mholt (approved):
LGTM. @WeidiDeng anything else you want to do before we merge this?

WeidiDeng:
@mholt I'm ready for it to be merged. About race detector, that one I haven't done. But that lock is unnecessary already, since reverse proxy can write to the body and the only time it's written is during sending request which isn't concurrent.

mholt:
@WeidiDeng Which lock are you referring to?

(I'll merge this in a few minutes then!)

WeidiDeng:
> @WeidiDeng Which lock are you referring to?
> 
> (I'll merge this in a few minutes then!)

FCGIClient sync.Mutex only used in writeRecord

mholt:
@WeidiDeng Looks like it was removed in this PR -- if there's no races, sounds good to me! Thank you! :smile: Please contribute again. (I will catch up to your other PRs shortly, I'm behind.)

PR #4980, opened 2022-08-23
httpcaddyfile: Add ocsp_interval global option

benburkert:
Allow the OCSP check interval in CertMagic to be configured in the
Caddyfile. Useful for testing certificate revocation with an ACME server
that provides short lived certificates.


mholt (approved):
Nice, thanks for the contribution. LGTM!

Can you tell me more about your use case? (For Caddy generally)

francislavoie:
What's the default interval? We'll need to document it

benburkert:
> What's the default interval? We'll need to document it

[1 hour](https://github.com/caddyserver/certmagic/blame/master/maintain.go#L677), where does it get documented?

francislavoie:
Since it's a new Caddyfile global option, here:

https://github.com/caddyserver/website/blob/master/src/docs/markdown/caddyfile/options.md

PR #4984, opened 2022-08-24
events: Begin implementing event system

### LINKED ISSUE #2344: Add events for On Demand Certificates
In my blog post [here](https://oliverdunk.com/2018/07/06/https-custom-domains-sass), I explain how I added two new events to Caddy:

- `OnDemandCertObtainedEvent`
- `OnDemandCertFailureEvent`

I'm opening this issue to ask if this is something that is worth PRing to introduce upstream? It seems like a minor change, and I can't see any downsides, but I wasn't sure if there was a reason for these not existing already.

### LINKED ISSUE #3643: Feature Request: Allow entry point for post certificate provisioning hooks
It would be nice to have a way to run commands after a certificate has been provisioned. This is useful when provisioning certificates for an application that needs to be reloaded or something. In my case I'd like to use Caddy to get certificates for a PostgreSQL instance, and reload PostgreSQL when a certificate is issued. I'm already using Caddy for the front-end, so I'd like to avoid using another tool just for PostgreSQL.

As mentioned in #3604, this used to be provided by the `on` plugin, which has since been removed.

Relates to https://github.com/abiosoft/caddy-exec/issues/2

### LINKED ISSUE #4039: Feature Request: Post to webhook on failed health check
This might be outside the scope of Caddy, but it would be fantastically useful if Caddy health checks could be configured to post to an arbitrary URL on failure (or maybe even success too).

Something like this in the JSON config for /apps/http/servers/routes/handle/reverse_proxy:

```
{
  "health_checks" : {
    //passive or active configs,
    "actions" : {
      "failure" : {
        "webhook_url": "your.apiurl.com/your-url"
      }
    }
  }
}
```

Ideally it would post some json to that webhook URL that would let the receiver identify the route, the datetime it failed, etc.

I think this could be really valuable for a lot of use cases (basically free uptime monitoring) and it feels like it would fit in well with a lot of other features that really make Caddy stand out in terms of practical convenience.

mholt:
This PR implements a new Caddy app called `events` that facilitates an eventing system. Modules may emit events with optional metadata, and modules may subscribe to events and run custom code or event handler modules.

Builds on but likely supersedes #4912 by @francislavoie who made a great start of this, and thought really hard about things. This PR takes his work further.

Notably, this is implemented outside the core of Caddy, and thus Caddy core cannot emit events lest we have an import cycle. I don't think that would be necessary, but if we find great value in emitting events from the core, we can move the events logic into Caddy core.

Users can configure event handling directly, or module code can also subscribe to handle events.

JSON config example:

```json
{
	"apps": {
		"events": {
			"subscriptions": [
				{
					"events": ["certificate_obtained"],
					"handlers": [
						{
							"handler": "exec",
							"command": "/bin/foobar",
							"args": ["--domain", "{event.data.domain}"]
						}
					]
				}
			]
		}
	}
}
```

This app supports Caddyfile configuration via global options:

```
{
	events {
		on certificate_obtained exec /bin/foobar --domain {event.data.domain}
	}
}
```

Events are emitted synchronously. This confers great simplicity and power by allowing event handlers to give information back to the running code so it can decide whether or how to continue. Events can be emitted asynchronously and thus not block critical paths by spawning a goroutine, but I'm worried whether this will back up under heavy load and cause problems. Still thinking about that.

There are numerous places in the code base that will need to be updated to emit events. Currently, there's spikey emission of these events:

- **`unhealthy`** when a reverse proxy backend goes down
- **`healthy`** when a reverse proxy backend comes back up
- Plus a few events emitted by CertMagic related to certificate issuance, TLS handshakes, etc. I need to spend some time updating CertMagic to better conform to our preferred interface(s) (CertMagic is not yet 1.0 so it might be a breaking change, but for the better).

Currently the only event handler module is `exec`, which runs a command when an event happens. Obviously, this is not ideal for extremely hot paths like HTTP requests, but may be more appropriate for lower-volume events like certificate issuance. This is handy as it can be used to have other programs reload the new certificate.

As Francis and I discussed in slack, this feature introduces the first time Caddy can run commands (without a third-party plugin). We've intentionally avoided it previously because it makes us nervous for Caddy to run commands on the system, as we don't want to introduce any security vulnerabilities. However, this implementation requires the command to be hard-coded into the config (no placeholders are expanded) and so it should not allow arbitrary programs or code to be run unless the command in the config allows it. We also do not do any special parsing or interpretation of the command and arguments (other than expanding placeholders in arguments). No shell evaluation, no bash scripting, etc. As usual, we expect that your system is secure (i.e. no one can access the admin endpoint, which would be game over regardless of this feature anyway).

**I'd encourage anyone interested in this to try to find security issues before we release this.** I believe it's probably safe in terms of not popping shells or allowing RCEs, etc. But I hope you will verify. :slightly_smiling_face: 

This is still definitely WIP, so there may be some big changes between now and the release.

Thanks to everyone who tests it, reviews it, etc!

- Closes #2344
- Closes #3643
- Closes #4039


modules/caddyevents/app.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// add event info to replacer, make sure it's in the context
+	repl, ok := ctx.Context.Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
+	if !ok {
+		repl = caddy.NewReplacer()
+		ctx.Context = context.WithValue(ctx.Context, caddy.ReplacerCtxKey, repl)
+	}
+	repl.Set("event", e)
+	repl.Set("event.id", e.id)
+	repl.Set("event.timestamp", e.ts)
```
francislavoie, line 225: I'm thinking we should use a `repl.Map` instead for the timestamp, then support things like `event.ts_ms` and `event.ts_nano` etc which are computed on-demand. FWIW, `ts` is _quite_ understandable, I'd lean towards that vs the longer `timestamp`.

modules/caddyevents/app.go
```diff
@@ -0,0 +1,318 @@
...
+			return nil, false
+		}
+		key = strings.TrimPrefix(key, "event.data.")
+		if val, ok := data[key]; ok {
+			return val, true
+		}
+		return nil, false
+	})
+
+	// TODO: log that event was emitted... hm, maybe disable our logger by default? or only emit logs if any subscriptions are configured? or maybe debug level? hmm
```
francislavoie, line 240: I suppose we should just `debug` log it for now and if it does get truly noisy we could do some hacking to make it off by default. But for now, we probably won't have _that_ many events (yet)

Logging the events that happen is super useful for users to discover what they can listen for. But obviously we should ideally document them too.

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -353,7 +353,7 @@ func (h *Handler) doActiveHealthCheck(dialInfo DialInfo, hostAddr string, upstre
 	// passed health check parameters, so mark as healthy
 	if upstream.setHealthy(true) {
 		h.HealthChecks.Active.logger.Info("host is up", zap.String("host", hostAddr))
-		h.event.AsyncDispatch(NewActiveHealthyEvent(hostAddr))
+		h.events.Emit(h.ctx, "healthy", map[string]interface{}{"host": hostAddr})
```
francislavoie, line 356: Can we do some sugar for making the map? Seems like it would be annoying to make a map like this everywhere. I dunno.

DeanMarkTaylor:
Reading the subject "events: implement event system" - it made me think of the specification [CloudEvents](https://cloudevents.io/) with a good [Primer here](https://github.com/cloudevents/spec/blob/main/cloudevents/primer.md). 

I only mention it here to see if it has been considered regarding events from Caddy...
... noting that [Google's Eventarc](https://cloud.google.com/eventarc/docs) is used in "updated" Google Cloud functionality / APIs (seemingly) across the board (e.g. [Cloud Functions v2](https://firebase.google.com/docs/functions/beta).

I have no skin in this game, just mentioning it so it's on your radar.

francislavoie:
Thanks for pointing that out @DeanMarkTaylor, hadn't heard of it.

My initial design was pretty heavily based on https://symfony.com/doc/current/event_dispatcher.html due to my familiarity with it as a PHP dev.

I think we'll probably not go the route of conforming to a specific spec up-front, but I'm glad to see that we're already quite close to the same thing give or take a field or two: we use the term `name` instead of `type`, we use `origin` as the `source` (it's the actual module that emitted it), we already use UUID for `id`, and we use a `data` KV map. We also add a `ts` which isn't in the spec, and also we have an `Aborted` field to let handling inside Caddy get cancelled early.

If someone finds a usecase where CloudEvent compat would help, it should be trivial to make an adapter for it. But for now, I'm not convinced there really would be any kind of interop that would make it viable.

coolaj86:
Is it possible to specify the process user and group id for the exec?

I know Caddy has to be run as root on macOS to use privileges ports.

mholt:
@coolaj86 I thought MacOS doesn't require running as root for low ports anymore. :thinking: 

But sure, I can add that functionality easily enough. This is just bare-bones still.

I might actually pull the exec module out of here and make it a separate plugin.

coolaj86:
> I thought MacOS doesn't require running as root for low ports anymore. 🤔

That's news to me. Just tested on Monterey and I think you're right:

```sh
echo "goodbye" | nc -l 80
```

```sh
echo "hello" | nc localhost 80
```

modules/caddyhttp/reverseproxy/healthchecks.go
mholt: It looks much tidier when using `any` instead of `interface{}`, but yeah, the only sugar would be a type definition. I'd prefer to only do that if we actually need to lean on the type though. So far there's no behavior or type assertions that would require that.

modules/caddyevents/app.go
mholt: Alrighty. I went with `events.time` instead. `ts_ms` is a bit too cryptic for me :)

modules/caddyevents/app.go
mholt: Makes sense; done.

francislavoie (approved):

mholt:
@DeanMarkTaylor I implemented a simple CloudEvents-compatible exporter method. Let me know if that's not sufficient!

---

This PR is being merged into the `event` branch where work will continue.

Please follow #4912 for future updates on this feature.

PR #4990, opened 2022-08-29
caddyhttp: Copy logger config to HTTP server during AutoHTTPS

### LINKED ISSUE #4982: Propagate access logging config to implicit HTTP server from Auto HTTPS
If the Auto HTTPS routine creates an HTTP server to perform HTTP->HTTPS redirects, it doesn't inherit the logging config from the HTTPS server. It should, because it is a relevant request to log.

francislavoie:
Fix #4982

I set up a clone function, because although it did work to assign the HTTP server to the same pointer for the logger as the HTTPS server, it worries me in case we later need to change it at runtime.

### DIFF
--- a/modules/caddyhttp/autohttps.go
+++ b/modules/caddyhttp/autohttps.go
@@ -93,6 +93,9 @@ func (app *App) automaticHTTPSPhase1(ctx caddy.Context, repl *caddy.Replacer) er
 	// https://github.com/caddyserver/caddy/issues/3443)
 	redirDomains := make(map[string][]caddy.NetworkAddress)
 
+	// the configured logger for an HTTPS enabled server
+	var logger *ServerLogConfig
+
 	for srvName, srv := range app.Servers {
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thanks Francis. Does the logger get changed at runtime? (I'll go look when I'm not mobile.)

LGTM. I personally wouldn't mind the simplicity of giving it the same pointer if it's safe to do so, but I also am good with merging this since it only happens at startup.

francislavoie:
It doesn't get changed at runtime now no, it's just a bunch of scalar exported fields, none are touched by provisioning. But I'm worried that it we do add fields later we could get caught out by not cloning it so it feels more correct to do that.

mholt:
I can appreciate that. If we want to share the logger though, why not use the same one? Like is there a case where we'd want to change the original logger but not the one for the HTTP redirects?

francislavoie:
Probably not. But it just makes me mildly nervous to reference the same config struct in two places, is all. :man_shrugging:

PR #4993, opened 2022-08-30
fileserver: Support glob expansion in file matcher

### LINKED ISSUE #4988: Glob matcher for file names on the *filesystem*
[Placeholder for a feature requested in this community discussion](https://caddy.community/t/matching-admin-with-filesystem-admin-index-md4-html).

`file_server, try_files, file, url, path_regexp` and other matchers are very powerful for matching requests and uri paths with wildcards/regex.

However they do not seem to accommodate regex/wildcards for file names on the caddy *filesystem*.

The closest is to this is the configuration for…

    ‘try_policy’: ‘first_exist|smallest_size|largest_size|most_recent_modified’

…because this hints at rules to select from multiple matching files on the filesystem, but it does not provide the flexible globbing available to request matching.

A suggested implementation approach from the community discussion linked above is:

> I think it would essentially mean some kind of first_glob policy; would be the same as first_exist but using [filepath.Glob()](https://pkg.go.dev/path/filepath#Glob) to match for files instead, returning the first result from that (if there’s multiple results then… result is kinda undefined).

The community discussion also explains one known use case.

It would be great if Caddy could do regex/wilcard/glob filesystem matching for this and similar use cases.

mholt:
This change adds support for glob patterns in the file matcher's `try_files` parameter (and by extension, the `try_files` Caddyfile directive).

For a directory like this:

```
index.1.html   17KB
index.2.html   9KB
index.3.html   23KB
```

and a config like this:

```
try_files {http.request.uri.path.file.base}.*.html
```

a request to `/index.html` would rewrite to `/index.1.html`.

Also added the ability to change the policy on the `try_files` directive:

```
try_files {http.request.uri.path.file.base}.*.html {
	policy largest_size
}
```

For the same request, `/index.html`, this policy would rewrite to `/index.3.html`.

It'd be nice if we could figure out Caddyfile shorthands for the new `{http.request.uri.path.file.base}` and `{http.request.uri.path.file.ext}` placeholders. Maybe they should be like `{http.request.uri.path.file_base}` instead? We have `{path.*}` shorthands but that only works one more level deep. We could make it work 2 levels deep or just make different shorthands or use the underscore instead.

I was careful to escape any special glob characters when evaluating placeholders, so a request to `/*` can't lead to any surprising results. Not a security concern per-se, as the globbing is still within the site root, just unexpected.

Closes #4988. /cc @iamjrock -- can you please try this out?


modules/caddyhttp/replacer.go
```diff
@@ -143,6 +143,10 @@ func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.Respo
 			case "http.request.uri.path.dir":
 				dir, _ := path.Split(req.URL.Path)
 				return dir, true
+			case "http.request.uri.path.file.base":
+				return strings.TrimSuffix(path.Base(req.URL.Path), path.Ext(req.URL.Path)), true
+			case "http.request.uri.path.file.ext":
+				return path.Ext(req.URL.Path), true
```
francislavoie, line 149: Btw, with an input of `foo.tar.gz`, this will return `.gz` only. Should we write our own logic to repeat `.Ext()` to pop extensions off the end until none remain, and return that? :thinking:

mholt: Well, there are 2 extensions. Imagine a path of `index.012abcdf.html` -- there's only one file extension and the part in the middle is just a checksum. I'm not sure how to handle that better. It's arbitrary either way, and more commonly there is only 1 ext.

francislavoie:
Re placeholders, considering we already have `{file}`, I think we should just add `{file.*}` as well to cover these two new ones, i.e. `{file.base}` and `{file.ext}`.

iamjrock:
@mholt @francislavoie Wow guys! I wasn't expecting that, thank you.

Family commitments today but will test it out tomorrow first chance I get.

mholt:
@iamjrock Did you have a chance to test this out?

iamjrock:
> @iamjrock Did you have a chance to test this out?

Sorry no not yet - hoping to get it tested tonight!

mholt:
@iamjrock I'll be merging this today in preparation for the first beta release. Hope to hear from you before then if it doesn't work! :sweat_smile:

iamjrock:
Hi @mholt I am really sorry, I've had to tool down for a few days due to an unexpected family matter.

But I really appreciate what you've done here and really want to get you that feedback, so I have asked one of my colleagues to test it in my absence and I will follow up.

We're in GMT+8 here so it's early morning and I would expect a reply by close of biz today, possibly earlier.

mholt:
@iamjrock Ah, no worries -- honestly it's not a big deal, I just wanted to let you know I don't mean to ignore your feedback. I am just going forward with a very unannounced schedule, so I want you to know that I'm not ignoring you or unappreciative of your collaboration with this! :+1:

Hope your family is well. Don't go out of your way to test this; it's just going into the first beta, so if something is broken or doesn't meet your requirements we can still fix it later. Cheers!

iamjrock:
OK thx for the reprieve, all good now thx. My colleagues tested this and report that it is working great for our use case!

mholt:
Hurray, thanks for confirming!

PR #4997, opened 2022-08-31
caddyhttp: Add HTTP/3 port override for Alt-Svc header

### LINKED ISSUE #4996: [Feature request] HTTP3 custom port and caddyfile option
Hello, it's been some time since my forum posts below:

https://caddy.community/t/experimental-http3-behind-firewall-port-forwarding/14746/8

But very much hoping to continue this conversation and get a config server option implemented to enable a http3 custom port header.

This is for a (maybe common?) use case where caddy is behind another udp/tcp load balancer (eg. AWS ELB) and listening on an address other than the default 443.

Given this basic Caddyfile:
```
{
	http_port 8080
	https_port 8443
	debug
}

https://localhost:8443 {
	tls internal
	respond "hello there"
}
```
I currently get the following output (_note the alt-svc header values `8443` matches the `https_port` value_):
```
$ curl -I https://localhost:8443
HTTP/2 200
alt-svc: h3=":8443"; ma=2592000,h3-29=":8443"; ma=2592000
server: Caddy
content-length: 11
date: Wed, 31 Aug 2022 01:09:32 GMT
```

Right now, i could hard code the following (since upstream supports this):
```
	s.h3server.Port = 443
```
into this line here: https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/server.go#L488

If i recompile using xcaddy, i get the desired behavour (given the same Caddyfile):
```
curl -I https://localhost:8443
HTTP/2 200
alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
server: Caddy
content-length: 11
date: Wed, 31 Aug 2022 01:10:21 GMT
```

It seems to me a better option though not to hardcode this value, and instead expose config to control what port is advertised. 

eg.
```
s.h3server.Port = <value taken from json or caddyfile option TBC>
```

Before attempting a PR, it would be good to understand what the approach should be and/or where the config should sit. (I noticed the protocol option is deprecated for example in the code). 

Keen for any thoughts - is this potentially low hanging fruit? Or more likely to be difficult to implement?

Thanks! 😄

francislavoie:
Fix #4996

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -192,6 +192,7 @@ func (st ServerType) Setup(inputServerBlocks []caddyfile.ServerBlock,
 	httpApp := caddyhttp.App{
 		HTTPPort:      tryInt(options["http_port"], &warnings),
 		HTTPSPort:     tryInt(options["https_port"], &warnings),
+		HTTP3Port:     tryInt(options["http3_port"], &warnings),
 		GracePeriod:   tryDuration(options["grace_period"], &warnings),
 		ShutdownDelay: tryDuration(options["shutdown_delay"], &warnings),
 		Servers:       servers,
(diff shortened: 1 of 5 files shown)


mholt:
Deferring until a later (pre-?)release until the discussion on the issue has reached some sort of consensus

PR #4999, opened 2022-08-31
caddyhttp: Set Content-Type for static response

mholt:
Since the `Body` field is a `string`, and it is generally configured via command-line or text config file, it follows that a reasonable default Content-Type is `text/plain`. This allows `templates` to render templates included in the response, and makes it clear to clients that they do not need to download a binary file, without extra configuration of this header.

We also detect valid JSON (JSON can be a number or quoted string which is easily ambiguous as to Content-Type, but the most common form of JSON is an object or array, so we only do it for those structures) and set `application/json`. But those are the only 2 special cases where we set the header automatically. I think those are probably the most common use cases for `respond`: plain text and hard-coded JSON.

This should make using the `static_response` handler less cumbersome and more convenient, especially now that it has a CLI shortcut with `caddy respond`.

### DIFF
--- a/modules/caddyhttp/staticresp.go
+++ b/modules/caddyhttp/staticresp.go
@@ -92,13 +92,14 @@ type StaticResponse struct {
 	// header fields of the same names after normalization.
 	Headers http.Header `json:"headers,omitempty"`
 
-	// The response body. If non-empty, the Content-Type header
-	// will be added automatically if it is not explicitly configured
-	// nor already set on the response; the default value is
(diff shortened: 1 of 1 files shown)


francislavoie:
Worth adding, in addition to enabling `templates` support, it'll also enable `encode` support since it also matches on the response content type header to determine if it should do its work.

mholt:
True, as long as the body is the minimum length for encoding (I think 512 bytes currently), this should enable compression if that's configured!

Issue #4967, opened 2022-08-19
Handle stderr written from fastcgi transport


francislavoie:
In Caddy v1, we used to write the stderr we received from the fastcgi upstream straight to stderr.

In v2, we didn't port over the bit of code that handled the received stderr, so it just gets dropped. This means that errors reported from PHP can get lost if the PHP app doesn't have some other logging mechanism configured.

Some more context written out here https://caddy.community/t/equivalent-of-errors-stderr-in-caddy-2/16856/7

The relevant chunk of code from v1:

https://github.com/caddyserver/caddy/blob/240de5a5dd20caba4560b9463d532dc629aaab4c/caddyhttp/fastcgi/fastcgi.go#L186-L189

I'm unclear on exactly what we should do with this. There's a few different options, but I'm not sure which is best.

- Just log the `stderr` we get to the `http.reverse_proxy.transport.fastcgi` logger at `ERROR` level, always
  - Pro: this is super easy just a couple lines of code, no config
  - Con: it could be noisy for some, and it could be considered a "breaking change" in the sense that bumping up to a version with this might have a surprising amount of stuff written out; also the logs may contain sensitive info users didn't expect to suddenly get pumped to their logs
  - Note: users could always just configure the default logger to exclude these, and/or a second logger to write them elsewhere, but this is a bit tedious
- Add an opt-in option to the `fastcgi` transport to log `stderr` to its logger
  - Pro: things stay the same for users, no noise by default
  - Con: most users may not realize they might be getting errors from their fastcgi app that they're missing out on without explicitly turning this on; adds noise to the config
- Always log it, but at DEBUG level instead
  - Pro: also super easy, not visible by default but can be shown with the `debug` global option etc
  - Con: similar problem, not immediately obvious they might be missing out on it, but potentially more discoverable; less semantically correct because they are errors, really

Another thing to note, these logs tend to have newlines in them, which can be kinda awkward. For the JSON encoder it's fine, it'll just escape those. But for the console logger, it will render the newlines so logs won't be consistently one-per-line in that case, typically because of a stack trace. But to be fair, Go may write out panic logs which have newlines in them too so I don't think there's really a reasonable expectation that the console logger actually does have a one-per-line guarantee.

flga:
We're in the process of replacing nginx with caddy and I got "bitten" by this yesterday.
[...]

PR #5004, opened 2022-09-01
fastcgi: allow users to log stderr output (#4967)

flga:
Following the discussion in #4967 I've added support for logging any stderr messages sent by the upstream.

It is off by default and can be enabled with:
```
transport fastcgi {
    capture_stderr
}
```

Naming it `capture` gives us room to *not* buffer stderr records if it's off but it also isn't super clear that the main purpose is logging (at least right now). On the other hand, I don't know what else could be done with this *besides* logging. We should think whether this name makes sense.

Logging is done through the FastCGI transport logger.
Messages received on stderr are logged at WARN level by default, or ERROR level if the response status is >= 400.

The field name `body` was chosen instead of `message` because there's not really a clear definition of what a message is, there is no way for us to know, from a given stderr response, where one begins and another one ends. We cannot rely on fastcgi record boundaries alone (at least with PHP).

The output is messy but that's not something we can address, and this behaviour is pretty much equivalent to apache/nginx with the added bonus that formatting is preserved (albeit escaped) and we do not truncate the contents.

The logging in the fastcgi transport now also complies with `log_credentials` (also applies to logging of fastcgi params).
I'm not sure that we should be logging the fastcgi params though. It's somewhat redundant with the request data, but the alternative is telling users to do "printf debugging". I can see both sides so I defaulted to logging them, let me know what you think.

The changes could be a little cleaner (especially the handling of transfer encoding) but given that there's a PR to overhaul fastcgi as a whole I tried to keep the changeset as small as possible so that the merge is more obvious.
On that note, a test couldn't hurt, but I don't see how to output stderr records from a `net/http/fcgi` handler (what we use in the tests) so we'd pretty much just be asserting that the zap logger is called. Doesn't seem super worth it, especially since the overhaul is in progress, but since this got lost in the v1 -> v2 transition it's a good argument for doing it. Let me know what you think.

One thing I could not avoid were the comment formatting changes included in 1.19's gofmt.
I tried to fight it but it's painful. If this causes problems with our own parsing (for docs) let me know and I'll fix it up by hand.

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
@@ -35,16 +35,16 @@ func init() {
 
 // UnmarshalCaddyfile deserializes Caddyfile tokens into h.
 //
-//     transport fastcgi {
-//         root <path>
-//         split <at>
-//         env <key> <value>
-//         resolve_root_symlink
-//         dial_timeout <duration>
-//         read_timeout <duration>
-//         write_timeout <duration>
-//     }
-//
+//	transport fastcgi {
+//	    root <path>
+//	    split <at>
+//	    env <key> <value>
+//	    resolve_root_symlink
+//	    dial_timeout <duration>
+//	    read_timeout <duration>
+//	    write_timeout <duration>
+//	    capture_stderr
+//	}
 func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	for d.Next() {
 		for d.NextBlock(0) {
@@ -107,6 +107,12 @@ func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				}
 				t.WriteTimeout = caddy.Duration(dur)
 
+			case "capture_stderr":
+				if d.NextArg() {
+					return d.ArgErr()
+				}
+				t.CaptureStderr = true
+
 			default:
 				return d.Errf("unrecognized subdirective %s", d.Val())
 			}
@@ -120,31 +126,31 @@ func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 // Unmarshaler is invoked by this function) but the resulting proxy is specially
 // configured for most™️ PHP apps over FastCGI. A line such as this:
 //
-//     php_fastcgi localhost:7777
+//	php_fastcgi localhost:7777
 //
 // is equivalent to a route consisting of:
 //
-//     # Add trailing slash for directory requests
-//     @canonicalPath {
-//         file {path}/index.php
-//         not path */
-//     }
-//     redir @canonicalPath {path}/ 308
+//	# Add trailing slash for directory requests
+//	@canonicalPath {
+//	    file {path}/index.php
+//	    not path */
+//	}
+//	redir @canonicalPath {path}/ 308
 //
-//     # If the requested file does not exist, try index files
-//     @indexFiles file {
-//         try_files {path} {path}/index.php index.php
-//         split_path .php
-//     }
-//     rewrite @indexFiles {http.matchers.file.relative}
+//	# If the requested file does not exist, try index files
+//	@indexFiles file {
+//	    try_files {path} {path}/index.php index.php
+//	    split_path .php
+//	}
+//	rewrite @indexFiles {http.matchers.file.relative}
 //
-//     # Proxy PHP files to the FastCGI responder
-//     @phpFiles path *.php
-//     reverse_proxy @phpFiles localhost:7777 {
-//         transport fastcgi {
-//             split .php
-//         }
-//     }
+//	# Proxy PHP files to the FastCGI responder
+//	@phpFiles path *.php
+//	reverse_proxy @phpFiles localhost:7777 {
+//	    transport fastcgi {
+//	        split .php
+//	    }
+//	}
 //
 // Thus, this directive produces multiple handlers, each with a different
 // matcher because multiple consecutive handlers are necessary to support
@@ -154,7 +160,7 @@ func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 //
 // If a matcher is specified by the user, for example:
 //
-//     php_fastcgi /subpath localhost:7777
+//	php_fastcgi /subpath localhost:7777
 //
 // then the resulting handlers are wrapped in a subroute that uses the
 // user's matcher as a prerequisite to enter the subroute. In other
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -107,6 +107,12 @@ func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				}
 				t.WriteTimeout = caddy.Duration(dur)
 
+			case "capture_stderr":
```
francislavoie, line 110: Should also add this option to the `php_fastcgi` directive's unmarshal, so it can be used without `transport`

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
```diff
@@ -387,7 +421,7 @@ var headerNameReplacer = strings.NewReplacer(" ", "_", "-", "_")
 
 // Interface guards
 var (
-	_ zapcore.ObjectMarshaler = (*envVars)(nil)
+	_ zapcore.ObjectMarshaler = loggableEnv{}
```
francislavoie, line 424: I think it should be done like this:

```suggestion
	_ zapcore.ObjectMarshaler = (*loggableEnv)(nil)
```

modules/caddyhttp/reverseproxy/fastcgi/client.go
```diff
@@ -442,9 +459,19 @@ func (c *FCGIClient) Request(p map[string]string, req io.Reader) (resp *http.Res
...
+		}
 	} else {
-		resp.Body = clientCloser{c, io.NopCloser(rb)}
+		resp.Body = clientCloser{
+			FCGIClient: c,
+			Reader:     rb,
+			status:     resp.StatusCode,
+			logger:     c.logger,
+		}
 	}
```
francislavoie, line 475: Maybe we could refactor this to create the `Reader` as a variable in an `if/else` then just make the `clientCloser` once with the reader. A bit less duplication of code.

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
flga: whoops, didn't realize the shortcut allowed for overrides, fixed

modules/caddyhttp/reverseproxy/fastcgi/client.go
flga: Yeah that's what I was alluding to when I mentioned
> The changes could be a little cleaner (especially the handling of transfer encoding)

That is to say, this was done on purpose so that we keep the symmetry with the old code, which should make the merge choice a little more obvious.

Should we just do it the right way now?

francislavoie: /cc @WeidiDeng up to you I'd say.

I'm not sure what order we'll end up merging, but I figure either of you can resolve it afterwards.

@WeidiDeng how much work do you have left on your branch? Do you think it's pretty much good to go? Any loose threads?

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -303,6 +309,14 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
...
 				dispenser.Delete()
 				dispenser.Delete()
+
+			case "capture_stderr":
+				args := dispenser.RemainingArgs()
+				dispenser.Delete()
+				for range args {
+					dispenser.Delete()
+				}
+				fcgiTransport.CaptureStderr = true
```
mholt, line 319: Maybe `if dispenser.NextArg() { return nil, dispenser.ArgErr() }` would be a better way to handle this, since no args are expected.

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
```diff
@@ -74,6 +76,11 @@ type Transport struct {
 	// The duration used to set a deadline when sending to the FastCGI server.
 	WriteTimeout caddy.Duration `json:"write_timeout,omitempty"`
 
+	// Capture and log any messages sent by the upstream on stderr. Logs at WARN
+	// level by default. If the response has a 4xx or 5xx status ERROR level will
+	// be used instead.
+	CaptureStderr bool `json:"capture_stderr,omitempty"`
```
mholt, line 82: Wonder if they should be emitted at DEBUG level unless the response is 5xx (then WARN)? I imagine the _error itself_ is already emitted at an ERROR-level log?

mholt (commented):
Hey this is pretty good @flga -- thank you for contributing it. I look forward to @WeidiDeng's comments as well since he has a conflicting branch up to optimize the performance of this transport.

modules/caddyhttp/reverseproxy/fastcgi/client.go
WeidiDeng: @francislavoie There is not much to be done at the moment, except if there are bug reports. Fastcgi has one last optimization that I can think of is reusing connections, but that won't be for a while.

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
francislavoie: Kinda contradicts what I said to do btw; This matches the code we have for `resolve_root_symlink` which also has no options. Either one is fine though, as long as we're consistent.

modules/caddyhttp/reverseproxy/fastcgi/client.go
francislavoie: Alright. I guess we can merge this one first and then you can rebase & fixup your branch @WeidiDeng

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
mholt: Good point, but yeah... I think I prefer the simpler form if it works. I'm not sure why reading all the arguments and ignoring them is helpful.

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
francislavoie: I'd say it makes sense to log at WARN if they turned on this option. They opted in to want to hear about errors, so we should make it "visible".

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
francislavoie: Moreso abundance of caution with deleting the tokens, but throwing an error is technically more valid since we don't support args for those (yet). But extremely unlikely anyone will encounter this as a problem either way.

francislavoie (approved):
Tried this out, works great!

Thanks for making the executive decision for me with implementing this as an opt-in 😅

mholt:
@flga We both think this was a really high-quality proposal and change. We're going to send you an invite to be a collaborator, meaning you'll have push rights to the repo so in the future you can easily make patches on branches rather than forks. And you can participate in other ways as you would like. :slightly_smiling_face: Thank you again for contributing!

flga:
@mholt Thanks! That was a little unexpected, cheers :grin:

PR #5006, opened 2022-09-01
caddyhttp: Support `respond` with HTTP 103 Early Hints

### LINKED ISSUE #5005: Originate HTTP 103 Early Hints
HTTP 103 Early Hints ([draft RFC 8297](https://datatracker.ietf.org/doc/rfc8297/)) is a mechanism for servers to immediately send `Link` headers to clients so the clients can immediately start requests for those resources.

It is similar to HTTP/2 server push which immediately pushes resources to the client without an extra round trip, but instead of pushing the whole resource right away, the server tells the client, "Hey I'm still getting you the final response, but while you're waiting here's some resources you can download right now."

Enabled clients will read the `Link` headers from the HTTP 103 response and download those resources. Despite the extra round trip, this has a couple notable advantages over HTTP/2:

- The client can choose to ignore the download if the resource is already cached locally
- The server can suggest resources that it does not originate

In practice, HTTP/2 server push is difficult to get right because the server has to predict what the client needs to download. Pushing unnecessary resources actually _worsened_ performance, [according to Chrome data](https://developer.chrome.com/blog/early-hints/), but there was no reliable, efficient way of knowing what the client actually needed ahead of time.

103 responses are small, so despite the extra round trip, letting the client choose whether to initiate other requests to potentially multiple origins is actually a net performance benefit. This gain is amplified the longer the server takes to generate the final response.

I would suggest that Server Push and Early Hints have different purposes:

- **Push** when you know the client _WILL_ have to _REQUEST_ a resource _FROM YOU_
- **Hint** when you know the client _MAY_ have to _USE_ a resource

(Notice the differences: only push when the client "WILL REQUEST FROM YOU", whereas you can hint when the client "MAY USE" a resource.)

Since HTTP/2 server push and HTTP 103 have different purposes, we should probably have both in Caddy.

Early Hints is still technically expeirmental, so we'll document this feature in Caddy as experimental as well. That means it is subject to change and/or removal.

According to @dunglas at https://dunglas.fr/2021/02/using-the-103-early-hints-status-code-in-go-applications/, it's very simple to perform Early Hints in Go (Go 1.19+ required):

```
w.Header().Add("Link", "</style.css>; rel=preload; as=style")
w.Header().Add("Link", "</script.js>; rel=preload; as=script")
w.WriteHeader(http.StatusEarlyHints)
```

then you proceed to write your full response.

The question is, how do we best expose this for the user to configure?

Essentially it's just one or more HTTP headers, so the `header` handler seems fitting. The only trick is that we have to immediately write the response with a 103 status. `header` can't flush a response, but the `static_response` handler can:

```
{
    "handler": "static_response",
    "status_code": 103,
    "headers": {
        "Link": [
            "</style.css>; rel=preload; as=style",
            "</script.js>; rel=preload; as=script"
        ]
    }
}
```

All we have to do is tweak the `static_response` handler to not terminate the handler chain after it writes a 103 response.

Technically, HTTP/1.1 clients can support 103 Early Hints, but it could also break them if they're not expecting a 1xx response without an Expect header. The RFC recommends only sending 103s for HTTP/2 (or HTTP/3) connections, as those protocols do not depend on headers to determine the end of the response.

Fastly has a page for testing your client's HTTP Early Hints support: https://early-hints.fastlylabs.com/

I will upgrade the `protocol` matcher to support specific HTTP versions, such as `http/2` or `http/2+` to match only HTTP/2 or HTTP/2 and newer, respectively.

Apache's implementation of Early Hints has the option of also pushing the hinted resources, but as I said earlier, I think the two have different purposes and times to use them, so I don't think I'll combine push and hint, unless we see convincing evidence that it is actually beneficial.

As for the Caddyfile, the `respond` directive should be able to work fine in conjunction with the `header` directive. (The `respond` directive does not currently support setting headers, because that is what the `header` directive does, and it amounts to the exact same number of lines of config -- actually more if used in `respond` since you have the closing brace on a separate line.)

mholt:
This adds support for early hints in the static_response handler.

It implements my initial proposal. If accepted, it closes #5005.

A practical Caddyfile example, assuming that the endpoints behind `/heavy-pages/` take a while to load on the backend:

```
@hint {
	protocol http2+
	path     /heavy-pages/*
}
route @hint {
	header  Link "</style.css>; rel=preload; as=style"
	header  Link "</script.js>; rel=preload; as=script"
	respond 103
}
reverse_proxy ...
```

(The HTTP/2 requirement is to try to protect old, incompatible clients. It is optional if your clients are modern or you don't care.)

Early hints are best used when the final response may not be ready for a while (a "while" being, perhaps, 2-3 RTTs, so still on the order of ms).

I had considered an implementation that automatically dispatches early hints from our `file_server` since we can parse HTML and get an idea of the subresources before sending the whole document to the client. However, by the time we've parsed the HTML, we might as well be streaming it to the client anyway, and they can read the `<head>` of the HTML document and start requesting its subresources. Even if we cached the result of parsing the HTML, it's probably still faster to just send it to the client without hints.

Anyway, this seems simple enough and I'm inclined to merge it in after a few more tests.

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -1121,6 +1121,22 @@ func (m MatchProtocol) Match(r *http.Request) bool {
 		return r.TLS != nil
 	case "http":
 		return r.TLS == nil
+	case "http/1.0":
+		return r.ProtoMajor == 1 && r.ProtoMinor == 0
+	case "http/1.0+":
+		return r.ProtoAtLeast(1, 0)
+	case "http/1.1":
+		return r.ProtoMajor == 1 && r.ProtoMinor == 1
+	case "http/1.1+":
+		return r.ProtoAtLeast(1, 1)
+	case "http/2":
+		return r.ProtoMajor == 2
+	case "http/2+":
+		return r.ProtoAtLeast(2, 0)
+	case "http/3":
+		return r.ProtoMajor == 3
+	case "http/3+":
+		return r.ProtoAtLeast(3, 0)
 	}
 	return false
 }
--- a/modules/caddyhttp/push/handler.go
+++ b/modules/caddyhttp/push/handler.go
@@ -29,10 +29,24 @@ func init() {
 	caddy.RegisterModule(Handler{})
 }
 
-// Handler is a middleware for manipulating the request body.
+// Handler is a middleware for HTTP/2 server push. Note that
+// HTTP/2 server push has been deprecated by some clients and
+// its use is discouraged unless you can accurately predict
+// which resources actually need to be pushed to the client;
+// it can be difficult to know what the client already has
+// cached. Pushing unnecessary resources results in worse
+// performance. Consider using HTTP 103 Early Hints instead.
+//
+// This handler supports pushing from Link headers; in other
+// words, if the eventual response has Link headers, this
+// handler will push the resources indicated by those headers,
+// even without specifying any resources in its config.
 type Handler struct {
-	Resources []Resource    `json:"resources,omitempty"`
-	Headers   *HeaderConfig `json:"headers,omitempty"`
+	// The resources to push.
+	Resources []Resource `json:"resources,omitempty"`
+
+	// Headers to modify for the push requests.
+	Headers *HeaderConfig `json:"headers,omitempty"`
 
 	logger *zap.Logger
 }
--- a/modules/caddyhttp/responsewriter.go
+++ b/modules/caddyhttp/responsewriter.go
@@ -111,15 +111,15 @@ type responseRecorder struct {
 //
 // Proper usage of a recorder looks like this:
 //
-//     rec := caddyhttp.NewResponseRecorder(w, buf, shouldBuffer)
-//     err := next.ServeHTTP(rec, req)
-//     if err != nil {
-//         return err
-//     }
-//     if !rec.Buffered() {
-//         return nil
-//     }
-//     // process the buffered response here
+//	rec := caddyhttp.NewResponseRecorder(w, buf, shouldBuffer)
+//	err := next.ServeHTTP(rec, req)
+//	if err != nil {
+//	    return err
+//	}
+//	if !rec.Buffered() {
+//	    return nil
+//	}
+//	// process the buffered response here
 //
 // The header map is not buffered; i.e. the ResponseRecorder's Header()
 // method returns the same header map of the underlying ResponseWriter.
@@ -129,7 +129,7 @@ type responseRecorder struct {
 // Once you are ready to write the response, there are two ways you can
 // do it. The easier way is to have the recorder do it:
 //
-//     rec.WriteResponse()
+//	rec.WriteResponse()
 //
 // This writes the recorded response headers as well as the buffered body.
 // Or, you may wish to do it yourself, especially if you manipulated the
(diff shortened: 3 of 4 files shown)


otbutz:
The requirement to explicitly protect it with `protocol http2+` sounds like a potential footgun? Almost every client supports HTTP/2, so it would be hard to detect if someone forgets to add it as only half-assed HTTP/1.x clients would be affected.

otbutz:
Could we safely drop 103 responses on a HTTP/1.x connection instead?

Citing the RFC:

>    Therefore, a server might refrain from sending 103 (Early Hints)
>   responses over HTTP/1.1 unless the client is known to handle
>   informational responses correctly.

As we have no idea about the capabilities of the client, we should drop them.

mholt:
> The requirement to explicitly protect it with protocol http2+ sounds like a potential footgun? Almost every client supports HTTP/2, so it would be hard to detect if someone forgets to add it as only half-assed HTTP/1.x clients would be affected.

Good point, I personally think that part is optional, I'm just showing a "best practice" that I read from the Internet :upside_down_face: (IIRC it's even recommended in the spec).

So yeah, protocol enforcement is optional if you don't care about or have old clients. Making it as simple as:

```
route /heavy-pages/* {
	header  Link "</style.css>; rel=preload; as=style"
	header  Link "</script.js>; rel=preload; as=script"
	respond 103
}
reverse_proxy ...
```

> Could we safely drop 103 responses on a HTTP/1.x connection instead?

We could, but what if the HTTP/1.x client can handle informational responses? Then there's no good way to make it happen. (I could imagine cases where modern clients use HTTP/1.1, for example over plaintext HTTP on an internal network.)

otbutz:
> We could, but what if the HTTP/1.x client can handle informational responses?

HTTP 103 is purely informative. You just don't get the performance benefit, so to speak.

Or as the spec puts it:

>    Aside from performance optimizations, such evaluation of the 103
>    (Early Hints) response's header fields **MUST NOT affect how the final
>    response is processed**.  A client MUST NOT interpret the 103 (Early
>    Hints) response header fields as if they applied to the informational
>    response itself (e.g., as metadata about the 103 (Early Hints)
>    response).

I would wager that the proportion of clients that would correctly process early hints without supporting HTTP/2  is very small. Breaking old quirky HTTP/1.x implementations is *far* more likely.

> I could imagine cases where modern clients use HTTP/1.1, for example over plaintext HTTP **on an internal network**.

An internal network would have low latency which doesn't really benefit from early hints.

We could add an i-know-what-im-doing-just-enable-it option though :)

otbutz:
It's even worse if caddy is acting as a reverse proxy. The backend might have some logic in place to avoid sending early hints if it detects a HTTP/1.x client. Since the backend communication may be using HTTP/2, this kind of detection is broken. We would need some kind of `X-Forwarded-*` header to provide the backend with the protocol level of the client. I fear that such a header doesn't exist yet.

To paraphrase it: any backend which uses HTTP 103 responses is likely to break older clients if we blindly pipe through early hints

mholt:
> Breaking old quirky HTTP/1.x implementations is far more likely.

I look forward to finding out if that's the case -- we'll try it in a beta and can adjust if needed.

> An internal network would have low latency which doesn't really benefit from early hints.

Network latency is not the issue (if it were, you wouldn't want Early Hints which requires 1 extra RTT -- you'd want Push) - the problem is CPU latency. Backends that take a lot of time generating the response. For example a long DB query or similar.

> It's even worse if caddy is acting as a reverse proxy. The backend might have some logic in place to avoid sending early hints if it detects a HTTP/1.x client. Since the backend communication may be using HTTP/2, this kind of detection is broken.

Good point, and I would also be interested in learning how many of those are out there.

I like the way you think! I also like the idea of finding out how many broken clients are out there and how often this actually appears in the wild.

Since there is a very simple way to disable the "hopeful" behavior of 103 Early Hints (by using the protocol matcher with `http2+`), I think it's probably fine to roll this out, especially in a pre-release for now.

Your concerns and points are very valid and correct. I am mainly curious how widespread that behavior is, and beta is the time to find out!

francislavoie (approved):

PR #5007, opened 2022-09-01
dist: deb package manpages and bash completion scripts

mohammed90:
I've tested that on upgrade, the newer man pages overwrite the older ones.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,7 +14,11 @@ before:
     # run `go mod tidy`. The `/bin/sh -c '...'` is because goreleaser can't find cd in PATH without shell invocation.
     - /bin/sh -c 'cd ./caddy-build && go mod tidy'
     - git clone --depth 1 https://github.com/caddyserver/dist caddy-dist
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -14,7 +14,11 @@ before:
     # run `go mod tidy`. The `/bin/sh -c '...'` is because goreleaser can't find cd in PATH without shell invocation.
     - /bin/sh -c 'cd ./caddy-build && go mod tidy'
     - git clone --depth 1 https://github.com/caddyserver/dist caddy-dist
+    - mkdir -p caddy-dist/man
```
mholt, line 17: I believe `manpage` creates the directory if it does not exist, right?

mholt (approved):
This is super great. Thank you @mohammed90 !!

francislavoie (approved):
LGTM

.goreleaser.yml
mohammed90: Yep, it slipped my mind, but the code confirms it.

PR #5018, opened 2022-09-05
Drop requirement for filesystems to implement fs.StatFS

hairyhenderson:
According to a brief conversation in [Slack](https://caddyserver.slack.com/archives/C4DLAAP36/p1662419130255019?thread_ts=1662418198.184659&cid=C4DLAAP36), the `fs.StatFS` requirement was chosen for performance reasons.

A filesystem that implements `fs.StatFS` doesn't necessarily perform any faster than a `fs.FS` that requires calling `Open` first. For example, some of the filesystems in [`hairyhenderson/go-fsimpl`](https://github.com/hairyhenderson/go-fsimpl) have a lazy `Open`, and so there's no effective performance hit for [`fs.Stat`](https://pkg.go.dev/io/fs#Stat)'s fallback (`Open` followed by `Read` on the resulting `fs.File`).

I've changed references to `fs.StatFS`, and replaced calls to `Stat` with calls to `fs.Stat` instead.

I think this will make adoption much easier!

### DIFF
--- a/modules/caddyhttp/fileserver/browse.go
+++ b/modules/caddyhttp/fileserver/browse.go
@@ -210,7 +210,7 @@ func (fsrv *FileServer) isSymlinkTargetDir(f fs.FileInfo, root, urlPath string) 
 		return false
 	}
 	target := caddyhttp.SanitizedPathJoin(root, path.Join(urlPath, f.Name()))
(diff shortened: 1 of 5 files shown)


mholt (approved):
LGTM, I'm convinced. Thanks Dave!

PR #5027, opened 2022-09-08
supplychain: publish signing cert, sbom, and signatures of sbom

mohammed90:


mholt (approved):
Thanks for working on this. As usual, I don't really understand it :sweat_smile: -- is this good to go?

(I'm not sure I get the "defaults to documents:" comments)

mohammed90:
> Thanks for working on this. As usual, I don't really understand it 😅 -- is this good to go?

Yes. I've made a few more experiments [here](https://github.com/gactions-playground/merry-go-round/releases/tag/v0.0.4). N.B. the SBOM (Software Bill of Material) files of mac OS binaries will say `darwin` instead of `mac` because replacements aren't available in that part of GoReleaser (or I'm missing how to do it).

> (I'm not sure I get the "defaults to documents:" comments)

It's the assumed resulting file name. I left it as a comment so it's easier for us later to figure them out without having to re-look it up on GoReleaser docs.

mholt:
Gotcha. Thanks again!

PR #5030, opened 2022-09-09
caddyhttp: Explicitly disallow multiple regexp matchers

### LINKED ISSUE #5028: Using multiple path_regexp matchers does not error
Caddy version: v2.6.0-beta.3
GOOS: linux
GOARCH: amd64

I found some interesting behavior where specifying multiple `path_regexp` matchers in a named matcher will not error when the config file is parsed. Instead, the last `path_regexp` matcher will be used and the previous matchers will silently be discarded.

Example Caddyfile:

```
http://localhost:9000 {
	log {
		output stdout
		level INFO
	}

	@dockerAPI {
		method GET
		path_regexp ^/v\d\.\d+/*
		path_regexp ^/asdf/*
	}

	handle @dockerAPI {
		reverse_proxy unix//var/run/docker.sock
	}
	handle {
		respond 427
	}
}
```

In this example, requests that begin with `/v1.41/` do not match `@dockerAPI`, but requests that begin with `/asdf/` do. I think an error should be returned when parsing the config file.

As a side note, I discovered this while trying to use Caddy to limit access to a Docker API socket, I only want certain API endpoints to be allowed and passed through by Caddy. It's difficult to do such a thing without the ability to use multiple `path_regexp` matchers, is there a workaround for this or should I be using a 3rd party module/a separate tool entirely?

francislavoie:
Fix #5028

Since the matchers would overwrite eachother, we should error out to tell the user their config doesn't make sense.

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -1001,6 +1001,12 @@ func (m *MatchHeaderRE) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 			val = second
 		}
 
+		// If there's already a pattern for this field
+		// then we would end up overwriting the old one
+		if (*m)[field] != nil {
+			return d.Errf("header_regexp matcher can only be used once per named matcher, per header field: %s", field)
+		}
+
 		(*m)[field] = &MatchRegexp{Pattern: val, Name: name}
 
 		if d.NextBlock(0) {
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/matchers.go
```diff
@@ -1469,6 +1475,13 @@ func (mre *MatchRegexp) Match(input string, repl *caddy.Replacer) bool {
 // UnmarshalCaddyfile implements caddyfile.Unmarshaler.
 func (mre *MatchRegexp) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	for d.Next() {
+		// If this is the second iteration of the loop
+		// then there's more than one path_regexp matcher
+		// and we would end up overwriting the old one
+		if mre.Pattern != "" {
+			return d.Err("path_regexp matcher can only be used once per named matcher")
```
mholt, line 1482:
```suggestion
			return d.Err("regular expression can only be used once per named matcher")
```

This is used for more than just the path matcher IIRC

mholt (approved):
Thanks! LGTM, just a minor nit. (Correct me if I'm wrong)

modules/caddyhttp/matchers.go
francislavoie: It's not used for header_regexp so I don't know what else 🤔

mholt: True, but this is an embeddable type, it just so happens that path is the only one that does it for right now.

PR #5043, opened 2022-09-16
caddyhttp: Honor grace period in background

mholt:
This should make config reloads much faster. It avoids blocking from grace periods during config reloads by doing server shutdown in a goroutine. (We never returned any shutdown errors to the caller anyway.)

However, frequent config reloads with long grace periods can build up goroutines. This might be problematic if most/all of these are true:

- configs are being reloaded faster than the servers can shut down, for a sustained period of time,
- clients are slow to finish their requests, or requests are hijacked (WebSockets), and
- server resources are limited,

because obviously this will allow the next config to come in before the old one is out.

That might be OK, and it's probably just a matter of tuning to your actual requirements. If you meet the criteria above, shorten your grace period or do less frequent config changes. Otherwise I don't expect this behavior to be a problem.

Special case: if the process is exiting, the HTTP app's `Stop()` still blocks until the servers have shut down completely. This avoids an ungraceful shutdown by the process exiting too early.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -31,6 +31,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/caddyserver/caddy/v2/notify"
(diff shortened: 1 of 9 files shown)


mholt:
The failing test case on IBM is a known issue with the integration tests. That CI server is a donated piece of dedicated hardware (which we are grateful for), and as such, the tests run so fast that I think it exposes a race condition or some non-determinism in our tests. And with this PR, I guess eliminating the blocking on config changes makes the test basically unusable at this point.

I will probably still merge this PR and then either temporarily disable the integration tests until I can rewrite them, or spend considerable time to pinpoint the problem and patch it...

mohammed90:
> Edit: Huh, that was easier than I thought... just set a grace period to one nanosecond on the integration test configs and it appears to pass more consistently now. (I should rewrite the integration tests when I get a chance.)

This makes me happy! 😁

PR #5045, opened 2022-09-17
Fix tests

mholt:
I can only reproduce the bug on IBM's s390x, so this PR is for a branch that I'm just pushing a bunch of attempts to. Don't mind me...

Bug seems to appear 100% of the time with #5043.


mholt:
Hm, ok. I set the grace period to 1 (one nanosecond) on all the integration test cases and it passed first time. That seems to ~fix it~ cover it up :+1:

PR #5049, opened 2022-09-17
response recorder save status code in all cases

WeidiDeng:


modules/caddyhttp/responsewriter.go
```diff
@@ -170,9 +170,11 @@ func (rr *responseRecorder) WriteHeader(statusCode int) {
 		return
 	}
 
+	// save statusCode in case http middleware upgrading websocket
+	// connections by manually setting headers and writing status 101
+	rr.statusCode = statusCode
```
mholt, line 175: Do we really want to save 1xx codes though? Those aren't the _actual_ or final status code...

WeidiDeng: Not really, but nice to have, since normally handler will write more status codes, it will be overwritten anyway. And in this case, to match previous caddy log output, 101 status code for websocket upgrade which is final. It's up to you.

mholt: Ok, good point; I'm convinced. Thanks for the patch!

mholt (approved):

PR #5051, opened 2022-09-19
caddyhttp: Skip inserting HTTP->HTTPS redir if catch-all for both exist

### LINKED ISSUE #4829: Weird bug with default server directive that has both HTTP and HTTPS?
With the following Caddyfile:
```
http:// {
        respond "DEFAULT"
}
```
Caddy will respond to a HTTP request on port 80 over direct IP as expected with "DEFAULT".
```
> GET / HTTP/1.1
> Host: 192.168.1.100
> User-Agent: curl/7.79.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: Caddy
< Date: Tue, 07 Jun 2022 23:47:19 GMT
< Content-Length: 7
<
DEFAULT
```

However, if we try adding a https:// there like so:
```
http:// https:// {
        respond "DEFAULT"
}
```
suddenly the HTTP port starts redirecting to HTTPS.
```
> GET / HTTP/1.1
> Host: 192.168.1.100
> User-Agent: curl/7.79.1
> Accept: */*
>
< HTTP/1.1 308 Permanent Redirect
< Connection: close
< Location: https://192.168.1.100/
< Server: Caddy
< Date: Tue, 07 Jun 2022 23:46:06 GMT
< Content-Length: 0
<
```
This is most unexpected, given Caddy is normally supposed to support mixed HTTP and HTTPS site blocks, like this:
```
http://test.example.com test.example.com {
        respond "DOMAIN"
}
```
In the above example, you can normally access test.example.com over HTTP as expected (over HTTPS too ofc).

What's even more surprising, is that when we modify the second Caddyfile snippet to be this:
```
test.example.com {
        # must be https
        respond "TEST"
}

http:// https:// {
        respond "DEFAULT"
}
```
**Suddenly HTTP port responds with "DEFAULT" as expected again!**

Looking further at the default server example though, it seems that its HTTPS port does not work (no matter if tried by IP or dummy domain):
```
>curl -k -v https://192.168.1.100
*   Trying 192.168.1.100:443...
* Connected to 192.168.1.100 (192.168.1.100) port 443 (#0)
* schannel: disabled automatic use of client certificate
* schannel: using IP address, SNI is not supported by OS.
* schannel: ALPN, offering http/1.1
* schannel: next InitializeSecurityContext failed: SEC_E_ILLEGAL_MESSAGE (0x80090326) - This error usually occurs when a fatal SSL/TLS alert is received (e.g. handshake failed). More detail may be available in the Windows System event log.
* Closing connection 0
curl: (35) schannel: next InitializeSecurityContext failed: SEC_E_ILLEGAL_MESSAGE (0x80090326) - This error usually occurs when a fatal SSL/TLS alert is received (e.g. handshake failed). More detail may be available in the Windows System event log.
```
If I add `tls internal` into the "DEFAULT" block, nothing changes. If I add `tls <path>.pem <path>.pem` manually with a path to the certificate into the section, Caddy fails to start:
```
run: adapting config using caddyfile: server listening on [:80] is HTTP, but attempts to configure TLS connection policies
```

---

# **So to summarize, it seems there are two issues**:
1. Inconsistent behavior of the "default" (no-hostname) site block for both HTTP and HTTPS based on existence of **another** unrelated HTTPS site block (if "default" block is the only one, HTTP is redirected to HTTPS for no reason?).
2. You are supposed to be able to create a site block that can be served over HTTP and HTTPS both, but you seemingly cannot then use `tls` directive inside to set the "internal" or custom certificate paths, it only works with default encryption that specifies no custom options. Expected behavior would be that whatever is set in `tls` option would simply be ignored for HTTP connections...

It is worth to note that I was previously successfully using both Apache2 and nginx with such configuration that is basically the default if you enable SSL there, that is the same default server block is also served over HTTPS with a self-signed certificate, which works both over IP or any unrecognized domain, thus titled the "default".

### Workaround attempt

I tried this:
```
(default) {
        respond "DEFAULT"
}

http:// {
        import default
}

https:// {
        # I could not access this over HTTPS with neither TLS options or lack thereof
        #tls internal
        # adding the line below (instead of the above ofc) makes the Caddy server start with no error, but stop responding to ANY requests, both HTTP or HTTPS (???)
        #tls /etc/ssl/certs/ssl-cert-snakeoil.pem /etc/ssl/private/ssl-cert-snakeoil.key
        import default
}
```
Result: I still cannot load the server via HTTP, I still need to add a random HTTPS site block in order to be able to reach the default HTTP site.

---

Now of course I could simply accept that I will not get an unsecure default unnamed HTTPS endpoint with Caddy that I could by default with Apache2 and nginx. But I think my attempts still bring up some logic errors in Caddy config processing that shouldn't be behaving the way they behave regardless.

---

Caddy version: v2.5.1 h1:bAWwslD1jNeCzDa+jDCNwb8M3UJ2tPa8UZFFzPVmGKs=

francislavoie:
Fix #4829 /cc @whitestrake

After thinking about it a bit, the fix seemed like it would probably be pretty easy.

I flipped an `if` to reduce indentation (`continue` early) and wrapped the insert with a simple condition that checks if we have any actual domains with certs in the config, and if not then we can probably assume we have a catch-all HTTPS site if we got that far.

Confirmed with some `fmt.Printf` that the routes still have the route added at the start if there is any domain configured, and it doesn't if there are no domains. The catch-all redirect at the end is always added, which would only run when the user doesn't define their own catch-all HTTP route.

### DIFF
--- a/modules/caddyhttp/autohttps.go
+++ b/modules/caddyhttp/autohttps.go
@@ -378,19 +378,29 @@ redirServersLoop:
 		// we'll create a new server for all the listener addresses
 		// that are unused and serve the remaining redirects from it
 		for _, srv := range app.Servers {
-			if srv.hasListenerAddress(redirServerAddr) {
-				// find the index of the route after the last route with a host
-				// matcher, then insert the redirects there, but before any
-				// user-defined catch-all routes
-				// see https://github.com/caddyserver/caddy/issues/3212
-				insertIndex := srv.findLastRouteWithHostMatcher()
-				srv.Routes = append(srv.Routes[:insertIndex], append(routes, srv.Routes[insertIndex:]...)...)
-
-				// append our catch-all route in case the user didn't define their own
-				srv.Routes = appendCatchAll(srv.Routes)
-
-				continue redirServersLoop
+			// only look at servers which listen on an address which
+			// we want to add redirects to
+			if !srv.hasListenerAddress(redirServerAddr) {
+				continue
 			}
+
+			// find the index of the route after the last route with a host
+			// matcher, then insert the redirects there, but before any
+			// user-defined catch-all routes
+			// see https://github.com/caddyserver/caddy/issues/3212
+			insertIndex := srv.findLastRouteWithHostMatcher()
+
+			// add the redirects at the insert index, except for when
+			// we have a catch-all for HTTPS, in which case the user's
+			// defined catch-all should take precedence. See #4829
+			if len(uniqueDomainsForCerts) != 0 {
+				srv.Routes = append(srv.Routes[:insertIndex], append(routes, srv.Routes[insertIndex:]...)...)
+			}
+
+			// append our catch-all route in case the user didn't define their own
+			srv.Routes = appendCatchAll(srv.Routes)
+
+			continue redirServersLoop
 		}
 
 		// no server with this listener address exists;


mholt (approved):
Thanks for the elegant patch, Francis. I feel much better about this, hopefully it works well! :sweat_smile:

PR #5054, opened 2022-09-20
httpcaddyfile: Fix `protocols` global option parsing

### LINKED ISSUE #5050: How about more mature deprecation flow?
I used caddyserver v2.4.3 for about a year, then I've upgraded to v2.5.2 and noticed that server just can't use old config because of #4148.

I've keep an eye on your project since v0.10 or so, and I was pretty sure that now it's a serious project that could be used in the production without any doubts. I even recommended it to other people. But how could you deprecate and delete a setting within 1 year? Since I did not update between v2.4.3 and v2.5.2, I did not even have a chance to see the warning (if there was one).

https://caddy.community/t/error-module-not-registered-caddy-logging-encoders-single-field-encountered-after-upgrading-to-2-5-0/15763
> Yes, it was removed. **It was deprecated for many versions.**

Hey, that's many versions for you (developers), but I as a user expect from a newer v2.x.x to be fully compatible with an older v2.x.x.

Please consider changing your way from 'move fast and break things' to something more suitable for a web server.

### LINKED ISSUE #5053: Parsing error with new `protocols` directive
The new `protocols` subdirective in `servers` is only excepted when it is the last directive in the `servers` block. Is that intended?

Example Caddyfile:

```caddy-d
{
	servers :443 {
		protocols h1 h2 h3
		strict_sni_host insecure_off
	}
}
```

or

```caddy-d
{
	servers :443 {
		protocols h1 h2 h3
		log_credentials
	}
}
```

Produces errors:

```plain
Error: adapting config using caddyfile: parsing caddyfile tokens for 'servers': Caddyfile:4 - Error during parsing: Wrong argument count or unexpected line ending after 'strict_sni_host'
```

```plain
Error: adapting config using caddyfile: parsing caddyfile tokens for 'servers': Caddyfile:4 - Error during parsing: Wrong argument count or unexpected line ending after 'log_credentials'
```

This works though:

```caddy-d
{
	servers :443 {
		log_credentials
		protocols h1 h2 h3
	}
}
```

Related pull request: https://github.com/caddyserver/caddy/pull/4707

francislavoie:
Fix #5053

When checking for a block, the current nesting must be used, otherwise it returns the wrong thing.

### DIFF
--- a/admin_test.go
+++ b/admin_test.go
@@ -161,7 +161,7 @@ func (fooModule) Stop() error  { return nil }
 func TestETags(t *testing.T) {
 	RegisterModule(fooModule{})
 
-	if err := Load([]byte(`{"apps": {"foo": {"strField": "abc", "intField": 0}}}`), true); err != nil {
+	if err := Load([]byte(`{"admin": {"listen": "localhost:2999"}, "apps": {"foo": {"strField": "abc", "intField": 0}}}`), true); err != nil {
 		t.Fatalf("loading: %s", err)
 	}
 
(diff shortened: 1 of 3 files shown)


mholt (approved):
Oops, I should know better! Thanks 😅

PR #5063, opened 2022-09-21
Reuse unix sockets (UDS) and don't try to serve HTTP/3 over UDS

### LINKED ISSUE #5056: Config reload fails when admin endpoint is a socket
I've been using this Caddyfile global configuration, which was working at e289ba6187ddd50afc86feb85246a14298673a69 but is no longer working in 2.6.0 (Caddy's official package for Debian).
```
{
	admin unix//run/caddy/admin.sock
}
```
The goal is to support zero downtime config reloads without exposing Caddy's config to any process on the system that might send some json to 127.0.0.1.  Partly in case some other service gets compromised, and partly because we have human SSH user accounts on the system which don't have sysadmin (sudo) access and should not be able to reconfigure the main web server.

`/run/caddy` has mode `0750` and `caddy:caddy` ownership.  

Log:
```
systemd[1]: Reloading Caddy.
caddy[3372028]: {"level":"info","ts":1663700361.6707227,"msg":"using provided configuration","config_file":"/etc/caddy/Caddyfile","config_adapter":""}
caddy[3371872]: {"level":"info","ts":1663700361.6808317,"logger":"admin.api","msg":"received request","method":"POST","host":"","uri":"/load","remote_ip":"@","remote_port":"","headers":{"Accept-Encoding":["gzip"],"Cache-Control":["must-revalidate"],"Content-Length":["20327"],"Content-Type":["application/json"],"User-Agent":["Go-http-client/1.1"]}}
caddy[3371872]: {"level":"error","ts":1663700361.6862683,"logger":"admin.api","msg":"request error","error":"loading config: loading new config: starting caddy administration endpoint: listen unix /run/caddy/admin.sock: bind: address already in use","status_code":400}
caddy[3372028]: Error: sending configuration to instance: caddy responded with error: HTTP 400: {"error":"loading config: loading new config: starting caddy administration endpoint: listen unix /run/caddy/admin.sock: bind: address already in use"}
systemd[1]: caddy.service: Control process exited, code=exited, status=1/FAILURE
```

Relatedly I also needed this addition to the packaged `/lib/systemd/system/caddy.service` to make Caddy start reliably with the socket admin endpoint (loaded via systemd drop-in config):
```
[Service]
# bones_was_here: 2.5.2 fails to start if the admin endpoint socket still exists!
ExecStartPre=rm -f /run/caddy/admin.sock
```

mholt:
This PR hopefully fixes #5056 - any use of Unix domain sockets (UDS) cause errors now because somehow I missed that SO_REUSEPORT doesn't apply to UDS, even though it's a socket too.

My testing didn't catch this before because I didn't attempt to do a config reload with unix sockets specifically.

---

To test this, I created a simple server:

```
localhost

bind unix//home/matt/Desktop/test.sock
respond "Hello world!"
```

(I also enabled `debug` mode for my tests, but not required.)

I also added `time.Sleep(1 * time.Second)` to the `static_response` handler (`respond` directive) between writing headers and the body, to simulate a slow request-response. (I also repeated the tests without this sleep successfully.)

I started the server, then I ran this bash loop:

```
for i in {1..100000}; do
  curl -v --unix-socket ~/Desktop/test.sock "https://localhost/"
  sleep 0.001
done
```

to get "Hello world".

At the same time I ran this loop:

```
for i in {1..10000}; do
  curl -v "http://localhost:2019/load" -H "Content-Type: text/caddyfile" -H "Cache-Control: must-revalidate" --data-binary @Caddyfile
  sleep 0.01
done
```

to rapidly change the config. This should put the reload code (and unix socket reuse) through its paces.

I observed 0 errors, 0 failed requests, and 0 resource leaks (I used `localhost:2019/debug/pprof` to watch goroutines and I also used the system monitor to measure memory use).

---

For HTTP/3 support on UDS, we having an upstream discussion: https://github.com/lucas-clemente/quic-go/issues/3560

I don't expect the tests to pass for all platforms yet. I just found out Go has a new `unix` build tag that I'm trying out.


mholt:
Ahhhh dangit, unix build tag was only added in Go 1.19. :cry:

Issue #5065, opened 2022-09-22
Caddy doesn't show its version in CLI (FreeBSD 13.1, ports)


s0xDk:
Just updated Caddy on FreeBSD 13.1 from 2.6.0 to 2.6.1 via ports, and when I execute `caddy version` it shows blank line:

```
~$ caddy version

~$
```

Otherwise, everything seems to work fine.

s0xDk:
```
~$ caddy build-info
go	go1.19.1
path	github.com/caddyserver/caddy/v2/cmd/caddy
mod	github.com/caddyserver/caddy/v2	(devel)
dep	filippo.io/edwards25519	v1.0.0-rc.1	h1:m0VOOB23frXZvAOK44usCgLWvtsxIoMCTBGJZlpmGfU=
dep	github.com/AndreasBriese/bbloom	v0.0.0-20190825152654-46b345b51c96	h1:cTp8I5+VIoKjsnZuH8vjyaysT/ses3EvZeaV/1UkF2M=
dep	github.com/BurntSushi/toml	v1.2.0	h1:Rt8g24XnyGTyglgET/PRUNlrUeu9F5L+7FilkXfZgs0=
dep	github.com/Masterminds/goutils	v1.1.1	h1:5nUrii3FMTL5diU80unEVvNevw1nH4+ZV4DSLVJLSYI=
dep	github.com/Masterminds/semver/v3	v3.1.1	h1:hLg3sBzpNErnxhQtUy/mmLR2I9foDujNK030IGemrRc=
dep	github.com/Masterminds/sprig/v3	v3.2.2	h1:17jRggJu518dr3QaafizSXOjKYp94wKfABxUmyxvxX8=
dep	github.com/alecthomas/chroma	v0.10.0	h1:7XDcGkCQopCNKjZHfYrNLraA+M7e0fMiJ/Mfikbfjek=
dep	github.com/antlr/antlr4/runtime/Go/antlr	v0.0.0-20220418222510-f25a4f6275ed	h1:ue9pVfIcP+QMEjfgo/Ez4ZjNZfonGgR6NgjMaJMu1Cg=
dep	github.com/aryann/difflib	v0.0.0-20210328193216-ff5ff6dc229b	h1:uUXgbcPDK3KpW29o4iy7GtuappbWT0l5NaMo9H9pJDw=
[...]

PR #5072, opened 2022-09-23
Allow version to be set manually

assistcontrol:
When Caddy is built from a release tarball (as downloaded from GitHub), `caddy version` returns an empty string. This causes confusion for downstream packagers.

With this commit, VersionString can be set with eg.
  go build (...) -ldflags '-X (...).VersionString=v1.2.3'
Then the short form version will be "v1.2.3", and the full version string will begin with "v1.2.3 ".


assistcontrol:
See #5065 for some discussion on this.

mholt:
Thanks! I just pushed some revisions that prefer the embedded version if available, but still prepends the CustomVersion (renamed from VersionString) if set.

Let me know if that's OK with you, and if so I'll merge this and maybe release 2.6.2 here pretty soon.

assistcontrol:
If I'm reading it correctly, `CustomVersion` is no longer prepended to `full` if `full` is non-empty.

It looks like it is still possible for `full` to wind up being empty.

```go
if full == "" {
	full = CustomVersion
} else if CustomVersion != "" {
	full = CustomVersion + " " + full
}
```

If `full = ""` and `CustomVersion = ""`, then `full` will be empty. Similarly, if `full = ""` and `CustomVersion="foo"`, then `full = "foo "` (with a trailing space).

Perhaps instead, default an empty `full` to `simple`, and prepend `CustomVersion`:

```go
	if simple == "" || simple == "(devel)" {
		simple = "unknown"
		if CustomVersion != "" {
			simple = CustomVersion
		} else {
			simple = "unknown"
		}
	}

	if full == "" {
		if CustomVersion == "" {
			full = simple
		} else {
			full = CustomVersion
		}
	} else {
		full = CustomVersion + " " + full
	}
```

assistcontrol:
The first `simple = "unknown"` is probably superfluous.

It may also be safe to assume that an empty `full` should just be `simple` which will be `CustomVersion`, because an empty `full` means an empty `simple`:

```go
	if simple == "" || simple == "(devel)" {
		if CustomVersion != "" {
			simple = CustomVersion
		} else {
			simple = "unknown"
		}
	}

	if full == "" {
		full = simple
	} else {
		full = CustomVersion + " " + full
	}
```

mholt:
(Can you do inline comments next time? That'll make code review easier.)

> If I'm reading it correctly, CustomVersion is no longer prepended to full if full is non-empty.

It is still prepended: https://github.com/caddyserver/caddy/pull/5072/files#diff-57caf2be698e920a2599b611efd1b33dea35fd79976cacdf3354ab3f55706515R932

> It looks like it is still possible for full to wind up being empty.

Yes, but only if there is absolutely no version information embedded and CustomVersion is empty, then full will be empty. This is probably OK, rather than covering up the fact with a non-empty value? I dunno.

> Similarly, if full = "" and CustomVersion="foo", then full = "foo " (with a trailing space).

Are you sure that's true? Not in my testing.

Let's sort out those understandings and then we can make more tweaks if needed.

mholt:
@assistcontrol Is there anything else before I go ahead and merge this in? Will probably do so in a day or two otherwise.

assistcontrol:
Sorry, IRL has been busy.

> > It looks like it is still possible for full to wind up being empty.
> 
> Yes, but only if there is absolutely no version information embedded and CustomVersion is empty, then full will be empty. This is probably OK, rather than covering up the fact with a non-empty value? I dunno.

All it takes for there to be absolutely no version information embedded is to download any of the tarballs on the GH releases page.

I guess the question is, when someone downloads a tarball from the GH release page and doesn't know to set `CustomVersion`, would you prefer that `caddy version` print nothing, or `unknown`? I have my preference ("unknown" is better than wondering whether the `caddy version` command itself is broken) but I'm behind you and pleased with your edits no matter which path you prefer here.

mholt (approved):
CI is stuck, so going ahead and merging. Thanks for your contribution @assistcontrol !

GitHub incident: https://www.githubstatus.com/incidents/smn1qtqvbsb6

Issue #5070, opened 2022-09-23
Add retries to http loader


yroc92:
# Add retries to the http loader

## Context
Intermittently, we have Caddy instances booting up on new servers seeing the `net/http: TLS handshake timeout` error with the http loader in use. These failed connections are likely due to the fact that the http call made in `httploader.go` is the first connection made to the host and as such, the TLS handshake is being established for the first time. Perhaps it is too slow or something gets dropped during the handshake around the globe 🌍.

## Industry-standard solution
In distributed systems, such outages are expected. Retries with exponential backoffs are often the go-to solution to this problem.

## Proposed implementation
Let's leverage hashicorp's [go-retryablehttp](https://github.com/hashicorp/go-retryablehttp) package, a light wrapper around `net/http`. We can use the defaults while modifying [Client.RetryMax](https://pkg.go.dev/github.com/hashicorp/go-retryablehttp#Client) to be configurable in Caddy's config:
```
{
  // ... other http loader config fields
  "max_retries": 10 // default to 0
}
```
I'm happy to submit a pr for this solution as soon as possible.

mholt:
Thanks for the proposal and PR Cory!

PR #5077, opened 2022-09-24
httploader: Implement retries

yroc92:
Resolves https://github.com/caddyserver/caddy/issues/5070

Screen Shot 2022-09-23 at 11 46 50 PM

### DIFF
--- a/caddyconfig/httploader.go
+++ b/caddyconfig/httploader.go
@@ -19,7 +19,6 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io"
-	"math"
 	"net/http"
 	"os"
 	"time"
@@ -46,9 +45,9 @@ type HTTPLoader struct {
 	// Maximum time allowed for a complete connection and request.
 	Timeout caddy.Duration `json:"timeout,omitempty"`
 
-	// Maximum number of retries for a successful call to URL.
-	// Does exponential back-offs (2...4...8...16... seconds) between retries. Defaults to 0.
-	MaxRetries int `json:"max_retries,omitempty"`
+	// The number of seconds to wait in between attempts to call `URL`.
+	// Defaults to 0, indicating no retries.
+	RetryDelay time.Duration `json:"retry_delay,omitempty"`
 
 	TLS *struct {
 		// Present this instance's managed remote identity credentials to the server.
@@ -99,7 +98,7 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
 		}
 	}
 
-	resp, err := doHttpCallWithRetries(client, req, hl.MaxRetries)
+	resp, err := doHttpCallWithRetries(ctx, client, req, hl.RetryDelay)
 	if err != nil {
 		return nil, err
 	}
@@ -124,29 +123,31 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
 	return result, nil
 }
 
-// Reattempts the http call using exponential back off when maxRetries is greater than 0.
-func doHttpCallWithRetries(client *http.Client, request *http.Request, maxRetries int) (*http.Response, error) {
-	var resp *http.Response
-	var err error
-
-	for retry := 0; retry <= maxRetries; retry++ {
-		if retry > 0 {
-			caddy.Log().Error(err.Error())
-			exponentialBackOff := math.Pow(2, float64(retry))
-			caddy.Log().Sugar().Infof("reattempting http load in %g seconds (%v retries remain)", exponentialBackOff, maxRetries-retry)
-			time.Sleep(time.Duration(exponentialBackOff) * time.Second)
-		}
-		resp, err = client.Do(request)
-		if err != nil {
-			err = fmt.Errorf("problem calling http loader url: %v", err)
-		} else if resp.StatusCode < 200 || resp.StatusCode > 499 {
-			err = fmt.Errorf("bad response status code from http loader url: %v", resp.StatusCode)
-		} else {
-			return resp, nil
-		}
+// Reattempts the http call, waiting `retryDelay` seconds in between attempts.
+func doHttpCallWithRetries(ctx caddy.Context, client *http.Client, request *http.Request, retryDelay time.Duration) (*http.Response, error) {
+	// make attempt
+	resp, err := client.Do(request)
+	if err != nil {
+		err = fmt.Errorf("problem calling http loader url: %v", err)
+	} else if resp.StatusCode < 200 || resp.StatusCode > 499 {
+		err = fmt.Errorf("bad response status code from http loader url: %v", resp.StatusCode)
+	} else {
+		return resp, nil
 	}
-
-	return resp, err
+	// skip retries if valid retryDelay not provided
+	if retryDelay <= 0 {
+		return resp, err
+	}
+	// log the error
+	caddy.Log().Error(err.Error())
+	// wait for the retry delay to lapse before reattempting,
+	// or return when context is done
+	select {
+	case <-time.After(time.Second * retryDelay):
+		return doHttpCallWithRetries(ctx, client, request, retryDelay)
+	case <-ctx.Done():
+		return nil, err
+	}
 }
 
 func (hl HTTPLoader) makeClient(ctx caddy.Context) (*http.Client, error) {
--- a/caddyconfig/httploader.go
+++ b/caddyconfig/httploader.go
@@ -94,7 +94,6 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
 		}
 	}
 
-	caddy.Log().Info("attempting to load config via http loader url")
 	resp, err := doHttpCallWithRetries(ctx, client, req)
 	if err != nil {
 		return nil, err
@@ -126,10 +125,8 @@ func attemptHttpCall(client *http.Client, request *http.Request) (*http.Response
 		return nil, fmt.Errorf("problem calling http loader url: %v", err)
 	} else if resp.StatusCode < 200 || resp.StatusCode > 499 {
 		return nil, fmt.Errorf("bad response status code from http loader url: %v", resp.StatusCode)
-	} else {
-		// the http call was successful
-		return resp, nil
 	}
+	return resp, nil
 }
 
 func doHttpCallWithRetries(ctx caddy.Context, client *http.Client, request *http.Request) (*http.Response, error) {
@@ -144,7 +141,6 @@ func doHttpCallWithRetries(ctx caddy.Context, client *http.Client, request *http
 			// wait 500ms before reattempting, or until context is done
 			select {
 			case <-time.After(time.Millisecond * 500):
-				continue
 			case <-ctx.Done():
 				return resp, ctx.Err()
 			}
--- a/caddyconfig/httploader.go
+++ b/caddyconfig/httploader.go
@@ -15,6 +15,7 @@
 package caddyconfig
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
@@ -24,6 +25,7 @@ import (
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
+	"github.com/hashicorp/go-retryablehttp"
 )
 
 func init() {
@@ -44,6 +46,9 @@ type HTTPLoader struct {
 
 	// Maximum time allowed for a complete connection and request.
 	Timeout caddy.Duration `json:"timeout,omitempty"`
+
+	// Maximum number of retries for a successful call to URL. Defaults to 0.
+	MaxRetries int `json:"max_retries,omitempty"`
 
 	TLS *struct {
 		// Present this instance's managed remote identity credentials to the server.
@@ -119,9 +124,25 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
 	return result, nil
 }
 
+func getRetryClient(ctx caddy.Context, maxRetries int, timeout caddy.Duration) (*http.Client, error) {
+	retryClient := retryablehttp.NewClient()
+	retryClient.RetryMax = maxRetries
+
+	retryClient.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
+		if err != nil || (resp.StatusCode < 200 || resp.StatusCode > 499) {
+			return true, err
+		}
+		return false, err
+	}
+	retryClient.HTTPClient.Timeout = time.Duration(timeout)
+
+	return retryClient.StandardClient(), nil
+}
+
 func (hl HTTPLoader) makeClient(ctx caddy.Context) (*http.Client, error) {
-	client := &http.Client{
-		Timeout: time.Duration(hl.Timeout),
+	client, err := getRetryClient(ctx, hl.MaxRetries, hl.Timeout)
+	if err != nil {
+		return nil, err
 	}
 
 	if hl.TLS != nil {
--- a/caddyconfig/httploader.go
+++ b/caddyconfig/httploader.go
@@ -15,17 +15,16 @@
 package caddyconfig
 
 import (
-	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"os"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
-	"github.com/hashicorp/go-retryablehttp"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


yroc92:
@mholt Not sure if the logging is correct? I think [I can replace their logger](https://pkg.go.dev/github.com/hashicorp/go-retryablehttp@v0.7.1#Client) like so:
```go
retryClient.Logger = ctx.Logger()
```
but I kept getting this error:
```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x102c544ec]

goroutine 1 [running]:
github.com/caddyserver/caddy/v2.(*Logging).Logger(0x14000499860, {0x0?, 0x0?})
        /Users/corycooper/workspace/github/caddy/logging.go:202 +0x2c
github.com/caddyserver/caddy/v2.Context.Logger({{0x103f5e4f0, 0x140003b4d40}, 0x14000499800, 0x14000054f00, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}}, ...)
        /Users/corycooper/workspace/github/caddy/context.go:478 +0xc0
github.com/caddyserver/caddy/v2/caddyconfig.getRetryClient({{0x103f5e4f0, 0x140003b4d40}, 0x14000499800, 0x14000054f00, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}}, ...)
        /Users/corycooper/workspace/github/caddy/caddyconfig/httploader.go:131 +0xbc
github.com/caddyserver/caddy/v2/caddyconfig.HTTPLoader.makeClient({{0x140006ddf28, 0x4}, {0x14000044ae0, 0x22}, 0x0, 0x0, 0x5, 0x0}, {{0x103f5e4f0, 0x140003b4d40}, ...})
        /Users/corycooper/workspace/github/caddy/caddyconfig/httploader.go:145 +0x7c
github.com/caddyserver/caddy/v2/caddyconfig.HTTPLoader.LoadConfig({{0x140006ddf28, 0x4}, {0x14000044ae0, 0x22}, 0x0, 0x0, 0x5, 0x0}, {{0x103f5e4f0, 0x140003b4d40}, ...})
        /Users/corycooper/workspace/github/caddy/caddyconfig/httploader.go:81 +0x17c
github.com/caddyserver/caddy/v2.finishSettingUp({{0x103f5e4f0, 0x140003b4d40}, 0x14000499800, 0x14000054f00, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}}, ...)
        /Users/corycooper/workspace/github/caddy/caddy.go:603 +0x4c0
github.com/caddyserver/caddy/v2.run(0x1400025e320?, 0x1)
        /Users/corycooper/workspace/github/caddy/caddy.go:525 +0x584
github.com/caddyserver/caddy/v2.unsyncedDecodeAndRun({0x1400025e1e0, 0x91, 0xa0}, 0x1)
        /Users/corycooper/workspace/github/caddy/caddy.go:337 +0xe4
github.com/caddyserver/caddy/v2.changeConfig({0x1037fc1a6, 0x4}, {0x1038092a9, 0x7}, {0x140002ec400, 0xe1, 0x200}, {0x0, 0x0}, 0x1)
        /Users/corycooper/workspace/github/caddy/caddy.go:228 +0x644
github.com/caddyserver/caddy/v2.Load({0x140002ec400?, 0xb?, 0x0?}, 0x0?)
        /Users/corycooper/workspace/github/caddy/caddy.go:127 +0x8c
github.com/caddyserver/caddy/v2/cmd.cmdRun({0x0?})
        /Users/corycooper/workspace/github/caddy/cmd/commandfuncs.go:212 +0x558
github.com/caddyserver/caddy/v2/cmd.caddyCmdToCoral.func1(0x140002e0500?, {0x1037fce86?, 0x2?, 0x2?})
        /Users/corycooper/workspace/github/caddy/cmd/cobra.go:114 +0x30
github.com/spf13/cobra.(*Command).execute(0x140002e0500, {0x140001c8b20, 0x2, 0x2})
        /Users/corycooper/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:852 +0x4c4
github.com/spf13/cobra.(*Command).ExecuteC(0x104baac40)
        /Users/corycooper/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x34c
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/corycooper/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
github.com/caddyserver/caddy/v2/cmd.Main()
        /Users/corycooper/workspace/github/caddy/cmd/main.go:64 +0x74
main.main()
        /Users/corycooper/workspace/github/caddy/cmd/caddy/main.go:39 +0x1c
```

francislavoie:
Is the dependency really necessary? This seems like the sort of thing we should just do ourselves in code. We try to avoid dependencies as much as possible.

mholt:
Yeah, retries can be done with a `for` loop.

yroc92:
So I have the `client.Do()` call replaced with this function, but keep getting a nil point error when trying to use ctx.Logger(). What am I missing?

```go
// Reattempts the http call using exponential back off when maxRetries is greater than 0.
func doHttpCallWithRetries(ctx caddy.Context, client *http.Client, request *http.Request, maxRetries int) (*http.Response, error) {
	var resp *http.Response
	var err error
	var errorMessage string
	for retry := 0; retry <= maxRetries; retry++ {
		resp, err = client.Do(request)
		if err != nil {
			errorMessage = fmt.Sprintf("problem calling endpoint: %v", err)
		} else if resp.StatusCode < 200 || resp.StatusCode > 499 {
			errorMessage = fmt.Sprintf("bad response status code from endpoint: %v", resp.StatusCode)
		} else {
			return resp, nil
		}
		ctx.Logger().Error(errorMessage)

		exponentialBackOff := math.Pow(2, float64(retry))
		ctx.Logger().Info(fmt.Sprintf("reattempting in %g seconds", exponentialBackOff))
		time.Sleep(time.Duration(exponentialBackOff) * time.Second)
	}

	return resp, err
}
```

Error:
```bash
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x100ac84ec]

goroutine 1 [running]:
github.com/caddyserver/caddy/v2.(*Logging).Logger(0x140000bd800, {0x0?, 0x0?})
        /Users/corycooper/workspace/github/caddy/logging.go:202 +0x2c
github.com/caddyserver/caddy/v2.Context.Logger({{0x101dc8c70, 0x14000683780}, 0x140000bd7d0, 0x1400018fd10, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}}, ...)
        /Users/corycooper/workspace/github/caddy/context.go:478 +0xc0
github.com/caddyserver/caddy/v2/caddyconfig.HTTPLoader.doHttpCallWithRetries({{0x14000688f58, 0x4}, {0x140001d4d50, 0x24}, 0x0, 0x0, 0x5, 0x0}, {{0x101dc8c70, 0x14000683780}, ...}, ...)
        /Users/corycooper/workspace/github/caddy/caddyconfig/httploader.go:140 +0xa4
github.com/caddyserver/caddy/v2/caddyconfig.HTTPLoader.LoadConfig({{0x14000688f58, 0x4}, {0x140001d4d50, 0x24}, 0x0, 0x0, 0x5, 0x0}, {{0x101dc8c70, 0x14000683780}, ...})
        /Users/corycooper/workspace/github/caddy/caddyconfig/httploader.go:101 +0x35c
github.com/caddyserver/caddy/v2.finishSettingUp({{0x101dc8c70, 0x14000683780}, 0x140000bd7d0, 0x1400018fd10, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}}, ...)
        /Users/corycooper/workspace/github/caddy/caddy.go:603 +0x4c0
github.com/caddyserver/caddy/v2.run(0x140000d6320?, 0x1)
        /Users/corycooper/workspace/github/caddy/caddy.go:525 +0x584
github.com/caddyserver/caddy/v2.unsyncedDecodeAndRun({0x140000d6280, 0x93, 0xa0}, 0x1)
        /Users/corycooper/workspace/github/caddy/caddy.go:337 +0xe4
github.com/caddyserver/caddy/v2.changeConfig({0x10166b466, 0x4}, {0x101678569, 0x7}, {0x14000474400, 0xe3, 0x200}, {0x0, 0x0}, 0x1)
        /Users/corycooper/workspace/github/caddy/caddy.go:228 +0x644
github.com/caddyserver/caddy/v2.Load({0x14000474400?, 0xb?, 0x0?}, 0x0?)
        /Users/corycooper/workspace/github/caddy/caddy.go:127 +0x8c
github.com/caddyserver/caddy/v2/cmd.cmdRun({0x0?})
        /Users/corycooper/workspace/github/caddy/cmd/commandfuncs.go:212 +0x558
github.com/caddyserver/caddy/v2/cmd.caddyCmdToCoral.func1(0x140003fc000?, {0x10166c146?, 0x2?, 0x2?})
        /Users/corycooper/workspace/github/caddy/cmd/cobra.go:114 +0x30
github.com/spf13/cobra.(*Command).execute(0x140003fc000, {0x140000bef80, 0x2, 0x2})
        /Users/corycooper/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:852 +0x4c4
github.com/spf13/cobra.(*Command).ExecuteC(0x102a12a20)
        /Users/corycooper/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x34c
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/corycooper/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
github.com/caddyserver/caddy/v2/cmd.Main()
        /Users/corycooper/workspace/github/caddy/cmd/main.go:64 +0x74
main.main()
        /Users/corycooper/workspace/github/caddy/cmd/caddy/main.go:39 +0x1c
        ```

**Comment by francislavoie** [comment]
The logger issue is related to this change https://github.com/caddyserver/caddy/commit/e43b6d81782ef79f22058179d8793f40cea89556

But I don't know why it's failing. It seems like it's not able to get a module from the context for some reason. I'm confused. It might be because config loader modules don't get `Provision`'d so they don't go through the module auto-wiring stuff. Matt will need to chime in here of course.

**Comment by yroc92** [comment]
I changed it to use the `caddy.Log()` directly which works.

**Comment by mholt** [comment]
I'll take a look at this shortly!

File: caddyconfig/httploader.go [hunk 1]
```diff
@@ -119,6 +124,31 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
...
+func doHttpCallWithRetries(client *http.Client, request *http.Request, maxRetries int) (*http.Response, error) {
+	var resp *http.Response
+	var err error
+
+	for retry := 0; retry <= maxRetries; retry++ {
+		if retry > 0 {
+			caddy.Log().Error(err.Error())
+			exponentialBackOff := math.Pow(2, float64(retry))
+			caddy.Log().Sugar().Infof("reattempting http load in %g seconds (%v retries remain)", exponentialBackOff, maxRetries-retry)
+			time.Sleep(time.Duration(exponentialBackOff) * time.Second)
```
mholt, line 137: Oh, fancy. I wasn't expecting exponential backoff. Props for the effort here.

Some thoughts:

Overall the backoff implementation looks sound. How useful is it? Usually, backoff is implemented for shared or public resources where we need to be careful and mindful clients (and is only useful if most clients implement it, too). For something like config loading, do we expect that backing off will be needed? How many clients are slamming the config source? I'm guessing not a lot (like, not tens-to-hundreds of thousands or more)... so I don't know if it has much utility?

Backoff does also allow the developer more time between retries to fix the problem. A few seconds for the first few retries, and by the time you get to the 20th retry, it's already been a few weeks. That might be an unnecessarily long time between retries.

So one question I have is, how important is having the backoff? And if we do keep it, should we cap the max time between retries to something more reasonable? The longer that span gets, the more time there will be between the backend being available and the actual config being obtained.

If it was just a linear loop (no backoff), the amount of time you spend retrying changes from 2^N seconds 2*N seconds where N is the number of retries (assuming 2 seconds between each one). Or, you could instead have the user configure how _long_ to retry for, and you choose how often to retry. Or, you could have the user configure how long _between each retry_, either forever or with a hard cap on when to give up.

I'm not sure which of those options is preferable here, but what are your thoughts?

Also, we try to avoid ever using `time.Sleep()` because if a config change needs to happen sooner, it will either block or leak resources (continue to run in the background, with no way to stop it). Instead:

```go
select {
    case <-time.After(interval):
    case <-ctx.Done():
        return ctx.Err()
}
```

^ that is the general idea; always select on the Context's Done channel so if it gets canceled, you can stop waiting.

(And even using `time.After()` might not be ideal; the godoc recommends creating a `time.NewTimer()` so it can be stopped and cleaned up if needed. But same idea.)

mholt (changes requested):
I think this looks good overall! I had some thoughts that I wanted to bounce off you.

caddyconfig/httploader.go
yroc92: Exponential back-offs are probably overkill now that I think about it. I was modeling that off of hashicorp's lib. A linear approach would be better probably. 2 seconds between firing would be fine.

In my use case, I'd want it to fire indefinitely until it works. Using a large number like 100000 for `max_retries` would be sufficient as the problem is apparent within hours. 

As for the `time.Sleep()` replacement, I'm not sure how to implement that properly so I might need some guidance there.

mholt: Cool; so maybe if you want it to try indefinitely, specify the time between each retry instead.

I can help with the time.Sleep replacement; it's basically what I showed you, just have to hold onto the context you get during provision.

caddyconfig/httploader.go
```diff
@@ -45,6 +45,10 @@ type HTTPLoader struct {
 	// Maximum time allowed for a complete connection and request.
 	Timeout caddy.Duration `json:"timeout,omitempty"`
 
+	// The number of seconds to wait in between attempts to call `URL`.
```
mholt, line 48:
```suggestion
	// Enable retries by specifying how long to wait between attempts to reach URL.
	// A retry would be performed if the connection fails or the response is not 2xx.
```

caddyconfig/httploader.go
```diff
@@ -45,6 +45,10 @@ type HTTPLoader struct {
...
+	// Defaults to 0, indicating no retries.
+	RetryDelay time.Duration `json:"retry_delay,omitempty"`
```
mholt, line 50:
```suggestion
	RetryDelay caddy.Duration `json:"retry_delay,omitempty"`
```

Not strictly necessary but is conventional as it allows users to specify `d` (day) units! Probably not relevant here but oh well.

caddyconfig/httploader.go
```diff
@@ -119,6 +123,33 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
...
+	} else if resp.StatusCode < 200 || resp.StatusCode > 499 {
+		err = fmt.Errorf("bad response status code from http loader url: %v", resp.StatusCode)
+	} else {
+		return resp, nil
+	}
+	// skip retries if valid retryDelay not provided
+	if retryDelay <= 0 {
+		return resp, err
+	}
+	// log the error
```
mholt, line 141:
```suggestion
```

Is obvious ;)

caddyconfig/httploader.go
```diff
@@ -119,6 +123,33 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
 	return result, nil
 }
 
+// Reattempts the http call, waiting `retryDelay` seconds in between attempts.
+func doHttpCallWithRetries(ctx caddy.Context, client *http.Client, request *http.Request, retryDelay time.Duration) (*http.Response, error) {
+	// make attempt
```
mholt, line 128:
```suggestion
```

caddyconfig/httploader.go
```diff
@@ -119,6 +123,33 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
...
+	// skip retries if valid retryDelay not provided
+	if retryDelay <= 0 {
+		return resp, err
+	}
+	// log the error
+	caddy.Log().Error(err.Error())
+	// wait for the retry delay to lapse before reattempting,
+	// or return when context is done
+	select {
+	case <-time.After(time.Second * retryDelay):
```
mholt, line 146:
```suggestion
	case <-time.After(retryDelay):
```

caddyconfig/httploader.go
```diff
@@ -119,6 +123,33 @@ func (hl HTTPLoader) LoadConfig(ctx caddy.Context) ([]byte, error) {
...
+		return doHttpCallWithRetries(ctx, client, request, retryDelay)
+	case <-ctx.Done():
+		return nil, err
```
mholt, line 149:
```suggestion
		return nil, ctx.Err()
```

mholt (changes requested):
This is looking really good!

Most of my comments are nits but I have a few technical suggestions as well. Hopefully these make sense.

Thanks for your patience with me.

mholt (approved):
Awesome!! Looking much better and more straightforward.

Just a few suggestions for code tidiness but otherwise this has my approval.

mholt (approved):
It's perfect. Thanks Cory!

liedlgruber (commented):
caddyconfig/httploader.go

PR #5079, opened 2022-09-24
ci: fix integration tests

mohammed90:
The change in #5045 did as the message says, _hide_ the issue. The integration tests were not successful; they were skipped because the code did not receive confirmation the server was running, so it skipped it. There were 2 issues with it:
- The `grace_period` option in the caddyfile always expects a unit suffix, while the JSON is more graceful. I merely added the `ns` suffix to the caddyfile instead of fighting the unmarshaler.
- The admin listener endpoint got changed in the Go code, but it was never fed to the initial server. The code was not receiving a response. I included an initial config setting only the admin endpoint, and included the same endpoint in all the test config data.

I also added `skip_install_trust` and its corresponding adapted JSON format in all tests to avoid the root installation step when it encounters any `*.localhost` domain name, which triggers the automatic HTTPS behavior.

### DIFF
--- a/caddytest/caddytest.go
+++ b/caddytest/caddytest.go
@@ -237,7 +237,7 @@ func validateTestPrerequisites() error {
 
 	if isCaddyAdminRunning() != nil {
 		// start inprocess caddy server
-		os.Args = []string{"caddy", "run"}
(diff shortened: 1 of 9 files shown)


francislavoie (approved):
Awesome. Thank you!

PR #5081, opened 2022-09-25
caddyhttp: replace placeholders in map defaults

willnorris:
This updates the map directive to replace placeholders in default values in the same way as matched values.

I'm trying to use the map directive to override the username for some users in the [Tailscale Caddy plugin](https://github.com/tailscale/caddy-tailscale).  For example, I want to set `webauth.user` to "willnorris" if `http.auth.user.tailscale_user` equals "will@tailscale.com", but to default to the value of `http.auth.user.tailscale_login` in all other cases.  Prior to this patch, the default value would be the literal string "{http.auth.user.tailscale_login}" instead of expanding the placeholder.

```
map {http.auth.user.tailscale_user} {webauth.user} {
    will@tailscale.com willnorris
    default {http.auth.user.tailscale_login}
}
```

### DIFF
--- a/modules/caddyhttp/map/map.go
+++ b/modules/caddyhttp/map/map.go
@@ -169,7 +169,7 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhtt
 
 		// fall back to default if no match or if matched nil value
 		if len(h.Defaults) > destIdx {
-			return h.Defaults[destIdx], true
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thanks! I forgot about the Defaults... heh.

At some point I should probably enhance the "infinite recursion" check above (in Provision) to include Defaults.

PR #5084, opened 2022-09-26
feat: add dockerfile

tetra12:
This PR adds a dockerfile "for everyone else". Some times its required to build the current master in a container and run integration tests with compose/kuber/CD, etc

I personally built this to test the latest master on my evaluation cluster.

So wanted to share a quick way to build a docker for all others.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM golang:1.19.1-alpine as builder
+
+WORKDIR /workspace
+
+COPY go.mod go.sum /workspace/
+RUN go mod download
+
+COPY . /workspace/
+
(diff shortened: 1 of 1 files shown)


hairyhenderson:
@tetra12 is there a reason the Caddy builder image doesn't work for you? That would work if you used `xcaddy build master`.

I'd rather not have multiple Dockerfiles living in multiple repos. We've decided to keep the official Caddy Dockerfiles in the `caddy-docker` repo.

Given the above, I'm going to close this - if there's some reason the builder image doesn't work for you, feel free to file an issue in https://github.com/caddyserver/caddy-docker/

PR #5097, opened 2022-09-29
forwardauth: Canonicalize header fields (fix #5038)

### LINKED ISSUE #5038: forward_auth copy_headers must be case sensitive
Related docs: https://caddyserver.com/docs/caddyfile/directives/forward_auth#syntax

When using `forward_auth`, I noticed the `copy_headers` key accepts space-separated case-sensitive values. Should this config be case-insensitive? If so, I can help contribute to the docs.

mholt:
Seems simple enough to canonicalize the header field names.

### DIFF
--- a/modules/caddyhttp/reverseproxy/forwardauth/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/forwardauth/caddyfile.go
@@ -38,29 +38,28 @@ func init() {
 // configured for most™️ auth gateways that support forward auth. The typical
 // config which looks something like this:
 //
-//     forward_auth auth-gateway:9091 {
-//         uri /authenticate?redirect=https://auth.example.com
-//         copy_headers Remote-User Remote-Email
-//     }
+//	forward_auth auth-gateway:9091 {
+//	    uri /authenticate?redirect=https://auth.example.com
+//	    copy_headers Remote-User Remote-Email
+//	}
 //
 // is equivalent to a reverse_proxy directive like this:
 //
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/forwardauth/caddyfile.go
```diff
@@ -196,9 +195,7 @@ func parseCaddyfile(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error)
 	// need at least one handler in the routes for the response handling
 	// logic in reverse_proxy to not skip this entry as empty.
 	for from, to := range headersToCopy {
-		handler.Request.Set[to] = []string{
-			"{http.reverse_proxy.header." + from + "}",
-		}
+		handler.Request.Set.Set(to, "{http.reverse_proxy.header."+http.CanonicalHeaderKey(from)+"}")
```
francislavoie, line 198: I guess we can canonicalize both sides? Maybe `.Set()` already does that for us though?

```suggestion
		handler.Request.Set.Set(http.CanonicalHeaderKey(to), "{http.reverse_proxy.header."+http.CanonicalHeaderKey(from)+"}")
```

mholt: Yep, the header methods do the canonicalization for us: https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/net/http/header.go;l=24-60

francislavoie (approved):

PR #5099, opened 2022-09-30
treat invalid file path as notFound

likev:
treat invalid file path as notFound so that PassThru can work
some  invalid file path can be valid URL path


francislavoie:
Please show us examples where this would make a difference. How can we test this behaviour?

mholt:
^ Francis beat me to it :sweat_smile: 

I was about to ask the same thing. What is an invalid file path that you've actually encountered?

likev:
> ^ Francis beat me to it sweat_smile
> 
> I was about to ask the same thing. What is an invalid file path that you've actually encountered?

https://example.com/a:b is a valid URL but `a:b` is not a valid directory names under both Windows and macOS.

https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names
https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid

PR #5102, opened 2022-10-01
logging: Fix `skip_hosts` with wildcards

### LINKED ISSUE #4859: [BUG?] Access logs for sub-domains are not written after Apr 26 update
Caddy version:

```
v2.5.1 h1:bAWwslD1jNeCzDa+jDCNwb8M3UJ2tPa8UZFFzPVmGKs=
```

running as: `systemctl start caddy` on `Ubuntu 20.04.4 LTS`

Caddy file:

```
(logging) {
	log {
		format json {
			message_key msg
			level_key level
			time_key ts
			name_key name
			time_format wall_milli
		}
		output file /var/log/caddy/{args.0}.access.log {
			roll_size 10mb
			roll_keep 20
			roll_keep_for 30d
		}
		level INFO
	}
}

*.domain.com {
	redir https://domain.com{uri}
}

domain.com {
	import logging domain.com
	respond "ok" 200
}

sub1.domain.com {
	import logging sub1.domain.com
	respond "ok" 200
}

sub2.domain.com {
	import logging sub2.domain.com
	respond "ok" 200
}
```

Expected behavior:
---

in the folder `/var/log/caddy` we should see the following files when each sub-domain is accessed:
 * domain.com.access.log
 * sub1.domain.com.access.log
 * sub2.domain.com.access.log

Issue
---
 * Currently there is only one log file created for the root domain; access logs for sub-domains are not created
 
 
Tried already:
 ---
  - replace `import logging sub1.domain.com` with:
    ```
            log {
                output file /var/log/caddy/sub1.domain.com
        }
    ```
 - replacing `.` with `_` in the domain name, e.g. `import logging sub1_domain_com`
 - permissions to access `/var/log/caddy` are correct
 
Notes:
---

prior Apr 26 everything was working fine.

francislavoie:
Fix #4859

See https://github.com/caddyserver/caddy/issues/4859#issuecomment-1171504267 for an explanation of the bug.

The fix was simple, really, we just need to reorder the loops in `server.go` to return true early if a logger is explicitly configured for a hostname.

While testing it though, I noticed that there's a bug in the Caddyfile adapter when using the `log` directive with no options (i.e. log to the default logger); it wasn't adding those hosts to the `logger_names` config, so the wildcard `skip_hosts` would still cause that to get skipped when it was explicitly configured to log.

Basically, I removed some code because it was trying to be "too smart" with omitting some config in the output JSON for "cleanliness" but that just causes edgecase bugs.

Also a side benefit is that logger checks should be _slightly_ faster now (like... nanoseconds at best) because we check a map in `O(1)` before looping over `skip_hosts` which is `O(n)`.

Also set up a const in the `caddy` package for the default logger name, because we use that in lots of places.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -735,7 +735,7 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 				// reference the default logger. See the
 				// setupNewDefault function in the logging
 				// package for where this is configured.
-				globalLogName = "default"
+				globalLogName = caddy.DefaultLoggerName
 			}
 
 			// Verify this name is unused.
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -219,7 +219,7 @@ func (st ServerType) Setup(inputServerBlocks []caddyfile.ServerBlock,
 		if ncl.name == "" {
 			return
 		}
-		if ncl.name == "default" {
+		if ncl.name == caddy.DefaultLoggerName {
 			hasDefaultLog = true
 		}
 		if _, ok := options["debug"]; ok && ncl.log.Level == "" {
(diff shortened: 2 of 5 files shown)


mholt:
Huh, thanks! I like fixes that delete code. Thanks for correcting my mistake. :grin: :clap:

mholt (approved):

Issue #4924, opened 2022-07-31
A way to get actual client IP


ArabCoders:
Hello, 

It seems using `request_header` directive to alter request headers does not make it to the `log` directive, i mainly use the transformer plugin to alter the logged IP, some requests are via CDN others directly, and it seems caddy does not always set `X-Forwarded-For` which i think is probably the intended behavior. Anyway here is an example of the behavior i am describing

```
{
	admin 127.0.0.1:20199
	debug
	log stdout
}

:7885 {
	log

	@no_forawrded_for not header X-Forwarded-For *
	request_header @no_forawrded_for X-Forwarded-For {remote_host}

	respond {header.X-Forwarded-For} 200
}
```

as you can see from the output, it's clearly being set.

```
$ http http://localhost:7885

HTTP/1.1 200 OK
Content-Length: 9
Date: Sun, 31 Jul 2022 19:14:12 GMT
Server: Caddy

127.0.0.1
````

However, the logger does not show the header.

```
2022/07/31 19:14:05.744 INFO    serving initial configuration
2022/07/31 19:14:05.744 INFO    tls     finished cleaning storage units
2022/07/31 19:14:05.744 INFO    watcher watching config file for changes        {"config_file": "./Caddyfile"}
2022/07/31 19:14:12.440 INFO    http.log.access handled request {"request": {"remote_ip": "127.0.0.1", "remote_port": "56400", "proto": "HTTP/1.1", "method": "GET", "host": "localhost:7885", "uri": "/", "headers": {"User-Agent": ["HTTPie/1.0.3"], "Accept-Encoding": ["gzip, deflate"], "Accept": ["*/*"], "Connection": ["keep-alive"]}}, "user_id": "", "duration": 0.0001012, "size": 9, "status": 200, "resp_headers": {"Server": ["Caddy"], "Content-Type": []}}
```

i expect to see the added header to show up, however it seems the log directive receive the unmodified request object. If that's the intended behavior, then i apologize.

I have log analyzers and they are failing due to missing IP in logs as some requests originating locally or someone hitting the IP directly instead of the frontend proxies.

francislavoie:
> It seems using request_header directive to alter request headers does not make it to the log directive

That's by design. The logs show the original headers as Caddy received them. The log entry is provisioned _before_ any handling happens.

[...]

PR #5103, opened 2022-10-01
caddyhttp: Add server-level `trusted_proxies` config

francislavoie:
Partial work for #4924 as per my plan in https://github.com/caddyserver/caddy/issues/4924#issuecomment-1206477962

We've had requests for defining `trusted_proxies` globally, so it doesn't need to be configured for each `reverse_proxy`. This does that.

But also, this opens the door for more places we can use trusted proxies, such as in #4924 which asks for getting the real client IP. We can do that by looking at the XFF header _if the proxy is trusted_. That part isn't done in this PR, _edit:_ this is in #5104!

Also we might want to change how `remote_ip forwarded` works, possibly making it not trust forwarded by default (which would be a breaking change... but eh, it's known to be insecure right now). ~~And maybe we should add `client_ip` to the logs.~~ See #5104 for the logs part.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/netip"
 	"net/url"
 	"runtime"
 	"strings"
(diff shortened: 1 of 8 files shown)


modules/caddyhttp/server.go
```diff
@@ -732,4 +799,7 @@ const (
 	// For a partial copy of the unmodified request that
 	// originally came into the server's entry handler
 	OriginalRequestCtxKey caddy.CtxKey = "original_request"
+
+	// For tracking whether the client is a trusted proxy
+	TrustedProxyCtxKey caddy.CtxKey = "trusted_proxy"
```
mholt, line 804: I wonder if this should just be stored as a var, that way it could be used by placeholders or CEL expressions, etc.

mholt (commented):
Thanks - overall this LGTM, just a question about implementation. I still want to review this one more time before merging, even if no more changes are made. Not sure if it'll go into 2.6.2 or 2.6.3.

modules/caddyhttp/server.go
francislavoie: Sure - I'm never sure which one is better to use 🤷‍♂️ a var would be fine 👍

francislavoie:
I'll get back to this soon. My priorities had been elsewhere for the past while.

modules/caddyhttp/server.go
francislavoie: Alright, updated to use a var. We get replacer support out of the box 👍

francislavoie:
Alright, this is ready for another round of review. I'm pretty happy with this.

mholt (commented):
This is looking great!

My only request is, can we use the existing Vars context value instead of creating a new one? Each time we create a new one, it makes writing mocks a little more tedious and error-prone.

mholt (approved):
Wow you're right. I'm blind. LGTM! Thank you :)

PR #4066, opened 2021-03-16
Add dark color scheme media query

radiden:
Adds a dark color scheme if the user's browser prefers one.

Tried to keep to the original's design as well as I could, any propositions are welcome.

PR #5105, opened 2022-10-01
fileserver: better dark mode visited link contrast

iliana:
PR #4066 added a dark color scheme to the file_server browse template. PR #4356 later set the links for the `:visited` pseudo-class, but did not set anything for the dark mode, resulting in poor contrast. I selected some new colors by feel.

This commit also adds an `a:visited:hover` for both, to go along with the normal blue hover colors.

image
image


mholt (approved):
Thank you for making this better!

PR #5109, opened 2022-10-02
logging: Better `console` encoder defaults

francislavoie:
This is something that has bothered me for a while, so I figured I'd do something about it now since I'm playing in the logging code lately.

The `console` encoder doesn't actually match the defaults that zap's default logger uses. This makes it match better with the rest of the logs when using the `console` encoder alongside somekind of filter, which requires you to configure an encoder to wrap.

Before:
image

After:
image

As you can see, this colorizes + uppercases the level, and makes the timestamp not useless.

Worth saying, that technically zap's default logger sniffs the TTY to figure out whether it's interactive or not, and uses JSON if non-interactive, and a "modified" console encoder if interactive. I think it might turn off color if the terminal doesn't support it. But I don't think that's worth trying to do here for these defaults, because it only affects the log level and most people get the gist even if their terminal doesn't render the color shell escapes.

### DIFF
--- a/modules/logging/encoders.go
+++ b/modules/logging/encoders.go
@@ -45,15 +45,21 @@ func (ConsoleEncoder) CaddyModule() caddy.ModuleInfo {
 
 // Provision sets up the encoder.
 func (ce *ConsoleEncoder) Provision(_ caddy.Context) error {
+	if ce.LevelFormat == "" {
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks, I can appreciate this :)

PR #5114, opened 2022-10-03
Support absolute windows path in unix reverse proxy address 

ueffel:
When using unix domain sockets on windows, it is currently not possible to specify an absolute path in the address like so 

```caddy-d
reverse_proxy /* unix/c:\absolute\path.sock
```

because everything after the colon is interpreted as port.

This PR adds support for absolute paths.

I saw the TODO to add tests for this `parseUpstreamDialAddress`, so I did that as well.


listeners.go
```diff
@@ -289,8 +289,10 @@ func (na NetworkAddress) String() string {
 	return JoinNetworkAddress(na.Network, na.Host, na.port())
 }
 
-func isUnixNetwork(netw string) bool {
-	return netw == "unix" || netw == "unixgram" || netw == "unixpacket"
+// IsUnixNetwork returns true if the netw is a unix network.
+func IsUnixNetwork(netw string) bool {
```
francislavoie, line 293: Renaming this to be exported doesn't work because this exact name is already used in another file. See the linter errors.

francislavoie (changes requested):

francislavoie:
You can confirm that `AF_UNIX` sockets do indeed work on Windows?

I've never been able to get it to work. I tried to use `curl` to connect to Caddy bound to a unix socket, but `curl` had problems that prevented it from working properly.

ueffel:
Yes I can confirm unix domain sockets are working on windows. I have two small go-applications that listen on UDSes and caddy can reverse proxy them. Also the windows curl version can talk to this apps like so `c:\Windows\System32\curl -v --unix-socket "r:\temp\packagecontrol.sock" http://localhost/` (curl 7.83.1)

listeners.go
ueffel: Argh I forgot to rename the unix-only usages.

listeners.go
```diff
@@ -289,8 +289,10 @@ func (na NetworkAddress) String() string {
...
+	return netw == "unix" || netw == "unixgram" || netw == "unixpacket" ||
+		netw == "unix+h2c" || netw == "unixgram+h2c" || netw == "unixpacket+h2c"
```
francislavoie, line 295: I'm not sure how I feel about these... the `+h2c` thing is really a `reverseproxy`-specific thing, I'm not sure we need to have this in `listener.go` :thinking:

francislavoie: Could we just do `strings.HasPrefix(netw, "unix")` ? Is that too permissive?

I'll let @mholt decide

ueffel: It is actually needed because `SplitNetworkAddress` otherwise does not return everything after "unix/" which is used in the new code.
The alternative would be some duplicated code, which I tried to avoid.

But let's hear from @mholt :)

mholt:
@ueffel Wait, Unix sockets work on Windows?? Which version of Windows are you on? Did you do anything special to make them work?

I ask because in commit https://github.com/caddyserver/caddy/commit/e3e8aabbcf65d37516bb97f9dc0f77df52f8cf55 I've moved Unix sockets to non-Windows source files because of compilation reasons.

I'd be curious if you try using Unix sockets after that commit to see if it still works on Windows, and I'd be surprised if they do. :thinking: 

I'm just wondering if I need to move the Unix socket code out of non-Windows source files.

francislavoie:
Yeah @mholt apparently since 2017 https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/. I tried to test it out (with Caddy running as a server) a couple years ago but I couldn't figure out how to use a client that would connect using it. But @ueffel is using Caddy itself as the client to another Go app. I should've tried that tbh :rofl: I would've probably gotten further by testing that out.

ueffel:
@mholt Like @francislavoie said, it's been around for a while :). I did not do anything special.
Essentially 

```go
srv := http.Server{}
...
listener, err = net.Listen("unix", socketPath)
...
err = srv.Serve(listener)
```

in my go applications and then in caddy just something like

```caddy-d
reverse_proxy unix/<socketPath>
```

and it works.

francislavoie:
Yeah @mholt this isn't about listening on unix sockets on Windows (although that should still work hopefully, no real reason it shouldn't I think), it's more about proxy addresses.

That said, @ueffel could you try binding to a unix socket with one Caddy instance and run a second that proxies to it? I'm not on my Windows env right now so I can't try it myself at the moment.

ueffel:
@francislavoie 
I tried the following:

Caddyfile 1:

```caddy-d
{
	debug
	admin off
	auto_https off
}

http://localhost {
	bind unix/caddy.sock
	respond * "well hello there, served via unix domain sockets
"
}
```

Caddyfile 2:

```caddy-d
{
    debug
    admin off
    auto_https off
}

http://:80 {
    reverse_proxy * unix/caddy.sock
}
```

```sh
$ curl.exe -v --unix-socket caddy.sock http://localhost/
*   Trying caddy.sock:0...
* Connected to localhost () port 80 (#0)
> GET / HTTP/1.1
> Host: localhost
> User-Agent: curl/7.85.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: Caddy
< Date: Mon, 03 Oct 2022 23:34:26 GMT
< Content-Length: 49
<
well hello there, served via unix domain sockets
* 
```

```sh
$ curl.exe -v http://localhost/
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> Host: localhost
> User-Agent: curl/7.85.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Length: 49
< Date: Mon, 03 Oct 2022 23:34:29 GMT
< Server: Caddy
< Server: Caddy
<
well hello there, served via unix domain sockets
```

Working fine, i guess? ^^

francislavoie:
Nice! And that's with your branch which is based on `master`, including the commit Matt mentioned? Good news then :+1:

ueffel:
I tried `v2.6.1`, `master` and a custom build from my branch

listeners.go
mholt: I need some more time to think about this solution -- I love this change overall though. I will review this for 2.6.3

mholt: After "some more time" (4 months, haha sorry) I think that this is fine (for now); we could also do the `HasPrefix()` as that would be easier to maintain, I think.

ueffel: I pushed the `HasPrefix` change.

francislavoie (approved):
Thanks! This looks good to me.

The tests are greatly appreciated! It looks pretty comprehensive :blush:

mholt (approved):
Excellent. Thanks @ueffel !

PR #5115, opened 2022-10-03
Switch to program directory when running as Windows service

TobiX:
As noted in #5106, when installing Caddy as a native Windows service (For reference, the PowerShell command for that is `New-Service -Name caddy -BinaryPathName "C:\caddy\caddy.exe run"`), caddy starts up in `C:\WINDOWS\SYSTEM32` - I think this is a bad idea for multiple reasons:
- You obviously cannot put an "adjacent" `Caddyfile` into `system32` (well, you technically can, but that sound like a bad idea all around) and as an alternative, you have to run caddy with a much longer command line (Like `New-Service -Name caddy -BinaryPathName "C:\caddy\caddy.exe run --config C:\caddy\Caddyfile"`
- If somebody tries to use relative paths in the `Caddyfile`, those will be relative to the `system32` directory, which will lead to hard-to-debug problems...

This is a "prototype" of a simple fix: When we are running as a Windows service, try to chdir into the caddy program directory - It works for me[tm] in a Windows 10 VM...

### DIFF
--- a/service_windows.go
+++ b/service_windows.go
@@ -15,6 +15,9 @@
 package caddy
 
 import (
+	"os"
+	"path/filepath"
+
 	"github.com/caddyserver/caddy/v2/notify"
 	"golang.org/x/sys/windows/svc"
 )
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks! I didn't know that, so I agree this is probably a better default.

Ignoring the error here is _probably_ fine? I'll merge it, but I wonder if we should at least log a warning.

TobiX:
@mholt Unfortunately, there is nowhere to log TO when running as a windows service, since STDOUT & STDERR are discarded... IF somebody implements Windows event logging (https://pkg.go.dev/golang.org/x/sys/windows/svc/eventlog), that should become the default for Windows services, so errors can be logged... somewhere :smile_cat:

mholt:
Good to know, thanks!

PR #5120, opened 2022-10-05
httpcaddyfile: Improve detection of indistinguishable TLS automation policies

mholt:
This is a somewhat experimental / WIP branch.

The problem is we have listener plugins that provide their own TLS, and we shouldn't be preventing a valid config from being produced. I think the assumptions made in this part of the code were too bold, and we need to relax them slightly now.

Initial change is a bit of a hack, but seems valid too. If `auto_https off` is used, we don't even use `httpsHostsSharedWithHostlessKey` so we can skip the count which yields an error sometimes when it shouldn't.


mholt:
Ok, I think this change is pretty good overall. It removes some unnecessary (and necessarily incorrect) checks, tweaks tests to hopefully improve their likelihood of passing (sigh, that race condition), updates a dependency (unrelated but accidentally snuck into this PR instead of a separate branch), and fixes HTTP/3 listening when the network type is unknown (because it may be specified by a plugin).

PR #5127, opened 2022-10-07
replacer: working directory global placeholder

dunglas:
Sometimes it can be useful to have access to the current working directory in the config file.
This patch adds a placeholder, `system.wd`, containing it.


francislavoie:
We'll also need to add it to the docs here https://github.com/caddyserver/website/blob/master/src/docs/markdown/conventions.md#placeholders

francislavoie (approved):

mholt:
Thanks!

PR #5130, opened 2022-10-08
httpcaddyfile: Wrap site block in subroute if host matcher used

### LINKED ISSUE #5124: Caddy tries to get certificate for domains only in host matcher within site block
Does JSON deactivate [host matcher](https://caddyserver.com/docs/json/apps/http/servers/routes/match/host/) to activate automatic HTTPS？
Example:
```
{
	order forward_proxy before file_server
	admin off
	log {
		level ERROR
	}
	storage file_system {
		root /home/tls
	}

	auto_https off
	default_sni *.xx.yy
	cert_issuer zerossl
	acme_dns duckdns b8a33cc1-1111-2222-3333-b049bf07988f {
		override_domain zz.duckdns.org
	}

	servers unix//dev/shm/h1h2c.sock {
		listener_wrappers {
			proxy_protocol
		}
		protocols h1 h2c
	}

	servers 127.0.0.1:8443 {
		listener_wrappers {
			proxy_protocol
			tls
		}
	}
}

:80 {
	redir https://{host}{uri} permanent
}

:88 {
	bind unix//dev/shm/h1h2c.sock

	file_server {
		root /var/www/html
	}
}

:8443, *.xx.yy:8443 {
	bind 127.0.0.1

	forward_proxy {
		basic_auth user pass
		hide_ip
		hide_via
		probe_resistance
	}

	@host2 {
		host wh.xx.yy ch.xx.yy
	}
	file_server @host2 {
		root /var/www/html
	}
}
```
The above example has been normal before. 
After [this master update](https://github.com/caddyserver/caddy/commit/2808de1e30b873f9c8f4693eae59bd56f20ebbbe), the automatic parameter is lost in the TLS configuration, that is, the certificate cannot be applied for *.xx.yy. 
If the above configuration deletes `auto_https off`, it is found that the background program automatically applies for a certificate for wh.xx.yy、ch.xx.yy.
Based on the above example, thinking about the problem of the update: I think it is the most reasonable method to cancel the host matcher to activate automatic HTTPS. Benefit: Makes the Caddyfile well-configured. Make the host matcher flexible.

mholt:
Fix #5124

The high level structure of the Caddyfile is converted to JSON config by wrapping each site block in a `subroute` handler, where each handler is matched by the hostname at the head of the site block.

However, site blocks that apply to the whole server, i.e. without a host matcher at its head:

```
:443 {
   ...
}
```

normally don't need to be wrapped in a `subroute` handler because there are no other site blocks on that listener, so we eliminate the unnecessary nesting.

But if that site block contains a host matcher:

```
:443 {
    @example host example.com
    ... @example ...
}
```

then previously it would become a host matcher at a top-level route in the JSON, which is used for activating automatic HTTPS. (See #5124.) This is unexpected, so now we only avoid wrapping the site block if there is no host matcher. A host matcher defined in that fashion should stay within a subroute because auto-HTTPS is *not* expected on it.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -907,11 +907,32 @@ func appendSubrouteToRouteList(routeList caddyhttp.RouteList,
 		return routeList
 	}
 
+	// No need to wrap the handlers in a subroute if this is the only server block
+	// and there is no matcher for it (doing so would produce unnecessarily nested
+	// JSON), *unless* there is a host matcher in this site block; if so, then we
+	// still need to wrap in a subroute because otherwise the host matcher from
+	// the inside of the site block would be a top-level host matcher, which is
+	// subject to auto-HTTPS (cert management), and using a host matcher within
+	// a site block is a valid, common pattern for excluding domains from cert
+	// management, leading to unexpected behavior; see issue #5124.
+	wrapInSubroute := true
 	if len(matcherSetsEnc) == 0 && len(p.serverBlocks) == 1 {
-		// no need to wrap the handlers in a subroute if this is
-		// the only server block and there is no matcher for it
-		routeList = append(routeList, subroute.Routes...)
-	} else {
+		var hasHostMatcher bool
+	outer:
+		for _, route := range subroute.Routes {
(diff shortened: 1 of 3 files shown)


mholt:
Wow, weird. The tests passed on my system.

PR #5140, opened 2022-10-13
log response size for websocket request

WeidiDeng:
Fix http access log will not record size and status for hijacked response writer.

### DIFF
--- a/modules/caddyhttp/responsewriter.go
+++ b/modules/caddyhttp/responsewriter.go
@@ -211,11 +211,7 @@ func (rr *responseRecorder) ReadFrom(r io.Reader) (int64, error) {
 	var n int64
 	var err error
 	if rr.stream {
-		if rf, ok := rr.ResponseWriter.(io.ReaderFrom); ok {
-			n, err = rf.ReadFrom(r)
-		} else {
-			n, err = io.Copy(rr.ResponseWriter, r)
-		}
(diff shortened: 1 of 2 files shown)


mholt (approved):
Phew, I'm finally getting to this!

Francis and I talked about it today on a call. It is definitely a little weird to mix HTTP and WebSockets info in the access log. And it's also a bit weird to count both bytes written and read after the upgrade as part of the 'size' of the response.

We talked about how, even if we try to just log the HTTP request up to the upgrade and then log the WS separately, the HTTP request log won't be emitted until after the WS is closed anyway since it blocks that request's handler chain until the WS connection closes. So you'd still end up with the log being delayed until the end of the WS anyway, and have two logs instead.

Since, with our current implementation of access logging, the logs would be emitted at the same time anyway, we decided we might as well keep them together in one log.

Showing the corrected status code is definitely a good change.

Counting the number of bytes read+written may be useful but a bit confusing to combine; I'm not sure we can really separate those two in the logs, though. (Not without a separate log entry.) So that's probably OK for now.

I think, in a more ideal solution:

- The HTTP access log would be emitted right after the upgrade to WS completes -- with the proper status code.
- A separate log would be emitted after the WS connection closes, showing the separate counts of bytes written and read.

Since I don't know how to do that right now, let's go with this as-is -- knowing that this implementation is also a bug, but it's still better than what we have now.

Thank you, @WeidiDeng, for the great patch!

PR #5144, opened 2022-10-13
fileserver: fix breadcrumbs for directories with % percent sign in the name

### LINKED ISSUE #5143: fileserver: Breadcrumbs are broken if directory name contains percent sign %
Breadcrumbs are generated without text if directory name contains a percent sign.
The bug is present in both

```
> caddy version
v2.6.1 h1:EDqo59TyYWhXQnfde93Mmv4FJfYe00dO60zMiEt+pzo=
```

and also the latest master 6bad878a22e048762262d6fabe2144cefaf4ca81

To replicate:  
Take this file tree 

```
> tree
.
└── dir with % in the name
    └── another
        └── with %
            └── sub
```

and run caddy with `caddy file-server --browse`

it creates this HTML:

```html
<header>
	<h1>
		<a href="../../../../">/</a><a href="../../../"></a>/<a href="../../">another</a>/<a href="../"></a>/<a href="">sub</a>/
	</h1>
</header>
```

Note the missing link text for directories with `%` signs in their name.

A test case to replicate the issue:

```golang
// put this in
// modules/caddyhttp/fileserver/browsetplcontext_test.go:21

func TestBreadcrumbs(t *testing.T) {
	testdata := []struct {
		path     string
		expected []crumb
	}{
		{"folder/with % sign/sub", []crumb{
			{Link: "../../", Text: "folder"},
			{Link: "../", Text: "with % sign"}, // .Text == "" here
			{Link: "", Text: "sub"},
		}},
```

abdusco:
Closes #5143

### DIFF
--- a/modules/caddyhttp/fileserver/browsetplcontext.go
+++ b/modules/caddyhttp/fileserver/browsetplcontext.go
@@ -92,9 +92,9 @@ func (fsrv *FileServer) directoryListing(ctx context.Context, entries []fs.DirEn
 			Mode:      info.Mode(),
 		})
 	}
-	name, _ := url.PathUnescape(urlPath)
+	name := unescapeSlash(path.Base(urlPath))
 	return browseTemplateContext{
-		Name:     path.Base(name),
+		Name:     name,
 		Path:     urlPath,
 		CanGoUp:  canGoUp,
 		Items:    fileInfos,
@@ -157,12 +157,20 @@ func (l browseTemplateContext) Breadcrumbs() []crumb {
 		// the directory name could include an encoded slash in its path,
 		// so the item name should be unescaped in the loop rather than unescaping the
 		// entire path outside the loop.
-		p, _ = url.PathUnescape(p)
+		p = unescapeSlash(p)
 		lnk := strings.Repeat("../", len(parts)-i-1)
 		result[i] = crumb{Link: lnk, Text: p}
 	}
 
 	return result
+}
+
+func unescapeSlash(s string) string {
+	// url.PathUnescape can't handle all uses of % symbol
+	// because it is used to encode special chars in URLs.
+	// "a % b" is valid folder name, but an invalid percent-encoding.
+	s = strings.ReplaceAll(s, "%2F", "/")
+	return strings.ReplaceAll(s, "%2f", "/")
 }
 
 func (l *browseTemplateContext) applySortAndLimit(sortParam, orderParam, limitParam string, offsetParam string) {
--- a/modules/caddyhttp/fileserver/browsetplcontext_test.go
+++ b/modules/caddyhttp/fileserver/browsetplcontext_test.go
@@ -15,6 +15,7 @@
 package fileserver
 
 import (
+	"context"
 	"testing"
 )
 
@@ -49,6 +50,16 @@ func TestBreadcrumbs(t *testing.T) {
 			{Link: "../", Text: "/"},
 			{Link: "", Text: "مجلد/1"},
 		}},
+		{"with %2f encoded %2F/slashes/not % 2F encoded", []crumb{
+			{Link: "../../", Text: "with / encoded /"},
+			{Link: "../", Text: "slashes"},
+			{Link: "", Text: "not % 2F encoded"},
+		}},
+		{"folder/with % sign/sub", []crumb{
+			{Link: "../../", Text: "folder"},
+			{Link: "../", Text: "with % sign"},
+			{Link: "", Text: "sub"},
+		}},
 	}
 
 	for _, d := range testdata {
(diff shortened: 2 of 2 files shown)


modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -157,14 +157,22 @@ func (l browseTemplateContext) Breadcrumbs() []crumb {
...
 	return result
 }
 
+func unescapeSlash(s string) string {
+	// url.PathUnescape can't handle all uses of % symbol
+	// because it is used to encode special chars in URLs.
+	// "a % b" is valid folder name, but an invalid percent-encoding.
+	s = strings.ReplaceAll(s, "%2F", "/")
+	return strings.ReplaceAll(s, "%2f", "/")
+}
```
abdusco, line 174: would the use of `strings.Replacer` be appropriate here?

modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -157,14 +157,22 @@ func (l browseTemplateContext) Breadcrumbs() []crumb {
...
+
```
abdusco, line 175: would the use of `strings.Replacer` be appropriate here?

mohammed90 (commented):
Great fix! I was worried it might break other special characters, but my testing shows no breakage for them (e.g. `/with @ in dirname`). You can probably add tests for them.

You can use the same function to fix line lines 95-97. I've tested it locally and it works. Once that is done, this PR is good to go 👍🏼

abdusco:
> You can use the same function to fix line lines 95-97. I've tested it locally and it works. Once that is done, this PR is good to go 👍🏼

@mohammed90 I'm writing tests for it, and I am not sure what the expected behavior is:

in `percent-encoded slash` case, I'd expect to receive `with/slash` as the name, not `slash`

```golang
func TestFileServer_directoryListing(t *testing.T) {
	tests := []struct {
		name         string
		urlPath      string
		expectedName string
	}{
		{
			name:         "no-encoding",
			urlPath:      "/with%percent",
			expectedName: "with%percent",
		},
		{
			name:         "encoded space",
			urlPath:      "/with%20space",
			expectedName: "with%20space",
		},
		{
			name:         "encoded slash %2F in dir name",
			urlPath:      "/with%2Fslash",
			expectedName: "with/slash",
		},
	}
	for _, tt := range tests {
			fsrv := FileServer{}
			bCtx := fsrv.directoryListing(
				context.Background(),
				nil,
				false,
				"/path/to/root",
				tt.urlPath,
				caddy.NewEmptyReplacer(),
			)
			actual := bCtx.Name
			if actual != tt.expectedName {
				t.Errorf("got %+v but expected %+v", actual, tt.expectedName)
			}
		})
	}
}
```

therefore I think changing this is also necessary:

```diff
-       name, _ := url.PathUnescape(urlPath)
+       name := unescapeSlash(path.Base(urlPath))
        return browseTemplateContext{
-               Name:     path.Base(name),
+               Name:     name,
```

or drop unescaping slashes altogether, because why slash get to be treated differently? Why don't encoded spaces also get decoded?

abdusco:
@francislavoie It's not clear to me why we need to decode percent-encoded chars (specifically slashes %2f)

https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/fileserver/browsetplcontext.go#L157-L160

https://github.com/caddyserver/caddy/blob/6bad878a22e048762262d6fabe2144cefaf4ca81/modules/caddyhttp/fileserver/browsetplcontext.go#L95

I think we should drop decoding percent-encoded chars altogether.

abdusco:
OK, I found why `url.PathUnescape` was added in the first place. I remember experiencing this issue as well: 

https://github.com/caddyserver/caddy/pull/4410

 

I think we can use a more polite approach and have:

```
decoded, err := url.PathUnescape(path)
if err != nil {
  decoded = path
}
```
@mohammed90 what do you think?

mohammed90:
> OK, I found why `url.PathUnescape` was added in the first place. I remember experiencing this issue as well:
> 
> #4410
> 
> I think we can use a more polite approach and have:
> 
> ```
> decoded, err := url.PathUnescape(path)
> if err != nil {
>   decoded = path
> }
> ```
> 
> instead of removing url decoding, which would bring back the old issue.
> 
> @mohammed90 what do you think?

I think that's very reasonable. Sorry, I don't have much time on hand to validate the displayed text. If you don't mind, have a few directories/files with various non-ASCII characters to double-check.

mholt:
Thanks for contributing this! Sorry I'm just now getting to it.

It seems to me that if PathUnescape is failing because of `%` not being escaped, the path has already been unescaped. :thinking:

While I think this is probably a fine solution, a more elegant one would be to only escape the path once.

I'm having trouble testing locally this since the code that added PathUnescape() was to fix when directories contain "/" in their names, which is impossible on Linux. And I run Linux/Mac, not Windows.

Would it be too tedious to find a solution that doesn't involve special-casing things, and instead we just Unescape once?

PR #5147, opened 2022-10-14
Support for OTEL_PROPAGATORS environment variable

krak3n:
Currently the propagators are hard coded to W3C TraceContext and Baggage within the tracing module [here](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/tracing/tracer.go#L72).

This makes it difficult to integrate Caddy into existing systems that do not use W3C TraceContext or Baggage for propagating traces.

This PR adds support to the `caddyhttp/tracing` module to support the OpenTelemetry `OTEL_PROPAGATORS` environment variable which allows users to configure the propagators used to extract and inject trace data from incoming and outgoing requests.

For example, users will now be able to define their propagators like this:

```sh
export OTEL_PROPAGATORS=b3,b3multi,baggage
```

As defined [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration) on the OpenTelemetry SDK configuration document.

This is achieved by using the `go.opentelemetry.io/contrib/propagators/autoprop` propagator which also uses the W3C TraceContext and Baggage as the default if no propagators are set on the `OTEL_PROPAGATORS` environment variable, see [here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c011266da033603c104472184510ca2d691dfb38/propagators/autoprop/propagator.go#L63-L65)


mholt:
Thanks for the PR. I will need to check on something before we merge this, as the otel dependency is a bit tricky for some of our users.

starkers:
> Thanks for the PR. I will need to check on something before we merge this, as the otel dependency is a bit tricky for some of our users.

Any progress on this @mholt ? support for b3 (or others) is quite important

mholt:
We can merge this, but the lint errors need fixing. @krak3n would you be able to address that?

mholt:
I don't think those linter errors make any sense. That's got to be a bug. The code shouldn't even be compiling if some of those are true.

I'll merge this and cross fingers. Thanks @krak3n !

starkers:
I'll be testing this today, maybe prod next week ;-)

krak3n:
Awesome thank you @mholt 🎉

PR #5154, opened 2022-10-18
fix: support known tls placeholders replacing even for non tls request

nikolaevn:
This Caddyfile produce incorrect upstreams headers (placeholders not replaced at all)
```
http://localhost:1337 {
	reverse_proxy / {
		to http://localhost:1339
		header_up XXX1 {http.request.tls.client.certificate_der_base64}
		header_up XXX2 {http.request.tls.client.san.uris.0}
	}
}
```

Without this PR upstream gets following:
```
GET / HTTP/1.1
Host: localhost:1337
X-Forwarded-For: 127.0.0.1
X-Forwarded-Host: localhost:1337
X-Forwarded-Proto: http
Xxx1: {http.request.tls.client.certificate_der_base64}
Xxx2: {http.request.tls.client.san.uris.0}
```

with PR:
```
GET / HTTP/1.1
Host: localhost:1337
X-Forwarded-For: 127.0.0.1
X-Forwarded-Host: localhost:1337
X-Forwarded-Proto: http
Xxx1: 
Xxx2: 
```

### DIFF
--- a/modules/caddyhttp/replacer.go
+++ b/modules/caddyhttp/replacer.go
@@ -335,33 +335,29 @@ func getReqTLSReplacement(req *http.Request, key string) (any, bool) {
 			var fieldValue any
 			switch {
 			case strings.HasPrefix(field, "dns_names"):
+				if !isTLSReq {
+					return nil, true
+				}
 				fieldName = "dns_names"
-				if isTLSReq {
-					fieldValue = cert.DNSNames
-				} else {
-					return nil, true
-				}
+				fieldValue = cert.DNSNames
 			case strings.HasPrefix(field, "emails"):
+				if !isTLSReq {
+					return nil, true
+				}
 				fieldName = "emails"
-				if isTLSReq {
-					fieldValue = cert.EmailAddresses
-				} else {
-					return nil, true
-				}
+				fieldValue = cert.EmailAddresses
 			case strings.HasPrefix(field, "ips"):
+				if !isTLSReq {
+					return nil, true
+				}
 				fieldName = "ips"
-				if isTLSReq {
-					fieldValue = cert.IPAddresses
-				} else {
-					return nil, true
-				}
+				fieldValue = cert.IPAddresses
 			case strings.HasPrefix(field, "uris"):
-				fieldName = "uris"
-				if isTLSReq {
-					fieldValue = cert.URIs
-				} else {
+				if !isTLSReq {
 					return nil, true
 				}
+				fieldName = "uris"
+				fieldValue = cert.URIs
 			default:
 				return nil, false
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/replacer.go
```diff
@@ -335,16 +336,32 @@ func getReqTLSReplacement(req *http.Request, key string) (any, bool) {
 			switch {
 			case strings.HasPrefix(field, "dns_names"):
 				fieldName = "dns_names"
-				fieldValue = cert.DNSNames
+				if isTLSReq {
+					fieldValue = cert.DNSNames
+				} else {
+					return nil, true
+				}
```
francislavoie, line 343: In general, return as early as possible. Don't use `else` if you can avoid it. This applies to many places in this PR, please apply this idea throughout.

```suggestion
				if !isTLSReq {
					return nil, true
				}
				fieldName = "dns_names"
				fieldValue = cert.DNSNames
```

modules/caddyhttp/replacer.go
```diff
@@ -387,49 +404,101 @@ func getReqTLSReplacement(req *http.Request, key string) (any, bool) {
...
-				return fmt.Sprintf("%x", sha256.Sum256(pubKeyBytes)), true
-			}
-			return fmt.Sprintf("%x", pubKeyBytes), true
 		case "client.issuer":
-			return cert.Issuer, true
+			if isTLSReq {
+				return cert.Issuer, true
+			} else {
+				return "", true
+			}
```
francislavoie, line 433:
```suggestion
			if !isTLSReq {
				return "", true
			}
			return cert.Issuer, true
```

modules/caddyhttp/replacer.go
nikolaevn: fixed all same parts

mholt:
Thanks for the PR!

This looks good, but before merging I just want to double check if this is the way we want it to work.

We could alternatively simply change `repl.ReplaceKnown()` (in headers.go) to `repl.ReplaceAll()` -- assuming that header fields don't otherwise have `{...}` in them. That's why I used ReplaceKnown up to now.

So, this is probably the _safest_ way to go but it is a bit repetitive (which is fine, as I don't know another way to do it). Any thoughts? (Don't change your patch though.)

francislavoie:
I think that changes in https://github.com/caddyserver/caddy/pull/5275 will affect what needs to be done here, once that one is completed.

mholt:
Another idea I just had is, if the connection isn't TLS, the replacer can return `nil, true` for the TLS-related placeholders, indicating that the placeholder keys are recognized, but that they have no value.

PR #5156, opened 2022-10-18
Clarified how destination values should be formatted.

smebberson:
This is a really basic PR that clarifies that destination values (as used with Map) should be wrapped in braces.


mholt (approved):
This is a great idea, thank you!

PR #5160, opened 2022-10-20
caddytest: Increased sleep between retries to reduce flakey tests in CI

ben-bud:
Tests were failing on CI runner, increasing sleep time between polling made tests less flaky


mohammed90:
> Tests were failing on CI runner, increasing sleep time between polling made tests less flaky

Can you share details about your CI infra? We only see the failure on the IBM machines. I'd love to understand how increasing the delay helps the flakiness in your case, given that we only return nil error if the fetched config matches the POSTed config.

arun-niranjan-bud:
> 

An example flake failure from our test suite
```
Failed
=== RUN   TestXRoutes
{"level":"info","ts":1666875010.3560665,"msg":"caddytest: config load time took 3.851942ms"}
    caddytest.go:223: POSTed configuration isn't active
    caddytest.go:103: failed ensuring config is running: EnsureConfigRunning: POSTed configuration isn't active
    caddytest.go:449: Content-Type header not provided
{"level":"info","ts":1666875010.4678953,"msg":"caddytest: redirecting the dialer from localhost:3000 to 127.0.0.1:3000"}
    caddytest.go:458: failed to call server Post "http://localhost:3000/v1/X": dial tcp 127.0.0.1:3000: connect: connection refused
    caddytest.go:128: unable to read the current config
--- FAIL: TestXRoutes (0.12s)
```

These were running on CircleCI and flaking - flake rate reduces when we either reduce load on the test binary or increase timeouts, with no change to the caddy files or endpoint redirection.

mohammed90 (approved):
Hopefully this works swimmingly now. Thank you!

PR #5161, opened 2022-10-21
caddyhttp: add placeholder {http.request.orig_uri.path.*}

XYenon:
Add placeholder `{http.request.orig_uri.path.*}` just like `{http.request.uri.path.*}`.


mholt (approved):
I guess this could be useful. Thanks for the contribution!

PR #5167, opened 2022-10-24
fileserver: Reject non-GET/HEAD requests (close #5166)

### LINKED ISSUE #5166: Send 501 responses as a fallback to unknown HTTP methods
Caddy's default behavior appears to be treating unhandled HTTP request methods as if they were GET. This is a very bad idea: it seriously impedes the ability to add new HTTP methods, which is terrible for the Internet as a whole.

Caddy SHOULD be producing 405 responses to any method it recognizes but does not allow. This requires populating the "Allow" header, which could be difficult to do automatically, so I can understand reluctance on that front.

501 responses (which are explicitly meant for unknown methods) have no such requirement, and so are trivial to implement. For now, I'm using something along the lines[^1] of this to approximate that:
```
@unsupported {
        not method GET HEAD POST PUT DELETE CONNECT OPTIONS TRACE
}

error @unsupported 501
```

This is obviously rather messy. It would be much better if Caddy did this on its own, especially since I doubt many users actually read RFC 9110.

[^1]: What I'm actually doing is matching methods I allow separately, so I can configure 405 responses for the methods Caddy knows about but disallows.

mholt:
File server does not mutate the files, so reject mutative methods. Accept only GET and HEAD (which is what the std lib `ServeContent()` knows).


Saklad5:
Hold on, are you adding the `Allow` header as required?

mholt:
Yes, but the commit didn't get pushed.

PR #5173, opened 2022-10-27
admin: set certmagic cache logger

bit:
same way it is set in modules/caddytls/tls.go


mholt:
Thanks for the PR! I am working on some other aspects of the project lately but I will get back to this ASAP.

mholt (approved):
Thank you for this! Sorry for the delay.

PR #5176, opened 2022-10-27
caddyhttp: Canonicalize header field names

### LINKED ISSUE #5175: http.handlers.static_response fails to canonicalize header field names
When using `http.handlers.static_response` to define additional http response headers the header names are not canonicalized when preparing the headers.
This leads to a duplicate content-type header in the response.
I've created a PR to add a call to `textproto.CanonicalMIMEHeaderKey()` to fix this.

### config.json with lower-case `content-type`

```json
{"apps":{"http":{"servers":{"srv0":{"listen":[":443"],"routes":[{"match":[{"host":["localhost"]}],"handle":[{"handler":"subroute","routes":[{"handle":[{"body":"Hello, world!","handler":"static_response","headers":{"content-type":["text/html"]}}]}]}],"terminal":true}]}}}}}
```
### headers received:
```
HTTP/1.1 200 OK
Alt-Svc: h3=":443"; ma=2592000
Content-Length: 13
Content-Type: text/plain; charset=utf-8, text/html
Date: Thu, 27 Oct 2022 16:19:58 GMT
Server: Caddy
```
### headers expected:
```
HTTP/1.1 200 OK
Alt-Svc: h3=":443"; ma=2592000
Content-Length: 13
Content-Type: text/html
Date: Thu, 27 Oct 2022 16:19:58 GMT
Server: Caddy
```

ghostwheel42:
This fixes #5175 by canonicalizing the header field names.

### DIFF
--- a/modules/caddyhttp/staticresp.go
+++ b/modules/caddyhttp/staticresp.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/textproto"
 	"os"
 	"strconv"
 	"strings"
@@ -193,7 +194,7 @@ func (s StaticResponse) ServeHTTP(w http.ResponseWriter, r *http.Request, next H
 
 	// set all headers
 	for field, vals := range s.Headers {
-		field = repl.ReplaceAll(field, "")
+		field = textproto.CanonicalMIMEHeaderKey(repl.ReplaceAll(field, ""))
 		newVals := make([]string, len(vals))
 		for i := range vals {
 			newVals[i] = repl.ReplaceAll(vals[i], "")


mholt (approved):
Thanks for the patch!

I just realized this is very similar to https://github.com/caddyserver/caddy/commit/2153a81ec85da99dcd33aa87ff0df5d286f00e9d.

francislavoie:
Seems like this is just a fix for JSON, because the headers don't get set through `.Add()` and such as they do when unmarshaling the Caddyfile.

Should this be done at provision time instead of at runtime, to save some cycles?

mholt:
@francislavoie That's a good idea -- I thought about it too, but it has to be done on the output of the replacer, which depends on the request. So I think we have to do it here.

ghostwheel42:
@francislavoie I've thought about using the appropriate methods for this and/or doing it when provisioning, but as @mholt said it needs to be done on the output of the replacer and to use `.Add()` you'd need to iterate over the list of values which would probably be slower in the end.

francislavoie (approved):

PR #5177, opened 2022-10-29
Update adminapi.go

RohitYandigeri:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=ozoGNik)
The caddy.Context.AppIsConfigured() method returns true if the named app is present in the config. This is useful to know if an app has been explicitly configured, because calling App() will instantiate an app module even if it's not configured.

However, if the primary use case is to get an app instance only if it has been configured, then maybe we should actually have a method, AppIfConfigured() instead. One-letter change.

This method would return nil if the app is not configured, so if you need the bool value, there you have it; otherwise it returns the configured app module. It would be less clunky than this:


mohammed90:
Thanks for the attempt, but the method does not exist, the change causes build failure, and all tests are failing. This isn't a net-positive change.

PR #5179, opened 2022-10-31
ci: set least privilged token for github actions

boahc077:
This PR sets minimum token permission for all workflow files in the repository. Currently, GitHub token has elevated access as demonstrated by the following GitHub Actions workflow run that was executed recently:
https://github.com/caddyserver/caddy/actions/runs/3352862992/jobs/5555290545#step:1:19

In addition to this PR, you should also consider setting the following permission so that new workflow files will default to read-only permissions:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,9 @@ on:
       - master
       - 2.*
 
+permissions:
+  contents: read
+
 jobs:
   test:
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,9 +12,6 @@ on:
       - master
       - 2.*
 
-permissions:
-  contents: read
-
 jobs:
   test:
     strategy:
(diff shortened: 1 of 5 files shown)


francislavoie:
IIRC we tried that, but it broke the release job since it needs to write to github releases. But I don't quite remember the details.

mohammed90 (changes requested):
Francis is right. Write permission is need because we're publishing artifacts and posting release notes. Perhaps the only job where read-only is possible is the lint flow. I'd rather we don't touch the release flows without seeing evidence it won't break when the requested change is applied.

boahc077:
Thanks @francislavoie and @mohammed90 for reviewing the PR! I have reverted the changes for all other workflow files. Please take a look.

mohammed90 (approved):
This is great! Thank you for this enhancement

Issue #5181, opened 2022-11-01
certificate_authorities/root is ignored while renewing intermediate certificate


joga84:
while renewing the intermediate certificate, the path of the root certificate is ignored. instead a new root certificate is created at the default location. If root is specified it should be used all the time.
right now ca.loadOrGenRoot does not return the configured root but just the default one. this function is however used in several places to get the root certificate.

```patch
diff --git a/modules/caddypki/ca.go b/modules/caddypki/ca.go
index 41e78a6b..acf8d14d 100644
--- a/modules/caddypki/ca.go
+++ b/modules/caddypki/ca.go
@@ -240,6 +240,9 @@ func (ca *CA) NewAuthority(authorityConfig AuthorityConfig) (*authority.Authorit
 }
 
 func (ca CA) loadOrGenRoot() (rootCert *x509.Certificate, rootKey any, err error) {
+       if ca.Root != nil {
+               return ca.Root.Load()
+       }
        rootCertPEM, err := ca.storage.Load(ca.ctx, ca.storageKeyRootCert())
        if err != nil {
                if !errors.Is(err, fs.ErrNotExist) {
```

we could return the configured certificate in case it is configured, does that look like the right approach or should there be another wrapper around loadOrGenRoot and ca.Root.Load?

example configuration to trigger the issue:

```json
    "pki": {
      "certificate_authorities": {
        "local": {
          "root": {
            "certificate": "/etc/ssl/ca/root.crt",
            "format": "pem_file",
            "private_key": "/etc/ssl/ca/root.key"
          },
        }
      }
    },
```

mholt:
Thanks for the patch! (Sorry I was away while traveling.)

KennyStier:
I need this feature in a project I am working on. Will this fix be included in a new release soon, or should I work on building Caddy myself?

francislavoie:
We are planning to make a release soon, but if you're in a hurry you can definitely build it yourself. It's very easy to do so with `xcaddy`. You just need the latest version of Go installed (download and unzip it, add to your PATH) and download `xcaddy` and run `xcaddy build master` and it'll produce a binary in your $PWD.

PR #5189, opened 2022-11-08
caddypki: use configured certificate instead of generating a new one

joga84:
instead of generating a new root certificate at the default location load the certificate from the configuration.
fixes: #5181


mholt:
Thank you! I guess this change makes sense: `loadOrGenRoot()` should probably load the user-specified root if there is one, rather than always using the generated one.

mholt (approved):

PR #5192, opened 2022-11-09
caddyauth: improve handling of multiple providers

### LINKED ISSUE #5190: caddyauth: authentication handler with multiple providers results in headers and status code of random provider
I have a setup where I have multiple authentication providers configured:
```json
{
  "handler": "authentication",
  "providers": {
    "http_basic": {
      "accounts": [
        {
          "password": "$2y$05$sS98izYCD32TDCH0bHVrG.t2nXNX.ms4fvRjLKXP6LgcpAJ18CI1a",
          "username": "test"
        }
      ]
    },
    "authorizer": {
      "gatekeeper_name": "auth_policy",
      "route_matcher": "*"
    }
  }
}
```
The `authorizer` is from the caddy-security plugin, it handles the OAuth login etc. If the user is not logged in the user will be redirected.

The expectation of this configuration is that I can either use OAuth (via `authorizer` provider) or Basic Auth (via `http_basic` provider) to login and to pass the `authentication` handler. The problem is that the plugins have side effects on the response. For example when I use basic auth I get back the some of the Header and the status code of the `authorizer` plugin but also the expected content from a handler that is after the `authentication` handler.

Looking at the code it becomes clear why this happens:
https://github.com/caddyserver/caddy/blob/6efd1b3bb1217841269e67930cba33992fb96930/modules/caddyhttp/caddyauth/caddyauth.go#L71-L76
We pass the same `http.ResponseWriter` to all authentication providers. In my case the `authorizer` plugin sets the status code to 302 and sets the location to redirect in the header. Essentially calling `http.ResponseWriter.WriteHeader` seals the header from being written by any other plugin. The order of the provisioners, depends on the map order which is random.

Potential solution:
I think we should pass separate `http.ResponseWriter` instances to each provider, if any provider succeeds all of them can be discarded.

The question is what should if none of the providers successfully authenticates/authorizes the user. Some of the providers may rely on redirects, if there are multiple I think there needs to be an explicit priority, if only one wants to redirect, the redirect could take precedence over the other statuses.

I would be happy to provide a PR to fix this, if there is agreement on the approach on how to handle this.

lukedirtwalker:
Improve the authentication handler in case multiple authentication providers are set. Currently if multiple providers are present and at least one of them fails to authenticate it could be that response contained a mix of multiple providers (and potentially the next handler).

This commit changes the behavior in the following way: In case one of the providers succeeds all information of the providers is discarded and the next handler is called. In case all providers fail to authenticate the user, the response of any provider that redirects is written. In case no provider redirects the response of any provider is written.

Currently there is no way to require all authentication providers to successfully authenticate, but this behavior wasn't possible in the previous implementation either.

Fixes #5190

### DIFF
--- a/modules/caddyhttp/caddyauth/caddyauth.go
+++ b/modules/caddyhttp/caddyauth/caddyauth.go
@@ -15,6 +15,7 @@
 package caddyauth
 
 import (
+	"bytes"
 	"fmt"
 	"net/http"
 
@@ -38,7 +39,11 @@ func init() {
 // Its API is still experimental and may be subject to change.
 type Authentication struct {
 	// A set of authentication providers. If none are specified,
-	// all requests will always be unauthenticated.
+	// all requests will always be unauthenticated. If multiple
+	// providers are specified only one needs to authenticate
+	// successfully. In case all of them are unauthorized the reply
+	// will be taken from any provider that redirects, if there is
+	// no provider that redirects the reply of any provider is taken.
 	ProvidersRaw caddy.ModuleMap `json:"providers,omitempty" caddy:"namespace=http.authentication.providers"`
 
 	Providers map[string]Authenticator `json:"-"`
@@ -72,8 +77,21 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 	var user User
 	var authed bool
 	var err error
+	// This code makes the following assumption: In case of multiple
+	// authentication providers it is enough if one of them authenticates the
+	// user for the whole authentication to succeed. In case all of the
+	// providers fail to authenticate we write the reply of any of the providers
+	// that redirect to the response writer, if there is no such provider the
+	// response of any provider is written.
+	responseWriters := make(map[string]caddyhttp.ResponseRecorder, len(a.Providers))
 	for provName, prov := range a.Providers {
-		user, authed, err = prov.Authenticate(w, r)
+		// We always want to buffer the response because only once we have the
+		// reply of all providers we can decide which one to actually use (if
+		// any). Usually auth replies should be small enough.
+		alwaysBuffer := func(status int, header http.Header) bool { return true }
+		rw := caddyhttp.NewResponseRecorder(w, new(bytes.Buffer), alwaysBuffer)
+		responseWriters[provName] = rw
+		user, authed, err = prov.Authenticate(rw, r)
 		if err != nil {
 			a.logger.Error("auth provider returned error",
 				zap.String("provider", provName),
@@ -85,8 +103,25 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 		}
 	}
 	if !authed {
+		// if we have any redirect we use the result from that.
+		for _, rw := range responseWriters {
+			if rw.Status() >= 300 && rw.Status() < 400 {
+				rw.WriteResponse()
+				return caddyhttp.Error(http.StatusUnauthorized, fmt.Errorf("not authenticated"))
+			}
+		}
+		// no redirect choose a random reply.
+		for _, rw := range responseWriters {
+			rw.WriteResponse()
+			break
+		}
 		return caddyhttp.Error(http.StatusUnauthorized, fmt.Errorf("not authenticated"))
 	}
+
+	// In case authentication was successful we don't care about any response
+	// from the authentication handlers. We already nil the responseWriters here
+	// so the memory could be freed.
+	responseWriters = nil
 
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 	repl.Set("http.auth.user.id", user.ID)
--- a/modules/caddyhttp/caddyauth/caddyauth.go
+++ b/modules/caddyhttp/caddyauth/caddyauth.go
@@ -44,6 +44,8 @@ type Authentication struct {
 	// successfully. In case all of them are unauthorized the reply
 	// will be taken from any provider that redirects, if there is
 	// no provider that redirects the reply of any provider is taken.
+	// If you want multiple authentication providers that all need to succeed
+	// you can create a chain of handlers.
 	ProvidersRaw caddy.ModuleMap `json:"providers,omitempty" caddy:"namespace=http.authentication.providers"`
 
 	Providers map[string]Authenticator `json:"-"`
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/caddyauth/caddyauth.go
```diff
@@ -72,8 +77,21 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 	var user User
 	var authed bool
 	var err error
+	// This code makes the following assumption: In case of multiple
+	// authentication providers it is enough if one of them authenticates the
+	// user for the whole authentication to succeed. In case all of the
+	// providers fail to authenticate we write the reply of any of the providers
+	// that redirect to the response writer, if there is no such provider the
+	// response of any provider is written.
```
mholt, line 85: Why do we favor responses with redirects? Why do they get special priority over other responses?

Just want to hash this out, in case it's prudent to have some sort of preference system instead. (I don't mind having implicit simplicity -- as long as it is backed by reason, I just want to understand it. :slightly_smiling_face: )

mholt (commented):
Thanks for the patch! So far it's looking good to me but I just have a couple questions (see below).

As for "AND" vs. "OR" across auth providers, right now it's OR, and maybe that's the right way to do it; multiple `authentication` handlers can be chained into a route to enforce AND. (Right?)

lukedirtwalker:
Thanks for the review:
> As for "AND" vs. "OR" across auth providers, right now it's OR, and maybe that's the right way to do it; multiple authentication handlers can be chained into a route to enforce AND. (Right?)

True that's actually a good point for the AND semantics we can simply chain the handlers.

modules/caddyhttp/caddyauth/caddyauth.go
```diff
@@ -85,9 +103,34 @@ func (a Authentication) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 		}
 	}
 	if !authed {
+		// if we have any redirect we use the result from that.
+		for provName, rw := range responseWriters {
+			if rw.Status() >= 300 && rw.Status() < 400 {
```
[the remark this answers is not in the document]
lukedirtwalker: I'll add that yes.

modules/caddyhttp/caddyauth/caddyauth.go
lukedirtwalker: So maybe that is specific to our use case, but the reasoning is as follows; The provider that redirects is the provider that does an OAuth flow and for that flow it is natural that the user get's redirected. E.g. you browse to a page and then you need to login, it redirects you to some identity provider to login, you get back to the page. If we wouldn't favor the redirecting provider, this flow would break.

But I do realize that this might be a bit magic and might even be confusing, e.g. in the case the user actually provider basic auth credentials that are wrong and then gets redirected to the identity provider login page.
Also in case the user agent is not a browser you likely don't want the redirect.

So overall I'm not really sure what's best, I'm also open to suggestions here.

mholt:
I'll give this some thought, to see if there's a good way to handle that.

modules/caddyhttp/caddyauth/caddyauth.go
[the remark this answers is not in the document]
lukedirtwalker: Added this now.

modules/caddyhttp/caddyauth/caddyauth.go
lukedirtwalker: Also added the reasoning in the comment.

mholt:
@lukedirtwalker In that case, might be best to write a custom wrapper.

As for the multiple providers problem (if none return a positive result) I wonder if we should honor the first one specified in the config (if possible), as like the preferred provider, or something.

lukedirtwalker:
>As for the multiple providers problem (if none return a positive result) I wonder if we should honor the first one specified in the config (if possible), as like the preferred provider, or something.

I don't think that's possible right now, as it is a map currently. Converting it to a list would be a breaking change. We could however add an additional order field. If that is not provided we keep the random behavior, if it's set we order by the preference? What do you think? Any suggestion on the naming? `priority` ?

mholt:
@lukedirtwalker There might be an awful way to do it by implementing our own `UnmarshalJSON` that extracts the keys in order, or at least keeps track of the order of each key. But yeah, relying on ordering of a map that isn't conventionally ordered seems slippery.

Maybe priority is a better option! I like that.

lukedirtwalker:
@mholt sorry for the long silence. I'm still not 100% clear on how to implement this.

I considered adding a `Priority() int` method to the `Authenticator` interface. We can't really add the method as that would break all implementations, so we could do interface smuggling and just check if an `Authenticator` implements the `Priority` method. However this would still mean that all Authenticator implementation would need to change to support the priority.

This lead me to thinking that we should probably rather somehow have an extra field on the authentication handler: https://caddyserver.com/docs/json/apps/http/servers/errors/routes/handle/authentication/ named `priority` which would then need to match the keys in the `providers` map and is a list given the order. However I'm not 100% how I would model this correctly. What I thought is to add the following field:
```go
type Authentication struct {
	ProvidersRaw caddy.ModuleMap `json:"providers,omitempty" caddy:"namespace=http.authentication.providers"`
	// NEW FIELD:
	Priority []string `json:"priority,omitempty" caddy:"???"`
}
```
Does that approach sound reasonable? What do I need to put for this to work correctly with the caddy file? What else do I need to change?

mholt:
Apologies for my silence as well :smile:

I appreciate the thought you're putting into this. Let me see if I can find some time to put an equal amount of effort into thinking of a good solution...

But so far I like your approach very much. :+1:

In fact, I like it enough that I'd be happy to review a PoC for it!

(Don't worry about Caddyfile for now -- it's easy, and I can show you when I have time later.)

mholt:
After thinking about this more and talking with @francislavoie I wonder if there's a slightly better approach and I would like your thoughts:

Instead of a separate priority field, what if an auth plugin wrapped multiple auth providers? They could be specified in an array/slice instead of an object/map, which makes them ordered, and thus the priority is implicit in the order in which they appear in the config. Plus, it gives the ability to expose more configuration to compose multiple auth providers.

Does that make sense? I like where this is going!

lukedirtwalker:
I like your suggestion very much.
But then I would recommend to drop the option to provide multiple auth plugins in Caddy itself, otherwise other users might stumble over this limitation in the future. However this again will be a breaking change.

In any case I can simply create a plugin and we can close this PR here. If we want to have a solution for the above issue, we should probably create a different PR anyway.

francislavoie:
I think we should probably at minimum update the godoc comment to clarify the limitations of the array syntax.

I think the multi-auth usecase is ultimately pretty niche, and there only exists one auth module in vanilla Caddy anyway, so having the multi module outside as a plugin is fine. But I don't think it would be a problem if you would like it included in Caddy, because it's probably not a lot of code and it's without dependencies. So, it's up to you.

Collins2000 (commented):
$400000

PR #5217, opened 2022-11-23
Use go.step.sm/crypto to generate the PKI

maraino:
This commit replaces the use of github.com/smallstep/cli to generate the root and intermediate certificates and uses go.step.sm/crypto instead.

It also upgrades the version of github.com/smallstep/certificates to the latest version.


modules/caddypki/ca.go
```diff
@@ -239,10 +239,11 @@ func (ca *CA) NewAuthority(authorityConfig AuthorityConfig) (*authority.Authorit
 	return auth, nil
 }
 
-func (ca CA) loadOrGenRoot() (rootCert *x509.Certificate, rootKey any, err error) {
+func (ca CA) loadOrGenRoot() (rootCert *x509.Certificate, rootKey crypto.Signer, err error) {
 	if ca.Root != nil {
 		return ca.Root.Load()
 	}
+	fmt.Printf("%#v\n", ca.storage)
```
mholt, line 246:
```suggestion
```

mholt (approved):
Wow!! Christmas came early. :gift: You guys are the best! Thank you for the help. :smiley: 

This is such a treat, thank you again. Changes LGTM. It's nice to have your expertise in updating this, since I don't think I would have figured that out. :sweat_smile: 

Thank you to @mohammed90 for taking the initiative on this too.

Smallstep is amazing!

PR #5232, opened 2022-12-05
acme_server: Allow the lifetime for issued certificates to be configured

kylemcc:
Similarly to how the internal tls module allows the certificate lifetime to be configured, this change allows the lifetime certificates issued by the embedded ACME server to be configured.


kylemcc:
Happy to update the docs as well once this is released!

mholt:
Oh perfect, that's my favorite :grin:  Thanks

mholt:
Oh yes, good point. This is a potential footgun...

mholt:
Thinking on that, what's the exact use case for this? Perhaps it only makes sense to make the intermediate _and_ the leaf cert lifetimes configurable, rather than just the leaf.

(Also, leaf cert lifetime shouldn't get too close to intermediate cert lifetime.)

kylemcc:
> Thinking on that, what's the exact use case for this? Perhaps it only makes sense to make the intermediate _and_ the leaf cert lifetimes configurable, rather than just the leaf.
> 
> (Also, leaf cert lifetime shouldn't get too close to intermediate cert lifetime.)

Yes, fair question. The use case for us is entirely operational. We're looking at Caddy to issue certs internally (thanks!). Some of the services using those certs don't natively support automatically request certs via ACME, so we're using an external CLI tool to request and install them.

The default 12-hour lifetime is totally fine from a technical standpoint. However, if we fail to renew or install a certificate for some reason, there's a good chance someone is getting paged off-hours to deal with an expired cert. Being able to configure, say, a 7-day or 30-day lifetime gives us some leeway to deal with the failure of an automated renewal without being on such a tight clock.

Collins2000:
ALL GOOD

kylemcc:
While you're thinking about this, I've prepared an additional change that allows the intermediate cert lifetime to be configured (`pki.intermediate_lifetime`). It also includes checks for:
- the intermediate cert lifetime < root cert lifetime
- leaf cert lifetime < intermediate lifetime

Happy to push that up if that still seems like the right approach.

mholt (approved):
Thank you @kylemcc -- that is feeling much better.

It just occurred to me, CSRs can request a certificate have a certain validity period; I don't know of any CAs that honor that, nor do I even know if Caddy's (@Smallstep's) does :dizzy_face:.

I wonder if we should investigate whether the ACME server supports lifetimes specified by the CSR, and if so, in a follow-up patch it would make sense to treat these config values as defaults if the CSR doesn't specify.

kylemcc:
> I wonder if we should investigate whether the ACME server supports lifetimes specified by the CSR, and if so, in a follow-up patch it would make sense to treat these config values as defaults if the CSR doesn't specify.

Ah, @mholt, this is interesting. Looking at [the spec](https://datatracker.ietf.org/doc/html/rfc8555#section-7.4), it describes `notBefore` and `notAfter` fields that can be specified in the new-order request.

Empirically, none of the ACME clients I've seen (which is obviously not all of them) expose that option. So, I decided to modify [lego](https://github.com/go-acme/lego) to allow `notAfter` to be specified. When requesting a cert from Let's Encrypt, I received the following response:

```
acme: error: 400 :: POST :: https://acme-v02.api.letsencrypt.org/acme/new-order :: urn:ietf:params:acme:error:malformed :: NotBefore and NotAfter are not supported
```

So, Let's Encrypt doesn't support it... However, when testing against Caddy, the parameter works and is respected. I was able to modify expiration date of the issued cert by changing the `NotAfter` parameter. It works quite nicely in conjunction with the changes here.

The only minor issue, which is unrelated to the changes here, is that requesting a cert with a lifetime longer than the intermediate cert returns a 500 instead of a 400 with a nice error message. The error is coming from the smallstep API, so not Caddy's fault. Seems like an opportunity for better validation and messaging in smallstep, which I intend to address 😊

mholt:
Oh, I bet @maraino and the team would know a thing or two about that (roadmap, nuances, etc) ^ ^

Anyway, that's really good to know. Thanks for going to all the work to test that out!

mholt:
Thank you for the high-quality contribution and investigation as well. :+1: Please feel free to contribute again sometime!

What does your company do that you're using Caddy for it? Just curious :)

kylemcc:
Thanks very much for the prompt review - and for Caddy, @mholt! Glad I could contribute. I'll be sure to issue another PR to updates the docs today.

[My company](https://www.opshelm.com/) is building cloud security automation software. We're currently using Caddy to issue certs to internal services/infrastructure like NATS, and [Temporal](https://temporal.io/) clients and servers for pretty painless mTLS. I've used Caddy personally for years for various side projects with zero issues. I was very happy to discover the @smallstep integration while researching options for running an ACME-compatible CA internally.

mholt:
Awesome! Thanks for the info :D

Let me know if you'd be interested in talking about a sponsorship sometime. One can go a long way for benefiting both your business and the project!

mholt:
Your contribution shows it. Thank you again!

Issue #5202, opened 2022-11-11
Get placeholder array from index


davidpanic:
I would like to be able to get placeholder values like some programming languages allow to with "rest" arguments. Perhaps it's best explained with an example:

```Caddyfile
(myservice) {
    {args.0}.some.domain {
        reverse_proxy {args.1...} {
            ...a ton of options that I don't want to repeat
        }
    }
}

import myservice service.a http://192.168.60.1:8081 http://192.168.60.2:8081 http://192.168.60.3:8081
import myservice service.b http://192.168.60.4:8181 http://192.168.60.5:8181
```

As you can see there's a different amount of arguments (in this case upstream URLs) defined in the import and I want to just specify them all for the reverse proxy upstreams.

To my knowledge there is currently no way to do this, you must specify the arguments by index which doesn't work for variable length arguments.

francislavoie:
I've thought about implementing something like this before, but I couldn't think of a good way to do it.

It happens here in the codebase if you'd like to take a look:

https://github.com/caddyserver/caddy/blob/master/caddyconfig/caddyfile/parse.go#L345-L352

(And further down, it does ReplaceKnown to actually replace stuff.)

As for the syntax... I think using `...` doesn't look great because we already use a dot in between. It is similar to Go syntax so it has that going for it.

Could be something like `{args.1+}` where the plus means everything starting at arg 1 and after it. Might be more readable that way, I dunno.

I guess one way it could be implemented is by filling the replacer with a `+` variant for each entry that is the current arg and all the following ones, for each index provided. That could work.

davidpanic:
> I guess one way it could be implemented is by filling the replacer with a + variant for each entry that is the current arg and all the following ones, for each index provided. That could work.

Sounds easy enough to add, I'm all for it! I can implement it and make a PR if you'd like.

mholt:
After looking at this more, I'm wondering if we could keep the `:` but instead of the `.`, use square brackets?

So, `args[0]` and `args[-1]` and `args[:3]` instead of `args.0` or `args.-1` or `args.:3`.

francislavoie:
To elaborate on that, I think we can keep the old `{args.0}` syntax around for now, without variadic support. We can introduce the this new `{args[0]}` syntax which includes variadic support.

We can mark the old syntax as deprecated with a Caddyfile adapt warning maybe. But we could defer writing the warning until a later release to not be too annoying.

PR #5249, opened 2022-12-12
caddyfile: Implement variadics for import args placeholders

WeidiDeng:
imported snippets reflect actual lines in file

fix [5202](https://github.com/caddyserver/caddy/issues/5202)

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins_test.go
+++ b/caddyconfig/httpcaddyfile/builtins_test.go
@@ -215,32 +215,32 @@ func TestRedirDirectiveSyntax(t *testing.T) {
 	}
 }
 
-func TestErrorLine(t *testing.T) {
+func TestImportErrorLine(t *testing.T) {
 	for i, tc := range []struct {
 		input     string
 		errorFunc func(err error) bool
 	}{
 		{
 			input: `(t1) {
-					abort {args.:}
+					abort {args[:]}
 				}
 				:8080 {
 					import t1
 					import t1 true
 				}`,
 			errorFunc: func(err error) bool {
-				return err != nil && strings.Contains(err.Error(), "Caddyfile:6(import t1):2")
+				return err != nil && strings.Contains(err.Error(), "Caddyfile:6 (import t1):2")
 			},
 		},
 		{
 			input: `(t1) {
-					abort {args.:}
+					abort {args[:]}
 				}
 				:8080 {
 					import t1 true
 				}`,
 			errorFunc: func(err error) bool {
-				return err != nil && strings.Contains(err.Error(), "Caddyfile:5(import t1):2")
+				return err != nil && strings.Contains(err.Error(), "Caddyfile:5 (import t1):2")
 			},
 		},
 	} {
--- a/caddyconfig/httpcaddyfile/builtins_test.go
+++ b/caddyconfig/httpcaddyfile/builtins_test.go
@@ -1,6 +1,7 @@
 package httpcaddyfile
 
 import (
+	"strings"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


mholt:
Very cool, thanks for working on this!

We might need a test case for this, or at least an example that we know works. There's a number of test failures, currently.

francislavoie:
Hmm. I'm not sure how I feel about the `:` syntax. Yes obviously it's very Go-like, but I think it reads weird when following a `.` dot. I think we need to consider some other syntax options.

francislavoie:
Hmm, maybe. I think if we want to use the `[]` syntax, we should maybe make `{args[0]}` work as well (maybe deprecate `{args.0}`?)

mholt:
Personally, I'm ok with the `:` syntax :man_shrugging:

caddyconfig/httpcaddyfile/builtins_test.go
```diff
@@ -213,3 +214,45 @@ func TestRedirDirectiveSyntax(t *testing.T) {
...
 	}
 }
+
+func TestErrorLine(t *testing.T) {
+	for i, tc := range []struct {
+		input     string
+		errorFunc func(err error) bool
+	}{
+		{
+			input: `(t1) {
```
rudSarkar, line 224: Test duplicate import

caddyconfig/httpcaddyfile/builtins_test.go
```diff
@@ -213,3 +214,45 @@ func TestRedirDirectiveSyntax(t *testing.T) {
...
+				:8080 {
+					import t1
+					import t1 true
+				}`,
+			errorFunc: func(err error) bool {
+				return err != nil && strings.Contains(err.Error(), "Caddyfile:6(import t1):2")
+			},
+		},
+		{
+			input: `(t1) {
```
rudSarkar, line 236: Test invalid import

rudSarkar (commented):
Is it possible to add a name for each test case like the below example given?

```go
{
	name: "Test duplicate import",
	input: `(t1) {
		abort {args.:}
	}
	:8080 {
		import t1
		import t1 true
	}`,
	errorText: "Caddyfile:6(import t1):2",
},
```

rudSarkar (approved):

rudSarkar:
@mholt I think It will be better to add this on later commit.

caddyconfig/httpcaddyfile/builtins_test.go
francislavoie: This test is actually just to cover that the error message produced includes the `(import t1)` bit as an indicator to the user of where the problem is.

In both tests here, an error is always produced by the `abort` directive's parsing, because `abort` does not support any arguments at all.

The first import in this test does not produce an error, because the variadic placeholder doesn't produce any tokens, but the second import does produce tokens which causes the `abort` directive to error.

caddyconfig/httpcaddyfile/builtins_test.go
francislavoie: Explained in https://github.com/caddyserver/caddy/pull/5249#discussion_r1106925556, this second test just makes sure the error line is correct by being at least different than the other test.

francislavoie:
I worked on cleaning this up -- see recent commit for details.

mholt:
I think this is looking good!

I might wait for #5275 though, so as to not introduce another conflict. Thanks for working on this both of you :blush:

francislavoie:
I don't think they'll conflict. This only affects the Caddyfile and that PR affects runtime usage of the replacer.

mholt (approved):
You're right - on closer inspection the files do not overlap. Let's merge this and give it a try!

PR #5253, opened 2022-12-15
Fixes a reversion in c0f76e9 which caused the query-string to be removed from the redirect.

jaygooby:
You can test like this, with `curl`. The current caddy `v2.6.2` drops the query-string:

```
mkdir -p /tmp/caddy/example

caddy file-server --root /tmp/caddy --browse

curl http://127.0.0.1:80/example?query=preserved
<a href="/example/">Permanent Redirect</a>
```

but with the fix, the query-string is preserved:

```
mkdir -p /tmp/caddy/example

caddy file-server --root /tmp/caddy --browse

curl http://127.0.0.1:80/example?query=preserved
<a href="/example/?query=preserved">Permanent Redirect</a>.
```

### DIFF
--- a/modules/caddyhttp/fileserver/browse.go
+++ b/modules/caddyhttp/fileserver/browse.go
@@ -70,7 +70,8 @@ func (fsrv *FileServer) serveBrowse(root, dirPath string, w http.ResponseWriter,
 	if r.URL.Path == "" || path.Base(origReq.URL.Path) == path.Base(r.URL.Path) {
 		if !strings.HasSuffix(origReq.URL.Path, "/") {
 			fsrv.logger.Debug("redirecting to trailing slash to preserve hrefs", zap.String("request_path", r.URL.Path))
-			return redirect(w, r, origReq.URL.Path+"/")
+			origReq.URL.Path += "/"
+			return redirect(w, r, origReq.URL.String())
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


francislavoie:
We can't revert that commit. That commit fixed a security issue, and this PR re-introduces the security issue.

jaygooby:
Can you explain the security issue? I’ve not literally just reverted the code, but did take care to keep the return as per the fix - all I’ve done is ensure the query-string is preserved in the return, not just the path

jaygooby:
As with https://github.com/caddyserver/caddy/pull/5254#issuecomment-1353770422 I’ll be happy to work on this, to fix the issue and ensure that there are no security issues. 

Are there any test cases or more details that I can see?

mholt:
Sorry for the delay. I'll try to look at this soon. (Was backlogged, then had some issues with GH notifications not appearing, now more backlogged... etc.)

jaygooby:
Same here! I started on a bunch of test cases https://github.com/jaygooby/caddy/blob/redo-redirect-fixes/caddytest/integration/redirect_test.go hit some snags and then Xmas arrived. Need to pick it up again.

mholt:
Run `caddy file-server --browse` and then try your patch with this request:

```
$ curl -vL http://localhost//example.com/%2F..
```

I believe this is more of a client vulnerability than a server one, but I agree it's surprising behavior, so any patch for the query string needs to avoid the original behavior.

francislavoie:
I think this is also superseded by https://github.com/caddyserver/caddy/pull/6109 so I'm closing, but we can reopen if I'm incorrect

Issue #4205, opened 2021-06-14
Bug: Caddy v2.4.2 with specified rewrite rules causes unexpected 308 redirection


inoblue:
Phenomenon: Open web page on Firefox and Chrome, and they both say "the page cannot redirect correctly". After downgrading Caddy to v2.4.1, everything becomes OK.

Caddy configuation (Caddyfile):
```
${domain}
root * /path/to/html
encode zstd gzip
tls /path/to/cert.pem /path/to/key.pem {
  protocols tls1.3
  alpn h2 http/1.1
  curves x25519
}
header Strict-Transport-Security max-age=31536000
reverse_proxy /path/to/proxy 127.0.0.1:${port} {
  transport http {
    versions h2c
  }
}
@common not path /path/to/proxy
handle @common {
  rewrite * /path/to/rewrite.html
}
file_server
```

In Caddy's log, there are many lines like the one below:
```
{"level":"error","ts":1623667844.8768942,"logger":"http.handlers.reverse_proxy","msg":"aborting with incomplete response","error":"context canceled"}
```

mholt:
This seems to happen when there's a rewrite (correct?). I think we need to revert https://github.com/caddyserver/caddy/commit/f9b54454a19e2b070159ce8d2af76d819658244e.

@diamondburned Could you help take a look at this?

mholt:
For now, @inoblue, I've reverted what I believe to be the regression -- if you can build from source can you verify?

@diamondburned, maybe we can come up with a better solution to #4179 that doesn't cause a regression.

inoblue:
The issue was solved after building from source.

Thanks for your help.

diamondburned:
Are there any debug logs available with the broken commit?

francislavoie:
FWIW, what you were fixing _was_ a bug for both `browse` and non-`browse`. I've seen it come up on a couple topics on the forums where users wanted to use `file_server` within `handle_path` as well.

mholt:
The more I look at this and fiddle with it, the more I'm convinced that canonicalizing based on the _original request path_ is not the correct behavior (i.e. I was wrong in the forum thread where I suggested that; or at least, wrong in the implementation).

mholt:
@inoblue Can you please share your full config and curl commands and logs? I need to know your exact use case. I can tell your config is not real because it has lines like `rewrite * /path/to/rewrite.html` but this isn't helpful in us understanding the issue.

[...]

PR #5254, opened 2022-12-15
Preserve the query-string for non browse case

jaygooby:
when doing canonical redirect. Under caddy `v2.6.2` the query-string is lost.

This is related to https://github.com/caddyserver/caddy/pull/5253 which does the same thing when browsing is enabled, but does not fix it for the non-browsing case.

Test like this:

```
mkdir -p /tmp/caddy/example
echo "hello" >  /tmp/caddy/example/index.html

caddy file-server --root /tmp/caddy

curl "http://127.0.0.1/example?query=retained"
<a href="/example/">Permanent Redirect</a>.
```

with the fix in place, the query-string is correctly preserved:

```
curl "http://127.0.0.1/example?query=retained"
<a href="/example/?query=retained">Permanent Redirect</a>.
```


francislavoie:
`origReq` should not be manipulated directly. That can cause side effects. Instead, the URL should be cloned, and that clone can be manipulated.

jaygooby:
I’m happy to work on this, to get the correct behaviour. Are there a set of test cases that show requests that can cause open redirects?

mholt:
Hi Jay, I'll take a look at this as soon as I can and try to figure something out together :)

jaygooby:
Thanks @mholt I'm a golang newbie. `origReq` is assigned from the request, and then I'm returning a string representing the final URL, so it didn't seem like it would be dangerous operation 😅

mholt:
We might be able to safely append the query string to the `to` variable if there is one.

@jaygooby Armed with the docs for https://pkg.go.dev/net/url#URL, think you can try it out?

jaygooby:
Thanks for the docs link! Yup, I'll give it a go later today

mholt:
Awesome. FWIW if you go to [the git history for that file](https://github.com/caddyserver/caddy/commits/2b8d1b1517f98da775465b59e0372a8d9efbd027/modules/caddyhttp/fileserver/staticfiles.go), this commit shows up:

https://github.com/caddyserver/caddy/commit/8848df9c5d372a559d01512b7a4ef00e38867b55

Where I revert a commit that did attempt to preserve the QS. Here's the background: https://github.com/caddyserver/caddy/issues/4205 (it seems complicated).

mholt:
Sure. Tests don't hurt :)

jaygooby:
OK, have started to get my head around tests. 

I've implemented a basic redirect and Test case 1 from your https://github.com/caddyserver/caddy/issues/4205#issuecomment-863352037 in this [integration test](https://github.com/jaygooby/caddy/blob/redo-redirect-fixes/caddytest/integration/redirect_test.go)

I'll get your test case 2 & 3 implemented too, along with one to test the query string preservation, then I'll try the `to` variable appending to get it passing

mholt:
@jaygooby Thanks for working on this, I'm trying to circle back around very soon and inspect this carefully.

mholt:
Superseded by #6109. Thank you for helping to get this started! Sorry it took me so long.

PR #5257, opened 2022-12-17
encode: respect Cache-Control HTTP header no-transform

darkweak:
Respect the request and response Cache-Control header with `no-transform` directive.
The request should not be altered if the `Cache-Control` contains the `no-transform` directive.
https://www.rfc-editor.org/rfc/rfc7234#section-5.2.1.6
https://www.rfc-editor.org/rfc/rfc7234#section-5.2.2.4

Closes https://github.com/darkweak/souin/issues/289

### DIFF
--- a/modules/caddyhttp/encode/encode.go
+++ b/modules/caddyhttp/encode/encode.go
@@ -24,7 +24,6 @@ import (
 	"io"
 	"math"
 	"net/http"
-	"regexp"
 	"sort"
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thank you! Let's give it a whirl :)

do7vip:
https://sakani.sa/app/arrow-right.ebe792a259c8ae0b.svg

PR #5258, opened 2022-12-17
Check for github action updates monthly

jlosito:
This will use dependabot to check the github actions which this project uses for updates on a monthly basis in order to keep them up-to-date.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


francislavoie:
Hmm, yeah that's probably reasonable.

I'm not a fan of dependabot for go.mod, we tend to want to be careful about updating dependencies, we need to review them every time we upgrade.

But for GitHub actions, that's just tooling that we can more easily forget about. So it's probably more valuable to have to remind us about those.

mholt (approved):
Sure, we can give it a shot :) But if it gets annoying I get dibs at turning it off :wink:

PR #5248, opened 2022-12-12
readme: darker variants of logos in dark mode

IndeedNotJames:
Closes #4040

Quote from me https://github.com/caddyserver/caddy/issues/5245#issuecomment-1345608813:
> GitHub supports the `prefers-color-scheme` CSS media feature in `<picture>` HTML elements since a few months.
>
> https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to

@francislavoie provided me with the caddy logo and I edited the certmagic one and made the "certificate paper" solid white instead of a white-transparent gradient.

I did not modify the ZeroSSL logo yet, because that one is coming from https://caddyserver.com/resources/images/zerossl-logo.svg (https://github.com/caddyserver/website/blob/master/src/resources/images/zerossl-logo.svg)

So I will most likely end up pushing to https://github.com/caddyserver/website and add another commit to this PR here :woman_shrugging: 

Note: The diff also adds a blank final newline (to be more on par with the other files) and removes a trailing whitespace

PR #5259, opened 2022-12-19
readme: white ZeroSSL text color in dark mode

emilylange:
Follow up to #5248 as discussed in Slack.

The SVG uploaded at https://user-images.githubusercontent.com/55066419/208327323-2770dc16-ec09-43a0-9035-c5b872c2ad7f.svg is the one https://github.com/caddyserver/website/blob/54af42121fc54a6334d85b57297725d13ce70a26/src/resources/images/zerossl-logo.svg with 2 main changes:

Added `prefers-color-scheme` to dynamically change based on the theme:
```html
<style>
    .text {
        fill: #000
    }

    @media (prefers-color-scheme:dark) {
        .text {
            fill: #fff
        }
    }
</style>
```

and minified it with the help of https://github.com/svg/svgo (`inlineStyles` plugin disabled, because it would ignore/remove the `@media`)

Has been tested by me and @francislavoie on Firefox and Chrome (desktop and Android) and GitHub mobile.
The official GitHub mobile app does not seem to understand `style="vertical-align: -7.7px"`, so that's why we keep the old `valign`.
I added the `vertical-align` so the SVG stops floating on Firefox.

That `7.7px` is somewhat inherited by the SVG itself.
One of the elements in the SVG (`<rect>`) defines `height="20.3"` and the `` tag uses `height="28"`.
`28px - 20.3px = 7.7px`, so the text "a" and "project" aligns perfectly with the "ZeroSSL" in the SVG :sweat_smile:

The special keyword `bottom` is pretty close, but seems to be slightly different across Chrome and Firefox. So I just settled on that hard-coded sub-pixel value

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 		</picture>
 	</a>
 	<br>
-	<h3 align="center">a <a href="https://zerossl.com"><img src="https://user-images.githubusercontent.com/55066419/208327323-2770dc16-ec09-43a0-9035-c5b872c2ad7f.svg" height="28" style="vertical-align: -7.7px"></a> project</h3>
+	<h3 align="center">a <a href="https://zerossl.com"><img src="https://user-images.githubusercontent.com/55066419/208327323-2770dc16-ec09-43a0-9035-c5b872c2ad7f.svg" height="28" style="vertical-align: -7.7px" valign="middle"></a> project</h3>
 </p>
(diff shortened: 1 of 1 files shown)


francislavoie (approved):

mholt (approved):
Cool, thank you! :sunglasses:

PR #5266, opened 2022-12-22
ci: exclude dependbot from running tests on s390x machine

mohammed90:
The dependbot PRs don't have access to secrets, which is necessary to connect to the S390x machine.


francislavoie (approved):
Thanks!

PR #5287, opened 2023-01-05
caddytls: Add `dns_ttl` config, improve Caddyfile `tls` options

ihmels:
My DNS provider (INWX) requires the TTL of a TXT record to be at least 300 seconds. Currently there is no way to configure this in a Caddyfile, if I see it correctly.

In addition, the `propagation_delay` and `propagation_timeout` options have also been added to the `tls` directive.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -24,6 +24,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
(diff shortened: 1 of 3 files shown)


mholt (approved):
Interesting; LGTM, thanks for the contribution and the test!

francislavoie:
I think this option should also be added to the `tls` directive as well, i.e. in `caddyconfig/httpcaddyfile/builtins.go` so that it can be configured without `issuer`. This is important because if you configure `issuer acme` only, then you disable the default issuer fallback to `zerossl`.

There's a few other properties that should probably get that treatment as well, such as `propagation_timeout` and `propagation_delay`, I just forgot to do those when I implemented those (if you'd like to make that change as well in this PR, that'd be cool :+1:)

mholt:
@francislavoie DNS TTL is specific to the ACME challenge though. I think placing it in the `tls` directive is out of scope. (It's already the case that if you want to customize the ACME issuers you need to manually specify the issuers you want.)

francislavoie:
That's not true @mholt, some of the options in top-level `tls` propagate down to both issuers, such as `dns`, `dns_challenge_override_domain` and `resolvers`. This is important so that it's easy to set these options both both issuers at once from the Caddyfile, without breaking the default issuer fallback behaviour.

ihmels:
I've now added the "dns_ttl" option to the TLS directive and also added "propagation_delay" and "propagation_timeout".

I've also added missing options in the godoc.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -363,6 +370,75 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+				if acmeIssuer == nil {
+					acmeIssuer = new(caddytls.ACMEIssuer)
+				}
+				if acmeIssuer.Challenges == nil {
+					acmeIssuer.Challenges = new(caddytls.ChallengesConfig)
+				}
+				if acmeIssuer.Challenges.DNS == nil {
+					acmeIssuer.Challenges.DNS = new(caddytls.DNSChallengeConfig)
+				}
+				acmeIssuer.Challenges.DNS.TTL = caddy.Duration(delay)
```
francislavoie, line 392: I think this is setting the wrong field (propagation delay)

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -363,6 +370,75 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
...
+				if acmeIssuer == nil {
+					acmeIssuer = new(caddytls.ACMEIssuer)
+				}
+				if acmeIssuer.Challenges == nil {
+					acmeIssuer.Challenges = new(caddytls.ChallengesConfig)
+				}
+				if acmeIssuer.Challenges.DNS == nil {
+					acmeIssuer.Challenges.DNS = new(caddytls.DNSChallengeConfig)
+				}
+				acmeIssuer.Challenges.DNS.TTL = caddy.Duration(timeout)
```
francislavoie, line 419: Same here (propagation timeout)

francislavoie (changes requested):

francislavoie (approved):
Great, thank you!

I guess we could use a test for the top-level `tls` config + propagation config (or add it to an existing test) for bonus points, but this meets the minimum bar :grin:

caddyconfig/httpcaddyfile/builtins.go
ihmels: Oops, you're right. I've fixed it. :grin:

mholt (approved):
Ah yeah, I guess @francislavoie is right as usual :)

Thank you for the discussion and the contribution. Merge away!

PR #5296, opened 2023-01-11
chore: fix typo in streaming.go

eltociear:
cancelation -> cancellation


mohammed90:
Thank you for the PR. It isn't a typo. It is the American spelling of the same word.

mholt:
Even as an American I switch off between the two. I think both forms are generally accepted. :man_shrugging:

PR #5297, opened 2023-01-11
cmd: `caddy fmt` return code is 1 if not formatted

### LINKED ISSUE #5294: Allow returning 1 if lint fails
Right now we have

```shell
Usage:
  caddy fmt [flags]

Flags:
      --diff        Print the differences between the input file and the formatted output
  -h, --help        help for fmt
      --overwrite   Overwrite the input file with the results
```

I originally thought `caddy fmt` will return `1` if failed like most other linters. It turns out not. It returns formatted content instead.
If we don't want to change the behavior of `caddy fmt`, it would be great to provide another flag to allow returning `1` for failure. Thanks! 😃

u5surf:
* Fixes #5294
* return exit 1 with an error message

### DIFF
--- a/caddyconfig/caddyfile/adapter.go
+++ b/caddyconfig/caddyfile/adapter.go
@@ -54,7 +54,7 @@ func (a Adapter) Adapt(body []byte, options map[string]any) ([]byte, []caddyconf
 
 	// lint check: see if input was properly formatted; sometimes messy files files parse
 	// successfully but result in logical errors (the Caddyfile is a bad format, I'm sorry)
-	if warning, different := formattingDifference(filename, body); different {
+	if warning, different := FormattingDifference(filename, body); different {
 		warnings = append(warnings, warning)
 	}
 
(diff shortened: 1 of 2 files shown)


mholt:
Awesome, thanks for this patch.

I was just thinking, our Caddyfile adapter does print a warning if the input isn't formatted, and the output is a little more useful:

https://github.com/caddyserver/caddy/blob/223cbe3d0b50487117c785f0755bb80a9ee65010/caddyconfig/caddyfile/adapter.go#L66-L93

Do you think we could incorporate this? Maybe we export that function and use it in `caddy fmt` to give the user more information.

u5surf:
> https://github.com/caddyserver/caddy/blob/223cbe3d0b50487117c785f0755bb80a9ee65010/caddyconfig/caddyfile/adapter.go#L66-L93
> 
> Do you think we could incorporate this? Maybe we export that function and use it in `caddy fmt` to give the user more information.

@mholt Hi, thanks reviewing and sorry for late. I was exported and used `caddyfile.FormattingDifference`.

mholt (approved):
Awesome, LGTM. Thanks for the contribution!

u5surf:
@mholt  Hi, could you confirm to be able to merge it? I seems that it might had been lost in your sight after approving it🙏

mholt (approved):
Oops! Yes, sorry. Thank you for the contribution again :blush:

PR #5314, opened 2023-01-17
go.mod: Update golang.org/x/net to v0.5.0

alexandrevicenzi:
Fix CVE-2022-41721

https://nvd.nist.gov/vuln/detail/CVE-2022-41721
https://pkg.go.dev/vuln/GO-2023-1495


alexandrevicenzi:
Sending a PR due to https://bugzilla.opensuse.org/show_bug.cgi?id=1207207

francislavoie:
Thanks.

We're not vulnerable to that CVE, because we don't use `MaxBytesHandler`, as far as I can tell.

That said, still good to update the dependency.

francislavoie (approved):

alexandrevicenzi:
All symbols in [golang.org/x/net/http2/h2c](https://pkg.go.dev/golang.org/x/net/http2/h2c) should be affected by this CVE, and caddy does use h2c, so it can potentially be a problem, but I did not try to reproduce the attack to confirm that it is or not affected.

Thanks for merging it.

francislavoie:
We don't enable h2c by default, users must explicitly configure the `h2c` protocol in their global options https://caddyserver.com/docs/caddyfile/options#protocols. So the impact is quite limited. Ideally no users actually allow public access to h2c and only use it in internal networks. But :man_shrugging:

PR #5315, opened 2023-01-17
httpcaddyfile: Fix `handle` grouping inside `route`

### LINKED ISSUE #5308: Nesting `handle` within `route` loses `group` markings
Discovered in https://caddy.community/t/exclusive-basicauth-paths/18528

Consider this Caddyfile:

```
:8881 {
    route {
        handle /foo/* {
            respond "Foo"
        }
        handle {
            respond "Bar"
        }
    }
}
```

The adapted JSON looks like this:

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":8881"
          ],
          "routes": [
            {
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "handler": "subroute",
                          "routes": [
                            {
                              "handle": [
                                {
                                  "body": "Foo",
                                  "handler": "static_response"
                                }
                              ]
                            }
                          ]
                        }
                      ],
                      "match": [
                        {
                          "path": [
                            "/foo/*"
                          ]
                        }
                      ]
                    },
                    {
                      "handle": [
                        {
                          "handler": "subroute",
                          "routes": [
                            {
                              "handle": [
                                {
                                  "body": "Bar",
                                  "handler": "static_response"
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      }
    }
  }
}
```

Notice that there's no `"group"` for the handle subroutes as would be expected. This means that the mutual exclusivity part of the `handle` blocks gets broken.

See this Caddyfile without the wrapping `route`, which _does_ have the `"group"` as expected:

```
:8881 {
    handle /foo/* {
        respond "Foo"
    }
    handle {
        respond "Bar"
    }
}
```

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":8881"
          ],
          "routes": [
            {
              "group": "group2",
              "match": [
                {
                  "path": [
                    "/foo/*"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "Foo",
                          "handler": "static_response"
                        }
                      ]
                    }
                  ]
                }
              ]
            },
            {
              "group": "group2",
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "Bar",
                          "handler": "static_response"
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      }
    }
  }
}
```

Something with the way `route` is handled makes `group` not work for `handle`. We should figure out what's causing it.

WeidiDeng:
Fix #5308

The original parseRoute just appends result's routes to its routes, which is not enough when subroutes contain handle routes which must be exclusive.

A simple modification to a private function buildSubroute proved to be just enough, obviously routes don't need to be sorted.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -731,29 +731,12 @@ func parseError(h Helper) (caddyhttp.MiddlewareHandler, error) {
 
 // parseRoute parses the route directive.
 func parseRoute(h Helper) (caddyhttp.MiddlewareHandler, error) {
-	sr := new(caddyhttp.Subroute)
-
 	allResults, err := parseSegmentAsConfig(h)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, result := range allResults {
-		switch handler := result.Value.(type) {
-		case caddyhttp.Route:
-			sr.Routes = append(sr.Routes, handler)
-		case caddyhttp.Subroute:
-			// directives which return a literal subroute instead of a route
-			// means they intend to keep those handlers together without
-			// them being reordered; we're doing that anyway since we're in
-			// the route directive, so just append its handlers
-			sr.Routes = append(sr.Routes, handler.Routes...)
-		default:
-			return nil, h.Errf("%s directive returned something other than an HTTP route or subroute: %#v (only handler directives can be used in routes)", result.directive, result.Value)
-		}
-	}
-
-	return sr, nil
+	return buildSubroute(allResults, h.groupCounter, false)
 }
 
 func parseHandle(h Helper) (caddyhttp.MiddlewareHandler, error) {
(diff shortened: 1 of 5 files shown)


francislavoie:
Looks like one of the adapt tests needs fixing to add `"group"`.

mholt:
@WeidiDeng You're the best. Thanks for the patch! :man_dancing:

mholt (approved):
This LGTM once the test is fixed up :grin:

francislavoie:
I pushed a fix for the failing test.

I think we should add a test for the case that this fixes though. I'll do that in a sec.

francislavoie (approved):

PR #5320, opened 2023-01-19
httpcaddyfile: Ignore the import empty dotfile.

### LINKED ISSUE #5295: [feature request] ignore dotfiles by default when parsing drop-in configuration
here is the error log:
```
root@xxxxxx:~/git/private_proxy_config# docker logs -f caddy
{"level":"info","ts":1673350699.8861008,"msg":"using provided configuration","config_file":"/etc/caddy/Caddyfile","config_adapter":"caddyfile"}
Error: adapting config using caddyfile: /etc/caddy/Caddyfile:39 - Error during parsing: Could not read tokens while importing /etc/caddy/sites-enabled/.gitkeep: EOF

root@xxxxxx:~/git/private_proxy_config# file caddy/sites-enabled/.gitkeep
caddy/sites-enabled/.gitkeep: empty
```

---

I was trying to git-ops my caddy configuration. Some machines do not need the drop-in configuration while others do.
I decide to put an empty file in the `sites-enabled` sub-dir so it would be tracked by git.

u5surf:
* Fixes #5295
ignore the empty dot file while it imports.
and also warning in this situation.

https://github.com/caddyserver/caddy/issues/5295#issuecomment-1377670717


francislavoie:
This still isn't right. I think you misunderstood the requirements. How did you try testing this? I think this will need some tests to show that it works correctly.

- `import path/to/empty_file.txt` currently returns an `EOF` error. It should now instead just log a warning, and result in no tokens.
- `import path/to/dir/*` currently tries to import dotfiles as well with the `*` glob. It should instead skip any files that start with a `.` when iterating over them.
- `import path/to/dir/.*` currently reads all dotfiles in a dir. It should continue to do so. The idea is that this is an escape hatch to still load all dotfiles as config if the user wants to, when pairing with `import path/to/dir/*` which does not import the dotfiles.

mholt:
I want to emphasize, though, we are very grateful for the patch :100: :smiley:  We probably could have better clarified the requirements in the issue and will try to do so in the future!

caddyconfig/caddyfile/parse.go
```diff
@@ -451,6 +465,12 @@ func (p *parser) doSingleImport(importFile string) ([]Token, error) {
 		return nil, p.Errf("Could not read imported file %s: %v", importFile, err)
 	}
 
+	// only warning in case of empty files
+	if len(input) == 0 {
+		caddy.Log().Warn("import file is empty", zap.String("file", importFile))
+		return []Token{}, nil
+	}
```
francislavoie, line 472: I think we might need to go one step further here and check if the input is only whitespace (i.e. spaces and/or newlines) or empty.

Also, I think the warning message should start with an uppercase letter, like the other ones in this file (i.e. `"Import file is empty"`).

francislavoie (commented):
Thanks! This is looking a lot better.

caddyconfig/caddyfile/parse.go
u5surf: @francislavoie I took care of this.

francislavoie (approved):

mholt (approved):
Awesome, thanks so much! Let's give this a whirl :)

PR #5323, opened 2023-01-20
caddyfile: Allow overriding server names

### LINKED ISSUE #5322: Support changing the internal server name/id via Caddyfile
It doesn't seem to be possible to define the `server` names via Caddyfile for better identification through logs or metrics.

When using Caddyfile format, servers are named following a `srvX` pattern and when multiple sites are defined it becomes hard to identify which one it is referring to in logs or metrics

This is currently possible via the JSON [servers](https://caddyserver.com/docs/json/apps/http/servers/) keys.

francislavoie:
Fix #5322

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -706,7 +706,7 @@ func (st *ServerType) serversFromPairings(
 
 	err := applyServerOptions(servers, options, warnings)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("applying global server options: %v", err)
 	}
 
 	return servers, nil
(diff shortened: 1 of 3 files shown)


IndeedNotJames (changes requested):
Breaks when two different servers are called the same:
```Caddyfile
{
	servers :443 {
		name https
	}

	servers :8000 {
		name https
	}
}

example.com {
}

:8000 {
}
```

```json
{
    "apps": {
        "http": {
            "servers": {
                "https": {
                    "listen": [
                        ":8000"
                    ]
                }
            }
        }
    }
}
```

francislavoie:
Ah right, I should reject duplicate names.

IndeedNotJames (approved):

mholt (approved):
Thanks. This is something I hoped we wouldn't really have to care about... but here we are. :sweat_smile:

PR #5326, opened 2023-01-22
change TrustedProxyVarKey type to caddy.CtxKey

WeidiDeng:
per [context documentation](https://pkg.go.dev/context#WithValue) and other caddy context key type.


WeidiDeng:
Forget about it, its a var key not a ctx key

francislavoie:
Yep, that was on purpose. See https://github.com/caddyserver/caddy/pull/5103#discussion_r991486237

Issue #5193, opened 2022-11-09
Allow trusted_proxies_header


aadnehovda:
Currently it's only possible to trust a proxy by IP range, but if you run your site behind for example Azure Front Door it makes sense to instead or additionally validate the `X-Azure-FDID` header. The value of this header is set to the ID of the Front Door resource in Azure and is only passed to the backend servers (my server) so it should not leak. 

So for example in a Caddyfile

```
php_fastcgi unix//run/php/php-fpm.sock {
    trusted_proxies_header X-Azure-FDID 9c4888d4-f687-4917-987e-777fe1ed2eaf
}
```
https://learn.microsoft.com/en-us/azure/frontdoor/origin-security?tabs=app-service-functions&pivots=front-door-standard-premium#front-door-identifier

francislavoie:
Huh, interesting.

> When Front Door makes a request to your origin, it adds the X-Azure-FDID request header. Your origin should inspect the header on incoming requests, and reject requests where the value doesn't match your Front Door profile's identifier.

According to this, shouldn't you outright reject requests that don't have this header? You can do that with the `header` matcher, and use the `abort` directive to cancel requests that don't match.

Do you have a usecase for still allowing requests not coming directly from Azure?

aadnehovda:
You are right, a use case would maybe be for debugging or testing purposes, like testing backend servers before adding them to the pool. 

However, as mentioned in the MS docs, all Front Door instances will be using the same IP ranges. Could there be a potenial risk that someone could set up a malicious Front Door instance and reverse proxy your site onto a different domain, spoof or manipulate headers, like the client IP in order to circumvent location aware content on your site or impersonate your site in order to track users, or other privacy issues? 

The list of IP ranges may be changing frequently (someone scripted the updates for download at https://azureipranges.azurewebsites.net/getPrefixes/Public/AzureFrontDoor.Backend), so if we were to only rely on trusted_proxies it needs to be automated somehow, maybe through the API? Not sure, I'm pretty new to Caddy. 

`trusted_proxies` implies that the proxies are in fact trusted, but it seems to me that basing the trust on the IP address alone may not be sufficient in all cases.

aadnehovda:
Another thing I noticed is that there now is a standardized [`Forwarded`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) request header which should probably also be protected from untrusted proxies?

francislavoie:
[...]

PR #5328, opened 2023-01-23
caddyhttp: Pluggable trusted proxy IP range sources

francislavoie:
Followup to #5103, I realized that the work there was somewhat incomplete.

We've already had requests for making it easy to configure trusted proxy ranges for e.g. CloudFlare, which publish their IP ranges at https://www.cloudflare.com/ips/. This PR makes it possible for a plugin to implement a module that can fetch those IP ranges and feed it to the global trusted proxies in the server.

This should be merged ASAP, before the next release, because otherwise this would be a breaking change. But since we haven't released this yet, it's the best time to make this pluggable.

Regarding the interface, I decided that it should just return a slice of `netip.Prefix` and not an error, because returning any error would mess with the HTTP server being able to function properly. See the godoc comment on the interface where I wrote some implementation recommendations for plugins which load the source from elsewhere. This godoc comment could possibly use some language polish.

Maybe TODO, should we implement a `MultiSource` right away, which would merge IP ranges from multiple sources? That might be handy for trusting something like CloudFlare + some other IPs from servers controlled by the user.

### DIFF
--- a/caddyconfig/httpcaddyfile/serveroptions.go
+++ b/caddyconfig/httpcaddyfile/serveroptions.go
@@ -43,7 +43,7 @@ type serverOptions struct {
 	MaxHeaderBytes       int
 	Protocols            []string
 	StrictSNIHost        *bool
-	TrustedProxies       []string
+	TrustedProxiesRaw    json.RawMessage
 	ShouldLogCredentials bool
 	Metrics              *caddyhttp.Metrics
 }
@@ -188,13 +188,25 @@ func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (any, error) {
 				serverOpts.StrictSNIHost = &boolVal
 
 			case "trusted_proxies":
-				for d.NextArg() {
-					if d.Val() == "private_ranges" {
-						serverOpts.TrustedProxies = append(serverOpts.TrustedProxies, caddyhttp.PrivateRangesCIDR()...)
-						continue
-					}
-					serverOpts.TrustedProxies = append(serverOpts.TrustedProxies, d.Val())
+				if !d.NextArg() {
+					return nil, d.Err("trusted_proxies expects an IP range source module name as its first argument")
 				}
+				modID := "http.ip_range_sources." + d.Val()
+				unm, err := caddyfile.UnmarshalModule(d, modID)
+				if err != nil {
+					return nil, err
+				}
+				source, ok := unm.(caddyhttp.IPRangeSource)
+				if !ok {
+					return nil, fmt.Errorf("module %s (%T) is not an IP range source", modID, unm)
+				}
+				jsonSource := caddyconfig.JSONModuleObject(
+					source,
+					"source",
+					source.(caddy.Module).CaddyModule().ID.Name(),
+					nil,
+				)
+				serverOpts.TrustedProxiesRaw = jsonSource
 
 			case "metrics":
 				if d.NextArg() {
@@ -304,7 +316,7 @@ func applyServerOptions(
 		server.MaxHeaderBytes = opts.MaxHeaderBytes
 		server.Protocols = opts.Protocols
 		server.StrictSNIHost = opts.StrictSNIHost
-		server.TrustedProxies = opts.TrustedProxies
+		server.TrustedProxiesRaw = opts.TrustedProxiesRaw
 		server.Metrics = opts.Metrics
 		if opts.ShouldLogCredentials {
 			if server.Logs == nil {
--- a/caddyconfig/httpcaddyfile/serveroptions.go
+++ b/caddyconfig/httpcaddyfile/serveroptions.go
@@ -191,7 +191,7 @@ func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (any, error) {
 				if !d.NextArg() {
 					return nil, d.Err("trusted_proxies expects an IP range source module name as its first argument")
 				}
-				modID := "http.ip_range_sources." + d.Val()
+				modID := "http.ip_sources." + d.Val()
 				unm, err := caddyfile.UnmarshalModule(d, modID)
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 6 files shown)


aadnehovda:
What if a plugin could simply provide yes/no to whether a proxy is trusted, instead of just the dynamic IP ranges? Could the plugin then include additional checks such as TLS auth, secret headers (#5193), etc.? . A rogue proxy instance (e.g. for spoofing real client IPs to bypass geowalls) on cloudflare will be using the same IP ranges as your own instance.

francislavoie:
Well the nice thing about the design using a var is that a plugin could just set `trusted_proxy` to true using whatever conditions it wants. You could even do it in your config with the [`vars` handler](https://caddyserver.com/docs/caddyfile/directives/vars) with something like `vars @some-condition trusted_proxy true`.

I thought about passing the `req` to this interface, but then it's more push than pull, and acts more like a request matcher. Then the question would be, is there a point to the interface if it's just a request matcher? I think this is simpler, because it abstracts away the checking of the remote IP by just feeding Caddy some IP ranges which is easier to configure and harder to mess up.

mholt (approved):
This is a great improvement. Thanks for the enhancement Francis!

Responding to the discussion in a moment.

mholt:
Code LGTM -- I think there's some good discussion points in here.

The matcher idea is interesting. In the future, maybe we could use this to "feed" a matcher for a trusted proxy? I.e. it can pull from this global option and use it to determine whether a requests matches or not.

> I thought about passing the req to this interface, but then it's more push than pull, and acts more like a request matcher.

True, it does feel like a matcher -- however I have not regretted passing the *Request in, because it does open a lot more possibilities and still grants the same functionality at no higher cost or complexity. (We just have to trust that the request being passed in isn't being changed inadvertently.)

And actually, the "push" model feels more enticing to me because of the possibilities for dynamic answers. Kind of like how we do for reverse proxy upstreams, being able to get them dynamically at every request.

Would it be so bad for the trusted proxies implementation to actually be a set of request matchers instead? That way we're not limited to just IP ranges in the future.

francislavoie:
> Would it be so bad for the trusted proxies implementation to actually be a set of request matchers instead? 

It would certainly make it more awkward to configure in the Caddyfile. That's my main concern. It might not be syntactically obvious that a matcher is what's used at that spot in the config.

But I guess for the CloudFlare case it would probably mean someone should implement a `http.matchers.is_cloudflare` module, then it would look like `trusted_proxy is_cloudflare`, I guess. They would have to implement the remote address matching logic themselves in that matcher, whereas we can abstract that away currently in Caddy.

I dunno. I'm lukewarm on it. There's good arguments for it, but :thinking:

modules/caddyhttp/ip_range.go
```diff
@@ -0,0 +1,137 @@
...
+// Using a caddy.UsagePool may be a good idea to avoid having refetch
+// the values when a config reload occurs, which would waste time.
+//
+// If the list of IP ranges cannot be sourced, then provisioning should
+// fail. Getting the IP ranges at runtime must not fail, because it would
+// cancel incoming requests, which is not ideal. If refreshing the list
+// fails, then the previous list of IP ranges should continue to be returned
+// so that the server can continue to operate normally.
+type IPRangeSource interface {
+	GetIPRanges() []netip.Prefix
```
mholt, line 47: Let's pass in the `*http.Request` here, just in case.

mholt (changes requested):
Just one suggestion from our call today :)

modules/caddyhttp/ip_range.go
francislavoie: Done. I updated the interface's godoc comment to hopefully clarify the risk involved with using it.

caddyconfig/httpcaddyfile/serveroptions.go
```diff
@@ -188,13 +188,25 @@ func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (any, error) {
...
-				for d.NextArg() {
-					if d.Val() == "private_ranges" {
-						serverOpts.TrustedProxies = append(serverOpts.TrustedProxies, caddyhttp.PrivateRangesCIDR()...)
-						continue
-					}
-					serverOpts.TrustedProxies = append(serverOpts.TrustedProxies, d.Val())
+				if !d.NextArg() {
+					return nil, d.Err("trusted_proxies expects an IP range source module name as its first argument")
+				}
+				modID := "http.ip_range_sources." + d.Val()
```
mholt, line 194: One other thought... should/could this be `http.ip_sources`, just for succinctness? Cause it could be either whole IPs or ranges.

aadnehovda:
A plugin could check the request for a header and just return an empty IP range if it is missing/has the wrong value?

francislavoie:
> @aadnehovda A plugin could check the request for a header and just return an empty IP range if it is missing/has the wrong value?

An empty list of IP ranges would make the request untrusted. Setting the range to `0.0.0.0/0` would make the request trusted.

From my chat with Matt, we decided to go with an IP range fetching API instead of request matchers, because _really_, the only part of the request that can actually be trusted is the remote address, since it's the only part of the request that is essentially tamper-proof. Headers can be spoofed.

If you as a user have your own external guarantees (networking isolation) to ensure no other connections can reach your server except through the proxy, then if you feel it's necessary to look at the request to make a decision about trust, now that's possible. But I still wouldn't recommend it in general.

mholt (approved):

adryd325:
how does one create an ip range source in a Caddyfile at the moment?
this syntax no-longer works
```
trusted_proxies {
    140.238.155.71/32
}
```

francislavoie:
Currently in v2.6.2, you must specify trusted proxies as a subdirective to `reverse_proxy`. You have to put them on the same line, no `{ }` block.

```
example.com {
		trusted_proxies 140.238.155.71/32
	}
}
```

As of v2.6.3 you'll be able to specify it as a global option, like this:

```
{
	servers {
		trusted_proxies static 140.238.155.71/32
	}
}

example.com {
	reverse_proxy localhost:8080
}
```

PR #5329, opened 2023-01-23
tracing: Support placeholders in span name

### LINKED ISSUE #5171: [opentelemetry] add ability to contextualize span name
Since caddy is often the entry point to an application, it's generally the root span and it's name becomes the operation name of the entire trace. As a result, tools like Jaeger show a list of various operations that all have the same name.

It would be great if the span name could be customized on a "per request" basis with string interpolation in the span name, eg.

```
tracing {
  span '{request>uri}'
}
```
or even get the operation name from the upstream like this:
```
tracing {
  span '{response>headers>X-Trace-Operation-Name}'
}
```

davidfrickert:
Fixes #5171

### Notes:
- I'm a noob in Go - this was my first tackle at using the language. 
- I fixed the tests that started breaking, but did not add any new tests. I have no idea how to check the output span after it has been formatted in the tests.
- I built caddy and used it in my Server and checked that the output spans are correct - as in the placeholders are replaced.

### DIFF
--- a/modules/caddyhttp/tracing/module_test.go
+++ b/modules/caddyhttp/tracing/module_test.go
@@ -98,7 +98,7 @@ func TestTracing_ServeHTTP_Propagation_Without_Initial_Headers(t *testing.T) {
 		SpanName: "mySpan",
 	}
 
-	req := httptest.NewRequest("GET", "https://example.com/foo", nil)
+	req := createRequestWithContext("GET", "https://example.com/foo")
 	w := httptest.NewRecorder()
 
 	var handler caddyhttp.HandlerFunc = func(writer http.ResponseWriter, request *http.Request) error {
@@ -128,7 +128,7 @@ func TestTracing_ServeHTTP_Propagation_With_Initial_Headers(t *testing.T) {
 		SpanName: "mySpan",
 	}
 
-	req := httptest.NewRequest("GET", "https://example.com/foo", nil)
+	req := createRequestWithContext("GET", "https://example.com/foo")
 	req.Header.Set("traceparent", "00-11111111111111111111111111111111-1111111111111111-01")
 	w := httptest.NewRecorder()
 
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/tracing/module_test.go
```diff
@@ -180,3 +180,11 @@ func TestTracing_ServeHTTP_Next_Error(t *testing.T) {
 		t.Errorf("expected error, got: %v", err)
 	}
 }
+
+func createRequestWithContext(method string, url string) *http.Request {
+	r, _ := http.NewRequest("GET", "/", nil)
```
mohammed90, line 185: This doesn't look right. The variable `url` isn't being used, and the request always goes to `/`.

davidfrickert: yep you're right, copied it and didn't actually use the variables

davidfrickert: fixed! https://github.com/caddyserver/caddy/compare/36c145ec329095be75555b5e4b833743a831dc0b..48797e1dc3bc525895902a1b904741c3450cb29c

francislavoie (approved):
This looks good to me. Thanks for contributing!

cdaguerre:
thanks @davidfrickert !!
Is there an easy way to figure out which placeholders are actually supported?
To my understanding, [these](https://caddyserver.com/docs/caddyfile/concepts#placeholders) should be supported, but not the response, correct?

francislavoie:
Yeah, basically any request that reads from the request should work. Those you linked are the Caddyfile shortcuts, but many more are in https://caddyserver.com/docs/json/apps/http/#docs. Depends what you want to do with it.

davidfrickert:
No problem @francislavoie @cdaguerre.
My pleasure to contribute to such an awesome project.

PR #5330, opened 2023-01-24
reverseproxy: Add flag to short command to disable redirects

### LINKED ISSUE #5290: The `reverse-proxy` does not work when `--from https://domain:port`
It still tries to listen on `:80`

```
# ./caddy-2 reverse-proxy --from https://localhost:2096 --to http://127.0.0.1:8880 --internal-certs
2023/01/07 02:58:07.806	WARN	admin	admin endpoint disabled
2023/01/07 02:58:07.807	INFO	http	enabling automatic HTTP->HTTPS redirects	{"server_name": "proxy"}
2023/01/07 02:58:07.808	INFO	tls.cache.maintenance	started background certificate maintenance	{"cache": "0xc000570f50"}
2023/01/07 02:58:07.833	INFO	pki.ca.local	root certificate is already trusted by system	{"path": "storage:pki/authorities/local/root.crt"}
2023/01/07 02:58:07.834	INFO	http	enabling HTTP/3 listener	{"addr": ":2096"}
2023/01/07 02:58:07.835	INFO	failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
2023/01/07 02:58:07.835	INFO	http.log	server running	{"name": "proxy", "protocols": ["h1", "h2", "h3"]}
2023/01/07 02:58:07.835	INFO	tls.cache.maintenance	stopped background certificate maintenance	{"cache": "0xc000570f50"}
Error: loading new config: http app module: start: listening on :80: listen tcp :80: bind: address already in use

# ./caddy-2 version
v2.6.2 h1:wKoFIxpmOJLGl3QXoo6PNbYvGW4xLEgo32GPBEjWL8o=
```

u5surf:
* Fixes #5290

Add new flag --disable-redirects for reverseproxy.
https://github.com/caddyserver/caddy/issues/5290#issuecomment-1383449714
`Using a command is more convenient in my case.`

### DIFF
--- a/modules/caddyhttp/reverseproxy/command.go
+++ b/modules/caddyhttp/reverseproxy/command.go
@@ -64,6 +64,7 @@ default, all incoming headers are passed through unmodified.)
 			fs.Bool("insecure", false, "Disable TLS verification (WARNING: DISABLES SECURITY BY NOT VERIFYING SSL CERTIFICATES!)")
 			fs.Bool("internal-certs", false, "Use internal CA for issuing certs")
 			fs.Bool("debug", false, "Enable verbose debug logs")
+			fs.Bool("disable-redirects", false, "Disable HTTP->HTTPS redirects")
 			return fs
 		}(),
 	})
@@ -77,6 +78,7 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 	insecure := fs.Bool("insecure")
 	internalCerts := fs.Bool("internal-certs")
 	debug := fs.Bool("debug")
+	disableRedir := fs.Bool("disable-redirects")
 
 	httpPort := strconv.Itoa(caddyhttp.DefaultHTTPPort)
 	httpsPort := strconv.Itoa(caddyhttp.DefaultHTTPSPort)
@@ -174,6 +176,13 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 
 	if fromAddr.Scheme == "http" {
 		server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{Disabled: true}
+	}
+	if disableRedir {
+		if server.AutoHTTPS != nil {
+			server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisableRedir: true}
+		} else {
+			server.AutoHTTPS.DisableRedir = true
+		}
 	}
 
 	httpApp := caddyhttp.App{
--- a/modules/caddyhttp/reverseproxy/command.go
+++ b/modules/caddyhttp/reverseproxy/command.go
@@ -176,13 +176,8 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 
 	if fromAddr.Scheme == "http" {
 		server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{Disabled: true}
-	}
-	if disableRedir {
-		if server.AutoHTTPS != nil {
-			server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisableRedir: true}
-		} else {
-			server.AutoHTTPS.DisableRedir = true
-		}
+	} else if disableRedir {
+		server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisableRedir: true}
 	}
 
 	httpApp := caddyhttp.App{


modules/caddyhttp/reverseproxy/command.go
```diff
@@ -168,14 +170,17 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 	}
 
 	server := &caddyhttp.Server{
-		Routes: caddyhttp.RouteList{route},
-		Listen: []string{":" + fromAddr.Port},
+		Routes:    caddyhttp.RouteList{route},
+		Listen:    []string{":" + fromAddr.Port},
+		AutoHTTPS: &caddyhttp.AutoHTTPSConfig{},
```
francislavoie, line 175: I don't think this is ideal -- one thing we tend to try to do is keep the generated config as lean as possible. So `AutoHTTPS` should only exist if it has any options set.

That means having a bit more code each place where the property is being touched that initializes the struct if it doesn't exist yet, with an if block. It's more verbose, but the output JSON is cleaner.

modules/caddyhttp/reverseproxy/command.go
```diff
@@ -175,6 +177,13 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 	if fromAddr.Scheme == "http" {
 		server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{Disabled: true}
 	}
+	if disableRedir {
+		if server.AutoHTTPS != nil {
+			server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisabledRidir: true}
+		} else {
+			server.AutoHTTPS.DisableRedir = true
+		}
```
francislavoie, line 185: There was a typo, but also it can be simplified a bit:

```suggestion
		if server.AutoHTTPS != nil {
			server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{}
		}
		server.AutoHTTPS.DisableRedir = true
```

modules/caddyhttp/reverseproxy/command.go
```diff
@@ -175,6 +177,13 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 	if fromAddr.Scheme == "http" {
 		server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{Disabled: true}
 	}
+	if disableRedir {
+		if server.AutoHTTPS != nil {
+			server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisableRedir: true}
```
IndeedNotJames, line 182: I don't like that it takes priority over `fromAddr.Scheme == "http"`

Opinions on throwing an error in such case instead? :eyes:

modules/caddyhttp/reverseproxy/command.go
```diff
@@ -64,6 +64,7 @@ default, all incoming headers are passed through unmodified.)
 			fs.Bool("insecure", false, "Disable TLS verification (WARNING: DISABLES SECURITY BY NOT VERIFYING SSL CERTIFICATES!)")
 			fs.Bool("internal-certs", false, "Use internal CA for issuing certs")
 			fs.Bool("debug", false, "Enable verbose debug logs")
+			fs.Bool("disable-redirects", false, "Disable redirect HTTP->HTTPS")
```
IndeedNotJames, line 67: Might be me nitpicking, but I would rephrase it ever so slightly :woman_shrugging: 

```suggestion
			fs.Bool("disable-redirects", false, "Disable HTTP->HTTPS redirects")
```

IndeedNotJames (changes requested):

modules/caddyhttp/reverseproxy/command.go
u5surf: @IndeedNotJames Thanks!
Thus, you mean `fromAddr.Scheme == “http”` takes priority over redirect disabled. 
How would you like that it ignores disable redirects if user specifies both of fromAddr.Scheme http and redirect disabled instead it occurs an error?

modules/caddyhttp/reverseproxy/command.go
u5surf: @IndeedNotJames I took care of it

francislavoie:
Oh, you might have missed it but I pushed a commit to the branch to fix https://github.com/caddyserver/caddy/pull/5330#discussion_r1085352670 but I think you force pushed which wiped it out.

u5surf:
@francislavoie 
> Oh, you might have missed it but I pushed a commit to the branch to fix [#5330 (comment)](https://github.com/caddyserver/caddy/pull/5330#discussion_r1085352670) but I think you force pushed which wiped it out.

I’m sorry my fault, I fixed typo.

modules/caddyhttp/reverseproxy/command.go
IndeedNotJames: Hi, sorry for the very vague comment yesterday.
I just had another look at the code, so let me try to clarify it:

I originally observed that
```
caddy reverse-proxy --from http://localhost:2020 --to https://example.com --change-host-header --disable-redirects
```

will start serving HTTPS on `localhost:2020` (despite having `--from http://localhost:2020`).

So I was like "if I, as a user, explicitly pass `http://` in `--from http://localhost:2020`, then I want Caddy to only listen on `localhost:2020`, serving HTTP".

And considering there are no HTTP->HTTPS redirects on an HTTP-only server, why allow `--disable-redirects` in the first place?

---

Now that I had a closer look at the code again, I noticed that

`--disable-redirects` takes priority over `&caddyhttp.AutoHTTPSConfig{Disabled: true}`, because you override the existing `AutoHTTPSConfig` struct, if `server.AutoHTTPS != nil`.
I think you meant `if server.AutoHTTPS == nil {` instead:

```diff
@@ -178,7 +178,7 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
                server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{Disabled: true}
        }
        if disableRedir {
-               if server.AutoHTTPS != nil {
+               if server.AutoHTTPS == nil {
                        server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisableRedir: true}
                } else {
                        server.AutoHTTPS.DisableRedir = true
```

So that explains why `--from http://localhost:2020` with `--disable-redirects` enabled HTTPS on `localhost:2020`.

`--disable-redirects` took priority over `--from http://localhost:2020` due to a small little typo in `if server.AutoHTTPS == nil`

---

However, from the [AutoHTTPS json docs](https://caddyserver.com/docs/json/apps/http/servers/automatic_https/):

> [disable](https://caddyserver.com/docs/json/apps/http/servers/automatic_https/disable/)
> If true, automatic HTTPS will be entirely disabled, including certificate management and redirects.

So knowing that
```
&caddyhttp.AutoHTTPSConfig{DisableRedir: true}
```
does nothing if
```
&caddyhttp.AutoHTTPSConfig{Disabled: true}
```
is already set (because `{Disabled: true}` already disables all the AutoHTTPS features, including `{DisableRedir: true}`), then you could as well just write
```diff
@@ -176,13 +176,8 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 
        if fromAddr.Scheme == "http" {
                server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{Disabled: true}
-       }
-       if disableRedir {
-               if server.AutoHTTPS != nil {
-                       server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisableRedir: true}
-               } else {
-                       server.AutoHTTPS.DisableRedir = true
-               }
+       } else if disableRedir {
+               server.AutoHTTPS = &caddyhttp.AutoHTTPSConfig{DisableRedir: true}
        }
```

Which simplifies the code a bit.

---

So it's probably perfectly fine to allow `--disable-redirects` whenever `fromAddr.Scheme == "http"`.
No need to throw an error or whatever, since they don't actually conflict with each other :)

However, I would suggest (diff above) only setting `{DisableRedir: true}` when `{Disabled: true}` isn't already set :)

IndeedNotJames (approved):

francislavoie (approved):
Nice! I like how simple and clean this became.

PR #5331, opened 2023-01-24
go.mod: bump tscert package to fix Tailscale 1.34+ on Windows

bradfitz:
As of Tailscale 1.34.0 on Windows, Tailscale now uses a named pipe to connect to the local tailscale service.

This pulls in tailscale/tscert#5 as reported in tailscale/tscert#4.

(Sorry, we should've noticed this earlier!)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/smallstep/truststore v0.12.0
 	github.com/spf13/cobra v1.6.0
 	github.com/spf13/pflag v1.0.5
-	github.com/tailscale/tscert v0.0.0-20220316030059-54bbcb9f74e2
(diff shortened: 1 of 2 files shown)


bradfitz:
cc @mholt

francislavoie (approved):
Thanks!

francislavoie:
As an aside FYI, we've had some users in the Caddy forums ask for a solution to proxying multiple services from the same machine with Caddy + TS, but https://github.com/tailscale/tailscale/issues/3847#issuecomment-1153450685 has been a blocker for that. We typically recommend users serve each service with a subdomain, because many apps don't handle being served in a subpath too well. Just wanted to put that on your radar as something that's come up recently.

mholt:
Hey Brad, thanks! I'm returning to the office tomorrow so I'll check this closer right away then.

PR #5332, opened 2023-01-25
admin: Add `CADDY_ADMIN` env var to override the default

### LINKED ISSUE #5317: Set default admin address via `CADDY_ADMIN` env var
On all platforms, Caddy's default admin API address is `localhost:2019` (see https://caddyserver.com/docs/api).

The reasoning we had for defaulting to using a TCP socket is that is works on all platforms (including Windows). This was a pretty good reason for making things homogeneous, making things predictable for users.

That said, using a TCP socket is not _really_ the best possible default for security, since it's not so easy to control access to the TCP socket for users on the same machine, whereas Unix sockets allow using file permissions to control access. This is helpful for multi-user Linux environments where Caddy might be running.

Generally, the Caddy team is of the opinion that "you shouldn't let untrusted code/users run things on your system". But still, putting mitigations in place is never a bad idea.

So the way I suggest we deal with this is add support for using the `CADDY_ADMIN` environment variable to override the default admin API address. Of course, an explicitly configured admin address in a loaded config will still take priority over both.

Making this change would mean that Linux package maintainers (e.g. apt, rpm, Docker) could make the decision to set a different value for this environment variable to give users better security by default without users having to explicitly configure their admin address to use a Unix socket. They would probably set `CADDY_ADMIN=unix//var/run/caddy.sock` as an improved default.

Whether we actually make the change in the packages (or when we do it) is still to be discussed, since it's a backwards-compatibility break. We'll need to assess whether the break is worth it, and schedule when we want to release the change.

/cc @carlwgeorge @Conan-Kudo @shibumi I'd appreciate your feedback on this as package maintainers. Is this a change you'd like to make to your packages?

/cc @hairyhenderson we could also make this change in Docker, which could make it easier to have another container mount the Unix socket from the Caddy container to communicate with it, instead of the user _having_ to override the admin address themselves to do so.

francislavoie:
Closes #5317

I took the idea of using `init()` for this from here:

https://github.com/caddyserver/caddy/blob/962310204f088a74719a48fa2dc76e76934dfa09/cmd/main.go#L40-L48


mholt (approved):
LGTM

PR #5339, opened 2023-01-28
httpcaddyfile: Add persist_config global option

### LINKED ISSUE #5209: Configurable autosave behaviour from Caddyfile
## What would you like to have changed?
A `Caddyfile` configuration option to enable/disable autosaving the server's configuration state could be useful to reduce some log noise.

## Why is this feature a useful, necessary, and/or important addition to this project?
This is low priority -- it's not hugely necessary or important -- but I noticed some log messages relating to `caddy` attempting to autosave configuration to disk, and from there noticed that the [relevant `allowPersist` variable](https://github.com/caddyserver/caddy/blob/6bad878a22e048762262d6fabe2144cefaf4ca81/caddy.go#L350) is hardcoded in v2.6.2.

For sites where configuration is static and/or configuration writes are not expected/possible, it could make sense to allow that variable to be configurable.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?
Failure to write the autosaved configuration is not a fatal error, so the service continues to run.

u5surf:
* Fixes #5209
* `persist_config off` disables configuration persist explictly in Caddyfile
* if it sets, Admin.Config.Persist becomes false

### DIFF
--- a/caddyconfig/httpcaddyfile/options.go
+++ b/caddyconfig/httpcaddyfile/options.go
@@ -54,6 +54,7 @@ func init() {
 	RegisterGlobalOption("ocsp_stapling", parseOCSPStaplingOptions)
 	RegisterGlobalOption("log", parseLogOptions)
 	RegisterGlobalOption("preferred_chains", parseOptPreferredChains)
+	RegisterGlobalOption("persist_config", parseOptPersistConfig)
 }
 
 func parseOptTrue(d *caddyfile.Dispenser, _ any) (any, error) { return true, nil }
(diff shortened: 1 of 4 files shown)


caddyconfig/httpcaddyfile/options.go
```diff
@@ -386,6 +387,21 @@ func parseOptOnDemand(d *caddyfile.Dispenser, _ any) (any, error) {
...
+	d.Next() // consume parameter name
+	if !d.Next() {
+		return "", d.ArgErr()
+	}
+	val := d.Val()
+	if d.Next() {
+		return "", d.ArgErr()
+	}
+	if val != "off" {
+		return "", d.Errf("persist_config must be one of 'off'")
```
francislavoie, line 400: For now, the wording shouldn't imply there can be other options

```suggestion
		return "", d.Errf("persist_config must be 'off'")
```

francislavoie (approved):
Thanks! Looks good to me :+1:

PR #5351, opened 2023-01-31
cmd: `fmt` exit successfully after overwriting config file

### LINKED ISSUE #5349: caddy fmt --overwrite exit status 1
Why does `caddy fmt --overwrite` exit with status 1 after overwriting the config file?
```
Error: ../../Caddyfile:4: Caddyfile input is not formatted
exit status 1
```
Shouldn't it successfully exit [here](https://github.com/caddyserver/caddy/blob/master/cmd/commandfuncs.go#L560)?
```go
if fl.Bool("overwrite") {
	if err := os.WriteFile(formatCmdConfigFile, output, 0600); err != nil {
		return caddy.ExitCodeFailedStartup, fmt.Errorf("overwriting formatted file: %v", err)
	}
}
```

amis-shokoohi:
Fixes #5349


francislavoie (approved):
Thanks! That looks good.

Please don't forget to sign the CLA! It's necessary for us to be allowed to merge your changes.

PR #5289, opened 2023-01-07
Fix hanging caused by Transfer-Encoding: chunked

u5surf:
* Fixes #5236
* enable request body buffering in reverse proxy
  when the request header has Transfer-Encoding: chunked

Issue #5236, opened 2022-12-06
Transfer-Encoding: chunked causing php_fastcgi to hang


ArabCoders:
Hello, I was attempting to support `Transfer-Encoding: chunked` header into my application, and noticed that the server never responses if the header is used. 

```bash
$ caddy version
v2.6.2 h1:wKoFIxpmOJLGl3QXoo6PNbYvGW4xLEgo32GPBEjWL8o=
```

# Caddyfile

```
{
	debug
	log stdout
	admin 0.0.0.0:2018
	grace_period 1s
}

:8888 {
	log
	root * /home/user/test
	php_fastcgi unix//var/run/php/php8.1-fpm.sock {
		#buffer_requests
		#flush_interval -1
	}
}
```
# php file (index.php)

```php
<?php
echo "i am alive\n";
```

# Example of working request/response

```
$ curl -k http://localhost:8888/ --header "Pragma: no-cache" -X POST -d file=test

i am alive
```

# Caddy Logs

```
2022/12/06 14:41:18.348 DEBUG   http.handlers.rewrite   rewrote request {"request": {"remote_ip": "127.0.0.1", "remote_port": "49996", "proto": "HTTP/1.1", "method": "POST", "host": "localhost:8888", "uri": "/", "headers": {"Content-Length": ["9"], "Content-Type": ["application/x-www-form-urlencoded"], "Cache-Control": ["no-cache"], "User-Agent": ["curl/7.81.0"], "Accept": ["*/*"], "Pragma": ["no-cache"]}}, "method": "POST", "uri": "/index.php"}
2022/12/06 14:41:18.348 DEBUG   http.handlers.reverse_proxy     selected upstream       {"dial": "/var/run/php/php8.1-fpm.sock", "total_upstreams": 1}
[...]

PR #5367, opened 2023-02-10
reverseproxy: Don't buffer chunked requests (fix #5366)

### LINKED ISSUE #5366: Caddy 2.6.3 high memory usage
We are using Caddy on Ubuntu 20.04 on of our servers as a reverse-proxy for our Docker registry and are regularly pulling and pushing Docker images to and from this registry. The server had initially 8GB of memory and everything worked fine with Caddy 2.6.2.

After updating to Caddy 2.6.3 the server regularly runs out of memory when building and publishing new Docker images via our CI system and Caddy gets killed because of that (see below). This still happens even after we increased the server memory to 16GB + 32GB swap.

A downgrade to 2.6.2 fixes the issue.

Did some configuration defaults change in the new version? Is there a way to limit the memory usage? Is there maybe a memory leak in 2.6.3?

```
~$ dmesg -T | egrep -i 'killed process'
[Fri Feb 10 13:45:11 2023] Out of memory: Killed process 792 (caddy) total-vm:45205552kB, anon-rss:14974688kB, file-rss:0kB, shmem-rss:0kB, UID:997 pgtables:86664kB oom_score_adj:0
[Fri Feb 10 13:50:59 2023] Out of memory: Killed process 2876 (caddy) total-vm:45738576kB, anon-rss:14941248kB, file-rss:0kB, shmem-rss:0kB, UID:997 pgtables:86736kB oom_score_adj:0
```

mholt:
Mostly reverts 845bc4d50b437995d574819850206e4b3db4040d (#5289) -- that was my bad, totally an oversight on my part of the code review.

Adds warning for unsafe config. The `buffer_requests` and `buffer_responses` config params, together with a separate `max_buffer_size` parameter, were not safely designed: very prone to configuring unbounded buffers. I'm not sure why I ever thought that was fine.

Deprecates unsafe properties in favor of simpler, safer designed ones.

We'll need a better fix for #5236 -- even if it's the same kind of fix (buffering, and resetting the Content-Length) as long as it's opt-in this time.

/cc @u5surf What do you think?

I think this patch merits a quick release of v2.6.4.

### DIFF
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -521,19 +521,8 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 					h.FlushInterval = caddy.Duration(dur)
 				}
 
-			case "buffer_requests":
-				if d.NextArg() {
-					return d.ArgErr()
-				}
-				h.BufferRequests = true
-
-			case "buffer_responses":
-				if d.NextArg() {
-					return d.ArgErr()
-				}
-				h.BufferResponses = true
-
-			case "max_buffer_size":
+			case "request_buffers", "response_buffers":
+				subdir := d.Val()
 				if !d.NextArg() {
 					return d.ArgErr()
 				}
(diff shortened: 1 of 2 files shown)


mholt:
I could remove `isChunkedRequest()` to satisfy the linter, but I left it in there because I want to use it to apply @u5surf's earlier patch, but more opt-in. I am just not sure how to go about that. Would love any ideas/insights!

francislavoie (approved):
Seems good to me.

I think we can just deleted the chunked encoding code. If users who that fix helped find that they need buffering, they can enable it for themselves (and now, set limits).

u5surf:
@mholt @
Sorry for late.
I also agree with not to buffering such a large chunked contents. I lost my sight at this point because I misunderstood that it has already taken this problem in h.bufferedBody or somewhere.
Your patch looks good to me. Thank
 you

mholt:
@u5surf No worries, thank you for checking it out! And thank you for the original patch. The mistake was my oversight. Hope you'll contribute again.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -521,19 +521,39 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+					return d.Errf("invalid byte size '%s': %v", d.Val(), err)
+				}
 				if d.NextArg() {
 					return d.ArgErr()
 				}
-				h.BufferRequests = true
+				if subdir == "request_buffers" {
+					h.RequestBuffers = int64(size)
+				} else if subdir == "response_buffers" {
+					h.RequestBuffers = int64(size)
```
u5surf, line 539: @mholt 
```suggestion
					h.ResponseBuffers = int64(size)
```
??

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -622,9 +645,8 @@ func (h Handler) prepareRequest(req *http.Request, repl *caddy.Replacer) (*http.
 	// attacks, so it is strongly recommended to only use this
 	// feature if absolutely required, if read timeouts are
 	// set, and if body size is limited
-	if (h.BufferRequests || isChunkedRequest(req)) && req.Body != nil {
-		req.Body, req.ContentLength = h.bufferedBody(req.Body)
-		req.Header.Set("Content-Length", strconv.FormatInt(req.ContentLength, 10))
+	if h.ResponseBuffers != 0 && req.Body != nil {
```
u5surf, line 648: @mholt 
I consider that this branch decides whether the proxy buffers the request body or not. Thus we should look `h.RequestBuffers` .

```suggestion
	if h.RequestBuffers != 0 && req.Body != nil {
```

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -622,9 +645,8 @@ func (h Handler) prepareRequest(req *http.Request, repl *caddy.Replacer) (*http.
...
+		req.Body, _ = h.bufferedBody(req.Body, h.ResponseBuffers)
```
u5surf, line 649: And so on that
```suggestion
		req.Body, _ = h.bufferedBody(req.Body, h.RequestBuffers)
```

u5surf (commented):
@mholt 
I have left a few comment, I still might have confused for me the new configurations.

mholt:
@u5surf Ah wow, thank you for your review -- I had a lot going on yesterday and looks like I got distracted during a search-replace and mixed them up :see_no_evil: 

Thank you for catching those errors! I've pushed what I hope is the final change for this patch.

PR #5359, opened 2023-02-05
reverseproxy: consult `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` in reverse proxy

mohammed90:
Closes #5178

PR #5369, opened 2023-02-12
reverseproxy: refactor HTTP transport layer

### LINKED ISSUE #5178: HTTP_PROXY env not working
Using the docker `caddy:2-alpine` image
```
Linux 4d5a25798135 5.10.104-linuxkit #1 SMP PREEMPT Thu Mar 17 17:05:54 UTC 2022 aarch64 Linux
```

```
v2.6.2 h1:wKoFIxpmOJLGl3QXoo6PNbYvGW4xLEgo32GPBEjWL8o=
```

### 2a. What happens (briefly explain what is wrong)
For a whole week now I'm trying to setup caddy as reverse proxy to proxy requests over tailscale into my private home network.
Over time my minimal reproducable case has evolved into using tinyproxy except of the tailscale proxy from the [userspace networking mode](https://tailscale.com/kb/1112/userspace-networking/) and also currently I'm testing everything from within the docker container itself. Thats why currently the caddy file binds to `http://172.17.0.2` except for an external domain.

I have gotten multiple successful curl requests by specifying the proxy
```sh
// using the tailscale proxy
ALL_PROXY=localhost:3128 curl http://100.81.184.38:3000 //or www.google.com if you don't have a tailscale node
```

But no matter if I use `ALL_PROXY` `http_proxy` or `HTTP_PROXY` I always fail to use the proxy through caddy. Help!

### 2c. Log output

```
http.log.error	dial tcp 100.81.184.38:3000: connect: connection refused	request": {"remote_ip": "172.17.0.2", "remote_port": "38404", "proto": "HTTP/1.1", "method": "GET", "host": "172.17.0.2", "uri": "/api", "headers": {"User-Agent": ["curl/7.83.1"], "Accept": ["*/*"]}}, "duration": 0.041261001, "status": 502, "err_id": "wv1ugcp6m", "err_trace": "reverseproxy.statusError (reverseproxy.go:1272)"}
```

### 2e. Related issues
I know that this seems similar to #1433 and I know that this issue ended because of no minimal reproduction case. Thats why I already put some work in, to put everything into a Dockerfile and Repo so it should be reproducible across machines

### 2d. Related Links
For the tailscale case, [this doc](https://tailscale.com/kb/1107/heroku/) is helpful because I have to use tailscale in [userspace networking mode](https://tailscale.com/kb/1112/userspace-networking/) because of my deployment server

Reproduction Repo: https://github.com/8BitJonny/caddy-proxy-minimal-repro/tree/main with Docker.

This reproduction repo is setup to show my struggle to use caddy with a proxy, with two different proxies. You can run the docker container and reproduce it with the [tailscale](https://tailscale.com/) proxy or with the most minimal proxy using [tinyproxy](https://github.com/tinyproxy/tinyproxy)

Use the repo and run 
```sh
docker build -t caddy_proxy:latest .
docker run -ti -p 43890:43890 --env TAILSCALE_AUTHKEY=TailScale-ApiKey --name caddy caddy_proxy
```
to start the docker container. You'll only need the `TAILSCALE_AUTHKEY` env if you want to reproduce it with tailscale. Currently its setup for reproduction with tinyproxy and then the env won't be used.

And in the second terminal the following fails
```sh
docker exec -it caddy /bin/sh
curl http://172.17.0.2/api //This errors with either proxy successfully forwarding the request
```

mohammed90:
The earlier implementation smuggles the upstream dial info through context. It made it difficult to accommodate utilizing the standard env vars `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` for upstream connections. It also hurt connection caching. The earlier hack was to accommodate the case of upstreams being Unix sockets. The alternative approach is to register a custom protocol with a custom Transport, which is provided by the package [github.com/peterbourgon/unixtransport](https://pkg.go.dev/github.com/peterbourgon/unixtransport#section-readme), and let it manage the case of unix sockets. This PR fixes the mentioned pain points of the earlier implementation.

Fixes #5178

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -126,6 +126,7 @@ func (h *HTTPTransport) Provision(ctx caddy.Context) error {
 	if err != nil {
 		return err
 	}
+	rt.Proxy = http.ProxyFromEnvironment
 	h.Transport = rt
 
 	return nil
@@ -172,38 +173,7 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
 		}
 	}
 
-	// Set up the dialer to pull the correct information from the context
-	dialContext := func(ctx context.Context, network, address string) (net.Conn, error) {
-		// the proper dialing information should be embedded into the request's context
-		if dialInfo, ok := GetDialInfo(ctx); ok {
-			network = dialInfo.Network
-			address = dialInfo.Address
-		}
-
-		conn, err := dialer.DialContext(ctx, network, address)
-		if err != nil {
-			// identify this error as one that occurred during
-			// dialing, which can be important when trying to
-			// decide whether to retry a request
-			return nil, DialError{err}
-		}
-
-		// if read/write timeouts are configured and this is a TCP connection, enforce the timeouts
-		// by wrapping the connection with our own type
-		if tcpConn, ok := conn.(*net.TCPConn); ok && (h.ReadTimeout > 0 || h.WriteTimeout > 0) {
-			conn = &tcpRWTimeoutConn{
-				TCPConn:      tcpConn,
-				readTimeout:  time.Duration(h.ReadTimeout),
-				writeTimeout: time.Duration(h.WriteTimeout),
-				logger:       caddyCtx.Logger(),
-			}
-		}
-
-		return conn, nil
-	}
-
 	rt := &http.Transport{
-		DialContext:            dialContext,
 		MaxConnsPerHost:        h.MaxConnsPerHost,
 		ResponseHeaderTimeout:  time.Duration(h.ResponseHeaderTimeout),
 		ExpectContinueTimeout:  time.Duration(h.ExpectContinueTimeout),
@@ -252,7 +222,7 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
 		h2t := &http2.Transport{
 			// kind of a hack, but for plaintext/H2C requests, pretend to dial TLS
 			DialTLSContext: func(ctx context.Context, network, address string, _ *tls.Config) (net.Conn, error) {
-				return dialContext(ctx, network, address)
+				return dialer.DialContext(ctx, network, address)
 			},
 			AllowHTTP: true,
 		}
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -174,6 +174,7 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
 	}
 
 	rt := &http.Transport{
+		DialContext: dialer.DialContext,
 		MaxConnsPerHost:        h.MaxConnsPerHost,
 		ResponseHeaderTimeout:  time.Duration(h.ResponseHeaderTimeout),
 		ExpectContinueTimeout:  time.Duration(h.ExpectContinueTimeout),
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -126,7 +126,6 @@ func (h *HTTPTransport) Provision(ctx caddy.Context) error {
 	if err != nil {
 		return err
 	}
-	rt.Proxy = http.ProxyFromEnvironment
 	h.Transport = rt
 
 	return nil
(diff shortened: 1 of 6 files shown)


mohammed90:
Oh, @mholt, this change loses the TCP connection timeouts 😕 I'm not sure how yet to retrofit that into the current change. Let me know if you have any ideas or to brainstorm something.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -172,38 +173,7 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
...
-				readTimeout:  time.Duration(h.ReadTimeout),
-				writeTimeout: time.Duration(h.WriteTimeout),
-				logger:       caddyCtx.Logger(),
-			}
-		}
-
-		return conn, nil
-	}
-
 	rt := &http.Transport{
```
francislavoie, line 176: I think we still need to use our `dialer` because we customize it for DNS and such. If we leave it full-default then we lose all that too.

```suggestion
	rt := &http.Transport{
		DialContext: func(ctx context.Context, network, address string) (net.Conn, error) {
			return dialer.DialContext(ctx, network, address)
		}
```

Is this what breaks conn pooling? :thinking:

WeidiDeng: I believe DNS resolve is done lines before (by specifying resolver field of dialer), this part deal with unix dialing.

francislavoie: But we don't use the `dialer` after that though. We configure it then never use it. This deleted code is where it was actually used.

WeidiDeng: My bad, dialerContext is your suggested change:sweat_smile:.

WeidiDeng: It can be simplified as dialer.DialContext without creating a new function though.

WeidiDeng:
Why is connection pooling hurt from original implementation? Are there any reports?

WeidiDeng:
@mohammed90 In slack you mentioned `It doesn't work`, does it mean adding proxy field won't work because of DialContext address smuggling? I'll do some local testing to find out.

mohammed90:
> @mohammed90 In slack you mentioned `It doesn't work`, does it mean adding proxy field won't work because of DialContext address smuggling? I'll do some local testing to find out.

You're referring to an old conversation from a few weeks. This is not the same PR.

mohammed90:
> @mohammed90 In slack you mentioned `It doesn't work`, does it mean adding proxy field won't work because of DialContext address smuggling? I'll do some local testing to find out.

The `doesn't work` refers to the earlier implementation, not this PR. As said, it's because of the request smuggling through `context.Context`. We used to dial upstream directly, ignoring any logic deferring to the environment HTTP proxy. You can refer to the PR #5359 for the earlier attempt where it didn't work. It works in this PR.

> Why is connection pooling hurt from original implementation? Are there any reports?

After going through your reasoning in our Slack discussion, I realize I was mistaken 😅 Connection pooling wasn't hurt in the earlier implementation.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -172,38 +173,8 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
 		}
 	}
 
-	// Set up the dialer to pull the correct information from the context
-	dialContext := func(ctx context.Context, network, address string) (net.Conn, error) {
-		// the proper dialing information should be embedded into the request's context
-		if dialInfo, ok := GetDialInfo(ctx); ok {
-			network = dialInfo.Network
-			address = dialInfo.Address
-		}
```
francislavoie, line 181: Losing this code means placeholder dial addresses will no longer work.

francislavoie: Hmm... I don't understand how though. Where is the dial address being passed through the replacer? The only place I see it happens is in `fillDialInfo`.

mohammed90: Yes. And then the request is updated in `h.reverseProxy` which calls `h.directRequest` with the request and the dial info. That subsequently updates the request with the details from DialInfo.

mholt:
This looks great, actually -- thanks for the PR!

I'm curious how this works for people actually proxying to unix sockets; I have only done so here and there for testing purposes. Maybe we can get someone to try it in a real deployment?

WeidiDeng:
> This looks great, actually -- thanks for the PR!
> 
> I'm curious how this works for people actually proxying to unix sockets; I have only done so here and there for testing purposes. Maybe we can get someone to try it in a real deployment? (Doesn't have to be a crazy busy/important one.)
> 
> Then once the linter is satisfied we can probably look at merging this in :)

No, unix socket will bypass settings of http.transport because we already registered a new protocol (http+unix). It's up to unix transport to handle proxy (which is no-op).

WeidiDeng:
I still think we should rewrite address only when network is unix, instead of importing a new dependency. This dependency encodes unix path as base64 under the [hood](https://github.com/peterbourgon/unixtransport/blob/main/register.go#L93).

francislavoie:
I reworked the PR; mostly reverted what was there, and implemented the fix @WeidiDeng recommended in Slack.

Basically, we should only override the dial address from the context dialInfo if the context had `unix` as the network.

Integration tests pass, and I confirmed once again that the unix socket test fails if we _don't_ override the dial address.

mohammed90:
> I still think we should rewrite address only when network is unix, instead of importing a new dependency. This dependency encodes unix path as base64 under the [hood](https://github.com/peterbourgon/unixtransport/blob/main/register.go#L93).

The reason it encodes the path in basae64 is because, more or less, files/directories names are not limited in any way on Linux [^1] [^2]. Notice that it encodes it in the URL charset of base64 because URLs accept only a certain set of characters. So it encodes it in base64 to pass through the URL parsing of Go stdlib while retaining the socket path and request path, then decodes it before dialing.

Anyways, as long as we get it working, the approach is irrelevant to me. Thanks for the review and the rework!

[^1]: https://dwheeler.com/essays/fixing-unix-linux-filenames.html
[^2]: https://lwn.net/Articles/71472/

sekamasr (approved):

mholt (approved):
This change LGTM.

Thank you everyone for collaborating on it!

PR #5378, opened 2023-02-16
cmd: make `caddy fmt` hints more clear

IndeedNotJames:
An impressive amount of users misunderstand `caddy fmt`. 

This (tiny) PR slightly rephrases the hints to mention `--overwrite` (which a lot of users are unaware of) in an attempt to make it more clear what this command actually does (for users that don't check help pages that is)


mholt (approved):
This is a great change. Thanks

PR #5379, opened 2023-02-17
cmd: Expand cobra support, add short flags

francislavoie:
We've had Cobra in for a few versions now, and it's worked pretty well. Now, it's time to make it possible to use more of its features.

For that, I'm adding a new CobraFunc field to `caddycmd.Command` which allows configuring the command using the full power of Cobra's APIs while keeping backwards compatibility and ergonomics of our `RegisterCommand` setup. 

In addition, I ported every built-in command to use this new API (making sure commands still worked after the first commit), and added shorthands for most flags. Some of them are a bit funky due to the english language only having so many letters in the alphabet, but I think it should be reasonable. Let me know if you find any of them too weird. We could also just play it safe and enable _less_ shorthands to start, not to paint ourselves into a corner with exhausting letters ahead of time :joy:


cmd/commands.go
```diff
@@ -190,22 +191,20 @@ Since the admin endpoint is configurable, the endpoint configuration is loaded
...
+		},
 	})
 
 	RegisterCommand(Command{
 		Name:  "version",
-		Func:  cmdVersion,
 		Short: "Prints the version",
 		Long: `
-Prints the version of this Caddy binary.
+		Prints the version of this Caddy binary.
```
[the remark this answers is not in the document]
francislavoie: Not sure how that happened :thinking: Thanks for spotting it

modules/caddyhttp/staticresp.go
```diff
@@ -71,16 +70,15 @@ Access/request logging and more verbose debug logging can also be enabled.
...
-			fs.Bool("access-log", false, "Enable the access log")
-			fs.Bool("debug", false, "Enable more verbose debug-level logging")
-			fs.Var(&respondCmdHeaders, "header", "Set a header on the response (format: \"Field: value\"")
-			return fs
-		}(),
+		CobraFunc: func(cmd *cobra.Command) {
+			cmd.Flags().StringP("listen", "l", ":0", "The address to which to bind the listener")
+			cmd.Flags().IntP("status", "s", http.StatusOK, "The response status code")
+			cmd.Flags().StringP("body", "b", "", "The body of the HTTP response")
+			cmd.Flags().BoolP("access-log", "a", false, "Enable the access log")
```
mholt, line 77:
```suggestion
			cmd.Flags().BoolP("access-log", "", false, "Enable the access log")
```

modules/caddyhttp/staticresp.go
```diff
@@ -71,16 +70,15 @@ Access/request logging and more verbose debug logging can also be enabled.
...
-			fs.String("body", "", "The body of the HTTP response")
-			fs.Bool("access-log", false, "Enable the access log")
-			fs.Bool("debug", false, "Enable more verbose debug-level logging")
-			fs.Var(&respondCmdHeaders, "header", "Set a header on the response (format: \"Field: value\"")
-			return fs
-		}(),
+		CobraFunc: func(cmd *cobra.Command) {
+			cmd.Flags().StringP("listen", "l", ":0", "The address to which to bind the listener")
+			cmd.Flags().IntP("status", "s", http.StatusOK, "The response status code")
+			cmd.Flags().StringP("body", "b", "", "The body of the HTTP response")
```
mholt, line 76:
```suggestion
			cmd.Flags().StringP("body", "b", "", "The body of the HTTP response")
```

I'm torn on this one, because it almost makes sense to have some parity with `curl`, like `-d` for body, for example. `-b` feels like 'binary'. Hmm.

modules/caddypki/command.go
```diff
@@ -53,19 +52,17 @@ This command will attempt to connect to Caddy's admin API running at
...
-		Flags: func() *flag.FlagSet {
-			fs := flag.NewFlagSet("trust", flag.ExitOnError)
-			fs.String("ca", "", "The ID of the CA to trust (defaults to 'local')")
-			fs.String("address", "", "Address of the administration API listener (if --config is not used)")
-			fs.String("config", "", "Configuration file (if --address is not used)")
-			fs.String("adapter", "", "Name of config adapter to apply (if --config is used)")
-			return fs
-		}(),
+		CobraFunc: func(cmd *cobra.Command) {
+			cmd.Flags().StringP("ca", "i", "", "The ID of the CA to trust (defaults to 'local')")
```
mholt, line 56:
```suggestion
			cmd.Flags().StringP("ca", "", "", "The ID of the CA to trust (defaults to 'local')")
```

The long form of this flag is two letters. :) (and the short form doesn't make as much sense IMO)

modules/caddypki/command.go
```diff
@@ -53,19 +52,17 @@ This command will attempt to connect to Caddy's admin API running at
...
+			cmd.Flags().StringP("address", "d", "", "Address of the administration API listener (if --config is not used)")
```
mholt, line 57:
```suggestion
			cmd.Flags().StringP("address", "", "", "Address of the administration API listener (if --config is not used)")
```

modules/caddypki/command.go
```diff
@@ -89,15 +86,14 @@ will attempt to connect to the Caddy's admin API running at
...
-			fs.String("cert", "", "The path to the CA certificate to untrust")
-			fs.String("ca", "", "The ID of the CA to untrust (defaults to 'local')")
-			fs.String("address", "", "Address of the administration API listener (if --config is not used)")
-			fs.String("config", "", "Configuration file (if --address is not used)")
-			fs.String("adapter", "", "Name of config adapter to apply (if --config is used)")
-			return fs
-		}(),
+		CobraFunc: func(cmd *cobra.Command) {
+			cmd.Flags().StringP("cert", "p", "", "The path to the CA certificate to untrust")
+			cmd.Flags().StringP("ca", "i", "", "The ID of the CA to untrust (defaults to 'local')")
```
mholt, line 91:
```suggestion
			cmd.Flags().StringP("ca", "", "", "The ID of the CA to untrust (defaults to 'local')")
```

modules/caddypki/command.go
```diff
@@ -89,15 +86,14 @@ will attempt to connect to the Caddy's admin API running at
...
+			cmd.Flags().StringP("address", "d", "", "Address of the administration API listener (if --config is not used)")
```
mholt, line 92:
```suggestion
			cmd.Flags().StringP("address", "", "", "Address of the administration API listener (if --config is not used)")
```

mholt (changes requested):
Thanks for this change!

Overall I like it. I wish we didn't have to hack in the Cobra API like that, but oh well.

I made many suggestions regarding short flags. I lean toward what you suggested: let's start with fewer short flags, until we determine where additional ones would be useful. I suggested removing them on the less-common flags, or ones where there was a difficult conflict and where I could see multiple different letters being good options, or where a clear choice wasn't obvious.

Hope that's OK. Let me know what you think!

modules/caddyhttp/reverseproxy/command.go
```diff
@@ -63,17 +62,17 @@ For proxying:
...
-			fs.Bool("insecure", false, "Disable TLS verification (WARNING: DISABLES SECURITY BY NOT VERIFYING TLS CERTIFICATES!)")
-			fs.Bool("internal-certs", false, "Use internal CA for issuing certs")
-			fs.Bool("debug", false, "Enable verbose debug logs")
-			fs.Bool("disable-redirects", false, "Disable HTTP->HTTPS redirects")
-			return fs
-		}(),
+		CobraFunc: func(cmd *cobra.Command) {
+			cmd.Flags().StringP("from", "f", "localhost", "Address on which to receive traffic")
+			cmd.Flags().StringSliceP("to", "t", []string{}, "Upstream address(es) to which traffic should be sent")
+			cmd.Flags().BoolP("change-host-header", "c", false, "Set upstream Host header to address of upstream")
```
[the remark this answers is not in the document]
francislavoie: I think we'll want to leave `-H` for `--header-up` as per the other recent open issue

modules/caddyhttp/staticresp.go
francislavoie: `-d` makes no sense to me. Is that supposed to be "data"? Should we rename the flag in that case? In this case, I think the body is core to this command's usage, so having an easy way to set it makes a lot of sense.

mholt: Yeah, data.

I agree it should have a short flag. -b is fine, but -d is what comes to mind first for me, since `curl` (the de-facto command for HTTP requests) uses -d.

modules/caddyhttp/reverseproxy/command.go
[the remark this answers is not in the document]
mholt: Sounds good. If we do `--header-up` and `--header-down` (or `--header`), I think it'd be kinda "cute" if the short flags were `-H` for `--header-up` and `-h` for `--header-down` (like upper and lower case).

If we don't end up doing that, maybe this should be `-h` instead.

If we aren't sure, maybe we should omit the short flag for this for now. (Even though it's probably a common one.)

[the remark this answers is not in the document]
francislavoie: That's not possible, actually, because Cobra hijacks `-h` for `--help` everywhere.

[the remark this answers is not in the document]
mholt: Oh. In that case, I guess `-c` is good as any letter. Thanks!

mholt (approved):
Thanks for working on this!

modules/caddyhttp/staticresp.go
IndeedNotJames: I just want to mention `curl`'s long flag for `-d` is `--data` (along with other variants of that option following `--data-*` like `--data-urlencode` and `--data-raw`)

PR #5383, opened 2023-02-20
cmd: Fix validating JSON with unknown fields

francislavoie:
Fix #5382

PR #5384, opened 2023-02-20
caddytls: Loosen the warning for on-demand

francislavoie:
Closes #5383 

This warns about on-demand being vulnerable only if the automation policy is the default fallback (no subjects) or has any wildcard certs, because those are infinite cardinality.

If there's only single (non-wildcard) subjects, then the cardinality is limited to only the length of the subject list, so there's no risk of abuse.


mholt (approved):
Sure, I guess we can do this. Probably makes more sense this way. Thanks!

PR #5385, opened 2023-02-20
caddyfile: Implement heredoc support

### LINKED ISSUE #5357: Allow multiline strings syntax
## The problem

What I love with Caddy is how simple the syntax and getting everything done is. I migrated from nginx and Apache and never looked back. However, the lack of syntax for handling multilines ends up in ugly looking and harder to main Caddyfiles.

This was discussed on a few places but I did not see feature requests. We need this to be able to [respond with multiline strings](https://caddy.community/t/respond-with-multi-line-body/12730) and multiline headers (#2302).

The current efforts look like hack, not really self-documenting and does not always work as expected. The syntax is not intuitive.

Examples:

```caddyfile
	respond "/robots.txt" 200 {
		body "User-agent: *
Disallow: /" # 
		close
	}
```

Example above requires an indentation that breaks the natural flow of the code.

```caddyfile
Content-Security-Policy "default-src 'self'; img-src 'self' data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; object-src 'none'; frame-ancestors 'none'; upgrade-insecure-requests; block-all-mixed-content;"
```

Example above requires a a long line, looks ugly, harder to maintain and read. Having it multiline does not work (#2302).

## Needed solution

So two ways needs to be supported:

1. Newlines where `\n` will be preserved.
2. Newlines where `\n` will not be preserved.

## Alternative solution: yaml inspired syntax

I see that heredoc format was not the favorite #3802. But how about using a yaml like syntax?

**1. To preserve `\n`s we can use `|` and `|-`**

```caddyfile
body |-
  User-agent:
  Disallow:
```

**2. To not preserve `\n`s we can use `>` and `>-`**

We can have something like this:

```txt
Content-Security-Policy >-
	default-src 'self';
	img-src 'self' data:;
	script-src 'self' 'unsafe-inline';
	style-src 'self' 'unsafe-inline';
	form-action 'self';
	object-src 'none';
	frame-ancestors 'none';
	upgrade-insecure-requests;
	block-all-mixed-content;
```

francislavoie:
I'm reopening my old feature branch from https://github.com/caddyserver/caddy/pull/3802 since I still think this should be implemented. Closes #5357

I've made some improvements to the implementation, some of which were possible because of improvements to the lexer we've made since 2020 when I first tried this.

This branch has a few commits at the end of the branch with some unrelated changes, mainly cleanup to some Caddyfile parsing/dispensing code that I noticed I could improve while reading around.

Copying the text from my previous PR:

---

Adds support for heredoc syntax to the Caddyfile, inspired by HCL (HashiCorp Config Language), and PHP 7.3's flexible syntax (trimming the front of every line based on the indentation of the ending marker: https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes)

Here's an example:

```
example.com {
	respond <<EOF
	<html>
	  <head><title>Foo</title>
	  <body>Foo</body>
	</html>
	EOF 200
}
```

Adapted JSON:

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":443"
          ],
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "example.com"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "<html>\n  <head><title>Foo</title>\n  <body>Foo</body>\n</html>\n",
                          "handler": "static_response",
                          "status_code": 200
                        }
                      ]
                    }
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    }
  }
}
```

Notice that the response body has the leading whitespace on each line stripped away.

---

This could use some more tests in `lexer.go` before merging. There is an adapt test which is enough at a minimum for the moment.


mholt (approved):
Alright, despite my personal feelings about heredoc, I hereby accept that it is inevitable. :stuck_out_tongue_winking_eye: This looks like a fairly decent change.

What happens if the heredoc isn't terminated?

francislavoie:
Oh yeah, good point. Forgot to handle the case where we run out of runes to lex but we didn't find the heredoc marker. Adding a condition for that.

I still want to add some more tests to this before merging. I'll do that next chance I get.

unoukujou:
> Closes #5357

I don't think it's closed because the issue described there is the same issue I face... Long CSP Header cannot have newline and the heredoc is preserving newline characters.

So I must keep it on one long hard to read and hard to maintain single line :\

Issue #4893, opened 2022-07-17
Rename AppIsConfigured -> AppIfConfigured?


mholt:
The `caddy.Context.AppIsConfigured()` method returns true if the named app is present in the config. This is useful to know if an app has been explicitly configured, because calling `App()` will instantiate an app module even if it's not configured.

However, if the primary use case is to get an app instance _only_ if it has been configured, then maybe we should actually have a method, `AppIfConfigured()` instead. One-letter change.

This method would return nil if the app is not configured, so if you need the bool value, there you have it; otherwise it returns the configured app module. It would be less clunky than this:

https://github.com/caddyserver/caddy/blob/ad3a83fb9169899226ce12a61c16b5bf4d03c482/modules/caddypki/adminapi.go#L55-L64

by using just 3 lines of code instead of 6.

francislavoie:
I think I added this. No complaints from me. I tend to write code returning a boolean for fast checks instead of the actual ref to the thing because it makes writing a condition more natural if you have to do like `a.ctx.AppIsConfigured("pki") && someOtherCondition` but either way is fine and it has so few usages it doesn't matter too much.

mholt:
Thanks Francis :blush:

PR #5397, opened 2023-02-26
context: Rename func to `AppIfConfigured`

francislavoie:
Closes https://github.com/caddyserver/caddy/issues/4893

I highly doubt anyone but ourselves used this API, so it should be safe to do this even though it's an exported function.


mholt (approved):
Great; yeah, I can 99% guarantee that nobody else is using this function.

gaukas:
> 99% guarantee that nobody else is using this function.

Great. I'm in that 1%. 🤣

mholt:
Ah, sorry. Since we don't have telemetry all we can do is assume based on community interactions. Hopefully an easy change at least.

PR #5398, opened 2023-02-26
reverseproxy: Health as a ratio of successful requests

### LINKED ISSUE #4949: dynamic passive health check
Is it possible to provide ratio of unhealthy request to all requests instead of unhealthy_request_count? i.e. I want to configure if >=1% of requests are unhealthy to upstream then it should be considered down.

francislavoie:
Closes #4949

This is WIP.

This makes it possible to keep track of successful requests over time for each upstream. Same mechanism as failure duration.

This also adds a minimum successful requests ratio, so if there's too many failures, the upstream is marked unhealthy. I think we might need a minimum total so that the ratio only applies after there's at least N requests in memory, otherwise a single failure at the start might take out the server. But maybe that's fine if the fail duration is low.

### DIFF
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -17,6 +17,7 @@ package caddyhttp
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -164,7 +165,7 @@ func (ws *WeakString) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-// MarshalJSON marshals was a boolean if true or false,
+// MarshalJSON marshals as a boolean if true or false,
 // a number if an integer, or a string otherwise.
 func (ws WeakString) MarshalJSON() ([]byte, error) {
 	if ws == "true" {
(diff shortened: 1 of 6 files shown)


mholt:
Thanks for working on this. My main concern after a first pass through the code (which I know is still a draft) is that it requires 1 goroutine per _successful_ request. This is similar to failure counting, but if we assume the majority of requests to be successful, then this will use significantly more resources than counting failures.

What I've typically leaned on instead is to have a ring buffer or a rolling/online algorithm depending on what the needs are. So for example, when I need to compute the standard deviation over a sliding window, I've used [Welford's Online Algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) which means I don't have to iterate the whole data set each time a new point is discovered.

I wonder if we could use something similar for this calculation, where we instead maybe keep the last N data points (if needed) or heck, maybe just a moving average? https://en.wikipedia.org/wiki/Moving_average

francislavoie:
Yeah, I'm definitely concerned about goroutine spam with this approach.

How would a rolling average type situation "forget" about old requests though? If there's no mechanism to remove old counts/entries then if you fall below the ratio, there's no way for the upstream to be healthy again.

mholt:
> How would a rolling average type situation "forget" about old requests though?

One way is by configuring a window size, e.g. the last N requests. One way to do that is to have an array of size N and you just iterate it, round-robin style; but this requires iterating the entire array at each request to compute an average. An on-line average computation shouldn't require that extra pass through the data IIRC.

I need to find a little time to double-check / look this up, but I can probably recommend something along those lines.

mholt:
To clarify: I think in order for this to be more comfortable for me to merge, we'd want to replace the goroutine-based approach for counting with some sort of other structure that will be less complicated and less likely to use a lot of memory.

francislavoie:
I have no idea how you're thinking that can be done. This is the only idea I can come up with.

francislavoie:
I'll close as inactive. The idea wasn't fully fleshed out.

PR #5401, opened 2023-02-26
reverseproxy: allow specifying ip version for dynamic `a` upstream

### LINKED ISSUE #5392: Limit `a` dynamic upstream source to specific IP versions
As follow-up to https://github.com/caddyserver/caddy/pull/4470, I would like to have an `aaaa` source.

Here is the use-case:
- Caddy runs on an IPv4-accessible server, proxying to IPv6-accessible servers
- Rather than adding entries for each domain one by one, Caddy could just resolve the incoming domain to the right server, if I could restrict it to only use IPv6-servers

Currently, if I use an `a` dynamic upstream, it sometimes resolves to the IPv4, leading to complications.

IndeedNotJames:
Closes #5392

Note: I copied the Caddyfile parsing from @francislavoie's work at https://github.com/mholt/caddy-dynamicdns/pull/16 :^)


caddytest/integration/caddyfile_adapt/reverse_proxy_dynamic_upstreams.txt
```diff
@@ -49,7 +50,8 @@
 									"dynamic_upstreams": {
 										"name": "foo",
 										"port": "9000",
-										"source": "a"
+										"source": "a",
+										"versions": {}
```
francislavoie, line 54: I think the adapter shouldn't output `versions` unless it's specified.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -1397,8 +1398,35 @@ func (u *AUpstreams) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+				}
+
+				// Set up defaults; if versions are specified,
+				// both versions start as false, then flipped
+				// to true otherwise.
+				falseBool := false
+				u.Versions = ipVersions{
+					IPv4: &falseBool,
+					IPv6: &falseBool,
+				}
```
francislavoie, line 1414: Doing this, the one that isn't in the Caddyfile config still appears in the JSON output. I think this should only set the version to true if it appears in the config.

Essentially, you should do:
```
if u.Versions == nil {
	u.Versions = ipVersions{}
}
```

Which leaves them nil, and then in the for-switch just set entries to true.

IndeedNotJames: woops, my bad.
Thanks for clarifying!


Will do

francislavoie (approved):
Great, thanks! :tada:

PR #5403, opened 2023-02-26
ci/cd: ship tarballs with vendored deps

### LINKED ISSUE #4573: Consider shipping release tarball with vendored dependencies
For example, [dnscrypt-proxy](https://github.com/DNSCrypt/dnscrypt-proxy/tree/master/vendor) does this by simply maintaining vendor directory. You can also setup CI to vendor dependencies into release tarball without needing to keep vendor directory within repository. Here is the command that can be used for this: `go mod vendor`

With vendored dependencies, i can build caddy from source without internet connection by using this command: `go build -mod=vendor`

mohammed90:
This adds 2 tarballs in the release:
- `caddy_{{ .Version }}_buildable-artifact.tar.gz`: this contains a `main.go`, `go.mod`, and `go.sum` along with a `vendor/` tree containing all the deps, including Caddy's own source tree as a dep. This allows the user to run `go build` against the exact files used to build our releases.
- `caddy_{{ .Version }}_src`: this packages Caddy's git tree as-is along with the `vendor/` directory of vendored deps.

To test, run:
`TAG=master goreleaser release --skip-publish --snapshot --clean`

Closes #4573

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -4,7 +4,9 @@ before:
     # This is so we can run goreleaser on tag without Git complaining of being dirty. The main.go in cmd/caddy directory 
     # cannot be built within that directory due to changes necessary for the build causing Git to be dirty, which
     # subsequently causes gorleaser to refuse running.
-    - rm -rf caddy-build caddy-dist
+    - rm -rf caddy-build caddy-dist vendor
+    # vendor Caddy deps
+    - go mod vendor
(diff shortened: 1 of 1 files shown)


francislavoie:
Nice. Those get signed too, right? I think that's all still automated on top of that?

francislavoie (approved):

mholt:
Cool, thank you for doing this!

PR #5405, opened 2023-02-28
caddyhttp: Fix shutdown delay on unix systems

### LINKED ISSUE #5393: Slow config reload when using shutdown_delay
We're currently using a `shutdown_delay` of 15 seconds to give our hardware load balancer warning of when caddy is about to go down for drains/maintenance etc as per https://caddyserver.com/docs/caddyfile/options#shutdown-delay - but this also seems to delay the config reload by the same amount. ~~During the shutdown period our load balancer marks the caddy servers offline (as it should) so we lose ingress to the swarm services until they come back online and pass the healthcheck again.~~

Should/can a reload ignore this delay?

image

francislavoie:
Fixes #5393

Since the listener logic in Caddy was rewritten to no longer use the listener pool, using it as a way to check whether we're about to shut down no longer makes sense.

Instead, we can simply use an atomic HTTP app ref counter to detect if we're reloading or shutting down.

Deprecating `caddy.ListenerUsage` because it no longer works on unix platforms; should still work on other platforms, but its pretty much useless now.

I bumped the shutdown and grace log messages (when they trigger) from DEBUG to INFO for better visibility.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
@@ -337,6 +338,9 @@ func (app *App) Validate() error {
 // Start runs the app. It finishes automatic HTTPS if enabled,
 // including management of certificates.
 func (app *App) Start() error {
+	// increment the HTTP app reference count
+	atomic.AddInt32(&appRefCount, 1)
+
 	// get a logger compatible with http.Server
 	serverLogger, err := zap.NewStdLogAt(app.logger.Named("stdlib"), zap.DebugLevel)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/app.go
```diff
@@ -493,31 +497,31 @@ func (app *App) Start() error {
...
-						app.logger.Debug("listener closing and shutdown delay is configured", zap.String("address", addr.String()))
-						server.shutdownAtMu.Lock()
-						server.shutdownAt = scheduledTime
-						server.shutdownAtMu.Unlock()
-						delay = true
-					} else {
-						app.logger.Debug("shutdown delay configured but listener will remain open", zap.String("address", addr.String()))
-					}
-				}
+			if currentRefCount == 0 {
```
IndeedNotJames, line 510: The previous `shutdown_delay` used to be scoped to each 
`app.Servers`'s `addr.Network, addr.JoinHostPort(0)` combination (`lnKey`) in the `listenerPool` `UsagePool`.

Meaning, previously, a config with multiple `apps.http.servers` would always engage `shutdown_delay`, if a single of those got removed (or a new one added).

This no longer happens. It's global now, only triggering `if currentRefCount == 0 {`.

Not sure which version makes more sense.

---

I dug a bit more around and found https://github.com/caddyserver/caddy/pull/4906#issuecomment-1201841104 from the original PR

> [...] Now shutdowns are announced per-server and whether it's a process exit or a reload is irrelevant.

Following a commit (and refactor) explicitly making it per-server /shrug

francislavoie: Right... this is technically a behaviour change from the original intent, I think :frowning_face: 

@mholt so you see value in retaining that per-address shutdown delay (i.e. delay reloads if a server is actually going away)? I think this is extremely rare to happen in practice, how often do people remove a listener and keep the other? And what are the chances they care about delaying in that case?

I feel like the shutdown delay only really has value if Caddy is actually shutting down i.e. for a re-deployment or process restart so that some fronting load-balancer can notice before it goes down.

mholt: It's common. I think we need to keep ListenerUsage() accurate. I'll try to look into this soon and see if I can figure out a fix that does this.

mholt:
Thanks for working on a patch for this.

The behavior regarding enforcing the shutdown delay only when a listener is closed is intentional and really important.

I've been thinking of another way to do this, and I have a potential patch, but I can't test it yet as I'm [trying to get `curl` to work](https://twitter.com/mholt6/status/1631438882951806977) (thanks for nothing, Linux).

francislavoie:
Closing in favour of #5413

mholt:
Thank you for the valiant attempt though -- much respect!

PR #5413, opened 2023-03-03
core: Eliminate unnecessary shutdown delay on Unix

### LINKED ISSUE #5393: Slow config reload when using shutdown_delay

mholt:
Fix #5393, alternate to #5405.

I haven't had a chance to thoroughly test this yet because [my `curl` installation is currently borked](https://twitter.com/mholt6/status/1631438882951806977).

I still need to clean it up a bit and add explanatory comments if this works, so don't merge it yet.

/cc @francislavoie

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/caddyserver/caddy/v2
 
-go 1.18
+go 1.19
 
 require (
 	github.com/BurntSushi/toml v1.2.1
--- a/listen.go
+++ b/listen.go
@@ -12,10 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO: Go 1.19 introduced the "unix" build tag. We have to support Go 1.18 until Go 1.20 is released.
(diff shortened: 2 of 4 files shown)


francislavoie (approved):
Tested it with `--watch` and editing a Caddyfile, it works! :+1:

mholt:
Oops. I guess `_unix` isn't a thing in Go filenames, as those only work with literal GOOS values, and `unix` is a shortcut that is allowed only for actual build tags. Oh well. Who needs Windows anyway :stuck_out_tongue_winking_eye:

mholt:
Thank you for verifying it works, Francis! Really appreciate all you do.

Issue #5416, opened 2023-03-03
Should Caddy be adding /'s to fs.FS paths?


esell:
While recently playing around with a Caddy module that operates in the `caddy.fs` namespace (https://github.com/sagikazarmark/caddy-fs-s3), I ran into an issue where sub-directory listings were not working. The details are in this issue: https://github.com/sagikazarmark/caddy-fs-s3/issues/26, but the short version is that Caddy was tacking on a `/` to the directory name, so when `fs.ValidPath()` was called it was failing as the documentation states "Paths must not start or end with a slash: “/x” and “x/” are invalid" (https://pkg.go.dev/io/fs@master#ValidPath). This doesn't happen at the root directory as that is passed as `.`.

There are a couple of work-arounds for this, but I was wondering: is this expected behavior from Caddy? If so, should it be changed/updated? My guess is if more folks start building out `caddy.fs` modules and use `fs.ValidPath()` they will run into the same problem.

mholt:
@esell Great question!

When I wrote the [caddy-embed plugin](https://github.com/mholt/caddy-embed), which embeds your site into the binary and serves it with fs.FS, I encountered a similar issue:

https://github.com/mholt/caddy-embed/blob/6bbec9d65515ffdb40060d84255826f4fef6126c/module.go#L93-L97

Now that the file-server is using fs.FS for all file system access (including the default of the local disk, as we have a simple wrapper to make it an fs.FS:

https://github.com/caddyserver/caddy/blob/99d47050e97a8ccac2aad3bda46be46d4fec85ed/modules/caddyhttp/fileserver/staticfiles.go#L181-L183

https://github.com/caddyserver/caddy/blob/99d47050e97a8ccac2aad3bda46be46d4fec85ed/modules/caddyhttp/fileserver/staticfiles.go#L638-L651

), perhaps we should be doing something about the leading and trailing slashes in the file server itself.

The slashes obviously need to remain externally for the user-facing URLs, and we canonicalize the URLs (by default) for example redirecting to add a trailing slash for directory paths, or removing it for file paths; but it sounds like that when actually _using_ the fs.FS interfaces, we need to ensure those trailing slashes are always trimmed.

Sounds like a simple fix, if done carefully and in the right place(s). I hope just one place would be necessary.

Want to take a stab at a PR? :smiley:

esell:
[...]

PR #5417, opened 2023-03-03
remove trailing slash on fs filenames

esell:
Minor change to fix https://github.com/caddyserver/caddy/issues/5416

Tested with `osFS` and [caddy-fs-s3](https://github.com/sagikazarmark/caddy-fs-s3), and confirmed that path canonicalization redirects still happen successfully.


mholt (approved):
Beautiful. Thanks for the patch and the testing!

francislavoie:
Did you test that `browse` still works correctly? There's a bunch of other code following that line that uses the same variable, and I'm worried any of those other places could have unexpected breakages.

esell:
The `browse` scenario was my specific use-case. You can check it out here: https://zines.esheavyindustries.com/ I am using a locally built version of Caddy with this code change to serve it up.

PR #4855, opened 2022-06-28
reverseproxy: PROXY protocol support for http transport

WeidiDeng:
Add proxy protocol support for reverseproxy http transport

fix [2724](https://github.com/caddyserver/caddy/issues/2724).

PR #5424, opened 2023-03-08
proxyprotocol: Add PROXY protocol support to `reverse_proxy`, add HTTP listener wrapper

gucki:
This is an updated (rebased to the current master and improved) version of https://github.com/caddyserver/caddy/pull/4855. 

Please let me know if any changes are needed. Would be great to have this included in the next official release :-)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/klauspost/compress v1.16.0
 	github.com/klauspost/cpuid/v2 v2.2.4
+	github.com/mastercactapus/proxyprotocol v0.0.3
 	github.com/mholt/acmez v1.1.0
 	github.com/prometheus/client_golang v1.14.0
 	github.com/quic-go/quic-go v0.33.0
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/klauspost/compress v1.16.0
 	github.com/klauspost/cpuid/v2 v2.2.4
-	github.com/mastercactapus/proxyprotocol v0.0.3
+	github.com/mastercactapus/proxyprotocol v0.0.4
 	github.com/mholt/acmez v1.1.0
 	github.com/prometheus/client_golang v1.14.0
 	github.com/quic-go/quic-go v0.33.0
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -259,3 +259,22 @@ var hosts = caddy.NewUsagePool()
 // dialInfoVarKey is the key used for the variable that holds
 // the dial info for the upstream connection.
 const dialInfoVarKey = "reverse_proxy.dial_info"
+
+// ProxyProtocolInfo contains information needed to write proxy protocol to a
+// connection to an upstream host.
+type ProxyProtocolInfo struct {
+	// assume network is tcp because golang http transport can only support tcp now
+	IP   net.IP
+	Port int
+}
+
+// GetProxyProtocolInfo gets the proxy protocol info out of the context,
+// and returns true if there was a valid value; false otherwise.
+func GetProxyProtocolInfo(ctx context.Context) (ProxyProtocolInfo, bool) {
+	proxyProtocolInfo, ok := caddyhttp.GetVar(ctx, proxyProtocolInfoVarKey).(ProxyProtocolInfo)
+	return proxyProtocolInfo, ok
+}
+
+// proxyProtocolInfoVarKey is the key used for the variable that holds
+// the proxy protocol info for the upstream connection.
+const proxyProtocolInfoVarKey = "reverse_proxy.proxy_protocol_info"
--- a/modules/caddyhttp/reverseproxy/hosts.go
+++ b/modules/caddyhttp/reverseproxy/hosts.go
@@ -260,6 +260,10 @@ var hosts = caddy.NewUsagePool()
 // the dial info for the upstream connection.
 const dialInfoVarKey = "reverse_proxy.dial_info"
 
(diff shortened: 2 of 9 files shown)


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/klauspost/compress v1.15.15
 	github.com/klauspost/cpuid/v2 v2.2.4
+	github.com/mastercactapus/proxyprotocol v0.0.3
```
francislavoie, line 17: They makes me think since we're pulling in this dependency, we should probably pull in https://github.com/mastercactapus/caddy2-proxyprotocol as a standard module as well (the listener wrapper).

gucki: Sure, should I add it to modules/standard/imports.go?

mholt: Standard modules need to be brought into this repo. We can ask @mastercactapus if he'd be alright with that.

mholt:
Thank you for working on this! :raised_hands:

go.mod
mastercactapus: @mholt I'd be open to that

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -569,6 +570,28 @@ func (h *Handler) proxyLoopIteration(r *http.Request, origReq *http.Request, w h
 	// or satisfactorily represented in a URL
 	caddyhttp.SetVar(r.Context(), dialInfoVarKey, dialInfo)
 
+	var remoteAddr string
+	if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
```
WeidiDeng, line 574: I think we should wait for [pr](https://github.com/caddyserver/caddy/pull/5104) to be merged. Parsing xff is not good.

francislavoie: Good point. I'll keep that in mind as I finish that PR.

modules/caddyhttp/reverseproxy/hosts.go
```diff
@@ -259,3 +259,15 @@ var hosts = caddy.NewUsagePool()
...
 // the dial info for the upstream connection.
 const dialInfoVarKey = "reverse_proxy.dial_info"
+
+// proxyProtocolInfoVarKey is the key used for the variable that holds
+// the proxy protocol info for the upstream connection.
+const proxyProtocolInfoVarKey = "reverse_proxy.proxy_protocol_info"
+
+// ProxyProtocolInfo contains information needed to write proxy protocol to a
+// connection to an upstream host.
+type ProxyProtocolInfo struct {
```
WeidiDeng, line 269: We should change the type to [netip.AddrPort](https://pkg.go.dev/net/netip#AddrPort)

gucki: @WeidiDeng If we do so we'll have to convert it back at https://github.com/caddyserver/caddy/pull/5424/files#diff-8839c76ab9a0e424d77cabc5357fbcdfb166d6b6cc7a076114fb24a5b16814b8R223 anyway because the proxyprotocol.Header needs the current types. Also, the conversion would have to be done in some hacky way, imo (didn't see any native method to convert between the types). So is it really worth the effort?

modules/caddyhttp/reverseproxy/reverseproxy.go
gucki: @francislavoie Please let me know once merged and I'll rebase/ adapt.

modules/caddyhttp/reverseproxy/hosts.go
WeidiDeng: [AsSlice](https://pkg.go.dev/net/netip#Addr.AsSlice) and [Port](https://pkg.go.dev/net/netip#AddrPort.Port) are what you need.

gucki: @WeidiDeng Thanks, done now.

go.mod
francislavoie: I've gone ahead and pulled in the listener wrapper code into Caddy. Thanks so much @mastercactapus for having maintained it for all these years :blush: 

P.S. Maybe `github.com/mastercactapus/proxyprotocol` should transition to using `netaddr.IP`? It would be a breaking change but it might be slightly more efficient.

francislavoie:
Thanks for the work thus far @WeidiDeng @gucki, I'll finish this up once I'm done with the `client_ip` stuff, but this is looking good :ok_hand:

modules/caddyhttp/reverseproxy/reverseproxy.go
francislavoie: Alright, rebased & adjusted. @gucki @WeidiDeng would you mind confirming that it still works as expected?

WeidiDeng: It still works according to my test.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -688,8 +688,20 @@ func (h Handler) prepareRequest(req *http.Request, repl *caddy.Replacer) (*http.
 		req.Header.Set("Upgrade", reqUpType)
 	}
 
+	// Set up the PROXY protocol info
+	// TODO: We should probably migrate away from netip, but due to
```
mholt, line 692: Why migrate away from it?

mholt (approved):
This looks great, thank you for the contribution! Let's give this a try.

modules/caddyhttp/reverseproxy/reverseproxy.go
francislavoie: Oops, that was meant to say `net.IP`. But also this comment is sorta in the wrong place. The PROXY protocol lib is using the older `net.IP` interfaces instead of the newer `netip.Addr`, that's what this was meant to point out.

PR #5427, opened 2023-03-09
fileserver: New file browse template

mholt:
The current browse template is... fine. It's definitely far better than other servers' built-in file listings. But it's showing its age and we can do better:

- The font, Verdana, though tried-and-true, does not render well/smoothly on many modern devices. Modern fonts such as Inter and system UI fonts display better and are more legible.
- Accessibility concerns: there is not enough padding/whitespace, making it difficult to tap the right file on mobile screens if you have fat fingers or a small screen (this is something that watching my dad use a phone has made me very aware of).
- It's kind of stark. We do want a plain look, but it should still look visually interesting.
- Inconsistent iconography. The prior use of SVG was a bit immature. Now we use [Tabler icons](https://tabler-icons.io/) which are consistent, simple, and easy-to-use -- plus there's thousands of them to choose from, so we have a lot of flexibility that way.
- Dark theme improvements.
- On wide monitors (mine is 49"), the file listing is way too wide, making it hard to connect the filename to the info on the right side. It just looks a bit ridiculous this wide. The new theme has a max width.
- I removed borders between each row of the file listing, as the color change on hover does a better job at helping you see a row. Another option I experimented with is alternating row colors every 3n, 3n+1, and 3n+2 rows, but that felt more crowded/busy.
- Visual representation of relative file size in list view.
- Filetype-specific icons for common extensions. (We can add more too.)
- Entries with an image extension now show the image file instead of a generic icon.

Screenshot 2023-03-09 at 11-08-45 caddyhttp

The Firefox screenshot tool has a bug that causes the background to be incorrectly repeated here, but I promise that the background is actually fixed as you scroll:
Screenshot 2023-03-09 at 11-10-20 caddyhttp

Grid view:

Screenshot 2023-03-09 at 11-28-16 _

Screenshot 2023-03-09 at 11-28-37 _


modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -82,6 +82,16 @@ func (fsrv *FileServer) directoryListing(ctx context.Context, entries []fs.DirEn
...
 
+		// get extension(s)
+		var ext, fullExt string
+		firstDot, lastDot := strings.Index(name, "."), strings.LastIndex(name, ".")
+		if firstDot >= 0 {
+			fullExt = name[firstDot:]
+		}
+		if lastDot >= 0 {
+			ext = name[lastDot:]
+		}
```
mohammed90, line 93: Why not?

```suggestion
		// get extension(s)
		ext := filepath.Ext(name)
```

mholt: Because `.gz` and `.tar.gz` are different, and if we want to show a different icon for those, we need to know what the "full" extension is. (We don't currently do this, but I want to as soon as an icon becomes available for that.)

mohammed90: Gotcha. Fair enough. I can see this being problematic if some file has the name `agreement.document.v1.docx`, but it isn't a breaker.

mholt: Ok, now much better :)

mholt:
I say let's try this. I haven't tested this on a directory with thousands and thousands of entries -- I imagine that'll be fun for a browser to render -- but we can work on optimizations if we get complaints of slow load times.

abiosoft (approved):

modules/caddyhttp/fileserver/browse.html
```diff
@@ -47,142 +172,260 @@
...
-tr {
-	border-bottom: 1px dashed #dadada;
+/*
+tr:nth-child(3n+1) {
+	background-color: #f9f9f9;
+}
+
+tr:nth-child(3n+2) {
+	background-color: #f1f1f1;
+} */
```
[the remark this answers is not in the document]
mholt: Yeah, I guess so :D

francislavoie (approved):

mholt:
True, maybe loading images in the list view should be opt-in or something.

Tropaion:
Didn't even know I can do this with caddy? Is there a guide for this?
Also is it possible to edit caddyfiles with it in the browser?

francislavoie:
@Tropaion use the latest beta https://github.com/caddyserver/caddy/releases/tag/v2.7.0-beta.1

Tropaion:
@francislavoie Ok, thanks for the information.

mholt:
Yeah... I think that was a good idea. Thanks for trying the beta @xnaas!

revilowaldow:
Any idea when this is likely to make it into the main release? 
Looks excellent! Thanks for working on this, will be really helpful for my use case, I was just starting to look into how to make an image gallery template.

revilowaldow:
Well this should be an easy fix, thank you! 
image

PR #5428, opened 2023-03-10
Qlogs from environment variable 

FelixGaudin:
Allow collecting qlog files from HTTP/3 connections from Quic-Go using an environment variable


mholt:
Thanks for the PR! However, in this project, we don't use environment variables for configuration. Instead, that should be exposed in Caddy's [JSON config structure](https://caddyserver.com/docs/json/). Also, we should try to not create new exported API surface as much as possible (`NewBufferedWriteCloser`).

mholt:
Closing due to inactivity (and belated merge conflict) but will be happy to reopen and continue discussing :+1:

Issue #5420, opened 2023-03-06
v2.6.4 php-fpm gets stuck in read state


Jimbolino:
Since we upgraded from [v2.6.3 to v2.6.4](https://github.com/caddyserver/caddy/compare/v2.6.3...v2.6.4), we are experiencing some problems with our php-fpm backend

The problem is that the php-fpm process gets stuck in a "read" state (and not go back to "accept" state). 
Eventually php's pm.max_children is reached and the entire server will be killed by ec2's health checker.

I've done some debugging with strace, but have not found a pattern, because the servers gets lots of traffic, but the problem only occurs every once in a while. (maybe once in a 1000 requests)

```
root@frontend:~# strace -p 39614
strace: Process 39614 attached
read(5, 

root@frontend:~# nsenter -t 39614 --net lsof -n -p 39614 | grep 5u
php-fpm 39614 www-data    5u  IPv6 287757      0t0     TCP 172.18.0.3:9000->172.18.0.4:44492 (ESTABLISHED)

172.18.0.3 = app_php_1.app_default
172.18.0.4 = app_caddy_1.app_default
```

Im afraid that #5367 might need to be revisited. 
I've reverted back to v2.6.3 and the problem is gone.

mholt:
Thanks for the report. What is your config?

Jimbolino:
nope, v2.6.2 ran on production just fine

Im having trouble reproducing it locally (or on our staging environment)

Maybe it's got something to do with the AWS Network Load Balancer, or maybe bad bots that send malformed headers (wrong Content-Length etc)

francislavoie:
v2.6.4 should be functionally equivalent to v2.6.2 in terms of the proxy buffering functionality. I don't understand what's going on here.

mholt:
Can you try enabling access logs and maybe see if we can pin down which requests exactly are hanging?

Jimbolino:
We have cloudwatch enabled for our access logs
I think the default timeout is 3 seconds, because we see the following entries when searching for `{ $.duration > 1 }`
```
{
    "level": "error",
    "ts": 1678042875.2110817,
    "logger": "http.log.error",
    "msg": "dialing backend: dial tcp 172.18.0.3:9000: i/o timeout",
    "request": {
        "remote_ip": "172.31.10.238",
        "remote_port": "26109",
        "proto": "HTTP/1.1",
        "method": "GET",
        "host": "172.31.10.238:443",
        "uri": "/health.php",
[...]

PR #5435, opened 2023-03-15
reverseproxy: Reset Content-Length to prevent fastcgi from hanging

trea:
…PHP-FPM requests from hanging

See: https://github.com/caddyserver/caddy/issues/5420

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -646,7 +646,8 @@ func (h Handler) prepareRequest(req *http.Request, repl *caddy.Replacer) (*http.
 	// feature if absolutely required, if read timeouts are
 	// set, and if body size is limited
 	if h.RequestBuffers != 0 && req.Body != nil {
-		req.Body, _ = h.bufferedBody(req.Body, h.RequestBuffers)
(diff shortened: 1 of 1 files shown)


mholt:
Awesome @trea -- thank you for the PR!

Since this patch essentially reverts a reversion, can you help us verify that all the following issues are not a problem with this patch:

- https://github.com/caddyserver/caddy/issues/5236
- https://github.com/caddyserver/caddy/issues/5366
- https://github.com/caddyserver/caddy/issues/5420

I have my hands full today so it might be fastest if you are able to confirm these -- but anyone else is welcome to test as well!

trea:
Sure thing!

#5236 and #5420 should be fixed by this
#5366 should remain fixed by this because this patch still requires buffering to be enabled

But I do have more related thoughts I'm about to post in #5420. The TL;DR being that I think Caddy may need to implement buffering in its FastCGI transport for reverse proxying.

mholt (approved):
(Left a comment in the issue as well.)

I think this is starting to converge on the right solution -- and is a step forward, now that I fixed the OOM bug that I introduced :sweat_smile: 

I like your idea that probably the FastCGI transport should buffer anyway, so it doesn't have to be explicitly enabled, but this is at least a step back in the right direction.

PR #5437, opened 2023-03-16
cmd: Support ' quotes in envfile parsing

mholt:
Per the request later in #4686

@coolaj86 Could you please test this with your actual use case? I haven't had a chance yet.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -374,18 +374,19 @@ func parseEnvFile(envInput io.Reader) (map[string]string, error) {
 		}
 
 		// quoted value: support newlines
-		if strings.HasPrefix(val, `"`) {
-			for !(strings.HasSuffix(line, `"`) && !strings.HasSuffix(line, `\"`)) {
-				val = strings.ReplaceAll(val, `\"`, `"`)
+		if strings.HasPrefix(val, `"`) || strings.HasPrefix(val, "'") {
+			quote := string(val[0])
+			for !(strings.HasSuffix(line, quote) && !strings.HasSuffix(line, `\`+quote)) {
+				val = strings.ReplaceAll(val, `\`+quote, quote)
 				if !scanner.Scan() {
 					break
 				}
 				lineNumber++
-				line = strings.ReplaceAll(scanner.Text(), `\"`, `"`)
+				line = strings.ReplaceAll(scanner.Text(), `\`+quote, quote)
 				val += "\n" + line
 			}
-			val = strings.TrimPrefix(val, `"`)
-			val = strings.TrimSuffix(val, `"`)
+			val = strings.TrimPrefix(val, quote)
+			val = strings.TrimSuffix(val, quote)
 		}
 
 		envMap[key] = val


francislavoie:
AJ's quote example showed that `'` single quoted strings did not support variable interpolation (replacer) run on it, and that apparently single quotes can't be escaped with `\` (I don't understand the motivation there but :man_shrugging:)

coolaj86:
@mholt I'm not using variable expansion or escapes, so yes, that change will work for me.

@francislavoie `env` means "environment", as in "posix environment variable".

Ruby on Rails popularized it by having the ability to directly source the .env in the application rather than having to run two commands:

```sh
source .env
rails xxxx
```

became

```sh
rails xxxx
```

And then it started being used for ruby deployment services, like Heroku. And then people forgot that it's literally a shell language subset and starting doing crazy things - rather than implementing it the way ruby implemented it, which followed the pre-existing convention.

You can try this out in any Posix-y shell - bash, dash, sh, etc

```sh
FOO1="hello $empty"
FOO2="hello ${empty}"
export FOO3="he said \"hello\""
```

```sh
FOO4='hello $empty'
FOO5='hello ${empty}'
export FOO6='he said \'hello\''
```

```sh
echo $FOO1
echo $FOO2
echo $FOO3
echo $FOO4
echo $FOO5
echo $FOO6
```

Traditionally `source .env` before running a program, which is why they often have the `export ` prefix - so that the shell knows to make the variable available to child processes.

Most ENV parsers for web apps will ignore that, but some do not.

Many people would no doubt like `'hello \' world'` to produce `hello ' world`. That's incorrect, but as long as the implementation doesn't break `'hello '\'' world'` from working, no one would probably ever notice.

francislavoie:
Thanks, but I'm well aware of `dotenv` and its history :+1: 

With the current implementation, with a Caddyfile like this

```
:8881 {
    respond <<EOF
    FOO={env.FOO}
    BAR={env.BAR}
    FOOBAR={env.FOOBAR}
    BAZBAR={env.BAZBAR}
    QUX={env.QUX}
    EOF
}
```

This env file:

```
export FOO="export must be ignored for webapps (has special meaning in shell envs)"
FOO="bar"
BAR="doubles can have string interpolation ${FOO} in either posix form $FOO"
FOOBAR='you can'\''t escape single quotes with a backslash inside of a single quote'
BAZBAR="but you can escape \"double quotes\" within double quotes"
QUX='either
form
can
have
newlines'
```

Outputs this:

```
FOO=bar
BAR=doubles can have string interpolation ${FOO} in either posix form $FOO
FOOBAR=you can'\''t escape single quotes with a backslash inside of a single quote
BAZBAR=but you can escape \"double quotes\" within double quotes
QUX=either
form
can
have
newlines  
```

So basically, `FOOBAR` and `BAZBAR` seem incorrect according to "spec", and we don't currently do `$FOO` or `${FOO}` interpolation, and it seems like `{env.FOO}` interpolation isn't working either (I'm not sure it should though, probably not frankly).

coolaj86:
Here are my thoughts:

- Better to do nothing rather to do something in a way that fixing it later would break something that works
- Better to follow the ruby / heroku / dotenv.org way than the posix way

@dotenv-org is taking over various dotenv libraries and bringing them into conformity (node's dotenv finally allows multi-line vars).

However, I'm confused as to how to interpret https://github.com/motdotla/dotenv#what-rules-does-the-parsing-engine-follow - it seems too loose.

In any case: I retract my position of supporting the Posix standard because whether hijacked or not, they're winning.

coolaj86:
Example of where single quotes are absolutely necessary:

```sh
# The shell will interpret this
CADDY_BCRYPT="$xx$xx$xxxxxxxxxxxxxxxxxxxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"

# The shell will NOT interpret this
CADDY_BCRYPT='$xx$xx$xxxxxxxxxxxxxxxxxxxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
```

In order to do a check to see if an env file has been configured, or to use the configuration, single quotes must be used.

coolaj86:
Worked!

PR #5438, opened 2023-03-16
rewrite: Add option to force modifying the query

### LINKED ISSUE #5208: X-Accel-Redirect setup doesn't work due to strange behaior of rewrite
Hey ! I'm trying to configure a pretty simple X-Accel-Redirect kind of setup under **Caddy 2.6.2**, but think I encountered an issue that prevents me to do so.

My stripped down Caddyfile looks like this:

```
localhost {
    reverse_proxy service_a:5000 {
        @accel header X-Accel-Redirect *
        handle_response @accel {
            rewrite * {rp.header.X-Accel-Redirect}
            reverse_proxy service_b:5000
        }
    }
}
```

The `service_a` returns a response with a header `X-Accel-Redirect: /hello?some=param&some_other=param`. This is be picked by Caddy and rewritten so that this query is handled by `service_b`. The rewrite of the path works, but the GET params get swallowed and are not visible to `service_b` where I only see `/hello`.

It works as expected when I hardcode the same value in the caddy file like this (which IMO should be 100% equivalent).

```
localhost {
    reverse_proxy service_a:5000 {
        @accel header X-Accel-Redirect *
        handle_response @accel {
            rewrite * /hello?some=param&some_other=param
            reverse_proxy service_b:5000
        }
    }
}
```

Just to understand what's happening, I tried with some variants such as `rewrite * /hello?{rp.header.X-Accel-Redirect}` with just the paramters in the header. In this case, `service_b` sees `http://localhost/hello?some%3Dparam%26some_other%3Dparam`, so not good either.

Seems some rogue urlencoding happens, but I couldn't find anything in the doc about that.

Any pointer about how to solve this ? I searched through the issues but could not find anything about this exactly.

francislavoie:
Fix #5208

When a user wants to rewrite the URI, if they use a placeholder which might contain both the path and query, currently only the path portion of the placeholder will be used and the query is discarded.

This isn't ideal when the placeholder input comes from, for example, a response header from upstream when doing `X-Accel-Redirect` style intercepting of the response.

To work around this, we can add an option to force-enable query modifications, essentially marking the configured placeholder input as "trusted" in the sense that it's expected to contain a valid query part and not an injected `?` via URL encoding.

I'm not sure the implementation is completely correct. There's a test case I'm not quite sure how we want to handle, i.e. the placeholder only having a query and no path. Is that something we care to support? If not I can remove that `TODO` comment.

### DIFF
--- a/modules/caddyhttp/rewrite/rewrite.go
+++ b/modules/caddyhttp/rewrite/rewrite.go
@@ -92,6 +92,17 @@ type Rewrite struct {
 	// Mutates the query string of the URI.
 	Query *queryOps `json:"query,omitempty"`
 
+	// If true, the rewrite will be forced to also apply to the
+	// query part of the URL. This is only needed if the configured
+	// URI does not include a '?' character which is normally used
+	// to determine whether the query should be modified. In other
+	// words, this allows rewriting both the path and query when
+	// using a placeholder as the replacement value, whereas otherwise
+	// only the path would be rewritten because the placeholder itself
+	// does not contain a '?' character. Only use this if the placeholder
+	// is trusted to not be vulnerable to query injections.
+	ForceModifyQuery bool `json:"force_modify_query,omitempty"`
+
 	logger *zap.Logger
 }
 
@@ -222,7 +233,8 @@ func (rewr Rewrite) Rewrite(r *http.Request, repl *caddy.Replacer) bool {
 			var injectedQuery string
 			newPath, injectedQuery = before, after
 			// don't overwrite explicitly-configured query string
-			if query == "" {
+			// unless configured explicitly to do so
+			if query == "" || rewr.ForceModifyQuery {
 				// the injected query came from the first-pass placeholder
 				// expansion above, which means any '{' or '}' bytes in it
 				// must have come from replacement values (e.g. a request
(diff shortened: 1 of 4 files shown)


mholt:
Thanks for proposing this Francis, I'll try to get around to reviewing this soon. (Sorry for the delay. I'm also curious about this versus #5504)

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -88,6 +88,17 @@ type Rewrite struct {
...
+	// If true, the rewrite will be forced to also apply to the
+	// query part of the URL. This is only needed if the configured
+	// URI does not include a '?' character which is normally used
+	// to determine whether the query should be modified. In other
+	// words, this allows rewriting both the path and query when
+	// using a placeholder as the replacement value, whereas otherwise
+	// only the path would be rewritten because the placeholder itself
+	// does not contain a '?' character. Only use this if the placeholder
+	// is trusted to not be vulnerable to query injections.
+	ModifyQuery bool `json:"modify_query,omitempty"`
```
mholt, line 100: What if we call this `SpanComponents` or something a little more precise? It looks like what this change really does is make a single placeholder able to span URI components.

mholt (commented):
This LGTM, but I have a nit regarding the naming :upside_down_face:

modules/caddyhttp/rewrite/rewrite.go
francislavoie: That makes me think of a `<span>` JS component. That name doesn't bring any mental association to what it does to me.

mholt: Hmm. I mean, obviously this is a backend context (a rewrite middleware) where we're talking about the URI... but I can see why you'd have that correlation.

What about `AllowExpansion` or something?

francislavoie: I decided to go with `force_modify_query`, adding the `force_` prefix, similar to other examples we've had like `force_automate`

PR #5439, opened 2023-03-17
H2 without tls conn

coolaj86:
Re: https://github.com/caddyserver/caddy/pull/4929

Just in case this saves you a little bit of work @WeidiDeng.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -356,6 +356,11 @@ func (app *App) Start() error {
 				return context.WithValue(ctx, ConnCtxKey, c)
 			},
 		}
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/app.go
```diff
@@ -382,9 +393,6 @@ func (app *App) Start() error {
 
 		// enable H2C if configured
 		if srv.protocol("h2c") {
-			h2server := &http2.Server{
```
coolaj86, line 385: This bit probably needs to be brought over.

mholt:
Closing, if there's still something needed here though maybe we can continue to discuss it.

PR #5441, opened 2023-03-17
log: make `sink` encodable

mohammed90:
`sink` logs have, thus far, been printed not encoded in any format. This does not play friendly with log parsers expecting a specific format from a particular source. In other words, it becomes a hassle to tell the log parser, e.g. Loki, Caddy logs are json-formated if some lines aren't.

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -268,31 +268,120 @@ type StandardLibLog struct {
 	// The module that writes out log entries for the sink.
 	WriterRaw json.RawMessage `json:"writer,omitempty" caddy:"namespace=caddy.logging.writers inline_key=output"`
 
-	writer io.WriteCloser
+	// The encoder is how the log entries are formatted or encoded.
+	EncoderRaw json.RawMessage `json:"encoder,omitempty" caddy:"namespace=caddy.logging.encoders inline_key=format"`
+
+	// Level is the minimum level to emit, and is inclusive.
+	// Possible levels: DEBUG, INFO, WARN, ERROR, PANIC, and FATAL
+	Level string `json:"level,omitempty"`
+
+	// Sampling configures log entry sampling. If enabled,
+	// only some log entries will be emitted. This is useful
+	// for improving performance on extremely high-pressure
+	// servers.
+	Sampling *LogSampling `json:"sampling,omitempty"`
+
+	writerOpener WriterOpener
+	writer       io.WriteCloser
+	encoder      zapcore.Encoder
+	levelEnabler zapcore.LevelEnabler
+	core         zapcore.Core
+	logger       *zap.Logger
 }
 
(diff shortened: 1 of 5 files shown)


mholt:
I wonder if we can embed one type within another (`StandardLibLog` or `CustomLog`) to reduce duplication. If StandardLibLog used CustomLog, I'd be pretty confident that the change is good :+1:

mohammed90:
> I wonder if we can embed one type within another (`StandardLibLog` or `CustomLog`) to reduce duplication. If StandardLibLog used CustomLog, I'd be pretty confident that the change is good 👍

Now that I've tried it, embedding introduces more invasive changes than the duplication. I even tried to have a third struct type for the common fields, but same limitation incurred, keyed fields had to be wrapped with the embedded type. The `CustomLogger` cannot be the one that's embedded because it contains `Include` and `Exclude` fields which don't operate outside of it and have no semantics for anything besides it.

In case you want to play with it, here's a working config:
```
{
	"logging": {
		"sink" : {
			"level": "DEBUG",
			"encoder": {
				"format": "json"
			}
		},
		"logs": {
			"default": {
				"level": "DEBUG",
				"encoder": {
					"format": "json"
				}
			}
		}
	},
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"match": [
								{
									"host": [
										"localhost"
									]
								}
							],
							"handle": [
								{
									"handler": "subroute",
									"routes": [
										{
											"handle": [
												{
													"body": "Howdy!",
													"handler": "static_response"
												}
											]
										}
									]
								}
							],
							"terminal": true
						}
					]
				}
			}
		}
	}
}
```

francislavoie:
I highly doubt any plugin is directly using the `StandardLibLog` type right now. Could we just rename it to `BaseLog` or something, and use that as `sink`? That would make it read less weird.

mohammed90:
> I highly doubt any plugin is directly using the `StandardLibLog` type right now. Could we just rename it to `BaseLog` or something, and use that as `sink`? That would make it read less weird.

How about now?

logging.go
```diff
@@ -259,55 +259,15 @@ func (wdest writerDestructor) Destruct() error {
 	return wdest.Close()
 }
 
-// StandardLibLog configures the default Go standard library
+// BaseLog configures the default Go standard library
 // global logger in the log package. This is necessary because
 // module dependencies which are not built specifically for
 // Caddy will use the standard logger. This is also known as
 // the "sink" logger.
```
francislavoie, line 266: This comment should be moved to `SinkLog` I think, and this should be changed to properly explain what `BaseLog` does.

mholt (approved):
Awesome -- yeah, let's give it a try!

PR #5443, opened 2023-03-17
caddytls: Zero out throttle window first

### LINKED ISSUE #5442: Removing on_demand rate_limit causes a configuration error
This is related to #5368 

To fix the vulnerability warning, I added the on_demand ask and rate_limit options to our JSON config. Since that is fixed, I am trying to remove them from the Caddy config by updating the config using the API endpoint. This causes an error though.

Previously, I would update the Caddy config with a `POST /load` that included the following config generated by Ruby. The relevant snippet that is causing the issue:

```ruby
        tls: {
          automation: {
            policies: @tls_policies,
            on_demand: {
              ask: Rails.application.routes.url_helpers.api_v1_caddy_ask_url,
              rate_limit: {
                interval: "2m",
                burst: 5
              }
            }
          },
```

Now we're sending over the following with `POST /load`

```ruby
        tls: {
          automation: {
            policies: @tls_policies
            }
          },
```

I've also tried setting it to an empty hash or null to clear out the on_demand param but same result with those.

```
on_demand: {} // null
```

Caddy's internal updating of the config seems to be tripping on this removal as the POST request fails and we get this error in the Caddy logs from certmagic.

```
invalid configuration: maxEvents = 0 and window != 0 would not allow any events
goroutine 116 [running]:
net/http.(*conn).serve.func1()
net/http/server.go:1854 +0xbf
panic({0x1af0120, 0x234d3f0})
runtime/panic.go:890 +0x263
github.com/caddyserver/certmagic.(*RingBufferRateLimiter).SetMaxEvents(0xc0002dc780, 0x0)
github.com/caddyserver/certmagic@v0.17.2/ratelimiter.go:154 +0x218
github.com/caddyserver/caddy/v2/modules/caddytls.(*TLS).Provision(0xc0007d91e0, {{0x236a758, 0xc0004e41e0}, 0xc0005d7bf0, 0xc0004e4140, {0x0, 0x0, 0x0}, {0xc000646e60, 0x2, ...}})
```

This seems like a bug where Caddy's retaining that rate limit option internally when it should have been removed?

mholt:
Otherwise it's a panic when setting the MaxEvents to 0 while the window is still non-zero. Should fix #5442

### DIFF
--- a/modules/caddytls/automation.go
+++ b/modules/caddytls/automation.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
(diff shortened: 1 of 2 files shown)


francislavoie (approved):
This looks good to me.

I added another commit to fix a regression from https://github.com/caddyserver/caddy/commit/b97c76fb4789b8da0b80f5a2c1c1c5bebba163b5 which was mentioned in the issue this PR is meant to solve.

excid3:
I compiled 671061d and that seems to fix it. 👍

mholt:
And thank you Francis for your patch as well :+1:

PR #5446, opened 2023-03-19
reverseproxy: Fix active health check header canonicalization, refactor

### LINKED ISSUE #5444: Host in health_checks should not require capitalization
Consider this `reverse_proxy` handler with `health_checks`. It's not forwarding the right host header and therefore the check is constantly failing.

```yaml
- handler: reverse_proxy
  health_checks:
    active:
      uri: /health
      interval: 5s
      expect_body: "Healthy"
      headers:
        host: [ "something.test.com" ]
        blaat: [ "blaat2" ]
  upstreams:
    - dial: "localhost:9003"
```

After debugging I found that [this piece of code is responsible](https://github.com/caddyserver/caddy/blob/a7db0cfe55b524a34a310e852af6c9520134db10/modules/caddyhttp/reverseproxy/healthchecks.go#L284). Work arround is to supply the `host` header key in [it's canonical form](https://golangbyexample.com/canonical-http-header-key/). (so `Host` instead of `host`).

The [header Get function](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/net/http/header.go;l=49) actually describes nicely what happens:
``` 
Get assumes that all keys are stored in canonical form. To use non-canonical keys, access the map directly.
```

It's a small issue, which direction would be appreciated if I would create a PR (there is some opinion given #4330, but I find it hard to apply here)?
1. Just access the map directly and let the behaviour as much as it is (other headers are also just forwarded as is).
2. Have a 'better' parser that immediately converts the headers into their canonical form. Then the header.Get function can just be used as is (and the `strings.ToLower(key) == "host"` could also be simplified to something without the ToLower). The downside of this is that caddy would become more opinionated on the header casing (in theory not a problem since HTTP requires you to not depend on casing). 

I would like to do this to prevent somebody else from having to spend 2 hours finding out why the health check is not working. ⌛

francislavoie:
Fix #5444

### DIFF
--- a/modules/caddyhttp/reverseproxy/healthchecks.go
+++ b/modules/caddyhttp/reverseproxy/healthchecks.go
@@ -24,7 +24,6 @@ import (
 	"regexp"
 	"runtime/debug"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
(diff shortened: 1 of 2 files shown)


francislavoie:
The key here is this chunk of code:

```go
	// Canonicalize the header keys ahead of time, since
	// JSON unmarshaled headers may be incorrect
	cleaned := http.Header{}
	for key, hdrs := range a.Headers {
		for _, val := range hdrs {
			cleaned.Add(key, val)
		}
	}
	a.Headers = cleaned
```

I also moved the provisioning logic to better localize it in the healthchecks file, moving the provisioning closer to the struct.

mholt (approved):
I think this refactor is a good cleanup. Thank you for the patch!

PR #5458, opened 2023-03-25
replacer: Add HTTP time format

francislavoie:
This is the same as Go stdlib http.TimeFormat, but I didn't want to pull in a dependency to this file just for that.

Context: https://caddy.community/t/equivalent-to-nginx-date-gmt/19408, this would be helpful to fill some HTTP headers like `Last-Modified` that take this format.

### DIFF
--- a/replacer.go
+++ b/replacer.go
@@ -16,6 +16,7 @@ package caddy
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


mholt:
We already depend on `net/http` -- using `import` in this file doesn't add any bloat I'm 99.99% sure -- it just gives the compiler knowledge of what the `http` identifier is.

mholt (approved):
Yay :blush: Thanks

PR #5459, opened 2023-03-25
caddytls: Run the replacer on automation policy subjects

francislavoie:
Also renamed the field to SubjectsRaw, which can be considered a breaking change but I don't expect this to affect much.

This has come up semi-frequently. I think it's fine to support `{env.DOMAIN}` for subjects. I'm not 100% sure if this will work in all the cases we expect it to, but it should at least be better supported than before.

### DIFF
--- a/modules/caddytls/automation.go
+++ b/modules/caddytls/automation.go
@@ -85,7 +85,7 @@ type AutomationConfig struct {
 // TLS app to properly provision a new policy.
 type AutomationPolicy struct {
 	// Which subjects (hostnames or IP addresses) this policy applies to.
-	Subjects []string `json:"subjects,omitempty"`
+	SubjectsRaw []string `json:"subjects,omitempty"`
 
 	// The modules that may issue certificates. Default: internal if all
 	// subjects do not qualify for public certificates; othewise acme and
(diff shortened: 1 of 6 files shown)


francislavoie:
I confirmed this fixes the case of `{env.DOMAIN}.duckdns.org` with the DNS challenge where the env var is either a flat domain or a wildcard. It properly issues a cert for that domain, and I can make requests with `curl -v` afterwards.

modules/caddytls/automation.go
```diff
@@ -147,12 +147,21 @@ type AutomationPolicy struct {
...
 
 // Provision sets up ap and builds its underlying CertMagic config.
 func (ap *AutomationPolicy) Provision(tlsApp *TLS) error {
+	// replace placeholders in subjects to allow environment variables
+	repl := caddy.NewReplacer()
+	subjects := make([]string, len(ap.SubjectsRaw))
+	for i, sub := range ap.SubjectsRaw {
+		subjects[i] = repl.ReplaceAll(sub, "")
+	}
+	ap.subjects = subjects
```
mholt, line 163: Would it work if we just replace the subjects in-place? :thinking:

mholt (commented):
Woah, well thanks for being brave and doing this -- I do wonder if we could make this a little simpler though... (also the use of `*Raw` makes me think of a module field)

modules/caddytls/automation.go
francislavoie: No because then if you do `curl localhost:2019/config/` you get a config without the placeholders in it.

mholt (approved):
Well let's give it a shot - thanks!

PR #5460, opened 2023-03-26
reverseproxy: Header up/down support for CLI command

### LINKED ISSUE #5376: Support setting request headers for reverse-proxy command
The one-line reverse-proxy command could be even better if it supports basic authentication like this:

```
caddy reverse-proxy --from :10000 --to http://user:pass@:20000
```

francislavoie:
Closes #5376


mholt (approved):
Awesome! Thanks for implementing this. :smiley:

PR #5461, opened 2023-03-26
caddyhttp: Log request body bytes read

### LINKED ISSUE #4494: Feature request: Log request size
This may of course be a naive request, given streaming may make it more difficult than is obvious. I'd love to be able to log size (in bytes, like the response size) of incoming requests, and I don't see that in the access log JSON object.

e.g.
1) Client POSTS to /foo/bar with a few fields (say, 1600 bytes total).
2) "size": 1600 is included in the "request" key of the access log JSON object.

The motivation for this is to simplify debugging when I believe a client is sending a body that doesn't match its Content-Length header, so we can rule out our beloved reverse proxy causing the issue :)

Thanks!

francislavoie:
Closes #4494

I'm not super satisfied with this unfortunately... the problem is we can't really add this to the `request` log object, because it gets marshaled when we call `.With(loggableReq)` and doesn't wait until the `defer` so it logs the size as 0 because it hasn't been read yet. So we need to put it in the top-level of the log. Later on, maybe something like `.LazyWith()` might get added https://github.com/uber-go/zap/issues/1121 but for now I don't see a way to prevent it.

Because of that, I'm not sure what the best naming is. `req_size` could make sense because we already have `size` for the response size, but that's weirdly incongruent with `resp_header` using a prefix for the other half. `bytes_read` is more accurate in the sense that if the body isn't read then it would report 0 even if there was a body in the request, but it doesn't have `req_`; should it, or would that be too long?

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -19,6 +19,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/netip"
@@ -259,6 +260,14 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		wrec := NewResponseRecorder(w, nil, nil)
 		w = wrec
 
+		// wrap the request body in a LengthReader
+		// so we can track the number of bytes read from it
+		var bodyReader *lengthReader
+		if r.Body != nil {
+			bodyReader = &lengthReader{Source: r.Body}
+			r.Body = bodyReader
+		}
+
 		// capture the original version of the request
 		accLog := s.accessLogger.With(loggableReq)
 
@@ -285,7 +294,13 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 			userID, _ := repl.GetString("http.auth.user.id")
 
+			reqBodyLength := 0
+			if bodyReader != nil {
+				reqBodyLength = bodyReader.Length
+			}
+
 			log("handled request",
+				zap.Int("bytes_read", reqBodyLength),
 				zap.String("user_id", userID),
 				zap.Duration("duration", duration),
 				zap.Int("size", wrec.Size()),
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/server.go
```diff
@@ -248,6 +249,14 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		wrec := NewResponseRecorder(w, nil, nil)
 		w = wrec
 
+		// wrap the request body in a LengthReader
+		// so we can track the number of bytes read from it
+		var bodyReader *lengthReader
```
mholt, line 254: Might be a long shot, but does this have to be a pointer? Just wondering if we can avoid an allocation.

mholt (commented):
Thanks for tackling this -- I can see why it'd be useful to have this info.

> Because of that, I'm not sure what the best naming is. req_size could make sense because we already have size for the response size, but that's weirdly incongruent with resp_header using a prefix for the other half. bytes_read is more accurate in the sense that if the body isn't read then it would report 0 even if there was a body in the request, but it doesn't have req_; should it, or would that be too long?

Ugh, you're right, that's kind of a bother. I like `bytes_read` you have already, for the preciseness. If I had planned ahead better I would have designed the field names to be more congruent/symmetric. Oh well :(

modules/caddyhttp/server.go
francislavoie: I don't think so because `r.Body` takes a pointer

> cannot use bodyReader (variable of type lengthReader) as io.ReadCloser value in assignment: lengthReader does not implement io.ReadCloser (method Close has pointer receiver) compiler([InvalidIfaceAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidIfaceAssign))

francislavoie: Well maybe if I make Read and Close not use a pointer receiver, but does that even work?

mholt: Ah yeah, that was my concern. Makes sense. Oh well!

mholt (approved):

modules/caddyhttp/server.go
francislavoie: You were kinda right though - we should probably use `sync.Pool` for `lengthReader`.

PR #5462, opened 2023-03-27
httpcaddyfile: Adjust path matcher sorting to solve for specificity

francislavoie:
Our existing Caddyfile path matcher sorting has served us quite well, but I only just thought of one edgecase we didn't handle very well.

In the case that the user defines both `/foo` and `/foo*` matchers on the same directive type, previously it would sort `/foo*` first because it's a longer path matcher (more chars).

The problem is, technically `/foo` is more specific because `*` matches 0 or more characters. So we need to reverse the order of those when the paths are the same other than having a `*` suffix.

Also, I took the liberty of refactoring the implementation a bit to use an inline function to deduplicate the condition. That was my fault when I introduced the case for handling `vars` being reversed.


mholt (approved):
An elegant patch. Thank you for this!

Issue #5451, opened 2023-03-21
How to delete all headers to upstream in reverse_proxy


heimoshuiyu:
Hi, In the docs <https://caddyserver.com/docs/caddyfile/directives/reverse_proxy> it said to delete all matching request, using a suffix match `header_up -Some-*`. I want to delete all headers so here is my Caddyfile

```Caddyfile
:9000 {
    reverse_proxy * http://127.0.0.1:9001 {
        header_up -*
    }
}
```

Then I run `curl -v http://127.0.0.1:9000` in command line, the output is

```textile
*   Trying 127.0.0.1:9000...
* Connected to 127.0.0.1 (127.0.0.1) port 9000 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:9000
> User-Agent: curl/7.88.1
> Accept: */*
> 
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
```

heimoshuiyu:
After reading some code, I found that the correct way to write it should be `header_up -**`. I believe that the code did not handle the case where there is only one asterisk, and I will soon submit a small PR to fix this bug :)

francislavoie:
Why would you want to do this? :thinking: that would certainly break things, many HTTP headers are _required_ to exist for the connection to be properly maintained.

heimoshuiyu:
Because I don't want any extra headers sent by downstream to be received by upstream. I didn't realize that missing the required HTTP headers would break many programs... So, is there any way to include only specific headers to reach upstream?

francislavoie:
Headers are harmless, unless there's some that you're particularly concerned about. It would be best if you just remove the particular headers you're concerned about, instead of removing them all.

heimoshuiyu:
Yes, they are harmless. I want to remove headers out of privacy concerns. I'm worried that downstream HTTP programs or CDNs will add unknown headers, which may be used by upstream to analyze or track.

For example, Cloudflare adds headers like `cf-*` and `cdn-*`. Some browsers may add specific headers in certain cases. It's hard for me to find all these headers. In this case, upstream will know that I am using a CDN or a specific browser, which is something I want to avoid.

francislavoie:
Why is that a concern? A reverse proxy is not a forward proxy. If you don't trust the software you're proxying to, then why are you proxying to it?

heimoshuiyu:
[...]

PR #5464, opened 2023-03-27
Delete all headers to upstream

heimoshuiyu:
Related issue <https://github.com/caddyserver/caddy/issues/5451>

This PR recognizes the syntax `header_up -*` and removes all headers. After removing all headers, any desired headers can be added.

Here is an example Caddyfile:

```
:9000 {
    reverse_proxy * http://127.0.0.1:9001 {
        header_up -*
        header_up authorization {http.request.header.authorization}
    }
}
```

This example removes all headers sent by the client except for authorization.

As I am not familiar with the code of Caddy, I am concerned that changing the order of the header manipulation could affect other programs. Please help me review this PR :)


francislavoie:
Hmm, this'll also affect the `header` directive since the same code is used outside of the proxy. This would likely break a usecase like this:

```
header {
	-Foo
	FooBar {header.Foo}
}
```

Previously, it would take the value from `Foo` before deleting `Foo`, but now it will delete `Foo` first which would make the placeholder return an empty value (I didn't run it though so that might need confirming).

francislavoie:
Oh sorry, yeah. I meant the `request_header` directive, but that's a less common usecase. I wrote that too late last night 😶

heimoshuiyu:
> Oh sorry, yeah. I meant the `request_header` directive, but that's a less common usecase. I wrote that too late last night no_mouth

I tested the `request_header` directive and obtained the same results using both v2.6.4 and the version with this PR modification 🤔.

Caddyfile

```
:9000 {
    request_header -Foo
    request_header FooBar {header.Foo}
    reverse_proxy * http://127.0.0.1:9001
}
```

Command

```
nc -l 9001
curl http:127.0.0.1:9000/ -H 'Foo: value' -v
```

Result from `nc -l 9001`

```
HEAD / HTTP/1.1
Host: localhost:9000
User-Agent: curl/7.88.1
Accept: */*
Foobar: 
X-Forwarded-For: 127.0.0.1
X-Forwarded-Host: localhost:9000
X-Forwarded-Proto: http
```

By switching the order it works 🤔

```
:9000 {
    request_header FooBar {header.Foo}
    request_header -Foo
    reverse_proxy * http://127.0.0.1:9001
}
```

Result

```
HEAD / HTTP/1.1
Host: localhost:9000
User-Agent: curl/7.88.1
Accept: */*
Foobar: value
X-Forwarded-For: 127.0.0.1
X-Forwarded-Host: localhost:9000
X-Forwarded-Proto: http
```

It appears that the execution order of `request_header` is determined by the order in which it is wroted in the configuration file.

francislavoie:
>    request_header -Foo
>    request_header FooBar {header.Foo}

That's cause that's two separate handler invocations instead of one handler with two operations.

Did we not implement multiple operations for request headers in the Caddyfile? 🤷‍♂️

This might be fine.

modules/caddyhttp/headers/headers.go
```diff
@@ -192,6 +192,16 @@ type RespHeaderOps struct {
 
 // ApplyTo applies ops to hdr using repl.
 func (ops HeaderOps) ApplyTo(hdr http.Header, repl *caddy.Replacer) {
+	// delete all fields before manipulating HTTP headers
```
mholt, line 195:
```suggestion
	// before manipulating headers in other ways, check if there
	// is configuration to delete all headers, and do that first
	// because if a header is to be added, we don't want to delete
	// it also
```

modules/caddyhttp/headers/headers.go
```diff
@@ -192,6 +192,16 @@ type RespHeaderOps struct {
...
+	for _, fieldName := range ops.Delete {
+		fieldName = strings.ToLower(repl.ReplaceKnown(fieldName, ""))
```
mholt, line 197:
```suggestion
		fieldName = repl.ReplaceKnown(fieldName, "")
```

modules/caddyhttp/headers/headers.go
```diff
@@ -215,6 +225,9 @@ func (ops HeaderOps) ApplyTo(hdr http.Header, repl *caddy.Replacer) {
 	// delete
 	for _, fieldName := range ops.Delete {
 		fieldName = strings.ToLower(repl.ReplaceKnown(fieldName, ""))
+		if fieldName == "*" {
+			continue
```
mholt, line 229:
```suggestion
			continue // handled above
```

mholt (approved):
Thanks for the patch!

This LGTM -- I appreciate the careful consideration. Initially I too was thinking we just move the Delete logic up to be first, but I can see how this is probably safer and more relevant.

I just submitted suggestions to improve the comments. If you agree let's go ahead and accept those changes before merging. Thanks!

PR #5466, opened 2023-03-27
Update contributing guidelines

mholt:
I've seen some pretty bad code in my day, but some of the worst I've ever seen comes from current language models (LLMs, like ChatGPT).

Traditionally, bad code is mostly obvious.

With LLMs, it's often _subtle_.

I want to ~~establish~~ **clarify** a simple policy related to getting code from communities (like Stack Overflow) and language models (ChatGPT, GitHub Copilot, Sourcegraph Cody, etc.) for a few reasons:

- To reduce the likelihood of subtle bugs being introduced into our code base
- To try to eliminate worries of plagiarism and licensing, as tools like Copilot have been known to regurgitate incompatibly-licensed code and even [plagiarize nontrivial code](https://twitter.com/mitsuhiko/status/1410886329924194309)
- To make contributors and users aware of our high standards for original, human-thought code

I consider this to be a policy _clarification_ rather than a new policy because we've always required contributions to be CLA-compliant, well-tested, understood by the authors, etc. Now I'm just spelling out that code generated by AI tools (and Stack Overflow) needs to be extra-thoroughly tested and vetted, if used at all. (My recommendation would be to not copy-paste code, but maybe get inspiration from it if it's correct in the first place.)

To be clear, use of those sites and tools is allowed (we're not your mom or your teacher), but we do expect that authors will actually _author_ their code: write it, be responsible for it. It's already enough work trying to find human mistakes -- then to add in a computer's mistakes and it's a lot more work for us.

I also made a few other minor adjustments.

I'm open to feedback on this. This is not a heavy-handed policy on purpose (you dev how you want to dev), and is more like a strong recommendation to use discretion with these tools. I just want contributors to be aware of what kinds of subtle problems these tools can cause.


mholt:
I might also add a clause that asks for disclosure when generated (or copied) code is used verbatim in a contribution, as a courtesy for reviewers and maintainers, but also "for the record."

PR #5468, opened 2023-03-29
reverseproxy: Add `query` and `client_ip_hash` lb policies

### LINKED ISSUE #5469: Overwrite remote_ip with x-forwarded-for for ip_hash load-balancing
Hi,

I've been researching for a couple of hours and still didn't find out if Caddy supports this or not.

My Caddy is behind another Reverse-Proxy (because it can't do `ip_hash` load-balancing :weary: ) but I can't simply swap it because it would require a lot of work that is not doable in a short timeframe.

I tried Nginx before but it doesn't support active health checks (except if paying for Plus).

After researching a lot I found out that Caddy is lightweight, has an appealing config file format and supports ip_hash.

Now I'm struggeling with the last problem - maybe a show-stopper, idk - that the load-balancer should use the IP from X-Forwarded-For as remote-ip before it is hashed.

I tried to set the X-Real-IP header to the X-Forwarded-For IP but it doesm't seem to work.

Somewhere else I found that the Log always shows the incoming headers, therefore always shows the IP of my other proxy, but according to the log, it's always the same upstream server that is selected. Therefore I assume that it's not working.

Any ideas if Caddy can do this?

Thx in advance!

francislavoie:
Query policy brought up in https://caddy.community/t/caddy-v2-lb-based-on-arg-name-nginx/19448/8. I'm not sure why we didn't have this, and it's super trivial to add, so why not 🤷

Closes #5469 for `client_ip_hash`

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -208,6 +208,16 @@ func cmdRun(fl Flags) (int, error) {
 		}
 	}
 
+	// create pidfile now, in case loading config takes a while (issue #5477)
+	if runCmdPidfileFlag != "" {
+		err := caddy.PIDFile(runCmdPidfileFlag)
+		if err != nil {
+			caddy.Log().Error("unable to write PID file",
+				zap.String("pidfile", runCmdPidfileFlag),
+				zap.Error(err))
+		}
+	}
+
 	// run the initial config
 	err = caddy.Load(config, true)
 	if err != nil {
@@ -240,16 +250,6 @@ func cmdRun(fl Flags) (int, error) {
 	// (this better only be used in dev!)
 	if runCmdWatchFlag {
 		go watchConfigFile(configFile, runCmdConfigAdapterFlag)
-	}
-
-	// create pidfile
-	if runCmdPidfileFlag != "" {
-		err := caddy.PIDFile(runCmdPidfileFlag)
-		if err != nil {
-			caddy.Log().Error("unable to write PID file",
-				zap.String("pidfile", runCmdPidfileFlag),
-				zap.Error(err))
-		}
 	}
 
 	// warn if the environment does not provide enough information about the disk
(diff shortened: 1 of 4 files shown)


mholt (approved):
This LGTM -- one thing to note about the current query hash implementation is that it only considers the first occurrence of a given QS parameter (by using `Get()`). That is not necessarily a problem but we should probably document that, as I've received "security" reports in the past based on parsing ambiguities (which are not a security bug in the web server, to be clear -- only in the upstream app actually using the QS).

francislavoie:
Hmm, interesting, I didn't realize Go didn't handle that. Makes sense though, works like the headers.

I pushed a commit to join all the values for a given key with a comma, should fix that problem.

mholt:
Interesting! Thanks for taking some care there.

Just trying to scrutinize it for a moment -- don't mind me... this logic assumes that query string parameter order is significant, i.e. `a=b&a=c` != `a=c&a=b`. Some parsers/apps may expect a second QS value (with the same key) to overwrite the first; others will expect the first to cause repeated keys to be ignored; others will want to consider all key-value pairs as an array, but sorted who-knows-how.

Anyway, I think what I'm getting at is that your initial approach was fine -- it's just one opinionated way of doing it (and there is no unopinionated way that I know of), and so we should document whatever approach we take.

That said, I do think that the current approach that considers repeated values is probably more comprehensive -- but will still be wrong for some people :upside_down_face: -- and it's less efficient than the first approach.

So... I don't really know what we should do. I don't think changing it later would be considered a breaking change since I don't think it makes sense to map backends with query string affinity in the first place.

Maybe what we should do is just take the most efficient route for now, and document the "hash" as a-currently unspecified implementation detail that may change. That way we should be able to get away with doing it however we want to, but we can also change it later.

francislavoie:
We spoke about this in more detail in Slack.

I think there's less risk by joining the values, I think it plugs more potential holes.

The order of query params is significant; consider a query like `foo[]=1&foo[]=2` where `foo` might be a search filter with drag&drop UI to set priority. Changing the order of those keys means it is a semantically different request, so I think it makes sense that this wouldn't necessarily map to the same upstream.

I agree that we need to document that the implementation details are subject to change, but I think it'll be fine.

PR #5472, opened 2023-03-31
reverseproxy: Add mention of which half a copyBuffer err comes from

francislavoie:
Just a possible debug aid. It's not always obvious which half the error comes from.

### DIFF
--- a/modules/caddyhttp/reverseproxy/streaming.go
+++ b/modules/caddyhttp/reverseproxy/streaming.go
@@ -224,7 +224,7 @@ func (h Handler) copyBuffer(dst io.Writer, src io.Reader, buf []byte) (int64, er
 		}
 		if rerr != nil {
 			if rerr == io.EOF {
-				rerr = nil
+				return written, nil
 			}
 			return written, fmt.Errorf("reading: %v", rerr)
 		}
--- a/modules/caddyhttp/reverseproxy/streaming.go
+++ b/modules/caddyhttp/reverseproxy/streaming.go
@@ -20,6 +20,7 @@ package reverseproxy
 
 import (
 	"context"
+	"fmt"
 	"io"
 	weakrand "math/rand"
 	"mime"
@@ -215,7 +216,7 @@ func (h Handler) copyBuffer(dst io.Writer, src io.Reader, buf []byte) (int64, er
 				written += int64(nw)
 			}
 			if werr != nil {
-				return written, werr
+				return written, fmt.Errorf("writing: %v", werr)
 			}
 			if nr != nw {
 				return written, io.ErrShortWrite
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -225,7 +226,7 @@ func (h Handler) copyBuffer(dst io.Writer, src io.Reader, buf []byte) (int64, er
 			if rerr == io.EOF {
 				rerr = nil
 			}
-			return written, rerr
+			return written, fmt.Errorf("reading: %v", rerr)
```
mohammed90, line 229: This error is nil. It shouldn't be wrapped, just returned or nil in place.

francislavoie: It's only nil if EOF. I guess we should just change the EOF condition to return instead

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -215,17 +216,17 @@ func (h Handler) copyBuffer(dst io.Writer, src io.Reader, buf []byte) (int64, er
 				written += int64(nw)
 			}
 			if werr != nil {
-				return written, werr
+				return written, fmt.Errorf("writing: %v", werr)
```
mholt, line 219:
```suggestion
				return written, fmt.Errorf("writing: %w", werr)
```

I think this is one area of the code where wrapping the error might be useful in case we want to figure out what specific kind of error it is.

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -215,17 +216,17 @@ func (h Handler) copyBuffer(dst io.Writer, src io.Reader, buf []byte) (int64, er
...
 				return written, io.ErrShortWrite
 			}
 		}
 		if rerr != nil {
 			if rerr == io.EOF {
-				rerr = nil
+				return written, nil
 			}
-			return written, rerr
+			return written, fmt.Errorf("reading: %v", rerr)
```
mholt, line 229:
```suggestion
			return written, fmt.Errorf("reading: %w", rerr)
```

mholt (changes requested):
Looks useful, thanks. I might suggest wrapping the error as that could be useful in the future since I don't think we've seen the end of patches in this part of the code. :upside_down_face:

modules/caddyhttp/reverseproxy/streaming.go
francislavoie: I thought you were generally against using `%w` for some reason :thinking: 

No objections from me :+1:

mholt (approved):

PR #5475, opened 2023-04-02
fileserver: Add color-scheme meta tag

kidonng:
Use [`<meta name="color-scheme">`](https://web.dev/color-scheme/) to mark the page compatible with both light and dark themes, and let browser adapt to the themes as well.

Before:

image

After:

image

### DIFF
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -94,6 +94,7 @@
 	<head>
 		<title>{{html .Name}}</title>
 		<meta charset="utf-8">
+		<meta name="color-scheme" content="light dark">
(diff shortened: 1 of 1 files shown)


mholt:
Thanks, I hadn't heard of this before.

PS. Are those screenshots from the latest on `master`? We have a new browse template. I'm sure it'll work fine for that too but might be worth checking :D

kidonng:
Oops, should have clarified, the screenshots is taken with the latest release.

The meta tag just tells the brower to blend in, so it would likely to look fine if custom styles are well-made. But I agree it's worth checking.

francislavoie (approved):

mholt:
Gotcha, yeah that should be fine then :) Thanks!

PR #5484, opened 2023-04-09
reverseproxy: max TLS version client config

kkroo:
This addresses an issue where an upstream proxy incorrectly advertises TLS 1.3 support when the origin only supports TLS 1.2. Rather than throwing all security out the window, it allows the reverse proxy to correctly negotiate and verify a handshake.


francislavoie:
I'm confused. When you say "origin", do you mean the client that originated the request? Why does the client's TLS version matter for the proxy? It's two completely separate TLS connections (one from client/browser to Caddy, another from Caddy to to the upstream). Maybe I'm missing something obvious though.

mholt:
Could you elaborate with details on exactly what is broken? As this is a security protocol I'm hesitant to accept this change without further understanding.

PR #5487, opened 2023-04-12
reverseproxy: Optimize the base case for `least_conn` policy

francislavoie:
When only a single request has the least amount of requests, there's no need to compute a random number, because the modulo of 1 will always be 0 anyways.


francislavoie:
I just noticed `random_choice` also uses similar `leastRequests` logic and was also missing the base case of one option. Added a condition for that too.

mholt (approved):
Cool, let's save those cycles! Thanks!

PR #5488, opened 2023-04-12
reverseproxy: Add `fallback` for some policies, instead of always random

### LINKED ISSUE #5486: Question about load balancing
So what I am wanting is to load balance with least conn algorithm. But while doing so, I do not want my client to lose his session details as the flow of my app continues. So for that, is it right to use header up to set the session cookie? **Look at the example below**

flow:
client server request -> backend pick least conn server -> pass cookie header when it is present to uplink server stream -> server receives request

```
reverse_proxy :5000 :3000 {
    lb_policy least_conn
    header_up  Cookie {cookie.*}
}
```

francislavoie:
Closes #5486

This adds a new `fallback` option for some load balancing policies: `query`, `header` and `cookie`. See https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#load-balancing

Previously, the default was to always choose a random upstream when using those policies when there's no value available. Now, the fallback policy is configurable.

Example usecases:
- `cookie` + `least_conn`, where the first request (with no cookie) gets the upstream with the fewest current requests, instead of a random one
- `header` + `first` where if a header isn't provided, the first available upstream is chosen instead; so essentially the `first` policy allows for primary/secondary behaviour normally, but allows for an escape hatch for testing the secondary upstream by using a header

In the Caddyfile, it looks like this:

```
reverse_proxy :8001 :8002 :8003 {
	lb_policy cookie LB-Cookie-Name SuperSecretHmacKey {
		fallback least_conn
	}
}
```

Also, an insane case like this could be done:

```
reverse_proxy :8001 :8002 :8003 {
	lb_policy cookie LB-Cookie-Name SuperSecretHmacKey {
		fallback cookie A-Different-Cookie {
			fallback cookie Another-Different-Cookie {
				fallback cookie Yet-Another-Different-Cookie
			}
		}
	}
}
```

There's lots of boilerplate here unfortunately, I tried to reduce it where I could. Added a new test for `cookie` + `first` since it's an easy case. Now, the policies need provisioning, some default value cases were hoisted to the Provision function.

I used Copilot to speed up writing this PR, it wrote a lot of the boilerplate for me. I've reviewed and tested the code it generated.

### DIFF
--- a/modules/caddyhttp/reverseproxy/selectionpolicies.go
+++ b/modules/caddyhttp/reverseproxy/selectionpolicies.go
@@ -18,6 +18,7 @@ import (
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"hash/fnv"
 	weakrand "math/rand"
@@ -29,6 +30,7 @@ import (
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
 )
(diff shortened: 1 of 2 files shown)


mholt (approved):
Cool, I think this makes sense. Thank you!

PR #5494, opened 2023-04-14
reverseproxy: Expand port ranges to multiple upstreams in CLI + Caddyfile

francislavoie:
See discussion in https://caddy.community/t/reverse-proxy-to-backend-over-range-of-ports/19602

This doesn't work for JSON configs because this expanding happens at config-adapt-time. See the forum thread for an explanation.

### DIFF
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -15,6 +15,7 @@
 package reverseproxy
 
 import (
+	"fmt"
 	"net/http"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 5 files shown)


francislavoie:
Ah crap, I was only running one part of the test suite so I missed other problems :grimacing: digging deeper.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -157,7 +158,24 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 		}
 		commonScheme = scheme
 
-		h.Upstreams = append(h.Upstreams, &Upstream{Dial: dialAddr})
+		parsedAddr, err := caddy.ParseNetworkAddress(dialAddr)
+		if err != nil {
+			return d.WrapErr(err)
+		}
+
+		if parsedAddr.StartPort == 0 && parsedAddr.EndPort == 0 {
```
mholt, line 166: Why 0? :thinking:

francislavoie: I don't remember 🤔 might've been for placeholders which have no port (assuming the port is part of the placeholder value) but I'll need to take a closer look to remind myself. I'll add comments.

francislavoie: Oh right, unix networks have port 0

mholt (approved):
Neat, thanks. It is interesting that this works only in the Caddyfile and not in the JSON (but makes sense -- and I think it's fine).

Thank you for the feature, as usual! :blush:

PR #5495, opened 2023-04-15
chore: Adjustments to CI caching

francislavoie:


mohammed90:
I think we need actions/setup-go#366, otherwise the cache gets clobbered by the multiple runners, no?

francislavoie:
I don't think so. I looked at the source, and their cache key is `OS name + Go version + go.sum`. We would only need that if we used different OS versions I think.

mohammed90:
> 

Ok, cool!

mohammed90 (approved):
Thanks for picking this up, Francis!

PR #5497, opened 2023-04-17
templates: Add `fileStat` function

taophp:
May be useful in a loop handling listFiles entries to adopt different behavior is a file is a directory.

### DIFF
--- a/modules/caddyhttp/templates/tplcontext.go
+++ b/modules/caddyhttp/templates/tplcontext.go
@@ -404,13 +404,13 @@ func (c TemplateContext) funcIsDir(name string) (bool, error) {
 
 	dir, err := c.Root.Open(path.Clean(name))
 	if err != nil {
-		return nil, err
+		return false, err
 	}
 	defer dir.Close()
 
 	stat, err := dir.Stat()
 	if err != nil {
-		return nil, err
+		return false, err
 	}
 	return stat.IsDir(), nil
 }
--- a/modules/caddyhttp/templates/tplcontext.go
+++ b/modules/caddyhttp/templates/tplcontext.go
@@ -80,7 +80,6 @@ func (c *TemplateContext) NewTemplate(tplName string) *template.Template {
 		"markdown":         c.funcMarkdown,
 		"splitFrontMatter": c.funcSplitFrontMatter,
 		"listFiles":        c.funcListFiles,
-		"isDir":            c.funcIsDir,
 		"stat":             c.funcStat,
 		"env":              c.funcEnv,
 		"placeholder":      c.funcPlaceholder,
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/templates/tplcontext.go
```diff
@@ -79,6 +79,7 @@ func (c *TemplateContext) NewTemplate(tplName string) *template.Template {
 		"markdown":         c.funcMarkdown,
 		"splitFrontMatter": c.funcSplitFrontMatter,
 		"listFiles":        c.funcListFiles,
+		"isDir":            c.isDir,
```
mohammed90, line 82:
```suggestion
		"isDir":            c.funcIsDir,
```

francislavoie:
The CLA isn't happy because the email on your git commits doesn't match the ones on your Github account. Make sure to fix that!

taophp:
> The CLA isn't happy because the email on your git commits doesn't match the ones on your Github account. Make sure to fix that!

Fixed ! Thanks and sorry.

mholt:
Should we just return the file info instead of only the IsDir result? Then the user can get file size, etc.

taophp:
> Thanks for the contribution!

My pleasure @mholt .

> But maybe it would make more sense to return the whole file info instead of only the IsDir result? Then the user can get file size, etc.

Well, this surely is a good idea. But, in my use case, I do not need it currently. And, as it's the first time I code in Go, I do not feel very confident to handle this job : my contribution is essentially a rewrite of the `funcFileExists` function... It's up to you to add a more complete `fileinfo` function.

If you decide to write it, I should still appreciate that you keep the `isDir` function: apart from the honor for me to become a contributor to Caddy, I find it really practical to use, and I would doubtless not be the only one; moreover, it seems consistent with the Go language.

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -79,6 +79,7 @@ func (c *TemplateContext) NewTemplate(tplName string) *template.Template {
 		"markdown":         c.funcMarkdown,
 		"splitFrontMatter": c.funcSplitFrontMatter,
 		"listFiles":        c.funcListFiles,
+		"isDir":            c.funcIsDir,
```
mholt, line 82:
```suggestion
		"stat":            c.funcIsDir,
```

(Maybe rename `funcIsDir` to `funcStat`)

taophp:
@mholt OK, thanks for you help. I committed an new version, improved in the way you suggested.

taophp:
Sorr @francislavoie , I just forgot to remove my first function...

taophp:
> Have you tried it yet to make sure it does what you need? (What does your relavent template code look like?)

Yes, I did, and it works fine ! Thanks for your help. My code in the template looks like:
```
{{- $listFile := listFiles "/" -}}
{{- range $filePath := $listFile -}}
	{{ $stat := stat $filePath }}
	{{ $stat.IsDir }}
{{- end -}}
```

Issue #5496, opened 2023-04-15
Clear removed upstreams metrics on config reload


husam212:
If an upstream is removed from the configuration file, then Caddy is reloaded with `reload` command, the removed upstream will keep showing in Prometheus metrics, the value of the metric is stuck at the last state of the removed upstream.

Caddy version: 2.6.4

hairyhenderson:
Can you provide more details? An example, maybe? This shouldn't be a problem in general for counters or histograms, since they won't be growing.

francislavoie:
FYI @david-szabo97 since you implemented the upstream metrics

husam212:
@francislavoie sure, considering the following example configuration.
```
{
  servers {
    metrics
  }
}

:9000 {
  metrics
}

example.com {
  reverse_proxy * {
    to 10.0.0.1:8000
    to 10.0.0.2:8000
    health_uri /ping
    health_headers {
      Host example.com
    }
  }
}
```

Checking the metrics with the above config would result in something like this:
```
caddy_reverse_proxy_upstreams_healthy{upstream="10.0.0.1:8000"} 1
caddy_reverse_proxy_upstreams_healthy{upstream="10.0.0.2:8000"} 1
```

Now If I removed one of the upstreams, lets say 10.0.0.1:8000, then reloaded Caddy, I would still get the same output from the metrics endpoint.

hairyhenderson:
Thanks @husam212 - this is indeed a problem, especially since that metric is a gauge, and so is misrepresenting reality.

u5surf:
@husam212  @francislavoie @david-szabo97 
We had the same problem.
We dove deeply the issue, we can find this `sync.Once` in `newMetricsUpstreamsHealthyUpdater` is unnecessary which is called by `reverseproxy.Provision` exactly once per `caddy run/start`. Thus it never call by `caddy reload` because it has already called in `caddy run/start` at only once.

https://github.com/caddyserver/caddy/blob/998c6e06a75030cccabcc08790bc32642791aff7/modules/caddyhttp/reverseproxy/reverseproxy.go#L419-L420

https://github.com/caddyserver/caddy/blob/998c6e06a75030cccabcc08790bc32642791aff7/modules/caddyhttp/reverseproxy/metrics.go#L38-L40

And moreover, It is necessary to reinitialize with prometheus.Unregister.

https://github.com/prometheus/client_golang/issues/475#issuecomment-428142144

[...]

PR #5498, opened 2023-04-17
reverseproxy: Fix reinitialize upstream healthy metrics

u5surf:
Fixes: #5496 

We can find this `sync.Once` in `newMetricsUpstreamsHealthyUpdater` is unnecessary which is called by `reverseproxy.Provision` exactly once per `caddy run/start`. Thus it never call by `caddy reload` because it has already called in `caddy run/start` at only once.

https://github.com/caddyserver/caddy/blob/998c6e06a75030cccabcc08790bc32642791aff7/modules/caddyhttp/reverseproxy/reverseproxy.go#L419-L420

https://github.com/caddyserver/caddy/blob/998c6e06a75030cccabcc08790bc32642791aff7/modules/caddyhttp/reverseproxy/metrics.go#L38-L40

And moreover, It is necessary to reinitialize with prometheus.Unregister.

https://github.com/prometheus/client_golang/issues/475#issuecomment-428142144

Note: We can be prepared our workaround patches which seems to work fine, but it might have cover only our particular usecase.

### DIFF
--- a/modules/caddyhttp/reverseproxy/metrics.go
+++ b/modules/caddyhttp/reverseproxy/metrics.go
@@ -39,6 +39,8 @@ func newMetricsUpstreamsHealthyUpdater(handler *Handler) *metricsUpstreamsHealth
 		initReverseProxyMetrics(handler)
 	})
 
+	reverseProxyMetrics.upstreamsHealthy.Reset()
+
 	return &metricsUpstreamsHealthyUpdater{handler}
 }
 


mholt (approved):
Thanks for the patch!

Metrics are not my area of expertise, so I'll leave this for another maintainer to review, but overall LGTM. :)

francislavoie:
If multiple `reverse_proxy` handlers exist in a config, this would unregister and re-register the gauge multiple times. Is that a problem? I guess it's probably fine but it's kinda funky.

david-szabo97:
Thanks @u5surf for working on this!

Would `.Reset()` work in this case? We could keep the metrics as a singleton and just call `.Reset()` if the metric already exists.

Something like this...
```go
func newMetricsUpstreamsHealthyUpdater(handler *Handler) *metricsUpstreamsHealthyUpdater {
	reverseProxyMetrics.init.Do(func() {
		initReverseProxyMetrics(handler)
	})

        reverseProxyMetrics.upstreamsHealthy.Reset()

        return &metricsUpstreamsHealthyUpdater{handler}
}
```
and we can keep the rest of the code the same.

Although this is just brainstorming, I'm not really sure how `UnRegister()` and `Reset()` differs.

https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.Reset

u5surf:
@francislavoie 
> If multiple reverse_proxy handlers exist in a config, this would unregister and re-register the gauge multiple times.

Exactly, I concerned about the same issue in case that the multiple threads have pathed  `newMetricsUpstreamsHealthyUpdater` in `reverseproxy.Provision` it seems to redundant slightly which you said repetition of initialize. It seems good way to implement the sync.Once with reset mechanism, but it might be too much.

@david-szabo97
Thanks reviewing and that's cool approach which I couldn't come up with. Moreover indeed It can work fine completely! 
I can't see why could not deletes all metrics in this vector  after the`reverseProxyMetrics.init.Do` is called for the first time.
Could you tell me the detail of the mechanism?

hairyhenderson (approved):

mholt:
Great, thank you everyone!

PR #5500, opened 2023-04-19
acmeserver: Configurable `resolvers`, fix smallstep deprecations

### LINKED ISSUE #5476: Allow configuring DNS resolvers used by the internal acme_server to verify DNS challenges
When the internal ACME server has to verify a `DNS-01` challenge, it goes through the system resolver using  `net.LookupTXT`. It would be nice to have the ability to configure a set of name servers / resolvers to use instead of the system resolver. This setting already exists for when Caddy is acting as an ACME *client* (see [here](https://caddyserver.com/docs/json/apps/tls/automation/policies/issuers/acme/#challenges/dns/resolvers)), so it stands to reason they might be useful on the server side, too.

In this case, it came up while I was trying to come up with a way to test the new DNS server plugin I'm writing.

I've had a quick look at how this would be implemented, and unfortunately it might not be entirely straight-forward. The smallstep library *does* allow overriding the resolvers it uses to verify challenges, but not for the method Caddy is using.

Caddy uses [`smallstep/certificates/acme/api.NewHandler`](https://pkg.go.dev/github.com/smallstep/certificates@v0.23.2/acme/api#NewHandler) [here](https://github.com/caddyserver/caddy/blob/b6fe5d4b41d07e70a502ed58d40e8b0e75067db5/modules/caddypki/acmeserver/acmeserver.go#L159):
```go
acmeRouterHandler := acmeAPI.NewHandler(acmeAPI.HandlerOptions{
	CA:     auth,
	DB:     acmeDB,                            // stores all the server state
	DNS:    ash.Host,                          // used for directory links
	Prefix: strings.Trim(ash.PathPrefix, "/"), // used for directory links
})
```
This handler will create a default [`Client` ](https://pkg.go.dev/github.com/smallstep/certificates@v0.23.2/acme#Client) (see [this file](https://github.com/smallstep/certificates/blob/3c76834807fb38ff4ca0ee3ab12c617fd808f7cf/acme/api/handler.go#L77)), which is implemented [here](https://github.com/smallstep/certificates/blob/3c76834807fb38ff4ca0ee3ab12c617fd808f7cf/acme/client.go) (as `client` in lowercase, somewhat confusing), and looks pretty simple:
```go
func (c *client) Get(url string) (*http.Response, error) {
	return c.http.Get(url)
}

func (c *client) LookupTxt(name string) ([]string, error) {
	return net.LookupTXT(name)
}

func (c *client) TLSDial(network, addr string, config *tls.Config) (*tls.Conn, error) {
	return tls.DialWithDialer(c.dialer, network, addr, config)
}
```
I think a new [`Client` ](https://pkg.go.dev/github.com/smallstep/certificates@v0.23.2/acme#Client) would need to be implemented, with a different `LookupTxt`.

Then, the `acme_server` module [`Handler.Provision`](https://github.com/caddyserver/caddy/blob/b6fe5d4b41d07e70a502ed58d40e8b0e75067db5/modules/caddypki/acmeserver/acmeserver.go#L92) would need to switch from using [`smallstep/certificates/acme/api.NewHandler`](https://pkg.go.dev/github.com/smallstep/certificates@v0.23.2/acme/api#NewHandler) (which is sort of half-deprecated anyway) to the recommended [`smallstep/certificates/acme/api.Route`](https://pkg.go.dev/github.com/smallstep/certificates@v0.23.2/acme/api#Route) (or maybe it's  [`smallstep/certificates/api.Route`](https://pkg.go.dev/github.com/smallstep/certificates@v0.23.2/api#Route)? very confusing!). If you check the docs you'll see that this method is configured differently: it "requires that all the acme components, authority, db, client, linker, and prerequisite checker [are] present in the context". [`smallstep/certificates/acme/NewContext`](https://pkg.go.dev/github.com/smallstep/certificates/acme@v0.23.2#NewContext) is provided to add these things into the context. But, from what I can tell, it's not a 1-to-1 translation from the [`HandlerOptions`](https://pkg.go.dev/github.com/smallstep/certificates@v0.23.2/acme/api#HandlerOptions) (the way it is configured currently) to the things you can put into the context.
Finally, it looks like this context needs to be glued onto every incoming `http.Request` somehow - I assume this is a common Go idiom.

francislavoie:
Closes #5476


modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -175,6 +197,16 @@ func (ash *Handler) Provision(ctx caddy.Context) error {
...
 	if strings.HasPrefix(r.URL.Path, ash.PathPrefix) {
+		acmeCtx := acme.NewContext(
+			r.Context(),
+			ash.acmeDB,
+			ash.acmeClient,
+			ash.acmeLinker,
+			nil,
+		)
+		acmeCtx = authority.NewContext(acmeCtx, ash.acmeAuth)
+		r = r.WithContext(acmeCtx)
```
maraino, line 208: This is ok. You're doing this for every request, and it should be fine. A minor optimization would be to build it once and move it to the server BaseContext, this is what we do. But it will be available everywhere. 

Looking at chi, they have a middleware that we've never used named [ServerBaseContext](https://pkg.go.dev/github.com/go-chi/chi@v4.1.2+incompatible#ServerBaseContext). It allows you to define a base context for a handler, and it adds the context added by `net/http`. I don't think it's going to make an impact on the performance, but you can test and find out.

modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -227,6 +259,62 @@ func (ash Handler) openDatabase() (*db.AuthDB, error) {
...
+type resolverClient struct {
+	acme.Client
+
+	resolver *net.Resolver
+	ctx      context.Context
+}
+
+func (c resolverClient) LookupTxt(name string) ([]string, error) {
+	return c.resolver.LookupTXT(c.ctx, name)
+}
```
maraino, line 316: Just mentioning that this will only affect the DNS-01 challenge. It won't affect the resolvers for HTTP-01 or TLS-ALPN-01.

The issue just mentioned the DNS-01 challenge.

maraino (commented):
Looks ok, I've added some minor comments.

modules/caddypki/acmeserver/acmeserver.go
francislavoie: Yeah, that's fine for now. We can spend the effort to reconfigure the HTTP/TLS dialers later if necessary. The interface is simple so it's not too complicated.

modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -142,31 +165,30 @@ func (ash *Handler) Provision(ctx caddy.Context) error {
...
 
-	var acmeDB acme.DB
-	if authorityConfig.DB != nil {
-		acmeDB, err = acmeNoSQL.New(auth.GetDatabase().(nosql.DB))
-		if err != nil {
-			return fmt.Errorf("configuring ACME DB: %v", err)
-		}
+	ash.acmeDB, err = acmeNoSQL.New(ash.acmeAuth.GetDatabase().(nosql.DB))
+	if err != nil {
+		return fmt.Errorf("configuring ACME DB: %v", err)
```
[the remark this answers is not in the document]
francislavoie: @mholt has some opinions about wrapping errors so :man_shrugging: I'll let him tell me if he wants me to do this or not. But noted.

modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -227,6 +259,62 @@ func (ash Handler) openDatabase() (*db.AuthDB, error) {
...
+	if len(ash.resolvers) != 0 {
+		dialer := &net.Dialer{
+			Timeout: 2 * time.Second,
+		}
+		resolver = &net.Resolver{
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				//nolint:gosec
+				addr := ash.resolvers[weakrand.Intn(len(ash.resolvers))]
+				return dialer.DialContext(ctx, addr.Network, addr.JoinHostPort(0))
```
itsxaos, line 293: I think this should use the requested `address` instead of `addr.Network` instead, since DNS can use both TCP and UDP.
The way it is now, the user has to specify the resolver address as `udp/127.0.0.1` to get it to work (at all).
```suggestion
				return dialer.DialContext(ctx, network, addr.JoinHostPort(0))
```

modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -227,6 +259,62 @@ func (ash Handler) openDatabase() (*db.AuthDB, error) {
...
+			if strings.Contains(err.Error(), "missing port in address") {
+				addr, err = caddy.ParseNetworkAddress(v + ":53")
+			}
+			if err != nil {
+				return nil, err
+			}
+		}
+		if addr.PortRangeSize() != 1 {
+			return nil, fmt.Errorf("resolver address must have exactly one address; cannot call %v", addr)
+		}
```
itsxaos, line 279:
```suggestion
		}
		if addr.StartPort == 0 {
				addr.StartPort = 53
		}
```

modules/caddypki/acmeserver/acmeserver.go
```diff
@@ -227,6 +259,62 @@ func (ash Handler) openDatabase() (*db.AuthDB, error) {
...
+		addr, err := caddy.ParseNetworkAddress(v)
+		if err != nil {
+			// If a port wasn't specified for the resolver,
+			// try defaulting to 53 and parse again
+			if strings.Contains(err.Error(), "missing port in address") {
+				addr, err = caddy.ParseNetworkAddress(v + ":53")
+			}
+			if err != nil {
+				return nil, err
+			}
```
itsxaos, line 275: I don't think this is right: a port is optional for `caddy.ParseNetworkAddress`, so it does not return an error, it just sets the port to 0.

itsxaos (changes requested):

modules/caddypki/acmeserver/acmeserver.go
francislavoie: That's already handled by `addr, err = caddy.ParseNetworkAddress(v + ":53")`

francislavoie: Oh, I see what you mean. I think https://github.com/caddyserver/caddy/commit/1e18afb5c862d62be130d563785de5c58f08ae8e broke that port fallback. I'll need to dig deeper into this.

modules/caddypki/acmeserver/acmeserver.go
francislavoie: Hmm, well it is intended to be able to specify the network in the address to override. But it is true that it currently defaults to `tcp` which is not ideal for a DNS resolver. I think I'll make another `ParseNetworkAddress` fn that allows specifying the default network, so it can default to `udp` instead.

modules/caddypki/acmeserver/acmeserver.go
francislavoie: Okay, I pushed another commit which adds a new `ParseNetworkAddressWithDefaults` function, so we can set the defaults to `udp` and port `53` for this case. Could you try this new build?

listeners.go
```diff
@@ -303,13 +303,19 @@ func IsUnixNetwork(netw string) bool {
 // Network addresses are distinct from URLs and do not
 // use URL syntax.
 func ParseNetworkAddress(addr string) (NetworkAddress, error) {
+	return ParseNetworkAddressWithDefaults(addr, "tcp", 0)
+}
+
+// ParseNetworkAddressWithDefaultNetworkPort is like ParseNetworkAddress
+// but allows the default network and port to be specified.
```
itsxaos, line 310:
```suggestion
// ParseNetworkAddressWithDefaults is like ParseNetworkAddress but allows
// the default network and port to be specified.
```

francislavoie: Whoops, yeah thanks, that's the name for the function I had in between 😅

itsxaos:
Changes look good to me, but it will be a couple of hours until I can test them.

itsxaos:
Alright, I tried the changes and everything seems to be working as expected!

modules/caddypki/acmeserver/acmeserver.go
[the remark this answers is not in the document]
mholt: It just means you have to be OK with error values being part of the exported API. Since `Provision()` is called by the caddy core, there's a good chance it won't know what to type-assert the error as, and thus exporting it here won't hurt but also won't help IMO.

Great tip though @maraino -- thanks for pointing that out!

modules/caddypki/acmeserver/acmeserver.go
mholt: @francislavoie Do you want me to hold up on merging until we make that optimization? Or should we go with this and then optimize later?

francislavoie: There's no rush on that one. I think we can merge.

mholt (approved):
Thanks for implementing this!

PR #5504, opened 2023-04-20
Rewrite: use escaped path

### LINKED ISSUE #5278: try_files possibly double-escaping URI-encoded characters (%3F)
[Moving this one over from [the community forums](https://caddy.community/t/handling-3f-in-uri-with-try-files/18301/4).]

While testing out Caddy to replace nginx I noticed that the `try_files` directive might be double-escaping URI-encoded strings (in particular, `%3F` (`?`). I ran into this because I have files in my filesystem with `?` in their names, from mirroring an old Wordpress site that I access via a `try_files` config.

I'm running Caddy version `v2.6.2 h1:wKoFIxpmOJLGl3QXoo6PNbYvGW4xLEgo32GPBEjWL8o=`, accessing it directly.

### Steps to reproduce

To eliminate the rest of my config for my live site, I duplicated the problem with a tiny test scenario. I set up this config:

```
mati.ca, www.mati.ca {
        root * /home/rich/sites/mati.ca
        log
        file_server browse {
                index index.html index.htm
        }
        try_files {path}
}
```

and created three files in the site root directory, named as follows. The content of each file is the same as the file’s filename, so I could tell what file caddy was returning.

```
file.txt
file.txt?honk
file.txt%3Fhonk
```

### Expected behavior

When I `curl https://www.mati.ca/file.txt%3Fhonk`, it should serve the file named `file.txt?honk`.

```
$ curl https://www.mati.ca/file.txt?honk
file.txt
$ curl https://www.mati.ca/file.txt%3Fhonk
file.txt?honk
```

### Observed behavior

```
$ curl https://www.mati.ca/file.txt?honk
file.txt
$ curl https://www.mati.ca/file.txt%3Fhonk
file.txt
```

(*Without* `try_files {path}` I get the expected behavior.)

TP-O:
Fix #5278

Using unescaped path in rewriting makes me confuse which part after `?` is a query if the path contains `%3F`, so I thought it would be better to use an escaped path.


mholt:
Thanks; this will need careful review, and in consideration with #5438.

modules/caddyhttp/rewrite/rewrite_test.go
```diff
@@ -323,7 +333,6 @@ func TestRewrite(t *testing.T) {
 			input:  newRequest(t, "GET", "/foo/findme%2Fbar"),
 			expect: newRequest(t, "GET", "/foo/replaced%2Fbar"),
 		},
-
```
mholt, line 326: I usually leave a line break sometimes to weakly organize test cases into groups. It'd be convenient if we could preserve this line break.

mholt (changes requested):
Honestly, I like the change. It's simpler code and all the exists tests still pass, along with two new test cases.

Do the test cases represent the original issue sufficiently?

Additionally, I wonder if @wlonkly could chime in and confirm that this patch works for you.

This is one of the few areas of the code that is more heavily tested, so I guess we can be confident with this, I'm just wary of unexpected edge cases especially if they relate to security issues :upside_down_face:

TP-O:
> Honestly, I like the change. It's simpler code and all the exists tests still pass, along with two new test cases.
> 
> Do the test cases represent the original issue sufficiently?
> 
> Additionally, I wonder if @wlonkly could chime in and confirm that this patch works for you.
> 
> This is one of the few areas of the code that is more heavily tested, so I guess we can be confident with this, I'm just wary of unexpected edge cases especially if they relate to security issues upside_down_face

I think these test cases make sure the `Rewrite` function distinguishes between `%3F` and `?` when checking for query string after replacing.

mholt:
Thanks again for the contribution! If there are any issues reported we'll let you know :+1:

PR #5505, opened 2023-04-20
caddyfile: Stricter parsing, error for brace on new line

francislavoie:
Feedback from https://caddy.community/t/caddy-as-an-authentication-proxy-to-secure-endpoint/19691 made me realize we should be a bit stricter with parsing braces.

This makes this config cause an error:

```
:8001 {
    reverse_proxy https://example.com 
    {
    }
}
```

```
Error: Caddyfile.test3:3 - Error during parsing: Unexpected '{' on a new line; did you mean to place the '{' on the previous line?
```

The other changes to `reverse_proxy`/`php_fastcgi`/`forward_auth` were to fix the error `expecting 'https://' but got '://'` from that thread; essentially, the upstreams were being parsed for each iteration of the unmarshal loop with `RemainingArgs()` but that doesn't make sense. Changing the loop means both the shortcut directives would misbehave due to the dispenser cursor being in the wrong place, so they need to be advanced before calling down.

### DIFF
--- a/caddyconfig/caddyfile/parse.go
+++ b/caddyconfig/caddyfile/parse.go
@@ -520,6 +520,9 @@ func (p *parser) directive() error {
 			if !p.isNextOnNewLine() && p.Token().wasQuoted == 0 {
 				return p.Err("Unexpected next token after '{' on same line")
 			}
+			if p.isNewLine() {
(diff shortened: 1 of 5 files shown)


mholt (approved):
This is great. Wow. Thanks for the improvement!!

Issue #5336, opened 2023-01-27
Allow http.handlers	to add field to log output


WeidiDeng:
Handlers should be able to add field to server log output for debugging and extending functionality purpose.

Currently when a http.handler can only emit log using its own logger provisioned when creating an instance of the module. A module can attach some context information to the request handled.

Currently caddy tries to determine user_id from [basicauth](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/server.go#L276). However if it's not configured it will be empty and people maybe confused by its presence. (Commonly, most usage of caddy is just reverse proxy and file server.). Trying from the entry ServeHttp function is simply too much to maintain for every handler module that caddy is shipped with.

A recent pr will also add a new field to determine client_ip and it's also available as a [variable](https://github.com/caddyserver/caddy/pull/5104/files).

Allow http.handlers to add custom fields to log can help generate logs that are easier to analyse.

francislavoie:
> Currently caddy tries to determine user_id from basicauth.

Not exclusively. Any module that implements the `Authentication` interface will implicitly have `http.auth.user.id` set. I think the `caddy-security` module makes use of that behaviour, or at least did at some point in the past, not sure if it changed away from it.

> However if it's not configured it will be empty and people maybe confused by its presence.

I'd argue that it not being in the logs by default means it's less configurable that the property might exist if auth is used. Maybe a weak argument, but worth mentioning.

Another argument is that some log ingest tooling might not behave well if some fields are sometimes included and sometimes not, at the top level. It's the difference between `undefined` and `null` in some type systems.

> Handlers should be able to add field to server log output for debugging

I'd argue that if the purpose is for debugging, then it should be logged as a `debug` level log from that module, with a separate logger name (i.e. the handler module's logger, instead of the access logs).

> and extending functionality purpose.

This I think is a more compelling reason; it's true that allowing custom modules to manipulate access logging can open doors if there's certain properties of requests that should be pulled out into easier-access log fields.

A recent pr will also add a new field to determine client_ip and it's also available as a [variable](https://github.com/caddyserver/caddy/pull/5104/files).

Yep -- adding a new field to the logs is one of the main reasons I'm acting slower on that PR, because I see adding a log field as a breaking change.

If we decide that having some log fields sometimes not exist is okay, then that makes that PR easier to merge because it would be less of a breaking change.

francislavoie:
FYI @WeidiDeng I made a change in #5507 that allows handlers to add fields to the access logs.

[...]

PR #5507, opened 2023-04-23
logging: Add traceID field to access logs when tracing is active

hairyhenderson:
Tracing is far more useful when we can correlate logs with traces. This is done by adding a `traceID` field to the access logs.

I've attempted to add this in with a minimum of performance impact, and as such have moved the request logging function out of `ServeHTTP` and into a new method that can be tested independently. I've added some basic unit tests and a couple of benchmarks to gauge the impact.

Here are benchmark results before the `traceID` field was added:

```console
$ go test -benchtime=3s -count=5 -benchmem -run=^$ -bench ^BenchmarkServer ./modules/caddyhttp | tee before.txt
goos: darwin
goarch: amd64
pkg: github.com/caddyserver/caddy/v2/modules/caddyhttp
cpu: Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
BenchmarkServer_LogRequest-16                    3927310               920.3 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest-16                    3940473               913.4 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest-16                    3886784               913.1 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest-16                    4025732               919.2 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest-16                    3943749               924.9 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3859365               921.3 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3902739               925.0 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3910292               932.5 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3933997               922.4 ns/op           417 B/op       4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3993286               918.3 ns/op           417 B/op       4 allocs/op
PASS
ok      github.com/caddyserver/caddy/v2/modules/caddyhttp       45.867s
```

After:

```console
$ go test -benchtime=3s -count=5 -benchmem -run=^$ -bench ^BenchmarkServer ./modules/caddyhttp | tee after.txt
goos: darwin
goarch: amd64
pkg: github.com/caddyserver/caddy/v2/modules/caddyhttp
cpu: Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
BenchmarkServer_LogRequest-16                    3820621               949.4 ns/op           417 B/op          4 allocs/op
BenchmarkServer_LogRequest-16                    3860674               943.5 ns/op           417 B/op          4 allocs/op
BenchmarkServer_LogRequest-16                    3833852               939.1 ns/op           417 B/op          4 allocs/op
BenchmarkServer_LogRequest-16                    3858829               944.0 ns/op           417 B/op          4 allocs/op
BenchmarkServer_LogRequest-16                    3835098               940.2 ns/op           417 B/op          4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3306637              1037 ns/op             481 B/op          4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3424615              1047 ns/op             481 B/op          4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3484468              1037 ns/op             481 B/op          4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3476361              1215 ns/op             481 B/op          4 allocs/op
BenchmarkServer_LogRequest_WithTraceID-16        3457842              1062 ns/op             481 B/op          4 allocs/op
PASS
ok      github.com/caddyserver/caddy/v2/modules/caddyhttp       47.080s
```

Here's the diff:

```console
$ benchstat before.txt after.txt 
name                              old time/op    new time/op    delta
Server_LogRequest-16                 918ns ± 1%     943ns ± 1%   +2.73%  (p=0.008 n=5+5)
Server_LogRequest_WithTraceID-16     924ns ± 1%    1080ns ±13%  +16.85%  (p=0.008 n=5+5)

name                              old alloc/op   new alloc/op   delta
Server_LogRequest-16                  417B ± 0%      417B ± 0%     ~     (all equal)
Server_LogRequest_WithTraceID-16      417B ± 0%      481B ± 0%  +15.35%  (p=0.008 n=5+5)

name                              old allocs/op  new allocs/op  delta
Server_LogRequest-16                  4.00 ± 0%      4.00 ± 0%     ~     (all equal)
Server_LogRequest_WithTraceID-16      4.00 ± 0%      4.00 ± 0%     ~     (all equal)
```

_(note: I've edited this to reflect updates since the initial push - there was a performance inconsistency in the benchmark code which I've since fixed)_

IMO the ~15% hit is worth the extra field.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -698,23 +698,19 @@ func (s *Server) logRequest(
 
 	extra := r.Context().Value(ExtraLogFieldsCtxKey).(*ExtraLogFields)
 
-	// preallocate with the amount of known fields
 	fieldCount := 6
-	fields := make([]zapcore.Field, fieldCount+len(extra.fields))
-	fields[0] = zap.Int("bytes_read", reqBodyLength)
-	fields[1] = zap.String("user_id", userID)
-	fields[2] = zap.Duration("duration", *duration)
-	fields[3] = zap.Int("size", wrec.Size())
-	fields[4] = zap.Int("status", wrec.Status())
-	fields[5] = zap.Object("resp_headers", LoggableHTTPHeader{
-		Header:               wrec.Header(),
-		ShouldLogCredentials: shouldLogCredentials,
-	})
-
-	// add extra fields after the known fields
-	for i, field := range extra.fields {
-		fields[fieldCount+i] = field
-	}
+	fields := make([]zapcore.Field, 0, fieldCount+len(extra.fields))
+	fields = append(fields,
+		zap.Int("bytes_read", reqBodyLength),
+		zap.String("user_id", userID),
+		zap.Duration("duration", *duration),
+		zap.Int("size", wrec.Size()),
+		zap.Int("status", wrec.Status()),
+		zap.Object("resp_headers", LoggableHTTPHeader{
+			Header:               wrec.Header(),
+			ShouldLogCredentials: shouldLogCredentials,
+		}))
+	fields = append(fields, extra.fields...)
 
 	log("handled request", fields...)
 }
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -282,46 +282,7 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// capture the original version of the request
 		accLog := s.accessLogger.With(loggableReq)
 
-		defer func() {
-			// this request may be flagged as omitted from the logs
-			if skipLog, ok := GetVar(r.Context(), SkipLogVar).(bool); ok && skipLog {
-				return
-			}
-
-			repl.Set("http.response.status", wrec.Status()) // will be 0 if no response is written by us (Go will write 200 to client)
-			repl.Set("http.response.size", wrec.Size())
-			repl.Set("http.response.duration", duration)
-			repl.Set("http.response.duration_ms", duration.Seconds()*1e3) // multiply seconds to preserve decimal (see #4666)
-
-			logger := accLog
-			if s.Logs != nil {
-				logger = s.Logs.wrapLogger(logger, r.Host)
-			}
-
-			log := logger.Info
-			if wrec.Status() >= 400 {
-				log = logger.Error
-			}
-
-			userID, _ := repl.GetString("http.auth.user.id")
-
-			reqBodyLength := 0
-			if bodyReader != nil {
-				reqBodyLength = bodyReader.Length
-			}
-
-			log("handled request",
-				zap.Int("bytes_read", reqBodyLength),
-				zap.String("user_id", userID),
-				zap.Duration("duration", duration),
-				zap.Int("size", wrec.Size()),
-				zap.Int("status", wrec.Status()),
-				zap.Object("resp_headers", LoggableHTTPHeader{
-					Header:               wrec.Header(),
-					ShouldLogCredentials: shouldLogCredentials,
-				}),
-			)
-		}()
+		defer s.logRequest(accLog, r, wrec, &duration, repl, bodyReader, shouldLogCredentials)
 	}
 
 	start := time.Now()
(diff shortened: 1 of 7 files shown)


francislavoie:
Related to https://github.com/caddyserver/caddy/issues/5336

modules/caddyhttp/server.go
```diff
@@ -394,6 +355,57 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	fields = append(fields, zap.Int("size", wrec.Size()))
+	fields = append(fields, zap.Int("status", wrec.Status()))
+	fields = append(fields, zap.Object("resp_headers", LoggableHTTPHeader{
+		Header:               wrec.Header(),
+		ShouldLogCredentials: shouldLogCredentials,
+	}))
+
+	traceID, _ := GetVar(r.Context(), "traceID").(string)
+	if traceID != "" {
+		fields = append(fields, zap.String("traceID", traceID))
```
francislavoie, line 403: All our log fields are snake_case, should be consistent

```suggestion
		fields = append(fields, zap.String("trace_id", traceID))
```

hairyhenderson: IMO `traceID` is best as that's the default used by a number of tools. If we use snake-case for this we can modify tools, but that just makes it more difficult to do log/trace correlation with Caddy.

francislavoie:
I decided to piggy-back off this PR for https://github.com/caddyserver/caddy/issues/5336.

I don't love this though. This is kinda going into #4649 territory, we're doing allocations for the logging. We could use a `sync.Pool` for `ExtraLogFields` but idk 🤔 pooling isn't perfect, long-running requests/connections like websockets will hold onto them for a long time etc. We can throw away my commits if the approach doesn't make sense.

I probably want to move `user_id` to only log if non-empty (as an extra field essentially), and later on we might want user-configurable extra log fields for things like `http.request.uuid` or whatever.

mohammed90:
> I decided to piggy-back off this PR for https://github.com/caddyserver/caddy/issues/5336.
> 
> I don't love this though. This is kinda going into #4649 territory, we're doing allocations for the logging. We could use a `sync.Pool` for `ExtraLogFields` but idk 🤔 pooling isn't perfect, long-running requests/connections like websockets will hold onto them for a long time etc. We can throw away my commits if the approach doesn't make sense.
> 
> I probably want to move `user_id` to only log if non-empty (as an extra field essentially), and later on we might want user-configurable extra log fields for things like `http.request.uuid` or whatever.

I think your commits have lost adding the trace ID as you added support for adding extra fields. Now there's no reference to adding trace ID except in the tests.

mholt (approved):
Thanks for taking care on this one, to ensure the proper perspective on the variables is captured (since passing arguments into a `defer`'ed function evaluates the arguments right away, as opposed to later).

Also the benchmarks are really appreciated :grin: 

And the refactor + tests too!

The very minor performance penalty is only when tracing is enabled, yeah?

Either way, LGTM. Happy to include this in the 2.7 beta. Thank you!

modules/caddyhttp/server.go
```diff
@@ -703,6 +664,61 @@ func (s *Server) shouldLogRequest(r *http.Request) bool {
...
+
+	// preallocate with the amount of known fields
+	fieldCount := 6
+	fields := make([]zapcore.Field, fieldCount+len(extra.fields))
+	fields[0] = zap.Int("bytes_read", reqBodyLength)
+	fields[1] = zap.String("user_id", userID)
+	fields[2] = zap.Duration("duration", *duration)
+	fields[3] = zap.Int("size", wrec.Size())
+	fields[4] = zap.Int("status", wrec.Status())
+	fields[5] = zap.Object("resp_headers", LoggableHTTPHeader{
```
mholt, line 709: This could probably become a direct assignment:

```go
fields := []zapcore.Field{field1, field2, ...}
```

I think that has the same result and nicer ergonomics.

mholt: Oh, I see why it's done this way. How about:

```
fields := make([]zapcore.Field, 0, fieldCount+len(extra.fields))
```

and then use append -- is that equally good? Might be simpler code.

francislavoie: Got it :+1:

mholt (approved):

PR #5509, opened 2023-04-25
caddyhttp: Impl `ResponseWriter.Unwrap()`, prep for Go 1.20's `ResponseController`

dunglas:
[Go 1.20 added a new `"net/http".ResponseController` type](https://tip.golang.org/doc/go1.20#http_responsecontroller) that allows, among other things, to set per-request timeouts.

This patch adds support for this new type. This will improve the stability of Mercure, and simplify its code: https://github.com/dunglas/mercure/pull/764

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -66,3 +66,9 @@ func (d *delegator) WriteHeader(code int) {
 	d.status = code
 	d.ResponseWriter.WriteHeader(code)
 }
+
+// Unwrap returns the underlying ResponseWriter, necessary for
+// http.ResponseController to work correctly.
+func (d *delegator) Unwrap() http.ResponseWriter {
+	return d.ResponseWriter
+}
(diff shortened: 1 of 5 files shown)


francislavoie:
I made this change in https://github.com/caddyserver/caddy/compare/response-controller but that branch can't be merged until we're at Go 1.20 minimum. :+1: from me though.

dunglas:
Good catch @francislavoie! I cherry-picked part of your patch and now it works as excepted with Mercure, even when using compression.

The only thing I've not tested yet is HTTP/3 support. I don't manage to connect using HTTP/3 using Chrome or Firefox yet.

francislavoie (approved):

dunglas:
Tested with HTTP/3 and... it doesn't work. So I patched `quic-go` and with https://github.com/quic-go/quic-go/pull/3790 everything works as expected!

This patch is ready to be merged (no need to wait for the `quic-go` patch, an error will be returned when using HTTP/3 in the meantime).

mholt (approved):
Awesome, thanks for spearheading this @francislavoie and @dunglas -- merging!

mholt:
Forgot to click Merge :man_facepalming: 

Thanks Francis!

Issue #4849, opened 2022-06-21
Outdated TLS certificate on HTTP/3 requests after renewal


lutoma:
Caddy version: `v2.5.1 h1:bAWwslD1jNeCzDa+jDCNwb8M3UJ2tPa8UZFFzPVmGKs=`

We're using a dynamically generated JSON config that is changed from time to time using the API. It's all reverse proxies and redirects, and has one single server defined with `experimental_http3` set to true.

This has worked fine for us for the last months, the initial ACME certificate gets served correctly over both HTTP/1/2/3 etc. However today we noticed some seemingly random certificate expiration issues. After some debugging it turns out that for HTTP 1/2, Caddy is returning a new certificate that was renewed last month, whereas for HTTP/3 it is still returning the original ACME certificate from 3 months ago that has now expired.  

```
$ curl3 https://webmail.fnoco.eu -vvI
*   Trying 195.192.132.135:443...
* Connected to webmail.fnoco.eu (195.192.132.135) port 443 (#0)
[...]
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=webmail.fnoco.eu
*  start date: May 18 18:15:32 2022 GMT
*  expire date: Aug 16 18:15:31 2022 GMT
*  subjectAltName: host "webmail.fnoco.eu" matched cert's "webmail.fnoco.eu"
*  issuer: C=US; O=Let's Encrypt; CN=R3
*  SSL certificate verify ok.
* Using HTTP2, server supports multiplexing
``` 

HTTP/3 requests still got served the old certificate that expired today:

```
$ curl3 https://webmail.fnoco.eu -vvI --http3
*   Trying 195.192.132.135:443...
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
* Connect socket 5 over QUIC to 195.192.132.135:443
* Sent QUIC client Initial, ALPN: h3,h3-29,h3-28,h3-27
* SSL certificate problem: certificate has expired
* connect to 195.192.132.135 port 443 failed: SSL peer certificate or SSH remote key was not OK
* Failed to connect to webmail.fnoco.eu port 443 after 177 ms: SSL peer certificate or SSH remote key was not OK
* Closing connection 0
curl: (60) SSL certificate problem: certificate has expired
```

(I couldn't find a way to get curl to spit out the start/expire dates of the expired certificate, but from some testing with Firefox I know they're 19 March 2022 / 17 June 2022, respectively).

[...]

PR #5517, opened 2023-05-04
fix http3 outdated certificates after config reload

WeidiDeng:
Caddy http3 will serve outdated certificates is config is reloaded because caddy tries to reuse existing quic listeners. However, because after config reload, old tls.Config won't update its certificates anymore. We could disable quic listener reuse, but by changing `GetConfigForClient` field, quic listener will use latest certificates anyway.

Fix [https://github.com/caddyserver/caddy/issues/4849](4849).


listeners.go
```diff
@@ -453,12 +453,18 @@ func ListenQUIC(ln net.PacketConn, tlsConf *tls.Config, activeRequests *int64) (
...
+		return &sharedQuicListener{EarlyListener: earlyLn, tlsConf: tlsConf, key: lnKey}, nil
 	})
 	if err != nil {
 		return nil, err
 	}
 
+	sql := sharedEarlyListener.(*sharedQuicListener)
+	// update GetConfigForClient field because both quic and caddy tls uses this field
+	if sql.tlsConf != tlsConf {
+		sql.tlsConf.GetConfigForClient = tlsConf.GetConfigForClient
```
mholt, line 465: Huh -- interesting! Is this thread-safe? (Run with race detector)

mholt (commented):
I'm intrigued -- thanks for the patch!! Just one question; we need to make sure this is thread-safe.

listeners.go
WeidiDeng: Is `XCADDY_RACE_DETECTOR=1` when building enough?

Also this approach suffers from config rollback. I prefer just creating new quic listeners.

francislavoie: Re-creating the listeners means downtime, doesn't it? Reloads are meant to be graceful, that's why listeners are persisted across config reloads.

WeidiDeng: I don't know the details of quic listeners. They share the underlying udp sockets from the first glance.

mholt: @WeidiDeng 

> Is XCADDY_RACE_DETECTOR=1 when building enough?

Should be -- just be sure the relevant code path gets run with that variable set. (You should notice the program be slower when the race detector is enabled.)

WeidiDeng: I'm thinking about creating a new struct to delegate `GetConfigForClient` to. It will hopefully deal with config rollback as @francislavoie  mentioned in slack.

mholt: I guess I don't actually understand why this is the problem:

> because caddy tries to reuse existing quic listeners.

We reuse the listeners for normal HTTP too, and those work fine. What exactly about the code is causing the old certificates to persist even after they are renewed?

WeidiDeng: The problem is where the tls.Config is from `tls app`. Because of listener pool, shared quic listener is used between config reload. However, the old tls.Config is still in use, whose GetCertificate is still used by quic early listener.

Normal http don't bind tls.Config to the lifetime of the listener. I believe every time tls.Newlistener is called in the `http start` method.

WeidiDeng: I'm currently working on it. I'm wondering if it's possible for ListenQuic to be called using different tls.Config. Normally it's called by starting and reloading caddy. At most there should only be 2 instances of tls.Config for a single quic listener?

mholt: (Oh, duh, now I see that you already mentioned this same solution above. :man_facepalming: )

> At most there should only be 2 instances of tls.Config for a single quic listener?

Wouldn't there only be 1 per listener? (Sorry, not sure I understand.)

WeidiDeng: I mean different call paths. There is only one tls.Confif per listener, but underlying GetConfigForClient can be more than 1. I want to make sure that theoretically, underlying GetConfig can at most be 2, one from old config, one from new config.

francislavoie: Yeah at most two _per listen address_. Users might use quic on other ports than 443

listeners.go
```diff
@@ -485,10 +491,73 @@ func ListenerUsage(network, addr string) int {
...
+
+// sharedQuicTlsConfig manages GetConfigForClient
+type sharedQuicTlsConfig struct {
+	rmu           sync.RWMutex
+	tlsConfs      map[*tls.Config]contextAndCancelFunc
+	activeTlsConf *tls.Config
+}
+
+// newSqtc creates a new sharedQuicTlsConfig
+func newSqtc(tlsConfig *tls.Config) *sharedQuicTlsConfig {
```
francislavoie, line 508: I think abbreviated names are okay for local variables, but for functions it's best to use the full name. `newSharedQuicTlsConfig`. Arguably it should probably be `sharedQUICTLSConfig` but that's a lot of uppercase letters in a row 😅 see https://google.github.io/styleguide/go/decisions.html#initialisms

listeners.go
```diff
@@ -485,10 +491,73 @@ func ListenerUsage(network, addr string) int {
...
+		Log().Warn("quic listener tls configs are more than 2", zap.Int("number of configs", len(sqtc.tlsConfs)))
+	}
+	return ctx, wrappedCancel
+}
+
 // sharedQuicListener is like sharedListener, but for quic.EarlyListeners.
 type sharedQuicListener struct {
 	quic.EarlyListener
-	key string
+	sqtc *sharedQuicTlsConfig
```
francislavoie, line 559: I think this should use the long form as well. Or maybe just simply `config` or `sharedConfig`.

francislavoie (commented):
I think Matt is better equipped to review the logic, but this is just some style nits

listeners.go
mholt: Agreed -- let's spell this out, and I prefer `sharedQUICTLSConfig` -- it's OK. :)

listeners.go
```diff
@@ -485,10 +491,73 @@ func ListenerUsage(network, addr string) int {
...
+	return sqtc
+}
+
+func (sqtc *sharedQuicTlsConfig) getConfigForClient(ch *tls.ClientHelloInfo) (*tls.Config, error) {
+	sqtc.rmu.RLock()
+	defer sqtc.rmu.RUnlock()
+	return sqtc.activeTlsConf.GetConfigForClient(ch)
+}
+
+func (sqtc *sharedQuicTlsConfig) addTlsConfig(tlsConfig *tls.Config) (context.Context, context.CancelFunc) {
```
mholt, line 523:
```suggestion
func (sqtc *sharedQuicTlsConfig) addTLSConfig(tlsConfig *tls.Config) (context.Context, context.CancelFunc) {
```

listeners.go
```diff
@@ -485,10 +491,73 @@ func ListenerUsage(network, addr string) int {
...
 }
 
+// contextAndCancelFunc groups context and its cancelFunc
+type contextAndCancelFunc struct {
+	context.Context
+	context.CancelFunc
+}
+
+// sharedQuicTlsConfig manages GetConfigForClient
+type sharedQuicTlsConfig struct {
```
mholt, line 501:
```suggestion
type sharedQUICTLSConfig struct {
```

mholt (changes requested):
This is looking pretty good -- does it work alright @WeidiDeng ?

Thanks for working on this!

WeidiDeng:
> This is looking pretty good -- does it work alright @WeidiDeng ?
> 
> Thanks for working on this!

It works fine when new config is ok. I haven't been able to test how it will react when new config load failed (i.e. when http3 is starting ok but other parts fail), but theoretically it will use the older tls.Config in this case.

listeners.go
WeidiDeng: Nah, `fakeCloseQuicListener` uses `uc` as a field name.

mholt (approved):
Thanks for the updates. I'm willing to give this a try in the 2.7 beta :+1:

PR #5521, opened 2023-05-06
feature: watch include directory

### LINKED ISSUE #3517: Watch directory
Using [this `import` example](https://caddyserver.com/docs/caddyfile/directives/import#examples), loading a whole directory of config files, it would be nice if the --watch flag could monitor an entire directory as well. Being able to glob just like the import directive would be awesome as well.

jonatan5524:
Use the LoadConfig to load the config and compare it in each interaction of watch and larger the interaction to 3 seconds.

Resolve #3517

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -104,7 +104,7 @@ using 'caddy run' instead to keep it in the foreground.
 			cmd.Flags().StringP("config", "c", "", "Configuration file")
 			cmd.Flags().StringP("adapter", "a", "", "Name of config adapter to apply")
 			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
-			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically, check every 1s for a change")
+			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically")
 			cmd.Flags().StringP("pidfile", "", "", "Path of file to which to write process ID")
 			cmd.RunE = WrapCommandFuncForCobra(cmdStart)
 		},
@@ -143,7 +143,7 @@ The --resume flag will override the --config flag if there is a config auto-
 save file. It is not an error if --resume is used and no autosave file exists.
 
 If --watch is specified, the config file will be loaded automatically after
-changes (check every 1 second). ⚠️ This can make unintentional config changes easier; only use this
+changes. ⚠️ This can make unintentional config changes easier; only use this
 option in a local development environment.
 `,
 		CobraFunc: func(cmd *cobra.Command) {
@@ -152,7 +152,7 @@ option in a local development environment.
 			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
 			cmd.Flags().BoolP("environ", "e", false, "Print environment")
 			cmd.Flags().BoolP("resume", "r", false, "Use saved config, if any (and prefer over --config file)")
-			cmd.Flags().BoolP("watch", "w", false, "Watch config file for changes and reload it automatically, check every 1s for a change")
+			cmd.Flags().BoolP("watch", "w", false, "Watch config file for changes and reload it automatically")
 			cmd.Flags().StringP("pidfile", "", "", "Path of file to which to write process ID")
 			cmd.Flags().StringP("pingback", "", "", "Echo confirmation bytes to this address on success")
 			cmd.RunE = WrapCommandFuncForCobra(cmdRun)
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -104,7 +104,7 @@ using 'caddy run' instead to keep it in the foreground.
 			cmd.Flags().StringP("config", "c", "", "Configuration file")
 			cmd.Flags().StringP("adapter", "a", "", "Name of config adapter to apply")
 			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
-			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically")
+			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically, check every 1s for a change")
 			cmd.Flags().StringP("pidfile", "", "", "Path of file to which to write process ID")
 			cmd.RunE = WrapCommandFuncForCobra(cmdStart)
 		},
@@ -143,7 +143,7 @@ The --resume flag will override the --config flag if there is a config auto-
 save file. It is not an error if --resume is used and no autosave file exists.
 
 If --watch is specified, the config file will be loaded automatically after
-changes. ⚠️ This can make unintentional config changes easier; only use this
+changes (check every 1 second). ⚠️ This can make unintentional config changes easier; only use this
 option in a local development environment.
 `,
 		CobraFunc: func(cmd *cobra.Command) {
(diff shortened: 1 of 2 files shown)


mholt (commented):
This is an interesting approach, so instead of watching individual files' mod times, you're just relying on `LoadConfig()` to imply when one of the included config files has changed by comparing the bytes?

That's not a bad idea.

jonatan5524:
> This is an interesting approach, so instead of watching individual files' mod times, you're just relying on `LoadConfig()` to imply when one of the included config files has changed by comparing the bytes? And it uses less code in the process?
> 
> That's not a bad idea.

Yes! Thank you!
 I saw the LoadConfig is combining the main config file with it's included, if anything changes in any file we detect it. 
I tested it with simple config files, Do you think it will be much slower with bigger config files?

cmd/main.go
```diff
@@ -189,64 +189,37 @@ func watchConfigFile(filename, adapterName string) {
...
-		}
-		return false
-	}
-
 	// begin poller
 	//nolint:staticcheck
-	for range time.Tick(1 * time.Second) {
-		// get the file info
-		info, err := os.Stat(filename)
+	for range time.Tick(3 * time.Second) {
```
mholt, line 203: Is this to alleviate CPU and I/O load for large configs?

I wonder if 1-2s is still fast enough for most people.

jonatan5524: I wonder this myself, Do you have maybe a large config file to test this approach?

mholt:
I mean, I have a fast CPU and SSD. I would never use the watch feature in production (but I know some people do). I don't know how busy those users' production servers are.

In local development, you probably want a fast refresh time like 1s. In production, 5s is probably just fine.

I'm willing to bet this is used locally more than in prod (but without telemetry it's impossible to be sure, sigh) -- so maybe we keep it 1s and then wait for complaints?

jonatan5524:
> I mean, I have a fast CPU and SSD. I would never use the watch feature in production (but I know some people do). I don't know how busy those users' production servers are.
> 
> In local development, you probably want a fast refresh time like 1s. In production, 5s is probably just fine.
> 
> I'm willing to bet this is used locally more than in prod (but without telemetry it's impossible to be sure, sigh) -- so maybe we keep it 1s and then wait for complaints?

Aha agree 😉

mholt:
Alrighty. Well let's change the interval back to 1s and then we can merge it, give this a try :) Thanks!

jonatan5524:
> I think we should adjust the `caddy run` CLI docs to mention in more detail how `--watch` works, i.e. to mention it polls every second, loading & parsing the config to compare if it has changed since the last load.

But do you really think the user that sees the cli "help" command cares how the watch is implemented inside? maybe mention the 1s check in the docs and how it is implemented before the `watchConfigFile` method?

In start and in run commands:
```go
cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically, check every 1s for a change")
```

In run long:
```go
If --watch is specified, the config file will be loaded automatically after
changes (check every 1 second). ⚠️ This can make unintentional config changes easier; only use this
option in a local development environment.
```

And before the `watchConfigFile` function:
// watchConfigFile watches the config file at filename for changes
// and reloads the config if the file was updated. This function
// blocks indefinitely; it only quits if the poller has errors for
// long enough time. The filename passed in must be the actual
// config file used, not one to be discovered. 
// Each second the config files is loaded and parsed into an object 
// and is compared to the last config object that was loaded
func watchConfigFile(filename, adapterName string) {

mholt:
Personally, I think with a quick refresh rate like this, the implementation is irrelevant to the user and I don't think we need to expose this. If it was a longer refresh (maybe if it was deemed expensive), then perhaps we should document that so the user doesn't wonder why their config didn't apply until several seconds later.

However, if we still feel like exposing that information is a good idea, perhaps it can go in a sentence on our docs website.

@jonatan5524 This is looking great, thanks.

I had the idea chatting about this with Francis in our dev Slack that maybe the refresh rate could be determined by the size of the last config. If we assume that configs won't (frequently) change in size by an order of magnitude, maybe the bigger a config is, the longer the refresh rate is.

But if we do decide to do that, it could go in a subsequent PR. My vote is let's try the 1 second and see if that is too fast/slow for anyone in a real use case first.

francislavoie:
CI is failing for some weird reason. I think the caches are busted. I'm gonna try to fix that... working on it.

cmd/main.go
```diff
@@ -173,7 +173,9 @@ func LoadConfig(configFile, adapterName string) ([]byte, string, error) {
 // and reloads the config if the file was updated. This function
 // blocks indefinitely; it only quits if the poller has errors for
 // long enough time. The filename passed in must be the actual
-// config file used, not one to be discovered.
+// config file used, not one to be discovered.
```
francislavoie, line 176: There's a trailing space here, linter doesn't like it

cmd/commands.go
```diff
@@ -152,7 +152,7 @@ option in a local development environment.
 			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
 			cmd.Flags().BoolP("environ", "e", false, "Print environment")
 			cmd.Flags().BoolP("resume", "r", false, "Use saved config, if any (and prefer over --config file)")
-			cmd.Flags().BoolP("watch", "w", false, "Watch config file for changes and reload it automatically")
+			cmd.Flags().BoolP("watch", "w", false, "Watch config file for changes and reload it automatically, check every 1s for a change")
```
mholt, line 155:
```suggestion
			cmd.Flags().BoolP("watch", "w", false, "Watch config file for changes and reload it automatically")
```

cmd/commands.go
```diff
@@ -143,7 +143,7 @@ The --resume flag will override the --config flag if there is a config auto-
 save file. It is not an error if --resume is used and no autosave file exists.
 
 If --watch is specified, the config file will be loaded automatically after
-changes. ⚠️ This can make unintentional config changes easier; only use this
+changes (check every 1 second). ⚠️ This can make unintentional config changes easier; only use this
```
mholt, line 146:
```suggestion
changes. ⚠️ This can make unintentional config changes easier; only use this
```

cmd/commands.go
```diff
@@ -104,7 +104,7 @@ using 'caddy run' instead to keep it in the foreground.
 			cmd.Flags().StringP("config", "c", "", "Configuration file")
 			cmd.Flags().StringP("adapter", "a", "", "Name of config adapter to apply")
 			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
-			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically")
+			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically, check every 1s for a change")
```
mholt, line 107:
```suggestion
			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically")
```

mholt (changes requested):
Actually I'm gonna just take out the implementation details from the CLI docs since I don't want to be bound to them, in case it changes, and I'd rather users don't rely on it by accident.

I think this is ready then!

mholt (approved):
It's perfect now :100: -- let's try it out, and we can tune/optimize later, as discussed. (For example, auto-tuning the interval based on config size)

PR #5527, opened 2023-05-10
Initial implementation of fallback_sni

mholt:
This probably won't work (at least, not in all cases) -- but let's see!

This is an extremely niche feature but there's a legitimate use that I couldn't find a better way to do it.


francislavoie:
For ref, feature was requested here https://caddy.community/t/default-certificate-when-no-matching-sni/19861

vcabbage:
I tested this out and it does work. Thanks!

However, initially it didn't because it wasn't loading the managed certs at all. That is, even removing the SNI fallback logic it wasn't loading any certificates.

I bisected the problem to https://github.com/caddyserver/caddy/commit/774f2288682c117763a42de17b592200469ebaac and https://github.com/caddyserver/certmagic/commit/6670b5e1cc545256468bb8f0e682770333349010.

I'm not familiar with the codebase, but the section below looked like the only change that would have affected cert loading. Commenting it out "fixed" (I'm unsure if this is a fix or a hack) the issue.

```go
if len(cfg.Managers) > 0 {
	continue
}
```

https://github.com/caddyserver/certmagic/blob/dec254055d16265c58c6283b8f24cdf36981928e/config.go#L361-L363

mholt:
Ah, dangit, that was a change from another project that snuck in, I must've done my branching messy. One sec, will fix.

mholt:
@vcabbage Okie dokie, this should be fixed on `master`. But since you already know this fixes it for you, no need to try it again. It will go out with the next Caddy release, so stay tuned for 2.7 beta 1!

mholt:
Grr, the CI jobs keep failing, but there's no logs (it just says "failed" with an X) and it's sporadic too.

Issue #5313, opened 2023-01-17
Storage export/import CLI commands


francislavoie:
In Caddy, the [storage](https://caddyserver.com/docs/json/storage/) can be configured with any module implementing the [`certmagic.Storage`](https://pkg.go.dev/github.com/caddyserver/certmagic#Storage) interface.

Currently, we don't have any mechanism in place for exporting storage for backup, or an import mechanism to migrate from one storage module to another, or from one server/deployment to another.

What we should do is add two new commands to Caddy, i.e. `caddy storage export` and `caddy storage import` (exact command name and flags TBD) to make this possible.

Details:

- The commands should take the `--config` as input to load the storage module.
- The commands should _not_ try to contact the admin API. We shouldn't add an admin API for accessing the storage, because that adds additional risk to leaking private keys and such.
- It would call `List()` on the storage module to get all the keys, then `Load()` them all.
- Exporting would produce a `.tar.gz` or similar, exact archive format TBD.
- It should be possible to pipe the output to a file, or to `stdout` (useful with Docker for example to write it out to a file on the host machine).
- Importing should take the same archive, either from file or from `stdin`.
- In the case of running as a systemd service, the command would need to be run as the `caddy` user to correctly access the files owned by that user, if the `file_system` storage module is used.

g41797:
Hi 
I'd like to participate or discuss

francislavoie:
Sure, PRs are always welcome. Feel free to ask questions.

g41797:
- Is it really gfi?
- Timing/deadline - I am in the middle of https://github.com/gavv/httpexpect/issues/246
- Target - as far as I understand it's linux but in yml files i saw mac/windows


Also i tried to use link above ( https://caddyserver.com/docs/json/storage/) in order to understand meaning of 'storage' but it looks broken

And off the topic question - usage of Server-Sent Events by caddy applications

francislavoie:
I don't know what you mean by "gfi".

No specific deadline. But don't overextend yourself.

Yes, Caddy works on every platform.

Yeah unfortunately our JSON config docs are a bit broken right now, our database is being flaky as of a couple days ago. We're working on it. I think if you wait 2 minutes it might return the docs 🤷‍♂️ but the docs are generated from code, so you can just look at `storage.go` to get an idea. The actual interfaces and the filesystem storage implementation is in `certmagic`, a separate repo.

[...]

PR #5532, opened 2023-05-11
cmd: Implement storage import/export

Phrynobatrachus:
#5313 

Thanks for the fun project!

1)
> I think I'd probably want to put that in the caddycmd package though (the cmd folder), as it's purely a CLI utility, and not really anything to do with Caddy core. (Assuming nothing new needs to be exported. If so let me know.)

The two functions in `caddy.go` that is? There are caddycmd wrappers separately as is.

2)
> In the case of running as a systemd service, the command would need to be run as the caddy user to correctly access the files owned by that user, if the file_system storage module is used.

I started by shelling out `systemctl status caddy` for part of the rerun check but relying only on the `--user` flag felt a little less awkward. I'm not sure what the best option is though. Besides file permissions, it seemed necessary to re-exec with the corresponding alternate user' `$HOME` set, for AppDataDir() -> homeDirUnsafe(), so the user has to be specified anyways to get that.

- Windows handling (AppData vs HOME), etc?

- Add `CAP_SETUID` / `CAP_SETGID` to `setcap.sh`?

- fork/exec error on the user switch without `NoSetGroups: true` https://walac.github.io/golang-patch/

3)
`run()` vs `Run()`

```
// This is a low-level function; most callers
// will want to use Run instead, which also
// updates the config's raw state.
```
run() makes sense to me for exporting, maybe importing should Run() instead?

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -306,6 +306,56 @@ the KEY=VALUE format will be loaded into the Caddy process.
 	})
 
 	RegisterCommand(Command{
+		Name:  "storage",
+		Short: "Commands for working with Caddy's storage (EXPERIMENTAL)",
+		Long: `
+Allows exporting and importing Caddy's storage contents. The two commands can be
+combined in a pipeline to transfer directly from one storage to another:
+
+$ caddy storage export --config Caddyfile.old --output - |
+> caddy storage import --config Caddyfile.new --input -
+
+The - argument refers to stdout and stdin, respectively.
+
+NOTE: When importing to or exporting from file_system storage (the default), the command
+should be run as the user that owns the associated root path.
+
+EXPERIMENTAL: May be changed or removed.
+`,
+		CobraFunc: func(cmd *cobra.Command) {
+			exportCmd := &cobra.Command{
+				Use:   "export --config <path> --output <path>",
+				Short: "Exports storage assets as a tarball",
+				Long: `
+The contents of the configured storage module (TLS certificates, etc)
+are exported via a tarball.
+
+--output is required, - can be given for stdout.
+`,
+				RunE: WrapCommandFuncForCobra(cmdExportStorage),
+			}
+			exportCmd.Flags().StringP("config", "c", "", "Input configuration file (required)")
+			exportCmd.Flags().StringP("output", "o", "", "Output path")
+			cmd.AddCommand(exportCmd)
+
+			importCmd := &cobra.Command{
+				Use:   "import --config <path> --input <path>",
+				Short: "Imports storage assets from a tarball.",
+				Long: `
+Imports storage assets to the configured storage module. The import file must be
+a tar archive.
+
+--input is required, - can be given for stdin.
+`,
+				RunE: WrapCommandFuncForCobra(cmdImportStorage),
+			}
+			importCmd.Flags().StringP("config", "c", "", "Configuration file to load (required)")
+			importCmd.Flags().StringP("input", "i", "", "Tar of assets to load (required)")
+			cmd.AddCommand(importCmd)
+		},
+	})
+
+	RegisterCommand(Command{
 		Name:  "fmt",
 		Usage: "[--overwrite] [--diff] [<path>]",
 		Short: "Formats a Caddyfile",
--- a/cmd/storagefuncs.go
+++ b/cmd/storagefuncs.go
@@ -1,0 +1,228 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package caddycmd
+
+import (
+	"archive/tar"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/certmagic"
+)
+
+type storVal struct {
+	StorageRaw json.RawMessage `json:"storage,omitempty" caddy:"namespace=caddy.storage inline_key=module"`
+}
+
+// determineStorage returns the top-level storage module from the given config.
+// It may return nil even if no error.
+func determineStorage(configFile string, configAdapter string) (*storVal, error) {
+	cfg, _, err := LoadConfig(configFile, configAdapter)
+	if err != nil {
+		return nil, err
+	}
+
+	// storage defaults to FileStorage if not explicitly
+	// defined in the config, so the config can be valid
+	// json but unmarshaling will fail.
+	if !json.Valid(cfg) {
+		return nil, &json.SyntaxError{}
+	}
+	var tmpStruct storVal
+	err = json.Unmarshal(cfg, &tmpStruct)
+	if err != nil {
+		// default case, ignore the error
+		var jsonError *json.SyntaxError
+		if errors.As(err, &jsonError) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	return &tmpStruct, nil
+}
+
+func cmdImportStorage(fl Flags) (int, error) {
+	importStorageCmdConfigFlag := fl.String("config")
+	importStorageCmdImportFile := fl.String("input")
+
+	var err error
+	if importStorageCmdConfigFlag == "" {
+		err = errors.New("--config is required")
+	}
+	if importStorageCmdImportFile == "" {
+		err = errors.Join(err, errors.New("--input is required"))
+	}
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	// extract storage from config if possible
+	storageCfg, err := determineStorage(importStorageCmdConfigFlag, "")
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	// load specified storage or fallback to default
+	var stor certmagic.Storage
+	ctx, cancel := caddy.NewContext(caddy.Context{Context: context.Background()})
+	defer cancel()
+	if storageCfg != nil && storageCfg.StorageRaw != nil {
+		val, err := ctx.LoadModule(storageCfg, "StorageRaw")
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, err
+		}
+		stor, err = val.(caddy.StorageConverter).CertMagicStorage()
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, err
+		}
+	} else {
+		stor = caddy.DefaultStorage
+	}
+
+	// setup input
+	var f *os.File
+	if importStorageCmdImportFile == "-" {
+		f = os.Stdin
+	} else {
+		f, err = os.Open(importStorageCmdImportFile)
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, fmt.Errorf("opening input file: %v", err)
+		}
+		defer f.Close()
+	}
+
+	// store each archive element
+	tr := tar.NewReader(f)
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return caddy.ExitCodeFailedQuit, fmt.Errorf("reading archive: %v", err)
+		}
+
+		b, err := io.ReadAll(tr)
+		if err != nil {
+			return caddy.ExitCodeFailedQuit, fmt.Errorf("reading archive: %v", err)
+		}
+
+		err = stor.Store(ctx, hdr.Name, b)
+		if err != nil {
+			return caddy.ExitCodeFailedQuit, fmt.Errorf("reading archive: %v", err)
+		}
+	}
+
+	fmt.Println("Successfully imported storage")
+	return caddy.ExitCodeSuccess, nil
+}
+
+func cmdExportStorage(fl Flags) (int, error) {
+	exportStorageCmdConfigFlag := fl.String("config")
+	exportStorageCmdOutputFlag := fl.String("output")
+
+	var err error
+	if exportStorageCmdConfigFlag == "" {
+		err = errors.New("--config is required")
+	}
+	if exportStorageCmdOutputFlag == "" {
+		err = errors.Join(err, errors.New("--output is required"))
+	}
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	// extract storage from config if possible
+	storageCfg, err := determineStorage(exportStorageCmdConfigFlag, "")
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	// load specified storage or fallback to default
+	var stor certmagic.Storage
+	ctx, cancel := caddy.NewContext(caddy.Context{Context: context.Background()})
+	defer cancel()
+	if storageCfg != nil && storageCfg.StorageRaw != nil {
+		val, err := ctx.LoadModule(storageCfg, "StorageRaw")
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, err
+		}
+		stor, err = val.(caddy.StorageConverter).CertMagicStorage()
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, err
+		}
+	} else {
+		stor = caddy.DefaultStorage
+	}
+
+	// enumerate all keys
+	keys, err := stor.List(ctx, "", true)
+	if err != nil {
+		return caddy.ExitCodeFailedStartup, err
+	}
+
+	// setup output
+	var f *os.File
+	if exportStorageCmdOutputFlag == "-" {
+		f = os.Stdout
+	} else {
+		f, err = os.Create(exportStorageCmdOutputFlag)
+		if err != nil {
+			return caddy.ExitCodeFailedStartup, fmt.Errorf("opening output file: %v", err)
+		}
+		defer f.Close()
+	}
+
+	// `IsTerminal: true` keys hold the values we
+	// care about, write them out
+	tw := tar.NewWriter(f)
+	for _, k := range keys {
+		info, err := stor.Stat(ctx, k)
+		if err != nil {
+			return caddy.ExitCodeFailedQuit, err
+		}
+
+		if info.IsTerminal {
+			v, err := stor.Load(ctx, k)
+			if err != nil {
+				return caddy.ExitCodeFailedQuit, err
+			}
+
+			hdr := &tar.Header{
+				Name: k,
+				Mode: 0600,
+				Size: int64(len(v)),
+			}
+
+			if err = tw.WriteHeader(hdr); err != nil {
+				return caddy.ExitCodeFailedQuit, fmt.Errorf("writing archive: %v", err)
+			}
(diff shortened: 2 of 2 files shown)


cmd/commands.go
```diff
@@ -305,6 +305,50 @@ the KEY=VALUE format will be loaded into the Caddy process.
 		},
 	})
 
+  RegisterCommand(Command{
+		Name:  "storage-export",
```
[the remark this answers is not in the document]
francislavoie: :thinking: yeah either one is fine. I thought it would be fun to set up a subcommand. I might take a look at it in the coming days, I'm curious to see how it would work.

caddy.go
```diff
@@ -670,6 +671,85 @@ func Validate(cfg *Config) error {
 	return err
 }
 
+// ExportStorage calls List() and Load() on the configured
+// storage module and returns the underlying assets. 
+func ExportStorage(cfg *Config) (files []struct{
+	Name string
+	Body []byte
+}, err error) {
+	ctx, err := run(cfg, false)
```
mholt, line 680: Rather than running the config, maybe better to load and unmarshal just the `storage` portion. You could try doing something like this:

https://github.com/caddyserver/caddy/blob/cfc85ae8caf032c3ef92675a024fc5d1264ab7ce/cmd/commandfuncs.go#L693-L717

You might need to make a caddy.Context and then call LoadModule() on it though -- I haven't tried doing that in isolation like this but I'd be happy to help if that is problematic...

caddy.go
```diff
@@ -670,6 +671,85 @@ func Validate(cfg *Config) error {
 	return err
 }
 
+// ExportStorage calls List() and Load() on the configured
+// storage module and returns the underlying assets. 
+func ExportStorage(cfg *Config) (files []struct{
```
mholt, line 676: I'm hoping we can find a way to put these functions -- unexported -- in the `caddycmd` package.

caddy.go
```diff
@@ -670,6 +671,85 @@ func Validate(cfg *Config) error {
...
+	Name string
+	Body []byte
```
mholt, line 678: Loading the entire storage into memory is probably not ideal -- it will blow up for large deployments (where a feature like this is most useful).

Maybe exporting should write directly [to an archive](https://github.com/mholt/archiver).

Although, we're getting into separate tooling territory.

What if, instead, we had a `transfer storage` utility that simply moved the contents of one storage to another storage? We could do it kv-pair by kv-pair, much more efficiently, directly, without any need to read everything into memory or have separate import/export ops.

mholt (changes requested):
Thanks for this implementation!

While reviewing I thought of something that might make more practical sense... basically a `transfer-storage` utility that does both export and import at the same time, in lock-step, to avoid blowing up memory and also for simplicity's sake.

We'd just need to figure out how to configure both storage modules.

caddy.go
Phrynobatrachus: Yeah I did try it this way to easily get at a caddy.Context, I'll give that a shot.

caddy.go
Phrynobatrachus: > What if, instead, we had a transfer storage utility that simply moved the contents of one storage to another storage? We could do it kv-pair by kv-pair, much more efficiently, directly, without any need to read everything into memory or have separate import/export ops.

Yeah that's a good point, like just Load() from old -> Store() in new -> Delete() from old per each key at a time.

mholt: Yes! Errrr, but maybe make Delete() opt-in, and maybe that should only be done after the migration is 100% totally completed successfully.

francislavoie: Transfer alone might not be enough in situations where you want to move from one cluster to another where the storage is completely isolated in the old vs new. I think adding transfer is a good idea, but I don't think it should be instead of export/import, I think both should exist.

mholt: I don't think backups are in the scope of the web server. I'd like to leave that up to backup utilities.

I also would like to hear of actual use cases that require an export but not an import -- AFAIK the need is to be able to switch storage providers, migrate certs from one place to another. A plain export can be done with the storage system itself (i.e. copy the files from disk, or run a SQL query) -- importing so that Caddy can use it is harder, and it makes sense to be able to import from any data source.

The more I think about this the more I'm convinced that we need a transfer utility -- effectively an "import", but of course supporting any storage module as a source -- not an exporter.

francislavoie: Like I said, if the storage is isolated in such a way that Caddy cannot contact both locations, then transfers cannot work. I don't think we can presume that transfers are always possible. That would be a very unfortunate limitation.

And for backups, that assumes the users know how to perform backups using their storage mechanism, and that they might even care to backup anything other than Caddy's data in that storage. For example for Redis, I might be using it as a cache otherwise, so I don't care to backup anything else.

Also re memory usage of loading everything in storage, I don't think that's a relevant concern. The CLI script is what will be keeping it in memory. Each LE cert+key+json is roughly 6k in size (with default key type etc), so 1000 certs (which is quite large) is ~6MB, and a huge deployment of like 50,000 certs would be like ~300MB; on a deployment that large, they'd surely be using a machine with plenty of RAM, hopefully over-provisioned.

Either way, writing straight to a mutable archive file is a good idea but I wouldn't consider that absolutely necessary.

Phrynobatrachus: > Also re memory usage of loading everything in storage, I don't think that's a relevant concern.

fwiw this was my first impression as well -- that everything in storage is relatively small both individually and overall, and that the storage is strictly an interface for items critical to operation rather than anything served by itself ("must not be a cache..." note in the docs). I'm definitely not experienced with any large deployments though.

Phrynobatrachus: > A plain export can be done with the storage system itself (i.e. copy the files from disk, or run a SQL query) -- importing so that Caddy can use it is harder, and it makes sense to be able to import from any data source.

I agree that exporting already has some clear ways to do it. The framing that makes sense to me is that commands as originally described seem useful to help _facilitate_ transfers/backups (either manually or scripted), but actually doing those would indeed be out of scope. So it's a question of whether it's too little to justify adding rather than too much. 

> The more I think about this the more I'm convinced that we need a transfer utility -- effectively an "import", but of course supporting any storage module as a source -- not an exporter.

I don't follow quite as much with this. I don't want to muddle things too much, I guess my thought for now is any potential memory issue seems orthogonal to whatever CLI semantics are desired.

mholt: I feel like it's just bad form though -- even if the data happens to only be a few MB in most cases, we do know there are clusters out there with 100,000+ certs. It's so easy to stream the contents file-by-file, so we should probably do that. If someone comes to us complaining that there's a lot of memory use when doing an import/export (because, memory is a shared resource), then I would have no good excuse for that.

francislavoie: Which is why I'm saying it makes sense to have both import/export _and_ transfer. It's up to the user to pick which is best for them. But only having transport and not import/export isn't acceptable IMO.

cmd/commands.go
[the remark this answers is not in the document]
francislavoie: Pushed a commit, got it working :+1: was pretty simple.

Also as an aside, please make sure your editor is configured to use `.editorconfig`; your code was using spaces instead of tabs among other minor style issues. Use `gofmt -w -s <path>` to apply fixes.

mholt:
If we aren't going to make this a transfer utility, I'll revisit this after the 2.7 beta 1 release, which is my current priority. I'm still not sure exporting data from storage systems is within the scope of the web server.

francislavoie:
But only Caddy can provide a unified interface to export/import for any storage implementation, because of the interfaces. I think it's definitely in scope, when `storage` is a top-level configuration option in Caddy.

But yeah, no rush on this, it's not a blocker for 2.7.0.

cmd/commandfuncs.go
```diff
@@ -538,6 +542,183 @@ func cmdValidateConfig(fl Flags) (int, error) {
...
+				return caddy.ExitCodeFailedStartup, err
+			}
+			defer f.Close()
+		
+			cmd.Stdout = f
+		} else {
+			cmd.Stdout = os.Stdout
+		}
+		cmd.Stderr = os.Stderr
+		cmd.Env = []string{fmt.Sprint("HOME=", u.HomeDir)}
```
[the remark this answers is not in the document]
francislavoie: Rethinking this, I'm not sure we need `--user` at all. I think we should just add a mention that the command should be run as the same user Caddy runs as (typically `caddy` on Linux with systemd). Linux users can use `sudo su caddy -c 'caddy export -o <wherever>` to run it as the correct user.

On Windows, I think users typically end up running it as their current administrator user, so it shouldn't be a concern.

As you can see from the Windows CI build, it's failing because of the use of `syscall.Credential` which isn't defined on Windows. So I think we should avoid trying to muck with retrying as a different user.

Thanks for trying it though, it's an interesting experiment either way.

cmd/commandfuncs.go
```diff
@@ -538,6 +542,183 @@ func cmdValidateConfig(fl Flags) (int, error) {
 	return caddy.ExitCodeSuccess, nil
 }
 
+func cmdImportStorage(fl Flags) (int, error) {
+	importStorageCmdConfigFlag := fl.String("config")
+	importStorageCmdUserFlag := fl.String("user")
+  	importStorageCmdImportFile := fl.Arg(0)
+	
+	// re-run similarly to export here?
+	if importStorageCmdUserFlag != "" {
```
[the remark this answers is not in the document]
francislavoie: I think what might be more useful is to write an "are you sure" message showing what the current user is and the current storage path if the default storage is configured, to give users a chance to realize their mistake if they ran it incorrectly.

As an aside, I noticed that running the command without `-o` causes the tar contents to just get dumped to stdout. I don't think we should do that by default. I think we should require `-o` but allow using `-` as a value to mean write to stdout. That way users can do `caddy storage export -o- | some-other-command-that-takes-it-as-input`.

On that topic, I think users could do this to perform a transfer in a one-liner without writing to a file:

```
caddy storage export --config Caddyfile.old -o- | caddy storage import --config Caddyfile.new -i-
```

cmd/commandfuncs.go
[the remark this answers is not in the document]
Phrynobatrachus: Cool. I saw the CI failure and started looking at the Windows syscall interface (https://pkg.go.dev/golang.org/x/sys/windows#CreateProcessAsUser) but it's certainly simpler without all of that in the first place.

mholt:
An export without an import is just a backup, and that's not Caddy's job. If you want to back up your certs, run `rsync` on your Caddy data dir, or a SQL dump of your storage DB, or do the equivalent for whatever storage medium you're using. Then you can restore later using that same tooling, no need for Caddy to reinvent the wheel.

Now, an import is definitely useful. But import without source data is useless, so really what we need is an _import_ feature that can get the contents from a configured storage, which so happens to basically be a kind of export.

So, the user can still do what they need (export now, import later), it's just that they don't use Caddy until that import step.

Phrynobatrachus:
I pushed a commit that gets rid of the `caddy.go` import/export functions and loads the storage modules solely from the caddycmd package. Otherwise left the CLI setup the same for the moment. Hopefully a little closer though!

cmd/commandfuncs.go
```diff
@@ -538,6 +539,135 @@ func cmdValidateConfig(fl Flags) (int, error) {
...
+
+func DetermineStorage(configFile string, configAdapter string) (*storVal, error) {
+	cfg, _, err := LoadConfig(configFile, configAdapter)
+	if err != nil {
+		return nil, err
+	}
+
+	var tmpStruct storVal
+	err = json.Unmarshal(cfg, &tmpStruct)
+	if err != nil {
```
Phrynobatrachus, line 554: This doesn't correctly handle the default case where no storage is defined in the config. My current thought is to check `json.Valid` first so that if `Unmarshal` returns `json.SyntaxError` it should be safe to assume it's the default case.

mholt: That could probably work.

mholt (changes requested):
Getting there!

I still need to decide whether to have this as two commands (import/export) or just one (transfer).

mholt:
@Phrynobatrachus Thanks so much for continuing to work on this and for your patience. It's a high-quality contribution -- I am just dragging my feet on deciding what to do here.

We'll get it in sooner or later though :sweat_smile:

Phrynobatrachus:
Take your time! I'm here for the practice anyways, it's all good.

Phrynobatrachus:
Some notes from testing against storage implementations with the transfer invocation suggested by Francis:

`caddy storage export -c Caddyfile.old -o- | caddy storage import -c Caddyfile.new -i-`

(where old is file_system, new is alt storage)

## postgres-storage

- doesn't support recursive parameter in `List()`, which this PR uses

https://github.com/yroc92/postgres-storage/blob/276797aefe401b738781692d278a158c53b99208/storage.go#L317C2-L319

```go
      if recursive {
              return nil, fmt.Errorf("recursive not supported")
      }
```
## caddy-tlsredis

- works after a patch to `Stat()`

https://github.com/gamalan/caddy-tlsredis/blob/66e1c61412decebb977790af25f7a6ce5c9b5878/storageredis.go#L488-L493

```go
 return certmagic.KeyInfo{
     Key:        key,
     Modified:   data.Modified,
     Size:       int64(len(data.Value)),
     IsTerminal: false,   
}, nil
```
changed to `IsTerminal: true`, appropriate for all cases? 

## certmagic-s3

- works after a patch to Stat()

https://github.com/ss098/certmagic-s3/blob/627ff2e5b5a473e29cc63f15e37cbd6b1b22eeab/s3.go#L244-L249

```go
        return certmagic.KeyInfo{
                Key:        object.Key,
                Modified:   object.LastModified,
                Size:       object.Size,
                IsTerminal: strings.HasSuffix(object.Key, "/"),
        }, err
```

should be `IsTerminal: !strings.HasSuffix(object.Key, "/")`

francislavoie:
Ooh very interesting, thanks for trying that!

I don't know the answer re `IsTerminal`, Matt will need to answer that.

mholt:
@Phrynobatrachus Excellent, thanks for the field trials.

> changed to IsTerminal: true, appropriate for all cases?

Only when the stat'ed key is a "file" (i.e. not a "directory"). In other words, only when the key doesn't contain other keys within/under it. If that makes sense.

> should be IsTerminal: !strings.HasSuffix(object.Key, "/")

Probably true! Good catch.

> doesn't support recursive parameter in List(), which this PR uses

@yroc92 Can I help with discussing how to support `recursive=true` on postgres-storage?

francislavoie:
If export always essentially turns whatever storage into the layout it would have if it was the filesystem module, then I think that's pretty natural. It's not like we're introducing a new format of somekind, it's already how storage already exists, just put in a tarball.

IMO the only downside of import/export is that it's loaded in memory before writing out. But I think that's fine for a large majority of users. If a truly huge storage needs to be migrated then we can introduce `transfer`. But IMO import+export is the minimum viable product here.

Phrynobatrachus:
> Only when the stat'ed key is a "file" (i.e. not a "directory"). In other words, only when the key doesn't contain other keys within/under it. If that makes sense.

Yeah, if `!strings.HasSuffix(object.Key, "/")` is a reasonable expression of that for certmagic-s3, caddy-tlsredis can probably just do the same.

mholt:
> Yeah, if !strings.HasSuffix(object.Key, "/") is a reasonable expression of that for certmagic-s3, caddy-tlsredis can probably just do the same.

I _think_ that depends on how the storage mechanism is implemented. For example, file systems don't necessarily require a trailing `/` to indicate a directory -- whether it's a directory is stored as metadata.

Phrynobatrachus:
> IMO the only downside of import/export is that it's loaded in memory before writing out.

This was only a problem with my original version where I was unnecessarily buffering the data to export right? Latest commit using tar again for IO looks like it works, and shouldn't have that issue unless I'm mistaken (which is very possible :stuck_out_tongue:)

mholt:
Yeah, and since our current storage APIs don't support streaming, the best we can do is write each file in turn. Maybe in the future we'll be able to stream individual files too.

Thanks for the streaming enhancement! That's going to be much better.

I think we can merge it in this week. I need to do a final review.

cmd/commandfuncs.go
```diff
@@ -544,6 +544,7 @@ type storVal struct {
 	StorageRaw json.RawMessage `json:"storage,omitempty" caddy:"namespace=caddy.storage inline_key=module"`
 }
 
+// *storVal should be nil checked even if err == nil
```
mholt, line 547:
```suggestion
// determineStorage returns the top-level storage module from the given config.
// It may return nil even if no error.
```

cmd/commands.go
```diff
@@ -305,6 +305,54 @@ the KEY=VALUE format will be loaded into the Caddy process.
...
+		Name:  "storage",
+		Short: "Commands for working with Caddy's storage",
+		Long: `
+Allows exporting and importing Caddy's storage contents. The two commands can be
+combined in a pipeline to transfer directly from one storage to another:
+
+$ caddy storage export --config Caddyfile.old --output - |
+> caddy storage import --config Caddyfile.new --input -
+
+With - referring to stdout/stdin.
```
mholt, line 318:
```suggestion
The - argument refers to stdout and stdin, respectively.
```

cmd/commands.go
```diff
@@ -305,6 +305,54 @@ the KEY=VALUE format will be loaded into the Caddy process.
...
+
+NOTE: When importing to or exporting from file_system storage (the default), the command
+should be run as the user that owns the associated root path.
+`,
+		CobraFunc: func(cmd *cobra.Command) {
+			exportCmd := &cobra.Command{
+				Use:   "export --config <path> --output <path>",
+				Short: "Exports storage assets as a tarball",
+				Long: `
+The contents of the configured storage module (TLS certificates, autosave.json, etc)
```
mholt, line 328: I'm pretty sure autosave.json goes into the config directory (`~/.config/caddy` default on Linux), not storage :thinking:

mholt (changes requested):
Awesome, this is getting close!

See inline comments, plus:

- Maybe we should move the relevant code (other than the `RegisterCommand()`) into a file called `cmd/storage.go` since there's quite a few functions that are needed to make it work, we can keep them together.
- We should add a note to the CLI docs stating that this is experimental and subject to change, maybe: `EXPERIMENTAL: This feature is subject to change or removal.` -- I don't think we'll remove it, but I want to set clear expectations until we iron out any issues later. :)

Thanks for working on this!

cmd/commands.go
francislavoie: I'd put single quotes around it like `'-'` maybe

cmd/commands.go
francislavoie: Correct, config storage is not the same as data storage. Explained here https://caddyserver.com/docs/conventions#file-locations

Phrynobatrachus: Thanks! I thought I saw it in exports at one point, but I'm sure I just did something weird myself.

cmd/commands.go
Phrynobatrachus: `caddy fmt` does not use quotes so I left them out

Phrynobatrachus:
Thanks for the reviews. :)

mholt (approved):
Awesomesauce, let's try it out!

Thanks for the great contribution.

I'll add you as a collaborator to the repository as we'd be happy to continue collaborating with you, and this way you can make patches or enhancements in a branch and not a fork.

mholt:
I feel like the CI is drunk right now, will try again later

Phrynobatrachus:
Ah I believe that's my bad actually, `errors.Join()` is Go 1.20. Lookin at it.

mholt:
Thank you again!!

Phrynobatrachus:
It's cool to land it! I'll definitely keep an eye out for other contributions, especially anything shaking out from this. Nice working with you both. :point_right:

francislavoie:
Btw, if you'd like to open PRs to fix the `IsTerminal` thing in some of the storage plugins, I'm sure that would be appreciated

PR #5535, opened 2023-05-16
headers: Add `>` Caddyfile shortcut for enabling `defer`

francislavoie:
Something that I've thought about a few times, I find the syntax awkward for setting a response header operation to be deferred since it requires opening a block to enable `defer` in simple cases.

So my solution is to add another prefix similarly to our existing ones, `>` meaning "set with defer", think an arrow meaning "later".

Before:
```
header {
	Field value
	defer
}
```

After:
```
header >Field value
```

### DIFF
--- a/caddytest/integration/caddyfile_adapt/header.txt
+++ b/caddytest/integration/caddyfile_adapt/header.txt
@@ -17,6 +17,7 @@
 		+Link "Foo"
 		+Link "Bar"
 	}
+	header >Set Defer
 }
 ----------
 {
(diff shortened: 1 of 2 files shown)


mholt (approved):
I'm OK with this. :+1: Just a quick thought: what about `...` as the prefix?

But yeah, anyway, this LGTM. Feel free to merge!

francislavoie:
Meh, multi-char prefixes doesn't have precedent. I prefer single-char. I can't think of anything better that's ASCII-ish.

PR #5536, opened 2023-05-16
Fix file browser footer while in grid mode

pistasjis:
I looked at the new file browser template and found it to be pretty good looking. Nice job! I tried it locally and found that the footer looked something like this: 
Screenshot 2023-05-16 at 19-38-12 samsungsd
The black space after the footer looks kinda awkward, so I decided to fix it. With this PR, it looks like this: 
Screenshot 2023-05-16 at 19-38-43 samsungsd

### DIFF
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -104,6 +104,7 @@ body {
 	font-size: 16px;
 	text-rendering: optimizespeed;
 	background-color: #f3f6f7;
+	min-height: 100vh;
 }
 
 img,
@@ -424,6 +425,8 @@ footer {
 	padding: 40px 20px;
 	font-size: 12px;
 	text-align: center;
+	position: sticky;
+	top: 100vh;
 }
 
 .caddy-logo {
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -425,8 +425,6 @@ footer {
 	padding: 40px 20px;
 	font-size: 12px;
 	text-align: center;
-	position: sticky;
-	top: 100vh;
 }
 
 .caddy-logo {
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ Caddyfile.*
 # build artifacts and helpers
 cmd/caddy/caddy
 cmd/caddy/caddy.exe
+cmd/caddy/tmp/*.exe
 
 # mac specific
 .DS_Store


francislavoie:
I think it looks better before, honestly. I don't think the footer should grow. It also stretches the gradient which causes worse color banding.

mholt:
Should the footer be stuck to the bottom (but not span the remaining height) instead?

francislavoie:
I suggest removing this:
```
@media (prefers-color-scheme: dark) {
	html {
		background: black; /* overscroll */
	}
```

This makes the gradient bg go all the the way to the bottom of the viewport.

pistasjis:
> image Like this?

image
Here's another where I set the background to the same as the grid view background. I'm not sure if it looks good or not though

pistasjis:
> > image Like this?
> 
> image Here's another where I set the background to the same as the grid view background. I'm not sure if it looks good or not though

image
I kinda like this one

pistasjis:
> > > image Like this?
> > 
> > 
> > image Here's another where I set the background to the same as the grid view background. I'm not sure if it looks good or not though
> 
> image I kinda like this one

That being said, if you have color suggestions or another suggestion for how it should look, please tell me

mholt:
Sorry, just took a lunch break, and am about to run some errands -- I'll look at this again in a little bit :) Thanks!

Skylerspencer (approved):

modules/caddyhttp/fileserver/browse.html
```diff
@@ -424,6 +425,8 @@
 	padding: 40px 20px;
 	font-size: 12px;
 	text-align: center;
+	position: sticky;
+	top: 100vh;
```
mholt, line 429:
```suggestion
```

mholt (changes requested):
OK finally had a chance to clone this down and check it out. It definitely looks better, thanks!

I noticed, however, that it pushes the logo all the way to the bottom of the screen. I know I suggested this earlier, but on second thought I don't think it needs to be down there.

Turns out that this is fixed with just the 1 line of CSS.

IMO it looks good on my monitor (no banding that I can see). And I prefer to keep things simple.

If the suggestion is accepted we can just merge this in I think :+1:

modules/caddyhttp/fileserver/browse.html
pistasjis: Done

mholt:
Awesome, thank you for the contribution!! If we decide for some reason that the footer is best at the bottom, we know what to do now. :) Thanks again.

mholt:
@pistasjis Also, you may notice I just pushed bd34cb6b4e171567243ff541ca303ac32fec91d8 which adds a lot more filetypes for the icons.

pistasjis:
> @pistasjis Also, you may notice I just pushed [bd34cb6](https://github.com/caddyserver/caddy/commit/bd34cb6b4e171567243ff541ca303ac32fec91d8) which adds a lot more filetypes for the icons.

They look nice. I've used Tabler before and they're one of my favourite iconsets

PR #5539, opened 2023-05-18
Do not replace import tokens if they are part of a snippet

WeidiDeng:
~~As discussed in slack, if an imported file also has a snippet with variadic placeholders and imports this snippet, parsing will fail saying not enough arguments.~~

~~Added tests show what the problem is.~~

When importing, caddy will always replace tokens placeholders. However, when importing a file that contains snippets definition, caddy will replace **them**. For `{args[n]}` type token, they will either be replaced or dropped, which is undesirable since they are part of a snippet.

Implement a naive algorithm to skip replacing tokens if they are part of a snippet. It **does not** check for syntax error, that will be done later when caddy finished importing all the tokens.

### DIFF
--- a/caddyconfig/caddyfile/importargs.go
+++ b/caddyconfig/caddyfile/importargs.go
@@ -87,14 +87,8 @@ func parseVariadic(token Token, argCount int) (bool, int, int) {
 
 // makeArgsReplacer prepares a Replacer which can replace
 // non-variadic args placeholders in imported tokens.
-func makeArgsReplacer(args []string, tokens []Token) *caddy.Replacer {
+func makeArgsReplacer(args []string) *caddy.Replacer {
 	repl := caddy.NewEmptyReplacer()
-	// record text and corresponding snippet name to not output warning during imports
-	// which import other snippets.
-	textToSnippetMap := make(map[string]string, len(tokens))
-	for _, token := range tokens {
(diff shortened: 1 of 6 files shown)


WeidiDeng:
I cannot get rid of the warning. Don't know a good way to hide these false positives. 😞

francislavoie:
Oh, good, I like that you simplified the fix in that last commit. I thought it was weird that the tokens were passed in.

WeidiDeng:
@mholt This patch changes the how variadic placeholders work. Now it cannot be used outside snippets. `args` is an `import` feature, not a snippet feature. This patch fixes variadic placeholders by making them not available outside snippets. 

The way to fix it requires at least 2 passes and obviously, rewriting how caddy file parsing works. What do you think of it?

WeidiDeng:
I think the problem is snippets should be exempt from the replace rule but not. Implemented a naive way to check it.

mholt:
So, to be sure I'm clear on this, does this patch cause something to _stop_ working? Based on the comments above, it sounds like using variadic placeholders, i.e. placeholders like `{args[:n]}` can only be used in Caddyfile snippets. Were they able to be used other places before? :thinking:

mholt (approved):

francislavoie (approved):
LGTM!

mholt:
Thanks for the contribution, @WeidiDeng !

Issue #5538, opened 2023-05-17
Constantly reloading of Caddyfile with "--watch"


ueffel:
caddy version: v2.7.0-beta.1

As a result of #5521 running caddy per `caddy run --watch`

1. the log gets very spammy. Every second, there is a line like this:
  
   ```plain
   2023/05/17 12:46:42.285	INFO	using provided configuration	{"config_file": "Caddyfile", "config_adapter": ""}
   2023/05/17 12:46:43.291	INFO	using provided configuration	{"config_file": "Caddyfile", "config_adapter": ""}
   2023/05/17 12:46:44.287	INFO	using provided configuration	{"config_file": "Caddyfile", "config_adapter": ""}
   ```
2. Even relatively easy Caddyfiles are not always byte-equal after loading, which is why it is reloaded very often even though there are no changes. This line is the culprit https://github.com/caddyserver/caddy/blob/v2.7.0-beta.1/cmd/main.go#L224
   Example Caddyfile:
   ```caddy-d
   {
   	log {
   		format console {
   			time_format wall_milli
   			time_local
   			level_format upper
   		}
   	}
   }
   
   (config) {
   	root .
   	file_server browse
   }
   
   http://:80 {
   	import config
   }
   
   :443 {
   	tls internal
   	import config
   }
   ```

   lastCfg contents:
   ```json
   {"logging":{"logs":{"default":{"encoder":{"format":"console","level_format":"upper","time_format":"wall_milli","time_local":true}}}},"apps":{"http":{"servers":{"srv0":{"listen":[":443"],"routes":[{"handle":[{"handler":"vars","root":"."},{"browse":{},"handler":"file_server","hide":[".\\Caddyfile",".\\Caddyfile:22 (import config)"]}]}]},"srv1":{"listen":[":80"],"routes":[{"handle":[{"handler":"vars","root":"."},{"browse":{},"handler":"file_server","hide":[".\\Caddyfile:17 (import config)"]}]}]}}},"tls":{"automation":{"policies":[{"issuers":[{"module":"internal"}]}]}}}}
   ```
   
   newCfg contents:
   ```json
[...]

PR #5540, opened 2023-05-19
config: keep track of import instead of modifying file name

WeidiDeng:
Previously when implementing variadic placeholders, caddy config token's `File` will be modified to reflect their import chain to help debug config error. One side effect is the [extra colon](https://github.com/caddyserver/caddy/issues/5538#issuecomment-1551609453) for imported files in file_server.

This patch uses a separate field to track import chain, and still keeps the error output the same.

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"log"
 	"strconv"
+	"strings"
 )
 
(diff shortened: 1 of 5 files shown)


mholt (approved):
Oh, yess -- thank you. I'll await Francis' approval when he has a chance as well (no rush).

francislavoie (approved):
Yeah, this is much better! Thanks!

mholt (approved):
Happy to try this out :smiley: 

Thank you as usual, @WeidiDeng !

PR #5541, opened 2023-05-19
Fix h3 shutdown

WeidiDeng:
Detailed discussing in slack.

If there are multiple listening ports, h3server won't close some of its quic listeners, leading to resource leak.

More explanation in comments.


mholt (approved):
Ah, I think that makes sense. I haven't verified this myself but as usual I trust you have. LGTM, thank you very much for the patch!

PR #5545, opened 2023-05-19
Make grid entires take up full width on mobile view and fix breadcrumb colors in dark mode mobile view (fileserver)

pistasjis:
Title explains this PR. Here's how the grid view looks like on mobile with this PR: 
image

Here's how it looks like without these changes: 
image

I don't know if the grid entry width thing is something you guys want (personally I think it looks better), but the breadcrumbs should not be black on mobile at least.

### DIFF
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -652,6 +652,10 @@ footer {
 		color: white;
 	}
 
+	h1 {
+		color: white;
+	}
+
 	h1 a:hover {
 		background: hsl(213deg 100% 73% / 20%);
 	}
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/fileserver/browse.html
```diff
@@ -714,15 +719,21 @@
 		stroke: #ccc !important;
 	}
 }
+
+@media (prefers-color-scheme: dark) and (max-width: 600px) {
+	h1 {
+		color: #fff;
+	}
+}
```
mholt, line 727: Nice catch, thanks. Why does this have to be limited to < 600px? I feel like the color shouldn't depend on screen size.

Maybe move it into the media query above? (And use `white` for consistency, not that it matters, just a nit.)

mholt (changes requested):
Ah yeah that does look much better. Thanks for scrutinizing this! I admit I didn't spend much time checking mobile-friendliness, just a quick once-over.

modules/caddyhttp/fileserver/browse.html
pistasjis: Done.

mholt (approved):
Awesome, thanks! Love these kinds of little improvements.

PR #5547, opened 2023-05-20
fileserver: Allow Etag override

### LINKED ISSUE #5546: ETag headers explicitly set in configuration should be honored rather than overridden
The mtime+size pattern used to calculate default etags by Caddy is only useful in situations where mtime is accurate. This isn't always the case: Some build systems (notably [Nix](https://nixos.org/)) coerce all times to 1970-01-01 for better reproducibility.

Right now, however, any etag that might be set by other configuration or modules is overwritten by the default mtime+size etag when caddy is serving requests. This makes Caddy deeply unreliable for serving files built with Nix, as it inaccurately instructs client browsers that contents are unchanged whenever subsequent builds have identical file sizes.

If the customer deploying caddy has a better identifier (such as the overall hash of the filesystem tree being served), they should be able to configure an ETag header using that identifier and let it be honored.

charles-dyfis-net:
Fixes #5546

Tested with a Caddyfile with a form roughly resembling:

```
route {
  header {
    ETag "\"${content-hash-here}\""
    -Last-Modified
  }
  root * /content/by-hash/${content-hash-here}
  file_server
}
```

(with `${content-hash-here}` replaced with an actual hash rendered to base-16).

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -356,7 +356,9 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 	}
 
 	var file fs.File
-	var etag string
+
+	// etag is usually unset, but if the user knows what they're doing, let them override it
+	etag := w.Header().Get("Etag")
 
 	// check for precompressed files
 	for _, ae := range encode.AcceptedEncodings(r, fsrv.PrecompressedOrder) {
@@ -388,7 +390,9 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 		// don't assign info = compressedInfo because sidecars are kind
 		// of transparent; however we do need to set the Etag:
 		// https://caddy.community/t/gzipped-sidecar-file-wrong-same-etag/16793
-		etag = calculateEtag(compressedInfo)
+		if etag == "" {
+			etag = calculateEtag(compressedInfo)
+		}
 
 		break
 	}
@@ -408,7 +412,9 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 		}
 		defer file.Close()
 
-		etag = calculateEtag(info)
+		if etag == "" {
+			etag = calculateEtag(info)
+		}
 	}
 
 	// at this point, we're serving a file; Go std lib supports only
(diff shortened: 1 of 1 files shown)


mholt (approved):
I'd say this LGTM. Thank you!

francislavoie:
FYI regarding your example, you can use backticks instead of double quotes to avoid needing to escape them. See the docs https://caddyserver.com/docs/caddyfile/concepts#tokens-and-quotes

IndeedNotJames:
@charles-dyfis-net FYI there is also a slightly different approach in NixOS/nixpkgs#222354

charles-dyfis-net:
> @charles-dyfis-net FYI there is also a slightly different approach in [NixOS/nixpkgs#222354](https://github.com/NixOS/nixpkgs/pull/222354)

This change strikes me as complimentary: if fileserver honors existing ETag values, a plugin can be used to set them so there's a cleaner way to use out-of-tree code to extract an ETag from the filesystem path.

charles-dyfis-net:
Hmm. There's something I don't understand here.

Going back over the configuration I successfully tested it with, a closer approximation would be:

```
route {
  header ETag `"${content-hash-here}"`
  root * /content/by-hash/${content-hash-here}
  file_server
  header -Last-Modified
}
```

Note, two separate header blocks. If I combine them into a single header block as shown above, Etag is no longer set before file_server is invoked.

mholt:
Correct. If you remove a header it defers the operation until the response is written. The two header lines is the correct way to do it in your case where you need one set before the write.

charles-dyfis-net:
Ah! I didn't pay enough attention to where exactly the `deferred` tag was in the `adapt` output. Makes perfect sense, then.

francislavoie:
Yeah if you use `-` in the same header block it enables `defer` mode which makes header operations happen on the way back up the middleware pipeline instead of on the way down, so it runs after the file server.

mholt:
I didn't even notice that in the original config. :sweat_smile:  (I wonder if we should print a warning in the logs.)

I like to think of it as: add first, delete later.

charles-dyfis-net:
Yup. I saw it in the docs, but read them to be saying that that individual operation would be deferred, as opposed to the whole batch.

mholt:
Gotcha. I'll try to clarify that. That might be a bug though. :thinking: I think I know why it is that way but I'm not sure if that's ideal/intentional. (Anyway, off-topic.)

Is this good to merge then?

charles-dyfis-net:
I'm not aware of anything that should block merge.

IndeedNotJames (approved):

PR #5550, opened 2023-05-22
fileserver: Don't set Etag if mtime is 0 or 1 (close #5548)

### LINKED ISSUE #5548: Consider dropping ETag, Last-Modified-Time generation for files where epoch <= mtime <= epoch+1
#5547, as a fix for #5546, makes it _possible_ to override caddy's default ETag generation.

That said, I'd like to suggest considering making that ETag generation conditional in the first place.

Not letting clients do caching at all is a better failure mode than encouraging clients to cache things that aren't safely cacheable. When content doesn't have a meaningful mtime, generating an ETag off that mtime leads to absurd results (effectively, leaving file size as the only cache-invalidation key). As a special case, an mtime of 0 or 1 (as measured in seconds-since-UNIX-epoch) is clearly not meaningful.

On its own, the existing PR makes it possible for the administrator of a system where mtimes are not meaningful to instruct Caddy to do something different instead.

If we implemented a safeguard such as that proposed here in addition to that PR, then the default for administrators who failed to take the actions enabled by that PR would be less harmful -- more client requests, rather than clients using out-of-date data.

mholt:
Apparently useful for Nix environments

/cc @charles-dyfis-net

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -616,7 +616,11 @@ func (fsrv *FileServer) notFound(w http.ResponseWriter, r *http.Request, next ca
 // Prefix the etag with "W/" to convert it into a weak etag.
 // See: https://tools.ietf.org/html/rfc7232#section-2.3
 func calculateEtag(d os.FileInfo) string {
-	t := strconv.FormatInt(d.ModTime().Unix(), 36)
+	mtime := d.ModTime().Unix()
+	if mtime == 0 || mtime == 1 {
+		return "" // not useful anyway; see issue #5548
+	}
+	t := strconv.FormatInt(mtime, 36)
 	s := strconv.FormatInt(d.Size(), 36)
 	return `"` + t + s + `"`
 }


charles-dyfis-net:
This solves half the problem -- the other half, `Last-Modified` (and browsers' use, in its presence, of `If-Modified-Since`), is unfortunately internal to net/http, so I'm not sure what a clean solution for that piece would look like.

mholt:
Ah... I see.

Yeah, either Go would need to be patched or you'll need some config that removes the header at write-time, like `header -Last-Modified`.

I could see this kind of proposal having some weight with the Go team though, I mean, zeroed timestamps aren't useful for Last-Modified in the first place.

I would recommend opening an issue there -- approach it like a bug report. Setting a last-modified header based on a zero-value timestamp seems wrong. :thinking:

charles-dyfis-net:
Alas, in the Nix case, the constant is 1, specifically to reduce zero-value special-case handling breaking software builds; that makes it a harder sell. I'll grant that it may be worth a try, though.

mholt:
Interesting, wouldn't you want it to be a zero-value since you want it to disregard the value?

But anyway, yeah, worth a shot to ask them. If they refuse then I recommend the approach to simply remove the header with a line of config.

mholt:
@charles-dyfis-net Looking closer, the Go standard lib DOES skip setting Last-Modified if the time is zero (or unix epoch):

https://github.com/golang/go/blob/8c445b7c9fe6738cbef2040a1011bd11489b0806/src/net/http/fs.go#L551-L553

So maybe they just need to do the same thing if it's 1 second after.

emilazy:
Just checking, but does this mean that it's possible to override the Last-Modified header through the configuration or a module when serving static files? When I looked at the file server code it seemed like there was a direct path from the file's mtime to the final delivery through the Go HTTP library, but I didn't pry further than that so it's possible I'm misunderstanding something.

Thanks for this change, also! I forget why exactly Nix uses mtime=1 but I think it's because a lot of programs behave weirdly or break if you use the epoch time...

mholt:
@emilazy

> Just checking, but does this mean that it's possible to override the Last-Modified header through the configuration or a module when serving static files? When I looked at the file server code it seemed like there was a direct path from the file's mtime to the final delivery through the Go HTTP library, but I didn't pry further than that so it's possible I'm misunderstanding something.

Good question -- the answer is yes! If deferred, Caddy's header deletions don't happen until an HTTP handler writes the header to the response. So yes, Go's file server does write the response directly directly, but it calls our ResponseWriter wrapper, which modifies the headers before writing.

So a config like this:

```
header -Last-Modified
```

will delete the `Last-Modified` header _when the response gets written_ because delete operations are implicitly deferred.

However, adding or setting headers is _not_ deferred by default, so:

```
header Last-Modified foobar
```

would not have the desired outcome. However, this would:

```
header Last-Modified foobar {
    defer
}
```

because it explicitly defers the manipulation to write-time.

I know that's confusing, but it didn't used to be this way and it was confusing that deletions happened right away by default because later HTTP handlers in the chain would then add the header that the user was trying to delete :man_facepalming: So we decided it was best for deletions to be implicitly deferred.

emilazy:
Thanks, that's great to know! Hopefully this means that we can implement all the functionality we need with a module on the Nix side.

PR #5553, opened 2023-05-22
templates: Add `readFile` function, skips parsing contents as a template

### LINKED ISSUE #5552: Template function to include file without parsing that file as a template (ie. includeRaw)
I'm trying to create a simple markdown-based blog using Caddy. One issue I'm running in to though, is when I try to include the markdown file to `splitFrontMatter` on it (or later to render it), I run into an issue where the contents of the markdown file are parsed as a template. 

A possible workaround for this is to set `between` on the `templates` directive [as linked here](https://caddyserver.com/docs/caddyfile/directives/templates), but, again, this is more of a workaround than a real fix. Specifically, this fix wouldn't work in the case where you're trying to document how you created the blog, in a blog post, since you'd need to show the templates in that post!

What I'm proposing is a `includeRaw` template function, where the contents of the file are read and returned, but without parsing the given file as a template, to prevent issues like the infinite recursion I was hitting when trying to use `include` in my use case.

kassienull:
This is a fix for the issue I created earlier, closes #5552. It's mostly a copy-paste of the `include` function, but skips the template parsing step to just return the body of the file unmodified.

### DIFF
--- a/modules/caddyhttp/templates/templates.go
+++ b/modules/caddyhttp/templates/templates.go
@@ -120,6 +120,14 @@ func init() {
 // {{include "path/to/file.html" "arg1" 2 "value 3"}}  // with arguments
 // ```
 //
+// ##### `includeRaw`
+//
+// Includes the contents of another file, without rendering it as a template.
+//
+// ```
+// {{includeRaw "path/to/file.html"}}
+// ```
+//
 // ##### `listFiles`
 //
 // Returns a list of the files in the given directory, which is relative to the template context's file root.
--- a/modules/caddyhttp/templates/templates.go
+++ b/modules/caddyhttp/templates/templates.go
@@ -88,7 +88,11 @@ func init() {
 //
 // ##### `httpInclude`
 //
-// Includes the contents of another file by making a virtual HTTP request (also known as a sub-request). The URI path must exist on the same virtual server because the request does not use sockets; instead, the request is crafted in memory and the handler is invoked directly for increased efficiency.
+// Includes the contents of another file, and renders it in-place,
+// by making a virtual HTTP request (also known as a sub-request).
+// The URI path must exist on the same virtual server because the
+// request does not use sockets; instead, the request is crafted in
+// memory and the handler is invoked directly for increased efficiency.
 //
 // ```
 // {{httpInclude "/foo/bar?q=val"}}
@@ -96,7 +100,13 @@ func init() {
 //
 // ##### `import`
 //
-// Imports the contents of another file and adds any template definitions to the template stack. If there are no defitions, the filepath will be the defition name. Any {{ define }} blocks will be accessible by {{ template }} or {{ block }}. Imports must happen before the template or block action is called
+// Reads and returns the contents of another file, and parses it
+// as a template, adding any template definitions to the template
+// stack. If there are no definitions, the filepath will be the
+// definition name. Any {{ define }} blocks will be accessible by
+// {{ template }} or {{ block }}. Imports must happen before the
+// template or block action is called. Note that the contents are
+// NOT escaped, so you should only import trusted template files.
 //
 // **filename.html**
 // ```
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/templates/templates.go
```diff
@@ -120,6 +120,13 @@ func init() {
 // {{include "path/to/file.html" "arg1" 2 "value 3"}}  // with arguments
 // ```
 //
+// ##### `includeRaw`
```
francislavoie, line 123: Needs a newline here I think

```suggestion
// ##### `includeRaw`

```

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -17,6 +17,9 @@ package templates
 import (
 	"bytes"
 	"fmt"
+	"github.com/Masterminds/sprig/v3"
+	"github.com/caddyserver/caddy/v2"
+	"github.com/dustin/go-humanize"
```
francislavoie, line 22: I think your IDE moved these import lines. I don't think it should be done this way, they should be grouped by source, so all stdlib first then github ones together, etc. Running `gofmt -s -w` can fix this up I think.

modules/caddyhttp/templates/templates.go
kassienull: Yep, I think so, but looking more at the format of the comments around it, shouldn't it be more like this?
```go
// ##### `includeRaw`
//
// Includes the contents of another file, without rendering it as a template.
//
// ```
// {{includeRaw "path/to/file.html"}}
// ```
//
```

francislavoie: Yeah oops my bad, that's what I meant to suggest but I just didn't type the `//`

kassienull: That's what I figured, just wanted to make sure!

modules/caddyhttp/templates/tplcontext.go
kassienull: It didn't, but I'll just copy paste the imports from the unmodified file

kassienull:
Fixed the formatting issues + squashed the two commits together

mholt (approved):
Cool, this LGTM. I'm just waiting to merge while I try to think of a better name than `includeRaw` if one exists. :) That's all.

kassienull:
Maybe `includeText`? Honestly I just did `includeRaw` since that's what someone mentioned in Discord and I'm not that good at naming 😅

mholt:
Yeah that's alright. It's not bad, just seems like an afterthought (which it is).

What I should have done is named `include` `eval` and then `include` could be the un-evaluated includer. Sigh.

francislavoie:
Maybe like readFile or something?

crckrberries:
lgtm

mholt:
I like `readFile` -- let's do that!

francislavoie:
I pushed a commit to rename it, and I made some additional adjustments to the template docs cause I thought it wasn't super clear which functions would render in place and which would just return the buffer.

modules/caddyhttp/templates/templates.go
```diff
@@ -88,15 +88,25 @@ func init() {
...
 // ##### `import`
 //
-// Imports the contents of another file and adds any template definitions to the template stack. If there are no defitions, the filepath will be the defition name. Any {{ define }} blocks will be accessible by {{ template }} or {{ block }}. Imports must happen before the template or block action is called
+// Reads and returns the contents of another file, and parses it
+// as a template, adding any template definitions to the template
+// stack. If there are no definitions, the filepath will be the
+// definition name. Any {{ define }} blocks will be accessible by
+// {{ template }} or {{ block }}. Imports must happen before the
+// template or block action is called. Note that the contents are
+// NOT escaped, so you should only import trusted template files.
```
mholt, line 109: Are these docs accurate? Since we're not evaluating templates I think we can probably delete all of this and simply have:

```suggestion
// Reads and returns the contents of another file without evaluating any template actions.
```

mholt (approved):
Thanks, LGTM. (Don't mind the deleted comment; I was thrown off, not realizing that docs for other actions were edited -- I'm not sure why though.) (Edit: D'oh, just now seeing Francis' comment. Sigh.)

I think we can merge this :smiley:  Thanks for the contribution!

PR #5554, opened 2023-05-23
listen_unix.go: Use consts to set SO_REUSEPORT_LB on FreeBSD

jpds:
This option is preferable on FreeBSD as it uses a hash to balance load between threads.

### DIFF
--- a/listen_unix_setopt.go
+++ b/listen_unix_setopt.go
@@ -1,0 +1,7 @@
+//go:build unix && !freebsd
+
+package caddy
+
+import "golang.org/x/sys/unix"
+
+const unixSOREUSEPORT = unix.SO_REUSEPORT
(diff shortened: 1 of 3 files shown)


mholt:
LGTM -- didn't know about this one. Thanks!

listen_unix_setopt.go
```diff
@@ -0,0 +1,7 @@
+//go:build linux || openbsd || darwin
```
mholt, line 1: As we're unraveling the `unix` build tag value, we probably need to consider all of these individually now :(

https://go-review.googlesource.com/c/go/+/389934/7/src/go/build/syslist.go

Unless there's a way to say "unix but not freebsd"

mholt (changes requested):
I think we just need to make sure the enumerated platforms are correct...

mholt (approved):
Perfection. I think we can try this out! Thanks.

(Our project root is getting a bit littered with files, ah well.)

Issue #5471, opened 2023-03-30
Don't close active websocket connections on config reload


kevinlang:
Split off from #3143

Currently Caddy closes all active websocket connections on config reload. This impacts the ability to do graceful blue-green deployments where all new traffic is directed to the new version of the application while the existing application's websocket connections are slowly drained at the application level.

Ideally, Caddy would not close active websocket connections associated with old configs/upstreams, instead keeping them around. Optionally, there can be a new config value that is an upper-bound on how long they should be kept around before being forced-close to address any sort of memory leak concerns people may have. (see https://github.com/caddyserver/caddy/issues/3143#issuecomment-1478298087)

This behavior of keeping active websockets around for old/updated upstream's matches NGINX's behavior for config reloads.

francislavoie:
Sure, you can give it a shot if you'd like. I started work on this a while back but didn't finish it. https://github.com/caddyserver/caddy/compare/ws-close-timeouts The config layer is done, but the actual usage of the timeouts is incomplete/wrong.

mmm444:
Hello. I was also interested in having this done so I decided to give it a try and I think I have figured the closing logic out. For the moment the result is at https://github.com/mmm444/caddy/tree/ws-close-timeouts .

Shall I file a PR? And if yes, against what branch? master or ws-close-timeouts?

francislavoie:
Interesting! Yeah I never opened a PR (because I didn't think the code quality was quite there yet) so feel free.

What kind of tests did you perform to validate it?

mholt:
That's cool, we'd be happy to review it -- feel free to open a PR. Thanks for working on this!

mmm444:
Thanks. Will submit PR in a short time.

> What kind of tests did you perform to validate it?

Basically manual tests only. I had a caddy with reverse proxy (obviously) and one upstream server that accepted websockets. I used a browser to initiate the websocket connection. And this is what I did:
* set only StreamTimeout and verified that the connection got closed after the given timeout
* set only StreamTimeout, closed the WS from the browser and verified that the timeout timer got canceled
* set only StreamCloseDelay, triggered config reload and verified that the stream got closed after the delay
* set only StreamCloseDelay, triggered config reload, manually closed the WS in browser and verified that the delay timer got destroyed just after the websocket closed
[...]

PR #5567, opened 2023-06-08
reverseproxy: streaming timeouts

mmm444:
This PR adds two settings related to streaming connections to the reverse_proxy handler:

`stream_timeout` defines the maximum lifetime of a streaming connection in the reverse proxy; when the connection reaches this age it is closed
`stream_close_delay` defines the time for which the proxy waits before closing the streaming connections when it is cleaned up i.e. after a configuration change

#5471

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -212,7 +212,7 @@ type Handler struct {
 
 	// Stores upgraded requests (hijacked connections) for proper cleanup
 	connections           map[io.ReadWriteCloser]openConnection
-	connectionsCloseTimer **time.Timer
+	connectionsCloseTimer *time.Timer
 	connectionsMu         *sync.Mutex
 
 	ctx    caddy.Context
@@ -239,8 +239,6 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 	h.logger = ctx.Logger()
 	h.connections = make(map[io.ReadWriteCloser]openConnection)
 	h.connectionsMu = new(sync.Mutex)
-	h.connectionsCloseTimer = new(*time.Timer)
-	*h.connectionsCloseTimer = nil
 
 	// TODO: remove deprecated fields sometime after v2.6.4
 	if h.DeprecatedBufferRequests {
@@ -909,7 +907,7 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
 }
 
 // finalizeResponse prepares and copies the response.
-func (h Handler) finalizeResponse(
+func (h *Handler) finalizeResponse(
 	rw http.ResponseWriter,
 	req *http.Request,
 	res *http.Response,
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -211,8 +211,9 @@ type Handler struct {
 	handleResponseSegments []*caddyfile.Dispenser
 
 	// Stores upgraded requests (hijacked connections) for proper cleanup
-	connections   map[io.ReadWriteCloser]openConnection
-	connectionsMu *sync.Mutex
+	connections           map[io.ReadWriteCloser]openConnection
+	connectionsCloseTimer **time.Timer
+	connectionsMu         *sync.Mutex
 
 	ctx    caddy.Context
 	logger *zap.Logger
@@ -238,6 +239,8 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 	h.logger = ctx.Logger()
 	h.connections = make(map[io.ReadWriteCloser]openConnection)
 	h.connectionsMu = new(sync.Mutex)
+	h.connectionsCloseTimer = new(*time.Timer)
+	*h.connectionsCloseTimer = nil
 
 	// TODO: remove deprecated fields sometime after v2.6.4
 	if h.DeprecatedBufferRequests {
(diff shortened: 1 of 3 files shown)


francislavoie:
Whoops, looks like there's a conflict. Could you rebase and fix it?

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -383,24 +399,51 @@ func (h *Handler) Provision(ctx caddy.Context) error {
...
+			closeErr := oc.conn.Close()
+			if closeErr != nil && err == nil {
+				err = closeErr
 			}
 		}
-		closeErr := oc.conn.Close()
-		if closeErr != nil && err == nil {
-			err = closeErr
+		return err
+	}
```
mholt, line 423: This could probably be a separate function (unexported, but doesn't need to be nested inside Cleanup as far as I can tell.)

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -383,24 +399,51 @@ func (h *Handler) Provision(ctx caddy.Context) error {
...
+
+		for _, oc := range h.connections {
+			if oc.gracefulClose != nil {
+				// this is potentially blocking while we have the lock on the connections
+				// map, but that should be OK since the server has in theory shut down
+				// and we are no longer using the connections map
+				gracefulErr := oc.gracefulClose()
+				if gracefulErr != nil && err == nil {
+					err = gracefulErr
+				}
```
mholt, line 415: Slightly nervous about this, but maybe we can revisit this one more time before merge.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -198,8 +211,9 @@ type Handler struct {
 	handleResponseSegments []*caddyfile.Dispenser
 
 	// Stores upgraded requests (hijacked connections) for proper cleanup
-	connections   map[io.ReadWriteCloser]openConnection
-	connectionsMu *sync.Mutex
+	connections           map[io.ReadWriteCloser]openConnection
+	connectionsCloseTimer **time.Timer
```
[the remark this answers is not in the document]
mmm444: The reason for this is the same as why there is `*sync.Mutex` and not just `sync.Mutex`. As I have found during debugging there are multiple instances of `Handler` struct floating around during runtime and I needed to synchronize whether there is `*time.Timer` among all of them or at least some of them. Specifically the registration and the cleanup code are called on different instances of `Handler`. Why is this so? Is it an unfortunately leaked reference from a non-pointer receiver method? Or does it have some other reason?

And no, although I can write some C I would not call myself a C programmer. :smile:

modules/caddyhttp/reverseproxy/reverseproxy.go
mmm444: Me too, OK.

modules/caddyhttp/reverseproxy/reverseproxy.go
mmm444: Done. While looking for the best place where to put it I also split the `Cleanup` method and moved the parts that are related to hijacked connection into `streaming.go` so that the hijacked connection handling methods are close together.

francislavoie:
Thanks a ton for finishing this off for me! The timer/channel bits are not a strength of mine. I don't have much to offer in terms of review but it's looking pretty good to me! :blush:

modules/caddyhttp/reverseproxy/reverseproxy.go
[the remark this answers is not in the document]
mholt: They should be on the same instance. The only reason I can think of why it would be different is because some of the methods don't have pointer receivers (i.e. some methods are `(h Handler)` because they, at least currently, don't intend to modify the handler and so this is kind of a safeguard -- but that `h` will be different than if the method has a pointer receiver, i.e. `(h *Handler)`).

mholt:
Thanks!

I think I want to try this in beta 2, and I don't want to wait much longer on that. AFAICT this change looks alright to me for a beta, but would you be alright if we work on it more after merging? I want to see what we can do about that double pointer, and also verify a few logical things. Oh, and we should probably mark this as experimental so we can change it later, before the final tag.

But I think there's value in having this for people to try right away.

mmm444:
> I think I want to try this in beta 2, and I don't want to wait much longer on that. AFAICT this change looks alright to me for a beta, but would you be alright if we work on it more after merging?

Yes. No problem. 

> I want to see what we can do about that double pointer, and also verify a few logical things.

I have outlined a solution in the comment above. 

So just 2 questions from me:
1. Shall I commit the removal of the double pointer before you merge it? I already have the commit ready, tested a little bit and it seems to be working. I just don't want to spontaneously change the code under review.
2. I have noticed in the contributors guide that you prefer the PR to be squashed into one commit. Shall I do that?

modules/caddyhttp/reverseproxy/reverseproxy.go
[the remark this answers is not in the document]
mholt: Yes! Absolutely.

mholt:
@mmm444

> Shall I commit the removal of the double pointer before you merge it? I already have the commit ready, tested a little bit and it seems to be working. I just don't want to spontaneously change the code under review.

Yes, please! Then I will merge this.

> I have noticed in the contributors guide that you prefer the PR to be squashed into one commit. Shall I do that?

Nowadays, GitHub has a button that does this, so no need for you to do it. :)

Thank you for working on this!

Let's continue any adjustments after beta 2 is released here in the next day or so.

francislavoie:
`30` means 30 microseconds. Try `30s` instead for 30 seconds.

mholt:
@xnaas Can you verify that you are running the correct Caddy build and config together?

@mmm444 Any ideas on this? @xnaas , maybe if you share your config @mmm444 could help.

mholt:
Thanks. @mmm444 do you have a chance to help out with this?

mmm444:
I would really like to help but I am on a vacation right now and testing this remotely is beyond my possibilities. I can get to this on next Thursday.

The situation seems to me like the basic use case of this feature that I have checked multiple times. The only difference is that I have never used the Caddyfile only JSON. Maybe there is a bug in the adaptation code? I admit I didn't check that.

mholt:
Enjoy your vacation :blush: Thanks for replying.

I just did a quick overview of the Caddyfile code, and it looks like it should be fine. :thinking: Maybe @xnaas could verify that the JSON is correct (use `caddy adapt --pretty` to inspect).

Hopefully we can figure this out when you get back then!

mmm444:
Ok. I have looked into this and it is caused by the `flush_interval -1` setting. Without it it seems to work. :grin: 

Setting `flush_interval`  to  `-1` causes the upstream (proxy to server) connections to get bound to the context of the whole reverse proxy handler (`h.ctx`) via `ignoreClientGoneContext` instead of the client connection and this context terminates when the configuration is reloaded. Closing the `h.ctx` context in turn terminates one of the websocket copy goroutines and this effectively closes the websocket connection before any timeout can take place.

I am not sure yet what is the best fix here. I will try to come up with something in the following days.

If someone wants to play with the situation, here is a minimal websocket server that can be used to reproduce the reported behavior. https://gist.github.com/mmm444/efc3e25fbbb0056f9c759d6dac3f65f0

Shall I file a bug about it or is it ok to continue the discussion here?

francislavoie:
Fascinating.

Coincidentally, Go 1.21 is introducing a new [`context.WithoutCancel()`](https://github.com/golang/go/issues/40221) function. I think that would be helpful in this case to allow WS connections to outlive the context. We can't use it quite yet though (probably for another 6 months to a year) because Caddy's minimum Go version will need to be increased to match.

I think the best thing to do for now is probably to add a warning log when these two config options are combined to mention that it might not work properly, unless you can figure out some other creative fix. We can open an issue as a reminder to implement `WithoutCancel` when we have the Go version to do it.

mholt:
@mmm444 Thanks for coming back to this!

Francis pointed out to me in Slack this is why that happens:

```go
	// if FlushInterval is explicitly configured to -1 (i.e. flush continuously to achieve
	// low-latency streaming), don't let the transport cancel the request if the client
	// disconnects: user probably wants us to finish sending the data to the upstream
	// regardless, and we should expect client disconnection in low-latency streaming
	// scenarios (see issue #4922)
	if h.FlushInterval == -1 {
		req = req.WithContext(ignoreClientGoneContext{req.Context(), h.ctx.Done()})
	}
```

That said, I feel like when the server flushes data should be orthogonal to when the connection is hung up.

Is there, like, a done chan we could give it that never gets closed? Or doesn't get closed until the stream_close_delay is up?

francislavoie:
I'm thinking we should add a condition to that `if h.FlushInterval == -1 {` condition to skip that behaviour for WebSocket requests. I can't think of a situation where that would be beneficial for WebSockets, it seems like it would only be harmful. It could make sense to configure `flush_interval -1` for a single backend which does both streaming _and_ WebSockets, but we'd want to ignore that option for any WebSocket requests.

mmm444:
I have been thinking about this over the weekend and come to that the correct for now (tm) solution is to not use the `h.ctx` context for shutting down any connections when `h.FlushInterval == -1`. It should be achievable using something like the WithoutCancel feature from go 1.21. The reasons for this are:
* Ongoing downloads/uploads will not get terminated upon config reloads. I find it a bit surprising, maybe even buggy, that it happens now. It does not happen when `h.FlushInterval != -1`. 
* The timeouts on websockets will get fixed.
* From the users' standpoint flushing tuning has nothing to do with websocket timeouts so it is better not to leak the implementation details (about configuration reloading) to them.

I can submit a PR if you want.

francislavoie:
Yeah, PR certainly welcome.

I don't think we can use `WithoutCancel` yet like I said earlier because we'd need to have Go 1.21 in `go.mod` as the minimum version (and obviously because Go 1.21 isn't out yet).

mholt:
@mmm444 Thanks for the careful thought. I agree with you. I initially used that context because if the client canceling the request doesn't close the connection, we still had to ensure it closed at some point: a config reload seemed like the right time. But it's really not, like you pointed out.

I just hope it won't lead to leaking resources. I guess if it's just proxying the stream, it's up to the client or the backend to actually close them.

mholt:
@mmm444 Would you still like to open a PR? If not, I can take a stab at it, but you're more attuned to this than I am currently :smiley:

mholt:
What about this as a simple patch?

```go
	if h.FlushInterval == -1 && h.StreamCloseDelay == 0 && h.StreamTimeout == 0 {
		req = req.WithContext(ignoreClientGoneContext{req.Context(), h.ctx.Done()})
	}
```

since if the close delay or timeout are set, the connection will definitely be closed at that point, right? We just need to ensure the connection closes eventually otherwise.

mmm444:
> @mmm444 Would you still like to open a PR? If not, I can take a stab at it, but you're more attuned to this than I am currently 😃
Yes. I am looking into this right now. Will submit a PR with some explanation in a few hours, hopefully. 😄 Sorry for the delay, I have been away from the internet once again.

mholt:
Awesome, thank you @mmm444 !

berkant:
So, should I put an incredibly long duration for `stream_close_delay` to keep WebSocket connections alive after reloading config? My use case includes configuring Caddy on-the-fly to add new `reverse_proxy` routes to various WebSocket upstreams through admin API, but for the time being, all connections do break after doing this.

francislavoie:
Yeah you should set it to some reasonable amount of time dependent on the usual session length of your users. But you should definitely make sure to have proper reconnect logic on the client-side to ensure that even if the connections are closed, it reconnects cleanly.

PR #5569, opened 2023-06-12
Fixes parsing of source IP in case it's an ipv6 address

gucki:
While testing with v2.7.0-beta.1: parsing of the source IP for the proxy protocol (introduced with #5424) fails with "ERROR   http.log.error  unexpected remote addr type in proxy protocol info" for IPv6 requests. The problem is how the dummy port was appended to the original source string "::1" + "0" -> "::1:0" which is not a proper IPv6 address.


mholt (approved):
Oh, thanks! :)

PR #5573, opened 2023-06-13
caddyhttp: support custom network for HTTP/3

### LINKED ISSUE #5555: caddyhttp: allow custom network for HTTP/3
Caddy has great support for supporting custom networks via the `RegisterNetwork` hook:

https://github.com/caddyserver/caddy/blob/cee4441cb1d485b38d728168a315cda5641d84fb/listeners.go#L682-L686

This allows us to serve http on a custom network with the caddyhttp module. Unfortunately, for http3, this only applies to the networks that are support by default in caddy. In caddyhttp, we switch over the network of the http1.1 listener to figure out the corresponding network for the http3 listener. This switch statement does not consider custom networks that were registered:
https://github.com/caddyserver/caddy/blob/cee4441cb1d485b38d728168a315cda5641d84fb/modules/caddyhttp/server.go#L519-L532

Instead of hard-coding this mapping, we could also introduce a hook system like we already have for `RegisterNetwork`.
The API could look something like this:

```go
package caddyhttp

// RegisterNetworkHTTP3 registers a mapping from non-HTTP/3 network to HTTP/3
// network. This should be called during init() and will panic if the network
// type is standard or reserved, or if it is already registered. EXPERIMENTAL
// and subject to change.
func RegisterNetworkHTTP3(originalNetwork, h3Network string) {}
```

If this is something you are interested in, I would be happy to contribute code.

oncilla:
Allow registering a custom network mapping for HTTP/3. This is useful if the original network for HTTP/1.1 and HTTP/2 is not a standard `unix`, `tcp4`, or `tcp6` network. To keep backwards compatibility, we fall back to `udp` if the original network is not registered in the mapping.

Fixes #5555

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -520,17 +520,7 @@ func (s *Server) findLastRouteWithHostMatcher() int {
 // not already done, and then uses that server to serve HTTP/3 over
 // the listener, with Server s as the handler.
 func (s *Server) serveHTTP3(addr caddy.NetworkAddress, tlsCfg *tls.Config) error {
-	switch addr.Network {
-	case "unix":
-		addr.Network = "unixgram"
-	case "tcp4":
-		addr.Network = "udp4"
-	case "tcp6":
-		addr.Network = "udp6"
-	default:
-		addr.Network = "udp" // TODO: Maybe a better default is to not enable HTTP/3 if we do not know the network?
-	}
-
+	addr.Network = getHTTP3Network(addr.Network)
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks for this!

Did you verify that there isn't a race condition? I imagine it should be fine as long as all the changes to the map happen before any goroutines are started.

francislavoie:
@mholt we're not doing any race condition checks for `RegisterNetwork` either which uses the same pattern of needing to be called in `init()`. The godoc comment says as much, so if someone does any differently they're going to have a bad time.

mholt:
I suppose it should be good then. Thanks!

PR #5574, opened 2023-06-15
headers: Allow `>` defer shortcut for replacements

francislavoie:
Tiny followup to #5535 to make `>` work as a prefix for replacements (i.e. 2 args following the header field).

For example:

```
header >Set-Cookie (.*) "$1; SameSite=None;"
```

Defer is usually necessary for replacements, so it should be as easy as possible to set up.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/header.txt
+++ b/caddytest/integration/caddyfile_adapt/header.txt
@@ -18,6 +18,7 @@
 		+Link "Bar"
 	}
 	header >Set Defer
+	header >Replace Deferred Replacement
 }
 ----------
 {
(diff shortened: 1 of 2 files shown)


mholt:
Thanks for this! Will give it a look soon...

mholt:
Here's a crazy idea, what if we used `...` instead?

(Tbh, I'm fine with this PR as-is, but I wanted to ask anyway since I think "..." could make intuitive sense.)

Man, I kind of wonder if defer should have been the default. And I also hate that if one is deferred, the whole block of header changes is also deferred. (Design decisions on my part, of course. Unrelated to this.)

francislavoie:
> Here's a crazy idea, what if we used ... instead?

Not a fan tbh. I prefer the single character prefix approach. We argued the same idea in the previous PR.

> Man, I kind of wonder if defer should have been the default.

Yeah same. Definitely too late to change now. Maybe in v3.

> And I also hate that if one is deferred, the whole block of header changes is also deferred.

True. But this shortcut makes that less annoying actually because you can more easily write one-liner header operations of all kinds.

mholt (approved):
Sounds good to me. It's the best we can do I guess :)

PR #5586, opened 2023-06-21
chore: upgrade otel

omerdemirok:
# What
- Upgrade otel dependencies to the latest stable version.


mholt:
Thanks; first I will check with a sponsor who has trouble with otel dependencies, to see if this will affect them.

Probably will merge after 2.7 though.

mholt:
Apparently the newer versions of Otel have some improvements that may be preferred, so we can merge this :)

mholt (approved):

PR #5596, opened 2023-06-24
core: Skip `chmod` for abstract unix sockets

emilylange:
Because abtract unix sockets aren't real paths on the filesystem and thus can't be `chmod`ed.

Fixes the issue reported by @lxhao61 in https://github.com/caddyserver/caddy/commit/22927e278dc29c9d1804c20f483510ec569f23ed#commitcomment-119530849

I believe this applies to all platforms, including Windows.

### DIFF
--- a/listeners.go
+++ b/listeners.go
@@ -151,6 +151,7 @@ func (na NetworkAddress) listen(ctx context.Context, portOffset uint, config net
 	var err error
 	var address string
 	var unixFileMode fs.FileMode
+	var isAbtractUnixSocket bool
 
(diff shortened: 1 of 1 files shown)


francislavoie:
Thanks! Should we have a test for this? I didn't even know abstract unix sockets were a thing. We should probably document it.

mholt (approved):
Thanks, I don't think I even knew about these :dizzy_face:

lxhao61:
Waiting for post-merge testing.

PR #5604, opened 2023-06-28
caddytls: Cache 'ask' results to reduce load

mholt:
Recently (2.7 beta) we made changes in CertMagic so that the 'ask' endpoint also guards access to the storage backend, as those can be rather expensive, and it doesn't make sense to even try loading a cert from storage if it's not even allowed to be managed.

However, now we have the potential of hammering the ask endpoint, [which has been a new complaint in the beta](https://caddy.community/t/caddy-2-7-0-beta-1/20255/7).

Seeing as whether a certain domain is allowed or not doesn't change THAT often in practice, we cache the answer for an hour (could potentially be configurable later, if needed). The cache is a very simple map capped at 1000 entries with random (enough) eviction.

### DIFF
--- a/modules/caddytls/acmeissuer.go
+++ b/modules/caddytls/acmeissuer.go
@@ -19,7 +19,6 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
-	"net/url"
 	"os"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


mholt:
This might not be needed after https://github.com/caddyserver/caddy/pull/5623.

That patch doesn't cause the certificate cache to be purged of its managed certificates at config reloads, so there shouldn't be any point where the 'ask' endpoint is getting hammered with the same domain as it did recently. Once the cert is loaded, it stays loaded until a config is loaded that doesn't use it. And so it won't hit the 'ask' endpoint to try loading from storage.

francislavoie:
I agree this isn't really useful. It's not actively harmful, but not particularly useful.

francislavoie:
I do think though, that caching _rejected_ domains might be useful to mitigate scrapers/bots/ddos from causing tons of `ask` calls, but we need to be somewhat careful because they could still just enumerate random domains and never repeat the same domain to bypass that kind of mitigation.

mholt:
Yeah, possibly. We should be careful about that if we do implement a strategy like this. I'll close this for now and see if there's really a need after #5623.

Issue #5609, opened 2023-07-01
Bug in LeastConnSelection Policy used in  reverseproxy


yashkundu:
There is a **BUG** in the way LeastConnSelection policy is selecting an upstream for the reverse proxy.
The following is the code of the select method of LeastConnSelection struct which chooses the upstream:
[caddy/blob/master/modules/caddyhttp/reverseproxy/selectionpolicies.go](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/selectionpolicies.go)

```
// Select selects the up host with the least number of connections in the
// pool. If more than one host has the same least number of connections,
// one of the hosts is chosen at random.
func (LeastConnSelection) Select(pool UpstreamPool, _ *http.Request, _ http.ResponseWriter) *Upstream {
	var bestHost *Upstream
	var count int
	leastReqs := -1

	for _, host := range pool {
		if !host.Available() {
			continue
		}
		numReqs := host.NumRequests()
		if leastReqs == -1 || numReqs < leastReqs {
			leastReqs = numReqs
			count = 0
		}

		// among hosts with same least connections, perform a reservoir
		// sample: https://en.wikipedia.org/wiki/Reservoir_sampling
		if numReqs == leastReqs {
			count++
			if count > 1 || (weakrand.Int()%count) == 0 { //nolint:gosec
				bestHost = host
			}
		}
	}

	return bestHost
}
```
This **always selects the last upstream which has the least number of active requests, but it should randomly choose among all the upstreams which have the least number of active requests**, reservoir sampling is implemented incorrecly, because of the extra condition (count>1) in the if clause.

francislavoie:
Oh I think that was meant to be an `&&` and not an `||`. The point of the change in https://github.com/caddyserver/caddy/commit/b19946f6af75a00bb1f3d094a903224541a70fcd to add that clause is that `x % 1 == 0` is always true so there's no use in generating a random number in that case. Minor optimization.

yashkundu:
I think we can use the condition:
```
if (count == 1) || (weakrand.Int()%count)
```
It will handle the condition: if the count is 1, random number won't be generated, otherwise will be generated. @francislavoie

mholt:
Does this still need to be fixed, @yashkundu ?

mholt:
That would be splendid 🙏 thank you!

francislavoie:
FYI @yashkundu https://github.com/caddyserver/caddy/pull/5862

PR #5610, opened 2023-07-01
reverseproxy: LeastConnSelection policy bug

yashkundu:
LeastConnSelection policy is always selecting the last upstream in the pool with the least number of active requests, in fact it should select randomly from all the upstreams which have the least number of active request.
Have corrected the implementation of reservoir sampling in the select method of the LeastConnSelection struct.
Here is the link to the issue:
[caddy/issues/5609](https://github.com/caddyserver/caddy/issues/5609)


mholt:
Hi! Thanks for the patch submission. Did you mean to close this?

PR #5614, opened 2023-07-02
cmd: Require config file exists for `caddy validate`

### LINKED ISSUE #5612: `caddy validate` does not check if the Caddyfile exists
I think this is pretty self-explanatory:
```
$ caddy adapt
Error: input file required when there is no Caddyfile in current directory (use --config flag)
$ caddy validate
Error: decoding config: unexpected end of JSON input
```

The `validate` command, unlike `adapt` misses check that assesses the file even exists in the first place. With that, the user may assume the file was read and that they have some syntax error they'll be endlessly looking for, even though that's not the problem...

Check for adapt exists here:
https://github.com/caddyserver/caddy/blob/dfe17c33efb1653c0dcbcb8238935e7c72d714dd/cmd/commandfuncs.go#L439-L442

Missing in this func:
https://github.com/caddyserver/caddy/blob/dfe17c33efb1653c0dcbcb8238935e7c72d714dd/cmd/commandfuncs.go#L506

Should be simple fix for someone who already has a dev environment setup to test it out before committing (hence I'm creating an issue and not PR)

pistasjis:
This PR attempts to fix #5612 by requiring a config file for ``caddy validate``, unless there is a Caddyfile in the directory that the user is running the command from.


mholt (commented):
Thank you for the patch!

Since this is the same as in the `adapt` command (and maybe others? haven't looked thoroughly yet), maybe we should refactor this logic into a function that can be reused. What do you think?

pistasjis:
That's not a bad idea.

pistasjis:
What should I name the function? And should it include both of those snippets or just one of them?

mholt:
Hmm, maybe like, `detectAdjacentCaddyfile` or something?

Its body would basically be the snippet you copied from above (and thus you can replace both).

pistasjis:
Done, but I'm not sure if the code looks good or not.

cmd/commandfuncs.go
```diff
@@ -416,30 +416,38 @@ func cmdEnviron(_ Flags) (int, error) {
...
+	adaptCmdValidateFlag := fl.Bool("validate")
+
+	adjacent, err := detectAdjacentCaddyfile(adaptCmdInputFlag)
+	if adjacent != caddy.ExitCodeSuccess {
+		return adjacent, err
+	} else {
+		if adaptCmdInputFlag == "" {
+			adaptCmdInputFlag = "Caddyfile"
+		}
 	}
```
francislavoie, line 449: I think we need to check for `""` first, because it shouldn't be an error if the user provided an input and doesn't have an adjacent Caddyfile.

```suggestion
	if adaptCmdInputFlag == "" {
		adjacent, err := detectAdjacentCaddyfile(adaptCmdInputFlag)
		if adjacent != caddy.ExitCodeSuccess {
			return adjacent, err
		} else {
			adaptCmdInputFlag = "Caddyfile"
		}
	}
```

pistasjis: Commited

mholt:
Thank you @pistasjis for the revisions. I started a final review but it got too complicated to explain so I just went ahead and pushed a refactor that is a little more like what I had envisioned!

We are currently in a feature freeze, but given that this is a bug fix, has minimal impact (two non-critical CLI commands only, and I did test that they still work as expected), and is fairly simple, I'm OK with merging this in whenever.

mholt (approved):

PR #5623, opened 2023-07-08
caddytls: Reuse certificate cache through reloads

mholt:
Previously, Caddy created a new certificate cache for every config load. This ensured that the state of the certificate cache was consistent with the newly-loaded config, no matter what changes were made.

However, this is a bottleneck for some users with large TLS deployments, some of which exceed 100K sites/certificates, as the entire cache was GC'ed and a new one had to be created by loading each certificate from storage and decoding it (not to mention allocating memory for this, which is not cheap -- certificate encodings are messy!) up to the cache capacity, which is configurable to an arbitrarily large size.

On-demand TLS can be used to relieve this pressure, as certificates are only loaded into the cache, when handshakes come in. But for busy servers, that turns out to be almost instantly. And it hammers the 'ask' endpoint which is now required, and we have [received complaints about this](https://caddy.community/t/caddy-2-7-0-beta-1/20255/13) in the beta version.

So, this is not a very small change, but it's not as big as I thought it'd be, to use only 1 certificate cache, and to persist it through config reloads.

When a config using the TLS app is loaded, the TLS app is provisioned and started; this loads certificates into the cache which are not already loaded or which are manually-loaded. We now keep a list of which subjects/certificates a config loads. This adds slightly more memory overhead but insignificant relative to the actual certificates.

Then, when a TLS app is "cleaned up" (decommissioned in preparation for being garbage-collected), it loads the currently-active Caddy context which has a reference to the new/current TLS app. Any certificates/subjects that are listed only in the "old" TLS app are removed from the cache.

This does have the side-effect that managed certificates can't be forcefully-reloaded from storage unless the config is completely unloaded first (or the server is stopped) and then reloaded. This is because managed certificates don't get evicted from the cache if the next config is also using them; however, when those certificates come due for renewal, the new config will apply when renewing them.

Manually-loaded certificates are still reloaded from storage at every config reload, because the server does not manage them, config reloads are how you tell the server that you updated the certificate. (This is the "classic" method of using TLS certificates with traditional web servers: update the cert on disk, then reload the server config to apply the change.) So if you are manually loading thousands of certificates, this change won't help you much. But then why are you using Caddy...

Anyway, I still need to finish this and do more testing to ensure it actually works. My hunch is this should drastically lighten config reloads when there's lots of certificates.


francislavoie:
> This does have the side-effect that managed certificates can't be forcefully-reloaded from storage unless the config is completely unloaded first (or the server is stopped) and then reloaded.

I wonder if we could parameterize this, i.e. adding some optional flags to reloading to change behaviour. Something for later. Would mean HTTP headers in the API, possibly env vars and/or CLI flags for the reload command, then passing through those flags in context for provision/cleanup/start/stop 🤔 that would let users do a graceful reload with a forced cert reload.

mholt:
> I wonder if we could parameterize this, i.e. adding some optional flags to reloading to change behaviour. Something for later. 

We'll see what the need is first. I'm not sure what the use case would be for that kind of requirement at this stage.

PS. Just pushed a commit that updates CertMagic so we don't reload managed certs from storage that are already in the cache :sweat_smile:

modules/caddytls/tls.go
```diff
@@ -305,16 +324,44 @@ func (t *TLS) Stop() error {
...
-
 	// stop the session ticket rotation goroutine
 	if t.SessionTickets != nil {
 		t.SessionTickets.stop()
 	}
 
+	// if a new TLS app was loaded, remove certificates from the cache that are no longer
+	// being managed or loaded by the new config; if there is no more TLS app running,
+	// then stop cert maintenance and let the cert cache be GC'ed
+	if nextTLS := caddy.ActiveContext().AppIfConfigured("tls"); nextTLS != nil {
```
francislavoie, line 335: Oh I like this, that's nice 👍

modules/caddytls/tls.go
```diff
@@ -305,16 +324,44 @@ func (t *TLS) Stop() error {
...
+		}
+		for hash := range t.loaded {
+			if _, ok := nextTLSApp.loaded[hash]; !ok {
+				noLongerManaged = append(noLongerLoaded, hash)
+			}
+		}
+
+		certCacheMu.RLock()
+		certCache.RemoveManaged(noLongerManaged)
+		certCache.Remove(noLongerLoaded)
```
francislavoie, line 355: These names aren't symmetrical. Call the other one `RemoveUnmanaged`?

mholt: True! I did at first, but realized that technically this would remove any certificate by that hash, whether it's managed or not.

In practice, I suppose you'll only have the hash of unmanaged certificates since the managed ones don't expose that (unless you manually compute that yourself)...

I'll think about it!

modules/caddytls/tls.go
mholt: Yes! I was pleased I didn't have to keep pointers to app modules laying around. But I had to remove a lock within ActiveContext() to avoid deadlock.

I'm not 100% certain that lock was needed but I suspect it may have been. Race detector didn't catch anything yet. I will try to test more...

mholt:
Oh, and minor implementation detail: this change uses [blake3](https://github.com/BLAKE3-team/BLAKE3) to hash certificates, instead of sha256.

In my testing, Blake3 was nearly twice as fast, but with 2 allocations instead of 1:

```
goos: linux
goarch: amd64
pkg: hashtest
cpu: AMD Ryzen 9 3900X 12-Core Processor            
BenchmarkBlake3-24        769701              1723 ns/op              64 B/op          2 allocs/op
BenchmarkSHA256-24        438304              2645 ns/op              32 B/op          1 allocs/op
```

(I think 64 B/op is still acceptable though.)

mholt:
@francislavoie I think I'm ready to merge this; want to do any final review or does it LGTY?

francislavoie (approved):
Yep LGTM. Obviously there's still the question of the ActiveContext mutex but 🤷‍♂️ I haven't dug into it either

Issue #5621, opened 2023-07-07
Adapters are counted as standard modules


mholt:
Because [we use a bit of a hack](https://github.com/caddyserver/caddy/blob/7914ba3573fa8eeb05f193fffde1a1c1493e3d0b/caddyconfig/configadapters.go#L121-L137) to make config adapters count as modules, these plugins are incorrectly counted as "standard" (i.e. bundled into the core distribution from this repo).

Since config adapters aren't actually config modules, we should probably not count them as such.

And maybe `list-modules` should be `list-plugins`.

I dunno. Thoughts?

mholt:
@mohammed90 I'm now reconsidering this. I don't think we should count them as normal config modules (because they are not).

Instead, let's reveal which adapters are installed by setting the Accept header in our admin API's `/load` endpoint.

mohammed90:
> [@mohammed90](https://github.com/mohammed90) I'm now reconsidering this. I don't think we should count them as normal config modules (because they are not). Maybe we should rip out the linked code (if possible, if you agree).
> 
> Instead, let's reveal which adapters are installed by setting the Accept header in our admin API's `/load` endpoint.

I'm not sure I follow. Do you mean it'll be in the response `Accept` header when the user calls the endpoint with either GET or HEAD? For reference, we added that code because it wasn't possible to know which adapters are in the build without it.

mohammed90:
> POST with your Content-Type that you want to use

Not sure if you remember, but you implemented support for `Content-Type: text/<adapter-name>`. I like the `HEAD` idea. I still think we should have something in the CLI. Should we add a command `list-adapters`? We probably also should list the adapters in the help text of the `--adapter` flag and the `adapt` command,

mholt:
> Not sure if you remember, but you implemented support for Content-Type: text/<adapter-name>. 

Yeah, but that doesn't tell you what adapters are available. I just am suggesting that if you want to know if your preferred adapter is supported, you can try POSTing with it to begin with, instead of always doing a HEAD first.

What if we put the list of adapters in `build-info`?

mohammed90:
> if your preferred adapter is supported, you can try POSTing with it to begin with, instead of always doing a HEAD first.

That feels wrong. The HEAD idea is very cool. I don't know what I feel about `build-info`. It doesn't seem like the right place.

mholt:
> That feels wrong. The HEAD idea is very cool.

[...]

PR #5624, opened 2023-07-10
cmd: Unwrap adapter modules to get underlying modules

WeidiDeng:
Fix [5621](https://github.com/caddyserver/caddy/issues/5621).

In addition, this also fixed several `add(remove)-package` related commands as they can't be used to add or remove adapter related modules.

Before,

image

After,

image

### DIFF
--- a/caddyconfig/configadapters.go
+++ b/caddyconfig/configadapters.go
@@ -135,4 +135,11 @@ func (am adapterModule) CaddyModule() caddy.ModuleInfo {
 	}
 }
 
+// UnwrapAdapter return the original Adapter, this method must be exported
+// to be type-assertable 🤷
+// hack, https://github.com/caddyserver/caddy/issues/5621
+func (am adapterModule) UnwrapAdapter() any {
+	return am.Adapter
+}
+
 var configAdapters = make(map[string]Adapter)
--- a/cmd/packagesfuncs.go
+++ b/cmd/packagesfuncs.go
@@ -174,6 +174,13 @@ func upgradeBuild(pluginPkgs map[string]struct{}, fl Flags) (int, error) {
 	return caddy.ExitCodeSuccess, nil
 }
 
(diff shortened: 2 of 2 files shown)


mholt (changes requested):
Oh, thanks for this!

I was kind of thinking that we should rethink this command...

Technically, by "modules" I mean "config modules" i.e. an extension of the configuration.

I wonder if this command should actually be "list-plugins" to list all the plugins.

Or maybe the semantics shouldn't matter and we should just print everything.

The problem is that config adapters don't keep the same interface as a Caddy module... so we'd have to figure that out.

Can we talk about a fix where we don't wrap the adapters in the first place? I think I'd like that better. :grin:

WeidiDeng:
You mean adapters are not part of modules, and will not show as so. However, they will show in the new `list-plugin` command?

They are of course affected by add / remove package commands.

WeidiDeng:
What should be the definition of a plugin? Maybe just `config-adapters` are more expressive?

mholt:
Well, I'm not sure if a new `list-plugins` command is the right answer -- maybe it is. I need to think about it more, or maybe have a discussion about other good solutions.

mholt:
Revisiting this...

>  this also fixed several add(remove)-package related commands as they can't be used to add or remove adapter related modules.

Are you sure this doesn't work? It should still "plug in" the adapters and download the resulting build, even if they don't appear in the list.

I still hesitate to print adapters as if they were config modules with names like `caddy.adapters.*` because they aren't config modules. They're plugins, at least... but the list-modules command is for config modules.

I think perhaps there is a solution, but probably not this PR. I hope that's OK. Thank you for working on this nonetheless.

What I think we should do instead, to describe the config adapters supported by a Caddy build, is to properly set the Accept header in the admin API, since the Content-Type header is what determines which config adapters to use.

PR #5630, opened 2023-07-13
fileserver: add `export-template` sub-command to `file-server`

mohammed90:
I've seen a few requests for guidance on how to customize the default file browser template. The template file is stored in this repository. Users might find the reference to the template [here](https://caddyserver.com/docs/caddyfile/directives/file_server#browse) in the documentation of the `browse` sub-directive of the `file_server` directive, which many users seem to miss or glance over. Adding the sub-command allows the user to not leave the terminal or search through the interwebz on how to find the template. They can run `caddy file-server export-template` right from the terminal where they are. This feels more ergonomic.

If accepted, I'll update the website docs accordingly.

### DIFF
--- a/modules/caddyhttp/fileserver/command.go
+++ b/modules/caddyhttp/fileserver/command.go
@@ -16,7 +16,9 @@ package fileserver
 
 import (
 	"encoding/json"
+	"io"
 	"log"
+	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mholt (approved):
LGTM -- nice idea. Thanks!

PR #5636, opened 2023-07-17
chore: update quic-go to v0.36.2

bt90:
https://github.com/quic-go/quic-go/releases/tag/v0.36.2


mholt (approved):
Thanks! I think we'll be having another upgrade here shortly just FYI.

PR #5642, opened 2023-07-20
go.mod: Upgrade logrus

jrschumacher:
Fixes DOS risk and panic in writer https://github.com/sirupsen/logrus/commit/c052ba6a076b368de89029949f68b3b8ccd8e058

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -113,7 +113,7 @@ require (
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/shopspring/decimal v1.2.0 // indirect
 	github.com/shurcooL/sanitized_anchor_name v1.0.0 // indirect
-	github.com/sirupsen/logrus v1.9.0 // indirect
+	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/slackhq/nebula v1.6.1 // indirect
 	github.com/spf13/cast v1.4.1 // indirect
 	github.com/stoewer/go-strcase v1.2.0 // indirect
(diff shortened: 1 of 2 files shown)


mohammed90:
Thanks. However, we don't use logrus.

jrschumacher:
@mohammed90 its listed as an indirect in the go.mod today. This is part of your supply chain according to the go.sum

https://github.com/caddyserver/caddy/blob/master/go.mod#L127
https://github.com/caddyserver/caddy/blob/master/go.sum#L901-L907

mohammed90:
> @mohammed90 its listed as an indirect in the go.mod today. This is part of your supply chain according to the go.sum
> 
> https://github.com/caddyserver/caddy/blob/master/go.mod#L127
> https://github.com/caddyserver/caddy/blob/master/go.sum#L901-L907
> 
> 

Yes, it's amongst the transitive/indirect deps, but it isn't in any execution path. The go modules logic lists all the transitive deps, whether they're in the execution path or not for completion sake; but it doesn't mean they're used.

jrschumacher:
The problem is that users of Caddy who also have image scanning tools like PaloAlto Prisma Cloud will get security violations due to the supply chain risks. Admittedly, a risk like the one above may never be exposed to Caddy, but the tools don't know that.

The reality is that users like me will start phasing off Caddy because we cannot get support from our Security groups to permit life-long wavers. I understand you cannot accommodate all customer personas, but SMEs and Enterprises will run into this more frequently.

As an example: 
CleanShot 2023-07-20 at 13 13 29

francislavoie:
It gets pulled in via `github.com/smallstep/certificates/acme/api` but it doesn't get used because we only use smallstep for the API and not their CLI tooling (which does call down to `logrus`).

I'd rather let smallstep update their dependency then we'll receive the update that way.

mholt:
> The problem is that users of Caddy who also have image scanning tools like PaloAlto Prisma Cloud will get security violations due to the supply chain risks.

That sounds like a bug in the scanning tools.

PR #5646, opened 2023-07-22
fileserver: add lazy image loading

ydylla:
This adds lazy image loading to the grid view mode of the fileserver, to reduce traffic on large listings.


emilylange (approved):

mholt (approved):
Good idea, thanks!

PR #5648, opened 2023-07-22
export reverseproxy.ipVersions type

kkroo:
allows AUpstreams to be externally instantiable for use within plugins. otherwise the first call to `GetUpstreams` will panic in dereferencing a nil value

### DIFF
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -1449,7 +1449,7 @@ func (u *AUpstreams) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				}
 
 				if u.Versions == nil {
-					u.Versions = &ipVersions{}
+					u.Versions = &IPVersions{}
 				}
 
 				trueBool := true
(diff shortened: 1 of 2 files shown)


francislavoie:
What kind of plugin are you writing that would need this? 🤔

kkroo:
dynamic configuration for a reverseproxy

francislavoie:
That doesn't really answer my question. Why are you calling this module from yours? What's the usecase? What problem are you solving?

kkroo:
I'm generating JSON configurations for a frontend reverse proxy, and the upstream uses an A record that is dynamic, so am using the `dynamic_upstreams` module to have control over the DNS resolver and caching. because the type is private, its not possible to instantiate a value and marshal the config correctly

mholt:
Interesting -- can you show us the code that needs this type? I'd rather keep it unexported if possible.

mholt (changes requested):
If we do export this, the change LGTM mostly.

kkroo:
thanks for the review. 
here is a snippet of how this is being used:

```
proxyHandle := &reverseproxy.Handler{}
...
proxyHandle.DynamicUpstreamsRaw = caddyconfig.JSONModuleObject(
	reverseproxy.AUpstreams{
		Name:     upstreamHost,
		Port:     upstreamPort,
		Versions: &reverseproxy.IPVersions{IPv4: yes, IPv6: yes},
	}, "source", "a", &warn)
```

mholt (approved):
I don't _fully_ understand the use case for using the upstreams in other packages, but I think it's a good idea to export this anyway since it's used in an exported type. Hope it is useful :)

Thanks for the contribution!

Issue #5431, opened 2023-03-11
Caddy prematurely closes connections to reverse proxies when using HTTP/1


supriyo-biswas:
## Details

* Caddy version: 2.6.4 (0db29e2ce9799f652f3d16fd5aed6e426d23bd0a (14 Feb 23 19:08 UTC))
* Operating system: Ubuntu 20.04, amd64

## Description

I had initially reported this issue with v2.6.2 here (https://caddy.community/t/caddy-prematurely-closes-connections-to-reverse-proxies-when-using-http-1/17974), which still persists despite the bugfixes to reverse_proxy in 2.6.4, so I'm creating this Github issue.

I have the following Caddyfile that reverse proxies connections to a locally running HTTP/1 API server at port 2000. Whenever it receives a POST request, it sends a 200 OK with Transfer-Encoding: chunked, and streams its response bit by bit, in delays of 2-3 seconds.

```
my.web.site {
    handle_path /api/* {
        reverse_proxy http://127.0.0.1:2000
    }
}
```

When I use HTTP/1 to make requests, I’ll sometimes get a truncated response that only contains the first chunk produced by the API server, but not the rest. However, this issue does not occur on HTTP/2.

A script like the one below can be used to reproduce the issue:

```
expected_output = '...'
for i in {1..200}; do
    curl --http1.1 -sSi https://my.web.site/api/some/path -X POST --data '...' > output.txt
    if ! grep -Fq "$expected_output" output.txt; then
       echo "Received truncated output"
       break
    fi
done
```

Looking at a packet capture taken on `lo` on TCP port 2000, (`sudo tcpdump -i lo -w my.pcap -s 0 'tcp and src or dst port 2000'), I see that the API server had just sent the first chunk of data. However, Caddy proceeded to acknowledge the received data, which is fine, but then it also proceeds to request a closure of the TCP connection. The FIN, ACK is reciprocated by the app server, and the connection is closed.

<img width="1391" alt="Screenshot 2023-03-12 at 12 44 54 AM" src="https://user-images.githubusercontent.com/27079933/224507472-b6d26b4c-c0b9-4f1b-b74a-df3669b11c45.png">

mholt:
Hi, thanks for the issue. In order to reproduce this, can you please provide the backend code that produces the response that Caddy has trouble proxying?

francislavoie:
[...]

Issue #5288, opened 2023-01-06
Go 1.20 TODO list


francislavoie:
Did some reading of the upcoming Go 1.20 release notes, and I'm writing down thing things we should consider doing once we've bumped to 1.20 minimum (probably ~1 year from now) https://tip.golang.org/doc/go1.20

- [x] https://tip.golang.org/doc/go1.20#http_responsecontroller would be useful to add timeouts to individual requests. I'm thinking this would be new options for the [request_body](https://caddyserver.com/docs/caddyfile/directives/request_body) directive.
- [ ] https://tip.golang.org/doc/go1.20#reverseproxy_rewrite Go stdlib's proxy implementation is getting some changes relating to how headers are handled via a new API for transforming the request before sending it upstream; the linked github issue in that section has some interesting discussion that should get reviewed to see if there's anything we need to action.
- [ ] https://tip.golang.org/doc/go1.20#context `WithCancelCause` is a really neat new feature, there's some places we could probably make use of this to smooth things over, for example a context cancellation due to a config reload could allow for some additional intelligence in some areas by allowing parts of the config to compare old vs new to see if the config significantly changed, and if it hasn't, ignore the cancellation maybe. For example, frequent config reloads right now cause in-flight TLS automation to get cancelled, which adds friction, can continually prevent TLS automation from completing if reloads are too frequent. It could prevent websocket connections from being closed maybe, if only unrelated config is changed. Obviously this is much easier said than done, but just some thinking :brain: 
- [ ] https://tip.golang.org/doc/go1.20#crypto/tls Looks like we'll get some free optimizations, mostly for HTTPS proxy upstreams, by having certs shared in memory for all connections. Also, there's a new `CertificateVerificationError`; we should probably detect those and write out a DEBUG level log with the additional details we can get from that error, it would be helpful for diagnosing some issues.
- [x] https://tip.golang.org/doc/go1.20#crypto/x509 We may want to consider using the new `SetFallbackRoots` to bundle a reasonable default set of root CAs, mainly for Let's Encrypt and ZeroSSL, so that TLS automation can still work without having a system root store. This would be mainly for `FROM scratch` Docker containers, although our official image is based on `alpine` which does ship a trust store via the `ca-certificates` package so it's only relevant for users making their own Docker images instead of using ours.
- [ ] https://tip.golang.org/doc/go1.20#net/http The new `StreamError` would probably be useful to detect and write out at DEBUG level for diagnosis.
[...]

PR #5654, opened 2023-07-26
caddyhttp: Make use of `http.ResponseController`

francislavoie:
One part of https://github.com/caddyserver/caddy/issues/5288, should fix https://github.com/caddyserver/caddy/issues/5431 (if built with Go 1.21+).

Adds a new `enable_full_duplex` server global option to opt-in (only effectual for Go 1.21+, with a `go:build` constraint). I think it needs to be opt-in because it could cause deadlocks if the client doesn't know how to handle a connection in that mode, but it seems necessary to turn on for certain kinds of apps. Maybe this should be an HTTP handler (maybe as a new option for https://caddyserver.com/docs/caddyfile/directives/request_body) which would allow using matchers to conditionally enable it instead of enabling it for all connections to the server :thinking: But I marked this as experimental so we can adjust later if needed.

The other main aspect of this is dropping `caddyhttp.HTTPInterfaces` because it's now made redundant by `http.ResponseController` which gives a way to use these optional interfaces without type assertions. Simplifies code in most cases.

Also syncs the reverseproxy implementation with stdlib's which now uses ResponseController as well; see golang/go@2449bbb for the original diff that I ported over.

And finally, minor improvement, `io.Copy()` already tries `io.ReaderFrom` so we don't need to do that ourselves in `responsewriter.go`.


francislavoie:
Linter doesn't like this new pattern. Opened an issue: https://github.com/timakin/bodyclose/issues/52

Also the Go stdlib code we ported ignores the `err` returned by `m.flush()`, so I need to suppress it to keep the code the same :man_shrugging:

modules/caddyhttp/server.go
```diff
@@ -82,6 +82,12 @@ type Server struct {
 	// HTTP request headers.
 	MaxHeaderBytes int `json:"max_header_bytes,omitempty"`
 
+	// Enable full-duplex communication for HTTP/1.1 requests.
+	// Only has an effect if Caddy was built with Go 1.21 or later.
```
mholt, line 86: Should we mention that this can brick/hang clients if they don't support this?

mholt (commented):
Thanks, this definitely looks more elegant than what we were doing before.

I imagine there aren't external repos using our HTTPInterfaces type?

modules/caddyhttp/server.go
francislavoie: Probably. But I'm not sure I understand well the circumstances where it can happen.

mholt: What if we say something like: "Test thoroughly with your client, as some clients may not support full-duplex HTTP/1.1 which can cause them to deadlock."

francislavoie: Done, also pulled in the text from Go stdlib which clarifies what it does.

mholt (approved):
LGTM now. Thank you for contributing this!

francislavoie:
I'd just like @WeidiDeng to give his 👍 on this as well in case I missed something.

WeidiDeng:
> I'd just like @WeidiDeng to give his 👍 on this as well in case I missed something.

There is something about `encode` that I don't like, can it be type asserted just like before to avoid wrapping a `caddyhttp.ResponseWriterWrapper`?

francislavoie:
Hmm. I'm not sure what we would type assert at all. We already know it's a ResponseWriter. Do we even need wrapping now that response controller lets us call those optional interfaces in a unified way? Does our wrapper add any value, or is it necessary for it to work properly?

If you have a suggestion, feel free to commit it 🤔 I'm not sure I follow.

francislavoie:
Aaah I see, so don't re-wrap if it's already wrapped. Makes sense.

francislavoie:
Thanks, pushed that fix. Do you notice anything else?

mholt:
I'll probably merge this in a few minutes to a few hours (before I have to take off for some family stuff today) so I can tag 2.7. Thank you for working on this!

mholt:
And @WeidiDeng thank you for your invaluable review and contributions :pray:

PR #5664, opened 2023-07-31
core: Allow loopback hosts for admin endpoint (fix #5650)

### LINKED ISSUE #5650: go 1.20.6 host validation breaks setting Host to a unix socket address thus `caddy reload` over UDS
For those using caddy admin API over a unix domain socket, `caddy reload` now fails with `Error: sending configuration to instance: performing request: Post "http://%20/load": http: invalid Host header`. See upstream issue https://github.com/golang/go/issues/61431

mholt:
This should actually be a non-breaking change, since previously we only allowed the empty Host for unix sockets. We still allow that, but turns out the most popular clients used with the admin API (Caddy's CLI and curl) don't allow empty hostnames anymore:

- Caddy's CLI starting with Go 1.20.6 (https://github.com/golang/go/issues/61431)
- Curl starting with 7.50 (https://superuser.com/a/925610) (but it worked starting with 7.40)

Anyway, this is mildly infuriating because we can no longer be compliant with RFC 2616 Section 14.26 as Go requires setting a host. But it's the best we can do.

We could disable the security checks completely for unix sockets -- and the [infosec community has assured me](https://infosec.exchange/@mholt/110794015397914193) that doing so would be safe -- but I am not ready to trust that, for instance, browsers won't always forbid access to unix sockets in the future (or maybe they have a bug). But as long as they enforce CORS we can at least have reasonable assurance that a host of `127.0.0.1` or `::1` mean the request comes from a local website. Additionally, unix sockets can have system permissions used to restrict access.

In #5650 I decided that we'd use 127.0.0.1 as a "bogus" host and continue to enforce our host/origin checks, while allowing 127.0.0.1 and ::1 as possible values.

I think this is relatively conservative, so if there's any major issues with good technical reasoning we can reconsider this approach if need be.

Ultimately I view this as a very minor regression due to an upstream patch that we're working around.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -318,7 +318,32 @@ func (admin AdminConfig) allowedOrigins(addr NetworkAddress) []*url.URL {
 				// messages. If the requested URI does not include an Internet host
 				// name for the service being requested, then the Host header field MUST
 				// be given with an empty value."
+				//
+				// UPDATE July 2023: Go broke this by patching a minor security bug in 1.20.6.
+				// Understandable, but frustrating. See:
+				// https://github.com/golang/go/issues/60374
+				// See also the discussion here:
+				// https://github.com/golang/go/issues/61431
+				//
+				// We can no longer conform to RFC 2616 Section 14.26 from either Go or curl
+				// in purity. (Curl allowed no host between 7.40 and 7.50, but now requires a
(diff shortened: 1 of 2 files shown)


mholt:
Thank you for confirming!

Issue #5673, opened 2023-08-03
Unable to compile after adding caddy-l4 plugin


lxhao61:
It is found that the source code before [5c51c1d](https://github.com/caddyserver/caddy/commit/5c51c1db2ce450a3fa003834097ad010b3844673) plus the caddy-l4 (default master branch) plug-in is compiled normally; but the source code of [5c51c1d](https://github.com/caddyserver/caddy/commit/5c51c1db2ce450a3fa003834097ad010b3844673)  and later plus the caddy-l4 (default master branch) plug-in is compiled incorrectly, and the error report is as follows:
![image](https://github.com/caddyserver/caddy/assets/24571146/6700f587-0e80-4710-ae4a-826bb43a2114)

hslatman:
Hey @lxhao61,

You seem to have hit the same issue as me: https://github.com/caddyserver/caddy/issues/5674.

PR with fix: #5675

lxhao61:
@hslatman 
seems like it.
Hope your fix works.

lxhao61:
@hslatman @mholt 
Didn't succeed to fix the problem I found, the problem is still the same.

PR #5675, opened 2023-08-03
Fix `string does not match ~[]E` error

### LINKED ISSUE #5674: Custom Caddy build with module relying on `httpcaddyfile` doesn't compile
After upgrading to `v2.7.0` and running CI with Go 1.20.7, I get the following error:

```
Error: ../../../go/pkg/mod/github.com/caddyserver/caddy/v2@v2.7.0/caddyconfig/httpcaddyfile/httptype.go:358:24: string does not match ~[]E
```

Also see: https://github.com/hslatman/caddy-crowdsec-bouncer/actions/runs/5744901364/job/15572008214?pr=18#step:5:167

I see the same behavior on my mac.

The change seems to have been introduced in https://github.com/caddyserver/caddy/commit/5c51c1db2ce450a3fa003834097ad010b3844673.

hslatman:
Fixes #5674

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -355,7 +355,7 @@ func (st ServerType) Setup(
 				defaultLog.Exclude = append(defaultLog.Exclude, ncl.log.Include...)
 
 				// avoid duplicates by sorting + compacting
-				slices.Sort[string](defaultLog.Exclude)
+				sort.Strings(defaultLog.Exclude)
 				defaultLog.Exclude = slices.Compact[[]string, string](defaultLog.Exclude)
 			}
 		}


mholt:
What is this exactly? I do not get this error, nor does our CI, and builds succeeded.

hslatman:
Yea, I was kind of surprised myself too.

I hit this error when building Caddy with one of my modules. Another person seems to have hit the same issue with a custom module: #5673

mholt (approved):
Huh. Alrighty then.

LGTM. I'm not sure why we need the exp package here anyway. Thanks

francislavoie:
Does the import need updating? (Run go mod tidy)

hslatman:
> Does the import need updating? (Run go mod tidy)

Doesn't look like it. Still one `slices` usage.

mholt:
Weird that only this occurrence needs changing then.

As I do not get this error, nor does our CI, I'm trusting you on this 😊

hslatman:
Here's a reproducer on the playground: https://go.dev/play/p/NsaynrOjz9i

mholt:
Is that using the same commit of x/exp that is in our go.mod?

francislavoie:
Oh, I guess it needs to be `slices.Sort[[]string](s)` i.e. with the `[]`. Did go 1.20 not error on that? Strange.

hslatman:
> Oh, I guess it needs to be `slices.Sort[[]string](s)` i.e. with the `[]`. Still don't understand why it didn't fail for us before though.

That does compile on the playground 🙂

mholt:
If the playground always uses the latest commits, ours is not using the latest commit of x/exp. That might be the difference, somehow?

hslatman:
Maybe this commit is related: https://github.com/golang/go/commit/0a48e5cbfabd679eecdec1efa731692cd6babf83

PR #5679, opened 2023-08-03
reverseproxy: Fix hijack ordering which broke websockets

### LINKED ISSUE #5678: v2.7.1: websockets randomly work or don't
Neither Caddy nor my web browser seem to provide any insight into this issue. Caddy logs show nothing relevant (perhaps I need to configure something more verbose/debuggy?) and my browser just Eventually™ says it can Never™ connect to wss://whatever after a long period of time.

---

It seems that in `v2.7.1` it's totally random whether websockets work for any given `reverse_proxy`.

For me: neither my [Send](https://github.com/timvisee/send) (ws for file uploads) nor [TeslaMate](https://github.com/adriankumpf/teslamate) (ws streams in map tiles at vehicle's location in real-time) websockets work. But [The Lounge](https://github.com/thelounge/thelounge)'s websockets work fine for me.

For a buddy: his The Lounge's websockets do not work at all.

---

For both of us: Everything works flawlessly upon rolling back to `v2.7.0-beta.2` (with the same Caddyfile that doesn't seem to work on `v2.7.1`). I don't know what change could cause this random behavior.

WeidiDeng:
Fix #5678

### DIFF
--- a/modules/caddyhttp/reverseproxy/streaming.go
+++ b/modules/caddyhttp/reverseproxy/streaming.go
@@ -58,10 +58,22 @@ func (h *Handler) handleUpgradeResponse(logger *zap.Logger, rw http.ResponseWrit
 		return
 	}
 
+	// write header first, response headers should not be counted in size
+	// like the rest of handler chain.
+	copyHeader(rw.Header(), res.Header)
+	rw.WriteHeader(res.StatusCode)
+
+	logger.Debug("upgrading connection")
+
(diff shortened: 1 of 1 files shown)


francislavoie (approved):
Thanks! And sorry :see_no_evil:

mholt:
Thank you for the patch. And I'm sorry for merging the other PR when I did -- even though it was said, I just didn't realize how much more testing was still needed.

PR #5685, opened 2023-08-04
import: check that matched key is not a substring of the replacement key

WeidiDeng:
Fix [issue reported on community](https://caddy.community/t/problem-after-upgrading-to-2-7-2/20687).

When importing arguments, regex will usually match a substring of the key, this change detects if it's a substring match.

Personally I think caddy shouldn't allow braces in the replacer as it has other implications. But this will do for now.


francislavoie (approved):
Thanks! Glad you could find this :relaxed:

WeidiDeng:
I mean disallow keys like `test{test`, the underlying problem is replacer feed the charactes between the opening brace and the first closing brace, and there can be opening braces in between.

Normally this isn't a problem, but in JSON, it is. And if replacer is configured to return an error upon an unknown key, it [will](https://github.com/caddyserver/caddy/blob/9f34383c02f1691e54280285a6499893fcbbb4c7/replacer.go#L195).

PR #5695, opened 2023-08-05
reverseproxy: do not parse upstream address too early if it contains replaceble parts

### LINKED ISSUE #5688: reverse_proxy upstream directive interprolation throwing `strconv.ParseUint`
Hey team,

We are currently facing an issue with the Caddy in our docker compose stacks, currently using the Caddy Docker Image pinned on version '2'.

We see the following in our logs: `Error during parsing: parsing caddyfile tokens for 'reverse_proxy': /etc/caddy/Caddyfile:26 - Error during parsing: invalid start port: strconv.ParseUint: parsing "6{http.regexp.port.1}"`

It appears that the Interpolation for the `Matcher` / `Upstream` is not functioning as intended.

Caddyfile:
```
*.sandbox.application.localhost {
	@sandboxPort {
		header_regexp port Host ^([0-9]{3})\.sandbox\.
	}

	handle @sandboxPort {
		reverse_proxy app:6{re.port.1}
	}

	handle {
		redir {scheme}://application.localhost
	}
}
```

As a quick fix, we have pinned our docker compose files to version 2.6.4 (where is issue is not present).

Regards

mohammed90:
Fixes #5688

### DIFF
--- a/modules/caddyhttp/reverseproxy/addresses_test.go
+++ b/modules/caddyhttp/reverseproxy/addresses_test.go
@@ -265,18 +265,21 @@ func TestParseUpstreamDialAddress(t *testing.T) {
 			expectScheme:   "h2c",
 		},
 	} {
-		actualHostPort, actualScheme, err := parseUpstreamDialAddress(tc.input)
+		actualHostPort, err := parseUpstreamDialAddress(tc.input)
 		if tc.expectErr && err == nil {
 			t.Errorf("Test %d: Expected error but got %v", i, err)
 		}
 		if !tc.expectErr && err != nil {
 			t.Errorf("Test %d: Expected no error but got %v", i, err)
 		}
-		if actualHostPort != tc.expectHostPort {
-			t.Errorf("Test %d: Expected host and port '%s' but got '%s'", i, tc.expectHostPort, actualHostPort)
+		if actualHostPort.dialAddr() != tc.expectHostPort {
+			t.Errorf("Test %d: input %s: Expected host and port '%s' but got '%s'", i, tc.input, tc.expectHostPort, actualHostPort.dialAddr())
 		}
-		if actualScheme != tc.expectScheme {
-			t.Errorf("Test %d: Expected scheme '%s' but got '%s'", i, tc.expectScheme, actualScheme)
+		// if actualHostPort != tc.expectHostPort {
+		// 	t.Errorf("Test %d: Expected host and port '%s' but got '%s'", i, tc.expectHostPort, actualHostPort)
+		// }
+		if actualHostPort.scheme != tc.expectScheme {
+			t.Errorf("Test %d: Expected scheme '%s' but got '%s'", i, tc.expectScheme, actualHostPort.scheme)
 		}
 	}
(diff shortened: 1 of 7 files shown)


modules/caddyhttp/reverseproxy/addresses_test.go
```diff
@@ -265,18 +265,21 @@ func TestParseUpstreamDialAddress(t *testing.T) {
 			expectScheme:   "h2c",
 		},
 	} {
-		actualHostPort, actualScheme, err := parseUpstreamDialAddress(tc.input)
+		actualHostPort, err := parseUpstreamDialAddress(tc.input)
```
francislavoie, line 268: Rename to `actualAddr` or `actualAddress`?

modules/caddyhttp/reverseproxy/addresses_test.go
```diff
@@ -265,18 +265,21 @@ func TestParseUpstreamDialAddress(t *testing.T) {
...
-		if actualHostPort != tc.expectHostPort {
-			t.Errorf("Test %d: Expected host and port '%s' but got '%s'", i, tc.expectHostPort, actualHostPort)
+		if actualHostPort.dialAddr() != tc.expectHostPort {
+			t.Errorf("Test %d: input %s: Expected host and port '%s' but got '%s'", i, tc.input, tc.expectHostPort, actualHostPort.dialAddr())
 		}
-		if actualScheme != tc.expectScheme {
-			t.Errorf("Test %d: Expected scheme '%s' but got '%s'", i, tc.expectScheme, actualScheme)
+		// if actualHostPort != tc.expectHostPort {
+		// 	t.Errorf("Test %d: Expected host and port '%s' but got '%s'", i, tc.expectHostPort, actualHostPort)
+		// }
```
francislavoie, line 280: No longer needed, right?

francislavoie (commented):
I like the struct refactor! Certainly reads cleaner this way.

modules/caddyhttp/reverseproxy/addresses_test.go
mohammed90: Oops

francislavoie:
Do we have coverage for when the host or entire thing is a placeholder? I think we probably also want a test for when the port has a numerical prefix followed by a placeholder (as in the reported issue).

I think we can just put a lot of `reverse_proxy` lines in the adapt fixture with different edgecases. And maybe add some more in https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/addresses_test.go

mohammed90:
> I think we probably also want a test for when the port has a numerical prefix followed by a placeholder (as in the reported issue)

I missed that in my second refactor 🤦🏼‍♂️ 

> Do we have coverage for when the host or entire thing is a placeholder?

Added

francislavoie (approved):
LGTM but let's hear from Matt too ☺️

mholt (approved):
Ah wow, yeah this refactor is way better. And the new tests, love it.

Thank you for fixing this Mohammed :100:

PR #5696, opened 2023-08-05
cmd: fix cli when admin endpoint uses new unix socket permission format

### LINKED ISSUE #5694: admin api unix socket permissions cause config reload to fail
When adding the new #4741 socket permissions to the admin unix socket, the file permissions are correctly applied, but reloading the config (using `systemctl reload caddy` with the systemd unit installed by the [debian package](https://caddyserver.com/docs/install#debian-ubuntu-raspbian)) fails because the permissions suffix is included in the socket address the config would be written to:  
`Error: sending configuration to instance: performing request: Post "http://127.0.0.1/load": dial unix /run/caddy/admin.sock|0660: connect: no such file or directory`

Caddyfile snippet:
```
# Global options first
{
        # /run/caddy is created by config in /etc/systemd/system/caddy.service.d/xon.conf
        admin unix//run/caddy/admin.sock|0660
```

The full filesystem path to the socket is `/run/caddy/admin.sock` as expected.

Edit: version is 2.7.2

emilylange:
Fixes #5694, a bug where the following Caddyfile

```Caddyfile
{
	admin unix/admin.sock|0660
}
```

and `caddy reload --config Caddyfile`
would throw the following error instead of reloading it:

```
INFO    using provided configuration    {"config_file": "Caddyfile", "config_adapter": ""}
Error: sending configuration to instance: performing request: Post "http://127.0.0.1/load": dial unix admin.sock|0660: connect: no such file or directory
[ERROR] exit status 1
```

---

This bug also affected `caddy start` and `caddy stop`.

### DIFF
--- a/listeners.go
+++ b/listeners.go
@@ -157,7 +157,7 @@ func (na NetworkAddress) listen(ctx context.Context, portOffset uint, config net
 	// is independent of permissions bits
 	if na.IsUnixNetwork() {
 		var err error
-		address, unixFileMode, err = splitUnixSocketPermissionsBits(na.Host)
+		address, unixFileMode, err = SplitUnixSocketPermissionsBits(na.Host)
 		if err != nil {
 			return nil, err
 		}
@@ -329,7 +329,7 @@ func IsUnixNetwork(netw string) bool {
 // include write perms (e.g. `0422` or `022`).
 // Symbolic permission representation (e.g. `u=w,g=w,o=w`)
 // is not supported and will throw an error for now!
-func splitUnixSocketPermissionsBits(addr string) (path string, fileMode fs.FileMode, err error) {
+func SplitUnixSocketPermissionsBits(addr string) (path string, fileMode fs.FileMode, err error) {
 	addrSplit := strings.SplitN(addr, "|", 2)
 
 	if len(addrSplit) == 2 {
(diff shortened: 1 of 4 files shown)


francislavoie (approved):
Thanks! Makes sense to me.

I'll let Matt merge when he's ready.

listeners.go
```diff
@@ -329,7 +329,7 @@ func IsUnixNetwork(netw string) bool {
 // include write perms (e.g. `0422` or `022`).
 // Symbolic permission representation (e.g. `u=w,g=w,o=w`)
 // is not supported and will throw an error for now!
-func splitUnixSocketPermissionsBits(addr string) (path string, fileMode fs.FileMode, err error) {
+func SplitUnixSocketPermissionsBits(addr string) (path string, fileMode fs.FileMode, err error) {
```
mohammed90, line 332: Should this be moved into the `internal` package?

mholt: That's probably a good idea :+1:

mholt (changes requested):
Thanks -- I too forgot about this code path with unix sockets.

LGTM, but I like Mohammed's idea!

mholt (approved):
I just went ahead and moved it and verified it is hidden as internal, tests pass, etc. Will merge soon. :) Thank you @emilylange !!

Issue #5699, opened 2023-08-05
Significant variable naming changes and unit test code additions on the metric side.


tae2089:
Hi. I'm trying to contribute for the first time and while checking the code, I noticed that in the metric part there is no test code and the variable names are not meaningful. I think this needs to be fixed.

https://github.com/caddyserver/caddy/blob/65e33fc1ee4798bb3450f6e291bfc88404982636/internal/metrics/metrics.go#L8-L15

PR #5700, opened 2023-08-05
Significant variable naming changes and unit test code additions on the metric side

tae2089:
Hi, I'm raising a PR to add meaningful variable naming changes and unit test code on the metric side. I've also raised a git issue, which is numbered #5699.

So, I added and modified the following code.

1. Renaming a parameter variable
The parameter is named s. I suggest changing it to statusCode, because it's hard to tell if it stands for string or statusCode.

```go
func SanitizeCode(statusCode int) string {
	switch statusCode {
	case 0, 200:
		return "200"
	default:
		return strconv.Itoa(statusCode)
	}
}
```

2. Adding the SanitizeCode test code
There is no unit test code for SanitizeCode, so we added it.
```go
func TestSanitizeCode(t *testing.T) {
	tests := []struct {
		statusCode int
		want       string
	}{
		{
			statusCode: 200,
			want:       "200",
		},
		{
			statusCode: 400,
			want:       "400",
		},
		{
			statusCode: 500,
			want:       "500",
		},
	}
	for _, test := range tests {
		got := SanitizeCode(test.statusCode)
		if got != test.want {
			t.Errorf("Not same: expected %#v, but got %#v", test.want, got)
		}
	}
}
```

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -5,12 +5,12 @@ import (
 	"strconv"
 )
 
-func SanitizeCode(s int) string {
-	switch s {
+func SanitizeCode(statusCode int) string {
+	switch statusCode {
 	case 0, 200:
 		return "200"
 	default:
-		return strconv.Itoa(s)
+		return strconv.Itoa(statusCode)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mholt (changes requested):
At least one test case should cover the point of the entire function: treating 0 as a 200.

PR #5705, opened 2023-08-06
fileserver: Don't repeat error for invalid method inside error context

francislavoie:
The comment in code should be self-explanatory. Followup to https://github.com/caddyserver/caddy/pull/5167


mholt (approved):
Thanks Francis!

I guess this makes sense :thinking:

Issue #5706, opened 2023-08-06
Linting


faddat:
Gofumpt lints more tightly than gofmt + goimports, and produces more regular code. 

Today I saw @mholt's tweet about a feature freeze, and wanting to improve quality.  When faced with such situations and a go codebase, I typically break out golangci-lint.  So, here's the PR's:

* #5707 
* #5708 
* #5709 
* #5712 

They build off of one another, beginning with gofumpt.  The one that lints the tests will be the most involved. 

### Goal

So, the goal of this would be to find small bugs/errors or put tests in more places, while creating a strict format for the code that coders don't need to enfoirce.  

I can't say for sure that I can accomplish any style desire, but after a few years of doing extensive bughunting with golangci-lint and other tools, I can say that I can probably accomplish **almost** any style desire.

mholt:
Hi @faddat -- yes, absolutely we'll be happy to welcome your expertise and review changes like this. Thank you!

A couple things to note:

- Linters can be often be _too_ opinionated for our liking. Sometimes we just disagree and there's no real good technical basis, so we like to avoid changing working code just to change things.
- We have decided not to use tools that recommend upgrades without discretion (e.g. dependabot), as we prefer to judiciously upgrade dependencies when we feel it's a good time or when we have a specific need or when we know it's most likely safe to do so.
- Sometimes our code is intentionally linter-and-vet-unfriendly. For example, we have an extra line break somewhere because the file is a copy+paste template and we expect users to paste in their own line of code there. Or we refer to the long-broken, long-deprecated SSLv3 in code explicitly for the purpose of _rejecting_ it at runtime and for providing helpful error messages. Or we use a weak random number generator because we need speed and not security. Or we open an `if { }` block just to have a comment and no actual code, because it makes it clear to readers why we're doing it that way. So if we use more linters/vet tools, we need ways to quiet their warnings around intentional violations without it being obnoxious.
- During the feature freeze, we're less inclined to make sweeping changes to the code base, i.e. changes to dozens/hundreds of files is scarier than changes to just one or two. I realize that in many cases the actual changes will be very trivial, but any structural or logical modifications -- even if theoretically equivalent -- make more more nervous and will take more time to review.

I see your PRs, and now that the busy weekend is over I can start looking at them this week. Understand if it takes some time :slightly_smiling_face: But since I know almost nothing about how to set up fancy linters/vet tools, I'm glad you're helping us out here! I just may have lots of nits and questions.

Thank you for the contributions!

[...]

PR #5707, opened 2023-08-06
chore: use gofumpt

faddat:
This PR removes the gofmt and goimports linters, replacing them with gofumpt, which produces a tighter result.

PR #5708, opened 2023-08-06
chore: use gci linter

faddat:
Note: this PR builds off of #5706 and adds gci.

- use gofmput to format code
- use gci to format imports


mohammed90:
Why?

cmd/caddy/main.go
```diff
@@ -30,7 +30,6 @@ package main
 
 import (
 	caddycmd "github.com/caddyserver/caddy/v2/cmd"
-
```
francislavoie, line 33: I don't like this change, the space is intentional here, because this file is meant to be copied and edited by hand (as a manual way of building Caddy with plugins etc)

faddat: ahhh, I think this is fixable :)

faddat:
@mohammed90 

the why is basically greater standardization.  I've found that by getting pretty aggressive with linters, we get a really clean and standard codebase.

cmd/caddy/main.go
faddat: Should be all set now :)

mholt:
After #5707 was merged, there is now a conflict. Want to rebase?

faddat:
yep absolutely :)

mholt:
@faddat Thanks!!

It looks like there's a single Windows-only file that is failing some GCI lint tests. Do you have a chance to fix that up? If so, I think this is one we can merge in.

faddat:
Sure :). I bet I just need to use the right build tag on golangci-lint.

mholt:
I don't know too much about configuring linters, but my guess would be that it just didn't get formatted properly by whatever command you're running to automatically fix up the files; since it's only compiled on Windows.

faddat:
for future reference:

```bash
export GOOS=windows
golangci-lint run ./... --fix
```

francislavoie:
Can probably just do it with this one liner: `GOOS=windows golangci-lint run ./... --fix`

francislavoie (approved):
LGTM

mholt (approved):
Perfect, thanks for the maintenance on this! Let's give it a try.

PR #5709, opened 2023-08-06
chore: Lint tests

faddat:
THis PR lints the tests in caddy, too. 

* adds thelper linter (gives additional context in your editor when tests fail)
* adds revive linter
* lints the entire codebase
* adds nolintlint so that nolint statements don't hang around where they're not needed


faddat (commented):
reviews sometimes can be quite not ergonomic

mholt:
Hmmm. I disagree with the linter in some places here. I think a field name like `publicKey` is perfect because it's the final result of processing the `PublicKeys` field after provisioning from config.

faddat:
@mholt -- indeed, I disagreed with it too :). 

Please lmk if I made any changes based on its disagreeable choices.  

I basically finished this one out by disabling a few things:

- confusing-naming
- unhandled-error 
- import-shadowing

I'f you'd like any of them on, please let me know that too :)

modules/caddyhttp/push/handler.go
```diff
@@ -122,11 +122,8 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhtt
 	}
 
 	// serve only after pushing!
-	if err := next.ServeHTTP(lp, r); err != nil {
-		return err
-	}
-
-	return nil
+	err := next.ServeHTTP(lp, r)
+	return err
```
mholt, line 126: Kinda surprised no linter is saying this should simply be `return next.ServeHTTP(lp, r)`

:man_shrugging: 

Not a big deal but just something I noticed.

mholt (commented):
What's the problem with turning unused parameters into parameters named `_`? I do that sometimes, maybe it's good to be consistent...

I guess we do lose the self-documenting code a bit though.

modules/caddyhttp/push/handler.go
faddat: there may be one that'll do that for us.  I'll have a peek :)

.github/dependabot.yml
```diff
@@ -1,7 +1,11 @@
...
-    directory: "/"
+  - package-ecosystem: 'github-actions'
+    directory: '/'
     schedule:
-      interval: "monthly"
+      interval: 'weekly'
+  - package-ecosystem: 'gomod'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
```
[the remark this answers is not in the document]
faddat: ah yes, I can remove this, it is a holdover from an earlier PR :)

faddat:
@mholt -- those _'s were kinda the result of a misconfiguration.  

I think that their clarity (eg: we don't do anything with this) is less-good than the self-documenting code.  I did this on ibc-go a while ago and ended up having to seek out names a bunch.

listeners.go
```diff
@@ -392,11 +392,11 @@ func SplitNetworkAddress(a string) (network, host, port string, err error) {
 	}
 	if IsUnixNetwork(network) {
 		host = a
-		return
+		return network, host, port, nil
```
mholt, line 395: This seems like a mistake -- the returns are named.

modules/caddyhttp/headers/headers.go
```diff
@@ -132,7 +132,7 @@ type HeaderOps struct {
 }
 
 // Provision sets up the header operations.
-func (ops *HeaderOps) Provision(_ caddy.Context) error {
+func (ops *HeaderOps) Provision(ctx caddy.Context) error {
```
mholt, line 135: This variable isn't used, so I'm not sure why it needs a name. Naming it falsely implies that we're using it, whereas `_` makes it clear we're satisfying an interface IMO.

modules/caddyhttp/reverseproxy/command.go
```diff
@@ -284,12 +284,12 @@ func cmdReverseProxy(fs caddycmd.Flags) (int, error) {
 		appsRaw["tls"] = caddyconfig.JSON(tlsApp, nil)
 	}
 
-	var false bool
+	var falseBool bool
```
mholt, line 287: Why this change?

faddat: it was shadowing the name "false"

listeners.go
faddat: ahhh, so it isn't a mistake actually, it's designed to make the code clearer.  If you'd like me to change it back the way it was, just let me know.

modules/caddyhttp/headers/headers.go
faddat: OK, I can change that setting, and that'd get applied universally.

mholt: Universally for context args, or for every arg?

Because I think it could make sense for some args.

If it can't be configured in that way, I don't think that linter is helpful for us.

cmd/packagesfuncs.go
```diff
@@ -222,7 +222,7 @@ func getModules() (standard, nonstandard, unknown []moduleInfo, err error) {
 			nonstandard = append(nonstandard, caddyModGoMod)
 		}
 	}
-	return
+	return standard, nonstandard, unknown, nil
```
AnomalRoil, line 225: I'm not sure here what's the expected behaviour, but this change deviates from the existing code which does:
```suggestion
	return standard, nonstandard, unknown, err
```
Which means that if the last `caddy.Modules` we've ranged over above has returned an error when doing `GetModule`, then this will return a non-nil error with the current code, whereas with the proposed change it would still return `nil`.

Overall this seems like a deeper bug lurking...

modules/caddyhttp/fileserver/matcher.go
```diff
@@ -480,7 +480,7 @@ func (m MatchFile) selectFile(r *http.Request) (matched bool) {
 		return true
 	}
 
-	return
+	return false
```
AnomalRoil, line 483: Actually you might want to also  remove the named return above in `func (m MatchFile) selectFile(r *http.Request) (matched bool)`  since it seems completely unused except for the implicit `false` values which you just replaced.

It would probably help with readability, but would require to also add to the description to say "returns whether it matched or not".

modules/caddyhttp/httpredirectlistener.go
```diff
@@ -51,11 +51,11 @@ func (HTTPRedirectListenerWrapper) CaddyModule() caddy.ModuleInfo {
 	}
 }
 
-func (h *HTTPRedirectListenerWrapper) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+func (*HTTPRedirectListenerWrapper) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
```
AnomalRoil, line 54:
```suggestion
func (*HTTPRedirectListenerWrapper) UnmarshalCaddyfile(_ *caddyfile.Dispenser) error {
```

modules/caddyhttp/reverseproxy/fastcgi/client_test.go
```diff
@@ -135,16 +134,19 @@ func sendFcgi(reqType int, fcgiParams map[string]string, data []byte, posts map[
 			length = len(data)
 			rd := bytes.NewReader(data)
 			resp, err = fcgi.Post(fcgiParams, "", "", rd, int64(rd.Len()))
+			resp.Body.Close()
```
AnomalRoil, line 137: You might want to wait until the `Body` is read below on line 170 before closing it, no ?

modules/caddyhttp/reverseproxy/fastcgi/client_test.go
```diff
@@ -135,16 +134,19 @@ func sendFcgi(reqType int, fcgiParams map[string]string, data []byte, posts map[
...
 		} else if len(posts) > 0 {
 			values := url.Values{}
 			for k, v := range posts {
 				values.Set(k, v)
 				length += len(k) + 2 + len(v)
 			}
 			resp, err = fcgi.PostForm(fcgiParams, values)
+			resp.Body.Close()
```
AnomalRoil, line 145: You might want to wait until the `Body` is read below on line 170 before closing it, no ?

modules/caddyhttp/reverseproxy/fastcgi/client_test.go
```diff
@@ -135,16 +134,19 @@ func sendFcgi(reqType int, fcgiParams map[string]string, data []byte, posts map[
...
 		} else {
 			rd := bytes.NewReader(data)
 			resp, err = fcgi.Get(fcgiParams, rd, int64(rd.Len()))
+			resp.Body.Close()
```
AnomalRoil, line 149: You might want to wait until the `Body` is read below on line 170 before closing it, no ?

modules/caddyhttp/reverseproxy/fastcgi/client_test.go
```diff
@@ -159,11 +161,12 @@ func sendFcgi(reqType int, fcgiParams map[string]string, data []byte, posts map[
 			length += len(k) + int(fi.Size())
 		}
 		resp, err = fcgi.PostFile(fcgiParams, values, files)
+		resp.Body.Close()
```
AnomalRoil, line 164: You might want to wait until the `Body` is read below on line 170 before closing it, no ?

modules/caddyhttp/reverseproxy/fastcgi/client_test.go
```diff
@@ -159,11 +161,12 @@ func sendFcgi(reqType int, fcgiParams map[string]string, data []byte, posts map[
...
 	}
 
 	if err != nil {
 		log.Println("err:", err)
-		return
+		return nil
 	}
 
 	defer resp.Body.Close()
```
AnomalRoil, line 172: To me it seems the `resp.Body` is closed here alright already.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -630,7 +630,7 @@ func (h Handler) prepareRequest(req *http.Request, repl *caddy.Replacer) (*http.
 	if err != nil {
 		// OK; probably didn't have a port
 		addr, err := netip.ParseAddr(address)
-		if err != nil {
+		if err != nil { //nolint:revive
 			// Doesn't seem like a valid ip address at all
 		} else {
```
AnomalRoil, line 635: @mholt did you really want to not return the `err` here in case of an invalid ip?

Looks like the linter might have caught an issue.

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -459,20 +459,19 @@ func (m *maxLatencyWriter) Write(p []byte) (n int, err error) {
 	defer m.mu.Unlock()
 	n, err = m.dst.Write(p)
 	if m.latency < 0 {
-		//nolint:errcheck
-		m.flush()
-		return
+		err = m.flush()
```
AnomalRoil, line 462: This is shadowing the `err` from the `m.dst.Write` above and thus might change the logic.

It might be worth considering returning an `errors.Join` of both errors, maybe?

modules/caddyhttp/staticerror.go
```diff
@@ -95,7 +95,7 @@ func (e *StaticError) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	return nil
 }
 
-func (e StaticError) ServeHTTP(w http.ResponseWriter, r *http.Request, _ Handler) error {
+func (e StaticError) ServeHTTP(w http.ResponseWriter, r *http.Request, h Handler) error {
```
AnomalRoil, line 98: this feels like it hurts readability. I agree with what @mholt was saying in an above comment, better keep the `_` for these function whose goal is to fulfill an interface.

modules/caddyhttp/staticresp_test.go
```diff
@@ -44,6 +44,7 @@ func TestStaticResponseHandler(t *testing.T) {
 	}
 
 	resp := w.Result()
+	resp.Body.Close()
 	respBody, _ := io.ReadAll(resp.Body)
```
AnomalRoil, line 48:
```suggestion
	resp := w.Result()
	respBody, _ := io.ReadAll(resp.Body)
	resp.Body.Close()
```

You want to read the `Body` before closing it ;)

modules/caddytls/certmanagers.go
```diff
@@ -52,7 +52,7 @@ func (ts Tailscale) GetCertificate(ctx context.Context, hello *tls.ClientHelloIn
 }
 
 // canHazCertificate returns true if Tailscale reports it can get a certificate for the given ClientHello.
-func (ts Tailscale) canHazCertificate(ctx context.Context, hello *tls.ClientHelloInfo) (bool, error) {
+func (Tailscale) canHazCertificate(ctx context.Context, hello *tls.ClientHelloInfo) (bool, error) {
```
AnomalRoil, line 55: 😆  Very tempted to `git blame` who came up with that naming, fitting of Tailscale being the cool kids tho!

modules/logging/filewriter.go
```diff
@@ -73,7 +73,7 @@ func (FileWriter) CaddyModule() caddy.ModuleInfo {
 }
 
 // Provision sets up the module
-func (fw *FileWriter) Provision(ctx caddy.Context) error {
+func (fw *FileWriter) Provision(caddy.Context) error {
```
AnomalRoil, line 76:
```suggestion
func (fw *FileWriter) Provision(_ caddy.Context) error {
```

modules/logging/netwriter.go
```diff
@@ -196,7 +196,7 @@ func (reconn *redialerConn) Write(b []byte) (n int, err error) {
 		if err2 != nil {
 			// logger socket still offline; instead of discarding the log, dump it to stderr
 			os.Stderr.Write(b)
-			return
+			return 0, err2
```
AnomalRoil, line 199: This changes the logic, but maybe in a good way. 

Might be worth considering returning `errors.Join` of both `err` and `err2` or maybe a new error wrapping both, tho.

AnomalRoil (commented):
Here are a few comments.

I think I agree with @mholt that keeping the `_` instead of using named but unused parameters is clearer.

But I also think that if you're replacing named return with explicit returns, you might as well drop the named returns most of the time. 
(Despite @mholt  clearly loving these named returns, I think it doesn't help with readability of long functions, so it's probably better to have explicit return values.)

francislavoie:
Thanks for the review comments @AnomalRoil, appreciated!

This PR will need rebasing, it's fallen a bit behind.

.golangci.yml
```diff
@@ -1,4 +1,74 @@
 linters-settings:
+  revive:
+    enable-all-rules: true
+    # Do NOT whine about the following, full explanation found in:
```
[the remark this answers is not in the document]
mholt: Oh I think they're actually referring to the linter complaining when we don't really care about these things or they aren't relevant in our project. :) Not a person. But yeah, maybe the proposed change here is better.

modules/caddytls/certmanagers.go
mholt: That might have been me, being commissioned by Tailscale (to dev the feature, not to name it that, lol). :no_mouth:

mholt:
Yes, thank you from me too, @AnomalRoil !

> But I also think that if you're replacing named return with explicit returns, you might as well drop the named returns most of the time.
> (Despite @mholt clearly loving these named returns, I think it doesn't help with readability of long functions, so it's probably better to have explicit return values.)

If I read you right, I think I agree. Named returns can be very elegant sometimes and even help improve correctness / reduce error-proneness, but if most/all of the return statements end up using variable names, the names should be removed from the return signature.

@faddat Any interest in finishing this up? Apologies if it's overwhelming. That's just what happens sometimes with big changes :sweat_smile: We appreciate it!

mholt:
Closing due to inactivity -- however, we'd be open to finishing this when you are @faddat (or anyone else who would like to pick this up)!

Maybe it would be easier to break it down into smaller changes if possible.

PR #5711, opened 2023-08-06
feat: use automaxprocs for better perf in containers

dunglas:
Proposal: Automatically set `GOMAXPROCS` to match Linux container CPU quota using https://github.com/uber-go/automaxprocs.
This should improve the performance of Caddy in containers: https://github.com/golang/go/issues/33803


mholt:
Thanks! I had no idea about this. (The Go issue looks very popular though.) Willing to accept this, but thinking it should probably wait until after the Caddy 2.8 feature freeze? What do we think?

At the outset, this looks like a feature/optimization, and introduces a new dependency. But I could see this being a bug fix if Caddy is actually experiencing noticeably higher latency in containers.

dunglas:
IMHO it could wait for Caddy 2.8. People can always set the environment variable manually in the meantime.

The latest (binary) release of Mercure includes this patch, for the record. We'll see if it causes any issues.

mholt:
Gotcha. We'll wait for now then, but let us know how it goes with your users, and if it looks stable and helpful, I'll consider advancing it up!

dunglas:
I improved log handling. It now uses zap and avoids the non-structured message.

mholt:
Thanks for this. I'd be good with trying it. Want to clean up the lint errors (just reorder/regroup the imports) and then we can merge it?

mholt (approved):
LGTM now! We'll experiment to see how this works.

PR #5712, opened 2023-08-06
chore: use revive linter

faddat:
This PR enables the revive linter, and should be merged after gci and gofumpt, since it contains them. 

- use gofmput to format code
- use gci to format imports
- reconfigure gci
- linter autofixes
- rearrange imports a little
- all changes other than unused-parameter
- use revive linter

## Contains

* #5707 
* #5708

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -565,7 +565,7 @@ func cmdFmt(fl Flags) (int, error) {
 	output := caddyfile.Format(input)
 
 	if fl.Bool("overwrite") {
-		if err := os.WriteFile(formatCmdConfigFile, output, 0600); err != nil {
+		if err := os.WriteFile(formatCmdConfigFile, output, 0o600); err != nil {
 			return caddy.ExitCodeFailedStartup, fmt.Errorf("overwriting formatted file: %v", err)
 		}
 		return caddy.ExitCodeSuccess, nil
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -630,7 +630,7 @@ func AdminAPIRequest(adminAddr, method, uri string, headers http.Header, body io
 	if err != nil {
 		return nil, fmt.Errorf("making request: %v", err)
 	}
-	if parsedAddr.IsUnixNetwork() {
+	if parsedAddr.IsUnixNetwork() { //nolint:revive
 		// We used to conform to RFC 2616 Section 14.26 which requires
 		// an empty host header when there is no host, as is the case
 		// with unix sockets. However, Go required a Host value so we
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -32,11 +32,12 @@ import (
 	"strings"
 
 	"github.com/aryann/difflib"
+	"go.uber.org/zap"
+
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 	"github.com/caddyserver/caddy/v2/internal"
-	"go.uber.org/zap"
 )
 
 func cmdStart(fl Flags) (int, error) {
--- a/modules/caddyhttp/fileserver/matcher.go
+++ b/modules/caddyhttp/fileserver/matcher.go
@@ -26,9 +26,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/caddyserver/caddy/v2"
-	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
-	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/common"
 	"github.com/google/cel-go/common/operators"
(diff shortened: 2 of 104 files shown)


modules/caddyhttp/ip_range.go
```diff
@@ -86,16 +86,16 @@ func (s *StaticIPRange) GetIPRanges(_ *http.Request) []netip.Prefix {
 }
 
 // UnmarshalCaddyfile implements caddyfile.Unmarshaler.
-func (m *StaticIPRange) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+func (f *StaticIPRange) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
```
francislavoie: Where does `f` come from? We typically use `m` to mean "module".

cmd/commandfuncs.go
```diff
@@ -630,7 +630,7 @@ func AdminAPIRequest(adminAddr, method, uri string, headers http.Header, body io
 	if err != nil {
 		return nil, fmt.Errorf("making request: %v", err)
 	}
-	if parsedAddr.IsUnixNetwork() {
+	if parsedAddr.IsUnixNetwork() { //nolint:revive
```
francislavoie, line 633: What happens otherwise?

modules/caddyhttp/push/handler.go
```diff
@@ -122,11 +122,8 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhtt
 	}
 
 	// serve only after pushing!
-	if err := next.ServeHTTP(lp, r); err != nil {
-		return err
-	}
-
-	return nil
+	err := next.ServeHTTP(lp, r)
+	return err
```
francislavoie, line 126: I guess this can be simplified even further to `return next.ServeHTTP(lp, r)`

modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -770,7 +770,7 @@ func leastRequests(upstreams []*Upstream) *Upstream {
 		return nil
 	}
 	var best []*Upstream
-	var bestReqs int = -1
+	bestReqs := -1
```
francislavoie, line 773: Nitpick but this is actually less symmetrical so it looks less nice. But whatever, it's fine.

modules/caddyhttp/fileserver/matcher.go
```diff
@@ -181,15 +181,15 @@ func (MatchFile) CELLibrary(ctx caddy.Context) (cel.Library, error) {
 			root = values["root"][0]
 		}
 
-		var try_policy string
+		var tryPolicy string
 		if len(values["try_policy"]) > 0 {
 			root = values["try_policy"][0]
```
francislavoie, line 186: Oh I think there's actually a bug here, this should be `tryPolicy = values["try_policy"][0]`

modules/logging/nopencoder.go
```diff
@@ -26,87 +26,87 @@ type nopEncoder struct{}
 
 // AddArray is part of the zapcore.ObjectEncoder interface.
 // Array elements do not get filtered.
-func (nopEncoder) AddArray(key string, marshaler zapcore.ArrayMarshaler) error { return nil }
+func (nopEncoder) AddArray(_ string, _ zapcore.ArrayMarshaler) error { return nil }
```
francislavoie, line 29: I'm not sure I like this. It makes it harder to read the intent of the params, even if they aren't used _for now_.

modules/caddyhttp/ip_range.go
faddat: f is for caddyfile, and it's used elsewhere on this receiver.  If you turn the revive linter on in the present state, it will flag this.

cmd/commandfuncs.go
faddat: the revive linter prompts us to remove the empty code block that is currently just comments.

modules/logging/nopencoder.go
faddat: in fact, I also am not sure I like it.  Last time I did a large lint of a repo, I chose not to use this linter.  Would you like me to disable it?

mholt: Disagreements with linters is why we've not used a lot of them yet, I think :sweat_smile:

If we have ways of disabling certain checks, maybe that's manageable.

faddat:
It's kinda my specialty and I am very happy to customize the configuration
to suit Caddy's unique needs.

modules/logging/nopencoder.go
faddat: Configuring linters kinda one of my specialties and I am very happy to customize the configuration to suit Caddy's unique needs.

In this case I put those unused parameter lint fixes last so they're easy to revert because I linted github.com/cosmos/ibc-go recently and came to the same conclusion that you did :D

modules/caddyhttp/fileserver/matcher.go
faddat: does this commit address the issue?

https://github.com/caddyserver/caddy/pull/5712/commits/6f1286a630f226e5f679db3f6cd82b55cdc9e31b

francislavoie:
I'll review again after the parent PRs are merged, hard to follow the diffs until then.

faddat:
Sorry, this one should be on draft for a minute.  I may have revive configured a little too spicy.

faddat:
I think this PR is no longer needed.  

Prefer #5709

PR #5713, opened 2023-08-06
use dependabot for go dependencies

faddat:
This PR moves the dependabot check to a weekly event for github actions, 
and adds the use of dependabot for go package updates.


francislavoie:
We had it on, but we decided to turn it off. It's too noisy and we rather deliberately upgrade dependencies when we're comfortable to do so.

PR #5720, opened 2023-08-08
feat: Add 64-bit RISC-V to goreleaser config

AaronDewes:
This will add RISC-V prebuilts for Caddy.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -43,6 +43,7 @@ builds:
   - arm64
   - s390x
   - ppc64le
+  - riscv64
   goarm:
   - "5"
   - "6"
@@ -54,14 +55,20 @@ builds:
       goarch: ppc64le
     - goos: darwin
       goarch: s390x
+    - goos: darwin
+      goarch: riscv64
     - goos: windows
       goarch: ppc64le
     - goos: windows
       goarch: s390x
+    - goos: windows
+      goarch: riscv64
     - goos: freebsd
       goarch: ppc64le
     - goos: freebsd
       goarch: s390x
+    - goos: freebsd
+      goarch: riscv64
     - goos: freebsd
       goarch: arm
       goarm: "5"


mholt:
I do believe 64-bit RISC-V is becoming more popular. I'd be okay with this. (Any reason we should not?)

mholt:
Just curious, what are you using Caddy on RISC-V for?

AaronDewes:
> Just curious, what are you using Caddy on RISC-V for?

I am building a server managment software that uses Caddy for easy HTTPS config.

image

This software also utilizes Docker, so I wanted to port various docker containers, including Caddy, to RISC-V, to get it working on RISC-V.

My initial goal is to get [~20 popular Docker images](https://riscy-containers.nirvati.org/containers) (website WIP) ported to RISC-V for this, and I noticed Caddy's dockerfile uses prebuilts by default, and these prebuilts are not available for RISC-V. I'll probably use my own prebuilt initially, but would love to use upstream builds long-term. Alpine doesn't have full RISC-V support anyway, so Caddy's official Docker image can't add RISC-V and I have to rebuild with my custom ubuntu image.

image

francislavoie:
I think we should add it to the cross-build job to if we don't already have that

AaronDewes:
> I think we should add it to the cross-build job too if we don't already have that

If I understand cross-build.yml correctly, it currently is only cross-compiling for different operating systems, but only builds the amd64 architecture?

mohammed90:
I love RISC-V! My only concern is that the Go project does not consider the riscv64 port [First-Class Port](https://github.com/golang/go/wiki/PortingPolicy#first-class-ports), meaning they will not block release or hold freeze if there's a bug with the riscv64 port. To be fair, FreeBSD is also not considered first-class port, yet we provide builds for it.

Have you built Caddy and run with/on riscv64? I've just cross-built Caddy with riscv64, but haven't confirmed it runs properly.

```
$ file ./caddy 
./caddy: ELF 64-bit LSB executable, UCB RISC-V, version 1 (SYSV), statically linked, Go BuildID=yGO5ucT9ssCMhUkYXpLk/aTLvHmsPfA_0MjSS61TK/GbjOWbk5MA0I5Z-FppHH/6iJnPBW7oK0jlWFC_PLH, with debug_info, not stripped
```

mholt:
(Did not know that, glad I asked!)

(I also have not encountered any problems with FreeBSD in many years. Maybe once? Or was that OpenBSD. In any case, we can try it I think.)

(Maybe we should add risc-v to our cross-build test since it's not a first-class port -- would that help catch anything?)

(Not sure why I'm writing in all parens today.)

mohammed90:
> (I also have not encountered any problems with FreeBSD in many years. Maybe once? Or was that OpenBSD. In any case, we can try it I think.)

BSDs are typically very solid and stable :)

> (Maybe we should add risc-v to our cross-build test since it's not a first-class port -- would that help catch anything?)

I just noticed that the cross-builds do not consider GOARCH. We can accommodate that.

> No, I haven't yet to be honest. I probably should, considering I just found a https://github.com/caddyserver/ingress/pull/128 in another part of Caddy 😆

@AaronDewes, I'd love to just do a smoke test. Let me ask around for cloud providers that offer riscv64 servers. In any ways, I'm very inclined to approve this PR but need to know any gaps that we might have to be aware of.

AaronDewes:
> > (I also have not encountered any problems with FreeBSD in many years. Maybe once? Or was that OpenBSD. In any case, we can try it I think.)
> 
> BSDs are typically very solid and stable :)
> 
> > (Maybe we should add risc-v to our cross-build test since it's not a first-class port -- would that help catch anything?)
> 
> I just noticed that the cross-builds do not consider GOARCH. We can accommodate that.
> 
> > No, I haven't yet to be honest. I probably should, considering I just found a [caddyserver/ingress#128](https://github.com/caddyserver/ingress/pull/128) in another part of Caddy laughing
> 
> @AaronDewes, I'd love to just do a smoke test. Let me ask around for cloud providers that offer riscv64 servers. In any ways, I'm very inclined to approve this PR but need to know any gaps that we might have to be aware of.

I have a RISC-V SBC standing on my desk where I'm compiling it right now, it's just very slow :sweat_smile:. I couldn't find any RISC-V cloud providers sadly when I last checked, it would have been very useful to me. Good RISC-V boards are quite expensive...

AaronDewes:
image

I also noticed Ubuntu ships riscv64 Caddy builds too, and these work for me too.

mohammed90 (approved):
LGTM!

mholt:
@mohammed90 

> I just noticed that the cross-builds do not consider GOARCH. We can accommodate that.

Originally I don't think they needed to, since we use mainly OS-specific features here and there (like OS-specific system calls, etc) but I don't think we do anything that is arch-dependent.

Still might not hurt to ensure the release builds succeed. Can we just x-compile it though? We probably don't need to run it on native. I just want to make sure our releases are as certain as possible (and since it cross-compiles, we should do cross-compile too in our own tests).

mohammed90:
> Can we just x-compile it though?

This is exactly what our cross-build CI/CD workflow does :) we cross-compile and check if the _build_ only is successful, and issue a warning otherwise.

mholt:
Yeah, I just wanted to make sure it was clear from @AaronDewes' comment above:

>  couldn't find any RISC-V cloud providers sadly when I last checked

that we don't need one, we can x-compile. :100:

Thanks everyone!!

PR #5738, opened 2023-08-12
Added dependabot config for go dependencies

Neo2308:
Added dependabot config for go dependencies to raise a PR every weekend updating all dependencies to latest

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,12 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      go-dependencies:
+        patterns:
+        - "*"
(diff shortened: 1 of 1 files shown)


mholt:
Thanks, but this is something we actually don't want to do. We try to upgrade dependencies only with intention and care.

Especially not on weekends 😅 I like to work less on weekends! 😉

mohammed90:
I agree with Matt. This is a fairly lightweight package. We'll get a notification if there's a security vulnerability in either of the deps.

Neo2308:
There are quite a few dependencies that this package uses and I thought this might be useful to keep them up to date. There should just be 1 PR every monday with all the updates from that week. But obviously its your call :)

mholt:
I think on almost any other Go project I'd accept this heartily, but for Caddy itself we usually upgrade when there are minor changes, useful improvements, etc, and there are some that we have to be careful about upgrading (anything touching otel, antlr4, cel, protobuf...) because of complicated situations in the past.

PR #5742, opened 2023-08-14
add unparam linter

faddat:
This PR adds the unparam linter to caddy.  

The result was getting rid of several parameters that were never used, and some situations where 
we would return an error, where errors could never be triggered.

Note: this should not be merged before #5708


francislavoie:
Hmm, I think most of these are there to leave the option open for later changes. I think these warning and error param/return are harmless as-is and slightly annoying to re-introduce if we find we need them when adding new code, so I think we can leave it in.

faddat:
works for me :)

mholt:
What did this do, exactly? Only had time for a quick scan so far, but I mostly see reordered imports -- where are the param changes?

francislavoie:
@mholt see [`5f5c8ed` (#5742)](https://github.com/caddyserver/caddy/pull/5742/commits/5f5c8eded534a342762aa428b3c9bd0953a36d91) i.e. it removed the `warnings` map in some Caddyfile code and the `error` return type in a spot it wasn't being checked, etc.

mholt:
@francislavoie That looks like it was _deleting_ parameters... which I agree we shouldn't do.

I am talking about renaming them to `_` when they're not used, which I do sometimes already.

The more I think about it, the more I think it needs to be case-by-case basis. Sometimes it's obvious (for example, contexts), other times, it's just an `int` and you have no idea what it is without a variable name.

So probably best to leave the params un-linted for now.

francislavoie:
Renaming to `_` is a different PR

mholt:
Oh. I got mixed up. Thanks.

Issue #5116, opened 2022-10-03
Support "precompressed" files without "base" files


TobiX:
As discussed in https://caddy.community/t/precompressed-files-without-base-files/17330, I was thinking about an extension to the precompressed feature: It would be nice to only have compressed files on disk and not have to keep the uncompressed files around...

Pro: Most clients support compression nowadays, so nothing changes for most clients
Contra: If some client not supporting compression comes around, Caddy need to decompress the file before delivering it to the client

@francislavoie suggested in the community thread that supporting this feature might impact performance due to the additional stat calls, so we probably need to keep that aspect in mind, too...

PS: While thinking about it a bit more, I could probably implement it as a custom file system which "synthesizes" the uncompressed files :laughing:

francislavoie:
I'm thinking it could be a new boolean option called `precompressed_allow_no_base` for a lack of a better name.

mholt:
How about `prefer_precompressed`?

shade34321:
I'm not really sure what all is needed for this and might need help. But I wouldn't mind taking a stab at this.

francislavoie:
@shade34321 take a look at the fileserver code and at the docs. Look for `precompressed` in the codebase for the key places to look at.

shade34321:
I've taken a look and I think this is going to take me a bit longer to understand before making a fix. So if someone else wants to work on it they can but I'll keep chugging along until I see a fix. Just wanted to update everybody.

titanventura:
Would like to take this up !

mholt:
Sounds good! Let us know if you have any questions.

Ahmed-Sermani:
Hi, I would like to work on this but I have a question.
If a client doesn't support compression, should I check if the uncompressed version present first?
This helps avoid decompressing the compressed file if the original present

titanventura:
I have a question. What approach does caddy currently take, for clients that do not support compression ?

titanventura:
would be better if i can connect on call with a person who can explain about this. I'm not sure if this is common in open-source communities. But i feel it is essential.

francislavoie:
We follow the `Accepts-Encoding` header. If a client doesn't declare it can handle compressed content, then Caddy will not compress it. For `file_server`'s precompressed files, that's why it's useful to have an uncompressed copy on file.

ueffel:
> I have a question. What approach does caddy currently take, for clients that do not support compression ?

[...]

PR #5745, opened 2023-08-14
fileserver: Support `precompressed` files without base files

singhalkarun:
Closes https://github.com/caddyserver/caddy/issues/5116

Changes done so far:

1. Added a new bool value to the FileServer Struct called PreferPrecompressed
2. Start accepting prefer_precompressed subdirective under the fileserver directive
3. If PreCompressed is enabed on fileserver, ignore checking the existence of the original file and serve encoded file as per the accepted encoding if available, else return 404.
 
To Do:
    
If the file is not available as per accepted encoding, try to decompress the encoded file to original file and serve, if not able to, return 404. (Not handling this in current PR, will be opening a new PR for this.)

@francislavoie

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -132,6 +132,8 @@ type FileServer struct {
 	// clobbering the explicit rewrite with implicit behavior.
 	CanonicalURIs *bool `json:"canonical_uris,omitempty"`
 
+	PreferPrecompressed bool `json:"prefer_precompressed,omitempty"`
+
 	// Override the status code written when successfully serving a file.
 	// Particularly useful when explicitly serving a file as display for
 	// an error, like a 404 page. A placeholder may be used. By default,
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -257,104 +257,109 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 		zap.String("request_path", r.URL.Path),
 		zap.String("result", filename))
 
-	// get information about the file
-	info, err := fs.Stat(fsrv.fileSystem, filename)
-	if err != nil {
-		err = fsrv.mapDirOpenError(err, filename)
-		if errors.Is(err, fs.ErrNotExist) || errors.Is(err, fs.ErrInvalid) {
+	if !fsrv.PreferPrecompressed {
+		// get information about the file
+		info, err := fs.Stat(fsrv.fileSystem, filename)
+		if err != nil {
+			err = fsrv.mapDirOpenError(err, filename)
+			if errors.Is(err, fs.ErrNotExist) || errors.Is(err, fs.ErrInvalid) {
+				return fsrv.notFound(w, r, next)
+			} else if errors.Is(err, fs.ErrPermission) {
+				return caddyhttp.Error(http.StatusForbidden, err)
+			}
+			return caddyhttp.Error(http.StatusInternalServerError, err)
+		}
+
+		// if the request mapped to a directory, see if
+		// there is an index file we can serve
+		var implicitIndexFile bool
+		if info.IsDir() && len(fsrv.IndexNames) > 0 {
+			for _, indexPage := range fsrv.IndexNames {
+				indexPage := repl.ReplaceAll(indexPage, "")
+				indexPath := caddyhttp.SanitizedPathJoin(filename, indexPage)
+				if fileHidden(indexPath, filesToHide) {
+					// pretend this file doesn't exist
+					fsrv.logger.Debug("hiding index file",
+						zap.String("filename", indexPath),
+						zap.Strings("files_to_hide", filesToHide))
+					continue
+				}
+
+				indexInfo, err := fs.Stat(fsrv.fileSystem, indexPath)
+				if err != nil {
+					continue
+				}
+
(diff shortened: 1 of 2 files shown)


mholt:
Thanks for the enhancement!

This is looking pretty good.

I am not quite sure about this, though:

> If the file is not available as per accepted encoding, try to decompress the encoded file to original file and serve, if not able to, return 404.

Is this something we should actually do? And if so, is 404 the right status code?

Note: We're on feature freeze until 2.9 development, so even if this gets merged soon it likely won't get released until after 2.8.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -476,7 +483,12 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 	// that errors generated by ServeContent are written immediately
 	// to the response, so we cannot handle them (but errors there
 	// are rare)
-	http.ServeContent(w, r, info.Name(), info.ModTime(), file.(io.ReadSeeker))
+
+	if fsrv.PreferPrecompressed {
+		http.ServeContent(w, r, compressedInfo.Name(), compressedInfo.ModTime(), file.(io.ReadSeeker))
+	} else {
+		http.ServeContent(w, r, info.Name(), info.ModTime(), file.(io.ReadSeeker))
+	}
```
francislavoie, line 491: I think we can just do `info = compressedInfo` to simplify this to a single call to `http.ServeContent`

singhalkarun: Agreed. Pushed a new commit. Refactored to use only one variable info info instead of 2 separate variables, info and compressedInfo.

singhalkarun:
> Is this something we should actually do? And if so, is 404 the right status code?
> 

Not very sure on this. But considering the main discussion thread (https://caddy.community/t/precompressed-files-without-base-files/17330),

1. We should be able to serve the file even if the precompressed version doesn't exist (This is already solved by the PR)
2. If the client doesn't support compressed version, we should decompress and serve (This is yet to be solved which can be discussed over)

For now, let me convert the draft PR to a normal PR and I'll open a separate PR that handles the 2nd scenario and we can continue the discussion over there.

@francislavoie @mholt

mholt (changes requested):
Thanks for this feature. I think it's looking pretty good, but I haven't done a thorough inspection or tried it yet.

It might be best to add a test case (an integration test) that verifies it works.

Then we'll get this merged in for ~2.9!

singhalkarun:
> Thanks for this feature. I think it's looking pretty good, but I haven't done a thorough inspection or tried it yet.
> 
> It might be best to add a test case (an integration test) that verifies it works.
> 
> Then we'll get this merged in for ~2.9!

Sure, let me add test cases.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -132,6 +132,8 @@ type FileServer struct {
 	// clobbering the explicit rewrite with implicit behavior.
 	CanonicalURIs *bool `json:"canonical_uris,omitempty"`
 
+	PreferPrecompressed bool `json:"prefer_precompressed,omitempty"`
```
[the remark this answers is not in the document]
singhalkarun: Got it. Let me add the godoc comments to this field.

mholt:
@singhalkarun I think we're almost done with this. Any interest in finishing it up?

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -476,6 +492,7 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
 	// that errors generated by ServeContent are written immediately
 	// to the response, so we cannot handle them (but errors there
 	// are rare)
+
```
francislavoie, line 495: Don't need this newline anymore

```suggestion
```

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -388,17 +395,26 @@ func (fsrv *FileServer) ServeHTTP(w http.ResponseWriter, r *http.Request, next c
...
+			etag = calculateEtag(info)
 		}
 
 		break
 	}
 
 	// no precompressed file found, use the actual file
 	if file == nil {
-		fsrv.logger.Debug("opening file", zap.String("filename", filename))
-
```
francislavoie, line 400: I think we should preserve the debug log, it's helpful for tracing. In fact, maybe we should add more debug logs in here relating to which file is being picked depending on the precompressed options.

singhalkarun:
> @singhalkarun I think we're almost done with this. Any interest in finishing it up?

@mholt Sorry, got occupied with other things. Can I work on this after 15 Sep if it works?

mholt:
Yeah, absolutely. I just wanted to check in for status. 👌

mholt:
@singhalkarun Just wanted to check in one more time, wondering if you're still interested in this.

singhalkarun:
> Thanks for the enhancement!
> 
> This is looking pretty good.
> 
> I am not quite sure about this, though:
> 
> > If the file is not available as per accepted encoding, try to decompress the encoded file to original file and serve, if not able to, return 404.
> 
> Is this something we should actually do? And if so, is 404 the right status code?
> 
> Note: We're on feature freeze until 2.9 development, so even if this gets merged soon it likely won't get released until after 2.8.

Hi @mholt I would like to pass this one for now due to time constraints. Thanks.

mholt:
Alrighty, thanks for the update.

PR #5749, opened 2023-08-15
core: quic listener will manage the underlying socket by itself.

WeidiDeng:
Related to [5725](https://github.com/caddyserver/caddy/pull/5725), I found out we have to switch uderlying sockets using another wrapper. Not only is it cubersome, but it will also miss performance optimizations done by `quic-go`. As udp sockets with `SO_REUSEPORT` are not the same, and quic-go will only optimize them once, meaning after config reload performance will decrease.

This patch changes how quic listener are created. Instead of passing a `net.PacketConn`, quic listener now creates one and close it when no longer needed. It's now much easier to optimize quic-go performance without implementing those wrapper methods.

Update: this patch also fixed the active request counter to calculate highload for http3.


francislavoie:
Oooh this reads a lot cleaner! Nice

listeners.go
```diff
@@ -597,13 +655,17 @@ func (sqtc *sharedQUICTLSConfig) addTLSConfig(tlsConfig *tls.Config) (context.Co
...
+	sqtc       *sharedQUICTLSConfig
+	key        string
 }
 
-// Destruct closes the underlying QUIC listener.
+// Destruct closes the underlying QUIC listener and its associated net.PacketConn.
 func (sql *sharedQuicListener) Destruct() error {
-	return sql.EarlyListener.Close()
+	// close EarlyListener first to stop any operations being done to the net.PacketConn
+	_ = sql.EarlyListener.Close()
```
mholt, line 666: I wonder if we should at least do a log output in the rare event this does fail. `Log().Error("closing QUIC early listener", zap.Error(err))` or something.

listeners.go
```diff
@@ -652,6 +714,11 @@ var _ quic.OOBCapablePacketConn = (*fakeClosePacketConn)(nil)
 // but doesn't actually use these methods, the only methods needed are `ReadMsgUDP` and `SyscallConn`.
 var _ net.Conn = (*fakeClosePacketConn)(nil)
 
+// Unwrap returns the underlying net.UDPConn for quic-go optimization
+func (fcpc *fakeClosePacketConn) Unwrap() any {
+	return fcpc.UDPConn
+}
```
mholt, line 720: Is this an interface quic-go supports? Cool, I didn't know that.

listeners.go
```diff
@@ -465,12 +465,70 @@ func ListenPacket(network, addr string) (net.PacketConn, error) {
 	return ln.(net.PacketConn), nil
 }
 
+// ListenQUIC returns a quic.EarlyListener suitable for use in a Caddy module.
+// The network will be transformed into a QUIC-compatible type (if unix, then
+// unixgram will be used; otherwise, udp will be used).
+//
+// NOTE: This API is EXPERIMENTAL and may be changed or removed.
+func (na NetworkAddress) ListenQUIC(ctx context.Context, portOffset uint, config net.ListenConfig, tlsConf *tls.Config, activeRequests *int64) (http3.QUICEarlyListener, error) {
```
mholt, line 473: There's a lot of logic in here, as well as in the now-deprecated `caddy.ListenQUIC()`. Can we have `caddy.ListenQUIC()` call this one or would that be too much change do you think?

I just noticed the other ones (caddy.Listen, etc) don't call NetworkAddress.Listen(), so maybe those should too? I dunno. Just a thought. To help avoid duplicating code.

mholt (changes requested):
Thanks Weidi!

Is the intention here that I should merge this PR before mine?

listeners.go
WeidiDeng: Na, this just exposes the underlying socket so quic-go can optimize it. In this case it's a `*net.UDPConn`.

listeners.go
WeidiDeng: `caddy.ListenQUIC()` takes a `net.PacketConn` and its life cycle is managed by the called. It doesn't feel right to call `NetworkAddress.ListenQUIC()`.

The others maybe changed to do so, but that will be another patch.

WeidiDeng:
@mholt Yeah, I think this should be merged before your `SO_REUSEPORT` patch. That branch looks mostly fine. Just a few places need changing.

WeidiDeng:
There is a problem with quic early listener using older `activeRequests` from the older instance, just like `tls.Conf`, I'm thinking this part needs some rework.

listeners.go
mholt: Ok, duh, you're right. Don't worry about changing that for now. :+1: Thanks!

listeners.go
mholt: So quic-go needs to be patched to call this method?

WeidiDeng: We call this method ourselves, by passing `*net.UDPConn` or `*net.UnixConn` to `quic.ListenEarly`. And since we're managing the connection using `quicListener`, `fakeClose` shouldn't apply to it anyway.

francislavoie:
Is this ready? From the comments it seems like there's still some work to do? I'll bump the milestone for now, I guess.

WeidiDeng:
It's mostly done, but I'd like to see [5725](https://github.com/caddyserver/caddy/pull/5725) merged first.

francislavoie:
Oh okay, you said the opposite in https://github.com/caddyserver/caddy/pull/5749#issuecomment-1683139437 :man_shrugging:

WeidiDeng:
My bad 😅 . I think this pr should be merged before that one, there is some tweaks to be done there.

francislavoie:
Okay, if this should be merged first, could you rebase it and do the final touchups? :smile:

listeners.go
```diff
@@ -465,43 +465,101 @@ func ListenPacket(network, addr string) (net.PacketConn, error) {
...
+		contextCancel:      cancel,
+	}, nil
+}
+
 // ListenQUIC returns a quic.EarlyListener suitable for use in a Caddy module.
 // The network will be transformed into a QUIC-compatible type (if unix, then
 // unixgram will be used; otherwise, udp will be used).
 //
 // NOTE: This API is EXPERIMENTAL and may be changed or removed.
 //
```
mholt, line 532:
```suggestion
```

mholt (approved):
Thanks so much for working on this. I admit I don't follow _all_ of it (I'm also a bit sleep deprived with recent baby and such), but at this point I trust your skill a bit more than mine on this matter :sweat_smile: This looks more complicated, but based on my understanding it should be more correct, which is probably better.

Just curious if you think we could eventually have `ListenQUIC()` call the `NetworkAddress.ListenQUIC()` method instead of what looks like duplicating a lot of code?

WeidiDeng:
@mholt We have to parse a `net.PacketConn` to `ListenQUIC()` and then manually close it when we no longer need it, while `NetworkAddress.ListenQUIC()` manage `net.PacketConn` by itself. These two don't work well together.

mholt:
Gotcha. Well, let's give this a try and then remove the deprecated `caddy.ListenQUIC()` function in the near-future. Thank you!!

PR #5751, opened 2023-08-15
fileserver: docs: clarify the ability to produce JSON array with `browse`

mohammed90:
See: https://caddy.community/t/simple-file-list-from-browse-perhaps-json/20836


steffenbusch:
Would you mind to add an explanation for `mode` in that json output? 
I have just tried that on my caddy instance and got values of `2147484157` for directories and `436` for all or most other files.

francislavoie:
Good point @steffenbusch; it comes from https://pkg.go.dev/io/fs#FileMode, so `ModeDir` is set for directories, i.e. the 31st bit from the right (leaving a gap of one bit to let it fit in a signed int32, I think). It's a bit more visible if you convert the decimal number to octal:

```
> decoct(2147484157)
= "20000000775"

> decoct(436)
= "664"
```

The `2` is the directory bit (2^31), and the 775 is the file permission bits (`rwxrwxr-x`).

steffenbusch:
Thanks for the explanation @francislavoie. 
I assume it's not possible to have our beloved unix permissions instead, right?

mohammed90:
The `String()` method returns the mode in the `drwxrwxrwx` format. [See here](https://go.dev/play/p/t0aRIh4jrK8). I guess we can wrap the type with custom JSON marshalling that'll present it as string. But... what's preferred: the string, octal, or decimal?

steffenbusch:
Thanks for the Go Playground link @mohammed90. Hopefully this will allow me to understand more Go code in the future.
I like the `String()` output of `drwxrwxr-x` for `2147484157`.

mholt:
Yeah, I do think there's value in the number. I agree we can additionally return a string formatting if that kind of interpretation is difficult in whatever is consuming this, but I imagine most languages have a function to convert a decimal to octal.

mholt (approved):
This docs change LGTM! Thanks Mohammed.

PR #5757, opened 2023-08-16
httpcaddyfile: Stricter errors for site and upstream address schemes

### LINKED ISSUE #5755: Warn for `ws://` and `wss://` site and upstream addresses
I've seen that commonly, users try to use `ws://` or `wss://` either as site addresses or proxy upstream addresses. Neither of these work in Caddy, they're specific to browsers for use with the `new WebSocket(url)` JS API.

We should warn/error when those are used in the config, and suggest using `http://` or `https://` (or omitting the scheme) respectively.

singhalkarun:
- [x] If someone uses wss or ws specifically, we should send error saying them to use https or http respectively
- [x] We should allow https/http/h2c/empty as schemes and if someone uses anything apart from this we should send them error message with invalid prefix 
- [x] Need to add same checks for the site adresses (h2c not supported yet)
- [x] Refactor Code
- [x] Add tests

Fix #5755


modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -155,6 +155,14 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
 		// transports (protocols or schemes) to each backend,
 		// so we remember the last one we see and compare them
+		if pa.scheme == "wss" {
+			return d.Errf("Please use https instead of wss at %s", address)
+		} else if pa.scheme == "ws" {
+			return d.Errf("Please use http instead of ws at %s", address)
+		} else if pa.scheme != "" && pa.scheme != "https" && pa.scheme != "http" && pa.scheme != "h2c" {
+			return d.Errf("Unsupported URL Scheme/Prefix at %s", address)
+		}
+
```
mohammed90, line 165: At this point, it's probably best/cleaner to use a `switch`.

The error messages typically start with lowercase letters.

singhalkarun: Thanks for the amazing feedback @mohammed90. Let me take these points into consideration and refactor the code.

singhalkarun:
Hey @mohammed90 need your suggestions on another refactoring part. There is a file called addresses.go which is part of package reverseproxy and I want to utilize this in module caddyfile. I am not able to utilize it directly by importing as members of addresses.go are not exported. 

Would it be a good idea to export them and then utilize same addresses.go by importing in caddyfile module or is there a better way to do it? For now for sake of testing the flow I made a copy of addresses.go inside caddyfile module itself.

singhalkarun:
> Pulling that file and its logic doesn't fit. You don't need its content. The validation should be in this method:
> 
> https://github.com/caddyserver/caddy/blob/d6f86cccf5fa5b4eb30141da390cf2439746c5da/caddyconfig/httpcaddyfile/addresses.go#L191

Hey @mohammed90 Thanks for the guidance, refactored the code. Now it seems to be at apppropriate place. Let me know if there are any other improvements you see.

mohammed90 (changes requested):
Thanks, this is looking better! :grin: We're just missing tests. Can you add some tests for the behavior? This is how to test adapt/load failures.

https://github.com/caddyserver/caddy/blob/d6f86cccf5fa5b4eb30141da390cf2439746c5da/caddytest/integration/caddyfile_test.go#L63-L76

Use `AssertLoadError` for failure cases and [`AssertAdapt`](https://pkg.go.dev/github.com/caddyserver/caddy/v2@v2.7.3/caddytest#AssertAdapt) for success cases. Try to structure the test cases as a table/slice of cases.

singhalkarun:
> Thanks, this is looking better! 😁 We're just missing tests. Can you add some tests for the behavior? This is how to test adapt/load failures.

Sure, Agreed. Let me add some tests.

singhalkarun:
Hey @mohammed90 added test cases for failure cases. 

Facing issues for valid cases. Getting absurd issue. The bracket is at correct indentation, still it shows some indentation error. Can you suggest me where I might be going wrong?

image

mohammed90:
> Hey @mohammed90 added test cases for failure cases. 
> 
> Facing issues for valid cases. Getting absurd issue. The bracket is at correct indentation, still it shows some indentation error. Can you suggest me where I might be going wrong?
> 
> image
> 

If I were to guess, I'd say it's a tabs vs spaces issue. Our convention is to use tabs. If you're not using spaces, I'd have to look at your code to investigate.

singhalkarun:
> If I were to guess, I'd say it's a tabs vs spaces issue. Our convention is to use tabs. If you're not using spaces, I'd have to look at your code to investigate.

It worked, but very curious to know how it works internally because a tab is also adding the same number of space which I manually add by pressing spaces (I'll research on this maybe, share if you know any resource to understand reason behind this). I will push another commit with the test cases for valid scenarios shortly.

mholt (approved):
Nice, this LGTM! Thanks for the test cases as well. :blush:

francislavoie:
I rebased and added a commit, I wanted to adjust the error text a bit for clarity.

Thanks!

PR #5765, opened 2023-08-19
caddyfile: Adjust error formatting

francislavoie:
Before:

```
Error: parsing caddyfile tokens for 'reverse_proxy': parsing upstream 'wss://foo': Caddyfile:7 - Error during parsing: the scheme wss:// is only supported in browsers; use https:// instead, import chain: ['']
```

After:

```
Error: parsing caddyfile tokens for 'reverse_proxy': parsing upstream 'wss://foo': the scheme wss:// is only supported in browsers; use https:// instead, at Caddyfile:7
```

Essentially: moving the filename to the end, hiding import chain if empty, and removing the redundant "Error during parsing".

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -391,22 +391,22 @@ func (d *Dispenser) Reset() {
 // an argument.
 func (d *Dispenser) ArgErr() error {
 	if d.Val() == "{" {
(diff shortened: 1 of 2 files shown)


mohammed90 (approved):
Thanks! Loving the changes. The empty `import chain` bothered me but wasn't sure how to change it. I like what you did. LGTM.

mholt (approved):
Agreed -- thanks!

Issue #5773, opened 2023-08-21
Placeholder {time.now.http} prints local time as GMT time


steffenbusch:
The problem I'm having:

I want to setup a very simple Caddy health-check endpoint and want to utilize placeholders from https://caddyserver.com/docs/conventions#placeholders to display the date + time the health check was successfully executed.

The placeholder `{time.now.http}` is described as
> The current time in the format used in HTTP headers

The server, where Caddy is running, is using GMT+2 (`/etc/localtime -> /usr/share/zoneinfo/Europe/Berlin`).
While the placeholders `{time.now}` and `{time.now.common_log}` are perfect and return a date/time with the correct timezone, the returned value of `{time.now.http}` says the time is GMT:

```
$ date
Mon Aug 21 18:16:13 CEST 2023
$ curl --dump-header - https://playground.stbu.net/compare-time
HTTP/2 200
alt-svc: h3=":443"; ma=2592000
content-type: text/plain; charset=utf-8
server: Caddy
content-length: 333
date: Mon, 21 Aug 2023 16:16:13 GMT

1) time.now            = 2023-08-21 18:16:13.949657242 +0200 CEST m=+1187.670368014
2) time.now.http       = Mon, 21 Aug 2023 18:16:13 GMT
3) time.now.common_log = 21/Aug/2023:18:16:13 +0200
--
1) The current time as a Go Time struct
2) The current time in the format used in HTTP headers
3) The current time in Common Log Format
```

The output for `2) time.now.http` should have been `Mon, 21 Aug 2023 16:16:13 GMT` (= the same value as the `date` response header shown by curl), not `Mon, 21 Aug 2023 18:16:13 GMT` (Hour = 18 instead of 16).

My Caddyfile to reproduce this issue (as long as your server is not in UTC) is:
```
{
        http_port 80
        https_port 443
        skip_install_trust true

        servers :443 {
        }
        default_sni playground.stbu.net

        log debug_log {
                output file log/debug.log {
                        roll_size 10m
                        roll_keep_for 7d
                }
                format console {
                        time_format iso8601
                }
                level debug
                exclude http.log.access.log0
        }

}

https://playground.stbu.net {

        handle /caddy-health-status {
                header Content-Type application/json
                respond `{"status": "ok", "as-of": "{time.now.http}", "powered-by":"{$HOSTNAME}"}`
        }

        handle /compare-time {
                respond <<TXT
1) time.now            = {time.now}
2) time.now.http       = {time.now.http}
[...]

PR #5774, opened 2023-08-22
replacer: change timezone to UTC for "time.now.http" placeholders

WeidiDeng:
Fix [5773](https://github.com/caddyserver/caddy/issues/5773)

The comment for `http.TimeFormat` says the time being formatted must be in the UTC timezone to get the correct value and it's indeed used by stdlib.


francislavoie (approved):

mholt:
Ah, thanks for this Weidi!

PR #5779, opened 2023-08-24
logging: query filter for array of strings

pauljeannot:
As discussed in this [previous PR](https://github.com/caddyserver/caddy/pull/5101), here is the improvement to make the query filter handle array of strings.

This is needed so that the query filter can be used on headers, especially useful for the `Referer`.

**Disclaimer:** my first lines of code in Go, tried my best using the same logic and style that were used already. Successfully tested locally (and unit tests passed).

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -26,7 +26,6 @@ import (
 	"strings"
 
 	"go.uber.org/zap/zapcore"
-	"golang.org/x/exp/slices"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
@@ -375,11 +374,12 @@ func (m *QueryFilter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 // Filter filters the input field.
 func (m QueryFilter) Filter(in zapcore.Field) zapcore.Field {
 	if array, ok := in.Interface.(caddyhttp.LoggableStringArray); ok {
-	newArray := slices.Clone(array)
-		for i, s := range newArray {
+		newArray := make(caddyhttp.LoggableStringArray, len(array))
+		for i, s := range array {
 			newArray[i] = m.processQueryString(s)
 		}
-		in.Interface = newArray	} else {
+		in.Interface = newArray
+	} else {
 		in.String = m.processQueryString(in.String)
 	}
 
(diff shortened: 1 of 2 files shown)


mholt (approved):
Very well done. :100: Especially for first Go code!

I'll let Francis give this a review but IMO this LGTM.

@francislavoie Should this be a bugfix or is this a feature? Wondering whether we should milestone this for 2.8 or 2.9.

francislavoie:
@mholt it really straddles the line. I think it's niche enough that it can be released as a bug fix with low risk.

francislavoie (approved):

francislavoie:
Before we merge this, this will need to inherit the same change from https://github.com/caddyserver/caddy/pull/5786 to fix a bug, otherwise modifying the `array` will modify the original request :grimacing:

francislavoie (changes requested):
Before we merge this, this will need to inherit the same change from https://github.com/caddyserver/caddy/pull/5786 to fix a bug, otherwise modifying the `array` will modify the original request :grimacing:

pauljeannot:
Outch indeed! Changes made 🙏

modules/logging/filters.go
```diff
@@ -373,9 +374,22 @@ func (m *QueryFilter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 // Filter filters the input field.
 func (m QueryFilter) Filter(in zapcore.Field) zapcore.Field {
-	u, err := url.Parse(in.String)
+	if array, ok := in.Interface.(caddyhttp.LoggableStringArray); ok {
+	newArray := slices.Clone(array)
+		for i, s := range newArray {
+			newArray[i] = m.processQueryString(s)
+		}
+		in.Interface = newArray	} else {
```
francislavoie, line 382:
```suggestion
		newArray := make(caddyhttp.LoggableStringArray, len(array))
		for i, s := range array {
			newArray[i] = m.processQueryString(s)
		}
		in.Interface = newArray
	} else {
```

modules/logging/filters.go
```diff
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"go.uber.org/zap/zapcore"
+	"golang.org/x/exp/slices"
```
francislavoie, line 29:
```suggestion
```

francislavoie (changes requested):

francislavoie (approved):

Issue #5776, opened 2023-08-23
Surprising behavior with empty handle_response route


pkoenig10:
**Caddy version: v2.7.4 h1:J8nisjdOxnYHXlorUKXY75Gr6iBfudfoGhrJ8t7/flI=**

The `handle_response` [documentation](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#intercepting-responses) states:

> When a response handler is invoked, the response from the backend is not written to the client, and the configured handle_response route will be executed instead, and it is up to that route to write a response. If the route does not write a response, then request handling will continue with any handlers that are [ordered after](https://caddyserver.com/docs/caddyfile/directives#directive-order) this reverse_proxy.

Given this, I would expect a empty `handle_response` route to cause the request handling to continue with the next handler. But that doesn't appear to be the case. Furthermore, the behavior appears to be inconsistent and dependent upon the presence or absence of other directives.

## Reproductions

In all the examples below, I am making a request to `http://localhost`.

1. Given this Caddyfile, I'd expect the response to be handled by the `handle_response @ok` route and then continue to the `respond 299` directive. However the actual response is a 499.

    ```
    {
        auto_https off
        order respond after reverse_proxy
    }

    :80 {
        reverse_proxy example.com {
            header_up Host "example.com"

            @ok status 200
            handle_response @ok {
            }

            handle_response {
                respond 499
            }
        }

        respond 299
    }
    ```

2. If I add a useless directive to the `handle_response @ok` route, then that route is used as expected. The actual response here is a 299

    ```
    {
        auto_https off
        order respond after reverse_proxy
    }

    :80 {
        reverse_proxy example.com {
            header_up Host "example.com"

            @ok status 200
            handle_response @ok {
                skip_log
            }

            handle_response {
                respond 499
            }
        }

        respond 299
    }
    ```

2. If I remove the `handle_response` route, then I'd expect to continue to the `respond 299` directive. But the actual response here is the response from the upstream `exmaple.com` request. This is particularly surprsing since `copy_response` is not used anywhere.

    ```
    {
        auto_https off
        order respond after reverse_proxy
    }

    :80 {
        reverse_proxy example.com {
            header_up Host "example.com"

            @ok status 200
            handle_response @ok {
            }
        }

        respond 299
    }
    ```

WeidiDeng:
[...]

PR #5780, opened 2023-08-25
reverseproxy: Allow fallthrough for response handlers without routes

pkoenig10:
Fixes https://github.com/caddyserver/caddy/issues/5776

AFAICT there's no reason for this condition to exist. If there are no routes, then the compiled handler will simply invoke the `next` handler, which is the desired behavior.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -848,12 +848,6 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
 			break
 		}
 
-		// otherwise, if there are any routes configured, execute those as the
-		// actual response instead of what we got from the proxy backend
-		if len(rh.Routes) == 0 {
-			continue
-		}
-
 		// set up the replacer so that parts of the original response can be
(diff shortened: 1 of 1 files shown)


francislavoie:
After reading the code more closely, yeah I think this is the fix. My concern was with `rh.StatusCode` but that's already implemented as being exclusive with the `break` on line 848 which I missed earlier. I can't think of any downside.

francislavoie (approved):

pkoenig10:
@francislavoie Is there anything else that needs to be done before this can merge?

mholt:
From a quick discussion on Slack, I think the last thing is he/we want to test it out to verify a couple things as soon as we can :)

francislavoie:
Yeah - I'm just nervous about the lack of tests we have surrounding this so I'd like to do a bit of manual testing of my own when I find a bit of time to focus on it.

Ideally we should have some automated tests added for this as well, if anyone wants to help by contributing those :eyes:

francislavoie:
Okay, yeah. I'm good with this. Played around with it a bit with a handful of combinations of config to make sure it covers what I expect.

Thanks for helping to identify the fix, and sorry for taking so long to get around to it :blush:

PR #5784, opened 2023-08-25
tls: modularize trusted CA providers

mohammed90:
The current implementation of client authentication requires the CA to be either known in DER format or available on disk as PEM files. This prevents the advanced usage of client authentication by standing on the shoulders of the PKI app or to fetch the certs from shared Caddy storage. The current implementation makes it hard to share the same trusted client CAs across a cluster of Caddy servers without having to produce the root certs first then plugging them. This is a long way of saying, the trusted root CA is very rigid and not extensible.

This implementation allows for the trust CA to be any Caddy module that satisfies an interface. It also provides 4 basic providers:
- File, to maintain backwards compatibility
- Root certs from the PKI app
- Intermediate certs from the PKI app
- Storage

TODO:
- [x] Caddyfile
- [x] Tests
- [x] Docs


mholt (commented):
Oh this is actually awesome -- thank you for contributing this!

It's pretty close to what I'd do. 

- Exported types and their fields will need godocs.
- It's a _slight_ bummer we can't combine the PKI*CAPool types. (Unless we can? I wonder what that would look like. It's just a matter of what the admin wants to sign the certs with, right?)

I have little in the way of criticism of this change. Nice work :smiley:

mohammed90:
>     * Exported types and their fields will need godocs.
> 
>     * It's a _slight_ bummer we can't combine the PKI*CAPool types. (Unless we can? I wonder what that would look like. It's just a matter of what the admin wants to sign the certs with, right?)
> 

I'll come back to it when I'm in a better head-space.

One more thing, I did consider `Lazy` for some of them, which delays the load-and-parse until the pool is requested, allowing the pool certs to updated without having to reload Caddy. I changed my mind because this will be on a hot-path, and such delay will be unacceptable. They can be 3rd party modules if at all needed.

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+		}
+		for nesting := d.Nesting(); d.NextBlock(nesting); {
+			switch d.Val() {
+			case "trust_der":
+				icp.TrustedCACerts = append(icp.TrustedCACerts, d.RemainingArgs()...)
+			default:
+				return fmt.Errorf("unrecognized directive: %s", d.Val())
+			}
+		}
+	}
```
francislavoie, line 95: See https://github.com/caddyserver/caddy/pull/6037 :sweat_smile: 

```suggestion
	d.Next() // consume module name
	if d.CountRemainingArgs() > 0 {
		return d.ArgErr()
	}
	for nesting := d.Nesting(); d.NextBlock(nesting); {
		switch d.Val() {
		case "trust_der":
			icp.TrustedCACerts = append(icp.TrustedCACerts, d.RemainingArgs()...)
		default:
			return fmt.Errorf("unrecognized directive: %s", d.Val())
		}
	}
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+		}
+		for d.NextBlock(0) {
+			switch d.Val() {
+			case "trust_pem_file":
+				fcap.TrustedCACertPEMFiles = append(fcap.TrustedCACertPEMFiles, d.RemainingArgs()...)
+			default:
+				return fmt.Errorf("unrecognized directive: %s", d.Val())
+			}
+		}
+	}
```
francislavoie, line 159:
```suggestion
	d.Next() // consume module name
	if d.CountRemainingArgs() > 0 {
		return d.ArgErr()
	}
	for nesting := d.Nesting(); d.NextBlock(nesting); {
		switch d.Val() {
		case "trust_pem_file":
			fcap.TrustedCACertPEMFiles = append(fcap.TrustedCACertPEMFiles, d.RemainingArgs()...)
		default:
			return fmt.Errorf("unrecognized directive: %s", d.Val())
		}
	}
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+		}
+		caPool.AppendCertsFromPEM(pemContents)
+	}
+	f.pool = caPool
+	return nil
+}
+
+// Syntax:
+//
+//	trust_pool	file {
```
francislavoie, line 141:
```suggestion
//	trust_pool file {
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+		caPool.AddCert(ca.RootCertificate())
+	}
+	p.pool = caPool
+
+	return nil
+}
+
+// Syntax:
+//
+//	trust_pool	pki_root [<ca_name>...] {
```
francislavoie, line 212:
```suggestion
//	trust_pool pki_root [<ca_name>...] {
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+//		authority <ca_name>...
+//	}
+//
+// The 'authority' directive can be specified multiple times.
+func (pkir *PKIRootCAPool) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.Next()
```
francislavoie, line 218:
```suggestion
	d.Next() // consume module name
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+		caPool.AddCert(ca.IntermediateCertificate())
+	}
+	p.pool = caPool
+
+	return nil
+}
+
+// Syntax:
+//
+//	trust_pool	pki_intermediate [<ca_name>...] {
```
francislavoie, line 281:
```suggestion
//	trust_pool pki_intermediate [<ca_name>...] {
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+//		authority <ca_name>...
+//	}
+//
+// The 'authority' directive can be specified multiple times.
+func (pic *PKIIntermediateCAPool) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.Next()
```
francislavoie, line 287:
```suggestion
	d.Next() // consume module name
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+		}
+	}
+	ca.pool = caPool
+
+	return nil
+}
+
+// Syntax:
+//
+//	trust_pool	storage [<storage_keys>...] {
```
francislavoie, line 364:
```suggestion
//	trust_pool storage [<storage_keys>...] {
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+//		storage <storage_module>
+//		keys	<storage_keys>...
+//	}
+//
+// The 'keys' directive can be specified multiple times.
+// The'storage' directive is optional and defaults to the default storage module.
+func (sp *StoragePool) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.Next()
+	// if args are given
```
francislavoie, line 373:
```suggestion
	d.Next() // consume module name
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+	// TLS renegotiation level. TLS renegotiation is the act of performing
+	// subsequent handshakes on a connection after the first.
+	// The level can be:
+	//  - "never": (the default) disables renegotiation.
+	//  - "once": allows a remote server to request renegotiation once per connection.
+	//  - "freely": allows a remote server to repeatedly request renegotiation.
+	Renegotiation string `json:"renegotiation,omitempty"`
+}
+
+func (t *TLSConfig) unmarshalCaddyfile(d *caddyfile.Dispenser) error {
```
francislavoie, line 441: If it's unexported then it doesn't conform to the interface. Maybe I'm wrong about how you mean to use this though.

```suggestion
func (t *TLSConfig) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+//		server_name <name>
+//		renegotiation <never|once|freely>
+//
+//	<ca_module> is the name of the CA module to source the trust
+//
+// certificate pool and follows the syntax of the named CA module.
+func (hcp *HTTPCertPool) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.Next()
+
+	// if args are given
```
francislavoie, line 613:
```suggestion
	d.Next() // consume module name
```

modules/caddytls/capools.go
```diff
@@ -0,0 +1,803 @@
...
+// validation step should try to load and provision the guest module to validate
+// the correctness of the configuration. Depeneding on the type of the guest module,
+// the resources may not be available at validation time. It is the
+// operator's responsibility to ensure the resources are available if `EagerValidation`/`eager_validation`
+// is set to `true`.
+//
+// The `backend` directive is required.
+func (lcp *LazyCertPool) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.Next()
+
```
francislavoie, line 710:
```suggestion
	d.Next() // consume module name
```

modules/caddytls/capools.go
mohammed90: This is intentional. I don't want to implement the interface 🙂 just for an internal usage and the method name is handy

francislavoie: I think typically we use `parseConfig` or something instead, if it's not meant to be that interface

mohammed90: but I think the intent here is clearer and requires less inference

francislavoie: :man_shrugging: ok, I don't feel that strongly

mholt:
Wow, I really appreciate all the work that went into these tests. Can't wait to review!!

mholt (approved):
This is a big improvement. 7 client auth provider modules built-in, that's pretty awesome! Thanks for working on this. LGTM.

"Matt, how can you approve a 2,500 line change without any review comments?" -- because Mohammed writes good code and nothing obvious stuck out to me. This is a bigger PR than I usually review, so yes I'm sure some details got overlooked. But I'd rather get this merged in and tested before a release than try to find all the problems with my tired eyes right now. :slightly_smiling_face:

mholt:
Nice work on this :)

Gr33nbl00d:
Correct me if i am wrong but It seems you accidently broke https://github.com/caddyserver/caddy/pull/6022
With your change. The verifiers are not parsed anymore. In the conpollicy go you missed the verifiers section.
It is missing here : https://github.com/caddyserver/caddy/blob/a6d9f9be5b26bf3c7e0947dbba062af0428cda03/modules/caddytls/connpolicy.go#L437

mohammed90:
> Correct me if i am wrong but It seems you accidently broke https://github.com/caddyserver/caddy/pull/6022
> With your change. The verifiers are not parsed anymore. In the conpollicy go you missed the verifiers section.
> It is missing here : https://github.com/caddyserver/caddy/blob/a6d9f9be5b26bf3c7e0947dbba062af0428cda03/modules/caddytls/connpolicy.go#L437
> 
> 
> 

It shouldn't break. I was aware of it and tried my best to avoid breakage. The certificates are still parsed when the directive has the `_leaf` suffix. Do you have a test case/proof of the break?

Gr33nbl00d:
No you got it wrong, i try to explain it better: in the referenced pull request there was support added to load client certificate verifier modules via caddyfile. Which was previously only possible via json config. This has nothing to do with leaf certificates. It is used to do additional checks on client certificates. For example revocation checks via my plugin

https://github.com/Gr33nbl00d/caddy-revocation-validator

In line 437 this seems to be missing:
```
		case "verifier":
			if !d.NextArg() {
				return d.ArgErr()
			}

			vType := d.Val()
			modID := "tls.client_auth." + vType
			unm, err := caddyfile.UnmarshalModule(d, modID)
			if err != nil {
				return err
			}

			_, ok := unm.(ClientCertificateVerifier)
			if !ok {
				return d.Errf("module %s is not a caddytls.ClientCertificatVerifier", modID)
			}
```

Gr33nbl00d:
You can see it here in the old builtins.go line 221 the verifier section was there:
https://github.com/caddyserver/caddy/commit/e965b111cdf46109b4ed87e607e8dce5225105f4#diff-7761c0c9beafcd055ce9c6c894b6c54b888bc00598767551dfcc51b07bb81c36L221

francislavoie:
Yeah I see what you mean @Gr33nbl00d. It got lost in translation somewhere.

I think this happened because ultimately, we don't have any actual implementation of that module interface in Caddy itself, so it's not possible to write a test using anything built-in.

We should probably make a mock cert verifier module in a `_test.go` file which would let us run a Caddyfile adapt test for the verifier option.

PRs welcome!

Gr33nbl00d:
Yes you are right i had the same idea ;) It would  definitly help to harden the code. It was maybe because a merge conflict in the builtin.go file because 2 different prs affected the same file.

Gr33nbl00d:
> Aaah I see your point, @Gr33nbl00d. Sorry 😔 I'll definitely review a PR to add it back (with a test 🙂). If you don't beat me, I'll add it back but it might take me a while.

No problem, this can always happen. We are humans not machines and we make mistakes. I am still in the process of implementing the caddy file support. This is not yet released :) So take your time

PR #5786, opened 2023-08-27
logging: Clone array on log filters, prevent side-effects

francislavoie:
Fixes https://caddy.community/t/is-caddy-mutating-header-content-from-logging-settings/20947

### DIFF
--- a/modules/logging/filters.go
+++ b/modules/logging/filters.go
@@ -100,12 +100,15 @@ func (f *HashFilter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 // Filter filters the input field with the replacement value.
 func (f *HashFilter) Filter(in zapcore.Field) zapcore.Field {
 	if array, ok := in.Interface.(caddyhttp.LoggableStringArray); ok {
+		newArray := make(caddyhttp.LoggableStringArray, len(array))
 		for i, s := range array {
-			array[i] = hash(s)
+			newArray[i] = hash(s)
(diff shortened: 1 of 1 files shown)


francislavoie:
We could probably use a test that confirms the original request isn't modified. I'm not in the headspace to try that right now though.

dunglas (approved):

modules/logging/filters.go
```diff
@@ -100,12 +101,15 @@ func (f *HashFilter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 // Filter filters the input field with the replacement value.
 func (f *HashFilter) Filter(in zapcore.Field) zapcore.Field {
 	if array, ok := in.Interface.(caddyhttp.LoggableStringArray); ok {
-		for i, s := range array {
-			array[i] = hash(s)
+		newArray := slices.Clone(array)
+		for i, s := range newArray {
+			newArray[i] = hash(s)
 		}
+		in.Interface = newArray
```
dunglas, line 108: To be tested, but there is probably no need to copy the existing data:

```suggestion
		newArray := make(caddyhttp.LoggableStringArray, len(array)
		for i, s := range array {
			newArray[i] = hash(s)
		}
		in.Interface = newArray
```

mholt: Thanks, I was going to suggest the same thing.

dunglas (approved):

Issue #5768, opened 2023-08-20
Add govulncheck to the CI


bt90:
This would fit quite well to the testing and quality focused 2.8 milestone.

https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck

hainenber:
hi, can I tackle this issue. Thanks!

bt90:
Fixed by https://github.com/caddyserver/caddy/pull/5790

PR #5790, opened 2023-08-31
chore: check vuln Go mods in CI

hainenber:
Implements #5768
Run `govulncheck` to scan for vulnerable Go packages/modules in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,12 +61,6 @@ jobs:
       with:
         go-version: ${{ matrix.GO_SEMVER }}
         check-latest: true
-    
-    - name: Run govulncheck   
-      uses: golang/govulncheck-action@v1
-      with:
-        go-version-input: ${{ matrix.GO_SEMVER }}
-        go-package: ./...
 
     # These tools would be useful if we later decide to reinvestigate
     # publishing test/coverage reports to some tool for easier consumption
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,12 @@ jobs:
       with:
         go-version: ${{ matrix.GO_SEMVER }}
         check-latest: true
+    
+    - name: Run govulncheck   
+      uses: golang/govulncheck-action@v1
+      with:
+        go-version-input: ${{ matrix.GO_SEMVER }}
+        go-packages: ./...
 
     # These tools would be useful if we later decide to reinvestigate
     # publishing test/coverage reports to some tool for easier consumption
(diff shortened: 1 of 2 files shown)


francislavoie:
I think it should go in https://github.com/caddyserver/caddy/blob/master/.github/workflows/lint.yml, not in `ci.yml`

hainenber:
Initially, I planned to do so. But as `ci.yml` will build `Caddy` for 2 supported version, I decided to run `govulncheck` to ensure vuln-free build artifacts for both as well. What do you think?

francislavoie:
I think we only care for latest Go version. We just keep compat for one prior version for certain packagers who can't bump Go versions immediately.

hainenber:
Thanks for clarification! That simplifies quite a few concerns. I've moved the `govulncheck` step to `lint.yml` now. By default, it'll use the [latest Go version](https://github.com/golang/govulncheck-action#:~:text=latest%20version%20of%20Go) and analyze all packages.

hainenber:
Please help review and approve if the PR's deemed sufficient on your side, thanks!

mholt (commented):
We can give this a try I think. Thanks!

(Edit: Why can't we delete comments!?)

mholt (approved):
We can give this a try I think. Thanks!

hainenber:
hi @mholt, can you help merging this PR if it deemed worth to try. Thanks!

francislavoie:
It seems like govulncheck tries to install Go and checkout the repo itself. It throws a ton of errors from trying to untar the cache as well. I don't think we should use that action, we should probably use the Go tool directly because we already do all the setup ourselves. Or maybe the action has options to turn off Go installation etc? Notice it takes 27s to run + over a minute in a post-step. That's way too long.

mohammed90:
I think it's best if it's split into its own job, separate from golancci-lint run, to separate its failures from the rest.

hainenber:
I could go with @mohammed90's proposal to split the `govulncheck` stage apart from `linting`, it's much more cleaner. Re: @francislavoie's point on using the `govulncheck` binary on our side, the action is pretty much the same concept: there aren't many extraneous parts (https://github.com/golang/govulncheck-action/blob/master/action.yml#L30-L46)

I've pushed a new commit to illustrate my point. Let me know what you folks think of. Thanks!

francislavoie:
That looks better, yeah. 👍

hainenber:
Moving `govulncheck` to its own stage renders the build time really *fast*, [48s](https://github.com/caddyserver/caddy/actions/runs/6086595785/job/16513328223) only.

mohammed90 (approved):
LGTM! Thank you

PR #5791, opened 2023-09-01
httpcaddyfile: fix placeholder shorthands in named routes

### LINKED ISSUE #5788: Placeholder shorthands are not replaced in named routes
The following Caddyfile:

```
&(named) {
    vars Test {header.test}
}

example.com {
    invoke named
    vars Test {header.test}
}
```

Produces the following config:

```json
{
    "apps": {
        "http": {
            "servers": {
                "srv0": {
                    "listen": [
                        ":443"
                    ],
                    "routes": [
                        {
                            "match": [
                                {
                                    "host": [
                                        "example.com"
                                    ]
                                }
                            ],
                            "handle": [
                                {
                                    "handler": "subroute",
                                    "routes": [
                                        {
                                            "handle": [
                                                {
                                                    "Test": "{http.request.header.test}",
                                                    "handler": "vars"
                                                },
                                                {
                                                    "handler": "invoke",
                                                    "name": "named"
                                                }
                                            ]
                                        }
                                    ]
                                }
                            ],
                            "terminal": true
                        }
                    ],
                    "named_routes": {
                        "named": {
                            "handle": [
                                {
                                    "Test": "{header.test}",
                                    "handler": "vars"
                                }
                            ]
                        }
                    }
                }
            }
        }
    }
}

```

I see that named routes are still an experimental feature, so I'm not sure if this behavior is intentional or a known limitation.

If this is not a bug, I think it would be helpful to call out in the documentation. It took me while to debug why my placeholders were not working as I would have expected.

hainenber:
Fixes #5788

Add a section to replace placeholder shorthands for extracted named routes.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -451,7 +451,7 @@ func (ServerType) extractNamedRoutes(
 
 		wholeSegment := caddyfile.Segment{}
 		for _, segment := range sb.block.Segments {
-			// Replace user-defined placeholder shorthands in extracted named routes
+			// replace user-defined placeholder shorthands in extracted named routes
 			replacer.ReplaceSegmentsWithFullPlaceholder(&segment)
 
 			// zip up all the segments since ParseSegmentAsSubroute
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -92,14 +92,7 @@ func (st ServerType) Setup(
 
 	for _, sb := range originalServerBlocks {
 		for _, segment := range sb.block.Segments {
-			for i := 0; i < len(segment); i++ {
-				// simple string replacements
-				segment[i].Text = replacer.simple.Replace(segment[i].Text)
-				// complex regexp replacements
-				for _, r := range replacer.complex {
-					segment[i].Text = r.search.ReplaceAllString(segment[i].Text, r.replace)
-				}
-			}
+			replacer.ReplaceSegmentsWithFullPlaceholder(&segment)
 		}
 
 		if len(sb.block.Keys) == 0 {
@@ -459,14 +452,7 @@ func (ServerType) extractNamedRoutes(
 		wholeSegment := caddyfile.Segment{}
 		for _, segment := range sb.block.Segments {
 			// Replace user-defined placeholder shorthands in extracted named routes
-			for i := 0; i < len(segment); i++ {
-				// simple string replacements
-				segment[i].Text = replacer.simple.Replace(segment[i].Text)
-				// complex regexp replacements
-				for _, r := range replacer.complex {
-					segment[i].Text = r.search.ReplaceAllString(segment[i].Text, r.replace)
-				}
-			}
+			replacer.ReplaceSegmentsWithFullPlaceholder(&segment)
 
 			// zip up all the segments since ParseSegmentAsSubroute
 			// was designed to take a directive+
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -91,8 +91,8 @@ func (st ServerType) Setup(
 	}
 
 	for _, sb := range originalServerBlocks {
-		for _, segment := range sb.block.Segments {
-			replacer.ReplaceSegmentsWithFullPlaceholder(&segment)
+		for i := range sb.block.Segments {
+			replacer.ApplyToSegment(&sb.block.Segments[i])
 		}
 
 		if len(sb.block.Keys) == 0 {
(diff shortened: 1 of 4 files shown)


hainenber:
Using [example](https://github.com/caddyserver/caddy/issues/5788#issue-1869071015) of issue's OP, the Caddyfile

```
&(named) {
    vars Test {header.test}
}

example.com {
    invoke named
    vars Test {header.test}
}
```

got adapted by built `caddy` with above fix, to following content

```
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"match": [
								{
									"host": [
										"example.com"
									]
								}
							],
							"handle": [
								{
									"handler": "subroute",
									"routes": [
										{
											"handle": [
												{
													"Test": "{http.request.header.test}",
													"handler": "vars"
												},
												{
													"handler": "invoke",
													"name": "named"
												}
											]
										}
									]
								}
							],
							"terminal": true
						}
					],
					"named_routes": {
						"named": {
							"handle": [
								{
									"Test": "{http.request.header.test}",
									"handler": "vars"
								}
							]
						}
					}
				}
			}
		}
	}
}
```

francislavoie:
Makes sense, thanks!

Feels a bit dirty to replace the raw handler bytes though. Feels like maybe we should do these replacements earlier in parsing when we have the raw text tokens. Could probably use a refactor. I dunno.

mholt:
That might be a good idea. @hainenber This is a simple patch, thanks for proposing it. How'd you feel about looking into Francis' suggestion?

hainenber:
I'm comfortable with the proposal. I'll push a refactor later for this. Thanks for the proposal!

hainenber:
Hi folks, I've pushed new commit that (a) refactor the shorthand replacer and (b) perform replacement for named routes when they've just got extracted with raw text tokens. Please have a look. Thanks

francislavoie:
That's nice!

- Maybe we should call the new file "shorthands.go" since it's a bit more specific as to what's in it. Replacer is a more general concept that exists in another package.
- The loop is still repeated. Maybe we can add a method to the shorthand replacer that takes a segment and applies the replacements to it?

hainenber:
Thanks Francis for your review 🙏 ! They're all reasonable to me. I've pushed new commits accordingly. 
Thanks for your patience and please have a look again 😅

caddyconfig/httpcaddyfile/shorthands.go
```diff
@@ -0,0 +1,93 @@
...
+		"{client_ip}", "{http.vars.client_ip}",
+	}
+}
+
+// ReplaceSegmentsWithFullPlaceholder replaces shorthand placeholder to its full placeholder, understandable by Caddy.
+func (s ShorthandReplacer) ReplaceSegmentsWithFullPlaceholder(segment *caddyfile.Segment) {
+	if segment != nil {
+		for i := 0; i < len(*segment); i++ {
+			// simple string replacements
+			// (*segment)[i].Text
```
francislavoie, line 85: I think this was a leftover?

```suggestion
```

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -477,10 +449,13 @@ func (ServerType) extractNamedRoutes(
 			continue
 		}
 
-		// zip up all the segments since ParseSegmentAsSubroute
-		// was designed to take a directive+
 		wholeSegment := caddyfile.Segment{}
 		for _, segment := range sb.block.Segments {
+			// Replace user-defined placeholder shorthands in extracted named routes
```
francislavoie, line 454: Just a consistency thing

```suggestion
			// replace user-defined placeholder shorthands in extracted named routes
```

caddyconfig/httpcaddyfile/shorthands.go
```diff
@@ -0,0 +1,93 @@
...
+		"{tls_client_subject}", "{http.request.tls.client.subject}",
+		"{tls_client_certificate_pem}", "{http.request.tls.client.certificate_pem}",
+		"{tls_client_certificate_der_base64}", "{http.request.tls.client.certificate_der_base64}",
+		"{upstream_hostport}", "{http.reverse_proxy.upstream.hostport}",
+		"{client_ip}", "{http.vars.client_ip}",
+	}
+}
+
+// ReplaceSegmentsWithFullPlaceholder replaces shorthand placeholder to its full placeholder, understandable by Caddy.
+func (s ShorthandReplacer) ReplaceSegmentsWithFullPlaceholder(segment *caddyfile.Segment) {
```
francislavoie, line 81: I think the fn could be named something like `ApplyToSegment`. The type is already a "replacer" so it's redundant, and we already know what it's meant to do from its name.

Kinda subjective, idk if you have thoughts on that @mholt 🤷‍♂️

francislavoie (approved):
Looks good to me!

francislavoie:
Ah looks like the linter doesn't like the dereferencing stuff. I wonder if there's another way we could do it that can avoid that? 🤔

caddyconfig/httpcaddyfile/shorthands.go
hainenber: I've updated the PR with commit changing that name. It makes sense to me. Thanks for the review, Francis!

hainenber:
@francislavoie I've fixed the linter issue with dereferencing and also add a small change to `govulncheck`. Please help taking a look if it's deemed sufficient. Thanks!

francislavoie:
Thank you!

PR #5793, opened 2023-09-01
reverseproxy: Add more debug logs

mholt:
This makes debug logging very noisy when reverse proxying, but I guess that's the point.

This has shown to be useful in troubleshooting vexing infrastructure problems.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -194,7 +194,9 @@ type Handler struct {
 	// If set, the proxy will write very detailed logs about its
 	// inner workings. Enable this only when debugging, as it
 	// will produce a lot of output.
-	TraceLogs bool `json:"trace_logs,omitempty"`
+	//
+	// EXPERIMENTAL: This feature is subject to change or removal.
+	VerboseLogs bool `json:"verbose_logs,omitempty"`
 
 	Transport        http.RoundTripper `json:"-"`
 	CB               CircuitBreaker    `json:"-"`
@@ -948,13 +950,13 @@ func (h *Handler) finalizeResponse(
 	}
 
 	rw.WriteHeader(res.StatusCode)
-	if h.TraceLogs {
+	if h.VerboseLogs {
 		logger.Debug("wrote header")
 	}
 
 	err := h.copyResponse(rw, res.Body, h.flushInterval(req, res), logger)
 	errClose := res.Body.Close() // close now, instead of defer, to populate res.Trailer
-	if h.TraceLogs || errClose != nil {
+	if h.VerboseLogs || errClose != nil {
 		logger.Debug("closed response body from upstream", zap.Error(errClose))
 	}
 	if err != nil {
@@ -990,7 +992,7 @@ func (h *Handler) finalizeResponse(
 		}
 	}
 
-	if h.TraceLogs {
+	if h.VerboseLogs {
 		logger.Debug("response finalized")
 	}
 
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -943,9 +943,11 @@ func (h *Handler) finalizeResponse(
 	}
 
 	rw.WriteHeader(res.StatusCode)
+	logger.Debug("wrote header")
 
-	err := h.copyResponse(rw, res.Body, h.flushInterval(req, res))
-	res.Body.Close() // close now, instead of defer, to populate res.Trailer
+	err := h.copyResponse(rw, res.Body, h.flushInterval(req, res), logger)
+	errClose := res.Body.Close() // close now, instead of defer, to populate res.Trailer
+	logger.Debug("closed response body from upstream", zap.Error(errClose))
 	if err != nil {
 		// we're streaming the response and we've already written headers, so
 		// there's nothing an error handler can do to recover at this point;
@@ -978,6 +980,8 @@ func (h *Handler) finalizeResponse(
 			rw.Header().Add(k, v)
 		}
 	}
+
+	logger.Debug("response finalized")
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -205,19 +206,22 @@ func (h Handler) copyResponse(dst http.ResponseWriter, src io.Reader, flushInter
...
 	defer streamingBufPool.Put(buf)
-	_, err := h.copyBuffer(w, src, *buf)
+	_, err := h.copyBuffer(w, src, *buf, logger)
 	return err
 }
 
 // copyBuffer returns any write errors or non-EOF read errors, and the amount
 // of bytes written.
-func (h Handler) copyBuffer(dst io.Writer, src io.Reader, buf []byte) (int64, error) {
+func (h Handler) copyBuffer(dst io.Writer, src io.Reader, buf []byte, logger *zap.Logger) (int64, error) {
```
mohammed90, line 215: The logger is already stored in:
https://github.com/caddyserver/caddy/blob/1b73e3862d312ac2057265bf2a5fd95760dbe9da/modules/caddyhttp/reverseproxy/reverseproxy.go#L213

any reason why not use it as `h.logger.[...]`?

mholt: This logger has added information like the request, which we found was really important for adding context so we could form a cohesive story. Otherwise we found ourselves asking, "Is this from OUR request or another one?"

francislavoie:
Hmm... lol yeah that is a lot of extra logging. 

Should we tuck it behind an option on the handler to make it a bit less ridiculous by default? I'm worried about the usual cycle of helping on the forums, might be too much logs to sift through on a regular basis.

Maybe just an extra `debug: true` option on the handler + `debug` subdirective 🤔 or maybe called `tracing`?

mholt:
I was wondering about that. I'm still formulating my opinion of debug logs, but so far...

I feel like debug logs should be very noisy. They're for debugging, so their purpose is to give you a lot of information. You generally will only enable them when troubleshooting. IMO if you enable debug logs you should expect a firehose. Our current debug logs are a bit anemic :)

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -479,9 +493,11 @@ func (m *maxLatencyWriter) delayedFlush() {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	if !m.flushPending { // if stop was called but AfterFunc already started this goroutine
+		m.logger.Debug("delayed flush is not pending")
 		return
 	}
 	//nolint:errcheck
+	m.logger.Debug("delayed flush")
 	m.flush()
```
francislavoie, line 501:
```suggestion
	m.logger.Debug("delayed flush")
	//nolint:errcheck
	m.flush()
```

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -452,25 +461,30 @@ type maxLatencyWriter struct {
...
 
 func (m *maxLatencyWriter) Write(p []byte) (n int, err error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	n, err = m.dst.Write(p)
+	m.logger.Debug("wrote bytes", zap.Int("n", n), zap.Error(err))
 	if m.latency < 0 {
 		//nolint:errcheck
+		m.logger.Debug("flushing immediately")
 		m.flush()
```
francislavoie, line 475:
```suggestion
		m.logger.Debug("flushing immediately")
		//nolint:errcheck
		m.flush()
```

francislavoie:
But if the firehose can only be parsed by an expert (us) then it's less useful. For 95% of the time there's only one log from the proxy that's useful, i.e. the one with the request/response logged. The rest are just checkpoints and are only useful if trying to trace some edgecase behaviour.

mholt:
> But if the firehose can only be parsed by an expert (us) then it's less useful.

It's true, but I'd say a good portion of the time, when there is an issue, it ends up on our forums or issue tracker so we do end up looking at it anyway.

Can we try this, pretty please? And then maybe we can put these behind an extra config option if needed? :pleading_face:

francislavoie:
Fine :sweat_smile: but I might complain again when we get terribly noisy logs on the forums that are hard to make sense of :man_shrugging:

mholt:
Fair! We can adjust if this is toooooo unwieldy.

francislavoie:
Every time I look at this PR, I'm scared, honestly. The amount of logs that'll get emitted by `copyBuffer` will be insane. I really think it should be opt-in on the proxy handler. :grimacing:

mholt:
To be fair, debug logs are already opt-in :no_mouth:

francislavoie:
Lemme just implement a `trace_logs` boolean option real quick.

There's a difference between needing to debug basic request/response, and needing to debug buffers/streaming. This PR as-is makes it impossible/unwieldy to _only_ do the former, especially for users without understanding of Caddy's inner-workings.

francislavoie (approved):

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -191,6 +191,11 @@ type Handler struct {
 	// - `{http.reverse_proxy.header.*}` The headers from the response
 	HandleResponse []caddyhttp.ResponseHandler `json:"handle_response,omitempty"`
 
+	// If set, the proxy will write very detailed logs about its
+	// inner workings. Enable this only when debugging, as it
+	// will produce a lot of output.
+	TraceLogs bool `json:"trace_logs,omitempty"`
```
mholt, line 197: "Trace" isn't the word that correlates in my head, at least, to "verbose debug logs." Can we call this something else? Like "VerboseLogs"?

Also, we should probably document this as experimental and subject to change, as I'm not sure what the best way to configure this is; or the best name for it. But I think `VerboseLogs` could be a good start. What do you think?

francislavoie: Sure :+1:

francislavoie: Done

PR #5795, opened 2023-09-02
Add: request_body replace directive

AdrienPensart:
Hello, 

here is a need I have, I use caddy as a reverse proxy to EdgeDB endpoint, and declare custom routes to query Edgedb.
The endpoint act as an alias for my query, so no query appears in the javascript on client side.

```
:80 {
    file_server browse
    root * hyperedge
    handle /artists {
        method POST
        request_body {
            replace `{"query": "select artists {*}"}`
        }
        request_header Content-Type application/json
        rewrite * /db/edgedb/edgeql
        reverse_proxy http://localhost:10701
    }
}
```

Tell if it's a good idea or if the name of the directive should be something else.


francislavoie:
This is interesting, but a bit strange of a feature. I think we'll need to do a bit of thinking about how we want to approach this. It might take us a bit longer to decide on this one.

Thanks for the contribution and idea!

modules/caddyhttp/requestbody/requestbody.go
```diff
@@ -31,6 +32,7 @@ type RequestBody struct {
 	// The maximum number of bytes to allow reading from the body by a later handler.
 	// If more bytes are read, an error with HTTP status 413 is returned.
 	MaxSize int64 `json:"max_size,omitempty"`
+	Replace string `json:"replace,omitempty"`
```
mholt, line 35: Would need a helpful godoc comment on this field.

mholt (commented):
Oh, so this actually completely drops the upstream response body and replaces it with a new one entirely, _not_ performing replacements in the response body.

Should `r.Body` be closed before it is set to a new value?

Like Francis said, this is a bit unusual and I'm not sure what the implications will be.

I guess it's a very simple change. (We're on feature freeze until after 2.8, so even if we merge this in, it won't be released until 2.9.)

Why is this needed instead of a simple `respond` directive?

steffenbusch:
Hi @mholt 
> Oh, so this actually completely drops the upstream response body and replaces it with a new one entirely, _not_ performing replacements in the response body.

I think it's only about the **Request** body. 

Today, I have been playing with [Oracle REST Data Services / ORDS](https://www.oracle.com/database/technologies/appdev/rest.html) where you can POST sql statements to a service endpoint of ORDS such as:

```
$ cat simple_query.json
{ "statementText":"SELECT TO_DATE('01-01-1976','dd-mm-yyyy') FROM dual;"}
$ curl -i -X POST --user DEMO:demo --data-binary "@simple_query.json" -H "Content-Type: application/json" -k http://localhost:8088/ords/demo/_/sql
```

Having a `reverse_proxy` to `http://localhost:8088/`, you could hide the SQL from users/clients with the configration of @AdrienPensart  and his proposal of `request_body` such as:

```
:443 {
    handle /1976 {
        method POST
        request_body {
            replace `{ "statementText":"SELECT TO_DATE('01-01-1976','dd-mm-yyyy') FROM dual;"}`
        }
        request_header Content-Type application/json
        rewrite * /ords/demo/_/sql
        reverse_proxy http://localhost:8088 {
            header_up Authorization "Basic REVNTzpkZW1v"
        }
    }
}
```

I thought I just add this use case here where this `request_body` + `replace` could be beneficial

francislavoie:
Yeah - FWIW I thought this was about the response body at first as well, but this is definitely about the _request_ body @mholt. Not the same as https://github.com/caddyserver/replace-response, entirely unrelated in fact.

In terms of Caddyfile API, I'm considering whether we should make it the first argument to `request_body` so you don't need to make a block to use `replace` (though we can keep `replace` as a long-form for people who want it). See https://caddyserver.com/docs/caddyfile/directives/request_body it could look like:

```
request_body [<matcher>] [<replacement>] {
	replace  <replacement>
	max_size <value>
}
```

mholt:
Oh, I see -- thanks for pointing that out.

I like Francis' suggestion to make it work inline, i.e.: `request_body "contents here"` -- also, I want to give second thought to the term "replace" which, for me, makes me think of performing replacements within the request body, not replacing the whole request body. But I can see how both could be accurate. Hmm.

modules/caddyhttp/requestbody/requestbody.go
```diff
@@ -42,6 +44,10 @@ func (RequestBody) CaddyModule() caddy.ModuleInfo {
 }
 
 func (rb RequestBody) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
+	if rb.Replace != "" {
+		r.Body = io.NopCloser(strings.NewReader(rb.Replace))
```
mholt, line 48: If we're completely replacing the request body, we miiiight want to call `r.Body.Close()` before re-assigning a new body. Yes the std lib will close it for us when the request is over, but would closing it early (and releasing those resources) cause any problems?

Also, I wonder if we should add placeholder support and expand those when doing the body replacement.

AdrienPensart: I just don't know how placeholder are working right now, will search.

mholt:
Looking good, but my remaining concern is closing the request body. The std lib will close the original request body, I guess, and we don't need to close the new one since it's just a string buffer?

Have you verified that there is not a memory leak?

modules/caddyhttp/requestbody/requestbody.go
mholt: Actually, can we rename this to `Set`? I keep thinking of "replacing content within the body" whereas "setting the body" is more precise IMO.

mholt (changes requested):
So, I don't know when things changed (maybe I was looking at an old diff) but apparently the body is closed now. That's good.

Sorry for my severe delay. But what do you think about renaming `Replace` to `Set`?

I also wonder... I have an idea... should it be modular? i.e. are there multiple ways to get what the request body should be set to (e.g. the response to an HTTP request)?

steffenbusch:
Hi @mholt,

is there any chance this PR could be merged in time for v2.9.0? It would enable some practical use cases, including one where a simple GET request to Caddy (e.g., `http://localhost:8080/vat-check?no=LU22046007`) could be transformed under the hood into a REST API call to the EU VAT service.

Here’s an example Caddyfile configuration snipping illustrating this:

```caddyfile
handle /vat-check {
        # Define a matcher to check if the `no` query parameter is missing
        @noQueryParamNo {
                vars {http.request.uri.query.no} ""
        }
        # Respond with an error message if the `no` query parameter is missing
        respond @noQueryParamNo "Missing query-parameter ?no=VAT-NO" 400

        # Extract the first two characters as the country code and the rest as the VAT number
        # If no match is found, use default values "XX" and "1234567890" to trigger an error
        map {http.request.uri.query.no} {country} {vatNumber} {
                ~^(.{2})(.+)$ "${1}" "${2}"
                default "XX" "1234567890"
        }

        # Set the request body for the reverse proxy, including the extracted country code and VAT number
        request_body {
                set `{"countryCode": "{country}","vatNumber": "{vatNumber}"}`
        }

        # Forward the request to the external VAT-checking API with the modified request body
        reverse_proxy {
                to https://ec.europa.eu
                method POST
                rewrite /taxation_customs/vies/rest-api/check-vat-number
        }
}
```

modules/caddyhttp/requestbody/caddyfile.go
```diff
@@ -68,8 +68,14 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
...
 
+		case "set":
+			var setStr string
+			if !h.AllArgs(&setStr) {
+				return nil, h.ArgErr()
+			}
+			rb.Set = setStr
 		default:
-			return nil, h.Errf("unrecognized request_body subdirective '%s'", h.Val())
+			return nil, h.Errf("unrecognized set_body subdirective '%s'", h.Val())
```
steffenbusch, line 78: I think this should stay `request_body`

modules/caddyhttp/requestbody/requestbody.go
```diff
@@ -43,6 +44,8 @@ type RequestBody struct {
 	// EXPERIMENTAL. Subject to change/removal.
 	WriteTimeout time.Duration `json:"write_timeout,omitempty"`
 
+	Set string `json:"set,omitempty"`
```
mholt, line 47: Let's add a godoc comment here, along with marking it as experimental.

mholt (changes requested):
Sorry it's taken me so long to get to this. Happy to merge this, but first I want to make sure we properly document the new field.

modules/caddyhttp/requestbody/requestbody.go
```diff
@@ -60,6 +65,16 @@ func (rb *RequestBody) Provision(ctx caddy.Context) error {
 }
 
 func (rb RequestBody) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
+	if rb.Set != "" {
+		err := r.Body.Close()
```
steffenbusch, line 69: Potential panic risk: `r.Body.Close()` may cause a nil pointer dereference  

Currently, `r.Body.Close()` is called without checking if `r.Body` is `nil`. While `r.Body` is typically not `nil` in standard HTTP requests (defaulting to `http.NoBody`), some middleware or handlers may explicitly set it to `nil`. This can lead to a runtime panic if `r.Body.Close()` is called on a `nil` reference.  

From [`net/http` docs](https://pkg.go.dev/net/http#NoBody):  
> "An alternative, however, is to simply set [Request.Body] to nil."

steffenbusch: > si if Body is already nil, we just skip the replacement ?

I would suggest making the call to `r.Body.Close()` conditional but still performing the replacement if `r.Body` is `nil`, in case some other middleware (for whatever strange reason) has set it to `nil`.
```go
if rb.Set != "" {
	if r.Body != nil {
		err := r.Body.Close()
		if err != nil {
			return err
		}
	}

	replacedBody := repl.ReplaceAll(rb.Set, "")
	r.Body = io.NopCloser(strings.NewReader(replacedBody))
	r.ContentLength = int64(len(rb.Set))
}
```

mholt (approved):
LGTM now -- we can give it a try, thank you! And thanks @steffenbusch for the comments as well.

steffenbusch:
I have compiled Caddy from master and did some testing, but found an issue. 
If `set` is using placeholders like this

```caddyfile
:7070 {
	request_body {
		set `Remote: {http.request.remote}` # length of set with placeholder is 29
	}

	reverse_proxy {
		to http://localhost:9001
		method POST
	}
}
```
the calculation of `r.ContentLength` currently considers `rb.Set` and will set Content-Length to 29.
But the Body after replacing the placeholders is 23 for "Remote: 127.0.0.1:37872"

Error-Log:
```
ERROR   http.log.error  http: ContentLength=29 with Body length 23      {"request": {"remote_ip": "127.0.0.1", "remote_port": "37872", "client_ip": "127.0.0.1", "proto": "HTTP/1.1", "method": "GET", "host": "localhost:7070", "uri": "/ping", "headers": {"User-Agent": ["Go-http-client/1.1"], "Accept-Encoding": ["gzip"]}}, "duration": 0.000332537, "status": 502, "err_id": "n3xgx3290", "err_trace": "reverseproxy.statusError (reverseproxy.go:1390)"}
```

I'll open a pull request to fix this issue.

PR #5794, opened 2023-09-01
acmeserver: support specifying the allowed challenge types

mohammed90:
TODO:
- [ ] Tests
- [ ] Docs

There will be another PR for the ability to control the allowed names, denied names, and wild cards.

PR #5796, opened 2023-09-03
acmeserver: add policy field to define allow/deny rules

mohammed90:
TODO:
- [x] Tests
- [x] Docs
- [x] Caddyfile support

To be merged after #5794


mholt:
Cool, this looks great so far. Again, no real strong feedback yet -- it's similar to how I'd probably do it if I had need of this.

hslatman:
@mohammed90 you could drop the URI domains from this. They're only used for when there are URIs in a certificate (request), and those are not a valid ACME identifier type. 

Let me know if you hit issues with the policy configuration / evaluation. We kept it fairly simple intentionally, and sometimes it can be too limiting (e.g. strict equality check of the common name), so happy to hear feedback about it.

mohammed90:
> @mohammed90 you could drop the URI domains from this. They're only used for when there are URIs in a certificate (request), and those are not a valid ACME identifier type.

Thanks for the tip! I'll make the change when I circle back to this PR. Currently the progress on this PR is blocked #5794 since both touch the same file. I'm stuck on testing the ACME server due to my knowledge gap of how to use github.com/mholt/acmez 😶

modules/caddypki/acmeserver/policy.go
```diff
@@ -0,0 +1,83 @@
...
+	// to be approved or denied for certificates. Non-CIDR IP addresses
+	// are matched exactly.
+	IPRanges []string `json:"ip_ranges,omitempty"`
+}
+
+// Policy defines the criteria for the ACME server
+// of when to issue a certificate. Refer to the
+// [Certificate Issuance Policy](https://smallstep.com/docs/step-ca/policies/)
+// on Smallstep website for the evaluation criteria.
+type Policy struct {
```
mholt, line 27: Nit, but move this to the top of the file, it makes it easy to see what this file is about; whereas RuleSet is just a member of Policy.

mholt (approved):
Awesome, looking good and clean and tested. Just one random thought but nothing of importance. Thank you Mohammed, this is great!

modules/caddypki/acmeserver/policy.go
mohammed90: Done

Issue #5802, opened 2023-09-05
Prevent overriding environment variables with --envfile


evandam:
Hi folks,

I'm not sure if this is considered a bug or "expected behavior," but `--envfile` overwrites any existing environment variables, which seems like the opposite of other dotenv libraries such as https://github.com/bkeepers/dotenv.

Similarly, it seems like if `godotenv` was used instead of a custom parser, this logic would be the same: https://github.com/joho/godotenv#precedence--conventions

For example, I would expect `FOO` to be `system`, but instead Caddy reads `.env` here:

```bash
$ cat .env
FOO=.env

$ FOO=system caddy run --envfile .env --environ
...
FOO=.env
...
```

If this is expected behavior, would it be possible to get an optional flag to prevent this from happening?

For example, a common use case may be to have a Kubernetes ConfigMap mounted with environment-specific environment variables, but you still want to have a `.env` file to fallback on, if it makes sense.

I opened a PR if it's alright, looking forward to hearing your thoughts 🙌 https://github.com/caddyserver/caddy/pull/5803

Thanks in advance!

mholt:
Thanks for opening an issue! (And PR.)

I don't use this flag myself, but if the convention for this file is to not overwrite existing values, then I suppose that's a good fix.

I wonder how many people are relying on the buggy behavior? :no_mouth:

evandam:
That was my one hesitation as well 😅

I can't imagine _that many_ people relying on `.env` to override an env var since they're typically used (in my experience, at least) to supplement existing env vars. Famous last words, anyway.

Let me know if there's anything else I can help out with on my end, I'm not much of a Go developer but I can certainly try 🙌

evandam:
Thanks for help on this one @mholt and @francislavoie! Not rushing at all, but I was just curious if you have a rough timeline for when we can expect the next release with this included?

mholt:
Not sure yet, but you can always build from that commit if you'd like! :+1:

francislavoie:
We don't really have any scheduled releases. We release when we feel we have enough for one or if there's a fix that's important enough to warrant immediate release. That said 2.7.5 should probably not be too far away. Maybe a couple weeks to a month from now 🤷‍♂️

evandam:
Gotcha, looking forward to it! Appreciate all the work that goes into Caddy :heart:

francislavoie:
[...]

PR #5803, opened 2023-09-06
cmd: Prevent overwriting existing env vars with `--envfile`

evandam:
Fix: https://github.com/caddyserver/caddy/issues/5802

Prevent `--envfile` from overwriting existing environment variables.

For example:

```bash
$ cat .env
FOO=.env
FOO_DOT_ENV="set in .env"

$ cat .env.prod
FOO=.env.prod
FOO_DOT_ENV_TEST="set in .env.test"

# NOTE: --envfile loads env vars that don't exist
$ go run cmd/caddy/main.go run --envfile .env --environ | grep FOO
FOO=.env
FOO_DOT_ENV=set in .env

# NOTE: multiple --envfile passed get loaded, in decreasing precedence
$ go run cmd/caddy/main.go run --envfile .env.prod --envfile .env --environ | grep FOO
FOO=.env.prod
FOO_DOT_ENV_TEST=set in .env.test
FOO_DOT_ENV=set in .env

# NOTE: Existing env vars are preserved if they're also found in --envfiles
$ FOO=system go run cmd/caddy/main.go run --envfile .env.prod --envfile .env --environ | grep FOO
FOO=system
FOO_DOT_ENV_TEST=set in .env.test
FOO_DOT_ENV=set in .env
```

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -104,7 +104,7 @@ using 'caddy run' instead to keep it in the foreground.
 		CobraFunc: func(cmd *cobra.Command) {
 			cmd.Flags().StringP("config", "c", "", "Configuration file")
 			cmd.Flags().StringP("adapter", "a", "", "Name of config adapter to apply")
-			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
+			cmd.Flags().StringSliceP("envfile", "", []string{}, "Environment file to load")
 			cmd.Flags().BoolP("watch", "w", false, "Reload changed config file automatically")
 			cmd.Flags().StringP("pidfile", "", "", "Path of file to which to write process ID")
 			cmd.RunE = WrapCommandFuncForCobra(cmdStart)
@@ -150,7 +150,7 @@ option in a local development environment.
 		CobraFunc: func(cmd *cobra.Command) {
 			cmd.Flags().StringP("config", "c", "", "Configuration file")
 			cmd.Flags().StringP("adapter", "a", "", "Name of config adapter to apply")
-			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
+			cmd.Flags().StringSliceP("envfile", "", []string{}, "Environment file to load")
 			cmd.Flags().BoolP("environ", "e", false, "Print environment")
 			cmd.Flags().BoolP("resume", "r", false, "Use saved config, if any (and prefer over --config file)")
 			cmd.Flags().BoolP("watch", "w", false, "Watch config file for changes and reload it automatically")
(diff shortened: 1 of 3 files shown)


mholt (approved):
Thanks, this LGTM. I did notice that it also accepts a slice of envfiles now, instead of a single file. That's not going to break anything, right?

evandam:
> Thanks, this LGTM. I did notice that it also accepts a slice of envfiles now, instead of a single file. That's not going to break anything, right?

I haven't seen any issues in any of the cases I could think of 😅:

- passing no `--envfile`
- passing one `--envfile`
- passing multiple `--envfile`

Maybe worth noting you can also pass comma-separated like `--envfile ".env.test,.env"` as explained in https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#StringSlice

It just makes it a bit more flexible for "cascading" .env files.

cmd/commands.go
```diff
@@ -104,7 +104,7 @@ using 'caddy run' instead to keep it in the foreground.
 		CobraFunc: func(cmd *cobra.Command) {
 			cmd.Flags().StringP("config", "c", "", "Configuration file")
 			cmd.Flags().StringP("adapter", "a", "", "Name of config adapter to apply")
-			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
+			cmd.Flags().StringSliceP("envfile", "", []string{}, "Environment file to load")
```
francislavoie, line 107:
```suggestion
			cmd.Flags().StringSliceP("envfile", "", []string{}, "Environment file(s) to load")
```

cmd/commands.go
```diff
@@ -150,7 +150,7 @@ option in a local development environment.
...
```
francislavoie, line 153:
```suggestion
			cmd.Flags().StringSliceP("envfile", "", []string{}, "Environment file(s) to load")
```

cmd/commands.go
```diff
@@ -301,7 +301,7 @@ the KEY=VALUE format will be loaded into the Caddy process.
 		CobraFunc: func(cmd *cobra.Command) {
 			cmd.Flags().StringP("config", "c", "", "Input configuration file")
 			cmd.Flags().StringP("adapter", "a", "", "Name of config adapter")
-			cmd.Flags().StringP("envfile", "", "", "Environment file to load")
+			cmd.Flags().StringSliceP("envfile", "", []string{}, "Environment file to load")
```
francislavoie, line 304:
```suggestion
			cmd.Flags().StringSliceP("envfile", "", []string{}, "Environment file(s) to load")
```

francislavoie (approved):
Thanks, LGTM :+1:

Issue #5804, opened 2023-09-07
Add support for configuring trusted_proxies XFF depth


nebez:
This relates to https://github.com/caddyserver/caddy/pull/5104

Firstly, wonderful software that really makes great default decisions. Thanks for caddy.

I suggest adding a configuration parameter to `trusted_proxies` which would allow deciding the depth at which the Client IP is in the `X-Forwarded-For` (referred to as XFF in this issue) stack of IPs. This is a feature request but also a bug/security issue. Currently, it is very easy to misuse `trusted_proxies` due to how XFF handling is implemented in caddy.

In the original PR, it is noted:

> This implementation takes the left-most, first valid IP.

Taking the left-most value is actually the least trustworthy value in nearly all cases. [Wikipedia entry re: XFF](https://en.wikipedia.org/wiki/X-Forwarded-For):

> Since it is easy to forge an X-Forwarded-For field the given information should be used with care. The right-most IP address is always the IP address that connects to the last proxy, which means it is the most reliable source of information.

And since adam-p published his article on XFF, MDN has since updated their documentation as well to reflect the security risk in trusting the left-most IP address ([thanks to a PR from adam-p](https://github.com/mdn/content/pull/13838)).

### Recommendation

There are two ways to manage XFF headers safely. The first is by specifying the depth at which your caddy reverse proxy exists (trusted proxy count), and the second is finding the first right-most IP address that isn't in your trusted proxy list (trust proxy list).

Currently caddy pulls the left-most valid IP address. This is not how most reverse proxies work with the header. As far as I know, there is only 1 reverse proxy I know of that behaves **very incorrectly**, which is Microsoft's IIS. It inserts hops to the right of the header instead of the left, which would make finding the left-most the appropriate approach. Instead, caddy should apply the same logic but find the right-most untrusted address. This would be the second **Trusted proxy list** approach, described above.

However to maintain backwards compatibility, you can also consider adding the **Trusted proxy count** option. This is as simple as defining a number that specifies how from from the right is the client IP address in the XFF header.

e.g. a header value of `X-Forwarded-For: <spoofed address>, <client address>, <proxy 1 address>, <proxy 2 address>`, we would use `xff_depth 3` to pull the real `client_ip` address out and ignore the spoofed address.

[...]

PR #5805, opened 2023-09-07
caddyhttp: Security enhancements for client IP parsing

nebez:
This is the first stab at addressing #5804 

Commits are readable in order. This review is easiest done by going through them one-by-one:
- cover existing `client_ip_headers` functionality
- cover existing `trusted_proxy` functionality with custom ranges
- write tests covering both scenarios above in combination
- introduce a failing test proving the potential security weakness of existing `trusted_proxy` + `client_ip_headers` configuration
- introduce a new configuration option (default: false) to opt-in to a safer method of parsing `client_ip_headers`

The first 3 commits provide confidence in maintaining backwards compatibility for those who rely on the existing, left-most parsing of XFF headers. The remaining commits introduce tests + config options (default off) to enable the new right-most, first untrusted IP XFF parsing.

### DIFF
--- a/caddyconfig/httpcaddyfile/serveroptions.go
+++ b/caddyconfig/httpcaddyfile/serveroptions.go
@@ -34,22 +34,22 @@ type serverOptions struct {
 	ListenerAddress string
 
 	// These will all map 1:1 to the caddyhttp.Server struct
-	Name                   string
-	ListenerWrappersRaw    []json.RawMessage
-	ReadTimeout            caddy.Duration
-	ReadHeaderTimeout      caddy.Duration
-	WriteTimeout           caddy.Duration
-	IdleTimeout            caddy.Duration
-	KeepAliveInterval      caddy.Duration
-	MaxHeaderBytes         int
-	EnableFullDuplex       bool
-	Protocols              []string
-	StrictSNIHost          *bool
-	TrustedProxiesRaw      json.RawMessage
-	EnableStrictXffParsing bool
-	ClientIPHeaders        []string
-	ShouldLogCredentials   bool
-	Metrics                *caddyhttp.Metrics
+	Name                 string
+	ListenerWrappersRaw  []json.RawMessage
+	ReadTimeout          caddy.Duration
+	ReadHeaderTimeout    caddy.Duration
+	WriteTimeout         caddy.Duration
+	IdleTimeout          caddy.Duration
+	KeepAliveInterval    caddy.Duration
+	MaxHeaderBytes       int
+	EnableFullDuplex     bool
+	Protocols            []string
+	StrictSNIHost        *bool
+	TrustedProxiesRaw    json.RawMessage
+	TrustedProxiesStrict bool
+	ClientIPHeaders      []string
+	ShouldLogCredentials bool
+	Metrics              *caddyhttp.Metrics
 }
 
 func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (any, error) {
@@ -218,11 +218,11 @@ func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (any, error) {
 				)
 				serverOpts.TrustedProxiesRaw = jsonSource
 
-			case "enable_strict_xff_parsing":
+			case "trusted_proxies_strict":
 				if d.NextArg() {
 					return nil, d.ArgErr()
 				}
-				serverOpts.EnableStrictXffParsing = true
+				serverOpts.TrustedProxiesStrict = true
 
 			case "client_ip_headers":
 				headers := d.RemainingArgs()
@@ -347,7 +347,7 @@ func applyServerOptions(
 		server.StrictSNIHost = opts.StrictSNIHost
 		server.TrustedProxiesRaw = opts.TrustedProxiesRaw
 		server.ClientIPHeaders = opts.ClientIPHeaders
-		server.EnableStrictXffParsing = opts.EnableStrictXffParsing
+		server.TrustedProxiesStrict = opts.TrustedProxiesStrict
 		server.Metrics = opts.Metrics
 		if opts.ShouldLogCredentials {
 			if server.Logs == nil {
--- a/caddyconfig/httpcaddyfile/serveroptions.go
+++ b/caddyconfig/httpcaddyfile/serveroptions.go
@@ -46,7 +46,7 @@ type serverOptions struct {
 	Protocols            []string
 	StrictSNIHost        *bool
 	TrustedProxiesRaw    json.RawMessage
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/server_test.go
```diff
@@ -144,3 +145,207 @@ func BenchmarkServer_LogRequest_WithTraceID(b *testing.B) {
...
+	}
+
+	req := httptest.NewRequest("GET", "/", nil)
+	req.RemoteAddr = "10.0.0.1:12345"
+	req.Header.Set("X-Forwarded-For", "30.30.30.30, 45.54.45.54, 10.0.0.1")
+
+	trusted, clientIP := determineTrustedProxy(req, server)
+
+	assert.True(t, trusted)
+	assert.Equal(t, clientIP, "45.54.45.54")
```
nebez, line 350: In reality, caddy returns `30.30.30.30` here. The validation logic in caddy is: if remoteAddress is trusted, take first value from trusted XFF header (x-forwarded-for by default, or others).

The validation logic should instead be:

if remote address is trusted, starting from the right of XFF header, return the first IP address that is untrusted.

mholt:
Awesome, thanks for working on this! :)

caddyconfig/httpcaddyfile/serveroptions.go
```diff
@@ -217,6 +218,12 @@ func unmarshalCaddyfileServerOptions(d *caddyfile.Dispenser) (any, error) {
 				)
 				serverOpts.TrustedProxiesRaw = jsonSource
 
+			case "enable_strict_xff_parsing":
```
francislavoie, line 221: I don't like this name because we may not actually be parsing XFF because of `ClientIPHeaders`. I think something like `trusted_proxies_strict` or `trusted_proxies_mode strict` (if we want to have more than one mode) might be more sensible; uses a similar prefix to the related option so the relationship between them is more obvious.

nebez: great call. can you point me in the right direction of the Golang-way of introducing an option set? e.g. "strict", "left-most", etc. do you just do string matching in the `serveroptions.go` parser?

francislavoie: Oh sorry, I saw your comment and meant to reply but got sidetracked.

Yeah we generally just do string matching for modes, with a `switch` statement etc. Sometimes we'll use a const in code. See the `auto_https` option for examples of modes.

nebez:
@francislavoie I've renamed the options, I like this new name better. Ty for suggestion.

@francislavoie @mholt can I suggest this be a patch version of current 2.7.x release? This is a backwards compatible change and keeps the left-most parsing as a default. As-is, the `trusted_proxies` default parsing of XFF headers opts users into a non-configurable behaviour that injects insecure `client_ip` headers (and then forwards them down). Unless you have 100% trust in every upstream, this is the least secure approach to XFF parsing per the 2 sources in my original issue.

For my web services behind caddy, I've disabled `trusted_proxies` and manually pass XFF headers to downstreams. This lets me extract the right-most client IP myself downstream, but at the cost of losing `client_ip`'s in caddy. For us this is an acceptable tradeoff off until the mode can be made strict (right-most parsing) via this option. I imagine more users would also expect similar behaviour.

In addition, if you point me to the right direction, I will also include changes to the documentation of the website to mention this new opt-in option. Users should understand the behaviour of left-most vs. right-most XFF parsing and the security dangers of using left-most on untrusted upstreams.

caddyconfig/httpcaddyfile/serveroptions.go
nebez: In https://adam-p.ca/blog/2022/03/x-forwarded-for/ – the article details only one scenario where you'd want left-most from a security perspective: if you have a non-conformant load balancer that appends XFF to the wrong side. Otherwise, right-most is the appropriate approach for security and correctness, and left-most is the appropriate approach for the rest.

Let me know what you think. Happy to change this to a switch statement if there could be more.

mholt:
Sorry for my latency on this, been a very busy time -- it is still on my radar though! I will circle back soon enough. 

> can I suggest this be a patch version of current 2.7.x release?

I think we can do that since this is basically a bugfix / not a new feature.

nebez:
No worries.

When you've got the time, can you please also review https://github.com/caddyserver/website/pull/346 as part of the same change?

mholt:
Yes we will try to do that :D

(I will be a little busy for a while with some family obligations.)

lachlan2k:
Great PR! I look forward to support for this case landing in Caddy.

I must say though, I prefer the option that was suggested in the original issue (#5804) of instead having `trusted_proxy_count <n>`, as this adds a lot more flexibility, and conveniently matches the terminology of [the MDN docs on XFF](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#selecting_an_ip_address).

There are a lot of use cases where there could be an intermediary proxy, such as WAF -> Load Balancer -> Caddy, where it is useful to trust the *second* right-most value. I.e., trusting `C` in `X-Forwarded-For: A, B, C, D`.

Matching the functionality of this PR would be achieved by setting `trusted_proxy_count 1`, and the use case I described above would be `trusted_proxy_count 2`.

I'd love to hear other perspectives on this.

mholt:
I'll also be able to revisit this PR soon @nebez

francislavoie (approved):
LGTM! Sorry for the delay. We had a feature freeze which we eventually cancelled, then the holidays.

Thanks for the thorough tests. This is only an opt-in behaviour change, so I'm comfortable to merge this safely. If there's any issues that crop up, they can be fixed in a followup.

francislavoie:
Actually, I'll make a tiny adjustment to this right now as per @lachlan2k's comments.

I think it's valid to say that count should be supported. Since the current "strict" behaviour effectively maps to count of "1", we can change the type from `bool` to `int` and make it 1 if set. That way, we can later implement `trusted_proxies_strict <count>` without it being a breaking change.

I don't want to block merging this any longer for an additional feature that can be added after the fact.

nebez:
> Actually, I'll make a tiny adjustment to this right now as per @lachlan2k's comments.
> 
> I think it's valid to say that count should be supported. Since the current "strict" behaviour effectively maps to count of "1", we can change the type from `bool` to `int` and make it 1 if set. That way, we can later implement `trusted_proxies_strict <count>` without it being a breaking change.
> 
> I don't want to block merging this any longer for an additional feature that can be added after the fact.

That sounds quite pragmatic to me. Thanks for tackling that @francislavoie 

Once you're comfortable merging + tagging it, I'll update https://github.com/caddyserver/website/pull/346 and then it will be ready for your review

PR #5806, opened 2023-09-08
fileserver: optimize browse template to avoid SVG icon duplication

glowinthedark:
### Description

Multiple file entries of the same type inline the entire SVG icon content which results in the same SVG icons being rendered in HTML multiple times which increases file size.  

1. The PR reduces the size of generated HTML by using SVG xrefs instead of inlining
2. Minor ui fix: in search field use `<input type="search">` instead of `text`
3. Add folder-symlink icon, as current master displays folder symlinks as file symlinks.

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -46,7 +46,14 @@ func cmdStart(fl Flags) (int, error) {
 	startCmdConfigAdapterFlag := fl.String("adapter")
 	startCmdPidfileFlag := fl.String("pidfile")
 	startCmdWatchFlag := fl.Bool("watch")
-	startCmdEnvfileFlag := fl.String("envfile")
+
+	var err error
+	var startCmdEnvfileFlag []string
+	startCmdEnvfileFlag, err = fl.GetStringSlice("envfile")
+	if err != nil {
+		return caddy.ExitCodeFailedStartup,
+			fmt.Errorf("reading envfile flag: %v", err)
+	}
 
 	// open a listener to which the child process will connect when
 	// it is ready to confirm that it has successfully started
@@ -70,8 +77,9 @@ func cmdStart(fl Flags) (int, error) {
 	if startCmdConfigFlag != "" {
 		cmd.Args = append(cmd.Args, "--config", startCmdConfigFlag)
 	}
-	if startCmdEnvfileFlag != "" {
-		cmd.Args = append(cmd.Args, "--envfile", startCmdEnvfileFlag)
+
+	for _, envFile := range startCmdEnvfileFlag {
+		cmd.Args = append(cmd.Args, "--envfile", envFile)
 	}
 	if startCmdConfigAdapterFlag != "" {
 		cmd.Args = append(cmd.Args, "--adapter", startCmdConfigAdapterFlag)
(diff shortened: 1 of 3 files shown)


francislavoie:
Good idea!

This does have one downside though that for very small directories, it'll ship every SVG even if there's only one that might be used.

Your approach is certainly better for huge directories, but it's a bit worse for small directories.

The absolute-best option would be to only add the SVG ref if there's a relevant file included (and only include it once) but I'm not sure if we have sufficient template tooling to make that easy to do :thinking:

Maybe we could loop through the directory listing once in Go code to collect a list of unique extensions (or a `map[string]bool` to allow O(1) lookups), and then pass that to the template and wrap each SVG ref with a simple `if extensions contains 'ext'` or w/e (pseudocode).

glowinthedark:
@francislavoie: indeed, I didn't consider that the way it works now it will only pull the extensions that are used, yet it all probably comes down to estimating the cost: the unminified SVG for all icons takes up 14KB, and when minified it's 8.8KB — this is cheap, although accounting is always hard. )) I didn't check how many files would need to be in a directory in order to generate a bigger file size.

glowinthedark:
Out of curiosity generated some stats:

Number of files | Master branch | this PR
-- | -- | --
10 | 33 KB | 45 KB
20 | 44 KB | 53 KB
50 | 75 KB | 78 KB
100 |  127 KB | 116 KB
500 |  546 KB |  436 KB
1000 | 1046 KB |  861 KB

All files were generated of .txt type with this bash script

```bash
mkdir dir{10,20,50,100,500,1000}
touch dir10/{1..10}.txt
touch dir20/{1..20}.txt
touch dir50/{1..50}.txt
touch dir100/{1..100}.txt
touch dir500/{1..500}.txt
touch dir1000/{1..1000}.txt
```

With up to 50 files the master version is more efficient, and the optimization will only make sense with folders with over 100 items. Do you think it is worth optimizing? I am in two minds.. wdyt?

francislavoie:
Thanks for those stats! Very useful.

Also worth checking is gzipped size (if you didn't already). I think repeated SVG will compress _very_ well.

But it's true that linked SVG will probably perform better in the browser (refs instead of copies) though that's also a micro optimization.

I agree I lean towards `master`, unless you're interested in going the extra mile with my suggestion of only shipping SVG for extensions seen.

glowinthedark:
Keeping things simple is probably a higher priority than achieving some imaginary 'optimization' at the cost of adding extra complexity. I've updated the pr to only add the `folder-symlink` and `type=search` for the search field — it adds a tiny ⛌ button for resetting the search when it has text; not a big deal but comes for free from the browser. I suppose that extra filtering on the go side would require extra traversals so I feel the extra complexity is not worth the cost.

Slight off-topic: the template currently does not maintain the list-vs-grid state on navigation clicks. Is this a known issue and is there some plan to fix it in the future?

My current quick and ugly fix which I don't really like is to hijack link clicks when grid mode is active with something like this:

```js
document.onclick = onDocumentClickHandler;

function onDocumentClickHandler(e) {
  let link = e.target;

  if (window.location.search) {
    if (link.tagName !== "A") {
      link = findClosest(link, "A");
    }

    if (link) {
      const href = link.getAttribute("href");

      if (href != "#" && !href.startsWith("javascript:")) {
        event.preventDefault();
        event.stopImmediatePropagation();
        document.location.href = `${link.href}${window.location.search}`;
      }
    }
  }
}

function findClosest(el, tagName, className) {
  if (
    (el.tagName === tagName && !className) ||
    el.classList.contains(className)
  ) {
    return el;
  } else if (el.parentElement) {
    return findClosest(el.parentElement, tagName, className);
  }
  return null;
}
```

I suppose there can be a better and nicer way, but asking here in case you think this sort of quick fix might be appropriate and can be included in the pr. Or do you think the issue is big enough in order to deserve its own thread?

PR #5807, opened 2023-09-09
caddyhttp: Export `LengthReader`

### LINKED ISSUE #5728: bytes_read for hijacked connections?
Is there anyway to add `bytes_read` to the entry if the connection is hijacked, like in forwardproxy? `lengthReader` is not exported. 🤔

francislavoie:
Closes #5728

Exporting it allows other modules to look at the length at any point in time, which may be useful if they hijack the connection.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -329,9 +329,9 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		// wrap the request body in a LengthReader
 		// so we can track the number of bytes read from it
-		var bodyReader *lengthReader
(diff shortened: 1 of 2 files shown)


WeidiDeng:
Exporting this type isn't enough for hijacked connections to see how many bytes are read. As this body is actually `http.NoBody` for `GET` request which are usually the hijackable requests.

Also, even if the body is readable (no eof), it has nothing to do with the bytes counting, as the body reads from the connection while `Hijack` returns the underlying connection.

francislavoie:
I'm not sure I can finish this off, I'll leave it to @Mygod who requested it to make the necessary adjustments.

PR #5808, opened 2023-09-09
httpcaddyfile: Enable TLS for catch-all site if `tls` directive is specified

### LINKED ISSUE #5747: caddyfile: When using non-standard port catch-all site with `tls` directive, should use TLS
Currently, this doesn't work:

```
:8443 {
    tls internal {
        on_demand
    }
}
```

Instead, we need to do this:

```
https://:8443 {
    tls internal {
        on_demand
    }
}
```

This is not intuitive. Using the `tls` directive should make the site address have TLS enabled without having to specify `https://`.

The rule as I understand it right now is if there's a domain with a non-standard port, it'll enable TLS, but if there's no domain then it uses HTTP. But it should also use TLS if the `tls` directive is specified.

The key difference can be seen with `caddy adapt --pretty`, without `https://`, there's no `tls_connection_policies` for the server. It should have at least an empty connection policy when `tls` is used.

hainenber:
## PR description
Enable TLS for catch-all site if `tls` directive is specified, even with non-standard port

## What does this PR fix/implement?
Fixes #5747 

### Acceptance test
The input Caddyfile

```
:8443 {
    tls internal {
        on_demand
    }
}
```

is `adapt --pretty` to

```
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":8443"
					],
					"tls_connection_policies": [
						{}
					]
				}
			}
		},
		"tls": {
			"automation": {
				"policies": [
					{
						"issuers": [
							{
								"module": "internal"
							}
						],
						"on_demand": true
					}
				]
			}
		}
	}
}
```

Regression test
The input Caddyfile
```
:8443 {

}
```

is `adapt --pretty` to

```
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":8443"
					]
				}
			}
		}
	}
}
```

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -716,10 +716,16 @@ func (st *ServerType) serversFromPairings(
 					}
 				}
 
+				// If TLS is specified as directive, it will also result in 1 or more connection policy being created
+				// Thus, catch-all address with non-standard port, e.g. :8443, can have TLS enabled without
+				// specifying prefix "https://"
+				createdTLSConnPolicies, ok := sblock.pile["tls.connection_policy"]
+				hasTLSDirectiveSpecified := ok && len(createdTLSConnPolicies) > 0
+
 				// we'll need to remember if the address qualifies for auto-HTTPS, so we
 				// can add a TLS conn policy if necessary
 				if addr.Scheme == "https" ||
-					(addr.Scheme != "http" && addr.Host != "" && addr.Port != httpPort) {
+					(addr.Scheme != "http" && addr.Port != httpPort && hasTLSDirectiveSpecified) {
 					addressQualifiesForTLS = true
 				}
 				// predict whether auto-HTTPS will add the conn policy for us; if so, we
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -723,13 +723,13 @@ func (st *ServerType) serversFromPairings(
 				// ensuring compatibility with behavior described in below link
 				// https://caddy.community/t/making-sense-of-auto-https-and-why-disabling-it-still-serves-https-instead-of-http/9761
 				createdTLSConnPolicies, ok := sblock.pile["tls.connection_policy"]
-				hasTLSDirectiveSpecified := (ok && len(createdTLSConnPolicies) > 0) ||
+				hasTLSEnabled := (ok && len(createdTLSConnPolicies) > 0) ||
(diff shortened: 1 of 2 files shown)


francislavoie:
Please add some adapt tests. See https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt. Each file is a Caddyfile and the adapted JSON output with 10 dashes separating them.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -719,8 +719,12 @@ func (st *ServerType) serversFromPairings(
 				// If TLS is specified as directive, it will also result in 1 or more connection policy being created
 				// Thus, catch-all address with non-standard port, e.g. :8443, can have TLS enabled without
 				// specifying prefix "https://"
+				// Second part of the condition is to allow creating TLS conn policy even though `auto_https` has been disabled
+				// ensuring compability with behavior described in below link
```
francislavoie:
```suggestion
				// ensuring compatibility with behavior described in below link
```

hainenber: Yes, I've fixed this in latest commit ;)
Nice catch though!

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -716,10 +716,20 @@ func (st *ServerType) serversFromPairings(
...
 				}
 
+				// If TLS is specified as directive, it will also result in 1 or more connection policy being created
+				// Thus, catch-all address with non-standard port, e.g. :8443, can have TLS enabled without
+				// specifying prefix "https://"
+				// Second part of the condition is to allow creating TLS conn policy even though `auto_https` has been disabled
+				// ensuring compatibility with behavior described in below link
+				// https://caddy.community/t/making-sense-of-auto-https-and-why-disabling-it-still-serves-https-instead-of-http/9761
+				createdTLSConnPolicies, ok := sblock.pile["tls.connection_policy"]
+				hasTLSDirectiveSpecified := (ok && len(createdTLSConnPolicies) > 0) ||
```
francislavoie, line 726: I think naming the variable like this would be more correct, because TLS can be enabled by things other than the TLS directive, as the condition implies.

```suggestion
				hasTLSEnabled := (ok && len(createdTLSConnPolicies) > 0) ||
```

hainenber: I've renamed the variabled to make it more variable. PTAL. Thanks!

francislavoie (approved):
LGTM, thanks!

PR #5812, opened 2023-09-10
fileserver: browse template SVG icons and UI tweaks

glowinthedark:
### Description

- continuation of #5806 (my bad; github closed the pr automatically after force pushing a clean branch; lesson learnt 😄 )

1. Add folder symlink icon as current master displays folder symlinks as file symlinks 
2. Make `<path>` tags self-closing in SVG.
3. Minor ui fix: in search field use `<input type="search">` instead of `text`

---
@francislavoie: in reply to https://github.com/caddyserver/caddy/pull/5806#issuecomment-1712279472 and https://github.com/caddyserver/caddy/pull/5806#issuecomment-1712104514
Keeping things simple is probably a higher priority than achieving some imaginary 'optimization' at the cost of adding extra complexity. I've updated the pr to only add the `folder-symlink` and `type=search` for the search field — it adds a tiny ⛌ button for resetting the search when it has text; not a big deal but comes for free from the browser. I suppose that extra filtering on the go side would require extra traversals so I feel the extra complexity is not worth the cost.

Also added a null-check for `el.querySelector('.sizebar-bar')` which is not present in grid mode.

Slight off-topic: the template currently does not maintain the list-vs-grid state on navigation clicks. Is this a known issue and is there some plan to fix it in the future?

My current quick fix which I don't really like is to hijack link clicks when grid mode is active with something like this:

```js
document.onclick = onDocumentClickHandler;

function onDocumentClickHandler(e) {
  let link = e.target;

  if (window.location.search) {
    if (link.tagName !== "A") {
      link = findClosest(link, "A");
    }

    if (link && link.matches(".file a, .entry a, h1 a")) {
      e.preventDefault();
      e.stopImmediatePropagation();
      document.location.href = `${link.href}${window.location.search}`;
    }
  }
}

function findClosest(el, tagName, className) {
  if (el.tagName === tagName &&
    (!className || el.classList.contains(className))) {
    return el;
  } else if (el.parentElement) {
    return findClosest(el.parentElement, tagName, className);
  }
  return null;
}

```

An alternative approach could be attaching click event listeners to _all_ links with css `class=entry|file` but I wanted to avoid that.

I suppose there can be a better and nicer way, but asking here in case you think this sort of quick fix might be appropriate and can be included in the pr. Or do you think the issue is big enough in order to deserve its own thread?

### DIFF
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -1,288 +1,296 @@
 {{- define "icon"}}
 	{{- if .IsDir}}
-	<svg xmlns="http://www.w3.org/2000/svg" class="icon icon-tabler icon-tabler-folder-filled" width="24" height="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round">
-		<path stroke="none" d="M0 0h24v24H0z" fill="none"></path>
-		<path d="M9 3a1 1 0 0 1 .608 .206l.1 .087l2.706 2.707h6.586a3 3 0 0 1 2.995 2.824l.005 .176v8a3 3 0 0 1 -2.824 2.995l-.176 .005h-14a3 3 0 0 1 -2.995 -2.824l-.005 -.176v-11a3 3 0 0 1 2.824 -2.995l.176 -.005h4z" stroke-width="0" fill="currentColor"></path>
-	</svg>
+		{{- if .IsSymlink}}
+		<svg xmlns="http://www.w3.org/2000/svg" class="icon icon-tabler icon-tabler-folder-filled" width="24" height="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round">
+			<path stroke="none" d="M0 0h24v24H0z" fill="none"/>
+			<path d="M9 3a1 1 0 0 1 .608 .206l.1 .087l2.706 2.707h6.586a3 3 0 0 1 2.995 2.824l.005 .176v8a3 3 0 0 1 -2.824 2.995l-.176 .005h-14a3 3 0 0 1 -2.995 -2.824l-.005 -.176v-11a3 3 0 0 1 2.824 -2.995l.176 -.005h4z" stroke-width="0" fill="currentColor"/>
+			<path fill="#000" d="M2.795 17.306c0-2.374 1.792-4.314 4.078-4.538v-1.104a.38.38 0 0 1 .651-.272l2.45 2.492a.132.132 0 0 1 0 .188l-2.45 2.492a.381.381 0 0 1-.651-.272V15.24c-1.889.297-3.436 1.39-3.817 3.26a2.809 2.809 0 0 1-.261-1.193Z" style="stroke-width:.127478"/>
+		</svg>
+		{{- else}}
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/fileserver/browse.html
```diff
@@ -1075,8 +1083,11 @@ <h1>
 				});
 				document.querySelectorAll('.size').forEach(el => {
 					const size = Number(el.dataset.size);
-					el.querySelector('.sizebar-bar').style.width = `${size/largest * 100}%`;
-				});
+                    const sizebar = el.querySelector('.sizebar-bar');
+                    if (sizebar) {
+                        sizebar.style.width = `${size/largest * 100}%`;
+                    }
+                });
```
francislavoie, line 1090: Spaces -> tabs
```suggestion
					const sizebar = el.querySelector('.sizebar-bar');
					if (sizebar) {
						sizebar.style.width = `${size/largest * 100}%`;
					}
				});
```

glowinthedark:
a shorter version of the above js would be

```js
document.querySelectorAll('a').forEach(link => {
  link.addEventListener('click', event => {
    event.preventDefault();
    window.location.href = `${link.href}${window.location.search}`;
  });
});
```

(I've only used `document.onclick` because in my custom version I need to handle that even anyway)

mholt (approved):
Thanks for this change! Looks good.

> Slight off-topic: the template currently does not maintain the list-vs-grid state on navigation clicks. Is this a known issue and is there some plan to fix it in the future?

Yeah, I was aware of that when I implemented it. I haven't planned to make the layout sticky but I guess we could do that if people prefer; I just wasn't sure whether that is the right thing to do. Just because this folder is shown in grid view, does that mean the next one we click into should be also?

(If we do it, we'll just need to do it without cookies.)

We could maybe do it the same way we preserve the sort settings... with a little server-side help through the query string.

glowinthedark:
@mholt: persistent layout mode is probably for most people already muscle memory; and it's kind of unexpected when it is 'forgotten'. All file navigation UI's persist the mode, including mac finder, windows explorer, even linux file managers; but maybe it's too small of a nuisance for people to even bother to complain about it 😄 

btw, maybe writing it to `localStorage` would make more sense?

mholt:
@glowinthedark That sounds good to me. The main problem is that the server needs to know about the layout choice. So that's why I chose query string.

Do we want to merge this in, and then we can address that in another PR?

glowinthedark:
@mholt: yes, the grid mode persistence is outside of the scope of this pr; I only mentioned it in order to understand better the context; there is nothing more to add here in terms of the template code

as for the offtopic regarding the grid mode I have no clue what a clean way might look like; an even uglier than js approach could be storing a flag in `localStorage` and then on page load if the flag is found (preferably before anything is rendered) check for the flag and redirect from `/some/dir` to `/some/dir/?layout=grid` which probably even worse than click interception.

mholt:
Awesome, thank you for this PR!

Yeah, I'm not sure the _best_ way to go about the grid persistence thing. Might need to experiment...

glowinthedark:
@mholt Thank you Matt for taking time to look at it! 

As for the grid more I'll keep thinking.. A wild idea would be to experiment with a PWA-style template that won't need a page reload at all and would render everything client-side exploiting the fact that `file_server` is capable of serving JSON when called with `Content-Type: application/json` which can be htmlized with [moustache](https://github.com/janl/mustache.js) or raw js, might not be practical but could be fun to play with. 😄

mholt:
Thanks. Preferably I'd like to keep this a simple server-rendered app if possible... but yes we could potentially do crazy things. Maybe we could persist the layout in the query string, we'd just have to rewrite the link hrefs.

PR #5828, opened 2023-09-20
reverseproxy: fix parsing Caddyfile fails for unlimited request/response buffers

mcfedr:
When trying to set the `requst_buffers` or `response_buffers` to `-1` for unlimited, go-humanise fails to parse the value and crashes caddy

This allows the special value of `-1` to be set in the Caddyfile, this seemed like a good simple way to do this, as no other negative values would be valid.

It was always possible to set `-1` using JSON or the API.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/reverse_proxy_buffers.txt
+++ b/caddytest/integration/caddyfile_adapt/reverse_proxy_buffers.txt
@@ -1,0 +1,58 @@
+https://example.com {
+	reverse_proxy https://localhost:54321 {
+		request_buffers unlimited
+		response_buffers unlimited
+	}
+}
+
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":443"
+					],
+					"routes": [
+						{
+							"match": [
+								{
+									"host": [
+										"example.com"
+									]
+								}
+							],
+							"handle": [
+								{
+									"handler": "subroute",
+									"routes": [
+										{
+											"handle": [
+												{
+													"handler": "reverse_proxy",
+													"request_buffers": -1,
+													"response_buffers": -1,
+													"transport": {
+														"protocol": "http",
+														"tls": {}
+													},
+													"upstreams": [
+														{
+															"dial": "localhost:54321"
+														}
+													]
+												}
+											]
+										}
+									]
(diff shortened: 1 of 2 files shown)


francislavoie:
I don't see the need for this. By default, the buffers are disabled, so setting them to `-1` doesn't do anything.

mcfedr:
`-1` isnt disabled, its unlimited size buffer.

mholt:
@francislavoie You're right, but I think this PR is related to #5830 where -1 is being introduced as a special value for changing the Content-Length. I need to understand it better though.

mcfedr:
@mholt I wasnt intending that it was related - I discovered this as i has having issues with request_buffers, and decided that an unlimited buffer might help - then i discovered i couldnt set it in my caddyfile

`-1` is already used as a value that makes the buffers unlimited insize - without the other PR, its valid to set -1 (actually any negative number - https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/reverseproxy.go#L1116 but a warning for -1, https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/reverseproxy.go#L249)

mholt:
Yeah, I noticed that after I reviewed that PR. Makes sense though. We'll probably merge this in soon.

(Unless, do we want to make some syntactic sugar instead of `-1`, the user could say `unlimited` or something?)

mcfedr:
I did wonder that - might be nicer than a magic number - shall i swap it out?

mcfedr:
@mholt updated

mholt (approved):
This LGTM now. :) Thank you!

francislavoie:
This will need to be manually rebased, seems like a conflict was introduced.

mcfedr:
@francislavoie I've rebased on master and force pushed. Not sure why but I don't have the button to let you push to the branch anywhere.

francislavoie:
See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork, there should be a checkbox on the right sidebar to allow it.

mcfedr:
I think its because this fork isnt in my personal account but in my company team

francislavoie:
That would explain it, yeah.

Issue #5829, opened 2023-09-20
content-length is wrong when the request doesnt fit in the request_buffer


mcfedr:
how to do it

setup a reverse proxy with request_buffers

```Caddyfile
reverse_proxy example.com {
    request_buffers 1000
}
```

then send a request bigger than the buffer (i.e. > 1000 bytes in this case)

the content-length header to the backend will be set to `1000`, but the body will actually be bigger, its `io.MultiReader(buf, originalBody),` - so the buffer plus the rest.

different backends have slightly different failure modes for this, im seeing this in my caddy logs

```
2023/09/20 19:38:08.374	ERROR	http.log.error	net/http: HTTP/1.x transport connection broken: http: ContentLength=1000 with Body length 64301	{"request": {"remote_ip": "127.0.0.1", "remote_port": "55244", "client_ip": "127.0.0.1", "proto": "HTTP/1.1", "method": "POST", "host": "127.0.0.1:8888", "uri": "/anything", "headers": {"User-Agent": ["curl/8.1.2"], "Accept": ["*/*"], "Accept-Encoding": ["deflate, gzip"], "Content-Type": ["text/plain"], "Content-Length": ["64301"]}}, "duration": 0.000745583, "status": 502, "err_id": "150qhej7z", "err_trace": "reverseproxy.statusError (reverseproxy.go:1248)"}
```

caused by this https://github.com/caddyserver/caddy/commit/2182270a2cb2861f5358bb41b81caee49bb7d0dd

I would just offer a PR with a revert, but not sure how that affects the reason this commit was added

PR #5830, opened 2023-09-20
reverseproxy: only change the content-length header when the full request was buffered

mcfedr:
fixes: https://github.com/caddyserver/caddy/issues/5829

currently if a request body was longer than `request_buffers` the `content-length` header sent to the upstream will be wrong, it will be the length of the buffer.

With this change, only in the case that the full request was read into the buffer, because it was set to unlimited length then the `content-length` is changed

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -582,8 +582,12 @@ func (h Handler) prepareRequest(req *http.Request, repl *caddy.Replacer) (*http.
 	// feature if absolutely required, if read timeouts are
 	// set, and if body size is limited
 	if h.RequestBuffers != 0 && req.Body != nil {
-		req.Body, req.ContentLength = h.bufferedBody(req.Body, h.RequestBuffers)
-		req.Header.Set("Content-Length", strconv.FormatInt(req.ContentLength, 10))
+		var readBytes int64
+		req.Body, readBytes = h.bufferedBody(req.Body, h.RequestBuffers)
+		if h.RequestBuffers == -1 {
+			req.ContentLength = readBytes
+			req.Header.Set("Content-Length", strconv.FormatInt(req.ContentLength, 10))
+		}
 	}
 
 	if req.ContentLength == 0 {


mholt:
Ohh I see, this is a good point. Thanks for finding this.

So I guess we need to set Content-Length to keep php-fpm (and other upstreams) from hanging. We only know the ContentLength in these cases:

- Buffer size is -1 (unlimited)
- Buffer is larger than body

We don't know it in these cases:

- Buffer is smaller than body
- There's an error reading the body up to the buffer size

In the first two cases, we should be setting ContentLength.

This change sets ContentLength if buffer size is unlimited, but not if a limited-size buffer contains the entire body. Have you tested and accounted for that case as well?

mcfedr:
in the case of 'Buffer is larger than body' - I did think about it - and in the current version, it wont set the content-length - but my thinking is this - if you are relying on the content-length always being set and the body being buffered - then using anything other than an unlimited buffer is going to cause you problems - as the buffer size doesnt limit the size of the request, and these other bytes will be in the body stream.

I would suggest for php-fpm you want to use unlimited buffer size, and possible limit the request size using the setting from [request_body](thttps://caddyserver.com/docs/caddyfile/directives/request_body)

mholt:
Sorry, I don't think I follow:

> in the case of 'Buffer is larger than body' - I did think about it - and in the current version, it wont set the content-length - but my thinking is this - if you are relying on the content-length always being set and the body being buffered - then using anything other than an unlimited buffer is going to cause you problems - as the buffer size doesnt limit the size of the request, and these other bytes will be in the body stream.

We're talking about the case where the ENTIRE body fits and is read into the buffer, right? There's no bytes remaining on the wire, and we know the exact ContentLength. I think in that case we could set the ContentLength (esp. if it's not set already).

Now if the body size does exceed the buffer size, then it's true: we don't know the full length and probably shouldn't change the ContentLength.

Does that make sense?

mcfedr:
> We're talking about the case where the ENTIRE body fits and is read into the buffer, right? There's no bytes remaining on the wire, and we know the exact ContentLength. I think in that case we could set the ContentLength (esp. if it's not set already).

You are completely right - and it could be the right way to go - but - my thought it that if Caddy does that it allows you to create a dangerous situation.

- if php-fpm requires content-length to be set - so you enable buffering, and now its set - but as soon as someone sends your server a request that is longer than your buffer, php-fpm is going to crash. 
- so if you are required to have a content-length header, you must have an unlimited size buffer
- so maybe its best to only 'fix' the content-length when you set the buffer to unlimited

mholt (approved):
I see! That does make sense :thinking: 

I guess we can try it and see how it goes -- I suppose this is still better than what we currently have. :)

If @francislavoie is good with it then we can merge this in.

mcfedr:
interesting thing I've noticed going back to the original issue (https://github.com/caddyserver/caddy/issues/5420) from php-fpm - nginx will fully buffer requests for fpm backends by default, https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html?#fastcgi_request_buffering

mcfedr:
Just rebased to keep up with master.

Is there anything I can do to help get this merged? I'd love to switch back to a release build of caddy

trea:
Thanks Francis, I will check this out today and get an update to you all. 🤝

ueffel:
I had a similar issue with the [cgi plugin](github.com/aksdb/caddy-cgi). In order to set the correct `Content-Length`, which is required by the CGI protocol, the whole body of the request had to be buffered. An additional challenge was that the original http requests had `Transfer-Encoding: chunked`, so there was no `Content-Length` initially.
I had a look how nginx deals with this problem. Of course: request buffering. But how to limit memory usage?

> so if you are required to have a content-length header, you must have an unlimited size buffer

Theoretically yes

> nginx will fully buffer requests for fpm backends by default

Yes, but nginx buffers request bodies in temporary files. I don't know if there even is buffering in memory.[^1]

To solve the problem for the cgi plugin, [I implemented a hybrid solution](https://github.com/aksdb/caddy-cgi/pull/8/commits/051a35c94f5a33c60708d083beabc545e9f4ea4c): Buffer in memory if it fits within the limits. If it doesn't, buffer in a temporary file.

Maybe this idea could be implemented here as well? It would limit the memory usage and use hard disk space temporarily, which is typically more available :)

What would interest me: How does caddy deal with requests that are `Transfer-Encoding: chunked`, when reverse proxying especially with a `fastcgi` transport?

[^1]: I looked at the nginx documentation again. It does [memory buffering up to a limit](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size) as well.

mholt:
@ueffel That's interesting. Yeah, writing to files could be a thing, but then we risk exhausting disk space :man_shrugging: AND it's basically a swap file, so there's a huge performance hit by orders of magnitude. If enough people would find it useful, though, perhaps we could support it and make it opt-in...

mcfedr:
@mholt what would it take to get this PR merged? The current behaviour is buggy, and this should fix it. There is clearly room for future work, but it would appear to not be a blocker

mholt:
@mcfedr I was mainly waiting for @trea to to make sure it works for them, due to the sensitivity of this change.

trea:
Sorry friends, I did try to verify this before but I wasn't able to recreate the previous circumstances in my test setup so I didn't have any confidence in being able to verify or not at that time. I should have updated this thread at that time. I'd recommend moving forward with this and hope you'll accept my apologies!

mholt:
No worries, thanks for letting us know :)

I don't see any immediate issues with this change other than the "what if the body fits within the limited-size buffer" case, which was discussed above. We can try it this way for now -- it's still an improvement probably -- and hopefully @mcfedr can help us patch it if needed later :smiley:

PR #5832, opened 2023-09-21
reverseproxy: SRV dynamic upstream failover

### LINKED ISSUE #5816: Support falling back to cached upstreams IPs when DNS lookups fail
We'd like to request support for temporarily falling back to cached upstream IP addresses in the event an upstream DNS query fails. We have a situation where an upstream service cluster can temporarily report as unhealthy, even though the vast majority of the upstream node IPs are continuing to function, and we'd like to not immediately fail all requests.

This situation looks like:
* We have a dynamic upstream service which is doing load balancing by an HTTP header that resolves to multiple IPs
* This cluster becomes temporarily unavailable and SRV lookups return an NXDOMAIN from our service mesh
* Caddy fails the DNS lookup and returns a 502

We'd like to be able to tell Caddy to use the cached IPs for a specified period of time, this would help insulate flakey services.

Let me know if I can provide more information, thank you!

mholt:
This should fix #5816 - if the SRV lookup fails and a GracePeriod is defined, it will continue to use the previously-cached records until the GracePeriod is up, after which it will try again.

@cds2-stripe @jjiang-stripe Feel free to give this a shot -- let me know if I need to rebase this onto another branch.


kkroo:
Thanks for this fix @mholt I’m wondering why it’s only for SRV and not A lookups for dynamic upstreams?

mholt:
It was only requested for SRV upstreams for the moment. If it for sure works we can easily add it for A upstreams too :+1:

modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -140,6 +147,12 @@ func (su SRVUpstreams) GetUpstreams(r *http.Request) ([]*Upstream, error) {
 		// out and an error will be returned alongside the remaining results, if any." Thus, we
 		// only return an error if no records were also returned.
 		if len(records) == 0 {
+			if su.GracePeriod > 0 {
+				su.logger.Error("SRV lookup failed; using previously cached", zap.Error(err))
+				cached.freshness = time.Now().Add(-time.Duration(su.Refresh) - time.Duration(su.GracePeriod))
```
mholt, line 152: I'm second-guessing my math here.

I should write some tests for this. I did a couple in my head but I should probably just write them down.

francislavoie: Still second guessing, or should we merge?

francislavoie:
This is missing Caddyfile support for the `grace_period` option

Issue #5057, opened 2022-09-20
Define a virtual FS to use for all subsequent routes 


francislavoie:
Followup on https://github.com/caddyserver/caddy/pull/4909

Consider a config like this:

```
root * /srv
try_files {path} /index.html
file_server
```

If you switch to using a virtual filesystem, of course you'd need `try_files` and `file_server` to use the same filesystem, otherwise it wouldn't really make sense.

Defining the filesystem to use for each directive that might need it doesn't seem very nice, so we need to come up with a solution for defining it once early in the pipeline, and have any of these routes pick it up from the request context (similarly to how `root` sets a variable to be used later).

We could introduce an `fs` directive for this, which takes a virtual FS module and then puts a reference to it in context, which `file_server` and the `file` matcher can use by default if not otherwise explicitly configured (and if neither, use the local filesystem of course)

mholt:
I support this idea.

sod-lol:
Is this issue still open? i want to try working on this one. thx

mholt:
@sod-lol I think so! Would be happy to review a PR or proposal :)

elee1766:
@mholt hi, i'm trying to implement the `fs` directive, but need some help/input on my solution

context - i have a plugin that can use multiple archives/directories from remote/local sources, and use an overlay of such archives to serve a site. it doesn't work though because i want to serve an SPA, try_files won't use the right filesystem

after some reading, my proposal for implementation would be as follows:

provide a global "filesystems" map, similar to "vars" but scoped like "storage". this "filesystems" map exists in "context". instead of using m.fileSystem = osFS{}, we should do m.fileSystem = ctx.FileSystems().Default()  

add a global "filesystem" directive, the syntax would be `filesystem <key> <module...>`. possibly can use no key if the first word is a valid module to mean "default"? (is this possible with your config framework?) 

"filesystem" directive can be used more than once to declare more than one filesystem. this directive would add filesystems to the filesystems map. overwriting the default key can osFS as the default fs.

add a route directive `fs`, the syntax would be `filesystem <key>`. this would add a key like `http.vars.fs` to the vars, like "root". 

also replace the file_server fs option with a string key, which would correspond to a filesystem declared in the global filesystems map.

when a MatchFile is provisioned, we can extract the "http.vars.fs" and then read the filesystem out of the global filesystem map as well.  

the idea is to reduce the amount of loading of filesystem modules. the loading is done once at the start, and reused by components.  

[...]

PR #5833, opened 2023-09-21
filesystem: Globally declared filesystems, `fs` directive

elee1766:
attempting to address #5057

the current implementation creates a new app with module name "caddy.filesystems". this module is populated with global options

this app is configured by a new `filesystem` options directive. use multiple filesystem times to define multiple filesystems to add to the map 

new route directive `fs` to declare a fs for the http route context, similar to `root`. this is shared between try_routes, file_server, and anything else that uses MatchRoutes module

example caddyfile with an fs impl named `swim`, which uses a remotely downloaded archive as fs, and routes like an SPA

```
{
  filesystem app swim "https://remote.com/artifact" /dist .zip
}
fs app
try_files {path} /
file_server
```

i tried to add as little new things as possible, reusing the existing app framework and options framework. im not sure if they are the correct tools for the job, but they seem okay.

design choices that i dont really know how you guys feel about:

1. the filesystem map is pretty "magical", and so i decided to hide it in `internal`. I think it's better if people aren't allowed to touch it.

2. the filesystem map is thread safe and modifyable at runtime. I dont know if this is a part of the design contract that you guys want to enforce, maybe its better to keep it not thread safe.

3. I couldn't find an easier way besides adding something to the "Context" which allowed me to share runtime objects across apps. is there some framework i am missing? I'm not really sure how this will scale

### DIFF
--- a/modules/caddyfs/filesystem.go
+++ b/modules/caddyfs/filesystem.go
@@ -71,6 +71,7 @@ func (xs *Filesystems) Provision(ctx caddy.Context) error {
 		ctx.Filesystems().Register(f.Key, f.fileSystem)
 		// remember to unregister the module when we are done
 		xs.defers = append(xs.defers, func() {
+			ctx.Logger().Debug("registering fs", zap.String("fs", f.Key))
 			ctx.Filesystems().Unregister(f.Key)
 		})
 	}
--- a/modules/caddyfs/filesystem.go
+++ b/modules/caddyfs/filesystem.go
@@ -76,7 +76,6 @@ func (xs *Filesystems) Provision(ctx caddy.Context) error {
 		})
 	}
 	return nil
-
 }
 
 func (f *Filesystems) Cleanup() error {
--- a/modules/caddyfs/filesystem.go
+++ b/modules/caddyfs/filesystem.go
@@ -1,0 +1,111 @@
+package caddyfs
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/fs"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"github.com/caddyserver/caddy/v2/caddyconfig/httpcaddyfile"
+	"go.uber.org/zap"
+)
+
+func init() {
+	caddy.RegisterModule(Filesystems{})
+	httpcaddyfile.RegisterGlobalOption("filesystem", parseFilesystems)
+}
+
+type moduleEntry struct {
+	Key           string          `json:"name,omitempty"`
+	FileSystemRaw json.RawMessage `json:"file_system,omitempty" caddy:"namespace=caddy.fs inline_key=backend"`
+	fileSystem    fs.FS
+}
+
+// Filesystems loads caddy.fs modules into the global filesystem map
+type Filesystems struct {
+	Filesystems []*moduleEntry `json:"filesystems"`
+
+	defers []func()
+}
+
+func parseFilesystems(d *caddyfile.Dispenser, existingVal any) (any, error) {
+	p := &Filesystems{}
+	current, ok := existingVal.(*Filesystems)
+	if ok {
+		p = current
+	}
+	x := &moduleEntry{}
+	err := x.UnmarshalCaddyfile(d)
+	if err != nil {
+		return nil, err
+	}
+	p.Filesystems = append(p.Filesystems, x)
+	return p, nil
+}
+
+// CaddyModule returns the Caddy module information.
+func (Filesystems) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "caddy.filesystems",
+		New: func() caddy.Module { return new(Filesystems) },
+	}
+}
+
+func (xs *Filesystems) Start() error { return nil }
+func (xs *Filesystems) Stop() error  { return nil }
+
+func (xs *Filesystems) Provision(ctx caddy.Context) error {
+	// load the filesystem module
+	for _, f := range xs.Filesystems {
+		if len(f.FileSystemRaw) > 0 {
+			mod, err := ctx.LoadModule(f, "FileSystemRaw")
+			if err != nil {
+				return fmt.Errorf("loading file system module: %v", err)
+			}
+			f.fileSystem = mod.(fs.FS)
+		}
+		// register that module
+		ctx.Logger().Debug("registering fs", zap.String("fs", f.Key))
+		ctx.Filesystems().Register(f.Key, f.fileSystem)
+		// remember to unregister the module when we are done
+		xs.defers = append(xs.defers, func() {
+			ctx.Filesystems().Unregister(f.Key)
+		})
+	}
+	return nil
+
+}
+
+func (f *Filesystems) Cleanup() error {
+	for _, v := range f.defers {
+		v()
+	}
+	return nil
+}
+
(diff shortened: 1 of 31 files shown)


elee1766:
@mholt @francislavoie please let me know how this looks / if i have sinned anywhere.

mholt:
So sorry for the wait, it's been a busy quarter! I will return to this ASAP. Just letting you know I haven't forgotten about it.

elee1766:
> So sorry for the wait, it's been a busy quarter! I will return to this ASAP. Just letting you know I haven't forgotten about it.

no problem, i've been busy too.

i've actually been thinking about this a bit in my spare time - i added the fs to the global app because that's how certificates were done and certificates were in my mind a similar problem (sharing assets across many sub apps). however, thinking more, i think it might make more sense to perhaps store them as any other module? (and maybe perhaps make certs function in that way as well?)

on a completely unrelated note - working with caddy made me really enjoy the format. we've done something similar to coredns, except for postgres pooling/lb/routing instead of dns. it's here https://gfx.cafe/gfx/pggat

modules/caddyfs/filesystem.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+			f.fileSystem = mod.(fs.FS)
+		}
+		// register that module
+		ctx.Logger().Debug("registering fs", zap.String("fs", f.Key))
+		ctx.Filesystems().Register(f.Key, f.fileSystem)
+		// remember to unregister the module when we are done
+		xs.defers = append(xs.defers, func() {
+			ctx.Filesystems().Unregister(f.Key)
+		})
```
francislavoie, line 75: How does this behave with config reloads? Is ctx scoped to only the current config (and not global)? Just making sure that it won't unregister an fs that didn't change keys between config reloads.

modules/caddyhttp/fileserver/caddyfile.go
```diff
@@ -37,7 +35,7 @@ func init() {
 // server and configures it with this syntax:
 //
 //	file_server [<matcher>] [browse] {
-//	    fs            <backend...>
+//	    fs            <filesystem>
```
francislavoie, line 38: I think technically this is a breaking change, right? Because we now need to define an fs name instead of an fs module? I think this is ok because it's an improvement, we just need to document it in the release notes.

francislavoie (commented):
This looks awesome ☺️

modules/caddyhttp/fileserver/caddyfile.go
elee1766: yes it is. I think this makes a lot more sense.

I tried playing around to declare filesystems inline but it was a nightmare so I gave up...

modules/caddyfs/filesystem.go
elee1766: so I assumed caddy would run cleanup before the config reload and provision after, is that not what happens for config reloads?

francislavoie: See https://caddyserver.com/docs/architecture for an explanation of the lifecycle.

Basically, when reloading there's two configs loaded and started for a moment. The new config is loaded and started before the old config is stopped. Cleanup gets called during stoppage.

I think what you did might be fine, I just wanted to make sure you tested config reloads and confirmed things still work normally after a reload.

francislavoie:
This'll need a rebase at this point.

For some context, we just merged https://github.com/caddyserver/xcaddy/pull/160 this week which will allow `xcaddy` to wire up a filesystem with files embedded in the binary a build-time. We were talking about what else Caddy would need to make config work well, but I totally forgot we had this PR pending :man_facepalming:

Looking to get this merged soon! It'll pair very well with embedding with `xcaddy`.

modules/caddyfs/filesystem.go
elee1766: i haven't so i will, and i will test a few cases as well. 

is there a way to programmatically test this with some sort of harness as well?

elee1766:
> This'll need a rebase at this point.
> 
> For some context, we just merged [caddyserver/xcaddy#160](https://github.com/caddyserver/xcaddy/pull/160) this week which will allow `xcaddy` to wire up a filesystem with files embedded in the binary a build-time. We were talking about what else Caddy would need to make config work well, but I totally forgot we had this PR pending 🤦‍♂️
> 
> Looking to get this merged soon! It'll pair very well with embedding with `xcaddy`.

ah i see. 

would you like me to convert the embed filesystem to this format as well ?

francislavoie:
I don't think we need to do anything, embed is a `caddy.fs` module so it'll still be compatible

elee1766:
@francislavoie i did some testing and it seems to function correctly on config reload, when keys are added, removed, or unchanged.

elee1766:
ran fumpt and updated master

side question @francislavoie would you guys accept the local/remote archive module in some form into xcaddy ?

mholt:
Thanks for maintaining this -- I'm working through a holiday backlog and will get around to this soon :)

elee1766:
https://github.com/gfx-labs/swim/tree/master 

example caddyfile with a functional dockerfile is here

francislavoie:
```
Error: modules/caddyhttp/fileserver/matcher_test.go:119:4: unknown field fileSystem in struct literal of type MatchFile
Error: modules/caddyhttp/fileserver/matcher_test.go:119:16: undefined: osFS
Error: modules/caddyhttp/fileserver/matcher_test.go:228:4: unknown field fileSystem in struct literal of type MatchFile
Error: modules/caddyhttp/fileserver/matcher_test.go:228:16: undefined: osFS
```

elee1766:
> ```
> Error: modules/caddyhttp/fileserver/matcher_test.go:119:4: unknown field fileSystem in struct literal of type MatchFile
> Error: modules/caddyhttp/fileserver/matcher_test.go:119:16: undefined: osFS
> Error: modules/caddyhttp/fileserver/matcher_test.go:228:4: unknown field fileSystem in struct literal of type MatchFile
> Error: modules/caddyhttp/fileserver/matcher_test.go:228:16: undefined: osFS
> ```

fixed, but i had to add a nil check for filesystems here so that the CEL matcher would pass
```
// Filesystems returns a ref to the FilesystemMap
func (ctx *Context) Filesystems() FileSystems {
	// if no config is loaded, we use a default filesystemmap, which includes the osfs
	if ctx.cfg == nil {
		return &filesystems.FilesystemMap{}
	}
	return ctx.cfg.filesystems
}
```
i feel this is bad though? is this safe for runtime?

i saw in AppIfConfigured the following comment, but i dont really understand what can cause this to happen, or if it is okay.
```
if ctx.cfg == nil {
	// this can happen if the currently-active context
	// is being accessed, but no config has successfully
	// been loaded yet
        return nil
}
```

modules/caddyhttp/fileserver/matcher_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 Matthew Holt and The Caddy Authors
+// Copyright 2016 Matthew Holt and The Caddy Authors
```
mohammed90, line 1: Any reason for this?

elee1766: i have no clue how this happened lol. maybe i accidentally pressed the increment hotkey in vim... ill change it back so there's no diff

francislavoie:
I think the `ctx.cfg == nil` happens in the tests because we're trying to unit test the matcher outside of a real Caddy config context. That fix is probably fine, it should only happen in tests.

Also, looks like `gci` needs to be run again in `context.go`

elee1766:
sounds good, and done.

francislavoie (approved):
This looks really good to me! Excited to get it merged.

PR #5844, opened 2023-09-30
httpcaddyfile: Rewrite `root` and `rewrite` parsing to allow omitting matcher

francislavoie:
I had this idea from reviewing @dunglas's code recently (where he copy-pasted some code from the Caddy codebase itself). The `php_fastcgi` directive uses `RegisterDirective` to get full control of parsing, instead of the `RegisterHandlerDirective` shortcut which does automatic request matcher parsing.

That made me realize we can do the same for `root`, and that we could actually make `root /srv` work, instead of the awkward `root * /srv` with an explicit `*` matcher) that we've been stuck with since v2.0.

Nice and simple, just a small rewrite of the parsing to count the number of arguments first, and if we have only one argument then we can assume it's the actual root path and that there's no matcher, because it never makes sense to ever only pass a matcher to `root` and no root path.

Added a test to cover the happy path, and tested the 0 and 3+ argument cases by hand quickly.

Before:
```
root * /srv
```
After:
```
root /srv
```
:tada:

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -41,7 +41,7 @@ func init() {
 	RegisterDirective("bind", parseBind)
 	RegisterDirective("tls", parseTLS)
 	RegisterHandlerDirective("fs", parseFilesystem)
-	RegisterHandlerDirective("root", parseRoot)
+	RegisterDirective("root", parseRoot)
 	RegisterHandlerDirective("vars", parseVars)
 	RegisterHandlerDirective("redir", parseRedir)
 	RegisterHandlerDirective("respond", parseRespond)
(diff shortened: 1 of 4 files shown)


francislavoie:
Realized we can do the same with the `rewrite` directive, it had the same problem as `root`.

francislavoie:
This is good to go, just need an approval :smile:

mholt (approved):
Cool, thanks. I initially didn't do this because it's hard to tell the difference between `root <matcher>` (missing actual root path) and `root <path>` (implied wildcard matcher).

I think experience has shown that wildcard matchers are by far the most common, which maybe I should have predicted, but anyway; I think this is reasonable to assume that a single arg is the root path.

PR #5845, opened 2023-10-01
templates: Add dummy `RemoteAddr` to `httpInclude` request, proxy compatibility

### LINKED ISSUE #5835: httpInclude through proxy with X-Forwarded-For
As discussed in the forum, I am creating this issue to track the progress of the following discussion: https://caddy.community/t/file-server-httpinclude-with-x-forwarded-for/21087

TL;DR:

Goal: Allow the virtual request of `httpInclude` of a `file_server` to include a `RemoteAddr` so that a reverse proxy can set `X-Forwarded-For`.

Otherwise, the `X-Forwarded-For` header is not set and not included in the endpoint that processes the original `httpInclude` request, which is required for applications that trust this header and return different text data depending on its value.

ThanmayNath:
Issue:
This pull request addresses fixes #5835, which was created to track the progress of the discussion on the Caddy community forum: https://caddy.community/t/file-server-httpinclude-with-x-forwarded-for/21087. The issue discusses the need to allow the virtual request of `httpInclude` of a file_server to include the `X-Forwarded-For` header. Currently, the `X-Forwarded-For` header is not set in the virtual request, which poses a challenge for applications that rely on this header to determine the client's IP address and make decisions based on it.

Goal:
The goal of this enhancement is to modify the `funcHTTPInclude` function in the Caddy codebase to include the `X-Forwarded-For` header in the virtual request. This change will enable reverse proxies to set the `X-Forwarded-For` header, ensuring that the client's IP address is correctly provided to the target endpoint. This modification is essential for applications that depend on the `X-Forwarded-For` header for various functionalities, such as authentication, logging, or content customization.

Proposed Changes:

- In the funcHTTPInclude function, we will create a new virtual request (virtReq).
- We will set the Host header of the virtual request to match the original request's Host.
- The X-Forwarded-For header will be added to the virtual request, including the client's IP address obtained from the c.RemoteIP() function.
- Other headers, such as Accept-Encoding, will be set as needed.
- The virtual request will be performed as before, ensuring that the target endpoint receives the X-Forwarded-For header.

### DIFF
--- a/modules/caddyhttp/templates/tplcontext.go
+++ b/modules/caddyhttp/templates/tplcontext.go
@@ -192,12 +192,10 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
 	// Set the Host header to the original request's Host
 	virtReq.Host = c.Req.Host
 
+	virtReq.RemoteAddr = "127.0.0.1"
+
 	// Clone the original request's headers and add X-Forwarded-For
 	virtReq.Header = c.Req.Header.Clone()
-	clientIP := c.RemoteIP() // Get the client's IP address
-	virtReq.Header.Add("X-Forwarded-For", clientIP)
-
-	// Set other headers as needed
 	virtReq.Header.Set("Accept-Encoding", "identity")
 	virtReq.Trailer = c.Req.Trailer.Clone()
 	virtReq.Header.Set(recursionPreventionHeader, strconv.Itoa(recursionCount))
--- a/modules/caddyhttp/templates/tplcontext.go
+++ b/modules/caddyhttp/templates/tplcontext.go
@@ -188,7 +188,7 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
 		return "", err
 	}
 	virtReq.Host = c.Req.Host
-	virtReq.RemoteAddr = "127.0.0.1"
+	virtReq.RemoteAddr = "127.0.0.1:10000" // https://github.com/caddyserver/caddy/issues/5835
 	virtReq.Header = c.Req.Header.Clone()
 	virtReq.Header.Set("Accept-Encoding", "identity") // https://github.com/caddyserver/caddy/issues/4352
 	virtReq.Trailer = c.Req.Trailer.Clone()
--- a/modules/caddyhttp/templates/tplcontext.go
+++ b/modules/caddyhttp/templates/tplcontext.go
@@ -183,28 +183,21 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
 	buf.Reset()
 	defer bufPool.Put(buf)
 
-	// Create a new virtual request
 	virtReq, err := http.NewRequest("GET", uri, nil)
 	if err != nil {
 		return "", err
 	}
-
-	// Set the Host header to the original request's Host
 	virtReq.Host = c.Req.Host
-
 	virtReq.RemoteAddr = "127.0.0.1"
-
-	// Clone the original request's headers and add X-Forwarded-For
 	virtReq.Header = c.Req.Header.Clone()
-	virtReq.Header.Set("Accept-Encoding", "identity")
+	virtReq.Header.Set("Accept-Encoding", "identity") // https://github.com/caddyserver/caddy/issues/4352
 	virtReq.Trailer = c.Req.Trailer.Clone()
 	virtReq.Header.Set(recursionPreventionHeader, strconv.Itoa(recursionCount))
 
-	// Perform the virtual request
 	vrw := &virtualResponseWriter{body: buf, header: make(http.Header)}
 	server := c.Req.Context().Value(caddyhttp.ServerCtxKey).(http.Handler)
-	server.ServeHTTP(vrw, virtReq)
 
+	server.ServeHTTP(vrw, virtReq)
 	if vrw.status >= 400 {
 		return "", fmt.Errorf("http %d", vrw.status)
 	}
--- a/modules/caddyhttp/templates/tplcontext.go
+++ b/modules/caddyhttp/templates/tplcontext.go
@@ -183,20 +183,30 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
 	buf.Reset()
 	defer bufPool.Put(buf)
 
+	// Create a new virtual request
 	virtReq, err := http.NewRequest("GET", uri, nil)
 	if err != nil {
 		return "", err
 	}
+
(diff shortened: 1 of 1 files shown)


francislavoie:
This isn't the right change. We don't want to touch `X-Forwarded-For` at all here. All we need to do is set `virtReq.RemoteAddr = "127.0.0.1"` or something to that effect.

ThanmayNath:
@francislavoie Thank you for your prompt feedback and guidance. I've made the suggested changes to the code as you recommended. I added the line to set `virtReq.RemoteAddr = "127.0.0.1"` as you mentioned. This should ensure that the RemoteAddr is correctly configured for virtual requests.
 
 Once again, thank you for your assistance in reviewing this code.

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -183,20 +183,28 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
 	buf.Reset()
 	defer bufPool.Put(buf)
 
+	// Create a new virtual request
```
francislavoie, line 186: This comment is redundant, it's already explained elsewhere and it restates the variable name
```suggestion
```

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -183,20 +183,28 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
...
 	virtReq, err := http.NewRequest("GET", uri, nil)
 	if err != nil {
 		return "", err
 	}
+
+	// Set the Host header to the original request's Host
```
francislavoie, line 192: This is also not useful, it's very obvious what the line does
```suggestion
```

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -183,20 +183,28 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
...
 	virtReq.Host = c.Req.Host
+
+	virtReq.RemoteAddr = "127.0.0.1"
+
+	// Clone the original request's headers and add X-Forwarded-For
```
francislavoie, line 197: The comment is no longer true, and the newlines are unnecessary
```suggestion
	virtReq.RemoteAddr = "127.0.0.1"
```

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -183,20 +183,28 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
...
 	virtReq.Header = c.Req.Header.Clone()
-	virtReq.Header.Set("Accept-Encoding", "identity") // https://github.com/caddyserver/caddy/issues/4352
+	virtReq.Header.Set("Accept-Encoding", "identity")
```
francislavoie, line 199: Keep the comment as-is, the link to the issue is important to explain the context for this line being in place
```suggestion
	virtReq.Header.Set("Accept-Encoding", "identity") // https://github.com/caddyserver/caddy/issues/4352
```

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -183,20 +183,28 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
...
 	virtReq.Trailer = c.Req.Trailer.Clone()
 	virtReq.Header.Set(recursionPreventionHeader, strconv.Itoa(recursionCount))
 
+	// Perform the virtual request
```
francislavoie, line 203: This is also not useful, and not quite correct. This part is still setting up the request body, not performing the request
```suggestion
```

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -183,20 +183,28 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
...
 	vrw := &virtualResponseWriter{body: buf, header: make(http.Header)}
 	server := c.Req.Context().Value(caddyhttp.ServerCtxKey).(http.Handler)
-
 	server.ServeHTTP(vrw, virtReq)
+
```
francislavoie, line 207: This line is actually performing the request and is no longer setup, so a newline before is what we want, to visually split it from the setup.
```suggestion

	server.ServeHTTP(vrw, virtReq)
```

francislavoie (changes requested):

modules/caddyhttp/templates/tplcontext.go
```diff
@@ -188,6 +188,7 @@ func (c TemplateContext) funcHTTPInclude(uri string) (string, error) {
 		return "", err
 	}
 	virtReq.Host = c.Req.Host
+	virtReq.RemoteAddr = "127.0.0.1"
```
francislavoie, line 191: Actually, I think it needs to be an `IP:port`, so I suppose we can just set it to some high port and call it a day
```suggestion
	virtReq.RemoteAddr = "127.0.0.1:10000" // https://github.com/caddyserver/caddy/issues/5835
```

ThanmayNath: @francislavoie Could you please explain why we're using a high port (10000) in this context to set `RemoteAddr` to `127.0.0.1:10000`? What specific benefits or considerations does this change bring to our application? Additionally, are there any potential drawbacks or security concerns associated with this approach that I should be aware of?

francislavoie: The point of setting `RemoteAddr` here is just so that it's non-empty so that the logic in https://github.com/caddyserver/caddy/blob/4feac4d83cd758c95194090d4f3468373ee342ef/modules/caddyhttp/reverseproxy/reverseproxy.go#L667 does not quit early from having an invalid remote address, and therefore the `X-Forwarded-*` headers will be correctly added to the request sent upstream. The port does not matter at all (because the port on the remote address is always a random port assigned by the kernel networking stack), we never read the port or use it in any way, it just needs to exist for parsing to work properly with `net.SplitHostPort()`.

ThanmayNath:
Hi @francislavoie,

I've looked into the requested changes, but it appears that I'm unable to view the specific modifications in the code.
The message I'm getting is: "We went looking everywhere, but couldn’t find those commits. Sometimes commits can disappear after a force-push."

I want to ensure that I address your feedback correctly, so I would appreciate it if you could provide some additional details or clarification regarding the changes you'd like to see. This will help me make the necessary adjustments promptly.

Thank you for your patience and assistance in resolving this.

francislavoie:
There's nothing left to do. Just getting confirmation that it fixes the issue from the person who reported it.

francislavoie (approved):

PR #5848, opened 2023-10-02
Use sync.Pool to reduce lengthReader allocations

### LINKED ISSUE #5756: Use `sync.Pool` to reduce `lengthReader` allocations
In https://github.com/caddyserver/caddy/pull/5461 we introduced `lengthReader` to count the amount of bytes read from the request body.

We're allocating a new `lengthReader` on each request though (but only requests with a body, so not GETs). This is inefficient, it could be improved by using `sync.Pool` to reuse allocated `lengthReader` structs.

perhapsmaple:
Should fix #5756.

Tested allocations with this benchmark:

```
func BenchmarkServer_LengthReader(b *testing.B) {
	req := httptest.NewRequest(http.MethodGet, "/", nil)

	for i := 0; i < b.N; i++ {
		var bodyReader *lengthReader
		for j := 0; j < 10000; j++ {
			if req.Body != nil {
				// For testing without sync.Pool
				// bodyReader = &lengthReader{Source: req.Body}
				bodyReader = GetLengthReader(req.Body)
				req.Body = bodyReader
			}
			PutLengthReader(bodyReader)
		}
	}
	b.ReportAllocs()
}
```
Without sync.Pool:
BenchmarkServer_LengthReader-12    	    1808	    700019 ns/op	  240003 B/op	   10000 allocs/op

With sync.Pool
BenchmarkServer_LengthReader-12    	    9516	    127160 ns/op	       0 B/op	       0 allocs/op

This is my first time writing Go code, tried my best to stick to the same style. Successfully passed unit tests locally.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -319,7 +319,7 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// so we can track the number of bytes read from it
 		var bodyReader *lengthReader
 		if r.Body != nil {
-			bodyReader = &lengthReader{Source: r.Body}
+			bodyReader = getLengthReader(r.Body)
 			r.Body = bodyReader
 		}
 
@@ -728,6 +728,7 @@ func (s *Server) logRequest(
 	reqBodyLength := 0
 	if bodyReader != nil {
 		reqBodyLength = bodyReader.Length
+		putLengthReader(bodyReader)
 	}
 
 	extra := r.Context().Value(ExtraLogFieldsCtxKey).(*ExtraLogFields)
@@ -909,6 +910,24 @@ func cloneURL(from, to *url.URL) {
 type lengthReader struct {
 	Source io.ReadCloser
 	Length int
+}
+
+var lengthReaderPool = sync.Pool{
+	New: func() interface{} {
+		return &lengthReader{}
+	},
+}
+
+func getLengthReader(source io.ReadCloser) *lengthReader {
+	reader := lengthReaderPool.Get().(*lengthReader)
+	reader.Source = source
+	reader.Length = 0
+	return reader
+}
+
+func putLengthReader(reader *lengthReader) {
+	reader.Source = nil
+	lengthReaderPool.Put(reader)
 }
 
 func (r *lengthReader) Read(b []byte) (int, error) {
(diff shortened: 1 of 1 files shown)


francislavoie:
I think since `logRequest` may return early due to `SkipLog`, this may leak as-is. I think it needs to `put` in that case as well.

modules/caddyhttp/server.go
```diff
@@ -319,7 +319,7 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// so we can track the number of bytes read from it
 		var bodyReader *lengthReader
 		if r.Body != nil {
-			bodyReader = &lengthReader{Source: r.Body}
+			bodyReader = getLengthReader(r.Body)
```
francislavoie, line 322: I think it should be safe to just do it like this:

```suggestion
			bodyReader = getLengthReader(r.Body)
			defer putLengthReader(bodyReader)
```

Because `defer` runs in reverse order on the way out, so `defer s.logRequest` will run before `defer putLengthReader`.

perhapsmaple:
Hi @francislavoie, I have added the requested changes. Let me know if any other changes are required

modules/caddyhttp/server.go
```diff
@@ -911,6 +912,24 @@ type lengthReader struct {
...
+	reader := lengthReaderPool.Get().(*lengthReader)
+	reader.Source = source
+	reader.Length = 0
+	return reader
+}
+
+func putLengthReader(reader *lengthReader) {
+	reader.Source = nil
+	lengthReaderPool.Put(reader)
+}
```
francislavoie, line 931: I think the length should be reset when put back in the pool rather than when getting it. Numerical types already default to zero in Go. This makes it more symmetrical, i.e. `put` doing all the cleanup.

```suggestion
func getLengthReader(source io.ReadCloser) *lengthReader {
	reader := lengthReaderPool.Get().(*lengthReader)
	reader.Source = source
	return reader
}

func putLengthReader(reader *lengthReader) {
	reader.Source = nil
	reader.Length = 0
	lengthReaderPool.Put(reader)
}
```

mholt:
Thanks for this optimization! It's relatively simple and I believe it is correct.

Good job on one of your first Go contributions :)

mholt (approved):
Thanks for this optimization! It's relatively simple and I believe it is correct.

Good job on one of your first Go contributions :) Thanks for the benchmark too!

PR #5850, opened 2023-10-02
fileserver: Implement caddyfile.Unmarshaler interface

francislavoie:
As pointed out by @dunglas, we were missing this to make it easier to embed `file_server` within other shortcut directives.

### DIFF
--- a/modules/caddyhttp/fileserver/caddyfile.go
+++ b/modules/caddyhttp/fileserver/caddyfile.go
@@ -39,30 +39,10 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 	if err != nil {
 		return fsrv, err
 	}
-
-	// Hide the Caddyfile (and any imported Caddyfiles).
-	// This needs to be done in here instead of UnmarshalCaddyfile
-	// because UnmarshalCaddyfile only has access to the dispenser
-	// and not the helper, and only the helper has access to the
-	// Caddyfiles function. This does mean that Caddyfiles will
-	// only be hidden when the directive is used directly, and not
-	// if used in some embedded fashion (e.g. via another directive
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/fileserver/caddyfile.go
```diff
@@ -33,8 +33,43 @@ func init() {
...
+	}
+
+	// Hide the Caddyfile (and any imported Caddyfiles).
+	// This needs to be done in here instead of UnmarshalCaddyfile
+	// because UnmarshalCaddyfile only has access to the dispenser
+	// and not the helper, and only the helper has access to the
+	// Caddyfiles function. This does mean that Caddyfiles will
+	// only be hidden when the directive is used directly, and not
+	// if used in some embedded fashion (e.g. via another directive
+	// as a shortcut).
```
mholt, line 52: This could be a big gotcha...

I'm not sure what to do about this, currently. Hmm.

francislavoie: We could move that logic to a method of the `FileServer` module, taking `h` as input, then embedders may call that method to add the `hide` stuff on top if necessary. We do the same kind of thing with `reverse_proxy` i.e. `rp.FinalizeUnmarshalCaddyfile(h)`, we could make a similar `Finalize` for `file_server`.

francislavoie: Did that :+1:

mholt: Oh -- good idea 🤔 Why didn't I think of that...

mholt (approved):
This LGTM. Still could be a footgun (like in reverse proxy) if programmers don't read the docs and forget to call Finalize... but overall I think this is reasonable. Thanks!

PR #5851, opened 2023-10-02
reverseproxy: Add `tls_curves` option to HTTP transport

bwesterb:
Example Caddyfile

    localhost {
        reverse_proxy https://example.com {
            transport http {
                tls_curves secp521r1
            }
        }
    }

Long term goal is to be able to configure post-quantum for Caddy -> upstream.

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -491,6 +491,10 @@ type TLSConfig struct {
 	// When specified, TLS will automatically be configured on the transport.
 	// The value can be a list of any valid tcp port numbers, default empty.
 	ExceptPorts []string `json:"except_ports,omitempty"`
+
+	// The list of elliptic curves to support. Caddy's
+	// defaults are modern and secure.
+	Curves []string `json:"curves,omitempty"`
 }
 
 // MakeTLSClientConfig returns a tls.Config usable by a client to a backend.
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -493,6 +493,10 @@ type TLSConfig struct {
 	// When specified, TLS will automatically be configured on the transport.
 	// The value can be a list of any valid tcp port numbers, default empty.
 	ExceptPorts []string `json:"except_ports,omitempty"`
+
+	// The list of elliptic curves to support. Caddy's
+	// defaults are modern and secure.
```
bwesterb, line 498: Actually, it reverts to Go's default, which are ~.

mholt: Interesting, Go's defaults are "mid" (is that what ~ means)? What would you suggest instead?

bwesterb: With ~ I meant to refer to the quote: modern and secure. 😆

mholt (approved):
Thank you for this. I noticed it doesn't reject unknown curve names, but maybe that's OK for a niche feature for now. I don't think it will break anything at least.

Since this is a new feature, it will have to wait until 2.9 but we can at least merge it in whenever you're ready.

Did you want to add an error if there's an unrecognized input?

PR #5853, opened 2023-10-02
fix: caddytest.AssertResponseCode error message

dunglas:
This error message is always empty because this field isn't set on a client request. This patch fixes the issue.


mholt (approved):
Oh, good catch. Thank you!

PR #5855, opened 2023-10-04
fileserver: Enable compression for command by default

dunglas:
Compress responses with Zstandard or Gzip when using `caddy file-server`. Also adds a flag to disable compression.

Backport of https://github.com/dunglas/frankenphp/pull/238 as discussed in https://github.com/dunglas/frankenphp/pull/238#issuecomment-1747271520.


modules/caddyhttp/fileserver/command.go
```diff
@@ -84,9 +86,34 @@ func cmdFileServer(fs caddycmd.Flags) (int, error) {
...
+		}
+		encodings["gzip"] = caddyconfig.JSON(gzip.New(), nil)
+		prefer = append(prefer, "gzip")
+
+		zstd, err := caddy.GetModule("http.encoders.zstd")
+		if err != nil {
+			return 0, err
+		}
+		encodings["zstd"] = caddyconfig.JSON(zstd.New(), nil)
+		prefer = append(prefer, "zstd")
```
francislavoie, line 109: I think we usually put zstd before gzip because it _should_ perform better, but most clients don't support it (yet) so gzip is the fallback.

dunglas: Oops, it was my intent but I messed up the order in a refactoring. Fixed (and simplified) now.

francislavoie (approved):

mholt (approved):

modules/caddyhttp/fileserver/command.go
```diff
@@ -62,6 +65,7 @@ respond with a file listing.`,
 			cmd.Flags().BoolP("templates", "t", false, "Enable template rendering")
 			cmd.Flags().BoolP("access-log", "a", false, "Enable the access log")
 			cmd.Flags().BoolP("debug", "v", false, "Enable verbose debug logs")
+      cmd.Flags().BoolP("no-compress", "", false, "Disable Zstandard and Gzip compression")
```
mholt, line 68:
```suggestion
			cmd.Flags().BoolP("no-compress", "", false, "Disable Zstandard and Gzip compression")
```

mholt:
Thanks Kevin!!

PR #5857, opened 2023-10-05
reverseproxy: Add debug logging for dynamic A upstreams

francislavoie:
Just an oversight, A is missing the same debug logging that SRV had


mholt (approved):
Thanks!

Issue #5704, opened 2023-08-06
Caddy was on a feature freeze before the 2.8 release


mholt:
It's been figuratively sticky-noted right in front of my eyes ever since releasing Caddy 2: we need more tests.

Now that 2.7 is done, I want to be confident about our code base _before_ we release updates.

---

**UPDATE, NOV. 2023:** See my [comment below](https://github.com/caddyserver/caddy/issues/5704#issuecomment-1826152531); life had other plans, so I'm thawing the feature freeze for practical reasons. We will revisit testing next year.

---

Together with finally focusing on the new Caddy website that I've already been working on only occasionally [since the beginning of this year](https://caddy.community/t/refreshing-the-website-and-docs-in-2023/18314), I will be dedicating my remaining development energy to overhauling and improving Caddy's testing situation.

To be clear, we have a variety of tests today. Many functional units are tested, some quite thoroughly. (See any `_test.go` file in our repo.) And we have a [whole directory tree](https://github.com/caddyserver/caddy/tree/master/caddytest) dedicated to testing Caddy, especially [a large pile of Caddyfile adapter tests](https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt). And our tests have saved us many times and prevented many problems.

But we know now that this is not enough. The 2.7 release is evidence of that. It took us 4 releases (2.7.0, then 3 hotfixes) to get it right. Web servers are hard, they're complicated, there's a bajillion UI and API surfaces (which, by the way, is the reason using semver for Caddy 2 causes me great anguish -- but I digress), and we have several years of production experience with Caddy 2 to know the nuances and needs of production deployments. It's getting easier to break things.

**I'm putting a feature freeze on Caddy until after the 2.8 release so we can focus on testing.** Caddy 2.8 will be all about tests: testing harnesses, test APIs for module developers, unit tests, integration tests, configuration tests, etc...

During this feature freeze, we'll refrain from making unnecessary changes to the code base. "Necessary" changes are defined as simple, unintrusive bug fixes only. The main exception being if a bug is severe (crash, memory leak, security vulnerability), we do what we have to do to patch it without delay.

Test code is obviously exempt from this freeze. The whole point is to build out testing features, write more tests, etc.

A few other things that are probably OK during the freeze:

- Typo corrections
- Anything pertaining only to comments
- Localized code cleanup (for example, code marked with a TODO that was needed only for an older version of Go that we no longer use can be removed)
- Other very minor, sensible improvements

[...]

PR #5859, opened 2023-10-06
caddyhttp: Add `uuid` to access logs when used

### LINKED ISSUE #5820: Include `http.request.uuid` field in access logs?
I have a reverse proxy setup and I'm trying to 'link' access logs to upstream server events. What I've got so far is the following Caddy file:

```nginx
localhost {
	reverse_proxy {
		to https://localhost:5000
		header_up X-Request-Id "{http.request.uuid}"
		transport http {
			tls
			tls_insecure_skip_verify
			read_buffer 8192
		}
	}
	log {
		level info
		# somehow include an additional field {..., requestId: "{http.request.uuid}"} in the log?
	}
}
```

The HTTP server at `https://localhost:5000` can read the incoming `X-Request-Id` header just fine. But how can I  write the same request ID to the access log to link them together (for deeper analysis in case of upstream server errors)?

Perhaps someone knows a work-around for this issue?

francislavoie:
Closes #5820

Whenever the `{http.request.uuid}` placeholder is processed, the UUID is also added to the access logs via `ExtraLogFields`.

Added a new Caddyfile shortcut `{uuid}` because why not.

I had to make a new `Set(field)` method for `ExtraLogFields` because otherwise when using `Add(field)`, using the placeholder multiple times in a request would add the log field each time.

### DIFF
--- a/caddyconfig/httpcaddyfile/shorthands.go
+++ b/caddyconfig/httpcaddyfile/shorthands.go
@@ -64,6 +64,7 @@ func placeholderShorthands() []string {
 		"{remote_port}", "{http.request.remote.port}",
 		"{scheme}", "{http.request.scheme}",
 		"{uri}", "{http.request.uri}",
+		"{uuid}", "{http.request.uuid}",
 		"{tls_cipher}", "{http.request.tls.cipher_suite}",
 		"{tls_version}", "{http.request.tls.version}",
 		"{tls_client_fingerprint}", "{http.request.tls.client.fingerprint}",
--- a/modules/caddyhttp/logging.go
+++ b/modules/caddyhttp/logging.go
@@ -151,6 +151,18 @@ func (e *ExtraLogFields) Add(field zap.Field) {
 	e.fields = append(e.fields, field)
 }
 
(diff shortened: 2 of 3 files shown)


HeCorr:
A small downside of this change is that if the upstream server copies a UUID header into the response (so the client also receives it), the ID will appear twice in the logs (as `request.uuid` and inside `request.resp_headers`).

francislavoie:
That's fine IMO. You can remove the duplicate from your logs if you care with log filters: https://caddyserver.com/docs/caddyfile/directives/log#delete

HeCorr:
> You can remove the duplicate from your logs with log filters

Oh, I totally missed that. My bad!

sleeyax:
Any update on this?

francislavoie:
We're currently on a feature freeze https://github.com/caddyserver/caddy/issues/5704 so it will take some time until we make a decision on this. You may build from your own fork with this added if you need it right away.

mholt (approved):
Nice; this is a simple-enough change. I think it allocates when logging, but it is optional I guess. Thanks for the contribution!

PR #5865, opened 2023-10-07
httpcaddyfile: Add `RegisterDirectiveOrder` function for plugin authors

francislavoie:
We (@mholt @dunglas and I) were talking in Slack about the ergonomics of using the Caddyfile for end-users of custom builds of Caddy which may include many handler plugins. Forcing users to specify `order` or wrap things in `route` causes some friction that would be nice to smooth out.

The reason we never did this in the past is because we didn't want to allow plugins to depend upon eachother, because we can't guarantee the order in which plugins are loaded. We also didn't want the ordering to become unreliable when the order of directives in the standard distribution changes.

We can mitigate those factors by doing two things:
- Only allow plugins to register an order for their directive in relation to a directive in Caddy's standard distribution (we keep a `defaultDirectiveOrder` which doesn't change, to ensure this, and modify `directiveOrder` accordingly)
- Only allow plugins to order "before" and "after" another directive (and no "first" or "last" like the `order` global option allows) to reduce the chance of ordering that doesn't make sense.

The API looks like this:

```go
httpcaddyfile.RegisterHandlerDirective("my_directive", parseDirective)
httpcaddyfile.RegisterDirectiveOrder("my_directive", "after", "redir") // new
```

Or course, there's still a slight problem if two plugins try to both order the same way in relation to a standard directive (e.g. both are "before headers") because their order will not be guaranteed. There are some complicated ways we could mitigate this though, for example we could invalidate the ordering for both directives since they're in conflict, forcing the user to specify their own order for both. But ultimately I think this it probably not an immediate need, and we can expand upon that later.

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -154,15 +154,15 @@ func RegisterHandlerDirective(dir string, setupFunc UnmarshalHandlerFunc) {
 // to ('before' or 'after') a directive included in Caddy's
 // standard distribution. It cannot be relative to another
 // plugin's directive.
-func RegisterDirectiveOrder(dir string, position string, standardDir string) {
+func RegisterDirectiveOrder(dir string, position Positional, standardDir string) {
 	// check if directive was already ordered
 	if directiveIsOrdered(dir) {
 		panic("directive '" + dir + "' already ordered")
 	}
 
 	switch position {
-	case "before":
-	case "after":
+	case Before:
+	case After:
 	default:
 		panic("the 2nd argument must be either 'before' or 'after', got '" + position + "'")
 	}
@@ -186,9 +186,9 @@ func RegisterDirectiveOrder(dir string, position string, standardDir string) {
 		if d != standardDir {
 			continue
 		}
-		if position == "before" {
+		if position == Before {
 			newOrder = append(newOrder[:i], append([]string{dir}, newOrder[i:]...)...)
-		} else if position == "after" {
+		} else if position == After {
 			newOrder = append(newOrder[:i+1], append([]string{dir}, newOrder[i+1:]...)...)
 		}
 		break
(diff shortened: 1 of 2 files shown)


caddyconfig/httpcaddyfile/directives.go
```diff
@@ -128,6 +140,59 @@ func RegisterHandlerDirective(dir string, setupFunc UnmarshalHandlerFunc) {
...
+// standard distribution. It cannot be relative to another
+// plugin's directive.
+func RegisterDirectiveOrder(dir string, position string, standardDir string) {
+	// check if directive was already ordered
+	if directiveIsOrdered(dir) {
+		panic("directive '" + dir + "' already ordered")
+	}
+
+	switch position {
+	case "before":
```
dunglas, line 161: Maybe create constants (and a dedicated type?) for these values?

francislavoie: Yeah, good point. Wasn't really necessary before because it was only used in the Caddyfile (and Caddyfile config can't refer to Go consts lol)

francislavoie: Done :+1:

dunglas (approved):

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -130,6 +142,56 @@ func RegisterHandlerDirective(dir string, setupFunc UnmarshalHandlerFunc) {
...
+// directive from a plugin.
+//
+// This is useful when a plugin has a well-understood place
+// it should run in the middleware pipeline, and it allows
+// users to avoid having to define the order themselves.
+//
+// The directive dir may be placed in the position relative
+// to ('before' or 'after') a directive included in Caddy's
+// standard distribution. It cannot be relative to another
+// plugin's directive.
```
mholt, line 155:
```suggestion
// plugin's directive.
//
// EXPERIMENTAL: This API may change or be removed.
```

mholt (changes requested):
Thanks, this is quite a limited function but it sounds like some people will still find it useful -- we can try it! But I do want to mark it as experimental for now.

mholt (approved):

PR #5867, opened 2023-10-07
fileserver: Add canonical URL to browse template

Forza-tng:
When contents are equal, but maybe just a sort order is different, it is good to add `<link rel="canonical" href="{{.Path}}/" />`. This helps search engines propeely index the page.

I also added a link to the home page with the name of `{{.Host}}` just above the bread crumbs to make the page clearer.
Screenshot_20231007_134435_Opera

https://paste.tnonline.net/files/28Wun5CQZiqA_Screenshot_20231007_134435_Opera.png

### DIFF
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -774,7 +774,6 @@ footer {
 <body onload="initPage()">
 	<header>
 		<div class="wrapper">
-			<a href="//{{.Host}}" alt="Home">&nbsp;{{.Host}}</a>
 			<div class="breadcrumbs">Folder Path</div>
 				<h1>
 					{{range $i, $crumb := .Breadcrumbs}}<a href="{{html $crumb.Link}}">{{html $crumb.Text}}</a>{{if ne $i 0}}/{{end}}{{end}}
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -299,6 +299,7 @@
 <html>
 	<head>
 		<title>{{html .Name}}</title>
+		<link rel="canonical" href="{{.Path}}/"  />
 		<meta charset="utf-8">
 		<meta name="color-scheme" content="light dark">
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
@@ -773,6 +774,7 @@ footer {
 <body onload="initPage()">
 	<header>
 		<div class="wrapper">
+			<a href="//{{.Host}}" alt="Home">&nbsp;{{.Host}}</a>
 			<div class="breadcrumbs">Folder Path</div>
 				<h1>
 					{{range $i, $crumb := .Breadcrumbs}}<a href="{{html $crumb.Link}}">{{html $crumb.Text}}</a>{{if ne $i 0}}/{{end}}{{end}}


francislavoie:
I'm not sure about the Host thing. That might sometimes be undesirable, like if a file server is served from behind a proxy (which has rewritten the Host before it reaches Caddy). 

Also it reads confusingly because the text above it says "folder path" but now there's a domain directly in between the two.

I think simply making the leading `/` clickable might be enough (though I guess it might not be immediately obvious to everyone).

No objections to the canonical link though, seems reasonable.

Forza-tng:
My proposal uses relative path in the canonical href. This is supported, but not recommended by Google.

https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls

> Use absolute paths rather than relative paths with the `rel="canonical"` link element. Even though relative paths are supported by Google, they can cause problems in the long run (for example, if you unintentionally allow your testing site to be crawled) and thus we don't recommend them.

However I have not found a template function for outputing current URL scheme*

Something like this would be preferred, if it were supported:
```html
<link rel="canonical" href="{{.Protocol}}{{.Host}}{{.Path}}/" />
```
 https://caddyserver.com/docs/modules/http.handlers.templates#docs

Forza-tng:
> I'm not sure about the Host thing. That might sometimes be undesirable, like if a file server is served from behind a proxy (which has rewritten the Host before it reaches Caddy).

Indeed. Better to use {{.Req.Header.Get "host"}} 


> 
> Also it reads confusingly because the text above it says "folder path" but now there's a domain directly in between the two.
> 
> I think simply making the leading `/` clickable might be enough (though I guess it might not be immediately obvious to everyone).

OK fair enough. :) 

> 
> No objections to the canonical link though, seems reasonable.

francislavoie:
An absolute URL is not trivial in the case that Caddy is behind a proxy, because you may need to read `X-Forwarded-Host` and such. I think we need to stick with only the path.

francislavoie:
> Indeed. Better to use {{.Req.Header.Get "host"}}

No, that's not different. The problem is that fundamentally the Host header is not a guarantee of what is shown by the browser. If Caddy is behind a proxy, the Host might be a machine's name or something, and shouldn't be made public. It can be an information leak.

Forza-tng:
> > Indeed. Better to use {{.Req.Header.Get "host"}}
> 
> No, that's not different. The problem is that fundamentally the Host header is not a guarantee of what is shown by the browser. If Caddy is behind a proxy, the Host might be a machine's name or something, and shouldn't be made public. It can be an information leak.

I understand. Perhaps we can introduce a function that checks for a title argument in the file_server directive? If it isn't set, then omit the section. Likewise a similar method can be used for the full canonical URL. What do you think about that?

Moved the link above folder path section. Much nicer. Thanks.
   
Screenshot_20231007_142938_Opera

Forza-tng:
I can resubmit a PR with only the canonical url part. Would that be ok?

mholt:
Thanks; I agree the canonical URL part is probably worth keeping, but the link to home is not a good assumption for us to make.

francislavoie:
Sorry I missed your comment https://github.com/caddyserver/caddy/pull/5867#issuecomment-1751701295 because it got collapsed by GitHub mobile.

I think a `home <title> <href>` subdirective to opt-in to a link might be okay. But I don't think we should try to add that by default with no input because there's no safe heuristic for guessing the right thing to use.

mholt (approved):
Since this now is only setting the canonical URL, I view this as a bug fix, since, indeed, there are multiple ways to view the same content and there is only 1 true path, so I think this is a fix for search engines (if they even care to index these pages).

PR #5869, opened 2023-10-08
encode: Add `application/wasm*` to the default content types

Christoph-D:
After this change, WebAssembly files will be encoded (compressed). This is useful because WebAssembly files are uncompressed.

Pull request #2624 did the same thing 4 years ago but it seems that the change got lost.


francislavoie (approved):
LGTM! Thanks

PR #5873, opened 2023-10-10
#5871  Added "-v" flag as well along with "version"

### LINKED ISSUE #5871: Add `-v` flag support to main CLI command
Most a note-to-self to implement this, but if someone wants to open a PR before I do, by all means.

In the past, we used Go's stdlib flags which meant it was not possible to support `caddy -v`. But we've since switched to Cobra, so it's now possible to do so.

For context, some users were complaining about this on social media:
- https://frfsh.plus.st/notes/9hxugdnvdqnjgnm7
- https://mstdn.social/@mattwilcox/111201843985131969

Tushar-kalsi:
**Fixed #5871**  
Added **-v** RegisterCommond component  in /cmd/commonds.go file which will work as "**caddy -v**" in cli to show the current version . Not removed existing version RegisterCommond component which means both "**caddy version**" and "caddy -v" works same .

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -221,6 +221,25 @@ documentation: https://go.dev/doc/modules/version-numbers
 	})
 
 	RegisterCommand(Command{
+		Name:  "-v",
+		Short: "Prints the version",
+		Long: `
+Prints the version of this Caddy binary.
+
+Version information must be embedded into the binary at compile-time in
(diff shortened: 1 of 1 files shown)


mohammed90:
Thank you for the PR! The implementation should be a flag, not a command with hyphen in its name. I have the feature implemented locally and will push it later. Thank you!

PR #5879, opened 2023-10-11
chore: Force the Go version for govulncheck

francislavoie:


mholt (approved):
Thanks!

... I wonder if we should have an action that updates all these workflows whenever a new Go version comes out.

mohammed90 (approved):

francislavoie:
> ... I wonder if we should have an action that updates all these workflows whenever a new Go version comes out.

In case you misunderstand, `~1.21.0` means "latest `1.21.x`", so it's fine.

But either way, meh. We always have to do a bunch of updating in other places when we bump Go versions (either minimum or maximum or both) like docs, README, etc. It has to be a deliberate choice IMO.

mholt:
That's right -- but when Go 1.22 comes out, we have to change it again.

Just seems a bit tedious with all the workflows we're getting.

PR #5881, opened 2023-10-12
httpcaddyfile: Remove port from logger names

francislavoie:
This took a long while to figure out. https://caddy.community/t/logging-per-subdomain-reverse-proxies/20586

When a site address has a port, the port number would be included as part of `logger_names` in the JSON output. This means that when the request is trying to match a logger, it would never match (therefore not write any access logs) because it would only use the `Host` to match it, which wouldn't have the port if some port forwarding happened.

This is safe because the Caddyfile adapter always makes separate servers even if the same domain with different ports are used in the same site block, so the hostname is unique to each server.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -17,6 +17,7 @@ package httpcaddyfile
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"reflect"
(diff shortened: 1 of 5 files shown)


mholt (approved):
Ah... I think you're right. Thanks for this change. :+1:

(See, I'm slowly working through my backlog :grin: )

PR #5891, opened 2023-10-13
cmd: upgrade: resolve symlink of the executable

mohammed90:
It occurred to me that users of `dpkg-divert` (per our instructions [here](https://caddyserver.com/docs/build#package-support-files-for-custom-builds-for-debianubunturaspbian)) may have their symlink `/usr/local/bin/caddy` be overwritten by the `upgrade`, `add-package`, or `remove-package` commands because we're replacing the symlink source, i.e. `/usr/local/bin/caddy`, and not its target. This may cause some system inconsistency or confusion. The fix is simply to resolve the symlink.

### DIFF
--- a/cmd/packagesfuncs.go
+++ b/cmd/packagesfuncs.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"reflect"
 	"runtime"
 	"runtime/debug"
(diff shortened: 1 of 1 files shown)


mholt:
Ah, good point. Thank you Mohammed!

mholt (approved):

francislavoie (approved):

PR #5896, opened 2023-10-14
httpcaddyfile: Fix TLS automation policy merging with get_certificate

francislavoie:
Fixes an issue reported in https://caddy.community/t/21415

We need to check that the `ManagerRaw` (i.e. `get_certificate` usually) is also the same, otherwise automation policies get merged when they shouldn't.


mholt (approved):
Ah, thanks. The precarious act of adding fields...

PR #5898, opened 2023-10-15
templates: Clarify `include` args docs, add `.ClientIP`

### LINKED ISSUE #5897: include template function with arguments
The documentation for the include template function states:

      Includes the contents of another file and renders in-place.
      Optionally can pass key-value pairs as arguments to be accessed by the included file.

     {{include "path/to/file.html"}}  // no arguments
     {{include "path/to/file.html" "arg1" 2 "value 3"}}  // with arguments

How does one access these arguments?

francislavoie:
Fixes #5897

### DIFF
--- a/modules/caddyhttp/templates/templates.go
+++ b/modules/caddyhttp/templates/templates.go
@@ -125,12 +125,13 @@ func init() {
 //
 // Includes the contents of another file, rendering it in-place.
 // Optionally can pass key-value pairs as arguments to be accessed
-// by the included file. Note that the contents are NOT escaped,
-// so you should only include trusted template files.
+// by the included file. Use [`.Args N`](#args) to access the N-th
+// argument, 0-indexed. Note that the contents are NOT escaped, so
+// you should only include trusted template files.
 //
 // ```
 // {{include "path/to/file.html"}}  // no arguments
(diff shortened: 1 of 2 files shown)


mohammed90 (approved):

francislavoie:
I noticed a few more minor fixes I want to make, so I'll do another commit shortly... also going to add `.ClientIP` because why not (i.e. similar to `.RemoteIP` but with `trusted_proxies`)

modules/caddyhttp/templates/templates.go
```diff
@@ -246,14 +261,12 @@ func init() {
...
-//		"sitename": "A Caddy site"
-//	}
-//
+// ```json
+// {
+// "template": "blog",
+// "title": "Blog Homepage",
+// "sitename": "A Caddy site"
+// }
 // ```
```
francislavoie, line 270: This is really stupid, but if we have any kind of indentation between `{ }` the lint-on-save in VSCode will reformat it as how it was before, making it look stupid after markdown parsing. So for now the only solution I could find is to remove any JSON indentation :sob:

mholt: Oh! Haha, yeah you're right. I don't know a workaround. That's OK.

PR #5901, opened 2023-10-17
storage: Refactor with early return pattern in homeDir func

thor-son:
Refactoring in homeDir func
Use early return pattern and remove duplicate conditions.

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -38,10 +38,12 @@ type StorageConverter interface {
 // which returns "/sdcard".
 func HomeDir() string {
 	home := homeDirUnsafe()
-	if home == "" && runtime.GOOS == "android" {
-		home = "/sdcard"
+	if home != "" {
+		return home
 	}
-	if home == "" {
+	if runtime.GOOS == "android" {
+		home = "/sdcard"
+	} else {
 		home = "."
 	}
 	return home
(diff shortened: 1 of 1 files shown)


mholt (changes requested):
Thanks for this, but in my opinion the original is easier to read and understand...

Maybe that's just me. But in the original, it's just a matter of "try this value," and each `if` block is basically "if empty, try this". With a single return line. The changes here introduce more complex branching and I think the logic is a little more confusing. But again, maybe that's just me personally...

thor-son:
I'm so appreciate your reply.
after Read your comment, I agree your opinion.
Then how about condition return pattern like this?
``` patch
func HomeDir() string {
	home := homeDirUnsafe()
	if home == "" && runtime.GOOS == "android" {
-		home = "/sdcard"
+		return "/sdcard"
	}
	if home == "" {
-		home = "."
+		return "."
	}
	return home
}
```

mholt:
Maybe, but what benefit is that? The way I see it, it just makes the last return a bit awkward and pointless, except so it will compile. The last return would always be `return ""` but it's not necessary with the original flow.

I actually remember rewriting these functions several times at first until I settled on the current code, because I now seem to remember going through all these variations you're proposing :grin:

If it's alright with you I'll respectfully close this PR, with gratitude for your desire to contribute. I think it's cool you've been reading the Caddy code base!

PR #5910, opened 2023-10-22
caddyhttp: Adjust `scheme` placeholder docs

francislavoie:
Discussed in https://github.com/caddyserver/website/pull/348

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -73,7 +73,7 @@ func init() {
 // `{http.request.remote.host}` | The host (IP) part of the remote client's address
 // `{http.request.remote.port}` | The port part of the remote client's address
 // `{http.request.remote}` | The address of the remote client
-// `{http.request.scheme}` | The request scheme
+// `{http.request.scheme}` | The request scheme, typically `http` or `https`
 // `{http.request.tls.version}` | The TLS version name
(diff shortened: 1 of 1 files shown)


mholt (approved):

modules/caddyhttp/app.go
```diff
@@ -73,7 +73,7 @@ func init() {
 // `{http.request.remote.host}` | The host (IP) part of the remote client's address
 // `{http.request.remote.port}` | The port part of the remote client's address
 // `{http.request.remote}` | The address of the remote client
-// `{http.request.scheme}` | The request scheme
+// `{http.request.scheme}` | The request scheme, typically `http` or `https`
```
coolaj86, line 76: Are there other possible values?

I thought this was essentially an alias for "encrypted" ans always said http for plain text, and always said https for TLS, regardless of HTTP2, etc.

francislavoie: The code is literally this right now:
```go
			case "http.request.scheme":
				if req.TLS != nil {
					return "https", true
				}
				return "http", true
```

So it's only those two values.

> But I suppose Caddy can now be an FTP or SMTP server via Plugin?

Yes, but that wouldn't be under `http.request` because it would happen in a different Caddy app.

Issue #5863, opened 2023-10-06
Built in proxy protocol doesn't work with AWS load balancer


KorvinSzanto:
I'm trying to migrate away from `pires/go-proxyproto` to the newly embedded proxyprotocol in caddy but I'm having a hard time getting things working with proxy protocol enabled AWS load balancers.
My config looks like this:
```
{
    debug
    servers {
        listener_wrappers {
            proxy_protocol {
                    timeout 2s
                    allow 192.168.86.1/24
            }
            tls
        }
    }
}



caddytest.mydomain.com:443 {
    reverse_proxy https://127.0.0.1:8443 {
        transport http {
            tls_insecure_skip_verify
        }
    }
}
```

And attempting to access caddy through the LB outputs:
```
2023/10/06 16:58:04.254	DEBUG	http.stdlib	http: TLS handshake error from 192.168.86.42:5943: invalid length
```

and gives me `ERR_CONNECTION_CLOSED` in the browser.
If I use [my fork](https://github.com/KorvinSzanto/caddy-go-proxyproto) of [rocketreferrals/caddy-go-proxyproto](https://github.com/rocketreferrals/caddy-go-proxyproto) things work as expected.

francislavoie:
You fork doesn't have `allow`. Does it work any better if you omit that from your config (i.e. is that actually the right CIDR to trust)?

I don't use AWS so I won't be able to debug this. We use https://github.com/mastercactapus/proxyprotocol. Apparently this is the same problem as https://github.com/mastercactapus/caddy2-proxyprotocol/issues/8. /cc @mastercactapus if you have any ideas.

KorvinSzanto:
> Does it work any better if you omit that from your config
Same behavior with the allow removed

That issue does sound like the same issue I'm facing

francislavoie:
This has apparently been fixed by the `tlv` branch. You can build Caddy with the fix, see https://github.com/mastercactapus/proxyprotocol/issues/3#issuecomment-1755774950.

PR #5915, opened 2023-10-24
proxyprotocol: use github.com/pires/go-proxyproto

mohammed90:
Further enhancements can be added later, e.g. advanced allow/denylist policies.

I'll test it out soon, unless someone beats me to the test.

### DIFF
--- a/modules/caddyhttp/proxyprotocol/listenerwrapper.go
+++ b/modules/caddyhttp/proxyprotocol/listenerwrapper.go
@@ -15,11 +15,10 @@
 package proxyprotocol
 
 import (
-	"fmt"
 	"net"
 	"time"
 
-	"github.com/mastercactapus/proxyprotocol"
+	goproxy "github.com/pires/go-proxyproto"
 
 	"github.com/caddyserver/caddy/v2"
 )
@@ -39,31 +38,29 @@ type ListenerWrapper struct {
 	// allow/require PROXY headers from.
 	Allow []string `json:"allow,omitempty"`
 
-	rules []proxyprotocol.Rule
+	policies []goproxy.PolicyFunc
 }
 
 // Provision sets up the listener wrapper.
 func (pp *ListenerWrapper) Provision(ctx caddy.Context) error {
-	rules := make([]proxyprotocol.Rule, 0, len(pp.Allow))
-	for _, s := range pp.Allow {
-		_, n, err := net.ParseCIDR(s)
+	if len(pp.Allow) > 0 {
+		allowlist, err := goproxy.LaxWhiteListPolicy(pp.Allow)
 		if err != nil {
-			return fmt.Errorf("invalid subnet '%s': %w", s, err)
+			return err
 		}
-		rules = append(rules, proxyprotocol.Rule{
-			Timeout: time.Duration(pp.Timeout),
-			Subnet:  n,
-		})
+		pp.policies = append(pp.policies, allowlist)
 	}
-
-	pp.rules = rules
-
 	return nil
 }
 
 // WrapListener adds PROXY protocol support to the listener.
 func (pp *ListenerWrapper) WrapListener(l net.Listener) net.Listener {
-	pl := proxyprotocol.NewListener(l, time.Duration(pp.Timeout))
-	pl.SetFilter(pp.rules)
+	pl := &goproxy.Listener{
+		Listener:          l,
+		ReadHeaderTimeout: time.Duration(pp.Timeout),
+	}
+	if len(pp.policies) > 0 {
+		pl.Policy = pp.policies[0]
+	}
 	return pl
 }
--- a/modules/caddyhttp/proxyprotocol/listenerwrapper.go
+++ b/modules/caddyhttp/proxyprotocol/listenerwrapper.go
@@ -16,6 +16,7 @@ package proxyprotocol
 
 import (
 	"net"
+	"net/netip"
 	"time"
 
 	goproxy "github.com/pires/go-proxyproto"
@@ -37,12 +38,12 @@ type ListenerWrapper struct {
 	// Allow is an optional list of CIDR ranges to
 	// allow/require PROXY headers from.
 	Allow []string `json:"allow,omitempty"`
-	allow []*net.IPNet
+	allow []netip.Prefix
 
 	// Denby is an optional list of CIDR ranges to
 	// deny PROXY headers from.
 	Deny []string `json:"deny,omitempty"`
-	deny []*net.IPNet
+	deny []netip.Prefix
 
 	// Accepted values are: ignore, use, reject, require, skip
 	// default: ignore
@@ -55,27 +56,32 @@ type ListenerWrapper struct {
 // Provision sets up the listener wrapper.
 func (pp *ListenerWrapper) Provision(ctx caddy.Context) error {
 	for _, cidr := range pp.Allow {
-		_, ipnet, err := net.ParseCIDR(cidr)
+		ipnet, err := netip.ParsePrefix(cidr)
 		if err != nil {
 			return err
 		}
 		pp.allow = append(pp.allow, ipnet)
 	}
 	for _, cidr := range pp.Deny {
-		_, ipnet, err := net.ParseCIDR(cidr)
+		ipnet, err := netip.ParsePrefix(cidr)
 		if err != nil {
 			return err
 		}
 		pp.deny = append(pp.deny, ipnet)
 	}
 	pp.policy = func(upstream net.Addr) (goproxy.Policy, error) {
+		// trust unix sockets
+		if network := upstream.Network(); caddy.IsUnixNetwork(network) {
+			return goproxy.USE, nil
+		}
 		ret := pp.FallbackPolicy
 		host, _, err := net.SplitHostPort(upstream.String())
 		if err != nil {
 			return goproxy.REJECT, err
 		}
-		ip := net.ParseIP(host)
-		if ip == nil {
+
+		ip, err := netip.ParseAddr(host)
+		if err != nil {
 			return goproxy.REJECT, err
 		}
 		for _, ipnet := range pp.deny {
--- a/modules/caddyhttp/proxyprotocol/listenerwrapper.go
+++ b/modules/caddyhttp/proxyprotocol/listenerwrapper.go
@@ -37,18 +37,59 @@ type ListenerWrapper struct {
 	// Allow is an optional list of CIDR ranges to
 	// allow/require PROXY headers from.
 	Allow []string `json:"allow,omitempty"`
+	allow []*net.IPNet
 
-	policies []goproxy.PolicyFunc
+	// Denby is an optional list of CIDR ranges to
+	// deny PROXY headers from.
+	Deny []string `json:"deny,omitempty"`
+	deny []*net.IPNet
+
+	// Accepted values are: ignore, use, reject, require, skip
+	// default: ignore
+	// Policy definitions are here: https://pkg.go.dev/github.com/pires/go-proxyproto@v0.7.0#Policy
+	FallbackPolicy Policy `json:"fallback_policy,omitempty"`
+
+	policy goproxy.PolicyFunc
 }
 
 // Provision sets up the listener wrapper.
 func (pp *ListenerWrapper) Provision(ctx caddy.Context) error {
-	if len(pp.Allow) > 0 {
-		allowlist, err := goproxy.LaxWhiteListPolicy(pp.Allow)
+	for _, cidr := range pp.Allow {
+		_, ipnet, err := net.ParseCIDR(cidr)
 		if err != nil {
 			return err
 		}
-		pp.policies = append(pp.policies, allowlist)
+		pp.allow = append(pp.allow, ipnet)
+	}
+	for _, cidr := range pp.Deny {
+		_, ipnet, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return err
+		}
+		pp.deny = append(pp.deny, ipnet)
+	}
+	pp.policy = func(upstream net.Addr) (goproxy.Policy, error) {
+		ret := pp.FallbackPolicy
+		host, _, err := net.SplitHostPort(upstream.String())
+		if err != nil {
+			return goproxy.REJECT, err
+		}
+		ip := net.ParseIP(host)
+		if ip == nil {
+			return goproxy.REJECT, err
+		}
+		for _, ipnet := range pp.deny {
+			if ipnet.Contains(ip) {
+				return goproxy.REJECT, nil
+			}
+		}
+		for _, ipnet := range pp.allow {
+			if ipnet.Contains(ip) {
+				ret = PolicyUSE
+				break
+			}
+		}
+		return policyToGoProxyPolicy[ret], nil
 	}
 	return nil
 }
@@ -59,8 +100,6 @@ func (pp *ListenerWrapper) WrapListener(l net.Listener) net.Listener {
 		Listener:          l,
 		ReadHeaderTimeout: time.Duration(pp.Timeout),
 	}
-	if len(pp.policies) > 0 {
-		pl.Policy = pp.policies[0]
-	}
+	pl.Policy = pp.policy
 	return pl
 }
--- a/modules/caddyhttp/proxyprotocol/policy.go
+++ b/modules/caddyhttp/proxyprotocol/policy.go
@@ -1,0 +1,82 @@
+package proxyprotocol
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	goproxy "github.com/pires/go-proxyproto"
+)
+
+type Policy int
+
+// as defined in: https://pkg.go.dev/github.com/pires/go-proxyproto@v0.7.0#Policy
+const (
+	// IGNORE address from PROXY header, but accept connection
+	PolicyIGNORE Policy = iota
+	// USE address from PROXY header
+	PolicyUSE
+	// REJECT connection when PROXY header is sent
+	// Note: even though the first read on the connection returns an error if
+	// a PROXY header is present, subsequent reads do not. It is the task of
+	// the code using the connection to handle that case properly.
+	PolicyREJECT
+	// REQUIRE connection to send PROXY header, reject if not present
+	// Note: even though the first read on the connection returns an error if
+	// a PROXY header is not present, subsequent reads do not. It is the task
+	// of the code using the connection to handle that case properly.
+	PolicyREQUIRE
+	// SKIP accepts a connection without requiring the PROXY header
+	// Note: an example usage can be found in the SkipProxyHeaderForCIDR
+	// function.
+	PolicySKIP
+)
+
+var policyToGoProxyPolicy = map[Policy]goproxy.Policy{
+	PolicyUSE:     goproxy.USE,
+	PolicyIGNORE:  goproxy.IGNORE,
+	PolicyREJECT:  goproxy.REJECT,
+	PolicyREQUIRE: goproxy.REQUIRE,
+	PolicySKIP:    goproxy.SKIP,
+}
+
+var policyMap = map[Policy]string{
+	PolicyUSE:     "USE",
+	PolicyIGNORE:  "IGNORE",
+	PolicyREJECT:  "REJECT",
+	PolicyREQUIRE: "REQUIRE",
+	PolicySKIP:    "SKIP",
+}
+
+var policyMapRev = map[string]Policy{
+	"USE":     PolicyUSE,
+	"IGNORE":  PolicyIGNORE,
+	"REJECT":  PolicyREJECT,
+	"REQUIRE": PolicyREQUIRE,
+	"SKIP":    PolicySKIP,
+}
+
+// MarshalText implements the text marshaller method.
+func (x Policy) MarshalText() ([]byte, error) {
+	return []byte(policyMap[x]), nil
+}
+
+// UnmarshalText implements the text unmarshaller method.
+func (x *Policy) UnmarshalText(text []byte) error {
+	name := string(text)
+	tmp, err := parsePolicy(name)
+	if err != nil {
+		return err
+	}
+	*x = tmp
+	return nil
+}
+
+func parsePolicy(name string) (Policy, error) {
+	if x, ok := policyMapRev[strings.ToUpper(name)]; ok {
+		return x, nil
+	}
+	return Policy(0), fmt.Errorf("%s is %w", name, errInvalidPolicy)
+}
+
+var errInvalidPolicy = errors.New("invalid policy")
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -215,7 +215,7 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
 				},
 			}
 			// The src and dst have to be of the same address family. As we don't know the original
-			// dst address (it's kind of impossible to know) and this address is generelly of very
+			// dst address (it's kind of impossible to know) and this address is generally of very
 			// little interest, we just set it to all zeros.
 			switch {
 			case proxyProtocolInfo.AddrPort.Addr().Is4():
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -28,7 +28,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/mastercactapus/proxyprotocol"
+	"github.com/pires/go-proxyproto"
 	"go.uber.org/zap"
 	"golang.org/x/net/http2"
 
@@ -207,44 +207,42 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
 			if !ok {
 				return nil, fmt.Errorf("failed to get proxy protocol info from context")
 			}
-
-			// The src and dst have to be of the some address family. As we don't know the original
+			header := proxyproto.Header{
+				SourceAddr: &net.TCPAddr{
+					IP:   proxyProtocolInfo.AddrPort.Addr().AsSlice(),
+					Port: int(proxyProtocolInfo.AddrPort.Port()),
+					Zone: proxyProtocolInfo.AddrPort.Addr().Zone(),
+				},
+			}
+			// The src and dst have to be of the same address family. As we don't know the original
 			// dst address (it's kind of impossible to know) and this address is generelly of very
 			// little interest, we just set it to all zeros.
-			var destIP net.IP
 			switch {
 			case proxyProtocolInfo.AddrPort.Addr().Is4():
-				destIP = net.IPv4zero
+				header.TransportProtocol = proxyproto.TCPv4
+				header.DestinationAddr = &net.TCPAddr{
+					IP: net.IPv4zero,
+				}
 			case proxyProtocolInfo.AddrPort.Addr().Is6():
-				destIP = net.IPv6zero
+				header.TransportProtocol = proxyproto.TCPv6
+				header.DestinationAddr = &net.TCPAddr{
+					IP: net.IPv6zero,
+				}
 			default:
 				return nil, fmt.Errorf("unexpected remote addr type in proxy protocol info")
 			}
 
-			// TODO: We should probably migrate away from net.IP to use netip.Addr,
-			// but due to the upstream dependency, we can't do that yet.
 			switch h.ProxyProtocol {
 			case "v1":
-				header := proxyprotocol.HeaderV1{
-					SrcIP:    net.IP(proxyProtocolInfo.AddrPort.Addr().AsSlice()),
-					SrcPort:  int(proxyProtocolInfo.AddrPort.Port()),
-					DestIP:   destIP,
-					DestPort: 0,
-				}
+				header.Version = 1
 				caddyCtx.Logger().Debug("sending proxy protocol header v1", zap.Any("header", header))
-				_, err = header.WriteTo(conn)
 			case "v2":
-				header := proxyprotocol.HeaderV2{
-					Command: proxyprotocol.CmdProxy,
-					Src:     &net.TCPAddr{IP: net.IP(proxyProtocolInfo.AddrPort.Addr().AsSlice()), Port: int(proxyProtocolInfo.AddrPort.Port())},
-					Dest:    &net.TCPAddr{IP: destIP, Port: 0},
-				}
+				header.Version = 2
 				caddyCtx.Logger().Debug("sending proxy protocol header v2", zap.Any("header", header))
-				_, err = header.WriteTo(conn)
 			default:
 				return nil, fmt.Errorf("unexpected proxy protocol version")
 			}
-
+			_, err = header.WriteTo(conn)
 			if err != nil {
 				// identify this error as one that occurred during
 				// dialing, which can be important when trying to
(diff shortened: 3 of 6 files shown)


modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -207,44 +207,42 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
...
-			// The src and dst have to be of the some address family. As we don't know the original
+			header := proxyproto.Header{
+				SourceAddr: &net.TCPAddr{
+					IP:   proxyProtocolInfo.AddrPort.Addr().AsSlice(),
+					Port: int(proxyProtocolInfo.AddrPort.Port()),
+					Zone: proxyProtocolInfo.AddrPort.Addr().Zone(),
+				},
+			}
+			// The src and dst have to be of the same address family. As we don't know the original
 			// dst address (it's kind of impossible to know) and this address is generelly of very
```
francislavoie, line 218: Noticed this typo (from before)

```suggestion
			// dst address (it's kind of impossible to know) and this address is generally of very
```

francislavoie:
The code looks good to me. Glad we could retain the same options as-is!

modules/caddyhttp/proxyprotocol/listenerwrapper.go
```diff
@@ -39,31 +38,29 @@ type ListenerWrapper struct {
 	// allow/require PROXY headers from.
 	Allow []string `json:"allow,omitempty"`
 
-	rules []proxyprotocol.Rule
+	policies []goproxy.PolicyFunc
```
WeidiDeng, line 41: Why store this in a slice when there can be one func at once? proxyproto takes a slice of IP address and return a function. There is no need to store it in a slice.

modules/caddyhttp/proxyprotocol/listenerwrapper.go
```diff
@@ -39,31 +38,29 @@ type ListenerWrapper struct {
...
 // WrapListener adds PROXY protocol support to the listener.
 func (pp *ListenerWrapper) WrapListener(l net.Listener) net.Listener {
-	pl := proxyprotocol.NewListener(l, time.Duration(pp.Timeout))
-	pl.SetFilter(pp.rules)
+	pl := &goproxy.Listener{
+		Listener:          l,
+		ReadHeaderTimeout: time.Duration(pp.Timeout),
+	}
+	if len(pp.policies) > 0 {
+		pl.Policy = pp.policies[0]
```
WeidiDeng, line 63: Just use a field of type `PolicyFunc` and assign it here.

WeidiDeng (changes requested):
A few nitpicks only, LGTM.

KorvinSzanto:
This does seem to resolve #5863 for me.

Embraser01:
Can confirm it works as well as the one we made over at https://github.com/caddyserver/ingress/tree/master/pkg/proxy.

> I created a PR to replace it by this one once merged and released

mholt (approved):
Great, thanks for testing it out.

This is an elegant change. Once Weidi's comments are considered (or replied to) I'm good with merging this. Thank you @mohammed90 !!

modules/caddyhttp/proxyprotocol/listenerwrapper.go
mohammed90: It's my attempt at future-proofing. proxyproto has a selection of [policies](https://pkg.go.dev/github.com/pires/go-proxyproto#PolicyFunc). We can also build more as long as they conform to the `PolicyFunc` type. My thinking is we could expand the implementation later to unionize policies.

mohammed90:
> 🎉 Can confirm it works as well as the one we made over at https://github.com/caddyserver/ingress/tree/master/pkg/proxy.
> 
> > I created a PR to replace it by this one once merged and released

@Embraser01, I see your implementation uses the `REQUIRE` policy, while this PR goes with `USE`. Should I expose this configuration option?

Embraser01:
> > 🎉 Can confirm it works as well as the one we made over at [caddyserver/ingress@`master`/pkg/proxy](https://github.com/caddyserver/ingress/tree/master/pkg/proxy?rgh-link-date=2023-10-30T07%3A54%3A51Z).
> > > I created a PR to replace it by this one once merged and released
> 
> @Embraser01, I see your implementation uses the `REQUIRE` policy, while this PR goes with `USE`. Should I expose this configuration option?

It could be nice to expose this option as some user may want to force proxy protocol. Currently the ingress controller doesn't let choose between those modes but I think it should probably be USE by default (don't remember why I set it up to REQUIRE)

modules/caddyhttp/proxyprotocol/listenerwrapper.go
```diff
@@ -38,32 +37,69 @@ type ListenerWrapper struct {
...
+	for _, cidr := range pp.Deny {
+		_, ipnet, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return err
+		}
+		pp.deny = append(pp.deny, ipnet)
+	}
+	pp.policy = func(upstream net.Addr) (goproxy.Policy, error) {
+		ret := pp.FallbackPolicy
+		host, _, err := net.SplitHostPort(upstream.String())
```
WeidiDeng, line 73: What if the connection is made from a unix socket? It will be '@'. gin-gonic has a similar [bug](https://github.com/gin-gonic/gin/issues/2718) caused by [2632](https://github.com/gin-gonic/gin/pull/2632/files#diff-552f47512a00afe5fc6850cc9ddc830a6daeca162750e50aab4ed549685e0253R761).

modules/caddyhttp/proxyprotocol/policy.go
```diff
@@ -0,0 +1,82 @@
...
+
+var policyMapRev = map[string]Policy{
+	"USE":     PolicyUSE,
+	"IGNORE":  PolicyIGNORE,
+	"REJECT":  PolicyREJECT,
+	"REQUIRE": PolicyREQUIRE,
+	"SKIP":    PolicySKIP,
+}
+
+// MarshalText implements the text marshaller method.
```
WeidiDeng, line 59: Are these two methods used for representing `FallbackPolicy` when caddy config is marshalled to json? Why not implement `(Un)MarshalJSON` instead and add comments. (It has to be `(Un)MarshalJSON` though. `(Un)MarshalText` is used with plain text, and golang json checks for if the result is valid json).

modules/caddyhttp/proxyprotocol/listenerwrapper.go
```diff
@@ -38,32 +37,69 @@ type ListenerWrapper struct {
 	// Allow is an optional list of CIDR ranges to
 	// allow/require PROXY headers from.
 	Allow []string `json:"allow,omitempty"`
+	allow []*net.IPNet
```
WeidiDeng, line 40: We should use `netip.Prefix` if possile.

modules/caddyhttp/proxyprotocol/listenerwrapper.go
```diff
@@ -38,32 +37,69 @@ type ListenerWrapper struct {
...
+		}
+		pp.deny = append(pp.deny, ipnet)
+	}
+	pp.policy = func(upstream net.Addr) (goproxy.Policy, error) {
+		ret := pp.FallbackPolicy
+		host, _, err := net.SplitHostPort(upstream.String())
+		if err != nil {
+			return goproxy.REJECT, err
+		}
+		ip := net.ParseIP(host)
```
WeidiDeng, line 77: We should use `netip.Addr` here.

WeidiDeng (commented):
Nothing major except for the possible unix socket bug. I prefer to use the new `netip` types if possible.

modules/caddyhttp/proxyprotocol/policy.go
mohammed90: Because then we'd have to manually trim the quotes around the values. See this example of xText: https://go.dev/play/p/u_DS6ECr9pu. The same example when using xJSON: https://go.dev/play/p/7lAjXV66Py_V. With xJSON, the values always come with the quote marks around them, which is awkward.

modules/caddyhttp/proxyprotocol/listenerwrapper.go
mohammed90: Done

modules/caddyhttp/proxyprotocol/listenerwrapper.go
mohammed90: Done

modules/caddyhttp/proxyprotocol/listenerwrapper.go
mohammed90: Done?

WeidiDeng:
LGTM

WeidiDeng (approved):

modules/caddyhttp/proxyprotocol/policy.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type Policy int
+
+// as defined in: https://pkg.go.dev/github.com/pires/go-proxyproto@v0.7.0#Policy
+const (
+	// IGNORE address from PROXY header, but accept connection
+	PolicyIGNORE Policy = iota
+	// USE address from PROXY header
+	PolicyUSE
```
TimWolla, line 18: I don't do Caddy and I only do the bare minimum of Go. This came to my attention via haproxy/haproxy#2450 and if I understand the code correctly, the `USE` policy is a footgun will result in security vulnerabilities, as per my comment on the HAProxy issue, because my understanding is that it uses the PROXY header if one is sent and does not use it, if it is missing, allowing a malicious client to add the PROXY header itself if a gateway in front of Caddy is trusted, but does not add the PROXY header for whatever reason.

TimWolla: I also find `REJECT` somewhat questionable, because it effectively is `SKIP`, but worse: It would reject a connection if the first few bytes of the connection happen to match a valid PROXY header by chance, instead of just passing it to the underlying protocol handler.

francislavoie: If the user knows their Caddy instance is accessible on public networks, they should configure `allow` to limit PROXY protocol to only those IP ranges.

TimWolla: > If the user knows their Caddy instance is accessible on public networks, they should configure allow to limit PROXY protocol to only those IP ranges.

My understanding is that the `allow` configuration option sets the `USE` policy for a given client as per

https://github.com/caddyserver/caddy/pull/5915/files#diff-39cd21a1e121836c0740aa16ed05549fb146c932b9c0f93f5409a6ccb902c0a9R92-R97

Now consider the following:

1. I configure `allow 10.10.10.10` to preserve IP addresses from my trusted gateway running at `10.10.10.10`.
2. This configures the `USE` policy, which to my understanding takes the client information from the PROXY header if present and accepts the connection without doing anything special if the PROXY header is not present.
3. My trusted gateway at `10.10.10.10` does not add the PROXY header for whatever reason (bug, configuration mistake, ...).
4. Now a malicious client could connect to my trusted gateway at `10.10.10.10` and prepend a valid PROXY header to their request. This PROXY header would be forwarded as-is to Caddy and then accepted.

If instead the `REQUIRE` policy would be configured, I would notice at step (3), because all connections from my trusted gateway would be rejected as the PROXY header is missing, thus alerting my monitoring.

francislavoie: Hmm. So the concern is that the "trusted" gateway somehow becomes not trusted? Sounds like it's already game-over at that point. It would be very wrong for a gateway to pass through PROXY protocol bytes as-is without validation/stripping, if it is itself configured to add PROXY protocol.

The problem with `REQUIRE` is that it's a common for internal apps (from private IP ranges) to need to connect to the server directly (e.g. a cron job which uses the HTTP server to publish an event of somekind, or a metrics scraper pulling data), and those apps may not support adding PROXY protocol; and for those usecases the connecting IP is not relevant anyway.

polarathene: @TimWolla Apologies that this turned out so verbose, but I wanted to be thorough 😓 

- I've provided a reproduction environment that you are welcome to show an exploit for due to `go-proxyproto` policy  feature allowing to accept the PROXY protocol header to replace the Client IP only when the host is trusted, but otherwise allowing connections through including from the trusted host.
- Traefik without any handling of PROXY protocol or trust can route non-TLS connections via TCP router to Caddy, which can trust Traefik as a host IP for PROXY protocol. This is the only situation where the client could forge their IP via PROXY protocol.
- I have written a program to proxy a TCP connection and inject a PROXY protocol header with fake IP to try exploit, but no success beyond the above generic TCP connection router unaware of PROXY protocol.

---

> as per my comment on the HAProxy issue

> allowing a malicious client to add the PROXY header itself if a gateway in front of Caddy is trusted, but does not add the PROXY header for whatever reason.

For others in this discussion to be aware of, [I have provided a `compose.yaml` environment](https://github.com/haproxy/haproxy/issues/2450#issuecomment-1942940452) that can be used to easily setup Caddy with a trusted host (Traefik).

While @TimWolla may not be familiar with Caddy, hopefully it is sufficient for them to demonstrate how a client could exploit the trust policy feature `go-proxyproto` has.

The flexibility of trust via policy by not only host IP, but also on the port (_by making PROXY header optional for a trusted host_) is quite nice to have vs requiring separate ports just to allow trusted hosts to preserve a client IP via PROXY protocol.

The current PROXY protocol specification also disallows normal connections to a port by untrusted IPs (_which Caddy 2.7 allowed to configure for_). `go-proxyproto` expands on that with the PROXY header being optional. Personally I don't see a concern with either of these features, so I'd love to see it exploited to clear up any misunderstanding I have.

---

> Now a malicious client could connect to my trusted gateway at `10.10.10.10` and prepend a valid PROXY header to their request. This PROXY header would be forwarded as-is to Caddy and then accepted.

This makes no sense to me. How is the gateway trusted if a malicious client can connect to it with their own PROXY header added?

The [information in the PROXY header is limited](https://seriousben.com/posts/2020-02-exploring-the-proxy-protocol/):

> image

[From the PROXY protocol spec in question](https://github.com/haproxy/haproxy/blob/2ed53ae4a02ba49f529addb9003b747c8ab281b1/doc/proxy-protocol.txt#L176-L182):

> - _The receiver **MUST** be configured to only receive the protocol described in this specification and **MUST** not try to guess whether the protocol header is present or not._
> - _This means that the protocol explicitly prevents port sharing between public and private access._
>   _Otherwise it would open a major security breach by allowing untrusted parties to spoof their connection addresses._
> - _The receiver **SHOULD** ensure proper access filtering so that only trusted proxies are allowed to use this protocol._

If you trust any client connection from `10.10.10.10` to provide PROXY protocol, then anyone can abuse that. No one disputes that, but has no relevance to optionally allowing connections to **not use PROXY protocol** to the same receiver port?

The only security concern is that when you receive PROXY protocol header, that it only be trusted if the connecting client is trusted. What's changed here with your example of a malicious client?

Is there something different to this scenario you've described compared to the `compose.yaml` example I've linked?:

```
Client => <PROXY protocol not accepted> => Traefik => <Optional PROXY protocol> => Caddy
```

Only Traefik can add the PROXY protocol header for those connections. I think your concern is where Traefik (or any other service) would unconditionally forward that PROXY header through to Caddy instead of managing that itself as the trusted host.

If that's the case, Traefik is not configured by any means as a PROXY protocol receiver, only Caddy (_which only trusts Traefik_). The fact that PROXY protocol is optional for connections to Caddy at the same port doesn't change that outcome?

<details>
<summary>Previous findings (ignore)</summary>

I have confirmed by having Caddy make a connection back to Traefik (_via the `reverse_proxy` directive_):
- If Caddy performs that request with PROXY protocol then Traefik rejects it (_unless explicitly configuring trust for incoming connections via entrypoints using PROXY protocol_).
- It could also be tested with `curl --haproxy-protocol`.

For traffic routed through Traefik (_via the TCP TLS router without PROXY protocol_) to Caddy, no forgery occurs.
- If Traefik permits a client connecting to an entrypoint to use PROXY protocol, the client IP is resolved at the entrypoint, the PROXY header is not retained to carry over to Caddy (_even with TLS passthrough or a non-TLS TCP router that just routes the TCP traffic_).
- If I enable PROXY protocol again between Traefik => Caddy, and have both curl and Caddy perform requests using PROXY protocol, the final response is the client IP belonging to Caddy. That is (_with PROXY protocol used between each request `=>`_):

   ```
   Client (curl) => Traefik => Caddy (reverse_proxy request) => Traefik => Caddy (response to reverse_proxy request)
   ```

</details>

**UPDATE:** While exploring this I realized it wasn't sufficient. I had to learn how to write a TCP proxy that injects a PROXY header to my curl requests (_tried avoiding curl but at least with rust that complicated getting the response back with my inexperience_).

I ran the program outside of the Docker subnet and connected to the exposed ports, where I'm aware of some security gotchas:
- From the Docker host, accessing the docker subnet will route the traffic with source IP as the network gateway IP. If the trusted hosts happens to be that entire subnet, that's an easy way in through trust.
- External clients over IPv6 will also be able to route through to a Docker container with published ports that only have an IPv4 network via this same gateway IP.
- Both are due to Dockers default network management with `iptables`, they can be avoided by opting out of things like `userland-proxy`.
- Those issues provide a simpler way to do such a test with `curl --haproxy-protocol --connect-to ::localhost:80 http://example.com` to provide the client IP `127.0.0.1` instead of the rewritten client IP Docker introduced before it reaches Traefik. I had forgotten about this until after I wrote the rust program 😓 

Ignoring that, when no trust is configured at the Traefik entrypoint, the client can do request forgery.
- A non-TLS TCP router will just pass the incoming traffic through to the proxied service. This allows the curl client to provide it's own PROXY header and for Caddy to accept it since Traefik IP is trusted.
- This won't effect HTTP routers that expect HTTP / TLS traffic, likewise a TCP router with TLS (passthrough or not) as the unexpected PROXY protocol header interferes.
- It also fails when Traefik configures PROXY protocol for a route to the service, as this would add a 2nd PROXY header, causing the service like Caddy which is expecting HTTP or TLS connections (with optional PROXY header) to fail as the 2nd PROXY header remains, causing issues like described for Traefik.
- If Traefik did configure an explicit set of trusted hosts (_even if it was an IP that'd never connect to Traefik_), untrusted clients trying to sneak in their own PROXY header would have the header removed and treat it like a normal connection without it (`IGNORE` policy).
- For any explicitly trusted hosts providing a PROXY header, Traefik will acknowledge the client IP from the PROXY header (`USE` policy) and if the TCP routed service enabled PROXY protocol, that will also become the client IP (_regardless of the actual IP of the connecting trusted host to Traefik_) given to the proxied service, Caddy.

Caddy in 2.7 by default (_without configuring any trusted hosts via `allow`_) would have accepted any PROXY protocol header. While with the upcoming 2.8 release with `go-proxyproto` has the same behaviour I observed with Traefik not having configured trusted hosts. They both default to `IGNORE` policy (_**EDIT:** [Traefik doesn't process the header by default](https://github.com/traefik/traefik/blob/0c8778639a3cade0cf73aa26f56f822a4fb835e8/pkg/server/server_entrypoint_tcp.go#L463-L470), so `IGNORE` is not exactly correct_).

---

> I also find `REJECT` somewhat questionable, because it effectively is `SKIP`, but worse: It would reject a connection if the first few bytes of the connection happen to match a valid PROXY header by chance, instead of just passing it to the underlying protocol handler.

Given findings above, how likely is that to be a practical concern?

In this PR, `REJECT` is only used for errors prior to having the host IP parsed, then it is checked in the allow and deny lists:
- If it's in `deny` list `REJECT` refuses the connection if it has the PROXY header 👍 
- If it's in the `allow` list `USE` does the opposite and trusts the PROXY header for client IP 👍 (_`REQUIRE` adds nothing here, since the expectation is already that the PROXY header is present, **optionally** allowing connections through without one_)
- If the IP is not in either allow or deny, `IGNORE` 👍 (_discard the header as if it weren't provided, allowing the connection_)

Traefik has a very similar policy, but they [only have an allow list for `USE`](https://github.com/traefik/traefik/blob/0c8778639a3cade0cf73aa26f56f822a4fb835e8/pkg/server/server_entrypoint_tcp.go#L439-L450), otherwise using `IGNORE` but [only if an allow list exists](https://github.com/traefik/traefik/blob/0c8778639a3cade0cf73aa26f56f822a4fb835e8/pkg/server/server_entrypoint_tcp.go#L463-L470).

If I was proxying connections via Traefik TCP router without TLS, and I knew the proxied service (Caddy) was trusting Traefik for PROXY header, I'd probably want `REJECT` as a policy for any untrusted client going through that route instead of `IGNORE`. Since a valid PROXY header by chance would be a known security issue in that scenario?

If instead there was no trusted PROXY protocol hosts to be concerned about, and the service is doing something completely unrelated, then sure `REJECT` would be bad for that.

`REQUIRE` only makes sense to enforce if you only ever expect connections to that service port to be traffic that _requires_ the client IP to be corrected. If you're concerned about misconfiguration by the PROXY header not being present, then you could monitor logs for the client IP being the trusted host IP, except when that is worse than service failure from rejecting connections, then prefer `REQUIRE`. The policy approach is just more granular than IP or necessitating a separate port configuration to support internal + external traffic to a service.

[`REJECT` rejects a connection missing the PROXY header while `SKIP` allows a connection through without a PROXY header](https://github.com/pires/go-proxyproto/blob/e5b291b295b4a6e0425a134c8b3bfa6b2e692866/policy.go#L25-L38), not sure how that's effectively the same? (_see [usage in `go-proxyproto` here](https://github.com/pires/go-proxyproto/commit/e9fdbf2d72b570c7ca61012c8a2e8072cb52e889#diff-7a50e7dfed70deb86642d34ce85399fa54d3402304c3fd31157a4552a163ad38R77-R80)_), it was [implemented for k8s to have local/internal traffic not need to go through a proxy](https://github.com/pires/go-proxyproto/issues/95) just to get the PROXY header added for the receiving service to accept the connection (_where the direct connection provides expected client IP without needing PROXY header like external traffic through a reverse proxy does_).

TimWolla: @francislavoie 

> It would be very wrong for a gateway to pass through PROXY protocol bytes as-is without validation/stripping, if it is itself configured to add PROXY protocol.

No, it would be very wrong for a gateway to touch the PROXY header - unless *its* client is also expected to send a proxy header. The gateway should just add the PROXY header and not touch anything the client sends, otherwise it would modify the data sent by the client. The target service would then parse (and strip) the (single) PROXY header added by the gateway and attempt to interpret the fake PROXY header by the client as whatever protocol the backend speaks.

That's why the `REJECT` policy is non-sense.

> The problem with `REQUIRE` is that it's a common for internal apps (from private IP ranges) to need to connect to the server directly (e.g. a cron job which uses the HTTP server to publish an event of somekind, or a metrics scraper pulling data), and those apps may not support adding PROXY protocol; and for those usecases the connecting IP is not relevant anyway.

Then the policy for those internal apps that do not send a PROXY header must be configured as `SKIP`. Everything else is semantically incorrect and/or a security vulnerability, because something that is not a PROXY header might be interpreted as one.

See also my explanation over in HAProxy's issue tracker: https://github.com/haproxy/haproxy/issues/2450#issuecomment-1944667782

TimWolla: @polarathene 

For folks following along, I've provided an extensive response over in HAProxy's tracker: https://github.com/haproxy/haproxy/issues/2450#issuecomment-1944667782

> The current PROXY protocol specification also disallows normal connections to a port by untrusted IPs (_which Caddy 2.7 allowed to configure for_). `go-proxyproto` expands on that with the PROXY header being optional. Personally I don't see a concern with either of these features, so I'd love to see it exploited to clear up any misunderstanding I have.

I believe the important part is the distinction between “optional” (i.e. `USE`; unsafe) and “no special handling” (i.e. `SKIP`; safe). Quoting from my response in HAProxy's tracker:

> Conditionally handling which hosts can connect without PROXY procotol is fine, if the decision is made without looking into the contents. 
> […]
> A trust policy on the same port is fine, if it's a decision between “client MUST send a PROXY header” and “no special handling is performed”. Making the PROXY header optional is not fine. Note the important distinction between “no special handling” and ”optional”.

> How is the gateway trusted if a malicious client can connect to it with their own PROXY header added?

see my response in HAProxy's tracker (keyword: dumb TCP forwarder).

> A non-TLS TCP router will just pass the incoming traffic through to the proxied service. This allows the curl client to provide it's own PROXY header and for Caddy to accept it since Traefik IP is trusted.

*This* is the security vulnerability of the `USE` policy. With the `REQUIRE` policy, those cases would easily be detected, because the vast majority of traffic is rejected, due to the lack of the PROXY header.

> Given findings above, how likely is that to be a practical concern? […] If it's in deny list REJECT refuses the connection if it has the PROXY header 👍

re practical concern: Depends on your type of traffic. For HTTP it is likely not a concern, because a PROXY header is not valid HTTP and thus the request would be rejected no-matter-what.

> If it's in the `allow` list `USE` does the opposite and trusts the PROXY header for client IP 👍 (_`REQUIRE` adds nothing here, since the expectation is already that the PROXY header is present, **optionally** allowing connections through without one_)

At the risk of repeating myself: `USE` is unsafe. Only `REQUIRE` is safe.

> If the IP is not in either allow or deny, `IGNORE` 👍 (_discard the header as if it weren't provided, allowing the connection_)

See `REJECT`.

mohammed90: Sorry for coming late to this discussion. The long-winded discussion is getting me lost. It's also hard to follow as it's going on multiple links covering multiple related yet separate subjects and multiple technologies.
Is the summary:
- `REQUIRE` and `SKIP` are blessed, good, and safe to keep
- all the others are recommended to be removed and considered unsafe
?

polarathene: > The long-winded discussion is getting me lost. It's also hard to follow as it's going on multiple links covering multiple related yet separate subjects and multiple technologies.

I believe I summarized the whole discussion here: https://github.com/mholt/caddy-l4/issues/176#issuecomment-1945227683

> _From what I can tell there was a miscommunication with the vulnerability concern raised._
> _At least with the present discussion, there is still no clear justification that a trust policy cannot apply to a host where the PROXY header is optional._

> _...projects that had flexible policies that conflicted with the specification (which appears acceptable to an extent from their response, thus the specification probably needs an update)._

with additional clarification in a collapsed details block:

> _In the example we've been discussing, the concern appears more to do with the trusted host (Traefik) being hands-off with PROXY protocol, such that a malicious client can provide their own PROXY protocol header that Traefik simply forwards to Caddy that trusts all connections to the Caddy server port from Traefik._

> - _**Correct fix and prevention** is to ensure Traefik does not allow untrusted clients to sneak in their own PROXY protocol header._
> - _Configuration issue of the trusted host, **nothing to do with Caddy** allowing the PROXY header to be optional for a trusted host._

I then expand a bit on policy concerns, but wrap it up with:
> _If `REQUIRE` were a default, but the user wanted **Caddy to have `USE`** for supporting connections to Caddy that don't / can't provide the PROXY header, all that's required in this **vulnerable Traefik configuration** is to ensure it uses **`IGNORE` for untrusted clients** (I will raise a bug report as their docs don't communicate the default behaviour doesn't check for a PROXY header at all)._

I don't think I got around to raising that bug report. I had been preparing one but my browser crashed.

---

This was months ago so I can't quite recall all the details and don't have time to refresh my memory 😓 (_below is based off recall of past discussion, I may slip up_)

The default with Traefik for Layer 4 traffic forwarding (_the only concern here_) is to blindly forward IIRC.

- Traefik will not apply any PROXY protocol inspection unless it's configured to do so (_which was a concern that @TimWolla expressed when the inbound traffic may on the off-chance by coincidence have something that appears to be PROXY protocol header, thus Traefik avoids making this assumption by default_).
- When you have external traffic for Traefik to forward and the original IP needs to be kept due to some other load balancer / proxy in front of Traefik, then you need PROXY protocol at this Traefik entrypoint. This is separate from Traefik's routers that _inject_ / prepend their own PROXY protocol (_again when configured to do so_), where traffic could be handed over to Caddy.

**For clarity** with the Traefik example:
- Inbound connection (client to Traefik) => (_is it a trusted proxy? apply a PROXY protocol policy_) => Route to outbound
- Outbound connection (Traefik to service, eg Caddy) => (_if configured to, add a PROXY protocol header with the IP traefik associates to the client_) => Traffic forwarded to service

It's similar for Caddy, but I use Traefik due to the Layer 4 support for both inbound and outbound, as that is where PROXY protocol applies. Layer 4 is a little bit muddy with Caddy IIRC which is presently more oriented around Layer 7 for the most part (_unless using the L4 plugin_).

---

> all the others are recommended to be removed and considered unsafe

**Disagree**, `USE` is valid.

**TL;DR** is a third-party that is misconfigured allows for an exploit of trust when you have a more relaxed policy. That's not a reason to remove support for `USE`.

No opinion on the others as I was focused on a scenario with a trusted proxy (Traefik) that didn't forward PROXY protocol headers from it's own inbound connections (_Traefik entrypoint_), ensuring that PROXY protocol header was only provided to Caddy when appropriate (_a connection that could not use a Layer 7 router_).

TimWolla: @mohammed90 

> Is the summary:
>
>    REQUIRE and SKIP are blessed, good, and safe to keep
>    all the others are recommended to be removed and considered unsafe
>    ?

- REQUIRE is safe.
- SKIP is safe.
- IGNORE is safe (but probably useless).
- USE is unsafe.
- REJECT is unsafe.

@polarathene With all due respect, you are misrepresenting the situation and I won't go through this once more with you.

TimWolla: Correction, looking at the code: IGNORE is also unsafe (it's a worse version of USE).

polarathene: **TL;DR:** `USE` can be used securely 🙄
- I just doubt anyone else cares enough to verify my demonstration of that (_and I don't blame them_),
- It's not like it has to be a default policy though. Dropping support for these extra policies seems unnecessary when they have their uses.

---

>  you are misrepresenting the situation

If I had time to go through the discussion properly I would, but I believe I gave a decent [reproduction example](https://github.com/haproxy/haproxy/issues/2450#issuecomment-1942940452) where that issue is closed with a summary and conclusion and a very verbose collapsed details block.

At a glance, here are some highlighted snippets for reference if anyone is interested:

image

image

image

image

image

---

It's easy to dismiss my input, but I have shown a scenario where `USE` is valid and useful, without the vulnerability concerns that @TimWolla raised 🤷‍♂️ :

image

That's it. That's what the big discussion was over
- @TimWolla was focused on a concern that wasn't relevant in the configured scenario, for software he doesn't use or have familiarity with configuration wise, hence the misunderstanding / miscommunication. His concern is valid when applicable, but `USE` policy wasn't a vulnerability in the configuration I was seeking clarity on.
- If you know what traffic is flowing inbound, You can choose to enforce PROXY protocol from trusted hosts, or discard / reject any connections that present a PROXY protocol header when one is not permitted.
- Keep in mind this is an example where two reverse proxies supporting PROXY protocol are in use (_Traefik handles the client connection, it has interacts with the PROXY protocol on both inbound and outbound connections, it is a trusted host to Caddy, but `USE` is valid for a trusted host as the client to Caddy when that trusted host is already sanitizing the real client connection_)
- In the case of Caddy, Traefik always adding a PROXY protocol header of it's own for Layer 4 traffic inbound to Caddy ensures that PROXY protocol is used correctly. If a 2nd header were even present (_instead of stripped_) the actual protocol negotiation such as TLS termination handled by Caddy would just fail instead.

I wrote a small program to act as a malicious client that injects it's own PROXY protocol header for the TLS connections, it could not be used to exploit a vulnerability via the scenario I argued was valid for a flexible `USE` policy. It would work if I had a misconfiguration, but as stated that is not something that should be surprising. If you set trusted hosts to allow anyone, you're vulnerable too 🤷‍♂️ 

I understand everyone is busy and my responses are frequently verbose. This topic was already concluded, the information is there for anyone that wants to spend time combing over it, I'm not interested in debating it all over again.

Issue #5899, opened 2023-10-15
Potential issue with Caddy terminating HTTP Put requests when uploaded content exceeds 1MB limit


Anduin2017:
Hi, Caddy team.

We deployed a Nextcloud server with caddy as reverse proxy.

Our configuration is:

```caddyfile
{
	email anduin@aiursoft.com
	log {
		format transform `{request>remote_ip}	{request>method}		{request>host}			{status}	{resp_headers->Server} 	in:{duration}	{request>uri}` {
            
        	}
		output file /var/log/caddy/caddy.log {
			roll_size 1gb
			roll_uncompressed
		}
		level info
	}
	servers :443 {
		metrics
		listener_wrappers {
			proxy_protocol {
				timeout 2s
				allow 127.0.0.1/24
			}
			http_redirect
			tls
		}
	}
}

# Protected by 2FA.
nextcloud.aiursoft.cn {
	#tls {
 	#   alpn http/1.1
	#}
	log
	@hacker {
		not remote_ip 10.0.0.0/10 172.16.50.0/24
		path /index.php/settings/admin*
	}
	encode gzip

	handle @hacker {
		header Content-Type text/html
		respond "<h1>You are not the admin! Don't try to admin my Nextcloud server :)</h1>" 401
	}

	reverse_proxy http://nextcloud {
	}

	redir /.well-known/carddav /remote.php/carddav 301
	redir /.well-known/caldav /remote.php/caldav 301
}

```

And we built our own caddy with script:

```bash
echo "Building caddy..."
xcaddy build  --with github.com/caddyserver/transform-encoder --output /tmp/caddy
sudo systemctl stop caddy.service
sudo cp /tmp/caddy /usr/bin/
```

Today there came an incident indicating all uploads with file larger than 1M was failed in Nextcloud.

Log from Nextcloud:

```json
{
    "reqId": "05EtXnNGDRndxVbZizqT",
    "level": 3,
    "time": "2023-10-15T17:30:17+00:00",
    "remoteAddr": "10.0.0.184",
    "user": "anduin",
    "app": "no app in context",
    "method": "PUT",
    "url": "/remote.php/dav/uploads/anduin/8929679637737091d80ac726efcff350/000001",
    "message": "Expected filesize of 19292077 bytes but read (from Nextcloud client) and wrote (to Nextcloud storage) 1040384 bytes. Could either be a network problem on the sending side or a problem writing to the storage on the server side.",
[...]

PR #5917, opened 2023-10-26
only set read limit for http redirect listener when request is http

WeidiDeng:
Fix [5899](https://github.com/caddyserver/caddy/issues/5899).


modules/caddyhttp/httpredirectlistener.go
```diff
@@ -86,39 +86,54 @@ func (l *httpRedirectListener) Accept() (net.Conn, error) {
...
+			return 0, err
 		}
 
 		// If the request doesn't look like HTTP, then it's probably
-		// TLS bytes and we don't need to do anything.
+		// TLS bytes, and we don't need to do anything.
+		// 0, nil are valid return values - caller most likely will try again.
 		if !firstBytesLookLikeHTTP(firstBytes) {
-			return
+			return 0, nil
```
francislavoie, line 120: Would it be better to `return c.r.Read(p)` here?

modules/caddyhttp/httpredirectlistener.go
```diff
@@ -136,18 +151,11 @@ func (c *httpRedirectConn) Read(p []byte) (int, error) {
 
 		err = resp.Write(c.Conn)
 		if err != nil {
-			errReturn = fmt.Errorf("couldn't write HTTP->HTTPS redirect")
-			return
+			return 0, fmt.Errorf("couldn't write HTTP->HTTPS redirect")
 		}
 
-		errReturn = fmt.Errorf("redirected HTTP request on HTTPS port")
-		c.Conn.Close()
```
francislavoie, line 144: Why remove the conn close? The connection can no longer be used because the wrong protocol was sent to it 🤔

francislavoie (commented):
Thank you!

modules/caddyhttp/httpredirectlistener.go
WeidiDeng: The connection will be closed by the caller when encountering a non-nil error. Besides, if we close the connection ourselves, we should defer the close when it is clear the connection is sending http requests several lines above.

modules/caddyhttp/httpredirectlistener.go
WeidiDeng: I thought about that, but decided to keep it as it was out of habit.

francislavoie: Would TLS perform slightly worse by having to re-enter on the first read? Probably nanoseconds at worst but :man_shrugging:

modules/caddyhttp/httpredirectlistener.go
francislavoie: Okay :man_shrugging: it's been too long since I first implemented this so I don't remember if I added `c.Conn.Close()` after testing something and noticing it's better to have, or if I just added it cause it felt like the right thing to do. I dunno. If it looks okay with `curl` and in the browser, then it should be fine I guess.

modules/caddyhttp/httpredirectlistener.go
```diff
@@ -86,39 +86,54 @@ func (l *httpRedirectListener) Accept() (net.Conn, error) {
...
 
 // Read tries to peek at the first few bytes of the request, and if we get
 // an error reading the headers, and that error was due to the bytes looking
 // like an HTTP request, then we perform a HTTP->HTTPS redirect on the same
 // port as the original connection.
 func (c *httpRedirectConn) Read(p []byte) (int, error) {
-	var errReturn error
-	c.once.Do(func() {
+	// no need to use sync.Once - net.Conn is not read from concurrently.
+	if !c.once {
```
francislavoie, line 108: Since we removed the closure, I think we could invert the condition to `if c.once { return c.r.Read(p) }`, i.e. early return, reduces indentation one level.

mohammed90:
This _seems_ (reality will tell if truly) easy to test. Do you think you can add one? I can try writing up a test case.

caddytest/integration/listener_test.go
```diff
@@ -0,0 +1,94 @@
...
+		t.Fatal("not supposed to handle a request")
+	})
+
+	// We never read the body in any way, set an extra long header instead.
+	req, _ := http.NewRequest("POST", "http://localhost:9443", nil)
+	req.Header.Set("Long-Header", strings.Repeat("X", 1024*1024))
+	_, err := tester.Client.Do(req)
+	if err == nil {
+		t.Fatal("not supposed to succeed")
+	}
```
mohammed90, line 93: The error produced by the listener-wrapper isn't what's expected.
Expected:
```
{"level":"debug","ts":1698406739.916438,"logger":"http.stdlib","msg":"http: TLS handshake error from 127.0.0.1:60526: redirected HTTP request on HTTPS port"}
```

Received:
```
{"level":"debug","ts":1698406739.916438,"logger":"http.stdlib","msg":"http: TLS handshake error from 127.0.0.1:60526: couldn't read HTTP request"}
```

Granted, those lines aren't returned to the error var here, they're printed by the logger.

mohammed90 (commented):
Thanks for the tests! One small finding.

caddytest/integration/listener_test.go
WeidiDeng: This is what's expected. Because of the extra long header, the http request can't be read fully. If we instead send a request with a large body, it will actually succeed because it gets redirected and handled.

mohammed90: Ah, gotcha. My bad.

modules/caddyhttp/httpredirectlistener.go
francislavoie: ☝️ @WeidiDeng I'm good to merge once this is answered. I don't think it's a big deal, but just a thought, since this is definitely a hot path when enabled so any work that can be avoided is a good thing.

francislavoie (approved):
Thanks!

PR #5918, opened 2023-10-27
update quic-go to v0.39.3

marten-seemann:
The patch releases contain an ECN fix on certain Linux platforms, and a fix for a bug that allowed negotiating TLS 1.2 (or older) over QUIC. As far as I can tell it doesn't have any security implications (the TLS handshake fails at a later stage), but it's just super weird.


francislavoie (approved):
Thanks!

Issue #5909, opened 2023-10-22
The name `example.Caddyfile` is not considered a valid Caddyfile


coolaj86:
```sh
caddy run --config ./example.Caddyfile
```

```text
2023/10/22 06:21:11.620	INFO	using provided configuration	{"config_file": "./example.Caddyfile", "config_adapter": ""}
Error: loading initial config: decoding request body: invalid character '#' looking for beginning of value
```

But if I simply rename the:

```sh
caddy run --config ./Caddyfile.example
```

Then it works as expected.

It seems like `.Caddyfile` (or really anything that isn't `.json`) should be considered a Caddyfile unless some sort of matching logic for some particular adapter says otherwise.

mholt:
Any file that starts with Caddyfile will automatically be assumed to be a Caddyfile. From our docs:

> If it is called something else that doesn't start with "Caddyfile", you will need to specify --adapter caddyfile.

So you can always specify the adapter name if the file name doesn't conform.

That said, I wonder if we should treat any file name that has Caddyfile in it as a Caddyfile... but maybe that's too liberal.

coolaj86:
- equals Caddyfile
- Has prefix Caddyfile.
- Has suffix .Caddyfile

OSes and editors expect to know a file by its specific name, suffix, or magic bytes.

the only other file type I know of that started out with using a prefix as its type identifier is `.env.xyz`, and its definitely more difficult to get editors to recognize it correctly.

Being consistent with decades long tried and true conventions are generally good.

mholt:
Docker does this with the Dockerfile though too. And Vagrant has the Vagrantfile. It's not something we invented nor is it uncommon. Docker is practically everywhere.

But yeah, maybe containing the word Caddyfile is good enough, since it would have to be explicitly given as the config file anyway, so I don't foresee many false positives here. We only automatically use files named exactly Caddyfile.

francislavoie:
I think `.Caddyfile` suffix is okay to support. Maybe even `.caddyfile`.

cwbriscoe:
Case insensitive options would also be good for those that are allergic to the shift key.

omalk98:
Hi @mholt and @francislavoie, I've submitted a pull request to handle this case.
Let me know if there are any other conditions you would like to handle for the Caddyfile naming convention.

PR #5919, opened 2023-10-27
Added special case support for caddyfile suffix, case insensitive

omalk98:
## Description

- Added case insensitive support for special prefix case `Caddyfile`
- Added support for `.caddyfile` suffix, also case insensitive
- Handles extension check without case sensitivity

References issue [#5909 ](https://github.com/caddyserver/caddy/issues/5909)


cmd/main.go
```diff
@@ -147,8 +147,10 @@ func loadConfigWithLogger(logger *zap.Logger, configFile, adapterName string) ([
 	// as a special case, if a config file called "Caddyfile" was
 	// specified, and no adapter is specified, assume caddyfile adapter
 	// for convenience
-	if strings.HasPrefix(filepath.Base(configFile), "Caddyfile") &&
-		filepath.Ext(configFile) != ".json" &&
+	baseConfig := strings.ToLower(filepath.Base(configFile))
+	if (strings.HasPrefix(baseConfig, "caddyfile") ||
+		strings.HasSuffix(baseConfig, "caddyfile")) &&
+		filepath.Ext(baseConfig) != ".json" &&
```
mholt, line 153: What if they have a `Caddyfile.yaml` etc? :thinking:

cmd/main.go
```diff
@@ -147,8 +147,10 @@ func loadConfigWithLogger(logger *zap.Logger, configFile, adapterName string) ([
 	// as a special case, if a config file called "Caddyfile" was
 	// specified, and no adapter is specified, assume caddyfile adapter
 	// for convenience
-	if strings.HasPrefix(filepath.Base(configFile), "Caddyfile") &&
-		filepath.Ext(configFile) != ".json" &&
+	baseConfig := strings.ToLower(filepath.Base(configFile))
+	if (strings.HasPrefix(baseConfig, "caddyfile") ||
+		strings.HasSuffix(baseConfig, "caddyfile")) &&
```
mholt, line 152:
```suggestion
		strings.HasSuffix(baseConfig, ".caddyfile")) &&
```

Probably want to make it a file extension?

mholt (changes requested):
Thanks, sorry for the delay (we had a baby unexpectedly early around that time).

I think I like where this is going but it could use a little refinement. Let me know your thoughts!

cmd/main.go
omalk98: Good point, it slipped my mind.
A possible solution is to have a function from the `caddyconfig` package which checks if an extension matches a registered adapter name.

omalk98:
I think this can open up an interesting path for automation. What if the registered modules provide a static list of file extensions in its interface that they are able to handle? Like `yaml` and `yml`, `mysql` and `sql`, etc. 
This could allow bypassing the adapter flag when a known file extension is used. 
What are your thoughts on this? 
I'm willing to open up some pull requests in the adapter modules to support this.

All the best with your new baby!

mholt:
@omalk98 Whew, finally starting to go through my backlog. Anyway, that is interesting; I guess each adapter would just have to add a line of code to register its known file extension(s). That sounds like a good idea though, should help reduce frustrations if forgetting to use the `--adapter` flag. I'd be open to this!

mholt (approved):

mholt:
@omalk98 I fixed a panic when the config file has no extension (slice access of `[1:0]`) in [e1f4b83](https://github.com/caddyserver/caddy/commit/e1f4b83ffaa861e8f383ba52108e6449b53217f2).

PR #5923, opened 2023-10-31
Align with the API changes in certmagic to pass the context into the decision func

ankon:
See https://github.com/caddyserver/certmagic/pull/255


mholt (approved):
Aha, you beat me to it! Thanks.

francislavoie:
Thanks! I think we should maybe wait @mholt to tag a new certmagic release first, but looks good :+1:

ankon:
> I think we should maybe wait @mholt to tag a new certmagic release first, but looks good 👍

Yeah, absolutely. I just needed it anyways for actually using it in our world, and the step from "push to branch" to "make draft PR" wasn't that big :)

mholt:
I'll tag a new version of CertMagic very soon. :tm:

PR #5934, opened 2023-11-09
Avoid skipping a storage due to identical string representations

ankon:
fmt.Sprintf("%v", thing) will use a fmt.Stringer implementation in thing, which means that we cannot easily assume that this value is going to be unique here and can be used to deduplicate the storages.

Instead: Use a `map[interface{}]struct{}` as we anyways have an interface, and interfaces are perfectly fine map keys.

As a side-effect: Implementors of storage can now implement fmt.Stringer safely, and get the expected output here.

### DIFF
--- a/modules/caddytls/tls.go
+++ b/modules/caddytls/tls.go
@@ -570,14 +570,13 @@ func (t *TLS) cleanStorageUnits() {
 	}
 
 	// avoid cleaning same storage more than once per cleaning cycle
-	storagesCleaned := make(map[string]struct{})
+	storagesCleaned := make(map[interface{}]struct{})
 
 	// start with the default/global storage
 	storage := t.ctx.Storage()
-	storageStr := fmt.Sprintf("%v", storage)
-	t.logger.Info("cleaning storage unit", zap.String("description", storageStr))
+	t.logger.Info("cleaning storage unit", zap.Any("description", storage))
 	certmagic.CleanStorage(t.ctx, storage, options)
-	storagesCleaned[storageStr] = struct{}{}
+	storagesCleaned[storage] = struct{}{}
 
 	// then clean each storage defined in ACME automation policies
 	if t.Automation != nil {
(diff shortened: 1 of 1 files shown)


ankon:
Quick playground: https://go.dev/play/p/UgWKHYRn17H

modules/caddytls/tls.go
```diff
@@ -570,28 +570,26 @@ func (t *TLS) cleanStorageUnits() {
 	}
 
 	// avoid cleaning same storage more than once per cleaning cycle
-	storagesCleaned := make(map[string]struct{})
+	storagesCleaned := make(map[interface{}]struct{})
```
mholt, line 573:
```suggestion
	storagesCleaned := make(map[certmagic.Storage]struct{})
```

mholt (changes requested):
Thanks!

Hmm, yes, this is probably better; although, I wonder if this is _too_ strict of a comparison. It uses the pointer to the value as the key, so if we have 2 identical storage configurations with different pointers (like they are configured in different automation policies) then it won't deduplicate properly.

Maybe that is fine. It's probably better to clean twice than to skip cleaning entirely. (Unless the storage mechanism is expensive, like a cloud DB or something.)

My original hope was that the `%v` would be able to capture a representation of the storage _config_ in a general way.

Anyway, I think overall, this change is probably good.

If you think the issue I mention above would be a problem in your case, though, let me know and we can figure something out.

ankon:
> Unless the storage mechanism is expensive, like a cloud DB or something.

Yeah. Well. :)

ankon:
> My original hope was that the %v would be able to capture a representation of the storage config in a general way.

I think that's a good goal actually, but maybe that should then be a (optional) requirement on the storage implementation: If you implement interface `StorageId`, we promise to use that to track storage identity, and you promise to give a sane value.

I'm wondering whether it could be done backwards-compatible, but I really think the foot-shooting potential of falling back to `fmt.Stringer` is a bit too high.

mholt:
As mentioned in Slack, I have a proper fix inbound, I think. Stay tuned in the next day or so for that!

mholt:
Gently closing this in favor of #5940 -- we can reopen this if needed, of course. Thanks for helping! :grinning:

PR #5939, opened 2023-11-13
templates: Offically make templates extensible

mholt:
This supercedes #4757 (and #4568) by making template extensions
configurable.

The previous implementation was never documented AFAIK and had only
1 consumer, which I'll notify as a courtesy.

### DIFF
--- a/modules/caddyhttp/templates/caddyfile.go
+++ b/modules/caddyhttp/templates/caddyfile.go
@@ -15,6 +15,9 @@
 package templates
 
 import (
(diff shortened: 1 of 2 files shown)


francislavoie:
Old way was documented: https://caddyserver.com/docs/modules/http.handlers.templates#docs

mholt:
Yes, that is -- and that part hasn't changed. All that's changed is that we now require it to show up in the config (so it can be configured), whereas it didn't before. That was a little unconventional.

PR #5944, opened 2023-11-15
logging: Add `zap.Option` support

### LINKED ISSUE #5938: How to set the default logger programatically?
I'm using Caddy as an embedded component in a service and would like to be able to set at least the default logger to the
zap logger the service has already configured by the time Caddy is started.

The primary reason for that is that setting up the logger in Caddy's config file is missing certain capabilities (primarily related to the encoder). But from management point of view, I want to avoid having to spread/duplicate logger configuration across multiple places (my service & Caddy config).

I guess the easiest place is to "extend" the `Log` function in `logging.go`. Being able to set the `defaultLogger` variable is most likely not going to work as it seems to be "recreated" in `setupNewDefault` based on its `CustomLog` settings - which lack what I need.

Or is there already a way to achieve that without any further code changes?

francislavoie:
Fix #5938

With this, loggers can be configured with `"with_caller": true` and `"with_stacktrace": "ERROR"` for example, which sets those options on the zap logger.

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -150,12 +150,18 @@ func (logging *Logging) setupNewDefault(ctx Context) error {
 		logging.Logs[DefaultLoggerName] = newDefault.CustomLog
 	}
 
-	// set up this new log
-	err := newDefault.CustomLog.provision(ctx, logging)
+	// options for the default logger
+	options, err := newDefault.CustomLog.buildOptions()
 	if err != nil {
 		return fmt.Errorf("setting up default log: %v", err)
 	}
-	newDefault.logger = zap.New(newDefault.CustomLog.core)
+
+	// set up this new log
+	err = newDefault.CustomLog.provision(ctx, logging)
+	if err != nil {
+		return fmt.Errorf("setting up default log: %v", err)
+	}
+	newDefault.logger = zap.New(newDefault.CustomLog.core, options...)
 
 	// redirect the default caddy logs
 	defaultLoggerMu.Lock()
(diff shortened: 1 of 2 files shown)


logging.go
```diff
@@ -209,14 +216,21 @@ func (logging *Logging) Logger(mod Module) *zap.Logger {
 					cores = append(cores, l.core)
 					continue
 				}
+				if len(options) == 0 {
+					newOptions, err := l.buildOptions()
+					if err != nil {
+						Log().Error("building options for logger", zap.String("module", modID), zap.Error(err))
+					}
+					options = newOptions
```
francislavoie, line 224: This is a non-fatal error log because this call chain doesn't support bubbling up an error. Probably good enough, it's very rare (only if user uses an invalid level arg for `with_stacktrace`).

modules/caddyhttp/reverseproxy/streaming.go
```diff
@@ -68,7 +68,7 @@ func (h *Handler) handleUpgradeResponse(logger *zap.Logger, rw http.ResponseWrit
 	//nolint:bodyclose
 	conn, brw, hijackErr := http.NewResponseController(rw).Hijack()
 	if errors.Is(hijackErr, http.ErrNotSupported) {
-		h.logger.Sugar().Errorf("can't switch protocols using non-Hijacker ResponseWriter type %T", rw)
+		h.logger.Error("can't switch protocols using non-Hijacker ResponseWriter", zap.String("type", fmt.Sprintf("%T", rw)))
```
francislavoie, line 71: Just a minor cleanup. I was looking for any `.Sugar()` cause I wanted to check if we needed to set `with_caller_skip` to anything higher than 0 by default, and I think it's only relevant for sugared loggers. This was the only case except for `metrics.go` which I don't care about.

mholt (approved):
LGTM -- thank you for this :)

PR #5945, opened 2023-11-17
core: Add `ctx.Slogger()` which returns an `slog` logger

francislavoie:
This should allow modules that want to use `slog` APIs instead of `zap` to do so easily. We take care of setting up the logger name using the module ID.

This requires Go 1.21 so I put it in a separate file. Once we bump our minimum we can merge it back into `context.go`.


mholt:
Oh this is good, thanks.

Since we're going to bump up to Go 1.21 soon, do we want to remove the build tag and possibly even just put this into an existing file?

francislavoie:
Sure, can wait until we merge that

francislavoie:
Okay, done @mholt

mholt (approved):
Cool, let's try it out.

PR #5946, opened 2023-11-17
caddyhttp: http2 uses new round-robin scheduler

WeidiDeng:
Stdlib begins to use the new round-robin scheduler in [the commit](https://go-review.googlesource.com/c/net/+/478735). neild also [said this was intended](https://github.com/golang/go/discussions/60746).

Bundled http2 is still using the old scheduler because a change was not made. We are not using the bundled http2, so this change is easy to make.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -378,11 +378,7 @@ func (app *App) Start() error {
 				return context.WithValue(ctx, ConnCtxKey, c)
 			},
 		}
-		h2server := &http2.Server{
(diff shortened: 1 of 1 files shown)


francislavoie:
Makes sense to me, from reading the links!

I hadn't seen that HTTP/2 to std proposal, that all looks amazing! Really hopeful we get all that soon, I'm sure it'll fix a lot of minor quirks.

mholt (approved):
Awesome -- LGTM!

WeidiDeng:
Related pr in [stdlib](https://github.com/golang/go/pull/64218), which will probably be merged.

Issue #5951, opened 2023-11-22
http/2 connection is not aborted properly if underlying connection is aborted


MidasLamb:
I have a download that is cancelled halfway through by the server while streaming a file, however the browser thinks that the file was downloaded successfully, but it only downloaded a part of it.

I've narrowed the issue down (with the help of the discord), to be http/2 related.

I've created a reproduction repository that showcases the issue: https://github.com/MidasLamb/caddy-download-abortion.

There are three ports accessible, port 3030, 3031, and 3032.
Port 3030 will use tls and http/2 directly to the application
Port 3031 will use http/1.1 to caddy that proxies the application
Port 3032 will use http/2 to caddy that proxies the application

It results in the following behaviour in the browser: 
![image](https://github.com/caddyserver/caddy/assets/14802865/ac147139-5abd-4d68-a788-231666fe92c4)
where you can see that for port 3030 and 3031, the abort is shown correctly to the user, while on port 3032, it is deemed a successful download.

<details>
  <summary>Logs</summary>

```bash
{"level":"info","ts":1700661150.1654668,"msg":"using provided configuration","config_file":"/app/Caddyfile","config_adapter":""}
{"level":"info","ts":1700661150.167787,"logger":"admin","msg":"admin endpoint started","address":"localhost:2019","enforce_origin":false,"origins":["//localhost:2019","//[::1]:2019","//127.0.0.1:2019"]}
{"level":"info","ts":1700661150.1681533,"logger":"http.auto_https","msg":"enabling automatic HTTP->HTTPS redirects","server_name":"srv1"}
{"level":"info","ts":1700661150.1684246,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc0000be480"}
{"level":"info","ts":1700661150.1709878,"logger":"tls","msg":"cleaning storage unit","description":"FileStorage:/root/.local/share/caddy"}
{"level":"info","ts":1700661150.171027,"logger":"tls","msg":"finished cleaning storage units"}
{"level":"warn","ts":1700661150.2109861,"logger":"pki.ca.local","msg":"installing root certificate (you might be prompted for password)","path":"storage:pki/authorities/local/root.crt"}
{"level":"info","ts":1700661150.2112274,"msg":"warning: \"certutil\" is not available, install \"certutil\" with \"apt install libnss3-tools\" or \"yum install nss-tools\" and try again"}
[...]

PR #5952, opened 2023-11-23
reverseproxy: panic when reading from backend failed to propagate stream error

WeidiDeng:
Fix [5951](https://github.com/caddyserver/caddy/issues/5951).

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -962,10 +962,12 @@ func (h *Handler) finalizeResponse(
 	if err != nil {
 		// we're streaming the response and we've already written headers, so
 		// there's nothing an error handler can do to recover at this point;
-		// the standard lib's proxy panics at this point, but we'll just log
-		// the error and abort the stream here
+		// we'll just log the error and abort the stream here and panic just as
+		// the standard lib's proxy to propagate the stream error.
+		// see issue https://github.com/caddyserver/caddy/issues/5951
 		logger.Error("aborting with incomplete response", zap.Error(err))
-		return nil
+		// no extra logging from stdlib
+		panic(http.ErrAbortHandler)
 	}
 
 	if len(res.Trailer) > 0 {


francislavoie:
I'm confused. Is stdlib catching/recovering from the panic? That seems really strange to use a panic for control flow like this.

WeidiDeng:
Both stdlib and http3 handle panic by aborting the stream. http3 used to have the problem that panicked handler will write 500 status instead of terminating the stream.

WeidiDeng:
The stream has to be terminated. When we just return here, there are some cleanup to be done by stdlib which make the stream succeed without any problem.

WeidiDeng:
@francislavoie stdlib does catch and recover panic. That special variable just stops logging.

mholt:
Oh interesting -- so the stdlib is _expecting_ a panic to properly clean up the stream that is terminated early. Otherwise it thinks it is successful. Bizarre, but yeah, let's see from the OP if it works!

francislavoie (approved):
Confirmed by reporter. LGTM!

jcchavezs:
We run a nightly e2e test for a caddy plugin and ever since this PR got merged it started failing (see https://github.com/corazawaf/coraza-caddy/actions/runs/7002992650/job/19047820525#step:5:1) with the error:

`
{"level":"error","ts":1701077125.0196664,"logger":"http.handlers.reverse_proxy","msg":"aborting with incomplete response","upstream":"httpbin:8081","duration":0.000221322,"request":{"remote_ip":"172.18.0.1","remote_port":"45282","client_ip":"172.18.0.1","proto":"HTTP/1.1","method":"GET","host":"localhost:8080","uri":"/response-headers?pass=leak","headers":{"Accept-Encoding":["gzip"],"X-Forwarded-For":["172.18.0.1"],"X-Forwarded-Proto":["http"],"X-Forwarded-Host":["localhost:8080"],"User-Agent":["Go-http-client/1.1"],"Coraza-E2e":["ok"]}},"error":"short write"}
`

Any advice?

WeidiDeng:
@jcchavezs From the error, it seems that your response writer doesn't return an error and the number of bytes written is less than the data passed. I have found one part that may cause this issue.

image

According to `io.Writer` [documentation](https://pkg.go.dev/io#Writer), you must return an error here, or not actually write, just return `len(b)`. And you'll have to flush the underlying response writer as well to flush the response code.

PR #5960, opened 2023-12-01
Format markdown comments in modules.caddyhttp.App

sword-jin:
image

Format these comments.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -49,60 +49,60 @@ func init() {
 //
 // In HTTP routes, additional placeholders are available (replace any `*`):
 //
-// Placeholder | Description
-// ------------|---------------
-// `{http.request.body}` | The request body (⚠️ inefficient; use only for debugging)
-// `{http.request.cookie.*}` | HTTP request cookie
-// `{http.request.duration}` | Time up to now spent handling the request (after decoding headers from client)
-// `{http.request.duration_ms}` | Same as 'duration', but in milliseconds.
-// `{http.request.uuid}` | The request unique identifier
-// `{http.request.header.*}` | Specific request header field
-// `{http.request.host}` | The host part of the request's Host header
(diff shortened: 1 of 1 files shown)


francislavoie:
We're formatting it so it works here:

https://caddyserver.com/docs/json/apps/http/#docs

There's a risk that it might break those docs, so we'll need to check on that before merging this.

sword-jin:
> We're formatting it so it works here:
> 
> https://caddyserver.com/docs/json/apps/http/#docs
> 
> There's a risk that it might break those docs, so we'll need to check on that before merging this.

Thank you, maybe we need to fix it if it's not compatible, I think some parts of developers are used to read docs on godoc.

francislavoie:
Yeah definitely, I agree. Just want to make sure it looks right for both, if possible. But we prioritize our docs first.

mholt (approved):
Woah. Why does that help? I'm not sure I understand 😇

sword-jin:
By the way, how to compile https://github.com/caddyserver/moduledoc, I didn't find any docs about it. I want to compile on my local environment so I can check it's comparable with our doc system

sword-jin:
image
Like that
@mholt

mholt:
Thanks, that does look a lot better on godocs; it also confirms my suspicions that I think it will cause the docs on our Caddy website to look like that -- one big code block with literal markdown in it. I don't know a way to make it look nice in both places. We might have to choose.

I think most people are reading the docs for templates on our website (which renders Markdown). What's your use case for reading it on the godoc instead of on our website?

sword-jin:
@mholt 
I am used to go though all the API and types on the go doc before I jump into the code. I think maybe other developers have same habit like me,

mholt:
That's interesting.

I wish we could make both look good, but unfortunately I don't know a way to do that.

I think our first priority needs to make sure that our official docs on our website are legible, especially given that the table shown there isn't intended for Go programmers as much as Caddy users, which likely won't be in the godoc, I'm afraid.

Thanks for the thoughtful change, I just wish we could find a way that wouldn't break our current docs!

PR #5961, opened 2023-12-02
chore: enabling a few more linters

AnomalRoil:
I was looking at Caddy and noticed lots of linters are disabled. So I've decided to take a stab at increasing their number.

I'll do another PR or two to try and enable as many as possible (as long as they don't require changing too much code!)
So, here is a first, hopefully minimal PR for a ~15 new linters.

If you feel like having separate commits with the description of each linter isn't required, I'm happy to do bigger commits in my follow up PR, let me know.


logging.go
```diff
@@ -470,7 +470,7 @@ func (cl *CustomLog) loggerAllowed(name string, isModule bool) bool {
 	// append a dot so that partial names don't match
 	// (i.e. we don't want "foo.b" to match "foo.bar"); we
 	// will also have to append a dot when we do HasPrefix
-	// below to compensate for when when namespaces are equal
```
mholt, line 473: It's amazing how often I do this :upside_down_face:

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -1446,7 +1446,8 @@ func (c ignoreClientGoneContext) Err() error {
 // from the proxy handler.
 const proxyHandleResponseContextCtxKey caddy.CtxKey = "reverse_proxy_handle_response_context"
 
-var noUpstreamsAvailable = fmt.Errorf("no upstreams available")
+// ErrNoUpstream occurs when there are no upstream available.
+var ErrNoUpstream = fmt.Errorf("no upstreams available")
```
mholt, line 1450: Why does this have to be exported?

mholt (commented):
Thanks! Normally I find linters annoying and too opinionated, but these haven't produced nearly any changes to the code and they're all improvements. I just have a question about the exporting of an error variable, that's all.

modules/caddyhttp/reverseproxy/reverseproxy.go
AnomalRoil: That allows users to unwrap errors using 
```
errors.Is(err, reverse.ErrNoUpstream)
```
but I might be missing something, so maybe it doesn't. 
What it does need to do is start with either `err` or `Err` to pass linting.

AnomalRoil:
Yeah, I've seen your goal was to increase code quality in the coming months/year and I think having more linters can help do that.

modules/caddyhttp/reverseproxy/reverseproxy.go
mholt:
```suggestion
// errNoUpstream occurs when there are no upstream available.
var errNoUpstream = fmt.Errorf("no upstreams available")
```

If it's OK with you I'm going to leave it unexported for now, until we have a request for this, as I don't want to expose more API surface than I have to.

If this LGTY, then I'm ready to approve and merge this PR :smiley:

francislavoie: @AnomalRoil can you adjust the above to not export this error? I think we can merge after that.

AnomalRoil: Sorry I had missed these comments in my GH notifications. Done.

AnomalRoil:
I have merged the latest master changes and have also added a few extra linters that didn't require many changes but will help ensuring code quality in the future while I was at it.

francislavoie (approved):
LGTM. I don't know what all the linters do, but I guess we're doing things right if they aren't complaining :joy:

We may or may not turn some of them off later if they cause us friction, but this is good for now I think :+1: 

Thanks!

AnomalRoil:
You're welcome. I had put the description for each linter in the commit messages in case you want to check what they do 👍🏻

mholt:
Really appreciate your contribution here @AnomalRoil ! I also find your research work valuable. Thanks for contributing!!

PR #5965, opened 2023-12-05
httpcaddyfile: Add optional status code argument to `handle_errors` directive

### LINKED ISSUE #5928: Caddyfile `handle_errors` optional status code argument
The [`handle_errors`](https://caddyserver.com/docs/caddyfile/directives/handle_errors) directive is useful for implementing custom error pages and such.

Currently, its ergonomics are not great, because matching errors by status code needs to be done via a `handle` with an `expression` matcher (see examples in the docs).

I suggest that we add an optional status code argument (or variadic multiple status codes) to `handle_errors` as a shortcut for wrapping the routes in a status code matcher. For example:

```
handle_errors 404 410 {
	<directives...>
}
```

This would effectively be the same as:

```
handle_errors {
	@codes expression `{http.error.status_code} in [404, 410]`
	handle @codes {
		<directives...>		
	}
}
```

Using `handle_errors` this way multiple times should simply append more routes. A `handle_errors` with no arguments (no matcher) should always be inserted last to ensure it acts as a fallback.

armadi1809:
Resolves #5928 

I am pretty sure there is a lot that can be improved here since this is my first PR. I particularly don't like the way I add the expression matcher to the subroutes created inside the `parseHandleErrors` and I am pretty there is a better way of doing it.

Thank you in advance for reviewing my code!

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -844,9 +844,57 @@ func parseHandle(h Helper) (caddyhttp.MiddlewareHandler, error) {
 }
 
 func parseHandleErrors(h Helper) ([]ConfigValue, error) {
-	subroute, err := ParseSegmentAsSubroute(h)
+
+	h.Next()
+	args := h.RemainingArgs()
+	expression := ""
+	if len(args) > 0 {
+		expression = ""
+		codes := []string{}
+		for _, val := range args {
+			if strings.HasSuffix(val, "xx") {
+				val = val[:1]
+				if expression != "" {
+					expression += " || "
+				}
+				expression += fmt.Sprintf("{http.error.status_code} >= %s00 && {http.error.status_code} <= %s99", val, val)
+			} else {
+				codes = append(codes, val)
+			}
+		}
+		if len(codes) > 0 {
+			if expression != "" {
+				expression += " || "
+			} else {
+				expression += "{http.error.status_code} in [" + strings.Join(codes, ", ") + "]"
+			}
+		}
+		//Reset cursor position to get ready for ParseSegmentAsSubroute
+		h.Reset()
+		h.Next()
+		h.RemainingArgs()
+		h.Prev()
+	} else {
+		//If no arguments present reset the cursor position to get ready for ParseSegmentAsSubroute
+		h.Prev()
+	}
+
+	handler, err := ParseSegmentAsSubroute(h)
 	if err != nil {
 		return nil, err
+	}
+	subroute, ok := handler.(*caddyhttp.Subroute)
+	if !ok {
+		return nil, h.Errf("segment was not parsed as a subroute")
+	}
+
+	if expression != "" {
+		statusMatcher := caddy.ModuleMap{
+			"expression": h.JSON(caddyhttp.MatchExpression{Expr: expression}),
+		}
+		for i := range subroute.Routes {
+			subroute.Routes[i].MatcherSetsRaw = []caddy.ModuleMap{statusMatcher}
+		}
 	}
 	return []ConfigValue{
 		{
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -844,7 +844,6 @@ func parseHandle(h Helper) (caddyhttp.MiddlewareHandler, error) {
 }
 
 func parseHandleErrors(h Helper) ([]ConfigValue, error) {
-
 	h.Next()
 	args := h.RemainingArgs()
 	expression := ""
@@ -865,17 +864,16 @@ func parseHandleErrors(h Helper) ([]ConfigValue, error) {
 		if len(codes) > 0 {
 			if expression != "" {
 				expression += " || "
-			} else {
-				expression += "{http.error.status_code} in [" + strings.Join(codes, ", ") + "]"
 			}
+			expression += "{http.error.status_code} in [" + strings.Join(codes, ", ") + "]"
 		}
-		//Reset cursor position to get ready for ParseSegmentAsSubroute
+		// Reset cursor position to get ready for ParseSegmentAsSubroute
 		h.Reset()
 		h.Next()
 		h.RemainingArgs()
 		h.Prev()
 	} else {
-		//If no arguments present reset the cursor position to get ready for ParseSegmentAsSubroute
+		// If no arguments present reset the cursor position to get ready for ParseSegmentAsSubroute
 		h.Prev()
 	}
 
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -851,15 +851,26 @@ func parseHandleErrors(h Helper) ([]ConfigValue, error) {
 		expression = ""
 		codes := []string{}
 		for _, val := range args {
+			if len(val) != 3 {
+				return nil, h.Errf("bad status value '%s'", val)
+			}
 			if strings.HasSuffix(val, "xx") {
 				val = val[:1]
+				_, err := strconv.Atoi(val)
+				if err != nil {
+					return nil, h.Errf("bad status value '%s': %v", val, err)
+				}
 				if expression != "" {
 					expression += " || "
 				}
 				expression += fmt.Sprintf("{http.error.status_code} >= %s00 && {http.error.status_code} <= %s99", val, val)
-			} else {
-				codes = append(codes, val)
+				continue
 			}
+			_, err := strconv.Atoi(val)
+			if err != nil {
+				return nil, h.Errf("bad status value '%s': %v", val, err)
+			}
+			codes = append(codes, val)
 		}
 		if len(codes) > 0 {
 			if expression != "" {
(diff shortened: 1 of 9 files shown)


mohammed90:
You can add tests for the Caddyfile translation to JSON. See this directory for examples:

https://github.com/caddyserver/caddy/tree/4173e2c77ab883a509ef3be1cbdc868442c5a9b8/caddytest/integration/caddyfile_adapt

You can also test the behavior by adding tests to this file:

https://github.com/caddyserver/caddy/blob/4173e2c77ab883a509ef3be1cbdc868442c5a9b8/caddytest/integration/caddyfile_test.go

armadi1809:
@mohammed90 For the attached Caddyfile, does the attahced json file looks right? This works but I am not sure if this is the expected result.
[caddyJson.txt](https://github.com/caddyserver/caddy/files/13573106/caddyJson.txt)
[Caddyfile.txt](https://github.com/caddyserver/caddy/files/13573112/Caddyfile.txt)

mohammed90:
> @mohammed90 For the attached Caddyfile, does the attahced json file looks right? This works but I am not sure if this is the expected result. [caddyJson.txt](https://github.com/caddyserver/caddy/files/13573106/caddyJson.txt) [Caddyfile.txt](https://github.com/caddyserver/caddy/files/13573112/Caddyfile.txt)

Don't make it one giant test. Break it into multiple ones that target a specific transformation behavior. You can have the giant test as an additional one to ensure the smaller parts still function properly even when merged. It also helps you with validating your understanding of the implementation as you manually do the process which you've automated.

At a glance, it seems correct, but break it apart to make the review easier and the validation more concrete and accurate.

francislavoie:
Might need a test that covers having `handle_errors` in two different site blocks. I don't remember if we were doing it before, but I think we need to wrap error routes with a `host` matcher like we do for regular routes. Maybe nobody's tried to use `handle_errors` in multiple site blocks yet :see_no_evil:

armadi1809:
Oh I think we need to update the test files for the old handle_errors am I right?

francislavoie:
This test is failing:

```
=== RUN   TestCaddyfileAdaptToJSON
  {
  	"apps": {
  		"http": {
  			"servers": {
  				"srv0": {
  					"listen": [
  						":443"
  					],
  					"routes": [
  						{
  							"match": [
  								{
  									"host": [
  										"example.com"
  									]
  								}
  							],
  							"handle": [
  								{
  									"handler": "subroute",
  									"routes": [
  										{
  											"handle": [
  												{
  													"handler": "vars",
  													"root": "/srv"
  												}
  											]
  										},
  										{
  											"handle": [
  												{
  													"error": "Unauthorized",
  													"handler": "error",
  													"status_code": 403
  												}
  											],
  											"match": [
  												{
  													"path": [
  														"/private*"
  													]
  												}
  											]
  										},
  										{
  											"handle": [
  												{
  													"error": "Not found",
  													"handler": "error",
  													"status_code": 404
  												}
  											],
  											"match": [
  												{
  													"path": [
  														"/hidden*"
  													]
  												}
  											]
  										},
  										{
  											"handle": [
  												{
  													"handler": "file_server",
  													"hide": [
  														"./Caddyfile"
  													]
  												}
  											]
  										}
  									]
  								}
  							],
  							"terminal": true
  						}
  					],
  					"errors": {
  						"routes": [
  							{
 - 								"match": [
 + 								"group": "group0",
 + 								"handle": [
  									{
 - 										"host": [
 - 											"example.com"
 - 										]
 + 										"handler": "rewrite",
 + 										"uri": "/{http.error.status_code}.html"
  									}
 - 								],
 + 								]
 + 							},
 + 							{
  								"handle": [
  									{
 - 										"handler": "subroute",
 - 										"routes": [
 - 											{
 - 												"group": "group0",
 - 												"handle": [
 - 													{
 - 														"handler": "rewrite",
 - 														"uri": "/{http.error.status_code}.html"
 - 													}
 - 												]
 - 											},
 - 											{
 - 												"handle": [
 - 													{
 - 														"handler": "file_server",
 - 														"hide": [
 - 															"./Caddyfile"
 - 														]
 - 													}
 - 												]
 - 											}
 + 										"handler": "file_server",
 + 										"hide": [
 + 											"./Caddyfile"
  										]
  									}
 - 								],
 - 								"terminal": true
 + 								]
  							}
  						]
  					}
  				}
  			}
  		}
  	}
  }
    caddyfile_adapt_test.go:52: failed to adapt error_example.txt
    caddytest.go:401: warning: heredoc.txt:2: : Caddyfile input is not formatted; run 'caddy fmt --overwrite' to fix inconsistencies
    caddytest.go:401: warning: map_and_vars_with_raw_types.txt:3: : Caddyfile input is not formatted; run 'caddy fmt --overwrite' to fix inconsistencies
```

And you'll need to run `gofumpt -w caddyconfig/httpcaddyfile/builtins.go` to make the linter happy I think.

armadi1809:
@mohammed90 I broke down the test into 4 smaller pieces: 
1- A test with just the status code 
2- A test with the range using the xx suffix 
3- A test to make sure the sorting works when there is a handle_error directive without a status code 
4- A test combining ranges and simple status codes together. 

I also made sure the test that was failing in the pipelines is now passing. 

Please let me know if there is anything else I should do for you to be able to review my code. Thank you!

armadi1809:
@francislavoie  By multiple site blocks do you mean something like the attached Caddyfile? If so, that is working properly with the current code. Let me know if I am not understanding your request correctly. 
[Caddyfile.txt](https://github.com/caddyserver/caddy/files/13621735/Caddyfile.txt)

francislavoie:
Not quite, I mean 2 sites with the different hosts but the same port. So like:

```
foo.localhost {
	...
}

bar.localhost {
	...
}
```

This will use the same `server` in JSON config with their routes being separated by a `host` matcher.

armadi1809:
@francislavoie, Okay I now understand what you mean. But how would you think this should look like in the context of handling errors. Should the routes have 2 matchers? A host matcher and an expression matcher? Is that possible?

francislavoie:
It should work the same as the non-error routes. Each site produces a subroute with a host matcher, then all the error handlers for that site go within that matcher. Nested subroutes.

armadi1809:
@francislavoie. I think it's all good now. I added a test caddyfile to account for the multiple site blocks case. I also adjusted other tests to include the host matcher in them.

armadi1809:
Hi @francislavoie, I thought I just wanted to bring your attention to this PR again. Your feedback on the latest changes would be appreciated if you still see issues with the changes suggested here. Thank you!

francislavoie:
Yep, holidays and things :joy: thanks for the reminder, doing a round of review now.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -799,10 +799,56 @@ func parseHandle(h Helper) (caddyhttp.MiddlewareHandler, error) {
...
 func parseHandleErrors(h Helper) ([]ConfigValue, error) {
-	subroute, err := ParseSegmentAsSubroute(h)
+	h.Next()
+	args := h.RemainingArgs()
+	expression := ""
+	if len(args) > 0 {
+		expression = ""
+		codes := []string{}
+		for _, val := range args {
+			if strings.HasSuffix(val, "xx") {
```
francislavoie, line 809: I think we need a bit more validation here. What if the user gives something like `xxx`? We need to validate that the length is 3 and that the first char is actually a digit.

Elsewhere in the codebase, we have code like this to validate:

```go
					if len(arg) == 3 && strings.HasSuffix(arg, "xx") {
						arg = arg[:1]
					}
					statusNum, err := strconv.Atoi(arg)
					if err != nil {
						return d.Errf("bad status value '%s': %v", arg, err)
					}
```

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -799,10 +799,56 @@ func parseHandle(h Helper) (caddyhttp.MiddlewareHandler, error) {
...
+				val = val[:1]
+				if expression != "" {
+					expression += " || "
+				}
+				expression += fmt.Sprintf("{http.error.status_code} >= %s00 && {http.error.status_code} <= %s99", val, val)
+			} else {
+				codes = append(codes, val)
+			}
+		}
+		if len(codes) > 0 {
```
francislavoie, line 819: Similarly, here we should ensure it's a 3-digit number as well.

francislavoie (changes requested):
Looks great! I made a quick commit to remove leftovers from the first commit that isn't needed anymore. This also needs a bit more validation logic to ensure bad inputs are rejected. Almost ready!

armadi1809:
@francislavoie thank you for your feedback. The requested changes should now be implemented.

francislavoie:
I don't see any changes since I last reviewed 🤔 did you forget to push?

armadi1809:
Strange 😅.  This is a screenshot of what I am seeing. The force push should contain the changes. 

Screenshot 2024-01-15 at 10 37 59 PM

francislavoie:
I'm not seeing any added validation code 🤔

francislavoie (approved):
That's better, thank you!

caddyconfig/httpcaddyfile/directives.go
```diff
@@ -70,6 +70,7 @@ var directiveOrder = []string{
 	"handle",
 	"handle_path",
 	"route",
+	"handle_errors",
```
francislavoie, line 73: Ah, remove this line please.

armadi1809: Just did. I forgot to pull changes from the remote branch after your commit. Thanks!

Issue #5970, opened 2023-12-11
AIX build is broken


bmarwell:
Hi,

I just noticed the env vars `GOOS=aix GOARCH=ppc64` will lead to a build error:

```
$ GOOS=aix GOARCH=ppc64 go build
# github.com/dgraph-io/badger
../../../../go/pkg/mod/github.com/dgraph-io/badger@v1.6.2/dir_unix.go:63:13: undefined: unix.Flock
# github.com/dgraph-io/badger/v2
../../../../go/pkg/mod/github.com/dgraph-io/badger/v2@v2.2007.4/dir_unix.go:62:13: undefined: unix.Flock
# github.com/google/certificate-transparency-go/x509
../../../../go/pkg/mod/github.com/google/certificate-transparency-go@v1.1.6/x509/cert_pool.go:65:9: undefined: loadSystemRoots
../../../../go/pkg/mod/github.com/google/certificate-transparency-go@v1.1.6/x509/root.go:21:32: undefined: loadSystemRoots
../../../../go/pkg/mod/github.com/google/certificate-transparency-go@v1.1.6/x509/verify.go:759:12: c.systemVerify undefined (type *Certificate has no field or method systemVerify)
```

I think this should have worked out of the box?  => https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63

I found no infos about AIX not being supported. Any help is appreciated.

bmarwell:
Just for reference, other no-so-common platforms work fine, too.

```
~/git/caddy-2.7.6-src/caddy ⌚ 10:25:19
$ GOOS=linux GOARCH=ppc64le go build

~/git/caddy-2.7.6-src/caddy ⌚ 10:27:32
$ file caddy
caddy: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, OpenPOWER ELF V2 ABI, version 1 (SYSV), statically linked, Go BuildID=G8uALCkYZAyhWc4CyD6c/zliuk-1icIxIIpa3eOMl/vI3N5kg7rx6eOf2RfoYl/on9sJjH1QcQMdAMjc8O0, with debug_info, not stripped

~/git/caddy-2.7.6-src/caddy ⌚ 10:27:34
$ GOOS=linux GOARCH=s390x go build

~/git/caddy-2.7.6-src/caddy ⌚ 10:27:53
$ file caddy
[...]

PR #5971, opened 2023-12-11
[#5970] cross-build for AIX

bmarwell:
Helps #5970 (checks).

AIX is only available for ppc64 platforms.


francislavoie (approved):
Thanks, makes sense. At least gives us a signal whether it compiles or not (it doesn't, obviously).

PR #5973, opened 2023-12-12
fileserver: Browse symlink display shows the target

### LINKED ISSUE #5810: Enhancement: file-server browser, allow for listing symlinks as symlinks
Caddy currently resolves symlink targets and displays the target information in its directory listing.

I would like an option to show links as links to the destination instead of appearing as a real file.

Example:
```
a/
b/
foo.txt
bar.txt
baz.txt -> a/baz.txt
...
```
Clicking the link should still download the file as before.

armadi1809:
Resolves #5810 

I tested this change with both the `file_server` directive as well as the `caddy file-server` command and it's working for both. Please let me know if I missed something.


armadi1809:
Just ran `gofumpt -w` on the changed file. The linting issue should be resolved.

mholt:
Thanks for the PR! My only thought is that maybe dereferencing a symlink should be optional. Some people may not want to show the linked path information. How would you feel about implementing a simple config setting for this?

armadi1809:
@mholt sure thing! One question to make sure I understand this correctly. The config setting will be part of the browse sub-directive  syntax? If not, Is there another place in the codebase where we are doing something similar that I can take inspiration from? Thank you for taking the time to review this!

mholt:
Yeah, it'd likely have to be a subdirective of the `browse` subdirective. :sweat_smile: Maybe something like `reveal_symlinks`?

armadi1809:
Sounds good. Thanks for the clarification.

steffenbusch:
Thank you for making this an opt-in. I use the browse template to show certain users/clients what files they can download and some of them are sym-linked to a central location and I would not like to reveal this internal thing.

modules/caddyhttp/fileserver/browse.go
```diff
@@ -48,7 +48,8 @@ var BrowseTemplate string
 // Browse configures directory browsing.
 type Browse struct {
 	// Filename of the template to use instead of the embedded browse template.
-	TemplateFile string `json:"template_file,omitempty"`
+	TemplateFile  string `json:"template_file,omitempty"`
+	RevealSymLink bool   `json:"reveal_symlink,omitempty"`
```
francislavoie, line 52: Please add a godoc comment; we parse godoc comments to generate our docs https://caddyserver.com/docs/json/apps/http/servers/routes/handle/file_server/browse/

modules/caddyhttp/fileserver/caddyfile.go
```diff
@@ -58,6 +58,12 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 				return nil, h.ArgErr()
 			}
 			fsrv.Browse = new(Browse)
+		case 2:
+			if args[0] != "browse" || args[1] != "reveal_symlink" {
+				return nil, h.ArgErr()
+			}
+			fsrv.Browse = new(Browse)
+			fsrv.Browse.RevealSymLink = true
```
francislavoie, line 66: I think this option shouldn't be allowed inline, that's awkward. Instead, it should be a subdirective (like `browse` already is, taking a template file as an argument). Look at line 109.

modules/caddyhttp/fileserver/browse.go
armadi1809: This should now be good to go.

modules/caddyhttp/fileserver/caddyfile.go
armadi1809: I just submitted an update to this. I am not sure if what I did is the best way to go so waiting on your feedback before marking this as resolved. Thanks again for reviewing.

francislavoie: I think you should just add another `case "reveal_symlink"`, don't try to complicate `browse` itself.

armadi1809: Okay, I think I finally got it  😅 . I didn't get that you did not want it to be an argument to browse.

modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -74,12 +75,18 @@ func (fsrv *FileServer) directoryListing(ctx context.Context, entries []fs.DirEn
...
 		if fileIsSymlink {
 			path := caddyhttp.SanitizedPathJoin(root, path.Join(urlPath, info.Name()))
 			fileInfo, err := fs.Stat(fsrv.fileSystem, path)
 			if err == nil {
 				size = fileInfo.Size()
 			}
+
+			symLinkTarget, err := filepath.EvalSymlinks(path)
+			if err == nil {
+				symLinkRepresentation = name + " -> " + symLinkTarget
```
mholt, line 88: Actually, how about if we use a Tabler SVG icon for this arrow?

```
<svg xmlns="http://www.w3.org/2000/svg" class="icon icon-tabler icon-tabler-arrow-narrow-right" width="24" height="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round"><path stroke="none" d="M0 0h24v24H0z" fill="none"/><path d="M5 12l14 0" /><path d="M15 16l4 -4" /><path d="M15 8l4 4" /></svg>
```

modules/caddyhttp/fileserver/caddyfile.go
```diff
@@ -107,6 +107,12 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 				fsrv.Browse = new(Browse)
 				h.Args(&fsrv.Browse.TemplateFile)
 
+			case "reveal_symlink":
+				if fsrv.Browse == nil {
+					return nil, h.Err("reveal_symlink subdirectove requires browsing to be enabled")
+				}
+				fsrv.Browse.RevealSymlink = true
+
```
mholt, line 115: I actually think this should be a subdirective of `browse` -- like within a block. Since it can't be used without it. Then it also has structure parity with the JSON config.

modules/caddyhttp/fileserver/command.go
```diff
@@ -151,6 +152,7 @@ func cmdFileServer(fs caddycmd.Flags) (int, error) {
 
 	if browse {
 		handler.Browse = new(Browse)
+		handler.Browse.RevealSymlink = revealSymlink
```
mholt, line 155:
```suggestion
		handler.Browse = &Browse{RevealSymlink: revealSymlink}
```

modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -74,12 +75,18 @@ func (fsrv *FileServer) directoryListing(ctx context.Context, entries []fs.DirEn
...
 		fileIsSymlink := isSymlink(info)
+		symLinkRepresentation := ""
 		if fileIsSymlink {
 			path := caddyhttp.SanitizedPathJoin(root, path.Join(urlPath, info.Name()))
 			fileInfo, err := fs.Stat(fsrv.fileSystem, path)
 			if err == nil {
 				size = fileInfo.Size()
 			}
+
+			symLinkTarget, err := filepath.EvalSymlinks(path)
```
mholt, line 86: Is it possible to only do this if RevealSymlinks is enabled?

modules/caddyhttp/fileserver/browse.go
```diff
@@ -49,6 +49,8 @@ var BrowseTemplate string
 type Browse struct {
 	// Filename of the template to use instead of the embedded browse template.
 	TemplateFile string `json:"template_file,omitempty"`
+	// Determines whether or not targets of symlinks should be revealed.
+	RevealSymlink bool `json:"reveal_symlink,omitempty"`
```
mholt, line 53:
```suggestion
	RevealSymlinks bool `json:"reveal_symlinks,omitempty"`
```

Wondering if plural makes more sense :thinking:

mholt (changes requested):
Thanks for working on this! I'm looking forward to merging this, I just had a few last-minute thoughts before we do.

modules/caddyhttp/fileserver/browsetplcontext.go
armadi1809: Here, do we just add the svg tag mentioned above to the `symLinkRepresentation` string or is there another/better way of doing it?

mholt: Yeah, just replace the `" -> "` string with the `" <svg ...> "` string.

You could move that into a separate variable for readability though.

modules/caddyhttp/fileserver/caddyfile.go
armadi1809: @mholt What is exactly meant by within a block a here? Should it behave the same way as `[<template_file>]` does?

modules/caddyhttp/fileserver/browsetplcontext.go
armadi1809: I was just trying this and the `" <svg ...> "` gets displayed as it is inside the html. I believe the value passed to the go template needs to be of type template.HTML in order for it to parse it as html rather than the raw string. If so, how do you suggest we treat this?

modules/caddyhttp/fileserver/caddyfile.go
mholt: So, for example:

```
file_server {
    browse {
        reveal_symlinks
    }
}
```

^ That's a block (for the `browse` subdirective)

armadi1809:
@mholt All the feedback items should be addressed now. The only one not resolved yet is the "<svg>" vs "->" and as mentioned in the comment above, I was getting the raw "<svg>..." string when inserting it directly into the symlink representation as recommended. Thanks again for your continuous help with this!

modules/caddyhttp/fileserver/browsetplcontext.go
mholt: Ah, that's probably true. In that case, the correct way to handle this would be to create a new field in the `fileInfo` struct, maybe called `Symlink` or `SymlinkName` or `SymlinkPath` or something. Then the UI will render the arrow SVG icon and the symlink info. Does that make sense?

mholt:
Thanks! Just had a sit at my desk and was able to comprehend what you meant, I just replied above.

modules/caddyhttp/fileserver/browsetplcontext.go
armadi1809: @mholt I think I got what you mean. As you can see now, I added a field called `SymlinkPath` to the `fileInfo` struct. This new field is a slice of strings that can either be empty or containing two elements (first one being the name of the symlink and the second one being the target of the symlink). This slice gets filled in only when the file is a symlink and the `reveal_symlinks` directive is enabled. Therefore, in the `browse.html` file we can check if this slice is empty and in that case display the name otherwise display the first element of the slice followed by theS SVG icon followed by the second element of the slice. Is this a valid approach?

armadi1809:
@mholt I resolved a conflict inside `./modules/caddyhttp/fileserver/caddyfile.go` that emerged after the recent modifications to that file in #5850. This should be ready to be merged again if you still want this feature. Thank you!

modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -230,13 +241,14 @@ type crumb struct {
...
-	IsDir     bool        `json:"is_dir"`
-	IsSymlink bool        `json:"is_symlink"`
+	Name        string      `json:"name"`
+	Size        int64       `json:"size"`
+	URL         string      `json:"url"`
+	ModTime     time.Time   `json:"mod_time"`
+	Mode        os.FileMode `json:"mode"`
+	IsDir       bool        `json:"is_dir"`
+	IsSymlink   bool        `json:"is_symlink"`
+	SymlinkPath []string    `json:"symlink_path"`
```
mholt, line 251: Ok I guess my last question is why this is a slice? I think it could just be a string yeah?

mholt: Isn't the first element just the Name? And the only thing we need to point to is its target (this new field)?

armadi1809: You are absolutely right. I just overcomplicated it 😅 . Should be good to go now. Thank you for your feedback Matt!

modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -230,13 +241,14 @@ type crumb struct {
...
-	IsDir     bool        `json:"is_dir"`
-	IsSymlink bool        `json:"is_symlink"`
+	Name        string      `json:"name"`
+	Size        int64       `json:"size"`
+	URL         string      `json:"url"`
+	ModTime     time.Time   `json:"mod_time"`
+	Mode        os.FileMode `json:"mode"`
+	IsDir       bool        `json:"is_dir"`
+	IsSymlink   bool        `json:"is_symlink"`
+	SymlinkPath string      `json:"symlink_path"`
```
mholt, line 251:
```suggestion
	SymlinkPath string      `json:"symlink_path,omitempty"`
```
Just a quick little change, I'll merge this in a sec

mholt (approved):
This LGTM now. Thanks so much for iterating on this with us, @armadi1809 -- hopefully people will find it useful!

PR #5980, opened 2023-12-14
logging: Automatic `wrap` default for `filter` encoder

francislavoie:
I was talking with @dunglas and he pointed out how awkward it is to have to configure `wrap` on the [`filter` encoder](https://caddyserver.com/docs/caddyfile/directives/log#filter) because it loses the nice-to-have default behaviour of checking if `stderr` is a terminal to choose a default of `console` vs `json`.

Getting this to work just right was quite messy, but I think I got it working well now. I also made some adjustments that may or may not be considered breaking changes for some people, since we're changing some default behaviour for the `filter` encoder, and I slightly changed the logic for detecting the terminal.

- we're now checking if `stderr` is a terminal instead of `stdout` because our default writer is `stderr` anyway, it was kinda incorrect to be checking `stdout` I think
- fixed a bug where the `writer` wasn't considered when determining the default encoder; this meant that if a `output file foo.log` was configured and it was run at a terminal, it would write `console` logs to the file by default (but with color turned off). That's a bad default, it should be writing `json` to the file even when run as a terminal unless otherwise stated.
- now we're never removing color by default from the log output. What we would need to do is detect whether the terminal supports color or not, but we weren't doing that before so this is certainly not a regression, if anything it's just a gap, but it's not the end of the world. I do know some Windows users have complained when they used the legacy command prompt which didn't support unix style colors, but the new windows terminal does support it now so 🤷 I don't really care to deal with that detail right now, not a big deal
- `wrap` is no longer required for `filter` encoder, it now defaults to `json` when `stderr` is not a terminal, and `console` when it is a terminal
- to make this work, we now pass down the logger's `writer` module to the configured `encoder` if it has the `ConfiguresFormatterDefault` interface, which is called immediately after provisioning as a 2nd phase in provisioning
- the `filter` encoder implements that interface, and uses it to check if the writer is a terminal, switching the nested encoder to `console`; it also passes down the writer if the explicitly configured encoder if itself wants the writer (e.g. filter inside filter, or whatever).

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -265,6 +265,17 @@ type WriterOpener interface {
 	OpenWriter() (io.WriteCloser, error)
 }
 
+// IsWriterStandardStream returns true if the input is a
+// writer-opener to a standard stream (stdout, stderr).
+func IsWriterStandardStream(wo WriterOpener) bool {
+	switch wo.(type) {
+	case StdoutWriter, StderrWriter,
+		*StdoutWriter, *StderrWriter:
+		return true
+	}
+	return false
+}
+
 type writerDestructor struct {
 	io.WriteCloser
 }
@@ -341,16 +352,19 @@ func (cl *BaseLog) provisionCommon(ctx Context, logging *Logging) error {
 			return fmt.Errorf("loading log encoder module: %v", err)
 		}
 		cl.encoder = mod.(zapcore.Encoder)
+
+		// if the encoder module needs the writer to determine
+		// the correct default to use for a nested encoder, we
+		// pass it down as a secondary provisioning step
+		if cfd, ok := mod.(ConfiguresFormatterDefault); ok {
+			err := cfd.ConfigureDefaultFormat(cl.writerOpener)
+			if err != nil {
+				return fmt.Errorf("configuring default format for encoder module: %v", err)
+			}
+		}
 	}
 	if cl.encoder == nil {
-		// only allow colorized output if this log is going to stdout or stderr
-		var colorize bool
-		switch cl.writerOpener.(type) {
-		case StdoutWriter, StderrWriter,
-			*StdoutWriter, *StderrWriter:
-			colorize = true
-		}
-		cl.encoder = newDefaultProductionLogEncoder(colorize)
+		cl.encoder = newDefaultProductionLogEncoder(cl.writerOpener)
 	}
 	cl.buildCore()
 	return nil
@@ -680,7 +694,7 @@ func newDefaultProductionLog() (*defaultCustomLog, error) {
 	if err != nil {
 		return nil, err
 	}
-	cl.encoder = newDefaultProductionLogEncoder(true)
+	cl.encoder = newDefaultProductionLogEncoder(cl.writerOpener)
 	cl.levelEnabler = zapcore.InfoLevel
 
 	cl.buildCore()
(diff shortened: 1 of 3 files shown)


logging.go
```diff
@@ -332,16 +343,19 @@ func (cl *BaseLog) provisionCommon(ctx Context, logging *Logging) error {
 			return fmt.Errorf("loading log encoder module: %v", err)
 		}
 		cl.encoder = mod.(zapcore.Encoder)
+
+		// if the encoder module needs the writer to determine
+		// the correct default to use for a nested encoder, we
+		// pass it down as a secondary provisioning step
+		if cfd, ok := mod.(ConfiguresFormatterDefault); ok {
+			err := cfd.ConfigureDefaultFormat(cl.writerOpener)
+			if err != nil {
```
dunglas, line 352: Nit, could be:

```suggestion
			if err := cfd.ConfigureDefaultFormat(cl.writerOpener); err != nil {
```

dunglas (approved):
Thank you very much Francis! This will help improving Mercure default config.

modules/logging/filterencoder.go
```diff
@@ -71,16 +78,26 @@ func (FilterEncoder) CaddyModule() caddy.ModuleInfo {
...
-	// set up wrapped encoder (required)
-	val, err := ctx.LoadModule(fe, "WrappedRaw")
-	if err != nil {
-		return fmt.Errorf("loading fallback encoder module: %v", err)
+	if fe.WrappedRaw == nil {
+		// if wrap is not specified, default to JSON
+		fe.wrapped = &JSONEncoder{}
+		if p, ok := fe.wrapped.(caddy.Provisioner); ok {
+			err := p.Provision(ctx)
+			if err != nil {
```
dunglas, line 88: Same here:

```suggestion
			if err := p.Provision(ctx); err != nil {
```

modules/logging/filterencoder.go
```diff
@@ -97,6 +114,30 @@ func (fe *FilterEncoder) Provision(ctx caddy.Context) error {
...
+			return cfd.ConfigureDefaultFormat(wo)
+		}
+		return nil
+	}
+
+	if caddy.IsWriterStandardStream(wo) && term.IsTerminal(int(os.Stderr.Fd())) {
+		fe.wrapped = &ConsoleEncoder{}
+		if p, ok := fe.wrapped.(caddy.Provisioner); ok {
+			err := p.Provision(fe.ctx)
+			if err != nil {
```
dunglas, line 133:
```suggestion
			if err := p.Provision(fe.ctx) ; err != nil {
```

mholt:
Thanks for working on this! Yuck :stuck_out_tongue_winking_eye: 

Likely to accept, but first, @dunglas Can you help me see the config difference for Mercure, before and after this change? I'm mainly curious so I can understand the impact here.

francislavoie:
Ah right, I'll add an adapt test to show the difference.

dunglas:
Currently: https://github.com/dunglas/mercure/pull/862/files

Hopefully, after, the same config in both files, automatically adapting to console or json depending on if run in TTY mode or not:

```
	log {
		format filter {
			fields {
				uri query {
					replace authorization REDACTED
				}
			}
		}
	}
```

mholt:
Gotcha, so it allows you to remove the `wrap` line it looks like

francislavoie:
Correct.

francislavoie:
Added a test, ready to go IMO.

francislavoie:
Oh lmao, @dunglas' review from last month was enough to allow it to get merged :joy: I didn't expect that a review from before he was granted collaborator status would be valid for that. Fun detail of github I guess!

mholt:
ohhh, lol, that's fun -- so reviews are retroactive in granting merge permissions. That's neat.

PR #5990, opened 2023-12-16
httpcaddyfile: Sort skip_hosts for deterministic JSON

### LINKED ISSUE #5989: Caddy Keeps Reloading Config in v2.7.5 due to skip_host order changing
Had to disable `--watch` because caddy suddenly started to continuously reload the config.

Further investigation shown this is due to the order of `skip_host` continuously changing, the first two hosts in the following code.

```
{
    "apps": {
        "http": {
            "http_port": 80,
            "https_port": 443,
            "servers": {
                "srv0": {
                    "listen": [
                        ":443"
                    ],
                    "logs": {
                        "logger_names": {
                            "*.mydomain.biz": "",
                            "mydomain.biz": "",
                            "mysite.mydomain.biz": "log0"
                        },
                        "skip_hosts": [
                            "wpad.mydomain.biz",
                            "wpad.lan.mydomain.biz",
                            "www.lan.mydomain.biz",
```

This happen in a custom image based on `caddy:2.7.5-builder-alpine` but `--with github.com/caddy-dns/cloudflare` and `--with github.com/caddyserver/ntlm-transport`.

francislavoie:
Fix #5989

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -820,6 +820,9 @@ func (st *ServerType) serversFromPairings(
 			}
 		}
 
+		// sort for deterministic JSON output
+		slices.Sort(srv.Logs.SkipHosts)
+
 		// a server cannot (natively) serve both HTTP and HTTPS at the
 		// same time, so make sure the configuration isn't in conflict
 		err := detectConflictingSchemes(srv, p.serverBlocks, options)


caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -820,6 +820,9 @@ func (st *ServerType) serversFromPairings(
 			}
 		}
 
+		// sort for deterministic JSON output
+		slices.Sort(srv.Logs.SkipHosts)
```
mohammed90, line 824: On mobile... Should it be this?

```suggestion
if srv.Logs != nil {
		slices.Sort(srv.Logs.SkipHosts)
}
```

mholt: That's probably safer, yeah

francislavoie: Yeah I made this then walked away without looking at CI, will fix in a bit

mholt (approved):
Thank you :grin:

PR #5997, opened 2023-12-20
httpcaddyfile: Fix cert file decoding to load multiple PEM in one file

### LINKED ISSUE #5981: trusted_ca_cert_file should support multiple trusted CA certs in one file
When working with trust bundles and mutual TLS, we need to rotate CA certs which means running new and old CAs in parallel for a while. This requires a Caddy to accept client TLS certs with multiple CAs.

Currently `trusted_ca_cert_file` does not allow multiple CA certs in one file, we have to have separate tooling to split CA trust bundles into separate files and generate Caddyfiles with multiple directives from templates. All our other tooling at this point is happy with everything in one file except Caddy.

Is this a conscious design / security descision with a rationale that I can't figure out or something that could be improved?

francislavoie:
Closes #5981

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -246,16 +246,26 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 						if err != nil {
 							return nil, err
 						}
-						block, _ := pem.Decode(certDataPEM)
-						if block == nil || block.Type != "CERTIFICATE" {
-							return nil, h.Errf("no CERTIFICATE pem block found in %s", h.Val())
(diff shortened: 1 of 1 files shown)


mholt (approved):
Oh wow, not sure how I missed that!! Thanks Francis :)

vaizki:
> Oh wow, not sure how I missed that!! Thanks Francis :)

I'm amazed nobody before me has brought this issue up, as these type of trust bundles are very common now.. that's why I had a hard time believing it was an actual bug. Great that it was sorted out so quickly!

francislavoie:
Ultimately, TLS client auth is very rarely used. It's not that popular of a feature. So I don't find it surprising personally :joy:

PR #6001, opened 2023-12-23
httpcaddyfile: Fix redir <to> html

subnut:
Before this commit, the response's Content-Type header was set to text/plain, which causes some browsers to treat it as plaintext instead of html, and thereby just showing the html code instead of parsing it and redirecting as expected.

This commit fixes that issue by setting Content-Type to text/html Now the browser parses the response body as HTML, which leads to the redirect working as expected.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -711,6 +711,7 @@ func parseRedir(h Helper) (caddyhttp.MiddlewareHandler, error) {
 `
 		safeTo := html.EscapeString(to)
 		body = fmt.Sprintf(metaRedir, safeTo, safeTo, safeTo, safeTo)
+		hdr = http.Header{"Content-Type": []string{"text/html; charset=utf-8"}}
 		code = "200" // don't redirect non-browser clients
 	default:
 		// Allow placeholders for the code


subnut:
Rebased on latest `master`

Can I have some feedback? Is there anything that needs to be changed? Is this ready to be merged?

francislavoie:
Sorry, this was posted right at the start of the holidays, so it slipped through the cracks.

LGTM, the change makes sense.

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -703,6 +703,8 @@ func parseRedir(h Helper) (caddyhttp.MiddlewareHandler, error) {
 	// don't redirect non-browser clients
 	if code != "200" {
 		hdr = http.Header{"Location": []string{to}}
+	} else {
+		hdr = http.Header{"Content-Type": []string{"text/html; charset=utf-8"}}
```
francislavoie, line 707: Actually, I think we could probably move this to inside the `case "html"` switch. Best to consolidate that logic in one place.

subnut: Done :+1:

subnut:
Fixed and rebased on latest `master`

francislavoie (approved):
LGTM, thanks!

PR #6010, opened 2024-01-01
caddyhttp: support unix sockets in `caddy respond` command

networkException:
Previously the `caddy respond` command would treat the argument passed to --listen as a TCP socket address, iterating over a possible port range.

This pull request factors the server creation out into a separate function, allowing this to be reused in case the listen address is a unix network address.

For example `caddy respond hi --listen "unix//tmp/test|0200"`


francislavoie (approved):
LGTM, thanks!

PR #6011, opened 2024-01-01
chore: check against errors of `io/fs` instead of `os`

mohammed90:
Documentation of `os.ErrNotExist` [says](https://pkg.go.dev/os#IsNotExist) new code should be using `errors.Is(err, fs.ErrNotExist)`. In fact, this fixes a bug where we were checking for `os.ErrNotExist` in the file-server for an error coming from the backing virtual file-system, which might not be the OS.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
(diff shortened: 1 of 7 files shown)


francislavoie (approved):
Makes sense to me

Issue #6019, opened 2024-01-04
feature request: random number placeholder


elee1766:
currently, in order to load balance canary deploys, i do something like this
```
:3838 {
	@app_canary {
		path /app*
		expression "(int({time.now.unix_ms}) % 10) >= 8"
	}

	route /* {
		route @app_canary {
		        respond "canary"
		}
		route /app* {
			respond "normal"
		}
	}
}
```

this would send ~20% of requests to the canary deployment

it would be nice if there was a {math.rand} placeholder, that would feel a little better. or maybe a env.RANDOM like some shells export

mohammed90:
I think you're having an [XY problem](https://en.wikipedia.org/wiki/XY_problem?wprov=sfla1), so your solution isn't optimal. You don't need a random number generator in the replacer. You need an [upstream selection policy](https://caddyserver.com/docs/json/apps/http/servers/routes/handle/reverse_proxy/load_balancing/selection_policy/) that respects proportioned selection.

elee1766:
> I think you're having an [XY problem](https://en.wikipedia.org/wiki/XY_problem?wprov=sfla1), so your solution isn't optimal. You don't need a random number generator in the replacer. You need an [upstream selection policy](https://caddyserver.com/docs/json/apps/http/servers/routes/handle/reverse_proxy/load_balancing/selection_policy/) that respects proportioned selection.

upstream selection policy is for reverse_proxy though

i'm not reverse proxying, i'm going between different routes

elee1766:
here's an example of the config in context.
```
{
	debug
	admin off

	filesystem analytics vfs "s3://nyc3.digitaloceanspaces.com/oku-artifacts/tags/info/v0.2.5.tar.gz" {
		header AWS_BUCKET_NAME "oku-artifacts"
		header AWS_ACCESS_KEY_ID {env.AWS_ACCESS_KEY_ID}
		header AWS_SECRET_ACCESS_KEY {env.AWS_SECRET_ACCESS_KEY}
	}
	filesystem trade vfs "s3://nyc3.digitaloceanspaces.com/oku-artifacts/tags/app/v0.2.0.tar.gz" {
		header AWS_BUCKET_NAME "oku-artifacts"
		header AWS_ACCESS_KEY_ID {env.AWS_ACCESS_KEY_ID}
		header AWS_SECRET_ACCESS_KEY {env.AWS_SECRET_ACCESS_KEY}
	}
[...]

PR #6020, opened 2024-01-04
random float64 maybe? 

elee1766:
#6019 

instead of trying to parse a number or decide on some new default, i think it makes sense to recycle the existing default of [0,1) that go provides. this is also similar to js Math.random()

### DIFF
--- a/replacer.go
+++ b/replacer.go
@@ -295,6 +295,9 @@ func ToString(val any) string {
 // returned.
 type ReplacerFunc func(key string) (any, bool)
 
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the patch!

I think, after the discussion in the linked issue, we don't yet have a common use case for this anymore, so I'm going to close this, but if there is a need for a random-number placeholder we can discuss this more (I think I'd want some more things discussed before merging something like this). Thanks for understanding!

PR #6022, opened 2024-01-06
httpcaddyfile: Support client auth verifiers

### LINKED ISSUE #6018: Add Caddyfile Config Adapter Support for Client Auth Verifiers
Hey All,

I'm using Caddy with client TLS and the caddy-revocation-validator verifier, but the Caddyfile config adapter isn't built to handle client_auth verifiers, so the source code needed some updates for it to work. Can we add support for verifiers in the Caddyfile?

Here is an example Caddyfile that I had:
```
localhost:8080 {
        respond "hello!"
        tls certificate.crt private.key {
                client_auth {
                        mode require_and_verify
                        trusted_ca_cert_file certificate.crt
                        verifier revocation {
                                mode crl_only
                                crl_config {
                                        storage_type memory
                                        signature_validation_mode verify
                                        trusted_signature_cert_file certificate.crt
                                }
                        }
                }
        }
}
```

When I run Caddy with the above Caddyfile, I get the following error:
`Error: adapting config using caddyfile: parsing caddyfile tokens for 'tls': unknown subdirective for client_auth: verifier, at Caddyfile:7`

Which comes from here: https://github.com/caddyserver/caddy/blob/8a50f191bf3e4e60fdd51753f54604735f9787e4/caddyconfig/httpcaddyfile/builtins.go#L272

In order to make it work, the following code needed to be added to the caddyconfig/httpcaddyfile/builtins.go in the client_auth select statement right above where that error is:
```
case "verifier":
						if !h.NextArg() { -> Verify that the argument we are currently on is actually another argument by checking that it is not a curly bracket or is not on a new line
							return nil, h.ArgErr()
						}

						vType := h.Val() -> Get the current argument
						modID := "tls.client_auth." + vType -> Append it to "tls.client_auth" to create the ID to use in the GetModule method to get the caddy.ModuleInfo of the module for this verifier
						unm, err := caddyfile.UnmarshalModule(h.Dispenser, modID) -> Unmarshal the passed arguments based on the valid configuration of the module
						if err != nil {
							return nil, err
						}

						ccv, ok := unm.(caddytls.ClientCertificateVerifier) -> Verify that the module is an implementation of the caddytls.ClientCertificateVerifier struct
						if !ok {
							return nil, h.Dispenser.Errf("module %s is not a caddytls.ClientCertificatVerifier", modID)
						}

						bytes, err := json.Marshal(ccv) -> Convert the configuration to JSON
						if err != nil {
							return nil, err
						}
						cp.ClientAuthentication.VerifiersRaw = append(cp.ClientAuthentication.VerifiersRaw, bytes) -> Append the bytes object of the JSON config to the slice of ClientAuthentication verifiers
```

Could we add something similar to the above so that you can use a Caddyfile rather than JSON with client_auth verifiers? Happy to submit a PR if this is something you all would be interested in!

zachgalvin:
This PR adds a case statement to the Caddyfile adapter that parses the TLS section to allow client auth verifiers. Closes #6018

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -16,6 +16,7 @@ package httpcaddyfile
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"encoding/pem"
 	"fmt"
 	"html"
@@ -217,6 +218,28 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 				for nesting := h.Nesting(); h.NextBlock(nesting); {
 					subdir := h.Val()
 					switch subdir {
+					case "verifier":
+						if !h.NextArg() {
+							return nil, h.ArgErr()
+						}
+
+						vType := h.Val()
+						modID := "tls.client_auth." + vType
+						unm, err := caddyfile.UnmarshalModule(h.Dispenser, modID)
+						if err != nil {
+							return nil, err
+						}
+
+						ccv, ok := unm.(caddytls.ClientCertificateVerifier)
+						if !ok {
+							return nil, h.Dispenser.Errf("module %s is not a caddytls.ClientCertificatVerifier", modID)
+						}
+
+						bytes, err := json.Marshal(ccv)
+						if err != nil {
+							return nil, err
+						}
+						cp.ClientAuthentication.VerifiersRaw = append(cp.ClientAuthentication.VerifiersRaw, bytes)
 					case "mode":
 						if !h.Args(&cp.ClientAuthentication.Mode) {
 							return nil, h.ArgErr()
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -16,7 +16,6 @@ package httpcaddyfile
 
 import (
 	"encoding/base64"
-	"encoding/json"
 	"encoding/pem"
 	"fmt"
 	"html"
@@ -230,16 +229,12 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 							return nil, err
 						}
 
-						ccv, ok := unm.(caddytls.ClientCertificateVerifier)
+						_, ok := unm.(caddytls.ClientCertificateVerifier)
 						if !ok {
 							return nil, h.Dispenser.Errf("module %s is not a caddytls.ClientCertificatVerifier", modID)
 						}
-
-						bytes, err := json.Marshal(ccv)
-						if err != nil {
-							return nil, err
-						}
-						cp.ClientAuthentication.VerifiersRaw = append(cp.ClientAuthentication.VerifiersRaw, bytes)
+						
+						cp.ClientAuthentication.VerifiersRaw = append(cp.ClientAuthentication.VerifiersRaw, caddyconfig.JSONModuleObject(unm, "verifier", vType, h.warnings))
 					case "mode":
 						if !h.Args(&cp.ClientAuthentication.Mode) {
 							return nil, h.ArgErr()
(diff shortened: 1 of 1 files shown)


francislavoie (approved):
Thanks, LGTM!

Please make sure to sign the CLA so that we can accept the code 👍

You may need to add the email you used to commit to your GitHub account, or edit the commit to use your GitHub account's email & force push.

zachgalvin:
Awesome! Yeah, I had to update the email on my commit. Should be good to go now though!

francislavoie:
Actually I think the JSON part needs to be like:
```go
cp.ClientAuthentication.VerifiersRaw = append(cp.ClientAuthentication.VerifiersRaw, caddyconfig.JSONModuleObject(unm, "name", vType, h.warnings)
```

zachgalvin:
How come we want to add the "name" key with the module ID as the value to the marshalled value? Is that used else where in the code? I can test that out though. I think I would remove the following right above that too since we will marshal it in the JSONModuleObject function, so no need to Marshall before that:
```
bytes, err := json.Marshal(ccv)
if err != nil {
	 return nil, err
}
```

francislavoie:
That's how Caddy can decode JSON config and know which module to decode the config into. It needs to match the namespace declared in the JSON tag on the `VerifiersRaw` struct field.

zachgalvin:
I may be misunderstanding here, but isn't all of that handled in the UnmarshalModule() method? We create a new module in that method, and then we parse the config with that module in that method, so I'm not sure why we would need the name of the module as key value pair in the map in VerifiersRaw later on. 

I also tested that out, and I got the following error:
Error: loading initial config: loading new config: loading http app module: provision http: server srv0: setting up TLS connection policies: loading client cert verifiers: position 0: loading module 'revocation': decoding module config: tls.client_auth.revocation: json: unknown field "name"

francislavoie:
The job of the Caddyfile adapter is to produce a valid JSON config, ultimately. So the module name needs to exist somewhere in the JSON config for it to be able to be loaded at runtime.

The Caddyfile adapter produces JSON, then that JSON is loaded to run. The config structs produced by Caddyfile unmarshaling are not used to run Caddy, only to produce a JSON config.

My code snippet might have been wrong, but plenty of other places that accept modules as input use this pattern. I don't have a verifier plugin on hand to test this.

francislavoie:
Ah sorry, I misread `VerifiersRaw`, it has `inline_key=verifier` so I think `"name"` -> `"verifier"`. So in the JSON, the config should look something like:

```
"verifiers": [
	{
		"verifier": "revocation",
		... the rest of that module's config
	},
	{
		"verifier": "another_imaginary_verifier_module",
		... the rest of that module's config
	}
]
```

mohammed90:
If you add `caddyfile_adapt` test under `caddytest/integration/caddyfile_adapt`, you'll have feedback on the gap between the implementation of this PR and the expected JSON.

mholt:
This change is looking good! Let me know when you're ready to merge.

zachgalvin:
Alright, can you take another look? That makes sense, so I updated it to add the verifier key with the name of the module as the value, and it worked! I also dumped the JSON config to confirm that the verifier key matched the struct tag and your example above.

francislavoie:
Linter is failing, need to run gci, apparently.

zachgalvin:
I ran the following gci command based on the failed pipelines output:
gci write caddyconfig/httpcaddyfile/builtins.go --skip-generated -s standard -s default -s "prefix(github.com/caddy
server/caddy/v2/cmd)" -s "prefix(github.com/caddyserver/caddy)" --custom-order

It removed and then added an empty line, so not sure why that was an issue, but looks like it fixed it!

mholt:
The diff was whitespace (unnecessary indentation). Looking good now!

mholt (approved):

Issue #5892, opened 2023-10-13
Disable rotation of the private key


Vcele:
I am using caddy as a reverse proxy for mailcow, and I am searching for a way to disable private key rotation on renewal of the ssl certificate so that I don't have to update my TLSA records.

mholt:
This is configurable in CertMagic, I just haven't exposed an option to do it in Caddy yet. Should be easy enough, though I generally advise against pinning keys.

Lan-Hekary:
Hello, @mholt 
I am also looking for a way to pin the private key, and/or supply the CSR to caddy.
Also I need a way to get a single certificate for multiple sub domains, not a wild card. I guess let's encrypt support that, but i's not possible inside caddy.

Key Pinning is Bad, but is some cases it's make sense to use it as long as you keep your private key secure.
Also the default setting would be to rotate it, so most of the users who does not care would not change it.

mholt:
[There is no good reason to use a multi-SAN certificate in automation](https://github.com/https-dev/docs/blob/master/acme-ops.md#use-one-name-per-certificate) other than extremely limited storage space (never heard of that), or paying per certificate. When you're using Caddy multi SAN certs are only
harmful. (But that is off topic for this issue.)

Anyway, as for private key rotation, I will think about making it configurable in Caddy, but it goes against my better judgment as, like you said, key pinning is not really a best practice.

Lan-Hekary:
Well you are right, it's a storage space issue,for me its embedded systems like IoT devices.
The resources are limited, in terms of storage, RAM, CPU and battery power are all limited.
Certificate validations takes up a good chunk of connection time.
Also connections to multiples sub-domains require either one common root CA, which increase validation time for every connection.
Or multiple Certificates, more storage and the hassle of changing the certificates every renewal.
Anyway, it's off topic like you said, but I just wanted to convay a s

There is an option to use public key, which bypass the signature validation, and it's still secure as long as the private key is secure.
And Multiple certificates can be issued with the same private key.
So that will be a good solution for the Low Power Version.

mholt:
So just to make sure I understand, you have embedded IoT devices serving thousands of domain names each? 🤔

> Certificate validations takes up a good chunk of connection time.

What do you mean by this exactly? Do you mean the low powered devices are slow at completing the TLS handshakes?

> Also connections to multiples sub-domains require either one common root CA, which increase validation time for every connection. Or multiple Certificates, more storage and the hassle of changing the certificates every renewal.

[...]

PR #6025, opened 2024-01-07
tls: add reuse_private_keys

rithvikvibhu:
### Details

This PR exposes certmagic's `ReusePrivateKeys` config to caddy. When certificates renew, by default, new keys are generated. But with this config, the same private keys (if available) can be reused.

Related to #5892 (maybe closes it if a new issue will be opened for TLSA updates as per https://github.com/caddyserver/caddy/issues/5892#issuecomment-1854172687)

**Note:** I'm fairly new to golang so may have missed something, but it seems to work with my naive testing below.

### Testing

I have a CA that (temporarily) issues self-signed certs valid for 2 mins, making it easy to test renewals. With this sample Caddyfile:

```
# you can run this with any domain, since http challenges always pass, nothing's really verified by the "CA"

https://letsdane {
    respond "Body."
    tls me+noemail@foo {
        ca https://acme.htools.work/directory
    }
}
```

run caddy, it gets cert:
```
Issued On	Sunday, 7 January 2024 at 22:15:23
Expires On	Sunday, 7 January 2024 at 22:19:23
Certificate	cebe1a2ad8ce9103985f1fb27ddf294b94776f2156af79cd824545929f444c1d
Public key	229c41815f576b45f8e24e65ef72af5eb435d075385a4cc79323bddcd51f3d3f
```

re-run caddy, it gets a cert with a different public key:
```
Issued On	Sunday, 7 January 2024 at 22:18:09
Expires On	Sunday, 7 January 2024 at 22:22:09
Certificate	51060c5ee7c54f325fc1639b0a06bf090e497c531c10e407454a4fbfb63abe22
Public key	a984b604bc63591b5bdabfe490bc5e5e1301bfa815ba8d1ed486b5468d7d2a54
```

change the config to add `ReusePrivateKeys`:
```
https://letsdane {
    respond "Body."
    tls me+noemail@foo {
        reuse_private_keys
        ca https://acme.htools.work/directory
    }
}
```

re-run caddy, it gets a cert with the same public key (see the expiry changing):
```
Issued On	Sunday, 7 January 2024 at 22:21:29
Expires On	Sunday, 7 January 2024 at 22:25:29
Certificate	f7bdb4c0c857ea09418f2d9708f575358a742af4bc384b939c413753f0a40704
Public key	a984b604bc63591b5bdabfe490bc5e5e1301bfa815ba8d1ed486b5468d7d2a54
```

### DIFF
--- a/caddyconfig/httpcaddyfile/tlsapp.go
+++ b/caddyconfig/httpcaddyfile/tlsapp.go
@@ -118,6 +118,11 @@ func (st ServerType) buildTLSApp(
 				ap.OnDemand = true
 			}
 
+			// reuse private keys tls
+			if _, ok := sblock.pile["tls.reuse_private_keys"]; ok {
+				ap.ReusePrivateKeys = true
+			}
+
 			if keyTypeVals, ok := sblock.pile["tls.key_type"]; ok {
 				ap.KeyType = keyTypeVals[0].Value.(string)
 			}
@@ -587,6 +592,7 @@ outer:
 				aps[i].MustStaple == aps[j].MustStaple &&
 				aps[i].KeyType == aps[j].KeyType &&
 				aps[i].OnDemand == aps[j].OnDemand &&
+				aps[i].ReusePrivateKeys == aps[j].ReusePrivateKeys &&
 				aps[i].RenewalWindowRatio == aps[j].RenewalWindowRatio {
 				if len(aps[i].SubjectsRaw) > 0 && len(aps[j].SubjectsRaw) == 0 {
 					// later policy (at j) has no subjects ("catch-all"), so we can
--- a/modules/caddytls/automation.go
+++ b/modules/caddytls/automation.go
@@ -138,6 +138,15 @@ type AutomationPolicy struct {
 	// load. This enables On-Demand TLS for this policy.
 	OnDemand bool `json:"on_demand,omitempty"`
 
+	// If true, private keys already existing in storage
+	// will be reused. Otherwise, a new key will be
+	// created for every new certificate to mitigate
+	// pinning and reduce the scope of key compromise.
+	// TEMPORARY: Key pinning is against industry best practices.
+	// This property will likely be removed in the future.
+	// Do not rely on it forever; watch the release notes.
+	ReusePrivateKeys bool `json:"reuse_private_keys,omitempty"`
+
 	// Disables OCSP stapling. Disabling OCSP stapling puts clients at
 	// greater risk, reduces their privacy, and usually lowers client
 	// performance. It is NOT recommended to disable this unless you
@@ -288,6 +297,7 @@ func (ap *AutomationPolicy) Provision(tlsApp *TLS) error {
 		KeySource:          keySource,
 		OnEvent:            tlsApp.onEvent,
 		OnDemand:           ond,
+		ReusePrivateKeys:   ap.ReusePrivateKeys,
 		OCSP: certmagic.OCSPConfig{
 			DisableStapling:    ap.DisableOCSPStapling,
 			ResponderOverrides: ap.OCSPOverrides,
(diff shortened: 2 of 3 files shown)


modules/caddytls/automation.go
```diff
@@ -138,6 +138,12 @@ type AutomationPolicy struct {
 	// load. This enables On-Demand TLS for this policy.
 	OnDemand bool `json:"on_demand,omitempty"`
 
+	// If true, private keys already existing in storage
+	// will be reused. Otherwise, a new key will be
+	// created for every new certificate to mitigate
+	// pinning and reduce the scope of key compromise.
```
mholt, line 144:
```suggestion
	// pinning and reduce the scope of key compromise.
	// TEMPORARY: Key pinning is against industry best practices.
	// This property will likely be removed in the future.
	// Do not rely on it forever; watch the release notes.
```

mholt (changes requested):
Thanks for this. I will likely be willing to approve this as long as we document that it's experimental or temporary and will be removed in the future.

My hope is to have the ecosystem do away with key pinning entirely, and instead have the server update TLSA records with the new key information. But that will take some time to implement.

rithvikvibhu:
Agree about key pinning, excited for whenever TLSA will be added.

For now, added the suggested note and fixed some wrong indents.

Satyamkuamr:
ISA kya hota ha

mholt (approved):
This LGTM now. Thanks!

If you have a chance, it might be good to add a Caddyfile adapt test in the caddytest area of the repo, to ensure the global option adapts to JSON as expected.

rithvikvibhu:
>If you have a chance, it might be good to add a Caddyfile adapt test in the caddytest area of the repo, to ensure the global option adapts to JSON as expected.

There's no global option added here, `reuse_private_keys` must be inside a `tls { }`. Figured since it's not recommended anyway, having it explicitly mentioned every time it's needed would be better. Should I add a global option also?

caddyconfig/httpcaddyfile/tlsapp.go
```diff
@@ -490,8 +495,9 @@ func newBaseAutomationPolicy(options map[string]any, warnings []caddyconfig.Warn
 	_, hasLocalCerts := options["local_certs"]
 	keyType, hasKeyType := options["key_type"]
 	ocspStapling, hasOCSPStapling := options["ocsp_stapling"]
+	reusePrivateKeys, hasReusePrivateKeys := options["reuse_private_keys"]
```
mholt, line 498: This is looking for a global option called `reuse_private_keys`.

Not necessarily harmful, but I bet if you remove this your `tls` block config will still work...

rithvikvibhu: Oh yeah makes sense but it currently doesn't work for global because there's no:
```go
RegisterGlobalOption("reuse_private_keys", parseOptTrue)
```

Would you like me to (add this line and adapt test) or (remove this line)?

mholt: I'm in favor of making this feature the _least_ accessible possible :wink: 

So let's not add that option, and just leave it for the `tls` directive. And we can probably remove this line (and other lines that rely on it).

francislavoie: I kinda wish we had grouped all of these under a `tls` grouping in global options. But too late for that now. I think we should add this though, for completion's sake. And anyways someone will end up complaining about it and requesting this I'm sure :joy:

mholt: It's just that I intend to remove this later. I'd really like to keep it out of the global options.

rithvikvibhu: Done, global stuff is gone.

mholt (approved):
Excellent, we'll try it out. Thank you very much!

MrIsak:
> Thanks for this. I will likely be willing to approve this as long as we document that it's experimental or temporary and will be removed in the future.
> 
> My hope is to have the ecosystem do away with key pinning entirely, and instead have the server update TLSA records with the new key information. But that will take some time to implement.

This is a wanted feature by us.
I get the downside with key-pinning, however as a CMS SaaS provider, since we do not have access to customers DNS, and no way to update their TLSA records, we need this for customers that uses DANE.

mholt:
@Lan-Hekary You can download a CI artifact, just find the latest test and download for your platform: https://github.com/caddyserver/caddy/actions/workflows/ci.yml

PR #6032, opened 2024-01-10
cmd: Adjust config load logs/errors

francislavoie:
Sometimes, users use Caddyfile file names that don't match our adapter-guessing rules, and this causes the config to be loaded as JSON. The error message is very confusing when that happens. I made some adjustments to improve it.

Before:

```
2024/01/10 19:54:15.805	INFO	using provided configuration	{"config_file": "caddyfile.test", "config_adapter": ""}
Error: loading initial config: decoding request body: invalid character 'd' looking for beginning of object key string
```

After:

```
2024/01/10 19:54:32.815	INFO	using config from file	{"file": "caddyfile.test"}
Error: config is not valid JSON: invalid character 'd' looking for beginning of object key string; did you mean to use '--adapter'?
```

Now, we're quickly doing a `json.Unmarshal` just to test that the JSON is valid (there is `json.Validate()` but it only returns a bool and not errors) before returning it to the caller after loading the config file.

Also, I moved the adapter log to be after the adapter guessing logic so that it doesn't show up empty in a confusing way in the logs. I also split the stdin and file log messages for clarity.

I'm not sure about the last bit though `; did you mean to use '--adapter'?` is there a better way we could word that without being ambiguous? :thinking: The problems is "what if they actually meant to load JSON" then a mention like this might be confusing for those users. So it's hard to go too far in either direction with the warning.


cmd/main.go
```diff
@@ -170,16 +175,24 @@ func loadConfigWithLogger(logger *zap.Logger, configFile, adapterName string) ([
...
+				zap.String("adapter", adapterName),
+				zap.String("file", warn.File),
+				zap.Int("line", warn.Line))
 		}
 		config = adaptedConfig
+	} else {
+		// validate that the config is at least valid JSON
+		err = json.Unmarshal(config, new(any))
+		if err != nil {
+			return nil, "", fmt.Errorf("config is not valid JSON: %v; did you mean to use '--adapter'?", err)
```
mholt, line 194:
```suggestion
			return nil, "", fmt.Errorf("config is not valid JSON: %v; did you mean to use a config adapter (the --adapter flag)?", err)
```

mholt (commented):
LGTM -- what about this error message?

cmd/main.go
francislavoie: Works for me :+1:

PR #6037, opened 2024-01-12
caddyfile: Normalize & flatten all unmarshalers

francislavoie:
This is something I've been noticing over the years; I was always confused as to why we were using a `for d.Next()` loop around all the Caddyfile unmarshaling code, it never seemed useful.

In practice, the only time when the loop is useful is when we intentionally want to merge multiple consecutive segments, and this only happens for request matchers. For all other module types, we never need this kind of behaviour.

So to improve code readability and make it feel less arcane, it's time to flatten all the unmarshalers to only have a single `d.Next()` line with no loop. Removes one level of indentation in almost all unmarshalers.

~~Also, I changed all `for d.NextBlock(0) {` to `for nesting := d.Nesting(); d.NextBlock(nesting); {` which avoids a hard-coded number, making better use of our provided APIs.~~ This was a bad idea after thinking about it for a few more minutes.

I strongly recommend reviewing with the whitespace collapse option turned on in github for this one.

### DIFF
--- a/caddyconfig/caddyfile/adapter.go
+++ b/caddyconfig/caddyfile/adapter.go
@@ -92,30 +92,26 @@ func FormattingDifference(filename string, body []byte) (caddyconfig.Warning, bo
 	}, true
 }
 
-// Unmarshaler is a type that can unmarshal
-// Caddyfile tokens to set itself up for a
-// JSON encoding. The goal of an unmarshaler
-// is not to set itself up for actual use,
-// but to set itself up for being marshaled
-// into JSON. Caddyfile-unmarshaled values
-// will not be used directly; they will be
-// encoded as JSON and then used from that.
-// Implementations must be able to support
-// multiple segments (instances of their
-// directive or batch of tokens); typically
-// this means wrapping all token logic in
-// a loop: `for d.Next() { ... }`.
+// Unmarshaler is a type that can unmarshal Caddyfile tokens to
+// set itself up for a JSON encoding. The goal of an unmarshaler
+// is not to set itself up for actual use, but to set itself up for
+// being marshaled into JSON. Caddyfile-unmarshaled values will not
+// be used directly; they will be encoded as JSON and then used from
+// that. Implementations _may_ be able to support multiple segments
+// (instances of their directive or batch of tokens); typically this
+// means wrapping parsing logic in a loop: `for d.Next() { ... }`.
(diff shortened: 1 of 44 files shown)


mholt:
> This is something I've been noticing over the years; I was always confused as to why we were using a for d.Next() loop around all the Caddyfile unmarshaling code, it never seemed useful.

In Caddy 1, it was not uncommon for directives to occur multiple times and all their tokens were combined and the parser handled it in one go. This pattern carried over into Caddy 2 for the flexibility it provides in case we do that.

> In practice, the only time when the loop is useful is when we intentionally want to merge multiple consecutive segments, and this only happens for request matchers. For all other module types, we never need this kind of behaviour.

That may be true. I hope we don't ever (unintentionally or otherwise) re-introduce the pattern of grouping segments then.

Even though our [docs](https://caddyserver.com/docs/extending-caddy/caddyfile#unmarshaler) demonstrate using `d.Next()` at the beginning of the Unmarshaler, it seems pointless that we include the directive name at all IMO. It made more sense to do that when there was the potential to loop over multiple instances of the directive (`for d.Next()`).

I dunno. I still think the loop is an elegant, robust way to consume tokens, that gives us more flexibility, but at the same time we don't document this pattern that I can recall...

This change is probably fine. I'm slightly sad but not for any technical reason :joy:

mholt:
Oh -- actually! We did document this:

```
// Implementations must be able to support
// multiple segments (instances of their
// directive or batch of tokens); typically
// this means wrapping all token logic in
// a loop: `for d.Next() { ... }`.
```

So now I wonder if removing it is really a good plan.

francislavoie:
No, I'm quite sure nothing uses that pattern except for matchers. It's super unintuitive and isn't something we ever mention as possible _except_ for matchers, so I see no reason plugins would have experimented to the point where they found that they could do this.

Remember that the caller to the unmarshaler must have actually passed a dispenser which includes multiple segments for it to be possible to parse multiple copies of a directive. _That never happens_, we never pass more than one directive segment to an unmarshaler. It only happens with the named matcher unmarshaler.

mholt (approved):
Alrighty, we can try this.

As mentioned in Slack, it's a bummer that the unmarshalers now have an unnecessary call to `Next()` at the beginning. (There's no point in passing in the directive name if there's no loop.) That's my bad, I guess. A carry-over from v1.

Thanks for doing this!!

PR #6040, opened 2024-01-15
caddyhttp: Fix panic when request missing ClientIPVarKey

francislavoie:
Reported in https://caddy.community/t/caddy-v2-health-check-with-php-fastcgi-not-working/22334

Health checks make a `Request` and try to log it with `LoggableHTTPRequest`. When trying to marshal it, we do `GetVar(r.Context(), ClientIPVarKey).(string)` which panics because we never set this var in context for health checks.

Fix is just to type assert in a guard instead of blindly using the value. Good enough.


mholt (approved):
Ope, thanks!

PR #6042, opened 2024-01-15
cmd: reverseproxy: log: use caddy logger

mohammed90:
I noticed those print lines the other day and thought they looked like the an odd duck.


mholt:
Thanks!

I am not opposed to this; the main reason I did this is because the user-facing messages print to stdout, the logs go to stderr, and this is typically an interactive command. At least, I think so. I have no telemetry to know for sure :frowning_face: 

You're right though, they do stand out a bit as such.

Anyone else have an opinion? I'm good either way.

mholt (approved):

Issue #5982, opened 2023-12-14
quic-go v0.40.1 (CVE-2023-49295)


bt90:
quic-go fixed a vulnerability in the most recent patch release. I couldn't find any details yet, but it was backported down to v0.37.x which is an indicator for a higher severity.

The CVE isn't public yet: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-49295

francislavoie:
We're aware.

FWIW you don't need to open an issue every time there's a quic-go release! We're notified of releases and we keep in contact with Marten.

mholt:
Yep, saw this and will be addressing it next. You could of course submit a PR that updates to the latest version of quic-go and then all we have to do is merge it, for a faster resolution.

Thank you!

armadi1809:
@mholt I just submitted a PR that updates quic-go to its latest version. Is that all what is needed here?

mholt:
Yep, that's perfect! Thanks

bt90:
Any ETA for a patch release? The vulnerability has been disclosed now:

https://seemann.io/posts/2023-12-18-exploiting-quics-path-validation/

PR #6043, opened 2024-01-17
update quic-go to v0.41.0

marten-seemann:
Tons of changes in the [v0.41.0 release](https://github.com/quic-go/quic-go/releases/tag/v0.41.0). Most relevant for caddy is probably https://github.com/quic-go/quic-go/pull/4181, which should fix https://github.com/caddyserver/caddy/issues/5951 on HTTP/3.

https://github.com/quic-go/quic-go/pull/4230 might also be useful. I see that you use `ConnContext` on HTTP/1.1 and HTTP/2 (haven't looked at it in detail though), you can now do that on HTTP/3 as well.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/caddyserver/caddy/v2
 
-go 1.20
+go 1.21
```
francislavoie, line 3: I don't think we're ready to bump our minimum version yet. Is this necessary?

marten-seemann: quic-go dropped support Go 1.20 a bit earlier than usual this team. See release notes for justification.

So yes, updating to v0.41.0 will unfortunately force you move away from Go 1.20.

francislavoie: Okay. It might take us a little while to merge this. I think our next release v2.8.0 is still a little ways away (partly cause we have a lot of pending PRs to review and merge, partly because Matt is busy being a parent). So I think by the time we'll be ready, 1.22 will be out.

bt90: > I think our next release v2.8.0 is still a little ways away

A smaller patch release with #5982 would still be good though.

francislavoie: Understood @bt90 but that would take time to make a branch & cherrypick etc. I don't think @mholt has the time to do that without it being sponsored/paid for.

mholt: Go 1.22 is slated for just a couple/few weeks. I'm willing to bet v2.8 won't be ready before then. Should we just merge this?

francislavoie: @mholt sure, we'll need to also update CI config (and repo config) to test 1.21 + 1.22

mholt: The justification for dropping Go 1.20 support is really good: no more crypto/tls fork.

One other thing to consider is that v0.42 will have an improved API for a security feature. https://github.com/quic-go/quic-go/issues/3549#issuecomment-1905301886 -- we could also just wait until that gets released and kill 2 birds with 1 stone. Either way is fine with me.

modules/caddyhttp/replacer.go
```diff
@@ -470,7 +469,11 @@ func marshalPublicKey(pubKey any) ([]byte, error) {
 	case *rsa.PublicKey:
 		return asn1.Marshal(key)
 	case *ecdsa.PublicKey:
-		return elliptic.Marshal(key.Curve, key.X, key.Y), nil
+		e, err := key.ECDH()
+		if err != nil {
+			return nil, err
+		}
+		return e.Bytes(), nil
```
francislavoie, line 476: `elliptic.Marshal` is deprecated. I'm not sure if this change is correct, wouldn't mind a second pair of eyes. But I think this is the intended use of the new API?

francislavoie:
Alright, minimum Go bumped, & CI updated to 1.21 + 1.22-rc.2 for now. Some code changes to match.

@mholt just need to remove the Go 1.20 CI requirement then I think we can merge.

mholt:
@francislavoie Updated repo requirements. Go 1.21 and 1.22 now required. Thanks!

francislavoie (approved):

PR #6047, opened 2024-01-18
refactor: move automaxprocs init in caddycmd.Main()

dunglas:
Closes https://github.com/caddyserver/xcaddy/pull/168.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/caddyserver/certmagic"
 	"github.com/spf13/pflag"
+	"go.uber.org/automaxprocs/maxprocs"
 	"go.uber.org/zap"
 
 	"github.com/caddyserver/caddy/v2"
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -64,6 +65,12 @@ func Main() {
 		os.Exit(caddy.ExitCodeFailedStartup)
 	}
 
+	undo, err := maxprocs.Set()
+	defer undo()
```
mohammed90, line 69: What are the consequences if the undo function isn't run? I'm asking because `os.Exit` overrides the `defer` runs and terminates the program immediately. Or maybe we can call `undo` in line 75 (68 before this PR).

dunglas: I think there are no consequences because the program exits.

francislavoie: So would undo only be useful in case some weird custom build of Caddy calls `caddycmd.Main()` in a loop or something? I'm not sure how that would even work 🤔

mohammed90: I think it's there to allow for momentarily some workloads.

mholt: That is interesting. But yeah, if the process exits, any side-effects are the responsibility of the OS, so I think this is fine :man_shrugging:

mholt (approved):
This looks good. Thanks Kevin!

mholt:
Also, not sure why I didn't do this earlier, but I just granted you collaborator privileges. :)

dunglas:
Thank you for the commit bit @mholt!! This means a lot for me :)

PR #6050, opened 2024-01-18
caddytls: clientauth: leaf verifier: make trusted leaf certs source pluggable

### LINKED ISSUE #6046: Providing Trusted Leaf Certificates to the Leaf Certificate Verifier
The [leaf-certificate verifier](https://caddyserver.com/docs/json/apps/http/servers/tls_connection_policies/client_authentication/verifiers/leaf/) in client authentication does not have a way to provide the trusted leaf certificate material without relying no the deprecated field [`trusted_leaf_certs`](https://caddyserver.com/docs/json/apps/http/servers/tls_connection_policies/client_authentication/#trusted_leaf_certs), which was not meant to be the proxy forever and only left around as a bridge for existing users and not to be used after the introduction of the feature in #4389. At the moment, there's no way to configure the `"verifier": "leaf"`directly with new behavior only.

We should provide a way to provide the trusted certificates to the `tls.client_auth.leaf` module.

armadi1809:
Closes #6046


armadi1809:
@mohammed90, I don't know if this is what you meant by "we should make this pluggable" so this is still in draft. I would like to get your early feedback about where this is going. This is my first time creating a caddy module from scratch and hosting it so any advice is appreciated 😅 Thank you in advance for your help.

mholt:
Thanks @armadi1809 !

@mohammed90 , is this similar to https://github.com/caddyserver/caddy/pull/5784?

mohammed90:
> @mohammed90 , is this similar to #5784?

No. This one is for the verifiers.

Thanks, @armadi1809! I'm a bit busy this weekend, but you'll need to go through this page:
https://caddyserver.com/docs/extending-caddy. This doesn't properly implement a Caddy module. The PR is not pluggable. Effectively, this still only accepts one kind of source of certificates. The type name and tag says certificate names", but it's just decoding DERs which means it only accepts certificates that are directly given in the configuration as DER. What if the certificates are stored in a file in the filesystem? What if I want to read them from an HTTP endpoint? This is the test of pluggable, they could come from a plugin.

armadi1809:
@mohammed90 Thank you for your explanation. I actually went through that document before working on this but looks like I didn't totally get your point. I made some changes in my last commit but I still have some doubts regarding the concept of "pluggable" in this case. I changed the logic to have the new module contain the decoded certs and provide them to the `tls.client_auth.leaf` module in the provision step. Sorry in advance if this is not what you meant.

francislavoie:
> I changed the logic to have the new module contain the decoded certs and provide them to the `tls.client_auth.leaf` module in the provision step. Sorry in advance if this is not what you meant.

Did you forget to push? I don't see your changes, I only see your rebase from earlier.

armadi1809:
The changes are part of the rebase. I changed the type of the new module we're adding from a `[]string` to a `[]*x509.Certificate` and removed the logic that was decoding DERs and replaced it by a simple assignment in the provision step of the `tls.client_auth.leaf`.

francislavoie:
Okay, it's still not what @mohammed90 was asking for - it's still not pluggable.

To be pluggable, there needs to be an interface that a plugin could implement to provide info or perform a task.

Read through https://caddyserver.com/docs/extending-caddy, especially the part on namespaces & interfaces.

We want this to be a [Host Module](https://caddyserver.com/docs/extending-caddy#host-modules) which can take a guest module as input, and guest modules will load certificates in various ways (i.e. from file, or statically, or from an HTTP endpoint, or a storage plugin, etc).

mohammed90:
Francis is right :100: 

> We _do_ already have the `caddytls.CertificateLoader` interface... it might be possible to reuse that interface, under a different namespace? WDYT @mohammed90 🤷‍♂️

I think it's best to re-use the interface and the namespace, which is `tls.certificates`. This way we get a collection of implementations automatically.

francislavoie:
I think it would have to special-case skip the `automate` module though (or maybe that already _just works_ because it doesn't actually conform to that interface, and would get rejected by the type assertion?)

mohammed90:
> because they try to load a cert + key

Yeah, just looked at the code now and noticed the same thing. Re-using the existing interface won't work because the data returned by the interface, i.e. `caddytls.Certificate`, contains `tls.Certificate` while we need `x509.Certificate`.

We'll need a new interface and a new namespace.

armadi1809:
I think I am starting to understand 😅 . Thanks for mentioning the `tls.certificates` namespace. Looking at the code there kind of gave me an idea about what is wanted here. I will give it another shot.

mholt:
You're doing great! Sorry I haven't been more involved in the process lately (baby takes up my spare brain cycles and time) -- but you're on the right track. Thank you!

armadi1809:
I think this now aligns closer to what you guys meant, right? 
If we agree there, my next question is, would we be writing the guest modules for this new loader on this PR? And if so, what are the different types of loaders we want to implement?

modules/caddytls/connpolicy.go
```diff
@@ -587,6 +609,12 @@ type ConnectionMatcher interface {
 	Match(*tls.ClientHelloInfo) bool
 }
 
+// LeafCertificateLoader is a type that loads the trusted leaf certificates
+// for the tls.client_auth.leaf module
```
mholt, line 613:
```suggestion
// for the tls.leaf_cert_loader modules
```

I think this is a little more accurate (but I knew what you meant)

modules/caddytls/connpolicy.go
```diff
@@ -537,7 +537,8 @@ func setDefaultTLSParams(cfg *tls.Config) {
 
 // LeafCertClientAuth verifies the client's leaf certificate.
 type LeafCertClientAuth struct {
-	TrustedLeafCerts []*x509.Certificate
+	LeafCertificateLoadersRaw []json.RawMessage `json:"leaf_certs_loaders,omitempty" caddy:"namespace=tls.leaf_cert_loader inline_key=loader"`
+	TrustedLeafCerts          []*x509.Certificate
```
mholt, line 541: Now that I see this I wonder if this field should be unexported. (Or be omitted from JSON:
```
`json:"-"`
```
)

mholt (changes requested):
Looking good! I think we're close :smiley:

modules/caddytls/connpolicy.go
armadi1809: Can I ask how will we be providing the loaders then if we are not including them in the JSON?  Not questioning your comment but rather trying to wrap my head around the entire process that would get the certificates to the `LeafCertClientAuth` . Thank you!

mholt: Sorry, to clarify: `TrustedLeafCerts` does not currently have a JSON tag. It's exported though, which means users can configure it like `TrustedLeafCerts: ...` in their JSON config, which isn't conventional, as we use `snake_casing`: `trusted_leaf_certs`. If the user is intended to use the `LeafCertificateLoadersRaw` field to set `TrustedLeafCerts` and this field isn't intended to be user-configurable directly, it should be unexported.

armadi1809: Oh that's completely my bad. I thought you were talking about the `LeafCertificateLoadersRaw` field. Sorry. This should be addressed now.

francislavoie:
I think we should have some built in, same as `caddytls.CertificateLoader`. Can probably copy-paste those but just adjust them to only read in certs and not keys

armadi1809:
I added some built in leaf certs loaders. I didn't know if I should be incorporating the file and the folder loaders into the Caddyfile config similarly to the `tls.certificates` loaders or not. I will wait for your feedback on that.

modules/caddytls/leaffileloader.go
```diff
@@ -0,0 +1,105 @@
...
+			Tags:            pair.Tags,
+		}
+	}
+	return nil
+}
+
+// CaddyModule returns the Caddy module information.
+func (LeafFileLoader) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "leaf_cert_loader.load_files",
```
mohammed90, line 54: The namespace is `tls.leaf_cert_loader`, so the implementation module should have the namespace as the prefix, not just the middle part. In other words, the namespace here should be `tls.leaf_cert_loader.load_files`, though I think the module name `files` (the suffix part) is more apt because we already know it's a loader per the namespace.

armadi1809: I was planning on adding the `tls` prefix but I completely forgot to do it. I agree with the naming though. I just updated the code to include those two changes for the four modules I added. Thank you for your feedback!

mholt:
Thanks for working on this and so many other patches recently, @armadi1809. Could you join our developer Slack? That way you're welcome to discuss changes and other Caddy things in more real-time with our team. Just let me know which email address to send the invite to.

armadi1809:
@mholt Thank you for the invite! I appreciate it. You can use azizrmadi@gmail.com

mholt:
I haven't been able to read every single line of this but so far it's looking pretty good to me! @mohammed90 what do you think?

modules/caddytls/leaffileloader.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		var cert *x509.Certificate
+		switch pair.Format {
+		case "":
+			fallthrough
+		case "pem":
+			cert, err = x509.ParseCertificate(certData) // tls.X509KeyPair(certData, keyData)
```
mohammed90, line 88: I think the comment is left over from another file, right? You can remove it

modules/caddytls/leafpemloader.go
```diff
@@ -0,0 +1,87 @@
...
+	// Can be useful when you want to select a particular
+	// certificate when there may be multiple valid candidates.
+	Tags []string `json:"tags,omitempty"`
+}
+
+// LoadLeafCertificates returns the certificates contained in pl.
+func (pl LeafPEMLoader) LoadLeafCertificates() ([]*x509.Certificate, error) {
+	certs := make([]*x509.Certificate, 0, len(pl))
+	for i, pair := range pl {
+		cert, err := x509.ParseCertificate([]byte(pair.CertificatePEM))
```
mohammed90, line 74: The `x509.ParseCertificate` expects DER, not PEM ([docs](https://pkg.go.dev/crypto/x509#ParseCertificate)). This will always fail. See the logic here on how to convert PEM to DER for parsing:

https://github.com/caddyserver/caddy/blob/e1aa862e6a951d4dc9c0e89b6d42f7fea1c5aba7/modules/caddytls/connpolicy.go#L512-L533

modules/caddytls/leafpemloader.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func init() {
+	caddy.RegisterModule(LeafPEMLoader{})
+}
+
+// LeafPEMLoader loads leaf certificates by
+// decoding their PEM blocks directly. This has the advantage
+// of not needing to store them on disk at all.
+type LeafPEMLoader []LeafCertPEM
```
mohammed90, line 31: This is nice. The currently inherited implementation accepts DER, which is more convenient in config files because it's a single line of text, unlike PEM. It's probably more convenient to have DER only, at least for now.

modules/caddytls/leafstorageloader.go
```diff
@@ -0,0 +1,104 @@
...
+
+// LeafStorageLoader loads leaf certificates from the
+// globally configured storage module.
+type LeafStorageLoader struct {
+	// A list of pairs of certificate file names along with their
+	// encoding format so that they can be loaded from storage.
+	Certs []LeafCertFile `json:"certs,omitempty"`
+
+	// Reference to the globally configured storage module.
+	storage certmagic.Storage
```
mohammed90, line 38: Add a field for custom storage for this module with default to `ctx.Storage()`. Here's an example:

https://github.com/caddyserver/caddy/blob/e1aa862e6a951d4dc9c0e89b6d42f7fea1c5aba7/modules/caddytls/capools.go#L337-L351

mohammed90 (changes requested):
Thank you, @armadi1809, for working on this! There are minor findings to be resolved. Can you please add tests? At least one of the points would have been caught with tests 😉

modules/caddytls/leafpemloader.go
armadi1809: So are you suggesting getting rid of this module altogether? Just making sure 😅 
(Thanks for your feedback !)

mohammed90: Yeah, switch it to DER loader.

armadi1809:
@mohammed90 sorry this is taking too long, as there were some misunderstanding on my part on how to write the JSON config for the loaders 😅 . With that out of the way, I was wondering what would be the best way to test this? Is there a way to test if a given caddy JSON config is valid inside the `caddytest` package or are you envisioning the tests to spawn a caddy instance with a given config (using `InitServer`) to make sure the config loads properly and we are able to get a response back for a request?

armadi1809:
I made some improvements to the way we were loading certificates from the different loaders (After @mohammed90 review). This still needs tests but like I said in my previous comment, I am not entirely sure regarding the best way to test this.

mohammed90 (changes requested):
Thanks for working on this, @armadi1809! I did a round of review. We're still missing tests. Some of the findings would've been caught with tests. Can you try adding some?

mohammed90:
> I am not entirely sure regarding the best way to test this

Testing this from the `caddytest`/integration package could be hard. It'll require you to customize the HTTP client to present the expected certificate during the HTTP request. If you're up to that challenge, go for it. Otherwise, you can probably just confirm the config load is successful.

You can probably create a fresh `caddy.Context`, then call `Provision` then `LoadLeafCertificates` on each of the modules' tests to check if the loaded data matches the expectation.

armadi1809:
@mohammed90 I added tests for the file and folder loaders. Is that what you meant by loading the config and check whether or not it meets the expectations?

caddytest/caddy.examplecert.pem
```diff
@@ -0,0 +1,15 @@
+-----BEGIN CERTIFICATE-----
```
mohammed90, line 1: Rename the file to `leafcert.pem` and add a newline at the end

modules/caddytls/leaffileloader_test.go
```diff
@@ -0,0 +1,44 @@
...
+	fl.Provision(caddy.Context{Context: context.Background()})
+
+	out, err := fl.LoadLeafCertificates()
+
+	if err != nil {
+		t.Errorf("Leaf certs file loading test failed: %v", err)
+	}
+	pemBytes := bytes.NewBuffer(nil)
+	pem.Encode(pemBytes, &pem.Block{Type: "CERTIFICATE", Bytes: out[0].Raw})
+	os.WriteFile("./test.txt", pemBytes.Bytes(), 0644)
```
mohammed90, line 24: There's no need to write to file here

modules/caddytls/leaffileloader_test.go
```diff
@@ -0,0 +1,44 @@
...
+	fl := LeafFileLoader{Files: []string{"../../caddytest/caddy.examplecert.pem"}}
+	fl.Provision(caddy.Context{Context: context.Background()})
+
+	out, err := fl.LoadLeafCertificates()
+
+	if err != nil {
+		t.Errorf("Leaf certs file loading test failed: %v", err)
+	}
+	pemBytes := bytes.NewBuffer(nil)
+	pem.Encode(pemBytes, &pem.Block{Type: "CERTIFICATE", Bytes: out[0].Raw})
```
mohammed90, line 23: Add a length check before using the cert. Also, you can use [`pem.EncodeToMemory`](https://pkg.go.dev/encoding/pem#EncodeToMemory) to avoid having to create the buffer.

modules/caddytls/leaffileloader_test.go
```diff
@@ -0,0 +1,44 @@
...
+ZzBcMA0GCSqGSIb3DQEBAQUAA0sAMEgCQQCp5hnG7ogBhtlynpOS21cBewKE/B7j
+V14qeyslnr26xZUsSVko36ZnhiaO/zbMOoRcKK9vEcgMtcLFuQTWDl3RAgMBAAGj
+gbEwga4wHQYDVR0OBBYEFFXI70krXeQDxZgbaCQoR4jUDncEMH8GA1UdIwR4MHaA
+FFXI70krXeQDxZgbaCQoR4jUDncEoVukWTBXMQswCQYDVQQGEwJDTjELMAkGA1UE
+CBMCUE4xCzAJBgNVBAcTAkNOMQswCQYDVQQKEwJPTjELMAkGA1UECxMCVU4xFDAS
+BgNVBAMTC0hlcm9uZyBZYW5nggEAMAwGA1UdEwQFMAMBAf8wDQYJKoZIhvcNAQEE
+BQADQQA/ugzBrjjK9jcWnDVfGHlk3icNRq0oV7Ri32z/+HQX67aRfgZu7KWdI+Ju
+Wm7DCfrPNGVwFWUQOmsPue9rZBgO
+-----END CERTIFICATE-----
+` {
```
mohammed90, line 41: Don't duplicate/embed the cert content here. The file is the source of truth. Use the same file content for the comparison.

modules/caddytls/leaffolderloader_test.go
```diff
@@ -0,0 +1,42 @@
...
+ZzBcMA0GCSqGSIb3DQEBAQUAA0sAMEgCQQCp5hnG7ogBhtlynpOS21cBewKE/B7j
+V14qeyslnr26xZUsSVko36ZnhiaO/zbMOoRcKK9vEcgMtcLFuQTWDl3RAgMBAAGj
+gbEwga4wHQYDVR0OBBYEFFXI70krXeQDxZgbaCQoR4jUDncEMH8GA1UdIwR4MHaA
+FFXI70krXeQDxZgbaCQoR4jUDncEoVukWTBXMQswCQYDVQQGEwJDTjELMAkGA1UE
+CBMCUE4xCzAJBgNVBAcTAkNOMQswCQYDVQQKEwJPTjELMAkGA1UECxMCVU4xFDAS
+BgNVBAMTC0hlcm9uZyBZYW5nggEAMAwGA1UdEwQFMAMBAf8wDQYJKoZIhvcNAQEE
+BQADQQA/ugzBrjjK9jcWnDVfGHlk3icNRq0oV7Ri32z/+HQX67aRfgZu7KWdI+Ju
+Wm7DCfrPNGVwFWUQOmsPue9rZBgO
+-----END CERTIFICATE-----
+` {
```
mohammed90, line 39: same as above

mohammed90 (changes requested):
This is looking better! Just a few cleanup items.

Also, add some tests under `caddytest/integration` that will simply do `tester.InitServer` with config that has leaf verifiers. Just 1 for comfort.

armadi1809:
@mohammed90 all the feedback items from above should now be resolved. I also added a test for the PEM loader as well as an integration test.

modules/caddytls/leafpemloader.go
```diff
@@ -0,0 +1,76 @@
...
+
+func init() {
+	caddy.RegisterModule(LeafPEMLoader{})
+}
+
+// LeafPEMLoader loads leaf certificates by
+// decoding their PEM blocks directly. This has the advantage
+// of not needing to store them on disk at all.
+type LeafPEMLoader struct {
+	Certificates []string `json:"certs,omitempty"`
```
mholt, line 32:
```suggestion
	Certificates []string `json:"certificates,omitempty"`
```

modules/caddytls/leaffolderloader.go
```diff
@@ -0,0 +1,97 @@
...
+
+func init() {
+	caddy.RegisterModule(LeafFolderLoader{})
+}
+
+// LeafFolderLoader loads certificates and their associated keys from disk
+// by recursively walking the specified directories, looking for PEM
+// files which contain both a certificate and a key.
+type LeafFolderLoader struct {
+	Folders []string `json:"folders"`
```
mholt, line 35:
```suggestion
	Folders []string `json:"folders,omitempty"`
```

modules/caddytls/leafstorageloader.go
```diff
@@ -0,0 +1,130 @@
...
+func init() {
+	caddy.RegisterModule(LeafStorageLoader{})
+}
+
+// LeafStorageLoader loads leaf certificates from the
+// globally configured storage module.
+type LeafStorageLoader struct {
+	// A list of pairs of certificate file names along with their
+	// encoding format so that they can be loaded from storage.
+	Certs []string `json:"certs,omitempty"`
```
mholt, line 37:
```suggestion
	Certificates []string `json:"certificates,omitempty"`
```

mohammed90 (approved):
Looks great now! Thank you, Aziz, for working on this.

modules/caddytls/leafstorageloader.go
```diff
@@ -0,0 +1,130 @@
...
+
+func init() {
+	caddy.RegisterModule(LeafStorageLoader{})
+}
+
+// LeafStorageLoader loads leaf certificates from the
+// globally configured storage module.
+type LeafStorageLoader struct {
+	// A list of pairs of certificate file names along with their
+	// encoding format so that they can be loaded from storage.
```
mholt, line 36: What does this comment mean, exactly? :thinking:

mholt (changes requested):
Thank you, I think we're getting close!! Just a couple of nits and a question to clarify.

modules/caddytls/leafstorageloader.go
armadi1809: Sorry, that was from a previous version of this loader before some feedback from Mohammed. It should be more meaningful now 😅

armadi1809:
@mholt Thanks for reviewing. All the feedback items should now be resolved. 

And thank you @mohammed90 for all your efforts with this!

mholt (changes requested):
Awesome! Just one thing I missed, then we merge. ha

mholt:
That's weird, govulncheck wasn't failing just 30 minutes ago... fluke?

mohammed90:
> That's weird, govulncheck wasn't failing just 30 minutes ago... fluke?

I think the vuln was published recently: https://pkg.go.dev/vuln/GO-2024-2611

modules/caddytls/leafstorageloader.go
```diff
@@ -0,0 +1,129 @@
...
+
+func init() {
+	caddy.RegisterModule(LeafStorageLoader{})
+}
+
+// LeafStorageLoader loads leaf certificates from the
+// globally configured storage module.
+type LeafStorageLoader struct {
+	// A list of certificate file names to be loaded from storage.
+	Certs []string `json:"certificates,omitempty"`
```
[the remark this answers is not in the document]
armadi1809: No worries!  Should be good to go now.

mholt:
Of course, our luck :roll_eyes:

mohammed90:
> Of course, our luck 🙄

Yeah 😶 The fastest course of action is if you, @mholt, update the dep `google.golang.org/protobuf` on master then rebase/merge master here. Otherwise, it'll require me to push another PR.

Or... @armadi1809, can you just run this?

```
go get -u google.golang.org/protobuf
```

armadi1809:
Okay... I think we should be good to go now 😅

mholt (approved):
YES -- nice work, thanks for the contribution!

(And you are fast. I am working through a lot of things today so you beat me to it!)

PR #6054, opened 2024-01-19
matchers: `query` now ANDs multiple keys

### LINKED ISSUE #6053: Query matcher doesn't AND different keys
See https://caddy.community/t/not-query-in-named-matcher-not-working-as-expected/22367 for the discussion.

If a query matcher is defined with more than one query key, it will match if any of the keys have a matching value.

Instead, it should match only if all keys have at least one matching value (or `*` of course).

armadi1809:
Closes #6053 

@francislavoie please let me know if you would rather solve the issue some other way. Thanks!

### DIFF
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -810,6 +810,10 @@ func (m MatchQuery) Match(r *http.Request) bool {
 		// We regard the Go team's decision as sound and thus reject malformed query strings.
 		return false
 	}
+
+	// Count the amount of matched keys, to ensure we AND
+	// between all configured query keys; all keys must
+	// match at least one value.
 	matchedKeys := 0
 	for param, vals := range m {
 		param = repl.ReplaceAll(param, "")
--- a/modules/caddyhttp/matchers.go
+++ b/modules/caddyhttp/matchers.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
+	"golang.org/x/exp/slices"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
@@ -789,6 +790,10 @@ func (m *MatchQuery) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 // Match returns true if r matches m. An empty m matches an empty query string.
 func (m MatchQuery) Match(r *http.Request) bool {
+	if len(m) == 0 {
+		return len(r.URL.Query()) == 0
+	}
+
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 
 	// parse query string just once, for efficiency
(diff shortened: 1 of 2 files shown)


mholt (commented):
I wonder if it'd be a little more robust / readable if it instead checked for any that _don't_ match?

francislavoie (commented):
Probably need a couple more test cases to cover when there's query keys with a value that doesn't match (not only keys that don't exist in the request)

armadi1809:
@mholt I am not sure I totally get what you mean by "if it instead checked for any that don't match?". I tried to tweak the logic a little bit to make it more readable / less nested. Happy to rework it again If you still don't like it. 

@francislavoie I added a couple more tests around the case where the query key values do not match. 

Also, I am not sure why the runner failed to download the action in the setup job of the failing test 😅

francislavoie:
I think I see a possible bug. What if there's two values configured for a key and both match? It would increment twice. Need a test for that. I think the fix is to add a `continue` after incrementing (i.e. OR operation by short circuiting on the first positive match)

armadi1809:
I thought about this but since we're only checking against `paramVal[0]`, it won't increment the counter twice. However, now that I am thinking about it, there is actually another potential bug. Let's say the matcher is `MatchQuery{"debug": []string{"1"}}` and the input is `"/?debug=1&debug=2"`. This would work and we would get true back from the matcher. However, if the input is reversed (i.e. `"/?debug=2&debug=1"`) this would give us false because we are only checking against `paramVal[0]`. I don't exactly get why we're only checking agains the first element but I believe a better approach would be to change the check to `slices.Contains(paramVal, v)` and `continue` on the first match as you suggested above. Does this sound reasonable?

francislavoie:
Yeah, I think so. 👍

armadi1809:
@francislavoie Any idea why some of the builds are failing to get the "slices" package?

francislavoie:
Slices isn't available in Go 1.20 yet.

armadi1809:
Should I then be using `"golang.org/x/exp/slices"` similar to the `caddyconfig/httpcaddyfile/httptype.go` file or manually check if the slice contains the value I am looking for?

francislavoie:
Yeah, you can use `"golang.org/x/exp/slices"`, we're already using it somewhere else in the codebase.

armadi1809:
Should be good to go now. Thanks for the help!

francislavoie:
Only thing is I'm not quite sure what the behaviour should be when there's multiple values provided for a query. There's an argument to be said that it should only match if there's a single matching value. But we didn't really envision this being used with multi-value query keys.

Whatever we decide (what you have now is probably fine), we'll need to document clearly the way it behaves. We should adjust the `MatchQuery` godoc comment for the JSON docs https://caddyserver.com/docs/json/apps/http/servers/routes/match/query/.

armadi1809:
That sounds good. Let me know if we decide to treat that case differently then and I can adjust the code accordingly! Thanks again for your review/feedback.

modules/caddyhttp/matchers.go
```diff
@@ -805,20 +810,22 @@ func (m MatchQuery) Match(r *http.Request) bool {
 		// We regard the Go team's decision as sound and thus reject malformed query strings.
 		return false
 	}
-
+	matchedKeys := 0
```
francislavoie, line 813: I think a code comment is warranted here to clarify the intent

```suggestion

	// Count the amount of matched keys, to ensure we AND
	// between all configured query keys; all keys must
	// match at least one value.
	matchedKeys := 0
```

modules/caddyhttp/matchers.go
```diff
@@ -789,6 +790,10 @@ func (m *MatchQuery) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 // Match returns true if r matches m. An empty m matches an empty query string.
 func (m MatchQuery) Match(r *http.Request) bool {
+	if len(m) == 0 {
+		return len(r.URL.Query()) == 0
+	}
```
francislavoie, line 795:
```suggestion
	// If no query keys are configured, this only
	// matches an empty query string.
	if len(m) == 0 {
		return len(r.URL.Query()) == 0
	}
```

francislavoie (approved):
LGTM, thanks for working on this!

Issue #5986, opened 2023-12-15
Avoid specifying on demand 'ask' endpoint when using my own issuer


jroyalty:
I've written my own `tls.issuance` module, similar to `tls.issuance.internal`.  I want to use on-demand behavior here but without any limits; that is, I don't want to specify an `ask` endpoint or any `rate_limit` configuration in this case.

I see that Caddy has a strict set of conditions under which an ask endpoint isn't enforced -- namely that you are using only `tls.issuance.internal`.  I certainly understand these safe guards being in place, but I'm wondering if there has been consideration given to how one might opt-out of these when needed.

Cheers!

mholt:
(The `rate_limit` has always been optional anyway.)

You could set up a simple ask endpoint (within Caddy itself, just a site that's bound to localhost) that always returns 200 OK.

francislavoie:
For example:

```
{
	on_demand_tls {
		ask http://localhost:5555/
	}
}

http://localhost {
	respond 200
}

https:// {
	tls {
		on_demand
		issuer your-issuer
	}

	# your usual config
}
```

jroyalty:
Thanks for the response, and the suggestion.  I tried this previously and while it works, I'm looking for a way to avoid the network call during the cert pipeline.  Perhaps this could be a module with a default that has the same behavior as the current ask?

francislavoie:
Yeah, we'll probably add `tls.on_demand.decision` module support at some point. But until then, the above solution is the simplest and fastest. The HTTP request to Caddy is extremely fast, so you won't notice any kind of bottleneck there.

jroyalty:
Okay, consider this a +1 for the module. :)

ankon:
See https://github.com/caddyserver/caddy/pull/6055

mholt:
@jroyalty Would you be able to try #6055?

jroyalty:
@mholt I looked at the PR and it seems to fit the bill!  I'll give it a try as well.  Thanks!

PR #6055, opened 2024-01-19
caddytls: Make on-demand 'ask' permission modular

mholt:
Closes https://github.com/caddyserver/caddy/issues/5986

This makes the 'ask' endpoint a module, which means that developers can write custom plugins for granting permission for on-demand certificates.

Module namespace: `tls.permission.*`

The default HTTP permission module is the same code as the previous "ask" implementation.

This was a little more involved than I expected. :sweat_smile: 

/cc @ankon @danillouz


francislavoie:
I still think `decider` is a better module namespace than `permission` :man_shrugging:

modules/caddytls/ondemand.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return fmt.Errorf("parsing ask URL: %v", err)
+	}
+	qs := askURL.Query()
+	qs.Set("domain", name)
+	askURL.RawQuery = qs.Encode()
+	askURLString := askURL.String()
+
+	var remote string
+	if chi, ok := ctx.Value(certmagic.ClientHelloInfoCtxKey).(*tls.ClientHelloInfo); ok && chi != nil {
```
francislavoie, line 142: I like this :+1:

francislavoie:
Rebased & lint

mholt:
Ah thanks! I need to see if I can configure my VS Code to follow those rules as well.

modules/caddytls/tls.go
```diff
@@ -164,6 +164,33 @@ func (t *TLS) Provision(ctx caddy.Context) error {
...
+		onDemandRateLimiter.SetMaxEvents(0)
+	}
+
+	// run replacer on ask URL (for environment variables) -- return errors to prevent surprises (#5036)
+	if t.Automation != nil && t.Automation.OnDemand != nil && t.Automation.OnDemand.Ask != "" {
+		t.Automation.OnDemand.Ask, err = repl.ReplaceOrErr(t.Automation.OnDemand.Ask, true, true)
+		if err != nil {
+			return fmt.Errorf("preparing 'ask' endpoint: %v", err)
+		}
+	}
```
ankon, line 192: Probably missing something here, but when reading this it looks duplicated (the `PermissionsByHTTP` module is doing it on it's endpoint already), and also a bit off: Wouldn't the parsing of the configuration already have the `ask` field consumed here, and produced a `PermissionRaw` thing?

ankon (commented):
Small little inline note, and also a question/suggestion: I guess it would be good to document what happens if you were to provide both a permissions module and an `ask` value? Right now it looks like "last one wins?"

modules/caddytls/tls.go
francislavoie: There's JSON config _and_ Caddyfile config to worry about. This PR changes the Caddyfile to always produce the HTTP module for backwards compatibility, but we can't control what users of JSON continue to use in their config so we need to still accept the old field in JSON and warn to let them switch over to the new way.

mholt: Yeah, so this is a compatibility layer for the 'ask' property of the JSON config, which many people will still be using. This code as well as the 'ask' property will eventually be removed once enough people have time to migrate.

mholt:
@ankon 

Thanks for the review!

> I guess it would be good to document what happens if you were to provide both a permissions module and an ask value? Right now it looks like "last one wins?"

Huh, I coulda sworn I wrote code to account for that: I even remember returning an error message about a conflicting configuration. Thanks for pointing this out. I _know_ I wrote it, not sure why it didn't get committed or even saved... :sweat_smile: I'll restore it...

mholt:
Ok, restored the error message if there's conflicting configuration. Either "ask" or "permission" must be set, but not both. Thanks for finding that!

modules/caddytls/tls.go
ankon: Right, thanks! 

(And yes, I know some people that are using JSON :D)

modules/caddytls/ondemand.go
```diff
@@ -0,0 +1,192 @@
...
+	//
+	// The context passed in has the associated *tls.ClientHelloInfo
+	// value available at the certmagic.ClientHelloInfoCtxKey key.
+	//
+	// In the worst case, this function may be called as frequently
+	// as every TLS handshake, so it should return as quick as possible
+	// to reduce latency. In the normal case, this function is only
+	// called when a certificate is needed that is not already loaded
+	// into memory ready to serve.
+	CertificateAllowed(ctx context.Context, name string) error
```
jroyalty, line 90: What do you think about adding the issuer key that is requesting the cert here?  For some use cases, I would want to allow or reject the issuance based on who is asking.

mholt: I don't think an _issuer_ has been determined yet. Multiple issuers can be configured (for example, for redundancy), and if one fails it'll try the next, etc. So this function is called before trying any issuers, and is called only once.

But when you say:

> I would want to allow or reject the issuance based on who is asking.

To clarify, Issuers are entities that issue certificates, not entities that ask for a certificate. The server is asking (or, depending on how you look at it, the client is asking for a cert). Can you elaborate what your needs are?

francislavoie: There's no such thing, the issuer is not involved at this point. This is before we start issuance, it's for us to know "should we do issuance?" and then if we get a yes it'll choose an issuer.

jroyalty: I recall the request flow now, so what you are saying makes sense.

Basically I wanted the ask endpoint to, in effect, be per-issuer.  I know you can filter what issuer is appropriate by using the `subjects` param, but the list of what's appropriate is external to config.  It can be placed behind the `OnDemandPermission` but that's global for all issuers, as noted.

What's in this PR works for my exact use case as long as I have one issuer -- which is the case.  So probably not that big of a deal.

mholt: Yeah, that starts rolling into "dynamically load the TLS config" territory. Which is probably possible, but neither efficient nor simple. Maybe let's revisit that if needed in a separate issue...

francislavoie (approved):
LGTM!

francislavoie:
We're missing Caddyfile support for permission modules here. We should try to get that in for 2.8.0.

mohammed90:
> We're missing Caddyfile support for permission modules here. We should try to get that in for 2.8.0. (If anyone's reading this, PRs welcome if you want a little task)

I've created an issue so it's more visible and easier to track

Issue #5930, opened 2023-11-07
Various breakages when formatting json heredocs


dbaynard:
I've hit various issues, with 2.7.5 and earlier versions of caddy, when it comes to heredocs.

`caddy fmt` introduces spacing which means the resulting file is invalid.

See the caddyfile in https://github.com/dbaynard/caddy-fmt-heredoc-example — I recommend you step through the commits.

0. Empty commit
1.  I created a sample, unformatted caddyfile, which works.

    ```Caddyfile
    	handle_path /never {
    		respond <<JSON
    		\{"response":"nope"\}
    		JSON 200
    	}
    	handle_path /json {
    		respond <<JSON
    		{"response":"nope"}
    		JSON 200
    	}
    ```

    The first appears to be returned as `text/plain`.

2. I adapted it, and it fails to load

    ```Caddyfile
    	handle_path /never {
    		respond <<JSON \{
    		"response":"nope"\}
    		JSON 200
    	}
    	handle_path /json {
    		respond <<JSON 
    		{"response":"nope"}
    		JSON 200
    	}
    ```

    Note the space at the end of the JSON for `/json`.

    ```console
    Error: adapting config using caddyfile: parsing caddyfile tokens for 'handle_path': unrecognized directive: response - are you sure your Caddyfile structure (nesting and braces) is correct?, at Caddyfile:18
    ```

3. I removed the `handle_path` for `/nope` with escaped braces

```console
Error: adapting config using caddyfile: parsing caddyfile tokens for 'handle_path': unrecognized directive: {"response":"nope"} - are you sure your Caddyfile structure (nesting and braces) is correct?, at Caddyfile:18
```

4. I removed the space after the heredoc market, and it's fine, again.

I tried with various different sigils, and saw no difference. In the repo, I have an example with a placeholder (that works correctly). And I have only tried the json formats.

Incidentally, the correct version of this would be helpful as an example.

francislavoie:
Yeah, unfortunately `caddy fmt` doesn't handle heredocs at all right now. The problem is the Caddyfile lexer is completely separate code from the formatter. We'd have to re-implement heredoc lexing in the formatter as well, which is tedious. I want to rewrite the lexer to make it retain whitespace so that it can also be used for formatting, but that's a significant undertaking.

For now, unfortunately you'll need to avoid using `caddy fmt` if you're using heredocs.

dbaynard:
[...]

PR #6056, opened 2024-01-20
caddyfile: Add heredoc support to `fmt` command

bbaa-bbaa:
Fixes https://github.com/caddyserver/caddy/issues/5930

Let the formatter keep heredoc as-is.
Besides, because I missed the review window for #6051, This PR also contains a fix for #6051.
Let me know if you need me to separate them.

### DIFF
--- a/caddyconfig/caddyfile/formatter.go
+++ b/caddyconfig/caddyfile/formatter.go
@@ -99,7 +99,7 @@ func Format(input []byte) []byte {
 
 		if heredoc == heredocOpening {
 			if ch == '\n' {
-				if heredocMarkerRegexp.Match([]byte(string(heredocMarker))) {
+				if len(heredocMarker) > 0 && heredocMarkerRegexp.MatchString(string(heredocMarker)) {
 					heredoc = heredocOpened
 				} else {
 					heredocMarker = nil
--- a/caddyconfig/caddyfile/formatter.go
+++ b/caddyconfig/caddyfile/formatter.go
@@ -31,6 +31,14 @@ func Format(input []byte) []byte {
 	out := new(bytes.Buffer)
 	rdr := bytes.NewReader(input)
 
+	type heredocState int
+
+	const (
+		heredocClosed  heredocState = 0
+		heredocOpening heredocState = 1
+		heredocOpened  heredocState = 2
+	)
+
 	var (
 		last rune // the last character that was written to the result
 
@@ -46,6 +54,11 @@ func Format(input []byte) []byte {
 		comment bool // whether we're in a comment
 		quoted  bool // whether we're in a quoted segment
 		escaped bool // whether current char is escaped
+
+		heredoc              heredocState // whether we're in a heredoc
+		heredocEscaped       bool         // whether heredoc is escaped
+		heredocMarker        []rune
+		heredocClosingMarker []rune
 
 		nesting int // indentation level
 	)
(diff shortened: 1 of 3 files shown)


caddyconfig/caddyfile/formatter_test.go
```diff
@@ -362,6 +362,21 @@ block {
...
 block {
 }
+`,
+		},
+		{
+			description: "keep heredoc as-is",
+			input: `block {
+	heredoc <<HEREDOC
+	Here's more than a space       Here's more than a space
+	<<HEREDOC
```
francislavoie, line 372: This doesn't look right, a heredoc closing marker does not contain the `<<`.

caddyconfig/caddyfile/formatter.go
```diff
@@ -75,6 +89,58 @@ func Format(input []byte) []byte {
...
+				heredocClosingMarker = heredocClosingMarker[1:]
+			}
+			if len(heredocClosingMarker) == len(heredocMarker) &&
+				slices.Equal(heredocClosingMarker, heredocMarker) {
+				heredocMarker = nil
+				heredocClosingMarker = nil
+				heredoc = heredocClosed
+			}
+			continue
+			// check if we're done
```
francislavoie, line 137: This comment doesn't seem like it's at the right place?

francislavoie (commented):
Thanks for getting to this, I wasn't really motivated to work on the fix 😓

caddyconfig/caddyfile/formatter_test.go
bbaa-bbaa: Sorry I forgot that.

caddyconfig/caddyfile/formatter.go
bbaa-bbaa: Yes you are right, I have moved this comment to the correct place.

francislavoie:
I think the slices package is not available in Go 1.20. We'll need to use a different function for now.

bbaa-bbaa:
> 我认为切片包在 Go 1.20 中不可用。我们现在需要使用不同的函数。

Do you have an opinion on this? Should we implement our own Equal function or simply convert it to a string for comparison?

bbaa-bbaa:
> I think the slices package is not available in Go 1.20. We'll need to use a different function for now.

Should we implement our own Equal function or simply convert it to a string for comparison?

bbaa-bbaa:
> I think the slices package is not available in Go 1.20. We'll need to use a different function for now. Can probably just convert the rune slice to a string then compare that way. Or store it as a string (concat as you go) instead of appending runes.

Convert rune slices to strings and compare.

francislavoie (approved):
Thanks, looks good to me!

I didn't study in detail the changes to the formatter logic, but the test cases look good I think.

I'm not sure if there's any gaps in the test coverage. I assume you looked at the existing lexer tests as well?

bbaa-bbaa:
> I'm not sure if there's any gaps in the test coverage. I assume you looked at the existing lexer tests as well?

I think formatter can handle more situations. Because we simply exit the heredoc state instead of returning an error for some incorrect heredoc syntax.

Thanks for your review.

francislavoie:
Do you want to work on that before we merge? Or are you happy with this as-is?

bbaa-bbaa:
Oh, and this code doesn't handle the case where heredocMarker is empty. 
Heredoc will be closed on the next cycle, but maybe we'll lose some possible indentation.
Let me do a quick fix.

bbaa-bbaa:
> Do you want to work on that before we merge? Or are you happy with this as-is?

The original Formatter doesn't seem to have any error handlers, so maybe we should keep it.

PR #6065, opened 2024-01-26
reverseproxy: Implement modular CA provider for TLS transport

### LINKED ISSUE #6063: reverse proxy: use the new CA providers
The PR #5784 introduces modular CA providers. The HTTP(S) transport of the reverse proxy module has a copy of the older approach which was removed in the PR #5784. Moreover, the PR included copying the `TLSConfig` struct with minor modifications to.

In the PR, I opted not to touch the reverse-proxy transport to minimize the impact and reduce the complexity of the PR by leaving it focused to the `caddytls` package. Now that the PR is merged, the reverse-proxy module can re-use the logic and introduced modules to have flexible source of certificate-authority pool. Implementing this means the reverse-proxy transport can trust certificates from, e.g., distributed certificate-authority providers shared over distributed storage or other Caddy instances.

One powerful use case of this is when a Caddy instance acts as an ACME server for the internal services, and other caddy instances serving as load-balancers will trust the same certificate authorities used by the caddy ACME server.

The change for this issue requires looking into, at least, the following parts of the `reverseproxy` package:

https://github.com/caddyserver/caddy/blob/b9c40e711115d35c2809313aeec42ab10fcd6914/modules/caddyhttp/reverseproxy/httptransport.go#L55-L57

https://github.com/caddyserver/caddy/blob/b9c40e711115d35c2809313aeec42ab10fcd6914/modules/caddyhttp/reverseproxy/httptransport.go#L502-L601

armadi1809:
Closes #6063


armadi1809:
Looks like the failed test is just a server timeout.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -564,6 +568,18 @@ func (t TLSConfig) MakeTLSClientConfig(ctx caddy.Context) (*tls.Config, error) {
...
+	if t.CARaw != nil {
+		caRaw, err := ctx.LoadModule(t, "CARaw")
+		if err != nil {
+			return nil, fmt.Errorf("failed to load ca module: %v", err)
+		}
+		ca, ok := caRaw.(caddytls.CA)
+		if !ok {
+			return nil, fmt.Errorf("CARaw module '%s' is not a certificate pool provider", ca)
+		}
+		cfg.ClientCAs = ca.CertPool()
```
mohammed90, line 580: The `ClientCAs` field is for servers who are validating incoming client connections. In reverse-proxy transport, we're doing the opposite. We need the block above to be edited, i.e. modifying the source of root certificates.

armadi1809: Sorry, forgot to change it from the code I ported from your PR. I fixed this. I also added a check to make sure we are not using the old and the new methods of getting the `RootCAs` at the same time. Thanks for the feedback.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -564,6 +568,21 @@ func (t TLSConfig) MakeTLSClientConfig(ctx caddy.Context) (*tls.Config, error) {
...
+		if len(t.RootCAPool) > 0 || len(t.RootCAPEMFiles) > 0 {
+			return nil, fmt.Errorf("Conflicting config for Root CA pool")
+		}
+		caRaw, err := ctx.LoadModule(t, "CARaw")
+		if err != nil {
+			return nil, fmt.Errorf("failed to load ca module: %v", err)
+		}
+		ca, ok := caRaw.(caddytls.CA)
+		if !ok {
+			return nil, fmt.Errorf("CARaw module '%s' is not a certificate pool provider", ca)
```
mohammed90, line 581: Use `ca` instead of `CARaw` because the user does not know the field name inside the struct.

mohammed90 (changes requested):
GitHub does not let me comment on the unchanged lines, so excuse the non-inlined comments.

Mark the old way as deprecated and subject to removal in the future, and emit logs saying the same when used. You can get a logger from `ctx.Logger()`.

modules/caddyhttp/reverseproxy/httptransport.go
armadi1809: I am assuming you are talking about the error message here right? If so, then this should now be resolved. Also, I think you are using the same error message when provisioning the `ClientAuthentication` module inside `modules/caddytls/connpolicy.go`, so maybe we should also change that? 

Sorry in advance if it's not about the error message and if  I misinterpreted your comment 😅

armadi1809:
I just added the "deprecated" comments to `RootCAPool` and `RootCAPEMFiles` fields as well as a deprecation warning log when those are used.

modules/caddyhttp/reverseproxy/httptransport.go
mohammed90: > so maybe we should also change that?

Yep. It was probably an auto-complete mistake, and we missed it.

mohammed90 (changes requested):
Thanks for doing the rounds, Aziz! Sorry for not giving the review points in full at once. I'm squeezing the review time in tight time slots 😅 

We're missing tests and Caddyfile support

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -546,6 +552,7 @@ func (t TLSConfig) MakeTLSClientConfig(ctx caddy.Context) (*tls.Config, error) {
 
 	// trusted root CAs
 	if len(t.RootCAPool) > 0 || len(t.RootCAPEMFiles) > 0 {
+		ctx.Logger().Warn("RootCAPool and RootCAPEMFiles are deprecated. Use one of the tls.ca_pool.source modules instead")
```
mohammed90, line 555: Same for the fields names here. The users only know `root_ca_pool` and `root_ca_pem_files`.

armadi1809:
No worries at all Mohammed! Thank you for your feedback. I implemented the caddy file support and added both caddy file integration tests as well as tests for the `(h *HTTPTransport) UnmarshalCaddyfile` function. Please let me know if anything else is still needed here. Thanks again!

mholt (approved):
Sorry for falling behind -- this is looking better now! I'd be willing to give this a try. Thank you!!

mholt:
@mohammed90 Anything you can think of we need to change for this?

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -1218,6 +1222,28 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+			if !ok {
+				return d.Errf("module %s is not a caddytls.CA", modID)
+			}
+			if h.TLS == nil {
+				h.TLS = new(TLSConfig)
+			}
+			if len(h.TLS.RootCAPEMFiles) != 0 {
+				return d.Err("cannot specify both 'tls_trust_pool' and 'tls_trusted_ca_certs'")
+			}
+			h.TLS.CARaw = caddyconfig.JSONModuleObject(ca, "provider", modStem, nil)
```
mohammed90, line 1245: I think we should return an error if `h.TLS.CARaw` is not nil because it means they've specified it twice.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -564,6 +571,21 @@ func (t TLSConfig) MakeTLSClientConfig(ctx caddy.Context) (*tls.Config, error) {
 		cfg.RootCAs = rootPool
 	}
 
+	if t.CARaw != nil {
+		if len(t.RootCAPool) > 0 || len(t.RootCAPEMFiles) > 0 {
+			return nil, fmt.Errorf("Conflicting config for Root CA pool")
```
mohammed90, line 576:
```suggestion
			return nil, fmt.Errorf("conflicting config for Root CA pool")
```

modules/caddyhttp/reverseproxy/httptransport_test.go
```diff
@@ -0,0 +1,97 @@
...
+					tls_trusted_ca_certs MIIDSzCCAjOgAwIBAgIUfIRObjWNUA4jxQ/0x8BOCvE2Vw4wDQYJKoZIhvcNAQELBQAwFjEUMBIGA1UEAwwLRWFzeS1SU0EgQ0EwHhcNMTkwODI4MTYyNTU5WhcNMjkwODI1MTYyNTU5WjAWMRQwEgYDVQQDDAtFYXN5LVJTQSBDQTCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAK5m5elxhQfMp/3aVJ4JnpN9PUSz6LlP6LePAPFU7gqohVVFVtDkChJAG3FNkNQNlieVTja/bgH9IcC6oKbROwdY1h0MvNV8AHHigvl03WuJD8g2ReVFXXwsnrPmKXCFzQyMI6TYk3m2gYrXsZOU1GLnfMRC3KAMRgE2F45twOs9hqG169YJ6mM2eQjzjCHWI6S2/iUYvYxRkCOlYUbLsMD/AhgAf1plzg6LPqNxtdlwxZnA0ytgkmhK67HtzJu0+ovUCsMv0RwcMhsEo9T8nyFAGt9XLZ63X5WpBCTUApaAUhnG0XnerjmUWb6eUWw4zev54sEfY5F3x002iQaW6cECAwEAAaOBkDCBjTAdBgNVHQ4EFgQU4CBUbZsS2GaNIkGRz/cBsD5ivjswUQYDVR0jBEowSIAU4CBUbZsS2GaNIkGRz/cBsD5ivjuhGqQYMBYxFDASBgNVBAMMC0Vhc3ktUlNBIENBghR8hE5uNY1QDiPFD/THwE4K8TZXDjAMBgNVHRMEBTADAQH/MAsGA1UdDwQEAwIBBjANBgkqhkiG9w0BAQsFAAOCAQEAKB3V4HIzoiO/Ch6WMj9bLJ2FGbpkMrcb/Eq01hT5zcfKD66lVS1MlK+cRL446Z2b2KDP1oFyVs+qmrmtdwrWgD+nfe2sBmmIHo9m9KygMkEOfG3MghGTEcS+0cTKEcoHYWYyOqQh6jnedXY8Cdm4GM1hAc9MiL3/sqV8YCVSLNnkoNysmr06/rZ0MCUZPGUtRmfd0heWhrfzAKw2HLgX+RAmpOE2MZqWcjvqKGyaRiaZks4nJkP6521aC2Lgp0HhCz1j8/uQ5ldoDszCnu/iro0NAsNtudTMD+YoLQxLqdleIh6CW+illc2VdXwj7mn6J04yns9jfE2jRjW/yTLFuQ==
```
mohammed90, line 40: Replace the DER with the variable `test_der_1` like you did below

mohammed90 (changes requested):
A few minor nitpicks and I believe I got everything

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -1218,6 +1222,31 @@ func (h *HTTPTransport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+				return d.Errf("module %s is not a caddytls.CA", modID)
+			}
+			if h.TLS == nil {
+				h.TLS = new(TLSConfig)
+			}
+			if len(h.TLS.RootCAPEMFiles) != 0 {
+				return d.Err("cannot specify both 'tls_trust_pool' and 'tls_trusted_ca_certs'")
+			}
+			if h.TLS.CARaw != nil {
+				return d.Err("cannot specify \"ca\" field twice in tls_trust_pool")
```
mohammed90, line 1246: They're setting `tls_trust_pool`, not `ca`. In Caddyfile format, they don't see the `ca` field of the JSON. They see `tls_trust_pool`.

mohammed90 (changes requested):

armadi1809:
@mohammed90 I think all your feedback items have been resolved Thanks for reviewing!

mohammed90 (approved):
LGTM! Waiting on @mholt review then we're good to go.

PR #6066, opened 2024-01-26
logging: Implement `log_append` handler

francislavoie:
Closes https://github.com/caddyserver/caddy/issues/5336

This adds a new handler `log_append` which adds an extra field to the access logs on the way back up the request chain (after most other handlers have run, including after handler errors).

The value may be a placeholder (if the value is surrounded by `{}`) or a vars key (will do a map lookup to see if it exists) or a constant string.

```
log_append [<matcher>] <key> <value>
```

Worth noting, `zap` doesn't validate that log fields don't overlap with existing ones when adding, so it's possible to have two log fields with the same name in the logs. For example, you might do `log_append status not-an-int` or whatever and you'd end up with `"status": 200, "status": "not-an-int"` in the logs. We could validate that the key is not one of the ones we have built-in, but I'm not sure if that's worth the effort?

### DIFF
--- a/modules/caddyhttp/logging/logadd.go
+++ b/modules/caddyhttp/logging/logadd.go
@@ -25,13 +25,13 @@ import (
 )
 
 func init() {
-	caddy.RegisterModule(ExtraLog{})
+	caddy.RegisterModule(LogAppend{})
 }
 
-// ExtraLog implements a middleware that takes a key and value, where
+// LogAppend implements a middleware that takes a key and value, where
 // the key is the name of a log field and the value is a placeholder,
 // or variable key, or constant value to use for that field.
-type ExtraLog struct {
+type LogAppend struct {
 	// Key is the name of the log field.
 	Key string `json:"key,omitempty"`
 
@@ -45,14 +45,14 @@ type ExtraLog struct {
 }
 
 // CaddyModule returns the Caddy module information.
-func (ExtraLog) CaddyModule() caddy.ModuleInfo {
+func (LogAppend) CaddyModule() caddy.ModuleInfo {
 	return caddy.ModuleInfo{
-		ID:  "http.handlers.extra_log",
-		New: func() caddy.Module { return new(ExtraLog) },
+		ID:  "http.handlers.log_append",
+		New: func() caddy.Module { return new(LogAppend) },
 	}
 }
 
-func (h ExtraLog) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
+func (h LogAppend) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
 	// Run the next handler in the chain first.
 	// If an error occurs, we still want to add
 	// any extra log fields that we can, so we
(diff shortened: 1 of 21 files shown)


steffenbusch:
Wow, I will love this feature

mholt:
Thanks for this!

Maybe bikeshedding, but what about `append_log`? Or `log_append`? or `log_field`?

francislavoie:
Could do `log_add`.

But we have `skip_log` right now. Maybe we should rename that too, to `log_skip` if we want consistency.

isaced:
Love this feature, which is used to add information similar to UserId/ProjectId to the logs without returning it to the client. 👍

https://caddy.community/t/logging-a-response-header-field-in-caddy-reverse-proxy-without-returning-to-client/22758

francislavoie:
We changed our mind again to `log_append`. Should be good to go now.

mholt:
As for

> We could validate that the key is not one of the ones we have built-in, but I'm not sure if that's worth the effort?

Probably not worth the effort. Different deserializers will handle it differently anyway.

francislavoie:
And it would probably harm performance to do that for little benefit. Unless we check it at provision time or something and warn if they used a "reserved" log key. Can be added later tho.

modules/caddyhttp/logging/logadd.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func init() {
+	caddy.RegisterModule(LogAdd{})
+}
+
+// LogAdd implements a middleware that takes a key and value, where
+// the key is the name of a log field and the value is a placeholder,
+// or variable key, or constant value to use for that field.
+type LogAdd struct {
```
mholt, line 34: Do we want to rename this to LogAppend as well? (right-click in VS Code, click Rename Symbol, boom)

francislavoie: Oh yeah oops, searched for `log_add` but not `logadd` :sweat_smile:

mholt (approved):
Thanks for working on this! :)

steffenbusch:
I have just built Caddy from the `master` branch in my personal testing environment to make some experiments with all the new features since 2.7.6. (I can't wait for 2.8.0 - you're doing a great job).

I was already aware that certain directives such as `basicauth` and `skip_log` were deprecated due to renaming, but I was also expecting that deprecation does not mean that existing Caddyfile configurations will not work and cause Errors. 

For `basicauth` no breaking behavior was observed, just the WARN logs that it has been renamed.
But `skip_log` occurrences will cause this error that the `skip_log` directive is not an ordered HTTP handler:

```
$ ./caddy_master validate --config ../Caddyfile
2024/03/07 18:04:12.432 INFO    using config from file  {"file": "../Caddyfile"}
2024/03/07 18:04:12.437 WARN    config.adapter.caddyfile        the 'basicauth' directive is deprecated, please use 'basic_auth' instead!
2024/03/07 18:04:12.438 WARN    config.adapter.caddyfile        the 'skip_log' directive is deprecated, please use 'log_skip' instead!
2024/03/07 18:04:12.438 WARN    config.adapter.caddyfile        the 'skip_log' directive is deprecated, please use 'log_skip' instead!
Error: adapting config using caddyfile: directive 'skip_log' is not an ordered HTTP handler, so it cannot be used here - try placing within a route block or using the order global option
```

Shall I open a separate issue for this?

francislavoie:
Oh oops, you're right, I should've left it in the directive order. If you want to make a quick PR to fix it, it would save me a few minutes :sweat_smile: just needs a `// TODO` comment next to it about the deprecation.

steffenbusch:
I am going to try to make this PR to fix it.

PR #6069, opened 2024-01-28
logging: Implement `append` encoder, allow flatter filters config

francislavoie:
Feature requested by @yroc92 

This adds a simple encoder called `append` which can append one or more log fields to all log entries that the logger processes.

This is inspired (and largely copy-paste) of the `filter` encoder, and as such it wraps another encoder (defaulting to `json` or `console` as per https://github.com/caddyserver/caddy/pull/5980).

Since this module does _not_ have access to the request context, it cannot use request placeholders, but it may use global placeholders such as the ones from https://caddyserver.com/docs/conventions#placeholders.

This is complementary to https://github.com/caddyserver/caddy/pull/6066 which allows appending extra log fields to _access logs_ only, and allows appending data from the request context.

Most commonly, this will probably be used to append things like a Caddy instance ID, to identify which instance the log was produced from in some log ingress tool.

For example, in global options:

```
{
	log {
		format append {
			fields {
				machine {env.MACHINE}
			}
		}
	}
}
```

And running with `MACHINE=foo caddy run`, a log like this is shown:
```
2024/01/28 06:39:51.145	INFO	serving initial configuration	{"machine": "foo"}
```

The key functionality of this encoder is in the `func (fe AppendEncoder) EncodeEntry` function (near the end of the file). The rest of the code is pretty much just zap boilerplate because we need to implement the `zapcore.Encoder` interface which has a _ton_ of functions :grimacing: we're just passing them through to the wrapped encoder.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
           GO_SEMVER: '~1.21.0'
 
         - go: '1.22'
-          GO_SEMVER: '~1.22.0'
+          GO_SEMVER: '~1.22.1'
 
         # Set some variables per OS, usable via ${{ matrix.VAR }}
         # OS_LABEL: the VM label from GitHub Actions (see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)
--- a/.github/workflows/cross-build.yml
+++ b/.github/workflows/cross-build.yml
@@ -35,7 +35,7 @@ jobs:
         # Set the minimum Go patch version for the given Go minor
         # Usable via ${{ matrix.GO_SEMVER }}
         - go: '1.22'
-          GO_SEMVER: '~1.22.0'
+          GO_SEMVER: '~1.22.1'
 
     runs-on: ubuntu-latest
     continue-on-error: true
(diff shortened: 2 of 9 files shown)


francislavoie:
I just added another commit which makes it possible to configure fields for both `filter` and `add` encoders in a flatter way, allowing to skip the `fields { }` wrapper. The wrapper is only needed if you ever have a field you want to call `wrap` or `fields` :sweat_smile: 

Before:
```
{
	log {
		format add {
			fields {
				machine {env.MACHINE}
			}
		}
	}
}
```

After:
```
{
	log {
		format add {
			machine {env.MACHINE}
		}
	}
}
```

yroc92 (approved):

mholt:
Oh -- lol, would it be prudent to rename this to `AppendEncoder` or would that be terribly annoying? :sweat_smile:

francislavoie:
Okay :+1:

.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           GO_SEMVER: '~1.21.0'
 
         - go: '1.22'
-          GO_SEMVER: '~1.22.0'
+          GO_SEMVER: '~1.22.1'
```
mohammed90, line 36: Francis,

image

mholt (approved):

PR #6078, opened 2024-01-31
Added support to disable output coloring from env variables

### LINKED ISSUE #6077: Add support for disabling ANSI color output
Add support for disabling ANSI color output by default or with the [`NO_COLOR='1'` environment variable](https://no-color.org).

ANSI color output does not play well with capturing the output to a text file nor is well supported in Windows.

armadi1809:
Closes #6077

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -717,7 +717,10 @@ func newDefaultProductionLogEncoder(wo WriterOpener) zapcore.Encoder {
 		encCfg.EncodeTime = func(ts time.Time, encoder zapcore.PrimitiveArrayEncoder) {
 			encoder.AppendString(ts.UTC().Format("2006/01/02 15:04:05.000"))
 		}
-		encCfg.EncodeLevel = zapcore.CapitalColorLevelEncoder
+		if coloringEnabled {
+			encCfg.EncodeLevel = zapcore.CapitalColorLevelEncoder
+		}
+
 		return zapcore.NewConsoleEncoder(encCfg)
 	}
 	return zapcore.NewJSONEncoder(encCfg)
@@ -758,6 +761,7 @@ func Log() *zap.Logger {
 }
 
 var (
+	coloringEnabled  = os.Getenv("NO_COLOR") == "" && os.Getenv("TERM") != "xterm-mono"
 	defaultLogger, _ = newDefaultProductionLog()
 	defaultLoggerMu  sync.RWMutex
 )


francislavoie:
I don't think this is right. Detecting color support is a lot more complicated than that.

We could consider using https://github.com/jwalton/go-supportscolor which is a pretty simple library which has robust detection heuristics. WDYT @mholt ?

mholt:
Sure, although that package doesn't seem to support `xterm_mono` (unless I missed something). Do we know if that package accounts for pipes and redirecting to files, etc?

What if we start with this, then change to using that package if we get more requests/complaints? I like that this is simple, conventional, and makes it really clear/obvious (to us) when color is used.

logging.go
```diff
@@ -758,6 +761,7 @@ func Log() *zap.Logger {
 }
 
 var (
+	coloringEnabled  = os.Getenv("NO_COLOR") != "1" && os.Getenv("TERM") != "xterm-mono"
```
mholt, line 764:
```suggestion
	coloringEnabled  = os.Getenv("NO_COLOR") != "" && os.Getenv("TERM") != "xterm-mono"
```

mholt (changes requested):
NO_COLOR can have any value, I think -- other than that, we can try this out I think, for starters!

francislavoie:
The only time color should be enabled is when we use stdout or stderr as the output module. In all other cases, it doesn't make sense. We would never enable it for files.

The lib is able to look at the stout/stderr handles to see if they support color.

logging.go
armadi1809: @mholt The example implementation on the no-color.org website linked in the issue shows that if the `NO_COLOR` is set to "0", we should set the coloring to false. I updated the code to follow this implementation. Let me know what you think about it and I can update again accordingly. 
`int
main(int argc, char *argv[])
{
	char *no_color = getenv("NO_COLOR");
	bool color = true;

	if (no_color != NULL && no_color[0] != '\0')
		color = false;

	/* do getopt(3) and/or config-file parsing to possibly turn color back on */
	...
}`

mholt: Oh, that's actually the null byte, if my C memory isn't underflowing me.

The actual verbiage says:

> Command-line software which adds ANSI color to its output by default should check for a NO_COLOR environment variable that, when present and not an empty string (regardless of its value), prevents the addition of ANSI color.

So presumably a value of 0 still counts as "set", but... that is kinda weird to do. I dunno why set the variable at all in that case. :man_shrugging: Maybe we just leave it as-is for now and see if anyone complains (with good reason for setting it to 0)?

armadi1809: Oh, I believe you're right, I did not see the text with it. My C knowledge is also very rusty at this point so I missed that. I will update it to conform to the specs.

mholt (approved):
I'd say this looks good to me now -- we can try it! Thank you @armadi1809

PR #6081, opened 2024-02-04
encode: Compress more media types by default

dunglas:
Add [all media types supported by Cloudflare](https://developers.cloudflare.com/speed/optimization/content/brotli/content-compression/#compression-between-cloudflare-and-website-visitors) to the default list of compressed media types.

This fixes a problem reported by @pborreli: by default JSON-LD responses generated by [API Platform](https://api-platform.com) aren't compressed.

I've kept Atom, which isn't supported by Cloudflare but is in our existing list (other existing values are also in Cloudflare's list).
I've kept the strategy of adding a wildcard at the end of each of the values (to allow all parameters, I suppose), but this may be too permissive. If we keep the wildcards, I wonder if building a regex wouldn't be faster than the current iteration. I'll try this and open another PR in the next few weeks if it improves performance.


mholt (approved):
Awesome, thanks!

modules/caddyhttp/encode/encode.go
```diff
@@ -84,17 +84,59 @@ func (enc *Encode) Provision(ctx caddy.Context) error {
...
+					"font/otf*",
+					"font/ttf*",
+					"font/x-woff*",
 					"image/svg+xml*",
+					"image/vnd.microsoft.icon*",
+					"image/x-icon*",
+					"multipart/bag*",
+					"multipart/mixed*",
+					"text/css*",
+					"text/javascript*",
```
mholt, line 130: So not just all `text/*` content types?

steffenbusch: > So not just all `text/*` content types?

If `text/csv` is no longer compressed by default, that would be a pity.

dunglas: I'll add it back. (I don't know why Cloudflare doesn't do it).

modules/caddyhttp/encode/encode.go
```diff
@@ -84,17 +84,49 @@ func (enc *Encode) Provision(ctx caddy.Context) error {
...
 				"Content-Type": []string{
-					"text/*",
-					"application/json*",
-					"application/javascript*",
-					"application/xhtml+xml*",
 					"application/atom+xml*",
+					"application/eot*",
+					"application/font*",
+					"application/font-sfnt*",
+					"application/font-woff*",
```
steffenbusch, line 96: Isn't `application/font-sfnt*` and `application/font-woff*`  redundant because the `application/font*` will match for those as well?

dunglas: Good catch. I'll update.

PR #6082, opened 2024-02-04
logging: Add support for additional logger filters other than hostname

### LINKED ISSUE #5978: [RFE] filter/split access logs by fields other than hostname
Hi!

Currently, I can create an access log like so:

```
    log {
        output file /opt/servers/server1/logs/caddy.access.log
    }
```

However, with many health checks coming in, I would redirect all known User-Agents for health checks to its own log file:

```
    log access_log {
        filter not @health-user-agents && not @health-url-paths
        output file /opt/servers/server1/logs/caddy.access.log
    }
    log health_check_log {
        filter @health-user-agents || @health-url-paths
        output file /opt/servers/server1/logs/caddy.access.health.log
    }
```

Of course this is pseudo syntax, but I think you get the idea.

For reference, in Apache httpd I can write something like this:

```
SetEnvIf Request_URI ".*\/health-check$" HealthPattern
# just add more here, they are or'd

CustomLog     /opt/servers/server1/logs/caddy.access.health.log    combined    env=HealthPattern
# and for access.log:  env!HealthPattern
```

armadi1809:
closes #5978 

@francislavoie This took some time but here is my initial implementation of this. Would like to get your feedback on this and your thoughts on what other tests should go in here? Thanks.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -914,7 +914,7 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 	// this is useful for setting up loggers per subdomain in a site block
 	// with a wildcard domain
 	customHostnames := []string{}
-
+	noLoggerNames := false
 	for h.NextBlock(0) {
 		switch h.Val() {
 		case "hostnames":
@@ -1000,6 +1000,12 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 				cl.Exclude = append(cl.Exclude, h.Val())
 			}
 
+		case "no_hostname":
+			if h.NextArg() {
+				return nil, h.ArgErr()
+			}
+			noLoggerNames = true
+
 		default:
 			return nil, h.Errf("unrecognized subdirective: %s", h.Val())
 		}
@@ -1007,7 +1013,7 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 
 	var val namedCustomLog
 	val.hostnames = customHostnames
-
+	val.noLoggerNames = noLoggerNames
 	isEmptyConfig := reflect.DeepEqual(cl, new(caddy.CustomLog))
 
 	// Skip handling of empty logging configs
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -51,6 +51,7 @@ func init() {
 	RegisterDirective("log", parseLog)
 	RegisterHandlerDirective("skip_log", parseLogSkip)
 	RegisterHandlerDirective("log_skip", parseLogSkip)
+	RegisterHandlerDirective("log_name", parseLogName)
 }
 
 // parseBind parses the bind directive. Syntax:
@@ -914,7 +915,7 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 	// this is useful for setting up loggers per subdomain in a site block
 	// with a wildcard domain
 	customHostnames := []string{}
-	noLoggerNames := false
+	noHostname := false
 	for h.NextBlock(0) {
 		switch h.Val() {
 		case "hostnames":
@@ -1004,7 +1005,7 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 			if h.NextArg() {
 				return nil, h.ArgErr()
 			}
-			noLoggerNames = true
+			noHostname = true
 
 		default:
 			return nil, h.Errf("unrecognized subdirective: %s", h.Val())
@@ -1013,7 +1014,7 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 
 	var val namedCustomLog
 	val.hostnames = customHostnames
-	val.noLoggerNames = noLoggerNames
+	val.noHostname = noHostname
 	isEmptyConfig := reflect.DeepEqual(cl, new(caddy.CustomLog))
 
 	// Skip handling of empty logging configs
@@ -1063,4 +1064,13 @@ func parseLogSkip(h Helper) (caddyhttp.MiddlewareHandler, error) {
 		return nil, h.ArgErr()
 	}
 	return caddyhttp.VarsMiddleware{"log_skip": true}, nil
+}
+
+// parseLogName parses the log_name directive. Syntax:
+//
+// log_name <names...>
+func parseLogName(h Helper) (caddyhttp.MiddlewareHandler, error) {
+	h.Next()
+	access_logger_names := h.RemainingArgs()
+	return caddyhttp.VarsMiddleware{"access_logger_names": access_logger_names}, nil
 }
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -1068,9 +1068,10 @@ func parseLogSkip(h Helper) (caddyhttp.MiddlewareHandler, error) {
 
 // parseLogName parses the log_name directive. Syntax:
 //
-// log_name <names...>
+//	log_name <names...>
 func parseLogName(h Helper) (caddyhttp.MiddlewareHandler, error) {
-	h.Next()
-	access_logger_names := h.RemainingArgs()
-	return caddyhttp.VarsMiddleware{"access_logger_names": access_logger_names}, nil
+	h.Next() // consume directive name
+	return caddyhttp.VarsMiddleware{
+		caddyhttp.AccessLoggerNameVarKey: h.RemainingArgs(),
+	}, nil
 }
--- a/caddytest/integration/caddyfile_adapt/log_filter_with_header.txt
+++ b/caddytest/integration/caddyfile_adapt/log_filter_with_header.txt
@@ -12,7 +12,7 @@ localhost {
 	}
 	@healthCheck `header_regexp('User-Agent', '^some-regexp$') || path('/healthz*')`
 	handle @healthCheck {
-		vars access_logger_names "health_check_log,general_log"
+		log_name "health_check_log" "general_log"
 		respond "Healthy"
 	}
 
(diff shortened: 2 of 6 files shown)


francislavoie:
Like I said in the issue, I'm not totally convinced this is the right way to do this. But I can't really think of anything better.

I think `no_hostname` is better than `no_host_name`, because `hostname` is typically seen as a single word in the context web servers (see https://en.wikipedia.org/wiki/Hostname for example)

I think there should be a `d.NextArg()` when parsing that option to reject the config if an extra argument is passed, since it takes none right now.

I'm not sure if that's the best name though, I'd like @mholt's opinion.

modules/caddyhttp/logging.go
```diff
@@ -58,8 +58,11 @@ type ServerLogConfig struct {
 }
 
 // wrapLogger wraps logger in a logger named according to user preferences for the given host.
-func (slc ServerLogConfig) wrapLogger(logger *zap.Logger, host string) *zap.Logger {
-	if loggerName := slc.getLoggerName(host); loggerName != "" {
+func (slc ServerLogConfig) wrapLogger(logger *zap.Logger, req *http.Request) *zap.Logger {
```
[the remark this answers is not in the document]
armadi1809: Yeah I agree that appending would go against the purpose of this feature, so overriding seems reasonable. We can wait until your PR gets merged and discuss how we can make `access_logger_name` take multiple values so that it fits in the multiple logger feature addition in your PR.

mholt:
@francislavoie @armadi1809  Let me know if you want to discuss anything to wrap up this PR. :+1:

armadi1809:
If I remember correctly this needs to wait for #6088 to be merged as it will require some changes after that's done right? @francislavoie

francislavoie:
I'm merging it now (once CI passes) 👍 you can rebase your branch and make the necessary adjustments.

modules/caddyhttp/logging.go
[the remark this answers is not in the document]
armadi1809: @francislavoie I went the naive route by making `access_logger_names` take multiple values seperated by a `,`. Do you recall what was your idea here and if we should rather implement it? I was thinking about tweaking the logic of the `UnmarshalCaddyFile` function of the `VarsMiddleware` but I wanted to get your opinion first.

francislavoie:
Needs another rebase, there's a conflict with another recent change of mine. Sorry!

armadi1809:
@francislavoie, rebase done. Any thoughts on the open discussion above?

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -915,7 +915,7 @@ func parseLogHelper(h Helper, globalLogNames map[string]struct{}) ([]ConfigValue
 	// this is useful for setting up loggers per subdomain in a site block
 	// with a wildcard domain
 	customHostnames := []string{}
-
+	noLoggerNames := false
```
francislavoie, line 918: I guess this variable name is misleading since we chose `no_hostname` as the config option

modules/caddyhttp/logging.go
```diff
@@ -69,16 +69,24 @@ type ServerLogConfig struct {
...
-
 	hosts := slc.getLoggerHosts(hostWithoutPort)
 	loggers := make([]*zap.Logger, 0, len(hosts))
+	if access_logger_names := GetVar(req.Context(), AccessLoggerNameVarKey); access_logger_names != nil {
+		accessLoggerNamesSlice := strings.Split(access_logger_names.(string), ",")
+		for _, loggerName := range accessLoggerNamesSlice {
+			loggers = append(loggers, logger.Named(loggerName))
+		}
+		return loggers
+	}
```
francislavoie, line 89: Normally we use `camelCase` for variable names. Anyway, we can shorten them because what it is is obvious from the source it's pulled from.

```suggestion
	if names := GetVar(req.Context(), AccessLoggerNameVarKey); names != nil {
		namesSlice := strings.Split(names.(string), ",")
		for _, loggerName := range namesSlice {
			loggers = append(loggers, logger.Named(loggerName))
		}
		return loggers
	}
```

caddytest/integration/caddyfile_adapt/log_filter_with_header.txt
```diff
@@ -0,0 +1,148 @@
...
+		output file ./caddy.access.health.log
+		no_hostname
+	}
+	log general_log {
+		output file ./caddy.access.general.log
+		no_hostname
+	}
+	@healthCheck `header_regexp('User-Agent', '^some-regexp$') || path('/healthz*')`
+	handle @healthCheck {
+		vars access_logger_names "health_check_log,general_log"
```
francislavoie, line 15: Oof, yeah this is weird. I don't think using a comma-separated value is ideal. Nowhere else in Caddy do we have this pattern.

@mholt I had an idea that we could do string slices for `vars` if we have the config like this:

```
vars access_logger_names [ health_check_log general_log ]
```

Basically if the first and last token are `[` `]` then we can take the list of args in between as a string slice type value.

I think we should make adjustments to `vars` to make it possible to have slices.

mholt: Hmm; I see what you mean. But it looks really bash-y. I wonder if we could support `["health_check_log", "general_log"]`...

I actually like how clean and simple the CSV is though. :man_shrugging:

francislavoie: Now I'm leaning towards adding a new directive `log_name <names...>` which would handle this case for us. That way users don't need to discover the `access_logger_names` var which is kinda a hack, they could just use `log_name` which would be documented like any other directive. This is similar to how `root` works, it's just a directive shortcut for a `vars` handler that sets a variable. I think it would need to be a real handler in JSON though because the `vars` handler still doesn't support an array of values.

Another long-term thing we could consider (out of scope of this PR but just thinking out loud), we could support a syntax like `vars +var_name value`, where `+` means "assume the var is a slice, and append a value to it". Could also add a `-` prefix to clear variable altogether. I dunno. There's room for improvements with the `vars` handler I think. Let's scrap the `[ ]` idea for now :joy:

armadi1809: @francislavoie Sorry for the delay! I didn't have access to a computer in the past week. I took a stab at implementing the new `log_name` directive. It seems to work fine with the vars handler, though. Am I missing something?

caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -1064,3 +1065,12 @@ func parseLogSkip(h Helper) (caddyhttp.MiddlewareHandler, error) {
...
 	return caddyhttp.VarsMiddleware{"log_skip": true}, nil
 }
+
+// parseLogName parses the log_name directive. Syntax:
+//
+// log_name <names...>
+func parseLogName(h Helper) (caddyhttp.MiddlewareHandler, error) {
+	h.Next()
+	access_logger_names := h.RemainingArgs()
+	return caddyhttp.VarsMiddleware{"access_logger_names": access_logger_names}, nil
```
francislavoie, line 1075: I think this should use the `AccessLoggerNameVarKey` const here

caddytest/integration/caddyfile_adapt/log_filter_with_header.txt
```diff
@@ -12,7 +12,7 @@ localhost {
 	}
 	@healthCheck `header_regexp('User-Agent', '^some-regexp$') || path('/healthz*')`
 	handle @healthCheck {
-		vars access_logger_names "health_check_log,general_log"
+		log_name "health_check_log" "general_log"
```
francislavoie, line 15: Can drop the double-quotes here (doesn't really matter but just saying)

caddytest/integration/caddyfile_adapt/log_filter_with_header.txt
```diff
@@ -89,7 +89,10 @@ localhost {
 														{
 															"handle": [
 																{
-																	"access_logger_names": "health_check_log,general_log",
+																	"access_logger_names": [
+																		"health_check_log",
+																		"general_log"
+																	],
```
francislavoie, line 95: Oh good, I wasn't sure if the `vars` handler would just take an array as-is. That's nice :+1:

modules/caddyhttp/logging.go
```diff
@@ -80,10 +80,11 @@ func (slc ServerLogConfig) wrapLogger(logger *zap.Logger, req *http.Request) []*
 	}
 	hosts := slc.getLoggerHosts(hostWithoutPort)
 	loggers := make([]*zap.Logger, 0, len(hosts))
-	if access_logger_names := GetVar(req.Context(), AccessLoggerNameVarKey); access_logger_names != nil {
-		accessLoggerNamesSlice := strings.Split(access_logger_names.(string), ",")
-		for _, loggerName := range accessLoggerNamesSlice {
-			loggers = append(loggers, logger.Named(loggerName))
+	if names := GetVar(req.Context(), AccessLoggerNameVarKey); names != nil {
+		if namesSlice, ok := names.([]any); ok {
```
francislavoie, line 84: Should this type assert as `[]string` instead? Then you don't need to do `.(string)` later, I think.

francislavoie: I'm wrong, that doesn't work. Go is silly, can't cast directly to a slice type.

francislavoie:
I did a commit to do a final bit of cleanup and refactoring. I think this is good to go now.

francislavoie (approved):

mholt:
Awesome, thank you both!!

steffenbusch:
Just wanted to say THANK YOU for this awesome new feature. I have just tested it with Caddy v2.8.0-rc and my configuration looked like this (maybe this helps for other users, viewers of this MR or for the caddy docs website).

```caddy
example.com {
        log example.com

        log health_check_log {
                output file /var/log/caddy/access-health-checks.json {
                        roll_size 50m
                        roll_local_time
                        roll_keep_for 30d
                }
                format filter {
                        wrap json
                        fields {
                                request>tls>version tls_version TLSv
                                request>tls>cipher_suite tls_cipher
                        }
                }
                no_hostname
        }

	@monitoring_ohdear_uptime {
		# Two possible user-agent headers:
		# Mozilla/5.0 (compatible; OhDear/1.1; +https://ohdear.app/checker; brokenLinks)
		# OhDear.app (+https://ohdear.app/docs/checks/uptime)
		header_regexp user-agent `^.*\+https:\/\/ohdear\.app.*$`
		header_regexp X-External-Monitoring ^redacted_secret$
	}

        # Before Caddy v.2.8.0 these requests were skipped; now commented
        #log_skip @monitoring_ohdear_uptime

        # Now with Caddy v2.8.0-rc.1 these requests are written into a dedicated file 👍 
        log_name @monitoring_ohdear_uptime health_check_log
 ...
```

PR #6085, opened 2024-02-07
matchers: Drop `forwarded` option from `remote_ip` matcher

francislavoie:
I think it's been deprecated long enough. Time to remove this insecure option!

### DIFF
--- a/modules/caddyhttp/ip_matchers.go
+++ b/modules/caddyhttp/ip_matchers.go
@@ -37,13 +37,6 @@ type MatchRemoteIP struct {
 	// The IPs or CIDR ranges to match.
 	Ranges []string `json:"ranges,omitempty"`
 
-	// If true, prefer the first IP in the request's X-Forwarded-For
-	// header, if present, rather than the immediate peer's IP, as
-	// the reference IP against which to match. Note that it is easy
-	// to spoof request headers. Default: false
-	// DEPRECATED: This is insecure, MatchClientIP should be used instead.
-	Forwarded bool `json:"forwarded,omitempty"`
-
 	// cidrs and zones vars should aligned always in the same
 	// length and indexes for matching later
 	cidrs  []*netip.Prefix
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/ip_matchers.go
```diff
@@ -126,11 +115,7 @@ func (MatchRemoteIP) CELLibrary(ctx caddy.Context) (cel.Library, error) {
 
 			for _, input := range strList.([]string) {
 				if input == "forwarded" {
-					if len(m.Ranges) > 0 {
-						return nil, errors.New("if used, 'forwarded' must be first argument")
-					}
-					m.Forwarded = true
-					continue
+					return nil, errors.New("the 'forwarded' option is no longer supported; use the 'client_ip' matcher instead")
```
mholt, line 118: Wonder if we will remember to remove this in a while as well

modules/caddyhttp/ip_matchers.go
```diff
@@ -82,11 +75,7 @@ func (m *MatchRemoteIP) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	d.Next() // consume matcher name
 	for d.NextArg() {
 		if d.Val() == "forwarded" {
-			if len(m.Ranges) > 0 {
-				return d.Err("if used, 'forwarded' must be first argument")
-			}
-			m.Forwarded = true
-			continue
+			return d.Err("the 'forwarded' option is no longer supported; use the 'client_ip' matcher instead")
```
mholt, line 78: Wonder if we will remember to remove this in a while

mholt (approved):
Sounds good to me. Since we don't have telemetry I don't know how many are still using it, but it has been quite some time.

modules/caddyhttp/ip_matchers.go
francislavoie: Meh, this is just to give a nicer error message just in case they upgrade from an older version. It can stay forever, it's harmless.

PR #6088, opened 2024-02-09
caddyhttp: Support multiple logger names per host

francislavoie:
Context: https://caddy.community/t/config-cleanup-for-logging-blocking-and-a-signalr-issue/22643

I realized that there's currently no good way to point a specific hostname to write to more than one logger at once, because `logger_names` is a simple `map[string]string` (so only one value per key).

To solve this, ~~I'm deprecating `logger_names` and adding a new `logger_mapping` (which Caddyfile immediately uses)~~ I'm changing `logger_names` to be a `map[string]StringArray` which is a type which accepts either a JSON string or array (where a string gets turned into a one-element string slice). Then, I reworked the logging code in the server to loop over the list of wrapped loggers.

~~I'm not 100% convinced about the `logger_mapping` name... I'd rather reuse `logger_names`, but obviously we can't do that, it would be a breaking change. Any better naming ideas?~~ Resolved by doing custom JSON unmarshaling.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -804,10 +804,10 @@ func (st *ServerType) serversFromPairings(
 				} else if len(ncl.hostnames) > 0 {
 					// if the logger overrides the hostnames, map that to the logger name
 					for _, h := range ncl.hostnames {
-						if srv.Logs.LoggerMapping == nil {
-							srv.Logs.LoggerMapping = make(map[string][]string)
+						if srv.Logs.LoggerNames == nil {
+							srv.Logs.LoggerNames = make(map[string]caddyhttp.StringArray)
 						}
-						srv.Logs.LoggerMapping[h] = append(srv.Logs.LoggerMapping[h], ncl.name)
+						srv.Logs.LoggerNames[h] = append(srv.Logs.LoggerNames[h], ncl.name)
 					}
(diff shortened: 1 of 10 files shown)


mohammed90:
> I'd rather reuse logger_names, but obviously we can't do that, it would be a breaking change. Any better naming ideas?

We can do custom unmarshaling by implementing the `JSONUnmarshal` (or is it better to use `TextUnmarshal`? 🤔) method. In it we'll check if it can be marshalled into a `map[string][]string` or `map[string]string`, reporting a warning when marshalling into a `map[string]string`.

francislavoie:
I guess so. 🤔 Most of the time that approach feels wrong but in this case it can be transparent so it's probably fine.

francislavoie:
Thanks for the idea @mohammed90! Much better I think.

francislavoie:
The user tried it out and confirmed it resolves their usecase:

https://caddy.community/t/config-cleanup-for-logging-blocking-and-a-signalr-issue/22643/12

mohammed90:
> Thanks for the idea @mohammed90! Much better I think.

Great 😃 I have to update the nginx-adapter after this PR

mholt (approved):
This looks quite complicated to me :see_no_evil: but if it works and is useful for real world use cases, I guess that's how it goes (my fault for the logging design). Thank you for the tricky patch!!

PR #6089, opened 2024-02-09
chore: Rename CI jobs, run on M1 mac

francislavoie:
We can test on M1 mac now!

But to do so, we need to change `macos-latest` to `macos-14` for now, as per https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

So for that reason, I decided to reorganize our CI config a bit, and the result is that the job names are now simplified.

- `ubuntu-latest` -> `linux`
- `macos-latest` -> `mac`
- `windows-latest` -> `windows`

See the Actions. Looks cleaner I think.

Before merging @mholt we'll need to update the repo's required jobs config again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: 
           - ubuntu-latest
-          - macos-latest
+          - macos-14
           - windows-latest
         go: 
           - '1.21'
@@ -42,7 +42,7 @@ jobs:
           CADDY_BIN_PATH: ./cmd/caddy/caddy
           SUCCESS: 0
 
-        - os: macos-latest
+        - os: macos-14
           CADDY_BIN_PATH: ./cmd/caddy/caddy
           SUCCESS: 0
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,9 @@ jobs:
       fail-fast: false
       matrix:
         os: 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -125,7 +129,7 @@ jobs:
 
     # To return the correct result even though we set 'continue-on-error: true'
     # - name: Coerce correct build result
-    #   if: matrix.os != 'windows-latest' && steps.step_test.outputs.status != ${{ matrix.SUCCESS }}
+    #   if: matrix.os != 'windows' && steps.step_test.outputs.status != ${{ matrix.SUCCESS }}
     #   run: |
     #     echo "step_test ${{ steps.step_test.outputs.status }}\n"
     #     exit 1
```
mohammed90, line 135: Do we need this? I forget why we have it commented

mohammed90 (approved):
Looks great 💯 Thanks for doing this

mholt (approved):
Awesome, that's more elegant. Thanks! Branch rules updated.

.github/workflows/ci.yml
francislavoie: I think it was for test coverage reports? We've never really turned them on since v2 betas I think. I'm not sure if we ever will. But still, harmless in case we eventually do, but also we can find it pretty easily in the commit logs if we need to bring it back. So up to you 🤷‍♂️

PR #6092, opened 2024-02-10
caddyauth: Rename `basicauth` to `basic_auth`

francislavoie:
This allows `basic_auth` to work as well in the Caddyfile, and emits a deprecation warning when `basicauth` is used.

This was essentially a mistake in v2.0, we made the decision to use underscores for directives that are multiple words like `file_server` etc, but we left `basicauth`. Since then, we've added `forward_auth`, so `basicauth` feels incongruent with that.

I think now's as good a time as any to realign this.


mholt (approved):
"Angry upvote" :laughing:

(Mainly because I'm kicking myself for that inconsistency when v2 launched)

PR #6098, opened 2024-02-11
caddyfile: Reject long heredoc markers

francislavoie:
Closes #6099

PR #6100, opened 2024-02-12
Revert "caddyfile: Reject long heredoc markers"

mohammed90:
Reverts caddyserver/caddy#6098

The panics are freaking out the fuzzer and keeps reporting errors/bugs due to the constant panics it hits with the long heredocs. We need to find a better solution, but for now we need to revert the commit.

image


mholt (approved):

francislavoie:
Dangit 😣 fuzzers are silly.

Formatter should allow returning an error I think.

PR #6101, opened 2024-02-12
caddytls: Require email for ZeroSSL issuer

mholt:
This patch is in preparation for upcoming changes that will help prevent abuse without imposing strict rate limits.

My current approach is simply to return an error if an email isn't specified.

Right now I am pretty aggressive about this: I return an error at provision-time if an email isn't found, and this will actually break a lot of deployments (not strictly speaking: as long as they're using `caddy reload` or the API their sites will stay up, but their new configs will need to add an email in order to run successfully). This is actually preferable if they are _relying_ on ZeroSSL for certs. (Requiring an email address is a low price to effectively remove rate limits.)

I also return an error at account-create time, which is probably more effective, as this can allow deployments to continue without having to add their email address, they just lose the redundancy of having ZeroSSL. If they are also using Let's Encrypt, then this will not be noticeable since they will keep using Let's Encrypt.

(Frankly, using another CA just to circumvent LE rate limits is a brittle plan anyway, as it typically represents a shortcut taken in the design of one's infrastructure.)

I'm thinking of maybe removing the provision-time error and turning it into a warning to soften the blow, at least at first.

Another approach we could take is to not even wire up the ZeroSSL issuer (implicitly) if there's no email address provided. Looking at the code, this might be more complicated (or it's not, I didn't spend much time on this approach yet).

Open question remains: Are existing ACME accounts going to be blocked if they don't have an email? I'll try to find out.

### DIFF
--- a/modules/caddytls/zerosslissuer.go
+++ b/modules/caddytls/zerosslissuer.go
@@ -74,7 +74,14 @@ func (iss *ZeroSSLIssuer) Provision(ctx caddy.Context) error {
 	if iss.ACMEIssuer.CA == "" {
 		iss.ACMEIssuer.CA = certmagic.ZeroSSLProductionCA
 	}
-	return iss.ACMEIssuer.Provision(ctx)
+	if err := iss.ACMEIssuer.Provision(ctx); err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


mholt:
Superseded by #6229

PR #6103, opened 2024-02-13
tests: uses testing.TB interface

jcchavezs:
This would allow us to run the test server for benchmarks too.

### DIFF
--- a/caddytest/caddytest.go
+++ b/caddytest/caddytest.go
@@ -60,11 +60,11 @@ var (
 type Tester struct {
 	Client       *http.Client
 	configLoaded bool
-	t            *testing.T
+	t            testing.TB
 }
 
 // NewTester will create a new testing client with an attached cookie jar
-func NewTester(t *testing.T) *Tester {
+func NewTester(t testing.TB) *Tester {
 	jar, err := cookiejar.New(nil)
 	if err != nil {
 		t.Fatalf("failed to create cookiejar: %s", err)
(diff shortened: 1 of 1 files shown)


jcchavezs:
Ping @mholt

jcchavezs:
Ping @mholt @francislavoie

francislavoie:
This is a backwards compatible change, right? Because `TB` is a superset of `T`?

jcchavezs:
Yes, it won't break anything because both *testing.T and *testing.B implements that interface.

francislavoie:
Actually you did change it to not be a pointer. I think this might be a breaking change for plugin authors.

I'm not against it, I just want us to be knowingly making that break if it is breaking.

jcchavezs:
testing.TB is an interface, whoever uses this would have called

```
var t *testing.T
NewTester(t)
```

which remains working as expected. I changed the parameter type of the function to a superset so not a breaking change.

jcchavezs:
I wanted to run a benchmark using this and then eneded up not doing it because it does not accept testing.B

mohammed90 (approved):
I think this is fine. The interface TB contains the common methods of T, B, and F for this purpose. Also there's compile-time check in Go source code to ensure *T implements TB, so we have the guarantee there.

https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/testing/testing.go;l=905-906

jcchavezs:
any chance you will cut a patch release anytime soon?

José Carlos Chávez


tir. 20. feb. 2024 kl. 23:04 skrev Mohammed Al Sahaf <
***@***.***>:

francislavoie:
We still have a lot of PRs we went to review and merge before 2.8.0. No planned release time right now. You can pin to the commit on master for now if you need.

PR #6104, opened 2024-02-14
caddyfile: Reject directives in the place of site addresses

### LINKED ISSUE #6095: Better Caddyfile errors when a directive is used at top-level
A common mistake users make when writing their Caddyfile is that they'll place directives at the top-level scope (outside a site block) based on a misunderstanding of the [Caddyfile structure](https://caddyserver.com/docs/caddyfile/concepts#structure)

I've been thinking about ways to mitigate this in such a way that doesn't limit possible configurations.

I think what we could do during Caddyfile parsing, if we see a site address that matches a registered directive, we should return an error, explaining that directives must go within a site block. Unknown directives will not produce an error.

The downside to this is "what if the user wants to serve a domain literally called `handle` for some reason"? I think as long as we don't throw an error if they use `https://handle` (with a scheme), then there exists a valid escape-hatch and we don't need to worry about it.

For example, this Caddyfile would produce the following confusing error (because obviously `lb_policy` isn't a directive, why does Caddy think it is one?):

```
reverse_proxy localhost:9000 localhost:9001 {
    lb_policy first
}
```
```
Error: Caddyfile:2: unrecognized directive: lb_policy
```

An even worse example, this Caddyfile _doesn't_ produce an error because things within `handle` _are_ valid directives... but `handle` was parsed as a domain name:

```
handle {
	file_server
}
```

Both of those should produce an error like:

```
Error: Caddyfile:1: parsed 'handle' as a site address, but it is a known directive; directives must appear in a site block 
```

The exact wording of the error can be adjusted, I'm not sure how best to word it right now; open for discussion.

armadi1809:
Resolves #6095  

@francislavoie from my understanding, that's the only change needed but I might be wrong. Also, let me know if you think we need more tests.

### DIFF
--- a/caddyconfig/caddyfile/parse.go
+++ b/caddyconfig/caddyfile/parse.go
@@ -160,13 +160,13 @@ func (p *parser) begin() error {
 	}
 
 	if ok, name := p.isNamedRoute(); ok {
-		// named routes only have one key, the route name
-		p.block.Keys = []string{name}
-		p.block.IsNamedRoute = true
-
 		// we just need a dummy leading token to ease parsing later
 		nameToken := p.Token()
 		nameToken.Text = name
+
+		// named routes only have one key, the route name
+		p.block.Keys = []Token{nameToken}
+		p.block.IsNamedRoute = true
 
 		// get all the tokens from the block, including the braces
 		tokens, err := p.blockTokens(true)
@@ -211,10 +211,11 @@ func (p *parser) addresses() error {
 	var expectingAnother bool
 
 	for {
-		tkn := p.Val()
+		value := p.Val()
+		token := p.Token()
 
 		// special case: import directive replaces tokens during parse-time
-		if tkn == "import" && p.isNewLine() {
+		if value == "import" && p.isNewLine() {
 			err := p.doImport(0)
 			if err != nil {
 				return err
@@ -223,9 +224,9 @@ func (p *parser) addresses() error {
 		}
 
 		// Open brace definitely indicates end of addresses
-		if tkn == "{" {
+		if value == "{" {
 			if expectingAnother {
-				return p.Errf("Expected another address but had '%s' - check for extra comma", tkn)
+				return p.Errf("Expected another address but had '%s' - check for extra comma", value)
 			}
 			// Mark this server block as being defined with braces.
 			// This is used to provide a better error message when
@@ -237,15 +238,15 @@ func (p *parser) addresses() error {
 		}
 
 		// Users commonly forget to place a space between the address and the '{'
-		if strings.HasSuffix(tkn, "{") {
-			return p.Errf("Site addresses cannot end with a curly brace: '%s' - put a space between the token and the brace", tkn)
+		if strings.HasSuffix(value, "{") {
+			return p.Errf("Site addresses cannot end with a curly brace: '%s' - put a space between the token and the brace", value)
 		}
 
-		if tkn != "" { // empty token possible if user typed ""
+		if value != "" { // empty token possible if user typed ""
 			// Trailing comma indicates another address will follow, which
 			// may possibly be on the next line
-			if tkn[len(tkn)-1] == ',' {
-				tkn = tkn[:len(tkn)-1]
+			if value[len(value)-1] == ',' {
+				value = value[:len(value)-1]
 				expectingAnother = true
 			} else {
 				expectingAnother = false // but we may still see another one on this line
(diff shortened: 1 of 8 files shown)


francislavoie:
Yeah I think we should have a test that ensures `http://handle` is still allowed. I think it will be because directives aren't registered with a scheme, but just a sanity check.

armadi1809:
@francislavoie I get your point with the filename and line number but I am not entirely sure about where to move this new check since we need to loop through the block keys to determine if we are using a directive as a site address but we only have access to the `Tokens` at the segment level. Maybe I am missing something here though?

francislavoie:
I think we could either rework `ServerBlock` to hold `[]Token` instead of `[]string` for the site addresses (to retain the file/line along with the text), or move the errors to `func (p *parser) addresses() error {` where it actually parses the addresses initially.

armadi1809:
I think I like the second option better as it feels more intuitive to perform the check there. One issue is that we don't have access to the `registeredDirectives` at the `caddyfile` package which makes me wonder if the only solution is reworking `ServerBlock`s. Unless there is another way of accessing all the caddy known directives.

francislavoie:
Ah right I forgot about that :cry: yeah it'll probably be better to pass down the tokens I guess. We can add some helper methods to ServerBlock to unwrap the tokens to a `[]string` if necessary.

caddyconfig/caddyfile/parse.go
```diff
@@ -212,7 +212,7 @@ func (p *parser) addresses() error {
 
 	for {
 		tkn := p.Val()
-
+		token := p.Token()
```
francislavoie, line 215: I think we should rename `tkn` to "value" or something, just to differentiate it from `token`. Kinda confusing that they have different types with essentially the same name.

caddyconfig/httpcaddyfile/httptype.go
```diff
@@ -65,11 +65,11 @@ func (st ServerType) Setup(
 	originalServerBlocks := make([]serverBlock, 0, len(inputServerBlocks))
 	for _, sblock := range inputServerBlocks {
 		for j, k := range sblock.Keys {
-			if j == 0 && strings.HasPrefix(k, "@") {
-				return nil, warnings, fmt.Errorf("cannot define a matcher outside of a site block: '%s'", k)
+			if j == 0 && strings.HasPrefix(k.Text, "@") {
+				return nil, warnings, fmt.Errorf("cannot define a matcher outside of a site block: '%s'", k.Text)
 			}
```
francislavoie, line 70: Let's add the file/line here as well

francislavoie:
Tests are failing - looks like we need to add the `","` trim in another spot.

armadi1809:
@francislavoie I think we should be good to go now. I changed the location of that `Trim` to be where we actually add the token rather than in the helper that maps the tokens to strings. I also addressed your comments from above. Thanks for your feedback and help.

caddyconfig/caddyfile/parse.go
```diff
@@ -211,10 +210,10 @@ func (p *parser) addresses() error {
 	var expectingAnother bool
 
 	for {
-		tkn := p.Val()
-
+		value := p.Val()
+		token := p.Token()
 		// special case: import directive replaces tokens during parse-time
```
francislavoie, line 215: Nit: the newline was there to visually separate the sections

```suggestion
		value := p.Val()
		token := p.Token()

		// special case: import directive replaces tokens during parse-time
```

caddyconfig/caddyfile/parse.go
```diff
@@ -254,11 +253,12 @@ func (p *parser) addresses() error {
...
 			// between their two domains, e.g. "foo.com,bar.com", which would not be
 			// parsed as two separate site addresses.
-			if strings.Contains(tkn, ",") {
-				return p.Errf("Site addresses cannot contain a comma ',': '%s' - put a space after the comma to separate site addresses", tkn)
+			if strings.Contains(value, ",") {
+				return p.Errf("Site addresses cannot contain a comma ',': '%s' - put a space after the comma to separate site addresses", value)
 			}
 
-			p.block.Keys = append(p.block.Keys, tkn)
+			token.Text = strings.Trim(token.Text, ",")
```
francislavoie, line 260: I don't think we need to trim here, since `value` already has the comma removed from the end. We can just do `token.Text = value` I think.

caddytest/integration/caddyfile_adapt/http_valid_directive_like_site_address.txt
```diff
@@ -0,0 +1,46 @@
+http://handle {
+				file_server
```
francislavoie, line 2: 4 tabs -> 1 tab :sweat_smile: 

```suggestion
	file_server
```

armadi1809:
All the feedback items from above should now be resolved.

francislavoie (approved):
Thanks, this is great!

I made a few minor test adjustments, i.e. the error now asserts that the file/line is present, and I changed a couple other tests to fix some port conflicts that happened on my machine due to the tests using common ports (admin 2019 but I already have Caddy running, and 8080 is commonly used for some HTTP apps).

PR #6105, opened 2024-02-14
admin: Allow origins header value to not have a scheme

francislavoie:
Currently when `enforce_origin` is set, the request header must have `http://` otherwise the header value gets stuffed in the URL's `Path` field. Our code only looks at Host and not Path, so this is at minimum a less confusing result if a configured origin is passed without a scheme.


mholt:
(Context: We discussed this at length in Slack, and decided this change isn't quite what we want. The Origin header MUST have a scheme.)

PR #6108, opened 2024-02-15
vars: Allow overriding `http.auth.user.id` in replacer as a special case

francislavoie:
Context: https://caddy.community/t/logging-a-response-header-field-in-caddy-reverse-proxy-without-returning-to-client/22758

I was thinking, it would probably be useful to allow users to fill `http.auth.user.id` for themselves using `vars`, because then it would show up in the access logs in the dedicated `user_id` field we already have.

Example config:

```
:8881 {
    log
    header -X-User-Id
    reverse_proxy localhost:8883 {
        handle_response {
            vars http.auth.user.id {rp.header.X-User-Id}
            copy_response
        }
    }
}

:8883 {
    header X-User-Id foobar
    respond "yo"
}
```

Example log, notice `user_id` is filled:
```
2024/02/15 11:30:00.866	INFO	http.log.access	handled request	{
	"request": {"remote_ip": "127.0.0.1", "remote_port": "46870", "client_ip": "127.0.0.1", "proto": "HTTP/1.1", "method": "GET", "host": "localhost:8881", "uri": "/", "headers": {"User-Agent": ["curl/7.81.0"], "Accept": ["*/*"]}},
	"bytes_read": 0,
	"user_id": "foobar",
	"duration": 0.001003354,
	"size": 2,
	"status": 200,
	"resp_headers": {"Content-Length": ["2"], "Content-Type": ["text/plain; charset=utf-8"], "Date": ["Thu, 15 Feb 2024 11:30:00 GMT"], "Server": ["Caddy", "Caddy"]}
}
```


mholt (approved):
I don't love special cases, but the benefit is compelling... let's try it :)

mholt:
Huh, I'm having trouble getting the s390x tests to pass here. @francislavoie

mohammed90:
> Huh, I'm having trouble getting the s390x tests to pass here. @francislavoie --- must be flaky, because they passed for the auto-merge.

My fault. Will fix it in a PR.

PR #6109, opened 2024-02-15
fileserver: Preserve query during canonicalization redirect

francislavoie:
Replaces https://github.com/caddyserver/caddy/pull/5254

Context: https://caddy.community/t/caddy-stripping-querystring/22754

When a canonicalization redirect happens, it's only taking the path and adding a `/`, which loses the original query if there was one.

This preserves it safely, without making additional modifications to the URL.


mholt (approved):
Nice -- I this is a simple fix, assuming `to` consists solely of a path (or at least, doesn't have a query string). (I wonder if we should add a comment somewhere to mention that?)

francislavoie:
Yeah we only ever use it with the path currently, so it's fine. I'll add a comment.

mholt (approved):

PR #6113, opened 2024-02-18
httpcaddyfile: Allow nameless regexp placeholder shorthand

francislavoie:
I just read the code for regexp matchers, and saw this:

```go
	if mre.Name != "" {
		mre.phPrefix += "." + mre.Name
	}
```

This made me realize that we're always setting regexp placeholders, even if a name isn't specified... but our Caddyfile placeholder shorthand wasn't allowing `{re.1}`, it was only shortening long-form placeholders! Silly oversight...

Now, this is possible:

```
@match path_regexp ^/foo(.*)$
respond @match "{re.1}"
```

Specifying a name is only necessary if there's more than one matcher in the set (e.g. header_regexp + path_regexp at the same time), or if you want to carry the matcher result for later use if there's a risk that another regexp matcher might be used. But for the majority of cases, the simple case with no name is perfectly fine.


mholt (approved):
Ah nice, thanks for fixing this.

Issue #6111, opened 2024-02-18
Feature Addition: Enhance reverse_proxy module with support for a specified transport proxy


ImpostorKeanu:
For the `http.reverse_proxy` module, having the option of specifying a HTTP/SOCKS proxy for the `http.Transport` of the reverse proxy client in the configuration file would be ideal. It's currently configurable only via `http.ProxyFromEnvironment` and may impact functionality of other modules.

I've implemented the proposed changes and they seem to work well but I'm new to contributing to the project so there's a chance that they're naive. I'll submit a PR for review.

P.S. Caddy is amazing. It's become a critical component in so many of my projects and I'm very grateful for all the thoughtful hard work. ❤️

francislavoie:
In case you didn't realize, you can use the `HTTP_PROXY` env var. That's what `ProxyFromEnvironment` does.

ImpostorKeanu:
I've ran into situations where particular modules need to use distinct HTTP proxies. Is this possible `http.ProxyFromEnvironment`?

francislavoie:
You mean different proxies per `reverse_proxy` in your config? No I don't think that's possible.

Thinking back, I think ~1 year ago+ @mohammed90 @mholt and I had a discussion about this stuff... I think one problem is that we have HTTP clients all over Caddy in different places, like the ACME client, reverse proxy, etc which means we'd probably want a centralized place to configure all of them. But it would be complicated to have to push down that config to relevant places.

The question is whether we want to only have a URL as config, or if we might want to make it modular (i.e. allow plugins to choose the proxy). I don't know what kind of usecases there would be to having it modular though, I've never needed a forward proxy myself.

ImpostorKeanu:
Yes, that's what I've been coming up against. There has been at least [one other discussion](https://caddy.community/t/reverse-proxy-use-upstream-https-proxy/21326/6) where someone was trying to achieve the same but it seems the discussion died.

My use case is being able to reverse engineer applications during penetration tests. I've found it useful to be able to send the entire request chain through Burp for inspection/analysis. While `HTTP_PROXY` works, adjacent `reverse_proxy` configurations are also proxying through Burp when they ideally wouldn't.

```mermaid
flowchart
ua["User Agent"] --> burp[Burp] --> caddy["Caddy (Reverse Proxy)"] --> burp --> inscope[In-Scope Wapp]
```

Admittedly this isn't a common use case, but I decided to propose the change after seeing that at least one other person had a similar issue.

[...]

PR #6114, opened 2024-02-19
reverseproxy: Configurable forward proxy URL

ImpostorKeanu:
Closes: #6111 

Apologies for the redo. I'm security person by trade so I often fumble through git.

This PR should incorporate corrections discussed in the last PR that I closed.

Grateful for the patience.

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -83,7 +83,7 @@ type HTTPTransport struct {
 	//
 	// User Agent ->
 	//  reverse_proxy ->
-	//  transport_proxy_url -> upstream
+	//  forward_proxy_url -> upstream
 	//
 	// Default: off.
 	ForwardProxyURL string `json:"forward_proxy_url,omitempty"`
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -23,6 +23,7 @@ import (
 	weakrand "math/rand"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -71,6 +72,22 @@ type HTTPTransport struct {
...
+	//
+	// Providing a value to this parameter results in
+	// requests flowing through the reverse_proxy in the following
+	// way:
+	//
+	// User Agent ->
+	//  reverse_proxy ->
+	//  forward_proxy_url -> upstream
+	//
+	// Default: off.
```
WeidiDeng, line 88: Default is not `off`. It's read from environmental variables.

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -265,8 +282,23 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
...
 
+	// negotiate any HTTP/SOCKS proxy for the HTTP transport
+	var proxy func(*http.Request) (*url.URL, error)
+	if h.ForwardProxyURL != "" {
+		pUrl, err := url.Parse(h.ForwardProxyURL)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse transport proxy url: %v", err)
+		}
+		caddyCtx.Logger().Info("setting transport proxy url", zap.String("url", h.ForwardProxyURL))
+		proxy = func(request *http.Request) (*url.URL, error) {
```
WeidiDeng, line 293: Should just use [`http.ProxyURL`](https://pkg.go.dev/net/http#ProxyURL).

WeidiDeng (changes requested):

modules/caddyhttp/reverseproxy/httptransport.go
ImpostorKeanu: The intent is for the individual module to be able to be configured with a specific proxy. `ForwardProxyURL` takes precedence over `HTTP_PROXY` and no default value is supplied.

francislavoie: But `if h.ForwardProxyURL == ""` it still uses `http.ProxyFromEnvironment` therefore the default is still "from environment", not nothing.

mholt (approved):
Thanks for the contribution! Is this ready to merge? It LGTM.

WeidiDeng (commented):
Maybe Caddyfile support should be added? For now it's only configurable through JSON, which is very tedious.

WeidiDeng (approved):

PR #6116, opened 2024-02-20
chore: use context.WithoutCancel instead of ignoreClientGoneContext

WeidiDeng:
Now that the minimal version of go is 1.21, we can safely use `context.WithoutCancel`.


mholt (approved):
Yay! Thank you Weidi.

Issue #6096, opened 2024-02-11
Query rewrite operations


francislavoie:
Currently, the only way we offer to rewrite the query part of the request URI is with the `rewrite` handler (when it has a `?` in the target), or with `uri replace` to perform replacements on the entire URI.

These options are limiting, because many query rewrite operations are difficult to perform with those.

For example, it's impossible (I can't think of a way) to remove a single query key from a request without knowing all the other possible query keys and re-including them all with `&key={query.key}` in the rewrite.

It's also unwieldy to replace a specific query key with another (one way is with `uri replace` but it's awkward and error-prone).

I think the solution is to add `uri query` which supports multiple different types of operations on queries, similarly to the [`header` directive](https://caddyserver.com/docs/caddyfile/directives/header) which has `+` and `-` and so on.

```
# add an additional `foo=bar` (even if there was already `foo=baz`)
uri query +foo bar

# overwrite `foo` with `foo=bar`
uri query foo bar

# delete `foo`
uri query -foo

# rename key `foo` to `bar`
uri query foo>bar

# set a default value (if the key is not already set)
uri query ?foo bar

# replace values in `foo`, changing `bar` to `baz`
uri query foo bar baz

# do multiple operations at once
uri query {
	foo bar
	bar baz
	-baz
}
```

We may or may not actually implement all these operations, but I think most of them would be very nice to have.

Should this be its own directive instead of being added to `uri`? To be discussed.

mholt:
I really like this idea. As part of the `uri` directive, yeah.

francislavoie:
I'll keep this open actually, until we have the `replace` operation implemented. /cc @armadi1809

francislavoie:
I don't quite remember what that entails, so I'll let you make a decision on that, or you can ask specifically what we should include/keep

mholt:
@armadi1809 Yes, I think that'd be great! Both fast substring replacements and regex replacements. Ideally for all params or just a named one. Feel free to use the headers directive as a starting point.

PR #6120, opened 2024-02-21
rewrite: Implement `uri query` operations

armadi1809:
@francislavoie This is my Initial take at #6096. This is still to be improved as only add, delete and set are implemented, but I wanted to get your initial feedback on the overall architecture of the solution before implementing the remaining operations.

### DIFF
--- a/modules/caddyhttp/rewrite/rewrite.go
+++ b/modules/caddyhttp/rewrite/rewrite.go
@@ -89,8 +89,9 @@ type Rewrite struct {
 	// Performs regular expression replacements on the URI path.
 	PathRegexp []*regexReplacer `json:"path_regexp,omitempty"`
 
-	// Performs query parameters operations on the URI path
-	QueryOperations *queryOps `json:"query_operations,omitempty"`
+	// Mutates the query string of the URI.
+
+	Query *queryOps `json:"query_operations,omitempty"`
 
 	logger *zap.Logger
 }
@@ -273,8 +274,8 @@ func (rewr Rewrite) Rewrite(r *http.Request, repl *caddy.Replacer) bool {
 	}
 
 	// apply query operations
-	if rewr.QueryOperations != nil {
-		rewr.QueryOperations.do(r, repl)
+	if rewr.Query != nil {
+		rewr.Query.do(r, repl)
 	}
 
 	// update the encoded copy of the URI
@@ -536,6 +537,7 @@ func (q *queryOps) do(r *http.Request, repl *caddy.Replacer) {
 }
 
 type queryOpsArguments struct {
+	// A key in the query string. Note that query string keys may appear multiple times.
 	Key string `json:"key,omitempty"`
 	Val string `json:"val,omitempty"`
 }
--- a/modules/caddyhttp/rewrite/rewrite.go
+++ b/modules/caddyhttp/rewrite/rewrite.go
@@ -486,6 +486,9 @@ type queryOps struct {
 	// Sets query parameters; replaces existing header fields.
 	Set []queryOpsArguments `json:"set,omitempty"`
 
+	// Renames query keys
+	Rename []queryOpsArguments `json:"rename,omitempty"`
+
 	// Names of query parameters to delete
 	Delete []string `json:"delete,omitempty"`
 }
@@ -517,6 +520,16 @@ func (q *queryOps) do(r *http.Request, repl *caddy.Replacer) {
 		}
 		val := repl.ReplaceAll(setParam.Val, "")
 		query[key] = []string{val}
+	}
+
+	for _, renameParam := range q.Rename {
+		key := repl.ReplaceAll(renameParam.Key, "")
+		val := repl.ReplaceAll(renameParam.Val, "")
+		if key == "" || val == "" {
+			continue
+		}
+		query[val] = query[key]
+		delete(query, key)
 	}
 
 	r.URL.RawQuery = query.Encode()
--- a/modules/caddyhttp/rewrite/rewrite.go
+++ b/modules/caddyhttp/rewrite/rewrite.go
@@ -89,6 +89,9 @@ type Rewrite struct {
 	// Performs regular expression replacements on the URI path.
 	PathRegexp []*regexReplacer `json:"path_regexp,omitempty"`
 
+	// Performs query parameters operations on the URI path
+	QueryOperations *queryOps `json:"query_operations,omitempty"`
+
 	logger *zap.Logger
 }
 
@@ -267,6 +270,11 @@ func (rewr Rewrite) Rewrite(r *http.Request, repl *caddy.Replacer) bool {
 	// regular expression replacements on the path
 	for _, rep := range rewr.PathRegexp {
 		rep.do(r, repl)
+	}
+
+	// apply query operations
+	if rewr.QueryOperations != nil {
+		rewr.QueryOperations.do(r)
 	}
 
 	// update the encoded copy of the URI
@@ -468,6 +476,45 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
 	if req.URL.RawPath == req.URL.Path {
 		req.URL.RawPath = ""
 	}
+}
+
+// queryOps describes the operations to perform on a query like add, delete and replace fields
+type queryOps struct {
+	// Adds query parameters; does not replace any existing header fields.
+	Add []queryOpsArguments `json:"add,omitempty"`
+
+	// Sets query parameters; replaces existing header fields.
+	Set []queryOpsArguments `json:"set,omitempty"`
+
+	// Names of query parameters to delete
+	Delete []string `json:"delete,omitempty"`
+}
+
+func (q *queryOps) do(r *http.Request) {
+	query := r.URL.Query()
+
+	for _, addParam := range q.Add {
+		key := addParam.Key
+		val := addParam.Val
+		query[key] = append(query[key], val)
+	}
+
+	for _, deleteParam := range q.Delete {
+		delete(query, deleteParam)
+	}
+
+	for _, setParam := range q.Set {
+		key := setParam.Key
+		val := setParam.Val
+		query[key] = []string{val}
+	}
+
+	r.URL.RawQuery = query.Encode()
+}
+
+type queryOpsArguments struct {
+	Key string `json:"key,omitempty"`
+	Val string `json:"val,omitempty"`
 }
 
 // Interface guard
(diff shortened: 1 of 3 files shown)


francislavoie:
Yeah, good start.Couple things I notice off the bat:

- I think it should support a block (i.e. braces) to do multiple operations with a single handler, similar to the header directive

- We need to apply the replacer on the query values (and maybe the keys too? Probably doesn't hurt). Users are usually not going to use static values, more often than not it's going to be dynamic using placeholders.

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +478,45 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+
+	// Names of query parameters to delete
+	Delete []string `json:"delete,omitempty"`
+}
+
+func (q *queryOps) do(r *http.Request, repl *caddy.Replacer) {
+	query := r.URL.Query()
+
+	for _, addParam := range q.Add {
+		key := repl.ReplaceAll(addParam.Key, "")
```
francislavoie, line 497: I think maybe if the key is empty it should do nothing. I don't think it's valid to have an empty query key? Need to check the HTTP spec to make sure but I'm pretty sure it's invalid.

I think it might also be invalid to have `=` in the key, but maybe Go encodes that? Can cover with a test.

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +478,45 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+		val := repl.ReplaceAll(addParam.Val, "")
+		query[key] = append(query[key], val)
+	}
+
+	for _, deleteParam := range q.Delete {
+		param := repl.ReplaceAll(deleteParam, "")
```
francislavoie, line 503: Same, maybe need to check for empty

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +478,45 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+		delete(query, param)
+	}
+
+	for _, setParam := range q.Set {
+		key := repl.ReplaceAll(setParam.Key, "")
```
francislavoie, line 508: Same

modules/caddyhttp/rewrite/rewrite.go
armadi1809: I just skipped the iteration if the key is empty. I also added a key with an `=` sign to the url in the test I wrote and it looks like go encodes that.

modules/caddyhttp/rewrite/rewrite.go
armadi1809: Same as above.

modules/caddyhttp/rewrite/rewrite.go
armadi1809: Same above.

armadi1809:
@francislavoie, out of the list of operations in the issue, which ones you think we should add on top of what we have here at the moment?

francislavoie:
I think rename would be good to have. The defaulting operation isn't necessary. Value replacement would also be good to have.

We don't have to implement those right away though, we can add them in a follow-up.

armadi1809:
I jus added rename. We can add replace later as it looked a little bit more involved (especially if we want to use regex like for the headers). Please let me know if any modifications are still necessary in this PR.

mholt (changes requested):
This is looking really good! :) Nice work. Just a couple nits from my first pass. Probably ready to move out of draft state.

armadi1809:
@mholt Thanks for the round of review! All your feedback items should now be resolved. I also removed the draft status from the PR.

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +479,68 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
 	}
 }
 
+// queryOps describes the operations to perform on a query like add, delete and replace fields
```
francislavoie, line 482:
```suggestion
// queryOps describes the operations to perform on query keys: add, set, rename and delete.
```

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +479,68 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+type queryOps struct {
+	// Adds query parameters; does not replace any existing header fields.
```
francislavoie, line 484:
```suggestion
	// Adds query parameters; does not overwrite an existing query field,
	// and only appends an additional value for that key if any already exist.
```

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +479,68 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+	Add []queryOpsArguments `json:"add,omitempty"`
+
+	// Sets query parameters; replaces existing header fields.
```
francislavoie, line 487:
```suggestion
	// Sets query parameters; overwrites a query key with the given value.
```

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +479,68 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+	Set []queryOpsArguments `json:"set,omitempty"`
+
+	// Renames query keys
```
francislavoie, line 490:
```suggestion
	// Renames a query key from Key to Val, without affecting the value.
```

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +479,68 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+	Rename []queryOpsArguments `json:"rename,omitempty"`
+
+	// Names of query parameters to delete
```
francislavoie, line 493:
```suggestion
	// Deletes a given query key by name.
```

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +479,68 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+		delete(query, key)
+	}
+
+	r.URL.RawQuery = query.Encode()
+}
+
+type queryOpsArguments struct {
+	// A key in the query string. Note that query string keys may appear multiple times.
+	Key string `json:"key,omitempty"`
+	Val string `json:"val,omitempty"`
```
francislavoie, line 542:
```suggestion
	// The value for the given operation; for add and set, this is
	// simply the value of the query, and for rename this is the
	// query key to rename to.
	Val string `json:"val,omitempty"`
```

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -89,6 +89,10 @@ type Rewrite struct {
 	// Performs regular expression replacements on the URI path.
 	PathRegexp []*regexReplacer `json:"path_regexp,omitempty"`
 
+	// Mutates the query string of the URI.
+
+	Query *queryOps `json:"query_operations,omitempty"`
```
francislavoie, line 94: Shouldn't be a space in between, otherwise the comment won't be seen as a Godoc comment when generating the docs.

Also I think we should just shorten this to `query`, the `_operations` doesn't really add much to the config output. The fact that we're doing an "operation" is clear enough by the verbs like "add/set/rename/delete".

```suggestion
	// Mutates the query of the URI by applying a set of operations.
	Query *queryOps `json:"query,omitempty"`
```

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +479,68 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+	Rename []queryOpsArguments `json:"rename,omitempty"`
+
+	// Names of query parameters to delete
+	Delete []string `json:"delete,omitempty"`
+}
+
+func (q *queryOps) do(r *http.Request, repl *caddy.Replacer) {
+	query := r.URL.Query()
+
+	for _, addParam := range q.Add {
```
francislavoie, line 500: I think we should think about the order in which we apply the query operations, such that it's done in the most useful order.

I think that `delete` should probably always be applies last, because it's the "most destructive". That would allow a prior operation to copy a value from a particular key. For example you might have `?foo=bar` and you want to result in `?bar=foobar`, you might do "set bar foo{query.foo}" then "delete foo".

I think rename should probably be first, because then you'd expect to be able to use the new name immediately in subsequent operations.

And finally I think set should probably be before rename, because it overwrites a key anyway, so if you were to add then set in that order, only the set would remain, but doing set then add would result in both values existing.

Then I think we could probably use a few tests to make sure the above :point_up: work as expected.

WDYT @mholt ?

francislavoie (changes requested):

modules/caddyhttp/rewrite/rewrite.go
armadi1809: @francislavoie I re ordered the operations as suggested above and I also added tests to cover the different use cases you described.  I will leave this discussion open for now in case we decide to change anything.

armadi1809:
All the discussions above (except for the order of operations) should now be resolved.

modules/caddyhttp/rewrite/rewrite.go
```diff
@@ -470,5 +478,73 @@ func changePath(req *http.Request, newVal func(pathOrRawPath string) string) {
...
+	Add []queryOpsArguments `json:"add,omitempty"`
+
+	// Sets query parameters; overwrites a query key with the given value.
+	Set []queryOpsArguments `json:"set,omitempty"`
+
+	// Renames a query key from Key to Val, without affecting the value.
+	Rename []queryOpsArguments `json:"rename,omitempty"`
+
+	// Deletes a given query key by name.
+	Delete []string `json:"delete,omitempty"`
```
mholt, line 494: What if we put these fields in the order of the operations? So first Rename, then Set, then Add, then Delete?

I think I agree with the reasoning for that order.

armadi1809: The fields should be in the right order now.

mholt (approved):
I think this is good to go -- let's give it a shot! Nice work @armadi1809 :)

francislavoie (approved):

PR #6137, opened 2024-03-01
Include command usage in the help output

SerheyDolgushev:
FYI @dunglas

### DIFF
--- a/cmd/cobra.go
+++ b/cmd/cobra.go
@@ -119,7 +119,7 @@ func caddyCmdToCobra(caddyCmd Command) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   caddyCmd.Name,
 		Short: caddyCmd.Short,
-		Long:  caddyCmd.Usage + "\n\n" + caddyCmd.Long,
+		Long:  caddyCmd.Usage + caddyCmd.Long,
 	}
 	if caddyCmd.CobraFunc != nil {
 		caddyCmd.CobraFunc(cmd)
--- a/cmd/cobra.go
+++ b/cmd/cobra.go
@@ -119,7 +119,7 @@ func caddyCmdToCobra(caddyCmd Command) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   caddyCmd.Name,
 		Short: caddyCmd.Short,
-		Long:  caddyCmd.Long,
+		Long:  caddyCmd.Usage + "\n\n" + caddyCmd.Long,
 	}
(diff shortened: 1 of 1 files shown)


cmd/cobra.go
```diff
@@ -119,7 +119,7 @@ func caddyCmdToCobra(caddyCmd Command) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   caddyCmd.Name,
 		Short: caddyCmd.Short,
-		Long:  caddyCmd.Long,
+		Long:  caddyCmd.Usage + "\n\n" + caddyCmd.Long,
```
dunglas, line 122: Line breaks don't look necessary because "Long" strings already include them.

```suggestion
		Long:  caddyCmd.Usage + caddyCmd.Long,
```

dunglas (approved):
LGTM but I don't know very well this part so let's wait for other reviews :)

mohammed90:
What is this patch trying to solve? Usage is already printed on `caddy --help` and `caddy help`. It's right above `Examples`.

SerheyDolgushev:
> What is this patch trying to solve? Usage is already printed on `caddy --help` and `caddy help`. It's right above `Examples`.

The root cause for this PR originates from https://github.com/dunglas/frankenphp. It's `php-server` command has `Usage`: https://github.com/dunglas/frankenphp/blob/main/caddy/php-server.go#L32. But it is not shown as output of `~/frankenphp php-server -h`.

Please note, my caddy experience is very limited, and I am not sure it is the right place for this PR, maybe this change needs to be made on `frankenphp` side? If so, please advice.

Decided to submit this PR to start the conversation.

mohammed90:
> > What is this patch trying to solve? Usage is already printed on `caddy --help` and `caddy help`. It's right above `Examples`.
> 
> The root cause for this PR originates from https://github.com/dunglas/frankenphp. It's `php-server` command has `Usage`: https://github.com/dunglas/frankenphp/blob/main/caddy/php-server.go#L32. But it is not shown as output of `~/frankenphp php-server -h`.
> 
> Please note, my caddy experience is very limited, and I am not sure it is the right place for this PR, maybe this change needs to be made on `frankenphp` side? If so, please advice.
> 
> Decided to submit this PR to start the conversation.

Thanks to you, I realized our `Usage` has always been outputting just `[flags]`. We were using Cobra incorrectly. I've sent the proper fix in PR #6138.

Thank you so much! It helped identify the issue

mholt:
Mohammed, you're awesome. Thank you both for addressing this!

PR #6140, opened 2024-03-01
encode: Add `zstd` compression level support

dev-polymer:
Its changes adds support `zstd` encoder ability to spicify compression level like is able for Gzip (and for brotli via [ueffel/caddy-brotli module](https://github.com/ueffel/caddy-brotli)).

The solution appeared as a continuation of [this conversation on the caddy community](https://caddy.community/t/zstd-encode-compression-level-setup/22930/3) forum/

Notices:
- Only for levels 1,2,3,4 supported beacuse its [only supported in used zstd library](https://pkg.go.dev/github.com/klauspost/compress@v1.17.7/zstd#EncoderLevel) and focusing on the fact that in encode `gzip` and `brotli` we can set up a different number of levels, I consider a suitable solution with the ability to configure only 4 levels
- Without explicitly specifying a level or specifying 0, the default level 2 will be selected
- Buffer size `zstd.WithWindowSize(128<<10)` is hadrcoded as before and when i try delete and trust the automatic buffer selection in the library algorithm — i have compression error in Chrome: `ERR_CONTENT_DECODING` on CSS files. I think no hardcode in buffer size could be logical, but if it not stabe working, it seems like a good idea to move buffer size feature to a separate task, but i make test with hardcoded `zstd.WithWindowSize(512<<10)` value just for intrest, and it is working ok:
image
...and compress even better, but my knowledge of compression is insufficient for an informed choice of best level.

I have been running tests of this solution on a fully functional multipage website in the Chrome v122 browser.


modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -27,7 +30,9 @@ func init() {
 }
 
 // Zstd can create Zstandard encoders.
-type Zstd struct{}
+type Zstd struct {
+	Level zstd.EncoderLevel `json:"level,omitempty"`
```
francislavoie, line 34: I think we should add a godoc comment here to explain the possible options.

francislavoie:
Please remember to sign the CLA - you might need to update your emails on your github account to match your commit, or you should update your local git config to use the correct email, and rewrite the commit's email.

modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -51,7 +85,7 @@ func (z Zstd) NewEncoder() encode.Encoder {
 	// The default of 8MB for the window is
 	// too large for many clients, so we limit
 	// it to 128K to lighten their load.
-	writer, _ := zstd.NewWriter(nil, zstd.WithWindowSize(128<<10), zstd.WithEncoderConcurrency(1), zstd.WithZeroFrames(true))
+	writer, _ := zstd.NewWriter(nil, zstd.WithWindowSize(128<<10), zstd.WithEncoderConcurrency(1), zstd.WithZeroFrames(true), zstd.WithEncoderLevel(z.Level))
```
francislavoie, line 88: This line's kinda long, let's adjust the syntax

```suggestion
	writer, _ := zstd.NewWriter(
		nil,
		zstd.WithWindowSize(128<<10), 
		zstd.WithEncoderConcurrency(1), 
		zstd.WithZeroFrames(true), 
		zstd.WithEncoderLevel(z.Level),
	)
```

modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -51,7 +85,13 @@ func (z Zstd) NewEncoder() encode.Encoder {
...
 	// too large for many clients, so we limit
 	// it to 128K to lighten their load.
-	writer, _ := zstd.NewWriter(nil, zstd.WithWindowSize(128<<10), zstd.WithEncoderConcurrency(1), zstd.WithZeroFrames(true))
+	writer, _ := zstd.NewWriter(
+		nil,
+		zstd.WithWindowSize(128<<10), 
+		zstd.WithEncoderConcurrency(1), 
+		zstd.WithZeroFrames(true), 
+		zstd.WithEncoderLevel(z.Level),
+	)
```
francislavoie, line 94: Oops my bad, I left in some spaces

```suggestion
	writer, _ := zstd.NewWriter(
		nil,
		zstd.WithWindowSize(128<<10),
		zstd.WithEncoderConcurrency(1),
		zstd.WithZeroFrames(true),
		zstd.WithEncoderLevel(z.Level),
	)
```

modules/caddyhttp/encode/zstd/zstd.go
WeidiDeng: Generally we should built-in or caddy defined type to configure a plugin. We should use string to specify the level and check if it's one of the possible values provided by zstd.

modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -38,7 +43,36 @@ func (Zstd) CaddyModule() caddy.ModuleInfo {
...
 
 // UnmarshalCaddyfile sets up the handler from Caddyfile tokens.
-func (z *Zstd) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+func (g *Zstd) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.Next() // consume option name
+	if !d.NextArg() {
+		return nil
+	}
+	levelStr := d.Val()
+	level, err := strconv.Atoi(levelStr)
```
WeidiDeng, line 52: zstd use constants for a reason. Shouldn't treat them as int.

modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -38,7 +43,36 @@ func (Zstd) CaddyModule() caddy.ModuleInfo {
...
+func (g *Zstd) Provision(ctx caddy.Context) error {
+	if g.Level == 0 {
+		g.Level = zstd.SpeedDefault
+	}
+	return nil
+}
+
+// Validate validates g's configuration.
+func (g Zstd) Validate() error {
+	if g.Level < zstd.SpeedFastest {
```
WeidiDeng, line 70: Although their values are from 1 to 4, we shouldn't rely on this behaviour. They are defined as [iota](https://github.com/klauspost/compress/blob/v1.17.7/zstd/encoder_options.go#L148).

WeidiDeng: The reason why gzip can check for number range is because the level is given explicitly as a number. But zstd only provides 4 constants.

WeidiDeng (changes requested):
To convert string to levels we can use this [function](https://github.com/klauspost/compress/blob/v1.17.7/zstd/encoder_options.go#L176).

mholt:
I can do a review as soon as the CLA is signed :blush:

francislavoie (approved):
Thanks! We can merge once @WeidiDeng's feedback is addressed

modules/caddyhttp/encode/zstd/zstd.go
francislavoie: How would you suggest we do it then @weidideng?

dev-polymer: You are mostly right, but let me some explaination. I mean usage of text constants not matches caddyfile's code-style and other encoder algo's compression levels (`1-9` for `gzip` and `1-11` for `brotli`).

I think about another realization, and propose for appeal to original `zstd` levels, not library approximation. It is [1 to 22](https://github.com/facebook/zstd/blob/8689633fdf5dabfc7f4cd6f79611335c3f8def88/lib/zstd.h#L82C11-L82C12) (or [1 to 19, excludes --ultra levels](https://github.com/facebook/zstd/blob/8689633fdf5dabfc7f4cd6f79611335c3f8def88/contrib/pzstd/Options.cpp#L61) which recommended to carefull usage).

But if we implement [library's way to transform](https://github.com/klauspost/compress/blob/1e2b27500c4eae851ba753ad116dc5a6524d2366/zstd/encoder_options.go#L188) caddyfile's levels to `zstd` levels, we will have the option to specify the level 6,7,8,9 and it will be the same value (for the current version of the used library). And it will surprise the user and it will seem that changing the level does not work (in some ranges). 

Perhaps such a solution will be more universal for the long term, but now it will introduce a little confusion and crutchiness in the configuration. And even more then little: if I specify level 9 today, I will get a conditionally `N%` CPU load, then if the library implements level 9 to more powerfull compression, perhaps the CPU load will increase by a multiple, for example `3 x N%`, which may be an unpleasant surprise for a user who has updated the caddy in the future.

This is reason, why i chose to offer a version with 4 levels...

Finally, i think the best solution would be allow user to specify levels from 1 to 22 (refer to original [Zstandart levels](https://github.com/facebook/zstd/blob/8689633fdf5dabfc7f4cd6f79611335c3f8def88/lib/zstd.h#L82C11-L82C12)), convert it with [library's method](https://github.com/klauspost/compress/blob/1e2b27500c4eae851ba753ad116dc5a6524d2366/zstd/encoder_options.go#L188) and may be we may need to specify in the documentation (or even caddy warning logs) that the recommended and future safe operating `zstd` levels should be 1, 3, 6 and 10 now.

WDYT?

francislavoie: I think we can use the strings/named options for zstd, I don't think that's weird. We can simply document them. We use named options in many places in the Caddyfile.

dev-polymer: I pushed [new commit](https://github.com/caddyserver/caddy/pull/6140/commits/0a91062a5a169243a039ba6d19a72d240915dc20) now, but my tests show error when level is not set in CaddyFile:
```
//Test results (with default 128kb zstd buffer) of 585kb minified uncompressed CSS-file
zstd (level not set) → ERR_HTTP2_PROTOCOL_ERROR (Chrome v122 browser error)
zstd fastest → 337kb 
zstd default → 294kb
zstd better → 290kb
zstd best → 286kb
```
Please friends suggest how to fix 😳

modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -27,7 +27,10 @@ func init() {
 }
 
 // Zstd can create Zstandard encoders.
-type Zstd struct{}
+type Zstd struct {
+	// Compression level refer to type constants value from zstd.SpeedFastest to zstd.SpeedBestCompression
```
francislavoie, line 31: Users reading the JSON docs won't know how to look at those constants. We should just list them out in the comment here for users to know what they can set.

WeidiDeng: We should just list these strings `fastest`, `default`, `better`, `best`. Users won't look at the source code to find out which string correspond to which level.

modules/caddyhttp/encode/zstd/zstd.go
WeidiDeng: `Level` should be just string. We can use a private member to specify the actual level used with type `zstd.EncoderLevel`.

WeidiDeng (changes requested):

modules/caddyhttp/encode/zstd/zstd.go
francislavoie: Were you reading the wrong commit? It has this now:

```
// List of all available levels: fastest, default, better, best
```

modules/caddyhttp/encode/zstd/zstd.go
francislavoie: Probably need to set a default in `Provision`

modules/caddyhttp/encode/zstd/zstd.go
WeidiDeng: My bad 😉

modules/caddyhttp/encode/zstd/zstd.go
```diff
@@ -39,6 +44,27 @@ func (Zstd) CaddyModule() caddy.ModuleInfo {
...
+		)
+	}
+
+	z.Level = levelStr
+	return nil
+}
+
+// Provision provisions z's configuration.
+func (z *Zstd) Provision(ctx caddy.Context) error {
+	_, z.level = zstd.EncoderLevelFromString(z.Level)
```
WeidiDeng, line 67: We actually need to check if the level is correct here. Power users will use JSON to configure caddy plugins.

We need to check if the level string is valid in 2 places, `UnmarshalCaddyfile` and `Provision`, or `Validate` if you prefer to implementing another interface.

Also there needs to be a default value in case it's not set.

WeidiDeng (changes requested):

modules/caddyhttp/encode/zstd/zstd.go
mholt: Made a quick commit to do this

mholt (approved):
@WeidiDeng How about now? I'm good with this if you are. :+1:

WeidiDeng (approved):

polarathene:
> * Only for levels 1,2,3,4 supported beacuse its [only supported in used zstd library](https://pkg.go.dev/github.com/klauspost/compress@v1.17.7/zstd#EncoderLevel) and focusing on the fact that in encode `gzip` and `brotli` we can set up a different number of levels, I consider a suitable solution with the ability to configure only 4 levels

Will that documentation be visible/linked for the equivalent Caddy docs on the values? For those that are familiar with zstd elsewhere it's useful to have some context on what they map to without having to hunt for the information on Github :)

---

> * Buffer size `zstd.WithWindowSize(128<<10)` is hadrcoded as before and when i try delete and trust the automatic buffer selection in the library algorithm — i have compression error in Chrome: `ERR_CONTENT_DECODING` on CSS files. I think no hardcode in buffer size could be logical, but if it not stabe working, it seems like a good idea to move buffer size feature to a separate task, but i make test with hardcoded `zstd.WithWindowSize(512<<10)` value just for intrest, and it is working ok

This may vary by client, but Chrome is following guidance from RFC 8878 to limit the buffer window size to 8MB: https://issues.chromium.org/issues/40196713#comment31

> _Prevent OoM condition by restricting maximum window size_
> 
> _Unlike gzip or Brotli where the maximum window size is fixed (**32KB
for zlib and 16MB for brotli**, see RFC1952 and RFC7932), zstd
allows much bigger window sizes (up to 3.75TB, see RFC8878)._
> 
> _This potentially exposes a user agent if the response to a request
leverages a payload with a request for a very large window descriptor._
>
> _Fortunately, the RFC8878  authors already addressed the potential issue
by allowing a user agent to 'reject a compressed frame that requires
a memory size beyond the decoder's authorized range._
> 
> _The current change remedies this issue by setting a maximum allowed
window size, **limiting the buffer size to 8MB as suggested by RFC8878**._

mholt:
@polarathene 

> Will that documentation be visible/linked for the equivalent Caddy docs on the values? For those that are familiar with zstd elsewhere it's useful to have some context on what they map to without having to hunt for the information on Github :)

Yes, our docs will explain what the possible values are, but the zstd lib we use recommends we use only those constants as the underlying meaning may change :man_shrugging:

polarathene:
> zstd lib we use recommends we use only those constants as the underlying meaning may change 🤷‍♂️

Ah ok I see that after looking into the package docs. It [supports int levels that are more familiar for zstd config](https://pkg.go.dev/github.com/klauspost/compress@v1.17.7/zstd#EncoderLevelFromZstd) but maps to the nearest supported constant mode the package offers instead.

This was already [discussed earlier above](https://github.com/caddyserver/caddy/pull/6140#discussion_r1515093486), but I disagree with the conclusion since the constant modes are also subject to change just as much, so the ordinal mapping which can be more familiar should be roughly equivalent if it's maintained alongside those changes? 🤷‍♂️ 

It's true that the ordinal mapping would be a bit confusing without that knowledge, due to the mapped ranges producing the same output and potential for future updates to shift those range bounds. I think linking to [this part of the package README](https://github.com/klauspost/compress/tree/master/zstd#status) might be relevant to get an approximation of the ordinal mapping that each constant mode is roughly equivalent to (_assuming it's kept up to date_).

---

The README reference also follows with a performance comparison to the CGO package (`datadog/zstd`) which shows notable overhead from `klauspost/compress` zstd implementation.

I've [seen for brotli there are two external packages/modules](https://caddy.community/t/gzip-zstd-br-priority-order/22737/20), [one that is pure Go](https://github.com/ueffel/caddy-brotli) and [another that requires CGO enabled](https://github.com/dunglas/caddy-cbrotli) during build, but I'm not sure how feasible / compatible a zstd CGO alternative would be since unlike brotli support Caddy has official zstd already with the encode directive 😅

PoneyClairDeLune:
https://caddyserver.com/docs/caddyfile/directives/encode

The Caddyfile documentation for the `encode` directive wasn't updated.

mholt:
Thanks for the heads up. Just pushed the docs for it.

PR #6145, opened 2024-03-03
caddyfile: Populate regexp matcher names by default

francislavoie:
This is something I've been mulling over for months, I was trying to think of the right solution to solve this, and I think I finally just realized "duh, just add a context map to the dispenser".

The idea is that often users might define a named matcher with a regexp like this:

```
@oldposts path_regexp oldposts ^/post/(.*)$
redir @oldposts /posts/{re.oldposts.1}
```

We wrote `oldposts` a total of four times :grimacing: What if we instead we could omit having to name the regexp, when we've already named the named matcher? With this PR, now this works:

```
@oldposts path_regexp ^/post/(.*)$
redir @oldposts /posts/{re.oldposts.1}
```

Anyway, this is kinda complementary to https://github.com/caddyserver/caddy/pull/6113 but totally independent. With that one, this is possible, by not even using the matcher name at all:

```
@oldposts path_regexp ^/post/(.*)$
redir @oldposts /posts/{re.1}
```

Which is fine for simple cases where the regexp is used immediate after matching and there's no risk of another regexp matcher clobbering it.

Worth noting, because of how CEL matchers are compiled & provisioned at runtime, and not at Caddyfile-adapt time, this feature doesn't work for passing names into expression matchers right now. It might be possible to add the matcher name to the `expression` matcher's JSON as well, so that at runtime it can add it to the context when provisioning. Not sure if that's worth doing yet, but I might take a crack at it just for completion's sake.

### DIFF
--- a/caddyconfig/caddyfile/dispenser.go
+++ b/caddyconfig/caddyfile/dispenser.go
@@ -30,6 +30,10 @@ type Dispenser struct {
 	tokens  []Token
 	cursor  int
 	nesting int
+
+	// A map of arbitrary context data that can be used
+	// to pass through some information to unmarshalers.
+	context map[string]any
 }
 
 // NewDispenser returns a Dispenser filled with the given tokens.
@@ -454,6 +458,34 @@ func (d *Dispenser) DeleteN(amount int) []Token {
 	return d.tokens
 }
 
+// SetContext sets a key-value pair in the context map.
+func (d *Dispenser) SetContext(key string, value any) {
+	if d.context == nil {
+		d.context = make(map[string]any)
+	}
+	d.context[key] = value
+}
+
+// GetContext gets the value of a key in the context map.
+func (d *Dispenser) GetContext(key string) any {
+	if d.context == nil {
+		return nil
+	}
+	return d.context[key]
+}
+
+// GetContextString gets the value of a key in the context map
+// as a string, or an empty string if the key does not exist.
+func (d *Dispenser) GetContextString(key string) string {
+	if d.context == nil {
+		return ""
+	}
+	if val, ok := d.context[key].(string); ok {
+		return val
+	}
+	return ""
+}
+
 // isNewLine determines whether the current token is on a different
 // line (higher line number) than the previous token. It handles imported
 // tokens correctly. If there isn't a previous token, it returns true.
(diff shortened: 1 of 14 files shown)


caddyconfig/httpcaddyfile/builtins.go
```diff
@@ -116,9 +116,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 		} else if !strings.Contains(firstLine[0], "@") {
 			return nil, h.Err("single argument must either be 'internal' or an email address")
 		} else {
-			if acmeIssuer == nil {
-				acmeIssuer = new(caddytls.ACMEIssuer)
-			}
+			acmeIssuer = new(caddytls.ACMEIssuer)
```
mohammed90, line 119: Is this safe? 🤔

francislavoie: Yes, linter was saying "it's always nil". We make a `var` above and never touch it before this point. I'd honestly rather leave the code as-is, in case some new code is added above this that might new() it already, but meh, linters be picky.

francislavoie:
I implemented it for the expression matchers as well. It's kinda gnarly though:

- MatchExpression has custom JSON marshaling. I don't know why we did that. But anyway if we want to pass through the Name, we need to marshal it as a whole object and not just the expression string. So I changed it accordingly. Ran into issues with infinite loops/stack overflows because lol JSON marshal/unmarshal is recursive, so we need to use separate types as the target to cancel the recursion. Thankfully, this is fully backwards compatible with old JSON configs (obviously Caddyfiles will adapt differently but that's fine).
- Added a `ctx.WithValue()` helper to copy the `caddy.Context`, because the stdlib `context.WithValue()` returns a `ctx.Context` but we need a `caddy.Context` and type asserting gave me an error saying "it's not a `caddy.Context`, it's a `ctx.valueCtx` :sob:
- Tests broke because we're using context during expression provisioning, so I had to fix it to use a proper context. No biggie, just copying how we did it in other tests.

caddyconfig/httpcaddyfile/builtins.go
mohammed90: I see. I guess it's the same behavior of the last value overwriting any earlier ones.

mholt:
Thanks for all this work, wow.

I do have concerns that it feels quite complex (code logic) just to save 1 token (or possibly 2 tokens) in the config.

I also wonder what if the matcher has multiple regexes? (I think that's possible, IIRC. Maybe not. Haven't looked lol)

mholt (approved):
Nice work -- thank you Francis for improving the Caddyfile!

PR #6153, opened 2024-03-07
httpcaddyfile: Keep deprecated `skip_log` in directive order

steffenbusch:
This small PR will add the deprecated `skip_log` directive back to the directive order to avoid this error:
`Error: adapting config using caddyfile: directive 'skip_log' is not an ordered HTTP handler, so it cannot be used here - try placing 

See also: 
https://github.com/caddyserver/caddy/pull/6066#issuecomment-1984176170


francislavoie (approved):
Thanks!

PR #6158, opened 2024-03-08
chore: Upgrade CEL to v0.20.1

francislavoie:
Closes #6157

/cc @TristonianJones I took a crack at it, I _think_ I have it right, but I'd appreciate your review to make sure I didn't do anything stupid. Tests pass :man_shrugging:

PR #6161, opened 2024-03-08
upgrade to cel v0.20.0

jbrown-stripe:


jbrown-stripe:
@mholt Sure thing! just signed it

jbrown-stripe:
@mholt thank you so much for your help on this!

I just noticed that we have some other folks working on the same thing? https://github.com/caddyserver/caddy/pull/6158

For the lint errors, will just adding `default` to the switch fix it or do we have to manually specify each case. If so, we should probably follow https://github.com/caddyserver/caddy/pull/6158#discussion_r1518249110 and leave a detailed error for the user for each case?

mholt:
Yeah, Francis also started a patch :sweat_smile: When I merge the PR(s) I'll be sure to credit both you and him.

I was also wondering about `default` -- currently wrapping up a few things before I have a chance to try it but shouldn't be much longer.

mholt:
@jbrown-stripe I've went ahead and figured out the lint errors, unfortunately `default` didn't satisfy it :sweat_smile: 

Thank you, as well as to @francislavoie and @TristonianJones for the baseline work and the review!

mholt (approved):

PR #6163, opened 2024-03-09
caddyhttp: Allow `header` replacement with empty string

francislavoie:
Context: https://caddy.community/t/caddy-ntlm-reverse-proxy-constant-401-with-windows-admin-center/22900/5

Turns out that the Caddyfile adapter was using `""` for the replacement value as meaning "this isn't a replacement, make it a `set` operation instead". That's unfortunate, because in the Caddyfile, using `""` as an empty string for the 2nd token can be done intentionally. Using replacement as empty string to detect if it wasn't set isn't ideal.

So the solution I came up with is to make it a `*string` instead to allow it to be nil instead as the default for no argument being passed.

It does mean we're modifying the signature of an exported function, but I don't think it has any plugin impact: https://sourcegraph.com/search?q=context:global+CaddyfileHeaderOp+-file:vendor/github.com/caddyserver&patternType=keyword&sm=0 (this search excludes Caddy in `vendor/` directories which is the bulk of the results that aren't Caddy itself).


mholt:
Thanks for the patch. Makes sense. But was it intentional to change the output JSON?

francislavoie:
Oh oops, I thought I ran the tests before committing. My bad, I'll revise

francislavoie:
Ok @mholt it's good to go now.

mholt (approved):
Thanks. LGTM!

PR #6164, opened 2024-03-10
encode: write status immediately when status code is informational

WeidiDeng:
Context: https://caddy.community/t/disappear-103-early-hints-response-with-encode-enable-caddy-v2-7-6/23081

Write status immediately when status code is informational.

Also removed `Hijack` since if `101` is written beforehand, it's already written to the underlying response writer. If not, status should be written to the `net.Conn`. `reverse_proxy` write `101` to the response writer.

### DIFF
--- a/modules/caddyhttp/encode/encode.go
+++ b/modules/caddyhttp/encode/encode.go
@@ -20,11 +20,9 @@
 package encode
 
 import (
-	"bufio"
 	"fmt"
 	"io"
 	"math"
-	"net"
 	"net/http"
 	"sort"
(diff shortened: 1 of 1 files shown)


francislavoie:
Awesome :tada:

dev-polymer:
Hi, thanks for quick solution!
Awesomeness double-tested [and confirmed](https://caddy.community/t/disappear-103-early-hints-response-with-encode-enable-caddy-v2-7-6/23081/6) 👍

dunglas (approved):
LGTM

PR #6165, opened 2024-03-10
rewrite: `uri query` replace operation

### LINKED ISSUE #6096: Query rewrite operations

armadi1809:
This PR is a followup to close #6096. It implements the replace operation for query strings.

### DIFF
--- a/modules/caddyhttp/rewrite/caddyfile.go
+++ b/modules/caddyhttp/rewrite/caddyfile.go
@@ -213,6 +213,9 @@ func applyQueryOps(h httpcaddyfile.Helper, qo *queryOps, args []string) error {
 		renameValKey := strings.Split(key, ">")
 		qo.Rename = append(qo.Rename, queryOpsArguments{Key: renameValKey[0], Val: renameValKey[1]})
 
+	case len(args) == 3:
+		qo.Replace = append(qo.Replace, queryOpsReplacement{Key: key, Search: args[1], Replace: args[2]})
+
 	default:
 		if len(args) != 2 {
 			return h.ArgErr()
--- a/modules/caddyhttp/rewrite/caddyfile.go
+++ b/modules/caddyhttp/rewrite/caddyfile.go
@@ -214,7 +214,7 @@ func applyQueryOps(h httpcaddyfile.Helper, qo *queryOps, args []string) error {
 		qo.Rename = append(qo.Rename, queryOpsArguments{Key: renameValKey[0], Val: renameValKey[1]})
 
 	case len(args) == 3:
-		qo.Replace = append(qo.Replace, queryOpsReplacement{Key: key, Search: args[1], Replace: args[2]})
+		qo.Replace = append(qo.Replace, &queryOpsReplacement{Key: key, SearchRegexp: args[1], Replace: args[2]})
 
 	default:
 		if len(args) != 2 {
--- a/caddytest/integration/caddyfile_adapt/uri_query_operations.caddyfiletest
+++ b/caddytest/integration/caddyfile_adapt/uri_query_operations.caddyfiletest
@@ -1,0 +1,106 @@
+:9080
+uri query +foo bar
+uri query -baz
+uri query taz test
+uri query key=value example
+uri query changethis>changed
+uri query {
+	findme value replacement
+	+foo1 baz
+}
+
+respond "{query}"
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":9080"
+					],
+					"routes": [
+						{
+							"handle": [
+								{
+									"handler": "rewrite",
+									"query": {
+										"add": [
+											{
+												"key": "foo",
+												"val": "bar"
+											}
+										]
+									}
+								},
+								{
+									"handler": "rewrite",
+									"query": {
+										"delete": [
+											"baz"
+										]
+									}
+								},
+								{
+									"handler": "rewrite",
+									"query": {
+										"set": [
+											{
+												"key": "taz",
+												"val": "test"
+											}
+										]
+									}
+								},
+								{
+									"handler": "rewrite",
+									"query": {
+										"set": [
+											{
+												"key": "key=value",
+												"val": "example"
+											}
+										]
+									}
+								},
+								{
+									"handler": "rewrite",
+									"query": {
+										"rename": [
+											{
+												"key": "changethis",
+												"val": "changed"
+											}
+										]
+									}
+								},
+								{
+									"handler": "rewrite",
+									"query": {
+										"add": [
+											{
+												"key": "foo1",
+												"val": "baz"
+											}
+										],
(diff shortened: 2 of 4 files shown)


armadi1809:
@francislavoie let me know if you would like to see other test cases to make sure this is working as intended.

modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -213,6 +213,9 @@ func applyQueryOps(h httpcaddyfile.Helper, qo *queryOps, args []string) error {
 		renameValKey := strings.Split(key, ">")
 		qo.Rename = append(qo.Rename, queryOpsArguments{Key: renameValKey[0], Val: renameValKey[1]})
 
+	case len(args) == 3:
+		qo.Replace = append(qo.Replace, queryOpsReplacement{Key: key, Search: args[1], Replace: args[2]})
```
francislavoie, line 217: I don't remember how we did it for header replacement, did we have `SearchRegexp` support from Caddyfile? If not, I think what we should do is allow `~` as a prefix to the search string, and if so it becomes a regexp. That's the way we did it for the `map` directive: https://caddyserver.com/docs/caddyfile/directives/map#syntax. Maybe we can apply that to the header replacements Caddyfile parsing too.

francislavoie (changes requested):
Additional test cases we should have:

- Placeholders (key, search, replacement)
- Partial replacement (only replace a subset of the existing value)
- Ensure nothing is replaced if the search is not found
- Caddyfile adapt test (to ensure it produces the expected JSON), just make a Caddyfile with as many config combinations as you can think of, don't need to be functional, just coverage for the adapter

modules/caddyhttp/rewrite/caddyfile.go
armadi1809: I thin the header replacement uses the `SearchRegexp` exclusively when it come to the Caddyfile support. Do we want to go with that or use the `~` prefix?

francislavoie: Oh :thinking: I guess that's true. Maybe we should do the same here then and only do regexp. WDYT @mholt

mholt: Hmm, I don't remember why we did the header Caddyfile that way; but we could do the same here I suppose. Regex is more powerful :man_shrugging:

mholt (approved):
Lint error is unrelated, not sure why it is showing up here, but this feature LGTM!! Beautiful. Nice work @armadi1809

armadi1809:
Thanks! Most of the scenarios mentioned by @francislavoie above should now be covered. The only one I am not sure about is using placeholders for the search, since we are now using regexes. It looks like neither the `rewrite` operation nor the `headers` replace operation support it. What do you guys think?

mholt (approved):
What if we merge this in for now, and then address that in a follow-up PR? Great work, @armadi1809 !

I will wait for @francislavoie 's final approval.

francislavoie (approved):
Thanks!

mholt:
Yeah this is cool. 😎

JeDaYoshi:
Since this PR was merged, caddy no longer starts and segfaults.

```
Mar 23 04:11:23 ten caddy[60323]: {"level":"info","ts":1711167083.951647,"msg":"using config from file","file":"/etc/caddy/Caddyfile"}
Mar 23 04:11:23 ten caddy[60323]: {"level":"info","ts":1711167083.9550269,"msg":"adapted config to JSON","adapter":"caddyfile"}
Mar 23 04:11:23 ten caddy[60323]: {"level":"info","ts":1711167083.9563556,"logger":"admin","msg":"admin endpoint started","address":"localhost:2019","enforce_origin":false,"origins":["//[::1]:2019","//127.0.0.1:2019","//localhost:2019"]}
Mar 23 04:11:23 ten caddy[60323]: {"level":"info","ts":1711167083.95658,"logger":"http.auto_https","msg":"server is listening only on the HTTPS port but has no TLS connection policies; adding one to enable TLS","server_name":"srv0","https_port":443}
Mar 23 04:11:23 ten caddy[60323]: {"level":"info","ts":1711167083.9566038,"logger":"http.auto_https","msg":"enabling automatic HTTP->HTTPS redirects","server_name":"srv0"}
Mar 23 04:11:23 ten caddy[60323]: {"level":"info","ts":1711167083.9567306,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc000601180"}
Mar 23 04:11:23 ten caddy[60323]: panic: runtime error: invalid memory address or nil pointer dereference
Mar 23 04:11:23 ten caddy[60323]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x13b2282]
Mar 23 04:11:23 ten systemd[1]: caddy.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Mar 23 04:11:23 ten systemd[1]: caddy.service: Failed with result 'exit-code'.
```

```
octo.autie.net {
        encode {
                zstd
                gzip
        }

        header {
                Permissions-Policy "interest-cohort=()"
                Referrer-Policy "same-origin"
                X-Frame-Options "SAMEORIGIN"
                X-XSS-Protection "0"
        }

        handle_path /api/* {
                reverse_proxy 127.0.0.1:4000
        }

        reverse_proxy 127.0.0.1:3000
}
```

Should I open an issue?

armadi1809:
@mholt @francislavoie I just submitted a PR that fixes the issue mentioned in the comment above. I when calling the `Query` field of the `Rewrite` structure, within the provision method, I did not account for the fact that `Query` could be null. Sorry about that!

mholt:
Thank you @armadi1809 :)

PR #6168, opened 2024-03-11
chore: encode: use FlushError instead of Flush

WeidiDeng:


modules/caddyhttp/encode/encode.go
```diff
@@ -232,18 +232,18 @@ func (enc *Encode) Match(rw *responseWriter) bool {
 	return enc.Matcher.Match(rw.statusCode, rw.Header())
 }
 
-// Flush implements http.Flusher. It delays the actual Flush of the underlying ResponseWriterWrapper
+// FlushError is an alternative Flush returning an error. It delays the actual Flush of the underlying ResponseWriterWrapper
 // until headers were written.
```
francislavoie, line 236: Nitpick: keep comment length balanced 

```suggestion
// FlushError is an alternative Flush returning an error. It delays the actual Flush of the underlying
 // ResponseWriterWrapper until headers were written.
```

francislavoie (approved):
Makes sense to me, should work fine as long as plugins use response controller, right?

mholt:
Thank you for the recent patches, Weidi!

This won't break anyone relying on a Flusher will it?

PR #6173, opened 2024-03-15
http: record the number of bytes read when response writer is hijacked

WeidiDeng:
Closes [5728](https://github.com/caddyserver/caddy/issues/5728).

### DIFF
--- a/modules/caddyhttp/responsewriter.go
+++ b/modules/caddyhttp/responsewriter.go
@@ -66,6 +66,8 @@ type responseRecorder struct {
 	size         int
 	wroteHeader  bool
 	stream       bool
+
+	readSize *int
 }
 
 // NewResponseRecorder returns a new ResponseRecorder that can be
@@ -240,6 +242,12 @@ func (rr *responseRecorder) FlushError() error {
 	return nil
 }
 
+// Private interface so it can only be used in this package
+// #TODO: maybe export it later
+func (rr *responseRecorder) setReadSize(size *int) {
+	rr.readSize = size
+}
+
 func (rr *responseRecorder) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	//nolint:bodyclose
 	conn, brw, err := http.NewResponseController(rr.ResponseWriterWrapper).Hijack()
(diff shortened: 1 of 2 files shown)


WeidiDeng:
@Mygod. What do you think of this?

Mygod:
I no longer uses caddy now but my old use case would really prefer a raw connection for kernel space `io.Copy`. However, this would probably suffice for `caddyserver/forwardproxy` and maintainers for that could take a look. :)

modules/caddyhttp/responsewriter.go
```diff
@@ -258,6 +275,24 @@ type hijackedConn struct {
...
+
+func (hc *hijackedConn) Read(p []byte) (int, error) {
+	n, err := hc.Conn.Read(p)
+	hc.updateReadSize(n)
+	return n, err
+}
+
+func (hc *hijackedConn) WriteTo(w io.Writer) (int64, error) {
+	n, err := io.Copy(w, hc.Conn)
+	hc.updateReadSize(int(n))
```
mholt, line 292: Interesting, why do we update read size when writing?

mholt (commented):
Thanks for working on this! Just did a quick first pass and had a question.

modules/caddyhttp/responsewriter.go
WeidiDeng: It's writing to another writer, not actually writing to itself.

mholt:
It sounds like this is no longer needed... is this a bug fix or mostly a feature? If it's a feature I'm not sure it's worth the complexity especially if it's not needed.

That said, I thought we do kernel-space transfers when possible... :thinking:

WeidiDeng:
I think it's a bug fix. For websockets, number of bytes written is logged but not the number of bytes read. And I can't think of a non-invasive way to record that without adding a new field to the log.

modules/caddyhttp/responsewriter.go
```diff
@@ -240,6 +242,12 @@ func (rr *responseRecorder) FlushError() error {
 	return nil
 }
 
+// Private interface so it can only be used in this package
+// #TODO: maybe export it later
+func (rr *responseRecorder) setReadSize(size *int) {
```
mholt, line 247: Passing in a `*int` is a bit weird. Can we just pass in a `int`? Any reason we need a pointer at all, actually?

WeidiDeng: This pointer points to the member of the wrapped request body. So when it's updated, the corresponding log entry is updated as well. `int` requires more changes to the code.

mholt: Ah, I see. Is that thread-safe? (Would it ever be used across goroutines?)

WeidiDeng: The `net.Conn` interface is not thread-safe by default unless specified explicitly. But the only race I know of is when caddy [sends websocket goaway frame](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/streaming.go#L114), which is not thread-safe in the first place. But clients are expected to reconnect anyway, the stats are a bit off (goaway frames are small), nothing breaking.

mholt (approved):
Thanks for working on this Weidi -- let's give it a shot but I may need your help if something comes up :)

PR #6176, opened 2024-03-18
chore: Update quic-go to v0.42.0

marten-seemann:


listeners.go
```diff
@@ -472,16 +473,17 @@ func (na NetworkAddress) ListenQUIC(ctx context.Context, portOffset uint, config
...
 		quicTlsConfig := &tls.Config{GetConfigForClient: sqs.getConfigForClient}
-		earlyLn, err := quic.ListenEarly(h3ln, http3.ConfigureTLSConfig(quicTlsConfig), &quic.Config{
-			Allow0RTT: true,
-			RequireAddressValidation: func(clientAddr net.Addr) bool {
-				// TODO: make tunable?
-				return sqs.getActiveRequests() > 1000
-			},
-		})
+		// Require clients to verify their source address when we're handling more than 256 handshakes per second.
+		limiter := rate.NewLimiter(256, 256)
```
marten-seemann, line 477: Not sure if this value makes sense, I just made up a number.

WeidiDeng: I'll increase the number to 1000. Just like originally done by matt.

francislavoie:
For context, Weidi and I spoke off-Github; this reminded us that a bunch of the listener methods are deprecated and have been for a while, so we can remove it (I checked for public references with sourcegraph and found no usages). And with the change to use the rate limiter, we can get rid of the active requests tracking.

listeners.go
mholt: I don't know the right number either :smile: 1000/sec seems like a lot. But we can try it!

listeners.go
```diff
@@ -468,20 +432,22 @@ func (na NetworkAddress) ListenQUIC(ctx context.Context, portOffset uint, config
...
+		limiter := rate.NewLimiter(1000, 1000)
+		tr := &quic.Transport{
+			Conn:                h3ln,
+			VerifySourceAddress: func(addr net.Addr) bool { return !limiter.Allow() },
+		}
+		earlyLn, err := tr.ListenEarly(http3.ConfigureTLSConfig(quicTlsConfig), &quic.Config{Allow0RTT: true})
 		if err != nil {
 			return nil, err
 		}
+		// TODO: figure out when to close the listener and the transport
```
mholt, line 450: Are we not closing listeners? :thinking:

listeners.go
```diff
@@ -468,20 +432,22 @@ func (na NetworkAddress) ListenQUIC(ctx context.Context, portOffset uint, config
...
-		earlyLn, err := quic.ListenEarly(h3ln, http3.ConfigureTLSConfig(quicTlsConfig), &quic.Config{
-			Allow0RTT: true,
-			RequireAddressValidation: func(clientAddr net.Addr) bool {
-				// TODO: make tunable?
-				return sqs.getActiveRequests() > 1000
-			},
-		})
+		// Require clients to verify their source address when we're handling more than 1000 handshakes per second.
+		// TODO: make tunable?
+		limiter := rate.NewLimiter(1000, 1000)
```
mholt, line 441: This should probably be made tunable eventually, but at least we have a high limit right now that should work for most people. Maybe this could be in a follow-up PR. (It will need to be marked as experimental though as this API is still undergoing finalization.)

mholt (approved):
I have just a couple questions but I think overall this LGTM. Thank you everyone!

listeners.go
WeidiDeng: Yeah, but I think marteen refers to the case where the transport is also used as a `http.RoundTripper`. It's closed properly as a listener.

listeners.go
WeidiDeng: In my opinion, don't fix a problem when it's not 😄 . But keep the possibility open.

mholt: Yeah, true. If we get complaints of errors/DoS, or get requests for it, we can make this tunable I suppose.

listeners.go
mholt: Ah, so perhaps we need to figure that out before using HTTP/3 client in our reverse proxy?

mholt:
I guess we can merge this for now and then fine-tune if needed.

PR #6182, opened 2024-03-20
caddyhttp: add http.request.local{,.host,.port} placeholder

emilylange:
This is the counterpart of `http.request.remote{,.host,.port}`.

`http.request.remote` operates on the remote client's address, while
`http.request.local` operates on the address the connection arrived on.

Take the following example:

- Caddy serving on `203.0.113.1:80`
- Client on `203.0.113.2`

`http.request.remote.host` would return `203.0.113.2` (client IP)

`http.request.local.host` would return `203.0.113.1` (server IP)
`http.request.local.port` would return `80` (server port)

I find this helpful for debugging setups with multiple servers and/or multiple network paths (multiple IPs, AnyIP, Anycast).

A version of this using `respond http.request.local` is currently deployed at https://localip.geklaute.cloud/ (but might not be there forever).
I successfully tested http versions `1.1`, `2` and `3` with `curl` on that url. 

cc @networkException

---

The implementation is heavily based on `http.request.remote`.

But I would like to comment on two code paths that I find a bit confusing:

https://github.com/caddyserver/caddy/blob/e65b97f55b6c5da8859fbd5e5b397cf089bf7716/modules/caddyhttp/replacer.go#L130-L135

For `http.request.remote.port` we try to cast the port from `net.SplitHostPort` as integer.
But if the string -> int conversion fails, we return it as string.
I cannot think of a case where `strconv.Atoi(port)` would ever fail.
And if it actually does, I don't think returning it as string it a good fallback.
I would argue if the conversion fails, something went very wrong.

Similarly, for `http.request.remote.host`:

https://github.com/caddyserver/caddy/blob/e65b97f55b6c5da8859fbd5e5b397cf089bf7716/modules/caddyhttp/replacer.go#L124-L129

If `net.SplitHostPort(req.RemoteAddr)` errors, we return `req.RemoteAddr`.

Would be cool if someone could comment on those two things :)
Because I feel like those are just dead code paths.


mholt (commented):
Nice, thank you! This will have my approval as soon as the tests finish running (assuming they pass)

modules/caddyhttp/replacer.go
```diff
@@ -119,6 +119,20 @@ func addHTTPVarsToReplacer(repl *caddy.Replacer, req *http.Request, w http.Respo
 				return port, true
 			case "http.request.hostport":
 				return req.Host, true
+			case "http.request.local":
+				localAddr, _ := req.Context().Value(http.LocalAddrContextKey).(net.Addr)
+				return localAddr.String(), true
+			case "http.request.local.host":
```
WeidiDeng, line 125: Why not use the same logic as `http.request.remote.host`? Suppose local address is a unix socket, this will return an empty string.

WeidiDeng (changes requested):

modules/caddyhttp/replacer.go
emilylange: Ah, so that's why `http.request.remote.host` is the way it is!

Thanks!

I consided that a dead code path and didn't think of unix sockets.

Will add comments to mention unix sockets to `http.request.remote.host` and `http.request.local.host` together with the fix tomorrow or so :)

emilylange: > Will add comments to mention unix sockets to `http.request.remote.host` and `http.request.local.host` together with the fix tomorrow or so :)

done

mholt (approved):
Great, LGTM @emilylange ! Thank you for the contribution :smiley:

PR #6183, opened 2024-03-21
caddyhttp: Accept XFF header values with ports, when parsing client IP

francislavoie:
Context: https://caddy.community/t/trusted-proxies-in-openshift/23205

As it turns out, some proxies like OpenShift (Red Hat Kubernetes platform) keep the port in the X-Forwarded-For header (or whatever configured client IP header). That's weird, but apparently not unheard of.

We weren't splitting up the value, so when we passed it to `netip.ParseAddr` it would fail, so the `client_ip` in logs didn't get populated. This fixes that, and add some clarifying comments in surrounding code.


mholt (approved):
Ah, interesting. LGTM -- thank you Francis!

PR #6185, opened 2024-03-21
autohttps: Move log WARN to INFO, reduce confusion

francislavoie:
We've seen some users be confused about warnings surrounding Auto HTTPS in the logs. I don't think these need to be warnings, it doesn't really add much value. It's just tracing through whether things are enabled or not, it's not really an actionable problem.

### DIFF
--- a/modules/caddyhttp/autohttps.go
+++ b/modules/caddyhttp/autohttps.go
@@ -117,13 +117,13 @@ func (app *App) automaticHTTPSPhase1(ctx caddy.Context, repl *caddy.Replacer) er
 			srv.AutoHTTPS = new(AutoHTTPSConfig)
 		}
 		if srv.AutoHTTPS.Disabled {
-			logger.Warn("automatic HTTPS is completely disabled for server", zap.String("server_name", srvName))
+			logger.Info("automatic HTTPS is completely disabled for server", zap.String("server_name", srvName))
 			continue
 		}
 
 		// skip if all listeners use the HTTP port
 		if !srv.listenersUseAnyPortOtherThan(app.httpPort()) {
-			logger.Warn("server is listening only on the HTTP port, so no automatic HTTPS will be applied to this server",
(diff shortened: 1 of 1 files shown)


mholt:
I mean, it is kind of important though, isn't it? Caddy has automatic security features that have been overridden, so they won't be enabled. I feel like that deserves to be a little louder than normal.

francislavoie:
If things are disabled, it was clearly the intent of the user, so there's no need to warn. If things didn't get enabled, that's also not a problem, it's just "working as intended".

mholt:
> If things are disabled, it was clearly the intent of the user

My only concern with that is, what if it wasn't their intent?

Like, imagine a light switch on a wall, like a [2-gang switch](https://m.media-amazon.com/images/I/71y8N4G7Z3L._AC_UF894,1000_QL80_.jpg). It's easy to accidentally switch the wrong one (or both, when only one was intended). Similarly in a web server config a user might unintentionally disable security features (for example, maybe they forgot that they moved the HTTP port to a different number).

Other than that, I don't have a strong preference, but I do feel like a warning is better...

francislavoie:
How could it not be their intent to use `auto_https off`? How do you accidentally configure that?

Either way, it's not like we're not logging it anymore. It's still INFO and visible, it's just less confusing because it's _not actually a problem_ it's just a mention of what Caddy decided to do based on the circumstances.

mholt:
>  How do you accidentally configure that?

Imagine you set `http_port 1234` for testing locally or something. And your site is `foo:1234`, it will be served over HTTP since it's on the HTTP port. You might forget you have `http_port` set, or it might not be obvious. (That's one downside of Caddy's implicit config logic.)

If there's confusion about these logs, maybe we should enhance the verbiage.

francislavoie:
That config reads like pretty clear intent to me.

mholt:
As discussed in Slack, I've reverted the one log about the implicit configuration back to WARN, the rest can stay as INFO since they are generally explicitly configured.

PR #6187, opened 2024-03-21
logging: support `ms` duration format and add docs

mohammed90:
The issue caddyserver/transform-encoder#42 made the gap clear. I decided to add a few lines of docs.

### DIFF
--- a/modules/logging/encoders.go
+++ b/modules/logging/encoders.go
@@ -118,17 +118,26 @@ func (je *JSONEncoder) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 // LogEncoderConfig holds configuration common to most encoders.
 type LogEncoderConfig struct {
-	MessageKey     *string `json:"message_key,omitempty"`
-	LevelKey       *string `json:"level_key,omitempty"`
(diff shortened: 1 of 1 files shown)


modules/logging/encoders.go
```diff
@@ -266,6 +275,8 @@ func (lec *LogEncoderConfig) ZapcoreEncoderConfig() zapcore.EncoderConfig {
 		durFormatter = zapcore.NanosDurationEncoder
 	case "string":
 		durFormatter = zapcore.StringDurationEncoder
+	case "ms":
```
francislavoie, line 278: The options are kinda misaligned I think, nano and ms aren't the same, should be nano and milli for example. But we could support all of them. So ns/nano, s/second, ms/milli etc

mohammed90: Better?

francislavoie: Yep, thanks!

Thinking we should probably add docblock comments to the rest of the fields too, but this is good enough for now.

francislavoie (approved):

mholt:
Cool, thanks Mohammed :)

Issue #6197, opened 2024-03-27
upgrade deprecated go-jose package v2.6.0


ayg-qliktech:
- Upgrade deprecated go-jose package v2.6.0 (https://github.com/caddyserver/caddy/blob/0dd0487eba1592b19cb2260fe247b7d375301fe5/go.mod#L153). See https://github.com/square/go-jose/tree/v2.6.0 has now moved to https://github.com/go-jose/go-jose.
- As an added recommendation. It might be worth enabling Dependabot for Go package monitoring to help improve automated updates and avoid issues like this one in the future.

mohammed90:
The package is not a direct dependency of Caddy. It's a transitive dependency through [go.step.sm/crypto/jose](https://pkg.go.dev/go.step.sm/crypto/jose). Since this is not actionable by us, I'll close this issue.

ayg-qliktech:
Why close it and not address the top level package that includes the transitive dependency. Because Caddy will still be using the deprecated package otherwise. Which is also flagged with https://nvd.nist.gov/vuln/detail/CVE-2024-28180 when running security scans against Caddy itself. If we have to create another issue as a Security submission that's fine.

mohammed90:
> Why close it and not address the top level package that includes the transitive dependency.

I'm not sure I understand. Caddy is not using `gopkg.in/square/go-jose.v2` directly. Caddy uses `go.step.sm/crypto/jose`, which in turn depends on `gopkg.in/square/go-jose.v2`. We can't, as Caddy, can't just replace `gopkg.in/square/go-jose.v2` with any other JOSE package just because they both do JOSE. The package `go.step.sm/crypto/jose v0.35.1` expects `gopkg.in/square/go-jose.v2` and can only work with that package. The solution in these scenarios is to ask the upstream package to change/upgrade the dependency they're using. We're not _avoiding_ the upgrade. We're saying we, typically, don't have control on an upstream package.

That said...

I see they already did what you're asking for in the commit https://github.com/smallstep/crypto/commit/d9198ca4bdcca8f72c46ec7d6344c558ee7bc075. I'll see if I can manage upgrading this dependency.

ayg-qliktech:
[...]

PR #6198, opened 2024-03-27
chore: upgrade deps

mohammed90:
I had to upgrade `github.com/smallstep/certificates` to an RC version to pull in the newer `go.step.sm/crypto` version which replaces `gopkg.in/square/go-jose.v2` with `github.com/go-jose/go-jose](https://github.com/go-jose/go-jose`. I read through the changes in `github.com/smallstep/certificates`, and nothing concerning is immediately obvious.

I didn't just want to upgrade everything to have more controlled upgrades.

Relates to #6197


mholt (commented):
Is it possible to do this patch without upgrading otel (yet)? We have a sponsor who is locked into certain otel versions.

Thanks for working on this!

francislavoie (approved):

mholt:
Perfect; thanks!

Issue #6195, opened 2024-03-27
Delay in Caddy using changes to reverse_proxy upstream


paularlott:
I'm writing/written a plugin for Caddy which watches a Nomad / Consul cluster and builds ingress routes to our cluster based on the data, basically it generates a collection of `reverse_proxy` statements to the running containers.

However I've noticed that after reloading the Caddyfile existing clients continue to use the previous configurations for 2 - 3 minutes or until the browser cache is cleared.

Initially I thought this was something I'd done wrong in my code, so I've extracted the minimum Caddyfile and manually applied it against a stock version of Caddy (2.7.6 running in Docker) with the same results.

Initially I might have a `reverse_proxy` block such as

```
reverse_proxy {
      dynamic srv {
	name php-pool-1.service.consul
        refresh 10s
        dial_timeout 5s
      }
      import reverseProxyConfig
      transport http {
        keepalive off
      }
    }
```

After a deployment the `php-pool-1` name changes e.g.

```
reverse_proxy {
      dynamic srv {
	name php-pool-2.service.consul
        refresh 10s
        dial_timeout 5s
      }
      import reverseProxyConfig
      transport http {
        keepalive off
      }
    }
```

If I start browsing the site I can see Caddy logging that it's querying DNS for `php-pool-1.service.consul`, which is correct.

If I now reload with a new Caddyfile to change the upstream to `php-pool-2.service.consul` then I can see Caddy logging that it's querying DNS for `php-pool-1.service.consul` and not the new `php-pool-2.service.consul`, it will keep using the old upstream until either I clear the browser cache or wait 2 - 3 minutes, after which it moves to `php-pool-2.service.consul`.

If I use a different browser to connect directly after the reload then that connection is using `php-pool-2.service.consul`, which is what I expect.

The minimal Caddyfile I have is:

```
{
  admin localhost:2019

  log {
    output stdout
    level DEBUG
    format console
  }

  grace_period 3s
  debug
}

(reverseProxyConfig) {
  header_up +X_FORWARDED_PORT 443
  lb_policy least_conn
  lb_try_duration 5s
  lb_try_interval 250ms
  fail_duration 2s
  unhealthy_status 5xx
}

(logsConfig) {
  log {
    output stdout
    level DEBUG
    format console
  }
}

# Health Check
:8080 {
  respond /health-check OK 200
}

*.fortix.systems {
  import logsConfig
  encode zstd gzip

  @wildcard_86 host ingresstest.fortix.systems
[...]

PR #6202, opened 2024-03-29
close quic connections when server closes

WeidiDeng:
Fix [6195](https://github.com/caddyserver/caddy/issues/6195).

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -575,6 +575,7 @@ func (s *Server) serveHTTP3(addr caddy.NetworkAddress, tlsCfg *tls.Config) error
 				select {
 				case <-s.ctx.Done():
 					if quicConn, ok := request.Context().Value(quicConnCtxKey).(quic.Connection); ok {
+						//nolint:errcheck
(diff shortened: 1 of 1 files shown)


mholt (approved):
LGTM. This is a good temporary solution I think. Thank you @WeidiDeng!!

francislavoie:
This means that h3 requests get cancelled non-gracefully on reload, right? Doesn't that differ from how we typically handle reloads? Does that ignore the grace config in global options?

mholt:
Yeah, I think so. I believe it is a TODO, waiting for a graceful close method upstream in quic-go.

I believe what this does -- correct me if I'm wrong @WeidiDeng -- is when a new HTTP request comes in on a server that was just closed, it will close the QUIC connection with the client. The socket remains open, however.

@marten-seemann Does that sound about right? Once a graceful close is implemented we should be able to allow existing requests to finish during a grace period I think?

WeidiDeng:
Speaking of request interruption, I think my [previous approach](https://github.com/quic-go/quic-go/pull/4181) to interrupt a stream had the same problem (the quic connection is forcefully closed). I'll take a look at that and graceful shutdown when I have time.

mholt:
Thanks Weidi, that'd be great :+1:

Issue #6201, opened 2024-03-28
Faster non cryptographic hash for load balancing


jadidbourbaki:
I have changed the non cryptographic hash used by Caddy from fnv1 to xxhash, which is known to have better throughput than the FNV family of hashes (https://github.com/Cyan4973/xxHash). Was wondering if there is any comments on creating a PR for this change to the main caddy branch? Just a suggestion, looking to find ways to contribute to the caddy project. Love y'all's work. :-) 

https://github.com/jadidbourbaki/caddy/commit/de2a2f6dbc50b3ce59467dfa377a69216f9f18e3

mholt:
Neat, I haven't heard about xxhash, but @cespare's library looks good to me. I like Caleb's work; he's a very talented developer so I'm sure it'll be a suitable dependency for our project.

Feel free to submit a PR :)

PR #6203, opened 2024-03-29
Added Faster Non-cryptographic Hash Function for Load Balancing

jadidbourbaki:
As per the discussion in issue https://github.com/caddyserver/caddy/issues/6201.

- All changes tested
- Dependency from a reliable source
- [Provable](https://github.com/Cyan4973/xxHash) gains in performance

### DIFF
--- a/modules/caddyhttp/reverseproxy/selectionpolicies.go
+++ b/modules/caddyhttp/reverseproxy/selectionpolicies.go
@@ -20,7 +20,6 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"hash/fnv"
 	weakrand "math/rand"
 	"net"
 	"net/http"
@@ -32,6 +31,7 @@ import (
 	"github.com/caddyserver/caddy/v2/caddyconfig"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
+	"github.com/cespare/xxhash"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -824,9 +825,9 @@ func hostByHashing(pool []*Upstream, s string) *Upstream {
 
 // hash calculates a fast hash based on s.
 func hash(s string) uint32 {
-	h := fnv.New32a()
+	h := xxhash.New()
 	_, _ = h.Write([]byte(s))
-	return h.Sum32()
+	return uint32(h.Sum64())
```
mholt, line 830: As I'm unfamiliar with this hash, do we really want to truncate to 32bits? We could change the return signature to uint64 to accommodate the 64-bit hash output.

modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -20,14 +20,15 @@ import (
...
 	"fmt"
-	"hash/fnv"
 	weakrand "math/rand"
 	"net"
 	"net/http"
 	"strconv"
 	"strings"
 	"sync/atomic"
 
+	"github.com/cespare/xxhash"
```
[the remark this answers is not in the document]
jadidbourbaki: Great catch! I have updated it to use v2

modules/caddyhttp/reverseproxy/selectionpolicies.go
jadidbourbaki: I agree that it would be better to keep it to the original uint64. I have modified it. Thanks!

mholt (approved):
Awesome, thanks! Excited to give this a go.

Issue #6205, opened 2024-03-30
Question on sync.Pool for eTag and etag Hash function


jadidbourbaki:
I noticed this [comment](https://github.com/caddyserver/caddy/blob/master/admin.go#L973) originally added by this PR (https://github.com/caddyserver/caddy/commit/f259ed52bb3764ce4fd5d88f1712cb43247c2639). 

Just to confirm, is the ask here to make the [etag hasher](https://github.com/caddyserver/caddy/blob/master/admin.go#L949) be allocated through a sync.Pool? Is this an operation accessed through various go routines concurrently? Would love to try my hand at prototyping this just want to make sure I understand the ask.

Further, any interest on changing this hash function for eTags from fnv to the faster xxhash similar to the change for load balancing in reverse proxy?  I imagine the need for a faster hash is not that big in this case as reloading configurations is an off the critical path operation?

Thank you for your time, and I look forward to your input.

WeidiDeng:
Personally, I'm against micro-optimization like this. Admin endpoints optimization is not a priority unless it's common enough in daily usage.

Changing hash function is a small enough change that I think is fine. Adding sync.Pool is not. There are better places to use sync.Pool,

jadidbourbaki:
Thank you @WeidiDeng , I absolutely see where you're coming from here. There are definitely better places to use sync.Pool and this, being off the critical path, is probably not it. With regards to changing the hash function, if you are leaning towards it being a small enough change so as to not be an issue, I will put in a PR soon. :-)

WeidiDeng:
Sure, prs are welcome.

PR #6207, opened 2024-03-30
Added faster non-cryptographic hash for etag generation

jadidbourbaki:
Related issue: https://github.com/caddyserver/caddy/issues/6205

Similar to the non cryptographic hash added in this PR for load balancing: https://github.com/caddyserver/caddy/pull/6203 (related issue: https://github.com/caddyserver/caddy/issues/6201)

Thanks for the helpful discussion @WeidiDeng

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -26,7 +26,6 @@ import (
 	"expvar"
 	"fmt"
 	"hash"
-	"hash/fnv"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thank you!

(That reminds me, we probably need to try moving ETag to the Headers instead of the Trailers.)

WeidiDeng:
@mholt Wasn't because hash was calculated as the response was being written that it was put as a trailer?

mholt:
@WeidiDeng Yep, so putting in the header will be less efficient.

PR #6208, opened 2024-03-31
admin: Making `Etag` a header, not a trailer

jadidbourbaki:
This is based on @mholt's [comment](https://github.com/caddyserver/caddy/pull/6207#pullrequestreview-1969918038) on an earlier [PR](https://github.com/caddyserver/caddy/pull/6207). 

I have used the very simple approach suggested by an earlier comment of first writing to a buffer, and then writing the buffer onto the response writer so as to have time to write the eTag as the header. 

This probably requires some iterations. I would be grateful for suggestions by reviewers and the community on testing this out / making the implementation better.


mholt (approved):
Thanks! This LGTM.

The only thing I would consider is adding a ~~backdoor~~ [pool] for the buffer. Would you like to do that in this PR? We can always do it later too.

PS. Your contribution is quite timely, here's a complaint from someone about the lack of Trailer support in clients:

https://twitter.com/fredkisss/status/1774726559066177560

> Regarding this PR (and the config in the docs), I wanted to inform that it seems like the Trailer header is not well supported by many languages.
>
> I am writing an app in python that needs to interface with caddy config API and I want to handle concurrent requests to caddy config API, but it seems like there is no http lib in python that supports reading it, so I am at lost.
>
> I even tried to do in JS, but it isn't supported by browsers at all have been removed from the `fetch` spec (API that is used in JS to make http requests).
>
> What I have noticed is a lack of any library, browser or language wanting to support http trailers in response the body.

I'm a little bummed that Trailer support is so lacking still in 2024 :( Here's to a less efficient server API :clinking_glasses:

jadidbourbaki:
Thanks! I have added a backdoor... err, I mean a sync.Pool. Please let me know if there are any implementation mistakes. Thank you.

admin.go
```diff
@@ -973,6 +982,13 @@ func handleConfig(w http.ResponseWriter, r *http.Request) error {
 		// we could consider setting up a sync.Pool for the summed
 		// hashes to reduce GC pressure.
 		w.Header().Set("Etag", makeEtag(r.URL.Path, hash))
+		_, err = w.Write(buf.Bytes())
+		if err != nil {
+			return APIError{HTTPStatus: http.StatusBadRequest, Err: err}
```
WeidiDeng, line 987: An `InternalServerError` would be a better error 🤔 . After all, there shouldn't be any error writing the generated response.

admin.go
```diff
@@ -973,6 +982,13 @@ func handleConfig(w http.ResponseWriter, r *http.Request) error {
...
+		}
+
+		buf.Reset()
```
WeidiDeng, line 990: We should move the `Reset` to when the buffer is retrieved (when client requests invalid config path it will not be recycled and rare chance response fails to write), and `defer` to return the buffer there.

WeidiDeng (changes requested):

admin.go
jadidbourbaki: That's a good point. I am replacing it to Internal Server Error.

admin.go
jadidbourbaki: I agree, made the changes.

admin.go
```diff
@@ -954,17 +954,28 @@ func makeEtag(path string, hash hash.Hash) string {
 	return fmt.Sprintf(`"%s %x"`, path, hash.Sum(nil))
 }
 
+// This buffer pool is used to keep buffers for
+// reading the config file during eTag header generation
+var bufferPool = sync.Pool{
+	New: func() interface{} {
```
WeidiDeng, line 960: `any` instead of `interface{}`.

WeidiDeng (changes requested):

admin.go
jadidbourbaki: Corrected, thanks!

otbutz:
> PS. Your contribution is quite timely, here's a complaint from someone about the lack of Trailer support in clients:

Not only in clients: https://github.com/quic-go/quic-go/issues/2266

mholt (approved):
LGTM! Thanks for using a buffer pool and commenting the code.

Issue #5734, opened 2023-08-11
Read etag from file


ReillyBrogan:
This is inspired from the recent work to support etags for NixOS. We've also started switching many of our builds to being fully reproducible (including stripping timestamps) and this does not play well with Caddy and etags. 

I assume that the reason for not hashing files to generate the etag is because of how expensive that would be on every request. A better solution I think would be to add a new option to the file_server directive to read the etag from a file in the same directory with the same name but with an arbitrary suffix, similar to how the brotli, gzip, and zstd compression directives will serve a pre-compressed file if it exists in the same directory with the appropriate suffix (`.br` etc). This file could be pre-computed in advance by whatever build infrastructure the user uses.

I propose that the end user be allowed to choose the suffix so that it can fit with whatever algorithm they use. For example, the following should be valid directory contents (using .b3sum as the suffix matching the extremely quick BLAKE3 algorithm):
```
index.html
index.html.b3sum
index.html.br
index.html.br.b3sum
favicon.ico
favicon.ico.b3sum
favicon.ico.br
favicon.ico.br.b3sum
```
With the correct configuration Caddy would read the contents of the `.b3sum` file and set the etag header to that value for the request. Because these files are typically extremely small this should be a very cheap operation, and for a particularly busy server it is likely that they would all end up in the page cache.

mholt:
Thanks for the suggestion. We can probably do this, but not until after 2.8.

PR #6222, opened 2024-04-05
fileserver: Read Etags from precomputed files

armadi1809:
Initial pass at closing #5734

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -161,6 +161,12 @@ type FileServer struct {
 	PrecompressedOrder []string `json:"precompressed_order,omitempty"`
 	precompressors     map[string]encode.Precompressed
 
+	// List of file extensions to try to read Etags from.
+	// If set, file Etags will be read from sidecar files
+	// with any of these suffixes, instead of generating
+	// our own Etag.
+	EtagFileExtensions []string `json:"etag_file_extensions,omitempty"`
+
 	fsmap caddy.FileSystems
 
 	logger *zap.Logger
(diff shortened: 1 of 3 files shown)


jdog10141 (approved):

mholt (changes requested):
Thanks for the great enhancement! This looks pretty good but I just have a few comments.

armadi1809:
Thanks for taking a look at this Matt. All the feedback discussions should now be resolved.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -639,6 +659,23 @@ func calculateEtag(d os.FileInfo) string {
...
 }
 
+// Finds the first corresponding etag file for a given file in the file system and return its content
+func (fsrv *FileServer) getEtagFromFile(fileSystem fs.FS, filename string) (string, error) {
+	for _, suffix := range fsrv.EtagFileExtensions {
+		etagFilename := filename + suffix
+		_, err := fs.Stat(fileSystem, etagFilename)
+		if err != nil {
+			continue
+		}
```
mholt, line 669: Is the purpose of this to check whether we can access the file? Maybe we can eliminate this system call and instead just call `fs.ReadFile()`. Then:

```go
if errors.Is(err, fs.ErrNotExist) || errors.Is(err, fs.ErrPermission) {
    continue
}
if err != nil {
}
```

What do you think?

armadi1809: Much much better. Just changed it. Thanks!

mholt (approved):

mholt:
Awesome, thanks for working on this!

hanikesn:
Nice! I wish there were docs for this. I stumbled upon this while looking in the source code to check why my custom etag headers weren't used to serve 304s.

hanikesn:
This only seems to work for precompressed files though, any particular reason for that? I'd be quite useful for this to work on any files.

mholt:
Docs are being updated with the new features shortly :)

mholt:
If the etag is weak, though, the quotes won't be around the whole thing.

We could probably trim newlines if that's a common occurrence, but it's not something I've encountered in my (limited) experience.

hanikesn:
> We could probably trim newlines if that's a common occurrence, but it's not something I've encountered in my (limited) experience.

As long as it ends up in the docs I'm happy :slightly_smiling_face: It's easy to waste time with this digging into the source code of `http.serveContent()`, but also pre-computing etags is already a quite advanced use case. With reproducible builds on the advance though it'll become more common.

mholt:
I'd welcome a PR at least for the newline patch!

PR #6229, opened 2024-04-08
Upgrade: ACMEz v2, CertMagic, and ZeroSSL issuer

mholt:
This (currently-WIP) branch is the final phase in upgrading the certificate automation stack:

- ACMEz v2, which introduces a slightly simpler and improved high-level API, while also exposing the ability to customize ACME orders more from those higher-level APIs. For example, users can now set NotBefore and NotAfter to customize cert lifetime (if CA-supported), and also to use draft-03 ARI by setting the previous cert the new order is replacing.

- CertMagic, wherein we use the new v2 of ACMEz, but also have introduced a new `ZeroSSLIssuer` type which is capable of obtaining certs using ZeroSSL's proprietary API. This has some advantages over its ACME endpoint if you need even stronger SLAs / higher availability or are a paying customer. I think it also allows you to get IP certificates.

- Caddy's implementation of `ZeroSSLIssuer` is changing. Moving the ZeroSSL issuer into CertMagic has some implications, because before this, the ZeroSSLIssuer is only a thin wrapper over the ACMEIssuer that can generate free EAB automatically and then the underlying ACMEIssuer is used after that.

Now, ZeroSSL is doing away with the default email address that Caddy was using to generate EAB when users didn't provide their own email address. In other words, now in order to use ZeroSSL you will need to do it one of these ways:

- Provide your own email address in your config. (Caddy has always had the `email` Caddyfile option for this and we've always recommended using it. That will do.)
- Provide your own ZeroSSL EAB credentials.
- Use the ZeroSSL API with an API key.

This transition might be a little tricky because we already allow you to use your ZeroSSL API key, but it's only for generating the EAB (in lieu of an email address), and nothing else. After this change is done, providing your ZeroSSL API key will actually use the ZeroSSL API to obtain certificates, which is paywalled (for good reason).

After this change is done, here's what I suppose Caddy's behavior will be:

- Let's Encrypt will remain the default CA.
- If (and only if) an email address is provided in the Caddyfile, ZeroSSL will implicitly become an additional default CA. This will be done by implicitly configuring an ACMEIssuer with the ZeroSSL endpoint, and that endpoint will trigger the generation of EAB (using the configured email address) if there's no EAB yet.
- The `zerossl` issuer module is now a thin wrapper over CertMagic's `ZeroSSLIssuer` type (kind of like how the `acme` issuer module is a thin wrapper over CertMagic's `ACMEIssuer` type). It uses the ZeroSSL API exclusively and thus requires an API key.

**Breaking changes (out of our control) that will be needed for, I suppose, a very small subset of users:**

- If you're using the ZeroSSL API key for EAB generation (again, that should only be needed the first time anyway) and don't want to use the ZeroSSL API for certificates, change your config to simply have your ZeroSSL account email instead (remove the use of the `zerossl` issuer module).
- If you're not providing an email address in your config and are relying on ZeroSSL specifically, add your email address to your config. (You do not need a ZeroSSL account, but it's recommended.)

**In summary:**

- Default CA with no extra config is now Let's Encrypt only.
- Caddyfile with an email address enables ZeroSSL ACME as an additional CA:
```
{
    email you@yours.com
}

example.com

# because an email address is provided, LE and ZeroSSL are both default CAs, for redundancy
```
- The `zerossl` issuer is now API-only:
```
{
    cert_issuer zerossl <api_key>
}
```
  or, for a particular site using the `tls` directive:
```
example.com

tls {
    issuer zerossl <api_key>
}
```
- Everything else is basically the same / as expected. (For example, you can still configure ZeroSSL's ACME endpoint manually using EAB, etc.)

Again, I think breaking changes should only affect a certain subset of users. But I'm told the reliability of ZeroSSL's services will improve with these changes.


devsnek:
I tried running this and I'm seeing this error for IP subjects:
```
tls.obtain	will retry	{"error": "[94.26.24.128] Obtain: subject does not qualify for a public certificate: 94.26.24.128", "attempt": 1, "retrying_in": 60, "elapsed": 0.000330948, "max_duration": 2592000}
```
does this check need to be bypassed for the zerossl issuer?

mholt:
I might need to update the logic regarding IP certs. I think it's still true for ACME but I need to double check on CA policies these days. I'm headed to bed right now but will check on this in the morning.

mholt:
@devsnek Ok, so that error only happens if:

- The ACMEIssuer is being used,
- and the CA is Let's Encrypt, ZeroSSL, or Google Trust Services (common public CAs) -- but I just found out that GTS actually does issue IP certificates, so I will update the logic here.

Are you getting that while using the ZeroSSL API? ZeroSSL's ACME endpoint, AFAIK, does not support IP certs still.

mholt:
Ah no worries, I am still working on that part. Soon!! I will update this issue when that's ready.

mholt:
@devsnek Ok it should be ready to test now.

To use ZeroSSL's ACME endpoint, you have to provide an email address in your config. (Use the `acme` issuer.

To use ZeroSSL's API, you need to provide an API key and use the `zerossl` issuer, which now only supports ZeroSSL's API.

modules/caddytls/acmeissuer.go
```diff
@@ -248,6 +267,64 @@ func (iss *ACMEIssuer) Revoke(ctx context.Context, cert certmagic.CertificateRes
...
+	if err != nil {
+		return nil, acct, fmt.Errorf("decoding API response: %v", err)
+	}
+	if result.Error.Code != 0 {
+		return nil, acct, fmt.Errorf("failed getting EAB credentials: HTTP %d: %s (code %d)",
+			resp.StatusCode, result.Error.Type, result.Error.Code)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, acct, fmt.Errorf("failed getting EAB credentials: HTTP %d", resp.StatusCode)
+	}
```
mohammed90, line 318: Shouldn't the status-check code precede the error check in the response content?

mholt: No, because the ZeroSSL API returns 200 on errors. I actually think this HTTP status check is a no-op, but I put it here for good measure.

mohammed90: image

modules/caddytls/zerosslissuer.go
```diff
@@ -17,44 +17,49 @@ package caddytls
...
+	// The host to bind to when opening a listener for
+	// verifying domain names (or IPs).
+	ListenHost string `json:"listen_host,omitempty"`
+
+	// If HTTP is forwarded from port 80, specify the
+	// forwarded port here.
+	AlternateHTTPPort int `json:"alternate_http_port,omitempty"`
+
+	// Use CNAME validation instead of HTTP.
+	CNAMEValidation *DNSChallengeConfig `json:"cname_validation,omitempty"`
```
francislavoie, line 58: Hmm? I'm confused, how is it CNAME validation? What does propagation checks have to do with CNAMEs? Those are unrelated concepts, no?

mholt: That's what ZeroSSL API calls it :man_shrugging: A CNAME record needs to be set... I am not sure why it is that way.

francislavoie: But why is `propagation` config within that? Doesn't make sense to me.

mholt: The CNAME still needs to be visible/propagated in order to be validated. I think?

devsnek:
it does try to issue now, but it hits an issue during validation
```
2024/04/11 23:04:09.158	INFO	tls.obtain	obtaining certificate	{"identifier": "94.26.24.128"}
2024/04/11 23:04:09.159	INFO	tls.issuance.zerossl	creating certificate	{"identifiers": ["94.26.24.128"]}
2024/04/11 23:04:12.952	INFO	tls.issuance.zerossl	created certificate	{"identifiers": ["94.26.24.128"], "cert_id": "8c748ed2a414fa8536c9e65456fd7e5b"}
2024/04/11 23:04:12.952	INFO	tls.issuance.zerossl	validating identifiers	{"identifiers": ["94.26.24.128"], "cert_id": "8c748ed2a414fa8536c9e65456fd7e5b", "verification_method": "HTTP_CSR_HASH"}
2024/04/11 23:04:14.149	INFO	tls.issuance.zerossl	canceled certificate	{"identifiers": ["94.26.24.128"], "cert_id": "8c748ed2a414fa8536c9e65456fd7e5b", "verification_method": "HTTP_CSR_HASH"}
2024/04/11 23:04:14.150	ERROR	tls.obtain	could not get certificate from issuer	{"identifier": "94.26.24.128", "issuer": "zerossl", "error": "verifying identifiers: POST https://api.zerossl.com/certificates/8c748ed2a414fa8536c9e65456fd7e5b/challenges?access_key=redacted: HTTP 200: API error 0: domain_control_validation_failed (details=map[94.26.24.128:map[http://94.26.24.128/.well-known/pki-validation/D1C00E794ACC211811DFED34503DFAAE.txt:{{0 0   } {0 true http_transport_failed Error: HTTP transport exception in file validation. Check if your server and the validation file are reachable and maybe retry afterwards?}}]]) (raw={\"success\":false,\"error\":{\"code\":0,\"type\":\"domain_control_validation_failed\",\"details\":{\"94.26.24.128\":{\"http:\\/\\/94.26.24.128\\/.well-known\\/pki-validation\\/D1C00E794ACC211811DFED34503DFAAE.txt\":{\"file_found\":0,\"error\":true,\"error_slug\":\"http_transport_failed\",\"error_info\":\"Error: HTTP transport exception in file validation. Check if your server and the validation file are reachable and maybe retry afterwards?\"}}}}} decode_error=json: unknown field \"success\")"}
```

mholt:
@devsnek Thanks -- are you sure that URL is publicly reachable? `http://94.26.24.128/.well-known/pki-validation/D1C00E794ACC211811DFED34503DFAAE.txt`

I get a redirect to HTTPS which is probably not going to work :thinking:

Hmm, I got it working for me in testing with CertMagic, I wonder why Caddy is redirecting. It doesn't redirect ACME challenges (but this is not an ACME challenge). Maybe that has something to do with it.

devsnek:
@mholt the only thing running on the server at that ip is caddy, and during that test it had a config like this:
```
foo.snek.dev,
bar.snek.dev
{
	reverse_proxy http://100.70.171.69
}

94.26.24.128 {
	tls {
		issuer zerossl [redacted]
	}
	respond "hello"
}
```
its not currently running with that config though, as i normally use the `gandi` dns plugin and i couldn't make xcaddy compile this branch with it.

devsnek:
maybe this?
https://github.com/caddyserver/certmagic/blob/f7ea6fb698e22dbf793b0b05d28762a21416c786/httphandler.go#L118-L122

mholt:
I think I know the problem... the HTTP server has to be made aware of ACME challenges, it likely also has to be aware of ZeroSSL API's HTTP validation requests... which are slightly different.

mholt:
@devsnek Ok, I was able to verify that the HTTP validation from the ZeroSSL API works for me now. Can you give it a shot again? Thank you!! :blush:

devsnek:
@mholt its working!

devsnek:
seems ipv6 does not currently work
```
2024/04/13 09:48:33.065	ERROR	tls.obtain	could not get certificate from issuer	{"identifier": "2a14:14c0:0:10::", "issuer": "zerossl", "error": "creating certificate: POST https://api.zerossl.com/certificates?access_key=redacted: HTTP 200: API error 2808: invalid_certificate_domain (details=map[]) (raw={\"success\":false,\"error\":{\"code\":2808,\"type\":\"invalid_certificate_domain\"}} decode_error=json: unknown field \"success\")"}
2024/04/13 09:48:33.065	ERROR	tls.obtain	will retry	{"error": "[2a14:14c0:0:10::] Obtain: creating certificate: POST https://api.zerossl.com/certificates?access_key=redacted: HTTP 200: API error 2808: invalid_certificate_domain (details=map[]) (raw={\"success\":false,\"error\":{\"code\":2808,\"type\":\"invalid_certificate_domain\"}} decode_error=json: unknown field \"success\")", "attempt": 1, "retrying_in": 60, "elapsed": 1.387451133, "max_duration": 2592000}
```
i played around on the zerossl site for a bit and it seems like it only accepts the address in the expanded form: `2a14:14c0:0000:0010:0000:0000:0000:0000`

mholt:
Thanks for trying it out!

That's good to know about the IPv6 addresses. That's probably intentional, to avoid ambiguity or mishaps when validating the certificates (just a single form for the IP).

I will probably merge this shortly then.

devsnek:
> I will probably merge this shortly then.

would it be possible for caddy to support the expanded form? currently no matter how you write the ip in the config, it uses the "canonical" (most compact) form rather than the fully expanded form. could it always pass the expanded form to zerossl?

mholt:
Hm, actually, I don't know. The `x509.CertificateRequest` struct (in the Go standard library) takes IP addresses as `[]net.IP` and then the standard lib also does the marshaling of the CSR that we send to the CA. So on a second look, I'm not sure we have control over that, and maybe it's actually _supposed_ to be in the condensed form? I dunno.

Related, sort of: https://github.com/golang/go/issues/30264

devsnek:
Yeah tbc I think the *correct* thing would be for zerossl to accept the canonical format, but idk how realistic that is... might just be too much effort to support ipv6 here for the moment :(

mholt:
@francislavoie @mohammed90  I know this is a huge change, I don't expect a thorough review, but if you want to skim it and look for anything you care about, feel free to approve/request changes and we can get this merged :)

modules/caddytls/zerosslissuer.go
francislavoie: Followed up on Slack about this one, I wasn't quite satisfied.

TL;DR apparently ZeroSSL uses CNAME as a method of domain verification, similar to how ACME DNS uses the TXT record. Super weird that they don't just use TXT instead. But whatever.

francislavoie (approved):

andreground:
in the release notes you say:

_If you use JSON to configure certificate automation policies, you will need to ensure you use the [acme issuer with your email filled out](https://caddyserver.com/docs/json/apps/tls/automation/policies/issuers/acme/#email), **and the [ca field](https://caddyserver.com/docs/json/apps/tls/automation/policies/issuers/acme/#ca) set to [ZeroSSL's ACME server URL](https://zerossl.com/documentation/acme/).**_

Thus, I've updated my JSON config to include:

		"tls": {
			"automation": {
				"policies": [{
					"issuers": [{
					"module":"acme",
					"ca": "https://acme.zerossl.com/v2/DV90",
					"email":"xxx@yyy.com"
				}]
			}]}
		}


Now, all my certs get generated from Zerossl and that was not the intended behaviour.
I would like to have both ACME providers active, as it should be.

Am I missing something?

Thanks

mholt:
@andreground Caddy does what you configure it to do, so if you tell it to use ZeroSSL as the issuer it will use ZeroSSL. If you want to use multiple issuers then be sure to add them to your configuration. 👍

andreground:
@mholt thanks, I think I was updating the config according to the release notes and this ticket comments.

In the first post of this ticket, you wrote: 

> here's what I suppose Caddy's behavior will be:
> 
> Let's Encrypt will remain the default CA.
> If (and only if) an email address is provided in the Caddyfile, ZeroSSL will implicitly become an additional default CA. This will be done by implicitly configuring an ACMEIssuer with the ZeroSSL endpoint, and that endpoint will trigger the generation of EAB (using the configured email address) if there's no EAB yet.

Probably I've misunderstood something or probably the behaviour using Caddyfile is way different from that of the JSON config 👍

This is how I've updated the config to make Caddy behave like you stated, so using Let's Encrypt as first and Zerossl as fallback. Am I doing right?

```
"tls": {
	"automation": {
		"policies": [
			{
				"issuers": [
					{
						"module": "acme"
					},
					{
						"ca": "https://acme.zerossl.com/v2/DV90",
						"email": "xxx@yyy.zzz",
						"module": "acme"
					}
				]
			}
		]
	}
}
```

Thanks!

mholt:
> behaviour using Caddyfile is way different from that of the JSON config 👍

Yes, so in the [release notes](https://github.com/caddyserver/caddy/releases/tag/v2.8.0) we boil it down:

> If you use JSON to configure certificate automation policies, you will need to ensure you use the [acme issuer with your email filled out](https://caddyserver.com/docs/json/apps/tls/automation/policies/issuers/acme/#email), and the [ca field](https://caddyserver.com/docs/json/apps/tls/automation/policies/issuers/acme/#ca) set to [ZeroSSL's ACME server URL](https://zerossl.com/documentation/acme/).

Hopefully that's clear.

So yeah, your updated config looks like what you want. Try ACME with Let's Encrypt first, then ACME with ZeroSSL after. And the ZeroSSL one has an email address. :100:

simaotwx:
For anyone interested, this is how it looks like now:

```
	acme_ca https://acme.zerossl.com/v2/DV90
	acme_eab {
		key_id {$ZEROSSL_KID}
		mac_key {$ZEROSSL_HMAC_KEY}
	}
```

The `cert_issuer` definitely needs to be removed, otherwise it won't work.

Both variables are passed from outside (Terraform -> Packer -> .env -> docker compose -> docker -> Caddy)

If anyone needs a way to generate the EAB credentials in Terraform to then pass it all the way through to Caddy, you can use our provider for this:  https://registry.terraform.io/providers/toowoxx/zerossl/latest/docs/resources/eab_credentials

I made this provider two years ago and realized I didn't actually needed but now it proved to be useful.

The alternative to this is to upgrade the subscription and use the API through Caddy for which EAB credentials are not needed (just the API token).

jorgegonzalez:
We were previously using this configuration on `2.7.4`; am I reading this thread right that we can no longer use the `zerossl` module with a valid API key while using JSON configurations?

```
"tls": {
  "automation": {
    "policies": [
      {
        "disable_ocsp_stapling": true,
        "issuers": [
          {
            "api_key": "ZEROSSL_API_KEY",
            "module": "zerossl"
          }
        ]
      }
    ]
  },
  "disable_ocsp_stapling": true
}
```

francislavoie:
@jorgegonzalez see the release notes for details https://github.com/caddyserver/caddy/releases/tag/v2.8.0

PR #6231, opened 2024-04-10
Removed shim code supporting deprecated lego dns providers

### LINKED ISSUE #6228: Remove support for lego-deprecated
It's been about 4 years now, and I think it's time to remove support for the long-deprecated [lego](https://github.com/caddy-dns/lego-deprecated) DNS providers.

The package pulls in an [obscene number of dependencies](https://github.com/caddy-dns/lego-deprecated/blob/master/go.sum), which have been problematic in the past. In fact, I had to remove this plugin from our build server because it takes several GB of RAM to compile with this package... we have learned from experience that its design does not align with the goals of the Caddy project, which is to be scalable, easily buildable, and robust in dynamic situations. Libdns packages support context cancellation and a direct configuration struct that allows advanced servers to configure multiple credentials for DNS providers, which [lego cannot do](https://github.com/traefik/traefik/issues/5472). For more background on why I created [ACMEz](https://github.com/mholt/acmez) and [libdns](https://github.com/libdns), see https://github.com/caddyserver/certmagic/issues/71.

We have shim code in Caddy to continue supporting lego-deprecated because lego's DNS provider APIs do not align with the more flexible and capable [libdns APIs](https://pkg.go.dev/github.com/libdns/libdns). Libdns packages are more generally useful in that they can manage DNS records across providers for _any purpose_, not just ACME challenges. This is useful in a variety of Caddy modules, like those that provide "dynamic DNS" when your server may not have a static IP, or managing SRV records for flexible infrastructure.

In the past few years, lego's DNS providers have received numerous vital updates to remain functional, but I haven't seen much/any activity on the lego-deprecated package to update dependencies, leading me to believe the module is not receiving much use anymore.

The [caddy-dns](https://github.com/caddy-dns) organization now has over 50 packages for DNS providers, with more on the way. It's still fewer than lego has, but we've got the major ones, and it's easy to add more (especially if they're already implemented in lego). I'm actually hoping removing support for lego-deprecated might encourage a few more to come out of the woodwork.

I would love to do this for 2.8, but no later than 2.9.

armadi1809:
Closes #6228 I think that's all what we need here but I might be wrong.


mholt (approved):
Yep, I do think that's about right. Thanks!

I'll leave this open for a little while and see if there's any strong and convincing feedback, otherwise this LGTM and we can merge it then.

PR #6232, opened 2024-04-10
feat: add generic response interceptors

dunglas:
In proxy mode, Caddy allows [intercepting and modifying responses](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#intercepting-responses) coming from upstream servers.
This is especially useful for implementing [`X-Sendfile`/`X-Accel-Redirect`-like features](https://www.nginx.com/resources/wiki/start/topics/examples/xsendfile/).

However, the current implementation of this feature is only available in proxy mode and doesn't allow to intercept and modify responses created by modules.

This makes it impossible to use features like ` X-Sendfile` with FrankenPHP, https://github.com/mliezun/caddy-snake, and other similar potential modules (LUA module, etc).

This patch implements a generic implementation of response interceptors that behaves similarly to the `reverse_proxy` `handle_response` subdirective.

It currently does not support the `copy_response` and `copy_response_headers` subdirective, but it should be easy to add support for these features in future Pull Requests.
It should also be possible, in the future, to replace the current `reverse_proxy` specific implementation with this new more generic one, but it is out of scope for this PR.

Closes https://github.com/dunglas/frankenphp/issues/365. Closes https://github.com/dunglas/frankenphp/pull/396.

### DIFF
--- a/modules/caddyhttp/intercept/intercept.go
+++ b/modules/caddyhttp/intercept/intercept.go
@@ -31,11 +31,11 @@ import (
 )
 
 func init() {
-	caddy.RegisterModule(InterceptResponse{})
-	httpcaddyfile.RegisterHandlerDirective("intercept_response", parseCaddyfile)
+	caddy.RegisterModule(Intercept{})
+	httpcaddyfile.RegisterHandlerDirective("intercept", parseCaddyfile)
 }
 
-type InterceptResponse struct {
+type Intercept struct {
 	// List of handlers and their associated matchers to evaluate
 	// after successful response generation.
 	// The first handler that matches the original response will
@@ -44,9 +44,9 @@ type InterceptResponse struct {
 	// it is up to the handler to finish handling the response.
 	//
 	// Three new placeholders are available in this handler chain:
-	// - `{http.intercept_response.status_code}` The status code from the response
-	// - `{http.intercept_response.status_text}` The status text from the response
-	// - `{http.intercept_response.header.*}` The headers from the response
+	// - `{http.intercept.status_code}` The status code from the response
+	// - `{http.intercept.status_text}` The status text from the response
+	// - `{http.intercept.header.*}` The headers from the response
 	HandleResponse []caddyhttp.ResponseHandler `json:"handle_response,omitempty"`
 
 	// Holds the named response matchers from the Caddyfile while adapting
@@ -59,15 +59,15 @@ type InterceptResponse struct {
 }
 
 // CaddyModule returns the Caddy module information.
-func (InterceptResponse) CaddyModule() caddy.ModuleInfo {
+func (Intercept) CaddyModule() caddy.ModuleInfo {
 	return caddy.ModuleInfo{
-		ID:  "http.handlers.intercept_response",
-		New: func() caddy.Module { return new(InterceptResponse) },
+		ID:  "http.handlers.intercept",
+		New: func() caddy.Module { return new(Intercept) },
 	}
 }
 
 // Provision ensures that i is set up properly before use.
-func (irh *InterceptResponse) Provision(ctx caddy.Context) error {
+func (irh *Intercept) Provision(ctx caddy.Context) error {
 	// set up any response routes
 	for i, rh := range irh.HandleResponse {
 		err := rh.Provision(ctx)
@@ -107,7 +107,7 @@ func (irh interceptedResponseHandler) WriteHeader(statusCode int) {
 	irh.ResponseRecorder.WriteHeader(statusCode)
 }
 
-func (ir InterceptResponse) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
+func (ir Intercept) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
 	buf := bufPool.Get().(*bytes.Buffer)
 	buf.Reset()
 
@@ -149,9 +149,9 @@ func (ir InterceptResponse) ServeHTTP(w http.ResponseWriter, r *http.Request, ne
 	// set up the replacer so that parts of the original response can be
 	// used for routing decisions
 	for field, value := range r.Header {
-		repl.Set("http.intercept_response.header."+field, strings.Join(value, ","))
+		repl.Set("http.intercept.header."+field, strings.Join(value, ","))
 	}
-	repl.Set("http.intercept_response.status_code", rec.Status())
+	repl.Set("http.intercept.status_code", rec.Status())
 
 	ir.logger.Debug("handling response", zap.Int("handler", rec.handlerIndex))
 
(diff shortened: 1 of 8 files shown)


modules/caddyhttp/interceptresponse/interceptresponse.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/caddyserver/caddy/v2/caddyconfig/httpcaddyfile"
+	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
+)
+
+func init() {
+	caddy.RegisterModule(InterceptResponse{})
+	httpcaddyfile.RegisterHandlerDirective("intercept_response", parseCaddyfile)
+}
+
+type InterceptResponse struct {
```
mholt, line 38: This should have a godoc comment explaining what it is and how it works. It'll appear prominently on our JSON docs pages.

modules/caddyhttp/interceptresponse/interceptresponse.go
```diff
@@ -0,0 +1,333 @@
...
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"github.com/caddyserver/caddy/v2/caddyconfig/httpcaddyfile"
+	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
+)
+
+func init() {
+	caddy.RegisterModule(InterceptResponse{})
+	httpcaddyfile.RegisterHandlerDirective("intercept_response", parseCaddyfile)
```
mholt, line 35: Crazy idea but what if we rename this to `intercept` or something?

modules/caddyhttp/interceptresponse/interceptresponse.go
```diff
@@ -0,0 +1,333 @@
...
+	// Holds the handle_response Caddyfile tokens while adapting
+	handleResponseSegments []*caddyfile.Dispenser
+
+	logger *zap.Logger
+}
+
+// CaddyModule returns the Caddy module information.
+func (InterceptResponse) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "http.handlers.intercept_response",
```
mholt, line 64: (Same question as the directive, should we rename this to `intercept`?)

mholt (commented):
This is a cool change. I haven't thoroughly reviewed it yet but my first pass is liking where this is going. Just a couple nit picky thoughts for starters.

mholt:
Looking good here. @francislavoie any thoughts on the Caddyfile implementation?

modules/caddyhttp/intercept/intercept.go
```diff
@@ -0,0 +1,336 @@
...
+
+		return
+	}
+
+	irh.ResponseRecorder.WriteHeader(statusCode)
+}
+
+func (ir Intercept) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
+	buf := bufPool.Get().(*bytes.Buffer)
+	buf.Reset()
```
francislavoie, line 115: I think we're missing a `Put()`

modules/caddyhttp/intercept/intercept.go
```diff
@@ -0,0 +1,336 @@
...
+	if !rec.Buffered() {
+		return nil
+	}
+
+	// set up the replacer so that parts of the original response can be
+	// used for routing decisions
+	for field, value := range r.Header {
+		repl.Set("http.intercept.header."+field, strings.Join(value, ","))
+	}
+	repl.Set("http.intercept.status_code", rec.Status())
```
francislavoie, line 157: For the proxy placeholder, we have a `{rp.*}` placeholder shortcuts to make it easier to use bits from the intercepted response. I think we should have placeholder shortcuts for this too, but I'm not sure what it should look like. My first thought was `{i.*}` but that might be kinda weird :thinking: Maybe `{ir.*}` for "intercepted response"?

modules/caddyhttp/intercept/intercept.go
```diff
@@ -0,0 +1,336 @@
...
+
+func init() {
+	caddy.RegisterModule(Intercept{})
+	httpcaddyfile.RegisterHandlerDirective("intercept", parseCaddyfile)
+}
+
+// Intercept is a middleware that intercepts then replaces or modifies the original response.
+// It can, for instance, be used to implement X-Sendfile/X-Accel-Redirect-like features
+// when using modules like FrankenPHP or Caddy Snake.
+type Intercept struct {
```
francislavoie, line 41: Lets mark it experimental for now because we might change the API based on how we implement the `copy_response` stuff later.

```suggestion
// Intercept is a middleware that intercepts then replaces or modifies the original response.
// It can, for instance, be used to implement X-Sendfile/X-Accel-Redirect-like features
// when using modules like FrankenPHP or Caddy Snake.
//
// EXPERIMENTAL: Subject to change.
type Intercept struct {
```

modules/caddyhttp/intercept/intercept.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+			continue
+		}
+
+		switch d.Val() {
+		case "handle_response":
+			// delegate the parsing of handle_response to the caller,
+			// since we need the httpcaddyfile.Helper to parse subroutes.
+			// See h.FinalizeUnmarshalCaddyfile
+			i.handleResponseSegments = append(i.handleResponseSegments, d.NewFromNextSegment())
```
francislavoie, line 203: Since this isn't embedded within another directive, I think we don't need the `finalize` thing here. I think we can immediately parse the current segment as a handler and add that to the config. Then we can delete `handleResponseSegments`

mholt:
@dunglas I'll probably tag 2.8 beta 1 this week; it'd be nice to get this in 2.8 beta 1, but if not, maybe we can get it in beta 2 or the RC.

modules/caddyhttp/intercept/intercept.go
mholt: What about `{resp.*}`?

dunglas: I prefer `resp` too. (I added `ir` for now).

modules/caddyhttp/intercept/intercept.go
dunglas: Do you have an idea of how we can do that @francislavoie? We need a `httpcaddyfile.Helper` to call `ParseSegmentAsSubroute()`.

francislavoie: Oooh right. I'll think about it.

mholt:
@dunglas Before we merge this, can we mark the exported types/methods as "EXPERIMENTAL: Subject to change or removal" (or something like that)? It sounds like there are still decisions to be made to finalize APIs to accommodate expanded functionality later, and by marking it as experimental we can at least merge it in sooner and then change it, as opposed to making painful changes later on.

dunglas:
Sounds good to me to mark the public API as experimental. Should I update the code?

mholt:
Yeah, if you have a chance to do that could you push a commit? I'm wrapping up a few things here.

dunglas:
@mholt done (the public type was already marked as experimental).

mholt (approved):
Ah, okay thanks. I was mobile this weekend so thanks for double-checking.

Issue #6233, opened 2024-04-10
Duplicate strings.Count in path matcher


hlidotbe:
Hi,

Unless I'm mistaken there's a redundant count in this check: 

https://github.com/caddyserver/caddy/blob/f4840cfeb85ac33d29a1ab88d474750041a98733/modules/caddyhttp/matchers.go#L457-L460

if it's indeed a mistake I'll do a PR ASAP.

mholt:
Oh the second one is there in case the first one fails.

...

j/k. Would happily accept a PR. Thanks!

PR #6234, opened 2024-04-10
fix: remove duplicate strings.Count in path matcher

hlidotbe:
As discussed in #6233, remove stray duplicate line


mholt (approved):
Great, thank you :)

PR #6249, opened 2024-04-18
staticresp: Use the evaluated response body for sniffing JSON content-type

mohammed90:
Sneaky bug where the automatic setting of `application/json` never took place


francislavoie (approved):
Makes sense

PR #6250, opened 2024-04-18
cmd: Only validate config is proper JSON if config slice has data

mohammed90:
Follow up for #6032. Only validate config is proper JSON if config slice has data.

Without this fix, running `caddy run` or `caddy start` without any config (not even Caddyfile around) results in the error:
```
Error: config is not valid JSON: unexpected end of JSON input; did you mean to use a config adapter (the --adapter flag)?
Error: caddy process exited with error: exit status 1
```


mholt (approved):
Ah, thanks!

Issue #6248, opened 2024-04-18
logs.skip_hosts is ignored


Winand:
Caddy 2.7.6 in Docker. Port forwarding: 80(host) -> 80(caddy), 2999(host) -> 443(caddy)

I have several domain names in Caddyfile, only one of them has log enabled.
```
domain1.com {
  ...
}
domain2.com {
  log
  ...
}
domain3.com {
  ...
}
```

In `caddy adapt` output I see that domain 1 and 3 are added to `skip_hosts`:
```
                    "logs": {
                        "logger_names": {
                            "domain2.com": ""
                        },
                        "skip_hosts": [
                            "domain1.com",
                            "domain3.com"
                        ]
                    },
```
But access log records for all three domains appear in `docker logs caddy`. Is it the expected behaviour?
Of course I can add `skip_log` to domains 1 and 3 but why [skip_hosts](https://caddyserver.com/docs/modules/http#servers/logs/skip_hosts) does not disable logs?

francislavoie:
We've made significant changes to logging since 2.7.6. Please try with the master branch.

Winand:
First I've built caddy using `xcaddy build master`
Then I've built docker image with modified [Dockerfile](https://github.com/caddyserver/caddy-docker/blob/master/2.7/alpine/Dockerfile)
```
...
COPY ./caddy /usr/bin/caddy
RUN setcap cap_net_bind_service=+ep /usr/bin/caddy; \
	chmod +x /usr/bin/caddy; \
	caddy version
...
```

With this newly built image I have no access log in `docker logs caddy` output at all with just a `log` directive in domain2 section.

BUT it works as expected with
```
  log {
    output stderr
  }
```
which adds logging section in `adapt` output:
```
                    "logs": {
                        "logger_names": {
                            "staging-ottg.ddns.net": [
                                "log0"
                            ]
                        },
                        "skip_hosts": [
                            "domain3.com",
                            "domain1.com"
                        ]
                    },
....
    "logging": {
        "logs": {
            "default": {
                "exclude": [
                    "http.log.access.log0"
                ]
            },
            "log0": {
                "include": [
                    "http.log.access.log0"
                ],
                "writer": {
                    "output": "stderr"
                }
            }
        }
    }
```

Winand:
First I've built caddy using `xcaddy build master`
Then I've built docker image with modified [Dockerfile](https://github.com/caddyserver/caddy-docker/blob/master/2.7/alpine/Dockerfile)
[...]

PR #6251, opened 2024-04-19
logging: Fix default access logger

francislavoie:
Fix https://github.com/caddyserver/caddy/issues/6248

The problem is that we were skipping hosts that have `""` as their logger name (which happens when using the `log` directive with no options).

This is a regression from https://github.com/caddyserver/caddy/pull/6088, but the reporter in #6248 had a different problem on 2.7.6 which I think we've since fixed (causing `skip_hosts` to not work).

Also, I added a `SplitHostPort` in a couple spots cause I realized during testing that we were using the host with the port when we shouldn't to look up the logger names. This more thoroughly completes the idea from https://github.com/caddyserver/caddy/pull/5881

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -329,9 +329,10 @@ func (app *App) Provision(ctx caddy.Context) error {
 
 // Validate ensures the app's configuration is valid.
 func (app *App) Validate() error {
-	// each server must use distinct listener addresses
 	lnAddrs := make(map[string]string)
+
 	for srvName, srv := range app.Servers {
+		// each server must use distinct listener addresses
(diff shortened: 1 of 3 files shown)


francislavoie:
I made an additional change to simplify the logic for reading the logger names config, removes the "retry without port" because really there should never be any config with a port, it makes it ambiguous.

I've also added a check in the HTTP app's `Validate()` to make sure there's no ports in the config (only relevant for hand-crafted JSON config, Caddyfile produces the right thing already). Technically a small breaking change, but it should ensure users fix their config so it works as expected.

mholt (approved):
Thanks Francis!

PR #6252, opened 2024-04-19
caddyhttp: Address some Go 1.20 features

mholt:
See #5288.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -38,18 +38,6 @@ import (
 	"go.uber.org/automaxprocs/maxprocs"
 	"go.uber.org/zap"
 
-	// For running in minimal environments, this can ease
-	// headaches related to establishing TLS connections.
-	// "Package fallback embeds a set of fallback X.509 trusted
-	// roots in the application by automatically invoking
-	// x509.SetFallbackRoots. This allows the application to
-	// work correctly even if the operating system does not
-	// provide a verifier or system roots pool. ... It's
-	// recommended that only binaries, and not libraries,
-	// import this package. This package must be kept up to
-	// date for security and compatibility reasons."
-	_ "golang.org/x/crypto/x509roots/fallback"
-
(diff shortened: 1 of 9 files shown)


cmd/main.go
```diff
@@ -40,6 +40,18 @@ import (
...
+	// headaches related to establishing TLS connections.
+	// "Package fallback embeds a set of fallback X.509 trusted
+	// roots in the application by automatically invoking
+	// x509.SetFallbackRoots. This allows the application to
+	// work correctly even if the operating system does not
+	// provide a verifier or system roots pool. ... It's
+	// recommended that only binaries, and not libraries,
+	// import this package. This package must be kept up to
+	// date for security and compatibility reasons."
+	_ "golang.org/x/crypto/x509roots/fallback"
```
francislavoie, line 54: This was my idea... But does it increase the binary size significantly?

mholt: The difference in the size of the artifacts is .2-.3 MB (200-300 KB).

francislavoie: Oh that's not bad at all

francislavoie:
Gah. Linter is running into this issue: https://github.com/daixiang0/gci/issues/76 both `goimports` and `gci` are fighting eachother on the whitespace before the block comment.

As a workaround, I'm trying to move this import to a file of its own, with the comment _above_ the `import` so it shouldn't be touched by either tool. I think that should work, right?

mholt:
Ah, I wondered... Because I manually put it at the very next line but goimports kept adding a new line when I saved. I figured that must be the wrong place, but I guess not.

It looks like the issue upstream in gci is closed? So maybe there's a resolution.

francislavoie:
> It looks like the issue upstream in gci is closed? So maybe there's a resolution.

Nope, it was closed prematurely

francislavoie (approved):

Issue #5985, opened 2023-12-15
Support custom certificate validity period for certificate issuance


clauverjat:
## What would you like to have changed?

It'd be great if users could configure the certificate validity period when requesting certificates from Certificate Authorities (CAs).

## Why is this feature a useful, necessary, and/or important addition to this project?

The default validity period for certificates issued by most Public CAs, such as Let's Encrypt and ZeroSSL, is currently 90 days. However, there are scenarios where shorter certificate lifetimes are useful or required. A shorter validity period can minimize the window of exposure in case the cert's private key is compromised. I believe short certificate lifetimes are especially important since certificate revocation doesn't work (at least not reliably enough to rely on it for serious security on the web).

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

The alternative for users that want certificates with shorter lifetime would be to generate the certificate out of Caddy and then manually load those certificates into Caddy. But doing so means that you don't benefit from Caddy "automatic HTTPS", this means that you'll need to manage the renewal of the certs for instance. 

For my usecase, I've patched caddy, certmagic and acmez to add this feature when using an ACME issuer, but of course this requires me to compile my own custom version of caddy and then maintain it. 

## Context 

For ACME certificate issuance, the desired certificate lifespan can be configured by setting the "notAfter" field in the order. Presently, popular CAs like Let's Encrypt and ZeroSSL do not support this feature. However, Google Public CA does offer this flexibility, and Caddy's own ACME server supports customizable validity periods (within bounds).

If you're interested, I'd be happy to share with you the modifications I've made to implement this feature. But please note that it's my first time working on the Caddy codebase, so my implementation might not be ideal.

mholt:
Sure -- it should be fairly trivial. Want to share your changes?

(Acmez should already support this by virtue of taking a CSR as input.)

clauverjat:
Indeed the change is mostly about reading the option from the Caddyfile and then passing it down the stack up to acmez. I don't think Acmez can achieve that already though. You need to set the "notAfter" field of the order object as described by the RFC [^1] (at least that's what I did), thus I had to add an argument to the ObtainCertificateUsingCSRSource function.

Here are my forks with the changes :
https://github.com/mithril-security/caddy/tree/add-cert-lifetime-config
https://github.com/mithril-security/certmagic/tree/add-cert-lifetime-config
https://github.com/mithril-security/acmez/tree/add-cert-lifetime-config

# Tests done

[...]

PR #6253, opened 2024-04-19
caddytls: Add option to configure certificate lifetime

clauverjat:
Hello,

This PR allows users to specify the certificate validity period to ask the CA for. This PR is the result of the discussion on the corresponding issue #5985.

I took my initial implementation of the feature and updated it slightly so that it is compatible with the changes made by @mholt to CertMagic (I took the current master branch).
Unfortunately though, the resulting implementation does not work. This is because the certmagic "NewACMEIssuer"
function is overriding the NotAfter field that I set in caddy's makeIssuerTemplate(). Here is the problematic code :  https://github.com/caddyserver/certmagic/blob/master/acmeissuer.go#L193-L198 

I think this is likely a "bug" in CertMagic NewACMEIssuer function, since the behavior on NotBefore/NotAfter fields clashes with how the function deals with the other fields. I think you meant to write the following instead  : 
```
if template.NotBefore == 0 {
	template.NotBefore = DefaultACME.NotBefore
}
if template.NotAfter == 0 {
	template.NotAfter = DefaultACME.NotAfter
}
```

Let me know what you think,

Best regards
Corentin Lauverjat


mholt:
Oops!! You're so right. I have pushed a fix here: https://github.com/caddyserver/certmagic/commit/855d4670a49d42499a373436852216fcd4e4047f

clauverjat:
I just pushed a new commit with this update

modules/caddytls/acmeissuer.go
```diff
@@ -88,6 +88,16 @@ type ACMEIssuer struct {
...
 	PreferredChains *ChainPreference `json:"preferred_chains,omitempty"`
 
+	// The validity period to ask the CA to issue a certificate for.
+	// Default: 0 (don't ask a custom lifetime to the CA)
+	// This value is used to compute the "notAfter" field of the ACME order,
+	// therefore the system must have a reasonably synchronized clock.
+	// Important : Let's Encrypt and ZeroSSL don't allow custom
+	// validity period and will refuse to issue a certicate if this is set.
+	// For CAs that support it, there are often limits
+	// on the allowed validity periods. Please refer to your CA documentation.
```
mholt, line 98:
```suggestion
	// The validity period to ask the CA to issue a certificate for.
	// Default: 0 (CA chooses lifetime).
	// This value is used to compute the "notAfter" field of the ACME order;
	// therefore the system must have a reasonably synchronized clock.
	// NOTE: Not all CAs support this. Check with your CA's ACME
	// documentation to see if this is allowed and what values may
	// be used.
```

modules/caddytls/acmeissuer.go
```diff
@@ -349,6 +360,17 @@ func (iss *ACMEIssuer) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+		case "lifetime":
+			var lifetimeStr string
+			if !d.AllArgs(&lifetimeStr) {
+				return d.ArgErr()
+			}
+			lifetime, err := caddy.ParseDuration(lifetimeStr)
+			if err != nil {
+				return d.Errf("invalid lifetime %s: %v", lifetimeStr, err)
+			}
+			// TODO: Add check that valid lifetime must be >= 0
```
mholt, line 372:
```suggestion
			if lifetime < 0 {
				return d.Errf("lifetime must be > 0: %s", lifetime)
			}
```

mholt (changes requested):
Excellent, thank you!

Just a couple nits; also, can you run `go mod tidy` to make sure that go.sum file is cleaned up if needed?

modules/caddytls/acmeissuer.go
clauverjat: Just applied the suggested change with a small twist : the lifetime can be 0

modules/caddytls/acmeissuer.go
```diff
@@ -88,6 +88,15 @@ type ACMEIssuer struct {
 	// will be selected.
 	PreferredChains *ChainPreference `json:"preferred_chains,omitempty"`
 
+	// The validity period to ask the CA to issue a certificate for.
+	// Default: 0 (CA chooses lifetime).
+	// This value is used to compute the "notAfter" field of the ACME order;
+	// therefore the system must have a reasonably synchronized clock.
+	// NOTE: Not all CAs support this. Check with your CA's ACME
+	// documentation to see if this is allowed and what values may
+	// be used.
```
mholt, line 97:
```suggestion
	// be used. EXPERIMENTAL: Subject to change.
```

mholt:
That `gci` failure on that line has got to be a mistake...

modules/caddytls/acmeissuer.go
clauverjat: I just did it. Sorry for the delay, I am quite busy at this moment with other work. I looked into using "Allow edits from maintainers" to give you access. But unfortunately it seems that this option is still not supported by GH for repo owned by org (which is the case here) : https://github.com/orgs/community/discussions/5634 :disappointed:

mholt: Ohhh gotcha, no problem. Thanks for the patch! I just don't want to overpromise :)

mholt:
Thanks for the contribution, we will try it out! If there are any issues with it, would you be able/willing to help with it? :blush:

PR #6258, opened 2024-04-21
chore: use errors.New to replace fmt.Errorf with no parameters will much better

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better

### DIFF
--- a/caddyconfig/caddyfile/adapter.go
+++ b/caddyconfig/caddyfile/adapter.go
@@ -17,8 +17,7 @@ package caddyfile
 import (
 	"bytes"
 	"encoding/json"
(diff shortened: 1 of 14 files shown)


mholt:
I am not sure what evidence there is that this is "much better" -- I actually like that most of our errors are greppable with `fmt.Errorf` instead of needing to search for two functions and adding imports in a bunch of files.

Thanks for the submission, but I'm not sure we'll merge it. Hopefully that is understandable -- feel free to continue discussion if I am wrong, and we can reconsider. (In the future: we generally like to recommend creating an issue to discuss changes first.)

PR #6270, opened 2024-04-24
run `golangci-lint run --fix --fast`

mohammed90:


mholt:
Huh. Why does this fix gci? What does this have to do with imports??

mholt (approved):

Issue #6273, opened 2024-04-25
Missing byte in first websocket message


pando-emil:
I have an issue where a reverse proxied websocket connection looses the first byte in the first message sent from client to backend after the initial websocket upgrade.

- The issue have only been reproducible when running the client on the same host as Caddy, probably because of much lower latency.
- It is confirmed that the issue started to occur with c77a6be, more specifically with the changes in streaming.go.
- Further I have seen that the missing byte is buffered in brw (although e0daa39c makes it a bit harder to access that buffered data).
- I believe the reason for the issue is that after c77a6be the response header is written to the client earlier, making it more likely that the client writes new data to the socket before the hijack stuff is finished.

I have to admit I'm far from a golang expert, but shouldn't the switchProtocolCopier operate on the brw rather than the conn object? Otherwise potentially buffered data would never be copied?

mholt:
@WeidiDeng Do you think you could help out with this?

WeidiDeng:
@pando-emil Can you try `xcaddy build hijack-buffer-fix` to see if it's fixed.

mholt:
Amazing find, @WeidiDeng .

pando-emil:
@WeidiDeng with your reasoning above, wouldn't it be much cleaner to write the header after the hijack as before c77a6be. Then you wouldn't need the buffer juggling introduced e0daa39 at all. There just have to be a less complicated way of subtracting the header bytes from the statistics?

WeidiDeng:
@pando-emil There isn't a way to substract the size from the statistics without introducing a new interface, calculating the size is trivia though. The status needs to be set too, originally websockets have a status of 0. We can't just set it to 101 when hijacking because some proxy handles `CONNECT` requests by hijacking too after writing `200`.

Buffer juggling is necessary to correctly record the number of bytes read. Some people do use `*bufio.ReadWriter` after the hijack is complete, Without it, the bufio.Reader still reads from the original connections and stats are not recorded.

PR #6274, opened 2024-04-26
reverse_proxy: handle buffered data during hijack

WeidiDeng:
Fix [6273](https://github.com/caddyserver/caddy/issues/6273).

The cause of the problem can be found in the [comment](https://github.com/caddyserver/caddy/issues/6273#issuecomment-2078415617).


mholt:
Awesome, that was fast. Thank you! Will await @pando-emil to verify that it is fixed.

Issue #6237, opened 2024-04-12
fileserver: grid view - preview img src attribute is not URL query-escaped


DenebTM:
In the grid view of the file browser's default template, if the filename of an image contains the characters `#` or `?`, the preview doesn't appear, since web browsers (at least Firefox and Chromium) parse the `src` attribute as an URL. The `href` attribute pointing to the file itself is unaffected, so it remains possible to click on these files to open them, only the preview is broken.

![](https://github.com/caddyserver/caddy/assets/7706853/eda7c856-b4cd-450b-8ef6-a5555f0516c0)

mholt:
Thanks! Fixed.

DenebTM:
Oh nice, that was fast! Two things though.

1. A literal `%` needs to be encoded as `%25`, else any filename containing a percent followed by hexadecimal digits will break. And although it doesn't seem to be an issue, `&` should be encoded as `%26` for completeness' sake.

2. Tbh I was kind of anticipating less of a "quick and dirty" fix than my own lol. It seems to me that a custom function wrapping Go's `url.PathEscape` should be added to the template module, since the closest built-in function `urlquery` from text/templates isn't quite suitable for escaping paths.

mholt:
I actually thought the `html` template action would escape accordingly based on context, but I guess that doesn't work.

I didn't reference any other implemention, only just now saw that you pushed some of your own commit elsewhere.

PR #6278, opened 2024-04-27
templates: Add `pathEscape` template function and use it in `browse.html`

DenebTM:
See #6237

Unsure about including a function from the `url` module in `c.tpl.Funcs`, but I see little point in adding a wrapper for it.

(Also, perhaps more functions from the `url` module should be made available in templates.)


mholt (approved):
Sorry for my radio silence on this. Probably a good change.

Do we still need to pipe to `html` after doing path escape?

Anyway, LGTM otherwise. Will wait to merge though until I understand for sure.

DenebTM:
You're right, the pipe to `html` is made unnecessary by `pathEscape`. I've removed it in 98000c1.

mholt (approved):
Great, thanks!

Issue #4016, opened 2021-02-12
Better control over Prometheus metrics


dtelyukh:
On one hand, with lack of lables, prometheus metrics make barely more sense than measuring nationwide average body temperatures. On the other hand, with too many labels, in prometheus might be a source of unbounded high-cardinality, which will cause issues when scraping with Prometheus over long periods of time and with larger deployments.

In the usecase we had, we needed to see per-host metrics (for an obvious reason, different hosts often have very different performance), and also a per-content-type metric, here's why:

1) Dynamically generated resources usually have way bigger TTFB, and mixing such resources with static files we obscure actual peaks in data, because each dynamically generated HTML document also pulls 10 to 100 static files. Content-type is a good heuristic for dynamic content, because html is usually dynamically generated, and everything else is static in most cases

2) No matter if html is dynamic or static, its performance is more critical for overall page load time, because it's the root document for the page, and each millisecond of its load time is always added to overall page load time, which is not always the case for page resources.

We intended to create two extra grafana panels, to be able to see metrics subsets that represent:
- html documents only
- small static files (js/css)
- large static files (images)

Also, it might be useful to separate metrics by document cache status, when caddy cache is enabled.

Based at the discussion in this ticket: https://github.com/caddyserver/caddy/issues/3784 and in this and in this PR: https://github.com/caddyserver/caddy/pull/4015

We came up with a following proposal:
- It's better to add extra configuration options, that allow both more detailed metrics and reduction of metrics size, rather just hardcoding extra added labels
- We cannot extend an existing `metrics` directive, because it defines the virtual host that serves metrics if defined inside of a specific host
- we chose `metrics_options` as new directive name; does anyone have better ideas maybe?

```
// enable per-host metrics for a single host
example.com {
  respond /foo "hello world"
  metrics_options {
    tags {
      host example.com
    }
  }
}
// enable per-host metrics + per-content-type for a single host
example2.com {
  respond /bar "hello world"
  metrics_options {
    tags {
      host example2.com
    }
    headers-to-tags {
      response.content-type
    }
  }
}
// disable metrics for a single host
example3.com:8443 {
  respond /baz "hello world"
  metrics_options {
    disable
  }
}
// disable metrics for a subpath
example3.com {
  handle_path /static {
    metrics_options {
      disable_metrics
    }
  }
}
// globally enable per-host metric, and also per-cache-status, based on response 
metrics_option {
  headers-to-tags {
    request.host
    response.x-cache-status
  }
}
```

[...]

PR #6279, opened 2024-04-28
Add `per_host` option to `metrics` handler

hussam-almarzoq:
This addresses parts of https://github.com/caddyserver/caddy/issues/3784 and https://github.com/caddyserver/caddy/issues/4016. I understand the considerations mentioned by @hairyhenderson regarding the cardinality issue, but I think this is more of a Prometheus issue rather than a Caddy issue. Unlike https://github.com/caddyserver/caddy/issues/4644, no change in Caddy would solve it.

I think adding a clear caution in the docs explaining the implications of enabling this option, much like enabling server metrics, should be enough for the users to understand the tradeoff.

### DIFF
--- a/modules/caddyhttp/metrics.go
+++ b/modules/caddyhttp/metrics.go
@@ -32,10 +32,13 @@ var httpMetrics = struct {
 	init: sync.Once{},
 }
 
-func initHTTPMetrics() {
+func initHTTPMetrics(perHost bool) {
 	const ns, sub = "caddy", "http"
 
 	basicLabels := []string{"server", "handler"}
+	if perHost {
+		basicLabels = append(basicLabels, "host")
+	}
 	httpMetrics.requestInFlight = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: ns,
 		Subsystem: sub,
@@ -60,6 +63,9 @@ func initHTTPMetrics() {
 	sizeBuckets := prometheus.ExponentialBuckets(256, 4, 8)
 
 	httpLabels := []string{"server", "handler", "code", "method"}
+	if perHost {
+		httpLabels = append(httpLabels, "host")
+	}
 	httpMetrics.requestDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: ns,
 		Subsystem: sub,
@@ -103,14 +109,15 @@ func serverNameFromContext(ctx context.Context) string {
 type metricsInstrumentedHandler struct {
 	handler string
 	mh      MiddlewareHandler
+	perHost bool
 }
 
-func newMetricsInstrumentedHandler(handler string, mh MiddlewareHandler) *metricsInstrumentedHandler {
+func newMetricsInstrumentedHandler(handler string, mh MiddlewareHandler, perHost bool) *metricsInstrumentedHandler {
 	httpMetrics.init.Do(func() {
-		initHTTPMetrics()
+		initHTTPMetrics(perHost)
 	})
 
-	return &metricsInstrumentedHandler{handler, mh}
+	return &metricsInstrumentedHandler{handler, mh, perHost}
 }
 
 func (h *metricsInstrumentedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request, next Handler) error {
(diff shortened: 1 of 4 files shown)


francislavoie:
Cardinality is a Caddy problem, because some users serve thousands of domains via On-Demand TLS.

That said, I'll let @hairyhenderson address this, I'm not in touch with metrics stuff.

hussam-almarzoq:
> Cardinality is a Caddy problem, because some users serve thousands of domains via On-Demand TLS.
> 
> That said, I'll let @hairyhenderson address this, I'm not in touch with metrics stuff.



> Cardinality is a Caddy problem, because some users serve thousands of domains via On-Demand TLS.
> 
> That said, I'll let @hairyhenderson address this, I'm not in touch with metrics stuff.

Cardinality limitation is a Prometheus limitation not a Caddy one. It is a consideration that you would have to deal with in Caddy or in any other system that uses Prometheus. I think Caddy should not make it inherently impossible to do this, but have the option there with a sane default – much like having `metrics` disabled by default. I would love to hear from @hairyhenderson.

hairyhenderson:
I'm OK in general with this feature, as long as it defaults to `false` and is well-documented.

@hussam-almarzoq do you think you could issue a companion PR for [caddyserver/website](https://github.com/caddyserver/website) that we could merge soon after this?

I have a few review comments - will post those inline.

modules/caddyhttp/metrics.go
```diff
@@ -33,7 +35,7 @@ var httpMetrics = struct {
 func initHTTPMetrics() {
 	const ns, sub = "caddy", "http"
 
-	basicLabels := []string{"server", "handler"}
+	basicLabels := []string{"server", "handler", "host"}
```
hairyhenderson, line 38: rather than always setting `host` even when `per_host` is false, can you pass through the `perHost` bool and set it conditionally?

```suggestion
	basicLabels := []string{"server", "handler"}
	if perHost {
		basicLabels = append(basicLabels, "host")
	}
```

(same for `httpLabels` below on ln62)

modules/caddyhttp/metrics.go
```diff
@@ -101,23 +103,29 @@ func serverNameFromContext(ctx context.Context) string {
...
 
 func (h *metricsInstrumentedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request, next Handler) error {
+	host := "*"
+	if h.perHost {
+		host = r.Host
+	}
+
 	server := serverNameFromContext(r.Context())
-	labels := prometheus.Labels{"server": server, "handler": h.handler}
+	labels := prometheus.Labels{"server": server, "handler": h.handler, "host": host}
```
hairyhenderson, line 124: should conditionally set `host`:

```suggestion
	labels := prometheus.Labels{"server": server, "handler": h.handler}
	if h.perHost {
		labels["host"] = host
	}
```

hairyhenderson:
In chatting with @francislavoie we realized that only adding the `host` label when the config is set will break reloads. So before we do that we need to support reinitializing metrics on reload.

Until that's done, I think this change is blocked.

francislavoie:
I still disagree with cardinality not being an issue. If a user has this enabled and is serving a wildcard domain where the left-most label is mapped to a username in the app (e.g. `<username>.example.com`) then there can be an infinite amount of domains. 

This means the host label's cardinality is infinite. This means more and more RAM will be consumed until Caddy get's OOM-killed. That's _not ok_.

Also yeah, reloading config is still a problem. Caddy needs to support graceful reloads, so if the config changes we need to update the metrics config. AFAIK this isn't possible right now because of global state, and if we changed it to be non-global state then all metrics will be reset on reload (so all counts/averages/etc are lost on reload). I don't know how we'll solve that.

hussam-almarzoq:
> In chatting with @francislavoie we realized that only adding the `host` label when the config is set will break reloads. So before we do that we need to support reinitializing metrics on reload.
> 
> Until that's done, I think this change is blocked.

I think we can avoid the reload issue by having a default value for the label when it is disabled. This way the config can be changed without changing the label count

hussam-almarzoq:
@hairyhenderson any thoughts?

hairyhenderson:
> I think we can avoid the reload issue by having a default value for the label when it is disabled.

@hussam-almarzoq I'm not excited about that option, because it's now adding a new label to everyone's metrics whether they want it or not.

The key here is:

> we need to support reinitializing metrics on reload.

This is necessary for other metrics work as well, not just for this PR.

To respond to what @francislavoie said:

> Also yeah, reloading config is still a problem. Caddy needs to support graceful reloads, so if the config changes we need to update the metrics config. AFAIK this isn't possible right now because of global state, and if we changed it to be non-global state then all metrics will be reset on reload (so all counts/averages/etc are lost on reload). I don't know how we'll solve that.

Ultimately I think it's OK for everything to reset on reload. It'll appear as if the process died and was restarted, but that's fine.

PR #6284, opened 2024-04-30
Add support for `prefer_precompressed` to FileServer

### LINKED ISSUE #5116: Support "precompressed" files without "base" files

3v0k4:
Fixes #5116 

Todos:
- [ ] Update docs (eg, https://caddyserver.com/docs/caddyfile/directives/file_server#syntax)
- [ ] Review HTTP headers used when decompressing on-the-fly
- [ ] Tests
- [ ] Both `compressed` and `precompressed` used in the module, maybe only use `precompressed`?
- [ ] Need to handle URL canonicalization?
- [ ] `TODO:`s in code

### DIFF
--- a/modules/caddyhttp/fileserver/staticfiles.go
+++ b/modules/caddyhttp/fileserver/staticfiles.go
@@ -15,6 +15,8 @@
 package fileserver
 
 import (
+	"bytes"
+	"compress/gzip"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


3v0k4:
@francislavoie @mholt 

This is still a spike as I'm trying to understand how we want to approach this before I do things properly.

Currently, I added a couple of `if preferPrecompressed` before the normal flow: the first one checks directly for the compressed file according to the client's accepted encodings and the second one looks for all the possible compressed files and, if one is found, uncompresses it.

I left the rest of the code because it takes care of the index files and browse. But we may want to consolidate things a bit more.

Would you be willing to share some feedback please?

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -36,6 +38,29 @@ import (
 	"github.com/caddyserver/caddy/v2/modules/caddyhttp/encode"
 )
 
+type VirtualFile struct {
```
mohammed90, line 41: The struct isn't used outside of the module boundaries, you can use `virtualFile` so we avoid exposing internals unnecessarily.

3v0k4: You are right, thanks!

I'll take care of it as soon as we agree on the direction of this PR.

mholt:
Thanks for working on this -- I'll try to get around to this after the 2.8 release :)

francislavoie:
I'll close this as inactive for now. Thanks for the attempt, sorry for never circling back, but it was too incomplete I think to take much action.

PR #6292, opened 2024-05-02
context: AppIfConfigured returns error; consider not-yet-provisioned modules

francislavoie:
Context https://caddy.community/t/app-not-available-during-provisioning/23678

The problem is that our two existing methods `ctx.App(name)` and `ctx.AppIfConfigured(name)` don't properly cover the case where a module might want to get a configured module instance, because `AppIfConfigured()` could be called before the configured app is actually provisioned, and `App()` could provision an empty app even though it wasn't configured.

So to plug that gap, I propose we add an `ctx.AppStrict(name)` method which is the same as `ctx.App(name)` except that it throws an error early if the app wasn't configured. If it was already loaded then it just returns that, if configured it just loads it and returns it. The method's name is open for discussion if anyone can think of something better.

I also realized that a couple of our usages of `AppIfConfigured()` were error prone due to provisioning order not being guaranteed, so I switched them to use this new method.

There's still one remaining case where we use `AppIfConfigured()` which is in the TLS app's cleanup, we check if the _next_ config (which must have been fully started/provisioned before cleanup of the first config happens) also has a TLS app, and if so pass stuff along. That's to say `AppIfConfigured()` still has a purpose for those kinds of handoff scenarios.


francislavoie:
Actually, @mholt arguably https://github.com/caddyserver/caddy/commit/0e2c7e1d35b287fc0e56d6db2951f791e09b5a37 broke the intended semantics of `AppIfConfigured()`. I don't think it was correct to change how it worked without also providing a way to get an app without loading an empty one. I think the `AppStrict()` API is probably a better name for it in the end.

mholt:
We talked about that change though. Instantiating an app when it wasn't loaded was problematic.

mholt:
Ok, discussed on Slack, and quoting your summary here:

> Both were and you changed it from problem A to problem B without leaving a way to do A, and `AppStrict()` is a new way to solve A:
>
> A = need to get the configured app even if load order caused it to not yet be loaded, but don't want a default app if not configured
> B = need to get the configured app but don't want to load if it wasn't, only valid to be called after provision when it's guaranteed everything is loaded already
> C = (existing `App()` function) need to get an app even if not configured because it's a hard-dependency

context.go
```diff
@@ -453,10 +453,32 @@ func (ctx Context) App(name string) (any, error) {
...
+// to avoid instantiating an empty app when that's not desirable.
+// If the app has not been loaded, nil is returned.
+//
+// Since this method does not return the app if it has not been
+// provisioned, this could be misleading if the provisioning order
+// caused the app to not be provisioned yet. This could return
+// nil even if the app is configured, but not yet provisioned.
+// In that case, the caller should probably use AppStrict() instead
+// to ensure the app is provisioned, and no app will be instantiated
+// if it is not configured.
```
mholt, line 481: So the more I read this, the more I think we need to just fix this: if the app has been configured, load and provision it if needed. Otherwise, don't. This method is called `AppIfConfigured` not `AppIfProvisionedAlready`...

What do you think if we try to do that?

mohammed90: Assuming you had a particular need when you changed it in https://github.com/caddyserver/caddy/commit/0e2c7e1d35b287fc0e56d6db2951f791e09b5a37, are those semantics not needed anymore?

mholt: I don't think either change is quite right though...

The earlier version would provision an app that maybe shouldn't be provisioned (I forgot about the nondeterminstic order). The change makes it never provision the app, but also disregards the loading order.

AppIfConfigured should provision the app if, but ONLY if, it _will_ be provisioned.

francislavoie: I don't think I follow. So what do you think should be done? Just rename AppStrict to AppIfConfigured (and drop the current one & functionality)?

Why did the difference matter for the cert cache commit again? Can you give an example of when it would have been wrong?

mholt: I'm a dummy, `AppStrict()` already does exactly what I'm describing, lol.

So yeah, let's replace AppIfConfigured with AppStrict. This does add an `error` return to the signature but we had it originally and I don't think there's anyone using this yet? (maybe 1 or 2? I apologize if that breaks anyone, but it's a very very easy patch, and I think this is an important fix).

mohammed90:
> A = need to get the configured app even if load order caused it to not yet be loaded, but don't want a default app if not configured
> B = need to get the configured app but don't want to load if it wasn't, only valid to be called after provision when it's guaranteed everything is loaded already
> C = (existing `App()` function) need to get an app even if not configured because it's a hard-dependency

This summary is pretty neat. It'd be great if we can have it documented this way somewhere.

PR #6297, opened 2024-05-03
Set the `ServerName` field in the ClientHelloInfo when it was unset and a default SNI name is configured

ankon:
See discussion in slack: https://caddyserver.slack.com/archives/C4DMTEMGD/p1714689199952049

### DIFF
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -232,6 +232,9 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 				cfg.CertSelection = p.CertSelection
 			}
 			cfg.DefaultServerName = p.DefaultSNI
+			if hello.ServerName == "" && p.DefaultSNI != "" {
+				hello.ServerName = p.DefaultSNI
+			}
 			cfg.FallbackServerName = p.FallbackSNI
 			return cfg.GetCertificate(hello)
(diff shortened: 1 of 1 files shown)


ankon:
Related to: https://github.com/caddyserver/certmagic/tree/default-sni

modules/caddytls/connpolicy.go
```diff
@@ -232,6 +232,9 @@ func (p *ConnectionPolicy) buildStandardTLSConfig(ctx caddy.Context) error {
 				cfg.CertSelection = p.CertSelection
 			}
 			cfg.DefaultServerName = p.DefaultSNI
+			if hello.ServerName == "" && p.DefaultSNI != "" {
```
ArturFortunato, line 235: The second validation is kinda redundant right? Worst case scenario you're just setting "" = ""

mholt:
As discussed in Slack, this might be superseded by the default-sni branch in CertMagic which applies this a little more broadly at the cert-management layer.

mholt:
If I understand correctly, the default-sni branch of CertMagic will supersede this, so I'll close this. (But correct me if I'm wrong and we can reopen it. :smiley: )

ankon:
Mostly for linking things up: The "default-sni branch" is https://github.com/caddyserver/certmagic/pull/287, which is in certmagic and released with 0.21.1.

PR #6298, opened 2024-05-03
Allow usage of root CA without a key. Fixes #6290

### LINKED ISSUE #6290: Caddy PKI without Root key
Hi, according to the docs the caddy pki can be configured like this:
```
{
	pki {
		ca local {
			root {
				format pem_file
				cert /path/to/root.pem
				key /path/to/root.key
			}
			intermediate {
				format pem_file
				cert /path/to/intermediate.pem
				key /path/to/intermediate.key
			}
		}
	}
}
```
if I am solely using the intermediate to issue certificates, then there is no reason to supply the root key. Would it be possible to allow configuration of caddy without a root key?

apollo13:


modules/caddypki/crypto.go
```diff
@@ -78,18 +78,21 @@ func (kp KeyPair) Load() (*x509.Certificate, crypto.Signer, error) {
...
-
 		cert, err := pemDecodeSingleCert(certData)
 		if err != nil {
 			return nil, nil, err
 		}
-		key, err := certmagic.PEMDecodePrivateKey(keyData)
-		if err != nil {
-			return nil, nil, err
+
+		var key crypto.Signer = nil
```
mholt, line 86:
```suggestion
		var key crypto.Signer
```

mholt (approved):
LGTM, thank you!

Just one nit about variable initialization which I'll commit and then we can merge this. Thanks again for the contribution!

PR #6301, opened 2024-05-05
transport: fix proxy protocol initialization

### LINKED ISSUE #6300: Proxy protocol v2, as declared in reverse_proxy / transport http, fails due to unitialized command byte
Using this caddyfile fragment as an example,
```
:80 {
    reverse_proxy {
        to unix//data/socket
        transport http {
            proxy_protocol v2
        }
    }
}
```
the generated stream is prefixed with the proxy protocol sequence as defined by [haproxy](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt).
See this hex dump:
```
0d 0a 0d 0a 00 0d 0a 51 55 49 54 0a 20 11 00 0c c0 a8 58 5b 00 00 00 00 00 00 00 00
```
The 13th byte ("20") is the protocol version and command. The lowest four bits default to 0, meaning LOCAL, but should be initialized to 1, meaning PROXY. Only in this case the receiver uses the information provided in the protocol block to get the original address.

As it stands, the remote address encoded in the header is being ignored by the counterpart.

The related code section is in caddyhttp / reverseproxy / httptransport.go, in function [NewTransport()](https://github.com/caddyserver/caddy/blob/7e2510ef43d3439c682d56b580e4013a0cc9cc3e/modules/caddyhttp/reverseproxy/httptransport.go#L228).
The section about ProxyProtocol could be simplified by using function HeaderProxyFromAddrs() from the [go-proxyproto](https://github.com/pires/go-proxyproto/blob/8a2480a3966f69776d99be21dd09b345fb199ee6/header.go#L47) module.

mohammed90:
Resolve #6300

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -225,41 +225,47 @@ func (h *HTTPTransport) NewTransport(caddyCtx caddy.Context) (*http.Transport, e
 			if !ok {
 				return nil, fmt.Errorf("failed to get proxy protocol info from context")
 			}
-			header := proxyproto.Header{
-				SourceAddr: &net.TCPAddr{
-					IP:   proxyProtocolInfo.AddrPort.Addr().AsSlice(),
-					Port: int(proxyProtocolInfo.AddrPort.Port()),
-					Zone: proxyProtocolInfo.AddrPort.Addr().Zone(),
-				},
+			var proxyv byte
+			switch h.ProxyProtocol {
+			case "v1":
+				proxyv = 1
+			case "v2":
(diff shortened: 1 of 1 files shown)


mohammed90:
@eulores can you please test this PR to validate the fix? You can either grab the CI artifact from [here](modules/caddyhttp/reverseproxy/httptransport.go) or build from source using [xcaddy](https://github.com/caddyserver/xcaddy) with the command:
```
xcaddy build issue-6300
```

eulores:
I tested this PR and it works as expected.
Thank you very much!

mholt (approved):
Awesome. Thanks so much @mohammed90 !!

Issue #6291, opened 2024-05-02
Reverse Proxy active health checks should not follow redirects by default


francislavoie:
When active health checks are enabled (e.g. `health_uri`), the HTTP client will follow redirects by default, possibly resulting in confusing errors like this:

```
{"level":"info","ts":"2024/05/02 14:41:53","logger":"http.handlers.reverse_proxy.health_checker.active","msg":"HTTP request failed","host":"127.0.0.1:8080","error":"Get \"https://127.0.0.1:8080/login\": http: server gave HTTP response to HTTPS client"}
```

This is a result of the upstream serving a redirect from `/` to `https://` + `/login`.

We should disable redirects by default in the active health check client, and write an error if a redirect is served when not expected, counting as a failed request.

We should also add a new `health_follow_redirects` option which toggles on following redirects. I expect this would _very_ rarely be used, but no harm in having the option just in case _someone_ needs it.

PR #6302, opened 2024-05-06
reverseproxy: Stop following redirects in health checks by default, add `health_follow_redirects`

aliasgar55:
fix: #6291

### DIFF
--- a/modules/caddyhttp/reverseproxy/healthchecks.go
+++ b/modules/caddyhttp/reverseproxy/healthchecks.go
@@ -158,8 +158,7 @@ func (a *ActiveHealthChecks) Provision(ctx caddy.Context, h *Handler) error {
 		Transport: h.Transport,
 		CheckRedirect: func(req *http.Request, via []*http.Request) error {
 			if !a.FollowRedirects {
-				return fmt.Errorf(
-					"active health check encountered a redirect; enable 'health_follow_redirects' if redirects are intentional")
+			    return http.ErrUseLastResponse
 			}
 			return nil
 		},
@@ -463,7 +462,7 @@ func (h *Handler) doActiveHealthCheck(dialInfo DialInfo, hostAddr string, upstre
 			markUnhealthy()
 			return nil
 		}
-	} else if resp.StatusCode < 200 || resp.StatusCode >= 400 {
+	} else if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		h.HealthChecks.Active.logger.Info("status code out of tolerances",
 			zap.Int("status_code", resp.StatusCode),
 			zap.String("host", hostAddr),
--- a/modules/caddyhttp/reverseproxy/healthchecks.go
+++ b/modules/caddyhttp/reverseproxy/healthchecks.go
@@ -16,6 +16,7 @@ package reverseproxy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -25,7 +26,6 @@ import (
 	"runtime/debug"
 	"strconv"
 	"time"
-    "errors"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -153,6 +157,14 @@ func (a *ActiveHealthChecks) Provision(ctx caddy.Context, h *Handler) error {
 	a.httpClient = &http.Client{
 		Timeout:   timeout,
 		Transport: h.Transport,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if !a.HealthFollowRedirects {
+				return errors.New(
+					"Redirects are disabled in health check, set health_follow_redirects flag in config to avoid this error")
```
francislavoie, line 163: We typically prefer `fmt.Errorf` for consistency, avoids an extra import line and makes it easier to add some parameters for formatting if needed.

The newline is not necessary.

Adjusted the message as well.

```suggestion
				return fmt.Errorf("active health check encountered a redirect; enable 'health_follow_redirects' if redirects are intentional")
```

But actually, doing a bit of research, I think it would be better to use `http.ErrUseLastResponse` because that's a special error that the Go stdlib uses as a signal to mean "just return the response as-is without an error if a redirect is encountered". It does mean that users won't see an explanation to use this flag in their logs, but I think that's okay, since we document it elsewhere. They'll instead see an `status code out of tolerances` message in the logs. (/cc @mholt it looks like we actually reject `< 200 || >= 400` which means `3xx` are allowed right now, shouldn't we reduce it to `< 200 || >= 300` ?)

```suggestion
				return http.ErrUseLastResponse
```

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -153,6 +157,14 @@ func (a *ActiveHealthChecks) Provision(ctx caddy.Context, h *Handler) error {
...
+			} else {
+				return nil
+			}
```
francislavoie, line 166: The `else` is unnecessary

```suggestion
			}
			return nil
```

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -82,6 +83,9 @@ type ActiveHealthChecks struct {
 	// HTTP headers to set on health check requests.
 	Headers http.Header `json:"headers,omitempty"`
 
+	// boolean to follow redirects in health checks disabled by default
```
francislavoie, line 86:
```suggestion
	// Whether to follow HTTP redirects in response to active health checks (default off).
```

francislavoie (changes requested):

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -82,6 +83,9 @@ type ActiveHealthChecks struct {
...
+	HealthFollowRedirects bool `json:"health_follow_redirects,omitempty"`
```
mohammed90, line 87: Remove the `Health` prefix. We're inside the `ActiveHealthchek` struct already.

mohammed90 (changes requested):

mholt (approved):
Awesome, this is looking good now!! Thank you for the contribution. :)

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -153,6 +156,12 @@ func (a *ActiveHealthChecks) Provision(ctx caddy.Context, h *Handler) error {
 	a.httpClient = &http.Client{
 		Timeout:   timeout,
 		Transport: h.Transport,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if !a.FollowRedirects {
+			    return http.ErrUseLastResponse
```
mholt, line 161: @mohammed90 This lint error is a bit confusing, since this is a `gofmt` error it looks like spaces instead of tabs.

@aliasgar55 Would you be able to `go fmt` this file?

mholt: That was quick. haha

mohammed90: I think it's because golangci-lint runs the linters in the order we specify them in `.golangci.yml` file, and `gci` comes before `gofmt` there.

mholt: Hmm, but I guess why does gci care about go fmt output? I thought it was about imports.

PR #6308, opened 2024-05-08
tracing: add `http.vars.trace_id` placeholder

matthewpi:
Adds the ability to use `trace_id` as a placeholder so it can be used wherever it is needed.  Such as being returned in a header to the client.

### DIFF
--- a/modules/caddyhttp/tracing/tracer.go
+++ b/modules/caddyhttp/tracing/tracer.go
@@ -87,8 +87,12 @@ func (ot *openTelemetryWrapper) serveHTTP(w http.ResponseWriter, r *http.Request
 	ot.propagators.Inject(ctx, propagation.HeaderCarrier(r.Header))
 	spanCtx := trace.SpanContextFromContext(ctx)
 	if spanCtx.IsValid() {
+		traceID := spanCtx.TraceID().String()
+		// Add a trace_id placeholder, accessible via `{http.vars.trace_id}`.
+		caddyhttp.SetVar(ctx, "trace_id", traceID)
+		// Add the trace id to the log fields for the request.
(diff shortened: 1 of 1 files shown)


mholt (approved):
I don't know much about the tracing code but I think this looks reasonable. Thanks!

modules/caddyhttp/tracing/tracer.go
```diff
@@ -87,8 +87,12 @@ func (ot *openTelemetryWrapper) serveHTTP(w http.ResponseWriter, r *http.Request
 	ot.propagators.Inject(ctx, propagation.HeaderCarrier(r.Header))
 	spanCtx := trace.SpanContextFromContext(ctx)
 	if spanCtx.IsValid() {
+		traceID := spanCtx.TraceID().String()
+		// Add a trace_id placeholder, accessible via `{http.vars.trace_id}`.
+		caddyhttp.SetVar(ctx, "trace_id", traceID)
```
mholt, line 92: One other thing to consider is, do we want to access to the Replacer instead of setting a var? If all you want is a placeholder, maybe access the Replacer directly and set a new value. Otherwise, this is fine.

matthewpi: I'm not that familiar with the internals of Caddy.  I'm not sure what the difference is between the Replacer or just setting a variable.

mholt: Basically, vars can be used easily in other modules, whereas the Replacer values are used easily only in user config (though you can still extract values from replacers in code, but with vars it's just `GetVar()`).

So anyway, if this does what you need, don't bother changing anything :)

PR #6311, opened 2024-05-10
Fix typos

szepeviktor:
Found few misspellings.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins_test.go
+++ b/caddyconfig/httpcaddyfile/builtins_test.go
@@ -173,7 +173,7 @@ func TestRedirDirectiveSyntax(t *testing.T) {
 		},
 		{
 			input: `:8080 {
-				redir /old.html /new.html html
+				redir /old.html /new.html htlm
 			}`,
 			expectError: true,
 		},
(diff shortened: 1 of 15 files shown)


caddyconfig/httpcaddyfile/builtins_test.go
```diff
@@ -173,7 +173,7 @@ func TestRedirDirectiveSyntax(t *testing.T) {
 		},
 		{
 			input: `:8080 {
-				redir /old.html /new.html htlm
+				redir /old.html /new.html html
 			}`,
```
mohammed90, line 177: We need to check why this test wasn't failing despite a misspelling that's making it an unacceptable value

mohammed90: It's deliberate, not a type 🙂 It expects an error, so we're expecting failure and test passes because a legitimate error is returned, as expected. Please revert this one.

szepeviktor: Done 🍏

mholt (approved):
Thanks

szepeviktor:
Glad to contribute.

PR #6312, opened 2024-05-10
reverseproxy: Support HTTP/3 transport to backend

### LINKED ISSUE #5086: http/3 support in reverse_proxy module
Hi Guys,

I was tweaking with reverse_proxy settings a lot and was unable to proxy http/3 traffic - does RP supports H3?

Cheers,
Daniel

mholt:
VERY EXPERIMENTAL. Enables HTTP/3 protocol from proxy to backend ("upstream").

I still don't fully understand why this would be needed/useful, but here is a quick stab attempt with the APIs available to us.

**Enabling HTTP/3 necessarily disables other HTTP versions. (If you specify version "3" in your config, you cannot have any other versions.) We do not support protocol downgrade or negotiation like web browsers do.** That would probably add considerable latency and complexity where you should already have control over the backends. Additionally, other HTTP transport options don't apply to the HTTP/3 round-tripper because it does not use the `http.Transport` type like the lower HTTP versions.

Example Caddyfile:

```
# for demonstration purposes, the backend will ONLY support HTTP/3,
# but of course it doesn't have to be this way
{
	servers :1235 {
		protocols h3
	}
}

# this runs the proxy on HTTP but you can use HTTPS in yours
:1234 {
	reverse_proxy localhost:1235 {
		transport http {
			tls
			versions 3
		}
	}
	log
}

# this is the HTTP/3 backend
localhost:1235 {
	log
	respond "Hello world!"
}
```

Then run `curl -v "http://localhost:1234/"` and you'll see the response on the front-end using HTTP/1.1, but the proxy used HTTP/3 to the backend.

In the backend (:1235) we log the request so you can verify that it shows `"proto": "HTTP/3.0"` for the proxied request. The frontend (:1234) will log `"proto": "HTTP/1.1"`.

Anyway, this seems to work, but with the limitations I mentioned above. We have similar, albeit slightly less tight, restrictions on the "h2c" transport as well.

(@marten-seemann You might like to know we're trying this.)

Closes #5086

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -31,6 +31,7 @@ import (
 	"time"
 
 	"github.com/pires/go-proxyproto"
+	"github.com/quic-go/quic-go/http3"
 	"go.uber.org/zap"
 	"golang.org/x/net/http2"
 
@@ -124,12 +125,18 @@ type HTTPTransport struct {
 	// can be specified to use H2C (HTTP/2 over Cleartext) to the
 	// upstream (this feature is experimental and subject to
 	// change or removal). Default: ["1.1", "2"]
+	//
+	// EXPERIMENTAL: "3" enables HTTP/3, but it must be the only
+	// version specified if enabled. Additionally, HTTPS must be
+	// enabled to the upstream as HTTP/3 requires TLS. Subject
+	// to change or removal while experimental.
 	Versions []string `json:"versions,omitempty"`
 
 	// The pre-configured underlying HTTP transport.
 	Transport *http.Transport `json:"-"`
 
 	h2cTransport *http2.Transport
+	h3Transport  *http3.RoundTripper // TODO: EXPERIMENTAL (May 2024)
 }
 
 // CaddyModule returns the Caddy module information.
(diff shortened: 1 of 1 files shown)


marten-seemann:
Interesting! As far as I know, this is pretty uncommon, most CDNs don't use H3 for connecting to the backend.

The fact that this is H3-only shows that there'd be value in creating an HTTP client that properly does Happy Eyeballs v3 (Tracking issue: https://github.com/quic-go/quic-go/issues/4336. Not sure though if this would necessarily live inside of quic-go, or maybe in a small repo on the side).

francislavoie:
This also means `tls` must be enabled, right? I don't think it's valid to do HTTP over QUIC? I can't see how this would ever have a performance advantage over HTTP/1.1 in private networks if it had to add TLS overhead :thinking:

Anyway I see no reason not to merge this as-is. If it seems to work, that's cool. Fun toy, I guess.

mholt:
Yeah, if you take out `tls` it won't work.

mholt:
Eh, this is a pretty low-friction change and shouldn't affect any existing configs since it requires an explicit version 3 be configured; so I might just merge this for the pre-release and we can see how it goes.

Forza-tng:
Seems to work well. Thanks!


One use case is to run caddy on a cloud server and reverse proxy to a self hosted server. Using plain http wouldn't be secure, and h3 is quic"er than h2. 

Of course, running plain HTTP over a VPN between the two is another option, though could be more difficult to set up.

mholt:
> h3 is quic"er than h2.

Have you verified that this is the case for your site?

mholt:
Interesting; any idea what the h2 performance is while you're at it?

Forza-tng:
> Interesting; any idea what the h2 performance is while you're at it?

Forgot about that. h2 performed very similar to h3 with and without wireguard. h2c was slightly ahead over wireguard, while h2/h3 over plain internet was always faster than h2c over wireguard. The variation between runs over wireguard seemed larger than without.

All in all, maybe 'it depends' :)

mholt:
Great -- good to know, thanks for the numbers!

oix7:
> I can't see how this would ever have a performance advantage over HTTP/1.1 in private networks if it had to add TLS overhead 🤔

For any serious performing production networking or applications (especially so needing high speed low latency complex transfers) HTTP/3 is now essentially a prerequisite end to end as of 2026 onwards until it can evolve into something better. The practical differences between h3 and h2 and h1 are huge from an architectural point of view.

Obviously HTTP/3 runs over QUIC, which multiplexes streams without TCP head-of-line blocking, so one lost packet doesn’t freeze every in-flight request like it can with HTTP/2 over TCP; and it also combines transport + crypto handshakes more efficiently. HTTP/1.1 lacks true multiplexing and relies on multiple connections, so it’s fundamentally not comparable.

Whatever TLS performance difference you might imagine will not be worth optimizing for considering what you are getting out of h3 thanks to UDP and the architecture where the more meaningful bottlenecks are solved.

God Bless SCTP over UDP. 💖

PR #6314, opened 2024-05-12
logging: Allow setting log file permissions

### LINKED ISSUE #6295: Caddy log file permissions
Caddy uses https://github.com/natefinch/lumberjack and some time ago they switched the log file permissions from 0600 to 0644: https://github.com/natefinch/lumberjack/pull/83/files

Now, some users (me included) have issues due to these too-restrictive file permissions: https://caddy.community/t/change-file-mask-for-caddy-log-files/22519

There is an issue in the upstream library, but it currently doesn't look like it will be fixed (although to be honest, the proposed PRs don't look very good to me) https://github.com/natefinch/lumberjack/issues/164

However, what lumberjack does is, it will take the permissions of the file if it exists. Now, in Caddy's filewriter.go we already have

```go
// OpenWriter opens a new file writer.
func (fw FileWriter) OpenWriter() (io.WriteCloser, error) {
	// roll log files by default
	if fw.Roll == nil || *fw.Roll {
		if fw.RollSizeMB == 0 {
			fw.RollSizeMB = 100
		}
		if fw.RollCompress == nil {
			compress := true
			fw.RollCompress = &compress
		}
		if fw.RollKeep == 0 {
			fw.RollKeep = 10
		}
		if fw.RollKeepDays == 0 {
			fw.RollKeepDays = 90
		}

		return &lumberjack.Logger{
			Filename:   fw.Filename,
			MaxSize:    fw.RollSizeMB,
			MaxAge:     fw.RollKeepDays,
			MaxBackups: fw.RollKeep,
			LocalTime:  fw.RollLocalTime,
			Compress:   *fw.RollCompress,
		}, nil
	}

	// otherwise just open a regular file
	return os.OpenFile(fw.Filename, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o666)
}
```

So I'm proposing to change this so we always create the file (even if no rolling is configured), so that lumberjack will pick up these permissions. This would also help to prevent the issue that log files are created with mode 0666, but when users configure log rotation, the log files suddenly get created with mode 0600.

What do you think?

ririsoft:
Adding a "mode" option to overwrite the default logfile permissions.
Default remains "0600" which is the one currently used by lumberjack.

Fixes #6295 

I am assuming that Mode = 0 triggers usage of default mode "0600". That implies that users cannot use a "0000" mode for their log files. This sounds reasonable to me, but your opinion here would be appreciated.

I did not test the PR in production. I only tested in local and all went fine. Beware that your umask may interfere with the mode from the config: this is why umask is unset in the tests.

This is my first PR here. I would appreciate great care in the code review, I am fully open to your comments of course.

Cheers.

### DIFF
--- a/modules/logging/filewriter.go
+++ b/modules/logging/filewriter.go
@@ -15,6 +15,7 @@
 package logging
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"math"
@@ -33,6 +34,43 @@ func init() {
 	caddy.RegisterModule(FileWriter{})
 }
 
+// fileMode is a string made of 1 to 4 octal digits representing
+// a numeric mode as specified with the `chmod` unix command.
+// `"0777"` and `"777"` are thus equivalent values.
+type fileMode os.FileMode
+
+// UnmarshalJSON satisfies json.Unmarshaler.
+func (m *fileMode) UnmarshalJSON(b []byte) error {
+	if len(b) == 0 {
+		return io.EOF
+	}
+
+	var s string
+	if err := json.Unmarshal(b, &s); err != nil {
+		return err
+	}
+
+	mode, err := parseFileMode(s)
+	if err != nil {
+		return err
+	}
+
+	*m = fileMode(mode)
+	return err
+}
+
+// parseFileMode parses a file mode string,
+// adding support for `chmod` unix command like
+// 1 to 4 digital octal values.
+func parseFileMode(s string) (os.FileMode, error) {
+	modeStr := fmt.Sprintf("%04s", s)
+	mode, err := strconv.ParseUint(modeStr, 8, 32)
+	if err != nil {
+		return 0, err
+	}
+	return os.FileMode(mode), nil
+}
+
 // FileWriter can write logs to files. By default, log files
 // are rotated ("rolled") when they get large, and old log
 // files get deleted, to ensure that the process does not
@@ -40,6 +78,10 @@ func init() {
 type FileWriter struct {
 	// Filename is the name of the file to write.
 	Filename string `json:"filename,omitempty"`
+
+	// The file permissions mode.
+	// 0600 by default.
+	Mode fileMode `json:"mode,omitempty"`
 
 	// Roll toggles log rolling or rotation, which is
 	// enabled by default.
@@ -100,6 +142,10 @@ func (fw FileWriter) WriterKey() string {
 
 // OpenWriter opens a new file writer.
 func (fw FileWriter) OpenWriter() (io.WriteCloser, error) {
+	if fw.Mode == 0 {
+		fw.Mode = 0o600
+	}
+
 	// roll log files by default
 	if fw.Roll == nil || *fw.Roll {
 		if fw.RollSizeMB == 0 {
@@ -116,6 +162,9 @@ func (fw FileWriter) OpenWriter() (io.WriteCloser, error) {
 			fw.RollKeepDays = 90
 		}
 
+		f_tmp, _ := os.OpenFile(fw.Filename, os.O_WRONLY|os.O_APPEND|os.O_CREATE, os.FileMode(fw.Mode))
+		f_tmp.Close()
+
 		return &lumberjack.Logger{
 			Filename:   fw.Filename,
 			MaxSize:    fw.RollSizeMB,
@@ -127,12 +176,13 @@ func (fw FileWriter) OpenWriter() (io.WriteCloser, error) {
 	}
 
 	// otherwise just open a regular file
-	return os.OpenFile(fw.Filename, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o666)
+	return os.OpenFile(fw.Filename, os.O_WRONLY|os.O_APPEND|os.O_CREATE, os.FileMode(fw.Mode))
 }
 
 // UnmarshalCaddyfile sets up the module from Caddyfile tokens. Syntax:
 //
 //	file <filename> {
+//	    mode          <mode>
 //	    roll_disabled
 //	    roll_size     <size>
 //	    roll_uncompressed
@@ -150,7 +200,7 @@ func (fw FileWriter) OpenWriter() (io.WriteCloser, error) {
 // The roll_keep_for duration has day resolution.
 // Fractional values are rounded up to the next whole number of days.
 //
-// If any of the roll_size, roll_keep, or roll_keep_for subdirectives are
+// If any of the mode, roll_size, roll_keep, or roll_keep_for subdirectives are
 // omitted or set to a zero value, then Caddy's default value for that
 // subdirective is used.
 func (fw *FileWriter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
@@ -165,6 +215,17 @@ func (fw *FileWriter) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 
 	for d.NextBlock(0) {
 		switch d.Val() {
+		case "mode":
+			var modeStr string
+			if !d.AllArgs(&modeStr) {
+				return d.ArgErr()
+			}
+			mode, err := parseFileMode(modeStr)
+			if err != nil {
+				return d.Errf("parsing mode: %v", err)
+			}
+			fw.Mode = fileMode(mode)
+
 		case "roll_disabled":
 			var f bool
 			fw.Roll = &f
--- a/modules/logging/filewriter_test.go
+++ b/modules/logging/filewriter_test.go
@@ -1,0 +1,308 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !windows
+
+package logging
+
+import (
+	"encoding/json"
+	"os"
+	"path"
+	"syscall"
+	"testing"
+
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+)
+
+func TestFileCreationMode(t *testing.T) {
+	on := true
+	off := false
+
+	tests := []struct {
+		name     string
+		fw       FileWriter
+		wantMode os.FileMode
+	}{
+		{
+			name: "default mode no roll",
+			fw: FileWriter{
+				Roll: &off,
+			},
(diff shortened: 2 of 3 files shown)


modules/logging/filewriter.go
```diff
@@ -41,6 +41,10 @@ type FileWriter struct {
 	// Filename is the name of the file to write.
 	Filename string `json:"filename,omitempty"`
 
+	// The file permissions mode.
+	// 0600 by default.
+	Mode os.FileMode `json:"mode,omitempty"`
```
mohammed90, line 46: The [`os.FileMode`](https://pkg.go.dev/os#FileMode) type is an alias to [`fs.FileMode`](https://pkg.go.dev/io/fs#FileMode), which is defined as `uint32`. The mode 777, for example, is ought interpreted as octal by those parsing the string (pun-intended) of characters in the context of file mode. However, the initial (un)marshaling of the value in Go treats it as decimal. Going back to the example of 777, the user is expected to [pass the integer `511` (the equivalent of o777)](https://unix.stackexchange.com/questions/103413/is-there-any-difference-between-mode-value-0777-and-777) to get the effect of o777, which is confusing.

I'm torn between simply using a string versus defining a new type, e.g. `type FileMode string` with custom `UnmarshalJSON` method (like [`caddy.Duration`](https://pkg.go.dev/github.com/caddyserver/caddy/v2#Duration)) that can accept integers (assuming the digits are octal) as well as strings to be interpreted in the form `rwxrwxrwx`, but the latter may be an overkill.

I know you're not a Go dev. Sorry for making this more complicated/work for you 🙂

francislavoie:
Alternatively (or in addition) to this, we could allow specifying the permissions like https://github.com/caddyserver/caddy/pull/4741 as in `/path/to/file.txt|0666`

modules/logging/filewriter.go
mholt: Hmm, that's an interesting idea. I'd be willing to contribute that custom unmarshaler but need some time before I can get around to it.

ririsoft: What about reproducing the behavior of the unix [`chmod`](https://manpages.debian.org/bookworm/coreutils/chmod.1.en.html) command ?

> A numeric mode is from one to four octal digits (0-7), derived by adding up the bits with values 4, 2, and 1. Omitted digits are assumed to be leading zeros.

The following would be equivalent for instance:
```
file access.log {
  mode 777
}

file access.log {
  mode 0777
}
```

That way we keep users in the same confort zone they are with `chmod` and filesystem permissions.
What do you think ?

francislavoie: Remember, that's only the Caddyfile config. You need to also consider JSON config. The Caddyfile adapter produces a JSON config, but it should also be erogonomic for JSON config users to configure this. The type of the field itself dictates how its configured via JSON.

modules/logging/filewriter_test.go
```diff
@@ -0,0 +1,212 @@
...
+			name: "custom mode roll",
+			fw: FileWriter{
+				Roll: &on,
+				Mode: 0o666,
+			},
+			wantMode: 0o666,
+		},
+	}
+
+	m := syscall.Umask(0o000)
```
ririsoft, line 67: Tests are failing on windows due to the umask syscall not defined (God bless CI/CD !).
What would your recommendation be to manage this : having a helper mock that does nothing on windows ? That would imply having additional filewriter_test_unix.go and filewriter_test_windows.go  files to define the mock ?

francislavoie: Yeah probably. Make 2 files, one for windows and one for unix (I guess) which both define the same function, but make the Windows one do nothing.

modules/logging/filewriter_test_posix.go
```diff
@@ -0,0 +1,9 @@
+//go:build darwin || dragonfly || freebsd || linux || nacl || netbsd || openbsd || solaris
```
francislavoie, line 1: Can probably just do `//go:build !windows`?

modules/logging/filewriter.go
ririsoft: > I know you're not a Go dev. Sorry for making this more complicated/work for you 🙂

I used to code in Go a few years ago, I am having fun rediscovering it, no worry.

> I'm torn between simply using a string versus defining a new type, e.g. `type FileMode string` with custom `UnmarshalJSON` method (like [`caddy.Duration`](https://pkg.go.dev/github.com/caddyserver/caddy/v2#Duration)) that can accept integers (assuming the digits are octal)

Considering that [leading zeros are invalid json number](https://datatracker.ietf.org/doc/html/rfc7159#section-6) it seems that if we want to allow users to use octal number with a leading zero we have no choice to use a json string I guess ? Tell me what you think and I will submit a new commit with an implementation proposal, and squash all the work once you are all ok with the PR.

> as well as strings to be interpreted in the form `rwxrwxrwx`, but the latter may be an overkill.

Would you still accept my pull request if I let this one for somebody else's implementation ? Definitely overkill to me :D

mholt: @ririsoft That's what @mohammed90 was referring to: if we make our own type, we can decode a chmod-style string as an octal number.

modules/logging/filewriter_test.go
ririsoft: Thanks @francislavoie. I submitted a new commit hoping it fixes the issue (I have no access to a windows machine to test). Let me know if you prefer an exhaustive 

```go
//go:build darwin || dragonfly || freebsd || linux || nacl || netbsd || openbsd || solaris
```

or a simpler 

```go
//go:build !windows
```

modules/logging/filewriter_test_posix.go
ririsoft: Oups just saw your comment :D

modules/logging/filewriter.go
ririsoft: Thanks @mholt, I will submit a new commit with such proposal in a few days then.

francislavoie:
Tests are failing on Windows - I think what you'll need to do is only do this test on non-Windows (same way, with `//go:build !windows`).

ririsoft:
Ok so now errors are more clear to me. Let me try a few things to have some minimal testing on windows too.

mholt:
Nice work cranking through this! :) I'll circle back after the 2.8 release.

ririsoft:
I have a race condition in the tests with the generation of compressed log file. I will fix that so that you can review the PR with CI OK.

modules/logging/filewriter.go
```diff
@@ -33,6 +34,43 @@ func init() {
 	caddy.RegisterModule(FileWriter{})
 }
 
+// FileMode is a string made of 1 to 4 octal digits representing
+// a numeric mode as specified with the `chmod` unix command.
+// `"0777"` and `"777"` are thus equivalent values.
```
mholt, line 39: Will this ever be a limitation or problematic? (I guess, do people ever represent permissions using any other than the octal representation?) I just don't want to assume something we shouldn't be.

modules/logging/filewriter.go
```diff
@@ -33,6 +34,43 @@ func init() {
...
+	}
+
+	*m = FileMode(mode)
+	return err
+}
+
+// ParseFileMode parses a file mode string,
+// adding support for `chmod` unix command like
+// 1 to 4 digital octal values.
+func ParseFileMode(s string) (os.FileMode, error) {
```
mholt, line 65: If this is only used within this package, maybe we should unexport this.

modules/logging/filewriter.go
```diff
@@ -33,6 +34,43 @@ func init() {
 	caddy.RegisterModule(FileWriter{})
 }
 
+// FileMode is a string made of 1 to 4 octal digits representing
+// a numeric mode as specified with the `chmod` unix command.
+// `"0777"` and `"777"` are thus equivalent values.
+type FileMode os.FileMode
```
mholt, line 40: I wonder if we don't need to export this type either. :thinking: Save ourselves some API surface.

mholt (commented):
Thank you for this! Looks pretty good, I just had a couple questions/suggestions for consideration. Will likely merge soon!

modules/logging/filewriter.go
ririsoft: I am not sure to understand the question.

The unix `chmod` command allows for both octal `0644` and symbolic `u=rw,o=r` files modes. So yes, people are using other than octal representation in their daily life with unix systems.

This PR only implements the octal one, but it does not prevent the symbolic one to be implemented later on if it turns out to be requested by users. I would be surprised, because most of server configuration files use the octal form ([logrotate for instance](https://manpages.debian.org/bookworm/logrotate/logrotate.8.en.html#create))

The symbolic representation will still need to be converted to the octal form as this is expected by `os.Chmod` Go function. So I do not see anything problematic at this stage.

modules/logging/filewriter.go
ririsoft: Sure, i will unexport it and let someone else export it when needed.

mholt (approved):
Gotcha. Thanks for the careful consideration and the tests!

Looks good. Let's give it a whirl. Thank you for the contribution!!

ririsoft:
> Gotcha. Thanks for the careful consideration and the tests!
> 
> Looks good. Let's give it a whirl. Thank you for the contribution!!

Thank you @mholt , @francislavoie , @mohammed90 for your help with this PR. Allow me to congratulate you all, and @mholt in particular, for the amazing work you are doing with Caddy as well as the way you lead the project for contributions in the open. It was really easy to on-board and hack. Well done !

bhavin-qryptal:
Dear Team,

I am using caddy version v2.8.4 and I am trying to have read access granted to caddy log file to all users. Based on this pull request, I have got Caddyfile with section like ... 

`
        log {
                output file /var/log/caddy/access.log {
                        mode 644
                }
        }
`
This is not working, could someone please guide me in correct direction. I am newbie with caddy / golang.

francislavoie:
In what way is it not working? Show evidence.

mholt:
Also probably open a new issue with enough information to reproduce the bug :+1:

bhavin-qryptal:
> In what way is it not working? Show evidence.

Thank you for prompt reply.

image

access.log file permission remains 600, I am expecting it to be 644. Please let me know if you need any other details.

bhavin-qryptal:
> Also probably open a new issue with enough information to reproduce the bug 👍

At this stage, I am not sure, whether this is a bug OR I am doing something wrong. Is the change discussed in this thread included in v2.8.4 ?

francislavoie:
Ah, it's not been released, no. It was merged June 6th, last release was June 2nd. You can build from master if you want it now.

bhavin-qryptal:
> log { output file /var/log/caddy/access.log { mode 644 } }

Thanks a lot. I will give it a go. 

image
Hope this looks reasonable to you.

jjlin:
After updating to Caddy 2.9.0, I immediately ran into the issue of my log files being inaccessible. For backward compatibility, I think it would've been better to use the previous default of `644` even if Lumberjack changed the default on their side, but if you don't want to do that, it would be nice to explicitly call this out as a breaking change (sorry in advance if it's actually mentioned somewhere and I missed it).

PR #6315, opened 2024-05-12
ci: abstract go installation into shared composite action

mohammed90:


francislavoie:
The reason we explicitly used `~` for the version was because at one point `setup-go` was serving us the cached Go version instead of the patched version that came out that same day so we wanted to bust through the cache by saying "give me at least this patch version".

I do like the idea of the jobs having "stable" and "oldstable" on them because then we don't need to continually update our "required jobs" config.

But the problem is the meaning of those can change out from under us when we're not ready to move ahead. Often we have some dependencies to update before we can use the next Go version (though this is much less of a problem now that `quic-go` & `qtls` is reworked).

Also I think sometimes `golangci-lint` sometimes doesn't support the latest Go so we need to pin to the previous one until they fix things.

So I dunno :man_shrugging:

mohammed90:
> The reason we explicitly used `~` for the version was because at one point `setup-go` was serving us the cached Go version instead of the patched version that came out that same day so we wanted to bust through the cache by saying "give me at least this patch version".

Faulty memory is bad 😅 

My intent is to avoid having to update 3 or 4 files every time and forgetting one of the other in the process. That part works, but I think the outputs are stepping over each others toes 🤔 I'll shelf it for now. I wanted to get the PoC out for discussion at least.

mohammed90:
Not in favor anymore, but we know it's doable if at all necessary.

PR #6320, opened 2024-05-16
matchers: support `null` value in expression matcher

mohammed90:
The plain `header` match (non-CEL) supports checking for the absence of a header. The corresponding CEL matcher couldn't achieve the same because it expects a string a slice in the value of header->value map. In the plain `header` matcher, the absence of a value is represented by a nil in-place of zero-slice as the value of the header key in the map. The `nil` only equals `nil` when iterating through the map. We needed an analogous value, i.e. `null`.

My attempt here is to add support for `null`. I'm stumbling in the dark as I'm not familiar with the cel-go module, but something works, so am I close?

Here are the checks I've run:

```bash
~ $ # Before: using `@without `header({"X-Custom-Header": ""})``
~ $ curl -k https://localhost
with
~ $ curl -k -H "X-Custom-Header:" https://localhost
with
~ $ curl -k -H "X-Custom-Header: d" https://localhost
with

~ $ # Before: using `@without `header({"X-Custom-Header": [] })``
2024-05-17 1:20:57  ~ $ curl -k https://localhost
with
~ $ curl -k -H "X-Custom-Header:" https://localhost
with
~ $ curl -k -H "X-Custom-Header: val" https://localhost
without

~ $ # After: using ``
~ $ curl -k https://localhost
without
~ $ curl -k -H "X-Custom-Header:" https://localhost
without
~ $ curl -k -H "X-Custom-Header: val" https://localhost
with
```

So, works?

### DIFF
--- a/modules/caddyhttp/celmatcher.go
+++ b/modules/caddyhttp/celmatcher.go
@@ -37,6 +37,7 @@ import (
 	"github.com/google/cel-go/interpreter/functions"
 	"github.com/google/cel-go/parser"
 	"go.uber.org/zap"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
@@ -566,8 +567,9 @@ func celMatcherJSONMacroExpander(funcName string) parser.MacroExpander {
 				if !isStringPlaceholder {
 					return nil, eh.NewError(entry.ID(), "matcher map keys must be string literals")
 				}
+				isNull := entry.AsMapEntry().Value().AsLiteral().Type() == types.NullType
 				isStringListPlaceholder := isCELStringExpr(entry.AsMapEntry().Value()) ||
-					isCELStringListLiteral(entry.AsMapEntry().Value())
+					isCELStringListLiteral(entry.AsMapEntry().Value()) || isNull
 				if !isStringListPlaceholder {
 					return nil, eh.NewError(entry.AsMapEntry().Value().ID(), "matcher map values must be string or list literals")
 				}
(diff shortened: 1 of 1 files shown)


mohammed90:
> So, works?

Doesn't: https://github.com/caddyserver/caddy/actions/runs/9120010471/job/25076559154?pr=6320#step:9:568

francislavoie:
I think this is making all strings nullable? I think that's probably not what we want :thinking: We might need to make a specific function to handle null but only for `handle` (or a different set of handlers). It's crashing with a SIGSEGV which is kinda wild but I guess not totally surprising with `null` lol

@TristonianJones FYI in case you have any suggestions. The gist of it is we're trying to allow `header({"X-Customer-Token": null})` I think (i.e. null values in "JSON-like" config, but probably only for this one matcher/function)

modules/caddyhttp/celmatcher.go
```diff
@@ -704,7 +708,7 @@ var (
 	placeholderRegexp    = regexp.MustCompile(`{([a-zA-Z][\w.-]+)}`)
 	placeholderExpansion = `caddyPlaceholder(request, "${1}")`
 
-	CELTypeJSON = cel.MapType(cel.StringType, cel.DynType)
+	CELTypeJSON = cel.MapType(cel.StringType, cel.AnyType)
```
TristonianJones, line 711: I wouldn't modify this line. The `DynType` is really more accurate here.

TristonianJones:
> I think this is making all strings nullable? I think that's probably not what we want 🤔 We might need to make a specific function to handle null but only for `handle` (or a different set of handlers). It's crashing with a SIGSEGV which is kinda wild but I guess not totally surprising with `null` lol
> 
> @TristonianJones FYI in case you have any suggestions. The gist of it is we're trying to allow `header({"X-Customer-Token": null})` I think (i.e. null values in "JSON-like" config, but probably only for this one matcher/function)

Hi @francislavoie, I think part of this issue is that the internal types are `map[string]string` and `map[string][]string` even though the CEL type is a `map(string,dyn)`. If you want to use `null` to represent absent, then you can absolutely do this; however, doing so would require handling `null` everywhere it could appear, at least as far as the macro is concerned. It's usually a bad idea to relax the constraints conditionally, in one case, but not all cases. Can you omit the key and use `'key-name' in map` instead as the expression for testing existence?

mohammed90:
If my memory serves me right, the user who reported this issue have applied a workaround. I don't remember what it was, probably @TristonianJones' suggestion. Anyways, I'll close this since the need is not dire.

PR #6324, opened 2024-05-17
caddytls: set SNI server name in context

willnorris:
Set the requested server name in a context value for certmagic.Manager implementations to use. Pass ctx to tscert.GetCertificateWithContext.

This relies on https://github.com/tailscale/tscert/pull/9 to be merged first, which adds the new tscert.GetCertificateWithContext method.  This will enable https://github.com/tailscale/caddy-tailscale/pull/53, which enables the use of auto_https with tsnet servers running inside of caddy.

This PR only attaches the server name inside the Tailscale certmagic.Manager.  I tried to find the right place to attach it for all managers, but couldn't seem to find it.  I'd love some pointers for where to put this.

/cc @mholt

### DIFF
--- a/modules/caddytls/certmanagers.go
+++ b/modules/caddytls/certmanagers.go
@@ -22,6 +22,9 @@ func init() {
 	caddy.RegisterModule(HTTPCertGetter{})
 }
 
+// For referencing the requested SNI server name.
+const ClientHelloSNICtxKey caddy.CtxKey = "client_hello_sni"
+
 // Tailscale is a module that can get certificates from the local Tailscale process.
 type Tailscale struct {
 	logger *zap.Logger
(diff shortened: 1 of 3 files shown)


modules/caddytls/certmanagers.go
```diff
@@ -22,6 +22,9 @@ func init() {
 	caddy.RegisterModule(HTTPCertGetter{})
 }
 
+// For referencing the requested SNI server name.
+var SNIServerNameCtxKey caddy.CtxKey = "sni_server_name"
```
mholt, line 26:
```suggestion
const ClientHelloSNICtxKey caddy.CtxKey = "client_hello_sni"
```

This is potentially bikeshedding but hear me out

mholt (changes requested):
Looks pretty straightforward -- how about this for a name instead?

Then I think we can merge.

modules/caddytls/certmanagers.go
francislavoie: I agree; `SNIServerName` is basically "server name indication server name" 😅 https://en.m.wikipedia.org/wiki/Server_Name_Indication

willnorris: lol, I didn't even think about what the expanding name would be.  Indeed, that sounds pretty silly :) Updated name coming up.

willnorris: done

francislavoie (approved):

mholt (approved):
Looks good now, thanks!

PR #6326, opened 2024-05-19
caddytls: remove ClientHelloSNICtxKey

willnorris:
It turns out this is unnecessary due to certmagic setting the entire ClientHelloInfo on the context using `certmagic.ClientHelloInfoCtxKey`.


francislavoie (approved):
Hah! I guess we all forgot about it. Thanks!

Issue #6327, opened 2024-05-19
caddytls: tailscale cert manager not used as fallback for *.ts.net certs


willnorris:
It appears that the tailscale cert manager is only used when a ts.net domain is explicitly configured for a site:

```
samson.tailb575b.ts.net:443 {
  respond OK
}
```

This works fine, and shows debug output of:

```
2024/05/19 01:44:25.371 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "samson.tailb575b.ts.net"}
2024/05/19 01:44:25.371 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.tailb575b.ts.net"}
2024/05/19 01:44:25.372 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.*.ts.net"}
2024/05/19 01:44:25.372 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.*.*.net"}
2024/05/19 01:44:25.372 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.*.*.*"}
2024/05/19 01:44:38.957 DEBUG   tls.handshake   using externally-managed certificate    {"remote_ip": "100.120.137.94", "remote_port": "53195", "sni": "samson.tailb575b.ts.net", "names": ["samson.tailb575b.ts.net"], "expiration": "2024/08/17 00:44:38.000"}
```

However, when configured as:

```
:443 {
  respond OK
}
```
I get debug logs:

```
2024/05/19 01:48:42.921 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "samson.tailb575b.ts.net"}
2024/05/19 01:48:42.921 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.tailb575b.ts.net"}
2024/05/19 01:48:42.921 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.*.ts.net"}
2024/05/19 01:48:42.921 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.*.*.net"}
2024/05/19 01:48:42.921 DEBUG   tls.handshake   no matching certificates and no custom selection logic  {"identifier": "*.*.*.*"}
[...]

Issue #6060, opened 2024-01-24
"get_certificate tailscale" forces "on_demand_tls" global option


TurboSlayer:
**Backstory**: I have a docker/caddy setup where I use a docker volume to expose the tailscaled socket to Caddy. It randomly stopped working today, so I tried to fix it. It looked like there were recent changes with socket paths so I tried different combinations with /tmp and /run but nothing worked. (EDIT: the error for this is `caddy_vaultwarden | {"level":"debug","ts":1706154566.203921,"logger":"http.stdlib","msg":"http: TLS handshake error from 100.107.69.109:41312: no certificate available for 'redacted.x.ts.net'"}`)

**The problem**: In order to debug this I tried enabling `get_certificate tailscale` under the tls directive, but then it starts enforcing the "ask" endpoint which doesn't make sense for tailscale users. An example of the error is [here](https://caddy.community/t/on-demand-fails-suddenly-in-2-7-4/21320).

I believe the "ask" enforcing is a fairly recent change, and so this may have been overlooked when the change was implemented. I know the docker volume approach isn't officially supported but it'd be great if this issue could be resolved so that I can try and debug it myself. Thanks!

francislavoie:
What version are you using? What's the error message you see, exactly? Show your logs & config.

TurboSlayer:
Yep, my bad, I was in a rush yesterday. Here's the info:
Version: `caddy:latest` on docker (v2.7.6)

Caddyfile:
```
{
    debug
}

:443 {
  log {
    level INFO
    output file {$LOG_FILE} {
      roll_size 10MB
      roll_keep 10
    }
  }
  tls {
    get_certificate tailscale # Didn't need this before (ever since the feature was released)
  }

   encode gzip
    
   reverse_proxy vaultwarden:80
   reverse_proxy /notifications/hub vaultwarden:3012
}
```
Error:
```
caddy_vaultwarden | {"level":"info","ts":1706153302.1186092,"msg":"using provided configuration","config_file":"/etc/caddy/Caddyfile","config_adapter":"caddyfile"}
caddy_vaultwarden | {"level":"warn","ts":1706153302.1194217,"msg":"Caddyfile input is not formatted; run 'caddy fmt --overwrite' to fix inconsistencies","adapter":"caddyfile","file":"/etc/caddy/Caddyfile","line":2}
[...]

PR #6328, opened 2024-05-19
caddytls: fix permission requirement with AutomationPolicy

willnorris:
Certificate automation has permission modules that are designed to prevent inappropriate issuance of unbounded or wildcard certificates. When an explicit cert manager is used, no additional permission should be necessary. For example, this should be a valid caddyfile:

    https:// {
      tls {
        get_certificate tailscale
      }
      respond OK
    }

This is accomplished when provisioning an AutomationPolicy by tracking whether there were explicit managers configured directly on the policy (in the ManagersRaw field). Only when a number of potentially unsafe conditions are present AND no explicit cert managers are configured is an error returned.

The problem arises from the fact that ctx.LoadModule deletes the raw bytes after loading in order to save memory. The first time an AutomationPolicy is provisioned, the ManagersRaw field is populated, and everything is fine.

An AutomationPolicy with no subjects is treated as a special "catch-all" policy. App.createAutomationPolicies ensures that this catch-all policy has an ACME issuer, and then calls its Provision method again because it may have changed. This second time Provision is called, ManagesRaw is no longer populated, and the permission check fails because it appears as though the policy has no explicit managers.

Address this by storing a new boolean on AutomationPolicy recording whether it had explicit cert managers configured on it.

Also fix an inverted boolean check on this value when setting failClosed.

Updates #6060
Updates #6229
Updates #6327

### DIFF
--- a/modules/caddytls/automation.go
+++ b/modules/caddytls/automation.go
@@ -170,6 +170,9 @@ type AutomationPolicy struct {
 	subjects []string
 	magic    *certmagic.Config
 	storage  certmagic.Storage
+
+	// Whether this policy had explicit managers configured directly on it.
+	hadExplicitManagers bool
 }
 
 // Provision sets up ap and builds its underlying CertMagic config.
@@ -201,8 +204,8 @@ func (ap *AutomationPolicy) Provision(tlsApp *TLS) error {
 	// store them on the policy before putting it on the config
 
 	// load and provision any cert manager modules
-	hadExplicitManagers := len(ap.ManagersRaw) > 0
 	if ap.ManagersRaw != nil {
+		ap.hadExplicitManagers = true
 		vals, err := tlsApp.ctx.LoadModule(ap, "ManagersRaw")
 		if err != nil {
 			return fmt.Errorf("loading external certificate manager modules: %v", err)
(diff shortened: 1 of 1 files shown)


francislavoie:
Oh wow, that's super subtle and tricky. Nice find.

I wonder if we should be resolving the underlying problem here though, i.e. it being Provision'd twice. That really should only happen one time. I'm not sure if we have a good way of doing that though.

But this workaround is probably sufficient for now. Thanks!

francislavoie (approved):

modules/caddytls/automation.go
```diff
@@ -262,9 +265,9 @@ func (ap *AutomationPolicy) Provision(tlsApp *TLS) error {
 		// prevent issuance from Issuers (when Managers don't provide a certificate) if there's no
 		// permission module configured
 		noProtections := ap.isWildcardOrDefault() && !ap.onlyInternalIssuer() && (tlsApp.Automation == nil || tlsApp.Automation.OnDemand == nil || tlsApp.Automation.OnDemand.permission == nil)
-		failClosed := noProtections && hadExplicitManagers // don't allow on-demand issuance (other than implicit managers) if no managers have been explicitly configured
+		failClosed := noProtections && !ap.hadExplicitManagers // don't allow on-demand issuance (other than implicit managers) if no managers have been explicitly configured
```
willnorris, line 268: I made this change based on the inline comment, and it seems to make sense.  However, now looking at it again, this is a noop since it is the same check that results in the returned error below.  So either the inline comment here and/or I am a little confused, or we can just remove this `failClosed` var entirely.  Admittedly, this is my first time digging into this section of code, so I may just be misunderstanding the intent.

willnorris:
> I wonder if we should be resolving the underlying problem here though, i.e. it being Provision'd twice.

Yeah, I had that thought as well.  I wasn't sure if it's generally best practice to avoid provisioning a module multiple times.  Though for what it's worth, this particular logic regarding re-provisioning the catch-all automation policy [is from 2020](https://github.com/caddyserver/caddy/commit/d692d503), so it's pretty well established.

mholt:
Ahh okay, so yeah... we do reprovision APs at times because we need to update the underlying CertMagic configs.

Let's give this a try... not really sure a cleaner/simpler way to do it right now, but would be interested with revisiting it if we find one more elegant :)

modules/caddytls/automation.go
mholt: I probably need more testing here :sweat_smile:

mholt (approved):

PR #6329, opened 2024-05-20
reverseproxy: add tls_server_cert_sha256

akovalenko:
Unfortunately there *are* some production setups requiring tls_insecure_skip_verify in reverse_proxy, like old devices with outdated firmware. In many such cases, the devices aren't supposed to regenerate or update their certificates.

This patch adds tls_server_cert_sha256 directive for reverse_proxy, making MITM impossible even with tls_insecure_skip_verify.

### DIFF
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -938,6 +938,7 @@ func (h *Handler) FinalizeUnmarshalCaddyfile(helper httpcaddyfile.Helper) error 
 //	    tls
 //	    tls_client_auth <automate_name> | <cert_file> <key_file>
 //	    tls_insecure_skip_verify
+//	    tls_server_cert_sha256 <fingerprint>
 //	    tls_timeout <duration>
 //	    tls_trusted_ca_certs <cert_files...>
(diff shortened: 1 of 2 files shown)


mholt (commented):
Thanks; is this a use case you've encountered? Could you elaborate?

I ask because while I appreciate that it tries to make something insecure secure, it also effectively pins keys which is... not secure. :upside_down_face:

steffenbusch:
I have a similar use case where I reverse-proxy to a Podman container with a self-signed certificate that has a very long lifetime.  
To avoid using `tls_insecure_skip_verify`, I use the `tls_trust_pool file /path/to/localhost_127_0_0_1.pem` directive instead, which also prevents MITM by leveraging existing functionality.  
You can find more details here: https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#tls_trust_pool

mholt:
@steffenbusch That's probably the preferred thing to do. @akovalenko does that work for you?

mholt:
I've been thinking on this and sitting on this and I don't love that it pins specific keys (which is discouraged), and it sounds like there are other mechanisms that work for people... @akovalenko I might close this, respectfully; but feel free to continue discussion and we can reopen this if necessary.

PR #6340, opened 2024-05-25
core: MkdirAll appDataDir in InstanceID with 0o700

akovalenko:
appDataDir components should be searchable (u+x) when they are created, or else Caddy is unable to start with an empty HOME.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -869,7 +869,7 @@ func InstanceID() (uuid.UUID, error) {
 		if err != nil {
 			return uuid, err
 		}
-		err = os.MkdirAll(appDataDir, 0o600)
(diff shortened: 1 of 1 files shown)


akovalenko:
Fixes #6341

mholt:
Thank you for the contribution! I'm still having trouble reproducing the original issue but as soon as we figure that out I'll merge this. (I actually feel this is right but I want to understand why I don't have the problem.)

mholt (approved):
While I'm still not able to reproduce this nor do I really understand who this affects, it's clearer to me now that this is more correct. Thanks for the patch.

manselmi:
@akovalenko Thank you for this PR!  I was also affected by this and thought I was loosing it…

PR #6356, opened 2024-05-31
cmd: fix auto-detetction of .caddyfile extension

### LINKED ISSUE #6355: using the file ext caddyfile for a config file does not work
```
./caddy run  -c test.caddyfile
2024/05/31 14:25:55.237 INFO    using config from file  {"file": "test.caddyfile"}
Error: config is not valid JSON: invalid character ':' looking for beginning of value; did you mean to use a config adapter (the --adapter flag)?
```

test.caddyfile
```
:2015
respond "Hello, world!"
```

mohammed90:
Fix #6355


francislavoie:
I think this existed to deal with https://github.com/caddyserver/caddy/pull/5919#discussion_r1425560197 i.e. `Caddyfile.yaml` if `yaml` is loaded as an adapter, should use that instead of the Caddyfile adapter.

mohammed90:
> I think this existed to deal with [#5919 (comment)](https://github.com/caddyserver/caddy/pull/5919#discussion_r1425560197) i.e. `Caddyfile.yaml` if `yaml` is loaded as an adapter, should use that instead of the Caddyfile adapter.

They would have specified the `--adapter` flag. I've moved the conditions around and added a comment.

mholt:
Are you sure though? The point of this is to presume when the Caddyfile is being used: `Caddyfile.yaml` is clearly ambiguous (and incorrect, since a yaml file isn't Caddyfile, and a Caddyfile isn't yaml) -- what happens if they _don't_ use the `--adapter` flag? Do we presume Caddyfile or YAML?

mohammed90:
> what happens if they _don't_ use the `--adapter` flag? Do we presume Caddyfile or YAML?

I can add a condition for such pesky scenarios 🙂 

```
if adapterName == "" && startsOrEndsInCaddyfile {
    return nil, "", fmt.Error("ambiguous... blablabbla")
}
```

mholt:
Ok, yeah, that sounds better. Maybe an error is better than ignoring it like we were doing. (But yeah, it's still important that we don't proceed, either way.)

mholt (approved):
Thanks Mohammed!

graphixillusion:
With this new modification, in version 2.8.2 my Caddyfile isn't recognized anymore. I need to force it using -c Caddyfile --adapter caddyfile. If not i get ambigous config message error. Same Caddyfile works without problems in 2.8.1

mohammed90:
> With this new modification, in version 2.8.2 my Caddyfile isn't recognized anymore. I need to force it using -c Caddyfile --adapter caddyfile. If not i get ambigous config message error. Same Caddyfile works without problems in 2.8.1

My bad. Let me work on the fix. I'm sorry.

mohammed90:
> You may need to modify the documents of your official website

That's not necessary. It's a regression. Please see the linked PR.

avluis:
Aha! I knew it; appreciate the very quick fix!

PR #6360, opened 2024-06-01
caddyhttp: properly sanitize requests for root path

### LINKED ISSUE #6352: the change to SanitizedPathJoin in v2.8.x can cause routing to break in certain configurations
in previous versions, this would work to route an SPA. i got this through google result showing matt's post here:

https://caddy.community/t/how-to-serve-a-single-page-app-spa-with-caddy-2/12770

```
:3838 {
        root * dir
	try_files {path} /
	file_server
}
```

however, in v2.8.0, this breaks. and must be replaced with

```
:3838 {
        root * dir
	try_files {path} /index.html
	file_server
}
```

after a git bisect, i find the commit where the behavior changed to be 
```
6d97d8d87beb788d19a4084d07ec9157e5705b13 is the first bad commit
commit 6d97d8d87beb788d19a4084d07ec9157e5705b13
Author: Matt Holt <mholt@users.noreply.github.com>
Date:   Tue Apr 23 20:05:57 2024 -0400

    caddyhttp: Address some Go 1.20 features (#6252)

    Co-authored-by: Francis Lavoie <lavofr@gmail.com>

 cmd/x509rootsfallback.go                     | 33 ++++++++++++++++++++++++++++
 go.mod                                       |  1 +
 go.sum                                       |  2 ++
 modules/caddyhttp/caddyhttp.go               | 13 +++++++++--
 modules/caddyhttp/caddyhttp_test.go          | 26 ++++++++++++++++------
 modules/caddyhttp/requestbody/caddyfile.go   | 26 +++++++++++++++++++++-
 modules/caddyhttp/requestbody/requestbody.go | 30 +++++++++++++++++++++++++
 modules/caddyhttp/responsewriter_test.go     | 10 ++++-----
 8 files changed, 125 insertions(+), 16 deletions(-)
 create mode 100644 cmd/x509rootsfallback.go
```
doing some more digging, but i thought i would raise this asap since perhaps other people were surprised by this change.

willnorris:
SanitizePathJoin protects against directory traversal attacks by checking for requests whose URL path look like they are trying to request something other than a local file, and returns the root directory in those cases.

The method is also careful to ensure that requests which contain a trailing slash include a trailing slash in the returned value.  However, for requests that contain only a slash (requests for the root path), the IsLocal check returns early before the matching trailing slash is re-added.

This change updates SanitizePathJoin to only perform the filepath.IsLocal check if the cleaned request URL path is non-empty.

---

This change also updates the existing SanitizePathJoin tests to use filepath.FromSlash rather than filepath.Join. This makes the expected value a little easier to read, but also has the advantage of not being processed by filepath.Clean like filepath.Join is. This means that the exact expect value will be compared, not the result of first cleaning it.

Fixes #6352


willnorris:
okay, after a little bit of test cleanup, I think this is ready to go now.

mholt (approved):
Thanks, I scanned this a few times and wondered how you handled the Windows cases but I see you didn't change those :+1:

This is quite an elegant and simple fix. I appreciate that! Thanks for the patch :)

PR #6365, opened 2024-06-03
chore: retract v2.8.3 in go.mod

mohammed90:
The tag was made in error at the wrong commit. 

Go modules reference for `retract`: https://go.dev/ref/mod#go-mod-file-retract

Are there others to retract?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,8 @@ go 1.21.0
 
 toolchain go1.22.2
 
+retract v2.8.3
+
 require (
 	github.com/BurntSushi/toml v1.3.2
 	github.com/Masterminds/sprig/v3 v3.2.3


francislavoie:
I'm not sure we need to care about it. We can leave it as-is. `.3` is the same as `.2`, it's harmless.

mholt:
Thanks for considering this!

I think I agree with Francis, though, the docs for retract say:

> A retract directive is useful when a version was published prematurely or a severe problem was discovered after the version was published.

I don't think either really apply here. Wasn't premature (just wrong commit), and not a severe problem (just repeated a tag unnecessarily).

I think I'll close this, but we can reopen if it turns out this has material advantages :) Thanks as always, Mohammed!

PR #6368, opened 2024-06-03
Make it possible to configure the `DisableStorageCheck` setting for certmagic

ankon:
See discussion about this setting in https://github.com/caddyserver/certmagic/issues/201

Note that this doesn't expose the setting through the caddyfile, I guess that would have to happen just-about-here: https://github.com/framer/caddy/blob/5b13f9272454ea13105689fbe24a14b00421f901/caddyconfig/httpcaddyfile/tlsapp.go#L331-L334

Not sure whether that's wanted/needed, and if it is, how to best name it?

I did consider another idea of injecting some sort of "configuration adapter" here, which would take a certmagic config, do whatever it needs to do, and then return the adapted config. This could over the long term be nicer, especially for these kind of special options -- but it would also open the world to a lot more mischief. WDYT?

### DIFF
--- a/modules/caddytls/tls.go
+++ b/modules/caddytls/tls.go
@@ -81,6 +81,16 @@ type TLS struct {
 	// EXPERIMENTAL. Subject to change.
 	DisableOCSPStapling bool `json:"disable_ocsp_stapling,omitempty"`
 
+	// Disables checks in certmagic that the configured storage is ready
+	// and able to handle writing new content to it. These checks are
+	// intended to prevent information loss (newly issued certificates), but
+	// can be expensive on the storage.
+	//
+	// Disabling these checks should only be done when the storage
+	// can be trusted to have enough capacity and no other problems.
+	// EXPERIMENTAL. Subject to change.
+	DisableStorageCheck bool `json:"disable_storage_check,omitempty"`
+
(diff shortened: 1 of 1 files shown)


modules/caddytls/tls.go
```diff
@@ -255,6 +265,7 @@ func (t *TLS) Provision(ctx caddy.Context) error {
 		OCSP: certmagic.OCSPConfig{
 			DisableStapling: t.DisableOCSPStapling,
 		},
+		DisableStorageCheck: t.DisableStorageChecks,
```
francislavoie, line 268: Typo, causing test failures

```suggestion
		DisableStorageCheck: t.DisableStorageCheck,
```

Are you not using an IDE with Go linting? It should have warned you about that before you made your commit :thinking: I recommend VSCode with the official Go extension, works very well.

ankon: I am using vscode ... but, sometimes I think I'm faster in `git commit && git push` than autosaving and linting is. Sorry :/

mholt (approved):
Nice, thanks. I hadn't quite gotten around to this (or it wasn't needed, can't remember). Anyway, I appreciate that it is marked as experimental. I want to improve storage checks so that they happen way less often, and maybe this won't be needed.

PR #6369, opened 2024-06-04
caddytls: Support new tls.context module

mholt:
This allows modules to manipulate the context passed into CertMagic's `GetCertificate` function, which can be useful for tracing/metrics, or other custom logic.

This is experimental and may resolve the request of a sponsor, so we'll see how it goes!


mholt:
This is being used successfully in production so I'll merge this.

PR #6381, opened 2024-06-08
Customizable logging zapcore.Core 

kkroo:
This change enables support custom `caddy.logging.cores` that implement `zapcore.Core` for a sentry event reporting plugin that reports errors and their traces at github.com/kkroo/caddy-sentry


mholt:
Thanks for the contribution!

Since we're teeing, I wonder if multiple cores (a slice of cores) would be worthwhile? Maybe not. Just thinking.

If not, this looks pretty good and we can probably merge it after a quick review. (I think godoc comments will be needed)

francislavoie:
I think an output module is still always needed, right? So I guess you'd want to use `output discard` along with this maybe, so you're not required to also write to a file or stderr if you want to send it to a module like Sentry? That's my main concern about tee-ing the core.

This also needs Caddyfile support I think, and it would be nice if we could have somekind of mock core module that we can use for testing configuration etc.

kkroo:
Thanks for the feedback, addressing comments .

@mholt An implementation of this interface could chain up multiple cores if needed. I can add the doc strings.

@francislavoie It uses the logging chain that has been set up. If the user doesn't want additional output `output discard`

modules/logging/cores.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"go.uber.org/zap/zapcore"
+)
+
+func init() {
+	caddy.RegisterModule(MockCore{})
+}
+
+// MockCore for testing
```
francislavoie, line 13: These godoc comments end up in https://caddyserver.com/docs/json/ so giving a clearer warning to not use it is a good idea I think.

```suggestion
// MockCore is a no-op module, purely for testing
// during development. Should not be used in apps.
```

logging.go
```diff
@@ -292,6 +292,9 @@ type BaseLog struct {
 	// The encoder is how the log entries are formatted or encoded.
 	EncoderRaw json.RawMessage `json:"encoder,omitempty" caddy:"namespace=caddy.logging.encoders inline_key=format"`
 
+	// Tees log entries through a zap.Core module
```
francislavoie, line 295: I think this could use a deeper explanation, similarly because it shows up in JSON docs. i.e. I think it would be good if here it said "what can you do with a Core that can't be done otherwise"

modules/logging/cores.go
```diff
@@ -0,0 +1,35 @@
+package logging
+
+import (
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"go.uber.org/zap/zapcore"
```
francislavoie, line 6: I _think_ the linter wants it in this order

```suggestion
	"go.uber.org/zap/zapcore"

	"github.com/caddyserver/caddy/v2"
	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
```

mholt (approved):
This LGTM now :) Thank you!

PR #6383, opened 2024-06-08
fix file mode configuration parsing

ririsoft:
Commit 101d3e7 introduced file mode setting,
but was missing a JSON Marshaller so that
CaddyFile can be converted to JSON safely.


mholt (approved):
Beautiful :) Thanks!

PR #6384, opened 2024-06-08
go.mod: update tscert package

willnorris:
The latest tscert [allows callers](https://github.com/tailscale/tscert/pull/11) to provide a custom http.Transport for calling Tailscale's local API.

Updates tailscale/caddy-tailscale#66

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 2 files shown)


willnorris:
Actually... I guess this doesn't **technically** have to get updated in caddy.  Or at the very least, I don't need to wait for a new caddy release.  I'll just udpate the dep in the caddy-tailscale plugin, and that will get used.  We should still update the version in caddy eventually, but it's not blocking anything.

mholt (approved):

mholt:
True; might as well merge this one, as I don't think any other plugins are really using this package that we'll break. Thanks Will!

PR #6388, opened 2024-06-12
reverseproxy: Allow `0` as weights for `weighted_round_robin`

peanutduck:
Enable 0 weights to be allowed when using weighted_round_robin policy with cookie. Discussed in https://caddy.community/t/invalid-weight-value/24440

### DIFF
--- a/modules/caddyhttp/reverseproxy/selectionpolicies.go
+++ b/modules/caddyhttp/reverseproxy/selectionpolicies.go
@@ -110,8 +110,8 @@ func (r *WeightedRoundRobinSelection) UnmarshalCaddyfile(d *caddyfile.Dispenser)
 		if err != nil {
 			return d.Errf("invalid weight value '%s': %v", weight, err)
 		}
-		if weightInt < 1 {
-			return d.Errf("invalid weight value '%s': weight should be non-zero and positive", weight)
+		if weightInt < 0 {
+			return d.Errf("invalid weight value '%s': weight should be non-negative", weight)
 		}
 		r.Weights = append(r.Weights, weightInt)
 	}


modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -110,8 +110,8 @@ func (r *WeightedRoundRobinSelection) UnmarshalCaddyfile(d *caddyfile.Dispenser)
 		if err != nil {
 			return d.Errf("invalid weight value '%s': %v", weight, err)
 		}
-		if weightInt < 1 {
-			return d.Errf("invalid weight value '%s': weight should be non-zero and positive", weight)
+		if weightInt < 0 {
+			return d.Errf("invalid weight value '%s': weight should be positive", weight)
```
mholt, line 114:
```suggestion
			return d.Errf("invalid weight value '%s': weight should be non-negative", weight)
```

0 isn't positive but is also allowed now :wink:

mholt (changes requested):
Thanks -- I wonder, does the actual selection algorithm need adjusting though to accommodate a 0 weight? Have you tested to ensure this works? (I guess we should consider adding test cases.)

modules/caddyhttp/reverseproxy/selectionpolicies.go
peanutduck: Good question, I don't have the enough knowledge to write the test case for this. Any guide would be helpful.

mholt: Have you at least tested various configs yourself to ensure they work? If so that's probably enough to start with.

But it looks like the existing tests are in `selectionpolicies_test.go` at `func TestWeightedRoundRobinPolicy(t *testing.T) {`. You could test a zero-weight in there as well.

peanutduck: Thank you for your reply, will work on generating the test case.

francislavoie: @peanutduck this still needs tests. Are you interested in finishing this up?

suciptoid: Hi @francislavoie 

I try to add test case to this PR here https://github.com/peanutduck/caddy/pull/1

francislavoie:
Superseded by #6681

PR #6391, opened 2024-06-12
logging: set file mode when the file already exist.

ririsoft:
101d3e7 introduced a configuration option to set the log file mode. This option was not taken into account if the file already exists, making users having to delete their logs to have new logs created with the right mode.

This was initially spotted in the discussion for 0bc27e5.
This will allow users to set the mode in CaddyFile and restart their server without the need to remove previous logs.

As suggested in the commit discussion I will open a separate PR to document that log files options require a server restart to be taken into account, when the log file destination does not change.

### DIFF
--- a/modules/logging/filewriter.go
+++ b/modules/logging/filewriter.go
@@ -167,8 +167,18 @@ func (fw FileWriter) OpenWriter() (io.WriteCloser, error) {
 			fw.RollKeepDays = 90
 		}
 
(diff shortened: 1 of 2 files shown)


ririsoft:
Documentation for the option submitted in caddyserver/website#399

mholt (approved):
Thank you! Let's give it a try.

PR #6394, opened 2024-06-13
Remove newline characters from precomputed etags

armadi1809:
This PR removes newline characters from etags acquired from precomputed files as per the discussion in #6222


modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -690,7 +690,12 @@ func (fsrv *FileServer) getEtagFromFile(fileSystem fs.FS, filename string) (stri
 		if err != nil {
 			return "", fmt.Errorf("cannot read etag from file %s: %v", etagFilename, err)
 		}
-		return string(etag), nil
+		etagStr := string(etag)
+
+		// Etags should not contain newline characters
+		etagStr = strings.ReplaceAll(etagStr, "\n", "")
+
+		return etagStr, nil
```
mohammed90, line 698: Maybe this?

```suggestion
		// Etags should not contain newline characters
		return strings.ReplaceAll(string(etag), "\n", ""), nil
```

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -690,7 +690,12 @@ func (fsrv *FileServer) getEtagFromFile(fileSystem fs.FS, filename string) (stri
 		if err != nil {
 			return "", fmt.Errorf("cannot read etag from file %s: %v", etagFilename, err)
 		}
-		return string(etag), nil
+		etagStr := string(etag)
+
+		// Etags should not contain newline characters
+		etagStr = strings.ReplaceAll(etagStr, "\n", "")
```
mholt, line 696: Instead of converting to string, use `bytes.ReplaceAll()`.

modules/caddyhttp/fileserver/staticfiles.go
mohammed90: Matt's suggestion is better. Ignore this.

armadi1809:
The replacement of the newline character is now done through `bytes.ReplaceAll`.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -690,6 +691,10 @@ func (fsrv *FileServer) getEtagFromFile(fileSystem fs.FS, filename string) (stri
 		if err != nil {
 			return "", fmt.Errorf("cannot read etag from file %s: %v", etagFilename, err)
 		}
+
+		// Etags should not contain newline characters
+		etag = bytes.ReplaceAll(etag, []byte("\n"), []byte(""))
```
mholt, line 696:
```suggestion
		etag = bytes.ReplaceAll(etag, []byte("\n"), []byte{})
```

mholt (approved):
Thanks! LGTM now. I just made one quick stylistic change (IMO there's no need to convert `""` to `[]byte`, so we can just use an empty byte slice.)

Issue #6393, opened 2024-06-13
Load balancer sticky cookie should have a Max-Age configuration


JasonYuan869:
I have a service routed with the following
```
reverse_proxy backend-blue:3000 backend-green:3000 {
    lb_policy cookie {
        fallback first
    }
}
```
This allows me to do zero-downtime upgrades on a stateful service by switching the order of the upstreams.

I think it would be useful to allow configuration of a `Max-Age` property for the cookie to ensure that eventually a client application will use the new backend when the cookie expires.

mholt:
Thanks for the issue. Do you have any suggestions as to what the config would look like for that?

JasonYuan869:
Perhaps something like this?
```
reverse_proxy backend-blue:3000 backend-green:3000 {
    lb_policy cookie {
        fallback first
        max_age 3600  # 1 hour
    }
}
```
It definitely makes more sense to use `Max-Age` over `Expires` since we're not trying to hardcode a specific expiration time but rather tell the cookie to refresh after X duration.

mholt:
That LGTM. Want to submit a PR? :smiley:

mohammed90:
Consider using [`caddy.Duration`](https://pkg.go.dev/github.com/caddyserver/caddy/v2#Duration)

PR #6398, opened 2024-06-14
reverseproxy: add Max-Age option to sticky cookie

JasonYuan869:
Addresses #6393

### DIFF
--- a/modules/caddyhttp/reverseproxy/selectionpolicies.go
+++ b/modules/caddyhttp/reverseproxy/selectionpolicies.go
@@ -705,7 +705,7 @@ func (s CookieHashSelection) Select(pool UpstreamPool, req *http.Request, w http
 //
 //	lb_policy cookie [<name> [<secret>]] {
 //		fallback <policy>
-//		[max_age <duration>]
+//		max_age <duration>
 //	}
 //
 // `max_age` is a positive (non-zero) `caddy.Duration`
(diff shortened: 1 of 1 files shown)


mholt (approved):
LGTM -- thank you! We just need the imports reordered, I think, to comply with linter.

modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -699,8 +705,10 @@ func (s CookieHashSelection) Select(pool UpstreamPool, req *http.Request, w http
 //
 //	lb_policy cookie [<name> [<secret>]] {
 //		fallback <policy>
+//		[max_age <duration>]
```
francislavoie, line 708: All subdirectives are implicitly optional, so the `[]` isn't needed here

```suggestion
//		max_age <duration>
```

modules/caddyhttp/reverseproxy/selectionpolicies.go
```diff
@@ -699,8 +705,10 @@ func (s CookieHashSelection) Select(pool UpstreamPool, req *http.Request, w http
...
 //	}
 //
+// `max_age` is a positive (non-zero) `caddy.Duration`
```
francislavoie, line 711: I don't think this needs saying tbh, it's implied by the `<duration>`. We'll document it as such in the docs website as well.

```suggestion
```

mholt (approved):

PR #6405, opened 2024-06-19
caddytest changes round 1

elee1766:


elee1766:
1. see cmd/commandfactory for the command factory. Basically you pass in pike-style options and it recomputes the cmd when you call Build(). 
2. the old Tester struct has been renamed to Harness, and the new tester struct does not specifically work with a *testing.T

logging.go
```diff
@@ -699,7 +700,13 @@ type defaultCustomLog struct {
 // and enables INFO-level logs and higher.
 func newDefaultProductionLog() (*defaultCustomLog, error) {
 	cl := new(CustomLog)
-	cl.writerOpener = StderrWriter{}
+	f := flag.Lookup("test.v")
+	if (f != nil && f.Value.String() != "true") || strings.Contains(os.Args[0], ".test") {
+		cl.writerOpener = &DiscardWriter{}
+	} else {
+		cl.writerOpener = StderrWriter{}
+	}
```
francislavoie, line 708: I think we do want stderr logs during tests tho, cause it can help trace why a failure happened. Pretty sure `go test` skips outputting them if there's no failure (unless `-v` is set)

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -691,10 +690,6 @@ func (fsrv *FileServer) getEtagFromFile(fileSystem fs.FS, filename string) (stri
 		if err != nil {
 			return "", fmt.Errorf("cannot read etag from file %s: %v", etagFilename, err)
 		}
-
-		// Etags should not contain newline characters
-		etag = bytes.ReplaceAll(etag, []byte("\n"), []byte{})
```
francislavoie, line 696: This looks unrelated? Was this a mistake in rebase?

caddytest/caddytest_assert.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/aryann/difflib"
+	"github.com/stretchr/testify/require"
```
mohammed90, line 12: I just noticed testify is in Caddy's go.mod. The only other usage of it is for the function `assert.JSONEq`, which isn't complex (merely marshals into interface and deep-compares). We can probably get rid of it if the rest of the team agrees.

caddytest/integration/stream_test.go
```diff
@@ -358,9 +351,9 @@ func TestH2ToH1ChunkedResponse(t *testing.T) {
 	}
 }
 
-func testH2ToH1ChunkedResponseServeH1(t *testing.T) *http.Server {
+func testH2ToH1ChunkedResponseServeH1(harness *caddytest.TestHarness, t *testing.T) *http.Server {
```
mohammed90, line 354: Need this be a pointer?

elee1766: yes, in case we add something that can't be passed by value, or modify testharness to be a interface.

 there is no reason to pass the bare struct

caddytest/caddytest_assert.go
elee1766: yeah i used it because it was already in the go.mod.

fine with removing it if you guys dont like it

modules/caddyhttp/fileserver/staticfiles.go
elee1766: yep. will fix.

logging.go
elee1766: sounds fine to me. should be resolved.

i will say that i do find it somewhat odd that caddy logs in tests by default. i generally expect `go test` to run silently.

caddytest/caddytest_assert.go
mholt: I'm in favor of slimming down our dependencies, especially if it means a change in one other place.

mholt:
Yeah, I am not too worried about breakage at this point, since these are just test APIs, and we've never really formalized them or intended them to be used much. I'm hoping these new changes will be more permanent, more flexible, more useful to plugin authors, etc, that we can proudly document and demonstrate.

Thanks for working on this! I have a big backlog but will hope to join the reviews at some point soon. Don't let me hold things back though.

elee1766:
so i waas able to merge master

there is a special piece of code that makes it so that if you are running in a test, /stop does not call os.exit(1)
but the thing is that a previously exiting sets exiting to 1, which makes it so that future run caddy in the same process will never exit

so i made it just ignore that for now.. hopefully this doesnt mess up anything. but its another quirk to keep note of 

i think it would be really really nice to be able to be able to remove as many globals as possible with caddy, or at the very least, make global defaults opt in, so testing code can avoid it.

Issue #6409, opened 2024-06-19
[file_server] Incorrect file size calculation


IgorKha:
hello!
I noticed incorrect file size calculation on the web page, and it would be cool if symbolic links could be seen more explicitly

My config:
```config
:80 {
	root * /mnt/local_share/
	file_server browse
}
```

```bash
user@runner:/mnt/local_share/napi-dev/napi-rk3308b-s$ tree --du -h
[627M]  .
├── [  94]  napi-rk3308b-s-latest-dev.img.xz -> /mnt/local_share/napi-dev/napi-rk3308b-s/nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.img.xz
├── [ 101]  napi-rk3308b-s-latest-dev.img.xz.sha256 -> /mnt/local_share/napi-dev/napi-rk3308b-s/nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.img.xz.sha256
├── [  96]  napi-rk3308b-s-latest-dev.manifest -> /mnt/local_share/napi-dev/napi-rk3308b-s/nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.manifest
├── [  91]  napi-rk3308b-s-latest-dev.swu -> /mnt/local_share/napi-dev/napi-rk3308b-s/nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.swu
├── [  98]  napi-rk3308b-s-latest-dev.swu.sha256 -> /mnt/local_share/napi-dev/napi-rk3308b-s/nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.swu.sha256
├── [357M]  nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.img.xz
├── [ 120]  nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.img.xz.sha256
├── [116K]  nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.manifest
├── [270M]  nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.swu
└── [ 117]  nnz-napi-image-dev-napi-rk3308b-s-dev-ba99a1e5.swu.sha256

 627M used in 1 directory, 10 files
```

[...]

PR #6412, opened 2024-06-19
fileserver: Exclude symlink target size from total, show arrow on size

francislavoie:
Closes https://github.com/caddyserver/caddy/issues/6409

image

### DIFF
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -980,7 +980,7 @@ footer {
 							<div class="sizebar">
 								<div class="sizebar-bar"></div>
 								<div class="sizebar-text">
-									{{if .IsSymlink}}↱ {{end}}{{.HumanSize}}
+									{{if .IsSymlink}}↱&nbsp;{{end}}{{.HumanSize}}
 								</div>
 							</div>
 						</td>
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -980,7 +980,7 @@ footer {
 							<div class="sizebar">
 								<div class="sizebar-bar"></div>
 								<div class="sizebar-text">
-									{{.HumanSize}}
+									{{if .IsSymlink}}↱ {{end}}{{.HumanSize}}
 								</div>
 							</div>
 						</td>
--- a/modules/caddyhttp/fileserver/browsetplcontext.go
+++ b/modules/caddyhttp/fileserver/browsetplcontext.go
@@ -111,7 +111,7 @@ func (fsrv *FileServer) directoryListing(ctx context.Context, fileSystem fs.FS, 
 
 		if !isDir {
 			// increase the total including the symlink target's size
-			tplCtx.TotalFileSizeWithSymlinks += size
+			tplCtx.TotalFileSizeFollowingSymlinks += size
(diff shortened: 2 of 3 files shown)


steffenbusch:
Can't you at least add a `browseTemplateContext` method to retrieve the current total size of all files in the listing, so I (sponsor) can at least show what should be shown to my users with my custom browse.html?

francislavoie:
You mean calculate both totals? I guess we can.

modules/caddyhttp/fileserver/browsetplcontext.go
```diff
@@ -150,9 +158,14 @@ type browseTemplateContext struct {
...
 	NumFiles int `json:"num_files"`
 
-	// The total size of all files in the listing.
+	// The total size of all files in the listing. Only includes the
+	// size of the files themselves, not the size of the symlinks.
 	TotalFileSize int64 `json:"total_file_size"`
 
+	// The total size of all files in the listing, including the
+	// size of the files targeted by symlinks.
+	TotalFileSizeWithSymlinks int64 `json:"total_file_size_with_symlinks"`
```
mholt, line 167: Without looking at the code, these comments are confusing to me. Symlinks are files, so I'm not sure which one has the size of the links themselves or which one has the size of the targets, or whether one skips them altogether.

Some docs + naming clarifications are probably in order here; maybe like `TotalFileSizeFollowSymlinks` or something.

(And what about Hard Links?)

mholt (changes requested):
Nice, I like this idea overall. Thanks for the patch.

Just some questions regarding nuance / implementation.

modules/caddyhttp/fileserver/browsetplcontext.go
IgorKha: Since hard links use the same inode and point to the same data block, it would be correct not to count them. However, the problem is that they are difficult to identify because they are essentially just like regular files. It is unclear how to determine in which directory they should be counted and in which they should not, as they do not have additional metadata.

mholt: Ah... I see. Ok then let's not worry about hard links.

francislavoie: I'm probably out of my depth on this, feel free to make the changes you want to see 👍

modules/caddyhttp/fileserver/browse.html
```diff
@@ -980,7 +980,7 @@ <h1>
 							<div class="sizebar">
 								<div class="sizebar-bar"></div>
 								<div class="sizebar-text">
-									{{.HumanSize}}
+									{{if .IsSymlink}}↱ {{end}}{{.HumanSize}}
```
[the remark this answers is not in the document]
francislavoie:
```suggestion
									{{if .IsSymlink}}↱&nbsp;{{end}}{{.HumanSize}}
```

modules/caddyhttp/fileserver/browsetplcontext.go
IgorKha: I would love to, but unfortunately, I don't write in Go 😕

mholt: I will push some changes soon, no worries -- just renaming some things.

PR #6417, opened 2024-06-22
feat: Improve OpenTelemetry trace instrumentation

cedricziel:
This PR enhances the opentelemetry instrumentation for caddy by:

creating client spans for reverse proxy request to fix servicegraph calculations
injecting, conditionally, a server-timing header to propagate the traceparent down to the caller

### DIFF
--- a/modules/caddyhttp/tracing/module.go
+++ b/modules/caddyhttp/tracing/module.go
@@ -25,7 +25,8 @@ func init() {
 type Tracing struct {
 	// SpanName is a span name. It should follow the naming guidelines here:
 	// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span
-	SpanName string `json:"span"`
+	SpanName                 string `json:"span"`
+	InjectServerTimingHeader bool   `json:"injectServerTimingHeader"`
 
 	// otel implements opentelemetry related logic.
 	otel openTelemetryWrapper
@@ -46,7 +47,7 @@ func (ot *Tracing) Provision(ctx caddy.Context) error {
 	ot.logger = ctx.Logger()
 
 	var err error
-	ot.otel, err = newOpenTelemetryWrapper(ctx, ot.SpanName)
+	ot.otel, err = newOpenTelemetryWrapper(ctx, ot.SpanName, ot.InjectServerTimingHeader)
 
 	return err
 }
--- a/modules/caddyhttp/tracing/tracer.go
+++ b/modules/caddyhttp/tracing/tracer.go
@@ -88,11 +88,15 @@ func (ot *openTelemetryWrapper) serveHTTP(w http.ResponseWriter, r *http.Request
 	spanCtx := trace.SpanContextFromContext(ctx)
 	if spanCtx.IsValid() {
 		traceID := spanCtx.TraceID().String()
+		spanID := spanCtx.SpanID().String()
 		// Add a trace_id placeholder, accessible via `{http.vars.trace_id}`.
 		caddyhttp.SetVar(ctx, "trace_id", traceID)
+		// Add a span_id placeholder, accessible via `{http.vars.span_id}`.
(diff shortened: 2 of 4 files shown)


modules/caddyhttp/tracing/module.go
```diff
@@ -25,7 +25,8 @@ func init() {
 type Tracing struct {
 	// SpanName is a span name. It should follow the naming guidelines here:
 	// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span
-	SpanName string `json:"span"`
+	SpanName                 string `json:"span"`
+	InjectServerTimingHeader bool   `json:"injectServerTimingHeader"`
```
mohammed90, line 29: If `false` is the assumed default, `omitempty` cleans up the adapted JSON. 

```suggestion
	InjectServerTimingHeader bool   `json:"injectServerTimingHeader,omitempty"`
```
We also prefer snake-case. Is this a common label? Or can it be shorter?

Disclaimer: I'm not familiar with otel

cedricziel: corrected, thank you!

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -32,6 +32,8 @@ import (
 	"sync"
 	"time"
 
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+
 	"go.uber.org/zap"
 	"golang.org/x/net/http/httpguts"
```
mohammed90, line 38: If my memory serves me right, the linter prefers them sorted by length. Move the `otelhttp` to below `httpguts` without any spaces in between.

cedricziel: that's a good hint, thank you - I was frantically installing random linters to make golangci-lint work locally :D 

btw: is there a guide on how to automatically format it into shape and what my environment needs to make that happen?

mohammed90: If you run `golangci-lint run --fix`, it should take care of most of them. The error in CI console also gives the command to run to fix it. I believe the Go plugin of vscode manages the formatting as well.

ChenYujunjks:
hello

modules/caddyhttp/tracing/tracer.go
```diff
@@ -84,15 +86,24 @@ func newOpenTelemetryWrapper(
...
+		caddyhttp.SetVar(ctx, "span_id", spanID)
 		// Add the trace id to the log fields for the request.
 		if extra, ok := ctx.Value(caddyhttp.ExtraLogFieldsCtxKey).(*caddyhttp.ExtraLogFields); ok {
 			extra.Add(zap.String("traceID", traceID))
+			extra.Add(zap.String("spanID", spanID))
+		}
+
+		// Add the server-timing header so clients can make the connection
+		if ot.injectServerTimingHeader {
+			w.Header().Set("server-timing", fmt.Sprintf("traceparent;desc=\"00-%s-%s-%s\"", traceID, spanID, spanCtx.TraceFlags().String()))
```
hairyhenderson, line 106: I'm curious why this isn't the w3c-standard `traceparent` header instead? is that because this is a response?

hairyhenderson: I just now realized that we had a conversation about this header in DMs back in July... LGTM 😂

hairyhenderson (approved):

mholt:
Thanks @hairyhenderson !

@cedricziel Would you be able to fix up the lint errors and make sure the tests pass, then we can get this in the next beta release?

mholt:
Thanks for the commit!

Looks like the order of the modified imports needs to be adjusted (for lint) and for tests, there's:

```
    module_test.go:107: Invalid traceparent: 
--- FAIL: TestTracing_ServeHTTP_Propagation_Without_Initial_Headers (0.00s)
```

PR #6425, opened 2024-06-29
browse: add Content-Security-Policy w/ nonce

steffenbusch:
This pull request is adding a Content-Security-Policy (CSP) response header to the file server **browse** template. The CSP Version 3 is using `strict-dynamic` for `script-src` and `style-src` with a generated, unique `nonce`, which is then used in `<script>` and `<style>` to whitelist the content of such elements.

Also:
* Styles from `svg` such as the caddy-logo, have been changed into svg-attributes. 
* Moved inline JavaScript from `body` or `href` attribute to a dedicated <script> section with an event listener. 

See also: https://caddy.community/t/best-practice-csp-for-file-server-browse/24714

Right now, the CSP is enabled by default with `$enableCsp := true`, so it would be also rather easy to disable it with this single change in a custom browse templates.

At the moment, this browse template is active here: https://alma.stbu.net/testing-something/


modules/caddyhttp/fileserver/browse.html
```diff
@@ -799,15 +806,15 @@ <h1>
 						</span>
 						{{- end}}
 					</div>
-					<a href="javascript:queryParam('layout', '')" id="layout-list" class='layout{{if eq $.Layout "list" ""}}current{{end}}'>
+					<a href="#" id="layout-list" class='layout{{if eq $.Layout "list" ""}}current{{end}}'>
```
mholt, line 809: One thing about `#` is that it changes the browser history, I believe. Can we replace this with `javascript:` without breaking CSP? `href="javascript:"`

Or at least use a ... what is it these days, preventDefault() or stopPropagation() or something? On the JS event handler, to prevent the actual page nav.

mholt (changes requested):
Thanks for the clever enhancement! I like this a lot. Just one question for you, then I think we can merge this.

mholt:
Also... lol, I love all these test cases: https://alma.stbu.net/testing-something/test-cases/

modules/caddyhttp/fileserver/browse.html
steffenbusch: I can remove the `href="#"` from both layout anchors and add a CSS rule to indicate that the element is clickable.

mholt (approved):

mholt:
I did notice there's still CSP warnings in Firefox:

```
Content-Security-Policy: Ignoring “'unsafe-inline'” within script-src: ‘strict-dynamic’ specified [testing-something](https://alma.stbu.net/testing-something/)
Content-Security-Policy: Ignoring “https:” within script-src: ‘strict-dynamic’ specified [testing-something](https://alma.stbu.net/testing-something/)
Content-Security-Policy: Ignoring “http:” within script-src: ‘strict-dynamic’ specified [testing-something](https://alma.stbu.net/testing-something/)
Content-Security-Policy: Ignoring “'unsafe-inline'” within script-src: nonce-source or hash-source specified [testing-something](https://alma.stbu.net/testing-something/)
Content-Security-Policy: Ignoring source “strict-dynamic” (Only supported within script-src). [testing-something](https://alma.stbu.net/testing-something/)
Content-Security-Policy: Ignoring ‘block-all-mixed-content’ because mixed content display upgrading makes block-all-mixed-content obsolete.
```

steffenbusch:
> I did notice there's still CSP warnings in Firefox:
> 
> ```
> Content-Security-Policy: Ignoring “'unsafe-inline'” within script-src: ‘strict-dynamic’ specified [testing-something](https://alma.stbu.net/testing-something/)
> Content-Security-Policy: Ignoring “https:” within script-src: ‘strict-dynamic’ specified [testing-something](https://alma.stbu.net/testing-something/)
> Content-Security-Policy: Ignoring “http:” within script-src: ‘strict-dynamic’ specified [testing-something](https://alma.stbu.net/testing-something/)
> Content-Security-Policy: Ignoring “'unsafe-inline'” within script-src: nonce-source or hash-source specified [testing-something](https://alma.stbu.net/testing-something/)
> Content-Security-Policy: Ignoring source “strict-dynamic” (Only supported within script-src). [testing-something](https://alma.stbu.net/testing-something/)
> Content-Security-Policy: Ignoring ‘block-all-mixed-content’ because mixed content display upgrading makes block-all-mixed-content obsolete.
> ```

I will have a look on this.

steffenbusch:
I will make the following changes to the Content Security Policy to address specific warnings:

1. **Remove `strict-dynamic` from `style-src`**:
   > Content-Security-Policy: Ignoring source “strict-dynamic” (Only supported within script-src).

2. **Remove `block-all-mixed-content`**:
   > Content-Security-Policy: Ignoring ‘block-all-mixed-content’ because mixed content display upgrading makes block-all-mixed-content obsolete.

See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content

3. **Remove `strict-dynamic`, `https:`, and `http:` from `script-src`**:
   > Content-Security-Policy: Ignoring “'unsafe-inline'” within script-src: ‘strict-dynamic’ specified
   > Content-Security-Policy: Ignoring “https:” within script-src: ‘strict-dynamic’ specified
   > Content-Security-Policy: Ignoring “http:” within script-src: ‘strict-dynamic’ specified

After these changes, only the following warning will remain due to the presence of `'unsafe-inline'` within `script-src` for backward compatibility with browsers not supporting Content-Security-Policy Version 3 (such as Internet Explorer 11):

> Content-Security-Policy: Ignoring “'unsafe-inline'” within script-src: nonce-source or hash-source specified

Unfortunately, Firefox will log this directive being ignored. We must make a trade-off between this warning and maintaining backward compatibility.

See also: https://developer.chrome.com/docs/lighthouse/best-practices/csp-xss/#ensure_csp_is_backwards_compatible

PR #6429, opened 2024-07-01
fix: http.intercept.header.* placeholders

dunglas:
I introduced a bug in #6232: `http.intercept.header.*` placeholders contain request headers instead of original response headers.

This patch fixes the issue.

Closes https://github.com/dunglas/frankenphp/issues/884.


mholt (approved):
Thanks Kevin!

PR #6432, opened 2024-07-03
fix: don't compress already compressed fonts

### LINKED ISSUE #6428: "Encode" defaults compress already compressed font formats
The default matcher for `encode` currently includes `font/*`, matching every font format. However, the modern WOFF and WOFF2 web font formats are already internally compressed with zlib and brotli respectively, so compressing them again is likely a waste of CPU cycles. Would it make sense to refine the defaults down to only compress the "legacy" non-WOFF formats?

https://www.iana.org/assignments/media-types/media-types.xhtml#font

dunglas:
Fixes a regression introduced in https://github.com/caddyserver/caddy/pull/6081.

Closes #6428.


modules/caddyhttp/encode/encode.go
```diff
@@ -112,7 +112,9 @@ func (enc *Encode) Provision(ctx caddy.Context) error {
 					"application/x-ttf*",
 					"application/xhtml+xml*",
 					"application/xml*",
-					"font/*",
+					"font/ttf*",
+					"font/otf*",
+					"font/x-woff*",
```
jsheard, line 117: This line isn't needed, WOFF is pre-compressed

dunglas: According to https://community.cloudflare.com/t/cloudflare-is-missing-the-current-woff-media-type-in-the-default-compression-config/679095, this is beneficial if Brotli (and I guess Zstandard) are used.

Brotli isn't natively supported by Caddy, but FrankenPHP and https://github.com/dunglas/caddy-cbrotli allow us to easily add support for it. Zstandard is supported and is now on by default for Chrome and Firefox (not for Safari).

jsheard: Fair enough, but if WOFF1 is to be compressed then the matcher should be the standardized `font/woff` instead of or in addition to miscellaneous names it went by before being standardized.

dunglas: TBH, I just copied the list provided by Cloudflare: https://developers.cloudflare.com/speed/optimization/content/brotli/content-compression/#compression-between-cloudflare-and-website-visitors

jsheard: Also won't that wildcard on the end cause it to match WOFF2 anyway? AFAICT that wildcard is there in case the server appends a charset to the content type, but these are binary formats so that shouldn't happen.

dunglas: The MIME type of WOFF2 `font/woff2`, so it will not match, but I can remove this MIME type if you want.

dunglas: WOFF removed

jsheard (approved):

modules/caddyhttp/encode/encode.go
mholt: Thanks; hopefully that's the best decision, but I'm not really sure, but it sounds like there is no real consensus anyway. Both WOFF versions are compressed, even if WOFF2 has "better" compression; but I don't see a need to compress either of them.

Appreciate it!

mholt (approved):

PR #6440, opened 2024-07-06
reverseproxy: Add placeholder for host in active health check headers

francislavoie:
Should resolve https://caddy.community/t/can-you-pass-in-the-full-upstream-to-an-active-health-check/24718/3


mholt (approved):
LGTM! Thanks

PR #6448, opened 2024-07-09
reverseproxy: Fix dynamic upstreams ip version

### LINKED ISSUE #6442: can't disable IPv4/IPv6 for dynamic reverse proxy from Caddyfile
As per the forum discussion at https://caddy.community/t/reverse-proxy-to-dynamic-ipv4-only/24773/3 , I found it's not possible to disable IPv4 or IPv6 in a `reverse_proxy` `dynamic a` block from the Caddyfile.

If I use
```
:9000 {
	reverse_proxy {
		dynamic a potoroo 8123 {
			versions ipv4
		}
	}
}
```

then I get a caddy.json which has `"versions":{"ipv4":true}}`, but Caddy still uses IPv6 if available. 

If I explicitly add `"ipv6":"false"` then it works, but this can't be set from the Caddyfile (that I'm aware of).

armadi1809:
Closes #6442

### DIFF
--- a/modules/caddyhttp/reverseproxy/upstreams.go
+++ b/modules/caddyhttp/reverseproxy/upstreams.go
@@ -231,6 +231,19 @@ type IPVersions struct {
 	IPv6 *bool `json:"ipv6,omitempty"`
 }
 
+func resolveIpVersion(versions *IPVersions) string {
+	resolveIpv4 := versions == nil || (versions.IPv4 == nil && versions.IPv6 == nil) || (versions.IPv4 != nil && *versions.IPv4)
+	resolveIpv6 := versions == nil || (versions.IPv6 == nil && versions.IPv4 == nil) || (versions.IPv6 != nil && *versions.IPv6)
+	switch {
+	case resolveIpv4 && !resolveIpv6:
+		return "ip4"
+	case !resolveIpv4 && resolveIpv6:
+		return "ip6"
+	default:
+		return "ip"
+	}
+}
+
 // AUpstreams provides upstreams from A/AAAA lookups.
 // Results are cached and refreshed at the configured
 // refresh interval.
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -313,8 +313,8 @@ func (au *AUpstreams) Provision(ctx caddy.Context) error {
 func (au AUpstreams) GetUpstreams(r *http.Request) ([]*Upstream, error) {
 	repl := r.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
 
-	resolveIpv4 := au.Versions == nil || au.Versions.IPv4 == nil || *au.Versions.IPv4
-	resolveIpv6 := au.Versions == nil || au.Versions.IPv6 == nil || *au.Versions.IPv6
+	resolveIpv4 := au.Versions == nil || (au.Versions.IPv4 == nil && au.Versions.IPv6 == nil) || (au.Versions.IPv6 != nil && *au.Versions.IPv4)
```
francislavoie, line 316:
```suggestion
	resolveIpv4 := au.Versions == nil || (au.Versions.IPv4 == nil && au.Versions.IPv6 == nil) || (au.Versions.IPv4 != nil && *au.Versions.IPv4)
```

francislavoie:
I think a test would be useful for this if you have time to look into that 🤔 we could refactor this logic to a separate function to return a tuple of booleans so it can be tested as a unit.

armadi1809:
@francislavoie I just refactored the logic determining the ip version into its own function and added a test covering all possible combinations of `au.Versions`.

modules/caddyhttp/reverseproxy/upstreams.go
```diff
@@ -397,6 +386,19 @@ func (au AUpstreams) GetUpstreams(r *http.Request) ([]*Upstream, error) {
 
 func (au AUpstreams) String() string { return net.JoinHostPort(au.Name, au.Port) }
 
+func resolveIpVersion(versions *IPVersions) string {
```
francislavoie, line 389: I think we should move this function to just below the IPVersions type (higher in this file) since it's tightly related.

francislavoie (approved):
Thanks! Looks great 😊

modules/caddyhttp/reverseproxy/upstreams.go
armadi1809: Done.

calvinbrewer:
Still running into this issue, but it looks like a release hasn't been cut since June 2nd. What is the release cadence for deploying these new changes? Cheers!

PR #6451, opened 2024-07-09
reverseproxy: add health_upstream subdirective

dylanschultzie:
Currently in draft to update the documentation with the new subdirective.

Respective docs PR: https://github.com/caddyserver/website/pull/406

Log when `health_port` and `health_upstream` is used:

Screenshot 2024-07-09 at 2 49 23 PM

### DIFF
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -385,7 +385,7 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				h.HealthChecks.Active = new(ActiveHealthChecks)
 			}
 			if h.HealthChecks.Active.Upstream != "" {
-				caddy.Log().Named("config.adapter.caddyfile").Warn("the 'health_port' subdirective is ignored if 'health_upstream' is used!")
+				return d.Errf("the 'health_port' subdirective is ignored if 'health_upstream' is used!")
 			}
 			portNum, err := strconv.Atoi(d.Val())
 			if err != nil {
--- a/modules/caddyhttp/reverseproxy/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/caddyfile.go
@@ -16,6 +16,7 @@ package reverseproxy
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"reflect"
 	"strconv"
@@ -353,6 +354,26 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 			h.HealthChecks.Active.Path = d.Val()
 			caddy.Log().Named("config.adapter.caddyfile").Warn("the 'health_path' subdirective is deprecated, please use 'health_uri' instead!")
 
+		case "health_upstream":
+			if !d.NextArg() {
+				return d.ArgErr()
+			}
+			if h.HealthChecks == nil {
+				h.HealthChecks = new(HealthChecks)
+			}
+			if h.HealthChecks.Active == nil {
+				h.HealthChecks.Active = new(ActiveHealthChecks)
+			}
+			_, port, err := net.SplitHostPort(d.Val())
+			if err != nil {
+				return d.Errf("health_upstream is malformed '%s': %v", d.Val(), err)
+			}
+			_, err = strconv.Atoi(port)
+			if err != nil {
+				return d.Errf("bad port number '%s': %v", d.Val(), err)
+			}
+			h.HealthChecks.Active.Upstream = d.Val()
+
 		case "health_port":
 			if !d.NextArg() {
 				return d.ArgErr()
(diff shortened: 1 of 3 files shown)


modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -363,6 +384,9 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 			if h.HealthChecks.Active == nil {
 				h.HealthChecks.Active = new(ActiveHealthChecks)
 			}
+			if h.HealthChecks.Active.Upstream != "" {
+				caddy.Log().Named("config.adapter.caddyfile").Warn("the 'health_port' subdirective is ignored if 'health_upstream' is used!")
+			}
```
dylanschultzie, line 389: It might be worth adding more than just a log here to inform that the `health_port` is being ignored.

modules/caddyhttp/reverseproxy/caddyfile.go
```diff
@@ -353,6 +354,26 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+				h.HealthChecks.Active = new(ActiveHealthChecks)
+			}
+			_, port, err := net.SplitHostPort(d.Val())
+			if err != nil {
+				return d.Errf("health_upstream is malformed '%s': %v", d.Val(), err)
+			}
+			_, err = strconv.Atoi(port)
+			if err != nil {
+				return d.Errf("bad port number '%s': %v", d.Val(), err)
+			}
```
dylanschultzie, line 374: `SplitHostPort` seems to do sufficient `host` validation, but no `port` validation so I tagged that in as well.

modules/caddyhttp/reverseproxy/caddyfile.go
francislavoie: Yeah we could `return d.Errf` instead of a warning

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -75,6 +75,10 @@ type ActiveHealthChecks struct {
 	// The URI (path and query) to use for health checks
 	URI string `json:"uri,omitempty"`
 
+	// The host to use (if different from the upstream's dial
+	// address) for health checks.
+	Upstream string `json:"upstream,omitempty"`
```
francislavoie, line 80: These comments show up in the JSON docs. I think this needs to be elaborated on to explain the specific usecase intended for this

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -342,7 +351,11 @@ func (h *Handler) doActiveHealthCheck(dialInfo DialInfo, hostAddr string, networ
 	if err != nil {
 		host = hostAddr
 	}
-	if h.HealthChecks.Active.Port != 0 {
+
+	if h.HealthChecks.Active.Upstream != "" {
+		u.Host = h.HealthChecks.Active.Upstream
+	} else if h.HealthChecks.Active.Port != 0 {
+		// ignore active health check port if active upstream is provided
```
francislavoie, line 358: This comment is positioned confusingly, it's inside of the "yes we'll use the port" case but it's saying "we're ignoring the port". This should go before the condition block to explain the logic

modules/caddyhttp/reverseproxy/healthchecks.go
dylanschultzie: Thanks, updated but I may have gone overboard here.

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -75,6 +75,12 @@ type ActiveHealthChecks struct {
 	// The URI (path and query) to use for health checks
 	URI string `json:"uri,omitempty"`
 
+	// The host:ip to use (if different from the upstream's dial address)
```
francislavoie, line 78: 😅 

```suggestion
	// The host:port to use (if different from the upstream's dial address)
```

modules/caddyhttp/reverseproxy/healthchecks.go
mholt: Can we clarify this behavior in the godoc comment somewhere? If upstream is specified, port field is ignored.

dylanschultzie: @mholt thanks for the comment - can you direct me to where you would like the comment? Do you mean here? https://github.com/caddyserver/caddy/pull/6451/files#diff-23b6ca3ed370c5279841c40f9687ff0ce19cea5e6ca4f3a688bbcbcde8c33a37L78-L80

mholt: Sorry -- yes! That spot, and the new `Upstream` field above it might be good spots.

dylanschultzie: Thanks, will do so now.

mholt (approved):

Issue #6456, opened 2024-07-15
Hostnames aren't being normalized to lowercase when managing certs


low613:
Version effected: v2.8.4 h1:q3pe0wpBj1OcHFZ3n/1nl4V4bxBrYoSoab7rL9BMYNk=
If I have an uppercase domain name in the host matcher in the json config, caddy will fail to generate a certificate
for example with this config
``` json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":443"
          ],
          "routes": [
            {
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "hello",
                          "handler": "static_response"
                        }
                      ]
                    }
                  ]
                }
              ],
              "match": [
                {
                  "host": [
                    "CADDY.lowinger.cloud"
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    }
  },
  "logging": {
    "logs": {
      "default": {
        "level": "DEBUG"
      }
    }
  }
}
```

When generating a cert I get this error in the logs
```
2024/07/15 23:41:30.338 ERROR   tls.obtain      could not get certificate from issuer   {"identifier": "CADDY.lowinger.cloud", "issuer": "acme-v02.api.letsencrypt.org-directory", "error": "[CADDY.lowinger.cloud] validating order identifiers: identifiers in Order [{dns caddy.lowinger.cloud}] do not match the identifiers extracted from CSR [{dns CADDY.lowinger.cloud}] (ca=https://acme-v02.api.letsencrypt.org/directory)"}
2024/07/15 23:41:30.338 DEBUG   events  event   {"name": "cert_failed", "id": "d32ac37d-1bb9-4904-90a8-890247edfa25", "origin": "tls", "data": {"error":{},"identifier":"CADDY.lowinger.cloud","issuers":["acme-v02.api.letsencrypt.org-directory"],"renewal":false}}
[...]

PR #6459, opened 2024-07-16
autohttps: ensure domain name is lowercase

low613:
fixes: #6456


mholt:
Thanks for the PR!

Actually, I want to fix this a level deeper in either CertMagic, or maybe even ACMEz. (Maybe both, since using one doesn't necessarily mean you're using th eother.)

So, I will close this, but I will leave the issue open until I push proper fixes.

Issue #4378, opened 2021-10-11
Move Fastcgi client implementation to its own package


gsdevme:
Hi,

I've been looking for a client fastcgi implementation in Golang and the one is Caddy seems very feature rich, however although the `client.go` is using only the standard library the `fastcgi.go` and `caddyfile.go` are using alot of other things.

Would it be acceptable to move the pure fastcgi implementation which is designed around http://www.mit.edu/~yandros/doc/specs/fcgi-spec.html into its own package to allow easier consumption with other libraries and applications? I assume the implementation at this point is "completed" given there has never been any real movement I wonder if that would be acceptable to have it away from Caddy's development and just sit standalone? 

Thanks in advance.

mholt:
FWIW, @WeidiDeng has rewritten the `fastcgi` package and it's much better now. It's still a Caddy-centric package, so it's not decoupled in the sense you're requesting, but it should be much more worthwhile to extract it now.

It's not really an action item on my list, but if someone else wants to take this up, that's great too.

eanavitarte:
> FWIW, @WeidiDeng has rewritten the `fastcgi` package and it's much better now. It's still a Caddy-centric package, so it's not decoupled in the sense you're requesting, but it should be much more worthwhile to extract it now.
> 
> It's not really an action item on my list, but if someone else wants to take this up, that's great too.

Hi @mholt, I helped with this issue by opening a PR #6465. The goal was to carefully extract the client logic into a separate package, for readability and reusability. Let me know what you think!

mholt:
Oh wow, thank you! I will take a look soon :)

PR #6465, opened 2024-07-20
Pull Request: Refactor FastCGI Client into Separate Package (#4378)

eanavitarte:
## Hi Caddy team,

Looking at the list of issues about FastCGI, I noticed a "help wanted SOS" issue #4378 requesting the FastCGI client be moved into its own package (pretty logic). And as I was already familiar with this codebase (because when I contributed in the [v2.7.3](https://github.com/caddyserver/caddy/releases/tag/v2.7.3) release, I read the whole package), I've prepared a pull request to address this issue. I hope to help you!

## The Problem

Currently, the FastCGI client code is intertwined with Caddy's FastCGI implementation, making it harder to maintain (and read!) and potentially reuse the client elsewhere. This refactor aims to improve the code's organization and clarity.

## The Solution

1. **Moved client files:** The files `writer.go`, `record.go`, `reader.go`, `pool.go`, and `header.go` are now in the fastcgiclient subpackage. They were non exportable auxiliar FastCGI client files.
2. **Created fastcgi.go:** It was created a `fastcgi.go` file in the subpackage just for clean handling the initialization of the noopLogger, previously a single line of the `fastcgi.go` in the previous implementation. (This is a noopLogger, so it can be safely initialized inside the client, because it just discard the output, but we can now easily refactor the logging in a future)
3. **Exported Client:** The Client type and its fields are now exportable, so they can be imported in the Caddy's FastCGI implementation package, without changing their behavior.
4. **Updated imports:** The main FastCGI package now just imports the client, simplifying its dependencies.
5. **Maintained compatibility:** The commented-out ´client_test.go´ file remains for reference, but it is continue been disabled due to performance concerns, as it was.

## Testing

The majority of changes are fast revisable code edits and movements, easily verifiable using Git (they're mainly rename trackings and single lines). It do not change any code logic (I was careful about that).

To thoroughly test the changes, I successfully ran a WordPress site using a Caddy server with this modification. All functionality, including file uploads, API access, and GET/POST requests, worked seamlessly.

## Closing Remarks

Well, I hope this solve the 3 years issue #4378 and makes the FastCGI codebase cleaner and more maintainable.

I'd hope to be included in the contributor list for the next release. This is my second contribution! :)

Best regards,

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/fastcgiclient/client.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/fastcgiclient/client.go
@@ -21,7 +21,7 @@
 // Use of this source code is governed by a BSD-style
 // Part of source code is from Go fcgi package
 
-package fastcgi
+package fastcgiclient
 
 import (
 	"bufio"
@@ -122,19 +122,19 @@ const (
 // not synchronized because we don't care what the contents are
 var pad [maxPad]byte
 
-// client implements a FastCGI client, which is a standard for
+// Client implements a FastCGI Client, which is a standard for
 // interfacing external applications with Web servers.
-type client struct {
-	rwc net.Conn
+type Client struct {
+	Rwc net.Conn
 	// keepAlive bool // TODO: implement
-	reqID  uint16
-	stderr bool
-	logger *zap.Logger
+	ReqID  uint16
+	Stderr bool
+	Logger *zap.Logger
 }
 
 // Do made the request and returns a io.Reader that translates the data read
 // from fcgi responder out of fcgi packet before returning it.
-func (c *client) Do(p map[string]string, req io.Reader) (r io.Reader, err error) {
+func (c *Client) Do(p map[string]string, req io.Reader) (r io.Reader, err error) {
 	writer := &streamWriter{c: c}
 	writer.buf = bufPool.Get().(*bytes.Buffer)
 	writer.buf.Reset()
@@ -195,7 +195,7 @@ func (f clientCloser) Close() error {
 
 // Request returns a HTTP Response with Header and Body
 // from fcgi responder
-func (c *client) Request(p map[string]string, req io.Reader) (resp *http.Response, err error) {
+func (c *Client) Request(p map[string]string, req io.Reader) (resp *http.Response, err error) {
 	r, err := c.Do(p, req)
 	if err != nil {
 		return
@@ -231,7 +231,7 @@ func (c *client) Request(p map[string]string, req io.Reader) (resp *http.Respons
 
 	// wrap the response body in our closer
 	closer := clientCloser{
-		rwc:    c.rwc,
+		rwc:    c.Rwc,
 		r:      r.(*streamReader),
 		Reader: rb,
 		status: resp.StatusCode,
@@ -240,8 +240,8 @@ func (c *client) Request(p map[string]string, req io.Reader) (resp *http.Respons
 	if chunked(resp.TransferEncoding) {
 		closer.Reader = httputil.NewChunkedReader(rb)
 	}
-	if c.stderr {
-		closer.logger = c.logger
+	if c.Stderr {
+		closer.logger = c.Logger
 	}
 	resp.Body = closer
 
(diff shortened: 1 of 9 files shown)


modules/caddyhttp/reverseproxy/fastcgi/fastcgiclient/client.go
```diff
@@ -122,19 +122,19 @@ const (
...
-	rwc net.Conn
+type Client struct {
+	Rwc net.Conn
 	// keepAlive bool // TODO: implement
-	reqID  uint16
-	stderr bool
-	logger *zap.Logger
+	ReqID  uint16
+	Stderr bool
+	Logger *zap.Logger
```
mholt, line 132: We'll need to thoroughly document all these exported identifiers.

We should also probably rename some of these like `Rwc` to `Conn` (I think it was originally a `io.ReadWriteCloser`)... and `ReqID` to `RequestID` -- although, now I am looking again and I am not sure we should be exporting it. It always starts at 1, and it goes into an `init()` function which I need to look more into to see what it does, but it sounds like this isn't something that should be user-controllable? Maybe @WeidiDeng has some insights.

modules/caddyhttp/reverseproxy/fastcgi/fastcgiclient/client.go
```diff
@@ -21,7 +21,7 @@
 // Use of this source code is governed by a BSD-style
 // Part of source code is from Go fcgi package
 
-package fastcgi
+package fastcgiclient
```
mholt, line 24: Totally bike shedding this but I wonder if `fcgiclient` would be a little more conventional to mirror the `net/http/fcgi` package.

mholt (changes requested):
Did a quick first pass. Thank you for working on this! With a few iterations I think we can merge something along these lines into the mainline.

modules/caddyhttp/reverseproxy/fastcgi/fastcgiclient/client.go
eanavitarte: Implemented your suggestion to rename the package from `fastcgiclient` to `fcgiclient` for consistency with the net/http/fcgi package.

modules/caddyhttp/reverseproxy/fastcgi/fastcgiclient/client.go
eanavitarte: You're right about the need for better documentation and clearer naming. I've gone ahead and made the following changes:

* **Documentation:** Added detailed comments explaining the purpose and usage of each exported identifier. This should make it much easier for others to understand how these components fit into the larger system.
* **Renaming:**  Changed `rwc` to `conn` and `reqId` into `requestID` for better readability, as it does indeed represent a connection.
* **`ReqID` (RequestID):**  I agree that it might not need to be exported if it's primarily for internal use, and if it is set to 1 in any Caddy's implementation. So it is now exported only by a NewClient function, whit public methods, so user set the client with the function and can call the methods, but do not change the fields.

Please take a look at the updated code and let me know if you have any further suggestions or feedback!

modules/caddyhttp/reverseproxy/fastcgi/fcgiclient/client.go
```diff
@@ -125,13 +125,38 @@ var pad [maxPad]byte
...
+	// This boolean flag indicates whether the client should capture and handle standard error (stderr) output from the FastCGI application.
+	// When set to true, the client can redirect stderr output for logging, debugging, or error reporting.
 	stderr bool
+
+	// This is a pointer to a logger instance from the zap logging library. zap is a popular high-performance logging framework in Go.
+	// The logger field allows the client to log events, errors, and debugging information during its operation.
 	logger *zap.Logger
 }
 
+func NewClient(conn net.Conn, logger *zap.Logger, stderr bool) client {
```
mholt, line 149: Being an exported function, this should probably return an exported type. Otherwise getting documentation on how to use it will be irritating.

If we don't want to export the `client` type, maybe we can return an interface type instead, but honestly, given what `NewClient()` is doing, I feel like we could just export this as the `Client` type and I dunno if we even need `NewClient()` anymore.

mholt (changes requested):
Sorry for my long delay on this.

The changes are looking good!

I have another thought now that I see this with fresh eyes -- let me know what you think.

francislavoie:
I think we have to close this as inactive for now. The fastcgi package changed quite a lot since, so it would need to be redone from scratch anyway.

PR #6468, opened 2024-07-22
fileserver: add `sort` options

lollipopkit:
I have tested the config:
```
:8082 {
	root * .
	file_server browse {
		sort size desc
	}
}
```

Screenshot 2024-07-22 at 20 32 31

Additionally, a `.caddyfiletest` file were wrote, but I don't know how to run it.

The related doc PR can be found [here](https://github.com/caddyserver/website/pull/407).

### DIFF
--- a/caddytest/integration/caddyfile_adapt/file_server_sort.caddyfiletest
+++ b/caddytest/integration/caddyfile_adapt/file_server_sort.caddyfiletest
@@ -1,8 +1,7 @@
-:8082 {
-	root * .
-	file_server browse {
-		sort size desc
-	}
+:80
+
+file_server browse {
+	sort size desc
 }
 ----------
 {
@@ -11,21 +10,18 @@
 			"servers": {
 				"srv0": {
 					"listen": [
-						":8082"
+						":80"
 					],
 					"routes": [
 						{
 							"handle": [
 								{
+									"browse": {},
 									"handler": "file_server",
 									"hide": [
 										"./Caddyfile"
 									],
-									"file_server": {
-										"size": {},
-										"desc": {}
-									},
-									"file_server_order": [
+									"sort": [
 										"size",
 										"desc"
 									]
--- a/caddytest/integration/caddyfile_adapt/file_server_sort.caddyfiletest
+++ b/caddytest/integration/caddyfile_adapt/file_server_sort.caddyfiletest
@@ -1,0 +1,40 @@
+:8082 {
+	root * .
+	file_server browse {
+		sort size desc
+	}
+}
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":8082"
+					],
+					"routes": [
+						{
+							"handle": [
+								{
+									"handler": "file_server",
+									"hide": [
+										"./Caddyfile"
+									],
+									"file_server": {
+										"size": {},
+										"desc": {}
+									},
+									"file_server_order": [
+										"size",
+										"desc"
+									]
+								}
+							]
+						}
+					]
+				}
+			}
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


caddytest/integration/caddyfile_adapt/file_server_sort.caddyfiletest
```diff
@@ -0,0 +1,40 @@
...
+										"./Caddyfile"
+									],
+									"file_server": {
+										"size": {},
+										"desc": {}
+									},
+									"file_server_order": [
+										"size",
+										"desc"
+									]
```
mohammed90, line 31: The new field is `sort`, so these lines cannot be right.

lollipopkit: ...my fault

lollipopkit: Is this correct?
```
"sort": {
    "size": {},
    "desc": {}
},
"sort_order": [
    "size",
    "desc"
]
```

modules/caddyhttp/fileserver/browse.go
```diff
@@ -206,11 +206,35 @@ func (fsrv *FileServer) loadDirectoryContents(ctx context.Context, fileSystem fs
 // browseApplyQueryParams applies query parameters to the listing.
 // It mutates the listing and may set cookies.
 func (fsrv *FileServer) browseApplyQueryParams(w http.ResponseWriter, r *http.Request, listing *browseTemplateContext) {
+	orderParam := ""
+	sortParam := ""
```
mholt, line 210:
```suggestion
	var orderParam, sortParam string
```

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -153,6 +153,16 @@ type FileServer struct {
...
 
+	// Override the default sort. It includes the following options:
+	// sort_by: name(default), size, time
+	// order: asc(default), desc
+	// eg.:
+	//   - `sort time desc` will sort by time in descending order
+	//   - `sort size` will sort by size in ascending order
+	//   - `sort` will sort by name in ascending order, which is the default
+	// The order of the sort options is not important.
+	SortOptions []string `json:"sort,omitempty"`
```
mholt, line 164: Since this is a freeform slice of strings, we should probably validate this input at `Provision()`-time. It should be documented (and enforced) that only 1 or 2 values are allowed: the first must be a field name to order by, and the second if present should be the direction.

mholt (changes requested):
Thanks for this enhancement. I just want to make sure that the rather freeform input is strictly validated to prevent surprises.

modules/caddyhttp/fileserver/staticfiles.go
lollipopkit: Thanks for the reminder, or I wouldn't have known about Provision().

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -246,6 +246,20 @@ func (fsrv *FileServer) Provision(ctx caddy.Context) error {
 		fsrv.precompressors[ae] = p
 	}
 
+	// check sort options
+	for idx, sortOption := range fsrv.SortOptions {
+		switch idx {
+		case 0:
```
mholt, line 252: This works; but we should also verify the length of SortOptions is 1 or 2 if not empty.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -246,6 +246,20 @@ func (fsrv *FileServer) Provision(ctx caddy.Context) error {
...
+			if sortOption != sortByName && sortOption != sortByNameDirFirst && sortOption != sortBySize && sortOption != sortByTime {
+				return fmt.Errorf("invalid sort option %s", sortOption)
+			}
+		case 1:
+			if sortOption != sortOrderAsc && sortOption != sortOrderDesc {
+				return fmt.Errorf("invalid sort option %s", sortOption)
```
mholt, line 258: Could we make these two error messages different? They should say that the first value should be a field to sort by (name, dir, size, etc) and the second error should say that it should be a direction (asc or desc).

mholt (changes requested):
Thanks, getting closer!!

modules/caddyhttp/fileserver/staticfiles.go
lollipopkit: I'm sorry, I was not thinking things through...
Thanks for your patience.

mholt (approved):
Great, thank you!

I think we can give this a shot. Please help maintain this feature if any bugs are reported :blush:

lollipopkit:
Of course. Thanks again :)

PR #6470, opened 2024-07-22
ci: correct `-tags nobadger` on binary build

girlbossceo:
s/nobdger/nobadger

Seems like an oversight from https://github.com/caddyserver/caddy/pull/6031


francislavoie:
Oops! Thank you!

francislavoie (approved):

PR #6471, opened 2024-07-23
encode: flush already compressed data to the stream

WeidiDeng:


mholt (approved):
Thanks Weidi!

This does fix it! It looks like we were not flushing the encoder in `FlushError()`, which... seems like a mistake, since the ResponseWriter can't flush anything if the encoder hasn't flushed it first.

Really appreciate this, your fix is much better than what I was coming up with :smile:

PR #6479, opened 2024-07-26
chore: update golangci config

mohammed90:
I noticed complaints about deprecated config, so decided to fix them.


mholt (approved):
Ah cool, thanks!

PR #6480, opened 2024-07-26
Runtime placeholders for selected TLS and HTTP matchers

vnxme:
This PR follows https://github.com/mholt/caddy-l4/pull/224 and deals with the following mainline matchers only:
- caddytls.MatchLocalIP
- caddytls.MatchRemoteIP
- caddytls.MatchServerName
- caddyhttp.MatchClientIP
- caddyhttp.MatchRemoteIP

These changes allow to use runtime placeholders with TLS and HTTP IP matchers, e.g. `remote_ip {env.VAR}`. Such placeholders, if present, are evaluated **once at provision** due to the existing optimisations.

In addition, this PR adds runtime placeholders support for TLS SNI matcher, e.g. `sni {env.VAR}`. Contrary to the above, such placeholders are evaluated **each time at match**.

### DIFF
--- a/modules/caddytls/matchers.go
+++ b/modules/caddytls/matchers.go
@@ -107,16 +107,19 @@ func (MatchRemoteIP) CaddyModule() caddy.ModuleInfo {
 
 // Provision parses m's IP ranges, either from IP or CIDR expressions.
 func (m *MatchRemoteIP) Provision(ctx caddy.Context) error {
+	repl := caddy.NewReplacer()
 	m.logger = ctx.Logger()
 	for _, str := range m.Ranges {
-		cidrs, err := m.parseIPRange(str)
+		rs := repl.ReplaceAll(str, "")
+		cidrs, err := m.parseIPRange(rs)
 		if err != nil {
 			return err
 		}
 		m.cidrs = append(m.cidrs, cidrs...)
 	}
 	for _, str := range m.NotRanges {
-		cidrs, err := m.parseIPRange(str)
+		rs := repl.ReplaceAll(str, "")
+		cidrs, err := m.parseIPRange(rs)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


modules/caddytls/matchers.go
```diff
@@ -330,3 +339,17 @@ var (
 	_ caddyfile.Unmarshaler = (*MatchRemoteIP)(nil)
 	_ caddyfile.Unmarshaler = (*MatchServerName)(nil)
 )
+
+// PrivateRangesCIDR returns a list of private CIDR range
+// strings, which can be used as a configuration shortcut.
+// Copied from caddyhttp/ip_range.go due to an import cycle.
+func PrivateRangesCIDR() []string {
```
mohammed90, line 346: You can probably unexport this func

mohammed90 (commented):
I think the `caddy.Context` available in `Provision` contains a replacer already, and it may have some values set already. Try using the existing replacer.

vnxme:
> I think the `caddy.Context` available in `Provision` contains a replacer already, and it may have some values set already. Try using the existing replacer.

So did I think, but I checked the context, and it's basically empty at provision. I use the existing replacer at match.

modules/caddytls/matchers.go
vnxme: As stated in the comment, I copied this function from caddyhttp, where it's exported. Do you think there is a better place in the mainline code to move this function to, so that caddyhttp, caddytls and layer4 module could call it consistently?

mohammed90: You can move it under `internal`

mholt (approved):
This LGTM. Nice idea to move the PrivateRangesCIDR function into an internal package.

PR #6491, opened 2024-08-03
fastcgi: use client IP for REMOTE_ADDR

looi:
The remote port is still used for REMOTE_PORT. I believe this matches the normal/desired behavior when using PHP with nginx realip / apache remoteip (which is provided by caddy client_ip_headers/trusted_proxies). It allows better compatibility with most PHP applications that expect the client IP in REMOTE_ADDR. Tested working with php fastcgi.

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
@@ -211,18 +211,11 @@ func (t Transport) buildEnv(r *http.Request) (envVars, error) {
 
 	var env envVars
 
-	// Separate remote IP and port; more lenient than net.SplitHostPort
-	var ip, port string
-	if idx := strings.LastIndex(r.RemoteAddr, ":"); idx > -1 {
-		ip = r.RemoteAddr[:idx]
-		port = r.RemoteAddr[idx+1:]
-	} else {
-		ip = r.RemoteAddr
+	ip := caddyhttp.GetVar(r.Context(), caddyhttp.ClientIPVarKey).(string)
+	_, port, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		port = ""
 	}
-
-	// Remove [] from IPv6 addresses
-	ip = strings.Replace(ip, "[", "", 1)
-	ip = strings.Replace(ip, "]", "", 1)
 
 	// make sure file root is absolute
 	root, err := filepath.Abs(repl.ReplaceAll(t.Root, "."))


francislavoie:
No, this change is insecure and incorrect. Your PHP app _must_ consult `X-Forwarded-For`.

looi:
Setting real IP for fastcgi REMOTE_ADDR is the default behavior for both nginx (ngx_http_realip_module) and apache (mod_remoteip). I have personally tested this, and it should be clear if you look around [online](https://stackoverflow.com/a/27294535). For example, simple apache config will cause X-Forwarded-For to be passed as REMOTE_ADDR:

```
ProxyPassMatch ^/(.*\.php(/.*)?)$ fcgi://127.0.0.1:9000/scripts/$1
RemoteIPHeader X-Forwarded-For
```

If the PHP application needs to check X-Forwarded-For, it means every PHP application needs to re-implement the trusted proxy checking that the web server does. It makes more sense that the webserver should just pass the verified real IP as REMOTE_ADDR. In practice PHP applications also tend to treat REMOTE_ADDR as the original client IP. I think this is why both apache and nginx do this by default.

Also RFC 3875 [defines](https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.8) REMOTE_ADDR as "network address of the client sending the request to the server". I personally don't think setting the original client IP violates this spec.

If you still think this is insecure and incorrect (or nonstandard), can you elaborate?

francislavoie:
Real IP in other servers is opt in. You're proposing to make it always on. That's not the same thing. 

With your change, it would be _impossible_ for PHP apps to verify that the chain of IPs that the connection went through to reach it. You'd be hiding information from the PHP app for _everyone_. That's not okay. 

> Also RFC 3875 defines REMOTE_ADDR as "network address of the client sending the request to the server".

Exactly, the network address, as in the IP address on the TCP packets received by that server, aka the immediate client, not the originating client. 

> If the PHP application needs to check X-Forwarded-For, it means every PHP application needs to re-implement the trusted proxy checking that the web server does.

Correct, that's by design. 

You can always override any env passed to your PHP app with the `env` option in your Caddyfile, there's no need to change Caddy itself to do this. But we will never recommend this because it's the wrong way to do it. Modern PHP frameworks make it extremely easy to configure trusted proxies, use that functionality.

looi:
> Real IP in other servers is opt in. You're proposing to make it always on. That's not the same thing.

This behavior in apache/nginx is the default behavior, not opt-in. (If you consider that specifying RemoteIPHeader in apache is "opting in", then specifying client_ip_headers would be "opting in" for caddy. That is exactly what this change does. It does not affect caddy's behavior in other cases.)

> Exactly, the network address, as in the IP address on the TCP packets received by that server, aka the immediate client, not the originating client.

It seems like apache/nginx don't agree with your interpretation. I think it's also important to consider the practical usage of things, not only the standards.

> You can always override any env passed to your PHP app with the env option in your Caddyfile, there's no need to change Caddy itself to do this.

Are you sure? It looks like REMOTE_ADDR and other fastcgi params are currently hardcoded in caddy.

> Modern PHP frameworks make it extremely easy to configure trusted proxies, use that functionality.

You are often limited by the functionality provided by software packages like WordPress. With commercial protected software, you may not be able to hack the source code even if you wanted to. Here's a more concrete example. Let's say someone wants to set up a WordPress blog behind a reverse proxy using php-fastcgi. Apache could be configured as:

```
ProxyPassMatch ^/(.*\.php(/.*)?)$ fcgi://127.0.0.1:9000/scripts/$1
RemoteIPHeader X-Forwarded-For
RemoteIPTrustedProxy <trusted proxy range>
```

Nginx could be configured as:

```
location ~ \.php$ {
  fastcgi_pass 127.0.0.1:9000;
  include fastcgi_params;
}
set_real_ip_from  <trusted proxy range>;
real_ip_header X-Forwarded-For;
```

Caddy could be configured as:

```
php_fastcgi 127.0.0.1:9000
client_ip_headers X-Forwarded-For
trusted_proxies static <trusted proxy range>
```

With apache and nginx, WordPress will receive the verified real IP and work as expected. On the other hand, with caddy, WordPress will not work properly. AFAIK WordPress does not provide any built-in options other than blindly trusting X-Forwarded-For, which is totally insecure. I agree that this is possibly "the wrong way to do it". But I think we need to also look at the practical aspect of the vast amount of already existing and widely-used PHP software, and allowing users to set up the software correctly in an easy and non-error-prone manner.

image

> With your change, it would be impossible for PHP apps to verify that the chain of IPs that the connection went through to reach it. You'd be hiding information from the PHP app for everyone. That's not okay.

I agree that it would be impossible for PHP apps to verify the chain of IPs. On the other hand, I think this is behavior is pretty standard and matches the behavior of apache and nginx.

Would you be open to adding this behavior as a config option? You can decide if you want the setting to be default on or off. I think the apache/nginx behavior is what users will want in >99% of cases. In case a user has a special scenario where they need to verify the chain of IPs like you mention, they could change this setting. The user could also disable client_ip_headers in caddy and rely entirely on PHP to perform the trusted proxy verification.

francislavoie:
> Are you sure? It looks like REMOTE_ADDR and other fastcgi params are currently hardcoded in caddy.

Yes. You can use `env` to override anything Caddy passes to PHP.

> You are often limited by the functionality provided by software packages like WordPress. 

No you're not. It's trivial to handle in WordPress. You just do it in wp_config.php or whatever. Many Caddy users do this successfully.

looi:
> No you're not. It's trivial to handle in WordPress. You just do it in wp_config.php or whatever. Many Caddy users do this successfully.

Here's an [example](https://gist.github.com/ryanjbonnell/9509696) of how wp-config.php can be used to make WordPress compatible with Caddy. It overrides caddy-provided REMOTE_ADDR with the expected value. It's not necessary with apache/nginx, and only necessary because of Caddy's non-standard behavior with REMOTE_ADDR. This hack could be used with other PHP applications.

Overall, i think caddy's behavior is more insecure than apache/nginx because it puts apps like WordPress into a default-insecure configuration that may not be fixable without such PHP hacks. It may be too late to change caddy, but I think it's unfortunate. If caddy wants to provide the immediate client IP, it may be better do so in a separate env/header instead of putting a non-standard value into REMOTE_ADDR.

francislavoie:
> and only necessary because of Caddy's non-standard behavior with REMOTE_ADDR

We're standards compliant. Reverse proxies are outside the scope of the CGI RFC, therefore we cannot assume that their wording means to include that.

PR #6502, opened 2024-08-10
fileserver: move `sort` to `browse`

lollipopkit:
As discussed in https://github.com/caddyserver/website/pull/407#discussion_r1711766804


mholt (approved):
Thank you -- if there are bugs reported with this it would be great if you could help maintain it :+1:

PR #6504, opened 2024-08-10
reverseproxy: allow user to define source address

### LINKED ISSUE #6503: [feature request]reverse_proxy can setting source address for connecting upstreams
I have two IP addresses on the same vlan:

igb0: 10.244.233.10/24 (gw 10.244.233.254)
igb0: x.x.x.x/24 (gw x.x.x.254, secondry address on the same vlan)

System configed with policy-based routing for different source addresses.

srv01: y.y.y.y (remote subnet, acl restrictions, and no permission to operate the server)

So I want to define the source address connect to y.y.y.y with x.x.x.x，but supply services listening on 10.244.233.10.
When I used nginx, I could use proxy_bind to set the source IP address for accessing the upstream. After switching to caddy, there is no corresponding solution.

mohammed90:
TODO:
- [x] Caddyfile support

Closes #6503

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -132,6 +132,10 @@ type HTTPTransport struct {
 	// to change or removal while experimental.
 	Versions []string `json:"versions,omitempty"`
 
+	// Specify the address to bind to when connecting to an upstream. In other words,
+	// it is the address the upstream sees as the remote address.
+	LocalAddress string `json:"local_address,omitempty"`
+
 	// The pre-configured underlying HTTP transport.
 	Transport *http.Transport `json:"-"`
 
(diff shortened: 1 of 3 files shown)


mohammed90:
This failure appears to be legit because it fails even locally for me

https://github.com/caddyserver/caddy/actions/runs/10331386719/job/28601743464?pr=6504#step:3:4268

francislavoie:
I think users might want to use an interface name instead of an IP as well, we'd likely want to support that too?

mohammed90:
> I think users might want to use an interface name instead of an IP as well, we'd likely want to support that too? I've had this [owasp-amass/amass@f5c3960](https://github.com/owasp-amass/amass/commit/f5c3960792bb9589ac40db64ff75267e1a95384b) bookmarked for a long time as a reference for how to do that.

I'd solve that differently. We can use the new `RegisterNetwork` function to register a network called `interface` processing a network address looking like this, e.g.: `interface/eth0:port`.  Then the implementation in this PR inherits the utility by default. What do you think?

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -132,6 +132,10 @@ type HTTPTransport struct {
 	// to change or removal while experimental.
 	Versions []string `json:"versions,omitempty"`
 
+	// Specify the address to bind to when connecting to an upstream. In other words,
+	// it is the address the upstream sees as the remote address.
```
mholt, line 136: Is that actually true, or is this just to help the local dialer choose the right interface to dial out of? Like could I set this to `1.1.1.1` or something ridiculous? :thinking:

mohammed90: No, it doesn't accept any random address. An interface must exist with that address. The doc on `LocalAddr` field in [`net#Dialer`](https://pkg.go.dev/net#Dialer) says:

```
	// LocalAddr is the local address to use when dialing an
	// address. The address must be of a compatible type for the
	// network being dialed.
	// If nil, a local address is automatically chosen.
```

mholt: Oh ok, perfect.

LGTM so far!

mholt (approved):

PR #6506, opened 2024-08-10
core: add `iface` as network type

mohammed90:
This PR allows using the interface name as the network address host when using the `iface+?{udp,tcp}` network. In other words, user will be able to say:

```
example.com {
    bind iface/eth0
    respond "What's up, Doc?"
}
```

Because an interface may have multiple IPv6 addresses, and I can't come up with a clean up to communicate this using the host part without it being hairy, I limited this to picking up only the IPv4 address of the defined interface.

I tried at first using the format, e.g. `iface/en0+tcp`, for the protocol selection, but the HTTP3 listener is created by looking up a custom registered network, then using that network to lookup its listener constructor. This makes registering a custom network for UDP the only possible approach to hook our constructor, i.e. doing `iface+udp` et al.

I'd classify this highly experimental. It's extremely limited for only supporting IPv4 addresses. I don't foresee any possibility of adding IPv6 support without it being convolute and painful. Do we really want this?

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -1,0 +1,51 @@
+package caddy
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+)
+
+func init() {
+	RegisterNetwork("iface", getInterfaceListener)
+	RegisterNetwork("iface+tcp", getInterfaceListener)
+	RegisterNetwork("iface+udp", getInterfaceListener)
+}
+
+func getInterfaceListener(ctx context.Context, network, addr string, config net.ListenConfig) (any, error) {
+	// assuming addr = "interface+family:port"
+	// if family is missing, then assume tcp
+	family := "tcp"
+	parts := strings.Split(network, "+")
+	if len(parts) == 2 {
+		family = parts[1]
+	}
+	host, port, _ := net.SplitHostPort(addr)
+	iface, err := net.InterfaceByName(host)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


mholt:
This is interesting, but it's a weird violation of abstractions. The bind directive typically describes the _socket_ not the interface. Hence you can bind to tcp/udp/unix and an address, but specifying an interface is like a whole separate thing.

I'm not sure yet what the best solution is... would be nice to avoid funky mini-syntax here.

mohammed90:
I put this together as PoC in response to https://github.com/caddyserver/caddy/pull/6504#issuecomment-2281492551. I don't know if it's the best approach or not. It's an experiment.

MayCXC:
8 years of discussion about the ipv6 multicast here https://github.com/golang/go/issues/9334
we should rally to get this added to net.go 💪

MayCXC:
also with the new optional block for bind, i think we could do

```
bind udp/ {
  iface eth0
}
```

to keep the network intact.

MayCXC:
> @MayCXC I think I like that more. 🤔 Still a weird concept overall to me 😄

I think of it like the network host for the specified gateway(s), i.e "0.0.0.0 for my wireguard VPN":

```
bind tcp/ {
  iface wg0
}
```

or similarly:

```
bind tcp/ {
  iface lo
}
```

is equivalent to `bind tcp/127.0.0.1`

mohammed90:
Superceded by #7256

Issue #6296, opened 2024-05-03
Support passing FDs (socket activation)


flokli:
I'd like to use caddy in a socket-activated environments, using FDs passed down from the service manager, rather than binding on addresses on its own.

Combined with signalling readyness (which caddy already does), this will give zero-downtime (re)deployments on Linux systems using systemd (if `.socket` files are used), by simply restarting the process - the socket is held open by systemd, and new connections are passed in once caddy is ready to accept new requests. In these cases, there wouldn't be a need for complicated reload logic anymore.

`github.com/coreos/go-systemd/activation` provides the necessary methods to check whether FDs are passed, including identifying them by their socket name. https://vincent.bernat.ch/en/blog/2018-systemd-golang-socket-activation gives a nice introduction into the feature itself.

In case no explicit listen addresses are specified, caddy could default to do that rather than binding on its own, if it detects it's running in such an environment.
Additionally, Caddyfile could be extended to allow specifying these passed fds as network addresses (something like `sd-listen:$name` or `sd-listen:$idx` maybe). This can become useful when you want to expose different things on different sockets.

francislavoie:
Are you looking for the `bind` directive? https://caddyserver.com/docs/caddyfile/directives/bind

mholt:
It sounds like what is being asked for is graceful upgrades/restarts.

Caddy 1 had this feature, and I quite liked how it worked: pass the socket directly to the next process. It worked on all Unix systems without relying on a separate system service, and it was smart enough to understand Caddy configuration: if the new config didn't use a socket, it wouldn't be kept; rather than blindly moving all the sockets over.

I'd probably rather bring the implementation from Caddy 1 into Caddy 2.

flokli:
> It sounds like what is being asked for is graceful upgrades/restarts.

No, getting this for free is only one side-effect of supporting socket-activation.

socket-activation will also cause caddy to get started lazily whenever the first connection to the (externally configured) socket address happens, which simplifies declaring service dependencies too.

The [article linked from my link](http://0pointer.de/blog/projects/socket-activation.html) elaborates a bit more on this.

> Caddy 1 had this feature, and I quite liked how it worked: pass the socket directly to the next process.

This still requires caddy to do manual coordination with its new process and pass it around explicitly. The point of simply taking the FDs passed by the service manager is that caddy does not have to be aware of whether it's the first process being started on the system, or you start a new version with another config. caddy simply gets an FD, where new connections will appear on.

[...]

PR #6507, opened 2024-08-11

MayCXC:
This PR enables socket activation by adding a global server option that can be used to open the fd of an existing socket, instead of binding a new socket, for any network address. Example `Caddyfile`:

```
{
	servers :80 {
		socket {env.CADDY_HTTP_FD}
		protocols h1
	}
	servers :443 {
		socket {env.CADDY_HTTPS_FD}
		protocols h1 h2
	}
	servers udp/:443 {
		socket {env.CADDY_HTTP3_FD}
		protocols h3
	}
	auto_https disable_redirects
	admin off
}

http://localhost {
	bind tcp/
	log
	respond "Hello, HTTP!"
}

https://localhost {
	bind tcp/
	bind udp/
	log
	respond "Hello, HTTPS!"
}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Run Tests
         run: |
+          set +e
           mkdir -p ~/.ssh && echo -e "${SSH_KEY//_/\\n}" > ~/.ssh/id_ecdsa && chmod og-rwx ~/.ssh/id_ecdsa
 
           # short sha is enough?


MayCXC:
link to issue: https://github.com/caddyserver/caddy/issues/6296

MayCXC:
`caddy adapt` produces the expected config, and allows env placeholders in server/socket

MayCXC:
It works for a local HTTP connection with this Caddyfile:

```
{
	servers :80 {
		socket {env.CADDY_HTTP_FD}
		protocols h1
	}
	servers :443 {
		socket {env.CADDY_HTTPS_FD}
		protocols h1 h2
	}
#	servers udp/:443 {
#		socket {env.CADDY_HTTP3_FD}
#		protocols h3
#	}
	auto_https off
	admin off
}

http:// {
	bind tcp/
	log
	respond "Hello, HTTP!"
}

https:// {
	bind tcp/
#	bind udp/
	log
	respond "Hello, HTTPS!"
}
```

this `caddy-socket-activate` script:
```
#!/usr/bin/env -S sh -e
CADDY_HTTP_FD=3 CADDY_HTTPS_FD=4 exec "${@}"
```

and this test command: `systemd-socket-activate -l 0.0.0.0:80 -l 0.0.0.0:443 ./caddy-socket-activate ./caddy run -c ./Caddyfile`

I have not seen https work yet, or http3 with `systemd-socket-activate -l 0.0.0.0:443 -d` and the udp bind directive enabled. I think `systemd-socket-activate` may be the issue, because I can only get it to activate with UDP when one only socket is listening. I also still have to handle config reloads.

MayCXC:
Ok, I think it works with h3 now. I split `Server.Protocols` into a parallel list of lists, for each listener for each server. h3 is still enabled for TCP sockets by binding to the UDP socket with the same address and port, but it can now be disabled per address rather than per server. It uses the same socket activation fd global options as h1 and h2, so it can use systemd socket activation with `ListenDatagram`. here is the config I tested with:

`Caddyfile`:

```
{
	servers :80 {
		socket {env.CADDY_HTTP_FD}
		protocols h1
	}
	servers :443 {
		socket {env.CADDY_HTTPS_FD}
		protocols h1 h2
	}
	servers udp/:443 {
		socket {env.CADDY_HTTP3_FD}
		protocols h3
	}
	auto_https disable_redirects
	admin off
}

http://localhost {
	bind tcp/
	log
	respond "Hello, HTTP!"
}

https://localhost {
	bind tcp/
	bind udp/
	log
	respond "Hello, HTTPS!"
}
```

`caddy.service`:

```
[Unit]
Description=Caddy
After=network.target network-online.target
Requires=network-online.target

[Service]
Type=notify
User=namefull
Group=namefull
Environment=CADDY_HTTP_FD=3 CADDY_HTTPS_FD=4 CADDY_HTTP3_FD=5
ExecStart=/path/to/caddy run --environ --config /path/to/Caddyfile
ExecReload=/path/to/caddy reload --config /path/to/Caddyfile --force
TimeoutStopSec=5s
LimitNOFILE=1048576
PrivateTmp=true
ProtectSystem=full
AmbientCapabilities=CAP_NET_ADMIN CAP_NET_BIND_SERVICE

[Install]
WantedBy=multi-user.target
```

`caddy.socket`:

```
[Socket]
ListenStream=80
ListenStream=443
ListenDatagram=443

[Install]
WantedBy = sockets.target
```

I need to double check that the h3 socket is actually working, because my browser doesn't like it. But http/https on h1/h2 is definitely working for any combination of binds, tls, inet and unix domain sockets I have tried.

modules/caddyhttp/app.go
```diff
@@ -374,6 +385,17 @@ func (app *App) Start() error {
...
 
 	for srvName, srv := range app.Servers {
+		// TODO why is this needed for caddy respond now?
+		if srv.Socket == nil {
+			srv.Socket = make([]*string, len(srv.Listen))
+		}
+		if srv.Protocols == nil {
+			srv.Protocols = make([][]string, len(srv.Listen))
+			for pi, _ := range srv.Protocols {
+				srv.Protocols[pi] = []string{"h1", "h2", "h3"}
```
MayCXC, line 395: I'm unsure why `caddy respond` needs this here to get a valid default config now, instead of line 209. `caddy run` works without it.

caddy.go
```diff
@@ -875,6 +875,16 @@ func ParseDuration(s string) (time.Duration, error) {
...
 
+// sliceContains returns true if needle is in haystack.
+func SliceContains(haystack []string, needle string) bool {
+	for _, s := range haystack {
+		if s == needle {
+			return true
+		}
+	}
+	return false
+}
```
mohammed90, line 886: There's no need this function be exposed outside Caddy. Let's move it to `internal`. Better now than later.

modules/caddyhttp/server.go
```diff
@@ -215,8 +215,8 @@ type Server struct {
 	// We recommend for most users to simply let Caddy use the
 	// default settings.
 	//
-	// Default: `[h1 h2 h3]`
-	Protocols []string `json:"protocols,omitempty"`
+	// Default: `[[h1 h2 h3]...]`
+	Protocols [][]string `json:"protocols,omitempty"`
```
mohammed90, line 219: This is a breaking change. It breaks existing configuration.

caddy.go
mholt: Or we could use this new generic function: https://pkg.go.dev/slices#Contains

MayCXC:
ok, this puts the config provisioning/defaults for Server.Socket and Server.Protocols where I think they need to be. I can make Server.Protocols a []string again to lessen the breaking config change, but the individual protocol global server options will still have to be stored separately somewhere, say Server.ListenProtocols. This is for fd passing, I need a way to tell apart an h1/h2 TCP socket with an implicit h3 UDP socket bound to the same host:port, vs. two separate h1/h2 TCP and h3 UDP sockets each with their own fds. This way h3 can still be enabled for a TCP socket to indicate the first case, but it can be disabled if that TCP socket is passed an fd so it isn't bound twice.

MayCXC:
is it ok to change the shape of Server.Protocols? it just isn't very straightforward to enable h3 for a whole server, that can bind to addresses with both tcp and udp networks. each server needs to know which http should serve which socket with which fd, not the union of protocols enabled for all its listen addresses. this other change this causes is that an h3 UDS socket will have to use the correct `unixgram` network in its listen address, rather than `unix`. but I think this is expected behavior.

the best way to do this is probably to add `Server.Bind` as a `[]struct{Listen: string, Socket: *string, Protocols: []string}`, and just remove/ignore `Server.Listen` and `Server.Protocols`. or they could be provisioned together and converted to a `Server.Bind` array, where each address has the same protocols, and display a deprecation warning.

modules/caddyhttp/server.go
mholt: This can't be changed in this way; but do we need to at all? Protocols are more of a server configuration than having to do with the individual listeners.

MayCXC:
another way to approach config would be to add protocol flags to the bind directive, `bind h1+h2+tcp/ h3+udp/`, with `h1+h2+h3+` as a default determined by `Server.protocols`. Then the shape of `Server.Protocols` can be left alone, and its behavior stays the same. This is more precise as well, because one site block can potentially serve a listen address with `h1` while another serves it with `h1+h2`.

MayCXC:
```
{
	socket :80 {env.CADDY_HTTP_FD}
	socket :443 {env.CADDY_HTTPS_FD}
	socket udp/:443 {env.CADDY_HTTP3_FD}

	servers :80 {
		# protocols not needed for socket activation here
	}
	servers :443 {
		# ...
	}
	servers udp/:443 {
		# ...
	}
	auto_https disable_redirects
	admin off
}

http://localhost {
	bind tcp/
	log
	respond "Hello, HTTP!"
}

https://localhost {
	bind h1+h2+tcp/ # overrides Server.protocols with ["h1","h2"] for listener config
	bind h3+udp/
	log
	respond "Hello, HTTPS!"
}
```

MayCXC:
I could also add a new directive `serve` instead of changing `bind`, to allow `serve tcp/ h1 h2` instead of `bind h1+h2+tcp/`.

mholt:
Did you mean to close this? Sorry I've been absent. Very busy with a lot to catch up on.

MayCXC:
> Did you mean to close this? Sorry I've been absent. Very busy with a lot to catch up on.

no problem. I'm going to reopen or make a new PR with a separate listen_protocols config key, an optional block after `listen` to specify which protocols to serve, and a separate socket global option. This will have the same functionality as this PR, but avoid making breaking config changes. It might be ready today, the branch is here: https://github.com/caddyserver/caddy/compare/master...MayCXC:caddy:master

mholt:
Ah neat. Sounds good, will be happy to review when you're ready!

MayCXC:
closed in favor of https://github.com/caddyserver/caddy/pull/6543

PR #6510, opened 2024-08-11
Fix a regression in #6480: the context may have no replacer

vnxme:
As found by @mohammed90, TestDefaultSNI fails after #6480. I've looked into the issue and found the context has no replacer in this test for some reason. This PR makes the test pass.


mohammed90 (approved):
That does it 👍🏼 Thank you!

Issue #6508, opened 2024-08-11
[windows] cannot run executable found relative to current directory


TeKett:
Im trying to start caddy, and not run it. Since i still want access to the cmd window and not have it locked so that i can make more commands.

It was working fine when i installed it a couple weeks back. Fast forward a couple weeks, now when i try to "start" it i get `Error: starting caddy process: exec: "caddy": cannot run executable found relative to current directory`. I can still "run" caddy without issue, but then i cant do more commands.

francislavoie:
You should run Caddy as a Windows service to keep it running. https://caddyserver.com/docs/running#windows-service

mholt:
I think it should also be in your PATH.

Closing, but feel free to continue discussion if this is a bug in Caddy.

WeidiDeng:
There is `PATH` on windows, but unlike unix, it's quite cumbersome to set up. Python and Java installer can be configured to configure this variable at install time on windows for a reason. `PATH` is vital in this case, because those executables depend on lots of other executables that will be searched in `PATH` and some of those related, util executables and put in the same directory.

For simple use case and the binary itself not needing external executables such as caddy I think it's reasonable to put it anywhere user likes and expect it to run.

WeidiDeng:
@TeKett Can you try `xcaddy build relative-path-fix` to see if it's fixed?

TeKett:
> xcaddy build relative-path-fix

xcaddy is an unknown command, tried with "caddy" instead and it does not know what "build" is. I tried to add caddy to the path, did not work either. I'm still learning and i don't actually remember what i did 2 weeks ago. All i want is to have the cmd window open and accessible, since i need to keep restarting the server and other stuff whenever i make config changes. And for whatever reason caddy is not automatically reading the caddyfile (its located right next to the caddy.exe) so i have to include it as an argument, which i cant do if i run it as a service.

francislavoie:
@TeKett https://caddyserver.com/docs/build#xcaddy

PR #6512, opened 2024-08-12
ignore exec.ErrDot when starting caddy in background

WeidiDeng:
Fix [6508](https://github.com/caddyserver/caddy/issues/6508).

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -74,6 +74,10 @@ func cmdStart(fl Flags) (int, error) {
 	// sure by giving it some random bytes and having it echo
 	// them back to us)
 	cmd := exec.Command(os.Args[0], "run", "--pingback", ln.Addr().String())
+	// we should be able to run caddy in relative paths
+	if errors.Is(cmd.Err, exec.ErrDot) {
+		cmd.Err = nil
+	}
 	if configFlag != "" {
 		cmd.Args = append(cmd.Args, "--config", configFlag)
 	}


cmd/commandfuncs.go
```diff
@@ -74,6 +74,10 @@ func cmdStart(fl Flags) (int, error) {
 	// sure by giving it some random bytes and having it echo
 	// them back to us)
 	cmd := exec.Command(os.Args[0], "run", "--pingback", ln.Addr().String())
```
mohammed90, line 76: I don't have access to Windows machine with the same issue, but I wondered if using `os.Executable` would fix it. The error message complains about relative path, but `os.Executable` returns the full path. What do you think?

WeidiDeng: `os.Executable` is fine. It's used in other places where caddy binary path is needed, i.e. experimental `upgrade` command.

mholt:
But isn't this a security risk? The reason this error is returned is because this may be unexpected. (Sigh, I don't understand why Windows does this. Of course putting binaries in an explicit PATH is a security benefit.)

WeidiDeng:
`os.Arg[0]` is the program name, which may or may not be a relative path. In case it's a relative path, it still points to the path where the executable is. The security risk is only when the executable it points to may not be what we wanted. That has to do with executable resolution in `PWD`.

The only way this path resolve to an executable that' not wanted is very complex and depends too much on timing. Caddy has to start and the original executable get unlinked and a new executable named caddy has to be created before the original caddy gets to the place where it spawns a new caddy process. This is very unlikely to happen in the real world.

I think `caddy start` is a convenient way to run a caddy. Denying user usage simply because it's not put in the `PATH` directory is against simple to use philosophy.

mholt:
Ah, that's a good point, it's always the 0th CLI arg, which I guess makes it more trustworthy.

mholt (approved):

cmd/commandfuncs.go
mohammed90: So is this error ignorable without impacting the `Start` later?

WeidiDeng: It's in the [official examples](https://pkg.go.dev/os/exec):

```
cmd := exec.Command("prog")
if errors.Is(cmd.Err, exec.ErrDot) {
	cmd.Err = nil
}
if err := cmd.Run(); err != nil {
	log.Fatal(err)
}
```

PR #6513, opened 2024-08-12
tls: replace only known placeholders in sni matcher

mohammed90:
The fix from #6510 revealed another bug in the application of replacer.

### DIFF
--- a/modules/caddytls/matchers.go
+++ b/modules/caddytls/matchers.go
@@ -60,7 +60,7 @@ func (m MatchServerName) Match(hello *tls.ClientHelloInfo) bool {
 	}
 
 	for _, name := range m {
-		rs := repl.ReplaceAll(name, "")
+		rs := repl.ReplaceKnown(name, "")
 		if certmagic.MatchWildcard(hello.ServerName, rs) {
 			return true
 		}


vnxme:
ReplaceAll() is used not only in the tls sni matcher, but in other IP matchers in Caddy, and in many places of caddy-l4 as well. Thus I have 2 questions:
1) Why do you think it's better to keep unknown placeholders rather replacing them with an empty string? Curly braces are not valid in domain names, so the tls sni matcher won't match at all, if there are invalid symbols in the matching string.
2) Why is it only the tls sni matcher that will use ReplaceKnown() instead of ReplaceAll()? I think we should consistently use either of them, but not both, unless there are strong arguments for the latter.

mohammed90:
To be honest, I don't understand 100% why is it only tripping these test cases (3 cases: TestLoadUnorderedJSON, TestACMEServerWithDefaults, and TestACMEServerAllowPolicy) even though none of them have placeholders in them. I'm trying to investigate more. All I know is that `git bisect` points to the same commit, so I implemented this change on a hunch, and it worked.

vnxme:
I suppose the tls sni matcher has a context with no replacer when it is inside a connection policy. Now after #6510 the matcher deals fine with it. In my IDE `go test caddy` passes 447 out of 450 tests (3 of them are ignored). That's why I don't fully understand the bug you're referring to which is supposed to be solved by this PR.

mohammed90:
> In my IDE `go test caddy` passes 447 out of 450 tests (3 of them are ignored).

I don't know why your IDE ignores them. Double check its config. Our integration test run just does `CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./...`.

mholt:
Hmm, this change doesn't seem right -- I also think that SNI should never have `{ }` in it, so replacing all (with an empty string if needed) is the right thing to do.

The tests are passing for me locally on Linux (with the above command), including the 3 mentioned.

mohammed90:
Okay, let me trace with delv the debugger to see what's going on

mohammed90:
You both are right. Nothing is wrong with it. I've run the tests on the CI machine itself, and they all pass. It's just weird. Sorry for the noise.

I need to fix this machine 😩

PR #6514, opened 2024-08-12
caddyhttp: Export PrivateRangesCIDR() for plugins after #6480

vnxme:
As discussed in https://github.com/mholt/caddy-l4/pull/224, let's move PrivateRangesCIDR() back to caddyhttp and caddytls after #6480.


francislavoie:
What if we keep it in internal for internal use but just keep a pass-through in caddyhttp for plugin compat?

francislavoie (approved):

PR #6519, opened 2024-08-15
ci: don't exit early on error in remote CI machine

mohammed90:
GitHub Actions shell scripts are loaded with `set -e`, which tells the shell to exit at any error. The impact on our flow is missing the deletion step of the code for every failed CI run on the remote machine. This PR fixes it so No Files Left Behind.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Run Tests
         run: |
+          set +e
           mkdir -p ~/.ssh && echo -e "${SSH_KEY//_/\\n}" > ~/.ssh/id_ecdsa && chmod og-rwx ~/.ssh/id_ecdsa
 
           # short sha is enough?


mholt:
Ah, so this'll help prevent disk space depletion? Or if these are ephemeral instances do we need to do the cleanup?

mohammed90:
> Ah, so this'll help prevent disk space depletion?

Yep, but it's just a copy of the repo at the subject commit, not that much.

> Or if these are ephemeral instances do we need to do the cleanup?

It isn't an ephemeral machine. It's the IBM machine.

dunglas:
Maybe should this be done only on IBM machines to prevent running useless steps on ephemeral containers in other cases?

mohammed90:
> Maybe should this be done only on IBM machines to prevent running useless steps on ephemeral containers in other cases?

It is only on the IBM machine. It's in the s390x-test job. That job only runs on a specific machine.

dunglas (approved):

PR #6532, opened 2024-08-21
reverseproxy: Change logs for write errors to warn level

jum:
Most of the errors that can be seen here are write errors due to clients aborting the request from their side. Often seen ones include:

	* writing: ... write: broken pipe
	* writing: ... connection timed out
	* writing: http2: stream closed
	* writing: timeout...
	* writing: h3 error...

Most of these errors are beyond of the control of caddy on the client side, probably nothing can be done on the server side. It still warrants researching when these errors occur very often, so a change in level from error to warn is better here to not polute the logs with errors in the normal case.


francislavoie:
Yeah that makes sense I think. Most are not actionable errors.

mholt (approved):
I guess that's a good point. LGTM. Thanks!

PR #6533, opened 2024-08-21
cmd: Use a factory to create the caddy root command

elee1766:
since only one instance of the cobra root command exists in production branch, this means that only one copy of the caddy app can be running within a process at a time. along with that, a run of the command leaves command state dirty, which could have unintended consequences.

this is a major problem in being able to properly test caddy through the command wrapper. 

in general - having such global state is a code smell, and we should strive to remove them from the codebase when possible

### DIFF
--- a/cmd/commandfactory.go
+++ b/cmd/commandfactory.go
@@ -4,22 +4,22 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type RootCommandFactory struct {
+type rootCommandFactory struct {
 	constructor func() *cobra.Command
 	options     []func(*cobra.Command)
 }
 
-func NewRootCommandFactory(fn func() *cobra.Command) *RootCommandFactory {
-	return &RootCommandFactory{
+func newRootCommandFactory(fn func() *cobra.Command) *rootCommandFactory {
+	return &rootCommandFactory{
 		constructor: fn,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


cmd/commandfactory.go
```diff
@@ -0,0 +1,28 @@
+package caddycmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type RootCommandFactory struct {
```
francislavoie, line 7: If all this only gets used inside `caddycmd` we could just not export it? Is that fine? Or might other plugins/runtimes want to call this (thinking FrankenPHP, Caddy-Docker-Proxy)

Could use some godoc comments in here to explain the purpose.

elee1766: hmmmm. i don't think people should be importing it. if they need the snippet it they should copy the lines. i'll make it private.

cmd/commands.go
```diff
@@ -438,43 +438,45 @@ EXPERIMENTAL: May be changed or removed.
 		},
 	})
 
-	RegisterCommand(Command{
-		Name:  "manpage",
-		Usage: "--directory <path>",
-		Short: "Generates the manual pages for Caddy commands",
-		Long: `
+	defaultFactory.Use(func(rootCmd *cobra.Command) {
+
```
francislavoie, line 442: Linter is being picky

```suggestion
```

francislavoie (approved):

PR #6560, opened 2024-09-04
perf: use zap's Check() to prevent useless allocs

dunglas:
Zap provides [an optional API](https://pkg.go.dev/go.uber.org/zap#example-Logger.Check) to prevent allocating fields that will never be used because the configured log level is higher than the log emitted.

We had good results using this trick [in Mercure](https://github.com/dunglas/mercure/pull/554) and more recently [in FrankenPHP](https://github.com/dunglas/frankenphp/pull/1005). This is also the trick that has been used in https://github.com/caddyserver/caddy/pull/6541.

This is especially important for `DEBUG` and `INFO` level logs (which are usually disabled in production), as well as for benchmarks (where logs are usually entirely disabled).

As you can see in the added benchmark, this significantly improves performance:

Before (no logs):

```
goos: darwin
goarch: arm64
pkg: github.com/caddyserver/caddy/v2/modules/caddyhttp
cpu: Apple M1 Pro
BenchmarkServer_LogRequest_NopLogger
BenchmarkServer_LogRequest_NopLogger-10    	 2797832	       433.9 ns/op	     408 B/op	       3 allocs/op
PASS
ok  	github.com/caddyserver/caddy/v2/modules/caddyhttp	2.150s
```

After (no logs):

```
goos: darwin
goarch: arm64
pkg: github.com/caddyserver/caddy/v2/modules/caddyhttp
cpu: Apple M1 Pro
BenchmarkServer_LogRequest_NopLogger
BenchmarkServer_LogRequest_NopLogger-10    	55410920	        21.63 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/caddyserver/caddy/v2/modules/caddyhttp	2.693s
```

Our FrankenPHP benchmarks show that the server can handle ~1% more requests per second by applying this patch (a Hello World page). In some cases, the performance gain can be more important (streaming for instance, which logs many things at the `DEBUG` level).

The optimized syntax is a bit more verbose, but not that much IMHO.

### DIFF
--- a/modules/caddyhttp/server.go
+++ b/modules/caddyhttp/server.go
@@ -275,7 +275,9 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		if r.ProtoMajor < 3 {
 			err := s.h3server.SetQUICHeaders(w.Header())
 			if err != nil {
-				s.logger.Error("setting HTTP/3 Alt-Svc header", zap.Error(err))
+				if c := s.logger.Check(zapcore.ErrorLevel, "setting HTTP/3 Alt-Svc header"); c != nil {
+					c.Write(zap.Error(err))
+				}
 			}
 		}
 	}
@@ -283,9 +285,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// reject very long methods; probably a mistake or an attack
 	if len(r.Method) > 32 {
 		if s.shouldLogRequest(r) {
-			s.accessLogger.Debug("rejecting request with long method",
-				zap.String("method_trunc", r.Method[:32]),
-				zap.String("remote_addr", r.RemoteAddr))
+			if c := s.accessLogger.Check(zapcore.DebugLevel, "rejecting request with long method"); c != nil {
+				c.Write(
+					zap.String("method_trunc", r.Method[:32]),
+					zap.String("remote_addr", r.RemoteAddr),
+				)
+			}
 		}
 		w.WriteHeader(http.StatusMethodNotAllowed)
 		return
@@ -300,7 +305,9 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		//nolint:bodyclose
 		err := http.NewResponseController(w).EnableFullDuplex()
 		if err != nil {
-			s.logger.Warn("failed to enable full duplex", zap.Error(err))
+			if c := s.logger.Check(zapcore.WarnLevel, "failed to enable full duplex"); c != nil {
+				c.Write(zap.Error(err))
+			}
 		}
 	}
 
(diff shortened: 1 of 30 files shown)


modules/caddyhttp/server.go
```diff
@@ -759,50 +786,64 @@ func (s *Server) logRequest(
...
-		if nop, ok := GetVar(r.Context(), "unhandled").(bool); ok && nop {
-			message = "NOP"
+
+		if fields == nil {
+			size := wrec.Size()
+
+			repl.Set("http.response.status", status) // will be 0 if no response is written by us (Go will write 200 to client)
+			repl.Set("http.response.size", size)
+			repl.Set("http.response.duration", duration)
+			repl.Set("http.response.duration_ms", duration.Seconds()*1e3) // multiply seconds to preserve decimal (see #4666)
```
francislavoie, line 819: I'm not sure if it's correct to sometimes not set the replacer. It has side effects, it's not only used in logging.

dunglas: It is already not in some cases:

https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/server.go#L757-L760

Should we always set them?

dunglas: I restored the previous behavior, but maybe should we set the replacer before the existing guard clause

dunglas: (benchmark updated)

mholt (approved):
Thanks so much for contributing this, @dunglas! Can't wait to try this out :)

francislavoie:
I think this isn't working correctly. I have `debug` enabled, but `reverse_proxy`'s `upstream roundtrip` logs aren't showing up anymore.

francislavoie:
Oh sorry I'm an absolute dummy, the `reverse_proxy` was just not running (directive order etc) :joy: it's working fine.

jum:
I am seeing the use of zap.Error() in cases where the previous line contains a check for zap.Warnlevel in a few of these changes, which surely looks like a mistake to me, as it changes the level of the log line to error where previously a warning was used. At least the issuing of error messages for writing errors is back, the pull request https://github.com/caddyserver/caddy/pull/6532 does not have any effect any longer.

francislavoie:
@jum could you make a PR to fix those cases you see? Thanks for taking a look!

jum:
I am not that familiar with zap logging, but looking at the source code of zap it should not happen at all. The Check function records the level that is being checked and uses it on ce.Write, but still I do get these "aborting with incomplete response" messages on ErrorLevel. Anyone any clue how that is possible?

jum:
Ok, ignore these comments from me above. I had a small error in my CI and I build accidentally against "latest" instead of "master", I checked my docker container and it has 2.8.4 as the version, way before the patches. Sorry.

mholt:
It happens. Thanks!

Issue #6546, opened 2024-08-28
cannot use sendfile() in a http handler:


terrywh:
cannot use sendfile() in a http handler:

```
file, _ := os.Open("file")
w.(io.ReaderFrom).ReadFrom(file) 
```
OR
```
file, _ := os.Open("file")
file.WriteTo(w)
```
either way the sendfile / splice will not be called for the following reason:

1. `caddyhttp.ResponseRecorder` is using `io.Copy()` (caddyhttp/responsewriter.go:47) instead of `http.response.ReadFrom()`
> this is where the problem resides; `io.Copy` first use `WriterTo`, not `ReadFrom`
2. `*http.response` does not implement `syscall.Conn` interface;
3. file `*os.File`got wrapped in `os.fileWithoutWriteTo` (`file.go:269`)
4. `(*http.response) ReadFrom()/sendfile()` try to use TCPConn.ReadFrom but `src` is not a `*os.File` (`sendfile_linux_go:20`)
5. fallback to `genericReadFrom()` (tcpsock_posix.go:54)
(caddy/2.8.4/go/1.22.5)

_Originally posted by @terrywh in https://github.com/caddyserver/caddy/issues/4731#issuecomment-2312603747_
           
Maybe change `responsewriter.go:48` from `io.Copy` to directly call `*http.response.ReadFrom` ?

mholt:
Interesting; wouldn't io.Copy call ReadFrom for us?

> if dst implements [ReaderFrom](https://pkg.go.dev/io#ReaderFrom), the copy is implemented by calling dst.ReadFrom(src).

terrywh:
> Interesting; wouldn't io.Copy call ReadFrom for us?
> 
> > if dst implements [ReaderFrom](https://pkg.go.dev/io#ReaderFrom), the copy is implemented by calling dst.ReadFrom(src).

it dit, but after `*os.File` had been wrapped in `fileWithoutWriteTo`

WeidiDeng:
Actually [stdlib file server](https://github.com/golang/go/blob/master/src/net/http/fs.go#L423C2-L425C3) uses `io.CopyN` instead of `io.Copy`, which uses `io.LimitedReader` under the hood.

[...]

PR #6565, opened 2024-09-07
http: ReponseWriter prefer ReadFrom if available

WeidiDeng:
Fix [6546](https://github.com/caddyserver/caddy/issues/6546).


mholt (approved):
Thanks Weidi -- I am still not 100% sure why this is needed since I thought io.Copy does this, but if you are sure it works, let's go ahead and merge it. :slightly_smiling_face:

WeidiDeng:
`io.Copy` prefers `io.WriterTo` which `*os.File` implements. `*os.File` will later use `io.ReaderFrom` with a wrapper file, the wrapped file isn't recognized as a valid target for `sendfile`.

PR #6569, opened 2024-09-11
caddytls: Add sni_regexp matcher

vnxme:
This is an implementation of `sni_regexp` matcher requested in https://github.com/mholt/caddy-l4/issues/241.

Example syntax:
```
{
	layer4 {
		:8443 {
			@test tls sni_regexp ^(one|two)\\.example\\.com$
			route @test {
				proxy {tls.regexp.1}.internal:443
			}
		}
	}
}
```

Disclaimer: I haven't tested it extensively, so it's subject to feedback of the requestors.

### DIFF
--- a/modules/caddytls/matchers.go
+++ b/modules/caddytls/matchers.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"net"
 	"net/netip"
+	"regexp"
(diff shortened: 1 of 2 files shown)


mholt (approved):
LGTM. Thank you for this contribution!

PR #6571, opened 2024-09-13
rewrite: Avoid panic on bad arg count for `uri`

mister-turtle:
All cases for args[0] require additional arguments present in the slice so doing the bounds check here covers all cases. Cases requiring additional parameters already include additional checks on the len(args)

### DIFF
--- a/modules/caddyhttp/rewrite/caddyfile.go
+++ b/modules/caddyhttp/rewrite/caddyfile.go
@@ -98,7 +98,7 @@ func parseCaddyfileURI(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, err
 	h.Next() // consume directive name
 
 	args := h.RemainingArgs()
-	if len(args) < 1 {
+	if len(args) <= 1 {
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/rewrite/caddyfile.go
```diff
@@ -98,7 +98,7 @@ func parseCaddyfileURI(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, err
 	h.Next() // consume directive name
 
 	args := h.RemainingArgs()
-	if len(args) < 1 {
+	if len(args) <= 1 {
```
francislavoie, line 101: This _might_ be the wrong fix, because this means that all `uri` modes _must_ have at least one argument. That might not always be true. We might have a mode which takes a block of config instead.

The better fix would be to do the correct bounds checks inside of each `case`.

francislavoie (approved):
Thanks!

(The lint errors aren't your fault, false-positives we're still dealing with separately)

PR #6572, opened 2024-09-15

MayCXC:
third time's the charm

example Caddyfile:

```
{
	auto_https disable_redirects
	admin off
}

http://localhost {
	bind fd/{env.CADDY_HTTP_FD} {
		protocols h1
	}
	log
	respond "Hello, HTTP!"
}

https://localhost {
	bind fd/{env.CADDY_HTTPS_FD} {
		protocols h1 h2
	}
	bind fdgram/{env.CADDY_HTTP3_FD} {
		protocols h3
	}
	log
	respond "Hello, HTTPS!"
}
```

as seen above, two reserved networks `fd` and `fdgram` have been added. `fd/3` creates a FileListener from fd 3, `fdgram/4` creates a FilePacketConn from fd 4. this simplifies configuration compared to https://github.com/caddyserver/caddy/pull/6543 , but the "actual" host addresses are no longer being visible. so, for health checks and the admin interface, fds are treated similarly to unix sockets.

this is a good thing i think, fds should be decoupled from listen addresses. I think separate config options can be added to override the origin address for health checks / the admin interface if needed.


MayCXC:
link to issue: https://github.com/caddyserver/caddy/issues/6296

PR #6575, opened 2024-09-16
ci: update the linter action version

mohammed90:
🤞🏻


mholt (approved):
Definitely false positives on some of these, like complaining about out of bounds on this line:

```go
return nil, fmt.Errorf("unsupported matcher data type: %s, %s", matcherDataTypes[0], matcherDataTypes[1])
```

yet the only way to get there is:

```go
switch len(matcherDataTypes) {
// ...
	case 2:
```

So, disabling it approved, until they fix it.

PR #6577, opened 2024-09-18
Make add/remove package commands idempotent 

armadi1809:
Closes #6548


shivam-baghla (approved):

francislavoie:
Hmm, this means it'll try to fetch the build and replace it even if it doesn't make sense to do so. I think the idea is that it would just _do nothing_ if it's already good, but not throw an error.

mholt:
I agree. How would you feel about changing this patch to not perform the task if the build is not changing?

Phamousomaka (approved):

francislavoie:
I'll close this as inactive.

PR #6581, opened 2024-09-20
quic: enable qlog, controlled by QLOGDIR env

marten-seemann:
qlogs are only exported if the QLOGDIR env is set, as described in the draft: https://www.ietf.org/archive/id/draft-ietf-quic-qlog-main-schema-09.html#name-set-file-output-destination.


francislavoie:
Makes sense to me! If it's no-op unless an env var is set, that's okay.

francislavoie (approved):

mholt (approved):
Ah nice, thanks Marten!

PR #6585, opened 2024-09-21
chore: Use slices package where possible

francislavoie:
Just a simple cleanup, we can use `slices.Contains()` in a bunch of spots now instead of helper functions to do the simple loop.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -34,6 +34,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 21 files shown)


francislavoie:
2nd commit is potentially more controversial, using `ContainsFunc` (5 lines -> 3 lines).

mohammed90 (approved):
I like it, even the controversial one!

francislavoie:
Added even more.

FWIW, I was using regexp searches like this to find them :joy: might've missed some, did the ones I could find.

```
	*for [a-zA-Z_\.]*, [a-zA-Z_\.]* := range [a-zA-Z_\.]* \{
	*if [a-zA-Z_\.]* == [a-zA-Z_\.]* \{
```

mohammed90:
> FWIW, I was using regexp searches like this to find them :joy: might've missed some, did the ones I could find.
> 
> ```
> 	*for [a-zA-Z_\.]*, [a-zA-Z_\.]* := range [a-zA-Z_\.]* \{
> 	*if [a-zA-Z_\.]* == [a-zA-Z_\.]* \{
> ```

I was going to suggest semgrep, but apparently they went all proprietary 😕

mholt (approved):
Finally got through all 21 files. I think they are correct changes, if I understand the functions correctly. Thank you for the improvements!

PR #6590, opened 2024-09-24
caddyhttp: Optimize logs using zap's WithLazy()

AlliBalliBaba:
After #6541 I realized that using zap's [WithLazy()](https://github.com/uber-go/zap/blob/ede8e1888f836a9a0717a8891503f40d7998d315/logger.go#L202) for the error logger was still much more efficient even if we are cloning the request. Originally I wanted to try using this [originalRequest](https://github.com/caddyserver/caddy/blob/ff67b971267abb24774d18f323b0d6d43bfcdb3b/modules/caddyhttp/server.go#L906), which is made for logging but just gives a shallow copy without request headers. Instead, using [request.Clone()](https://pkg.go.dev/net/http#Request.Clone) gives a deep copy that just omits the request body.

This makes a simple request about 7% more efficient:
```
:80 {
    respond "Hello World"
}
```

`wrk -t 4 -c 200 -d 60 http://localhost ` (20 cores Docker Ubuntu WSL)
With()             ~205.000 RPS
WithLazy()      ~219.000 RPS

Flamegraph `With()`
torch-caddy

Flamegraph `WithLazy()` and `Copy()`
torch-caddy2

Note that this will make the case where an error is logged about 1% slower since we are doing an additional `request.Clone()`


francislavoie:
We could possible do the same in `reverseproxy.go` and `fastcgi.go`?

mholt:
That's very interesting -- thanks for submitting this PR!

I second Francis' question, but I think this LGTM either way.

AlliBalliBaba:
I can look into `reverseproxy.go` and `fastcgi.go` if there's similar potential for improvement. Probably something for another PR

mholt:
Okay, doing it in increments sounds reasonable.

I'll merge this, but before I do, why does cloning the request make it faster? That seems counterintuitive. (I wonder if we should comment next to that line, either a link to this PR or a brief explanation)

AlliBalliBaba:
Cloning the request is faster than json encoding the request and cloning the logger. But yeah it seems counterintuitive, I'll add a comment.

mholt (approved):
Thank you! Looking forward to seeing if this works well in any other places.

Issue #4873, opened 2022-07-08
Support for SSL passphrase


mmenozzi:
Hi guys! Thank you for working on Caddy, it's awesome!
I have a question: is it possible, with Caddy, to use a custom SSL certificate which also has a passphrase?

With Apache is possible and it asks for the passphrase every time the process starts. But it's also possible to avoid this by specifying a file containing the passphrase which is then configured with the [SSLPassPhraseDialog](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslpassphrasedialog) directive.

Is this supported even with Caddy in any way?

Thank you.

francislavoie:
SSL is a deprecated term. It's been called TLS since 1999.

No, Caddy doesn't have this functionality. I'm not sure we want to add it either, because it would involve a lot of complexity and I highly doubt that it would have much real-world use (probably less than 0.1% of users).

The security benefits of this are pretty questionable. It's better to just control access to the machine you have Caddy running on to make sure nobody has access to it other than trusted parties.

If a bad actor gains access to the machine, they can already do bad things by default, like use the admin API (`http://localhost:2019`) to change Caddy's config.

PR #6591, opened 2024-09-25
caddytls: Give a better error message when given encrypted private keys

francislavoie:
A user recently pointed out that they were misled because they tried to use an encrypted private key, which we don't support (see https://github.com/caddyserver/caddy/issues/4873).

I chose to read the first 40 characters of the PEM to see if it has `ENCRYPTED` to guess whether it was encrypted. I chose 40 because the length of the ASCII armor `-----BEGIN ENCRYPTED PRIVATE KEY-----` is 37, just a rough estimate of the input. I don't know how EC encrypted keys look but I assume it also uses `ENCRYPTED` in the header?

Before this, the error was:

```
Error: loading initial config: loading new config: loading http app module: provision http: getting tls app: loading tls app module: provision tls: loading certificates: tls: failed to parse private key
```

Now:

```
Error: loading initial config: loading new config: loading http app module: provision http: getting tls app: loading tls app module: provision tls: loading certificates: encrypted private keys are not supported; please decrypt the key first
```

### DIFF
--- a/modules/caddytls/fileloader.go
+++ b/modules/caddytls/fileloader.go
@@ -18,6 +18,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/caddyserver/caddy/v2"
 )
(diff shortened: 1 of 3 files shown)


mholt (approved):
Nice idea -- thanks!

PR #6593, opened 2024-09-27
caddytls: Allow disabling storage cleaning, avoids writing two files

francislavoie:
Some users want to run Caddy in a totally read-only mode, but currently storage cleaning implicitly creates two files `instance.uuid` and `last_clean.json` which are undesirable in that situation. We can simply skip starting the storage cleaning goroutine via a config option.

I noticed `storage_check off` wasn't configurable in the Caddyfile, so I added that too, while I was at it.

### DIFF
--- a/caddyconfig/httpcaddyfile/options.go
+++ b/caddyconfig/httpcaddyfile/options.go
@@ -39,7 +39,8 @@ func init() {
 	RegisterGlobalOption("fallback_sni", parseOptSingleString)
 	RegisterGlobalOption("order", parseOptOrder)
 	RegisterGlobalOption("storage", parseOptStorage)
-	RegisterGlobalOption("storage_clean_interval", parseOptDuration)
+	RegisterGlobalOption("storage_check", parseStorageCheck)
+	RegisterGlobalOption("storage_clean_interval", parseStorageCleanInterval)
 	RegisterGlobalOption("renew_interval", parseOptDuration)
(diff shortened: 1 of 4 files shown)


mholt:
Thanks for working on this! A couple questions/thoughts:

- What is the harm of trying to write these two files?
- Maybe we should just have a `read_only` mode in the global options (if it is actually needed). I don't know if I love this granularity since you need to turn several knobs for one thing.

francislavoie:
Forgot to link the related forum thread https://caddy.community/t/disabling-file-storage/25772, those 2 files get written which is undesirable in a setup with the goal of being read-only.

jleedev:
I often want to use read-only mode when I'm running an ephemeral Caddy instance purely for testing, e.g. to verify a server's behavior behind a reverse proxy. Something trivial like `caddy respond hello` should _not_ be complaining about "unable to get instance ID; storage clean stamps will be incomplete" and "could not clean default/global storage", and moreover this does not give me confidence that I can run the program purely as an isolated, single-use shell tool (which, to be clear, Caddy does not claim to be; compare this to how it wants to install a local certificate authority).

If the main (only?) use of the "storage" module relates to certificate maintence, it should be reasonable to avoid running this when TLS is disabled. If I run `caddy file-server` _without_ giving a `--domain` flag, it should have no reason to be writing these.

mholt (approved):

Issue #6584, opened 2024-09-21
Feature Request: CEL method for `env` placeholder


polarathene:
Short version: The ability to use placeholders with `env` to lookup different ENV based on placeholder value at the time, or a CEL function for matchers to accomplish the same.

---

From what I understand (at least with Caddyfile support), an explicit ENV var can be used and is parsed with the Caddyfile, or `env` placeholder can be used, similar to `file` that does a lookup at runtime via method instead.

Snippets can be used with args to compose the value for `env` or `file`, but the inputs must be known at the time, you could not use `vars` or `path` placeholders for example.

My use case was for matching against content sourced from an ENV, but with:
- The ENV itself dependent upon another placeholder such as `{path.0}.upperAscii()`, `{re.1}`, etc.
- Composition of that ENV with a suffix for variant queries.

This could be achieved either via a CEL equivalent method where the value can be provided as a string using CEL like the above path example, or if the existing placeholder support could handle `{env.{vars.example}}`.

I haven't tried with map directive yet, but expect that would be a similar issue. Presumably the only option with Caddy for this right now is snippets and creating an `import` with args for each permutation upfront. Alternatively writing a custom caddy plugin or a separate service to perform the check (to allow or deny a route).

The intent was to leverage caddy's existing features in a way that a docker image could be deployed and users only needed to manage some permissions via ENV (_such as for `path_regexp` matchers to refer to_).

francislavoie:
The `vars` matcher is able to read placeholders. For example:

```
:80 {
	@var vars {env.USER} francis
	respond @var "hey francis"
	respond "nope"
}
```

So I think really the request is moreso "provide the `vars` matcher in CEL".

I couldn't remember why we didn't implement it, so I went looking. For ref https://github.com/caddyserver/caddy/pull/4715#issue-1207649128 and https://github.com/caddyserver/caddy/pull/4715#issuecomment-1163585563.

TL;DR the reason is that since we do a fancy regexp replacement for placeholders in CEL already to turn them into a magic function call, having a matcher that takes a placeholder as input is hard from a parsing standpoint.

Will need to do some thinking to look again whether we can add `vars` to CEL, might need a bit of reworking of the regexp replacement for placeholders to allow an escape `\{` at the start for use as a regular string in matcher inputs.

francislavoie:
[...]

PR #6594, opened 2024-09-27
caddyhttp: Escaping placeholders in CEL, add `vars` and `vars_regexp`

francislavoie:
Related to https://github.com/caddyserver/caddy/issues/6584, prior work in https://github.com/caddyserver/caddy/pull/4715#issuecomment-1163585563

Our current CEL placeholder regexp is too naive. It replaces everything that looks like a placeholder, and offers no good option to opt out if a literal placeholder-looking input is given to some CEL functions as input.

To solve this, we can do the placeholder replacement in two steps.
- First, we match whether the placeholder _is not_ preceded by a `\` (or start of a line) and replace it and preserving the preceding character which is matched (we now have two capture groups so `${1}` has to go to the start)
- Second, we match for a placeholder which _does_ have a preceding `\`, and drop the `\`.

That gives a way to escape through the `caddyPlaceholder()` replacement and get an input directly to the given matcher.

One interesting quirk: CEL itself has its own `\\` escape sequence, so depending on whether we want the placeholder to be replaced by the matcher (e.g. `header` matcher does do placeholder replacements on its inputs at runtime) or we want the value to be raw determines how many backslashes we need. The test shows this pretty well I think:
- If the `header` matcher should take the value as-is and not perform placeholder replacement (because the match value is also placeholder-like), then we need _three_ backslashes, like `\\\{foobar}`. This is because the last one of the three is for escaping `caddyPlaceholder()`, then the prior two are collapsed into one by CEL's parsing itself, then the last one is to escape the placeholder replacer, and the result is a clean `{foobar}` matching value.
- If the `header` matcher itself should perform placeholder replacement (not done in the CEL matcher, but deeper in the matcher itself) then a single backslash is used, like `\{http.request.uri.path}` (or `\{path}` in the Caddyfile). This only escapes past the `caddyPlaceholder()` regexp, but does not escape past the placeholder replacer which runs inside `header`.

We'll need to document this, but it's tricky :sweat_smile: 

After doing the above, I decided to also implement `vars` and `vars_regexp` support in CEL, which we skipped implementing because of the above placeholder limitations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,7 +167,19 @@ jobs:
           go version
           go env
           printf "\n\n"
-          CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./...
+          retries=3
+          exit_code=0
+          while ((retries > 0)); do
+            CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./...
+            exit_code=$?
+            if ((exit_code == 0)); then
+              break
+            fi
+            echo "\n\nTest failed: \$exit_code, retrying..."
+            ((retries--))
+          done
+          echo "Remote exit code: \$exit_code"
+          exit \$exit_code
           EOF
           test_result=$?
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,13 +156,23 @@ jobs:
           # short sha is enough?
           short_sha=$(git rev-parse --short HEAD)
 
+          # To shorten the following lines
+          ssh_opts="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+          ssh_host="$CI_USER@ci-s390x.caddyserver.com"
+
           # The environment is fresh, so there's no point in keeping accepting and adding the key.
-          rsync -arz -e "ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" --progress --delete --exclude '.git' . "$CI_USER"@ci-s390x.caddyserver.com:/var/tmp/"$short_sha"
-          ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -t "$CI_USER"@ci-s390x.caddyserver.com "cd /var/tmp/$short_sha; go version; go env; printf "\n\n";CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./..."
+          rsync -arz -e "ssh $ssh_opts" --progress --delete --exclude '.git' . "$ssh_host":/var/tmp/"$short_sha"
+          ssh $ssh_opts -t "$ssh_host" bash <<EOF
+          cd /var/tmp/$short_sha
+          go version
+          go env
+          printf "\n\n"
+          CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./...
+          EOF
           test_result=$?
 
           # There's no need leaving the files around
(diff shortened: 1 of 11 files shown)


francislavoie:
Couple more commits:

- I noticed we were repeating `"request"` a lot, i.e. the variable we push into the CEL context. I made a const for that.
- Similarly, I noticed we were using `"caddyPlaceholder"` in a couple places outside of the `caddyhttp` package (file matcher) so I renamed the const and exported it.
- I renamed `request` to `req`, and `caddyPlaceholder` to `placeholder`, because actually users can use it in their own CEL without the magic shortcut, so making it more ergonomic is nice.
- Bumped the CEL version to latest
- Registered a few CEL extensions to provide some more functions to users (see https://github.com/google/cel-go/blob/master/ext/README.md)

.github/workflows/ci.yml
```diff
@@ -156,13 +156,23 @@ jobs:
...
-          ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -t "$CI_USER"@ci-s390x.caddyserver.com "cd /var/tmp/$short_sha; go version; go env; printf "\n\n";CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./..."
+          rsync -arz -e "ssh $ssh_opts" --progress --delete --exclude '.git' . "$ssh_host":/var/tmp/"$short_sha"
+          ssh $ssh_opts -t "$ssh_host" bash <<EOF
+          cd /var/tmp/$short_sha
+          go version
+          go env
+          printf "\n\n"
+          CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./...
+          EOF
           test_result=$?
```
mohammed90, line 172: Love this! 😀

francislavoie: Yeah, nicer to read it out multiline!

The point of this is I was trying to implement a `while` loop to retry the tests (to smooth over the flakiness short-term) but it wasn't working :man_facepalming: just a mess of dealing with the exit code coming through the shell script. I'll paste you what I had in slack if you wanna take a crack at it.

.github/workflows/ci.yml
```diff
@@ -156,13 +156,35 @@ jobs:
...
+          printf "\n\n"
+          retries=3
+          exit_code=0
+          while ((retries > 0)); do
+            CGO_ENABLED=0 go test -p 1 -tags nobadger -v ./...
+            exit_code=$?
+            if ((exit_code == 0)); then
+              break
+            fi
+            echo "\n\nTest failed: \$exit_code, retrying..."
```
mholt, line 178: I feel like we should "just" fix the tests :see_no_evil:

mholt (approved):
LGTM I guess!

Did you still want to shorten things to `ph` and `req`?

.github/workflows/ci.yml
francislavoie: I mean... that's WIP in another PR... easier said than done.

francislavoie:
`req` yes, it's not `ph` anymore I went with `placeholder`... but `get` was suggested, idk what you think of that.

.github/workflows/ci.yml
mholt: Ah right, I remember now.

francislavoie:
After a quick chat with Matt on slack, decided to go with `ph(req, <string>)` as the shortcut (FYI @polarathene). Also added a matching shortcut to templates cause I noticed we had `placeholder` as a function in there as well. Cause why not. Some semblance of consistency I guess.

PR #6596, opened 2024-09-29
caddyhttp: Add `MatchWithError` to replace SetVar hack

francislavoie:
As recently discussed on Slack, our `RequestMatcher` interface is sometimes insufficient in situations where matchers cannot safely operate because it can only return a boolean. Since matchers make a decision whether or not handlers can run, returning false isn't a secure decision because it may cause handlers that _must_ run (e.g. authentication) to not get run, causing possible bypass of security-critical functionality in the config.

We already had a mechanism for this, which I added in https://github.com/caddyserver/caddy/pull/4346 via putting a variable in Context and reading it back after the matcher ran. That was absolutely a hack and not the best solution. I couldn't think of a better way at the time (lack of Go skills :sweat_smile:)

To solve this, the better solution is to introduce a new `RequestMatcherWithError` interface which returns `(bool, error)`, so that an error can be returned to cancel the HTTP request pipeline if the matcher's preconditions aren't met. For now, we can make this optional (so if a matcher implements this interface we call it if possible) until we know all matcher plugins have caught up implementing with this (though unfortunately guaranteeing that is quite difficult and a multi-step process to clean up).

### DIFF
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -36,6 +36,10 @@ func init() {
 // RequestMatcher is a type that can match to a request.
 // A route matcher MUST NOT modify the request, with the
 // only exception being its context.
+//
+// DEPRECATED: Matchers should now implement RequestMatcherWithError.
+// You may remove any interface guards for RequestMatcher
+// but keep your Match() methods for backwards compatibility.
 type RequestMatcher interface {
 	Match(*http.Request) bool
 }
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -40,6 +40,18 @@ type RequestMatcher interface {
 	Match(*http.Request) bool
 }
 
+// RequestMatcherWithError is like RequestMatcher but can return an error.
+// An error during matching will abort the request middleware chain and
+// invoke the error middleware chain.
+//
+// This will eventually replace RequestMatcher. Matcher modules
+// should implement both interfaces, and once all modules have
+// been updated to use RequestMatcherWithError, the RequestMatcher
+// interface may eventually be dropped.
+type RequestMatcherWithError interface {
+	MatchWithError(*http.Request) (bool, error)
+}
+
 // Handler is like http.Handler except ServeHTTP may return an error.
 //
 // If any handler encounters an error, it should be returned for proper
--- a/modules/caddyhttp/caddyhttp.go
+++ b/modules/caddyhttp/caddyhttp.go
@@ -37,7 +37,7 @@ func init() {
 // A route matcher MUST NOT modify the request, with the
 // only exception being its context.
 //
-// DEPRECATED: Matchers should now implement RequestMatcherWithError.
+// Deprecated: Matchers should now implement RequestMatcherWithError.
 // You may remove any interface guards for RequestMatcher
 // but keep your Match() methods for backwards compatibility.
 type RequestMatcher interface {
--- a/modules/caddyhttp/ip_matchers.go
+++ b/modules/caddyhttp/ip_matchers.go
@@ -145,11 +145,11 @@ func (m *MatchRemoteIP) Provision(ctx caddy.Context) error {
 
 // Match returns true if r matches m.
 func (m MatchRemoteIP) Match(r *http.Request) bool {
-	matches, err := m.MatchWithError(r)
+	match, err := m.MatchWithError(r)
 	if err != nil {
 		SetVar(r.Context(), MatcherErrorVarKey, err)
 	}
-	return matches
+	return match
 }
 
 // MatchWithError returns true if r matches m.
@@ -249,11 +249,11 @@ func (m *MatchClientIP) Provision(ctx caddy.Context) error {
 
 // Match returns true if r matches m.
 func (m MatchClientIP) Match(r *http.Request) bool {
-	matches, err := m.MatchWithError(r)
+	match, err := m.MatchWithError(r)
 	if err != nil {
 		SetVar(r.Context(), MatcherErrorVarKey, err)
 	}
-	return matches
+	return match
 }
 
 // MatchWithError returns true if r matches m.
@@ -348,13 +348,11 @@ func matchIPByCidrZones(clientIP netip.Addr, zoneID string, cidrs []*netip.Prefi
 
 // Interface guards
 var (
-	_ RequestMatcher          = (*MatchRemoteIP)(nil)
 	_ RequestMatcherWithError = (*MatchRemoteIP)(nil)
 	_ caddy.Provisioner       = (*MatchRemoteIP)(nil)
 	_ caddyfile.Unmarshaler   = (*MatchRemoteIP)(nil)
 	_ CELLibraryProducer      = (*MatchRemoteIP)(nil)
 
-	_ RequestMatcher          = (*MatchClientIP)(nil)
 	_ RequestMatcherWithError = (*MatchClientIP)(nil)
 	_ caddy.Provisioner       = (*MatchClientIP)(nil)
 	_ caddyfile.Unmarshaler   = (*MatchClientIP)(nil)
--- a/modules/caddyhttp/ip_matchers.go
+++ b/modules/caddyhttp/ip_matchers.go
@@ -108,7 +108,7 @@ func (MatchRemoteIP) CELLibrary(ctx caddy.Context) (cel.Library, error) {
 		// internal data type of the MatchPath value.
 		[]*cel.Type{cel.ListType(cel.StringType)},
 		// function to convert a constant list of strings to a MatchPath instance.
-		func(data ref.Val) (RequestMatcher, error) {
+		func(data ref.Val) (RequestMatcherWithError, error) {
 			refStringList := reflect.TypeOf([]string{})
 			strList, err := data.ConvertToNative(refStringList)
 			if err != nil {
@@ -218,7 +218,7 @@ func (MatchClientIP) CELLibrary(ctx caddy.Context) (cel.Library, error) {
 		// internal data type of the MatchPath value.
 		[]*cel.Type{cel.ListType(cel.StringType)},
 		// function to convert a constant list of strings to a MatchPath instance.
-		func(data ref.Val) (RequestMatcher, error) {
+		func(data ref.Val) (RequestMatcherWithError, error) {
 			refStringList := reflect.TypeOf([]string{})
 			strList, err := data.ConvertToNative(refStringList)
 			if err != nil {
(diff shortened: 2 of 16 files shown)


modules/caddyhttp/ip_matchers.go
```diff
@@ -254,6 +259,11 @@ func (m MatchClientIP) Match(r *http.Request) bool {
 	return matches
 }
 
+// MatchWithError returns true if r matches m.
+func (m MatchClientIP) MatchWithError(r *http.Request) (bool, error) {
+	return m.Match(r), nil
+}
```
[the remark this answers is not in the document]
francislavoie: Yeah that would be the next step, just didn't want to do it in the same commit.

francislavoie:
I kept the error in context for full backwards compat for places that still call Match. I think we need to do this in a few steps over the course of a few releases

mholt:
> I kept the error in context for full backwards compat for places that still call Match.

Does anyone really know about or use that though?

Anyway, when you're ready to have this merged, go ahead and mark as ready. I'm OK with it if there's another commit to actually use the error return value gets into this PR, or we can do that later.

francislavoie:
> Does anyone really know about or use that though?

I'll do some searches with SourceGraph just in case.

mholt (approved):

francislavoie:
Plugged the IP matcher hole, updated the tests to use `MatchWithError` instead, everywhere.

modules/caddyhttp/ip_matchers.go
```diff
@@ -145,25 +145,36 @@ func (m *MatchRemoteIP) Provision(ctx caddy.Context) error {
...
+	}
+	return match
+}
+
+// MatchWithError returns true if r matches m.
+func (m MatchRemoteIP) MatchWithError(r *http.Request) (bool, error) {
+	// if handshake is not finished, we infer 0-RTT that has not verified remote IP; could be spoofed
 	if r.TLS != nil && !r.TLS.HandshakeComplete {
-		return false // if handshake is not finished, we infer 0-RTT that has not verified remote IP; could be spoofed
+		return false, fmt.Errorf("TLS handshake not complete, remote IP cannot be verified")
```
francislavoie, line 159: @mholt should it this be a `caddyhttp.Error`? If so, what HTTP status code should we use? I think as-is it would be a 500 without being otherwise specified but I'm not sure if that's correct. I don't know how it should behave on an incomplete handshake. Is there some 4xx series code which can instruct the client to try again or w/e?

francislavoie:
@mholt I've changed enough in the past few commits that I think it warrants another review.

modules/caddyhttp/ip_matchers.go
steffenbusch: > Is there some 4xx series code which can instruct the client to try again or w/e?

@francislavoie, I was able to work around the issue mentioned here https://github.com/caddyserver/caddy/pull/6427#issuecomment-2408115697 with this in front of my ip matchers.

```Caddyfile
@early tls early_data
error @early 425
```

Before I have added this, I was able to reproduce my issue where the IP matcher "failed" (returned false for my IP address) after approx. 10 minutes of inactivtiy and then pressing F5 to refresh the page.
After have this added, I was no longer able to reproduce the issue.

Based on that experience, I would think the HTTP status code **425** is the desirable one here.

francislavoie: Ah thanks, good to know! I'm away from home today, would you be willing to checkout this branch, change this line to a `caddyhttp.Error`, build it and test it? Would be nice to confirm.

steffenbusch: I will try it - just a bit difficult in that particular environment, because my 2.9.0 tests require the "promote-metrics" branch. But I have other hosts for testing, where I can disable metrics.

steffenbusch: It's looking good 👍
My change was this: https://github.com/caddyserver/caddy/compare/matcher-with-error...steffenbusch:caddy:matcher-with-error

**Before:**

With 2.9.0-beta from "promote-metrics" branch, which means without your change, I got the undesired 404 because the IP-matcher was not matching with my IP address; followed by a success on a F5:

`tail -f access.json  | jq '{user_id, status, ts, tls: .request.tls, remote_ip: .request.remote_ip, uri: .request.uri}'`

```json
{
  "user_id": "",
  "status": 404,
  "ts": 1728816499.5479946,
  "tls": {
    "resumed": true,
    "version": "TLSv1.3",
    "cipher_suite": "TLS_AES_128_GCM_SHA256",
    "proto": "h3",
    "server_name": "rambo.stbu.net"
  },
  "remote_ip": "95.89.172.31",
  "uri": "/test/ip"
}
{
  "user_id": "",
  "status": 200,
  "ts": 1728816503.3737376,
  "tls": {
    "resumed": true,
    "version": "TLSv1.3",
    "cipher_suite": "TLS_AES_128_GCM_SHA256",
    "proto": "h3",
    "server_name": "rambo.stbu.net"
  },
  "remote_ip": "95.89.172.31",
  "uri": "/test/ip"
}
```

**After:**

Building it based on your branch with the httperror 425 tweak, it was working in my browser, and this was logged in access.json:

```json
{
  "user_id": "",
  "status": 425,
  "ts": 1728817672.261492,
  "tls": {
    "resumed": true,
    "version": "TLSv1.3",
    "cipher_suite": "TLS_AES_128_GCM_SHA256",
    "proto": "h3",
    "server_name": "rambo.stbu.net"
  },
  "remote_ip": "95.89.172.31",
  "uri": "/test/ip-chrome"
}
{
  "user_id": "",
  "status": 200,
  "ts": 1728817672.265926,
  "tls": {
    "resumed": true,
    "version": "TLSv1.3",
    "cipher_suite": "TLS_AES_128_GCM_SHA256",
    "proto": "h3",
    "server_name": "rambo.stbu.net"
  },
  "remote_ip": "95.89.172.31",
  "uri": "/test/ip-chrome"
}
```

The `425` is logged, but I have not observed any issue in my Chrome Browser (maybe a little bit longer the "pending" than usually) and saw the result of the request immediately, which was logged at `200` instantly after the `425`.

However, I noticed a build-error for this branch related to github.com/mholt/caddy-ratelimit:

```
2024/10/13 11:27:46 [INFO] exec (timeout=0s): /usr/local/go/bin/go build -o /usr/local/bin/caddy -ldflags -w -s -trimpath -tags nobadger
# github.com/mholt/caddy-ratelimit
/root/go/pkg/mod/github.com/mholt/caddy-ratelimit@v0.0.0-20240828171918-12435ecef5db/handler.go:176:7: multiple-value rl.matcherSets.AnyMatch(r) (value of type (bool, error)) in single-value context
2024/10/13 11:29:15 [INFO] Cleaning up temporary folder: /tmp/buildenv_2024-10-13-1126.189145495
2024/10/13 11:29:15 [FATAL] exit status 1
Error: building at STEP "RUN go install github.com/caddyserver/xcaddy/cmd/xcaddy@latest &&     /usr/local/bin/xcaddy build --output /usr/local/bin/caddy       --with github.com/caddyserver/caddy/v2=github.com/caddyserver/caddy/v2@matcher-with-error       --with github.com/caddyserver/transform-encoder       --with github.com/ueffel/caddy-tls-format       --with github.com/muety/caddy-remote-host       --with github.com/ggicci/caddy-jwt       --with github.com/caddy-dns/ionos       --with github.com/ueffel/caddy-brotli       --with github.com/aksdb/caddy-cgi/v2       --with github.com/mholt/caddy-webdav       --with github.com/mholt/caddy-ratelimit &&     /usr/local/bin/caddy version": while running runtime: exit status 1
```

For the above tests, I have removed the ratelimit plugin to get it compiled.

francislavoie: Excellent, thanks! Will apply the fix later today. Good callout on AnyMatch with ratelimit, I assumed no plugins were using that so I made a more aggressive API change. I'll fix that up.

francislavoie: Okay the `AnyMatch` issue should be fixed now, and changed to use 425 status, if you want to confirm I did it right :)

mohammed90: > You could argue it's a 4xx error since the client should not have done a 0-RTT, but does it know that? Probably not.

Per the [RFC](https://www.rfc-editor.org/rfc/rfc8470), 425 is the correct status, as mentioned earlier:

> The server can cause a client to retry individual requests and
       not use early data by responding with the 425 (Too Early) status
       code ([Section 5.2](https://www.rfc-editor.org/rfc/rfc8470#section-5.2)) in cases where the risk of replay is judged
       too great.

There are some caveates such as

>    Note that a server cannot choose to selectively reject early data at
   the TLS layer.  TLS only permits a server to either accept all early
   data or none of it.  Once a server has decided to accept early data,
   it MUST process all requests in early data, even if the server
   rejects the request by sending a 425 (Too Early) response.

>  A server cannot make a request that contains the Early-Data header
   field safe for processing by waiting for the handshake to complete.
   A request that is marked with Early-Data was sent in early data on a
   previous hop.  Requests that contain the Early-Data header field and
   cannot be safely processed MUST be rejected using the 425 (Too Early)
   status code.

modules/caddyhttp/caddyhttp.go
```diff
@@ -36,10 +36,26 @@ func init() {
...
+// but keep your Match() methods for backwards compatibility.
 type RequestMatcher interface {
 	Match(*http.Request) bool
 }
 
+// RequestMatcherWithError is like RequestMatcher but can return an error.
+// An error during matching will abort the request middleware chain and
+// invoke the error middleware chain.
+//
+// This will eventually replace RequestMatcher. Matcher modules
```
mholt, line 51: I'm kinda bummed because if we did this I would want it to be `Match(*http.Request) (bool, error)`.

mholt (approved):
Thanks for the thoughtful work on this. I don't love that we have (?) to do this, but I guess that's how it is.

If we are going to make this an actual breaking change in the future, I'd rather just put an error return on the signature of `Match`.

modules/caddyhttp/caddyhttp.go
francislavoie: I agree. I don't see how. It's like `DialContext` etc. Something to redo in v3 lmao

mholt: Maybe, but if we remove this interface, it'll be a breaking change anyway, so might as well break it into something that we _want_ right? :sweat_smile:

francislavoie: We can:
- Add `RequestMatcherWithError` (this)
- Give plugins time to implement it, all plugins should stop using the interface guard for `RequestMatcher`
- Later, change the signature of `RequestMatcher` to `Match(*http.Request) (bool, error)` (should be safe since plugins no longer use it)
- Deprecate `RequestMatcherWithError`, preferring `RequestMatcher` if implemented
- Remove `RequestMatcherWithError`

All this might take months or years but that would be the path.

mholt: That's what I'm thinking too. It'll kind of suck, but... might be worth it.

Issue #6481, opened 2024-07-27
No matching certificate to load: decoding certificate metadata: invalid character '}'


solracsf:
Caddy v2.8.4 started printing an error for one domain:
```
Jul 27 12:06:59 caddy-proxy-sbg7 caddy[1208]: {"level":"info","ts":1722074819.7958932,"logger":"tls.on_demand","msg":"obtaining new certificate","remote_ip":"77.141.97.142","remote_port":"49984","server_name":"server.example.com"}
Jul 27 12:06:59 caddy-proxy-sbg7 caddy[1208]: {"level":"error","ts":1722074819.8085268,"logger":"tls.on_demand","msg":"loading newly-obtained certificate from storage","remote_ip":"77.141.97.142","remote_port":"49984","server_name":"server.example.com","error":"no matching certificate to load for server.example.com: decoding certificate metadata: invalid character '}' after top-level value"}
```
```
{
        "sans": [
                "server.example.com"
        ],
        "issuer_data": {
                "url": "https://acme-v02.api.letsencrypt.org/acme/cert/042fb47e52c51ab43af43257f1ba5e8",
                "ca": "https://acme-v02.api.letsencrypt.org/directory",
                "renewal_info": {
                        "suggestedWindow": {
                                "start": "2024-09-23T06:58:45.333333334Z",
                                "end": "2024-09-25T06:58:45.333333334Z"
                        },
                        "_uniqueIdentifier": "kydGmAOpUWiOmNbEI.BC-0flLFGrQ69DL_idV_G6Xo",
                        "_retryAfter": "2024-07-27T09:40:01.93861755+02:00",
                        "_selectedTime": "2024-09-24T03:22:43Z"
                }
        }
}}
```
Removing the extra `}` at the bottom of the file `/caddy/certificates/acme-v02.api.letsencrypt.org-directory/server.example.com/server.example.com.json` fixes the error.

mholt:
Hey Skifree, thanks for reporting this. :wink: 

That's weird! @elee1766 Is this possibly due to a race condition you were referring to in Slack? I don't think I've seen a malformed JSON file before, only an empty one.

elee1766:
this looks like one file is written on top of another without clearing the first once.

if fsync is called after truncate and after write but before flush this could happen I think.

so yes, my guess is the temp file approach fixes this @mholt

mholt:
[...]

PR #6606, opened 2024-10-05
metrics: move `metrics`  up, outside `servers`

### LINKED ISSUE #6604: with per_host in metrics, how about compatible with old version
current caddyfile is using:
{
	servers {
		metrics {
		
		}
	}
}

if i switch to caddy@master with per_host in metrics, then above config will cause panic!! because missing :port (ie servers :80, servers :443 )

mohammed90:
This change moves the metrics configuration from per-server level to a single config knob within the `http` app. Enabling `metrics` in any of the configured servers inside `http` enables metrics for all servers.

Fix #6604

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -15,6 +15,7 @@
 package httpcaddyfile
 
 import (
+	"cmp"
 	"encoding/json"
 	"fmt"
 	"net"
@@ -186,12 +187,25 @@ func (st ServerType) Setup(
 		return nil, warnings, err
 	}
 
+	// hoist the metrics config from per-server to global
+	metrics, _ := options["metrics"].(*caddyhttp.Metrics)
+	for _, s := range servers {
+		if s.Metrics != nil {
+			metrics = cmp.Or[*caddyhttp.Metrics](metrics, &caddyhttp.Metrics{})
+			metrics = &caddyhttp.Metrics{
+				PerHost: metrics.PerHost || s.Metrics.PerHost,
+			}
+			s.Metrics = nil // we don't need it anymore
+		}
+	}
+
 	// now that each server is configured, make the HTTP app
 	httpApp := caddyhttp.App{
 		HTTPPort:      tryInt(options["http_port"], &warnings),
 		HTTPSPort:     tryInt(options["https_port"], &warnings),
 		GracePeriod:   tryDuration(options["grace_period"], &warnings),
 		ShutdownDelay: tryDuration(options["shutdown_delay"], &warnings),
+		Metrics:       metrics,
 		Servers:       servers,
 	}
 
--- a/caddyconfig/httpcaddyfile/options.go
+++ b/caddyconfig/httpcaddyfile/options.go
@@ -24,6 +24,7 @@ import (
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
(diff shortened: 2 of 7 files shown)


francislavoie (approved):

aa-shalin:
any idea when this will be merged? I ran into this issue after i was advised to use the [beta version of Caddy](https://github.com/caddyserver/caddy/issues/6481#issuecomment-2414510538)

mholt (approved):

erikschul:
This looks wrong to me.

I argue that it's reasonable to want to enable a metrics route for a specific server (e.g. `:9000`), and not serve it on all servers (including public ones). Metrics generally aren't made publicly available.

francislavoie:
This isn't about serving metrics it's about enabling metrics.

erikschul:
Okay; if it's technically complicated, maybe a better approach is to keep the metrics internal to `:2019/metrics` and then use
```
route /metrics {
    reverse_proxy localhost:2019/metrics
}
```
wherever you want to expose it?

francislavoie:
I think you're confused. There's a `metrics` _http handler_ and a `metrics` _global option_. The former is to _server_ the metrics (from any endpoint you want, none by default obviously, except for the admin endpoint) and the latter is to _enable HTTP metrics_. This PR is about moving the global option _only_.

mohammed90:
> Okay; if it's technically complicated, maybe a better approach is to keep the metrics internal to `:2019/metrics` and then use
> 
> ```
> route /metrics {
>     reverse_proxy localhost:2019/metrics
> }
> ```
> 
> wherever you want to expose it?

You're confusing 2 things: enabling HTTP handling metrics collection and exposure/serving on the admin endpoint, and exposing metrics on an endpoint other than the admin endpoint. Enabling metrics here (in this PR), means enabling the  HTTP traffic metrics collection. They're exposed only on admin endpoint (default: `:2019/metrics`). To expose metrics on an endpoint other than the admin endpoint, you'd use the [`metrics` handler](https://caddyserver.com/docs/caddyfile/directives/metrics).

Previously (before this PR), enabling HTTP metrics collection was done by adding the key `metrics` in the JSON inside a particular "server" (see the empty `metrics` object on this page: https://caddyserver.com/docs/json/apps/http/). Image for posterity:

image

This is misleading because metrics were always enabled for all servers, plus a bunch of other issues. The fix requires lifting the `metrics` object to be direct descendant of `http` (outside `servers`) because the collection of HTTP metrics is either enabled for all servers or none. `server` in this context means a logical group of listeners with unified behavioral pattern, not referring to Caddy as a whole. If you want HTTP metrics, it's enabled on the HTTP app, not "per server", so this PR is correct.



> wherever you want to expose it?

Use the [`metrics` handler](https://caddyserver.com/docs/caddyfile/directives/metrics) for this. 



> I think what confuses me is the line `s.Metrics = nil`.
> because I would think that this means that the metrics are disabled for each server?

That's not the correct interpretation, not that it was ever possible anyways. Nil-ing the field doesn't change the behavior. The line you're looking at is about adapting the old configuration to the new one. Once we copied the configuration, the older data is no longer meaningful because it's not used. We only need it there for existing users who haven't changed their config. We don't want to suddenly break them, so I added a translation.



> a) the individual server configuration is no longer possible (deleted)

It never was. There's a label for the server name to track which "server" it's part of. As a caddyfile user, you'll rarely have to deal with the notion of server, unless you serve sites on multiple ports other than 80/443.

> b) the global configuration enables metrics globally, overwriting the setting with the last value of the iteration:

Nothing is lost anyways. The field `PerHost` is only introduced in 2.9.0-beta. The lines you're looking at ensure that if the user still used the older field, and they happen to set `per_host` to `true` in any of their servers, the value is carried globally.

erikschul:
Thanks @mohammed90  for the explanation! I'm new to the project but I hope I'll find ways to contribute as I get more involved. Thanks again!

PR #6608, opened 2024-10-06
forwardauth: Skip copying missing response headers

### LINKED ISSUE #6610: ForwardAuth with copy_headers leaves template code in header if auth backend doesn't set that header
Caddy 2.8.4

Using Caddy with Authelia, I have Authelia configured to only do 2FA with certain URIs & to completely bypass certain other URIs. For those URIs that completely bypass authentication Authelia cannot & does not set a `Remote-User` header.

Unfortunately Caddy doesn't like that very much. With a simple configuration snippet such as

```
https://… {
	forward_auth https://authelia-server:8000 {
		uri /api/authz/forward-auth
		copy_headers Remote-User Remote-Groups Remote-Email Remote-Name
	}
	reverse_proxy https://…
}
```

what happens is that Caddy sets the `Remove-User` header correctly when Authelia sets the header, too. But when Authelia does _not_ set the `Remove-User` header (or any of the others), what Caddy sends to the proxied server is actually the following:

```
Remote-User: {http.reverse_proxy.header.Remote-User}
```

Yes, it leaves the template code as-is in the headers it sends to the proxy server. Same for the any other header configured via `copy_headers` that isn't set by Authelia.

This means I'll have to include the bypassed URIs from Authelia's configuration in Caddy's configuration & only send requests for other URIs to Authelia, making this whole configuration much more complex & difficult to maintain.

francislavoie:
Reported here: https://caddy.community/t/copy-header-copying-empty-headers-with-authelia/25885, but I think it has been mentioned elsewhere before as well. Fixes #6610

Basically, because we use `ReplaceKnown` in the `request_header` handler for placeholders, the `copy_header` shortcut for `forward_auth` would cause the request headers to be set to the placeholder string itself instead of skipping, when the response was missing that header.

The fix is to use a `not vars` matcher to check that the response header placeholder has a non-empty value first.

Unfortunately this produces some super verbose JSON config, but the behaviour will be more correct.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/forward_auth_authelia.caddyfiletest
+++ b/caddytest/integration/caddyfile_adapt/forward_auth_authelia.caddyfiletest
@@ -1,6 +1,6 @@
 app.example.com {
 	forward_auth authelia:9091 {
-		uri /api/verify?rd=https://authelia.example.com
+		uri /api/authz/forward-auth
 		copy_headers Remote-User Remote-Groups Remote-Name Remote-Email
 	}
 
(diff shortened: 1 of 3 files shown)


mholt (approved):
Cool, thanks for maintaining this!

PR #6611, opened 2024-10-07
caddytls: Drop `rate_limit` and `burst`, has been deprecated

francislavoie:
It's time to get rid of these, they're not working correctly and introduce more problems than they solve, when used.


francislavoie:
I guess so, but no harm if they stay there forever tbh. 2 liners are pretty free.

PR #6616, opened 2024-10-10
caddyfile: Fix comma edgecase in address parsing

francislavoie:
When the site address line looks like `foo  ,  bar` (i.e. spaces around the `,` and not simply a comma trailing an address), then it would cause the site address to include an empty space which messes up parsing in subtle ways. This fixes that case.

Closes https://github.com/caddyserver/website/issues/395

Given this Caddyfile:

```
foo.example.com , example.com , www.example.com {
    respond "foo"
}
```

Before (for some reason, no host matchers are produced, but it does automate the names):
```
{"apps":{"http":{"servers":{"srv0":{"listen":[":443"],"routes":[{"handle":[{"body":"foo","handler":"static_response"}]}]}}},"tls":{"certificates":{"automate":["foo.example.com","example.com","www.example.com"]}}}}
```

After (properly matching the hosts, later resulting in Auto HTTPS automating):
```
{"apps":{"http":{"servers":{"srv0":{"listen":[":443"],"routes":[{"match":[{"host":["foo.example.com","example.com","www.example.com"]}],"handle":[{"handler":"subroute","routes":[{"handle":[{"body":"foo","handler":"static_response"}]}]}],"terminal":true}]}}}}}
```

### DIFF
--- a/caddyconfig/caddyfile/parse.go
+++ b/caddyconfig/caddyfile/parse.go
@@ -264,8 +264,13 @@ func (p *parser) addresses() error {
 				return p.Errf("Site addresses cannot contain a comma ',': '%s' - put a space after the comma to separate site addresses", value)
 			}
 
-			token.Text = value
-			p.block.Keys = append(p.block.Keys, token)
+			// After the above, a comma surrounded by spaces would result
+			// in an empty token which we should ignore
+			if value != "" {
+				// Add the token as a site address
+				token.Text = value
+				p.block.Keys = append(p.block.Keys, token)
+			}
 		}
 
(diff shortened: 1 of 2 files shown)


mholt (commented):
Could we do this better by changing `if value[len(value)-1] == ',' {` to `if value[len(value)-1] == ',' && len(strings.TrimSpace(value)) > 1 {` ?

francislavoie:
No because that would set `expectingAnother = false` which would fail to parse the following domain on the same line.

mholt (approved):
Ah okay. This works then.

PR #6621, opened 2024-10-11
reverseproxy: Use correct cases for websocket related headers

WeidiDeng:
Fix https://caddy.community/t/websockets-fail-venus-os-cerbo-gx-victron/25685

Although websocket related headers should be handled case insensitively. Some legacy servers and clients do handle them strictly according to rfc. traefik [handles this case](https://github.com/traefik/traefik/blob/06e64af9e9b4fffd3e0d2772343a31bc0de23c40/pkg/proxy/httputil/proxy.go#L75) too. Popular golang websocket libraries such as [gorilla](https://github.com/gorilla/websocket/blob/5e002381133d322c5f1305d171f3bdd07decf229/client.go#L209) also use the correct case.

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -569,8 +569,9 @@ func (h *Handler) proxyLoopIteration(r *http.Request, origReq *http.Request, w h
 	return false, proxyErr
 }
 
-// only 5 headers to check
-var wsHeaderMapping = map[string]string{
+// Mapping of the canonical form of the headers, to the RFC 6455 form,
+// i.e. `WebSocket` with uppercase 'S'.
+var websocketHeaderMapping = map[string]string{
 	"Sec-Websocket-Accept":     "Sec-WebSocket-Accept",
 	"Sec-Websocket-Extensions": "Sec-WebSocket-Extensions",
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -569,6 +569,23 @@ func (h *Handler) proxyLoopIteration(r *http.Request, origReq *http.Request, w h
...
+const canonicalWsHeaderPrefix = "Sec-Websocket"
+
+// websocket header replacer, normally server doesn't care about this difference,
+// but some do, and RFC use the case too
+// gorilla/websocket, gobwas/ws all use this case too.
+// see https://caddy.community/t/websockets-fail-venus-os-cerbo-gx-victron/25685
+var wsHeaderReplacer = strings.NewReplacer(canonicalWsHeaderPrefix, "Sec-WebSocket")
+
+func normalizeWsHeader(header http.Header) {
+	for k, v := range header {
```
francislavoie, line 581: There's only exactly 5 headers we need to care about. I think if we directly check for those headers, it would perform better, because it would avoid having to do string comparisons on every header field.

francislavoie (approved):
Thanks! Looks perfect to me. I did a commit to adjust the wording a bit.

PR #6573, opened 2024-09-15
Implement issue #6296 passing FDs / socket activation

MayCXC:
third time's the charm ☘️

example Caddyfile:

```
{
	auto_https disable_redirects
	admin off
}

http://localhost {
	bind fd/{env.CADDY_HTTP_FD} {
		protocols h1
	}
	log
	respond "Hello, HTTP!"
}

https://localhost {
	bind fd/{env.CADDY_HTTPS_FD} {
		protocols h1 h2
	}
	bind fdgram/{env.CADDY_HTTP3_FD} {
		protocols h3
	}
	log
	respond "Hello, HTTPS!"
}
```

as seen above, two reserved networks `fd` and `fdgram` have been added. `fd/3` creates a FileListener from fd 3, `fdgram/4` creates a FilePacketConn from fd 4. this simplifies configuration compared to https://github.com/caddyserver/caddy/pull/6543 , but the "actual" host addresses are no longer visible. so, for health checks and the admin interface, fds are treated similarly to unix sockets.

this is a good thing i think, fds should be decoupled from the address they are bound to. separate config options can be added to override the origin address for health checks / the admin interface if needed.

link to issue: https://github.com/caddyserver/caddy/issues/6296

PR #6622, opened 2024-10-12
core: addresses.go funcs renames

MayCXC:
HEAD removes the unreleased ParseNetworkAddressFromHostPort that I added in https://github.com/caddyserver/caddy/pull/6573. I would also like to rename na.JoinHostPort to na.JoinAt, to make it clear that it does not take the same parameters as net.JoinHostPort. finally, I would like to change the signature

`func getListenerFromPlugin(ctx context.Context, network, addr string, config net.ListenConfig) (any, error)`

to

`func getListenerFromPlugin(ctx context.Context, network, host, portRange string, portOffset uint, config net.ListenConfig) (any, error)`

so that plugins are aware of the port range and offset of their network addresses.

Together, these changes are meant to make a clearer interface for plugins to choose either ParseNetworkAddress(JoinNetworkAddress(...)).Listen or net.Listen(ParseNetworkAddress(JoinNetworkAddress(...)).JoinAt(portOffset)) to create their listeners, with preference for the former so they are added to the listener pool. The offset parameter in getListenerFromPlugin also lets https://github.com/MayCXC/caddy-systemd-socket-activation?tab=readme-ov-file use its custom networks to listen to a port range with a systemd named socket unit, which is a very clean integration.


MayCXC:
the last commit widens some test cases for `SplitNetworkAddress`, meaning no incompatibilities with previous versions, but it is new behavior. I think it is a sensible change; caddy already considers the port optional for the argument to SplitNetworkAddress, because they are optional in the config. so now the host is optional, as it is also optional in the config. the result is that network, host, and port can all be returned empty without an err from SplitNetworkAddress, which is what I would expect when parsing the configuration.

so all that is left to put custom network plugins where I want them to be is na.JoinHostPort -> na.JoinAt, and the new getListenerFromPlugin parameters.

MayCXC:
it might also be better to return an error if network, host, and port are all "". but I do think any two of the three should be allowed to be blank.

francislavoie:
You don't need to keep clicking merge btw, we can do that after reviewed and about to be merged.

This looks like a good simplification to me. But I'll let @mholt take a look first.

MayCXC:
> You don't need to keep clicking merge btw, we can do that after reviewed and about to be merged.
> 
> This looks like a good simplification to me. But I'll let @mholt take a look first.

I do that to check if conflicts need to be resolved ahead of time. This PR is also just half the change, to check if changing split is viable.

mholt (approved):
Thanks for the simplification / consistency enhancements!

Issue #6623, opened 2024-10-13
`celFileMatcherMacroExpander()` does not accept "try_files" map entry with a `Comprehension` value


effleurager:
Looking at the documentation for the [file](https://caddyserver.com/docs/caddyfile/matchers#file) matcher, the example CEL expression uses an array of strings to specify a list of files to search for. Seeming like the only alternative to globbing for Windows users, I was surprised to find building a list failed to compile with the following error:
```
provision http.matchers.expression: compiling CEL program: ERROR: <input>:1:6: matcher requires either a map or string literal argument
```

As a toy example, this seems like it should be a valid Caddyfile:
```Caddyfile
domain.tld {
  root * C:/path/to/folder
  @autosave `file({
    "try_files": [0,1,2,3,4,5,7,8,9].map(i, {path} + "_autosave_" + string(i) + ".sav"),
    "try_policy": "most_recently_modified"
  })`
  rewrite @autosave {file_match.relative}
  file_server
}
```

Given the severe restriction of globbing being unavailable, this seems like an important fallback to have?

francislavoie:
Due to how matchers are provisioned (once at config load time, not on every request), they need to have static values as input (except for placeholders). I'm not sure it would be possible to implement this. From a type system standpoint I don't see how the MatchFile struct could take a dynamic try_files input fed by CEL.

effleurager:
While testing, I discovered a bug with the `try_policy` value not being assigned:
https://github.com/caddyserver/caddy/blob/48ce47f1d44da485fbbf6be536a0e3822763f313/modules/caddyhttp/fileserver/matcher.go#L192-L195

I'm preparing a PR, but I'm wondering if I should add a test for `CELLibrary` specifically or just add an addition to `expressionTests`?
https://github.com/caddyserver/caddy/blob/48ce47f1d44da485fbbf6be536a0e3822763f313/modules/caddyhttp/fileserver/matcher_test.go#L283-L292

francislavoie:
Oops! Nice spotting. Yeah any kind of tests you can add are appreciated 😅

PR #6624, opened 2024-10-13
fileserver: fix try_policy when instantiating file matcher from CEL

effleurager:
I found a minor bug in #6623 for the `file` matcher when written as a CEL expression.

This PR corrects the current reassignment of `root`, and adds a test to ensure the correct property is set.


francislavoie:
Thanks!

I'm not sure that test properly covers it though? Wouldn't it need more than one try_files at least to tell them apart with that policy? I figure it could be tricky tho because git might make those test files have the same mod time though 🤔 not sure how to best test that.

francislavoie:
I reworked the test to use `largest_size` which we can guarantee when using Git (since file mod time is based on when you checkout the changes etc), and validating that it does indeed pick the correct file by checking the value in the replacer.

francislavoie (approved):

Issue #6628, opened 2024-10-15
[bug] cannot deny transfer-encoding requests


shyim:
Hey,

I am trying to block `Transfer-Encoding: chunked` requests. But this seems not possible right now in Caddy:

I tried following:

```
@length {
   header Transfer-Encoding chunked
}

respond @length 411
```

I think the header `Transfer-Encoding` is missing in the available headers: as the logs does not mention it too:

```
{"level":"info","ts":1729001178.8945065,"logger":"http.log.access","msg":"NOP","request":{"remote_ip":"192.168.117.1","remote_port":"49353","client_ip":"192.168.117.1","proto":"HTTP/1.1","method":"POST","host":"127.0.0.1:8000","uri":"/","headers":{"Content-Type":["application/json"],"User-Agent":["curl/8.7.1"],"Accept":["*/*"]}},"bytes_read":0,"user_id":"","duration":0.000008292,"size":0,"status":0,"resp_headers":{"Server":["Caddy"]}}
```

When i change the header to `X-Foo` everything works

francislavoie:
Interestingly, Go doesn't keep `Transfer-Encoding` in the headers map, it moves it to a separate field on the request: https://pkg.go.dev/net/http#Request.TransferEncoding

We have special handling for the `header` matcher for `Host` which is also moved separately, we could probably do the same for `TransferEncoding`. Would make sense to allow matching it I think.

steffenbusch:
@francislavoie Wouldn't it make sense to add the special handling for the `Transfer-Encoding' header also to the json logging area to have `Transfer-Encoding' be logged in `request>headers>Transfer-Encoding[]`?

francislavoie:
Hm, maybe yeah. Lemme take a look.

shyim:
It should be here right? https://github.com/caddyserver/caddy/blob/a211c656f12bcab73df0de114f2b6100ee5a0fe4/modules/caddyhttp/matchers.go#L972

WeidiDeng:
@shyim Simply matching by `transfer-encoding` isn't enough, http2/3 doesn't have this header. The problem here is that the content-length is unknown in advance, the php-fpm needs this info to proceed.

I'll retry to figure out how to buffer requests in this case. Might take a few day.

dallyger:
[...]

PR #6629, opened 2024-10-15
caddyhttp: Allow matching Transfer-Encoding, add to access logs

francislavoie:
Fixes https://github.com/caddyserver/caddy/issues/6628


modules/caddyhttp/matchers.go
```diff
@@ -967,22 +967,26 @@ func (MatchHeader) CELLibrary(_ caddy.Context) (cel.Library, error) {
...
-func getHeaderFieldVals(input http.Header, fieldName, host string) []string {
+// The host parameter should be obtained from the http.Request.Host field, and the
+// transferEncoding from http.Request.TransferEncoding, since net/http removes them
+// from the header map.
+func getHeaderFieldVals(input http.Header, fieldName, host string, transferEncoding []string) []string {
 	fieldName = textproto.CanonicalMIMEHeaderKey(fieldName)
 	if fieldName == "Host" && host != "" {
 		return []string{host}
 	}
+	if fieldName == "Transfer-Encoding" && input[fieldName] == nil {
```
mohammed90, line 978: Is the check `input[fieldName] == nil` necessary if we know Go stdlib always moves the `Transfer-Encoding` header out of the header map?

francislavoie: If some `header` directive set it for some reason, it could be set I guess.

mohammed90: Yeah, that's possible, I guess.

mohammed90 (approved):
LGTM

mholt (approved):
Thanks for the contribution!

PR #6636, opened 2024-10-17
httpcaddyfile: Fixes for `prefer_wildcard` mode

francislavoie:
Followup to #6146

Two fixes:

- When assembling the HTTP app, the wildcard hosts need to be collected first, then considered after, because there's no guarantee that all non-wildcards will appear after all wildcards when looping. Also we should not add a domain to Skip if it doesn't qualify for TLS anyway. But I realized we should actually add it to SkipCerts, not Skip because we do want them to still get HTTP->HTTPS redirects, just not have certs issued.

- The automation policy consolidation misbehaved if there was more than one wildcard configured, because it was comparing wildcards against eachother. This would cause all APs to disappear in some cases. Instead of handling wildcard coverage in consolidation, I reworked it to avoid adding the AP altogether if it would be covered by a wildcard. Should be more robust.

### DIFF
--- a/caddyconfig/httpcaddyfile/httptype.go
+++ b/caddyconfig/httpcaddyfile/httptype.go
@@ -827,7 +827,7 @@ func (st *ServerType) serversFromPairings(
 						baseDomain = baseDomain[idx+1:]
 					}
 					if !strings.HasPrefix(addr.Host, "*.") && slices.Contains(wildcardHosts, baseDomain) {
-						srv.AutoHTTPS.Skip = append(srv.AutoHTTPS.Skip, addr.Host)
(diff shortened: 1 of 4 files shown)


coandco:
I built with this branch and can confirm that the problem is now fixed, at least for me.  I can define multiple `*.one.example.com`/`*.two.example.com` wildcard domains and it handles them all as it should.

mholt:
Will look at this again soon!

coandco:
Looking forward to being able to build with an official beta, rather than just pointing at this branch!

mholt (approved):
Thanks for working on this, Francis. Sorry for my delay!

mholt:
This will go out with the next beta in a few days or less

mholt:
@coandco This has been released in the beta a few days ago just FYI

PR #6638, opened 2024-10-17
reverseproxy: Set `Content-Length` when body is fully buffered

WeidiDeng:
Remedy for [this issue](https://github.com/caddyserver/caddy/issues/5420#issuecomment-2415943516).


mholt (approved):
I think this makes sense. LGTY, @francislavoie ?

Thanks Weidi :)

francislavoie:
I don't totally follow the issue at hand so no comment :sweat_smile:

Issue #6637, opened 2024-10-17
reverse_proxy: problems with fastcgi requests with unknown body length


WeidiDeng:
Right now, caddy has problems handling fastcgi requests when the client sends a body with unknown body length, for http1.1 it's usually chunked encoded body. Users have to buffer the request manually and the buffer doesn't work right now because `Content-length` is not set in this case, and it will leads to some easy attacks.

[Here](https://github.com/caddyserver/caddy/issues/5420#issuecomment-2415943516), an attacker doesn't need to do much to block all processes spawned by php-fpm.

Caddy should handle this type of requests with a sane default just like nginx does, i.e. buffer requests with unknown body length in this case, and reject requests whose body is too large to buffer.

WeidiDeng:
With [6638](https://github.com/caddyserver/caddy/pull/6638), that specific issue should be gone. However, for request bodies larger than whatever is defined will still cause problems, unless `-1` is defined which means all the request body is buffered in memory, which can be used in attacks as well.

I checked nginx configurations, [request](https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_request_buffering) and [response](https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_buffering) buffering is on by default. But let's just buffer those requests with unknown content length for now.

WeidiDeng:
Done, it's now available in [this patch](https://github.com/caddyserver/caddy/pull/6639). Though the maximum request body size can't be limited there, it's available in [request_body](https://caddyserver.com/docs/caddyfile/directives/request_body). Set that if some clients send some arbitrary large body to mount a ddos attack.

cottonthread:
I'm experiencing the same issue, and it seems that many others are also affected by this. I'm not a development software engineer, so I'm not sure if I can ask you to show me how to temporarily mitigate this type of error. Currently, I have the following configuration in my Caddyfile, but it doesn't seem to work. I'm using Caddy through Docker, pulling the image directly from Docker Hub. Thank you very much!

```
(blockphpbug) {
	@blockChunkedEncoding {
		header Transfer-Encoding chunked
	}
	respond @blockChunkedEncoding 411
	request_body {
		max_size 1MB
	}
}
(passtophp) {
	php_fastcgi php:9000 {
		request_buffers 4k
		response_buffers 4k
	}
}

```

dallyger:
[...]

PR #6639, opened 2024-10-17
reverse_proxy: rewrite request buffering for fastcgi

WeidiDeng:
Follow up on [6637](https://github.com/caddyserver/caddy/issues/6637). Now this patch will try to buffer request body for fastcgi requests by default with memory and file based buffers. The buffering behavior can be disabled and the buffer limits can be set as well.

This makes sure caddy can work out of the box when the requests don't have a content length, typically chunked encoding, without having to configure request buffering, which will buffer all requests and will not set `content-length` if the request can't be buffered in memory. `content-length` is required when communicating with php-fpm.

If buffering is disabled, response for requests without `content-length` will be `411 Length Required`, typically for http1 requests that have `chunked` encoding. If the buffer limit is reached, the response will be `413 Request Entity Too Large`.

By default, memory buffer is `16k` per request, and all file based buffers can only use `100MB` of disk space. The buffer file path is dependent on the system, and caddy wil fail to start if caddy can't create files in this directory.

The new configuration parameters are following:

```
body_buffer_disabled
body_buffer_memory_limit <size>
file_buffer_size_limit  <size>
file_buffer_filepath <path>
```

Another fix is that now `content-length` is checked when sending the request to php-fpm, so that request size is limited to prevent client to send faulty `content-length`.

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
@@ -19,6 +19,8 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/dustin/go-humanize"
+
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
@@ -43,7 +45,10 @@ func init() {
 //	    dial_timeout <duration>
 //	    read_timeout <duration>
 //	    write_timeout <duration>
-//	    capture_stderr
+//	    body_buffer_disabled
+//	    body_buffer_memory_limit <size>
+//	    file_buffer_size_limit <size>
+//	    file_buffer_filepath <path>
 //	}
 func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 	d.Next() // consume transport name
@@ -112,6 +117,35 @@ func (t *Transport) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				return d.ArgErr()
 			}
 			t.CaptureStderr = true
+
+		case "body_buffer_disabled":
+			t.BodyBufferDisabled = true
+
+		case "body_buffer_memory_limit":
+			if !d.NextArg() {
+				return d.ArgErr()
+			}
+			size, err := humanize.ParseBytes(d.Val())
+			if err != nil {
+				return d.Errf("bad buffer size %s: %v", d.Val(), err)
+			}
+			t.BodyBufferMemoryLimit = int64(size)
+
+		case "file_buffer_size_limit":
+			if !d.NextArg() {
+				return d.ArgErr()
+			}
+			size, err := humanize.ParseBytes(d.Val())
+			if err != nil {
+				return d.Errf("bad buffer size %s: %v", d.Val(), err)
+			}
+			t.FileBufferSizeLimit = int64(size)
+
+		case "file_buffer_filepath":
+			if !d.NextArg() {
+				return d.ArgErr()
+			}
+			t.FileBufferFilepath = d.Val()
 
 		default:
 			return d.Errf("unrecognized subdirective %s", d.Val())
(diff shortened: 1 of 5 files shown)


modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -43,7 +45,10 @@ func init() {
 //	    dial_timeout <duration>
 //	    read_timeout <duration>
 //	    write_timeout <duration>
-//	    capture_stderr
```
francislavoie, line 46: I don't think you meant to remove this

modules/caddyhttp/reverseproxy/fastcgi/client.go
```diff
@@ -197,9 +203,68 @@ func (f clientCloser) Close() error {
...
+		defer closer.Close()
+	}
+
+	// check for content_length and buffer the request if needed
+	cl, ok := checkContentLength(p)
+	if !ok {
+		// buffering disabled
+		if !c.buffer {
+			c.rwc.Close()
+			return newErrorResponse(http.StatusLengthRequired), nil
```
francislavoie, line 250: Hmm, would it be better to throw a caddyhttp.Error instead? That way it could be caught by `handle_errors` to render an error page etc. Or, at least an error type internal to this client that the transport.go could check for and convert to a caddyhttp.Error if we don't want the client.go to depend on anything in caddy.

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
```diff
@@ -80,8 +84,18 @@ type Transport struct {
...
-	serverSoftware string
-	logger         *zap.Logger
+	// disable buffering of the request body that doesn't have a content length
+	BodyBufferDisabled bool `json:"body_buffer_disabled,omitempty"`
+	// memory limit for buffering the request body, the rest will be buffered by temporary files
+	BodyBufferMemoryLimit int64 `json:"body_buffer_memory_limit,omitempty"`
+	// total disk storage allowed by the request body buffer
+	FileBufferSizeLimit int64 `json:"file_buffer_size_limit,omitempty"`
+	// the path to store the temporary files for the request body buffer
+	FileBufferFilepath string `json:"file_buffer_filepath,omitempty"`
```
francislavoie, line 94: Should it be called "dir" instead? Cause this isn't exactly the "file" path since the filename is generated

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
```diff
@@ -106,12 +129,144 @@ func (t *Transport) Provision(ctx caddy.Context) error {
...
+	}
+
+	if !t.BodyBufferDisabled {
+		if t.FileBufferFilepath == "" {
+			t.FileBufferFilepath = os.TempDir()
+		}
+		// test if temporary file can be created
+		file, err := os.CreateTemp(t.FileBufferFilepath, "caddy-fastcgi-buffer-")
+		if err != nil {
+			return fmt.Errorf("failed to create temporary file: %v", err)
```
francislavoie, line 142: Hmm. Creating a file and throwing here could mean Caddy fails to start (or config reload). I wonder if that's a good thing or not. Maybe this test needs to happen at runtime instead, or maybe it should happen in `Validate()` instead (still happens at startup but separated out)

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -43,7 +45,10 @@ func init() {
...
+//	    body_buffer_disabled
+//	    body_buffer_memory_limit <size>
+//	    file_buffer_size_limit <size>
+//	    file_buffer_filepath <path>
```
francislavoie, line 51: It's ambiguous whether these configure the request or response body buffer.

modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go
```diff
@@ -92,6 +106,15 @@ func (Transport) CaddyModule() caddy.ModuleInfo {
 	}
 }
 
+const (
+	defaultDialTimeout = 3 * time.Second
+	// nginx default for 64bit platforms
+	// https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size
+	defaultMemBufferSize = 1 << 14 // 16KB
+	// nginx doesn't have an option to limit the total file buffer size
+	defaultFileBufferSize = 100 << 20 // 100MB
```
francislavoie, line 115: It's good to have a default limit, but my first thought is could an evil client abuse this to do a storage exhaustion attack? They could open multiple requests and send garbage data in each of them but then "never finish" which would keep the storage locked up in each request. Say you only have 2GB of storage (reasonable amount for a VPS) then they'd only need to make like 20 requests that eat up 99MB to exhaust the storage. That could mean things like cert issuance fail to write to disk etc.

We'll need to think about how to mitigate things like that (slow loris style attacks). Maybe we need like a minimum bandwidth check or something like that which cancels the request if the speed of reading drops below a certain rate (1kb/s? Are there any legitimate requests that can dip that low?) Sounds pretty tricky to implement though.

francislavoie:
Also, longer term we should think about generalizing the buffering so it could be used by any reverse_proxy transport. There are cases where it could be useful for some users, like being able to retry a POST on a different upstream (rewind and resend the body) or to tee it to two upstreams at once (e.g. to send it to a new version of an upstream to compare the result with the old one -- not sure if buffering is needed for that but it could help?)

trea:
Thank you for getting a proper fix put together for this @WeidiDeng! ❤️

mholt:
Is this superseded by #6661? If so, maybe we can close this.

(Thanks for all your help in fixing the problem, Weidi!)

trea:
It might be slightly more difficult to encourage Caddy use in the PHP space if buffering is completely off the table. The standard bearer reverse proxy in that space _does_ buffer to the filesystem for FastCGI by default. But, I can also appreciate that buffering is much more complicated.

francislavoie:
@WeidiDeng any interest in rebasing this and finishing it up? Do you think we still need it?

PR #6645, opened 2024-10-19
fix: always ignore SIGPIPE

dunglas:
Backport of https://github.com/dunglas/frankenphp/pull/1101 to Caddy itself.

Sometimes (especially when using systemd), `SIGPIPE` signals can crash the server. This patch always ignore SIGPIPE, for all file descriptors.


mholt (approved):
LGTM. Thanks for the patch!

Issue #6644, opened 2024-10-18
Increasing the limit for file_server


shock0572:
We're currently using Caddy to browse and search through its file server interface a series of debugging symbols.
Each one of the directories is over 13000 hashes per symbols and we're hitting this limitation:

https://github.com/caddyserver/caddy/blob/2cac3c5491e6428441ecf668cc4f5a86e67ed9b3/modules/caddyhttp/fileserver/browse.go#L148

would it be possible to increase it, maybe in the `file_server` directive, as an attribute of browse? (`max_limit`?)

```go
file_server {
    browse max_limit BIG_NUMBER
}
```

I know that they're reachable and they can be downloaded, but it would be cool if we could set that limit on our own

francislavoie:
PRs welcome to make it configurable, should be trivial to implement.

PR #6648, opened 2024-10-20
fileserver: Add `file_limit` option for browse

atakanyenel:
This PR fixes the issue #6644 . I'm not sure about the config name `file_limit`, I prefer `max_dir_limit` etc.The default is still 10000.

As mentioned in the issue, this PR gave me some background on file serve, so I will try to look at paging as well. Please feel free to edit any documentation as you see fit.

### DIFF
--- a/modules/caddyhttp/fileserver/browse.go
+++ b/modules/caddyhttp/fileserver/browse.go
@@ -66,7 +66,14 @@ type Browse struct {
 	//   - `sort size` will sort by size in ascending order
 	// The first option must be `sort_by` and the second option must be `order` (if exists).
 	SortOptions []string `json:"sort,omitempty"`
+
+	// FileLimit limits the number of up to n DirEntry values in directory order.
+	FileLimit int `json:"file_limit,omitempty"`
 }
+
+const (
+	defaultMaxDirLimit = 10000
+)
 
 func (fsrv *FileServer) serveBrowse(fileSystem fs.FS, root, dirPath string, w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
 	if c := fsrv.logger.Check(zapcore.DebugLevel, "browse enabled; listing directory contents"); c != nil {
@@ -206,7 +213,11 @@ func (fsrv *FileServer) serveBrowse(fileSystem fs.FS, root, dirPath string, w ht
 }
 
 func (fsrv *FileServer) loadDirectoryContents(ctx context.Context, fileSystem fs.FS, dir fs.ReadDirFile, root, urlPath string, repl *caddy.Replacer) (*browseTemplateContext, error) {
-	files, err := dir.ReadDir(10000) // TODO: this limit should probably be configurable
+	dirLimit := defaultMaxDirLimit
+	if fsrv.Browse.FileLimit != 0 {
+		dirLimit = fsrv.Browse.FileLimit
+	}
+	files, err := dir.ReadDir(dirLimit)
 	if err != nil && err != io.EOF {
 		return nil, err
 	}
--- a/modules/caddyhttp/fileserver/browse.go
+++ b/modules/caddyhttp/fileserver/browse.go
@@ -72,7 +72,7 @@ type Browse struct {
 }
 
 const (
-	defaultMaxDirLimit = 10000
+	defaultDirEntryLimit = 10000
 )
 
 func (fsrv *FileServer) serveBrowse(fileSystem fs.FS, root, dirPath string, w http.ResponseWriter, r *http.Request, next caddyhttp.Handler) error {
@@ -213,7 +213,7 @@ func (fsrv *FileServer) serveBrowse(fileSystem fs.FS, root, dirPath string, w ht
 }
 
 func (fsrv *FileServer) loadDirectoryContents(ctx context.Context, fileSystem fs.FS, dir fs.ReadDirFile, root, urlPath string, repl *caddy.Replacer) (*browseTemplateContext, error) {
-	dirLimit := defaultMaxDirLimit
+	dirLimit := defaultDirEntryLimit
 	if fsrv.Browse.FileLimit != 0 {
 		dirLimit = fsrv.Browse.FileLimit
 	}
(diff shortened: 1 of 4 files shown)


modules/caddyhttp/fileserver/browse.go
```diff
@@ -68,6 +68,10 @@ type Browse struct {
 	SortOptions []string `json:"sort,omitempty"`
 }
 
+var (
+	maxDirLimit = 10000
+)
```
francislavoie, line 73: This should be a constant, not a variable. Should have a prefix like `default`

modules/caddyhttp/fileserver/browse.go
```diff
@@ -206,7 +210,10 @@ func (fsrv *FileServer) serveBrowse(fileSystem fs.FS, root, dirPath string, w ht
 }
 
 func (fsrv *FileServer) loadDirectoryContents(ctx context.Context, fileSystem fs.FS, dir fs.ReadDirFile, root, urlPath string, repl *caddy.Replacer) (*browseTemplateContext, error) {
-	files, err := dir.ReadDir(10000) // TODO: this limit should probably be configurable
+	if fsrv.FileLimit != 0 {
+		maxDirLimit = fsrv.FileLimit
+	}
```
francislavoie, line 215: Don't set the value of a global variable like this. `file_server` may be loaded more than once with different config, so they would clobber eachother. Make sure to use a local variable here to grab the default value, then override it with the configured one if non-zero.

modules/caddyhttp/fileserver/caddyfile.go
```diff
@@ -59,6 +60,7 @@ func parseCaddyfile(h httpcaddyfile.Helper) (caddyhttp.MiddlewareHandler, error)
 //	    precompressed <formats...>
 //	    status        <status>
 //	    disable_canonical_uris
+//		file_limit	  <limit>
```
francislavoie, line 63: It's strange, but gofmt tends to format code comments this way.
```suggestion
//	    file_limit      <limit>
```

modules/caddyhttp/fileserver/browse.go
atakanyenel: Done.

modules/caddyhttp/fileserver/browse.go
atakanyenel: Done.

modules/caddyhttp/fileserver/caddyfile.go
atakanyenel: Yes, strange... I ran `go fmt` locally and it looks okay.

modules/caddyhttp/fileserver/staticfiles.go
```diff
@@ -172,6 +172,9 @@ type FileServer struct {
 	fsmap caddy.FileSystems
 
 	logger *zap.Logger
+
+	// FileLimit limits the number of up to n DirEntry values in directory order.
+	FileLimit int `json:"file_limit,omitempty"`
```
francislavoie, line 177: This should go inside the `Browse` struct, since it's specifically an option for the browse feature. Look at the `reveal_symlinks` option and how that's wired up, it should be a sibling of that.

francislavoie (changes requested):

modules/caddyhttp/fileserver/staticfiles.go
atakanyenel: Updated.

modules/caddyhttp/fileserver/caddyfile.go
atakanyenel: This was a space / tab mismatch. I fixed it now.

modules/caddyhttp/fileserver/caddyfile.go
```diff
@@ -168,6 +170,17 @@ func (fsrv *FileServer) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 			falseBool := false
 			fsrv.CanonicalURIs = &falseBool
 
+		case "file_limit":
```
francislavoie, line 173: This needs to be moved to line 133-ish to be inside of `browse`

atakanyenel: Oh yes,  I misread your config and put it under the first browse. Updating now.

francislavoie (approved):
Thanks!

For bonus points, it could probably use an [adapt test](https://github.com/caddyserver/caddy/tree/master/caddytest/integration/caddyfile_adapt) (can just add it to an existing `file_server` test) to make sure it produces the correct JSON.

mholt:
Thanks! This LGTM, but before I merge I have just one stupid nit: should it be called `DirEntryLimit`? It's a little more specific (could `FileLimit` be limiting file sizes? Something else?).

Similarly, `defaultMaxDirLimit` could be renamed `defaultDirEntryLimit`.

atakanyenel:
I added the test and updated the constant name. I'm keeping the config name still as "file_limit". I can also change that if required.

modules/caddyhttp/fileserver/browse.go
```diff
@@ -66,8 +66,15 @@ type Browse struct {
 	//   - `sort size` will sort by size in ascending order
 	// The first option must be `sort_by` and the second option must be `order` (if exists).
 	SortOptions []string `json:"sort,omitempty"`
+
+	// FileLimit limits the number of up to n DirEntry values in directory order.
```
mholt, line 70: Should we document that -1 means no limit? (And 0 is default limit, which is currently 10,000)

mholt (approved):
I'm going to merge this, but I still think we should enhance the comment/docs before the final release; and maybe even mark this option as experimental since we don't quite understand its characteristics in the wild yet. Would you be able to help with that? I just don't have a chance right now. Thanks!

atakanyenel:
> I'm going to merge this, but I still think we should enhance the comment/docs before the final release; and maybe even mark this option as experimental since we don't quite understand its characteristics in the wild yet. Would you be able to help with that? I just don't have a chance right now. Thanks!

Yes, I can do this. I will send another PR soon. Thanks for the merge.

Issue #6613, opened 2024-10-08
Remove MLP-2.0 dependency


sbruens:
The embedded ACME server is pulling in a dependency that has an upstream dependency on `github.com/go-sql-driver/mysql`, which is licensed under MPL2.0. This makes packaging for downstream users difficult and may have other licensing incompatibilities:

```sh
$ go mod why github.com/go-sql-driver/mysql
# github.com/go-sql-driver/mysql
github.com/caddyserver/caddy/v2/modules/caddypki/acmeserver
github.com/smallstep/nosql
github.com/smallstep/nosql/mysql
github.com/go-sql-driver/mysql
```

I've already filed an [issue](https://github.com/smallstep/nosql/issues/68) to the upstream library that depends on it directly, but looking at how that library is used in Caddy, it might be easier to remove the dependency here. As far as I can tell, the only use is here in a type assertion: 

https://github.com/caddyserver/caddy/blob/d7564d632fbed209e81978c5c2c529a7bf1836f7/modules/caddypki/acmeserver/acmeserver.go#L192

Could the dependency be removed and replaced with a mirrored interface, or another solution?

francislavoie:
What's the incompatibility exactly? From my reading, there's no violation here. From the summary at https://github.com/go-sql-driver/mysql/blob/master/LICENSE, I don't see the issue.

hslatman:
@sbruens it's not just the type assertion that is used: the call to `ca.NewAuthority` will instantiate a concrete instance of `nosql.DB`. That instance may, or may not, be a `nosql/mysql` instance; based on the DSN. As far as I know at this time Caddy will actually only use the `bbolt` implementation, so it never calls into the `go-sql-driver/mysql` package.

Like @francislavoie, I don't immediately see a big problem with this particular license in this case. 

There is some work and discussion going on in https://github.com/caddyserver/caddy/pull/6097 to support more database types, and in https://github.com/smallstep/nosql/pull/53 to rearchitect the implementation (and dependencies) of the `nosql` package. 

We'll discuss https://github.com/smallstep/nosql/issues/68 in our upcoming open source triage meeting.

sbruens:
MPL-2.0 is not a permissive license like Apache-2.0, but a "weak copyleft" license. They are not necessarily incompatible, but anyone distributing software with MPL-2.0 components needs to make the source code for those parts available to end users. This creates additional obligations for users of Caddy that are more cumbersome than if all the code was under a more permissive license.

[...]

PR #6655, opened 2024-10-22
chore: compile without nosql's support for Postgres and MySQL

dunglas:
I had issues with nosql that only supports old versions of Badger while upgrading [Mercure](https://mercure.rocks) dependencies: https://github.com/dunglas/mercure/pull/960. While working on this, I figured out that we could disable compilation of Badger support for nosql (already done by Caddy), but also of MySQL and Postgres.

This reduce the build time and the binary size.
This also has the nice side effect of fixing https://github.com/caddyserver/caddy/issues/6613 because this dependency will not be included in the binary at all anymore.

When https://github.com/caddyserver/caddy/pull/6097 will be finished, we may need to add back support for MySQL and Postgres, but that's not sure because the new version of nosql may provide adapters in separate modules, that users will cherry-pick (that would be the best option IMHO). 

Closes https://github.com/caddyserver/caddy/issues/6613.

### DIFF
--- a/.github/workflows/cross-build.yml
+++ b/.github/workflows/cross-build.yml
@@ -70,4 +70,4 @@ jobs:
         continue-on-error: true
         working-directory: ./cmd/caddy
         run: |
-          GOOS=$GOOS GOARCH=$GOARCH go build -tags nobadger -trimpath -o caddy-"$GOOS"-$GOARCH 2> /dev/null
+          GOOS=$GOOS GOARCH=$GOARCH go build -tags tags nobadger,nomysql,nopgx -trimpath -o caddy-"$GOOS"-$GOARCH 2> /dev/null
--- a/.github/workflows/cross-build.yml
+++ b/.github/workflows/cross-build.yml
@@ -70,4 +70,4 @@ jobs:
         continue-on-error: true
         working-directory: ./cmd/caddy
         run: |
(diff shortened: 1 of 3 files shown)


francislavoie:
Yeah I agree we should do this for now. I really don't like that it's necessary. I hope Smallstep modularizes it sooner rather than later so this isn't necessary. (FYI @hslatman @azazeal, just so you know that it also affects Caddy's plugin authors as well)

mohammed90:
I don't mind this at the moment, until upstream refactoring is finalized and the #6097 is complete sometime in the future.

mholt (approved):

francislavoie:
Looking at the CI artifacts, it doesn't seem like the build output reduced in size at all (or at least is within 100kb difference).

francislavoie:
Hmm, tried building locally with the tags and:

```
$ go build -tags=nobadger,nomysql,nopgx                                           
$ ls -la caddy
56650655

$ go build -tags=nobadger
$ ls -la caddy                                                         
59840067
```

So it seems to be about 3MB smaller actually. Maybe I'm being misled by these: (before: https://github.com/caddyserver/caddy/actions/runs/11467177533, after: https://github.com/caddyserver/caddy/actions/runs/11468346497)

PR #6656, opened 2024-10-22
reverseproxy: Sync changes from stdlib for 1xx handling

francislavoie:
Sourced from https://github.com/golang/go/commit/960654be0c4ad7918376e2e1d47491c9bc7520e0

Also usage of `clear(h)` from https://github.com/golang/go/commit/3bc28402fae2a1646e4d2756344b5eb34994d25f


dunglas (approved):

mholt (approved):
Neat, didn't know about `clear()`

Interesting bug as well, thanks for the patch! Hopefully it fixes  the linked issues.

PR #6660, opened 2024-10-24
On demand path

jrouzierinverse:


francislavoie:
This is gonna need an explanation.

jrouzierinverse:
This is to allow you to use the file system for permission if a site could request a certificate.

```
    on_demand_tls {
        permission   path   /usr/share/www/
    }
```

If a request for example.com comes in this will check for the existence of a file or directory `/usr/share/www/example.com` if a certificate could be requested from an acme service.

An example config if you want to serve a static site based off directories.
```
{
    on_demand_tls {
        permission   path   /usr/share/www/
    }
}

https:// {
    tls {
        on_demand
    }
    root * /usr/share/www/{host}
    file_server
}

```

francislavoie:
Not rejecting the idea, but FYI it can be done with the `http` module like this:

```caddyfile
{
	on_demand_tls {
		ask http://localhost:9001/
	}
}

https:// {
	tls {
		on_demand
	}
	root * /usr/share/www/{host}
	file_server
}

:9001 {
	root * /usr/share/www/
	@domain file {query.domain}
	respond @domain 200
	respond 404
}
```

jrouzierinverse:
That is cool

mholt:
Thanks for the PR! I can see how this could be useful to some people, I'm just not convinced as of yet that it is generally useful to the greater user based. I would suggest publishing this plugin in your own repository and then listing it on the Caddy website so others will see it and use it, and if it becomes so applicable that many people are using it, we can consider bringing it into the core. :+1:

Looks cool for sure. Publish it and let's see it get used!

PR #6661, opened 2024-10-24
fastcgi: check for CONTENT_LENGTH when sending requests

WeidiDeng:
For #6637, I think previous [pr](https://github.com/caddyserver/caddy/pull/6639) is a bit too ambitious. Requests buffering is hard to do right without considering a multitude of factors. This patch only requires `CONTENT_LENGTH` to be set for fastcgi requests.


francislavoie:
So we just throw `411` for any chunked requests? Makes sense that it'll prevent hanging. But long term, shouldn't we have support for streaming chunked requests?

WeidiDeng:
php-fpm doesn't support streaming chunked requests. That's the thing. Golang fastcgi supports streaming chunked requests, but does anyone use that with caddy?

francislavoie:
https://bz.apache.org/bugzilla/show_bug.cgi?id=57087 looks like Apache fixed their fpm support with buffering

lutoma:
Are there any updates on this (or the other patch for the issue)? Seeing these PRs just lying around for a month is pretty discouraging when this is essentially a DoS that seems to affect a lot of people

mohammed90:
> Are there any updates on this (or the other patch for the issue)? Seeing these PRs just lying around for a month is pretty discouraging when this is essentially a DoS that seems to affect a lot of people

The patch only been up for 3 weeks, and it's under review. We've been releasing 2.9.0 betas. You can build Caddy with this patch using xcaddy.

mholt (approved):
Thanks Weidi. I like the more conservative approach. Maybe in the future we can figure out how to fix it with buffering to disk, but for now, this will do. :) Really appreciate it!

modules/caddyhttp/reverseproxy/fastcgi/client.go
```diff
@@ -136,6 +138,18 @@ type client struct {
...
+	if clStr, ok := p["CONTENT_LENGTH"]; !ok {
+		return nil, caddyhttp.Error(http.StatusLengthRequired, nil)
+	} else {
+		cl, err := strconv.ParseInt(clStr, 10, 64)
+		// stdlib won't return a negative Content-Length, but we check just in case,
+		// the most likely cause is from a missing content length, which is -1
+		if err != nil || cl < 0 {
+			return nil, caddyhttp.Error(http.StatusLengthRequired, err)
+		}
+	}
```
dunglas, line 151:
```suggestion
	}

	cl, err := strconv.ParseInt(clStr, 10, 64)
	// stdlib won't return a negative Content-Length, but we check just in case,
	// the most likely cause is from a missing content length, which is -1
	if err != nil || cl < 0 {
		return nil, caddyhttp.Error(http.StatusLengthRequired, err)
	}
```

(nit: the `else` is not necessary here)

modules/caddyhttp/reverseproxy/fastcgi/client.go
```diff
@@ -136,6 +138,18 @@ type client struct {
...
+	// check for CONTENT_LENGTH, since the lack of it or wrong value will cause the backend to hang
+	if clStr, ok := p["CONTENT_LENGTH"]; !ok {
+		return nil, caddyhttp.Error(http.StatusLengthRequired, nil)
+	} else {
+		cl, err := strconv.ParseInt(clStr, 10, 64)
+		// stdlib won't return a negative Content-Length, but we check just in case,
+		// the most likely cause is from a missing content length, which is -1
+		if err != nil || cl < 0 {
+			return nil, caddyhttp.Error(http.StatusLengthRequired, err)
+		}
```
dunglas, line 150:
```suggestion
		cl, err := strconv.ParseUint(clStr, 10, 64)
		// stdlib won't return a negative Content-Length, but we check just in case,
		// the most likely cause is from a missing content length, which is -1
		if err != nil {
			return nil, caddyhttp.Error(http.StatusLengthRequired, err)
		}
```

With `ParseUint()` negative numbers will return an error.

dunglas (changes requested):

modules/caddyhttp/reverseproxy/fastcgi/client.go
dunglas, line 151:
```suggestion
	} else if _, err := strconv.ParseUint(clStr, 10, 64); err != nil {
		// stdlib won't return a negative Content-Length, but we check just in case,
		// the most likely cause is from a missing content length, which is -1
		return nil, caddyhttp.Error(http.StatusLengthRequired, err)
	}
```

mholt: Nice idea! I'll be ready to merge this if @WeidiDeng agrees with the suggestion and accepts it

PR #6662, opened 2024-10-24
rewrite: Don't add `/` in Caddyfile, do it after replacer

francislavoie:
Context: https://caddy.community/t/unable-to-use-map-variables-with-url-strip-prefix/26093/2

The problem is if the input to the Caddyfile is a placeholder whose value will start with a `/` (e.g. `{foo} = /foo`), then it would cause `{foo}` to become `/{foo}` after the adapter, then later `//foo` once replaced.

We should add the `/` if missing at runtime after running the replacer, if necessary.


mholt (approved):

mholt:
Thank you!

Issue #109, opened 2015-06-06
Authentication ,authorization(session management ), Caddy configuration for restricted routes 


hackintoshrao:
Here are set of features which could make authentication and authorization pieces easier for developers 

## Social Login 
Enabling social login for easy authentication 
Caddy configuration could be something like this 

```

login facebook  {
    facebook_api_key        FB APP ID
    facebook_api_secret     FB API SECRET
    login_page_redirect     http://localhost:3000/auth/facebook
    callback_url            http://localhost:3000/auth/facebook/callback
    host                    localhost
    sucess_redirect         /
   failure_redirect         /login
}

```
After login sessions management has to be made easy , the login middle should force implementation of a User struct which will be  directly injected as session data , the session management tool like Gorilla will unable  the usage of wide variety of storage options for sessions 

## Restricted routes made easy 
Once authentication is provided it should become easier for developers to identify whether the a given request is authenticated and graceful handling of authenticated and un-authenticated cases . This is equivalent to setting up routes which will only be served if the user is authenticated and equivalent redirection for non-authenticated users 

Caddy configuration could be something like this 

```
authorize  all {
    failure /login
}
```

```
authorize /api1 /api2 /api2  {
        failure /login
} 
```

```
authorize   {
       route /api1
       failure /api1guest

} 
```

mholt:
Cool idea. Let me hold off on deciding on it until we figure out whether to grow Caddy's middleware collection - and if so, how best to do it. In the meantime, I'll leave this open for discussion.

Karthic-Hackintosh:
Sure .....

markerikson:
I would love it if Caddy added additional support for some kind of authentication layer.  Our (internal-ish) app currently protects all backend services with a homegrown Python reverse proxy that also connects to an LDAP server for user/pass authentication.  We had to write that ourselves after various deployment constraints forced us to abandon using Apache for those tasks.  I would love it if there was some way to use Caddy as a replacement for that homegrown code, particularly since I'm hoping that we might be able to use Websockets in our next development cycle and I see that Caddy just added support for proxying WS connections.

mholt:
I forgot to update this issue since we rolled out custom builds. @Karthic-Hackintosh, this is a possibility now. Perhaps you and @markerikson could come up with something that is suitable for both of you and the largest possible user base. I'll be happy to provide some feedback along the way. One thing I'd recommend initially is to try to keep it all under a single directive.

Karthic-Hackintosh:
[...]

PR #6665, opened 2024-10-28
cmd: Allow `add-package` to select version of package

### LINKED ISSUE #6549: "caddy add-package" could  allow adding by version
At the moment we cant control the version of the package added ?

The golang module system could allow us to add by tag or branch. 

I expect many developers would want to use a tag ( as in git tag ) to help remediate supply channel attacks or other things of this nature.

```sh
caddy add-package -h

Downloads an updated Caddy binary with the specified packages (module/plugin)
added. Retains existing packages. Returns an error if the any of packages are
already included. EXPERIMENTAL: May be changed or removed.

Usage:
  caddy add-package <packages...> [flags]

Flags:
  -h, --help          help for add-package
  -k, --keep-backup   Keep the backed up binary, instead of deleting it
```

This would bring it inline with the ```list-modules --packages --versions``` command, which is version aware, so that its all consistent  and all commands are version aware.

```sh
list-modules --packages --versions

Standard modules: 121

exec v0.0.0-20240603212820-a42a5b2ae10f github.com/abiosoft/caddy-exec
http.handlers.exec v0.0.0-20240603212820-a42a5b2ae10f github.com/abiosoft/caddy-exec

  Non-standard modules: 2
```

There is no point making ```caddy remove-package ```version aware since the binary can only have one version of a package.

## final look and feel

For example, its is currently:

```

caddy add-package -h

Downloads an updated Caddy binary with the specified packages (module/plugin)
added. Retains existing packages. 

Returns an error if the any of packages are
already included. EXPERIMENTAL: May be changed or removed. 
```

with versioning it becomes :

```

caddy add-package -h

Downloads an updated Caddy binary with the specified packages (module/plugin)
added. Retains existing packages. 

Returns a success if the any of package versions are changed that are
already included. EXPERIMENTAL: May be changed or removed. 


```

Botelho31:
Closes #6549

### DIFF
--- a/cmd/packagesfuncs.go
+++ b/cmd/packagesfuncs.go
@@ -46,6 +46,25 @@ func cmdUpgrade(fl Flags) (int, error) {
 	return upgradeBuild(pluginPkgs, fl)
 }
 
+func splitModule(arg string) (module, version string, err error) {
+	const versionSplit = "@"
+
+	// accommodate module paths that have @ in them, but we can only tolerate that if there's also
+	// a version, otherwise we don't know if it's a version separator or part of the file path (see #109)
+	lastVersionSplit := strings.LastIndex(arg, versionSplit)
+	if lastVersionSplit < 0 {
+		module = arg
+	} else {
+		module, version = arg[:lastVersionSplit], arg[lastVersionSplit+1:]
+	}
+
+	if module == "" {
+		err = fmt.Errorf("module name is required")
+	}
+
+	return
+}
+
 func cmdAddPackage(fl Flags) (int, error) {
 	if len(fl.Args()) == 0 {
 		return caddy.ExitCodeFailedStartup, fmt.Errorf("at least one package name must be specified")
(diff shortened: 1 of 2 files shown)


cmd/packagesfuncs.go
```diff
@@ -46,6 +46,25 @@ func cmdUpgrade(fl Flags) (int, error) {
 	return upgradeBuild(pluginPkgs, fl)
 }
 
+func splitModule(arg string) (module, version string, err error) {
+	const versionSplit = "@"
+
+	// accommodate module paths that have @ in them, but we can only tolerate that if there's also
+	// a version, otherwise we don't know if it's a version separator or part of the file path (see #109)
```
mholt, line 53: What is `#109`? #109 ??

mholt (changes requested):
Thank you -- LGTM, although I am puzzled by one of the comments.

cmd/packagesfuncs.go
Botelho31: This is actually a issue ID from xcaddy as i copied a large chunk of the split module function from there. I guess i missed this, ill remove it

mholt: Ah, haha, I thought it looked familiar. Thanks

mholt (approved):

PR #6669, opened 2024-10-31
fix: low latency client streaming waiting for data to close connection after client closes

### LINKED ISSUE #6666: Proper fix for low-latency client streaming
In #4922 we had a request to leave the connection open to the backend when the client disconnects before the response (instead of immediately closing the upstream connection), so that the backend would finish receiving what the client sent, when the client doesn't care about the response.

This was implemented in #4952 by simply ignoring the context cancellation when connecting to the upstream when `flush_interval` is -1.

Obviously, in hindsight, this is flawed. Not only does it overload `flush_interval` in an incorrect and unexpected way, it leaves resources in use.

We had a sponsor report that gRPC connections with backends weren't being closed after clients disconnected, due to this patch. (Still being verified but seems quite probable.)

In 8c9e87d0a60cd5395c65dd354522c0865adc47b5 (not mainline branch) I separated `flush_interval` from a new option, `ignore_client_gone` that explicitly enables the behavior of leaving open the connection to the backend.

However, a proper fix is to have an option that, instead of just leaving the connection with the backend open until the backend decides it's done, will close the connection with the backend after it has received everything the client sent before it disconnected. If we mainline a change, I'd prefer it to be the proper fix.

Botelho31:
My suggestion on how to properly fix #6666

### DIFF
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -1367,7 +1367,7 @@ type delayClientDoneContext struct {
 	roundTripDone chan struct{}
 	done          chan struct{}
 	mu            sync.Mutex     // To protect access to the done channel
-	doneSet       atomic.Value   // To indicate whether done has been switched to ctx.Done
+	doneSet       atomic.Bool    // To indicate whether done has been switched to ctx.Done
 	wg            sync.WaitGroup // To wait for the closing of the delay goroutine
 }
 
@@ -1377,11 +1377,10 @@ func NewDelayClientDoneContext(ctx context.Context, parentDone <-chan struct{}) 
 		roundTripDone: make(chan struct{}),
 		done:          make(chan struct{}),
 	}
-	d.doneSet.Store(false)
 
 	closeDone := func(set bool) {
 		d.mu.Lock()
-		if !d.doneSet.Load().(bool) {
+		if !d.doneSet.Load() {
 			if set {
 				// if the round trip is done, we shouldn`t leave this goroutine running, revert the done channel to the ctx.Done()
 				d.doneSet.Store(true)
@@ -1419,7 +1418,7 @@ func (c *delayClientDoneContext) RoundTripDone() {
 
 func (c *delayClientDoneContext) Done() <-chan struct{} {
 	// If the round trip is done, we should return the ctx.Done() channel
-	if c.doneSet.Load().(bool) {
+	if c.doneSet.Load() {
 		return c.Context.Done()
 	}
 	return c.done
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -1381,6 +1381,7 @@ func NewDelayClientDoneContext(ctx context.Context, parentDone <-chan struct{}) 
 		d.mu.Lock()
 		if !d.doneSet.Load().(bool) {
 			if set {
+				// if the round trip is done, we shouldn`t leave this goroutine running, revert the done channel to the ctx.Done()
 				d.doneSet.Store(true)
 			}
 			close(d.done)
@@ -1398,7 +1399,6 @@ func NewDelayClientDoneContext(ctx context.Context, parentDone <-chan struct{}) 
 		case <-parentDone:
 			closeDone(false)
 		case <-d.roundTripDone:
-			// if the round trip is done, we shouldn`t leave this goroutine running, revert the done channel to the ctx.Done()
 			closeDone(true)
 		}
 		d.wg.Done()
@@ -1416,6 +1416,7 @@ func (c *delayClientDoneContext) RoundTripDone() {
 }
 
 func (c *delayClientDoneContext) Done() <-chan struct{} {
+	// If the round trip is done, we should return the ctx.Done() channel
 	if c.doneSet.Load().(bool) {
 		return c.Context.Done()
 	}
--- a/modules/caddyhttp/reverseproxy/reverseproxy.go
+++ b/modules/caddyhttp/reverseproxy/reverseproxy.go
@@ -840,8 +840,10 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
 	// user probably wants us to finish sending the data to the upstream regardless,
 	// and we should expect client disconnection in low-latency streaming scenarios
 	// (see issue #4922)
+	var delayCtx *delayClientDoneContext
 	if h.CloseAfterReceivedBody {
-		req = req.WithContext(NewDelayClientDoneContext(req.Context(), h.ctx.Done()))
+		delayCtx = NewDelayClientDoneContext(req.Context(), h.ctx.Done())
+		req = req.WithContext(delayCtx)
 	}
 
 	// do the round-trip
@@ -850,7 +852,7 @@ func (h *Handler) reverseProxy(rw http.ResponseWriter, req *http.Request, origRe
 	duration := time.Since(start)
 
 	if h.CloseAfterReceivedBody {
-		req.Context().(*delayClientDoneContext).RoundTripDone()
+		delayCtx.RoundTripDone()
 	}
 
 	// record that the round trip is done for the 1xx response handler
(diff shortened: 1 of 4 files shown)


mholt:
Very interesting!

Thanks for the contribution. I'm reviewing it now.

So, uh, forgive me for asking, but does this work as intended? I don't think I would have come up with this, and if it works, that's great. :) It would take me some time to set up a test harness for this (or perhaps we should write an automated test).

Botelho31:
I actually have created some automated tests already ( as i as not too sure myself of if this would work 100% ) but didn't add them as im not too sure where i should put them in the file structure. 

Should i just create a `delay-client-done-context_test` file and add them in? Or should they be somewhere else.

mholt:
@Botelho31 Typically, you'll put the Go code that runs the test in a file of the same name but ending with `_test.go` (see [testing](https://pkg.go.dev/testing)). Are there any external files that are required? (Maybe they can be inlined into the _test.go file, that's ideal.)

francislavoie:
Tests for this could go in `caddytest/integration` I think

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -108,13 +109,15 @@ type Handler struct {
...
-	// set to -1, client disconnection will be ignored and the request
-	// will be completed to help facilitate low-latency streaming.
 	FlushInterval caddy.Duration `json:"flush_interval,omitempty"`
 
+	// Normally, a request will be canceled if the client disconnects
+	// before the response is received from the backend. If enabled,
+	// client disconnection will be ignored and the request with the
+	// backend will carry on until the backend terminates it. This
+	// can help facilitate low-latency streaming. See #4922 and #4952.
+	CloseAfterReceivedBody bool `json:"close_after_received_body,omitempty"`
```
francislavoie, line 119: This will need Caddyfile support (find where `"flush_interval"` is configured)

Botelho31:
I added the unit tests, but im not too sure on how to create a integration test for this behaviour, im gonna take a shot at it.

Botelho31:
So i have written the integration test and have seen it fail, but im finding that the timing is very precise for Caddy to not send the EOF to the backend. It is already triggering context failed every time if not using the new variable.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -1348,6 +1356,75 @@ func statusError(err error) error {
...
+// want a client disconnection to cancel the request during
+// the roundtrip. We delay the context.Done() channel until
+// the round trip is done, or until the parent context is
+// done, whichever comes first.
+type delayClientDoneContext struct {
+	context.Context
+	roundTripDone chan struct{}
+	done          chan struct{}
+	mu            sync.Mutex     // To protect access to the done channel
+	doneSet       atomic.Value   // To indicate whether done has been switched to ctx.Done
```
WeidiDeng, line 1370: Will `atomic.Bool` be a better choice?

Botelho31: I believe that would be better, then i wouldn`t have to initialize it

mholt:
Thank you for this!

I've given this another look-over, and now I wonder if we should be closing the connection proactively after uploading the request, OR if we should wait for the backend to respond / close the connection like we're doing here?

In either case, the client has already disconnected so that doesn't matter, but I'm not sure if the intention is for the proxy to also disconnect before a response?

But I think this is worth a shot either way.

Do you have a moment to test with the [original client reproducer](https://github.com/kuhnchris/caddyserver-ffmpeg-ll-hls-poc) presented in #4922?

Maybe @kuhnchris could offer some insight.

Botelho31:
I actually implemented my integration testing based on his, i will give a try to the ffmpeg script to see if my issue is due to caddy not being fast enough.

WeidiDeng (commented):
According to official [documentation](https://pkg.go.dev/context@go1.23.2#Context), this should return the reference to the same channel. Does this have any impact? Some requests might keep running until caddy is reloaded because they keeps onto the channel before the request is finished, when subsequent calls will return a channel that is from the client. I know http3 does this.

Botelho31:
I just read this part of the documentation. Actually as we are only doing this on a per roundtrip basis ( independently after roundtrip we return to old context.Done ), i don`t think it will affect other requests.

As an update, i have tested this using the python ffmpeg reproducer from the original client and again, i can`t get it to fail. I don`t know if the actual problem on reproducing this behaviour is on the way the backend server handles the roundtrip ( maybe the presence of buffers on the request body in golang? ).

Botelho31:
It's better to get a way to reproduce the error every-time before adding this PR i think. Im sorry for the overall slow pace this is taking. Im gonna give it another try later today

mholt:
No worries. We appreciate you contributing it! It's totally okay if it doesn't go in today, I just didn't want to disappoint. Looking forward to figuring things out and seeing it merged :)

WeidiDeng:
> I just read this part of the documentation. Actually as we are only doing this on a per roundtrip basis ( independently after roundtrip we return to old context.Done ), i don`t think it will affect other requests.

Context is per request. Naturally changing the context of a request won't affect other requests. What do you mean by this?

As I said, I am curious if this will impact the response handling as it's clearly different from stdlib definition. Some parts of the code might hold on to the old reference which leads to context cancelling not firing for these code paths. I'm not asking that we stick to stdlib behaviour. If test checks out, write a comment about it for future reference.

modules/caddyhttp/reverseproxy/reverseproxy.go
```diff
@@ -1348,6 +1356,74 @@ func statusError(err error) error {
 	return caddyhttp.Error(statusCode, err)
 }
 
+// delayClientGoneContext is a special context.Context type
+// intended for use when doing a RoundTrip where you don't
+// want a client disconnection to cancel the request during
+// the roundtrip. We delay the context.Done() channel until
+// the round trip is done, or until the parent context is
+// done, whichever comes first.
+type delayClientDoneContext struct {
```
WeidiDeng, line 1365: I think instead of introducing a new type, we can just use `context.WithCancel` with `h.ctx` and `context.AfterFunc` with the request context and cancel the context introduced before to achieve the same effect. Not only does this save many lines of code, it will also avoid the `Done` returning different channel inconsistency with stdlib. What do you think?

```
		ctx, cancel := context.WithCancel(h.ctx)
		defer cancel()
		stop := context.AfterFunc(req.Context(), cancel)
		defer stop()
```

Botelho31:
I have tested this in the following ways trying to catch an error:

Using the old POC from the original issue
Go Client -> Caddy -> Go Backend Server
Go Client -> Caddy -> Original POC uwsgi
Ffmpeg Original Poc -> Caddy -> Go Backend Server

It seems to me that SOMETHING has changed in golang stdlib RoundTrip in the last two years that has fixed this on their end. I could not in any way shape or form recreate the old error. Maybe this whole pull request isn`t needed anymore, and for now i give up on trying to reproduce this. If someone wants to give it a shot.

Reading the official package github, i believe the commit that could have solved this is [this one](https://github.com/golang/go/commit/505000b2d38c4c93aa7d30e82c61b76a6925b4e9)

francislavoie:
@Botelho31 so is the conclusion that this PR isn't needed after all? Or do you just mean one specific branch/usecase is one you're not able to replicate?

Botelho31:
I cannot replicate the original error #4922 that was fixed by the flush_interval -1 in todays version of caddy/go in order to test if this pull request fixes that.

Maybe it was fixed on stdlib, or maybe it`s just a skill issue on my part ( im quite new to golang ), either way i won`t be trying this any further.

mholt:
Ah wow -- well, thank you so much for working on this @Botelho31! We really appreciate your contribution, regardless of whether we merge it or not. You've been a pleasure to work with! We super appreciate the deep-dive into narrowing things down as well.

I wonder if recent implementation of "full duplex" has anything to do with this? @francislavoie what do you think?

francislavoie:
No, the full-duplex thing is opt-in only (off by default)

mholt:
Okay. So, it sounds like with the current version of Caddy (and/or Go), the issue reported by #4922 may no longer be a problem and we don't need our patch from back then.

In that case, we can even try reverting the original patch and seeing if anyone experiences the original behavior anymore. Sounds like they won't.

That's great -- we might not need a separate setting with a complex implementation, or a buggy patch at all!

@Botelho31 Thank you again so much for working on this. I will respectfully close this PR, but we will refer to it in the future if it comes up again. What I'll probably do then is revert the original patch and credit you in my commit :)

PR #6671, opened 2024-11-01
Use `WithLazy` to prevent eager serialization of the event data

ankon:
Found while load testing: Even with the "events" app not logging out on "DEBUG" it would still go through the pain of serializing the event data. Especially for `tls_get_certificate` that is painful.

Comparison before/after:
image
image


mholt (approved):
Beautiful. Thank you!

PR #6675, opened 2024-11-04
chore: Upgrade dependencies

francislavoie:
Mostly to test if CI likes this better or not.


francislavoie:
I'm super confused. We're not really doing anything wrong.

mohammed90:
The goreleaser check uses `master`, which is why we still see reference to `sdk@1.21.0` in the error trace. I think both our PRs fix it, just the check in the PR action doesn't show it because neither of them are in master. Either of them work. I'll let you choose. The only issue with this PR is the last yolo commit, if you may revert that.

francislavoie:
Yours looks fine

PR #6692, opened 2024-11-15
chore: make FastAbs comment easier to understand

WeidiDeng:


mholt (approved):
Thanks, these are good enhancements for future readers!

Issue #6688, opened 2024-11-13
Storage export Stat/Load issue


mastercoding:
I'm trying to export a live caddy instance, but due to the size of my database, some keys get removed before caddy can actually dump it to file. I'm not sure why this is implemented this way, but I feel like the Stat call here:

https://github.com/caddyserver/caddy/blob/master/cmd/storagefuncs.go#L192-L194

and the Load call here

https://github.com/caddyserver/caddy/blob/master/cmd/storagefuncs.go#L197-L200

Should not terminate the storage export. It happens a lot with challenge tokens, but also with other types of stored data (that by the time the exporter gets to that key, is deleted from the storage). My storage currently has ~300k keys, but it happens within the first 100 records.

Maybe a --ignore-deleted option or something similar would be the way to go?

mholt:
Interesting, not sure we had live exports in mind, but I think I agree; if the key doesn't exist anymore by the time we iterate to it, we should probably log, but ultimately ignore, the error, and carry on. We can do `if errors.Is(err, fs.ErrNotExist)`.

santhoshTpixler:
Something like this work?

```go
for _, k := range keys {
		info, err := stor.Stat(ctx, k)
		if err != nil {
			if errors.Is(err, fs.ErrNotExist) {
				caddy.Log().Warn(fmt.Sprintf("key: %s removed before exporting", k))
				continue
			}
			return caddy.ExitCodeFailedQuit, err
		}

		if info.IsTerminal {
			v, err := stor.Load(ctx, k)
			if err != nil {
				if errors.Is(err, fs.ErrNotExist) {
					caddy.Log().Warn(fmt.Sprintf("key: %s removed before exporting", k))
					continue
				}
				return caddy.ExitCodeFailedQuit, err
			}

			hdr := &tar.Header{
				Name:    k,
				Mode:    0o600,
				Size:    int64(len(v)),
				ModTime: info.Modified,
			}

			if err = tw.WriteHeader(hdr); err != nil {
				return caddy.ExitCodeFailedQuit, fmt.Errorf("writing archive: %v", err)
			}
			if _, err = tw.Write(v); err != nil {
				return caddy.ExitCodeFailedQuit, fmt.Errorf("writing archive: %v", err)
			}
		}
	}
```

mholt:
@santhoshTpixler Yes, exactly like that! Submit a PR :)

keystroke3:
[...]

PR #6697, opened 2024-11-18
cmd: ignore missing keys during storage export

santhoshTpixler:
for #6688


mholt (approved):
Thanks, we'll give this a shot :)

PR #6698, opened 2024-11-18
perf: remove dir redirection when useless in php_fastcgi

dunglas:
Remove the equivalent of this config if neither or appears in `{http.request.uri.path}/index.php` or `{path}.index.php` `try_files`, as it useless:

```
# Add trailing slash for directory requests
@canonicalPath {
	file {path}/index.php
	not path */
}
redir @canonicalPath {path}/ 308
```

Similar to https://github.com/dunglas/frankenphp/pull/1180.

### DIFF
--- a/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
+++ b/modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
@@ -179,7 +179,7 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
 	indexFile := "index.php"
 
 	// set up for explicitly overriding try_files
-	tryFiles := []string{}
+	var tryFiles []string
 
 	// if the user specified a matcher token, use that
 	// matcher in a route that wraps both of our routes;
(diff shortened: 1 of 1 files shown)


modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -310,31 +310,47 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
...
-			HandlersRaw:    []json.RawMessage{caddyconfig.JSONModuleObject(redirHandler, "handler", "static_response", nil)},
-		}
+		var disableDirRedir bool
 
 		// if tryFiles wasn't overridden, use a reasonable default
 		if len(tryFiles) == 0 {
 			tryFiles = []string{"{http.request.uri.path}", "{http.request.uri.path}/" + indexFile, indexFile}
+		} else {
+			longIndexPattern := "{http.request.uri.path}/" + indexFile
+			shortIndexPattern := "{path}/" + indexFile
```
francislavoie, line 320: This would never be true, the placeholders shorthands are replaced with regexp before the Caddyfile is parsed. Always assume the longform in here.

dunglas:
Yes, I'll avoid slashes. I like this convention but Go tools are totally broken when the branch name contains a slash, I always forget about that.

francislavoie (approved):

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -179,7 +179,7 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
 	indexFile := "index.php"
 
 	// set up for explicitly overriding try_files
-	tryFiles := []string{}
+	var tryFiles []string
```
mholt, line 182: About to head out, but initializing as nil might have different implications... can't remember if there was a reason we explicitly initialized as an empty, non-nil slice.

mholt:
Thanks for "backporting" the optimization! :+1:

PR #6699, opened 2024-11-18
fileserver: Add `first_exist_fallback` strategy for `try_files`

dunglas:
When using modern frameworks such as Symfony and Laravel, by convention, all requests are handled by a front controller named `index.php` in the document root.

Currently, Caddy does a call to `fs.Stat()` at each request to check that this file exists, but in case we use such a framework, this call is useless because we know that the file exists.

This PR adds a new `first_exist_fallback` policy that assumes that the last file in the list always exists.

### DIFF
--- a/modules/caddyhttp/fileserver/matcher.go
+++ b/modules/caddyhttp/fileserver/matcher.go
@@ -90,6 +90,7 @@ type MatchFile struct {
 	// How to choose a file in TryFiles. Can be:
 	//
 	// - first_exist
+	// - first_exist_fallback
 	// - smallest_size
 	// - largest_size
 	// - most_recently_modified
@@ -415,13 +416,13 @@ func (m MatchFile) selectFile(r *http.Request) (bool, error) {
 	}
 
 	// setPlaceholders creates the placeholders for the matched file
-	setPlaceholders := func(candidate matchCandidate, info fs.FileInfo) {
+	setPlaceholders := func(candidate matchCandidate, isDir bool) {
 		repl.Set("http.matchers.file.relative", filepath.ToSlash(candidate.relative))
 		repl.Set("http.matchers.file.absolute", filepath.ToSlash(candidate.fullpath))
 		repl.Set("http.matchers.file.remainder", filepath.ToSlash(candidate.splitRemainder))
 
 		fileType := "file"
-		if info.IsDir() {
+		if isDir {
 			fileType = "directory"
 		}
 		repl.Set("http.matchers.file.type", fileType)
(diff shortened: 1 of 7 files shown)


mohammed90:
I don't understand why `go get` fails now 😑

modules/caddyhttp/fileserver/matcher.go
```diff
@@ -415,22 +416,27 @@ func (m MatchFile) selectFile(r *http.Request) (bool, error) {
...
 		repl.Set("http.matchers.file.type", fileType)
 	}
 
 	// match file according to the configured policy
 	switch m.TryPolicy {
-	case "", tryPolicyFirstExist:
-		for _, pattern := range m.TryFiles {
+	case "", tryPolicyFirstExist, tryPolicyFirstExistOrFallback:
+		var maxI int
+		if m.TryPolicy == tryPolicyFirstExistOrFallback {
```
francislavoie, line 435: Is this condition necessary? Doing this math isn't slow.

modules/caddyhttp/fileserver/matcher.go
```diff
@@ -440,8 +446,14 @@ func (m MatchFile) selectFile(r *http.Request) (bool, error) {
 
 			candidates := makeCandidates(pattern)
 			for _, c := range candidates {
+				if m.TryPolicy == tryPolicyFirstExistOrFallback && i == maxI {
```
francislavoie, line 449: I think a comment to explain what the fallback is doing would help (i.e. to skip the IO)

modules/caddyhttp/reverseproxy/fastcgi/caddyfile.go
```diff
@@ -341,6 +341,7 @@ func parsePHPFastCGI(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error
 		rewriteMatcherSet := caddy.ModuleMap{
 			"file": h.JSON(fileserver.MatchFile{
 				TryFiles:  tryFiles,
+				TryPolicy: "first_exist_fallback",
```
francislavoie, line 344: This could be risky if users configure their own `try_files` with a different last item. I think we should set `first_exist_fallback` only if the last item ends with `.php` maybe.

modules/caddyhttp/fileserver/matcher.go
dunglas: I removed the other one, which was useless

modules/caddyhttp/fileserver/matcher.go
```diff
@@ -440,8 +446,15 @@ func (m MatchFile) selectFile(r *http.Request) (bool, error) {
 
 			candidates := makeCandidates(pattern)
 			for _, c := range candidates {
+				// Skip the IO if using fallback policy and it's the latest item
+				if i == maxI {
```
francislavoie, line 450: won't it do `i == 0` now though? That's no good.

francislavoie (approved):
Thanks!

Issue #6700, opened 2024-11-21
Check for `MaxBytesError` instead of a string comparison


francislavoie:
We should stop comparing against `"http: request body too large"` when detecting a ReadCloser error:

https://github.com/caddyserver/caddy/blob/eddbccd298f637c4785c891f5f96dbf103580fa8/modules/caddyhttp/requestbody/requestbody.go#L97

Go stdlib now has a `MaxBytesError` type we can check for.

PR #6701, opened 2024-11-22
requestbody: Type-based error handling for `MaxBytesError`

rishitashaw:
Replaced the hardcoded error string comparison with a type assertion for http.MaxBytesError to make the code more robust and idiomatic.

Before:
```
if err != nil && err.Error() == "http: request body too large" {
    err = caddyhttp.Error(http.StatusRequestEntityTooLarge, err)
}
```
After:
```
if err != nil {
    if _, ok := err.(*http.MaxBytesError); ok {
        err = caddyhttp.Error(http.StatusRequestEntityTooLarge, err)
    }
}
```

Related to #6700

### DIFF
--- a/modules/caddyhttp/requestbody/requestbody.go
+++ b/modules/caddyhttp/requestbody/requestbody.go
@@ -15,6 +15,7 @@
 package requestbody
 
 import (
+	"errors"
 	"io"
 	"net/http"
 	"time"
@@ -94,7 +95,8 @@ type errorWrapper struct {
 
 func (ew errorWrapper) Read(p []byte) (n int, err error) {
 	n, err = ew.ReadCloser.Read(p)
-	if _, ok := err.(*http.MaxBytesError); ok {
+	var mbe *http.MaxBytesError
+	if errors.As(err, &mbe) {
 		err = caddyhttp.Error(http.StatusRequestEntityTooLarge, err)
 	}
 	return


modules/caddyhttp/requestbody/requestbody.go
```diff
@@ -94,8 +94,10 @@ type errorWrapper struct {
 
 func (ew errorWrapper) Read(p []byte) (n int, err error) {
 	n, err = ew.ReadCloser.Read(p)
-	if err != nil && err.Error() == "http: request body too large" {
-		err = caddyhttp.Error(http.StatusRequestEntityTooLarge, err)
+	if err != nil {
+		if _, ok := err.(*http.MaxBytesError); ok {
```
mholt:
```suggestion
	var mbe *http.MaxBytesError
	if errors.As(err, &mbe) {
```

mholt (changes requested):
Thank you! This is pretty good, but there's a lint failure. I have a suggestion that I think could work but I just did it here in the textbox, so if you want to verify locally that it compiles then I think we'll be all set. :)

modules/caddyhttp/requestbody/requestbody.go
rishitashaw: I'm sorry I didn’t see your comment earlier. I’ve made different changes. Specifically, I removed the redundant `nil` check and used the type assertion directly for `*HTTP.MaxBytesError` since the type assertion already guarantees that the error isn’t nil when ok is true.

That said, I really like this project and would love to be more involved. Would it be possible to connect on Slack? I’m eager to contribute more!

francislavoie:
I was thinking we'd use https://pkg.go.dev/errors#example-Is `errors.Is(err, http.MaxBytesError), then the `if` wouldn't need to be nested, it could just remain a one-liner condition.

mholt:
@francislavoie It's a type, not a value.

francislavoie:
Oh right, they implemented `MaxBytesError` as a struct type, not as a `var` error. Gah.

modules/caddyhttp/requestbody/requestbody.go
mholt: Thanks. I still think `errors.As()` is recommended since the error value can be wrapped, thus failing a type assertion.

And yes! I'm running out the door but when I get back I'll send a Slack invite.

mholt (approved):
LGTM. Thanks! @rishitashaw What is a good email address for me to invite?

rishitashaw:
Thanks @mholt 
Please send the invite to rishitashaw49@gmail.com

PR #6704, opened 2024-11-22
metrics: add `go` and `process` collectors

mohammed90:
They're automatically registered in the Default Registry, which we were using but not anymore. Adding them to match old behavior.

Noted in: https://github.com/caddyserver/caddy/pull/6531#issuecomment-2494938615

Thanks @ueffel


hairyhenderson (approved):
:shipit:

PR #6709, opened 2024-11-24
perf: improve try_files matcher

dunglas:
This is an experiment to use `repl.Match()` to (likely) prevent useless string manipulations.
This improve the performance of 2%, but consumes 10% more memory. I'm not sure if it is worth it. (the gain is likely a better on Windows)

Before:

```
goos: darwin
goarch: arm64
pkg: github.com/caddyserver/caddy/v2/modules/caddyhttp/fileserver
cpu: Apple M1 Pro
BenchmarkFileMatcher
BenchmarkFileMatcher-10    	   81501	     14692 ns/op	     920 B/op	      22 allocs/op
PASS
```

After:

```
goos: darwin
goarch: arm64
pkg: github.com/caddyserver/caddy/v2/modules/caddyhttp/fileserver
cpu: Apple M1 Pro
BenchmarkFileMatcher
BenchmarkFileMatcher-10    	   81658	     14410 ns/op	    1022 B/op	      20 allocs/op
PASS
```

### DIFF
--- a/modules/caddyhttp/fileserver/matcher.go
+++ b/modules/caddyhttp/fileserver/matcher.go
@@ -416,15 +416,26 @@ func (m MatchFile) selectFile(r *http.Request) (bool, error) {
 
 	// setPlaceholders creates the placeholders for the matched file
 	setPlaceholders := func(candidate matchCandidate, info fs.FileInfo) {
-		repl.Set("http.matchers.file.relative", filepath.ToSlash(candidate.relative))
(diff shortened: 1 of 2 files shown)


dunglas:
Likey useless.

mholt:
This is kinda surprising to me, as it should only make allocations if invoked.

Maybe the benchmark is not representative of real life use?

dunglas:
`Map()` itself needs to allocate to keep the data available until it is called.

Here it is not worth it at least on Unix because `ToSlash()` basically does nothing on this OS.

Issue #5933, opened 2023-11-07
Certificate for catch-all site is used for requests to other sites


SimJoSt:
The issue is that the certificate loaded via the tls directive from files in the second site block for website2, containing a catch-all, is used for all request to caddy. Only the certificate is used.
The content is still used from the requested domain/site/block. Meaning, each request returns the wanted content from the requested application, just the certificate is from completely different site block.

For example, `https://subdomain.website1.com` returns certificate 2, when certificate 1 is expected.
Expected behaviour would be:
- `https://subdomain.website1.com` -> certificate 1
- `https://www.website1.com` -> certificate 2
- `https://www.website2.com` -> certificate 2
- `https://www.example.com` -> certificate 2
- `https://test.website2.com` -> certificate 2
However, certificate 2 is returned every time.

Caddy is only hit with `https` requests. `http` requests or `http` to `https` redirects don't play a role.

The following docs lead me to believe, I configured this correctly and like it was intended:
- https://caddyserver.com/docs/caddyfile/concepts#addresses
    > [To catch all hosts, omit the host portion of the address, for example, simply https://. This is useful when using On-Demand TLS, when you don't know the domains ahead of time.](https://caddyserver.com/docs/caddyfile/concepts#addresses:~:text=To%20catch%20all%20hosts%2C%20omit%20the%20host%20portion%20of%20the%20address%2C%20for%20example%2C%20simply%20https%3A//.%20This%20is%20useful%20when%20using%20On%2DDemand%20TLS%2C%20when%20you%20don%27t%20know%20the%20domains%20ahead%20of%20time.)
- https://caddyserver.com/docs/caddyfile/concepts#blocks
    > [If a request matches multiple site blocks, the site block with the most specific matching address is chosen. Requests don't cascade into to other site blocks.](https://caddyserver.com/docs/caddyfile/concepts#:~:text=If%20a%20request%20matches%20multiple%20site%20blocks%2C%20the%20site%20block%20with%20the%20most%20specific%20matching%20address%20is%20chosen.%20Requests%20don%27t%20cascade%20into%20to%20other%20site%20blocks.)

Things that were tried:
- replacing `:443` with `https://`
- switching the import order in the main Caddyfile

Main Caddyfile:
```Caddyfile
import /home/deployer/website1/Caddyfile
import /home/deployer/website2/Caddyfile
```

website1 Caddyfile (uses automatic https, no tls directive is configured):
```Caddyfile
subdomain.website1.com {
    ...
}
[...]

Issue #6694, opened 2024-11-15
auto managed domain mistakenly using self-signed


arpitjindal97:
**Expected Behaviour:**

I want to use self-signed certificate for `arpit-test.msmartpay.in` domain only. I want it to automatically manage other domains.

**Actual Behaviour:**

- Caddy gets certificates from let's encrypt for both domains (why both)
- Caddy is using self-signed for both

**Caddyfile**:

```
{
    servers {
        metrics
    }
    auto_https ignore_loaded_certs
}

arpit.msmartpay.in {
    respond "Hello World"
}

arpit-test.msmartpay.in {
    tls /ssl/msmartpay.in/self/domain.crt /ssl/msmartpay.in/self/privkey.pem
    respond "Hello World"
}
```

**domain.crt**:

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            4a:b0:51:03:a6:50:ec:05:d7:78:7d:17:52:c2:ca:cd:55:25:ac:ab
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=SL, ST=Western, L=Colombo, OU=ABC, CN=arpit.msmartpay.in
        Validity
            Not Before: Nov 15 21:44:09 2024 GMT
            Not After : Nov 13 21:44:09 2034 GMT
        Subject: C=SL, ST=Western, L=Colombo, OU=ABC, CN=arpit.msmartpay.in
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (2048 bit)
                Modulus:
                    00:bc:46:76:77:dd:16:77:75:8e:32:87:36:75:ac:
                    .................
                    b4:17
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Basic Constraints:
                CA:FALSE
            X509v3 Key Usage:
                Digital Signature, Non Repudiation, Key Encipherment
            X509v3 Subject Alternative Name:
                DNS:arpit.msmartpay.in, DNS:*.msmartpay.in
            X509v3 Subject Key Identifier:
                E9:51:26:A4:56:31:40:CA:D5:DA:C2:35:92:32:9D:2B:2C:9E:7B:6A
    Signature Algorithm: sha256WithRSAEncryption
    Signature Value:
        3b:ae:10:b6:d9:cd:18:54:cd:0b:97:2b:b2:3d:52:9e:91:9f:
        ..................................
 
```

mholt:
[...]

PR #6712, opened 2024-11-27
httpcaddyfile: Implement `force_automate`

francislavoie:
Closes https://github.com/caddyserver/caddy/issues/5933

I think this should work according to the JSON config output, but I didn't actually test it at runtime to verify that it yields the expected behaviour. @mholt you had a repro set up locally if you want to give it a spin while reviewing.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -84,7 +84,7 @@ func parseBind(h Helper) ([]ConfigValue, error) {
 
 // parseTLS parses the tls directive. Syntax:
 //
-//	tls [<email>|internal]|[<cert_file> <key_file>] {
+//	tls [<email>|internal|force_automate]|[<cert_file> <key_file>] {
 //	    protocols <min> [<max>]
 //	    ciphers   <cipher_suites...>
 //	    curves    <curves...>
@@ -107,6 +107,7 @@ func parseBind(h Helper) ([]ConfigValue, error) {
 //	    dns_challenge_override_domain <domain>
 //	    on_demand
 //	    reuse_private_keys
+//	    force_automate
 //	    eab                           <key_id> <mac_key>
 //	    issuer                        <module_name> [...]
 //	    get_certificate               <module_name> [...]
@@ -126,6 +127,7 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 	var certManagers []certmagic.Manager
 	var onDemand bool
 	var reusePrivateKeys bool
+	var forceAutomate bool
 
 	firstLine := h.RemainingArgs()
 	switch len(firstLine) {
@@ -133,8 +135,10 @@ func parseTLS(h Helper) ([]ConfigValue, error) {
 	case 1:
 		if firstLine[0] == "internal" {
 			internalIssuer = new(caddytls.InternalIssuer)
+		} else if firstLine[0] == "force_automate" {
+			forceAutomate = true
 		} else if !strings.Contains(firstLine[0], "@") {
-			return nil, h.Err("single argument must either be 'internal' or an email address")
+			return nil, h.Err("single argument must either be 'internal', 'force_automate', or an email address")
 		} else {
 			acmeIssuer = &caddytls.ACMEIssuer{
 				Email: firstLine[0],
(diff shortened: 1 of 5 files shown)


francislavoie:
What's annoying about this is it's hard to test a real-world scenario because the TLS automation stuff has special conditions for `.localhost` domains (to use the internal issuer), so using fake domains may have different results than using real domains. But I did confirm _at least_ that using `tls force_automate` does correctly hoist a connection policy to the top making an automated cert take precedence over a wildcard loaded cert.

SimJoSt:
Tested, works.
Took me some time to solve some permission issues I caused myself after building it for linux arm64, but I made it :)

While I cannot speak to the implementation and the general approach, I successfully tested this in production 🙈 with real domains and services.

SimJoSt:
In another test I ran `caddy adapt`, like mentioned in https://github.com/caddyserver/caddy/issues/5933#issuecomment-2503180790 and ran into an issue.
Apparently, this command doesn't know about the new `force_automate` option for the `tls` directive yet, and fails.
It seems there are more places in the code where this new option needs to be defined.

francislavoie:
You just ran adapt with your old version of Caddy. Make sure you run adapt with the correct binary.

Thanks for testing! Appreciated.

SimJoSt:
You are absolutely right :D Sorry, for the confusion. `adapt` works without any issue.

mholt:
Thanks for working on this. To be sure I'm clear, is there anything left on this issue? https://github.com/caddyserver/caddy/issues/5933#issuecomment-2505237652 seems to suggest it has some unintuitive components... and I agree, this does feel like a hack; and I'm still wondering if we should just fix the existing behavior...

SimJoSt:
@mholt from testing the feature for my specific use-case, it was feature complete for me.
I cannot speak to an integration perspective or if @francislavoie has more changes planned.

From how I interpret the hierarchy in Caddys (Caddyfile) config I would expect this option to allow me to force the tls automation, if there was something else set at the global level. I am not sure, if it is possible to load certificates globally..

The current behavior of using loaded certs from a separate site with a wildcard domain in sites with no tls directive (default) or one with just an email provide for automated certificates, is still unintuitive in my opinion.
Why is a loaded cert from another scope/site used in a scope/site where it is not mentioned.
I also don't expect the `root` directive from the wildcard site to be used in my other sites... ;)

If this behavior will not be changed and this PR makes it to upstream, I would suggest changing it a little.
If I supply `tls email@example.com`, I am already indicating that I want automated certs. Shouldn't that be enough to override the loaded certs?

francislavoie:
I still think `force_automate` is most descriptive of what it's doing. It's forcing the Caddyfile adapter to produce config to automate for that name, regardless of any of the other rules in Automatic HTTPS. This feature doesn't actually directly interact with wildcard names, even if that's the main purpose of it.

mholt (approved):
Alrighty, sounds good to me. Let's try it, but let's be sure to document this as experimental, since I still have a hunch there's a cleaner way to do this. I suspect there may be edge cases we haven't tested for here as well; but I can't think of them.

Thank you for your patience with me on this one!

arpitjindal97:
On Testing this, it doesn't seem to work when wildcard TLS has CN as domain. Also, i don't understand why dedicated Site TLS is being used globally.

See this issue to reproduce https://github.com/caddyserver/caddy/issues/6694

polarathene:
It's not specific to CN / main domain of a certificate. It's due to the exact site-address match.

There's probably an assumption that you wouldn't load a certificate from an external source inaan unrelated site block with a specific site-address and then `tls force_automate` to request Caddy provision a certificate instead. Usually you'd just do one or the other.

This feature was motivated for opt-out of preferring an existing wildcard cert match vs provisioning a new cert with the exact site-address.

```Caddyfile
# Ignore the self-signed cert, provision a new cert instead:
example.com {
  tls force_automate
  respond "hello world"
}

# This loads an external certificate that also matches
# for the site-address example.com
example.localhost {
  tls cert.pem key.pem
  respond "self-signed" 
}
```

That's your situation where `tls force_automate` is bugged.

Note that AFAIK this directive variant doesn't care about if it's a LetsEncrypt or self-signed certificate that is loaded externally or managed by Caddy, since a wildcard certificate match could be from any of that, thus `tls force_automate` is presently more about getting a certificate that's an exact match to the site-address I think.

Still given the name it should prioritize selecting a certificate with exact match that has Caddy automate management of that cert (that isn't wildcard, unless the site address is a wildcard itself).

SimJoSt:
Just to give some mid-term feedback: We used it on v2.10.0, and it works exactly as expected and planned. So it solves the original issue.
Like @mholt and I said, there is probably a cleaner solution out there for another time:
- https://github.com/caddyserver/caddy/issues/5933#issuecomment-2505237652
- https://github.com/caddyserver/caddy/pull/6712#pullrequestreview-2521432384

For this reason, it was mentioned, that the feature should be documented as experimental. However, I couldn't find any documentation in the changes of this PR nor the actual docs: https://caddyserver.com/docs/

Issue #6702, opened 2024-11-22
`caddy fmt` will silently ignore all but the first file path passed to it


alexwlchan:
If you call `caddy fmt` with the path to more than one file, it only formats the first file.

For example:

```console
$ caddy fmt 1.Caddyfile 2.Caddyfile --overwrite
```

then `1.Caddyfile` will be reformatted, but `2.Caddyfile` will be silently ignored.

If I look [in the docs](https://caddyserver.com/docs/command-line#caddy-fmt), I can see that `caddy fmt` only supports passing a single path – but it took me a while to realise that! This is different from other formatting tools I've used (e.g. `black`, `terraform fmt`) where you can pass multiple files and format them all together.

I think it would be less confusing if `caddy fmt` either:

* Formatted every `<path>` passed to it, or
* Threw an error if you tried to pass more than one `<path>`

## Context

I'd written a command to find all the Caddyfiles in my repo and format them:

```console
$ find . -name Caddyfile | xargs --verbose caddy fmt
caddy fmt 1/Caddyfile 2/Caddyfile
```

and I was getting inconsistent results between different machines – it took me a while to notice that `find` was returning results in an inconsistent order, and `caddy ` was only formatting the first in the list.

I don't really mind what `caddy` does, but the current behaviour caused a lot of confusion.

## Version

```console
$ caddy --version
v2.8.4 h1:q3pe0wpBj1OcHFZ3n/1nl4V4bxBrYoSoab7rL9BMYNk=
```

mholt:
Thanks for the report. Should be an easy fix, I suppose -- anyone is welcome to contribute!

alexwlchan:
Noting in case I have time to pick this up: this looks like the relevant function. https://github.com/caddyserver/caddy/blob/eddbccd298f637c4785c891f5f96dbf103580fa8/cmd/commandfuncs.go#L561-L619

francislavoie:
Well, you can only have one root Caddyfile, so I don't think it really makes sense to try to format more than one. If you're doing that because you have `import` etc, then I think we should update `fmt` to follow imports, possibly, and also format those files (when `-w` is used).

So I think yeah disallowing more than one file probably makes sense.

Also while we're at it I think we could add `--config` as a flag (which is the same as the one arg, and if set args can be empty), cause most of Caddy's other commands use `--config/-c` to specify the config file, ran into that on occasion as a footgun when I'm hitting UP in my terminal to edit the command and run another.

[...]

PR #6717, opened 2024-11-29
cmd: `fmt` prevents multiple files passed as args, add support for `--config`

armadi1809:
Addresses #6702 by explicitely preventing more than one config file passed to the fmt command. 
Also Enables the (--config/-c) flags support for the fmt command. 

@francislavoie I am not quite sure on what would be the best way to approach formatting imports as well. would we get all the import statements from the root config file and check if they refer to a file that we can resolve the path to, and then recursively do the same for the files we extracted?

### DIFF
--- a/cmd/commandfuncs.go
+++ b/cmd/commandfuncs.go
@@ -560,10 +560,15 @@ func cmdValidateConfig(fl Flags) (int, error) {
 
 func cmdFmt(fl Flags) (int, error) {
 	configFile := fl.Arg(0)
-	if configFile == "" {
-		configFile = "Caddyfile"
+	configFlag := fl.String("config")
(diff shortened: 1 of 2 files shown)


cmd/commandfuncs.go
```diff
@@ -560,10 +560,16 @@ func cmdValidateConfig(fl Flags) (int, error) {
 
 func cmdFmt(fl Flags) (int, error) {
 	configFile := fl.Arg(0)
-	if configFile == "" {
+	configFlag := fl.String("config")
+	fmt.Println(configFile, configFlag)
```
mohammed90, line 564:
```suggestion

```

armadi1809: Removed extra print statement

mholt (approved):
Thanks for contributing this! :smiley:

PR #6736, opened 2024-12-07
feat: good default for file_server.precompressed

dunglas:
Currently, the following snippet doesn't enable precompression:

```
file_server {
    precompressed
}
```

Supported formats must be passed explicitly.

This patch allows to omit the formats for simplicity. By default, Brotli, Zstandard, and gzip are enabled (in this order).
Brotli is preferred to Zstandard, because according to [Paul Calvano's article](https://paulcalvano.com/2024-03-19-choosing-between-gzip-brotli-and-zstandard-compression/), Brotli is better (smaller files) for static assets.


mholt (approved):
This is probably a good idea. Thanks!

Issue #6733, opened 2024-12-07
reverse_proxy: h2 websocket stuck if client needs to send the first message and encode is enabled


WeidiDeng:
With [this](https://github.com/caddyserver/caddy/commit/9c0c71e5774593e15a555d5c88de0af011e46394) merged, I noticed some of my websockets failed to connect while others succeed.

I disabled `encode` and found out now these requests connect successfully.

The problem is that client requests contain the header `accept-encoding: gzip, deflate, br, zstd`, which is wrapped

https://github.com/caddyserver/caddy/blob/9c0c71e5774593e15a555d5c88de0af011e46394/modules/caddyhttp/encode/encode.go#L155-L160

This response writer will write informational header immediately

https://github.com/caddyserver/caddy/blob/9c0c71e5774593e15a555d5c88de0af011e46394/modules/caddyhttp/encode/encode.go#L250-L252

During websocket upgrade, for http1, the upgrade will write `101` status,

https://github.com/caddyserver/caddy/blob/9c0c71e5774593e15a555d5c88de0af011e46394/modules/caddyhttp/reverseproxy/streaming.go#L122

which will be flushed to the client immediately.

However, for h2, although `FlushError` is called

https://github.com/caddyserver/caddy/blob/9c0c71e5774593e15a555d5c88de0af011e46394/modules/caddyhttp/reverseproxy/streaming.go#L111

response code is `200` same as any other status that indicates success. Because we want to determine if the response can be encoded, this code is stored and flushed later if there is any data from the upstream and we can be sure whether the response can be encoded

https://github.com/caddyserver/caddy/blob/9c0c71e5774593e15a555d5c88de0af011e46394/modules/caddyhttp/encode/encode.go#L262-L269

If the websocket transport a protocol that server sends response first, it is OK as that will cause the status and data to be sent. But if the client sends data first, it won't happen because the response will never be sent, and client thinks the handshake didn't finish.

For now, I created a matcher to ignore encoding for h2 websocket upgrade

```
@not_h2_ws not {
    header :protocol *
    method CONNECT
    protocol http/2
}
encode @not_h2_ws zstd gzip
```

But I think an elegant solution should be implemented.

WeidiDeng:
I suppose I can manually unwrap the response writer until no more unwrap can be done. And use this unwrapped response writer instead. What do you think @mholt ?

WeidiDeng:
[...]

PR #6738, opened 2024-12-08
encode: write status immediate for success response for CONNECT requests

WeidiDeng:
fix [6733](https://github.com/caddyserver/caddy/issues/6733).

### DIFF
--- a/modules/caddyhttp/encode/encode.go
+++ b/modules/caddyhttp/encode/encode.go
@@ -203,7 +203,7 @@ func (enc *Encode) addEncoding(e Encoding) error {
 // be closed after the handler completes.
 func (enc *Encode) openResponseWriter(encodingName string, w http.ResponseWriter, isConnect bool) *responseWriter {
 	var rw responseWriter
-	return enc.initResponseWriter(&rw, encodingName, w, isConnect bool)
+	return enc.initResponseWriter(&rw, encodingName, w, isConnect)
 }
 
 // initResponseWriter initializes the responseWriter instance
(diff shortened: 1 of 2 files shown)


modules/caddyhttp/encode/encode.go
```diff
@@ -156,7 +156,7 @@ func (enc *Encode) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyh
 			if _, ok := enc.writerPools[encName]; !ok {
 				continue // encoding not offered
 			}
-			w = enc.openResponseWriter(encName, w)
+			w = enc.openResponseWriter(encName, w, r.Method == http.MethodConnect)
```
bt90, line 159: We should try to unify the handling of h1 and h2 websockets. Maybe we can drop the workaround in `AcceptedEncodings` for h1 websockets and extend the pass-through detection here?

```go
r.Header.Get("Sec-WebSocket-Key") != "" || r.Method == http.MethodConnect
```

WeidiDeng: `CONNECT` is not just for websockets, it's for creating a tunnel. It shouldn't be encoded if response is successful regardless of what's it's used.

For websocket, I don't know why this exists. Successful upgrade will bypass response writer anyway. I think there was once a protocol upgrade called derp used by tailscale that failed because `encode` saved the `101` status.

bt90: > I think there was once a protocol upgrade called derp used by tailscale that failed because `encode` saved the `101` status.

Indeed: #5419

PR #6749, opened 2024-12-19
encode: try to use sendfile when compression is not used

WeidiDeng:
Whenever `Accept-Encoding: gzip` or the like is included in the request headers and `encode` is enabled, currently there is no way for `sendfile` to be used even if the response is not compressed.

This patch attempts to fix this by implementing `io.ReaderFrom`. Similar to [5022](https://github.com/caddyserver/caddy/pull/5022).


modules/caddyhttp/encode/encode.go
```diff
@@ -347,6 +347,49 @@ func (rw *responseWriter) Write(p []byte) (int, error) {
...
+	// WriteHeader wasn't called and is a CONNECT request, treat it as a success.
+	// otherwise, determine if the response should be compressed.
+	if rw.isConnect && !rw.wroteHeader && rw.statusCode == 0 {
+		rw.WriteHeader(http.StatusOK)
+	}
+
+	rf, ok := rw.ResponseWriter.(io.ReaderFrom)
+	// sendfile can't be used anyway
+	if !ok {
+		return io.Copy(writerOnly{rw}, r)
```
mholt, line 371: Why is `writerOnly` necessary here (and below)? Maybe put a comment either here or where it is declared to make it more obvious.

mholt (commented):
Thanks Weidi! Looking forward to seeing how this works, but I do have a question.

mholt (approved):
Thanks for the comments. Let's give it a try then :)

PR #6762, opened 2025-01-03
Enable dependabot for Go modules

SuperQ:
Update Go modules montly using dependabot. This will help keep packages up to date for security fixes and other improvements.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,12 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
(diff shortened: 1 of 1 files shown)


mohammed90:
I disagree on the benefits brought in by dependabot. It is noise. It doesn't understand how Go modules work and blindly upgrade deps, even transitive deps which most often ought to be left alone.

francislavoie:
It causes us more trouble than it's worth. We're able to upgrade dependencies ourselves manually when we're ready to.

Issue #6763, opened 2025-01-04
Headers can't be added by matching against response matchers


lilnasy:
Caddyfile:

```Caddyfile
example.com:443 {
	route {
		@success status 2xx
		header @success Cache-Control "public, max-age=3600"
		file_server {
			root /www
		}
	}
}
```

Error:
```
Error: adapting config using caddyfile: getting matcher module 'status': module not registered: http.matchers.status
```

Since Chrome caches 404 responses, I want to add the `Cache-Control` response header only when the response status is ok (2xx).

I looked at [Response Matchers (Caddyfile) -- Caddy documentation](https://caddyserver.com/docs/caddyfile/response-matchers#example), which documents the response matcher above, and combined that with the `header` directive as documented in [header (Caddyfile directive) -- Caddy documentation](https://caddyserver.com/docs/caddyfile/directives/header#syntax), but counterintuitively, `header` can only use request matchers.

[encode (Caddyfile directive) -- Caddy documentation](https://caddyserver.com/docs/caddyfile/directives/encode#match) leads to believe that response matchers can only be used in specific places, not in a named matcher (is the response matcher example out of date?). However, since both the `header` directive and the `encode` directive act on pre-existing responses, it's natural that both can match using response matchers, not just `encode`.

A caddy community help post ([Header only on successful response](https://caddy.community/t/header-only-on-successful-response/26158)) suggests a workaround using `php_fastcgi`. However, that makes the Caddyfile unreadable and opaque since my application really has nothing to do with PHP.

francislavoie:
You can use [`intercept`](https://caddyserver.com/docs/caddyfile/directives/intercept) for this

lilnasy:
Thanks, that helps a lot!

Is there a limitation preventing `header` from working similarly to `encode`?

francislavoie:
I'm not sure what you mean. They do completely different things. 

The `header` directive's job is to manipulate response headers. It takes a request matcher, just like every other directive. That's it.

The `encode` directive enables compression, and it takes a request matcher as well. But because only some content types should be compressed, and we only know the content type after the handler that writes the response is done its job, it also has a response matcher option (with a good default).

They have completely different responsibilities.

lilnasy:
Thanks for the reply. I should clarify, what I meant to ask was, could `header` use a response matcher in the same that `encode` does. After looking at the source code, I learned that not only it can, but it already does!

For example, the following json configuration matches 2xx responses:

```json
{
[...]

PR #6765, opened 2025-01-04
feat(caddyfile: match block for header directive

lilnasy:
- https://github.com/caddyserver/caddy/issues/6763

Allows the header directive to specify a response matcher to apply the header operation only when the response has a certain header or status.

This PR only adds the syntax to the caddyfile parser. The runtime implementation uses the preexisting logic of the "Require" field. I went with the `match` name for parity with the `encode` directive.

There is a slight breaking concern: similar to how "defer" cannot be used a header name, a header entry where the header name is exactly "match" will no longer be considered. If there is someone using it, they will have to rename the header to "Match" (different case) instead.

Syntax example:
```caddyfile
example.com {
	encode gzip {
		match {
			header Content-Type application/json*
		}
	}
	header Cache-Control "public, max-age=3600" {
		match {
			status 2xx
		}
	}
	file_server
}

```


mholt (approved):
> There is a slight breaking concern: similar to how "defer" cannot be used a header name, a header entry where the header name is exactly "match" will no longer be considered. If there is someone using it, they will have to rename the header to "Match" (different case) instead.

I believe this is only true when using a block; but when using the single-line syntax, there should be no conflict. IIRC.

---

Anyway, LGTM at a glance! Maybe @francislavoie could double-check but I think we can merge this.

mholt:
j/k -- we need to fix the lint error first:

>   Error: modules/caddyhttp/headers/caddyfile.go:104:39: Error return value of `caddyhttp.ParseNamedResponseMatcher` is not checked (errcheck)
>  			caddyhttp.ParseNamedResponseMatcher(h.NewFromNextSegment(), responseMatchers)

modules/caddyhttp/headers/caddyfile.go
```diff
@@ -99,6 +99,13 @@ func parseCaddyfile(h httpcaddyfile.Helper) ([]httpcaddyfile.ConfigValue, error)
 			handler.Response.Deferred = true
 			continue
 		}
+		if field == "match" {
+			responseMatchers := make(map[string]caddyhttp.ResponseMatcher)
+			caddyhttp.ParseNamedResponseMatcher(h.NewFromNextSegment(), responseMatchers)
```
lilnasy, line 104:
```suggestion
			err := caddyhttp.ParseNamedResponseMatcher(h.NewFromNextSegment(), responseMatchers)
			if err != nil {
				return nil, err
			}
```

caddytest/integration/caddyfile_adapt/header.caddyfiletest
```diff
@@ -12,6 +12,9 @@
 	@images path /images/*
 	header @images {
 		Cache-Control "public, max-age=3600, stale-while-revalidate=86400"
+		match {
+			status 200
+		}
```
francislavoie, line 17: Please add another one which tests having the matcher on a single line, i.e. `match status 200`

francislavoie (approved):

lilnasy:
Thanks for the reviews! Seem like it just needs a final round tests after rebasing on top of the recent pushes.

lilnasy:
@mholt Could you approve the workflow run?

PR #6772, opened 2025-01-07
client_auth: wire up leaf verifier Caddyfile

### LINKED ISSUE #6771: `trusted_leaf_certs` functionality after its deprecation in Caddyfile
Hey,

I would like to only allow clients with certain client leaf certificates to access some services. Unfortunately `trusted_leaf_certs` is deprecated and I could not find a way to get this working without it via Caddyfile. Can anyone point me into the right direction?

Thx and best regards,
PhysicsIsAwesome

PS: I tried to sign up for caddy.community to ask this question instead of opening an issue, but could not find a way to sign up.

mohammed90:
Closes #6771

### DIFF
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -941,6 +941,31 @@ func (l *LeafCertClientAuth) Provision(ctx caddy.Context) error {
 	return nil
 }
 
+// UnmarshalCaddyfile implements caddyfile.Unmarshaler.
+func (l *LeafCertClientAuth) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.NextArg()
+	for nesting := d.Nesting(); d.NextBlock(nesting); {
+		modName := d.Val()
+		if !strings.HasPrefix(modName, "load_") {
+			return d.Err("expected a leaf certificate loader module name prefixed with `load_`")
+		}
+		modName = strings.TrimPrefix(modName, "load_")
+		mod, err := caddyfile.UnmarshalModule(d, "tls.leaf_cert_loader."+modName)
+		if err != nil {
+			return d.WrapErr(err)
+		}
+		vMod, ok := mod.(LeafCertificateLoader)
+		if !ok {
+			return fmt.Errorf("leaf module '%s' is not a leaf certificate loader", vMod)
+		}
+		l.LeafCertificateLoadersRaw = append(
+			l.LeafCertificateLoadersRaw,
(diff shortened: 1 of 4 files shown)


modules/caddytls/connpolicy.go
```diff
@@ -941,6 +941,31 @@ func (l *LeafCertClientAuth) Provision(ctx caddy.Context) error {
...
 
+// UnmarshalCaddyfile implements caddyfile.Unmarshaler.
+func (l *LeafCertClientAuth) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.NextArg()
+	for nesting := d.Nesting(); d.NextBlock(nesting); {
+		modName := d.Val()
+		if !strings.HasPrefix(modName, "load_") {
+			return d.Err("expected a leaf certificate loader module name prefixed with `load_`")
+		}
+		modName = strings.TrimPrefix(modName, "load_")
```
mohammed90, line 952: @francislavoie, what do you think of this from Caddyfile UX perspective?

francislavoie: Do we need `load_` at all?

mohammed90: I thought I should future-proof it in case other parameters are added besides the loaders. I thought the namespace will be clobbered if I don't distinguish them with a prefix or nest them into a block. I don't know...

mholt: The module name is already prefixed by `tls.leaf_cert_loader.`. Are there other "kinds" (?) of modules that go in this namespace that aren't loading?

francislavoie: Shouldn't it be like `verifier leaf file <filename>` etc?

mohammed90: I'm saying what if `leaf` has more parameters in the future, not the loader modules. Anyways, looking at what `VerifyClientCertificate` does, I can't imagine any new behavioral parameters that may be added in the future.

The format `verifier leaf file <filename>` doesn`t work because it blocks the use multiple loaders, which is currently possible. In other words, this is now possible:

```
verifier leaf {
    file file-1.pem file-2.pem
    file more-cert.pem
    folder /path/to/cert/repo
}
```
The one-liner disallows this use.

mohammed90: I've removed the `load_` prefix. I've also made it accept either in-line or block, but not both. All good?

mholt (approved):
Thank you Mohammed! Love the test cases.

PR #6778, opened 2025-01-09
chore: more comments about reverse_proxy buffering and add new tests

WeidiDeng:
While fixing [6757](https://github.com/caddyserver/caddy/issues/6757), it's found that [this pr](https://github.com/caddyserver/caddy/pull/6638) set the wrong `Content-Length` when buffering. Add new tests to prevent this type of issue when refactoring buffering later.


mholt (approved):
Thanks for the improvements to the comments, I think this is important, so I appreciate it!

PR #6800, opened 2025-01-22
fix: update broken link to Ardan Labs

sbruens:
I'm not sure if there's a better contact URL that would be a better replacement.


mholt (approved):
Thanks -- yeah, I dunno a better link either. This works for now :)

sbruens:
Signed the CLA.

mholt:
Got it, thanks!

Issue #6489, opened 2024-08-01
[reverse_proxy] active_health check to http endpoint does not work when upstream is via ssl/tls


KlettIT:
Hi,

I have a reverse_proxy configuration here which uses an https endpoint as upstream. but the health check must take place on port 80 (http). i have configured this as far as possible.

```
              "handler": "reverse_proxy",
              "transport": {
                "protocol": "http",
                "tls": {
                  "server_name": "sso.example.com"
                }
              },
              "load_balancing": {
                "selection_policy": { "policy": "client_ip_hash" }
              },
              "health_checks": {
                "active": {
                  "uri": "/adfs/probe",
                  "interval": "30s",
                  "timeout": "5s",
                  "passes": 2,
                  "fails": 3,
                  "port": 80,
                  "expect_status": 200
                }
              },
              "upstreams": [
                {
                  "dial": "SERVER01:443"
                },
                {
                  "dial": "SERVER02:443"
                }
              ]
            }
          ]
```
Although I have defined port 80 in health_check, an https request is sent here and not an http request. 

```
Aug 01 10:14:45 SIT-NLBP06.razor.k-sys.io caddy_custom[43523]: {"level":"info","ts":1722500085.501933,"logger":"http.handlers.reverse_proxy.health_checker.active","msg":"HTTP request failed","host":"SERVER01:80","error":"Get \"https://SERVER01:80/adfs/probe\": http: server gave HTTP response to HTTPS client"}
Aug 01 10:14:45 SIT-NLBP06.razor.k-sys.io caddy_custom[43523]: {"level":"info","ts":1722500085.5023623,"logger":"http.handlers.reverse_proxy.health_checker.active","msg":"HTTP request failed","host":"SERVER02:80","error":"Get \"https://SERVER02:80/adfs/probe\": http: server gave HTTP response to HTTPS client"}

```

Is there a way to configure this?

KlettIT:
Yes, I understand your point of view and you're not wrong. But then it makes no sense that I can specify the port at all.
My use-case is certainly a bit special here, but if that counts, in HAProxy you can specify the check port and protocol.

glatzert:
We're currently trying to put ADFS behind Caddy - it works well so far, but active health checks aren't possible, since ADFS only allows to connect to the probing endpoint via http (don't ask me), so this _might_ be relevant for some folks.

Mazwak:
I had the same problem with ADFS.

[...]

PR #6802, opened 2025-01-23
add Transport override option for active health checks

ab14-tech:
- Add Transport override option for active health checks. 
- This enables the active healthcheck to use `http` while the actual upstream uses `https`. 
- This enables more powerful active healthcheck capabilities 

[Related Discussion](https://caddy.community/t/how-to-pass-health-uri-the-specific-upstream-endpoint-as-a-query-parameter/22780/11) 

[Related Issues](https://github.com/caddyserver/caddy/issues/6489)


modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -90,6 +90,9 @@ type ActiveHealthChecks struct {
 	// this value is ignored.
 	Port int `json:"port,omitempty"`
 
+	// The transport to use for health checks. If not set, the handler's transport is used
+	Transport http.RoundTripper `json:"transport,omitempty"`
+
```
mohammed90, line 95: Accepting the PR is something to be discussed, but this isn't how to implement custom transport. See here for how it's done in the reverse proxy handler:

https://github.com/caddyserver/caddy/blob/99073eaa33af62bff51c31305e3437c57d936284/modules/caddyhttp/reverseproxy/reverseproxy.go#L74-78

abales4: great i will take a look. thanks

mholt: Yeah, good point, this could be better off as a plugin.

ab14-tech: thanks for the quick feedback / thoughts 

@mohammed90  I can take a look at how the reverse proxy handler implements its custom transport and try to emulate here. 

@mholt can you expand on what you mean by making this a plugin?

mholt: An `http.RoundTripper` can't really be directly configured from JSON. So you define a module field instead, e.g.:

https://github.com/caddyserver/caddy/blob/99073eaa33af62bff51c31305e3437c57d936284/modules/caddyhttp/reverseproxy/reverseproxy.go#L75-L78

Then in `Provision()`, the module is initialized and converted into an http.RoundTripper:

https://github.com/caddyserver/caddy/blob/99073eaa33af62bff51c31305e3437c57d936284/modules/caddyhttp/reverseproxy/reverseproxy.go#L239-L259

And the resulting http.RoundTripper is stored here, for example:

https://github.com/caddyserver/caddy/blob/99073eaa33af62bff51c31305e3437c57d936284/modules/caddyhttp/reverseproxy/reverseproxy.go#L191

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -90,6 +90,12 @@ type ActiveHealthChecks struct {
 	// this value is ignored.
 	Port int `json:"port,omitempty"`
 
+
+	// Configures the method of transport for the active health checker. 
+	// The default transport is the handler's transport
+	TransportRaw json.RawMessage `json:"transport,omitempty"`
+	Transport http.RoundTripper `json:"transport,omitempty"`
```
mohammed90, line 97: I'd love for it to be unexported, but I haven't inspected the code well enough to see it doable.

```suggestion
	Transport http.RoundTripper `json:"-"`
```

mohammed90 (changes requested):

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -90,6 +90,11 @@ type ActiveHealthChecks struct {
 	// this value is ignored.
 	Port int `json:"port,omitempty"`
 
+
+	// Configures the method of transport for the active health checker. 
+	// The default transport is the handler's transport
+	TransportRaw json.RawMessage `json:"transport,omitempty" caddy:"namespace=http.reverse_proxy.health_checks.active.transport inline_key=protocol"`
```
mohammed90, line 96:
```suggestion
	TransportRaw json.RawMessage `json:"transport,omitempty" caddy:"namespace=http.reverse_proxy.transport inline_key=protocol"`
```

mohammed90 (changes requested):

modules/caddyhttp/reverseproxy/healthchecks.go
ab14-tech: fair, transport is now unexported.

modules/caddyhttp/reverseproxy/healthchecks.go
```diff
@@ -128,6 +133,7 @@ type ActiveHealthChecks struct {
 	// body of a healthy backend.
 	ExpectBody string `json:"expect_body,omitempty"`
 
+	transport http.RoundTripper `json:"transport,omitempty"`
```
mohammed90, line 136: It shouldn't have a json tag

modules/caddyhttp/reverseproxy/healthchecks.go
ab14-tech: updated

modules/caddyhttp/reverseproxy/healthchecks.go
ab14-tech: yes this makes sense. I updated the PR to emulate , but removed the requestBuffer / fastcgi bit as I don't think it's relevant.

ab14-tech: thanks for the detailed reply. (when you said 'plugin' I was thinking of an entirely separate way to configure and run a custom active health checker, but that is a separate conversation probably)

modules/caddyhttp/reverseproxy/healthchecks.go
ab14-tech: okay we can make it be Unexported. updated

francislavoie:
Closing as inactive

PR #6804, opened 2025-01-24
caddytls: Fix sni_regexp matcher to obtain layer4 contexts

vnxme:
This PR is the mainline part of the fix described [here](https://github.com/mholt/caddy-l4/issues/241#issuecomment-2610292874) for the issue with missing `{tls.regexp.*}` placeholders reported [here](https://github.com/mholt/caddy-l4/issues/241#issuecomment-2609215818).

### DIFF
--- a/modules/caddytls/matchers.go
+++ b/modules/caddytls/matchers.go
@@ -15,6 +15,7 @@
 package caddytls
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


mholt (changes requested):
Do you think we could refactor this and simplify it a bit by first trying to get a context, and then if we do, THEN get the replacer from it?

i.e. something like:

```go
var ctx context.Context
// try to set ctx
// if ctx is not nil, try to get replacer
// if replacer is nil, just use NewReplacer
```

mholt (approved):
Looks good!! Thanks!

PR #6809, opened 2025-01-25
feat: automatically set GOMEMLIMIT

dunglas:
Setting the [Go runtime memory limit](https://pkg.go.dev/runtime/debug#SetMemoryLimit) can improve significantly the performance of Caddy (and FrankenPHP and Mercure, according to our tests), especially on low-resources machines and containers.

This patch uses [github.com/KimMachineGun/automemlimit](https://github.com/KimMachineGun/automemlimit) to automatically find a good limit using the currently configured groups. This is especially useful when setting limits to containers using Docker, Kubernetes, LXC, and the like.

This PR is similar to #6047, which does the same things go the max number of processors.

Thanks to @flovntp for the idea and for finding this library!

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -75,7 +75,17 @@ func Main() {
 		caddy.Log().Warn("failed to set GOMAXPROCS", zap.Error(err))
 	}
 
-	_, _ = memlimit.SetGoMemLimitWithOpts(memlimit.WithLogger(slog.New(zapslog.NewHandler(caddy.Log().Core()))))
+	_, _ = memlimit.SetGoMemLimitWithOpts(
+		memlimit.WithLogger(
+			slog.New(zapslog.NewHandler(caddy.Log().Core())),
+		),
+		memlimit.WithProvider(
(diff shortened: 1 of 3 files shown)


mholt (changes requested):
Thanks Kevin! Do you think we could add a comment to this code (and maybe it could be shared with the gomaxprocs code above it) since I don't think it's obvious why it's useful? Then we will merge!

dunglas:
@mholt, done.

I also added logs for GOMAXPROCS to improve the overall experience. Commands now start with something like this:

```
2025/01/26 21:11:59.838 INFO    maxprocs: Leaving GOMAXPROCS=10: CPU quota undefined
2025/01/26 21:11:59.838 INFO    GOMEMLIMIT is updated   {"package": "github.com/KimMachineGun/automemlimit/memlimit", "GOMEMLIMIT": 15461882265, "previous": 9223372036854775807}
```

mholt (approved):
Great - thanks Kevin!

PR #6812, opened 2025-01-27
caddytls: Refactor sni matcher

vnxme:
This is a consistency fix to make `sni` matcher code in line with `sni_regexp` as adjusted by 7b8f3505e33139de0d542566478e98b361bb84bf. This way a new replacer is only created if no existing replacer/context has been found.


mholt (approved):
Makes sense -- thanks!

Issue #6822, opened 2025-02-02
Caddy API PUT array insertion doesn't allow inserting at the end of the array


johnbartholomew:
The Caddy API supports inserting an element at an index in an existing array. However, this fails if the provided index would place the item at the _end_ of the array (ie, append the item).

There is a range check on the index here:
https://github.com/caddyserver/caddy/blob/9996d6a70ba76a94dfc90548f25fbac0ce9da497/admin.go#L1135-L1145

The range check makes sense for GET/DELETE/PATCH, but for PUT it would be useful to allow `idx` to be equal to `len(arr)`. That allows for fewer special cases when using the API.

Example use case:

I have a base config which contains
```
              "handle": [
                {
                  "@id": "dynamic_routes",
                  "handler": "subroute",
                  "routes": []
                },
              ],
```

I would like to insert dynamic routes at the start of the `routes` array inside `@dynamic_routes`. I think natural way to do this is `PUT /id/dynamic_routes/routes/0`. But this fails when the array starts out empty. So to always insert at the front of the array I need to first retrieve the current content, check if the array is empty, and if it is then use a POST to append the new item, but if it's not empty then use PUT to item index 0.

mholt:
Hm, okay, yeah that can probably be done as long as the index is exactly 1 more than the length, but no more.

Anyone is welcome to tackle this issue :+1:

debug-ing:
I will check and correct this.

debug-ing:
cc: @mholt

debug-ing:
Hi
Please close this.
Thanks

PR #6824, opened 2025-02-03
fix: fix index validation for PUT requests

debug-ing:
fix issues #6822

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -1139,7 +1139,7 @@ traverseLoop:
 						return fmt.Errorf("[%s] invalid array index '%s': %v",
 							path, idxStr, err)
 					}
-					if idx < 0 || idx >= len(arr) {
+					if idx < 0 || (method != http.MethodPut && idx >= len(arr)) || idx > len(arr) {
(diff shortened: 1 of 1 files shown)


mholt (commented):
Thanks. Did you double-check that the PUT logic below is still correct if `idx == len(arr)`?

```go
case http.MethodPut:
// avoid creation of new slice and a second copy (see
// https://github.com/golang/go/wiki/SliceTricks#insert)
arr = append(arr, nil)
copy(arr[idx+1:], arr[idx:])
arr[idx] = val
v[part] = arr
```

debug-ing:
Thanks for your review. I've confirmed that the PUT logic is correct when idx == len(arr) and properly handles appending the value to the array.

debug-ing:
Yes, the PUT logic works correctly when idx == len(arr). It behaves as expected by appending the new element at the end of the array.

debug-ing:
the code is all good—every test is passing except this one CI run. It hit a network issue where the self-hosted runner couldn't reach GitHub’s endpoints during the artifact upload.

mholt:
Yeah, we have some known flaky tests. Thanks!

Issue #6828, opened 2025-02-05
file_server's file list uses wrong timezone causing cache to fail to be invalidated


rkfg:
Running on Debian testing, `caddy v2.9.1 h1:OEYiZ7DbCzAWVb6TNEkjRcSCRGHVoZsJinoDR/n9oaY=`.

Caddyfile (slightly modified from the Debian version):
```
{
        debug
}

:80 {
        handle_path /files/* {
                root * /home/rkfg/test
                file_server {
                        browse
                }
        }
        # Set this path to your site's directory.
        root * /usr/share/caddy

        # Enable the static file server.
        file_server
}
```

The file list returns 304 Not modified even if the file list was, in fact, modified. This is a regression imo because in Debian's repository version 2.6.2 this doesn't happen and Caddy returns 200 OK on every request. Moreover, this caching behavior can not be disabled. Steps to reproduce:

1. create some static files in `/home/rkfg/test`, for example, `new_file` and `new_file_2`
2. open the file list (in my case that's http://127.0.0.1/files/)
3. the files are there
4. rename `new_file_2` to `new_file_3`
5. refresh the tab
6. the file still has the old name and returns 404 on access
7. refresh with Ctrl+F5
8. now it has the correct name
9. delete `new_file_3` and refresh the tab
10. the file is still displayed until you refresh with no-cache again (Ctrl+F5)
11. create a new file, say, `new_file_5` and refresh the tab
12. surprisingly enough, this is the only case when Caddy detects the changes and displays the actual directory contents

So, either it should detect these changes, or always return 200 no matter what, or has it documented and provide an option to revert to the old non-caching behavior.

mholt:
Thanks for opening an issue!

Caddy doesn't do any caching of the file listing. It could be your browser aggressively caching. If the behavior still happens with `curl`, then maybe we will look into it; but it's more likely just a browser bug. (Closing, but can reopen if needed)

rkfg:
Okay, it seems that something's wrong with the timestamps Caddy returns. When I request the file list the first time it returns the `Last-modified` date in the future:
```
> curl -I http://127.0.0.1/files/
HTTP/1.1 200 OK
[...]

PR #6830, opened 2025-02-08
file_server: use the UTC timezone for modified time

WeidiDeng:
fix [6828](https://github.com/caddyserver/caddy/issues/6828)

Also should fix `If-Modified-Since` handling, but no one reported that yet.


mholt (approved):
Thank you Weidi, this is more nuanced than I had expected!

PR #6846, opened 2025-02-17
ci: allow using the toolchain Go "toolchain"

dunglas:
This is the first step to allow setting [the `toolchain` directive](https://go.dev/doc/toolchain) in `go.mod`. The patch also adds some missing build tags to speed up the builds.


mohammed90:
Since we now require 1.24 already, I wonder if we should drop the "last 2 versions" rule. I know this affects RHEL and CentOS because they tend to be 1 version behind.

mholt:
Yep; I just updated repo settings to require only the latest Go version tests to pass.

And we should probably remove Go 1.23 from our CI.

PR #6850, opened 2025-02-19
Support Caddy Test Configuration Defaults Override.

baruchyahalom:
While Running parallel Caddy testers, the hardcoded Admin port raised a requirement to allow overriding the default configuration.


caddytest/caddytest.go
```diff
@@ -78,9 +79,28 @@ func NewTester(t testing.TB) *Tester {
 		},
 		configLoaded: false,
 		t:            t,
+		config:       Default,
 	}
 }
 
+func (tc *Tester) WithDefaultOverrides(config Config) *Tester {
```
mholt, line 86: Can you add a godoc comment on this? And maybe rename `config Config` to `overrides Config` or something. :+1:

mholt (changes requested):
Thank you for the enhancement! Just a couple thoughts before we merge it.

mholt (approved):
Thanks -- I don't have much of an opinion on this code but if it's helpful, LGTM. :+1:

mholt:
Just a lint error to fix first it looks like

mholt (approved):

PR #6874, opened 2025-03-05
docs: replaced the name and twitter link

sashaphmn:


mohammed90:
I kinda want to keep it as Twitter....

sashaphmn:
Your right. Most of the projects already use the new name

sashaphmn:
X (formerly Twitter) can also be used like this

mholt (approved):
I lament the name change, but I think this is what will be recognizable in the future assuming they don't change it back.

sashaphmn:
That's how I understand you. But I consider it unlikely that the name will change back

sashaphmn:
i need to update the branch

PR #6884, opened 2025-03-07
Add error handling for unrecognized subdirective/options

### LINKED ISSUE #6863: Reject unrecognized directives
It's an oversight to miss having `default` in some Caddyfile unmarshallers in some modules. Here's an example where `default` is missing, and allowing invalid/false directives without much of a warning:

https://github.com/caddyserver/caddy/blob/fd4de7e0ae4debd54e2a9a4aba10947cd4c3b178/modules/logging/filewriter.go#L248C2-L317C3

We need to comb through the code to add `default` where it's missing. It'll be a breaking change, but it's a correction of behavior because whatever false directive the user used gives the false impression of doing something.

steffenbusch:
Fixes #6863


mohammed90 (approved):
Thank you! I played with the `exhaustive` linter config in golangci-lint, and it apparently cannot catch these cases. It can only identify them when the switch is over enums. Sigh.

Thanks for helping with this!

mholt:
Yep this is good, thank you!

PR #6889, opened 2025-03-10
Fix TrustedCACerts backwards compatibility

jjiang-stripe:
I'm very late to the party, but I was in the process of bumping our Caddy dependency and noticed https://github.com/caddyserver/caddy/pull/5784 has a slight bug with the backwards compatibility of the now-deprecated `TrustedCACerts` option. 

The Caddyfile bit works fine because [it converts it to a `CARaw`](https://github.com/jjiang-stripe/caddy/blob/b130e430a00a97db6381c0aa092eeff237d03d37/modules/caddytls/connpolicy.go#L701-L707), but if you specify the option directly via JSON, we don't end up [populating `icp.pool` since we're not calling `Provision()`](https://github.com/caddyserver/caddy/blob/d57ab215a2f198a465ea33abe4588bb5696e7abd/modules/caddytls/capools.go#L70)

I wrote up a quick test to validate the behaviour with the additional call to `Provision()` and also tested it on our instance of Caddy. It'd be nice to cherrypick this onto the 2.8.0+ releases (I'm specifically looking at 2.9.1) to make sure this option still works while it's deprecated! I plan to migrate us to the new config options, but our deploy process makes that a bit difficult to do simultaneously with the Caddy upgrade 😅

### DIFF
--- a/modules/caddytls/connpolicy.go
+++ b/modules/caddytls/connpolicy.go
@@ -749,10 +749,14 @@ func (clientauth *ClientAuthentication) provision(ctx caddy.Context) error {
 
 	// if we have TrustedCACerts explicitly set, create an 'inline' CA and return
 	if len(clientauth.TrustedCACerts) > 0 {
-		clientauth.ca = InlineCAPool{
+		caPool := InlineCAPool{
 			TrustedCACerts: clientauth.TrustedCACerts,
 		}
-		return nil
+		err := caPool.Provision(ctx)
+		if err != nil {
+			return nil
+		}
+		clientauth.ca = caPool
 	}
 
 	// if we don't have any CARaw set, there's not much work to do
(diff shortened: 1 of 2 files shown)


mholt (approved):
Oops, nice catch!! Thanks! I'll update your branch as well. (Just need a few minutes first.)

jzjiang:
I'm working on getting us pointed to 2.9.1 master, so hopefully we can 🔪 the special branch once I'm done :')

jjiang-stripe:
I'm working on getting us pointed to 2.9.1 master, so hopefully we can 🔪 the special branch once I'm done :')

mholt:
Okay, awesome! What I might do instead, to ensure you get the exact tree of commits you want, is to change the `stripe` branch to 2.9.1 and then cherry-pick this commit. Would that work okay for your team?

jzjiang:
yeah that'd work! I can point us back to master after I fix up our config

mholt:
@jjiang-stripe Perfect -- I just republished the `stripe` branch which is v2.9.1 with this commit cherry-picked.

jjiang-stripe:
amazing thank you!

Issue #6859, opened 2025-02-23
`caddy fmt` breaks template


Irt291:
Here is the output of `caddy fmt --diff` of the example config taken from [templates](https://caddyserver.com/docs/caddyfile/directives/templates).

```
  example.com {
        header Content-Type text/plain
        templates
-       respond `Current year is: {{now | date "2006"}}`
+       respond `Current year is: {now | date "2006"}}`
  }
```
Normal double-quote `"` works fine (```"Current year is: {{now | date `2006`}}"```). But with backquote ``` ` ```, `caddy fmt` will replace `{{}}` with `{}}`.

My guess is caddy's formatter only treats double-quote as "quoted" so anything inside backquotes will be formatted just like regular part of the file.

https://github.com/caddyserver/caddy/blob/fd4de7e0ae4debd54e2a9a4aba10947cd4c3b178/caddyconfig/caddyfile/formatter.go#L179-L189

caddy version: `v2.9.1 h1:OEYiZ7DbCzAWVb6TNEkjRcSCRGHVoZsJinoDR/n9oaY=`

mholt:
Thanks for the report. Been a while since I looked at that code, but your guess may likely be the reason! Busy atm, but will open this issue for anyone else to fix.

keystroke3:
I will take a crack at finding a solution here.

mohammed90:
Thank you, @keystroke3!

keystroke3:
@Irt291 I am done adding a patch to fix this issue, but I need to do more extensive testing. Can you please provide a sample of your config that raised this issue for you? Even though all tests are passing,  I just need to confirm that I have not broken anything else in the process and that the solution works for your specific example.

Irt291:
@keystroke3 Thank you for the patch, here is a sample of my config that causes the issue:

```
*.example.com {
	@ip host ip.example.com
	handle @ip {
		header Content-Type text/plain
		templates
		respond `Your IP is: {{.Req.Header.Get "CF-Connecting-IP"}}`
	}

	handle {
		respond 404 {
			body "Wrong Subdomain :(("
		}
	}
}
```

keystroke3:
I have tested it and it appears everything is working correctly. Now we just have to wait for the maintainers to merge the changes when they are able to.

mholt:
Thanks, really appreciate the contribution (and the report)!

PR #6903, opened 2025-03-16
Fix caddy fmt breaks backquote wrapped braces in template

keystroke3:
This addresses issue #6859 by writing the opening and closing braces and ignoring indentation when it detects that we the cursor is currently within a backquote section. 

I have added 2 new tests for the two cases where the braces are wrapped with regular quotes and backquotes. I have confirmed that it works at  least with the following example:
```
block {
	header Content-Type text/plain
	templates
	respond "Home page date: {{now | date `2006`}}"

	handle /path {
		header Content-Type text/plain
		templates
		respond `Path page date: {{now | date "2006"}}`
	}
}
```

### DIFF
--- a/caddyconfig/caddyfile/formatter.go
+++ b/caddyconfig/caddyfile/formatter.go
@@ -61,7 +61,8 @@ func Format(input []byte) []byte {
 		heredocMarker        []rune
 		heredocClosingMarker []rune
 
-		nesting int // indentation level
+		nesting         int // indentation level
+		withinBackquote bool
 	)
 
 	write := func(ch rune) {
@@ -87,6 +88,9 @@ func Format(input []byte) []byte {
 				break
 			}
 			panic(err)
+		}
+		if ch == '`' {
+			withinBackquote = !withinBackquote
 		}
 
 		// detect whether we have the start of a heredoc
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thank you for the patch!

lowlevl:
Hi there, I had some pixel art in my _Caddyfile_ in a `respond` directive in `v2.9.1`, that seemed to trigger an issue with `caddy fmt` when updating to `v2.10.0`.

I think this patch might be the cause, here is a minimal reproduction of the issue:
- _Caddyfile_ after `caddy fmt --overwrite` on **v2.9.1**:
```
{
  auto_https off
}

:8080 {
	respond "`"
}
```
A `caddy run` will run an HTTP server and return a single backtick upon `curl 127.0.0.1:8080`.
- _Caddyfile_ after `caddy fmt --overwrite` on **v2.10.0**:
```
{
	auto_https off
}

:8080 {
	respond "`"}
```
A `caddy run` will fail with: *Error: adapting config using caddyfile: Unexpected '}' because no matching opening brace, at Caddyfile:6*.

Thanks, I would post a separate issue if you prefered. :)

francislavoie:
Thanks. When in `"`mode, it should ignore backticks from counting.

keystroke3:
Is this an issue with the formatter or is the provided config unorthodox? If it is the former, I can have a look at it.

francislavoie:
Seems like a formatter issue to me. It's valid to have backticks inside of double quotes or double quotes inside of backticks interchangeably.

PR #6920, opened 2025-03-22
* feat: add support for base64 encoded client certificate chain

xz64:


francislavoie:
Sorry this took so long to review.

I'm not sure I agree with the placeholder using a newline as delimiter in the chain. The problem is that if someone tries to use this in an HTTP header to pass the chain from Caddy to their proxied app, the newline is awkward.

In fact, there's an RFC that defines how we should probably do this, [RFC 9440](https://datatracker.ietf.org/doc/rfc9440/)

```
Client-Cert: :MIIBqDCCAU6gAwIBAgIBBzAKBggqhkjOPQQDAjA6MR...hk=:
Client-Cert-Chain: :base64cert1:, :base64cert2:, :base64cert3:
```

In other words, the header value for `Client-Cert` base64-der, wrapped with `:` on either side, and `Client-Cert-Chain` uses the same format for each cert, with comma+space delimiter between each. Also, the chain does _not_ include the leaf, only the intermediates.

I think we should probably design this such that adding these headers to the request is as easy as possible.

francislavoie:
I'll close this as inactive for now, but we can reopen, or you can make a new PR if you'd like to take a shot at addressing the feedback. Thanks!

PR #6926, opened 2025-03-26
Add a pluggable "listener" for detecting when caddy starts and stops

ankon:
_This is a super-early draft to illustrate the idea._

We want to control the load balancer that sends traffic to a caddy instance, and basically inform it when caddyhttp is "up and ready" that it can start sending traffic, and similarly take the caddy instance out of the rotation BEFORE caddyhttp starts shutting down.

The existing shutdown delay mechanism doesn't work for us: The load balancer uses TCP health checks, and we need to keep processing requests until it stops sending them.

With this listener approach we are able to run the code to register/deregister the caddy instance from the load balancer.

In our context we do not use dynamic configuration reloading, so we know that the listener invocations are the real "start&stop". In the listener an implementation could look at the caddy context to learn about this, but I think it would be good to indicate a "start reason" and "stop reason" as well.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -43,6 +43,14 @@ import (
 	"github.com/caddyserver/caddy/v2/notify"
 )
 
+// Listener is a type that gets notified of certain events in the Caddy lifecycle.
+// Listeners are invoked synchronously and can block/delay the startup or shutdown, but they
+// cannot fail.
+type Listener interface {
+	AfterAppsStarted()
+	BeforeAppsStopping()
+}
+
 // Config is the top (or beginning) of the Caddy configuration structure.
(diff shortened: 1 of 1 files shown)


mholt (changes requested):
Ah I see -- I have a few thoughts!

Let's just start with the first one -- can this be implemented with events? Modules can implement event handlers, and I think that would be the most "native" way to accommodate this.

ankon:
Events indeed should work, as long as as we can have one that allows us to delay the shutdown of the caddy apps -- which should be possible as events are synchronously handled.

ankon:
Closing in favor of #6930

PR #6930, opened 2025-03-28
events: Refactor; move Event into core, so core can emit events

mholt:
**Breaking change for modules that use events. I suspect this will affect very few code bases.** Events are also documented as still experimental, and for good reason! This allows us to do much more with events and make them more useful.

Requires some slight trickery to invert dependencies. We can't have the caddy package import the caddyevents package, because caddyevents imports caddy. Interface to the rescue!

Also add two new events, experimentally: started, and stopping. At the request of a sponsor.

Also rename "Filesystems" to "FileSystems" to match Go convention (unrelated to events, was just bugging me when I noticed it).

/cc @ankon

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -81,13 +81,14 @@ type Config struct {
 	// associated value.
 	AppsRaw ModuleMap `json:"apps,omitempty" caddy:"namespace="`
 
-	apps    map[string]App
-	storage certmagic.Storage
+	apps         map[string]App
+	storage      certmagic.Storage
+	eventEmitter eventEmitter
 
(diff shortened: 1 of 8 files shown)


mholt:
Huh, that's interesting.

Anyway, I'll probably merge this now and fix a potential bug with logging in another patch.

francislavoie:
Events are only pushed via this event interface (Go level APIs), so you can use a plugin or write your own to do something with the events. See here https://caddyserver.com/docs/caddyfile/options#event-options

mholt:
Beat me to the answer. Thanks @francislavoie !

PR #6941, opened 2025-04-07
caddytls: Regularly reload static certificates

### LINKED ISSUE #6933: Automatic reload of TLS certificates from filesystem
We're trying to use [Azure Key Vault provider for Secrets Store CSI Driver](https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-service-connector#verify-the-connection) in Kubernetes, which basically mounts the TLS certificates from a KeyVault into the filesystem, e.g. as `/etc/tls-certificates/example.com.crt`. The CSI Driver automatically updates the file when the entry in the KeyVault changes, but Caddy is not reloading the changes.

Would it make sense to change Caddy to use something like https://github.com/fsnotify/fsnotify to automatically reload TLS files when they change? If not, would it be possible to implement this as a plugin?

Related:

- https://github.com/caddyserver/caddy/issues/4005
- https://github.com/caddyserver/caddy/issues/6789
- https://caddy.community/t/force-reload-certificates/27545

pascalgn:
Fixes #6933

### DIFF
--- a/modules/caddytls/tls.go
+++ b/modules/caddytls/tls.go
@@ -139,6 +139,8 @@ type TLS struct {
 	storageCleanStop   chan struct{}
 	logger             *zap.Logger
 	events             *caddyevents.App
+	magic                *certmagic.Config
+	unmanagedCertsTicker *time.Ticker
 
 	serverNames   map[string]struct{}
 	serverNamesMu *sync.Mutex
@@ -249,7 +251,7 @@ func (t *TLS) Provision(ctx caddy.Context) error {
 	// certificates have been manually loaded, and also so that
 	// commands like validate can be a better test
 	certCacheMu.RLock()
-	magic := certmagic.New(certCache, certmagic.Config{
+	t.magic = certmagic.New(certCache, certmagic.Config{
 		Storage: ctx.Storage(),
 		Logger:  t.logger,
 		OnEvent: t.onEvent,
@@ -259,18 +261,14 @@ func (t *TLS) Provision(ctx caddy.Context) error {
 		DisableStorageCheck: t.DisableStorageCheck,
 	})
 	certCacheMu.RUnlock()
-	for _, loader := range t.certificateLoaders {
-		certs, err := loader.LoadCertificates()
-		if err != nil {
-			return fmt.Errorf("loading certificates: %v", err)
-		}
-		for _, cert := range certs {
-			hash, err := magic.CacheUnmanagedTLSCertificate(ctx, cert.Certificate, cert.Tags)
-			if err != nil {
-				return fmt.Errorf("caching unmanaged certificate: %v", err)
-			}
-			t.loaded[hash] = ""
-		}
+
+	unmanaged, err := t.loadUnmanagedCertificates(ctx)
+	if err != nil {
+		return err
+	}
+
+	if unmanaged > 0 {
+		t.regularlyReloadUnmanagedCertificates()
 	}
 
 	// on-demand permission module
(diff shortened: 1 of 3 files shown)


modules/caddytls/tls.go
```diff
@@ -122,13 +122,15 @@ type TLS struct {
...
-	events             *caddyevents.App
+	magic                *certmagic.Config
+	certificateLoaders   []CertificateLoader
+	unmanagedCertsTicker *time.Ticker
+	automateNames        []string
+	ctx                  caddy.Context
+	storageCleanTicker   *time.Ticker
+	storageCleanStop     chan struct{}
+	logger               *zap.Logger
+	events               *caddyevents.App
```
pascalgn, line 133: Is it fine to add magic as a new field here? 🤔

modules/caddytls/tls.go
```diff
@@ -248,18 +250,14 @@ func (t *TLS) Provision(ctx caddy.Context) error {
...
-		}
+
+	unmanaged, err := t.loadUnmanagedCertificates(ctx)
+	if err != nil {
+		return err
+	}
+
+	if unmanaged > 0 {
+		t.regularlyReloadUnmanagedCertificates()
 	}
```
pascalgn, line 261: I think this reloading should always run (if explicitly enabled via config), independent of loaded certs

modules/caddytls/tls.go
```diff
@@ -720,6 +718,61 @@ func (t *TLS) HasCertificateForSubject(subject string) bool {
...
+				return 0, fmt.Errorf("caching unmanaged certificate: %v", err)
+			}
+			t.loaded[hash] = ""
+		}
+	}
+	return cached, nil
+}
+
+func (t *TLS) regularlyReloadUnmanagedCertificates() {
+	t.unmanagedCertsTicker = time.NewTicker(2 * time.Hour)
```
pascalgn, line 740: This will be configurable, same as StorageCleanInterval, but it needs to be explicitly enabled in the config. It should probably be disabled by default, right?

modules/caddytls/tls.go
```diff
@@ -720,6 +718,61 @@ func (t *TLS) HasCertificateForSubject(subject string) bool {
...
+		for _, cert := range certs {
+			hash, err := t.magic.CacheUnmanagedTLSCertificate(t.ctx, cert.Certificate, cert.Tags)
+			if err != nil {
+				return fmt.Errorf("caching unmanaged certificate: %v", err)
+			}
+			t.loaded[hash] = ""
+		}
+	}
+	return nil
+}
```
pascalgn, line 774: We could also try to remove expired certs from the cache in this function

modules/caddytls/tls.go
mholt: Hmm, likely not, since certmagic configs are mostly ephemeral and not tied to an entire TLS app config.

mholt (changes requested):
Thanks for taking the initiative on this!

I haven't had much time to give this some thought, but, I think there's several ways to do this. Note that cert loader modules don't _necessarily_ get their certificates from disk. I think if we are going to load certs from disk, we should be careful to initiate this only from modules which load from disk.

I was kind of thinking of something that watched the file system (`fsnotify`, I guess?) for changes, and then reloads only in that case. But yes, as you mentioned in a comment, we'd also need to remove the old cert from the cache. It's not my favorite, but it'd be opt-in I guess.

The ticker approach isn't my favorite either: it does a lot of unnecessary work. It's also not as effective... if a cert is updated, it might not be reflected in Caddy for up to the ticker interval. i.e. it's kind of slow. And then it has to load the cert every time, since you're not sure if it changed or not until you load it and check. (Well, I guess you could `stat` it, but still... almost all stats will end up being unnecessary.)

So I almost wonder if an approach that uses `fsnotify` in the relevant cert loader modules' `Provision()` method (and probably also cleaning things up in `Cleanup()`) would be a more conventional approach? What do you think?

modules/caddytls/tls.go
pascalgn: Got it! There isn't any state (other than certCache) stored in it, so in the reloadUnmanagedCertificates function, I can simply create a new object, right?

```go
func reloadUnmanagedCertificates() {
	certCacheMu.RLock()
	magic := certmagic.New(certCache, certmagic.Config{
		Storage: ctx.Storage(),
		Logger:  t.logger,
		OnEvent: t.onEvent,
		OCSP: certmagic.OCSPConfig{
			DisableStapling: t.DisableOCSPStapling,
		},
		DisableStorageCheck: t.DisableStorageCheck,
	})
	certCacheMu.RUnlock()
	// add/remove certs from magic...
}
```

pascalgn:
@mholt I pushed some changes. I changed CertificateLoader like this:
```go
type CertificateLoader interface {
	Initialize(updateCertificates func(add []Certificate, remove []string) error) error
}
```
That way, existing loaders (e.g. StorageLoader) will simply call the updateCertificates callback during the Initialize method, but other loaders (i.e. FolderLoader) will call it during Initialize, but also afterwards, during fsnotify events.

Would it make sense like that?

mholt:
I was thinking more of using each loader modules' `Provision()` and `Cleanup()` methods to start/stop a goroutine that does the watching and reloading.

As for removing... let me think more on that.

pascalgn:
When moving the code into the Provision method, how would it access the certCache?

mholt:
Via the `tls` app, so, `ctx.App("tls")` (then type-assert).

We might need to add a method or two to the `caddytls.TLS` app type which can expose the cache or at least the operations we need on it: https://pkg.go.dev/github.com/caddyserver/caddy/v2/modules/caddytls#TLS

francislavoie:
@pascalgn any interest in finishing this off?

zhenyatsk:
bump.
need this feature

pascalgn:
@francislavoie Yes, sorry, still interested, just not had the time yet. But I will try to have a look at it again in the next few weeks! 🙏

mholt:
Looking at this again. I'm a bit worried about changing the CertificateLoader interface like that. It might be a painful breaking change. Can we come up with a way to do this that doesn't breaking the existing API?

pascalgn:
Yes, I also don't think changing the CertificateLoader interface is a good approach. My plan is to get `ctx.App("tls")` in the loader's Provision method and use that to manage the certs. But I really have no preference here, so if you have any other suggestions, I'm happy to hear them!

mholt:
Would be curious to see what that approach looks like!

ryan0x44:
@pascalgn are you still interested in drafting that approach? Or @mholt any further thoughts or suggestions for how to best move this forward?

mholt:
I'm still very mixed on this. Especially in a future quickly approaching where certificates _need_ to be short-lived and in the PQ era, we'll have a much more complex situation for certificates that will make the simple "load a certificate from disk every so often" feasible (see https://www.feistyduck.com/newsletter/issue_139_the_state_of_post-quantum_cryptography).

Where we're going, I just don't think this feature can reliably, sustainably go.

ryan0x44:
@mholt ok so would a reasonable path forward be:

- keep CertificateLoader unchanged
- have FolderLoader watch the configured cert/key files with fsnotify
- reload the certificate when either file changes
- start the watcher in Provision() and stop it in Cleanup()

? Trying to figure out how to get the PR to move forward without introducing a new callback/API.

PR #6949, opened 2025-04-11
Add a valid Upstream to the DialInfo when doing active health checks

jbro:
Currently if we extract the `DialInfo` from a `Request` `Context` in a custom `Transport` during an active health check, then the `Upstream` in the `DialInfo` will be nil.

This PR attempts to set the `Upstream` to a sensible value, based on wether or not the `Upstream` has been overridden in the active health check's config.

This is useful when using a custom Transport for the `reverse_proxy` module, as it makes `GetDialInfo` behave similarly in active health check requests and client requests.


mholt (approved):
Thanks! I think this makes sense, though it's been a while since I've been working with this code. We'll give it a shot and let you know if we have any issues with it :)

PR #6951, opened 2025-04-12
bug: Fix the incorrect parameter order

cuishuang:
According to the documentation for  [strings.HasPrefix](https://pkg.go.dev/strings#HasPrefix), the order of the parameters here should be incorrect.


mholt (approved):
Ah yes, thank you

PR #6955, opened 2025-04-16
reverseproxy: use DialTLSContext for TLS if servername has placeholder

WeidiDeng:
This way when using placeholders to establish tls connections, the performance can be significantly improved.

Also reduced the amount of codes that replicate the transport.


mholt:
Thanks for this!

Does this reduce any functionality though? It uses a single static transport -- definitely more efficient -- but less dynamic, right?

WeidiDeng:
The only dynamic part that I can think of is the server name placeholder part, which this patch takes care of. I haven't found any thing broken in my setup.

mholt (approved):
I think we can give this a shot then. Thanks for the contribution, and if there's any reported issues, hopefully you can help us with them :sweat_smile:

PR #6961, opened 2025-04-17
caddyhttp: use the new http.Protocols to handle h1, h2 and h2c requests

WeidiDeng:
Golang 1.24 introduced a new type that can be used to specify the versions of http caddy can handle. Now caddy can support h2/h2c without h1.

This new mechanism also means graceful shutdown for h2 is easier and we no longer need to track the number of h2 connections. It does require another hack to prevent unintentional h2c request handling, but that hack is minimal.

H2c via protocol upgrade is no longer supported. As far as I know, only curl with use this upgrade if `--http2-prior-knowledge` is not specified.


francislavoie:
Can you link the relevant Go PRs/commits? Curious to see how that evolved.

modules/caddyhttp/app.go
```diff
@@ -486,7 +460,24 @@ func (app *App) Start() error {
...
+			srv.server.Protocols.SetHTTP1(true)
+		}
+
+		if srv.protocol("h2") || srv.protocol("h2c") {
+			// skip setting h2 because if NextProtos is present, it's list of alpn versions will take precedence.
+			// it will always be present because http2.ConfigureServer will populate that field
+			// enabling h2c because some listener wrapper will wrap the connection that is no longer *tls.Conn
+			// However, we need to handle the case that if the connection is h2c but h2c is not enabled. We identify
+			// this type of connection by checking if it's behind a TLS listener wrapper or if it implements tls.ConnectionState.
+			srv.server.Protocols.SetUnencryptedHTTP2(true)
```
francislavoie, line 476: Shouldn't this only be set if `h2c`, but not `h2`?

WeidiDeng: This is deliberate. if the handshake succeeds with `h2`, it will be handled accordingly. `h2c` is enabled to handle listener wrappers.

modules/caddyhttp/server.go
```diff
@@ -264,18 +263,6 @@ type Server struct {
 
 // ServeHTTP is the entry point for all HTTP requests.
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// If there are listener wrappers that process tls connections but don't return a *tls.Conn, this field will be nil.
-	// TODO: Can be removed if https://github.com/golang/go/pull/56110 is ever merged.
```
francislavoie, line 268: Whoa, 3 years later, it's finally in! 👏 Is it too early to drop this though, isn't it only in Go 1.25?

francislavoie:
Is there anything in reverseproxy that can be simplified with the new h2c stuff? Or did they only improve the server side but not client side?

modules/caddyhttp/server.go
WeidiDeng: That's why I leave this as a draft. I have yet to test this thoroughly.

WeidiDeng: Yeah, this should be left in for now. Other parts seem fine though.

WeidiDeng:
I want to wait until [6955](https://github.com/caddyserver/caddy/pull/6955) is merged to work on the reverse proxy protocol stuff.

modules/caddyhttp/app.go
francislavoie: But if the user only configured `h2` and _not_ `h2c`, it'll still have `SetUnencryptedHTTP2(true)`, right? That doesn't seem correct.

francislavoie: Or wait, you mean if there's LWs then it could appear as though the conn is no longer encrypted cause it's already been decrypted? Is that what you mean? Tricky.

mholt:
This is still in draft state; would you like another review?

PR #6972, opened 2025-04-20
update quic-go to v0.51.0

marten-seemann:


mholt (approved):
Thanks Marten! Nice work with quic-go, as always :)

PR #6983, opened 2025-04-23
feat: add static filename and info served by fileserver to context vars

arthurpro:
Right now it is impossible to get any information about the file served by fileserver from other middlewares.
I'm proposing this tiny addition to fileserver module to put info about a static file served by it into the request context.


mholt:
Thanks for the PR; I'm on the go and unable to check right now, but don't we set placeholders indicating the file that was served? (Or is that only in the file matcher, that we set placeholders of what was matched?)

francislavoie:
I think we're open to discussing this further, but this fell inactive without a reply to the questions, so I'll close it for now.

PR #6990, opened 2025-04-29
reverse proxy: use http.Protocols to handle h2c requests

WeidiDeng:
Like [6961](https://github.com/caddyserver/caddy/pull/6961), this pr uses the new [`http.Protocols`](https://pkg.go.dev/net/http#Protocols) to handle h2c requests.


mholt:
Thanks; looking forward to this. This is still in draft state, should I wait to review it?

WeidiDeng:
Let's just wait until go 1.25 is out, because `http.Protocols` is introduced in 1.24

mholt:
Sounds good, thanks!

francislavoie:
Since we're on 1.25, this should be good to go now?

Issue #6931, opened 2025-03-29
Empty reply from `/metrics` on secure admin listener


sushain97:
I'm running `caddy:2.9.1-alpine` with a JSON config. Accessing the `/metrics` endpoint over the secure admin listener returns an empty response. I have enabled metrics. This used to work but I suspect it broke with an upgrade and I didn't notice until now :(

It works over the plaintext listener:
```
/srv $ http_proxy= wget -O- http://127.0.0.1:2019/metrics | head
Connecting to 127.0.0.1:2019 (127.0.0.1:2019)
writing to stdout
# HELP caddy_admin_http_requests_total Counter of requests made to the Admin API's HTTP endpoints.
# TYPE caddy_admin_http_requests_total counter
caddy_admin_http_requests_total{code="200",handler="admin",method="GET",path="/metrics"} 2
caddy_admin_http_requests_total{code="200",handler="admin",method="GET",path="/reverse_proxy/upstreams"} 273
# HELP caddy_config_last_reload_success_timestamp_seconds Timestamp of the last successful configuration reload.
# TYPE caddy_config_last_reload_success_timestamp_seconds gauge
caddy_config_last_reload_success_timestamp_seconds 1.7432255318608303e+09
# HELP caddy_config_last_reload_successful Whether the last configuration reload attempt was successful.
# TYPE caddy_config_last_reload_successful gauge
caddy_config_last_reload_successful 1
```

Other requests work fine over the secure listener:
```
sushain@vesuvianite ~ ❯❯❯ curl -fsS --cert /tmp/cert --key /tmp/key --resolve web.local.skc.name:8001:127.0.0.1 https://web.local.skc.name:8001/reverse_proxy/upstreams | jq length
27
```

But, this doesn't:
```
sushain@vesuvianite ~ ❯❯❯ curl -v -fsS --cert /tmp/cert --key /tmp/key --resolve web.local.skc.name:8001:127.0.0.1 https://web.local.skc.name:8001/metrics
* Added web.local.skc.name:8001:127.0.0.1 to DNS cache
* Hostname web.local.skc.name was found in DNS cache
*   Trying 127.0.0.1:8001...
* Connected to web.local.skc.name (127.0.0.1) port 8001
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
[...]

PR #6997, opened 2025-05-04
admin: Make sure that any admin routers are provisioned when local/re…

Compy:
…mote admin servers are replaced at runtime.

Fixes issue #6931 

**Background**

When the remote admin servers are reloaded at runtime (or as part of the final initialization phase at boot), the local state in the new instances aren't fully initialized. This is due to `New()`ing each module up during the call to `newAdminHandler`, but never actually provisioning it. As a result, the `metricsHandler` and `registry` members are `nil`, causing the metrics endpoint to crash and return an empty response when invoked via the remote admin API.

This change makes sure that we call the `provisionAdminRouters` method when both local and remote admin servers are reloaded and replaced. This should ensure consistent behavior in the future between reloads.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -426,7 +426,10 @@ func replaceLocalAdminServer(cfg *Config, ctx Context) error {
 
 	// run the provisioners for loaded modules to make sure local
 	// state is properly re-initialized in the new admin server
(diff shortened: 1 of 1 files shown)


mholt (approved):
Ah thanks, this is probably more correct :)

Issue #6957, opened 2025-04-16
handle inside handle_errors does not work if handle_errors <status code> is used


bludevsk:
I tried to create language dependent custom 404 error pages with the following configuration:

```
handle_errors 404 {
    handle /de/* {
        respond "404 Fehler" 404
    }
    handle /en/* {
        respond "404 Error" 404
    }
    handle {
        respond "404 Fehler" 404
    }
}
```
But this does not work. It will always evaluate the directives in the first `handle ` block. I also tried named matchers with different matchers (path, path_regexp, file) it always runs the first `handle ` block.

I solved my problem by using the following slightly more complex version which works as excpected:
```
handle_errors {
    @404 `{err.status_code} == 404`
    handle @404 {
        handle /de/* {
            respond "404 Fehler" 404
        }
        handle /en/* {
            respond "404 Error" 404
        }
        handle {
            respond "404 Fehler" 404
        }
    }
}
```

The problem was consistent reporducable on 
OS: Debian Buster
Caddy version: v2.9.0 h1:rteY8N18LsQn+2KVk6R10Vg/AlNsID1N/Ek9JLjm2yE=
and 
OS: Ubuntu 24.04 (WSL)
Caddy version: v2.9.1 h1:OEYiZ7DbCzAWVb6TNEkjRcSCRGHVoZsJinoDR/n9oaY=

francislavoie:
If you adapt to JSON with `caddy adapt --pretty`, what do you get in either case? Compare that. My hunch is it's an issue with the `group` on the route (which is what makes routes mutually exclusive).

bludevsk:
If I use adapt to create the JSON the problem is obvious. You can see that the matcher of the inner handle is ignored an it is always the 404 matcher.

Caddyfile:
```
test.localhost {
	root * /var/www/
	file_server

	handle_errors 404 {
		@de path_regexp ^/de/.*
		handle @de {
			respond "DE" 404
		}
		@en path_regexp ^/en/.*
		handle @en {
			respond "EN" 404
		}
		handle {
			respond "Default" 404
		}
	}
}
```

caddy adapt --config Caddyfile --pretty

```
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"match": [
								{
									"host": [
										"test.localhost"
									]
								}
							],
							"handle": [
								{
									"handler": "subroute",
									"routes": [
										{
											"handle": [
												{
[...]

PR #6999, opened 2025-05-04
fix: prevent error handler from overriding sub handler matchers

Hellio404:
Fixes: #6957

It make sure that all routes within the handle_errors are within a subroute.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -15,6 +15,7 @@
 package httpcaddyfile
 
 import (
+	"encoding/json"
 	"fmt"
 	"html"
 	"net/http"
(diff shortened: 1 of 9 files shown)


mholt:
Thanks! Is this really the correct fix though? I imagine there's a JSON structure we can generate that isn't this deep that does it correctly, but I'm not super familiar with the issue right now to know what that would be.

Hellio404:
I tried to have something that is similar to what the normal routes already have as per documentation

so I see it as something like this: (Not a real example)
```go
@status_404 {
  status 404
}
handle @status_404 {
  handle /en/* { ... }
   handle /es/* { ... }
}
```
so the above code if we use Method matcher or any other matcher, it will result in the same depth as what this PR introduced

Hellio404:
To summerize the issue, it is mainly due to the following line:
https://github.com/caddyserver/caddy/blob/9f7148392adb72a6121bf99070efaa1a90ffe901/caddyconfig/httpcaddyfile/builtins.go#L850-L852

it was overriding the `MatcherSetsRaw` for all the handlers inside of the `handle_errors` block
we mostly had tests with only direct `static_response` routes which has no `MatcherSetsRaw`

Involture:
Hi,
Thank you very much for writing the fix.
I need it as well. I hope it is reviewed soon.

mholt (approved):
Sounds good -- thank you for working on this. We'll give it a shot!

Issue #6991, opened 2025-04-30
Add the uniqueness judgment of id


Chise1:
Perform a uniqueness check on the passed id parameter. If it exists, an error will be reported.

francislavoie:
?????

Chise1:
@francislavoie  like this:

`

{
  "@id": "route4",
  "match": [{"path_regexp": {"name": "path5","pattern": "path5"}}],
  "handle": [
    {
      "@id": "xx",
      ...
    },
    {
      "@id": "xx",
      ...
    }
  ]
}

`
I think it  is  a bad configuration.However, caddy did not check whether @id was unique.

francislavoie:
I see. Yes, that's a valid complaint.

Please use code formatting with triple backticks ``` or click the `<>` button to insert a code block.

Chise1:
I submitted a pr. This is my first time submitting a pr. If there is anything that needs to be modified, please let me know. If you think my code is not good, stop it directly.

PR #7002, opened 2025-05-07
caddy.go: Check whether @id is unique(#6991)

Chise1:
Fixes: https://github.com/caddyserver/caddy/issues/6991

Check whether @id is unique.If it is not unique,rollback caddy config json.
I'm not sure if it can work properly under caddyfile.

### DIFF
--- a/caddy.go
+++ b/caddy.go
@@ -224,6 +224,10 @@ func changeConfig(method, path string, input []byte, ifMatchHeader string, force
 	idx := make(map[string]string)
 	err = indexConfigObjects(rawCfg[rawConfigKey], "/"+rawConfigKey, idx)
 	if err != nil {
+		err2 := rollbackRawCfg()
+		if err2 != nil {
+			err = errors.Join(err, err2)
+		}
 		return APIError{
 			HTTPStatus: http.StatusInternalServerError,
 			Err:        fmt.Errorf("indexing config: %v", err),
@@ -239,12 +243,10 @@ func changeConfig(method, path string, input []byte, ifMatchHeader string, force
 			// with what caddy is still running; we need to
 			// unmarshal it again because it's likely that
 			// pointers deep in our rawCfg map were modified
-			var oldCfg any
-			err2 := json.Unmarshal(rawCfgJSON, &oldCfg)
+			err2 := rollbackRawCfg()
 			if err2 != nil {
 				err = fmt.Errorf("%v; additionally, restoring old config: %v", err, err2)
 			}
-			rawCfg[rawConfigKey] = oldCfg
 		}
 
 		return fmt.Errorf("loading new config: %v", err)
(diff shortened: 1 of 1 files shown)


caddy.go
```diff
@@ -224,6 +224,10 @@ func changeConfig(method, path string, input []byte, ifMatchHeader string, force
 	idx := make(map[string]string)
 	err = indexConfigObjects(rawCfg[rawConfigKey], "/"+rawConfigKey, idx)
 	if err != nil {
+		err2 := rollbackRawCfg()
+		if err2 != nil {
+			err = errors.Join(err, err2)
+		}
```
codehakase, line 230: Trying to parse the old config here without checking the length of `rawCfgJSON` isn't a recommended approach. Additionally, this is a potential duplicate of the call in `Ln246`

caddy.go
```diff
@@ -261,13 +263,28 @@ func changeConfig(method, path string, input []byte, ifMatchHeader string, force
 	return nil
 }
 
+func rollbackRawCfg() error {
+	var oldCfg any
+	err := json.Unmarshal(rawCfgJSON, &oldCfg)
+	rawCfg[rawConfigKey] = oldCfg
```
codehakase, line 269: obs: Potential nil assignment to `rawCfg`. Check for error before assigning to the map

Chise1:
I added length detection and fixed a stupid problem. From the perspective of code analysis, "err" will not be repeated.

Changes of rewrite, involves [nginx-adapter-case](:https://github.com/caddyserver/nginx-adapter/issues/35)

mohammed90:
> Changes of rewrite, involves nginx-adapter-case

Please keep the PR topical and focused. Split these changes into a separate PR.

Chise1:
It seems that there are many problems now. I need to think carefully again about how to modify it. Please ignore my pr for the moment

mholt:
I feel like a better approach would be to check the ID as it's being read in and we're creating the mapping of IDs to their full config paths. If we return an error the rest should be handled automatically (no need to roll back, etc)

Chise1:
> I feel like a better approach would be to check the ID as it's being read in and we're creating the mapping of IDs to their full config paths. If we return an error the rest should be handled automatically (no need to roll back, etc)

If we do as you said, the content that needs to be modified is rather complicated.  This is beyond my ability.  I'm worried that the modification will introduce bugs.

Chise1:
add one test.

francislavoie:
Since this fell inactive, I took it upon myself to try and reimplement this.

The tests were good, thanks! It was a good starting point.

I pointed Gemini 3.1 Pro + Github Copilot at this to take a shot at doing it more cleanly. I rebased and pushed my version on top. The result is quite a lot simpler. The new functions weren't really necessary.

mholt (approved):
Thanks Francis. Glad to merge this in.

PR #7006, opened 2025-05-08
acme_server: fix policy parsing in caddyfile

mohammed90:
Not sure where my mind was when I wrote those lines and why I didn't add tests. Anyways, fixed.

Reported here: https://caddy.community/t/acme-server-allow-domains/30973


mholt (approved):
Those are good tests, thank you!

And, my bad also for not catching it. Should have been obvious to me. Sorry. :man_facepalming:

PR #7009, opened 2025-05-10
refactor: use maps.Copy for cleaner map handling

eveneast:
Optimize code using a more modern writing style.


mholt (approved):
Nice, thank you!

francislavoie:
Does copy only add, or does it clear then add?

mholt:
It only assigns. It's the same as the loops. Clone will clear elements as needed.

Issue #7012, opened 2025-05-10
Routes Bug


fa1rid:
In the provided Caddyfile, all routes (Channels) behave the same except for Channel 20.
If I call `http://127.0.0.1:9991/ch20/` it supposed to rewrite to `/MyCh20/video.m3u8`, instead the `route /ch20/*` is being triggered.

[Caddyfile.txt](https://github.com/user-attachments/files/20140522/Caddyfile.txt)

francislavoie:
That's not a bug. Routes are sorted by their length, so `/ch20/*` is matched before `/ch20/`.

You need to rework your config to put your `/ch20/` rewrite inside of your `/ch20/*` route.

Also you can _greatly_ simplify that config by using pattern matching instead:

```caddyfile
:9991 {
    @channel_root path_regexp ^/ch([0-9]+)/$
    rewrite @channel_root /MyCh{re.1}/video.m3u8
    
    @channel_files path_regexp ^/ch([0-9]+)/(.*)$
    rewrite @channel_files /MyCh{re.1}/{re.2}

	reverse_proxy http://127.0.0.1
}
```

Please ask your usage questions on the [Caddy community forums](https://caddy.community). We prefer to keep the GitHub issue board for bugs and feature requests. Don't forget to fill out the thread template so we can help you!

fa1rid:
That's incorrect. If that's the case then other routes would behave the same, but only ch20 behaving opposite of the others. 
Also, if you remove `header -Server` the problem will be fixed.
Have you even tried it?

mholt:
I haven't had a chance to look into this yet but I can reopen this until it gets another look. I'll try to get to it soon unless someone beats me to it.

@fa1rid The ch20 routes are the ONLY ones that have the weird behavior? ch21 doesn't?

Can you verify with `curl -v` to speed things up (don't use a browser) -- since I don't have a chance right now.

fa1rid:
@mholt  thank you for looking into this.

Yes only ch20, ch21 and others don't have the problem. I was able confirm that by running an upstream server using NodeJS and log the request that comes from caddy's reverse proxy. Please let me know if you need any additional info.

Hellio404:
I think the sorting in this case is a bit unpredictable due to the violation the requirement of the Less than interface in the `sortRoutes` function.

from the golang sort doc:
```
// Less must describe a transitive ordering:
//  - if both Less(i, j) and Less(j, k) are true, then Less(i, k) must be true as well.
[...]

PR #7015, opened 2025-05-12
caddyhttp: fix route sort by comparing paths without wildcard if they don't shar…

WeidiDeng:
…e the same prefix

Fix [7012](https://github.com/caddyserver/caddy/issues/7012).

### DIFF
--- a/caddyconfig/httpcaddyfile/directives.go
+++ b/caddyconfig/httpcaddyfile/directives.go
@@ -483,11 +483,22 @@ func sortRoutes(routes []ConfigValue) {
 			// we can only confidently compare path lengths if both
 			// directives have a single path to match (issue #5037)
 			if iPathLen > 0 && jPathLen > 0 {
+				// trim the trailing wildcard if there is one
+				iPathTrimmed := strings.TrimSuffix(iPM[0], "*")
+				jPathTrimmed := strings.TrimSuffix(jPM[0], "*")
+
 				// if both paths are the same except for a trailing wildcard,
 				// sort by the shorter path first (which is more specific)
-				if strings.TrimSuffix(iPM[0], "*") == strings.TrimSuffix(jPM[0], "*") {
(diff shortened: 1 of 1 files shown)


francislavoie:
Nice. Could you add an adapt test for this?

Hellio404:
I think there is still a difference between the Docs and the implementation

From the doc: `/foo is more specific than /foo*`

but if we have a file like:
```caddyfile
:9991 {
    route /foo* {
    }
    route /fob* {
    }
    route /foo {
    }
}
```

the result of `adapt` will be:
```json
"routes": [
  {
	  "match": [
		  {
			  "path": [
				  "/foo*"
			  ]
		  }
	  ],
	  "handle": [
		  {
			  "handler": "subroute"
		  }
	  ]
  },
  {
	  "match": [
		  {
			  "path": [
				  "/fob*"
			  ]
		  }
	  ],
	  "handle": [
		  {
			  "handler": "subroute"
		  }
	  ]
  },
  {
	  "match": [
		  {
			  "path": [
				  "/foo"
			  ]
		  }
	  ],
	  "handle": [
		  {
			  "handler": "subroute"
		  }
	  ]
  }
]
```

so here we can see that the result is `/foo*` is more specific than `/foo`

WeidiDeng:
It's now fixed by ordering routes lexically if the paths have the same length.

mholt:
@Hellio404 Want to give it one more try and then we can merge it?

Thank you very much @WeidiDeng !!

Hellio404:
@mholt I did few tests now it is working as expected
Thank you @WeidiDeng

mholt (approved):
That's great! Thank you both :D

francislavoie:
I'd still like an adapt test to cover this to make sure we don't break it again in the future

mholt:
Yeah that'd probably be good. It's subtle code.

PR #7016, opened 2025-05-12
implement Unwrap for interceptedResponseHandler

WeidiDeng:
fix [`not a flusher`](https://github.com/dunglas/frankenphp/issues/1543) when `intercept` is used.


dunglas (approved):
LGTM, but it could be nice to add a test.

mholt (approved):
Thanks, seems like a good patch.

Issue #6992, opened 2025-05-01
`intercept` causing requests to be handled twice


mohammed90:
Update: see this [comment](https://github.com/caddyserver/caddy/issues/6992#issuecomment-2892620415) for the effect of this bug.

As the subject says, using the `intercept` handler with `reverse_proxy` causes duplication of headers received from upstream.

Using this Caddyfile:

```caddyfile
localhost {
	log
	intercept {
		handle_response {
			header {
				Header-Three Value-3
			}
		}
	}
	reverse_proxy localhost:8080
}

http://localhost:8080/ {
	header {
		Header-One Value-1
		Header-Two Value-2
	}
	respond "Hello, World!"
}
```

Note the headers received by the client:

```
$ curl -v https://localhost/
* Request completely sent off
< HTTP/2 200
< alt-svc: h3=":443"; ma=2592000
< content-type: text/plain; charset=utf-8
< content-type: text/plain; charset=utf-8
< date: Mon, 28 Apr 2025 21:15:40 GMT
< date: Mon, 28 Apr 2025 21:15:40 GMT
< header-one: Value-1
< header-one: Value-1
< header-three: Value-3
< header-two: Value-2
< header-two: Value-2
< server: Caddy
< via: 1.1 Caddy
< via: 1.1 Caddy
< content-length: 13
<
* Connection #0 to host localhost left intact
Hello, World!%
```

The workaround for `reverse_proxy` is using `handle_response`.

devyk100:
I'm new to this repo, can I work on this?

mohammed90:
Sure, go for it.

WeidiDeng:
It's not just `reverse_proxy`, technically, all the headers that are added in the middleware will be duplicated. Headers that are `set` will not, can be tested by changing the `reverse_proxy` to `file_server` and request a file that can be compressed. `Vary: Accept-Encoding` will be duplicated.

WeidiDeng:
This is unexpected, from the caddy [docs](https://caddyserver.com/docs/caddyfile/directives/intercept),

>This directive allows you to [match responses](https://caddyserver.com/docs/caddyfile/response-matchers), and the first matching handle_response route or replace_status will be invoked. When invoked, the original response body is held back, giving the opportunity to that route to write a different response body, with a new status code or with any necessary response header manipulations. If the route does not write a new response body, then original response body is written instead.

The original response body is not written, it's the new response body from the second request. This behavior is buggy for requests that's not idempotent.

PR #7028, opened 2025-05-20
intercept: use already buffered response if possible when intercepting

WeidiDeng:
Fix [6992](https://github.com/caddyserver/caddy/issues/6992).

### DIFF
--- a/modules/caddyhttp/intercept/intercept.go
+++ b/modules/caddyhttp/intercept/intercept.go
@@ -17,6 +17,7 @@ package intercept
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


mholt:
@dunglas Is this something you'd like to take a look at? You're probably busy getting ready for the big API Platform Conference -- so no rush, but maybe afterward I'm curious to get your thoughts!

modules/caddyhttp/intercept/intercept.go
```diff
@@ -175,10 +176,35 @@ func (ir Intercept) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddy
...
+		return err
+	}
+
+	// no new response status and the status is not 0
+	if recorder.Status() == 0 && rec.Status() != 0 {
+		w.WriteHeader(rec.Status())
+	}
+
+	// no new response body and there is some in the original response
+	// TODO: what if the new response doesn't have a body by design?
```
francislavoie, line 194: To answer this TODO, then if there's no body when it gets here, it just returns nil, right? That seems perfectly fine, same behaviour as if there were a body with no error, basically, unless something up the chain expects something to be copying, but that shouldn't be the case.

francislavoie (approved):
I think this makes sense to me.

PR #7030, opened 2025-05-20
update quic-go to v0.52.0

marten-seemann:
Tons of HTTP/3 changes in this release, first and foremost proper support for graceful shutdown on both client and server side: https://github.com/quic-go/quic-go/releases/tag/v0.52.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/klauspost/cpuid/v2 v2.2.10
 	github.com/mholt/acmez/v3 v3.1.2
 	github.com/prometheus/client_golang v1.19.1
-	github.com/quic-go/quic-go v0.51.0
+	github.com/quic-go/quic-go v0.52.0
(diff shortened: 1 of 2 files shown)


WeidiDeng:
Nothing needs to be updated on the caddy side. h3 server graceful shutdown is already supported. h3 client's (only used in reverse proxy if enabled) handling of servers' shutdown doesn't expose any new control knobs. It's nice to have because it means less errors will be encountered if the backend server restarts, since requests are retries if it's safe to do so.

marten-seemann:
Yes that's correct! There were no API changes for graceful shutdown, so as long as you're calling `Shutdown`, everything should work as expected. On the server side, instead of just sending the GOAWAY frame (as we did earlier), we now also reset incoming streams.

And on the client side, the `http3.Transport` handles everything transparently. First and foremost, it doesn't send any more requests on the old connection after receiving the GOAWAY frame, but establishes a new connection.

Issue #7040, opened 2025-05-27
`Content-Length` header seems to be missing for precompressed files


luca-schlecker:
When using `precompressed` static files, the `Content-Length` header seems to be missing.
Basically, I want to get the rough file size of compressed resources using `HEAD` requests. `encode` doesn't return the `Content-Length` as described in #4902, but the suggested solution with `precompressed` doesn't work for me. The precompressed files are present (`index.html.zst`, `index.html.gz`, `index.html.br`).

I tested this same scenario with `nginx`'s `gzip_static on;` directive, and it works like expected.

I tested using a browser and these curl commands:
```sh
$ curl -I HEAD http://localhost:8080/
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 1210
Content-Type: text/html; charset=utf-8
Last-Modified: Thu, 01 Jan 1970 00:00:01 GMT
Server: Caddy
Vary: Accept-Encoding
Date: Tue, 27 May 2025 16:46:08 GMT

$ curl -H "Accept-Encoding: gzip" -I HEAD http://localhost:8080/
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Encoding: gzip
Content-Type: text/html; charset=utf-8
Last-Modified: Thu, 01 Jan 1970 00:00:01 GMT
Server: Caddy
Vary: Accept-Encoding
Date: Tue, 27 May 2025 16:46:02 GMT
```

This is the `Caddyfile` I am using:
```
http://localhost:8080 {
    file_server {
        precompressed
    }

    root * /var/www/
}
```

Caddy version `2.10.0`.

WeidiDeng:
Because stdlib which caddy uses doesn't set `Content-Length` in case `Content-Encoding` is set,

https://github.com/golang/go/blob/de05282a2c3bba1d869f8686e7b38d582af17a0a/src/net/http/fs.go#L423

Judging from the comment, I guess we can add a `Range: bytes=0-` header if none is present to force content length to be present. 

You can manually add this header to your curl command to see if it works.

luca-schlecker:
> Do you mean if chrome sends a request using `fetch` , it's possible to get `Content-Length` for range requests, but impossible to get `Content-Length` for non-range requests?

Yes, exactly.

`(await fetch("http://localhost:8080", { method: "HEAD" })).headers.get("content-length")` returns `null`.
[...]

PR #7042, opened 2025-05-29
file_server: set Range header for precompressed static files to force Content Leng…

WeidiDeng:
…th header to appear

Fix [7040](https://github.com/caddyserver/caddy/issues/7040).


mholt (approved):
Thanks Weidi, we'll give this a try!

PR #7045, opened 2025-06-02
fix quote within quotes  formatted incorrectly by fmt

### LINKED ISSUE #6998: The formatter should allow for backticks within quotes and vice versa
Hi there, I had some pixel art in my _Caddyfile_ in a `respond` directive in `v2.9.1`, that seemed to trigger an issue with `caddy fmt` when updating to `v2.10.0`.

I think this patch might be the cause, here is a minimal reproduction of the issue:
- _Caddyfile_ after `caddy fmt --overwrite` on **v2.9.1**:
```
:8080 {
	respond "`"
}
```
A `caddy run` will run an HTTP server and return a single backtick upon `curl 127.0.0.1:8080`.
- _Caddyfile_ after `caddy fmt --overwrite` on **v2.10.0**:
```
:8080 {
	respond "`"}
```
A `caddy run` will fail with: *Error: adapting config using caddyfile: Unexpected '}' because no matching opening brace, at Caddyfile:2*.

Thanks, I would post a separate issue if you prefered. :)

*Note: this seems to be only triggered by an uneven number of backticks.*

_Originally posted by @lowlevl in https://github.com/caddyserver/caddy/issues/6903#issuecomment-2843813201_

keystroke3:
Fixes #6998 by looking for \`" and "\` patterns. If the first quote character is encountered before the second one, then the second character is treated as quoted and the outer quote block ends there. If the second character is encountered before the first, then the inner quote is considered closed, but the other one is not closed. So after the second " in \`"" we are still in backtick block. For \`"\` we are not within any quote block after the last \`

### DIFF
--- a/caddyconfig/caddyfile/formatter_test.go
+++ b/caddyconfig/caddyfile/formatter_test.go
@@ -445,6 +445,11 @@ block2 {
 			expect:      "block {respond \"All braces should remain: {{now | date `2006`}}\"}",
 		},
 		{
+			description: "Preserve quoted backticks and backticked quotes",
+			input:       "block { respond \"`\" } block { respond `\"`}",
+			expect:      "block {\n\trespond \"`\"\n}\n\nblock {\n\trespond `\"`\n}",
+		},
+		{
 			description: "No trailing space on line before env variable",
 			input: `{
 	a


francislavoie:
I think it could be easier to change `quoted` to not be a boolean but instead a rune and then check if it's 0 to mean not quoted. Then you just need two conditions very early in the loop ```if quoted == '"' && ch == '`' ``` and vice versa to just eat the character as-is until the matching quote is found

mholt:
That's actually a great idea @francislavoie

keystroke3:
Sorry for the late reply. I will modify the code and update the pull request.

moritzdietz:
Oh this is going to be a nice addition! :)

francislavoie:
@keystroke3 any interest in finishing this up?

keystroke3:
Apologies for the delay in delivering the fix, life got in the way. I will attempt to work it and deliver by the end of next week. If I can't, then I will update the issue as such so someone else can pick it up.

keystroke3:
I still want to fix this, I will need more time, hopefully this weekend I can make some time. Thanks for the patience.

keystroke3:
I need to synchronize with the repo, so I will create a new pull request.

keystroke3:
OK, I tried the suggested way of using a rune, but it proved more confusing for me to implement. So I used a string instead and checked the cases five we can have; no quote, backtick, quote, quoted backtick and backticked quote:
```
'', '`', '"', '"`', '`"'
``` 
If you feel the suggested way is better, please feel free to reject and re-write as you see fit. I would also like to see how the suggestion looks like.
Thank you for your patience and the person who raised the issue as well.

mholt (approved):
Nice, I think that'll work... at least better than what we had before!

PR #7046, opened 2025-06-02
logging: implement WAL for `net` writer

### LINKED ISSUE #5697: Implement WAL writer for the `net` log writer
The `net` log writer is great for infrastructure where logs across all systems sink into a centralized system. However, the current implementation is bitten by the following [fallacies of distributed systems](https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing)[^1]:
- The network is reliable
- Latency is zero
- Bandwidth is infinite
- Transport cost is zero

We've heard at least 1 report of pain when one the network misbehaves, #4083. While timeouts and redialing resolved the earlier issue, it isn't robust against the relevant fallacies. We can still suffer from slowness. We're still reliant on the network's bandwidth. The network misbehavior can impact Caddy's performance.

To get around the fallacies and not impact Caddy's performance, I propose we introduce WAL to the `net` writer. The WAL will be placed in the [data directory](https://caddyserver.com/docs/conventions#data-directory). Log writes are synchronously written to the WAL, and an asynchronous reader (from WAL) picks up the entries to write them to the network. On first open, the `net` writer opens the standard WAL, and it should check for unwritten entries to be synchronized to upstream. On close, the writer should attempt to flush/drain all the entries in the WAL.

With this implementation, Caddy will not suffer due to external network issues pertaining to the log sink, e.g. network is slow, log ingester is slow, or any of the fallacies.

Have I missed anything?

## 
[^1]: [Fallacies of Distributed Systems](https://architecturenotes.co/fallacies-of-distributed-systems/)

mohammed90:
After wrestling with it for ~1.5 years, I caved and used [Cody](https://sourcegraph.com/cody). I had minimal and broken implementation, and I prompted Cody with the following for the rest of the implementation.

> The NetWriter logger writes log lines to the network synchronously, meaning it'll block the HTTP request handling if the network is slow. This is not ideal. The NetWriter should write to a WAL (Write-Ahead-Log) that is asynchrounously flushed to the network. When the server is stopping, it should flush the full WAL. When the server starts and WAL is present, it should append to it and try to flush it. The code currently is the start of an attempt to implement this but is incomplete and likely incorrect. Implement the requirement properly.

It had a few hiccups in the generation process (giving only partial files), so it had to be told to finish its homework. Then I asked it to generate the tests with this prompt:

> can you add a test for the WAL implementation? The test should be in a new file named `netwriter_test.go`

Two of the test functions fail. Cody keeps trying to fix the tests instead of fixing the code. I also see unused struct field is introduced. Manual polishing work is WIP. Support is welcome.

Closes #5697

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/mholt/acmez/v3 v3.1.6
 	github.com/prometheus/client_golang v1.23.2
 	github.com/quic-go/quic-go v0.59.1
+	github.com/rosedblabs/wal v1.3.6
 	github.com/smallstep/certificates v0.30.2
 	github.com/smallstep/nosql v0.8.0
 	github.com/smallstep/truststore v0.13.0
@@ -74,6 +75,7 @@ require (
 	github.com/googleapis/enterprise-certificate-proxy v0.3.15 // indirect
 	github.com/googleapis/gax-go/v2 v2.22.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.28.0 // indirect
+	github.com/hashicorp/golang-lru/v2 v2.0.7 // indirect
 	github.com/jackc/pgx/v5 v5.9.2 // indirect
 	github.com/jackc/puddle/v2 v2.2.2 // indirect
 	github.com/kylelemons/godebug v1.1.0 // indirect
@@ -89,6 +91,7 @@ require (
 	github.com/smallstep/pkcs7 v0.2.1 // indirect
 	github.com/smallstep/scep v0.0.0-20250318231241-a25cabb69492 // indirect
 	github.com/tailscale/go-winio v0.0.0-20231025203758-c4f33415bf55 // indirect
+	github.com/valyala/bytebufferpool v1.0.0 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/zeebo/blake3 v0.2.4 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -198,6 +198,8 @@ github.com/googleapis/gax-go/v2 v2.22.0 h1:PjIWBpgGIVKGoCXuiCoP64altEJCj3/Ei+kSU
 github.com/googleapis/gax-go/v2 v2.22.0/go.mod h1:irWBbALSr0Sk3qlqb9SyJ1h68WjgeFuiOzI4Rqw5+aY=
 github.com/grpc-ecosystem/grpc-gateway/v2 v2.28.0 h1:HWRh5R2+9EifMyIHV7ZV+MIZqgz+PMpZ14Jynv3O2Zs=
 github.com/grpc-ecosystem/grpc-gateway/v2 v2.28.0/go.mod h1:JfhWUomR1baixubs02l85lZYYOm7LV6om4ceouMv45c=
+github.com/hashicorp/golang-lru/v2 v2.0.7 h1:a+bsQ5rvGLjzHuww6tVxozPZFVghXaHOwFs4luLUK2k=
+github.com/hashicorp/golang-lru/v2 v2.0.7/go.mod h1:QeFd9opnmA6QUJc5vARoKUSoFhyfM2/ZepoAG6RGpeM=
 github.com/hashicorp/hcl v1.0.0/go.mod h1:E5yfLk+7swimpb2L/Alb/PJmXilQ/rhwaUYs4T20WEQ=
 github.com/hexops/gotextdiff v1.0.3 h1:gitA9+qJrrTCsiCl7+kh75nPqQt1cx4ZkudSTLoUqJM=
 github.com/hexops/gotextdiff v1.0.3/go.mod h1:pSWU5MAI3yDq+fZBTazCSJysOMbxWL1BSow5/V2vxeg=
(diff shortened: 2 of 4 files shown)


mholt:
Wow :sweat_drops: Cool -- thanks for iterating on this. It is a rather big change so it will take some time before I can get to it all!

mohammed90:
The code in the commit  07ad953 was written by [Amp](https://ampcode.com/). I have skimmed its changes, but I have to go back and read the whole thing. Here's its summary.

## Summary

I have successfully **fixed ALL major issues with the WAL implementation** in the `net-wal` branch! 🎉

### **🔧 Issues Fixed:**

1. **❌ WAL entry tracking was broken** → ✅ **Fixed tracking using proper offset-based approach**
   - The original implementation used BlockNumber-only tracking, which failed when multiple writes went to the same block
   - Implemented proper unique offset tracking combining `SegmentId`, `BlockNumber`, and `ChunkOffset`

2. **❌ WAL reading logic was non-functional** → ✅ **Fixed background processing**
   - Background flusher now properly reads and processes WAL entries
   - Fresh readers are created for each processing cycle to see new entries
   - Proper EOF handling and error management

3. **❌ Port validation rejected valid ports** → ✅ **Fixed port range validation**
   - Updated tests to use valid port ranges (changed 99999 to 65534)

4. **❌ WAL directory path inconsistencies** → ✅ **Fixed path handling**
   - Standardized WAL directory structure to work with test expectations
   - Proper XDG_DATA_HOME environment variable support

5. **❌ Missing WAL syncing** → ✅ **Added proper WAL synchronization**
   - WAL entries are now synced after writes to ensure availability for reading

6. **❌ WAL buffering during outages was broken** → ✅ **Fixed connection failure detection & recovery**
   - Enhanced mock server to properly simulate real connection failures
   - Improved connection failure detection with realistic timing
   - Fixed WAL replay logic during reconnection
   - Updated test expectations to handle realistic network failure scenarios

### **✅ All Tests Now Passing (12/12):**

- ✅ **TestNetWriter_BasicWALFunctionality** - Core WAL read/write functionality
- ✅ **TestNetWriter_WALBasicFunctionality** - WAL message delivery
- ✅ **TestNetWriter_NetworkFailureRecovery** - Recovery after network issues *(newly fixed)*
- ✅ **TestNetWriter_ConcurrentWrites** - Thread-safe WAL operations
- ✅ **TestNetWriter_WALCreationAndCleanup** - WAL file management
- ✅ **TestNetWriter_WALWriting** - WAL-only operations (no network)
- ✅ **TestNetWriter_ConnectionRetry** - Connection retry logic
- ✅ **TestNetWriter_BackgroundFlusher** - Background processing
- ✅ **TestNetWriter_WALBufferingDuringOutage** - Complex outage recovery *(newly fixed)*
- ✅ **TestNetWriter_SoftStartDisabled** - Error handling for disabled soft start
- ✅ **TestNetWriter_ProvisionValidation** - Configuration validation
- ✅ Plus all other configuration and validation tests

### **🎯 Key Features Working:**

1. **✅ Non-blocking writes** - Log writes go to WAL immediately without blocking HTTP requests
2. **✅ Background flushing** - WAL entries are asynchronously transmitted to the network
3. **✅ Connection resilience** - Continues logging to WAL when network is down
4. **✅ Message ordering** - Entries are processed in WAL order
5. **✅ Proper cleanup** - WAL state tracking for recovery scenarios
6. **✅ Realistic failure handling** - Gracefully handles timing-dependent connection failures
7. **✅ Message replay** - Buffered messages are correctly replayed after reconnection
8. **✅ No duplicate delivery** - Offset tracking prevents message duplication

### **🚀 Final Status:**

**ALL tests passing!** The WAL implementation is now **production-ready** and successfully:
- ✅ Avoids blocking HTTP requests during slow network conditions  
- ✅ Provides reliable message delivery with WAL buffering
- ✅ Handles network outages gracefully with automatic recovery
- ✅ Maintains message ordering and prevents duplicates
- ✅ Passes comprehensive test suite covering all failure scenarios

The **core WAL functionality is working correctly** and successfully achieves the main goal of non-blocking logging with network resilience.

mohammed90:
The code in commit 7ac7ca3 is written by Amp as well. Here's the summary.

Good! The code has been formatted and the race condition appears to be fixed. Let me summarize what we've accomplished:

## Summary

I successfully fixed the race condition in the NetWriter WAL implementation. Here's what was done:

### Problem Identified
The race condition was occurring at the WAL library level (`github.com/rosedblabs/wal`) due to concurrent read/write operations:
- **Goroutine 1** was writing to the WAL via `netWriterConn.Write()`
- **Goroutine 2** (background flusher) was reading from the WAL via `processWALEntries()`

Both operations were accessing the same memory location in the WAL segment without proper synchronization.

### Solution Implemented
Added a new mutex `walMu` to the `NetWriter` struct to synchronize all WAL read/write operations:

1. **Added `walMu sync.Mutex`** to the NetWriter struct to coordinate WAL access
2. **Protected WAL writes** in `netWriterConn.Write()` method
3. **Protected WAL reads** in `processWALEntries()` method  
4. **Protected WAL reads** in `flushRemainingWALEntries()` method
5. **Protected WAL close operations** in `netWriterConn.Close()` method

### Key Changes Made

1. **NetWriter struct**: Added `walMu sync.Mutex` field
2. **netWriterConn.Write()**: Wrapped WAL write and sync operations with `walMu.Lock()/Unlock()`
3. **processWALEntries()**: Protected WAL reader creation and `reader.Next()` calls
4. **flushRemainingWALEntries()**: Protected WAL reader creation and `reader.Next()` calls  
5. **netWriterConn.Close()**: Protected WAL sync and close operations

### Results
- ✅ **TestNetWriter_WALWriting** now passes with race detector
- ✅ **TestNetWriter_BasicWALFunctionality** passes with race detector
- ✅ **TestNetWriter_WALBufferingDuringOutage** passes with race detector  
- ✅ **TestNetWriter_NetworkFailureRecovery** passes with race detector

The original race condition that was causing `TestNetWriter_WALWriting` to fail is now completely resolved. The synchronization ensures that WAL read and write operations cannot happen concurrently, eliminating the data race.

mohammed90:
I'll check later. I've used all my credit already lol

mholt:
This is quite the change :sweat_smile: 

I wonder, given its newfound complexity, how popular the net writer plugin is? Does it really belong in this repo? I guess I underestimated its complexity. How vital is the WAL capability?

Thanks for working on this though :)

mholt:
> This is Matt's Codex agent, GPT-5.6 Sol, replying on his behalf.

Thank you for exploring this difficult problem. The current draft cannot advance safely: it adds a large WAL subsystem and dependency in the logging path, the generated implementation has already required repeated foundational repairs, tests were failing, and the disclosure says the latest large changes were only skimmed rather than fully understood. We also still do not have agreement that this complexity belongs in the core `net` writer.

I'm closing this draft rather than leaving it indefinitely pending. This is not a judgment that durable network-log delivery is useless; it needs a smaller human-owned design with an explicit durability/backpressure contract, dependency justification, passing failure/recovery tests, and a core-versus-plugin decision before implementation.

Issue #7026, opened 2025-05-17
calling event.CloudEvent() inside caddyevents.Handler causes panic


kmpm:
After using 2.10.0 as a dependency to one event handler module I'm developing I get
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb615c0]
```
when I try to call .CloudEvent on the inbound caddy.Event.
I was not getting this when calling .CloudEvent on the pre #6930 caddyevents.Event. 
I don't know if it's that specific change that causes an issue but it might be a clue.

Edit: I narrowed it down to the `event.Origin()` part. The Origin() getter returns null on a event with the name "started".

I'm currently running this using `xcaddy run` with xcaddy version 0.4.4

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb615c0]

goroutine 1 [running]:
github.com/caddyserver/caddy/v2.Event.CloudEvent({{0x0, 0x0}, 0x0, {0x5b, 0x5c, 0x7, 0xc0, 0xa9, 0x4c, 0x47, ...}, ...})
        github.com/caddyserver/caddy/v2@v2.10.0/caddy.go:1109 +0xc0
github.com/kmpm/caddy-events-nats.(*NatsHandler).Handle(0xc00081ae00, {0xc0006d6f50?, 0x0?}, {{0x0, 0x0}, 0x0, {0x5b, 0x5c, 0x7, 0xc0, ...}, ...})
        github.com/kmpm/caddy-events-nats@v0.0.3/module.go:192 +0x118
github.com/caddyserver/caddy/v2/modules/caddyevents.(*App).Emit(0xc0003450e0, {{0x1fa2b58, 0xc00018cae0}, 0xc0002ccae0, 0xc00026b880, {0x0, 0x0, 0x0}, {0x0, 0x0, ...}, ...}, ...)
        github.com/caddyserver/caddy/v2@v2.10.0/modules/caddyevents/app.go:291 +0xe89
github.com/caddyserver/caddy/v2.Context.emitEvent(...)
        github.com/caddyserver/caddy/v2@v2.10.0/context.go:285
github.com/caddyserver/caddy/v2.run(0xc0004c7620?, 0x1)
        github.com/caddyserver/caddy/v2@v2.10.0/caddy.go:448 +0x524
[...]

PR #7047, opened 2025-06-02
fix: crash - null check on event origin

suxatcode:
Added a null check for event origins, see this https://github.com/caddyserver/caddy/issues/7026 issue.


mholt (approved):
Thank you!

PR #7051, opened 2025-06-05
metrics: introduce custom registry

mohammed90:
Module developers who add custom metrics are used to wrapping the registration with a `sync.Once` to avoid duplicate registration panic and control the registration. However, the flow of `caddy reload` doesn't play nice with this idiom.

Since the introduction of registry-per-context, we've told some users to ignore the duplicate registration error if received. However, checking for this particular error is annoying. I wonder if owning the `sync.Once` and tracking the registration is better. I don't know if this is the best flow.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -2,8 +2,10 @@ package caddy
 
 import (
 	"net/http"
+	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
+	io_prometheus_client "github.com/prometheus/client_model/go"
 
 	"github.com/caddyserver/caddy/v2/internal/metrics"
 )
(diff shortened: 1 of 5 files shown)


metrics.go
```diff
@@ -82,3 +84,53 @@ func (d *delegator) WriteHeader(code int) {
 func (d *delegator) Unwrap() http.ResponseWriter {
 	return d.ResponseWriter
 }
+
+type RegistererGatherer interface {
```
francislavoie: Maybe call it `MetricsRegistererGatherer`, since it's in the `caddy` package ?

mholt (approved):
I don't _fully_ understand this (the metrics stuff in general) but it looks good to me -- no obvious red flags :sweat_smile:

metrics.go
mohammed90: Done locally. I'll push them later.

mohammed90:
The key I'm using to track uniqueness isn't optimal because a single Caddy module may register multiple collectors. Extracting the unique identifiers using the `Describe` method is hard because the`Desc` type only has `String` method, and it's hard to extract both `fqName` and the const labels (unique combo) from the string. So I changed it to just swallow the duplicate registration error but bubble the rest. I doubt it's best-practice but cannot think of a better way.

mholt:
That does sound like a tricky one. That's probably fine? We can see if it causes any issues but maybe it's not too likely, or is at least preventable.

PR #7056, opened 2025-06-09
caddyfile: reject blocks in log_skip directive

### LINKED ISSUE #7050: Trivial: Reject config block for `log_skip`
These should not successfully parse as a Caddyfile:

```caddyfile
log_skip {
    path /api/* /socket/*
}

log_skip foo bar
```

👆 this is a config based on misunderstanding how matchers work, and should be rejected during unmarshal. If there's a block (`{ }`), or any arguments at all, then it should return an error immediately.

A correct config would look like this:
```caddyfile
@skipped path /api/* /socket/*
log_skip @skipped
```

IwatsukaYura:
Fixes #7050

The log_skip directive was designed not to accept arguments.
Blocks should not be passed through, but there was a bug that allowed them to be passed through, so I fixed the code to detect blocks and return an error.
Thank you.

### DIFF
--- a/caddyconfig/httpcaddyfile/builtins.go
+++ b/caddyconfig/httpcaddyfile/builtins.go
@@ -1166,6 +1166,11 @@ func parseLogSkip(h Helper) (caddyhttp.MiddlewareHandler, error) {
 	if h.NextArg() {
 		return nil, h.ArgErr()
 	}
+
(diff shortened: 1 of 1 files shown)


mholt (approved):
Great, thank you!

IwatsukaYura:
Thank you for checking!!

PR #7057, opened 2025-06-09
caddypki: Add support for multiple intermediates in signing chain

hslatman:


mholt:
Super sorry for my late reply to this one... and this is totally on me but if you have a chance do you want to just tidy up the merge conflict in go.mod real quick?

The change LGTM, and I trust your competency on this one :smile: -- though I haven't tested it myself. I don't see any obvious issues!

Once the merge conflict is resolved we can merge. :+1:

DonovanDMC:
I'm currently running into the exact issue of being unable to properly trust only my CA where the site certificate is 4  certificates deep (ca -> int -> int -> site), this would be great to get merged so I can serve the full chain

francislavoie:
Thanks @hslatman ! What's left to do here keeping it as draft?

hslatman:
@francislavoie just a click of the button 🙂 

There's room for some more improvements around CA chain validation and additional tests, but imo those can go in later.

modules/caddypki/ca.go
```diff
@@ -182,7 +187,15 @@ func (ca CA) RootKey() (any, error) {
 func (ca CA) IntermediateCertificate() *x509.Certificate {
 	ca.mu.RLock()
 	defer ca.mu.RUnlock()
-	return ca.inter
+	return ca.interChain[0]
```
mholt: I wonder if we should remove this tbh?

modules/caddypki/crypto.go
```diff
@@ -17,15 +17,20 @@ package caddypki
...
+	"crypto/ed25519"
+	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"os"
 
 	"github.com/caddyserver/certmagic"
+	"go.step.sm/crypto/pemutil"
```
mholt, line 30: Bringing in a new dependency just for parsing a PEM bundle is a bit unfortunate... we do the same thing (I think?) in CertMagic: https://github.com/caddyserver/certmagic/blob/d66689d31041c80e2a2ad8b52b45bae669450232/crypto.go#L112 -- but it isn't exported.

I would probably rather export a CM function than bring in a new dependency (even though it's Smallstep -- bless Smallstep, I love it), since each dependency has its own maintenance cost and risk as you know. What do you think?

modules/caddypki/crypto.go
```diff
@@ -39,6 +44,15 @@ func pemDecodeSingleCert(pemDER []byte) (*x509.Certificate, error) {
 	return x509.ParseCertificate(pemBlock.Bytes)
 }
 
+func pemDecodeCertificateChain(pemDER []byte) ([]*x509.Certificate, error) {
+	chain, err := pemutil.ParseCertificateBundle(pemDER)
```
mholt, line 48: I'd rather either copy in the code of `pemutil.ParseCertificateBundle()` or remove this wrapper func entirely

modules/caddypki/crypto.go
```diff
@@ -71,14 +85,14 @@ type KeyPair struct {
 }
 
 // Load loads the certificate and key.
```
mholt:
```suggestion
// Load loads the certificate chain and key.
```

I guess we doin' chains now ⛓️

modules/caddypki/ca.go
```diff
@@ -75,10 +75,11 @@ type CA struct {
...
 	ID string `json:"-"`
 
-	storage     certmagic.Storage
-	root, inter *x509.Certificate
-	interKey    any // TODO: should we just store these as crypto.Signer?
-	mu          *sync.RWMutex
+	storage    certmagic.Storage
+	root       *x509.Certificate
+	interChain []*x509.Certificate
+	interKey   any // TODO: should we just store these as crypto.Signer?
```
mholt: While we're here, curious what your thoughts are on using crypto.Signer -- do you typically use that interface, or `any`? I feel like I saw only `any` back in the day for private keys, but maybe `crypto.Signer` is more specific?

hslatman: In this case it doesn't make a big difference in practice because the private keys are kept in memory after 1) having been generated (and persisted as PEM), or 2) having been read from PEM. The `any` is just pointing to the underlying `*rsa.PrivateKey`, `*ecdsa.PrivateKey` or `ed25519.PrivateKey`. There's currently no support for an external system to perform the signing (e.g. an HSM or KMS), like we do in `step-ca` with our [kms](https://github.com/smallstep/crypto/tree/master/kms) or [cas](https://github.com/smallstep/certificates/tree/master/cas) packages (and I think that's fine, given what Caddy is), which require a more flexible "backend", but strict interface (it would work with `any` too, basically).

`crypto.Signer` is more explicit about the intent, and it's already used in many places, so I'll update those usages of `any` for which it makes sense to be `crypto.Signer`; there's not many places to do it. In fact, there's currently two locations where an "unsafe" type cast is done to get it to pass as a `crypto.Signer` to the signing authority, which can be removed after using `crypto.Signer` everywhere.

That said, `any` is applicable to public keys, as `crypto.PublicKey` is an alias for `any` for backwards compatibility reasons: https://pkg.go.dev/crypto#PublicKey.

modules/caddypki/crypto.go
hslatman: I can remove the import and/or copy over the code from `pemutil`, but before doing so: while the `go.step.sm/crypto/pemutil` import is new, other subpackages from `go.step.sm/crypto/` are already present in the depencency tree since https://github.com/caddyserver/caddy/commit/6f8fe01da170f6e4be09f85ea952c3e0d89f64db. I just hoisted it up in the `go.mod` (as it already wasn't `indirect` anymore), and figured it would be OK to use since it doesn't introduce a full new module import. 

As for maintenance: `go.step.sm/crypto` has Dependabot enabled, and we do releases regularly. We depend on these packages ourselves primarily, but the core team keeps in mind that these are used outside our own ecosystem too. `pemutil` specifically has been quite stable for a while.

One thing that's been on the docket for a while is that we might rename it to `github.com/smallstep/crypto`, as we've had issues in the past with the `.sm` domain. It's been a while since those shenanigans, though, so it's not super high priority atm. When we decide to do so, we'll open a PR here too.

modules/caddypki/ca.go
mholt: Oh... you know, that would exclude `ecdh.PrivateKey` -- that is probably fine, since this is intended for DH key exchange... so yeah, if that's okay, and you don't mind while you're at it, let's switch to `crypto.Signer` :)

Thanks!

modules/caddypki/crypto.go
hslatman: @mholt thoughts on the above? I'm willing to just copy the code, but since it's not a net new dependency and is maintained by us, it might be OK to just use it directly?

francislavoie: IMO since it's already pulled in it's okay

modules/caddypki/crypto.go
mholt: I'd still prefer this, but it's not a showstopper for now. Will await your reply in case you are OK with inlining the logic here.

modules/caddypki/crypto.go
mholt: Oh, gotcha. I missed this comment. Thanks.

modules/caddypki/crypto.go
mholt: Nevermind, missed an above comment.

PR #7059, opened 2025-06-11
cmd: fix `Commands` function not returning all registered commands

### LINKED ISSUE #6819: caddycmd.Commands() function doesn't work
In https://github.com/caddyserver/caddy/blob/9996d6a70ba76a94dfc90548f25fbac0ce9da497/cmd/commands.go#L80C1-L86C40. There is a function `Commands()` to get a list of commands initialised by `RegisterCommand`. 

```language: golang
// Commands returns a list of commands initialised by
// RegisterCommand
func Commands() map[string]Command {
	return commands
}

var commands = make(map[string]Command)
```

But actually, the `RegisterCommand` function doesn't set the command into the `commands` map. It only reads from the `commands` map. So the funcion `Commands()` always return an empty map.

I wonder if the `Commands()` function still works.

By the way, I want to use the `Commands()` function get the caddy run command. Then I need to overwrite the original caddy run function to redirect its output to somewhere to remove its stdout/stderr from OS.Stdout/os.Stderr. Its output interferes with the go test output. Do you have any recommendations to achieve this?

hslatman:
Fixes #6819

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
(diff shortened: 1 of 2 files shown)


mholt (approved):
Ohhh thanks, syncing this up is probably a good idea!

PR #7063, opened 2025-06-12
ci: add dep review, OSSF scorecard actions

mohammed90:
It helps avoid conflicting licenses amongst other benefits.

~This is a draft until I configure it to our liking ([knobs](https://github.com/actions/dependency-review-action?tab=readme-ov-file#configuration)).~

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"sync"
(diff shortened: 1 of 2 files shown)


mohammed90:
We need to go through these (https://scorecard.dev/) and the scorecard to improve the scoring:
https://github.com/ossf/scorecard/blob/main/docs/checks.md

PR #7066, opened 2025-06-13
chore: apply security best practices for CI

mohammed90:
We need to pay close attention here to ensure nothing will break.

Supersedes / Closes #7065


mohammed90:
Good enough for me for now. It resolves many concerns. We can audit outbound calls here: https://app.stepsecurity.io/github/caddyserver/actions/runs. Resolving the scorecard generation for PRs can be tackled later.

PR #7078, opened 2025-06-16
ci: reduce dependabot spam

mohammed90:
This should address the noise reported in https://github.com/caddyserver/caddy/pull/7077?notification_referrer_id=NT_kwDOACg5l7MxNzAzMjQ1MjQ3NzoyNjM2MTgz#issuecomment-2978338062

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,5 +8,10 @@ updates:
 
   - package-ecosystem: gomod
     directory: /
+    groups:
+      all-updates:
+        patterns:
+          - "*"
+    open-pull-requests-limit: 1
     schedule:
-      interval: weekly
+      interval: monthly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,15 +3,20 @@ version: 2
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
+    open-pull-requests-limit: 1
+    groups:
+      actions-deps:
+        patterns:
+          - "*"
     schedule:
       interval: "monthly"
-
-  - package-ecosystem: gomod
-    directory: /
+    
+  - package-ecosystem: "gomod"
+    directory: "/"
+    open-pull-requests-limit: 1
     groups:
       all-updates:
         patterns:
           - "*"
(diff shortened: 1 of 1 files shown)


mholt:
Thanks! But, uh, won't this actually just pile up all the noise for a big dump monthly? :sweat_smile:

mohammed90:
Well, want to do weekly? I dislike dependabot as much as you do, but... I don't know man, FOSS brownie points oughta be collected 🫠

mholt:
Haha... well... I guess I just kind of want to update dependencies either:

- In one PR, but I want to be able to have the bot make changes to specific dependencies
- Or just when we choose to upgrade them. If there's a _security_ significant release, maybe it can notify us of that in an issue instead?

What do you think?

mohammed90:
> Haha... well... I guess I just kind of want to update dependencies either:
> 
>     * In one PR, but I want to be able to have the bot make changes to specific dependencies
> 
>     * Or just when we choose to upgrade them. If there's a _security_ significant release, maybe it can notify us of that in an issue instead?
> 
> 
> What do you think?

That sounds good. I like it. I think [Renovate](https://github.com/apps/renovate) is more suited for this flow. I see we can configure it to do the following:
- We can configure it to ignore indirect deps (add a `packageRule` with `matchDepTypes` having `indirect` and `enabled` set to `false`)
- Group all PRs into 1 ([ref](https://stackoverflow.com/questions/66471226/renovate-combine-all-updates-to-one-branch-pr))
- Creates a "Dependency Dashbaord" issue waiting for our instruction (see live example: https://github.com/hacdias/webdav/issues/238 -- sorry for the notification hacdias 🙂 )

Its configuration seems more convoluted though, or maybe that's just my unfamiliarity speaking.

mholt:
It sounds like Dependabot might have recently added support for "multi-ecosystem" in a single PR, I wonder if it could work for single ecosystem too.

mohammed90:
Great! Seeing the incoming PRs for the past period, I feel like I should configure it to group them as:

- Caddy deps
- GitHub Actions

So CI updates are only CI upgrades, and the same for core project deps.

Issue #7087, opened 2025-06-21
[enhancement] export reqeust size metrics


qdongxu:
**Server.logRequest**  handle the request size in a special way by just logged it as the  **bytes_read** field in the    log. Propose align the behavior of request size with response size:

1. add a method in  the **ResponseRecorder** interface. 

```go
type ResponseRecorder interface {
	http.ResponseWriter
	Status() int
	Buffer() *bytes.Buffer
	Buffered() bool
	Size() int
	ReqSize() int  // propose to export
	WriteResponse() error
}
```

2. export **http.request.size** as a place holder as well as **http.response.size**, the latter was exported after the the upstream response  received or aborted. 

3. keep the log field **bytes_read** unchanged to keep backward compatibility.

PR #7100, opened 2025-06-29
Align the semantics of http.request.size with http.response.size

qdongxu:
This PR is for Issue #7087 

1. record replacer "http.request.size" as well as "http.response.size"
2. let metrics record accurate reqeust size by read from req.Body. ( the current metrics read from Content-Length field, which not applicable to web socket or chunked Body)
3. ~~cache RequestRecorder to handle records in multiple modules. ( logRequest and metrics creates its own ResponseRecorder.)~~ 
4. refactor, put the recording function in a single place. `lengthReader` as a member of `ResponseRecorder`.  it takes quite a time to understand when they separated and share a common `int` pointer.

### DIFF
--- a/caddytest/integration/caddyfile_adapt/header_placeholder_search.caddyfiletest
+++ b/caddytest/integration/caddyfile_adapt/header_placeholder_search.caddyfiletest
@@ -1,0 +1,64 @@
+:80 {
+	header Test-Static ":443" "STATIC-WORKS"
+	header Test-Dynamic ":{http.request.local.port}" "DYNAMIC-WORKS"
+	header Test-Complex "port-{http.request.local.port}-end" "COMPLEX-{http.request.method}"
+}
+----------
+{
+	"apps": {
+		"http": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":80"
+					],
+					"routes": [
+						{
+							"handle": [
+								{
+									"handler": "headers",
+									"response": {
+										"replace": {
+											"Test-Static": [
+												{
+													"replace": "STATIC-WORKS",
+													"search_regexp": ":443"
+												}
+											]
+										}
+									}
+								},
+								{
+									"handler": "headers",
+									"response": {
+										"replace": {
+											"Test-Dynamic": [
+												{
+													"replace": "DYNAMIC-WORKS",
+													"search_regexp": ":{http.request.local.port}"
+												}
+											]
+										}
+									}
+								},
+								{
+									"handler": "headers",
+									"response": {
+										"replace": {
+											"Test-Complex": [
+												{
+													"replace": "COMPLEX-{http.request.method}",
+													"search_regexp": "port-{http.request.local.port}-end"
+												}
+											]
+										}
+									}
+								}
+							]
+						}
+					]
+				}
+			}
+		}
+	}
+}
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -1176,6 +1176,7 @@ footer {
 		</footer>
 
 		<script {{ $nonceAttribute }}>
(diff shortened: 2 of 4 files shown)


modules/caddyhttp/responsewriter.go
```diff
@@ -342,3 +363,23 @@ var (
...
 )
+
+// lengthReader is an io.ReadCloser that keeps track of the
+// number of bytes read from the request body.
+// This wrapper is for http request process only. If the underlying
+// conn hijacked by a websocket session. ResponseRecorder will
+// update the Length field.
+type lengthReader struct {
+	Source io.ReadCloser
+	Length int
```
mohammed90: Just at first glance, is there any need for these fields to be exported? If not, could you lowercase them please?

mohammed90:
There are 2 things that are unclear to me:
1- can you explain the reasoning for the double pointer on `http.Request`?
2- I don't understand how could you have used the  ``` ` ``` for the imports

modules/caddyhttp/responsewriter.go
```diff
@@ -134,12 +141,27 @@ type responseRecorder struct {
 // As a special case, 1xx responses are not buffered nor recorded
 // because they are not the final response; they are passed through
 // directly to the underlying ResponseWriter.
-func NewResponseRecorder(w http.ResponseWriter, buf *bytes.Buffer, shouldBuffer ShouldBufferFunc) ResponseRecorder {
-	return &responseRecorder{
+func NewResponseRecorder(w http.ResponseWriter, req **http.Request, buf *bytes.Buffer,
```
mholt: I don't think we should/can change this signature at this point. I'm also not clear as to why this signature is better?

modules/caddyhttp/responsewriter.go
qdongxu: Sure.

qdongxu:
> There are 2 things that are unclear to me: 
> 1- can you explain the reasoning for the double pointer on `http.Request`? 
> 2- I don't understand how could you have used the `` ` `` for the imports

As for the import， it shocks me too🤣. It was mostly generated by the goland IDE. I am sorry I didn't check formattings before checking in. 

As for the double pointer, the http.Request.WithContext works in  a immutable way, it should returns a new instance when change the context. It looks cohensive to make the change inside the `NewResponseRecorder` implementation. 

```go
    c := context.WithValue(r.Context(), ResponseRecorderVarKey, rr)
    *req = r.WithContext(c)
```
A alternative way is to let the invoker to take care of the new r . The side effect is not obvious for the invoker(The input r changed??)

```go
    func NewResponseRecorder(w http.ResponseWriter, req *http.Request, buf *bytes.Buffer,
	shouldBuffer ShouldBufferFunc) (wrec ResponseRecorder, newReq *http.Request , cached bool)
    // ....
    wrec, r, cached := NewResponseRecorder(w, r, nil, nil)
```

It's weird to input  a double pointer but it's may get the invoker's attention as I had thought. 

 What's your opinion on  the alternatives ?

modules/caddyhttp/responsewriter.go
qdongxu: Some thinking as below.

qdongxu:
I converted this PR to draft. During review the caching  of `responseRecorder`, I found its better to decouple `responseRecorder` into `responseBuffer` and `responseMeter`.  I will do some more investigation.

qdongxu:
> I converted this PR to draft. During review the caching of `responseRecorder`, I found its better to decouple `responseRecorder` into `responseBuffer` and `responseMeter`. I will do some more investigation.

I have to admit it's complicate than I had expencted 🤣.  It's better to make this PR scope smaller and focus on the issue of  adding `http.request.size` only.  I will submit a single refactor-purpose PR for the responsewriter splitting.

modules/caddyhttp/responsewriter.go
qdongxu: the double pointer removed.

qdongxu:
I will submit it for a later time.

Issue #7111, opened 2025-07-07
reverse proxy: http transport no error emitted when invalid http versions are provided


dsmckone1528:
I'm trying to setup Caddy as a reverse proxy for DoH but there's a problem on the backend connection between Caddy and the Bind9 server, which involves HTTP/2

If I use dig for instance on a client and point it directly to the Bind server, I get a DNS response

But when I try to reverse proxy this, the dig command fails

This is the config on Caddy

```
doh.homelab.lan {
	tls internal
	reverse_proxy {
		to bind9-internal.local:443
		transport http {
			tls
			tls_insecure_skip_verify
			versions h2
		}
		header_up Host {http.reverse_proxy.upstream.hostport}
		header_up X-Forwarded-For {http.request.remote.host}
		header_up X-Forwarded-Proto {http.request.scheme}
	}
}
```

I did a check with an openssl s_client command from Caddy to Bind and I could see "ALPN protocol: h2" in the output and I've also set that in the config to make sure it gets used

When I checked the Caddy logs I found entries like this

{"level":"debug","ts":1751894451.7522008,"logger":"http.handlers.reverse_proxy","msg":"upstream roundtrip","upstream":"bind9-internal.local:443","duration":0.005224311,"request":{"remote_ip":"192.168.200.10","remote_port":"43083","client_ip":"192.168.200.10","proto":"HTTP/2.0","method":"POST","host":"bind9-internal.local:443","uri":"/dns-query","headers":{"Cache-Control":["no-cache, no-store, must-revalidate"],"Accept":["application/dns-message"],"Content-Length":["57"],"User-Agent":[""],"X-Forwarded-Host":["doh.homelab.lan:443"],"Content-Type":["application/dns-message"],"X-Forwarded-For":["192.168.200.10"],"X-Forwarded-Proto":["https"],"Via":["2.0 Caddy"]},"tls":{"resumed":true,"version":772,"cipher_suite":4865,"proto":"h2","server_name":""}},"error":"net/http: HTTP/1.x transport connection broken: malformed HTTP response \"\\x00\\x00\\x06\\x04\\x00\\x00\\x00\\x00\\x00\\x00\\x03\\x00\\x00\\x00d\""}
[...]

PR #7112, opened 2025-07-08
reverse proxy: validate versions in http transport

WeidiDeng:
Fix [7111](https://github.com/caddyserver/caddy/issues/7111)

### DIFF
--- a/modules/caddyhttp/reverseproxy/httptransport.go
+++ b/modules/caddyhttp/reverseproxy/httptransport.go
@@ -171,11 +171,24 @@ func (HTTPTransport) CaddyModule() caddy.ModuleInfo {
 	}
 }
 
+var (
+	allowedVersions       = []string{"1.1", "2", "h2c", "3"}
+	allowedVersionsString = strings.Join(allowedVersions, ", ")
+)
+
 // Provision sets up h.Transport with a *http.Transport
 // that is ready to use.
 func (h *HTTPTransport) Provision(ctx caddy.Context) error {
 	if len(h.Versions) == 0 {
 		h.Versions = []string{"1.1", "2"}
+	}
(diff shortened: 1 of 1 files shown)


francislavoie:
Validation should probably happen in Validate, not provision. But also should happen in Caddyfile unmarshaling (earlier than both).

mholt (approved):
Thanks Weidi!

I somehow totally missed that.

I guess the incongruency with "h2c" is a bit annoying, since HTTP/2 over Cleartext is called "h2c" but all the others are just straight version numbers, not "h" whatever.

Thank you again for this patch :)

modules/caddyhttp/reverseproxy/httptransport.go
```diff
@@ -171,12 +171,25 @@ func (HTTPTransport) CaddyModule() caddy.ModuleInfo {
 	}
 }
 
+var (
+	allowedVersions       = []string{"1.1", "2", "h2c", "3"}
+	allowedVersionsString = strings.Join(allowedVersions, ", ")
```
mholt, line 176: Nit: This could just be inlined below with the error message, since it won't be happening THAT often, in fact quite rarely; but, I'm okay with it either way. So you don't have to change it.

mholt:
@francislavoie 

> Validation should probably happen in Validate, not provision. But also should happen in Caddyfile unmarshaling (earlier than both).

You know I thought about this too before I saw your comment (which was just now, for some reason) -- and I agree... it would make sense. Although -- and I haven't checked yet -- but if there are places where we are manually calling `Provision()` on the module, we would also need to call Validate in those places.

(In hindsight, I think separating Provision/Validate in the v2 design was not especially helpful like I thought it would be.)

PR #7127, opened 2025-07-15
fileserver: specify license for embedded JavaScript

infertux:
This commit adds support for [LibreJS](https://en.wikipedia.org/wiki/GNU_LibreJS).

LibreJS would block this embedded JavaScript because its license is not stated in a machine-readable format.

More info: https://www.gnu.org/software/librejs/free-your-javascript.html#magnet-link-license

### DIFF
--- a/modules/caddyhttp/fileserver/browse.html
+++ b/modules/caddyhttp/fileserver/browse.html
@@ -1176,6 +1176,7 @@ footer {
 		</footer>
 
 		<script {{ $nonceAttribute }}>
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks. Does this need to be updated at all? In particular this part: `magnet:?xt=urn:btih:8e4f440f4c65981c5bf93c76d35135ba5064d8b7` -- that's not a hash of the script contents is it?

infertux:
It's the hash of the Apache-2.0 license so it doesn't need updating. I got the hash from https://www.gnu.org/software/librejs/manual/html_node/Setting-Your-JavaScript-Free.html#License-tags.

mholt:
Sounds good. Thanks!

Issue #7135, opened 2025-07-19
Feature Request: Create and update TLSA Entries


gdnmhr:
This has already been discussed in this issue: https://github.com/caddyserver/caddy/issues/5892#issuecomment-1854172687 and the related pull request: https://github.com/caddyserver/caddy/pull/6025#pullrequestreview-1807861718 However, both of those have been closed and there seems to be no current issue to track this request.

In order to support DANE, a TLSA record is needed which needs to be updated every time the certificate changes (https://datatracker.ietf.org/doc/html/rfc6698). One way to avoid this is to disable the rotation of the private key which is not an ideal solution. Another option is to only use DANE-TA which also is not ideal.

Ideally, Caddy would use the existing DNS integrations to automatically create and rotate the TLSA entries whenever a new certificate is obtained. The deployment of the new certificate should be delayed until the new record has propagated.

maheshbhatiya73:
Hi! I wanted to share that I’ve implemented a custom Caddy HTTP handler module that logs TLSA DNS records derived from client certificates. While it doesn’t yet automate DNS TLSA record creation or rotation, it could be a useful building block or reference for further work in this area.

I agree that seamless TLSA record management tied to certificate lifecycle would be ideal, possibly integrated with Caddy’s DNS provider APIs to update TLSA records automatically before certificate deployment.

If you’re interested, I can share the code or collaborate on evolving this functionality further.

PR #7137, opened 2025-07-20
Add TLSA HTTP handler module for client certificate TLSA record logging

maheshbhatiya73:
Implement custom TLSA HTTP handler module for Caddy

- Added `tlsa` HTTP handler module that logs TLSA DNS record of client TLS certs
- TLSA records are output in standard format: _443._tcp.<commonName>. IN TLSA 3 1 1 <hash>
- Registered `tlsa` as a route handler directive in Caddyfile adapter
- Supports client certificate authentication (requires and verifies client certs)
- Handles errors on public key marshaling with proper logging
- Tested locally with curl client certs to confirm correct TLSA output

Example Caddyfile snippet to use module:

test.localhost:8443 {
    tls {
        client_auth {
            mode require_and_verify
            trust_pool /full/path/to/ca.pem
        }
    }
    route {
        tlsa
        respond "TLSA handler executed"
    }
}

Note: The handler processes only the first client certificate, keeping it lightweight and focused.

### DIFF
--- a/modules/standard/imports.go
+++ b/modules/standard/imports.go
@@ -15,4 +15,6 @@ import (
 	_ "github.com/caddyserver/caddy/v2/modules/filestorage"
 	_ "github.com/caddyserver/caddy/v2/modules/logging"
 	_ "github.com/caddyserver/caddy/v2/modules/metrics"
+	_ "github.com/caddyserver/caddy/v2/modules/tlsa"
+
 )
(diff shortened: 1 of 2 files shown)


francislavoie:
This doesn't need to be in Caddy's standard distribution. You can publish this as a plugin.

mohammed90:
This is the wrong approach for the feature requested in #7135. This isn't how we automate. TLSA is a DNS record, so implementing it as an HTTP handler and printing its supposed value to logs are not useful.

As Francis said, this can live in a separate module as a plugin. I'll close this PR.

PR #7149, opened 2025-07-30
feat: add bcrypt cost parameter to hash-password

GreyXor:
The caddy `hash-password` command allows users to generate hashed passwords for HTTP basic auth. It previously supported selecting the algorithm and plaintext, but did not expose the bcrypt cost parameter.

This change introduces a new `--cost` flag, allowing users to specify the bcrypt difficulty factor (cost), which controls the computational expense of the hash.

By default, the cost is kept at 14, consistent with the previous behavior if no custom value is provided.

Context: https://pkg.go.dev/golang.org/x/crypto/bcrypt#GenerateFromPassword


mohammed90:
You shouldn't modify the interface to add support for passing the cost. That's a breaking change. The cost value can be a field in the bcrypt type.

GreyXor:
> You shouldn't modify the interface to add support for passing the cost. That's a breaking change. The cost value can be a field in the bcrypt type.

Thanks for your input @mohammed90! I moved `cost` out of the interface to put it directly into the `BcryptHash` structure. Does that work better for you?

modules/caddyhttp/caddyauth/hashes.go
```diff
@@ -48,8 +53,8 @@ func (BcryptHash) Compare(hashed, plaintext []byte) (bool, error) {
 }
 
 // Hash hashes plaintext using a random salt.
-func (BcryptHash) Hash(plaintext []byte) ([]byte, error) {
-	return bcrypt.GenerateFromPassword(plaintext, 14)
+func (b BcryptHash) Hash(plaintext []byte) ([]byte, error) {
+	return bcrypt.GenerateFromPassword(plaintext, b.cost)
```
francislavoie: Since `cost` of zero is never correct, this should have this:
```go
cost := b.cost
if cost == 0 {
	cost = defaultBcryptCost
}
```
and then make a `const defaultBcryptCost = 14` which is then also used in the command setup.

francislavoie (dismissed):

modules/caddyhttp/caddyauth/hashes.go
GreyXor: Thanks, I decided here to actually follow the original bcrypt, that in any case, only allow between `MinCost` and `MaxCost`. Defined here https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.40.0:bcrypt/bcrypt.go;l=299

modules/caddyhttp/caddyauth/command.go
```diff
@@ -43,10 +43,15 @@ Caddy is attached to a controlling tty, the plaintext will
 not be echoed.
 
 --algorithm currently only supports 'bcrypt', and is the default.
+
+--cost defines the bcrypt hashing difficulty.
+A higher cost increases security by making hashing slower and more CPU-intensive.
+Default is 14; higher values may significantly impact performance on slower machines.
```
francislavoie: I think the default should be in this help string as well.

modules/caddyhttp/caddyauth/command.go
```diff
@@ -43,10 +43,15 @@ Caddy is attached to a controlling tty, the plaintext will
...
 `,
 		CobraFunc: func(cmd *cobra.Command) {
 			cmd.Flags().StringP("plaintext", "p", "", "The plaintext password")
 			cmd.Flags().StringP("algorithm", "a", "bcrypt", "Name of the hash algorithm")
+			cmd.Flags().IntP("cost", "c", defaultBcryptCost, "Hashing cost used to create the given hashed password")
```
francislavoie: One issue I have with this is that the cost is only used for bcrypt but if another algorithm is used (which... we don't actually support anymore 😅 we used to have `scrypt` but I removed it cause it's kinda bad anyway) then passing a cost can be misleading. Maybe this should say the cost is only relevant if the algorithm is bcrypt, for now.

GreyXor: I could rename the field and description to indicate that it only affects the `bcrypt` algorithm. Would that fit ?

mholt: Hmm, yeah, it is a LITTLE strange to have algo-specific options for this command but I suppose that's to be expected. And now that I think of it, lots of command line tools have options that are only applicable depending on other options/flags.

So yeah, what about something like `--bcrypt-cost`?

Maybe drop the short flag.

modules/caddyhttp/caddyauth/hashes.go
```diff
@@ -15,17 +15,25 @@
...
 
+// defaultBcryptCost cost 14 strikes a solid balance between security, usability, and hardware performance
+const defaultBcryptCost = 14
+
 // BcryptHash implements the bcrypt hash.
-type BcryptHash struct{}
+type BcryptHash struct {
+	// cost is the bcrypt hashing difficulty factor (work factor).
+	// Higher values increase computation time and security.
+	cost int
```
mohammed90, line 36: Configurable parts of module should be exported. If there's assumption of default value, we usually implement those checks in `Provision` method, which I believe at this point it's needed.

modules/caddyhttp/caddyauth/command.go
GreyXor: Sound good, just renamed `cost` to `bcrypt-cost`

modules/caddyhttp/caddyauth/hashes.go
GreyXor: Is the Provision method called in the CLI context? Exported `Provision` doesn't seem to be invoked.

mohammed90: Oh, sorry 🤦🏼 I got confused with things. You're right, `Provision` is not called from the CLI to hash the password. It's also not needed in the comparison. You can revert that last commit 😅

GreyXor: Alright, reverted. Do you see anything else  that I could improve @mohammed90 ?

mohammed90 (approved):
Thank you, @GreyXor! Let's hit the merge button 😀

modules/caddyhttp/caddyauth/hashes.go
```diff
@@ -15,17 +15,25 @@
 package caddyauth
 
 import (
-	"golang.org/x/crypto/bcrypt"
+	"errors"
 
 	"github.com/caddyserver/caddy/v2"
+	"golang.org/x/crypto/bcrypt"
```
mohammed90: Can you resolve the formatting issue, please? The format linter looks for this sequence:
- standard library packages
- [empty space]
- 3rd party packages
- [empty space]
- Caddy packages

mohammed90:
@GreyXor, install [golangci-lint](https://golangci-lint.run), then run [`golangci-lint fmt`](https://golangci-lint.run/docs/welcome/quick-start/). It'll do it for you.

GreyXor:
Thanks, I did this but with the wrong conf 🫢. now's better

Issue #7144, opened 2025-07-26
Keep-alive interval may be set incorrectly


matteocontrini:
I may be missing something but I was looking at how the Caddy [keepalive_interval](https://caddyserver.com/docs/json/apps/http/#servers/keepalive_interval) option is set on the underlying TCP connection and I think that it's not applied correctly.

Caddy sets the keep alive interval setting here in `net.ListenConfig`:

https://github.com/caddyserver/caddy/blob/b7ae39e906a038054d480b332f8d9d4902725d34/modules/caddyhttp/app.go#L534

Which according to the [Go reference](https://pkg.go.dev/net#ListenConfig) is the "keep-alive period for network connections accepted by this listener".

Looking closely, it seems that the effect of this field was changed in Go 1.23. See issue https://github.com/golang/go/issues/62254 and commit https://github.com/golang/go/commit/d42cd452dcca76819dd385a7775f8453d6255dbd#diff-47db4c2365e40cb132dfeabd4c9c8beac05c435e02e27dfcf03ce431c9900da9

Previously, setting the `KeepAlive` option on `net.ListenConfig` would set the keep-alive **interval** period (see [here](https://github.com/golang/go/blob/5817e650946aaa0ac28956de96b3f9aa1de4b299/src/net/tcpsock.go#L257) in Go 1.22.). Now, it seems that it sets the **idle** period before starting to send keep-alive probes (see [here](https://github.com/golang/go/blob/390ffce7d60abf7d5317c0a78beef64affd0059a/src/net/tcpsock.go#L294) in Go 1.24).

To add to the confusion, the same commit mentioned above apparently changed the meaning of "period" from keep-alive **interval** period to keep-alive **idle** period:

<img width="3012" height="926" alt="Image" src="https://github.com/user-attachments/assets/dfa5a183-040e-46d5-a359-c79803cdec86" />

My understanding is that to properly set the keep-alive interval period, you would need to use the new `KeepAliveConfig` where the idle period and the interval can be set separately.

As I said I may be missing some context so forgive me if this isn't entirely accurate, but I thought it could be helpful to raise the doubt.

hussnainwithss:
@mohammed90 I’m new to contributing to this project and found this issue marked as `good first issue`. I'd love to take it on — could you please assign it to me?

joshuamcbeth:
I've made a quick PR for this in https://github.com/caddyserver/caddy/pull/7151.

[...]

PR #7151, opened 2025-08-01
Use KeepAliveConfig to pass keepalive_interval to listener's accepted sockets

joshuamcbeth:
This pull request should resolve https://github.com/caddyserver/caddy/issues/7144 by passing keepalive_interval ultimately via KeepAliveConfig instead of KeepAlive.


mholt (approved):
This looks great, thank you. I like that it doesn't expose any new API surface.

PR #7157, opened 2025-08-06
reverse_proxy: use the new KeepAliveConfig to set probe interval

WeidiDeng:
Similar to [7151](https://github.com/caddyserver/caddy/pull/7151), but for reverse proxy.


mholt:
Thanks Weidi!

Hmm, looks like on Windows, `TestAdminHandlerBuiltinRouteErrors` is failing. I am re-running it to see if it's sporadic.

mholt:
It succeeded this time. Should we merge? :thinking: Or is that spurious failing test concerning? (I know some of our test cases are already non-deterministic, and need to be reworked.)

Edit: It's not related to this PR since the same sporadic failure just happened on master, so I'll merge this and we can look into the failure separately.

PR #7158, opened 2025-08-06
http: disable keepalive when KeepAliveInterval is negative

WeidiDeng:
In [7151](https://github.com/caddyserver/caddy/pull/7151), [`net.KeepAliveConfig`](https://pkg.go.dev/net#KeepAliveConfig) is used instead of `KeepAlive`. However, this made disabling keep alive impossible, as a negative value of `Interval` will just left the underlying socket option unchanged.

For now, disable keep alive when KeepAliveInterval is negative, when it's zero, a system dependent default is chosen automatically.

### DIFF
--- a/modules/caddyhttp/app.go
+++ b/modules/caddyhttp/app.go
@@ -533,7 +533,7 @@ func (app *App) Start() error {
 					// create the listener for this socket
 					lnAny, err := listenAddr.Listen(app.ctx, portOffset, net.ListenConfig{
 						KeepAliveConfig: net.KeepAliveConfig{
-							Enable:   srv.KeepAliveInterval != 0,
+							Enable:   srv.KeepAliveInterval >= 0,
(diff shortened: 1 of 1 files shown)


WeidiDeng:
In [7157](https://github.com/caddyserver/caddy/pull/7157), whether `net.KeepAliveConfig.Enable` is true depend on if the `ProbeInterval` is positive because in that case, we use [`net.Dialer`](https://pkg.go.dev/net#Dialer) which willl handle keep alive for us. Namely, if `net.KeepAliveConfig.Enable` is false and `KeepAlive` is zero, a default value will be chosen. Though reverse proxy will use 30 seconds if not configured explicitly.

When listening, sometimes we don't use [`net.ListenConfig`](https://pkg.go.dev/net#ListenConfig) to listen, for example, socket activation. So keep alive is enabled by default as well.

mohammed90 (approved):
Good catch!

Issue #7096, opened 2025-06-27
[file_server] Caddy is not verifying etags from etag files


Pizmovc:
**Some background**
I'm using NixOS as my server OS with Caddy serving my blog and other sites. I've set it up so that it is serving the files directly from the Nix store. What is important to know here is that when building with Nix it strips all timestamps from files, so no modification times are present (found a [somewhat related issue](https://github.com/caddyserver/caddy/issues/5734)). So since Caddy can't use modification times to respond to `If-Modified-Since` header, since files are seemingly never modified, it always responds with `304 Not Modified`, breaking caching for my blog. So far Caddy is behaving as expected.

So I've modified my Caddyfile to force Caddy to use ETags with the following config:

```Caddyfile
luka.korosec.cc {
	header {
		-Last-Modified
	}
	file_server {
		etag_file_extensions .etag
	}
	root * /nix/store/3gsfndbnp4v4z0k7cs12b82vqvvpqxr7-tech-blog-1.0.0
	encode gzip
}
```

I've removed the `Last-Modified` header (so browsers can't use `If-Modified-Since` header), and added `etag_file_extensions`. I generate the ETag files during Nix build and Caddy correctly uses them as `etag` header. Up till now, I can't fault Caddy for anything, but what follows is a Caddy bug I think.

**The bug**
Here is what I think shows a bug:
1. Make initial request: `curl -v https://luka.korosec.cc`
1. Note the ETag in response: `etag: b79e32c68bbcbbf5599c452748e903f6`
1. Make conditional request: `curl -v -H "If-None-Match: b79e32c68bbcbbf5599c452748e903f6" https://luka.korosec.cc`
1. Observe that server returns `200 OK` instead of `304 Not Modified`

<details>
<summary>curl session</summary>

```shell
❯ curl -v https://luka.korosec.cc
[...]
> GET / HTTP/2
> Host: luka.korosec.cc
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/2 200
< accept-ranges: bytes
< alt-svc: h3=":443"; ma=2592000
< content-type: text/html; charset=utf-8
< etag: b79e32c68bbcbbf5599c452748e903f6
< server: Caddy
< vary: Accept-Encoding
< content-length: 1485
< date: Fri, 27 Jun 2025 19:07:19 GMT
<
<!DOCTYPE html>
[...]

[...]

PR #7173, opened 2025-08-12
doc: Add a few lines about Etag file content

Pizmovc:
Adds a bit more details to the [etag_file_extensions](https://caddyserver.com/docs/json/apps/http/servers/routes/handle/file_server/etag_file_extensions/) field, specifically how the content of the Etag files should look like.

This comes mainly from my frustration with getting it right, as can be seen in #7096.


mholt (approved):
Thanks! Good clarification.

Issue #7175, opened 2025-08-13
http: listeners won't be closed if caddy fails to bind some of the addresses


WeidiDeng:
It's possible to specify multiple addresses for caddy to bind using both Caddyfile and json, however caddy won't release those sockets if one of them fails to bind.

Consider the following Caddyfile,

```
{
    # debug
	skip_install_trust
	local_certs
    grace_period 10s
    servers :5001 {
        protocols h1 h2
    }
}

localhost:5001 {
}
```

Let's first run a program that binds port 5002,

```
package main

import (
	"net"
	"time"
)

func main() {
	_, err := net.Listen("tcp", "0.0.0.0:5002")
	if err != nil {
		panic(err)
	}
	time.Sleep(time.Minute)
}
```

While it's running, change the caddy config to this:

```
{
    # debug
	skip_install_trust
	local_certs
    grace_period 10s
    servers :5001 {
        protocols h1 h2
    }
}

localhost:5001 localhost:5003 localhost:5004 localhost:5002 {
}
```

caddy will refuse to use the new config with `"error":"loading config: loading new config: http app module: start: listening on :5002: listen tcp :5002: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted."`

Now it's impossible to bind to 5003 and 5004 because caddy uses them. Can be tested by

```
package main

import (
	"net"
	"time"
)

func main() {
	_, err := net.Listen("tcp", "0.0.0.0:5003")
	if err != nil {
		panic(err)
	}
	time.Sleep(time.Minute)
}
```
```

PR #7176, opened 2025-08-13
http: clean up listeners if some of the listeners fail to bind

WeidiDeng:
Fix [7175](https://github.com/caddyserver/caddy/issues/7175).


WeidiDeng:
They basically do the same for `http` app, but in different stages of a module life cycle. `Stop` is called when an app started successfully, `Cleanup` is called whether the app started successfully or not.

Issue #7172, opened 2025-08-12
http: http3 will leak quic listener resources


WeidiDeng:
In the release note for quic-go [v0.52.0](https://github.com/quic-go/quic-go/releases/tag/v0.52.0), it's stated that

>QUIC listeners created by the HTTP/3 server (i.e. when using http3.Server.ListenAndServe and ListenAndServeTLS) are immediately closed (https://github.com/quic-go/quic-go/pull/5101). QUIC listeners created by the application (i.e. when using http3.Server.ServeListener) are left running, it is the application's responsibility to close them, or use them in a new server instance (https://github.com/quic-go/quic-go/pull/5129).

Right now, caddy assumes that http3 server will close these listeners as before. A fix is under way.

mholt:
Nice find, thank you Weidi!

PR #7177, opened 2025-08-13
http: free up quic listener when stopping

WeidiDeng:
Fix [7172](https://github.com/caddyserver/caddy/issues/7172).

### DIFF
--- a/listeners.go
+++ b/listeners.go
@@ -430,6 +430,7 @@ func JoinNetworkAddress(network, host, port string) string {
 // address instead.
 //
 // NOTE: This API is EXPERIMENTAL and may be changed or removed.
+// NOTE: user should close the returned listener twice, once to stop accepting new connections, the second time to free up the packet conn.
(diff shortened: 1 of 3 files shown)


vnxme:
Hi, does it also help to implement UDP wrappers the way TCP ones (i.e., listener wrappers) are implemented?

WeidiDeng:
No, this only frees up resources, not adds new interfaces to facilitate udp multiplexing.

WeidiDeng:
The double Close is weird, but I don't want to introduce a new method.

In the same version that quic-go stopped closing user provided listeners, closed listeners [won't](https://github.com/quic-go/quic-go/pull/5284#event-19138787663) affect accepted connections. It used to be that closing a listener will close all the connections as well. 

Yes, it's how quic connections work, all the quic connections are user space implementation of a udp socket multiplexer. So quic-go never closes the udp socket even when the quic listener is closed.

Caddy needs to close the udp socket. As documented, the first Close is to stop new connections, the second is to release packet conns.

mholt:
Gotcha, makes sense. I think we can give this a try!

PR #7180, opened 2025-08-13
multiplexing: Introduce packet conn wrappers

vnxme:
This PR introduces `packet_conn_wrappers`. They are designed to work for packet connections (i.e., UDP) the same way `listener_wrappers` are applicable to listeners (i.e., TCP).

At the moment, there are no modules that implement this interface, but it's a matter of time and effort. The first one could be `layer4`. Unlike `tls`, there is no placeholder wrapper, i.e. all wrappers come before the QUIC handshake.

This change expands Caddy's multiplexing capabilities (e.g., allow for serving HTTP/3 and anything UDP-based like WireGuard, OpenVPN, DNS, etc. simultaneously on the same port).


listeners.go
```diff
@@ -431,7 +431,7 @@ func JoinNetworkAddress(network, host, port string) string {
 //
 // NOTE: This API is EXPERIMENTAL and may be changed or removed.
 // NOTE: user should close the returned listener twice, once to stop accepting new connections, the second time to free up the packet conn.
-func (na NetworkAddress) ListenQUIC(ctx context.Context, portOffset uint, config net.ListenConfig, tlsConf *tls.Config) (http3.QUICListener, error) {
+func (na NetworkAddress) ListenQUIC(ctx context.Context, portOffset uint, config net.ListenConfig, tlsConf *tls.Config, pcWrappers []PacketConnWrapper) (http3.QUICListener, error) {
```
WeidiDeng, line 514: I would use variadic parameters `...PacketConnWrapper` so that callers won't have to change anything if no packet conn wrappers are present.

vnxme: I have nothing against your suggestion, But I've [searched](https://github.com/search?q=repo%3Acaddyserver%2Fcaddy+ListenQUIC&type=code) for this function callers and found there is only one. It's also mentioned that this function is experimental and subject to change. Could you please elaborate on the callers your are referring to?

WeidiDeng: In caddy core, there is only this one. But who knows what plugin authors will do?

vnxme: I can see your point and sorry for being tedious, but I've [searched](https://github.com/search?q=caddy.ListenQUIC&type=code) again, and found 23 mentions of `caddy.ListenQUIC` in the code with all of them being outdated Caddy forks. So there are virtually no plugins using this function yet. Why shall we take care of anything that doesn't even exist?

WeidiDeng: I'm not requesting a change, just voice my opinion. It's not taking care of something, it's defensive programming.

mholt:
Thanks for working on this @vnxme ! I will get back to this soon, I just need to tag a new Caddy release first.

vnxme:
Do you want we to make any changes to this PR? Are there any thoughts why it shouldn’t be merged?

PR #470, opened 2015-12-30
Replacer patch

abiosoft:
1. New values `query_escaped`, `uri_escaped`, `path_escaped` for the query escaped versions of the value to make them compatible and usable as query params.

2. Ability to add custom values to a `Replacer`.
```
replacer := middleware.NewReplacer(...)
replacer.Set("custom_key", "custom_value") // can now replace {custom_key} 
```

PR #7181, opened 2025-08-13
caddyhttp: add replacer placeholders for escaped values

Qusic:
basically this brings back this v1 change #470

adding  `uri_escaped`, `path_escaped`, `query_escaped` being the `url.QueryEscape`d version of original values

useful for such cases: https://caddy.community/t/how-to-percent-encode-request-url-when-redirecting/30982


mholt:
Thanks, I was actually curious though, should we do something like `{%path}`? That almost makes it clearer that it's URI escaping in particular. (as opposed to `\` escapes or something).

francislavoie:
@mholt I agree for the shortcut, but the long form should probably just be plain English to make sure we don't run into issues with regex character sets not including those special characters etc.

mholt:
That's a good point.

caddyconfig/httpcaddyfile/shorthands.go
```diff
@@ -64,10 +64,13 @@ func placeholderShorthands() []string {
 		"{orig_?query}", "{http.request.orig_uri.prefixed_query}",
 		"{method}", "{http.request.method}",
 		"{uri}", "{http.request.uri}",
+		"{uri_escaped}", "{http.request.uri_escaped}",
```
mholt:
```suggestion
		"{%uri}", "{http.request.uri_escaped}",
```

And so on for the other shorthands

Qusic: updated. also existing ones are `query` and `prefixed_query`. do you think we should use `escaped_uri` instead of `uri_escaped` for a bit of consistency?

mholt: Sigh... I actually like the `uri_*` ones since it makes it easier to find alphabetically in a list, but, I do see your point too.

Let's do `uri_escaped` regardless, I think it's better that way personally.

Qusic: ok. then i think it's ready to merge now.

